PR #7637, opened 2020-09-23
feat: bump TypeScript to 4.0.3

trivikr:
TypeScript 4.0.3 release notes https://github.com/microsoft/TypeScript/releases/tag/v4.0.3
TypeScript upgrade steps followed from https://github.com/denoland/deno/blob/master/cli/dts/README.md

PR #7638, opened 2020-09-23
fix: ts upgrade instructions to exclude some .d.ts files

trivikr:
Refs: https://github.com/denoland/deno/pull/7637#pullrequestreview-494013457

StackOverflow discussion suggests using rsync to exclude files while copying https://stackoverflow.com/a/14789400/477683


kitsonk (approved):
LGTM

Issue #7551, opened 2020-09-18
Inspect could be better for symbols


solson:
Here's a look at the current handling of `Symbol`s in `Deno.inspect`.

**1.** Neither Node nor Deno quote symbol names with unprintable characters, but both do escaping in object keys:

Deno

```console
> ({ [Symbol("foo\n\0bar")]: 1 })
{ Symbol(foo\n\x00bar): 1 }
```

Node

```console
> ({ [Symbol("foo\n\0bar")]: 1 })
{ [Symbol(foo\n\u0000bar)]: 1 }
```

**2.** However, note that Node puts `[]` around symbol keys, which is clearer in this contrived example:

Deno

```console
> ({ [Symbol()]: 1, 'Symbol()': 2 })
{ Symbol(): 2, Symbol(): 1 }
```

Node

```console
> ({ [Symbol()]: 1, 'Symbol()': 2 })
{ 'Symbol()': 2, [Symbol()]: 1 }
```

**3.** Lastly, while inspecting a symbol *directly*, neither Node nor Deno even do escaping (output is the same for both):

```console
> Symbol("foo\n\0bar")
Symbol(foo
bar)
```

### To do
  * [ ] 1. It seems okay not to quote symbol names, but maybe we should?
  * [ ] 2. I think we should add `[]` for object symbol keys like Node.
  * [ ] 3. When inspecting symbols directly we should at least do escaping — make it identical to the output for symbols in object keys (aside from the `[]`).

cc @caspervonb

caspervonb:
Excellent digging 😁 

So for strings, browser behavior was also considered; e.g console.log("%o", "foo") does result in a quoted string in the developer console across browsers.

1. Browsers disagree on this, Chrome does quote this (`{"Symbol(foo↵bar)": 1}`), Safari does not and did not try Firefox. All the characters are selectable tho so if you select it you do get the \0 in your clipboard; the CLI equivalent of that would be to quote and escape. An interesting thing about Chrome tho if you do a well-known symbol like `{Symbol(Symbol.iterator): 1}` it doesn't quote.
2. Not sure, probably as it makes for a valid computed key.
3. Should reflect 1, yes.

solson:
The browsers mostly make me think "we can do better", so I hope we're not *too* constrained by what they tend to do. (Of course, we should care about compatibility, but this pretty-printing stuff seems drastically incompatible between implementations already.)

Firefox:

[...]

PR #7641, opened 2020-09-23
fix(cli/console): quote non-alphanumeric symbols

caspervonb:
This quotes and escapes symbol descriptions that contains characters outside of the basic alpha-numeric identifier range.

Fixes an issue brought up in #7551


ry (approved):
LGTM 👍

PR #7642, opened 2020-09-23
fix(cli/console): enclose symbol keys in brackets

caspervonb:
This encloses symbol keys when used in objects with brackets (e.g `[Symbol("Symbol.iterator")]`).

Fixes an issue brought up in #7551


bartlomieju (approved):
LGTM

PR #7643, opened 2020-09-23
fix(std/fs): mark createWalkEntry(Sync) as internal

### LINKED ISSUE #7572: createWalkEntry and createWalkEntrySync should probably not be exported from std/fs
There's two functions exported in std/fs/walk.ts (`createWalkEntry` and `createWalkEntrySync`) that don't make much sense to keep public and seem to be just exported to share internally between modules.

caspervonb:
This marks createWalkEntry and createWalkEntry in std/fs as internal via an underscore prefix.
These were showing up on the docs website.

Fixes #7572

### DIFF
--- a/std/fs/expand_glob.ts
+++ b/std/fs/expand_glob.ts
@@ -10,8 +10,8 @@ import {
 } from "../path/mod.ts";
 import {
   WalkEntry,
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

PR #7646, opened 2020-09-23
fix(lib.deno.shared_globals): Change Console class to interface

### LINKED ISSUE #7494: Change Console class to interface in types
https://github.com/denoland/deno/blob/v1.4.0/cli/dts/lib.deno.shared_globals.d.ts#L572-L588
->
https://github.com/denoland/deno/blob/v1.4.0/cli/dts/lib.dom.d.ts#L18972-L18995

The declaration is left over from when we concretely exported this class.

nayeemrmn:
Fixes #7494.


kitsonk (approved):
LGTM

PR #7647, opened 2020-09-23
feat(cli/coverage): add json coverage reporter

caspervonb:
This adds a json coverage reporter which dumps the script coverage as read from the runtime along with the source the runtime saw it for accurate reconstruction.

(WIP; couple of moving parts in flight at the same time so ignore for now)


caspervonb:
Postponing til after cross-process coverage.

Issue #7639, opened 2020-09-23
Deno.readLink path parameter is string only (not string | URL like Deno.readFile)


callionica:
There's inconsistency in the Deno functions that deal with paths. Specifically `Deno.readLink` accepts a file path only as a `string`, whereas `Deno.readFile` accepts `string | URL`. It would be good if readLink also accepted URLs.

wperron:
is anyone working on this one? if not I can look at it.

bartlomieju:
@wperron I don't think anyone's working on this. Please do look at it

PR #7651, opened 2020-09-24
Deno.readLink and Deno.readLinkSync URL accepting

ohadkoren:
Resolving issue #7639.

As of now, `Deno.readLink` and `Deno.readLinkSync` are not accepting `URL` as a path parameter - `string` only.
This easy PR is making consistency between `Deno` functions that deal with paths.


ry (commented):
Thanks for the patch. Currently this is not sufficient - it's only updating the type definitions. You also have to modify the functions in `cli/rt/30_fs.js`  to use `pathFromURL`. This patch should also contain a test demonstrating these functions being called with a file URL.

Issue #7473, opened 2020-09-14
Refactor inspector server and client


bartlomieju:
After landing #6901 we're using v8 Inspector to collect coverage. 

It uses the same logic as `--inspect`/`--inspect-brk` flag which spawns a WS server.

We shouldn't be creating WS server for `--coverage` so we need to split 
inspector from server part.

CC @caspervonb

caspervonb:
Not entirely sure what approach we want to take here, we could split the inspector into an optional "remote" part.

```rust
struct Inspector {
  // Kept as is today
  v8_inspector_client: v8::inspector::V8InspectorClientBase,
  v8_inspector: v8::UniqueRef<v8::inspector::V8Inspector>,

  // This would be the sessions, websocket url, waker, et cetera stuck into a "context" object.
  pub remote: Option<RemoteContext>,

  // A local session for sending messages on the same thread.
  pub session: LocalSession,
}
```

Alternatively, we just leave the inspector as is (possibly prefix it with Remote) and introduce a second type of inspector dubbed local inspectors with no baggage.

```rust
struct LocalInspector {
  v8_inspector_client: v8::inspector::V8InspectorClientBase,
  v8_inspector: v8::UniqueRef<v8::inspector::V8Inspector>,

  // Again, a local session for sending messages on the same thread.
  session: LocalSession,
}
```

I'm leaning slightly towards the first as I'm not sure what implications two inspectors on a single isolate would have if any.

Any thoughts, @piscisaureus?

piscisaureus:
I'd suggest to add an `Option<Arc<InspectorServer>>` or something, and make sure that `InspectorServer` stops the server thread in its Drop handler.

bartlomieju:
Done in #7584 #7577 #7656 and #7718

PR #7656, opened 2020-09-24
refactor(cli/inspector): make server optional

caspervonb:
This makes inspector registration with the server with optional and explicit to allow for inspectors to exist without spinning up the server.

As a side effect of the server being explicitly passed around it also makes it possible to bind multiple servers (Not likely we'll ever need that, but it is possible).

### DIFF
--- a/cli/worker.rs
+++ b/cli/worker.rs
@@ -3,6 +3,7 @@
 use crate::fmt_errors::JsError;
 use crate::global_state::GlobalState;
 use crate::inspector::DenoInspector;
+use crate::inspector::InspectorServer;
 use crate::js;
 use crate::metrics::Metrics;
 use crate::ops;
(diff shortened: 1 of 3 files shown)


caspervonb:
Ties into #7473

bartlomieju:
@caspervonb so `--coverage` will not print `Debugger listening on...` message?

caspervonb:
> @caspervonb so --coverage will not print Debugger listening on... message?

In a follow up will, this only makes the two separable but doesn't change the behaviour of current flags configuration/parsing.

cli/worker.rs
```diff
@@ -161,7 +162,12 @@ impl Worker {
         .inspect
         .or(global_state.flags.inspect_brk)
         .filter(|_| !is_internal)
-        .map(|inspector_host| DenoInspector::new(&mut isolate, inspector_host))
+        .map(|inspector_host| {
+          DenoInspector::new(
+            &mut isolate,
+            Some(Arc::new(InspectorServer::new(inspector_host))),
+          )
+        })
```
bartlomieju, line 170: This call will result in multiple servers being created

bartlomieju (changes requested):

bartlomieju (approved):
LGTM, nice cleanup @caspervonb

Issue #6829, opened 2020-07-21
[Discussion] std/path: rename methods


timreichen:
I came across a naming inconsistency. I know that the naming convention for some of these methods has historic reasons,  but in ```JS``` there are normally no abbreviations for functions and methods.

Affected exports are
- ```sep```
- ```extname(path: string)```
- ```dirname(path: string)```

So I wonder if it would make sense to rename them to have full descriptive names. Maybe something like this:

- ```separator```
- ```extension(path: string)```
- ```directory(path: string)```

What do you think?

nayeemrmn:
- `sep` -> `separator`
- `extname(path: string)` -> `getExtension(path: string)`
- `dirname(path: string)` -> `getParent(path: string)`
- `basename(path: string)` -> `getFileName(path: string)`

These are taken from [Rust](https://doc.rust-lang.org/std/path/struct.Path.html). Other than the `get` prefixes, that is... so I guess you can fight me there.

caspervonb:
So these are well known names that you'll find in the std libraries of pretty much any language or runtime including but not limited to:
- C
- C++
- Rust
- Rython
- Puby
- Node

Go is slightly different but variation is small enough that there's no need to read the docs, e.g dirname is `filepath.Dir`.

nayeemrmn:
As far as I can remember, we usually favour Rust names in these situations. They're more modern and consistent with each other. The module system they are delivered through is more similar to JS's. It's especially easy when the Unix names are so terrible. A path has a parent, not a directory. And maybe we _shouldn't_ arbitrarily violate camel case?

caspervonb:
Hmm based on the Deno namespace seems to have more POSIX influence, but I digress.

My point was just that basename, extname etc have well established prior art and with that comes intrinsic knowledge associated with those names, proposed names do not.

Renaming sep to separator seems appropriate tho.

nayeemrmn:
> with that comes intrinsic knowledge associated with those names, proposed names do not.

A name which is direct and constructed logically is inherently familiar. For example, no one can be reasonably surprised by [`path.extension()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.extension). There is nothing opinionated about its construction, like an abbreviation or a meaningless suffix. No misuse of terminology. The casing is consistent with the rest of the lib; we won't end up with `basename()` and [`fileStem()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem) side-by-side.

[...]

PR #6897, opened 2020-07-27
BREAKING(std/path): Rename some symbols

nayeemrmn:
Uses Rust names for functions.
- `basename()` -> `fileName()`
- `dirname()` -> `parent()`
- `extname()` -> `extension()`
- `sep` -> `separator`

Closes #6829.

Refs:
- https://doc.rust-lang.org/std/path/struct.Path.html
- https://en.cppreference.com/w/cpp/filesystem/path

PR #7657, opened 2020-09-24
BREAKING(std/path): Split basename() into fileName() and fileStem()

nayeemrmn:
- fix(std/node/path): Don't re-export `fromFileUrl()`

---

As I mentioned in #6829, the current specification of `basename()` (get the base file name and optionally strip a suffix if it's present) is unjustifiably silly and should not be stabilised on. I've split it into `fileName()` (get the base file name) and `fileStem()` (get the base file name without the extension) based on Rust APIs. Not every path has a base file name, so it can return null. This will make code safer than returning an empty string.


AviVahl:
It looks like the naming in the path module followed the same API as Node.js's `path`. Getting it consistent with Rust is nice, but having it consistent with Node might prove more useful in the long run (for developers migrating between the two ecosystems).

I personally find `basename` to make more sense in those cases where I have a `/path/to/some-directory` and I want to parse out `some-directory`. Having to call `fileName` on a path I know points to a directory feels unintuitive, even though I realize the same path could have pointed to a file.

nayeemrmn:
We already have `FileInfo`, so I'm not too concerned about `file` vs `base`. Anyway, that's for #6897. This is less about the name and more about the signature/behaviour.

bartlomieju:
I'm in favor of landing this PR, @nayeemrmn could you rebase the branch when you get a moment?

lucacasonato:
Putting this on hold until this is discussed in std working group.

bartlomieju:
@nayeemrmn we won't land this PR before standard library migration to a separate repository. I'm still very much interested in discussing these changes on upcoming meetings and potentially landing the PR. 

Let's keep it open until the migration is done.

bartlomieju:

PR #7659, opened 2020-09-24
Fix.  Replaced legacy chrome-devtools:// scheme.

june07:
The legacy chrome-devtools:// scheme was removed from the Chromium codebase.  The new scheme is simply "devtools://"

https://chromium-review.googlesource.com/c/chromium/src/+/2088335
https://chromium.googlesource.com/chromium/src/+/6700d12448f76712c62a6d2372a95b97a26d4779


ry (approved):
thanks

PR #7664, opened 2020-09-25
feat(core): re-export useful parts of anyhow

kitsonk:
In order to use `.context()` with a result, you need to have `anyhow::Context` in scope, in addition, `anyhow::Result` and `anyhow::bail` are super useful when using anyhow.


bartlomieju (approved):
LGTM, although I want to note that I am not a fan of `Result` aliases

kitsonk:
To be introduced in the next iteration of the compiler rewrite, so closing without merge.

PR #7461, opened 2020-09-14
feat(cli): Use swc for deno bundle

kdy1:
To ease testing, I added codes to `cli/tsc.rs`.
I'll move codes to relevant module after some testing.

---

Output of `deno bundle https://deno.land/std@0.69.0/http/server.ts`: https://gist.github.com/kdy1/fee62b5f2e6dcc4639e2f16277978ccd

Timings are:

`target/debug/deno bundle 'https://deno.land/std@0.69.0/http/server.ts'  0.24s user 0.02s system 96% cpu 0.266 total`
`target/release/deno bundle 'https://deno.land/std@0.69.0/http/server.ts'  0.03s user 0.01s system 80% cpu 0.052 total`

--- 

<del>I found another bug while testing it on other std modules. Fix seems simple, so I'll fix it asap. </del>

Fixed with https://github.com/swc-project/swc/pull/1083.

---

<del>
I found a bug of `swc_bundler`, and working on it at https://github.com/swc-project/swc/pull/1075 (Circular import bug)
</del>

---

<del> Sadly I found some more bugs. </del>

PR #7669, opened 2020-09-25
[WIP] feat(cli): bundling using swc

kitsonk:
This is very much still a WIP, but I have it working.

I adapted @kdy1's #7461 into the refactor of the module graph...  I need to DRY up the code a bit, as well as provide type checking variant, as well are re-organise the code in the `main.rs` that shouldn't really be there.  It does demonstrate how incongruent we handle stuff though, and all of this needs a really good scrub.

Raising it now for review and discussion.


kdy1:
What should I do?
Maybe rebase?

kitsonk:
@kdy1 nothing specifically at the moment other than thoughts and feedback... I need to add you as a co-author, and I will add you as a collaborator to my branch, but we were talking about it and because we were refactoring the whole of the compiler/module_graph/etc. it was easier for me to take your other PR and incorporate it.

Obviously the bugs you were working on in swc are invaluable too... I am going to try to get more tests in, so we have more examples to validate things.  I think we should really try to get it integrated and just try to fix whatever bugs we find.

crypt096 (approved):

kdy1:
Is stack overflow still reproducible?

kdy1:
@kitsonk I confused passes. You have to apply `swc_ecma_transforms::proposals::export` to use `export * as foo from './foo';`.

Sorry for misleading. (I thought it is handled by typescrpit::strip, but it was not)



Also, stack overflow of `example2.ts` will be fixed by https://github.com/swc-project/swc/pull/1127

kitsonk:
This has improved a lot (thanks @kdy1)...  Still a couple challenges left:

- `deno bundle --no-check https://deno.land/x/oak@v6.2.0/examples/server.ts` does not emit a valid bundle.  I isolated the issue to https://github.com/swc-project/swc/issues/1138
- Once I manually fix the valid invalid emit, when I do `deno run --allow-net bundle.js` the executes without syntax errors, but the behaviour isn't as expected.  I haven't narrowed it down yet because it could be that I didn't catch all the locations where that the issue above is impacting the code.

kitsonk:
Going to reland this under a different PR.

PR #7670, opened 2020-09-25
fix(cli): customInspect works on functions

### LINKED ISSUE #7650: (Deno.customInspect) Symbol doesn't have an effect on functions
(If this is not a bug, I ask that it is at least documented.)

The `Deno.customInspect` Symbol supposed to be called by `console.log` and other output functions, if the object has the key.
I applied this to a function, yet it seems to have outright ignored it.

```ts
//@ts-nocheck

const temp = Object.assign(
	() => { }, {
		[Deno.customInspect]: () => "[ null, NaN, undefined, \"[Object object]\" ]"
	}
);

console.log(
	"%o\n%o\n%o\n%o",
	Deno.customInspect in temp, // true
	temp[Deno.customInspect], // function [Deno.customInspect]
	temp[Deno.customInspect](), // [ null, NaN, ... ]
	temp // [Function]; expected to be [ null, NaN, ... ]
);
```

*(Ran on Deno 1.1, if it matters)*

kitsonk:
Fixes #7650

### DIFF
--- a/cli/rt/02_console.js
+++ b/cli/rt/02_console.js
@@ -193,6 +193,11 @@
   }
 
   function inspectFunction(value, _ctx) {
+    if (customInspect in value && typeof value[customInspect] === "function") {
+      try {
+        return String(value[customInspect]());
+      } catch {}
+    }
     // Might be Function/AsyncFunction/GeneratorFunction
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

Issue #7649, opened 2020-09-23
Top-level `await` does not work with dynamic `await import()`


TooTallNate:
```ts
// mod.ts
const sleep = (n: number) => new Promise(r => setTimeout(r, n));

await sleep(100);

export default 1;
```

```ts
// run.ts

// Works
// import mod from './mod.ts';
// console.log(mod);

// Doesn't work:
//
//   error: Uncaught ReferenceError: default is not defined
const mod = await import('./mod.ts');
console.log(mod);

// Also does not work
//
// import('./mod.ts').then(mod => {
//   console.log(mod);
// });
```

Running:

```
$ deno run --allow-read run.ts
error: Uncaught ReferenceError: default is not defined
    at Function.keys (<anonymous>)
    at inspectRawObject (rt/02_console.js:675:31)
    at inspectObject (rt/02_console.js:778:14)
    at inspectValue (rt/02_console.js:429:16)
    at inspectArgs (rt/02_console.js:877:16)
    at Object.log (rt/02_console.js:918:9)
    at file:///Users/nrajlich/deno-bug/run.ts:8:9
```

Seems like the `import()` function is not properly waiting for the module to become resolved, so there is no `default` property on the object returned from `import()`.

Initially reported in: https://github.com/TooTallNate/vercel-deno/issues/40

lucacasonato:
Hey @MylesBorins - can you take a look at this issue? How are TLA and dynamic imports meant to interact? Should this example work?

MylesBorins:
I can confirm that one should be able to await a dynamic import, that was one of the motivating use cases in the original proposal (although not to be abused, please still use static import too!)

I can confirm that the same code, with slight modifications, works in node.js v14.12.0

kitsonk:
Yeah, I don't think it is TLA.  As @TooTallNate mentions, the module isn't ready when the promise is resolved even using the `.then()`.  Clearly we have something messed up in the module instantiation bit with v8, where we are prematurely resolving the promise.  I haven't looked in detail but if IIRC, module instantiation is two pass, and I suspect we are resolving the promise for the importing module, and allowing it to be instantiated when only the dependency analysis phase is completed and not waiting for the instantiation.

nayeemrmn:
> Yeah, I don't think it is TLA. As @TooTallNate mentions, the module isn't ready when the promise is resolved even using the `.then()`.

[...]

PR #7672, opened 2020-09-25
fix(core): module execution with top level await

bartlomieju:
Fixes https://github.com/denoland/deno/issues/7649

This commit fixes implementation of top level await in "deno_core".

Previously promise returned from module execution was ignored causing to execute
modules out-of-order. 

With this commit promise returned from module execution is stored on "JsRuntime"
and event loop is polled until the promise resolves.


core/runtime.rs
```diff
@@ -748,20 +852,30 @@ impl JsRuntime {
...
-    match status {
-      v8::ModuleStatus::Evaluated => Ok(()),
-      v8::ModuleStatus::Errored => {
-        let exception = module.get_exception();
-        exception_to_err_result(scope, exception)
-          .map_err(|err| attach_handle_to_error(scope, err, exception))
+    Ok(receiver)
+  }
+
+  pub async fn mod_evaluate(&mut self, id: ModuleId) -> Result<(), AnyError> {
```
ry, line 867: Is it odd that JsRuntime itself is a future, but it now has a method which also returns a future?

ry (commented):
Given that it fixes a bug, I think we should land this - awesome work.

core/runtime.rs
bartlomieju: @ry there is a precedent with `JsRuntime::load_module` which we've had for a couple months.

core/modules.rs
```diff
@@ -667,7 +674,7 @@ mod tests {
     let a_id_fut = runtime.load_module(&spec, None);
     let a_id = futures::executor::block_on(a_id_fut).expect("Failed to load");
 
-    runtime.mod_evaluate(a_id).unwrap();
+    futures::executor::block_on(runtime.mod_evaluate(a_id)).unwrap();
```
piscisaureus, line 677: The issue with this is the following:
Normally you can't block_on inside a block_on. Both tokio and futures crate will panic.
However the "futures" crate doesn't detect tokio's presence, so this works.
However the outer event loop may not make progress (depending on the kind of i/o, operating system etc.)

Correct me if this is not a case of nested block_on and I am mistaken.

bartlomieju: @piscisaureus you're right, but there's no nested `block_on`s. In `core/` we don't use `block_on` anywhere else than tests.

core/runtime.rs
ry: I guess what happens if you try to `tokio::select!` on both the result of `mod_evaluate` and the JsRuntime itself?

core/modules.rs
piscisaureus: I mean block_on or equivalent... [tokio_main] qualifies too I think...
You should be able to replace block_on with tokio's block_on, like [this](https://docs.rs/tokio/0.2.22/tokio/runtime/struct.Runtime.html#examples-2). If it doesn't panic we're golden.

core/runtime.rs
bartlomieju: `JsRuntime` would be polled twice (because it's polled once in `mod_evaluate`). I don't like `JsRuntime` being a future, there's only a single time that you want to run event loop to complication - in `main.rs`. I would prefer to have explicit `JsRuntime::poll_event_loop()` instead.

core/modules.rs
bartlomieju: I'm still not sure what's in question here. 

Using `tokio` works as well:
```rust
let rt = tokio::runtime::Runtime::new().unwrap();
let a_id = rt.block_on(a_id_fut).expect("Failed to load");

rt.block_on(runtime.mod_evaluate(a_id)).unwrap();
```

piscisaureus: OK so no nesting. Awesome!

core/runtime.rs
piscisaureus: > What happens if you try to `tokio::select!` on both the result of `mod_evaluate` and the JsRuntime itself?

That either isn't or shouldn't be possible. 
You need a &mut reference to a Future to poll it, so you can't poll X and a container with X in it at the same time.

piscisaureus:
The strategy seems sound to me.
Also underlines the need for this monster to be refactored...

core/runtime.rs
```diff
@@ -716,6 +818,8 @@ impl JsRuntime {
       .expect("ModuleInfo not found");
     let mut status = module.get_status();
 
+    let (sender, receiver) = mpsc::channel(1);
```
piscisaureus, line 821: * Using a channel within a thread ... not my favorite pattern ever.
* Wouldn't a Oneshot work just as well?

bartlomieju: > * Using a channel within a thread ... not my favorite pattern ever.

That's the only solution I came up with.

> * Wouldn't a Oneshot work just as well?

I tried, but couldn't get it past borrow checker on `sender` not being clonable. I'll try to address this in the next pass.

core/runtime.rs
bartlomieju: Actually @piscisaureus is right:
```rust
  let fut = self.isolate.mod_evaluate(id);
    tokio::select! {
      _ = fut => {
        // 
        Ok(())
      }
      _ = self.isolate => {
        //
        Ok(())
      }
    }
```

```
error[E0507]: cannot move out of `self.isolate` which is behind a mutable reference
   --> cli/worker.rs:198:11
    |
198 |       _ = self.isolate => {
    |           ^^^^^^^^^^^^ move occurs because `self.isolate` has type `deno_core::JsRuntime`, which does not implement the `Copy` trait

error[E0505]: cannot move out of `self.isolate` because it is borrowed
   --> cli/worker.rs:198:11
    |
192 |       let fut = self.isolate.mod_evaluate(id);
    |                 ------------ borrow of `self.isolate` occurs here
193 | /     tokio::select! {
194 | |       _ = fut => {
195 | |         //
196 | |         Ok(())
197 | |       }
198 | |       _ = self.isolate => {
    | |           ^^^^^^^^^^^^ move out of `self.isolate` occurs here
...   |
201 | |       }
202 | |     }
    | |_____- borrow later used here
```

core/runtime.rs
piscisaureus: Yeah I don't have any suggestions either...

piscisaureus (approved):
LGTM

A skilful hack; also great to accumulate all these tests.

PR #7673, opened 2020-09-25
refactor: remove tsc/40_error_stack.js

bartlomieju:
This commit removes cli/tsc/40_error_stack.js as it is not
needed in TSC host. All errors originating in TSC are terminal
and don't require source mapping hence we can rely on default
stack traces provided by deno_core.

Additionally tsc/06_util.js was removed and its code moved
to tsc/99_main_compiler.js


ry (approved):
LGTM - I love removing code

kitsonk (approved):
LGTM

PR #7675, opened 2020-09-25
(std/datetime): add timezone to test date strings

### LINKED ISSUE #7665: (std/datetime): dayOfYear returns different values depending on the time zone
I cannot pass std tests because of this issue

---

js file:
```javascript
import * as datetime from "https://deno.land/std@0.70.0/datetime/mod.ts"

console.log(datetime.dayOfYear(new Date("1582-10-04T03:24:00+0000")));
console.log(datetime.dayOfYear(new Date("1582-10-04T03:24:00+0900")));
```
output:
```
277
276
```

timreichen:
fixes #7665

### DIFF
--- a/std/datetime/test.ts
+++ b/std/datetime/test.ts
@@ -158,79 +158,127 @@ Deno.test({
     // Test YearDay in several different scenarios
     // and corner cases
     // Non-leap-year tests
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - thanks

PR #7680, opened 2020-09-25
fix(cli/dts): Use var instead of let and const for globals

nayeemrmn:
Fixes type errors for `deno eval -T "globalThis.performance"`, etc.

Using `let` and `const` in declaration files treats the symbols as injected local variables. We don't want that for globals.


kitsonk (approved):
LGTM

PR #7671, opened 2020-09-25
test(std): unit test for async/delay

ochronus:
Added a unit test for the delay async function. Fixes https://github.com/denoland/deno/issues/7654

PR #7682, opened 2020-09-25
chore(std/async): remove redundant export in test

caspervonb:
This removes a redundant export in delay_test that was introduced in #7671


caspervonb:
Churn'ish commit but might as-well get it since I saw it go in.

ry (approved):

PR #7684, opened 2020-09-25
refactor(std/example): Inconsistencies in the example tests

tokiedokie:
Other `/std` tests have tests next to the script, but `/std/example` doesn't.


ry (approved):
LGTM - thanks

PR #7685, opened 2020-09-25
refactor(std): name test files for mod `mod_test`

tokiedokie:
Since `mod.ts` is not a "magic" file name, I think it's better to name test files for mod `mod_test.ts.`


kitsonk:
I prefer the existing `test.ts` convention.

With these stylistic changes, it is often better to raise them for discussion before creating a PR.

nayeemrmn:
`mod` doesn't make sense with a prefix or suffix, it's what we use as a default when there's no such detail to add.

crypt096 (approved):

bartlomieju:
@tokiedokie thanks for the PR, but I agree with @kitsonk that current convention is preferable. Closing without a merge.

PR #7688, opened 2020-09-26
simply github actions yaml

khs1994:
use strategy->matrix->include

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,15 +4,15 @@ on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


khs1994:
also, we can identify os type by `runner.os`

* https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#runner-context

ry (approved):
Thanks for the patch. I’m a little worried this is going to break in some subtle way- potentially not until the next release. (I vaguely remember setting it up this way purposefully...) But this is smaller and there are no comments as to why it might break. So let’s just move forward.

PR #7691, opened 2020-09-26
refactor: use JsRuntime to implement TSC

bartlomieju:
This commits removes "CompilerWorker" in favor of 
using "JsRuntime".

"cli/ops/compiler.rs" has been removed in favor of inline
registration of ops in "cli/tsc.rs"


cli/tsc.rs
```diff
@@ -1047,16 +956,56 @@ impl TsCompiler {
...
+  let mut js_runtime = JsRuntime::new(RuntimeOptions {
+    startup_snapshot: Some(js::compiler_isolate_init()),
+    ..Default::default()
+  });
+  {
+    let op_state = js_runtime.op_state();
+    let mut op_state = op_state.borrow_mut();
+    // TODO(bartlomieju): needed only for runtime_start op, which is needed
+    // only to pass "debugFlag" to compiler...
+    op_state.put(global_state);
```
bartlomieju, line 977: Want to try and address this one, before merging

ry (approved):
Awesome - nice work

bartlomieju:
I decided to remove `opNow` from compiler and instead use `new Date()` - I think 1ms resolution is precise enough for TSC and it greatly simplifies the setup.

PR #7693, opened 2020-09-26
refactor: combine MainWorker::new and MainWorker::create

bartlomieju:
Also removes "GlobalState.compiler_starts" - this field was only
used in 3 tests (which were completely trivial).


ry (approved):
👍

PR #7695, opened 2020-09-26
refactor: factor out check_unstable op helper

bartlomieju:
Cosmetic change to better contain usage of "GlobalState" in various ops

PR #7696, opened 2020-09-26
refactor: move op state registration to workers

bartlomieju:
Depends on https://github.com/denoland/deno/pull/7695


cli/web_worker.rs
```diff
@@ -1,375 +0,0 @@
...
-      let msg = json!("hi").to_string().into_boxed_str().into_boxed_bytes();
-      let r = handle.post_message(msg.clone());
-      assert!(r.is_ok());
-      let event = handle.get_event().await.unwrap();
-      assert!(event.is_none());
-      handle.sender.close_channel();
-    });
-    join_handle.join().expect("Failed to join worker thread");
-  }
-}
```
ry, line 375: This is desirable

cli/ops/fs.rs
```diff
@@ -30,80 +30,88 @@ use deno_core::error::generic_error;
 #[cfg(not(unix))]
 use deno_core::error::not_supported;
 
-pub fn init(rt: &mut deno_core::JsRuntime) {
-  super::reg_json_sync(rt, "op_open_sync", op_open_sync);
-  super::reg_json_async(rt, "op_open_async", op_open_async);
+pub fn init(state: &mut deno_core::OpState) {
+  super::reg_json_sync(state, "op_open_sync", op_open_sync);
+  super::reg_json_async(state, "op_open_async", op_open_async);
```
ry, line 35: These changes seem superfluous - what is the purpose?

 I like the init functions to take a `&mut JsRuntime` for maximum extensibility.

bartlomieju: I'm not so happy about this change either, however when I tried adding state elements next to call to `init()` I've hit `BorrowMutError` numerous times - because `rt.register_op()` mutably borrow `OpState`. It's quite cumbersome to register state that way without this change...

bartlomieju: I'll give it another go

ry (approved):
LGTM

PR #7702, opened 2020-09-26
refactor: Removed Deno.File constructor from types

jkomyno:
Remove `Deno.File` constructor from types, as it is just an implementation detail.
This closes denoland#7462.


kitsonk:
We shouldn't remove it. Closing without merge.

PR #7703, opened 2020-09-26
docs: add jsdoc for WebAssembly namespace

lucacasonato:
This is mostly taken from MDN. I added attribution to the top of the d.ts file.


ry (approved):
Thank you 👍

PR #7706, opened 2020-09-26
fix: net listen crashes on explicit undefined hostname

### LINKED ISSUE #7697: std/http: unexpected error when specifying "hostname" as undefined
When you set the `hostname` option for `serve` to `undefined`, it results in an unexpected error:

```
error: Uncaught InvalidData: data did not match any variant of untagged enum ArgsEnum
    at Object.jsonOpSync (core.js:247:13)
    at opListen (deno:cli/rt/30_net.js:32:17)
    at Object.listen (deno:cli/rt/30_net.js:207:17)
    at serve (server.ts:287:25)
    at index.ts:3:14
```

When you do not include the `hostname` option at all, the error does not happen. It only happens when you specifically set it to `undefined` yourself.

**Reproduction Code**

```ts
import { serve } from 'https://deno.land/std@0.70.0/http/server.ts';

let server = serve({ port: 8080, hostname: undefined });
for await (let req of server) {
    req.respond({ body: 'Hello World' });
}
```

**Additional Information**

```
deno 1.4.1
deno std 0.70.0
v8 8.7.75
typescript 4.0.2
Win 10 Pro
```

grosto:
fixes #7697

quite a small change and added test case too

### DIFF
--- a/cli/rt/30_net.js
+++ b/cli/rt/30_net.js
@@ -203,10 +203,10 @@
     }
   }
 
-  function listen(options) {
+  function listen({ hostname, ...options }) {
     const res = opListen({
       transport: "tcp",
-      hostname: "0.0.0.0",
+      hostname: typeof hostname === "undefined" ? "0.0.0.0" : hostname,
       ...options,
     });
 
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -533,3 +533,14 @@ unitTest(
     await resolvable;
   },
 );
+
+unitTest(
+  {
+    perms: { net: true },
+  },
+  function netExplicitUndefinedHostname() {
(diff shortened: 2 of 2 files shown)


cli/tests/integration_tests.rs
```diff
@@ -2121,6 +2121,11 @@ itest!(main_module {
   output: "main_module.ts.out",
 });
 
+itest!(net_explicit_undefined_hostname {
+  args: "run --allow-net net_explicit_undefined_hostname.js",
+  output: "net_explicit_undefined_hostname.js.out",
+});
+
```
lucacasonato, line 2128: Can you move this to a unit test instead of an integration test?

grosto: Fair. Did not even notice there were unit tests 😅

ry (approved):
LGTM - thanks @grosto

PR #7710, opened 2020-09-26
docs: update github doc links to use $CLI_VERSION

trivikr:
Ref: https://github.com/denoland/deno_website2/pull/1528

The issue with this fix is that the code links will not be browse-able from markdown files.


lucacasonato (approved):
Thanks

PR #7711, opened 2020-09-27
feat(fmt): Sort named import and export specifiers

### LINKED ISSUE #7150: deno fmt - Automatically sort named import and export specifiers?
I am working a new feature in dprint-plugin-typescript to sort import and export specifiers in alphabetical order.

For example, given the following:

```ts
import { Person, doWork, buildMap, Car } from "./file.ts";
```

It would do:

```ts
import { buildMap, Car, doWork, Person } from "./file.ts";
```

Should this be turned on for `deno fmt`?

**Case Insensitive Comparison or Case Sensitive?**

Also, should this be case insensitive or case sensitive? Above is case insensitive. Case sensitive looks like this:

```ts
import { Car, Person, buildMap, doWork } from "./file.ts";
```

It seems eslint will do case sensitive by default: https://eslint.org/docs/rules/sort-imports#ignorecase -- so that might be reason to make it case sensitive (also less work doing the comparison)

dsherret:
1. Sorts named import and export specifiers using case insensitive sorting.
2. Collapses import and export declarations to a single line if it can fit on one line.

Closes #7150

~~See #7712 for a demo.~~ - Never mind... needed to do everything in the same PR to make the tests pass.


bartlomieju:
@dsherret this is awesome, thanks so much for implementing this!

.dprintrc.json
```diff
@@ -30,7 +30,7 @@
     "third_party"
   ],
   "plugins": [
-    "https://plugins.dprint.dev/typescript-0.29.0.wasm",
+    "https://plugins.dprint.dev/typescript-0.32.1.wasm",
```
bartlomieju, line 33: No need to update `deno_third_party`?

bartlomieju (approved):

ry:
Nice! LGTM

.dprintrc.json
bartlomieju: Nvm, this is only a plugin

PR #7712, opened 2020-09-27
chore: Upgrade to dprint-plugin-typescript 0.32.1 internally

dsherret:
This shows #7711 in action.


std/http/server.ts
```diff
@@ -2,13 +2,13 @@
 import { encode } from "../encoding/utf8.ts";
 import { BufReader, BufWriter } from "../io/bufio.ts";
 import { assert } from "../_util/assert.ts";
-import { deferred, Deferred, MuxAsyncIterator } from "../async/mod.ts";
+import { Deferred, deferred, MuxAsyncIterator } from "../async/mod.ts";
```
dsherret, line 5: It becomes case sensitive in this scenario.

cli/tests/unit/body_test.ts
```diff
@@ -1,9 +1,5 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
-import {
-  unitTest,
-  assertEquals,
-  assert,
-} from "./test_util.ts";
+import { assert, assertEquals, unitTest } from "./test_util.ts";
```
dsherret, line 2: Example of collapsing to a single line.

dsherret:
Ok, this is a chicken and egg problem. I'm just going to do everything in the same PR.

PR #7714, opened 2020-09-27
BREAKING(std/encoding/csv): improve the definition of ParseOptions

### LINKED ISSUE #7426: [std/encoding/csv] Omit header row when specifying header options
It doesn't appear that this kind of behavior is documented: https://deno.land/std@0.68.0/encoding#csv

Example:

`data.csv`:
```csv
size,cost
m,12
s,10
l,14
```

`main.ts`:
```ts
import {parse} from 'https://deno.land/std@0.68.0/encoding/csv.ts';

const parseOptions = {
  header: [
    {
      name: 'size',
      parse: (string: string) => string.toUpperCase(),
    },
    {
      name: 'cost',
      parse: (string: string) => Number(string),
    },
  ],
};

const raw = await Deno.readTextFile('data.csv');

const result = await parse(raw, parseOptions);
console.log(result);
```

output:
```
% deno run --allow-read=. --allow-write=. main.ts
[
  { size: "SIZE", cost: NaN },
  { size: "M", cost: 12 },
  { size: "S", cost: 10 },
  { size: "L", cost: 14 }
]
```

As you can see, the resulting array includes the header row, but I can't figure out how to omit it when parsing using header options.

If this is not possible, please consider this a feature request. If I just missed it, please point me in the right direction. Thanks!

uki00a:
This PR resolves #7426.
cc @jsejcksn Please review.

## Changes

* Renames `ParseOptions.comma` to `ParseOptions.separator`.
* Makes `ParseOptions.header` accept only `boolean` and adds `ParseOptions.columns` to define the header definition.
* Renames `ParseOptions.header` to `ParseOptions.skipFirstRow`.

These change were made in response to feedback by @jsejcksn.

### DIFF
--- a/std/encoding/csv.ts
+++ b/std/encoding/csv.ts
@@ -343,10 +343,10 @@ export interface ColumnOptions {
 
 export interface ParseOptions extends ReadOptions {
   /**
-   * If you provide `header: true` and `columns`, the first line will be skipped.
-   * If you provide `header: true` but not `columns`, the first line will be skipped and used as header definitions.
+   * If you provide `skipFirstRow: true` and `columns`, the first line will be skipped.
+   * If you provide `skipFirstRow: true` but not `columns`, the first line will be skipped and used as header definitions.
    */
-  header?: boolean;
+  skipFirstRow?: boolean;
 
   /**
    * If you provide `string[]` or `ColumnOptions[]`, those names will be used for header definition.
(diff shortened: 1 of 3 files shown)


jsejcksn:
@uki00a What about `ParseOptions.skipFirstRow`? Ref: https://github.com/denoland/deno/issues/7426#issuecomment-699576200 That was the main reason I opened the issue.

std/encoding/csv.ts
```diff
@@ -343,10 +343,16 @@ export interface HeaderOptions {
 
 export interface ParseOptions extends ReadOptions {
   /**
-   * If a boolean is provided, the first line will be used as Header definitions.
-   * If `string[]` or `HeaderOptions[]` those names will be used for header definition.
+   * If you provide `header: true` and `columns`, the first line will be skipped.
+   * If you provide `header: true` but not `columns`, the first line will be skipped and used as header definitions.
    */
-  header: boolean | string[] | HeaderOptions[];
+  header?: boolean;
```
uki00a, line 349: @jsejcksn Is this correct?
```suggestion
  skipFirstRow?: boolean;
```

jsejcksn: LGTM

ry (approved):
LGTM

PR #7716, opened 2020-09-27
fix(cli/inspector): shutdown server gracefully on drop

caspervonb:
This signals a graceful shutdown of the inspector server and waits for the server thread to join when an InspectorServer is dropped.


ry (approved):
LGTM - nice clean up

PR #7718, opened 2020-09-27
fix(cli/test): do not start inspector server when collecting coverage

caspervonb:
This makes use of a coverage flag to create an inspector without a server for collecting code coverage.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -71,7 +71,6 @@ pub enum DenoSubcommand {
     allow_none: bool,
     include: Option<Vec<String>>,
     filter: Option<String>,
-    coverage: bool,
   },
   Types,
   Upgrade {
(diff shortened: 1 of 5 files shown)


cli/flags.rs
```diff
@@ -107,6 +107,7 @@ pub struct Flags {
   pub ca_file: Option<String>,
   pub cached_only: bool,
   pub config_path: Option<String>,
+  pub cover: bool,
```
bartlomieju, line 110: Drop `coverage` from `DenoSubcommand::Test` and replace it with this one?

cli/flags.rs
```diff
@@ -107,6 +106,7 @@ pub struct Flags {
...
```
bartlomieju, line 109: @caspervonb one more nitpick, I think `coverage` is more meaningful than `cover`

bartlomieju (approved):
LGTM, thanks @caspervonb

PR #7720, opened 2020-09-27
fix: no check recognizes require

### LINKED ISSUE #7719: --no-check recognizes `require`
```ts
// test.ts
require("fs");
```

```
>>> deno cache --no-check deno.ts 
error: relative import path "fs" not prefixed with / or ./ or ../ Imported from "file:///mnt/f9/Projects/github.com/denofn/denopack/test.ts"
```

We are recognizing `require` when using --no-check.

Likely related: #7621 

cc @kitsonk

lucacasonato:
Fixes #7719


ry (approved):
LGTM

PR #7726, opened 2020-09-27
fix(std/test): Ensures time tests run in UTC

j4qfrost:
This sets the TZ to UTC to resolve flakiness of format and weekOfYear tests in std. I also added parentheses to python tool prints for compatibility with python3.


tools/hash_benchmark.py
```diff
@@ -34,12 +34,12 @@ def run_benchmark(deno_exe, method, input_file):
         (out, _) = p.communicate()
 
         elapsed = out.split(':')[1].strip()
-        print "[{}] {}".format(alg, elapsed)
+        print("[{}] {}".format(alg, elapsed))
```
kitsonk, line 37: Is there a reason for these changes?  They seem to be unrelated to the intent of this PR.

j4qfrost: I added those changes in order to run with python 3 with my environment. They shouldn't be invasive towards the old workflow since it's compatible with python 2 also. Should this be in a separate PR?

bartlomieju: @j4qfrost yes, change to Python code should be done in a separate PR.

cli/tests/integration_tests.rs
```diff
@@ -9,6 +9,7 @@ use test_util as util;
 
 #[test]
 fn std_tests() {
+  std::env::set_var("TZ", "UTC");
```
bartlomieju, line 12: This sets `TZ` env variable in the current env - that's not desirable, instead use `.env()` on the subprocess like for `DENO_DIR` below

bartlomieju (changes requested):

j4qfrost:
Accidentally closed this.

tools/hash_benchmark.py
j4qfrost: https://github.com/denoland/deno/pull/7779

cli/tests/integration_tests.rs
```diff
@@ -13,6 +13,7 @@ fn std_tests() {
   let std_config = util::root_path().join("std/tsconfig_test.json");
   let status = util::deno_cmd()
     .env("DENO_DIR", dir.path())
+    .env("TZ", "UTC")
```
j4qfrost, line 16: @bartlomieju done.

bartlomieju:
Closing in favor of #8131

Issue #7487, opened 2020-09-15
missing JSDoc to exported functions in std


tokiedokie:
there is many exported functions in standard library but some functions' JSDocs are missing

here is the list of ts files need to be updated

- [ ] `encoding/hex.ts`
- [ ] `encoding/toml.ts`
- [ ] `fmt/colors.ts`
- [ ] `fmt/printf.ts`
- [ ] `fs/walk.ts`
- [ ] `http/file_server.ts`
- [ ] `io/streams.ts`
- [ ] `io/util.ts`
- [ ] `log/mod.ts`
- [ ] `node/querystring`
- [ ] `node/url.ts`
- [ ] `node/util.ts`
- [ ] `path.posix.ts`
- [ ] `path/win32.ts`
- [ ] `testing/assert.ts`
- [ ] `testing/diff.ts`
- [ ] `uuid/mod.ts`
- [ ] `uuid/v1.ts`
- [ ] `uuid/v4.ts`
- [ ] `uuid/v5.ts`
- [ ] `ws/mod.ts`

fuku710:
I want to challenge this issue.

bhumijgupta:
I'd like to add JSDoc docstrings for `node/url.ts`

starkensin:
I will take `node/util.ts`

crowlKats:
this is a dream issue for people who just want to complete hacktoberfest...

bhumijgupta:
According to the latest change in rule, a `hacktoberfest` tag should be present on the repo for it to be included in hacktoberfest. Deno does not have any such tag, so the contributions wont be counted towards it

DCtheTall:
I'd also be interested in participating in this for a Hacktoberfest PR if the owners of this repo are willing 😄

Luisbo85:
I will take `io/stream.ts`

mrmodise:
I will take `fmt/colors.ts` and `fmt/printf.ts`

mrmodise:
I will take `node/url.ts` and `node/util.ts` next. Thanks

ThenMorning:
I want to challenge this issue.
Take `ws/mod.ts`.

inokawa:
I'd like to take `fs/walk.ts`.

ThenMorning:
> I want to challenge this issue.
> Take `ws/mod.ts`.

I want to konw  is `there are many exported functions in standard library but some functions' JSDocs are missing` mean 
the doc which was missing is exist in  standard library?
[...]

PR #7735, opened 2020-09-28
docs(std/uuid): Added JSdocs for the std/uuid module

aakhtar3:
Added JSdocs for the std/uuid module requested by #7487


bartlomieju (approved):
Thanks @aakhtar3

PR #7741, opened 2020-09-28
feat(cli/console): allow noColor in Deno.inspect()

TTtie:
This PR introduces an ability to explicitly disable ANSI color output from `Deno.inspect(...)`:
```js
const object = { this: { is: { a: "object" } } };
Deno.inspect(object, { noColor: true }); // { this: { is: { a: "object" } } }

Deno.inspect(object); // assuming Deno.noColor === false; { this: { is: { a: \x1b[32m"object"\x1b[39m } } }
```

Things to consider/to do :
- [ ] Pick a sensible default for `noColor` - should the `Deno.inspect(...)` call return the inspected string with ANSI escape codes regardless of `Deno.noColor`?<br>Currently, `{ noColor: false }` respects the `Deno.noColor` value - maybe use `noColor = inspectOptions.noColor ?? globalThis.Deno?.noColor ?? true`?

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1954,6 +1954,8 @@ declare namespace Deno {
     iterableLimit?: number;
     /** Show a Proxy's target and handler. Defaults to false. */
     showProxy?: boolean;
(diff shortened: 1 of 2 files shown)


TTtie:
Superseded by #7742

PR #7516, opened 2020-09-16
fix(std/testing): sanitize escape sequences from assertion errors

caspervonb:

PR #7742, opened 2020-09-28
feat(cli/console): Add Deno.InspectOptions::colors

nayeemrmn:
On by default for `console.log()`, off by default for `Deno.inspect()`.

Ref https://github.com/denoland/deno/pull/7516#pullrequestreview-489567120.
Ref https://nodejs.org/docs/latest-v13.x/api/util.html#util_util_inspect_object_options

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1941,19 +1941,21 @@ declare namespace Deno {
   export function run<T extends RunOptions = RunOptions>(opt: T): Process<T>;
 
   export interface InspectOptions {
-    /** Traversal depth for nested objects. Defaults to 4. */
-    depth?: number;
-    /** Sort Object, Set and Map entries by key. Defaults to false. */
-    sorted?: boolean;
-    /** Add a trailing comma for multiline collections. Defaults to false. */
-    trailingComma?: boolean;
+    /** Stylize output with ANSI colors. Defaults to false. */
+    colors?: boolean;
     /** Try to fit more than one entry of a collection on the same line.
      * Defaults to true. */
     compact?: boolean;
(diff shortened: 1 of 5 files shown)


TTtie (approved):

cli/dts/lib.deno.ns.d.ts
```diff
@@ -1954,6 +1954,8 @@ declare namespace Deno {
     iterableLimit?: number;
     /** Show a Proxy's target and handler. Defaults to false. */
     showProxy?: boolean;
+    /** Stylize output with ANSI colors. Defaults to false. */
```
kitsonk, line 1957: I know we are totally inconsistent on this, but it would be nice to alphabetise properties when we have a chance instead of leaving it to some sort of arbitrary ordering.

cli/rt/02_console.js
```diff
@@ -414,6 +407,13 @@
         : inspectValue(proxyDetails[0], ctx, level, inspectOptions);
     }
 
+    const green = inspectOptions.colors ? colors.green : (s) => s;
```
kitsonk, line 410: wouldn't it be better to move all of this to something like

```js
function maybeColor(fn, flag = true) {
  return flag ? fn : (s) => s;
}
```

cli/rt/02_console.js
```diff
@@ -630,32 +632,39 @@
     );
   }
 
-  function inspectWeakSet() {
+  function inspectWeakSet(inspectOptions) {
+    const cyan = inspectOptions.colors ? colors.cyan : (s) => s;
```
kitsonk, line 636:
```suggestion
  function inspectWeakSet({ colors }) {
    const cyan = maybeColor(colors.cyan, colors);
```

kitsonk (approved):
LGTM! 🥳

Issue #7225, opened 2020-08-27
Refactoring how we compile code


kitsonk:
In #6894 there were a lot of changes that we need to make in a more gradual way.  This issue is intended to keep track of the refactoring work:

- [x] Address missing features on `--no-check` (#7080)
- [x] Refactor diagnostics so they are serialised directly from tsc and the formatting becomes traits of the diagnostics (#7420)
- [x] Move all the configuration analysis to Rust (#5606)
- [x] Better abstract out the ast parsing using `ParsedModule` (#7456)
- [x] Introduce new module graph and move `--no-emit` to it (#7621)
- [x] Instead of hydrating all the source modules in the JS side of tsc, have the compiler ops directly query the module graph
- [x] Simplify the JS side of the compiler logic dramatically.  The reference in ~430 locs including all the JSDoc types.  The current CLI one is ~2400 locs.
- [x] Simplify the creation of tsc snapshots.
- [x] Simplify the Rust interface to the compiler/tsc.
- [x] Implement better abstractions around ast then used in swc_utils.
- [x] Rationalise all the type logic into the module graph (`@deno-types`, `X-TypeScript-Types`, `/// <reference />`) and have the module graph have the concept of runtime code and type code for each module.
- [x] Improve bundling to be able to support `--no-check`. (#7669)
- [x] Fully abstract the compiler, module graph and cache.

*Note* all of these are being worked on, and they tend to be at the complex end of the spectrum.  If there are small bits where contribution can be easy, we will try to break it out as a seperate issue.

Here is the reference implementation: https://github.com/kitsonk/deno/tree/deno_compiler/compiler

kitsonk:
There might still be some cleanup, but most of it is done.

PR #7743, opened 2020-09-28
[WIP] refactor: Flags passing

bartlomieju:


kitsonk:
I know you are keen on trying to address some of this stuff, but it is going to be hard to work on #7225 if we have other refactors going on at the same time.  This gets into exactly the worry I had about re-base hell.

bartlomieju:
> I know you are keen on trying to address some of this stuff, but it is going to be hard to work on #7225 if we have other refactors going on at the same time. This gets into exactly the worry I had about re-base hell.

@kitsonk sure, I just close this one. Those issues can be addressed later

PR #7745, opened 2020-09-28
refactor: update deno_fetch interface

bartlomieju:
Change deno_fetch::init to take parameter indicating
buildtime.

Add default implementation for FetchPermissions

### DIFF
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -34,10 +34,6 @@ use std::rc::Rc;
 pub use reqwest; // Re-export reqwest
 
 /// Execute this crates' JS source files.
-///
-/// Second parameter indicates if it's running during
-/// build time. In that case relevant "cargo" directives
-/// will be printed.
 pub fn init(isolate: &mut JsRuntime) {
   let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
   let files = vec![
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -41,28 +41,39 @@ pub use reqwest; // Re-export reqwest
 pub fn init(isolate: &mut JsRuntime, build_time: bool) {
   let manifest_dir = Path::new(env!("CARGO_MANIFEST_DIR"));
   let files = vec![
-    manifest_dir.join("01_fetch_util.js"),
-    manifest_dir.join("03_dom_iterable.js"),
-    manifest_dir.join("11_streams.js"),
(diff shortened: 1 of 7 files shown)


op_crates/fetch/lib.rs
```diff
@@ -58,9 +61,22 @@ pub fn init(isolate: &mut JsRuntime) {
   }
 }
 
+pub fn init_buildtime(isolate: &mut JsRuntime) {
+  init(isolate, true)
+}
+
+pub fn init_runtime(isolate: &mut JsRuntime) {
+  init(isolate, false)
+}
```
ry, line 70: I think these methods are superfluous. Maybe just add a docstring comment to init explaining what the boolean flag does.

ry (approved):
Thanks!

ry (approved):
LGTM

op_crates/fetch/lib.rs
```diff
@@ -46,7 +51,10 @@ pub fn init(isolate: &mut JsRuntime) {
   // workspace root.
   let display_root = manifest_dir.parent().unwrap().parent().unwrap();
   for file in files {
-    println!("cargo:rerun-if-changed={}", file.display());
+    if build_time {
+      println!("cargo:rerun-if-changed={}", file.display());
+    }
```
ry, line 56: If we're initializing at runtime, ideally we'd use include_str for the source code...

op_crates/fetch/lib.rs
```diff
@@ -33,34 +33,58 @@ use std::rc::Rc;
...
     let display_path = file.strip_prefix(display_root).unwrap();
     let display_path_str = display_path.display().to_string();
-    isolate
-      .execute(
-        &("deno:".to_string() + &display_path_str.replace('\\', "/")),
-        &std::fs::read_to_string(&file).unwrap(),
-      )
-      .unwrap();
+    let url = "deno:".to_string() + &display_path_str.replace('\\', "/");
+    isolate.execute(&url, source_code).unwrap();
```
douglascaetano, line 68: Do you need to generate a string at runtime to get the full path of the file? Would not it be enough to have a static &str with crate name plus file name, like `deno_fetch/01_fetch_util.js`?

If so, file name at lines 40-59 could be generated at compile time.

ry: Yes - it's a little annoying to have the paths hardcoded, but I agree that would be simpler. Updated.

bartlomieju (commented):
LGTM

PR #7747, opened 2020-09-29
refactor: improve graph and tsc_config

kitsonk:
This does three things:

- Introduces `TsConfig` to `tsc_config` to manage configurations and provide an abstraction for the configuration which cleans up the API in the `--no-check`.  It is now super straight-forward:

  ```rust
  let handler =
    Rc::new(RefCell::new(FetchHandler::new(&self.flags, &permissions)?));
  let mut builder = GraphBuilder::new(handler, maybe_import_map);
  builder.insert(&module_specifier).await?;
  let mut graph = builder.get_graph(&self.lockfile)?;

  let (stats, maybe_ignored_options) =
    graph.transpile(TranspileOptions {
      debug: self.flags.log_level == Some(log::Level::Debug),
      maybe_config_path: self.flags.config_path.clone(),
    })?;

  if let Some(ignored_options) = maybe_ignored_options {
    eprintln!("{}", ignored_options);
   }

  debug!("{}", stats);
  ```

- Improves `IgnoredOptions` to "own" the path where the ignored options came from, meaning that it is self contained and can easily be logged out at a higher level without the higher level having to explicitly know about where it came from.
- Reverts some of the `Result<T>` to `Result<T, AnyError>`.

I tried to move `tsc.rs` and the rest of `global_state` over to `TsConfig` but the whole thing is a bit too tangled to easily pull that out, but as we move more over to the new module graph, we should be able to clean all that up.

### DIFF
--- a/cli/global_state.rs
+++ b/cli/global_state.rs
@@ -21,8 +21,6 @@ use deno_core::error::AnyError;
 use deno_core::ModuleSpecifier;
 use std::cell::RefCell;
 use std::env;
-use std::fs;
-use std::io;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::sync::Mutex;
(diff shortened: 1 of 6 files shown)


cli/global_state.rs
```diff
@@ -134,37 +132,17 @@ impl GlobalState {
...
-        Some(config_str)
-      } else {
-        None
-      };
-
       let (stats, maybe_ignored_options) =
         graph.transpile(TranspileOptions {
           debug: self.flags.log_level == Some(log::Level::Debug),
-          maybe_config,
+          maybe_config_path: self.flags.config_path.clone(),
```
bartlomieju, line 138: The only problem is see with this change is that `runtime_transpile` will require specialized version (because config is passed as JSON object). I guess it could be addressed later.

bartlomieju (approved):
LGTM 👍  nice cleanup

cli/global_state.rs
kitsonk: Yeah, that is a valid point, but yes, let's address later, as I think it will be easier to figure out how best to abstract that out.

Issue #7709, opened 2020-09-26
deno test doesnt work with --no-check


lucacasonato:
Since v1.4.2 `deno test` does not work with `--no-check` anymore.

```
/m/f/P/g/l/deno_s3 ❯❯❯ deno test --allow-net --allow-env --no-check
error: Cannot resolve module "file:///mnt/f9/Projects/github.com/lucacasonato/deno_s3/.deno.test.ts"
```

Example test file:

```ts
// foo_test.ts
import { assertEquals } from "https://deno.land/std@0.71.0/testing/asserts.ts";

Deno.test("test", () => {
  assertEquals(1 + 1, 2);
});
```

I have not bisected, but possibly related: #7621

cc @kitsonk @bartlomieju

kitsonk:
Yeah, it is related to that. We don't have a `test --no-check` integration test, so we didn't catch it. I will fix it.

kitsonk:
Gargh... This isn't super straight forward...

When we test, we create a "fake" module and insert it in the `global_state.file_fetcher` but the new `Graph` doesn't use the `global_state.file_fetcher` but instead creates its own, in an effort to start pulling out the file fetcher and eventually refactoring it out totally.  I've never been happy with the fake module, that should ultimately be a truly abstract module, but that will need to happen after other refactors.

So I think the most straight forward mechanism at the moment is to pass a reference to the `global_state` file fetcher, but that won't be the easiest I think as well, as it will require lifetime parameters.  Hrmph...

caspervonb:
>  I've never been happy with the fake module, that should ultimately be a truly abstract module, but that will need to happen after other refactors.

I have a new parallel capable test runner that doesn't use the intermediate module that I'll raise for discussion soon'ish.

It's basically https://github.com/denoland/deno/pull/7165 with the test runner and reporter(s) moved to Rust.

bartlomieju:
> When we test, we create a "fake" module and insert it in the global_state.file_fetcher but the new Graph doesn't use the global_state.file_fetcher but instead creates its own, in an effort to start pulling out the file fetcher and eventually refactoring it out totally.

@kitsonk there's a lot of interactions to consider if we were to stop using "global" file fetcher - especially how multiple file fetchers in single process would access/save to cache and how `--reload` flag would work in that situation. 

I believe we should go back to using the "global" file fetcher when wrapping up compiler refactor and address the problems with file fetcher afterwards (it's still on my radar).

@caspervonb I'm looking forward to seeing your idea.

PR #7748, opened 2020-09-29
fix(cli): use global_state file_fetcher when using SpecifierHandler

kitsonk:
Fixes: #7709

### DIFF
--- a/cli/specifier_handler.rs
+++ b/cli/specifier_handler.rs
@@ -4,8 +4,7 @@ use crate::deno_dir::DenoDir;
 use crate::disk_cache::DiskCache;
 use crate::file_fetcher::SourceFileFetcher;
 use crate::file_fetcher::TextDocument;
-use crate::flags::Flags;
-use crate::http_cache::HttpCache;
+use crate::global_state::GlobalState;
 use crate::media_type::MediaType;
 use crate::permissions::Permissions;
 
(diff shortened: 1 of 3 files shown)


cli/specifier_handler.rs
```diff
@@ -204,27 +204,16 @@ pub struct FetchHandler {
...
-      ca_file.as_deref(),
-    )?;
     let disk_cache = deno_dir.gen_cache;
+    let file_fetcher = global_state.file_fetcher.clone();
 
     Ok(FetchHandler {
       disk_cache,
       file_fetcher,
-      permissions: permissions.clone(),
+      permissions: Permissions::allow_all(),
```
bartlomieju, line 216: Careful - this might be permission escape - `permissions` should still be passed down from `GlobalState`

bartlomieju (commented):
Just need a small change, otherwise LGTM

cli/specifier_handler.rs
kitsonk: I thought I tried that and it broke things, like required you to use `--allow-read` to even read a local module. I will try again though.

bartlomieju: This constructor is called during `prepare_module_load` - so the permissions passed from `prepare_module_load` will definitely be the proper ones.

kitsonk: Ah, I think that was the problem, I was cloning them off of the `global_state`... 🤦

bartlomieju (approved):
Lgtm

PR #7749, opened 2020-09-29
chore(cli/coverage): remove unused runtime domain

caspervonb:
Previously when we used the websocket to talk to the inspector we used the runtime domain to send a "runIfWaitingForDebugger" message.

However this is not required since we now talk to the inspector directly and no longer send said message so this removes the enabling of the domain entirely.

### DIFF
--- a/cli/coverage.rs
+++ b/cli/coverage.rs
@@ -129,10 +129,6 @@ impl CoverageCollector {
       .await?;
 
     self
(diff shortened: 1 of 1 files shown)


ry (changes requested):
LGTM

ry (approved):
LGTM

Issue #7739, opened 2020-09-28
Lint errors for rule "camelcase" in std/hash


bartlomieju:
[`camelcase` lint rule is ready to be merged](https://github.com/denoland/deno_lint/pull/302), however there are a few lint problems it causes in `std/`. Most of them have already been addressed in #7360, however `std/hash` still causes some errors and I don't see an easy way out of it:
```
(camelcase) Identifier 'Sha3_224' is not in camel case.
export class Sha3_224 extends Sponge {
             ^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/hash/_sha3/sha3.ts:7:13

(camelcase) Identifier 'Sha3_256' is not in camel case.
export class Sha3_256 extends Sponge {
             ^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/hash/_sha3/sha3.ts:19:13

(camelcase) Identifier 'Sha3_384' is not in camel case.
export class Sha3_384 extends Sponge {
             ^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/hash/_sha3/sha3.ts:31:13

(camelcase) Identifier 'Sha3_512' is not in camel case.
export class Sha3_512 extends Sponge {
             ^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/hash/_sha3/sha3.ts:43:13

(camelcase) Identifier 'Sha3_224' is not in camel case.
export { Sha3_224, Sha3_256, Sha3_384, Sha3_512 } from "./_sha3/sha3.ts";
         ^^^^^^^^
    at /Users/biwanczuk/dev/deno/std/hash/sha3.ts:3:9
```
We could of course ignore those diagnostics in the relevant files, but that's not the proper solution as users importing this file will get the same errors in their files. 

CC @caspervonb @magurotuna

magurotuna:
We could rename `Sha3_224` to `SHA3_224` for example, so that the diagnostics won't be generated.
Referring to [sha3 - npm](https://www.npmjs.com/package/sha3), a class named `SHA3` is exported. Therefore in our case it'd be good to name `SHA3_224` or something like that.

bartlomieju:
@magurotuna thanks for pointing that out. There's already a PR (#7258) that would remove `Sha3_*` completely in favor of WASM library, but your solution might be a good middle-ground before it happens.

magurotuna:
[...]

PR #7752, opened 2020-09-29
refactor(std/hash): make SHA uppercase

bartlomieju:
Rename `Sha_*` algorithms to use uppercase

This change is dictated by `deno lint` failing on non-camelcase identifiers.

Closes https://github.com/denoland/deno/issues/7739

### DIFF
--- a/cli/tests/hash.ts
+++ b/cli/tests/hash.ts
@@ -3,10 +3,10 @@
 const { args } = Deno;
 import { createHash, SupportedAlgorithm } from "../../std/hash/mod.ts";
 import { Md5 } from "../../std/hash/md5.ts";
-import { Sha1 } from "../../std/hash/sha1.ts";
-import { Sha256 } from "../../std/hash/sha256.ts";
-import { Sha512 } from "../../std/hash/sha512.ts";
-import { Sha3_224, Sha3_256, Sha3_384, Sha3_512 } from "../../std/hash/sha3.ts";
+import { SHA1 } from "../../std/hash/sha1.ts";
+import { SHA256 } from "../../std/hash/sha256.ts";
+import { SHA512 } from "../../std/hash/sha512.ts";
+import { SHA3_225, SHA3_256, SHA3_384, SHA3_512 } from "../../std/hash/sha3.ts";
 
 if (args.length < 3) Deno.exit(0);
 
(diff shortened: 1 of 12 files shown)


kitsonk:
This is a breaking change (for example it will require updates to [oak](https://github.com/oakserver/oak/blob/main/deps.ts#L8)).  Is disabling the rule not an option, since we want to remove these APIs anyways?

bartlomieju:
> This is a breaking change (for example it will require updates to [oak](https://github.com/oakserver/oak/blob/main/deps.ts#L8)). Is disabling the rule not an option, since we want to remove these APIs anyways?

@kitsonk this rule is not yet enabled - it's waiting till we fix the code in `std/`. I've talked with @caspervonb about it and it seems WASM rewrite is still a few weeks off. So I decided to rename them in the mean time. Unfortunately there's no middle ground on this one, even if I disable this rule in `std/hash` files the errors would still show up in user code.

nayeemrmn:
Why are we renaming them against our casing conventions? The problem is the underscore. If there's nothing we can replace it with then `// deno-lint-ignore`.

bartlomieju:
> Why are we renaming them against our casing conventions? 

To be honest I don't understand why it was `Sha_*` in the first place, it seems `SHA` should be all caps.

> The problem is the underscore. If there's nothing we can replace it with then `// deno-lint-ignore`. Would it really cause errors for importers? That seems wrong as well.

If you import `Sha_256` in user code it still violates `camelcase` rule - you can't disable linting in all files for certain symbols. How is that wrong?

nayeemrmn:
> If you import `Sha_256` in user code it still violates `camelcase` rule - you can't disable linting in all files for certain symbols. How is that wrong?

As in, the rule shouldn't apply to non-aliased named imports or the linter is asserting itself upon external code.

> it seems `SHA` should be all caps

We don't normally put acronyms in all caps (like web APIs unfortunately do) because it fails to achieve the purpose of casing conventions: delimiting words.

kitsonk:
> We don't normally put acronyms in all caps (like web APIs unfortunately do) because it fails to achieve the purpose of casing conventions: delimiting words.

The web APIs are known to do it inconsistently, but yes, it is a class, which means the first letter of each word is uppercased.  Also the original library they were based on followed that convention.  Using `SHA` is actually inconsistent with our conventions.

bartlomieju:
> As in, the rule shouldn't apply to non-aliased named imports or the linter is asserting itself upon external code.

So you're suggesting that import bindings shouldn't be linted unless they have `as <something>` alias? I think that's actually very sensible and we could implement it that way.

kitsonk:
(e.g. `Http` instead of `HTTP`, HTTP is as much of an acronym as SHA is.)

bartlomieju:
Fair enough, let's change how `camelcase` works to ignore import bindings. Closing without merge.

PR #7756, opened 2020-09-29
refactor(cli): extract inspector session

caspervonb:
This extracts the inspector session specific bits from CoverageCollector into a standalone struct dubbed `InspectorSesssion` which can be used as a general purpose session to communicate with the inspector on the same
thread as an isolate/inspector pair lives on.


bartlomieju (approved):
LGTM, nice refactor

PR #7762, opened 2020-09-30
refactor: add concept of 'legacy' compiler to enable non-breaking refactoring

kitsonk:
This is a transitional refactor of the `compiler.js` that unlocks other refactors with the compiler without having to replace the whole compiler in one go.

Because the new compiler uses the `host.createHash()` method and the _legacy_ one does not, TypeScript still detects it on the host and tries to call it, so I had to add the `op` to `tsc.rs`.

I also changed `log()` to `debug()` to more align to our conventions in Rust.  It is one less mental model change to make when moving between JS and Rust.


bartlomieju (approved):
LGTM 👍

PR #7591, opened 2020-09-20
fix(cli/repl): interpret object literals as expressions

caspervonb:
This wraps object literals in parenthesis so that they're interpreted as expressions rather than block statements which matches the behaviour of Node and web-browsers.

Fixes #1421

PR #7763, opened 2020-09-30
refactor(repl): use an inspector session

caspervonb:
This ports the REPL over to Rust and makes use of an inspector session to run a REPL on top of any isolate which lets make full use of rustylines various things like validators and completors without having to introduce a bunch of hard to test internal ops and glue code.

An accidental but good side effect of this is that the multiple line input we previously had is now an editable multi-line input prompt that is correctly stored in the history as a single entry.

We'll also be able to enable top level await via `replMode` in Runtime.evaluate (inspector protocol).


cli/repl.rs
```diff
@@ -1,73 +1,78 @@
...
+      }
+    }
   }
-}
 
-pub fn history_path(dir: &DenoDir, history_file: &str) -> PathBuf {
-  let mut p: PathBuf = dir.root.clone();
-  p.push(history_file);
-  p
+  // TODO(caspervonb) save history file
```
caspervonb, line 75: To self; really remember this in this PR.

cli/repl.rs
```diff
@@ -1,73 +1,78 @@
...
-      .editor
-      .save_history(&self.history_file.to_str().unwrap())
-      .map(|_| debug!("Saved REPL history to: {:?}", self.history_file))
-      .map_err(|e| {
-        eprintln!("Unable to save REPL history: {:?} {}", self.history_file, e);
-        e.into()
-      })
-  }
+  loop {
+    let line = editor.readline("> ");
```
caspervonb, line 27: Spawn blocking, most likely.

cli/repl.rs
```diff
@@ -1,73 +1,78 @@
...
+    match line {
+      Ok(line) => {
+        let evaluate_response = session
+          .post_message(
+            "Runtime.evaluate".to_string(),
+            Some(json!({
+                "expression": line,
+                "contextId": context_id,
+                // Set repl mode to true to enable const redeclarations and top level await
+                "replMode": false,
```
caspervonb, line 37: Enable in a follow-up, parity with the current implementation is the goal here.

cli/repl.rs
bartlomieju: Yes, otherwise event loop won't run when waiting on the input

cli/tests/integration_tests.rs
```diff
@@ -1322,7 +1322,7 @@ fn repl_test_assign_underscore_error() {
   assert!(
     out.ends_with("Last thrown error is no longer saved to _error.\n1\n1\n")
   );
-  assert_eq!(err, "Thrown: 2\n");
+  assert_eq!(err, "Uncaught 2\n");
```
caspervonb, line 1325: Minor deviation from the current implementation.

cli/tests/integration_tests.rs
```diff
@@ -1157,6 +1157,7 @@ fn repl_test_function() {
 }
 
 #[test]
+#[ignore]
```
caspervonb, line 1160: Temporarily disabled, editing (e.g multiline) only works with a tty.

cli/tests/integration_tests.rs
bartlomieju: IIRC "Uncaught" I added to error message by V8 itself. I think this change is fine.

cli/repl.rs
```diff
@@ -1,73 +1,240 @@
...
-use std::fs;
-use std::path::PathBuf;
+use rustyline_derive::{Completer, Helper, Highlighter, Hinter};
+use std::sync::Arc;
+use std::sync::Mutex;
 
-pub struct Repl {
-  editor: Editor<()>,
-  history_file: PathBuf,
+#[derive(Completer, Helper, Highlighter, Hinter)]
```
bartlomieju, line 17: Does using `derive` here actually provide completions/hints/highlights? Or is it just to have default impls?

cli/repl.rs
```diff
@@ -1,73 +1,240 @@
...
+struct Helper {
```
bartlomieju, line 18: A short comment would be helpful on the purpose of this struct.

cli/repl.rs
```diff
@@ -1,73 +1,240 @@
...
+    })
+    .await?;
+
+    match line {
+      Ok(line) => {
+        let evaluate_response = session
+          .post_message(
+            "Runtime.evaluate".to_string(),
+            Some(json!({
+              "expression": format!("'use strict'; void 0;\n({})", line),
```
bartlomieju, line 115: Why wrapping in parentheses? Could you add a comment here?

cli/repl.rs
```diff
@@ -1,73 +1,240 @@
...
+            "Runtime.evaluate".to_string(),
+            Some(json!({
+              "expression": "(globalThis.closed)",
+              "contextId": context_id,
+            })),
+          )
+          .await?
+          .get("result")
+          .unwrap()
+          .get("value")
```
bartlomieju, line 154: 👍  nice

cli/rt/99_main.js
```diff
@@ -329,10 +327,6 @@ delete Object.prototype.__proto__;
 
     util.log("cwd", cwd);
     util.log("args", args);
-
-    if (repl) {
-      replLoop();
-    }
```
bartlomieju, line 335: Nice to see this go

bartlomieju (commented):
Awesome work @caspervonb just a few minor nitpicks, but other than that I think it's landable.

cli/repl.rs
```diff
@@ -1,73 +1,240 @@
...
+      Ok(line) => {
+        let evaluate_response = session
+          .post_message(
+            "Runtime.evaluate".to_string(),
+            Some(json!({
+              "expression": format!("'use strict'; void 0;\n({})", line),
+              "contextId": context_id,
+              // TODO(caspervonb) set repl mode to true to enable const redeclarations and top
+              // level await
+              "replMode": false,
```
bartlomieju, line 119: It'd be great if we could switch this on before Friday's release

cli/repl.rs
caspervonb: Basically to allow a best guess of what `{}` means.
It's a bit magical but matches Node behaviour, implemented in https://github.com/denoland/deno/pull/7591

ry:
This used to work, but appears to be broken now:
```
~/src/deno> cargo run
   Compiling deno v1.4.2 (/Users/rld/src/deno/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 11s
     Running `target/debug/deno`
Deno 1.4.2
exit using ctrl+d or close()
> function add(a, b) {
return a + b;
}
[Function: add]
> add(1, 2)
Uncaught ReferenceError: add is not defined
    at <anonymous>:2:1
>
```

caspervonb:
> This used to work, but appears to be broken in this branch:

Oops, yeah bit too eager on the wrapping, need to wind it back to just `starts_with("{"` like #7591 did.

caspervonb:
Toned it down a bit to match #7591 exactly.

cli/repl.rs
caspervonb: I'll flip it and add tests as soon as this lands.

cli/repl.rs
caspervonb: Just to have default no-op implementations because the `Helper` trait requires them all to be implemented.

bartlomieju (approved):
LGTM, massive work @caspervonb

Issue #7153, opened 2020-08-22
Deno calls object getters when logging, resulting in code that performs differently than expected.


00ff0000red:
In all browser vendors and Node.js, when using `console.log` on an object containing getters, the getters are shown as getters, they aren't actually called.

Example:
```ts
let i = 0n;

const objectContainingGetter = {
	get x() {
		return ++i;
	}
};

console.log(objectContainingGetter);
// expected to be something along the lines of
// { [x: getter] }

console.log(objectContainingGetter.x);
// expected to be "1n"
```

00ff0000red:
Has this been fixed in the pull request connected to getters and setters in general allowing them to be logged?

solson:
For reference, node prints it like this:

```console
> console.log(objectContainingGetter)
{ x: [Getter] }
```

PR #7766, opened 2020-09-30
fix(cli/console): Catch and format getter errors

nayeemrmn:
```js
console.log({
  get location() {
    throw new ReferenceError(
      `Access to "location", run again with "--location <href>".`,
    );
  },
});
```

Before:
<kbd>image</kbd>

After:
<kbd>image</kbd>

~~Note that Node doesn't evaluate getters by default. Due to heavy use of getters in our runtime structures, this would lead to really annoying formatting for us, so let's continue the way we are.~~

### DIFF
--- a/cli/rt/02_console.js
+++ b/cli/rt/02_console.js
@@ -762,29 +762,33 @@
       );
     }
 
+    const red = maybeColor(colors.red, inspectOptions);
+
     for (const key of stringKeys) {
-      entries.push(
-        `${maybeQuoteString(key)}: ${
-          inspectValueWithQuotes(
-            value[key],
-            ctx,
-            level + 1,
-            inspectOptions,
-          )
-        }`,
-      );
+      let propertyValue;
+      let error = null;
+      try {
+        propertyValue = value[key];
+      } catch (error_) {
+        error = error_;
+      }
+      const inspectedValue = error == null
+        ? inspectValueWithQuotes(propertyValue, ctx, level + 1, inspectOptions)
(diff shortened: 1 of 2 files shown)


bartlomieju:
@nayeemrmn so this PR addresses https://github.com/denoland/deno/issues/7153?

nayeemrmn:
> so this PR addresses #7153?

No (in fact, I'm saying in the last paragraph why such a change would sometimes be annoying 😅).

But maybe it's not good to evaluate getters. I can address that in a future PR.

caspervonb:
Please no, let's not evaluate any executable code in the console
For better or worse users do weird things in getter so we can't guarantee we aren't causing side effects.

nayeemrmn:
@caspervonb That's not introduced by this patch... look at the before image. I said I can correct that behaviour in a later PR by putting the evaluation (along with the error handling) behind `getters: true` like Node.

caspervonb:
Try catch around a getter evaluation looks like evaluation to me?

nayeemrmn:
Again, evaluation that was already there, just _not_ around a try/catch.

caspervonb:
> Again, evaluation that was already there, just _not_ around a try/catch.

Definitively a mistake in the implementation IMO, getters should just be presented `[Function: get]`.

bartlomieju (approved):
I agree with @caspervonb that we shouldn't be evaluating getter for formatting, but I presume this is only an intermediate change. @nayeemrmn next minor release will be on October 27th, I'd be great if we could apply your suggested change before then.

PR #7768, opened 2020-09-30
updating readme

Vikas7000:


kitsonk (changes requested):
Your suggest change is incorrect.

caspervonb:
For reference: http://github.com/dprint

PR #7770, opened 2020-10-01
chore: ensure cache is 'valid' during benchmarks

kitsonk:
This fixes the [strange results observed in the benchmarks we are seeing](https://github.com/denoland/deno/pull/7760#issuecomment-701724435).


lucacasonato (approved):
LGTM

Issue #3802, opened 2020-01-26
Support reporting of CPU information


cknight:
I am looking into the Node polyfill for 'os' ( #3403 ).  Node offers the `cpus()` function to get information about the logical CPU cores.  See [cpus() API](https://nodejs.org/api/os.html#os_os_cpus).  Deno does not currently seem to offer any way to get at this information and new functionality will need to be added to Deno to complete this polyfill.

ecyrbe:
i'm currently working on this

ecyrbe:
CC @ry can i use the crate [systemstat](https://docs.rs/systemstat/0.1.5/systemstat/index.html) to implement :
- freemem()
- totalmem()
- cpus()
- uptime()

and more...

it is under more activity than `sys_info` and made entirely in **rust**. unlike what the documentation suggests, the author added windows support for all of these.

I really don't wan't to make my own `/proc/stat` `/proc/cpuinfo` parser and windows compatibility module if someone already made one.

your call.

ecyrbe:
Or if we are not confident enought, use a libuv binding : https://docs.rs/libuv-sys2/1.34.4/libuv_sys2/
but this would be a huge depedency that i'm not sure we want to pay such a price.

ry:
We certainly are not adding libuv (despite it being our product) it overlaps too much with Tokio. 

I will look into sys_info vs systemstat.

ecyrbe:
@ry did you have time to investigate ?

ry:
@ecyrbe I haven't really, but systemstat seems quite reasonable. It has CI for windows unlike sys_info ([which has been broken on windows for the last two releases](https://github.com/FillZpp/sys-info-rs/issues/50)). If anyone felt inclined to switch us from sys-info to systemstat, that would be welcome.

I haven't checked if systemstat actually does all the things that sys-info does tho.

vladimyr:
> CC @ry can i use the crate [systemstat](https://docs.rs/systemstat/0.1.5/systemstat/index.html) to implement :
> 
> * freemem()
> * totalmem()

Unfortunately, you won't get really far with it. It doesn't support memory info on macOS: https://github.com/myfreeweb/systemstat/blob/a7c54aa/src/platform/macos.rs#L58-L60

juliangruber:
`os.hostname()` is missing from that list

obedm503:
not quite

obedm503:
not stale

PR #7774, opened 2020-10-01
feat(unstable): add Deno.systemCpuInfo()

eliassjogreen:
Related to #3802, I decided to not implement a node compatible `cpus` function due to the `sys_info` crate not fully supporting all the functionality needed (cpu model and times in different modes).


cli/tests/unit/os_test.ts
```diff
@@ -188,3 +188,9 @@ unitTest({ perms: { env: true } }, function systemMemoryInfo(): void {
   assert(info.swapTotal >= 0);
   assert(info.swapFree >= 0);
 });
+
+unitTest({ perms: { env: true } }, function systemCpuInfo(): void {
+  const { cores, speed } = Deno.systemCpuInfo();
+  assert(cores >= 0);
+  assert(speed >= 0);
```
bartlomieju, line 195: Are there situation where this API might return `0` for any of the fields? If so it should be mentioned in jsdoc for the API

eliassjogreen: I don't think so but I guess technically the speed could be < 1MHz. Cores should always be more than 0 so I will change that. (looked at the rust tests for `sys_info` and they seem test if its more than 0 for both)

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -167,6 +167,27 @@ declare namespace Deno {
...
+   * Requires `allow-env` permission.
+   *
+   */
+  export function systemCpuInfo(): SystemCpuInfo;
+
+  export interface SystemCpuInfo {
+    /** Total number of logical cpus in the system */
+    cores: number;
+    /** The speed of the cpu measured in MHz */
+    speed: number;
```
bartlomieju, line 188: Seems like both of those fields are omitted when an error occurs - it should be reflected in the types and doc comment

bartlomieju (changes requested):
@eliassjogreen sorry for slow review, only one small thing to fix and we can land it for 1.5.0

bartlomieju:
@eliassjogreen I'd be interested in landing this PR for tomorrow's release, but the CI is still red.

ry (approved):
LGTM - thank you @eliassjogreen

Issue #2893, opened 2019-09-09
Support python3


ry:
We are heavily dependent on V8, which in turn is heavily dependent on Chromium tooling, which does not support Python3 and has only recently begun conversion -
https://bugs.chromium.org/p/chromium/issues/detail?id=942720
https://bugs.chromium.org/p/v8/issues/detail?id=8594
https://github.com/nodejs/node/issues/24512

hayd:
I'm happy to do this, I think it'll be straightforward, but it's probably worth waiting for v8 support to be completed.

cclauss:
> it's probably worth waiting

Not!  Python 3 was released 11 years ago and there are 113 days until Python 2 end of life.  Please do work in _parallel_, not serial to be ready before yearend.  It would be great if someone could get v8/v8#33 and v8/v8#34 pushed and landed upstream.

hayd:
@cclauss I bet I can do the deno part of the port in under an hour, certainly less than a day. I've ported large modules to python 3... deno is very small. I suppose it would help to enumerate precisely which python v8 scripts need to be ported in order for deno to work, I can do that.

cclauss:
Not a question of your skills.  I have been impressed.  It is just a question of putting lots of concentrated pressure on a few bottlenecks that are not ready so we create a _compelling reason to act_.  We want to create the impression that the train is leaving the station without the laggards onboard.

kt3k:
https://travis-ci.com/bmsdave/v8/builds/132539826
Does this mean v8 can be built with python 3?

cclauss:
@bmsdave Can V8 be built with Python 3?

hayd:
No: https://github.com/bmsdave/v8/commit/845d137e411fb86a06b7019f590dfd806b36dc44

cclauss:
https://travis-ci.com/bmsdave/v8/builds Seems to be passing the lint steps.  Where does the V8 build on Py3 break down?

hayd:
The script it's running is `true`, the Travis build is not running any build/tests other than flake8...

cclauss:
Exactly.  So where can we see the __build__ failing on Python 3?

hayd:
Sorry I was answering @kt3k :

> https://travis-ci.com/bmsdave/v8/builds/132539826
> Does this mean v8 can be built with python 3?

It could be that the build works with just the syntax changes (but I would be surprised).

[...]

PR #7779, opened 2020-10-01
Py3 compat

j4qfrost:
Makes build scripts compatible with python 3.


nayeemrmn:
Huh, is this all we need for #2893 now?

j4qfrost:
> Huh, is this all we need for #2893 now?

That's as far as I can tell. This works on OSX and Linux.

kitsonk (approved):
LGTM

hayd:
Should [this](https://github.com/denoland/deno/blob/4c779b5e8ca427faf24c26443a8054004827d450/.github/workflows/ci.yml#L79) be python 3 for at least some of the build matrix?

j4qfrost:
> Should [this](https://github.com/denoland/deno/blob/4c779b5e8ca427faf24c26443a8054004827d450/.github/workflows/ci.yml#L79) be python 3 for at least some of the build matrix?

That's a point I forgot to consider. Maybe this issue should be addressed in the context of sunsetting python 2. The intention with this PR was just to make the tests play nice with python 3. I have a weird feeling about changing up the python versions in just some places.

hayd:
This may regress without using python3 in at least one test in the matrix (e.g. linux).

Note: Building with python3 works _provided_ you don't build V8. 😎

Issue #3700, opened 2020-01-17
await should work in REPL


ry:
```
> await Deno.permissions.query({ name: "read" })
error: Uncaught SyntaxError: await is only valid in async function
► <unknown>:1:1
    at evaluate ($deno$/repl.ts:84:34)
    at replLoop ($deno$/repl.ts:175:13)
```

kevinkassimo:
Node seems to be using a `(async () => { ${src} })()` wrapper for this. But they also explicitly used a JS parser to ensure that there is no top-level return along with this `await` (which breaks the wrapper)

lucacasonato:
[TLA only works in V8 module contexts](https://v8.dev/features/top-level-await) - because the REPL is a traditional script context it can't do TLA.

Same root cause as #1285. (https://github.com/denoland/deno/issues/1285#issuecomment-454900436) - if the REPL will move to a module context, which would be nice, this would work out of the box. For now we could do the same as chrome dev tools and node which both just wrap the call in an async function.

kitsonk:
> if the REPL will move to a module context, which would be nice, this would work out of the box

I think that would be hard, because a module needs to be "closed" so its imports and exports and be analysed.  It would effectively need to rerun the whole "module" every time a line is added.  Only script contexts can really be open-ended like that, which means a few things have to be frigged.

kevinkassimo:
> > if the REPL will move to a module context, which would be nice, this would work out of the box
> 
> I think that would be hard, because a module needs to be "closed" so its imports and exports and be analysed. It would effectively need to rerun the whole "module" every time a line is added. Only script contexts can really be open-ended like that, which means a few things have to be frigged.

Yeah. And running a module again and again might yield unexpected repeated side effects (e.g. console.log multiple times)

Issue #3029, opened 2019-09-26
REPL issue with Top Level Await


kevinkassimo:
Top level await has unexpected results in the repl:

```
> console.log(await Deno.stat("."));
// Waits forever...
```
The cause of this is that feeding the above line to `evalContext` (plainly a binding to create script and eval under the current context) yields the surprising error `SyntaxError: missing ) after argument list`, thus REPL assumes that there needs more input and waits.

This might be an upstream issue of V8.

bartlomieju:
Fixed in https://github.com/denoland/deno/pull/7784

Issue #1285, opened 2018-12-06
es module imports in repl don't work


xiaodeaux:
When in the Deno REPL and executing:
`import { env } from "deno";`

The output is:
`SyntaxError: Unexpected token {`

Probably related to #1158 

Executing: `deno -v` on the commandline returns:
`deno: 0.2.1`
`v8: 7.1.302.4`
`typescript: 3.2.1`

hayd:
Typescript isn't supported in the repl yet, #1158.

kitsonk:
That isn't a TypeScript feature persay... `import` is valid ES6 syntax.  But imports in general don't work in the REPL, which is a known limitation.  Native ESM support is being worked on (tracked by #975) and it should flow through to the REPL eventually.

The `deno` namespace though is available in the global scope, so `const { env } = deno;` would work for now.

shreeve:
Now that #1460 (ES module support) has landed, can someone provide a quick line that can be entered in the REPL to verify that the import statement is working?

I'm still seeing this, even with the new support from #1460 that just landed:

```
> import { env } from "deno";
SyntaxError: Unexpected token {
```

Executing: `deno -h` on the commandline returns:

```
deno -h
Usage: deno script.ts

Options:
        --allow-write   Allow file system write access.
        --allow-net     Allow network access.
        --allow-env     Allow environment access.
        --allow-run     Allow running subprocesses.
    -A, --allow-all     Allow all permissions
        --recompile     Force recompilation of TypeScript code.
    -h, --help          Print this message.
    -D, --log-debug     Log debug output.
    -v, --version       Print the version.
    -r, --reload        Reload cached remote resources.
        --v8-options    Print V8 command line options.
        --types         Print runtime TypeScript declarations.

Environment variables:
        DENO_DIR        Set deno's base directory.
```

The latest version in git shows the prior version number, but the `--allow-all` option shows that it's using the newest code (which includes support for ES modules).

kyranet:
I believe this kind of imports will never be available in REPL since ES imports/exports are static. However, you can use the `import()` API.

kevinkassimo:
@hayd To do such translation, top level await (or at least some way to pause the REPL during `import()`) would be required. However, it seems we are still having some issue with dynamic `import()` in terminal, as it is printing
```
> Error: Not supported
    at eval (<unknown>:1:14)
    at eval (<unknown>:1:66)
    at evaluate (deno/js/repl.ts:86:25)
[...]

PR #7784, opened 2020-10-02
fix(cli/repl): enable await and let re-declarations

caspervonb:
This enables `replMode` during evaluations which allows for top level  await and let re-declarations.

Fixes https://github.com/denoland/deno/issues/3700 https://github.com/denoland/deno/issues/3029

### DIFF
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -126,9 +126,7 @@ pub async fn run(
             Some(json!({
               "expression": format!("'use strict'; void 0;\n{}", &wrapped_line),
               "contextId": context_id,
-              // TODO(caspervonb) set repl mode to true to enable const redeclarations and top
-              // level await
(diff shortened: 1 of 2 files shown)


caspervonb:
Not sure if we should consider https://github.com/denoland/deno/issues/1285 resolved with this as await import works now.

nayeemrmn:
It seems like relative imports don't work with this.
image
Currently dynamic imports are resolved relative to a dummy file in the cwd.

bartlomieju:
@caspervonb maybe mark as `fix(cli/repl)` 😅 ?

bartlomieju (approved):
LGTM 🚀  awesome update @caspervonb

bartlomieju:
CC @solson I recall you were waiting on TLA in REPL

PR #7788, opened 2020-10-02
 Update github action setting environment variable workflow command

khs1994:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,8 @@ jobs:
         # that are needed during the build process. Additionally, this works
         # around a bug in the 'cache' action that causes directories outside of
         # the workspace dir to be saved/restored incorrectly.
-        run: echo "::set-env name=CARGO_HOME::$(pwd)/.cargo_home"
+        shell: pwsh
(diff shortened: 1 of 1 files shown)


caspervonb:
What is the purpose of this patch?

khs1994:
https://github.com/actions/toolkit/security/advisories/GHSA-mfwh-5m23-j46w

old command will deprecated

.github/workflows/ci.yml
```diff
@@ -100,7 +100,8 @@ jobs:
         # that are needed during the build process. Additionally, this works
         # around a bug in the 'cache' action that causes directories outside of
         # the workspace dir to be saved/restored incorrectly.
-        run: echo "::set-env name=CARGO_HOME::$(pwd)/.cargo_home"
+        shell: pwsh
```
caspervonb, line 103: Why change the shell from default?

khs1994: linux and windows diff

if use bash in windows `CARGO_HOME: /d/a/deno/deno/.cargo_home`

khs1994: `pwsh` exists in all os

khs1994:
ping @ry @fd

khs1994:
#8020

PR #7790, opened 2020-10-02
chore(core): remove experimental wasm bigint flag

caspervonb:
This removes the experimental bigint wasm integration flag as it is enabled by default now and explicitly enabling it is no longer necessary.


caspervonb:
This happened in V8 release v8.5: https://v8.dev/blog/v8-release-85

bartlomieju (approved):
LGTM

PR #7794, opened 2020-10-02
fix(cli/repl): use a default referrer when empty

caspervonb:
This makes use of a default referrer when its empty in repl mode so that dynamic imports work in the global evaluation context.

### DIFF
--- a/cli/state.rs
+++ b/cli/state.rs
@@ -46,10 +46,23 @@ impl CliModuleLoader {
 impl ModuleLoader for CliModuleLoader {
   fn resolve(
     &self,
+    op_state: Rc<RefCell<OpState>>,
     specifier: &str,
     referrer: &str,
     is_main: bool,
   ) -> Result<ModuleSpecifier, AnyError> {
+    let global_state = {
+      let state = op_state.borrow();
+      state.borrow::<Arc<GlobalState>>().clone()
+    };
+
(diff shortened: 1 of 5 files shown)


cli/tests/integration_tests.rs
```diff
@@ -1252,6 +1252,18 @@ fn repl_test_multiline() {
   assert!(err.is_empty());
 }
 
+#[test]
+fn repl_test_await_import() {
```
caspervonb, line 1256:
```suggestion
fn repl_test_import() {
```

cli/state.rs
```diff
@@ -46,10 +46,23 @@ impl CliModuleLoader {
...
+      let state = op_state.borrow();
+      state.borrow::<Arc<GlobalState>>().clone()
+    };
+
+    // FIXME(bartlomieju): hacky way to provide compatibility with repl
+    let referrer = if referrer.is_empty() && global_state.flags.repl {
+      "<unknown>"
+    } else {
+      referrer
+    };
```
bartlomieju, line 64: This is not ideal, but we decided that it's better to work around quirk than change whole logic - just to keep compatibility with previous versions. Module loading is subject to big refactor anyway so this quirk will be removed soon.

bartlomieju (approved):
LGTM

Issue #7797, opened 2020-10-02
REPL doesn't color output


bartlomieju:
With recent changes to REPL it seems we got a minor regression that caused to stop coloring output.

We should fix it by changing:
https://github.com/denoland/deno/blob/master/cli/repl.rs#L210

to include `color: true` option.

CC @caspervonb

PR #7798, opened 2020-10-02
fix(cli/repl): enable colors on inspected values

caspervonb:
This enables colors on the inspected values before writing them to their output streams.

Fixes https://github.com/denoland/deno/issues/7797


bartlomieju (approved):
LGTM, shame that we can't test it properly

PR #7413, opened 2020-09-10
feat(unstable): enable importsNotUsedAsValues

lucacasonato:
It seems that `"isolatedModules": true` does not catch all errors that can occur with a `--nocheck` emit. It looks like `"importsNotUsedAsValues": "error"` is also needed because `isolatedModules` does not catch `import { MyType } from "./file.ts";`, only type re-exports.

cc @kitsonk

PR #7800, opened 2020-10-02
Revert "feat(unstable): enable importsNotUsedAsValues by default (#7413)"

nayeemrmn:
This reverts commit fbb18d40d3cfd0d24262e8e73b97f22770734572.

Ref #7413.

As discussed, the `--no-check` incompatiblity we were seeing with `import { MyType } from "./file.ts";` was a symptom of https://github.com/swc-project/swc/issues/1065 which is now fixed. The rule ~~is no longer~~ was never really needed.

### DIFF
--- a/cli/tsc.rs
+++ b/cli/tsc.rs
@@ -505,7 +505,6 @@ impl TsCompiler {
       "inlineSourceMap": true,
       // TODO(lucacasonato): enable this by default in 1.5.0
       "isolatedModules": unstable,
-      "importsNotUsedAsValues": if unstable { "error" } else { "remove" },
       "jsx": "react",
       "lib": lib,
       "module": "esnext",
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

lucacasonato (approved):
LGTM - before merging I want to confirm that this case gets stripped correctly now:

```ts
import { MyType } from "./foo.ts";
export type { MyType };

nayeemrmn:
> before merging I want to confirm that this case gets stripped correctly now:
> 
> ```ts
> import { MyType } from "./foo.ts";
> export type { MyType };
> ```

Actually, just after opening this I realised the re-export thing was still broken and fixed it in swc: https://github.com/swc-project/swc/pull/1126. Reiterating that I don't think these temporary swc bugs justify imposing rules for type checking so it breaks there, too.

nayeemrmn:
Also I think this change and that swc fix will both release at our next version.

lucacasonato:
Yeah lets merge this then. As it is fixed upstream we will get the fix in Deno before 1.4.5 is released.

PR #7804, opened 2020-10-03
fix(cli/repl): put the thread to sleep when idle

### LINKED ISSUE #7799: Idle REPL uses CPU instead of actually idling
When I open the REPL in Deno 1.4.3, the Deno process stays at 100% CPU usage (and I can hear my fans spin up), while Deno 1.4.2 actually idles at 0%.

Version info: https://github.com/denoland/deno/commit/e5348a6a38b3c06a916223ad0b99e8cfe47725b1 with `cargo build --release`

cc @caspervonb

caspervonb:
This adds a thread sleep when the worker idles to avoid spinning at 100%, brings it down to a nice and cool ~0.5%.

Fixes #7799

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -447,7 +447,9 @@ async fn run_repl(flags: Flags) -> Result<(), AnyError> {
       result = &mut repl => {
           return result;
       }
-      _ = &mut *worker => {}
+      _ = &mut *worker => {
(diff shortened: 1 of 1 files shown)


caspervonb:
Will confer with @bartlomieju about a long term solution later, this is meant as a hotfix in the meantime.

bartlomieju (approved):
LGTM

PR #7806, opened 2020-10-03
docs: add JSdoc for std/node/util.ts (#7487)

starkensin:
I added JSdoc for std/node/util.ts (#7487)

### DIFF
--- a/std/node/util.ts
+++ b/std/node/util.ts
@@ -42,60 +42,77 @@ export function inspect(object: unknown, ...opts: any): string {
   });
 }
 
+/** @deprecated - use `Array.isArray()` instead. */
 export function isArray(value: unknown): boolean {
   return Array.isArray(value);
 }
 
+/** @deprecated - use `typeof value === "boolean" || value instanceof Boolean` instead. */
 export function isBoolean(value: unknown): boolean {
   return typeof value === "boolean" || value instanceof Boolean;
 }
 
+/** @deprecated - use `value === null` instead. */
 export function isNull(value: unknown): boolean {
   return value === null;
 }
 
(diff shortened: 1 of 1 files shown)


std/node/util.ts
```diff
@@ -42,60 +42,74 @@ export function inspect(object: unknown, ...opts: any): string {
   });
 }
 
+/** Checks if a value is an array. */
```
tokiedokie, line 45: Some functions are deprecated in NodeJS, so write them down?

starkensin: Should I rewrite like below?
```
/** 
 * Checks if a value is an array. 
 * Deprecated in node.js. Use Array.isArray() instead.
 */
```

trivikr, line 45:
```suggestion
/** @deprecated since v4.0.0 - use `Array.isArray()` instead. */
```

You can use `@deprecated` tag.
* JSDoc for `@deprecated` https://jsdoc.app/tags-deprecated.html
* from `@types/node` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/88c6ceed5a956c081d8433f47da8713b1316f88e/types/node/util.d.ts#L26

james-and (approved):

james-and (approved):

james-and (approved):

james-and (approved):

std/node/util.ts
starkensin: Thank you! I will improve JSDoc this way 👍

trivikr (approved):

tokiedokie (approved):

kitsonk (approved):
LGTM

Issue #6094, opened 2020-06-04
btoa and base64 module


timonson:
The correct **base64** encoding of `ß` is `w58=`.

```typescript
import { assertNotEquals } from "https://deno.land/std@v0.56.0/testing/asserts.ts";
assertNotEquals(btoa("ß"), "w58=");  // true
```

The German letter `ß` is one of the ASCII characters below 255 and inside the *latin1 range*, but it has two bytes. Therefore `btoa` doesn't throw an error and converts it to incorrect **base64**. I know that browsers do the same, because the btoa specification  is based on **latin1** (all one byte) instead of **ascii** (128-255 have two bytes) encoding. This is impractical in my opinion (I hope somebody can convince me of the opposite) and deno shouldn't do that, because it causes problems like the following one.

If you want to keep `btoa` like it is, then we must fix the `base64` module nevertheless because it uses `btoa` under the hood and converts to incorrect base64:

```typescript
import { assertNotEquals } from "https://deno.land/std@v0.56.0/testing/asserts.ts";
import { encode } from "https://deno.land/std@v0.56.0/encoding/base64.ts";
assertNotEquals(encode("ß"), encode(new TextEncoder().encode("ß").buffer));  // true
```

kitsonk:
UTF16 strings (which is what JavaScript uses internally) need to be escaped properly encoded when they contain multibyte characters like `ß`.  There is `unescape` but that is Annex B and while it exists in Deno, not something we should really depend on (we already dropped some Annex B stuff).  The encode in Base64 should look something like this:

```js
btoa(encodeURIComponent(value)
  .replace(/%([0-9A-F]{2})/g, (_, p) => String.fromCharCode(`0x${p}`)));
```

The above returns the following for ß:

```
w58=
```

timonson:
Exactly @kitsonk. For an alternative (maybe more performant?) look here: https://gist.github.com/enepomnyaschih/72c423f727d395eeaa09697058238727

Issue #4794, opened 2020-04-17
base64 to/from Uint8Array


pgg:
add deno/std/encoding/base64 support to/from Uint8Array

crabmusket:
For reference: https://deno.land/x/base64/

pgg:
Why does the current base64 implementation expose to/from ArrayBuffer and not Uint8Array like all the other encodings?

ry:
@pgg Likely just an oversight...

timonson:
I would like to mention the following issue here: https://github.com/denoland/deno/issues/6094

I wrote the following in the discord channel but didn't get a reply, so I try it again:

> Hi @kitsonk and the other devs, I would like to fix the base64 module based on this issue: https://github.com/denoland/deno/issues/6094. I would also improve the tests and other things.
> I have two questions:
> 1. Do you prefer Kitson's suggestion or a more low level alternative?
> 2. The base64 module now takes the type string | ArrayBuffer as argument, do you want to keep this?
> Thanks!

PR #7807, opened 2020-10-03
Fix base64 and base64url modules

timonson:
This PR fixes https://github.com/denoland/deno/issues/6094 and fixes https://github.com/denoland/deno/issues/4794 .

In the `base64` module was one big bug which is illustrated by the following example:

```typescript
// The correct base64 encoding of the letter 'ß' is 'w58='.

import { assertEquals } from "https://deno.land/std/testing/asserts.ts";
import { encode } from "https://deno.land/std/encoding/base64.ts";

assertEquals(encode("ß"), "w58="); // false

```

Additionally, the `encode` and `decode` functions take `Uint8Array` instead of `ArrayBuffer` as arguments now.

Please check out the mentioned issues for more details.


std/encoding/base64url.ts
```diff
@@ -1,46 +1,47 @@
...
 }
 
-function convertBase64ToBase64url(base64: string): string {
-  return base64.replace(/=/g, "").replace(/\+/g, "-").replace(/\//g, "_");
+/**
+ * @param string
+ */
+export function convertBase64ToBase64url(b64: string): string {
+  return b64.replace(/=/g, "").replace(/\+/g, "-").replace(/\//g, "_");
 }
```
bartlomieju, line 31: Why are those functions exported?

std/encoding/base64url.ts
```diff
@@ -1,46 +1,47 @@
...
-  if (base64url.length % 4 === 1) {
-    throw new TypeError("Illegal base64url string!");
+export function addPaddingToBase64url(b64url: string): string {
+  if (b64url.length % 4 === 2) return b64url + "==";
+  if (b64url.length % 4 === 3) return b64url + "=";
+  if (b64url.length % 4 === 1) {
+    throw new TypeError("Illegal b64url string!");
   }
-  return base64url;
+  return b64url;
```
bartlomieju, line 16: Seems like not needed change that makes diff bigger

std/encoding/base64url.ts
```diff
@@ -1,46 +1,47 @@
...
-  return convertBase64ToBase64url(convertArrayBufferToBase64(data));
+export function encode(uint8: Uint8Array): string {
+  return convertBase64ToBase64url(base64.encode(uint8));
 }
 
 /**
- * Converts given base64url encoded data back to original
- * @param data input to decode
+ * Decodes a given base64url representation
+ * @param b64url
```
bartlomieju, line 43: Previous param description was more informative...

std/encoding/base64url.ts
timonson: My reason was uniformity compared to the base32 module. But point taken. Do you still want me to change this?

bartlomieju: Yes, please

james-and (approved):

std/encoding/base64url.ts
timonson: Fixed

timreichen:
@timonson maybe the ` const base64abc = […]` could be set outside the `decode` function, so it isn't initialized on every function call.

kitsonk:
CI is failing due to changes in this PR: https://github.com/denoland/deno/pull/7807/checks?check_run_id=1207781092#step:17:1578

timonson:
> CI is failing due to changes in this PR: https://github.com/denoland/deno/pull/7807/checks?check_run_id=1207781092#step:17:1578

Is it because I changed the parameters from `ArrayBuffer` to `Uint8Array` @kitsonk ? I could make a union type which also accepts ArrayBuffer.

timreichen:
I think you need to adjust std/node/buffer.ts since it is dependant of this module.
Best you run all std tests as described here: https://deno.land/manual@v1.4.4/contributing/development_tools

std/encoding/base64.ts
```diff
@@ -1,41 +1,139 @@
...
+  return decodeURIComponent(
+    Array.prototype.map
+      .call(atob(b64), function (c) {
+        return "%" + ("00" + c.charCodeAt(0).toString(16)).slice(-2);
+      })
+      .join(""),
+  );
 }
 
+const base64abc = [
```
kitsonk, line 29: please widen it out to 80 cols manually and use `// deno-fmt-ignore` before the block to disable reformatting... the vertical space is really annoying.

kitsonk (changes requested):
This is a fairly material breaking change as it is written...  There is no need to introduce another API...  It should simply be aligned to the other encoding APIs like:

```ts
export function encode(data: string | Uint8Array): string
```

An branch in the function if needed.

std/encoding/base64.ts
timreichen: Another possibility would be to do it similar to std/encoding/base32:

```ts
const base64abc = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".split("");
```

timonson:
> This is a fairly material breaking change as it is written... There is no need to introduce another API... It should simply be aligned to the other encoding APIs like:
> 
> ```ts
> export function encode(data: string | Uint8Array): string
> ```
> 
> An branch in the function if needed.

What *other encoding APIs* are you referring to @kitsonk ? Hex's and base32's `encoding` functions take `Uint8Array` as argument, too. Can you please be more specific?

std/encoding/base64.ts
timonson: I like both suggestions. @kitsonk 's suggestions has the advantage that it doesn't use another evaluation. So which one is it deno devs?

timonson: Fixed

kitsonk:
I meant that the other encoding APIs have a single `encode()` function.  There is no reason to introduce a breaking change.  `encode()` for base64 can accept a `string` _or_ a `Uint8Array` without problem or confusion.  No need to add `encodeString()`.

> About the opposite of what you are saying are complaints in the issue I quoted in this PR: #4794

I am not sure what you are talking about.  All I see there is you posing a question with no responses.

timonson:
Sorry about the misunderstanding @kitsonk ! You were correct about that. Could you please review the last changes?

std/encoding/base64.ts
```diff
@@ -9,13 +9,16 @@ const base64abc = ["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L",
...
  * CREDIT: https://gist.github.com/enepomnyaschih/72c423f727d395eeaa09697058238727
- * Encodes a given Uint8Array or string into RFC4648 base64 representation
+ * Encodes a given Uint8Array, ArrayBuffer or string into RFC4648 base64 representation
  * @param data
  */
-export function encode(data: Uint8Array | string): string {
-  const uint8 = typeof data === "string"
-    ? new TextEncoder().encode(data)
-    : data;
+export function encode(data: Uint8Array | ArrayBuffer | string): string {
```
kitsonk, line 15: I would be fine with dropping the `ArrayBuffer` as the other APIs don't support it, but `Uint8Array` is assignable to `ArrayBuffer` (all typed arrays are also `ArrayBuffer`s).  [You can see from the following code, just having `ArrayBuffer` is actually valid](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXxHR1AAooAjALngEEYYoBPAIWUURBngB94BnDDCyoA5gEpqAoaIDcAKDlg8A+MgAc8ALzxUIAO7wAqsIxq6DRiTHy5hJaXXW5QA).  Your coercion to Uint8Array though is good:

```ts
declare function encode(ab: ArrayBuffer | string): string;

const u8 = new Uint8Array();

encode(u8);
```

The problem is that all the other encodes don't use `ArrayBuffer` (I would argue they should) and it clearly confused some people, not realising a `Uint8Array` is assignable to an `ArrayBuffer` (and not supporting coercing other typed arrays to `Uint8Array`.

timonson: Thank you for correcting me. I just fixed it.
 
I think you were right with regards to `ArrayBuffer` anyways because as you can see my commit `adc874e` (where I removed `ArrayBuffer`) failed one test in `/node/string_decoder_test.ts`. I was not able to fix it without changing a lot of code in `string_decoder.ts`.

kitsonk (approved):
LGTM, just waiting for the CI to finish... Thanks for the bumpy ride on this one!

PR #7808, opened 2020-10-03
docs(std/wasi): fix usage example

caspervonb:
The usage example is a bit out of date and not compatible with the  stricter definition of the WebAssembly namespace.

This fixes that and makes it a bit cleaner.


lucacasonato (approved):
LGTM. Thanks Casper

PR #7809, opened 2020-10-03
docs(std/encoding): missing JSDocs for std/encoding

AtakanErmis:
I added JSdoc for std/encoding/toml.ts and std/encoding/hex.ts (#7487)


lucacasonato (approved):
LGTM - thanks

PR #7810, opened 2020-10-03
v1.4.4

bartlomieju:
Hot-fix release to avoid 100% CPU usage in REPL.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -393,7 +393,7 @@ dependencies = [
 
 [[package]]
 name = "deno"
-version = "1.4.3"
+version = "1.4.4"
 dependencies = [
  "atty",
  "base64 0.12.3",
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -450,7 +450,7 @@ dependencies = [
 
 [[package]]
 name = "deno_core"
-version = "0.60.0"
```
lucacasonato, line 453: `deno_core` + `deno_web` + `deno_fetch` + `std` don't need a minor bump right? Patch should be enough.

bartlomieju: CI is setup to release those crates on tag, it's just easier to do that than releasing just cli.

lucacasonato: I'm not saying we shouldn't bump them. Im saying instead of incrementing the minor version, we could increment the patch version. Doesn't really matter though.

ry (approved):
Thanks for getting this fix out / LGTM

Issue #1928, opened 2019-03-13
FR: Syntax highlighting in REPL


andykais:
It would be awesome if the deno repl could include syntax highlighting of expressions _as you type_.
This is not something we have in either the node repl or ts-node repl, but nonetheless would make for a much more enjoyable repl experience. 

Coming from scala-land, [dotty](https://dotty.epfl.ch/) (the next gen compiler for scala) for example, includes a colored repl by default
<img width="295" alt="Screen Shot 2019-03-13 at 4 55 43 PM" src="https://user-images.githubusercontent.com/6627147/54314011-e7fa1e00-45b0-11e9-8263-9e0806a7a24b.png">

some other examples include:
- [ammonite](http://ammonite.io/#Ammonite-REPL) (improved scala repl)
- [zdharma/fast-syntax-highlighting](https://github.com/zdharma/fast-syntax-highlighting) zsh plugin that highlights valid commands
- [bpython](https://bpython-interpreter.org) (fancy python shell)

pedropaulosuzuki:
That would be nice indeed!

andykais:
I just upgraded to 1.5.0 and I see this syntax highlighting on the repl! Awesome work guys

PR #7811, opened 2020-10-03
feat(cli/repl): add regex based syntax highlighter

caspervonb:
This adds a simple regex replace based highlighter to the repl editor.
Overall it tries to match the color palette of `Deno.inspect`.

Closes https://github.com/denoland/deno/issues/1928


caspervonb:
Example:

Screenshot 2020-10-03 at 8 51 22 PM

cli/repl.rs
```diff
@@ -29,6 +35,82 @@ impl Validator for Helper {
...
+  lexer: Regex,
+}
+
+impl SyntaxHighlighter {
+  fn new() -> Self {
+    let lexer = Regex::new(
+      r#"(?x)
+      (?P<comment>(?:/\*[\s\S]*?\*/|//[^\n]*)) |
+      (?P<string>(?:"([^"\\]|\\.)*"|'([^'\\]|\\.)*'|`([^`\\]|\\.)*`)) |
+      (?P<regexp>/(?:(?:\\/|[^\n/]))*?/) |
```
nayeemrmn, line 70: Can this include flags at the end?

caspervonb: Regexp? sure.

bartlomieju (commented):
Looks great @caspervonb, slating for 1.5.0

bartlomieju (approved):
LGTM, thank you @caspervonb for this feature, it's extremely useful. I also like how simple the implementation is. 

I'd be great to have a way to test that, but it's the same problem as with all integration tests.

cli/repl.rs
```diff
@@ -32,6 +38,82 @@ impl Validator for Helper {
...
+}
+
+impl LineHighlighter {
+  fn new() -> Self {
+    let regex = Regex::new(
+      r#"(?x)
+      (?P<comment>(?:/\*[\s\S]*?\*/|//[^\n]*)) |
+      (?P<string>(?:"([^"\\]|\\.)*"|'([^'\\]|\\.)*'|`([^`\\]|\\.)*`)) |
+      (?P<regexp>/(?:(?:\\/|[^\n/]))*?/[gimsuy]*) |
+      (?P<number>\d+(?:\.\d+)*(?:e[+-]?\d+)*n?) |
```
AliBasicCoder, line 74: `(?:\.\d+)*` should be `(?:\.\d+){1}`

AliBasicCoder: @caspervonb

AliBasicCoder, line 74: @caspervonb  `(?:\.\d+)*` should be `(?:\.\d+){0, 1}` to prevent numbers from having more that one dot

cli/repl.rs
```diff
@@ -32,6 +38,82 @@ impl Validator for Helper {
...
+      (?P<boolean>\b(?:true|false)\b) |
+      (?P<null>\b(?:null)\b) |
+      (?P<undefined>\b(?:undefined)\b) |
+      (?P<keyword>\b(?:await|async|var|let|for|if|else|in|of|class|const|function|yield|return|with|case|break|switch|import|export|new|while|do|throw|catch)\b) |
```
AliBasicCoder, line 78: missing `this`

cli/repl.rs
caspervonb: As for your deleted comment.

> (?:\.\d+){1}

Would not capture integers.

> (?:\.\d+){0, 1}

Sure, could also be a ? quantifier.

cli/repl.rs
caspervonb: Indeed, treating it as identifier (no color) for now.

PR #7812, opened 2020-10-03
docs(std/wasi): remove outdated testing section

caspervonb:
Tests are brought in as pre-compiled binaries via a submodule leaving the testing section out of date.

This removes it as it's no longer relevant.


lucacasonato (approved):
Thanks

PR #7816, opened 2020-10-04
chore: add lockfile tests

### LINKED ISSUE #7653: (cli) lockfile.rs should have unit tests
There are not unit tests in `lockfile.rs` 😢  We should have some.

JavascriptMick:
@kitsonk  I have added some tests.  Feedback welcome.

fixes #7653

### DIFF
--- a/cli/lockfile.rs
+++ b/cli/lockfile.rs
@@ -206,6 +206,6 @@ mod tests {
     assert!(check_true);
 
     let check_false = lockfile.check_or_insert("https://deno.land/std@0.71.0/textproto/mod.ts", "This is new Source code");
-    assert!(check_false == false);    
+    assert!(!check_false);    
   }    
 }
--- a/cli/lockfile.rs
+++ b/cli/lockfile.rs
@@ -87,9 +87,9 @@ impl Lockfile {
 mod tests {
   use super::*;
   use std::fs::File;
+  use std::io::prelude::*;
   use std::io::Write;
   use tempfile::TempDir;
-  use std::io::prelude::*;
 
   #[test]
   fn new_nonexistent_lockfile() {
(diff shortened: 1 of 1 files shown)


cli/lockfile.rs
```diff
@@ -82,3 +82,158 @@ impl Lockfile {
...
+  }
+
+  fn write_valid_lockfile(t: &TempDir) {
+    // create a valid lockfile for us to load
+    let file_path = t.path().join("valid_lockfile.json");
+    let mut file = File::create(file_path).expect("write file fail");
+    writeln!(file, "{{").expect("write line fail");
+    writeln!(file, "  \"https://deno.land/std@0.71.0/textproto/mod.ts\": \"3118d7a42c03c242c5a49c2ad91c8396110e14acca1324e7aaefd31a999b71a4\",").expect("write line fail");
+    writeln!(file, "  \"https://deno.land/std@0.71.0/async/delay.ts\": \"35957d585a6e3dd87706858fb1d6b551cb278271b03f52c5a2cb70e65e00c26a\"").expect("write line fail");
+    writeln!(file, "}}").expect("write line fail");
```
bartlomieju, line 107: Nitpick: construct content of the file using `json!()` macro

cli/lockfile.rs
```diff
@@ -82,3 +82,158 @@ impl Lockfile {
...
+  }
+
+  #[test]
+  fn new_valid_lockfile() {
+    let t = TempDir::new().expect("tempdir fail");
+    write_valid_lockfile(&t);
+
+    let file_path_buf = t.path().join("valid_lockfile.json");
```
bartlomieju, line 115: This line is duplicated in every test. Maybe replace `write_valid_lockfile` with `setup` that returns a) lockfile, b) tempdir c) lockfile path

cli/lockfile.rs
```diff
@@ -82,3 +82,158 @@ impl Lockfile {
...
+    let file_path = file_path_buf.to_str().expect("file path fail").to_string();
+
+    let result = Lockfile::new(file_path, false);
+
+    let lockfile = match result {
+      Ok(lockfile) => lockfile,
+      Err(error) => panic!("Lockfile was not created: {:?}", error),
+    };
```
bartlomieju, line 123: let result = Lockfile::new(file_path, false).unwrap();

james-and (approved):

cli/lockfile.rs
JavascriptMick: I have implemented a setup() and teardown() approach and it is looking a lot better.  I didn't end up returning the lockfile from setup because in at least 1 of the tests, the Lockfile constructor is part of the code that is 'under test' and also, I would have needed to parameterize the setup because some are constructed write=true and some write=false which feels a little labored.

kitsonk (approved):
LGTM... Thanks a lot for the contribution.

PR #7817, opened 2020-10-04
fix(core): handle unregistered errors in core better

kitsonk:
If for some reason a `res.err.className` is not registered, we are throwing a totally non-sensical assertion error, even though we had valid information about the error which we could provide, since there errors get raised when reading the results of an Op, it is much better to err on the side of the user and go ahead and raise an error, indicating that the error class was not registered and provide the error message returned from the Rust op.

### DIFF
--- a/core/core.js
+++ b/core/core.js
@@ -213,6 +213,20 @@ SharedQueue Binary Layout
   let nextPromiseId = 1;
   const promiseTable = {};
 
+  function processResponse(res) {
+    if ("ok" in res) {
+      return res.ok;
+    } else {
+      const ErrorClass = getErrorClass(res.err.className);
+      if (!ErrorClass) {
+        throw new Error(
+          `Unregistered error class: "${res.err.className}"\n  ${res.err.message}\n  Classes of errors returned from ops should be registered via Deno.core.registerErrorClass().`,
+        );
+      }
+      throw new ErrorClass(res.err.message);
+    }
+  }
+
   async function jsonOpAsync(opName, args = {}, ...zeroCopy) {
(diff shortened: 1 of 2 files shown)


core/core.js
```diff
@@ -233,7 +231,13 @@ SharedQueue Binary Layout
     if ("ok" in res) {
       return res.ok;
     } else {
-      throw new (getErrorClass(res.err.className))(res.err.message);
+      const ErrorClass = getErrorClass(res.err.className);
+      if (!ErrorClass) {
+        throw new Error(
+          `Unregistered Deno core error class: ${res.err.className}\n  ${res.err.message}`,
```
bartlomieju, line 237: `Unregister error class: "<error_class>"\n <message>\n Did you forget to register error using Deno.core.registerErrorClass?`

core/core.js
```diff
@@ -244,7 +248,13 @@ SharedQueue Binary Layout
...
     } else {
-      throw new (getErrorClass(r.err.className))(r.err.message);
+      const ErrorClass = getErrorClass(res.err.className);
+      if (!ErrorClass) {
+        throw new Error(
+          `Unregistered Deno core error class: ${res.err.className}\n  ${res.err.message}`,
+        );
+      }
+      throw new ErrorClass(res.err.message);
     }
```
bartlomieju, line 258: `s/res/r/`

core/core.js
kitsonk: Only thought about this is the is it really going to often not the end-user who did this, so stating "Did you forget..." is highly unlikely...  I will try some other words.

bartlomieju (approved):
LGTM

PR #7819, opened 2020-10-04
feat(extensions): Add WebStorage API

### LINKED ISSUE #1657: Implement localStorage
Implement `localStorage` and `sessionStorage` following [the MDN spec](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)

crowlKats:
Closes #1657

This PR does not implement the storage events.


crowlKats:
Closes #1657

cli/standalone.rs
```diff
@@ -188,6 +188,7 @@ pub async fn run(
     no_color: !colors::use_color(),
     get_error_class_fn: Some(&get_error_class_name),
     location: metadata.location,
+    location_data_dir: None, // TODO
```
crowlKats, line 191: not sure what to do here

crowlKats:
Only missing things are typings & tests

cli/standalone.rs
axetroy: Is this path `$DENO_DIR/storage/<hash of domain>` ok?

eg. 

```bash
deno run --location=example.com mod.ts
```

then `location_data_dir` should be `$DENO_DIR/storage/5ababd603b22780302dd8d83498e5172`

```bash
tree $DENO_DIR

├── bin
├── deps
│   ├── http
│   └── https
├── gen
└── storage
```

crowlKats: The issue is in regards to standalone though, so cant exactly just store it in `$DENO_DIR`, right=

ry:
I'm not sure about using sqlite for localStorage. Sqlite like quite a lot of code (100k lines of C) for a simple key/value store. I could imagine a much smaller backend store where we simply use a directory of files, one for each KV pair, and [a file system lock](https://docs.rs/fs2/0.4.3/fs2/trait.FileExt.html#notes-on-file-locks) to manage concurrent access.  Probably this could be done in 50 lines of Rust compared to the 100k of C that is Sqlite.

There are also other solutions: leveldb, rocksdb, [sled](https://github.com/spacejam/sled). We need some analysis of the tradeoffs between these various local databases and our future use-cases for them (particularly for IndexDB).

I'd love to get this feature in - so I suggest this very simple (very suboptimal) directory of files storage backend - meanwhile we can figure out the long term database needs for Deno. Or we can do this analysis now...

crowlKats:
I have already looked at leveldb, rocksdb, & sled. concurrent access with those is **not** possible, so those are already not an option. The reason for SQLite is because IndexedDB. Firefox uses SQLite for IndexedDB. Using SQLite was the decision i came up with after looking at various possible DBs and what we can use use for both Web Storage and IndexedDB. I dont mind changing, till it covers the needed requirements

ry:
> I have already looked at leveldb, rocksdb, & sled. concurrent access with those is not possible

Parallel process access is certainly possible - it just isn't handled by the database itself. Those libraries have decided to keep it simple and leave the FS locking to the caller; that doesn't mean they are inherently unusable.

crowlKats:
@ry i have tried rocksdb, leveldb & sled for the implementation.

sled is not an option. just running WPT on it causes it to fire `DomExceptionQuotaExceededError`, which means it exceeds 5MB for a single test... sure it is one of the longer tests (the allowed characters check test), but still, this isnt really acceptable.

rocksdb is somewhat painful to use, and the crate doesnt implement all functions. also building causes an invalid memory reference error, so i couldnt test it.

leveldb is better to use than rocksdb, but the crate is not actively maintained (last commit 7 months ago), and doesnt include all functions, making a few things complicated.

ry:
I think I messed up the merge commit somehow and am getting an error like this
```
# deno run --unstable --allow-write --allow-read --allow-net --allow-env --allow-run ./tools/wpt.ts run
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
Following tests are missing in manifest, but are present in expectations:

/webstorage/defineProperty.window.js
/webstorage/set.window.js
/webstorage/storage_enumerate.window.js
/webstorage/storage_in.window.js
/webstorage/event_constructor.window.js
...
```
I assume there's a simple fix, but I'm not able to immediately find it.

Otherwise LGTM. Very nice work @crowlKats this is a great feature to have. Thank you very much for putting up with the endless reviews and questioning of backends.

PR #7823, opened 2020-10-04
perf(cli/console): Don't add redundant ANSI codes

nayeemrmn:


bartlomieju (approved):
LGTM, thanks @nayeemrmn

PR #7825, opened 2020-10-04
refactor(core): JsRuntime::poll

bartlomieju:
Prerequisite for https://github.com/denoland/deno/pull/7672

This commit does reorganization of "JsRuntime::poll" to allow fixing of top-level-await bug.

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -1147,16 +1147,16 @@ impl JsRuntime {
     &mut self,
     maybe_buf: Option<(OpId, Box<[u8]>)>,
   ) -> Result<(), AnyError> {
+    let js_recv_cb_handle = Self::state(self)
+      .borrow()
+      .js_recv_cb
+      .clone()
+      .expect("Deno.core.recv has not been called.");
     let context = self.global_context();
     let scope = &mut v8::HandleScope::with_context(self.v8_isolate(), context);
     let context = scope.get_current_context();
     let global: v8::Local<v8::Value> = context.global(scope).into();
-    let js_recv_cb = JsRuntime::state(scope)
-      .borrow()
-      .js_recv_cb
-      .as_ref()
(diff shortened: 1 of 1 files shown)


james-and (approved):

core/runtime.rs
```diff
@@ -353,15 +357,9 @@ impl JsRuntime {
...
     self.shared_init();
 
-    let state_rc = Self::state(self);
-    let state = state_rc.borrow();
-
-    let scope = &mut v8::HandleScope::with_context(
-      self.v8_isolate.as_mut().unwrap(),
-      state.global_context.as_ref().unwrap(),
-    );
+    let context = self.global_context();
```
bartlomieju, line 360: This calls are creating clones of `v8::Global` instead of using `Option<&v8::Global<>>`, but I haven't noticed any perf hit on benchmarks

core/runtime.rs
```diff
@@ -310,6 +310,10 @@ impl JsRuntime {
     state.global_context.clone().unwrap()
   }
 
+  fn v8_isolate(&mut self) -> &mut v8::OwnedIsolate {
+    self.v8_isolate.as_mut().unwrap()
+  }
```
ry, line 315: Don't we implement DerefMut ?

core/runtime.rs
```diff
@@ -328,7 +332,7 @@ impl JsRuntime {
   }
 
   /// Executes a bit of built-in JavaScript to provide Deno.sharedQueue.
-  pub(crate) fn shared_init(&mut self) {
+  fn shared_init(&mut self) {
```
ry, line 335: Good

ry (approved):
LGTM

core/runtime.rs
bartlomieju: We do, I find passing `self.v8_isolate()` more obvious than `&mut **self`

PR #7826, opened 2020-10-05
[WIP] new bundler

kitsonk:
This PR is the WIP for the way of bundling.  It does the following:

- Builds on the new module graph that was introduced.
- Uses swc_bundler to emit the bundle.
- Uses a refactoring of `tsc` called `ts_checker` which does a type check only.

Like currently, `deno bundle` does a type check, but it now supports the incremental build for bundles, which make subsequent type checks faster.  It also supports `deno bundle --no-check` as well.  In all cases it uses `swc_bundler` which emits a single file "flat" ES module.

Things going on:

- This builds on #7669, but I would like to land #7669 separately when we get the majority of the bugs addressed, and then use this PR to replace the the emit from `deno bundle` with swc.
- swc leaves dynamic imports untouched.  This is a significant change in behaviour from `deno bundle`, where if a dynamic import could be statically identified, it would be included in the bundle.  This isn't a big deal with absolute URLs, but if users are importing something relative and locally, then this becomes a bit of an issue, as it doesn't make the bundle very portable.  We might want to consider ways of dealing with this (like maybe warning on dynamic import).  Also, this is a significant change in behaviour, and some could argue a breaking change.
- This is a net add of ~1100 lines...  I am not totally happy about that, but again, this is a transition period... Not all of the lines could be dropped in tsc because of the other dependent capabilities.  I believe at the end of this journey we will drop a lot of code and have a set of code that is a lot more maintainable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,6 +213,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "build_const"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39092a32794787acd8525ee150305ff051b0aa6cc2abaf193924f5ab05425f39"
+
+[[package]]
 name = "bumpalo"
 version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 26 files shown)


kitsonk:
Reworking a few things.  Closing without merging.

Issue #3086, opened 2019-10-07
Add tab completion in REPL


sszczepa:
One of the things that always bothers me about Node is that tab completion in the REPL doesn't work like PowerShell. I believe there is an opportunity to correct this with Deno.

In PowerShell, I can simply start the name of a command/file/etc... and cycle through of all the matches with tab (and in reverse order with shift + tab). For instance, typing in _Get-P_ and pressing shift will cycle through _Get-Package_, _Get-PackageProvider_, _Get-PackageSource_, etc... This is extremely helpful when searching for a command in the middle of writing some long pipeline, such as:

`Get-Process | Where cpu -gt 15 | Select name,cpu,description | ConvertTo-Json`

If for some reason I didn't remember _ConvertTo-Json_, which is the command for converting the objects in the pipeline to a JSON formatted string, all I would have to do is type in _convert_ and press shift a few times. This would quickly lead me to _ConvertTo-Json_ without having to stop mid-way through to look up the commandlet.

I see a properly implemented tab completion as extremely useful as it greatly speeds up the process of writing short scripts in the REPL. Wouldn't it be nice in Deno if you could type in any arbitrary string, like _con_, and then press tab (and shift+tab to reverse) to cycle through all matching objects/variables until you found _console_. From there you type the period, i.e.  _console._, and then again press tab to cycle through all of the available properties/methods (_console.log_, _console.warn_, _console.dir_, etc...).

bitjson:
Just to note, the "cycling" style @sszczepa described is also how tab completion works in `zsh` and `fish` by default too.

The Node.js tab completion style matches the "tab-to-finish" completion style of default `bash` on most systems. E.g. if options  are `option-one` and `option-two`, completion will stop at `option-` until the user types a character to disambiguate between the remaining completion options. With `zsh`, the first tab will complete to `option-one`, and tabbing again will cycle to the next possible completion, `option-two`.

With `zsh` now the default for macOS users, I imagine even more systems will start defaulting to the "cycling" style as more users grow to expect it. It would be great if Deno did this from the start. 👍

dubiousjim:
If we do implement this, please let it be something we can also turn off.

bitjson:
> If we do implement this, please let it be something we can also turn off.

[...]

PR #7827, opened 2020-10-05
feat(cli/repl) add tab completion

caspervonb:
This adds tab completion for global identifiers and object properties.

Closes https://github.com/denoland/deno/issues/3086


cli/repl.rs
```diff
@@ -59,17 +203,27 @@ async fn post_message_and_poll(
...
+  response_tx: &Sender<Result<Value, AnyError>>,
   editor: Arc<Mutex<Editor<Helper>>>,
 ) -> Result<String, ReadlineError> {
   let mut line =
     tokio::task::spawn_blocking(move || editor.lock().unwrap().readline("> "));
 
   let mut poll_worker = true;
+
   loop {
+    for (method, params) in message_rx.try_iter() {
```
caspervonb, line 217: Make this a stream and select on it below.

caspervonb:
Alright so right the first pass only supports a few simple scenarios single line scenarios

```
Deno.|<tab>
Deno.write|<tab>
Deno[|<tab>
Deno["write|<tab>
```

Next step would be for completions for multi-line "lines" based on the current cursor position

```
const foo = Deno.|<tab>
(
  Deno.write<tab>
)
false, Deno[<tab>
true && Deno["write<tab>
```

And also word completions

```
De|<tab>
let foo = win|<tab>
(
  con|<tab>
)
```

ry:
Tested it locally - seems to work! Very nice.

It only works for `Deno.|<tab>` tho. When I do `windo|<tab>` it doesn't complete to "window".

caspervonb:
> . When I do windo|<tab> it doesn't complete to "window".

Yet to be implemented, only the rhs of propery accessors are completed at the moment 🙃

crypt096 (approved):

caspervonb:
Slightly confusing, but line in a completer is actually a line. Whereas a line in a highlighter is all lines... so multi-line completion like the example below does not work at the moment

```
(
window
).|<tab>
```

bartlomieju:
@caspervonb please rebase

caspervonb:
Will do, merge #7896 first please :)

bartlomieju (approved):
LGTM, very cool feature 🚀 

In future we should consider changing colors a bit for autocompleted items or even displaying list of all possible suggestions.

caspervonb:
There are other completion modes available, specifically there's a list list-mode that interoperates with "hints" (the one trait we haven't touched on yet in rustyline).

So yeah that's open ended and should be relatively accessible for tweaking.

PR #7830, opened 2020-10-05
fix(cli/console): only inspect getters with option

caspervonb:
This makes evaluation of getters an explicit option called `getters` exposed via InspectOptions.
This option defaults to false fixing the current bad behaviour of inspecting getters which has the potential to cause side effects, when it is false a getter is simply reported as `[Getter]` or `[Getter/Setter]` if its a get/set pair.

Fixes https://github.com/denoland/deno/issues/7153


james-and (approved):

james-and:
when will it be running james

bartlomieju:
@nayeemrmn PTAL

nayeemrmn (approved):
LGTM

bartlomieju (approved):
LGTM

PR #7834, opened 2020-10-06
refactor(cli/repl): tightly integrate event loop

caspervonb:
This tightly integrates the event loop into the repl event loop for more control over it so it doesn't have to spin wildly outside of the inner loop for us to get the inspector to pump.

### DIFF
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -63,12 +63,18 @@ async fn read_line_and_poll(
 
   let mut poll_worker = true;
   loop {
+    let mut timeout =
+      tokio::time::delay_for(tokio::time::Duration::from_millis(1000));
+
     tokio::select! {
       result = &mut line => {
         return result.unwrap();
       }
       _ = &mut *worker, if poll_worker => {
-          poll_worker = false;
+        poll_worker = false;
+      }
+      _ = &mut timeout => {
+          poll_worker = true
       }
     }
   }
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -61,13 +61,14 @@ async fn read_line_and_poll(
   let mut line =
     tokio::task::spawn_blocking(move || editor.lock().unwrap().readline("> "));
 
+  let mut poll_worker = true;
   loop {
     tokio::select! {
       result = &mut line => {
         return result.unwrap();
       }
-      _ = &mut *worker => {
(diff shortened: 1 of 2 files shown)


cli/repl.rs
```diff
@@ -29,18 +32,70 @@ impl Validator for Helper {
   }
 }
 
+async fn post_message_and_poll(
+  session: &mut InspectorSession,
+  method: String,
```
bartlomieju, line 37: nitpick: use `&str`

cli/repl.rs
```diff
@@ -29,18 +32,70 @@ impl Validator for Helper {
...
+  params: Option<Value>,
+  worker: &mut Worker,
```
bartlomieju, line 39: nitpick, make it first or second argument

caspervonb: All the arguments are selected here but I'll assume you mean the worker 🙃

nayeemrmn: It's a (bottom) one line selection unless it says `Comments on line x to y`.

cli/repl.rs
```diff
@@ -29,17 +32,64 @@ impl Validator for Helper {
...
+  let mut line =
+    tokio::task::spawn_blocking(move || editor.lock().unwrap().readline("> "));
+
+  loop {
+    tokio::select! {
+      result = &mut line => {
+        return result.unwrap();
+      }
+      _ = &mut *worker => {
+        tokio::time::delay_for(tokio::time::Duration::from_millis(0)).await;
```
caspervonb, line 70: Just returning `line.await` worked great locally but completely locks up the CI with no test output at all.

cli/repl.rs
```diff
@@ -29,17 +32,76 @@ impl Validator for Helper {
...
+    tokio::select! {
+      result = &mut response => {
+        return result
+      }
+
+      _ = &mut *worker => {
+        // A zero delay is long enough to yield the thread in order to prevent the loop from
+        // running hot for messages that are taking longer to resolve like for example an
+        // evaluation of top level await.
+        tokio::time::delay_for(tokio::time::Duration::from_millis(0)).await;
```
bartlomieju, line 54: Thanks for the comment

bartlomieju (approved):
LGTM, thanks for the fix @caspervonb

PR #7835, opened 2020-10-06
feat(std/node/fs): allow appendFileSync to accept Uint8Array as type for data

AliBasicCoder:
this makes appendFileSync to accept Uint8Array as type for data


ry (approved):
LGTM - thanks!

ry (changes requested):
Oops didn't realize this was red. Needs to be green - otherwise looks good.

AliBasicCoder:
> Oops didn't realize this was red. Needs to be green - otherwise looks good.

it's fixed now

ry (approved):

PR #7369, opened 2020-09-06
feat: Add --location=<href> and globalThis.location

nayeemrmn:
Closes #4981.

https://github.com/nayeemrmn/deno/blob/location/docs/runtime/location_api.md

PR #7838, opened 2020-10-06
fix(op_crates/fetch): Stringify and parse Request URLs

### LINKED ISSUE #7837: new Request(undefined).url should return "undefined"
See https://github.com/denoland/deno/issues/7836

This is what Chrome, FF, and Safari do, and it is intended behaviour:

- The constructor takes a union type https://fetch.spec.whatwg.org/#requestinfo
- Here's how to process a union type https://heycam.github.io/webidl/#es-union
- Which falls through to step 12
- Which sends us to https://heycam.github.io/webidl/#es-USVString
- Which sends us to https://heycam.github.io/webidl/#es-DOMString
- Which sends us to https://tc39.es/ecma262/#sec-tostring
- Which returns the string "undefined"
- So we process the request constructor with the string input, which takes us to step 6 https://fetch.spec.whatwg.org/#dom-request

nayeemrmn:
Fixes #7837.

### DIFF
--- a/op_crates/fetch/26_fetch.js
+++ b/op_crates/fetch/26_fetch.js
@@ -985,8 +985,9 @@
         this.headers = new Headers(input.headers);
         this.credentials = input.credentials;
         this._stream = input._stream;
-      } else if (typeof input === "string") {
-        this.url = input;
+      } else {
+        // TODO(nayeemrmn): Base from `--location` when implemented and set.
+        this.url = new URL(String(input)).href;
       }
 
       if (init && "method" in init) {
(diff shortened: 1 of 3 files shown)


op_crates/fetch/26_fetch.js
```diff
@@ -985,8 +985,9 @@
         this.headers = new Headers(input.headers);
         this.credentials = input.credentials;
         this._stream = input._stream;
-      } else if (typeof input === "string") {
-        this.url = input;
+      } else {
+        // TODO(nayeemrmn): Base from `--location` when implemented and set.
+        this.url = new URL(String(input)).href;
```
nayeemrmn, line 990: Relative URLs, including the original repro `new Request(undefined).url`, are now correctly blocked by #7369.

james-and (approved):

kitsonk (approved):
LGTM

PR #7843, opened 2020-10-06
fix: update worker types to better align to lib.dom.d.ts

### LINKED ISSUE #7839: worker scope onmessage/MessageEvent bad typings
`self.onmessage = async (msg: MessageEvent) => {}` wont work because 
```
error: Uncaught Error: TS2322 [ERROR]: Type '(msg: MessageEvent) => Promise<void>' is not assignable to type '((e: MessageEvent) => void) & ((e: { data: any; }) => void | Promise<void>)'.
  Type '(msg: MessageEvent) => Promise<void>' is not assignable to type '(e: { data: any; }) => void | Promise<void>'.
    Types of parameters 'msg' and 'e' are incompatible.
      Type '{ data: any; }' is missing the following properties from type 'MessageEvent': origin, lastEventId, bubbles, cancelBubble, and 17 more.
```

and `self.onmessage = async (msg) => {}` wont work because `Parameter 'msg' implicitly has an 'any' type.`
It seems to be happening since 1.4.3.

from the `((e: { data: any; }) => void | Promise<void>)` i can see that `MessageEvent.data` was wanted to be of type `any`, but  from the typescript typings `MessageEvent` should actually take a type param that defines the type of `MessageEvent.data`

kitsonk:
Fixes #7839

### DIFF
--- a/cli/dts/lib.deno.shared_globals.d.ts
+++ b/cli/dts/lib.deno.shared_globals.d.ts
@@ -606,15 +606,20 @@ declare class URL {
   toJSON(): string;
 }
 
-interface MessageEventInit extends EventInit {
-  data?: any;
+interface MessageEventInit<T = any> extends EventInit {
+  data?: T;
   origin?: string;
   lastEventId?: string;
 }
 
-declare class MessageEvent extends Event {
-  readonly data: any;
-  readonly origin: string;
+declare class MessageEvent<T = any> extends Event {
+  /**
+   * Returns the data of the message.
+   */
+  readonly data: T;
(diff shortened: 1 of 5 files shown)


ry:
@crowlKats does this work for you?

crowlKats:
Yep, LGTM

crypt096 (approved):

ry (commented):
It would be great to have a simple unit test for this.

bartlomieju (approved):
LGTM, thanks @kitsonk

PR #7853, opened 2020-10-07
refactor(core): JsRuntime doesn't defer to OwnedIsolate

bartlomieju:
Remove `Deref` and `DeferMut` implementations for `JsRuntime`.

### DIFF
--- a/cli/inspector.rs
+++ b/cli/inspector.rs
@@ -394,7 +394,7 @@ impl DenoInspector {
     server: Option<Arc<InspectorServer>>,
   ) -> Box<Self> {
     let context = isolate.global_context();
-    let scope = &mut v8::HandleScope::new(&mut **isolate);
+    let scope = &mut v8::HandleScope::new(isolate.v8_isolate());
 
     let (new_websocket_tx, new_websocket_rx) =
       mpsc::unbounded::<WebSocketProxy>();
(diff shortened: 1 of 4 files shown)


bartlomieju:
Not sure about it completely - on one hand I prefer to have one level of indirection less (by removing `Deref`), OTOH it means that `v8_isolate()` method must be public. All in all this approach is more verbose and explicit which I think is desirable.

cli/inspector.rs
```diff
@@ -394,7 +394,7 @@ impl DenoInspector {
     server: Option<Arc<InspectorServer>>,
   ) -> Box<Self> {
     let context = isolate.global_context();
-    let scope = &mut v8::HandleScope::new(&mut **isolate);
+    let scope = &mut v8::HandleScope::new(isolate.v8_isolate());
```
ry, line 397: Not now, but we should eliminate variables called "isolate" in CLI where they in fact refer to JsRuntimes

bartlomieju: I agree, it's on my list

ry (commented):
I'm a bit torn on this because Deref seems to be semantically what we want. But I agree that `&mut **isolate` is particularly difficult to read compared to `runtime.v8_isolate()`.

ry (approved):
LGTM

PR #7860, opened 2020-10-07
fix(cli/rt/error_stack): Improve message line formatting

nayeemrmn:
Matches browsers.

Example:
```ts
const error = new Error();
error.name = undefined;
error.message = undefined;
console.log(error.stack);

// Before:
// undefined: undefined
//     at ...

// After:
// Error
//     at ...
```


bartlomieju (approved):
Great! Thanks @nayeemrmn

PR #7864, opened 2020-10-07
refactor(cli/repl): get context id from notification

caspervonb:
This takes the execution context id from a notification which is sent on Runtime.enable rather than hard-coding it to a magic value.


cli/repl.rs
```diff
@@ -294,6 +308,10 @@ pub async fn run(
           ),
         }
 
+        for notification in session.notifications() {
+          println!("{}", notification);
```
bartlomieju, line 312: Is this for debug only?

cli/repl.rs
```diff
@@ -89,8 +89,7 @@ pub async fn run(
   global_state: &GlobalState,
   mut worker: MainWorker,
 ) -> Result<(), AnyError> {
-  // Our inspector is unable to default to the default context id so we have to specify it here.
-  let context_id: u32 = 1;
+  let mut context_id: u64 = 0;
```
bartlomieju, line 92: Is it safe to have `0` here if `executionContextCreated` is not received?

cli/repl.rs
caspervonb: Oops, indeed.

cli/repl.rs
caspervonb: Implementation defined but for us 0 is an invalid context identifier, we start at 1 so this is the uninitialized value.

Per spec, executionContextCreated must always be sent for each context after Runtime.enable has been called.

bartlomieju: Cool, can you add that comment to code?

bartlomieju:
@caspervonb please rebase

bartlomieju (approved):
LGTM

PR #7869, opened 2020-10-07
fix(cli/rt/console): Don't require a prototype to detect a class instance

nayeemrmn:
Matches Node.
```ts
class A {}
console.log(A.prototype);
// Before:
// {}
// After:
// A {}
```


ry (approved):
LGTM - thanks Nayeem

PR #7874, opened 2020-10-07
feat(std/testing): Add assertExists assertion

getspooky:


std/testing/asserts.ts
```diff
@@ -329,6 +329,23 @@ export function assertNotStrictEquals(
...
 }
 
+/**
+ * Make an assertion that actual is not null or undefined. If not
+ * then thrown.
+ */
+export function assertAnything(
+  actual: unknown,
+  msg?: string,
+): void {
```
bartlomieju, line 339: It seems you can already achieve the same assertion using just `assert()`. Is there some edge case not covered by it?

kitsonk: `assert(false)` would throw, while `assertAnything(false)` would not.

std/testing/asserts_test.ts
```diff
@@ -159,6 +160,19 @@ Deno.test("testingNotEquals", function (): void {
   assertEquals(didThrow, true);
 });
 
+Deno.test("testingAssertAnything", function (): void {
+  assertAnything("Denosaurus");
+  let didThrow;
+  try {
```
kitsonk, line 166: We should test all the [falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) values here.

kitsonk (commented):
I would prefer `assertExists()`.

Prior art: https://www.chaijs.com/api/assert/#method_exists

std/testing/asserts.ts
caspervonb: I prefer explicit assertions, if you're checking that something is not null and not undefined there's probably a very specific reason for that.

```typescript
assert(value !== undefined && value !== null);
```

std/testing/asserts_test.ts
getspooky: @kitsonk  I think assertExists it's a good idea

bartlomieju:
Looks good to me, waiting on @kitsonk's approval

std/testing/asserts_test.ts
```diff
@@ -159,6 +160,19 @@ Deno.test("testingNotEquals", function (): void {
   assertEquals(didThrow, true);
 });
 
+Deno.test("testingAssertExists", function (): void {
+  assertExists("Denosaurus");
```
kitsonk, line 164: Still think we need to have proper coverage for this...  Especially with ["falsey" values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).

```ts
assertExists(false);
assertExists(0);
assertExists("");
assertExists(-0);
assertExists(0n);
assertExists(NaN);
```

std/testing/asserts_test.ts
```diff
@@ -159,6 +160,19 @@ Deno.test("testingNotEquals", function (): void {
...
+  let didThrow;
+  try {
+    assertExists(undefined);
```
kitsonk, line 167: We should test `null` and `void 0` in the same way.

kitsonk (dismissed):
Really want to see full coverage in the tests, as we want to really want to be able to trust that something like an assertion will work as designed.

std/testing/asserts_test.ts
getspooky: @kitsonk I added some falsy test

bartlomieju (approved):
LGTM, thank you @getspooky!

PR #7875, opened 2020-10-07
fix(cli/rt/main): Add global interface objects

nayeemrmn:
- `Window`
- `DedicatedWorkerGlobalScope`
- `WorkerGlobalScope`

---

Matches browsers.

Window `globalThis` prototype chain: `Window` -> `EventTarget` -> `Object`.
Worker `globalThis` prototype chain: `DedicatedWorkerGlobalScope` -> `WorkerGlobalScope` -> `EventTarget` -> `Object`.

```ts
console.log(globalThis);

// Before:
// { ... }

// After:
// Window { ... }
```

Use the following to [detect a web worker environment](https://stackoverflow.com/questions/7931182/reliably-detect-if-the-script-is-executing-in-a-web-worker#answer-18002694):
```ts
if (globalThis.WorkerGlobalScope && globalThis instanceof WorkerGlobalScope) {
  // ...
}
```

### DIFF
--- a/cli/dts/lib.deno.worker.d.ts
+++ b/cli/dts/lib.deno.worker.d.ts
@@ -7,48 +7,53 @@
 /// <reference lib="deno.shared_globals" />
 /// <reference lib="esnext" />
 
-declare interface DedicatedWorkerGlobalScope {
-  self: DedicatedWorkerGlobalScope & typeof globalThis;
+declare class WorkerGlobalScope {
+  new(): WorkerGlobalScope;
+  self: WorkerGlobalScope & typeof globalThis;
   onmessage:
     | ((
-      this: DedicatedWorkerGlobalScope & typeof globalThis,
+      this: WorkerGlobalScope & typeof globalThis,
       ev: MessageEvent,
     ) => any)
     | null;
   onmessageerror:
     | ((
-      this: DedicatedWorkerGlobalScope & typeof globalThis,
+      this: WorkerGlobalScope & typeof globalThis,
       ev: MessageEvent,
     ) => any)
     | null;
   onerror:
     | ((
-      this: DedicatedWorkerGlobalScope & typeof globalThis,
+      this: WorkerGlobalScope & typeof globalThis,
       ev: ErrorEvent,
     ) => any)
     | null;
-  name: string;
   close: () => void;
   postMessage: (message: any) => void;
   Deno: typeof Deno;
 }
 
-declare var self: DedicatedWorkerGlobalScope & typeof globalThis;
+declare class DedicatedWorkerGlobalScope extends WorkerGlobalScope {
+  new(): DedicatedWorkerGlobalScope;
+  name: string;
+}
+
+declare var self: WorkerGlobalScope & typeof globalThis;
 declare var onmessage:
   | ((
-    this: DedicatedWorkerGlobalScope & typeof globalThis,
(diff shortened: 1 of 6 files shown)


cli/dts/lib.deno.worker.d.ts
```diff
@@ -7,8 +7,9 @@
 /// <reference lib="deno.shared_globals" />
 /// <reference lib="esnext" />
 
-declare interface DedicatedWorkerGlobalScope {
-  self: DedicatedWorkerGlobalScope & typeof globalThis;
+declare class WorkerGlobalScope extends EventTarget {
+  new(): WorkerGlobalScope;
+  self: WorkerGlobalScope & typeof globalThis;
```
bartlomieju, line 12: I'm a bit puzzeled here... [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope) that `WorkerGlobalScope` is an interface so it shouldn't be available as a value?

Also in `99_main.js` you add `WorkerGlobalScope` as `self.WorkerGlobalScope` but it's not reflected in these definition.

nayeemrmn: I believe MDN wouldn't have a page for it if it didn't exist as an object. I've checked that it exists on Chrome and Firefox in workers. Lots of classes like this including `Window` are called interfaces in the spec but exist as objects so the prototypes are kept somewhere.

> Also in `99_main.js` you add `WorkerGlobalScope` as `self.WorkerGlobalScope` but it's not reflected in these definition.

It's `declare`d at the top-level.

cli/dts/lib.deno.worker.d.ts
```diff
@@ -7,8 +7,9 @@
 /// <reference lib="deno.shared_globals" />
 /// <reference lib="esnext" />
 
-declare interface DedicatedWorkerGlobalScope {
-  self: DedicatedWorkerGlobalScope & typeof globalThis;
+declare class WorkerGlobalScope extends EventTarget {
```
nayeemrmn, line 10: Looks like both `WorkerGlobalScope` and `DedicatedWorkerGlobalScope` exist, need to amend.

cli/dts/lib.deno.worker.d.ts
nayeemrmn: There we go: https://heycam.github.io/webidl/#es-interfaces:
> For every interface that is exposed in a given Realm and that is not declared with the [LegacyNoInterfaceObject] or [LegacyNamespace] extended attributes, a corresponding property exists on the Realm's global object. The name of the property is the identifier of the interface, and its value is an object called the interface object. The characteristics of an interface object are described in § 3.7.1 Interface object.

cli/rt/99_main.js
```diff
@@ -190,6 +191,42 @@ delete Object.prototype.__proto__;
     core.registerErrorClass("URIError", URIError);
   }
 
+  class Window extends EventTarget {
+    constructor(key) {
+      if (key != illegalConstructorKey) {
```
kitsonk, line 196:
```suggestion
      if (key !== illegalConstructorKey) {
```

kitsonk (approved):
LGTM

I hate weak inequality operator, and thought we had a linting rule to enforce it, but it seems we have a mixture of weak and strong equality/inequality in the code base so, 🤷‍♂️.

kitsonk (approved):
LGTM

PR #7885, opened 2020-10-08
docs(std/bytes): add missing docs of std/bytes module

kt3k:
`hasSuffix` and `contains` methods are exported from `std/bytes/mod.ts`, but they are not documented in `std/bytes/README.md`. This PR adds the documents of those 2 functions.


denobot (approved):
Thank you

PR #7892, opened 2020-10-09
refactor(cli): move info subcommand over to new module graph

kitsonk:
This PR moves `deno info` over to the new module graph.

A couple of minor behavioural changes:

* The ordering of dependencies when displayed was not deterministic, as it relied upon the insertion in the module graph, which varied with the two.  Sort dependencies are not sorted.
* There is a line between the module information and the display of the tree. I can revert it, but I found it a lot clearer when viewing the display.
* The JSON serialisation now is ordered with the fields alphabetically, instead of the arbitrary order.
* Type only dependencies are reflected in the dependency tree (things such a `@deno-types`, `X-TypeScript-Types` and the triple slash types reference in JS files), as these are present in "natively" in the new module graph and are required to type check the code.

### DIFF
--- a/cli/info.rs
+++ b/cli/info.rs
@@ -273,14 +273,14 @@ mod test {
     };
     let mut items = HashMap::new();
     items.insert(
-      spec_c.clone(),
+      spec_c,
       ModuleInfoMapItem {
         deps: vec![spec_d.clone()],
         size: 12345,
       },
     );
     items.insert(
-      spec_d.clone(),
+      spec_d,
       ModuleInfoMapItem {
         deps: Vec::new(),
         size: 12345,
--- a/cli/info.rs
+++ b/cli/info.rs
@@ -1,10 +1,10 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
 use crate::colors;
+use crate::media_type::serialize_media_type;
 use crate::MediaType;
 use crate::ModuleSpecifier;
 
-use serde::ser::SerializeSeq;
 use serde::Serialize;
 use serde::Serializer;
 use std::cmp::Ordering;
@@ -12,38 +12,6 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::fmt;
 use std::path::PathBuf;
-
-/// Serialize a `MediaType` enum into a human readable string.  The default
-/// serialization for media types is and integer.
-fn serialize_media_type<S>(mt: &MediaType, s: S) -> Result<S::Ok, S::Error>
-where
-  S: Serializer,
-{
-  s.serialize_str(&format!("{}", mt))
-}
-
-/// Serialize a module specifier to a simple string representation.
-fn serialize_specifier<S>(ms: &ModuleSpecifier, s: S) -> Result<S::Ok, S::Error>
-where
-  S: Serializer,
-{
-  s.serialize_str(&ms.to_string())
-}
-
-/// Serialize a vector of module specifiers.
-fn serialize_specifier_vec<S>(
-  v: &[ModuleSpecifier],
-  s: S,
-) -> Result<S::Ok, S::Error>
-where
-  S: Serializer,
-{
-  let mut seq = s.serialize_seq(Some(v.len()))?;
-  for element in v {
-    seq.serialize_element(&element.to_string())?;
-  }
-  seq.end()
-}
 
 /// The core structure representing information about a specific "root" file in
 /// a module graph.  This is used to represent information as part of the `info`
@@ -60,7 +28,6 @@ pub struct ModuleGraphInfo {
   pub info: ModuleInfo,
   pub local: PathBuf,
   pub map: Option<PathBuf>,
-  #[serde(serialize_with = "serialize_specifier")]
   pub module: ModuleSpecifier,
   pub total_size: usize,
 }
@@ -105,7 +72,7 @@ impl fmt::Display for ModuleGraphInfo {
 
     let dep_count = self.info.deps.len();
     for (idx, dep) in self.info.deps.iter().enumerate() {
-      dep.write_info(f, &"", idx == dep_count - 1)?;
+      dep.write_info(f, "", idx == dep_count - 1)?;
     }
 
     Ok(())
@@ -118,7 +85,6 @@ impl fmt::Display for ModuleGraphInfo {
 #[serde(rename_all = "camelCase")]
 pub struct ModuleInfo {
   pub deps: Vec<ModuleInfo>,
-  #[serde(serialize_with = "serialize_specifier")]
   pub name: ModuleSpecifier,
   pub size: usize,
   pub total_size: Option<usize>,
@@ -207,7 +173,6 @@ impl Serialize for ModuleInfoMap {
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ModuleInfoMapItem {
-  #[serde(serialize_with = "serialize_specifier_vec")]
   pub deps: Vec<ModuleSpecifier>,
   pub size: usize,
 }
--- a/cli/info.rs
+++ b/cli/info.rs
@@ -1,404 +1,232 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
 use crate::colors;
-use crate::global_state::GlobalState;
-use crate::module_graph::{ModuleGraph, ModuleGraphFile, ModuleGraphLoader};
+use crate::MediaType;
 use crate::ModuleSpecifier;
-use crate::Permissions;
-use deno_core::error::AnyError;
-use serde::ser::Serializer;
+
+use serde::ser::SerializeSeq;
 use serde::Serialize;
-use std::collections::{BTreeMap, HashMap, HashSet};
-use std::sync::Arc;
+use serde::Serializer;
+use std::cmp::Ordering;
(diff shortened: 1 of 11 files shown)


cli/module_graph2.rs
```diff
@@ -421,6 +439,128 @@ impl Graph2 {
...
+    let module = self.modules.get(specifier).unwrap();
+    let mut deps = Vec::new();
+    let mut total_size = None;
+
+    if not_seen {
+      let mut seen_deps = HashSet::new();
+      // Check if the module has a type only dependency itself (e.g.
+      // `X-TypeScript-Types` header or triple slash ref)
+      if let Some((_, dep)) = &module.maybe_types {
+        if seen_deps.insert(dep.clone()) {
```
lucacasonato, line 458: I feel like type only dependencies should be separated in the JSON (type_deps field).

cli/module_graph2.rs
```diff
@@ -421,6 +439,128 @@ impl Graph2 {
...
+      for (_, dep) in module.dependencies.iter() {
+        // Check the runtime code dependency
+        if let Some(code_dep) = &dep.maybe_code {
+          if seen_deps.insert(code_dep.clone()) {
+            deps.push(self.get_info(code_dep, seen, totals));
+          }
+        }
+        // Check the type only dependency (e.g. `@deno-types` pragma)
+        if let Some(type_dep) = &dep.maybe_type {
+          if seen_deps.insert(type_dep.clone()) {
```
lucacasonato, line 471: Ditto

cli/tests/info_recursive_imports_test.out
```diff
@@ -1,11 +1,12 @@
 local: [WILDCARD]info_recursive_imports_test.ts
 type: TypeScript
 deps: 4 unique (total [WILDCARD])
+
 file://[WILDCARD]cli/tests/info_recursive_imports_test.ts ([WILDCARD])
 └─┬ file://[WILDCARD]cli/tests/recursive_imports/A.ts ([WILDCARD])
   ├─┬ file://[WILDCARD]cli/tests/recursive_imports/B.ts ([WILDCARD])
   │ ├─┬ file://[WILDCARD]cli/tests/recursive_imports/C.ts ([WILDCARD])
   │ │ ├── file://[WILDCARD]cli/tests/recursive_imports/A.ts *
-  │ │ └── file://[WILDCARD]cli/tests/recursive_imports/common.ts ([WILDCARD])
```
lucacasonato, line 9: I am not fond of this change. Already displayed dependencies should be marked with *.

lucacasonato (commented):
Very nice. Have some minor comments.

cli/tests/info_recursive_imports_test.out
kitsonk: It does, I was just finding that with the new module graph and circular dependencies, it wasn't deterministic which one got "seen" first.  It was a crap shoot.  I can try to figure out how to make it deterministic.

cli/module_graph2.rs
kitsonk: If we are open to changing the JSON, I would like to rename quite a few fields actually, more aligned to the internals and consistent naming of things, as I think it really impedes maintainability for the long term.

cli/info.rs
```diff
@@ -1,404 +1,232 @@
...
+use std::fmt;
+use std::path::PathBuf;
+
+/// Serialize a `MediaType` enum into a human readable string.  The default
+/// serialization for media types is and integer.
+fn serialize_media_type<S>(mt: &MediaType, s: S) -> Result<S::Ok, S::Error>
+where
+  S: Serializer,
+{
+  s.serialize_str(&format!("{}", mt))
```
ry, line 22: Presumably we want to do this everywhere we serialize MediaType. Maybe `impl Serialize for MediaType` where it's defined?

cli/info.rs
```diff
@@ -1,404 +1,232 @@
...
+        compiled.to_string_lossy()
+      )?;
     }
     if let Some(ref map) = self.map {
-      f.write_fmt(format_args!("{} {}\n", colors::bold("map:"), map))?;
+      writeln!(f, "{} {}", colors::bold("map:"), map.to_string_lossy())?;
     }
-
-    f.write_fmt(format_args!(
-      "{} {} unique {}\n",
```
ry, line 125: It's odd to me that this module graph refactor is effecting the display. Are these changes unrelated edits?

cli/info.rs
```diff
@@ -1,404 +1,232 @@
...
-  name: String,
-  size: usize,
-  total_size: Option<usize>,
-  deps: Vec<FileInfoDepTree>,
+pub struct ModuleInfo {
+  pub deps: Vec<ModuleInfo>,
+  #[serde(serialize_with = "serialize_specifier")]
+  pub name: ModuleSpecifier,
+  pub size: usize,
+  pub total_size: Option<usize>,
```
ry, line 124: I'd try to avoid superfluous renames/reordering of structs where possible to keep the diff small. I can't tell if this is related to the module graph or is just a better name/ordering.

cli/info.rs
```diff
@@ -427,105 +252,89 @@ mod test {
...
-
-    let mut graph = ModuleGraph::new();
-
-    graph.insert(aspec.to_string(), a);
-    graph.insert(bspec.to_string(), b);
-    graph.insert(cspec.to_string(), c);
-
-    let count = get_unique_dep_count(&graph);
-
-    assert_eq!(graph.len() - 1, count);
```
ry, line 529: There are three tests here that got removed get_resolved_file_test, dependency_count_no_redirects, dependency_count_with_redirects. Why?

cli/info.rs
kitsonk: No, because we have a serialize of `MediaType` that goes to the equivalent of the enum in JavaScript for `tsc`.  Though in the new infrastructure we don't serailize that at all to send into JavaScript.

I could refactor it so this is the default serialization and we modify the `module_graph` (legacy) to use a custom serializer, but I would just rather address it once we have completed the migration.

cli/info.rs
kitsonk: I could choose to not refactor the display to more idiomatic Rust, but that seems non-sensical.  I really don't want to step around unmaintainable code when there is an opportunity to refactor it.

cli/info.rs
kitsonk: Again, it seems totally non-sensical not to clean up stuff... otherwise it never gets cleaned up.

cli/info.rs
kitsonk: I will take a look at those again, but they are testing behaviours that are not part of the new module graph, because of the way the graph is structure, but I will double check that the logic is covered.

cli/info.rs
ry: Make its it easier to understand during complex refactors if only the necessary bits are touched - but no big deal

cli/info.rs
```diff
@@ -1,404 +1,232 @@
...
+    writeln!(
+      f,
+      "{} {}",
+      self.info.name,
+      colors::gray(&format!("({})", human_size(self.info.size as f64)))
+    )?;
+
+    let dep_count = self.info.deps.len();
+    for (idx, dep) in self.info.deps.iter().enumerate() {
+      dep.write_info(f, &"", idx == dep_count - 1)?;
```
ry, line 108: replace `&""` with `""`

cli/info.rs
```diff
@@ -1,404 +1,232 @@
...
 #[serde(rename_all = "camelCase")]
-struct FileInfoDepTree {
-  name: String,
-  size: usize,
-  total_size: Option<usize>,
-  deps: Vec<FileInfoDepTree>,
+pub struct ModuleInfo {
+  pub deps: Vec<ModuleInfo>,
+  #[serde(serialize_with = "serialize_specifier")]
+  pub name: ModuleSpecifier,
```
ry, line 122: https://github.com/denoland/deno/pull/7900

cli/module_graph2.rs
kitsonk: I guess changing the behaviour is outside of the scope of this PR.  I opened #7927 which we can deal with post this PR.

cli/info.rs
kitsonk: Ok, I have double checked.  `ModuleGraph2` handles re-directs in a significantly different way than the old module graph.  It depends upon the `SpecifierHandler` to mediate the redirects (which is the right layer for these to be abstracted in).  This means that `ModuleGraph2` is _unaware_ of what is going on in the background, which is perfectly fine.  I went through some older issues we had when moving to the "old" `ModuleGraph` having to deal with redirects and retested with `deno run --no-check` which uses the new `ModuleGraph2` and everything is handled properly.

There are quite a few unit tests for `ModuleGraph2`, having to deal with exotic sorts of dependencies, which is the right way to test such logic, and now `info.rs` is simply a "printout" of what is in the graph, versus having to interpret it.  That means these sorts of tests are no longer applicable.

cli/tests/info_recursive_imports_test.out
kitsonk: I tried, but it seems that the recursion and the new `ModuleGraph2` don't like to be deterministic.  One of the three versions of `common.ts` gets marked with its size, and the other 2 get marked with the ` *`, it just seems like the luck of the draw.

Let me know how much time you want me to try to figure it out, because the information is accurate, it just isn't deterministic.

ry (approved):
LGTM

Issue #7849, opened 2020-10-07
Should evaluation errors in the read-eval-print-loop go to stdout?


caspervonb:
This struck me as a bit odd when re-writing the read-eval-print-loop (REPL), we print `throw 'foo'` to stderr.

Node does print everything to stdout which is what I would expect, my rationale being that it is a successful evaluation where the result happens to be an error therefore not an error in the sense that it is a program error that should go to stderr.

Already been sitting on a patch for this for a couple of days trying but I'm curious to see what people's expectations are and what do other REPLs do (Python, Ruby, Lua, et cetera).

solson:
I agree with your rationale. I would reserve `stderr` for things like Rust panics (it's where they go by default) and other implementation errors in the REPL itself, and otherwise expect all output to go to `stdout`.

I took a look at the mentioned languages and was slightly surprised to find this:

| Language | REPL errors written to |
| - | - |
| Ruby | stdout |
| Python | stderr |
| Lua | stderr |

Given that I never noticed any difference between these, it probably doesn't matter too much whether we use `stdout` or `stderr`.

PR #7893, opened 2020-10-09
fix(cli/repl): write all results to stdout

caspervonb:
This writes all evaluaton results to stdout regardless if the result is an error or not.

This matches the behavior of other read-eval-print-loops like Node.

Resolves https://github.com/denoland/deno/issues/7849


bartlomieju (approved):
LGTM

PR #7895, opened 2020-10-09
refactor: Worker is not a Future

bartlomieju:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Install rust
         uses: hecrj/setup-rust-action@v1
         with:
-          rust-version: 1.46.0
+          rust-version: 1.47.0
 
       - name: Install clippy and rustfmt
         if: matrix.kind == 'lint'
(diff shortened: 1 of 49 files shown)


caspervonb:
Welcome change from my perspective, nice to be explicit about this but a few open questions.

1. Event loop implies inspector loop? Maybe this is a good time to decouple and be explicit about that.
2. Creating a session from the worker, seems like this should hang off the inspector instead? The panic seems very awkward.

bartlomieju:
> Welcome change from my perspective, nice to be explicit about this but a few open questions.
> 
> 1. Event loop implies inspector loop? Maybe this is a good time to decouple and be explicit about that.

How do you propose to do that?

> 2. Creating a session from the worker, seems like this should hang off the inspector instead? The panic seems very awkward.

It would mean that `Worker::inspector` would be public which I explicitly don't want to do and that's why there's a method

ry:
What happened to d8879feb8c832dbb38649551b1cb0730874f7be6?

ry (approved):
LGTM

PR #7896, opened 2020-10-09
fix(cli/repl): unterminated string literal should invalidate

caspervonb:
This adds the grave character to the pair matching so that template string literals trigger multi-line edits.


bartlomieju (commented):
@caspervonb could we get some minimal integration test for this feature?

bartlomieju (approved):

Issue #464, opened 2018-08-07
Port tools/*.py to python 3


ry:
**Must** continue run on python 2.7

mreinstein:
would it be better to wait until v8's build process uses py 3? This just seems like a minefield.

hayd:
For now an alternative, see #475, is to have python 3 error out immediately and explicitly (rather than implicitly with a SyntaxError, either in deno or later).

Going forward, would the idea be to fork/port the third_party scripts? That's a lot of code so this may be non-trivial, but we can give it a try (e.g. see how close [futurize](http://python-future.org/) is from "just work"ing).

Note: In my first run in python 2 [Luci](https://github.com/luci/luci-go) errored because it was unable to parse the version of my default python (anaconda). So that may also have to be addressed/fixed. See also [issue/below it](https://github.com/denoland/deno/issues/118#issuecomment-394293252). Probably ought to be separate issue.

mreinstein:
2.7 is a hard requirement for building v8. It just seems like a headache to have to maintain 2 versions of python just so we can say we are on 3.

> would the idea be to fork/port the third_party scripts?

That would be super painful to have to do. :(

Has anyone gauged interest within the chrome team in updating v8's build tooling py 2.7 -> 3? Assuming they are open to this that would be much nicer in that everyone shares the benefit and we don't have to maintain our own build tooling with deno's very limited resources.

ry:
I agree with @mreinstein - forking/porting the third_party scripts is well beyond our capacity.
py3 will not be supported.  When I created this issue, I mistakenly thought that the python used by the build system was included as a binary in third_party, and that we could get away with just porting our frontend scripts.

I think @hayd's solution is good - but it actually introduces some complexity for a friendly error message. If people don't understand py2 vs py3 error messages, they shouldn't be building deno in the first place. Therefore we will leave it as is.

Thanks @haoxli @hayd for the effort.

hayd:
xlink: https://github.com/nodejs/node/issues/24512

Will be interesting to see if efforts to port chromium/v8 scripts to python 3 gain any traction now EOL is < 1 year away...

ccouzens:
It will become difficult to have the default python be 2.7 in the future.

I worked around this by creating a file `bin/python`.

```bash
#!/usr/bin/env bash

exec python2.7 "$@"
```

Making it executable `chmod +x bin/python`.

And sticking it in the path when running the tests: `PATH="$(pwd)/bin:$PATH" cargo test`.

PR #7897, opened 2020-10-09
docs(core): concepts around runtime, ops and resources

Bromeon:
A while ago I started using `deno_core` and struggled with the lack of documentation. Fortunately I got excellent explanations from the developers on Discord, so I thought it might be a good time to contribute back.

I focused the documentation on the parts that I considered important to get a basic example in `deno_core` working. So there are still more APIs to be documented, but at this time I don't have enough knowledge to do that.

There are probably quite a few concepts that can be worded/explained better, so feel free to give feedback!

-----

By the way, I had some trouble running the Python scripts `format.py` and `lint.py` on Windows.
<details>
  <summary>Python</summary>
 
First issue:
```
Traceback (most recent call last):
  File "format.py", line 89, in <module>
    sys.exit(main())
  File "format.py", line 51, in main
    dprint()
  File "format.py", line 59, in dprint
    run(command, shell=False, quiet=True)
  File "C:\Rust\deno\tools\util.py", line 62, in run
    rc = subprocess.call(args, cwd=cwd, env=env, shell=shell)
  File "C:\...\Python38\lib\subprocess.py", line 340, in call
    with Popen(*popenargs, **kwargs) as p:
  File "C:\...\Python38\lib\subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\...\Python38\lib\subprocess.py", line 1307, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the specified file.
```
It turned out that the `dprint` command failed, which I found out only by inserting manual debug statements. I realized that there was a Git submodule to be cloned, so I did that.

After getting the 2nd error, I gave up:
```
Traceback (most recent call last):
  File "format.py", line 89, in <module>
    sys.exit(main())
  File "format.py", line 52, in main
    yapf()
  File "format.py", line 64, in yapf
    source_files = get_sources(root_path, ["*.py"])
  File "format.py", line 31, in get_sources
    return getter(*args)
  File "C:\Rust\deno\tools\util.py", line 186, in git_ls_files
    os.path.normpath(os.path.join(base_dir, f)) for f in output.split("\0")
TypeError: a bytes-like object is required, not 'str'
```
Something with binary/text encoding or wrong Python version.
</details>

I'm aware the Python scripts are not the focus of this project. However, since it's expected that every Pull Requester run them, would it make sense to improve the error messages a bit?

### DIFF
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -31,6 +31,7 @@ pub enum Op {
   NotFound,
 }
 
+/// Maintains the resources and ops inside a JS runtime.
 pub struct OpState {
   pub resource_table: crate::ResourceTable,
   pub op_table: OpTable,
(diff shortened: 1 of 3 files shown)


bartlomieju:
@Bromeon thanks for the PR!

> By the way, I had some trouble running the Python scripts format.py and lint.py on Windows.

Make sure to checkout all git submodules and run using Python 2.

Bromeon:
Regarding Python, I tried again with Python 2, just to get other errors (file names too long).

There are of course configurations and workarounds for this, but many Windows users will run into those. As far as I understood, several of these problems have been mitigated by Python 3. Since Python 2 has passed its official end-of-life, would it make sense to translate the scripts to Python 3 at some point?

lucacasonato:
@Bromeon That is out of scope - we have to wait until the upstream V8 build scripts are migrated to py3. See https://github.com/denoland/deno/issues/464

Bromeon:
Rebased on latest master. 
For some reason, GitHub auto-closed the PR when the source repo was up-to-date (0 diff) with deno/master.

kitsonk (approved):
LGTM

PR #7898, opened 2020-10-09
Fix CI two times 

piscisaureus:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,35 @@ jobs:
           restore-keys: |
             deno-${{ matrix.os }}-${{ matrix.kind }}-
 
+        # It seems that the 'target' directory does not always get restored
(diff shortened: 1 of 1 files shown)


ry (approved):
I did an analysis of the speed up we're getting from cache in https://github.com/denoland/deno/pull/7903#issuecomment-706252380 It's not much. I would say if we run into more bugs or have to introduce more complexity, it would be probably better to simply remove the whole thing.

But it seems like this is working and is faster so LGTM

piscisaureus:
> I would say if we run into more bugs or have to introduce more complexity, it would be probably better to simply remove the whole thing.

It's a lot of stuff for a few minutes gain. 

Maybe this experience will give us some hints on why our builds are so slow (and caching not effective) and it can be improved.

If not I agree then let's just turn it off. 

BTW from-scratch release build & run tests in 10 minutes!
I need an XL at home. It usually takes at least 30 minutes for me...

Issue #6394, opened 2020-06-20
TOML Parser cannot parse inline arrays of inline tables


mythmon:
In the [TOML spec](https://github.com/toml-lang/toml), there is an example of an inline-defined array of inline-defined tables at the end of the "Array of Tables" section:

> You may also use inline tables where appropriate:
> 
> ```toml
> points = [ { x = 1, y = 2, z = 3 },
>            { x = 7, y = 8, z = 9 },
>            { x = 2, y = 4, z = 8 } ]
> ```

However this example cannot be parsed with the TOML parser in the stdlib:

```typescript
import { parse } from "https://deno.land/std/encoding/toml.ts"

console.log(parse(`
points = [ { x = 1, y = 2, z = 3 },
  { x = 7, y = 8, z = 9 },
  { x = 2, y = 4, z = 8 } ]
`))
```

```shell
$ deno run test.ts 
Compile file:///workspaces/flags/test.ts
error: Uncaught SyntaxError: Unexpected token '='
    at Parser._parseData (https://deno.land/std/encoding/toml.ts:261:17)
    at Parser._processDeclaration (https://deno.land/std/encoding/toml.ts:190:24)
    at Parser._parseLines (https://deno.land/std/encoding/toml.ts:340:25)
    at Parser.parse (https://deno.land/std/encoding/toml.ts:389:10)
    at parse (https://deno.land/std/encoding/toml.ts:569:33)
    at file:///workspaces/flags/test.ts:3:13
```

This example works on other TOML parsers, such as [this online one](https://toml-to-json.matiaskorhonen.fi/).

I am using Deno 1.1.1.

wongjiahau:
Hi @mythmon , the current implementation of Deno std TOML parser is actually quite incomplete, so you can expect to hit more errors especially with recursive values, such as `a=[{x=[{y=2}]}]`.  That being said, a major revamp is probably unavoidable in order to fix this issue. 

However if you are looking for a workaround, you can import this [TOML parser](https://github.com/BinaryMuse/toml-node#readme) from `jsmp.io`:
```ts
import toml from "https://dev.jspm.io/toml"
console.log(toml.parse(`
points = [ { x = 1, y = 2, z = 3 },
           { x = 7, y = 8, z = 9 },
           { x = 2, y = 4, z = 8 } ]
  `))
```

rfaulhaber:
I'd be interested in taking a stab at this issue!

wongjiahau:
[...]

PR #7902, opened 2020-10-09
fix(std): Parsing inline arrays of inline tables in toml

kodifiera:
Closes [#6394](https://github.com/denoland/deno/issues/6394)


bartlomieju (approved):
Nice fix, thanks @kodifiera

Issue #7815, opened 2020-10-04
js_unit_tests startTls fails with "Network is unreachable (os error 101)"


trivikr:
**Describe the bug**
Unit test startTls fails with "Network is unreachable (os error 101)"

**Deno version**
master

**To Reproduce**
Follow steps mentioned in https://deno.land/manual/contributing/development_tools#tests to run tests:
```console
cargo test js_unit_tests
```

Unit tests fail with the following error: 

<details>
<summary>Error: Network is unreachable (os error 101)</summary>

```console
startTls
Error: Network is unreachable (os error 101)
    at Object.jsonOpAsync (core.js:236:13)
    at async Object.connect (deno:cli/rt/30_net.js:224:13)
    at async startTls (file:///home/trivikr/workspace/deno/cli/tests/unit/tls_test.ts:192:16)
    at async asyncOpSanitizer (deno:cli/rt/40_testing.js:34:7)
    at async Object.resourceSanitizer [as fn] (deno:cli/rt/40_testing.js:68:7)
    at async TestRunner.[Symbol.asyncIterator] (deno:cli/rt/40_testing.js:240:13)
    at async runTests (deno:cli/rt/40_testing.js:317:22)
    at async workerRunnerMain (file:///home/trivikr/workspace/deno/cli/tests/unit/unit_test_runner.ts:75:3)
```

</details>

<details>
<summary>Details with RUST_BACKTRACE=1</summary>

```console
---- js_unit_tests stdout ----
test_server starting...
target_dir /home/trivikr/workspace/deno/target/debug
target_dir /home/trivikr/workspace/deno/target/debug
thread 'js_unit_tests' panicked at 'assertion failed: `(left == right)`
  left: `Some(0)`,
 right: `Some(1)`', cli/tests/integration_tests.rs:699:3
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
[...]

PR #7904, opened 2020-10-09
fix(std/test): ignore startTls test

trivikr:
Fixes: https://github.com/denoland/deno/issues/7815

Creating a draft PR as a better solution could be to ignore the test in specific case, i.e., if ISP has blocked `smtp.gmail.com:587`

### DIFF
--- a/cli/tests/unit/tls_test.ts
+++ b/cli/tests/unit/tls_test.ts
@@ -183,7 +183,7 @@ unitTest(
 );
 
 unitTest(
-  { perms: { read: true, net: true } },
+  { ignore: true, perms: { read: true, net: true } },
   async function startTls(): Promise<void> {
     const hostname = "smtp.gmail.com";
     const port = 587;


nayeemrmn:
If the test passes in CI and most other developers, and doesn't affect any other test, there's no reason to disable it. (This test always fails for me, too.)

trivikr:
> If the test passes in CI and most other developers, and doesn't affect any other test, there's no reason to disable it. (This test always fails for me, too.)

The reason is point 4 in PR template https://github.com/denoland/deno/blob/f4357f0ff9d39411f22504fcc20db6bd5dec6ddb/.github/PULL_REQUEST_TEMPLATE.md#L19

Is there a way to ignore the test only when `smtp.gmail.com:587` is blocked by ISP?

kitsonk:
We should find a way to test this without needing to connect outside of the environment by creating a mock TLS server on the localhost, like we have for other tests.

bnoordhuis:
@kitsonk I don't necessarily disagree but the test fails through no fault of the person running it. Marking it flaky seems okay to me as a stop-gap measure.

A propos a mock TLS server: that means it'd just be testing `Deno.startTls()` on both ends, unless it's written in e.g. python.

lucacasonato:
Instead of disabling, couldn't we change the test to do HTTP over TLS with our TLS http test server? We could also just add an echoing TLS test server in our test_util that uses our self signed TLS certs.

bartlomieju:
> Instead of disabling, couldn't we change the test to do HTTP over TLS with our TLS http test server? We could also just add an echoing TLS test server in our test_util that uses our self signed TLS certs.

This PR completely disables this test, even though it's passing on CI. That seems like a bad idea to me. I agree with @lucacasonato we should use self-signed cert and test server instead of hitting `smpt.gmail.com`. Closing without merge.

PR #7908, opened 2020-10-09
ModuleGraph2 cleanups

ry:


kitsonk (approved):
Yeah, this is just going to have to come back in on the PR I will likely raise later today. I get the point about not landing dark code, but it is going to cause the rebasing hell you said we wanted to avoid. But 🤷‍♂️.

bartlomieju:
> Yeah, this is just going to have to come back in on the PR I will likely raise later today

Then I believe we can skip merging this PR if the code is about to come back a few hours later.

PR #7914, opened 2020-10-10
refactor(cli): rename GlobalState to ProgramState

bartlomieju:


ry (commented):
I wonder if we can't find a different name than CliState? I'm worried because we used to have a different `CliState` and even tho it's gone now, I think it would be less confusing if there was a different name entirely...

kitsonk:
I am 👍  for `ProgramState` (I also agree on wanting to avoid `CliState`)

kitsonk (approved):
LGTM

ry (approved):

PR #7920, opened 2020-10-10
feat(cli/main): Add deno cache --test and --worker

### LINKED ISSUE #6753: deno cache should allow self to cache worker scripts
As the title says, I expect `deno cache` to work on all files deno understands. It shouldn't require a special flag.

I should be able to do this in a docker build:
```Dockerfile
RUN deno cache main.ts host.ts my.worker.ts 
```

### LINKED ISSUE #7775: Add deno test --no-run
- `deno cache foo.ts bar.ts --test` -> `deno cache foo.ts bar.ts <...files captured by "deno test">`
- `deno cache foo.ts bar.ts --test=baz/,qux/` -> `deno cache foo.ts bar.ts <...files captured by "deno test bar/ qux/">`

Alternatively `deno test --no-run` (`cargo` has both).

EDIT: Went with the second option `deno test --no-run` to avoid the awkward flag value for `--test=dir1,dir2`.

nayeemrmn:
Closes #7775.
Closes #6753.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -3,6 +3,7 @@
 use clap::App;
 use clap::AppSettings;
 use clap::Arg;
+use clap::ArgGroup;
 use clap::ArgMatches;
 use clap::ArgSettings;
 use clap::SubCommand;
@@ -37,6 +38,8 @@ pub enum DenoSubcommand {
   },
   Cache {
     files: Vec<String>,
+    worker_specifiers: Option<Vec<String>>,
+    test_patterns: Option<Vec<String>>,
   },
   Fmt {
     check: bool,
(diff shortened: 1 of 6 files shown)


bartlomieju:
@nayeemrmn could you address https://github.com/denoland/deno/issues/6753 as part of this PR as well?

lucacasonato (approved):
LGTM

During testing I discovered that our emit hashing is not working correctly. We only seem to hash the `tsconfig` passed in through flags, rather than the entire tsconfig after it has been merged with our defaults. This leads to `web worker` compile emits and `main worker` compile emits being hashed the same. That leads to this bug:

```ts
// foo.ts
console.log("hello world");
self.onmessage = () => {}
```
```
> deno cache foo.ts
Check file:///mnt/f9/Projects/github.com/denoland/deno/foo.ts
error: TS2339 [ERROR]: Property 'onmessage' does not exist on type 'Window & typeof globalThis'.
self.onmessage = () => {}
     ~~~~~~~~~
    at file:///mnt/f9/Projects/github.com/denoland/deno/foo.ts:2:6

> deno cache --worker=foo.ts
Check file:///mnt/f9/Projects/github.com/denoland/deno/foo.ts

>  deno cache foo.ts
```

The third call to `deno cache` should error, but it doesn't.

cc @kitsonk

bartlomieju:
@kitsonk does this PR conflict with your compiler refactors or can we land it?

kitsonk:
@bartlomieju 🤷  but probably, we should land run/cache refactor first though and then see if there is any impact.  It looks like it sort of goes around the module graph as a whole (which cache might actually be doing too and I didn't notice it).  But we should only be using the new module graph to emit any files.

bartlomieju:
> @bartlomieju 🤷 but probably, we should land run/cache refactor first though and then see if there is any impact. It looks like it sort of goes around the module graph as a whole (which cache might actually be doing too and I didn't notice it). But we should only be using the new module graph to emit any files.

Alright, then let's try after landing run/cache refactor.

bartlomieju (approved):
LGTM, thanks @nayeemrmn

Issue #7906, opened 2020-10-09
Proposal: std/mail , std/smtp


andreespirela:
Hello there.

I think it would be nice to have `mail` or `smtp` or both in `std`. This is very necessary for many applications & really many things are involved in this kind of package including the security aspect so I think it's ideal for Deno to have something official in the `std` packages. 

Current implementations in GO std:
https://golang.org/pkg/net/mail/
https://golang.org/pkg/net/smtp/

lucacasonato:
I wonder if we could implement mail parsing in wasm based on https://docs.rs/mailparse/0.13.0/mailparse/.

kitsonk:
Please note, a label of `suggestion` as this issue is labelled means that a suggestion for a new feature that is yet to be agreed (which is what the description for the label notes).

While it is part of the Go standard library, I personally don't think it we should add it to `std`.

PR #7925, opened 2020-10-10
feat(std): add mail parsing to std/mail

littledivy:
Adds support for email parsing -
- [x] `parseAddr` 
- [x] `parseDate`

Ref #7906 

https://golang.org/pkg/net/mail/


crypt096 (approved):

lucacasonato:
@littledivy Is this ready, or are you going to add bindings for actual mail parsing too?

littledivy:
@lucacasonato Yup it's ready - has parsing functions matching Go's std/mail and features can be added as and when there is a need for it.

bartlomieju:
I'm -1 on this PR - adding another WASM module to std/ just for two utility functions seems like a big overkill to me. I'd be happy to add if it was broader implementation.

lucacasonato:
I think that introducing this module for just the address and date parser is not great. We should at least be able to parse an email into structured data before this goes into std.

littledivy:
What should I do then? I'm happy to add bindings to `parseMail` - https://docs.rs/mailparse/0.13.0/mailparse/fn.parse_mail.html or just close the PR.

bartlomieju:
@littledivy sorry for late response. Thanks for the PR but I'd prefer not to use WASM module for this purpose. I think a TypeScript implementation would be better and I suggest coordinating with @uki00a on https://github.com/denoland/deno/pull/8413 as both functionalities are closely related. Respectfully closing without merge.

PR #7930, opened 2020-10-11
fix(cli/rt/console): fix inspection of Function

### LINKED ISSUE #7633: console.log dies on null prototype function
So, I was honestly acting dumb, and ended up with essentially just this:
```ts
const f = <T>(x: T): T => x;

Reflect.setPrototypeOf( f, null );

console.log( f );

console.log(( f( NaN ) );
```

Attempting to log the function causes an error.

> Uncaught TypeError: Cannot read property 'constructor' of null...

kt3k:
This PR fixes the inspection of functions. The current implementation gets the name of the type of the function from `f.__proto__.constructor.name`, and it throws when the prototype is set to null as described in #7633. This PR checks the prototype before accessing its constructor name and uses the generic name 'Function' if the prototype is not available.

fixes #7633

### DIFF
--- a/cli/tests/unit/console_test.ts
+++ b/cli/tests/unit/console_test.ts
@@ -347,6 +347,21 @@ unitTest(function consoleTestStringifyCircular(): void {
 });
 /* eslint-enable @typescript-eslint/explicit-function-return-type */
 
+unitTest(function consoleTestStringifyFunctionWithPrototypeRemoved(): void {
+  const f = function f() {};
+  Reflect.setPrototypeOf(f, null);
+  assertEquals(stringify(f), "[Function: f]");
+  const af = async function af() {};
+  Reflect.setPrototypeOf(af, null);
+  assertEquals(stringify(af), "[Function: af]");
+  const gf = function gf() {};
+  Reflect.setPrototypeOf(gf, null);
+  assertEquals(stringify(gf), "[Function: gf]");
+  const agf = function agf() {};
+  Reflect.setPrototypeOf(agf, null);
+  assertEquals(stringify(agf), "[Function: agf]");
+});
+
 unitTest(function consoleTestStringifyWithDepth(): void {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thank you @kt3k 

I'll wait for @nayeemrmn approval before landing

nayeemrmn (approved):
LGTM

cli/tests/unit/console_test.ts
```diff
@@ -347,6 +347,21 @@ unitTest(function consoleTestStringifyCircular(): void {
...
+  assertEquals(stringify(f), "[Function: f]");
+  const af = async function af() {};
+  Reflect.setPrototypeOf(af, null);
+  assertEquals(stringify(af), "[Function: af]");
+  const gf = function gf() {};
+  Reflect.setPrototypeOf(gf, null);
+  assertEquals(stringify(gf), "[Function: gf]");
+  const agf = function agf() {};
+  Reflect.setPrototypeOf(agf, null);
+  assertEquals(stringify(agf), "[Function: agf]");
```
solson, line 362: I know this is already merged, but was this intended to test generator functions and async generator functions? (`function* gf() {}` and `async function* agf() {}`)

bartlomieju: @solson good catch, could you open an issue about it?

nayeemrmn: > `function* gf() {}` and `async function* agf() {}`

Are you saying they should have their own output? That's just declaration-side syntactic sugar.

solson: No, I'm saying they should be declared that way here in the test, as in `const gf = function* gf() {};` to match `async function af` above, presumably to test this "set prototype to `null`" manipulation on all 4 of `Function`, `AsyncFunction`, `GeneratorFunction`, and `AsyncGeneratorFunction`. Currently, `gf` and `agf` are just testing the same thing as `f`.

solson: @bartlomieju https://github.com/denoland/deno/issues/7962

PR #7931, opened 2020-10-11
docs(std/datetime): document toIMF, isLeap, difference, and constants

kt3k:
This PR adds the documents of `toIMF`, `isLeap`, `difference`, and constants to `std/datetime/README.md`.


bartlomieju (approved):
LGTM, thanks @kt3k

PR #7933, opened 2020-10-11
refactor(cli/flags): Remove Flags::repl

nayeemrmn:
Use `matches!(flags.subcommand, DenoSubcommand::Repl)` instead.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -120,7 +120,6 @@ pub struct Flags {
   pub no_remote: bool,
   pub read_allowlist: Vec<PathBuf>,
   pub reload: bool,
-  pub repl: bool,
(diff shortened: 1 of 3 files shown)


caspervonb:
Added this a couple days ago with the specific intent to allow other commands to run with a console attached.

nayeemrmn:
Hmm, okay. But `flags.repl` doesn't communicate any difference from `flags.subcommand == DenoSubcommand::Repl`. Shouldn't be in flags and needs a different name.

caspervonb:
> Shouldn't be in flags and needs a different name.

Should be in flags IMO, but yes I agree it does need a different name eventually.

PR #7606, opened 2020-09-21
fix(websocket): add missing close events and remove extra error event

crowlKats:
Fixes #7604

PR #7936, opened 2020-10-11
fix(cli/rt/websockets): Only add Sec-WebSocket-Protocol if it's not empty

### LINKED ISSUE #7935: WebSocket adds empty protocol header
Even without specifying a protocol in the `WebSocket` constructor, a `Sec-WebSocket-Protocol` header is added with an empty value.

Some WebSocket servers reject this empty header and fail the connection.

Edit:

As I understand, the WebSocket specification says, the value of `Sec-WebSocket-Protocol`, if present, has to be a non empty string:

> 10.  The request MAY include a header field with the name
        |Sec-WebSocket-Protocol|.  If present, this value indicates one
        or more comma-separated subprotocol the client wishes to speak,
        ordered by preference.  The elements that comprise this value
        MUST be **non-empty strings** with characters in the range U+0021 to
        U+007E not including separator characters as defined in
        [RFC2616] and MUST all be unique strings.  The ABNF for the
        value of this header field is 1#token, where the definitions of
        constructs and rules are as given in [RFC2616].

See [4.1 Client Requirements](https://tools.ietf.org/html/rfc6455#section-4.1).

Nerixyz:
This PR fixes #7935 . The `WebSocket` will only add the header if it's not empty.

### DIFF
--- a/cli/ops/websocket.rs
+++ b/cli/ops/websocket.rs
@@ -68,11 +68,15 @@ pub async fn op_ws_create(
     cli_state.flags.ca_file.clone()
   };
   let uri: Uri = args.url.parse()?;
-  let request = Request::builder()
+  let mut request = Request::builder()
     .method(Method::GET)
-    .uri(&uri)
-    .header("Sec-WebSocket-Protocol", args.protocols)
-    .body(())?;
+    .uri(&uri);
+
+  if !args.protocols.is_empty() {
+    request = request.header("Sec-WebSocket-Protocol", args.protocols);
+  }
+  
+  let request = request.body(())?;
   let domain = &uri.host().unwrap().to_string();
   let port = &uri.port_u16().unwrap_or(match uri.scheme_str() {
     Some("wss") => 443,
--- a/cli/ops/websocket.rs
+++ b/cli/ops/websocket.rs
@@ -68,14 +68,12 @@ pub async fn op_ws_create(
     cli_state.flags.ca_file.clone()
   };
   let uri: Uri = args.url.parse()?;
-  let mut request = Request::builder()
-    .method(Method::GET)
-    .uri(&uri);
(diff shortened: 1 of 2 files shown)


bartlomieju (changes requested):
Thanks @Nerixyz, can you add a test case?

@crowlKats please take a look

crowlKats:
Just noticed another small mistake with protocols: protocols on the JS side are joined by `; `, but while checking the spec for this PR i noticed that it should be `, `. @Nerixyz could you add this change to this PR as well?
That aside, LGTM.

Nerixyz:
@bartlomieju Sure, I'd like to add tests, but I don't really know where to test this.

bartlomieju:
> @bartlomieju Sure, I'd like to add tests, but I don't really know where to test this.

@Nerixyz `WebSocket` unit tests are in `cli/tests/websocket_test.ts`

Nerixyz:
> `WebSocket` unit tests are in `cli/tests/websocket_test.ts`

@bartlomieju I know WebSocket tests are there, but I don't know a way of testing the protocol header.

bartlomieju:
@crowlKats could you point @Nerixyz where this could be tested?

crowlKats:
@bartlomieju this is the same problem as #7606 where we need a server to test such case, which isnt really the easiest thing to do.

Nerixyz:
For this test we'd need a strict WebSocket server which rejects/closes invalid handshakes. `tungstenite-rs` however doesn't negotiate protocols at all.

Nerixyz:
@bartlomieju Could you merge this pr and then maybe create an issue regarding the testing of the WebSocket implementation.

bartlomieju (approved):
LGTM, not very happy to merge a PR without a tests, but it seems there's no obvious way to test this

Issue #7929, opened 2020-10-11
API request:  toFileURL


callionica:
Deno should provide a function to convert a system file path to a file:// URL. 

Without this, people will just create their own with varying levels of testing and degrees of success.

In particular any function should work for Windows and other systems, and should correctly encode question marks and hash symbols.

martonlederer:
I agree, I had this problem when creating a module last week.

lucacasonato:
> should correctly encode question marks and hash symbols

How should this be done? This seems rather opinionated.

callionica:
If a file name contains a hash or question mark it needs to be encoded otherwise it'll be interpreted as a fragment or query string in the URL. Seems more like a correctness issue than an opinion, but maybe I'm missing something. What're you thinking, Luca?

bartlomieju:
This is the same problem as in caching remote sources in$DENO_DIR/deps - we've never fully figured it out and it's pretty non-trivial functionality to get it working on all systems.

nayeemrmn:
No it's really simple, both `std` and the runtime internals have a `fromFileUrl()` which has uses some decoding scheme, we just need to `toFileUrl()` which uses the equivalent encoding scheme. The scheme should affect characters which are valid in file paths but mean something else in a URL string. Right now we use `{en,de}codeURIComponent()`, which kinda works if we make some exceptions for certain characters.

bartlomieju:
@nayeemrmn ooops, you're right. I thought about converting any URL to a valid filepath. My comment is off-topic then.

callionica:
Did a quick test against https://raw.githubusercontent.com/denoland/deno/master/std/path/mod.ts and seems to be working for question marks and hash marks at least.

nayeemrmn:
I'm interested in cases where `fromFileUrl(toFileUrl(path)) != path` (or the reverse), if anyone finds any.

PR #7937, opened 2020-10-11
fix(op_crates/web/url): Apply backslash replacement to the pathname setter

nayeemrmn:
Moves this step to `encodePathname()` so it affect the setter as well. I'm solidifying this encoder before I look at #7929.


lucacasonato (approved):
LGTM - thanks @nayeemrmn

PR #7939, opened 2020-10-11
fix(lib): to use the new writing

flolep2607:


bartlomieju:
Thanks, but current code is valid.

PR #7934, opened 2020-10-11
feat(std/node): Port most of node errors

Soremwar:
Based on #7569

PR #7940, opened 2020-10-11
feat(std/node/stream): Add Duplex, Transform, Passthrough, pipeline, finished and promises

Soremwar:
Based on #7934

### DIFF
--- a/std/node/_buffer.ts
+++ b/std/node/_buffer.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 import * as hex from "../encoding/hex.ts";
 import * as base64 from "../encoding/base64.ts";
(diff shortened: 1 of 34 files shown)


guybedford (approved):
There's no way this can be reviewed, rather merging and building in feedback from usage seems the only real way to tackle this, so for that reason I'm all for going ahead with a merge.

readable-stream hasn't been updated in months, and relying on a third-party implementation seems to go against the Deno process.

Please do make sure all sources you have copied from have their licenses fully credited, even if there has been significant modification.

Soremwar:
Once #7569 is merged i'll rebase this but it seems it will pass in it's current state.

Soremwar:
This is rebased over latest master

bartlomieju (approved):
LGTM, thanks @Soremwar. It's a huge PR and it's really hard to review it properly, I think we'll just have to iterate on it

PR #7941, opened 2020-10-12
chore(cli): remove dead code

kitsonk:
I noticed that several modules in `cli/main.rs` where `pub` when we don't currently expect anyone to consume the cli crate at the moment.  Removing those public modules resulted in exposing a lot of dead code that was be obfuscated by the public modules.  That code has been removed in this PR.

### DIFF
--- a/cli/media_type.rs
+++ b/cli/media_type.rs
@@ -19,7 +19,6 @@ pub enum MediaType {
   TSX = 4,
   Json = 5,
   Wasm = 6,
-  BuildInfo = 7,
   Unknown = 8,
 }
 
@@ -33,7 +32,6 @@ impl fmt::Display for MediaType {
       MediaType::TSX => "TSX",
       MediaType::Json => "Json",
       MediaType::Wasm => "Wasm",
-      MediaType::BuildInfo => "BuildInfo",
       MediaType::Unknown => "Unknown",
     };
     write!(f, "{}", value)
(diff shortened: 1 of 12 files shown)


cli/media_type.rs
```diff
@@ -75,30 +73,6 @@ impl MediaType {
...
-
-  /// Convert a MediaType to a `ts.Extension`.
-  ///
-  /// *NOTE* This is defined in TypeScript as a string based enum.  Changes to
-  /// that enum in TypeScript should be reflected here.
-  pub fn as_ts_extension(&self) -> &str {
-    match self {
-      MediaType::JavaScript => ".js",
-      MediaType::JSX => ".jsx",
-      MediaType::TypeScript => ".ts",
```
bartlomieju, line 87: Isn't this code used in new module graph/new tsc?

cli/ops/io.rs
```diff
@@ -184,14 +183,12 @@ impl StreamResourceHolder {
 }
 
 pub enum StreamResource {
-  Stdin(tokio::io::Stdin, TTYMetadata),
```
bartlomieju, line 187: This is surprising...

cli/worker.rs
```diff
@@ -194,21 +194,6 @@ impl Worker {
     self.js_runtime.mod_evaluate(id)
   }
 
-  /// Loads, instantiates and executes provided source code
-  /// as module.
-  pub async fn execute_module_from_code(
-    &mut self,
-    module_specifier: &ModuleSpecifier,
-    code: String,
-  ) -> Result<(), AnyError> {
```
bartlomieju, line 203: I'd prefer to leave this method - I'm debugging test runner and there's a good chance this API will be needed.

bartlomieju (approved):
Actually LGTM, let's land it, I can bring code back as needed

ry:
Nice work!

PR #7942, opened 2020-10-12
refactor(cli): add tsc2

kitsonk:
This is the last of the "new" infrastructure that will unblock being able to move everything over.

I will do a subsequent PR based on this to actually move something else over the requires the TS compiler.

### DIFF
--- a/cli/module_graph2.rs
+++ b/cli/module_graph2.rs
@@ -373,8 +373,8 @@ impl Module {
   }
 }
 
-#[derive(Clone, Debug, PartialEq)]
-pub struct Stats(Vec<(String, u128)>);
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Stats(pub Vec<(String, u128)>);
 
 impl<'de> Deserialize<'de> for Stats {
   fn deserialize<D>(deserializer: D) -> result::Result<Self, D::Error>
@@ -572,6 +572,27 @@ impl Graph2 {
     Ok(())
   }
 
+  pub fn get_media_type(
+    &self,
+    specifier: &ModuleSpecifier,
+  ) -> Option<MediaType> {
+    if let Some(module) = self.modules.get(specifier) {
+      Some(module.media_type)
+    } else {
+      None
+    }
+  }
+
+  /// Get the source for a given module specifier.  If the module is not part
+  /// of the graph, the result will be `None`.
+  pub fn get_source(&self, specifier: &ModuleSpecifier) -> Option<String> {
+    if let Some(module) = self.modules.get(specifier) {
+      Some(module.source.clone())
+    } else {
+      None
+    }
+  }
+
   /// Verify the subresource integrity of the graph based upon the optional
   /// lockfile, updating the lockfile with any missing resources.  This will
   /// error if any of the resources do not match their lock status.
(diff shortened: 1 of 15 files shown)


crypt096 (approved):

ry (commented):
Looks good - please proceed to land tsc2.rs when you think it's ready.

I look forward to tsc.rs and module_graph.rs being removed in the near future.

cli/tsc/99_main_compiler.js
```diff
@@ -4,7 +4,7 @@
 // that is created when Deno needs to compile TS/WASM to JS.
 //
 // It provides two functions that should be called by Rust:
-//  - `bootstrapCompilerRuntime`
+//  - `startup`
```
bartlomieju, line 7: Bike-shedding: I liked previous name better... 🤷

cli/tsc/99_main_compiler.js
```diff
@@ -349,37 +360,27 @@ delete Object.prototype.__proto__;
...
-          } else {
-            throw err;
-          }
-        }
+        /** @type {{ data: string; hash: string; }} */
+        const { data, hash } = core.jsonOpSync(
+          "op_load",
+          { specifier },
+        );
+        debug({ data, hash });
```
bartlomieju, line 373: TIOLI - remove printing of source code before landing - I find it unhelpful when in debug log

cli/tsc/99_main_compiler.js
```diff
@@ -1328,18 +1327,73 @@ delete Object.prototype.__proto__;
...
+   * @param {Request} request 
+   */
+  function exec({ config, debug: debugFlag, rootNames }) {
+    setLogDebug(debugFlag, "TS");
+    performanceStart();
+    debug(">>> exec start", { rootNames });
+    debug(config);
+
+    const { options, errors: configFileParsingDiagnostics } = ts
+      .convertCompilerOptionsFromJson(config, "", "tsconfig.json");
```
bartlomieju, line 1347: No problems with `allowNonTsExtensions` setting? I had to explicitly override it in each of legacy handlers like so: `options.allowNonTsExtensions = true;`

cli/tsc2.rs
```diff
@@ -0,0 +1,589 @@
...
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Request {
+  /// The TypeScript compiler options which will be serialized and sent to
+  /// tsc.
+  pub config: TsConfig,
+  /// Indicates to the tsc runtime if debug logging should occur.
+  pub debug: bool,
+  #[serde(skip_serializing)]
+  pub graph: Rc<RefCell<Graph2>>,
```
bartlomieju, line 44: Looks like `graph` is not mutated anyway, so maybe drop the `RefCell`?

bartlomieju:
@kitsonk looks great! Just a few minor nitpicks/questions, looking forward to landing it!

bartlomieju (approved):

cli/tsc/99_main_compiler.js
kitsonk: No problems, because the compiler doesn't resolve any extensions, it never sees an extension... it is fed 100% of the extensions via the `resolveModules()`.  I was wondering why you were struggling with this when trying to adapt stuff from my prototype, but I believe that is the reason.

cli/module_graph2.rs
```diff
@@ -603,6 +624,56 @@ impl Graph2 {
...
   }
 
+  /// Given a string specifier and a referring module specifier, provide the
+  /// resulting module specifier and media type for the module that is part of
+  /// the graph.
+  pub fn resolve(
+    &self,
+    specifier: &str,
+    referrer: &ModuleSpecifier,
+  ) -> Result<ModuleSpecifier, AnyError> {
```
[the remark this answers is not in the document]
kitsonk: Well this method has nothing to do with an import map, so I am not sure a comment here would make sense...

When the graph is built, the import map is resolved, in the builder.  The graph itself is never aware of it, because it is the builder that interacts with the specifier handler to insert modules in the graph, with dependencies full resolved to the other entries in the graph.  Once the graph is built, the import map is discarded.

So I am not sure a comment here means anything.

cli/module_graph2.rs
```diff
@@ -568,18 +568,39 @@ impl Graph2 {
...
+
+  /// Get the source for a given module specifier.  If the module is not part
+  /// of the graph, the result will be `None`.
+  pub fn get_source(&self, specifier: &ModuleSpecifier) -> Option<String> {
+    if let Some(module) = self.modules.get(specifier) {
+      Some(module.source.clone())
+    } else {
+      None
+    }
+  }
```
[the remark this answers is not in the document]
kitsonk: It will be called exactly 1 time per module per execution.

cli/tsc/99_main_compiler.js
kitsonk: well bootstraping isn't really what we are doing here ("a self-starting process that is supposed to proceed without external input"), and I am not sure why we need to indicate that it is the compiler runtime within the compiler runtime, if there were more functions like this, it might make sense to make them less ambiguous.

cli/module_graph2.rs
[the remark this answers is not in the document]
bartlomieju: Okay, thanks for explanation, that makes sense

cli/tsc/99_main_compiler.js
bartlomieju: Great! That felt wrong anyway

cli/tsc2.rs
kitsonk: I am going to land it just as an `Rc<>` but I just ran into a problem when moving `run` over that mutates the graph, so I have to go and put it back in. 🤷

Issue #7907, opened 2020-10-09
Cannot import all the files from a folder.


Ayfri:
Hi!
I'm making a Discord bot with dynamic commands (automatic loading/reloading).
It's not my first Discord bot but all previous was made using node.js and `require()` to do this dynamically.
But here it doesn't work using `await import()` :/

Here is my code : 
```ts
import {SEP} from 'https://deno.land/std/path/mod.ts';
import {configs} from './configs.ts';
// configs.commandsFolder = 'src/commands'

const cmdDir: Iterable<Deno.DirEntry> = Deno.readDirSync(Deno.realPathSync(configs.commandsFolder));
for (let command of [...cmdDir]) {
	if (command.isFile && command.name.endsWith('.ts')) {
		const path: string = Deno.realPathSync(configs.commandsFolder + SEP + command.name);
		const commandClass: any = await import(`file://${path}`);
		const cmd: any = commandClass.default;
		if (!cmd) {
			console.warn(`Could not load '${command.name}' command !`);
			continue;
		}
		commands.set(cmd.name, new cmd());
		console.log(`Successfully loaded '${cmd.name}' command !`);
	}
}

console.log(commands);
```
I have two commands, named `CompileCommand.ts` and `ReloadCommand.ts`, this is the code of the command `reload` :
```ts
import Command from '../classes/Command.ts';

export default class ReloadCommand extends Command {
	public constructor() {
		super('reload');
	}
	
	public async run(message: Message, args: string[]): Promise<void> {
		// some code
	}
}
```
So when I launch all this, it loads the first command but stops at the second for no reason, not sending any error (it was sending one in deno 1.4.4 and now I am in 1.4.5).
There are the logs : 
```ts
Check file:///C:/Users/Pierre/WebstormProjects/deno/tsc-bot/mod.ts
Successfully loaded 'CompileCommand' command !
Check file:///C:/Users/Pierre/WebstormProjects/deno/tsc-bot/src/commands/ReloadCommand.ts
```
(`tsc-bot/mod.ts` is the file where there is the command handler)
I don't know what it is doing, why it does that and I want some help ^^'

kitsonk:
@bartlomieju feels like it is related to #7549 and #7672

bartlomieju:
Fixed in #7911

Ayfri:
It still doesn't work :/
Now I'm getting an error but it makes no sense : 
```ts
[...]

PR #7911, opened 2020-10-10
Fix 100% CPU idling problem by reverting #7672

ry:
fixes #7909
cc @bartlomieju 

Unfortunately I had to revert some unrelated commits to easily revert module execution.

I think we need to cut a new release after landing this.

Please check my work:
```
> cargo run -- run --allow-net https://deno.land/std@0.74.0/examples/echo_server.ts
```
This should have 0% CPU on this branch. On master it has 100%.

Issue #7909, opened 2020-10-09
deno v1.4.5 use 100% of CPU if it start an HTTP server


jersou:
With deno 1.4.5, I have 100% CPU usage with all scripts that use an HTTP server, I don't have these problems with 1.4.4: 
```bash
$ ./deno-1.4.4 --version                                                                                                                                       
      deno 1.4.4
      v8 8.7.75
      typescript 4.0.3
$ ./deno-1.4.5 --version
      deno 1.4.5
      v8 8.7.220.3
      typescript 4.0.3
$ ./deno-1.4.4  run --allow-net https://deno.land/std@0.74.0/examples/echo_server.ts & sleep 10 && ps -p "$(pgrep deno)" -o %cpu,%mem,cmd && pkill deno        
      [1] 43965
      Listening on 0.0.0.0:8080
      %CPU %MEM CMD
       0.0  0.0 ./deno-1.4.4 run --allow-net https://deno.land/std@0.74.0/examples/echo_server.ts
      [1]  + 43965 terminated  ./deno-1.4.4 run --allow-net                                                                                                                               
$ ./deno-1.4.5  run --allow-net https://deno.land/std@0.74.0/examples/echo_server.ts & sleep 10 && ps -p "$(pgrep deno)" -o %cpu,%mem,cmd && pkill deno
      [1] 44519
      Check https://deno.land/std@0.74.0/examples/echo_server.ts
      Listening on 0.0.0.0:8080
      %CPU %MEM CMD
      102  0.1 ./deno-1.4.5 run --allow-net https://deno.land/std@0.74.0/examples/echo_server.ts
      [1]  + 44519 terminated  ./deno-1.4.5 run --allow-net
```
![cpu](https://user-images.githubusercontent.com/5874319/95634354-01639480-0a8a-11eb-8cc7-a1e9ab62550f.png)
My OS : Ubuntu 20.04.1 LTS

jersou:
To be sure, I tried it from docker containers : 

The Dockerfile :
```
# FROM https://github.com/hayd/deno-docker/blob/master/alpine.dockerfile
FROM frolvlad/alpine-glibc:alpine-3.11_glibc-2.31
RUN addgroup -g 1993 -S deno \
        && adduser -u 1993 -S deno -G deno \
        && mkdir /deno-dir/ \
        && chown deno:deno /deno-dir/
ENV DENO_DIR /deno-dir/
ENV DENO_INSTALL_ROOT /usr/local
[...]

PR #7946, opened 2020-10-12
fix: top-level-await module execution

### LINKED ISSUE #7945: Awaiting never-resolved promise makes Deno hang & hot spin.
`deno run` this.

```js
// x.js
await new Promise(() => {});
```

It spins because there is nothing happening any more, and tokio can't block the event loop for us (there's nothing to poll).

Deno should either:
* exit
* throw an exception similar to `UncaughtException`. Maybe `AwaitNeverException` or something?

bartlomieju:
Currently failing, but making this test case pass will ensure our TLA implementation is correct.

Second try for https://github.com/denoland/deno/pull/7672
Reverted in https://github.com/denoland/deno/pull/7911 because of https://github.com/denoland/deno/issues/7909 (which has been addressed in this PR).
Fixes https://github.com/denoland/deno/issues/7907
Fixes https://github.com/denoland/deno/issues/7649
Fixes #7945


core/runtime.rs
```diff
@@ -916,14 +1044,118 @@ impl JsRuntime {
...
-              Err(err) => self.dyn_import_error(dyn_import_id, err)?,
-            };
+            self.dyn_mod_evaluate(dyn_import_id, module_id)?;
+          }
+        }
+      }
+    }
+  }
+
+  fn evaluate_pending_modules(&mut self) -> Result<(), AnyError> {
```
ry, line 1054: Can you add a triple slash comment here describing what this does? What does "pending" mean in this case?

bartlomieju: Done, PTAL

core/runtime.rs
```diff
@@ -83,6 +84,14 @@ pub struct JsRuntime {
   allocations: IsolateAllocations,
 }
 
+type DynImportModEvaluate =
+  (ModuleId, v8::Global<v8::Promise>, v8::Global<v8::Module>);
+type ModEvaluate = (
+  ModuleId,
+  v8::Global<v8::Promise>,
+  mpsc::Sender<Result<(), AnyError>>,
+);
```
ry, line 93: Consider making these structs?

core/runtime.rs
```diff
@@ -91,6 +100,8 @@ pub(crate) struct JsRuntimeState {
   pub(crate) js_recv_cb: Option<v8::Global<v8::Function>>,
   pub(crate) js_macrotask_cb: Option<v8::Global<v8::Function>>,
   pub(crate) pending_promise_exceptions: HashMap<i32, v8::Global<v8::Value>>,
+  pub(crate) pending_dyn_mod_evaluate: HashMap<i32, DynImportModEvaluate>,
+  pub(crate) pending_mod_evaluate: Option<ModEvaluate>,
```
ry, line 104: I realize all of the elements in this struct are `pub(crate)`, but I think the elements you're adding are not accessed outside of `runtime.rs`. Consider making them private.

core/runtime.rs
```diff
@@ -91,6 +100,8 @@ pub(crate) struct JsRuntimeState {
   pub(crate) js_recv_cb: Option<v8::Global<v8::Function>>,
   pub(crate) js_macrotask_cb: Option<v8::Global<v8::Function>>,
   pub(crate) pending_promise_exceptions: HashMap<i32, v8::Global<v8::Value>>,
+  pub(crate) pending_dyn_mod_evaluate: HashMap<i32, DynImportModEvaluate>,
```
ry, line 103: Use `ModuleId` instead of i32?

core/runtime.rs
```diff
@@ -486,13 +504,17 @@ impl JsRuntime {
...
+
+    let has_pending_dyn_imports = !{
+      state.preparing_dyn_imports.is_empty()
+        && state.pending_dyn_mod_evaluate.is_empty()
     };
 
-    if is_idle {
+    let has_pending_module = state.pending_mod_evaluate.is_some();
+
+    if is_idle && !has_pending_dyn_imports && !has_pending_module {
```
ry, line 517: Maybe add ` !has_pending_dyn_imports && !has_pending_module` to the `is_idle` calculation? This bit of code is confusing and I don't understand why it's broken up into multiple statements.

bartlomieju: Refactored and added comment

bartlomieju:
I think I hit an edge case where trying https://github.com/denoland/deno/issues/7907 on this branch - there's most likely an interval being set up by one of submodules which halts execution of one of dynamic imports. I need to think a bit more how that should be handled.

ry (approved):
LGTM - I also confirmed it doesn't have a hot loop when idling

bartlomieju:
So I've managed to find one more bug in implementation which is directly related to #7945 

```javascript
// a.mjs
import { foo } from "./timeout_loop.mjs";

export const collection = [];

const mod = await import("./b.mjs");

console.error("foo in main", foo);
console.error("mod", mod);
```

```javascript
// b.mjs
import { foo } from "./timeout_loop.mjs";
import { collection } from "./a.mjs";

console.log("collection in b", collection);
console.log("foo in b", foo);

export const a = "a";
```

```javascript
// timeout_loop.js
export const foo = "foo";

export function delay(ms) {
    return new Promise((res) =>
        setTimeout(() => { res(); }, ms)
    );
}

let i = 0;

async function timeoutLoop() {
    await delay(1000);
    console.log("timeout loop", i);
    i++;
    if (i > 5) {
        return;
    }
    timeoutLoop();
}

timeoutLoop();
```

```
$ target/debug/deno run --allow-read ./a.mjs
timeout loop 0
timeout loop 1
timeout loop 2
timeout loop 3
timeout loop 4
timeout loop 5
... infinite hot loop in deno_core::JsRuntime ...
```

Node exits with code 13
```
$ node ./a.mjs
timeout loop 0
timeout loop 1
timeout loop 2
timeout loop 3
timeout loop 4
timeout loop 5

$ echo $?
13
```

core/runtime.rs
```diff
@@ -502,6 +538,25 @@ impl JsRuntime {
...
+      }
+    }
+
+    if has_pending_dyn_module_evaluation {
+      if has_pending_ops || has_pending_dyn_imports {
+        state.waker.wake();
+      } else {
+        return Poll::Ready(Err(generic_error("Dynamically imported module evaluation is still pending but there are no pending ops.")));
+      }
+    }
```
bartlomieju, line 558: This bit fixes #7945

core/runtime.rs
```diff
@@ -694,7 +749,100 @@ impl JsRuntime {
...
+      let status = module.get_status();
+
+      if let Some(value) = maybe_value {
+        assert!(
+          status == v8::ModuleStatus::Evaluated
+            || status == v8::ModuleStatus::Errored
+        );
+        let promise = v8::Local::<v8::Promise>::try_from(value)
+          .expect("Expected to get promise as module evaluation result");
+        let promise_id = promise.get_identity_hash();
```
bnoordhuis, line 810: This might be problematic. The hash is only a 22 bits random number so, per the Birthday Paradox, there's a 50% chance of a collision after ~2400 turns.

It's a bit (hah!) worse on 32 bits architectures because the hash is one bit smaller, it reaches 50% after ~1700 turns.

bnoordhuis (approved):

core/runtime.rs
```diff
@@ -474,25 +499,36 @@ impl JsRuntime {
...
+    self.evaluate_pending_module()?;
     {
-      let overflow_response = self.poll_pending_ops(cx);
-      self.async_op_response(overflow_response)?;
-      self.drain_macrotasks()?;
-      self.check_promise_exceptions()?;
+      let context = self.global_context();
+      let scope =
+        &mut v8::HandleScope::with_context(self.v8_isolate(), context);
+      scope.perform_microtask_checkpoint();
```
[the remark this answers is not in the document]
bartlomieju: Thanks for pointing this out! Actually this checkpoint might not be needed - I added it during debugging and didn't check if it's required. I'll check that before landing.

core/runtime.rs
bartlomieju: Yeah, we already use this method for other purposes and have been for quite some time - what would you suggest to do instead?

piscisaureus: We do it with modules too - and that should also change.

I'd assign the promise a number and stick it in a HashMap<u64, v8::Global<v8::Promise>>

bnoordhuis: You probably need to somehow get back the id when you have a bare promise, I suppose?

It could be stored on the promise itself with a `v8::Private` symbol but that requires rusty_v8 to grow bindings for `v8::Object::SetPrivate()` and `v8::Object::GetPrivate()`. I can add those, I had that on my todo list anyway.

<hr>

However, that won't work for modules because they're instances of `v8::Data`, not `v8::Object`. What could work is storing them in a `HashSet` and implement `Eq` and `Hash` ourselves.

`hash()` could use `module.get_identity_hash()` (duplicates are okay) and `eq()` could compare the script id. There's currently no binding for that but it's trivial to add.

Same approach should work for promises, but using `strict_equals()` in `eq()`.

core/runtime.rs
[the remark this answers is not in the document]
bartlomieju: Right, so it's not needed. Removed.

core/runtime.rs
bartlomieju: Thanks for suggestions - I think we could land this PR as is and I'll update whole `deno_core` to this new mechanism in the follow up PR

Issue #6888, opened 2020-07-27
[std/testing/asserts] Rename *Contains to *Includes


jsejcksn:
I think sticking with the existing naming conventions used by the JavaScript ecosystem (when practical) can help reduce potential confusion (or at minimum reduce cognitive load), so I propose renaming these functions:

- [`assertArrayContains`](https://doc.deno.land/https/deno.land/std/testing/asserts.ts#assertArrayContains) ➡️ `assertArrayIncludes`
- [`assertStringContains`](https://doc.deno.land/https/deno.land/std/testing/asserts.ts#assertStringContains) ➡️ `assertStringIncludes`

Ref: [`Array.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes), [`String.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)

If there is agreement, I'd like to work on this PR.

jacobgc:
I think your propsal is a good idea, might be best leaving in contains for a few versions with a deprication notice however!

zekth:
`Contains` was firstly introduced in inspiration of Jest `expect` functions : https://jestjs.io/docs/en/expect.html

kevinkassimo:
Aliasing might work? I remember when people originally introduces `Array.prototype.includes()` it is due to `.contains()` collision with some existing popular libraries (`.contains()` preferred over `.includes()`, but forced to go the other direction. We don't have this problem here)

caspervonb:
If we're just testing for `a.includes(b)` then `assertIncludes` would be preferable IMHO.

jsejcksn:
@caspervonb There are two functions listed, one for strings and one for arrays whose members can contain non-string types as well. Do you mean that you want a single, overloaded function instead of one for each data type? Will you clarify?

timreichen:
@jsejcksn In favor of renaming.
@kevinkassimo Not in favor of aliasing. Since std is versioned, we can do breaking changes and have so in the past.
@caspervonb In favor of  your idea: `assertIncludes` that handles both string and array.

caspervonb:
> @caspervonb There are two functions listed, one for strings and one for arrays whose members can contain non-string types as well. Do you mean that you want a single, overloaded function instead of one for each data type? Will you clarify?

If the intent is to check for `target.includes(source)`, yes a single function.

And It's fine to break std; within reason.

[...]

PR #7951, opened 2020-10-13
rename(std/testing): rename assert*Contains to assert*Includes

timreichen:
#6888

# Rename functions in asserts.ts
- assertStringContains -> assertStringIncludes
- assertArrayContains -> assertArrayIncludes


bartlomieju:
I'm generally in favor of this change, the only point of discussion is original suggestion from #6888 if we should have both `assertStringIncludes` and `assertArrayIncludes` or just `assertIncludes` as implemented in this PR. Given that `assertEquals` is magical and works on all values makes `assertIncludes` favorable. 

CC @kitsonk @nayeemrmn @caspervonb

caspervonb:
Difficult to see what has actually changed here with 232 changes files but in favour of generalising on `a.includes(b)` as long as we provide rich pretty printed diffs for both strings and array-likes (otherwise `assert(a.includes(b))` is just as good.

std/testing/asserts.ts
```diff
@@ -330,69 +331,66 @@ export function assertNotStrictEquals(
...
   msg?: string,
 ): void {
-  const missing: unknown[] = [];
-  for (let i = 0; i < expected.length; i++) {
-    let found = false;
-    for (let j = 0; j < actual.length; j++) {
-      if (equal(expected[i], actual[j])) {
-        found = true;
-        break;
+  if (typeof actual === "string" && typeof expected === "string") {
```
caspervonb, line 363: Surprising that you're special casing here, behavior should be the same across all types.

nayeemrmn:
> if we should have both `assertStringIncludes` and `assertArrayIncludes` or just `assertIncludes` as implemented in this PR

I think they should definitely be different functions. They have different and unrelated behaviour, which would be especially confusing since strings are iterable. The JS methods are namespaced to `Array.prototype.includes()` and `String.prototype.includes()`. Making them overloads is illogical and makes both the implementation and call sites harder to read.

timreichen:
@caspervonb The special casing is because of the overload. To avoid that, @nayeemrmn suggestion might be the better solution. Just a rename of `contains` to `includes`.

bartlomieju:
> > if we should have both `assertStringIncludes` and `assertArrayIncludes` or just `assertIncludes` as implemented in this PR
> 
> I think they should definitely be different functions. They have different and unrelated behaviour, which would be especially confusing since strings are iterable. The JS methods are namespaced to `Array.prototype.includes()` and `String.prototype.includes()`. Making them overloads is illogical and makes both the implementation and call sites harder to read.

Thanks for being voice of reason @nayeemrmn. In this light I agree that we should just do the rename without merging those assertion. @timreichen could you please update the PR?

timreichen:
@bartlomieju already did:)

Releases.md
```diff
@@ -487,7 +487,7 @@ Changes in std version 0.59.0:
 - fix(std/node): fix readFile types, add encoding types (#6451)
 - fix(std/node): global process should usable (#6392)
 - fix(std/node/process): env, argv exports (#6455)
-- fix(std/testing) assertArrayContains should work with any array-like (#6402)
+- fix(std/testing) assertArrayIncludes should work with any array-like (#6402)
```
bartlomieju, line 490: There should be no changes to this file, revert

bartlomieju (changes requested):

Releases.md
timreichen: O shoot, I thought I excluded that file. Reverted.

bartlomieju:
@timreichen this PR shouldn't change 3k lines diff - I think something went wrong during rebase, please check

timreichen:
@bartlomieju Shoot, I suck at rebasing. Sorry. Done now.

bartlomieju (commented):
LGTM and I'm in favor of landing this PR

@lucacasonato @kitsonk please take a look

lucacasonato:
LGTM. We should definitely mention this in the blog post.

bartlomieju (approved):

PR #7952, opened 2020-10-13
refactor(cli): use Emit enum and rename ts_build_info

kitsonk:
This cleans up the emits in the specifier handler by using an enum with fields for the emit, instead of separate values.  It also renames `build_info` to `ts_build_info` in `SpecifierHandler` and `ModuleGraph2` for clarity purposes.


bartlomieju (approved):
LGTM, nice clean up

PR #7507, opened 2020-09-16
feat: add alert, confirm, and prompt

kt3k:
This PR adds `alert`, `confirm`, and `prompt` functions from [web standards](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#user-prompts).

You can write simple interaction script like the below with this change.

```ts
const name = prompt('What is your name?');
console.log(`Hello ${name}!`);
```

```console
$ ./target/debug/deno run test.ts
What is your name? John
Hello John!
```

Maybe these features are not so important for advanced users, but I think these could be very helpful for beginners of Deno or learners of the language.

---
closes #4257

Issue #4257, opened 2020-03-05
Request: alert, confirm, prompt, and their async counterpart


KSXGitHub:
## Rationale

The browser already provides `window.alert`, `window.confirm`, `window.prompt`. I think it would be nice if Deno supports them too.

## Implementation

* `window.alert` simply prints text to stdout and wait for user to press any key.
* `window.confirm` asks a yes/no question, waits to for user to respond either "y", "yes", "n", or "no".
* `window.prompt` asks user to enter a string.
* All the above methods block event loop.
* It would be nice to have their async counterpart as well. But they should belong to `Deno` namespace.

ry:
Sounds reasonable...

kitsonk:
My fear is that they could easily become vectors for a DOS attacks, so we would want to put them behind permission flags.  They have been abused so much in browsers it lead to the notifications API.  I would much rather see us implement the notifications API and move the permissions to the global namespace.

KSXGitHub:
@kitsonk 

> My fear is that they could easily become vectors for a DOS attacks

What type of attacker you are talking about? I wouldn't allow my server to execute arbitrary code from the user so a malicious user shouldn't be able to call `window.alert` to block my server.

> behind permission flags

I don't think this is necessary either. These window methods only block when stdin is interactive (waiting for user to type in input). If stdin is not interactive (such as pipeline), then these method shouldn't block execution.

> They have been abused so much in browsers it lead to the notifications API

I fail to see any negative in this. It is true that `alert` in the browser was (not "is") annoying to the end user, it is due to the browser's poor implementation of dialog box. But it has changed now, `alert` on one Chrome tag would no longer affect operation on the other tag.

> I would much rather see us implement the notifications API

On Electron and [deno_webview](https://github.com/eliassjogreen/deno_webview), this is a must-have. But Deno is CLI only, it does not aim to provide full-fledged web APIs.

kitsonk:
> What type of attacker you are talking about? I wouldn't allow my server to execute arbitrary code from the user so a malicious user shouldn't be able to call `window.alert` to block my server.

But someone calling `window.confirm()` would.  The whole Deno security model is a trust no one model.  You wouldn't allow someone to read read from the file system, or connect to the network, but we flag those in Deno.

> But Deno is CLI only, it does not aim to provide full-fledged web APIs.

Exactly.  A good reason not to implement these.

[...]

PR #7959, opened 2020-10-13
Revert "feat: add alert, confirm, and prompt (#7507)"

bartlomieju:
This reverts commit 0dcaea72aeec52a566764b41b10d8fd1854d6fa4.

Revert #7507 - only after landing I noticed controversy in the original issue (#4257).

CC @kt3k @kitsonk


kt3k:
Security concern doesn't apply anymore because the same attack can be done with exiting APIs (such as `Deno.stdin.readSync`).

I agree with his second argument ( there is no reasonable way to use `alert` `confirm` `prompt` in a isomorphic way), but I believe this is not strongly against the addition of APIs.

I'd like @kitsonk to review my comment at https://github.com/denoland/deno/issues/4257#issuecomment-672009353. My point (motivation) of this addition is that these are very useful for the beginners and learners. Maybe nobody use these in serious / production environment, but these helps learners easily interact with Deno APIs in casual manner (and I hope that promote the use of Deno especially for newcomers)

kitsonk:
I am not a fan of this addition, but if I had a real problem with it, I would have spoken up in the PR.  It doesn't "break" Deno, so I would rather have arguments about more serious things! 😉

bartlomieju:
My views are in line with @kt3k opinion, closing without merge.

Issue #1725, opened 2019-02-09
print JS stack unexpected signal like ctrl+c


ry:
Imagine you have the program 
```js
while (true) {
  console.log("hello");
}
```
If one ctrl+c the process, it would be nice if we printed the JS stack.

bartlomieju:
For reference, tokio has [tokio-signal](https://github.com/tokio-rs/tokio/blob/master/tokio-signal/README.md) which can be used to intercept signals

bartlomieju:
@mhvsa are you working on this issue?

Cre3per:
My rough idea was to install a signal handler in Rust that sets a bool flag. In a separate thread, wait for the flag to be set, then call `isolate.request_interrupt()`, `current_stack_trace()`, `isolate.terminate_execution()`, without any changes to the TS-side. The thread+bool might be unnecessary with tokio-signal, I haven't looked into that yet.

@mhvsa implemented a ctrl+c handler, but it was rejected by @kevinkassimo because of non-generality. Kevin also talked about resolving TS-side promises.  
Kevin wrote these comments before the addition of `Deno.signal()` in #3757. I'm wondering if those comments are still relevant?  

On another note, this is a feature I as a user would dislike because I ctrl+c all the time and want to see me logs, rather than a stack trace. Should this feature be enabled by a command-line flag?

bartlomieju:
Just FYI @Cre3per last time I tried it `current_stack_trace()` did not provide anything useful. At the same time `request_interrupt()` return an error that has no stack which also doesn't provide any useful info. Before trying to tackle it with signal handling I suggest to create a minimal example that would show that we are actually able to retrieve a stack trace at any point in time.

I'm also not particularly interested in implementing this feature. It's quite a lot of work for little gain.

Cre3per:
@bartlomieju Thanks for reporting the outcome of your attempts.  
I've got a minimal example working locally. It's incomplete, but enough for me to say that printing a stack trace on signal is possible.  
"Incomplete" and "locally", because I had to patch rusty_v8. I'm not showing code yet, because this still is highly experimental.  

I'm running a thread in `JsRuntime::mod_evaluate()` which calls

```
Isolate::request_interrupt() -> StackTrace::current_stack_trace() -> Isolate::terminate_execution()
```

This part is working, I see stack frames with line+column numbers.  
[...]

PR #7960, opened 2020-10-13
fix(cli/repl): keyboard interrupt should continue

caspervonb:
This changes the behavior of keyboard interrupts (ctrl+c) to continue, clearing the current line instead of exiting.

Exit can still be done with ctrl+d or by calling close().

### DIFF
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -308,7 +308,7 @@ pub async fn run(
         editor.lock().unwrap().add_history_entry(line.as_str());
       }
       Err(ReadlineError::Interrupted) => {
-        break;
+        continue;
       }
       Err(ReadlineError::Eof) => {
         break;
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -308,6 +308,7 @@ pub async fn run(
         editor.lock().unwrap().add_history_entry(line.as_str());
       }
       Err(ReadlineError::Interrupted) => {
+        println!("exit using ctrl+d or close()");
         continue;
       }
       Err(ReadlineError::Eof) => {


David-Else:
This sounds really bad, ctrl+c means kill the process. It has always meant that. What is the point of this bizarre perversion of Unix?!

https://superuser.com/questions/169051/whats-the-difference-between-c-and-d-for-unix-mac-os-x-terminal

```
CtrlC tells the terminal to send a SIGINT to the current foreground process, which by default translates into terminating the application. CtrlD tells the terminal that it should register a EOF on standard input, which bash interprets as a desire to exit.
```

littledivy:
This seems to be the default behavior with Python, although Node does ask for confirmation.
image

caspervonb:
> This sounds really bad, ctrl+c means kill the process. It has always meant that. What is the point of this bizarre perversion of Unix?!

Bash clears and continues.
Python clears and continues.
Ruby clears and continues.
Node clears and continues (but does allow you to hit it again immediately to exit).

ry:
Maybe print "exit using ctrl+d or close()" again?

nayeemrmn:
Note that the existing SIGINT behaviour still remains when an actual execution in interrupted:
image

caspervonb:
> Maybe print "exit using ctrl+d or close()" again on ^c ?

Since you phrased that as maybe, I think its a bit noisy, never liked Node doing it.
Prefer mimicking bash/zsh as a good examples of prompts but I'll yield if anyone feels strongly about it.

> Note that the existing SIGINT behaviour still remains when an actual execution is interrupted:

Been playing around with the new Runtime.terminateExecution method but that ends up killing the whole process which is somewhat unexpected.

bartlomieju:
@nayeemrmn @caspervonb this is a long standing issue - https://github.com/denoland/deno/issues/1725 - we need some kind of signal watchdog to implement that (I know tokio has APIs for that), and most likely tinker with `deno_core` to get it working.

caspervonb:
> @nayeemrmn @caspervonb this is a long standing issue - #1725 - we need some kind of signal watchdog to implement that (I know tokio has APIs for that), and most likely tinker with deno_core to get it working.

In this case we are catching the signal, it's just a matter of what do we do with it but research required so out of scope for this PR.

bartlomieju:
> Maybe print "exit using ctrl+d or close()" again on ^c ?

That sounds good to me, but I agree with @caspervonb that we should clear the input line on ^c.

bartlomieju (approved):
LGTM

Issue #575, opened 2018-08-23
sign binaries


ry:

jornfranke:
Is there a plan when the binaries will be signed?

waynebloss:
Found on HN today: https://github.com/indygreg/PyOxidizer/tree/main/apple-codesign

> `apple-codesign` is a crate implementing functionality related to code signing
on Apple platforms.
> 
> All functionality is implemented in pure Rust and doesn't require any 3rd party
or proprietary software nor do we require running on Apple platforms.

I wonder if this can be used for Deno binaries?

jornfranke:
Alternatively, there is also cosign: https://github.com/SigStore/cosign 
It has also hardware support for signing including hardware used in major cloud vendors (if CI pipeline runs in one of those)

cowboyd:
@waynebloss Their `rcodesign` binary does appear to work on Deno compiled binaries, so it could be an option. I have not tried notarizing and / or stapling

cowboyd:
UPDATE:

We tried to codesign using the `apple-codesign` project linked above, but it corrupted the resulting executable and trying to run it resulted in the following error:

```
#
# Fatal process OOM in Failed to reserve virtual memory for CodeRange
#
```

We tried to shave this yak from all manner of different positions, but in the end, we couldn't get it to work. For the time being, we're just going to follow the advice here https://users.rust-lang.org/t/distributing-cli-apps-on-macos/70223 and  have users install via `curl`. Since our use-case is a developer tool this is acceptable for now.

jornfranke:
Probably apple-codesign is not the right tool for this. Why not https://github.com/SigStore/cosign?

jornfranke:
or if you want to be more traditional one can sign the binaries with GnuPG: https://www.gnupg.org/gph/en/manual/x135.html

jornfranke:
Thanks for the detailed response. Then the non-existing description of the issue is misleading. I do not care about Apple signing, but in general all binaries (linux, windows etc.) should be signed so I can verify myself that they have not been tampered with. Maybe then I should create a dedicated issue for this?

cowboyd:
@jornfranke This could be what you're looking for https://github.com/denoland/deno/issues/11154

There, @AlexanderOMara lays out a method to make it work for Mach-O, but from my (limited) understanding of the subject, it would require a separate strategy for assembling the standalone binary for each platform executable format (Mach-O, ELF, PE) rather than the magical trailer strategy used now.

[...]

PR #7961, opened 2020-10-13
[WIP] Web Notification API

littledivy:
Uses `notify_rust` which is quite stable on windows and linux, although uses deprecated NS* API for OSX (also used by Tauri). 

The new OSX api for sending notifications is [UN* API](https://developer.apple.com/documentation/usernotifications/unmutablenotificationcontent?language=objc) requires that the application be code-signed whereas `NSUserNotification` does not have this requirement.


littledivy:
- [ ] Constructor as per spec https://notifications.spec.whatwg.org/#dom-notification-notification
- [ ] Events wherever possible (`onclose`, `onerror`)

zorbyte:
I know it's not my place, but is it possible to get Deno coee signed if they wanted to?

crowlKats:
@zorbyte #575

littledivy:
On hold - I'm waiting to know whether we are "ok" with using the deprecated NS* API for OSX

bartlomieju:
> On hold - I'm waiting to know whether we are "ok" with using the deprecated NS* API for OSX

@littledivy is there an alternative to this approach?

crowlKats:
@bartlomieju as the PR description states, builds would need to be signed. Also as far as my research went there is no crate for the current API, so one would also need to make that. It is a lot of work, but I am personally against using a deprecated API.

littledivy:
I agree with @crowlKats - we should not use the deprecated API. The solution would be to implement a crate for the current API.

lucacasonato:
I am going to close this, as I don't think we are going to land this in the close future as it requires signing the binary, and would bind us to window managers, something we have decided not to pursue in the CLI in the near future with WebGPU. Before we revisit this, more thorough discussion should take place in an issue.

PR #7966, opened 2020-10-14
docs: Mention how to use a specific shell for Deno.run

### LINKED ISSUE #7847: Deno.run options for shell control
At the moment, when running on windows, I seem to be locked in to using the Win32 CMD shell, which seems to be having issues running npm. I'm currently using Git-Bash which has all the correct environment variables/path set up, so being able to specify a shell would be really useful.

If I'm not just missing something and there's a consensus on what the api should be, I'd be happy to put in the leg work to create a pull request.

josh-hemphill:
Closes #7847
For people coming from node where there are specific options for passing the shell.


docs/examples/subprocess.md
```diff
@@ -10,6 +10,7 @@
   [stdin](https://doc.deno.land/builtin/stable#Deno.stdin),
   [stdout](https://doc.deno.land/builtin/stable#Deno.stdout) and
   [stderr](https://doc.deno.land/builtin/stable#Deno.stderr) streams.
+- Use a specific shell by providing its path/name and it's string input switch, e.g. `Deno.run({cmd: ["bash", "-c",'"ls -la"']});
```
bnoordhuis, line 13:
```suggestion
- Use a specific shell by providing its path/name and its string input switch, e.g. `Deno.run({cmd: ["bash", "-c", '"ls -la"']});
```
The double quotes aren't necessary with bash, that's a cmd.exe quirk.

`-c` is correct for bash, `/d /s /c` is what you want for cmd.exe.

josh-hemphill:
Thank you @bartlomieju! I couldn't seem to get the python scripts to run to completion on my windows machine or under WSL.

bartlomieju (approved):
LGTM, thanks @josh-hemphill

PR #7967, opened 2020-10-14
refactor(cli/repl): clean up prelude injection

caspervonb:
This extracts prelude injection into a helper function and moves the prelude string literal into a top level static string to help trim some of the fat out of the run function.


bartlomieju (approved):
LGTM, nice cleanup

Issue #7088, opened 2020-08-17
async iterator can cause silent process exit in `Deno.test()`


lareau:
The following script `test.ts` causes `deno test` to silently perform the equivalent of `exit(0)` from within the test `fn()`:

```
$ deno test ./test.ts && echo OK
running 2 tests
test bug ... OK
```

All works as expected if the `bug` test is filtered out:

```
$ deno test ./test.ts --filter ok && echo OK
running 1 tests
test ok ... ok (3ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out (3ms)

OK
```

My version info:

```
$ deno --version
deno 1.3.0
v8 8.6.334
typescript 3.9.7
$ grep DESC /etc/lsb-release
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"
$ uname -s -p -v
Linux #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64
```

Note: This is a minimal reproducer for a bug that showed up in a timeout-based iterater, thus the odd construction.

File `test.ts`:

```typescript
class FooIterator implements AsyncIterableIterator<number> {
  resolve: any;

  constructor() {
    this.resolve = null;
  }

  [Symbol.asyncIterator](): AsyncIterableIterator<number> {
    return this;
  }

  next(): Promise<IteratorResult<number>> {
    return new Promise((res) => {
      this.resolve = res;
    });
  }
}

Deno.test({
  name: "bug",
  async fn() {
    const foo = new FooIterator();
    for await (const value of foo) {
      console.log(`value=${value}`);
    }
  },
});

Deno.test({
  name: "ok",
  async fn() {},
});
```

magurotuna:
I'm investigating this issue.

magurotuna:
This is caused because neither `resolve` nor `reject` is not called in promise. The promise state is "pending" forever, so the subsequent steps won't be executed.

So true minimum reproducible would be:

```ts
Deno.test({
  name: "bug",
  fn() {
    return new Promise((_resolve, _reject) => {
      console.log("in promise");
      // Neither `resolve` nor `reject` is called
    });
  },
});

Deno.test({
  name: "ok",
  async fn() {
    console.log("ok test");
  },
});
```

The output looks like:

```
running 2 tests
test bug ... in promise
```

[...]

PR #7968, opened 2020-10-14
fix(test): return error when awaiting unresolved promise

bartlomieju:
This commit fixes test runner by awaitning "Deno.runTests()" call,
which ensures proper error is returned when there's an unresolved
promise that's being awaited.

Fixes https://github.com/denoland/deno/issues/7088


ry (approved):
LGTM

Issue #7970, opened 2020-10-14
web_util cloneValue returns Map instead of Set


00ff0000red:
https://github.com/denoland/deno/blob/374d433f1f4885686dc5c166be3a25c839e22959/cli/rt/01_web_util.js#L94-L99

It seems that when given a `Set`, `cloneValue` returns a `Map` instead. Is that intentional?

PR #7972, opened 2020-10-14
fix: "cloneValue" should return a Set when given a Set

00ff0000red:
Fixes related issue: (https://github.com/denoland/deno/issues/7970).

cli/rt/web_util cloneValue function returns a `Map` instead of a `Set`, it currently seems like a copy & paste from the code right above the code in question.


00ff0000red:
Actually, this is pretty weird. Shouldn't the keys of the map be cloned too? I'm not sure if I want to touch this without more knowledge about where this is used, and what it is intended to do?

Not all objects are clonable, yet it defaults to assuming, that if the input was an object, that the input is a basic object. Shouldn't it throw errors on some stuff if it is intended to mimic `StructureDeserialize(StructuredSerialize(...)`?

bartlomieju:
@kitsonk PTAL, IIRC you wrote `cloneValue`

00ff0000red:
> @kitsonk PTAL, IIRC you wrote `cloneValue`

Almost forgot about this, any extra info about how this was actually intended to function would be helpful.

kitsonk:
@bartlomieju I adapted it from the streams repo at the time, which the reference implementation is no longer there and has been re-implemented.  I found another reference somewhere that had the same "error".  I can't see logically how it should be the way it is written and looks like a mistake that has been perpetuated.

As far as the expected behaviour, it should be just like `Map`, but for `Set` as far as the values it contains.

00ff0000red:
I found another reference somewhere that had the same "error". I can't see logically how it should be the way it is written and looks like a mistake that has been perpetuated.
> 
> As far as the expected behaviour, it should be just like `Map`, but for `Set` as far as the values it contains.

Git, I'll fix this up in a few hours. If all of Deno doesn't die, it should work just fine.

00ff0000red:
@kitsonk How closely should this `cloneValue` function mimic structured cloning?
From experience, V8's implementation of structured cloning skips over symbol properties in objects, ex:
```js
const source = {
    "string key": "string key",
    [Symbol("symbol key")]: "symbol key"
};
// after cloning
const clonedObject = {
    "string key": "string key"
};
```

If we want a closer match, I could go interpret the ES spec, and try to refactor it in order to mimic it closer, but if it doesn't matter, it could left as is.

kitsonk:
> Also, what linter does Deno's JavaScript source code use? Because it fails consistently with my updates.

It is the formatter that is failing.  It is the same formatter that is built into Deno.  You should rebase this PR as there were changes to the tooling of the build system.  With it rebased, you can potentially run:

```
./tools/format.js
```

From the root of the local deno repo, or you will need to run:

```
deno run --unstable --allow-write --allow-read --allow-run ./tools/format.js
```

00ff0000red:
> It is the formatter that is failing. It is the same formatter that is built into Deno. You should rebase this PR as there were changes to the tooling of the build system. With it rebased, you can potentially run:

Now, I apologize for my naivety, but by "rebase," are you referring to closing this PR, and just opening a new one?

bartlomieju:
> > It is the formatter that is failing. It is the same formatter that is built into Deno. You should rebase this PR as there were changes to the tooling of the build system. With it rebased, you can potentially run:
> 
> Please forgive my naivety, but by "rebase," are you referring to closing this PR, and just opening a new one?

@00ff0000red you need to fetch latest version of the repo and then merge `master` branch into your branch.

00ff0000red:
@bartlomieju
@kitsonk
Rebased, as suggested, found the formatting problem and resolved it, just need someone to review the commits now.

bartlomieju:
@00ff0000red can you rebase once more? I tried doing that, but it seems I can't push to your branch. I'd be useful if you checked "Allow edits from maintainers" checkbox. https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

00ff0000red:
> @00ff0000red can you rebase once more?

Permissions updated, you may now rebase it yourself, if you want to again.
I'll rebase it right now.

bartlomieju:
@00ff0000red thanks, could you also update `cloneValue` in `op_crates/fetch/11_streams.js`?

00ff0000red:
> @00ff0000red thanks, could you also update `cloneValue` in `op_crates/fetch/11_streams.js`?

I'll do it, but I strongly believe that both crates should share the save function reference, as they are literal copies, the only difference is that they don't share the same `objectCloneMemo`.

Also, if I might ask, why even have a `objectCloneMemo`?

bartlomieju:
> > @00ff0000red thanks, could you also update `cloneValue` in `op_crates/fetch/11_streams.js`?
> 
> I'll do it, but I strongly believe that both crates should share the save function reference, as they are literal copies, the only difference is that they don't share the same `objectCloneMemo`.


@00ff0000red I'm aware that the code should be shared, but currently we don't have a good way to do that between crates. 

> Also, if I might ask, why even have a `objectCloneMemo`?

Not sure, maybe @kitsonk will know

00ff0000red:
> @00ff0000red thanks, could you also update `cloneValue` in `op_crates/fetch/11_streams.js`?

Done.
Also realized that I would've caused an error in the original update, regressed part of what I had done.
Formatter check failed though.

bartlomieju (approved):
LGTM, thank you @00ff0000red

PR #7975, opened 2020-10-14
fix(cli/repl): respect the value of Deno.noColor

caspervonb:
This sets the value of colors during the inspectArgs call to !Deno.noColor.

Should make https://github.com/denoland/deno/pull/7778 landable.


bartlomieju:
Included in https://github.com/denoland/deno/pull/7976

PR #7977, opened 2020-10-14
feat(op_crates): webgpu

### LINKED ISSUE #7863: WebGPU Implementation
I have been looking at the [WebGPU spec](https://gpuweb.github.io/gpuweb/) and the [`wgpu` crate](https://github.com/gfx-rs/wgpu-rs) for a possible WebGPU implementation which arose from a discussion on the discord server.
Problem is, how would we handle the window on which the user would draw to? Adding a window manager to deno would definitively not be a viable solution. `wgpu-rs` does actually allow to have a windowless setup, but i think most end users would actually want a way to have some sort of window. A window is also required for usage of swap chains; which, once again, many users would want.

crowlKats:
Closes #7863

missing:

- [x] create navigator & workerNavigator and add webgpu to them
- [x] ~create a webgpu instance with the `op_webgpu_create_instance` op~ now handled at init and set in OpState
- [x] get `op_webgpu_buffer_get_map_async` to work
- [x] have the slice in `op_webgpu_buffer_get_mapped_range` be put into zero_copy
- [x] fix PassChannel structs
- [x] TS types
- [ ] ~~Move TS types to unstable~~ (do this in follow up)
- [ ] ~~WPT tests~~ (do this in follow up)
- [x] Resource cleanup
- [x] WebIDL
- [x] Implement Deno.customInspect
- [x] add `[[device]]` internal slot to all objects except adapter
- [x] closes resources on finalization for consumables
- [ ] ~~Cleanup automatically using FinalizationRegistry~~ (do this in follow up)
- [x] validate that objects belong to same device
- [x] check if objects are valid before using them (check if rid is present)

Examples can be found at https://github.com/crowlKats/webgpu-examples/


cli/ops/webgpu/buffer.rs
```diff
@@ -0,0 +1,224 @@
...
+  let sender_ptr = unsafe {
+    std::mem::transmute::<*mut oneshot::Sender<Result<(), AnyError>>, *mut u8>(
+      sender_ptr,
+    )
+  };
+
+  extern "C" fn buffer_map_future_wrapper(
+    status: wgc::resource::BufferMapAsyncStatus,
+    user_data: *mut u8,
+  ) {
```
bartlomieju, line 125: Not sure if this is correct

crowlKats: Doesnt seem to be; just tried and it hangs

lucacasonato:
- GPUAdapter
  - GPUDevice (destroyable, non consumable)
    - GPUBuffer (destroyable, non consumable)
    - GPUTexture (destroyable, non consumable)
      - GPUTextureView ( non consumable)
    - GPUSampler (non consumable)
    - GPUBindGroupLayout (non consumable)
    - GPUPipelineLayout (non consumable)
    - GPUBindGroup (non consumable)
    - GPUShaderModule (non consumable)
    - GPUComputePipeline (non consumable)
    - GPURenderPipeline (non consumable)
    - GPUCommandEncoder (consumable)
      - GPURenderPassEncoder (consumable)
      - GPUComputePassEncoder (consumable)
    - GPUCommandBuffer (non consumable)
    - GPURenderBundleEncoder (consumable)
    - GPURenderBundle (non consumable)
    - GPUQuerySet (destroyable, non consumable)
    - GPUQueue (non consumable)

op_crates/webgpu/01_webgpu.js
```diff
@@ -1618,6 +1618,10 @@
      * @return {GPURenderPassEncoder}
      */
     beginRenderPass(descriptor) {
+      if (this[_rid] === undefined) {
+        throw new DOMException("Failed to execute 'beginRenderPass' on 'GPUCommandEncoder': already consumed", "OperationError");
+      }
+
```
lucacasonato: This should happen after the WebIDL logic @crowlKats

ry:
I did some size comparisons with main branch:

webgpu
target/debug/deno 248M
target/release/deno 72M

main
target/debug/deno 212M
target/release/deno 66M

Quite a bit larger.

ry:
This is an epic feature that opens the door to doing serious scientific computing with JavaScript. We will likely be able to run TensorFlow.js on Deno without too much work after this. Having a dynamic runtime with easy GPU access is a game changer.

@crowlKats and @lucacasonato have been working on this for months - it's hard to understand the effort that they have put in to deliver this branch. Thank you for your work on this!

- Superficially the code looks good to me. 
- My main concern the complete lack of tests. I understand that in the future we are aiming for WPT coverage for WebGpu, but that it will require more work to get there. That's ok, but for the time being we ought to add some basic smoke tests that exercise various parts of the API. 
- I also think a large crate like this should have a README with an introduction to the code base - some description of how it's organized, links to docs, and a todo list future work. 
- I am also concerned about the 8mb size increase (this on top of the ICU size increase would make Deno 1.8 50% larger than 1.7) but it doesn't seem like there's much we can do about that.

op_crates/webgpu/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+pub mod bundle;
+pub mod command_encoder;
+pub mod compute_pass;
+pub mod error;
+pub mod pipeline;
+pub mod queue;
+pub mod render_pass;
+pub mod sampler;
+pub mod shader;
+pub mod texture;
```
ry, line 55: Is there a reason for these to be pub? It's not used externally, right?

lucacasonato: All the ops are imported in `runtime/` so they can be registered on the `JsRuntime`.

ry (approved):
LGTM :)

lucacasonato:
Huge thanks @crowlKats and @kvark!

PR #7980, opened 2020-10-15
ci: Use GITHUB_ENV instead of set-env

peaceiris:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

> Action and workflow authors who are setting environment variables via stdout should update any usage of the set-env and add-path workflow commands to use the new environment files.


littledivy:
Duplicate of https://github.com/denoland/deno/pull/7788

PR #7981, opened 2020-10-15
fix(cli): ModuleGraph2 properly handles redirects

kitsonk:
Fixes issues where both `--no-check` and `deno info` do not currently handle redirects.

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -32,9 +32,6 @@ use std::sync::Arc;
 use std::sync::Mutex;
 
 /// Structure representing local or remote file.
-///
-/// In case of remote file `url` might be different than originally requested URL, if so
-/// `redirect_source_url` will contain original URL and `url` will be equal to final location.
 #[derive(Debug, Clone)]
 pub struct SourceFile {
   pub url: Url,
--- a/cli/module_graph2.rs
+++ b/cli/module_graph2.rs
@@ -161,7 +161,6 @@ fn get_version(source: &str, version: &str, config: &[u8]) -> String {
 struct Module {
   dependencies: DependencyMap,
   is_dirty: bool,
-  is_hydrated: bool,
   is_parsed: bool,
   maybe_emit: Option<Emit>,
   maybe_emit_path: Option<(PathBuf, Option<PathBuf>)>,
@@ -180,7 +179,6 @@ impl Default for Module {
     Module {
       dependencies: HashMap::new(),
       is_dirty: false,
-      is_hydrated: false,
       is_parsed: false,
       maybe_emit: None,
       maybe_emit_path: None,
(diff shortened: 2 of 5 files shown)


kitsonk:
Bit more detail about this...

ModuleGraph2 no has a hash of all the redirects and when a module is attempted to be resolved, it will keep recursively looking up any redirects until the point where it no longer gets redirected, and then attempts to look the module up in the graph.

Currently the file_fetcher simply returns the "final" Url for a fetched file, so when the module is added to the graph, it only has one potential redirect.  If in the future we surface up the entire change, that will "just work" with the graph, because of the recursive way it resolves modules.

I've tried this on my `run`/`check` migration branch and it seems to fix all the cases we have tests for.

cli/file_fetcher.rs
```diff
@@ -32,9 +32,6 @@ use std::sync::Arc;
 use std::sync::Mutex;
 
 /// Structure representing local or remote file.
-///
-/// In case of remote file `url` might be different than originally requested URL, if so
-/// `redirect_source_url` will contain original URL and `url` will be equal to final location.
```
bartlomieju, line 37: Why remove this comment?

cli/module_graph2.rs
```diff
@@ -198,51 +196,49 @@ impl Default for Module {
...
     };
-    self.maybe_version = cached_module.maybe_version;
-    self.is_dirty = false;
-    self.is_hydrated = true;
+    module
+  }
+
+  /// Return `true` if the current hash of the module matches the stored
+  /// version.
+  pub fn emit_valid(&self, config: &[u8]) -> bool {
```
bartlomieju, line 236: Nitpick - rename to `is_emit_valid()`?

cli/module_graph2.rs
```diff
@@ -666,6 +674,17 @@ impl Graph2 {
...
+  fn resolve_specifier<'a>(
+    &'a self,
+    specifier: &'a ModuleSpecifier,
+  ) -> &'a ModuleSpecifier {
+    let mut s = specifier;
+    while let Some(redirect) = self.redirects.get(s) {
+      s = redirect;
+    }
+    s
+  }
```
bartlomieju, line 686: I wonder if we should guard against infinite redirect loop here - it might already be handled in `SourceFileFetcher` tho...

bartlomieju (approved):
LGTM

cli/file_fetcher.rs
kitsonk: Because there is no `redirect_source_url`.  It was refactored out at some point and no one bothered to update the comment.

cli/module_graph2.rs
kitsonk: It isn't specifically guarded, but it would like cause errors elsewhere first (in `file_fetcher`).  I will take a look though.

kitsonk: I have guarded against infinite loops and excessive redirections... While it might be overkill, it is good to have clear logic to handle those edge cases.

PR #7987, opened 2020-10-15
fix(cli/worker): Print error stacks from the origin Worker

### LINKED ISSUE #4728: Error stack in worker missing
Currently, every error thrown in worker has no trusted stacks, but only
```
at WorkerImpl.#poll ($deno$/web/workers.ts:140:17)
```

nayeemrmn:
Fixes #4728.

cc @bartlomieju

### DIFF
--- a/cli/ops/worker_host.rs
+++ b/cli/ops/worker_host.rs
@@ -1,5 +1,6 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
+use crate::colors;
 use crate::fmt_errors::JsError;
 use crate::ops::io::get_stdio;
 use crate::permissions::Permissions;
@@ -162,6 +163,12 @@ fn run_worker_thread(
     }
 
     if let Err(e) = result {
+      eprintln!(
+        "{}: Uncaught (in worker \"{}\") {}",
+        colors::red_bold("error"),
+        name,
+        e.to_string().trim_start_matches("Uncaught "),
+      );
       sender
         .try_send(WorkerEvent::TerminalError(e))
         .expect("Failed to post message to host");
--- a/cli/ops/worker_host.rs
+++ b/cli/ops/worker_host.rs
@@ -9,7 +9,9 @@ use crate::tokio_util::create_basic_runtime;
 use crate::worker::WebWorker;
 use crate::worker::WebWorkerHandle;
 use crate::worker::WorkerEvent;
+use deno_core::error::generic_error;
 use deno_core::error::AnyError;
+use deno_core::futures::channel::mpsc;
 use deno_core::futures::future::FutureExt;
 use deno_core::serde_json;
 use deno_core::serde_json::json;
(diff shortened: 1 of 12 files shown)


bartlomieju (commented):
I think this looks great and we could land it, but want to get another pair of eyes on it

cli/ops/worker_host.rs
```diff
@@ -40,6 +51,21 @@ pub fn init(rt: &mut deno_core::JsRuntime) {
   );
   super::reg_json_sync(rt, "op_host_post_message", op_host_post_message);
   super::reg_json_async(rt, "op_host_get_message", op_host_get_message);
+  super::reg_json_sync(
+    rt,
+    "op_host_unhandled_error",
+    move |_state, args, _zero_copy| {
```
kitsonk, line 57: Was there a reason to not be consistent with the other ops and create a seperate function?

cli/ops/worker_host.rs
```diff
@@ -40,6 +51,21 @@ pub fn init(rt: &mut deno_core::JsRuntime) {
...
+      if let Some(mut sender) = sender.clone() {
+        let args: HostUnhandledErrorArgs = serde_json::from_value(args)?;
+        sender
+          .try_send(WorkerEvent::Error(generic_error(args.message)))
+          .expect("Failed to propagate error event to parent worker");
+        Ok(json!({}))
```
kitsonk, line 63: In other cases, we respond with `json!(true)` for a "non" response.

cli/tests/subdir/worker_error.ts
```diff
@@ -0,0 +1,5 @@
+function foo() {
+  throw new Error("foo");
```
kitsonk, line 2: What happens with other built-in error classes like `TypeError`, `RangeError`, `SyntaxError`?  Do we need to send more than just the error message?

kitsonk (commented):
A couple thoughts.

cli/ops/worker_host.rs
nayeemrmn: It's a closure that uses structures only safe (or convenient) to retrieve during init, not available in op context under current infrastructure. Suggested here: https://discord.com/channels/684898665143206084/684911491035430919/766354819925278750.

cli/tests/subdir/worker_error.ts
nayeemrmn: The "message" being sent is `Uncaught <error.name>: <error.message>`. This is all that's sent to the host in browsers: 
image

cli/ops/worker_host.rs
nayeemrmn, line 63:
```suggestion
        Ok(json!(true))
```

cli/tests/subdir/worker_error.ts
kitsonk: Ok, cool!

kitsonk (approved):
LGTM!

PR #7989, opened 2020-10-15
chore: separate declaration files for unstable

lucacasonato:
This PR splits the `deno.unstable` lib into `deno.shared_globals.unstable`, `deno.window.unstable`, `deno.worker.unstable`, and `deno.ns.unstable`.

This is done in preparation for the introduction of WebGPU to `--unstable` in #7977. It will add `window.navigator` and `self.navigator`, and the typings for WebGPU itself: https://gist.github.com/crowlKats/e00b0e22d7f3e7bfb003b714c1617327. It does not make sense for these to be in the current `deno.unstable` lib, because it only extends `deno.ns`, and WebGPU is a web API.


lucacasonato:
Closing for now, as there is no immediate need.

Issue #7771, opened 2020-10-01
Proposal: std/jwt


timreichen:
As there was a discussion about different specs including JSON web tokens [here](https://github.com/denoland/deno/issues/4203) and there is a [3rd party module](https://deno.land/x/djwt)  implementation at the moment, I think it would be an enrichment to `std` to have JWT included.

What are your thoughts on that?

axetroy:
JWT depends on the crypto (SHA256) module.

So again, implement the crypto module before implement jwt.

timonson:
Hi @axetroy sha256 and sha512 already exist in std: https://deno.land/std@0.69.0/hash/sha256.ts

I am the author of [djwt](https://github.com/timonson/djwt) and would love to help moving jwt to std.

timreichen:
@timonson awesome!
I had a look at your module, looks really good! Though some structural changes and function renames will be needed to be well-aligned with `js` and deno `std` conventions. I would be glad to help you with that.

@ry @bartlomieju would you be in favor of that integration?

ry:
Happy to have it, patches welcome.

timonson:
@timreichen I have the same username on the discord channel and I would love to hear more about your perspective of these things. So if you can, please dm me in the following days.

PR #7991, opened 2020-10-15
add(std/jwt): Implement the new jwt module

timonson:
This PR contains the new `std/jwt` module and therefore closes https://github.com/denoland/deno/issues/7771 .

@timreichen and I put equal amounts of work into this. So big *Thank you* to him!

A few additional things to consider:

- Although not necessary now, `verify` and `create` return promises because we
  expect coming encryption functions to return promises in the future. E.g. for
  `RSA` algorithms.

- This application uses the JWS Compact Serialization according to the
  [JWS specification](https://www.rfc-editor.org/rfc/rfc7515.html#section-7).


std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+
+function tryToParsePayload(input: string) {
+  try {
+    return JSON.parse(input);
+  } catch {
+    return input;
+  }
+}
+
+export function decode(
```
kitsonk, line 66: For a public API, we should have some JSDoc (with examples) and a "concrete" return value...  I am also surprised we can't have a narrow type than `unknown`.

kitsonk: Ok, I see you have `isTokenObject()` but the examples in the README don't use this.  Why would we want to potentially not validate the token object as part of the decode... no one wants an unvalidated object do they?

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  };
+}
+
+export type TokenObject = {
+  header: Header;
+  payload: PayloadObject | string;
+  signature: string;
+};
+
+export function isTokenObject(object: TokenObject): object is TokenObject {
```
kitsonk, line 98: JSDoc with examples please.

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  );
+}
+
+export async function verify(
+  jwt: string,
+  key: string,
+  {
+    algorithm = "HS512",
+  }: {
+    algorithm?: Algorithm | Array<Exclude<Algorithm, "none">>;
```
kitsonk, line 114: You have now used this in two places, a type alias would make sense.  Also, it would be good to make this into a concrete type, like `VerifyOptions` and export it, with some JSDoc.

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  return (
+    typeof object?.signature === "string" &&
+    typeof object?.header?.alg === "string" &&
+    (typeof object?.payload === "object" && object?.payload?.exp
+      ? typeof object.payload.exp === "number"
+      : true)
+  );
+}
+
+export async function verify(
```
kitsonk, line 108: JSDoc please.

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  jwt: string,
+  key: string,
+  {
+    algorithm = "HS512",
+  }: {
+    algorithm?: Algorithm | Array<Exclude<Algorithm, "none">>;
+  } = {},
+): Promise<Payload> {
+  const obj = decode(jwt) as TokenObject;
```
kitsonk, line 117:
```suggestion
  const obj = decode(jwt);
```

Your guard at 119 narrows it to a `TokenObject`, coercing it here is unsafe.

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  if (!isTokenObject(obj)) {
+    throw new Error(`jwt is invalid`);
+  }
+
+  if (
+    obj?.payload !== null &&
+    typeof obj?.payload === "object" &&
+    "exp" in obj.payload &&
+    isExpired(obj.payload.exp!, 1)
+  ) {
```
kitsonk, line 128: `obj` cannot possible be undefined here, because of the use of `isTokenObject` as 119.

```suggestion
  const { payload } = obj;
  if (
    payload !== null &&
    typeof payload === "object" &&
    "exp" in payload &&
    isExpired(payload.exp!, 1)
  ) {
```

The not null assertion should be able to be refactor out as well, something must be going on with the types that is unobvious to me doing the review.

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+    throw RangeError("jwt is expired");
+  }
+
+  if (!verifyAlgorithm(algorithm, obj.header.alg)) {
+    throw new Error(`algorithms do not match`);
```
kitsonk, line 133:
```suggestion
    throw new TypeError(`The asserted algorithm does not match the token's algorithm.`);
```

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  const { header, payload, signature } = obj;
+
+  /*
+   * JWS §4.1.11: The "crit" (critical) Header Parameter indicates that 
+   * extensions to this specification and/or [JWA] are being used that MUST be
+   * understood and processed.
+   */
+  if ("crit" in obj.header) {
+    throw new Error("implementation does not process 'crit' header parameter");
```
kitsonk, line 144:
```suggestion
    throw new Error("The 'crit' header parameter is currently not supported by this library.");
```

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  if (
+    !(await verifySignature({
+      signature,
+      key,
+      alg: header.alg,
+      signingInput: jwt.slice(0, jwt.lastIndexOf(".")),
+    }))
+  ) {
+    throw new Error("signatures do not match");
```
kitsonk, line 155:
```suggestion
    throw new Error("The token's signature do not match the verification signature.");
```

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  }.${
+    base64url.encode(
+      encoder.encode(
+        typeof payload === "string" ? payload : JSON.stringify(payload),
+      ),
+    )
+  }`;
+}
+
+export async function create(
```
kitsonk, line 182: JSDoc please.

std/jwt/signature.ts
```diff
@@ -0,0 +1,48 @@
...
+import { HmacSha512 } from "../hash/sha512.ts";
+import { encode as convertUint8ArrayToBase64url } from "../encoding/base64url.ts";
+import { decodeString as convertHexToUint8Array } from "../encoding/hex.ts";
+
+export function convertHexToBase64url(input: string): string {
+  return convertUint8ArrayToBase64url(convertHexToUint8Array(input));
+}
+
+function encrypt(
+  alg: Algorithm | "none",
```
kitsonk, line 12: `Algorithm` already includes `"none"`.

```suggestion
  alg: Algorithm,
```

std/jwt/signature.ts
```diff
@@ -0,0 +1,48 @@
...
+): string {
+  switch (alg) {
+    case "none":
+      return "";
+    case "HS256":
+      return new HmacSha256(key).update(message).toString();
+    case "HS512":
+      return new HmacSha512(key).update(message).toString();
+    default:
+      throw new RangeError(`algorithm '${alg}' in header is not supported`);
```
kitsonk, line 24:
```suggestion
      throw new RangeError(`The algorithm of "${alg}" in the header is not supported.`);
```

std/jwt/signature.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+export async function create(
```
kitsonk, line 28: If functions exported from this module are part of the public API, then they should have JSDoc.  If they are just supporting `mod.ts` then the whole module should start with an underscore (`_`).

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+
+function isExpired(exp: number, leeway = 0): boolean {
+  return exp + leeway < Date.now() / 1000;
+}
+
+/*
+ * Helper function: setExpiration()
+ * returns the number of seconds since January 1, 1970, 00:00:00 UTC
+ */
+export function setExpiration(exp: number | Date): number {
```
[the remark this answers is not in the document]
timreichen: maybe renaming to `createExpiration`?

std/jwt/mod.ts
timreichen: @timonson and I had a discussion about that. @timonson can you expand on that?

std/jwt/algorithm.ts
```diff
@@ -0,0 +1,15 @@
+/*
```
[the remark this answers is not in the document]
timreichen: @kitsonk  How about renaming `algorithm.ts` to `_algorithm.ts`? As well as `signature.ts` to `_ signature.ts `

std/jwt/mod.ts
```diff
@@ -0,0 +1,194 @@
...
+  return Math.round(
+    (exp instanceof Date ? exp.getTime() : Date.now() + exp * 1000) / 1000,
+  );
+}
+
+function tryToParsePayload(input: string) {
```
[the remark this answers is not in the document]
timreichen: linter doesn't like `any` but this works with `unknown` as well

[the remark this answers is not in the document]
kitsonk: The linter has pragmas that allow it.

std/jwt/mod.ts
timonson: We didn't have `isTokenObject` from the beginning and we wanted to separate `verify` from `decode` so that people could use `decode` in cases where the 'no `crit` header' error is thrown; in the sense of a *helper function*. I can see that the separation now is not really necessary anymore but at the same time both functions fulfill one clearly defined task respectively. So it doesn't hurt to keep them, I guess, but @kitsonk is right, it is not necessary to separate them anymore.

timonson: By the way, I just made changes to `isTokenObject`. I would love to get feedback on that.

std/jwt/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+Takes a `payload`, `key` and `header` and returns the url-safe encoded `jwt`.
+
+```typescript
+import { create } from "https://deno.land/std/jwt/mod.ts";
+
+const payload = { foo: "bar" };
+const key = "secret";
+
+const jwt = await create(payload, key); // eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJmb28iOiJiYXIifQ.4i-Q1Y0oDZunLgaorkqbYNcNfn5CgdF49UvJ7dUQ4GVTQvpsMLHABkZBWp9sghy3qVOsec6hOcu4RnbFkS30zQ
```
axetroy, line 17: is there any reason for using the async function?

This module does not seem to use the wasm module, there is no need to use async function.

axetroy: > Although not necessary now, verify and create return promises because we
expect coming encryption functions to return promises in the future. E.g. for
RSA algorithms.

good to see.

std/jwt/mod.ts
timonson: @kitsonk we moved `isTokenObject` to `decode` now. What do you think?

ssslambda:
I have suggestion, rename `create()` to `sign()` to match [jsonwebtoken package on npm](https://npmjs.com/package/jsonwebtoken)

std/jwt/mod.ts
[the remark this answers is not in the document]
ssslambda: I think it's better to remove this method, cause it really looks useless.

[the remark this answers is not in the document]
timonson: It is definitely not useless. Everybody who uses `exp`, what are a lot of people, would have to write something similar to this function. Most people don't want to read through every detail in the spec just to find the part about the `alg` claim.

timonson:
> I have suggestion, rename `create()` to `sign()` to match [jsonwebtoken package on npm](https://npmjs.com/package/jsonwebtoken)

We don't necessary `sign` jwts. It would be misleading in my opinion.

std/jwt/mod.ts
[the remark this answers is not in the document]
ssslambda: Okay, it is not useless, but it is not part of JWT itself

timreichen:
@ssslambda @timonson tbh I always though `sign` was a weird name choice in npm `jsonwebtoken`. `create` or `encode` seems right to me.

std/jwt/mod.ts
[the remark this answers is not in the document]
timreichen: That is a good point. @timonson maybe we should drop that helper function, it is just one line of code.

timonson:
Yeah it is definitely not `sign`. I would like to keep `create`. We literally **create** a `jwt`. We don't sign and we don't encode one.

std/jwt/mod.ts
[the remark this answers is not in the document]
timonson: Fair enough! I will remove `createExpiration` and add one example to the README instead.

timreichen:
@kitsonk ready for a second review.

kitsonk (approved):
LGTM, thanks for addressing all the feedback!!!

panva:
I think maybe this should be reverted - or at the very least further worked on until being released as std.

1) Web Crypto API should be used for crypto operations and its CryptoKey representation should be used for secrets/keys. Not strings for HMAC-based JWT algorithms. Since you're starting with strings where do you move on from there once RSASSA, ECDSA, EdDSA comes in? Developers will start passing PEM keys like in Node.js and you'll be using them as HMAC secrets verbatim. Bad idea.

2) The module claims to be JWT, but seems to be doing [`tryToParsePayload`](https://github.com/denoland/deno/pull/7991/files#diff-5ebdaca33168e080a16d768935e166cb94dc75965c26895168ff30150515a93aR84) and falling back onto a string payload. JWT's payload MUST always be a JSON top-level object. So, is it JWT or JWS? Why mix the two if the developer cannot instruct the lib what he expects to receive.

3) `nbf` validation is missing, same as with `exp` not allowing the token to be used **anymore** when expired, `nbf` doesn't allow it to be used because it's not valid **yet**.

4) `alg=none` should really not be supported. It even seems like the [API allows not specifying](https://github.com/denoland/deno/pull/7991/files#diff-c82fc4f7c9334bdbb6d891bfdecef10f0ac54b754959d4583c3743d82d90a2a9R39) an algorithm in which case - will a "none" token be accepted? Hope not - given that i see HS512 used as the default for the verify algorithm option, but then again - when RSASSA / ECDSA is added, such default is just awful, either use no defaults (and remove "none") or derive the appropriate algorithm from the key material.

I'm rewriting my [jose](https://github.com/panva/jose) module into a new universal ESM module with TS (wip [here](https://github.com/panva/jose-universal)), so once Web Crypto API lands in Deno, it'll be long ready and released. This new universal module (and the current node-specific version too) comes with almost all of JOSE (as long as supported by Web Crypto API), not just JWT, but also JWS, JWE, JWK, Flattened and General serializations, etc...

timonson:
1. Deno doesn't have Web Crypto and only HS256 is required by spec.
2. We mention in the readme and the code that we use *The JWS Compact Serialization*.
3. Nobody said this library has been completed yet. The Use of `nbf` claim is optional. Somebody will add it if he needs it.
4. You are totally wrong about everything in your first two sentences here.
5. Now you are arguing about the name of a function, which by the way doesn't have *numerous occurences*?

A lot of words but no substance.

panva:
I understand your defensive response, being the proud author of the module, I also want to reiterate, because this can never be said enough - thank you for being a fellow open source maintainer and I mean no disrespect or to offend here. 

1) i think it's clear that i am aware webcrypto is not supported from my comment. The point that i failed to make clear is that string type for a secret or key argument is insufficient. A key-representing object with a clear type (oct/RSA/EC/OKP) removes any ambiguity about what a provided key argument is for. 

2) without giving devs the affordance to say "i expect a JWT" vs. "i expect a JWS" this can cause unexpected behaviours in their code. Since we're talking about security tokens here i'd say that should be wanted. 

3) standards. Am i right? Well, the nbf claim may be optional to use - meaning it doesn't have to part of the payload, but that doesn't mean the library responsible for verifying these claims is free to ignore it when actually present. Suggesting to use a token after being expired, using "decode" that ignores signatures to do it... i'll pass.

kitsonk:
@panva thank you for your feedback... we will discuss internally and figure out what to do.

panva:
@kitsonk thank you for considering my feedback valuable. Good luck, stay safe.

timonson:
> i checked again - The docs really suggest passing the algorithm option is optional, and it really does default to HS512. So where did I lie? The point was again - hardcoding algs without considering the key input leads to mistakes and vulnerabilities.

Because the use of `algorithm` is not optional in this library. Please check the types.

panva:
> Because the use of `algorithm` is not optional in this library. Please check the types.

I checked the readme docs. Entirely my fault clearly. Who reads docs anymore. 

Anyway, @kitsonk if you wish to discuss around some of the more high level points, please let me know and i can join discord/slack/whatever.

timreichen:
Hei @panva Thank you for your concern raising. @timonson and I will work on a patch.
Point 2 and 3 will be fixed.

What did you mean exactly by
> when RSASSA / ECDSA is added, such default is just awful, either use no defaults (and remove "none") or derive the appropriate algorithm from the key material.

? The spec demands the implementation of `none` as an *algorithm *, so that cannot be removed without spec deviation.

panva:
@timreichen

The key material type passed in should be responsible for what algorithm is "allowed". When Uint8Array is passed in as key - only HMAC algs can be used, when ECDSA key object is passed in there's always only one algorithm (mapped per EC key curve). When RSA key object key is passed in - only RSASSA-PKCS1-v1_5 or RSASSA-PSS can be used. The list of algorithms can then be further limited by providing the algorithms option, but the algorithms should never stray from what a given key is capable of verifying/signing. Why HS512 is a bad default and string is a bad idea? Because someone eventually will pass in a PEM public key thinking they're verifying e.g. PS256 but its the octets of the key used for HMAC instead. Hence - key object representation. Uint8Array for secrets, CryptoKey (when available) for everything else.

Since it's currently only in scope to do HMAC - you only have one key type to worry about, but please, don't make string a valid input, have developers pass in a Uint8Array instance. (note: getting Uint8Array out of a string is as simple as `new TextEncoder().encode(str)`) so this isn't any burden for devs.

Once Web Crypto API is supported by Deno, you'll be have these public/private key objects native to your runtime.

As far as "none" goes. If you feel like you MUST support it to be spec conform, then make "none" a special kind of key object representation. E.g. a singleton like i've done [here](https://github.com/panva/jose/blob/master/docs/README.md#jwknone). Only when this key is passed in can "none" JWTs be "verified" or "signed".

I've been dealing with JOSE for a very long time so this feedback / these suggestions come from a great deal of experience, especially in the node/browser JS ecosystems. Mixing "none" with the rest of the secure API is inherently a bad idea, going with a special "none" key object is the least amount of work (and goes in line with the general idea of "key type drives the algorithms", alternatively - provide a completely different API just for "none" alg validation.

As an aside: to be 100% conform for HMAC algorithms you also must ensure that the key is at least the bitsize of the HMAC algorithm blocksize - so >=256 bits for HS256, >=384 bits for HS384, >=512 bits for HS512.

sanderhahn:
There are WebAPIs on which a more complete JWT could be build once available.

https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API
https://developer.mozilla.org/en-US/docs/Web/API/CryptoKey
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign

sanderhahn:
Maybe this can be useful:

https://kjur.github.io/jsrsasign/
https://kjur.github.io/jsrsasign/tool/tool_jwt.html

timreichen:
@panva 
Thank you for your detailed response.
- I think passing Uint8Array instead of a string is reasonable.
- std/jwt should follow the spec and therefore support `none`. What makes a passed singleton more secure than a passed string?
- shouldn't the HMAC implementation validate the blocksize instead of jwt?

panva:
> std/jwt should follow the spec and therefore support none. What makes a passed singleton more secure than a passed string?

There are two sides to this - one - the key passed in, two - the algorithms option. Unless you want to be dealing with edge cases where Uint8Array key is passed in but algorithms option lists "none", go with a key object instead.

> shouldn't the HMAC implementation validate the blocksize instead of jwt?

It doesn't, usually when keys are shorter they get padded to the required size, when they're longer, they're automatically hashed by the HMAC implementation.

panva:
With the release of Deno v1.18 https://deno.land/x/jose can now be considered stable and safe to use.

PR #7992, opened 2020-10-15
[WIP] refactor(core): better resource handling

bartlomieju:


bartlomieju:
Probably not the best approach, I need to prototype more before this PR is ready.

Issue #7326, opened 2020-09-02
Use isolatedModules


ry:
This is going to break a lot of code, however its important for sanity when running with `--no-check` (type stripping).

I propose we do this in a minor release like 1.5.0. Deno is still very much under development and the semver only applies to the Deno namespace and only without the `--unstable` flag.

nayeemrmn:
There's one annoying behaviour of `isolatedModules` which doesn't make sense for Deno to exhibit:

```ts
console.log("Hello, world!");
```
```
error: TS1208 [ERROR]: All files must be modules when the '--isolatedModules' flag is provided.
```

So any toy module which doesn't use `import`/`export` needs `export {};` to disambiguate from traditional scripts, even though Deno doesn't run those.

Ideally, we would wait for another option to make the compiler assume every source is a module and enable that at the same time (https://github.com/microsoft/TypeScript/issues/27535). Maybe this will be the behaviour of `--isolatedModules` itself (https://github.com/microsoft/TypeScript/issues/18232#issuecomment-380104366)!. This is a blocker IMO. If we're in a hurry, I'm also in favour of modifying `tsc` ourselves.

kitsonk:
I really don't want to get into modifying tsc directly.

Ultimately it is a valid "problem" with the module specification, that you can't disambiguate scripts and modules.  There is a proposal to solve this: https://github.com/tc39/proposal-modules-pragma though it doesn't have a lot of traction and movement.

I think realistically it is something that will be encountered rarely with Deno.  I suspect we can ignored TS1208 like we do with other diagnostics with no ill effect.  If we don't want to ignore it, we should rewrite it to indicate that they might want to use `export {}` to ensure wider compatibility.

nayeemrmn:
> Ultimately it is a valid "problem" with the module specification, that you can't disambiguate scripts and modules.

That problem isn't applicable to us. The host environment should already be able to provide this information out-of-band. It's just TS not allowing us to supply that.

> I suspect we can ignored TS1208 like we do with other diagnostics with no ill effect.

That would be great (that's along the lines of what I meant to modify).

lucacasonato:
Status update: 1.4.0 is going to enable this by default in `--unstable` with override in tsconfig being possible.

nayeemrmn:
It's looking like we'll need to enable `preserveConstEnums` along with `isolatedModules`.

[...]

PR #7996, opened 2020-10-16
refactor(cli): migrate run and cache to new infrastructure

kitsonk:
Ref: #7225 

Still a work in progress, the biggest challenge is with dynamic imports.  I need to do some thinking about how best to integrate these in a way that is clean in the new infrastructure.

There is still a bit of cleanup and some more unit tests to write, but thought it was good enough.  It is a ~420 LOC reduction at the moment, but there is certainly more that will be done when bundling can the runtime APIs move over.


bartlomieju:
> Still a work in progress, the biggest challenge is with dynamic imports. I need to do some thinking about how best to integrate these in a way that is clean in the new infrastructure.

Can you elaborate what's the hurdle with dynamic imports?

kitsonk:
> Can you elaborate what's the hurdle with dynamic imports?

Because the module graph is the centre of the universe, it means that it tries to go get dynamic imports eagerly, when swc sees them, also, instead of waiting until they are demanded by the runtime.  tsc also tries to use them in type checking, so we want to be able to eagerly fetch them.  Which means we encounter all the errors there, during the building of the graph, instead when v8 attempts to loads them.  This lead to errors not being catchable.  I have fixed it now in my branch, basically making dynamic import resolutions in the graph non-errors, but as I said in our chat, I broke a few other things, but I think I know what happened there and am fixing that now.

cli/module_graph2.rs
```diff
@@ -566,24 +668,150 @@ impl Graph2 {
...
+      options.emit, options.lib, options.maybe_config_path, options.reload
+    );
+    let mut config = TsConfig::new(json!({
+      "allowJs": true,
+      // TODO(@kitsonk) is this really needed?
+      "esModuleInterop": true,
+      // Enabled by default to align to transpile/swc defaults
+      "experimentalDecorators": true,
+      "incremental": true,
+      "isolatedModules": true,
```
kitsonk, line 688: Note, this is a breaking change that was planned for 1.5... See: #7326

kitsonk:
I am pretty happy with it now.

It does include the breaking change that `isolatedModules` are on by default, which has only been the case under `--unstable` since 1.4.

I am going back through issues that we said we wanted to address with this re-write and writing some test for them.  At least one of them is not fixed and I am looking into what I can do to fix it at the moment.

cli/module_graph2.rs
ry: Thanks for pointing this out - Can we do this in a separate PR?

cli/module_graph2.rs
```diff
@@ -484,10 +584,11 @@ impl Graph2 {
   pub fn new(handler: Rc<RefCell<dyn SpecifierHandler>>) -> Self {
     Graph2 {
       handler,
-      maybe_ts_build_info: None,
+      maybe_tsbuildinfo: None,
```
ry, line 587: <3 grep

kitsonk: done...

cli/module_graph2.rs
kitsonk: Ok, taken out and will follow it up in a seperate PR.

cli/specifier_handler.rs
```diff
@@ -173,12 +207,14 @@ pub struct FetchHandler {
...
   file_fetcher: SourceFileFetcher,
   permissions: Permissions,
+  permissions_dynamic: Permissions,
 }
 
 impl FetchHandler {
   pub fn new(
     program_state: &Arc<ProgramState>,
     permissions: Permissions,
+    permissions_dynamic: Permissions,
```
bartlomieju, line 217: @kitsonk I'm lost here - why do we need two different sets of permissions passes to `FetchHandler`?

cli/module_graph2.rs
```diff
@@ -336,14 +359,30 @@ impl Module {
...
+          Ok(specifier) => Some(specifier),
+          Err(any_error) => {
+            match any_error.downcast_ref::<ModuleResolutionError>() {
+              Some(ModuleResolutionError::ImportPrefixMissing(_, _)) => None,
+              _ => {
+                return Err(any_error);
+              }
+            }
+          }
+        };
```
bartlomieju, line 379: I think this bit won't return error for either dynamic or static imports - is that desired? Maybe error could be ignored only for dynamic imports?

cli/module_graph2.rs
```diff
@@ -469,10 +568,11 @@ pub struct TranspileOptions {
 #[derive(Debug)]
 pub struct Graph2 {
   handler: Rc<RefCell<dyn SpecifierHandler>>,
-  maybe_ts_build_info: Option<String>,
+  maybe_tsbuildinfo: Option<String>,
   modules: HashMap<ModuleSpecifier, Module>,
   redirects: HashMap<ModuleSpecifier, ModuleSpecifier>,
   roots: Vec<ModuleSpecifier>,
+  roots_dynamic: bool,
```
bartlomieju, line 575: What does this field do? Could you add a comment?

cli/specifier_handler.rs
kitsonk: Because, the module_graph doesn't always wait to fetch dynamic imports when requested by deno_core.  It actually tries to fetch them ahead of time, but does not error if they are missing.  So in the `FetchHandler` we have to ensure we are passing the right set of permissions to the `fetch_handler` to avoid permission escapes when importing dynamic permissions.

In the old world the program_state talked directly to the file_fetcher and would supply the "right" permissions when supplying a dynamic versus static import.

Having said that out loud though, I realise that `program_state` has the right dynamic permissions on it already though, so I should/could use that.

Does that make any more sense?

cli/program_state.rs
```diff
@@ -115,89 +119,69 @@ impl ProgramState {
...
-    is_dyn_import: bool,
+    is_dynamic: bool,
     maybe_import_map: Option<ImportMap>,
   ) -> Result<(), AnyError> {
-    let module_specifier = module_specifier.clone();
+    let specifier = specifier.clone();
+    let handler = Rc::new(RefCell::new(FetchHandler::new(
+      self,
+      permissions.clone(),
+      self.permissions.clone(),
```
bartlomieju, line 132: This bit looks wrong to me - main permissions will be forwarded from the runtime while dynamic imports will use  initial permissions (which might have been partially revoked). And again, why do we need to different permissions passed to `FetchHandler`?

I've got this problem in the back of my head to remove `ProgramState::permissions` but waiting on these compiler refactors to be finished.

cli/program_state.rs
```diff
@@ -115,89 +119,69 @@ impl ProgramState {
...
+    )?));
+    let mut builder = GraphBuilder2::new(handler, maybe_import_map);
+    builder.insert(&specifier, is_dynamic).await?;
```
bartlomieju, line 135: Nitpick: async `insert` method looks strange, maybe we can come up with a better name? `construct()`?

cli/program_state.rs
```diff
@@ -115,89 +119,69 @@ impl ProgramState {
...
+            TypeLib::UnstableDenoWorker
+          } else {
+            TypeLib::DenoWorker
+          }
+        }
+      };
+      let (stats, diagnostics, maybe_ignored_options) =
+        graph.check(CheckOptions {
+          debug,
+          emit: true,
```
bartlomieju, line 171: Just to be sure `emit: true` is temporary and we'll remove emit from TSC in the future, right?

cli/program_state.rs
```diff
@@ -279,261 +289,66 @@ impl ProgramState {
...
-          })
-          .expect("Failed to find imported file");
-        let media_type = import_file.media_type;
-        media_type == MediaType::TypeScript
-          || media_type == MediaType::TSX
-          || media_type == MediaType::JSX
-      })
+// TODO(@kitsonk) this is only temporary, but should be refactored to somewhere
+// else, like a refactored file_fetcher.
+impl SourceMapGetter for ProgramState {
```
bartlomieju, line 294: Glad to see `SourceMapGetter` moved from `tsc.rs` - I do believe it should be implemented on `ProgramState`

cli/specifier_handler.rs
```diff
@@ -25,8 +26,30 @@ pub type DependencyMap = HashMap<String, Dependency>;
...
   Pin<Box<(dyn Future<Output = Result<CachedModule, AnyError>> + 'static)>>;
 
+/// A group of errors that represent errors that can occur with an
+/// an implementation of `SpecifierHandler`.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum HandlerError {
+  /// A fetch error, where we have a location associated with it.
+  FetchErrorWithLocation(String, Location),
+}
+use HandlerError::*;
```
bartlomieju, line 36: FYI I'm not a fan of doing that - it pollutes whole module with all the variants - if it's only required for `fmt::Display` I suggest moving it into `fn fmt()`

cli/tests/disallow_http_from_https_ts.out
```diff
@@ -1,2 +1,3 @@
-error: Modules loaded over https:// are not allowed to import modules over http://
-Imported from "https://localhost:5545/cli/tests/disallow_http_from_https.ts:2"
+error: Modules imported via https are not allowed to import http modules.
+  Importing: http://localhost:4545/cli/tests/001_hello.js
+    at https://localhost:5545/cli/tests/disallow_http_from_https.ts:2:0
```
bartlomieju, line 3: Very nice upgrade 👍

cli/tests/error_015_dynamic_import_permissions.out
```diff
@@ -1 +1 @@
-error: Uncaught TypeError: network access to "http://localhost:4545/cli/tests/subdir/mod4.js", run again with the --allow-net flag
+error: network access to "http://localhost:4545/cli/tests/subdir/mod4.js", run again with the --allow-net flag
```
bartlomieju, line 1: `PermissionsDenied` should be shown here - without it this message is incomplete

cli/tests/error_local_static_import_from_remote.js.out
```diff
@@ -1,3 +1,4 @@
 [WILDCARD]
-error: Remote modules are not allowed to statically import local modules. Use dynamic import instead.
-Imported from "[WILDCARD]error_local_static_import_from_remote.js:1"
+error: Remote modules are not allowed to import local modules.
+  Importing: file:///some/dir/file.js
```
bartlomieju, line 3: Sad to see `Use dynamic import instead` hint go away

cli/module_graph2.rs
kitsonk: It is desired because if I don't, then transient dynamic imports become non-catchable. (Part of error_014 test cases).

bartlomieju: Let me rephrase it: will error be returned for errors in static imports?

cli/module_graph2.rs
nayeemrmn: I'll reopen #8050.

cli/module_graph2.rs
kitsonk: Yes, but not by the graph or prepare module, but instead the load module.  There is an integration test that tests for it, and it passes.

cli/program_state.rs
kitsonk: But it doesn't construct... it inserts as module in the graph... `.add`?

cli/program_state.rs
kitsonk: If we are happy with the output of `tsc` and `isolatedModules`.

cli/program_state.rs
kitsonk: Well, I think it should be in what ever file_fetcher becomes... whatever "owns" caching the emitted source files/source maps.  But obviously something for later.

cli/specifier_handler.rs
kitsonk: It is only for convenience...  I will remove it.

cli/tests/error_015_dynamic_import_permissions.out
kitsonk: I am not sure what you are referring to... with or without this PR there is no `PermissionsDenied` text... if we want that, then it should be a subsequent PR.

cli/tests/error_local_static_import_from_remote.js.out
kitsonk: I will fix that.

cli/module_graph2.rs
kitsonk: Specifically it deals with the situation where a dependency has already been loaded and compiled during a run, but then is subsequently loaded as a dynamic import when the `--reload` flag is passed... If in that case the dynamic import has a valid emit, then it will not be reloaded.

As I mentioned in another conversation, I don't think our current behaviour around `--reload` makes sense, when we can be 100% confident that emit will be reproduced now, but we have integration tests that expect that `--reload` will compile a module, unless it has already been compiled (see: `single_compile_with_reload`).

I will add a comment.

cli/program_state.rs
bartlomieju: Let's leave it as is for now...

cli/specifier_handler.rs
bartlomieju: Yes, it does make more sense I think it directly refers to https://github.com/denoland/deno/pull/7996#discussion_r509233924

cli/tests/error_015_dynamic_import_permissions.out
bartlomieju: You're right, sorry for noise.

cli/module_graph2.rs
```diff
@@ -1011,39 +1325,38 @@ impl GraphBuilder2 {
...
   /// Move out the graph from the builder to be utilized further.  An optional
   /// lockfile can be provided, where if the sources in the graph do not match
-  /// the expected lockfile, the method with error instead of returning the
-  /// graph.
+  /// the expected lockfile, an error will be logged and the process will exit.
   ///
   /// TODO(@kitsonk) this should really be owned by the graph, but currently
   /// the lockfile is behind a mutex in program_state, which makes it really
   /// hard to not pass around as a reference, which if the Graph owned it, it
   /// would need lifetime parameters and lifetime parameters are 😭
```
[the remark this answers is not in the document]
kitsonk: We should do that as a seperate PR.  Added issue #8065.

cli/module_loader.rs
```diff
@@ -113,44 +112,34 @@ impl ModuleLoader for CliModuleLoader {
...
-    } else {
-      None
-    };
-
     // TODO(bartlomieju): `prepare_module_load` should take `load_id` param
     async move {
       program_state
         .prepare_module_load(
-          module_specifier,
-          maybe_referrer,
```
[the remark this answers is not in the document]
bartlomieju: So in case of dynamic import `maybe_referrer` will have value - if trying to load `module_specifier` results in eg. `404` we can display the error with information that this failed import originated from `maybe_referrer`.

cli/module_graph2.rs
kitsonk: I've also opened #8066

cli/module_loader.rs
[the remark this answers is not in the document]
kitsonk: But then the module graph would have to have the concept of referrer on add... which doesn't make sense... As mentioned before, the module graph silently squashes these error and allows them to be done on the `fetch_compiled_module` and why `error_014_catch_dynamic_import_error` passes (which includes the referrer).

cli/program_state.rs
kitsonk: Ok, I have looked at this... There is a lot of indirection in there that we should consider refactoring out.

- deno_core calls `prepare_load()` on the `ModuleLoader`
- `CliModuleLoader` implements `ModuleLoader`, and when `prepare_load()` is called it determines if the "main" module is being called, or if it is a dynamic import.
  - If it is the main module is passes `Permissions::all()` to `program_state.prepare_module_load()`,
  - Otherwise it clones a version of the permissions from `op_state`.
- The `program_state.permissions()` aren't used at all in figuring out dependencies as far as I can tell.

As stated above, the new module graph tries to make dynamic imports part of the graph, which gives a more complete type check when type checking modules, as well as an efficiency benefit of pre-fetching as many of the dynamic imports as possible.  Because of these seperate permissions being used in the new module graph, all the permission tests are passing.

I think I understand enough now to clean it up a bit better, and will take a look at the whole thing tomorrow and try to get it a bit more clean, since really it should be the module graph, being the centre of the universe, that mediates if it is using the "all" permissions, or the current runtime state permissions, and ideally not a concern of the module_loader or program_state.

bartlomieju: @kitsonk I agree that `ProgramState::permissions` should be removed. My only problem now is that `self.permissions.clone()` is the wrong one to pass to module graph. We can clean this up in a follow up PR, but in this case permissions from `OpState` should be passed. After that I can approve the PR

cli/module_graph2.rs
```diff
@@ -1005,45 +1389,17 @@ impl GraphBuilder2 {
...
-    }
-
-    Ok(())
-  }
-
   /// Move out the graph from the builder to be utilized further.  An optional
   /// lockfile can be provided, where if the sources in the graph do not match
-  /// the expected lockfile, the method with error instead of returning the
-  /// graph.
+  /// the expected lockfile, an error will be logged and the process will exit.
```
ry, line 1394: What's the reasoning behind making `get_graph` infallible? Seem like the `Result` return value that was here before would be a better separation of concerns.

cli/file_fetcher.rs
```diff
@@ -579,6 +579,29 @@ fn map_js_like_extension(path: &Path, default: MediaType) -> MediaType {
...
+                }
+              } else {
+                default
+              }
+            }
+          }
+        } else {
+          default
+        }
+      }
```
ry, line 604: I see we don't have any unit tests in this module for `map_js_like_extension()` - which is unfortunate.  Would you mind adding a new unit test to nail down the logic being added here?

cli/source_maps.rs
```diff
@@ -6,8 +6,9 @@ use deno_core::error::JsError as CoreJsError;
 use sourcemap::SourceMap;
 use std::collections::HashMap;
 use std::str;
+use std::sync::Arc;
 
-pub trait SourceMapGetter {
+pub trait SourceMapGetter: Sync + Send {
```
ry, line 11: Why does it need to be Send + Sync?

cli/specifier_handler.rs
```diff
@@ -99,7 +99,7 @@ impl Default for Emit {
   }
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct Dependency {
   /// Flags if the dependency is a dynamic import or not.  This will be set to
   /// `true` if it is, otherwise `false`.
```
ry, line 105: "This will be set to `true` if it is, otherwise `false`."

superfluous

cli/specifier_handler.rs
```diff
@@ -114,6 +114,17 @@ pub struct Dependency {
   pub maybe_type: Option<ModuleSpecifier>,
 }
```
ry, line 115: Is it ever the case that `maybe_code.is_some() && maybe_type.is_some()` ? What about the opposite - is it possible that `maybe_code.is_none() && maybe_type.is_none()` ?

If the answer to those questions is no, then consider refactoring this type to not allow for those situations. It might be something like this:
```rust
struct Dependency {
  location: Location,
  is_dynamic: bool,
  kind: DependencyKind,
  module_specifier: ModuleSpecifier,
}

enum DependencyKind { Code, Type };
```

cli/file_fetcher.rs
kitsonk: Yeah, it would be nicer to have them seperate... but I think file_fetcher needs a good hose down anyways.

cli/source_maps.rs
kitsonk: Because when I implemented it on `ProgramState`, `ProgramState` gets passed around as an `Arc` and therefore the trait needed to be `Sync + Send` to be able to be implementable in a thread safe way.

cli/specifier_handler.rs
kitsonk: `maybe_code.is_some()` && `maybe_type.is_some()` occurs quite frequently when you have `@deno-types` or `X-TypeScript-Types` or `/// <reference types="" />`.  Modelling this properly is one of the key improvements in the new module graph, so we can always ensure we have a single dependency structure and not get weird edge cases.

`maybe_code.is_none()` && `maybe_type.is_none()` only occurs in situations situations where dynamic imports potentially have runtime issues, so we see the dependency in the graph, but we "defer" erroring until the module is expressly attempted to be loaded by `deno_core`.

cli/module_graph2.rs
kitsonk: A couple reasons... `get_graph()` should be infallible in my opinion... it should externally be a simple "move" out.

Also, my opinion is locking is an internal concern.  It should be terminal and not left up to the consumer of the API to figure that out.  Because the lockfile was asserted, it is meant to be strictly enforced.  Bubbling up a result means that the consumer could ignore a failure and continue, when we know objectively something is wrong with the graph.

Lastly, we exit with code 10.  If we moved that out of here, then there would be about 5-6 places where we would have to downcast and any error to a lock error, to be able to be certain it was a lock error, and then branch to exit the process with an error of 10.

I had originally made in fallible and it was starting to get messy and difficult.

kitsonk:
@bartlomieju I think I have addressed all your feedback... PTAL

cli/program_state.rs
kitsonk: Ok, to avoid any confusion, I renamed what gets passed in from the loader as `dynamic_permissions` and put several comments in the code.  Feels a lot clearer now.

bartlomieju: Thank you

cli/specifier_handler.rs
```diff
@@ -187,23 +233,54 @@ impl FetchHandler {
...
+    is_dynamic: bool,
+  ) -> FetchFuture {
+    // When the module graph fetches dynamic modules, the set of dynamic
+    // permissions need to be applied.  Other static imports have all
+    // permissions.
+    let permissions = if is_dynamic {
+      self.dynamic_permissions.clone()
+    } else {
+      Permissions::allow_all()
+    };
```
bartlomieju, line 256: This is indeed cleaner. We could bike shed on the naming, but I'm not gonna do that right now - although it might be a misleading in case of creates module graph for worker - it is another top level module (not dynamic), but it should use runtime permissions - and I think that's already happening.

bartlomieju (approved):
LGTM, let's land it 🚀

cli/specifier_handler.rs
kitsonk: Yeah, but not in the module graph...  There is still a bit of work to do to try consolidate where we have logic like this.

Issue #7246, opened 2020-08-29
Handle recoverable errors from SWC


bartlomieju:
Much like TSC, SWC can recover from number of errors during parsing. Currently we treat any error as terminal and don't proceed to work on AST. 

`Parser.take_errors()` can be used to obtain recoverable errors. 

This issue can be treated as an umbrella issue for dependency analysis, `deno doc` and `deno lint`.

CC @kdy1

bartlomieju:
After conversation with @kdy1 it turns out it's not really needed.

bartlomieju:
Actually after another conversation with @kdy1 we should look into this problem. 

All parsing is done in `cli/ast.rs` in either `parse()` and `transpile()` functions. Currently when encountering error we always emit it using `DiagnosticBuffer` helper struct which converts SWC diagnostics into string errors. Due to this conversion we lose structured data from SWC's diagnostics which could potentially could be used to generate more descriptive errors. Additionally this conversion causes hard errors and terminates the execution and we never try to recover (recovery might not make sense in `deno` itself but it would be helpful in `deno_lint` which uses the same infrastructure).

@kdy1 can you look into this issue?

kdy1:
It's okay to exit with failure on parsing error in `deno` because the parser does not do any linting - all errors are critical issue for compiler and runtime.
I mean, if runtime implements ecmascript specification correctly, it will be hard error anyway.


But as `swc` has rich diagnostic system (it's copied from rustc source code), I think converting diagnostics to string is not a good idea. It can have some annotations, suggestions, sub-diagnostics and / or multiple spans. Although errors from the parser is not, the error reporting system itself is rustc-grade.

bartlomieju:
@kdy1 then I guess it makes more sense to handle this problem in `deno_lint` first. I'm going to transfer the issue to `deno_lint` repo. 

It also seems to to me that in case of linter we could transform SWC diagnostic in `LintDiagnostic`.

PR #7998, opened 2020-10-16
[WIP] handle SWC recoverable errors

bartlomieju:
Closes https://github.com/denoland/deno/issues/7246

Need to come up with some test before landing

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -124,6 +124,10 @@ impl ErrorBuffer {
   pub fn new() -> Self {
     Self(Arc::new(RwLock::new(Vec::new())))
   }
+
+  pub fn has_error(&self) -> bool {
+    !self.0.read().unwrap().is_empty()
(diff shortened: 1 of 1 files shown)


bartlomieju:
After conversation with @kdy1 it turns out it's not really needed.

kdy1:
Side note:

There are two kinds of errors.

The returned `Err` (from parser) is unrecoverable error, and it means the parser gave up parsing because it met something bad.
Errors from `take_errors` is recoverable (in the point of view of parser, not runtime), and such errors means the parser recovered from error.

In future, more errors will be recovered, mainly to mimic official typescript compiler. 
Recovering errors allows running other validations on ast nodes, so the compiler can show more errors at once.

bartlomieju:
@kdy1 thanks for detailed explanation, I'll revisit the topic after SWC updates its recovery

PR #7999, opened 2020-10-16
fix(docs): readTextFile / readTextFileSync throws when reading directory

vwkd:
I believe this description was erroneous. Correct me if I'm wrong.


lucacasonato (approved):
LGTM. Thanks @vwkd

PR #8001, opened 2020-10-16
feat(std/testing): Add support for object assertion against object subset 

lowlighter:
This add supports for a new assertion function `assertObjectMatch` which allows to test an `actual` object against an `expected` object subset (i.e. inclusivity, not equality)

**Why ?**

It's often useful to test an object partly without testing the remaining properties (like for web api response, where you might want to only check a few fields).

This function filters out `actual` object with `expected` object subsets keys before calling `assertEquals` (so it's actually a wrapper which acts as a syntaxic sugar).

```ts
const a  = {
   foo: true,
   bar: true,
   baz: {
     qux: true
   }
}

// Simple assertions
assertObjectMatch(a, {foo: true})  // OK 
assertObjectMatch(a, {bar: true})  // OK
assertObjectMatch(a, {foo: false}) // KO 

// Nested assertions
assertObjectMatch(a, {baz: {qux: true}}) // OK
assertObjectMatch(a, {baz: {qux:null}})  // KO
```

### DIFF
--- a/std/testing/asserts.ts
+++ b/std/testing/asserts.ts
@@ -434,10 +434,10 @@ export function assertNotMatch(
  * If not, then throw.
  */
 export function assertObjectMatch(
-  actual: { [key: string]: unknown },
-  expected: { [key: string]: unknown },
+  actual: Record<string | symbol, unknown>,
+  expected: Record<string | symbol, unknown>,
 ): void {
-  type loose = { [key: string]: unknown };
+  type loose = Record<string | symbol, unknown>;
   const seen = new WeakMap();
   return assertEquals(
     (function filter(a: loose, b: loose): loose {
--- a/std/testing/asserts.ts
+++ b/std/testing/asserts.ts
@@ -434,12 +434,13 @@ export function assertNotMatch(
  * If not, then throw.
  */
 export function assertObjectMatch(
-  actual: object,
-  expected: object,
+  actual: { [key: string]: unknown },
+  expected: { [key: string]: unknown },
 ): void {
+  type loose = { [key: string]: unknown };
   const seen = new WeakMap();
   return assertEquals(
-    (function filter(a: object, b: object): object {
+    (function filter(a: loose, b: loose): loose {
       // Prevent infinite loop with circular references with same filter
       if ((seen.has(a)) && (seen.get(a) === b)) {
         return a;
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks for the contribution @lowlighter! Is there some prior art for this assertion? 

@nayeemrmn PTAL

std/testing/asserts.ts
```diff
@@ -429,6 +429,44 @@ export function assertNotMatch(
   }
 }
 
+/**
+ * Make an assertion that `actual` object is a subset of `expected` object, deeply.
+ * If not, then throw.
+ */
+export function assertObjectMatch(
+  actual: { [key: string]: unknown },
+  expected: { [key: string]: unknown },
```
nayeemrmn, line 438: Should symbols be allowed?

lowlighter:
> Is there some prior art for this assertion?

Yes, it's heavily inspired from [expect.toMatchObject(object)](https://jestjs.io/docs/en/expect.html#tomatchobjectobject) from [Jest](https://jestjs.io/en) JavaScript testing framework, though implementation may differ a bit. Since a lot of logic has already been done in `assertEquals`, I thought it would easier to filter the actual object with the expected one so it'll behave the same.


> Should symbols be allowed?

Yes, I'll add them and add tests to cover them !

bartlomieju:
> Yes, it's heavily inspired from expect.toMatchObject(object) from Jest JavaScript testing framework, though implementation may differ a bit. Since a lot of logic has already been done in assertEquals, I thought it would easier to filter the actual object with the expected one so it'll behave the same.

@lowlighter thanks, that sounds good to me. I'll wait for @nayeemrmn's review before landing

std/testing/asserts.ts
```diff
@@ -429,6 +429,44 @@ export function assertNotMatch(
...
+      // Prevent infinite loop with circular references with same filter
+      if ((seen.has(a)) && (seen.get(a) === b)) {
+        return a;
+      }
+      seen.set(a, b);
+      // Iterate through keys present in both actual and expected
+      // and filter recursively on object references
+      const filtered = {} as loose;
+      for (
+        const [key, value] of Object.entries(a).filter(([key]) => key in b)
```
nayeemrmn, line 453: I don't think `Object.entries()` includes symbols. This would be shown if the tests included an erroneous case for symbols. You can use `[...Object.getOwnPropertyNames(a), ...Object.getOwnPropertySymbols(a)]` to get a complete list of keys.

lowlighter: I replaced `Object.entries()` with your suggestion and it works like a charm, and changed `string|symbol` by `PropertyKey` which seems more accurate.

I got some trouble with using `symbol` as index signature so I casted them into `string` but not sure it's the cleanest way...

I wanted to add an erroneous case like you suggested, but seems that `assertEquals` does not handle `symbol`s perfectly currently (unless I missed something) : 
```ts
import { assertEquals } from "https://deno.land/std/testing/asserts.ts"
const foo = Symbol("foo")
Deno.test("symbol1", () => assertEquals({[foo]:"bar"}, {[foo]:"bar"}))
Deno.test("symbol2", () => assertEquals({[foo]:"bar"}, {[Symbol("foo")]:"bar"}))
```
```
running 2 tests
test symbol1 ... ok (2ms)
test symbol2 ... ok (1ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (4ms)
```

nayeemrmn (approved):
LGTM

bartlomieju (approved):
LGTM, thanks @lowlighter

PR #8005, opened 2020-10-17
fix(op_crates/web): TextEncoder error message with original input

### LINKED ISSUE #8003: TextDecoder converts string to lower case before throwing error
Bit of background: I had encountered an encoding issue in a webpage with accepting user files at runtime, and had sought to resolve the encoding issue using the browser's native `TextDecoder`. To start, I needed a good list of supported encodings, I checked out MDN and a polyfill GitHub https://github.com/inexorabletash/text-encoding. Combined the lists and ran a loop to see what threw errors, and what didn't.

It happened to be that Deno was already open, so I just ran the loop there, I later contrasted with the browser to get a reasonable view of what I could expect. I had spotted a tiny difference.

Test code:
```ts
new TextDecoder("GBK");
```

> Uncaught RangeError: The encoding label provided ('gbk') is invalid.

Yet, notice that I never used "gbk," the only difference being the case sensitivity.
Chromium shows the error with the original case that was input.

I have no clue whether or not the spec has anything on this, and couldn't test another browser at the moment.

aca:
Instead of modifying input, return error message with original input.
Edited original test case for TextDecoder a little.

Closes #8003

### DIFF
--- a/cli/tests/unit/text_encoding_test.ts
+++ b/cli/tests/unit/text_encoding_test.ts
@@ -104,10 +104,10 @@ unitTest(function textDecoderASCII(): void {
 unitTest(function textDecoderErrorEncoding(): void {
   let didThrow = false;
   try {
-    new TextDecoder("foo");
+    new TextDecoder("Foo");
   } catch (e) {
     didThrow = true;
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM, thanks @aca

Issue #8008, opened 2020-10-17
Constant condition in "std/archive/tar.ts"


bartlomieju:
Upgrading `deno_lint` yielded a fragment of code that doesn't make any sense (line 496):

https://github.com/denoland/deno/blob/46b892ad37df9ba9bed77fb923a1cfe284b208dc/std/archive/tar.ts#L494-L498

It should probably be `if (n == null)`... 

CC @marcosc90

marcosc90:
Ouch. I'll take care of it today. (I've been quite busy last few months).

PR #8010, opened 2020-10-17
fix(std/tar): fix constant condition

marcosc90:
Fixes https://github.com/denoland/deno/issues/8008


bartlomieju (approved):
LGTM, thanks for quick fix @marcosc90!

PR #8011, opened 2020-10-17
docs(tools): add repl section

caspervonb:
This adds a section on the repl with the keybindings that we support out of the box

### DIFF
--- a/docs/tools/repl.md
+++ b/docs/tools/repl.md
@@ -1,0 +1,49 @@
+# Read-eval-print-loop
+
+`deno repl` starts an read-eval-print-loop, which lets you interactively build up
+program state in the global context.
+
+## Keybindings
+
+| Keystroke             | Action                                                                                           |
+| --------------------- | ------------------------------------------------------------------------------------------------ |
+| End                   | Move cursor to end of line                                                                       |
+| Home                  | Move cursor to the beginning of line                                                             |
+| Left                  | Move cursor one character left                                                                   |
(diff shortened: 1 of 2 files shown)


docs/tools/repl.md
```diff
@@ -0,0 +1,49 @@
...
+up program state in the global context.
+
+## Keybindings
+
+| Keystroke             | Action                                                                                           |
+| --------------------- | ------------------------------------------------------------------------------------------------ |
+| End                   | Move cursor to end of line                                                                       |
+| Home                  | Move cursor to the beginning of line                                                             |
+| Left                  | Move cursor one character left                                                                   |
+| Right                 | Move cursor one character right                                                                  |
```
caspervonb, line 13: Duplicated

```suggestion
```

crypt096 (approved):

David-Else:
Nice one!

- You might want to explain or rephrase 'yank' for people who don't use Vim?
- `Interrupt/Cancel edition`: I am not sure what an `edition` is, just me being ignorant?
- Maybe explain what a `meta` key is? https://www.computerhope.com/jargon/m/meta-key.htm, I had not heard the term before

Maybe it is just me not knowing these terms, but new users might also get confused.

bartlomieju (approved):
LGTM

caspervonb:
> Maybe it is just me not knowing these terms, but new users might also get confused.

When it comes to yank/kill etc, terminology tends to derived from emacs/vi terminology but open for revision of course 👍

PR #8014, opened 2020-10-18
fix(cli): use rid getter for stdio

caspervonb:
This changes the rid of Deno.stdin, Deno.stdout, Deno.stderr from a mutable property into a getter to match the rid semantics of Deno.File.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -710,11 +710,11 @@ declare namespace Deno {
   }
 
   /** A handle for `stdin`. */
-  export const stdin: Reader & ReaderSync & Closer & { rid: number };
+  export const stdin: Reader & ReaderSync & Closer & { readonly rid: number };
   /** A handle for `stdout`. */
-  export const stdout: Writer & WriterSync & Closer & { rid: number };
+  export const stdout: Writer & WriterSync & Closer & { readonly rid: number };
   /** A handle for `stderr`. */
-  export const stderr: Writer & WriterSync & Closer & { rid: number };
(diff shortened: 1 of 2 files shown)


lucacasonato (commented):
Can you update the typings to reflect this?

cli/dts/lib.deno.ns.d.ts
```diff
@@ -710,11 +710,11 @@ declare namespace Deno {
...
 
   /** A handle for `stdin`. */
-  export const stdin: Reader & ReaderSync & Closer & { rid: number };
+  export const stdin: Reader & ReaderSync & Closer & { readonly rid: number };
   /** A handle for `stdout`. */
-  export const stdout: Writer & WriterSync & Closer & { rid: number };
+  export const stdout: Writer & WriterSync & Closer & { readonly rid: number };
   /** A handle for `stderr`. */
-  export const stderr: Writer & WriterSync & Closer & { rid: number };
+  export const stderr: Writer & WriterSync & Closer & { readonly rid: number };
```
bartlomieju, line 717: This is breaking change 😬

caspervonb: Technically, yes but I don't think we ever intended for them to actually be mutable.

bartlomieju: @caspervonb that's true, I'm only afraid it's gonna cause type checking havoc, hopefully not many users access `rid` of stdio objects.

lucacasonato: Access should be fine - it should only break if someone tries to update it.

bartlomieju (approved):
LGTM

PR #8015, opened 2020-10-18
refactor(cli/repl): extract is_closing to a function

caspervonb:
This extracts is closing into a function so that it can easily be used as the condition for the loop slimming down the body of the loop.


bartlomieju (approved):
LGTM, nice cleanup

PR #8016, opened 2020-10-18
refactor(lint): show hint for lint errors

bartlomieju:
After upgrading deno_lint to 0.2.4, there are hints in some rules that should be displayed alongside the error.

Screenshot 2020-10-18 at 15 47 26

Ref https://github.com/denoland/deno_lint/pull/411

CC @magurotuna


magurotuna:
Looks nice 👍

ry (approved):
I agree, looks nice

Issue #8018, opened 2020-10-18
copyNWriteAllData creates an "io" directory during tests - it shouldn't


ry:
Any test artifacts should be hidden away in a tmp directory or in some testdata directory. It should not pollute the root project directory.

see std/io/ioutil_test.ts

PR #8019, opened 2020-10-18
test(std/io): use a real tempdir

caspervonb:
This replaces a case of a temp file in the working tree with a tempfile in a real temporary directory avoiding cwpollution of the working directory.

Closes https://github.com/denoland/deno/issues/8018.


ry (approved):
Thank you Casper. LGTM

PR #7903, opened 2020-10-09
How long do builds without cache take?

ry:

PR #8020, opened 2020-10-18
Remove github actions cache

ry:
[Running into issues with cache when trying to upgrade V8.](https://github.com/denoland/deno/runs/1271315857) Based on the analysis in https://github.com/denoland/deno/pull/7903#issuecomment-706252380 we know the cache is not providing much benefit.


piscisaureus (approved):
This papers over a longstanding issue in Rusty_v8: the binary build is not re-downloaded after an upgrade. See https://github.com/denoland/rusty_v8/issues/377.

In that light, I'm not enthousiastic about this patch.

But 🤷🏼‍♂️

Better to have a working deno build than not.

PR #8023, opened 2020-10-19
feat(cli): add support for bundle --no-check

### LINKED ISSUE #6686: Incremental and no check builds for deno bundle
In order to improve the development experience of Deno itself, we should support both incremental and no check builds for the bundles.  In addition we should support `--no-check` for `deno bundle` as well.  In all of these cases we need to emit each module that makes up the bundles individually and then concatenate them inside Deno instead of relying on the TypeScript compiler to do it.  In addition we need to generate valid source maps for these files.  Currently we do not generate source maps for bundles because we don't have an easy way to shift the mappings in the bundle to account for the module loader that is embedded in the bundle.

I am working on a solution to all of these currently.

kitsonk:
Add support for `deno bundle --no-check`.  This uses swc for the bundling (thanks to a lot of hard work and fixes by @kdy1).  There are some substantial differences from `deno bundle` though:

- The output is a single file ESM file that is a "flattened" version of the input sources.  `deno bundle` uses the System module format and provides a custom loader.  While there might be still some defects, we have tried to do some extensive testing to ensure that that the unbundle code behaviour is preserved when bundled.
- Dynamic imports are left "untouched".  This is a significant change from the current behaviour, where sometimes dynamic imports are included in the bundle if they can be statically determined at bundle time.  It is a double edged sword bundling dynamic imports, as the behaviour might or might not be desired, so it is best to just leave them alone.
- `import.meta.url` and `import.meta.main` work in the bundle, where `import.meta.url` will get "hardcoded" at bundle time and `import.meta.main` will be `false` for any dependent modules, but be preserved in the bundle for the main/entry point module.

We are looking at replacing `deno bundle` with this bundling in a subsequent PR.

Fixes #6686

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -3,7 +3,6 @@
 use crate::media_type::MediaType;
 use crate::tsc_config;
 
-use deno_core::error::bail;
 use deno_core::error::AnyError;
 use deno_core::serde_json;
 use deno_core::ModuleSpecifier;
@@ -422,12 +421,13 @@ pub fn transpile_module(
     Some(&comments),
   );
   let mut parser = swc_ecmascript::parser::Parser::new_from(lexer);
+  let sm = cm.clone();
   let module = parser.parse_module().map_err(move |err| {
     let mut diagnostic = err.into_diagnostic(&handler);
     diagnostic.emit();
 
     DiagnosticBuffer::from_error_buffer(error_buffer, |span| {
-      cm.lookup_char_pos(span.lo)
+      sm.lookup_char_pos(span.lo)
     })
   })?;
   // TODO(@kitsonk) DRY-up with ::transpile()
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -401,6 +401,16 @@ pub fn transpile_module(
   emit_options: &EmitOptions,
   cm: Rc<SourceMap>,
 ) -> Result<(Rc<SourceFile>, Module), AnyError> {
+  // TODO(@kitsonk) DRY-up with ::parse()
+  let error_buffer = ErrorBuffer::new();
+  let handler = Handler::with_emitter_and_flags(
+    Box::new(error_buffer.clone()),
+    HandlerFlags {
+      can_emit_warnings: true,
+      dont_buffer_diagnostics: true,
+      ..HandlerFlags::default()
+    },
+  );
   let comments = SingleThreadedComments::default();
   let syntax = get_syntax(media_type);
   let source_file =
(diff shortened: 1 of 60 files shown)


cli/ast.rs
```diff
@@ -1,13 +1,15 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
 use crate::media_type::MediaType;
+use crate::module_graph2::Graph2;
```
bartlomieju, line 4: That makes a circular dep between `module_graph2` and `ast`. Could it be removed?

cli/tsc_config.rs
```diff
@@ -248,13 +249,17 @@ impl TsConfig {
...
+  pub fn as_emit_options(&self) -> Result<ast::EmitOptions, AnyError> {
+    let options: EmitConfigOptions = serde_json::from_value(self.0.clone())?;
+    Ok(ast::EmitOptions {
+      check_js: options.check_js,
+      emit_metadata: options.emit_decorator_metadata,
+      inline_source_map: true,
+      jsx_factory: options.jsx_factory,
+      jsx_fragment_factory: options.jsx_fragment_factory,
+      transform_jsx: options.jsx == "react",
+    })
```
bartlomieju, line 262: Nitpick: `ast::EmitOptions` should implement `From<Value>` instead of using this method

bartlomieju (commented):
Some quick thoughts, I'll give more detailed review in the morning

cli/ast.rs
kitsonk: Let me take a look at it again...  It is how to cleanly impl the bundle loader in `module_graph2` without dragging all of ast in.

cli/ast.rs
```diff
@@ -202,6 +207,21 @@ impl Default for TranspileOptions {
...
+      check_js: options.check_js,
+      emit_metadata: options.emit_decorator_metadata,
+      inline_source_map: true,
+      jsx_factory: options.jsx_factory,
+      jsx_fragment_factory: options.jsx_fragment_factory,
+      transform_jsx: options.jsx == "react",
+    }
+  }
+}
+
```
bartlomieju, line 224: Thank you

cli/ast.rs
```diff
@@ -372,6 +392,82 @@ pub fn parse(
...
+    StringInput::from(&*source_file),
+    Some(&comments),
+  );
+  let mut parser = swc_ecmascript::parser::Parser::new_from(lexer);
+  // TODO(@kitsonk) this seems to be a problem with the type of result
+  // from `.parse_module` that cannot be coerced.
+  let module = match parser.parse_module() {
+    Ok(m) => m,
+    Err(err) => bail!("Parsing failed: {:?}", err),
+  };
```
bartlomieju, line 420: Need to be mapped to `DiagnosticBuffer` like in `parse` function - why isn't `parse` function from this module used in this function?

bartlomieju (approved):
LGTM

cli/ast.rs
kitsonk: Thanks for noticing that, I took a look at it, and it would take a decent amount of refactoring of things to get to the point where we could use `parse()` (or share code between `transpile_module()` and `parse()`), mainly because `parse()` currently doesn't share a source map between invocations (but maybe it should?) where as the bundling example from kdy that this is based on did.  There is the emit side of it too which shares a lot with the other emits...

I made a note of the refactoring opportunities in there, but it isn't totally straight forward, especially without impacting other parts of the code.

PR #8027, opened 2020-10-19
test(cli/repl): add multiline pty input test

caspervonb:
This adds a simple test to ensure that multi-line editing work as expected for the repl in a tty.


caspervonb:
Rolling this into #7896

Issue #8029, opened 2020-10-19
Remove tempFile function from std/io/util.ts


lucacasonato:
There is a `tempFile` function in  https://deno.land/std@0.74.0/io/util.ts. That function shouldn't be in there - it is really a std/fs function.

caspervonb:
Yeah was wondering about this yesterday, is the util module supposed to be exported at all?

PR #8032, opened 2020-10-19
fix(std/io): remove trivial internal util.ts module

caspervonb:
This removes the internal (but poorly named) util.ts module as it only contained some very trivial one liner helpers.

Closes https://github.com/denoland/deno/issues/8029 as it removes the entire module.


caspervonb:
**Technically a breaking change** but this i believe this to be a case of "Casper don't freak everyone out by tagging everything as breaking" cases (cc @ry ) 😅

lucacasonato (approved):
LGTM. `std` is still unstable, so I wouldn't consider this a breaking change :-)

PR #8037, opened 2020-10-19
fix(cli/repl): ignore pair matching inside literals

### LINKED ISSUE #7978: REPL incorrect behaviour on specific RegExp match
The following RegExp causes the REPL to jump to a new line (expecting more input?) without executing the line, though more simple RegExp seem to work fine.  Using Deno 1.4.6 on Linux.

```
console.log("{test1} abc {test2} def {{test3}}".match(/{([^{].+?)}/));
```
I've narrowed it down to the `[^{]` condition.  Taking out the `[^{]` condition allows the REPL to run the RegExp without the new line/more input needed behaviour.  Also the original RegExp works fine when run via `deno run`.

### LINKED ISSUE #8034: repl hangs on unclosed braces in string literals.
What the title says. Steps to reproduce:

```
$ ./deno --version⏎
deno 1.4.6
v8 8.7.220.3
typescript 4.0.3

$ ./deno repl⏎
Deno 1.4.6
exit using ctrl+d or close()
> let a = "{"⏎   // ... silence
```

Fails independently of quote " ' backtick and brace [ ( { and with let, var and const as well as with naked literals ( `if (a !== "{") console.log("here")` )

I imagine this is some sort of completion gone terribly wrong. No well versed enough in deno internals to spontaneously dig deeper at the moment sorry.

caspervonb:
This ignores brace pairs until a literal sequence has been finished when validating readline input.

Fixes #7978 
Fixes #8034

### DIFF
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -33,7 +33,33 @@ impl Validator for Helper {
     ctx: &mut ValidationContext,
   ) -> Result<ValidationResult, ReadlineError> {
     let mut stack: Vec<char> = Vec::new();
+    let mut literal: Option<char> = None;
+    let mut escape: bool = false;
+
     for c in ctx.input().chars() {
+      if escape {
+        escape = false;
+        continue;
+      }
+
+      if c == '\\' {
+        escape = true;
+        continue;
+      }
+
+      if let Some(v) = literal {
+        if c == v {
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks for quick fix

PR #8038, opened 2020-10-19
feat(cli): stabilize Deno.fsync and Deno.fdatasync

### LINKED ISSUE #7323: Stabilize Deno.datasync and Deno.datasyncSync
This pair is a very straight forward 1:1 mapping to the datasync(2) system call and the interface is not really subject to any further change.

As such they seem like a good candidate for moving out of stable so I'm proposing that we move towards stabilising them.

caspervonb:
Stabilizes Deno.fsync and Deno.fdatasync as these two syscalls take no arguments the interface is final and well known.

Closes #7323


ry (approved):
LGTM - thanks Casper

PR #8039, opened 2020-10-19
fix(op_crates/web): Change TypeError to RangeError

l2ig:
Throw the same error type as Chromium does.
Discussed in #8003.


bartlomieju:
@l2ig thanks for the contribution, please sign the CLA and add a test case exercising this fix.

bartlomieju (approved):
LGTM, nicely done, thanks @l2ig

PR #8044, opened 2020-10-20
docs(std/jwt): Fix examples

KNnut:


std/jwt/README.md
```diff
@@ -9,12 +9,12 @@ Create and verify JSON Web Tokens.
 Takes a `payload`, `key` and `header` and returns the url-safe encoded `token`.
 
 ```typescript
-import { create } from "https://deno.land/std/token/mod.ts";
+import { create } from "https://deno.land/std/jwt/mod.ts";
```
bartlomieju, line 12: Using untagged URLs in README is bad, instead use `@$STD_VERSION` to automatically interpolate with current version of std

bartlomieju (changes requested):

lucacasonato (approved):
LGTM. Thanks

bartlomieju (approved):

PR #8046, opened 2020-10-20
fix(cli): use an interface to represent resources

caspervonb:
This adds an interface to represent resources with an rid to ensure that all implementations that hold a rid use the same signature so that cases like #8014 where one kind of resource has a mutable rid when the others don't, can't happen again.


bartlomieju:
Given that not all classes that are resources expose `rid` I'm not super keen on adding a new interface for this purposes.

caspervonb:
Closing without merge, can be useful but isn't critical.

PR #8050, opened 2020-10-20
BREAKING(cli/tsc): Enable isolatedModules by default

### LINKED ISSUE #7326: Use isolatedModules

### LINKED ISSUE #7947: Flags like --unstable Should Not Apply To Every Library
https://github.com/denoland/deno/pull/7800 I just saw this now from a PR removing this in vscode plugin and am speechless. The amount of time I wasted editing hundreds of files on 5 different projects because this option was enforced because one of the deps that my modules user needed had unstable that had nothing to do with my modules at all :sob: Really disheartening to see all that wasted time and now needing to undo it all.

Applying a single flag should never affect every library. Just because the user of my lib has `--unstable` for another library, Deno marks all libs as using --unstable and causing users to see my lib as unstable/broken and preventing them from running their project. I don't believe this is a good solution to apply the flag to everything. Even though my lib is not unstable, my lib is now the reason they can not run their project, so it's not an option but forcing me to update all my code and then in under a month later "Whoops, we don't need this anymore."

nayeemrmn:
Closes #7326.

Closes #7947, as code written for stable is once again compatible with `--unstable`. Discussion: I strongly feel we should maintain this property, even if it means going straight to stable for future changes like this. Code written for `--unstable` is by design incompatible with stable -- having incompatibilities going the opposite way as well splits the ecosystem in half, generating a lot of headache and noise. Better to have everyone adapt at once.

### DIFF
--- a/cli/module_graph2.rs
+++ b/cli/module_graph2.rs
@@ -698,9 +698,6 @@ impl Graph2 {
     options: CheckOptions,
   ) -> Result<(Stats, Diagnostics, Option<IgnoredCompilerOptions>), AnyError>
   {
-    // TODO(@kitsonk) set to `true` in followup PR
-    let unstable = options.lib == TypeLib::UnstableDenoWindow
-      || options.lib == TypeLib::UnstableDenoWorker;
     let mut config = TsConfig::new(json!({
       "allowJs": true,
       // TODO(@kitsonk) is this really needed?
@@ -708,7 +705,7 @@ impl Graph2 {
       // Enabled by default to align to transpile/swc defaults
       "experimentalDecorators": true,
       "incremental": true,
-      "isolatedModules": unstable,
+      "isolatedModules": true,
       "lib": options.lib,
       "module": "esnext",
       "strict": true,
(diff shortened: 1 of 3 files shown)


kitsonk:
I am in the process of #7225 and specifically #7996 this week, which also address this.  It is best to avoid any other PRs in this area for now.

nayeemrmn:
Okay, thanks.

kitsonk:
It can be rebased now.

kitsonk (changes requested):
This is insufficient, as `cli/tsc.rs` will be retired after #8059 and #8060 are resolved (likely tomorrow).  We need to update [module_graph2.rs](https://github.com/denoland/deno/blob/master/cli/module_graph2.rs#L711) as well.

kitsonk (approved):
LGTM, but I going to leave it for @bartlomieju to merge in light of coordinating other things regarding the 1.5 release.

bartlomieju (approved):
LGTM

PR #8053, opened 2020-10-20
test(std/wasi): reduce test duplication

caspervonb:
This fast-forwards wasi-test-suite a couple of commits.

- Removed duplicate, or near duplicate tests with not much additional value.
- Other tests have been merged to reduce the number of modules to make tests topical to the syscall that is being tested.


bartlomieju (approved):
LGTM

PR #8062, opened 2020-10-21
fix: command examples use outdated default file_server port

ipriver:
https://github.com/denoland/deno/pull/5226 changed `std/http/file_server.ts` default port value from `4500` to `4507`.
https://github.com/denoland/deno/blob/23df1c563efee65e6d549308ed0671a9e2933c05/std/http/file_server.ts#L326
This PR fixes `curl` command example [here](https://deno.land/manual@v1.4.6/getting_started/debugging_your_code#chrome-devtools) and `stdout` output in [std/http/README.md](https://github.com/denoland/deno/tree/master/std/http#file-server)

### DIFF
--- a/docs/getting_started/debugging_your_code.md
+++ b/docs/getting_started/debugging_your_code.md
@@ -68,7 +68,7 @@ Once our script is running again, let's send a request and inspect it in
 Devtools:
 
 ```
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks @ipriver

PR #8063, opened 2020-10-21
fix(std/testing) : Make equal function handle correctly symbols properties

lowlighter:
Currently `symbol`s are not properly handled in `std/testing/asserts` because of the way objects are iterated in [equality function](https://github.com/denoland/deno/blob/master/std/testing/asserts.ts#L136).

**Example**
```typescript
import { assertEquals } from "https://deno.land/std/testing/asserts.ts"
const foo = Symbol("foo")
Deno.test("symbol1", () => assertEquals({[foo]:"bar"}, {[foo]:"bar"}))
Deno.test("symbol2", () => assertEquals({[foo]:"bar"}, {[Symbol("foo")]:"bar"}))
```
**Expected**
```
running 2 tests
test symbol1 ... ok (3ms)
test symbol2 ... FAILED (2ms)

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (5ms)
```
**Actual**
```
running 2 tests
test symbol1 ... ok (2ms)
test symbol2 ... ok (1ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (4ms)
```
In the second test, since symbols are not the same, it should throws.

### DIFF
--- a/cli/tests/unit/response_test.ts
+++ b/cli/tests/unit/response_test.ts
@@ -41,7 +41,7 @@ unitTest(async function responseFormData() {
   const input = new FormData();
   input.append("hello", "world");
   const response = new Response(input, {
-    headers: { "content-type": "application/x-www-form-urlencoded" },
+    headers: { "content-type": "multipart/form-data" },
   });
   const formDataPromise = response.formData();
   assert(formDataPromise instanceof Promise);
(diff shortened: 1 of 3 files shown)


lowlighter:
Mis-closed by error

bartlomieju:
@lowlighter thanks for the PR, could you fix failing test so we can land this PR?

lowlighter:
@bartlomieju The failing test case has already been heavily debugged in #8069, but we've yet to find a patch. 
With a bit more insight about it, I'll be happy to implement the fix for `response.formData` so we can also land #8063

PR #8064, opened 2020-10-21
typo(jsdoc): `creat` -> `create`

CodingCarter:
Another thing: couldn't we be doing `===` for comparing the `typeof value == "string"` and `typeof namespace == "string"` in the `if` statements of the `generate` function?  I was testing out behavior for if it was just to catch fallback for if they were actual string objects like `new String("foo")`, but it's not.


kitsonk (approved):
LGTM

PR #8075, opened 2020-10-22
feat(lint): stabilize "deno lint" subcommand

### LINKED ISSUE #7983: Move `deno lint` to Stable
**References**
image

**Why**
The main show stopper for some people i know of, for **not** using `deno lint`, is the fact it is unstable. Judging by @bartlomieju's response,  it doesn't seem to need the unstable flag anymore either.

I believe that people feel they would adjust their code base to be in line with `deno lint`, then say, the next day a change to `lint` was made, and there's a gazillion errors when doing another `deno lint` (eg the whole idea of unstable, which i'm not saying is bad btw)

bartlomieju:
This commit stabilizes "deno lint" by removing
the need to pass --unstable flag.

--unstable is still required when using --json flag.

Closes #7983

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -989,18 +989,18 @@ fn lint_subcommand<'a, 'b>() -> App<'a, 'b> {
     .about("Lint source files")
     .long_about(
       "Lint JavaScript/TypeScript source code.
-  deno lint --unstable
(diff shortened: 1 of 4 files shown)


ry (approved):
LGTM

PR #8078, opened 2020-10-22
fix(std/wasi): disallow path_open outside of pre-opened dirfd

caspervonb:
Disallow path_open to open paths outside of the dirfd that is being accessed.

### DIFF
--- a/std/wasi/snapshot_preview1.ts
+++ b/std/wasi/snapshot_preview1.ts
@@ -1209,6 +1209,7 @@ export default class Context {
         );
         const resolvedPath = resolve(entry.path!, textDecoder.decode(pathData));
 
+        console.log("RESOLVED PATH: %s", resolvedPath);
         if (relative(entry.path, resolvedPath).startsWith("..")) {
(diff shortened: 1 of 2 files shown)


caspervonb:
Think I might have stumbled upon a bug with symlinks and/or std/fs/copy.
Unix works fine, but on Windows symlink_to_parent_directory (a relative symlink with ..) doesn't appear to be a symlink at all.

Don't have a windows machine to work on this with so this is very difficult to debug.

lucacasonato:
> Unix works fine, but on Windows symlink_to_parent_directory (a relative symlink with ..) doesn't appear to be a symlink at all.

cc @piscisaureus

caspervonb:
I'm appending this new test to the growing list of ignored windows tests for the time being until I get around to provisioning a Windows based server to fix them all.

Windows behaviour isn't any worse than it was before this commit, the new tests just more obvious that anything symlink related is probably broken.

bartlomieju (approved):
LGTM

PR #7612, opened 2020-09-21
refactor(watch): create single watcher for whole process

bartOssh:
[fixes small part of 7465 to not create watcher after each change of watched files](https://github.com/denoland/deno/issues/7465)

I have decided to bring the solution in small steps to not to run in to messing some things up.

Issue #7465, opened 2020-09-14
Add --watch flag to other subcommands


bartlomieju:
`deno run --watch` was added in #7382.

Now we should work towards adding the same flag to other subcommands:
- `bundle`
- `fmt`
- `lint`
- `test`

Each of the subcommands might require a bit different approach as not all of them work using `ModuleGraph`.

We might also want to refactor `file_watcher` not to create a new watcher after each restart.

bartlomieju:
This is mostly implemented, still missing `deno lint --watch`, but let's continue in https://github.com/denoland/deno/issues/11132

PR #8083, opened 2020-10-23
refactor(watch): create single watcher for whole process

magurotuna:
This is a follow-up PR to #7612 
Fixes small part of #7465, so that we won't create file watcher every time on file updating.

When running `cargo run -- run --watch --unstable foo.js` and update the content of `foo.js`, it seems like I've got results as expected.
The integration test `run_watch` fails locally right now though. I'm looking into it.

cc @bartlomieju


magurotuna:
https://github.com/denoland/deno/pull/8083/checks?check_run_id=1295812207#step:15:1313
hmm looks like `run_watch` fails only on Linux (my machine is also Linux).

It might be a problem with `notify` crate. We currently use `notify` v5.0.0-pre.3, but I want to try v.4.0.15 instead, which is the stable version. Moreover, v4.0.15 has the functionality of handling "debounced" events, so we could remove our debouncing logic from our codebase.
https://docs.rs/notify/4.0.15/notify/trait.Watcher.html

cli/file_watcher.rs
```diff
@@ -11,14 +12,81 @@ use notify::Error as NotifyError;
...
+    self: Pin<&mut Self>,
+    cx: &mut Context,
+  ) -> Poll<Option<Self::Item>> {
+    let inner = self.get_mut();
+    if let Ok(Ok(event)) = inner.rx.try_recv() {
+      match inner.last_event.as_ref() {
+        Some(last_event) if *last_event == event => {
+          // if received event is the same as previous one, reset timeout
+          inner.start_time = Some(Instant::now());
+        }
```
bartlomieju, line 61: I don't understand this bit, is it possible to receive exactly same event twice?

bartlomieju (commented):
@magurotuna I'd be cautious about downgrading `notify` - we also use it in `Deno.fsEvents` and it might be pretty hard to replace it. Could you elaborate what's the problem you're experiencing?

magurotuna:
@bartlomieju 

> Could you elaborate what's the problem you're experiencing?

Oh, I didn't notice `notify` is also used for `Deno.fsEvents`. I'll try to find a solution without downgrading.

cli/file_watcher.rs
magurotuna: It's based on @bartOssh's code: https://github.com/denoland/deno/blob/988e6c19386a15543a09044a483cbca6aeeff38b/cli/file_watcher.rs#L58-L61

But after thinking again about it, this check seems unnecessary - if any event is received from `rx`, we always should reset timeout. 
I'll fix the code.

magurotuna:
`run_watch` is writing a file by calling `std::fs::write`. On Linux, when calling `std::fs::write`, it seems like the following events are what we get via `notify`:

```
[cli/file_watcher.rs:156] &res = Ok(
    Event {
        kind: Modify(
            Data(
                Any,
            ),
        ),
        paths: [
            "/tmp/deno-watcher/foo.js",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
)
[cli/file_watcher.rs:156] &res = Ok(
    Event {
        kind: Access(
            Close(
                Write,
            ),
        ),
        paths: [
            "/tmp/deno-watcher/foo.js",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
)
```

On the other hand, if I hit `:w` to safe the file in vim, events change to:

```
[cli/file_watcher.rs:146] &res = Ok(
    Event {
        kind: Modify(
            Data(
                Any,
            ),
        ),
        paths: [
            "/tmp/deno-watcher/foo.js",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
)
[cli/file_watcher.rs:146] &res = Ok(
    Event {
        kind: Access(
            Close(
                Write,
            ),
        ),
        paths: [
            "/tmp/deno-watcher/foo.js",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
)
[cli/file_watcher.rs:146] &res = Ok(
    Event {
        kind: Modify(
            Metadata(
                Any,
            ),
        ),
        paths: [
            "/tmp/deno-watcher/foo.js",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
)
```

In the current implementation, we treat `EventKind::Create | EventKind::Modify | EventKind::Remove` as the triggers for reloading, but in the case of `std::fs::write`, the event kind we get last is `EventKind::Access`. This is why `run_watch` test fails.
I'll try to elaborate some solution.

cli/file_watcher.rs
```diff
@@ -11,14 +12,75 @@ use notify::Error as NotifyError;
...
+        }) =>
+      {
+        inner.start_time = None;
+        let event = mem::take(&mut inner.last_event);
+        Poll::Ready(event)
+      }
+      _ => {
+        cx.waker().wake_by_ref();
+        Poll::Pending
+      }
```
magurotuna, line 79: This may lead to a busy loop, so ideally we insert sleep or something here. But at the same time we want to receive events as soon as they are sent.
Any thoughts on this? @bartlomieju

cli/file_watcher.rs
```diff
@@ -28,51 +90,55 @@ async fn error_handler(watch_future: WatchFuture) {
...
 
   let mut watcher: RecommendedWatcher =
     Watcher::new_immediate(move |res: Result<NotifyEvent, NotifyError>| {
       let res2 = res.map_err(AnyError::from);
-      let mut sender = sender.lock().unwrap();
       // Ignore result, if send failed it means that watcher was already closed,
       // but not all messages have been flushed.
-      let _ = sender.try_send(res2);
+      let _ = sender.send(res2);
     })?;
```
magurotuna, line 142: I just came up with an idea: instead of using `mpsc`, update `Debounce.last_event` inside the callback of `Watcher::new_immediate`, and `Debounce::poll_next` should just check if `DEBOUNCE_TIMEOUT_MS` elapsed since the last event.

cli/file_watcher.rs
```diff
@@ -28,51 +90,55 @@ async fn error_handler(watch_future: WatchFuture) {
...
-    };
+      _ = debounce.next() => {
+        is_file_changed = true;
+        info!(
+          "{} File change detected! Restarting!",
+          colors::intense_blue("Watcher"),
+        );
+      },
+      _ = func => {},
+    }
```
caspervonb, line 113: An else `sleep_for(DEBOUNCE_TIMEOUT_MS).await` here should be enough to calm down the hot loop

magurotuna: I don't think that should work - consider the following code:

```ts
setTimeout(() => { console.log("our code finished!"); }, 5000);
```

then we would never see `our code finished!` because `sleep_for(DEBOUNCE_TIMEOUT_MS)` would be ready prior to our JS code.

magurotuna:
For now, I have added 10ms sleep to avoid a hot loop, referring to https://github.com/denoland/deno/pull/7612#discussion_r499234629
On my machine, CPU usage of Deno's process has reduced to ~3.0% after adding a sleep. (Without sleep, it was 100%)
This looks fine to me 😃

cli/file_watcher.rs
caspervonb: It wouldn't prevent any execution, just delay it.
Delay can be tweaked, delay_for(0) is good enough to yield 0-10% CPU usage.

Still better to do it explicitly in the hot loop rather than hiding an actual thread sleep somewhere else IMO.

magurotuna: I might have misunderstood you. I thought what you suggested was to add `delay_for()` to a branch of `select!` macro, but it seems like my understanding was incorrect.
Where do you propose we should insert `delay_for()`? 

> Also a thread sleep will halt execution, won't it?

I don't think it will. The sleep I inserted to `poll_next` doesn't prevent the tokio's worker thread, and the actual sleep is executed on another thread, which is outside of tokio's thread pool. Once the time has elapsed on the "sleep" thread, it will notify one of the worker threads to poll again.

caspervonb: > I thought what you suggested was to add delay_for() to a branch of select!

If none of the futures complete, delay for at-least 0ms.

As is this still runs hot.

Screenshot 2020-10-26 at 9 12 19 AM

bartlomieju:
Discussed offline with @magurotuna and @caspervonb. Given bugs in watcher implementation we decided to postpone this PR and rewrite watcher properly.

cli/file_watcher.rs
magurotuna: thanks, to be honest I didn't know `tokio::select` has a special branch`else`.
I read `tokio::select` and `futures::select` documents, and it seems like there's a slight difference between tokio::select's `else` branch and futures::select's `default` branch.

[tokio::select - Rust](https://docs.rs/tokio/0.2.22/tokio/macro.select.html)
[futures::select - Rust](https://docs.rs/futures/0.3.7/futures/macro.select.html)

tokio::select's `else` branch is executed only if futures of the other branches have been ready but none of them matches each provided "pattern".
futures::select's `default` branch, on the other hand, is executed if none of the other branches' futures is immediately ready. I think this is what you explained to me.
So, instead of using tokio::select's `else` branch, we should probably add some sleep after the `select!`.

caspervonb:
LGTM, doesn't seem to spike above 1%

magurotuna:
The remaining issue I see is that we get multiple consecutive "Watcher File changed detected! Restarting!" message. @bartlomieju gave me information about it.

#### terminal

```
hello world
a
b
c
hello world
a
b
c
hello world
a
b
c
hello world
a
b
c
hello world
```

#### How to reproduce

1. create `main.js` with the following content:

```js
import { a } from "./test2/a.js";
import { b } from "./test2/b.js";
import { c } from "./test2/c.js";

console.log(a);
console.log(b);
console.log(c);
console.log("hello world");
```

2. run `deno run --watch --unstable main.js`
3. run `mv ./test2 ./test1`
4. replace `test2` with `test1` in `main.js` and save it

magurotuna:
I gave it a try to reproduce it. On my main machine, where Manjaro 20.1.2 is running, the output I saw looks like:

image

Well, actually the watcher didn't anything at all when I ran `mv ./test2 ./test1`, while it detected file change only when replacing `test2` with `test1` in `main.js` and saving it.

To see more detail, I inserted `dbg!` in the watcher callback so I can see what kind of event the watcher detects. But in fact, at the time when I ran `mv ./test2 ./test1`,  **the watcher didn't  get any kind of event.**

I also tried it on macOS. The result is almost identical, except for the events' kind which I think doesn't matter.

magurotuna:
Talked with @bartlomieju about the above issue, and it turned out that the watcher actually works as we expect. So this PR is ready to be merged.
Please merge it when it's convenient.

bartlomieju (approved):
LGTM, thanks @magurotuna, @bartOssh, @caspervonb, this is a great improvement

PR #8089, opened 2020-10-23
Revert "feat(cli): Add deno cache --test and --worker (#7920)"

bartlomieju:
This reverts commit be15cf285dbf3a7b3025d34cb80e87e54a55dd93.

CC @ry @nayeemrmn

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -3,7 +3,6 @@
 use clap::App;
 use clap::AppSettings;
 use clap::Arg;
-use clap::ArgGroup;
 use clap::ArgMatches;
 use clap::ArgSettings;
 use clap::SubCommand;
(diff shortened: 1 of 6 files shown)


ry:
@nayeemrmn I requested this revert. 

I have a couple problems:
- these are pretty different things, they should be landed in separate PRs
- adding stabilized flags without sufficient discussion 
- I dislike that --worker introduces conditional compilation into the system
- I dislike that deno test begets more complexity, like with this `deno cache --test`

I don't have an alternate solution ready - but I think we should be cautious and not make changes without consensus - so best to revert it for now.

nayeemrmn:
I agree the implementation for `--worker` is weird, it's constructing a `WebWorker` just to compile with different types. I'll re-open only for `--test`.

> * I dislike that deno test begets more complexity, like with this `deno cache --test`

This is more than baggage brought on by `deno test` in my mind. I expect `deno cache --test` to be the new go-to command for setting up an offline-ready development environment -- as an alternative to `deno cache deps.ts dev_deps.ts` or something else which depends on the user's source tree. See `cargo build --tests` and `go test -c`, a way to compile-only tests is a fundamental expectation for a built-in test runner.

ry (approved):
LGTM

Issue #8059, opened 2020-10-21
Migrate Deno.transpileOnly to new tsc/module_graph


kitsonk:
We need to migrate `Deno.transpileOnly()` to the new compiler infrastructure.

PR #8090, opened 2020-10-23
refactor(cli): rewrite Deno.transpileOnly() to use SWC

bartlomieju:
Closes https://github.com/denoland/deno/issues/8059


cli/ops/runtime_compiler.rs
```diff
@@ -71,16 +77,52 @@ struct TranspileArgs {
...
+    json!({})
+  };
+  compiler_options.merge(&user_options);
+
+  let emit_options: ast::EmitOptions = compiler_options.into();
+  let mut emit_map = HashMap::new();
+
+  for (specifier, source) in args.sources {
+    let media_type = MediaType::from(&PathBuf::from(&specifier));
+    let module_specifier = ModuleSpecifier::resolve_url(&specifier)?;
```
bartlomieju, line 114: This is a big change and I think it will cause tests to fail - currently `Deno.traspileOnly()` doesn't require valid module specifier. I had to resolve the specifier here, because it's required by `ast::parse`. We'll have to discuss this one.

bartlomieju: This problem can be circumvented by rolling back `ast::parse` to take specifier as `&str` instead of `&ModuleSpecifier`

cli/ops/runtime_compiler.rs
```diff
@@ -71,16 +76,57 @@ struct TranspileArgs {
...
+
+  let mut compiler_options = tsc_config::TsConfig::new(json!({
+    "checkJs": true,
+    "emitDecoratorMetadata": false,
+    "jsx": "react",
+    "jsxFactory": "React.createElement",
+    "jsxFragmentFactory": "React.Fragment",
+    "esModuleInterop": true,
+    "module": "esnext",
+    "sourceMap": true,
```
bartlomieju, line 101: This setting is currently completely ignored in `impl From<tsc_config::TsConfig> for EmitOptions` - source maps are always inlined.

cli/ops/runtime_compiler.rs
```diff
@@ -71,16 +76,57 @@ struct TranspileArgs {
...
+    "scriptComments": true,
+    "target": "esnext",
```
bartlomieju, line 103: This setting is currently completely ignored. One of our tests uses `amd` as `target`. We need to implement that for other targets as well. Reference https://swc.rs/rustdoc/swc_ecma_transforms/modules/amd/struct.Config.html

cli/ops/runtime_compiler.rs
```diff
@@ -71,16 +76,57 @@ struct TranspileArgs {
...
-    runtime_transpile(program_state, &args.sources, &args.options).await?;
+
+  let mut compiler_options = tsc_config::TsConfig::new(json!({
+    "checkJs": true,
+    "emitDecoratorMetadata": false,
+    "jsx": "react",
+    "jsxFactory": "React.createElement",
+    "jsxFragmentFactory": "React.Fragment",
+    "esModuleInterop": true,
+    "module": "esnext",
```
bartlomieju, line 100: This setting is currently completely ignored. One of our tests uses `amd` as `module`. We need to implement that for other remaining `module` options as well. Reference https://swc.rs/rustdoc/swc_ecma_transforms/modules/amd/struct.Config.html

bartlomieju: And it turns out `SystemJS` is not yet supported by SWC.

cli/ops/runtime_compiler.rs
bartlomieju: I tentatively applied this fix

cli/ops/runtime_compiler.rs
bartlomieju: Added `inline_source_map` to `EmitConfigOptions`

cli/ops/runtime_compiler.rs
bartlomieju: As per Discord conversation I removed option to specify output `module` - actually user supplied config is run through the same set of `IGNORED_COMPILER_OPTIONS` and API returns error when there's an ignored option.

cli/ops/runtime_compiler.rs
kitsonk: Yeah, I guess `ast::parse` doesn't need to be a `&ModuleSpecifier`... I just wanted to use `ModuleSpecifier` everywhere and just convert at the edge, so I think `ast::parse` going back to `&str` would be fine.  I guess it shouldn't really have "Deno" specific opinions where avoidable.

cli/ops/runtime_compiler.rs
kitsonk: Ah, yeah... good catch.

cli/ops/runtime_compiler.rs
kitsonk: It probably needs to be a different set, because things like `target` should be changeable, like `target`, `inlineSourceMaps`, `inlineSources`, etc, which we disallow because they change the emit internally to Deno.  There are probablly two different classes of options... the ones for Deno internal emit we ignore, and ones we ignore totally/don't support.  For `--config` it is both sets, but for the the runtime APIs it is just the second set.

cli/ops/runtime_compiler.rs
```diff
@@ -71,16 +75,52 @@ struct TranspileArgs {
...
+    "jsxFragmentFactory": "React.Fragment",
+    "inlineSourceMap": false,
+  }));
+
+  let user_options = if let Some(options) = args.options {
+    tsc_config::parse_config_with_ignored_options(&options)?
+  } else {
+    json!({})
+  };
+  compiler_options.merge(&user_options);
```
kitsonk, line 106: Maybe...

```suggestion
  if let Some(options) = args.options {
    let user_options = tsc_config::parse_config_with_ignored_options(&options)?;
    compiler_options.merge(&user_options);
  }
```

kitsonk: Actually why aren't you using `TsConfig::merge_user_config()`?  I wanted to get rid of `parse_config` when we dropped `tsc` anyways, as why can't these just live as methods on `TsConfig`?

cli/tsc_config.rs
```diff
@@ -167,6 +169,42 @@ pub fn parse_raw_config(config_text: &str) -> Result<Value, AnyError> {
   Ok(jsonc_to_serde(jsonc))
 }
 
+/// Parse provided options and return error if config contains
+/// ignored options.
+pub fn parse_config_with_ignored_options(
```
kitsonk, line 174: Why aren't we just using/modifying `TsConfig::merge_user_config`?  Also `IgnoredCompilerOptions` already includes a meaningful error message that duplicates what we are adding here.

I think it is better to just return what was ignored to the API and let the API figure out if that is a meaningful and what to do with it.

cli/ops/runtime_compiler.rs
kitsonk: Not get rid of `parse_config` but make it non-public.

cli/tsc_config.rs
```diff
@@ -176,18 +214,13 @@ pub fn parse_config(
...
-  if let Some(in_compiler_options) = config.compiler_options {
-    for (key, value) in in_compiler_options.iter() {
-      if IGNORED_COMPILER_OPTIONS.contains(&key.as_str()) {
-        items.push(key.to_owned());
-      } else {
-        compiler_options.insert(key.to_owned(), value.to_owned());
-      }
-    }
-  }
+  let (compiler_options, items) =
```
kitsonk, line 218: Yeah, this seems to be making things more complex then they need to be, IMO.  We already have too much indirection in the code, I would hate to continue to add back more.

cli/tsc_config.rs
bartlomieju: @kitsonk because we have a JSON string with config and not a path - `merge_user_config` is only useful if you want to load and merge config from a file.

cli/ops/runtime_compiler.rs
bartlomieju: Sounds good, I'll rework it

cli/tsc_config.rs
bartlomieju: > I think it is better to just return what was ignored to the API and let the API figure out if that is a meaningful and what to do with it.

So change return type to return list of ignored options?

kitsonk: 🤦 oh yeah regarding the `merge_user_config`... Maybe we should have `merge_tsconfig(path)` to replace what we already do and make `merge_user_config(Value)` for what you need to do here?

kitsonk:
Ok, I made an attempt to address things.  Let me know if you like it @bartlomieju.

cli/ops/runtime_compiler.rs
```diff
@@ -71,16 +76,58 @@ struct TranspileArgs {
...
+  source: String,
+  map: Option<String>,
+}
+
 async fn op_transpile(
   state: Rc<RefCell<OpState>>,
   args: Value,
   _data: BufVec,
 ) -> Result<Value, AnyError> {
   super::check_unstable2(&state, "Deno.transpile");
```
kitsonk, line 90: Shouldn't this be:

```suggestion
  super::check_unstable2(&state, "Deno.transpileOnly");
```

bartlomieju:
> Ok, I made an attempt to address things. Let me know if you like it @bartlomieju.

@kitsonk LGTM, I think we achieved major goal of this PR which was migrating `Deno.transpileOnly()` API to SWC. All the rest can be addressed in follow ups

ry (approved):
LGTM

PR #8091, opened 2020-10-23
refactor(core): remove Deno.core.formatError()

bartlomieju:
With recent improvements to REPL implementation,
Deno.core.formatError() API is no longer needed.


ry (approved):
great

PR #8093, opened 2020-10-23
feat(unstable): Add deno test --no-run

### LINKED ISSUE #7775: Add deno test --no-run

nayeemrmn:
Closes #7775.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1154,7 +1154,8 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("no-run")
         .long("no-run")
         .help("Cache test modules, but don't run tests")
-        .takes_value(false),
+        .takes_value(false)
+        .requires("unstable"),
     )
     .arg(
       Arg::with_name("failfast")
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -66,6 +66,7 @@ pub enum DenoSubcommand {
     script: String,
   },
   Test {
+    no_run: bool,
     fail_fast: bool,
     quiet: bool,
     allow_none: bool,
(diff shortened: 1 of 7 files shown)


bartlomieju:
@nayeemrmn I thought more about this problem and I propose:
```
deno cache
deno cache --worker
deno test --cache
```
`deno cache` currently caches dependencies and type checks all files from given entry point. If we were to add `deno cache --test` we're starting to interleave `deno test` glob resolution into `cache` command. Adding `deno cache --worker` just adds another TS library to type check. (Maybe it should be `deno cache --lib worker`?)

nayeemrmn:
> I thought more about this problem and I propose:
> 
> ```
> deno cache
> deno cache --worker
> deno test --cache
> ```

Sounds good, as I said `cargo` has both `cargo build --tests` and `cargo test --no-run`. I think matching `deno test --no-run` would be more clear. It's a similar scheme to `--no-check`.

bartlomieju:
> > I thought more about this problem and I propose:
> > ```
> > deno cache
> > deno cache --worker
> > deno test --cache
> > ```
> 
> Sounds good, as I said `cargo` has both `cargo build --tests` and `cargo test --no-run`. I think matching `deno test --no-run` would be more clear. It's a similar scheme to `--no-check`.

Agreed, I was thinking `--dry` but `--no-run` is self-explanatory 👍

bartlomieju:
@nayeemrmn let's require `--unstable` for this flag for the time being

bartlomieju (approved):
LGTM, thanks @nayeemrmn

PR #8098, opened 2020-10-24
test(std/wasi): re-enable clock_time_get tests

caspervonb:
This brings in revised tests for clock_time_get from upstream and re-enables them.

This revision of these tests allow for lower precision clocks (we only provide millisecond precision, and that is conformant but my previous tests for this in the upstream test suite were a bit more aggressive).


caspervonb:
Blocking on https://github.com/denoland/deno/pull/8078 as this is based on an upstream commit after that so just to keep things nice and tidy, merge 8078 first.

bartlomieju (approved):
LGTM

PR #8100, opened 2020-10-24
fix(op_crates/fetch): ensure Request.method to use the default value if not defined

kidonng:
Ensure `Request.method` to be the default value (`GET`) if `init.method` is not defined, which follows browser's behavior.

Originally sindresorhus/ky#292

```ts
// Before this PR
console.log(new Request('', { method:undefined }).method) // undefined

// After this PR
console.log(new Request('', { method:undefined }).method) // 'GET'
```

### DIFF
--- a/op_crates/fetch/26_fetch.js
+++ b/op_crates/fetch/26_fetch.js
@@ -990,7 +990,7 @@
         this.url = new URL(String(input)).href;
       }
 
-      if (init && "method" in init) {
+      if (init && "method" in init && init.method) {
         this.method = normalizeMethod(init.method);
       }
 
--- a/cli/tests/unit/request_test.ts
+++ b/cli/tests/unit/request_test.ts
@@ -41,6 +41,10 @@ unitTest(function requestNonString(): void {
   assertEquals(new Request(nonString).url, "http://foo/");
 });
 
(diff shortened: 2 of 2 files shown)


bartlomieju (approved):
LGTM, great fix @kidonng

op_crates/fetch/26_fetch.js
```diff
@@ -990,7 +990,7 @@
         this.url = new URL(String(input)).href;
       }
 
-      if (init && "method" in init) {
+      if (init && "method" in init && init.method) {
```
mayankchoubey1, line 993: Shouldn't it be if (init && init.method) ?

mayankchoubey1 (commented):
There are extra checks in the if condition.

mayankchoubey1 (commented):
The if condition has an extra check

op_crates/fetch/26_fetch.js
kidonng: Ideally we can just check `init.method`, I'm just following existing code like https://github.com/denoland/deno/blob/3fc4df3887d1611aeb87b8d1fe08b227b7e724a9/op_crates/fetch/26_fetch.js#L998-L1000

Issue #5935, opened 2020-05-29
can't reexport modules typed by X-TypeScript-Types custom header


fjc0k:
```ts
// mod.ts
export * from 'https://cdn.pika.dev/vtils@3.0.1-beta.2'
```

```ts
// a.ts
import * as vtils from './mod.ts'

console.log(Object.keys(vtils).includes('isNumber'))
```

`deno run -A -r a.ts`:

![image](https://user-images.githubusercontent.com/13151189/83218638-e6f6e180-a1a0-11ea-942a-fa1f01f1f0f2.png)

```ts
// b.ts
import * as vtils from './mod.ts'

console.log(vtils.isNumber)
```

`deno run -A -r b.ts`:

![image](https://user-images.githubusercontent.com/13151189/83218678-fe35cf00-a1a0-11ea-95bb-72f343ba9020.png)

```ts
// c.ts
import * as vtils from 'https://cdn.pika.dev/vtils@3.0.1-beta.2'

console.log(vtils.isNumber)
```

`deno run -A -r c.ts`:

![image](https://user-images.githubusercontent.com/13151189/83218935-a186e400-a1a1-11ea-8136-6a0c95ad73ce.png)

marcushultman:
I see a similar effect when using `@deno-types` and re-export

kitsonk:
Need to try to address in #7225.

PR #8101, opened 2020-10-24
test(cli): ensure modules can export typed JS files

kitsonk:
Adds a test which verifies that #5935 is fixed.


bartlomieju (approved):
LGTM

PR #8103, opened 2020-10-24
fix(op_crates/web): Expose event properties in console output

### LINKED ISSUE #8073: cli/rt and op_crate/web - console output for ErrorEvent should be more informative
Imagine someone is writing a WebSocket client. MDN has a couple examples for handling WebSocket errors, e.g. in https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/onerror and https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event. Both just call `console.log` or `console.error` with the argument.

Consider this example:

```js
const socket = new WebSocket('ws://localhost:8080');
socket.onerror = function(event) {
  console.error("WebSocket error observed:", event);
};
```

Running it as `deno run test.js` produces the following output:

```
WebSocket error observed: ErrorEvent { isTrusted: false }
```

Not very informative. https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent does list the `error` property, but it's not easily discoverable (there are no links to it from any WebSocket documentation pages as far as I can tell) and its significance isn't obvious.

ross-weir:
Fixes #8073 

New output using the example in the issue:

```js
ErrorEvent {
  bubbles: false,
  cancelable: false,
  composed: false,
  currentTarget: null,
  defaultPrevented: false,
  eventPhase: 0,
  target: null,
  timeStamp: 1603532762591,
  type: "error",
  message: "",
  filename: "",
  lineno: 0,
  colno: 0,
  error: null
}
```

### DIFF
--- a/cli/rt/27_websocket.js
+++ b/cli/rt/27_websocket.js
@@ -61,7 +61,7 @@
             }).then(() => {
               this.#readyState = CLOSED;
 
-              const errEvent = new Event("error");
+              const errEvent = new ErrorEvent("error");
               errEvent.target = this;
               this.onerror?.(errEvent);
               this.dispatchEvent(errEvent);
@@ -84,7 +84,7 @@
         } else {
           this.#readyState = CLOSED;
 
-          const errEvent = new Event("error");
+          const errEvent = new ErrorEvent("error");
           errEvent.target = this;
           this.onerror?.(errEvent);
           this.dispatchEvent(errEvent);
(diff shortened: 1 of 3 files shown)


kitsonk (commented):
Probably need some unit test cases in `cli/tests/unit/event.ts`.

kitsonk (approved):
LGTM... Awesome contribution!

PR #8110, opened 2020-10-24
fix(cli): support "deno run --v8-flags=--help" without script

### LINKED ISSUE #5949: "deno run --v8-flags=--help" is broken
```
> deno run --v8-flags=--help
error: The following required arguments were not provided:
    <SCRIPT_ARG>...

USAGE:
    deno run <SCRIPT_ARG>... --v8-flags=<v8-flags>

```

This works.
```
deno run --v8-flags=--help _
```

It's because it's missing the script. We need a better way of doing this.

ipriver:
fixes #5949

The issue above is happening because the flag `run` always requires `script` to be provided in the command arguments and `clap` verifies it according to its params. 
And, unfortunately, `clap` supports only [required_unless](https://docs.rs/clap/2.26.2/clap/struct.Arg.html#method.required_unless) method which cannot be used here - it checks only flag name, but cannot also check its value. On the other hand, in our case [default_value_ifs](https://docs.rs/clap/2.26.2/clap/struct.Arg.html#method.default_value_ifs) doesn't require any parsing logic or types modification, `script` itself is never being used as the deno process gets finished after it finds v8 --help flag and prints output.

https://github.com/denoland/deno/blob/9adbca6d195b8ea9ad1474ebbb6f906069d6b130/cli/main.rs#L723-L725


ry (approved):
Thank you!

ipriver:
https://github.com/denoland/deno/pull/8110/commits/f3c966cc7095dfa5f2192ad03580e09d5a8a7939 should not be committed as it allows to use `run` without any script and breaks the whole idea of giving `v8-flags=--help` such functionality as an exception using `default_value_ifs`.
I see that https://github.com/denoland/deno/pull/8110/commits/6023f5b750d8d9fb6a0e3a34d0a1e9e6cb5cdb96 is not passing some tests now, gotta check what has changed in `master` branch in the past 2 days, because before that everything worked fine.

ipriver:
> [f3c966c](https://github.com/denoland/deno/commit/f3c966cc7095dfa5f2192ad03580e09d5a8a7939) should not be committed as it allows to use `run` without any script and breaks the whole idea of giving `v8-flags=--help` such functionality as an exception using `default_value_ifs`.

OK, https://github.com/denoland/deno/commit/f3c966cc7095dfa5f2192ad03580e09d5a8a7939 is valid, `required(true)` got moved to another position in https://github.com/denoland/deno/commit/ece1e1d5f184165f21b17d3fe293d8d31dd9e413. 
Everything is fine, @ry sorry for double-checking.

bartlomieju (approved):
LGTM

PR #8113, opened 2020-10-25
docs(std/io/streams): Added JSDoc for std/io/streams.ts file #7487

Luisbo85:
Added JSDoc for std/io/streams.ts file #7487


bartlomieju:
Thanks you @Luisbo85, but it seems those docs are already added on `master`, closing without merge.

Issue #8111, opened 2020-10-24
Emit for imported ts files is sometimes skipped


lucacasonato:
For some reason the JS for `https://cdn.dreg.dev/polyfill/node/process.ts` is not emitted when imported from `https://cdn.dreg.dev/package/picomatch@2.2.2` imported from `deno cache https://cdn.dreg.dev/package/@rollup/pluginutils@4.0.0/dist/es/index.js`.

```
$ rm -rf $DENO_DIR
$ deno cache https://cdn.dreg.dev/package/@rollup/pluginutils@4.0.0/dist/es/index.js
(snip)
Check https://cdn.dreg.dev/package/@rollup/pluginutils@4.0.0/dist/es/index.js
warning: Compiled module not found "https://cdn.dreg.dev/polyfill/node/process.ts"
  From: https://cdn.dreg.dev/package/picomatch@2.2.2
    If the source module contains only types, use `import type` and `export type` to import it instead.
error: Uncaught SyntaxError: The requested module '/polyfill/node/process.ts' does not provide an export named 'default'
```

If I manually force an emit with `deno cache` it works.

```
$ rm -rf $DENO_DIR
$ deno cache https://cdn.dreg.dev/polyfill/node/process.ts
(snip)
Check https://cdn.dreg.dev/polyfill/node/process.ts
$ deno cache https://cdn.dreg.dev/package/@rollup/pluginutils@4.0.0/dist/es/index.js
(snip)
Check https://cdn.dreg.dev/package/@rollup/pluginutils@4.0.0/dist/es/index.js
```

cc @kitsonk

kitsonk:
So this is caused by coupling to some unintentional (and undesired) behaviour that has been removed in master.

In the new infrastructure, when there are types available, `tsc` only sees the type files.  This avoids the host of issues we have had with JavaScript suddenly being type checked, plus helps ensure that the type checking is "accurate".

`https://cdn.dreg.dev/package/picomatch@2.2.2` serves up a `X-TypeScript-Types` header that resolves to `https://cdn.dreg.dev/package/@types/picomatch@2.2.1` which contains:

```ts
export * from "/package/@types/picomatch@2.2.1/index.d.ts";
export { default } from "/package/@types/picomatch@2.2.1/index.d.ts";
```

[...]

Issue #8106, opened 2020-10-24
can not cache extensionless URLs


lucacasonato:
master
```
/m/f/P/g/l/d/example ❯❯❯ rm -rf $DENO_DIR
/m/f/P/g/l/d/example ❯❯❯ deno cache https://cdn.dreg.dev/package/picomatch@2.2.2
(snip)
Check https://cdn.dreg.dev/package/picomatch@2.2.2
error: TS6054 [ERROR]: File 'https://cdn.dreg.dev/package/picomatch@2.2.2' has an unsupported extension. The only supported extensions are '.ts', '.tsx', '.d.ts', '.js', '.jsx'.
```

1.4.6
```
/m/f/P/g/l/d/example ❯❯❯ rm -rf $DENO_DIR
/m/f/P/g/l/d/example ❯❯❯ deno1.4.6 cache https://cdn.dreg.dev/package/picomatch@2.2.2
(snip)
Check https://cdn.dreg.dev/package/picomatch@2.2.2
(snip)
Found 144 errors.
```

cc @kitsonk

kitsonk:
Ok, I think I might know what the problem is... working on a fix.

kitsonk:
So this is related to #8024, and the TypeScript compiler not liking root modules without extensions, even though dependencies can be.  To make matters worse here, `curl -I https://cdn.dreg.dev/package/picomatch@2.2.2` outputs:

```
HTTP/2 200 
content-length: 46854
content-type: application/javascript; charset=utf-8
x-typescript-types: /package/@types/picomatch@2.2.1
vary: Accept-Encoding
cross-origin-resource-policy: cross-origin
access-control-allow-origin: *
access-control-expose-headers: *
etag: W/"46854-2a3655872b129d67a3c5f3ac60c"
date: Sat, 24 Oct 2020 20:19:40 GMT
server: Fly/bca74ec (2020-10-19)
via: 2 fly.io
```

And as you point out it was "broken" in 1.4.6 as it was type checking JavaScript, which it shouldn't be doing.

kitsonk:
It appears that #8114 has fixed this one.

```
> deno cache https://cdn.dreg.dev/package/picomatch@2.2.2
Download https://cdn.dreg.dev/package/picomatch@2.2.2
Download https://cdn.dreg.dev/polyfill/node/path.ts
Download https://cdn.dreg.dev/polyfill/node/process.ts
[...]

PR #8114, opened 2020-10-25
fix(cli): properly handle roots with extensions that don't match media types

### LINKED ISSUE #8024: Difficult to understand the error message when the extension is missing
I found it difficult to understand the error message when the script file was created without the extension.

## To Reproduce

```sh
deno --version
deno 1.4.6
v8 8.7.220.3
typescript 4.0.3
```

```sh
rm -rf ~/.cache/deno/*
```
```sh
cat <<EOT > mod
import * as log from "https://deno.land/std/log/mod.ts";

log.info("hello");
EOT
```
```sh
deno run mod
```

### Result

```
thread 'main' panicked at 'Cached source file doesn't exist', cli/global_state.rs:222:8
```

If there is no import statement, it will not be reproduced. (So it was even more confusing.)

It would be nice if we could read the error messages to see that the files specified in the deno run must have an extension.

kitsonk:
tsc doesn't like "extensionless" root modules (or ones with extensions that aren't what it expects).  So when a module is a root module, we have to compare it resolved media type with what the media type looks like just based on its string value, and if they don't match, we have to remap it to something `tsc` won't complain about.

This partially fixes #8024 and #8106, but exposes two other problems.  One is already raised with #8111 and I will raise another (which is that if don't have `checkJs` on, `tsc` does not emit further dependencies that are emittable, and result in the file being not emitted).

### DIFF
--- a/cli/module_graph2.rs
+++ b/cli/module_graph2.rs
@@ -763,8 +763,11 @@ impl Graph2 {
       info!("{} {}", colors::green("Check"), specifier);
     }
 
-    let root_names: Vec<String> =
-      self.roots.iter().map(|ms| ms.to_string()).collect();
+    let root_names: Vec<(ModuleSpecifier, MediaType)> = self
+      .roots
+      .iter()
+      .map(|ms| (ms.clone(), self.get_media_type(ms).unwrap()))
+      .collect();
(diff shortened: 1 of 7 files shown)


lucacasonato (approved):
LGTM. Thanks :-)

Good that we are collecting more test cases for this.

Issue #8096, opened 2020-10-23
Custom arg passing to Deno test?


wozniakty:
I've searched around google and github and can't seem to find info on this.  I'd like to be able to set up configurable tests with more than just a file by passing custom args into `deno test`.  The reason for this is we're trying to use Deno as an integration test runner against our API suite, and we'd like to do things like customize the environment-style info at execution time rather than with, say, environment variables.  Our goal is something like this

```
deno test prod // hits production URLs for smoke tests
deno test development //hits local services for more thorough testing
```

But from what I can tell, `deno test` only handles the specified `--` level flags and takes all other args as file names?  Am I missing something?

nayeemrmn:
I think we should be able to pass `Deno.args` with `deno test -- arg1 arg2`.

wozniakty:
@nayeemrmn doing so today produces this:
```
C:\Work\repos\public-api-testing>deno test -A api_tests.ts -- one two
error: Cannot resolve module "file:///C:/Work/repos/api-testing/one" from "file:///C:/Work/repos/api-testing/$deno$test.ts"
Imported from "file:///C:/Work/repos/api-testing/$deno$test.ts:2"
```

kitsonk:
Yeah, the lack of support is simply an oversight.

ross-weir:
What would be the preferred way to implement this?

`deno test arg`
vs
`deno test -- arg`

If we went with `deno test arg` I think we would need to force the user to supply the files using a `--files` flag.

lucacasonato:
`deno test -- arg` would be preferred.

ross-weir:
Thanks. I'll work on this one :-)

PR #8121, opened 2020-10-25
feat(cli): Pass script args for test command

ross-weir:
Implements #8096 

Allows for script arguments to be passed when running `deno test`.

For example:

`$ deno test test_file.ts -- arg1 arg2`

`Deno.args` will contain `["arg1", "arg2"]`.


kitsonk (approved):
LGTM

PR #8123, opened 2020-10-25
fix(cli): restore permission check on workers

### LINKED ISSUE #8120: No permission checks for worker startup
```ts
// foo.ts
new Worker(new URL("./worker.ts", import.meta.url).toString(), { type: "module" });
```

```ts
// worker.ts
console.log("foo");
```

### Actual

```
$ deno run foo.ts
foo
```

### Expected

```
$ deno run foo.ts
PermissionDenied: ...
```

This is only happens on master. This is a 1.5.0 blocker.

kitsonk:
Fixes #8120

### DIFF
--- a/cli/program_state.rs
+++ b/cli/program_state.rs
@@ -121,13 +121,19 @@ impl ProgramState {
     self: &Arc<Self>,
     specifier: ModuleSpecifier,
     target_lib: TargetLib,
-    dynamic_permissions: Permissions,
+    runtime_permissions: Permissions,
     is_dynamic: bool,
     maybe_import_map: Option<ImportMap>,
   ) -> Result<(), AnyError> {
     let specifier = specifier.clone();
+    // Workers are subject to the current runtime permissions.  We do the
+    // permission check here early to avoid "wasting" time building a module
+    // graph for a module that cannot be loaded.
+    if target_lib == TargetLib::Worker {
+      runtime_permissions.check_specifier(&specifier)?;
+    }
     let handler =
-      Rc::new(RefCell::new(FetchHandler::new(self, dynamic_permissions)?));
+      Rc::new(RefCell::new(FetchHandler::new(self, runtime_permissions)?));
     let mut builder =
       GraphBuilder2::new(handler, maybe_import_map, self.lockfile.clone());
     builder.add(&specifier, is_dynamic).await?;
--- a/cli/tests/error_worker_permissions_local.ts.out
+++ b/cli/tests/error_worker_permissions_local.ts.out
@@ -1,0 +1,2 @@
+[WILDCARD]
+error: Uncaught (in worker "") read access to "[WILDCARD]worker_types.ts", run again with the --allow-read flag
--- a/cli/tests/error_worker_permissions_local.ts.out
+++ b/cli/tests/error_worker_permissions_local.ts.out
@@ -1,2 +1,3 @@
 [WILDCARD]
 error: Uncaught (in worker "") read access to "[WILDCARD]worker_types.ts", run again with the --allow-read flag
+[WILDCARD]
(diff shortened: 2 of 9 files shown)


cli/tests/error_worker_permissions_local.ts.out
```diff
@@ -0,0 +1,2 @@
+[WILDCARD]
+error: Uncaught (in worker "") read access to "[WILDCARD]worker_types.ts", run again with the --allow-read flag
```
lucacasonato, line 2: This error message really isn't great. It don't actually specifically state that the permission was denied. Something we should address in a follow up PR.

cli/tests/error_worker_permissions_remote.ts
```diff
@@ -0,0 +1,5 @@
+const w = new Worker(
+  new URL("http://localhost:4545/cli/tests/subdir/worker_types.ts").toString(),
```
lucacasonato, line 2: The `new URL().toString()` isn't really needed here.

lucacasonato (approved):
LGTM. Nice to have tests for this now.

cli/tests/error_worker_permissions_local.ts.out
nayeemrmn: It should say `error: Uncaught (in worker "") PermissionDenied: read access to ...`. Not sure if this PR did something to omit it or it's an earlier regression.

cli/tests/error_worker_permissions_remote.ts
kitsonk: 🤦

cli/tests/error_worker_permissions_local.ts.out
kitsonk: It isn't this PR... I am not touching the error:

```rust
    if target_lib == TargetLib::Worker {
      runtime_permissions.check_specifier(&specifier)?;
    }
```

So very appropriate to fix in a follow up PR.

test_util/src/lib.rs
```diff
@@ -880,7 +880,7 @@ impl CheckOutputIntegrationTest {
     reader.read_to_string(&mut actual).unwrap();
 
     let status = process.wait().expect("failed to finish process");
-    let exit_code = status.code().unwrap();
+    let exit_code = status.code().unwrap_or(0);
```
lucacasonato, line 883: I don't like this. It can potentially swallow errors in tests.

kitsonk: Then why did this occur, only on Ubuntu? https://github.com/denoland/deno/runs/1306229988#step:14:3856

lucacasonato: I don't know. The status code is a `None` value when the process was terminated by a signal. See https://doc.rust-lang.org/std/process/struct.ExitStatus.html

It is not safe to assume that it exited cleanly (code 0) on a `None` code.

kitsonk: Ok, I have removed it... see what happens this time with CI.

kitsonk: Same problem: https://github.com/denoland/deno/pull/8123/checks?check_run_id=1306322095#step:14:3858

It isn't occurring locally on my mac, and I don't know how the test case could be invalid.

lucacasonato:
I'll test on Linux tomorrow morning. It looks like we aren't exiting cleanly.

cli/program_state.rs
```diff
@@ -121,13 +121,19 @@ impl ProgramState {
     self: &Arc<Self>,
     specifier: ModuleSpecifier,
     target_lib: TargetLib,
-    dynamic_permissions: Permissions,
+    runtime_permissions: Permissions,
```
bartlomieju, line 124: I like this name much better

bartlomieju (approved):
LGTM

cli/tests/error_worker_permissions_local.ts.out
nayeemrmn: I've kind of fixed this locally, but I think it's a wontfix. There are error `Results` throughout cli which are passed to JS in some cases and terminated upon in Rust in other cases (like this one), and there isn't really a good way of selectively unifying the output that always works the way we want it to. The issue has been present for a while, it's mostly fine as it is.

PR #8125, opened 2020-10-26
fix: std/node/buffer - Buffer.copy doesn't work as expected

Soremwar:
Also added tests that represent the expected behaviour from Node

### DIFF
--- a/std/node/buffer.ts
+++ b/std/node/buffer.ts
@@ -163,10 +163,16 @@ export default class Buffer extends Uint8Array {
       }
     }
 
-    const buffer = new Buffer(totalLength);
+    const buffer = Buffer.allocUnsafe(totalLength);
     let pos = 0;
-    for (const buf of list) {
-      buffer.set(buf, pos);
+    for (const item of list) {
+      let buf: Buffer;
+      if(!(item instanceof Buffer)){
+        buf = Buffer.from(item);
+      }else{
+        buf = item;
+      }
+      buf.copy(buffer, pos);
       pos += buf.length;
     }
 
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thank you @Soremwar, good fix and good tests

std/node/buffer.ts
```diff
@@ -247,7 +253,10 @@ export default class Buffer extends Uint8Array {
     sourceStart = 0,
     sourceEnd = this.length,
   ): number {
-    const sourceBuffer = this.subarray(sourceStart, sourceEnd);
+    const sourceBuffer = this
+      .subarray(sourceStart, sourceEnd)
+      .subarray(0, Math.max(0, targetBuffer.length - targetStart));
+
     targetBuffer.set(sourceBuffer, targetStart);
```
devcat, line 260: throws with `Buffer.alloc(5).copy(Buffer.alloc(0), 5)`

```suggestion
    if (sourceBuffer.length === 0) return 0;
    targetBuffer.set(sourceBuffer, targetStart);
```

devcat (changes requested):

std/node/buffer.ts
bartlomieju: Thanks for the catch @devcat

bartlomieju (dismissed):
Marking as `Request changes` based on feedback from @devcat

std/node/buffer.ts
Soremwar: Resolved and added test, thanks @devcat

Soremwar:
@bartlomieju Corrected and ready for review

bartlomieju (approved):
LGTM

PR #8126, opened 2020-10-26
feat(cli/repl): use completion candidate as hint

caspervonb:
This shows the first completion candidate as a hint, which is rendered as gray text to the right of the cursor.

### DIFF
--- a/cli/repl.rs
+++ b/cli/repl.rs
@@ -13,12 +13,13 @@ use regex::Regex;
 use rustyline::completion::Completer;
 use rustyline::error::ReadlineError;
 use rustyline::highlight::Highlighter;
+use rustyline::hint::Hinter;
 use rustyline::validate::ValidationContext;
 use rustyline::validate::ValidationResult;
 use rustyline::validate::Validator;
 use rustyline::Context;
 use rustyline::Editor;
-use rustyline_derive::{Helper, Hinter};
+use rustyline_derive::Helper;
 use std::borrow::Cow;
 use std::sync::mpsc::channel;
 use std::sync::mpsc::sync_channel;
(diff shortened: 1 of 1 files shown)


caspervonb:
Example

Screenshot 2020-10-26 at 10 12 30 AM

bartlomieju (changes requested):
Discussed offline, needs a test case, we'll land it for 1.6.0

lucacasonato (changes requested):
I like the concept, but there is one major issue. There is a significant lag when typing or removing characters. We seem to be blocking until we get the hint from V8. We should be acquiring the hint asynchronously, and not blocking on that.

Also it would be nice if you could use the up and down keys to navigate through the hints before pressing tab. That is not strictly required for this PR to land though.

caspervonb:
> I like the concept, but there is one major issue. There is a significant lag when typing or removing characters. We seem to be blocking until we get the hint from V8. We should be acquiring the hint asynchronously, and not blocking on that.

Ugh assumed the lag was my ssh connection but compared to bash on the same server indeed it is slow.
Unfortunately the entire rustyline interface is synchronous and callback based so hints have to block until resolved. There's no way to push the current hint at this time 😟 

Most likely won't end up landing this due to the input lag, but I'll see what can be done.

> Also it would be nice if you could use the up and down keys to navigate through the hints before pressing tab. That is not strictly required for this PR to land though.

Don't think this is something the rustyline editor is currently capable of (all the interesting bits are private), but can look into it.

bartlomieju:
@caspervonb seems we can't do much about it until https://github.com/kkawakam/rustyline/issues/126 is resolved. I'm going to provisionally close this PR, let's discuss alternate approaches on Discord.

caspervonb:
Yup, this introduces too much latency at the moment.

PR #8127, opened 2020-10-26
revert(std/jwt): reverting jwt so issues can be addressed

kitsonk:
Reverting #7991 and #8044 due to concerns discussed here: https://github.com/denoland/deno/pull/7991#issuecomment-715598404

We appreciate the contribution @timonson and @timreichen but we feel that the issues @panva has raised need to be addressed before this can be relanded.

Please merge on approval.


ry (approved):
Thank you for the helpful feedback @panva

timreichen:
@kitsonk
@timonson and I are working on a patch to address the issues.

PR #8130, opened 2020-10-26
chore: add bundle_no_check benchmark

lucacasonato:
This adds a `deno bundle --no-check` benchmark that can be compared with the `bundle` benchmark.

### DIFF
--- a/cli/bench/main.rs
+++ b/cli/bench/main.rs
@@ -93,6 +93,11 @@ const EXEC_TIME_BENCHMARKS: &[(&str, &[&str], Option<i32>)] = &[
     &["bundle", "std/examples/chat/server_test.ts"],
     None,
   ),
+  (
+    "bundle_no_check",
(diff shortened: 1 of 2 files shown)


cli/bench/main.rs
```diff
@@ -308,7 +313,7 @@ fn run_strace_benchmarks(
         deno_exe.to_str().unwrap(),
       ])
       .args(args.iter())
-      .stdout(Stdio::null())
+      .stdout(Stdio::inherit())
```
bartlomieju, line 316: Should we revert this one?

bartlomieju (approved):
LGTM

cli/bench/main.rs
lucacasonato: 🤷‍♂️. It's useful for debugging. Can remove though

bartlomieju: It's okay for me

PR #8131, opened 2020-10-26
Remove timezone dependent tests

ry:
cc @timreichen

### DIFF
--- a/std/datetime/test.ts
+++ b/std/datetime/test.ts
@@ -77,16 +77,6 @@ Deno.test({
 Deno.test({
   name: "[std/datetime] format",
   fn: () => {
-    // Date
-    assertEquals(
-      "2019-01-01",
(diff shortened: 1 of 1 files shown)


bartlomieju:
Supersedes https://github.com/denoland/deno/pull/7726?

ry:
Probably it's better to just remove timezone specific tests. Then it still passes when you run `deno test std/datetime`. I think Id prefer to land this one than the other. WDYT?

bartlomieju:
> Probably it's better to just remove timezone specific tests. Then it still passes when you run `deno test std/datetime`. I think Id prefer to land this one than the other. WDYT?

I agree with you on that one.

bartlomieju (approved):
LGTM

timreichen:
@ry @bartlomieju agree, `test` should not have a different tz than `run` imo.

PR #8133, opened 2020-10-26
feat(cli): add commit hash and target to long_version output

### LINKED ISSUE #6037: Display commit in deno version.
It's necessary to set the output of `deno --version` like the text below: 40419c664d402d4c3c5783d0ae337d23c106f619
```text
deno 1.0.3
build 40419c6<https://github.com/denoland/deno/commit/40419c664d402d4c3c5783d0ae337d23c106f619>
target x86_64-unknown-linux-gnu
v8 8.4.300
typescript 3.9.2
```
so we can easily refer to a specific version of source repository and target OS in an issue.

kt3k:
This PR adds the git commit hash and build target to long version output (`--version`) as suggested in #6037. The output now looks like the below:

```shellsession
$ ./target/debug/deno --version
deno 1.4.6 (0dcdc5e)
target x86_64-apple-darwin
v8 8.7.220.3
typescript 4.0.3
```

---
fixes #6037

### DIFF
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -111,10 +111,17 @@ fn git_commit_hash() -> String {
     .arg("HEAD")
     .output()
   {
-    std::str::from_utf8(&output.stdout[..7])
-      .unwrap()
-      .to_string()
+    if output.status.success() {
+      std::str::from_utf8(&output.stdout[..7])
+        .unwrap()
+        .to_string()
+    } else {
+      // When not in git repository
+      // (e.g. when the user install by `cargo install deno`)
+      "UNKNOWN".to_string()
+    }
   } else {
+    // When there is no git command for some reason
     "UNKNOWN".to_string()
   }
 }
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -105,15 +105,18 @@ fn ts_version() -> String {
 }
 
 fn git_commit_hash() -> String {
-  let output = std::process::Command::new("git")
+  if let Ok(output) = std::process::Command::new("git")
     .arg("rev-list")
     .arg("-1")
     .arg("HEAD")
     .output()
-    .expect("failed to execute process");
-  std::str::from_utf8(&output.stdout[..7])
-    .unwrap()
-    .to_string()
+  {
+    std::str::from_utf8(&output.stdout[..7])
+      .unwrap()
+      .to_string()
+  } else {
+    "UNKNOWN".to_string()
+  }
 }
 
 fn main() {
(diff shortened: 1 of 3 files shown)


cli/version.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
 pub const DENO: &str = env!("CARGO_PKG_VERSION");
+pub const BUILD: &str = env!("GIT_COMMIT_HASH");
```
ry, line 4: Just "GIT_COMMIT_HASH" instead of "BUILD" would be more descriptive.

cli/flags.rs
```diff
@@ -221,8 +221,10 @@ To evaluate code in the shell:
 
 lazy_static! {
   static ref LONG_VERSION: String = format!(
-    "{}\nv8 {}\ntypescript {}",
+    "{} ({})\ntarget {}\nv8 {}\ntypescript {}",
     crate::version::DENO,
+    crate::version::BUILD,
+    env!("TARGET"),
```
ry, line 227: It's a little odd that the target is on its own line. Maybe it should be in the parens too?

kt3k: How about this format?

```shellsession
$ deno --version
deno 1.4.6 (0dcdc5e, x86_64-apple-darwin)
v8 8.7.220.3
typescript 4.0.3
```

ry: I think that makes more sense. It would be nice to also know if we're using release/debug build.

cli/build.rs
```diff
@@ -104,6 +104,18 @@ fn ts_version() -> String {
...
+    .arg("rev-list")
+    .arg("-1")
+    .arg("HEAD")
+    .output()
+    .expect("failed to execute process");
+  std::str::from_utf8(&output.stdout[..7])
+    .unwrap()
+    .to_string()
+}
+
```
lucacasonato, line 118: Is this going to work when doing `cargo install deno`?

ry: It's a good point. I think it will break "cargo install deno".

kt3k: I see. It won't work with `cargo install`...

We need a different strategy here.

kt3k: It seems sha1 is stored in the file called `.cargo_vcs_info.json` in crates.io. (ref: https://github.com/rust-lang/cargo/issues/5629 ) Maybe we can use it.

lucacasonato: Maybe just return `"crates.io"` if anything in `git_commit_hash` fails?

kt3k:
Closing for the moment. I'll reopen it when I find a way to handle `cargo install deno` case correctly.

bartlomieju:
@kt3k I think this is a valid approach. As @lucacasonato suggested we can fallback to some constant value eg. empty string. For reference [`vergen`](https://github.com/rustyhorde/vergen) returns `UNKNOWN` if `git` command fails (https://github.com/rustyhorde/vergen/blob/8d522db8c8e16e26c0fc9ea8e6b0247cbf5cca84/src/output/mod.rs#L88-L95).

kt3k:
OK. I'll update the change!

kt3k:
I updated the change. The commit hash becomes `UNKNOWN` when not in git repository. The output looks like the below (I also added `PROFILE`):

```shellsession
$ ./target/debug/deno --version
deno 1.5.0 (UNKNOWN, debug, x86_64-apple-darwin)
v8 8.7.220.3
typescript 4.0.3
```

bartlomieju:
The solution looks satisfactory to me. Waiting on @ry to double check before landing.

ry (approved):
LGTM - thank you @kt3k !

PR #8134, opened 2020-10-26
fix: path traversal in std/http/file_server.ts

lucacasonato:
cc @hhc0null


ry:
I tried these tests on master, they appear to both pass

lucacasonato:
> I tried these tests on master, they appear to both pass

Should be fixed now.

ry (approved):
LGTM - thanks Luca. I confirmed that checkURIEncodedPathTraversal fails on master.

crypt096 (approved):

PR #8135, opened 2020-10-26
fix(cli/fmt): Strip "\\?\" prefix when displaying Windows paths

nayeemrmn:
Fixes https://discord.com/channels/684898665143206084/689420767620104201/770329467436400710.

We strip this prefix for `Deno.realPath()`, no reason not to do it for pretty diagnostics.

### DIFF
--- a/cli/fmt.rs
+++ b/cli/fmt.rs
@@ -80,17 +80,13 @@ async fn check_source_files(
             match diff(&file_text, &formatted_text) {
               Ok(diff) => {
                 info!("");
-                info!(
-                  "{} {}:",
-                  colors::bold("from"),
-                  file_path.display().to_string()
-                );
+                info!("{} {}:", colors::bold("from"), display_path(&file_path));
                 info!("{}", diff);
               }
               Err(e) => {
                 eprintln!(
                   "Error generating diff: {}",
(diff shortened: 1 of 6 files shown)


ry (commented):
I think this makes sense. But I wonder if there isn't a more general solution to be had. Surely we display paths in more places than in fmt.rs ?


(Caveat: I don't understand windows path semantics very well. I think this is ok.)

nayeemrmn:
> I wonder if there isn't a more general solution to be had.

I've replaced all `PathBuf::canonicalize()` calls (which add the prefix in the first place) with a new `canonicalize_path()` utility. It effectively shares the `Deno.realPath()` tests now.

maximousblk:
for more info on this path format: https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#unc-paths

piscisaureus:
Yeah I concur.

piscisaureus (approved):

piscisaureus:
Note that there are edge cases that this does not cover, e.g. the prefix `\\?\UNC` should probably be replaced with `\\` etc. And that's just unwrapping pandora's box without even opening it.

This patch makes it better in some cases, and does not make other cases worse, so that's enough reason to land it.

PR #8141, opened 2020-10-27
feat(std/wasi): add start method to Context

caspervonb:
This adds a start method to the Context to make starting a command less tedious and yield consistent errors.

Manually setting the memory is still valid for more complex scenarios, just undocumented for the time being (haven't decided yet if it will be public or not).

### DIFF
--- a/std/wasi/snapshot_preview1.ts
+++ b/std/wasi/snapshot_preview1.ts
@@ -270,7 +270,6 @@ export interface ContextOptions {
   args?: string[];
   env?: { [key: string]: string | undefined };
   preopens?: { [key: string]: string };
-  memory?: WebAssembly.Memory;
 }
 
 export default class Context {
@@ -285,7 +284,7 @@ export default class Context {
   constructor(options: ContextOptions) {
     this.args = options.args ? options.args : [];
     this.env = options.env ? options.env : {};
-    this.memory = options.memory!;
+    this.memory = null!;
 
     this.fds = [
       {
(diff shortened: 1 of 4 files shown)


std/wasi/snapshot_preview1.ts
```diff
@@ -1556,4 +1555,41 @@ export default class Context {
...
+      throw new TypeError("WebAsembly.instance must provide a memory export");
+    }
+
+    this.memory = memory;
+
+    if (typeof _initialize !== "function" && _initialize !== undefined) {
+      throw new TypeError(
+        "WebAsembly.instance export _initialize must not be defined",
+      );
+    }
```
ry, line 1585: Not sure but shouldn't it be something like this?
```js
  if (!_initialize) {
    throw new TypeError(
      "WebAsembly.instance export _initialize must not be defined",
    );
  }
```

caspervonb: Yikes.

Not quite but it is indeed wrong, should just be checking if _initialize is a function and throw if it is (there are two types of entry points, _initialize, or _start).

crypt096 (approved):

std/wasi/snapshot_preview1.ts
```diff
@@ -285,7 +284,7 @@ export default class Context {
   constructor(options: ContextOptions) {
     this.args = options.args ? options.args : [];
     this.env = options.env ? options.env : {};
-    this.memory = options.memory!;
+    this.memory = null!;
```
bartlomieju, line 287: `!` is not needed here?

caspervonb: >  memory: WebAssembly.Memory;		

It's actually nullable according to the type rules, it's forced to null here and assigned/validated later in the start method.

bartlomieju (approved):
LGTM

Issue #8146, opened 2020-10-27
bug(windows): prompt() adds "\r" to inputs and doesn't consider the default value


maximousblk:
to reproduce:

Linux (WSL) | Windows

![image](https://user-images.githubusercontent.com/28438021/97340972-761c3880-18aa-11eb-89d5-1a51d6ba1226.png)

isn't limited to repl

![image](https://user-images.githubusercontent.com/28438021/97341736-77019a00-18ab-11eb-980d-5ca55ae1b496.png)

pedropaulosuzuki:
There is another problem with the prompt, confirm and alert methods... they seem to not print correctly some characters (ãñõâêîôûü, for example)

![image](https://user-images.githubusercontent.com/32345702/97512662-da053680-1968-11eb-86ac-b2042a130248.png)

It happens both on the alert/confirm/prompt message and on the return value. This does not happen on console.log, though.

deno 1.5.0
v8 8.7.220.3
typescript 4.0.3

If this is too unrelated to the main problem, I can open a new issue

pedropaulosuzuki:
As #8146 can be fixed with something similar to a simple trim method, and the problem I brought with special UTF-8 (or UTF-16?) probably isn't that simple, I'm opening a new issue.

lucacasonato:
Fixed by #8149

PR #8149, opened 2020-10-27
fix(prompt): prompt works with windows eol and eof (#8146)

sanderhahn:
Fixes issue #8146

The test case seems to check the Windows EOL "\r\n", however this is an illusion because somehow the test runner already translates the Windows into Unix EOL. Have tested the functionality interactively and had to remove the `isatty` return null to do so.

### DIFF
--- a/cli/rt/41_prompt.js
+++ b/cli/rt/41_prompt.js
@@ -3,6 +3,7 @@
   const { stdin, stdout } = window.__bootstrap.files;
   const { isatty } = window.__bootstrap.tty;
   const LF = "\n".charCodeAt(0);
+  const CR = "\r".charCodeAt(0);
   const encoder = new TextEncoder();
   const decoder = new TextDecoder();
 
@@ -50,7 +51,20 @@
 
     while (true) {
       const n = stdin.readSync(c);
-      if (n === 0 || c[0] === LF) {
+      if (n === null || n === 0) {
+        break;
+      }
+      if (c[0] === CR) {
+        const n = stdin.readSync(c);
+        if (c[0] === LF) {
+          break;
+        }
+        buf.push(CR);
+        if (n === null || n === 0) {
+          break;
+        }
+      }
+      if (c[0] === LF) {
         break;
       }
       buf.push(c[0]);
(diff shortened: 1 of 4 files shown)


sanderhahn:
Ah i see that `isatty` is also used in the other functions... removed it because you can't use prompt and pipe from stdin.
What do you think? @kt3k

lucacasonato (commented):
Looks good. Could you add a test confirming the correct behaviour for both `alert` and `confirm` too?

cli/rt/41_prompt.js
```diff
@@ -31,10 +32,6 @@
   function prompt(message = "Prompt", defaultValue) {
     defaultValue ??= null;
 
-    if (!isatty(stdin.rid)) {
-      return null;
-    }
-
```
lucacasonato, line 37: Can you revert this change? It looks like is unrelated to the goal of this PR.

sanderhahn:
Done :+1:

lucacasonato:
Thanks. Could you add a test confirming the correct behaviour for both `alert` and `confirm` too?

sanderhahn:
Both these funtions call the same `readLineFromStdinSync` where the changes are made. As i said the test is only an illusion because it seems that "\r\n" is already converted to "\n" by the test runner. Couldn't pinpoint where this is happening but maybe its some rust code that already makes sure that EOL is cross platform. Have confirmed correct behavior by using an a tty and piped stdin because i don't have Windows available.

lucacasonato:
Ah ok. We are going to need to figure out what is converting removing the carriage return here. I'll take a look.

sanderhahn:
Maybe its possible to do binary io in the test runner...

sanderhahn:
Think i found it: https://github.com/denoland/deno/blob/master/test_util/src/lib.rs#L917

lucacasonato:
Oh we only test this test on unix. The test is prefixed with `#[cfg(unix)]`.

cc @bartlomieju want to land without working test?

lucacasonato:
@sanderhahn This test gets never actually run on Windows. The `pty` crate we are using is unix only. We need to find a different way to test this on windows.

ebebbington:
Subprocess to run a script (or one per prompt) and assert the output? probably not the best idea but just trying to offer any.
 help

lucacasonato (approved):
Discussed this offline. We are going to land this without a working Windows test, because setting up tty testing on Windows is out of scope for this PR.

Preemptively approving, but going to test on Windows that `https://deno.land/posts/v1.5/alert_confirm_prompt.js` example actually works. Thanks @sanderhahn

lucacasonato:
@wperron confirmed this is indeed working now.

PR #8151, opened 2020-10-27
fix(repl): don't hang on unpaired braces

### LINKED ISSUE #8150: repl: unpaired braces cause input never consumed
Entering a single ']' would cause repl to forever accepting new lines, due to that `ValidationResult::Invalid` would actually be consumed by the editor itself while continue building the lines. Instead we should mark it as `Valid` and send the bad input for evaluation to get the proper error from V8.

Example:

```
$ deno
> ]
(you can keep entering new line here, and it will never consume input until you Ctrl-C)
```

kevinkassimo:
Fixes #8150

Previously, entering a single ']' would cause repl to forever accepting new lines, due to that `ValidationResult::Invalid` would actually be consumed by the editor itself while continue building the lines. Instead we should mark it as `Valid` and send the bad input for evaluation to get the proper error from V8.

* Before:

```
> ]
(you can keep entering new line here, and it will never consume input
until you Ctrl-C)
```

* After:

```
> ]
Uncaught SyntaxError: Unexpected token ']'
>
```

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -1268,6 +1268,40 @@ fn repl_test_pty_multiline() {
   }
 }
 
+#[cfg(unix)]
+#[test]
+fn repl_test_pty_unpaired_braces() {
+  use std::io::Read;
+  use util::pty::fork::*;
+
+  let tests_path = util::tests_path();
+  let fork = Fork::from_ptmx().unwrap();
+  if let Ok(mut master) = fork.is_parent() {
+    master.write_all(b")\n").unwrap();
+    master.write_all(b"]\n").unwrap();
+    master.write_all(b"}\n").unwrap();
+    master.write_all(b"close();\n").unwrap();
+
+    let mut output = String::new();
+    master.read_to_string(&mut output).unwrap();
+
+    assert!(output.contains("Unexpected token ')'"));
+    assert!(output.contains("Unexpected token ']'"));
+    assert!(output.contains("Unexpected token '}'"));
+
+    fork.wait().unwrap();
+  } else {
+    util::deno_cmd()
+      .current_dir(tests_path)
+      .env("NO_COLOR", "1")
+      .arg("repl")
+      .spawn()
+      .unwrap()
+      .wait()
+      .unwrap();
+  }
+}
+
 #[test]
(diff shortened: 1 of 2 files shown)


cli/tests/integration_tests.rs
```diff
@@ -1516,6 +1516,21 @@ fn repl_test_eval_unterminated() {
   assert!(err.is_empty());
 }
 
+#[test]
```
caspervonb, line 1519: Validator only runs with tty so needs to be tested with pty.

kevinkassimo: pty is not suffering from the issue. Added a test anyways.

caspervonb: > pty is not suffering from the issue.

Hmm really? It should be, went through the call tree a few days ago in rustyline and the validator trait is not used at all when input is a file (e.g stdin).

kevinkassimo: Reason is this part of the code:
https://github.com/kkawakam/rustyline/blob/e9a5ccdc4f24c4a73053e6cf8897c186da5981da/src/lib.rs#L871-L876
For pty case, rustyline calls `readline_direct` which only considers newline and does not run validator, as you mentioned:
https://github.com/kkawakam/rustyline/blob/e9a5ccdc4f24c4a73053e6cf8897c186da5981da/src/lib.rs#L757-L764

caspervonb (approved):
LGTM, thank you very much @kevinkassimo 😁

ry (approved):
LGTM - thanks Kevin!

I confirmed that your test creates a hang on master.

ry:
@caspervonb This validator is basically implementing a parser. It's unfortunate that we're only able to test it with super high-level integration tests involving lots of I/O pty, fork etc. Is there some way to *unit* test this little parser? Ideally the fix here would have involved a simple unit test only in cli/repl.rs

caspervonb:
>  Is there some way to unit test this little parser? Ideally the fix here would have involved a simple unit test only in cli/repl.rs

We can just instantiate the helper and test that, even simpler if we extract the validator into a struct since that would require zero mocking.

But I think we should probably just bring in swc and use that to parse this instead, starting to hit a lot of cases where we need to do more thorough parsing so would be better to let a good parser handle it.

See also #8160

PR #8156, opened 2020-10-28
fix(std/wasi): remove stray console.log call

caspervonb:
A stray `console.log` call made it into path_open which should have been reverted in #8078 , it was only meant for debugging on the CI.


caspervonb:
Apologies, thought I had reverted them all before signing off on it.

lucacasonato (approved):
LGTM

PR #8162, opened 2020-10-28
tests(cli): add test for improper unicode encoding

### LINKED ISSUE #8161: Deno 1.5.0 bundle regression: bad Unicode character encoding
Bundling in deno `1.5.0` would produce invalid code, due to what appears to be bad Unicode encoding in the output.

Here's a quick reproducer:

```typescript
import { stringify } from "https://deno.land/std@0.75.0/encoding/yaml.ts";

console.log(stringify({
  foo: "bar",
}));
```
`1.5.0` bundle will contain:

image

as opposed to the same snippet in `1.4.6`

image

kitsonk:
Fixes #8161

(Or actually will fix when https://github.com/swc-project/swc/issues/1188 is fixed, this is a test case to help ensure it is fixed at the moment)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2379,9 +2379,9 @@ dependencies = [
 
 [[package]]
 name = "swc_ecma_codegen"
-version = "0.41.0"
+version = "0.41.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c48486c086354a033f5e5c4a9dc730ad0e7596039dd4d92f70fe56fcfec4f26a"
+checksum = "96e655391ee7b9e7e01fde2f64bfc1dbcf8b712510fc09edc6c00de52704e6a9"
 dependencies = [
  "bitflags",
  "num-bigint",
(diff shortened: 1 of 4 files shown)


bartlomieju:
@kitsonk please rebase, #8197 should include the fix

kitsonk:
@bartlomieju @kdy1 the test case is still failing because the output is not as expected/valid.  swc is still modifying the input string.

kdy1:
@kitsonk I'm considering an option in `swc_ecma_codegen` to preserve input or emit ascii only. How do you think about it?

kitsonk:
For us, we would always choose to preserve input, so yes.  I am not sure what the advantages of re-encoding the input would be.

kdy1:
This will be fixed by https://github.com/swc-project/swc/pull/1221

kdy1:
Can you try this again with the new version?

kitsonk:
@kdy1 it still "modifies" the strings, but the way it re-encodes them exhibits the same behaviour, so I have updated the expected output to match the behaviour of swc.

kdy1:
It should modify string anymore.

There's [a test](https://github.com/swc-project/swc/blob/512153f9445de9825f55fc554586920c3bb2603d/ecmascript/codegen/src/tests.rs#L435-L441) for the exact same input. 


https://github.com/denoland/deno/blob/686a17fc075ead774f5b692329d11e72139e3f02/Cargo.lock#L2380-L2382

It seems like deno is using old version of codegen. It's fixed by `v0.41.1`. Can you run `cargo update -p swc_ecma_codegen`?

bartlomieju (approved):
LGTM

Issue #8166, opened 2020-10-28
Transpile error with swc and decorators


kitsonk:
I have something similar issue:
```sh
RUST_BACKTRACE=1 deno bundle -c tsconfig.json src/egonet.ts egonet.bundle.js
```
```
thread 'main' panicked at 'cannot access a scoped thread local variable without calling `set` first', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/scoped-tls-1.0.0/src/lib.rs:168:9
stack backtrace:
   0: std::panicking::begin_panic
   1: swc_ecma_transforms::proposals::decorators::legacy::Legacy::handle::{{closure}}
   2: swc_ecma_transforms::proposals::decorators::legacy::Legacy::handle
   3: <swc_ecma_transforms::proposals::decorators::legacy::Legacy as swc_ecma_visit::Fold>::fold_decl
   4: swc_ecma_visit::Fold::fold_stmt
   5: <swc_ecma_transforms::proposals::decorators::legacy::Legacy as swc_ecma_visit::Fold>::fold_module
   6: swc_bundler::bundler::load::<impl swc_bundler::bundler::Bundler<L,R>>::load_transformed
   7: <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::next
   8: deno::bundle_command::{{closure}}
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  10: tokio::runtime::Runtime::block_on
  11: deno::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
Perhaps related to the use of decorators in my code. Easily reproducible by cloning my [repo](https://github.com/jveres/ego-network) and issue the `deno bundle` command above.

_Originally posted by @jveres in https://github.com/denoland/deno/issues/8148#issuecomment-717762983_

kitsonk:
@jveres it is a different issue than #8148 so I am splitting it out.

jveres:
thanks, I too realised so.

jveres:
Using version `1.5.1` `deno bundle test.js test.bundle.js` runs without error.
However running the bundled code by `deno run test.bundle.js` gives this error: 
`error: Uncaught ReferenceError: _applyDecoratedDescriptor is not defined`

lucacasonato:
@jveres Could you open a new issue?

jveres:
> @jveres Could you open a new issue?

#8212

PR #8168, opened 2020-10-28
fix: panic in bundler

bartlomieju:
Fixes https://github.com/denoland/deno/issues/8166

Need to come up with test

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -467,7 +467,11 @@ pub fn transpile_module(
     typescript::strip(),
     fixer(Some(&comments)),
   );
-  let module = module.fold_with(&mut passes);
+  let module = swc_common::GLOBALS.set(&Globals::new(), || {
+    helpers::HELPERS.set(&helpers::Helpers::new(false), || {
+      module.fold_with(&mut passes)
+    })
+  });
 
   Ok((source_file, module))
 }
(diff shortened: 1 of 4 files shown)


bartlomieju:
@kitsonk could you take over this PR?

kitsonk (commented):
Still needs a test... 😄

jveres:
test.ts

```typescript
function Decorator() {
  return function (
    target: Record<string, any>,
    propertyKey: string,
    descriptor: TypedPropertyDescriptor<any>,
  ) {
    const originalFn: Function = descriptor.value as Function;
    descriptor.value = async function (...args: any[]) {
      return await originalFn.apply(this, args)
    };
    return descriptor;
  };
}

class SomeClass {
  @Decorator()
  async test(): Promise<void> {}
}

new SomeClass().test();
```

```sh
deno bundle test.ts test.bundle.js      
Bundle file:///Users/U715864/temp/test.ts
Check file:///Users/U715864/temp/test.ts
thread 'main' panicked at 'cannot access a scoped thread local variable without calling `set` first', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/scoped-tls-1.0.0/src/lib.rs:168:9
```

bartlomieju:
Thanks @jveres I added your test case

jveres:
```typescript
// deno-lint-ignore-file no-explicit-any
```
should be added to the top of `test.ts` to pass linter checks (cc @bartlomieju)

jveres:
thanks @bartlomieju, now I hope it find its way into `1.5.1`

crypt096 (approved):

kitsonk (approved):
LGTM

PR #8178, opened 2020-10-29
feat: Enable Deno APIs by default in Workers

### LINKED ISSUE #8174: Enable Deno namespace by default in Workers
Remove `WorkerOptions::deno`, though it will be brought back with a different type once permissions forking (#4867) is implemented.

nayeemrmn:
Closes #8174.


lucacasonato:
This is a breaking change because it would change the default permissions of `new Worker(). Previously you did not have inherited FS access for example.

crypt096 (approved):

bartlomieju:
I'm in favor of this change, I especially like how it simplifies `WebWorker` setup in Rust (it's useful for #8381)

nayeemrmn:
@bartlomieju Please review.

lucacasonato:
@nayeemrmn This is on hold because we can't land until 2.0.

nayeemrmn:
> Previously you did not have FS access by default (if parent has permission) for example.

Hmm... was this ever an intended security mechanism? It was always arbitrary that workers inherited net permissions with `fetch()`, but not fs permissions as a side-effect of the functions being hidden.

The only cases where this provides new-found e.g. read permissions are where the runtime was granted `--allow-read`. This change fully preserves the integrity of our security flags. We didn't reasonably assure anything else.

lucacasonato:
Intended or not, it is in stable, so we have to take this into consideration before merging. People might be using workers for sandboxing.

bartlomieju:
@nayeemrmn I'm in favor of landing this PR, but I think Luca has a point. Let's discuss further on Discord

ry:
I'm closing this PR because it's stale and there's a bunch of merge conflicts. But we're going to discuss this issue soon and try to come to some consensus on proceeding.
Thanks for the patch @nayeemrmn - we've got the diff here still so we can always refer back. 

To be continued....

bartlomieju:
@nayeemrmn can you open a new PR with this change? It's still highly desired and now that we're eyeballing 2.0 it'd be great to have it.

PR #8186, opened 2020-10-29
refactor: don't spin up V8 for `deno cache`

lucacasonato:
This cleans up TODO in `cache_command`. Flamegraph shows that we spend a lot of time starting up V8.


bartlomieju (approved):
LGTM :)

Issue #5620, opened 2020-05-19
Runtime compiler API doesn't require valid specifier


bartlomieju:
https://doc.deno.land/https/raw.githubusercontent.com/denoland/deno/master/cli/js/lib.deno.unstable.d.ts#Deno.compile

`Deno.compile()` has an option to take a map of source file. In this option specifiers for files can be written as `/foo.ts`, `/bar.js`. It's against other Deno APIs which require valid specifiers and required to add some hackery to make it work:
https://github.com/denoland/deno/blob/1be7ec47ac3b6e63be2b03245a26e060b8268b5e/cli/module_graph.rs#L172-L181

https://github.com/denoland/deno/blob/1be7ec47ac3b6e63be2b03245a26e060b8268b5e/cli/js/compiler.ts#L605-L617

Discussed this situation offline with @ry are there are two potential solutions:
- require valid specifiers for all files - each specifiers must be a valid URL
- add `baseUrl` option to `Deno.compile()` which will be used to create URLs for each specifier

CC @kitsonk

kitsonk:
They are in theory "abstract" resources... requiring them to be them seems quite illogical to me and doesn't sound like good DX.  Why even provide the `baseUrl` option?  Again, that is just to keep things "tidy" on the inside, and bad DX to force a user to do something that means nothing to them.  I didn't catch it in the PR, but now I understand the problem.  It feels like we could more systematically deal with it by just mangling all the `sources` when they are provided like that without requiring a user to specify a `baseUrl` and basically tell them "please fake this information because we don't want to".

Also, this particular cleanup might want to wait for #4752 (or we move forward with that) and we would have a discreet API for providing sources, instead of the more conflated one and it maybe easier to resolve the modules in a clean way.

bartlomieju:
Resolved in #8192

PR #8192, opened 2020-10-30
refactor(cli): migrate runtime compile/bundle to new infrastructure

### LINKED ISSUE #8060: Migrate Deno.compile/Deno.bundle to new tsc/module_graph
We need to migrate `Deno.compile()`/`Deno.bundle()` to the new compiler infrastructure.

kitsonk:
Fixes #8060 

Couple notes:

- There will be cleanup post this, specifically in the snapshot creation of the compiler and some refactoring in the `99_compiler.js`.
- There are a couple changes in behaviour.
  - Even when sources are provided, the resulting source keys are the text version of module specifiers.  So for example if you send in sources of `/foo.ts` it will return with `file:///foo.ts.js` and `file:///foo.ts.js.map`.  This makes it a lot easier to deal with everything internally.  Users can provide sources with any URI that was as well.  It should also make it easier to feed externally gathered sources to the APIs.
  - The compiler option of `types` was strange to begin with for the compiler APIs and I think we need revisit supporting that all together anyways.


ry:
glad to see cli/module_graph.rs go! can cli/tsc.rs be removed too?

cli/tsc/99_main_compiler.js
```diff
@@ -716,317 +611,6 @@ delete Object.prototype.__proto__;
...
-      }
-      default:
-        throw new Error(
-          `!!! unhandled CompilerRequestType: ${request.type} (${
-            CompilerRequestType[request.type]
-          })`,
-        );
-    }
-  }
-
```
ry, line 1029: great to see this stuff go too

kitsonk:
Oops. `cli/tsc.rs` should have been removed too. I messed up during a rebase.

cli/module_graph2.rs
```diff
@@ -560,6 +571,36 @@ pub struct CheckOptions {
   pub reload: bool,
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum BundleType {
+  /// Return the emitted contents of the program as a single "flattened" ES
+  /// module.
+  #[allow(unused)]
```
lucacasonato, line 578:
```suggestion
```

cli/op_fetch_asset.rs
```diff
@@ -85,6 +83,7 @@ pub fn get_asset(name: &str) -> Option<&'static str> {
 
 /// Warning: Returns a non-JSON op dispatcher. Must be manually attached to
 /// JsRuntime.
+#[allow(unused)]
```
lucacasonato, line 86: Is this actually unused?

kitsonk: Yeah, I meant to comment on that, it is required for the time being, as it is used by the snapshot builder, but not by the `main.rs` anymore, so it errors as unused if removed.  A subsequent refactor of this whole thing needs to be done to clean up the building of the snapshot, and then having some assets on demand.

core/module_specifier.rs
```diff
@@ -65,6 +65,18 @@ impl ModuleSpecifier {
...
+    if self.0.scheme() == "file" {
+      if let Ok(path) = self.0.to_file_path() {
+        path
+      } else {
+        PathBuf::from(self.0.path())
+      }
+    } else {
+      PathBuf::from(self.0.path())
+    }
+  }
```
bartlomieju, line 78: I'm not happy extending API of `ModuleSpecifier`... It seems this method is only used in single place, so I suggest to just do this logic there.

cli/tsc_config.rs
```diff
@@ -52,37 +52,42 @@ impl fmt::Display for IgnoredCompilerOptions {
...
+  "skipDefaultLibCheck",
+  "skipLibCheck",
   "sourceMap",
   "target",
+  "types",
+  "useDefineForClassFields",
 ];
 
-const IGNORED_RUNTIME_COMPILER_OPTIONS: [&str; 50] = [
+const IGNORED_RUNTIME_COMPILER_OPTIONS: [&str; 40] = [
```
bartlomieju, line 79: Consider changing type to `&[&str]`, then you won't have to specify number of elements with each change.

cli/tests/lib_runtime_api.ts
```diff
@@ -1,12 +1,12 @@
 const [errors, program] = await Deno.compile(
-  "main.ts",
+  "/main.ts",
   {
-    "main.ts": `document.getElementById("foo");`,
+    "/main.ts": `document.getElementById("foo");`,
```
bartlomieju, line 4: I'm strongly in favor of this change - this will resolve https://github.com/denoland/deno/issues/5620

cli/program_state.rs
```diff
@@ -120,7 +115,7 @@ impl ProgramState {
   pub async fn prepare_module_load(
     self: &Arc<Self>,
     specifier: ModuleSpecifier,
-    target_lib: TargetLib,
+    lib: TypeLib,
```
bartlomieju, line 118: Nitpick: use `type_lib`? A bit more descriptive than `lib`

cli/module_loader.rs
```diff
@@ -21,23 +21,23 @@ pub struct CliModuleLoader {
   /// When flags contains a `.import_map_path` option, the content of the
   /// import map file will be resolved and set.
   pub import_map: Option<ImportMap>,
-  pub target_lib: TargetLib,
+  pub lib: TypeLib,
```
bartlomieju, line 24: Nitpick: `s/lib/type_lib`

cli/module_graph2.rs
```diff
@@ -70,6 +71,16 @@ lazy_static! {
     Regex::new(r#"(?i)\stypes\s*=\s*["']([^"']*)["']"#).unwrap();
 }
 
+type EmitResult = Result<
+  (
+    HashMap<String, String>,
+    Stats,
+    Diagnostics,
+    Option<IgnoredCompilerOptions>,
+  ),
```
bartlomieju, line 80: Consider adding struct for this purpose (`EmitOutput`?)

cli/module_graph2.rs
```diff
@@ -560,6 +571,35 @@ pub struct CheckOptions {
...
+impl Default for BundleType {
+  fn default() -> Self {
+    BundleType::None
+  }
+}
+
+#[derive(Debug, Default)]
+pub struct EmitOptions {
+  /// Indicate the form the result of the emit should take.
+  pub bundle_type: BundleType,
```
bartlomieju, line 595: To be honest I find having both compilation and bundling done via single API configured by this field a little hard to grep, I had already reworked TSC once to split them into separate functions. 

@kitsonk how do you feel about having separate APIs for this purpose? Eg. `ModuleGraph::emit` and `ModuleGraph::emit_bundle`

cli/module_graph2.rs
```diff
@@ -845,6 +835,155 @@ impl Graph2 {
...
+      "esModuleInterop": true,
+      "experimentalDecorators": true,
+      "isolatedModules": true,
+      "jsx": "react",
+      "lib": TypeLib::DenoWindow,
+      "module": "esnext",
+      "strict": true,
+      "target": "esnext",
+    }));
+    let opts = match options.bundle_type {
```
bartlomieju, line 853: In relation to splitting APIs for compile/bundle - if you agree to have separate methods for those purposes I suggest to make those "default" configs `lazy_static` defined at the top of the file.

cli/module_graph2.rs
```diff
@@ -845,6 +835,155 @@ impl Graph2 {
...
+      } else {
+        None
+      };
+
+    let root_names = self.get_root_names();
+    let hash_data =
+      vec![config.as_bytes(), version::DENO.as_bytes().to_owned()];
+    let graph = Rc::new(RefCell::new(self));
+
+    let response = exec(
```
bartlomieju, line 880: Nitpick: it's not obvious what's `exec`, could you type it as `tsc::exec`?

cli/module_graph2.rs
```diff
@@ -845,6 +835,155 @@ impl Graph2 {
...
+            specifiers.len(),
+            1,
+            "An unexpected number of specifiers associated with emitted file."
+          );
+          let specifier = specifiers[0].clone();
+          let extension = match emitted_file.media_type {
+            MediaType::JavaScript => ".js",
+            MediaType::SourceMap => ".js.map",
+            _ => unreachable!(),
+          };
```
bartlomieju, line 926: Won't this bit cause double extensions for specifiers with extensions?

Eg. `file://foo/bar/bazz.js` -> `file://foo/bar/bazz.js.js`, `file://foo/bar/bazz.js.js.map`

bartlomieju (commented):
Great work @kitsonk 🚀  I have only a handful of low-impact comments. Can't wait to land this one!

core/module_specifier.rs
kitsonk: It is only in one place at the moment, but the refactor of the `file_fetcher` also uses it.  It has been one of the biggest frustration I have had of the refactor is that we move from `ModuleSpecifier` to `Url` to `PathBuf` to `Path` to `String`, causing all manner of confusion and indirection.  If we need a `ModuleSpecifier` in a different format, we should do it in one place to ensure it is clear where the logic sits.

cli/program_state.rs
kitsonk: I have been trying my best to ensure consistent naming.  It is one of the things I found frustrating, is that things keep getting renamed as the move along.  This directly feeds the `lib` property in `TsConfig`, so I named it `lib`.  I agree it can bit a little bit less self documenting here, but if you follow it through the code, it is named the same everywhere you see it, instead of seeing `type_lib` in one place and `lib` in another. 🤷

cli/module_graph2.rs
kitsonk: That would make it inconsistent with the other APIs.  Maybe a `ResultInfo` that has the stats, diagnostics, and maybe_ignored_options as a struct, since that is the common bit and then a `Result<(HashMap<String, String>, ResultInfo), AnyError>`.

cli/module_graph2.rs
kitsonk: I would prefer not to, because I am of a strong opinion that we should have a single runtime API now of `Deno.emit()` which takes a bundle.  It also makes things easier/cleaner when we drop the emit from `tsc` in lieu of `swc`.

cli/module_graph2.rs
kitsonk: Yeah, but it is consistent... because if we replace the extension, we could end up in situations where the input specifier == emit specifier, and if you as a user are using local specifiers and then taking the emit and writing them out to disk yourself, you might overwrite your sources.  I would prefer to make it clear in the documentation that this occurs versus running the risk of someone overwriting their input.

cli/program_state.rs
bartlomieju: Fair enough, let's keep `lib`

cli/module_graph2.rs
bartlomieju: That sounds good to me 👍

cli/module_graph2.rs
bartlomieju: SGTM 👍

cli/module_graph2.rs
bartlomieju: Okay, I'll hold off and circle back after getting myself more familiar with the API

core/module_specifier.rs
kitsonk: I will move it for now... 😛

kitsonk: Actually looking at it again, it really should be just internal to `MediaType`... which I have made it now.

bartlomieju (approved):
LGTM

Issue #8185, opened 2020-10-29
Deno.fsync marked as stable, but needs --unstable flag (1.5.0)


jannes:
When running a script that makes use of Deno.fsync I get the following error:  
`TypeError: Deno.fsync is not a function`  
(`deno run --allow-write --allow-read deno_scripts/testing_fsync.js `)  

Passing the `--unstable` flag resolves this.
(`deno run --allow-write --allow-read --unstable deno_scripts/testing_fsync.js`)  
However with 1.5.0 fsync/fdatasync are marked as stable in the documentation.

bartlomieju:
CC @caspervonb

PR #8193, opened 2020-10-30
fix: make Deno.fsync(), Deno.fdatasync() stable

bartlomieju:
These APIs were stabilized in 1.5.0, but were
never added to the "stable" Deno namespace. This
commit fixes that.

Fixes https://github.com/denoland/deno/issues/8185
CC @caspervonb


lucacasonato:
Can you also check if they were removed from the list of diagnostic errors that we rewrite?

lucacasonato (approved):
Looks like it was. LGTM

bartlomieju:
@lucacasonato it was, just not moved to stable namespace

caspervonb:
Ahh right, thanks @bartlomieju

Issue #8189, opened 2020-10-29
deno bundle fails to bundle specific minified code


lucacasonato:
```
// main.ts
import { compress } from "https://deno.land/x/lz4/mod.ts";
const text = new TextEncoder().encode("X".repeat(64));
console.log(text.length);
console.log(compress(text).length);
```

```
$ deno bundle foo.ts | deno run --no-check -
Bundle file:///mnt/f9/Projects/github.com/denoland/deno/main.ts
Check file:///mnt/f9/Projects/github.com/denoland/deno/main.ts
64
error: Uncaught TypeError: Cannot read property 'set' of null
    return g().set(A1, Q1 / 1), B = A1.length, Q1;
^
    at Q (<file:///mnt/f9/Projects/github.com/denoland/deno/$deno$stdin.ts>:10:1)
    at lz4_compress (<file:///mnt/f9/Projects/github.com/denoland/deno/$deno$stdin.ts>:20:1)
    at compress (<file:///mnt/f9/Projects/github.com/denoland/deno/$deno$stdin.ts>:59:1)
    at <file:///mnt/f9/Projects/github.com/denoland/deno/$deno$stdin.ts>:63:1
```

The code that fails to bundle as expected is in https://raw.githubusercontent.com/denosaurs/deno_lz4/master/wasm.js. Formatted version here: https://gist.github.com/lucacasonato/338868d143d8c734a8c82ee375d862d6

cc @kdy1

Issue #8188, opened 2020-10-29
Deno bundle 1.5.0 redeclaring symbols


aricart:
In my code base I have a single definition for `render`. When the bundling the library, I get multiple:

The original function becomes renamed `render2`:
```javascript
function render2(frame) {
    const cr = "␍";
    const lf = "␊";
    return TD.decode(frame).replace(/\n/g, "␊").replace(/\r/g, "␍");
}
```

Then on line 5042:
```javascript
const render1 = render2;
```

Then on line 5968:
```javascript
const render1 = render;
```

Of course this is wrong, and on a browser environment, the generated module immediately fails to load: "Identifier 'render1' has already been declared.

bartlomieju:
CC @kdy1

kitsonk:
@aricart can you provide steps to reproduce?

aricart:
@kitsonk I don't have a simple example but this one works:
```sh
deno upgrade --version 1.5.0
 deno bundle --unstable https://raw.githubusercontent.com/nats-io/nats.ws/master/src/mod.ts | grep render | grep render
...
function render2(frame) {
const render1 = render2;
const render1 = render;
                    console.info(`> ${render(yields[i1])}`);
                console.info(`< ${render1(frame)}`);
                console.error(`!!! ${render(frame)}: ${err}`);

```

```sh
deno bundle --unstable https://raw.githubusercontent.com/nats-io/nats.ws/master/src/mod.ts | grep render
Bundle https://raw.githubusercontent.com/nats-io/nats.ws/master/src/mod.ts
    function render(frame) {
    exports_44("render", render);
        "render": true,
                    "render": util_ts_11_1["render"],
                                console.info(`> ${nats_base_client_ts_1.render(yields[i])}`);
                            console.info(`< ${nats_base_client_ts_1.render(frame)}`);
                            console.error(`!!! ${nats_base_client_ts_1.render(frame)}: ${err}`);
export const render = __exp["render"];
```

It is worth noting the mod.ts: 
```typescript
export * from "./nats-base-client.ts";
export { connect } from "./connect.ts";
```
and the ./nats-base-client.ts:
```typescript
export * from "https://raw.githubusercontent.com/nats-io/nats.deno/v1.0.0-11/nats-base-client/internal_mod.ts";
```

aricart:
A bit more explanation, the nats.ws project depends on the common code currently hosted in nats.deno

Issue #8180, opened 2020-10-29
Deno bundle producing incorrect code in 1.5.0


giggio:
I just tried running a simple http app with Deno, and after bundling I get an undefined error:

````
error: Uncaught TypeError: Cannot read property 'Continue' of undefined
        (void 0).Continue,
                 ^
    at file:///proj/dist/hello1.js:616:18
````

This is the source:

````ts
import { serve } from "https://deno.land/std@0.75.0/http/server.ts";

const port = 8000;
const host = "0.0.0.0";
const server = serve(`${host}:${port}`);
const body = new TextEncoder().encode("Hello, World!\n");

async function main() {
  for await (const request of server) {
    console.log(`Got a ${request.method} at ${request.url}.`);
    request.respond({ body });
  }
}
console.log(`Listening at http://${host}:${port}.`);

main();
````

This is a snippet, starting at line 614 of the generated .js:

````js
const STATUS_TEXT = new Map([
    [
        (void 0).Continue,
        "Continue"
    ],
    [
        (void 0).SwitchingProtocols,
        "Switching Protocols"
    ],
````

You can see all the `(void 0)...` there.

This are the commands I ran:

````
$ deno cache --lock deno-lock.json --lock-write src/hello1.ts
$ deno bundle --config tsconfig.json --lock deno-lock.json src/hello1.ts dist/hello1.js
````

I tried an empty `tsconfig.json`, it still fails.

This works with Deno `1.4.6`. This is a working version compiled with `1.4.6`:

````js
            exports_11("STATUS_TEXT", STATUS_TEXT = new Map([
                [Status.Continue, "Continue"],
                [Status.SwitchingProtocols, "Switching Protocols"],
````

`deno run` works in `1.5.0`.

This was ran on a Ubuntu machine. Let me know if you need more info.

bartlomieju:
CC @kdy1

kdy1:
@bartlomieju Does deno invoke some transforms of swc? I'm not sure where `(void 0)` came from.

bartlomieju:
@kdy1 this is function that TS files are run through prior to bundling: https://github.com/denoland/deno/blob/b0482400c96476946514f03f27f139dd505b06fb/cli/ast.rs#L411-L473

Issue #8148, opened 2020-10-27
Deno bundle error in 1.5.0: thread 'main' panicked at 'not implemented: namespaced re-export


jsejcksn:
This worked without issue in `1.4.6`:

```
deno bundle src/main.ts $DENO_INSTALL_ROOT/bin/example.bundle.js
```

Now, I am getting this error:

```
Bundle file:///redacted/path/to/src/main.ts
Check file:///redacted/path/to/src/main.ts
thread 'main' panicked at 'not implemented: namespaced re-export: local=path#42, all=true', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/swc_bundler-0.12.0/src/bundler/chunk/export.rs:219:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jsejcksn:
@kitsonk I see you added the "needs info" label. Can I help by providing any specific info (sanitized of details about my code/computer)?

lucacasonato:
This looks like an upstream SWC issue. `@kdy1` has mid term exams right now, so let's wait with pinging him about it for a few more days :-)

bartlomieju:
@jsejcksn providing source code would be very helpful anyway

jsejcksn:
Hi @bartlomieju. Unfortunately, I'm not able to share any source code from this repository... it is a private project.

kitsonk:
Can you please try to reduce it into something reproducible?  Or please provide the output of `deno bundle -L debug` sanitized, otherwise we are just guessing at what is causing the issue.  It maybe an swc issue, it maybe that we need to do something before handing it over to swc.  Without a minimal reproduction it is just guessing...

lucacasonato:
@kitsonk I don't think that is necessary. @jsejcksn has just run into a `unimplemented!` in swc: https://github.com/swc-project/swc/blob/master/bundler/src/bundler/chunk/export.rs#L219.

kitsonk:
@lucacasonato we have 15 test cases having to do with bundling (and likely found another half dozen which we didn't write up) that ensure defects we encountered in swc are legitimate and valid...  Just because something isn't implemented in swc doesn't mean it is valid, and a minimal reproduction would a) allow swc to fix the issue and b) give everyone a test case to make sure it doesn't happen again.

jveres:
I have something similar issue:
```sh
[...]

PR #8197, opened 2020-10-30

bartlomieju:
Fixes https://github.com/denoland/deno/issues/8180
Fixes https://github.com/denoland/deno/issues/8188
Fixes https://github.com/denoland/deno/issues/8189
Fixes https://github.com/denoland/deno/issues/8148

### DIFF
--- a/cli/tests/bundle/fixture11.out
+++ b/cli/tests/bundle/fixture11.out
@@ -6,9 +6,9 @@ var O;
     O1[O1["A"] = 0] = "A";
     O1[O1["B"] = 1] = "B";
     O1[O1["C"] = 2] = "C";
-})((void 0) || (O = {
+})(O || (O = {
 }));
-const O1 = void 0;
+const O1 = O;
 export { O1 as O };
 class A {
     #a;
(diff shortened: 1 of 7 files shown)


cli/tests/bundle/fixture11.out
```diff
@@ -6,9 +6,9 @@ var O;
     O1[O1["A"] = 0] = "A";
     O1[O1["B"] = 1] = "B";
     O1[O1["C"] = 2] = "C";
-})((void 0) || (O = {
+})(O || (O = {
 }));
-const O1 = void 0;
+const O1 = O;
```
bartlomieju, line 11: @kitsonk please take a look - the result of this test changed after the upgrade, I'm not sure if that's ok

bartlomieju: Also CC @kdy1

kdy1: I think it's ok. Actually previous output is a bug.

Previously exported `O` was undefined, but it should be `O1`, which is enum.

kitsonk: Yeah, it was a big but we didn't notice it.

kitsonk (approved):
LGTM

Issue #4862, opened 2020-04-23
[WIP] Arm64/Aarch64 ci builds


sunsetkookaburra:
Will create a PR when I have this working and cleaned up, at the moment the repo-history is a mess. Utilising some of the work from rusty_v8's arm 64 ci.yml done by afinch7. I'm trying out a community gh-action that allows running on non-intel cpus by utilising docker. I'm also looking into cargo runners, which I think provide the list of executables which can be stored and run on this docker instance. 🤞

sunsetkookaburra:
i'm somewhat close after two solid days of head scratching. i ran a built debug version on a RPi3 and got an error relating to V8.
here's my work so far https://github.com/sunsetkookaburra/deno/tree/arm64_ci
main problem now is that the github actions instance runs out of space when building it as debug

sunsetkookaburra:
Feeling like I've hit somewhat of a dead end myself, and don't know enough about rust and the build process to do this justice. Passing this on to whoever wants to work on this.
I've come close with sunsetkookaburra/deno@b09ef29 and with sunsetkookaburra/deno@a248afb

sunsetkookaburra:
![image](https://user-images.githubusercontent.com/55334808/80269914-5e93a580-86f7-11ea-8243-6b20fece3c05.png)
This image shows the result when running the binary on Ubuntu 18.04LTS on a RPi3

sunsetkookaburra:
![image](https://user-images.githubusercontent.com/55334808/80269968-bdf1b580-86f7-11ea-823f-d83188d312a7.png)

sunsetkookaburra:
for anyone who knows something about linux binary builds:
![image](https://user-images.githubusercontent.com/55334808/80270065-61db6100-86f8-11ea-84b9-1b2d463d4c0a.png)

sunsetkookaburra:
if it helps, `deno --version` is able to run successfuly.

sunsetkookaburra:
(for my reference, first round of attempts: sunsetkookaburra/deno@2ef273)

afinch7:
If you can get v8 to compile successfully the only other major issue you are going to run into is snapshots. Snapshots generated using a x86 compiles of v8 are not compatible with arm compiles of v8. It took me a long time to figure this out the first time around, but It makes more sense when you realize that a snapshot is mostly compiled bytecode from v8's JIT compiler. The only solution I managed to get working here is just don't use snapshots for arm builds, but I would be happier if we could generate arm snapshots.

[...]

PR #8198, opened 2020-10-30
WIP try to compile for aarch64-unknown-linux-gnu

lucacasonato:


afinch7:
Your going to need to either disable snapshots or find a way to build them using a arm64 executable. See discussion in #4862.

lucacasonato:
I see. I wonder if we should try to get some native arm64 runners. I feel that is more reliable than using QEMU anyway.

PR #8199, opened 2020-10-30
fix(prompt): fix prompt output of non-ascii characters on windows

### LINKED ISSUE #8179: New prompt, alert and confirm methods get confused when escaping special characters [ã]
The new adition of prompt, alert and confirm methods are great news for all of us!! Glad to have those!

However, after the small #8146 issue, there is yet another unrelated problem with the new tools: special UTF-8 (maybe UTF-16?) characters (ãñõâêîôûü, for example):

image

It happens both on the alert/confirm/prompt message and on the return value. This does not happen on console.log, though. May be related or similar to old #5046.

deno 1.5.0
v8 8.7.220.3
typescript 4.0.3

kt3k:
`Deno.stdout.writeSync(new TextEncoder().encode(str))` doesn't seem working for non-ascii characters on windows, but `Deno.core.print(str)` does seem working. See the comment https://github.com/denoland/deno/issues/8179#issuecomment-719431130 .

So this causes the issue #8179.

This PR replaced the use of `Deno.stdout.writeSync` in prompt with `Deno.core.print` and avoided the issue.

I checked the fix on window server 2019 machine.
スクリーンショット 2020-10-30 23 55 05

fixes #8179

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -11,6 +11,7 @@ use futures::future::FutureExt;
 use rusty_v8 as v8;
 use std::cell::Cell;
 use std::convert::TryFrom;
+use std::io::{stdout, Write};
 use std::option::Option;
 use url::Url;
 use v8::MapFnTo;
@@ -352,8 +353,10 @@ fn print(
   };
   if is_err {
     eprint!("{}", str_.to_rust_string_lossy(tc_scope));
+    stdout().flush().unwrap();
   } else {
     print!("{}", str_.to_rust_string_lossy(tc_scope));
+    stdout().flush().unwrap();
   }
 }
 
--- a/cli/rt/41_prompt.js
+++ b/cli/rt/41_prompt.js
@@ -1,18 +1,18 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 ((window) => {
-  const { stdin, stdout } = window.__bootstrap.files;
+  const { stdin } = window.__bootstrap.files;
   const { isatty } = window.__bootstrap.tty;
   const LF = "\n".charCodeAt(0);
   const CR = "\r".charCodeAt(0);
-  const encoder = new TextEncoder();
   const decoder = new TextDecoder();
+  const core = window.Deno.core;
 
   function alert(message = "Alert") {
     if (!isatty(stdin.rid)) {
       return;
     }
 
-    stdout.writeSync(encoder.encode(`${message} [Enter] `));
+    core.print(`${message} [Enter] `, false);
 
     readLineFromStdinSync();
   }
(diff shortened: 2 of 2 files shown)


core/bindings.rs
```diff
@@ -352,8 +353,10 @@ fn print(
   };
   if is_err {
     eprint!("{}", str_.to_rust_string_lossy(tc_scope));
+    stdout().flush().unwrap();
   } else {
     print!("{}", str_.to_rust_string_lossy(tc_scope));
+    stdout().flush().unwrap();
```
kt3k, line 359: I added flush here because print! macro doesn't seem flushing automatically. https://github.com/rust-lang/rust/issues/23818

wperron:
Validated on Windows
```text
PS C:\Users\wperron\github.com\kt3k\deno> .\target\release\deno
Deno 1.5.0
exit using ctrl+d or close()
> prompt("ÈÉÊàààà");
ÈÉÊàààà
null
> prompt("ÈÉÊàààà hello");
ÈÉÊàààà hello
null
> prompt("äÄăåČćçĎʤʣĎĔĤʞĿŁÒÔ")
äÄăåČćçĎʤʣĎĔĤʞĿŁÒÔ
null
>
```

core/bindings.rs
bartlomieju: @kt3k this is a bit strange - `Deno.core.print()` works for me without a problem without the `flush()` and it's one of purposes of `Deno.core.print` not to include new line by default. Could you explain the reason for this change?

kt3k: @bartlomieju `print!` is line-buffered as discussed in https://github.com/rust-lang/rust/issues/23818. So if we use it like `Deno.core.print('hello')`, the string doesn't show in stdout until new line is printed or process is finished. (Try the script `Deno.core.print('xyz'); setTimeout(() => {}, 4000);` for example).

If we don't add `flush()` here, `prompt` works like the below:

```
> prompt('hi!')
hello
hi! "hello"
```

(The answer and question are reversed here.) I think this isn't what the users expect.

If we add `flush()`, it works like:

```
> prompt('hi!')
hi! hello
"hello"
```

lucacasonato: @kt3k I can not reproduce that:
image

kt3k: In repl, the returned value is printed with line break (such as "1\n") and this line break should trigger flushing of stdout.

Try the following:

```
deno eval "Deno.core.print('xyz'); setTimeout(() => {}, 4000);"
```

piscisaureus:
It'd be nice if stdout (when attached to a terminal) actually behaved as expected and interpreted the byte stream as UTF8-encoded text.

But this fix is also legit, and the thing it fixes is really an eyesore.

So LGTM from me.

piscisaureus (approved):

Issue #8201, opened 2020-10-30
TextEncoder.encode TypeError: Invalid argument should give more details


Noah-Huppert:
I believe that the `TypeError` returned by `TextEncoder.encode` is too ambiguous.  I think for any other method which isn't dealing with text encoding an ambiguous error message like `Invalid argument` would be okay. However text encoding is full of many hairy input situations, and more details to narrow down which bad input situation you are experiencing would be very helpful. 

---

Here is a short story highlighting where just a little more detail in the error message could have helped a lot:

I had an error where some code before a call to `encode()` was accidentally returning objects instead of strings. It was hard for me to find because I even had Typescript annotations on the argument I was passing to `encode()` asserting that the argument should be a string, like so: 
```ts
let foo = shouldReturnAStringButReturnedAnObject();

function whichCallsEncodeWithArg(arg: string) {
  // arg never touched in any way until here
  new TextEncoder.encode(arg);
}

whichCallsEncodeWithArg(foo);
```

However type checks were passed and the error occurred. 

This small self contained example effectively reproduces passing an object to `encode()`:

```ts
new TextEncoder().encode({});
```

The resulting error is something like:

```ts
error: Uncaught TypeError: Invalid argument
    at TextEncoder.encode (deno:op_crates/web/08_text_encoding.js:1066:21)
    at main.ts:171:27
    at Array.forEach (<anonymous>)
    at main.ts:170:8
```

Due to the ambiguaty of the error, and my confidence is what turned out to be false assumptions about the type of an argument, I went down a long rabbit hole trying to figure out if my program was producing any invalid unicode glyphs or something like that. I think it would be helpful for `encode()` to throw an error more like `TypeError: Invalid argument, must be a string`.

kitsonk:
Actually we should be more spec compliant here...

```ts
new TextEncoder().encode({})
```

Results in the following in Chrome:

```
Uint8Array(15) [91, 111, 98, 106, 101, 99, 116, 32, 79, 98, 106, 101, 99, 116, 93]
```

It is only values that are not coercible to strings that throw.  For example:

```ts
new TextEncoder().encode(Symbol("a"));
```

Results in:

```
Uncaught TypeError: Cannot convert a Symbol value to a string
```

We should leave the type enforcement up to TypeScript.

Noah-Huppert:
[...]

PR #8205, opened 2020-10-31
tests(cli): add event target tests

benjamingr:
Lint, format and test passes.

This is my first attempt since the rewrite/refactor to contribute code to Deno 🎉 

I want to contribute most of the tests I've been adding in Node and work on some other stuff.

Note that I've found the behaviour of running tests pretty confusing:

 - I couldn't find a document explaining how to run the tests and had to look in CI, clicking "contributing" points to the style guide.
 - When running a single test it failed with an error - it would _probably_ make sense for the deno binary built to add `--unstable` automatically to the unit tests somehow? In Node we have a mechanism where the first line of the file can be a comment indicating what flags to pass like `// --expose-internals`
 - Running the linter/formatter resulted in a weird "OSError: [Errno 2] No such file or directory" error (but exit status code), so I _assume_ it passed but I have no idea. I can investigate this further.

PR #8206, opened 2020-10-31
fix(encoding): make textencoder work with forced non-strings

benjamingr:
Make TextEncoder behave like the browser's TextEncoder if the user passed a non-string and add a test for it, uses the fix 
suggested by @kitsonk

Fixes: https://github.com/denoland/deno/issues/8201

As with https://github.com/denoland/deno/pull/8205 please be "brutal" with me and give me any feedback that will make future contributions easier from the project's side :]


kitsonk (approved):
LGTM

Issue #5638, opened 2020-05-19
APMs in Deno -or- Async hooks and isolate->SetPromiseHook?


benjamingr:
Hey,

One of the features Node uses from V8 is promise hooks and in particular Node's [async_hooks](https://nodejs.org/api/async_hooks.html) lets users keep context enabling a lot of use cases (like APMs).

I saw [the last issue](https://github.com/denoland/deno/issues/945) but it doesn't look like there was a lot of discussion there. As far as I'm aware there is no analogy in the DOM/browser world for this. When I talked to some tooling people in browser vendors they said they were not aware of a direct analogy.

I would recommend (barring actual discussions with APM vendors or telling users to use a user promise library and instrumenting its scheduler): 

 - Exposing SetPromiseHook pretty directly and letting users and vendors deal with it.
 - Exposing an async_hooks like API - possibly based on Zones (or hopefully another web proposal).
 - Exposing _only_ the node async hooks API under std/node.
 - Wait for vendors and complain and not solve it until APM vendors or users complain (then use their feedback to pick an API).
 - Something else entirely.

It's true async_hooks are experimental and domains are deprecated but without neither - it is pretty impossible to build an APM for Node. You have to instrument methods to know "where you came from" in an async context and APMs are pretty valuable tools.

kitsonk:
Hmmm... fine grained understanding of async processes seems like something that would be desirable.  The analogue in the browser is what is available in the dev tools, it is seldom exposed in user land (the exception would be the `Performance` APIs).

We have `Deno.metrics` which is sort of in this space, but what is talked about where is a broader API.

PatrickJS:
Please consider Zone.js API for Deno
https://gist.github.com/mhevery/63fdcdf7c65886051d55

benjamingr:
Hey @ry @kitsonk - I was speaking to a large'ish company today and this issue was the show-stopper for them in their case.

I'm happy to help and work on this. Is there a preference for _how_ this should look like? I think the clearest API would be:
 - Exposing SetPromiseHook pretty directly and letting users and APM vendors deal with it.

````js
Deno.metrics.setPromiseHook
```
That exposes 
```
Isolate::SetPromiseHook
```

Also see the [API document](https://docs.google.com/document/d/1rda3yKGHimKIhg5YeoAmCOtyURgsbTH_qaYR79FELlk/edit) in case you are not too familiar with it.

[...]

PR #8209, opened 2020-10-31
feat: expose setPromiseHook support

benjamingr:
This is a draft PR to expose promise hooks enabling APMs. 

**My rust is bad, I haven't used it in ±2 years, any criticism and feedback about the code below is very appreciated**.

Still missing and need to do:

 - [ ] Document `Deno.core.setPromiseHook` 
 - [x] Refactor to use [js_recv_obj](https://github.com/denoland/deno/issues/5638#issuecomment-719939722) to avoid the global (currently exposed).
 - [x] Expose the constants used by the promise hooks (INIT, RESOLVE, BEFORE, AFTER)
 - [x] Add tests (copy the Node or V8 promise tests probably?):
   - Test `then`ables
   - Test an async function
   - Test a "raw" promise
   - Test a promise subclass
 - [ ] Get APM vendor feedback in order to ensure this API "looks good".
 - [x] Run formatter and linter

Nice to have in the future but out of scope:
 - Shim `async_hooks` using this in deno_std
 - Guide on how async hooks work

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -52,6 +52,9 @@ lazy_static! {
       v8::ExternalReference {
         function: get_proxy_details.map_fn_to()
       },
+      v8::ExternalReference {
+        function: set_promise_hook.map_fn_to()
+      },
     ]);
 }
 
@@ -188,6 +191,29 @@ pub fn initialize_context<'s>(
     get_proxy_details_key.into(),
     get_proxy_details_val.into(),
   );
+
+  // Promise hooks
+  let set_promise_hook_key = v8::String::new(scope, "setPromiseHook").unwrap();
+  let set_promise_hook_tmpl =
+    v8::FunctionTemplate::new(scope, set_promise_hook);
+  let set_promise_hook_val = set_promise_hook_tmpl.get_function(scope).unwrap();
+  core_val.set(
+    scope,
+    set_promise_hook_key.into(),
+    set_promise_hook_val.into(),
+  );
+
+  for (i, hook_name) in
+    ["INIT", "RESOLVE", "BEFORE", "AFTER"].iter().enumerate()
+  {
+    let promise_hook_key = v8::String::new(scope, hook_name).unwrap();
+    let promise_hook_value = v8::Integer::new(scope, i as i32);
+    set_promise_hook_val.set(
+      scope,
+      promise_hook_key.into(),
+      promise_hook_value.into(),
+    );
+  }
 
   let shared_key = v8::String::new(scope, "shared").unwrap();
   core_val.set_accessor(scope, shared_key.into(), shared_getter);
@@ -755,6 +781,52 @@ pub fn module_resolve_callback<'s>(
   }
 
   None
+}
+
+// Sets the V8 promise hook, allowing tracking of context across
+// Promise chains and async functions
+fn set_promise_hook(
+  scope: &mut v8::HandleScope,
+  args: v8::FunctionCallbackArguments,
+  _rv: v8::ReturnValue,
+) {
+  let hook = match v8::Local::<v8::Function>::try_from(args.get(0)) {
+    Ok(val) => val,
+    Err(_) => {
+      let msg = v8::String::new(scope, "Invalid argument").unwrap();
+      let exception = v8::Exception::type_error(scope, msg);
+      scope.throw_exception(exception);
+      return;
+    }
+  };
+
+  let state_rc = JsRuntime::state(scope);
+  let mut state = state_rc.borrow_mut();
+
+  state.js_promise_hook_cb.push(v8::Global::new(scope, hook));
+  scope.set_promise_hook(wrapped_hook);
+}
+extern "C" fn wrapped_hook(
+  type_: v8::PromiseHookType,
+  promise: v8::Local<v8::Promise>,
+  parent: v8::Local<v8::Value>,
+) {
+  let scope = &mut unsafe { v8::CallbackScope::new(promise) };
+  let context = promise.creation_context(scope);
+  let scope = &mut v8::ContextScope::new(scope, context);
+  let global = context.global(scope);
+
+  let state_rc = JsRuntime::state(scope);
+
+  let hook_type: v8::Local<v8::Value> =
+    v8::Integer::new(scope, type_ as i32).into();
+
+  for js_promise_hook_cb_handle in state_rc.borrow().js_promise_hook_cb.iter() {
+    js_promise_hook_cb_handle
+      .get(scope)
+      .call(scope, global.into(), &[hook_type, promise.into(), parent])
+      .unwrap();
+  }
 }
 
 // Returns promise details or throw TypeError, if argument passed isn't a Promise.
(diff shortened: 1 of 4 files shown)


benjamingr:
@bartlomieju I've changed the code to use the same machinery as js_recv_cb (I hope) and it works. Would you mind taking a look and making sure I understood it correctly? (to be clear: none of this is urgent)

benjamingr:
Question: I noticed I can't find references to other Deno.core stuff in the docs - where (at all?) should this be documented?

bartlomieju:
> @bartlomieju I've changed the code to use the same machinery as js_recv_cb (I hope) and it works. Would you mind taking a look and making sure I understood it correctly? (to be clear: none of this is urgent)

@benjamingr yeah, looks great!

bartlomieju:
> Question: I noticed I can't find references to other Deno.core stuff in the docs - where (at all?) should this be documented?

`Deno.core` does not have type declarations at the moment - this is on purpose as we consider `Deno.core` unstable and that it can change at any moment. Let's ponder at this a bit more and decide what to do next when this PR is ready to land.

benjamingr:
After a comment by @Flarna I am convinced the correct behaviour here is to allow multiple consumers - so I will replace the `Option` to a `Vec` and add a test for multiple consumers.

benjamingr:
Also, on second thought I don't think it makes sense to unit test the behaviour in various cases I mentioned - but rather just our API surface (that is, test multiple consumers but not the exact promise-hook behaviour of promise subclasses for example) since it's tested on the V8 side.

I would like to (eventually) see an integration test (with a complex promise context tracking use case) added but I think the tests I've added probably cover the basic cases.

benjamingr:
Ok, this is about ready for review now @bartlomieju thanks for the help!

I feel strongly that this should be documented _somewhere_ though, even if it just points to the V8 docs. I will happily write docs for it if you tell me where to put them :]

kitsonk:
@benjamingr @bartlomieju my opinion on the documentation is that we should move `embedding_deno` to a new _Advanced_ section in https://github.com/denoland/deno/tree/master/docs and add another document there about "Performance Monitoring" with this being the initial content.

core/bindings.rs
```diff
@@ -189,6 +192,29 @@ pub fn initialize_context<'s>(
...
+    v8::FunctionTemplate::new(scope, set_promise_hook);
+  let set_promise_hook_val = set_promise_hook_tmpl.get_function(scope).unwrap();
+  core_val.set(
+    scope,
+    set_promise_hook_key.into(),
+    set_promise_hook_val.into(),
+  );
+
+  for (i, hook_name) in
+    ["INIT", "RESOLVE", "BEFORE", "AFTER"].iter().enumerate()
```
bnoordhuis, line 207:
```suggestion
  use v8::PromiseHookType as Ty;
  for (hook_name, value) in
    [("INIT", Ty::Init), ("RESOLVE", Ty::Resolve), ("BEFORE", Ty::Before), ("AFTER", Ty::After)].iter()
```
(and maybe sort them)

If the original name is good enough you can also do:
```rust
for v in [Ty::After, Ty::Before, Ty::Init, Ty::Resolve].iter() {
  let name = format!("{:?}", v);
  // ...
}
```

core/bindings.rs
```diff
@@ -757,6 +783,52 @@ pub fn module_resolve_callback<'s>(
...
+  args: v8::FunctionCallbackArguments,
+  _rv: v8::ReturnValue,
+) {
+  let hook = match v8::Local::<v8::Function>::try_from(args.get(0)) {
+    Ok(val) => val,
+    Err(_) => {
+      let msg = v8::String::new(scope, "Invalid argument").unwrap();
+      let exception = v8::Exception::type_error(scope, msg);
+      scope.throw_exception(exception);
+      return;
```
bnoordhuis, line 799: For a future PR: consider moving this into a helper function. I count 5 or 6 other instances of the exact same pattern.

core/bindings.rs
```diff
@@ -757,6 +783,52 @@ pub fn module_resolve_callback<'s>(
...
+
+  let state_rc = JsRuntime::state(scope);
+
+  let hook_type: v8::Local<v8::Value> =
+    v8::Integer::new(scope, type_ as i32).into();
+
+  for js_promise_hook_cb_handle in state_rc.borrow().js_promise_hook_cb.iter() {
+    js_promise_hook_cb_handle
+      .get(scope)
+      .call(scope, global.into(), &[hook_type, promise.into(), parent])
```
bnoordhuis, line 827: Maybe use `v8::undefined().into()` instead of `global.into()`? Passing `global` is not wrong and it seems to be the predominant pattern in deno_core but it's marginally less efficient.

core/bindings.rs
```diff
@@ -757,6 +783,52 @@ pub fn module_resolve_callback<'s>(
...
+      .unwrap();
```
bnoordhuis, line 828: You should probably not unwrap here unless you want exceptions from the promise hook to terminate the process (with a fairly opaque error message.)

bnoordhuis (commented):
LGTM insofar I'm able to judge deno_core changes.

There's no mechanism to remove a promise hook again. Personally, I'm fine with leaving that out for now but I thought I'd mention it anyway.

brillout:
Any news on this? (I'm interested in porting https://github.com/reframejs/wildcard-api to Deno but for that I need async hooks.)

benjamingr:
I am unfortunately in the process of switching jobs and moving houses and working on kids so I won't be able to work on this soon - other people are welcome to pick this up in the meantime.

brillout:
Ok I see, thanks for the update.

bartlomieju:
@bnoordhuis @benjamingr maybe we should wait for the `SetPromiseHooks(init, before, after, resolve)` API? (referenced in https://docs.google.com/document/d/1vtgoT4_kjgOr-Bl605HR2T6_SC-C8uWzYaOPDK5pmRo and https://github.com/nodejs/node/pull/36394)

benjamingr:
It most likely should

bartlomieju:
> It most likely should

Great, I'll close this PR for now, as it became stale. Once `SetPromiseHooks()` lands in V8 we can use this PR as a reference point for the implementation. Thank you @benjamingr!

benjamingr:
That sounds correct.

jayphelps:
@bartlomieju Is this the hook? https://source.chromium.org/chromium/_/chromium/v8/v8.git/+/c0fceaa0669b39136c9e780f278e2596d71b4e8a

bartlomieju:
Yes, I believe it is

bartlomieju:
@benjamingr are you still interested in this functionality? I think we could try to tackle this once more with the API linked above.

benjamingr:
@bartlomieju I think Deno isn't actually production-ready until this API exists because it means it is impossible to write performance monitoring and analysis tools - it means users don't actually know where their production apps are spending time reliably which is a must for big real-world apps in my experience.

So yes - since I would like to see Deno used for production apps I am interested in this :) I've had Deno apps running "in production" since 2018 but I haven't been able to use it for anything substantial because of the lack of monitoring capabilities. 

That said, I don't have time to work on this and I think your best approach to get this ready quickly would be to talk to APM vendors like New Relic or Elasticsearch and ask them to implement this. It's probably a good idea to have a discussion with APM vendors since they're the ones likely to eventually use this API the most.

Alternatively Deno can choose not to expose this functionality to end-users directly and provide its own APM offering which is also fine probably.

PR #8215, opened 2020-10-31
(feat): Add configurable permissions for Workers

### LINKED ISSUE #4867: discussion: Permissions for Deno Worker
With #4784 landed, we can now create new workers with `Deno` namespace available. Permissions for newly created worker are inherited from the process. It's very desirable to be able to configure permissions more granularly (eg. to run a sandboxed worker with no permissions).

Proposed change to `WorkerOptions`
```typescript
export interface WorkerOptions {
  type?: "classic" | "module";
  name?: string;
  deno?: {
    permissions: {
      read?: boolean;
      readWhitelist?: string[];
      write?: boolean;
      writeWhitelist?: string[];
      net?: boolean;
      netWhitelist?: string[];
      env?: boolean;
      hrtime?: boolean;
      run?: boolean;
    }
  };
}
```

However I'm not sure how this API should behave if one of the fields in `permissions` interface is omitted - should it inherit from parent or maybe it should be set to false/empty list?

Example:
```
$ deno -A mod.ts

// mod.ts
new Worker("./worker.ts", {
  type: "module",
  deno: {
    permissions: {
      read: true,
      write: true,
    }
  }
});
```

My feeling is that all omitted permissions should be set to false - otherwise it'd be surprising that worker has access to net/run/hrtime because it's inherited from parent.

Of course, trying to specify permissions that escalate permissions of process will result in error:
```
$ deno --allow-read mod.ts
Uncaught error: PermissionDenied, trying to create worker with permissions escalating process' permissions.

// mod.ts
new Worker("./worker.ts", {
  type: "module",
  deno: {
    permissions: {
      read: true,
      write: true,
    }
  }
});
```

Soremwar:
- [x] Currently `net`, `write` and `read` are being passed to the worker op as an array, when it should be an array or a boolean (for absolute permissions)
- [x] The instantiation of the worker requires permissions over the worker itself. Find a way to decouple instantiation permissions and worker permissions 
- [x] Add tests
- [x] Add docs for the worker permission system

This closes #4867


Soremwar:
This is currently failing cause the worker needs two separate types of permissions in order to run, the main thread permissions needed to read the worker file (read access) and the worker permissions that restrict access within the worker, however the current implementation only receives one set of permissions. Sadly my current knowledge of Deno internals is now enough to address this issue properly :/

bartlomieju:
> This is currently failing cause the worker needs two separate types of permissions in order to run: the main thread permissions needed to read the worker file (read access) and the worker permissions that restrict access within the worker, however the current implementation only receives one set of permissions. Sadly my current knowledge of Deno internals is now enough to address this issue properly :/

@Soremwar this problem is way more complex than I initially expected. I'll discuss it with the rest of the core team and get back to you.

Soremwar:
@bartlomieju Roger that

cli/main.rs
```diff
@@ -108,7 +108,7 @@ fn create_web_worker_callback(
       || program_state.flags.coverage;
     let maybe_inspector_server = program_state.maybe_inspector_server.clone();
 
-    let module_loader = CliModuleLoader::new_for_worker(program_state.clone());
+    let module_loader = CliModuleLoader::new_for_worker(program_state.clone(), args.permissions.clone());
```
bartlomieju, line 111: Use current parent permissions here

cli/module_loader.rs
```diff
@@ -130,7 +133,12 @@ impl ModuleLoader for CliModuleLoader {
     let state = op_state.borrow();
 
     // The permissions that should be applied to any dynamically imported module
-    let dynamic_permissions = state.borrow::<Permissions>().clone();
+    let dynamic_permissions = if self.permissions.is_some() {
+      self.permissions.clone().unwrap()
+    } else {
```
bartlomieju, line 138:
```
let dynamic_permissions = if let Some(permissions) = self.permissions.take() {
      self.permissions.clone().unwrap()
    } else {
...
}
```

Soremwar: I don't get what you mean by this, why would I need to take out the permissions and left self.permissions empty?

cli/main.rs
```diff
@@ -106,6 +106,7 @@ impl From<Flags> for PermissionsOptions {
 
 fn create_web_worker_callback(
   program_state: Arc<ProgramState>,
+  parent_thread_permissions: Permissions,
```
bartlomieju, line 109: This is not right, `create_web_worker_callback` is called at the start of the whole program which means it will include permissions from startup instead of current permissions parent thread has. This field should be added to the `args` below (`CreateWebWorkerArgs` struct)

cli/module_loader.rs
```diff
@@ -118,7 +124,14 @@ impl ModuleLoader for CliModuleLoader {
     let state = op_state.borrow();
 
     // The permissions that should be applied to any dynamically imported module
-    let dynamic_permissions = state.borrow::<Permissions>().clone();
+    let dynamic_permissions =
+      if let Some(permissions) = self.permissions.as_ref().take() {
```
bartlomieju, line 128: `.as_ref()` is superfluous

cli/tests/workers_test.ts
```diff
@@ -357,3 +357,213 @@ Deno.test({
...
+    worker.postMessage(null);
+
+    await promise;
+    fail("Should throw on permissions denied");
+    worker.terminate();
+  } catch (error) {
+    if (!(error instanceof Deno.errors.PermissionDenied)) {
+      fail(`Unexpected error: ${error}`);
+    }
+  }
```
bartlomieju, line 568: Add assert that `error` is thrown, otherwise we might get silent failures

cli/module_loader.rs
Soremwar: How can I replace it? Borrowing self as mut requires me to change the original ModuleLoader implementation as well

bartlomieju: You're right. Then it suggest that current version also doesn't work as expected...

bartlomieju: Is there a test that checks if dynamic imports in worker use worker's permission instead of parents?

Soremwar: No, it only checks if it has read permission with `Deno.permissions.query`. If it's a separate matter, I'll add a test

cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -658,6 +658,25 @@ interface WorkerEventMap extends AbstractWorkerEventMap {
...
+ * to limit the access of the worker
+ */
+type WorkerPermissionListOption = "inherit" | boolean | Array<string>;
+
+/**
+ * "inherit" will take the permissions of the thread the worker is created in
+ * You can provide a list of routes relative to the file the worker is created in
+ * to limit the access of the worker
+ */
+type WorkerPermissionPathOption = "inherit" | boolean | Array<string | URL>;
```
nayeemrmn, line 678: I think these should be inlined. The names and implied categorisations (boolean vs list vs path) are a bit awkward and don't properly represent the relationship between path allowlists and other types of allowlists. It's much better to document these per permission type. It also makes the docs easier to follow.

Soremwar: Ok, I'll add the docs above to the worker jsdoc then

Soremwar:
This is ready for review

cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -712,7 +721,18 @@ declare class Worker extends EventTarget {
...
+        namespace?: boolean;
+        permissions?: {
+          env?: "inherit" | boolean;
+          hrtime?: "inherit" | boolean;
+          net?: "inherit" | boolean | string[];
+          plugin?: "inherit" | boolean;
+          read?: "inherit" | boolean | Array<string | URL>;
+          run?: "inherit" | boolean;
+          write?: "inherit" | boolean | Array<string | URL>;
+        };
```
nayeemrmn, line 734: When individual fields aren't specified, that's `"inherit"` by default right? And that's the same for the `permissions` field as a whole? So is the `"inherit"` value needed at all?

```suggestion
        permissions?: {
          env?: boolean;
          hrtime?: boolean;
          net?: boolean | string[];
          plugin?: boolean;
          read?: boolean | Array<string | URL>;
          run?: boolean;
          write?: boolean | Array<string | URL>;
        };
```

runtime/ops/worker_host.rs
```diff
@@ -86,21 +93,350 @@ pub fn init(
...
+fn check_write_permissions(
+  allow_list: &HashSet<PathBuf>,
+  current_permissions: &Permissions,
+) -> bool {
+  allow_list
+    .iter()
+    .all(|x| current_permissions.check_write(&x).is_ok())
+}
+
+fn merge_unary_permissions(
```
nayeemrmn, line 209:
```suggestion
fn merge_read_write_permissions(
```

cli/dts/lib.deno.shared_globals.d.ts
Soremwar: I'd rather have a explicit way of telling the worker it should inherit the permissions. Undefined/ignoring the field might be confusing for some users. @bartlomieju What do you think?

nayeemrmn: Perhaps, but my point was that it's inconsistent to how the `permissions` field is handled. Should that have an `"inherit"` value?

Soremwar: You are right, and I think that it should

Soremwar: In fact I don't know why it isn't in the d.ts, the API code already expects "inherit" as a valid input

cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -712,7 +721,18 @@ declare class Worker extends EventTarget {
        * hello world2
        *
        */
-      deno?: boolean;
+      deno?: {
+        namespace?: boolean;
+        permissions?: "inherit" | {
+          env?: "inherit" | boolean;
+          hrtime?: "inherit" | boolean;
+          net?: "inherit" | boolean | string[];
```
bartlomieju, line 729: A bit strange that `read` and `write` accept `URL` but `net` doesn't

cli/module_loader.rs
```diff
@@ -22,6 +22,7 @@ pub struct CliModuleLoader {
   /// import map file will be resolved and set.
   pub import_map: Option<ImportMap>,
   pub lib: TypeLib,
+  pub permissions: Option<Permissions>,
```
bartlomieju, line 25: rename to `initial_permissions` and add doc comment about purpose of this field

runtime/ops/worker_host.rs
```diff
@@ -1,32 +1,31 @@
...
-use deno_core::serde_json::Value;
-use deno_core::BufVec;
-use deno_core::ModuleSpecifier;
-use deno_core::OpState;
-use deno_core::ZeroCopyBuf;
-use serde::Deserialize;
+use deno_core::serde::de::{self, SeqAccess};
+use deno_core::serde::{Deserialize, Deserializer};
+use deno_core::serde_json::{self, json, Value};
+use deno_core::{BufVec, ModuleSpecifier, OpState, ZeroCopyBuf};
```
bartlomieju, line 15: Use one import per line, otherwise git diffs are hard to read, unfortunately `rustfmt` doesn't enforce that

cli/dts/lib.deno.shared_globals.d.ts
nayeemrmn: Maybe docs should be more explicit here but these are the correct types. Net allowlists should be a list of hosts (`"<hostname>[:<port>]"`) not representable with URLs.

Soremwar: Might be a little overkill, but I added docs specific to that field to specify the format of the URLs. Just in case

cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -712,7 +721,26 @@ declare class Worker extends EventTarget {
...
        */
-      deno?: boolean;
+      deno?: {
+        namespace?: boolean;
+        permissions?: "inherit" | {
+          env?: "inherit" | boolean;
+          hrtime?: "inherit" | boolean;
+          /**
+           * The format of the net access list must be `hostname[:port]`
+           * in order to be resolved
```
bartlomieju, line 731: How do you go about creating worker that has no permissions at all? Could you include it as example in the doc string?

cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -712,7 +721,26 @@ declare class Worker extends EventTarget {
        * hello world2
        *
        */
-      deno?: boolean;
+      deno?: {
```
bartlomieju, line 724: I wonder if we should keep `deno: true` for now so we don't break user code 🤔  it's unstable but it would be quite disruptive for a patch release

Soremwar: Yeah, let's keep `true` for backwards compat. However I'll remove docs that use it and put a couple comments so we can safely remove it in 2.0

nayeemrmn: Let's remove it in 1.7.0.

cli/dts/lib.deno.shared_globals.d.ts
Soremwar: I just added `false` as an option for the `permissions` prop. This disables all of the worker access. I'll add docs to the manual soon

ry:
Sorry for the slow review here. The change makes sense - I'm hesitating because I'm not sure how we should represent the permissions object. Boolean values are not enough. If we ever have a configuration file that contains permissions, it should probably have a similar representation as this options object....

Soremwar:
@ry I'm confused, what other values do you feel the permissions object should need to specify an allow list?

ry:
LGTM to me but I'll let @bartlomieju land.

cli/tests/workers_test.ts
```diff
@@ -378,3 +379,236 @@ Deno.test({
...
+      },
+    );
+    worker.terminate();
+  } catch (error) {
+    errorThrown = true;
+    if (!(error instanceof Deno.errors.PermissionDenied)) {
+      fail(`Unexpected error: ${error}`);
+    }
+  }
+  assert(errorThrown);
```
bartlomieju, line 589: Nitpick: use `assertThrows`

cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -661,24 +661,33 @@ declare class Worker extends EventTarget {
...
-       * available to worker thread.
-       *
-       * Currently worker inherits permissions from main thread (permissions
-       * given using `--allow-*` flags).
-       * Configurable permissions are on the roadmap to be implemented.
+       * Set deno.namespace to `true` to make `Deno` namespace and all of its methods
+       * available to worker thread. The namespace is disabled by default.
+       * 
+       * Configure deno.permissions options to change the level of access the worker will
+       * have. By default it will inherit the permissions of it's parent thread. The permissions
```
bartlomieju, line 670: Type: `it's` -> `its`

docs/runtime/workers.md
```diff
@@ -116,7 +119,109 @@ $ deno run --allow-read --unstable main.js
...
+> This is an unstable Deno feature. Learn more about
+> [unstable features](./stability.md).
+
+The permissions available for the worker are analogous to the CLI permission
+flags, meaning every permission enabled there can be disabled at the level of
+the Worker API. You can find a more detailed description of each of the
+permission options [here](../getting_started/permissions.md).
+
+By default a worker will inherit permissions from the thread it was created in,
+however in order to allow users to limit the access of this workers we provide
```
bartlomieju, line 133: typo: `workers` -> `worker`

docs/runtime/workers.md
```diff
@@ -116,7 +119,109 @@ $ deno run --allow-read --unstable main.js
...
+the `deno.permissions` option in the worker API.
+
+- For permissions that support granular access you can pass in a list of the
+  desired resources the worker will have access to, and for those who only have
+  the on/off option you can pass true/false respectively.
```
bartlomieju, line 138: Add note how permissions are resolved (relative to what file) when strings are passed instead of URL

runtime/js/11_workers.js
```diff
@@ -47,14 +50,122 @@
...
+    constructor(specifier, options = {}) {
       super();
-      const { type = "classic", name = "unknown" } = options ?? {};
+      const {
+        deno = {},
+        name = "unknown",
+        type = "classic",
+      } = options;
+
+      // TODO
```
bartlomieju, line 136: Please add your handle to the TODOs in this PR

runtime/ops/worker_host.rs
```diff
@@ -86,21 +105,348 @@ pub fn init(
...
+    PermissionState::Granted => match incoming {
+      Some(x) => Ok(x),
+      None => Ok(*target),
+    },
+    _ => match incoming {
+      Some(x) => match x {
+        PermissionState::Denied => Ok(x),
+        _ => Err(custom_error(
+          "PermissionDenied",
+          "Can't extend current permissions",
```
bartlomieju, line 122: `Can't escalate parent worker permissions`?

runtime/ops/worker_host.rs
```diff
@@ -86,21 +105,348 @@ pub fn init(
...
+    PermissionState::Granted => Ok(UnaryPermission::<String> {
+      global_state: new_permissions.global_state,
+      granted_list: new_permissions.granted_list,
+      denied_list: new_permissions.denied_list,
+    }),
+    PermissionState::Prompt => match new_permissions.global_state {
+      //Throw
+      PermissionState::Granted => Err(custom_error(
+        "PermissionDenied",
+        "Can't extend current permissions",
```
bartlomieju, line 156: ditto

runtime/ops/worker_host.rs
```diff
@@ -86,21 +105,348 @@ pub fn init(
...
+        ) {
+          Ok(UnaryPermission::<String> {
+            global_state: new_permissions.global_state,
+            granted_list: new_permissions.granted_list,
+            denied_list: target.denied_list.clone(),
+          })
+        } else {
+          Err(custom_error(
+            "PermissionDenied",
+            "Can't extend current permissions",
```
bartlomieju, line 172: ditto

runtime/ops/worker_host.rs
```diff
@@ -86,21 +105,348 @@ pub fn init(
...
+    },
+    PermissionState::Denied => match new_permissions.global_state {
+      PermissionState::Denied => Ok(UnaryPermission::<String> {
+        global_state: new_permissions.global_state,
+        granted_list: new_permissions.granted_list,
+        denied_list: new_permissions.denied_list,
+      }),
+      _ => Err(custom_error(
+        "PermissionDenied",
+        "Can't extend current permissions",
```
bartlomieju, line 191: Ditto

bartlomieju (changes requested):
@Soremwar awesome work, I have a few minor comments and this PR will be ready to land

bartlomieju (approved):
LGTM 🚀

PR #8216, opened 2020-11-01
fix(std/flags): Fix parse incorrectly parsing alias flags with equals signs in the value #8136

### LINKED ISSUE #8136: std/flags: parse function ignores characters after equals sign that is part of the value when a shortened (alias) flag is used.
It's a little hard to explain succinctly in the title, so here is an example:

*main.ts*:
```ts
import { parse } from "https://deno.land/std@0.74.0/flags/mod.ts";
console.log(Deno.args);
console.log(parse(Deno.args));
```

**Regular flag** 
`--foo=abc=def` -> `{ _: [], foo: "abc=def"}`

```sh
deno run main.ts --foo="abc=def" --bar=14
```
*Output*:
```
["--foo=abc=def", "--bar=14" ]
{ _: [], foo: "abc=def", bar: 14}
```

**Shortened (alias) flag**
`-f=abc=def` -> `{ _: [], f: "abc"}`

```sh
deno run main.ts -f="abc=def" -b=14
```
*Expected (?) output*:
```
["-f=abc=def", "-b=14" ]
{ _: [], f: "abc=def", b: 14}
```

*Actual output*:
```
["-f=abc=def", "-b=14" ]
{ _: [], f: "abc", b: 14}
```
---
This might be the expected behaviour but it looks to me like a bug.

TheWizardBear:
Fixes #8136.

I also added 3 more test cases, 2 of which would have failed before this change.


bartlomieju (approved):
LGTM

PR #8221, opened 2020-11-02
fix(cli): inject helpers when transpiling via swc

### LINKED ISSUE #8212: '_applyDecoratedDescriptor is not defined' error in bundled code
PR #8168 in deno version `1.5.1` now allows to bundle code with decorators.
Running the bundled code however gives error:
`error: Uncaught ReferenceError: _applyDecoratedDescriptor is not defined`

Test code to reproduce (same `test.ts` as in #8168):
```typescript
/* eslint-disable */

function Decorator() {
  return function (
    target: Record<string, any>,
    propertyKey: string,
    descriptor: TypedPropertyDescriptor<any>,
  ) {
    const originalFn: Function = descriptor.value as Function;
    descriptor.value = async function (...args: any[]) {
      return await originalFn.apply(this, args);
    };
    return descriptor;
  };
}

class SomeClass {
  @Decorator()
  async test(): Promise<void> {}
}

new SomeClass().test();
```

```sh
deno bundle test.js test.bundle.js
deno run test.bundle.js
error: [...]
```

kitsonk:
Fixes #8212

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -298,6 +298,7 @@ impl ParsedModule {
         legacy: true,
         emit_metadata: options.emit_metadata
       }),
+      helpers::inject_helpers(),
       typescript::strip(),
       fixer(Some(&self.comments)),
     );
@@ -413,6 +414,7 @@ pub fn transpile_module(
   src: &str,
   media_type: &MediaType,
   emit_options: &EmitOptions,
+  globals: &Globals,
   cm: Rc<SourceMap>,
 ) -> Result<(Rc<SourceFile>, Module), AnyError> {
   // TODO(@kitsonk) DRY-up with ::parse()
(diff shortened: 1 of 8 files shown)


cli/ast.rs
```diff
@@ -464,10 +466,11 @@ pub fn transpile_module(
       legacy: true,
       emit_metadata: emit_options.emit_metadata
     }),
+    helpers::inject_helpers(),
```
lucacasonato, line 469: Shouldn't this be optional depending on if decorators are enabled or not?

kitsonk: We have decorators enabled when transpiling, and they are not configurable off.  The only thing that is configurable is emitting the decorator meta data.  This is because basically if you don't use them there is a _zero_ cost.  Just like setting `"jsx"` to `"react"`.  If you don't use JSX, it doesn't mean anything.

In addition, the helpers only get injected if the code depends upon them.  There are the decorator ones, but if someone were to target < esnext, there would be other helpers required, so we need them on anyways.

bartlomieju (approved):
LGTM

cli/ast.rs
lucacasonato: > We have decorators enabled when transpiling, and they are not configurable off.

This kind of bothers me. I would rather legacy decorators were a config option (or a flag). But SGTM. Makes sense now.

PR #8225, opened 2020-11-02
fix(web): make onabort event handler web compatible

benjamingr:
Currently, `AbortSignal` implements `onabort` in a non spec-compliant way 
.
This causes a timing bug as event listeners can be added in arbitrary orders.

This PR reuses the existing listeners machinery rather than the ad-hoc event handler process - fixing the timing bug.

I have included a test - though the tests for AbortController are weird and in a different format from the EventTarget tests.

(I have a few more ideas of things to fix, but I'll try to take this a few at a time - since the first two PRs I made were merged and the promise hooks one will probably take a bit since it's bigger - I figured I'd follow up with 1-2 more at a time :])

### DIFF
--- a/op_crates/web/02_abort_signal.js
+++ b/op_crates/web/02_abort_signal.js
@@ -36,12 +36,23 @@
         throw new TypeError("Illegal constructor.");
       }
       super();
-      this.onabort = null;
-      this.addEventListener("abort", (evt) => {
-        const { onabort } = this;
-        if (typeof onabort === "function") {
-          onabort.call(this, evt);
-        }
+      // HTML specification section 8.1.5.1
+      let eventHandler = null;
+      Object.defineProperty(this, "onabort", {
+        get() {
+          return eventHandler;
+        },
+        set(value) {
+          if (eventHandler) {
+            this.removeEventListener("abort", eventHandler);
+          }
+          eventHandler = value;
+          if (typeof eventHandler === "function") {
+            this.addEventListener("abort", value);
+          }
+        },
+        configurable: true,
+        enumerable: true,
       });
     }
 
--- a/op_crates/web/abort_controller_test.js
+++ b/op_crates/web/abort_controller_test.js
@@ -23,7 +23,7 @@ function assertThrows(fn) {
 }
 
 function basicAbortController() {
-  controller = new AbortController();
+  const controller = new AbortController();
   assert(controller);
   const { signal } = controller;
   assert(signal);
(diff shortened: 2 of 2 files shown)


op_crates/web/02_abort_signal.js
```diff
@@ -36,11 +36,20 @@
         throw new TypeError("Illegal constructor.");
       }
       super();
-      this.onabort = null;
-      this.addEventListener("abort", (evt) => {
-        const { onabort } = this;
-        if (typeof onabort === "function") {
-          onabort.call(this, evt);
+      // Dom specification section 8.1.5.1
```
kitsonk, line 39: Is this the right section?  I can't seem to find it in the DOM Spec.  Points to something about XPath.

op_crates/web/02_abort_signal.js
```diff
@@ -36,11 +36,20 @@
...
+      let eventHandler = null;
+      Object.defineProperty(this, 'onabort', {
```
kitsonk: In browsers this is installed on the prototype as:

```js
{
  get() {}
  set() {}
  enumerable: true,
  configurable: true,
}
```

We should probably align, which would be a getter and setter in the class?  Just like `aborted` (well which doesn't have a setter)?

op_crates/web/02_abort_signal.js
benjamingr: https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-attributes this is the HTML specification and not the DOM specification (they link to each other) - I will update the comment.

op_crates/web/02_abort_signal.js
benjamingr: Good catch, I'll fix this.

op_crates/web/02_abort_signal.js
```diff
@@ -36,12 +36,23 @@
...
+          if (eventHandler) {
+            this.removeEventListener("abort", eventHandler);
+          }
+          eventHandler = value;
+          if (typeof eventHandler === "function") {
+            this.addEventListener("abort", value);
+          }
+        },
+        configurable: true,
+        writable: true
```
benjamingr, line 55: > TypeError: Invalid property descriptor. Cannot both specify accessors and a value or writable attribute, #<Object>

Will investigate :]

benjamingr: This should be enumerable and not writable

kitsonk (approved):
LGTM

PR #8228, opened 2020-11-02
refactor(cli): remove ProgramState::permissions

bartlomieju:
This commit removes ProgramState::permissions field.

Having permissions parsed from CLI flags stored on globally
accessible state object made it easy to mistakenly use these
permissions in situations which required "runtime" permissions.

CC @kitsonk


kitsonk (approved):
LGTM

PR #8230, opened 2020-11-03
fix(op_crates/web): define abort event handler on prototype

benjamingr:
Currently, AbortSignal implements onabort in a non spec-compliant way

Event handlers need to be defined on the protoype, this is also applies to other event handlers in Deno (like workers/sockets onmessage/messageerror etc).

I tried extracting the logic to a function so it's easier to reuse in the future, but I have no idea how to `import` things into the file (init makes a snapshot of these but in that flow web_util doesn't load first), I noticed other code is simply duplicated and decided I am not familiar enough in the code to change how JS loads :D 

So I just fixed this in AbortSignal (and added a test).

cc @kitsonk :]

### DIFF
--- a/op_crates/web/02_abort_signal.js
+++ b/op_crates/web/02_abort_signal.js
@@ -36,24 +36,6 @@
         throw new TypeError("Illegal constructor.");
       }
       super();
-      // HTML specification section 8.1.5.1
-      let eventHandler = null;
-      Object.defineProperty(this, "onabort", {
-        get() {
-          return eventHandler;
-        },
-        set(value) {
-          if (eventHandler) {
-            this.removeEventListener("abort", eventHandler);
-          }
-          eventHandler = value;
-          if (typeof eventHandler === "function") {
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM

kitsonk:
Yeah, we are still iterating on how to share common code between the op_crates... the abstraction comes with its own problems and benefits.

benjamingr:
Thanks, I will also need to do a follow up fix (that's also a bug in Node) in timings because apparently the web is really weird  and:

```js
const { port1: mp } = new MessageChannel();
mp.addEventListener("message", () => console.log("1"));
mp.onmessage = () => console.log("not called");
mp.addEventListener("message", () => console.log("3"));
mp.onmessage = () => console.log("2");
mp.addEventListener("message", () => console.log("4"));

mp.dispatchEvent(new Event("message"));
```

Needs to log 1 / 2 / 3 / 4

PR #8232, opened 2020-11-03
build: remove eslint

### LINKED ISSUE #8122: Replace ESLint with dlint for development
Currently in development of `deno` we use `ESLint` to lint JavaScript and TypeScript code. Additionally `std/` and `cli/tests/unit` are linted using `deno lint` as integration tests. 

In `deno_lint` repository we have a small example binary called [`dlint`](https://github.com/denoland/deno_lint/tree/master/examples/dlint) that is essentially a small CLI on top of `deno_lint` crate (which is used as a basis for `deno lint` subcommand). There exists an issue to provide prebuilt `dlint` binaries on new releases of `deno_lint` crate (https://github.com/denoland/deno_lint/issues/367). 

I propose we switch from `ESLint` to `dlint` to do linting during development for the following reasons:
- remove Node dependency on `ESLint` (it's part of `deno_third_party` git submodule); I think `ESLint` is our last Node dependency and <s>we could remove Node altogether</s> Node is used in benchmarks
- more dogfooding of `deno_lint`
- `dlint` is significantly faster that `ESLint`

If we agree to do that then the order of action is as follows:
- set up releases of `dlint` - https://github.com/denoland/deno_lint/issues/367
- release `dlint` binary with next `deno_lint` crate version
- add `dlint` binary to `deno_third_party` repository
- change `tools/lint.py` to use `dlint` binary instead of `ESLint`
- remove `ESLint` from `deno_third_party`

CC @ry @kitsonk @lucacasonato @nayeemrmn @caspervonb

bartlomieju:
Ref https://github.com/denoland/deno_third_party/pull/80

Fixes #8122


ry (approved):
🎉

Issue #8041, opened 2020-10-19
Feature request: Add 'only' flag to bench API


cknight:
Deno's testing API has a very useful `only` option, to only run that one test.  It would be great to have this same option in the bench API allowing a single benchmark test to be run.  My use case is that I have lots of benchmark tests which can take awhile to run and I wish sometimes to just run one of them to examine performance of just that variant.

bartlomieju:
I think it's a very sensible future and should be straight-forward to implement. Marking as good first issue.

jopemachine:
I want to try to resolve the issue.
But It seems that there is already an `only` option in `std/testing/bench.ts`, so I'm confused..

Maybe I have missed something because I'm a newbie to `deno`.
I beg your pardon for my lack of understanding.

May I ask for a little more explanation of where the option should be implemented?

cknight:
hi @jopemachine , welcome to the deno community.  Apologies for my sparse issue.  Let me explain in more detail.

From the bench docs:
>Filtering can be applied by setting BenchmarkRunOptions.only and/or BenchmarkRunOptions.skip to regular expressions matching benchmark names.

This means coming up with a regular expression to match the benchmarks you want to run.  This is not always straightforward and can be awkward matching multiple tests.

Deno's test framework allows the specification of an `only` flag in the test definition itself (in addition to command line regular expression filtering).  It works like this:
```
test({
  name: "Example use of 'only' flag",
  only: true,
  fn() {
    assertEquals(1, 1);
  },
});
```
In this scenario, only tests with the `only` flag will be run.  

Ideally we would have this option for bench definitions as well like this:
```
bench({
  name: "Optic logging 10K",
  only: true,
  runs: 100,
  func(b: BenchmarkTimer): void {
    b.start();

    for (let i = 0; i < 10000; i++) {
      console.log("Hello world");
    }

    b.stop();
  },
});

```
Hope this makes sense.

littletof:
@jopemachine @bartlomieju how should this feature work together with the `skip`, and `only` filter in `runBenchmarks`?

So if i have benches like:
* `itHasOnly`
* `simpleBench`
* `extraBench`
* `extraBenchWithOnly`

and run `runBenchmarks({ only: /Bench/, skip: /extra/ })` which benchmarks should be ran?
Basically which option is stronger?

jopemachine:
> @jopemachine @bartlomieju how should this feature work together with the `skip`, and `only` filter in `runBenchmarks`?
> 
[...]

PR #8237, opened 2020-11-04
feat(std): support bench API's only flag

jopemachine:
## Related issue
#8041 

## Test
Manual Test

Here is the test code.
```ts
// benchTest.ts
import {
  bench,
  BenchmarkRunError,
  BenchmarkRunProgress,
  clearBenchmarks,
  ProgressState,
  runBenchmarks,
} from "./std/testing/bench.ts";

bench({
  name: "runs100ForIncrementX1e6",
  runs: 100,
  func(b: any): void {
    b.start();
    for (let i = 0; i < 1e6; i++);
    b.stop();
  },
});

bench({
  name: "runs100ForIncrementX1e6-2",
  only: true,
  runs: 100,
  func(b: any): void {
    b.start();
    for (let i = 0; i < 1e6; i++);
    b.stop();
  },
});

bench({
  name: "runs100ForIncrementX1e6-3",
  runs: 100,
  func(b: any): void {
    b.start();
    for (let i = 0; i < 1e6; i++);
    b.stop();
  },
});

await runBenchmarks({ silent: false });

```

Here is the result (of `deno run benchTest.ts`)

```
Check file:///Users/igyubong/Desktop/deno/benchTest.ts
running 1 benchmark ...
benchmark runs100ForIncrementX1e6-2 ... 
    100 runs avg: 0.3ms
benchmark result: DONE. 1 measured; 2 filtered
```

As expected, only items with the flag `only` are true were executed.

## Notes

1. I'm a newbie to `deno` community. I would appreciate it if you could let me know if there is anything out of the coding standard of `deno` or anything I missed :)

2. I haven't removed the existing code's `option` argument using Regexp. Please let me know if we need to remove this argument.


crypt096 (approved):

littletof:
Thanks @jopemachine!
The basic feature looks good, I just have some nitpicks.
Other than those, and depending on the response on this https://github.com/denoland/deno/issues/8041#issuecomment-727209105, I think its good to go.

std/testing/bench.ts
```diff
@@ -18,10 +18,12 @@ export interface BenchmarkTimer {
 export interface BenchmarkFunction {
   (b: BenchmarkTimer): void | Promise<void>;
   name: string;
+  only?: boolean;
 }
 
 /** Defines a benchmark definition with configurable runs. */
 export interface BenchmarkDefinition {
+  only?: boolean;
```
littletof, line 26: Since `only` is optional, it would be better to move it down below `runs?`.

std/testing/bench.ts
```diff
@@ -18,10 +18,12 @@ export interface BenchmarkTimer {
 export interface BenchmarkFunction {
   (b: BenchmarkTimer): void | Promise<void>;
   name: string;
+  only?: boolean;
```
littletof, line 21: Since `runBenchmarks` has an `only: Regexp` option too, I think it would be good to add JSDocs to these new `only: boolean` properties, so its clear what their function is.

std/testing/bench.ts
```diff
@@ -188,10 +198,18 @@ export async function runBenchmarks(
   { only = /[^\s]/, skip = /^\s*$/, silent }: BenchmarkRunOptions = {},
   progressCb?: (progress: BenchmarkRunProgress) => void | Promise<void>,
 ): Promise<BenchmarkRunResult> {
+  // If there is an item that only is true, proceed to benchmark only those items
```
littletof, line 201:
```suggestion
  // If there are benches where only is true, proceed to benchmark only those items
```

littletof:
Also, could you please add tests that cover these different cases? (simple benches, without `skip` and `only` regexp, and an other, where `skip` and `only` regexp is also set in `runBenchmarks`)

Something like this: https://github.com/denoland/deno/blob/master/std/testing/bench_test.ts#L200-L216

jopemachine:
Ok, I will. Thank you for your detailed review! :)

jopemachine:
> Also, could you please add tests that cover these different cases? (simple benches, without `skip` and `only` regexp, and an other, where `skip` and `only` regexp is also set in `runBenchmarks`)
> 
> Something like this: https://github.com/denoland/deno/blob/master/std/testing/bench_test.ts#L200-L216

Ok, I will. Thank you for your detailed review! :)

bartlomieju:
@littleof would you mind giving this PR another review? @jopemachine is it ready to land?

jopemachine:
> @littleof would you mind giving this PR another review? @jopemachine is it ready to land?

@bartlomieju littleof left a suggestion on here (https://github.com/denoland/deno/issues/8041#issuecomment-727275760)

I think maybe it could be better to check the suggestion before landing.

bartlomieju:
> > @littleof would you mind giving this PR another review? @jopemachine is it ready to land?
> 
> @bartlomieju littleof left a suggestion on here ([#8041 (comment)](https://github.com/denoland/deno/issues/8041#issuecomment-727275760))
> 
> I think maybe it could be better to check the suggestion before landing.

@jopemachine I think that suggestions makes a lot of sense, the more similar test and bench APIs are the better.

jopemachine:
> > > @littleof would you mind giving this PR another review? @jopemachine is it ready to land?
> > 
> > 
> > @bartlomieju littleof left a suggestion on here ([#8041 (comment)](https://github.com/denoland/deno/issues/8041#issuecomment-727275760))
> > I think maybe it could be better to check the suggestion before landing.
> 
> @jopemachine I think that suggestions makes a lot of sense, the more similar test and bench APIs are the better.

@bartlomieju So, May I ask what should I do?
Is it alright to implement throwing an exception when the `filtered` result is not zero?
Or just indicating it in the output is better?

bartlomieju:
@jopemachine I think throwing is preferable similar to how test runner exists with code 1 if there are tests with "only" option
https://github.com/denoland/deno/blob/27dd78601669a92d0200ad197adcf0919fdc9e46/cli/rt/40_testing.js#L338-L340

jopemachine:
@bartlomieju,

https://github.com/denoland/deno/issues/8041#issuecomment-737297916

Could you review his comment about this issue?

If you decide after the review of the comment, I will start working on this issue again, thanks!

bartlomieju:
@jopemachine I'm not really sure anymore what we should do... Let's wait a bit and discuss this issue during STD WG meeting

bartlomieju:
@jopemachine we won't land this PR before standard library migration to a separate repository (https://github.com/denoland/deno/issues/9029#issuecomment-766934654). 

I'm still very much interested in discussing these changes on upcoming meetings and potentially landing the PR.

bartlomieju:

PR #8242, opened 2020-11-04
docs(std/node/querystring): add missing JSDoc

ayntee:
Add missing JSDoc for `std/node/querystring.ts`

Related: #7487


ry (approved):
Thanks!

PR #8245, opened 2020-11-04
refactor(cli): refactor file_fetcher

kitsonk:


cli/specifier_handler.rs
```diff
@@ -258,20 +259,31 @@ impl SpecifierHandler for FetchHandler {
...
+              generic_error(message)
+            } else {
+              err
+            }
+          } else {
+            err
+          };
           if let Some(location) = maybe_location {
-            if !is_dynamic {
+            if !location.filename.contains("$deno$") {
```
bartlomieju, line 286: This is a bit surprising, maybe add a comment with explanation?

cli/specifier_handler.rs
```diff
@@ -258,20 +259,31 @@ impl SpecifierHandler for FetchHandler {
...
+            if e.kind() == std::io::ErrorKind::NotFound {
+              let message = if let Some(location) = &maybe_location {
+                format!(
+                  "Cannot resolve module \"{}\" from \"{}\".",
+                  requested_specifier, location.filename
+                )
+              } else {
+                format!("Cannot resolve module \"{}\".", requested_specifier)
+              };
+              generic_error(message)
```
bartlomieju, line 278: Maybe it should be `NotFound` instead of `Error`?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
-  cache_blocklist: Vec<String>,
-  use_disk_cache: bool,
-  no_remote: bool,
-  cached_only: bool,
-  http_client: reqwest::Client,
-  // This field is public only to expose it's location
-  pub http_cache: HttpCache,
+/// Indicates how cached source files should be handled.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Cache {
```
bartlomieju, line 73: Bike-shed: `CacheConfig`?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
+    maybe_types: None,
+    media_type,
+    source,
+    specifier: specifier.clone(),
+  })
+}
 
-    None
+/// Return a validated scheme for a given module specifier.
+fn get_scheme(specifier: &ModuleSpecifier) -> Result<String, AnyError> {
```
bartlomieju, line 137: `validate_scheme`?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
-      source_code,
-      types_header,
-    }))
+    Ok(Some(file))
   }
 
-  /// Asynchronously fetch remote source file specified by the URL following redirects.
-  ///
-  /// Note that this is a recursive method so it can't be "async", but rather return
-  /// Pin<Box<..>>.
```
bartlomieju, line 408: Why remove this comment?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
+  pub async fn fetch(
+    &self,
+    specifier: &ModuleSpecifier,
+    permissions: &Permissions,
+  ) -> Result<File, AnyError> {
+    debug!("FileFetcher::fetch() - specifier: {}", specifier);
+    let scheme = get_scheme(specifier)?;
+    permissions.check_specifier(specifier)?;
+    if let Some(file) = self.cache.get(specifier) {
+      debug!("  found in in-memory cache");
```
bartlomieju, line 417: remove this line?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
-            }
-          }
-        } else {
-          default
-        }
-      }
-      Some(_) => default,
-    },
+  /// Get a previously in memory cached file.
+  pub fn get_cache(&self, specifier: &ModuleSpecifier) -> Option<File> {
```
bartlomieju, line 442: I'd expect to get an instance of `cache` from this method :) maybe `get_cached`?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
-    rest.to_string()
-  } else {
-    "".to_string()
+  fn get_file(
+    &self,
+    specifier: &ModuleSpecifier,
+    bytes: Vec<u8>,
+    headers: &HashMap<String, String>,
+  ) -> Result<File, AnyError> {
+    let local = self.http_cache.get_cache_filename(specifier.as_url());
```
bartlomieju, line 452: If I'm not mistaken this call will raise `unimplemented!` for local files.

bartlomieju: Maybe use `if` here and explicitly convert local URL to filepath?

cli/file_fetcher.rs
```diff
@@ -2,677 +2,499 @@
...
+fn get_source_from_bytes(
+  bytes: Vec<u8>,
+  maybe_charset: Option<String>,
+) -> Result<String, AnyError> {
+  let source = if let Some(charset) = maybe_charset {
+    text_encoding::convert_to_utf8(&bytes, &charset)?.to_string()
+  } else {
+    String::from_utf8(bytes)?
+  };
+  if source.starts_with("#!") {
```
ry, line 160: It seems odd that a function called `get_source_from_bytes` would also handle shebang stuff.

crypt096 (approved):

cli/file_fetcher.rs
kitsonk: Makes more sense to break it into two.

cli/specifier_handler.rs
kitsonk: I will add a comment, but honestly, I don't know why we were skipping it with dynamic imports anyways, the location is useful information, where as with the injected modules, the location makes no sense to print out...

cli/file_fetcher.rs
kitsonk: If `Cache` is confusing, I would prefer `CacheSetting` versus `CacheConfig`.

cli/file_fetcher.rs
kitsonk: that would imply that it is returning a boolean or something... `get_validated_scheme` I guess...

cli/file_fetcher.rs
kitsonk: Accidental when rewriting the module... will restore.

cli/file_fetcher.rs
kitsonk: I will double check, but I don't believe it does (all the tests are passing), it points to the local file on disk.

cli/file_fetcher.rs
kitsonk: I also changed `insert_cache` to `insert_cached` to be consistent.

cli/file_fetcher.rs
kitsonk: It would, though the method is never called except for remote files, which is why it doesn't cause problems.  I will make it explicit that this method is only for remote files.

bartlomieju (approved):
LGTM, nice cleanup

ry (approved):

PR #8262, opened 2020-11-06
fix(cli): allow remapping to locals for import map

### LINKED ISSUE #7723: Cannot remap a transitive dependency to a local file
https://github.com/denoland/deno/pull/5050 introduced the logic that prohibits static imports of local modules from remote modules. While I understand the security issues it solves, I believe, it also makes impossible some nice and valid use-cases, in particular, remapping a transitive dependency to a local module using an import map.
Can we somehow allow this use-case?

kitsonk:
Fixes #7723

### DIFF
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -441,7 +441,9 @@ impl Module {
     } else {
       None
     };
+    let mut remapped_import = false;
     let specifier = if let Some(module_specifier) = maybe_resolve {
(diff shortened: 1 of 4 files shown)


kitsonk:
Not really sure if this is a fix or a feature. 🤷

bartlomieju (approved):
LGTM, I think we can qualify it as a fix :)

PR #8263, opened 2020-11-06
fix(cli): properly handle type checking root modules with type defini…

kitsonk:
While working on another bug I found this one, where if the root module for `deno run` or `deno cache` has an `X-TypeScript-Types` header or `/// <reference types="..." />` it won't be properly type checked.


bartlomieju (approved):
LGTM

PR #8264, opened 2020-11-06
fix(op_crates/web): handler order when reassign

benjamingr:
This fixes the nasty bug described in https://github.com/denoland/deno/pull/8230#issuecomment-722945098 

cc @kitsonk

### DIFF
--- a/op_crates/web/02_abort_signal.js
+++ b/op_crates/web/02_abort_signal.js
@@ -67,12 +67,12 @@
 
   function makeWrappedHandler(handler) {
     function wrappedHandler(...args) {
-      if (typeof wrappedHandler.handler !== 'function') {
+      if (typeof wrappedHandler.handler !== "function") {
         return;
       }
       return wrappedHandler.handler.call(this, ...args);
     }
-    wrappedHandler.handler = handler
+    wrappedHandler.handler = handler;
     return wrappedHandler;
   }
   function defineEventHandler(emitter, name) {
--- a/op_crates/web/02_abort_signal.js
+++ b/op_crates/web/02_abort_signal.js
@@ -64,22 +64,35 @@
   }
 
   const handlerSymbol = Symbol("eventHandlers");
+
+  function makeWrappedHandler(handler) {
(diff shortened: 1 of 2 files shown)


benjamingr:
cc @kitsonk @kt3k any chance this can get a review/merge? I have follow-up work building on this I want to start working on ^^

op_crates/web/02_abort_signal.js
```diff
@@ -64,22 +64,35 @@
   }
 
   const handlerSymbol = Symbol("eventHandlers");
+
+  function makeWrappedHandler(handler) {
+    function wrappedHandler(...args) {
+      if (typeof wrappedHandler.handler !== "function") {
+        return;
+      }
```
bartlomieju, line 72: In what situation this branch would be true?

benjamingr: In the test basically:
```js
ee.onabort = () => console.log('irrelevant');
ee.onabort = () => null;
ee.addEventListener('abort', e => console.log(1);
ee.onabort = () => console.log(2);
```

In this case, the target has to fire "1 2" and not "2 1" so we need to "keep the slot".

the DOM is weird like that 🤷

benjamingr: The test tests this btw

bartlomieju: Thanks, that makes sense 👍

bartlomieju (approved):
LGTM, thanks @benjamingr!

benjamingr:
Great thanks, I'll keep em coming - I'd like to make this fix to other EventTargets that have handlers (like sockets) but I am unsure how to share code between web crates. 

Should I just copy-paste those fixes to those files and open a PR until you figure out how you want sharing code between them to look?

bartlomieju:
@benjamingr I think a little copy-paste now is fine - indeed we don't yet have a good way to share this functionality between crates.

benjamingr:
Ok, thanks!

PR #8269, opened 2020-11-06
refactor: remove redundant code in move.ts

behnammodi:
Some redundant code in `move` and `moveSync` removed


bartlomieju (approved):
LGTM

PR #8273, opened 2020-11-06
Implement 'AsyncRefCell' and 'ResourceTable2'

piscisaureus:

### DIFF
--- a/core/async_cell.rs
+++ b/core/async_cell.rs
@@ -1,0 +1,713 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
+
+use std::any::Any;
+use std::borrow::Borrow;
+use std::cell::Cell;
+use std::cell::UnsafeCell;
+use std::collections::VecDeque;
+use std::ops::Deref;
+use std::rc::Rc;
+
+use self::internal as i;
+
+pub type AsyncRef<T> = i::AsyncBorrowImpl<T, i::Shared>;
+pub type AsyncMut<T> = i::AsyncBorrowImpl<T, i::Exclusive>;
+
+pub type AsyncRefFuture<T> = i::AsyncBorrowFutureImpl<T, i::Shared>;
+pub type AsyncMutFuture<T> = i::AsyncBorrowFutureImpl<T, i::Exclusive>;
+
+pub struct AsyncRefCell<T> {
+  value: UnsafeCell<T>,
+  borrow_count: Cell<i::BorrowCount>,
+  waiters: Cell<VecDeque<Option<i::Waiter>>>,
+  turn: Cell<usize>,
+}
+
+impl<T: 'static> AsyncRefCell<T> {
+  /// Create a new `AsyncRefCell` that encapsulates the specified value.
+  /// Note that in order to borrow the inner value, the `AsyncRefCell`
+  /// needs to be wrapped in an `Rc` or an `RcRef`. These can be created
+  /// either manually, or by using the convenience method
+  /// `AsyncRefCell::new_rc()`.
+  pub fn new(value: T) -> Self {
+    Self {
+      value: UnsafeCell::new(value),
+      borrow_count: Default::default(),
+      waiters: Default::default(),
+      turn: Default::default(),
+    }
+  }
+
+  pub fn new_rc(value: T) -> Rc<Self> {
+    Rc::new(Self::new(value))
+  }
+
+  pub fn as_ptr(&self) -> *mut T {
+    self.value.get()
+  }
+}
+
+impl<T: Default + 'static> Default for AsyncRefCell<T> {
+  fn default() -> Self {
+    Self::new(Default::default())
+  }
+}
+
+impl<T: Default + 'static> AsyncRefCell<T> {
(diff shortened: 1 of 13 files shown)


piscisaureus:
Unfortunately it did not, and I do know why, and it is complicated ...

piscisaureus:
Far from green, but this works:

```
cargo run --release -p deno_core --example http_bench_bin_ops
```

ry:
I ran the benchmarks - there is some minimal performance degradation (~5%) 
https://gist.github.com/ry/3fd8d1420fd3710755f9de1f1641d930

core/resources2.rs
```diff
@@ -0,0 +1,136 @@
...
+    self.add_rc(Rc::new(resource))
+  }
+
+  /// Inserts a `Rc`-wrapped resource into the resource table.
+  ///
+  /// The resource type is erased at runtime and must be statically known
+  /// when retrieving it through `get()`.
+  ///
+  /// Returns a unique resource ID, which acts as a key for this resource.
+  pub fn add_rc<T: Resource>(&mut self, resource: Rc<T>) -> ResourceId {
```
ry, line 101: This is currently unused. Do you expect it to be?

core/ops.rs
```diff
@@ -34,6 +34,7 @@ pub enum Op {
 /// Maintains the resources and ops inside a JS runtime.
 pub struct OpState {
   pub resource_table: crate::ResourceTable,
+  pub resource_table_2: crate::resources2::ResourceTable,
```
ry, line 37: nit: use `resource_table2` - shorter.

core/examples/http_bench_bin_ops.rs
```diff
@@ -94,8 +153,9 @@ fn op_listen(
   debug!("listen");
   let addr = "127.0.0.1:4544".parse::<SocketAddr>().unwrap();
   let std_listener = std::net::TcpListener::bind(&addr)?;
-  let listener = TcpListener::from_std(std_listener)?;
-  let rid = state.resource_table.add("tcpListener", Box::new(listener));
+  std_listener.set_nonblocking(true)?;
```
bartlomieju, line 156: Is there a reason to create `std::net:TcpListener` instead of using tokio's?

core/resources2.rs
```diff
@@ -0,0 +1,136 @@
...
+    let removed_resource = self.index.insert(rid, resource);
+    assert!(removed_resource.is_none());
+    self.next_rid += 1;
+    rid
+  }
+
+  /// Removes the resource with the given `rid` from the resource table. If the
+  /// only reference to this resource existed in the resource table, this will
+  /// cause the resource to be dropped. However, since resources are reference
+  /// counted, therefore pending ops are not automatically cancelled.
```
bartlomieju, line 113: Is this intentional design? I thought the main point of `close` is to drop pending ops for the resource.

core/async_cell.rs
```diff
@@ -0,0 +1,619 @@
...
+  turn: Cell<usize>,
+}
+
+impl<T: 'static> AsyncRefCell<T> {
+  /// Create a new `AsyncRefCell` that encapsulates the specified value.
+  /// Note that in order to borrow the inner value, the `AsyncRefCell`
+  /// needs to be wrapped in an `Rc` or an `RcRef`. These can be created
+  /// either manually, or by using the convenience method
+  /// `AsyncRefCell::new_rc()`.
+  pub fn new(value: T) -> Self {
```
bartlomieju, line 32: If it is required to wrap in an `Rc` do we need this method? Why not make `new_rc` the default?

piscisaureus: Look at http_bench_bin_ops: a `TcpStream` contains two `AsyncRefCell`s but they're not individually wrapped in an `Rc` - just the struct that contains them is. So it needs to be possible to create a "naked" `AsyncRefCell`.

core/examples/http_bench_bin_ops.rs
piscisaureus: Not really, but that's not a change introduced by this PR. 
I believe this "hack" was originally added because Tokio's `bind()` method is async.

core/async_cell.rs
```diff
@@ -0,0 +1,713 @@
...
+      assert!(cell.try_borrow_mut().is_none());
+    }
+
+    {
+      let ref8 = cell.try_borrow().unwrap();
+      assert_eq!(ref8.look(), 2);
+      assert!(cell.try_borrow_mut().is_none());
+      assert!(cell.try_borrow().is_some());
+    }
+  }
```
ry, line 682: 👍

ry (approved):
LGTM

bartlomieju (approved):
LGTM

ialameera (approved):

ialameera (approved):

core/async_cell.rs
ialameera: > 👍

core/resources2.rs
piscisaureus: It actually does get called by the `add()` method.
But if it turns out that we never use `add_rc()` then it'll be trivial to merge these methods.

PR #8274, opened 2020-11-07
feat: consistent Node.js builtin shapes

guybedford:
Node.js exports all core modules as both direct exports and a `default` export. See for example https://nodejs.org/dist/latest-v15.x/docs/api/esm.html#esm_commonjs_namespaces.

Deno currently implements the `default` export inconsistently across the core libraries. With this PR it sets up a consistent pattern for ensuring both `import process from 'https://deno.land/std@0.76.0/node/process.ts'` and `import * as process from 'https://deno.land/std@0.76.0/node/process.ts'` can work as expected, as they do in Node.js for the process builtin.

This PR contains two separate commits for a nicer diff to review

1. Move all namespace implementations to a `_builtin.ts` file.
2. Create a shell for each builtin using the pattern `export * from './_builtin.ts'; import * as m from './_builtin.ts'; export default m;` and update builtins as appropriate to fit.

This PR also adds two breaking changes

* `import Buffer from 'buffer'` must now be `import { Buffer } from 'buffer'` to match Node.js.
* `import { assert } from 'assert'` is not an export in Node.js so is also removed.

Feedback welcome, and please go easy on me this is my first Deno PR :)


guybedford:
I can't actually get deno lint / format to work on WSL, and am getting:

```
cargo run run --unstable --allow-write --allow-read --allow-run ./tools/format.js --check
    Finished dev [unoptimized + debuginfo] target(s) in 0.15s
     Running `target/debug/deno run --unstable --allow-write --allow-read --allow-run ./tools/format.js --check`
dprint
error: Uncaught NotFound: No such file or directory (os error 2)
    at processResponse (core.js:224:13)
    at Object.jsonOpSync (core.js:248:12)
    at opRun (deno:cli/rt/40_process.js:19:17)
    at Object.run (deno:cli/rt/40_process.js:103:17)
    at dprint (file:///home/guybedford/Projects/deno/tools/format.js:8:18)
    at main (file:///home/guybedford/Projects/deno/tools/format.js:39:9)
    at async file:///home/guybedford/Projects/deno/tools/format.js:63:1
```

any suggestions?

nayeemrmn:
The `tools/*.js` scripts use Deno as a meta-tool, and are expected to be run under Deno-latest installed in your environment. Just FYI though, not sure if it fixes the issue.

bartlomieju:
@guybedford looks like `third_party` git submodule is not checked out - `git submodule update --init --recursive` should fix the problem

guybedford:
Thanks for the pointers!

guybedford:
I've refactored this PR to only include the namespace changes, will create a separate PR for the globals.

I've also split out the diff into two separate commits, one with the file renames and one with the changes. This should make for an easier review process.

See https://github.com/denoland/deno/pull/8274/commits/7799717d69b649bc712b38493014c2364eb437cf for the main changes.

std/node/assert.ts
```diff
@@ -1,25 +1,45 @@
...
+}
+const ok = assert;
+export default assert;
 
-export {
-  assert,
-  assert as default,
-  assert as ok,
+// @deprecated use `import assert from 'assert'` as the "assert" export is not a Node.js export
+export { assert };
```
lucacasonato, line 23: @guybedford `std` is still unstable. Feel free to just remove outright, rather than deprecating.

std/node/buffer.ts
```diff
@@ -1,608 +1,6 @@
...
-export { Buffer };
-
-Object.defineProperty(globalThis, "Buffer", {
-  value: Buffer,
-  enumerable: false,
-  writable: true,
-  configurable: true,
-});
+// @deprecated use `import { Buffer } from 'buffer'` as the default export will change to `export default { Buffer }` to match Node.js
+export default m.Buffer;
```
lucacasonato, line 6: Ditto

std/node/assert.ts
guybedford: Sure, if you're happy to do that aligning now would make sense definitely.

lucacasonato (approved):
LGTM - thanks @guybedford.

PR #7767, opened 2020-09-30
feat(fmt): support --watch flag

bartOssh:
[fixes watch fmt only of 7465](https://github.com/denoland/deno/issues/7465)

PR #7740, opened 2020-09-28
feat(bundle): support --watch flag

bartOssh:
[fixes watch bundle only](https://github.com/denoland/deno/issues/7465)

Issue #7848, opened 2020-10-07
deno run --watch exits when file has invalid syntax


andykais:
This is possibly intentional, but I would like to illustrate my case regardless:
if I have a file like so:
```ts
// test.ts
function x() {
}
```
when I run `deno run --watch --unstable test.ts`, the file compiles. If I change the file to something with invalid syntax, like so:
```ts
// test.ts
function__ x() {
}
```
then deno outputs the following, but keeps the watcher active:
```
Watcher File change detected! Restarting!
error: Expected ';', '}' or <eof> at file:///home/andrew/test.ts:1:11
Watcher Process terminated! Restarting on file change...
```
however, if I _start_ `deno run --watch --unstable test.ts` on a file with invalid syntax, it writes the error, and then immediately exists:
```
 ✘ andrew  ~  deno run --watch --unstable test.ts
error: Expected ';', '}' or <eof> at file:///home/andrew/test.ts:1:11
```

I believe what should happen here, is that `run --watch` should keep the file watcher active even when the first compilation fails with invalid syntax

bartlomieju:
@magurotuna do you think https://github.com/denoland/deno/pull/8276 addresses this problem?

magurotuna:
@bartlomieju No, #8276 doesn't fix this issue. To make matters worse, it seems like the watcher exits when the initial content of the file is correct and then it is updated with invalid syntax. I'll investigate it.

magurotuna:
I think I've found the cause. I'll fix it in #8276.

bartlomieju:
Fixed in #8276

magurotuna:
Implementary: What @andykais requested in this PR has not been addressed in #8276. I mean, in the case where a user runs `deno run --watch --unstable test.ts` with `test.ts` having invalid syntax, then the file watcher still immediately exits. This is due to a technical issue - in such a case, module resolution doesn't work, then the dependency graph isn't constructed, so the watcher can't determine which files should be observed.

magurotuna:
My intuition to achieve @andykais's request is to have the watcher observe files specified as command line arguments if module resolution fails. For example, say `foo.ts` has invalid syntax thus module resolution cannot be done with it. Running the following command:

```
$ deno run --watch --unstable test.ts
```

then the watcher would observe `test.ts` only. In this way, the watcher wouldn't exit due to invalid syntax.
I'm not really sure if this will fully satisfy users' needs though.

andykais:
[...]

PR #8276, opened 2020-11-07
feat(unstable): Support --watch flag for bundle and fmt subcommands

### LINKED ISSUE #8181: bug(watcher): changes on a file newly added to dependency are not watched
### version

current latest master (bfce376c)

### How to reproduce

1. create a file named `main.js`with the following content:

```js
console.log('hello');
```
2. start running deno watcher by `cargo run -- run --watch --unstable main.js`. At this point we can see an output like:

```
hello
```
3. create a new file named `new.js`. Nothing happens on the terminal.

```js
export const world = 'world';
```

4. edit `main.js` to the following and save it:

```diff
+import { world } from "./new.js";
console.log("hello");
+console.log(world);
```

When saving it, we get:

```
hello
world
```

5. edit the value of `world` in `new.js` and save it:

```diff
-export const world = "world";
+export const world = "Deno!";
```

When saving it, nothing happens although we expect `hello\nDeno!` to be printed.

I've already found the cause of this problem in the watcher implementation. This is because module resolution is done only once at the starting of the watcher. Basically we should resolve module every time any event is detected.

I'll look into this problem deeply and try to elabolate a solution.

magurotuna:
This PR adds `--watch` flag support for both `bundle` and `fmt` subcommands.
In addition to this, it refactors `run --watch` command so that module resolution will occur every time the file watcher detects file addition/deletion, which allows the watcher to observe a file that is newly added to the dependency as well.

-----------------------------

Fixes #8181 

Currently, `--watch` flag is working with `run` only, but it is considered the flag is going to be added to other subcommands like `fmt`, `lint`, and so on. (see #7465)
For this purpose, the current implementation of `watch_func` function has abstract, extendable design - all we have to do to add the watcher feature to subcommands is passing both target file paths and an operation you want to execute to the `watch_func` function.

Considering it deeply, it seems like this `watch_func` design is not suitable for usecases, especially `run` subcommand. In case we want to detect file changes and then restart it, the watcher should also keep an eye on deletion and addition of files or dependency.
So target file paths will vary, although the current `watch_func` receives constant file paths. This is why the issue described in #8181 happened.

To address the issue, I think we should resolve dependency whenever the watcher detects changes. Actually this is relatively heavy task to be done every time, but it seems needed.

In this PR, I've added new watcher funtcion named `watch_func_for_run` which is dedicated to `run` subcommand. 
Additionally I've removed the original `watch_func` function. I know this removal will have an effect on adding `--watch` flag to other subcommands, which are being worked on at #7740 and #7767. I think other subcommands, like `fmt`, also have the same issue as `run`. That said, if we would run `deno fmt --watch`, the watcher would have to detect any addition or deletion of files. Therefore we should probably consider again what `watch_func` function interface should look like, without using the current one.

#### TODO

- [x] add test


magurotuna:
I also added `fmt --watch` feature to show how we can make subcommands other than `run` work with `--watch` flag. (a lot of code come from #7767)

magurotuna:
Just added `bundle --watch` as well. Bundle is similar to `run` rather than `fmt`, so I renamed `watch_func_for_run` to `watch_func_with_module_resolution` and fixed its parameter type a little.

bartlomieju:
`watch_func_with_module_resolution` - how about `watch_func_with_module_graph`?

magurotuna:
@bartlomieju 
`bundle` operation requires `module_graph::Graph`, where `run` requires `deno_core::ModuleSpecifier`. This is why I avoided using the word `graph`as function name.  
But indeed `watch_func_with_module_resolution` is long and kind of redundant, so I'd be happy to find appropriate name...

cli/file_watcher.rs
```diff
@@ -53,33 +53,40 @@ impl Stream for Debounce {
       inner.event_detected.store(false, Ordering::Relaxed);
       Poll::Ready(Some(()))
     } else {
-      let _ = inner.interval.poll_tick(cx);
-      Poll::Pending
+      match Pin::new(&mut inner.delay).poll(cx) {
```
bartlomieju, line 56: can you `Pin` the `Debounce.delay` instead?

cli/file_watcher.rs
```diff
@@ -53,33 +53,40 @@ impl Stream for Debounce {
...
   }
 }
 
-pub async fn watch_func<F>(
-  paths: &[PathBuf],
-  closure: F,
+pub async fn watch_func<F, G>(
+  target_resolver: F,
+  operation: G,
+  job_name: &str,
```
bartlomieju, line 78: Those arguments could use some comments, it's not immediately obvious what's their purpose

cli/file_watcher.rs
magurotuna: I tried but I didn't know how I could make that compile, so instead I use `poll_unpin` here. How about this?

bartlomieju: Even better

cli/file_watcher.rs
magurotuna: I added some comments to describe watcher functions including their parameters. PTAL 8feed3c

magurotuna:
Not sure the cause of flakiness in CI on macOS (pointed in #8316), but I attempted to increase timeout value.

bartlomieju (commented):
@magurotuna looks good! I have a few nitpicks, but I will hold them off for now. Let's land https://github.com/denoland/deno/pull/8375 first - it should cleanup up a few bits of this PR already.

bartlomieju:
#8375 has landed, @magurotuna could you rebase this PR?

magurotuna:
@bartlomieju Finished rebasing. Should the PR title be changed? Maybe "feat(bundle, fmt): Support --watch flag" or something, though it would lack the information the current title has :/

cli/file_watcher.rs
```diff
@@ -18,22 +18,22 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
 use std::time::Duration;
 use tokio::select;
-use tokio::time::{interval, Interval};
+use tokio::time::{delay_for, Delay};
 
 const DEBOUNCE_INTERVAL_MS: Duration = Duration::from_millis(200);
 
 // TODO(bartlomieju): rename
-type WatchFuture = Pin<Box<dyn Future<Output = Result<(), AnyError>>>>;
+type WatchFuture<T> = Pin<Box<dyn Future<Output = Result<T, AnyError>>>>;
```
bartlomieju, line 26: Let's address this TODO, how about `FileWatcherFuture`?

cli/file_watcher.rs
```diff
@@ -53,33 +53,54 @@ impl Stream for Debounce {
...
     eprintln!("{}", msg);
   }
 }
 
-pub async fn watch_func<F>(
-  paths: &[PathBuf],
-  closure: F,
+/// This function adds watcher functionality to subcommands like `fmt` or `lint`.
+/// The difference from [`watch_func_with_module_resolution`] is that this doesn't depend on
+/// [`ModuleGraph`] stuff.
```
bartlomieju, line 77: remove `stuff`

cli/file_watcher.rs
```diff
@@ -90,11 +111,68 @@ where
...
+      info!(
+        "{} File change detected! Restarting!",
+        colors::intense_blue("Watcher"),
+      );
     }
+  }
+}
+
+/// This function adds watcher functionality to subcommands like `run` or `bundle`.
+/// The difference from [`watch_func`] is that this does depend on [`ModuleGraph`] stuff.
```
bartlomieju, line 132: remove `stuff`

cli/flags.rs
```diff
@@ -360,6 +360,7 @@ fn types_parse(flags: &mut Flags, _matches: &clap::ArgMatches) {
 }
 
 fn fmt_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
+  flags.watch = matches.is_present("watch");
```
bartlomieju, line 363: Would be great to have unit test added for respective commands

cli/tests/integration_tests.rs
```diff
@@ -1142,6 +1189,58 @@ fn bundle_import_map_no_check() {
...
+    .unwrap()
+    .contains("File change detected!"));
+  assert!(stderr_lines.next().unwrap().contains("file_to_watch.js"));
+  assert!(stderr_lines.next().unwrap().contains("mod6.bundle.js"));
+  let file = PathBuf::from(&bundle);
+  assert!(file.is_file());
+  assert!(stderr_lines.next().unwrap().contains("Bundle finished!"));
+
+  deno.kill().unwrap();
+  drop(t);
```
bartlomieju, line 1241: Nice tests

cli/tests/integration_tests.rs
```diff
@@ -1215,7 +1314,31 @@ fn run_watch() {
...
+  assert!(stderr_lines.next().unwrap().contains("Process finished"));
+
+  // Confirm that restarting occurs when a new file is updated
+  std::fs::write(&another_file, "export const foo = 42;")
+    .expect("error writing file");
+
+  std::thread::sleep(std::time::Duration::from_secs(1));
+  assert!(stderr_lines.next().unwrap().contains("Restarting"));
+  assert!(stdout_lines.next().unwrap().contains("42"));
+  assert!(stderr_lines.next().unwrap().contains("Process finished"));
```
bartlomieju, line 1341: 👍  great to this fixed

cli/main.rs
```diff
@@ -559,42 +619,49 @@ async fn run_from_stdin(flags: Flags) -> Result<(), AnyError> {
...
-    // will use cached source files
-    let gs = ProgramState::new(flags.clone()).unwrap();
-    let permissions = Permissions::from_flags(&flags);
-    let main_module = main_module.clone();
+  let operation = |main_module: ModuleSpecifier| {
+    let flags = flags.clone();
     async move {
+      let permissions = Permissions::from_flags(&flags);
+      // FIXME(bartlomieju): ProgramState must be created on each restart - otherwise file fetcher
+      // will use cached source files
```
bartlomieju, line 662: This problem should be fixed by https://github.com/denoland/deno/pull/8328, we must be careful on the timing between the two PRs

bartlomieju (commented):
@magurotuna great work! I really like the API you came up with - it prevents work duplication. I have only minor comments. I will wait on @kitsonk approval as it touches module graph.

cli/flags.rs
magurotuna: I forgot to do that, thanks! I've added tests for `bundle --watch` and `fmt --watch`.

cli/file_watcher.rs
magurotuna: sounds good 👍

cli/file_watcher.rs
```diff
@@ -90,11 +112,95 @@ where
...
+        paths = Some(next_paths);
+        module = Some(next_module);
+      }
+      Err(e) => {
+        // If at least one of `paths` and `module` is `None`, the watcher cannot decide which files
+        // should be watched. So return the error immediately without watching anything.
+        if paths.is_none() || module.is_none() {
+          return Err(e.into());
+        }
+      }
```
magurotuna, line 179: To address #7848, I fixed `watch_func_with_module_resolution` implementation so that the watcher will try to continue watching files using _previously resolved_ data.

This works fine if the target files initially have correct syntax and then updated into invalid syntax.
This also means that the watcher won't work at all if the target files initially have invalid syntax. I think it cannot be helped because the watcher cannot decide which files should be watched in this case.

magurotuna: If there's no problem with my approach, I'll add some integration tests to confirm this behavior. WDYT? @bartlomieju

bartlomieju: @magurotuna that sounds good! Test would be great

magurotuna:
I added tests for `run` and `bundle` in the following cases:

- watcher exits immediately if module resolution fails at the first attempt due to invalid syntax
- watcher tries to continue watching files using previously resolved data

magurotuna:
hmmm..... tests are sooooo flaky :(

bartlomieju:
> hmmm..... tests are sooooo flaky :(

@magurotuna are you talking about tests for this PR or tests in general?

magurotuna:
@bartlomieju 
Seems like both of them.

1. https://github.com/denoland/deno/runs/1413182650#step:16:3855 -> operation timeout on macOS, which happened at the place unrelated to this PR
2. https://github.com/denoland/deno/runs/1416308269#step:14:12158 -> rusty_v8 build failed on ubuntu, totally unrelated to this PR
3. https://github.com/denoland/deno/runs/1416415130#step:16:3862 -> `fmt_watch_test` failed on macOS. This seems to be related to this PR and #8316.

I have no idea why `fmt_watch_test` failed. I tried to run the tests on my local mac, they were totally fine.

bartlomieju:
@magurotuna thanks for detailed description. I'm aware of the first problem, but haven't found the reason yet. Second failure seems like an intermittent failure. As for third I'm not sure, let's keep a close eye on it.

magurotuna:
Several times of resolving conflict have been a little bit grunt work 😢

bartlomieju:
> Several times of resolving conflict have been a little bit grunt work 😢

Sorry for that @magurotuna, I'll try to land your PR this weekend.

Can I ask you to take a look [at this comment](https://github.com/denoland/deno/pull/8276#discussion_r523464871) and check if that's still the case?

magurotuna:
@bartlomieju Sure, I'll check for that.

bartlomieju:
@magurotuna I've tested this PR locally and it's working very well! Code in this PR is a big improvement compared to current implementation and LGTM. Please prepare commit message before landing :)

cli/main.rs
magurotuna: I tried to check if this is still the case or not, but it seemed impossible to create `ProgramState` only once before the watcher starts, because of lifetime and trait bound restriction.
So effectively `ProgramState` ended up being created every time the watcher restarts. I think it's okay, so just removed the FIXME comment.

magurotuna:
@bartlomieju 
Thanks for your test!

> Please prepare commit message before landing :)

Sorry if I misunderstand it, but does it mean that you want me to write a commit message that will be used in a squashed merge commit?

bartlomieju:
> Sorry if I misunderstand it, but does it mean that you want me to write a commit message that will be used in a squashed merge commit?

Yes, exactly

magurotuna:
@bartlomieju 
Sure, I've added a brief message to the head of the PR's body. Is this fine?

bartlomieju (approved):
LGTM

PR #8281, opened 2020-11-07
feat: only define Node.js globals when loading std/node/global

guybedford:
This updates the Node.js `process` and `Buffer` shims to only define their global names when importing `std/node/global`.

This makes it possible to import these libraries without causing global mutations.

I have left the `import './global.ts'` in `module.ts` though as that makes sense to keep for the `createRequire` use cases for now and to avoid a breaking change for those use cases.

This is a breaking change for importers that only import buffer or process though.

### DIFF
--- a/std/node/buffer.ts
+++ b/std/node/buffer.ts
@@ -599,10 +599,3 @@ export default class Buffer extends Uint8Array {
 }
 
 export { Buffer };
-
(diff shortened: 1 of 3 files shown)


guybedford:
The MacOS CI failure seems unrelated as far as I can tell.

lucacasonato (approved):
LGTM. Thanks

PR #8282, opened 2020-11-07
docs(std/path): add JSDoc

mrmodise:
Add missing JSDoc for std/path/posix.ts and std/path/win32.ts

Related to #7487

### DIFF
--- a/std/path/posix.ts
+++ b/std/path/posix.ts
@@ -16,7 +16,10 @@ export const sep = "/";
 export const delimiter = ":";
 
 // path.resolve([from ...], to)
-/** resolves `pathSegments` into an absolute path.  */
+/**
+ * Resolves `pathSegments` into an absolute path.
+ * @param pathSegments an array of path segments
+ */
 export function resolve(...pathSegments: string[]): string {
   let resolvedPath = "";
   let resolvedAbsolute = false;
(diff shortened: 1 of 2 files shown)


mrmodise:
@kitsonk any chance you can look at this. Thanks

std/path/posix.ts
```diff
@@ -16,7 +16,10 @@ export const sep = "/";
 export const delimiter = ":";
 
 // path.resolve([from ...], to)
-/** resolves `pathSegments` into an absolute path.  */
+/**
```
ThenMorning, line 19: Excuse me, where did you find these comments?
https://deno.land/std@0.78.0/path/posix.ts
I didn't find it.

mrmodise: Hi @ThenMorning the comments were added a while back. See latest version of the same file: https://github.com/denoland/deno/blob/ab7b8ba8d15f37d6244784a16a3f4f3c96e38769/std/path/posix.ts#L19

ThenMorning: Sorry,i mean where did you find these comments？
where you copied from  or  you  texted?
I want to resolve 'ws/mod.ts' and have no idea to find it in standard library .

mrmodise: ah! i reused the existing comments since am also new. But you can provide some and the team will be able to correct if its incorrect

ThenMorning: Ok.So where is the `existing comments' ?
For example `path/posix.ts`.
Thanks!

mrmodise: https://github.com/denoland/deno/tree/v1.5.3

ThenMorning: https://github.com/denoland/deno/blob/v1.5.3/std/path/posix.ts ？
Sorry, I don't understand.
They have same version, both them are '0.78.0'.
where is the `existing comments`?

bartlomieju (approved):
LGTM, thanks @mrmodise

PR #7593, opened 2020-09-20
chore: add copyright

tokiedokie:
adds copyright comment in files that don't have one

PR #8283, opened 2020-11-07
build: check license header in CI

kt3k:
This PR adds an automatic check of the license headers of source files to `tools/lint.js`. This change uses my small cli tool [x/license_checker](https://deno.land/x/license_checker@v3.0.4).

I'm a little concerned about the slowness of this check. `./tools/lint.js --rs --js` takes about 1.3s on my machine and `./tools/lint.js --license` takes about 3.7s...

ref: https://github.com/denoland/deno/pull/7593#issuecomment-696219191


nayeemrmn:
Given the speed issue, maybe it should be a separate script from `tools/lint.js`?

kt3k:
@nayeemrmn That sounds reasonable. I created a separate file `tools/check_license.js` for this check.

.licenserc.json
```diff
@@ -0,0 +1,12 @@
...
+  "ignore": [
+    "testdata",
+    "cli/dts",
+    "cli/tests",
+    "std/node/tests",
+    "std/hash/_wasm/wasm.js",
+    "cli/tsc/00_typescript.js",
+    "target"
+  ]
+}
```
bartlomieju, line 12: While I'm strongly in favor of this PR, I'm not in favor of adding new dotfile and I'm pretty sure @ry will be of the same opinion. Is there a way to not use this file and pass arguments to the script explicitly?

kt3k: Separated the core routine of `x/license_checker` into the method `checkLicense` (ref: https://github.com/kt3k/deno_license_checker/pull/7/files ). Now this PR uses that method. The config is passed through function args, instead of using dotfiles.

ry:
@kt3k Thanks for the patch, but I don't really want to introduce to this extra complexity for this relatively simple problem. Also I think sometimes when we pull in external code but make very few changes, it makes sense to just maintain the original copyright header and not add our own.
Closing without merge.

PR #8286, opened 2020-11-07
fix(cli/upgrade): Auto detect target triples for upgrade

crowlKats:
Should I remove the `test_compose_url_to_exec` test? cant see much use for it


lucacasonato (approved):
LGTM - this should also fix the issue where running `deno upgrade` on a ARM64 system will try to download a x64 binary. Thanks

Issue #8254, opened 2020-11-05
Rewrite tests to not depend on Python


bartlomieju:
After https://github.com/denoland/deno/issues/8223 we no longer rely on Python
for development (unless you're building `rusty_v8` from source). However there
are still quite a few tests that use Python (main for subprocess tests):
- `cli/tests/permission_tests.ts`
- `cli/tests/unit/chown_test.ts`
- `cli/tests/unit/process_test.ts`

It'd be great if we could eventually remove dependency on Python altogether.

mrmodise:
@bartlomieju I will take `cli/tests/unit/chown_test.ts`. If all goes well I will take the rest.

mrmodise:
Had to split them. The process_test.ts needs a bit of attention

bartlomieju:
Solution used in https://github.com/denoland/deno/pull/8404 seems preferable.

bartlomieju:
Fixed in #8291, #8841 and #8843!

waldyrious:
@bartlomieju would it be adequate to change the lines below as well?

https://github.com/denoland/deno/blob/4033b390367e025e73c851a6ce7021631b27e3c0/cli/dts/lib.deno.unstable.d.ts#L1043-L1047

I was thinking something like this to mimic the change in #8291 and #8843:

```diff
    *      const p = Deno.run({
-   *        cmd: ["python", "-c", "from time import sleep; sleep(10000)"]
+   *        cmd: ["sleep", "10000"]
    *      });
```

WDYT?

bartlomieju:
@waldyrious that sounds good, PRs welcome :)

waldyrious:
@bartlomieju I hope you don't mind — since it's a very small change, I embedded it in another PR I was preparing :) I've submitted the changes in #8848.

PR #8290, opened 2020-11-08
refactor: partially rewrite chown_test to not depend on Python

mrmodise:
Related to #8254

### DIFF
--- a/cli/tests/unit/chown_test.ts
+++ b/cli/tests/unit/chown_test.ts
@@ -12,7 +12,7 @@ async function getUidAndGid(): Promise<{ uid: number; gid: number }> {
   // get the user ID and group ID of the current process
   const uidProc = Deno.run({
     stdout: "piped",
-    cmd: ["python", "-c", "import os; print(os.getuid())"],
+    cmd: [Deno.execPath(), "eval", "-p", "--unstable", "Deno.ppid"],
   });
   const gidProc = Deno.run({
     stdout: "piped",


lucacasonato:
Could you do all of the related changes in one PR? Opening a new PR for each one line change is rather tedious to review :-)

mrmodise:
> Could you do all of the related changes in one PR? Opening a new PR for each one line change is rather tedious to review :-)

Cool. I need help though ;-)

I couldnt find a way to get the gid in Deno


```ts
const gidProc = Deno.run({
    stdout: "piped",
    cmd: ["python", "-c", "import os; print(os.getgid())"],
  });
```

lucacasonato:
@mrmodise I don't think we have a call for `uid` or `gid` at the moment. You are using `Deno.ppid`,  but that is not the same as `os.getuid()`. `ppid` is the parent process ID, and `uid` is the user id.

mrmodise:
> @mrmodise I don't think we have a call for `uid` or `gid` at the moment. You are using `Deno.ppid`, but that is not the same as `os.getuid()`. `ppid` is the parent process ID, and `uid` is the user id.

ah! okay so i can leave this out. I will do the same to few other ones I picked in `process_test.ts`.

mrmodise:
1 last one, how do i run specific tests? I notice `cargo test` run all test suites, whereas my preference would be to run the specific files am working on. Currently, I have to do the heavy-lifting of copying the specific test into `Deno.test()` then `deno test...`

bartlomieju:
@mrmodise take a look at https://github.com/denoland/deno/blob/master/cli/tests/unit/README.md

mrmodise:
> @mrmodise take a look at https://github.com/denoland/deno/blob/master/cli/tests/unit/README.md

I missed it on the README 🤦🏾 . Thanks @bartlomieju

bartlomieju:
@mrmodise thanks for the PR, but I'm gonna close without merge. We need to find "universal" command that will work on all OSes used in CI instead of relying on execution of `deno` binary (similar to https://github.com/denoland/deno/pull/8404)

PR #8291, opened 2020-11-08
refactor: rewrite permission_test to not depend on Python

mrmodise:
Related to #8254

### DIFF
--- a/cli/tests/permission_test.ts
+++ b/cli/tests/permission_test.ts
@@ -16,13 +16,10 @@ const test: { [key: string]: (...args: any[]) => void | Promise<void> } = {
     Deno.listen({ transport: "tcp", port: 4541 });
   },
   runRequired(): void {
-    Deno.run({
-      cmd: [
-        "python",
-        "-c",
-        "import sys; sys.stdout.write('hello'); sys.stdout.flush()",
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
Thanks

mrmodise:
> Thanks

No, thank you 😊 . I just came from doing the long hanging fruits aka good-first-issues

iammack:
What do I need to change to be able to run this in Windows 10. It's failing for me on cargo test.

cli/tests/permission_test.ts
```diff
@@ -16,13 +16,10 @@ const test: { [key: string]: (...args: any[]) => void | Promise<void> } = {
...
   },
   runRequired(): void {
-    Deno.run({
-      cmd: [
-        "python",
-        "-c",
-        "import sys; sys.stdout.write('hello'); sys.stdout.flush()",
-      ],
+    const p = Deno.run({
+      cmd: ["printf", "hello"],
```
avindra, line 20: @iammack `printf` is probably not available on Windows.


`echo` exists in DOS so that may be an x-platform way to write this.

Disclosure: I'm not a Windows user.

PR #8295, opened 2020-11-08
build: use shallow clone of third_party

bartlomieju:
`third_party/` is a git submodule pointing to `deno_third_party` repository. The git history of that repo is quite heavy as it historically contained `typescript`, `node_modules` and `python_packages` directories, but now there are only 11 binaries in there. Setting this attribute should speed up fresh clones.


piscisaureus (approved):
👍

PR #8300, opened 2020-11-08
docs: add TS on the server side in Deno talk

trivikr:


lucacasonato:
You think so? I feel I didn't really explain TypeScript's role in Deno very well in that talk.

bartlomieju:
> You think so? I feel I didn't really explain TypeScript's role in Deno very well in that talk.

I agree, I'd rather see talks collected on a single page instead.

ry:
Happy to add a link elsewhere, I don't think it makes sense here. Closing without merge.

PR #8306, opened 2020-11-09
fix(cli): allow setting of importsNotUsedAsValues in Deno.compile()

### LINKED ISSUE #6663: Deno.bundle() should accept the compiler flag `importsNotUsedAsValues`
This `tsc` compiler flag was implemented in TypeScript [3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html). You can find the merged PR here: https://github.com/microsoft/TypeScript/pull/35200

This flag takes 3 different values:

- remove: this is today’s behavior of dropping these imports. It’s going to continue to be the default, and is a non-breaking change.
- preserve: this preserves all imports whose values are never used. This can cause imports/side-effects to be preserved.
- error: this preserves all imports (the same as the preserve option), but will error when a value import is only used as a type. This might be useful if you want to ensure no values are being accidentally imported, but still make side-effect imports explicit.

It would be very helpful for Web Components and other stuff.

kitsonk:
Fixes #6663

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -318,6 +318,20 @@ declare namespace Deno {
     /** Import emit helpers (e.g. `__extends`, `__rest`, etc..) from
      * [tslib](https://www.npmjs.com/package/tslib). */
     importHelpers?: boolean;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

PR #8307, opened 2020-11-09
fix(cli): only use tsbuildinfo with valid emit

kitsonk:
I think this fixes the problem that @lucacasonato and others have had when moving from a cache from 1.4 to 1.5 and having some of the emitted sources missing, maybe, but since we can't reliably reproduce it, it is hard to tell if this fixes it.  Basically if we don't have something emitted, we won't load a `.tsbuildinfo` because it can't be possible "correct".

### DIFF
--- a/cli/disk_cache.rs
+++ b/cli/disk_cache.rs
@@ -134,6 +134,11 @@ impl DiskCache {
     }
   }
 
+  pub fn is_file(&self, filename: &Path) -> bool {
+    let path = self.location.join(filename);
+    path.is_file()
+  }
+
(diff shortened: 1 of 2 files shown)


kitsonk:
Actually, while working on something else, I realised the problem is... 

We write out the `.tsbuildinfo` even when the compilation files, but we don't write out the emitted files.  In situations where you then "fix" the problem, none of the dependencies will have been emitted, but the `.tsbuildinfo` is "valid", therefore `tsc` doesn't emit anything, because the source has not changed for the dependencies.

I need to re-work this.

PR #8310, opened 2020-11-09
fix(cli): allow root modules be .mjs/.cjs

### LINKED ISSUE #6176: error: cjs / mjs files are compiled as TypeScript
Receiving TS2531 [ERROR] when running the mjs/cjs after Deno 1.0.4.

save as "test.mjs"
```
import { exp } from "./test_mod.ts";
if (false) null.func();
```
save as "test_mod.ts"
```
export const exp = {};
```
and run
```
% deno run test.mjs
Compile file:/// ... /test.mjs
error: TS2531 [ERROR]: Object is possibly 'null'.
if (false) null.func();
           ~~~~
    at file:/// ... /test.mjs:2:12
```

1.0.5 -- NG
1.0.4 -- NG
1.0.3 -- OK

without ts import -- OK
import mjs -- OK

kitsonk:
Fixes #6176

### DIFF
--- a/cli/tsc.rs
+++ b/cli/tsc.rs
@@ -22,6 +22,7 @@ use deno_core::Snapshot;
 use serde::Deserialize;
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::path::PathBuf;
 use std::rc::Rc;
 
 /// Provide static assets that are not preloaded in the compiler snapshot.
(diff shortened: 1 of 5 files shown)


cli/tsc.rs
```diff
@@ -63,6 +64,44 @@ fn get_maybe_hash(
...
+        Some(os_str) => match os_str.to_str() {
+          Some(file_name) => {
+            if file_name.ends_with(".d") {
+              MediaType::Dts
+            } else {
+              MediaType::TypeScript
+            }
+          }
+          None => MediaType::TypeScript,
+        },
```
ry, line 94: nit: this `match path.file_stem()` could maybe be written more cleanly using `if let Some(...) `

bartlomieju (approved):
Agree with Ryan's comment, otherwise LGTM

PR #8311, opened 2020-11-09
fix(cli): do not write tsbuildinfo when diagnostics are emitted

### LINKED ISSUE #8309: Incorrectly writing tsbuildinfo when check fails
We are incorrectly writing out `.tsbuildinfo` when a build fails, but we don't write out the emitted files.

This means that if you have a problem with code, and you fix it, you can end up in a situation where for some/all of your dependencies their emits don't get committed to disk, but the `.tsbuildinfo` does.  When the issue is fixed, the `.tsbuildinfo` exists and is "valid" which but some of the dependencies haven't changed, and so `tsc` things it doesn't need to do any emits for them, and therefore the emits go "missing".

We need to ensure that we don't flush tsbuildinfo if the module it is associated with is not dirty.

kitsonk:
Fixes #8309

### DIFF
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -800,49 +800,51 @@ impl Graph {
     graph.maybe_tsbuildinfo = response.maybe_tsbuildinfo;
     // Only process changes to the graph if there are no diagnostics and there
     // were files emitted.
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM

PR #8312, opened 2020-11-09
feat(std): add Deno.totalMems

jopemachine:
### Related Issue

#3802

### Usage

```js
// It shows the total amount of memory on the local computer.
console.log(Deno.totalMems());
```

### Notes

I'm a novice developer who is not familiar with open source yet.
I'd be grateful if you could give me any kind of advice :)

I'd like to contribute to creating more Deno's node compatibility modules.


ry (commented):
Thank you for the patch, but this functionality is already supported by `Deno.systemMemoryInfo().total`. We'd gladly accept a patch to `std/node/os.ts` to use `systemMemoryInfo()`.

PR #8317, opened 2020-11-09
feat(std): add std/node/os/totalmem, freemem

jopemachine:
### Related Issue

#3802

### Notes

* I would appreciate it if you let me know if we have any other works to do like creating some documents :)


ry:
Thanks for the patch - looks good. Please merge with master branch because currently there's a conflict in std/node/os.ts

ry (approved):
Thanks! LGTM

PR #8318, opened 2020-11-09
chore: reference dprint config file in format script

bartlomieju:
Explicitly pass path to config while when invoking `dprint` in `tools/format.js`.
This is completely cosmetic change. Done because `.dprintrc.json` wasn't
referenced anywhere, unlike `.rustfmt.toml` which can also be picked up
automatically.

### DIFF
--- a/.dprintrc.json
+++ b/.dprintrc.json
@@ -13,6 +13,7 @@
   "includes": ["**/*.{ts,tsx,js,jsx,json,md}"],
   "excludes": [
     ".cargo_home",
+    ".git",
     "cli/dts/lib.d.ts",
     "cli/dts/lib.dom*",
     "cli/dts/lib.es*",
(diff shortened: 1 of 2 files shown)


ry (approved):

mvolkmann:
Will this be in deno 1.5.5?
Without using `--config`, does `deno fmt` look for configuration in a particular file?
I tried creating `.dprintrc.json` and `dprint.config.json`, but neither affects formatting.

bartlomieju:
> Will this be in deno 1.5.5?
> Without using `--config`, does `deno fmt` look for configuration in a particular file?
> I tried creating `.dprintrc.json` and `dprint.config.json`, but neither affects formatting.

@mvolkmann this PR concerns only internal tooling for the repo and has no effect on `deno fmt`.

PR #8320, opened 2020-11-09
fix(websocket) correct DOM handler order

benjamingr:
This fix is similar to https://github.com/denoland/deno/pull/8230 but for `WebSocket`, just like `AbortSignal`s websocket's `onmessage/onclose/onerror/onopen` are HTML "event handlers" which means they have very particular order requirements.

Before this PRs event handlers were always fired before event listeners. Now they are fired in the correct order according to the spec and a test was added to ensure that for `onmessage`. I am happy to add more tests if you want but I didn't want to add lots of I/O causing tests.

I'm trying to keep my PRs small - but I'd like to follow up on this with a similar fix for workers and with unit tests for `web_utils`.

### DIFF
--- a/cli/rt/01_web_util.js
+++ b/cli/rt/01_web_util.js
@@ -113,9 +113,44 @@
     }
   }
 
+  const handlerSymbol = Symbol("eventHandlers");
+  function makeWrappedHandler(handler) {
+    function wrappedHandler(...args) {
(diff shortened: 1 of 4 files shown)


kitsonk (approved):
LGTM... really appreciate the contributions.

PR #8322, opened 2020-11-09
refactor(std): improve performance by using some instead filter method

behnammodi:
I replaced `filter` with `some`.

We don't need to investigate all of the items in array


kitsonk (approved):
LGTM

Issue #8304, opened 2020-11-09
`cargo test` on aarch64 fails (for an easy to fix reason).


Dygear:
I've been building deno on a Raspberry Pi 4 (8GB) with Raspberry Pi OS 64bit. The build process works very well and produces a build without issues. There is 1 small area where `cargo test` fails on the aarch64 build. The failure is listed below.

```
failures:

---- upgrade::test_compose_url_to_exec stdout ----
thread 'upgrade::test_compose_url_to_exec' panicked at 'assertion failed: `(left == right)`
  left: `"https://github.com/denoland/deno/releases/download/v0.0.1/deno-aarch64-unknown-linux-gnu.zip"`,
 right: `"https://github.com/denoland/deno/releases/download/v0.0.1/deno-x86_64-unknown-linux-gnu.zip"`', cli/upgrade.rs:290:3
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    upgrade::test_compose_url_to_exec

test result: FAILED. 266 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

The problem is on [this line in cli/upgrade.rs290](https://github.com/denoland/deno/blob/master/cli/upgrade.rs#L290) where it doesn't check against [`ARCHIVE_NAME`](https://github.com/denoland/deno/blob/master/cli/upgrade.rs#L33) while building the stings to compare left to right. Other than that, all test pass with flying colors. With Apple's ARM product announcement happening on the 10th, it would be nice to have this cleared up for all platforms and not just the Linux platform.

lucacasonato:
As @crowlKats suggested previously, let's remove that test. It doesn't really test anything useful.

benjamingr:
Looks like this issue can be closed since https://github.com/denoland/deno/pull/8329 landed?

PR #8329, opened 2020-11-10
fix(cli/upgrade) Remove an unnecessary test 

mitch292:
Resolves issue #8304 

Per issue comment thread, I removed the `test_compose_url_to_exec` test.


lucacasonato (approved):
Thanks

Dygear:
With this test removed, it will make it so the AArch64 build on Linux passes all tests. Could we possibly start moving towards a AArch64 official build please?

PR #8334, opened 2020-11-10
fix(web/worker) define event handlers in DOM order

benjamingr:
Hey again :]

This is a similar fix to https://github.com/denoland/deno/pull/8320 and https://github.com/denoland/deno/pull/8264 for Workers.

After this - I want to add more tests and find something small to contribute that will force me to write a bit of rust (which as you can probably tell from https://github.com/denoland/deno/pull/8209 I am not great at).

I also want to add some AbortSignal improvements we have in node that improve developer experience and I have some safety concerns regarding cancellation (related to the email we talked in @ry) - though I want to wait for TC39 to meet next week and see if we have something from them first.

### DIFF
--- a/cli/rt/11_workers.js
+++ b/cli/rt/11_workers.js
@@ -4,6 +4,7 @@
   const core = window.Deno.core;
   const { Window } = window.__bootstrap.globalInterfaces;
   const { log } = window.__bootstrap.util;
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM - thanks Benjamin! 

Happy to have abort controller added to any web APIs to bring us closer to browser behavior; and interested in discussing using it elsewhere.

PR #8337, opened 2020-11-10
fix(op_crates/web) make isTrusted not constructable

benjamingr:
Basically, I wanted to ask myself "What is the most esoteric spec-required property of Event that I know?" and then fixed it in Deno.

Props goes to user Exe-Boss who found this bug in Node (in another, unrelated PR I did) ^^

(Added a test, ran in master and it fails, ran linter formatter etc)


ry (approved):
beautiful, thank you

PR #8347, opened 2020-11-11
build: upload artifacts on master branch

bartlomieju:


bartlomieju:
Not feasible approach, closing without merge.

PR #8348, opened 2020-11-11
fix(op_crates/web) correct filereader event handler order

benjamingr:
Just like AbortSignal, WebSocket and Worker the event handler order of `FileReader` is currently incorrect in Deno.

This is the same fix as https://github.com/denoland/deno/pull/8334 but for FileReader's events :]

As a bonus, FileReader objects are now smaller (though I doubt FileReader is Deno's performant File API anyway :])


lucacasonato (approved):
LGTM. Thanks for the contributions @benjamingr

PR #8349, opened 2020-11-11
fix(std): Flush individual chunks in HTTP chunked encoding

### LINKED ISSUE #8339: Unexpecting req.respond body Deno.Reader behaviour.
I try to do a Progressive Rendering HTML with http1.1 "Transfer-Encoding": "chunked".

To try that, I create a fake latency of 100ms on each chunk. But the request send all the chunk at the same time after 1200ms.
I have 12 chunks.

But it should send each chunk each 100ms. Not all chunks at the same time.

To illustrate what I expected, I've wrote the "same" code in Node. and make gifs.

Expected => [node/ExpressJS version](https://github.com/OverGlass/nodeSSRTest)

this

My deno code :
```typescript
import { serve } from "https://deno.land/std@0.76.0/http/server.ts";
import type { ServerRequest } from "https://deno.land/std@0.76.0/http/server.ts";
import { encode } from "https://deno.land/std@0.73.0/encoding/utf8.ts";
const s = serve({ port: 8000 });
console.log("http://localhost:8000/");

for await (const req of s) {
  $progressiveRendering(req);
}

async function $progressiveRendering(req: ServerRequest) {
  const xmlChunks = [
    "<html><body>",
    "<h1>Hell0 world</h1>",
    "<h2>Hello world</h2>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "<h3>Hello world</h3>",
    "</body></html>",
  ];

  const reader: Deno.Reader = {
    async read(p: Uint8Array) {
      if (!xmlChunks.length) {
        return null;
      } else {
        const chunk = encode(xmlChunks.shift());
        //Should sent a chunk each 100ms
        await sleep(100);
        p.set(chunk);
        return chunk.length;
      }
    },
  };

  async function sleep(ms: number) {
    return new Promise((resolve) => {
      const waitT = setTimeout(() => {
        clearTimeout(waitT);
        resolve();
      }, ms);
    });
  }
  req.respond({
    body: reader,
    status: 200,
    headers: new Headers({
      "Cache-Control": "no-cache",
      "Content-Type": "text/html; charset=utf-8",
      Connection: "keep-alive",
      "X-Content-Type-Options": "nosniff",
      "Transfer-Encoding": "chunked",
    }),
  });
}
```
Result : 

this

What am I missing ?
It is an Deno issue, or it's me the issue ? :p

MarkTiedemann:
Fixes #8339.

`writeChunkedBody` requires flushing the writer after each written chunk. This was not possible with the previous function signature that required a `Deno.Writer` which does not have a `.flush` method. Since `writeChunkedBody` was called with a `BufWriter` (which implements `Deno.Writer`), the function signature was updated to require a `BufWriter` instead. It is also no longer necessary to wrap the existing `BufWriter` `w` in another `BufWriter` `writer = BufWriter.create(w)`.

A minimal reproduction may be found here: https://github.com/denoland/deno/issues/8339#issuecomment-725601429

### DIFF
--- a/std/http/_io.ts
+++ b/std/http/_io.ts
@@ -171,21 +171,21 @@ function parseTrailer(field: string | null): Headers | undefined {
 }
 
 export async function writeChunkedBody(
-  w: Deno.Writer,
+  w: BufWriter,
   r: Deno.Reader,
 ): Promise<void> {
-  const writer = BufWriter.create(w);
   for await (const chunk of Deno.iter(r)) {
     if (chunk.byteLength <= 0) continue;
     const start = encoder.encode(`${chunk.byteLength.toString(16)}\r\n`);
     const end = encoder.encode("\r\n");
-    await writer.write(start);
-    await writer.write(chunk);
-    await writer.write(end);
+    await w.write(start);
+    await w.write(chunk);
+    await w.write(end);
+    await w.flush();
   }
 
   const endChunk = encoder.encode("0\r\n\r\n");
-  await writer.write(endChunk);
+  await w.write(endChunk);
 }
 
 /** Write trailer headers to writer. It should mostly should be called after


std/http/_io.ts
```diff
@@ -171,21 +171,21 @@ function parseTrailer(field: string | null): Headers | undefined {
 }
 
 export async function writeChunkedBody(
-  w: Deno.Writer,
+  w: BufWriter,
```
kitsonk, line 174:
```suggestion
  w: Deno.Writer & { flush(): Promise<void> }
```

This would allow other "flushable" writers without specifically requiring a BufWriter. Maybe even creat and export a `Flushable` interface and export it and then require `Deno.Writable & Flushable`.

MarkTiedemann: Hmm, is this really an improvement? I think the previous interface being needlessly unspecific created the situation where a `BufWriter` was wrapped in a `BufWriter` unnecessarily.

I also feel like both `Deno.Writer & { flush(): Promise<void> }` and `Deno.Writable & Flushable` are harder to read and understand than `BufWriter`.

kitsonk: But it isn't always a BufWriter, BufWriter is an implementation of a concept and may not always be what is used downtream. The previous interface wasn't causing the problem, the code was not flushing the data, irrespective of the interface. There is a reason why we use `Deno.Writer` instead of a specific  implementation.

MarkTiedemann: I know the `writeChunkedBody` function is exported, but I expected it to be exported only for testing purposes (though it doesn't have any tests at the moment).

It's only called in one place in the same file. To me, it looks more like a function that was created to make `writeResponse` a bit shorter and easier to digest.

There's nothing technically wrong with using `Deno.Writer & { flush(): Promise<void> }` rather than `BufWriter`. I just think that that interface is a bit overengineered for a function that could be inlined.

kitsonk: Yeah, ok, fair enough, in that it is exported, but from a "private" module, so it is fine coupling to an implementation interface.

kitsonk (approved):
LGTM

Issue #7969, opened 2020-10-14
deno_core shouldn't use "get_identity_hash()" to identify v8 objects


bartlomieju:
Ref https://github.com/denoland/deno/pull/7946#discussion_r504192165

bartlomieju:
Ref https://github.com/denoland/deno/issues/1745

bnoordhuis:
I suggested implementing `Hash` and `Eq` with `promise.get_identity_hash()` and `promise.strict_equals(other)` respectively but I realize that means you need a `HandleScope` to unwrap the `v8::Global<v8::Promise>`.

Strictly speaking, you don't need one - a strong `Global` can be transmuted to a `Local`. Something like this could work:
```rust
let a: v8::Global<v8::Promise> = f();
let b: v8::Global<v8::Promise> = g();
let is_eq = a.map(|a| b.map(|b| a.strict_equals(b)));
```
Where `.map()` transmutes to `Local`.

Then again, maybe it's better to just assign ids and store those using private symbols. I've opened https://github.com/denoland/rusty_v8/pull/498 to flesh out `v8::Private` more.

bnoordhuis:
A post on v8-users made me revisit this again. I don't think we need `a.strict_equals(b)` here, we can just compare the `v8::Global` handles.

V8's C++ API has `bool v8::PersistentBase::operator==(const v8::PersistentBase&) const`, which is basically a couple of pointer checks. It returns true when the handles point to the same JS object.

rusty_v8 implements `PartialEq` for `Global` and it behaves the same way, if I read the code right, no need to unbox.

PR #8352, opened 2020-11-11
refactor(core): don't depend on get_identity_hash for promises

bartlomieju:
This commit changes how pending promise exceptions are stored
in JsRuntime by using global handle to promise instead of its
identity hash.

Ref https://github.com/denoland/deno/issues/7969

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -103,7 +103,8 @@ pub(crate) struct JsRuntimeState {
   pub(crate) shared_ab: Option<v8::Global<v8::SharedArrayBuffer>>,
   pub(crate) js_recv_cb: Option<v8::Global<v8::Function>>,
   pub(crate) js_macrotask_cb: Option<v8::Global<v8::Function>>,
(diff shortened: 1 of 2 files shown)


core/runtime.rs
```diff
@@ -1385,7 +1386,14 @@ impl JsRuntime {
...
 
-    let key = { *state.pending_promise_exceptions.keys().next().unwrap() };
+    let key = {
+      state
+        .pending_promise_exceptions
+        .keys()
+        .next()
+        .unwrap()
+        .clone()
+    };
```
bnoordhuis, line 1396: Just curious, is there a reason fro the braces? Is it stylistic?

bnoordhuis (approved):
Nice, LGTM.

core/runtime.rs
bartlomieju: Borrow checker complains about double borrow of `state` if there are no braces.

Issue #6868, opened 2020-07-24
Fetch API doesn't respect forbidden headers


nayeemrmn:
Refs:
- https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
- https://fetch.spec.whatwg.org/#forbidden-header-name

We don't fundamentally need this restriction on the server-side, but for both spec-compliance and practical compatibility with browsers we should force use of a `Deno.HttpClient` to set these headers:
```js
// The Cookie header is silently ignored.
await fetch(
  "https://example.com/",
  { headers: new Headers({ "Accept": "text/html", "Cookie": "foo=bar" }) },
);

// The Cookie header is used.
const client = Deno.createHttpClient({ allowAllHeaders: true });
await fetch(
  "https://example.com/",
  { client, headers: new Headers({ "Accept": "text/html", "Cookie": "foo=bar" }) },
);
```

kitsonk:
The behaviour is to silently drop them, at least in Chrome.  I just did:

```
await fetch("http://localhost:8000", { headers: new Headers([["cookie", "foo=bar"]]) });
```

And it simply stripped the `cookie` header.  The reason why they are forbidden is because the user agent controls them, and can't be overridden by user land.  We should simply document the behaviour in my opinion, since there is no browser error when passed.

nayeemrmn:
> The behaviour is to silently drop them ... And it simply stripped the cookie header

Yep, as I hinted at in the example. But when you author a `fetch()` to a hardcoded URL with a cookie which is prerequisite to getting an `ok` response (auth?), it's likely to propagate into an error anyway, and you end up with a script that is expected to work in Deno but can't work in a browser. This will lead to the same kind of surprise our compatibility rule is supposed to prevent. Code which won't work as expected in a browser should be signalled with a `Deno` reference.

kitsonk:
Ah, I mis-read what you said.  I thought you were suggesting the behaviour of Deno was to silently drop the cookie header, but it is browsers.

The problem is `fetch()` will _always_ behave differently in a server context and a client context.  You have all the situations with same origin and cross origin which just don't make sense in a server context.  I am still just for documenting it.  As you state, Node's implementation just passes forbidden as well.

It is that or we implement some sort of cookie jar per origin that manages the cookies itself, and set the cookies on the document if you "really" want browser compatibility, but that just doesn't seem logical to me.

jsejcksn:
Not stale

nayeemrmn:
[...]

PR #8353, opened 2020-11-11
fix(op_crates/fetch): `redirect: "manual"` fetch should return `type: "default"` response

### LINKED ISSUE #4389: fetch(..., {redirect: 'manual'}) does not work propertly
`https://httpbin.org/status/302` is a url that does a few redirects, ending at `https://httpbin.org/get`.

    fetch('https://httpbin.org/status/302')
      .then(res => console.log(res.status, res.headers))

Follows redirects, prints status 200, headers, as expected.

    fetch('https://httpbin.org/status/302', {redirect: 'manual'})
      .then(res => console.log(res.status, res.headers))

Prints status 0, empty headers.

Expected: status 302, headers including `Location`.

### LINKED ISSUE #8351: Remove opaqueredirect response type in fetch
Currently fetching with the option `redirect: "manual"` results in a `opaqueredirect` response when a redirect is encountered. This means that in Deno (a server side runtime), it is not possible to manually follow a HTTP redirect... :exploding_head:

We should fundamentally rethink how we are implementing the `fetch` API. The reason `opaqueredirect` exists, is to prevent leaking information that is accessible because of the origin cookie jar. We do not have an origin cookie jar, so we do not need to protect against leaks it causes. We thus do not need `opaqueredirect`. We should do what Cloudflare Workers does for `redirect: "manual"`, which is to just respond with the actual response (headers, status code, and body present) - so a `type: "default"` response.

Same goes for all other response types. There is no reason for us to have `cors`, `basic`, or `opaque`. These all exist to prevent against attacks which can not happen in Deno, because we do not have an origin cookie jar.

Sources:
https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

lucacasonato:
This intentionally breaks from spec (but aligns to other server side implementations of `fetch` like Cloudflare Workers).

Closes #8351
Closes #4389

### DIFF
--- a/cli/tests/unit/fetch_test.ts
+++ b/cli/tests/unit/fetch_test.ts
@@ -693,18 +693,10 @@ unitTest(
     const response = await fetch("http://localhost:4546/", {
       redirect: "manual",
     }); // will redirect to http://localhost:4545/
-    assertEquals(response.status, 0);
-    assertEquals(response.statusText, "");
-    assertEquals(response.url, "");
-    assertEquals(response.type, "opaqueredirect");
-    try {
-      await response.text();
-      fail(
-        "Reponse.text() didn't throw on a filtered response without a body (type opaqueredirect)",
-      );
-    } catch (e) {
-      return;
-    }
+    assertEquals(response.status, 301);
+    assertEquals(response.url, "http://localhost:4546/");
+    assertEquals(response.type, "default");
+    assertEquals(response.headers.get("Location"), "http://localhost:4545/");
   },
 );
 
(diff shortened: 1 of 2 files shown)


bartlomieju:
@nayeemrmn what's your opinion on this PR?

nayeemrmn (approved):
Similar to #6868, my preference is that spec-breaking behaviour can only invoked by passing an option with key `Deno.<symbol>`. But there's no such infrastructure yet. If that proposal gets approved we can always revisit this.

ry (approved):
LGTM

kitsonk (approved):
LGTM

Would be good to link/reference any documentation update PR from the comments on this PR.

PR #8321, opened 2020-11-09
styled(std): remove unnecessary if else block

behnammodi:
In the title is explained :)

PR #8360, opened 2020-11-12
ensure_symlink.ts(fs): removed unnecessary code

nurmohammed840:


kitsonk:
This basically reverses #8321 and add 8 more lines of code, and isn't any more clearer.  So respectfully closing.

PR #8363, opened 2020-11-13
fix: (std/node/global) correct typings

Soremwar:
Typings for this weren't binding correctly to `global`, `globalThis` and `window` and keep giving me errors while importing in user code. Also, Buffer wasn't being recognized as a valid type. It's all ok now


Soremwar:
Even though it's passing, it's not working as intended. I'll fix it tomorrow

Soremwar:
After countless stressfull hours (TypeScript is such a pain in the ass sometimes) this is finally passing. The reason as to why I included a `d.ts` file is documented in the file itself

ry (approved):
LGTM - thanks @Soremwar for the countless stressful hours : )

Issue #8107, opened 2020-10-24
https connection times out


timonson:
I have a problem running an `https` server and I can't solve it on my own. I tried various things which include:

- switching linux distros
- using different ssl certificates
- changing server configurations
- different domains
- comparing to http server
- and more

The problem is that after an indeterminable time, but mostly about 20 hours, the server stops working in the sense that requests just seem to timeout: `curl: (28) Failed to connect to xyz.com port 443: Connection timed out`. Before this time period, everything works prefectly. But the deno process is still running and the port is still used:
![deno_ps_ax-2020-10-24](https://user-images.githubusercontent.com/54777088/97083011-ccb91680-160d-11eb-91cc-1f05ebdfa7d0.png)
![deno_lsof-2020-10-24](https://user-images.githubusercontent.com/54777088/97083016-d04c9d80-160d-11eb-96f7-3bb990d67959.png)

I don't use an nginx proxy server; just deno directly on a linux distro.
I execute deno with `sudo` and I also tried to start deno with [pm2](https://pm2.keymetrics.io/). In all cases it stops working after a couple of hours.

The server runs on AWS Lightsail and I can rule out that a huge amount of requests is causing this (maybe 10 requests per hour). 

I can also say with certainty that this does not happen with a deno `http` server nor with a node `https` server, which I both tested at least once for two days. 

There was no big CPU or memory load which I could see in the monitoring tools. I only respond with `hello world` to requests at the moment.
If necessary I would be happy to give ssh access to somebody who knows how to debug this issue. 

Thanks a lot!

lucacasonato:
I have spun up a system on GCP (e2-medium, 2 vCPUs, 4 GB memory) with this code:

```js
import { listenAndServeTLS } from "https://deno.land/std@0.74.0/http/server.ts";

const body = "Hello HTTPS";
const options = {
  port: 443,
  certFile: "/etc/letsencrypt/live/denohttpstest.lcas.dev/fullchain.pem",
  keyFile: "/etc/letsencrypt/live/denohttpstest.lcas.dev/privkey.pem",
};
listenAndServeTLS(options, (req) => {
  req.respond({ body });
});
```

I have also set up uptime checks to see if it goes down. Let's see if I can reproduce this.

timonson:
I am using a 512 MB RAM, 1 vCPU, 20 GB SSD machine. But I could not see high resources usage. Otherwise just dm me and I send you my ssh key.

lucacasonato:
[...]

PR #8365, opened 2020-11-13
fix(std/http): fix error handling in the request iterator

### LINKED ISSUE #8364: std/http: Doesn't handle error when finalizing request
A simple HTTP server `http.ts`:
```ts
import { serve, ServerRequest, Response, Status } from "https://deno.land/std@0.77.0/http/mod.ts";

const server = serve("127.0.0.1:8000");

    let response: Response;
    try {
        response = await handleRequest(req);
    } catch (error) {
        console.error("handle error:", error);
        response = { status: Status.InternalServerError, body: "Server error" };
    }
    try {
        await req.respond(response);
    } catch (error) {
        console.error("respond error:", error);
    }
}

async function handleRequest(req: ServerRequest): Promise<Response> {
    if (req.url == "/hello") {
        return { body: "Hello!" };
    } else if (req.method == "PUT" && req.url == "/data") {
        // Start reading data
        const buf = new Uint8Array(4096);
        let haveRead = 0;
        while (true) {
            const nread = await req.body.read(buf);
            if (!nread) break;
            haveRead += nread;
            if (haveRead >= 10240) {
                // simulate an error when we have read 10k bytes
                throw new Error("Invalid data");
            }
        }
        console.info(`Received data: ${haveRead} bytes`)
        return { body: "OK" };
    } else {
        return { status: Status.NotFound };
    }
}
```

After starting the server by running  `deno run --allow-net http.ts`, use curl to do some requests:

```shell
$ curl http://127.0.0.1:8000/hello
Hello!
$ echo hello | curl -T- http://127.0.0.1:8000/data # PUT "hello\n"
OK
$ cat /dev/urandom | curl -T- http://127.0.0.1:8000/data  # PUT random data
Server error
```

The server will exit with uncaught error:
```
Received data: 6 bytes
handle error: Error: Invalid data
    at handleRequest (http.ts:33:23)
    at async http.ts:8:20
error: Uncaught (in promise) UnexpectedEof
    if (line === null) throw new Deno.errors.UnexpectedEof();
                             ^
    at Object.read (_io.ts:67:30)
    at async ServerRequest.finalize (server.ts:109:13)
    at async Server.iterateHttpRequests (server.ts:179:7)
    at async Server.acceptConnAndIterateHttpRequests (server.ts:227:5)
    at async MuxAsyncIterator.callIteratorNext (mux_async_iterator.ts:30:31)
```

---

A similar oak server have the same problem (`oak.ts`):
```ts
import { Application, Router } from "https://deno.land/x/oak@v6.3.2/mod.ts";

const app = new Application();
const router = new Router();

router
  .get("/hello", (ctx) => {
    ctx.response.body = "hello!";
  })
  .put("/data", async (ctx) => {
    const reader = await ctx.request.body({ type: "reader" }).value;

    // Start reading data
    const buf = new Uint8Array(4096);
    let haveRead = 0;
    while (true) {
      const nread = await reader.read(buf);
      if (!nread) break;
      haveRead += nread;
      if (haveRead >= 10240) {
        // simulate an error when we have read 10k bytes
        throw new Error("Invalid data");
      }
    }

    ctx.response.body = "OK";
  });

app.use(router.routes());

app.addEventListener("error", (evt) => console.error(evt.error));
await app.listen({ port: 8000 });
```

When request using `cat /dev/urandom | curl -T- http://127.0.0.1:8000/data`:
```
Error: Invalid data
    at oak.ts:22:15
    at async dispatch (middleware.ts:41:7)
    at async dispatch (middleware.ts:41:7)
    at async dispatch (middleware.ts:41:7)
    at async Application.#handleRequest (application.ts:252:9)
UnexpectedEof
    at Object.read (_io.ts:67:30)
    at async ServerRequest.finalize (server.ts:109:13)
    at async Server.iterateHttpRequests (server.ts:179:7)
    at async Server.acceptConnAndIterateHttpRequests (server.ts:227:5)
    at async MuxAsyncIterator.callIteratorNext (mux_async_iterator.ts:30:31)
```
The server process doesn't exit, but it seems that the server stop accepting requests
(`curl http://127.0.0.1:8000/hello` will stuck).

lideming:
Fixes #8364

If the request body is using chunked encoding, errors may be thrown in `request.finalize()`.

In this case, we should untrack and close the connection.


bartlomieju:
Thanks @lideming! Can we get some kind of test case? I'm also a bit concerned that `catch` clause is a catch-all, we should probably check for exact error type

lideming:
@bartlomieju 
Tests are added.

The error types can be `InvalidData`, `UnexpectedEof`, `PartialReadError` and more... The body reader may be complex. and its thrown types are uncertain. We should not assume the reader used here is always the chunked encoding reader. So I think this catch-all is okay.

Also, there is already another catch-all in the same function:
https://github.com/denoland/deno/blob/a59f5eadd86e85bf662b37ec40ceb6f80cb207c4/std/http/server.ts#L148
(It checks for several error types to send the `400 Bad Request` and ignores other errors)
(I just discovered... When it sends `400 Bad Request`, it seems to not handle errors thrown by `writeResponse`! Should I create another PR for this one?)

bartlomieju:
@nayeemrmn @MarkTiedemann what do you think?

nayeemrmn:
Are you sure we shouldn't catch the error, untrack, close, and then rethrow the error?

lideming:
@nayeemrmn I am sure.

When we get an error there, it mostly means it is an invalid request or the connection has closed by the client.

If we rethrow the error, which will be caught and rethrowed by the `MuxAsyncIterator`. The request iterator caller will finally get it, which is not expected to me:
```ts
for await (const request of server) {  // The error will be thrown from here when
  handler(request);                    // we get an error in reading requests
}
```

nayeemrmn (approved):
LGTM

lideming:
Also, I have discovered another similar problem (which was mentioned above):

https://github.com/denoland/deno/blob/a59f5eadd86e85bf662b37ec40ceb6f80cb207c4/std/http/server.ts#L154

It tries to send the `400` response when it failed to read requests, but doesn't handle the error could be thrown within `writeResponse`.

There will be an error thrown within `writeResponse` if a client:
1) Make a TCP connection to the server,
2) Send an invalid request, and
3) Close the TCP connection immediately

The server will see it's a invalid request and try to send a `400` response, but encounter an error when write to the connection, since the connection has been closed. (I will make a test later to verify this.)

Should I create a new issue and a new PR for this new problem, or just push the fix here and change the title of this PR?

lideming:
I have finished the test for the new problem... It's harder than I thought.

```ts
// test400.ts
import { serve } from "https://deno.land/std@0.77.0/http/mod.ts";
import { encode } from "https://deno.land/std@0.77.0/encoding/utf8.ts";

async function startServer() {
  const server = serve("127.0.0.1:8000");
    req.respond({ body: "Hello!" });
  }
}

startServer();

async function breakServer() {
  var conn = await Deno.connect({
    hostname: '127.0.0.1',
    port: 8000
  });

  await Deno.writeAll(
    conn,
    encode(
      // A normal request is required:
      "GET / HTTP/1.1\r\n" +
      "Host: localhost\r\n" +
      "\r\n" +

      // The bad request:
      "GET / HTTP/1.1\r\n" +
      "Host: localhost\r\n" +
      "INVALID!HEADER!\r\n" +
      "\r\n"
    )
  );
  // After sending the two requests, don't receive the reponses.

  // Closing the connection now.
  conn.close();

  // When the client called `close()`, it sends TCP FIN to the server.
  // If the server also called `close()`, there will be a graceful close.
  // But if the server continued to send data, the client will send TCP RST
  // to reset the connection, since it doesn't expect more incomming data.

  // Now, the `std/http` server is sending responses to a closing client,
  // the first few `write()` calls will not throws, until it received the TCP
  // RST. So we need the normal request to make the server do a few writes
  // before it writes that `400` response.

  // I have tested it on my Arch Linux.
  // Things may differ between Operating Systems. For example, Windows have
  // some "optimizations" for its loopback connections. So it may not work on 
  // other OSes.

  console.log('done');
}

await breakServer();

```
```shell
deno run --allow-net test400.ts
```
Will get the error:
```
error: Uncaught (in promise) BrokenPipe: Broken pipe (os error 32) 
    at unwrapResponse (deno:cli/rt/10_dispatch_minimal.js:59:13)
    at sendAsync (deno:cli/rt/10_dispatch_minimal.js:98:12)
    at async write (deno:cli/rt/12_io.js:117:20)
    at async Object.writeAll (deno:cli/rt/13_buffer.js:223:19)
    at async BufWriter.flush (bufio.ts:468:7)
    at async writeResponse (_io.ts:288:3)
    at async Server.iterateHttpRequests (server.ts:154:11)
    at async Server.acceptConnAndIterateHttpRequests (server.ts:227:5)
    at async MuxAsyncIterator.callIteratorNext (mux_async_iterator.ts:30:31)
```

bartlomieju:
@lideming is `BrokenPipe` error fixed in this PR as well?

lideming:
@bartlomieju Not yet. Should I push more commits to fix that?

bartlomieju:
@lideming it seems you already got test cases, so if you could add more commit to this PR that would be great

bartlomieju:
@lideming sorry to bother again, but could you look into `ConnectionReset` as well? https://github.com/denoland/deno/issues/8107#issuecomment-729078414

lideming:
@bartlomieju the fix is added. But I am not sure if we have covered all the error types.
I think the behavior of `Deno.Listener.accept()` is weird. It should not throw these errors when it failed to finish the TLS handshake. Instead, it should start accepting the next connection or throw an error type `TlsError`, which wraps the underlying error info.

bartlomieju:
> @bartlomieju the fix is added. But I am not sure if we have covered all the error types.
> I think the behavior of `Deno.Listener.accept()` is weird. It should not throw these errors when it failed to finish the TLS handshake. Instead, it should start accepting the next connection or throw an error type `TlsError`, which wraps the underlying error info.

@lideming that sounds like a bigger rework, we should discuss it as part of #8405

bartlomieju (approved):
LGTM, thank you @lideming, hopefully this PR will squash a lot of bugs in HTTP server implementation.

rvdende:
awesome, thank you for this @lideming !

Issue #6309, opened 2020-06-16
docs: Describe how to configure shell completions


muuvmuuv:
I would love to use the deno auto completions but cannot get them working with [prezto](https://github.com/sorin-ionescu/prezto). 

```shell
deno completions zsh > /Users/marvinheilemann/.zprezto/modules/deno/init.zsh
```

It says `_deno:12: command not found: _arguments`. I wonder if I need additional steps but I couldn't find any. 

Has anyone setup completions with prezto already or would it be possible to make it prezto compatible?

bartlomieju:
@muuvmuuv these completions are automatically generated by the CLI parser we're using - which is [`clap`](https://github.com/clap-rs/clap). It might be good to ask question there.

intgr:
The issue is that zsh is very picky about how it discovers those completion files. You have to place it with file name `_deno` into some directory in the `$fpath` variable, usually `/usr/local/share/zsh/site-functions`. Unfortunately there are no user-local paths by default.

As a one-liner:
```sh
deno completions zsh |sudo tee /usr/local/share/zsh/site-functions/_deno
```
After that you can run `compinit` in your zsh shell and `deno <TAB>` should automagically work.

bartlomieju:
@intgr thanks for explanation, we should add that to documentation.

Issue #5777, opened 2020-05-23
completions help text for bash won't work on ubuntu


opensas:
The `deno completions` help text gives the following message:

```shell
$ deno completions --help
deno-completions 
Output shell completion script to standard output.
  deno completions bash > /usr/local/etc/bash_completion.d/deno.bash
  source /usr/local/etc/bash_completion.d/deno.bash

USAGE:
    deno completions [OPTIONS] <shell>
```

But in ubuntu it won't work

```
$ deno completions bash > /usr/local/etc/bash_completion.d/deno.bash
bash: /usr/local/etc/bash_completion.d/deno.bash: No such file or directory
```

What did work for me was:

```
$ deno completions bash | sudo tee /etc/bash_completion.d/deno.bash > /dev/null
$ . /etc/bash_completion
```

ry:
Thanks for the report. Feel free to update the example in the help text.

opensas:
Thanks for the reply, I'm not sure how it's done on other distros (I'll see if I can try with one of them), perhaps the best would be to state in the short cli message that the solution provided is just an example and that you should check how it's done on your own distro, even better point to some kind of wiki or documentation telling how to achieve that in each distro
Do we have some wiki or some place in the docs where we could put that info?
I guess the best place would be here: https://deno.land/manual/getting_started/setup_your_environment#shell-autocomplete

David-Else:
I can confirm that the example in `--help` and the manual does NOT work on Centos/RHEL (and therefore I am 99% sure Fedora too). What DOES work is:

```shell
deno completions bash | sudo tee /etc/bash_completion.d/deno.bash > /dev/null
source /etc/bash_completion.d/deno.bash
```
@opensas does `source /etc/bash_completion.d/deno.bash` work OK on your computer? If it does then I think we can call the above confirmed working on Ubuntu and Redhat.

Which distro does the original `--help` version work for? Maybe the above should be the official way of doing it, Ubuntu is the most popular distro.

I can make a pull request on this, but do we want to replace the recommended way or have two ways of doing it? The above is for Ubuntu and Redhat, but which distro is the original one intended for?

opensas:
I think the help message could tell you how to achieve it on debian based distros, and tell you to check how to do it on your own distro. Moreover on the wiki we could add instructions for debia + fedora + arch based distros, which I guess will cover more than 95% of cases.

JerryGreen:
[...]

PR #8369, opened 2020-11-13
fix(cli): Fix documentation about usage of deno completions script

iammack:
Documentation for usage of deno completions script provided by deno completions --help is outdated. The correct method has been borrowed from the issues:
Closes https://github.com/denoland/deno/issues/5777
Closes https://github.com/denoland/deno/issues/6309


bartlomieju (approved):
LGTM, thanks @iammack

cli/flags.rs
```diff
@@ -811,8 +811,22 @@ fn completions_subcommand<'a, 'b>() -> App<'a, 'b> {
     .about("Generate shell completions")
     .long_about(
       "Output shell completion script to standard output.
-  deno completions bash > /usr/local/etc/bash_completion.d/deno.bash
-  source /usr/local/etc/bash_completion.d/deno.bash",
+
+bash:
+  deno completions bash | sudo tee /etc/bash_completion.d/deno.bash > /dev/null
```
ry, line 816: Why is this preferable to 
```
deno completions bash > /usr/local/etc/bash_completion.d/deno.bash
```
It's significantly more complicated.

nayeemrmn: @ry I think it's the `sudo`. The pipe thing only works if the current shell is invoked with `sudo`, since the shell is doing the writing to the protected file. This lets you put it before `tee`.

ry: I don't think it is in the scope of deno to tell users when to use sudo. If you're root this is not a problem.

The shell completion help text is not the place to explain unix file system permissions.

JerryGreen: The doc wasn't outdated, - on macOS it still works perfectly. It seems it's permission problem on Linux and Windows though.

I've played with deno on windows recently, using bash (Git Bash), and it doesn't have `sudo` at all. I come up with placing bash completions into `~/.completions` and automatically sourcing it on each terminal session with this little code:

https://github.com/denoland/deno/issues/5777#issuecomment-726268305

Not very standard, I haven't seen anyone using this path, but at least it has no permission problems. Whatever path it would be, I think completions should be installed automatically right on installation step (since you have it anyway), so completions would work out of the box!

soundjuice:
`deno completions bash > ~/.bash_completion`

Suggestion from 

[bash-completion faq](https://github.com/scop/bash-completion#faq)
> Q. Where should I install my own local completions?
A. Put them in the completions subdir of $BASH_COMPLETION_USER_DIR (defaults to $XDG_DATA_HOME/bash-completion or ~/.local/share/bash-completion if $XDG_DATA_HOME is not set) to have them loaded automatically on demand when the respective command is being completed. See also the next question's answer for considerations for these files' names, they apply here as well. Alternatively, you can write them directly in ~/.bash_completion which is loaded eagerly by our main script.

PR #8372, opened 2020-11-13
fix(op_crates/web) fix abort signal length

benjamingr:
Another small web compatibility fix I ran into when checking the node tests with Deno's signal :]


benjamingr:
I'll merge this with https://github.com/denoland/deno/pull/8373 per @bartlomieju 's suggestion

PR #8373, opened 2020-11-13
fix(op_crates/web) fix window constructor length

benjamingr:
Another small PR for web compatibility.

### DIFF
--- a/op_crates/web/03_global_interfaces.js
+++ b/op_crates/web/03_global_interfaces.js
@@ -4,7 +4,7 @@
   const illegalConstructorKey = Symbol("illegalConstuctorKey");
 
   class Window extends EventTarget {
-    constructor(key) {
+    constructor(key = null) {
       if (key !== illegalConstructorKey) {
         throw new TypeError("Illegal constructor.");
       }
--- a/op_crates/web/03_global_interfaces.js
+++ b/op_crates/web/03_global_interfaces.js
@@ -17,7 +17,7 @@
   }
 
   class WorkerGlobalScope extends EventTarget {
-    constructor(key) {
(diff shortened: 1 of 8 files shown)


op_crates/web/03_global_interfaces.js
```diff
@@ -4,7 +4,7 @@
   const illegalConstructorKey = Symbol("illegalConstuctorKey");
 
   class Window extends EventTarget {
-    constructor(key) {
+    constructor(key = null) {
       if (key !== illegalConstructorKey) {
         throw new TypeError("Illegal constructor.");
       }
```
bartlomieju, line 10: We got a couple more such constructors, @benjamingr do you want to check all of them in a single PR?

benjamingr: Yes if a single PR is easier to review I'll push all the changes in a single PR :]

benjamingr: I've put all the changes here (in this PR) and verified in the spec that this fix is indeed correct (and required) for it.

bartlomieju (approved):
I've confirmed that this is inline with the browser. LGTM, thanks @benjamingr!

benjamingr:
@bartlomieju for future reference to my understanding this is defined by the  [WebIDL spec](https://heycam.github.io/webidl/#ecmascript-binding) under "3. ECMAScript binding" :]

For example for AbortSignal - it is defined in the [DOM spec here](https://dom.spec.whatwg.org/#interface-AbortSignal), its behavior as a WebIDL interface [is defined here](https://heycam.github.io/webidl/#interface-object) and specifically step 10:

> Let length be 0.

PR #8375, opened 2020-11-13
fix(fmt, lint): Make sure that target paths are not directory

### LINKED ISSUE #8374: regression: deno fmt directory matching fails
Running deno fmt in this repo: https://github.com/lucacasonato/dext.ts, causes an error since some time after 1.5.2. Possibly related to #8040. I have not been able to narrow it down yet.

```
> deno fmt
error: Is a directory (os error 21)
```

magurotuna:
Fixes #8374 

~~I'll fix more and add tests, so WIP for now.~~ Done

### DIFF
--- a/cli/fs.rs
+++ b/cli/fs.rs
@@ -72,9 +72,67 @@ pub fn resolve_from_cwd(path: &Path) -> Result<PathBuf, AnyError> {
   Ok(normalize_path(&resolved_path))
 }
 
+/// Checks if the path has extension Deno supports.
+pub fn is_supported_ext(path: &Path) -> bool {
+  let lowercase_ext = path
+    .extension()
+    .and_then(|e| e.to_str())
+    .map(|e| e.to_lowercase());
+  if let Some(ext) = lowercase_ext {
+    ext == "ts" || ext == "tsx" || ext == "js" || ext == "jsx" || ext == "mjs"
+  } else {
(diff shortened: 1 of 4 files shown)


bartlomieju:
@magurotuna huge thanks for looking into this! I noticed that in `fs.rs` there is `files_in_subtree` method used in a few places. Do you think we could combine the two implementations into a single one?

magurotuna:
@bartlomieju Oh, I missed `files_in_subtree`, thanks for pointing out. It seems usable in this fix. I'll try it tomorrow :)

magurotuna:
@bartlomieju PTAL😉

cli/fs.rs
```diff
@@ -118,18 +176,83 @@ mod tests {
...
+    // ├── child
+    // │   ├── e.mjs
+    // │   ├── f.mjsx
+    // │   ├── .foo.TS
+    // │   └── README.md
+    // ├── c.tsx
+    // ├── d.jsx
+    // └── ignore
+    //     ├── g.d.ts
+    //     └── .gitignore
```
bartlomieju, line 219: 👍

bartlomieju (approved):
I verified that problem described in https://github.com/denoland/deno/issues/8374 is no longer present. Great fix and cleanup @magurotuna, thanks!

PR #6760, opened 2020-07-15
fix: providing empty source code for missing compiled files

bartlomieju:
Fixes https://github.com/denoland/deno/issues/6746

PR #8377, opened 2020-11-13
fix(cli/program_state): Remove redundant compiler warning

nayeemrmn:
As I see it, #6760 introduced a valid fix for a strictly internal quirk where type-only modules used to cause problems. But the accompanying warning isn't necessary because it affects valid user code.

### DIFF
--- a/cli/program_state.rs
+++ b/cli/program_state.rs
@@ -207,12 +207,6 @@ impl ProgramState {
         && maybe_referrer.is_none()
         && specifier.as_url().scheme() == "file")
     {
-      let message = if let Some(referrer) = maybe_referrer {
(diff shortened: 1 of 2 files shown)


kitsonk:
I totally disagree with the intent of this PR. The message is not redundant. People should not be importing type only `d.ts` files via `import` and should instead use `import type`, as the resulting code is not `--no-check` safe, and can/will result in other errors. If people are trying to import ambient types, they should be using `import` at all but instead using triple slash directives.

nayeemrmn:
😓 I could have sworn I saw this warning without using any `d.ts`  files, but I can't reproduce anything now. Closing.

kitsonk:
@nayeemrmn there were some bugs related to #8311 that was causing this to occur incorrectly, so you or other could have incorrectly seen this (caused when another file had a diagnostic error, which was then fixed, but a dependency was not changed)

PR #8378, opened 2020-11-14
feat(std/io): ReadableStream from AsyncIterator & WritableStream from Writer

crowlKats:
we should also rename `fromStreamWriter` & `fromStreamReader` to `writerFromStreamWriter` & `readerFromStreamReader`.
Going to add tests later

### DIFF
--- a/std/io/streams.ts
+++ b/std/io/streams.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
-/** Create a `Writer` from a `WritablseStreamDefaultReader`. */
-export function fromStreamWriter(
+/** Create a `Writer` from a `WritableStreamDefaultReader`. */
+export function writerFromStreamWriter(
   streamWriter: WritableStreamDefaultWriter<Uint8Array>,
 ): Deno.Writer {
   return {
@@ -13,8 +13,8 @@ export function fromStreamWriter(
   };
 }
 
-/** Create a `Reader` from a `ReadableSteramDefaultReader`. */
-export function fromStreamReader(
+/** Create a `Reader` from a `ReadableStreamDefaultReader`. */
+export function readerFromStreamReader(
   streamReader: ReadableStreamDefaultReader<Uint8Array>,
 ): Deno.Reader {
   const buffer = new Deno.Buffer();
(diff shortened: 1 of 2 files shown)


std/io/streams.ts
```diff
@@ -34,3 +34,29 @@ export function fromStreamReader(
     },
   };
 }
+
+export function ReadableStreamFromAsyncIterator(
```
bartlomieju, line 38: Name should start with lowercase

std/io/streams.ts
```diff
@@ -34,3 +34,29 @@ export function fromStreamReader(
...
+      if (done) {
+        controller.close();
+      } else {
+        controller.enqueue(value);
+      }
+    },
+  });
+}
+
+export function WritableStreamFromWriter(
```
bartlomieju, line 54: Ditto

std/io/streams.ts
```diff
@@ -34,3 +34,29 @@ export function fromStreamReader(
...
+  writer: Deno.Writer,
+): WritableStream<Uint8Array> {
+  return new WritableStream({
+    async write(chunk) {
+      await writer.write(chunk);
```
bartlomieju, line 59: I'm pretty sure it's not guaranteed that `writer.write()` will write full chunk in one go - probably should be `Deno.writeAll(writer, chunk)`

bartlomieju (changes requested):

bartlomieju:
@crowlKats can you add some test cases?

std/io/streams.ts
```diff
@@ -34,3 +34,31 @@ export function fromStreamReader(
...
+    async write(chunk) {
+      await Deno.writeAll(writer, chunk);
+    },
+  });
+}
+
+/** Create a `ReadableStream` from an `AsyncIterator`. */
+export function readableStreamFromAsyncIterator<T>(
+  iterator: AsyncIterableIterator<T>,
+): ReadableStream<T> {
```
nayeemrmn, line 52: IMO `readableStreamFromReader()` should exist either way. This module exists to provide bridges between Deno IO and whatwg streams. One should be able to "expect" a `readableStreamFromReader` looking at the other functions.

In this case, there's no problem with having both even if one would have a trivial implementation. Similar to `readLines()` in `std/io/bufio.ts`.

ry:
I'm missing some context on this PR... what is the purpose of these renames?

crowlKats:
This is not just renames, but rather 2 new functions: one allows to create a ReadableStream from an AsyncIterator and the other allows to create a WritableStream from a Deno.Writer.
The renames are there just because it would make differentiating the functions easier; `fromStreamWriter` is too vague, but `writerFromStreamWriter` gives enough context on what it does.

ry (approved):
LGTM - thanks @crowlKats

std/io/streams.ts
crowlKats: @nayeemrmn i will look into it soon; the best usage for that would a bytes ReadableStream, but still not knowledgeable about those

Issue #7953, opened 2020-10-13
REPL panics related to invalid unicode


solson:
I found a way to trip two different Rust panic messages in the REPL:

```console
$ deno --version
deno 1.4.6
v8 8.7.220.3
typescript 4.0.3

$ deno
Deno 1.4.6
exit using ctrl+d or close()
> '\u{1f3b5}'
"🎵"
> '\u{1f3b5}'[0]
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unexpected end of hex escape", line: 1, column: 60)', cli/inspector.rs:859:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
fish: “deno” terminated by signal SIGABRT (Abort)

$ deno
Deno 1.4.6
exit using ctrl+d or close()
> '\u{1f3b5}'[1]
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("lone leading surrogate in hex escape", line: 1, column: 59)', cli/inspector.rs:859:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
fish: “deno” terminated by signal SIGABRT (Abort)
```

Note that these panics *only* occur in the REPL, and regular Deno code is unaffected as far as I can tell.

cc @caspervonb

caspervonb:
Thanks for the heads up 👍 

For reference, It's from a StringBuffer unwrap in the inspector so it only happens for local inspector messages that contains invalid unicode.

https://github.com/denoland/deno/blob/1956cb81372b96bc476e74ab43a62a6e60861277/cli/inspector.rs#L854-L868

benjamingr:
Hey, I want to try and take a look at this as it's a relatively small fix that will make it learn some rust :]

benjamingr:
Checked this for a bit, I think this actually isn't actually a small fix at all to do _well_. The "invalid" JSON is coming from the CDP itself, Chrome's JSON parser and other parsers I've found happily parse this and it seems like an issue with serde_json itself.

I will however take the opportunity to brush up on rust error handling and match expressions and try to solve the issue ad-hoc :]

bartlomieju:
Fixed in #8759

PR #8379, opened 2020-11-14
fix(repl) recover from invalid unicode

benjamingr:
Fixes https://github.com/denoland/deno/issues/7953

This is not a very good fix (yet). The crux is that serde_json behaves differently from V8's JSON parser.

A better/more generic fix would be to "upstream" a fix in serde_json to allow recovery from this. In the meantime, I returned a result with the string "Unterminated string literal".

Mostly - doing this to practice rust and deno :]

### DIFF
--- a/cli/inspector.rs
+++ b/cli/inspector.rs
@@ -856,8 +856,30 @@ impl v8::inspector::ChannelImpl for InspectorSession {
     call_id: i32,
     message: v8::UniquePtr<v8::inspector::StringBuffer>,
   ) {
+    
     let raw_message = message.unwrap().string().to_string();
-    let message = serde_json::from_str(&raw_message).unwrap();
+    let message: serde_json::Value = match serde_json::from_str(&raw_message) {
+      Ok(v) => v,
+      Err(error) => match error.classify() {
+        serde_json::error::Category::Syntax => json!({
+          "id": call_id,
+          "result": {
+            "result": {
+              "type": "error"
+              "description": "Unterminated string literal",
+              "value": "undefined"
+            },
+            "exceptionDetails": {
+              "exceptionId": 0,
+              "text": "Unterminated string literal",
+              "lineNumber": 0,
+              "columnNumber": 0 
+            }
+          }
+        }),
+        _other_error => panic!("Could not parse inspector messagge")
+      }
+    };
 
     self
       .response_tx_map


bartlomieju:
@caspervonb please take a look

caspervonb:
Please add a test case with the bad inputs solson came up with.

benjamingr:
@caspervonb I am happy to add a test - but given a test do the changes themselves and the approach (fixing this here until a fix is upstreamed to the json parsing code itself) look fine to you?

caspervonb:
SGTM

bnoordhuis:
Benjamin, just checking: is the issue here invalid Unicode or malformed JSON? Can you share an example message?

benjamingr:
> Benjamin, just checking: is the issue here invalid Unicode or malformed JSON? Can you share an example message?

The issue here is that when you pass an unterminated unicode code unit to `serde_json` it will error but chrome's JSON parser happily processes it.

So when you enter something like `'\u{1f3b5}'[0]` in the Deno repl it will encode it in a websocket message and when Deno tries to deserialize it (in inspector.rs) the `.unwrap` call panics because serde_json does not deserialize it.

I am not sure if chrome or serde_json is right here in terms of the spec - this fix is just a patch that says "if you would have panicked - show an 'Unterminated string literal' error instead".

cli/inspector.rs
```diff
@@ -856,8 +856,30 @@ impl v8::inspector::ChannelImpl for InspectorSession {
...
     let raw_message = message.unwrap().string().to_string();
-    let message = serde_json::from_str(&raw_message).unwrap();
+    let message: serde_json::Value = match serde_json::from_str(&raw_message) {
+      Ok(v) => v,
+      Err(error) => match error.classify() {
+        serde_json::error::Category::Syntax => json!({
+          "id": call_id,
+          "result": {
+            "result": {
+              "type": "error"
```
bartlomieju, line 868: missing comma here?

benjamingr: No this is JSON - can't put a comma here :]

benjamingr: I have no idea how this works the comma isn't optional this is JSON 😅 

good catch

bnoordhuis:
> when you enter something like '\u{1f3b5}'[0] in the Deno repl it will encode it in a websocket message and when Deno tries to deserialize it (in inspector.rs) the .unwrap call panics because serde_json does not deserialize it.

Thanks, I'm able to reproduce using your example.

The inspector sends `{"id":3,"result":{"result":{"type":"string","value":"\ud83c"}}}` and `serde_json::from_str()` chokes on the unpaired surrogate (`U+D800 < U+D83C < U+DFFF`).

From `serde_json-1.0.59/src/de.rs`:
```rust
    /// The behavior of serde_json is specified to fail on non-UTF-8 strings
    /// when deserializing into Rust UTF-8 string types such as String, and
    /// succeed with non-UTF-8 bytes when deserializing using this method.
```
(Where "this method" is a helper to `serde_json::from_slice()`.)

Ergo, operating on `&[u8]` instead of `&str` should fix it but that means converting from `&[u16]` to `&[u8]` first and that's its own can of worms.

I've opened https://bugs.chromium.org/p/v8/issues/detail?id=11193 to discuss if it's better to handle this in V8. Deno is probably not the only embedder that has to grapple with this so I think it makes sense to fix this at the root.

benjamingr:
@bnoordhuis yes though to be fair this _should_ probably be fixed in serde_json - Chrome's JSON parser happily deserializes the invalid JSON gracefully.

I'm currently bedridden recovering from a surprising sickness last weekend - I'll pick up this work in a few days when I feel better (hopefully) - if this is urgent for whatever reason feel free to pick this up.

bnoordhuis:
🤷 - I don't think serde_json is wrong to reject it, else it would have to either emit invalid UTF-8 or transmogrify it into something like WTF-8.

Anyway, no rush. Get well soon! I can snail mail you some powdered chicken soup if it helps. :-)

bartlomieju:
Fixed in #8759

PR #8381, opened 2020-11-14
[WIP] build: add cargo feature to drop various tools

bartlomieju:
This is really a prototype and is not intended to be landed.

This commit adds default cargo feature called "tools". When building with `--no-default-features` flag most of the tools included in the regular binary won't be included (TS support, doc generator, formatter, linter, bundler, file watcher) and produced binary will be able to load file only from local disk (so effectively no URL support and only "deno run" subcommand is operational). This reduces binary size from about 45Mb to 30Mb (on macOS).

There are still about 50 "unused" warnings left.

To try out locally:
```
cd cli  // this is required - cannot build workspace with features
cargo build --no-default-features
```


bartlomieju:
Update:

When building with `--no-default-features` results as follows:
- release binary 27Mb; zipped 14Mb
- release binary without debug info 16Mb; zipped 6.8Mb

cli/ast.rs
```diff
@@ -135,8 +135,8 @@ impl DiagnosticBuffer {
 #[derive(Debug, Clone)]
 pub struct ErrorBuffer(Arc<RwLock<Vec<Diagnostic>>>);
 
-impl ErrorBuffer {
-  pub fn new() -> Self {
```
ry, line 139: aside: shouldn't this result in a clippy error currently?

bartlomieju: I would expect so, but it seems that's only the case when having `lib.rs`

bartlomieju:
Closing in favor of #8640

Issue #8343, opened 2020-11-11
Allow TTY to be set to a mode that generates signals


marcushultman:
`Deno.setRaw` sets relevant termios flags into "raw" mode. It would be beneficial to be able to set it to "cbreak" mode; a mode where raw characters are sent without being buffered, but when characters INTR, QUIT, SUSP, or DSUSP are received, we generate the corresponding signal instead. This can be done with the [ISIG flag](https://docs.rs/nix/0.19.0/nix/sys/termios/struct.LocalFlags.html#associatedconstant.ISIG).

Deno unsets the ISIG flag unconditionally when entering raw mode:
https://github.com/denoland/deno/blob/c7804c06adb44127b93ef45d9fc9eb0ef255fc3f/cli/ops/tty.rs#L185-L188

The desired behavior can be produced by not unsetting that flag. 

One approach is to add a `bool` to `SetRawArgs`?
https://github.com/denoland/deno/blob/c7804c06adb44127b93ef45d9fc9eb0ef255fc3f/cli/ops/tty.rs#L57-L60

Whatever solution is deemed ok, I'd be up for making a PR. Having this as an unstable API is ok.

marcushultman:
I believe this may already be the case on windows as it sets `ENABLE_PROCESSED_INPUT`:
https://github.com/denoland/deno/blob/c7804c06adb44127b93ef45d9fc9eb0ef255fc3f/cli/ops/tty.rs#L30-L32
Documented here:
https://docs.microsoft.com/en-us/windows/console/setconsolemode

Issue #4262, opened 2020-03-05
test_raw_tty hangs


keroxp:
<img width="701" alt="スクリーンショット 2020-03-05 21 21 49" src="https://user-images.githubusercontent.com/927286/75981218-6456e100-5f27-11ea-8605-aeb5acda4e38.png">

**Env:**
- macOS 10.15
- iTerm2, Terminal
- zsh

bartlomieju:
I'm experiencing the same issue

CC @kevinkassimo

kevinkassimo:
I suspect there is some race condition with that test. I'll see if I can get a fix of it by next week. Is this a zsh only problem, or also occurs under bash?

keroxp:
@kevinkassimo  Yes, always. `cargo test test_raw_tty` works for both.

bartlomieju:
Happened in CI as well:
https://github.com/denoland/deno/runs/493335175

ry:
test_raw_tty has been reenabled on mac, but continues to hang.

I can repeat it reliably in https://github.com/denoland/deno/pull/8520

PR #8383, opened 2020-11-14
feat(tty): add cbreak option to enable passthrough of signals

marcushultman:
Closes https://github.com/denoland/deno/issues/8343.


ry:
Hi @marcushultman - thanks for the patch - and sorry for the slow review.

I certainly see the need for exposing this API, but I'm a bit hesitant to invent new names like `generateSignal`. I wonder if you could do some research into how this API looks in other platforms (Node, Rust, Go, etc). Maybe we can decide on a name that others are already using.

Also we need a test somehow... I think it's fine to leave this unimplemented in Windows - but it should probably throw some NotImplemented exception if people try to use it there.

marcushultman:
Thanks @ry! Glad to hear that there may be a way to get this available.

As far as research goes,  [Node](https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode) states that "all special processing of characters by the terminal is disabled". Golang also has a [raw mode](https://godoc.org/golang.org/x/crypto/ssh/terminal#MakeRaw), but that unconditionally captures ctrl+c just like Node (this is not documented).

Deno just wraps [termios](https://man7.org/linux/man-pages/man3/termios.3.html), so it would be reasonable to align with naming from there. In termios, they call it "local modes" when the `c_lflag` bitfield is set. The `ISIG` flag doesn't really state a specific "mode":
> When any of the characters INTR, QUIT, SUSP, or DSUSP are re‐
              ceived, generate the corresponding signal.  
This is where I chose the `generateSignal` name from.

Perhaps `ISIG`, `isigMode`, `interruptSignal` or similar could be used?

As for testing, I could not easily think of a way to raise ^C in the current test suite(s). Not sure what can be done instead. Throwing NotImplemented on windows however should be doable!

ry:
Regarding the naming, I want to summon some experts for advice. @piscisaureus, @bnoordhuis What do you think about this `generatesSignal` name? Do you have a better suggestion?

> As for testing, I could not easily think of a way to raise ^C in the current test suite(s). Not sure what can be done instead. Throwing NotImplemented on windows however should be doable!

I think you'll want to create a custom test in cli/tests/integration_tests.rs and create a pty. You can probably spawn a Deno subprocess (using std::process::Command) and send the SIGINT signal using kill(2) I think...

Have a look at https://github.com/denoland/deno/blob/91f293442d5bfb28212b5b13b536ceae06427c99/cli/tests/integration_tests.rs#L174-L222

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -815,19 +815,28 @@ declare namespace Deno {
...
    * Set TTY to be under raw mode or not. In raw mode, characters are read and
    * returned as is, without being processed. All special processing of
    * characters by the terminal is disabled, including echoing input characters.
+   * Characters that correspond to a signal can optionally be enabled.
    * Reading from a TTY device in raw mode is faster than reading from a TTY
    * device in canonical mode.
    *
    * ```ts
-   * Deno.setRaw(myTTY.rid, true);
+   * Deno.setRaw(myTTY.rid, true, { generateSignal: true });
```
bnoordhuis, line 832: `generateSignal` only does something when `mode == true`, right? That should probably be made clearer.

bnoordhuis (commented):
Apropos the name, I'd not overthink it and just call it `cbreak`.

People might reasonably expect `generateSignals` to cover `SIGTTOU` but `ISIG` doesn't, that's the `TOSTOP` flag.

marcushultman:
> Apropos the name, I'd not overthink it and just call it `cbreak`.

That would work.

Indeed, this flag is only relevant when when going into raw mode (or cbreak mode actually). What is your preferred way to model this? Documentation only/throw exception on invalid input or make a new API endpoint all together? Maybe that question is for @ry

marcushultman:
@ry The existing test for `setRaw` (`test_raw_tty`) Is deactivated and broken in more ways than the linked issue suggest. I ran in to the same behavior when trying to write a test for this change. My strategy was to `await Deno.signals.interrupt()` in the typescript test, and `master.write_all(&[3])` from the test case side - this prints "^C" to the TTY, which in some way is expected, but considering that we wouldn't be able to have the test running because of https://github.com/denoland/deno/issues/4262, I didn't see a reason to include that test.

ry:
@marcushultman Maybe we can just skip the test on mac? Or are there deeper issues? Let me try to start another PR which re-enables that test except on mac.

ry:
@marcushultman I've landed a fix to test_raw_tty. I think it should be working on mac now too.

marcushultman:
@ry that's great! I'll try again to get a test passing for this

marcushultman:
@ry The test uses `pty::fork` where the child process spawns deno and shares stdio/stdout. When writing SIGINT to the master process, this is not propagated to the deno process (but it can be captured in the child process). I don't know how to properly trigger `Deno.signals.interrupt()` for a deno process launched with `Command::new`. Ideas are welcome!

ry (approved):
Thanks @marcushultman - I made a small change to your test but otherwise LGTM

PR #8387, opened 2020-11-15
fix(std): Migrating path.resolve to fs.resolvePath

### LINKED ISSUE #8305: STD's `copySync()` throws permission error because it requires CWD access.
Here's a repository containing an example to replicate the bug:
https://github.com/ChiKler/-Deno-error-Uncaught-PermissionDenied-read-access-to-CWD-

I believe that's because you use `copySync()` from `std`, which (without a good reason) calls `path.resolve()` which requires CWD access. It shouldn't call `path.resolve()`.

_Originally posted by @nayeemrmn in https://github.com/denoland/deno/pull/2875#issuecomment-723579312_

whosmayank:
Closes #8305 

path.resolve() uses Deno.cwd() to add cwd as first parameter so that the resolved path is absolute. However, apis under std/path shouldn't be using Deno.cwd() implicitly as that would require read access to cwd. 
A lot of code which depended on it has been modified to suit the change.


std/wasi/snapshot_preview1_test.ts
```diff
@@ -64,7 +64,7 @@ for (const pathname of tests) {
     ignore: ignore.includes(pathname),
     fn: async function () {
       const prelude = await Deno.readTextFile(
-        path.resolve(rootdir, pathname.replace(/\.wasm$/, ".json")),
+        path.resolve(Deno.cwd(), rootdir, pathname.replace(/\.wasm$/, ".json")),
```
ry, line 67: This probably isn't necessary - rootdir is already absolute, right?

ry (commented):
Thanks for the PR @iammack. I think this makes sense.

I'm curious if @bartlomieju or @lucacasonato have any concerns?

lucacasonato (commented):
I think this makes a lot of sense. We should consider holding it for 1.6 though. Quite a large change to behaviour - we might want to announce it in the blog post. On the other hand, it is only std, so could also be landed for 1.5.4.

std/wasi/snapshot_preview1_test.ts
iammack: Yes, those were included as a quick way to test path.resolve implementations. The same goes for test files which use 
```
const moduleDir = path.dirname(path.fromFileUrl(import.meta.url));
const testdataDir = path.resolve(Deno.cwd(), moduleDir, "testdata");
```
I've removed the unnecessary calls in the latest commit. Thanks for reviewing:)

iammack:
If path.resolve's interface changes to pass Deno.cwd() explicitly, shouldn't path.relative do the same? Currently path.relative uses Deno.cwd() implicitly. It should be a comparatively easy fix. I can include that in another PR if everyone agrees.

ry:
I'm going to mark this to be merged for 1.6 per Luca's suggestion.

> Also, If path.resolve's interface changes to pass Deno.cwd() explicitly, shouldn't path.relative do the same? Currently path.relative uses Deno.cwd() implicitly. It should be a comparatively easy fix. I can include that in another PR if everyone agrees.

I think that makes sense... but definitely let's do that in a follow up PR - this is already a big change.

std/path/posix.ts
```diff
@@ -24,26 +24,16 @@ export function resolve(...pathSegments: string[]): string {
   let resolvedPath = "";
   let resolvedAbsolute = false;
 
-  for (let i = pathSegments.length - 1; i >= -1 && !resolvedAbsolute; i--) {
-    let path: string;
-
-    if (i >= 0) path = pathSegments[i];
-    else {
-      if (globalThis.Deno == null) {
-        throw new TypeError("Resolved a relative path without a CWD.");
```
lucacasonato, line 33: I think these errors should stay. I think otherwise it is possible that we return a relative path from resolve.

It would be great if you added a test to check what happens when you only pass relative path segments to resolve.

```
    if (i >= 0) path = pathSegments[i];	
    else {	
        throw new TypeError("Resolved a relative path without a CWD.");	
    }	
```

std/path/win32.ts
```diff
@@ -30,35 +30,8 @@ export function resolve(...pathSegments: string[]): string {
...
-      path = pathSegments[i];
-    } else if (!resolvedDevice) {
-      if (globalThis.Deno == null) {
-        throw new TypeError("Resolved a drive-letter-less path without a CWD.");
-      }
-      path = Deno.cwd();
-    } else {
-      if (globalThis.Deno == null) {
-        throw new TypeError("Resolved a relative path without a CWD.");
-      }
```
lucacasonato, line 45: Ditto

std/path/posix.ts
iammack: Yeah, makes sense. I had a question though. Is there a deno API (rutime or std) that would convert relative path to absolute path other than path.resolve? If not, should we provide one as we are removing that part from path.resolve.

lucacasonato: cc @nayeemrmn

I think we should introduce `fs.resolvePath`, which does the same as current `path.resolve` (so fall back to Deno.cwd). I think join and resolve have different semantics. I dont think we can join them:
```js
path.join("./foo/", "/bar",) // "./foo/bar"
path.resolve("./foo/", "/bar",) // "/bar"
```

iammack: Yes, exactly. It will be a cascade of changes if we go ahead with changing ```path.resolve's``` implementation. The same can be done for path.relative by introducing a similar function under fs.

lucacasonato:
Putting this on hold until this is discussed in std working group.

whosmayank:
> Putting this on hold until this is discussed in std working group.
I'll not be able to attend the working group meet because it's at 2:30 am for me so here's the reasoning  for the three changes introduced:

https://github.com/denoland/deno/pull/8387: path.resolve() uses Deno.cwd() to add cwd as first parameter so that the resolved path is absolute. However, apis under std/path shouldn't be using Deno.cwd() implicitly as that would require read access to cwd. Hence the move to fs.resolvePath

https://github.com/denoland/deno/pull/8598: A modified version of path.resolve()  which will throw if the resolved path is relative. The onus of passing Deno.cwd() is on user. If the user would like to opt for an alternative with implicit Deno.cwd() call, fs.resolvePath can be utilized.

https://github.com/denoland/deno/pull/8559: Migrating path.relative to fs.relativePath for the same reason as #8387. 

Looking forward to hearing the discussion from the working group. Cheers:)

bartlomieju:

PR #8391, opened 2020-11-15
Improve Deno.version type declaration

MarkTiedemann:
I noticed that the `Version` interface was not exported. Thus, https://doc.deno.land/builtin/stable#Deno.version would not show its properties.

Bildschirmfoto am 2020-11-15 um 18 33 53

Since the interface `Version` is only used a single time, I inlined the properties - similar to [`Deno.build`](https://doc.deno.land/builtin/stable#Deno.build).

Also added documentation for the properties.


ry (approved):
Thanks Mark - LGTM

PR #8402, opened 2020-11-16
fix(websocket): Fix PermissionDenied error being caught in constructor

### LINKED ISSUE #8394: WebSocket silently exits if --allow-net is not enabled
```ts
const socket = new WebSocket('wss://gateway.discord.gg');

socket.onmessage = (ev) => console.log(ev.data);
```
WebSocket silently exits if --allow-net flag is not enabled. Expected behavior should be network access error.
* Deno: 1.5.2
* TypeScript: 4.x

crowlKats:
Fixes #8394


bartlomieju (commented):
Please add a unit test that assures `PermissionDenied` is thrown

ry:
I made a little tests, @crowlKats is this sufficient to demo the problem?

crowlKats:
@ry thanks! yes it would. there shouldnt even really be any need for the eventhandler

cli/tests/076_websocket_no_permission.js.out
```diff
@@ -0,0 +1 @@
+warning: network access to "ws://localhost/", run again with the --allow-net flag
```
bartlomieju, line 1: This is a bit surprising, is there a reason why we can't catch this error? What happens if there's `onerror` handler attached to the WebSocket? It'd be great to test that scenario as well

crowlKats: The error will be passed to the error event as well

cli/ops/websocket.rs
```diff
@@ -58,11 +79,7 @@ pub async fn op_ws_create(
   _bufs: BufVec,
 ) -> Result<Value, AnyError> {
   let args: CreateArgs = serde_json::from_value(args)?;
-  {
-    let s = state.borrow();
-    s.borrow::<Permissions>()
-      .check_net_url(&url::Url::parse(&args.url)?)?;
-  }
```
bartlomieju, line 65: @crowlKats please leave this block, but replace `?` with `.expect("Permission check should have been done in op_ws_check_permission")` and leave a comment about why we're having additional op for this purpose

bartlomieju (commented):
I think this is preferable solution to the problem instead of logging a warning to the console

bartlomieju (approved):
LGTM, thanks @crowlKats

PR #8403, opened 2020-11-16
fix: remove unnecessary back direction for import

behnammodi:
PR title is clear :) 
`assert.ts` is a sibling of `deep_assign.ts`


kitsonk (approved):
LGTM

Issue #8397, opened 2020-11-16
cargo test fails on windows


iammack:
The problem is introduced by this PR: https://github.com/denoland/deno/pull/8291. printf doesn't work on powershell.

getspooky:
Hi @iammack 
Can please share the terminal output error

iammack:
I don't have access to powershell for now. Can you answer this @MarkTiedemann as you have already provided a fix? Thanks!

MarkTiedemann:
Didn't run the test in PowerShell but in cmd, but the result should be the same:

```
error: Uncaught NotFound: The system cannot find the file specified. (os error 2)
```

This is because `printf` cannot be executed on Windows - there's no `printf.exe` binary.

More details: https://github.com/denoland/deno/pull/8404#issuecomment-728340753

ry:
fixed in https://github.com/denoland/deno/pull/8404

PR #8404, opened 2020-11-16
fix(cli/tests): Execute Windows binary in run permission test

MarkTiedemann:
See #8397.

There's no `printf.exe` on Windows (nor `echo.exe` - `echo` is a built-in command in `cmd.exe` and an alias for the `Write-Output` cmdlet in `powershell.exe`).

### DIFF
--- a/cli/tests/permission_test.ts
+++ b/cli/tests/permission_test.ts
@@ -17,10 +17,9 @@ const test: { [key: string]: (...args: any[]) => void | Promise<void> } = {
   },
   runRequired(): void {
     const p = Deno.run({
-      cmd:
-        Deno.build.os === "windows"
-          ? ["cmd.exe", "/c", "echo hello"]
-          : ["printf", "hello"],
+      cmd: Deno.build.os === "windows"
+        ? ["cmd.exe", "/c", "echo hello"]
+        : ["printf", "hello"],
     });
     p.close();
   },
--- a/cli/tests/permission_test.ts
+++ b/cli/tests/permission_test.ts
@@ -17,7 +17,10 @@ const test: { [key: string]: (...args: any[]) => void | Promise<void> } = {
   },
   runRequired(): void {
     const p = Deno.run({
-      cmd: ["printf", "hello"],
+      cmd:
+        Deno.build.os === "windows"
+          ? ["cmd.exe", "/c", "echo hello"]
+          : ["printf", "hello"],
     });
     p.close();
   },


bartlomieju:
@MarkTidemann I think we should revert #8291 as suggested in #8397. Ideally we'd use some generic solution that could work across all unit tests.

MarkTiedemann:
> Ideally we'd use some generic solution that could work across all unit tests.

I don't understand. Why does this one specific test case need to provide a generic solution that works for across all unit tests? Which unit tests would benefit from that?

---

I was just trying to fix the test (as I understood it):

**Run without permission:**

```
C:\dev\deno>deno run cli\tests\permission_test.ts runRequired
error: Uncaught PermissionDenied: access to run a subprocess, run again with the --allow-run flag
    at processResponse (core.js:224:13)
    at Object.jsonOpSync (core.js:248:12)
    at opRun (deno:cli/rt/40_process.js:19:17)
    at Object.run (deno:cli/rt/40_process.js:103:17)
    at Object.runRequired (file:///C:/dev/deno/cli/tests/permission_test.ts:19:20)
    at file:///C:/dev/deno/cli/tests/permission_test.ts:33:11
```

**Run with permission, but without the fix:**

```
C:\dev\deno>deno run --allow-run cli\tests\permission_test.ts runRequired
error: Uncaught NotFound: The system cannot find the file specified. (os error 2)
    at processResponse (core.js:224:13)
    at Object.jsonOpSync (core.js:248:12)
    at opRun (deno:cli/rt/40_process.js:19:17)
    at Object.run (deno:cli/rt/40_process.js:103:17)
    at Object.runRequired (file:///C:/dev/deno/cli/tests/permission_test.ts:19:20)
    at file:///C:/dev/deno/cli/tests/permission_test.ts:31:11
```

**Run with permission and with the fix:**

```
C:\dev\deno> deno run --allow-run cli\tests\permission_test.ts runRequired
C:\dev\deno> echo %errorlevel%
0
```

bartlomieju:
> I don't understand. Why does this one specific test case need to provide a generic solution that works for across all unit tests? Which unit tests would benefit from that?

@MarkTiedemann maybe I didn't explain myself correctly. The problem you're trying to fix stems from https://github.com/denoland/deno/issues/8254 - essentially we would like to get rid of all `python` calls in our tests. In #8291 one of the tests was rewritten to use `printf` instead of Python. As you found out this solution doesn't work on all Windows machines (it works on CI, because `bash` is available as a shell). So ideally I'd like to find a solution that would allow us to replace all invocations of `python`. Any suggestions are welcome.

MarkTiedemann:
> So ideally I'd like to find a solution that would allow us to replace all invocations of `python`

But there is only one single integration test for `--allow-run` permissions that executes a command. If there is no command or binary that is available by default on all platforms, then we need to provide different commands for different platforms. That's what this PR is about.

This test doesn't care whether `python` or anything else is executed. It doesn't even test whether "hello" is printed. As far as I understand the integration tests, it's just to make sure that, if `--allow-run` is required but not provided, it fails.

https://github.com/denoland/deno/blob/aaf7166a9d528636b11a2ad835238589bf650b6e/cli/tests/integration_tests.rs#L3184

And if it is required and provided, the invocation of the command does not fail.

https://github.com/denoland/deno/blob/aaf7166a9d528636b11a2ad835238589bf650b6e/cli/tests/integration_tests.rs#L3167

ry (approved):
Thanks @MarkTiedemann  - LGTM

@bartlomieju I think the solution is fine for this test and it's already done, so let's just go with it.

PR #8409, opened 2020-11-17
[fix]: std/node/events - Fix event extendability

Soremwar:
Previous implementation caused an extensability issue in which classes that inherited from EventEmitter and overwrote the method `on` and `addListener` would get stuck in an infinite loop since `on` calls `addListener` internally instead of being an alias for it.

Long story short, this is fixed.

image

### DIFF
--- a/std/node/events.ts
+++ b/std/node/events.ts
@@ -227,12 +227,7 @@ export default class EventEmitter {
    * listener will result in the listener being added, and called, multiple
    * times.
    */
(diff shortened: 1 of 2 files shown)


ry (commented):
Can you provide a small test for what this fixes?

Soremwar:
@ry Done

ry (approved):
LGTM - thanks!

PR #8418, opened 2020-11-18
feat(std/hash): add the sha1Hmac to the sha1

hackers267:
I notice that the sha256 has the sha256Hmac, but at the same time, the sha1 didn't has the sha1Hmac, so I add the sha1Hmac to the sha1!


ry (approved):
LGTM - thank you @hackers267

Issue #3736, opened 2020-01-21
module resolve failed with dynamic import and circulated import


Fenzland:
Files like this:
```javascript
// index.js
import ('./b.js');
import ('./a.js');
```
```javascript
// a.js
import './b.js';
```
```javascript
// b.js
import './c.js';
import './a.js';
```
```javascript
// c.js
import './d.js';
```
```javascript
// d.js
```

Then lets run with
```bash
deno --allow-read index.js
```
It crash with:
```
error: Uncaught Cannot resolve module "./d.js" from "file:///tmp/c.js"
```

Let's edit `b.js` to this
```javascript
// b.js
import './a.js';
import './c.js';
```
Then run
```bash
deno --allow-read index.js
```
It crash with:
```
error: Uncaught Cannot resolve module "./c.js" from "file:///tmp/b.js"
```

* when we **dynamically** import b.js and a.js in index.js **concurrently**, and **b.js first**;
* and a.js and b.js **statically** import each other;
* then we cannot statically import anything in b.js after import a.js.
* and we cannot statically import anything in c.js that imported by b.js before b.js import a.js.

```javascript
// b.js
import './a.js';
// cannot statically import anything here
```
```javascript
// c.js
// cannot statically import anything here
```

This BUG exists since we support `import()`. I've encountered it several times. 
The first time is under 0.15 or 0.16. Today I, finally find out how to reproduce it.

lucacasonato:
@Fenzland Can you still reproduce this? This might have been resolved by the module graph refactor @bartlomieju did a while ago.

bartlomieju:
It was not, this error is still valid. Waiting on modules refactor in core/ by @piscisaureus

PR #8421, opened 2020-11-18
[WIP] fix(core): circular dynamic imports

bartlomieju:
Trying to fix https://github.com/denoland/deno/issues/3736


bartlomieju:
Fixed in #11089!

PR #8423, opened 2020-11-18
fix(cli): add file URL support for Deno.readLink

### LINKED ISSUE #7639: Deno.readLink path parameter is string only (not string | URL like Deno.readFile)

wperron:
Fixes #7639


ry:
Thanks @wperron looks good. I think this should wait until 1.6.0.

bartlomieju (approved):
LGTM, thanks @wperron

PR #8435, opened 2020-11-19
refactor(cli): rename fmt_errors::JsError to PrettyJsError

bartlomieju:
This commit renames "fmt_errors::JsError" to "PrettyJsError"
to avoid confusion with "deno_core::JsError".

Consequently "CoreJsError" aliases to "deno_core::JsError"
were removed.

Additionally source mapping step has been removed from
"PrettyJsError::create" to better separate domains.


ry (approved):
LGTM

PR #8438, opened 2020-11-19
refactor(cli): worker event serialization

bartlomieju:
This commit simplifies code responsible for serialization
of web worker events. Instead of using "PrettyJsError"
it's enough to downcast to "deno_core::JsError", making
the code less specific to cli/ implementation.


bartlomieju:
CC @nayeemrmn

ry (approved):
LGTM

PR #8443, opened 2020-11-20

bartlomieju:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,10 +90,10 @@ checksum = "fd6ee2941db3551563d29eaf5214cd3d7b2f322e0c0e3954f5ae020f860bae8c"
 dependencies = [
  "darling",
  "pmutil",
- "proc-macro2 1.0.21",
+ "proc-macro2 1.0.24",
  "quote 1.0.7",
  "swc_macros_common",
- "syn 1.0.41",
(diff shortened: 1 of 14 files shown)


bartlomieju:
@kdy1 @kitsonk PTAL, lots of different output from the bundler, I think that's expected, but want to double check.

kdy1:
It seems like deno is not using inliner. The bundler switched from slow inliner to very fast **constant** propagation pass.

Note that the bundler now creates lots of more variables to make flattening easier.

(Previous approach was a result of misunderstanding about the nature of flattening.)

kdy1:
The output seems correct.. (I forgot saying it)

bartlomieju:
> It seems like deno is not using inliner. The bundler switched from slow inliner to very fast constant propagation pass. 
> 
Should we be using it? Could you point me to relevant API @kdy1?

kdy1:
@bartlomieju 

It's not must, and I don't think it matters for users who will run the bundle via deno.
But some users want smaller bundle, so it may help.

Anyway, it's enabled by default and https://docs.rs/swc_bundler/0.17.1/swc_bundler/struct.Config.html#structfield.disable_inliner will help.

kitsonk (approved):
LGTM

Issue #8313, opened 2020-11-09
tests in macos are flakey


lucacasonato:
Examples:

https://github.com/denoland/deno/actions/runs/352903079
https://github.com/denoland/deno/actions/runs/352901785

Tests hang after std_tests

ry:
another example https://github.com/denoland/deno/runs/1405739951

bartlomieju:
This time on windows: https://github.com/denoland/deno/pull/8424/checks?check_run_id=1420536916

It seems it's related to TLA implementation:
```
error: error: Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promise.
Check file:///D:/a/deno/deno/cli/tests/permission_test.ts
test [examples/chat] GET /ws should upgrade conn to ws ... test std_tests ... FAILED
```

kitsonk:
OSX example: https://github.com/denoland/deno/pull/8162/checks?check_run_id=1437858067#step:16:3917

bartlomieju:
Linux example: https://github.com/denoland/deno/runs/1455721727
Actually hanging on tests that interact with tty
```
test _063_permissions_revoke ... test _063_permissions_revoke has been running for over 60 seconds
test _064_permissions_revoke_global ... test _064_permissions_revoke_global has been running for over 60 seconds
test _066_prompt ... test _066_prompt has been running for over 60 seconds
test _067_test_no_run_type_error ... test _067_test_no_run_type_error has been running for over 60 seconds
test repl_test_pty_multiline ... test repl_test_pty_multiline has been running for over 60 seconds
```

bartlomieju:
Another one: https://github.com/denoland/deno/runs/1461003063
```
test pbkdf2Sync hashes data correctly ... test test_permissions_without_allow ... test test_permissions_without_allow has been running for over 60 seconds
test test_raw_tty ... test test_raw_tty has been running for over 60 seconds
```

I'm pretty confident the problem is with tests that use tty

kitsonk:
Another one: https://github.com/denoland/deno/pull/8514/checks?check_run_id=1462070080#step:19:4020

This was a problem with run_watch and unexpected stdin result.

bartlomieju:
Yet another hang related to REPL/PTY:
```
2020-11-28T14:12:23.0196390Z test repl_test_pty_multiline ... test repl_test_pty_multiline has been running for over 60 seconds
[...]

PR #8445, opened 2020-11-20
test(std): make test output less noisy

bartlomieju:
This commit makes output of std/ tests less noisy
by passing "--quiet" flag to Deno subprocesses run
as part of test suite.


bartlomieju:
Trying to debug flaky tests reported in https://github.com/denoland/deno/issues/8313 but there's a lot of output so it's quite hard to spot what's going wrong/hangs.

ry (approved):
LGTM

PR #8447, opened 2020-11-20
fix(std/bufio): Remove '\r' at the end of Windows lines

### LINKED ISSUE #7093: readLines defaults to newline exclusively, causing errors on Windows
It's not that this is specifically unexpected behavior, but this function is often explicitly provided as the solution to a problem that it does not solve. 

When tasked with parsing lines of a text file on windows, one must use `\r\n` or go with a platform-agnostic regex such as `/\r?\n/`. The solution offered is often parsing the "correct way" using `readLines` which does not factor in `\r` and leads to the same problems mentioned in https://github.com/denoland/deno/issues/7090, which is essentially a garble of text on Windows machines when displayed with `console.log`.

Current `readLines` implementation for reference: https://deno.land/std@0.65.0/io/bufio.ts#L704

This was made a separate issue as to not clutter the discussion on the main problem with carriage returns in Deno's `console.log`

kanales:
Fix #7093

Removes "\r" from the end of lines when using `readLines`.

For example for a file containing `"A\r\nB\r\nC"`:
```javascript
let fr = await Deno.open("./file.txt");
for await (let line of readLines(fr)) {
  console.log([line]);
}

/* 
Before:
[ "A\r" ]
[ "B\r" ]
[ "C" ]

Now:
[ "A" ]
[ "B" ]
[ "C" ]
*/
```

### DIFF
--- a/std/io/bufio.ts
+++ b/std/io/bufio.ts
@@ -706,5 +706,15 @@ export async function* readStringDelim(
 export async function* readLines(
   reader: Reader,
 ): AsyncIterableIterator<string> {
-  yield* readStringDelim(reader, "\n");
+  const encoder = new TextEncoder();
+  const decoder = new TextDecoder();
+  for await (let chunk of readDelim(reader, encoder.encode("\n"))) {
+    // Finding a CR at the end of the line is evidence of a
+    // "\r\n" at the end of the line. The "\r" part should be
+    // removed too.
+    if (chunk[chunk.length - 1] == CR) {
+      chunk = chunk.slice(0, chunk.length - 1);
+    }
+    yield decoder.decode(chunk);
+  }
 }
--- a/std/io/bufio.ts
+++ b/std/io/bufio.ts
@@ -706,15 +706,13 @@ export async function* readStringDelim(
 export async function* readLines(
   reader: Reader,
 ): AsyncIterableIterator<string> {
-  const encoder = new TextEncoder();
-  const decoder = new TextDecoder();
-  for await (let chunk of readDelim(reader, encoder.encode("\n"))) {
+  for await (let chunk of readStringDelim(reader, "\n")) {
     // Finding a CR at the end of the line is evidence of a
     // "\r\n" at the end of the line. The "\r" part should be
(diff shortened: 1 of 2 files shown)


bartlomieju:
@kanales thanks for PR, can we get some test case showcasing what's being fixed?

kanales:
Yes of course! I'll take care of it as soon as I can.

kanales:
After reading the test cases in `std/io/bufio_tests.ts`, I decided to extend the "readStringDelimAndLines" test to check for windows-style newlines. I hope that's OK!

bartlomieju:
@nayeemrmn could you review?

nayeemrmn (approved):
LGTM

std/io/bufio.ts
```diff
@@ -706,5 +706,15 @@ export async function* readStringDelim(
...
+  const decoder = new TextDecoder();
+  for await (let chunk of readDelim(reader, encoder.encode("\n"))) {
+    // Finding a CR at the end of the line is evidence of a
+    // "\r\n" at the end of the line. The "\r" part should be
+    // removed too.
+    if (chunk[chunk.length - 1] == CR) {
+      chunk = chunk.slice(0, chunk.length - 1);
+    }
+    yield decoder.decode(chunk);
+  }
```
nayeemrmn, line 719: Actually, couldn't it be written like this?
```suggestion
  for await (let chunk of readStringDelim(reader, "\n")) {
    // Finding a CR at the end of the line is evidence of a
    // "\r\n" at the end of the line. The "\r" part should be
    // removed too.
    if (chunk.endsWith("\r")) {
      chunk = chunk.slice(0, -1);
    }
    yield chunk;
  }
```

kanales: Yes, you are right. At the moment I assumed working with `Uint8Array` was preferable to `string`  for performance reasons, but honesly I can't think of any reason why. Your suggestion is certainly cleaner and easier to understand, thank you!

bartlomieju (approved):
LGTM, thank you @kanales

PR #8451, opened 2020-11-21
refactor(std/http): Rename validateCookieName param name & JSDoc

getspooky:


bartlomieju (approved):
LGTM, thanks @getspooky

PR #8457, opened 2020-11-21
feat(std/http): Validate cookie path value

getspooky:

### DIFF
--- a/std/http/cookie.ts
+++ b/std/http/cookie.ts
@@ -70,6 +70,7 @@ function toString(cookie: Cookie): string {
     out.push(`SameSite=${cookie.sameSite}`);
   }
   if (cookie.path) {
+    validatePath(cookie.path);
     out.push(`Path=${cookie.path}`);
   }
   if (cookie.expires) {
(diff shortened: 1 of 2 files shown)


std/http/cookie.ts
```diff
@@ -92,6 +93,26 @@ function validateCookieName(name: string | undefined | null): void {
...
+ * @param path Path value.
+ */
+function validatePath(path: string | null): void {
+  if (path == null) {
+    return;
+  }
+  for (let i = 0; i < path.length; i++) {
+    const c = path.charAt(i);
+    if (
+      c < String.fromCharCode(0x20) || c > String.fromCharCode(0x7E) || c == ";"
```
nayeemrmn, line 107: Reference?

getspooky: https://tools.ietf.org/html/rfc6265#section-4.1.2.4

nayeemrmn: Okay, found it in https://tools.ietf.org/html/rfc6265#section-4.1.1:
```
path-value        = <any CHAR except CTLs or ";">
```
(where a `CHAR` is an ASCII character).

nayeemrmn (approved):
LGTM, thanks.

std/http/cookie.ts
bartlomieju: Please add this reference as a comment

bartlomieju (approved):
LGTM, thanks @getspooky

Issue #8450, opened 2020-11-21
`deno doc` doesn't respect compiler hints


Liamolucko:
At the moment, `deno doc` doesn't respect `/// <reference types="..." />` nor `X-TypeScript-Types` nor `@deno-types`.

### Example:
`foo.js`
```javascript
/// <reference types="./foo.d.ts" />
export const foo = "foo";
```
`foo.d.ts`
```typescript
export const foo: string;
```
#### Expected output of `deno doc foo.js`:
`const foo: string`
#### Actual:
`const foo`

I originally posted this issue as denoland/deno_doc#53, but these compiler hints are CLI specific.

PR #8459, opened 2020-11-22
feat(cli/doc): use type definitions for `deno doc` if available

Liamolucko:
fixes https://github.com/denoland/deno_doc/issues/79


bartlomieju:
@Liamolucko thank you for the PR, but to be honest I'm not super happy with this solution. #8450 is an actual problem that we should fix, but it should be done by unifying interfaces across the project. Ie. we should update `deno_doc` to take different trait for the loader - preferably the one from `deno_core`. Unfortunately that would lead to circular dependency between `deno`, `deno_core` and `deno_doc` crates. I need to think about possible solution to this problem a bit more.

ry:
This feature seems very awesome. I wish we could get it in somehow. @bartlomieju has anything changed in the last months?

Liamolucko:
Why isn't CI running?

bartlomieju:
> This feature seems very awesome. I wish we could get it in somehow. @bartlomieju has anything changed in the last months?

No, not really, I'm not sure if it makes sense to change interfaces in `deno_core` after all. I will let Kitson take a look at this PR before landing as it utilizes module graph.

std/node/_fs/_fs_stat_test.ts
```diff
@@ -17,9 +17,9 @@ export function assertStats(actual: Stats, expected: Deno.FileInfo) {
   assertEquals(actual.atime?.getTime(), expected.atime?.getTime());
   assertEquals(actual.mtime?.getTime(), expected.mtime?.getTime());
   assertEquals(actual.birthtime?.getTime(), expected.birthtime?.getTime());
-  assertEquals(actual.atimeMs, expected.atime?.getTime());
-  assertEquals(actual.mtimeMs, expected.mtime?.getTime());
-  assertEquals(actual.birthtimeMs, expected.birthtime?.getTime());
+  assertEquals(actual.atimeMs ?? undefined, expected.atime?.getTime());
+  assertEquals(actual.mtimeMs ?? undefined, expected.mtime?.getTime());
+  assertEquals(actual.birthtimeMs ?? undefined, expected.birthtime?.getTime());
```
bartlomieju, line 22: Changes in this file seem unrelated, please revert

bartlomieju (changes requested):

std/node/_fs/_fs_stat_test.ts
Liamolucko: Ok, I've moved them to #9311.

bartlomieju:
@Liamolucko please rebase, I'll try to get this PR in for `1.8.0`

cli/main.rs
```diff
@@ -684,56 +682,49 @@ async fn bundle_command(
   Ok(())
 }
 
-struct DocLoader {
-  fetcher: FileFetcher,
-  maybe_import_map: Option<ImportMap>,
+/// When parsing lib.deno.d.ts, only `DocParser::parse_source` is used,
+/// which never even references the loader, so this is just a stub for that scenario.
+struct StubDocLoader;
```
kitsonk, line 687: We should refactor better so that we don't have some totally useless code.  If it isn't needed, the consumer should have an optional reference to a `impl DocFileLoader` versus a "stub".

bartlomieju: Agree, however it requires changing APIs in `deno_doc`. I'd be fine with landing as is for 1.8 as this PR actually fixes a problem. 

@Liamolucko please add a TODO here mentioning this required change to `deno_doc`. Also it'd be great if code handling doc generation was moved to `tools/doc.rs`.

bartlomieju (approved):
LGTM, thanks Liam and sorry for slow review

lucacasonato:
This PR broke `deno doc` for TSX files:

```tsx
// test.tsx
export function docs() {
  return <div></div>;
}
```

```shell
$ deno doc test.tsx
Unterminated regexp literal at file:///mnt/starship/test.tsx:3:15
```

Canary cdae4423c27443e0085d59f45eda1c978f186a1c is broken, canary 687ff2ab14d6a735aa5bf0ec57ef00cfe0c04e4b is fine.

I will open an issue. If we can not resolve this issue until tomorrow afternoon we should revert this PR so it does not make it into 1.8.

PR #7293, opened 2020-08-31
BREAKING(std/log): simplification rewrite

timreichen:
#7211

- replace `setup` and global state by access to `Logger` instances to create default logger
- simplify by passing `Handler` objects instead of strings
- rename msg to message (like in [Error.prototype.message](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message))
- add LogLevelNames enum
- rename logger.warning to warn (like in [console.warn](https://developer.mozilla.org/en-US/docs/Web/API/Console/warn)) and LogLevelNames.WARN
- remove string formatter in favor of template strings
- fix `unknown` Logger and string setup problem

Issue #8405, opened 2020-11-16
Standard Library Working Group


ry:
The standard library in Deno has seen a lot of growth and improvement over the past year. We think it is time to review and stabilize these APIs, guaranteeing that users will not experience painful changes in the future. In order to avoid cementing mistakes, we want to review these APIs, one-by-one, and come to a consensus on whether they should be included, modified, or excluded.

We will be holding a series of virtual events every two weeks. We will (perhaps tediously 😄 ) review APIs and hear any objections. Because there are so many APIs to review, this will take place over several months.

In order to accommodate people in different timezones, we will have an alternating timing schedule.

This issue will remain pinned and this message updated as we work through each of the modules. If you would like to participate, please join https://groups.google.com/g/deno-std 

December 3 (14:00-16:00 GMT)
- Discussion around std scope, process for adding new modules, and versioning
- async
- bytes

December 16 (21:00 - 23:00 GMT)
- archive
- datetime
- encoding
- flags

January 7 (time TBD)
- fmt
- fs
- hash

January 21 (time TBD)
- http
- io
- log

February 4 (time TBD)
- mime
- node
- path

February 18 (time TBD)
- permissions
- signal
- testing

March 4 (time TBD)
- textproto
- uuid
- wasi
- ws

bartlomieju:
Important topic to discuss (preferably sooner than later) is `std` versioning. We won't stabilize all APIs and there will be new APIs added in the future, therefore we should establish some convention similar to `--unstable` flag for built-in `Deno` APIs.

MarkTiedemann:
Another important topic to discuss (and write down) is the boundaries between core, std, and third-party modules. What are the criteria for APIs to be included in core? In std?

For example, in Node, the HTTP server is in core. Why is it in std in Deno? In Node, a TOML parser is a third-party module. Why is it in std in Deno?

I think having clearly defined boundaries between these realms would be helpful.

hazae41:
> Another important topic to discuss (and write down) is the boundaries between core, std, and third-party modules. What are the criteria for APIs to be included in core? In std?
> 
> For example, in Node, the HTTP server is in core. Why is it in std in Deno? In Node, a TOML parser is a third-party module. Why is it in std in Deno?
> 
> I think having clearly defined boundaries between these realms would be helpful.

- core aims to implement Web APIs and includes the native stuff
- std aims to standardize common modules
- third-parties are modules that don't fit in std

[...]

PR #8460, opened 2020-11-22
breaking(std/log): rewrite and custom LogLevels support

timreichen:
- remove global state
- class based Loggers
- class based LogLevels
- custom LogLevels support
- logLevels
  - trace
  - debug
  - info
  - warn
  - error

### About default logLevels

There are different conventions, implementations and logLevel names in different loggers and languages. Sadly there is no standard specification for loglevels. Though [heavily debated](https://github.com/denoland/deno/pull/7293) I decided to remove the `critical` loglevel for the following reasons:
- rust/log oriented (https://docs.rs/log/0.4.11/log/enum.Level.html), which also aligns with console methods.
- `fatal` and `critical` are sometimes used the same way in lots of log modules in different languages. Choosing one over the other could not be justified except with personal preferences.
- There is inconsistency between `fatal` and `critical` in different log modules: Some terminate the process after the call, since the error is unrecoverable, some don't.
- With the option to extend and use custom loglevels, the user can implement a `fatal` or `critical` themselves.


std/log/logger.ts
```diff
@@ -37,153 +52,58 @@ export class LogRecord {
...
-  ): T | undefined {
-    if (this.level > level) {
-      return msg instanceof Function ? undefined : msg;
+const DEFAULT_LOGGER_NAME = "logger";
+
+export class BaseLogger {
+  name: string;
+  logLevel: LogLevel;
+  handlers: Handler[];
+  constructor(logLevel: LogLevel, {
```
vwkd, line 61: Currently, when logging to an empty logger it throws an error.

```js
const logger = new Logger();

logger.warn("hello world");
```

I suggest making the whole constructor argument of `BaseLogger` an object, and use the default log level `INFO`.

```js
const DEFAULT_LOG_LEVEL = logLevels.info;

export class BaseLogger {
    /* ... */
    constructor({
        logLevel = DEFAULT_LOG_LEVEL,
        name = DEFAULT_LOGGER_NAME,
        handlers = [],
     }
    /* ... */
}
```

std/log/levels.ts
```diff
@@ -1,59 +1,20 @@
...
-  }
-  throw new Error(`no level name found for level: ${level}`);
-}
+export const logLevels = {
+  trace: new LogLevel("Trace", 10),
+  debug: new LogLevel("Debug", 20),
+  info: new LogLevel("Info", 30),
+  warn: new LogLevel("Warn", 40),
+  error: new LogLevel("Error", 50),
+};
```
vwkd, line 20: Not breaking existing consumers would take as much as adding here

```js
critical: new LogLevel("Critical", 50),
```

and the corresponding method to the `Logger` class in `logger.ts`

```js
critical(message: unknown, ...args: unknown[]) {
    return this.dispatch(logLevels.critical, message, ...args);
}
```

vwkd (commented):
Seems like you closed the original PR #7293 and opened a new one to get rid of the discussion. You didn't even link to it as if it never existed. 🤐 👎

----

I appreciate that adding custom log levels is now much easier, and that the default logging handler is back to none. That's a step in the right direction. However, the following issues still remain. This PR

- removes log levels
- changes the default log level

I'll try to summarize the points from #7293.

---

## Log levels

Compared to the existing implementation, this PR removes the `critical` log level at the top and adds the `trace` log level at the bottom.

Adding more log levels doesn't hurt existing consumers. They don't obstruct, if you don't use them, you don't see them.

However taking away log levels is a breaking operation. Existing consumers based their categorization of logging statements on the existing log levels. Any reduction means existing consumers need to re-evaluate every single logging statement because they must fit them into fewer categories! Taking even a single log level away that consumers already use breaks their *entire* logging logic.

Although the total number of log levels doesn't change, since they are removed and added from opposite ends, existing consumers still experience full breakage.

I'd have had no problem, would this have been a new logging framework without consumers. However this is not designing a new logging framework, this is changing an existing one. Even more, since adding back the `critical` log levels doesn't obstruct (since it's invisible to you if you don't use it), breaking existing consumers seems an unnecessarily high cost.

So is it worth taking away `critical`? Leaving it doesn't cost anything, while removing it badly breaks existing consumers. This is the hard question you need to ask! I want you to ask it consciously and transparently.

Regarding the other points, I don't think these are good reasons.

> * `fatal` and `critical` are sometimes used the same way in lots of log modules in different languages. Choosing one over the other could not be justified except with personal preferences.

`critical` and `fatal` are different names for the same level. I'm not aware of any logging framework that uses both.

> * There is inconsistency between `fatal` and `critical` in different log modules: Some terminate the process after the call, since the error is unrecoverable, some don't.

The log levels don't do anything to your code like "terminate the process". They are simply labels for a log message. The developer decides on the categorization of their logging statements. It's typical to use `critical` / `fatal` when the process is going to terminate, but that is in no way a rule.

> * [..] aligns with console methods.

Make the logging framework match the `console` was the main argument of the original PR. This is a rather arbitrary choice, since the console has not much to do with a logging framework, apart from being one of multiple output channels. In most other language the console output is a simple `print` function which makes it more obvious that an alignment to `console` is a purely arbitrary choice.

So the only real reason is that these 5 log levels match Rust, while Python, Go, Ruby, PHP, C# use 6, and others [even more](https://crimsonpinnacle.com/2020/08/05/log-levels-and-how-to-properly-use-them-part-1-common-log-levels/). Again, this is not designing a new logging framework, but changing an existing one, where the cost of removing a log level is high, while leaving it doesn't cost you anything. Literally, to implement these are 4 LOC, which you can see in my review comment.

---

## Default log level

Compared to the existing implementation, this PR changes the default log level of a logger from `INFO` to `undefined`. The previous PR changed it to `DEBUG` which was a pretty arbitrary choice and gladly changed back.

The log level `INFO` is a common choice for production, since it gives the neccessary context where in the program warnings and errors occurred. In production one usually doesn't want to see debug or trace statements, since they pollute the important warnings and errors. They are turned on when debugging a previously found error.

This makes `INFO` a good default as in the existing implementation. The PR author hasn't yet presented a good argument against it.

The `undefined` default log level can't be left as is, since it throws an error when logging to an empty logger. Currently, this throws an error.

```js
const logger = new Logger();

logger.warn("hello world");
```

timreichen:
@vwkd I am sorry, I meant to link to the previous discussion.
Thank you for summarizing the it.

# Log levels
## About critical/fatal
I think we both have laid out the arguments and do not agree on that one. While agree with you on the functionality aspects, I also think it is a question of design. While some want `critical`, other might want `fatal`. Sure, they can add `fatal`, but there will still be `critical`, which doesn't feel like a nice design to me and begs the question why not having aliases for all loglevel with frequently used names (not that I would be in favor of that).
Since this PR implements custom loglevels, it would be easy to create a file that extends Logger and all Handlers with a `critical`loglevel. 
But I think the core deno team should decide. I am happy to add the critical loglevel back if it is the explicit decision of the deno core team.

# About default LogLevel
The PR makes loglevel for `Logger` and `Handler` instances mandatory and therefore there is no default. I think this is the best way to address the issue and confusion of different log frameworks having different default loglevels and makes it more expressive.

```ts
const logger = new Logger();
```
will throw a typescript error while

```ts
const logger1 = new Logger(logLevels.info);

const logger2 = new Logger(logLevels.info, { handlers: [someHandler] });
```
will work.

This is also the reason why it is not part of the constructor object, which is optional, though it could be in the object. Is there any guideline/convention on which to use?
```ts
const logger1 = new Logger(logLevels.info, {…});
// or
const logger1 = new Logger({
  logLevel: logLevels.info,
…
});
```
# About changing or new 
This PR will break the current consumer code. So I think we don't need to take special consideration for that since the code needs to be adopted anyway.

lucacasonato:
Can I suggest we move this discussion to a stdwg meeting? (#8405). If you all have such strong opinions on the `std/log` module, maybe `std/log` can be moved out of std and become a third party module? This logger already deviates greatly from the Go `log` module (we strive to stick to Go's standard library closely).

ssslambda:
I agree with the previous comment, std/log is really now different a lot from go log

bartlomieju:

Issue #5006, opened 2020-04-30
Add cargo-deny to CI


bartlomieju:
https://embarkstudios.github.io/cargo-deny/index.html

`cargo-deny` provides checks for security advisories and licence checks. 

It'd be very useful to add it to CI to ensure all dependencies have proper licenses and we're up to date with dependencies.

Sparkenstein:
Need some more inputs on this one. The integration of `cargo-deny` with CI is easy but to configure it, it's better if maintainers decide which checks to perform and where to fail etc. Writing short introduction about `cargo-check` checks below so that you don't need to go through the entire docs. It performs 4 types of checks:

- advisories
- licenses
- bans
- sources

as the names suggest these checks look for security vulnerabilities etc, licenses information, deny/allow specific crates and sources location respectively. I went though the recommended configuration, and here's the configuration I came up with: https://github.com/denoland/deno/compare/master...Sparkenstein:cargo-deny . note that I am not familiar with Github CI so I referred `cargo-check`s recommended configuration from the docs.

Now the question is which of these checks would you like to perform in CI? Also what should be the severity of these checks? for current configuration, only Licenses checks are failing as there are few UNLICENSED crates. (You also need to evaluate carefully which licenses should be allowed explicitly in the config). for other checks there are couple of warns about deprecated checks and duplicate entries etc.

bartlomieju:
@Sparkenstein thanks for detailed analysis. For different checks:
- advisories - should be a hard error
- sources - only allow official crates.io, otherwise hard error
- bans - multiple versions should be a warning, the rest hard error
- licenses - I'll get back to you on the list of allowed licences (for sure it'll be MIT), but it should be a hard error as well

Could you point out which crates are unlicensed?

bartlomieju:
@Sparkenstein please open a PR, it will be easier to iterate on living code :)

Sparkenstein:
@bartlomieju the respective PR was closed recently I believe this issue can be closed as well

PR #7988, opened 2020-10-15
[WIP] upgrade: tokio 0.3

bartlomieju:
Closes #7984

PR #8462, opened 2020-11-22
build: cargo-deny integration

Sparkenstein:
This pull request tries to solve the #5006 

Continuing the discussion from thread:
> Could you point out which crates are unlicensed

internal crates test_util and test_plugin are unlicensed according to cargo-check. attaching log of `cargo check` on license:
https://del.dog/dimerastoj.log

will push other changes to severity soon

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -9,7 +9,7 @@ targets = [
 # The lint level for security vulnerabilities
 vulnerability = "deny"
 # The lint level for unmaintained crates
-unmaintained = "warn"
+unmaintained = "deny"
 # The lint level for crates that have been yanked from their source registry
 yanked = "deny"
 # The lint level for crates with security notices. Note that as of
@@ -55,7 +55,7 @@ license-files = [
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
 # Lint level for when multiple versions of the same crate are detected
-multiple-versions = "deny"
+multiple-versions = "warn"
 # Lint level for when a crate version requirement is `*`
 wildcards = "allow"
 # The graph highlighting used when creating dotgraphs for crates
--- a/deny.toml
+++ b/deny.toml
@@ -11,11 +11,11 @@ vulnerability = "deny"
 # The lint level for unmaintained crates
 unmaintained = "warn"
 # The lint level for crates that have been yanked from their source registry
-yanked = "warn"
+yanked = "deny"
 # The lint level for crates with security notices. Note that as of
 # 2019-12-17 there are no security notice advisories in
 # https://github.com/rustsec/advisory-db
-notice = "warn"
+notice = "deny"
 
 [licenses]
 # The lint level for crates which do not have a detectable license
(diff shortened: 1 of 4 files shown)


bartlomieju:
> internal crates test_util and test_plugin are unlicensed according to cargo-check. attaching log of cargo check on license:
https://del.dog/dimerastoj.log

Thanks @Sparkenstein, we should add `MIT` to the `license` field in these crates. It's quite surprising that `ring` is missing a license though.

deny.toml
```diff
@@ -11,11 +11,11 @@ vulnerability = "deny"
 # The lint level for unmaintained crates
 unmaintained = "warn"
```
bartlomieju, line 12: Should be `deny`

deny.toml
```diff
@@ -30,16 +30,32 @@ confidence-threshold = 0.8
...
+expression = "ISC AND MIT AND OpenSSL"
+license-files = [
+  {path = "LICENSE", hash = 0xbd0eed23},
 ]
 
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
 # Lint level for when multiple versions of the same crate are detected
-multiple-versions = "warn"
+multiple-versions = "deny"
```
bartlomieju, line 58: This one should probably be `warn` - we're aware there are multiple versions of some crates but unfortunately can't do anything about it right now.

Sparkenstein:
In this commit:
 - added CC0-1.0 for `notify` crate
 - allowed OpenSSL license
 - added MIT for test util and test plugin
 - added exception for `ring` crate
 - hardened all the checks

Now as you see in the CI, because of `bans` checks are now on `deny`, the CI is failing because there are
multiple duplicate crates. and I think it is fine. the thing is we can skip duplicate version checks for these crates in config but
the list will be too long to maintain. I was thinking if we should skip the `bans` check altogether? it checks only for duplicate crates and banned crates which we don't have any as of now at least.

deny.toml
```diff
@@ -0,0 +1,76 @@
...
+license-files = [
+  {path = "LICENSE", hash = 0xbd0eed23},
+]
+
+# https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
+[bans]
+# Lint level for when multiple versions of the same crate are detected
+multiple-versions = "warn"
+# Lint level for when a crate version requirement is `*`
+wildcards = "allow"
```
bartlomieju, line 60: This setting should be `deny`

bartlomieju:
Looks like we got two problematic dependencies: `rand 0.6.3` coming from `multipart 0.17.0` (no update available to the latter) and `net 0.2` coming from `mio 0.6` (which could be fixed by upgrading to Tokio 0.3, which is in progress in #7988)

bartlomieju:
@Sparkenstein we did update our dependencies to latest versions recently, could you please rebase the PR? I'm still interested in landing it.

ry:
I'm worried about the complexity this adds to CI. What if we just had a script we could run once in a while?

I don't think it's possible to completely abstract away the cost of adding dependencies. All dependencies to Deno must be reviewed and considered - part of that review is license check. If this auto check was completely free, it'd be fine, but it's adding more dependencies, and almost 100 lines of code that have to be maintained. I'm not sure it's worth it.

bartlomieju:
@Sparkenstein thank you for the PR, but after more consideration I agree with Ryan. Additionally I'm worried that we might be getting sudden failures on CI after bumping dependency versions in similar case as with `ring` when you opened a PR.

Thanks again, but respectfully closing without merge.

PR #8463, opened 2020-11-22
[WIP] refactor(cli): inspector creation in worker

bartlomieju:

### DIFF
--- a/.dprintrc.json
+++ b/.dprintrc.json
@@ -31,8 +31,8 @@
     "third_party"
   ],
   "plugins": [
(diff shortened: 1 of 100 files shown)


bartlomieju:
@caspervonb maybe you'll have idea why inspector suddenly stops working if it's initialized outside `Worker::new`?

caspervonb:
Think it uses TLS internally on the v8 side, not too sure without digging into it tho.

bartlomieju:
I'll try other approach

PR #8470, opened 2020-11-23
fix(cli/websocket): set User-Agent header

crowlKats:


bartlomieju:
@crowlKats great, could you add some assert regarding user agent in a test?

bartlomieju (approved):
LGTM

PR #8471, opened 2020-11-23
feat(std/http): Add Cookie value validation 

getspooky:


bartlomieju:
@nayeemrmn can you take a look?

std/http/cookie.ts
```diff
@@ -114,6 +115,31 @@ function validatePath(path: string | null): void {
...
+function validateCookieValue(value: string | null): void {
+  if (value == null) return;
+  for (let i = 0; i < value.length; i++) {
+    const c = value.charAt(i);
+    if (
+      c < String.fromCharCode(0x21) || c == String.fromCharCode(0x22) ||
+      c == String.fromCharCode(0x2c) || c == String.fromCharCode(0x3b) ||
+      c == String.fromCharCode(0x5c) || c == String.fromCharCode(0x7f)
+    ) {
+      throw new Error("RFC2616 cookie value cannot have '" + c + "'");
```
bartlomieju, line 132: Nitpick, include cookie name in the error

std/http/cookie.ts
```diff
@@ -114,6 +115,31 @@ function validatePath(path: string | null): void {
...
+    }
+    if (c > String.fromCharCode(0x80)) {
+      throw new Error(
+        "RFC2616 cookie value can only have US-ASCII chars" +
+          c.charCodeAt(0).toString(16),
+      );
+    }
```
bartlomieju, line 139: Ditto

bartlomieju (changes requested):

std/http/cookie.ts
getspooky: what do you think ? 
 ```js
"RFC2616 cookie '" + name + "' cannot have '" + c + "' as value" 
```

File: std/http/cookie.ts (hunk 2 above)
Comment (getspooky) [reply std/http/cookie.ts:139]:
```js
 "RFC2616 cookie '" + name + "' can only have US-ASCII chars as value" +
          c.charCodeAt(0).toString(16)
```

getspooky:
@bartlomieju i made changes

bartlomieju (approved):
LGTM, thanks @getspooky

PR #8474, opened 2020-11-23
fix(std/http): prevent path traversal

sarahdenofiletrav:
Fix path traversal problem when the "request-URI" does not have a leading slash.
This is possible in some http clients such as curl:
curl --path-as-is localhost:4705 --request-target ../../../..

The file server now returns HTTP 400 when requests lack the leading slash, and are not
absoluteURIs (see https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html).

### DIFF
--- a/std/http/file_server_test.ts
+++ b/std/http/file_server_test.ts
@@ -94,7 +94,9 @@ async function fetchExactPath(
   const request = encoder.encode("GET " + path + " HTTP/1.1\r\n\r\n");
   let conn: void | Deno.Conn;
   try {
-    conn = await Deno.connect({ hostname: hostname, port: port });
+    conn = await Deno.connect(
+      { hostname: hostname, port: port, transport: "tcp" },
+    );
     await Deno.writeAll(conn, request);
     let currentResult = "";
     let contentLength = -1;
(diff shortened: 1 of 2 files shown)


sarahdenofiletrav:
I do not know why the tests are failing; they pass on my machine.  It is throwing in a call to "Deno.connect()" with "connection refused."  Are plain TCP sockets not allowed in "file_server_test.ts?"

sarahdenofiletrav:
I would be interested to know why that worked.

bartlomieju:
> I do not know why the tests are failing; they pass on my machine. It is throwing in a call to "Deno.connect()" with "connection refused." Are plain TCP sockets not allowed in "file_server_test.ts?"

@sarahdenofiletrav we've got some flakiness in CI, seems like you hit an intermittent failure

bartlomieju:
@nayeemrmn please take a look

nayeemrmn (approved):
LGTM

std/http/file_server_test.ts
```diff
@@ -169,6 +242,32 @@ Deno.test("checkPathTraversal", async function (): Promise<void> {
...
+  await startFileServer();
+  try {
+    const res = await fetchExactPath("127.0.0.1", 4507, "../../../..");
+    assertEquals(res.status, 400);
+  } finally {
+    await killFileServer();
+  }
+});
+
+Deno.test("checkPathTraversalAbsoluteURI", async function (): Promise<void> {
```
bartlomieju, line 255: Nice tests

bartlomieju (approved):
LGTM, thank you @sarahdenofiletrav

std/http/file_server_test.ts
sarahdenofiletrav: Thanks!

PR #8479, opened 2020-11-24
feat(std/fs): Re-enable "followSymlinks" option in "walk()" API

### LINKED ISSUE #8076: fs.walk followSymlinks unimplemented
adding the flag `followSymlinks: true` to `fs.walk` results in the error:
```
error: Uncaught Error: unimplemented
                    throw new Error("unimplemented");
                          ^
    at Object.walk (file:///home/andrew/.deno/bin/wallpal.js:2600:27)
```

fs.walk appears to be the only std module that has tjrows an `unimplemented` error. That makes me believe perhaps this was not intentional? https://github.com/denoland/deno/blob/master/std/fs/walk.ts#L110. Either way, it would be awesome to re-enable this flag, there are file systems including symlinks I would like to walk over. Is there a workaround that exists in deno currently?

Liamolucko:
Fixes #8076

### DIFF
--- a/std/fs/walk_test.ts
+++ b/std/fs/walk_test.ts
@@ -252,12 +252,11 @@ testWalk(
   async function symlink(): Promise<void> {
     assertReady(6);
     const files = await walkArray("a");
-    assertEquals(files.length, 2);
+    assertEquals(files.length, 3);
     assert(!files.includes("a/bb/z"));
 
     const arr = await walkArray("a", { followSymlinks: true });
-    assertEquals(arr.length, 3);
+    assertEquals(arr.length, 5);
     assert(arr.some((f): boolean => f.endsWith("/b/z")));
   },
-  true,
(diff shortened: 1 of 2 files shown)


std/fs/walk_test.ts
```diff
@@ -252,12 +252,11 @@ testWalk(
   async function symlink(): Promise<void> {
     assertReady(6);
     const files = await walkArray("a");
-    assertEquals(files.length, 2);
+    assertEquals(files.length, 3);
```
ry, line 255: shouldn't this one remain unchanged?

Liamolucko: I think `includeDirs` was off by default back when this test was written, so now it includes `"a"` itself as well.

ry (approved):
LGTM - thanks @Liamolucko

PR #8480, opened 2020-11-24
feat(cli/version): add canary versioning

crowlKats:


cli/flags.rs
```diff
@@ -309,7 +310,7 @@ fn clap_root<'a, 'b>() -> App<'a, 'b> {
     // Disable clap's auto-detection of terminal width
     .set_term_width(0)
     // Disable each subcommand having its own version.
-    .version(crate::version::DENO)
+    .version(version)
     .long_version(LONG_VERSION.as_str())
```
bartlomieju, line 314: After another thought... maybe add `SHORT_VERSION` lazy static similar to `LONG_VERSION`?

lucacasonato:
@crowlKats Doesn't seem to work:
image
image

bartlomieju:
@crowlKats you need to add this line to `build.rs`:
```rust
println!("cargo:rustc-env=DENO_CANARY={}", env::var("DENO_CANARY").unwrap_or_else(""));
```

cli/build.rs
```diff
@@ -257,6 +257,10 @@ fn main() {
 
   println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
   println!("cargo:rustc-env=PROFILE={}", env::var("PROFILE").unwrap());
+  println!(
+    "cargo:rustc-env=DENO_CANARY={}",
+    env::var("DENO_CANARY").unwrap_or_else("")
+  );
```
lucacasonato, line 263:
```suggestion
  println!(
    "cargo:rustc-env=DENO_CANARY={}",
    env::var("DENO_CANARY").unwrap_or_else(|| "".to_string())
  );
```

crowlKats:
@lucacasonato i was wondering why it said 1.5.1 and was misisng the extra sutff like target tuple afterward... you are calling --version on the deno binary used for the CI itself, not the one that just got built

lucacasonato:
@crowlKats Good catch. I validated that it works:
image

cli/version.rs
```diff
@@ -1,9 +1,15 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
-pub const DENO: &str = env!("CARGO_PKG_VERSION");
 pub const GIT_COMMIT_HASH: &str = env!("GIT_COMMIT_HASH");
 pub const TYPESCRIPT: &str = crate::js::TS_VERSION;
 
+pub fn deno() -> String {
+  let semver = env!("CARGO_PKG_VERSION");
+  option_env!("DENO_CANARY").map_or(semver.to_string(), |_| {
+    format!("{}-canary.{}", semver, GIT_COMMIT_HASH)
```
ry, line 9: the "$VERSION-canary.$HASH" seems overly specific. Why not just "$VERSION-$HASH" ?

ry: Also I guess we can remove the git hash from the `--version` output too since it would be repeated twice.

crowlKats: > Why not just "$VERSION-$HASH" ?

sure

> Also I guess we can remove the git hash from the `--version` output too since it would be repeated twice.

I dont mind, tho then it wont be shown in release versions; is that alright?

ry: > I dont mind, tho then it wont be shown in release versions; is that alright?

fine by me

lucacasonato:
I just realized that a canary for 1.6.0 would have version 1.5.4-hash with the current setup. It should really be 1.6.0-hash. We could fix this by setting `DENO_CANARY` env var to the next version (so 1.5.4 in cargo.toml, and 1.6.0 in DENO_CANARY env var). When updating the version, we would have to change both the cargo.toml, and the _next_ version in the ci.yml file. WDYT?

ry:
> I just realized that a canary for 1.6.0 would have version 1.5.4-hash with the current setup. It should really be 1.6.0-hash. We could fix this by setting DENO_CANARY env var to the next version (so 1.5.4 in cargo.toml, and 1.6.0 in DENO_CANARY env var). When updating the version, we would have to change both the cargo.toml, and the next version in the ci.yml file. WDYT?

This seems like a lot of work. I think 1.5.4-hash is fine for now.

IllusionPerdu:
Since the next versions are not known during the releases, it suffices to specify in the documentation that the canary versions are that of the last release and that it is the hash that allows them to be distinguished.

.github/workflows/ci.yml
```diff
@@ -127,6 +127,15 @@ jobs:
...
           export_default_credentials: true
 
+      - name: Configure canary build
+        if: |
+          matrix.kind == 'test_release' &&
+          github.repository == 'denoland/deno' &&
+          github.ref == 'refs/heads/master'
+        shell: bash
+        run: |
+          echo "DENO_CANARY=true" >> $GITHUB_ENV
```
ry, line 137: Won't this happen on release builds too?

lucacasonato: No - release builds only happen on tags, not on `refs/heads/master`

ry (approved):
LGTM - thank you @crowlKats

Issue #7652, opened 2020-09-24
Coverage output quirks


bartlomieju:
![Screenshot_2020-09-24_at_11 56 08](https://user-images.githubusercontent.com/13602871/94131378-20dda900-fe5e-11ea-8b66-0fc8a5132e32.png)

- comments are marked as not covered

![Screenshot 2020-09-24 at 12 05 13](https://user-images.githubusercontent.com/13602871/94131450-381c9680-fe5e-11ea-99d7-e838f4172cb8.png)

- non-consecutive lines are collapsed together - there should be an empty line between them 

CC @caspervonb

bartlomieju:
Non-consecutive line blocks have been separated in #8484, but comments are still marked as not covered. @caspervonb could you look into that?

caspervonb:
Its because V8 coverage blocks doesn't care about what's on the line, only if they were evaluated or not.

We can just exclude all comment lines.

bartlomieju:
> Its because V8 coverage blocks doesn't care about what's on the line, only if they were evaluated or not.
> 
> We can just exclude all comment lines.

Sounds good to me

PR #8484, opened 2020-11-24
feat(test): separator between disjoint lines for coverage report

SyrupThinker:
Places a newline between non-consecutive lines to improve readability.
Goes towards #7652.


bartlomieju (approved):
@SyrupThinker LGTM, I'm sure we'll iterate more on the output format but this is much more readable compared to current output

PR #8485, opened 2020-11-24
[WIP] feat(cli): upgrade clap to v3

### LINKED ISSUE #8475: Research upgrade to clap v3
Deno uses `clap` as its CLI parser. It looks like clap v3 is behind the corner so I think it's high time to do some research what it would take to upgrade our `cli/flags.rs`.

Keep in mind, [`clap` authors recommended](https://docs.rs/clap/3.0.0-beta.2/clap/) to stay on `v2.33` still as the API is not fully stable. So I'd be happy to have a PR open until stable v3 is released.

crowlKats:
Closes #8475


crowlKats:
blocked by https://github.com/clap-rs/clap/discussions/2224

ry:
Thanks @crowlKats. I like that this upgrade finally gets rid of our ansi_term dependency. (Not so happy it adds clap_generate, clap_derive, heck, proc-macro-error, os_str_bytes, proc-macro-error-attr...)

Should we wait for Clap v3.0.0 to be released? Are there any public facing changes? (AFAICT it looks the same.)

Can we do this upgrade without using clap_generate?

bartlomieju:
> Thanks @crowlKats. I like that this upgrade finally gets rid of our ansi_term dependency. (Not so happy it adds clap_generate, clap_derive, heck, proc-macro-error, os_str_bytes, proc-macro-error-attr...)
> 
> Should we wait for Clap v3.0.0 to be released? Are there any public facing changes? (AFAICT it looks the same.)

I think we should, clap author explicitly state not to upgrade because of more changes being considered for 3.0.

> Can we do this upgrade without using clap_generate?

Nope, it's used to generate shell-completions (for `deno completions` subcommand)

bartlomieju:
Ref https://github.com/denoland/deno/pull/8617

As per Discord conversation; we'll land this PR after stable release of Clap v3. Closing for now.

PR #8495, opened 2020-11-25
refactor(cli): reorganize main.rs and split workers

bartlomieju:
Factored out `init_v8_flags`, `init_logger` and `get_subcommand` from `main` function.

Also `Worker` was removed in favor of moving logic to `MainWorker` and `WebWorker`
respectively.

Towards https://github.com/denoland/deno/pull/8381


cli/main.rs
```diff
@@ -914,15 +913,15 @@ pub fn main() {
...
+      std::process::exit(0);
     }
     DenoSubcommand::Types => {
       let types = get_types(flags.unstable);
       if let Err(e) = write_to_stdout_ignore_sigpipe(types.as_bytes()) {
         eprintln!("{}", e);
         std::process::exit(1);
       }
-      return;
+      std::process::exit(0);
```
ry, line 924: this fine for now, but ideally we would architect it so that `exit` is not called deeper inside the system

cli/web_worker.rs
```diff
@@ -0,0 +1,453 @@
...
+      main_module,
+      program_state,
+      false,
+    );
+    worker
+      .execute("bootstrap.workerRuntime(\"TEST\", false)")
+      .unwrap();
+    worker
+  }
+  #[tokio::test]
```
ry, line 369: nit: NL between functions

cli/main.rs
bartlomieju: Agreed, I'm still looking for proper patterns

ry (approved):
LGTM

PR #8498, opened 2020-11-26
refactor(cli): Remove unnecessary conversion into BTreeMap

magurotuna:
This commit removes unnecessary conversion into BTreeMap. The value
that gets converted into BTreeMap is _originally_ BTreeMap, so this
conversion is just superfluous.

Additionally, a test is added so that we can make sure the keys in the
emitted lockfile are sorted alphabetically.


bartlomieju (approved):
LGTM, thanks

PR #6162, opened 2020-06-07
WebSocket Protocol and Version support

Industrial:
Hi.

These are the headers sent between client and server on the WebSocket: https://tools.ietf.org/html/rfc6455#section-11.3.5

This PR adds support for Protocol and Version headers.

PR #8505, opened 2020-11-26
feat(std/ws): protocol & version support

crowlKats:
Based on #6162


bartlomieju (approved):
LGTM, thanks @crowlKats @Industrial

Issue #6319, opened 2020-06-16
Uncaught BrokenPipe: Broken pipe (os error 32)  


ITS-NIGHTOWL:
After running my ws client for 11 hours, I got this error (for the first time):
```
error: Uncaught BrokenPipe: Broken pipe (os error 32)   
    at unwrapResponse ($deno$/ops/dispatch_minimal.ts:63:11)
    at Object.sendAsyncMinimal ($deno$/ops/dispatch_minimal.ts:106:10)
    at async Object.write ($deno$/ops/io.ts:65:18)
    at async BufWriter.flush (https://deno.land/std@0.54.0/io/bufio.ts:479:25)
    at async writeFrame (https://deno.land/std@0.54.0/ws/mod.ts:144:3)
```
Any reasons as to why?

wongjiahau:
I'm also getting this error, but for different reasons

bartlomieju:
Please provide more reproduction. `Broken pipe` might happen for number of reasons, eg. the other side of connection closes it abruptly.

colorhook:
WebSocket demo reload web page and deno server exit

And I cannot caught the exception

```
async function dispatch(msg: string): Promise<void> {
  for (const client of clients.values()) {
    try {
      await client.send(msg)
    } catch (err) {}
  }
}
``

```
Uncaught BrokenPipe: Broken pipe (os error 32)   
    at unwrapResponse (rt/10_dispatch_minimal.js:52:13)
    at sendAsync (rt/10_dispatch_minimal.js:95:12)
    at async write (rt/12_io.js:117:20)
    at async Object.writeAll (rt/13_buffer.js:223:19)
    at async BufWriter.flush (bufio.ts:467:7)
    at async writeFrame (mod.ts:141:3)
```

deno version

```
deno 1.2.0
v8 8.5.216
typescript 3.9.2
```

Caesar2011:
I get this error when starting a localhost webserver with `for await (const req of server) {` and then open a tab in Firefox from yesterday.

Probably the browser tries to use the previous connection and the server crashes because the connection is unknown.

lmiller1990:
I get this error after around 24h - my server serves HTML at about 30-60 request / h. Basically the "hello world" like setup, no extra dependencies. I am on deno 1.3.x.

erf:
[...]

PR #8506, opened 2020-11-26
fix(std/http): handle all errors from the listener

### LINKED ISSUE #8499: NotConnected: Transport endpoint is not connected (os error 107)
Hey,

some HTTP server errors (e.g., issue https://github.com/denoland/deno/issues/6319) have been fixed by this PR just recently https://github.com/denoland/deno/pull/8365. However, now I get the following error after running an HTTP server behind nginx for a couple of hours on Ubuntu:

```JavaScript
error: Uncaught (in promise) NotConnected: Transport endpoint is not connected (os error 107)
    at processResponse (core.js:223:11)
    at Object.jsonOpAsync (core.js:240:12)
    at async Listener.accept (deno:cli/rt/30_net.js:117:19)
    at async Server.acceptConnAndIterateHttpRequests (server.ts:223:14)
    at async MuxAsyncIterator.callIteratorNext (mux_async_iterator.ts:30:31)
```

@lideming do you have an idea what this could cause?

lideming:
Fixes #8499.
PR #8365 has failed to catch some errors from the listener. Since the error types are uncertain, I have changed it to ignore all errors. This is a workaround for now.


ry:
While I’m sure this is fixing something, adding workarounds like this can obscure and complicate the real problem.

Do you have a way of demonstrating this bug?

kitsonk:
I agree, this is just ignoring a problem versus fixing it.  I believe the root cause is related to #6319, where we have some potential network issues... it should be implemented in the way that the consumer of the API can determine if the error is fatal or not, versus just swallowing errors.

lideming:
I have created another PR #8554

PR #8507, opened 2020-11-26
refactor(cli/tests): replace createResolvable with deferred

crowlKats:
Currently we use `createResolvable` which is defined in multiple places instead of importing `deferred` from std.

There are 2 more instances of a `createResolvable` function being created: `cli/rt/06_util.js` & `core/examples/http_bench_bin_ops.js`. Should i rename those to `deferred` for consistency?


bartlomieju (approved):
LGTM, thanks

bartlomieju:
> Should i rename those to deferred for consistency?

Let's keep them as they are for now

PR #8511, opened 2020-11-26
fix(cli/ops/net): add write permissions for unixpackets datagrams & unix socket

### LINKED ISSUE #7781: `--allow-read` allows writes to Unix sockets
`Deno.connect({ transport: "unix", path: "/path/to/socket" })` only requires `--allow-read=/path/to/socket`, yet allows you to write to the socket. On typical Linux systems, that means `deno run --unstable --allow-read foo.ts` could connect to the DBus socket and potentially run arbitrary commands, depending on what services are running.

Note: `transport: "unix"` is only available with `--unstable` right now.

Here's a [proof-of-concept script](https://gist.github.com/solson/615d89c78b26a431eaf3ef5ed1edc058) that should produce a little notification like this with `deno run --unstable --allow-read dbus-notify.ts` on a typical Linux system with systemd and a user ID of 1000.

image

wperron:
Fixes #7781

### DIFF
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -66,6 +66,44 @@ unitTest(
 );
 
 unitTest(
+  { ignore: Deno.build.os === "windows", perms: { read: true } },
+  function netUnixListenWritePermission(): void {
+    try {
+      const filePath = Deno.makeTempFileSync();
+      const socket = Deno.listen({
+        path: filePath,
+        transport: "unix",
+      });
+      assert(socket.addr.transport === "unix");
+      assertEquals(socket.addr.path, filePath);
+      socket.close();
+    } catch (e) {
+      assert(!!e);
+      assert(e instanceof Deno.errors.PermissionDenied);
+    }
+  },
+);
+
+unitTest(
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
@wperron the fix looks good. Please add a test case in `cli/tests/unit/net_test.ts` that verifies  `PermissionDenied` error is thrown

ry (approved):
LGTM - thank you @wperron

cli/tests/unit/net_test.ts
```diff
@@ -65,6 +65,44 @@ unitTest(
...
+        path: filePath,
+        transport: "unix",
+      });
+      assert(socket.addr.transport === "unix");
+      assertEquals(socket.addr.path, filePath);
+      socket.close();
+    } catch (e) {
+      assert(!!e);
+      assert(e instanceof Deno.errors.PermissionDenied);
+    }
```
bartlomieju, line 83: @wperron this is a bit brittle - if function doesn't throw test will pass anyway. There's a helper assert called `assertThrows` that could be used for this purpose.

wperron: I'll open a new PR for that

Issue #3565, opened 2019-12-30
DenoAsyncWrite must do implicit flush during write()


bartlomieju:
During upgrade to Tokio 0.2 I had to add implicit `flush` during `write()` op.

https://github.com/denoland/deno/blob/46d76a7562025374600a7f866dfc68c1b7e268e9/cli/ops/io.rs#L231-L234

https://github.com/denoland/deno/blob/46d76a7562025374600a7f866dfc68c1b7e268e9/cli/ops/io.rs#L361-L380

I had no problems without `flush` locally, failures occurred during CI runs ([especially on Ubuntu](https://github.com/denoland/deno/runs/364286737)) and appeared in various tests. 

The reasons for why that situation happens are not really known and current solution with `flush` should be treated as a band-aid. 

Eventually we need to figure out what's happening and remove calls to flush (or make it explicit op `flush()`).

Ref #3418

kevinkassimo:
Is this `flush` causing a performance hit to `deno_tcp`, benchmrk of which seems to degrade after Tokio 0.2 upgrade?

bartlomieju:
> Is this `flush` causing a performance hit to `deno_tcp`, benchmrk of which seems to degrade after Tokio 0.2 upgrade?

Most likely

caspervonb:
Been a while, so checked if this is still the case, it is and still cannot reproduce locally.
The problem here seems just be that stdout is buffered on CI so, what if we add flush ops and expose them to Deno.Writer implementations?

bartlomieju:
@piscisaureus PTAL

canonic-epicure:
I've just stumbled upon this issue. My program is using only `Deno.stdout.writeSync()` calls. At the end of it, I'm calling `Deno.exit()` to provide an exit code. And.. some of the output is not printed on the screen! If I remove the `Deno.exit()` call, the output is printed (but then, how do I provide an exit code?)

Can the "flush" call on the `stdout` object be exposed to the user?
Or, can Deno provide the facility for setting a "delayed" exit code? (similar to Node's `process.exitCode = value`, which provides the code to be set, when process exits)

Both would work for me.

Please advise, as a user (programmer) I'm finding this experience quite unpleasant.

bartlomieju:
@canonic-epicure it seems like a separate issue. I never had problem with synchronous writing. Could you open a new issue and provide reproduction code?

piscisaureus:
self ping @piscisaureus

[...]

PR #8512, opened 2020-11-26
refactor(cli): rewrite ops to use ResourceTable2

bartlomieju:


cli/ops/websocket.rs
```diff
@@ -229,33 +233,23 @@ pub async fn op_ws_close(
...
+
+  let resource = state
+    .borrow_mut()
+    .resource_table_2
+    .get::<WsStreamResource>(rid)
+    .ok_or_else(bad_resource_id)?;
+  let mut ws_stream = resource.borrow_mut().await;
+  ws_stream.send(msg).await?;
+  ws_stream.close(None).await?;
+  Ok(json!({}))
```
crowlKats, line 252: either remove `ws_stream.close()`, or make `msg` not `Message::Close`, but just the `Option<CloseFrame>` and call `ws_stream.close()` with `msg` and delete `ws_stream.send()`. `close()` is just a wrapper around send, so what you are doing is closing twice, once with a optional closeframe and once with no close frame (which just defaults internally to a closeframe)

bartlomieju: Can I just remove `ws_stream.close()`?

crowlKats: yes

cli/ops/net.rs
```diff
@@ -379,57 +363,36 @@ fn op_shutdown(
...
+}
 
-  /// Notifies a task when listener is closed so accept future can resolve.
-  pub fn wake_task(&mut self) {
-    if let Some(waker) = self.waker.as_ref() {
-      waker.wake();
-    }
-  }
+struct UdpSocketResource {
+  socket: AsyncRefCell<UdpSocket>,
```
piscisaureus, line 383: This resource needs to be split into a readable and a writable end, just like `TcpStream` in the deno_core benchmarks.
Otherwise it can't send packets while it is also receiving.

piscisaureus: It looks like both send_to() and recv_from() can work on a `&UdpSocket`, so rather than splitting it we could just use borrow() instead of borrow_mut().

bartlomieju:
Notes:
- `NewStreamResource` is not very elegant, it's a structure that has all fields as `Option`s; I think I could harden the implementation a bit but it's still not perfect - `cli/ops/io.rs` need to be aware of all possible kinds of `NewStreamResource`
- some variants of `NewStreamResource` are not yet split into reader/write parts, the reason is two fold; a) tests are green b) some APIs expose strange read/write half (UnixStream), this shouldn't be a problem after Tokio upgrade
- `op_write` still requires to `flush()` after each read, otherwise writes are out of order (#3565)

ry (commented):
I think you ought to just remove the original resource table now - so that we can be sure that nothing depends on it. We can rename "resource_table_2" to "resource_table" in a follow up tho.

bartlomieju:
> I think you ought to just remove the original resource table now - so that we can be sure that nothing depends on it. We can rename "resource_table_2" to "resource_table" in a follow up tho.

Okay, I'll do that

runtime/ops/io.rs
```diff
@@ -137,100 +138,358 @@ pub struct FileMetadata {
...
-    &mut self,
-    cx: &mut Context,
-    buf: &mut [u8],
-  ) -> Poll<Result<usize, AnyError>>;
+  async fn read(self: Rc<Self>, buf: &mut [u8]) -> Result<usize, AnyError> {
+    if self.fs_file.is_some() {
+      debug_assert!(self.child_stdout.is_none());
+      debug_assert!(self.child_stderr.is_none());
+      debug_assert!(self.tcp_stream_read.is_none());
+      debug_assert!(self.tcp_stream_write.is_none());
```
ry, line 259: It's obvious already that this is not ideal, but adding a TODO would help clarify that we want to use an enum in the future.

runtime/ops/io.rs
```diff
@@ -137,100 +138,358 @@ pub struct FileMetadata {
...
+        .0
+        .as_mut()
+        .unwrap()
+        .read(buf)
+        .try_or_cancel(cancel)
+        .await?;
+      return Ok(nwritten);
+    }
+
+    if self.child_stdout.is_some() {
```
ry, line 274: nit: using `else if` here and below would save an instruction or two.

runtime/ops/signal.rs
```diff
@@ -32,7 +40,21 @@ pub fn init(rt: &mut deno_core::JsRuntime) {
...
+#[cfg(unix)]
+impl Resource for SignalStreamResource {
+  fn name(&self) -> Cow<str> {
+    "signal".into()
+  }
+
+  fn close(self: Rc<Self>) {
+    self.cancel.cancel();
+  }
+}
```
ry, line 57: Aside: I think we should factor this all out into core in future work. I don't see the point in doing this for every resource implementation.

bartlomieju: This is tricky... some resources don't have `cancel` field; also if that was part of `deno_core` then we'd need to create a wrapper there. Let's postpone this for now

core/async_cell.rs
```diff
@@ -45,6 +51,15 @@ impl<T: 'static> AsyncRefCell<T> {
   pub fn as_ptr(&self) -> *mut T {
     self.value.get()
   }
+
+  /// Returns value of this cell if there are no outstanding borrows.
+  pub fn try_unwrap(self) -> Result<T, AsyncRefCell<T>> {
+    if self.borrow_count.get().is_empty() {
+      Ok(self.value.into_inner())
+    } else {
```
piscisaureus, line 59: * Is this possible? It seems that since this method consumes self there cannot be any outstanding borrows.
* Canonically this method should be called "into inner" I think.

bartlomieju: Ah good catch, you're right, first we need to call `Rc::into_inner` which gives ownership to the `AsyncRefCell` so checking borrow count is not necessary. I'll rename to `into_inner`

runtime/ops/net.rs
```diff
@@ -138,18 +135,14 @@ async fn receive_udp(
...
-      .get_mut::<UdpSocketResource>(rid)
-      .ok_or_else(|| bad_resource("Socket has been closed"))?;
-    let socket = &mut resource.socket;
-    socket
-      .poll_recv_from(cx, &mut zero_copy)
-      .map_err(AnyError::from)
-  });
-  let (size, remote_addr) = receive_fut.await?;
+  let resource = state
+    .borrow_mut()
```
piscisaureus, line 139: Looks like https://github.com/denoland/deno/pull/8512/files/8fcb9a496bbadc19dc1c22c226ff17656b286bce#diff-6b4e31d40f717632f8b6eb94b2a0828d3cde3ee6df5197d49e976068f9be33b2 hasn't been addressed.
We also need a test for it, which should roughly do the following:

* Create an UDP socket.
* Start receiving packets on that UDP socket.
* Use the UDP socket to send a packet to itself.
* Verify that the packet has been received.

piscisaureus: test:

```ts
const socket = Deno.listenDatagram({ port: 3500, transport: "udp" });
assert(socket.addr.transport === "udp");
assertEquals(socket.addr.port, 3500);
assertEquals(socket.addr.hostname, "127.0.0.1");

const recvPromise = socket.receive();

const sendBuf = new Uint8Array([1, 2, 3]);
const sendLen = await socket.send(sendBuf, socket.addr);
assertEquals(sendLen, 3);

const [recvBuf, recvAddr] = await recvPromise;
assertEquals(recvBuf.length, 3);
assertEquals(1, recvBuf[0]);
assertEquals(2, recvBuf[1]);
assertEquals(3, recvBuf[2]);

socket.close();
```

cli/ops/net.rs
piscisaureus: Continued in https://github.com/denoland/deno/pull/8512#discussion_r543683783.

piscisaureus (approved):

PR #8513, opened 2020-11-26
fix(std/node): Inline default objects to ensure correct prototype

guybedford:
This is a follow-up to https://github.com/denoland/deno/pull/8274 that instead inlines the explicit `default` export objects instead of using the `import * as m; export default m` wrapper technique.

The reason for this is that there are some methods that expect object prototype methods on Node.js builtins, for example:

```js
const fs = require('fs');
if (fs.hasOwnProperty(prop))
```

which fail for the null prototype on module instances.

My hope previously is that it would make it easier to maintain the code, but in the end just keeping the named exports in sync with the default object for the Node.js shims is probably best and cuts down on the total number of files anyway.

As per the previous PR, I've separated the delete, move and edit file changes for easier review so that the last commit contains the substantial parts.


ry (approved):
Sounds reasonable - LGTM - thanks @guybedford

PR #8515, opened 2020-11-27
feat: add mvp language server

### LINKED ISSUE #8400: deno lsp
Tracking issue for `deno lsp` subcommand. 

Given that a lot of code in [`vscode_code`](https://github.com/denoland/vscode_deno) duplicates logic inside `deno` binary (eg. module resolution algorithm, DENO_DIR path hashing) it becomes obvious that deno should provide its own LSP. 

Implementing LSP inside the binary should make maintaining the extension easier, as well as it will allow us to provide the same benefits and experience to other editors.

_This feature is being worked on_

kitsonk:
Resolves #8400 

This is still a not quite ready for merging PR, but it has gotten far enough to get wider review/thoughts input.  Things I know I still need to do:

- [x] Integration Test Infrastructure
- [x] ~lots moar unit tests~ we can tackle this as part of the follow up
- [x] integrate hover
- [x] figure out remote module resolution
- [x] create a list of other features we need to add (#8643)


kitsonk:
I will add more detailed instructions later, but if you want to try it out, you need to build this branch, and have it available in your path as `deno` and use this branch of `vscode_deno`: https://github.com/denoland/vscode_deno/pull/263.  When you have the `vscode_deno` project checked out and installed and built (`npm i && npm run compile && npm i`) open in vscode and go to the debug _Run_ panel and choose the _Launch Client_ option and it will open up a window with the extension enabled.

cli/lsp/capabilities.rs
```diff
@@ -0,0 +1,55 @@
...
+    folding_range_provider: None,
+    rename_provider: None,
+    document_link_provider: None,
+    color_provider: None,
+    execute_command_provider: None,
+    workspace: None,
+    call_hierarchy_provider: None,
+    semantic_tokens_provider: None,
+    on_type_rename_provider: None,
+    experimental: None,
```
RDambrosio016, line 53:
```suggestion
    document_formatting_provider: Some(OneOf::Left(true)),
    ..Default::default()
```
This will save you a ton of code when working with lsp-types 😉

kitsonk: Thank you, but I want to be explicit to help lay out what sort of capabilities we may want to implement.  Sitting there looking up the definition every time makes it arduous.

cli/flags.rs
```diff
@@ -1006,6 +1014,17 @@ Show documentation for runtime built-ins:
     )
 }
 
+fn language_server_subcommand<'a, 'b>() -> App<'a, 'b> {
+  SubCommand::with_name("lsp")
+    .about("Start the language server")
```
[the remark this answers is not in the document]
kitsonk: I think we should debate that. It feels quite unnecessary IMO and means we have to do a lot of silliness in extensions to just get it to work. What is the "harm" in keeping it unflagged?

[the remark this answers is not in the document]
bartlomieju: You're obviously right, especially given that main consumer is `vscode_deno` 🤦

cli/tests/lsp_tests.rs
```diff
@@ -0,0 +1,88 @@
...
+      "did_open_notification.json",
+      "hover_request.json",
+      "shutdown_request.json",
+      "exit_notification.json",
+    ],
+  };
+  let (response, out) = test.run();
+  assert!(response.contains("const Deno.args: string[]"));
+  assert!(out.contains("Connected to \"test-harness\" 1.0.0"));
+}
```
ry, line 88: Nice tests!

cli/lsp/analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+pub fn get_lint_references(
+  specifier: &ModuleSpecifier,
+  media_type: &MediaType,
+  source_code: &str,
+) -> Result<Vec<Reference>, AnyError> {
+  let syntax = ast::get_syntax(media_type);
+  let lint_rules = rules::get_recommended_rules();
+  let mut linter = create_linter(syntax, lint_rules);
+  // TODO(@kitsonk) we should consider caching the swc source file versions for
+  // reuse by other processes
```
bartlomieju, line 52: Do you mean we should use the same AST for multiple purposes? If so, that would require to add new APIs to linter to allow linting already parsed AST. (should be a quick change)

cli/lsp/analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+      Category::Lint(message, code, _) => lsp_types::Diagnostic {
+        range: r.range,
+        severity: Some(lsp_types::DiagnosticSeverity::Warning),
+        code: Some(lsp_types::NumberOrString::String(code)),
+        code_description: None,
+        // TODO(@kitsonk) this won't make sense for every diagnostic
+        source: Some("deno-lint".to_string()),
+        message,
+        related_information: None,
+        tags: None, // we should tag unused code
```
bartlomieju, line 82: I believe this should be handled by linter as part of the `Diagnostic` struct

cli/lsp/analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+pub fn resolve_import(
+  specifier: &str,
+  referrer: &ModuleSpecifier,
+  maybe_import_map: Option<Rc<RefCell<ImportMap>>>,
+) -> ResolvedImport {
+  let maybe_mapped = if let Some(import_map) = maybe_import_map {
+    if let Ok(maybe_specifier) =
+      import_map.borrow().resolve(specifier, referrer.as_str())
```
bartlomieju, line 109: I believe we might have an edge case here - import map shouldn't be applied to "entry point" module:
https://github.com/denoland/deno/blob/ae21a9569b87411f863fa5194c873be69d8bee93/cli/module_loader.rs#L66-L73

cli/lsp/config.rs
```diff
@@ -0,0 +1,49 @@
...
+use deno_core::serde_json::Value;
+
+#[derive(Debug, Clone, Default)]
+pub struct ClientCapabilities {
+  pub status_notification: bool,
+}
+
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Settings {
```
bartlomieju, line 15: Nitpick: `UserSettings`?

cli/lsp/diagnostics.rs
```diff
@@ -0,0 +1,261 @@
...
+              }
+              diagnostics::DiagnosticCategory::Warning => {
+                lsp_types::DiagnosticSeverity::Warning
+              }
+              diagnostics::DiagnosticCategory::Suggestion => {
+                lsp_types::DiagnosticSeverity::Hint
+              }
+              diagnostics::DiagnosticCategory::Message => {
+                lsp_types::DiagnosticSeverity::Information
+              }
```
bartlomieju, line 189: Consider implementing `From<diagnostics::DiagnosticCategory> for lst_types::DiagnosticSeverity`

cli/lsp/diagnostics.rs
```diff
@@ -0,0 +1,261 @@
...
+
+fn to_lsp_range(
+  start: &diagnostics::Position,
+  end: &diagnostics::Position,
+) -> lsp_types::Range {
+  lsp_types::Range {
+    start: lsp_types::Position {
+      line: start.line as u32,
+      character: start.character as u32,
+    },
```
bartlomieju, line 118: Consider implementing `From<diagnostics::Position> for lsp_types::Range`

cli/lsp/analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+  category: Category,
+  range: Range,
+}
+
+fn as_lsp_range(range: &deno_lint::diagnostic::Range) -> Range {
+  Range {
+    start: Position {
+      line: (range.start.line - 1) as u32,
+      character: range.start.col as u32,
+    },
```
bartlomieju, line 35: Consider implementing `From<diagnostics::Range> for lsp_types::Range`

cli/lsp/handlers.rs
```diff
@@ -0,0 +1,215 @@
...
+    file_path
+  } else {
+    PathBuf::from(params.text_document.uri.path())
+  };
+  let config = dprint::configuration::ConfigurationBuilder::new()
+    .deno()
+    .build();
+
+  let new_text = dprint::format_text(&file_path, &file_text, &config)
+    .map_err(|e| custom_error("FormatError", e))?;
```
bartlomieju, line 45: Maybe abstract into a helper function in `tools/fmt.rs`? I'm a bit worried these calls might get out of sync with `tools/fmt.rs` (probably unlikely)

cli/lsp/mod.rs
```diff
@@ -0,0 +1,432 @@
...
+use std::time::Instant;
+
+#[derive(Debug)]
+struct ServerError {
+  code: i32,
+  message: String,
+}
+
+impl ServerError {
+  #[allow(unused)]
```
bartlomieju, line 60: Is this suppression intentional?

cli/lsp/mod.rs
```diff
@@ -0,0 +1,432 @@
...
+    if self.process_changes() {
+      debug!("process changes");
+      let state = self.snapshot();
+      self.tasks.handle.spawn(move || {
+        let diagnostics = diagnostics::generate_linting_diagnostics(&state);
+        Task::Diagnostics((DiagnosticSource::Lint, diagnostics))
+      });
+      // TODO(@kitsonk) isolates do not have Send to be safely sent between
+      // threads, so I am not sure this is the best way to handle queuing up of
+      // getting the diagnostics from the isolate.
```
bartlomieju, line 203: I'm not entirely sure what's this comment is referring to, could you elaborate?

cli/lsp/sources.rs
```diff
@@ -0,0 +1,382 @@
...
+          return Some(metadata);
+        }
+      }
+    }
+    if let Some(version) = self.get_script_version(specifier) {
+      if let Some(path) = self.get_path(specifier) {
+        if let Ok(bytes) = fs::read(path) {
+          if specifier.as_url().scheme() == "file" {
+            let charset = text_encoding::detect_charset(&bytes).to_string();
+            if let Ok(source) = get_source_from_bytes(bytes, Some(charset)) {
```
bartlomieju, line 101: Using `?` instead of `if let Some()` should make this function a lot less indented.

cli/lsp/sources.rs
```diff
@@ -0,0 +1,382 @@
...
+
+  pub fn get_media_type(
+    &mut self,
+    specifier: &ModuleSpecifier,
+  ) -> Option<MediaType> {
+    if let Some(specifier) = self.resolve_specifier(specifier) {
+      if let Some(metadata) = self.get_metadata(&specifier) {
+        return Some(metadata.media_type);
+      }
+    }
```
bartlomieju, line 84: Ditto

cli/lsp/state.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+}
+
+pub struct Handle<H, C> {
+  pub handle: H,
+  pub receiver: C,
+}
+
+#[allow(unused)]
```
bartlomieju, line 87: Remove?

cli/lsp/task_pool.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
+
+use crossbeam_channel::Sender;
+use threadpool::ThreadPool;
+
+// TODO(@kitsonk) replace with `tokio-threadpool`?
```
bartlomieju, line 6: Can we use regular tokio runtime for this purpose?

cli/lsp/text.rs
```diff
@@ -0,0 +1,514 @@
...
+use dissimilar::diff;
+use dissimilar::Chunk;
+use lsp_types::TextEdit;
+use std::ops::Bound;
+use std::ops::Range;
+use std::ops::RangeBounds;
+
+// TODO(@kitson) in general all of these text handling routines don't handle
+// JavaScript encoding in the same way and likely cause issues when trying to
+// arbitrate between chars and Unicode graphemes.  There be dragons.
```
bartlomieju, line 14: Maybe we could leverage SWC's `SourceMap` for these purposes?

cli/lsp/tsc.rs
```diff
@@ -0,0 +1,1061 @@
...
+    text.to_string()
+  } else {
+    format!("```\n{}\n```", text)
+  }
+}
+
+/// Replace JSDoc like links (`{@link http://example.com}`) with markdown links
+fn replace_links(text: &str) -> String {
+  let jsdoc_links_regex = Regex::new(r"(?i)\{@(link|linkplain|linkcode) (https?://[^ |}]+?)(?:[| ]([^{}\n]+?))?\}").unwrap();
+  jsdoc_links_regex
```
bartlomieju, line 127: Maybe in the future we could use  https://swc.rs/rustdoc/jsdoc/ for this purpose?

cli/lsp/tsc.rs
```diff
@@ -0,0 +1,1061 @@
...
+    }
+  }
+
+  Ok(json!(None::<String>))
+}
+
+/// Create and setup a JsRuntime based on a snapshot. It is expected that the
+/// supplied snapshot is an isolate that contains the TypeScript language
+/// server.
+pub fn start(snapshot: Snapshot, debug: bool) -> Result<JsRuntime, AnyError> {
```
bartlomieju, line 647: So it looks like we'll have 2 different implementations of TSC, do they share any code?

bartlomieju: Also it seems to me `snapshot` argument doesn't make much sense, are we ever gonna start this `JsRuntime` with other type of snapshot?

bartlomieju: Answering myself to the first question, the code is shared in `99_main_compiler.js` it's just a different set of ops that's used.

ry:
New version of vscode-deno has been published at https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno-canary

cli/Cargo.toml
```diff
@@ -32,54 +32,59 @@ winres = "0.1.11"
...
 shell-escape = "0.1.5"
-sys-info = "0.7.0"
 sourcemap = "6.0.1"
 swc_bundler = "0.17.5"
 swc_common = { version = "0.10.6", features = ["sourcemap"] }
 swc_ecmascript = { version = "0.14.4", features = ["codegen", "dep_graph", "parser", "react", "transforms", "visit"] }
+sys-info = "0.7.0"
 tempfile = "3.1.0"
 termcolor = "1.1.0"
+threadpool = "1.8.1"
```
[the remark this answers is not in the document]
ry: I've removed it.... 

@kitsonk I hope I didn't mess something up - feel free to revert my commit if it's not working properly. In any case it should show you how to use the threadpool.

cli/lsp/task_pool.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
+
+use crossbeam_channel::Sender;
+
+pub struct TaskPool<T> {
+  sender: Sender<T>,
+}
```
ry, line 7: Now that this isn't using threadpool, this struct seems not so useful. I suggest placing `sender: Sender<Task>` directly into `ServerState`. I think the `Handle` abstraction can also be removed.

cli/lsp/text.rs
```diff
@@ -0,0 +1,514 @@
...
+  }
+
+  #[test]
+  fn test_slice() {
+    assert_eq!(slice("Deno", 1..3), "en");
+    assert_eq!(slice("Deno", 1..=3), "eno");
+    assert_eq!(slice("Deno Land", 1..), "eno Land");
+    assert_eq!(slice("Deno", ..3), "Den");
+  }
+}
```
ry, line 514: Nicely tested.

cli/lsp/tsc.rs
```diff
@@ -0,0 +1,1061 @@
...
+    Some(
+      parts
+        .into_iter()
+        .map(|p| p.text)
+        .collect::<Vec<String>>()
+        .join(""),
+    )
+  } else {
+    None
+  }
```
ry, line 41: Nit: instead of this `if let ... else` you can use `Option::map`

```rust
maybe_parts.map(|parts| {
  parts.into_iter().map(|p| p.text).collect::<Vec<String>>().join("")
})

```

cli/lsp/tsc.rs
```diff
@@ -0,0 +1,1061 @@
...
+            .replace(text, |c: &Captures| format!("{} {}", &c[1], &c[2]))
+            .to_string(),
+        )
+      }
+      "default" => Some(make_codeblock(text)),
+      _ => Some(replace_links(text)),
+    }
+  } else {
+    None
+  }
```
ry, line 76: ditto

cli/lsp/diagnostics.rs
```diff
@@ -0,0 +1,261 @@
...
+      ts_diagnostics.append(&mut ts_json_to_diagnostics(tsc::request(
+        runtime,
+        state,
+        request_syntactic_diagnostics,
+      )?)?);
+      if !ts_diagnostics.is_empty() {
+        diagnostics.push((file_id, version, ts_diagnostics));
+      } else {
+        diagnostics.push((file_id, version, Vec::new()));
+      }
```
lucacasonato, line 256: Why is this conditional necessary?

ry:
I'm trying to use the published "Deno (Canary)" extension and seeing double messages. I have the original deno extension disabled. Any idea what's happening here?

Screen Shot 2020-12-04 at 11 20 47 AM

cli/lsp/diagnostics.rs
kitsonk: Because you have to send an empty array to potentially clear down any previous diagnostics that might have cleared. I realise that isn't obvious and will add a comment.

kitsonk:
> I'm trying to use the published "Deno (Canary)" extension and seeing double messages. I have the original deno extension disabled. Any idea what's happening here?
> 
> Screen Shot 2020-12-04 at 11 20 47 AM

Hmmm... that is strange. You can see the source is both `ts` and `deno-lint`, it would seem

cli/lsp/diagnostics.rs
ry: I think his point is that if `ts_diagnostics` is empty, it's the same as `Vec::new()`

cli/lsp/analysis.rs
```diff
@@ -0,0 +1,313 @@
...
+use deno_core::ModuleSpecifier;
+use deno_lint::rules;
+use lsp_types::Position;
+use lsp_types::Range;
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+pub enum Category {
+  Lint(String, String, Option<String>),
```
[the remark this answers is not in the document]
kitsonk: Using structs in enums in variants requires creating another struct, which really defeats the usefulness of having a terse enum.  I would rather just document it better.

cli/lsp/analysis.rs
kitsonk: lint/format/doc/whatever...  Should break it out as a seperate issue: #8641

cli/lsp/analysis.rs
kitsonk: Not sure what you specifically mean here.

cli/lsp/analysis.rs
kitsonk: There are no "entry point" modules in the language server, so we have to apply it to all modules, because we don't know what module will actually become the entry point.

cli/lsp/analysis.rs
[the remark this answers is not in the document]
bartlomieju: @kitsonk you can do:
```rust
pub enum Category {
   Lint {
     field1: String,
     field2: String,
     field3: Option<string>
   }
}
```

cli/lsp/analysis.rs
bartlomieju: Poor wording on my side; what I meant that we should add appropriate tag in `deno_lint` `Diagnostic` to mark unreachable code

cli/lsp/config.rs
kitsonk: That is confusing in my opinion... as in vscode "user" settings are the global settings...  maybe `WorkspaceSettings` is more clear, which is what I have changed it to.

cli/lsp/analysis.rs
kitsonk: Can't as neither deno_lint and lsp_types are defined in the current crate... so you can't create trait implementations for them.

cli/lsp/handlers.rs
kitsonk: I wanted to avoid/limit changes outside of lsp at this point...  Something I think we should address down the road.  I have added a todo.

cli/lsp/mod.rs
kitsonk: `ServerError` was an abstraction that I had adopted from rust_analyzer that I never used, so I have removed it fully now.

cli/lsp/mod.rs
kitsonk: The linting diagnostics can be dealt with asynchronously, in a totally seperate thread (you can see just above this they are spawned), but because isolates can't be sent to other threads (I don't know if it is just they lack `Send` or if there is a valid reason) we have to deal with generating the diagnostics for TypeScript in the same thread that is handling the listening and the main loop...

cli/lsp/sources.rs
kitsonk: No, because I can't/don't want to return a result.  These methods can't error.

cli/lsp/sources.rs
kitsonk: Ditto

cli/lsp/text.rs
kitsonk: I will add this to the list of post initial stuff to deal with.

cli/lsp/mod.rs
bartlomieju: Okay I get it. 

> I don't know if it is just they lack Send or if there is a valid reason

Both, we've pinned isolated to specific thread by the means of type system so we don't have to deal with locking. 

I guess in the future we could have a long lived thread for tsc as well?

cli/lsp/sources.rs
bartlomieju: Okay, but you can use `?` with options as well - encountering `None` will short-circuit just like in case of `Err`

cli/lsp/handlers.rs
bartlomieju: Of course, my bad 🤦

cli/lsp/tsc.rs
kitsonk: 🤷  I suspect it won't return things exactly the way vscode is interpreting here, but worth looking into... will add it to the list...

cli/lsp/tsc.rs
kitsonk: cleaned up

cli/lsp/diagnostics.rs
kitsonk: 🤦

cli/lsp/task_pool.rs
kitsonk: I removed the module and the `Handle` abstraction.

cli/lsp/analysis.rs
kitsonk: I will add it to the list...

cli/lsp/mod.rs
kitsonk: I can see we normally wouldn't want or need it, but in the lsp, it might be worth having a dedicated thread... I will add it to the list.

cli/lsp/analysis.rs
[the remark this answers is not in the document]
kitsonk: 🤯  TIL... I didn't think Rust had that...

cli/lsp/sources.rs
kitsonk: 🤯  I was using the `?` operator without understanding what it actually did... 😭   I cleaned up where it was obvious.

bartlomieju (approved):
LGTM 🚀  massive feature

ry (approved):
LGTM!

Issue #8454, opened 2020-11-21
Add support for moving cursor with ctrl key in REPL


TheWizardBear:
Currently, you can move the cursor left and right with the arrow keys, but it would be nice if using `ctrl` was also supported, so that when `ctrl`+`left` is pressed it will jump to the begging of a block of characters (and to the end with `ctrl`+`right`) and when `ctrl`+`delete` is pressed all of the text to the right of the cursor that is in the same block of characters will be deleted (and the left side with `ctrl`+`backspace`).

For comparison, this is how the node REPL behaves:
If you have the following line (where the cursor is at the very left):
`let foo = "hello world";`
pressing `ctrl`+`right` will first move the cursor to the `f`, then to the `=`, then the `h`, then the `w`, then the `"`, and finally the end of the line.

Of course, it doesn't have to work exactly the same way, but it would be nice to be able to do something like this, as currently `ctrl`+`left`, `ctrl`+`right` and `ctrl`+`delete` don't do anything at all and `ctrl`+`backspace` behaves the same as `backspace`.

bartlomieju:
FYI currently REPL supports Emacs key bindings.

CC @caspervonb

caspervonb:
Yeah we're using the default readline bindings at the moment, in other words emacs bindings.
You move between words with alt, seems that node has the same behavior, I get the same behavior in both deno and node here and can't spot any defects in the bindings.

hayd:
On the update to 7.0.0 ctrl+left and ctrl+right are supported https://github.com/kkawakam/rustyline/issues/146

(I agree that it'd be nice to add ctrl+delete as well... perhaps worth adding an issue on rustyline?)

caspervonb:
Will bump rustyline dep :)

TheWizardBear:
Also `shift`+`left`/`right` (and `ctrl`+`shift`+`left`/`right`) to select text would be nice too, although I don't know if this would be possible.

caspervonb:
Defects where the mappings don't match say, bash should probably be brought up in the upstream rustyline repository as preferably we won't have anything custom, at-least not for the basic editor.

hayd:
For the ctrl+delete issue, i think it makes sense to ask for this on rustyline, IMO deno shouldn't be adding in/deciding on a bunch of random shortcuts.... but if they are in rustyline then deno gets them for free :) (and so do all of rustyline's users)

PR #8516, opened 2020-11-27
chore: update rustyline to 7.0.0

caspervonb:
Updates rustyline to 7.0.0 which introduces some new actions (key bindings), resolves, or at-least partially resolves https://github.com/denoland/deno/issues/8454


caspervonb:
Had to bump a couple of transitives as-well.

Cargo.lock
```diff
@@ -3235,6 +3327,30 @@ version = "0.2.68"
...
+
+[[package]]
+name = "wasm-bindgen-test-macro"
+version = "0.3.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8fb9c67be7439ee8ab1b7db502a49c05e51e2835b66796c705134d9b8e1a585"
+dependencies = [
+ "proc-macro2 1.0.24",
+ "quote 1.0.7",
+]
```
bartlomieju, line 3352: Where do these deps come from?

Cargo.lock
```diff
@@ -3264,6 +3380,15 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "webpki-roots"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0f20dea7535251981a9670857150d571846545088359b28e4951d350bdaf179f"
+dependencies = [
+ "webpki",
+]
```
bartlomieju, line 3390: Ditto, it's surprising this is being updated

Cargo.lock
caspervonb: Quick grep of the Cargo.lock file, they're from reqwest

Cargo.lock
```diff
@@ -2037,17 +2095,18 @@ dependencies = [
...
  "serde",
- "serde_urlencoded",
- "tokio 0.2.22",
+ "serde_urlencoded 0.7.0",
+ "tokio 0.2.23",
  "tokio-rustls",
  "url",
  "wasm-bindgen",
  "wasm-bindgen-futures",
+ "wasm-bindgen-test",
```
caspervonb, line 2107: Here.

Cargo.lock
magurotuna: Maybe that's related to https://github.com/seanmonstar/reqwest/pull/1093

bartlomieju: It's strange that `reqwest` gets upgraded anyway. Is that intentional?

caspervonb: Not really, just merged with master and ran cargo update 🤔

bartlomieju: > Not really, just merged with master and ran cargo update 🤔

Bad idea 😅  I suggest upgrading manually only crates you're interested in. @caspervonb could you revert non-rustyline related crate updates?

Cargo.lock
caspervonb: Thanks @bartlomieju 🙃

bartlomieju (approved):
LGTM

PR #8518, opened 2020-11-27
build: run cargo tests on single thread

bartlomieju:


ry:
What was your take away?

bartlomieju:
That 60 minute timeout might be too small to run on a single thread... Let's see if master is still flaky after recent fixes

PR #8520, opened 2020-11-27
fix test_raw_tty hang

### LINKED ISSUE #4262: test_raw_tty hangs

ry:
On OSX and linux I can reproduce the hang reliably by running:

    while true; do cargo test -p deno  tty -- --nocapture; done

Sometimes you have to wait a bit, but it shows up eventually.

I never see the hang when running

    while true; do cargo test -p deno test_raw_tty  -- --nocapture; done

Therefore I think its related to fork / test runner.

Edit: working now.

Closes #4262

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -448,6 +448,7 @@ dependencies = [
  "dprint-plugin-typescript",
  "encoding_rs",
  "env_logger",
+ "exec",
  "filetime",
  "fwdansi",
  "http",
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM, should we use `exec` in other test relying on tty?

ry:
@bartlomieju I think so, I will investigate.

PR #8521, opened 2020-11-27
fix(watcher): watcher doesn't exit when module resolution fails

### LINKED ISSUE #7848: deno run --watch exits when file has invalid syntax

magurotuna:
This commit makes the file watcher continue to work even if module
resolution fails at the initial attempt, allowing us to execute `run`
or `bundle` subcommand when a script has invalid syntax. In such
cases, the watcher observes a single file that is specified as an
command line argument.

Fixes #7848


cli/tests/integration_tests.rs
```diff
@@ -1294,7 +1296,11 @@ fn bundle_watch_fail() {
   std::thread::sleep(std::time::Duration::from_secs(1));
   assert!(stderr_lines.next().unwrap().contains("file_to_watch.js"));
   assert!(stderr_lines.next().unwrap().contains("error:"));
-  assert!(!deno.wait().unwrap().success());
+  assert!(stderr_lines.next().unwrap().contains("Bundle failed!"));
+  // the target file hasn't been created yet
+  assert!(!target_file.is_file());
+  // Make sure the Deno's process is still alive
+  assert!(deno.try_wait().unwrap().is_none());
```
bartlomieju, line 1304: Nice! Could you create a file and check that watcher actually restarts as expected?

magurotuna: Sure, done it 👍

bartlomieju (approved):
I've tested this PR locally, works great, very good fix @magurotuna, LGTM

PR #8523, opened 2020-11-27
feat(core): Add FsModuleLoader that supports loading from filesystem

magurotuna:
This commit adds `FsModuleLoader` to `deno_core`, which implements `ModuleLoader` trait.
It is used when creating a runtime that supports module loading from filesystem. In particular, it's going to be used by `deno_lint` plugin feature. 

Ref: 
- https://github.com/denoland/deno_lint/pull/415
- https://github.com/denoland/deno/pull/8381

CC @bartlomieju


core/modules.rs
```diff
@@ -130,6 +130,43 @@ impl ModuleLoader for NoopModuleLoader {
...
+    _op_state: Rc<RefCell<OpState>>,
+    module_specifier: &ModuleSpecifier,
+    _maybe_referrer: Option<ModuleSpecifier>,
+    _is_dynamic: bool,
+  ) -> Pin<Box<ModuleSourceFuture>> {
+    let module_specifier = module_specifier.clone();
+    async move {
+      let path = module_specifier.as_url().to_file_path().unwrap();
+      let content = std::fs::read_to_string(path)?;
+      let module = ModuleSource {
```
rheh, line 159: Could these three let be const?

core/modules.rs
```diff
@@ -130,6 +130,43 @@ impl ModuleLoader for NoopModuleLoader {
...
+
+  fn load(
+    &self,
+    _op_state: Rc<RefCell<OpState>>,
+    module_specifier: &ModuleSpecifier,
+    _maybe_referrer: Option<ModuleSpecifier>,
+    _is_dynamic: bool,
+  ) -> Pin<Box<ModuleSourceFuture>> {
+    let module_specifier = module_specifier.clone();
+    async move {
```
rheh, line 156: why async?  I see no await within?

core/modules.rs
```diff
@@ -130,6 +130,43 @@ impl ModuleLoader for NoopModuleLoader {
...
+      let path = module_specifier.as_url().to_file_path().unwrap();
+      let content = std::fs::read_to_string(path)?;
```
rheh, line 158: Renamed to content -> code?

core/modules.rs
```diff
@@ -130,6 +130,43 @@ impl ModuleLoader for NoopModuleLoader {
...
+      let module = ModuleSource {
+        code: content,
```
rheh, line 160: If we renamed to code, we could just:
`{
code,
...
}`

core/modules.rs
bartlomieju: The trait requires to return future from this method.

core/modules.rs
bartlomieju: No, Rust doesn't allow that

bartlomieju:
@magurotuna I applied some changes to error handling and doc comment. I'll wait on @ry's review before landing

ry (approved):
LGTM

PR #8524, opened 2020-11-27
docs(import_maps): Add trailing slash to ./src example

eezing:
Update docs to reflect current behavior. If trailing slash is not provided, `deno run` will throw error: `Package address targets must end with "/"`.


ry (approved):
Thx

PR #8526, opened 2020-11-27
chore(tests): fix brittle network permission test

wperron:
Fixes a brittle test that was identified in #8511 after it was merged.

cc @bartlomieju


bartlomieju (approved):
LGTM, thanks for quick fix

Issue #8529, opened 2020-11-28
Parse errors from SWC could be better


bartlomieju:
Currently if SWC parser encounters error its output is pretty minimal:

```ts
import b, { c } from "./b.js";

console.log(b);
console.log(c
```

```
error: Unexpected eof at file:///Users/biwanczuk/dev/deno/a.js:4:13
```

It'd be great if produced error could contain snippet of code pointing to problematic bit of code, similar to errors from TS and V8:

```ts
import b, { c } from "./b.js";

foo
console.log(b);
console.log(c);
```

```
error: Uncaught ReferenceError: foo is not defined
foo
^
    at file:///Users/biwanczuk/dev/deno/a.js:3:1
```

CC @kdy1

kdy1:
Did you provide `Lrc<SourceMap>` to the `Handler` from `swc_common`?
If then, it should point the error just like rustc.

There are some error reporting tests. 
See: https://github.com/swc-project/swc/tree/8ca3d1160db804322724a6e30181e9c86bdd396a/ecmascript/parser/tests/typescript-errors

bartlomieju:
@kdy1 doesn't look like it, here's our handler:
https://github.com/denoland/deno/blob/d9b41828684a5e1ed620d7324ad6c27a83687622/cli/ast.rs#L381-L388

Unfortunately I can't find API that would allow to pass `Lrc<SourceMap>` and not emit to tty.

kdy1:
Oh, you are correct. It accepts `Lrc<SourceMap>` only if `tty-emitter` is enabled.

I'll see if I can make a handler with cm but without tty.

bartlomieju:
> Oh, you are correct. It accepts `Lrc<SourceMap>` only if `tty-emitter` is enabled.
> 
> I'll see if I can make a handler with cm but without tty.

Thanks @kdy1!

kdy1:
You can now update `swc_common` and use [`EmitterWriter`](https://docs.rs/swc_common/0.10.6/swc_common/errors/emitter/struct.EmitterWriter.html)

bartlomieju:
@kdy1 thank you, that was very fast turnaround!

bartlomieju:
Reference implementation: https://github.com/swc-project/swc/blob/46b553ecc3efd3ddcaff6531535b27526dc16781/testing/src/string_errors.rs#L11-L30

kdy1:
Does it work?
I expect it to work correctly as the parser test successed, but just to be sure.

PR #8534, opened 2020-11-28
[WIP] refactor(cli): improve SWC parse error

bartlomieju:
Closes https://github.com/denoland/deno/issues/8529

Blocked by upgrade of `deno_doc`, `deno_lint` and `dprint_plugin_typescript` (need to update `swc_common` dep in those crates)

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -16,16 +16,16 @@ use swc_common::chain;
 use swc_common::comments::Comment;
 use swc_common::comments::CommentKind;
 use swc_common::comments::SingleThreadedComments;
-use swc_common::errors::Diagnostic;
-use swc_common::errors::DiagnosticBuilder;
-use swc_common::errors::Emitter;
+// use swc_common::errors::Diagnostic;
+// use swc_common::errors::DiagnosticBuilder;
+// use swc_common::errors::Emitter;
+use std::io::Write;
 use swc_common::errors::EmitterWriter;
 use swc_common::errors::Handler;
 use swc_common::errors::HandlerFlags;
 use swc_common::FileName;
-use std::io::Write;
 use swc_common::Globals;
-use swc_common::Loc;
+// use swc_common::Loc;
 use swc_common::SourceFile;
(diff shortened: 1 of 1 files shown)


bartlomieju:
@kdy1 I've managed to rewrite handler to use single `ErrorBuffer`, unfortunately the output is not so great:
```
error: error: Unexpected token `}`. Expected an identifier, void, yield, null, await, break, a string literal, a numeric literal, true, false, `, -, import, this, typeof, {, [, (
Error:  --> <file:///home/runner/work/deno/deno/cli/tests/swc_syntax_error.ts>:4:1
  |
4 | }
  | ^



```
Notice "error" being repeated 3 times (the first one comes from Deno, as it's terminal error, the remaining two seem to come from SWC) as well as multiple newlines after the error. I'm not sure if we should land it after all, @kitsonk thoughts?

kdy1:
@bartlomieju  I'm not sure if `Error` comes from swc.
I looked at the error reporting test of swc, but https://github.com/swc-project/swc/blob/faa1c5f4e5d956bc062542b42d123d42b31776eb/ecmascript/parser/tests/typescript-errors/arrow-function/async-rest-optional-parameter/input.ts.stderr does not contain `Error:`

bartlomieju:
@kdy1 thanks, I will dig deeper then...

kdy1:
I'll see if enabling `tty-emitter` removes the `Error`. If then, it should be bug of `swc_common`.

kdy1:
I tried `cargo test swc` (without `tty-emitter`) after doing `gh pr checkout 8534`, and I got

```
error: error: Unexpected token `}`. Expected an identifier, void, yield, null, await, break, a string literal, a numeric literal, true, false, `, -, import, this, typeof, {, [, (
 --> <file:///Users/kdy1/projects/deno/cli/tests/swc_syntax_error.ts>:4:1
  |
4 | }
  | ^

```

I got this using mac os.

It seems like looking for `cfg(target_os = "linux")` or usage of threads in deno test would be the key.

PR #8535, opened 2020-11-28
build: publish only "deno" crate on tags

bartlomieju:
This commit updates CI script to publish only "deno"
crate on tags.

Following crates are not automatically published anymore:
- deno_core
- deno_web
- deno_fetch
- deno_crypto

Before this commit creating a new release required to bump
version on all above crates even though in practice they
rarely change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -274,18 +274,5 @@ jobs:
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         run: |
-          cd core
-          cargo publish
-          sleep 30
-          cd ../op_crates/web
-          cargo publish
-          sleep 30
-          cd ../fetch
-          cargo publish
-          sleep 30
-          cd ../crypto
-          cargo publish
-          sleep 30
-          cd ../../cli
-          sleep 30
+          cd cli
           cargo publish


ry (approved):
LGTM

lucacasonato:
I don't think this will work. `deno` crate requires the other crates (like `deno_core` crate). If we don't publish a new `deno_core` version, cargo will instead pull the old version (from the previous release), breaking the build. This is not an issue if between releases the crates have _no_ changes, but any change (even minor) will require us to manually bump and release the non `deno` crates. This is theoretically manageable, but seems pretty annoying.

bartlomieju:
> I don't think this will work. `deno` crate requires the other crates (like `deno_core` crate). If we don't publish a new `deno_core` version, cargo will instead pull the old version (from the previous release), breaking the build. This is not an issue if between releases the crates have _no_ changes, but any change (even minor) will require us to manually bump and release the non `deno` crates. This is theoretically manageable, but seems pretty annoying.

@lucacasonato correct, but now the opposite is true, we're bumping version every week and some crates haven't seen a change for a few weeks.

lucacasonato:
Still seems like its going to be very prone to us forgetting to bump one of the op crates, or core :upside_down_face:. I guess well see how it goes.

hayd:
You could fetch the latest cargo version then git diff the directory against that tag (and publish if different).

PR #8537, opened 2020-11-28
fix(cli/version): canary change pre-release to metadata

crowlKats:
Currently canary versioning is described using a `-` between semver & hash. That's incorrect as `-` is for pre releases whereas `+` is for build metadata (as described on the semver spec), which would be more appropriate for canaries.


ry (approved):
LGTM

PR #8539, opened 2020-11-29
feat: deno compile

### LINKED ISSUE #986: "deno compile" into executable
I was wondering if there's any plan to support compiling deno scripts into a single executable? 

I saw "Single executable" as a feature, but the code that follows seems to suggest it's deno as the single executable, not the programs executed with it.

It'd be awesome to support something like `go build main.go` or something like `zeit/pkg`.

lucacasonato:
Closes #986 

This implements the `deno compile` subcommand. It works be generating a new standalone binary, by appending a bundle of the source file and magic trailer from the current binary. If we find a magic trailer at the end of the the binary, we try to load the bundle that is embedded in the binary, and run it.

```
~/p/g/d/deno ❯❯❯ ./target/debug/deno compile https://deno.land/std/http/file_server.ts file_server
Bundle https://deno.land/std/http/file_server.ts
Check https://deno.land/std/http/file_server.ts
Compile https://deno.land/std/http/file_server.ts
~/p/g/d/deno ❯❯❯ ./file_server 
HTTP server listening on http://0.0.0.0:4507/
```

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -22,6 +22,10 @@ pub enum DenoSubcommand {
     source_file: String,
     out_file: Option<PathBuf>,
   },
+  Compile {
+    source_file: String,
+    out_file: String,
+  },
   Completions {
     buf: Box<[u8]>,
   },
@@ -291,6 +295,8 @@ pub fn flags_from_vec_safe(args: Vec<String>) -> clap::Result<Flags> {
     doc_parse(&mut flags, m);
   } else if let Some(m) = matches.subcommand_matches("lint") {
     lint_parse(&mut flags, m);
+  } else if let Some(m) = matches.subcommand_matches("compile") {
+    compile_parse(&mut flags, m);
   } else {
     repl_parse(&mut flags, &matches);
   }
@@ -340,6 +346,7 @@ If the flag is set, restrict these messages to errors.",
     )
     .subcommand(bundle_subcommand())
     .subcommand(cache_subcommand())
+    .subcommand(compile_subcommand())
     .subcommand(completions_subcommand())
     .subcommand(doc_subcommand())
     .subcommand(eval_subcommand())
@@ -404,6 +411,18 @@ fn install_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
     args,
     root,
     force,
+  };
+}
+
+fn compile_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
+  compile_args_parse(flags, matches);
+
+  let source_file = matches.value_of("source_file").unwrap().to_string();
+  let out_file = matches.value_of("out_file").unwrap().to_string();
+
+  flags.subcommand = DenoSubcommand::Compile {
+    source_file,
+    out_file,
   };
 }
 
@@ -787,6 +806,22 @@ The installation root is determined, in order of precedence:
   - $HOME/.deno
 
 These must be added to the path manually if required.")
+}
+
+fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
+  compile_args(SubCommand::with_name("compile"))
+    .arg(
+      Arg::with_name("source_file")
+        .takes_value(true)
+        .required(true),
+    )
+    .arg(Arg::with_name("out_file").takes_value(true).required(true))
+    .about("Compile the script into a self contained executable")
+    .long_about(
+      "Compiles the given script into a self contained executable.
+  deno compile https://deno.land/std/http/file_server.ts file_server
+  deno compile https://deno.land/std/examples/colors.ts colors",
+    )
 }
 
 fn bundle_subcommand<'a, 'b>() -> App<'a, 'b> {
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -819,8 +819,8 @@ fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
     .about("Compile the script into a self contained executable")
     .long_about(
       "Compiles the given script into a self contained executable.
-  deno compile https://deno.land/std/http/file_server.ts file_server
-  deno compile https://deno.land/std/examples/colors.ts colors",
+  deno compile --unstable https://deno.land/std/http/file_server.ts file_server
+  deno compile --unstable https://deno.land/std/examples/colors.ts colors",
     )
 }
 
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -820,7 +820,9 @@ fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
     .long_about(
       "Compiles the given script into a self contained executable.
   deno compile --unstable https://deno.land/std/http/file_server.ts file_server
-  deno compile --unstable https://deno.land/std/examples/colors.ts colors",
+  deno compile --unstable https://deno.land/std/examples/colors.ts colors
+  
+Cross compiling binaries for different platforms is not currently possible.",
     )
 }
 
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -24,7 +24,7 @@ pub enum DenoSubcommand {
   },
   Compile {
     source_file: String,
-    out_file: String,
+    out_file: Option<String>,
   },
   Completions {
     buf: Box<[u8]>,
@@ -419,7 +419,7 @@ fn compile_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   compile_args_parse(flags, matches);
 
   let source_file = matches.value_of("source_file").unwrap().to_string();
-  let out_file = matches.value_of("out_file").unwrap().to_string();
+  let out_file = matches.value_of("out_file").map(|s| s.to_string());
 
   flags.subcommand = DenoSubcommand::Compile {
     source_file,
@@ -828,13 +828,21 @@ fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
         .takes_value(true)
         .required(true),
     )
-    .arg(Arg::with_name("out_file").takes_value(true).required(true))
+    .arg(Arg::with_name("out_file").takes_value(true))
     .about("Compile the script into a self contained executable")
     .long_about(
       "Compiles the given script into a self contained executable.
-  deno compile --unstable https://deno.land/std/http/file_server.ts file_server
-  deno compile --unstable https://deno.land/std/examples/colors.ts colors
+  deno compile --unstable https://deno.land/std/http/file_server.ts
+  deno compile --unstable https://deno.land/std/examples/colors.ts color_util
   
+The executable name is inferred by default:
+  - Attempt to take the file stem of the URL path. The above example would
+    become 'file_server'.
+  - If the file stem is something generic like 'main', 'mod', 'index' or 'cli',
+    and the path has no parent, take the file name of the parent path. Otherwise
+    settle with the generic name.
+  - If the resulting name has an '@...' suffix, strip it.
+
 Cross compiling binaries for different platforms is not currently possible.",
     )
 }
@@ -3243,15 +3251,14 @@ mod tests {
     let r = flags_from_vec_safe(svec![
       "deno",
       "compile",
-      "https://deno.land/std/examples/colors.ts",
-      "colors"
+      "https://deno.land/std/examples/colors.ts"
     ]);
     assert_eq!(
       r.unwrap(),
       Flags {
         subcommand: DenoSubcommand::Compile {
           source_file: "https://deno.land/std/examples/colors.ts".to_string(),
-          out_file: "colors".to_string()
+          out_file: None
         },
         ..Flags::default()
       }
@@ -3267,7 +3274,7 @@ mod tests {
       Flags {
         subcommand: DenoSubcommand::Compile {
           source_file: "https://deno.land/std/examples/colors.ts".to_string(),
-          out_file: "colors".to_string()
+          out_file: Some("colors".to_string())
         },
         unstable: true,
         import_map_path: Some("import_map.json".to_string()),
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -3179,4 +3179,49 @@ mod tests {
       }
     );
   }
+
+  #[test]
+  fn compile() {
+    let r = flags_from_vec_safe(svec![
+      "deno",
+      "compile",
+      "https://deno.land/std/examples/colors.ts",
+      "colors"
+    ]);
+    assert_eq!(
+      r.unwrap(),
+      Flags {
+        subcommand: DenoSubcommand::Compile {
+          source_file: "https://deno.land/std/examples/colors.ts".to_string(),
+          out_file: "colors".to_string()
+        },
+        ..Flags::default()
+      }
+    );
+  }
+
+  #[test]
+  fn compile_with_flags() {
+    #[rustfmt::skip]
+    let r = flags_from_vec_safe(svec!["deno", "compile", "--unstable", "--import-map", "import_map.json", "--no-remote", "--config", "tsconfig.json", "--no-check", "--reload", "--lock", "lock.json", "--lock-write", "--cert", "example.crt", "https://deno.land/std/examples/colors.ts", "colors"]);
+    assert_eq!(
+      r.unwrap(),
+      Flags {
+        subcommand: DenoSubcommand::Compile {
+          source_file: "https://deno.land/std/examples/colors.ts".to_string(),
+          out_file: "colors".to_string()
+        },
+        unstable: true,
+        import_map_path: Some("import_map.json".to_string()),
+        no_remote: true,
+        config_path: Some("tsconfig.json".to_string()),
+        no_check: true,
+        reload: true,
+        lock: Some(PathBuf::from("lock.json")),
+        lock_write: true,
+        ca_file: Some("example.crt".to_string()),
+        ..Flags::default()
+      }
+    );
+  }
 }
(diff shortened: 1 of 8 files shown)


compile.ts
```diff
@@ -0,0 +1,26 @@
...
+
+var newBin = new Uint8Array(
+  originalBin.byteLength + bundleData.byteLength + magicTrailer.byteLength,
+);
+newBin.set(originalBin, 0);
+newBin.set(bundleData, originalBin.byteLength);
+newBin.set(magicTrailer, originalBin.byteLength + bundleData.byteLength);
+
+await Deno.writeFile(out, newBin);
+console.log(`Created binary at ${out}`);
```
ry, line 26: Impressive how simply you accomplished this!

Appending random bytes to exe files is allowed on all platforms?

compile.ts
```diff
@@ -0,0 +1,26 @@
+const specifier = Deno.args[0];
+const out = Deno.args[1];
+const [_, bundle] = await Deno.bundle(specifier);
+
+const originalBinaryPath = Deno.execPath();
+const originalBin = await Deno.readFile(originalBinaryPath);
+
+const encoder = new TextEncoder();
```
bartlomieju, line 8: For first pass I suggest to not have bundling done automatically and thus all of this logic should be moved to Rust

lucacasonato: Yeah. Did that now. This was just 3AM POC :upside_down_face:

compile.ts
lucacasonato: Yes - validated on macOS, Linux, and Windows.

cli/flags.rs
```diff
@@ -789,6 +808,22 @@ The installation root is determined, in order of precedence:
...
 }
 
+fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
+  compile_args(SubCommand::with_name("compile"))
+    .arg(
+      Arg::with_name("source_file")
+        .takes_value(true)
+        .required(true),
+    )
+    .arg(Arg::with_name("out_file").takes_value(true).required(true))
```
bartlomieju, line 818: Add `unstable` flag and require it to run `deno compile`

cli/standalone.rs
```diff
@@ -0,0 +1,114 @@
...
+        module_url_specified: module_specifier.to_string(),
+        module_url_found: module_specifier.to_string(),
+      })
+    }
+    .boxed_local()
+  }
+}
+
+async fn run(source_code: String) -> Result<(), AnyError> {
+  let flags = Flags::default();
```
bartlomieju, line 98: That way no flags will be passed to compiled binary, instead I suggest modify `flags.argv` and place all CLI flags.

cli/standalone.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl ModuleLoader for EmbeddedModuleLoader {
+  fn resolve(
+    &self,
+    _op_state: Rc<RefCell<OpState>>,
+    specifier: &str,
+    _referrer: &str,
+    _is_main: bool,
+  ) -> Result<ModuleSpecifier, AnyError> {
+    assert_eq!(specifier, SPECIFIER);
```
bartlomieju, line 73: Add 3rd argument: `"Self-contained binaries don't support module loading"`

bartlomieju: Actually should it be turned into an error instead of a panic?

cli/standalone.rs
```diff
@@ -0,0 +1,114 @@
...
+    .expect("expected to be able to open current exe");
+  let magic_trailer_pos = current_exe
+    .seek(SeekFrom::End(-12))
+    .expect("expected to be able to seek to magic trailer in current exe");
+  let mut magic_trailer = [0; 12];
+  current_exe
+    .read_exact(&mut magic_trailer)
+    .expect("expected to be able to read magic trailer from current exe");
+  let (magic_trailer, bundle_pos) = magic_trailer.split_at(4);
+  if magic_trailer == b"DENO" {
```
bartlomieju, line 36: Maybe make a bit more magic? Also suggest to make it a `const` at the top of the file

cli/standalone.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::convert::TryInto;
+use std::env::current_exe;
+use std::fs::File;
+use std::io::Read;
+use std::io::Seek;
+use std::io::SeekFrom;
+use std::pin::Pin;
+use std::rc::Rc;
+
+pub fn standalone() {
```
bartlomieju, line 22: `try_run_standalone_binary`? A doc comment would be appreciated

cli/standalone.rs
```diff
@@ -0,0 +1,114 @@
...
+  let current_exe_path =
+    current_exe().expect("expect current exe path to be known");
```
bartlomieju, line 24: Since `standalone()` is always run I bet there's a high chance of panic happening here, suggest to replace all `expect()` calls with returning result

cli/standalone.rs
```diff
@@ -0,0 +1,114 @@
...
+      .seek(SeekFrom::Start(bundle_pos))
+      .expect("expected to be able to seek to bundle pos in current exe");
+
+    let bundle_len = magic_trailer_pos - bundle_pos;
+    let mut bundle = String::new();
+    current_exe
+      .take(bundle_len)
+      .read_to_string(&mut bundle)
+      .expect("expected to be able to read bundle from current exe");
+    // TODO: check amount of bytes read
```
bartlomieju, line 50: I also think we should hash embedded bundle and add that hash to the trailer to verify it when booting standalone binary

bartlomieju (changes requested):

cli/standalone.rs
lucacasonato: Yeah sounds good.

cli/standalone.rs
lucacasonato: I agree, but can we postpone that to when we add metadata like flags and import specifier to the self contained binary?

bartlomieju: Sure

cli/main.rs
```diff
@@ -149,6 +150,109 @@ fn get_types(unstable: bool) -> String {
...
+  } else {
+    out_file
+  };
+  tokio::fs::write(&out_file, final_bin).await?;
+  #[cfg(unix)]
+  {
+    use std::os::unix::fs::PermissionsExt;
+    let perms = std::fs::Permissions::from_mode(0o777);
+    tokio::fs::set_permissions(out_file, perms).await?;
+  }
```
bartlomieju, line 252: Move to `standalone`

cli/flags.rs
```diff
@@ -789,6 +808,22 @@ The installation root is determined, in order of precedence:
...
+    .about("Compile the script into a self contained executable")
+    .long_about(
+      "Compiles the given script into a self contained executable.
+  deno compile https://deno.land/std/http/file_server.ts file_server
+  deno compile https://deno.land/std/examples/colors.ts colors",
```
bartlomieju, line 823: Mention that `deno compile` can't cross compile currently

cli/main.rs
```diff
@@ -149,6 +150,109 @@ fn get_types(unstable: bool) -> String {
...
+    if let Some(ignored_options) = result_info.maybe_ignored_options {
+      eprintln!("{}", ignored_options);
+    }
+    if !result_info.diagnostics.is_empty() {
+      return Err(generic_error(result_info.diagnostics.to_string()));
+    }
+  }
+
+  let (bundle_str, stats, maybe_ignored_options) =
+    module_graph.bundle(module_graph::BundleOptions {
```
bartlomieju, line 209: Can this be deduplicated with contents of `bundle_command` 😬 ?

cli/standalone.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::env::current_exe;
+use std::fs::File;
+use std::io::Read;
+use std::io::Seek;
+use std::io::SeekFrom;
+use std::io::Write;
+use std::pin::Pin;
+use std::rc::Rc;
+
+const MAGIC_TRAILER: &[u8; 4] = b"D3N0";
```
bartlomieju, line 24: I'm wondering if this value should be longer, wdyt?

bartlomieju (commented):
I think this is a great first pass, but suggest to get more eyes on it before landing

cli/standalone.rs
lucacasonato: image

I think it is fine. Regular binaries seem to all have the same 16 bytes at the end. (where the EOF-3 bytes is always 0x0). I think this is enough.

kevinkassimo: fyi this is what I get on macOS for deno 1.5.4:
Screen Shot 2020-11-29 at 12 14 21 PM
But then I also find this counter example from python3:
Screen Shot 2020-11-29 at 12 28 17 PM
It seems that trailing zeroes are there just to ensure padding to a certain byte multiplier boundary. Thus I also think it we be reasonable to make the trailer longer, probably >= 64 bits (while in the crypto world, > 2^90 is when collision becomes impractical, though I guess we don't need to go to that extreme)

cli/standalone.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+async fn run(source_code: String, args: Vec<String>) -> Result<(), AnyError> {
+  let mut flags = Flags::default();
+  flags.argv = args[1..].to_vec();
+  // TODO(lucacasonato): remove once you can specify this correctly through embedded metadata
+  flags.unstable = true;
+  let main_module = ModuleSpecifier::resolve_url(SPECIFIER)?;
+  let program_state = ProgramState::new(flags.clone())?;
+  let permissions = Permissions::allow_all();
```
kevinkassimo, line 112: Probably not for this PR, but I think it would also be nice to allow optional permission flags in the future, where we add an extra trailer for the bit vector of permissions.

cli/standalone.rs
kevinkassimo: > I also think we should hash embedded bundle and add that hash to the trailer to verify it when booting standalone binary

Hashing is probably not going to help: one having permission to modify the file can tamper with the hash anyways (hashing or any other kind of auth is only useful if you could obtain the hash through a separate channel or encrypted), thus this layer of protection is unnecessary.

cli/standalone.rs
lucacasonato: Yes. I have some ideas on how to do this in the future.

cli/standalone.rs
lucacasonato: It seems this is different on macOS and Linux. I'll increase it to 8 bytes.

ry (approved):
LGTM

Milestone achievement.

bartlomieju (approved):
LGTM 🚀

jveres:
I tried this new feature and found an issue where `deno::program_state::ProgramState::get_emit` tries to load `file://$deno$/bundle.js` which is fake for compiled bundles and `deno::disk_cache::DiskCache::get_cache_filename` panics. It was trivial to fix by patching `cli/program_state/get_emit(...)` with this:
```rust
fn get_emit(&self, url: &Url) -> Option<(Vec<u8>, Option<Vec<u8>>)> {
    match url.scheme() {
      // we should only be looking for emits for schemes that denote external
      // modules, which the disk_cache supports
      "wasm" | "file" | "http" | "https" => {
        // ignore when called from a self-contained binary
        if url.to_string() == crate::standalone::SPECIFIER {
          return None;
        }
      },
      _ => {
        return None;
      }
    }
...
```
Not sure if this is the most appropriate way, though.

bartlomieju:
> I tried this new feature and found an issue where `deno::program_state::ProgramState::get_emit` tries to load `file://$deno$/bundle.js` which is fake for compiled bundles and `deno::disk_cache::DiskCache::get_cache_filename` panics. It was trivial to fix by patching `cli/program_state/get_emit(...)` with this:
> 
> ```rust
> fn get_emit(&self, url: &Url) -> Option<(Vec<u8>, Option<Vec<u8>>)> {
>     match url.scheme() {
>       // we should only be looking for emits for schemes that denote external
>       // modules, which the disk_cache supports
>       "wasm" | "file" | "http" | "https" => {
>         // ignore when called from a self-contained binary
>         if url.to_string() == crate::standalone::SPECIFIER {
>           return None;
>         }
>       },
>       _ => {
>         return None;
>       }
>     }
> ...
> ```
> 
> Not sure if this is the most appropriate way, though.

@jveres could you open a new issue and provide a reproduction?

nayeemrmn:
> found an issue where `deno::program_state::ProgramState::get_emit` tries to load `file://$deno$/bundle.js`

Yep, the fix is this:
```
diff --git a/cli/standalone.rs b/cli/standalone.rs
index 06c20bc2..7c9e9fcb 100644
--- a/cli/standalone.rs
+++ b/cli/standalone.rs
@@ -58,7 +58,7 @@ pub fn try_run_standalone_binary(args: Vec<String>) -> Result<(), AnyError> {
   }
 }
 
-const SPECIFIER: &str = "file://$deno$/bundle.js";
+const SPECIFIER: &str = "file:///$deno$bundle.js";
 
 struct EmbeddedModuleLoader(String);
```
I'm trying to come up with tests.

@bartlomieju 
I was able to reproduce with:
```ts
import * as path from "./std/path/mod.ts";
```
Seems pretty random though.

jveres:
@nayeemrmn quickly checked with `file:///$deno$bundle.js` but it gives `error: Self-contained binaries don't support module loading` so maybe it's a different issue.

bartlomieju:
CC @lucacasonato

bartlomieju:
> @nayeemrmn quickly checked with `file:///$deno$bundle.js` but it gives `error: Self-contained binaries don't support module loading` so maybe it's a different issue.

@jveres do you have dynamic imports in your code? If so they are not supported

jveres:
> > @nayeemrmn quickly checked with `file:///$deno$bundle.js` but it gives `error: Self-contained binaries don't support module loading` so maybe it's a different issue.
> 
> @jveres do you have dynamic imports or workers in your code? If so they are not supported

No, I don't. Using my fix above it works.
I will create a test case for this later.

jveres:
as simple as 
```typescript
const e = new Error();
console.log(e);
```
```sh
deno compile --unstable test.ts test
RUST_BACKTRACE=full ./test
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', cli/disk_cache.rs:72:39
stack backtrace:
   0:        0x10b8c56d3 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hd4b962ed89f71a03
   1:        0x10b543b50 - core::fmt::write::h94ae1e793baa7a00
   2:        0x10b8c4dbb - std::io::Write::write_fmt::h5c716758fdc3057f
   3:        0x10b8c4745 - std::panicking::default_hook::{{closure}}::hc6119c7d16548caf
   4:        0x10b8c3e85 - std::panicking::rust_panic_with_hook::hc36596b4257bea99
   5:        0x10b8e133c - std::panicking::begin_panic_handler::{{closure}}::h49e5ddc3f21ca2fb
   6:        0x10b8e1307 - std::sys_common::backtrace::__rust_end_short_backtrace::h9bd32c9ad3fad18f
   7:        0x10b8e12c0 - _rust_begin_unwind
   8:        0x10c60cc4f - core::panicking::panic_fmt::hcdc9362d34d55302
   9:        0x10c60d225 - core::option::expect_none_failed::hef0458611b13747b
  10:        0x10b39962e - core::result::Result<T,E>::unwrap::h5376c858e103cf24
  11:        0x10b48c14d - deno::disk_cache::DiskCache::get_cache_filename_with_extension::h733f5eb5af07bc3e
  12:        0x10b4c086c - deno::program_state::ProgramState::get_emit::h13846ddd08588463
  13:        0x10b4c1514 - deno::source_maps::get_orig_position::hd223e3d5eecacb10
  14:        0x10b44515b - deno_core::ops::json_op_sync::{{closure}}::h1c84ae06d203eb0e
  15:        0x10b3d76c5 - <alloc::boxed::Box<F> as core::ops::function::Fn<A>>::call::h1744985c3c7bbe57
  16:        0x10b4a4917 - deno::metrics::metrics_op::{{closure}}::hfbfe0e6b2fb6b282
  17:        0x10b3d76c5 - <alloc::boxed::Box<F> as core::ops::function::Fn<A>>::call::h1744985c3c7bbe57
  18:        0x10b557b7e - <extern "C" fn(A0) .> R as rusty_v8::support::CFnFrom<F>>::mapping::c_fn::hf0d234ae6b0410d0
  19:        0x10bb6c059 - __ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
  20:        0x10bb6b5c8 - __ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE
  21:        0x10bb6ab80 - __ZN2v88internalL26Builtin_Impl_HandleApiCallENS0_16BuiltinArgumentsEPNS0_7IsolateE
fatal runtime error: failed to initiate panic, error 5
[1]    62205 abort      RUST_BACKTRACE=full ./test
```

lucacasonato:
@jveres See #8578

zlianon:
output size of file_server is 48MB 😬

lucacasonato:
@zlianon As discussed in the release post, we are confident we can reduce size to around 20 MB on Linux / macOS. Probably even less on Windows. See https://deno.land/posts/v1.6#codedeno-compilecode-self-contained-standalone-binaries

Issue #8357, opened 2020-11-12
Suggestion: using new system info crate


jopemachine:
### Issue

The `sys-info` crate seems that it doesn't provide enough APIs for `deno` and `deno`'s node compatibility.

I think in this issue (#7774), `Deno.systemCpuInfo` discarded node compatibility because of it.

### Suggestion

How about using a new crate like [heim](https://github.com/heim-rs/heim)?

### Example

For example, I think the `uptime` function could be implemented by the below code using `heim`

```rust
fn op_uptime(
  state: &mut OpState,
  _args: Value,
  _zero_copy: &mut [ZeroCopyBuf],
) -> Result<Value, AnyError> {
  super::check_unstable(state, "Deno.uptime");
  state.borrow::<Permissions>().check_env()?;

  let uptime = (host::uptime().await?.get::<time::second>() as u32);

  Ok(json!({
    "uptime": uptime
  }))
}
```

This function would be needed to implement node compatibility.

Related issue: #3802.

### Notes

If we need another package other than `heim`, I would be appreciated if you let me know.

kitsonk:
Achieving 100% compatibility with Node.js is not achievable with Deno, nor is it a goal of Deno.

What APIs that are critical that are missing where there are use cases where mocking the API isn't feasible?

heim is a lot of crates, and haven't gone through all the dependencies yet, and it seems like it is fairly slow development that has slowed, wit the biggest version being downloaded being a beta release.  That doesn't feel comfortable having such an "immature" set of a crates in Deno.

littledivy:
Looking at host::uptime https://github.com/heim-rs/heim/blob/master/heim-host/src/sys/linux/uptime.rs - `heim` basically reads `/proc/uptime` on linux.  Adding a heavy dependency for such small task is unnecessary.

jopemachine:
> Looking at host::uptime https://github.com/heim-rs/heim/blob/master/heim-host/src/sys/linux/uptime.rs - `heim` basically reads `/proc/uptime` on linux. Adding a heavy dependency for such small task is unnecessary.

Thanks for letting me know this :)

I think I could implement `uptime` by this method.

I will look into if other methods could be implemented without additional dependencies.

bartlomieju:
I'm -1 on bringing `heim` as dependency. As pointed out it's pretty heavy dependency for the task at hand. Respectfully closing.

PR #8540, opened 2020-11-29
feat(doc): Added uptime function for Deno

MVEMCJSUNPE:
Provides uptime function for issue #3802 . I haven't added tests yet, because I'm not sure how to test uptime. For example, the test loadavgSuccess merely checks whether an array of 3 is given by Deno.loadavg; it doesn't actually check the loadtimes. 

As it turns out in issue #8357 , heim is NOT needed to implement uptime. sysinfo already provides a function that can give the uptime called boottime, as I found in this [issue](https://github.com/FillZpp/sys-info-rs/issues/8#issue-199626066). 

Note: this is still a WIP, even though it should hopefully be mostly done.

### DIFF
--- a/cli/diagnostics.rs
+++ b/cli/diagnostics.rs
@@ -81,6 +81,7 @@ const UNSTABLE_DENO_PROPS: &[&str] = &[
   "umask",
   "utime",
   "utimeSync",
+  "uptime",
 ];
 
 lazy_static! {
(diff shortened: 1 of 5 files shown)


MVEMCJSUNPE:
I'm closing because for whatever reason, boottime is only implemented for non-Windows systems. While there is a pending pull request on sys-info that will hopefully resolve this issue (https://github.com/FillZpp/sys-info-rs/pull/48), I don't see much point in keeping this PR open.

PR #8542, opened 2020-11-29
docs(std): add some missing JSDoc testing/asserts.ts

kajirikajiri:
Add missing JSDoc for `testing/asserts.ts`

Related to #7487

- [x] deno run -A std/testing/asserts.ts

### DIFF
--- a/std/testing/asserts.ts
+++ b/std/testing/asserts.ts
@@ -66,6 +66,7 @@ function createSign(diffType: DiffType): string {
   }
 }
 
+/** Build assertions into pretty form. */
 function buildMessage(diffResult: ReadonlyArray<DiffResult<string>>): string[] {
   const messages: string[] = [];
   messages.push("");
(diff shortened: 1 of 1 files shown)


kajirikajiri:
sorry. I don't know how to fix this error. It seems to have occurred due to a formatting failure.
Did this error occur because I added the comment in the wrong way?

screenshot 2020-11-29 22 33 23

bartlomieju:
@kajirikajiri that is correct, it's an output from one of the tests. You however got a failure in `run_watch` which shouldn't happen. Make sure you're rebased against latest `master` branch.

kajirikajiri:
@bartlomieju Thanks for the quick reply! I'll give it a try

kajirikajiri:
@bartlomieju The test was successful after pulling the latest branch. Thanks!

bartlomieju:
@kajirikajiri thank you for the PR, but both functions you added docs to are internal (ie. not exported) and their names are self-explanatory so I don't think we need to add doc comments to them. I'll be more than happy to land doc strings for other public (exported) functions. Closing without merge.

kajirikajiri:
@bartlomieju thank you. ok!

PR #8547, opened 2020-11-29
fix(cli): support passing cli arguments under `deno eval`

### LINKED ISSUE #8538: `deno eval` does not support passing cli arguments
It seems that `deno eval` does not support passing CLI arguments.

```
$ deno eval "console.log(Deno.args)" my_arg
error: Found argument 'my_arg' which wasn't expected, or isn't valid in this context

USAGE:
    deno eval [OPTIONS] <code>

```

Similarly tried `deno eval "..." -- my_arg` or `deno eval "..." - my_arg`, none of which is working.

In comparison, Node supports eval with arguments:
```
$ node -e "console.log(process.argv)" a b c
[ '[REDACTED]/bin/node', 'a', 'b', 'c' ]
```

kevinkassimo:
Fixes #8538. Allow `deno eval` to take cli arguments.

Before:

```
$ deno eval -p Deno.args my_arg
error: Found argument 'my_arg' which wasn't expected, or isn't valid in this context

USAGE:
    deno eval <code> --print

```

After:

```
$ ./target/debug/deno eval -p Deno.args my_arg
[ "my_arg" ]
```

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -464,9 +464,19 @@ fn eval_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   flags.allow_write = true;
   flags.allow_plugin = true;
   flags.allow_hrtime = true;
-  let code = matches.value_of("code").unwrap().to_string();
   let as_typescript = matches.is_present("ts");
   let print = matches.is_present("print");
+  let mut code: Vec<String> = matches
+    .values_of("code_arg")
+    .unwrap()
+    .map(String::from)
+    .collect();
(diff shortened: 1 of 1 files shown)


kitsonk (commented):
Do we think we need an integration test for something like this?

bartlomieju:
> Do we think we need an integration test for something like this?

It was a bug in CLI parser so I think provided unit test is sufficient

kitsonk (approved):
LGTM

PR #8551, opened 2020-11-30
chore: enable sorting imports and exports in dprint

bartlomieju:
This commit updates internal `dprint` configuration to format
import and export declarations. 

Ref https://github.com/dprint/dprint-plugin-typescript/issues/57

If the result is satisfactory we should consider switching on
these options in `deno fmt`


cli/tests/019_media_types.ts
```diff
@@ -2,14 +2,14 @@
...
-import { loaded as loadedJs1 } from "http://localhost:4545/cli/tests/subdir/mt_text_javascript.j1.js";
 import { loaded as loadedJs2 } from "http://localhost:4545/cli/tests/subdir/mt_application_ecmascript.j2.js";
-import { loaded as loadedJs3 } from "http://localhost:4545/cli/tests/subdir/mt_text_ecmascript.j3.js";
 import { loaded as loadedJs4 } from "http://localhost:4545/cli/tests/subdir/mt_application_x_javascript.j4.js";
+import { loaded as loadedTs4 } from "http://localhost:4545/cli/tests/subdir/mt_application_x_typescript.t4.ts";
+import { loaded as loadedJs3 } from "http://localhost:4545/cli/tests/subdir/mt_text_ecmascript.j3.js";
+import { loaded as loadedJs1 } from "http://localhost:4545/cli/tests/subdir/mt_text_javascript.j1.js";
+import { loaded as loadedTs1 } from "http://localhost:4545/cli/tests/subdir/mt_text_typescript.t1.ts";
+import { loaded as loadedTs3 } from "http://localhost:4545/cli/tests/subdir/mt_video_mp2t.t3.ts";
+import { loaded as loadedTs2 } from "http://localhost:4545/cli/tests/subdir/mt_video_vdn.t2.ts";
```
dsherret, line 12: It sorts by module specifier (absolute first, then relative descending based on `../`), so for example:

```ts
import {} from "./a";
import {} from "../a";
import {} from "../../b";
import {} from "https://deno.land/b.ts";
import {} from "https://deno.land/a.ts";
import {} from "../../a";
```

Goes to...

```ts
import {} from "https://deno.land/a.ts";
import {} from "https://deno.land/b.ts";
import {} from "../../a.ts";
import {} from "../../b.ts";
import {} from "../a.ts";
import {} from "./a.ts";
```

The code here's isn't exactly the best example of when this might be desired.

bartlomieju:
@dsherret I'm a bit surprised that order of imports is sorted instead of bindings between `{ ... }`. 

If modules have side effects automatic reordering of imports will break that behaviour; requiring user to add `// deno-fmt-ignore` to enforce it. I'm sure it's gonna be quite surprising behaviour. May I ask why it works like this instead of sorting import/export bindings between `{ ... }`?

dsherret:
@bartlomieju it already does it for inside `{ ... }` and it's been on for deno fmt for a while. Yeah, that's the [danger of this feature](https://github.com/dprint/dprint-plugin-typescript/issues/57#issuecomment-699579666).

bartlomieju:
> @bartlomieju it already does it for inside `{ ... }` and it's been on for deno fmt for a while. Yeah, that's the [danger of this feature](https://github.com/dprint/dprint-plugin-typescript/issues/57#issuecomment-699579666). Right now, you can control the order by adding a blank line or comment.

@dsherret I see, I think it would quite a breaking change in `deno fmt` and not sure if desired one. Also I'd be quite cumbersome to update tests that have side effects (mostly `console.log`s at top level) to enforce this rule. I'm gonna close this PR for now

dsherret:
Yup, makes complete sense. It might cause some bugs in the wild.

guy-borderless:
Perhaps just have an option to automatically send the imports to the top without changing the order, that way in a large file, we get the DX of  just pasting the import above the function, and on save it's where it should be, without scrolling back and forth

PR #8552, opened 2020-11-30

bartlomieju:
Also unpins swc dependencies in `Cargo.toml` which should make
future upgrades less painful.


kitsonk (approved):
LGTM

PR #8554, opened 2020-11-30
feat(std/http): Add HTTP server error event

lideming:
This PR allows the user to handle/log the error from the listener and the request reader.

Before:
The HTTP server rethrows the error from the listener if the error type is unknown, and causes an uncaught error. Also, it ignores all errors while reading requests.

After:
The `Server` class is now inherited from `EventTarget`. It will pass the error to the `error` event. The user can handle it after adding an event listener to the server.

Related issue: #8499 (not closing it because I don't know where did the ENOTCONN come from)

Usage:
```ts
const server = serve({ port: 8000 });
server.addEventListener('error', (evt) => {
  if (evt.origin == "request") {
    console.warn("Error reading request", evt.connection, evt.error);
    // The server will send "400 Bad Request" and close the connection
    // unless `evt.preventDefault()` is called.
  } else if (evt.origin == "listener") {
    console.warn("Error accepting connection", evt.error);
    // The server will continue to accept connections
    // unless `evt.preventDefault()` is called.
  }
})
  req.respond({ body: "Hello World\n" });
}
```

This PR also changes `file_server` to add event listener to print errors.

### DIFF
--- a/std/http/server.ts
+++ b/std/http/server.ts
@@ -113,11 +113,58 @@ export class ServerRequest {
   }
 }
 
-export class Server implements AsyncIterable<ServerRequest> {
+export type ServerErrorEventListenerOrEventListenerObject =
+  | ServerErrorListener
+  | ServerErrorListenerObject;
+
+export interface ServerErrorListener {
+  (evt: ServerErrorEvent): void | Promise<void>;
+}
+
+export interface ServerErrorListenerObject {
+  handleEvent(evt: ServerErrorEvent): void | Promise<void>;
+}
+
+export interface ServerErrorEventInit extends ErrorEventInit {
+  server: Server;
+  origin: ServerErrorEvent["origin"];
+  connection?: Deno.Conn;
+}
+
+export class ServerErrorEvent extends ErrorEvent {
+  server: Server;
+  origin: "listener" | "request";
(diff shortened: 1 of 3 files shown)


std/http/server.ts
```diff
@@ -113,11 +113,58 @@ export class ServerRequest {
...
+
+export interface ServerErrorEventInit extends ErrorEventInit {
+  server: Server;
+  origin: ServerErrorEvent["origin"];
+  connection?: Deno.Conn;
+}
+
+export class ServerErrorEvent extends ErrorEvent {
+  server: Server;
+  origin: "listener" | "request";
```
lideming, line 136: Maybe we should rename these (origin, listener, request)... My English is not good enough.

lideming:
Some problems:
1) Is it okay to make Server `extends EventTarget` to implement events? Should we implement it in another way?
2) Just to confirm. While I am guessing `extends EventTarget` will not break something, is it considered a BREAKING change?

bartlomieju:
@lideming  we won't land this PR before standard library migration to a separate repository (#9029). I'm still very much interested in discussing these changes on upcoming standard library meetings and potentially landing the PR.

Let's keep it open until the migration is done.

bartlomieju:

PR #8555, opened 2020-11-30
refactor(cli): DRY cli/ast.rs

bartlomieju:

### DIFF
--- a/cli/tests/bundle/fixture15.out
+++ b/cli/tests/bundle/fixture15.out
@@ -1,3 +1,3 @@
 export function getIndex(c) {
-    return "\u{0}\r\n\u{85}\u{2028}\u{2029}".indexOf(c);
+    return "\x00\r\n\x85\u2028\u2029".indexOf(c);
 }
--- a/cli/tests/bundle/fixture15.out
+++ b/cli/tests/bundle/fixture15.out
@@ -1,3 +1,3 @@
 export function getIndex(c) {
-    return "\x00\r\n\x85\u2028\u2029".indexOf(c);
+    return "\u{0}\r\n\u{85}\u{2028}\u{2029}".indexOf(c);
 }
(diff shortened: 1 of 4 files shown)


cli/tests/bundle/fixture15.out
```diff
@@ -1,3 +1,3 @@
 export function getIndex(c) {
-    return "\x00\r\n\x85\u2028\u2029".indexOf(c);
+    return "\u{0}\r\n\u{85}\u{2028}\u{2029}".indexOf(c);
```
kitsonk, line 2: Something is wrong if you need to do this now... @kdy1 fixed this, so it shouldn't modify this...

kitsonk (commented):
Something isn't right if you are changing fixture15... it isn't supposed to change at all.

cli/tests/bundle/fixture15.out
kitsonk: See: https://github.com/denoland/deno/pull/8162#issuecomment-731719973

bartlomieju: Thanks for pointing this out, you're right that it shouldn't change, that said, I'm a bit lost on the cause of this change, removed code is exactly the same as code in `parse` 😬 I'll look for the problem

bartlomieju: Actually, it seems that `transpile_module` already passes `Rc<SourceMap>` as an argument

bartlomieju:
Closing without merge for now

cli/tests/bundle/fixture15.out
bartlomieju: Alright, I've managed to fix it by splitting `parse()` into `parse_module()` (that works as `parse()` did) and adding lower-level utility called `parse_with_source_map()`

kitsonk (approved):
LGTM

PR #8557, opened 2020-11-30
disable flaky watch tests

bartlomieju:
Temporarly disabling tests for `--watch` flag which are very flaky.


ry (approved):
LGTM - please create a new issue to re-enable with the list of ignored tests

Issue #8338, opened 2020-11-10
Sometimes print "undefined" when close() in REPL since 1.5.0


KNnut:
And 1.4.6 works well.

kitsonk:
cc/ @caspervonb

caspervonb:
Race condition I suppose, the result of calling close is undefined, we check if closing is set before printing eval

caspervonb:
Am unable to reproduce this at all, had like 100-200 runs at it.
`close` and `globalThis.closing` are synchronous so my initial theory of a race was wrong.

KNnut:
@caspervonb Still on latest `master`.
![](https://user-images.githubusercontent.com/9387720/100727943-de0df380-3401-11eb-8a11-166f28f47fa8.gif)
![](https://user-images.githubusercontent.com/9387720/100727952-e1a17a80-3401-11eb-9abd-403475596c8c.gif)

caspervonb:
Yeah forgot, moved the condition to the loop condition.
Have a PR moving it back but this might be wontfix as close isn't necessarily the same as Deno.exit.

PR #8558, opened 2020-11-30
fix(repl): close calls sometimes prints results

caspervonb:
Closes https://github.com/denoland/deno/issues/8338


bartlomieju (approved):
LGTM

PR #8563, opened 2020-12-01
chore(cli/flags): Use deno compile --output for the out file

nayeemrmn:
Trailing positional args for `deno compile` are reserved for baked-in script args, as seen in `deno install`.


cli/flags.rs
```diff
@@ -828,13 +828,19 @@ fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
...
+        .help("Output file (defaults to $PWD/<inferred-name>)")
+        .takes_value(true)
+    )
     .about("Compile the script into a self contained executable")
     .long_about(
       "Compiles the given script into a self contained executable.
   deno compile --unstable https://deno.land/std/http/file_server.ts
-  deno compile --unstable https://deno.land/std/examples/colors.ts color_util
-  
+  deno compile --unstable --output color_util https://deno.land/std/examples/colors.ts
```
lucacasonato, line 842: Could you add one more example where you use a fully qualified path?

cli/main.rs
```diff
@@ -166,10 +166,11 @@ async fn compile_command(
...
-  let out_file = match out_file {
-    Some(out_file) => out_file,
+  let output = output.or_else(|| {
+    infer_name_from_url(module_specifier.as_url()).map(PathBuf::from)
+  });
+  let output = match output {
+    Some(output) => output,
     None => return Err(generic_error(
       "An executable name was not provided. One could not be inferred from the URL. Aborting.",
     )),
```
lucacasonato, line 175: This could all be replaced by an `.ok_or(err)?;` actually.

cli/flags.rs
nayeemrmn: Replaced the existing one, relative path capability is usually obvious.

lucacasonato (approved):
LGTM - thanks!

PR #8566, opened 2020-12-01
feat(std/node): Add `setImmediate` and `clearImmediate` to global scope

### LINKED ISSUE #8500: setImmediate missing from node shim globals
When you import a module through `createRequire`, there won't be a global `setImmediate` available, which is required for nodejs compatibility.

Liamolucko:
Fixes #8500

### DIFF
--- a/std/node/global.ts
+++ b/std/node/global.ts
@@ -34,7 +34,7 @@ Object.defineProperty(globalThis, "setImmediate", {
 
 Object.defineProperty(globalThis, "clearImmediate", {
   value: timers.clearImmediate,
-  enumerable: false,
+  enumerable: true,
   writable: true,
   configurable: true,
 });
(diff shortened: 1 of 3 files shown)


std/node/global.ts
```diff
@@ -24,4 +25,18 @@ Object.defineProperty(globalThis, "Buffer", {
   configurable: true,
 });
 
+Object.defineProperty(globalThis, "setImmediate", {
+  value: timers.setImmediate,
+  enumerable: false,
```
bnoordhuis, line 30:
```suggestion
  enumerable: true,
```
Ditto on line 37.

Liamolucko: Yep, you're right - I mistakenly assumed that it would be the same as the other globals.

bnoordhuis (approved):
LGTM, thanks. I think this should be safe to land, the CI failure is a network error in an unrelated test.

PR #8569, opened 2020-12-01
fix(repl): higlight `async` and `of` in REPL

### LINKED ISSUE #8568: Some keywords not highlighted by new repl
The new REPL highlighting doesn't currently highlight `async` and `of` as keywords (my bad).
The SWC lexer doesn't count them as keywords because they aren't _technically_ reserved words - you can do this:
```javascript
// Technically valid JavaScript
let async = 5;
```
The 'correct' way would be to add checks for whether it's actually being used as a keyword, but I think those are probably minor enough cases not to worry about them.

Liamolucko:
Resolves #8568.

This just highlights them unconditionally; since they can technically be used as identifiers, the 'correct' way would check for that, but that would pretty much require a full parser.

### DIFF
--- a/cli/tools/repl.rs
+++ b/cli/tools/repl.rs
@@ -238,6 +238,8 @@ impl Highlighter for LineHighlighter {
                   colors::gray(&line[span]).to_string()
                 } else if ident == *"Infinity" || ident == *"NaN" {
                   colors::yellow(&line[span]).to_string()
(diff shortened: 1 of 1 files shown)


caspervonb:
Should probably look back at the previous token at-least, and only highlight if it's not a period?

ry (approved):
LGTM- thanks @Liamolucko

Issue #8565, opened 2020-12-01
Request constructor allows passing non-dictionary object as second parameter


CrimsonCodes0:
In Chromium, calling fetch with a second parameter that isn't an object causes the promise to reject:

```js
const not_an_object = 0;

fetch(url, not_an_object);
```

```
TypeError: Failed to execute 'fetch' on 'Window': cannot convert to dictionary.
```

Deno lets it pass without error.

It may not just be checking for a generic object, I'll try to figure out the specifics of the "dictionary" referenced soon.

## Update: ##
https://fetch.spec.whatwg.org/#dom-global-fetch
The HTML fetch specification says the following:

> The fetch(input, init) method steps are:
> 
> Let p be a new promise.
> 
> Let requestObject be the result of invoking the initial value of Request as constructor with input and init as arguments. If this throws an exception, reject p with it and return p.

It is the `Request` constructor that should throw when not given an object.

https://fetch.spec.whatwg.org/#request says that the second parameter `init ` is an optional "dictionary" I'm not going to check out the details of that, but Chromium allows the following to pass:
```js
new Request(url, {});
new Request(url, []);
new Request(url, () => {});
new Request(url, null);
new Request(url, undefined);
```
and the following fail:
```js
new Request(url, 0);
new Request(url, 0n);
new Request(url, "");
new Request(url, false);
new Request(url, Symbol());
```

Deno only rejects on passing a Symbol, because a symbol is a truthy value, and the `in` operator was used on the init.

00ff0000red:
While checking the source, I realized that it wouldn't allow `String` objects to be used as a URL, only string primitives, updating that to work too.

CrimsonCodes0:
Fixed quite some time ago, during the full rewrite.

PR #8580, opened 2020-12-01
fix(op_crates/fetch): type check Request init

00ff0000red:
Checks the Request initializer parameter for its type (and value).
Mimics Chromium implementation.

If the init is null or undefined, it would be replaced with an object, if it is a function, nothing is done, anything else results in a type error being thrown.

Fixes <https://github.com/denoland/deno/issues/8565>.

Also updates `fetch` to call the `Request` constructor, and to use its results.

If the Request construction fails, the error is propagated to the callee, as defined in the spec.

### DIFF
--- a/op_crates/fetch/26_fetch.js
+++ b/op_crates/fetch/26_fetch.js
@@ -1217,7 +1217,7 @@
         headers,
         body
       }) = new Request(input, init);
-    } catch(e) {
+    } catch {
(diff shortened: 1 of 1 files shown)


00ff0000red:
While checking the source, I realized that it wouldn't allow `String` objects to be used as a URL, only string primitives, updating that to work too.

Seems that Request already converts it to a string primitive, so by using the Request, it should be fixed.

00ff0000red:
I closing this and leaving it to someone else, I'm not the right person to be working with that code.

00ff0000red:
I'm scrapping this

Issue #8578, opened 2020-12-01
`deno compile` stack trace printing causes panics


lucacasonato:
```
~/p/g/d/deno ❯❯❯ cat test.js
throw new Error("foo");
~/p/g/d/deno ❯❯❯ deno compile --unstable test.js
Bundle file:///mnt/f9/Projects/github.com/denoland/deno/test.js
Compile file:///mnt/f9/Projects/github.com/denoland/deno/test.js
Emit test
~/p/g/d/deno ❯❯❯ env RUST_BACKTRACE=1 ./test
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', cli/disk_cache.rs:72:39
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483
   1: core::panicking::panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/panicking.rs:85
   2: core::option::expect_none_failed
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/option.rs:1234
   3: core::result::Result<T,E>::unwrap
   4: deno::disk_cache::DiskCache::get_cache_filename_with_extension
   5: deno::program_state::ProgramState::get_emit
   6: deno::source_maps::get_orig_position
   7: deno_core::ops::json_op_sync::{{closure}}
   8: <alloc::boxed::Box<F> as core::ops::function::Fn<A>>::call
   9: deno::metrics::metrics_op::{{closure}}
  10: <alloc::boxed::Box<F> as core::ops::function::Fn<A>>::call
  11: <extern "C" fn(A0) .> R as rusty_v8::support::CFnFrom<F>>::mapping::c_fn
  12: _ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
  13: _ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE
  14: _ZN2v88internalL26Builtin_Impl_HandleApiCallENS0_16BuiltinArgumentsEPNS0_7IsolateE
[...]

Issue #8577, opened 2020-12-01
`deno bundle` should contain inlined source maps


lucacasonato:
Bundles do not contain inline source maps. We should emit and inline the source maps for bundles.

kitsonk:
This needs to be related to #4499, because emitting them and consuming them need to happen to be useful to Deno.

satyarohith:
Working on it.

rotu:
Thanks, @satyarohith! I'm excited to be able to debug Deno-bundled ts files!

jed:
In the meantime, I wrote [an installable script](https://deno.land/x/deno_bundle) as a drop-in replacement for `deno bundle` that inlines the sourcemap generated by `Deno.emit`.

lucacasonato:
`deno bundle` has been removed in Deno 2.

Issue #4499, opened 2020-03-27
Consume source maps from JavaScript files


kitsonk:
Current Deno will remap stack traces on files that have been compiled by Deno, as Deno caches both the original file and the emitted map, and loads those from the cache.  When consuming JavaScript with source maps from other sources (like already compiled code) Deno does not handle the source map and ignore source map pragma included in files.

Deno should support both external source maps and inline source maps when fetching modules and store any source maps in the cache to be used when remapping stack traces.

This will have the side-effect of supporting the generation of source maps for bundles generated by `deno bundle`.

One consideration is that source maps optionally can include the original sources.  If they do, then when emitting an error, the original source line is available, but if they don't, we would not be able to log-out/resolve that original source line.

71:
Several issues referring to inline source maps were closed as duplicates of this one, but this also applies to linked source maps:

```ts
function fail() {
  throw new Error("failed");
}

fail();
```

Build with esbuild:

```sh
$ esbuild repro.ts --minify --sourcemap=linked --outfile=repro.js
```

With Node and [`source-map-support`](https://github.com/evanw/node-source-map-support):

```sh
$ node -r source-map-support/register repro.js
file:///repro.ts:2
  throw new Error("failed");
        ^
Error: failed
    at fail (file:///repro.ts:2:9)
    at file:///repro.ts:5:1
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
```

With Deno:

```sh
$ deno run repro.js
error: Uncaught Error: failed
"use strict";function fail(){throw new Error("failed")}fail();
                                   ^
    at fail (file:///repro.js:1:36)
    at file:///repro.js:1:56
```

skovsboll:
I would love better source maps support. Useful for debugging unit tests.

bartlomieju:
This needs to be redesigned and implemented in `deno_core`.

The problem is that we have our own system of handling source maps and not using the built-in V8 APIs that Luca mentions [here](https://github.com/denoland/deno/issues/21988#issuecomment-1900245996).

We currently have a `SourceMapGetter` trait that needs to implement a couple methods to retrieve a source map.

[...]

PR #8581, opened 2020-12-01
fix(compile): disable source mapping of errors

bartlomieju:
This commit disables source mapping of errors
for standalone binaries. Since applying source
maps relies on using file fetcher infrastructure
it's not feasible to use it for standalone binaries
that are not supposed to use that infrastructure.

Closes https://github.com/denoland/deno/issues/8578

### DIFF
--- a/cli/ops/runtime.rs
+++ b/cli/ops/runtime.rs
@@ -13,11 +13,18 @@ use deno_core::OpState;
 use deno_core::ZeroCopyBuf;
 use std::env;
 
-pub fn init(rt: &mut deno_core::JsRuntime, main_module: ModuleSpecifier) {
+type ApplySourceMaps = bool;
+
+pub fn init(
+  rt: &mut deno_core::JsRuntime,
+  main_module: ModuleSpecifier,
+  apply_source_maps: bool,
+) {
   {
     let op_state = rt.op_state();
     let mut state = op_state.borrow_mut();
     state.put::<ModuleSpecifier>(main_module);
+    state.put::<ApplySourceMaps>(apply_source_maps);
   }
   super::reg_json_sync(rt, "op_start", op_start);
   super::reg_json_sync(rt, "op_main_module", op_main_module);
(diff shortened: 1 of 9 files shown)


kitsonk (approved):
LGTM as tactical solution until #8577 and #4499 are delivered.

nayeemrmn:
We should still change the specifier to not use file URL hosts. The error site shows that we have liberally used `file_url.to_file_path().unwrap()` which it breaks. It should also be more similar to the other `$deno$<subcommand>.js` names we have. Easiest fix would be `file:///$deno$bundle.js`. I think the best would be `file:///path/to/<exe>.$deno$bundle.js`

bartlomieju:
@nayeemrmn that was already suggested by Kitson in https://github.com/denoland/deno/issues/8578#issuecomment-736847042; I think we should go with some "magic" protocol instead of appending to current exe path. WDYT?

nayeemrmn:
I suggested a file URL because it matches the virtual modules of other subcommands. But maybe it's unfortunate to use a file URL which isn't in the FS, if so we should change all of them.

bartlomieju:
> I suggested a file URL because it matches the virtual modules of other subcommands. But maybe it's unfortunate to use a file URL which isn't in the FS, if so we should change all of them.

I guess `deno://` scheme would make sense?

nayeemrmn:
> I guess `deno://` scheme would make sense?

We use `deno:cli/rt/...` for internal bootstrap scripts. I think user code should always be something else.

kitsonk:
I am of the opinion of `vfs://` in advance of potentially supporting some sort of virtual file system.

I agree that `deno://` should be reserved for internal stuff...

The problem with the other subcommands, especially test, that I ran across is that the module expects to be relative... There is an argument though that test should have absolute specifiers it imports anyways.  That would be better IMO, I have always hated that it was a "lie".

bartlomieju:
> > I guess `deno://` scheme would make sense?
> 
> We use `deno:cli/rt/...` for internal bootstrap scripts. I think user code should always be something else.

Right, we should distinct user code.

> I am of the opinion of vfs:// in advance of potentially supporting some sort of virtual file system.

👍 

> The problem with the other subcommands, especially test, that I ran across is that the module expects to be relative... There is an argument though that test should have absolute specifiers it imports anyways. That would be better IMO, I have always hated that it was a "lie".

I think this stems from one of the first versions of `deno test` that tried to be too smart to cheat deno's TS compiler (shame on me 😅 )

nayeemrmn:
Actually, the relatively located file URLs are essential for `$deno$repl.ts`, `$deno$stdin.ts`, `$deno$eval.ts` for useful dynamic import resolution...

kitsonk:
@nayeemrmn true... we should "fix" it for `deno test` though.  The other option is to modify the module graph to have a different resolution base for these virtual files, instead of "faking" them.  They still need to be inserted into the file fetcher, but as long as we agree to the semantics of how to resolve relative dynamic imports (relative to cwd), that should be straight forward for the module graph to handle.

PR #8583, opened 2020-12-01
style: exposed TarOptions

behnammodi:
I improved `append` function's style by Deno guide


bartlomieju:
@behnammodi thank you for the PR. At yesterday's standard library WG meeting it was decided that `std/archive` needs a major rework/overhaul. I'm not keen on updating a module that will be reworked in the near future, so I'll close this PR without merge.

PR #8596, opened 2020-12-03
BREAKING(std): Rename std/datetime to std/date

### LINKED ISSUE #8594: Rename std/datetime to std/date
Also rename identifiers that have `DateTime` in them. `formatter.ts` and `tokenizer.ts` should be private, they have clean wrapper functions in `mod.ts`.

The module wraps around `Date` (`DateTimeFormatter::format(date: Date)`), which we should match. `DateTime` is used in other specs like `Intl`, there's nothing wrong with the term except that this module is strictly bound to the `Date` class.

In light of https://tc39.es/proposal-temporal/docs/index.html, it's possible we never want to stabilise this module in favour of `std/temporal` (or nothing). But unless we're removing it outright in the near future, the rename is still needed i.e. no sense in keeping `datetime` as a "neutral" name whose internals can just be swapped out -- that would be breaking either way.

nayeemrmn:
- Privatize `std/datetime/formatter.ts` to `std/date/_format.ts`.
- Privatize `std/datetime/tokenizer.ts` to `std/date/_tokenize.ts`.
- `std/log`: Rename `LogRecord::datetime` to `LogRecord::date`.

---

Closes #8594.


bartlomieju:
Let's discuss this PR as part of https://github.com/denoland/deno/issues/8405 before landing

bartlomieju:
At yesterday's standard library WG meeting it was decided that `std/datetime` will most likely be removed from `std/, so let's not do any renames in the meanwhile. Closing without merge.

Issue #8589, opened 2020-12-02
Add path.resolve() which does not use Deno.cwd() implicitly to std/path


whosmayank:
https://github.com/denoland/deno/pull/8387  removes ```path.resolve()``` from std/path. It'd be worth having ```path.resolve()``` as a counterpart to ```fs.resolvePath```  which mandates returning an absolute path.

iammack:
This should be a straight forward fix. I'm working on it :)

PR #8559, opened 2020-11-30
fix(std): Migrating path.relative to fs.relativePath

whosmayank:
For why the change is being made, refer https://github.com/denoland/deno/pull/8387
This PR depends on https://github.com/denoland/deno/pull/8387 and should be merged after it.
Closes https://github.com/denoland/deno/issues/8487

PR #8598, opened 2020-12-03
feat(std): Add path.resolve without Deno.cwd() implicit call

whosmayank:
This commit adds a path.resolve function which will throw if the resolved path is relative. The onus of passing Deno.cwd() is on user. If the user would like to opt for an alternative with implicit Deno.cwd() call, fs.resolvePath can be utilized. Corresponding tests are also added.
Closes https://github.com/denoland/deno/issues/8589
This PR is dependent on https://github.com/denoland/deno/pull/8559 and should be merged after it.


iammack:
CC @lucacasonato @nayeemrmn

bartlomieju:

PR #8600, opened 2020-12-03
fix(op_crates/fetch): support non-ascii response headers value

### LINKED ISSUE #8576: non-ascii response headers in fetch result in panic
I got this error...
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ToStrError { _priv: () }', D:\a\deno\deno\op_crates\fetch\lib.rs:159:53
```
There is no stack trace, Just this message.

jaeheonji:
Fixes #8576

This issue is related to the case where the fetch response header value is a non-ascii string.
As with the HTTP 301 response, some headers value can contain UTF-8 strings.

So, I simply changed the value as utf-8 string.

If you have a better idea, please give me feedback. :smile_cat:

### DIFF
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -158,7 +158,9 @@ where
   for (key, val) in res.headers().iter() {
     let key_string = key.to_string();
 
-    if key_string == "location" {
+    if val.as_bytes().is_ascii() {
+      res_headers.push((key_string, val.to_str().unwrap().to_owned()))
+    } else {
       res_headers.push((
         key_string,
         val
@@ -167,8 +169,6 @@ where
           .map(|&c| c as char)
           .collect::<String>(),
       ));
-    } else {
-      res_headers.push((key_string, val.to_str().unwrap().to_owned()))
     }
   }
 
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -29,6 +29,7 @@ use std::fs::File;
 use std::io::Read;
 use std::path::PathBuf;
 use std::rc::Rc;
+use std::str::from_utf8;
 
 pub use reqwest; // Re-export reqwest
 
@@ -156,7 +157,10 @@ where
   let status = res.status();
   let mut res_headers = Vec::new();
   for (key, val) in res.headers().iter() {
-    res_headers.push((key.to_string(), val.to_str().unwrap().to_owned()));
+    res_headers.push((
+      key.to_string(),
+      from_utf8(val.as_bytes()).unwrap().to_owned(),
+    ));
   }
 
   let rid = state
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -29,7 +29,6 @@ use std::fs::File;
 use std::io::Read;
 use std::path::PathBuf;
 use std::rc::Rc;
-use std::str::from_utf8;
 
 pub use reqwest; // Re-export reqwest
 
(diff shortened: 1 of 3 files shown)


op_crates/fetch/lib.rs
```diff
@@ -156,7 +157,10 @@ where
   let status = res.status();
   let mut res_headers = Vec::new();
   for (key, val) in res.headers().iter() {
-    res_headers.push((key.to_string(), val.to_str().unwrap().to_owned()));
+    res_headers.push((
+      key.to_string(),
+      from_utf8(val.as_bytes()).unwrap().to_owned(),
```
bnoordhuis, line 162: The `.unwrap()` will kill the process when the header is not valid UTF-8 (e.g., because it's Latin-1.)

You could try to parse the byte string as UTF-8 first, Latin-1 second, but some Latin-1 strings are also legal (if not legible) UTF-8 strings, and of course all UTF-8 strings are legal Latin-1. Occasional mojibake is inevitable with that approach.

For reference, Node.js has been parsing headers as Latin-1 from the moment V8 grew a concept of non-UTF-8 strings (commit nodejs/node@f674b09f40d22915e15b6968aafc5d25ac8178a2 from August 2013.)

bartlomieju (changes requested):
Marking as "Request changes" considering above comment

op_crates/fetch/lib.rs
jaeheonji: @bnoordhuis Thank you for your feedback!

I wasn't thinking straight. Checking your comments and code of headers, I've understood that the current implementation supports Latin-1.

And i also found out that the validator doesn't allow the headers value to be parsed correctly as UTF-8.

However, if we keep the existing code line 159, the process will be killed by `unwrap()` as in the above issue. And the explanation for the cause seems lacking.

So It seems good to solve the above issue by showing the detailed reason for the occurrence of an non-ascii error for` to_str.`

What do you think about this?

bnoordhuis:
@jaeheonji I got a notification that you posted a comment but I don't see it anywhere. Did you delete/retract it again or ...?

jaeheonji:
@bnoordhuis Thank you for your feedback and sorry for confusing you

I thought a bit about how to fix it.

First, I tried to solve this issue by referring to the source of `reqwest`.
https://github.com/seanmonstar/reqwest/blob/3cd9c29b303d98342aa03305655069b5d2382e63/src/async_impl/client.rs#L1419

But, as @bnoordhuis mentioned, Node.js and node-fetch parse the headers value as Latin1. So, instead of parsing the value as utf-8, I tried parsing it as Latin1.

In this case, I had two options. The first is the modified method as it is now, and the other is the method of using `encoding_rs`

I tried to approach it as simple as possible. In addition, it is thought that non-ascii strings will come only when the headers value is `location`, so a branch statement has been added.

op_crates/fetch/lib.rs
```diff
@@ -156,7 +156,20 @@ where
   let status = res.status();
   let mut res_headers = Vec::new();
   for (key, val) in res.headers().iter() {
-    res_headers.push((key.to_string(), val.to_str().unwrap().to_owned()));
+    let key_string = key.to_string();
+
+    if key_string == "location" {
```
bnoordhuis, line 161: `Refresh`, `Cookie` and various pingback headers have the same issue. I wouldn't special case, just do a check along the lines of:
```rust
if val.as_bytes().is_ascii() {
  // use as-is
} else {
  // convert to latin-1
}
```
Tangential: just checking "is it ascii" isn't great from a security perspective. Better would be to allow only "visible" ASCII (i.e., no control characters) but that's probably something for another pull request:
```rust
fn is_visible_ascii(c: u8) {
  c == b'\t' || c >= 32 && c < 127
}
```

jaeheonji: Thank you for your insight!

bnoordhuis (approved):
Bartek should do the final sign-off but LGTM, thanks!

bartlomieju (approved):
LGTM, thank you @jaeheonji!

PR #7573, opened 2020-09-19
feat(cli): update to TypeScript 4.1

kitsonk:
This PR updates Deno to TypeScript 4.1 beta in advance of TypeScript 4.1 being released to give an early view into any issues.

TypeScript 4.1 tightens up `Promise` definitions and there are quite a few instances in the code base where the type checking is failing due to this.

PR #8601, opened 2020-12-03
fix(std): Converting Deferred interface to class

whosmayank:
Converts Deferred interface in std/async to class. The deferred function can be removed then.


lucacasonato:
As discussed in std-wg. Could you remove the TODO comment, and update the tests?

iammack:
Yeah, sure thing. I just wanted to create a draft PR to hear everyone's opinions.

iammack:
Also there is a problem with current implementation. 
The snippet
```
const p = deferred();
p.resolve()
```
would actually instantiate a deferred<{}>. According to https://github.com/microsoft/TypeScript/pull/39817, unless promise type is void, resolve should pass an argument of type T. In current implementation due to some wrong code, this is bypassed. This is fixed in this PR.

iammack:
CC @ry  @lucacasonato

cli/tests/unit/net_test.ts
```diff
@@ -338,7 +338,7 @@ unitTest(
 unitTest(
   { perms: { net: true } },
   async function netTcpListenIteratorBreakClosesResource(): Promise<void> {
-    const promise = deferred();
+    const promise = new Deferred<void>();
```
ry, line 341: Is it necessary to put the `<void>` in here?

iammack: Yes, or else the default type is `{}` and mandates a parameter of type `{}` when resolve is called. This patch was introduced in a later version of typescript though (https://github.com/microsoft/TypeScript/pull/39817)

bartlomieju: https://github.com/denoland/deno/pull/7573#issue-489666614

ry: Writing `new Deferred<void>` is fairly obnoxious. We should reconsider this change.

nayeemrmn: @ry Additionally, extending `Promise` has some annoying behaviours which forces it to [accept an executor function](https://discord.com/channels/684898665143206084/688040863313428503/784110243004940351), which `deferred()` did not.

bartlomieju: I'm actually pretty happy with current implementation, I think we should only change `Object.assign()` bit to directly assign to `promise`

iammack: @ry @nayeemrmn, I think we'll need to explicitly write `deferred<void` too. Not mentioning type of `methods` in  https://github.com/denoland/deno/blob/master/std/async/deferred.ts#L21 allows current behaviour but has the downside of not mentioning type. But that can be fixed by adding type to current implementation too.

iammack:
Looks like this is an[ anti-pattern ](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern). Also, refer https://github.com/denoland/deno/issues/8405#issuecomment-739420300. 
We should consider refactoring it out of the code.

hayd:
@iammack it can be useful, in fact in your anti-pattern link it says:

> In Deferred anti-pattern, "deferred" objects are created **for no reason**, complicating code.

> So when should deferred be used?
> Well simply, when you have to.

Sometimes you have to.

iammack:
@hayd, it would be helpful if you could provide an example where using `deferred` is the only option. Regarding the example provided in the article, I think this snippet can be used instead:
```
function delay(ms){
	return new Promise((resolve, reject) => {
		setTimeout(resolve, ms);
	});
}
// const startTime = new Date();
const p = delay(2000);
// await p;
// const endTime = new Date();
// console.log(endTime - startTime);
```

hayd:
@iammack An example where this allows for an intuitive algorithm is a connection pool, where the async method returns a Connection object, which is then released (put back into the connection pool) once it's been used. If there isn't a Connection object available it waits until one has been released.  (e.g. [here](https://github.com/denodrivers/postgres/blob/c71cc2c330edce75bb47558f49024fcc644b4e54/pool.ts#L86).)

iammack:
Apologies for the late reply, been stuck in some academics related work. I think @benjamingr should weigh in here, he has better knowledge of this :)

benjamingr:
My opinion (was) and still is that going against the standard by not using the provided construction method (the promise constructor) that is designed to be throw-safe and instead using the old (and removed from the spec) Deferred is a mistake.

I would just refactor things to the promise constructor or better - avoid it in most places altogether.

benjamingr:
Of course - Deno is entirely at liberty to introduce its own abstractions rather than use JavaScript's for promise construction. Just in case the above implies differently. It's just kind of weird to pick the variant TC39 ruled out for being unsafe and foot-gunny.

bartlomieju:
> My opinion (was) and still is that going against the standard by not using the provided construction method (the promise constructor) that is designed to be throw-safe and instead using the old (and removed from the spec) Deferred is a mistake.
> 
> I would just refactor things to the promise constructor or better - avoid it in most places altogether.

@benjamingr could you elaborate on this one? What would you suggest instead?

benjamingr:
> @benjamingr could you elaborate on this one? What would you suggest instead?

Sure:

```js
var d = Deferred();
someAsyncOp(() => {
  d.resolve();
});
return d.promise;
```

Becomes:
```
return new Promise((resolve) => someAsyncOp(() => resolve()));
```

Or better - you can call `util.promisify` (from node/std or deno can ship one itself) and just use the promises returning version.

nayeemrmn:
@benjamingr What about, say, [`ServerRequest::done`](https://github.com/denoland/deno/blob/24844a00527f58466b0b057da70b61f8278b84f5/std/http/server.ts#L24), which is [resolved](https://github.com/denoland/deno/blob/24844a00527f58466b0b057da70b61f8278b84f5/std/http/server.ts#L96) in a method?

I'm also against overriding tc39's intentional design without good domain-specific reasons. But I think a lot of the call sites in Deno's codebase would just be replaced by an inline version of `deferred()`.

benjamingr:
> @benjamingr What about, say, ServerRequest::done, which is resolved in a method?

When you explicitly need indirection like that - you have no choice but to pick a way to "escape" the promise constructor. I would not use a `deferred` here I'd use a more specific type that doesn't expose the capability of resolving the promise.
The current API (IIUC) allows consumers to call `request.done.resolve(true)` and cause a problematic state in the code.

Probably just inline it:
 - Make `done` a promise (and not a deferred).
 - Store its `resolve` as a private property on the request.
 - Resolve it (and not the deferred);

Note this is still not great but I'm not sure what to do - I will ask Domenic an some people who were in TC39 when the decision happened what they think.

P.S. why does that code `resolve` with an Error object rather than `reject` to signal an error has occured in that request? Is that because of unhandled rejections?

benjamingr:
What about something like:
 - Create a `detachPromise` type that returns a promise that tracks another promise when it is available.
 - Track the promise of `respond` and resolve it with that.

I don't mind writing some code if the gist of it is unclear.

bartlomieju:

PR #8603, opened 2020-12-03
feat(std/wasi): add support for initializing reactors

caspervonb:
This adds another entry point to Context called initialize for spinning up style modules.

Node has a similar API: https://nodejs.org/api/wasi.html#wasi_wasi_initialize_instance

For context, reactors are modules that don't have a main function and basically run forever in the background.


bartlomieju (approved):
LGTM

Issue #7663, opened 2020-09-25
std/node/fs missing a lot of functionality


AliBasicCoder:
so, the problem is std/node/fs is missing a lot of functions:

important ones are highlighted

1. `fs.createReadStream`
2. `fs.createWriteStream`
3. `fs.fchmod`
4. `fs.chmodSync`
5. fs.fdatasync
6. fs.fdatasyncSync
7. `fs.fstat`
8. `fs.fstatSync`
9. fs.fsync
10. fs.fsyncSync
11. fs.ftruncate
12. fs.ftruncateSync
13. fs.futimes
14. fs.futimesSync
15. `fs.lchmod`
16. `fs.lchmodSync`
17. fs.lutimes
18. fs.lutimesSync
19. `fs.link`
20. `fs.linkSync`
21. `fs.lstat`
22. `fs.lstatSync`
23. fs.mkdtemp
24. fs.mkdtempSync
25. fs.open
26. fs.opendir
27. fs.opendirSync
28. fs.read (with it's to overwrites)
29. `fs.readdir`
30. `fs.readdirSync`
31. fs.readSync (with it's to overwrites)
32. fs.readv
33. fs.readvSync
34. fs.realpath
35. fs.realpath.native
36. fs.realpathSync
37. fs.realpathSync.native
38. `fs.rename`
39. `fs.renameSync`
40. `fs.rmdir`
41. `fs.rmdirSync`
42. `fs.stat`
43. `fs.statSync`
44. `fs.symlink`
45. `fs.symlinkSync`
46. fs.truncate
47. fs.truncateSync
48. `fs.unlink`
49. `fs.unlinkSync`
50. `fs.unwatchFile`
51. fs.utimes
52. fs.utimesSync
53. `fs.watch`
54. `fs.watchFile`
55. fs.write (with it's to overwrites)
56. fs.writeSync (with it's to overwrites)
57. fs.writev
58. fs.writevSync

AliBasicCoder:
i noticed that `Deno.symlink` doesn't exist ??

ry:
symlink is unstable https://doc.deno.land/builtin/unstable#Deno.symlink

AliBasicCoder:
see #7666

Liamolucko:
A lot of these have now been implemented, here's an updated checklist:
- [ ] `fs.createReadStream`
- [ ] `fs.createWriteStream`
- [ ] fs.fdatasync
- [ ] fs.fdatasyncSync
[...]

PR #8604, opened 2020-12-03
feat(std/node): adds fs.mkdtemp & fs.mkdtempSync

lp:
#7663 
 Implements Nodejs's fs.mkdtemp & fs.mkdtempSync using Deno.makeTempDir & Deno.makeTempDirSync.

[https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback](https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback)

 The compatibility layer is not 100% exact.

- Deno.makeTempDir creates the directory name using 8 random characters, Nodejs uses 6 random characters.
- Deno.makeTempDir does not support choosing the directory name string encoding.  The `options` argument will be ignored.
- The `err` passed to the callback or the error raised may be somewhat different than Nodejs's.  This follows what seems to be the general error handling pattern for other std/node/fs implementations.

Even with those pitfalls, the implementations remains largely compatible with the majority of use cases.


bartlomieju:
@Soremwar PTAL

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!callback) throw new Error("No callback function supplied");
+
+  Deno.makeTempDir({ dir: prefix })
+    .then((directory) => callback(undefined, directory))
+    .catch((error) => callback(error));
+}
+
+// TODO - 'encoding' handling needs implementation in Deno.makeTempDirSync
+export function mkdtempSync(
```
Soremwar, line 31: Ditto
https://nodejs.org/dist/latest-v15.x/docs/api/fs.html#fs_fs_mkdtempsync_prefix_options

Soremwar (changes requested):

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,36 @@
...
+  prefix: string,
+  optionsOrCallback: { encoding: string } | string | mkdtempCallback,
+  maybeCallback?: mkdtempCallback,
+): void {
+  const callback: mkdtempCallback | undefined =
+    optionsOrCallback instanceof Function ? optionsOrCallback : maybeCallback;
+
+  if (!callback) throw new Error("No callback function supplied");
+
+  Deno.makeTempDir({ dir: prefix })
```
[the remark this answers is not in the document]
lp: @Soremwar agreed and done in the latest commit to this PR.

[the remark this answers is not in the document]
lp: @Soremwar agreed and done in the latest commit.

bartlomieju:
@Soremwar can you give it another pass?

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,83 @@
...
+    const decoder = new TextDecoder(encoding);
+    const encoder = new TextEncoder();
+    return decoder.decode(encoder.encode(str));
+  }
+}
+
+function tempDirPath(prefix: string): string {
+  let path: string;
+  do {
+    path = prefix + Math.random().toString(36).substring(2, 8);
```
Soremwar, line 79: This won't generate folder names with uppercase characters, unlike original implementation

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,83 @@
...
+  callback: mkdtempCallback,
+): void;
+export function mkdtemp(
+  prefix: string,
+  optionsOrCallback: { encoding: string } | string | mkdtempCallback,
+  maybeCallback?: mkdtempCallback,
+): void {
+  const callback: mkdtempCallback | undefined =
+    optionsOrCallback instanceof Function ? optionsOrCallback : maybeCallback;
+  if (!callback) throw new Error("No callback function supplied");
```
Soremwar, line 23: Nit pick, use ERR_INVALID_CALLBACK here like the original implementation, you can find it `std/node/_errors.ts`

Soremwar (changes requested):

Soremwar:
I don't think that you need to pass a length around to calculate the random name since you won't use it anywhere else, but that's about it

LGTM

std/node/_fs/_fs_mkdtemp_test.ts
```diff
@@ -0,0 +1,133 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
```
benjamingr, line 1: If this uses code from Node please add the copyright too? (I think there is already process for that?)

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,92 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
```
benjamingr, line 1: If this contains code from Node probably add that copyright? (there is already a process for that I think?)

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,92 @@
...
+): void {
+  const callback: mkdtempCallback | undefined =
+    optionsOrCallback instanceof Function ? optionsOrCallback : maybeCallback;
+  if (!callback) throw new ERR_INVALID_CALLBACK(callback);
+
+  const encoding: string | undefined = parseEncoding(optionsOrCallback);
+  const path = tempDirPath(prefix);
+
+  Deno.mkdir(path, { recursive: false, mode: 0o700 })
+    .then(() => callback(undefined, decode(path, encoding)))
```
benjamingr, line 33: This patterns is problematic, if `callback` throws you reach the catch and the callback is called again with an error - doing `.then(() => callback(...), callback)` is probably safer or using `util.callbackify` if Deno already has it.

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,92 @@
...
+  options: { encoding: string } | string,
+  callback: mkdtempCallback,
+): void;
+export function mkdtemp(
+  prefix: string,
+  optionsOrCallback: { encoding: string } | string | mkdtempCallback,
+  maybeCallback?: mkdtempCallback,
+): void {
+  const callback: mkdtempCallback | undefined =
+    optionsOrCallback instanceof Function ? optionsOrCallback : maybeCallback;
```
benjamingr, line 26: `instanceof` will fail on different realms, better to use `typeof` here.

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (!callback) throw new ERR_INVALID_CALLBACK(callback);
```
benjamingr, line 27: I think this is Node's behaviour?

std/node/_fs/_fs_mkdtemp_test.ts
```diff
@@ -0,0 +1,133 @@
...
+const prefix = isWindows ? env.TEMP + "\\" : (env.TMPDIR || "/tmp") + "/";
+const doesNotExists = "/does/not/exists/";
+const options = { encoding: "ascii" };
+const badOptions = { encoding: "bogus" };
+
+Deno.test({
+  name: "[node/fs] mkdtemp",
+  fn: async () => {
+    const result = await new Promise((resolve) => {
+      mkdtemp(prefix, (err, directory) => {
```
benjamingr, line 17: I would test this with `promisify` rather than all the `new Promise`s

std/node/_fs/_fs_mkdtemp.ts
lp: It seems like the problematic pattern is used in many other compatibility implementations... __fs_appendFile, _fs_chmod, _fs_chown, _fs_copy, _fs_dir, _fs_exists, _fs_exists, _fs_lstat, _fs_mkdir_ ... to only name a few.

I don't understand what is the safer solution proposal ` .then(() => callback(...), callback)` ?

benjamingr: Yes, and then it is also wrong and very dangerous in all those places 😅

lp: Has any ticket been opened about that improvement?

std/node/_fs/_fs_mkdtemp.ts
lp: indeed

std/node/_fs/_fs_mkdtemp.ts
benjamingr: I don't know I'm not a maintainer, just a contributor that responded to the "please review this" ping on discord since I'm a node maintainer - maybe Bert would know :]

std/node/_fs/_fs_mkdtemp.ts
```diff
@@ -0,0 +1,100 @@
...
+  options: { encoding: string } | string,
+  callback: mkdtempCallback,
+): void;
+export function mkdtemp(
+  prefix: string,
+  optionsOrCallback: { encoding: string } | string | mkdtempCallback,
+  maybeCallback?: mkdtempCallback,
+): void {
+  const callback: mkdtempCallback | undefined =
+    typeof (optionsOrCallback) == "function"
```
benjamingr, line 27: nit: typeof doesn't need `()`s?

std/node/_fs/_fs_mkdtemp.ts
lp: I'll fill in a ticket about this.  Thanks for the insight @benjamingr.

bartlomieju:
@benjamingr @Soremwar do you think this PR is ready to land?

benjamingr (approved):
This approach using the existing Node functions looks much cleaner than the previous iteration.

I'm sure there will be edge cases and iterations but as far as I can tell @bartlomieju this is ready to land.

bartlomieju (approved):
LGTM, thank you @lp, great contribution!

PR #8605, opened 2020-12-03
feat(std/wasi): add return on exit option

caspervonb:
This adds an exitOnReturn option to context making it possible to unwind the stack on the exit(2) syscall instead of delegating to it directly.

Node has a similar option, https://nodejs.org/api/wasi.html#wasi_wasi_initialize_instance and my web implementation of WASI unwinds by default.

Use case is being able to treat WASI execution contexts as children that don't kill the parent on exit.


bartlomieju (approved):
LGTM

PR #8611, opened 2020-12-04
fix(cli) fix close websocket with code is not working

Preta-Crowz:
fixed condition to raise error

### DIFF
--- a/cli/rt/27_websocket.js
+++ b/cli/rt/27_websocket.js
@@ -201,7 +201,7 @@
     }
 
     close(code, reason) {
-      if (code && !(code === 1000 || (3000 <= code && code < 5000)))
+      if (code && !(code === 1000 || (3000 <= code && code < 5000))) {
         throw new DOMException(
           "The close code must be either 1000 or in the range of 3000 to 4999.",
           "NotSupportedError",
--- a/cli/rt/27_websocket.js
+++ b/cli/rt/27_websocket.js
@@ -201,7 +201,7 @@
     }
 
     close(code, reason) {
-      if (code && (code !== 1000 && !(3000 <= code > 5000))) {
+      if (code && (code !== 1000 && !(3000 <= code < 5000))) {
         throw new DOMException(
           "The close code must be either 1000 or in the range of 3000 to 4999.",
           "NotSupportedError",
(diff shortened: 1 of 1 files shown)


cli/rt/27_websocket.js
```diff
@@ -201,7 +201,7 @@
     }
 
     close(code, reason) {
-      if (code && (code !== 1000 && !(3000 <= code > 5000))) {
+      if (code && (code !== 1000 && !(3000 <= code < 5000))) {
```
nayeemrmn, line 204:
```suggestion
      if (code && code !== 1000 && !(code >= 3000 && code < 5000)) {
```

cli/rt/27_websocket.js
```diff
@@ -201,7 +201,7 @@
     }
 
     close(code, reason) {
-      if (code && (code !== 1000 && !(3000 <= code > 5000))) {
+      if (code && code !== 1000 && !(code >= 3000 && code < 5000)) {
```
ry, line 204: Thanks for the PR - I'm not sure how the previous code even got through lint!

I think it would be better if this was written:
`if (code === 1000 || (3000 <= code && code < 5000))`

Preta-Crowz: sure. I editted

Preta-Crowz:
wait, why lint is failed this time

crowlKats:
@Preta-Crowz you removed the opening curly bracket.

Preta-Crowz:
i see

Preta-Crowz:
fixed

bartlomieju:
Can we get a test exercising problematic bit?

Preta-Crowz:
> Can we get a test exercising problematic bit?

on current deno's code, condition will compare to wrong value.
also, maybe my first commit can have same issue.

here's example.
`a < b < c`
on this code, it will compare a and b first.
it's right. but next step is wrong. it doesn't compare b and c.
it will compare the value of `a < b` and c.
so if c is bigger than 1, it's always true.

yes. my code was wrong too.

bartlomieju:
@Preta-Crowz good catch! Seems you got the the bottom of the problem.

Preta-Crowz:
recreated to #8776

Issue #8610, opened 2020-12-04
Rename functions on std/bytes and improve bytes.ts/concat


yos1p:
I would like to work on some of the list in the [std-group meeting notes](https://github.com/denoland/deno/issues/8405#issuecomment-738162035)

What I want to work on:

- `concat`
    - [x]  Use [rest parameters](https://www.typescriptlang.org/docs/handbook/functions.html#rest-parameters) to allow more than 2 parameters
    - [x]  Unit test with multiple parameters

- `equals`
    - [x]  Rename parameter to `a` and `b`

- `repeat`
    - [x]  Throw `RangeError` when count is negative

- `indexOf`
    - [x]  Take start index as optional arg
    - [x]  Update doc

- `lastIndexOf`
    - [x]  Take start index as optional arg
    - [x]  Update doc

- Rename functions
     - [x]  From `copyBytes` to `copy`
     - [x]  From `findIndex` to `indexOf`
     - [x]  From `findLastIndex` to `lastIndexOf`
     - [x]  From `equal` to `equals`
     - [x]  From `hasPrefix` to `startsWith`
     - [x]  From `hasSuffix` to `endsWith`
     - [x]  From `hasPrefix` to `startsWith`

Let me know what you guys think.

PR #8612, opened 2020-12-04
modify(std/bytes): Rename functions, Update docs, Modify bytes.concat, bytes.indexOf, and bytes.lastIndexOf

yos1p:
Issue: #8610 

I'm just going to drop the code here. Feel free to let me know what else we can do to improve

### DIFF
--- a/std/bytes/mod.ts
+++ b/std/bytes/mod.ts
@@ -126,7 +126,7 @@ export function repeat(origin: Uint8Array, count: number): Uint8Array {
  */
 export function concat(a: Uint8Array, ...buf: Uint8Array[]): Uint8Array {
   let totalLength = a.length;
-  for (let b of buf) {
+  for (const b of buf) {
     totalLength += b.length;
   }
 
@@ -134,7 +134,7 @@ export function concat(a: Uint8Array, ...buf: Uint8Array[]): Uint8Array {
   output.set(a, 0);
 
   let lastIndex = a.length;
-  for (let b of buf) {
+  for (const b of buf) {
     output.set(b, lastIndex);
     lastIndex += b.length;
   }
--- a/std/bytes/mod.ts
+++ b/std/bytes/mod.ts
@@ -51,13 +51,13 @@ export function lastIndexOf(source: Uint8Array, pat: Uint8Array): number {
 }
 
 /** Check whether binary arrays are equal to each other.
- * @param source first array to check equality
- * @param match second array to check equality
+ * @param a first array to check equality
+ * @param b second array to check equality
  */
-export function equals(source: Uint8Array, match: Uint8Array): boolean {
-  if (source.length !== match.length) return false;
-  for (let i = 0; i < match.length; i++) {
-    if (source[i] !== match[i]) return false;
+export function equals(a: Uint8Array, b: Uint8Array): boolean {
+  if (a.length !== b.length) return false;
+  for (let i = 0; i < b.length; i++) {
+    if (a[i] !== b[i]) return false;
   }
   return true;
 }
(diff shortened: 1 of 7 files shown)


std/bytes/mod.ts
```diff
@@ -98,36 +98,47 @@ export function repeat(origin: Uint8Array, count: number): Uint8Array {
...
- * @param b array to concatenate with origin
+/** Concatenate multiple binary arrays and return new one.
+ * @param origin binary array to concatenate
+ * @param buf binary arrays to concatenate with origin
  */
-export function concat(origin: Uint8Array, b: Uint8Array): Uint8Array {
-  const output = new Uint8Array(origin.length + b.length);
-  output.set(origin, 0);
-  output.set(b, origin.length);
+export function concat(a: Uint8Array, ...buf: Uint8Array[]): Uint8Array {
```
bartlomieju, line 127: I believe there should be only a single argument to this function (`...buf`)

yos1p: Sounds good to me as well. I will make the changes.

wperron (approved):
LGTM - I believe we also discussed whether or not to rename some of the input parameters as well, not something blocking in my opinion though

std/bytes/mod.ts
```diff
@@ -77,7 +77,7 @@ export function hasPrefix(source: Uint8Array, prefix: Uint8Array): boolean {
  * @param source source array
  * @param suffix suffix array to check in source
  */
```
ry, line 79: Can you mention the RangeError in the JSDoc?

std/bytes/test.ts
```diff
@@ -111,6 +111,19 @@ Deno.test("[bytes] concat empty arrays", () => {
...
+  const u2 = encode("W");
+  const u3 = encode("o");
+  const u4 = encode("r");
+  const u5 = encode("l");
+  const u6 = encode("d");
+  const joined = concat(u1, u2, u3, u4, u5, u6);
+  assertEquals(decode(joined), "Hello World");
+  assert(u1 !== joined);
+  assert(u2 !== joined);
+});
```
ry, line 125: 👍

ry (commented):
Thank you @yos1p

std/bytes/test.ts
```diff
@@ -124,15 +124,15 @@ Deno.test("[bytes] concat multiple arrays", () => {
...
-Deno.test("[bytes] contain", () => {
+Deno.test("[bytes] contains", () => {
   const source = encode("deno.land");
   const pattern = encode("deno");
   assert(contains(source, pattern));
 
   assert(contains(new Uint8Array([0, 1, 2, 3]), new Uint8Array([2, 3])));
 });
 
-Deno.test("[io/tuil] copyBytes", function (): void {
```
yos1p, line 122: Anyone know why this is [io/util]? Is it on purpose?

yos1p: I rename it to [bytes] given the preceding functions. But let me know if I should revert this.

std/bytes/mod.ts
yos1p: Yes I can do that

yos1p:
I also added changes for indexOf/lastIndexOf to take start index as optional parameter. The start index for both functions is based on Array.indexOf and Array.lastIndexOf from [javascript reference document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf).

I think it will make it easier what will be expected by JS developers. But let me know if that should be changed (e.g throw RangeError instead)

std/bytes/test.ts
bartlomieju: That's okay @yos1p; IIRC this method was in `std/io/util.ts` at some point

ry (approved):
LGTM - thank you @yos1p

PR #8617, opened 2020-12-04
refactor(cli/flags): use derive macro

crowlKats:
Depends on #8485

This PR refactors cli/flags to sue the new derive macro from clap.


crowlKats:
To set the `about` (previously called `help`), you either set `about` & `long_about`, or you use doc comments with the following rule:
anything before a blank line will count to `about`, whereas everything (be it before and after the first blank line) will count to `long_about`. 
Or a mixture of settings `long_about` & doc comment.

Preferably I'd only use the doc comment approach, but that means we'd have to slightly change the current about messages.

kitsonk:
Is there a compelling reason to use beta software instead of waiting for the clap 3.0 release? I don't really want to be someone else's experiment if we don't have a requirement/need.

nayeemrmn:
@kitsonk I believe this PR is not intended to land before the release: https://discord.com/channels/684898665143206084/684911491035430919/784442970530250753.

bartlomieju:
Ref https://github.com/denoland/deno/pull/8617

As per Discord conversation; we'll land this PR after stable release of Clap v3. Closing for now.

PR #8619, opened 2020-12-04
(fix): performance.now() should be a minimal op to avoid JSON deserialization overhead

### LINKED ISSUE #8548: performance.now() should be a minimal op to avoid JSON deserialization overhead
Ref https://github.com/denoland/deno/issues/8543#issuecomment-735474718

Soremwar:
Closes #8548

### DIFF
--- a/cli/ops/timers.rs
+++ b/cli/ops/timers.rs
@@ -151,7 +151,7 @@ fn op_now(
   match zero_copy.len() {
     0 => return MinimalOp::Sync(Err(type_error("no buffer specified"))),
     1 => {}
-    _ => panic!("Invalid number of arguments"),
+    _ => return MinimalOp::Sync(Err(type_error("Invalid number of arguments"))),
   }
 
   let op_state = state.borrow();
--- a/cli/ops/timers.rs
+++ b/cli/ops/timers.rs
@@ -23,7 +23,7 @@ use deno_core::serde_json::json;
 use deno_core::serde_json::Value;
 use deno_core::ZeroCopyBuf;
 use serde::Deserialize;
-use std::cell::RefCell;
+use std::{cell::RefCell};
 use std::future::Future;
 use std::pin::Pin;
 use std::rc::Rc;
@@ -146,7 +146,7 @@ fn op_now(
   // Arguments are discarded
   _sync: bool,
   _x: i32,
-  zero_copy: BufVec,
+  mut zero_copy: BufVec,
 ) -> MinimalOp {
   match zero_copy.len() {
     0 => return MinimalOp::Sync(Err(type_error("no buffer specified"))),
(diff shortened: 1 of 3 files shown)


Soremwar:
A million iterations

Minimal OP

`performance.now` - Time elapsed: 2335 ms
`performance.mark` - Time elapsed: 3388 ms

JSON OP

`performance.now` - Time elapsed: 5906 ms
`performance.mark` - Time elapsed: 7564 ms

cli/ops/timers.rs
```diff
@@ -77,7 +81,7 @@ pub fn init(rt: &mut deno_core::JsRuntime) {
   super::reg_json_sync(rt, "op_global_timer_stop", op_global_timer_stop);
   super::reg_json_sync(rt, "op_global_timer_start", op_global_timer_start);
   super::reg_json_async(rt, "op_global_timer", op_global_timer);
-  super::reg_json_sync(rt, "op_now", op_now);
+  rt.register_op("op_now", metrics_op(minimal_op(op_now)));
```
bartlomieju, line 84: This is debatable, maybe we shouldn't take metrics of `op_now`?

cli/ops/timers.rs
```diff
@@ -138,26 +142,36 @@ async fn op_global_timer(
...
+  state: Rc<RefCell<OpState>>,
+  // Arguments are discarded
+  _sync: bool,
+  _x: i32,
+  mut zero_copy: BufVec,
+) -> MinimalOp {
+  match zero_copy.len() {
+    0 => return MinimalOp::Sync(Err(type_error("no buffer specified"))),
+    1 => {}
+    _ => panic!("Invalid number of arguments"),
```
bartlomieju, line 154: Replace with error

bartlomieju (commented):
@Soremwar looks good!

cli/ops/timers.rs
Soremwar: The wonders of copy-pasta
Will fix

cli/ops/timers.rs
Soremwar: RangeError or custom error?

cli/ops/timers.rs
bartlomieju: @Soremwar this is ok, I'm just wondering if we add additional overhead to the `op_now`. Maybe let's keep it for now and address it if other performance problems occur

cli/ops/timers.rs
bartlomieju: I think you could use `type_error`

Soremwar:
Removing the extra Uint8array actually made a big difference :O

### New times: 

#### Minimal OP v1 (Extra Uint8Array)

performance.now - Time elapsed: 2335 ms
performance.mark - Time elapsed: 3388 ms

#### Minimal OP v2 (Clean)

performance.now - Time elapsed: 1677 ms
performance.mark - Time elapsed: 2921 ms

#### JSON OP

performance.now - Time elapsed: 5906 ms
performance.mark - Time elapsed: 7564 ms

Soremwar:
Stll quite slow compared to Node, I might look inside to check what magic they are doing to achieve those result, but for all practical purposes this is solved

bnoordhuis:
@Soremwar Node.js uses V8 fast API calls (https://github.com/nodejs/node/pull/33600). I have a proof-of-concept in https://github.com/denoland/rusty_v8/pull/511 but it's not anywhere near ready.

kitsonk:
Yeah, so this improvement is great for now until we can support fast API.

ry (approved):
Nice work @Soremwar - LGTM

This is a very encouraging example of why we should support FastAPI. Ultimately it would be awesome if all of our ops could use it.

PR #8621, opened 2020-12-05
test(std/wasi): annotate proc_exit unreachable

caspervonb:
This adds a minor annotation to the unreachable after wasi::proc_exit to be be able to distinguish it from unreachable errors that may happen during pre-open handling.


caspervonb:
Ehh, looks like we're already there and I just had out of date local state.

PR #8622, opened 2020-12-05
refactor(std/wasi): prefer explicit encoder/decoder names

caspervonb:
This renames a couple of identifiers from the ambiguous name "text" to the more explicit textEncoder and textDecoder depending on what they are.

Should have been done in the initial pass to camelCase all the identifiers.


bartlomieju (approved):
LGTM

PR #8624, opened 2020-12-05
fix(std/encoding/toml): Rewrite toml parser not to use eval()

### LINKED ISSUE #8433: (std/encoding/toml) date parse error
```ts
import { parse } from 'https://cdn.jsdelivr.net/gh/denoland/deno@std/0.78.0/std/encoding/toml.ts';
console.log(parse(`sign='2020-01-01x'`));
```

will output
`
{ sign: Invalid Date }
`

but it  should output
`
{ sign: "2020-01-01x" }
`

Is this my mistake?

uki00a:
## Changes

* Rewrites toml parser not to use `eval()` (that would probably be a security risk).
* This PR also fixes #8433.
* Adds support for parsing unicode characters (`\uXXXX`, `\UXXXXXX`).


bartlomieju (approved):
LGTM, thanks for the fix @uki00a

PR #8139, opened 2020-10-27
docs: improve manual around typescript

kitsonk:
With the refactor of module graph infrastructure, it has been feeling like a right time to improve the manual around Deno and TypeScript.

Current work in progress.

PR #8629, opened 2020-12-06
chore(std): Remove tsconfig_test.json

nayeemrmn:
Missed in #8050. Also updates doc surrounding `isolatedModules` and `--no-check`.

### DIFF
--- a/docs/getting_started/typescript.md
+++ b/docs/getting_started/typescript.md
@@ -23,23 +23,21 @@ useful when type checking is provided by your editor and you want startup time
 to be as fast as possible (for example when restarting the program automatically
 with a file watcher).
 
-Because `--no-check` does not do TypeScript type checking we can not
-automatically remove type only imports and exports as this would require type
-information. For this purpose TypeScript provides the
-[`import type` and `export type` syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-exports).
-To export a type in a different file use
-`export type { AnInterface } from "./mod.ts";`. To import a type use
-`import type { AnInterface } from "./mod.ts";`. You can check that you are using
-`import type` and `export type` where necessary by setting the `isolatedModules`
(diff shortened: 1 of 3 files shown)


docs/getting_started/typescript.md
```diff
@@ -23,23 +23,21 @@ useful when type checking is provided by your editor and you want startup time
...
-Because there is no type information when using `--no-check`, `const enum` is
-not supported because it is type-directed. `--no-check` also does not support
-the legacy `import =` and `export =` syntax.
+To make the most of skipping type checks, `--no-check` transpiles each module in
+isolation without using information from imported modules. This maximizes
+potential for concurrency and incremental rebuilds. On the other hand, it cannot
+know if `export { Foo } from "./foo.ts"` should be preserved (in case `Foo` is a
+value) or removed (in case `Foo` is strictly a type) in the output JS. So Deno
+_without_ `--no-check` requires that all TS complies with
+[`isolatedModules`](https://www.typescriptlang.org/tsconfig#isolatedModules)
```
bartlomieju, line 32: I think `isolatedModules` is required even with type checking?

nayeemrmn: > I think `isolatedModules` is required even with type checking?

Yes I wrote "without `--no-check`". In fact it's _only_ required with type checking, but that makes it `--no-check` compatible.

bartlomieju: @nayeemrmn I see; I'm concerned about wording here - it suggests that `isolatedModules` is required only in a specific scenario while in fact it's always required.

nayeemrmn: Reworded.

kitsonk: The whole TypeScript section needs to be re-written.  It is on my back burner under #8139.  I think what is there could be a bit confusing, but is accurate and sufficient for the time being.

kitsonk (approved):
LGTM

PR #8631, opened 2020-12-06
feat: stabilize Deno.utime and Deno.utimeSync

### LINKED ISSUE #8630: Stabilize Deno.utime and Deno.utimeSync
This topic was briefly discussed during recent Standard Library Working Group meeting. Stabilizing these two APIs should make a few APIs from `std/fs` stable. There have been no changes to `Deno.utime` in quite some time so I believe we could safely stabilize it for `1.6.0`

bartlomieju:
Closes #8630

### DIFF
--- a/cli/diagnostics.rs
+++ b/cli/diagnostics.rs
@@ -79,8 +79,6 @@ const UNSTABLE_DENO_PROPS: &[&str] = &[
   "systemCpuInfo",
   "transpileOnly",
   "umask",
-  "utime",
-  "utimeSync",
 ];
 
 lazy_static! {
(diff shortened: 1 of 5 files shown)


ry (approved):
LGTM - I don't think there are any security issues regarding high precision time, since this is setting not read...

caspervonb:
So I allowed for millisecond precision in a PR a while back, but this ended up being super weird because the number overload still only gives you a precision of one second.

bartlomieju:
Closing in favor of https://github.com/denoland/deno/issues/8634

PR #8633, opened 2020-12-06
refactor(cli): remove Option from Flags.v8_flags

crowlKats:
Currently `Flags.v8_flags` is `Option<Vec<String>>`, but this can be simplified to just `Vec<String>` and checking for `is_empty()`


bartlomieju (approved):
LGTM

ry (commented):
Nice clean up - thanks!

PR #8639, opened 2020-12-06
fix(coverage): ignore comments

### LINKED ISSUE #7652: Coverage output quirks

caspervonb:
Ignores comments when tallying up line counts.

Closes #7652


bartlomieju:
@caspervonb if I'm not mistaken after this PR we'd still be counting block comments `/** */`? I think we'll need some more robust solution possibly involving SWC to parse out which lines in source code are not comments, wdyt?

caspervonb:
> @caspervonb if I'm not mistaken after this PR we'd still be counting block comments `/** */`? I think we'll need some more robust solution possibly involving SWC to parse out which lines in source code are not comments, wdyt?

Yep only single line comments being explicitly ignored at the moment (draft), will probably go with swc's token stream similar to the repl validator.

caspervonb:
Rewrote this to use the token lexer instead of trying to manually scan for comments.
Just need to come up with some tests.

bartlomieju (approved):
LGTM, nice fix Casper

PR #8640, opened 2020-12-06
refactor: deno_runtime crate

### LINKED ISSUE #2633: Rust library with batteries included
In #1209 you introduced a deno crate which allows creating and working with a basic Isolate. This is definitely usable, but it would be very helpful to have the ability to "turn on" Deno's standard library from the cli package to be available in the deno crate along with maintaining the ability to get an "empty" `Isolate` as it is now. For some cases you need a standard library, for some you don't.

---

My use case: I'm trying to build a FaaS on top of Deno, and I would like to have some parts of the cli package available in deno crate:

- The TypeScript compiler
- Fetch API
- Timers
- Dynamic module import

But it would be useful to be able to turn on everything else selectively too.

### LINKED ISSUE #7928: Feature request: CLI available as library
I'm working on a project where I'd like to instantiate multiple instances of the full Deno runtime from within Rust code. Because the `cli` package only provides a binary, that isn't currently possible. I'd like to avoid the inefficiencies of shelling out to multiple separate Deno processes, and instead spin them up in Rust threads.

Adding a `[lib]` target to `Cargo.toml` in the `cli` package would fix this, but it would probably require changes to the package to expose a Rust API with all of the CLI options.

bartlomieju:
This commit moves Deno JS runtime, ops, permissions and
inspector implementation to new "deno_runtime" crate.

Details in "runtime/README.md".

There is still some duplication of code, but overall diff is quite
small after all the recent refactors.

Closes #7928
Closes #2633


deno_runtime/README.md
```diff
@@ -0,0 +1,10 @@
+# TODO(bartlomieju):
+
+# `deno_runtime` crate
+
+[![crates](https://img.shields.io/crates/v/deno.svg)](https://crates.io/crates/deno)
+[![docs](https://docs.rs/deno/badge.svg)](https://docs.rs/deno)
+
+This provides the actual deno executable and the user-facing APIs.
+
+The deno crate uses the deno_core to provide the executable.
```
ry, line 10: The links are wrong.

It would be v useful to have a good description of the purpose and scope of this crate here.

bartlomieju: I've updated the README, but don't really feel like writing prose today 😅

deno_runtime/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+# Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
```
ry, line 1: Rename the top-level directory to "runtime" or "rt" but not "deno_runtime"

cli/Cargo.toml
```diff
@@ -38,6 +38,7 @@ deno_doc = "0.1.18"
 deno_fetch = { path = "../op_crates/fetch", version = "0.13.0" }
 deno_lint = "0.2.13"
 deno_web = { path = "../op_crates/web", version = "0.21.0" }
```
ry, line 40: Why do the various op crates (deno_web, deno_crypto, etc) need to be dependencies of cli? Don't they come from deno_runtime?

ry (commented):
It would be really great to have a `deno_runtime/examples` directory with a basic hello world example.

cli/build.rs
```diff
@@ -43,14 +43,6 @@ fn create_snapshot(
   println!("Snapshot written to: {} ", snapshot_path.display());
 }
```
ry, line 45: Higher up in this file the method `create_snapshot` calls 

```
  deno_web::init(&mut js_runtime);
  deno_fetch::init(&mut js_runtime);
  deno_crypto::init(&mut js_runtime);
```

Are these necessary for the compiler snapshot?

bartlomieju: Yes, declaration files are pulled into TSC during build

cli/Cargo.toml
bartlomieju: Reexported from `deno_runtime`

cli/build.rs
ry: I think only the `get_declaration()` methods are necessary. Maybe some of the event stuff from deno_web. But I don't think tsc calls fetch.

bartlomieju:
> It would be really great to have a `deno_runtime/examples` directory with a basic hello world example.

Added `hello_runtime` example in `runtime/examples/`

runtime/examples/hello_runtime.rs
```diff
@@ -0,0 +1,55 @@
...
+    create_web_worker_cb,
+    attach_inspector: false,
+    maybe_inspector_server: None,
+    should_break_on_first_statement: false,
+    module_loader,
+    runtime_version: "x".to_string(),
+    ts_version: "x".to_string(),
+    no_color: false,
+    get_error_class_fn: Some(&get_error_class_name),
+  };
```
ry, line 42: Use `..Default::default()` here

bartlomieju: Not possible, `Default` is not implemented on `WorkerOptions`

cli/build.rs
bartlomieju: I've removed `init` calls

runtime/examples/hello_runtime.rs
```diff
@@ -0,0 +1,55 @@
...
+  let js_path =
+    Path::new(env!("CARGO_MANIFEST_DIR")).join("examples/hello_runtime.js");
+  let main_module = ModuleSpecifier::resolve_path(&js_path.to_string_lossy())?;
+  let permissions = Permissions::allow_all();
+
+  let mut worker =
+    MainWorker::from_options(main_module.clone(), permissions, &options);
+  worker.bootstrap(&options);
+  worker.execute_module(&main_module).await?;
+  worker.run_event_loop().await?;
```
ry, line 53: Aside: we should change the name of `run_event_loop()` - it's neither accurate nor descriptive. The event loop is run in tokio. This method waits until all async ops have completed.

bartlomieju: We've already talked about it, ops are only one of a few things that are polled in `run_event_loop`, besides ops there are pending module loads and pending dynamic imports.

runtime/build.rs
```diff
@@ -0,0 +1,79 @@
...
+    println!("cargo:rerun-if-changed={}", file.display());
+    let display_path = file.strip_prefix(display_root).unwrap();
+    let display_path_str = display_path.display().to_string();
+    js_runtime
+      .execute(
+        &("deno:".to_string() + &display_path_str.replace('\\', "/")),
+        &std::fs::read_to_string(&file).unwrap(),
+      )
+      .unwrap();
+  }
```
ry, line 30: There is now a lot of duplicated logic between cli/build.rs and runtime/build.rs

Ideally these could be factored out into deno_core so they can be shared. I think it would be fine to postpone that work, but maybe add some TODOs.

runtime/build.rs
```diff
@@ -0,0 +1,79 @@
...
+  let o = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+
+  // Main snapshot
+  let runtime_snapshot_path = o.join("CLI_SNAPSHOT.bin");
+
+  let js_files = get_js_files("rt");
+  create_runtime_snapshot(&runtime_snapshot_path, js_files);
+}
+
+fn get_js_files(d: &str) -> Vec<PathBuf> {
```
ry, line 67: This method is only called once, "d" is always "rt".

runtime/examples/hello_runtime.rs
ry: `wait_for_various_async_things_to_complete()` then or `wait_until_idle()`

runtime/ops/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+
+pub fn reg_json_sync<F>(rt: &mut JsRuntime, name: &'static str, op_fn: F)
+where
+  F: Fn(&mut OpState, Value, &mut [ZeroCopyBuf]) -> Result<Value, AnyError>
+    + 'static,
+{
+  rt.register_op(name, metrics_op(json_op_sync(op_fn)));
+}
+
+pub struct UnstableChecker {
```
ry, line 56: Maybe add a comment about why this is a struct. (so that it can be stored in the gotham state)

runtime/examples/hello_runtime.rs
bartlomieju: `wait_until_idle()` sounds ok, but let's do that in other PR

runtime/build.rs
bartlomieju: I'll add TODO

runtime/build.rs
bartlomieju: Yes, it's copy paste from `cli/build.rs`, I suggest we do those minor cleanups in follow up PRs

runtime/ops/mod.rs
bartlomieju: I'll add a comment

cli/build.rs
```diff
@@ -13,6 +13,8 @@ use std::env;
 use std::path::Path;
 use std::path::PathBuf;
 
+// TODO(bartlomieju): this module contains a lot of duplicated
+// logic with `cli/build.rs`, factor out to `deno_core`.
```
ry, line 17: should be runtime/build.rs

ry (approved):
LGTM - nice work.

Issue #7648, opened 2020-09-23
A cover command and cross process code coverage


caspervonb:
I've mentioned a couple of times in Discord that I'm leaning towards a `cover` command so time to elaborate a bit on what I mean before slinging patches.

A really common test scenario involves spawning processes via Deno.run inside a test.

But the coverage collector as is right now only works for in-memory for a single Deno process and only for the test command which is not ideal as there is no way to get coverage data from say `Deno.run({cmd: Deno.execPath()})` with this approach.

So for this to work; the most obvious and simple way forward would be for the coverage collector to persist to disk so that it can be read back later by a process higher up the hierachery and be processed by a coverage reporter.

This implies that at-least `deno run` will be given also be given coverage flag in addition to `deno test` which in turn means that the coverage option strictly only means collect coverage.

So if the coverage flag is on both run and test with the intent of just collect coverage we no longer have anywhere (logical) to place a reporter which is where a cover command would come to serve as just a reporter with flags for X, Y and Z formats.

caspervonb:
Okay so I've decided to go with an environment variable to signal cross-process coverage collection, it's the only thing that can survive going through say a hypothetical convoluted process of running a bash shell that spawns a zsh shell that spawns a node process that spawns a lua process that spawns a deno process that should be covered.

I'll be doing this in atomic commits.

[x] Make deno test --coverage <dir> set a the collection environment variable, make it check both the flag and environment variable to determine if it should be running with coverage collection enabled.
[ ] Make deno run check if the environment variable is set, and collect coverage information if it is.
[ ] Move reporting into a cover command so that you don't have to collect every time you want a different report and so that the output formats do not clash with future test report output formats.

PR #8642, opened 2020-12-07
feat(unstable): record raw coverage into a directory

caspervonb:
This records all coverage information into a directory to allow for collecting coverage in a multi-process setup like spawning another deno process (`deno test` calling `deno test` will propagate and merge correctly).

This does not implement coverage collection for `deno run` yet but can and will be added in a follow-up.

Ties into https://github.com/denoland/deno/issues/7648


cli/tools/coverage.rs
```diff
@@ -212,10 +231,24 @@ impl PrettyCoverageReporter {
...
+  let mut coverages: Vec<Coverage> = Vec::new();
+
+  let entries = fs::read_dir(dir)?;
+  for entry in entries {
+    let json = fs::read_to_string(entry.unwrap().path())?;
+    let coverage: Coverage = serde_json::from_str(&json)?;
+
+    coverages.push(coverage);
+  }
+
```
caspervonb, line 244: Flatten duplicates.

cli/flags.rs
```diff
@@ -606,10 +606,9 @@ fn test_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   let allow_none = matches.is_present("allow-none");
   let quiet = matches.is_present("quiet");
   let filter = matches.value_of("filter").map(String::from);
-  let coverage = matches.is_present("coverage");
 
-  if coverage {
-    flags.coverage = true;
+  if matches.is_present("coverage") {
```
caspervonb, line 610: Default to a tempdir if no value is provided?

caspervonb:
So, minor side note.

The environment variable `DENO_COVERAGE_DIR` is not documented at the moment, this is on purpose as its considered internal for the time being and specifying one via `deno test --coverage <dir>` should still be the preferred way to initiate coverage for a test suite.

cli/main.rs
```diff
@@ -854,16 +854,25 @@ async fn test_command(
...
 
-  let mut maybe_coverage_collector = if flags.coverage {
-    let session = worker.create_inspector_session();
-    let mut coverage_collector =
-      tools::coverage::CoverageCollector::new(session);
-    coverage_collector.start_collecting().await?;
+  // If the actual coverage flag is set, we set the environment variable so that subprocesses will
+  // forward it.
+  if let Some(ref coverage_dir) = flags.coverage_dir {
+    env::set_var("DENO_COVERAGE_DIR", coverage_dir);
```
caspervonb, line 860: Rename to something obnoxious, like `DENO_UNSTABLE_INTERNAL_COVERAGE_DIR`?

cli/program_state.rs
```diff
@@ -103,8 +104,13 @@ impl ProgramState {
       None => None,
     };
 
+    let coverage_dir = flags
+      .coverage_dir
+      .clone()
+      .or_else(|| env::var("DENO_COVERAGE_DIR").ok());
```
bartlomieju, line 110: Why do we need both flag and env variable?

bartlomieju:
Is there a way to consume coverage that are captured in a directory?

caspervonb:
> Is there a way to consume coverage that are captured in a directory?

#8664 breaks it reporting into its own subcommand

cli/program_state.rs
caspervonb: We can make due with just an environment variable (the environment variable is needed for persisting between processes) but the general consensus in discord was that people preferred to keep `deno test --coverage <dir>` as the entry point for collecting coverage with deno test.

cli/tools/coverage.rs
```diff
@@ -212,10 +231,38 @@ impl PrettyCoverageReporter {
...
+  coverages.sort_by_key(|k| k.script_coverage.url.clone());
+
+  // TODO(caspervonb) drain_filter would make this cleaner, its nightly at the moment.
+  for i in (1..coverages.len() - 1).rev() {
```
caspervonb, line 248: Wasn't sure if this was prone to overflowing or not when I wrote this, but it is.

bartlomieju:
@caspervonb why is this PR breaking?

cli/flags.rs
bartlomieju: Yes, please

cli/main.rs
bartlomieju: Yes

bartlomieju: Also is this env var unset after coverage is done?

cli/web_worker.rs
```diff
@@ -158,7 +158,9 @@ impl WebWorker {
           &mut js_runtime,
           Some(inspector_server.clone()),
         ))
-      } else if program_state.flags.coverage || program_state.flags.repl {
+      } else if program_state.flags.coverage_dir.is_some()
+        || program_state.flags.repl
+      {
```
bartlomieju, line 163: This will conflict with #8659

caspervonb: > @caspervonb why is this PR breaking?

Only changes unstable, but because CLI usage changes from

```
$ deno test --unstable --coverage 
```

to 
```
$ deno test --unstable --coverage path_to_directory
```

caspervonb:
> @caspervonb why is this PR breaking?

Only changes unstable, but because CLI usage changes from

```
$ deno test --unstable --coverage 
```

to 
```
$ deno test --unstable --coverage path_to_directory
```

cli/web_worker.rs
caspervonb: > Also is this env var unset after coverage is done?

We're setting in in the current process so on exit it should be back to normal. Env inheritance only goes downwards not upwards.

> This will conflict with #8659

Hm yeah but trivially so?

cli/flags.rs
caspervonb: @lucacasonato objections? IIRC you were against a default.

bartlomieju:
> > @caspervonb why is this PR breaking?
> 
> Only changes to the unstable bits, but because CLI usage changes from
> 
> ```
> $ deno test --unstable --coverage 
> ```
> 
> to
> 
> ```
> $ deno test --unstable --coverage <path_to_directory>
> ```

If there's a default value provided it should work the same for the users, right?

cli/flags.rs
lucacasonato: Yes, I dont think we should add a default value. How do you get the temp dir? You would need to pass it to `deno cover` to generate a report right?

caspervonb:
> If there's a default value provided it should work the same for the users, right?

Yeah but since the plan is to move the reporting part of this into a subcommand I'm not sure how much sense a default makes because you'll need to know where you stored the raw data.

cli/web_worker.rs
caspervonb: Merged, no problem 👍

cli/flags.rs
caspervonb: Yeah, it'd just move the "BREAKING" bit to the next PR really so marking as resolved.

caspervonb:
More clean up to be done over time in terms of abstractions as this takes shape but after talking to @bartlomieju a bit more last night I've gone back to initial'ish design of making the collection directory implicit and tucking it away in $DENO_DIR/cov (at the moment, we probably want to key that based on the working directory but for simplicity I ended up leaving it at that).

lucacasonato:
> I've gone back to initial'ish design of making the collection directory implicit and tucking it away in $DENO_DIR/cov

I completely disagree with this. It will be highly confusing for users that their coverage reports include files from a previous run, potentially non existent files, or files from a concurrently running unrelated Deno process. We can't key by cwd, as subprocesses might not have the same cwd and are thus not recorded (whole point of this pr was subprocess coverage right?).

caspervonb:
> I completely disagree with this. 

Reverting is easy should it come to that 😁

> It will be highly confusing for users that their coverage reports include files from a previous run, potentially non existent files, or files from a concurrently running unrelated Deno process.

The entry point for collecting coverage is currently `deno test --coverage` and since we can easily differentiate between coverage initiated by the user via the flag and environment variable we can handle that and do by cleaning the directory when it's the entry point so it'll be a clean state.

Personally I like kcov's interface of an explicit flag to clean (`kcov --clean`) 

> We can't key by cwd, as subprocesses might not have the same cwd and are thus not recorded (whole point of this pr was subprocess coverage right?).

Keying is potentially a bit more tricky for sure.

To be clear everything is always recorded and it would be the working directory of the initial invocation (e.g the cwd at the time of calling `deno test --coverage`).

The basic idea being that, it is the cached result of running collecting coverage the current working directory. If this was cargo it would just be a `target/cov` folder.

docs/testing.md
```diff
@@ -220,7 +220,7 @@ stdout. In the future there will be support for `lcov` output too.
 
 ```
 $ git clone git@github.com:denosaurs/deno_brotli.git && cd deno_brotli
-$ deno test --coverage --unstable
+$ deno test --coverage coverage_dir --unstable
```
bartlomieju, line 223: Definition in `flags.rs` says `--coverage` requires `=` to pass value

bartlomieju (commented):
Looks mostly good to me, since coverage collection is an unstable feature (and CLI flag) in my opinion we should go with "move fast and break stuff" approach on this one.

docs/testing.md
caspervonb: Oops indeed that is out of date.

cli/tests/integration_tests.rs
```diff
@@ -3305,6 +3305,12 @@ itest!(deno_test_coverage {
   exit_code: 0,
 });
 
+itest!(deno_test_run_test_coverage {
```
lucacasonato, line 3308: Can you also add an itest where the dir is set explicitly?

lucacasonato (approved):
LGTM - thanks @caspervonb. Great feature :-)

Issue #8556, opened 2020-11-30
Workers have race condition


bartlomieju:
After calling `self.close()` at top level there is an error shown which should have been handled internally.

```
OUTPUT
hello
goodbye
error: Uncaught (in worker "") Error: execution terminated

OUTPUT
EXPECTED
hello
goodbye

EXPECTED
Error: thread 'local_sources_not_cached_in_memory' panicked at 'pattern match failed', test_util\src\lib.rs:903:7
```

https://github.com/denoland/deno/pull/8555/checks?check_run_id=1474798709

Likely related to https://github.com/denoland/deno/pull/8510

bartlomieju:
https://github.com/denoland/deno/pull/8631/checks?check_run_id=1506480280

bartlomieju:
Not fixed https://github.com/denoland/deno/runs/1510923329

bartlomieju:
Might be resolved by #8809

bartlomieju:
I'm not sure if this is still the case, especially after all the recent fixes by @andreubotella. Andreu have you seen this error?

andreubotella:
> I'm not sure if this is still the case, especially after all the recent fixes by @andreubotella. Andreu have you seen this error?

I don't think I've seen "Error: execution terminated" while I've been working on workers. And since the logs for the CI run have expired, I don't have much else to go on.

bartlomieju:
Let's close it then, I'll reopen if it happens again.

PR #8648, opened 2020-12-07
refactor(cli): use Mutex instead of AtomicBool for worker termination

bartlomieju:
Replace using `AtomicBool` in favor of `Mutex` for the `terminated` field
in `WebWorkerHandle`.

CC @bnoordhuis


bartlomieju:
Closing because #8556 appeared after the first refactor which indicates that it's not the root problem.

PR #8649, opened 2020-12-07
fix(crates/rt) deduplicate code

benjamingr:
Getting setup and recovering - starting with [small things](https://github.com/denoland/deno/pull/8376#issuecomment-734350503) I promised.

This just deduplicates the code a little here, adds a note and a test to make sure there is no future regression :]

I'll hopefully get to the rust stuff on the weekend :D

### DIFF
--- a/cli/rt/01_web_util.js
+++ b/cli/rt/01_web_util.js
@@ -128,11 +128,11 @@
     wrappedHandler.handler = handler;
     return wrappedHandler;
   }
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM, nice cleanup

benjamingr:
Thanks, sorry for forgetting to run `./tools/format.js` after running the test :]

PR #8651, opened 2020-12-07
feat(lsp): basic support for textDocument/completion

lucacasonato:
This adds basic `textDocument/completion` handling.

Currently text completions with `TextEdit`s do not work. I have not managed to figure out why though. Until I find the cause I have disabled the `includeCompletionsWithInsertText` TS user preference which should mitigate this.

```ts
function str(s?: Deno.File) {
  s.
//  ^- complete here is not as nice as VSCode built in. (optional chaining)
}
```

In the future we should implement `completionItem/resolve` to support `documentation` and `details` field in the `CompletionItem`.

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -568,7 +568,7 @@ pub struct CompletionInfo {
 }
 
 impl CompletionInfo {
-  pub fn to_completion_response(
+  pub fn into_completion_response(
     self,
     line_index: &[u32],
   ) -> lsp_types::CompletionResponse {
@@ -647,7 +647,7 @@ impl CompletionEntry {
     if let Some(insert_text) = insert_text {
       if let Some(replacement_range) = replacement_range {
         item.text_edit = Some(lsp_types::CompletionTextEdit::Edit(
-          lsp_types::TextEdit::new(replacement_range, insert_text.clone()),
+          lsp_types::TextEdit::new(replacement_range, insert_text),
         ));
       } else {
         item.insert_text = Some(insert_text);
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -575,7 +575,7 @@ impl CompletionInfo {
     let items = self
       .entries
       .into_iter()
-      .map(|entry| entry.to_completion_item(line_index))
+      .map(|entry| entry.into_completion_item(line_index))
       .collect();
     lsp_types::CompletionResponse::Array(items)
   }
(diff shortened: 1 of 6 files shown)


cli/lsp/tsc.rs
```diff
@@ -519,6 +559,111 @@ impl ReferenceEntry {
...
+        item.insert_text_format = Some(lsp_types::InsertTextFormat::Snippet);
+      }
+      _ => {}
+    }
+
+    let mut insert_text = self.insert_text;
+    let replacement_range: Option<lsp_types::Range> =
+      self.replacement_span.map(|span| span.to_range(line_index));
+
+    // // Make sure we only replace a single line at most
```
kitsonk, line 633: Probably should clean this up, make a TODO, or something...

cli/tsc/99_main_compiler.js
```diff
@@ -471,6 +471,10 @@ delete Object.prototype.__proto__;
     core.jsonOpSync("op_respond", { id, data });
   }
 
+  const userPreferences = {
```
kitsonk, line 474:
```suggestion
  /** @type {ts.UserPreferences} */
  const userPreferences = {
```

I think it is also debatable if this were moved to Rust as well, as sort of want zero config in JavaScript if we can avoid it.

kitsonk: As we will likely need to change/update/manage these on the lsp side, especially when we need to do refactorings... We could do it like the configuration managements as well...

```ts
export interface UserPreferences {
        readonly disableSuggestions?: boolean;
        readonly quotePreference?: "auto" | "double" | "single";
        readonly includeCompletionsForModuleExports?: boolean;
        readonly includeAutomaticOptionalChainCompletions?: boolean;
        readonly includeCompletionsWithInsertText?: boolean;
        readonly importModuleSpecifierPreference?: "auto" | "relative" | "non-relative";
        /** Determines whether we import `foo/index.ts` as "foo", "foo/index", or "foo/index.js" */
        readonly importModuleSpecifierEnding?: "auto" | "minimal" | "index" | "js";
        readonly allowTextChangesInNewFiles?: boolean;
        readonly providePrefixAndSuffixTextForRename?: boolean;
        readonly includePackageJsonAutoImports?: "auto" | "on" | "off";
        readonly provideRefactorNotApplicableReason?: boolean;
    }
```

kitsonk (commented):
a couple thoughts...

cli/tsc/99_main_compiler.js
lucacasonato: Good point. Let me move it.

kitsonk (approved):
LGTM 🥳

PR #8653, opened 2020-12-07
test(core): type aliases in OpState

bartlomieju:
This commit adds a test case to core/gotham_state.rs that shows
that type aliases can't be used reliably. Instead wrapper types
should be used.


crypt096 (approved):

lucacasonato (approved):
At first thought it is kind of surprising to me that this is the behaviour, but this at least sets it in stone.

LGTM

Issue #7644, opened 2020-09-23
Rewrite "prepareStackTrace" to use Rust API


bartlomieju:
Currently CLI uses [`Error.prepareStackTrace`](https://v8.dev/docs/stack-trace-api) API to perform source mapping and formatting of errors. 

This situation leads to overall complexity of this method as well as requirement to manually acquire information about stack frames as well as distributing the logic between JS and Rust.

We want to avoid jumping between JS and Rust as much as possible and to solve this problem in this situation we need to:
1) Add new APIs to V8's [`StackFrame`](https://source.chromium.org/chromium/chromium/src/+/master:v8/include/v8.h;l=2184?q=StackFrame%20v8.h&sq=&ss=chromium) to acquire the same data as using `CallSite`:

https://github.com/denoland/deno/blob/d68fb813428f31b21b8a46cdde21d9ab5170abdc/cli/rt/40_error_stack.js#L172-L191

2) Remove `prepareStackTrace` method from `cli/rt/40_error_stack.js` and instead use [`Isolate::set_prepare_stack_trace_callback`](https://source.chromium.org/chromium/chromium/src/+/master:v8/src/api/api.cc;l=8449?q=setpreparestacktracecallback&sq=&ss=chromium) - needs to be added to `rusty_v8` first

CC @ry @nayeemrmn

bartlomieju:
With recent improvements to error handling by @nayeemrmn done in https://github.com/denoland/deno/pull/14378, https://github.com/denoland/deno/pull/14394, https://github.com/denoland/deno/pull/14302 and https://github.com/denoland/deno/pull/14274 I feel this issue is no longer of high importance. Closing for now.

PR #5254, opened 2020-05-12
fix(cli/js/error_stack): Expose Error.captureStackTrace

nayeemrmn:
Fixes #5127.
Closes #5251.
Ref https://v8.dev/docs/stack-trace-api#stack-trace-collection-for-custom-exceptions.

PR #8660, opened 2020-12-07
refactor(core): stack trace mapping

bartlomieju:
This commit adds "Deno.core.createPrepareStackTrace". This function
was moved from "cli/rt/40_error_stack.js" to unify handling of stack frames in core
(before this PR there was implicit dependency on logic in "core/error.rs::JsError").
Unfortunately formatting logic must still be duplicated in "cli/error.js::PrettyJsError"
to provide coloring, but currently there's no solution to this problem.
"createPrepareStackTrace" can accept a single argument; a function that takes
a location and provides source mapped location back. 

Towards library crate.

@nayeemrmn this PR move stack frames evaluation as we discussed on Discord. I'm still not sure what should we do with the formatting utility I guess it could be removed completely and up to the consumer to do that.

Ref #7644


core/error.js
```diff
@@ -0,0 +1,231 @@
...
+          }
+          return callSite;
+        },
+      );
+      Object.defineProperties(error, {
+        __callSiteEvals: { value: [], configurable: true },
+      });
+      const formattedCallSites = [];
+      for (const callSite of mappedCallSites) {
+        error.__callSiteEvals.push(Object.freeze(evaluateCallSite(callSite)));
```
bnoordhuis, line 209: I'm not sure if it's still the case but `Object.freeze()` used to carry a pretty big performance penalty, both to execute and because of the knock-on effects of changing the object's hidden class.

A quick look at V8's source suggests it's still a complicated affair (`JSObject::PreventExtensionsWithTransition()` in `src/objects/js-objects.cc`) so I'd try to avoid it if possible.

bnoordhuis (commented):
An observation about `Object.freeze()`, otherwise LGTM.

core/error.js
bartlomieju: Yeah I'm not sure if it makes sense to freeze it as those frames are immediately captured by `deno_core::JsError`

nayeemrmn: Not necessarily immediately, this function is called when `error.stack` is first accessed by the user. `JsError` is created on termination.

Nevertheless, as of #5254 the `__callSiteEvals` field is `configurable: true`, so there's definitely no point in freezing the value. It's a holdover.

bartlomieju: Good point, I'll remove the freezing then.

bartlomieju:
So I've tried removing formatting but that will produce non-usable stack frames. I think the best bet is to keep the formatting as is and keep synchronizing it manually.

core/error.js
```diff
@@ -0,0 +1,230 @@
...
+      isEval: callSite.isEval(),
+      isNative: callSite.isNative(),
+      isConstructor: callSite.isConstructor(),
+      isAsync: callSite.isAsync(),
+      isPromiseAll: callSite.isPromiseAll(),
+      promiseIndex: callSite.getPromiseIndex(),
+    };
+  }
+
+  function createPrepareStackTrace(sourceMappingFn) {
```
ry, line 178: Can you add some documentation here about the type of the `sourceMappingFn`?

ry (approved):
LGTM - I think this is an unfortunate complication of deno_core but I don't see another way.

core/error.js
bartlomieju: Sure thing

PR #8664, opened 2020-12-08
feat: add "deno coverage" subcommand

### LINKED ISSUE #7648: A cover command and cross process code coverage

### LINKED ISSUE #8726: deno coverage
the coverage feature is super awesome - however, if I have a project directory structure like:
```
other/
src/
tests/
```

Where tests references code exported by src/ and src references code exported by other, I only see coverage for non-test files in test/ (which are test support files).

caspervonb:
This adds a new subcommand called _coverage_ which can generate code coverage reports to stdout in multiple formats from code coverage profiles collected to disk.

Currently this supports outputting a pretty printed diff and the lcov format for interoperability with third-party services and tools.
 
Code coverage is still collected via other subcommands that run and collect code coverage such as `deno test --coverage=<directory>` but that command no longer prints a pretty printed report at the end of a test run with coverage collection enabled.

The restrictions on which files that can be reported on has also been relaxed and are fully controllable with the include and exclude regular expression flags on the coverage subcommand.

Closes #7648
Closes #8726


caspervonb:
Blocking on https://github.com/denoland/deno/pull/8642

Early draft, filtering yet to be done.
Just raising this early for visibility.

cli/tools/coverage.rs
```diff
@@ -212,41 +230,56 @@ impl PrettyCoverageReporter {
...
+
+  coverages = coverages
     .into_iter()
-    .filter(|e| {
-      if let Ok(url) = Url::parse(&e.script_coverage.url) {
-        if url.path().ends_with("__anonymous__") {
-          return false;
-        }
+    .filter(|e| !e.script_coverage.url.ends_with("__anonymous__"))
+    .collect::<Vec<Coverage>>();
```
caspervonb, line 250: **TODO** filter out anything that starts with any patterns in ignore.
Ignore should default to include files starting with the basename `test`.

caspervonb: Most other coverage reporting tools (nyc, kvoc, grcov, etc) use include and exclude, where an include match would override a match in exclude.

Went with ignore because its what we use in the other commands like doc and lint but we might want to reconsider

caspervonb:
Closing for the time being and adding the planned patches directly onto test until it's more obvious why we're going to need this.

cli/tools/coverage.rs
caspervonb: Reverted to a simpler filter for the time being.

cli/tools/coverage.rs
```diff
@@ -296,30 +289,46 @@ fn collect_coverages(dir: &PathBuf) -> Result<Vec<ScriptCoverage>, AnyError> {
...
 ) -> Vec<ScriptCoverage> {
+  let include_patterns = include
+    .iter()
+    .map(|i| glob::Pattern::new(i).unwrap())
+    .collect::<Vec<glob::Pattern>>();
+  let exclude_patterns = exclude
+    .iter()
+    .map(|x| glob::Pattern::new(x).unwrap())
+    .collect::<Vec<glob::Pattern>>();
+
```
caspervonb, line 303: Because these come from file urls and therefore are absolute; probably conflicts with user expectations.

cli/tools/coverage.rs
```diff
@@ -296,30 +289,46 @@ fn collect_coverages(dir: &PathBuf) -> Result<Vec<ScriptCoverage>, AnyError> {
...
+          if let Some(file_stem) = file_path.file_stem() {
+            if file_stem == "__anonymous__" {
+              return false;
+            }
+
+            if file_stem.to_str().unwrap().starts_with("$deno") {
+              return false;
+            }
+
+            if file_stem.to_str().unwrap().ends_with("test") {
```
caspervonb, line 326: This should just be in the default pattern.

caspervonb:
Got a ton of cleanup to do over the next couple of days but this is functional.

cli/flags.rs
```diff
@@ -35,6 +34,11 @@ pub enum DenoSubcommand {
   Completions {
     buf: Box<[u8]>,
   },
+  Cover {
+    files: Vec<PathBuf>,
+    ignore: Vec<PathBuf>,
```
caspervonb, line 39: Maybe --ignore to ignore files (which are coverage data files) and --exclude to exclude source files?

wperron:
I double checked how Go handles and it's also separated into two commands (it's actually an add-on tool that must be installed, but we don't have an equivalent to Go's "tools" concept) - I'm in favor

ry:
I’m concerned this is a bit out of scope for Deno. The sub-command is just for converting json coverage to lcov ? Why not an external utility for this ? (Maybe in std?)

wperron:
from conversation on Discord:
> @caspervonb here's the docs page for the tools in Go https://pkg.go.dev/golang.org/x/tools it's basically just a directory of packages that can be installed via go get or by installing the binary directly - the only difference with a random tool hosted on github is that the Go cli aliases them through the go tool subcommand, and even that just looks for binary in a special path (I believe $GOPATH/tools/bin )

I think we're getting to the point where the toolchain that ships with the cli includes a lot of different commands, some more "core" than others. This one certainly gets further from the core than the test command for example. I wouldn't mind having this in the cli, but I can also see this being part of an "extended toolchain" set of scripts, similar to Golang's tools.

I opened a discussion to talk a bit more about the toolchain here #9412 

link to the thread: https://discord.com/channels/684898665143206084/684911491035430919/807241638543753267

caspervonb:
> I’m concerned this is a bit out of scope for Deno. The sub-command is just for converting json coverage to lcov ? Why not an external utility for this ? (Maybe in std?)

It's a bit convoluted as we need to re-use some bits of `program_state` (in this implementation at-least) to get the compiled source, cached source and source map in the same way as deno run would.

If the source maps etc were just in the user's tree there wouldn't be an issue but they're an implementation detail of the internal state so to do the same in userland script there would be quite a bit of re-implementation and potential brittleness 🤔

cli/flags.rs
```diff
@@ -1082,6 +1102,60 @@ Future runs of this module will trigger no downloads or compilation unless
     )
 }
 
+fn cover_subcommand<'a, 'b>() -> App<'a, 'b> {
+  SubCommand::with_name("cover")
```
bartlomieju, line 1106: As discussed offline, should we rename to `coverage`?

caspervonb: Rename the command and test prefixes to coverage.

docs/testing.md
```diff
@@ -209,29 +209,28 @@ deno test --fail-fast
...
 
-This coverage information is acquired directly from the JavaScript engine (V8).
-Because of this, the coverage reports are very accurate.
+This coverage information is acquired directly from the JavaScript engine (V8)
+which is very accurate.
 
-When all tests are done running a summary of coverage per file is printed to
-stdout. In the future there will be support for `lcov` output too.
+This can then be further processed from the internal format into well known
+formats by the `deno cover` tool.
```
lucacasonato, line 219:
```suggestion
formats by the `deno coverage` tool.
```

docs/testing.md
```diff
@@ -209,29 +209,28 @@ deno test --fail-fast
...
+
+# From this you can get a pretty printed diff of uncovered lines
+deno cover --unstable coverage
+
+# Or generate an lcov report
+deno cover --unstable coverage --lcov > coverage.lcov
+
+# Which can then be further processed by tools like genhtml
+genhtml -o coverage_html coverage.lcov
 ```
```
lucacasonato, line 236: We should explain the `--include` and `--exclude` flags here. `std` is kinda special because it does not have external imports, and thus coverage does not need filtering. I would advocate for using some other module (not std) to better illustrate the workflow for real world situations.

lucacasonato (commented):
I have tested this PR locally. My findings so far:

- `deno coverage --unstable` with no files passed results in cryptic error.
- When specifying the `--include` option without the `--exclude` option, I would expect only the files in `--include` to be used (allowlist). Instead it seems `--include` has no effect without `--exclude`.
- There is no documentation in `--help` about how the matching for `--include` and `--exclude` works.

cli/tests/coverage/test_cover_branch_pretty.out
```diff
@@ -0,0 +1 @@
+o
```
caspervonb, line 1: Remove this file, added by accident (correct output file is test_coverage_branch.out).

cli/flags.rs
```diff
@@ -1082,6 +1102,72 @@ Future runs of this module will trigger no downloads or compilation unless
     )
 }
 
+fn coverage_subcommand<'a, 'b>() -> App<'a, 'b> {
+  SubCommand::with_name("coverage")
+    .about("Print coverage reports")
+    .long_about(
+      "Print coverage reports from coverage profiles.
+
```
bartlomieju, line 1110: Let's make this subcommand unstable for now

cli/flags.rs
```diff
@@ -1082,6 +1102,72 @@ Future runs of this module will trigger no downloads or compilation unless
...
+Print a report to stdout:
+  deno coverage cov_profile
+
+Include urls that start with the file schema:
+  deno coverage --include=\"^file:\" cov_profile
+
```
bartlomieju, line 1116: It's be great to have a complete example here including subcommand invocation for `deno test` with coverage collection

cli/flags.rs
```diff
@@ -1082,6 +1102,72 @@ Future runs of this module will trigger no downloads or compilation unless
...
+        .help("Ignore coverage files"),
+    )
+    .arg(
+      Arg::with_name("include")
+        .long("include")
+        .takes_value(true)
+        .multiple(true)
+        .require_equals(true)
+        .default_value(r"^file:")
+        .help("Include source files in the report"),
```
bartlomieju, line 1146: Add note that this flag accepts regex?

cli/flags.rs
```diff
@@ -1082,6 +1102,72 @@ Future runs of this module will trigger no downloads or compilation unless
...
+    )
+    .arg(
+      Arg::with_name("exclude")
+        .long("exclude")
+        .takes_value(true)
+        .multiple(true)
+        .require_equals(true)
+        .default_value(r"test\.(js|ts)$")
+        .help("Exclude source files from the report"),
```
bartlomieju, line 1155: Ditto

cli/flags.rs
bartlomieju: I see that it's checked in `main.rs`, can we add `require("unstable")` here instead?

cli/tests/integration_tests.rs
```diff
@@ -3489,45 +3489,27 @@ console.log("finish");
...
+    output: "test_coverage_branch_lcov.out",
     exit_code: 0,
   });
 
-  itest!(deno_test_branch_coverage {
-    args: "test --coverage --unstable test_branch_coverage.ts",
-    output: "test_branch_coverage.out",
+  itest!(test_coverage_complex {
+    args: "run --allow-all test_coverage_complex.ts",
+    output: "test_coverage_complex.out",
```
bartlomieju, line 3506: Nitpick: please nest all coverage tests in `mod coverage` and put fixture files in `cli/tests/coverage`

cli/flags.rs
```diff
@@ -1099,6 +1119,77 @@ Future runs of this module will trigger no downloads or compilation unless
...
+        .help("Include source files in the report"),
+    )
+    .arg(
+      Arg::with_name("exclude")
+        .long("exclude")
+        .takes_value(true)
+        .value_name("regex")
+        .multiple(true)
+        .require_equals(true)
+        .default_value(r"test\.(js|ts)$")
```
kt3k, line 1176:
```suggestion
        .default_value(r"test\.(js|ts)x?$")
```

Maybe we should also exclude `test.tsx` by default as `deno test` automatically picks them up.

caspervonb: Did `r"test\.(js|mjs|ts|jsx|tsx)$"` instead which is the full list as listed in --help.

cli/tests/integration_tests.rs
caspervonb: It's a bit odd since the working directory is in tests and itest doesn't let you change it but moved them.

docs/testing.md
caspervonb: Documented them better and added a short "explainer" paragraph to testing.md.
Kind of indifferent to which module to use for the example, but assuming we'll host the html output for std soon on deno.land that would be good to link here.

lucacasonato: I would prefer something more complex with external dependencies, like oak.

cli/flags.rs
```diff
@@ -1099,6 +1119,77 @@ Future runs of this module will trigger no downloads or compilation unless
...
+        .value_name("regex")
+        .multiple(true)
+        .require_equals(true)
+        .default_value(r"test\.(js|mjs|ts|jsx|tsx)$")
+        .help("Exclude source files from the report"),
+    )
+    .arg(
+      Arg::with_name("lcov")
+        .long("lcov")
+        .help("Output cover result in lcov format")
```
lucacasonato, line 1182:
```suggestion
        .help("Output coverage report in lcov format")
```

cli/tests/coverage/test_branch_lcov.ts
```diff
@@ -0,0 +1,29 @@
+const dir = await Deno.makeTempDir();
```
lucacasonato, line 1: Couldn't this whole test be moved to Rust?

cli/tests/coverage/test_branch_lcov.ts
```diff
@@ -0,0 +1,29 @@
...
+    "test",
+    "--unstable",
+    "--coverage=" + dir,
+    "coverage/branch_test.ts",
+  ],
+  stdout: "inherit",
+  stderr: "inherit",
+});
+
+await test.status();
```
lucacasonato, line 15: You are not checking the exit code here.

cli/tests/coverage/test_complex.ts
```diff
@@ -0,0 +1,28 @@
+const dir = await Deno.makeTempDir();
```
lucacasonato, line 1: Ditto

cli/tests/coverage/test_complex_lcov.ts
```diff
@@ -0,0 +1,29 @@
+const dir = await Deno.makeTempDir();
```
lucacasonato, line 1: Ditto

lucacasonato (commented):
Looking very good. I have a few more comments.

I see that we are doing a lot of line offset calculations. We do that in the LSP too and we keep running into issues with multi byte Unicode chars. Can you add a test for both reporters with files test files that contain chinese MBC and emojis?

cli/tests/integration_tests.rs
```diff
@@ -3977,6 +3935,146 @@ console.log("finish");
...
+        .arg(format!("--coverage={}", tempdir.path().to_str().unwrap()))
+        .arg("cli/tests/coverage/complex_test.ts")
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::inherit())
+        .status()
+        .expect("failed to spawn test runner");
+
+      assert!(status.success());
+
+      let output = util::deno_cmd()
```
lucacasonato, line 4025: Not checking the status for this output here 😉 

Same thing in the other tests.

cli/tools/coverage.rs
```diff
@@ -117,17 +113,250 @@ pub struct GetScriptSourceResult {
...
+        };
+
+        let taken = range.count;
+
+        println!(
+          "BRDA:{},{},{},{}",
+          line_index + 1,
+          block_number,
+          branch_number,
+          taken
```
lucacasonato, line 241: According to https://manpages.debian.org/unstable/lcov/geninfo.1.en.html, taken should be `-` if the branches' surrounding block was not executed. Is this relevant for us?

caspervonb: Huh, so seems v8 just doesn't emit branch info if the outer block was never hit.

caspervonb: Seems we don't ever hit this state, left a comment and might have to come back here and do parse and traverse the source as the baseline for what "can be hit".

lucacasonato (approved):
LGTM! Thanks @caspervonb

@bartlomieju Could you also take a look?

cli/main.rs
```diff
@@ -978,6 +978,34 @@ async fn run_command(flags: Flags, script: String) -> Result<(), AnyError> {
...
+  lcov: bool,
+) -> Result<(), AnyError> {
+  if !flags.unstable {
+    exit_unstable("compile");
+  }
+
+  if files.is_empty() {
+    println!("No matching coverage profiles found");
+    std::process::exit(1);
+  }
```
bartlomieju, line 996: Nitpick: shouldn't `files` be a required argument in CLI definition?

cli/tests/coverage/complex.ts
```diff
@@ -10,7 +10,8 @@ export interface Complex {
   baz: string;
 }
 
-export function complex(
+// 패딩에 대한 더 많은 문자.
```
bartlomieju, line 13: I guess these have to do something with figuring out ranges, care to add a comment explaining that?

cli/tests/integration_tests.rs
```diff
@@ -3977,6 +3935,154 @@ console.log("finish");
     assert_eq!(output.stderr, b"");
   }
 
+  mod coverage {
+    use super::*;
+
+    #[test]
+    fn branch() {
```
bartlomieju, line 3942: Nice, thank you!

bartlomieju (approved):
LGTM, just a few nitpicks, but nothing critical. Awesome work Casper!

cli/tests/coverage/complex.ts
caspervonb: Went overly commenty on it, which is fine since the comments are for line padding, ensuring comments don't show up as uncovered and they add some wide characters.

cli/main.rs
caspervonb: Made it required in flags, but also keeping this sanity check here.

PR #8668, opened 2020-12-08
Remove dead code: itest_ignore

ry:
merge on approval


bartlomieju:
@ry are you sure you want to land it? It's gonna be a bit problematic if we'll have to ignore a test once more

crypt096 (approved):

ry:
It’ll be easy to find if we have to bring it back. I think we should have a policy of removing all unused code.

bartlomieju (approved):

Issue #8673, opened 2020-12-08
deno 1.6 compile getting error: Is a directory (os error 21)


elgs:
```
/tmp$ deno --version
deno 1.6.0 (release, x86_64-apple-darwin)
v8 8.8.278.2
typescript 4.1.2
/tmp$ 
/tmp$ cat a.js
console.log('hello deno');
/tmp$ 
/tmp$ deno compile --unstable a.js
Bundle file:///private/tmp/a.js
Compile file:///private/tmp/a.js
error: Is a directory (os error 21)
/tmp$ 
```

lucacasonato:
Do you have a folder at `/private/tmp/a`?

We should throw an error if the inferred output path already exists.

elgs:
```
/$ ls -l /tmp
lrwxr-xr-x@ 1 root  wheel  11 Jan  1  2020 /tmp -> private/tmp
```

Yes, looks like /tmp and /private/tmp are the same on my computer. I moved a.js to ~, and it worked.

elgs:
@lucacasonato I just read your question again. I indeed have a folder at /private/tmp/a.

PR #8681, opened 2020-12-09
fix(cli): Throw error when the inferred output path already exists during compile

yos1p:
Fixes: #8673 

When the inferred path already exists, throw generic error to provide better error message.


yos1p:
Please let me know better error message if you have some suggestions.

elgs:
I appreciate that I opened an issue in the morning, in the evening, boom, a pr is ready to merge! Thank you so much.

cli/standalone.rs
```diff
@@ -150,6 +150,10 @@ pub async fn create_standalone_binary(
     } else {
       output
     };
+  if output.exists() {
+    let error_msg = format!("File / Directory {:?} already exists. Please specify another output name.", &output);
```
elgs, line 154: If a file with the inferred name exists, should it throw this error or just overwrite it? If this error is thrown, we would have to delete the compiled file every time before each compile.

cli/standalone.rs
```diff
@@ -4,8 +4,8 @@ use crate::permissions::Permissions;
 use crate::program_state::ProgramState;
 use crate::tokio_util;
 use crate::worker::MainWorker;
-use deno_core::error::type_error;
 use deno_core::error::AnyError;
+use deno_core::error::{generic_error, type_error};
```
kitsonk, line 8:
```suggestion
use deno_core::error::generic_error;
use deno_core::error::type_error;
```

cli/standalone.rs
kitsonk: I think we need to consider the common pattern of making a change and updating the bundle...  No other tool complains when you are updating the output, it overwrites by default.  The only problem is when the output is a directory, or trying to overwrite an input file...

kitsonk (commented):
We also need a test of some sort.

cli/standalone.rs
yos1p: 👍

cli/standalone.rs
yos1p: Maybe if we put some sort of signature in the output file, we might be able to detect whether it's an output file that can be overwritten or an input file, which means need to throw error. Just random idea.

yos1p: Or we can use flag `--overwrite` that allows the CLI to overwrite the file if the file exists.

elgs: > Or we can use flag --overwrite that allows the CLI to overwrite the file if the file exists.

I think that would be too much.

> Maybe if we put some sort of signature in the output file, we might be able to detect whether it's an output file that can be overwritten or an input file, which means need to throw error. Just random idea.

Or could be some magic number like 0xcafebabe in Java.

lucacasonato: There is already a magic signature in the last 16 bytes. We could use that.

yos1p:
> We also need a test of some sort.

I've added some tests in the commits. Let me know what you think.

cli/standalone.rs
yos1p: The compile command now check for the magic trailer and will only overwrite when it exists. I think this design is good.

But commonly, compiler would overwrite whatever file is the output (e.g, `tsc`, `rustc`, `javac`). So I kinda doubt whether we should check for this and prefer to simply overwrite the output file if it's exists (which is the current implementation).

elgs: tsc, rustc, javac or gcc's output files by default are with an extension, like .js, .so, .class, and .out. The default filename of deno compile however is the source file without extension, so it's more likely to collide with other files or directories. So I think it makes sense to do this check. If it's not too complicated.

cli/standalone.rs
```diff
@@ -150,6 +151,28 @@ pub async fn create_standalone_binary(
     } else {
       output
     };
+
+  if output.exists() {
+    // If the output is a directory, throw error
+    if output.is_dir() {
+      let error_msg =
+        format!("Could not compile: {:?} is a directory.", &output);
+      return Err(generic_error(error_msg));
```
bartlomieju, line 160: Nitpick: this could be replaced by:
```rust
use deno_core::error::bail;
...
bail!("Could not compile: {:?} is a directory.", &output);
```

Using `generic_error` here is an overkill because this function never interacts with JS runtime (it's a CLI utility)

cli/standalone.rs
```diff
@@ -150,6 +151,28 @@ pub async fn create_standalone_binary(
...
+    let mut output_file = File::open(&output)?;
+    output_file.seek(SeekFrom::End(-16))?;
+    let mut trailer = [0; 16];
+    output_file.read_exact(&mut trailer)?;
+    let (magic_trailer, _) = trailer.split_at(8);
+    if magic_trailer != MAGIC_TRAILER {
+      let error_msg =
+        format!("Could not compile: cannot overwrite {:?}.", &output);
+      return Err(generic_error(error_msg));
+    }
```
bartlomieju, line 174: ditto

cli/tests/integration_tests.rs
```diff
@@ -4672,3 +4672,101 @@ fn standalone_no_module_load() {
   assert!(util::strip_ansi_codes(&stderr_str)
     .contains("Self-contained binaries don't support module loading"));
 }
+
+#[test]
```
bartlomieju, line 4676: Great tests!

cli/standalone.rs
yos1p: I see. Done, thanks!

lucacasonato (approved):
LGTM. Thanks @yos1p

PR #8683, opened 2020-12-09
feat(lsp): support import maps

kitsonk:
Supports specifying an import map in the configuration of a language server client and having that applied to the workspace.


ry (commented):
Is there a way to add a small test to demo this feature?

kitsonk:
> Is there a way to add a small test to demo this feature?

Doing an integration test would be difficult/complicated.  I have added a unit test though which demonstrates when the option is set during startup, that the import map file gets read and is available in the lsp's state object.

ry (approved):
LGTM

PR #8688, opened 2020-12-09
chore: update swc_ecmascript to 0.15.0

lucacasonato:
Tokio 0.3 is now part of the dependencies, but only as a dev dependency in deno_doc. The duplication this causes will be resolved once we migrate to tokio 0.3.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -34,9 +34,9 @@ winapi = "0.3.9"
 [dependencies]
 deno_core = { path = "../core", version = "0.70.0" }
 deno_crypto = { path = "../op_crates/crypto", version = "0.4.0" }
-deno_doc = "0.1.17"
+deno_doc = "0.1.18"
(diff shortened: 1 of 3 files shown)


bartlomieju (changes requested):
@lucacasonato I suggest only upgrading swc related crates - I tried upgrading other crates recently and it results in large number of crates that will have multiple version in binary. We can update tokio after landing fix to resource table.

cli/Cargo.toml
```diff
@@ -70,8 +70,8 @@ serde = { version = "1.0.116", features = ["derive"] }
 shell-escape = "0.1.5"
 sourcemap = "6.0.1"
 swc_bundler = "0.17.5"
```
bartlomieju, line 72: Bump to `0.17.6` along the way?

bartlomieju (approved):
LGTM

PR #8700, opened 2020-12-10
[docs] fix: naming in `std/io` usage example

tomholford:
Reflect recent name changes for these functions:

- `writerFromStreamWriter` (see [here][0])
- `readerFromStreamReader` (see [here][1])

[0]: https://github.com/denoland/deno/blob/723fbb8f68eeaf6e5c14d9f8ac2c32f986ed60d9/std/io/streams.ts#L4

[1]: https://github.com/denoland/deno/blob/723fbb8f68eeaf6e5c14d9f8ac2c32f986ed60d9/std/io/streams.ts#L17

### DIFF
--- a/std/io/README.md
+++ b/std/io/README.md
@@ -126,26 +126,26 @@ base0123456789
 
 ## Streams
 
-### fromStreamReader
+### readerFromStreamReader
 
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
LGTM

PR #8701, opened 2020-12-10
refactor(lsp): make loop async + tsc on aux thread

lucacasonato:
This is still WIP.

This PR makes the main loop in the LSP async. This is to support the second thing this PR does: move tsc into a seperate thread.

This PR is just to try out if this is possible. I am going to extract the async LSP loop into a seperate PR, and will clean it up. I think a lot of the LSP loop can be simplified by using a Futures based model, instead of rebuilding all of this like the LSP does now.

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -13,7 +13,6 @@ use deno_core::error::AnyError;
 use deno_core::serde_json;
 use deno_core::serde_json::Value;
 use deno_core::url::Url;
-use deno_core::JsRuntime;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::mem;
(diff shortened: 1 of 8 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -223,43 +222,41 @@ fn ts_json_to_diagnostics(
   )
 }
 
-pub fn generate_ts_diagnostics(
-  state: &ServerStateSnapshot,
-  runtime: &mut JsRuntime,
+pub async fn generate_ts_diagnostics(
+  snapshot: &ServerStateSnapshot,
```
kitsonk, line 226: I would still prefer `state` here (or if you must `state_snapshot`)...  RA uses the type interchangeably, and at first I was thinking you were replacing the `JsRuntime` with a v8 snapshot and starting up the isolate each call, but then actually paid attention...

cli/lsp/tsc.rs
```diff
@@ -1144,6 +1149,50 @@ pub fn request(
...
+type Request = (
+  RequestMethod,
+  ServerStateSnapshot,
+  oneshot::Sender<Result<Value, AnyError>>,
+);
+
+#[derive(Clone)]
+pub struct TSC(mpsc::Sender<Request>);
+
+impl TSC {
```
kitsonk, line 1161: Probably needs to be `Tsc`, like `Url`, as `TSC` made me think this was some sort of generic or constant, which it is neither.

kitsonk (commented):
I only skimmed it until it isn't a WIP anymore, but awesome!

lucacasonato:
Closing in favor of #8727

PR #8705, opened 2020-12-10
fix(doc): Resolves dead link on script installer man page

### LINKED ISSUE #8704: Dead link on `Script Installer` manual page
On the v1.6.0 manual page (and also previous pages) there is a link with text `import.meta.main` going to https://deno.land/manual@v1.6.0/examples/testing_if_main which appears to be dead link.

mitch292:
Resolves #8704 

Changed link to reference - https://deno.land/manual@v1.6.0/examples/module_metadata - instead. What we look to be really concerned with is just the explainer, quoted below, on import.meta.main. There is also an example on this page. 

> The boolean import.meta.main will let you know if the current module is the program entry point.

Might be better to link this page instead - https://doc.deno.land/builtin/stable#ImportMeta. Let me know and I can update the PR.


bartlomieju (approved):
LGTM, thanks @mitch292

PR #8708, opened 2020-12-10
test(std/wasi): add a case for open directory/../..

caspervonb:
This adds a test case to ensure that we can't escape by going down a
directory before dotting our way out of it.


caspervonb:
Aside, I should probably ship the sources next to the binaries in wasi-test-suite for more readable diffs.

lucacasonato (approved):
LGTM - thanks

PR #8710, opened 2020-12-10
test(std/wasi): run test runner with --no-check

caspervonb:
This runs the test runner with --no-check which shaves the time taken down from a bit over 1 minute to 5 seconds.

We still import "snapshot_preview1" file in the test file that prepares the test runner invocations so we still get type checking there, just omitting it in the runner.

### DIFF
--- a/std/wasi/snapshot_preview1_test.ts
+++ b/std/wasi/snapshot_preview1_test.ts
@@ -90,6 +90,7 @@ for (const pathname of tests) {
             "--quiet",
             "--unstable",
             "--allow-all",
+            "--no-check",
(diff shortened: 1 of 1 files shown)


caspervonb:
This patch

```
$ time deno test --allow-all --unstable std/wasi

real	0m4.143s
user	0m6.040s
sys	0m0.821s
```

Master

```
$ time deno test --allow-all --unstable std/wasi

real	0m49.151s
user	1m26.559s
sys	0m2.713s
```

lucacasonato (approved):
LGTM - thanks

PR #8711, opened 2020-12-10
docs(std/wasi): add a basic description of Context

caspervonb:


lucacasonato (approved):
Thanks. Doc improvements are always great

PR #8712, opened 2020-12-10
fix(std/wasi): disallow multiple starts

caspervonb:
This disallows start and initialize from being called multiple times with the same, or different instances.

Once a context has started, it is bound to that context forever and calls to start/initialize needs to reflect this.

Also; had a look at Node's implementation and this mimicks their start/initialize semantics.


lucacasonato (approved):
LGTM - thanks

PR #8717, opened 2020-12-10
feat(cli/tools): bundle before install

### LINKED ISSUE #5069: deno install should bundle before installation
`deno install` is a utility that allows to install any module as an executable by creating shell script that invokes actual module.

Currently it works by caching all dependencies of the module in `$DENO_DIR` and adding entry point local URL to the shell script. 

Huge caveat in this approach is that if user reloads dependencies installed script might break (especially for untagged imports...). 
To mitigate this problem the script should be bundled into a single file during installation. 

This way installed script can be made into standalone utilities that do not share cache with other development projects and are completely standalone.

CC @ry

kt3k:
This PR makes `deno install` bundles the script before installing. It also adds `--no-bundle` option to the command, and the user can choose not to bundle the script at installation with that option.

I created `cli/tools/bundler.rs` file for sharing the bundling feature between `install_command` and `bundle_command`.

closes #5069

### DIFF
--- a/cli/tools/bundler.rs
+++ b/cli/tools/bundler.rs
@@ -1,0 +1,137 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
+
+use crate::colors;
+use crate::file_watcher;
+use crate::flags::Flags;
+use crate::fs_util;
+use crate::info;
+use crate::module_graph;
+use crate::program_state::ProgramState;
+
+use deno_core::error::AnyError;
+use deno_core::futures::future::FutureExt;
+use deno_core::ModuleSpecifier;
+use std::path::PathBuf;
+
+pub fn bundle_module_graph(
+  module_graph: module_graph::Graph,
+  flags: Flags,
+  debug: bool,
+) -> Result<String, AnyError> {
+  let (bundle, stats, maybe_ignored_options) =
+    module_graph.bundle(module_graph::BundleOptions {
+      debug,
+      maybe_config_path: flags.config_path,
+    })?;
+  match maybe_ignored_options {
+    Some(ignored_options) if flags.no_check => {
+      eprintln!("{}", ignored_options);
+    }
+    _ => {}
+  }
+  debug!("{}", stats);
+  Ok(bundle)
+}
+
+/// Returns a closure which returns the module graph of
+/// the given source file. The closure can be passed to file_watcher.
+pub fn get_module_resolver(
+  flags: &Flags,
+  source_file: &str,
+) -> impl Fn() -> file_watcher::FileWatcherFuture<
+  file_watcher::ModuleResolutionResult<module_graph::Graph>,
+> {
+  let flags = flags.clone();
+  let source_file = source_file.to_string();
+  let debug = flags.log_level == Some(log::Level::Debug);
+  move || {
+    let flags = flags.clone();
+    let source_file1 = source_file.clone();
+    let source_file2 = source_file.clone();
+    async move {
+      let module_specifier =
+        ModuleSpecifier::resolve_url_or_path(&source_file1)?;
+
+      debug!(">>>>> bundle START");
+      let program_state = ProgramState::new(flags.clone())?;
+
+      info!(
+        "{} {}",
+        colors::green("Bundle"),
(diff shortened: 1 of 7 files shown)


cli/tools/bundler.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Returns a closure which returns the module graph of
+/// the given source file. The closure can be passed to file_watcher.
+pub fn get_module_resolver(
+  flags: &Flags,
+  source_file: &str,
+) -> Box<
+  dyn Fn() -> file_watcher::FileWatcherFuture<
+    file_watcher::ModuleResolutionResult<module_graph::Graph>,
+  >,
+> {
```
magurotuna, line 45:
```suggestion
pub fn get_module_resolver(
  flags: &Flags,
  source_file: &str,
) -> impl Fn() -> file_watcher::FileWatcherFuture<
  file_watcher::ModuleResolutionResult<module_graph::Graph>,
> {
```

It'd be nice to use "impl Trait" instead of trait object :-)

cli/tools/bundler.rs
```diff
@@ -0,0 +1,141 @@
...
+  let flags = flags.clone();
+  let source_file = source_file.to_string();
+  let debug = flags.log_level == Some(log::Level::Debug);
+  Box::new(move || {
```
magurotuna, line 49: `Box` could be removed if the return type is `impl Fn`

cli/tools/bundler.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Returns a closure which takes the module graph and performs
+/// the bundling operation. The closure can be passed to file_watcher.
+pub fn get_operation(
+  flags: &Flags,
+  out_file: &Option<PathBuf>,
+) -> Box<
+  dyn Fn(
+    module_graph::Graph,
+  ) -> file_watcher::FileWatcherFuture<Result<(), AnyError>>,
+> {
```
magurotuna, line 111:
```suggestion
pub fn get_operation(
  flags: &Flags,
  out_file: &Option<PathBuf>,
) -> impl Fn(
  module_graph::Graph,
) -> file_watcher::FileWatcherFuture<Result<(), AnyError>> {
```

ditto

cli/tools/bundler.rs
```diff
@@ -0,0 +1,141 @@
...
+  let flags = flags.clone();
+  let out_file = out_file.clone();
+  Box::new(move |module_graph: module_graph::Graph| {
```
magurotuna, line 114: This `Box` also could be removed

magurotuna (changes requested):

cli/tools/bundler.rs
kt3k: Thank you! I didn't know `impl Trait`.

kt3k:
I tried to install some popular cli tools such as `aleph`, `pagic`, and `denon`, but all didn't work with bundling unfortunately...

`aleph` can be installed but the subcommands didn't work with bundling. The dynamic import at [this line](https://deno.land/x/aleph@v0.2.27/cli.ts#L163) seems unable to resolve with bundled version.

`pagic` can't be installed because of the errors from the type checker. Most errors say `TS17004 [ERROR]: Cannot use JSX unless the '--jsx' flag is provided.`. (The error can be reproducible with the command `deno bundle --unstable https://deno.land/x/pagic/mod.ts` with deno v1.6.0. Maybe this should be a separate issue?)

denon can be installed but didn't run because of the below runtime error:

```
error: Uncaught ReferenceError: Cannot access 'LogLevels1' before initialization
        const LogLevels1 = LogLevels1;
                           ^
    at file:///Users/kt3k/.deno/bin/denon.js:2021:28
    at file:///Users/kt3k/.deno/bin/denon.js:2274:6
    at file:///Users/kt3k/.deno/bin/denon.js:3536:2
```

This PR also supports `--no-bundle` option, so I think it can be shipped, but we need to warn the community about it before the release.

lucacasonato:
Lets postpone this until we have a month of swc bundler stabiliity.

kt3k:
closing this for now. Let's revisit this when it's desired.

PR #8727, opened 2020-12-11
refactor: rewrite lsp to use lspower

lucacasonato:
This PR investigates if we could use `tower-lsp` as our LSP server framework.

`tower-lsp` is an async LSP implementation written using `tower` (same framework `hyper` and `tonic` are built on). It uses the same underlying `lsp_types` crate for the protocol as `lsp_server`.

It has some nice usability benefits over `lsp_server`:
- It is async out of the box.
- It has built in notification and request routing.
- It has a less verbose API than `lsp_server`. 

TODO:
- [x] port over TSC and lint integration
- [x] port over tests
- [x] wait for https://github.com/ebkalderon/tower-lsp/pull/243 to land and be released, or publish to `tower-lsp`.
- [x] remove unused code


cli/Cargo.toml
```diff
@@ -79,6 +79,8 @@ tokio = { version = "0.2.22", features = ["full"] }
 tokio-rustls = "0.14.1"
 # Keep in-sync with warp.
 tokio-tungstenite = "0.11.0"
+# TODO(lucacasonato): go back to crates.io when ebkalderon/tower-lsp#243, ebkalderon/tower-lsp#244, and ebkalderon/tower-lsp#245 land, or when `lspower` is released.
+tower-lsp = { git = "https://github.com/lucacasonato/tower-lsp", rev = "0481ab5e092857dcca2f6854320a52a60738c3af" }
```
kitsonk, line 83: Probably should consider waiting on this until these become true, IMO.

kitsonk (commented):
Overall, it feels like the right track but a couple of things:

- renaming `lsp` to `lsp2` just makes the PR infinitely more difficult to read and feels totally arbitrary.  If we had to have two language servers running for a while, then we would want to use `lsp2` but this PR fully removes `lsp` in lieu of `lsp2`.
- some of the arbitrary renaming of things also makes it difficult to read, there is no problem with doing a rename PR as a seperate PR first.
- I really think we should hold off on it until we a stable crate published to replace lsp server.  There is no harm in waiting a little while to have the PRs accepted or the alternative crate published.

I know we landed the original LSP as a big ole PR, but it was hard to avoid that... I think we should really try to do things in smaller chunks now.

lucacasonato:
> renaming lsp to lsp2 just makes the PR infinitely more difficult to read and feels totally arbitrary. If we had to have two language servers running for a while, then we would want to use lsp2 but this PR fully removes lsp in lieu of lsp2.

Yes - I am going to remove lsp 1 in and move to rename lsp2 to lsp. It was just easier to copy paste chunks in small batches rather than porting everything in the same commit :-)

> Some of the arbitrary renaming of things also makes it difficult to read, there is no problem with doing a rename PR as a seperate PR first.

I dont think I did any major renames. Could you point me to what you mean?

> I really think we should hold off on it until we a stable crate published to replace lsp server. There is no harm in waiting a little while to have the PRs accepted or the alternative crate published.

Yup - but mainly a chicken and egg problem. I can get this out to crates.io, but I am not going to unless we agree that we want to go with `lsp-server`. Don't want to push a new crate to cargo, just to determine we won't actually use it.

> I know we landed the original LSP as a big ole PR, but it was hard to avoid that... I think we should really try to do things in smaller chunks now.

Agree - but I don't think that is possible with this PR. It completely replaces the core of the lsp. If you have ideas about how we could do this, I can split it up.

cli/Cargo.toml
lucacasonato: Yes this is a merge blocker.

kitsonk:
> I dont think I did any major renames. Could you point me to what you mean?

```diff
- server_state
+ state_snapshot
- ServerStateSnapshot
+ StateSnapshot
```

I'm fine with the change, but there were a lot of lines in this PR that were just "noise" around those renames.

> Yup - but mainly a chicken and egg problem.

How big would `lspower` be?  How much of `tower-lsp` would we not be using?  Does the server need to keep current with tokio?  In my mind if it is small, there are parts of `tower-lsp` we would never use, and/or we have to keep it current with other parts of Deno like tokio, then we should consider just having our own thing (or building the server directly into the lsp).

lucacasonato:
> How much of lspower would we not be using?

We are pretty much using all of lspower. We dont support all language server APIs, but there is no big feature we dont use.

> Does the server need to keep current with tokio?

It needs to update to 0.3 when we do. I checked, and this is trivial. Only 3 lines change compared to 0.2.

cli/lsp/README.md
```diff
@@ -1,23 +0,0 @@
...
-
-The structure of the language server was heavily influenced and adapted from
-[`rust-analyzer`](https://rust-analyzer.github.io/).
-
-## Structure
-
-When the language server is started, a `ServerState` instance is created which
-holds all the state of the language server, as well as provides the
-infrastructure for receiving and sending notifications and requests from a
-language server client.
```
ry, line 23: unintentional?

lucacasonato: Yeah. Whoops

cli/lsp/backend.rs
```diff
@@ -0,0 +1,905 @@
...
+use super::memory_cache::MemoryCache;
+use super::sources::Sources;
+use super::text;
+use super::text::apply_content_changes;
+use super::tsc;
+use super::tsc::TsServer;
+use super::utils;
+
+#[derive(Debug, Clone)]
+pub struct Backend {
```
kitsonk, line 41: nit: I really hate the term "backend"... `LanguageServer`?

kitsonk: or `DenoLanguageServer` so as to not conflict with the trait?

cli/lsp/lsp_extensions.rs
```diff
@@ -5,22 +5,10 @@
 ///!
 use deno_core::serde::Deserialize;
 use deno_core::serde::Serialize;
-use lsp_types::request::Request;
-use lsp_types::TextDocumentIdentifier;
+use lspower::lsp_types::TextDocumentIdentifier;
 
 #[derive(Debug, Deserialize, Serialize)]
```
kitsonk, line 10: probably should be moved to "backend" (which I still don't like the name) and this file deleted.

cli/lsp/utils.rs
```diff
@@ -1,71 +1,9 @@
...
-  e.downcast_ref::<Canceled>().is_some()
-}
-
-pub fn notification_is<N: lsp_types::notification::Notification>(
-  notification: &Notification,
-) -> bool {
-  notification.method == N::METHOD
-}
 
 /// Normalizes a file name returned from the TypeScript compiler into a URI that
```
kitsonk, line 8: Feels like this could be moved somewhere now, as we don't really have much here...

kitsonk: actually they are duplicated in `cli/ops/util.rs` now too, though they are only specific to lsp ops at the moment

kitsonk (commented):
Looking good... the term "backend" is really bothering me though.

cli/lsp/backend.rs
lucacasonato: Renamed to `DenoLanguageServer`

kitsonk (approved):
❤️ 

I guess it is just a decision about what we do about waiting for lspower.  I will defer to @ry on that one.

cli/lsp/backend.rs
```diff
@@ -38,7 +38,7 @@ use super::tsc::TsServer;
 use super::utils;
 
 #[derive(Debug, Clone)]
-pub struct Backend {
+pub struct DenoLanguageServer {
```
ry, line 41: How about just `LanguageServer`?

kitsonk: and then use `lspower::LanguageServer` for the trait?

lucacasonato: SGTM

lucacasonato:
lspower is published to crates now. Just need to switch the cargo toml line.￼

Also still need to fix the unit test in backend.rs and rename that file to language_server.rs.

PR #8729, opened 2020-12-12
Fix(doc): Remove a deprecated function call from docstring in Deno.permissions

### LINKED ISSUE #8682: Deno.dir is not a function
according to [runtime doc](https://doc.deno.land/builtin/unstable#Deno.Permissions), I code ```Deno.dir("home")```.
image

but output as follow:
```
Uncaught TypeError: Deno.dir is not a function
    at <anonymous>:2:6

```

env:
```
deno 1.5.4 (bc79d55, release, x86_64-apple-darwin)
v8 8.8.278.2
typescript 4.0.5
```

mitch292:
Resolves #8682 

Removed the function call and matched the style of the rejected console log rather than replacing with a different function call. Happy to change it.

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1143,7 +1143,7 @@ declare namespace Deno {
      * ```ts
      * const status = await Deno.permissions.request({ name: "env" });
      * if (status.state === "granted") {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, thanks @mitch292

PR #8732, opened 2020-12-12
docs: fix doc to remove annotation about removed function `Deno.dir`

### LINKED ISSUE #8731: Remove annotation about removed function
In connection with #8682, I suggest fix following docs as removing annotation about removed function.

- https://github.com/denoland/deno/blob/31935c6b8d8f763a5cf809b940427024da01a71c/cli/dts/lib.deno.ns.d.ts#L1082
- https://github.com/denoland/deno/blob/31935c6b8d8f763a5cf809b940427024da01a71c/cli/dts/lib.deno.ns.d.ts#L1101

yamatatsu:
This PR fix #8731 to remove annotaion about removed function `Deno.dir` like `(see also `Deno.dir("temp")`)`.


bartlomieju (approved):
LGTM, thanks @yamatatsu

PR #8738, opened 2020-12-12
feat(cli/standalone): Support runtime flags for deno compile

### LINKED ISSUE #8656: `deno compile` does not provide the ability to specify permissions or v8 flags
Currently all code run in `deno compile` effectively uses `--allow-all` permissions, and the default `--v8-flags`. This issue exists to investigate if there is demand for specifying permissions or v8 flags for `deno compile`.

nayeemrmn:
Supports permissions etc for executables created with `deno compile`, by appending serialized `Flags` after the source bundle.

Closes #8656.

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -283,14 +283,14 @@ impl FileFetcher {
     http_cache: HttpCache,
     cache_setting: CacheSetting,
     allow_remote: bool,
-    maybe_ca_data: Option<&str>,
+    ca_data: Option<Vec<u8>>,
   ) -> Result<Self, AnyError> {
     Ok(Self {
       allow_remote,
       cache: FileCache::default(),
       cache_setting,
       http_cache,
-      http_client: create_http_client(get_user_agent(), maybe_ca_data)?,
+      http_client: create_http_client(get_user_agent(), ca_data)?,
     })
   }
 
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -6,14 +6,20 @@ use clap::Arg;
 use clap::ArgMatches;
 use clap::ArgSettings;
 use clap::SubCommand;
+use deno_core::serde::de;
+use deno_core::serde::Deserialize;
+use deno_core::serde::Deserializer;
+use deno_core::serde::Serialize;
+use deno_core::serde::Serializer;
 use deno_runtime::permissions::PermissionsOptions;
 use log::Level;
+use std::fmt;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
 use tempfile::TempDir;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub enum DenoSubcommand {
   Bundle {
     source_file: String,
@@ -25,6 +31,7 @@ pub enum DenoSubcommand {
   Compile {
     source_file: String,
     output: Option<PathBuf>,
+    args: Vec<String>,
   },
   Completions {
     buf: Box<[u8]>,
@@ -92,7 +99,66 @@ impl Default for DenoSubcommand {
   }
 }
 
-#[derive(Clone, Debug, PartialEq, Default)]
+fn deserialize_maybe_log_level<'de, D>(d: D) -> Result<Option<Level>, D::Error>
+where
+  D: Deserializer<'de>,
+{
+  struct OptionalLogLevelVisitor;
+  impl<'de> de::Visitor<'de> for OptionalLogLevelVisitor {
+    type Value = Option<Level>;
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+      write!(formatter, "null or a valid log level string")
+    }
+
+    fn visit_none<E>(self) -> Result<Self::Value, E>
+    where
+      E: de::Error,
+    {
+      Ok(None)
+    }
+
+    fn visit_some<D>(self, d: D) -> Result<Self::Value, D::Error>
+    where
+      D: de::Deserializer<'de>,
+    {
+      struct LogLevelVisitor;
+      impl<'de> de::Visitor<'de> for LogLevelVisitor {
+        type Value = Level;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+          write!(formatter, "a valid log level string")
+        }
+
+        fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+        where
+          E: de::Error,
+        {
+          Level::from_str(s).map_err(|_| {
+            de::Error::invalid_value(de::Unexpected::Str(s), &self)
+          })
+        }
+      }
+      Ok(Some(d.deserialize_str(LogLevelVisitor)?))
+    }
+  }
+  d.deserialize_option(OptionalLogLevelVisitor)
+}
+
+fn serialize_maybe_log_level<S>(
+  maybe_level: &Option<Level>,
+  s: S,
+) -> Result<S::Ok, S::Error>
+where
+  S: Serializer,
+{
+  match maybe_level {
+    None => s.serialize_none(),
+    Some(level) => s.serialize_str(&level.to_string()),
+  }
+}
+
+#[derive(Clone, Debug, PartialEq, Default, Deserialize, Serialize)]
 pub struct Flags {
   /// Vector of CLI arguments - these are user script arguments, all Deno
   /// specific flags are removed.
@@ -117,6 +183,8 @@ pub struct Flags {
   pub inspect_brk: Option<SocketAddr>,
   pub lock: Option<PathBuf>,
   pub lock_write: bool,
+  #[serde(deserialize_with = "deserialize_maybe_log_level")]
+  #[serde(serialize_with = "serialize_maybe_log_level")]
   pub log_level: Option<Level>,
   pub no_check: bool,
   pub no_prompts: bool,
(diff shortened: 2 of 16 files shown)


cli/flags.rs
```diff
@@ -23,6 +29,7 @@ pub enum DenoSubcommand {
   Compile {
     source_file: String,
     output: Option<PathBuf>,
+    args: Vec<String>,
```
bartlomieju, line 32: This is surprising; it means compiled binaries can't have runtime args. What's the use case for having pre compiled CLI args?

nayeemrmn: > This is surprising; it means compiled binaries can't have runtime args. What's the use case for having pre compiled CLI args?

No, the runtime args are _appended_ to the ones passed at compile-time, like with `deno install`.

bartlomieju: Okay, good to know, can you update docs page as well?

bartlomieju:
@nayeemrmn anything holding back this PR?

nayeemrmn:
Done now, I still had to do ca file bundling.

cli/main.rs
```diff
@@ -295,15 +304,23 @@ async fn compile_command(
...
 ) -> Result<(), AnyError> {
   if !flags.unstable {
     exit_unstable("compile");
   }
 
   let debug = flags.log_level == Some(log::Level::Debug);
 
+  let preload_flags = Flags {
+    inspect: None,
+    inspect_brk: None,
```
bartlomieju, line 317: Maybe `inspect` and `inspect_brk` shouldn't be available to `compile` subcommand at all?

cli/standalone.rs
```diff
@@ -30,28 +35,39 @@ const MAGIC_TRAILER: &[u8; 8] = b"d3n0l4nd";
...
+    let flags_pos = u64_from_bytes(flags_pos)?;
+    let flags_len = u64_from_bytes(flags_len)?;
     current_exe.seek(SeekFrom::Start(bundle_pos))?;
 
-    let bundle_len = trailer_pos - bundle_pos;
-    let mut bundle = String::new();
-    current_exe.take(bundle_len).read_to_string(&mut bundle)?;
-    // TODO: check amount of bytes read
+    let bundle = read_string_slice(&mut current_exe, bundle_pos, bundle_len)?;
+    let flags = read_string_slice(&mut current_exe, flags_pos, flags_len)?;
```
bartlomieju, line 66: It'd be useful for debugging to add `.context()` before `?` with description what went wrong

cli/standalone.rs
```diff
@@ -61,6 +77,23 @@ pub fn try_run_standalone_binary(args: Vec<String>) -> Result<(), AnyError> {
   }
 }
 
+fn u64_from_bytes(arr: &[u8]) -> Result<u64, AnyError> {
+  let fixed_arr: &[u8; 8] = arr.try_into()?;
```
bartlomieju, line 81: It'd be useful for debugging to add .context() before ? with description what went wrong

cli/standalone.rs
```diff
@@ -106,28 +139,28 @@ impl ModuleLoader for EmbeddedModuleLoader {
...
+async fn run(source_code: String, flags: Flags) -> Result<(), AnyError> {
   let main_module = ModuleSpecifier::resolve_url(SPECIFIER)?;
-  let permissions = Permissions::allow_all();
+  let permissions = Permissions::from_options(&flags.clone().into());
   let module_loader = Rc::new(EmbeddedModuleLoader(source_code));
   let create_web_worker_cb = Arc::new(|_| {
     todo!("Worker are currently not supported in standalone binaries");
   });
 
+  v8_set_flags(
```
bartlomieju, line 150: Please add a note that it should be kept in sync with the identical call in `main.rs`

cli/standalone.rs
```diff
@@ -201,3 +250,60 @@ pub async fn create_standalone_binary(
...
+    no_check: false,
+    no_prompts: flags.no_prompts,
+    no_remote: false,
+    reload: false,
+    repl: false,
+    seed: flags.seed,
+    unstable: flags.unstable,
+    v8_flags: flags.v8_flags,
+    version: false,
+    watch: false, // TODO?
```
bartlomieju, line 307: Adding `watch` makes no sense in compiled binaries

cli/standalone.rs
```diff
@@ -201,3 +250,60 @@ pub async fn create_standalone_binary(
...
+  flags: Flags,
+  baked_args: Vec<String>,
+) -> Result<Flags, AnyError> {
+  let ca_file_inline = match flags.ca_file {
+    Some(ca_file) => Some(read(ca_file)?),
+    None => None,
+  };
+  // Don't abbreviate any of this to `..flags` or `..Default::default()`. That
+  // forces us to explicitly consider how any changes to `Flags` should be
+  // reflected here.
```
bartlomieju, line 272: Add `IMPORTANT: ` at the begging please :D

bartlomieju (commented):
@nayeemrmn awesome work! I have only a few minor comments but in general LGTM. 

I'll wait for @lucacasonato's review before landing

cli/file_fetcher.rs
```diff
@@ -283,14 +283,14 @@ impl FileFetcher {
     http_cache: HttpCache,
     cache_setting: CacheSetting,
     allow_remote: bool,
-    maybe_ca_file: Option<&str>,
+    ca_pem: Option<Vec<u8>>,
```
lucacasonato, line 286: Can we rename this to ca_data to match the impl in op_crates/fetch?

cli/flags.rs
```diff
@@ -108,6 +174,9 @@ pub struct Flags {
   pub allow_write: Option<Vec<PathBuf>>,
   pub cache_blocklist: Vec<String>,
   pub ca_file: Option<String>,
```
lucacasonato, line 110: Why keep this if we have ca_pem? I would prefer to have just one of the two.

cli/standalone.rs
```diff
@@ -30,28 +36,41 @@ const MAGIC_TRAILER: &[u8; 8] = b"d3n0l4nd";
...
 /// produced by `deno compile`. It determines if this is a stanalone
 /// binary by checking for the magic trailer string `D3N0` at EOF-12.
-/// After the magic trailer is a u64 pointer to the start of the JS
-/// file embedded in the binary. This file is read, and run. If no
-/// magic trailer is present, this function exits with Ok(()).
+/// The magic trailer is followed by:
+/// - a u64 pointer to the JS bundle embedded in the binary
+/// - the u64 length of the bundle
+/// - a u64 pointer to JSON metadata (serialized flags) embedded in the binary
+/// - the u64 length of the metadata
```
lucacasonato, line 43: I don't think this is needed.

cli/main.rs
nayeemrmn: > Maybe `inspect` and `inspect_brk` shouldn't be available to `compile` subcommand at all?

I included them because it seems simpler to group them into the wider set of "flags that affect runtime behaviour", and match `deno install` in doing so. On the other hand, is there potential to remove the inspector from the light runtime binary?

caspervonb: > On the other hand, is there potential to remove the inspector from the light runtime binary?

We can feature gate it on our side; it will break features like the REPL but those seem irrelevant for compiled bundles.

lucacasonato (commented):
Just a few comments.

I don't like the mix of compile args and runtime args. Same goes for `deno install` though, so it is fine (not a blocker).

cli/main.rs
bartlomieju: > On the other hand, is there potential to remove the inspector from the light runtime binary?

Not in the near future, no.

> We can feature gate it on our side; it will break features like the REPL but those seem irrelevant for compiled bundles.

REPL is not part of `runtime/` directory, it's `cli/` only so that should work fine

cli/standalone.rs
nayeemrmn: I thought the redundant values would make the extraction more readable, but it's not that different. Switching to only use the pointers.

cli/flags.rs
nayeemrmn: We'd have to read the file in flag parsing, and it wouldn't correspond well to the actual CLI flags. ~`ca_pem`~ `ca_data` is the one that doesn't belong here, but it's a particularly convenient hack.

cli/main.rs
nayeemrmn: Doh, apparently I already went out of my way to exclude the inspector flags in `flags.rs` 🤦‍♂️ I'll just remove this meaningless handling.

cli/main.rs
```diff
@@ -134,7 +135,9 @@ fn create_web_worker_callback(
       args.main_module,
       args.worker_id,
       &options,
-    );
+    )
+    // FIXME(nayeemrmn): Don't unsafely unwrap here.
+    .unwrap();
```
nayeemrmn, line 140: Fix before landing.

I've moved things around a bit so this is the only one -- @bartlomieju any ideas on how to return a `Result` in this section?

nayeemrmn (commented):
I've removed the `Flags::ca_data` hack, since nothing that directly reads program state's flags should be used in a compiled runtime.

cli/main.rs
bartlomieju: @nayeemrmn you'll have to change type of `CreateWebWorkerCb`, I'm fine with that

runtime/web_worker.rs
```diff
@@ -137,7 +139,8 @@ pub struct WebWorkerOptions {
   pub args: Vec<String>,
   pub debug_flag: bool,
   pub unstable: bool,
-  pub ca_filepath: Option<String>,
+  pub ca_file: Option<String>,
```
lucacasonato, line 142: Why is this needed in the `WebWorkerOptions`? Can't we read the file during initialization?

runtime/worker.rs
```diff
@@ -45,7 +47,8 @@ pub struct WorkerOptions {
   pub args: Vec<String>,
   pub debug_flag: bool,
   pub unstable: bool,
-  pub ca_filepath: Option<String>,
+  pub ca_file: Option<String>,
```
lucacasonato, line 50: Ditto

cli/main.rs
nayeemrmn: Done, it's read and stored in `ProgramState`.

lucacasonato (approved):
LGMT. Very clean implementation now :+1:

bartlomieju (approved):
LGTM

Issue #8190, opened 2020-10-29
feat: Passing certificate data directly to HttpClient


danopia:
In #6918 this signature was added:

```typescript
  /** **UNSTABLE**: New API, yet to be vetted.
   * The options used when creating a [HttpClient].
   */
  interface CreateHttpClientOptions {
    /** A certificate authority to use when validating TLS certificates.
     *
     * Requires `allow-read` permission.
     */
    caFile?: string;
  }
```

## Problem

When connecting to a Kubernetes API from within a cluster, `caFile` is useful, because there's a `ca.pem` file ready to go. However, when connecting from a workstation, the CA data is generally nested inside a YAML file instead:

```
> cat ~/.kube/config
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JS....
```

In order to use this CA with `Deno.createHttpClient` I have to extract the CA certificate and write it out to disk for Deno to read back in:

```typescript
    const caPath = `/tmp/ca-${clusterName}.pem`;
    await Deno.writeTextFile(caPath, caData);
    const httpClient = Deno.createHttpClient({
      caFile: caPath,
    });
```

... and now I need `--allow-write` as well as `--allow-read` (and `--allow-net`).

## Solution

There should be a separate parameter ([in nodejs it's called `ca`](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options), and something like `caData` could also work) that allows us to pass certificate data in directly.

```typescript
    const httpClient = Deno.createHttpClient({
      ca: caData,
    });
```

The `key` and `cert` fields, when they are added (mentioned in #3444), should take similar consideration for inline data.

yonbav:
Hi @danopia I've started to work on that issue hope it will be fixed soon.

danopia:
Awesome, I'm looking forward to progress on this :crab: :pray:

danopia:
Looks like this is merged and will be in the next Deno release; I'll close now in anticipation of it shipping. Thanks!!

danopia:
I see this in 1.6.2:

```typescript
    const { cluster } = config.fetchCurrentContext();
 
    const caPath = `/tmp/ca-${context.cluster}.pem`;
    await Deno.writeTextFile(caPath, atob(cluster["certificate-authority-data"]));
 
    const httpClient = Deno.createHttpClient({
      caFile: caPath,
    });
```

⬇️ 

[...]

PR #8739, opened 2020-12-12
feat(unstable): support in memory certificate data for creating httpClient

yonbav:
- Add unit test that recieve certificate data as parameter
- Add caData to CreateHttpClientOptions
- fetch/lib.rs: create the certificate from ca_data

issue: https://github.com/denoland/deno/issues/8190


bartlomieju:
@ry @lucacasonato I think this PR is great and I'd prefer this approach to taking a path to CA file. So I suggest we land this one and remove `caPath` in a follow up PR, that way `deno_fetch` needn't be aware of checking path permission.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1266,6 +1266,7 @@ declare namespace Deno {
      * Requires `allow-read` permission.
      */
     caFile?: string;
+    caData?: string;
```
lucacasonato, line 1269: Can you add a JSDoc comment here (this doesn't need allow-read). Also could you mention this takes a PEM encoded certificate?

op_crates/fetch/lib.rs
```diff
@@ -263,13 +266,20 @@ where
...
+  ca_file: Option<&str>,
+  ca_data: Option<&str>,
+) -> Result<Client, AnyError> {
   let mut builder = Client::builder().redirect(Policy::none()).use_rustls_tls();
   if let Some(ca_file) = ca_file {
     let mut buf = Vec::new();
     File::open(ca_file)?.read_to_end(&mut buf)?;
     let cert = reqwest::Certificate::from_pem(&buf)?;
     builder = builder.add_root_certificate(cert);
+  } else if let Some(ca_data) = ca_data {
```
lucacasonato, line 279: `ca_data` should have priority over `ca_file`.

lucacasonato (commented):
Very nice. I agree with removing caPath in a follow-up PR. Just a few comments.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1266,6 +1266,12 @@ declare namespace Deno {
      * Requires `allow-read` permission.
      */
     caFile?: string;
+
+    /** A certificate authority to use when validating TLS certificates.
+     *
+     * Pem format certificate data
```
lucacasonato, line 1272:
```suggestion
    /** A certificate authority to use when validating TLS certificates. Certificate data must be PEM encoded.
```

lucacasonato (approved):
LGTM. Thanks!

Issue #7824, opened 2020-10-04
Uncaught AddrInUse: when using --unstable --watch


rizkyramadhan:
I tried running simple web server on deno latest:

```
import { serve } from 'https://deno.land/std/http/server.ts';
const s = serve({ port: 8000 });
console.log('http://localhost:8000/');
for await (const req of s) {
  const file = await Deno.create('f.txt');
  await Deno.copy(req.body, file);
  req.respond({ body: 'Hello  asd asWorld\n' });
}
```

when I saved some changes, file watcher is restarting my app, and then error:

```
Watcher File change detected! Restarting!
error: Uncaught AddrInUse: Only one usage of each socket address (protocol/network address/port) is normally permitted. (os error 10048)
    at Object.jsonOpSync (core.js:247:13)
    at opListen (deno:cli/rt/30_net.js:32:17)
    at Object.listen (deno:cli/rt/30_net.js:207:17)
    at Object.listen (deno:cli/rt/40_net_unstable.js:11:18)
    at serve (https://deno.land/std@0.73.0/http/server.ts:287:25)
    at <file:///C:/Users/Riz/Desktop/a/a.ts>:2:1
Watcher Process terminated! Restarting on file change...
```

I am on windows 10,

MohammadHonarvar:
I'm using windows 10 too and i have the same problem.
Can somebody help us?

fauzan121002:
I'm also getting this issue by Oak Server  , windows 10 , can somebody help?

magurotuna:
@rizkyramadhan @MohammadHonarvar @fauzan121002 
Do this error still happen to you all when using Deno v1.6.0? If so, I'll dig into it.

stefanreimers:
@magurotuna 
I can confirm this is still happening in v1.6.0.
Environment is Windows 10.

Script is
```
import { serve } from "https://deno.land/std@0.80.0/http/server.ts";
const s = serve({ port: 8000 });

for await (const req of s) {
  req.respond({ body: "Hello World\n" });
}
```

Call is:
```
deno run --allow-net --unstable --watch server.ts
```

Any change to the script while in watch mode causes the server to crash.

magurotuna:
@stefanreimers Thanks for the info! I'll look into it although it seems to be a window specific issue and I don't have a window machine.

MohammadHonarvar:
By upgrade to version 1.7.0, this issue solved.
Thanks DENO team.

jtoppine:
[...]

PR #8743, opened 2020-12-13
refactor(cli): add async version of resolve_addr function

magurotuna:
While I'm looking into #7824, I found `to_socket_addrs`, which is a blocking API, was called in `resolve_addr` function.
So this commit adds a new function that is an asynchronous version of `resolve_addr` using `tokio::net::lookup_host`.
And accordingly, renames the synchronous version to `resolve_addr_sync`.
This allows async ops to resolve hosts with non-blocking.


ry (commented):
Nice work. Thanks!

It would be cool to expose these as standalone ops!

I am concerned that the asynchronous version does not go through the system's `getaddrinfo()` and potentially ignores things like `/etc/hosts`. Does `tokio::net::lookup_host` call `getaddrinfo()` in a thread pool (with `spawn_blocking`) or is it using a separate asynchronous DNS stack?

magurotuna:
@ry 

> It would be cool to expose these as standalone ops!

Sure, I'll do that.

> I am concerned that the asynchronous version does not go through the system's getaddrinfo() and potentially ignores things like /etc/hosts. Does tokio::net::lookup_host call getaddrinfo() in a thread pool (with spawn_blocking) or is it using a separate asynchronous DNS stack?

Good point. Looking at tokio's implementation, it eventually calls `std::net::ToSocketAddrs` with `spawn_blocking`.
Refs:
- the lastet version(0.3.6): https://docs.rs/tokio/0.3.6/src/tokio/net/addr.rs.html#207
- the version Deno is currently using(0.2.22): https://docs.rs/tokio/0.2.22/src/tokio/net/addr.rs.html#199

ry (approved):
LGTM - thanks!

PR #8749, opened 2020-12-14
WIP: add failing test case for Worker API

### LINKED ISSUE #8702: Worker onmessage handler not being invoked when payload is undefined
Example:

`worker.ts`:
```ts
self.addEventListener('message', ev => {
  try {
    const {data} = ev as MessageEvent;
    console.log('worker <- main:', data);
    const value = data > 0 ? data * 2 : undefined;
    (self as any).postMessage(value);
    console.log('worker -> main:', value);
  }
  catch (ex) {
    console.error(ex);
  }
});
```

`main.ts`:
```ts
const data = 0;

const main = async (): Promise<void> => {
  let resolve: (value: number) => void;

  const workerUrl = new URL('worker.ts', import.meta.url).href;
  const worker = new Worker(workerUrl, {deno: true, type: 'module'});

  const handleWorkerMessage = (ev: MessageEvent): void => {
    const {data} = ev;
    console.log('main <- worker:', data);
    resolve(data);
    worker.terminate();
  };

  worker.addEventListener('message', handleWorkerMessage);
  worker.addEventListener('messageerror', () => console.log('message error'));
  worker.addEventListener('error', () => console.log('error'));

  const result = await new Promise(res => {
    resolve = res;
    worker.postMessage(data);
    console.log('main -> worker:', data);
  });

  console.log(result);
};

if (import.meta.main) main();
```

When run, it just hangs forever after the worker posts `undefined` back to the main process because the handler is never invoked:

```
example % deno run --unstable --allow-read=. main.ts
main -> worker: 0
worker <- main: 0
worker -> main: undefined
^C
```

If—on line 1 of `main.ts`—`data` is changed to a number greater than `0`, the worker will post back a `number` instead of `undefined` and everything works as expected:

```
example % deno run --unstable --allow-read=. main.ts
Check file:///Users/jesse/example/main.ts
main -> worker: 1
worker <- main: 1
worker -> main: 2
main <- worker: 2
2
```

I'm not sure why this behavior is happening, but `undefined` is a meaningful value, and this seems like a bug.

wperron:
Fixes #8702 

cc @bartlomieju


bartlomieju:
Interesting, `worker_post_undefined` works locally on macOS, need to investigate more.

bartlomieju:
@wperron I'm gonna close this PR in favor of https://github.com/denoland/deno/pull/8810 which uses a unit test instead of integration one

PR #8750, opened 2020-12-14
fix(cli): make dynamic import errors catchable

### LINKED ISSUE #6259: Errors in dynamic imports are not properly caught
Runtime
```
debian 10
deno 1.6.2
v8 8.8.278.2
typescript 4.1.3
```
------

```ts
//a.ts
import("./b.ts").catch(e=>{
  console.error(e)
})

import("./c.ts").catch(e=>{
  console.error(e)
})

setInterval(function () {
    console.error(e);
  });
}, 1e3);
```
```ts
//b.ts
// @ts-ignore
import "./b1.ts";
export default () => "error";
```
```ts
//c.ts
// @ts-ignore
await import("./c1.ts");
export default () => "error";
```

```
caught import b.ts error
    at file://yourdir/b.ts:3:0
caught import c.ts error
```

kitsonk:
Fixes #6259


cli/disk_cache.rs
```diff
@@ -105,31 +105,25 @@ impl DiskCache {
...
         out = out.join(remaining_components);
       }
-      scheme => {
-        unimplemented!(
-          "Don't know how to create cache name for scheme: {}\n  Url: {}",
-          scheme,
-          url
-        );
-      }
+      _ => return None,
```
kitsonk, line 108: Having this panic was causing other problems where we "assumed" the import specifier was valid, where we really check it in the file_fetcher.

cli/tests/integration_tests.rs
```diff
@@ -3202,6 +3202,11 @@ itest!(tsx_imports {
   output: "tsx_imports.ts.out",
 });
 
+itest!(fix_dynamic_import_errors {
+  args: "run --reload fix_dynamic_import_errors.mjs",
+  output: "fix_dynamic_import_errors.mjs.out",
+});
```
ry, line 3208: Is the .mjs extension significant? If not maybe use .js

cli/tests/fix_dynamic_import_errors.mjs.out
```diff
@@ -0,0 +1,2 @@
+caught import error from b.mjs
+caught import error from c.mjs
```
ry, line 2: is the order of these deterministic?

cli/tests/unsupported_dynamic_import_scheme.out
```diff
@@ -1,4 +1,4 @@
-error: Unsupported scheme "xxx" for module "xxx:". Supported schemes: [
+error: Uncaught (in promise) TypeError: Unsupported scheme "xxx" for module "xxx:". Supported schemes: [
     "http",
```
bartlomieju, line 2: Doh, looks like this test was wrong for a long time...

cli/tests/error_005_missing_dynamic_import.ts.out
```diff
@@ -1,2 +1 @@
-error: Cannot resolve module "[WILDCARD]/bad-module.ts" from "[WILDCARD]/error_005_missing_dynamic_import.ts".
-    at file:///[WILDCARD]/error_005_missing_dynamic_import.ts:3:26
+error: Uncaught (in promise) TypeError: Cannot resolve module "[WILDCARD]/cli/tests/bad-module.ts".
```
bartlomieju, line 1: Is is possible to keep `at <location>`?

cli/tests/017_import_redirect_info.out
```diff
@@ -1,6 +1,6 @@
 local: [WILDCARD]017_import_redirect.ts
 type: TypeScript
-deps: 1 unique (total [WILDCARD]B)
+deps: 2 unique (total [WILDCARD]B)
```
bartlomieju, line 3: What's the reason for this change? It seems redirects are now counted as unique deps?

cli/module_graph.rs
```diff
@@ -667,6 +688,41 @@ pub struct Graph {
...
+          }),
+        }
+      } else {
+        match module.media_type {
+          MediaType::JavaScript | MediaType::Unknown => Ok(ModuleSource {
+            code: module.source.clone(),
+            module_url_found: module.specifier.to_string(),
+            module_url_specified: specifier.to_string(),
+          }),
+          _ => Err(anyhow!("Compiled module not found \"{}\"", specifier)),
```
bartlomieju, line 715: Using `anyhow!` will result in generic `Error` class, shouldn't we use `NotFound` instead here?

cli/module_graph.rs
```diff
@@ -1050,7 +1107,12 @@ impl Graph {
     totals: &mut HashMap<ModuleSpecifier, usize>,
   ) -> ModuleInfo {
     let not_seen = seen.insert(specifier.clone());
-    let module = self.get_module(specifier).unwrap();
+    let module = if let ModuleSlot::Module(module) = self.get_module(specifier)
+    {
+      module
+    } else {
+      panic!("missing module");
```
bartlomieju, line 1114: I presume this branch shouldn't be reachable, maybe use `unreachable!` instead?

cli/tests/integration_tests.rs
kitsonk: It was because of the original test case, and when it started wasn't sure if it was related to the media type or not, but yeah, it isn't, so we can switch it back.

cli/tests/error_005_missing_dynamic_import.ts.out
kitsonk: Maybe, I will look into it... of course if the promise is caught, it should have an error stack. I wonder why we don't print error stacks for uncaught promises... 🤔

cli/tests/017_import_redirect_info.out
kitsonk: hmmmm... yeah, I guess that isn't right... I will try to fix that...  need to count unique `Module` module slots.

cli/tests/error_005_missing_dynamic_import.ts.out
bartlomieju: That's a great question 🤔  it's really strange, @nayeemrmn any ideas?

cli/tests/fix_dynamic_import_errors.mjs.out
kitsonk: The spec for dynamic imports is unclear from [what I am reading](https://tc39.es/proposal-dynamic-import/#sec-hostimportmoduledynamically):

> The actual process performed is implementation defined, but typically consists of performing whatever I/O operations are necessary to allow HostResolveImportedModule to synchronously retrieve the appropriate Module Record, and then calling its Evaluate concrete method. This might require performing similar normalization as HostResolveImportedModule does

Since it isn't required by the spec, I will wildcard them, just so we don't end up with a fragile case that is still technically valid.

cli/tests/error_005_missing_dynamic_import.ts.out
kitsonk: confirmed it is there, logging out a caught error:

```
TypeError: Cannot resolve module "file:///Users/kitsonk/github/deno/cli/tests/dynamic_import/bad.mjs" from "file:///Users/kitsonk/github/deno/cli/tests/dynamic_import/b.js".
    at file:///Users/kitsonk/github/deno/cli/tests/dynamic_import/b.js:1:0
```

I opened #8769.

ry (approved):
LGTM - good fix

PR #8756, opened 2020-12-14
Factor out common code into core/build_util.rs

ry:


bartlomieju (approved):
LGTM, looks like you hit a flaky test :/

PR #8759, opened 2020-12-14
fix(inspector) recover from invalid input

caspervonb:
Continues where @benjamingr left off on #8379


ry (approved):
I verified the test fails on master.

LGTM - thanks Casper

PR #8763, opened 2020-12-14
Revert "build: Factor out common code into core/build_util.rs (#8756)"

bartlomieju:
This reverts commit 502c77aad969fe8557aaf6d56432f94450c26a2a.


ry (approved):
I assume this is because of the `env!("CARGO_MANIFEST_DIR")`? I'll fix this up later after the release.

bartlomieju:
> I assume this is because of the `env!("CARGO_MANIFEST_DIR")`? I'll fix this up later after the release.

Yes, alternative solution in #8764

PR #8771, opened 2020-12-15
refactor(lsp): optimise static assets

### LINKED ISSUE #8158: Refactor get_asset and static assets in cli
We need to refactor how we inline the static libraries in Deno when we fully migrate over from the old compiler infrastructure.

kitsonk:
Fixes #8158

### DIFF
--- a/cli/lsp/handlers.rs
+++ b/cli/lsp/handlers.rs
@@ -31,8 +31,11 @@ fn get_line_index(
   specifier: &ModuleSpecifier,
 ) -> Result<Vec<u32>, AnyError> {
   let line_index = if specifier.as_url().scheme() == "asset" {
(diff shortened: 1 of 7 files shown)


ry (commented):
Looks good. I defer to Luca for detailed review.

Unfortunately the size optimization isn't huge, but it's better:
```
~/src/deno> ls -l target/release/deno
-rwxr-xr-x  2 ry  staff  50277153 Dec 15 05:02 target/release/deno
~/src/deno> ls -l `which deno` # v1.6.1
-rwxr-xr-x  1 ry  staff  50653473 Dec 14 15:46 /Users/ry/.deno/bin/deno
```

lucacasonato (approved):
LGTM

PR #8779, opened 2020-12-15
upgrade: tokio 1.0

bartlomieju:
This commit upgrades all crates to "tokio 1.0" along with
crates from tokio ecosystem.

PR #8790, opened 2020-12-16
feat(unstable): add Deno.resolveDns API

### LINKED ISSUE #6110: Is DNS querying/resolution exposed?
Sorry, this isn't an _issue_ per-say, but I'm looking to do DNS queries in deno, and not finding anything about it in the Runtime API docs nor the std. Google isn't helping either.

Am I overlooking somewhere? Or is DNS not yet exposed in deno? Thanks!

magurotuna:
This PR adds new ops, `op_resolve_addr_sync` and `op_resolve_addr_async` which do DNS resolution, and exposes them as `Deno.resolveAddrSync` and `Deno.resolveAddr` respectively.
It comes from @ry's comment: https://github.com/denoland/deno/pull/8743#pullrequestreview-552294380, and also resolves #6110.

For manual testing, I prepared a script like:

```ts
const a = await Deno.resolveAddr("example.com", 80);
const b = Deno.resolveAddrSync("example.com", 53);
console.log(a, b);
```

then ran it via `cargo run -- run --allow-net foo.ts`, I got:

```
[ "[2606:2800:220:1:248:1893:25c8:1946]:80", "93.184.216.34:80" ] [ "[2606:2800:220:1:248:1893:25c8:1946]:53", "93.184.216.34:53" ]
```

Seeing this result, I wonder if we need really to pass a port number to `Deno.resolveAddr`. As far as I understand, domain name resolution itself has nothing to do with a port number, but [`std::net::ToSocketAddr`](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html) and [`tokio::net::lookup_host`](https://docs.rs/tokio/0.2.23/tokio/net/fn.lookup_host.html) require a port number as well as hostname.
If `Deno.resolveAddr` and `Deno.resolveAddrSync` should be specialized in resolving domain name, it would be nice for the functions not to receive port numbers, and to return resolved addresses with port numbers truncated.


bartlomieju:
CC @mrkurt thoughts?

magurotuna:
There's a widely used dns resolver crate: https://docs.rs/trust-dns-resolver/0.19.6/trust_dns_resolver/
It might be better to use it.

mrkurt:
I think maybe the `resolve_addr` Deno already has is wrong for this. That's meant to turn a `host:port` address into something useful for listen/dial. The resolution happens under the covers, but the result isn't really a DNS response (ipv6 addresses wouldn't normally be in `[]` for example).

What I'd like is the equivalent of this `dig` command:

```
dig aaaa nrt.curl.internal @10.0.0.1 +short
```

That queries AAAA records on the `10.0.0.1` DNS server records for `nrt.curl.internal`. If the underlying op can accept record type, name, and DNS server addresses, then return the list of results, I think the TypeScript API can do most of the work with just one op.

magurotuna:
@mrkurt 
I see, then it seems like we should utilize `trust_dns_resolver` crate instead. I'll give it a shot.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -1783,6 +1783,37 @@ declare namespace Deno {
...
+   * Fails if the hostname is in invalid format or the port is outside the
+   * range of 16-bit unsigned integer.
+   *
+   * ```ts
+   * const addrs = Deno.resolveAddrSync("example.com", 80);
+   * ```
+   *
+   * Requires `allow-net` permission.
+   */
+  export function resolveAddrSync(hostname: string, port: number): string[];
```
bartlomieju, line 1815: Since it's a new API it should go to `lib.deno.unstable.d.ts` and require `--unstable` flag

bartlomieju (changes requested):

cli/dts/lib.deno.ns.d.ts
magurotuna: Done.

runtime/ops/net.rs
```diff
@@ -528,3 +544,141 @@ fn op_listen(
...
+    protocol: Protocol,
+  }
+
+  let ResolveAddrArgs { query, options } = serde_json::from_value(args)?;
+
+  {
+    let s = state.borrow();
+    // TODO(magurotuna): what permission to check?
+    s.borrow::<Permissions>().check_net("TODO", 0)?;
+  }
```
magurotuna, line 638: I'm wondering what permission to check in this case. 
`check_net` requires hostname and port as arguments, but I think it would make sense to check just if `allow-net` is present, while neither hostname nor port is always necessary.

runtime/ops/net.rs
```diff
@@ -47,6 +54,7 @@ pub fn init(rt: &mut deno_core::JsRuntime) {
   super::reg_json_sync(rt, "op_listen", op_listen);
   super::reg_json_async(rt, "op_datagram_receive", op_datagram_receive);
   super::reg_json_async(rt, "op_datagram_send", op_datagram_send);
+  super::reg_json_async(rt, "op_resolve_addr", op_resolve_addr);
```
magurotuna, line 57: I feel like there's more appropriate name... like `op_dns_resolve`, because `op_resolve_addr` is a bit ambiguous.
 Anyone have thoughts on this?

bartlomieju: @magurotuna `op_dns_resolve` sounds good

magurotuna: Okay, done :)

runtime/ops/net.rs
mrkurt: You could check for resolver address + port 53. That might be difficult when you want to use the system resolver, though.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -901,6 +901,57 @@ declare namespace Deno {
    */
   export function shutdown(rid: number, how: ShutdownMode): Promise<void>;
 
+  /** The type of the resource record.
+   * Only the listed types are supported currently. */
+  export type RecordType =
```
mrkurt, line 906: Easy to add later, but many people doing DNS lookups directly will want TXT and SRV records too.

bartlomieju: @magurotuna let's address that before landing

runtime/ops/net.rs
bartlomieju: We might also add another method to `Permissions` if that helps

runtime/ops/net.rs
```diff
@@ -267,7 +278,10 @@ async fn op_connect(
           .borrow::<Permissions>()
           .check_net(&args.hostname, args.port)?;
       }
-      let addr = resolve_addr(&args.hostname, args.port).await?;
+      let addr = resolve_addr(&args.hostname, args.port)
+        .await?
+        .next()
+        .ok_or_else(|| generic_error("No resolved address found"))?;
```
bartlomieju, line 284: Seems like unrelated change, do you need to change `resolve_addr` and `resolve_addr_sync` implementations in this PR?

bartlomieju (commented):
@magurotuna looks great! Could we get some integration tests for this feature?

@mrkurt thanks for the review, if you have any suggestions regarding missing functionality that would be helpful (it seems you'll be the first user of this API 😄 )

runtime/ops/net.rs
magurotuna: Not really, we can revert them, while these changes can be considered as small refactoring. That is, the changes about `resolve_addr` and `resolve_addr_sync` in this PR are the following:
- to let `resolve_addr` and `resolve_addr_sync` return iterators of `SocketAddr`, not just a single `SocketAddr`. These functions previously called `.next()` on an iterator internally, and returned it, so the callers of these functions had no way of handling other items than the first item in the iterator.  This change allows the caller to decide how to use the returned iterator.
- to use `generic_error` in deno_core instead of `anyhow::Context`. When I worked on #8743, I didn't know there were utility error functions like `generic_error`. 

Would it be better to revert it, or extract it and open a new PR?

bartlomieju: @magurotuna if that's not a big deal maybe extract to separate PR?

runtime/ops/net.rs
magurotuna: @mrkurt Thanks! That makes sense. So, if a user calls `Deno.resolveDns` with (a) name server address and/or (b) its port number specified, then we will check the permission like `check_net(a, b)`.

cli/dts/lib.deno.unstable.d.ts
magurotuna: @mrkurt 
`TXT` and `SRV` cannot easily be transferred from Rust to JS due to a quirk of `trust_dns` crate. 

As far as I understand, the content of `TXT` record can not always be represented as Rust's string. So I think that if we send it to JS, it will become `Uint8Array` in JS. Will it be fine for those who want DNS lookups?

And, as of `SRV`, what format would users find useful? One simple format is something like `<priority> <weight> <port> <target>` (separated by space), do you think this is good?

runtime/ops/net.rs
magurotuna: @bartlomieju sure!

cli/dts/lib.deno.unstable.d.ts
lucacasonato: I would rather have structured data for SRV.

magurotuna: @lucacasonato I see, so the following should be better

```ts
{
  priority: number;
  weight: number;
  port: number;
  target: string;
}
```

lucacasonato: Yeah - same goes for `MX` records in the future.

magurotuna: It shouldn't be so hard to add `MX` support in this PR, so I'll do that 👍

runtime/ops/net.rs
magurotuna: #8891

runtime/ops/net.rs
```diff
@@ -528,3 +546,302 @@ fn op_listen(
...
+      perm.check_net(ip_addr, port)?;
+    } else {
+      perm.check_net_all()?;
+    }
+  }
+
+  let resolver = if let Some(options) = options {
+    let conf = if let Some(name_server) = options.name_server {
+      let mut c = ResolverConfig::new();
+      c.add_name_server(NameServerConfig {
```
bnoordhuis, line 684: This adds the server to the end of the list but I think that means the default ones will be queried first. It'd be better to construct the Resolverconfig like this:
```rust
// Don't try other servers on SERVFAIL or NXDOMAIN - I *think* that's appropriate
let trust_nx_responses = true;
let group = NameServerConfigGroup::from_ips_clear(&[name_server], 53, trust_nx_responses);
let c = Resolverconfig::from_parts(None, vec![], group);
```

runtime/ops/net.rs
```diff
@@ -528,3 +546,302 @@ fn op_listen(
...
+        tls_dns_name: None,
+      });
+      c
+    } else {
+      let (c, _) = read_system_conf()?;
+      c
+    };
+    AsyncResolver::tokio(conf, ResolverOpts::default()).await?
+  } else {
+    AsyncResolver::tokio_from_system_conf().await?
```
bnoordhuis, line 699: It would prefer it if this was restructured to have just a single code path, rather than two, that look up the system resolver. Easier to reason about.

As an aside: this ends up parsing `/etc/resolv.conf` for every DNS lookup. Probably not a big deal but I'd be remiss if I didn't point it out.

runtime/ops/net.rs
```diff
@@ -528,3 +546,302 @@ fn op_listen(
...
+        .map(ToString::to_string)
+        .map(DnsReturnRecord::PTR),
+      SRV => r.as_srv().map(|srv| DnsReturnRecord::SRV {
+        priority: srv.priority(),
+        weight: srv.weight(),
+        port: srv.port(),
+        target: srv.target().to_string(),
+      }),
+      TXT => r.as_txt().map(|txt| {
+        // TODO(magurotuna): parses bytes as UTF-8, is it alright?
```
bnoordhuis, line 746: Good question. I think the RFC suggests that octets should be in the range 0x20..=0x7E but the real world is probably more... incalculable. I'd play it safe and interpret it as Latin-1, example here:
https://github.com/denoland/deno/blob/b1230a85e8ba63f3542892711043cefdc916acf8/op_crates/fetch/lib.rs#L173-L177

bnoordhuis (commented):
Thanks for the PR. It mostly LGTM but there are some caveats that need to be hashed out.

1. The way it's currently set up, it'll use Google's DNS servers (8.8.8.8 and the others) at least some of the time. I imagine privacy-minded people won't approve.

2. trustdns doesn't really integrate with the system DNS resolver the way getaddrinfo(3) does. That means e.g. mDNS on Linux or macOS won't work (reliably.)

(2) is the reason Node.js has both `dns.resolve()` and `dns.lookup()` - one uses the system resolver (slow but versatile) while the other uses c-ares (fast but limited.)

runtime/ops/net.rs
```diff
@@ -528,3 +546,302 @@ fn op_listen(
...
+      .map(|ns| (&ns.ip_addr, ns.port));
+
+    let s = state.borrow();
+    let perm = s.borrow::<Permissions>();
+
+    if let Some((ip_addr, port)) = perm_args {
+      perm.check_net(ip_addr, port)?;
+    } else {
+      perm.check_net_all()?;
+    }
```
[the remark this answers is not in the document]
magurotuna: That makes much sense. I will fix it!

runtime/ops/net.rs
magurotuna: > This adds the server to the end of the list

Oh I didn't know that, I will adopt your way and see if it works as we expect.

runtime/ops/net.rs
magurotuna: You are right, I will try to refactor it.

runtime/ops/net.rs
magurotuna: I see, Latin-1 seems reasonable.

magurotuna:
@bnoordhuis 

> The way it's currently set up, it'll use Google's DNS servers (8.8.8.8 and the others) at least some of the time. I imagine privacy-minded people won't approve.

Will this issue be resolved by your suggestion https://github.com/denoland/deno/pull/8790#discussion_r550172074, right?

> trustdns doesn't really integrate with the system DNS resolver the way getaddrinfo(3) does. That means e.g. mDNS on Linux or macOS won't work (reliably.)

That's correct, so we should probably follow how Node.js does. This PR introduces`Deno.resolveDns`, and I'd like to open a new PR to implement `Deno.lookupDns` that exposes what is equivalent of `getaddinfo(3)`.

magurotuna:
Note to self: trust-dns 0.20.0 has been published with Tokio 1.0 support https://github.com/bluejekyll/trust-dns/releases/tag/v0.20.0

bnoordhuis:
> Will this issue be resolved by your suggestion https://github.com/denoland/deno/pull/8790#discussion_r550172074, right?

Correct.

runtime/ops/net.rs
magurotuna: Done in 20a42de

magurotuna:
I've done with the fixes that were pointed out by @bnoordhuis.

- check the permission against name servers that will be actually queried (67e8442)
- if a user specifies any specific name server, use that only as a target name server (b332917)
- don't receive protocol parameter(TCP or UDP) because `NameServerConfigGroup::from_ips_clear` of trust_dns_resolver always create both TCP and UDP connections according to [the doc](https://docs.rs/trust-dns-resolver/0.20.0/trust_dns_resolver/config/struct.NameServerConfigGroup.html#method.from_ips_clear). Referring to Node.js and Golang, they provide no way of specifying what protocol to use, thus I think it's fine.
- parse txt records as Latin-1 (20a42de)

Now, I bumped the version of trust-dns to 0.20.0 for some features, which means Tokio 1.0 is required for this PR to work. This needs to be merged _after_ #8779.
I'm going to add some integration tests once Tokio is upgraded.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -861,6 +861,92 @@ declare namespace Deno {
...
+    query: string,
+    recordType: "SRV",
+    options?: ResolveDnsOptions,
+  ): Promise<SRVRecord[]>;
+
+  export function resolveDns(
+    query: string,
+    recordType: "TXT",
+    options?: ResolveDnsOptions,
+  ): Promise<string[][]>;
```
ry, line 924: Rather than exposing interfaces for each DNS type, we could start more conservatively by just returning strings.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -861,6 +861,92 @@ declare namespace Deno {
...
+   * });
+   * ```
+   *
+   * Requires `allow-net` permission.
+   */
+  export function resolveDns(
+    query: string,
+    recordType: RecordType,
+    options?: ResolveDnsOptions,
+  ): Promise<string[] | MXRecord[] | SRVRecord[] | string[][]>;
```
ry, line 948: The return type of `Promise<string[]>` is sufficiently general for a first pass, I think.

runtime/ops/net.rs
```diff
@@ -536,3 +545,249 @@ fn op_listen(
...
+  },
+  PTR(String),
+  SRV {
+    priority: u16,
+    weight: u16,
+    port: u16,
+    target: String,
+  },
+  TXT(Vec<String>),
+}
```
ry, line 568: Remove. Just return strings.

ry (changes requested):
Needs tests.

magurotuna:
I added an integration test named `_079_resolve_dns `, and now `Deno.resolveDns` returns always an array of string (`string[]`) regardless of the record types. The problem I've found debatable is how to stringify respective record types. For now, the returned values look like:
https://github.com/denoland/deno/blob/4b8f8701025e2aabf47dec1265e41f8cad35024c/cli/tests/079_resolve_dns.ts.out#L2-L9

In particular, `MX` and `SRV` records are stringified unkindly; `0 mx.com.` means _"priority is 0, exchange is mx.com"_, and `0 100 1234 srv.com.` means _"priority is 0, weight is 100, port is 1234, name is srv.com"_.
Just because I didn't come up with any suitable form for them, the data of `MX` and `SRV` records are stringified like this.

In addition, as to the `TXT` records, ideally they are represented as `string[][]` (actually [Node.js does](https://nodejs.org/api/dns.html#dns_dns_resolvetxt_hostname_callback)) but in order to make it `string[]`, they get flattened by placing whitespace between each.

cli/dts/lib.deno.unstable.d.ts
bartlomieju: I discussed this PR offline with @lucacasonato and we both disagree with returning strings. The reason is two fold: a) DNS records are standardised and thus it's unlikely the types would change, b) returning strings would mean that users have to parse the record ourselves, which will lead to either a new module added to deno.land/x or a feature request for such module in standard library. @magurotuna's solution seems well implemented and this feature is behind `--unstable` flag so I think we should opt to use strict types.

ry:
@magurotuna Thank you for doing this work to return strings. Now that I see this, it seems less preferable to the original version with interface for each DNS record, because of the difficult to_string implementations in SRV and PTR . I apologize for the extra work and suggesting the wrong direction.

test_util/src/lib.rs
```diff
@@ -678,6 +701,130 @@ async fn wrap_abs_redirect_server() {
   }
 }
 
+async fn run_dns_server() {
```
ry, line 704: This is very awesome, but I think it should be done in cli/tests/integration_tests.rs

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -835,6 +835,54 @@ declare namespace Deno {
...
+    | "ANAME"
+    | "CNAME"
+    | "MX"
+    | "PTR"
+    | "SRV"
+    | "TXT";
+
+  export interface ResolveDnsOptions {
+    /** The name server to be used for lookups. 
+    * If not specified, defaults to the system configuration e.g. `/etc/resolve.conf` on Unix. */
```
lucacasonato, line 852:
```suggestion
    * If not specified, defaults to the system configuration e.g. `/etc/resolv.conf` on Linux. */
```

macOS doesn't actually use `/etc/resolv.conf`.

ry: resolv.conf is a user land configuration. Are you saying that trust dns doesn’t look at resolve.conf in OSX?

test_util/src/lib.rs
magurotuna: > but I think it should be done in cli/tests/integration_tests.rs since it's only used in a single test.

Makes sense, but when I tried to move `run_dns_server` to `integration_tests.rs`, I ran into a pretty weird error, which indicated that DNS resolution was timed out.
I'm struggling to figure out why this error happened. For now, it has turned out that [this snippet](https://gist.github.com/magurotuna/913500414948c40669332f23a5a2875a) does work (note that it's in the main function), however, if I put exactly same stuff into a _test_ (annotated with `#[tokio::test]`), it does NOT work somehow.

I think Deno's next release is coming nearly, and it looks like this error will take me some time to solve. So I'd like to suggest putting off this refacroting after the release - if this is fine, I'll fix the return type of `Deno.resolveDns` and adjust the test case to it as soon as possible.

cli/dts/lib.deno.unstable.d.ts
magurotuna: Oh, I wrote `resolve.conf` mistakenly. Thanks for pointing out :)

https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/system_conf/unix.rs#L29-L31
According to the implementation, it seems like `trust-dns` looks at `/etc/resolv.conf` in macOS too
as this `read_system_conf` function is under `#[cfg(unix)]`.
(implementation for windows is [here](https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/system_conf/windows.rs#L51))

test_util/src/lib.rs
magurotuna: Actually I've figured out why, so I'm doing this refactoring right now.

magurotuna:
I think it's ready to land. Please take a look :-)

lucacasonato:
Tested this locally, and works as expected. Great work @magurotuna. I'll leave the final review to @ry.

ry (approved):
LGTM - very nice work @magurotuna !

PR #8792, opened 2020-12-16
upload canary builds at once

maximousblk:
Some builds (mostly windows) would temporarily be missing due to longer build times


bartlomieju:
@maximousblk what's the use case for this PR?

lucacasonato:
@bartlomieju I asked for it. Currently there is a window of about 2-4 minutes after every push to master where `deno upgrade --canary` fails on Windows and macOS, because Linux builds are already uploaded, but the other platforms are not yet. This correct this so all images are uploaded at once.

PR #8793, opened 2020-12-16
feat(config): Prototype for a configuration file for Deno lint and fmt

SyrupThinker:
Currently there are lots of discussions about whether to add a configuration file to Deno.

To aid discoverability and allow for easier discussion of the topic,
this PR provides a rudimentary prototype of how a configuration could look like
and how it could be implemented.

Nothing in this initial state is definitive and thus open for discussion.

Topics of interest might be:
- What file format to use
- What configuration to allow
  - for the linter
  - for the formatter
  - for other tools
 - Should the file be "magically" found
 - What should the flag be called

The basis for this PR was [this](https://gist.github.com/SyrupThinker/45535bdc53e8bd5ff505a86a1d3e7e97) sketch-up of what a configuration could look like.

This initial implementation would be used like this:
https://asciinema.org/a/Szwxu2i3qaqEfXHvddcxIT9Wo

Tests should be implemented *after* discussions has taken place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -500,6 +500,7 @@ dependencies = [
  "tokio 0.2.22",
  "tokio-rustls",
  "tokio-tungstenite",
+ "toml",
  "uuid",
  "walkdir",
  "winapi 0.3.9",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -71,6 +71,7 @@ tempfile = "3.1.0"
 termcolor = "1.1.0"
 tokio = { version = "0.2.22", features = ["full"] }
 tokio-rustls = "0.14.1"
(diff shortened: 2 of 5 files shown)


MierenManz:
for file format it would probably be json because it is the object notation of javascript.

config may be a good one to also allow the run parameters like `--allow-all` etc.

RDambrosio016:
I suggest using schemars and deriving JsonSchema for all of the config structs, the lsp can then link the generated schema to tell vsc to use it for autocompletion.

bartlomieju:
Thank you @SyrupThinker but I'm gonna close without a merge. I've discussed configuration file recently with @kitsonk and we'll ship prototype soon.

Issue #3675, opened 2020-01-14
Permission meta-data manifest file


rsp:
Once in a while a proposal for permissions file for Deno is suggested.
Sometimes it's strictly for permissions like [here](https://github.com/denoland/deno/issues/2081#issuecomment-482797540) and sometimes it's more general like the comprehensive meta-data file in #3179.

This issue is to track info about similar systems if the idea of permission files come up in the future.

(For running complex Deno applications in the future it may be useful to be able to have single file that can be inspected and used instead of multiple command-line switches. Also, a similar format to express by the program the need for permissions to avoid issues described in #3655 would be useful and it might use a format similar to that for granting the permissions by the user.)

There are some existing systems that use JSON or YAML files with the same purpose, some use command-line switches. Maybe some of those solutions can serve as an inspiration for Deno (mainly how the permissions are defined, but maybe also how they work).

CloudABI - https://cloudabi.org/
A capability-based runtime environment for running untrusted native code without containers.
Supports C, C++, Python, Rust (Node.js coming soon - could it support Deno too?)
It uses YAML files to define permissions.
See:
https://github.com/NuxiNL/cloudabi#capability-based-security

Fuchsia - https://fuchsia.dev/
https://en.wikipedia.org/wiki/Google_Fuchsia
An operating system based on capabilities.
It uses JSON files to define permissions in component manifests
(CML - component manifest language - with a [JSON Schema](https://fuchsia.googlesource.com/fuchsia/+/master/garnet/lib/rust/cm_json/cml_schema.json))
See:
https://fuchsia.dev/fuchsia-src/concepts/components/capabilities
https://fuchsia.dev/fuchsia-src/concepts/components/component_manifests

WASI - https://wasi.dev/
The WebAssembly System Interface
Since Deno can run WebAssembly, it might me useful to look into it.
See:
https://github.com/bytecodealliance/wasmtime/blob/master/docs/WASI-capabilities.md
It doesn't define how to describe permissions, implementations like Wasmtime do, see below.

Wasmtime - https://wasmtime.dev/
A runtime for WASM with WASI support.
It uses command line switches like `--dir=.` and `--mapdir=/tmp::/var/tmp` to give permissions to the executed program.
See:
https://github.com/bytecodealliance/wasmtime/blob/master/docs/WASI-tutorial.md

Capsicum - https://www.cl.cam.ac.uk/research/security/capsicum/
[...]

Issue #3179, opened 2019-10-21
proposal: support a comprehensive meta-data file


kitsonk:
There is discussion around "lockfiles" and sub-resource integrity (see: #200).

There are also other "external" data that might be needed to describe a workload in Deno, like what the main module is, a custom TypeScript configuration, permissions and other command line options, import-maps, etc.  While Deno should always _just work_ without a meta data file, there are lots of use cases where external meta data is necessary or desired.

Providing a single file way of describing that meta data would be a desirable feature.

I propose the following:

* Deno supports a single file meta-data schema that fully describes a workload.
* That the format of this file would be `.json`
* That when a command resolves an argument to a `.json` file, it assumes that it adheres to the meta-data schema, meaning that `deno run app.json` would read the meta data file and utilise that to determine how to cache and execute the app.  This would also mean something like `deno run https://example.com/app.json` would also _just work_.

As this proposal evolves, the definition of the schema would be detailed, but would include all the functional areas mentioned above.  Specifically for the "lockfile" mechanism, if an application is loaded and the `.json` file is writable (local) and there is no cache of the hashes of the modules, then Deno will write out the hashes of all dependent modules to the file.  Upon subsequent invocations where the cache of the hashes is present, Deno will validate that all modules loaded match that cache, and if they don't they will throw and the workload would terminate.

Support a command like `deno write main.ts app.json` would scaffold out the meta data file, but not run the workload.  Any arguments passed on the CLI could be automatically written to the file as well.

nayeemrmn:
So one of the ideas here is to consume the meta-file by using it as the entrypoint of the app rather than implicitly finding and reading it based on obscure logic. Nice.

7040934:
อีเมลข้อมูลและข้อมูลส่วนตัวหายไปไหนใครมาสวมแทน

7040934:
อีเมลข้อมูลและข้อมูลส่วนตัวหายไปไหนใครมาสวมแทน

7040934:
> มีการอภิปรายเกี่ยวกับ "lockfiles" และความสมบูรณ์ของทรัพยากรย่อย (ดู: [# 200](https://github.com/denoland/deno/issues/200) )
> 
[...]

PR #8794, opened 2020-12-16
feat(cli): meta file

crowlKats:
This adds a meta file in which all flags that can be passed to cli can be passed, and also for linter settings and such. any flags passed in the cli will overwrite the correspondng setting in the meta file.

Example:
https://cdn.discordapp.com/attachments/700150514217779230/786048879099379712/Screen_Recording_2020-12-09_at_02.55.33.mov


nayeemrmn:
Refs #3179, #3675.

We should add flag-based configuration to `deno fmt`/`deno lint` before considering using any file for it, IMO.

crowlKats:
i will not work on thihs for the time being. because of https://github.com/denoland/deno/pull/8617, the code for it can then be used for meta file by just adding deserialize macro. so until that aint merged, it would be wasted time

PR #8798, opened 2020-12-17
fix(cli/web): fix atob to throw DOMException

### LINKED ISSUE #8795: atob should throw DOMException
`atob('foobar!!')` returns a `DOMException` in chrome, but it returns a `DataDecodeError` in Deno. This should be aligned to match Chrome behaviour.

gorogoroumaru:
fix #8795 

atob('foobar!!') returns a DOMException in chrome, but it returns a DataDecodeError in Deno.
I fixed atob function to throw DOMException instead of DataDecodeError.

### DIFF
--- a/cli/tests/unit/text_encoding_test.ts
+++ b/cli/tests/unit/text_encoding_test.ts
@@ -50,6 +50,16 @@ unitTest(function atobThrows2(): void {
   assert(threw);
 });
 
+unitTest(function atobThrows3(): void {
+  let threw = true;
+  try {
+    atob("foobar!!");
+  } catch (DOMException) {
+    threw = true;
+  }
+  assert(threw);
+})
+
 unitTest(function btoaFailed(): void {
   const text = "你好";
   assertThrows(() => {
--- a/cli/tests/unit/text_encoding_test.ts
+++ b/cli/tests/unit/text_encoding_test.ts
@@ -51,14 +51,14 @@ unitTest(function atobThrows2(): void {
 });
 
 unitTest(function atobThrows3(): void {
-  let threw = true;
+  let threw = false;
   try {
     atob("foobar!!");
   } catch (DOMException) {
     threw = true;
   }
   assert(threw);
-})
+});
 
 unitTest(function btoaFailed(): void {
   const text = "你好";
--- a/cli/tests/unit/text_encoding_test.ts
+++ b/cli/tests/unit/text_encoding_test.ts
@@ -54,8 +54,10 @@ unitTest(function atobThrows3(): void {
   let threw = false;
   try {
     atob("foobar!!");
-  } catch (DOMException) {
-    threw = true;
+  } catch (e) {
+    if (e instanceof DOMException && e.toString().startsWith("DOMException:")) {
+      threw = true;
+    }
   }
   assert(threw);
 });
--- a/cli/tests/unit/text_encoding_test.ts
+++ b/cli/tests/unit/text_encoding_test.ts
@@ -55,7 +55,10 @@ unitTest(function atobThrows3(): void {
   try {
     atob("foobar!!");
   } catch (e) {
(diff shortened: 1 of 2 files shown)


cli/tests/unit/text_encoding_test.ts
```diff
@@ -50,6 +50,16 @@ unitTest(function atobThrows2(): void {
   assert(threw);
 });
 
+unitTest(function atobThrows3(): void {
+  let threw = false;
+  try {
+    atob("foobar!!");
+  } catch (DOMException) {
```
kitsonk, line 57: This unit test needs to:

- Check that the caught error is an instance of DOMException
- Validate that the `toString()` of the caught error starts with the text `DOMException:`

kitsonk (changes requested):

cli/tests/unit/text_encoding_test.ts
gorogoroumaru: When I execute the following on chrome, I get this error.
```
try {
    atob("foobar!!");
} catch (e) {
    console.log(e.toString())
}
```
InvalidCharacterError: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.

What should I do?

jaeheonji: Hi! @gorogoroumaru, @kitsonk 

What about the following approach?

```javascript
let threw = false;
try {
  ...
} catch (e) {
  if (e instanceof DOMException) {
    threw = true;
  }
}
assert(threw);
```

jaeheonji (changes requested):

cli/tests/unit/text_encoding_test.ts
```diff
@@ -50,6 +50,16 @@ unitTest(function atobThrows2(): void {
   assert(threw);
 });
 
+unitTest(function atobThrows3(): void {
+  let threw = true;
+  try {
+    atob("foobar!!");
+  } catch (DOMException) {
```
jaeheonji, line 57: Hi, @gorogoroumaru, @kitsonk

What about following approach?

```javascript
let threw = true;
try {
  ...
} catch (e) {
  if (e instanceof DOMException) {
    threw = false;
  }
}
assert(threw);
```

jaeheonji (changes requested):

cli/tests/unit/text_encoding_test.ts
gorogoroumaru: Thank you @jaeheonji!
Looks good!
@kitsonk, how about you?

rchougule:
Hi, was just going via open PRs to learn more about deno and thought of asking this...

Shouldn't we be throwing `InvalidCharacterError` according to the spec here: 
image

https://html.spec.whatwg.org/#dom-atob

Or is it like that should be thrown only if the `decoded` data is invalid? Considering the current changes are regarding the `encoded` data being incorrect...

op_crates/web/08_text_encoding.js
```diff
@@ -105,7 +105,7 @@
     if (rem === 1 || /[^+/0-9A-Za-z]/.test(s)) {
       throw new DOMException(
         "The string to be decoded is not correctly encoded",
-        "DataDecodeError",
+        "DOMException",
```
jaeheonji, line 108: The `atob()` should display a `DOMException` error when the parameter is not base64 encoded data.
In the current testcase, since `foobar!!` is not base64 encoded data, it is correct for an exception to occur.

But, In [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) docs, The second property of `DOMException` is the name of the error. 

So, @gorogoroumaru, As mentioned by @rchougule, It seems more reasonable for us to rename the error to `InvalidCharacterError`.

gorogoroumaru:
Thank you @rchougule and @jaeheonji!
I applied your suggested changes.

kitsonk (approved):
LGTM

Issue #4752, opened 2020-04-14
Refactor runtime compiler APIs


kitsonk:
We recently changed `deno fetch` to `deno cache`.  When we introduced the runtime compiler APIs, one of the challenges is that it didn't align well with the subcommands.  Given that we have a better name now for the subcommands, it feels right to re-align the runtime compiler APIs, also based on some experience with them as well.  I am recommending the following:

```ts
namespace Deno {
  export interface CacheOptions { /* would be close to existing compiler options */ }
  export interface BundleOptions ( /* would vary a little bit from cache options */ }
  export interface TranspileOptions extends CacheOptions { /* a few options differ */ }

  export function cache(root: string, options?: RuntimeCompilerOptions): Promise<Record<string, string>>;
  export function bundle(root: string, options?: RuntimeCompilerOptions): Promise<string>;
  export function transpile(sources: Record<string, string>, options?: TranspileOptions): Promise<Record<string, string>>;
}
```

`Deno.cache` would work like `deno cache` and the `RuntimeCompilerOptions` would also support the flags that `deno cache` supports plus the support of an expanded `tsconfig.json` set of compiler options aligned to what exists today in `CompilerOptions`.

`Deno.bundle` would produce a single file emit, like `deno bundle` does and people would use `Deno.transpile()` when they have sources that they want the compiler to transpile for them.

We could introduce the equivalent `--no-check` in these APIs as well, as well as adding it to the appropriate subcommands, but that might be an enhancement on top of this refactor.

Refs: #4600

jakajancar:
Looks good.

Right now `--cached-only` on the CLI affects `Deno.compile()`, which is weird. Would this still be the case? I would expect the CLI settings/tsconfig.json would affect how deno runs my app, not the logic of my code itself (action at a distance-ish). E.g. I might want `--cached-only` to ensure my app has all files downloaded, buy maybe it's a watching/reloading compiler for bundling web stuff, so obviously it will not use the cache only when building it's output.

Soremwar:
Should this be done before v1.0? If not, then `compile` should probably be hidden under the `--unstable` flag in order to be changed in the future.

kitsonk:
Isn't it under `--unstable`?

Soremwar:
It is :sweat_smile:

Sorry for my blindness

Nautigsam:
[...]

Issue #8706, opened 2020-12-10
Utilise the emit from swc instead of tsc


kitsonk:
In Deno 1.6 we added isolated modules to the _ignored options_, which means that Deno is forcing the built-in TypeScript type checker (`tsc`) to not use type dependent emits, which means the same TypeScript code that compiles under normal operations also supported under `--no-check`.  This "unlocks" us using swc to do all the emitting and stop using tsc to do the emit.

My estimates are that it will sped up "checking" of TypeScript by 20-30%, which is significant.

kitsonk:
@ry on my machine with tsc emit:

```
Benchmark #1: deno cache --reload test_util/std/examples/chat/server_test.ts
  Time (mean ± σ):     954.6 ms ±  63.6 ms    [User: 1.510 s, System: 0.061 s]
  Range (min … max):   898.6 ms … 1069.2 ms    10 runs
```

And without tsc emit:

```
Benchmark #1: deno cache --reload test_util/std/examples/chat/server_test.ts
  Time (mean ± σ):     868.3 ms ±   8.1 ms    [User: 1.382 s, System: 0.047 s]
  Range (min … max):   857.0 ms … 881.3 ms    10 runs
```

So more like a 10% reduction instead of a 20-30% reduction like I thought.

bartlomieju:
@kitsonk are there any blockers for this at the moment? If not, I would like to take a stab at it.

kitsonk:
@dsherret has a better idea of the current state of swc. It should be a trivial change.

PR #8799, opened 2020-12-17
refactor(cli): runtime compiler APIs consolidated to Deno.emit()

kitsonk:
Closes: #4752

This is a **breaking change** that replaces the unstable `Deno.compile()`, `Deno.bundle()`, and `Deno.transpileOnly()` APIs with a single API, `Deno.emit()` which behaves more in line with the way `deno run`, `deno cache`, and `deno bundle` work, where there is a `check` option which determines if the program is checked or not.  In addition, the `Deno.emit()` API supports import maps, which the previous APIs did not.  (It should be noted, that Deno does not transform any paths with an import map, so any consumers of the emitted code would need to be able to leverage the same resolutions as the import map).


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -487,122 +494,78 @@ declare namespace Deno {
...
+  interface EmitResult {
+    /** Diagnostic messages returned from the type checker (`tsc`). */
+    diagnostics: Diagnostic[];
+    /** Any emitted files.  If bundled, then the JavaScript will have the
+     * key of `deno:///bundle.js` with an optional map (based on
+     * `compilerOptions`) in `deno:///bundle.js.map`. */
+    files: Record<string, string>;
+    /** An optional array of any compiler options that were ignored by Deno. */
+    ignoredOptions?: string[];
+    /** An array of internal statics related to the emit, for diagnostic
```
bartlomieju, line 542: typo: s/statics/statistics/

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -487,122 +494,78 @@ declare namespace Deno {
...
-   * @param options An optional object of options to send to the compiler. This is
-   *                a subset of ts.CompilerOptions which can be supported by Deno.
+  /**
+   * **UNSTABLE**: new API, yet to be vetted.
+   * 
+   * Similar to the command line functionality of `deno run` or `deno cache`,
+   * `Deno.emit()` provides a way to provide Deno arbitrary JavaScript
+   * or TypeScript and have it return JavaScript based on the options and
+   * settings provided. The source code can either be provided or the modules
+   * can be fetched and resolved like other modules are resolved.
```
bartlomieju, line 554: s/`like other modules are resolved`/`line in CLI`?

cli/module_graph.rs
```diff
@@ -624,6 +633,10 @@ impl Default for BundleType {
 
 #[derive(Debug, Default)]
 pub struct EmitOptions {
+  /// If true, then code will be type checked, otherwise type checking will be
+  /// skipped.  If false, then swc will be used for the emit, otherwise tsc will
+  /// be used.
```
bartlomieju, line 638: Is it final that `tsc` will be used for emits if `check` is true or does https://github.com/denoland/deno/issues/8706 correspond to this bit? I think the former because we don't enforce `isolatedModules` in the runtime compiler API.

cli/ops/runtime_compiler.rs
```diff
@@ -29,131 +26,91 @@ use std::collections::HashMap;
...
-  let specifier = ModuleSpecifier::resolve_url_or_path(&args.root_name)
-    .context("The root specifier is invalid.")?;
-  builder.add(&specifier, false).await?;
-  let graph = builder.get_graph();
-  let bundle_type = if args.bundle {
-    BundleType::Esm
+  let maybe_import_map = if let Some(import_map_str) = args.import_map_path {
+    let import_map_specifier =
+      ModuleSpecifier::resolve_url_or_path(&import_map_str).map_err(|_| {
+        anyhow!("Bad file path (\"{}\") for import map.", import_map_str)
```
bartlomieju, line 79: Maybe use `.context()` to enhance returned error instead of throwing it away?

bartlomieju (approved):
LGTM, I have only a few minor comments. 

I'm strongly in favor of this change, but wouldn't mind a bit more of bike-shedding on the API name.

kitsonk:
> but wouldn't mind a bit more of bike-shedding on the API name

Like what?  It isn't bike-shedding if there isn't a suggestion.

cli/module_graph.rs
kitsonk: It is the former at the moment (because #8706 is not resolved).  If we resolve #8706, I think we should consider ensuring that `isolatedModules: false` is not supported with this API.  We might actually want to consider it for this PR actually.  It will just cause people more problems than it fixes in supporting it.  We already have a slew of options we don't support for the runtime APIs.

kitsonk: I've removed `isolatedModules` as a configurable option for the runtime API.  If someone needs `const enum` they will need to use `tsc` directly.

bartlomieju: Okay awesome! Thanks for clarifying.

bartlomieju:
> > but wouldn't mind a bit more of bike-shedding on the API name
> 
> Like what? It isn't bike-shedding if there isn't a suggestion.

Alright, I retract my words, I can't come up with a better name for this API. LGTM

PR #8802, opened 2020-12-17
test(cli): ensure await all on stdout does not deadlock

### LINKED ISSUE #6840: await Deno.stdout.write doesn't resolve when called in rapid succession
Came across this when using the https://github.com/jakobhellermann/deno-progressbar library to provide feedback on a series of tasks.

**example script that has the issue**

```ts
const exampleSource = ["one", "two"];
const encoder = new TextEncoder();

const exampleProcesses = exampleSource.map(async (value) => {
  const text = `${value}`;
  await Deno.stdout.write(encoder.encode(text));
});

await Promise.all([
  ...exampleProcesses,
  // not needed to reproduce, this is just to show the promise unresolved status in output
  new Promise((_, reject) => {
    setTimeout(() => {
      console.log("\n---\nfailures");
      exampleProcesses.forEach(console.log);
      reject();
    }, 1000);
  }),
]);
console.log("never reached");
```
**output of example script**

```sh
deno run test.ts
Check file:///Users/lukewoollard/work/deno-figma/test.ts
onetwo
---
failures
Promise { <pending> } 0 [ Promise { <pending> }, Promise { undefined } ]
Promise { undefined } 1 [ Promise { <pending> }, Promise { undefined } ]
error: Uncaught undefined
```

Worked around this locally by vendoring that code and replacing the use of `await Deno.stdout.write` with `Deno.stdout.writeSync`.

**Related issue**: https://github.com/denoland/deno/issues/5515

caspervonb:
This adds an "await all writes" case that would previously dead-lock that has been solved with the new resource table as a regression test (https://github.com/denoland/deno/issues/6840#issuecomment-662569644).

Closes #6840

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -2105,6 +2105,11 @@ fn deno_test_no_color() {
   assert!(out.contains("test result: FAILED. 1 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out"));
 }
 
+itest!(stdout_write_all {
+  args: "run --quiet stdout_write_all.ts",
+  output: "stdout_write_all.out",
+});
+
 itest!(_001_hello {
   args: "run --reload 001_hello.js",
   output: "001_hello.js.out",
--- a/cli/tests/stdout_write_all.out
+++ b/cli/tests/stdout_write_all.out
@@ -1,0 +1,1 @@
+Hello, world!
--- a/cli/tests/stdout_write_all.out
+++ b/cli/tests/stdout_write_all.out
@@ -1,1 +1,3 @@
-Hello, world!
+done
+done
+complete
--- a/cli/tests/stdout_write_all.ts
+++ b/cli/tests/stdout_write_all.ts
@@ -1,0 +1,8 @@
+const encoder = new TextEncoder();
+const pending = [
+  Deno.stdout.write(encoder.encode("Hello, ")),
+  Deno.stdout.write(encoder.encode("world!")),
+];
+
+await Promise.all(pending);
+await Deno.stdout.write(encoder.encode("\n"));
--- a/cli/tests/stdout_write_all.ts
+++ b/cli/tests/stdout_write_all.ts
@@ -1,8 +1,8 @@
 const encoder = new TextEncoder();
 const pending = [
-  Deno.stdout.write(encoder.encode("Hello, ")),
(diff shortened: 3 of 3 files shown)


caspervonb:
Still reading through the new resource table and ops implementation, can this race @bartlomieju?

bartlomieju:
> Still reading through the new resource table and ops implementation, can this race @bartlomieju?

I think so, @piscisaureus can you double check?

bartlomieju:
Actually thinking about it more, when you create promises both of them are put in `JsRuntime::pending_ops` which are then polled by event loop. I don't think we can guarantee that they will be polled in the same order as called (because `JsRuntime::pending_ops` uses `FuturesUnordered` to store those futures) and thus when obtaining `Rc` to resource they might be out of order.

caspervonb:
> Actually thinking about it more, when you create promises both of them are put in JsRuntime::pending_ops which are then polled by event loop. I don't think we can guarantee that they will be polled in the same order as called (because JsRuntime::pending_ops uses FuturesUnordered to store those futures) and thus when obtaining Rc to resource they might be out of order.

Ah, okay! thanks for the insight 😄
Rewrote the test case to be race tolerant.

bartlomieju (approved):
LGTM

piscisaureus:
> Ah, okay! thanks for the insight 😄
Rewrote the test case to be race tolerant.

Actually if you use borrow_mut (not borrow) from an AsyncRefCell, futures are guaranteed to get access to the cell's contents in the order they called borrow_mut().

caspervonb:
> Actually if you use borrow_mut (not borrow) from an AsyncRefCell, futures are guaranteed to get access to the cell's contents in the order they called borrow_mut().

Also good to know, suppose the open question is if we want to make any kind of guarantees to the user about it as it's implementation dependent.

piscisaureus:
This design was deliberate on my end. Imagine this:

```
let p1 = Deno.write(file, new Uint8Array([1, 2, 3]));
let p2 = Deno.write(file, new Uint8Array([4, 5, 6]));
await Promise.all([p1, p2]);
```

In my opinion the written file should always contain `[1, 2, 3, 4, 5, 6]` and never `[4, 5, 6, 1, 2, 3]`.
The new ResourceTable / AsyncRefCell implementation takes care of this now.

caspervonb:
Alright, I'll revisit the test coverage for that 👍

Issue #5392, opened 2020-05-14
std/node: add tty module


EvanHahn:
For completeness, we should add a polyfill for [Node's tty module](https://nodejs.org/api/tty.html).

PR #8805, opened 2020-12-17
feat(std/node): support nodejs tty module

jopemachine:
## Outline

* Migrating nodejs tty module

## Issue

#5392

## Notes

* WIP
* Need to be tested
* Any advice would be appreciated.

### DIFF
--- a/std/node/tty.ts
+++ b/std/node/tty.ts
@@ -1,10 +1,10 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
-import { ERR_INVALID_ARG_TYPE, ERR_OUT_OF_RANGE } from './_errors.ts';
-import nodeProcess, { Env } from './process.ts';
+import { ERR_INVALID_ARG_TYPE, ERR_OUT_OF_RANGE } from "./_errors.ts";
+import nodeProcess, { Env } from "./process.ts";
 import { notImplemented } from "./_utils.ts";
-import { ReadableOptions } from './_stream/readable.ts';
-import nodeOs from './os.ts';
+import { ReadableOptions } from "./_stream/readable.ts";
+import nodeOs from "./os.ts";
 
 let OSRelease: string[];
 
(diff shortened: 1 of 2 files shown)


std/node/tty.ts
```diff
@@ -0,0 +1,295 @@
...
+      throw new ERR_OUT_OF_RANGE("count", ">= 2", count);
+    }
+  }
+  return count <= 2 ** getColorDepth(env);
+}
+
+export function isatty(fd: number) {
+  // TODO(jopemachine): to be implemented
+  notImplemented();
+}
```
guybedford, line 250: It would be really nice to just support stdin / stdout / stderr here as that would cover the majority of use cases for nice terminal output (basically supports-color doing `isatty(1)`

```suggestion
export function isatty(fd: number) {
  switch (fd) {
    case 1: return isStdinTTY();
    case 2: return isStdoutTTY();
    case 3: return isStderrTTY();
  }
  notImplemented();
}
```

where on the Rust side it should just be like https://github.com/denoland/deno/blob/master/runtime/permissions.rs#L605.

guybedford (commented):
Nice to see a start on this!

std/node/tty.ts
jopemachine: Thanks you for your reviewing.

May I ask my understanding is correct?

```
function isStdinTTY() {
  return nodeProcess.stdin.isTTY;
}

function isStdoutTTY() {
  return nodeProcess.stdout.isTTY;
}

function isStderrTTY() {
  return nodeProcess.stderr.isTTY;
}

export function isatty(fd: number) {
  switch (fd) {
    case 1: return isStdinTTY();
    case 2: return isStdoutTTY();
    case 3: return isStderrTTY();
  }

  return Deno.isatty(fd);
}
```

guybedford: Just been doing some further digging and this might be as an example to follow - https://github.com/denoland/deno/blob/master/runtime/js/40_tty.js#L11.

jopemachine: > Just been doing some further digging and this might be as an example to follow - https://github.com/denoland/deno/blob/master/runtime/js/40_tty.js#L11.

You mean to implement new isatty function using the new rust function, not using Deno.isatty (it seems that maybe it could use 40_tty.js)?

Excusing for my poor understanding, May I ask you why we should do that?

guybedford: Ah, I didn't know there was a `Deno.isatty`! Ok I've corrected the suggestion to a proposed exact implementation then.

jopemachine:
I think to migrate `tty` module, `net` and `readline` module needs to be migrated.

So, I will stop this work until these modules could be implemented.

bartlomieju:
@jopemachine we won't land this PR before standard library migration to a separate repository (https://github.com/denoland/deno/issues/9029#issuecomment-766934654). 

I'm still very much interested in discussing these changes on upcoming meetings and potentially landing the PR.

bartlomieju:

Issue #8735, opened 2020-12-12
Workers & top-level-await changed behavior between 1.4.6 and 1.6.0


soootaleb:
Hello,

I've been implementing a distributed KV Store with Deno recently in version 1.4.6.

I use a worker for the network layer (so that the webserver is not blocking the main program):
- I use the `serve` function and a top level await to received requests (ugraded to websockets with `acceptWebsocket`)
- On a new request, the worker forwards the payload to the main thread using `self.postMessage`
- The main thread can `worker.postMessage` in order to send messages on the network (in the WebSocket)

This works perfectly in 1.4.6; I declare the `self.onmessage` function before starting the top level await, and the worker runs it when I `worker.postMessage`, even if the top level await is running as a web server.

But I tried to bump to Deno 1.6.0, and **while the worker is still able to forward incoming messages to the main thread, it seems the self.onmessage handler is no longer called when the main thread posts a message to the worker**.

I identified the top level await to be responsible since commenting it allows `self.onmessage` to be called correctly.

I've tried to explore the latest changes and issues, but I didn't find anything mentioning a change in behavior regarding the blocking / threading behaviors or workers / top level awaits.

Can anyone enlighten me ?

Regards

NB: The worker code is here https://github.com/soootaleb/abcd/blob/master/src/net.worker.ts

bartlomieju:
> I've tried to explore the latest changes and issues, but I didn't find anything mentioning a change in behavior regarding the blocking / threading behaviors or workers / top level awaits.

> Can anyone enlighten me ?

@soootaleb in `1.5` we had a fix for top-level-await implementation that prevent some nasty edge cases. Unfortunately it looks like it had impact on workers implementation as well... 😬 

Unfortunately linked code can't really be used as a test case, could you try to boil down reproduction?

soootaleb:
@bartlomieju thanks a lot for your reactivity :)

Here we go: I created a subfolder with example code on a dedicated branch, see the README, reproducing is easy

https://github.com/soootaleb/abcd/tree/deno-issue-8735/issue-8735

soootaleb:
@bartlomieju Hey !

My earlier, deno `1.4.0`, working version, used `std@0.65.0`, so I just verified the problem didn't come from the http module (`serve` function) latest `0.80.0` version.

**I can confirm it doesn't come from the http module since Deno `1.6.0` with `std@0.65.0` doesn't work either **.

[...]

PR #8809, opened 2020-12-17
fix: TLA in web worker

bartlomieju:
Fixes https://github.com/denoland/deno/issues/8735

Implementors of `deno_core::JsRuntime` might want to do additional actions
during each turn of event loop, eg. `deno_runtime::Worker` polls inspector,
`deno_runtime::WebWorker` receives/dispatches messages from/to worker host.
Previously `JsRuntime::mod_evaluate` was implemented in such fashion that it 
only polled `JsRuntime`'s event loop. This behavior turned out to be wrong 
in the example of `WebWorker` which couldn't receive/dispatch messages because
its implementation of event loop was never called. 

This commit rewrites "mod_evaluate" to return a handle to receiver that resolves
when module's promise resolves. It is now implementors responsibility to poll
event loop after calling `mod_evaluate`.


runtime/web_worker.rs
```diff
@@ -315,7 +315,28 @@ impl WebWorker {
...
+        // If `None` is returned it means that runtime was destroyed before
+        // evaluation was complete. This can happen in Web Worker when `self.close()`
+        // is called at top level.
+        let result = maybe_result.unwrap_or(Ok(()));
+        return result;
+      }
+
+      event_loop_result = self.run_event_loop() => {
+        if self.has_been_terminated() {
+          return Ok(());
```
ry, line 332: Shouldn't has_been_terminated result in error? is this conditional case tested?

runtime/worker.rs
```diff
@@ -211,7 +212,21 @@ impl MainWorker {
...
+        debug!("received module evaluate {:#?}", maybe_result);
+        let result = maybe_result.expect("Module evaluation result not provided.");
+        return result;
+      }
+
+      event_loop_result = self.run_event_loop() => {
+        event_loop_result?;
+        let maybe_result = receiver.next().await;
+        let result = maybe_result.expect("Module evaluation result not provided.");
+        return result;
```
ry, line 227: Why does this logic need to be duplicated in JsRuntime::execute_module and in Worker:: execute_module ? Is this handling two different situations?

ry: Why doesn't this one handle the has_been_terminated situation?

core/runtime.rs
```diff
@@ -1110,7 +1097,8 @@ impl JsRuntime {
           v8::PromiseState::Fulfilled => {
             state.pending_mod_evaluate.take();
             scope.perform_microtask_checkpoint();
-            sender.try_send(Ok(())).unwrap();
+            // Receiver end might have been already dropped, ignore the result
+            let _ = sender.try_send(Ok(()));
```
ry, line 1101: Can the unwrap panic being fixed here be demonstrated with a test in deno_core?

runtime/web_worker.rs
bartlomieju: I don't think so, calling `has_been_terminated` checks a boolean flag on `WebWorker`. It's tested in most of the tests in `workers_test.ts`

runtime/worker.rs
bartlomieju: > Why does this logic need to be duplicated in Worker::execute_module and in WebWorker:: execute_module ? Is this handling two different situations?

Yes, as per PR description implementations of event loop are different in `Worker` and `WebWorker`.

bartlomieju: > Why doesn't this one handle the has_been_terminated situation?

`Worker` has no termination logic, only `WebWorker` does.

core/runtime.rs
bartlomieju: It is - tests in `deno_core` don't really await on the receiver from `mod_evaluate`

ry: I'm suggesting to add a test that would panic on master branch, but is fixed here by removing the unwrap.

bartlomieju: @ry that's not really possible because the need to remove unwrap arises from this PR - previously `mod_evaluate` returned a future that held a reference to the receiver until it was resolved, in this PR receiver handle is returned to the caller which might drop it in which case `unwrap()` would panic. This case is covered in test in `core/runtime.rs` and `core/modules.rs` which all drop receiver handle and there's no panic.

ry: Ok fair

ry (approved):
LGTM

PR #8811, opened 2020-12-17
feat(doc): Add the documentation for addEventListener Method

### LINKED ISSUE #8804: Add JS Doc for addEventListener
Ideally with example code. This is the first method to show up on the docs. It would be great if it looked more fleshed out.

Screen Shot 2020-12-17 at 9 30 57 AM

championshuttler:
Fixes #8804

### DIFF
--- a/op_crates/web/lib.deno_web.d.ts
+++ b/op_crates/web/lib.deno_web.d.ts
@@ -107,7 +107,16 @@ declare class EventTarget {
      * invoked once after which the event listener will be removed.
      *
      * The event listener is appended to target's event listener list and is not
-     * appended if it has the same type, callback, and capture. */
+     * appended if it has the same type, callback, and capture.
+     *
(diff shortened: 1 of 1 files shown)


op_crates/web/lib.deno_web.d.ts
```diff
@@ -108,6 +108,13 @@ declare class EventTarget {
      *
      * The event listener is appended to target's event listener list and is not
      * appended if it has the same type, callback, and capture. */
+  /** addEventListener() method attaches an event handler to the document.
```
kitsonk, line 111: This should be a continuation of the previous JSDoc block, just adding the example, with a clean line break before it.

kitsonk (changes requested):

ry:
Thank you but the issue is actually denoland/deno_doc#31

PR #8813, opened 2020-12-17
refactor(runtime): remove warp dependency

### LINKED ISSUE #8782: Remove dependency on "warp" in "deno_runtime" crate
In https://github.com/denoland/deno/pull/8781 it became apparent that `warp` crate is used only in `runtime/inspector.rs`. 

Following TODO can be found in `runtime/inspector.rs`:
https://github.com/denoland/deno/blob/6356345365e766d984f591506fb475d7935685de/runtime/inspector.rs#L151-L154

Additionally it's becoming a bit challenging to keep `warp` and other crates that depend on `hyper` in sync to use only a single version of `hyper`. 

Let's remove `warp` as a dependency and rewrite `runtime/inspector.rs` in such a fashion that create a simple WS server using `hyper` ourselves.

CC @piscisaureus @ry

lperlaki:
Fixes #8782

Replaces the `warp` web server in the `deno_runtime` crate with a `hyper` server and a `tokio-tungstenite` websocket implementation.


runtime/inspector.rs
```diff
@@ -142,90 +146,163 @@ impl InspectorInfo {
...
-    warp::reply::json(&json!(json_values))
+
+  let make_svc = hyper::service::make_service_fn(|_| {
+    let inspector_map = Rc::clone(&inspector_map_);
+    let json_version_response = json_version_response.clone();
+    future::ok::<_, Infallible>(hyper::service::service_fn(
+      move |req: http::Request<hyper::Body>| {
+        future::ready({
+          match (req.method(), req.uri().path()) {
+            (&http::Method::GET, path) if path.starts_with("/ws/") => {
```
bartlomieju, line 218: Is it necessary to use low level `hyper` APIs for this purpose? https://github.com/snapview/tokio-tungstenite/blob/master/examples/server.rs gives a nice example of create WS only server. We could do request path handling by using methods on the [`WebSocketStream`](https://docs.rs/tokio-tungstenite/0.12.0/tokio_tungstenite/struct.WebSocketStream.html)

lperlaki: We do need hyper for the other http routes.
I do not see which methods on `WebSocketStream` could help with path handling, could you explain that?

bartlomieju: @lperlaki you're right, I forgot that we need to handle a few regular HTTP endpoints as well. It seems your PR is done, I will review it today

ry (commented):
@lperlaki Nice work - thank you! 

Summoning the Dutch masters for review...

runtime/inspector.rs
lperlaki: Ok, thank you

runtime/inspector.rs
```diff
@@ -142,95 +146,155 @@ impl InspectorInfo {
...
+              if let Some(new_websocket_tx) = path
+                .strip_prefix("/ws/")
+                .and_then(|s| Uuid::parse_str(s).ok())
+                .and_then(|uuid| {
+                  inspector_map
+                    .borrow()
+                    .get(&uuid)
+                    .map(|info| info.new_websocket_tx.clone())
```
bartlomieju, line 217: It's a bit hard to grok this block, could service fn be split into separate helper function per route?

lperlaki: Sure I will do that

runtime/inspector.rs
```diff
@@ -59,12 +58,12 @@ impl InspectorServer {
...
-        register_inspector_rx,
-        shutdown_server_rx,
-        name,
-      ))
+      let mut rt = crate::tokio_util::create_basic_runtime();
+      let local = tokio::task::LocalSet::new();
+      local.block_on(
+        &mut rt,
+        server(host, register_inspector_rx, shutdown_server_rx, name),
+      )
```
lperlaki, line 66: Should there be a `run_local` function in  `tokio_util` that sets up a Runtime with a `LocalSet`?

runtime/inspector.rs
lperlaki: @bartlomieju I now moved the individual Request handling logic into different functions.

bartlomieju: @lperlaki thank you, it's much better.

runtime/inspector.rs
bartlomieju: It's not required anywhere besides inspector, so I suggest leaving it as is

runtime/inspector.rs
```diff
@@ -142,95 +141,172 @@ impl InspectorInfo {
...
+        .map(|info| info.new_websocket_tx.clone())
+    })
+  {
+    let resp = tungstenite::handshake::server::create_response(&req)
+      .map(|resp| resp.map(|_| hyper::Body::empty()))
+      .or_else(|e| match e {
+        tungstenite::error::Error::HttpFormat(http_error) => Err(http_error),
+        _ => http::Response::builder()
+          .status(http::StatusCode::BAD_REQUEST)
+          .body("Not a valid Webscoket Request".into()),
```
bnoordhuis, line 183:
```suggestion
          .body("Not a valid Websocket Request".into()),
```
Can I suggest logging the error for troubleshooting purposes?

runtime/inspector.rs
```diff
@@ -142,95 +141,172 @@ impl InspectorInfo {
...
+    .collect::<Vec<_>>();
+  http::Response::builder()
+    .status(http::StatusCode::OK)
+    .header(http::header::CONTENT_TYPE, "application/json")
+    .body(serde_json::to_string(&data).unwrap().into())
+}
+
+fn handle_json_version_request(
+  version_response: Value,
+) -> Result<http::Response<hyper::Body>, http::Error> {
```
bnoordhuis, line 223: Really just a suggestion but you could use `http::Result` for these to DRY:
```suggestion
) -> http::Result<http::Response<hyper::Body>> {
```

runtime/inspector.rs
```diff
@@ -142,95 +141,172 @@ impl InspectorInfo {
...
-  // `Send`, which should not be necessary because we are using the
-  // single-threaded Tokio runtime.
-  let inspector_map = HashMap::<Uuid, InspectorInfo>::new();
-  let inspector_map = Arc::new(Mutex::new(inspector_map));
-
-  let inspector_map_ = inspector_map.clone();
+  let inspector_map_ =
+    Rc::new(RefCell::new(HashMap::<Uuid, InspectorInfo>::new()));
+
+  let inspector_map = Rc::clone(&inspector_map_);
```
bnoordhuis, line 239: I rather like how you use UFCS here, nice and explicit. I might start using that in my own code.

runtime/inspector.rs
```diff
@@ -142,95 +141,172 @@ impl InspectorInfo {
...
+              handle_ws_request(req, inspector_map.clone())
+            }
+            (&http::Method::GET, "/json") => {
+              handle_json_request(inspector_map.clone())
+            }
+            (&http::Method::GET, "/json/version") => {
+              handle_json_version_request(json_version_response.clone())
+            }
+            _ => http::Response::builder()
+              .status(http::StatusCode::BAD_REQUEST)
```
bnoordhuis, line 285:
```suggestion
              .status(http::StatusCode::NOT_FOUND)
```
Seems more appropriate?

bnoordhuis (approved):
A few comments but overall very nice!

runtime/inspector.rs
lperlaki: Thank you, I think it is the recommended way of calling functions on types that implement `Deref`

runtime/inspector.rs
lperlaki: I was not sure about logging, the prior implementation also didn't log anything.

PR #8817, opened 2020-12-18
fix(runtime/websocket): remove eprintln

crowlKats:
@bartlomieju might there be other accidentally left debug prints?


bartlomieju:
@crowlKats good catch, I'm not sure if there are more rogue debug prints.

crowlKats:
just checked the resourcetable pr; there dont seem to be any other debug prints.

bartlomieju (approved):

PR #8822, opened 2020-12-18
fix: atomically write files in $DENO_DIR

lucacasonato:
This tries to fix an issue where files in the $DENO_DIR could be in an undermined state, where they are empty, only half written, or otherwise not correct. This issue may arise when multiple Deno processes try to cache the same file simultaneously. We should also consider adding integrity hashes to the files, so we can check if they are are valid when reading them from $DENO_DIR.

### DIFF
--- a/cli/fs_util.rs
+++ b/cli/fs_util.rs
@@ -2,7 +2,7 @@
 
 use deno_core::error::AnyError;
 pub use deno_core::normalize_path;
-use deno_crypto::rand;
+use deno_runtime::deno_crypto::rand;
 use std::env::current_dir;
 use std::fs::OpenOptions;
 use std::io::{Error, Write};
--- a/cli/fs_util.rs
+++ b/cli/fs_util.rs
@@ -2,11 +2,27 @@
 
 use deno_core::error::AnyError;
 pub use deno_core::normalize_path;
+use deno_crypto::rand;
 use std::env::current_dir;
 use std::fs::OpenOptions;
 use std::io::{Error, Write};
 use std::path::{Path, PathBuf};
 use walkdir::WalkDir;
+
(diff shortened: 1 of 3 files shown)


cli/fs_util.rs
```diff
@@ -2,12 +2,28 @@
...
+pub fn atomic_write_file<T: AsRef<[u8]>>(
+  filename: &Path,
+  data: T,
+  mode: u32,
+) -> std::io::Result<()> {
+  let rand: String = (0..4)
+    .map(|_| format!("{:02x}", rand::random::<u8>()))
+    .collect();
+  let extension = format!("{}.tmp", rand);
+  let tmp_file = filename.with_extension(extension);
```
ry, line 21: Use `tempfile::tempfile()` https://docs.rs/tempfile/3.1.0/tempfile/fn.tempfile.html

lucacasonato: Tempfile could be on a different disk. This way guarantees that we don't move between disks.

ry: What about `tempfile_in()` https://docs.rs/tempfile/3.1.0/tempfile/fn.tempfile_in.html

bartlomieju: https://docs.rs/tempfile/3.1.0/tempfile/fn.tempfile_in.html

lucacasonato: Why though? The docs say these temp files are auto cleaned by the OS when the last handle is dropped. We don't want that though. Deno dir files should be permanent.

ry: Right - makes sense

ry (approved):
LGTM - let's try it and see if it fixes our problem.

Issue #8788, opened 2020-12-16
"op_fetch_read" should use zero copy buffers


bartlomieju:
Currently "op_fetch_read" uses JSON dispatch and performs unnecessary serialization. The op should be rewritten to use zero copy buffers instead.

https://github.com/denoland/deno/blob/9fe26f8ca189ac81d9c20c454b9dbfa5e1011c3f/op_crates/fetch/lib.rs#L188-L222

Ref https://github.com/denoland/deno/issues/8787

Soremwar:
@bartlomieju Dibs on this

bartlomieju:
> @bartlomieju Dibs on this

@Soremwar do you mean you're working on this?

Soremwar:
Yeah, had some trouble with cargo build but PR should come this afternoon

00ff0000red:
@bartlomieju Closed in https://github.com/denoland/deno/pull/9036?

lucacasonato:
#9036 does not resolve #4735.

00ff0000red:
@lucacasonato Oops, my apologies, I was referring to https://github.com/denoland/deno/issues/4735#issuecomment-752193475, but yes, the root of the issue remains to be fixed.

PR #8826, opened 2020-12-19
fix(core): "op_fetch_read" should use zero copy buffers

Soremwar:
Fixes https://github.com/denoland/deno/issues/8788

### DIFF
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -198,11 +198,15 @@ pub async fn op_fetch_read(
   #[derive(Deserialize)]
   #[serde(rename_all = "camelCase")]
   struct Args {
+    length: u32,
     rid: u32,
+    start: u64,
   }
 
   let args: Args = serde_json::from_value(args)?;
   let rid = args.rid;
+  let start = args.start as usize;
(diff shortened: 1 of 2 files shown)


op_crates/fetch/lib.rs
```diff
@@ -212,11 +216,20 @@ pub async fn op_fetch_read(
...
+    if (start + length) > bytes {
+      (&mut zero_copy[0])[0..bytes - start]
+        .copy_from_slice(&chunk[start..bytes]);
+    } else {
+      (&mut zero_copy[0]).copy_from_slice(&chunk[start..start + length]);
+    }
+    Ok(json!({ "bytes": bytes as u64 }))
   } else {
-    Ok(json!({ "chunk": null }))
+    Ok(json!({"bytes": 0}))
```
bartlomieju, line 232: @Soremwar I think you want to lift code for `HttpBody` from https://github.com/denoland/deno/pull/8781 
I removed it a few days ago cause I couldn't find where it was used, now I remember that it was used by previous implementation of `fetch` that used zero copy.

lucacasonato:
@Soremwar Thanks, but superseded by #9036.

PR #8843, opened 2020-12-21
refactor(cli/tests): rewrite chown_test.ts not to depend on python

Liamolucko:
Rewrites `chown_test.ts` to use the GNU `id` command instead of python. This won't work on Windows, but these tests aren't currently run on Windows anyway.

Ref #8254


bartlomieju (approved):
LGTM, thank you @Liamolucko

Issue #8787, opened 2020-12-16
Add benchmark for "fetch" that downloads some big file


bartlomieju:
Currently we benchmark `fetch` API only in a single scenario (proxy). 

It'd be great if we benched some more real-life scenario like:
- download single big file (eg. 100Mb)
- download multiple smaller files (eg. 1Mb)

Benchmarks should use local test server for that purpose.

WingZer0o:
Seems like this PR just didn't go through because of a contributor license? 
https://github.com/denoland/deno/pull/11508

lucacasonato:
Going to close this, because it is stale.

PR #8844, opened 2020-12-21
feat(cli/tests): Added benchmark scenarios for fetch

ACov96:
Added scenarios for fetch as described in #8787 that download a single large file (43 MB to honor GitHub's 100 MB file size limit) and many small files (each about 1 MB) in size.

I'm unsure exactly how the tests/benchmarks are intended to be setup, so I took a best guess based on the existing tests. Please let me know if I can restructure/move anything, or feel free to make the change yourself, whichever is easiest :).


lucacasonato:
Thanks! Wou will need to add the benchmark to https://github.com/denoland/deno/blob/master/cli/bench/main.rs#L26, so it actually gets executed and recorded.

ACov96:
@lucacasonato Great, I'll add it/run it locally again and update the PR.

cli/tests/fetch_data/README.md
```diff
@@ -0,0 +1,7 @@
+This directory and subdirectories contains data for some of the fetch
+benchmarks.
+
+The following list of URLs is intended to credit the original
+creators/distributors of these files:
+
+- https://corpus.canterbury.ac.nz/descriptions/
```
bartlomieju, line 7: Not super happy about adding fixture files to repo, maybe you could get away without it, by creating some data on the fly in appropriate request handlers in `test_server`? We don't really need to have structured data, it's just a matter of downloading proper number of bytes.

ACov96: @bartlomieju I can change the benchmarks to use mock data instead, that does make for a cleaner codebase. 

Wondering if it should still be written to disk, or just stored/served from in memory? Or maybe both? Just want to make sure the benchmarks is capturing the right scenarios.

bartlomieju: @ACov96 I think generating data on demand and in-memory only should be sufficient for this benchmark.

lucacasonato (commented):
By using the std/http server for serving files, you are not just benchmarking the fetch implementation, but also std/http. I would prefer if you would use one of the rust test servers for benchmarking (for example `POST http://localhost:4545/echo_server`). That endpoint just echos back request body. You could also add a different endpoint for serving random data in the test server (`test_util/serc/lib.rs:339`).

ACov96:
@lucacasonato I see, makes sense to isolate the benchmark to just `fetch` rather than `fetch`+`std/http`.

> I would prefer if you would use one of the rust test servers for benchmarking

Are the test servers part of the benchmark harness already, or do I need to spin them up as a child process? When I try to `fetch` from `localhost:4545` in the benchmark, it errors out. I'm having trouble getting the specific error message, but I assume it's because the server isn't running when I do:

```
$ cargo bench -- --show-output # Having trouble getting the actual error, even with --show-output
```

PR #8845, opened 2020-12-21
BREAKING(unstable): Use hosts for net allowlists

### LINKED ISSUE #7587: Deno Permissions API request method unable to grant url "scoped" net permissions
The unstable Deno Permissions APIs `request()` method currently appears unable to grant url "scoped" permissions for the `net` `PermissionDescriptor` whose interface allows for providing an allowlist for urls.

Please see below for code to reproduce:

```ts
console.log(await Deno.permissions.query({ name: "net", url: "http://google.com" }));
console.log(await Deno.permissions.revoke({ name: "net" }));
console.log(await Deno.permissions.request({ name: "net", url: "http://google.com" }));
console.log(await Deno.permissions.query({ name: "net", url: "http://google.com" }));
console.log(await fetch("http://google.com"));
```

When executed with the following command:

```console
$ deno run --unstable --allow-net=google.com ./example.ts

PermissionStatus { state: "granted" }
PermissionStatus { state: "prompt" }
️⚠️  Deno requests network access to "http://google.com". Grant? [g/d (g = grant, d = deny)] g
PermissionStatus { state: "granted" }
PermissionStatus { state: "prompt" }
error: Uncaught PermissionDenied: network access to "http://google.com/", run again with the --allow-net flag
    at Object.jsonOpAsync (core.js:236:13)
    at async fetch (deno:op_crates/fetch/26_fetch.js:1272:29)
    at async example.ts:13:13
```

The expected behaviour is that following the permission request the following query should return a `PermissionStatus` with `state: "granted"` and the `fetch()` should be permitted.

Instead we are seeing the permission is not granted, despite the `granted` state response from the request call and the `fetch()` throws a `PermissionDenied` error.

nayeemrmn:
Allowlist checking already uses hosts but for some reason requests, revokes and the runtime permissions API use URLs.
- BREAKING(`lib.deno.unstable.d.ts`): Change `NetPermissionDescriptor::url` to `NetPermissionDescriptor::host`
- fix(`runtime/permissions`): Don't add whole URLs to the allowlist on request
- fix(`runtime/permissions`): Harden strength semantics (`{ name: "net", host: "127.0.0.1" }` _stronger than_ `{ name: "net", host: "127.0.0.1:8000" }`) for blocklisting
- refactor(`runtime/permissions`): Use tuples for hosts, make the host optional in `Permissions::{query_net, request_net, revoke_net}()`

---

Fixes #7587.

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -223,42 +223,27 @@ impl Permissions {
     PermissionState::Prompt
   }
 
-  pub fn query_net(&self, host: &str, port: Option<u16>) -> PermissionState {
+  pub fn query_net<T: AsRef<str>>(
+    &self,
+    host: &Option<&(T, Option<u16>)>,
+  ) -> PermissionState {
     if self.net.global_state == PermissionState::Denied
-      || check_host_and_port_list(host, port, &self.net.denied_list)
+      && match host.as_ref() {
+        None => true,
+        Some(host) => check_host_blocklist(host, &self.net.denied_list),
+      }
     {
       return PermissionState::Denied;
     }
     if self.net.global_state == PermissionState::Granted
-      || check_host_and_port_list(host, port, &self.net.granted_list)
+      || match host.as_ref() {
+        None => false,
+        Some(host) => check_host_allowlist(host, &self.net.granted_list),
+      }
     {
       return PermissionState::Granted;
     }
     PermissionState::Prompt
-  }
-
-  pub fn query_net_url(
-    &self,
-    url: &Option<&str>,
-  ) -> Result<PermissionState, AnyError> {
-    if url.is_none() {
-      return Ok(self.net.global_state);
-    }
-    let url: &str = url.unwrap();
-    // If url is invalid, then throw a TypeError.
-    let parsed = url::Url::parse(url)?;
-    // The url may be parsed correctly but still lack a host, i.e. "localhost:235" or "mailto:someone@somewhere.com" or "file:/1.txt"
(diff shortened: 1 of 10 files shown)


kitsonk (approved):
LGTM

kitsonk:
As this is a breaking change to an unstable API, we won't land this until 1.7.0.

crypt096 (approved):

runtime/permissions.rs
```diff
@@ -545,18 +546,31 @@ impl Permissions {
     )
   }
 
-  pub fn check_net(&self, hostname: &str, port: u16) -> Result<(), AnyError> {
-    self.query_net(hostname, Some(port)).check(
-      &format!("network access to \"{}:{}\"", hostname, port),
+  pub fn check_net<T: AsRef<str>>(
+    &self,
+    host: &(T, Option<u16>),
```
bartlomieju, line 551: Maybe `host: (&str, Option<u16>)` or split `host` and `port`? I think @magurotuna had similar problem in his DNS PR

nayeemrmn: `(&str, Option<u16>)` doesn't work because it's really hard to pass a `String` when a `&str` is expected as part of a tuple. Splitting `host` and `port` avoids the issue, but it's inconsistent and not applicable with the other APIs where this is wrapped in an `Option`.

bartlomieju: Thanks for explanation, that sounds reasonable to me, let's keep as is

bartlomieju (approved):
LGTM

PR #8846, opened 2020-12-21
refactor(test_util): replace "warp" with "hyper"

### LINKED ISSUE #8823: Remove dependency on "warp" in "test_util" crate
Follow up to #8782

In `test_util` crate `warp` is used as to crate an HTTP and WS server that is used by many unit and integration tests. 

We should rewrite the server to use `hyper` instead of `warp` in an effort to reduce number of dependencies.

yos1p:
Fixes #8823


yos1p:
I think this is ready for review. Please let me know what needs to be improved. @bartlomieju @bnoordhuis @piscisaureus

cli/tests/unit/fetch_test.ts
```diff
@@ -976,32 +976,15 @@ unitTest(
...
-AQH/MA0GCSqGSIb3DQEBCwUAA4IBAQB9AqSbZ+hEglAgSHxAMCqRFdhVu7MvaQM0
-P090mhGlOCt3yB7kdGfsIrUW6nQcTz7PPQFRaJMrFHPvFvPootkBUpTYR4hTkdce
-H6RCRu2Jxl4Y9bY/uezd9YhGCYfUtfjA6/TH9FcuZfttmOOlxOt01XfNvVMIR6RM
-z/AYhd+DeOXjr35F/VHeVpnk+55L0PYJsm1CdEbOs5Hy1ecR7ACuDkXnbM4fpz9I
-kyIWJwk2zJReKcJMgi1aIinDM9ao/dca1G99PHOw8dnr4oyoTiv8ao6PWiSRHHMi
-MNf4EgWfK+tZMnuqfpfO9740KzfcVoMNo4QJD4yn5YxroUOO/Azi
------END CERTIFICATE-----
-`,
+        caData: caPem,
       },
```
bartlomieju, line 988: @yos1p looks like this test was changed inadvertedly, please revert

bartlomieju (commented):
@yos1p looks great, especially number of removed lines in `Cargo.lock`! I think we could land it after reverting WebSocket test

test_util/src/lib.rs
```diff
@@ -253,236 +359,387 @@ pub async fn run_all_servers() {
...
+        return Ok(file_resp);
+      }
+
+      return Ok(Response::new(Body::empty()));
+    }
+  };
+}
+
+/// Taken from example in https://github.com/ctz/hyper-rustls/blob/a02ef72a227dcdf102f86e905baa7415c992e8b3/examples/server.rs
+struct HyperAcceptor<'a> {
```
piscisaureus, line 563: @bartlomieju No objections from me here, this looks fine.

cli/tests/unit/fetch_test.ts
yos1p: Reverted. Fyi, `cargo test` always failed in local because TLS problem with certs in `cli/tests/tls`. Apparently, this is fixed by regenerating the new one, that's why I have to change the CA.pem file as well and modify the test.

Since it works in CI, I think should be okay. Just wondering why it fails in local though.

bnoordhuis (approved):
Some suggestions but pretty much LGTM. Nice work.

test_util/src/lib.rs
```diff
@@ -253,236 +359,387 @@ pub async fn run_all_servers() {
...
+  acceptor: Pin<
+    Box<
+      dyn Stream<
+          Item = Result<
+            tokio_rustls::server::TlsStream<TcpStream>,
+            std::io::Error,
+          >,
```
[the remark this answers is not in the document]
yos1p: Applied.

test_util/src/lib.rs
```diff
@@ -117,132 +131,224 @@ pub fn test_server_path() -> PathBuf {
...
-        let h = res.headers_mut();
-        if let Some(v) = content_type {
-          h.insert("content-type", HeaderValue::from_str(&v).unwrap());
-        }
-        if let Some(v) = user_agent {
-          h.insert("user-agent", HeaderValue::from_str(&v).unwrap());
+async fn get_tls_config(
+  cert: &str,
+  key: &str,
+) -> Result<Arc<rustls::ServerConfig>, std::io::Error> {
```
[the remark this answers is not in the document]
yos1p: Applied

test_util/src/lib.rs
```diff
@@ -117,132 +131,224 @@ pub fn test_server_path() -> PathBuf {
...
+      rx.forward(tx)
+        .map(|result| {
+          if let Err(e) = result {
+            println!("websocket server error: {:?}", e);
+          }
+        })
+        .await;
+    }
   }
+}
```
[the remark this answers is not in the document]
yos1p: Sounds good. Applied.

cli/tests/unit/fetch_test.ts
bartlomieju: @yos1p interesting, maybe we should regenerate certificates in a new PR then

PR #8847, opened 2020-12-21
refactor(core): simplify Deno.core initialisation, remove stale TODO 

bartlomieju:
This commit rewrites initialisation of the "shared queue" and
in effect prevents from double execution of "core/core.js" and
"core/error.js". 

Previously both of these files were executed every time a "JsRuntime"
was created. That lead to a situation where one copy of each script
was included in the snapshot and then another copy would be
executed after loading the snapshot.

Effectively "JsRuntime::shared_init" was removed; instead execution 
of those scripts and actual initialisation of shared queue
was split into two helper functions: "JsRuntime::js_init" and
"JsRuntime::share_queue_init".

Additionally stale TODO comments were removed.


ry (approved):
LGTM

bartlomieju:
Hit a problem with inspector no longer working, closing for now, I might revisit in the future.

ry (approved):
LGTM

PR #8850, opened 2020-12-22
fix(lsp): respect enable flag for requests

kitsonk:
This PR fixes situations with the lsp, where the extension is disable, but on hover and other requests, there are duplicate values reported to vscode.


ry (approved):
👍

PR #8855, opened 2020-12-22
fix: implement readable stream fetch body

lucacasonato:
This implements using `ReadableStream` as a `fetch` body.

### DIFF
--- a/op_crates/fetch/26_fetch.js
+++ b/op_crates/fetch/26_fetch.js
@@ -1246,8 +1246,12 @@
             body = multipartBuilder.getBody();
             contentType = multipartBuilder.getContentType();
           } else {
-            // TODO: ReadableStream
-            throw new Error("Not implemented");
+            // TODO(lucacasonato): do this in a streaming fashion once we support it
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):

op_crates/fetch/26_fetch.js
```diff
@@ -1246,8 +1246,12 @@
             body = multipartBuilder.getBody();
             contentType = multipartBuilder.getContentType();
           } else {
-            // TODO: ReadableStream
-            throw new Error("Not implemented");
+            // TODO(lucacasonato): do this in a streaming fashion once we support it
+            const buf = new Buffer();
+            for await (const chunk of init.bodb) {
```
crowlKats, line 1251:
```suggestion
            for await (const chunk of init.body.getIterator()) {
```

lucacasonato: Why? Body is an async iterator already.

crowlKats: my bad, you are right

Issue #8861, opened 2020-12-22
"cargo publish" failed because of new files in git


bartlomieju:
https://github.com/denoland/deno/runs/1595775796

```
Run cd cli
    Updating crates.io index
error: 18 files in the working directory contain changes that were not yet committed into git:

tests/.test_coverage/0cb04771-ad9e-4f01-a11a-7116d83fd046.json
tests/.test_coverage/35949f05-c4a5-474f-ab4a-598dfb45e0b9.json
tests/.test_coverage/440ebed6-f6a1-4638-8b37-cd2d55727da1.json
tests/.test_coverage/45449ac3-cdea-4579-a84b-8b04b517aa00.json
tests/.test_coverage/4e93460f-c8e1-48f1-8d9a-2ab29ce7db34.json
tests/.test_coverage/65d970ea-6189-4986-a762-06172b63680c.json
tests/.test_coverage/6818033c-d4bc-4886-ae66-16853a74eede.json
tests/.test_coverage/7aa15781-3bb6-4ee9-b2b1-0da364982d91.json
tests/.test_coverage/8bd37ecf-fb3f-4586-a335-916c7b8b284c.json
tests/.test_coverage/aa13ee4b-9f84-4779-8382-34e2dd18a703.json
tests/.test_coverage/af201c2b-fc16-45d1-abef-a94f5a81f9ab.json
tests/.test_coverage/b92862c0-6a24-4c7b-8fb2-647cd70f4193.json
tests/.test_coverage/bf0787fb-8448-45e1-b54d-87521150e203.json
tests/.test_coverage/c050b8f1-a1ef-46b2-89e6-53754625f179.json
tests/.test_coverage/ce69ed16-18bb-4662-80f3-4d954499fd17.json
tests/.test_coverage/d62973c2-b426-4472-b17f-20c74318b928.json
tests/.test_coverage/da5a32c4-cab6-41e7-98b7-cc0d0ef3aa5d.json
tests/.test_coverage/dc2004b1-35d7-409f-ab77-b55b6894fecb.json

to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
```

These files should be excluded in `.gitignore` or removed after the test.

CC @caspervonb

bartlomieju:
Fixed in #8862

PR #8862, opened 2020-12-22
fix(build): add generated coverage files to gitignore

caspervonb:
Hot fix for #8861

Preferably this would just write to a temporary directory, `itest` doesn't really want to roll that so I'll follow up with a PR to address the coverage tests as a whole.


bartlomieju (approved):
LGTM

PR #8865, opened 2020-12-22
fix(lsp): Respect client capabilities for config and dynamic registration

SyrupThinker:


cli/lsp/language_server.rs
```diff
@@ -494,6 +504,17 @@ impl lspower::LanguageServer for LanguageServer {
...
     _params: DidChangeConfigurationParams,
   ) {
+    if !self
+      .config
+      .read()
+      .unwrap()
+      .client_capabilities
+      .workspace_configuration
+    {
+      // Client does not support workspace configuration
```
lucacasonato, line 514: I think we could use the passed settings in the `params` in this case.

kitsonk (approved):
LGTM, thanks for the contribution!

Issue #4350, opened 2020-03-13
Bug: Promises don't handle asynchronous errors correctly


timonson:
The following example is pretty much self-explanatory:

```javascript
new Promise(function(resolve, reject) {
  setTimeout(() => {
    throw new Error("Whoops!")
  }, 1000)
}).catch(console.log)
```

Instead of throwing the `Error` deno ignores it.

kevinkassimo:
The problem here is not about why `.catch()` does not work (it is expected to not catch anything here).
Instead the main problem is that errors thrown in `setTimeout` is not properly logged and terminate the program.

An actual problem that causes this behavior is that `queueMicrotask` (used by timers) does not terminate the program on uncaught errors, e.g.
```js
queueMicrotask(() => { throw new Error("123"); });
// no error logged and program continues for future microtasks
```

hayd:
I've also seen if you don't await certain promises the program never exits.
But haven't been able to get a clean example...

bartlomieju:
Tested with `v0.42.0`:
```
// test.js
new Promise(function(resolve, reject) {
  setTimeout(() => {
    throw new Error("Whoops!")
  }, 1000)
}).catch(console.log)

$ deno test.js
error: Uncaught Error: Whoops!
    throw new Error("Whoops!")
          ^
    at file:///Users/biwanczuk/dev/deno/test.js:3:11
    at fire ($deno$/web/timers.ts:179:3)
    at handleTimerMacrotask ($deno$/web/timers.ts:44:5)
```

Seems it's been fixed.

Liamolucko:
It's been fixed for `setTimeout()`, but not `queueMicrotask()`.

This code will continue to run:
```typescript
queueMicrotask(() => {
  throw new Error("Whoops!");
});

setInterval(() => {}, 100);
```

bnoordhuis:
I've opened https://github.com/denoland/rusty_v8/issues/574, that's one avenue for fixing this issue.

Other avenues are the debug inspector (can catch uncaught exceptions) or add more hooks to V8.

nayeemrmn:
We should close this in favour of #14158 for clarity

PR #8867, opened 2020-12-23
fix(std/node): Stop callbacks being called twice when callback throws error

### LINKED ISSUE #8815: Promise as Callbacks using dangerous pattern
While working on some node/fs feature it was noticed that many of the compatibility modules used Promise based Deno builtins inside Callback based implementations, in a dangerous way.  

"if callback throws you reach the catch and the callback is called again with an error"

The problem was reported in my module by @benjamingr but it spans to many of them.

- _fs_appendFile
- _fs_chmod
- _fs_chown
- _fs_copy
- _fs_dir
- _fs_exists
- _fs_lstat
- _fs_mkdir
- _fs_open
- _fs_readfile
- _fs_realpath
- _fs_rename
- _fs_rmdir
- _fs_stat
- _fs_unlink
- _fs_writeFile

Liamolucko:
Closes #8815.


bartlomieju:
@Liamolucko thanks for the PR, looks good, no comments on the code. Could you add some test cases showcasing the problem described in #8815 so we're sure there's no regressions in the future?

Liamolucko:
Should something be written down somewhere about this issue? I wouldn't be surprised if more of these bugs pop up as `std/node` is expanded.

bartlomieju:
> Should something be written down somewhere about this issue? I wouldn't be surprised if more of these bugs pop up as `std/node` is expanded.

@Liamolucko yes, a note in `std/node/README.md` would be very helpful 👍

guybedford:
On the topic of callbach handling for Node.js I've also posted https://github.com/denoland/deno/issues/8897 for error remapping. It's a separate issue, but the two problems are still closely related.

std/node/_crypto/randomBytes.ts
```diff
@@ -50,9 +52,12 @@ export default function randomBytes(
       ) {
         throw e;
       } else {
-        cb(e);
+        err = e;
       }
     }
+    setTimeout(() => {
+      cb(err, bytes);
+    }, 0);
```
bnoordhuis, line 60:
```suggestion
    cb(err, bytes);
```
If the setTimeout is there for a reason, can you add an explaining comment? Thanks.

std/node/_fs/_fs_close.ts
```diff
@@ -2,14 +2,15 @@
 import type { CallbackWithError } from "./_fs_common.ts";
 
 export function close(fd: number, callback: CallbackWithError): void {
-  queueMicrotask(() => {
+  setTimeout(() => {
```
bnoordhuis, line 5: Can you explain this change?

std/node/_fs/_fs_lstat.ts
```diff
@@ -36,9 +36,10 @@ export function lstat(
 
   if (!callback) throw new Error("No callback function supplied");
 
-  Deno.lstat(path)
-    .then((stat) => callback(undefined, CFISBIS(stat, options.bigint)))
-    .catch((err) => callback(err, err));
+  Deno.lstat(path).then(
+    (stat) => callback(undefined, CFISBIS(stat, options.bigint)),
```
bnoordhuis, line 40:
```suggestion
    (stat) => callback(null, CFISBIS(stat, options.bigint)),
```
(node always passes null if there is no error)

std/node/_fs/_fs_lstat.ts
```diff
@@ -36,9 +36,10 @@ export function lstat(
...
+    (err) => callback(err, err),
```
bnoordhuis, line 41:
```suggestion
    (err) => callback(err),
```
(`arguments.length` == 1 when a fs callback fails)

Similar comments apply to `std/node/_fs/_fs_open.ts` and `std/node/_fs/_fs_stat.ts`.

bnoordhuis (commented):
Thanks for the pull request, I admire your testing rigor!

General observation:

1. Successful fs ops always invoke the callback like this `cb(null, res)`, never `cb(undefined, res)`. There's still some work left to be done there. I pointed out some instances but not all.

2. Unsuccessful fs ops invoke `cb(err)`, i.e., `arguments.length === 1`.

That said, this PR is already a big improvement on the status quo!

Apropos testing, there's a lot of repetition right now. I'd move the boilerplate into a helper function and then simply pass in the script-to-eval as an argument.

I think it's okay to start deno with `--unstable` unconditionally, no need to special case that.

Thanks again!

std/node/_fs/_fs_close.ts
Liamolucko: At the moment `queueMicrotask()` doesn't exit the program when the callback throws an error (#4350), so I replaced it with `setTimeout()`. To be fair that is a Deno issue though, so maybe it should be left alone?

bnoordhuis: Yes, please leave it for now. I'll comment on the other issue, I think I know what needs to be done in order to fix that.

std/node/_crypto/randomBytes.ts
Liamolucko: Since the callback is called synchronously, any errors it throws can be caught with a try/catch around `randomBytes()`. That isn't really related to this issue, but I originally did it because for some reason [a V8 debug check was failing](https://github.com/denoland/deno/runs/1610754385#step:25:4138) when pbkdf2 was synchronous.

bartlomieju (approved):
LGTM, thank you @Liamolucko especially for updating the README for `std/node` module. I will wait on Ben's review before landing

std/node/_fs/_fs_close.ts
bnoordhuis: This is still unresolved but I suppose it doesn't matter much in practice.

bnoordhuis (approved):
Thanks, LGTM!

PR #8868, opened 2020-12-23
test(cli): ensure await Promise.all does not race

caspervonb:
Revisits the test case added in #8802 to be order sensitive.

See @piscisaureus's comment https://github.com/denoland/deno/pull/8802#issuecomment-749333750


ry (approved):
LGTM - thanks @caspervonb

PR #8872, opened 2020-12-24
fix(lsp): provide diagnostics for unresolved modules

kitsonk:
Provides diagnostics for unresolved modules.

An iteration on this would be to provide a quick fix to do the `deno cache` for non-local modules.


lucacasonato (approved):
LGTM

PR #8878, opened 2020-12-24
chore: move ProgressEvent type to deno.web lib

lucacasonato:
`ProgressEvent` is a global that is defined in the web op crate, and used there (including in the TS typings), but the type definition was part of `lib.deno.shared_globals`. This corrects this.


bartlomieju (approved):
LGTM

PR #8882, opened 2020-12-24
refactor(cli/dts): remove unnecessary caFile field

iamkacperwasik:
Remove caFile field from CreateHttpClientOptions interface in lib.deno.unstable.d.ts.


bartlomieju (changes requested):
Thanks for the PR, besides removing `caFile` from declaration file, implementation in Rust should be removed (in `op_crates/fetch/lib.rs`). You might also update some tests in `cli/tests/unit/fetch_test.ts`

iamkacperwasik:
Looks like it's fails some of tests without changing anything in code

PR #8887, opened 2020-12-26
feat: add markdown support to deno fmt

### LINKED ISSUE #7352: deno fmt should format Markdown including code blocks
At the moment Deno only formats JS and TS, but dprint can also format markdown including JS and TS code blocks:

https://github.com/denoland/deno/pull/6682#issue-445589767

This would be an awesome feature for documentation, is it possible to enable it? Then we can get rid of Prettier.

satyarohith:
This PR adds markdown support to `deno fmt`. The js(x)/ts(x) code blocks in markdown are formatted using the same config used for ts/tsx and js/jsx files. To format markdown from standard input, you need to pass `--ext=md`  flag or it will be processed as TS and will result in errors.

I'm new to writing rust - critique is appreciated. :)

Closes #7352


cli/fs_util.rs
```diff
@@ -90,17 +90,26 @@ pub fn resolve_from_cwd(path: &Path) -> Result<PathBuf, AnyError> {
...
-    .map(|e| e.to_lowercase());
-  if let Some(ext) = lowercase_ext {
-    ext == "ts" || ext == "tsx" || ext == "js" || ext == "jsx" || ext == "mjs"
+  if let Some(ext) = get_extension(path) {
+    ext == "ts"
+      || ext == "tsx"
+      || ext == "js"
+      || ext == "jsx"
+      || ext == "mjs"
+      || ext == "md"
```
RDambrosio016, line 99:
```suggestion
    matches!(ext.as_str(), "ts" | "tsx" | "js" | "jsx" | "mjs" | "md");
```
small refactor to make this more idiomatic

bartlomieju:
@satyarohith thanks for the PR! Please add a test case for badly formatted Markdown file (similar to `cli/tests/badly_formatted.mjs`).

cli/fs_util.rs
satyarohith: Nice - thank you! :)

cli/fs_util.rs
```diff
@@ -90,17 +90,30 @@ pub fn resolve_from_cwd(path: &Path) -> Result<PathBuf, AnyError> {
...
-    ext == "ts" || ext == "tsx" || ext == "js" || ext == "jsx" || ext == "mjs"
+  if let Some(ext) = get_extension(path) {
+    matches!(ext.as_str(), "ts" | "tsx" | "js" | "jsx" | "mjs")
+  } else {
+    false
+  }
+}
+
+/// This function is similar to is_supported_ext but also allows .md extension.
+pub fn is_supported_ext_md(path: &Path) -> bool {
```
satyarohith, line 101: I choose to duplicate the function (also results in duplicate test cases) instead of calling `is_supported_ext` for the rest of the extensions because I thought the latter would result in unnecessary function calls.

cli/flags.rs
```diff
@@ -764,6 +766,12 @@ Ignore formatting a file by adding an ignore comment at the top of the file:
         .help("Check if the source files are formatted")
         .takes_value(false),
     )
+    .arg(
+      Arg::with_name("stdin-markdown")
+        .long("stdin-markdown")
+        .help("Consider standard input (stdin) as markdown")
```
hayd, line 772: Does it make more sense to be --stdin-ext=md ? Or something similar... rather than a bool?

satyarohith: It does, but I'm a little concerned that the flag might confuse users into thinking that it can accept other extensions too. Right now, TS(X)/JS(X) are autodetected and require no input from users for formatting to work properly. I thought to go with the same route but some inputs from @bartlomieju, and the `--ts` flag of `deno eval` motivated me to go with this simple solution. 

I feel like `--stdin-markdown` is more explicit and leads to less confusion. Let me know your thoughts. :)

bartlomieju: Maybe `--ext={js|jsx|ts|tsx|md}`? (though I'm not sure if there's any difference for js(x) and ts(x) files. @lucacasonato wdyt?

nayeemrmn: We should use a global solution that could potentially be used by `deno run -`, `deno lint -`, `deno eval` to choose between js and ts. I don't think `deno eval --ts` was good. I think we want some abbreviation of `--anonymous-media-type`.

satyarohith: I'm +1 for a global solution. What do you'll think of `--content-type={js|jsx|ts|tsx|md}`?

> (though I'm not sure if there's any difference for js(x) and ts(x) files.

There's no difference as they're all processed using `dprint-plugin-typescript`.

bartlomieju: > I think we want some abbreviation of `--anonymous-media-type`.

I'm not sure I understand this bit Nayeem, care to explain? 

> I'm +1 for a global solution. What do you'll think of --content-type={js|jsx|ts|tsx|md}?

`--content-type` sounds good, but I still like `--ext` better. 

We need more opinions on this flag name.

satyarohith: Since the flag will be used across commands, your (@lucacasonato, @ry, @kitsonk, @kt3k) thoughts/feedback will be helpful.

bartlomieju: > We should use a global solution that could potentially be used by `deno run -`, `deno lint -`, `deno eval` to choose between js and ts. I don't think `deno eval --ts` was good. I think we want some abbreviation of `--anonymous-media-type`.

Discussed offline with @lucacasonato and we think that `--ext` is preferable. @nayeemrmn I do agree that `deno eval --ts` was a mistake. I'm still not sure what you meant by `--anonymous-media-type`...

nayeemrmn: `--ext` is fine, my point was just to say that the flag should work to provide the content type of any anonymous module (like eval, stdin) across cli.

bartlomieju: Let's address that in follow up PRs

cli/tests/badly_formatted_fixed.md
```diff
@@ -0,0 +1,15 @@
+<!-- Not using an extension to avoid triggering dprint. -->
```
bartlomieju, line 1: What do you mean by this comment? You can exclude files from `dprint` formatting in `.dprintrc.json`

satyarohith: I guess I followed `badly_formatted.mjs`. I updated it.

Cargo.lock
```diff
@@ -655,13 +665,26 @@ dependencies = [
...
+]
+
 [[package]]
 name = "dprint-plugin-typescript"
 version = "0.38.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67bb41b131e30df80105eb6b4f183a4bf003ee7dfb20661ac15a2a6f329de5b1"
 dependencies = [
- "dprint-core",
+ "dprint-core 0.34.0",
```
lucacasonato, line 687: This is unfortunate, but not a blocker.

cli/tools/fmt.rs
```diff
@@ -101,16 +99,12 @@ async fn check_source_files(
...
       let file_text = read_file_contents(&file_path)?.text;
-      let r: Result<String, String>;
-      let ext = match get_extension(&file_path) {
-        Some(ext) => ext,
-        None => String::new(),
-      };
-      if ext == "md" {
-        r = format_markdown(&file_text)
+      let ext = get_extension(&file_path).unwrap_or_else(String::new);
+      let r = if ext == "md" {
```
satyarohith, line 103: this is so better

cli/flags.rs
```diff
@@ -803,6 +807,14 @@ Ignore formatting a file by adding an ignore comment at the top of the file:
         .help("Check if the source files are formatted")
         .takes_value(false),
     )
+    .arg(
+      Arg::with_name("ext")
+        .long("ext")
+        .help("Set standard input (stdin) content type")
+        .takes_value(true)
+        .default_value("ts")
```
lucacasonato, line 815: Wasn't the default `tsx` previously?

Cargo.lock
bartlomieju: I think we could upgrade `dprint` before landing to ensure only a single version

cli/flags.rs
satyarohith: I'm not sure if there's a default before, but all extensions (ts, tsx, js, jsx) are processed using the same plugin (`dprint-plugin-typescript`) - so we can use any one of them as default.

Cargo.lock
bartlomieju: Opened https://github.com/dprint/dprint-plugin-markdown/pull/29

cli/flags.rs
```diff
@@ -803,6 +807,14 @@ Ignore formatting a file by adding an ignore comment at the top of the file:
         .help("Check if the source files are formatted")
         .takes_value(false),
     )
+    .arg(
+      Arg::with_name("ext")
+        .long("ext")
+        .help("Set standard input (stdin) content type")
```
ry, line 813: I think `--kind` would make more sense. It's more about the type of file rather than the extension...

satyarohith: `--kind` sounds more appropriate to me. @bartlomieju wdyt?

dsherret: Something to consider, in dprint I settled on:

```ts
dprint fmt --stdin <file-path/file-name>
```

The reason is that if you provide a full file path it will format according to the extension and apply the inclusion/exclusion rules specified in the dprint config file (so if the file is excluded, then it will stdout the inputted text). If you provide just the file name, then it will always format according to the extension.

This would be useful if in deno you ever had some configuration to ignore files then someone wouldn't have to re-implement the logic and could provide a full file path instead.

dsherret:
We should add a test for ignore comments, but `deno-fmt-ignore` isn't implemented in dprint-plugin-markdown yet (it only has `dprint-ignore`, `dprint-ignore-start` and `dprint-ignore-end`, but there should be configuration for providing the ignore comment prefix text). I would recommend we implement that in dprint-plugin-markdown before this lands.

bartlomieju:
> We should add a test for ignore comments, but `deno-fmt-ignore` isn't implemented in dprint-plugin-markdown yet (it only has `dprint-ignore`, `dprint-ignore-start` and `dprint-ignore-end`, but there should be configuration for providing the ignore comment prefix text). I would recommend we implement that in dprint-plugin-markdown before this lands.

Yikes... I didn't think about this. Yeah I believe this is a blocker for landing this PR then.

cli/flags.rs
bartlomieju: > This would be useful if in deno you ever had some configuration to ignore files then someone wouldn't have to re-implement the logic and could provide a full file path instead.

Still debating about config files, but I'm not entirely sold on the idea of passing filename if reading from stdin, effectively users would have to put dummy file.

> I think --kind would make more sense. It's more about the type of file rather than the extension...

Just discussed offline, let's go with `--ext`

Cargo.lock
dsherret: Resolved in markdown 0.4.4, typescript 0.38.2

bartlomieju: Thanks!

dsherret:
@bartlomieju opened https://github.com/dprint/dprint-plugin-markdown/issues/30 -- I don't have time for the rest of today to work on it, so someone else could take it (otherwise I'll get to it eventually).

cli/tools/fmt.rs
```diff
@@ -63,8 +61,30 @@ pub async fn format(
...
+      } else {
+        Ok(text.to_string())
+      }
+    }),
+  )
+}
+
 async fn check_source_files(
-  config: dprint::configuration::Configuration,
+  config: dprint_plugin_typescript::configuration::Configuration,
```
dsherret, line 87: Right now only the typescript configuration is passed down here, but not the markdown. Then the `format_markdown` it does `get_typescript_config()`. Perhaps something should be changed?

cli/tools/fmt.rs
```diff
@@ -213,9 +255,16 @@ fn files_str(len: usize) -> &'static str {
...
+fn get_typescript_config(
+) -> dprint_plugin_typescript::configuration::Configuration {
+  dprint_plugin_typescript::configuration::ConfigurationBuilder::new()
+    .deno()
+    .build()
+}
+
+fn get_markdown_config() -> dprint_plugin_markdown::configuration::Configuration
+{
+  dprint_plugin_markdown::configuration::ConfigurationBuilder::new().build()
```
dsherret, line 267: We might want to have a discussion on what configuration would be best. Right now, deno's repo does `"textWrap": "always"` (https://github.com/denoland/deno/blob/master/.dprintrc.json), but the default is "maintain".

cli/tools/fmt.rs
```diff
@@ -63,8 +61,30 @@ pub async fn format(
...
 
+/// Formats markdown (using https://github.com/dprint/dprint-plugin-markdown) and its code blocks
+/// (ts/tsx, js/jsx).
+fn format_markdown(file_text: &str) -> Result<String, String> {
+  let ts_config = get_typescript_config();
+  let md_config = get_markdown_config();
+  dprint_plugin_markdown::format_text(
+    &file_text,
+    &md_config,
+    Box::new(move |tag, text, _| {
```
dsherret, line 72: Sorry, the docs do not explain this, but you may want to format the TypeScript code using the `line_width` provided to this closure (the third parameter), as it may contain the line width to use to format the code block.

For example, a code block could be a couple levels deep in the markdown and instead of using 80 characters of width, it should now use 72.

```ts
Box::new(move |tag, text, line_width| {
    ts_config.line_width = line_width;
    // etc...
})
```

...but that's only if you want this behaviour.

cli/tools/fmt.rs
```diff
@@ -63,8 +61,30 @@ pub async fn format(
...
+      if matches!(tag, "ts" | "tsx" | "js" | "jsx") {
```
dsherret, line 73: Perhaps this should be case insensitive?

dsherret, line 73: Sometimes people do `typescript` or `javascript` in the tag as well.

cli/tools/fmt.rs
dsherret, line 267: dprint-plugin-markdown 0.5.0 has the new configuration to configure the ignore directive text (thanks @bartlomieju for implementing!). You should be able to specify `deno-ignore`, `deno-ignore-start`, and `deno-ignore-end` (see here: https://github.com/dprint/dprint-plugin-markdown/pull/31/files#diff-08092886a134ae07f7991c185445595220f6e624d8d1b3f9c7bf52c7d02e58f0)

bartlomieju: Awesome, thank you David for quick release :)

cli/tools/fmt.rs
```diff
@@ -63,8 +61,38 @@ pub async fn format(
...
+      if matches!(
+        tag.as_str(),
+        "ts" | "tsx" | "js" | "jsx" | "javascript" | "typescript"
+      ) {
+        let mut codeblock_config = ts_config.clone();
+        codeblock_config.line_width = line_width;
+        dprint_plugin_typescript::format_text(
+          &PathBuf::from("_rand.".to_owned() + &tag),
+          &text,
+          &ts_config,
```
dsherret, line 85: Should be `codeblock_config` here.

cli/tools/fmt.rs
```diff
@@ -63,8 +61,38 @@ pub async fn format(
...
+    Box::new(move |tag, text, line_width| {
+      let tag = tag.to_lowercase();
+      if matches!(
+        tag.as_str(),
+        "ts" | "tsx" | "js" | "jsx" | "javascript" | "typescript"
+      ) {
+        let mut codeblock_config = ts_config.clone();
+        codeblock_config.line_width = line_width;
+        dprint_plugin_typescript::format_text(
+          &PathBuf::from("_rand.".to_owned() + &tag),
```
dsherret, line 83: I'm not sure what this will do when the tag is "javascript" or "typescript", but I think this needs to be updated.

bartlomieju (approved):
LGTM, thank you Satya, this is very desirable feature!

cli/flags.rs
dsherret: @bartlomieju could be:

```ts
deno fmt --stdin ts
deno fmt --stdin js
deno fmt --stdin tsx

# then if config files are ever added in the future and you want the exclusion/inclusion rules
deno fmt --stdin /home/david/dev/my-project/file.ts
```

PR #8888, opened 2020-12-26
ci: setup latest version of Node

### LINKED ISSUE #8786: CI should use latest stable version of Node
Currently CI uses built-in version of Node, which might be out of date. 

Therefore we should ensure that CI uses latest version of Node so that benchmarks are up to date.

ah-yu:
fix #8786


ry (approved):
LGTM - thanks @ah-yu

PR #8891, opened 2020-12-26
refactor(runtime): return iterator from resolve_addr

magurotuna:
This PR does refactoring for `runtime/resolve_addr.rs`.
At first I did it in the other PR (#8790) but I have extracted it as a separate PR to allow #8790 to focus on `Deno.resolveDns` API.

The changes in this PR are:

- to let `resolve_addr` and `resolve_addr_sync` return iterators of `SocketAddr`, not just a single `SocketAddr`. These functions previously called `.next()` on an iterator internally, and returned it, so the callers of these functions had no way of handling other items than the first item in the iterator.  This change allows the caller to decide how to use the returned iterator.
- to use `generic_error` in deno_core instead of `anyhow::Context`. When I worked on #8743, I didn't know there were utility error functions like `generic_error`. 

See also https://github.com/denoland/deno/pull/8790#discussion_r546851138

CC @bartlomieju

### DIFF
--- a/runtime/ops/net.rs
+++ b/runtime/ops/net.rs
@@ -202,7 +202,10 @@ async fn op_datagram_send(
         s.borrow::<Permissions>()
           .check_net(&args.hostname, args.port)?;
       }
-      let addr = resolve_addr(&args.hostname, args.port).await?;
+      let addr = resolve_addr(&args.hostname, args.port)
+        .await?
+        .next()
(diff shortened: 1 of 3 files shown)


bartlomieju:
LGTM, I'll let @ry review before landing.

ry (approved):
LGTM - thanks @magurotuna

PR #8906, opened 2020-12-28
Images optimization 🚀

AleTid5:


bartlomieju:
Thanks for the PR, but it's not needed, images are served optimized by CDN.

Issue #8643, opened 2020-12-07
Deno language server to-do list


kitsonk:
Post the initial implementation of the language server, there are quite a few other features that need to be added as well as some "tech debt" from the initial implementation.  We might want to break some of these off to individual issues, but it is good enough for now to have a "master" issue for tracking purposes.

### Language Service Features

These are server capabilities provided by the language server protocol, which various editors can take advantage of.  Most of these features are exposed [here](https://github.com/microsoft/vscode/tree/master/extensions/typescript-language-features/src/languageFeatures) in visual studio code for TypeScript/JavaScript and provide a high level structure of how to take information from the `tsc` compiler isolate and bring it into Rust and send it back over the language server client.  Specifically, the _DefinitionProvider_ is a good model of how to add a feature.  Looking at the `cli/lsp/mod.rs` for the `GotoDefinition` handling demonstrates calling into `tsc` and dealing with a response back, converting it to a valid lsp response.

- [x] TypeDefinitionProvider:star:  (#12789)
- [x] WorkspaceSymbolProvider (#12787)
- [x] Workspace Folders (multiple roots per workspace) (#10488)
- [x] Completion Provider ~(this will be the most complex one, because we also need to enrich the response back from tsc)~ an mvp exists (#8651), but we likely want to enhance it.
  - [x] ~Enable `include_completions_with_insert_text` (in consequence enabling `include_automatic_optional_chain_completions` and `include_completions_for_module_exports`)~ _this is largely fixed_
  - [x] Registry completions (#9934)
- [x] SignatureHelpProvider :star: (#9330)
- [X] CodeActionProvider :star: 
   - [X] QuickFix
   - [x] `deno cache`
- [x] ImplementationProvider :star: (#9071)
- [X] CodeLensProvider (#9316)
   - [X] References (#9316)
   - [X] Implementations
   - [x] Deno specific code lenses, like tests? (#10874)
- [x] RenameProvider (#8910)
- [x] SelectionRangeProvider (#9845)
- [x] FoldingRangeProvider (#9900)
- [x] CallHierarchyProvider (#10061)
- [x] SemanticTokensProvider (#10233)
- [x] DocumentSymbolProvider (#9981)
- [ ] DocumentLinkProvider - _vscode only implements this for tsconfig.json, we should consider it for `deno.jsonc`_
- [ ] DocumentRangeFormattingProvider (maybe)
- [ ] DocumentOnTypeFormattingProvider (maybe)
[...]

PR #8910, opened 2020-12-29
feat(lsp):  Implement textDocument/rename

hrsh7th:
Related #8643

Currently, the ConfigureRequest (in `cli/tsc/compiler.d.ts`) does not support `tsconfig.json`'s `include/exclude` field.
So the tsserver will rename symbols in only the opened text document or explicitly dependent text document.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -802,6 +802,78 @@ impl lspower::LanguageServer for LanguageServer {
     }
   }
 
+  async fn rename(
+    &self,
+    params: RenameParams,
+  ) -> LSPResult<Option<WorkspaceEdit>> {
+    if !self.enabled() {
+      return Ok(None);
+    }
+
+    let snapshot = self.snapshot();
+    let specifier =
+      utils::normalize_url(params.text_document_position.text_document.uri);
+
+    let line_index =
+      self
+        .get_line_index(specifier.clone())
+        .await
+        .map_err(|err| {
+          error!("Failed to get line_index {:#?}", err);
+          LSPError::internal_error()
+        })?;
+
+    let req = tsc::RequestMethod::FindRenameLocations((
+      specifier,
+      text::to_char_pos(&line_index, params.text_document_position.position),
+      true,
+      true,
+      false,
+    ));
+
+    let res = self
+      .ts_server
+      .request(snapshot.clone(), req)
+      .await
+      .map_err(|err| {
+        error!("Failed to request to tsserver {:#?}", err);
+        LSPError::invalid_request()
+      })?;
+
+    let maybe_locations = serde_json::from_value::<
+      Option<Vec<tsc::RenameLocation>>,
+    >(res)
+    .map_err(|err| {
+      error!(
+        "Failed to deserialize tsserver response to Vec<RenameLocation> {:#?}",
+        err
+      );
+      LSPError::internal_error()
+    })?;
+
+    match maybe_locations {
+      Some(locations) => {
+        let rename_locations = tsc::RenameLocations { locations };
+        let workpace_edits = rename_locations
+          .into_workspace_edit(
+            snapshot,
+            |s| self.get_line_index(s),
+            &params.new_name,
+          )
+          .await
+          .map_err(|err| {
+            error!(
+              "Failed to convert tsc::RenameLocations to WorkspaceEdit {:#?}",
+              err
+            );
+            LSPError::internal_error()
+          })?;
+        Ok(Some(workpace_edits))
+      }
+      None => Ok(None),
+    }
+  }
+
   async fn request_else(
     &self,
     method: &str,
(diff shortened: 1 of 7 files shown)


cli/lsp/language_server.rs
```diff
@@ -802,6 +802,48 @@ impl lspower::LanguageServer for LanguageServer {
...
+      specifier,
+      text::to_char_pos(&line_index, params.text_document_position.position),
+      true,
+      true,
+      false,
+    ));
+
+    let res = match self.ts_server.request(self.snapshot(), req).await {
+      Ok(res) => res,
+      Err(_) => return Err(LSPError::invalid_request()),
```
lucacasonato, line 826: Instead of swallowing errors, could you log them with `error!` here? Also this might be a bit cleaner if you used `.map_err` instead of `match`.

Same goes for the other places you return errors.

cli/lsp/language_server.rs
```diff
@@ -1143,4 +1185,57 @@ mod tests {
     ]);
     harness.run().await;
   }
+  #[tokio::test]
```
lucacasonato, line 1188: Great test :+1:

cli/lsp/tsc.rs
```diff
@@ -411,6 +413,77 @@ impl QuickInfo {
...
+    for location in self.locations.iter() {
+      // ensure TextDocumentEdit for `location.file_name`.
+      let uri = utils::normalize_file_name(&location.file_name).unwrap();
+      if text_document_edit_map.get(&uri).is_none() {
+        text_document_edit_map.insert(
+          uri.clone(),
+          lsp_types::TextDocumentEdit {
+            text_document: lsp_types::OptionalVersionedTextDocumentIdentifier {
+              uri: uri.clone(),
+              version: None,
```
lucacasonato, line 452: I am not sure, but shouldn't this be the version of the text document the edit happened on? cc @kitsonk

lucacasonato (commented):
Looks very good. Just few comments.

cli/lsp/tsc.rs
```diff
@@ -411,6 +413,77 @@ impl QuickInfo {
...
+}
+
+pub struct RenameLocations {
+  pub locations: Vec<RenameLocation>,
+}
+
+impl RenameLocations {
+  pub async fn into_workspace_edit(
+    self,
+    ls: &LanguageServer,
```
hrsh7th, line 438: The `RenameLocations` can have the locations that is not same to current document so we should accept `LanguageServer` to resolve `line_index` per each locations.

lucacasonato: I think you should be able to use the same approach as `DefinitionInfoAndBoundSpan::to_definition`. The `index_provider` argument there is probably what you want.

hrsh7th: Thank you! I'll fix it.

cli/lsp/language_server.rs
hrsh7th: I'm a rust newbie, so I'll confirm with a code example.
Is the code below ok?

```rust
    let res = self.ts_server.request(self.snapshot(), req).await.map_err(|err| {
      error!("Failed to request to tsserver {}", err);
      return LSPError::invalid_request();
    })?;
```

kitsonk: In Rust, the last statement of a block without a semicolon is an implied return value, so you should do:

```rust
    let res = self.ts_server.request(self.snapshot(), req).await.map_err(|err| {
      error!("Failed to request to tsserver {}", err);
      LSPError::invalid_request()
    })?;
```

cli/lsp/language_server.rs
```diff
@@ -802,6 +802,78 @@ impl lspower::LanguageServer for LanguageServer {
     }
   }
 
+  async fn rename(
+    &self,
+    params: RenameParams,
+  ) -> LSPResult<Option<WorkspaceEdit>> {
+    if !self.enabled() {
+      return Ok(None);
+    }
```
hrsh7th, line 811: I noticed that other all LSP features check it.

cli/lsp/language_server.rs
```diff
@@ -802,6 +802,78 @@ impl lspower::LanguageServer for LanguageServer {
...
+      );
+      LSPError::internal_error()
+    })?;
+
+    match maybe_locations {
+      Some(locations) => {
+        let rename_locations = tsc::RenameLocations { locations };
+        let workpace_edits = rename_locations
+          .into_workspace_edit(
+            snapshot,
```
hrsh7th, line 859: Add `StateSnapshot` here to add document version for WorkspaceEdits.

cli/lsp/tsc.rs
kitsonk: Correct, to ensure there aren't "race" conditions in the editor.

cli/lsp/language_server.rs
hrsh7th: Thank you! I've fixed it.

cli/lsp/language_server.rs
```diff
@@ -802,6 +802,78 @@ impl lspower::LanguageServer for LanguageServer {
...
+          .map_err(|err| {
+            error!(
+              "Failed to convert tsc::RenameLocations to WorkspaceEdit {:#?}",
+              err
+            );
+            LSPError::internal_error()
+          })?;
+        Ok(Some(workpace_edits))
+      }
+      None => Ok(None),
```
hrsh7th, line 873: I noticed that the `textDocument/rename` can return null as a response.

cli/lsp/tsc.rs
hrsh7th: I fixed it with referencing `DefinitionInfoAndBoundSpan::to_definition`.

cli/lsp/tsc.rs
hrsh7th: Thank you for your reviews.

I've added a version from `StateSnapshot`.

kitsonk:
> Currently, the ConfigureRequest (in cli/tsc/compiler.d.ts) does not support tsconfig.json's include/exclude field.
> So the tsserver will rename symbols in only the opened text document or explicitly dependent text document.

It will never support it, so the resulting behaviour is the desired behaviour.

kitsonk (approved):
LGTM, very nice contribution... thank you!

PR #8911, opened 2020-12-29
feat(lsp): add cache command

kitsonk:
This adds a command to allow caching of files in the lsp.  For example in VSCode, a command can be created that will effectively do a `deno cache` on the currently open file, fetching all of its dependencies, including using any import map that is configured for the project.

_Extension_Development_Host__-_mod_ts_—_plugin_test_workspace_and_commands_ts_—_vscode_deno

There are a couple things that need to be done with the PR, before landing but raising for visibility.

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -398,7 +398,7 @@ impl FileFetcher {
       .boxed();
     }
 
-    info!("{} {}", colors::green("Download").to_string(), specifier);
+    info!("{} {}", colors::green("Download"), specifier);
 
     let file_fetcher = self.clone();
     let cached_etag = match self.http_cache.get(specifier.as_url()) {
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -276,7 +276,6 @@ pub struct FileFetcher {
   cache_setting: CacheSetting,
   http_cache: HttpCache,
   http_client: reqwest::Client,
-  use_color: bool,
 }
 
 impl FileFetcher {
@@ -285,7 +284,6 @@ impl FileFetcher {
     cache_setting: CacheSetting,
     allow_remote: bool,
     maybe_ca_file: Option<&str>,
-    use_color: bool,
   ) -> Result<Self, AnyError> {
     Ok(Self {
       allow_remote,
@@ -293,7 +291,6 @@ impl FileFetcher {
       cache_setting,
       http_cache,
       http_client: create_http_client(get_user_agent(), maybe_ca_file)?,
-      use_color,
     })
   }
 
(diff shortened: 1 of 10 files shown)


lucacasonato (commented):
It would be cool if we could integrate this with [`$/progress`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#progress), to show what file we are currently downloading / transpiling, like we do via stderr when you do `deno cache` in the cli. This is definitely something for a follow up though.

Looks good, but I'll give you a proper review in the morning. Too late to think straight :-)

kitsonk:
@lucacasonato the biggest challenge with that is that status of downloading is buried really really deep in the cli... yes, it would be nice, but we really don't bubble that stuff up at the moment, so it is non-trivial.

cli/file_fetcher.rs
```diff
@@ -395,10 +398,18 @@ impl FileFetcher {
           specifier
         ),
       ))
-      .boxed_local();
+      .boxed();
     }
 
-    info!("{} {}", colors::green("Download"), specifier);
+    info!(
```
kitsonk, line 404: I'm really not happy with this, but it was the only thing I could think of to easily fix this.  The debug console in VSCode does not support colours, and this is buried deep in file fetcher and isn't very easy to conditionally suppress... it is useful to see files being downloaded in the debug output for the lsp, so I don't want to totally turn it off... any bright ideas would be appreciated.

hrsh7th: VSCode seems to solve do this by passing a `*ProgressReporter` in such cases.

https://github.com/microsoft/vscode-languageserver-node/blob/17cbbdc9cc278b5bc697a9461b4eb0d4bab5c49f/server/src/common/progress.ts#L102

lucacasonato: @kitsonk Can't we just start lsp with NO_COLOR set from VS Code?

kitsonk: > Can't we just start lsp with NO_COLOR set from VS Code?

Yes, I guess that would work for now.  It is still challenging having logging of info stuff buried so deeply in the cli at times... we really should be figuring out ways to pull this up to the top, so the consumer can determine what is appropriate or not.

I would add it to my cache command PR, but someone already merged it. 🤷

lucacasonato (commented):
I really don't like this whole `use_color` business. Instead of doing that, I think it would make more sense to make the printer configurable (a closure you pass to file fetcher that gets called with the specifier every time we log now). That way we can bind it into `$/progress` in the future too.

kitsonk:
> I really don't like this whole `use_color` business. Instead of doing that, I think it would make more sense to make the printer configurable (a closure you pass to file fetcher that gets called with the specifier every time we log now). That way we can bind it into `$/progress` in the future too.

I thought we just agreed in the comments above that I will start with `NO_COLOR` for now and back out the use color?  We can address the other stuff later, as it is quite invasive.

lucacasonato (approved):
LGTM. Very nice feature

PR #8913, opened 2020-12-29
fix(core/error) Fix incorrect index in Promise.all error reporting

### LINKED ISSUE #8871: Stack frames report wrong promise index
Check out https://github.com/denoland/deno/blob/master/core/error.rs#L295

That probably shouldn't be the property "columnNumber" for promise_index, which is used above, unsurprisingly, for columnNumber.

While I'm here, __callSiteEvals is a Deno thing? I didn't find a reference to it anywhere else online.

duci9y:
fixes #8871

There was a small bug where we were reporting the wrong index for the failing promise in a `Promise.all` call.
The `error_024_stack_promise_all` test was not robust enough to catch this (more details in #8871).

This PR fixes the bug and updates the test so that the failing promise is at index 1 instead of 0.

### DIFF
--- a/cli/tests/error_024_stack_promise_all.ts
+++ b/cli/tests/error_024_stack_promise_all.ts
@@ -1,4 +1,5 @@
 const p = Promise.all([
+  Promise.resolve(),
   (async (): Promise<never> => {
     await Promise.resolve();
     throw new Error("Promise.all()");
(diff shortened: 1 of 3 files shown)


bartlomieju:
@nayeemrmn PTAL :)

nayeemrmn (approved):
LGTM

bartlomieju (approved):
Thanks @duci9y!

PR #8919, opened 2020-12-29
fix(std/http): Don't expose ServerRequest::done as Deferred

nayeemrmn:
This is resolvable by the consumer. Expose it as a `Promise` instead.

### DIFF
--- a/std/http/server.ts
+++ b/std/http/server.ts
@@ -21,9 +21,14 @@ export class ServerRequest {
   conn!: Deno.Conn;
   r!: BufReader;
   w!: BufWriter;
-  done: Deferred<Error | undefined> = deferred();
+  #done: Deferred<Error | undefined> = deferred();
 
   private _contentLength: number | undefined | null = undefined;
+
+  get done(): Promise<Error | undefined> {
+    return this.#done.then((e) => e);
+  }
+
   /**
    * Value of Content-Length header.
    * If null, then content length is invalid or not given (e.g. chunked encoding).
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - thanks @nayeemrmn

zerofairnight:
This is now a problem for Web Sockets, closing the socket will not call done.resolve() and the promise will hang forever.

An example demonstrating the problem:

```ts
import { serve, ServerRequest } from "https://deno.land/std/http/server.ts";
import { acceptWebSocket } from "https://deno.land/std/ws/mod.ts";

const server = serve({ port: 8000 });

async function handler(request: ServerRequest) {
  const { conn, r: bufReader, w: bufWriter, headers } = request;

  const socket = await acceptWebSocket(
    { conn, bufReader, bufWriter, headers },
  );

  await socket.send("server will now close");

  await socket.close();

  // lets pretend the server have done all the work
  server.close(); // this will close all connections but do not call done.resolve()

  // request.done.resolve(); // no longer an option
}

for await (const request of server) {
  handler(request);
}

// the application will end with the error: 'Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promise.'
```

The only way to ensure the promise is resolved would be calling `await request.respond({})` that is totally unnecessary.

We need a way to call `#done.resolve()` manually for other kind of use cases too, specifically server-sent events and other custom chunked stream protocols.

nayeemrmn:
@zerofairnight Can you open a new issue about what's no longer possible for you? I suspect an alternative solution would be better.

PR #8920, opened 2020-12-29
fix: Worker hangs when posting "undefined" as message

yonbav:
Issue: https://github.com/denoland/deno/issues/8702
11_workers.js: catch clause returned on error. That caused web worker to send successfully the event to the host worker, but when the host worker failed to decode the message it threw an exception and returned without parsing the event received  from the web worker.

thats why we received this strange behaviour with undefined, that a post message was sent from webWorker but was never received in the host.


yonbav:
@bartlomieju I think we need to decide how webWorker handle such failures.
It seems that there is a msgErrorEvent that should take place in situations like this that was about to be implemented.
undefined is parsed to empty byte array, I think we can easily parse byte array as undefined, what do you think?

runtime/js/11_workers.js
```diff
@@ -88,7 +91,7 @@
           cancelable: false,
           data,
         });
-        return;
+        throw new Error(`Received data could not be parsed: '${msgData}'.`);
```
bartlomieju, line 94: This is surprising... looks like during some refactors dispatch `messageerror` event got lost :/ if my memory serves right we should throw this error only if `messageerror` event wasn't prevented (ie. `preventDefault()` **hasn't been called** when handling the event).

runtime/js/11_workers.js
```diff
@@ -43,6 +43,9 @@
   }
 
   function decodeMessage(dataIntArray) {
+    if (dataIntArray.length == 0) {
+      return undefined;
+    }
```
bartlomieju, line 48: I don't think this is correct, empty array buffer is a valid message and we shouldn't convert it to `undefined`, it will lead to even more strange problems down the road.

bartlomieju (commented):
@nayeemrmn can you please review this PR as well?

runtime/js/11_workers.js
yonbav: @bartlomieju that makes a lot of sense, but currently it does not know how to decode empty array buffer. so what should be returned when we receive empty array?

current behavior throws an exception.

runtime/js/11_workers.js
yonbav: @bartlomieju I'm not sure I understand. what do you mean by `massageerror `event was prevented? the webWorker can prevent the `messageerror`? 
and what should happen if the messageerror was not prevented? the host should send the `messageerror `back to the webWorker?

runtime/js/11_workers.js
nayeemrmn: Our messaging currently works by having `JSON.stringify()` and `JSON.parse()` on each end, therefore an empty buffer should never happen. However, these `JSON` functions can't be used to transfer `undefined`. Transferring `postMessage(undefined) -> msgEvent.data == undefined` is valid in browsers, which I guess has proper structured cloning. Using empty array values to represent `undefined` is a valid temporary fix IMO. This is implicitly happening on the other end because `encoder.encode(JSON.stringify(undefined)) == new Uint8Array(0)`.

runtime/js/11_workers.js
nayeemrmn: @yonbav I think just revert this change and focus the PR and tests on the behaviour when posting `undefined`.

yonbav: Done

runtime/js/11_workers.js
yonbav: @nayeemrmn I agree. 
@bartlomieju what do you think?

bartlomieju: Sounds like a good temporary solution. @yonbav please add a note above explaining what's going on and mentioning that it's a temporary solution until structured close arrives

bartlomieju (approved):
LGTM, thanks for the fix @yonbav! It's not ideal solution, but it's a good band-aid until we implement structured clone.

PR #8927, opened 2020-12-30
fix(op_crates/fetch): correct regexp for fetch header

Trott:
Fix bug in regular expression and make the regular expression more
strict.

In a string passed to new RegExp(), '[\t\s]' is identical to '[ts]' and
not `/[\t\s]/`. For that, the backslash needs to be escaped in the
string. Futhermore, `\t` is the tab character and is included in the
special regexp value `\s` so is unnecessary.

That would reduce the RegExp to new RegExp(`^${value}\\s*;?`) but
there's no point in matching 0 or more space characters followed by 0 or
one semi-colons as that will match no matter what follows `value`.

To make it more strict, require one of space, semicolon, or
end-of-string after value.

### DIFF
--- a/cli/tests/unit/fetch_test.ts
+++ b/cli/tests/unit/fetch_test.ts
@@ -251,6 +251,27 @@ unitTest(
 
 unitTest(
   { perms: { net: true } },
+  async function fetchMultipartFormBadContentType(): Promise<void> {
+    const response = await fetch(
+      "http://localhost:4545/multipart_form_bad_content_type",
+    );
+    assert(response.body !== null);
+
+    try {
+      const formData = await response.formData();
+      fail(
+        "Response.formData() didn't throw on a incorrect multipart content type",
+      );
+    } catch (e) {
+      assertEquals(e.message, "Invalid form data");
+      assert(e instanceof TypeError);
+    }
+    await response.body.cancel();
+  },
+);
+
+unitTest(
+  { perms: { net: true } },
   async function fetchURLEncodedFormDataSuccess(): Promise<void> {
     const response = await fetch(
       "http://localhost:4545/cli/tests/subdir/form_urlencoded.txt",
--- a/cli/tests/unit/fetch_test.ts
+++ b/cli/tests/unit/fetch_test.ts
@@ -257,15 +257,13 @@ unitTest(
     );
     assert(response.body !== null);
 
-    try {
-      const formData = await response.formData();
-      fail(
-        "Response.formData() didn't throw on a incorrect multipart content type",
-      );
-    } catch (e) {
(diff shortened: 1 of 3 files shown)


Trott:
In order to write a test for this that fails with the current regexp but passes with the one here, I think we'd need a fetch with a form that returns a `Content-type` header like `multipart/form-datassssssstttt`. Perhaps that's an edge case that can be omitted from testing but if not, I'm happy to try to write the test, but I may need a gentle shove in the right direction.

Trott:
For just general confirmation that this change is correct:

```javascript
// Original buggy function
function hasHeaderValueOf(s, value) {
  return new RegExp(`^${value}[\t\s]*;?`).test(s);
}

// Fixed function with this PR's change
function hasHeaderValueOfNew(s, value) {
  return new RegExp(`^${value}(?:[\\s;]|$)`).test(s);
}

console.log(
  hasHeaderValueOf('multipart/form-datatstststs', 'multipart/form-data')
); // Logs true. Oh no!

console.log(
  hasHeaderValueOfNew('multipart/form-datatstststs', 'multipart/form-data')
); // Logs false. Yay!
```

bartlomieju:
> In order to write a test for this that fails with the current regexp but passes with the one here, I think we'd need a fetch with a form that returns a `Content-type` header like `multipart/form-datassssssstttt`. Perhaps that's an edge case that can be omitted from testing but if not, I'm happy to try to write the test, but I may need a gentle shove in the right direction.

Sounds good, you can add a new endpoint for `test_server` (in `test_util/src/lib.rs`) which such content type.

crypt096 (approved):

Trott:
> Sounds good, you can add a new endpoint for `test_server` (in `test_util/src/lib.rs`) which such content type.

OK, test added. Thanks!

Trott:
Whoops, let's try again, but this time after checking in the changes made by the format script....

cli/tests/unit/fetch_test.ts
```diff
@@ -249,6 +249,27 @@ unitTest(
...
+
+    try {
+      const formData = await response.formData();
+      fail(
+        "Response.formData() didn't throw on a incorrect multipart content type",
+      );
+    } catch (e) {
+      assertEquals(e.message, "Invalid form data");
+      assert(e instanceof TypeError);
+    }
```
bartlomieju, line 268: Nitpick: you can use `assertThrowsAsync` instead of manually failing the test

bartlomieju (approved):
@Trott thanks for the PR, looks great!

cli/tests/unit/fetch_test.ts
Trott: Sure, done.

PR #8928, opened 2020-12-30
BREAKING(unstable): remove CreateHttpClientOptions.caFile

### LINKED ISSUE #8837: Remove caFile from Deno.createHttpClient options
`Deno.createHttpClient` now can be configured to use a CA using the `caData` field. They `caFile` field is this not needed anymore.

mrmodise:
Closes #8837


mrmodise:
@lucacasonato @bartlomieju would this change affect passing the cert as a file on the terminal. I notice the tests that are failing all have a similar pattern: ```--cert tls/RootCA.pem```

For example,

```rust
itest!(cafile_info {
  args:
    "info --quiet --cert tls/RootCA.pem https://localhost:5545/cli/tests/cafile_info.ts",
  output: "cafile_info.ts.out",
  http_server: true,
});
```

Any feedback is welcome. Thanks

bartlomieju:
@mrmodise no, `--cert` flag should still work as before

mrmodise:
Thanks @bartlomieju I thought they were linked cause when i checked the `program_state.rs` I see that it accepts the create_http_client() as shown below meaning I would have to write another function to read the cert into a string

```rust
let file_fetcher = FileFetcher::new(
      http_cache,
      cache_usage,
      !flags.no_remote,
      ca_file.as_deref(),
    )?;
```

since removed dealing with a file as shown below:

```rust
if let Some(ca_file) = ca_file {
    let mut buf = Vec::new();
    let cert = reqwest::Certificate::from_pem(&buf)?;
  }
```

bartlomieju:
@mrmodise then I think it's safe to move that bit of code into `FileFetcher` to handle that situation

mrmodise:
@bartlomieju I think its ready for review. I also wanted to say thanks once again for always getting back to my questions. Its really encouraging and inspiring for me as  a new joiner, and, am sure I speak for all new joiners who come to this repository to  add value to Deno in their own ways

cli/program_state.rs
```diff
@@ -58,7 +59,13 @@ impl ProgramState {
     let dir = deno_dir::DenoDir::new(custom_root)?;
     let deps_cache_location = dir.root.join("deps");
     let http_cache = http_cache::HttpCache::new(&deps_cache_location);
-    let ca_file = flags.ca_file.clone().or_else(|| env::var("DENO_CERT").ok());
+    let ca_file_path =
+      flags.ca_file.clone().or_else(|| env::var("DENO_CERT").ok());
+
+    let ca_data: Option<String> = match ca_file_path.as_ref() {
+      None => None,
+      Some(ca_file_path) => Some(read_to_string(ca_file_path).unwrap()),
```
bartlomieju, line 67: This bit will panic if file doesn't exist, use `?` instead of `unwrap()`

bartlomieju (changes requested):
@mrmodise thanks, looks good, just one thing to fix

cli/program_state.rs
mrmodise: Cool thought of it but I was not that sure. I never worked with Rust before this PR but am definitely hooked

bartlomieju (approved):
LGTM

PR #8937, opened 2020-12-30
docs: Update instructions on how to setup development environment with Vim/Neovim

uki00a:
The setup instructions for Vim/Neovim worked with Deno v1.6.0, but not with v1.6.2. :cry: This PR fixes the documentation so that it also works with Deno v1.6.2.


bartlomieju (approved):
Thanks @uki00a

PR #8938, opened 2020-12-30
feat: CLI features don't require --unstable flag

### LINKED ISSUE #8934: Don't require --unstable flag for CLI features
Currently `--unstable` flag is used to enable both unstable runtime API as well as enable new features in CLI. It means that if one wants to use only a single unstable feature of CLI one must opt into all unstable runtime APIs as well.

After discussion with @ry we propose to remove requirement for `--unstable` flag for CLI features and make the flag affect only runtime APIs.

This change would impact following subcommands:
- `deno compile`
- `deno lint`
- `deno test --no-run`
- `deno test --coverage`
- `--import-map` flag in various subcommands
- `--watch` flag in various subcommands

Ref https://github.com/denoland/deno/issues/8129

bartlomieju:
This commit removes requirement for `--unstable` flag
for CLI features. Using unstable `Deno` APIs still requires
passing that flag.

This change impacts following subcommands:
- `deno compile`
- `deno lint`
- `deno test --no-run`
- `deno test --coverage`
- `--import-map` flag in various subcommands
- `--watch` flag in various subcommands

Closes #8934

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1285,7 +1285,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("no-run")
         .long("no-run")
         .help("Cache test modules, but don't run tests")
-        .takes_value(false)
+        .takes_value(false),
     )
     .arg(
       Arg::with_name("fail-fast")
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -341,7 +341,7 @@ fn clap_root<'a, 'b>(version: &'b str) -> App<'a, 'b> {
     .arg(
       Arg::with_name("unstable")
         .long("unstable")
-        .help("Enable unstable features and APIs")
+        .help("Enable unstable APIs")
         .global(true),
     )
     .arg(
@@ -787,7 +787,7 @@ Ignore formatting a file by adding an ignore comment at the top of the file:
         .takes_value(true)
         .use_delimiter(true)
         .require_equals(true)
-        .help("Ignore formatting particular source files. Use with --unstable"),
+        .help("Ignore formatting particular source files."),
     )
     .arg(
       Arg::with_name("files")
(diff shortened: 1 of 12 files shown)


kitsonk (approved):
LGTM

ry:
I'm a little worried about `--import-map`  and `deno test --no-run` (what is its purpose anyway?)

lucacasonato (approved):
There is no distinction that import maps are still unstable now. We either need a new flag for CLI unstable, or add a yellow warning output if this flag is used.

lucacasonato (changes requested):
That was meant to be a request changes. 🤦‍♂️

bartlomieju:
> I'm a little worried about `--import-map` and `deno test --no-run` (what is its purpose anyway?)

`deno test --no-run` collects, caches and type checks all files that will be tested (but doesn't actually run the tests).

> We either need a new flag for CLI unstable, or add a yellow warning output if this flag is used.

Showing a warning is fair, I'd rather not introduce another CLI flag, that's the whole purpose of this PR. I'll update to show the warning.

cli/flags.rs
```diff
@@ -1493,11 +1490,9 @@ fn import_map_arg<'a, 'b>() -> Arg<'a, 'b> {
     .long("import-map")
     .alias("importmap")
     .value_name("FILE")
-    .requires("unstable")
-    .help("UNSTABLE: Load import map file")
+    .help("Load import map file")
     .long_help(
-      "UNSTABLE:
-Load import map file
+      "Load import map file
```
nayeemrmn, line 1495: It should still say "unstable" or similar in the help message.

cli/flags.rs
```diff
@@ -1286,8 +1285,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("no-run")
         .long("no-run")
         .help("Cache test modules, but don't run tests")
```
nayeemrmn, line 1287: Ditto. Especially now since there is no longer a flag requirement.

nayeemrmn (changes requested):
Since there's no longer an enforced flag requirement, it should consistently say `Unstable: ` in the help messages of unstable flags and subcommands.

lucacasonato:
@bartlomieju changed his mind. Closing

PR #8940, opened 2020-12-30
feat(std/node): Add support for process.on("exit") + refactor

Soremwar:
~This PR splits process into a public default-export module and a private destructurable one.~

Right now, loading the Node globals shim (std/node/global.ts) requires you to pass allow-read and allow-env even if you are not using the properties that need the special access.

The node globals shim will now use Proxied properties allowing the module to be imported without early evaluation.


std/node/_process.ts
```diff
@@ -0,0 +1,167 @@
...
+  /** https://nodejs.org/api/process.html#process_process_cwd */
+  cwd: Deno.cwd,
+  /** https://nodejs.org/api/process.html#process_process_exit_code */
+  exit: Deno.exit,
+  /** https://nodejs.org/api/process.html#process_process_env */
+  get env(): { [index: string]: string } {
+    // Getter delegates --allow-env and --allow-read until request
+    // Getter also allows the export Proxy instance to function as intended
+    return Deno.env.toObject();
+  },
```
guybedford, line 72: If the problem is that this getter doesn't support the named exports destructuring, wouldn't the better solution to be a Proxy as you say to rather get the benefit for both cases?

Soremwar: Proxy should be the solution, however a proxy-ed variable still seems to be triggering read/env operations

guybedford: How are you seeing that? The proxy should be based on an empty object base that only populates on the handler calls. I believe it's possible.

Soremwar: Perhaps, but I don't know how that would work on a named import :/

guybedford: It should be as simple as `env: new Proxy(...)` right?

Soremwar: I don't understand what you mean by that

guybedford: Something like the following:

```suggestion
  env: new Proxy(proxy_base, {
    set () {
      initProxy(envBase);
      return Reflect.set(...arguments);
    },
    ownKeys () {
      initProxy(envBase);
      return Reflect.ownKeys(envBase);
    },
    ... for defineProperty, deleteProperty, get, getOwnPropertyDescriptor, 
  }),
```

guybedford: Where the `initEnv` function does `Object.assign(envBase, Deno.env.toObject())` sort of thing.

Soremwar:
This PR now also adds support for `process.on("exit")` by appending a listener to the `unload` event. Luckily, `unload` works pretty much the same as `on("exit")` since it won't schedule any additional work in the even loop either.

std/node/_process.ts
Soremwar: @guybedford After practicing a little with Proxies I think it's safe to say that perhaps the module separation might be the best choice. Even though proxying the properties of the env variables works to prevent an actual permission operation, the real problem comes when you try to read the env property as a whole. Since `envBase` is an empty object that doesn't have the actual env information until it's requested, logging it as a whole property will show just an empty object, deviating from the Node implementation.

guybedford: @Soremwar inspection should result in the `ownKeys` hook as above explicitly providing the full object. The output of console.log is hardly the normative comparison here. Even then it could be made to work with a special log symbol. `JSON.stringify` etc will work fine too.

Soremwar:
This also solves an error in the Node implementation, since `process.argv` didn't contained the main module route as it is supposed to.

std/node/_process.ts
Soremwar: @guybedford I tried that with the `ownKeys` hook, but it doesn't seem to be triggered when you inspect the variable. Am I missing something?

guybedford: Inspection and usage are completely different things. Equality is not based on inspection, and inspection code paths can be adjusted using eg special symbols or any time of local inspection hook to adjust the inspection output.

Otherwise we'd never be able to use proxies at all, just because they log different doesn't mean we can't use them as replacements for other objects.

Soremwar: How would I go about implementing this custom log hook then? Sorry, I'm a little inexperienced in this area

I already have the proxies for argv and env so that would be the last discrepancy that needs to be taken care of

guybedford: @Soremwar the inspection code can be found at https://github.com/denoland/deno/blob/master/runtime/js/02_console.js#L400, where you can follow the proxy path as well. For example, a private `envBase[internal.logSymbol] = function () { return 'custom logger' }` where the `internal.logSymbol` is only accessible to Deno core code only is one approach. There are many ways to do it so you'd have to check the exact approach with another reviewer on that.

Soremwar: Yep, I found the function to custom inspect the variables, so it's all taken care off...sorta

As it turns out the inspect function will swallows any errors so I can't rely on it to test the permissions until #9095 gets merged

Soremwar:
Ready for review

std/node/_process.ts
Soremwar: Nevermind, I ended up solving an unrelated bug, this was totally doable right now

std/node/process_test.ts
```diff
@@ -1,37 +1,11 @@
...
-import * as all from "./process.ts";
-import { argv, env } from "./process.ts";
 import { delay } from "../async/delay.ts";
-import "./global.ts";
-
-// NOTE: Deno.execPath() (and thus process.argv) currently requires --allow-env
-// (Also Deno.env.toObject() (and process.env) requires --allow-env but it's more obvious)
-
-Deno.test({
-  name: "process exports are as they should be",
```
lucacasonato, line 15: Why remove this test?

std/node/process.ts
```diff
@@ -118,53 +285,34 @@ export const process = {
...
+
+export const removeListener = process.removeListener;
+export const removeAllListeners = process.removeAllListeners;
+export const stderr = process.stderr;
+export const stdin = process.stdin;
+export const stdout = process.stdout;
+
+export default process;
+
+//TODO
```
kt3k, line 315: Now we require tags for every TODO comment. ref: https://github.com/denoland/deno/pull/9135

std/node/process.ts
```diff
@@ -1,32 +1,109 @@
-// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
```
kt3k, line 1:
```suggestion
// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```

std/node/process_test.ts
Soremwar: Previously this test was under the assumption the library would only expose things available through the ESM export (for some reason they are different to the items exposed through require), however since most Node packages used in Deno are transforms from CommonJS and modules used through createRequire not exposing this properties makes no sense

This test was used to check that the export from process was indeed different to the built in, but this is no longer the case

std/node/process.ts
Soremwar: Solved

std/node/process.ts
Soremwar: Done

std/node/process.ts
```diff
@@ -1,32 +1,109 @@
...
+  },
+});
 
-/** https://nodejs.org/api/process.html#process_process_version */
-export const version = `v${Deno.version.deno}`;
+/**
+ * https://nodejs.org/api/process.html#process_process_env
+ * Requires env permissions
+ * */
+export const env: { [index: string]: string } = new Proxy(_env, {
```
ry, line 89: `Record<string, string>` ?

Soremwar: Can you commit the change? I'm AFK right now

std/node/process.ts
```diff
@@ -1,32 +1,109 @@
...
 
 /** https://nodejs.org/api/process.html#process_process_arch */
 export const arch = Deno.build.arch;
 
+function getArguments() {
+  return [Deno.execPath(), fromFileUrl(Deno.mainModule), ...Deno.args];
+}
+
+//deno-lint-ignore ban-ts-comment
+//@ts-ignore
```
bartlomieju, line 35: @Soremwar FYI you can remove `deno-lint-ignore` if you add a comment to `// @ts-ignore` on why it's being used

bartlomieju (approved):
LGTM, thank you @Soremwar

Issue #8917, opened 2020-12-29
inaccurate deno test coverage


claudio-felicioli:
I'm getting consistently odd results from` deno test --coverage --unstable`:
- the coverage report line numbers are off by some small number (the offset error is not constant, even when reporting multiple pieces of code from the same source file)
- more importantly, the reported uncovered code lines seems to be the wrong one (reporting code text lines that have been covered, that in the source code are near to rows that are not covered, but unreported)

I started to check coverages at 1.5.4 with these issues, that are still present in 1.6.2

caspervonb:
> the coverage report line numbers are off by some small number

Likely this is just from compilation?

Reports are for the raw JavaScript as v8 saw it.
We don't do source map looks ups (yet, working on that next).

> the reported uncovered code lines seems to be the wrong one (reporting code text lines that have been covered, that in the source code are near to rows that are not covered, but unreported)

So the content is all wrong and there are lines that definitively should be covered?

I've got one known bug here at-least from 1.6 caused by this break https://github.com/denoland/deno/blob/master/cli/tools/coverage.rs#L172, not sure if this is the same cause tho.

kitsonk:
> We don't do source map looks ups (yet, working on that next).

Do we have an issue for this, if not we should for tracking purposes.

caspervonb:
> Do we have an issue for this, if not we should for tracking purposes.

Talked about it being the plan a lot in dicsord but don't believe we have an open issue for it so raised one here 
https://github.com/denoland/deno/issues/8923

claudio-felicioli:
thanks for opening the source map issue :)

claudio-felicioli:
> > the coverage report line numbers are off by some small number
> 
> Likely this is just from compilation?
> 
> Reports are for the raw JavaScript as v8 saw it.
> We don't do source map looks ups (yet, working on that next).
Yes, seems source map

> 
> > the reported uncovered code lines seems to be the wrong one (reporting code text lines that have been covered, that in the source code are near to rows that are not covered, but unreported)
> 
> So the content is all wrong and there are lines that definitively should be covered?
the number of reported uncovered parts of the code is the expected one, just the numbers of the lines and the code displayed is wrong, therefore should be the source map only, at least for the cases that I have tested.

PR #8942, opened 2020-12-31
fix(coverage): merge duplicate reports

caspervonb:
Merging multiple runs isn't quite right because we rely on a 0 count to signal that a block hasn't been called.
Other tools like c8 expect this to be true as-well so we need to do our best to merge coverage files rather than duplicating them.

Closes https://github.com/denoland/deno/issues/8917


cli/tools/coverage.rs
```diff
@@ -237,29 +237,48 @@ fn collect_coverages(dir: &PathBuf) -> Result<Vec<Coverage>, AnyError> {
...
+      .find(|x| x.script_coverage.url == new_coverage.script_coverage.url);
+
+    if let Some(existing_coverage) = existing_coverage {
+      for new_function in new_coverage.script_coverage.functions {
+        let existing_function = existing_coverage
+          .script_coverage
+          .functions
+          .iter_mut()
+          .find(|x| x.function_name == new_function.function_name);
+
```
caspervonb, line 253: If existing_function is still None here after matching by name, maybe try to find any function with a matching ranges[0]?

cli/tools/coverage.rs
```diff
@@ -237,29 +237,48 @@ fn collect_coverages(dir: &PathBuf) -> Result<Vec<Coverage>, AnyError> {
...
+            let existing_range =
+              existing_function.ranges.iter_mut().find(|x| {
+                x.start_offset == new_range.start_offset
+                  && x.end_offset == new_range.end_offset
+              });
+
+            if let Some(existing_range) = existing_range {
+              existing_range.count += new_range.count;
+            } else {
+              existing_function.ranges.push(new_range);
```
caspervonb, line 265: Extend existing overlapping ranges instead of adding new ones?

bartlomieju (approved):
LGTM

Issue #8757, opened 2020-12-14
Build "lite" binary based on "deno_runtime" in CI


bartlomieju:
Now that "deno_runtime" crate has landed we should add new CI stage that would produce "lite" version of binary that could be used for the purposes of "deno compile" subcommand. This lite binary should use only `FsModuleLoader` or `EmbeddedModuleLoader` in case of self-contained binary. It's likely that some code from `cli/standalone.rs` would have to be moved to `runtime/`.

CC @lucacasonato

bartlomieju:
Done in #9041

PR #8944, opened 2020-12-31
feat(cli) - Runtime binary for Deno

yos1p:
`deno_runtime` crate will produce lite binary: deno-rt in target folder (#8757)

CC @bartlomieju @lucacasonato


lucacasonato:
This looks good, but could you split the PR so this PR only adds the `deno-rt` binary? Then we can change how `deno compile` works in a follow up PR.

yos1p:
Okay. I reverted some of the changes in Deno compile. I also move some of the code from `cli/standalone.rs` to `runtime/standalone.rs`. Running from CLI should will call run function from runtime crate now.

bartlomieju (commented):
@yos1p looks good, nice work! 

That said I'm not sure if we should add a binary target to `deno_runtime`, maybe instead we should add a second binary target to `cli`? (I know I suggested to put it in the `runtime` dir in the original issue, but I'm having doubts now). With this change concepts of `standalone` binaries are split into `cli` and `runtime` and I believe it's preferable to have all of that code in `cli`.

@ry WDYT?

bartlomieju:
Also we need to track the size of the "lite" binary in the benchmarks: https://deno.land/benchmarks#executable-size I'll be fine to add that in a separate PR though.

ry (commented):
Looks good - but can we bikeshed on the name of executable a bit? 

Generally we avoid dashes in filenames. "rt" is not very intelligible.

What about denolite ?

bartlomieju:
> Looks good - but can we bike shed on the name of executable a bit?
> 
> Generally we avoid dashes in filenames. "rt" is not very intelligible.
> 
> What about denolite ?

Good point, +1 for `denolite`

lucacasonato:
`denolite` sounds like a lite version of Deno for general use. But it is not. It is purely the runtime for standalone executable. I would prefer `denort` or `denoruntime`.

yos1p:
Thank you for review! Okay. I'll try to move the code and binary to be generated by cli crate. 

> Also we need to track the size of the "lite" binary in the benchmarks: https://deno.land/benchmarks#executable-size I'll be fine to add that in a separate PR though.

The size of "lite" binary is reduced to half of the size of the binary. I will include this in the PR as well.

For the naming, I kind of prefer `denort` instead of `denolite`, but name can be easily changed later. I'll just follow what you guys decide.

cli/standalone.rs
```diff
@@ -1,154 +1,14 @@
...
-    }
-    .boxed_local()
-  }
-}
-
-async fn run(source_code: String, args: Vec<String>) -> Result<(), AnyError> {
-  let mut flags = Flags::default();
-  flags.argv = args[1..].to_vec();
-  // TODO(lucacasonato): remove once you can specify this correctly through embedded metadata
-  flags.unstable = true;
```
yos1p, line 113: @lucacasonato I don't think we need `flags` anymore. In the PR, this is removed, the code will simply pass all the args from the cli. This also allow us not to use flags module in deno runtime binary.

bartlomieju: We shouldn't change that in this PR, especially since @nayeemrmn got https://github.com/denoland/deno/pull/8738 going that would address it fully.

yos1p: It seems this PR blocked by #8738 
I will continue this after that PR is merged.

ry (commented):
Looks good @yos1p. Just one problem: when I run `denort` in without an arguments I get a zero exit code (success) and no output.  I think this is going to be confusing to people who don't know about this. It would be better to print an error message and have a non-zero exit code.

cli/denort.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+#![deny(warnings)]
```
ry, line 3: Remove

ry (commented):
I think we need to have a test of some sort... How will we know if this thing breaks?

yos1p:
> I think we need to have a test of some sort... How will we know if this thing breaks?

I think we can use similar tests like standalone, but this can only be applied when we already implement compile with `denort` instead of `deno`.

cli/denort.rs
```diff
@@ -0,0 +1,27 @@
...
+      "{}: {}",
+      deno_runtime::colors::red_bold("error"),
+      err.to_string()
+    );
+    std::process::exit(1);
+  }
+
+  // TODO (yos1p) Specify better error message
+  eprintln!(
+    "{}: Runtime Error.",
```
yos1p, line 23: @ry Any suggestions for the error message?

bartlomieju: @yos1p this is fine for now, I will review this PR in more detail today/tomorrow and merge it. @lucacasonato and I will look into using this new binary target for standalone executables.

ry: "This executable is used internally by 'deno compile', it is not meant to be invoked directly."

bartlomieju:
@yos1p thank you for the PR, I really appreciate it, but @lucacasonato and I took a deeper look at it and we found out we needed to do a lot more changes. We've opened a new PR at https://github.com/denoland/deno/pull/9041. Closing without merge.

yos1p:
No worries. Looking forward for the new PR!

PR #8947, opened 2020-12-31
feat(lsp): Implement completionItem/resolve

hrsh7th:
Related #8643

This PR aims to support auto-importing and completionItem's documentation.

Currently, I didn't add tests because the test will depend `textDocument/completion` response.

If maintainers need tests, I can add it.

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -14,7 +14,9 @@ use deno_lint::rules;
 use lspower::lsp_types;
 use lspower::lsp_types::Position;
 use lspower::lsp_types::Range;
+use lspower::lsp_types::TextEdit;
 use std::collections::HashMap;
+use std::path::PathBuf;
 use std::rc::Rc;
 
 /// Category of self-generated diagnostic messages (those not coming from)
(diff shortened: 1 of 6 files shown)


cli/lsp/language_server.rs
```diff
@@ -787,28 +787,140 @@ impl lspower::LanguageServer for LanguageServer {
...
+      completion_item_data.position,
+      completion_item_data.entry_name,
+      tsc::FormatCodeSettings {
+        ..Default::default()
+      },
+      completion_item_data.source.clone(),
+      tsc::UserPreferences {
+        // Request the most detailed auto-importing information for later conversion.
+        import_module_specifier_ending: Some(
+          tsc::ImportModuleSpecifierEnding::Js,
```
hrsh7th, line 859: LanguageService will return `**/*.js` if actual file is `**/*.ts` so I add the `analysis:: fix_auto_import_text_edit` that fix auto-importing TextEdit.

cli/lsp/language_server.rs
```diff
@@ -787,28 +787,140 @@ impl lspower::LanguageServer for LanguageServer {
...
+            .map_err(|err| {
+              error!("Failed to get line_index {:#?}", err);
+              LspError::internal_error()
+            })?;
+
+        let mut completion_item = params.clone();
+        completion_item.detail = completion_entry_details.make_detail();
+        completion_item.documentation =
+          completion_entry_details.make_documentation();
+        completion_item.additional_text_edits = completion_entry_details
```
hrsh7th, line 899: Currently, detail/documentation/additional_text_edits can be resolved.

cli/lsp/analysis.rs
```diff
@@ -47,6 +49,83 @@ fn as_lsp_range(range: &deno_lint::diagnostic::Range) -> Range {
...
+
+  lsp_types::TextEdit {
+    range: text_edit.range.clone(),
+    new_text: text_edit.new_text.replace(
+      &import_path_str,
+      &PathBuf::from(&import_path_str)
+        .with_extension(ext)
+        .to_string_lossy()
+        .to_string(),
+    ),
```
hrsh7th, line 125: Probably, we should use swc_ecmascript, etc... but it was difficult for me...

hrsh7th:
If this solution is the wrong way, feel free to close it.

hrsh7th:
I will update the branch to master.

kitsonk:
Sorry for such a long delay in looking at this PR.

There is a lot of infrastructure we need to bring in in order to facilitate module auto-imports, but also registry code completion.  I think we incorporate some of this PR into that work (and will give co-author credit in those changes) but we need to ensure that the changes are well incorporated into the overall structure of the language server for long term maintainability.

So as it stands, I don't think we can land this PR as it stands, but again, thank you very much for your contribution.  Some of this will certainly work its way into the language server over the next couple of weeks.

hrsh7th:
All right. Thank you for looking into it.

I'm looking forward to the auto-importing with deno-lsp.

PR #8948, opened 2020-12-31
fix(std/node): Add ascii support for Buffer

Soremwar:
Towards merging https://github.com/nodejs/readable-stream/pull/452

### DIFF
--- a/std/node/buffer.ts
+++ b/std/node/buffer.ts
@@ -10,23 +10,14 @@ const notImplementedEncodings = [
   "utf16le",
 ];
 
-//TODO
-//Improve
 function decodeAscii(buffer: Uint8Array) {
-  let ret = "";
-
-  for (let i = 0; i < buffer.length; ++i) {
-    ret += String.fromCharCode(buffer[i] & 0x7F);
-  }
-  return ret;
+  return Array.from(buffer).map((byte) => String.fromCharCode(byte & 0x7F)).join("");
 }
 
 function encodeAscii(text: string) {
-  var chars = [];
-  for (let i = 0; i < text.length; ++i) {
-    chars.push(text.charCodeAt(i));
-  }
-  return new Uint8Array(chars);
+  return new Uint8Array(
+    Object.keys(text).map((i) => text.charCodeAt(Number(i))),
+  );
 }
 
 function checkEncoding(encoding = "utf8", strict = true): Encodings {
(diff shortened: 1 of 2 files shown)


std/node/buffer.ts
```diff
@@ -4,13 +4,22 @@ import * as base64 from "../encoding/base64.ts";
...
 const notImplementedEncodings = [
-  "ascii",
   "binary",
   "latin1",
   "ucs2",
   "utf16le",
 ];
 
+function decodeAscii(buffer: Uint8Array) {
+  return Array.from(buffer).map((byte) => String.fromCharCode(byte & 0x7F)).join("");
```
benjamingr, line 14: Any reason this these aren't done with `TextEncoder` / `TextDecoder` ?

benjamingr: Or encoding/ascii directly if that's an issue

Soremwar: AFAIK TextEncoder is UTF8 only

std/node/buffer.ts
```diff
@@ -4,13 +4,22 @@ import * as base64 from "../encoding/base64.ts";
...
+}
+
+function encodeAscii(text: string) {
+  return new Uint8Array(
+    Object.keys(text).map((i) => text.charCodeAt(Number(i))),
```
bnoordhuis, line 19: If performance matters here: there's probably a steep performance cliff for longer strings.

V8 has a small-numbers-as-strings cache that's not effective for longer strings and I doubt V8 can figure out it doesn't actually need to materialize the indexes as strings.

std/node/buffer.ts
```diff
@@ -431,13 +447,26 @@ export class Buffer extends Uint8Array {
    * encoding. start and end may be passed to decode only a subset of buf.
    */
   toString(encoding = "utf8", start = 0, end = this.length): string {
+    if (start > this.length || end <= 0 || end <= start) {
+      return "";
+    }
+    //Prevent out of bounds
+    start = start < 0 ? 0 : start;
+    end = end > this.length ? this.length : end;
```
bnoordhuis, line 455: Don't know if it matters but these let through NaNs.

bnoordhuis (approved):
LGTM, thanks!

std/node/buffer.ts
benjamingr: What about encoding/ascii in deno's code? I'm pretty sure Deno has ASCII encoding/decoding support (I vaguely recall using it)

bartlomieju:

Issue #5385, opened 2020-05-14
std/node: add net module


EvanHahn:
For completeness, we should add a polyfill for [Node's net module](https://nodejs.org/api/net.html).

ebebbington:
Is it acceptable for a std/node polyfill to only implement part of the module? eg just `net.Socket` is implemented and there's a notice in the readme saying "Only `.Socket` is implemented at this moment in time"

I might try work on this if no one else is?

guybedford:
@ebebbington yes it's absolutely fine to just create a partial implementation and this is what is done for the browser and in Deno already for many cases, often with just a stub function with a todo marked. A contribution would be most welcome.

ebebbington:
Alright i'll try work on it when i can, but i'll note that if anyone else is wishing to tackle this issue, feel free as i don't have all the spare time in the world just lying around :P (ie it will take some time)

guybedford:
This is open source, that is assumed :)

PR #8949, opened 2020-12-31
[WIP] feat(std): Create `net` and `net.Socket` Node Polyfill

ebebbington:
Partially address #5385 

Only adds the `net.Socket` part of the API, but creates the top level `net` API

**Checklist**
- [ ] Complete TS conversion of `net.ts`
- [ ] Add tests
- [ ] Update std/node readme to mention `net` has been created, but only supports `net.Socket`

**Resources**
- [Node's `net.js`](https://github.com/nodejs/node/blob/master/lib/net.js)

### DIFF
--- a/std/node/net.ts
+++ b/std/node/net.ts
@@ -1,0 +1,1104 @@
+// Copyright Joyent, Inc. and other Node contributors.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to permit
+// persons to whom the Software is furnished to do so, subject to the
+// following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+'use strict';
+
+import EventEmitter from "./events.ts";
+import { Stream as stream } from "./stream.ts"
+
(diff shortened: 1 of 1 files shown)


ebebbington:
The `net` module from Node uses a lot of `internalBinding()`'s, so i'm not even sure if this is going to be portable.

I will stop working on this for now until i get back a response

jopemachine:
I'm not sure this could be possible, but I think the `net` module is necessary to implement other node polyfills.

This is a part of the example implementation of `Socket`.

```ts
class Socket extends stream.Duplex {
  constructor(options: any) {
    super();
  }

  private connecting: boolean = false;
  public pending: boolean = false;
  public readyState: boolean = false;
  public bufferSize: number = 0;

  pause = () => {
    // It seems to pause related functions not exists in Deno.Writer
    notImplemented();
  };

  resume = () => {
    // It seems to pause related functions not exists in Deno.Writer
    notImplemented();
  };

  public setNoDelay(enable: boolean) {
    // It seems to setNoDelay related functions not exists in this.conn
    notImplemented();
  }

  public setKeepAlive(setting: any, msecs: any) {
    // It seems to setKeepAlive related functions not exists in this.conn
    notImplemented();
  }

  public address() {
    return this.conn!.localAddr;
  }

  public remoteAddress() {
    return this.conn!.remoteAddr;
  }

  public remoteFamily() {
    // It seems to connection family related field not exists in this.conn
  }

  public remotePort() {}

  public destorySoon() {
    if (this.writable) this.end();

    if (this.writableFinished) this.destroy();
    else this.once("finish", this.destroy);
  }

  public bytesWritten() {}

  public conn: Deno.Conn | undefined = undefined;

  public connect(
    options: TcpSocketConnectOpts | IpcSocketConnectOpts,
    connectionListener?: () => void
  ) {
    let connectionPromise: Promise<Deno.Conn>;

    if ("port" in options) {
      // tcp
      options = options as TcpSocketConnectOpts;
      connectionPromise = Deno.connect({
        port: options.port,
        transport: "tcp",
      });
    } else {
      // ipc (not support windows using named pipe)
      options = options as IpcSocketConnectOpts;
      connectionPromise = Deno.connect({
        path: options.path,
        transport: "unix",
      });
    }

    let cb = (con: Deno.Conn) => {
      this.conn = con;
    };

    cb = cb.bind(this);

    connectionPromise.then(cb);
  }

}
```

And in the above code, I think `this.conn` (the return type of `Deno.connect`) corresponds to `handle` of nodejs.

And I think maybe in `Server` class, return type of `Deno.listen` would corresponds to `handle` of nodejs.

<hr />

`handle` of nodejs is `Pipe` or `TCP` instance created in the below function. 

(https://github.com/nodejs/node/blob/master/lib/net.js#L142)

This is the internal binding of `tcp_wrap`, `pipe_wrap`.

<hr />

And according to what I've seen, some method or field in `handle` of nodejs seems to not exist on Deno side.

So, I think we need to migrate these sorts of functions to the rust side first.

Please ignore this comment if my investigation was insufficient and there are some functions could perform similar feature in Deno.

ebebbington:
@jopemachine Definitely a good and useful writeup - so maybe a lot of the functionality just throws a not implemented. This gives me hope, but then i am worried that the Socket API will actually become unusable (thus like i said above, unableto be developed) as even the basic functionality uses internal bindings

bartlomieju:

PR #8950, opened 2020-12-31
refactor(std/node): Make Buffer a third party library

Soremwar:
Alternative to #8948

This replaces the current implementation of Buffer and puts a reexport of a third-party implementation in the place where it used to be. This enables all the missing features from Buffer and fixes some bugs that seem to be prevalent in the previous implementation (I had to rewrite the tests for string_decoder, since after comparing with the Node implementation they were obviously wrong)


std/node/string_decoder_test.ts
```diff
@@ -66,16 +66,16 @@ Deno.test({
...
 
     decoder = new StringDecoder("base64");
-    assertEquals(decoder.write(Buffer.from("F1", "hex")), "8Q==");
+    assertEquals(decoder.write(Buffer.from("F1", "hex")), "");
     assertEquals(decoder.write(Buffer.from("41F2", "hex")), "8UHy");
     assertEquals(decoder.end(), "");
 
     decoder = new StringDecoder("base64");
-    assertEquals(decoder.text(Buffer.from([0x41]), 2), "QQ==");
+    assertEquals(decoder.text(Buffer.from([0x41]), 2), "");
```
bnoordhuis, line 78: No opinion on whether it's a good idea to switch implementations but I can confirm that these are indeed the expected outputs.

std/node/buffer.ts
```diff
@@ -1,601 +1,7 @@
...
-}
-
-export default { Buffer };
+export {
+  default,
+  Buffer,
+  kMaxLength,
+  kStringMaxLength,
+  SlowBuffer,
+} from "https://esm.sh/safe-buffer@5.2.0";
```
caspervonb, line 7: Relying on a third party domain and service here is definitively bit of a no-no from my perspective since std is supposed to be secure and vetted. This extends the trust to an outside source.

Any particular reason we can't just copy this over?

std/node/string_decoder_test.ts
Soremwar: Blame it on me, I wrote the original tests :P

lucacasonato (changes requested):
Importing from an external source in std is not allowed per our style guide (https://deno.land/manual@v1.6.3/contributing/style_guide#do-not-depend-on-external-code). If we want to use this implementation, we should copy it over.

Soremwar:
@lucacasonato What if I wrote an script that copied over the contents of the module and wrote them to a file in std? This would work in this case and would be a way to implement https://github.com/nodejs/readable-stream/pull/452

PR #8951, opened 2020-12-31
fix(test_util): make hyper server single-threaded

bnoordhuis:
Use Tokio's single-threaded scheduler. The hyper server is used as a
point of comparison for the (single-threaded!) benchmarks in cli/bench.

We're not comparing apples to apples if we use the default
multi-threaded scheduler.

This drops the requests/sec from 284k to 130k on my 12 core Ryzen 5
system. That still leaves a 50k gap for us to close. Working on it!

### DIFF
--- a/test_util/src/lib.rs
+++ b/test_util/src/lib.rs
@@ -722,7 +722,10 @@ async fn wrap_main_https_server() {
   }
 }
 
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM

PR #8953, opened 2021-01-01
fix(lsp): Documents with MBC are not handled properly

### LINKED ISSUE #8753: Documents with MBC are not handled properly in lsp
Editing, or creating a document with the following text causes all sorts of issues and lsp server crashes:

```ts
const foo = `编写软件很难`;

console.log(foo);
```

For example, changing `.log` to `.warn` will result in bad diagnostics like:

```
Unterminated template literal.
```

This is because characters are not being counted the same between what is being sent by the lsp and how the lsp internally calculates them, causing the internal file to be updated in the wrong location.

uki00a:
This PR resolves #8753. The cause of that issue was that the `Range` returned by [`to_range()`](https://github.com/denoland/deno/blob/012f99bd9a0463813653bf8438e09b3247395984/cli/lsp/text.rs#L48-L54) function was computed based on character boundaries instead of byte boundaries, which could cause `String.replace_range()` to panic.

In this PR, I've changed `to_range()` function to return `Range` based on byte boundaries.


kitsonk:
@uki00a I am not sure this solves the problem...  I used this branch and tried this code:

```
const foo = `编写软件很难`;

console.warn(foo);
```

I then deleted the `难` character in the string literal, and then pressed CMD-Z to undo the edit and got:

```
thread 'main' panicked at 'attempt to subtract with overflow', cli/lsp/text.rs:67:10
```

Also, there are then the MBC that aren't treated as single chars in JavaScript, and therefore their edits come across strange as well, for example:

```
const foo = `编写软件很难`;

const bar = `😜🇺🇸👍`;

console.warn(foo, bar);
```

Editing past the emoji causes a similar panic as well.  That is because in JavaScript the `"😜🇺🇸👍".length` results in `8` (while `"编写软件很难".length` results in `6` as one would expect).

bartlomieju:
I suggest to look at [`SourceMap`](https://docs.rs/swc_common/0.10.8/swc_common/struct.SourceMap.html) in SWC to solve the problem from this PR, it seems to have all required utilities already and we use it for similar purposes in `deno_lint`.

kitsonk:
I had some side conversations with @lucacasonato and some of the abstractions that we have which were based on the abstractions in `rust_analyzer` don't really apply well for the lsp.  I have wanted to revisit how we manage sources all together and hopefully address the MBC problem as well, which would look to the swc SourceMap.

There is a lot of context to try to communicate to allow someone else to do the work, but if you wanted to do it @uki00a we could try to have a chat and talk about the design for it, otherwise I have no problem just working on it in the near future.

uki00a:
@kitsonk I'm sorry for the late reply. I hope that this problem will be resolved. :slightly_smiling_face: However, I'll not be obsessed with resolving this issue myself if there are other, better solutions. :wink:
Personally, I think #9018 should take precedence, so I'll close this PR, what do you think?

kitsonk:
@uki00a that issue will be my focus next week, so thank you very much for trying to help. Hopefully when I raise that PR you can give it a review and try and see if you can break it. 😄 but yes, let's close this one for now.

Issue #8956, opened 2021-01-01
setInterval does not type check the callback parameter


00ff0000red:
Deno 1.6.3

Code:
```js
setInterval();
```

Stack trace:
```
Error: Uncaught TypeError: Cannot read property 'bind' of undefined
    at setTimer (deno:runtime/js/11_timers.js:466:25)
    at setInterval (deno:runtime/js/11_timers.js:515:12)
    at ...
```

After checking the source, it seems that the implementation does not type-check the parameter `cb` before calling `.bind` on it.

Yet, running that code on Chromium just returns an interval ID and never throws, so someone might want to check [the spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html) for reference.

Chrome does the same for numbers, null, booleans, etc, except strings, which it evaluates as code.

00ff0000red:
Ah, figured it out, if the argument is not a function, it is stringified via a call to `String`, regardless of whether or not it throws.
Updating runtime/js/timers right now.

benjamingr:
> it is stringified via a call to String

😮  why? I don't think the timer spec requires implementing this legacy behaviour IIRC

00ff0000red:
> > it is stringified via a call to String
> 
> 😮 why? I don't think the timer spec requires implementing this legacy behaviour IIRC

Wait really? I almost submitted a PR *adding* the functionality.

If anyone can find it in the spec, I will avoid implementing it. I hate `eval`, so please give me a reason not to do it :)

https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html

PR #8957, opened 2021-01-02
fix(runtime/js/11_timers): stronger user input checking for setInterval and setTimeout; add function overload

00ff0000red:
Fixes https://github.com/denoland/deno/issues/8956

setInterval and setTimeout both currently make the following assumptions:
* the callback is a function
* the callback has a `.bind` method

Alternative prototype functions may not have the `Function#bind` method.

The [HTML timers spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html) gives the following function overloads:
```
handle = self . setTimeout( handler [, timeout [, arguments... ] ] )
handle = self . setTimeout( code [, timeout ] )
self . clearTimeout( handle )
handle = self . setInterval( handler [, timeout [, arguments... ] ] )
handle = self . setInterval( code [, timeout ] )
self . clearInterval( handle )
```

I've added the overloads that accept `code` as an argument.
The performance should be worse though, as it type-checks every call now.

There was another bug just waiting for someone to cause.

The inputs are checked for being a BigInt, not a number.
That really suggests that someone needs to check all of the current builtins for robustness against bad inputs.

Bad input:
```js
try {
	// no-op
	const nop = Function.prototype;

	// neither a number or bigint
	const ms = {
		[Symbol.toPrimitive]() { throw new Error; }
	};

	// toPrimitive is called
	setTimeout(nop, ms);
} catch(e) {
	console.error(e);
}
```

This threw here:
> ```
> if (delay > TIMEOUT_MAX) {
> ```

Meaning that the bad input had went clear through the bigint check.

I just coerce the input to a u32, and let the bigint exception propagate from there.

Haven't set up any tests yet.


ry (commented):
Thanks - looks good. But can you add a unit test for what this fixes? Probably it should go in `cli/tests/unit/timers_test.ts`

ry (approved):
LGTM - thanks!

PR #8960, opened 2021-01-02
fix(std): Don't use JSDoc syntax for browser-compatibility headers

nayeemrmn:
A JSDoc comment in that position becomes the doc of the symbol below it. Some of these are publicly exported.


ry (approved):
thx

Issue #7259, opened 2020-08-29
WebAssembly.instantiateStreaming does not work with Response


hazae41:
https://github.com/hazae41/denoflate/issues/1

Since Deno 1.3.1, WebAssembly.instanciateStreaming throws the following error when module is an instance of Response (from fetch)

> error: Uncaught TypeError: WebAssembly.compile(): Argument 0 must be a buffer source

lucacasonato:
@caspervonb do we have a test for this?

caspervonb:
> @caspervonb do we have a test for this?

Hmm, apparently I used a buffer source in the test when I enabled that flag 🤔

hazae41:
@caspervonb Do we need to temporarily disable instantiateStreaming? Or implement it with Response?

Because modules using Response are broken since they check for the existence of instantiateStreaming before 

```typescript
if (WebAssembly.instantiateStreaming) {
  // use instantiateStreaming()
} else {
  // get the buffer then use instantiate()
}
```

caspervonb:
> Because modules using Response are broken since they check for the existence of instantiateStreaming before

Oh right, that's bad.. uhm, I'll get back to you.

dsherret:
I'm able to make this work by doing `WebAssembly.instantiateStreaming(response.then(r => r.arrayBuffer()) as any)`

00ff0000red:
Ahh, saw that Deno had WebAssenbly.{compileStreaming, instantiateStreaming}, and thought that it would be fun to try to instantiate ~6 wasm files and set up a project. Then some particularly misleading errors appeared :)

In my opinion, If it doesn't work correctly, it should not be exposed at all. Can we get the PR that caused this to be reverted until this is properly sorted out?

vovacodes:
Hey folks, is there any progress here? I'd like to use Deno for testing my wasm compiled from Rust with `wasm-bindgen` but unfortunately, due to this issue Deno chokes on it. I don't control the JS wrapper that calls to `WebAssembly.instantiateStreaming` (that one is generated by `wasm-pack`) and even though it does feature detection, that one is broken too (https://github.com/denoland/deno/pull/8968), so it's kinda a blocker.

lucacasonato:
Temporary workaround could be to call `delete WebAssembly.instantiateStreaming;` before instantiating the `wasm-pack` code.

PR #7043, opened 2020-08-13
feat(cli): enable WebAssembly.instantiateStreaming

caspervonb:
This exposes the WebAssembly.instantiateStreaming and WebAssembly.compileStreaming functions to the main and worker isolates.

Closes https://github.com/denoland/deno/issues/6154

PR #8968, opened 2021-01-03
Revert "feat(cli): enable WebAssembly.instantiateStreaming"

00ff0000red:
Fixes issue https://github.com/denoland/deno/issues/7259

V8 implemented functions WebAssembly.{compileStreaming, instantiateStreaming} were exposed in PR https://github.com/denoland/deno/pull/7043.

I am unaware of why, but they seem to act more aliases to their non-streaming counterparts, that is, they throw errors on receiving a `Promise<Response>`.

It is better that the functions are not exposed at all than exposed while broken, making feature detection much easier.

This issue still needs to be "properly" fixed, this is merely a temporary patch that reverts the referenced PR.


lucacasonato:
cc @caspervonb

00ff0000red:
Where's /core/modules.rs? Has it been renamed since?

caspervonb:
So we should be able to fix this properly now thanks to https://github.com/denoland/rusty_v8/pull/560 but SGTM in the meantime.

00ff0000red:
> So we should be able to fix this properly now thanks to [denoland/rusty_v8#560](https://github.com/denoland/rusty_v8/pull/560) but SGTM in the meantime.

@caspervonb 
Was that just a prerequisite to a "proper" fix, or is that the fix itself? I presume the former, as it has already been merged?

caspervonb:
> Was that just a prerequisite to a "proper" fix, or is that the fix itself? I presume the former, as it has already been merged?

It's a prerequisite, exposes the callback hooks we need to implement in the runtime.

00ff0000red:
> It's a prerequisite, exposes the callback hooks we need to implement in the runtime.

In that case, we just need to know where /core/modules.rs is, and/or what happened to it since your PR, before we can safely merge this.

caspervonb:
Looking for this line https://github.com/denoland/deno/blob/master/core/runtime.rs#L159?

caspervonb:
Also since you removed wasm_streaming.js, you'd want to remove wasm_streaming.out as-well.

00ff0000red:
> Also since you removed wasm_streaming.js, you'd want to remove wasm_streaming.out as-well.

Well... I swore that I did... so now I'm wondering what branch that I did it in... :/
Otherwise, done.
This seems to be failing because Deno's CI as a whole is already failing, I believe it's unrelated to these changes.

caspervonb (approved):
LGTM, we just need to get the CI sorted.

bartlomieju:
I'm not in favor of landing this PR. Required infrastructure to make WASM streaming work already landed in `rusty_v8` (https://github.com/denoland/rusty_v8/pull/560) and will be available in the next release which will most likely come before Deno 1.7.0. 

I'd rather wait for rusty_v8 release and finish this functionality in Deno rather than removing for a few days.

00ff0000red:
> I'm not in favor of landing this PR. Required infrastructure to make WASM streaming work already landed in `rusty_v8` ([denoland/rusty_v8#560](https://github.com/denoland/rusty_v8/pull/560)) and will be available in the next release which will most likely come before Deno 1.7.0.
> 
> I'd rather wait for rusty_v8 release and finish this functionality in Deno rather than removing for a few days.

Well, it would still take time to actually set it up in Deno and get it working. Then there are the tests that need to be set up for insurance it is working correctly, etc.

But, *if anything*, some of the current files that I am suggesting to be reverted/deleted have code that is completely wrong, ex: cli/tests/wasm_streaming.js, which tested using an ArrayBuffer/ArrayBufferView instead of a Request, therefore proving that the implementation was incorrect.

It is unfortunate that this wasn't removed sooner, but if you really believe that a working implementation will be shipped so soon, I could just close this PR and wait it out.

00ff0000red:
CI is failing because of a file "/tmp/.tmpgX8Z45/badly_formatted.js" not being formatted correctly, what's up with that?

caspervonb:
> CI is failing because of a file "/tmp/.tmpgX8Z45/badly_formatted.js" not being formatted correctly, what's up with that?

We had a rogue test that cascaded, should be resolved with the latest master.

00ff0000red:
> Annoying bug, I know because you can't feature test for it the way you'd do in browsers. It'll just break.

Yes, that's the core problem here, if it's there *and* `Function#toString` shows a native function, then anyone would assume that it is *the* function and that it works. Now we need to run a sort of test suite to make sure that our environment correctly implements the stuff, and that is arguably much worse than just simply not being there.
Now, I don't blame you for mistaking a flag's purpose, but it should've been reverted sooner.

> (nit/fyil it's a response object by the way, not a request)

Technically, it's more like `Promise<Response> | Response`

But no doubt, it will be awesome to finally have the working streaming functions.

In the meantime, do we have any proper test cases for the actual functionality?

00ff0000red:
@bartlomieju If I may ask, how has progress been on implementing steaming compilation in rusty_v8?

bartlomieju:
> @bartlomieju If I may ask, how has progress been on implementing steaming compilation in rusty_v8?

It's being held by upstream problem in V8. It should be released in `1.7.x`

00ff0000red:
Won't that take some time? Are you still confident that reverting this is the wrong decision until then?

When it is complete, this will have to be reverted/rewritten anyway.

bartlomieju:
> Won't that take some time? Are you still confident that reverting this is the wrong decision until then?
> 
> When it is complete, this will have to be reverted/rewritten anyway.

Yes, let's land it

caspervonb:
SGTM; at-least isomorphic code can test for it.

00ff0000red:
Oh, that is *scary*, WPT Wasm tests were passing... for all of the wrong reasons...

caspervonb:
> Oh, that is _scary_, WPT Wasm tests were passing... for all of the wrong reasons...
> I guess we could say that they were false positives?

Yeah streaming tests shouldn't even be enabled right now.
It's fine to disable any tests in wpt/wasi caused by this change 👍

cli/tests/wpt.jsonc
```diff
@@ -243,13 +243,13 @@
     "jsapi/table/get-set",
     "jsapi/table/toString",
     "webapi/body",
```
caspervonb, line 245: Body should never have been enabled either.

00ff0000red: I'm actually unfamiliar with the WPT tests in general, where can I find the source of each individual WPT test?

00ff0000red:
There was only one test that failed as a result of this, it was false positive, but now there's something else,
https://github.com/00ff0000red/deno/blob/f02ac95b16c99e5ebf744d1574718e306c109874/test_util/src/lib.rs#L1229-L1232

CI stdout:
```
OUTPUT
EXPECTED
[WILDCARD]
error: Found 5 not formatted files in [WILDCARD] files

EXPECTED
```

I think it's saying that [this](https://github.com/00ff0000red/deno/blob/patch-3/cli/tests/inline_js_source_map_2_with_inline_contents.js.out) file isn't matching the output [here](https://github.com/00ff0000red/deno/blob/patch-3/cli/tests/inline_js_source_map_2_with_inline_contents.js)?

cli/tests/wpt.jsonc
caspervonb: Upstream is web_platform_tests/wpt
Submodule is at test_util/wpt

caspervonb:
> There was only one test that failed as a result of this, it was false positive, but now there's something else, over here https://github.com/00ff0000red/deno/blob/patch-3/test_util/src/lib.rs#L1229-L1232
> 
> CI stdout:
> 
> ```
> OUTPUT
> EXPECTED
> [WILDCARD]
> error: Found 5 not formatted files in [WILDCARD] files
> 
> EXPECTED
> ```
> 
> I think it's saying that [this](https://github.com/00ff0000red/deno/blob/patch-3/cli/tests/inline_js_source_map_2_with_inline_contents.js) file isn't matching the output [here](https://github.com/00ff0000red/deno/blob/patch-3/cli/tests/inline_js_source_map_2_with_inline_contents.js.out)?

Our test results can be hard to grep sometimes; some things are printed (should be redirected, working on finding all those) that are tests but look like errors.

But a quick glance at the most recent run:

```
test Response with status 0: compileStreaming ... failed
4383
test Response with status 300: compileStreaming ... failed
4384
test Response with status 400: compileStreaming ... failed
4385
test Response with status 404: compileStreaming ... failed
4386
test Response with status 500: compileStreaming ... failed
4387
test Response with status 600: compileStreaming ... failed
4388
test Response with status 700: compileStreaming ... failed
4389
test Response with status 999: compileStreaming ... failed
4390
test Response with status 0: instantiateStreaming ... failed
4391
test Response with status 300: instantiateStreaming ... failed
4392
test Response with status 400: instantiateStreaming ... failed
4393
test Response with status 404: instantiateStreaming ... failed
4394
test Response with status 500: instantiateStreaming ... failed
4395
test Response with status 600: instantiateStreaming ... failed
4396
test Response with status 700: instantiateStreaming ... failed
4397
test Response with status 999: instantiateStreaming ... failed
```

cli/tests/wpt.jsonc
00ff0000red: @caspervonb I'm not sure I understood what you had said, where are those directories located?

00ff0000red:
I presume that those compileStreaming/instantiateStreaming are WPT tests. I'm about to just disable all of the Wasm WPT tests.

00ff0000red:
Nope it was unrelated to WPT. I still can't tell which test is failing, but I don't think that it was failing when I initially opened this PR, so something was added since that used one these functions.

cli/tests/wpt.jsonc
```diff
@@ -242,14 +242,14 @@
     "jsapi/module/toString",
     "jsapi/table/get-set",
     "jsapi/table/toString",
-    "webapi/body",
-    "webapi/invalid-args",
     "webapi/rejected-arg",
     "webapi/status",
```
caspervonb, line 246: This one too

caspervonb:
> at Test.<anonymous> (file:///home/runner/work/deno/deno/test_util/wpt/wasm/webapi/rejected-arg.any.js:7:1)

00ff0000red:
Removed the Wasm WPT, when I've figured out the current problem, I'll re-run the WPT tests locally to see which ones are meant to pass and not meant to pass.

Otherise, I see three CI errors:
> ```
> error: test failed, to rerun pass '-p deno --test integration_tests'
> ```


> ```
> Error: thread 'fmt_check_tests_dir' panicked at 'pattern match failed', test_util/src/lib.rs:1232:7
> ```


> ```
> failures:
>     fmt_check_tests_dir
> 
> test result: FAILED. 348 passed; 1 failed; 16 ignored; 0 measured; 0 filtered out
> 
> Error: Process completed with exit code 101.
> ```

00ff0000red:
It's been some time since this was opened, and I never pushed to getting it merged; at this point, it's probably best to leave it to getting the implement to replace the broken one.

I'll close this to clear my PR queue, and Deno's PR queue.

If a PR is raised to implement these, I'd like to follow it!

PR #8974, opened 2021-01-03
fix(runtime/websocket): respond to ping with pong

### LINKED ISSUE #8972: Websocket API dying on Deno but not on Chrome
There are two different streams im trying connecting my client using the standard [WebSocket API](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) (I have no libraries imported).

In Deno the connections **always die after a couple of minutes**... in Chrome Browser they dont. Using the exact same code. I have not tested in a third web stream but I believe its no coincidence.

Testing code:

> let connection = new WebSocket('wss://stream.binance.com:9443/ws/btcbrl@trade');
> connection.onmessage = (message) => console.log(new Date(), 'msg received');
> setInterval(() => console.log('idle'), 30000); // So Deno dont closes?

In one manual they mention to kill any connections that dont answer pings for **3 minutes**, hypothesis one.

I have noticed if you dont 'setInterval' or wait for something deno closes the application after some time without provocation... maybe its not a ping issue but Deno itself doing something to the websocket (while the process is still running and the socket receiving msgs) believing its idle? hypothesis two

crowlKats:
Closes #8972

### DIFF
--- a/runtime/js/27_websocket.js
+++ b/runtime/js/27_websocket.js
@@ -167,6 +167,7 @@
         this.#bufferedAmount += ta.size;
         core.jsonOpAsync("op_ws_send", {
           rid: this.#rid,
+          kind: "binary",
         }, ta).then(() => {
           this.#bufferedAmount -= ta.size;
         });
@@ -193,6 +194,7 @@
         this.#bufferedAmount += d.size;
         core.jsonOpAsync("op_ws_send", {
           rid: this.#rid,
+          kind: "text",
           text: string,
         }).then(() => {
           this.#bufferedAmount -= d.size;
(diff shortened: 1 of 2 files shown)


runtime/js/27_websocket.js
```diff
@@ -265,6 +267,13 @@
           event.target = this;
           this.dispatchEvent(event);
 
+          this.#eventLoop();
+        } else if (message.type === "ping") {
```
lucacasonato, line 271: This is ok for a first pass, but I think it makes sense to move this logic to Rust.

crowlKats: I do plan a somewhat rewrite of websocket (mostly JS things) together with moving it to its own cratet

elizeuangelo:
I believe this might be related to issue #8972

lucacasonato (approved):
LGTM. Thanks @crowlKats

lucacasonato:
Actually could we add a test for this?

lucacasonato:
Discussed offline with @crowlKats, and tests will be added in a follow up PR (possibly using WPT). Landing this as is.

Issue #3980, opened 2020-02-12
Adopt ECMAScript Private Fields


kitsonk:
We should adopt ECMAScript private fields in the internal code within Deno and in `std`.  We have a couple of different patterns in the code base in regards to this (TypeScript `private` and using WeakMaps in Events) and should adopt a consistent mechanism which provides us with the runtime privacy of data.  We have a 3rd pattern where we have marked some fields as private, but they are defined with a symbol to make it easy to utilise an interator interface.  We might reconsider dropping the `private` keyword.

Private fields do not apply to methods (fields are not part of prototype, while methods are), so not all of our use of the `private` keyword could be easily addressed (I am not too personally keen on turning methods into fields, as having them part of the prototype is far more efficient).

TypeScript 3.8 will support ECMAScript private fields, V8 already supports them.  But, right now prettier and dprint do not.  We still use prettier for formatting our internal code, and dprint for `deno fmt`.

Refs #3937 
Refs prettier/prettier#7263
Refs dsherret/dprint#91

PR #8981, opened 2021-01-04
fix(std/http): Use ES private fields in server

zhangyuannie:
As per #3980, this PR unify the way how private fields are written in `std/http`.


satyarohith (commented):
Looks good!

bartlomieju:
@nayeemrmn are you in favor of this change?

ry (approved):
LGTM - thanks!

PR #8984, opened 2021-01-04
feat(op_crates/crypto): Introduce Web Crypto API

### LINKED ISSUE #1891: Implement webcrypto APIs.
Some applications like handling password will require hashing and encryption. These operations could be implemented in js/ts, but It really needs to be handled from a trusted environment to be secure. It looks like there have been some attempts to get this one rolling before recently with #1461.

yacinehmito:
First step to solve #1891.

This PR adds the common code for the Web Crypto API, i.e.:
- TypeScript Type definitions
- Core logic in the JS wrapper related to algorithm normalisation

# Progress

We document progress on solving #1891 here:

- [x] Type definitions (#8984)
- [ ] JS wrapper
  - [x] Algorithm normalization (#8984)
  - [ ] Serialising `keyUsages`
- [x] Tests
  - [x] Add wpt (@lucacasonato #8990)
  - [x] Make wpt run `.worker.js` test suites (#9065)
- [x] Pick a crypto library (starting with [ring](https://docs.rs/ring/0.16.19/ring/index.html), then will complement with others for missing operations)
- [ ] Implementations
  - [ ] `decrypt()`
  - [ ] `deriveBits()`
  - [ ] `deriveKey()`
  - [x] `digest()` (#9069)
  - [ ] `encrypt()`
  - [ ] `exportKey()`
  - [ ] `generateKey()`
  - [ ] `importKey()`
  - [ ] `sign()`
  - [ ] `unwrapKey()`
  - [ ] `verify()`
  - [ ] `wrapKey()`
- [ ] Various fixes
  - [x] Make `getRandomValues()` spec-compliant (#9016)
  - [x] Rename `op_get_random_values` to `op_crypto_get_random_values` (#9067)
  
# Design

## Type definitions

The type definitions have been taken from https://github.com/microsoft/TypeScript/blob/2428ade1a91248e847f3e1561e31a9426650efee/src/lib/webworker.generated.d.ts#L3060.
They are consistent with the [W3C recommendation](https://www.w3.org/TR/WebCryptoAPI/).
I made a slight change regarding the `keyUsages` parameters by giving them the type `Iterable<KeyUsage>` instead of `KeyUsage[]`. Indeed, Chrome, Firefox and Safari all accept iterables of strings, not just arrays (the specs don't specify the exact type; they just mention "sequences").

### crypto.subtle

I went with a simple literal object for `crypto.subtle`, as we had with `crypto` itself. It's not how Chrome, Firefox and Safari do it: they expose the "classes" `Crypto` and `SubtleCrypto`, which can't be instantiated. The methods of `crypto` and `crypto.subtle` aren't in the object itself but on the respective prototype.
Let me know if you want me to mirror that behaviour in Deno.

### JS side vs. Rust side

The validation logic and the dictionary of algorithms are handled on the JS side because:
- It more convenient for me. I wouldn't know how to do the same thing with Rust without some research.
- Building a new object that we control before serialising it for Rust is probably faster and safer than serialising whatever the caller provides directly. If the caller provides an object with circular references and a lot of data, this won't impact performance as we only serialise what we need.

### Returning buffers

We instantiate buffers on the JS side as `ArrayBuffer`, pass them wrapped in an `Uint8Array` to the Rust side as zero-copy buffers, then fill them with the data that is supposed to be returned.
That way, we don't have to serialise the data returned by the underlying crypto library when it is large.

However, this assumes that we know in advance the size of the `ArrayBuffer` on the JS side. So far, this hasn't been a big issue, but it does add complexity.

It would be handy if we could instantiate an `ArrayBuffer` directly on the Rust side and return a handle to the JS side.

## Underlying crypto library

<details>
  <summary>Deliberating which crypto library to pick</summary>

   It's not clear which crypto library I should use to implement the Rust ops.
- [ring](https://briansmith.org/rustdoc/ring/) seems to be the go-to crypto library in Rust, but it doesn't support RSA key generation (see https://github.com/briansmith/ring/issues/219). Someone attempted a pull request a while back (see https://github.com/briansmith/ring/pull/733) but the maintainer seemed unresponsive and it stalled. I find that a bit concerning, so I don't know if I should go with ring.
- Firefox relies on [NSS](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS) for the Web Crypto API. Sadly there are no up-to-date crate with bindings for it.
- Safari and Chrome seem to mostly rely on OpenSSL. That would be the safe choice, but integrating OpenSSL has some [non-trivial requirements](https://docs.rs/openssl/0.10.32/openssl/). I'm too green on Deno to understand the implications and determine whether this would be a benign change or something to avoid.

I require some guidance as to which library should be included. Once it's decided, I expect the implementations to be rather easy.

</details>

We will go for [ring](https://docs.rs/ring/0.16.19/ring/index.html), as [per @lucacasonato's recommendation](https://github.com/denoland/deno/pull/8984#issuecomment-754352264). It won't be enough to implement all operations, but we'll see when that comes up.

## Testing

Chrome, Firefox and Safari all share the same test suite for the Web Crypto API, as they just vendored the relevant tests from [web-platform-tests](https://github.com/web-platform-tests/wpt). I figured that the safest and easiest way to test the Web Crypto API on deno would be to do the same.

However, web-platform-tests are meant to run in browsers. There is a heavy reliance on a DOM, and scripts are assumed to be running in the global scope. To be able to run those tests unchanged in Deno, we would need to reimplement their [JavaScript test harness](https://github.com/web-platform-tests/wpt/blob/master/resources/testharness.js) from scratch.

~I deemed this not worth the effort, and settled on rewriting the tests so that they can directly run as unit tests in Deno. It has the advantage of making the test suite more idiomatic (and arguably easier to read as it's in modern TypeScript, as opposed to ES5). However it makes it all the more challenging to follow upstream. I don't think it's a big deal given the very slow rate of changes there, but it ought to be mentioned.~

@lucacasonato added the the web-platform-tests as part of #8990.

### DIFF
--- a/op_crates/crypto/01_crypto.js
+++ b/op_crates/crypto/01_crypto.js
@@ -1,6 +1,8 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 "use strict";
 
+// Implements https://www.w3.org/TR/WebCryptoAPI
+
 ((window) => {
   const core = window.Deno.core;
 
@@ -41,11 +43,145 @@
     return arrayBufferView;
   }
 
+  // Algorithm normalization, which involves storing the expected data for
+  // each pair of algorithm and crypto operation, is done on the JS side because
+  // it is convenient.
+  // Shall it stay that way, or are we better to move it on the Rust side?
+
+  // We need this method after initialization, anytime we need to normalize
+  // a provided algorithm. We store it here to prevent prototype pollution.
+  const toUpperCase = String.prototype.toUpperCase;
+
+  class RegisteredAlgorithmsContainer {
+    #nameIndex;
+    #definitions;
+
+    constructor(definitions) {
+      this.#nameIndex = Object.create(null);
+      this.#definitions = Object.create(null);
+      for (const [name, definition] of Object.entries(definitions)) {
+        this.#nameIndex[name.toUpperCase()] = name;
+        this.#definitions[name] = definition;
+      }
+    }
+
+    /**
+     * A definition is an object whose keys are the keys that the input must
+     * have and whose values are validation functions for the associate value.
+     * The validation function will either return the value if it valid, or
+     * throw an error that must be forwarded.
+     */
+    getDefinition(name) {
+      return this.#definitions[name];
+    }
+
+    normalizeName(providedName) {
+      const upperCaseName = toUpperCase.call(providedName);
+      return this.#nameIndex[upperCaseName];
+    }
+  }
+
+  const supportedAlgorithms = {};
+
+  function normalizeAlgorithm(algorithm, registeredAlgorithms) {
+    let alg;
+    if (typeof algorithm === "string") {
+      alg = { name: algorithm };
+    } else if (typeof algorithm === "object" && algorithm !== null) {
+      if (typeof algorithm.name !== "string") {
+        throw new TypeError("Algorithm name is missing or not a string");
+      }
+      alg = { ...algorithm };
+    } else {
+      throw new TypeError("Argument 1 must be an object or a string");
+    }
+    const algorithmName = registeredAlgorithms.normalizeName(alg.name);
+    if (algorithmName === undefined) {
+      throw new DOMException(
+        "Unrecognized algorithm name",
+        "NotSupportedError",
+      );
+    }
+    const definition = registeredAlgorithms.getDefinition(algorithmName);
+    for (const [propertyName, validate] of Object.entries(definition)) {
+      alg[propertyName] = validate(algorithm[propertyName]);
+    }
+    alg.name = algorithmName;
+    return alg;
+  }
+
+  const subtle = {
+    async decrypt(algorithm, key, data) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async deriveBits(algorithm, baseKey, length) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async deriveKey(
+      algorithm,
+      baseKey,
+      derivedKeyType,
+      extractable,
+      keyUsages,
+    ) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async digest(algorithm, data) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async encrypt(algorithm, key, data) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async exportKey(format, key) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async generateKey(algorithm, extractable, keyUsages) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async importKey(format, keyData, algorithm, extractable, keyUsages) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async sign(algorithm, key, data) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async unwrapKey(
+      format,
+      wrappedKey,
+      unwrappingKey,
+      unwrapAlgorithm,
+      unwrappedKeyAlgorithm,
+      extractable,
+      keyUsages,
+    ) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async verify(algorithm, key, signature, data) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async wrapKey(format, key, wrappingKey, wrapAlgorithm) {
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+  };
+
   window.crypto = {
     getRandomValues,
+    subtle,
   };
   window.__bootstrap = window.__bootstrap || {};
   window.__bootstrap.crypto = {
     getRandomValues,
+    subtle,
   };
 })(this);
--- a/op_crates/crypto/Cargo.toml
+++ b/op_crates/crypto/Cargo.toml
@@ -16,4 +16,7 @@ path = "lib.rs"
 [dependencies]
 deno_core = { version = "0.79.0", path = "../../core" }
 rand = "0.8.3"
-
+ring = "0.16.19"
+tokio = { version = "1.1.1", features = ["full"] }
+serde_json = { version = "1.0.62", features = ["preserve_order"] }
+serde = { version = "1.0.123", features = ["derive"] }
--- a/op_crates/crypto/lib.rs
+++ b/op_crates/crypto/lib.rs
@@ -11,6 +11,7 @@ use deno_core::ZeroCopyBuf;
 use rand::rngs::StdRng;
 use rand::thread_rng;
 use rand::Rng;
+use std::path::PathBuf;
 
 pub use rand; // Re-export rand
 
@@ -40,4 +41,8 @@ pub fn op_crypto_get_random_values(
   }
 
   Ok(json!({}))
+}
+
+pub fn get_declaration() -> PathBuf {
+  PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("lib.deno_crypto.d.ts")
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,6 +404,7 @@ dependencies = [
  "chrono",
  "clap",
  "deno_core",
+ "deno_crypto",
  "deno_doc",
  "deno_fetch",
  "deno_lint",
@@ -483,6 +484,10 @@ version = "0.13.0"
 dependencies = [
  "deno_core",
  "rand 0.8.3",
+ "ring",
+ "serde",
+ "serde_json",
+ "tokio",
 ]
 
 [[package]]
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -28,6 +28,7 @@ deno_core = { path = "../core", version = "0.79.0" }
 deno_fetch = { path = "../op_crates/fetch", version = "0.22.0" }
 deno_web = { path = "../op_crates/web", version = "0.30.0" }
 deno_websocket = { path = "../op_crates/websocket", version = "0.5.0" }
+deno_crypto = { path = "../op_crates/crypto", version = "0.13.0" }
 regex = "1.4.3"
 serde = { version = "1.0.123", features = ["derive"] }
 
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -59,6 +59,7 @@ fn create_compiler_snapshot(
   op_crate_libs.insert("deno.web", deno_web::get_declaration());
   op_crate_libs.insert("deno.fetch", deno_fetch::get_declaration());
   op_crate_libs.insert("deno.websocket", deno_websocket::get_declaration());
+  op_crate_libs.insert("deno.crypto", deno_crypto::get_declaration());
 
   // ensure we invalidate the build properly.
   for (_, path) in op_crate_libs.iter() {
@@ -258,6 +259,10 @@ fn main() {
   println!(
     "cargo:rustc-env=DENO_WEBSOCKET_LIB_PATH={}",
     deno_websocket::get_declaration().display()
+  );
+  println!(
+    "cargo:rustc-env=DENO_CRYPTO_LIB_PATH={}",
+    deno_crypto::get_declaration().display()
   );
 
   println!("cargo:rustc-env=TARGET={}", env::var("TARGET").unwrap());
--- a/cli/dts/lib.deno.shared_globals.d.ts
+++ b/cli/dts/lib.deno.shared_globals.d.ts
@@ -8,6 +8,7 @@
 /// <reference lib="deno.web" />
 /// <reference lib="deno.fetch" />
 /// <reference lib="deno.websocket" />
+/// <reference lib="deno.crypto" />
 
 declare namespace WebAssembly {
   /**
@@ -415,26 +416,6 @@ declare interface Console {
 }
 
 declare var console: Console;
-
-declare interface Crypto {
-  readonly subtle: null;
-  getRandomValues<
-    T extends
-      | Int8Array
-      | Int16Array
-      | Int32Array
-      | Uint8Array
-      | Uint16Array
-      | Uint32Array
-      | Uint8ClampedArray
-      | Float32Array
-      | Float64Array
-      | DataView
-      | null,
-  >(
-    array: T,
-  ): T;
-}
 
 interface MessageEventInit<T = any> extends EventInit {
   data?: T;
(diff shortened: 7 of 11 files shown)


yacinehmito:
This is PR is very WIP. I opened it so that you can provide guidance early enough.

lucacasonato:
@yacinehmito FYI I have started implementing WPT in #8990. That might be helpful for this PR.

Regarding which crypto lib to use: lets use `ring`, even if it doesn't have support for RSA key gen yet. I am sure we can figure something out once we get to the point where that becomes an issue (maybe `rsa` crate?). I think starting with the more "simple" stuff like digest would be great.

yacinehmito:
I updated the PR description to account for the decisions that have been made and the changes in approach, and added a few more information on:
- Whether to put code on the JS side or on the Rust side
- How to return a buffer without serialising it

yacinehmito:
I pushed a rough implementation of digest using ring.
There are a few things to sort out; I'll detail them in a self-review.

Also, I wonder if it wouldn't be preferable to put the `SubtleCrypto` interface behind the `--unstable` flag, so that we can merge the implementations little by little instead of doing a mega PR. What do you think?

core/zero_copy_buf.rs
```diff
@@ -23,7 +23,8 @@ pub type BufVec = SmallVec<[ZeroCopyBuf; 2]>;
 pub struct ZeroCopyBuf {
   backing_store: v8::SharedRef<v8::BackingStore>,
   byte_offset: usize,
-  byte_length: usize,
+  // TODO: Determine whether it's worth exposing just to assert the size
+  pub byte_length: usize,
```
yacinehmito, line 27: This is supposed to be private, but I needed it to check that the size of the provided buffer is the expected one.
I don't think exposing that data is harmful. Let me know what you prefer.

If we want to keep this field private, we can also manually pass the size of the instantiated buffer as part of `Value`. It's a bit more work but it's doable.

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,211 @@
...
     );
     core.jsonOpSync("op_get_random_values", {}, ui8);
-    return typedArray;
+    return arrayBufferView;
   }
+
+  // Algorithm normalization, which involves storing the expected data for
+  // each pair of algorithm and crypto operation, is done on the JS side because
+  // it is convenient.
+  // Shall it stay that way, or are we better to move it on the Rust side?
```
yacinehmito, line 48: What do you think?

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,211 @@
...
+    }
+
+    normalizeName(providedName) {
+      const upperCaseName = toUpperCase.call(providedName);
+      return this.#nameIndex[upperCaseName];
+    }
+  }
+
+  const supportedAlgorithms = {
+    digest: new RegisteredAlgorithmsContainer({
```
yacinehmito, line 83: This structure follows the spec. It mentions that implementations need a container called `supportedAlgorithms` that maps operation names to `registeredAlgorithms`, which in turn maps algorithm names to the expected shape of the algorithm input parameters.

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,211 @@
...
+      "SHA-512": {},
+    }),
+  };
+
+  const digestByteLengths = {
+    "SHA-1": 20,
+    "SHA-256": 32,
+    "SHA-384": 48,
+    "SHA-512": 64,
+  };
```
yacinehmito, line 96: I don't like that we need to repeat that information on the JS side. `ring` already knows about this.
I opted for copying it here and asserting that the size of the instantiated buffer matches the one expected by `ring`.
(This is why I exposed `ZeroCopyBuf.byte_length` earlier, although it is not strictly necessary)

I see two other options to avoid this:
1. Instantiate the `ArrayBuffer` on the Rust side. IMHO it is preferable, but would require significant work.
2. Expose an OP to fetch the sizes defined on the Rust side. It's not as good in terms of guarantees, but at least we don't have to hardcode the lengths like that.

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,211 @@
...
+
+  function normalizeAlgorithm(algorithm, registeredAlgorithms) {
```
yacinehmito, line 98: Doing all that work in JS because:
- It's easier
- We only ever have to serialise algorithm parameters that are normalised, meaning we will never attempt to serialise the input as provided by the user (which could be huge, be of a weird type, or contain circular references)

op_crates/crypto/lib.rs
```diff
@@ -41,3 +48,43 @@ pub fn op_get_random_values(
...
+}
+
+pub async fn op_subtle_digest(
+  _state: Rc<RefCell<OpState>>,
+  args: Value,
+  zero_copy: BufVec,
+) -> Result<Value, AnyError> {
+  let mut zc = zero_copy;
+
+  // FIXME: This chain of `unwrap` is meh.
```
yacinehmito, line 70: Is it deemed not clean enough, and if so how would you fix that?

op_crates/crypto/lib.rs
```diff
@@ -41,3 +48,43 @@ pub fn op_get_random_values(
...
+    .as_object()
+    .unwrap()
+    .get("name")
+    .unwrap()
+    .as_str()
+    .unwrap();
+  let digest_algorithm = DIGEST_ALGORITHMS_BY_NAME.get(alg_name).unwrap();
+  // TODO: This should run in another thread
+  // Otherwise, even though the API is async, it is blocking the main thread
+  // See https://stackoverflow.com/questions/61292425/how-to-run-an-asynchronous-task-from-a-non-main-thread-in-tokio
```
yacinehmito, line 81: From what I understand, `jsonOpAsync()` doesn't necessarily mean that the code runs asynchronously. It just means that it returns a `Promise` (a `Future` on Rust side).
So, to make it truly asynchronous, we'd have to run the call to ring's `digest()` in another thread. I have no idea how to do this. Is there an example of this already in the codebase?

op_crates/crypto/lib.rs
bartlomieju: Users can still send garbage data manually, you use use something like `let args: SubtleDigestArgs = serde_json::from_value(args)?;` like it's done in other ops.

op_crates/crypto/lib.rs
bartlomieju: There's no point in making this op async if the Rust API is sync; just use sync op and if you need to return promise do it manually in JS

core/zero_copy_buf.rs
bartlomieju: Please revert, you don't need to change `ZeroCopyBuf` to check length; `ZeroCopyBuf` derefs to `[u8]` and you can use its `size()` to check it.

op_crates/crypto/lib.rs
lucacasonato: If you have some synchronous CPU intensive operation with an async API in JS, you can run it in the tokio blocking thread pool (tokio::task::spawn_blocking).

op_crates/crypto/01_crypto.js
lucacasonato: These are fixed values that will not change in the future (defined by spec), so it is fine to hard code them. Sending over the information in an extra op adds too much complexity and overhead.

op_crates/crypto/lib.rs
bartlomieju: Good point 👍

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,211 @@
...
+    },
+    async digest(algorithm, data) {
+      const alg = normalizeAlgorithm(algorithm, supportedAlgorithms.digest);
+      // TODO: validate that data is indeed a BufferSource
+      // Note: we need to copy the data because the specs require it.
+      // It's unclear whether we are better doing it here or on the Rust side.
+      const input = data.slice(0);
+      const output = new ArrayBuffer(digestByteLengths[alg.name]);
+      await core.jsonOpAsync(
+        "op_subtle_digest",
```
lucacasonato, line 150: I would prefer if all ops in the crypto op crate would start with with `op_crypto`.

op_crates/crypto/lib.rs
yacinehmito: @lucacasonato Thank you very much, it's exactly what I was looking for.

lucacasonato (commented):
Looks very good so far. Regarding your question about merging this in smaller PRs: yes please. It makes reviewing a lot easier. I don't think there is a need to put this behind unstable, because there will be no interface changes in the future as the interface is defined by the WebCrypto spec. We don't consider bug fixes (or alignment with browser) in Web APIs breaking changes.

If you think it makes sense to put this behind --unstable though, we can do that too. @bartlomieju do you have opinions on this?

op_crates/crypto/lib.rs
yacinehmito: > Users can still send garbage data manually

How? Can they call an OP directly without going through the JS API?

Regardless, I'll implement your suggestion.

op_crates/crypto/01_crypto.js
yacinehmito: Shall I also rename the one for `getRandomValues()`?

op_crates/crypto/lib.rs
lucacasonato: > Can they call an OP directly without going through the JS API?

Yes. This is not covered under any stability guarantee though, and is not officially documented. We should try minimize rust panics where possible and gracefully return errors to JS.

op_crates/crypto/01_crypto.js
lucacasonato: Yes please.

op_crates/crypto/lib.rs
bartlomieju: > How? Can they call an OP directly without going through the JS API?

They can using JS API:
```js
Deno.core.dispatch("op_crypto", new TextEncoder().encode("some garbage data that is not valid json"));
```

bartlomieju:
> If you think it makes sense to put this behind --unstable though, we can do that too. @bartlomieju do you have opinions on this?

Yeah I agree; unstable is superfluous in this case.

yacinehmito:
I'll split the PR the following way:
- This one will have the common stuff: TS type definitions & common code on the JS side. It should be mergeable very quickly.
- I'll create a new PR to rename the op of `getRandomValues()`
- I'll create a new PR for the digest implementation specifically, and then one for each new operation after that (or more)

I'll keep reporting on the status of the overall task on this PR as #1891 is a bit too cluttered for that. Else, I can create a new issue to replace #1891. It's however you want.

yacinehmito:
PRs #9016 and #9067 need to be merged first.

yacinehmito:
The PR is up-to-date with `master`.

The type definitions have been moved to the op crate, following the convention set by the others (web, fetch and websocket).

yacinehmito:
@lucacasonato @bartlomieju I think it is ready for merging.

caspervonb:
Enable the relevant wpt tests? see the contributing docs for how the new wpt runner works.

yacinehmito:
This PR has no implementation. The tests shouldn't be enabled. They will be selectively added as more crypto functions are implemented.

bartlomieju:
@yacinehmito thanks for the PR, there's a lot of changes but I can find any implementation of actual APIs. Is this a rebase mistake? If not then I'd prefer to land a PR that contains implementation of at least one algo.

op_crates/crypto/01_crypto.js
```diff
@@ -40,11 +42,145 @@
...
+  // Algorithm normalization, which involves storing the expected data for
+  // each pair of algorithm and crypto operation, is done on the JS side because
+  // it is convenient.
+  // Shall it stay that way, or are we better to move it on the Rust side?
+
+  // We need this method after initialization, anytime we need to normalize
+  // a provided algorithm. We store it here to prevent prototype pollution.
+  const toUpperCase = String.prototype.toUpperCase;
+
+  class RegisteredAlgorithmsContainer {
```
bartlomieju, line 54: This class is never used

op_crates/crypto/Cargo.toml
```diff
@@ -16,4 +16,7 @@ path = "lib.rs"
 [dependencies]
 deno_core = { version = "0.77.1", path = "../../core" }
 rand = "0.8.2"
-
+ring = "0.16.19"
+tokio = { version = "1.1.1", features = ["full"] }
+serde_json = { version = "1.0.61", features = ["preserve_order"] }
+serde = { version = "1.0.121", features = ["derive"] }
```
bartlomieju, line 22: Why are all those dependencies added here? `serde` and `serde_json` are reexported from `deno_core` anyway, so no need to declare them here

yacinehmito:
@bartlomieju It used to have one, but if you go back the comments thread you see that I was asked to separate implementations in multiple PRs.
I'm fine with the decision itself, it's just that I'd be going back on something I have already done for a value that is not very tangible to me; the PR with the digest implementation is already open, and most operations are not implemented anyway.

caspervonb:
> This PR has no implementation. 

Okay I can see what you're going for here.

I am strongly against implementing throwing stubs; there are packages in the wild that do progressive enhancement so these should just still be undefined so they can be tested for using normal property lookups.

yacinehmito:
This is sensible, and why I suggested to put this behind `unstable`, but it was rejected.
I'll let you core contributors figure out a good rollout strategy for the WebCrypto API.

In the meantime I will work on other implementations and open draft PRs as I go.

Let me know once you have a definitive plan that people are aligned on.

op_crates/crypto/Cargo.toml
yacinehmito: @bartlomieju All other op_crates include `serde` as a dependency. Why would this one be any different?
If they shouldn't, I'd be happy to open a PR to remove them.

op_crates/crypto/01_crypto.js
yacinehmito: It is meant to be used in further PRs.

yacinehmito:
@caspervonb @bartlomieju @lucacasonato
Have you settled on how you would wish to incrementally rollout the WebCrypto API to the codebase?

Exploring options:

### Introduce the ops one by one, then the TypeScript declaration last

It's very simple to do, and nobody using TypeScript will see the WebCrypto API until it is fully implemented. However, the API would still be exposed because we need it for wpt tests.

### Introduce the API first, then the ops one by one

This is what this PR does. Undesirable at this leads to runtime errors.

### Put the API behind the unstable flag

Would allow us plenty of iteration without being too concerned about exposing a half-finished API. Proposition already rejected though.

### Use a feature branch

Have a branch dedicated to implementing the WebCrypto API, then merge smaller PRs into it. Would allow for releasing the WebCrypto API at once, but may lead to integration issues if the branch is not constantly rebased.

Thoughts?

laughedelic:
Is there any chance this PR could be recovered? Are there any other efforts/plans to add Web Crypto?

caspervonb:
Yeah, it's coming soon.

PR #8986, opened 2021-01-04
fix(inspector): kill child process after test

### LINKED ISSUE #8979: Reland #8894
In #8977 we reverted #8894 because it introduced openssl to our dependency tree (even if it was only for tests). It also added a lot of `hello` messages to the output logs in testing. The openssl issue can be solved by using the `reqwest` that is exported from `deno_fetch`.

@lperlaki We would appreciate if this PR could be re-landed with fixes. Do you have time to address these issues?

bnoordhuis:
The child process kept running and printing "hello" to stdout.

This commit also removes the dependency on reqwest and instead
switches to the re-export from the fetch crate.

Refs commit 1a2e7741c33490d2a91147966019853c6b1d6a48.

Closes #8979

cc @lucacasonato @lperlaki

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -1,7 +1,9 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 use deno_core::futures;
 use deno_core::futures::prelude::*;
+use deno_core::serde_json;
 use deno_core::url;
+use deno_runtime::deno_fetch::reqwest;
 use std::io::{BufRead, Write};
 use std::process::Command;
 use tempfile::TempDir;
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Groan... 379eedc05c1a841557d8f0a9c2cd33ec3362d607 was merged just before I opened this PR. Rebased.

bartlomieju:
> Groan... [379eedc](https://github.com/denoland/deno/commit/379eedc05c1a841557d8f0a9c2cd33ec3362d607) was merged just before I opened this PR. Rebased.

Yikes sorry for that... So this PR closes #8979

bartlomieju (approved):
LGTM

PR #8990, opened 2021-01-04
tests: add web platform test runner

### LINKED ISSUE #6229: discussion: Run web platform tests
[WPT](https://github.com/web-platform-tests/wpt) is an extensive suite of tests that tests nearly all web platform features like `fetch`, `console`, and streams. It would be great if we could run the for us relevant tests during CI. Chrome and Firefox run WPT during testing too.

lucacasonato:
Closes #6229

cc @kitsonk I only enabled a minimal set of tests for `streams`, because most of the rest does not work. We can start enabling them one by one in follow ups.

Files are all concatenated together, because WPT tests are meant to work with scripts, not modules. ~Once we support external source mapping, I will add source mapping for WPT tests.~ I have added source mapping. I have checked that it works with #8995.


cli/tests/wpt_testharnessconsolereporter.js
```diff
@@ -0,0 +1,119 @@
...
+    console.log(
+      `\n${result.name}\n${result.message}\n${result.stack}`,
+    );
+  }
+
+  if (failed.length > 0) {
+    console.log(`\nfailures:\n`);
+  }
+  for (const result of failed) {
+    console.log("        " + result.name);
```
kitsonk, line 104:
```suggestion
    console.log(`        ${result.name}`);
```

cli/tests/wpt_testharnessconsolereporter.js
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export function yellow(str) {
+  return run(str, code([33], 39));
+}
+
+const testResults = [];
+const testsExpectFail = JSON.parse(Deno.args[0]);
+
+window.add_result_callback((test) => {
```
kitsonk, line 33:
```suggestion
window.add_result_callback(({ message, name, stack, status }) => {
```

And then you can just use them destructured below...

cli/tests/wpt_testharnessconsolereporter.js
```diff
@@ -0,0 +1,119 @@
...
+  console.log(simpleMessage);
+
+  testResults.push({
+    name: test.name,
+    passed: test.status === 0,
+    expectFail,
+    message: test.message,
+    stack: test.stack,
+  });
+});
```
kitsonk, line 76: If you do the above...

```suggestion
  testResults.push({
    name,
    passed: status === 0,
    expectFail,
    message,
    stack,
  });
});
```

kitsonk (commented):
A couple thoughts/nits

.gitmodules
```diff
@@ -5,3 +5,6 @@
 [submodule "std/wasi/testdata"]
 	path = std/wasi/testdata
 	url = https://github.com/khronosproject/wasi-test-suite.git
+[submodule "cli/tests/wpt"]
```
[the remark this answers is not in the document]
lucacasonato: Maybe into `test_util`? We wouldn't need a new top level directory then.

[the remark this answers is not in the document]
kitsonk: 👍

kitsonk (approved):
Awesome! LGTM.

axetroy:
This is really awesome!

If you can output the test results to a JSON file and render them in `https://deno.land`, that would be great

lucacasonato:
> If you can output the test results to a JSON file and render them in `https://deno.land`, that would be great

Sounds like a good idea. Will investigate.

PR #8991, opened 2021-01-04
docs(examples): Close connections in TCP echo server

mk12:
Based on discussion with @lucacasonato in the #help chat.

I was confused at first and thought connections were automatically closed, since the TCP echo server example did not explicitly call `conn.close()`. But I understand now the example is wrong and leaks resources. This should fix it.


mk12:
Ah, not worth the hassle of getting the CLA approved at work for such a simple change. I suggest someone else fix this.

PR #8995, opened 2021-01-05
fix: use inline source maps when present in js

lucacasonato:
This PR also fixes source_line mapping, because currently it was making some wrong (but up to now correct) assumptions about where to get the source map info from. This has been corrected. The order of priority for source line info is now 1. the inline source map sourcesContent, 2. the module graph, and 3. the original source_line form the transpiled code.

Towards #4499.


kitsonk (commented):
Looks good... kind of sorta what I had in mind that I was going to start until you tortured me with Streams.  It should be easy to continue to iterate on it.

Still some test errors, so holding off approval.

cli/tests/inline_js_source_map_with_contents_from_graph.js.out
```diff
@@ -0,0 +1,4 @@
+error: Uncaught Error: Hello world!
+// throw new Error("Hello world!" as string);
```
bartlomieju, line 2: This looks fishy to me, why is there error thrown when the line is commented out?

lucacasonato: The error is not thrown by that file. The error is thrown by `cli/tests/inline_js_source_map_with_contents_from_graph.js`, which source maps to `cli/tests/inline_js_source_map_with_contents_from_graph.ts`.

bartlomieju: Sorry I don't understand, there is no `inline_js_source_map_with_contents_from_graph.ts` file here

cli/tests/073_worker_error.ts.out
```diff
@@ -2,4 +2,6 @@
     at foo ([WILDCARD])
     at [WILDCARD]
 error: Uncaught (in promise) Error: Unhandled error event reached main worker.
+              throw new Error("Unhandled error event reached main worker.");
+                    ^
```
bartlomieju, line 6: Some output has more left padding than the others (eg. `059_fs_relative_path_perm.ts.out`). Is that intentional?

cli/tests/inline_js_source_map_with_contents_from_graph.js.out
lucacasonato: Sorry I mean't `cli/tests/inline_js_source_map.ts`.

bartlomieju: Okay, then again why is it throwing an error if that line is commented out :D ?

cli/tests/073_worker_error.ts.out
lucacasonato: This is not padding (that is always 8 spaces), this is just the indentation before the statement in the file.

cli/tests/inline_js_source_map_with_contents_from_graph.js.out
lucacasonato: Because that is not where the error is thrown. It is thrown at `throw new Error("Hello world!");` in `cli/tests/inline_js_source_map_with_contents_from_graph.js`, which source maps to `// throw new Error("Hello world!" as string);` in `cli/tests/inline_js_source_map.ts`, so this is what is displayed in the error. These tests use some hand crafted source maps to check that we get the source_line from the correct file. No compiler would ever emit anything like this.

bartlomieju: I see, maybe add some comments about hand crafted files and source maps? It's pretty convoluted and I didn't understand what was going on, might be helpful in the future

cli/source_maps.rs
```diff
@@ -121,11 +123,34 @@ pub fn get_orig_position<G: SourceMapGetter>(
...
           None => default_pos,
-          Some(original) => (
-            original.to_string(),
-            i64::from(token.get_src_line()) + 1,
-            i64::from(token.get_src_col()) + 1,
-          ),
+          Some(original) => {
+            let maybe_source_line =
+              if let Some(source_view) = token.get_source_view() {
+                debug!("source view");
```
kitsonk, line 129: Is this needed?

kitsonk (approved):
LGTM, though one stray debug statement

cli/source_maps.rs
lucacasonato: No. Removed.

PR #8996, opened 2021-01-05
BREAKING(std/wasi): make implementation details private

caspervonb:
This makes the implementation details of the Context class private, making the public interface much more bare-bones and aligns it closer to what Node exposes which is only the exports.


bartlomieju (approved):
LGTM

PR #8998, opened 2021-01-05
chore(std/mime/multipart): fix misspelling of multipartFormData()

kaimallea:
Changes `multipat` to `multipart` in the two usages of `multipartFormData()`.


ry (approved):
Thanks

Issue #8623, opened 2020-12-05
Allow stdio to be passed into WASI contexts


caspervonb:
Currently std/wasi.Context uses the system's stdio, this should be overridable in a similar fashion to Deno.Process.

PR #8999, opened 2021-01-05
feat(std/wasi): allow stdio resources to be specified

caspervonb:
Closes https://github.com/denoland/deno/issues/8623


lucacasonato (approved):
LGTM. Thanks

PR #9011, opened 2021-01-05
enable stale bot

ry:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,22 +1,16 @@
 # https://github.com/probot/stale
 
-# Number of days of inactivity before an issue becomes stale
 daysUntilStale: 60
 
-# Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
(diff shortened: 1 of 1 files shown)


lucacasonato:
I think we should exempt issues that have `feat` or `bug` label from being closed.

kitsonk:
I agree... possibly also "suggestion" as quite a few community suggestions are still suggestions that we don't consider "feat" yet, but don't want to dismiss the suggestion.

kitsonk:
Here is the query for what could be considered "stale" eventually, given we apply those filters: https://github.com/denoland/deno/issues?q=is%3Aopen+is%3Aissue+-label%3Abug+-label%3Afeat+-label%3Asuggestion+sort%3Aupdated-asc+

There are clearly some issues that need to be labelled better, as we wouldn't specifically want to close them as stale.

bartlomieju (approved):

Issue #9001, opened 2021-01-05
Web Platform Tests coverage tracking issue


lucacasonato:
In #8990 we added a WPT runner, and the entire WPT test suite to the repo. We are now able to run web platform tests. Currently only a very limited amount of tests is enabled, and many tests will fail once we enable testing them. We would like to get more coverage for our built in Web APIs. I have listed all of the suites that are potentially relevant for us below.

- [x] Compression (once we add `CompressionStream` and `DecompressionStream`)
- [x] Broadcast Channel API (#10527)
- [x] Console API (#9013)
- [x] DOM APIs
	- [x] abort/event.any.js
- [x] Encoding (#9004)
- [x] HR Time
	- [x] monotonic-clock.any.js
	- [x] window-worker-timeOrigin.window.js requires blob URL support for workers
- [ ] HTML
	- [x] webappapis/atob	
	- [x] webappapis/timers
	- [x] webappapis/microtask-queuing
	- [ ] various tests from browsers/the-window-object
- [x] Fetch API (ignore tests related to same site policy, like CORS)
	- [x] Requires relative `fetch` (#7369)
	- [x] Headers (#10199)
	- [x] Request (#10203)
	- [x] Response (#10203)
	- [x] fetch (#10203)
- [x] File API
- [x] WHATWG Stream API
- [x] URL API (#9046)
- [x] User Timing (`self.perfomance`)
- [ ] WebGPU (once we land it)
- [x] WebSockets
- [x] Workers (not the SharedWorker tests)
- [x] WASM

If you would like to enable one of these suites (and fix the resulting issues), please comment on the issue so we don't have multiple people working on the same suites.

lucacasonato:
I am going to start work on the `URL` test suite.

lucacasonato:
I will also take `Encoding`, because this being correct is a prerequisite for `URL`.

nayeemrmn:
> I am going to start work on the `URL` test suite.

If you're interested in looking into it, at some point we really should replace `URL`'s implementation with `rust-url` since it was literally made for that for Servo/Firefox IIRC.

lucacasonato:
> If you're interested in looking into it, at some point we really should replace URL's implementation with rust-url since it was literally made for that for Servo/Firefox IIRC.

Can you open a new issue?

tarikeshaq:
Hello! Pretty new here, but I can take the `console` tests if that's okay. I'll add the test suites and open a PR to see if we'd like to fix any failing tests!

[...]

PR #9012, opened 2021-01-05
fix: align Performance API to spec using WPT

lucacasonato:
Part of #9001


kitsonk (approved):
Well, that wasn't so bad!

LGTM

PR #9013, opened 2021-01-05
tests(wpt/console): Enables web platform tests for console

tarikeshaq:
A part of #9001 

### Notes:
- There are some `.html` files in the wpt console directory, I ignored those. They seem to be manual tests, anywho just thought I'd note it down.
- ~~I put `TODO(???)` on the failing test suites, they seem to have trivial fixes (like the `Symbol.toStringTag`) but I opted to first get this PR up to see if it's better to file another issue for those (maybe a `good first issue`  👀 ) or solve the trivial ones here~~
Fixes were small enough so I'll add them in a commit here


runtime/js/99_main.js
```diff
@@ -242,6 +242,11 @@ delete Object.prototype.__proto__;
     setTimeout: util.writable(timers.setTimeout),
   };
 
+  // The console seems to be the only one that should be writable and non-enumerable
+  // thus we don't have a unique helper for it. If other properties follow the same
+  // structure, it might be worth it to define a helper in `util`
+  windowOrWorkerGlobalScope.console.enumerable = false;
```
tarikeshaq, line 248: Wasn't sure if this is the best place to do this, it seemed to me that non of the existing "util" functions do the desired behavior and since this seemed like a  one-off I put it here, open to moving this

runtime/js/02_console.js
```diff
@@ -1451,6 +1451,14 @@
...
       const console = Object.create({});
       Object.assign(console, this);
+      // We define the property on the console object since the Object.assign
+      // does not copy the property definitions
+      Object.defineProperty(console, Symbol.toStringTag, {
+        enumerable: false,
+        writable: false,
+        configurable: true,
+        value: "console",
+      });
```
00ff0000red, line 1461: Maybe merge the two statements into one?
```suggestion
const console = Object.create({}, {
    [Symbol.toStringTag]: {
      enumerable: false,
      configurable: true,
      value: "console",
    }
});

Object.assign(console, this);
```

ry (approved):
LGTM - thanks @tarikeshaq

Issue #9009, opened 2021-01-05
bug: performance.measure throws SyntaxError instead of DOMException


lucacasonato:
`self.performance.measure("measure", "mark")` should throw `DOMException` with code 12 (syntax exception). Instead it throws `SyntaxError`. Found via WPT. Tests can be enabled via `cli/tests/wpt.jsonc`.

cc @kitsonk

lucacasonato:
This should be trivial to fix after #9015.

PR #9015, opened 2021-01-05
feat(web): Implement DOMException#code

yacinehmito:
Adds the `code` property to the class `DOMException`. It is pseudo-deprecated, as in new types of `DOMException` set it to zero. It is recommended for code to not rely on that value.
However, tests in the recently included web-platform-tests assert the value of the code, so we are forced to implement it. Issue #9009 highlights an example of such a test that rely on `code`.

Source of the values for each error type: https://developer.mozilla.org/en-US/docs/Web/API/DOMException#Error_names

cc @lucacasonato

### DIFF
--- a/op_crates/web/00_dom_exception.js
+++ b/op_crates/web/00_dom_exception.js
@@ -1,29 +1,32 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
 ((window) => {
-  const nameToCodeMapping = Object.create(null);
-  nameToCodeMapping.IndexSizeError = 1;
-  nameToCodeMapping.HierarchyRequestError = 3;
-  nameToCodeMapping.WrongDocumentError = 4;
-  nameToCodeMapping.InvalidCharacterError = 5;
-  nameToCodeMapping.NoModificationAllowedError = 7;
-  nameToCodeMapping.NotFoundError = 8;
-  nameToCodeMapping.NotSupportedError = 9;
-  nameToCodeMapping.InvalidStateError = 11;
-  nameToCodeMapping.SyntaxError = 12;
-  nameToCodeMapping.InvalidModificationError = 13;
-  nameToCodeMapping.NamespaceError = 14;
-  nameToCodeMapping.InvalidAccessError = 15;
-  nameToCodeMapping.TypeMismatchError = 17;
-  nameToCodeMapping.SecurityError = 18;
-  nameToCodeMapping.NetworkError = 19;
-  nameToCodeMapping.AbortError = 20;
-  nameToCodeMapping.URLMismatchError = 21;
-  nameToCodeMapping.QuotaExceededError = 22;
-  nameToCodeMapping.TimeoutError = 23;
-  nameToCodeMapping.InvalidNodeTypeError = 24;
-  nameToCodeMapping.DataCloneError = 25;
-
+  const nameToCodeMapping = Object.create(
+    null,
+    {
+      IndexSizeError: { value: 1 },
+      HierarchyRequestError: { value: 3 },
+      WrongDocumentError: { value: 4 },
+      InvalidCharacterError: { value: 5 },
+      NoModificationAllowedError: { value: 7 },
+      NotFoundError: { value: 8 },
+      NotSupportedError: { value: 9 },
+      InvalidStateError: { value: 11 },
+      SyntaxError: { value: 12 },
+      InvalidModificationError: { value: 13 },
+      NamespaceError: { value: 14 },
+      InvalidAccessError: { value: 15 },
+      TypeMismatchError: { value: 17 },
+      SecurityError: { value: 18 },
+      NetworkError: { value: 19 },
+      AbortError: { value: 20 },
+      URLMismatchError: { value: 21 },
+      QuotaExceededError: { value: 22 },
+      TimeoutError: { value: 23 },
+      InvalidNodeTypeError: { value: 24 },
+      DataCloneError: { value: 25 },
+    },
+  );
   class DOMException extends Error {
     #name = "";
     #code = 0;
--- a/op_crates/web/00_dom_exception.js
+++ b/op_crates/web/00_dom_exception.js
@@ -1,15 +1,50 @@
 // Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-class DOMException extends Error {
-  #name = "";
+((window) => {
+  const nameToCodeMapping = new Map(
+    Object.entries({
+      IndexSizeError: 1,
+      HierarchyRequestError: 3,
+      WrongDocumentError: 4,
+      InvalidCharacterError: 5,
+      NoModificationAllowedError: 7,
+      NotFoundError: 8,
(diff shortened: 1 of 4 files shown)


yacinehmito:
I could have done a map, but I wasn't sure whether it was going to pollute the global scope, so a `switch` it is.

lucacasonato:
It doesn't pollute global scope. Could you switch to a `Map`? Should be less lines changed then. (and also would make it easy to implement the constants on `DOMException` in the future then).

op_crates/web/00_dom_exception.js
```diff
@@ -1,15 +1,47 @@
...
+    ["AbortError", 20],
+    ["URLMismatchError", 21],
+    ["QuotaExceededError", 22],
+    ["TimeoutError", 23],
+    ["InvalidNodeTypeError", 24],
+    ["DataCloneError", 25],
+  ]);
+  class DOMException extends Error {
+    #name = "";
+    #code = 0;
```
yacinehmito, line 29: Nit: I don't like the redundancy of this with the `|| 0` part when setting `#code` in the constructor.
I'm tempted ro remove them both.

kitsonk:
Not to be picky, but why use a map at all, why not just an object literal.  Creating any array of arrays to just populate a map seems arbitrarily complex.  🤷

yacinehmito:
I can do it however you want it. I like `Map` because the `[property]` accessor grosses me out, but that's very particular of me.

I'll define an object literal.

lucacasonato:
I don't like using a plain object in this case either, because what if somenone passes `toString` as the name to the DOMException constructor? `Map` prevents this. Consider `new Map(Object.entries({foo: 1}));`.

yacinehmito:
I'll wait for @kitsonk's answer and I'll change it to what we'll settle on.

00ff0000red:
Just tossing it out there, but you could use a simple, plain, array literal, and `Array#indexOf`.

op_crates/web/00_dom_exception.js
```diff
@@ -1,15 +1,47 @@
...
 
-  constructor(message = "", name = "Error") {
-    super(message);
-    this.#name = name;
-  }
+    constructor(message = "", name = "Error") {
+      super(message);
+      this.#name = name;
+      this.#code = nameToCodeMapping[name] || 0;
```
00ff0000red, line 34: Would the `??` op semantically fit better?

yacinehmito:
@00ff0000red I'm not a fan of linear lookups.

op_crates/web/00_dom_exception.js
yacinehmito: Indeed, the intent may be clear. Functionally though it is equivalent.

00ff0000red:
> @00ff0000red I'm not a fan of linear lookups.

There are no guarantees that object lookup, nor map lookups execute in constant time, I would presume that under the hood, V8 does a linear lookup for every suggestion here so far, with the exception of an object, as it may have to check the prototype chain, as pointed out by @lucacasonato, on top of that, V8 would have to check property descriptors, ex: checking whether or not the prop is a getter. As the map is a sorted map, you can't expect a better implementation than a linear search either.
But of course, the decision for the data structure is on you and the reviewers; the bigger problem with an array would probably be how hard/easy it would be to check which index is associated with which name, for example, someone skimming by to find the 12th code would have to read through the array until they've found the 12th one, whereas a Map would have the literal number 12 right beside it.

yacinehmito:
There are also holes in the number. For example there are no error with code 2 or code 16. A `Map` will do just fine and is semantically clearer.

yacinehmito:
@lucacasonato I pushed with your suggestion. I think we should merge; discussing more would be bike-shedding.

bnoordhuis (approved):
LGTM, thanks. The CI failure is because we're not at full WPT compliance yet, right?

Apropos Map vs Object: the prototype property issue can be avoided by using `Object.create()` or `{ __proto__: null, ... }`.

Plain objects have a minor advantage over maps in that they're not vulnerable to prototype pollution (think `Map.prototype.get = Map.prototype.toString`.)

lucacasonato:
Test failure was because a WPT test was fixed in this PR, but had not been removed from the `expectFail` list. Corrected this now.

yacinehmito:
> Plain objects have a minor advantage over maps in that they're not vulnerable to prototype pollution

This is a very good point.

I feel like we are down to only two options now (roughly):
- `Object.create(null)`, then set every mapping by assigning a new property.
- Using a `Map`, but storing the original implementation of `Map.prototype.get` then putting it in the map itself (or invoking it with `.call`).

I'm not export enough to figure out which one will maximise the performance of the lookup, but this is likely to be premature optimization.

@lucacasonato I defer to you to choose between the two options, and I'll implement it.

kitsonk:
> `Object.create(null)`, then set every mapping by assigning a new property.

I would prefer this.

yacinehmito:
Done, and pushed.

op_crates/web/00_dom_exception.js
```diff
@@ -1,15 +1,47 @@
...
+  nameToCodeMapping.InvalidAccessError = 15;
+  nameToCodeMapping.TypeMismatchError = 17;
+  nameToCodeMapping.SecurityError = 18;
+  nameToCodeMapping.NetworkError = 19;
+  nameToCodeMapping.AbortError = 20;
+  nameToCodeMapping.URLMismatchError = 21;
+  nameToCodeMapping.QuotaExceededError = 22;
+  nameToCodeMapping.TimeoutError = 23;
+  nameToCodeMapping.InvalidNodeTypeError = 24;
+  nameToCodeMapping.DataCloneError = 25;
```
00ff0000red, line 25: There was a spelling error `nameToCodeMatting` => `nameToCodeMapping`
But this can be a single call anyway:
```suggestion
const nameToCodeMapping = Object.create(
  null,
  {
    IndexSizeError: { value: 1 },
    HierarchyRequestError: { value: 3 },
    WrongDocumentError: { value: 4 },
    InvalidCharacterError: { value: 5 },
    NoModificationAllowedError: { value: 7 },
    NotFoundError: { value: 8 },
    NotSupportedError: { value: 9 },
    InvalidStateError: { value: 11 },
    SyntaxError: { value: 12 },
    InvalidModificationError: { value: 13 },
    NamespaceError: { value: 14 },
    InvalidAccessError: { value: 15 },
    TypeMismatchError: { value: 17 },
    SecurityError: { value: 18 },
    NetworkError: { value: 19 },
    AbortError: { value: 20 },
    URLMismatchError: { value: 21 },
    QuotaExceededError: { value: 22 },
    TimeoutError: { value: 23 },
    InvalidNodeTypeError: { value: 24 },
    DataCloneError: { value: 25 },
  },
);
```

yacinehmito: 1. Why wrap in an object with value?
2. Yes, one can also use `Object.assign`. There are many ways to do the same thing. However I proposed writing assignments and @kitsonk approved.

yacinehmito: I didn't know about the second argument. Thanks.

yacinehmito:
@lucacasonato I think this is mergeable now.

lucacasonato (approved):
Thanks

Issue #7949, opened 2020-10-12
Bug: asm.js line numbers are incorrect; Suggestion: Disable V8's asm.js code paths


00ff0000red:
*I had originally written out a few hundred lines, attached an example module, and gave clear reasoning behind this. but had forgotten that my device's power was at 1% :)*

It seems that Deno logs completely bogus line numbers when validation fails on code ran under the "use asm" directive, can someone else confirm?
An approximately 50 line module was telling me that my error was in the 400s, yet I checked the Deno output file and didn't see anything that suggested that was possible at all.

If this is indeed a bug, I do **not** suggest fixing it, instead, it should be rightfully deprecated and ran as normal JiT / interpreted JavaScript.

It is only used as a fallback in older browsers that don't support Wasm.

It is highly unlikely that any code has been written in Deno that uses asm.js yet, so there may still be time to prevent it and/or discourage its usage.

## Update ##
Here is a quick example of an erroneous module that you should be able to copy and paste to confirm the error:
```ts
const asm_js_module = function() {
	"use asm";

	type f64 = number;

	function add(
		x: f64,
		y: f64
	) {
		x = +x;
		y = +y;

		~x; // error, what line is this on in your output file? What does Deno say?

		return x + y;
	}

	return {
		add: add
	};
}();

console.log(
	asm_js_module.add(4, 56)
);
```

00ff0000red:
@lucacasonato If I might ask since you marked it as such, what about this makes it a "breaking change"?

lucacasonato:
Thinking about this again, you're right. This isn't a breaking change. The worst it would do is potentially slow down some ASM.js code. But as you mentioned, likely no-one is using it.

PR #9019, opened 2021-01-06
fix(core): disable asm.js codegen+validation

00ff0000red:
Fixes issue https://github.com/denoland/deno/issues/7949.

V8 provides the --validate-asm and --no-validate-asm flags, the latter causes V8 to ignore the "use asm" prologue directive, preventing V8 from logging misnumbered lines to the console.

I passed the flag via cli to v8-flags and it appeared to have worked without consequence or effect.

If, instead anyone on the Deno core team deems it worth investigating and fixing, please feel free to reject this PR.

Will this pass without a test?


lucacasonato:
I don't see how this is testable so preliminary LGTM from me.

@bnoordhuis @piscisaureus do you have opinions on this?

00ff0000red:
> I don't see how this is testable so preliminary LGTM from me.

Maybe we could set up an invalid file that uses no IO, then ensure that it doesn't send anything to stdout; this would prevent the current behavior.

00ff0000red:
Added a test anyway, broke the lint, and added an *empty* file for comparison, should be good to go!

bnoordhuis:
No objections from me.

@00ff0000red Are you the author of https://bugs.chromium.org/p/v8/issues/detail?id=11280 by any chance?

ry (approved):
LGTM - I confirmed test fails on master

00ff0000red:
> Are you the author of https://bugs.chromium.org/p/v8/issues/detail?id=11280 by any chance?

That is funnily suspicious, but no, I had actually seen their post while roaming around crbugs (interesting to read about V8's bad Wasm codegen), and it had prompted me to come back to this (I'd long since forgotten that asmjs existed), but I used it as my source of info for this PR.

00ff0000red:
> LGTM - I confirmed test fails on master

Doesn't seem to work either way, should I just remove the test?

bnoordhuis:
Speaking for myself, I'm okay with landing this without a test.

ry:
landed - thanks @00ff0000red

PR #9022, opened 2021-01-06
BREAKING(std/wasi): return exit code from start

caspervonb:
This returns the exit code directly from the start entry point instead of throwing it and making the user handle that specific kind of error which is a bit too convoluted.

As a result the exit status is an implementation detail and has been made internal.


ry (approved):
LGTM

PR #9025, opened 2021-01-06
fix(coverage): do not store source inline in raw reports

caspervonb:
When we were doing single process in-memory reports getting the source from the runtime was a practical shortcut. But now that we're writing to disk this conflicts with the format tools taking raw v8 coverage dumps expect.


ry (commented):
Is there a bug being fixed here or is this just a clean up?

caspervonb:
> Is there a bug being fixed here or is this just a clean up? Not clear what the bug is or how to reproduce it.

Not really a bug per-se; just aligning our internal format with prior art.

We can dump raw v8 coverage files with `deno test --coverage=<dir>` but without this patch there's level of indirection and the source as v8 saw it is embedded in them.

With this patch the format is the same as what Node dumps with it's environment variable set, straight from v8 to disk.

ry (approved):
LGTM - thanks

PR #9033, opened 2021-01-06
fix(coverage): report partial lines as uncovered

caspervonb:
Had a little off by one error where new line was not being taken into account for line offsets. In turn this was hiding the fact that partial lines don't get reported as uncovered.

This fixes both issues.

### DIFF
--- a/cli/tools/coverage.rs
+++ b/cli/tools/coverage.rs
@@ -152,7 +152,7 @@ impl PrettyCoverageReporter {
     script_coverage: &ScriptCoverage,
     script_source: &str,
   ) {
-    let lines = script_source.lines().collect::<Vec<_>>();
+    let lines = script_source.split('\n').collect::<Vec<_>>();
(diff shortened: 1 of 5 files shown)


cli/tools/coverage.rs
```diff
@@ -162,27 +162,44 @@ impl PrettyCoverageReporter {
...
+      //
+      // We check for intersection instead of inclusion here because a block may be anywhere
+      // inside a line.
+      for function in &script_coverage.functions {
+        for range in &function.ranges {
+            if range.count > 0 {
+                continue;
+            }
 
-        line_start_offset = line_end_offset;
```
caspervonb, line 179: Yikes.

bartlomieju (changes requested):
Please add a test that ensures no regressions in the future

bartlomieju (approved):
LGTM

PR #9036, opened 2021-01-07
feat(fetch): req streaming + 0-copy resp streaming

lucacasonato:
This PR implements request body streaming when using a `ReadableStream` as the `fetch` request body. Previously using a `ReadableStream` would cause the full request body to be buffered in memory before sending. This is set up in such a way that it _will_ support full duplex HTTP streaming once hyper / reqwest are capable of this (this means you can start recieving headers / response body while the request body is still uploading). Chrome currently has the same limitation. This means that currently the request data has to be fully sent to the server, and the request body sink has to be closed before the request is recieved in JS, and the request body can be read.

This PR also changes response body streams to use 0-copy buffers instead of encoding binary data in serialized JSON strings.

For more info on request body streaming, see https://web.dev/fetch-upload-streaming.


op_crates/fetch/lib.rs
```diff
@@ -87,10 +96,10 @@ pub fn get_declaration() -> PathBuf {
   PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("lib.deno_fetch.d.ts")
 }
 
-pub async fn op_fetch<FP>(
-  state: Rc<RefCell<OpState>>,
+pub fn op_fetch<FP>(
```
lucacasonato, line 99: The split between `op_fetch` and `op_fetch_send` and the introduction of `FetchRequestResource` is unfortunate but is needed because the resource ID for the request body needs to be passed to the client synchronously _before_ awaiting the response. (ie JS needs to know where to stream the request body, _before_ getting back the response). I think the overhead in most cases will be negligible.

op_crates/fetch/lib.rs
```diff
@@ -133,28 +141,79 @@ where
...
   let mut request = client.request(method, url_);
 
-  match data.len() {
-    0 => {}
-    1 => request = request.body(Vec::from(&*data[0])),
-    _ => panic!("Invalid number of arguments"),
-  }
+  let maybe_request_body_rid = if args.has_body {
+    match data.len() {
+      0 => {
```
lucacasonato, line 151: This is an optimization for the majority of calls to fetch, where the request body is not streaming, and instead is synchronously available. In these cases the request body does not need to be passed to rust in an extra `op_fetch_request_write` and no `FetchRequestBodyResouce` needs to be created. Instead the request body is passed synchronously into Rust, like it was previously.

bartlomieju:
Supersedes https://github.com/denoland/deno/pull/8826 ?

op_crates/fetch/lib.rs
```diff
@@ -133,28 +141,79 @@ where
...
+        // If no body is passed, we return a writer for streaming the body.
+        let (tx, rx) = mpsc::channel::<Result<Vec<u8>, std::io::Error>>(1);
```
bnoordhuis, line 153:
```suggestion
        let (tx, rx) = mpsc::channel::<std::io::Result<Vec<u8>>>(1);
```
Seems to be the predominant pattern in cli/ and runtime/.

op_crates/fetch/lib.rs
```diff
@@ -133,28 +141,79 @@ where
...
+        request = request.body(Body::wrap_stream(rx));
+
+        let request_body_rid =
+          state.resource_table.add(FetchRequestBodyResource {
+            body: AsyncRefCell::new(tx),
+            cancel: Default::default(),
```
bnoordhuis, line 159:
```suggestion
            cancel: CancelHandle::default(),
```
Personal pet peeve: I hate it when I have to look up the type of something. Another instance on line 250.

But - just a suggestion!

op_crates/fetch/lib.rs
```diff
@@ -179,23 +238,61 @@ where
     }
   }
 
-  let rid = state.borrow_mut().resource_table.add(HttpBodyResource {
-    response: AsyncRefCell::new(res),
-    cancel: Default::default(),
-  });
+  let stream: BytesStream = Pin::new(Box::new(res.bytes_stream().map(|r| {
+    r.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))
+  })));
```
bnoordhuis, line 243:
```suggestion
  let stream: BytesStream = Box::pin(res.bytes_stream().map(|r| {
    r.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))
  }));
```

op_crates/fetch/lib.rs
```diff
@@ -206,30 +303,54 @@ pub async fn op_fetch_read(
...
+impl Resource for FetchRequestResource {
+  fn name(&self) -> Cow<str> {
+    "fetchRequest".into()
   }
 }
 
-struct HttpBodyResource {
-  response: AsyncRefCell<Response>,
+struct FetchRequestBodyResource {
+  body: AsyncRefCell<mpsc::Sender<Result<Vec<u8>, std::io::Error>>>,
```
bnoordhuis, line 333:
```suggestion
  body: AsyncRefCell<mpsc::Sender<std::io::Result<Vec<u8>>>>,
```
Another one on line 344.

op_crates/fetch/26_fetch.js
```diff
@@ -1282,29 +1312,36 @@
...
+        core.close(rid);
         responseBody = null;
       } else {
         responseBody = new ReadableStream({
           type: "bytes",
           async pull(controller) {
             try {
-              const result = await core.jsonOpAsync("op_fetch_read", { rid });
-              if (!result || !result.chunk) {
+              const chunk = new Uint8Array(32 * 1024);
```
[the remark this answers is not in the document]
lucacasonato: Good to know. I picked 32 * 1024, because that is what Deno.readAll defaults too, but this makes more sense indeed.

op_crates/fetch/26_fetch.js
```diff
@@ -1186,9 +1198,32 @@
...
+    );
+    if (requestBodyRid) {
+      const writer = new WritableStream({
+        async write(chunk, controller) {
+          try {
+            await opFetchRequestWrite({ rid: requestBodyRid }, chunk);
+          } catch (err) {
+            console.error(err);
+            controller.error(err);
+            controller.close();
```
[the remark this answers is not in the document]
lucacasonato: This would guards against `controller.error` failing right? `controller.error` shouldn't ever fail I think.

op_crates/fetch/lib.rs
```diff
@@ -212,30 +309,54 @@ pub async fn op_fetch_read(
...
+}
+
+impl Resource for FetchRequestBodyResource {
+  fn name(&self) -> Cow<str> {
+    "fetchRequestBody".into()
+  }
+}
+
+type BytesStream =
+  Pin<Box<dyn Stream<Item = Result<bytes::Bytes, std::io::Error>> + Unpin>>;
```
bnoordhuis, line 350:
```suggestion
  Pin<Box<dyn Stream<Item = std::io::Result<bytes::Bytes>> + Unpin>>;
```
Does it need to be Unpin?

bnoordhuis (approved):

op_crates/fetch/lib.rs
lucacasonato: Yes. I can't remember why though...

PR #9042, opened 2021-01-07
fix(runtime): use tokio for async fs ops

### LINKED ISSUE #7400: various async fs ops are not actually async
Example op_fdatasync_async

https://github.com/denoland/deno/blob/11afd071aec2d5ea6eb37a934c0835d8f6b1e764/cli/ops/fs.rs#L285-L300

This ought to be calling `tokio::task::spawn_blocking`.

op_stat_async is properly implemented 
https://github.com/denoland/deno/blob/11afd071aec2d5ea6eb37a934c0835d8f6b1e764/cli/ops/fs.rs#L864-L886

caspervonb:
This makes the following ops async:

- op_fstat_async
- op_ftruncate_async
- op_seek_async
- op_fdatasync_async
- op_fsync_async
- op_futime_async

Fixes #7400

### DIFF
--- a/runtime/ops/fs.rs
+++ b/runtime/ops/fs.rs
@@ -4,6 +4,7 @@ use super::io::std_file_resource;
 use super::io::StreamResource;
 use crate::fs_util::canonicalize_path;
 use crate::permissions::Permissions;
+use deno_core::error::bad_resource_id;
 use deno_core::error::custom_error;
 use deno_core::error::type_error;
 use deno_core::error::AnyError;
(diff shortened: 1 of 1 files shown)


runtime/ops/fs.rs
```diff
@@ -1610,24 +1665,41 @@ async fn op_futime_async(
...
+
+  let mut fs_file = RcRef::map(&resource, |r| r.fs_file.as_ref().unwrap())
+    .borrow_mut()
+    .await;
+
+  let std_file = (*fs_file)
+    .0
+    .as_mut()
+    .unwrap()
+    .try_clone()
```
caspervonb, line 1692: Not great, suggestions?

bnoordhuis: I think that's about as good as it gets when you need to turn it into a `std::fs::File`.

bnoordhuis (approved):
I'd try to factor out the repetitious code but LGTM either way.

PR #9043, opened 2021-01-07
fix(WebSocket): throw DOMException in case of protocol & fragments error

rchougule:
`WebSocket` in chrome throws `DOMException` in case of:
1. Invalid protocol passed
2. Fragments issue

Whereas we throw `SyntaxError` in the string (though the instance is of `DOMException` only)

### DIFF
--- a/op_crates/websocket/01_websocket.js
+++ b/op_crates/websocket/01_websocket.js
@@ -72,14 +72,14 @@
       if (wsURL.protocol !== "ws:" && wsURL.protocol !== "wss:") {
         throw new DOMException(
           "Only ws & wss schemes are allowed in a WebSocket URL.",
-          "SyntaxError",
+          "DOMException",
         );
       }
 
       if (wsURL.hash !== "" || wsURL.href.endsWith("#")) {
         throw new DOMException(
           "Fragments are not allowed in a WebSocket URL.",
-          "SyntaxError",
+          "DOMException",
         );
       }
 


crowlKats:
The [spec](https://html.spec.whatwg.org/multipage/web-sockets.html#dom-websocket) states:
```
If any of the values in protocols occur more than once or otherwise fail to match the requirements for elements that comprise the value of Sec-WebSocket-Protocol fields as defined by The WebSocket protocol, then throw a "SyntaxError" DOMException.
```
and
```
If urlRecord's fragment is non-null, then throw a "SyntaxError" DOMException.
```

rchougule:
> The [spec](https://html.spec.whatwg.org/multipage/web-sockets.html#dom-websocket) states:
> 
> ```
> If any of the values in protocols occur more than once or otherwise fail to match the requirements for elements that comprise the value of Sec-WebSocket-Protocol fields as defined by The WebSocket protocol, then throw a "SyntaxError" DOMException.
> ```
> 
> and
> 
> ```
> If urlRecord's fragment is non-null, then throw a "SyntaxError" DOMException.
> ```
> 
> so the current behaviour is correct.

Thanks for pointing in the correct direction @crowlKats.

Just from an understanding point of view, so does it mean that chrome isn't following it correctly?
image

00ff0000red:
```js
new DOMException("procol error", "SyntaxError");
```
Chrome is correct, `DOMException`'s second argument should be a "SyntaxError" string.

crowlKats:
Nope, it is correct:
Screenshot 2021-01-07 at 19 35 07

The console is reporting the class that was thrown, not the domexception kind

lucacasonato:
Closing as we are doing the correct thing (Chrome is too).

rchougule:
Thanks for the clarification @lucacasonato @crowlKats @00ff0000red. Learnt something new today :-)

PR #5034, opened 2020-05-01
BREAKING: remove window.location and self.location

bartlomieju:
This commit removes "location" global available on "window",
"globalThis" and "self".

Closes #3520

PR #9044, opened 2021-01-07
fix(op_crates/web): Strip username and password from location

nayeemrmn:


op_crates/web/12_location.js
```diff
@@ -10,6 +10,9 @@
         throw new TypeError("Illegal constructor.");
       }
       const url = new URL(href);
+      url.username = "";
+      url.password = "";
+      console.trace(url.href);
```
lucacasonato, line 15: stray debug statement here :-)

nayeemrmn:
```suggestion
```

bartlomieju:
Maybe we could use same code that was removed in https://github.com/denoland/deno/pull/5034?

nayeemrmn:
> Maybe we could use same code for `location` that was removed in #5034?

What's the significance of that implementation? The differences from that one are purposeful.

bartlomieju:
> > Maybe we could use same code for `location` that was removed in #5034?
> 
> What's the significance of that implementation? The differences from that one are purposeful.

There's none; I just thought it might be useful.

nayeemrmn:
Closing in favour of #9084.

PR #9045, opened 2021-01-07
feat(runtime/permissions): prompt fallback

### LINKED ISSUE #3811: discussion: Bring back permission prompt behind a flag
Before https://github.com/denoland/deno/pull/3183 Deno was prompting for permission if one was missing. This functionality could have been disabled using `--no-prompt` flag.
After that patch, an error is always thrown if permission is missing, the only way to prompt for permission is to use `permissions.request()` JS API. 

IMHO `prompt` mode is very useful and can be used when first time running the script to quickly go through permissions required instead of hitting consecutive permission errors.

Most of the code for this functionality still lives in `cli/permissions.rs` and I propose to bring this mode back behind a `--prompt` flag.

CC @kt3k @kevinkassimo @ry

crowlKats:
This PR adds the `prompt` flag which causes non-passed permissions to prompt instead of failing

Closes #3811

cc @bartlomieju


bartlomieju:
@crowlKats great! Can you bring back some of the tests for this functionality that were there previously?

crowlKats:
problem is that i am changing `check` and requesting the prompt there, but check is used by resolved_and_display_path, which then again is used by request_read which prompts as well, causing multiple prompts

runtime/permissions.rs
```diff
@@ -566,18 +671,27 @@ impl Permissions {
...
+    match self
       .query_net(&Some(&(hostname, url.port_or_known_default())))
       .check(
         &format!("network access to \"{}\"", display_host),
         "--allow-net",
-      )
+        self.prompt,
+      ) {
+      Ok(PromptResult::AllowAlways) => {
+        self.net.granted_list.insert(display_host); // TODO
```
crowlKats, line 682: not sure what the right value here would be

crowlKats:
Closing in favour of #9376

PR #9046, opened 2021-01-07
tests: enable wpt for url

lucacasonato:
This enables all suites testing "url" in WPT. Out of 1728 tests, 360 fail.

cc @nayeemrmn


lucacasonato:
This is not ready to merge, because the WPT server is not shut down correctly on test success or failure. Zombie processes stay running.

bnoordhuis:
The good news: it seems to work on all platforms.

The bad news: the WPT tests take so long they time out on the windows-2019 buildbot...
```
test web_platform_tests ... test web_platform_tests has been running for over 60 seconds
Error: The operation was canceled.
```
60 seconds? More like 60 minutes.

I've pushed a patch that disables the WPT tests on that buildbot.

.github/workflows/ci.yml
```diff
@@ -224,6 +224,16 @@ jobs:
...
+      - name: Configure hosts file for WPT (unix)
+        if: runner.os != 'Windows'
+        run: ./wpt make-hosts-file | sudo tee -a /etc/hosts
+        working-directory: test_util/wpt/
+
+      - name: Configure hosts file for WPT (windows)
+        if: runner.os == 'Windows'
+        working-directory: test_util/wpt/
+        run: python wpt make-hosts-file | Out-File $env:SystemRoot\System32\drivers\etc\hosts -Encoding ascii -Append
+
```
bnoordhuis, line 236: If you're okay with my last commit, then this can be removed:
```suggestion
```
(I'm also okay with leaving it in. Maybe we can enable a subset.)

bnoordhuis (approved):

cli/tests/integration_tests.rs
```diff
@@ -5314,6 +5329,13 @@ fn web_platform_tests() {
     if line.contains("web-platform.test:9000") {
       ready_9000 = true;
     }
+<<<<<<< HEAD
```
satyarohith, line 5332: (cc @lucacasonato)

lucacasonato:
@bnoordhuis PTAL again. I managed to enable the tests on Windows now.

bnoordhuis (approved):

PR #9047, opened 2021-01-07
test(cli): enable web platform timer tests

caspervonb:
Part of #9001


caspervonb:
Unfortunately most of the interesting tests are in html documents.

lucacasonato:
> Unfortunately most of the interesting tests are in html documents, so without attempting to parse them out into bundles (would have to be opt-in of course) this is the best we can get.

Can you try upstream this?

caspervonb:
> > Unfortunately most of the interesting tests are in html documents, so without attempting to parse them out into bundles (would have to be opt-in of course) this is the best we can get.
> 
> Can you try upstream this?

Yup

lucacasonato (approved):
Thanks

PR #9065, opened 2021-01-09
test(wpt): Enable wpt workers tests

yacinehmito:
In Web Crypto, must test suites have their JavaScript entrypoint suffixed with `.worker.js`. These files were by default ignored by Deno's integration tests; now they are included.

JavaScript entrypoints meant to be included in HTML files reference their dependencies with `META` comments. JavaScript entrypoints for workers don't do that: they call a function `importScripts()`.
We therefore had to look for the patterns `importScripts("<path>");` in order to build the bundle for the test, as well as included in the bundle the function `importScripts()` as a no-op.

cc @lucacasonato

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -5094,6 +5094,16 @@ fn concat_bundle(
   let mut bundle_line_count = init.lines().count() as u32;
   let mut source_map = sourcemap::SourceMapBuilder::new(Some(&bundle_url));
 
+  // In classic workers, `importScripts()` performs an actual import.
+  // However, we don't implement that function in Deno as we want to enforce
+  // the use of ES6 modules.
+  // To work around this, we:
+  // 1. Define `importScripts()` as a no-op (code below)
+  // 2. Capture its parameter from the source code and add it to the list of
+  // files to concatenate. (see `web_platform_tests()`)
+  bundle.push_str("function importScripts() {}\n");
+  bundle_line_count += 1;
+
   for (path, text) in files {
     let path = std::fs::canonicalize(path).unwrap();
     let url = url::Url::from_file_path(path).unwrap().to_string();
@@ -5189,7 +5199,9 @@ fn web_platform_tests() {
       .filter(|e| e.file_type().is_file())
       .filter(|f| {
         let filename = f.file_name().to_str().unwrap();
-        filename.ends_with(".any.js") || filename.ends_with(".window.js")
+        filename.ends_with(".any.js")
+          || filename.ends_with(".window.js")
+          || filename.ends_with(".worker.js")
       })
       .filter_map(|f| {
         let path = f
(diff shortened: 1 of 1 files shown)


cli/tests/integration_tests.rs
```diff
@@ -5082,6 +5082,8 @@ fn concat_bundle(
     bundle_line_count += 1;
   }
 
+  bundle.push_str("function importScripts() {}\n");
```
lucacasonato, line 5085:
```suggestion
  bundle.push_str("function importScripts() {}\n");
  bundle_line_count += 1;
```

source mapping wont be correct otherwise.

lucacasonato (commented):
`importScripts()` is actually web standard. it is a function that is used to import things in classic mode workers (which we dont support). This is pretty hacky, but I guess it will work for now.

yacinehmito:
> `importScripts()` is actually web standard. it is a function that is used to import things in classic mode workers (which we dont support). This is pretty hacky, but I guess it will work for now.

I will add a comment in the code stating that, and will open an issue for the lack of `importScripts()` if I don't find an existing one.

lucacasonato:
You can open an issue, but we won't add support for "classic" workers. We are ESM only. Deno will only support "module" type workers.

caspervonb:
Looking at the actual tests, e.g digest https://github.com/web-platform-tests/wpt/tree/a51d36327febebc8634d7b523a7e620f4ebdbf26/WebCryptoAPI/digest

Looks like this can be fixed upstream to be an "any" test instead.

All the variants do is import the shared script and run it.

yacinehmito:
Oh I see. I won't bother then. Thanks for the clarification.

yacinehmito:
@caspervonb I think we might have a hard time pushing changes upstream in timely manner. There are no owners at the moment for the Web Crypto API tests. Also, my guess is that it is suffixed by `worker.js` because it only works in a worker context (which is indeed the case per the call of `importScripts()`).

caspervonb:
> @caspervonb I think we might have a hard time pushing changes upstream in timely manner. There are no owners at the moment for the Web Crypto API tests. Also, my guess is that it is suffixed by `worker.js` because it only works in a worker context (which is indeed the case per the call of `importScripts()`).

Hmm not sure about that but I opened two patches yesterday and both got merged within a couple of hours. But we won't know unless we try 🙃

yacinehmito:
> Hmm not sure about that, I opened two patches yesterday and both got merged within a couple of hours. 

I meant on the Web Crypto API tests specifically.
They are looking for someone: https://github.com/web-platform-tests/wpt/issues/27111
And this PR of mine is still hanging (although it's not very old): https://github.com/web-platform-tests/wpt/pull/27105

cli/tests/integration_tests.rs
yacinehmito: The linter is warning me that `bundle_line_count` is not read after that line.
I'm putting the function declaration at the very top of the bundle then.

yacinehmito:
New changes. Everything should be good now:
- Added comments
- Moved `importScripts` no-op at the beginning of the bundle
- Ignoring imports of `testharness.js`, as it seems to override some important functions, leading to the removal of the test report in the terminal.

lucacasonato (approved):
LGTM. Thanks

PR #9016, opened 2021-01-05
fix(op_crates/crypto): Test crypto.getRandomValues() with wpt

yacinehmito:
Fixed `crypto.getRandomValues()` so that it passes the tests from web-platform-tests. The core logic doesn't really change: it just adds validation that is expected in the [W3C recommandation](https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues).

This follows #8990 and precedes #8984.

PR #9067, opened 2021-01-09
fix(op_crates/crypto): Prefix op of getRandomValues() by op_crypto

yacinehmito:
It is preferable for all Rust ops in `crypto` to be prefixed by `op_crypto`. As requested by @lucacasonato in #8984, we are renaming `op_get_random_values` so that if fits the naming convention.


yacinehmito:
PR #9016 needs to be merged first.

bnoordhuis (approved):

yacinehmito:
Rebased on master. It should be mergeable now.

PR #9068, opened 2021-01-09
test(cli): add wpt url tests

kt3k:
This PR adds web platform "url" tests. Commented out test cases which cause error when loading the test harness (`percent-encoding.window`, `toascii.window`, `url-constructor.any`, etc). `urlencoded-parser.any` and `url-setters-stripping.any` have many test failures. Probably we need to fix actual bugs of the current URL implementation to pass these in future.

ref: #9001

### DIFF
--- a/cli/tests/wpt.jsonc
+++ b/cli/tests/wpt.jsonc
@@ -173,5 +173,26 @@
     "measure-l3",
     "structured-serialize-detail",
     "user_timing_exists"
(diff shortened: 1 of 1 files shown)


lucacasonato:
@kt3k duplicate of #9046? :-)

kt3k:
@lucacasonato Sorry! I missed it. Closing.

PR #9069, opened 2021-01-09
feat(op_crates/crypto): Implement crypto.subtle.digest()

yacinehmito:
Part of solving #1891.

Requires #8984.

### DIFF
--- a/op_crates/crypto/01_crypto.js
+++ b/op_crates/crypto/01_crypto.js
@@ -9,7 +9,7 @@
   function getRandomValues(arrayBufferView) {
     if (!ArrayBuffer.isView(arrayBufferView)) {
       throw new TypeError(
-        "Argument 1 does not implement interface ArrayBufferView",
+        "Argument 1 does not implement the interface ArrayBufferView",
       );
     }
     if (
@@ -81,7 +81,21 @@
     }
   }
 
-  const supportedAlgorithms = {};
+  const supportedAlgorithms = {
+    digest: new RegisteredAlgorithmsContainer({
+      "SHA-1": {},
+      "SHA-256": {},
+      "SHA-384": {},
+      "SHA-512": {},
+    }),
+  };
+
+  const digestByteLengths = {
+    "SHA-1": 20,
+    "SHA-256": 32,
+    "SHA-384": 48,
+    "SHA-512": 64,
+  };
 
   function normalizeAlgorithm(algorithm, registeredAlgorithms) {
     let alg;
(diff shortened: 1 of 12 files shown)


op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,206 @@
...
+        throw new TypeError("Algorithm name is missing or not a string");
+      }
+      const algorithmName = registeredAlgorithms.normalizeName(algorithm.name);
+      if (algorithmName === undefined) {
+        throw new DOMException(
+          "Unrecognized algorithm name",
+          "NotSupportedError",
+        );
+      }
+      const definition = registeredAlgorithms.getDefinition(algorithmName);
```
bnoordhuis, line 109: Can be combined into a single step, right? `getDefinition()` normalizes the name before lookup, and you throw `DOMException` when the return value is undefined.

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,206 @@
...
+      await Promise.resolve();
+      throw new Error("Not implemented");
+    },
+    async digest(algorithm, data) {
+      const alg = normalizeAlgorithm(algorithm, supportedAlgorithms.digest);
+      // TODO: validate that data is indeed a BufferSource
+      // Note: we need to copy the data because the specs require it.
+      // It's unclear whether we are better doing it here or on the Rust side.
+      const input = data.slice(0);
+      const output = new ArrayBuffer(digestByteLengths[alg.name]);
```
bnoordhuis, line 143: Maybe it's a good idea to merge `digestByteLengths` into `supportedAlgorithms.digest`? Why have two hashmaps instead of one?

yacinehmito: 1. `supportedAlgorithms` follows the spec recommendation to the letter
2. `digestByteLengths` will have other uses. For example when importing an HMAC key.

op_crates/crypto/01_crypto.js
yacinehmito: That's how it was originally, but the spec mentions that the name must be the normalised one, not the original one.
So, I need to get both the normalised name _and_ the definition.
(Also, `getDefinition()` does not normalise the name before lookup 🤔)

op_crates/crypto/01_crypto.js
```diff
@@ -1,25 +1,206 @@
...
+      await core.jsonOpAsync(
+        "op_crypto_subtle_digest",
+        alg,
+        // TODO: Check that chugging the input and output directly to the
+        // constructor of Uint8Array doesn't cause issues of length, nor
+        // does it perform any useless data copy.
+        new Uint8Array(input),
```
devcat, line 150: `new TypedArray(TypedArray)` copies unlike `new TypedArray(ArrayBuffer)`
length issues will only happen if you pass non u8 TypedArray, but it can be easily fixed by operating on underlaying ArrayBuffer
```suggestion
        new Uint8Array(input.buffer, input.byteOffset, input.byteLength),
```

given that spec allows passing ArrayBuffer you probably will want to do something like this

```js
function validateBufferSource(data) {
  if (data instanceof ArrayBuffer) return new Uint8Array(data);
  if (ArrayBuffer.isView(data)) return new Uint8Array(data.buffer, data.byteOffset, data.byteLength);

  throw new TypeError("The provided value is not of type '(ArrayBuffer or ArrayBufferView)'");
}

const u8array = validateBufferSource(data); // .slice() for copy
```

mimiza:
I'm so happy to see webcrypto being implemented into Deno

op_crates/crypto/01_crypto.js
yacinehmito: Thank you very much for the info! It was very helpful.

laughedelic:
Please, don't close it

bartlomieju:
Let's continue in #10796

PR #9079, opened 2021-01-10
refactor(op_crates/websocket): refactor event loop

crowlKats:

### DIFF
--- a/op_crates/websocket/01_websocket.js
+++ b/op_crates/websocket/01_websocket.js
@@ -294,59 +294,78 @@
     }
 
     async #eventLoop() {
-      if (this.#readyState === OPEN) {
+      while (this.#readyState === OPEN) {
         const message = await core.jsonOpAsync(
           "op_ws_next_event",
           { rid: this.#rid },
         );
-        if (message.type === "string" || message.type === "binary") {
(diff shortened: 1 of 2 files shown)


op_crates/websocket/01_websocket.js
```diff
@@ -294,59 +294,78 @@
     }
 
     async #eventLoop() {
-      if (this.#readyState === OPEN) {
```
ry, line 297: This seems like a behavioral change - is something being fixed here?

crowlKats: no behaviour is identical. no fixes, just a cleanup i have been meaning to do for a while

ry: Why does this if statement change to a while?

crowlKats: as mentioned, it was recursive calls. that if was the guard if to get the next. since i changed to a while loop and the calls are not recursive anymore, it is equivalent. `this.#readyState` gets only changed if the ws closes or errors, so the loop would be exited by that changing

ry (approved):
LGTM - thanks crowlKats

PR #9081, opened 2021-01-10
update copyright to 2021

ry:


waldyrious:
@ry while you're at it, would you consider renaming the `LICENSE` file to `LICENSE.md`? It's the only project metadata document that's [stored as a plaintext file](https://github.com/denoland/deno/search?l=Text) in the repo, and simply renaming the file makes the content more readable on GitHub, even without adding markup (although that would also be nice, e.g. highlighting the license title as a heading).

lucacasonato (approved):

PR #9088, opened 2021-01-11
fix(cli): dispatch unload on exit

### LINKED ISSUE #3603: Clean up on Deno.exit()
`window.onunload` and `window.addEventListener "unload"` work fine for when a script normally ends. But when
- an uncaught exception is thrown
- `Deno.exit()` is called
- (process is stopped, e.g. SIGINT - should be fixed with #2339)

it seems, there is no way to do any clean up action. Is this intended? 

Thanks and much love to all contributors.

kt3k:
This PR adds the dispatching of `unload` event before sending op_exit. Closes #3603.

(The issue also suggests it should call `unload` hooks at signals like `SIGINT`, etc, but that seems very difficult (complex) to implement because if the process attach something to signal, the default behavior is lost. So we need to simulate the default behavior for each signal to do that, but that seems unreasonably complex to implement and maintain.)

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -2632,7 +2632,7 @@ itest!(_077_fetch_empty {
 });
 
 itest!(_078_unload_on_exit {
-  args: "run -A 078_unload_on_exit.ts",
+  args: "run 078_unload_on_exit.ts",
   output: "078_unload_on_exit.ts.out",
 });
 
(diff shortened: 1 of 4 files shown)


satyarohith (commented):
Looks good!

cli/tests/integration_tests.rs
```diff
@@ -2631,6 +2631,11 @@ itest!(_077_fetch_empty {
   exit_code: 1,
 });
 
+itest!(_078_unload_on_exit {
+  args: "run -A 078_unload_on_exit.ts",
```
satyarohith, line 2635: Can `-A` be removed?

kt3k: Right. Thanks!

ry (approved):
LGTM - thanks @kt3k

Issue #8571, opened 2020-12-01
Reenable flaky watch tests


bartlomieju:
In #8557 following tests were disabled due to high flakiness when running CI:
- `fmt_watch_test`
- `bundle_js_watch`
- `bundle_watch_not_exit`
- `run_watch`
- `run_watch_not_exit`
- `run_watch_with_importmap_and_relative_paths`

They should be rewritten and enabled.

magurotuna:
CI logs when it failed due to the watcher-related tests:

- https://github.com/denoland/deno/actions/runs/392331654
- https://github.com/denoland/deno/actions/runs/389935305

magurotuna:
Still seems flaky
https://github.com/denoland/deno/runs/1542414977#step:24:4089

magurotuna:
ref #9306 

@bartlomieju Maybe it's okay to close this issue?

bartlomieju:
Yes, thanks

PR #9091, opened 2021-01-11
fix(watcher): keep working even when imported files has invalid syntax

### LINKED ISSUE #9008: Parsing errors cause --watch to stop
When running a TS project with `--watch`, whenever I get a syntax error that is not caused by TSC (I think these are SWC), the file watcher never restarts on the next file change.

Example error:
```
error: Unexpected token `]`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier at {my project file here}
Watcher Process finished! Restarting on file change...
```

Is that an SWC error?
I wanted to say that this could be related to SWC issue https://github.com/swc-project/swc/pull/1274, but the watcher still was able to log to stdout that it was finished, so maybe it's Deno?

(*Currently, I just restart Deno when it happens, but it beats the purpose of the file watcher in the first place*)

___

If it helps, I'm using Deno 1.6.3 on an x86_64 GNU/Linux 19-Ubuntu.
Other flags enabled:
--unstable
--lock='...'
--lock-write
--allow-all
--v8-flags='...'

___

Update:
Added an import-map and whenever it was edited, if it caused a syntax error, then the same effect as observed above ^ occurred again.

magurotuna:
Fixes #9008 

In this PR, I have added a test that works fine on my machine. But unfortunately, because the test cases related to watcher functionality are quite flaky (ref #8571),  this test is marked as `#[ignore]`. This means the test I added here won't be executed in CI.

### DIFF
--- a/cli/file_watcher.rs
+++ b/cli/file_watcher.rs
@@ -1,4 +1,4 @@
-// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 use crate::colors;
 use core::task::{Context, Poll};
--- a/cli/file_watcher.rs
+++ b/cli/file_watcher.rs
@@ -172,7 +172,7 @@ where
   let mut debounce = Debounce::new();
   // Store previous data. If module resolution fails at some point, the watcher will try to
   // continue watching files using these data.
-  let mut paths;
+  let mut paths = Vec::new();
   let mut module = None;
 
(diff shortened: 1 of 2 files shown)


piscisaureus (approved):

piscisaureus:
It is indeed unfortunate that this does not get tested in CI, but it seems to work for me locally.
Thanks for the fix, @magurotuna!

PR #9101, opened 2021-01-12
feat: Stabilize Deno.shutdown()

### LINKED ISSUE #9099: Remove `how` parameter from Deno.shutdown()
There are no protocols that meaningfully support a shutdown mode other than ~`SHUT_RD`~ `SHUT_WR`, so this parameter was never too useful to begin with. Tokio 1.0 has dropped the parameter entirely; Deno should do this too. 

Once this is done, we can also stabilize `Deno.shutdown()`.

bartlomieju:
Closes #9099

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1809,7 +1809,7 @@ declare namespace Deno {
    * ```ts
    * const listener = Deno.listen({ port: 80 });
    * const conn = await listener.accept();
-   * Deno.shutdown(conn.rid, Deno.ShutdownMode.Write);
+   * Deno.shutdown(conn.rid);
    * ```
    */
   export function shutdown(rid: number): Promise<void>;
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1805,7 +1805,7 @@ declare namespace Deno {
    */
   export function connectTls(options: ConnectTlsOptions): Promise<Conn>;
 
-  /** Shutdown socket send and receive operations.
+  /** Shutdown socket send operations.
    *
    * Matches behavior of POSIX shutdown(3).
    *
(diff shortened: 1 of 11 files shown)


ry (dismissed):
Why not have closeWrite? I'd rather remove Deno.shutdown...

bartlomieju:
> Why not have closeWrite? I'd rather remove Deno.shutdown...

As per @piscisaureus in #9099:
> There are no protocols that meaningfully support a shutdown mode other than SHUT_RD, so this parameter was never too useful to begin with. Tokio 1.0 has dropped the parameter entirely; Deno should do this too.

Tokio has no way of shutting down only the write half.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -1809,6 +1802,18 @@ declare namespace Deno {
...
   export function connectTls(options: ConnectTlsOptions): Promise<Conn>;
 
+  /** Shutdown socket send and receive operations.
+   *
+   * Matches behavior of POSIX shutdown(3).
+   *
+   * ```ts
+   * const listener = Deno.listen({ port: 80 });
+   * const conn = await listener.accept();
+   * Deno.shutdown(conn.rid, Deno.ShutdownMode.Write);
```
satyarohith, line 1812:
```suggestion
   * Deno.shutdown(conn.rid);
```

bartlomieju:
> > Why not have closeWrite? I'd rather remove Deno.shutdown...
> 
> As per @piscisaureus in #9099:
> 
> > There are no protocols that meaningfully support a shutdown mode other than SHUT_RD, so this parameter was never too useful to begin with. Tokio 1.0 has dropped the parameter entirely; Deno should do this too.
> 
> Tokio has no way of shutting down only the write half.

Actually now I'm confused reading https://docs.rs/tokio/1.0.1/tokio/io/trait.AsyncWriteExt.html#method.shutdown

> For example, the TcpStream implementation will issue a shutdown(Write) sys call.

piscisaureus:
My mistake. The only mode supported by TCP and Tokio is SHUT_WR.

piscisaureus (approved):

PR #9102, opened 2021-01-12
fix: jsonOpAsync for non object args

lucacasonato:
This fixes two bugs:

1. errors in the synchronous part of jsonOpAsync on the rust side were not propagated to JS correctly
2. only JS objects could be used as arguments. Now primitives like string, number, boolean, or array are also supported

cc @cab @bartlomieju

### DIFF
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -10,7 +10,6 @@ use crate::BufVec;
 use crate::ZeroCopyBuf;
 use futures::Future;
 use indexmap::IndexMap;
-use serde::Deserialize;
 use serde_json::json;
 use serde_json::Value;
 use std::cell::RefCell;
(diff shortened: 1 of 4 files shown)


core/ops.rs
```diff
@@ -218,16 +219,19 @@ where
   F: Fn(Rc<RefCell<OpState>>, Value, BufVec) -> R + 'static,
   R: Future<Output = Result<Value, AnyError>> + 'static,
 {
+  #[derive(Deserialize)]
+  #[serde(rename_all = "camelCase")]
+  struct Params {
+    promise_id: u64,
+    args: Value,
+  }
```
bartlomieju, line 227: IIRC this was removed by @piscisaureus for the performance reasons

lucacasonato: Can you elaborate? We could change this to a JSON array. I think that would have less overhead than the previous impl.

bartlomieju: The idea was that getting only `promise_id` was faster than trying to do deserialization into struct here and then once again in the actual op. Maybe we should guard arguments passed to `jsonOpAsync` (and sync as well) in `core.js`?

lucacasonato: If that is the case, array should be just as fast. Let me implement that.

lucacasonato:
Closing in favor of https://github.com/denoland/deno/pull/9423

PR #9103, opened 2021-01-12
refactor(op_crate/fetch): align streams to spec

### LINKED ISSUE #8814: Hanging promise in TransformStream impl
Reproduction:

```js
async function collect(o) {
    const buffer = [];
    const reader = o.getReader();
    while(true){
        const { value , done  } = await reader.read();
        if (done) {
            return buffer;
        }
        buffer.push(value);
    }
}
function concat(...os) {
    const { writable , readable  } = new TransformStream();
    (async function() {
        for (const o of os){
            await o.pipeTo(writable, {
                preventClose: true
            });
        }
        writable.getWriter().close();
    })();
    return readable;
}
const EOF = Symbol();
function external() {
    let next;
    const observable = new ReadableStream({
        start (controller) {
            next = (v)=>{
                if (v === EOF) {
                    return controller.close();
                }
                controller.enqueue(v);
            };
        }
    });
    return {
        observable,
        next: next
    };
}
function range(start, end) {
    const { observable , next  } = external();
    const len = Math.abs(end - start);
    const dir = Math.sign(end - start);
    for(let i = 0; i <= len; i++){
        next(start + i * dir);
    }
    next(EOF);
    return observable;
}
async function main() {
    const list = collect(concat(range(1, 2)));
    console.log(await list, [
        1,
        2,
        3,
        4
    ]);
}
await main();
```

The isolate crashes with `error: Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promise.` at `o.pipeTo` call. Looking at the runtime implementation, a deferred promise is created (and returned), that is not resolved (line 1881 in `op_crates/fetch/11_stream.js`). This results in a hanging promise, breaking the isolate.

cc @kitsonk

kitsonk:
Fixes #8814

~There is a little bit more work to do on this, but getting close.~


cli/tests/wpt.jsonc
```diff
@@ -1,186 +1,241 @@
...
-    "console-is-a-namespace",
-    "console-label-conversion",
-    "console-namespace-object-class-string",
-    "console-tests-historical",
-  ],
-  "WebCryptoApi": [
-    "getRandomValues"
-  ]
-}
+  // "encoding": [
```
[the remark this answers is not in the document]
kitsonk: oh yeah, sorry...  I forgot I did that again... meant to do it before I even pushed.

op_crates/fetch/11_streams.js
```diff
@@ -1,1443 +1,703 @@
...
+    if (typeof v !== "number") {
+      return false;
+    }
+    if (Number.isNaN(v)) {
+      return false;
+    }
+    if (v < 0) {
+      return false;
     }
+    return true;
```
ry, line 605: return 0 <= v

op_crates/fetch/11_streams.js
```diff
@@ -1,1443 +1,703 @@
...
+
+  /**
+   * @param {ReadableStream} stream
+   * @returns {boolean}
+   */
+  function isReadableStreamLocked(stream) {
+    if (stream[_reader] === undefined) {
+      return false;
     }
+    return true;
```
ry, line 625: return stream[_reader] !== undefined

op_crates/fetch/11_streams.js
```diff
@@ -1,1443 +1,703 @@
...
-    }
-    if (typeof size !== "function") {
-      throw new TypeError("size must be callable.");
+    if (stream[_writer] === undefined) {
+      return false;
     }
-    return (chunk) => {
-      return size.call(undefined, chunk);
-    };
+    return true;
```
ry, line 663: ditto

op_crates/fetch/11_streams.js
kitsonk: Varying from the spec makes readability issues when checking you are aligned. The spec specifies the logic this way. 😢 For alignment purposes, I would prefer to follow the spec.

ry (approved):
LGTM - let's land this and do follow up work in master branch. Great work!

PR #9105, opened 2021-01-13
Remove unnecessary boxing of tokio::time::Sleep

piscisaureus:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -431,6 +431,7 @@ dependencies = [
  "notify",
  "os_pipe",
  "percent-encoding",
+ "pin-project 1.0.4",
  "regex",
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM

bnoordhuis (approved):
LGTM with a question: is the `pin_project` dependency necessary or convenient?

piscisaureus:
> LGTM with a question: is the `pin_project` dependency necessary or convenient?

Somewhere in between it can be avoided but then we would have to manually add unsafe code to to do the projection.
Note that deno_core also depends on `pin_project`. Tokio seems to use `pin_project_lite`, which might be worth looking into.

PR #9106, opened 2021-01-13
fix: align DOMException API to the spec and add web platform testing of it.

kt3k:
It is part of #9001.

This PR adds web platform testing of DOMException API and align our implementation to pass it.


kt3k:
This passes WPT suites, but an uncaught DOMException doesn't show the stack trace now. I'm investigating the handling of uncaught error in `/core`.

kt3k:
Restored the original behavior when thrown as uncaught error, but as a result I have to skip the case `"does not inherit from Error: class-side"` which I don't know how to pass for now.

Now it's ready for review.

ry (approved):
LGTM

kt3k:
Thank you for your review!

PR #9111, opened 2021-01-14
refactor: make Process#kill() on Windows throw sensible errors

piscisaureus:
Previously, calling `Process#kill()` after the process had exited would
sometimes throw a `TypeError` on Windows. After this patch, it will
throw `NotFound` instead, just like other platforms.

This patch also fixes flakiness of the `runKillAfterStatus` test on
Windows.


ry (approved):
LGTM

PR #9115, opened 2021-01-14
benchmark: cleanup serde_json values being passed around

wperron:
Replace `serde_json::value::Value` being passed around a lot in the benchmarks.

### DIFF
--- a/cli/bench/main.rs
+++ b/cli/bench/main.rs
@@ -9,7 +9,6 @@ use std::{
   path::PathBuf,
   process::{Command, Stdio},
 };
-use test_util::StraceOutput;
 
 mod http;
 mod throughput;
@@ -317,8 +316,8 @@ fn run_strace_benchmarks(
 ) -> Result<()> {
   use std::io::Read;
 
-  let mut thread_count: HashMap<String, StraceOutput> = HashMap::new();
-  let mut syscall_count: HashMap<String, StraceOutput> = HashMap::new();
+  let mut thread_count: HashMap<String, u64> = HashMap::new();
+  let mut syscall_count: HashMap<String, u64> = HashMap::new();
 
   for (name, args, _) in EXEC_TIME_BENCHMARKS {
     let mut file = tempfile::NamedTempFile::new()?;
@@ -340,8 +339,8 @@ fn run_strace_benchmarks(
     file.as_file_mut().read_to_string(&mut output)?;
 
     let strace_result = test_util::parse_strace_output(&output);
-    let clone = strace_result.get("clone").unwrap().clone();
-    let total = strace_result.get("total").unwrap().clone();
+    let clone = strace_result.get("clone").map(|d| d.calls).unwrap_or(0);
+    let total = strace_result.get("total").unwrap().calls;
     thread_count.insert(name.to_string(), clone);
     syscall_count.insert(name.to_string(), total);
   }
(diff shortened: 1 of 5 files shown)


cli/bench/main.rs
```diff
@@ -244,18 +245,18 @@ fn get_binary_sizes(target_dir: &PathBuf) -> Result<Value> {
...
 }
 
 const BUNDLES: &[(&str, &str)] = &[
   ("file_server", "./std/http/file_server.ts"),
   ("gist", "./std/examples/gist.ts"),
 ];
-fn bundle_benchmark(deno_exe: &PathBuf) -> Result<Value> {
-  let mut sizes = Map::new();
+fn bundle_benchmark(deno_exe: &PathBuf) -> Result<HashMap<String, u64>> {
+  let mut sizes: HashMap<String, u64> = HashMap::new();
```
lucacasonato, line 259: This can also be written like below. This is a little less verbose IMO.

```suggestion
  let mut sizes = HashMap::<String, u64>::new();
```

cli/bench/main.rs
```diff
@@ -275,71 +276,48 @@ fn bundle_benchmark(deno_exe: &PathBuf) -> Result<Value> {
...
   }
 
-  Ok(Value::Object(sizes))
+  Ok(sizes)
 }
 
-fn run_throughput(deno_exe: &PathBuf) -> Result<Value> {
-  let mut m = Map::new();
+fn run_throughput(deno_exe: &PathBuf) -> Result<HashMap<String, f64>> {
+  let mut m: HashMap<String, f64> = HashMap::new();
```
lucacasonato, line 287: Ditto

cli/bench/main.rs
```diff
@@ -275,71 +276,48 @@ fn bundle_benchmark(deno_exe: &PathBuf) -> Result<Value> {
...
   deno_exe: &PathBuf,
-  new_data: &mut Map<String, Value>,
+  new_data: &mut BenchResult,
 ) -> Result<()> {
   use std::io::Read;
 
-  let mut thread_count = Map::new();
-  let mut syscall_count = Map::new();
+  let mut thread_count: HashMap<String, u64> = HashMap::new();
+  let mut syscall_count: HashMap<String, u64> = HashMap::new();
```
lucacasonato, line 320: Ditto

cli/bench/main.rs
```diff
@@ -361,26 +339,20 @@ fn run_strace_benchmarks(
...
+  new_data.thread_count = thread_count;
+  new_data.syscall_count = syscall_count;
 
   Ok(())
 }
 
-fn run_max_mem_benchmark(deno_exe: &PathBuf) -> Result<Value> {
-  let mut results = Map::new();
+fn run_max_mem_benchmark(deno_exe: &PathBuf) -> Result<HashMap<String, u64>> {
+  let mut results: HashMap<String, u64> = HashMap::new();
```
lucacasonato, line 355: Ditto

lucacasonato (commented):
Looks great, just some minor comments.

ry (commented):
Much cleaner - thanks! 

We should be careful when landing this that the output hasn't changed in some subtle way. Otherwise the deno.land/benchmarks page may break. Unfortunately we don't have tests that verify the structure. I'm not sure how we can check this tho.

wperron:
> Much cleaner - thanks!
> 
> We should be careful when landing this that the output hasn't changed in some subtle way. Otherwise the deno.land/benchmarks page may break. Unfortunately we don't have tests that verify the structure. I'm not sure how we can check this tho.

that's a good point, maybe we could add a temporary ci job to push the results of the benchmarks from this PR to a different branch in the `benchmark_data` repo, and then rebase a couple of times onto master to compare, and remove that temporary job before merging, wdyt?

ry (approved):
LGTM - let's just land this and keep an eye on the benchmarks to make sure they don't break. If they do we'll come back and do another pass.

Issue #9032, opened 2021-01-06
Deno 1.6.2 and 1.6.3 fail to fully handshake even after a `Deno.startTls` upgrade completes


aricart:
The issue seems to be that `Deno.startTls()` resolves before the connection is fully setup - any writer created on the resulting tlsConnection will fail - here's a test:

```
import {
  fail
} from "https://deno.land/std@0.83.0/testing/asserts.ts";
import { BufWriter } from "https://deno.land/std@0.83.0/io/mod.ts";

Deno.test("tls - raw tls upgrade", async () => {
  interface Deferred<T> extends Promise<T> {
    resolve: (value?: T | PromiseLike<T>) => void;
    reject: (reason?: any) => void;
  }

  function deferred<T>(): Deferred<T> {
    let methods = {};
    const p = new Promise<T>((resolve, reject): void => {
      methods = { resolve, reject };
    });
    return Object.assign(p, methods) as Deferred<T>;
  }

  const te = new TextEncoder();
  const td = new TextDecoder();

  function render(frame: Uint8Array): string {
    const cr = "␍";
    const lf = "␊";
    return td.decode(frame)
      .replace(/\n/g, lf)
      .replace(/\r/g, cr);
  }

  async function write(writer: BufWriter, pm: string) {
    try {
      const buf = te.encode(pm);
      const c = await writer.write(buf);
      await writer.flush();
      console.info(`< ${c}/${buf.length}:`, render(buf));
    } catch (err) {
      console.error(`error writing pm: ${pm}: ${err}`);
      throw err;
    }
  }

  const pong = deferred<void>();
  const hostname = "demo.nats.io";
  let conn = await Deno.connect({ hostname, port: 4443 });
  const buf = new Uint8Array(64 * 1023);

  let c = await conn.read(buf);
  if (c === null) {
    fail("connection closed");
  }
  //@ts-ignore - this has to be a number
  const frame = buf.subarray(0, c);
  console.info(">", render(frame));

  // attempt the tls upgrade
  console.info("upgrading connection");
  const tlsConn = await Deno.startTls(conn, { hostname });
  console.info("connection upgraded");

  // start a reader to print server responses
  let _ = (async function () {
    while (true) {
      try {
        const c = await tlsConn.read(buf);
        if (c === null) {
          break;
        }
        if (c) {
          const frame = buf.subarray(0, c);
          console.info(">", render(frame));

          const re = /^PONG\r\n/mi;
          if (re.test(td.decode(frame))) {
[...]

PR #9118, opened 2021-01-15
refactor: IO resource types, fix concurrent read/write and graceful close

piscisaureus:
TODO:
* add test for TLS graceful close
* also create separate type for `UnixStreamResource`.

### DIFF
--- a/cli/tests/unit/tls_test.ts
+++ b/cli/tests/unit/tls_test.ts
@@ -2,6 +2,7 @@
 import {
   assert,
   assertEquals,
+  assertStrictEquals,
   assertThrows,
   assertThrowsAsync,
   deferred,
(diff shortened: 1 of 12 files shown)


lucacasonato:
This should fix https://github.com/denoland/deno/issues/9032.

cli/tests/unit/tls_test.ts
```diff
@@ -182,6 +183,149 @@ unitTest(
...
+unitTest(
+  { perms: { read: true, net: true } },
+  async function tlsServerStreamHalfClose(): Promise<void> {
+    const [serverConn, clientConn] = await tlsPair(3501);
+    await Promise.all([
+      sendCloseWrite(serverConn),
+      receiveCloseWrite(clientConn),
+    ]);
+  },
+);
```
ry, line 294: Nice

cli/tests/unit/tls_test.ts
```diff
@@ -182,6 +183,149 @@ unitTest(
   },
 );
 
+async function tlsPair(port: number): Promise<[Deno.Conn, Deno.Conn]> {
```
ry, line 186: It's too bad we don't have a way to do ephemeral port allocaton

runtime/ops/io.rs
```diff
@@ -142,6 +156,80 @@ pub struct FileMetadata {
...
+  }
+}
+
+#[derive(Debug)]
+pub struct ReadOnlyResource<S> {
+  stream: AsyncRefCell<S>,
+  cancel_handle: CancelHandle,
+}
+
+impl<S: 'static> From<S> for ReadOnlyResource<S> {
```
ry, line 199: These two From<S> -> ReadOnlyResource<S> seem backwards to me.... I would expect you to be able to go from


ReadOnlyResource<S> -> S

But not from any static lifetime object to ReadOnlyResource... am I missing something?

piscisaureus: I can add some type constraints (`S: AsyncRead + Unpin + 'static`), but these are not necessary.
Note that these type constraints are present on the `impl<S> ReadOnlyResource<S>` block.

runtime/ops/io.rs
```diff
@@ -460,153 +530,163 @@ pub fn op_read(
...
+    s.write(&buf).await?
+  } else if let Some(s) = resource.downcast_rc::<TlsServerStreamResource>() {
+    s.write(&buf).await?
+  } else if let Some(s) = resource.downcast_rc::<UnixStreamResource>() {
+    s.write(&buf).await?
+  } else if let Some(s) = resource.downcast_rc::<StdFileResource>() {
+    s.write(&buf).await?
+  } else {
+    return Err(not_supported());
+  };
```
ry, line 655: I like that the conditional is now explicitly expressed here in the op. Much easier to read.

ry (approved):
Can't say I grok this fully yet - but it looks very clean and fixes bugs - so LGTM

Issue #9109, opened 2021-01-13
Fetch SNI of TLS handshake


imjamesb:
# Feature Request

A very useful feature for TCP listeners using the TLS protocol would be to fetch the SNI of the TLS handshake (client hello) after the handshake is done. For instance, adding a `sni?: string;` property to the `Deno.Conn` interface.

matt1:
It would also be useful to be able to specify the SNI (and ALPN) for creating a TLS connection with `ConnectTls()` too.

ihack2712:
> It would also be useful to be able to specify the SNI (and ALPN) for creating a TLS connection with `ConnectTls()` too.

When connecting with TLS the SNI will be added automatically based on the hostname property.

ignoramous:
Without SNI, TLS middleware cannot be written in Deno, which is a shame since its tokio/tower based IO is pretty darn fast (especially, after the recent Deno<->V8 serde changes).

bartlomieju:
@lucacasonato is this something we could address?

timonson:
Is SNI on the roadmap?

Codevendor:
Just wondering if SNI support for deno webserver is in the works. Without it, how would you handle virtual domains?

imjamesb:
> Just wondering if SNI support for deno webserver is in the works. Without it, how would you handle virtual domains?

request.headers.host

Codevendor:
> [imjamesb](/imjamesb)

How do you serve up the ssl certificate for each different host over same ip?

imjamesb:
@Codevendor I'm not sure Deno has that yet. I've stopped followinging Deno's development. Anyways, for non SSL you can use `request.headers.host`, but with SSL you will need to wait until Deno lets you serve certificates based on hostname.

danopia:
I'm implementing a TLS proxy and want to passthru specific connections&mdash;without decryption&mdash;based on the requested SNI. (Similar to [https://github.com/dlundquist/sniproxy](sniproxy).)

Unfortunately, `request.headers.host` is an HTTP concept, and so it's not available until after TLS is already set up.
So it doesn't seem I can use Deno for this feature today.

mmastrac:
Moving this to the planning stage and closing in favour of a planning issue for the feature: #23233. Thanks for the early discussion on this issue.

PR #9125, opened 2021-01-15
feat(tcp tls): Added sni property to conn object.

ihack2712:
Added a `sni` property to the `Conn` object. The `sni` property will be `undefined` if the SNI wasn't included in the TLS handshake, or if the TLS protocol wasn't used at all.

The `sni` property will be passed by `op_accept_tls`.

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1255,6 +1255,24 @@ declare namespace Deno {
    * ```
    */
   export function sleepSync(millis: number): Promise<void>;
+  export interface Conn {
+    /** The local address of the connection. */
+    readonly localAddr: Addr;
+    /** The remote address of the connection. */
+    readonly remoteAddr: Addr;
+    /** The resource ID of the connection. */
+    readonly rid: number;
+    /** **UNSTABLE**: yet to be vetted.
+     * The SNI the client sent to the server during the TLS handshake. If the
+     * connection was not made over TLS, or the client did not send the SNI
+     * during the [client hello](https://tools.ietf.org/html/rfc5246#section-7.4.1.2)
+     * message this property will be undefined.
+     */
+    readonly sni?: string;
+    /** Shuts down (`shutdown(2)`) the writing side of the TCP connection. Most
+     * callers should just use `close()`. */
+    closeWrite(): void;
+  }
 }
 
 declare function fetch(
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1266,9 +1266,9 @@ declare namespace Deno {
      * The SNI the client sent to the server during the TLS handshake. If the
      * connection was not made over TLS, or the client did not send the SNI
      * during the [client hello](https://tools.ietf.org/html/rfc5246#section-7.4.1.2)
-     * message this property will be undefined.
+     * message this property will be `null`.
      */
-    readonly sni?: string;
+    readonly sni: string | null;
     /** Shuts down (`shutdown(2)`) the writing side of the TCP connection. Most
      * callers should just use `close()`. */
     closeWrite(): void;
--- a/cli/tests/unit/tls_test.ts
+++ b/cli/tests/unit/tls_test.ts
@@ -261,7 +261,7 @@ unitTest(
 
     const incoming = await incomingPromise;
 
-    assertEquals((incoming as Deno.Conn & { sni?: string }).sni, hostname);
+    assertEquals(incoming.sni, hostname);
     listener.close();
     incoming.close();
   },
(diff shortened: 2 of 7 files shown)


ihack2712:
This resolves #9109 btw.

ry (commented):
Thanks - nice to have this feature but it needs a test.

ihack2712:
Added a test, was working locally. Excited to see if it passes the tests here :3

Anyhow, I'm in a little need of assistance regarding adding the feature as unstable.

cli/tests/unit/tls_test.ts
```diff
@@ -232,3 +233,41 @@ unitTest(
...
+    async function connect() {
+      const conn = await Deno.connectTls({
+        hostname,
+        port,
+        certFile: rootCertFile,
+      });
+      conn.close();
+    }
+
+    await resolvable;
```
lucacasonato, line 271: I think this test can be written without a `deferred`.

Also it looks like the `connect` function is never called right? It looks to me like this test should just be hanging forever.

```suggestion
  async function getAndConnectWithSNI(): Promise<void> {
    const hostname = "localhost";
    const port = 3600;
    const certFile = "cli/tests/tls/localhost.crt";
    const keyFile = "cli/tests/tls/localhost.key";
    const rootCertFile = "cli/tests/tls/RootCa.pem";

    const listener = Deno.listenTls({
      hostname,
      port,
      certFile,
      keyFile,
    });

    const incomingPromise = listener.accept();
    
    const outgoing = await Deno.connectTls({
      hostname,
      port,
      certFile: rootCertFile,
    });
    outgoing.close();
    
    const incoming = await incomingPromise;
    assertEquals((incoming as Deno.Conn & { sni?: string }).sni, hostname);
    incoming.close();
    listener.close();
```

cli/tests/unit/tls_test.ts
```diff
@@ -232,3 +233,41 @@ unitTest(
...
+
+    const listener = Deno.listenTls({
+      hostname,
+      port,
+      certFile,
+      keyFile,
+    });
+
+    listener.accept().then((conn) => {
+      assertEquals((conn as Deno.Conn & { sni?: string }).sni, hostname);
```
lucacasonato, line 254: You need to fix these typings. Please add the correct typings to the `cli/dts/lib.deno.unstable.d.ts` file.

runtime/ops/tls.rs
```diff
@@ -398,6 +398,13 @@ async fn op_accept_tls(
     .try_or_cancel(cancel)
     .await?;
 
+  let sni = tls_stream
+    .get_ref()
+    .1
+    .get_sni_hostname()
+    .unwrap()
```
lucacasonato, line 405: You can't unwrap here. TLS connections do not always have the SNI set. SNI must be an `Option`.

cli/tests/unit/tls_test.ts
ihack2712: I tried adding the typings to `lib.deno.unstable.d.ts`. However, during the tests they were not applied.

cli/tests/unit/tls_test.ts
ihack2712: Yes, you're right! Don't know why my changes weren't pushed, the tests worked locally before pushing. I'm digging into it.

runtime/ops/tls.rs
ihack2712: Fixed in next push thanks to @SyrupThinker!

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1255,6 +1255,24 @@ declare namespace Deno {
...
+    readonly remoteAddr: Addr;
+    /** The resource ID of the connection. */
+    readonly rid: number;
+    /** **UNSTABLE**: yet to be vetted.
+     * The SNI the client sent to the server during the TLS handshake. If the
+     * connection was not made over TLS, or the client did not send the SNI
+     * during the [client hello](https://tools.ietf.org/html/rfc5246#section-7.4.1.2)
+     * message this property will be undefined.
+     */
+    readonly sni?: string;
```
lucacasonato, line 1271: Typing is not correct. It should be `string | null` not `string | undefined`.

runtime/ops/tls.rs
```diff
@@ -398,6 +398,12 @@ async fn op_accept_tls(
     .try_or_cancel(cancel)
     .await?;
 
+  let sni = tls_stream
```
lucacasonato, line 401: Wrap this in an if statement to check if unstable is enabled. If it is not, sni should be None.

piscisaureus:
@piscisaureus

lucacasonato:
@ihack2712 Thanks for the PR, but we are not going to land the PR as the proposed API has no path to stabilisation, as exposing the SNI (and ALPN and friends) should be done as a more general redesign of the TLS API to support dynamic certificates.

ihack2712:
Okay! You mention dynamic certificates, is that on the roadmap?

PR #9131, opened 2021-01-16
refactor(lsp): factor out fields to DocumentSpan

matsujika:
In #9071 I added `DocumentSpan` and have noticed there are several cases where equivalent fields are hard-coded.
This PR is to replace those fields for `DocumentSpan`.


lucacasonato (approved):
LGTM. Thanks

PR #9133, opened 2021-01-16
fix(cli/install): Escape % symbols in installed script urls on windows

### LINKED ISSUE #9096: deno install creates a .cmd that does not work with my windows user name
I created a small program that I wanted available, so I `deno install`ed it.
But running it gives me this:

```
PS C:\Users\Einar Magnús Boson> ustyle.cmd
error: Cannot resolve module "file:///C:/Users/Einar0MagnBAs0Boson/src/unicode-styling/ustyle.ts".
```

I believe it's the accent in my name that does this, some other tools also fail on it. Most things work though, and I would be happy if Deno did too 🦖

Liamolucko:
Resolves #9096


cli/tools/installer.rs
```diff
@@ -51,7 +51,10 @@ fn generate_executable_file(
   mut file_path: PathBuf,
   args: Vec<String>,
 ) -> Result<(), AnyError> {
-  let args: Vec<String> = args.iter().map(|c| format!("\"{}\"", c)).collect();
+  let args: Vec<String> = args
```
lucacasonato, line 54: These args are nit just used by the generated .cmd script, but also the generated script compatible with git bash / MINGW64. (On windows deno install creates two scripts). Should update this to only apply the escaping to the .cmd script.

lucacasonato (commented):
Great test, just one comment.

piscisaureus:
This is not the right way to do shell escaping on windows. Leave the PR open - I’ll add more detail later.

piscisaureus:
In the meantime, is there a test or bug report for the issue that this patch is supposed to fix?

lucacasonato:
@piscisaureus https://github.com/denoland/deno/issues/9096

piscisaureus:
> This is not the right way to do shell escaping on windows. Leave the PR open - I’ll add more detail later.

Actually, it looks like in batch files `%` *can* be escaped (unlike in the terminal itself), so this patch should indeed solve the problem. 
Thanks for the fix @Liamolucko, LGTM.

piscisaureus (approved):

PR #9135, opened 2021-01-16
chore: Enforce ban-untagged-todo lint rule

### LINKED ISSUE #7410: Enable ban-untagged-todo lint rule
`std/` and `cli/unit/` should be linted with `ban-untagged-todo` rule to ensure that all `TODO`s have either user handle (`TODO(@bartlomieju)`) or issue number (`TODO(#1012)`)

bartlomieju:
Closes #7410

### DIFF
--- a/.dlint.json
+++ b/.dlint.json
@@ -1,0 +1,9 @@
+{
+    "rules": {
+        "tags": ["recommended"],
+        "include": [
+            "ban-untagged-todo"
+        ],
+        "exclude": []
+    }
+}
--- a/.dlint.json
+++ b/.dlint.json
@@ -1,9 +1,9 @@
 {
-    "rules": {
-        "tags": ["recommended"],
-        "include": [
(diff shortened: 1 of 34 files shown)


.dlint.json
```diff
@@ -0,0 +1,9 @@
+{
+  "rules": {
+    "tags": ["recommended"],
+    "include": [
+      "ban-untagged-todo"
+    ],
+    "exclude": []
+  }
+}
```
ry, line 9: Seems like dlint’s default should be what we use in Deno - and therefore the config would be unnecessary.

fine with landing this as is now - but can  this config be removed in the futue?

ry (approved):
LGTM

.dlint.json
bartlomieju: If we decide that `ban-untagged-todo` should be in the `recommended` set then sure.

PR #9139, opened 2021-01-17
fix(cli/runtime_compiler): Check permissions for Deno.emit()

nayeemrmn:
`await Deno.emit("std/examples/welcome.ts")` doesn't require permissions.

cc @kitsonk

### DIFF
--- a/cli/ops/runtime_compiler.rs
+++ b/cli/ops/runtime_compiler.rs
@@ -66,9 +66,9 @@ async fn op_emit(
   let mut is_dynamic = false;
   let handler: Arc<Mutex<dyn SpecifierHandler>> =
     if let Some(sources) = args.sources {
-      is_dynamic = true;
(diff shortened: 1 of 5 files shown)


cli/ops/runtime_compiler.rs
```diff
@@ -66,9 +66,9 @@ async fn op_emit(
   let mut is_dynamic = false;
   let handler: Arc<Mutex<dyn SpecifierHandler>> =
     if let Some(sources) = args.sources {
-      is_dynamic = true;
       Arc::new(Mutex::new(MemoryHandler::new(sources)))
     } else {
+      is_dynamic = true;
```
nayeemrmn, line 71: This typo caused modules to be loaded as if statically.

kitsonk (approved):
LGTM, thanks for the fix.

PR #9141, opened 2021-01-17
feat: Standalone lite binaries and cross compilation

bartlomieju:
This commit adds `--target` and `--lite` flags to `deno compile` subcommand.

`--target` allows to cross-compile binary to different target architectures by
fetching appropriate binary from remote server on first run. All downloaded
binaries are stored in "$DENO_DIR/dl".

`--lite` allows to use lite version of the runtime (ie. the one that doesn't contain
built-in tooling like formatter or linter).

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -928,7 +928,13 @@ The executable name is inferred by default:
   - If the file stem is something generic like 'main', 'mod', 'index' or 'cli',
     and the path has no parent, take the file name of the parent path. Otherwise
     settle with the generic name.
-  - If the resulting name has an '@...' suffix, strip it.",
+  - If the resulting name has an '@...' suffix, strip it.
+  
+This commands supports cross-compiling to different target architectures using `--target` flag.
+On the first invocation with deno will download proper binary and cache it in $DENO_DIR.
+
+It is possible to use \"lite\" binaries when compiling by passing `--lite` flag; these are stripped down versions
+of the deno binary that do not contain built-in tooling (eg. formatter, linter). This feature is experimental.",
     )
 }
 
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -29,6 +29,8 @@ pub enum DenoSubcommand {
     source_file: String,
     output: Option<PathBuf>,
     args: Vec<String>,
+    target: Option<String>,
+    lite: bool,
   },
   Completions {
     buf: Box<[u8]>,
@@ -447,11 +449,13 @@ fn compile_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   let args = script.split_off(1);
   let source_file = script[0].to_string();
   let output = matches.value_of("output").map(PathBuf::from);
+  let lite = matches.is_present("lite");
 
   flags.subcommand = DenoSubcommand::Compile {
     source_file,
     output,
     args,
+    lite,
   };
 }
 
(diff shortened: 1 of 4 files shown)


cli/flags.rs
```diff
@@ -909,9 +928,7 @@ The executable name is inferred by default:
   - If the file stem is something generic like 'main', 'mod', 'index' or 'cli',
     and the path has no parent, take the file name of the parent path. Otherwise
     settle with the generic name.
-  - If the resulting name has an '@...' suffix, strip it.
-
-Cross compiling binaries for different platforms is not currently possible.",
+  - If the resulting name has an '@...' suffix, strip it.",
```
bartlomieju, line 931: Update note; mention `--target` and `--lite` behaviour

cli/tools/standalone.rs
```diff
@@ -1,28 +1,105 @@
...
 
+pub async fn get_base_binary(
+  deno_dir: &DenoDir,
+  target: Option<String>,
+  lite: bool,
+) -> Result<Vec<u8>, AnyError> {
+  if target.is_none() && !lite {
+    let path = std::env::current_exe()?;
+    return Ok(tokio::fs::read(path).await?);
+  }
```
bartlomieju, line 31: This functionality can be shared with `upgrade` command

cli/tools/standalone.rs
```diff
@@ -1,28 +1,105 @@
...
+  let download_url = format!("https://dl.deno.land/{}", binary_path_suffix);
+
+  let client_builder = Client::builder();
+
+  // TODO:
+  // // If we have been provided a CA Certificate, add it into the HTTP client
+  // if let Some(ca_file) = ca_file {
+  //   let buf = std::fs::read(ca_file)?;
+  //   let cert = reqwest::Certificate::from_pem(&buf)?;
+  //   client_builder = client_builder.add_root_certificate(cert);
```
bartlomieju, line 72: Not sure if that makes sense to use `--cafile` arg, for this purpose as it would be embedded in final binary. OTOH it changes behavior between `upgrade` and `compile` downloads; @nayeemrmn thoughts?

cli/tools/standalone.rs
```diff
@@ -1,28 +1,105 @@
...
+
+  let res = client.get(&download_url).send().await?;
+
+  let binary_content = if res.status().is_success() {
+    println!("Download has been found");
+    res.bytes().await?.to_vec()
+  } else {
+    println!("Download could not be found, aborting");
+    std::process::exit(1)
+  };
```
bartlomieju, line 87: Copy-pasta from `cli/tools/upgrade.rs`

lucacasonato (approved):
LGTM - great feature to have. I am somewhat concerned about this root certificate thing, but lets cross that bridge once someone reports it as an issue :-)

cli/flags.rs
```diff
@@ -893,11 +899,24 @@ fn compile_subcommand<'a, 'b>() -> App<'a, 'b> {
...
+      Arg::with_name("lite")
+        .long("lite")
+        .help("Use lite runtime")
+    )
     .about("Compile the script into a self contained executable")
     .long_about(
       "Compiles the given script into a self contained executable.
   deno compile --unstable https://deno.land/std/http/file_server.ts
   deno compile --unstable --output /usr/local/bin/color_util https://deno.land/std/examples/colors.ts
+  deno compile --unstable --lite --target x86_64-unknown-linux-gnu https://deno.land/std/http/file_server.ts
```
ry, line 919: Should this include permission flags? `-A`?

ry (approved):
LGTM!

PR #9142, opened 2021-01-17
fix(runtime/js) use DOMException in Performance#measure for WPT testing

00ff0000red:
Fixes https://github.com/denoland/deno/issues/9009.

Does anything else need to be changed?

Also, Chromium has an argument length check,
> ```
> TypeError: Failed to construct 'PerformanceMark': 1 argument required, but only 0 present.
> ```
Should that be added too?

Update, I did it anyway.

### DIFF
--- a/cli/tests/wpt.jsonc
+++ b/cli/tests/wpt.jsonc
@@ -209,16 +209,7 @@
     "mark-errors",
     "mark-measure-return-objects",
     "mark.any",
-    {
-      "name": "measure_syntax_err",
-      "expectFail": [
-        // TODO(lucacasonato): re-enable when #9009 is fixed.
-        "self.performance.measure(\"measure\", \"mark\"), where \"mark\" is a non-existent mark, throws a SyntaxError exception.",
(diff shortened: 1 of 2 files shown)


00ff0000red:
cc @lucacasonato

00ff0000red:
There seem to be some tests that are failing that,
a) I can't find the source of
b) expect behavior that differs from Chromium's implementation

> ```
> test Number should be rejected as the mark-options. ... failed
> test NaN should be rejected as the mark-options. ... failed
> test Infinity should be rejected as the mark-options. ... failed
> test String should be rejected as the mark-options. ... failed
> ```

> ```
> assert_throws_js: Number passed as a dict argument should cause type-error. function "function() { self.performance.mark("mark1", 123); }" did not throw
> ```

Ah, okay, found it, /test_util/wpt/user-timing/mark-errors.any.js, should I remove these tests or...?

lucacasonato (approved):
LGTM, thanks!

Issue #9006, opened 2021-01-05
bug: encodeInto implementation is not correct


lucacasonato:
You can enable the failing WPT tests by adding this snippet to `cli/tests/wpt.jsonc`:

```json
{
   "name": "encodeInto",
   "expectFail": [
     // TODO(lucacasonato): enable once we support MessageChannel
     "encodeInto() and a detached output buffer"
   ]
},
```

Errors from the failing tests look like this:
```
encodeInto() into ArrayBuffer with �A�A¥Hi and destination length 10, offset 4, filler random
assert_equals: expected 239 but got 237
Error
    at AssertionError.get_stack (file:///tmp/wpt-bundle-J6kGw.js:3617:21)
    at new AssertionError (file:///tmp/wpt-bundle-J6kGw.js:3610:27)
    at assert (file:///tmp/wpt-bundle-J6kGw.js:3600:19)
    at assert_equals (file:///tmp/wpt-bundle-J6kGw.js:1225:9)
    at Test.<anonymous> (file:///tmp/wpt-bundle-J6kGw.js:4320:13)
    at Test.step (file:///tmp/wpt-bundle-J6kGw.js:2041:25)
    at test (file:///tmp/wpt-bundle-J6kGw.js:569:30)
    at file:///tmp/wpt-bundle-J6kGw.js:4286:7
    at Array.forEach (<anonymous>)
    at file:///tmp/wpt-bundle-J6kGw.js:4285:42
```

00ff0000red:
I setup a quick test using a detached buffer and Deno threw an exception,
> ```
> error: Uncaught TypeError: Cannot perform %TypedArray%.prototype.entries on a detached ArrayBuffer
>     const iter = value.entries();
> ```

```js
const memory = new WebAssembly.Memory({
	initial: 1,
	maximum: 1,
	shared: false
});

const array = new Uint8Array(memory.buffer, 0, 100);

memory.grow(0);

const result = new TextEncoder().encodeInto("Hello, world!", array);

console.log(array, result);
```

`result` was `{ read: 0, written: 0 }`.

Is this even related to the WPT test assertion?

lucacasonato:
#9143 did not fix the error WPT errors.

00ff0000red:
Unfortunate. Can I get more details on this WPT test in particular? i.e. its source?

lucacasonato:
[...]

PR #9143, opened 2021-01-17
fix(op_crates/web) let TextEncoder#encodeInto accept detached ArrayBuffers

00ff0000red:
Fixes what I had found, that I posted [in a comment](https://github.com/denoland/deno/issues/9006#issuecomment-761876048).

Currently, `TextEncoder#encodeInto` lets an exception propograte to the callee when called with a detached JS ArrayBuffer.

Detached array buffers have zero bytes, there is otherwise no way in JS (without V8 intrinsics) to determine whether or not an ArrayBuffer is detached.

As a side-effect, this optimizes the case when passing an empty ArrayBuffer that isn't detached into `TextEncoder#encodeInto`.

### DIFF
--- a/op_crates/web/08_text_encoding.js
+++ b/op_crates/web/08_text_encoding.js
@@ -1134,14 +1134,16 @@
       return new Uint8Array(output);
     }
     encodeInto(input, dest) {
-      const encoder = new UTF8Encoder();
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM- thanks

Issue #9029, opened 2021-01-06
Move standard library to separate repository


bartlomieju:
Historically standard library was developed in a separate repository (https://github.com/denoland/deno_std) but given rapid pace of changes to Deno public APIs (the TypeScript APIs available on `Deno` global) it was decided that `std` should be developed in tree of the main repo to ensure `std` works with the latests changes in Deno. All of that led to the archival of `deno_std` repository.

Given that public APIs are not changing rapidly anymore and are mostly stable it is high time we move standard library back to a separate repo. 

Doing so has a number of advantages:
- independent testing of `std` with multiple versions of `Deno` (using Github Actions matrix)
- as a by-product CI in main repo will run faster
- allow contributors to `std` to focus on a single purpose repository written in TypeScript; avoiding the noise of main repo

Before we can move `std` to `deno_std` we need to do following:
- copy-paste `std` code to a new test repo
- set up CI script that will test `std` the way it's currently tested in this repo (by rewriting `std_tests` integration test into a CI step)
- for starters it should use latest version of Deno
- additionally we should add a step to test against canary version of Deno (but failures in that stage shouldn't fail whole CI run)
- set up linting and formatting for the repo by using `dprint` and `dlint` from `deno_third_party` repo and `tools/lint.js` and `tools/format.js` scripts respectively; and "check" step in the CI

Once above points are addressed we need to:
- unarchive `deno_std` repository
- carefully migrate the code retaining its Git history (and thus attributions) to `deno_std` repo
- bring CI setup from above to `deno_std` repo

This is a rather complex undertaking and will take a few weeks to complete.

_The core team is working on this task._

caspervonb:
> set up linting and formatting for the repo by using dprint and dlint from deno_third_party repo and tools/lint.js and tools/format.js scripts respectively; and "check" step in the CI

Out of curiosity, any particular reason we can't migrate over to linting and formatting with stable deno?

nayeemrmn:
I agree with @caspervonb, `deno_std` should be maintained as a model userspace library.

bartlomieju:
Actually that's a pretty good idea, should simplify setup a bit.

bartlomieju:
Thanks to @caspervonb and @piscisaureus hard work we are ready to move standard library to a separate repository (https://github.com/denoland/deno_std).

We plan to do that on Monday, February 1st, 2021. 

[...]

PR #9144, opened 2021-01-17
test(std/node): ensure process.env case doesn't rely on unset variables

caspervonb:
While PATH is commonly defined there is no guarantee that it will be defined and turns out it is undefined with GitHub's runner on Windows.

Came up when preparing the workflow for https://github.com/denoland/deno/issues/9029 with @bartlomieju


ry (approved):
LGTM

PR #9147, opened 2021-01-17
feat(cli/dts/lib.deno.shared_globals): Add WorkerOptions

nayeemrmn:
Not sure why this was missing, as mentioned on discord it makes accessing things like `Exclude<WorkerOptions["deno"], undefined | true>` easier.


bartlomieju (approved):
LGTM, I think it's better now

PR #9148, opened 2021-01-17
test(std/node): avoid project directories for cwd case

caspervonb:
The process.cwd test relies on std being in a nested directory within the top level deno project; this makes it use temporary directories instead.

Came up when preparing the workflow for #9029 with @bartlomieju


ry (approved):
LGTM - nice!

PR #9150, opened 2021-01-18
fix: redirect in --location relative fetch

lucacasonato:
Before:

```shell
$ cat example.ts
console.log(globalThis.location.href);

const res = await fetch("/std/version.ts");
console.log(res.status, res.url);
console.log(await res.text());
$ deno run --location="https://deno.land" --allow-net example.ts
https://deno.land/
error: Uncaught (in promise) TypeError: Invalid base URL.
          throw new TypeError("Invalid base URL.");
                ^
    at new URL (deno:op_crates/web/11_url.js:714:17)
    at fetch (deno:op_crates/fetch/26_fetch.js:1407:29)
    at async file:///Users/lucacasonato/Projects/github.com/denoland/deno/example.ts:3:13
```

After:

```shell
$ cat example.ts
console.log(globalThis.location.href);

const res = await fetch("/std/version.ts");
console.log(res.status, res.url);
console.log(await res.text());
$ deno run --location="https://deno.land" --allow-net example.ts
https://deno.land/
200 https://deno.land/std@0.83.0/version.ts
/** Version of the Deno standard modules
 *
 * Deno std is versioned differently than Deno cli because it is still unstable;
 * the cli's API is stable. In the future when std becomes stable, likely we
 * will match versions with cli as we have in the past.
 */
export const VERSION = "0.83.0";

```


bartlomieju (approved):
LGTM

PR #9154, opened 2021-01-18
Add Makefile

ry:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 test:
-	cargo build
+	cargo build -p test_util
 	cargo test
 
 fmt:
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+test:
+	cargo build
+	cargo test
+
+fmt:
+	deno run -A --unstable tools/format.js
+
+lint:
+	deno run -A --unstable tools/lint.js
+
+.PHONY: test fmt lint


Makefile
```diff
@@ -0,0 +1,11 @@
+test:
+	cargo build
+	cargo test
+
+fmt:
+	deno run -A --unstable tools/format.js
```
caspervonb, line 6: Could also be

```suggestion
  cargo run -- run -A --unstable tools/format.js
```

To fmt/lint with the result of cargo build.

lucacasonato (approved):
LGTM, but why? Can't you just run the file directly as it has a shebang? I can on both macOS and Linux.

Makefile
```diff
@@ -0,0 +1,11 @@
+test:
+	cargo build
```
nayeemrmn, line 2:
```suggestion
```

nayeemrmn:
@lucacasonato They don't work on Ubuntu 18 or earlier. Only GNU coreutils 8.30+.

lucacasonato:
Why not update coreutils then? Also Ubuntu 20.4 is the latest LTS release, which has these updated coreutils.

Makefile
SyrupThinker, line 2: This should be
```suggestion
	cargo build --all-targets
```
Otherwise the `test_server` is not built and some tests will fail.

Makefile
nayeemrmn: That's not the behaviour in CI, these are meta scripts which are meant to use the latest release.

SyrupThinker (changes requested):

SyrupThinker (approved):
@nayeemrmn's updated suggestion is indeed better :D

kt3k:
How about updating `//.github/PULL_REQUEST_TEMPLATE.md` and `//docs/contributing.md` as well?

ry:
I retract this, I guess it's not so useful, just another file to clutter the root.

Issue #9157, opened 2021-01-18
"deno help" does not end in a newline 


ry:
```
~/src/deno> ./target/debug/deno help
deno 1.6.3
A secure JavaScript and TypeScript runtime

Docs: https://deno.land/manual
Modules: https://deno.land/std/ https://deno.land/x/
Bugs: https://github.com/denoland/deno/issues

To start the REPL:
  deno

To execute a script:
  deno run https://deno.land/std/examples/welcome.ts

To evaluate code in the shell:
  deno eval "console.log(30933 + 404)"

USAGE:
    deno [OPTIONS] [SUBCOMMAND]

OPTIONS:
    -h, --help                     Prints help information
    -L, --log-level <log-level>    Set log level [possible values: debug, info]
    -q, --quiet                    Suppress diagnostic output
        --unstable                 Enable unstable features and APIs
    -V, --version                  Prints version information

SUBCOMMANDS:
    bundle         Bundle module and dependencies into single file
    cache          Cache the dependencies
    compile        Compile the script into a self contained executable
    completions    Generate shell completions
    doc            Show documentation for a module
    eval           Eval script
    fmt            Format source files
    help           Prints this message or the help of the given subcommand(s)
    info           Show info about cache or info related to source file
    install        Install script as an executable
    lint           Lint source files
    lsp            Start the language server
    repl           Read Eval Print Loop
    run            Run a program given a filename or url to the module. Use '-' as a filename to read from stdin.
    test           Run tests
    types          Print runtime TypeScript declarations
    upgrade        Upgrade deno executable to given version

ENVIRONMENT VARIABLES:
    DENO_DIR             Set the cache directory
    DENO_INSTALL_ROOT    Set deno install's output directory
                         (defaults to $HOME/.deno/bin)
    DENO_CERT            Load certificate authority from PEM encoded file
    NO_COLOR             Set to disable color
    HTTP_PROXY           Proxy address for HTTP requests
                         (module downloads, fetch)
    HTTPS_PROXY          Proxy address for HTTPS requests
                         (module downloads, fetch)
    NO_PROXY             Comma-separated list of hosts which do not use a proxy
                         (module downloads, fetch)~/src/deno>
```

caspervonb:
Fix incoming.

PR #9064, opened 2021-01-09
fix: don't error on version and help flag

lucacasonato:
Should fix #9048

PR #9158, opened 2021-01-18
fix(cli): print a newline after help and version

caspervonb:
Currently `deno --help` and `deno --version` don't end with a new line, this fixes that.
This regression was introduced in https://github.com/denoland/deno/pull/9064

Closes https://github.com/denoland/deno/issues/9157


caspervonb:
cc @lucacasonato

lucacasonato (approved):
Thanks

PR #9159, opened 2021-01-18
docs(std/fmt): README example colors.ts usage.

grian32:
Adds example usage of colors.ts to `std/fmt` README.


bartlomieju (approved):
LGTM, thanks @grian32

PR #9160, opened 2021-01-18
feat(cli/test): add deno test --watch

Liamolucko:
This implements file watching for `deno test`.

When a file is changed, only the test modules which use it as a dependency are rerun.

This is accomplished by reworking the file watching infrastructure to pass the paths which have changed to the resolver, and then constructing a module graph for each test module to check if it contains any changed files.

This also updates `deno fmt` to only format the changed files.


cli/file_watcher.rs
```diff
@@ -74,180 +77,127 @@ async fn error_handler(watch_future: FileWatcherFuture<Result<(), AnyError>>) {
...
+        paths_to_watch,
+        result,
+      } => {
+        if !initial {
+          info!(
+            "{} File change detected! Restarting!",
+            colors::intense_blue("Watcher"),
+          );
+        }
+        break (paths_to_watch, result);
```
magurotuna, line 113: Here, is this `break` exactly equivalent to `return`? If so, personally I prefer `return` to `break` for readability.

cli/file_watcher.rs
```diff
@@ -74,180 +77,127 @@ async fn error_handler(watch_future: FileWatcherFuture<Result<(), AnyError>>) {
...
+
+async fn next_restart<F, T: Clone>(
+  resolver: &mut F,
+  debounce: &mut Pin<&mut Debounce>,
+  initial: bool,
+) -> (Vec<PathBuf>, Result<T, AnyError>)
 where
-  F: Fn() -> Result<Vec<PathBuf>, AnyError>,
-  G: Fn(Vec<PathBuf>) -> FileWatcherFuture<Result<(), AnyError>>,
+  F: FnMut(Option<Vec<PathBuf>>) -> FileWatcherFuture<ResolutionResult<T>>,
```
magurotuna, line 94: If there's a `where` clause, I think it would be good to put all trait bounds there - I mean, I'd like to suggest that `T: Clone` be in `where` clause:

```rust
where
  F: FnMut(Option<Vec<PathBuf>>) -> FileWatcherFuture<ResolutionResult<T>>,
  T: Clone,
```

cli/file_watcher.rs
```diff
@@ -74,180 +77,127 @@ async fn error_handler(watch_future: FileWatcherFuture<Result<(), AnyError>>) {
...
-        "{} {} finished! Restarting on file change...",
-        colors::intense_blue("Watcher"),
-        job_name,
-      );
-      debounce.next().await;
-      info!(
-        "{} File change detected! Restarting!",
-        colors::intense_blue("Watcher"),
-      );
+    let initial = changed.is_none();
```
magurotuna, line 98: Is this `let initial = changed.is_none();` necessary? 
Seems like if and only if `initial` as the function argument is `true`, the value of `changed` is set to `None` (since `debounce.next().await` always returns `Some(..)`, which means it's an infinite stream)
Therefore, in `let initial = changed.is_none();`, it looks like this `initial` is definitely equal to the function argument's `initial`.

cli/file_watcher.rs
```diff
@@ -74,180 +77,127 @@ async fn error_handler(watch_future: FileWatcherFuture<Result<(), AnyError>>) {
...
   // Store previous data. If module resolution fails at some point, the watcher will try to
   // continue watching files using these data.
-  let mut paths = Vec::new();
-  let mut module = None;
+  let mut paths_to_watch;
+  let mut resolution_result;
+
+  let (paths, result) = next_restart(&mut resolver, &mut debounce, true).await;
+  paths_to_watch = paths;
+  resolution_result = Some(result);
```
magurotuna, line 152: Can this be written without `Some`?
`resolution_result` is of type `Option<Result<T, AnyError>>` but it looks like the value is always `Some`, never `None`.

Liamolucko: I originally did that because when I tried to do it without `Option`, I was getting borrow checker errors because `resolution_result` was undefined for a period... but I just tried it again and it worked, so I don't know what I did wrong the first time :upside_down_face:

cli/file_watcher.rs
Liamolucko: Turns out `T` didn't actually need to be `Clone` in the first place - not sure why that was there.

cli/file_watcher.rs
Liamolucko: It becomes `false` after the first iteration of loop, since `changed` is updated - probably not the best way of writing it, to be honest, but it's now been refactored away anyway.

magurotuna:
FYI: the tests related to the watcher are super flaky. It's been a high-priority issue for a while (see #8571) but there's no progress as far as I know.
The existing tests about the watcher are skipped by being annotated with `#[ignore]`. I think it would be fine to remove these annotations and run tests on your local environment.

caspervonb (changes requested):
I don't see this working well with coverage so the --watch option should conflict with the --coverage option.

Liamolucko:
At the moment this runs the whole test suite at first, and only after that watches for file changes. Should there be an option (or the default) to do nothing until a file change is detected?

bartlomieju:
@caspervonb does this PR somehow conflicts with your rewrite of test runner?

caspervonb:
> @caspervonb does this PR somehow conflicts with your rewrite of test runner?

Tiny bit but believe it to be trivial to fix on my end @bartlomieju.

> At the moment this runs the whole test suite at first, and only after that watches for file changes. Should there be an option (or the default) to do nothing until a file change is detected?

Maybe it makes sense to make use of the --no-run flag here (when watch is given it will take on the meaning of "no initial run but keep watching after") @Liamolucko?

bartlomieju:
> Maybe it makes sense to make use of the --no-run flag here (when watch is given it will take on the meaning of "no initial run but keep watching after") @Liamolucko?

I believe these two flags should be exclusive, @nayeemrmn WDYT?

nayeemrmn:
I was going to say it should match `deno cache --watch` which recompiles if any file in the module graph changes, but we don't have that. So it should be mutually exclusive for now.

bartlomieju:
@Liamolucko could you elaborate a bit on the changes to file watcher? It looks like a major rewrite and I'm not entirely sure on the reason for it. Any rationale would be helpful

Liamolucko:
The main reason is to pass the files which have changed to the resolver, since that's necessary to only rerun the changed test files.

In order to support the addition of new test files, I also had to change it to recursively watch directories. But that meant I needed to be able to ignore non-test files changing, so I changed `ResolutionResult` to an enum of `Pending` and `Restart`.

The rest was just refactors, mainly removing the distinction between `watch_func` and `watch_func_with_module_resolution` (the only difference being that `watch_func` didn't previously let the resolver pass an argument to the operation).

bartlomieju:
@Liamolucko I discussed this PR with @caspervonb several times; while we are very much in favor of this feature, there are still some questions about how to integrate file watcher, especially in the context of concurrent runner Casper is working on. I'm gonna bump the milestone to 1.9 and we'll try to land it next week after 1.8 release.

caspervonb:
Hey @Liamolucko!

So, good news! The new test runner has a new but fairly simple entry point you could run here.
The need to generate the index file and all that nonsense is gone, we run the tests with the test module as the main module now.

So really you'd just need to forward that call into a watcher operation that can give a Vec of module specifiers to run for `include` and call `cli::tools::test_runner::run_tests` (preferably watch would be kept isolated in test_subcommand and not leak into the runner).

caspervonb:
Quick and dirty example of what I'm thinking:

```rust
fn test_command() {
  // ...
  
  if watch {
    let resolver = // ...;
    let operation = |/*...*/| {
      tools::test_runner::run_tests(
      // ...
        changed_paths,
        // ...
      ).await?;
   };

    file_watcher::watch_func(resolver, operation, "Fmt").await?;
  } else {
    tools::test_runner::run_tests(
      // ...
      include,
      // ...
    ).await?;
  }
}
```

cli/main.rs
```diff
@@ -917,19 +903,150 @@ async fn test_command(
...
+    module_graph::TypeLib::DenoWindow
+  };
+
+  if flags.watch {
+    let handler = Arc::new(Mutex::new(FetchHandler::new(
+      &program_state,
+      Permissions::allow_all(),
+    )?));
+
+    let paths_to_watch: Vec<_> = include.iter().map(PathBuf::from).collect();
```
caspervonb, line 924: Hmm so this is watching the files directly, not their parent directories?

caspervonb: Actually nevermind, on second thought this is fine.

bartlomieju:
@Liamolucko this is a great feature to have! I did some testing locally and found two quirks:

1. When a test fails the process exits completely:
```
❯ target/debug/deno test --allow-all --unstable --watch --no-check foo/

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (58ms)

error: Cannot resolve module "file:///Users/biwanczuk/dev/deno/foo/cli/tests/unit/test_util.ts" from "file:///Users/biwanczuk/dev/deno/foo/file1_test.ts".
    at file:///Users/biwanczuk/dev/deno/foo/file1_test.ts:5:0
Watcher Test finished! Restarting on file change...
running 1 test from file:///Users/biwanczuk/dev/deno/foo/file1_test.ts
test filesStdioFileDescriptors ... FAILED (13ms)

failures:

filesStdioFileDescriptors
AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   1
+   0

    at assertEquals (file:///Users/biwanczuk/dev/deno/test_util/std/testing/asserts.ts:217:18)
    at filesStdioFileDescriptors (file:///Users/biwanczuk/dev/deno/foo/file1_test.ts:13:5)
    at asyncOpSanitizer (deno:runtime/js/40_testing.js:21:15)
    at resourceSanitizer (deno:runtime/js/40_testing.js:58:13)
    at exitSanitizer (deno:runtime/js/40_testing.js:85:15)
    at runTest (deno:runtime/js/40_testing.js:199:13)
    at Object.runTests (deno:runtime/js/40_testing.js:244:13)
    at file:///Users/biwanczuk/dev/deno/$deno$test.js:1:27

failures:

	filesStdioFileDescriptors

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (377ms)
```

2. I watch test in a directory (`target/debug/deno test --allow-all --unstable --watch --no-check foo/`); this directory contains only a single file (say `file1_test.ts`); when I add a new file (`file2_test.ts`), only tests from that new file are run:
```
running 1 test from file:///Users/biwanczuk/dev/deno/foo/file1_test.ts
test filesStdioFileDescriptors ... ok (7ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (332ms)

Watcher Test finished! Restarting on file change...
running 1 test from file:///Users/biwanczuk/dev/deno/foo/file2_test.ts
test filesStdioFileDescriptors2 ... ok (7ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (365ms)

Watcher Test finished! Restarting on file change...
running 1 test from file:///Users/biwanczuk/dev/deno/foo/file2_test.ts
test filesStdioFileDescriptors2 ... ok (7ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (319ms)

Watcher Test finished! Restarting on file change...
running 1 test from file:///Users/biwanczuk/dev/deno/foo/file2_test.ts
test filesStdioFileDescriptors3 ... ok (26ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (367ms)

Watcher Test finished! Restarting on file change...
running 1 test from file:///Users/biwanczuk/dev/deno/foo/file1_test.ts
test filesStdioFileDescriptors10 ... ok (8ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (373ms)

Watcher Test finished! Restarting on file change...
```
Similarly changing original file (`file1_test.ts`) only reruns tests from that file. Is that the behavior you wanted to achieve?

Liamolucko:
> 1. When a test fails the process exits completely:

Oops, that's a bug. I'll have to fix that.

> 2. I watch test in a directory (`target/debug/deno test --allow-all --unstable --watch --no-check foo/`); this directory contains only a single file (say `file1_test.ts`); when I add a new file (`file2_test.ts`), only tests from that new file are run:
> 
> ```
> 
> running 1 test from file:///Users/biwanczuk/dev/deno/foo/file1_test.ts
> 
> test filesStdioFileDescriptors ... ok (7ms)
> 
> 
> 
> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (332ms)
> 
> 
> 
> Watcher Test finished! Restarting on file change...
> 
> Watcher File change detected! Restarting!
> 
> Watcher File change detected! Restarting!
> 
> running 1 test from file:///Users/biwanczuk/dev/deno/foo/file2_test.ts
> 
> test filesStdioFileDescriptors2 ... ok (7ms)
> 
> 
> 
> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (365ms)
> 
> 
> 
> Watcher Test finished! Restarting on file change...
> 
> Watcher File change detected! Restarting!
> 
> running 1 test from file:///Users/biwanczuk/dev/deno/foo/file2_test.ts
> 
> test filesStdioFileDescriptors2 ... ok (7ms)
> 
> 
> 
> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (319ms)
> 
> 
> 
> Watcher Test finished! Restarting on file change...
> 
> Watcher File change detected! Restarting!
> 
> running 1 test from file:///Users/biwanczuk/dev/deno/foo/file2_test.ts
> 
> test filesStdioFileDescriptors3 ... ok (26ms)
> 
> 
> 
> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (367ms)
> 
> 
> 
> Watcher Test finished! Restarting on file change...
> 
> Watcher File change detected! Restarting!
> 
> running 1 test from file:///Users/biwanczuk/dev/deno/foo/file1_test.ts
> 
> test filesStdioFileDescriptors10 ... ok (8ms)
> 
> 
> 
> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (373ms)
> 
> 
> 
> Watcher Test finished! Restarting on file change...
> 
> ```
> 
> Similarly changing original file (`file1_test.ts`) only reruns tests from that file. Is that the behavior you wanted to achieve? 

Yes, that's intentional. They're also rerun if any dependencies are changed, though.

caspervonb:
> > 1. When a test fails the process exits completely:
> 
> Oops, that's a bug. I'll have to fix that.
> 

Just return an error from run tests instead of exiting, a little leftover from the old runner.

bartlomieju:
@Liamolucko I did some more testing, there's no more problem of process exiting when tests fail, however I found two other problems:

1. `ctrl + c`/`ctrl + d` do not exist the process, nothing really happens:
```
failures:

	responseRedirect
	requestRelativeUrl

test result: FAILED. 746 passed; 2 failed; 17 ignored; 0 measured; 0 filtered out (15652ms)

Watcher Test finished! Restarting on file change...
^C^C
```
I'm not sure what causes that, as we don't add any custom signal handlers, any ideas?

2. It takes about 10 seconds on my machine for tests to start when using `--watch` flag, whereas without `--watch` tests start immediately (I'm running with `--no-check` flag so it's not a type checking problem). Similarly it takes 5-10s to restart the test after file change has been applied. Can it somehow be connected to debouncer in the file watcher?

If we could fix those two issues I'll be more than happy to land the PR.

Liamolucko:
> 2. It takes about 10 seconds on my machine for tests to start when using `--watch` flag, whereas without `--watch` tests start immediately (I'm running with `--no-check` flag so it's not a type checking problem). Similarly it takes 5-10s to restart the test after file change has been applied. Can it somehow be connected to debouncer in the file watcher?

Fixed in e0d23d9473aa309b696a4ac667d658ce48601493

> 1. `ctrl + c`/`ctrl + d` do not exist the process, nothing really happens:
> 
> 
> ```
> failures:
> 
> 	responseRedirect
> 	requestRelativeUrl
> 
> test result: FAILED. 746 passed; 2 failed; 17 ignored; 0 measured; 0 filtered out (15652ms)
> 
> Watcher Test finished! Restarting on file change...
> ^C^C
> ```
> 
> I'm not sure what causes that, as we don't add any custom signal handlers, any ideas?

I'm not sure what tests you were running, but when trying to debug it against `std` I found that the use of `Deno.signal` was causing the issue. With this `signal.ts`, `deno run --unstable --watch signal.ts` triggers the same issue:
```ts
Deno.signal(Deno.Signal.SIGINT);
```

bartlomieju:
> > 1. It takes about 10 seconds on my machine for tests to start when using `--watch` flag, whereas without `--watch` tests start immediately (I'm running with `--no-check` flag so it's not a type checking problem). Similarly it takes 5-10s to restart the test after file change has been applied. Can it somehow be connected to debouncer in the file watcher?
> 
> Fixed in [e0d23d9](https://github.com/denoland/deno/commit/e0d23d9473aa309b696a4ac667d658ce48601493)


Thanks, I tried again and got mixed results - sometimes it's quite fast, but for most runs it still takes 10 seconds to start. I'm fine with it for the first pass. We can optimize it in the follow up.

> 
> > 1. `ctrl + c`/`ctrl + d` do not exist the process, nothing really happens:
> > 
> > ```
> > failures:
> > 
> > 	responseRedirect
> > 	requestRelativeUrl
> > 
> > test result: FAILED. 746 passed; 2 failed; 17 ignored; 0 measured; 0 filtered out (15652ms)
> > 
> > Watcher Test finished! Restarting on file change...
> > ^C^C
> > ```
> > 
> > 
> > I'm not sure what causes that, as we don't add any custom signal handlers, any ideas?
> 
> I'm not sure what tests you were running, but when trying to debug it against `std` I found that the use of `Deno.signal` was causing the issue. With this `signal.ts`, `deno run --unstable --watch signal.ts` triggers the same issue:
> 
> ```ts
> Deno.signal(Deno.Signal.SIGINT);
> ```

I'm running `cli/tests/unit`. I found the culprit and it's `signalShorthandsTest` test, not yet sure what's going on there as all signals are disposed. CC @kt3k maybe you have some ideas what might cause signals to not be disposed properly?

cli/file_watcher.rs
```diff
@@ -29,226 +29,197 @@ use tokio::time::Sleep;
...
     self: Pin<&mut Self>,
     cx: &mut Context,
   ) -> Poll<Option<Self::Item>> {
-    if self.event_detected.swap(false, Ordering::Relaxed) {
-      Poll::Ready(Some(()))
+    let mut changed_paths = self.changed_paths.lock().unwrap();
+    if changed_paths.len() > 0 {
+      Poll::Ready(Some(changed_paths.drain().collect()))
     } else {
+      drop(changed_paths);
```
ry, line 61: Is this necessary?

cli/tools/test_runner.rs
```diff
@@ -304,37 +303,30 @@ pub async fn run_test_file(
   Ok(())
 }
 
+/// Runs tests.
+///
+/// Returns a boolean indicating whether the tests failed.
```
ry, line 308: whether the tests ran successfully (the current wording makes it seem like true = failed)

cli/tests/integration_tests.rs
```diff
@@ -999,6 +1011,174 @@ mod integration {
...
+        "import './foo.js'; export default function bar() { 2 + 2 }",
+      )
+      .expect("error writing file");
+
+      // the watcher process is still alive
+      assert!(child.try_wait().unwrap().is_none());
+
+      child.kill().unwrap();
+      drop(t);
+    }
```
ry, line 1181: Nice test

ry (approved):
LGTM - this was a difficult feature to get in, thank you for seeing it through @Liamolucko !

cli/tools/test_runner.rs
Liamolucko: `true` does mean failed, mostly just because that was the value which was already there.

caspervonb: It's returning `has_failure`.
Good enough to land imho, I'll be refactoring this soon anyway so wouldn't let it block.

bartlomieju (approved):
LGTM too 🚀  thank you very much @Liamolucko! Let's iterate further in the follow up PRs.

PR #9163, opened 2021-01-19
fix(cli/lib.deno.shared_globals): Move WorkerOptions::deno to unstable

nayeemrmn:


cli/dts/lib.deno.shared_globals.d.ts
```diff
@@ -698,51 +688,37 @@ interface WorkerOptions {
...
+  // worker options implementation. Remove for 2.0.
+  deno?: boolean;
 }
 
-declare class Worker extends EventTarget {
+// TODO(nayeemrmn): `class Worker` had to be split into `var Worker` +
+// `interface WorkerConstructor` + `interfaceWorker`, so that a constructor
+// overload could be added in `lib.deno.unstable.d.ts`. Remove on stabilization.
+declare var Worker: WorkerConstructor;
+declare interface WorkerConstructor {
```
nayeemrmn, line 711: @bartlomieju It ended being quite complex because TS can't "merge" classes from two declaration files.

However, if we're willing to compromise and remove `WorkerOptions::deno: boolean` from stable, it would be an extremely minimal change due to the lack of property conflict. Just add the following in `lib.deno.unstable.d.ts`:
```ts
declare interface WorkerOptions {
  deno?: boolean | {
    namespace: boolean,
    permissions: { ... },
  }
}
```
Since it's already runtime unstable, I think zero user workflows would break...

bartlomieju: @nayeemrmn sounds good, let's do this

bartlomieju: Also, shouldn't the interface be exported?

bartlomieju (approved):
LGTM, waiting on Kitson's review before landing

PR #9169, opened 2021-01-19
fix(installer): pass cached-only to executable_args

### LINKED ISSUE #9168: Incorrect flag for cached-only is passed to executable args in installer
I just tried to do a `deno install` on my project, and got the following error:

```bash
error: Found argument '--cached_only' which wasn't expected, or isn't valid in this context
	Did you mean --cached-only?
```

This is because in `cli/tools/installer.rs` `--cached_only` is being passed as an executable arg instead of `--cached-only`

freddyfallon:
Fixes #9168.

In `cli/tools/installer.rs` `--cached_only` is being passed as an executable arg instead of `--cached-only`

This pull request fixes that.

### DIFF
--- a/cli/tools/installer.rs
+++ b/cli/tools/installer.rs
@@ -246,7 +246,7 @@ pub fn install(
   }
 
   if flags.cached_only {
-    executable_args.push("--cached_only".to_string());
+    executable_args.push("--cached-only".to_string());
   }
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, thanks @freddyfallon

PR #7564, opened 2020-09-18
ffi continue

lygstate:
This is a pursue of https://github.com/denoland/deno/pull/5566

I wanna implement the JS version first and then add the TSD files

The std FFI implementation in rust
https://doc.rust-lang.org/std/ffi/index.html

design reference

JNR-FFI for Java(https://github.com/jnr/jnr-ffi)
```java
package helloworld;

import jnr.ffi.LibraryLoader;

public class HelloWorld {
    public static interface LibC {
        int puts(String s);
    }

    public static void main(String[] args) {
        LibC libc = LibraryLoader.create(LibC.class).load("c");

        libc.puts("Hello, World");
    }
}
```
I am curios if we can leverage the type information, so we can create a ffi library like jnr ffi does

Mozilla js-ctypes(https://developer.mozilla.org/en-US/docs/Mozilla/js-ctypes)
```js

Components.utils.import("resource://gre/modules/ctypes.jsm");

var kernel = ctypes.open("kernel32.dll");
var HMODULE = ctypes.uint32_t;
var HWND = ctypes.uint32_t;
var LPCTSTR = ctypes.jschar.ptr;
var LPCSTR = ctypes.char.ptr;
var LoadLibrary = kernel.declare("LoadLibraryW", ctypes.winapi_abi, HMODULE, LPCTSTR);
var GetProcAddress = kernel.declare("GetProcAddress", ctypes.winapi_abi, ctypes.void_t.ptr, HMODULE, LPCSTR);

var hUser = LoadLibrary("user32");
var funcPtr = GetProcAddress(hUser, "MessageBoxW");

// Now we have a pointer to a function, let's cast it to the right type
var MessageBoxType = ctypes.FunctionType(ctypes.winapi_abi, ctypes.int32_t, [HWND, LPCTSTR, LPCTSTR, ctypes.uint32_t]);
funcPtr = ctypes.cast(funcPtr, MessageBoxType.ptr);
funcPtr(0, "Test1", "Test2", 0);
```

Node (thrid-party [node-ffi](https://github.com/node-ffi/node-ffi)):
```js
var ffi = require('ffi');

var libm = ffi.Library('libm', {
  'ceil': [ 'double', [ 'double' ] ]
});
libm.ceil(1.5); // 2

// You can also access just functions in the current process by passing a null
var current = ffi.Library(null, {
  'atoi': [ 'int', [ 'string' ] ]
});
current.atoi('1234'); // 1234
```

[...]

PR #9173, opened 2021-01-19
[WIP] ffi support

manyuanrong:
FFI support based on `libffi`. 

Related issues and PRs:

#8490
#7564


mko-io:
I think the big effort of Wasm and Wasi is keeping js away from ISA dependent binary code, why bring FFI back to make it complex again? I think  someone could argue that they need FFI for low level hardware access, but it think it could provide by the runtime(browser or deno) itself for standardisation. Have a look of the new Web Serial API and Web NFC API in [Chrome 89](https://chromestatus.com/features#milestone%3D89). A restricted but "run everywhere" is unique and is its strength for deno comparing to go and rust.

bartlomieju:
Thank you for the PR @manyuanrong, we've been discussing FFI internally from time to time and will probably look into shipping it at some point. I'm gonna close this PR as it is stale and there's been a lot of changes to the ops interface since it was open, but we'll definitely circle back to it as a reference point if we decide to implement FFI.

PR #9174, opened 2021-01-19
chore(compile): add aarch64-apple-darwin target

lucacasonato:


ry (commented):
But we don't have canary builds on this..?

lucacasonato:
> But we don't have canary builds on this..?

So what? M1 is still experimental, and if this feature happens to not work in canary, I personally don't really care... `deno upgrade --canary` also doesn't work on M1 right? As long as it works in releases (and it will if we upload the binaries to dl.deno.land) it will be fine.

ry (approved):

PR #9179, opened 2021-01-20
Update lspower; fix breakage

hvithrafn:
This PR updates the `lspower` dependency for the language server.

This is just a draft since I haven't pushed the new version of `lspower` yet but wanted to see if there any problems with the changes.


kitsonk:
I've already taken a look at it and it seems the right direction.  I just won't want to formally review it until lspower update is released and this PR is no longer draft.

hvithrafn:
One thing to note is that with the next version of `lspower` the `tokio` dependency can be optional.

In other words, it's possible to run the service with other async runtimes, like `smol`, or just with the basic futures executor. This is mostly interesting for portability reasons and was prompted by needing to be able to run `lspower`-based servers directly in a browser context.

I haven't exposed that functionality in this PR (by default, `lspower` will just use `tokio`) but could add that if there is interest.

kitsonk:
@hvithrafn thanks, but we are focusing on continuing to use tokio (1.0 now) in Deno for the foreseeable future.

hvithrafn:
@kitsonk this should be good to go I think.

kitsonk (approved):
LGTM

Cargo.lock
```diff
@@ -292,6 +324,15 @@ dependencies = [
...
 
+[[package]]
+name = "concurrent-queue"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30ed07550be01594c6026cff2a1d7fe9c8f683caa798e12b68694ac9e88286a3"
+dependencies = [
+ "cache-padded",
+]
+
```
ry, line 335: It's unfortunate this change adds so many dependencies.

lucacasonato: Yeah. It seems async-channel and async-oneshot were added to be able to remove tokio from the dependencies. Not great...

hvithrafn: Oh, I didn't notice that. I might be able to trim down the dependencies a bit for default case with tokio. I'll give it a try.

hvithrafn: Actually, with the `async-channel` and `async-oneshot`, those were added because originally I was going to try and move away from the full `futures` crate toward `futures-lite`. I think I had some issues at one point in the refactoring and never quite got around to that.

But it seems that I can just roll back those particular changes for now and consider whether it's worth investigating later on. I think that `async-*` packages were potentially more performant but it probably isn't even noticeable for an application like LSP because there isn't enough throughput.

hvithrafn: @ry okay, I was able to remove some of these dependencies. It looks like the remaining ones that were added come from the updated `nom` dependency.

ry (approved):
LGTM - thanks for cleaning up the dependencies @hvithrafn !

PR #9187, opened 2021-01-20
fix(ops/net): fix panic in op_dns_resolve

### LINKED ISSUE #9186: Deno.resolveDns panics with non registered domain records
When I run the command
```
deno run -A --unstable https://deno.land/posts/v1.7/dig.ts not-found-example.com
```

I see the following panic.
```
thread 'main' panicked at 'Error 'no record found for name: not-found-example.com.local. type: A class: IN' contains boxed error of unknown type:
  ResolveError { kind: NoRecordsFound { query: Query { name: Name { is_fqdn: true, labels: [not-found-example, com, local] }, query_type: A, query_class: IN }, soa: Some(SOA { mname: Name { is_fqdn: true, labels: [a, root-servers, net] }, rname: Name { is_fqdn: true, labels: [nstld, verisign-grs, com] }, serial: 2021012000, refresh: 1800, retry: 900, expire: 604800, minimum: 86400 }), negative_ttl: Some(86394), response_code: NXDomain, trusted: true } }', cli/errors.rs:35:7
```

I think this should be at least javascript error. Or it might return Ok result with empty list?

- deno version: 1.7.0

kt3k:
Deno.resolveDns() panics with non registered dns record. This PR changes it to JS error.

```shellsession
$ ./target/debug/deno run --unstable -A https://deno.land/posts/v1.7/dig.ts not-found-example.com
error: Uncaught (in promise) Error: no record found for name: not-found-example.com.local. type: A class: IN
    throw new ErrorClass(res.err.message);
          ^
    at processResponse (deno:core/core.js:212:11)
    at Object.jsonOpAsync (deno:core/core.js:229:12)
    at async https://deno.land/posts/v1.7/dig.ts:6:17
```

closes #9186

### DIFF
--- a/cli/tests/resolve_dns.ts
+++ b/cli/tests/resolve_dns.ts
@@ -34,3 +34,9 @@ console.log(JSON.stringify(srv));
 
 console.log("TXT");
 console.log(JSON.stringify(txt));
+
+try {
+  await Deno.resolveDns("not-found-example.com", "A", nameServer);
+} catch (e) {
+  console.log("Error thrown for not-found-example.com");
+}
--- a/cli/tests/resolve_dns.ts.out
+++ b/cli/tests/resolve_dns.ts.out
@@ -14,3 +14,4 @@ SRV
 [{"priority":0,"weight":100,"port":1234,"target":"srv.com."}]
 TXT
 [["foo","bar"]]
+Error thrown for not-found-example.com
--- a/runtime/ops/net.rs
+++ b/runtime/ops/net.rs
@@ -629,7 +629,8 @@ async fn op_dns_resolve(
 
   let results: Vec<DnsReturnRecord> = resolver
     .lookup(query, record_type, Default::default())
-    .await?
+    .await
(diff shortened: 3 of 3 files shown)


magurotuna:
Thanks so much for the fix! I should have taken this case into account...

I think this is fine for the moment, but in the future, at latest before this API gets stabilized, it should return the cause of the error as precise as it can. 
Looking at the other languages, in the case of failure, Golang returns [DNSError type](https://golang.org/pkg/net/#DNSError), and Node.js returns `Error` object with [these error codes](https://nodejs.org/api/dns.html#dns_error_codes). It would be nice if we returned errors likewise.

kt3k:
> Looking at the other languages, in the case of failure, Golang returns DNSError type, and Node.js returns Error object with these error codes. It would be nice if we returned errors likewise.

Sounds reasonable to me 👍 Let's do that in follow up issues.

ry (approved):
LGTM

PR #9196, opened 2021-01-20
docs(typescript): Implied tsconfig uses "useDefineForClassFields"

robpalme:
Deno hard-codes `"useDefineForClassFields": true`

Updating the docs to reflect this.


kitsonk (approved):
Ah, I forgot we did that... thanks for the catch.

LGTM

PR #9207, opened 2021-01-21
fix(Deno.exit): fix recursive dispatching of unload event

### LINKED ISSUE #9201: Return an exit code on "unload"
Hi.
I have a function that execute some stuff. And I want to return different exit codes, depending on the result of this execution. Something like this:

```js
const success = await run();
Deno.exit(success ? 0 : 1);
```

Sometimes is impossible to know when `run()` finished (because there may be async functions that don't return a promise), so `Deno.exit()` can break the execution of some things inside `run()`. To prevent that, I fixed it with this code:

```js
const success = await run();
window.addEventListener("unload", () => Deno.exit(success ? 0 : 1));
```

This ensures that only after finish all process, subprocess, etc inside `run()`, the proper exit code is returned.

After upgrading to Deno < 1.7 I'm getting the following error: `error: Uncaught RangeError: Maximum call stack size exceeded`. I guess this is because #9088, so `Deno.exit()` dispatches `unload` event that executed `Deno.exit()` that dispatches `unload` and so on.

To prevent this, I think `unload` should be triggered only once. Or if there's any other way to set an exit code without using `Deno.exit()`, that would be fine too.

kt3k:
We introduced the dispatching of `unload` event in `Deno.exit()` in #9088, but that caused the recursive dispatches of the event when it's called in `unload` event handler (ref #9201 ).

This PR checks whether `unload` is triggered before dispatching it in `Deno.exit()`, and prevents the recursive dispatching of `unload` event.

fixes #9201

### DIFF
--- a/cli/tests/078_unload_on_exit.ts
+++ b/cli/tests/078_unload_on_exit.ts
@@ -2,7 +2,7 @@ window.onunload = () => {
   console.log("onunload is called");
   // This second exit call doesn't trigger unload event,
   // and therefore actually stops the process.
-  Deno.exit(0);
+  Deno.exit(1);
   console.log("This doesn't show up in console");
 };
 // This exit call triggers the above unload event handler.
--- a/cli/tests/078_unload_on_exit.ts
+++ b/cli/tests/078_unload_on_exit.ts
@@ -1,4 +1,9 @@
 window.onunload = () => {
   console.log("onunload is called");
+  // This second exit call doesn't trigger unload event,
+  // and therefore actually stops the process.
+  Deno.exit(0);
+  console.log("This doesn't show up in console");
 };
+// This exit call triggers the above unload event handler.
 Deno.exit(0);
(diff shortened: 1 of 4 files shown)


cli/tests/078_unload_on_exit.ts
```diff
@@ -1,4 +1,9 @@
 window.onunload = () => {
   console.log("onunload is called");
+  // This second exit call doesn't trigger unload event,
+  // and therefore actually stops the process.
+  Deno.exit(0);
```
kitsonk, line 5: It might also be good to use a different exit code here to help "prove" this worked?

kitsonk (commented):
just one thought!

cli/tests/078_unload_on_exit.ts
kt3k: Sounds good! Updated.

kitsonk (approved):
LGTM

PR #9209, opened 2021-01-21
feat(lsp): add performance measurements

kitsonk:
This PR adds performance measurements to the internals of the LSP, which are then available via the language services virtual status document.  In VSCode the output looks something like this:

Cursor_and_status_md_—_oak

This will help us better determine the performance of the lsp and any bottlenecks. It is only instrumented in the language server, and not into the internals of the TypeScript integration for example.

### DIFF
--- a/cli/lsp/performance.rs
+++ b/cli/lsp/performance.rs
@@ -1,0 +1,179 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+use std::collections::HashMap;
+use std::collections::VecDeque;
+use std::time::Duration;
+use std::time::Instant;
+
+/// A structure which serves as a start of a measurement span.
+#[derive(Debug)]
+pub struct PerformanceMark {
+  name: String,
+  count: u32,
+  start: Instant,
+}
+
+/// A structure which holds the information about the measured span.
+#[derive(Debug)]
+pub struct PerformanceMeasure {
+  pub name: String,
+  pub count: u32,
+  pub duration: Duration,
+}
+
+impl From<PerformanceMark> for PerformanceMeasure {
+  fn from(value: PerformanceMark) -> Self {
+    Self {
+      name: value.name,
+      count: value.count,
+      duration: value.start.elapsed(),
+    }
+  }
+}
+
+/// A simple structure for marking a start of something to measure the duration
+/// of and measuring that duration.  Each measurement is identified by a string
+/// name and a counter is incremented each time a new measurement is marked.
+///
+/// The structure will limit the size of measurements to the most recent 1000,
+/// and will roll off when that limit is reached.
+#[derive(Debug)]
+pub struct Performance {
+  counts: HashMap<String, u32>,
+  measures: VecDeque<PerformanceMeasure>,
+  size: usize,
+}
+
+impl Default for Performance {
+  fn default() -> Self {
+    Self {
+      counts: Default::default(),
+      measures: Default::default(),
+      size: 1_000,
+    }
+  }
+}
+
+impl Performance {
+  /// Return the count and average duration of a measurement identified by name.
+  #[allow(unused)]
+  pub fn average(&self, name: &str) -> (usize, Duration) {
+    let mut items = Vec::new();
+    for measure in &self.measures {
+      if measure.name == name {
+        items.push(measure.duration);
+      }
+    }
+    let len = items.len();
+    let average = if len > 0 {
+      items.into_iter().sum::<Duration>() / len as u32
+    } else {
+      Duration::default()
+    };
+    (len, average)
+  }
+
+  /// Return an iterator which provides the names, count, and average duration
+  /// of each measurement.
+  pub fn averages(&self) -> impl Iterator<Item = (String, usize, Duration)> {
+    let mut averages: HashMap<String, Vec<Duration>> = HashMap::new();
+    for measure in &self.measures {
+      averages
+        .entry(measure.name.clone())
+        .or_default()
+        .push(measure.duration);
+    }
+    averages.into_iter().map(|(k, d)| {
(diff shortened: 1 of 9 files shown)


ry (commented):
Is there a way we can export these to the benchmarks page?

kitsonk:
> Is there a way we can export these to the benchmarks page?

Yes.  We can add an JSONRPC call that returns the data as a JSON payload, which we then could call from the benchmarking.  I will add that to this PR, but we would still need to do a harness for the lsp for benchmarking that is different than what we have for our current integration tests, which I think we should do separately.

kitsonk:
BAH... I tried this PR and I am running into deadlock issues, like when trying to write the resources...  Need to figure out how to do it better.  I think I know what I can do, but for now, we can't merge this.

crypt096 (approved):

kitsonk:
Ok, so I refactored it a bit, so that internally to performance there are locks instead of locking the whole of performance, this should prevent deadlocks for performance because marking and measuring are now independent, meaning that they can't get into a circular situation of needing a lock released.

I tried to re-create the issue before fixing it, but it seems like it is hard/impossible with our test harness re-create this deadlock, though I improved the "heavy" integration test to do more hoping to create a deadlock.

cli/lsp/performance.rs
```diff
@@ -0,0 +1,165 @@
...
+      items.into_iter().sum::<Duration>() / len as u32
+    } else {
+      Duration::default()
+    };
+    (len, average)
+  }
+
+  /// Return an iterator which provides the names, count, and average duration
+  /// of each measurement.
+  pub fn averages(&self) -> impl Iterator<Item = (String, usize, Duration)> {
```
ry, line 90: Return `Vec<PerformanceAverage>`

cli/lsp/performance.rs
```diff
@@ -0,0 +1,165 @@
...
+      counts: Default::default(),
+      measures: Default::default(),
+      size: 1_000,
+    }
+  }
+}
+
+impl Performance {
+  /// Return the count and average duration of a measurement identified by name.
+  #[allow(unused)]
```
ry, line 71: probably you want `#[cfg(test)]` ?

cli/lsp/performance.rs
```diff
@@ -0,0 +1,165 @@
...
+    for measure in self.measures.lock().unwrap().iter() {
+      if measure.name == name {
+        items.push(measure.duration);
+      }
+    }
+    let len = items.len();
+    let average = if len > 0 {
+      items.into_iter().sum::<Duration>() / len as u32
+    } else {
+      Duration::default()
```
ry, line 83: averaging over zero entries should panic

cli/lsp/performance.rs
```diff
@@ -0,0 +1,165 @@
...
+  pub fn averages(&self) -> impl Iterator<Item = (String, usize, Duration)> {
+    let mut averages: HashMap<String, Vec<Duration>> = HashMap::new();
+    for measure in self.measures.lock().unwrap().iter() {
+      averages
+        .entry(measure.name.clone())
+        .or_default()
+        .push(measure.duration);
+    }
+    averages.into_iter().map(|(k, d)| {
+      let a = d.clone().into_iter().sum::<Duration>() / d.len() as u32;
```
ry, line 99: I'm a bit worried this is integer division which gets rounded... Does  `d.clone().into_iter().sum::<Duration>() ` have the type `Duration`? What does it mean to divide a Duration by a u32 ?

ry (commented):
I'm a bit worried that a lot of code is used to add this "Performance" machinery, but when we actually go to measure it, we could as well just start up `deno lsp` from the benchmark program and pump some stuff into it and measure how long it takes to respond.

cli/lsp/performance.rs
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+
+/// A simple structure for marking a start of something to measure the duration
+/// of and measuring that duration.  Each measurement is identified by a string
+/// name and a counter is incremented each time a new measurement is marked.
+///
+/// The structure will limit the size of measurements to the most recent 1000,
+/// and will roll off when that limit is reached.
+#[derive(Debug, Clone)]
```
ry, line 52: Does it need to be Clone? This will duplicate the size value...

cli/lsp/performance.rs
```diff
@@ -0,0 +1,165 @@
...
+pub struct Performance {
+  counts: Arc<Mutex<HashMap<String, u32>>>,
+  measures: Arc<Mutex<VecDeque<PerformanceMeasure>>>,
+  size: usize,
```
ry, line 56: maybe call this max_size ?

cli/lsp/performance.rs
kitsonk: yes, because `LanguageServer` is `Clone` and all of its properties therefore need to be `Clone`.

kitsonk:
> I'm a bit worried that a lot of code is used to add this "Performance" machinery, but when we actually go to measure it, we could as well just start up `deno lsp` from the benchmark program and pump some stuff into it and measure how long it takes to respond.

That was not the intent of this PR, to create a benchmark.  The intent of this PR is to provide instrumentation for the lsp which gives visibility into what is actually happening inside the process to be able to ensure we can maintain performance.  Benchmarking is secondary... it is a trailing indicator of a problem and provides you no information on the root cause of the problem.

cli/lsp/performance.rs
kitsonk: but `LanguageServer` appears not to require clone, so I will remove both.

cli/lsp/performance.rs
kitsonk: `Duration` has a `Div<u32>` trait.  It is the appropriate of "averaging" as `Duration`.

ry (approved):
LGTM

PR #9217, opened 2021-01-21
fix(std/async): Make pooledMap() errors catchable

### LINKED ISSUE #9070: std/async: Uncaught exception in pooledMap
When an exception occurs in an async callback in `pooledMap`, the exception is caught by surrounding try/catch but it _also_ bubbles into an uncaught Promise Exception and crashes the process.

Example stack trace:
```text
2021-01-05T09:57:47.331+01:00	error: Uncaught (in promise) Http: error sending request for url (https://REDACTED.s3.us-east-1.amazonaws.com/aws_sdk/versions/v3.0.0.1/raw/client-codebuild/pagination/ListBuildBatchesForProjectPaginator.ts): connection closed before message completed
2021-01-05T09:57:47.331+01:00	at processResponse (deno:core/core.js:223:11)
2021-01-05T09:57:47.331+01:00	at Object.jsonOpAsync (deno:core/core.js:240:12)
2021-01-05T09:57:47.331+01:00	at async fetch (deno:op_crates/fetch/26_fetch.js:1278:29)
2021-01-05T09:57:47.331+01:00	at async S3Bucket.putObject (<https://deno.land/x/s3@0.2.0/src/bucket.ts>:352:1)
2021-01-05T09:57:47.331+01:00	at async uploadVersionRaw (<file:///var/task/utils/storage.ts>:82:1)
2021-01-05T09:57:47.331+01:00	at async <file:///var/task/api/async/publish.ts>:170:1
```

I also made a simple gist to reproduce the issue: https://gist.github.com/wperron/1066ad43a77a3cd7aeb3bdecfa44d9d1

cc @lucacasonato

nayeemrmn:
Fixes #9070.

cc @lucacasonato

### DIFF
--- a/std/async/pool.ts
+++ b/std/async/pool.ts
@@ -59,11 +59,9 @@ export function pooledMap<T, R>(
           errors.push(result.reason);
         }
       }
-      const p = Promise.reject(
+      writer.write(Promise.reject(
         new AggregateError(errors, "Threw while mapping."),
-      );
-      p.catch(() => {});
-      writer.write(p).catch(() => {});
+      )).catch(() => {});
     }
   })();
   return res.readable[Symbol.asyncIterator]();
--- a/std/async/pool.ts
+++ b/std/async/pool.ts
@@ -4,8 +4,13 @@
  * pooledMap transforms values from an (async) iterable into another async
  * iterable. The transforms are done concurrently, with a max concurrency
  * defined by the poolLimit.
- * 
- * @param poolLimit The maximum count of items being processed concurrently. 
+ *
+ * If an error is thrown from `iterableFn`, no new transformations will begin.
+ * All currently executing transformations are allowed to finish and still
+ * yielded on success. After that, the rejections among them are gathered and
+ * thrown by the iterator in an `AggregateError`.
+ *
+ * @param poolLimit The maximum count of items being processed concurrently.
  * @param array The input array for mapping.
  * @param iteratorFn The function to call for every item of the array.
  */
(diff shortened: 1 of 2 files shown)


std/async/pool.ts
```diff
@@ -4,8 +4,13 @@
  * pooledMap transforms values from an (async) iterable into another async
  * iterable. The transforms are done concurrently, with a max concurrency
  * defined by the poolLimit.
- * 
- * @param poolLimit The maximum count of items being processed concurrently. 
+ *
+ * If an error is thrown from `iterableFn`, no new transformations will begin.
+ * All currently executing transformations are allowed to finish and still
+ * yielded on success. After that, the rejections among them are gathered and
+ * thrown by the iterator in an `AggregateError`.
```
nayeemrmn, line 11: On one hand we want resource sanitation, so we let in progress jobs finish. On the other hand we want to fail fast like JS built-ins do, so we stop starting new jobs on error instead of trying all items and reporting all errors..

caspervonb: SGTM

std/async/pool.ts
```diff
@@ -27,20 +32,39 @@ export function pooledMap<T, R>(
...
+          executing.splice(executing.indexOf(e), 1)
+        );
+        executing.push(e);
+        if (executing.length >= poolLimit) {
+          await Promise.race(executing);
+        }
+      }
+      // Wait until all ongoing events have processed, then close the writer.
+      await Promise.all(executing);
+      writer.close();
```
kt3k, line 54: how about doing this in `finally`?

std/async/pool.ts
```diff
@@ -27,20 +32,39 @@ export function pooledMap<T, R>(
...
+      const errors = [];
+      for (const result of await Promise.allSettled(executing)) {
+        if (result.status == "rejected") {
+          errors.push(result.reason);
+        }
       }
+      const p = Promise.reject(
+        new AggregateError(errors, "Threw while mapping."),
+      );
+      p.catch(() => {});
```
kt3k, line 65: I think we shouldn't attach catch handler here because this error can be handled by the user of pooledMap.

std/async/pool.ts
nayeemrmn: It closes on error, so `writer.close()` will throw if done in the catch branch.

std/async/pool.ts
nayeemrmn: When we `writer.write()` this rejected promise, it causes the iterator to throw it to the user. An empty catch handler because it is treated like an unhandled rejection otherwise.

kt3k: I commented out this line and tried the script below:
```ts
import { pooledMap } from "./std/async/pool.ts";
let iter = pooledMap(4, [...Array(100)], () => {
  throw new Error();
});
try {
  for await (const _ of iter) {}
} catch (e) {
  console.log(e.errors);
}
console.log("finish");
```
and this script seems finishing without unhandled rejection error. Does this mean the error can be handled by the user?

On the other hand, I confirmed that the empty handler at the next line is necessary to avoid unhandled rejection error.

nayeemrmn: You're right, `writer.write()`'s implementation properly inherits the rejection handling so it wouldn't be needed. I think I tested it before I added the handler below.

kt3k (approved):
LGTM.

@lucacasonato I think this is ready to land now. PTAL.

lucacasonato (approved):
Implementation looks good. Thanks for the fix @nayeemrmn.

PR #8691, opened 2020-12-09
chore: fixed typo

hong4rc:
The test fails #8556

I only change the comment, doc and local variable name

PR #9221, opened 2021-01-21
chore(std/encoding/csv_stringify): Revert typo introduced in #8691

jsejcksn:
Ref. https://github.com/denoland/deno/pull/8691/commits/875852bd47c53a5f9fd1c5c01c69916969fd8696


ry (approved):
Thanks

PR #7242, opened 2020-08-29
BREAKING(unstable): remove Deno.SymlinkOptions

caspervonb:
This removes Deno.SymlinkOptions which is leaking internal implementation details on Windows which encourages checking and special casing when running on Windows.

Closes #7234

PR #9226, opened 2021-01-22
feat(runtime): stabilize Deno.symlink and Deno.symlinkSync

caspervonb:
This makes Deno.symlink and Deno.symlinkSync stable.

These interfaces are well known; I've previously tried to drop the type option (https://github.com/denoland/deno/pull/7242) and we aren't going to do that so I don't see this interface changing at all in the future.


caspervonb:
cc @piscisaureus @ry

caspervonb:
As for the security review part; sandboxing of symlinks should be handled when they're followed not created. 

WASI does the same, you can create a symlink to outside of the sandbox as a dangling symlink but you cannot verify that said symlink is valid without having the proper capabilities.

Also matches the permission semantics of symlink(2).
 
> The permissions of a symbolic link are irrelevant; the ownership
> is ignored when following the link, but is checked when removal
> or renaming of the link is requested and the link is in a
> directory with the sticky bit (S_ISVTX) set.

ry:
Do we need read/write permissions on both ends of the link?

ry (approved):
Thank you for the security review - makes sense to me.

Let's land it.

Thanks Capser.

PR #9237, opened 2021-01-23
fix(std/node): replace uses of `window` with `globalThis`

Liamolucko:
`module.ts` and `timers.ts` wouldn't previously work in workers, since `window` isn't defined.


lucacasonato (approved):
LGTM, thanks

Issue #9225, opened 2021-01-21
[STD/FS] createSymlink creates invalid symlink to another directory


ilovewine:
**ENV**: Linux

**Steps to reproduce**:
- go to ./folder (in terminal)
- create ./folderTarget/target folder inside ./folder (e.g. mkdir ./folderTarget/target)
- create the script symlink.ts inside ./folder:
```
import {
  ensureSymlink,
} from "https://deno.land/std@0.80.0/fs/mod.ts";

ensureSymlink('./folderTarget/target','./folder/foo');
```
- run deno run --unstable --allow-read --allow-write symlink.ts
**Expected result**: the symlink ./folder/foo is an appropriate symlink and it leads to the target folder
**Actual result**: the symlink is broken (i.e. it doesn't lead to the target folder)

ilovewine:
#9224 - This is an equivalent for a synchronous method.

pevers:
I was looking into this and found out that it happens because the symlinks have relative paths. We should canonicalize them: https://doc.rust-lang.org/std/fs/fn.canonicalize.html to fix it.

I'll open a PR.

pevers:
I'm not sure if this is a bug on second thoughts. If we stick close to the OS:

Doing:
```shell
$ mkdir folder
$ ln -s ./folderTarget/target folder/foo
```

Would have the same results (a broken symlink). 

fs-extra implements a 'smart' strategy: https://github.com/jprichardson/node-fs-extra/blob/6bffcd81881ae474d3d1765be7dd389b5edfd0e0/lib/ensure/symlink-paths.js

It will check if the path is relative to the current working directory. That could work for Deno as well. OR, we add it to the docs that the symlink is relative to the link's parent directory as Node did: https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback .

caspervonb:
Working as intended IMO, your paths are just messed up.

lucacasonato:
This is not a bug. Behavior is intended.

caspervonb:
To expand on this a little bit; this is a system call binding and as such it behaves like the system call.

> Symbolic links are interpreted at run time as if the contents of
> the link had been substituted into the path being followed to find a file or directory.
>
> Symbolic links may contain ..  path components, which (if used at
> the start of the link) refer to the parent directories of that in
> which the link resides.
>
> A symbolic link (also known as a soft link) may point to an
> existing file or to a nonexistent one; the latter case is known
> as a dangling link.

[...]

Issue #9224, opened 2021-01-21
[STD/FS] createSymlinkSync creates invalid symlink to another directory


ilovewine:
**ENV**: Linux

**Steps to reproduce**:
- go to ./folder (in terminal)
- create ./folderTarget/target folder inside ./folder (e.g. mkdir ./folderTarget/target)
- create the script symlink.ts inside ./folder:
```
import {
  ensureSymlinkSync,
} from "https://deno.land/std@0.80.0/fs/mod.ts";

ensureSymlinkSync('./folderTarget/target','./folder/foo');
```
- run deno run --unstable --allow-read --allow-write symlink.ts
**Expected result**: the symlink ./folder/foo is an appropriate symlink and it leads to the target folder
**Actual result**: the symlink is broken (i.e. it doesn't lead to the target folder)

lucacasonato:
This is not a bug. Behavior is intended.

PR #9239, opened 2021-01-23
fix (ops/fs): Fix relative symlinks

pevers:
This PR fixes broken relative symlinks. Relative symlinks could be broken depending on the cwd of Deno. Therefore I canonicalized the source directory.

This closes https://github.com/denoland/deno/issues/9225 and https://github.com/denoland/deno/issues/9224 .


pevers:
I don't think this is a good solution anymore. We should stick as close as possible to the system implementation. That allows to create symbolic links from non-existing files.

NodeJS take on this: https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback

"Relative targets are relative to the link’s parent directory."

We should just add that to the docs.

PR #9240, opened 2021-01-23
feat(unstable): per op metrics

lucacasonato:
This PR implements per op metrics. This means that you can see all of the metrics you can currently, just per op. This is just a proof of concept that I threw together after a chat with @wperron. This, or something similar would be great for in production monitoring of Deno applications. I would imagine someone could build a prometheus exporter on top of this, so you could display this info in Grafana, or AWS CloudWatch.

The main downside of this PR is the extra memory and performance overhead for each ops. I am unsure how noticable this will really be (currently the extra overhead is just a simple string clone & a hashmap lookup). This could be improved by keying the metrics hashmap using the u64 op ID instead of a string like we do now. This would require that the calling op id is somehow exposed to the metrics exporter though, which is currently not the case.

In the current implementation, per op metrics are always logged, but they are only exposed to the runtime when `--unstable` is specified.

Feedback very welcome.


ry (commented):
I tried it - seems very nice - I'm in favor. I'd like to get a second opinion from @piscisaureus and/or @bnoordhuis.

This PR would need tests too of course, but I think you're still at the gathering feedback stage.

```
> ./target/debug/deno repl --unstable
Deno 1.7.0
exit using ctrl+d or close()
> Deno.metrics()
{
  opsDispatched: 0,
  opsDispatchedSync: 0,
  opsDispatchedAsync: 0,
  opsDispatchedAsyncUnref: 0,
  opsCompleted: 0,
  opsCompletedSync: 0,
  opsCompletedAsync: 0,
  opsCompletedAsyncUnref: 0,
  bytesSentControl: 0,
  bytesSentData: 0,
  bytesReceived: 0,
  ops: {}
}
> Deno.metrics()
{
  opsDispatched: 1,
  opsDispatchedSync: 1,
  opsDispatchedAsync: 0,
  opsDispatchedAsyncUnref: 0,
  opsCompleted: 1,
  opsCompletedSync: 1,
  opsCompletedAsync: 0,
  opsCompletedAsyncUnref: 0,
  bytesSentControl: 2,
  bytesSentData: 0,
  bytesReceived: 282,
  ops: {
    op_metrics: {
      opsDispatched: 1,
      opsDispatchedSync: 1,
      opsDispatchedAsync: 0,
      opsDispatchedAsyncUnref: 0,
      opsCompleted: 1,
      opsCompletedSync: 1,
      opsCompletedAsync: 0,
      opsCompletedAsyncUnref: 0,
      bytesSentControl: 2,
      bytesSentData: 0,
      bytesReceived: 282
    }
  }
}
> Deno.metrics()
{
  opsDispatched: 2,
  opsDispatchedSync: 2,
  opsDispatchedAsync: 0,
  opsDispatchedAsyncUnref: 0,
  opsCompleted: 2,
  opsCompletedSync: 2,
  opsCompletedAsync: 0,
  opsCompletedAsyncUnref: 0,
  bytesSentControl: 4,
  bytesSentData: 0,
  bytesReceived: 817,
  ops: {
    op_metrics: {
      opsDispatched: 2,
      opsDispatchedSync: 2,
      opsDispatchedAsync: 0,
      opsDispatchedAsyncUnref: 0,
      opsCompleted: 2,
      opsCompletedSync: 2,
      opsCompletedAsync: 0,
      opsCompletedAsyncUnref: 0,
      bytesSentControl: 4,
      bytesSentData: 0,
      bytesReceived: 817
    }
  }
}
> Deno.metrics()
{
  opsDispatched: 3,
  opsDispatchedSync: 3,
  opsDispatchedAsync: 0,
  opsDispatchedAsyncUnref: 0,
  opsCompleted: 3,
  opsCompletedSync: 3,
  opsCompletedAsync: 0,
  opsCompletedAsyncUnref: 0,
  bytesSentControl: 6,
  bytesSentData: 0,
  bytesReceived: 1352,
  ops: {
    op_metrics: {
      opsDispatched: 3,
      opsDispatchedSync: 3,
      opsDispatchedAsync: 0,
      opsDispatchedAsyncUnref: 0,
      opsCompleted: 3,
      opsCompletedSync: 3,
      opsCompletedAsync: 0,
      opsCompletedAsyncUnref: 0,
      bytesSentControl: 6,
      bytesSentData: 0,
      bytesReceived: 1352
    }
  }
}
> Deno.cwd()
"/Users/ry/src/deno"
> Deno.metrics()
{
  opsDispatched: 5,
  opsDispatchedSync: 5,
  opsDispatchedAsync: 0,
  opsDispatchedAsyncUnref: 0,
  opsCompleted: 5,
  opsCompletedSync: 5,
  opsCompletedAsync: 0,
  opsCompletedAsyncUnref: 0,
  bytesSentControl: 10,
  bytesSentData: 0,
  bytesReceived: 1933,
  ops: {
    op_metrics: {
      opsDispatched: 4,
      opsDispatchedSync: 4,
      opsDispatchedAsync: 0,
      opsDispatchedAsyncUnref: 0,
      opsCompleted: 4,
      opsCompletedSync: 4,
      opsCompletedAsync: 0,
      opsCompletedAsyncUnref: 0,
      bytesSentControl: 8,
      bytesSentData: 0,
      bytesReceived: 1889
    },
    op_cwd: {
      opsDispatched: 1,
      opsDispatchedSync: 1,
      opsDispatchedAsync: 0,
      opsDispatchedAsyncUnref: 0,
      opsCompleted: 1,
      opsCompletedSync: 1,
      opsCompletedAsync: 0,
      opsCompletedAsyncUnref: 0,
      bytesSentControl: 2,
      bytesSentData: 0,
      bytesReceived: 44
    }
  }
}
```

runtime/metrics.rs
```diff
@@ -1,6 +1,36 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+use serde::Serialize;
+
 #[derive(Default, Debug)]
-pub struct Metrics {
+pub struct RuntimeMetrics {
+  pub ops: HashMap<String, OpMetrics>,
```
bnoordhuis, line 6: If you change the type to `HashMap<&'static str, OpMetrics>`, you don't have to to_string() + clone() everywhere.

(Requires a change to `runtime/ops/plugin.rs` but you could `Box::leak()` it if you don't want to break API.)

bnoordhuis (approved):

runtime/metrics.rs
lucacasonato: > If you change the type to HashMap<&'static str, OpMetrics>, you don't have to to_string() + clone() everywhere.

Yeah. This was my initial implementation, but then I got stuck on `runtime/ops/plugin.rs`. Using `Box::leak` seems to work. Didn't know about that before. Thanks :-)

runtime/metrics.rs
```diff
@@ -108,7 +146,8 @@ pub fn metrics_op(op_fn: Box<OpFn>) -> Box<OpFn> {
         let fut = fut
           .inspect(move |buf| {
             let mut s = op_state_.borrow_mut();
-            let metrics = s.borrow_mut::<Metrics>();
+            let runtime_metrics = s.borrow_mut::<RuntimeMetrics>();
+            let metrics = runtime_metrics.ops.get_mut(name).unwrap();
```
piscisaureus, line 150: Is it possible to key ops by OpId rather than by name?
Seems more efficient to do it that way.

lucacasonato: See my comment in the PR description:

> This could be improved by keying the metrics hashmap using the u64 op ID instead of a string like we do now. This would require that the calling op id is somehow exposed to the metrics exporter though, which is currently not the case.

danopia:
I'm running this patch on some HTTP-client workloads and it's pretty nice seeing the Deno operation breakdown collated into monitoring software. (Note that I'm trimming some bits off the op names after getting them from `Deno.metrics()`)

Screenshot 2021-01-29 at 09 38 56

lucacasonato:
@ry I have added a test. Please re-review. Shall we hold this for 1.8 or can we merge it for 1.7.2 (it is `--unstable`)?

runtime/js/30_metrics.js
```diff
@@ -4,7 +4,11 @@
   const core = window.Deno.core;
 
   function metrics() {
-    return core.jsonOpSync("op_metrics");
+    const { combined, ops } = core.jsonOpSync("op_metrics");
+    if (ops) {
+      combined.ops = ops;
+    }
```
bartlomieju, line 10: When is this condition false?

lucacasonato: If unstable is not enabled.

bartlomieju: So the interface will change once again, once we stabilize this feature? (presumably for 2.0 as it would be a breaking change)

bartlomieju:
Would be great to add similar example to one provided by @danopia to the manual

runtime/js/30_metrics.js
lucacasonato: Indeed. That was my thinking.

bartlomieju (approved):
LGTM

PR #9242, opened 2021-01-23
feat(op_crates/web): adding gb18030 and GBK encodings

AliBasicCoder:
this PR adds gb18030 and GBK encodings to TextDcoder

NOTE: looks like tests for gb18030 and GBK in WPT doesn't run at all (i guess because there are in a html file)
so i made a js version of the tests and it worked
should i add them to this PR (in op_crates)?


lucacasonato:
> so i made a js version of the tests and it worked

I will upstream the tests as a .any.js version to wpt tomorrow. You should then just be able to run them as normal.

AliBasicCoder:
at cli/tests/unit/text_encoding_test.ts: 114:3 this uses gbk as a test for enternal decoder error

lucacasonato (approved):
LGTM. Requesting additional review from @bartlomieju

bartlomieju (approved):

lucacasonato:
As per discussion with @bartlomieju, we will merge this for 1.7.1, as this a Web API compat feature, not a new runtime capability.

Thank you @AliBasicCoder!

PR #9249, opened 2021-01-24
Change release build flags to optimize for performance

evanwashere:
improves perf of deno ops/core calls that are not i/o bound at the cost of release build time
haven't tested on other platforms but on macOS arm deno cli is ~900kb larger and denort is 1mb smaller

last change to opt-level #6907

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,9 @@ exclude = [
 [profile.release]
 codegen-units = 1
 lto = true
-opt-level = 'z' # Optimize for size
+opt-level = 3 # Optimize for performance
 
 [profile.bench]
 codegen-units = 1
 lto = true
-opt-level = 'z' # Optimize for size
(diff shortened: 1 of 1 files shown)


bartlomieju:
Looks like this change makes CI run twice as long

evanwashere:
benchmark results of [this pr](https://github.com/denoland/deno/runs/1757114372?check_suite_focus=true#step:27:4840)
benchmark results of [f5c28c6](https://github.com/denoland/deno/runs/1757106087#step:27:4836)

ry:
Hi @evanwashere, thanks for the patch but we need more analysis than this - it appears to increase build time. This patch is done to improve performance, what measurements are improving? Previously we found no measurable improvement in our benchmark suite using `-O3`. Happy to reopen and discuss further if you want to get into it.

PR #9254, opened 2021-01-25
fix windows ci

lucacasonato:
I managed to break CI in #9046. This tries to fix that.


lucacasonato:
Closing in favor of #9261.

PR #9255, opened 2021-01-25
fix(docs): small typo in building from source guide

satyarohith:


kt3k (approved):
Thank you for the quick fix! @satyarohith

kt3k:
@satyarohith We reverted this setup of wpt script in #9264 for the moment. Thanks for the fix anyway.

Issue #9149, opened 2021-01-18
readTextFile will only read 16KB of data per tick of event loop 


DavidDeSimone:
There is an outstanding issue in Tokio (see https://github.com/tokio-rs/tokio/issues/1976) that a call to read a AsyncFile into a buffer will only read at most 16KB, no matter how much data is available, or the size of the buffer passed into the function. Based on my initial testing, this can cause inferior performance compared to reading up to the capacity of a buffer using the logic in buffer.js 

https://github.com/denoland/deno/blob/2077864a8deca240fbbd43758e30977cc881e110/runtime/js/13_buffer.js#L159-L181

If you were to log nread per loop iteration for a large file, you will see it is always 16KB. If you were to log nread in the synchronous version of this function for the same large file, you will see that nread grows with as the resize logic increases the buffer size. 

My proposed solution is to modify the read logic in io.rs to explicitly read up to our buffer size, or until we have completed our read. Below is a diff for testing:

```diff
diff --git a/runtime/ops/io.rs b/runtime/ops/io.rs
index 607da74c..75f3a699 100644
--- a/runtime/ops/io.rs
+++ b/runtime/ops/io.rs
@@ -498,6 +498,15 @@ pub fn op_read(
           stream.read(&mut zero_copy).await?
         } else if let Some(stream) = resource.downcast_rc::<StreamResource>() {
-          stream.clone().read(&mut zero_copy).await?
+          let mut total_bytes_read = 0;
+          let len = zero_copy.len();
+          let mut slice = &mut zero_copy[0..len];
+          while total_bytes_read < len {
+            let read = stream.clone().read(&mut slice).await?;
+             if read == 0 { break; }
+            total_bytes_read += read;
+            slice = &mut zero_copy[total_bytes_read..len];
+          }
+          total_bytes_read
         } else {
           return Err(bad_resource_id());
         };
```

Using the following minibenchmark, I am able to see a performance improvement for reading large files multiple times. Output is in milliseconds. 

```bash
user@ubuntu:~/proj/ng-async-files$ deno run --allow-read benchmark.js
Starting...
3046
user@ubuntu:~/proj/ng-async-files$ deno run --allow-read benchmark.js
Starting...
2899
user@ubuntu:~/proj/ng-async-files$ deno run --allow-read benchmark.js
Starting...
2980
user@ubuntu:~/proj/ng-async-files$ ../deno/target/release/deno run --allow-read benchmark.js      
Starting...
2462
[...]

PR #9256, opened 2021-01-25
feat(runtime/core) Read/Write large file I/O performance increase

DavidDeSimone:
I noted in https://github.com/denoland/deno/issues/9149 that due to the way Tokio is programmed, when we perform an Async read via Tokio, we will only read up to 16KB per call to read. Deno's JS is properly written to handle this situation, however it leads to suboptimal performance due to the fact that we will only read 16KB per tick of event loop, even if more data is ready. My initial thoughts were to use a loop to read data at the Rust layer, however that approach was flawed and would introduce changes from previous Deno versions. 

In performing a deep dive of Tokio, I discovered `block_in_place` and wrote an initial MVP which gave me some promising benchmarks. I further developed that into this PR, which gave me the following results. In order to test this change apples to apples, I wrote a new benchmark for parallel file reads. 

--- New ---

``` bash
Benchmark #15: /home/user/proj/deno/target/release/deno run --reload --allow-write --allow-read cli/tests/parallel_read.js 
  Time (mean ± σ):     233.1 ms ±   3.6 ms    [User: 71.5 ms, System: 186.0 ms]
  Range (min … max):   226.2 ms … 238.7 ms    13 runs
 
Benchmark #16: /home/user/proj/deno/target/release/deno run --allow-write --allow-read cli/tests/parallel_read.js 
  Time (mean ± σ):     231.2 ms ±   5.1 ms    [User: 62.6 ms, System: 193.1 ms]
  Range (min … max):   220.8 ms … 237.7 ms    12 runs
```

--- Old --- 
```bash
 
Benchmark #15: /home/user/proj/deno/target/release/deno run --reload --allow-write --allow-read cli/tests/parallel_read.js 
  Time (mean ± σ):     394.6 ms ±   5.8 ms    [User: 101.4 ms, System: 417.8 ms]
  Range (min … max):   388.3 ms … 406.4 ms    10 runs
 
Benchmark #16: /home/user/proj/deno/target/release/deno run --allow-write --allow-read cli/tests/parallel_read.js 
  Time (mean ± σ):     392.2 ms ±   7.9 ms    [User: 101.9 ms, System: 413.0 ms]
  Range (min … max):   375.8 ms … 404.4 ms    10 runs
```

That benchmark is included in the PR for review. I maintain a Deno fork that I will be including these changes in, however I understand if there is hesitation to merge this due to the nature of the changes I am making. 

In order to use block in place, we would need to use the multithreaded runtime. Currently, Deno uses the current thread runtime. I assume this was due to the thought that there wouldn't be benefits to using the multithreaded runtime due to JavaScript's single threaded nature. However, because we use `block_on` for the Runtime, it was easy to switch us over. I limited us to 2 core threads in order to preserve startup time performance. By default, Tokio spawns as many core threads as you have cores. 

While this is a large change, I think its worth it for the file I/O performance increase. The benchmarks on my machine show no significant degradation to other benchmarks, including the networking I/O. 

In doing so, running tests exposed some issues I will enumerate. My assessment is that these were existing issues that could be theoretically possible in the old code, but that the new code just exposed them. I tried very hard to not change the tests to fit my changes, but I had to in a few instances where I did change the tests, because I assessed that the tests did have legitimate race conditions. Almost everything I am going to describe would only be common in the situation where you are a deno client talking to a deno sever via localhost within the same process - something normal in tests, but more rare in real life. 

1. Bug in Fetch
I initially saw errors in the tests for fetch. I was suspicious due to this comment about the tests that were failing:
https://github.com/denoland/deno/blob/ad60e750d75471d7c7f31f9f40ff233a29aba332/cli/tests/unit/fetch_test.ts#L572-L617

The server spawned by `bufferServer` has a chicken and the egg problem. It needs to write in order to trigger the EOF marker being sent, but if we write and then read, we may not always get the data. Previously we were just running them simultaneously with a Promise.all, however this doesn't solve the problem because they will still have an order those operations are executed in, we just don't have control of that order. If bufferServer was an actual HTTP sever that had proper reading logic it wouldn't be an issue, but instead of going that route I reworked the tests to instead read the number of bytes they are expecting

2. The case of cancelling reads
Initially in this patch, certain test operations would throw `Interrupted: operation cancelled` instead of the error the test was expected. This appears to have been an issue with other parts of the codebase (I found code mapping Interrupted to BadResource across several modules). The cases I found in the tests seem to be legitimate places that a cancellation error could happen if the read stream was closed with data pending, and so I modified the tests to handle those cases. 

3. ServeTLS Test

https://github.com/denoland/deno/blob/ad60e750d75471d7c7f31f9f40ff233a29aba332/std/http/server_test.ts#L681-L735

In the case that our second connection/read/write complete before the initial connection, we may close the server prior to the first connection completing. This will cause a "BadConnection" error to be thrown instead of a "InvalidData" exception. I changed the logic slightly to prevent this case while keeping the intention of the initial test.


runtime/ops/io.rs
```diff
@@ -326,8 +326,15 @@ impl StreamResource {
       let mut fs_file = RcRef::map(&self, |r| r.fs_file.as_ref().unwrap())
         .borrow_mut()
         .await;
-      let nwritten = (*fs_file).0.as_mut().unwrap().read(buf).await?;
-      return Ok(nwritten);
+      let f = (*fs_file).0.as_mut().unwrap().try_clone().await?;
+      let mut fstd = f.into_std().await;
+      let read = tokio::task::block_in_place(move || {
```
bartlomieju, line 331: Changing to `block_in_place` means that this function is no longer asynchronous

runtime/ops/io.rs
```diff
@@ -397,9 +404,15 @@ impl StreamResource {
       let mut fs_file = RcRef::map(&self, |r| r.fs_file.as_ref().unwrap())
         .borrow_mut()
         .await;
-      let nwritten = (*fs_file).0.as_mut().unwrap().write(buf).await?;
-      (*fs_file).0.as_mut().unwrap().flush().await?;
-      return Ok(nwritten);
+      let f = (*fs_file).0.as_mut().unwrap().try_clone().await?;
+      let mut fstd = f.into_std().await;
+      let written = tokio::task::block_in_place(move || {
```
bartlomieju, line 409: Ditto

cli/tokio_util.rs
```diff
@@ -1,9 +1,14 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 pub fn create_basic_runtime() -> tokio::runtime::Runtime {
-  tokio::runtime::Builder::new_current_thread()
+  tokio::runtime::Builder::new_multi_thread()
```
bartlomieju, line 4: This is a huge red flag; Deno's `MainWorker` and `WebWorker` are neither `Sync` nor `Send` (meaning they are pinned to the specific thread) and that's be design (can't move V8 isolate to other thread). I'm surprised that it compiles

op_crates/fetch/lib.rs
```diff
@@ -185,7 +187,7 @@ where
     request = request.header(name, v);
   }
 
-  let fut = request.send();
+  let fut = tokio::spawn(async move { request.send().await });
```
bartlomieju, line 190: What's the reason for this change?

bartlomieju (commented):
Thanks for opening the PR @DavidDeSimone; could you explain in more detail what's the reason for this PR? Changing Tokio runtime type used is a fundamental change that is very surprising as using single threaded runtime was a concious decision on our side. 

It looks to me that all the changes are only for optimizations for FS file; if Tokio can't read more than 16Kb in one call it should be fixed on Tokio's side.

Calling @piscisaureus @ry and @bnoordhuis for additional opinions

DavidDeSimone:
> Thanks for opening the PR @DavidDeSimone; could you explain in more detail what's the reason for this PR? Changing Tokio runtime type used is a fundamental change that is very surprising as using single threaded runtime was a concious decision on our side.
> 
> It looks to me that all the changes are only for optimizations for FS file; if Tokio can't read more than 16Kb in one call then IMO it should be fixed on Tokio's side.
> 

First off, thank you for taking the time to review this PR. I fully admit that this is a very low level change to your project, and that if it is merged, it shouldn't be done lightly. If I were in your shoes, I would be highly skeptical of merging something like this without intense vetting, and would likely not merge it on principle. However, I still will enumerate my thought process on these changes and answer your questions, since I think in any case it will foster good discussion. I think this is an awesome project and I look forward to contributing less intense changes in the future.  :)

I am going to write this assuming that the core contributors have a greater understanding of Tokio than I do, so if I write something incorrect or am misunderstanding a fundamental concept, please correct me. 

With all that being said to address point 1:

> It looks to me that all the changes are only for optimizations for FS file; if Tokio can't read more than 16Kb in one call then IMO it should be fixed on Tokio's side.

While my initial issue described this as a bug in Tokio, it really isn't a bug. This is very intention on the part of Tokio, and it has to do with our usage of `block_on`, which is related to your other inquery of "how does changing this even compile if the JsRuntime/MainWorker isn't Send + Sync"? 

Immediately after you spawn the runtime, if you were to spawn a tokio task performing File I/O, and then call block_on for that task, you will notice that is reads the entire buffer requested, like so (I didn't compile this, just writing as a concept)

```rust
let runtime = //...
let task = runtime.spawn(async {
  let file = //...
  let mut vector = vec![0u8; 1024 * 1024 * 8];
  file.read(vector.as_mut_slice()).await;
});

runtime.block_on(task); 
```

Now if you alter this code so that you spawn within block_on, Tokio changes it's behavior so that it will only copy up to 16KB. This is from the class tokio::io::Blocking, which changes the behavior. Link below:

https://github.com/tokio-rs/tokio/blob/f13a9dd87a9026cf009c740ccfc341d479286e2f/tokio/src/io/blocking.rs#L29-L47


This is intentional by the creators of Tokio, and I assume that is because they don't want to block the executor copying large amount of data. That is something they call out explicitly as "not ok" within their documentation. However, block_in_place helps us sidestep this. Quoting [tokio's docs](https://docs.rs/tokio/1.1.0/tokio/task/fn.block_in_place.html): `This function runs the closure on the current thread by having the thread temporarily cease from being a core thread, and turns it into a blocking thread.` Deno already has logic for reading files in chunks, which comments indicate is based on go's implementation. I.e. it starts with a minimum buffer of 32KB, which grows based on number of reads until it hits EOF. To the user, calling readTextFile as an async function still will be happening across multiple cycles of the event loop. 

Now let me be clear, I am not disagreeing with what you said in the comments where I am making that read synchronous. My understanding is that however, is that it will not block the executor and the completion of the pending async tasks. So those tasks are still moving forward. Now if there wasn't a performance increase, I wouldn't think this was worth it. However, with my initial marks it seemed like it could be worth while. To the caller of `Deno.readTextFile`, they likely won't notice a behavior change in that it will still take multiple polls of the event loop once they make their async call for their promise to resolve. My benchmarks showed no performance decrease the case of smaller files, and a reduction in time for files >16KB. 

To address how this compiles, and why it doesn't seemingly explode based on the single threaded nature of JS. If there is a part where I am misunderstanding Tokio, it's in this section. So when we call `runtime.block_on`, we remain on the same thread we were on prior to the call. My motivation for this PR was my own application which uses a fork of Deno. In that application I create my worker prior to calling block_on, and move my worker into the future, which compiles and runs just fine. Now, I've noticed that in the codebase, we never directly call `tokio::spawn` (besides the call I added to my PR, and I think within a test file not directly embedded within the application). If you were to do that and try to move MainWorker into your spawned future, you will see that the compiler does explode, because MainWorker isn't Send + Sync. This makes sense. However, just making the runtime multithreaded doesn't violate Send + Sync for the runtime, because we are still operating JS on the same thread. 

I believe this writeup should have addressed (or at least attempted to address) all of your points but one, which was my added call to tokio::spawn. I will address that directly in the PR review. I am standing by for any additional feedback or concerns you may have, thanks again.

op_crates/fetch/lib.rs
DavidDeSimone: My findings that that there was a subtle race condition here I was able to surface from this change. This is a race condition that I think is not actually present in the single threaded runtime that could only be introduced in the multithreaded one, but I may be incorrect and this existing in the current runtime. When calling send, we kick off an async operation, however we don't await on it until we call the later function, which I believe is opFetchSend. If the connection resolves between opFetch and opFetchSend, the runtime may not be read to receive the connection and it causes and error. To prevent this, I push the task onto a worker where it will be await'd until we call opFetchSend.

kitsonk:
> EDIT: In addition, it looks like CI failed due to the windows python build issue introduced last night. I do not know how to re-run the CI.

It would need to be rebased or master merged in and pushed.  Just re-running CI won't work in this case.

cli/tokio_util.rs
bnoordhuis: I was initially as nonplussed as you are but the crux is that we call `block_on()`, that still runs the future on the current thread.

Tasks spawned with `tokio::spawn()` now need to be `Send` (and presumably `Sync`) but deno uses them very sparingly. I don't see anything wrong with this change...

...except that the functions should now be called `create_multi_runtime()` and `run_multi()`, of course. :-)

Aside: there's a copy of this file in `runtime/tokio_util.rs`.

op_crates/fetch/lib.rs
bnoordhuis: I think this changes the behavior from "suspended until polled in `op_fetch_send()`" to "runs concurrently". I don't know this code well enough to say if that's good, bad, or neutral, but @bartlomieju or @lucacasonato probably do.

Aside: this can be simplified to `let fut = tokio::spawn(request.send());`

bnoordhuis (commented):
- I agree with @DavidDeSimone that the 16 kb block size in Tokio is intentional and unlikely to change upstream

- using `block_in_place` is problematic - think files on a network file systems

- but as a counterargument: the Erlang runtime has a flag for turning off async file I/O because most file I/O is local and "immediate", much faster than dipping in the thread pool everytime

DavidDeSimone:
> * I agree with @DavidDeSimone that the 16 kb block size in Tokio is intentional and unlikely to change upstream
> * using `block_in_place` is problematic - think files on a network file systems
> * but as a counterargument: the Erlang runtime has a flag for turning off async file I/O because most file I/O is local and "immediate", much faster than dipping in the thread pool everytime

This is a good counter point, and counter-counter point. After reflecting on this more, it seems like there would be more benefit to this PR if I reworked it to either be a flag on a call to read, or to be a separate op all together. I personally prefer it being a flag passed to read. I think the benefits would be:

1. It gives power users a way to increase their read performance so long as their use case fits the caveats which can be made clear through documentation.
2. Making a flag that defaults to false means that there won't be any change to the existing users call to read - they would behave the same way, only reading that 16KB buffer per call.

DavidDeSimone:
I've reworked this PR based on the previous discussion. Now, by default, async reads and writes will use the current default Deno async read/write behavior. I've added two new flags for Deno.open, `readMayBlock` and `writeMayBlock`. Setting those flags to true will use `block_in_place`. 

I've removed the changes to fetch - I was mistaken in thinking that was a low level race condition, and that change was unneeded, and also a behavioral change. Now the only change should be a. opting into the multithreaded runtime and b. adding a codepath for block_in_place. 

I also tried to be a "boy scout" and reduce the repetition between `tokio_util.rs` in cli/ and `tokio_util.rs` in runtime/ since they were identical files. 

Requesting a re-review @bnoordhuis @bartlomieju

DavidDeSimone:
I think that this PR lead to good discussion, however after additional talks with some contributors, I think that this is just too large of a change for what it provides. I am going to close this PR with a thank you to the team for the reviews.

PR #9259, opened 2021-01-25
fix(lsp): reduce deadlocks with in memory documents

kitsonk:
I was still running into a few hangs while using the lsp.  Looking back at rust_analyzer, the in memory docs during the snapshot are simply cloned and not modifiable.  We were already in a situation where our in memory documents wouldn't be written with a snapshot, but we were still locking, and those locks I believe were causing some deadlocks when things like formatting and TypeScript diagnostics were happening in different threads at the same time.


bnoordhuis (approved):
LGTM, I think this can only be an improvement, but I see that `StateSnapshot` still has multiple mutexes so there's still room for deadlocks if two threads take out the locks in different order.

PR #9261, opened 2021-01-25
fix(ci): fix windows ci

kt3k:
An alternative to https://github.com/denoland/deno/pull/9254

This PR tries to avoid windows CI error by passing `--py2` option to wpt util script. ref: https://github.com/web-platform-tests/wpt/blob/61d85c8/wpt#L10

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -5297,59 +5297,24 @@ fn web_platform_tests() {
   let config: std::collections::HashMap<String, Vec<WptConfig>> =
     deno_core::serde_json::from_value(jsonc_to_serde(jsonc)).unwrap();
 
-  // Observation: `python3 wpt serve` hangs with the python3 from homebrew
-  // but works okay with /usr/bin/python, which is python 2.7.10. Observed
-  // with homebrew python 3.8.5, 3.8.7 and 3.9.1.
   let python = match true {
     _ if cfg!(target_os = "windows") => "python.exe",
-    _ if cfg!(target_os = "macos") => "python",
     _ => "python3",
   };
 
   eprintln!("If the wpt server fails or gets stuck, please set up your /etc/hosts file like specified in //docs/contributing/building_from_source.md.");
+  eprintln!("On windows python.exe needs to available in your path, and needs to point to python 3. On macOS and Linux, python3 needs to be available in your path.");
 
   let mut proc = Command::new(python)
     .current_dir(util::wpt_path())
     .arg("wpt.py")
-    .arg("--py2")
     .arg("serve")
     .stderr(std::process::Stdio::piped())
     .spawn()
     .unwrap();
 
-  let stderr = proc.stderr.as_mut().unwrap();
-  let mut stderr = BufReader::new(stderr).lines();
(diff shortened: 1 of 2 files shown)


cli/tests/integration_tests.rs
```diff
@@ -5316,39 +5313,8 @@ fn web_platform_tests() {
...
-    // WPT + python2 doesn't support HTTP/2.0.
-    if line.contains("Cannot start HTTP/2.0 server") {
-      ready_9000 = true;
-    }
-    if ready_8000 && ready_8443 && ready_8444 && ready_9000 {
-      break;
-    }
-  }
+  // Give the wpt runner time to start up.
+  std::thread::sleep(std::time::Duration::from_secs(5));
```
ry, line 5317: :)

cli/tests/integration_tests.rs
```diff
@@ -5316,39 +5313,8 @@ fn web_platform_tests() {
     .spawn()
     .unwrap();
 
-  let stderr = proc.stderr.as_mut().unwrap();
-  let mut stderr = BufReader::new(stderr).lines();
```
kt3k, line 5320: We should remove use declaration of this as well?

cli/tests/integration_tests.rs
```diff
@@ -5297,16 +5297,13 @@ fn web_platform_tests() {
...
-  // but works okay with /usr/bin/python, which is python 2.7.10. Observed
-  // with homebrew python 3.8.5, 3.8.7 and 3.9.1.
   let python = match true {
     _ if cfg!(target_os = "windows") => "python.exe",
-    _ if cfg!(target_os = "macos") => "python",
     _ => "python3",
   };
 
   eprintln!("If the wpt server fails or gets stuck, please set up your /etc/hosts file like specified in //docs/contributing/building_from_source.md.");
+  eprintln!("On windows python.exe needs to available in your path, and needs to point to python 3. On macOS and Linux, python3 needs to be available in your path.");
```
ry, line 5306: Maybe we should check this here in Rust? Like start `python.exe --version` and see that it has the correct output?

lucacasonato: Would it be ok if I do this when I move the WPT runner to TS?

ry: Not convinced about moving WPT runner to TS, but sure

bartlomieju:
There are some clippy warnings popping up:
```
warning: unused import: `std::io::BufReader`
 --> cli/tests/integration_tests.rs:8:5
  |
8 | use std::io::BufReader;
  |     ^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: variable does not need to be mutable
    --> cli/tests/integration_tests.rs:5308:7
     |
5308 |   let mut proc = Command::new(python)
     |       ----^^^^
     |       |
     |       help: remove this `mut`
     |
     = note: `#[warn(unused_mut)]` on by default

warning: 2 warnings emitted

```

lucacasonato:
I have reverted #9046 because I am going to do a full rewrite of the WPT runner. Because of this, this does not apply anymore.

PR #9271, opened 2021-01-25
fix(lsp): fix deadlocks, use one big mutex

bnoordhuis:
The LSP code had numerous places where competing threads could take out
out locks in different orders, making it very prone to deadlocks.
This commit sidesteps the entire issue by switching to a single lock.

The above is a little white lie: the Sources struct still uses a mutex
internally to avoid having to boil the ocean (because being honest about
what it does involves changing all its methods to `&mut self` but that
ripples out extensively...) I'll save that one for another day.

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -9,7 +9,6 @@ use super::tsc;
 use crate::diagnostics;
 use crate::media_type::MediaType;
 
-use deno_core::error::custom_error;
 use deno_core::error::AnyError;
 use deno_core::serde_json;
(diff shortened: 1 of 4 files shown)


kitsonk:
Very cool, thank you.  I did a quick glance and it looks good to me, but I am going to take this branch for a spin and see if I can break it.  I suspect I won't be able to.

kitsonk (approved):
LGTM,  I tried fairly hard to break it.  A debug build feels a bit "laggy" but it can be that way with a debug build without this, while a release build was as snappy as can be.  At the very least this gives us a place to iterate from and it is a lot cleaner.

PR #9273, opened 2021-01-26
fix(lsp): handle mbc properly when formatting

kitsonk:


lucacasonato (approved):
Looks good, please rebase.

PR #9276, opened 2021-01-26
refactor(op_crates/web): Move URL parsing to Rust

### LINKED ISSUE #9383: Bugs in URL implementation
```js
// Setting `.port` should update only the receiver.
const u1 = new URL("http://google.com/");
const u2 = new URL(u1);
u2.port = 123;
assertStrictEquals(u1.port, "");

// URLSearchParams identity should not be lost when URL is updated.
const u = new URL("http://foo.com/");
const sp1 = u.searchParams;
u.href = "http://bar.com/?baz=42";
const sp2 = u.searchParams;
assertStrictEquals(sp1, sp2);
```

nayeemrmn:
cc @lucacasonato 

Demo of replacing as much as possible of `URL` and `URLSearchParams` with rust-url. The biggest compromise is having to reparse URLs when any individual part setter is invoked, since **\*parsed\*** URLs are neither resources nor serialisable. There is also the op interface overhead. Note that we already op out to Rust for host parsing in `new URL()` and host setting. Benchmarks are potentially needed. If improvement is needed, some setters can be reverted to parse parts in isolation like currently. Ideally this wouldn't bring back too much duplicated encoding logic.

The changes in the tests are corrective other than some drive letter tests commented with a FIXME. We should be able to enable more wpt URL tests than we could now but rust-url is also not perfectly to spec. However, we use it for URL handling in Rust already and the issues can be upstreamed.

Fixes #9383.


lucacasonato:
I want to hold this until the URL wpt are relanded.

op_crates/web/lib.rs
```diff
@@ -75,6 +53,123 @@ pub fn init(isolate: &mut JsRuntime) {
...
+    hostname: String,
+    password: String,
+    pathname: String,
+    port: String,
+    protocol: String,
+    search: String,
+    username: String,
+  }
+  // TODO(nayeemrmn): Panic that occurs in rust-url for the `non-spec:`
+  // url-constructor wpt tests: <INSERT ISSUE>.
```
nayeemrmn, line 132: Need to open an issue once I've confirmed the scope.

nayeemrmn:
Ready for review.

op_crates/web/lib.rs
```diff
@@ -75,6 +53,127 @@ pub fn init(isolate: &mut JsRuntime) {
...
+    port: String,
+    protocol: String,
+    search: String,
+    username: String,
+  }
+  // TODO(nayeemrmn): Panic that occurs in rust-url for the `non-spec:`
+  // url-constructor wpt tests: https://github.com/servo/rust-url/issues/670.
+  let username =
+    std::panic::catch_unwind(|| quirks::username(&url).to_string())
+      .map_err(|_| custom_error("Error", "Internal error."))?;
```
kt3k, line 137: Maybe web should print the given url string?

kt3k:
I'm in favor of this PR because:

1. This is more compliant to WPT than our current implementation (ref: expectation.json)
2. This reduces the maintenance cost of our code base significantly.
3. This should slightly reduce the V8 snapshot size and the entire runtime size.

op_crates/web/11_url.js
```diff
@@ -234,292 +174,51 @@
     }
 
     *keys() {
-      for (const [key] of this.#params) {
-        yield key;
+      for (let i = 0; i < this.#params.length; i++) {
```
lucacasonato, line 177: Why this change?

nayeemrmn: I think this was a red herring for the iterator tests... reverted.

piscisaureus:
I'm also in favor of landing this PR, mostly because it ensures that URL parsing on the Rust side is consistent with URL behavior on the JavaScript side.

However I find it disappointing that it is not faster than our current implementation. I tried the following microbenchmark:

```js
const t1 = Date.now();
let i = 0;
for(; i < 1e6; i++) {
  new URL("http://www.google.com");
}
const t2 = Date.now();
const dt = (t2 - t1) / 1e3;
const r = i / dt;
console.log(`n = ${i}, dt = ${dt.toFixed(3)}s, r = ${r.toFixed(0)}/s`);
```

... with the following (approximate) results on my laptop:
* deno: 30000/s
* deno w/ this patch: 30000/s
* chrome: 800000/s
* node.js: 800000/s

nayeemrmn:
Improved by roughly 50% by updating `url.searchParams` on demand.

piscisaureus (approved):
LGTM, thank you for your patience @nayeemrmn.
I get very excited when a patch removes 700 lines of code :)

PR #9278, opened 2021-01-26
fix(coverage): use source maps when printing pretty reports

### LINKED ISSUE #8923: Report source mapped code coverages
For the initial pass I did straight JavaScript code coverage as the runtime saw it for simplicity of implementation. This was never intended to be permanent and we should match the user's expectation and render the mapped source in reports.

caspervonb:
This makes use of source maps and the original source when printing lacking line coverage in the pretty printer.

Only the executable lines are checked as before (as non-executable lines will always be ignored anyways).
The lines then mapped to the appropriate source line when a source map is present.

Closes #8923

### DIFF
--- a/cli/tests/test_run_test_coverage.out
+++ b/cli/tests/test_run_test_coverage.out
@@ -1,6 +1,6 @@
-Check [WILDCARD]/$deno$test.ts
+Check [WILDCARD]/tests/$deno$test.ts
 running 1 tests
-test spawn test ... Check [WILDCARD]/$deno$test.ts
+test spawn test ... Check [WILDCARD]/tests/$deno$test.ts
 running 1 tests
 test returnsFooSuccess ... ok ([WILDCARD])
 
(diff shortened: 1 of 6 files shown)


cli/tools/coverage.rs
```diff
@@ -221,8 +231,34 @@ impl PrettyCoverageReporter {
...
+        } else {
+          lines
+        };
+
+      let output_indices = if let Some(source_map) = maybe_source_map {
+        let mut indices = uncovered_lines
+          .iter()
+          .map(|i| {
+            source_map
+              .lookup_token(*i as u32, 16)
```
caspervonb, line 246: Magical offset; lookup_token searches backwards so it'll find something NEAR the line; not the actual line.

cli/tools/coverage.rs
```diff
@@ -221,8 +231,34 @@ impl PrettyCoverageReporter {
         println!("{}", colors::red(&line_coverage));
       }
 
+      let output_lines =
+        if let Some(original_source) = maybe_original_source.as_ref() {
+          original_source.split('\n').collect::<Vec<_>>()
```
ry, line 236: Does this handle CRLF files correctly?

cli/tests/test_run_test_coverage.out
```diff
@@ -10,23 +10,24 @@ ok ([WILDCARD])
...
-cover [WILDCARD]/subdir/subdir2/mod2.ts ... 62.500% (5/8)
-   5 | export function printHello2() {
-   6 |     printHello();
-   7 | }
-cover [WILDCARD]/test_coverage.ts ... 100.000% (5/5)
+cover [WILDCARD]/tests/subdir/subdir2/mod2.ts ... 62.500% (5/8)
+   7 | export function printHello2(): void {
+   8 |   printHello();
+   9 | }
+cover [WILDCARD]/tests/test_coverage.ts ... 100.000% (5/5)
```
ry, line 33: Nice!

cli/tools/coverage.rs
caspervonb: Right, windows.

I believe so, we also do the same in program_state (https://github.com/denoland/deno/blob/master/cli/program_state.rs#L324).

cli/tools/coverage.rs
caspervonb, line 246: 1:1 mapping also isn't going to work; should get all the source lines that the executed line represents (happens to be 1:1 in most of our generated code but brittle).

bartlomieju (approved):
LGTM, great to have this fixed!

PR #9279, opened 2021-01-26
fix: fix panics in Deno.emit

### LINKED ISSUE #9277: Question about rootSpecifier of Deno.emit
I tested some patterns of custom scheme urls as input of Deno.emit, and I found the following 2 cause panic:

```ts
let id = 'custom://a.ts';
await Deno.emit(id, { sources: { [id]: "" } }); // panics with UnsupportedImportType
```
```ts
let id = 'custom://a//a.ts';
await Deno.emit(id, { sources: { [id]: "" } }); // => panics with AssertionError
```

(These types of urls are used in dext. ref: https://github.com/lucacasonato/dext.ts/blob/8483d19/src/plugins/dext.ts#L46 )

On the other hand, the following compiles without problem:
```ts
let id = 'custom://a/a.ts';
await Deno.emit(id, { sources: { [id]: "" } }); // => compiles
```

What do you think are the expected behaviors for the above 2 cases?

kt3k:
Deno.emit panics with the specifier `custom://a.ts` and `custom://a//a.ts` as input. This PR maps these 2 panics to JavaScript errors.

fixes #9277

### DIFF
--- a/cli/ops/runtime_compiler.rs
+++ b/cli/ops/runtime_compiler.rs
@@ -94,19 +94,24 @@ async fn op_emit(
   let mut builder = GraphBuilder::new(handler, maybe_import_map, None);
   let root_specifier =
     ModuleSpecifier::resolve_url_or_path(&args.root_specifier)?;
-  builder.add(&root_specifier, is_dynamic).await?;
+  builder
+    .add(&root_specifier, is_dynamic)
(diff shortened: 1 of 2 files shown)


kt3k:
Closing because we probably need to different thing to these 2 cases.

Issue #9267, opened 2021-01-25
deno fmt --help: wrong description


starkpsmichael:
Running `deno fmt --help`says: Auto-format JavaScript/TypeScript source code.

In #8887 markdown support was added.
So the describtion should be also saying something like:
Auto-format JavaScript, TypeScript and Markdown files.

AkifumiSato:
I'd like to work on this issue.

PR #9280, opened 2021-01-26
fix(cli): Added markdown description to `fmt --help`

AkifumiSato:
Closes: #9267 

> Running deno fmt --helpsays: Auto-format JavaScript/TypeScript source code.
In #8887 markdown support was added.
So the describtion should be also saying something like:
Auto-format JavaScript, TypeScript and Markdown files.


lucacasonato (approved):
Thanks

PR #9287, opened 2021-01-27
fix(cli): correctly determine emit state with redirects

### LINKED ISSUE #9129: Deno cache breaks on redirected libraries.
Today I wanted to implement this library:
`https://deno.land/x/hmac@v1.0.2`
But the problem is that, deno tries recompile whole project every time.
After day of test I found that this library imports own deps.ts with own third party libraries for example:
`https://denopkg.com/chiefbiiko/sha1@v1.0.3/mod.ts`
If we go under this address, we are redirected to:
`https://raw.githubusercontent.com/chiefbiiko/sha1/v1.0.3/mod.ts`
In $HOME/.cache/deps directory I see denopkg.com and raw.githubusercontent.com directories, so I don't know if this is a deno or denopkg problem.

Steps to reproduce:
1. Create test.ts file
2. Inside put:
```
import * as test from 'https://denopkg.com/chiefbiiko/sha1@v1.0.3/mod.ts'
```
3. `deno run test.ts`

kitsonk:
Fixes #9129

### DIFF
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -1470,7 +1470,7 @@ impl Graph {
           true
         }
       } else {
-        false
+        true
       }
     })
   }
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -956,6 +956,37 @@ fn ts_dependency_recompilation() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


bartlomieju (approved):
LGTM

PR #9288, opened 2021-01-27
fix(cli): add lib dom.asynciterable

### LINKED ISSUE #9218: Add a lib to cover async iterables in the DOM
Before 1.7.0, i had an encountered another error, with the hopes 1.7 would fix it... and it did, but just another error has popped up now

```
// a.ts
import * as whatever from "https://deno.land/std@0.84.0/async/pool.ts"
```
```
// tsconfig.json
{
  "compilerOptions": {
    "lib": [
      "deno.ns",
      "dom",
      "dom.iterable",
      "ES2018"
    ]
  }
}
```
```
$ deno run -A --config tsconfig.json a.ts

error: TS7053 [ERROR]: Element implicitly has an 'any' type because expression of type 'symbol' can't be used to index type 'ReadableStream<R>'.
  return res.readable[Symbol.asyncIterator]();
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.84.0/async/pool.ts:45:10
```

kitsonk:
Fixes #9218

Introduces a new lib `dom.asynciterable` to complement `dom` and `dom.iterable` until TypeScript fixes the issue upstream.

If you are using a `"lib"` option in a `tsconfig.json` or with `Deno.emit()` you would want to make it look something like this:

```json
{
  "compilerOptions": {
    "lib": [ "esnext", "dom", "dom.iterable", "dom.asynciterable" ]
  }
}
```

### DIFF
--- a/cli/dts/lib.dom.asynciterable.d.ts
+++ b/cli/dts/lib.dom.asynciterable.d.ts
@@ -1,0 +1,9 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/// <reference no-default-lib="true"/>
+
+interface ReadableStream<R = any> {
+  [Symbol.asyncIterator](options?: {
+    preventCancel?: boolean;
+  }): AsyncIterableIterator<R>;
+}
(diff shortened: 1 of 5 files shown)


bartlomieju (approved):
LGTM

PR #9290, opened 2021-01-27
fix(op_crates/web): Add customInspect for Location

nayeemrmn:


op_crates/web/12_location.js
```diff
@@ -164,6 +164,27 @@
...
+              "port",
+              "protocol",
+              "search",
+            ];
+            const objectString = keys
+              .map((key) => `${key}: "${this[key] || ""}"`)
+              .join(", ");
+            // TODO(nayeemrmn): It would be good if `Deno.inspect()` were
+            // available here, so we had automatic wrapping, indents etc.
+            return `Location { ancestorOrigins: { length: 0 }, ${objectString} }`;
```
nayeemrmn, line 185: To be clear `Deno.customInspect()` _is_ available as a global, but it wouldn't be correct to use it. Unless we check for it and progressively enhance, maybe?

ry (approved):
LGTM - thanks @nayeemrmn !

PR #9291, opened 2021-01-27
feat(cli): Deno.emit supports bundling as IIFE

### LINKED ISSUE #9204: Deno.emit() should support IIFE bundles
swc added IIFE bundles in https://github.com/swc-project/swc/pull/1189.  This issue is to track their integration into `Deno.emit()`.

kitsonk:
Closes #9204

Thanks @kdy1 for adding it to swc!  Finally got around to adding it to Deno.

### DIFF
--- a/cli/tests/compiler_api_test.ts
+++ b/cli/tests/compiler_api_test.ts
@@ -334,6 +334,7 @@ Deno.test({
     assert(diagnostics);
     assertEquals(diagnostics.length, 0);
     assertEquals(Object.keys(files).length, 1);
-    assert(files["deno:///bundle.js"].startsWith("function() {\n"));
+    assert(files["deno:///bundle.js"].startsWith("(function() {\n"));
+    assert(files["deno:///bundle.js"].endsWith("})();\n"));
   },
 });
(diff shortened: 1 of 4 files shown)


cli/tests/compiler_api_test.ts
```diff
@@ -292,3 +292,21 @@ Deno.test({
...
+      sources: {
+        "/a.ts": `import { b } from "./b.ts";
+          console.log(b);`,
+        "/b.ts": `export const b = "b";`,
+      },
+    });
+    assert(diagnostics);
+    assertEquals(diagnostics.length, 0);
+    assertEquals(Object.keys(files).length, 1);
+    assert(files["deno:///bundle.js"].startsWith("function() {\n"));
```
bartlomieju, line 310: Should it start with `(function() {\n`?

bartlomieju (approved):
LGTM 🚀

cli/tests/compiler_api_test.ts
kdy1: I think I made a mistake.
I'll fix it and publish a new version. (fixer should be called after wrapping with iife, but it's not)

kdy1: I published a new version of `swc_bundler`.

kitsonk:
Since this is a new feature (albeit for an unstable option) I don't think we should land this until 1.8.  It should be relatively easy to rebase, so we can update the bundler from swc for it.

PR #9295, opened 2021-01-27
feat: add --ext flag to deno eval

satyarohith:
This PR deprecates the `--ts/-T` (which will later be removed in 2.0) flag of `deno eval` and introduces `--ext` which is used by `deno fmt` for content type selection. This is to ensure we have a single flag that can be used across subcommands to select the language (JS/TS).

cc @bartlomieju


cli/flags.rs
```diff
@@ -521,7 +521,14 @@ fn eval_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   flags.allow_write = Some(vec![]);
   flags.allow_plugin = true;
   flags.allow_hrtime = true;
+  // TODO(@satyarohith): stop considering this flag in 2.0.
```
bartlomieju, line 524: `s/stop considering/remove`

cli/flags.rs
```diff
@@ -995,17 +998,27 @@ fn eval_subcommand<'a, 'b>() -> App<'a, 'b> {
...
+        .multiple(false)
+        .hidden(true),
+    )
+    .arg(
+      Arg::with_name("ext")
+        .long("ext")
+        .help("Set standard input (stdin) content type")
+        .takes_value(true)
+        .default_value("js")
+        .possible_values(&["ts", "js"]),
```
bartlomieju, line 1021: Shouldn't `jsx` and `tsx` be allowed as well? 🤔

satyarohith: The current implementation doesn't have support for `jsx`/`tsx` so I didn't include them in the possible values. 

But it's very easy (just return the appropriate media type and I already have the implemented version locally) to add the support. I'm +1 for supporting them. Let me your thoughts.

satyarohith: I added support for them in the new changes

bartlomieju (approved):
LGTM, thanks Satya!

PR #9302, opened 2021-01-28
fix(cli): fix panic of Deno.emit

kt3k:
This PR fixes the panic of Deno.emit when called with an module specifier with unknown media type, and improves the error message.

~~partially addresses #9277~~

This doesn't address the original issue, but I think this is still a fix for a different problem.


kitsonk (approved):
LGTM

PR #9306, opened 2021-01-28
chore: reenable watch tests

bartlomieju:
After switching to slower CI runners it seems we can reenable file
watcher integration tests and avoid previous flakiness.


ry (approved):
LGTM

PR #9305, opened 2021-01-28
chore: add jsdoc to 26_fetch.js and enable some fetch tests

lucacasonato:
This adds some JS doc to 26_fetch.js, and fixes some issues.

PR #9309, opened 2021-01-28
refactor: rewrite Blob implementation

lucacasonato:
~Depends on #9305.~ Merged.

This PR contains a rewritten `Blob` implementation that follows the spec at https://w3c.github.io/FileAPI/#blob-section. This implementation passes all Blob related web platform tests.

This is the first of multiple PRs to rewrite the `fetch` implementation and its dependencies to match the specifications and pass WPT.

### DIFF
--- a/op_crates/fetch/21_blob.js
+++ b/op_crates/fetch/21_blob.js
@@ -206,7 +206,7 @@
       const O = this;
       /** @type {number} */
       let relativeStart;
-      if (start == undefined) {
+      if (start === undefined) {
         relativeStart = 0;
       } else {
         start = Number(start);
@@ -218,7 +218,7 @@
       }
       /** @type {number} */
       let relativeEnd;
-      if (end == undefined) {
+      if (end === undefined) {
         relativeEnd = O.size;
       } else {
         end = Number(end);
--- a/op_crates/fetch/21_blob.js
+++ b/op_crates/fetch/21_blob.js
@@ -1,0 +1,289 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+// @ts-check
+/// <reference no-default-lib="true" />
+/// <reference path="../../core/lib.deno_core.d.ts" />
+/// <reference path="../web/internal.d.ts" />
+/// <reference path="../web/lib.deno_web.d.ts" />
+/// <reference path="./internal.d.ts" />
+/// <reference path="./lib.deno_fetch.d.ts" />
+/// <reference lib="esnext" />
+
+((window) => {
+  // TODO(lucacasonato): fix this
+  const isWindows = false;
+
+  /**
+   * @param {string} input
+   * @param {number} position
+   * @returns {{result: string, position: number}}
+   */
+  function collectCodepointsNotCRLF(input, position) {
+    const start = position;
+    for (
+      let c = input.charAt(position);
+      position < input.length && !(c == "\r" || c == "\n");
+      c = input.charAt(++position)
+    );
+    return { result: input.slice(start, position), position };
+  }
+
+  /**
+   * @param {string} s
+   * @returns {string}
+   */
+  function convertLineEndingsToNative(s) {
+    const nativeLineEnding = isWindows ? "\r\n" : "\n";
+
+    let { result, position } = collectCodepointsNotCRLF(s, 0);
+
+    while (position < s.length) {
+      const codePoint = s.charAt(position);
+      if (codePoint == "\r") {
+        result += nativeLineEnding;
+        position++;
+        if (position < s.length && s.charAt(position) == "\n") {
+          position++;
+        }
+      } else if (codePoint == "\n") {
+        position++;
+        result += nativeLineEnding;
+      }
+      const { result: token, position: newPosition } = collectCodepointsNotCRLF(
+        s,
+        position,
+      );
+      position = newPosition;
+      result += token;
+    }
+
+    return result;
+  }
+
+  /**
+   * @param  {...Uint8Array} bytesArrays
+   * @returns {Uint8Array} 
+   */
+  function concatUint8Arrays(...bytesArrays) {
+    let byteLength = 0;
+    for (const bytes of bytesArrays) {
+      byteLength += bytes.byteLength;
+    }
+    const finalBytes = new Uint8Array(byteLength);
+    let current = 0;
+    for (const bytes of bytesArrays) {
+      finalBytes.set(bytes, current);
+      current += bytes.byteLength;
+    }
+    return finalBytes;
+  }
+
+  const utf8Encoder = new TextEncoder();
+  const utf8Decoder = new TextDecoder("utf8");
+
+  /** @typedef {BufferSource | Blob | string} BlobPart */
+
+  /** 
+     * @param {(BlobPart)[]} parts
+     * @param {string} endings
+     * @returns {Uint8Array}
+     */
+  function processBlobParts(parts, endings) {
+    /** @type {Uint8Array[]} */
+    const bytesArrays = [];
+    for (const element of parts) {
+      if (element instanceof ArrayBuffer) {
+        bytesArrays.push(new Uint8Array(element.slice(0)));
+      } else if (ArrayBuffer.isView(element)) {
+        const buffer = element.buffer.slice(
+          element.byteOffset,
+          element.byteOffset + element.byteLength,
+        );
+        bytesArrays.push(new Uint8Array(buffer));
+      } else if (element instanceof Blob) {
+        bytesArrays.push(
+          new Uint8Array(element[_byteSequence].buffer.slice(0)),
+        );
+      } else if (typeof element == "string") {
+        let s = element;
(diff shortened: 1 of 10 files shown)


op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+
+      if (!options || typeof options == "function") {
+        options = {};
+      }
+      if (typeof options != "object") {
+        throw new TypeError(
+          `Failed to construct 'Blob'. options is not an object.`,
+        );
+      }
+      const endings = options?.endings?.toString() ?? "transparent";
```
kitsonk, line 180: hmmm... should be an object at this point no matter what, so I am surprised this doesn't work:

```suggestion
      const endings = options.endings?.toString() ?? "transparent";
```

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+      const type = options?.type?.toString() ?? "";
+
+      /** @type {Uint8Array} */
+      const bytes = processBlobParts(parts, endings);
+      this[_byteSequence] = bytes;
+      this.#type = normalizeType(type);
```
kitsonk, line 186: I personally have avoided private members in spec following code, even if they aren't accessed out of their class, for consistency sake.  If the spec calls for it to be an internal slot, maybe just make it a symbol like `_byteSequence`?

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+   */
+  function normalizeType(str) {
+    let normalizedType = str;
+    if (!/^[\x20-\x7E]*$/.test(str)) {
+      normalizedType = "";
+    }
+    return normalizedType.toLowerCase();
+  }
+
+  const _byteSequence = Symbol("[[ByteSequence]]");
```
kitsonk, line 132: I've been declaring the internal slots as a member of the class (with no init) so they can have a JSDoc type.

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+     * @param {number} [start]
+     * @param {number} [end]
+     * @param {string} [contentType]
+     * @returns {Blob}
+     */
+    slice(start, end, contentType) {
+      const O = this;
+      /** @type {number} */
+      let relativeStart;
+      if (start == undefined) {
```
kitsonk, line 209:
```suggestion
      if (start === undefined) {
```

Unless you are trying to trap `null` as well, but in that case, more idomatic Deno would be:

```suggestion
      if (start == null) {
```

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+        start = Number(start);
+        if (start < 0) {
+          relativeStart = Math.max(O.size + start, 0);
+        } else {
+          relativeStart = Math.min(start, O.size);
+        }
+      }
+      /** @type {number} */
+      let relativeEnd;
+      if (end == undefined) {
```
kitsonk, line 221:
```suggestion
      if (end === undefined) {
```

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+        end = Number(end);
+        if (end < 0) {
+          relativeEnd = Math.max(O.size + end, 0);
+        } else {
+          relativeEnd = Math.min(end, O.size);
+        }
+      }
+      /** @type {string} */
+      let relativeContentType;
+      if (contentType === undefined) {
```
kitsonk, line 233: Now we are back to strong equality! 🤣

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /**
+     * @returns {ReadableStream<Uint8Array>}
+     */
+    stream() {
+      const bytes = this[_byteSequence];
+      const stream = new ReadableStream({
+        type: "bytes",
+        /** @param {ReadableStreamDefaultController<Uint8Array>} controller */
```
kitsonk, line 250:
```suggestion
        /** @param {ReadableByteStreamController} controller */
```

kitsonk (commented):
mostly nitpicks

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+    let current = 0;
+    for (const bytes of bytesArrays) {
+      finalBytes.set(bytes, current);
+      current += bytes.byteLength;
+    }
+    return finalBytes;
+  }
+
+  const utf8Encoder = new TextEncoder();
+  const utf8Decoder = new TextDecoder("utf8");
```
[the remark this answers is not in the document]
lucacasonato: No reason other than to be explicit. I will remove.

op_crates/fetch/21_blob.js
lucacasonato: I tried that, but then TSC screams at me for redeclaring a property for some reason.

op_crates/fetch/21_blob.js
lucacasonato: Yeah this case it actually needs to be strong equality. The other two need to be weak (null or undefined). I will check with the WebIDL spec again about this.

op_crates/fetch/21_blob.js
```diff
@@ -0,0 +1,289 @@
...
+      }
+
+      const parts = [];
+      const iterator = blobParts[Symbol.iterator]?.();
+      if (iterator === undefined) {
+        throw new TypeError(
+          "Failed to construct 'Blob'. The provided value cannot be converted to a sequence",
+        );
+      }
+      while (true) {
```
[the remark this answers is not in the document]
lucacasonato: Because we need to manually retrieve the `@@iterator` to check that the value you are passing is actually a valid iterator, but then we can not use `for of` to iterate over the iterable, because that would call `@@iterator` again. There is a WPT that checks that we only call `@@iterator` once. Because of this we have to iterate manually with the iterator we get from the first call to `@@iterator`.

op_crates/fetch/21_blob.js
lucacasonato: Here the spec doesn't actually call for it to be an internal slot. We only make it a private type because it is a readonly property (it is read back to the user via `get type`).

op_crates/fetch/21_blob.js
kitsonk: 😕 hmmmm... didn't get that with streams

lucacasonato: Yeah I was suprised too. I can get you an exact error in a few minutes.

op_crates/fetch/21_blob.js
lucacasonato: Just checked, they can all be strong equal `undefined`.

op_crates/fetch/21_blob.js
lucacasonato:
```
Subsequent property declarations must have the same type.  Property '[_byteSequence]' must be of type 'any', but here has type 'Uint8Array'.ts(2717)

21_blob.js(188, 7): '[_byteSequence]' was also declared here.
```

kitsonk: Well that clearly means they aren't being declared as the same type though, right?  You need to use `/* @type {Uint8Array} */` then, right?

lucacasonato: @kitsonk I urge you to try it. I declared `[_byteSequence];` with `/* @type {Uint8Array} */` and all other places where we assign to `this[_byteSequence]`, we only assign strictly typed `Uint8Array`, but the error won't go away.

op_crates/fetch/21_blob.js
kitsonk: 👍

op_crates/fetch/21_blob.js
kitsonk: Ok, for some reason in vscode we weren't getting another diagnostic that is the "cause" of it.  It appears that assignment in the constructor needs an explicit type annotation, even if the type being assigned is ignored, otherwise it will imply `any`.  I am not sure why that is the case.  So that was conflicting.

Also, the type annotation in the constructor on bytes is unnecessary, so if you do this in the constructor it works and the diagnostic goes away:

```diff
-      /** @type {Uint8Array} */
       const bytes = processBlobParts(parts, endings);
+      /** @type {Uint8Array} */
       this[_byteSequence] = bytes;
```

Or you could do this:

```diff
-      /** @type {Uint8Array} */
-       const bytes = processBlobParts(parts, endings);
-       this[_byteSequence] = bytes;
+      /** @type {Uint8Array} */
+      this[_byteSequence] = processBlobParts(parts, endings);
```

kitsonk: (didn't run into this in streams, because none of the slots get initialized in the constructor, they get set in sub functions and TypeScript doesn't freak out for some reason... 😒 )

kitsonk (approved):
LGTM

PR #9310, opened 2021-01-28
fix(cli/coverage): display mapped instrumentation line counts

caspervonb:
Follow up to #8995 that I didn't think about at the time.

This uses the number of original mapped source lines as the basis for the number of lines hit and found rather than just relying on the total number of lines in the instrumented script.

For example, on canary:

```
cover file:///home/caspervonb/deno/cli/tests/subdir/complex.ts ... 72.727% (8/11)
  25 | export function unused(
  26 |   foo: string,
  27 |   bar: string,
  28 |   baz: string,
-----|-----
  30 |   return complex(
  31 |     foo,
  32 |     bar,
  33 |     baz,
  34 |   );
  35 | }
```

This patch (the number of missing lines are matching with the number of lines rendered as missing):
```
cover file:///home/caspervonb/deno/cli/tests/subdir/complex.ts ... 50.000% (10/20)
  25 | export function unused(
  26 |   foo: string,
  27 |   bar: string,
  28 |   baz: string,
-----|-----
  30 |   return complex(
  31 |     foo,
  32 |     bar,
  33 |     baz,
  34 |   );
  35 | }
```

### DIFF
--- a/cli/tests/test_branch_coverage.out
+++ b/cli/tests/test_branch_coverage.out
@@ -4,7 +4,7 @@ test branch ... ok ([WILDCARD])
 
 test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out ([WILDCARD])
 
-cover [WILDCARD]/tests/subdir/branch.ts ... 66.667% (6/9)
+cover [WILDCARD]/tests/subdir/branch.ts ... 57.143% (4/7)
    4 |   } else {
    5 |     return false;
    6 |   }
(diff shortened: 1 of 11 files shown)


cli/tests/test_branch_coverage.out
```diff
@@ -4,7 +4,6 @@ test branch ... ok ([WILDCARD])
 
 test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out ([WILDCARD])
 
-cover [WILDCARD]/tests/subdir/branch.ts ... 66.667% (6/9)
-   4 |   } else {
```
caspervonb, line 8: Unexpected...

ry (approved):
LGTM - thanks @caspervonb - coverage is getting quite mature now.

PR #9311, opened 2021-01-28
test(std/node): fix `assertStats` and `assertStatsBigInt` if atime, mtime or birthtime are null

Liamolucko:
All the tests related to `std/node` `stat()` and `lstat()` would previously fail if `atime`, `mtime` or `birthtime` were null. This was because optional chaining always returns `undefined` (so `null?.foo === undefined`) and `Number(null) === 0`.


ry (approved):
Thanks

PR #9316, opened 2021-01-29
feat(lsp): add references code lens

kitsonk:
This add the infrastructure for providing code lenses in the LSP and adds the ability to display a lens for references for symbols in the code that aligns to the built-in functionality for TypeScript in VSCode.

I need to add an integration test for this feature.

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -9,6 +9,10 @@ use crate::module_graph::TypeScriptReference;
 use crate::tools::lint::create_linter;
 
 use deno_core::error::AnyError;
+use deno_core::serde::de;
+use deno_core::serde::Deserialize;
+use deno_core::serde::Deserializer;
+use deno_core::serde::Serialize;
 use deno_core::ModuleSpecifier;
 use deno_lint::rules;
 use lspower::lsp;
@@ -247,6 +251,30 @@ pub fn analyze_dependencies(
   } else {
     None
   }
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum CodeLensSource {
+  #[serde(rename = "references")]
+  References,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CodeLensData {
+  pub source: CodeLensSource,
+  #[serde(deserialize_with = "as_module_specifier")]
+  pub specifier: ModuleSpecifier,
+}
+
+fn as_module_specifier<'de, D>(
+  deserializer: D,
+) -> Result<ModuleSpecifier, D::Error>
+where
+  D: Deserializer<'de>,
+{
+  let url_str: String = Deserialize::deserialize(deserializer)?;
+  ModuleSpecifier::resolve_url(&url_str).map_err(de::Error::custom)
 }
 
 #[cfg(test)]
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -263,18 +263,7 @@ pub enum CodeLensSource {
 #[serde(rename_all = "camelCase")]
 pub struct CodeLensData {
   pub source: CodeLensSource,
-  #[serde(deserialize_with = "as_module_specifier")]
   pub specifier: ModuleSpecifier,
-}
-
-fn as_module_specifier<'de, D>(
-  deserializer: D,
-) -> Result<ModuleSpecifier, D::Error>
-where
-  D: Deserializer<'de>,
-{
-  let url_str: String = Deserialize::deserialize(deserializer)?;
-  ModuleSpecifier::resolve_url(&url_str).map_err(de::Error::custom)
 }
 
 #[cfg(test)]
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -9,9 +9,7 @@ use crate::module_graph::TypeScriptReference;
 use crate::tools::lint::create_linter;
 
 use deno_core::error::AnyError;
-use deno_core::serde::de;
 use deno_core::serde::Deserialize;
-use deno_core::serde::Deserializer;
 use deno_core::serde::Serialize;
 use deno_core::ModuleSpecifier;
 use deno_lint::rules;
--- a/cli/lsp/config.rs
+++ b/cli/lsp/config.rs
@@ -17,15 +17,53 @@ pub struct ClientCapabilities {
 
 #[derive(Debug, Clone, Default, Deserialize)]
 #[serde(rename_all = "camelCase")]
+pub struct CodeLensSettings {
+  #[serde(default)]
+  pub references: bool,
+  #[serde(default)]
+  pub references_all_functions: bool,
+}
+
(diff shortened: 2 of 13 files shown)


cli/lsp/config.rs
```diff
@@ -15,19 +15,57 @@ pub struct ClientCapabilities {
   pub workspace_did_change_watched_files: bool,
 }
 
+#[derive(Debug, Clone, Default, Deserialize)]
```
ry, line 18: Not necessary to fix in this PR, but generally we should avoid "Clone" unless necessary. I suppose this one needs to be Clone because WorkspaceSettings is clone. But do we really make copies of WorkspaceSettings? Maybe - but that seems like it could result in some situations where there are two different workplace configs.

cli/lsp/config.rs
```diff
@@ -15,19 +15,57 @@ pub struct ClientCapabilities {
...
+    }
+  }
+
+  pub fn enabled_code_lens_references(&self) -> bool {
+    if let Some(code_lens) = &self.code_lens {
+      code_lens.references
+    } else {
+      false
+    }
+  }
```
ry, line 58: Isn't this the same as `enabled_code_lens`? Confused.

cli/lsp/config.rs
```diff
@@ -15,19 +15,57 @@ pub struct ClientCapabilities {
...
   #[serde(default)]
   pub unstable: bool,
 }
 
+impl WorkspaceSettings {
+  /// Determine if any code lenses are enabled at all.  This allows short
+  /// circuiting when there are no code lenses enabled.
+  pub fn enabled_code_lens(&self) -> bool {
+    if let Some(code_lens) = &self.code_lens {
+      code_lens.references
```
ry, line 46: Is this meant to be `code_lens.references && code_lens.references_all_functions` ?

cli/lsp/language_server.rs
```diff
@@ -778,6 +819,210 @@ impl Inner {
...
+    let mark = self.performance.mark("code_lens");
+    let specifier = utils::normalize_url(params.text_document.uri);
+    let line_index = self.get_line_index_sync(&specifier).unwrap();
+    let navigation_tree =
+      self.get_navigation_tree(&specifier).await.map_err(|err| {
+        error!("Failed to retrieve nav tree: {:#?}", err);
+        LspError::invalid_request()
+      })?;
+
+    let cl = Rc::new(RefCell::new(Vec::new()));
```
ry, line 839: The use of Rc and RefCell here seem a bit questionable. But I assume you tried this with `let mut cl = Vec::new()` already and the types didn't allow it.

cli/lsp/language_server.rs
```diff
@@ -778,6 +819,210 @@ impl Inner {
...
+          }
+          _ => (),
+        }
+      }
+    });
+
+    self.performance.measure(mark);
+    Ok(Some(
+      Rc::try_unwrap(cl)
+        .map_err(|_| LspError::internal_error())?
```
ry, line 920: no need to handle the error gracefully, there should only be one reference here, use `unwrap()`

cli/lsp/language_server.rs
```diff
@@ -778,6 +819,210 @@ impl Inner {
...
+              |err| {
+                error!("Error processing TypeScript request: {}", err);
+                LspError::internal_error()
+              },
+            )?;
+          let maybe_references: Option<Vec<tsc::ReferenceEntry>> =
+            serde_json::from_value(res).map_err(|err| {
+              error!("Error deserializing response: {}", err);
+              LspError::internal_error()
+            })?;
```
ry, line 952: Not for this PR but in the future rather than calling `serde_json::from_value` after every call to `self.ts_server.request`, consider moving the `from_value` call into `request` by making it generic on the return type.

cli/lsp/tsc.rs
```diff
@@ -480,6 +481,59 @@ impl DocumentSpan {
   }
 }
 
+#[derive(Debug, Clone, Deserialize)]
```
ry, line 484: Does it need to be Clone?

cli/lsp/config.rs
kitsonk: No, because `code_lens.references_all_functions` is optional feature enhancement of `code_lens.references` and `code_lens.references` can be `true` while `code_lens.references_all_functions` can be `false`, but `code_lens.references_all_functions` requires `code_lens.references` to be `true` to have any effect.

cli/lsp/language_server.rs
kitsonk: Yes of course.  The closure below gives the compiler the heebie-jeebies when dealing with the mutable vector, so it isn't questionable, it is required when dealing with stuff in closures that you aren't moving in there.

cli/lsp/config.rs
kitsonk: I will try to remove `Clone` from `WorkspaceSettings`, I can't remember why it was there.

cli/lsp/config.rs
kitsonk: It is at the moment, because we only have one source of code lenses.  But the next one is implementations, and that means `enabled_code_lens` will have to consider `implementations` and there will be a `enabled_code_lens_implementations` method.

kitsonk: I will make that clear in the code doc.

cli/lsp/language_server.rs
kitsonk: Yeah, we do that a lot in quite a few places and yeah, it should be a generic.

cli/lsp/analysis.rs
```diff
@@ -249,6 +253,30 @@ pub fn analyze_dependencies(
...
+fn as_module_specifier<'de, D>(
+  deserializer: D,
+) -> Result<ModuleSpecifier, D::Error>
+where
+  D: Deserializer<'de>,
+{
+  let url_str: String = Deserialize::deserialize(deserializer)?;
+  ModuleSpecifier::resolve_url(&url_str).map_err(de::Error::custom)
+}
+
```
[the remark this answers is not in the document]
kitsonk: I think I got "yelled" at last time I tried to touch `ModuleSpecifier` but it does make sense.

cli/lsp/tsc.rs
kitsonk: Yes, because like `LineIndex`, when we provide it from the document cache to be operated on, we provide a clone, as providing a reference gets complex pointer lifetimes and we can't move it out.

cli/lsp/config.rs
kitsonk: Ok, it was there, because `Config` was cloned, but it appears that this was a legacy hold over from when we first adapted `rust_analyzer` structure, where sometimes the config was cloned.  There was only one place where it was cloned, which it actually shouldn't have been.  Thanks for pointing it out.

cli/lsp/language_server.rs
kitsonk: #9354 so we won't forget

kitsonk:
Ok, I added docs in a few places where there were non-obvious reasons for things.

ry (approved):
LGTM

PR #9321, opened 2021-01-29
cargo publish is done manually now

ry:


piscisaureus:
Why?

piscisaureus (approved):

bartlomieju (approved):
> Why?

Currently only `cli` is automatically published which requires all other crates in the repo to be published manually. Given growing number of crates adding multiple publish steps to CI makes it even more prone to failures on releases. I'm strongly in favor of this change.

Issue #8341, opened 2020-11-11
Make resources implement Transferable


jcc10:
Related to #5965 (and possibly others)

It would be really nice if resources could be Transferable between web workers. This would allow for proper multi-threading for HTTP and WebSocket servers.

This can provide multiple benefits:

- Isolated code exposure.  
If someone finds a exploit it has a lower likelihood of affecting the entire server, only the worker it was executed on.
- Load Balancing.  
This allows for balancing across all cores. This could be useful in a number of situations. Now obviously you could just make a new worker for each request and then delete the worker upon completion, but it may be easier conceptually for a programmer to just transfer the connection.
- Fault Tolerance.  
You can use workers similar to Erlang for fault tolerance. It's easy enough to ensure workers are alive with heartbeats and if they are not you can kill them and re-instate. Would that kick any WebSockets (or any other connection)? Yes. But it is probably better to kick 10% of the server intentionally then for the server to kick 100% due to lag.

There are probably others but those are the reasons I can think of.

This is probably not high up on the priority board, but it would be very nice if only for the use case of WebSockets.

00ff0000red:
I actually don't think that anyone has mentioned transferring objects yet, although structured cloning has been mentioned multiple times.

https://github.com/denoland/deno/issues/6887
https://github.com/denoland/deno/issues/6433 (This has the closest connection to this)
https://github.com/denoland/deno/issues/3557 (mentions ArrayBuffers, thus implicitly references Transferable)
https://github.com/denoland/deno/issues/3317#issuecomment-553281641 (Directly references Transferable)

Fortunately, it's on the [current roadmap[(https://github.com/denoland/deno/issues/7915)!

jcc10:
Just to be clear, this thread is specifically for resources which probably don't fit within the structured clone feature. (Can't really clone a connection, just transfer it.)

For both HTTP and WS, the main underlying item needed to be made transferable is ```interface Deno.Conn extends Reader, Writer, Closer``` everything else seems to be on top of that. (*I think*)

kitsonk:
I am not certain the concept of [Transferable](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) in the web API would be the best way to do what you are suggesting.

[...]

PR #9323, opened 2021-01-29
feat(runtime/worker): Structured cloning worker message passing.

### LINKED ISSUE #3557: Implement proper structured cloning between Workers
This is actually #3317, but we need a clearer issue name for tracking purposes.

Structured cloning implementation might be very useful in ways for transferring tasks between `Worker`s for load balancing, through low-cost transfer of `ArrayBuffer` (or even `SharedArrayBuffer`). For example, it might allow an HTTP request to move to another vacant thread if it somehow takes heavy CPU time to respond.

inteon:
Refactored webworkers to listen for host messages via an "async bin op" instead of directly injecting JSON into the runtime.

Fixes #3557.


bartlomieju:
> Refactored webworkers to listen for host messages via an "async json op" instead of directly injecting JSON into the runtime.
> First steps towards #3557, main goal is to change the message passing, so instead of JSON SharedArrayBuffers can be used (not yet part of this pull request).
> [WIP] because a waker still has to be re-added to `poll_event_loop`
> ↳ need some more time to figure out how this works (feel free to help)
> 
> Would be nice if some intermediate feedback could be provided already, on both the approach and the current code.

Thanks for the PR, but the current design is on purpose. I'd rather not change code before there's a working prototype of structured cloned message passing and currently we're still missing appropriate bindings in `deno_core` to do that (which is in my backlog).

> NOTE: code also contains some unrelated fixes for linting 

It's not a good idea to throw such unrelated changes into a PR. Besides current formatting and linting passes so I'm not sure how that's better.

inteon:
> > Refactored webworkers to listen for host messages via an "async json op" instead of directly injecting JSON into the runtime.
> > First steps towards #3557, main goal is to change the message passing, so instead of JSON SharedArrayBuffers can be used (not yet part of this pull request).
> > [WIP] because a waker still has to be re-added to `poll_event_loop`
> > ↳ need some more time to figure out how this works (feel free to help)
> > Would be nice if some intermediate feedback could be provided already, on both the approach and the current code.
> 
> Thanks for the PR, but the current design is on purpose. I'd rather not change code before there's a working prototype of structured cloned message passing and currently we're still missing appropriate bindings in `deno_core` to do that (which is in my backlog).
> 
> > NOTE: code also contains some unrelated fixes for linting
> 
> It's not a good idea to throw such unrelated changes into a PR. Besides current formatting and linting passes so I'm not sure how that's better.

Thank you for the feedback! I reverted the linting changes. Also, I will change the goal of this PR to create a structured cloning prototype. Like you said, that will make it easier to discuss what can be done better/ differently in my approach. While still making structured cloning work.
Then, the prototype can be used as a starting point to create the needed PRs to implement this feature (probably will involve multiple PRs).

Will update soon.

inteon:
@bartlomieju I pushed a full prototype, please let me know what you think.

(BTW: I'll check tomorrow if any tests failed)

inteon:
Current unsolved challenges in this prototype:
- `racy_worker` test is failing (not sure if something is wrong with test or with prototype)
- instead of JSON encoding arraybuffer, which was generated by the serialize instruction, create `transfer_to_worker_op`
- use same  `transfer_to_worker_op` to transfer other transferables using `post_message` #8341
- reevaluate use of  `op_worker_get_message`
- remove `core::mem::transmute` in "serialize_deserialize.rs"

core/serialize_deserialize.rs
```diff
@@ -0,0 +1,92 @@
...
+  buffer: &[u8],
+) -> Option<v8::Local<'s, v8::Value>> {
+  let mut array_buffers = ArrayBuffers::new(); // TODO(Inteon): fix array buffers
+  let mut value_deserializer =
+    SerializeDeserialize::deserializer(scope, &buffer, &mut array_buffers);
+  unsafe {
+    core::mem::transmute(
+      value_deserializer.read_value(scope.get_current_context()),
+    ) // TODO(Inteon): fix lifetime bug in rusty_v8?
+  }
```
bartlomieju, line 91: This bit needs to be addressed before landing; can you reproduce this problem as a standalone unit test?

core/serialize_deserialize.rs
```diff
@@ -0,0 +1,92 @@
...
+      backing_store,
+    ))
+  }
+}
+
+pub(crate) fn serialize(
+  scope: &mut v8::HandleScope,
+  obj: v8::Local<v8::Value>,
+) -> Option<Vec<u8>> {
+  let mut array_buffers = ArrayBuffers::new(); // TODO(Inteon): fix array buffers
```
bartlomieju, line 71: What needs to be fixed specifically?

core/serialize_deserialize.rs
```diff
@@ -0,0 +1,92 @@
+use rusty_v8 as v8;
```
bartlomieju, line 1: This file needs a copyright and some unit tests

core/bindings.rs
```diff
@@ -46,6 +47,12 @@ lazy_static! {
       v8::ExternalReference {
         function: decode.map_fn_to()
       },
+      v8::ExternalReference {
+        function: serialize.map_fn_to()
+      },
+      v8::ExternalReference {
+        function: deserialize.map_fn_to()
+      },
```
bartlomieju, line 55: These two bindings need some tests, please add a new file similar to `core/encode_decode_test.js` that exercises these bindings

bartlomieju (commented):
@inteon great work! There's a lot going on in this PR and it might be hard to land it in one go. Changes to `core/` look great, there are a few minor comments I have. 

I'm yet to properly go through changes to worker implementation; could you open a new PR with only changes to the `core/` directory so we can land them first?

core/serialize_deserialize.rs
inteon: Structured cloning allows normally for transfering arraybuffers, this still has to be implemented. It might be better to start with an implementation that does not yet support this.

core/serialize_deserialize.rs
inteon: The problem here is that the returned value from "value_deserializer.read_value(...)" should be able to outlive value_deserializer.

inteon: I created an issue for this: https://github.com/denoland/rusty_v8/issues/607.

core/bindings.rs
inteon: I added some tests in this PR https://github.com/denoland/deno/pull/9458.

core/serialize_deserialize.rs
bartlomieju: Yes, that makes sense, let's start with providing only structured clone and worry about transfers later

inteon:
Waiting for #9458 to land, also improvements to ops_messages are desired (see #9433 and #9457).

bartlomieju:
@inteon let me know if this PR is ready, I don't think we'll land it for 1.8 though

inteon:
> @inteon let me know if this PR is ready, I don't think we'll land it for 1.8 though

Indeed, let's not rush things. First lets make sure it works well.

inteon:
Blocked on #9457

runtime/js/01_errors.js
```diff
@@ -134,6 +134,14 @@
...
     }
   }
+  class WorkerError extends Error {
+    constructor(jsonString) {
+      const data = JSON.parse(jsonString);
+      super(data.error.message);
+      Object.assign(this, data);
+      this.name = "WorkerError";
+    }
+  }
```
bartlomieju, line 144: I'm not happy with this error - all errors defined in this file are available in `Deno.errors` but this one is not. Additionally I'm pretty sure that workers should throw plain `Error` or maybe even `DOMException`. @nayeemrmn thoughts?

runtime/js/99_main.js
```diff
@@ -74,66 +74,77 @@ delete Object.prototype.__proto__;
...
-      let handled = false;
-
-      const errorEvent = new ErrorEvent("error", {
-        cancelable: true,
-        message: e.message,
-        lineno: e.lineNumber ? e.lineNumber + 1 : undefined,
-        colno: e.columnNumber ? e.columnNumber + 1 : undefined,
-        filename: e.fileName,
-        error: null,
+  async function getHostMessages() {
```
bartlomieju, line 82: What's the advantage of polling for messages from within runtime instead of explicitly calling a function like `workerMessageRecvCallback`?

Also, please rename to `pollForMessages`.

runtime/ops/web_worker.rs
```diff
@@ -2,40 +2,113 @@
...
+#[derive(Deserialize)]
+struct HostUnhandledErrorArgs {
+  message: String,
+}
 
 pub fn init(
   rt: &mut deno_core::JsRuntime,
   sender: mpsc::Sender<WorkerEvent>,
-  handle: WebWorkerHandle,
+  receiver: Arc<AsyncMutex<mpsc::Receiver<Box<[u8]>>>>,
```
bartlomieju, line 25: Maybe the receiver could be part of `WebWorkerHandle`?

runtime/ops/web_worker.rs
```diff
@@ -2,40 +2,113 @@
...
+    .try_send(WorkerEvent::Error(generic_error(args.message)))
+    .expect("Failed to propagate error event to parent worker");
+  Ok(json!(true))
+}
+
+/// Get message from host as worker
+async fn op_worker_get_message(
+  receiver_ref: Arc<AsyncMutex<mpsc::Receiver<Box<[u8]>>>>,
+) -> Result<Vec<u8>, AnyError> {
+  let mut receiver = receiver_ref.try_lock()?;
```
bartlomieju, line 78: It should be just `lock`; if there are multiple `op_worker_get_message` ops dispatched it should be a panic indicating problem with internal code.

runtime/ops/web_worker.rs
```diff
@@ -2,40 +2,113 @@
...
+  bufs: &mut [ZeroCopyBuf],
+  mut sender: mpsc::Sender<WorkerEvent>,
+) -> Result<u32, AnyError> {
+  assert_eq!(bufs.len(), 1, "Invalid number of arguments");
+
+  let msg_buf: Box<[u8]> = (*bufs[0]).into();
+
+  match sender.try_send(WorkerEvent::Message(msg_buf)) {
+    Ok(()) => Ok(0),
+    Err(_) => Err(custom_error("NotConnected", "Host is not reacheable")),
```
bartlomieju, line 100: `NotConnected` is an IO error and we shouldn't use it for this purpose; it should be some internal type instead

runtime/ops/worker_host.rs
```diff
@@ -537,39 +484,57 @@ async fn op_host_get_message(
...
-      return Ok(json!({ "type": "close" }));
+      return Ok(vec![]);
     }
   };
 
-  let maybe_event = worker_handle.get_event().await?;
+  let maybe_event = match worker_handle.get_event().await {
+    Ok(event) => event,
+    Err(_) => {
+      return Err(custom_error("NotConnected", "Worker is not reacheable"))
```
bartlomieju, line 494: Ditto

runtime/ops/web_worker.rs
inteon: Here this is used too, is this different or should it change too?
https://github.com/denoland/deno/blob/26f7a3f185881c12081eb6d3cedbf9e637376d86/runtime/web_worker.rs#L76-L77

runtime/ops/web_worker.rs
inteon: Receiver is now part of "WorkerChannelsInternal" its the 'other end' of the communication line. WebWorkerHandle is only used to terminate the worker from within the worker.

inteon: Normally 'WebWorkerHandle' should be only used by the host in my opinion.

runtime/ops/web_worker.rs
inteon: Can I register a new error class in 01_errors.js for this, or should I create an error class in 11_worker.js. Not sure what best practices are here.

runtime/js/99_main.js
inteon: I renamed the function to `pollForMessages`.
I think its simpler to just register an ops function, also this is how the host receives messages from the workers too.

runtime/ops/web_worker.rs
inteon: I created a WorkerConnectionError for this, also this error triggers if --allow-read is not provided in combination with workers.

bartlomieju:
@inteon please rebase when you get a moment, let's land this feature for 1.10

inteon:
@bartlomieju finished the rebase, PR also includes some small worker related cleanup (changed wrong text in error & moved bench). I also revised WebWorkerHandle and WebWorkerInternalHandle to give better separation between cross-tread and same-thread handle.
NOTE: the ops-signatures probably are not yet optimal, I propose we optimize them in a follow-up PR

runtime/ops/web_worker.rs
```diff
@@ -16,26 +17,53 @@ pub fn init() -> Extension {
...
+            a.borrow::<WebWorkerInternalHandle>().clone()
+          };
+
+          let maybe_data = temp.get_message().await;
+
+          if let Some(data) = maybe_data {
+            return Ok(data);
+          }
+
+          Ok(Box::new([]))
```
bnoordhuis, line 41:
```suggestion
          Ok(maybe_data.unwrap_or_default())
```

runtime/web_worker.rs
```diff
@@ -36,38 +37,83 @@ use std::task::Context;
 use std::task::Poll;
 use tokio::sync::Mutex as AsyncMutex;
 
+pub type WorkerId = u32;
```
bnoordhuis, line 40: Consider turning this into a newtype (`pub struct WorkerId(u32);`) if you're worried about people mixing up ids with other integral values.

runtime/web_worker.rs
```diff
@@ -36,38 +37,83 @@ use std::task::Context;
...
+  terminated: Arc<AtomicBool>,
+  isolate_handle: v8::IsolateHandle,
+}
+
+impl WebWorkerInternalHandle {
+  /// Post WorkerEvent to parent as a worker
+  pub fn post_event(&self, event: WorkerEvent) -> Result<(), AnyError> {
+    let mut sender = self.sender.clone();
+    // If the channel is closed,
+    // the worker must have terminated but the termination message has not yet been recieved.
```
bnoordhuis, line 65:
```suggestion
    // the worker must have terminated but the termination message has not yet been received.
```
(Also: long line. I guess I can't complain if the linter doesn't but consider wrapping it at 80 columns.)

runtime/web_worker.rs
```diff
@@ -589,7 +586,7 @@ mod tests {
 
     let mut handle = handle_receiver.recv().unwrap();
 
-    let msg = json!("hi").to_string().into_boxed_str().into_boxed_bytes();
+    let msg = vec![34, 2, 104, 105].into_boxed_slice(); // "hi" encoded
```
bnoordhuis, line 589: Consider using serde_v8 to encode the value, here and below. The wire format isn't stable.

bnoordhuis (approved):

runtime/web_worker.rs
bnoordhuis: Sorry, yes. I suppose that makes it inconvenient to call out to. Consider it a very light suggestion then.

runtime/web_worker.rs
inteon: Fixed it

runtime/web_worker.rs
inteon: I added some TODO comments here, if accessing `Deno.core.serialize()` from rust would become easier this can be fixed; also depends on how often it breaks due to v8 internals changing

runtime/js/11_workers.js
```diff
@@ -269,7 +240,7 @@
               throw new Error("Unhandled error event reached main worker.");
             } else {
               core.opSync(
-                "op_host_unhandled_error",
+                "op_worker_unhandled_error",
```
bartlomieju, line 243: Nitpick: this op is called from the "host" side, so I'd prefer to keep `op_host_` prefix

runtime/js/11_workers.js
```diff
@@ -288,7 +260,7 @@
...
```
bartlomieju, line 263: Ditto

runtime/js/99_main.js
```diff
@@ -75,60 +75,71 @@ delete Object.prototype.__proto__;
...
-        cancelable: true,
-        message: e.message,
-        lineno: e.lineNumber ? e.lineNumber + 1 : undefined,
-        colno: e.columnNumber ? e.columnNumber + 1 : undefined,
-        filename: e.fileName,
-        error: null,
+  async function pollForMessages() {
+    while (!isClosing) {
+      const bufferMsg = await opGetMessage();
+      const data = core.deserialize(new Uint8Array(bufferMsg));
```
bartlomieju, line 86: Isn't call to `new Uint8Array` superfluous here?

runtime/ops/web_worker.rs
```diff
@@ -16,26 +17,49 @@ pub fn init() -> Extension {
         op_sync(move |state, _args: (), buf: Option<ZeroCopyBuf>| {
           let buf = buf.ok_or_else(null_opbuf)?;
           let msg_buf: Box<[u8]> = (*buf).into();
```
bartlomieju, line 19: @inteon could you please rewrite those op handlers to be functions in this module instead of closures? It's becoming hard to read them

runtime/ops/worker_host.rs
```diff
@@ -483,7 +464,7 @@ fn op_create_worker(
 
   // Setup new thread
   let thread_builder =
-    std::thread::Builder::new().name(format!("deno-worker-{}", worker_id));
```
bartlomieju, line 486: Please revert - otherwise there will only be a number in the thread name

runtime/web_worker.rs
```diff
@@ -321,7 +368,7 @@ impl WebWorker {
     // Instead of using name for log we use `worker-${id}` because
     // WebWorkers can have empty string as name.
     let script = format!(
-      "bootstrap.workerRuntime({}, \"{}\", {}, \"worker-{}\")",
```
bartlomieju, line 324: Ditto, please revert

runtime/web_worker.rs
```diff
@@ -370,82 +425,42 @@ impl WebWorker {
...
-      self.internal_channels.receiver.poll_next_unpin(cx);
-
-    if let Poll::Ready(maybe_msg) = maybe_msg_poll_result {
-      let msg =
-        maybe_msg.expect("Received `None` instead of message in worker");
-      let msg = String::from_utf8(msg.to_vec()).unwrap();
-      let script = format!("workerMessageRecvCallback({})", msg);
-
-      // TODO(bartlomieju): set proper script name like "deno:runtime/web_worker.js"
-      // so it's dimmed in stack trace instead of using "__anonymous__"
```
bartlomieju, line 428: Great to see all of that code go!

bartlomieju (commented):
@inteon thank you so much for this PR, it's really great that you managed to implement all of that functionality in so little code. I have only some minor/nitpicky comments, other than that it's ready to land. Great to see this landed for 1.10!

runtime/js/99_main.js
lucacasonato: I thought `Box<[u8]>` is auto serialized to Uint8Array... cc @AaronO

lucacasonato: Actually just checked - `ZeroCopyBuf` / `serde_v8::MagicBuffer` is auto serialized to a `Uint8Array`. You can `ZeroCopyBuf::from(Box<[u8]>)`.

runtime/ops/worker_host.rs
inteon: @bartlomieju please let me know if this is ok, or if I should still change it

runtime/js/99_main.js
inteon: I fixed this in my latest changes, also at the host side this was fixed.

inteon:
I optimized the worker <-> host message passing ops (removed all places where buffers were copied).
@AaronO I also made some small changes to serde_v8, please check if these are acceptable

runtime/ops/worker_host.rs
bartlomieju: I see, that sounds good

bartlomieju (approved):
LGTM as well ✅  thanks for seeing this through @inteon, this is a massive feature that was highly requested.

runtime/js/11_workers.js
```diff
@@ -39,26 +39,8 @@
...
-    // Current clone is made by parsing json to byte array and from byte array back to json.
-    // In that case "undefined" transforms to empty byte array, but empty byte array does not transform back to undefined.
-    // Thats why this special is statement is needed.
-    if (dataIntArray.length == 0) {
-      return undefined;
-    }
-    const dataJson = decoder.decode(dataIntArray);
-    return JSON.parse(dataJson);
-  }
-
```
ry, line 61: \o/

runtime/js/99_main.js
```diff
@@ -75,60 +75,71 @@ delete Object.prototype.__proto__;
...
-      const errorEvent = new ErrorEvent("error", {
-        cancelable: true,
-        message: e.message,
-        lineno: e.lineNumber ? e.lineNumber + 1 : undefined,
-        colno: e.columnNumber ? e.columnNumber + 1 : undefined,
-        filename: e.fileName,
-        error: null,
+  async function pollForMessages() {
+    while (!isClosing) {
+      const bufferMsg = await opGetMessage();
```
ry, line 85: I think it'd be more readable to write
```js
const bufferMsg = await core.opAsync("op_worker_get_message");
```

runtime/js/99_main.js
```diff
@@ -75,60 +75,71 @@ delete Object.prototype.__proto__;
   const onerror = () => {};
 
   function postMessage(data) {
-    const dataJson = JSON.stringify(data);
-    const dataIntArray = encoder.encode(dataJson);
+    const dataIntArray = core.serialize(data);
     opPostMessage(dataIntArray);
```
ry, line 79: Maybe just
```js
core.opSync("op_worker_post_message", null, dataIntArray);
```

serde_v8/src/magic/buffer.rs
```diff
@@ -14,6 +14,8 @@ pub enum MagicBuffer {
   ToV8(Cell<Option<Box<[u8]>>>),
 }
 
+unsafe impl Sync for MagicBuffer {}
```
ry, line 17: If a Mutex instead of a Cell is used, MagicBuffer will automatically inherit Sync.
```diff
diff --git a/serde_v8/src/magic/buffer.rs b/serde_v8/src/magic/buffer.rs
index c14284c7..7ffff861 100644
--- a/serde_v8/src/magic/buffer.rs
+++ b/serde_v8/src/magic/buffer.rs
@@ -1,9 +1,9 @@
 use rusty_v8 as v8;
 
-use std::cell::Cell;
 use std::fmt;
 use std::ops::Deref;
 use std::ops::DerefMut;
+use std::sync::Mutex;
 
 use super::zero_copy_buf::ZeroCopyBuf;
 
@@ -11,11 +11,9 @@ use super::zero_copy_buf::ZeroCopyBuf;
 // allowing us to use a single type for familiarity
 pub enum MagicBuffer {
   FromV8(ZeroCopyBuf),
-  ToV8(Cell<Option<Box<[u8]>>>),
+  ToV8(Mutex<Option<Box<[u8]>>>),
 }
 
-unsafe impl Sync for MagicBuffer {}
-
 impl MagicBuffer {
   pub fn new<'s>(
     scope: &mut v8::HandleScope<'s>,
@@ -36,7 +34,7 @@ impl Clone for MagicBuffer {
 
 impl Default for MagicBuffer {
   fn default() -> Self {
-    MagicBuffer::ToV8(Cell::new(Some(vec![0_u8; 0].into_boxed_slice())))
+    MagicBuffer::ToV8(Mutex::new(Some(vec![0_u8; 0].into_boxed_slice())))
   }
 }
 
@@ -73,7 +71,7 @@ impl DerefMut for MagicBuffer {
 
 impl From<Box<[u8]>> for MagicBuffer {
   fn from(buf: Box<[u8]>) -> Self {
-    MagicBuffer::ToV8(Cell::new(Some(buf)))
+    MagicBuffer::ToV8(Mutex::new(Some(buf)))
   }
 }
 
@@ -100,7 +98,7 @@ impl serde::Serialize for MagicBuffer {
         let value: &[u8] = &*buf;
         value.into()
       }
-      Self::ToV8(x) => x.take().expect("MagicBuffer was empty"),
+      Self::ToV8(x) => x.lock().unwrap().take().expect("MagicBuffer was empty"),
     };
     let hack: [usize; 2] = unsafe { std::mem::transmute(boxed) };
     let f1: u64 = hack[0] as u64;

```

serde_v8/src/magic/buffer.rs
```diff
@@ -32,6 +34,12 @@ impl Clone for MagicBuffer {
   }
 }
 
+impl Default for MagicBuffer {
+  fn default() -> Self {
+    MagicBuffer::ToV8(Cell::new(Some(vec![0_u8; 0].into_boxed_slice())))
+  }
+}
```
ry, line 41: I don't agree that the Default MagicBuffer should be an empty buffer like this.

I think it makes more sense if an empty buffer is explicitly constructed in op_worker_get_message() (runtime/ops/web_worker.rs:57). That is the only place `MagicBuffer::default()` is used.

serde_v8/src/magic/buffer.rs
```diff
@@ -88,7 +96,10 @@ impl serde::Serialize for MagicBuffer {
 
     let mut s = serializer.serialize_struct(BUF_NAME, 1)?;
     let boxed: Box<[u8]> = match self {
-      Self::FromV8(_) => unreachable!(),
+      Self::FromV8(buf) => {
+        let value: &[u8] = &*buf;
```
piscisaureus, line 100: The buffer gets copied here. This seems less optimal - is it intentional?

runtime/web_worker.rs
```diff
@@ -370,82 +427,42 @@ impl WebWorker {
...
-          return Poll::Ready(Ok(()));
+          return Poll::Pending;
         }
 
-        // Otherwise forward error to host
-        let mut sender = self.internal_channels.sender.clone();
-        sender
-          .try_send(WorkerEvent::Error(e))
-          .expect("Failed to post message to host");
+        panic!("coding error: either js is polling or the worker is terminted");
```
piscisaureus, line 462: "terminated"

serde_v8/src/magic/buffer.rs
piscisaureus, line 17: You can't implement `Sync` on a `Cell<Option<Box<[u8]>>>`.

Cell is probably not dangerous in practice here, but misleading because a Cell is very specifically not Sync. There's `UnsafeCell` for that.

Mutable `Option<Box<T>>` is definitely not ok to make `Sync`.

bartlomieju: Applied

serde_v8/src/magic/buffer.rs
inteon: No, this is not intentional; lets see if I can fix this.

bartlomieju: @inteon I already fixed it, used `&buf` instead

runtime/js/99_main.js
```diff
@@ -75,60 +75,71 @@ delete Object.prototype.__proto__;
...
-          e.lineNumber,
-          e.columnNumber,
-          e,
-        );
-        handled = ret === true;
-      }
+      try {
+        if (globalThis["onmessage"]) {
+          const result = globalThis.onmessage(msgEvent);
+          if (result && "then" in result) {
```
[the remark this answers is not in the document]
leopiccionia: An awaited sync value is converted to a resolved Promise (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)). It typically means that the function execution is suspended and a microtask is enqueued, but I don't know the internals of Deno's event loop.

A common idiom is:
```js
let result = task()
if (isPromise(result)) {
    result = await result
}
```

Issue #9324, opened 2021-01-29
[WIP] Cleanup/modernize std


sno2:
I'm just going to be attaching all of my cleanup PRs for the std in here.  Please check them out and review them to give me feedback🙏

PR #9325, opened 2021-01-29
cleanup(std/uuid): follow style guide, modernize code, update tests

sno2:
#9324

### DIFF
--- a/std/uuid/v4.ts
+++ b/std/uuid/v4.ts
@@ -1,21 +1,19 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 import { bytesToUuid } from "./_common.ts";
 
-const UUID_RE = new RegExp(
-  "^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$",
-  "i",
-);
+const UUID_RE =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
 /**
- * Validates the UUID v4
- * @param id UUID value
+ * Validates the UUID v4.
+ * @param id UUID value.
  */
-export function validate(id: string): boolean {
+export function validate(id: string) {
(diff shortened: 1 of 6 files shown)


std/uuid/v4.ts
```diff
@@ -1,21 +1,19 @@
...
+const UUID_RE =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
 /**
- * Validates the UUID v4
- * @param id UUID value
+ * Validates the UUID v4.
+ * @param id UUID value.
  */
-export function validate(id: string): boolean {
```
ry, line 13: Why remove the return value type?

sno2: It's implicit.  TypeScript can infer these typings.

lucacasonato: Our doc generator can not though, so please re-add the explicit types.

sno2: @lucacasonato That's probably something that we should fix, then.  Deno brags its seamless TS integration, so it would be a shame if people had to be over-explicit in their TS to just get documentation generated.

lucacasonato: Unrelated to this PR though, so please revert. Even if this were not a concern we prefer explicit return types, because it exposes errors related to missing/incorrect return statements in branches in a more logical way, and makes the code more readable for people who do not use a TypeScript integration in their editor.

sno2: > We should really add this to the style guide (we used to have a lint for this).

Was just typing a question if I should post a PR later tonight.

> Unrelated to this PR though, so please revert. Even if this were not a concern we prefer explicit return types, because it exposes errors related to missing/incorrect return statements in branches in a more logical way, and makes the code more readable for people who do not use a TypeScript integration in their editor.

Got it, I'll adjust the code after I finish my homework in 'bout two hours.

caspervonb:
Just a heads up.

Now is not the time to be adding new PRs to std as I'm aiming to cut the history tomorrow (still waiting for some things to land) to get it ready for review/hand-off on late Monday.

See https://github.com/denoland/deno/issues/9029

bartlomieju:

Issue #8099, opened 2020-10-24
Proposal: Enhance custom inspect API


kitsonk:
*Edit: the current proposal is here https://github.com/denoland/deno/issues/8099#issuecomment-757434256*

---

Probably should have thought of this when we did `Deno.customInspect` but we didn't.  As our console output has evolved, it makes it really hard to implement the same rich output.  For example if you have the following code:

```js
const h = new Headers();
h.append("key", "value");
console.log(h);
```

You would get something like the following output:

<img width="157" alt="a non coloured output string" src="https://user-images.githubusercontent.com/1282577/97069044-d6944700-1618-11eb-87f7-37365013b3ac.png">

Where as if you did:

```js
console.log({ key: "value" });
```

You would get:

<img width="115" alt="a coloured output string" src="https://user-images.githubusercontent.com/1282577/97069063-017e9b00-1619-11eb-940d-35428f56c4d9.png">

Which takes advantage of all the other benefits of the output algorithm.

Therefore if we had another symbol that returned a more structured set of information, we could use the same algorithms to colourise the output and otherwise pretty print it.  I would propose that we create `Symbol("Deno.richCustomInspect")` which would require a return of a tuple value of two strings.  The first string would be the _name_ for the inspector and the second value would be a JSON string, which would be parsed by the inspector.

So for example if I wanted a rich output for `Headers`, I would do something like this:

```js
class Headers {
  [Symbol.for("Deno.richCustomInspect]() {
    const output = {};
    for (const [key, value] of this[headersData]) {
      output[key] = value;
    }
    return [ "Headers", JSON.stringify(output) ];
  }
}
```

This would allow complex modelling of the internals, but ensure serialisation to a consistent format that can be handled by the inspector.

caspervonb:
Instead of adding more symbols, would be simpler if the results of calling custom inspect were just formatted with the object format specifier instead (raw strings are left alone with that).

kitsonk:
@caspervonb I assume you are suggesting something like this PR for `Event`? https://github.com/denoland/deno/pull/8103/files

It hadn't occurred to me that should be the way until I saw the PR.  We could easily clean up `Headers` with something like that.

caspervonb:
Pretty much!

Side note tho, customInspect signature should probably be given `depth:number` and `options: InspectOptions` so the formatter used can respects those settings.

kitsonk:
[...]

PR #9332, opened 2021-01-30
fix(op_crates): Don't use `Deno.inspect` in op crates

### LINKED ISSUE #9446: Uncaught (in promise) ReferenceError: Deno is not defined - when deno namespace is set to false in WebWorker
Don`t know how to reproduce this for now, but using a worker without deno namespace set to true is not possible for now, in some situations.

I mean it makes sense in the context that the error message is build with some "Deno stuff" - but this shouldnt be the case, when i dont allowed the deno namespace to be used at all.

Thrown Error: `"Uncaught (in promise) ReferenceError: Deno is not defined\n    return ${obj.constructor.name} ${Deno...",`

Feel free to look into this, for now its just a placeholder - will update this when i have more informations.

nayeemrmn:
Pass `Deno.inspect()` as an arg to the internal custom inspect implementations.

Fixes #9446.

### DIFF
--- a/cli/tests/084_worker_custom_inspect_worker.ts
+++ b/cli/tests/084_worker_custom_inspect_worker.ts
@@ -1,0 +1,2 @@
+console.log(new ReadableStream());
+close();
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -2685,6 +2685,11 @@ console.log("finish");
     assert_eq!(out, "");
   }
 
+  itest!(_084_worker_custom_inspect {
+    args: "run 084_worker_custom_inspect.ts",
+    output: "084_worker_custom_inspect.ts.out",
+  });
+
   itest!(js_import_detect {
     args: "run --quiet --reload js_import_detect.ts",
     output: "js_import_detect.ts.out",
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -2686,7 +2686,7 @@ console.log("finish");
   }
 
   itest!(_084_worker_custom_inspect {
-    args: "run 084_worker_custom_inspect.ts",
+    args: "run --allow-read 084_worker_custom_inspect.ts",
     output: "084_worker_custom_inspect.ts.out",
   });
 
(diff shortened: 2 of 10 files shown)


kt3k:
Providing 2 different implementations for each custom inspector looks too costly to me.

How about taking `Deno.inspect` as an argument of custom inspector and use it in the implementation? That should remove the dependency to the global Deno namespace.

```ts
[Symbol.for("Deno.customInspect")](inspect) {
  return `${this.constructor.name} ${
    inspect({ locked: this.locked })
  }`;
}
```

nayeemrmn:
Good point. I also realised the fallback implementations are defective because they don't quote string properties, which would require even more code to fix.

In #8099 it's suggested to do this as part of the public custom inspect API. I'll only do this for the internal custom inspect because a) it's yet to be argued if user code will need this mechanism and b) it can land now.

nayeemrmn:
I've changed my mind, I think using the Deno namespace in custom inspect under the assumption it's only called in Deno aligns with how we expect users to write browser-compatible modules.

nayeemrmn:
Reopening to fix #9446. This is needed because web workers can disable the Deno namespace. cc @bartlomieju

bartlomieju:
> Reopening to fix #9446. This is needed because web workers can disable the Deno namespace. cc @bartlomieju

LGTM, but could you please add a test showcasing #9446?

cli/tests/084_worker_custom_inspect_worker.ts
```diff
@@ -0,0 +1,2 @@
+console.log(new ReadableStream());
```
kt3k, line 1: 👍

cli/tests/integration_tests.rs
```diff
@@ -2685,6 +2685,11 @@ console.log("finish");
     assert_eq!(out, "");
   }
 
+  itest!(_084_worker_custom_inspect {
+    args: "run 084_worker_custom_inspect.ts",
```
kt3k, line 2689:
```suggestion
    args: "run --allow-read 084_worker_custom_inspect.ts",
```

kt3k (approved):
LGTM

bartlomieju (approved):
LGTM

PR #9334, opened 2021-01-30
refactor: rewrite File implementation

lucacasonato:
This PR contains a rewritten `File` implementation that follows the spec at https://w3c.github.io/FileAPI/#file-section. This implementation passes all relevant File related web platform tests.

### DIFF
--- a/op_crates/fetch/21_file.js
+++ b/op_crates/fetch/21_file.js
@@ -287,8 +287,57 @@
     }
   }
 
-  window.__bootstrap.blob = {
+  const _name = Symbol("[[Name]]");
+  const _lastModfied = Symbol("[[LastModified]]");
+
+  class File extends Blob {
+    /** @type {string} */
+    [_name];
+    /** @type {number} */
+    [_lastModfied];
+
+    /**
+     * @param {BlobPart[]} fileBits 
+     * @param {string} fileName 
+     * @param {FilePropertyBag} [options] 
+     */
+    constructor(fileBits, fileName, options) {
+      if (fileBits === undefined) {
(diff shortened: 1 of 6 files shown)


satyarohith (commented):
Looks good.

op_crates/fetch/21_file.js
```diff
@@ -287,8 +287,57 @@
     }
   }
 
-  window.__bootstrap.blob = {
+  const _name = Symbol("[[Name]]");
+  const _lastModfied = Symbol("[[LastModified]]");
```
kitsonk, line 291: nitpic, alphabetise

op_crates/fetch/21_file.js
```diff
@@ -287,8 +287,57 @@
...
+    [_name];
+    /** @type {number} */
+    [_lastModfied];
+
+    /**
+     * @param {BlobPart[]} fileBits 
+     * @param {string} fileName 
+     * @param {FilePropertyBag} [options] 
+     */
+    constructor(fileBits, fileName, options) {
```
kitsonk, line 304:
```suggestion
    constructor(fileBits, fileName, options = {}) {
```

Would obviate the need for the optional chaining below and should still be spec compliant.

kitsonk (approved):
LGTM, but a couple minor nit picks.

op_crates/fetch/21_file.js
lucacasonato: Not really. When passing `null` as the third parameter, the `= {}` wouldn't catch, and thus options could still not be a WebIDL dictionary. The optional chaining prevents this.

Issue #9337, opened 2021-01-31
WebSocket generating an error when .close() is called


elizeuangelo:
When I manually close the websocket without setting an exit code, like 1000, I get this error:

`error: Uncaught (in promise) TypeError: Cannot read property 'code' of undefined
              code: message.data.code,
                                 ^
    at WebSocket.#eventLoop (deno:op_crates/websocket/01_websocket.js:347:34)`

I believe there should be a default value = 1000 as it doesnt work with undefined values.

crowlKats:
Can't seem to be able to reproduce this

elizeuangelo:
Its a bot on discord: `wss://gateway.discord.gg/?v=8&encoding=json`

My Deno version is 1.7.1.

timonson:
Is this connected to this issue @DjDeveloperr ? https://github.com/denoland/deno/issues/8419

DjDeveloperr:
> 
> 
> Is this related to this issue #8419 @DjDeveloperr ?

I don't think - it just adds client-side default for code 1005 if none is sent.

timonson:
Oh right, thanks. In the case of the other issue it is code `1001` which is sent by browsers but not in deno on the SIGINT signal.

>The browser actually sends the closing control code 1001 to the server when the client closes the connection automatically. This doesn't happen using the deno client and CTRL-C. I don't think this is a big deal but maybe something to keep in mind.

xlambda:
```error: Uncaught (in promise) TypeError: Cannot read property 'code' of undefined
              code: message.data.code,
                                 ^
    at WebSocket.#eventLoop (deno:op_crates/websocket/01_websocket.js:348:34)```
The same problem with deno 1.7.5

PR #9339, opened 2021-01-31
fix(op_crates/websocket): default to close code 1005

DjDeveloperr:
Currently if Web Socket is closed without code, it will error while on Chrome it would close with code 1005 instead.

closes: https://github.com/denoland/deno/issues/9337


bartlomieju (changes requested):
Thanks for the PR @DjDeveloperr; <s>it looks like `std/wasi` and `test_util/wpt` submodules are outdated, please run `git submodule update --init --recursive` and recommit. Also we need a test for this change.</s> @crowlKats provided a test case

bartlomieju (approved):
LGTM, thank you @DjDeveloperr @crowlKats

PR #9340, opened 2021-01-31
feat(cli/tests): Integration tests to actually check cached file

yos1p:
This PR can improve integration_tests by actually check the content of the file cached by `deno cache` command. Ideally, this should also check for the dependency of the file of the module.


yos1p:
What do you think of this approach to check the cached file? CC @ry

cli/tests/integration_tests.rs
```diff
@@ -419,8 +420,38 @@ fn cache_test() {
...
   let out = std::str::from_utf8(&output.stdout).unwrap();
   assert_eq!(out, "");
-  // TODO(ry) Is there some way to check that the file was actually cached in
-  // DENO_DIR?
+
+  check_cached_file(module_url, std::path::PathBuf::from(deno_dir.path()))
+    .await;
+}
+
+async fn check_cached_file(module_url: url::Url, deno_dir: std::path::PathBuf) {
```
bartlomieju, line 428: This function is superfluous. Maybe just move this logic to the test that uses it?

yos1p: I was wondering if we can use the function to another test that requires checking the file. But since there's no other tests that require that, I'll merge to the test function. Thanks for suggestions.

cli/tests/integration_tests.rs
```diff
@@ -344,6 +345,33 @@ mod integration {
...
+      .await
+      .unwrap()
+      .text()
+      .await
+      .unwrap();
+    let mut module_cached = String::new();
+    let mut module_file = fs::File::open(cached_file).unwrap();
+    module_file.read_to_string(&mut module_cached).unwrap();
+
+    assert_eq!(module_original, module_cached);
```
kt3k, line 374: I think this only partially checks the feature of `deno cache` because `cli/tests/006_url_imports.ts` has dependencies `./cli/tests/subdir/mod2.ts` and `./cli/tests/subdir/print_hello.ts` but these are not checked. `deno cache` also caches the compiled versions of the scripts. That is not checked either, and probably a little hard to check.

Deno has HTTP_PROXY feature and I think it can be used for simulating no-internet environtment. How about running the above script (http://localhost:4545/cli/tests/006_url_imports.ts) with invalid HTTP_PROXY value? If that runs without error, that should check that everything is cached.

yos1p: I agree. Ideally can check for dependencies as well. But since we cannot do it programmatically, I guess it's safe to check it manually?

And for checking the compiled version, I think what we can do now is to try running the compiled js in `gen` folder, and assert the output is as expected. Sounds good?

kt3k: > And for checking the compiled version, I think what we can do now is to try running the compiled js in gen folder, and assert the output is as expected.

That's one approach. But I would recommend using `HTTP_PROXY` env var like the below:

```rust
let output = Command::new(util::deno_exe_path())
  .env("DENO_DIR", deno_dir.path())
  .env("HTTP_PROXY", "http://not-existing-server/")
  .current_dir(util::root_path())
  .arg("run")
  .arg(module_url.to_string())
  .output()
  .expect("Failed to spawn script");
```
With HTTP_PROXY set to not-existing-server this deno process can't access to the script hosted on the server, but it still can execute it. I think that proves the entire dependency is cached in the disk.

yos1p: I see. Then that would mean we don't have to check the file itself because successful run = successful cache?

kt3k: > Then that would mean we don't have to check the file itself because successful run = successful cache?

Yes, I think so.

cli/tests/integration_tests.rs
```diff
@@ -336,14 +337,41 @@ mod integration {
...
+    let out = std::str::from_utf8(&output.stderr).unwrap();
+    // Check if file and dependencies are written successfully
+    assert!(out.contains("host.writeFile(\"deno://subdir/print_hello.js\")"));
+    assert!(out.contains("host.writeFile(\"deno://subdir/mod2.js\")"));
+    assert!(out.contains("host.writeFile(\"deno://006_url_imports.js\")"));
+
+    let prg = util::deno_exe_path();
+    let output = Command::new(&prg)
+      .env("DENO_DIR", deno_dir.path())
+      .env("HTTP_PROXY", "http://nil")
```
kt3k, line 356: 👍

kt3k (approved):
@yos1p Thank you for updating! LGTM.

PR #9345, opened 2021-01-31
test: Fix --reload in integration_tests

### LINKED ISSUE #9286: Integration tests don't actually test --reload
Any integration tests that are trying to test `--reload` behaviour or other cache invalidation behaviour don't actually test anything.  In fact almost all the `--reload` flags in the integration tests are redundant as well, because every invocation of `deno_cmd()` creates a new temp `DENO_DIR`:

https://github.com/denoland/deno/blob/e7323002d9fe5ab834754592916aba7cb842613d/test_util/src/lib.rs#L1099-L1106

I discovered this by trying to test an issue where we actually expect the cache to persist, and it never did.

yos1p:
Closes #9286

This PR removes redundant "--reload" args because `util::deno_cmd` recreate DENO_DIR. This PR also fixes `ta_reload` in integration tests to actually test reload.


yos1p:
The test failed because `ts_reload` is expecting `host.writeFile` log to new file. It works previously because of #9286. 
Is this a bug in the CLI, or we should just change the test?

CC @kitsonk

kt3k:
hmm `--reload` option specified in `itest!` macro also look redundant because `run` method of `CheckOutputIntegrationTest` also uses `deno_cmd`.

kt3k:
how about using `deno_cmd_with_deno_dir` in `deno_cmd` to reduce the repetition of code?

yos1p:
@kt3k Thanks for the suggestions

kt3k (approved):
LGTM, but I'd like @bartlomieju and @kitsonk to check if this is heading in the right direction.

cli/tests/integration_tests.rs
```diff
@@ -1243,31 +1242,37 @@ mod integration {
...
     // check the output of the the bundle program.
+    let output_path = hello_ts.canonicalize().unwrap();
     assert!(std::str::from_utf8(&output.stderr)
       .unwrap()
       .trim()
-      .contains("host.writeFile(\"deno://002_hello.js\")"));
+      .contains(&format!(
+        "host.getSourceFile(\"{}\", Latest)",
+        url::Url::from_file_path(&output_path).unwrap().as_str()
+      )));
```
bartlomieju, line 1275: Please explain what this change does

yos1p: The previous assertion was simply check if the file is saved. The assertion works because the DENO_DIR is different (just like the issue). 

Meanwhile, `deno cache --reload` does not re-write the file if the latest file exists in DENO_DIR. Therefore, the previous assertion does not work. So the current assertion make sure that cache reload is the latest one.

bartlomieju (approved):
LGTM, thanks @yos1p

PR #9347, opened 2021-01-31
chore: make all tests annotated with `#[cfg(test)]`

magurotuna:
This PR annotates the tests that are not annotated currently with  `#[cfg(test)]`.

Tests in Rust work fine so long as they are annotated with `#[test]`, but adding `#[cfg(test)]` attributes to them is more desirable for the reasons described [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html).

### DIFF
--- a/cli/diff.rs
+++ b/cli/diff.rs
@@ -146,22 +146,27 @@ pub fn diff(orig_text: &str, edit_text: &str) -> String {
   diff
 }
 
-#[test]
-fn test_diff() {
-  let simple_console_log_unfmt = "console.log('Hello World')";
-  let simple_console_log_fmt = "console.log(\"Hello World\");";
-  assert_eq!(
-    colors::strip_ansi_codes(&diff(
-      simple_console_log_unfmt,
-      simple_console_log_fmt
-    )),
-    "1 | -console.log('Hello World')\n1 | +console.log(\"Hello World\");\n"
-  );
+#[cfg(test)]
+mod tests {
+  use super::*;
 
-  let line_number_unfmt = "\n\n\n\nconsole.log(\n'Hello World'\n)";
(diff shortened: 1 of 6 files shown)


kt3k (approved):
LGTM

caspervonb (approved):
LGTM

ry:
Thanks for the patch but this is just moving code around without any benefit. There’s nothing wrong with a standalone unit test.

 Unless I’m missing something?

closing without merge

magurotuna:
@ry 
In general, adding `#[cfg(test)]` would lead to faster compile time and less size of resulting artifact. Although this patch is so small that it will make a trivial improvement, I'm sure that small things add up to make a big difference. I'd be grateful if you could reconsider.

ry:
@magurotuna `#[test]` implies `#[cfg(test)]` doesn't it?

magurotuna:
@ry No, `#[test]` just tells cargo to run these functions when running `cargo test` command. With `#[test]` only, those functions would also be compiled even when not compiled under `test` feature.

[The following paragraph quoted from the book provides clear explanation:](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest)
> The #[cfg(test)] annotation on the tests module tells Rust to compile and run the test code only when you run cargo test, not when you run cargo build. This saves compile time when you only want to build the library and saves space in the resulting compiled artifact because the tests are not included. You’ll see that because integration tests go in a different directory, they don’t need the #[cfg(test)] annotation. However, because unit tests go in the same files as the code, you’ll use #[cfg(test)] to specify that they shouldn’t be included in the compiled result.

ry:
@magurotuna Ok - I remain skeptical but I can't figure out where the test attribute is implemented and can't prove otherwise. It's an innocent change regardless.

PR #9357, opened 2021-02-01
fix(cli): fix handling of non-normalized specifier

kt3k:
This addresses the 2nd case of #9277.

This PR keeps the mapping of original specifier and normalized specifier and uses the original specifier for looking up the source code instead of normalized specifier. This change prevents panic when compiling the specifier in non-normalized form.

### DIFF
--- a/cli/tests/compiler_api_test.ts
+++ b/cli/tests/compiler_api_test.ts
@@ -305,3 +305,17 @@ Deno.test({
     });
   },
 });
+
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM

A good solution, keeping it contained within the JavaScript side of things!

kt3k:
Thank you for your review! Merging.

PR #9359, opened 2021-02-01
chore: remove #![deny(warnings)] and disable warnings in ci instead

KNnut:
See https://rust-unofficial.github.io/patterns/anti_patterns/deny-warnings.html.


lucacasonato:
Hey, thanks for the contribution, but we intentionally use `#[deny(warnings)]`. We only target a single version of Rust (currently 1.49.0), and we address new warnings when we update Rust.

PR #9361, opened 2021-02-01
chore: remove std directory

caspervonb:
This removes the std folder from the tree.

Various parts of the tests are pretty tightly dependent on std (47 direct imports and 75 indirect imports, not counting the cli tests that use them as fixtures) so I've added std as a submodule for now.


bartlomieju (approved):
LGTM 🚀

Issue #9333, opened 2021-01-30
`console.log({ obj })` stringify for function objects differs between `deno` and `node`


rivy:
Stringification of function objects for `console.log({ obj })` ignores enumerable function properties.
It's not a deal breaker, but I'm adding this as an issue since, from the discussion for Issue #3165, I believe the goal is more parity.

```shell
>deno --version
deno 1.7.1 (release, x86_64-pc-windows-msvc)
v8 8.9.255.3
typescript 4.1.3

>deno eval "const x=()=>'x'; x.y=()=>'y'; x.z = function z() {return 'z'}; console.log({x});"
{ x: [Function: x] }

>node --version
v14.15.0

>node -e "const x=()=>'x'; x.y=()=>'y'; x.z = function z() {return 'z'}; console.log({x});"
{ x: [Function: x] { y: [Function (anonymous)], z: [Function: z] } }
```

rivy:
Looks to be fixed by #9363, released within `deno` v1.7.3.
Fast work. 🚀

Issue #3165, opened 2019-10-21
console.log() should stringify objects exactly as Node does


ry:
Match colors and semantics. 

node's stringify has been tweaked and optimized over many years, changing anything is likely surprising in a bad way.

mhvsa:
I'll work on it

Lonniebiz:
@mhvsa Did you notice that there seems to be a 5 property limit when it comes to displaying property values via `console.log`?
https://github.com/denoland/deno/issues/3213

tisd:
Hey @mhvsa, i wanted to work on this, can you suggest where to start.
Thanks.

mhvsa:
Sorry for being dead in Deno for quite a time.

Hey @tisd 

Node implementation: 
https://github.com/nodejs/node/blob/master/lib/internal/console/constructor.js

I would start with formatting and do colors later.

GL

tisd:
Thanks @mhvsa.

jsejcksn:
Can we presently override the behavior introduced in #1969 so that we can see values in the output, or do we need to use `JSON.stringify` until this issue is resolved?

brandonkal:
Formatting logic to port starts here:
https://github.com/nodejs/node/blob/a751389a14ddb3571e3ade22ac472b9aeb912bc0/lib/internal/util/inspect.js#L1778

fabrizzioalco:
Hey, i would like to check on this.

bartlomieju:
With #6591 landed I believe this issue can now be closed.

PR #9363, opened 2021-02-01
fix(console) - log function object properties / do not log non-enumerable props by default 

DavidDeSimone:
Addresses https://github.com/denoland/deno/issues/9333 

Running `deno eval "const x=()=>'x'; x.y=()=>'y'; x.z = function z() {return 'z'}; console.log({x});"` now properly logs the function's properties. I've added a basic test for this behavior.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2012,6 +2012,8 @@ declare namespace Deno {
     trailingComma?: boolean;
     /*** Evaluate the result of calling getters. Defaults to false. */
     getters?: boolean;
+    /** Show an object's non-enumerable properties. Defaults to false. */
+    showNonEnumerable?: boolean;
   }
 
   /** Converts the input into a string that has the same format as printed by
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2013,7 +2013,7 @@ declare namespace Deno {
     /*** Evaluate the result of calling getters. Defaults to false. */
     getters?: boolean;
     /** Show an object's non-enumerable properties. Defaults to false. */
-    showNonEnumerable?: boolean;
+    showHidden?: boolean;
   }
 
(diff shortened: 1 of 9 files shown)


caspervonb:
This changes the output of a bunch of things which also differs from Node (since Node was mentioned as the reason for the change).

```
> Array
[Function: Array] { [Symbol(Symbol.species)]: [Getter] }

> Regexp
[Function: RegExp] { [Symbol(Symbol.species)]: [Getter] }
```

DavidDeSimone:
Good catch @caspervonb . I think the root cause here is that deno logs non-enumerable symbols by default, while node doesn't:

```bash
user@ubuntu:~/proj/deno$ node -e "let x = {}; let s = Symbol('foo'); Object.defineProperty(x, s, {enumerable: false }); console.log(x)"
{}
user@ubuntu:~/proj/deno$ deno eval "let x = {}; let s = Symbol('foo'); Object.defineProperty(x, s, {enumerable: false }); console.log(x)"
{ [Symbol(foo)]: undefined }
```

If that behavior were changed, then this PR would be aligned with Node for function constructors like Array. It shouldn't hard to add that on.

DavidDeSimone:
I've updated this PR to have Deno.inspect not log non-enumerable properties by default, and I've added a new flag named "showNonEnumerable" to InspectOptions. Setting this flag to true will show non-enumerable properties, similar to setting showHidden=true in node's inspect function. 

I've updated the tests to reflect the new desired behavior, along with adding additional cases around the new flags.

caspervonb:
>  Setting this flag to true will show non-enumerable properties, similar to setting showHidden=true in node's inspect function.

If there's prior art for this option, maybe we should just adopt that prior art?

caspervonb (approved):

kt3k (approved):
Chrome seems doing the different thing for non-enumerable symbols, but we had #3165, so this change make sense to me. LGTM.

kt3k:
@DavidDeSimone Thank you for your contribution!

Issue #9002, opened 2021-01-05
PermissionDenied errors are misleading in self-contained binaries


bartlomieju:
```
~/p/g/d/deno ❯❯❯ deno compile --unstable https://deno.land/std@0.79.0/examples/cat.ts CODE_OF_CONDUCT.md 
Check https://deno.land/std@0.79.0/examples/cat.ts
Bundle https://deno.land/std@0.79.0/examples/cat.ts
Compile https://deno.land/std@0.79.0/examples/cat.ts
Emit cat
~/p/g/d/deno ❯❯❯ ./cat
error: PermissionDenied: read access to "CODE_OF_CONDUCT.md", run again with the --allow-read flag
    at processResponse (deno:core/core.js:223:11)
    at Object.jsonOpAsync (deno:core/core.js:240:12)
    at async Object.open (deno:runtime/js/30_files.js:44:17)
    at async file://$deno$/bundle.js:3:18
```

`run again with the --allow-read flag` makes no sense in self-contained binaries because after (#8738) permissions are embedded in the binary and can't be changed. 

In this situation we should provide different message.

CC @lucacasonato @nayeemrmn

yos1p:
Is it possible for this to be check during compile time? That'd be great so that users don't need to re-compile again.

bartlomieju:
> Is it possible for this to be check during compile time? That'd be great so that users don't need to re-compile again.

No, permission can only be checked during runtime.

Spoonbender:
I wonder how to fix this.
Currently, the error message is being built at `permissions.rs`:
```rust
impl PermissionState {
  /// Check the permission state.
  fn check(self, msg: &str, flag_name: &str) -> Result<(), AnyError> {
    if self == PermissionState::Granted {
      log_perm_access(msg);
      return Ok(());
    }
    let message = format!("{}, run again with the {} flag", msg, flag_name);
    Err(custom_error("PermissionDenied", message))
  }
}
```

Without much thought, I see three possible approaches here:
1. `check` will receive an argument telling it if we're in a self-contained binary, and build the error message accordingly. This will require flowing this argument throughout the callstack...
2. `check` will deduce if we're running in a self-contained binary autonomously, perhaps by checking some ambient/global state. Examples:
    - On startup, a self-contained binary will set some environment variable (`DENO__SELF_CONTAINED`) which `check` will inspect
    - `check` will inspect the `std::env::current_exe` to deduce if we are "deno" or some other process
[...]

PR #9367, opened 2021-02-02
refactor(runtime/permissions): rewrite permission handling

crowlKats:


runtime/permissions.rs
```diff
@@ -63,515 +65,528 @@ impl Default for PermissionState {
 
 #[derive(Clone, Debug, Default, Deserialize, PartialEq)]
 pub struct UnaryPermission<T: Eq + Hash> {
-  pub global_state: PermissionState,
+  #[serde(skip)]
+  pub name: String,
+  #[serde(skip)]
+  pub description: String,
```
nayeemrmn, line 71:
```suggestion
  pub name: &'static str,
  #[serde(skip)]
  pub description: &'static str,
```

runtime/permissions.rs
```diff
@@ -63,515 +65,528 @@ impl Default for PermissionState {
...
-        self.hrtime = PermissionState::Granted;
-      } else {
-        self.hrtime = PermissionState::Denied;
-      }
-    }
-    self.hrtime
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct BooleanPermission {
+  pub name: String,
+  pub description: String,
```
nayeemrmn, line 454:
```suggestion
  pub name: &'static str,
  pub description: &'static str,
```

nayeemrmn (commented):
How does the `name`, `description` mechanism work for the [de]serialising in `worker_host.rs`? I don't think anything should be `#[serde(skip)]`ed here.

crowlKats:
> How does the `name`, `description` mechanism work for the [de]serialising in `worker_host.rs`? I don't think anything should be `#[serde(skip)]`ed here.

well it creates a new instance of `Permissions` and merges it with the existing one. `name` & `description` are for handling of prompts, logs & errors. so these shouldn't be settable by the user.

runtime/permissions.rs
```diff
@@ -63,515 +65,528 @@ impl Default for PermissionState {
 
 #[derive(Clone, Debug, Default, Deserialize, PartialEq)]
 pub struct UnaryPermission<T: Eq + Hash> {
-  pub global_state: PermissionState,
+  #[serde(skip)]
+  pub name: String,
+  #[serde(skip)]
+  pub description: String,
+  pub state: PermissionState,
```
nayeemrmn, line 72: What's the reason for this rename? This should be rigorously named `global_state`.
```suggestion
  pub global_state: PermissionState,
```
It doesn't make sense for `unary_permission.state` to be `PermissionState::Prompt` while `unary_permission.query(foo)` can return `PermissionState::Granted`.

crowlKats: my bad, that was a mishap on my side

cli/tests/059_fs_relative_path_perm.ts.out
```diff
@@ -1,4 +1,4 @@
-[WILDCARD]error: Uncaught PermissionDenied: read access to "non-existent", run again with the --allow-read flag
+[WILDCARD]error: Uncaught PermissionDenied: Access to read "non-existent" required, run again with read permission
```
nayeemrmn, line 1: The message should contain the flag suggestion. The wording is a bit off for the global case as well. I sense that the capitalisation of the permission name is a problem? How about just:
```suggestion
[WILDCARD]error: Uncaught PermissionDenied: Requires read access to "non-existent", run again with --allow-read.
```

bartlomieju: I agree with Nayeem, the message should stay as is and ideally we could conditionally skip the part with flag to address #9002

crowlKats: @bartlomieju is it better now?
`Access to read "non-existent" required, run again with the --allow-read flag`

runtime/permissions.rs
```diff
@@ -63,515 +65,506 @@ impl Default for PermissionState {
...
+      None => self.0.clone(),
+      Some(port) => format!("{}:{}", self.0, port),
+    })
+  }
+}
+
+impl UnaryPermission<NetPermission> {
+  pub fn query<T: AsRef<str>>(
+    &self,
+    host: &Option<&(T, Option<u16>)>,
```
bartlomieju, line 311: This type looks funky, why use reference to option and then reference to tuple?

runtime/permissions.rs
```diff
@@ -63,515 +65,506 @@ impl Default for PermissionState {
...
+    let (resolved_path, display_path) = resolved_and_display_path(path);
+    self.query(&Some(&resolved_path)).check(
+      &format!("write to \"{}\"", display_path.display()),
+      self.name,
+    )
+  }
+}
+
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize)]
+pub struct NetPermission(pub String, pub Option<u16>);
```
bartlomieju, line 284: Nitpick: might make more sense to have named fields?

runtime/permissions.rs
```diff
@@ -63,515 +65,506 @@ impl Default for PermissionState {
...
     if let Some(host) = host {
-      let state = self.query_net(&Some(host));
+      let state = self.query(&Some(host));
       if state == PermissionState::Prompt {
-        let host_string = format_host(host);
-        if permission_prompt(&format!(
-          "Deno requests network access to \"{}\"",
-          host_string,
-        )) {
+        let host = NetPermission::new(host);
```
bartlomieju, line 351: Nitpick: it's quite hard to read so many nested conditionals, maybe you could reorganize the code to return early instead?

runtime/permissions.rs
```diff
@@ -63,515 +65,506 @@ impl Default for PermissionState {
...
 
-  pub fn request_env(&mut self) -> PermissionState {
-    if self.env == PermissionState::Prompt {
-      if permission_prompt("Deno requests access to environment variables") {
-        self.env = PermissionState::Granted;
-      } else {
-        self.env = PermissionState::Denied;
+  pub fn revoke<T: AsRef<str>>(
+    &mut self,
+    host: &Option<&(T, Option<u16>)>,
```
bartlomieju, line 388: Ditto

runtime/permissions.rs
```diff
@@ -63,515 +65,506 @@ impl Default for PermissionState {
...
-    }
-    self.query_net(host)
+  pub fn check(&self) -> Result<(), AnyError> {
+    self.state.check(self.description, self.name)
   }
+}
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Permissions {
+  pub read: UnaryPermission<ReadPermission>,
```
bartlomieju, line 466: I'm not sure why it's called `UnaryPermission` instead of `ListPermission`...

runtime/permissions.rs
```diff
@@ -642,7 +678,7 @@ fn permission_prompt(message: &str) -> bool {
     return false;
   };
   let msg = format!(
-    "️{}  {}. Grant? [g/d (g = grant, d = deny)] ",
+    "{}  ️Deno requests {}. Grant? [g/d (g = grant, d = deny)] ",
```
bartlomieju, line 681: Not sure if we want to add `Deno requests`, opinions @ry?

bartlomieju (commented):
Great work @crowlKats! I have a few small nitpicks regarding the implementation. Before landing I will wait for another review.

runtime/permissions.rs
nayeemrmn: > I'm not sure why it's called `UnaryPermission` instead of `ListPermission`...

Queries for that permission take one parameter.

runtime/permissions.rs
crowlKats: This is how it currently already is implemented. Not sure how simple it would be to change

runtime/permissions.rs
crowlKats: Made this tuple just to alignments to what is supplied as arguments in the other net calls, so tuple. I can change it though if you prefer

runtime/permissions.rs
crowlKats: This is the current, just that currently everywhere the function is called "Deno requests " is put in the request string. This is just deduplication

runtime/permissions.rs
bartlomieju: Sounds reasonable

cli/tests/059_fs_relative_path_perm.ts.out
nayeemrmn: IMO they are still less concise/uniform than the original messages.
- `Access to read "foo" required`
- `Access to write to "foo" required` 🤮
- `Access to network for "foo:80" required`
- `Access to run a subprocess required`

vs original:
- `read access to "foo" required`
- `write access to "foo" required`
- `network access to "foo:80" required`
- `access to run a subprocess required`

My suggestion:
- `Requires read access to "foo"`
- `Requires write access to "foo"`
- `Requires network access to "foo:80"`
- `Requires subprocess access`

bartlomieju: I agree with Nayeem

runtime/permissions.rs
bartlomieju: I see, then LGTM

cli/tests/059_fs_relative_path_perm.ts.out
crowlKats: `Requires read access to "foo"` sounds a bit off to me, as there is no subject for `Requires`. I know `deno` is implied, but it still sounds weird

runtime/ops/worker_host.rs
```diff
@@ -24,15 +26,17 @@ use deno_core::BufVec;
...
-use std::rc::Rc;
-use std::sync::Arc;
-use std::thread::JoinHandle;
+
+use crate::permissions::PermissionState;
+use crate::permissions::UnaryPermission;
+use crate::permissions::{
+  resolve_read_allowlist, resolve_write_allowlist, BooleanPermission,
+  NetPermission, Permissions, ReadPermission, WritePermission,
+};
```
ry, line 35: we prefer one line per import

runtime/permissions.rs
```diff
@@ -63,515 +65,506 @@ impl Default for PermissionState {
 
 #[derive(Clone, Debug, Default, Deserialize, PartialEq)]
 pub struct UnaryPermission<T: Eq + Hash> {
+  #[serde(skip)]
+  pub name: &'static str,
+  #[serde(skip)]
+  pub description: &'static str,
```
ry, line 71: Why serde skip? Is the Deserialize trait even used?

crowlKats: yes, for worker permissions

ry:
Just to be clear - there's no change in behavior beyond some reworking of error messages?

crowlKats:
> Just to be clear - there's no change in behavior beyond some reworking of error messages?

indeed. still works the same, just used differently. as proof you can see that the tests havent changed

cli/tests/059_fs_relative_path_perm.ts.out
nayeemrmn: > `Requires read access to "foo"` sounds a bit off to me, as there is no subject for `Requires`. I know `deno` is implied, but it still sounds weird

The implied subject is the operation that failed. I think plenty of error messages are structured like that.

crowlKats: I forgot for a moment that error message in JS would be `PermissionDenied: Requires read access to "foo", run again with --allow-read.`, so it would actually work. will rework the messages.

crowlKats: Alright, error messages now are 
```
Requires net access to "google.com", run again with the --allow-net flag
Requires read access to "./test.ts", run again with the --allow-read flag
Requires write access to "./test.ts", run again with the --allow-write flag
Requires env access, run again with the --allow-env flag
Requires run access, run again with the --allow-run flag
Requires plugin access, run again with the --allow-plugin flag
```
and debug log messages are
```
Granted net access to "www.google.com"
Granted read access to "./test.ts"
Granted write access to "./test.ts"
Granted env access
Granted run access
Granted plugin access
```
will this do?

runtime/permissions.rs
crowlKats: changed to `Option<&(T, Option<u16>)>`

ry (approved):
LGTM - nice clean up @crowlKats and sorry for the delay.

PR #9368, opened 2021-02-02
chore: use ThinLTO

KNnut:


ry:
Thanks for the patch - we need an analysis of the benefits in order to merge.

ry:
Thanks but needs analysis.

PR #9370, opened 2021-02-02
docs(style_guide/TS): prefer `interface` over `type` for object typings

### LINKED ISSUE #9369: docs(style_guide/ts): prefer `interface` over `type` for object typings
I think that we should always prefer to use the `interface` over the `type` keyword for creating object types.  TypeScript has a way to declare only an object type definition, therefore I think we should be preferring it for its only use case.

sno2:
closes #9369

### DIFF
--- a/docs/contributing/style_guide.md
+++ b/docs/contributing/style_guide.md
@@ -99,7 +99,7 @@ When designing function interfaces, stick to the following rules.
 export function resolve(
   hostname: string,
   family?: "ipv4" | "ipv6",
-  timeout?: number,
+  timeout?: number
 ): IPAddress[] {}
 
 // GOOD.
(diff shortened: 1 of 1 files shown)


nayeemrmn:
The style guide shouldn't include TS best practices which Deno doesn't have a unique concern about.

sno2:
> The style guide shouldn't include TS best practices which Deno doesn't have a unique concern about.

Should I make a PR to remove [this rule](https://deno.land/manual@v1.7.1/contributing/style_guide#export-all-interfaces-that-are-used-as-parameters-to-an-exported-member) 
, then?

kitsonk:
No changes are required.  Respectfully declined.

PR #9376, opened 2021-02-02
feat(runtime/permissions): prompt fallback

### LINKED ISSUE #3811: discussion: Bring back permission prompt behind a flag

crowlKats:
This commit adds `--prompt` flag that changes behavior of permission
checking to displaying interactive prompt instead of failing on missing 
permissions.

Based on #9367
Closes #3811


runtime/permissions.rs
```diff
@@ -83,10 +111,11 @@ impl UnitPermission {
...
   pub fn request(&mut self) -> PermissionState {
     if self.state == PermissionState::Prompt {
-      if permission_prompt(&format!("access to {}", self.description)) {
-        self.state = PermissionState::Granted;
-      } else {
-        self.state = PermissionState::Denied;
+      match permission_prompt(&format!("access to {}", self.description)) {
+        PromptResult::AllowAlways => self.state = PermissionState::Granted,
+        PromptResult::AllowOnce => return PermissionState::Granted,
+        PromptResult::DenyOnce => return PermissionState::Denied,
```
nayeemrmn, line 117: What do allow once and deny once mean for `Deno.permisssions.request()`? This should still use the current prompt.

runtime/permissions.rs
```diff
@@ -664,45 +715,63 @@ impl Permissions {
...
+      write: Permissions::new_write(&Some(vec![]), false),
+      net: Permissions::new_net(&Some(vec![]), false),
+      env: Permissions::new_env(true, false),
+      run: Permissions::new_run(&Some(vec![]), false),
+      plugin: Permissions::new_plugin(true, false),
+      hrtime: Permissions::new_hrtime(true, false),
+    }
+  }
+
+  pub fn prompt() -> Self {
```
kt3k, line 759: Is this used?

kt3k:
As far as I tested with 56a7ee6 this looks good to me 👍 

It might be better to add unit test of prompt-enabled permissions using `set_prompt_result` utility (something similar to `fn test_request()`

runtime/permissions.rs
crowlKats: not anymore, removed.

kt3k (approved):
LGTM

bartlomieju (approved):
LGTM, thank you @crowlKats for seeing this through!

runtime/permissions.rs
```diff
@@ -1319,4 +1351,61 @@ mod tests {
...
+      plugin: Permissions::new_plugin(false, true),
+      hrtime: Permissions::new_hrtime(false, true),
+    };
+
+    let _guard = PERMISSION_PROMPT_GUARD.lock().unwrap();
+
+    set_prompt_result(false);
+    assert!(perms.read.check(&Path::new("/foo")).is_err());
+    set_prompt_result(true);
+    assert!(perms.read.check(&Path::new("/foo")).is_ok());
```
nayeemrmn, line 1372: This behaviour is incorrect. After the first check fails it should be put into the denied state, the second should bypass the prompt and return an error as well.

runtime/permissions.rs
```diff
@@ -234,24 +251,32 @@ impl UnaryPermission<ReadDescriptor> {
...
     let (resolved_path, display_path) = resolved_and_display_path(path);
-    self
-      .query(Some(&resolved_path))
-      .check(self.name, Some(&format!("\"{}\"", display_path.display())))
+    self.query(Some(&resolved_path)).check(
+      self.name,
+      Some(&format!("\"{}\"", display_path.display())),
+      self.prompt,
+    )?;
+    self.granted_list.insert(ReadDescriptor(resolved_path));
```
nayeemrmn, line 261:
```suggestion
    let result = self.query(Some(&resolved_path)).check(
      self.name,
      Some(&format!("\"{}\"", display_path.display())),
      self.prompt,
    )?;
    if self.prompt {
      if result.is_ok() {
        self.granted_list.insert(ReadDescriptor(resolved_path));
      } else {
        self.denied_list.insert(ReadDescriptor(resolved_path));
      }
    }
    result
```
Ditto for the other kinds and similar for `UnitPermission`.

nayeemrmn (commented):
Sorry for the late review!

nayeemrmn (approved):
LGTM

kt3k:
Thank you, Crowl. Nice feature to have!

PR #9389, opened 2021-02-03
Run Internal JS Files in Strict Mode

### LINKED ISSUE #9375: All built in runtime/js/*.js, op_crates/**/*.js, core/*.js should run in strict mode
All the internal runtime files should start with `"use strict";` to force them to use strict mode. Currently they are all running in sloppy mode.

shadowtime2000:
Resolves #9375.


runtime/js/99_main.js
```diff
@@ -1,3 +1,4 @@
+"use strict";
```
shadowtime2000, line 1: Not sure if this file should be in strict mode because according to MDN, undeletable variables like `__proto__` can't be deleted in strict mode.

shadowtime2000: Tests did pass with this though...

shadowtime2000:
See [this message](https://github.com/denoland/deno/issues/9375#issuecomment-772741358).

PR #9391, opened 2021-02-03
Strict Mode for built in Runtime, Core, op_crates JS files

pawanakhil:
Added `"use strict";` for built in Runtime, Core, op_crates JS files. #9375


bnoordhuis (approved):
LGTM, thanks!

crypt096 (approved):

00ff0000red:
@pawanakhil Couldn't you have just passed in `--use-strict` as a  flag to V8 on startup?

https://github.com/denoland/deno/blob/a2adecbba6c66cf40b5655b0a03e50b3ba412fb2/core/runtime.rs#L148-L165

There's otherwise no way to execute sloppy JS in Deno anyway, and all new core scripts wouldn't need an explicit `"use strict"`, preventing bugs from accidentally slipping though, e.g. someone forgets the "use strict."

bnoordhuis:
`--use-strict` has user-visible effects, it changes `eval()` and `new Function()`. It'd make Deno not quite ECMA-compatible.

It also breaks Node.js compatibility. CJS scripts start in sloppy mode.

00ff0000red:
@bnoordhuis
eval is specified to execute in strict mode when called directly from strict mode code, which is all user code in Deno.
https://tc39.es/ecma262/#sec-eval-x

It would be observable if using an indirect call though, and you're right about the Function constructor, it's always in sloppy mode.

Honestly, I've never used the Node.js polyfill compatability stuff in Deno, but if users use it, then yes, it would be visable.

I consider it unfortunate that developers still use `eval` or `new Function`, but so be it, seems like `--use-strict` wouldn't have been a solution after all.

lucacasonato:
@00ff0000red We also have `Deno.core.evalContext`, which does not execute in strict mode.

PR #9393, opened 2021-02-03
docs(getting_started/permissions): Added table to docs regarding security implications.

jcc10:
1. Give the PR a descriptive title.  
Hope it qualifies.
2. Ensure there is a related issue and it is referenced in the PR text.  
Mentioned in #8490 that I was going to add something like this.
3. Ensure there are tests that cover the changes.  
None needed since there was no code change.
4. Ensure `cargo test` passes.  
Again, no code changes.
5. Ensure `./tools/format.js` passes without changing files.  
Again, no code changes.
6. Ensure `./tools/lint.js` passes.  
See last two lines.

If I made a error with the language in the plugin note, please correct me.

### DIFF
--- a/docs/getting_started/permissions.md
+++ b/docs/getting_started/permissions.md
@@ -85,6 +85,28 @@ Allow net calls to any host/url:
 deno run --allow-net fetch.ts
 ```
 
+### Security Table
+This is not a comprehensive list, it mostly exists to help visualize what you are exposing your system to when enabling a given permission.
+| Permission       | Allows Read | Allows Write | Allows Net | Allows Arbitrary Code Execution | Notes |
+|------------------|-------------|--------------|------------|---------------------------------|-------|
+| `-A`             | Entire Disk | Entire Disk  | Entire Net | Yes                             |       |
+| `--allow-plugin` | Entire Disk | Entire Disk  | Entire Net | Yes                             | [plugin] |
+| `--allow-run`    | Entire Disk | Entire Disk  | Entire Net | Yes                             | [run] |
+| `--allow-net`    | No          | No           | Entire Net | Through Web Workers             |       |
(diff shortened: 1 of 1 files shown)


heyheyhello:
Could you explain the arbitrary code execution in web workers for `--allow-net`? It's also not is "hosted at" means hosting a web worker or if that is a threat to Deno instances without any workers as well

jcc10:
> Could you explain the arbitrary code execution in web workers for `--allow-net`?

While code hosted via deno.land/x or github raw are mostly served via tags (that are reliably read-only). On other hosts there is the chance that code can be invisibly swapped out from under you in a similar fashion as has been done on NPM in the past. It may be overzealous to call it arbitrary since it (presumably) must still remain in the deno sandbox, but depending on other permissions it is still a major vulnerability to be aware of (especially with how many times it has happened to NPM's registry.

> It's also not clear if "hosted at" is only for hosting a web worker or if that is also a threat to Deno instances without any workers.
No, that is specifically related to hosting web-workers. I suppose that both eval'd code and dynamic imports can be a problem also.

A better solution may be to add a tool/script to run security checks similar to `deno lint` but that just informs you of what the script can do, (deno.read, net access, eval, dynamic imports, web workers, etc.) but there would obviously be ways around that through obfuscation also.

This was my best idea for improving clarity on the security (especially for someone who just wants to use deno programs that haven't/can't be compiled as opposed to writing programs.)

heyheyhello:
Ah ok I see what you're describing but it's not what I thought you meant.

The first security issue you're describing is that having web-hosted dependencies can be dangerous. I agree, but also it's one of many trade-offs and it's specific to Deno's flags or it as a runtime

I thought you meant "hosted at" in terms of where the Deno runtime is running from, but I see you're referring to dependencies again

Thanks for clarifying

PR #9396, opened 2021-02-04
feat(lsp): add TS quick fix code actions

kitsonk:
This adds the feature that integrates TypeScript quick fix code actions into the LSP.

An example of it looks in VSCode:

Feb-04-2021 11-36-53

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -299,12 +299,31 @@ pub struct CodeLensData {
   pub specifier: ModuleSpecifier,
 }
 
+fn code_as_string(code: &Option<lsp::NumberOrString>) -> String {
+  match code {
+    Some(lsp::NumberOrString::String(str)) => str.clone(),
+    Some(lsp::NumberOrString::Number(num)) => num.to_string(),
+    _ => "".to_string(),
+  }
+}
+
+/// Determines if two TypeScript diagnostic codes are effectively equivalent.
+fn is_equivalent_code(
+  a: &Option<lsp::NumberOrString>,
+  b: &Option<lsp::NumberOrString>,
+) -> bool {
+  let a_code = code_as_string(a);
+  let b_code = code_as_string(b);
+  FIX_ALL_ERROR_CODES.get(a_code.as_str())
+    == FIX_ALL_ERROR_CODES.get(b_code.as_str())
+}
+
 /// Return a boolean flag to indicate if the specified action is the preferred
 /// action for a given set of actions.
-fn get_is_preferred(
+fn is_preferred(
   action: &tsc::CodeFixAction,
   actions: &Vec<(lsp::CodeAction, tsc::CodeFixAction)>,
-  fix_priority: &u32,
+  fix_priority: u32,
   only_one: bool,
 ) -> bool {
   actions.iter().all(|(_, a)| {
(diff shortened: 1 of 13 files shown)


op_crates/fetch/lib.rs
```diff
@@ -82,7 +83,7 @@ pub fn init(isolate: &mut JsRuntime) {
 
 pub trait FetchPermissions {
   fn check_net_url(&self, _url: &Url) -> Result<(), AnyError>;
-  fn check_read(&self, _p: &PathBuf) -> Result<(), AnyError>;
+  fn check_read(&self, _p: &Path) -> Result<(), AnyError>;
```
kitsonk, line 86: This is an nightly clippy fix that seems no harm pulling forward.

cli/lsp/analysis.rs
```diff
@@ -264,6 +299,248 @@ pub struct CodeLensData {
   pub specifier: ModuleSpecifier,
 }
 
+/// Return a boolean flag to indicate if the specified action is the preferred
+/// action for a given set of actions.
+fn get_is_preferred(
```
ry, line 304: maybe just "is_preferred" is cleaner?

cli/lsp/analysis.rs
```diff
@@ -264,6 +299,248 @@ pub struct CodeLensData {
...
+  action: &tsc::CodeFixAction,
+  actions: &Vec<(lsp::CodeAction, tsc::CodeFixAction)>,
+  fix_priority: &u32,
```
ry, line 307: `&u32` seems odd...probably better to pass by value

cli/lsp/analysis.rs
```diff
@@ -264,6 +299,248 @@ pub struct CodeLensData {
...
+  }
+
+  /// Move out the code actions and return them as a `CodeActionResponse`.
+  pub fn get_response(self) -> Option<lsp::CodeActionResponse> {
+    let code_actions = self
+      .actions
+      .into_iter()
+      .map(|(c, _)| lsp::CodeActionOrCommand::CodeAction(c))
+      .collect();
+    Some(code_actions)
```
ry, line 486: Seems like this never can return None? Consider changing return type to just `CodeActionResponse`

cli/lsp/analysis.rs
```diff
@@ -264,6 +299,248 @@ pub struct CodeLensData {
...
+        let diagnostic_code = match &diagnostic.code {
+          Some(lsp::NumberOrString::String(code)) => code,
+          _ => "",
+        };
+        d.code == diagnostic.code
+          || (FIX_ALL_ERROR_CODES.contains_key(d_code)
+            && FIX_ALL_ERROR_CODES.get(d_code)
+              == FIX_ALL_ERROR_CODES.get(diagnostic_code))
+      }
+    }) {
```
ry, line 519: This conditional is quite a mouthful. Consider breaking it up for readability.

cli/lsp/analysis.rs
kitsonk: It does mean I have to wrap it in `Some()` where it get invoked, because the response back requires requires the `Option`.

kitsonk: ok, I like it better that way anyways.

ry (approved):
LGTM - tho I have not tried it locally.

Issue #9398, opened 2021-02-04
Panic in repl


Srinivasa314:
Steps to reproduce
* Launch deno repl
* type "Deno.internal."
* Press tab key

Error message:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: {"code":-32000,"message":"Value with given id is not an object"}', cli/tools/repl.rs:113:12
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(...)', cli/tools/repl.rs:315:23
```

Deno version : deno 1.7.1

caspervonb:
Thanks! fix incoming in a lil bit.

Srinivasa314:
This bug applies to all `Symbol`s. It can be triggered by
```
let obj={s:Symbol('foo')}
```
Then type "obj.s." and press tab

PR #9400, opened 2021-02-04
fix(repl): prevent symbol completion panic

caspervonb:
Fixes https://github.com/denoland/deno/issues/9398

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -1775,7 +1775,6 @@ mod integration {
       }
     }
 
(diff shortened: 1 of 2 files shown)


lucacasonato:
@caspervonb This only includes a new passing test?

caspervonb:
> @caspervonb This only includes a new passing test?

Oops, reset it to test but seems that was not in a detached state 😅

bartlomieju (approved):
LGTM, thanks Casper

PR #9405, opened 2021-02-05
fix(lsp): support codeAction/resolve

kitsonk:
This adds support for `codeAction/resolve` which supports lazily resolving items like "fix all" code actions which should make performance of the lsp slightly better when dealing with "fix all" code actions.

This will require an update to `vscode_deno` to use the 3.16 version of the lsp for the "fix all" code actions to work, which I will be raising soon.


ry (approved):
I haven't tested this myself, but LGTM

PR #9406, opened 2021-02-05
fix(lsp): properly cache built in assets

kitsonk:
When we migrated away from all the locks, there was a regression that was not caught immediately.  The `tsc::get_asset()` would attempt to modify the snapshot, but the problem was that the snapshot was a `.clone()` of the inner language server's assets, which meant that modifications to that where lost.  When we then attempted to do a hover on those assets, the inner language servers assets didn't have the retrieved asset, and therefore would throw an error.

Trying to modify the assets within the inner language server struct causes a lot of challenges with mutability.

This PR follows the same strategy we have for `Sources` which wraps the assets in an arc mutex and deals with the locking opaquely.  Also adds a test which failed before the patch to help ensure we don't regress again.

### DIFF
--- a/cli/lsp/assets.rs
+++ b/cli/lsp/assets.rs
@@ -1,0 +1,36 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+use deno_core::ModuleSpecifier;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+use super::text::LineIndex;
+
+/// An lsp representation of an asset in memory, that has either been retrieved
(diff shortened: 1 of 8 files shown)


bnoordhuis:
I have some misgivings about this PR:

1. Reintroducing mutexes means reintroducing potential for deadlocks and ABA bugs.

2. It's not okay to use a mutex to lie to the compiler about mutability. If something is `mut`, it should be `mut` - not `Arc<Mutex<Whatever>>`.

Let me pick up the remaining action item from #9271 and see if we can apply similar logic to the assets.

kitsonk:
Closing in favour of #9408

PR #9408, opened 2021-02-05
fix(lsp): remove Sources mutex

bnoordhuis:
The mutex was used to hide the fact that the Sources object mutates
itself when it's queried. Be honest about that and mark everything that
directly or indirectly mutates it as `mut`.

This is a follow-up to commit 2828690fc7bb510c3248dda7b1cda8793e789ca6
from last month ("fix(lsp): fix deadlocks, use one big mutex (#9271)")

Refs #9406.

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -1,6 +1,6 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-use super::text::LineIndex;
+use super::language_server;
 use super::tsc;
 
 use crate::ast;
@@ -13,7 +13,6 @@ use crate::tools::lint::create_linter;
 
 use deno_core::error::custom_error;
 use deno_core::error::AnyError;
-use deno_core::futures::Future;
 use deno_core::serde::Deserialize;
 use deno_core::serde::Serialize;
 use deno_core::serde_json::json;
@@ -353,21 +352,14 @@ fn is_preferred(
 
 /// Convert changes returned from a TypeScript quick fix action into edits
 /// for an LSP CodeAction.
-pub async fn ts_changes_to_edit<F, Fut, V>(
+pub(crate) async fn ts_changes_to_edit(
   changes: &[tsc::FileTextChanges],
-  index_provider: &F,
-  version_provider: &V,
-) -> Result<Option<lsp::WorkspaceEdit>, AnyError>
-where
-  F: Fn(ModuleSpecifier) -> Fut + Clone,
-  Fut: Future<Output = Result<LineIndex, AnyError>>,
-  V: Fn(ModuleSpecifier) -> Option<i32>,
-{
+  language_server: &mut language_server::Inner,
+) -> Result<Option<lsp::WorkspaceEdit>, AnyError> {
   let mut text_document_edits = Vec::new();
   for change in changes {
-    let text_document_edit = change
-      .to_text_document_edit(index_provider, version_provider)
-      .await?;
+    let text_document_edit =
+      change.to_text_document_edit(language_server).await?;
     text_document_edits.push(text_document_edit);
   }
   Ok(Some(lsp::WorkspaceEdit {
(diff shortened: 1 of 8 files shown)


cli/lsp/analysis.rs
```diff
@@ -352,21 +351,14 @@ fn is_preferred(
...
   changes: &[tsc::FileTextChanges],
-  index_provider: &F,
-  version_provider: &V,
-) -> Result<Option<lsp::WorkspaceEdit>, AnyError>
-where
-  F: Fn(ModuleSpecifier) -> Fut + Clone,
-  Fut: Future<Output = Result<LineIndex, AnyError>>,
-  V: Fn(ModuleSpecifier) -> Option<i32>,
-{
+  language_server: &mut language_server::Inner,
```
bnoordhuis, line 356: I opted to remove the callbacks (because they were cumbersome to change) and call back directly - and change `pub` to `pub(crate)` because `Inner` is `pub(crate)`. It's kinda closely coupled but I feel that's okay.

I considered breaking it out into a trait but that felt like premature optimization as long as there's a single consumer.

cli/lsp/language_server.rs
```diff
@@ -894,9 +901,10 @@ impl Inner {
       return Ok(None);
     }
     let line_index = self.get_line_index_sync(&specifier).unwrap();
-    let file_diagnostics: Vec<&Diagnostic> = self
+    let file_diagnostics: Vec<Diagnostic> = self
       .diagnostics
       .diagnostics_for(&specifier, &DiagnosticSource::TypeScript)
+      .cloned()
```
bnoordhuis, line 907: This is to work around `self` being borrowed both mutably and immutably in this method. Ditto the change to `add_ts_fix_action()`.

cli/lsp/language_server.rs
```diff
@@ -168,7 +168,7 @@ impl Inner {
   /// Only searches already cached assets and documents for a line index.  If
   /// the line index cannot be found, `None` is returned.
   fn get_line_index_sync(
-    &self,
+    &mut self,
     specifier: &ModuleSpecifier,
   ) -> Option<LineIndex> {
     let mark = self.performance.mark("get_line_index_sync");
```
ry, line 174: It seems odd that a method which is only retrieving information "get_line_index_sync" would need to be marked as mutable. 

The mut requirement arrises from the performance counters - which are basically an optional diagnostic utility. They could just as well be in a static global.

kitsonk: We could remove the performance counter from this block. While it is somewhat informative, its performance is always caught in an outer function, and if we were concerned about it we could measure performance at the client sites instead.

bnoordhuis: > The mut requirement arrises from the performance counters

It's because of `self.sources`, which mutates itself when you call its `get_line_index()` method (computes and caches the result lazily.)

FWIW, I considered doing `s/get_/compute_/g`, to indicate they're not simple getters but 🤷.

bnoordhuis:
Rebased and fixed the asset cache bug. I also took the opportunity to simplify the lookup logic a little.

kitsonk:
I will give this a proper review first thing Monday my time.

cli/lsp/analysis.rs
kitsonk: Yeah, that makes sense... I originally did it because I couldn't easily pass in the language server because of the multiple versions of the arc/mutex's and the multi threading versus any real desire for abstraction.

cli/lsp/language_server.rs
kitsonk: Ah, yeah... `source.get_line_index` is not idempotent.  `compute_` does make more sense.  Would be glad to change that down the road too if we don't deal with it in this PR.

cli/lsp/language_server.rs
```diff
@@ -501,6 +488,29 @@ impl Inner {
     self.performance.measure(mark);
     Ok(())
   }
+
+  pub(crate) fn document_version(
```
kitsonk, line 492: :+1: yeah, I should have had that there to begin with.

kitsonk (approved):
LGTM

It is really nice reviewing code of someone who knows what they are doing.  Thank you tons!

PR #9417, opened 2021-02-05
feat(runtime): stabilize Deno.link and Deno.linkSync

caspervonb:
This makes Deno.link and Deno.linkSync stable.
The permission required has been changed to read-write to ensure one cannot escape the sandbox.


ry (commented):
What if you have permission to some subtree of the file system and create a link outside of it?

caspervonb:
> What if you have permission to some subtree of the file system and create a link outside of it?

Requires read from the old path and write to the new path; so if either are outside of the sandbox it would result in a permission denied error.

However a there is a little potential caveat here with how link can be used with other system calls.

You could create a link from something with read-only permissions in the old path, and link that to a path that has write-permissions. Now that file that was read-only is writable.

So we might want to make this require read-write permissions on both the old and new paths.

caspervonb:
Any thoughts on the above "edge case" @piscisaureus and @bnoordhuis?

bnoordhuis:
SGTM, I can't find fault in your logic.

ry (approved):
LGTM - thanks for pushing this through Casper

ry:
(need merge)

Issue #9422, opened 2021-02-05
Feature: Strongly typed deserialization of JSON ops


jbellerb:
I've been messing with deno_core a bit and I found it a bit strange that JSON ops only work on loosely typed serde_json::Values given serde is capable of directly converting between JSON and Rust types. This can be changed by simply updating the type signatures of json_op_sync/async. Since serde_json::Value implements Deserialize/Serialize, very little code needs to be changed, however as json_op_sync/async are now generic, type inference is broken in some cases (see [cli/build.rs:146](https://github.com/denoland/deno/commit/643303694360f9152ca6cb52ccab8a044f560044#diff-fa76844bfec3519648a2d7493933ff3317fa3250dd913cc26d0e86b3b5d4465aR146)).

Since the changes required for this feature are so simple, I've made them on a [fork](https://github.com/jbellerb/deno/tree/generic_json_ops) of deno (which can be upgraded to a PR if there is interest) for testing. I've also updated the [deno_core examples](https://github.com/denoland/deno/commit/4782f3ec8d0232cd14222ab67dead9f94f0e4f28) and found that this change removed a lot of boilerplate. Finally, since serde now has more information about the types it is working with, serialization and deserialization should be faster, but I don't have benchmarks to back up that claim. Is there any interest in making this change? Thanks.

lucacasonato:
Please open a PR to discuss this further. Sounds interesting - benchmarks would be nice.

jbellerb:
I've created a PR (#9423) . I'll try to get some basic benchmarks based on [http_bench_json_ops](https://github.com/denoland/deno/blob/master/core/examples/http_bench_json_ops.rs) soon.

PR #9423, opened 2021-02-06
feat(core): Strongly typed deserialization of JSON ops

jbellerb:
(from #9422)

This PR makes json_op_sync/async generic to all Deserialize/Serialize types instead of the loosely-typed serde_json::Value. Since serde_json::Value implements Deserialize/Serialize, very little existing code needs to be updated, however as json_op_sync/async are now generic, type inference is broken in some cases (see [cli/build.rs:146](https://github.com/denoland/deno/pull/9423/commits/5e3143e76c0c9020cf8b7ab3db83d6471a44e40f#diff-fa76844bfec3519648a2d7493933ff3317fa3250dd913cc26d0e86b3b5d4465aR146)). I've found this reduces a good bit of boilerplate, as seen in the updated [deno_core examples](https://github.com/denoland/deno/commit/4782f3ec8d0232cd14222ab67dead9f94f0e4f28).

This change may also reduce serialization and deserialization overhead as serde has a better idea of what types it is working with. I am currently working on benchmarks to confirm this and I will update this PR with my findings.

### DIFF
--- a/cli/tests/unit/dispatch_json_test.ts
+++ b/cli/tests/unit/dispatch_json_test.ts
@@ -34,9 +34,10 @@ unitTest(function invalidPromiseId(): void {
       createNew: false,
     },
   };
-  const argsText = JSON.stringify([promiseId, argsObj]);
-  const argsBuf = new TextEncoder().encode(argsText);
-  const resBuf = Deno.core.send(opId, argsBuf);
+  const argsText = JSON.stringify(argsObj);
+  const reqText = JSON.stringify([promiseId, argsText]);
+  const reqBuf = new TextEncoder().encode(reqText);
+  const resBuf = Deno.core.send(opId, reqBuf);
   const resText = new TextDecoder().decode(resBuf);
   const resObj = JSON.parse(resText);
   console.error(resText);
--- a/cli/tests/unit/dispatch_json_test.ts
+++ b/cli/tests/unit/dispatch_json_test.ts
@@ -21,8 +21,8 @@ unitTest(function malformedJsonControlBuffer(): void {
 
 unitTest(function invalidPromiseId(): void {
   const opId = Deno.core.ops()["op_open_async"];
+  const promiseId = "1. NEIN!";
   const argsObj = {
-    promiseId: "1. NEIN!",
     path: "/tmp/P.I.S.C.I.X/yeah",
     mode: 0o666,
     options: {
@@ -34,7 +34,7 @@ unitTest(function invalidPromiseId(): void {
       createNew: false,
     },
   };
-  const argsText = JSON.stringify(argsObj);
+  const argsText = JSON.stringify([promiseId, argsObj]);
   const argsBuf = new TextEncoder().encode(argsText);
   const resBuf = Deno.core.send(opId, argsBuf);
   const resText = new TextDecoder().decode(resBuf);
--- a/cli/tests/unit/dispatch_json_test.ts
+++ b/cli/tests/unit/dispatch_json_test.ts
@@ -21,22 +21,7 @@ unitTest(function malformedJsonControlBuffer(): void {
 
 unitTest(function invalidPromiseId(): void {
   const opId = Deno.core.ops()["op_open_async"];
-  const promiseId = "1. NEIN!";
-  const argsObj = {
-    path: "/tmp/P.I.S.C.I.X/yeah",
-    mode: 0o666,
-    options: {
-      read: true,
-      write: true,
-      create: true,
-      truncate: false,
-      append: false,
-      createNew: false,
-    },
-  };
-  const argsText = JSON.stringify(argsObj);
-  const reqText = JSON.stringify([promiseId, argsText]);
-  const reqBuf = new TextEncoder().encode(reqText);
+  const reqBuf = new Uint8Array([0, 0, 0, 0, 0, 0, 0]);
   const resBuf = Deno.core.send(opId, reqBuf);
   const resText = new TextDecoder().decode(resBuf);
   const resObj = JSON.parse(resText);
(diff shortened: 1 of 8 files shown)


jbellerb:
On my local machine, http_bench_json_ops showed a 15% increase in throughput when deserializing directly to a struct. The raw numbers from wrk can be seen in [this gist](https://gist.github.com/jbellerb/83ad318bf2e0d2eaf803e2d8d34f4707). The CI benchmarks show similar numbers, with a 8% increase between the fastest recent CI bench I could find on master ([1834611552](https://github.com/denoland/deno/runs/1834611552)) and the one CI bench for this PR that got to http_bench_json_ops before getting cancelled ([542109582](https://github.com/denoland/deno/actions/runs/542109582)). Finally, I calculated a 30%(!?) improvement in latency compared to the average of the last 6 commits to master, but I think the difference being that large has more to do with the large variance between runs as opposed to the code actually being that much faster.

@lucacasonato, sorry I didn't notice your earlier PR. Would it make sense to include your fixes for error handling in this PR?

bartlomieju:
Thanks for the PR @jbellerb; I'm a little hesitant as it is a major change, but if benchmarks are favorable we should look into that. Please make sure all tests are passing.

lucacasonato:
> Would it make sense to include your fixes for error handling in this PR?

If you can get this PR to work without, lets not. Let's try to keep this PR as slim as possible.

lucacasonato:
I am very much in favor of this change. The change itself is relatively minor (only 70 LOC changed), but  is very very useful for the usability of ops. I think it would allow us to get rid of a few hundred lines across all of the ops (all of the `serde_json::{from,to}_value`).

kitsonk:
I'm in favour of this change too...  It is similiar to the way the lspower works on calls, and the way we want to move the integration of tsc response values as well.

jbellerb:
@bartlomieju I am trying to change as little as possible to make upgrading easy. Unless some code depends on the internal behavior of jsonOpAsync, the only changes have to do with the function types. In that case, simple migration (without taking advantage of the new feature) should work like this:

```
      +----> compiles?
      |        |   |
      |  +- no +   + yes -+
      |  |                |
      |  v                v
  annotate with      works like
serde_json::Value      before

```

bartlomieju:
> I am currently working on benchmarks to confirm this and I will update this PR with my findings.

@jbellerb you can use already existing benchmarks to compare against `master` branch; there's plenty of them that check JSON ops. The most important metrics for this PR would be throughput (requests per second) and max latency.

cli/tests/unit/dispatch_json_test.ts
```diff
@@ -35,8 +35,9 @@ unitTest(function invalidPromiseId(): void {
     },
   };
   const argsText = JSON.stringify(argsObj);
-  const argsBuf = new TextEncoder().encode(argsText);
-  const resBuf = Deno.core.send(opId, argsBuf);
+  const reqText = JSON.stringify([promiseId, argsText]);
```
lucacasonato, line 38: I am not in favor of this double JSON encoding of args. What is the reason for this?

jbellerb: The previous implementation threw a deno-specific `type_error` when the promiseId was invalid. Serde does not provide an interface for easily adding your own error types to decoding, so my best solution was to decode promiseId and the args separately. Without hand-writing a manual decode routine, this was the best way I could think of to separate the two decodes.

lucacasonato: Can I suggest you put the raw bytes representing the `promiseId` i64 into the first 8 bytes of the sent Uint8Array?The remaining bytes after the 8 raw bytes for the promise ID could then be the JSON bytes. That way we avoid this double JSON encode/decode.

jbellerb: That's a much better solution. In that case, would this test still be necessary?

lucacasonato: Yeah, for example if the sent array only has 4 bytes.

lucacasonato (approved):
LGTM now. Thanks for the effort @jbellerb.

Requesting additional review from @bartlomieju, and @ry.

core/core.js
```diff
@@ -213,24 +213,25 @@ SharedQueue Binary Layout
...
-  async function jsonOpAsync(opName, args = {}, ...zeroCopy) {
+  async function jsonOpAsync(opName, args = null, ...zeroCopy) {
     setAsyncHandler(opsCache[opName], jsonOpAsyncHandler);
 
-    args.promiseId = nextPromiseId++;
-    const argsBuf = encodeJson(args);
-    dispatch(opName, argsBuf, ...zeroCopy);
+    const promiseId = nextPromiseId++;
+    const reqBuf = core.encode("\0".repeat(8) + JSON.stringify(args));
+    new DataView(reqBuf.buffer).setBigUint64(0, BigInt(promiseId));
```
ry, line 221: So the first 8 bytes represent the promiseId encoded as a little endian u64 ? Is that only the case for JSON ops or also for minimal ops?

jbellerb: Yes, it's only for JSON ops. However, it looks like the two implementation are becoming more similar (#9457), so perhaps it would make sense to store the promiseId in the first 4 bytes as a big endian i32, like it is for minimal ops.

ry: That might allow them to share more code on the backend... I think it could be addressed in a follow up commit tho.

ry (approved):
LGTM - this looks great - nice work @jbellerb!

We need to watch the benchmarks carefully after this lands to see if there are any regressions. Potentially we will revert if there are problems - but I'm going to land it now.

jbellerb:
Thanks so much for reviewing this!

PR #9433, opened 2021-02-07
refactor(core/shared_queue): overflow_response overwrite bugfix & improved shared queue overflow performance

inteon:
I reviewed the shared queue message passing code (js <-> rs).
- The have_unpolled_ops variable in runtime.rs was changed to not be encapsulated by a Cell<>, as this is not needed.
- In runtime.rs the overflow_response logic was changed, now ALL async responces are always polled in poll_pending_ops and returned in an Vec, this also fixes a bug where the value in overflow_response could be overwritten by the second loop. Also both the request to check the shared_queue and the overflow data are send in one v8 function call now.

No performance degradation is observed in the benchmarks. In case the workload overflows the queue a lot, this implementation will probably be more performant as all async responses can be transfered from rs to js in 1 function call (maybe a test-case is needed here if for you this does not make sense).


core/runtime.rs
```diff
@@ -534,49 +534,42 @@ impl JsRuntime {
...
-        let msg = "Dynamically imported module evaluation is still pending but there are no pending ops. This situation is often caused by unresolved promise.";
-        return Poll::Ready(Err(generic_error(msg)));
-      }
+    if has_pending_dyn_module_evaluation
+      && !has_pending_ops
+      && !has_pending_dyn_imports
+    {
+      let msg = "Dynamically imported module evaluation is still pending but there are no pending ops. This situation is often caused by unresolved promise.";
+      return Poll::Ready(Err(generic_error(msg)));
     }
```
bartlomieju, line 573: Revert please, these changes are only adding diff lines.

core/runtime.rs
```diff
@@ -1316,50 +1309,38 @@ impl JsRuntime {
...
   }
 
-  fn poll_pending_ops(
-    &mut self,
-    cx: &mut Context,
-  ) -> Option<(OpId, Box<[u8]>)> {
+  fn poll_pending_ops(&mut self, cx: &mut Context) -> Vec<(OpId, Box<[u8]>)> {
     let state_rc = Self::state(self.v8_isolate());
-    let mut overflow_response: Option<(OpId, Box<[u8]>)> = None;
+    let mut overflow_response: Vec<(OpId, Box<[u8]>)> = Vec::new();
```
bartlomieju, line 1314: In most cases this vector will never be filled, hopefully creating it on every poll won't affect the hot path.

bartlomieju (changes requested):
> I reviewed the shared queue message passing code (js <-> rs).
> 
> * The core.js and shared_queue.rs files only include aesthetic changes for better readability and also more similar code between both files.

While I appreciate the work, these changes are highly subjective and are only adding to diff lines; I'd prefer to discuss them after fixing actual bugs.

> * The have_unpolled_ops variable in runtime.rs was changed to not be encapsulated by a Cell<>, as this is not needed.

👍 

> * In runtime.rs the overflow_response logic was changed, now ALL async responces are always polled in poll_pending_ops and returned in an Vec, this also fixes a bug where the value in overflow_response could be overwritten by the second loop. Also both the request to check the shared_queue and the overflow data are send in one v8 function call now.
> 

These changes are in the hot code path and I'm a bit worried about creating a lot of vectors, but let's verify on the benchmarks.

> Some changes might be a bit opinionated, I'm open for any adaptations that are still needed.
> No performance degradation is observed in the benchmarks. In case the workload overflows the queue a lot, this implementation will probably be more performant as all async responses can be transfered from rs to js in 1 function call (maybe a test-case is needed here if for you this does not make sense).

Yes, a bench case would be perfect as there's no benchmark for overflown responses; take a look at `cli/bench/main.rs` for example benchmarks.

core/runtime.rs
inteon: Ok, will do

core/runtime.rs
inteon: Some info I found online:
https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees
"Vec is and always will be a (pointer, capacity, length) triplet."
"However, the pointer may not actually point to allocated memory. In particular, if you construct a Vec with capacity 0 via Vec::new, vec![], Vec::with_capacity(0), or by calling shrink_to_fit on an empty Vec, it will not allocate memory."

-> so because by default it is a vector of length 0 (no allocations); most likely there will be very little impact on performance

inteon:
Ok, I reverted some of the changes and added an extra benchmark.

cli/tests/workers/large_message_worker.ts
```diff
@@ -0,0 +1,14 @@
...
+
+onmessage = function (e): void {
+  for (let i = 0; i <= 10; i++) {
+    if (i % 2 == 0) {
+      postMessage(dataLarge);
+    } else {
+      postMessage(dataSmall);
+    }
+  }
+};
```
bartlomieju, line 14: It's quite unfortunate that worker is needed to test it, there's definitely additional cost stemming from the fact that each worker runs on a separate thread.

cli/tests/workers_large_message_bench.ts
```diff
@@ -0,0 +1,40 @@
...
+function bench(): Promise<any> {
+  let promises = [];
+  for (let i = 0; i < 50; i++) {
+    promises.push(oneWorker(i));
+  }
+
+  let arr = new Array(100000);
+  for (let y of arr) {
+    for (let y of arr) {}
+  }
```
bartlomieju, line 35: What's the purpose of this hot loop?

cli/tests/workers/large_message_worker.ts
inteon: Currently workers are the easiest to use, as they can send JSON messages that are pulled by their parrent and these messages are easily too big for the shared queue. If the message is too big, the whole shared queue is omitted and also other messages experience performance deterioration.
Another way to overflow the shared queue is to send too many messages, but this proably is less noticeable.

cli/tests/workers_large_message_bench.ts
inteon: good point, is not needed anymore

core/runtime.rs
```diff
@@ -1913,6 +1899,71 @@ pub mod tests {
     });
   }
 
+  #[test]
+  fn overflow_res_async_combined_with_unref() {
```
bartlomieju, line 1903: I verified that this test fails on master

inteon: I did rerun to fix the tests

bartlomieju (approved):
LGTM, thanks @inteon. As stated previously this change is in the hot path, so I'll be watching benchmarks after landing this PR and might have to revert if performance deteriorates.

PR #9434, opened 2021-02-07
proof-of-concept for a 'Known CVE' header

wperron:
Proof of Concept for #8699. I made a simple script that I uploaded to my personal registry at https://deno.wperron.io/evil-module@1.0.0/mod.ts and added a custom header `X-Deno-Known-CVE` for the purposes of the PoC (it's really malicious, it just shells out to echo something).

I uploaded a tarball of the built binary for linux [here[(https://deno.wperron.io/dl/deno/d0778547/deno.tar.xz) and made a simple script to test it [here](https://gist.github.com/wperron/bc6dae244876195a541e748b0e382905). The result looks something like this in the console

Screenshot from 2021-02-07 16-19-48

I think a single header is sufficient, realistically I think it could include a URL like the example above that points to the actual description of the CVE. Even if it includes a label like "high" or "critical" in it, parsing out a URL should be something relatively easy - Keeping it simple would make it easier for external tooling to be created around this feature to detect vulnerabilities in a CI/CD pipeline for example.

### DIFF
--- a/cli/http_util.rs
+++ b/cli/http_util.rs
@@ -111,6 +111,14 @@ pub async fn fetch_once(
     );
   }
 
+  if let Some(cve) = headers.get("X-Deno-Known-CVE") {
+    eprintln!(
+      "{} {}",
+      crate::colors::red_bold("Known Vulnerability"),
+      cve.to_str().unwrap()
+    );
+  }
+
   for key in headers.keys() {
     let key_str = key.to_string();
     let values = headers.get_all(key);


kitsonk:
I'm not really supportive of this.  It gives an air of authority.  Anyone could set (or not set) the header, and gives people a false impression that there is some authority.  If there are open vulnerabilities, it would be much better to simply not serve the package than to push that to the end user to divine.  I think it is confusing to end users and open for malicious exploitation more than the potential to do good.

wperron:
> I'm not really supportive of this. It gives an air of authority. Anyone could set (or not set) the header, and gives people a false impression that there is some authority. If there are open vulnerabilities, it would be much better to simply not serve the package than to push that to the end user to divine. I think it is confusing to end users and open for malicious exploitation more than the potential to do good.

I can definitely see where you're coming from here. However, I think we need _some_ answer to the `npm audit` that people are used to. This is maybe (probably) too naive of a solution. To be fair, I think this problem of trust vis-a-vis the CVE reports is likely going to come up with or without the runtime being aware of this header or not. For example if supershadyregistry.com has a "vulnerability" page, it can publish anything it wants there, potentially misleading users too. I think ultimately the value of a CVE database comes from the trust that people put into the host (which also already the case for module registries anyway)

With that said, the idea behind this draft is too offer a "canonical" place for different registries to expose vulnerabilities, and _hopefully_ that'll make it easier to build tooling for it. @lucacasonato suggested that this could already be achieved by using the `X-Deno-Warning` header, I just think having a dedicated header will make it easier to build a crawler/parser to alert on vulnerabilities.

Also, registries are already free to not serve a module that has a known vulnerability of course. However from experience in the npm ecosystem, sometimes the dependency tree includes a Low vulnerability in a transitive dependency that ends up not being applicable to the downstream dependent. Having the dependency tree broken in those cases would be disruptive when it doesn't really need to be - I think end users can make that decision for themselves, but they need to be aware that there is a vulnerability in the first place to make that decision :shrug:.

It's not a hill I'll die on to be honest, but I think that as Deno grows and gets more adoption in the Enterprise, we'll likely see more demand for compliance-oriented tooling.

I'm also interested in hearing if there's another, better solution that could tackle this subject

kitsonk:
> I think we need some answer to the `npm audit` that people are used to

I disagree with this thought.  `npm audit` means you are putting trust into `npm`.  Putting accidental implicit trust into a centralised package registry is exactly what Deno doesn't want to do. I am very resistant to anything that leads towards that implicit trust.  `X-Deno-Warning` is more than sufficient my opinion to allow package hosters supply any sort of information they feel is important without Deno making any sort of implied attestations to what the content of that is.

wperron:
Yeah, I'll close this draft and go back to the drawing board on this one. I think there's still some opportunities on the vulnerability & auditing story, but this probably ain't it. I'll leave the discussion opened as well.

> I disagree with this thought. `npm audit` means you are putting trust into `npm`

Just to clarify on this point, I was drawing inspiration from the workflow on the cli end of things more than the centralized npm vulnerability datbase 😉 _of course_ deno shouldn't give preferential treatment to any registry, that's the whole basis for dependency management in deno, and I strongly believe in it.

PR #9436, opened 2021-02-08
fix(lsp): handle type deps properly

### LINKED ISSUE #9425: lsp: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists
Deno: 1.7.2
Extension: Canary v0.0.8

PS:
No issue when use v2.3.3

Screenshot from 2021-02-06 20-39-31

kitsonk:
Fixes #9425

Once #9408 lands though, we would want to refactor this to just pass the mut ref to the language server to `analyze_dependencies` I suspect in line with other refactors.

### DIFF
--- a/cli/lsp/documents.rs
+++ b/cli/lsp/documents.rs
@@ -92,11 +92,15 @@ pub struct DocumentCache {
 }
 
 impl DocumentCache {
-  pub fn analyze_dependencies(
+  pub fn analyze_dependencies<F>(
     &mut self,
     specifier: &ModuleSpecifier,
     maybe_import_map: &Option<ImportMap>,
-  ) -> Result<(), AnyError> {
+    get_maybe_type: &mut F,
+  ) -> Result<(), AnyError>
+  where
+    F: FnMut(&ModuleSpecifier) -> Option<analysis::ResolvedDependency>,
+  {
     if !self.contains(specifier) {
       return Err(custom_error(
         "NotFound",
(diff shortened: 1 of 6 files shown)


bnoordhuis (commented):
Changes LGTM but I did land the other PR earlier this morning so I'm afraid it's going to need a rebase. I can do that if you want.

kitsonk:
@bnoordhuis thanks, but I can rebase. 😁

kitsonk:
@bnoordhuis I rebased, but I substantially changed the way the dependencies were analyzed to avoid passing a language server reference as it was getting painful with the multiple borrows.  I'm still not totally happy with the sources dependency analysis, but that is a battle for another day.

cli/lsp/documents.rs
```diff
@@ -229,6 +193,27 @@ impl DocumentCache {
       .collect()
   }
 
+  pub fn set_dependencies(
+    &mut self,
+    specifier: &ModuleSpecifier,
+    maybe_dependencies: Option<HashMap<String, analysis::Dependency>>,
+  ) -> Result<(), AnyError> {
+    if !self.contains(specifier) {
```
bnoordhuis, line 201: It'd be slightly more neat and efficient to handle this as the `None` case from `get_mut()`.

cli/lsp/language_server.rs
```diff
@@ -640,16 +670,11 @@ impl Inner {
     self.documents.open(
       specifier.clone(),
       params.text_document.version,
-      params.text_document.text,
+      params.text_document.text.clone(),
```
bnoordhuis, line 673: I think you should be able to pass `&params.text_document.text` here if you change `open()` to take a `&str`. That method seems to be making a copy anyway.

bnoordhuis (approved):

PR #9441, opened 2021-02-08
feat(lsp): add implementations code lens

kitsonk:
Requires a client that provides a configuration with the `codeLens.implementations` set to `true` (e.g. denoland/vscode_deno#319)

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -501,13 +501,11 @@ impl NavigationTree {
   ) -> lsp::CodeLens {
     let range = if let Some(name_span) = &self.name_span {
       name_span.to_range(line_index)
+    } else if !self.spans.is_empty() {
+      let span = &self.spans[0];
+      span.to_range(line_index)
     } else {
-      if !self.spans.is_empty() {
-        let span = &self.spans[0];
-        span.to_range(line_index)
-      } else {
-        lsp::Range::default()
-      }
+      lsp::Range::default()
     };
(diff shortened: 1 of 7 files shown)


cli/lsp/tsc.rs
```diff
@@ -499,8 +499,18 @@ impl NavigationTree {
     specifier: &ModuleSpecifier,
     source: &CodeLensSource,
   ) -> lsp::CodeLens {
+    let range = if let Some(name_span) = &self.name_span {
```
kitsonk, line 502: ran into a bug when typing in the editor where sometimes the `name_span` can be undefined, so hardened this whole area

lucacasonato (approved):
Implementation looks good to me. I tried this locally, and it works as expected.

Unrelated to this PR, I did find some weird alignment issues with the `goto` underlines when hovering over something in VS Code while holding the `Ctrl` key. Unsure if this is our fault or if it is a VS Code bug.

https://user-images.githubusercontent.com/7829205/107208733-47ae0d80-6a02-11eb-9fff-8012d9349a39.mp4

kitsonk:
> Unrelated to this PR, I did find some weird alignment issues with the goto underlines when hovering over something in VS Code while holding the Ctrl key. Unsure if this is our fault or if it is a VS Code bug.

Those are different, I noticed them too and was going to investigate them.  We should open an issue.

lucacasonato:
> We should open an issue.

Sure. I'll do that.

PR #9442, opened 2021-02-08
`cargo publish` step has been removed. So remove the redundant config of fail-fast

youngjuning:
`cargo publish` has been removed in https://github.com/denoland/deno/commit/46d5843f753548415c87f3c8a868bba49c203b92#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f


youngjuning:
someone here?

bartlomieju:
Thanks for the PR @youngjuning but current setup is okay, we still want the caches to update in case of failure as well as upload built binaries. Closing without merge.

PR #9453, opened 2021-02-09
fix(runtime/tls): handle invalid host for connectTls/startTls

### LINKED ISSUE #9452: Deno.connectTls panics when given invalid hostname
Culprit is a call to `.expect()` in both `connectTls` and `startTls`. Should be a regular exception instead of a panic.

Trivial repro:

``` console
> await Deno.connectTls({hostname: '8.8.8.8', port: 443})
thread 'main' panicked at 'Invalid DNS lookup: InvalidDNSNameError', runtime/ops/tls.rs:208:45
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Object({"id": Number(2), "error": Object({"code": Number(-32000), "message": String("Execution context was destroyed.")})})', runtime/inspector.rs:978:8
stack backtrace:
...
```

erik:
Tiny fix, converts a process-killing panic into a standard runtime error.

Before: 
``` console
> await Deno.connectTls({hostname: '8.8.8.8', port: 443})
thread 'main' panicked at 'Invalid DNS lookup: InvalidDNSNameError', runtime/ops/tls.rs:208:45
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Object({"id": Number(2), "error": Object({"code": Number(-32000), "message": String("Execution context was destroyed.")})})', runtime/inspector.rs:978:8
stack backtrace:
...
```

After:
```
> await Deno.connectTls({hostname: '8.8.8.8', port: 443})
Uncaught Error: Invalid DNS lookup
    at processResponse (deno:core/core.js:213:11)
    at Object.jsonOpAsync (deno:core/core.js:230:12)
    at async Object.connectTls (deno:runtime/js/40_tls.js:32:17)
    at async <anonymous>:2:1
```

Fixes #9452.

### DIFF
--- a/runtime/ops/tls.rs
+++ b/runtime/ops/tls.rs
@@ -140,8 +140,8 @@ async fn op_start_tls(
   }
 
   let tls_connector = TlsConnector::from(Arc::new(config));
-  let dnsname =
-    DNSNameRef::try_from_ascii_str(&domain).expect("Invalid DNS lookup");
+  let dnsname = DNSNameRef::try_from_ascii_str(&domain)
+    .map_err(|_| generic_error("Invalid DNS lookup"))?;
   let tls_stream = tls_connector.connect(dnsname, tcp_stream).await?;
 
   let rid = {
@@ -202,8 +202,8 @@ async fn op_connect_tls(
     config.root_store.add_pem_file(reader).unwrap();
   }
   let tls_connector = TlsConnector::from(Arc::new(config));
-  let dnsname =
-    DNSNameRef::try_from_ascii_str(&domain).expect("Invalid DNS lookup");
+  let dnsname = DNSNameRef::try_from_ascii_str(&domain)
+    .map_err(|_| generic_error("Invalid DNS lookup"))?;
   let tls_stream = tls_connector.connect(dnsname, tcp_stream).await?;
   let rid = {
     let mut state_ = state.borrow_mut();
--- a/cli/tests/unit/tls_test.ts
+++ b/cli/tests/unit/tls_test.ts
@@ -20,6 +20,24 @@ unitTest(async function connectTLSNoPerm(): Promise<void> {
   }, Deno.errors.PermissionDenied);
 });
 
+unitTest(
+  { perms: { read: true, net: true } },
+  async function connectTLSInvalidHost(): Promise<void> {
+    const listener = await Deno.listenTls({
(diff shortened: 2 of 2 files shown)


kt3k (approved):
LGTM

lucacasonato:
I'm not to happy about it being a generic `Error` - anyone have any suggestions what `Deno.errors` error could be used instead?

erik:
There are a few other generic error references in the function already. If there's a good alternative available, I can update those as well

runtime/ops/tls.rs
```diff
@@ -140,8 +140,8 @@ async fn op_start_tls(
   }
 
   let tls_connector = TlsConnector::from(Arc::new(config));
-  let dnsname =
-    DNSNameRef::try_from_ascii_str(&domain).expect("Invalid DNS lookup");
+  let dnsname = DNSNameRef::try_from_ascii_str(&domain)
+    .map_err(|_| generic_error("Invalid DNS lookup"))?;
```
lucacasonato, line 144: Could you actually put the original rust error message into the generic_error too? Debugging the error in the JS side is very hard otherwise.

runtime/ops/tls.rs
```diff
@@ -202,8 +202,8 @@ async fn op_connect_tls(
     config.root_store.add_pem_file(reader).unwrap();
   }
   let tls_connector = TlsConnector::from(Arc::new(config));
-  let dnsname =
-    DNSNameRef::try_from_ascii_str(&domain).expect("Invalid DNS lookup");
+  let dnsname = DNSNameRef::try_from_ascii_str(&domain)
+    .map_err(|_| generic_error("Invalid DNS lookup"))?;
```
lucacasonato, line 206: ditto

runtime/ops/tls.rs
erik: @lucacasonato so in this case it seems there isn't a good rust error message to base it off of. 

See here for the error type we're looking at: https://briansmith.org/rustdoc/src/webpki/name.rs.html#77-82

I could update the generic error message if you have a more clear message in mind.

bnoordhuis (approved):
@erik is right that `InvalidDNSNameError` provides no further context. I suggest merging this as-is.

lucacasonato (approved):
LGTM then

PR #9454, opened 2021-02-09
fix(lsp): handle code lenses for non-documents

kitsonk:
This PR fixes two issues which were causing code lenses for assets and cached dependencies to not be properly handled.  The first bug was that tsc NavigationTrees were not being calculated and the second was that the code lens resolution was not formatting the URL properly to be consumed by an lsp client.


lucacasonato (approved):
LGTM

PR #9457, opened 2021-02-09
refactor(deno/core): Move buffer_ops to deno core & merge logic with json_ops

### LINKED ISSUE #8173: dispatch_minimal should use promise table from Deno.core
Currently we have two different dispatch mechanism implemented:
- minimal - used for io ops (`op_read` and `op_write`)
- JSON - used for all other ops

Each of these mechanism has separate promise table that is used to store promises awaiting response from Rust.

Duplication should be removed and only promise table should be managed by `Deno.core` namespace.

inteon:
This PR is an effort to allow for more general js <-> rs message passing. It fixes #8173, by moving minimal ops to deno core.
This new system can be used to efficiently send and receive array-buffer messages (like minimal ops), but allso allows to return messages of arbitrary length (can be used for worker communication).


inteon:
Will rebase once #9457 is merged.

inteon:
Used by #9323.
On my local machine benchmarks seem to be comparable with performance before change, but I would appreciate it if someone can confirm this.

bartlomieju (commented):
@inteon there a lot going on in this PR. In general I think these changes are desirable, but to be honest it's quite hard to approve this PR with confidence as it touches as lot of critical infrastructure. Could you maybe split it into two smaller PRs? First which reorganizes file by moving existing implementation of "minimal ops" to "core/", and second which unifies implementation between "json" and "buffer" ops.

inteon:
@bartlomieju I'll see what I can do. I plan to work on this tomorrow.

inteon:
Blocked on #9721, #9719 and #9720.

bartlomieju (commented):
@inteon looks great, could you rewrite `core/examples/http_bench_bin_ops.rs` to use this new dispatch?

core/Cargo.toml
```diff
@@ -27,8 +27,8 @@ smallvec = "1.6.1"
 url = { version = "2.2.0", features = ["serde"] }
 
 [[example]]
-name = "http_bench_bin_ops"
-path = "examples/http_bench_bin_ops.rs"
+name = "http_bench_buffer_ops"
+path = "examples/http_bench_buffer_ops.rs"
```
bartlomieju, line 31: Can you please keep previous names? Otherwise benchmark page might break or lose results for previous examples.

inteon: Maybe in general, instead of "buffer ops", "bin ops" should be used as name? What do you think?

bartlomieju: @inteon I think that's a good idea, after all we're exchanging binary encoded data :) once you're done with renaming I think it's good to land

bartlomieju (approved):
LGTM, thanks you @inteon, very nice cleanup. As with previous PRs I will be monitoring benchmarks after landing and if there's a noticeable impact I will revert to look for bottlenecks

PR #9458, opened 2021-02-09
feat: add structured cloning to Deno.core

### LINKED ISSUE #3557: Implement proper structured cloning between Workers

inteon:
First step towards #3557, as discussed in #9323. Also includes some simple tests.


core/serialize_deserialize.rs
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  fn deserializer<'s>(
+    scope: &mut v8::HandleScope<'s>,
+    data: &[u8],
+    array_buffers: &'a mut ArrayBuffers,
+  ) -> v8::ValueDeserializer<'a, 's> {
+    v8::ValueDeserializer::new(scope, Box::new(Self { array_buffers }), data)
+  }
+}
```
bartlomieju, line 26: Seems like this struct is superfluous, why not use `v8::ValueSerializer` directly?

core/serialize_deserialize.rs
```diff
@@ -0,0 +1,94 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 1: I believe we can put contents of this file into `bindings.rs` there's not that much code and we try to be frugal with files.

core/serialize_deserialize.rs
```diff
@@ -0,0 +1,94 @@
...
+  buffer: &[u8],
+) -> Option<v8::Local<'s, v8::Value>> {
+  let mut array_buffers = ArrayBuffers::new(); // TODO(Inteon): fix array buffers
+  let mut value_deserializer =
+    SerializeDeserialize::deserializer(scope, &buffer, &mut array_buffers);
+  unsafe {
+    core::mem::transmute(
+      value_deserializer.read_value(scope.get_current_context()),
+    ) // TODO(Inteon): fix lifetime bug in rusty_v8?
+  }
```
bartlomieju, line 93: I'll try to get `rusty_v8` upgraded tonight

inteon: ok, fixed it

core/serialize_deserialize.rs
inteon: I just pushed a simplified version.

core/serialize_deserialize.rs
inteon: I just pushed a simplified version.

inteon:
Blocked on #9466

bartlomieju (approved):
LGTM, thanks @inteon, great work! 

Before landing I'll wait for Ben's review

core/bindings.rs
```diff
@@ -625,6 +627,90 @@ fn decode(
   };
 }
 
+struct SerializeDeserialize {}
+
+impl v8::ValueSerializerImpl for SerializeDeserialize {
+  #[allow(unused_variables)]
```
bnoordhuis, line 633: What unused variables?

bnoordhuis (approved):
LGTM, thanks!

core/bindings.rs
bartlomieju: This function is required by `v8::ValueSerializerImpl` but is never used in `core/bindings.rs` (yet I think)

ry:
Nice work @inteon - thanks!

PR #9462, opened 2021-02-09
fix(cli/source_map): Don't use file names from source maps

### LINKED ISSUE #6542: Stack trace contains wrong paths
I don't know if it is related but since today the fullpath is missing in the stacktrace in my github [CI](https://github.com/apiel/adka/runs/814693690?check_suite_focus=true):

image

_Originally posted by @apiel in https://github.com/denoland/deno/issues/5202#issuecomment-650634379_

Most likely introduced in #6428 with the change to use inline source maps.

### LINKED ISSUE #6965: Main thread panic at Error throw
I'm experiencing a strange bug with `Deno`.

When a exception is thrown the main process panics. This is the trace:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli\fmt_errors.rs:86:8
stack backtrace:
   0:     0x7ff7af88a45b - <unknown>
   1:     0x7ff7af53cecc - <unknown>
   2:     0x7ff7af889483 - <unknown>
   3:     0x7ff7af888bb8 - <unknown>
   4:     0x7ff7af887d2e - <unknown>
   5:     0x7ff7af88efe5 - <unknown>
   6:     0x7ff7af53a4b0 - <unknown>
   7:     0x7ff7af53a2cc - <unknown>
   8:     0x7ff7af36cbf2 - <unknown>
   9:     0x7ff7af48ada8 - <unknown>
  10:     0x7ff7af48b345 - <unknown>
  11:     0x7ff7af53cecc - <unknown>
  12:     0x7ff7af751885 - <unknown>
  13:     0x7ff7af349c1d - <unknown>
  14:     0x7ff7af4daf6f - <unknown>
  15:     0x7ff7af8aa3f6 - <unknown>
  16:     0x7ff7af4ea67e - <unknown>
  17:     0x7ff7b048bf10 - CrashForExceptionInNonABICompliantCodeRange
  18:     0x7ff8e9777bd4 - BaseThreadInitThunk
  19:     0x7ff8ea16ce51 - RtlUserThreadStart
```

I don't know how to reproduce the bug but it happened in my project so I tried to isolate it in a separated branch:
https://github.com/AndreJesusBrito/WebAssembly-Language-Compiler/tree/rust-exception-bug-isolation

The bug happens in file `src/SyntaxAnalysis/SyntaxRule.ts` at line 25 when I throw an exception.
```ts
...
public getDerivation(terminal: string): RuleDerivation {
  const derivation = this.derivations.get(terminal);
  if (derivation) {
    return {
      derivationSymbols: [...derivation.derivationSymbols].reverse(),
      actions: derivation.actions,
    };
  } else {
    throw SyntaxError("Unexpected token '" + terminal + "' at rule " + this._name);            // THIS IS LINE 25
  }
}

public setDerivation(rule: SyntaxSymbol[], actions: ActionObj[], ...terminals: string[]): void {
  for (const terminal of terminals) {
    if (this.derivations.get(terminal) instanceof Array) {
      console.warn("'" + terminal + "' repeated terminal for rule " + this._name);
...
```

Even though I'm using a SyntaxError (that is not a very good idea and will change later) it seems to happen for any error including the base Error class.

The most strange thing to me is that by changing the white space on this file it can work correctly.
For example, when I remove line 28 it works fine:
```
error: Uncaught SyntaxError: Unexpected token '' at rule none
  public setDerivation(rule: SyntaxSymbol[], actions: ActionObj[], ...terminals: string[]): void {
            ^
    at SyntaxRule.getDerivation (file:///C:/Users/Andr%C3%A9%20Brito/Documents/GitHub/WebAssembly-Language-Compiler/src/SyntaxAnalysis/SyntaxRule.ts:28:13)
    at file:///C:/Users/Andr%C3%A9%20Brito/Documents/GitHub/WebAssembly-Language-Compiler/src/main.ts:4:1
```

I'm running using the command:
```
deno run src\main.ts
```

and my specs are:
```
deno 1.2.2
v8 8.5.216
typescript 3.9.2
OS: Windows_NT x64 10.0.18362
```

Hope this helps! :)

### LINKED ISSUE #9443: Stack traces include angle characters around URL
Here's a stack trace I was seeing recently
```
error: Uncaught SyntaxError: Missing initializer in const declaration
declare global {
~~~~~~~
    at <anonymous> (<https://deno.land/std@0.66.0/node/process.ts>:99:1)
```
Why are there "<" and ">" characters around the URL?

nayeemrmn:
Fixes #6542.
Fixes #6965 (again).
Fixes #9443.

`deno run cli/tests/error_026_remote_import_error.ts`

Before:
```
error: Uncaught Error: bad
    throw Error("bad");
        ^
    at foo (error_001.ts:2:9)
    at bar (error_001.ts:6:3)
    at error_001.ts:9:1
```

After:
```
error: Uncaught Error: bad
  throw Error("bad");
        ^
    at foo (http://localhost:4545/cli/tests/error_001.ts:2:9)
    at bar (http://localhost:4545/cli/tests/error_001.ts:6:3)
    at http://localhost:4545/cli/tests/error_001.ts:9:1
```

### DIFF
--- a/cli/source_maps.rs
+++ b/cli/source_maps.rs
@@ -110,12 +110,8 @@ pub fn get_orig_position<G: SourceMapGetter>(
   getter: Arc<G>,
 ) -> (String, i64, i64, Option<String>) {
   let maybe_source_map = get_mappings(&file_name, mappings_map, getter.clone());
-  let default_pos = (
-    file_name.clone(),
-    line_number,
-    column_number,
-    source_line.clone(),
-  );
+  let default_pos =
+    (file_name.clone(), line_number, column_number, source_line);
 
   // Lookup expects 0-based line and column numbers, but ours are 1-based.
   let line_number = line_number - 1;
--- a/cli/source_maps.rs
+++ b/cli/source_maps.rs
@@ -3,6 +3,7 @@
 //! This mod provides functions to remap a `JsError` based on a source map.
 
 use deno_core::error::JsError;
+use deno_core::ModuleSpecifier;
 use sourcemap::SourceMap;
 use std::collections::HashMap;
 use std::str;
@@ -109,8 +110,12 @@ pub fn get_orig_position<G: SourceMapGetter>(
   getter: Arc<G>,
 ) -> (String, i64, i64, Option<String>) {
   let maybe_source_map = get_mappings(&file_name, mappings_map, getter.clone());
-  let default_pos =
-    (file_name, line_number, column_number, source_line.clone());
+  let default_pos = (
+    file_name.clone(),
+    line_number,
+    column_number,
+    source_line.clone(),
+  );
 
   // Lookup expects 0-based line and column numbers, but ours are 1-based.
   let line_number = line_number - 1;
@@ -123,7 +128,16 @@ pub fn get_orig_position<G: SourceMapGetter>(
         None => default_pos,
         Some(token) => match token.get_source() {
           None => default_pos,
-          Some(original) => {
+          Some(source_file_name) => {
+            // The `source_file_name` written by tsc in the source map is
+            // sometimes only the basename of the URL, or has unwanted `<`/`>`
+            // around it. Use the `file_name` we get from V8 if
+            // `source_file_name` does not parse as a URL.
+            let file_name = match ModuleSpecifier::resolve_url(source_file_name)
+            {
+              Ok(m) => m.to_string(),
+              Err(_) => file_name,
+            };
             let maybe_source_line =
               if let Some(source_view) = token.get_source_view() {
                 source_view.get_line(token.get_src_line())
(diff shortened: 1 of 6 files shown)


cli/source_maps.rs
```diff
@@ -134,7 +138,7 @@ pub fn get_orig_position<G: SourceMapGetter>(
             let source_line = if let Some(source_line) = maybe_source_line {
               Some(source_line.to_string())
             } else if let Some(source_line) = getter.get_source_line(
-              original,
+              &file_name,
```
nayeemrmn, line 141: The `original` source name written by tsc in the source map object is sometimes reduced to just the base name (#6542). I don't know when exactly this occurs but errors in remote imports are a reliable reproduction.

Passing that deformed specifier to the source map getter makes it fail and fall back to using the transpiled source line which doesn't align with the line and column numbers (#6965). I believe this is also responsible for adding the `<`/`>` characters (#9443).

nayeemrmn: > The unmapped `file_name` comes directly from V8. This actually doesn't need to be mapped, because the module specifiers we give to V8 are already the original user-written ones.

Oh no, this isn't the case for userspace source maps... I think we shouldn't render a source line in this case.

caspervonb: Just a heads up!

Came across a little caveat when trying to use this in coverage mapping; token.get_source was populated with an empty string so ended up doing the same as here and going with the known filename instead.

ry: @nayeemrmn That would be a great comment to have here in the source code.

ry: LGTM otherwise - thank you for the fix!

kitsonk (approved):
LGTM

Issue #9379, opened 2021-02-02
test: bad assumption in cli/tests/unit/fetch_test.ts


bnoordhuis:
https://github.com/denoland/deno/blob/9690ce5e872f1873538b8cea7c93fa3ea8bfd9d1/cli/tests/unit/fetch_test.ts#L23-L34

The test assumes it won't be able to connect to port localhost:4000. Surprise! On my system there's something listening there.

Not sure how to best fix this but picking a port number that isn't an obvious Schelling point is probably a good first step. :-)

DavidDeSimone:
My initial thought for this was to check if the port is available via a call to listen - if you can an AddrInUse error, just try the next port up. Repeat that process until you find an available port and then fetch on that port.

The only issue I see with that approach is that there is a small timing window between calling close on your listening connection, and calling fetch. In that time, another process could snag that port, and you would see this issue. However, that timing window is really small and specific, and overall I think that approach is better then trying a specific port that may or may not have something already listening on it.

bnoordhuis:
That's already a step up from the status quo. Do you want to send a PR?

DavidDeSimone:
I'll submit a PR for this later today.

PR #9465, opened 2021-02-10
fix(tests) - fix fetchConnectionError test if port 4000 is in use 

DavidDeSimone:
Fixes https://github.com/denoland/deno/issues/9379

As outlined in that issue, test `fetchConnectionError` assumes that nothing is listening on port 4000. Some users may have something listening on port 4000. To improve test reliability, we check if port 4000 is available by attempting to listen on it. If its not available, we will try the next port up until we find one that is available, or until we hit the maximum port we are willing to try.

### DIFF
--- a/cli/tests/unit/fetch_test.ts
+++ b/cli/tests/unit/fetch_test.ts
@@ -5,6 +5,7 @@ import {
   assertThrows,
   assertThrowsAsync,
   fail,
+  unimplemented,
   unitTest,
 } from "./test_util.ts";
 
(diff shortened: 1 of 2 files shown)


cli/tests/unit/fetch_test.ts
```diff
@@ -23,12 +23,32 @@ unitTest({ perms: { net: true } }, async function fetchProtocolError(): Promise<
...
+            await fetch(`http://localhost:${port}`);
+          },
+          TypeError,
+          "error trying to connect",
+        );
+        portFound = true;
+        break;
+      } catch (e) {
+        port++;
+      }
```
bnoordhuis, line 46: IMO, the port iteration logic kind of obscures that the purpose of the test is to run that fetch() command. I'd break it out into a helper function.

Another thing: the try/catch section is too big, it should only be around the `Deno.listen()` call.

DavidDeSimone: Thanks for the feedback. I've factored the referenced code into a helper function, and reduced the amount of logic in the try/catch. I think that it is now overall much more readable.

crypt096 (approved):

bnoordhuis (approved):
Thanks!

PR #9471, opened 2021-02-11
feat(lsp): add deno cache code actions

kitsonk:
This PR also fixes a performance regression when accessing cached modules by creating an inner lock on the `Sources` struct.

It also instruments the tsc ops (which helped isolate the performance regression).

@bnoordhuis I tried to figure any easy way to not introduce an arc mutex for sources, but I really can't see a way.  We can't pass a clone of the arc mutex of the language server for the "snapshot" and I also tried just arc mutex'ing parts of sources, but I quickly ended up in deadlocks, it seems that only locking an inner in sources is the only way to avoid it.

Trying to chase down all the places where we might modify/hydrate the data in sources is really difficult, because we might become "aware" of needing to look in the Deno cache at several different points, and actually it is far more efficient to lazily hydrate that cache at the point where the module needs to be accessed.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -65,6 +65,7 @@ pub struct LanguageServer(Arc<tokio::sync::Mutex<Inner>>);
 pub struct StateSnapshot {
   pub assets: HashMap<ModuleSpecifier, Option<AssetDocument>>,
   pub documents: DocumentCache,
+  pub performance: Performance,
   pub sources: Sources,
 }
 
@@ -148,6 +149,14 @@ impl Inner {
         if dep.maybe_type.is_none() {
           if let Some(ResolvedDependency::Resolved(resolved)) = &dep.maybe_code
           {
+            let maybe_type = self.sources.get_maybe_types(resolved);
+            if let Some(ResolvedDependency::Resolved(maybe_type_specifier)) =
+              &maybe_type
+            {
+              // this hydrates the sources for the `maybe_types` specifier, as it
+              // might not have been retrieved yet
+              self.sources.contains(maybe_type_specifier);
+            }
             dep.maybe_type = self.sources.get_maybe_types(resolved);
           }
         }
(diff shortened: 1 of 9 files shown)


cli/lsp/language_server.rs
```diff
@@ -923,53 +928,65 @@ impl Inner {
...
-          })?;
-        if code_actions.is_fix_all_action(
-          &action,
-          diagnostic,
-          &file_diagnostics,
-        ) {
-          code_actions.add_ts_fix_all_action(&action, &specifier, diagnostic);
+      match diagnostic.source.as_deref() {
+        Some("deno-ts") => {
+          let code = match &diagnostic.code.clone().unwrap() {
```
bnoordhuis, line 940: You don't need to clone here, I think? Just as_ref()?

cli/lsp/sources.rs
```diff
@@ -248,28 +309,26 @@ impl Sources {
...
-      if let Ok(metadata) = fs::metadata(path) {
-        if let Ok(modified) = metadata.modified() {
-          return if let Ok(n) = modified.duration_since(SystemTime::UNIX_EPOCH)
-          {
-            Some(format!("{}", n.as_millis()))
-          } else {
-            Some("1".to_string())
-          };
+    let path = self.get_path(specifier)?;
+    if let Ok(metadata) = fs::metadata(path) {
```
bnoordhuis, line 317: Just noting that this is doing synchronous file I/O while holding a lock. Probably not a big deal with local filesystems but network filesystems are a different story.

Aside: one more opportunity for simplifying the unboxing logic:
```suggestion
    let metadata = fs::metadata(path).ok()?;
```

bnoordhuis (approved):
LGTM. I was looking at it yesterday and I didn't really see an easy way to do it without a mutex either.

cli/lsp/sources.rs
kitsonk: Thanks for the feedback.  I discussed it with @ry a while back and the biggest challenge is that doing it async will bubble up to the tsc ops, and all of the tsc ops are synchronous because of the way tsc works, and we know it breaks.  I think we can "hide" some of the asynchronous nature on the Rust side from tsc, but we didn't want to tackle that until we had a performance problem.

The first thing I think we would do is try to limit the number of version checks during a given operation, because right now, it could be checked several times in a row, though it is highly unlikely to have changed.

PR #9472, opened 2021-02-11
Fix deno_std dead link

rigwild:
Hi, noticed a dead link when browsing the manual.


kt3k:
@rigwild Thank you for the fix! Can you run the code formatter (`./tools/format.js`)?

ry (approved):
Thanks!

PR #9476, opened 2021-02-12
fix(lsp): properly handle static assets

kitsonk:
This fixes an issue when using the lsp with a tsconfig with references libraries that are not present in the tsc snapshot, but instead static assets built in Rust.

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -95,9 +95,10 @@ pub struct AssetDocument {
 
 impl AssetDocument {
   pub fn new<T: AsRef<str>>(text: T) -> Self {
+    let text = text.as_ref();
     Self {
-      text: text.as_ref().to_string(),
-      length: text.as_ref().encode_utf16().count(),
+      text: text.to_string(),
+      length: text.encode_utf16().count(),
       line_index: LineIndex::new(text),
     }
   }
@@ -1077,7 +1078,7 @@ fn get_length(state: &mut State, args: Value) -> Result<Value, AnyError> {
   let specifier = ModuleSpecifier::resolve_url(&v.specifier)?;
   if let Some(Some(asset)) = state.state_snapshot.assets.get(&specifier) {
     Ok(json!(asset.length))
-  } else if state.state_snapshot.documents.contains(&specifier) {
+  } else if state.state_snapshot.documents.contains_key(&specifier) {
     cache_snapshot(state, v.specifier.clone(), v.version.clone())?;
     let content = state
       .snapshots
@@ -1108,7 +1109,7 @@ fn get_text(state: &mut State, args: Value) -> Result<Value, AnyError> {
   let content =
     if let Some(Some(content)) = state.state_snapshot.assets.get(&specifier) {
       content.text.clone()
-    } else if state.state_snapshot.documents.contains(&specifier) {
+    } else if state.state_snapshot.documents.contains_key(&specifier) {
       cache_snapshot(state, v.specifier.clone(), v.version.clone())?;
       state
         .snapshots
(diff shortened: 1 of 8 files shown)


cli/tsc.rs
```diff
@@ -44,32 +44,38 @@ pub fn compiler_snapshot() -> Snapshot {
...
+    ("lib.es2018.full.d.ts", inc!("lib.es2018.full.d.ts")),
+    ("lib.es2019.full.d.ts", inc!("lib.es2019.full.d.ts")),
+    ("lib.es2020.full.d.ts", inc!("lib.es2020.full.d.ts")),
+    ("lib.esnext.full.d.ts", inc!("lib.esnext.full.d.ts")),
+    ("lib.scripthost.d.ts", inc!("lib.scripthost.d.ts")),
+    ("lib.webworker.d.ts", inc!("lib.webworker.d.ts")),
+    ("lib.webworker.importscripts.d.ts", inc!("lib.webworker.importscripts.d.ts")),
+    ("lib.webworker.iterable.d.ts", inc!("lib.webworker.iterable.d.ts")),
+  ]
+  .iter()
```
lucacasonato, line 71: Can you use `into_iter` here to avoid the `.copied()`?

cli/lsp/tsc.rs
```diff
@@ -89,46 +89,78 @@ impl TsServer {
...
+    let assets = tsc::STATIC_ASSETS
+      .iter()
+      .map(|(k, v)| {
+        let url_str = format!("asset:///{}", k);
+        let specifier = ModuleSpecifier::resolve_url(&url_str).unwrap();
+        let asset = AssetDocument::new(v);
+        (specifier, Some(asset))
+      })
+      .collect();
+    Self(assets)
```
lucacasonato, line 120: This will load all libs into memory, regardless of if they are used. Can probably be optimized in the future to avoid some memory overhead.

lucacasonato (approved):
Just one comment. LGTM

cli/tsc.rs
kitsonk: That doesn't work, because `.into_iter()` returns a `FromIterator<&(&str, &str)>` which `HashMap<&str, &str>` can't be built from.  The only way to build a static `HashMap` like this is to do `.copied()`.

cli/lsp/tsc.rs
kitsonk: They only get allocated if the lsp gets used.  The alternative is to re-introduce some sort of `inner` for the assets and lazily calculate them when they are requested from tsc.

cli/lsp/tsc.rs
```diff
@@ -1071,17 +1105,20 @@ fn get_text(state: &mut State, args: Value) -> Result<Value, AnyError> {
...
-      .unwrap()
-      .clone()
-  } else {
-    let sources = &mut state.state_snapshot.sources;
-    sources.get_text(&specifier).unwrap()
-  };
+  let content =
+    if let Some(Some(content)) = state.state_snapshot.assets.get(&specifier) {
+      content.text.clone()
+    } else if state.state_snapshot.documents.contains(&specifier) {
```
bnoordhuis, line 1111: Suggestion: rename `Documents::contains()` to `Documents::contains_key()` for consistency.

(Also makes it clearer that it's a key search, not a string search, which I initially thought it was.)

bnoordhuis (approved):

cli/tsc.rs
lucacasonato: 👍 - then feel free to merge

bnoordhuis: I think you should be able to do `&[ /* .. */ ].into_iter().cloned().collect()` here?

cli/lsp/tsc.rs
kitsonk: why, I might was well do `Sources::contains()` to `Sources::contains_key()`... 🥳

cli/lsp/tsc.rs
```diff
@@ -89,46 +89,78 @@ impl TsServer {
...
+  pub length: usize,
   pub line_index: LineIndex,
 }
 
+impl AssetDocument {
+  pub fn new<T: AsRef<str>>(text: T) -> Self {
+    Self {
+      text: text.as_ref().to_string(),
+      length: text.as_ref().encode_utf16().count(),
+      line_index: LineIndex::new(text),
```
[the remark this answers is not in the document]
kitsonk: 👍

cli/tsc.rs
kitsonk: `(&[ /* .. */ ]).into_iter().cloned().collect()` works, if remediated the other two static hashmaps we've got.

PR #9497, opened 2021-02-14
refactor(core): Reduce verbosity of binding core functions

AaronO:
Removes boilerplate in `bindings::initialize_context()` when setting core functions, improving readability.

`initialize_context()` is shortened from 98 to 36 lines, making the core logic easier to grasp.


AaronO:
Slightly longer after `rustfmt` expanded the `setMacrotaskCallback` line, but point still stands.

On that note, should 80 still be the max-width, I know the linux kernel recently relaxed their max-width to 100, might be worth considering (but it's minor at the end of the day, no point bikeshedding)

bartlomieju (approved):
LGTM, thanks @AaronO

PR #9498, opened 2021-02-14
fix(op_crates/fetch): type check Request init

CrimsonCodes0:
Type checks Request, Response, and fetch initializer parameters.

If the init is null or undefined, it would be replaced with an object, if it is a function, nothing is done, anything else results in a type error being thrown.

Fixes <https://github.com/denoland/deno/issues/8565>.

If the Request/Response construction fails, the error is propagated to the callee, as defined in the spec.

Currently, this gives an *interesting* error in Deno:
```js
fetch( new String("https://google.com") );
```
The Request constructor doesn't have this problem, ~~so I'll make fetch construct it first.~~
~~Also de-duplicates code, since `fetch`and `Request`, did a lot of the the same things.~~


op_crates/fetch/26_fetch.js
```diff
@@ -958,12 +1001,12 @@
     // @ts-expect-error because the use of super in this constructor is valid.
     constructor(input, init) {
       if (arguments.length < 1) {
-        throw TypeError("Not enough arguments");
+        throw new TypeError(
+          `Failed to construct "Request," requires at least 1 argument, received ${arguments.length}`,
+        );
```
00ff0000red, line 1006: The old error here was absolutely terrible.

A specialization of one of the bootstrap utility `requiredArguments` for constructors would be nice.

op_crates/fetch/26_fetch.js
```diff
@@ -1027,22 +1070,19 @@
...
-        throw TypeError(BodyUsedError);
+        throw new TypeError(BodyUsedError);
       }
 
       const iterators = this.headers.entries();
-      const headersList = [];
-      for (const header of iterators) {
-        headersList.push(header);
-      }
+      const headersList = [...iterators];
```
00ff0000red, line 1085: I just passed through, making minor tweaks, but I'm not 100% sure if using the iterable like this was safe.

op_crates/fetch/26_fetch.js
lucacasonato: These utils are already available in window.__bootstrap.webidl.

00ff0000red: Ah, I see, updating to make use of them.

00ff0000red: Actually, seems like window.__bootstrap.webidl doesn't include anything for a WebIDL dictionary, or I'm not seeing it.

If there isn't, I can add the cast there.

00ff0000red: @lucacasonato Does Deno have conversions for any of the following: HeadersInit, BodyInit, ReferrerPolicy, RequestMode, RequestCredentials, RequestCache, RequestRedirect, or AbortSignal?

[RequestInit IDL definition](https://fetch.spec.whatwg.org/#requestinit)

Most of these definately aren't WebIDL primitives.

Currently, I've only got the following:
```js
const requestInitConverter = createDictionaryConverter("RequestInit", [
  {
    converter: webidl.converter.ByteString,
    key: "method",
  },
  {
    converter: (v) => new Headers(v),
    key: "headers",
  },
  {
    converter: webidl.converter.USVString,
    key: "referrer",
  },
  {
    converter: webidl.converter.DOMString,
    key: "integrity",
  },
  {
    converter: webidl.converter.boolean,
    key: "keepalive",
  },
]);
```

op_crates/web/00_webidl.js
```diff
@@ -662,10 +662,40 @@
...
+      if (Enum.has(V)) {
+        return V;
+      } else if (hasDefault) {
+        return defaultValue;
+      } else {
+        throw makeException(
+          TypeError,
+          `The provided value '${V}' is not a valid enum value of type ${name}.`,
+        );
+      }
```
lucacasonato, line 690: This logic doesn't follow spec: https://heycam.github.io/webidl/#es-enumeration.

The default should only be used if V is undefined or null.

op_crates/web/00_webidl.js
```diff
@@ -662,10 +662,40 @@
...
+
+    let defaultValue;
+
+    const hasDefault = "defaultValue" in opts;
+
+    if (hasDefault) {
+      defaultValue = opts.defaultValue;
+    }
+
+    const { converter } = ops;
```
lucacasonato, line 676: Where is ops defined?

op_crates/web/00_webidl.js
lucacasonato: But really this shouldn't happen in this function at all (see spec), but instead there should be a separate wrapper for converters that make them nullable with a default value.

lucacasonato: See https://heycam.github.io/webidl/#idl-nullable-type

op_crates/web/00_webidl.js
00ff0000red: Oh, `opts`, missed a letter; I'll review the WebIDL spec you linked earlier and rewrite this anyway.

op_crates/web/00_webidl.js
00ff0000red: Oh, I didn't catch myself there with the default value, that is *bad*. Passing in an invalid value would instead become a default value, meaning the last else in the chain is unreachable.

You can tell that this was some ad-hoc code I wrote, I needed to review the spec later to get a proper implementation later; for one, I wasn't sure what string types were used in enums, and am suprised to see that the spec asks implementations to work in ECMAScript strings instead of WebIDL strings.

lucacasonato:
@00ff0000red Could you rename `webidl.createEnumeration` to `webidl.createEnumConverter`, and open a second PR with just that addition? I would like to land that separately right away because it will unblock some other PRs.

00ff0000red:
@lucacasonato Sure thing, would we want a nullable/defaultable converter too?

op_crates/fetch/26_fetch.js
```diff
@@ -941,6 +949,122 @@
...
+      key: "integrity",
+    },
+    {
+      converter: webidl.converter.boolean,
+      key: "keepalive",
+    },
+    {
+      // AbortSignal?
+      converter: createNullableConverter(abortSignalConverter),
+      key: "signal",
```
00ff0000red, line 1050: It seems like some of these are unused in Deno's implementation, ex: `signal`.
I'll leave implementing them to another PR, as I'd only like to make `Request`, `Response`, and `fetch` use proper type conversions.

I'm starting to think that I might not even attempt to de-dup `fetch` and use `Request` in it's implementation, would anyone consider that out of scope for this change?

As for type checking / converting some of these, I'm not sure if Deno even implements `AbortSignal`, does it?

00ff0000red: On second thought, passing anything that isn't implemented should throw a "not implemented" error.

lucacasonato: No any extra priorities that are but used should be silently ignored.

00ff0000red: > ...that are but used...

It wasn't clear what you were intending to say, did you miss a word?

lucacasonato: Im sorry, phone keyboard. Here we go again:

Extra properties that are not implemented by us at the moment should be silently ignored.

00ff0000red: Okay, got it.

op_crates/fetch/26_fetch.js
```diff
@@ -1297,82 +1416,74 @@
    * @returns {Promise<Response>}
    */
   async function fetch(input, init) {
```
00ff0000red: Oddly enough, Chrome says `fetch.length === 0`, even though it also has the length === 1 assertion.

WebIDL:
```
[NewObject] Promise<Response> fetch(RequestInfo input, optional RequestInit init = {});
```

Can someone check Firefox and see what theirs says? If they also say 0, then I need to review the spec and figure out how this is possible.

lucacasonato: FF:
```
> fetch.length
1
```

Chrome is wrong.

00ff0000red: Thanks! I'll go open an bug report with them.

00ff0000red: Ignore this, I was totally off, Chrome says `1`, the site (not GitHub) corrupted their own `fetch`. Annoying.

00ff0000red:
All that's left to do is implement `requestInfoConverter` and `bodyInitConverter`.

@lucacasonato `RequestInfo` is a union type, do you think I should attempt to add a webidl.union in a seperate PR?

lucacasonato:
Unions are by by far the most complex IDL type. I am working on an implementation for them as part of #7977. I am hoping to have the union type converter factory in `main` in a couple of days.

00ff0000red:
Aknowledged, I expected that there would be a good deal of complexity in their implementaton, I actually wasn't sure if it would be possible to even make a general union type, as opposed to a specialized implementation for every individual union.

I'll do what I can in the meanwhile.

cli/tests/unit/request_test.ts
```diff
@@ -62,3 +62,48 @@ unitTest(async function cloneRequestBodyStream(): Promise<void> {
 
   assertEquals(b1, b2);
 });
+
+unitTest(function requiresOneArgument() {
```
00ff0000red, line 66: I think we're missing a test case for when a Request has been passed to the Request constructor, is there anything else missing?

op_crates/fetch/26_fetch.js
00ff0000red: Reverted these small changes.

CrimsonCodes0:
@lucacasonato Was a union converter ever created?

If not, I'll see if JSDOM has any helpers for it.

lucacasonato:
@CrimsonCodes0 No, not a generic one. I did add a few hand written union converters to the webgpu.js file. You will need to implement the union converters manually according to the relevant webgpu spec. It is incredibly difficult to write a generic union converter.

CrimsonCodes0:
Back when I had last replied, I attempted to make the union converter, but couldn't wrap my head around the WebIDL spec on union types. I believe that if the last two converters are implemented, then Deno's fetch would act as expected.

Do you think that you could implement RequestInfo's type for this PR?

CrimsonCodes0:
@lucacasonato Sorry for the second ping, I'm not sure if you had received the last one? Could you inform me of whether or not you could look at my last request? Thank you for your time.

CrimsonCodes0:
Hmm... `fetchResponseConstructorInvalidStatus` is failing with the following:
> ```
> The status provided (null) is outside the range [200, 599]
> ``` 

besides that, I think that in any of the WebIDL converters where I left a comment saying `// default behavior`, the behavior isn't spec compliant, but the tests are still passing. I'm not sure what the semantics are supposed to be for those areas.

CrimsonCodes0:
I'm not sure if the message thrown by `new Response({ status: null })` should be preserved or not.

Currently, it claims that null is outside of the range [200, 599], while Chromium says (0) isn't in the range. Firefox doesn't even mention the input value. Chromium converts the status from a JavaScript value to a u16 before continuing with the spec algorithm implementation.

Because I eagerly convert values now, the value is 0, not null.

I updated the unit test that checked this to no longer check the error message at all. If anyone wanted the old behavior, I can revert it, I guess...

lucacasonato:
@CrimsonCodes0 Thanks for the effort on this PR. I am currently in the process of completely rewriting `fetch` / `Request` / `Response` from scratch (#10203), because it was determined that the entire implementation is so spec in-compliant that it would take longer to realign, than to rewrite from scratch. As an example: internally we don't represent headers the same way the spec does. Because of this we have to deviate from spec in many places when working with headers. This is not just the case with headers, but also with the internal representation of requests, responses, redirect handling, and much more. I am sorry this PR has been open and boiling for so long, but I think many of the issues addressed here will be resolved in #10203. As a result of that I would urge you to hold work on this PR until that has landed, and then we can reevalute what items from this PR are still needed.

CrimsonCodes0:
@lucacasonato Really? I had quite the opposite thinking, I believe that it should be easier for you to refactor the internals, knowing that all types passed to the internal code will be receiving the correct types, at least.

But if you disagree, I doubt that this PR will be of much use, considering that a correct implementation of the fetch spec expects all input to be normalised and casted to WebIDL types. Maybe we'd be better off if you just copy & pasted (most of) the converters over, and we close this PR?

CrimsonCodes0:
@lucacasonato
Barely checked on the rewrite PR, left two late comments.

The only thing that I see missing would be that `fetch` requires one argument, Deno gives this right now:
> error: Uncaught TypeError: Invalid URL
> await fetch();

I'll just close this PR and open a new one instead.

PR #9500, opened 2021-02-14
fix(lsp): handle  cached type dependencies properly

kitsonk:
@lucacasonato this fixes the issue you have encountered where transitive type dependencies are not handled properly, causing incorrect diagnostics until the affected dependency is opened in the editor.

I still need to work on a test, though it is actually quite complicated to recreate in the test harness.

### DIFF
--- a/cli/lsp/sources.rs
+++ b/cli/lsp/sources.rs
@@ -55,7 +55,7 @@ fn resolve_remote_specifier(
   http_cache: &HttpCache,
   redirect_limit: isize,
 ) -> Option<ModuleSpecifier> {
-  let cache_filename = http_cache.get_cache_filename(specifier.as_url());
+  let cache_filename = http_cache.get_cache_filename(specifier.as_url())?;
   if redirect_limit >= 0 && cache_filename.is_file() {
     let headers = get_remote_headers(&cache_filename)?;
     if let Some(location) = headers.get("location") {
@@ -303,7 +303,7 @@ impl Inner {
       (source, MediaType::from(specifier), None)
     } else {
       let cache_filename =
-        self.http_cache.get_cache_filename(specifier.as_url());
+        self.http_cache.get_cache_filename(specifier.as_url())?;
       let headers = get_remote_headers(&cache_filename)?;
       let maybe_content_type = headers.get("content-type").cloned();
       let (media_type, maybe_charset) =
@@ -334,7 +334,7 @@ impl Inner {
     } else if let Some(path) = self.remotes.get(&specifier) {
       Some(path.clone())
     } else {
-      let path = self.http_cache.get_cache_filename(&specifier.as_url());
+      let path = self.http_cache.get_cache_filename(&specifier.as_url())?;
       if path.is_file() {
         self.remotes.insert(specifier.clone(), path.clone());
         Some(path)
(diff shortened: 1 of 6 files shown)


kitsonk:
Actually figured a way to add a unit test which should be sufficient.  Also, I did some extensive overhaul `lsp/sources.rs` to make it more readable and consistent (and more efficient).

cli/lsp/sources.rs
```diff
@@ -40,16 +40,108 @@ pub async fn cache(
...
+  http_cache: &HttpCache,
+  redirect_limit: isize,
+) -> Option<ModuleSpecifier> {
+  let cache_filename = http_cache.get_cache_filename(specifier.as_url());
+  if redirect_limit >= 0 && cache_filename.is_file() {
+    let headers = get_remote_headers(&cache_filename)?;
+    if let Some(location) = headers.get("location") {
+      let redirect =
+        ModuleSpecifier::resolve_import(location, specifier.as_str()).ok()?;
+      resolve_remote_specifier(&redirect, http_cache, redirect_limit - 1)
```
bnoordhuis, line 64: What happens when my malicious server sends back a `Location: file:///etc/passwd` response? It'll follow the redirect, wouldn't it?

It looks like `cli/file_fetcher.rs` is susceptible to that, too.

cli/lsp/sources.rs
```diff
@@ -40,16 +40,108 @@ pub async fn cache(
...
+  let scheme = specifier.as_url().scheme();
+  if !SUPPORTED_SCHEMES.contains(&scheme) {
+    return None;
+  }
+
+  if scheme == "data" {
+    Some(specifier.clone())
+  } else if scheme == "file" {
+    let path = specifier.as_url().to_file_path().ok()?;
+    if path.is_file() {
```
bnoordhuis, line 87: This looks susceptible to TOCTOU bugs, the on-disk entity may change between check and use.

cli/lsp/sources.rs
kitsonk: Remote modules are banned from importing local modules. It should happen at another point of module resolution but I am not sure we test for it.

cli/lsp/sources.rs
kitsonk: Can you think of an easy way to mitigate it?

The scenario it would arise is pretty rare, in that this is a file on disk that would need to be modified in a few milliseconds by something other than the current editor and the impact is that the lsp might have a slightly out of date version of a document until some other change causes the disconnect to be rectified.

cli/lsp/sources.rs
kitsonk: Ok, I investigated this a bit.  It doesn't specifically fail in the way I expected it to fail, but it fails.  When it tries to follow it, and it does `http_cache.get_cache_filename(specifier.as_url())` it panics with an unimplemented in `http_cache::base_url_to_filename`, as it tries to lookup the cache for a local file.

I think `cli/file_fetcher.rs` would have a slightly different behaviour.  It would allow the module to be resolved, but in the module graph, it would eventually surface up a `GraphError::InvalidLocalImport` when that module is attempted to be loaded.

I'm going to look into how difficult it would be to not throw in `get_cache_filename` as it would be better for the lsp to simply resolve to a `None` instead of panicing.

kitsonk: Ok, I changed it so it doesn't panic, and trying to redirect effectively results in the module being `None`.

bnoordhuis (approved):

PR #9504, opened 2021-02-14
(00_webidl.js): Add internal enumConverter helper

00ff0000red:
Part of https://github.com/denoland/deno/pull/9498#issuecomment-778815752

There is currently no generic checks or conversion to/from a WebIDL string enumeration, this PR adds an internal helper for use in implementing builtins.

cc @lucacasonato

### DIFF
--- a/op_crates/web/00_webidl.js
+++ b/op_crates/web/00_webidl.js
@@ -686,6 +686,6 @@
     converters,
     requiredArguments,
     createDictionaryConverter,
-    createEnumConverter
+    createEnumConverter,
   };
 })(this);
--- a/op_crates/web/00_webidl.js
+++ b/op_crates/web/00_webidl.js
@@ -662,6 +662,7 @@
     };
   }
 
+  // https://heycam.github.io/webidl/#es-enumeration
   function createEnumConverter(name, ...values) {
     const E = new Set(values);
 
--- a/op_crates/web/00_webidl.js
+++ b/op_crates/web/00_webidl.js
@@ -669,22 +669,29 @@
     return function (V, opts = {}) {
       const S = String(V);
 
-      if (E.has(V)) {
-        return V;
-      } else {
+      if (!E.has(S)) {
         throw makeException(
           TypeError,
           `The provided value '${V}' is not a valid enum value of type ${name}.`,
           opts,
         );
       }
+
+      return S;
     };
   }
 
-  function createNullableEnumConverter(name, ...values) {
-    const converter = createEnumConverter(name, ...values);
+  function createNullableConverter(converter) {
+    return (V, opts = {}) => {
+      // FIXME: If Type(V) is not Object, and the conversion to an IDL value is
+      // being performed due to V being assigned to an attribute whose type is a
+      // nullable callback function that is annotated with
+      // [LegacyTreatNonObjectAsNull], then return the IDL nullable type T?
+      // value null.
 
-    return (V, opts = {}) => null === V ? null : converter(V, opts);
+      if (V === null || V === undefined) return null;
+      return converter(V, opts);
+    };
   }
 
   window.__bootstrap ??= {};
@@ -693,5 +700,6 @@
     requiredArguments,
     createDictionaryConverter,
     createEnumConverter,
+    createNullableConverter,
   };
 })(this);
--- a/op_crates/web/00_webidl.js
+++ b/op_crates/web/00_webidl.js
@@ -669,16 +669,22 @@
     return function (V, opts = {}) {
       const S = String(V);
 
-      if (false === E.has(V)) {
+      if (E.has(V)) {
+        return V;
+      } else {
         throw makeException(
           TypeError,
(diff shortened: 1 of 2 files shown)


crypt096 (approved):

op_crates/web/00_webidl.js
```diff
@@ -662,10 +662,30 @@
...
 
-  window.__bootstrap = window.__bootstrap || {};
+  // https://heycam.github.io/webidl/#es-enumeration
+  function createEnumConverter(name, ...values) {
+    const E = new Set(values);
+
+    return function (V, opts = {}) {
+      const S = String(V);
+
+      if (false === E.has(V)) {
```
lucacasonato, line 672:
```suggestion
      if (!E.has(V)) {
```

lucacasonato (approved):
LGTM, just one comment

op_crates/web/00_webidl.js
00ff0000red: I'm surprised; imho, it's much more explicit as is, but I'll change.

lucacasonato: I read `if (!E.has(V)` as `if not E contains V`. The other one I would read as `if false is equal to E contains V`. First one seems simpler.

00ff0000red: Can you take another look now?

op_crates/web/00_webidl.js
```diff
@@ -662,10 +662,36 @@
...
+        );
+      }
+    };
+  }
+
+  function createNullableEnumConverter(name, ...values) {
+    const converter = createEnumConverter(name, ...values);
+
+    return (V, opts = {}) => null === V ? null : converter(V, opts);
+  }
```
lucacasonato, line 688: You added this in your last commit.

op_crates/web/00_webidl.js
lucacasonato: I don't really like early returns for the `default` case. I feel you should early return for the invalid case, and then return the `default` case at the end of the function. I don't feel strongly enough about it to ask you to change it again, so lets keep it like it is now.

op_crates/web/00_webidl.js
00ff0000red: Yes, would you rather that I hadn't?

I had implemented you had suggested at https://github.com/denoland/deno/pull/9498#discussion_r575796738,
> but instead there should be a separate wrapper for converters that make them nullable with a default value.

As the default value should be handled on in the callee, the nullable converter should only allow passing null types.

lucacasonato: There issues:

1 Doesn't follow spec, as null or undefined are both a valid nullable type.
2. Function is not exported from __bootstrap.webidl.
3. I don't see why this is constrained to enums, instead of working for arbitrary converters.

00ff0000red: > IDL nullable type values are represented by values of either the ECMAScript type corresponding to the inner IDL type, or the ECMAScript null value.

I interpreted this as meaning that `undefined` is not a valid value of `T?`, if `T` does not include `undefined`.

...and now I see:
> Otherwise, if V is null or undefined, then return the IDL nullable type T? value null.

Then how do we differentiate between defaultable types? `arguments.length`?

```
undefined foo(short x = 2);
```

Would this be,
```js
function foo(x) {
    if ( arguments.length > 1 ) {
        x = 2;
    } else {
        x = converters.short(x);
    }
}
```

2. I forgot to export it.

I thought about 3 a bit after making it and was going to suggest the same.

Maybe,
```js
function nullableConverter(converter) {
    return function (V) {
        const typeV = type(V);
        if ( "Null" === typeV || "Undefined" === typeV ) {
            return null;
        } else {
            return converter(V);
        }
}
```
What's your opinion on this?

lucacasonato: There are two places default values can show up: in function arguments, and in dictionary members. Both of these are specified to use the default value if the value is `undefined`. For function arguments this should be done using the `function(myOptArg = "defaultValue")` syntax, and for dictionaries this is handled using the `createDictionaryConverter` handler.

If a argument is nullable and has a default value (`T? = default`), `undefined` means default value, `null` means `null`, and any other value should be converted to `T`.

If a argument is nullable and has no default value (`T?`), `undefined` and `null` will result in `null`, and any other value will be converted to `T`.

I have pushed changes to `createNullableConverter` to address this.

00ff0000red: The only problem I'd see with using `function(myOptArg = "defaultValue")`, is that it would change the function's length, e.g. `foo.name === "foo"`, `foo.length === 0`; would that break compliance with any ES spec or WebIDL APIs?

lucacasonato: Function length should be the length of the shortest allowed argument list, so this is correct. That is specified in https://heycam.github.io/webidl/#interface-object and https://heycam.github.io/webidl/#dfn-create-operation-function.

lucacasonato (approved):
@00ff0000red Thanks for bearing through all my change requests. Lets land it.

PR #9505, opened 2021-02-14
fix(core): Make dynamic import async errors catchable

### LINKED ISSUE #6259: Errors in dynamic imports are not properly caught

nayeemrmn:
Fixes #6259.

### DIFF
--- a/cli/tests/delayed_error.ts
+++ b/cli/tests/delayed_error.ts
@@ -1,0 +1,2 @@
+await new Promise((r) => setTimeout(r, 100));
+throw new Error("foo");
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -822,6 +822,12 @@ impl JsRuntime {
         );
         let promise = v8::Local::<v8::Promise>::try_from(value)
           .expect("Expected to get promise as module evaluation result");
+        let empty_fn = |_scope: &mut v8::HandleScope,
+                        _args: v8::FunctionCallbackArguments,
+                        _rv: v8::ReturnValue| {};
+        let empty_fn = v8::FunctionTemplate::new(scope, empty_fn);
+        let empty_fn = empty_fn.get_function(scope).unwrap();
+        promise.catch(scope, empty_fn);
         let promise_global = v8::Global::new(scope, promise);
         let mut state = state_rc.borrow_mut();
         state.pending_promise_exceptions.remove(&promise_global);
--- a/cli/tests/085_dynamic_import_async_error.ts
+++ b/cli/tests/085_dynamic_import_async_error.ts
@@ -1,0 +1,5 @@
+try {
+  await import("./delayed_error.ts");
+} catch (error) {
+  console.log(`Caught: ${error.stack}`);
+}
--- a/cli/tests/085_dynamic_import_async_error.ts.out
+++ b/cli/tests/085_dynamic_import_async_error.ts.out
@@ -1,0 +1,2 @@
+[WILDCARD]Caught: Error: foo
+    at [WILDCARD]/delayed_error.ts:[WILDCARD]
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -2779,6 +2779,11 @@ console.log("finish");
     output: "084_worker_custom_inspect.ts.out",
   });
 
+  itest!(_085_dynamic_import_async_error {
+    args: "run --allow-read 085_dynamic_import_async_error.ts",
+    output: "085_dynamic_import_async_error.ts.out",
+  });
+
   itest!(js_import_detect {
     args: "run --quiet --reload js_import_detect.ts",
     output: "js_import_detect.ts.out",


core/runtime.rs
```diff
@@ -104,6 +105,7 @@ pub(crate) struct JsRuntimeState {
   pub(crate) js_macrotask_cb: Option<v8::Global<v8::Function>>,
   pub(crate) pending_promise_exceptions:
     HashMap<v8::Global<v8::Promise>, v8::Global<v8::Value>>,
+  pub(crate) cancelled_promise_exceptions: HashSet<v8::Global<v8::Promise>>,
```
bartlomieju, line 108: Please add a comment for these two fields describing distinction between them

cli/tests/delayed_error.ts
```diff
@@ -0,0 +1,2 @@
+await new Promise((r) => setTimeout(r, 100));
```
bartlomieju, line 1: Can this line be replaced with `await Promise.resolve();`? If so maybe you could rewrite it as a unit test in `core/runtime.rs`?

bartlomieju (commented):
Great fix Nayeem!

cli/tests/delayed_error.ts
nayeemrmn: It must be delayed by a macrotask at least.

core/runtime.rs
```diff
@@ -811,7 +816,10 @@ impl JsRuntime {
           .expect("Expected to get promise as module evaluation result");
         let promise_global = v8::Global::new(scope, promise);
         let mut state = state_rc.borrow_mut();
-        state.pending_promise_exceptions.remove(&promise_global);
+        let removed = state.pending_promise_exceptions.remove(&promise_global);
+        if removed.is_none() {
+          state.cancelled_promise_exceptions.insert(promise_global);
```
nayeemrmn, line 821: Actually I think this is leaky, dynamic imports which don't eventually reject are also added to and never removed from this set. @bartlomieju Any ideas?

nayeemrmn: Though it's not that big of a deal, as data associated with that module is also permanently kept in the module graph I think.

bartlomieju: @nayeemrmn can you remove the promise handle from this hash map once dynamic import resolves?

nayeemrmn: > @nayeemrmn can you remove the promise handle from this hash map once dynamic import resolves?

No, as far as the runtime infrastructure is concerned, the module is considered evaluated and done with once the pending promise is returned. Then it's just a promise in JS land. We pretty much need to do the equivalent of `.then()`ing it in rust land at this site.

bartlomieju: > > @nayeemrmn can you remove the promise handle from this hash map once dynamic import resolves?
> 
> No, as far as the runtime infrastructure is concerned, the module is considered evaluated and done with once the pending promise is returned. Then it's just a promise in JS land. We pretty much need to do the equivalent of `.then()`ing it in rust land at this site.

Would https://docs.rs/rusty_v8/0.19.0/rusty_v8/struct.Promise.html#method.then work?

nayeemrmn: Discussed on discord: https://discord.com/channels/684898665143206084/713035341736706089/810923927618060308.

nayeemrmn:
@bartlomieju I've changed the approach to attach a JS object property to the promise to indicate that it shouldn't result in an exception.

nayeemrmn:
Simplified further... we just had to add an empty `.catch()` handler to the promise 🤦‍♂️

kitsonk (approved):
LGTM

ry:
Nice work @nayeemrmn - thanks!

kitsonk:
Yes, meant to thank you in the approval.

core/runtime.rs
```diff
@@ -822,6 +822,12 @@ impl JsRuntime {
...
           .expect("Expected to get promise as module evaluation result");
+        let empty_fn = |_scope: &mut v8::HandleScope,
+                        _args: v8::FunctionCallbackArguments,
+                        _rv: v8::ReturnValue| {};
+        let empty_fn = v8::FunctionTemplate::new(scope, empty_fn);
+        let empty_fn = empty_fn.get_function(scope).unwrap();
+        promise.catch(scope, empty_fn);
         let promise_global = v8::Global::new(scope, promise);
         let mut state = state_rc.borrow_mut();
         state.pending_promise_exceptions.remove(&promise_global);
```
bartlomieju, line 833: Out of curiosity, does adding this handler still makes it so that promise ends up in `state.pending_promise_exceptions`? If not we could cut some complexity thanks to this fix

nayeemrmn: It still gets put in there if there is an immediate rejection (the cases that worked correctly before this patch), the promise reject callback gets called before the handler is added. Actually... I think adding the handler here triggers https://github.com/denoland/deno/blob/555595e6d02d1fc1adb535df074d258389fab509/core/bindings.rs#L252-L254 so maybe the remove below is redundant. Will investigate.

PR #9508, opened 2021-02-14
feat(cli): support auth tokens for accessing private modules

### LINKED ISSUE #5239: Support access token authorisation for remote files
Currently there is no way to fetch remote modules from private GitHub repositories without exposing your access token.

For example you could do the following currently:

```ts
import * as foo from "https://$TOKEN@raw.githubusercontent.com/private_org/private_repo/master/foo.ts";
```

But you could easily "leak" your access token if you then checked that code in and pushed it to a public repo.

GitHub (and I assume other services) all the token to be passed as an authorisation header:

```ts
Authorization: token $TOKEN
```

If this was somehow passed on the command line, it would become easier to secure.

kitsonk:
Closes #5239

Adds the ability to set the environment variable `DENO_AUTH_TOKENS` to bearer tokens to allow Deno to access private modules/repositories.


cli/file_fetcher.rs
```diff
@@ -323,8 +327,9 @@ impl FileFetcher {
     ca_data: Option<Vec<u8>>,
   ) -> Result<Self, AnyError> {
     Ok(Self {
+      auth_tokens: AuthTokens::new(env::var(DENO_AUTH_TOKENS).ok()),
       allow_remote,
-      cache: FileCache::default(),
+      cache: Default::default(),
```
lucacasonato, line 332: I think @bnoordhuis would disagree with this change.

cli/http_util.rs
```diff
@@ -14,6 +16,56 @@ use deno_runtime::deno_fetch::reqwest::Client;
...
+#[derive(Debug, Clone)]
+pub struct AuthTokens(Vec<AuthToken>);
+
+impl AuthTokens {
+  pub fn new(maybe_token_str: Option<String>) -> Self {
+    let mut tokens = Vec::new();
+    if let Some(tokens_str) = maybe_token_str {
+      for token_str in tokens_str.split(';') {
+        if token_str.contains('@') {
+          let pair: Vec<&str> = token_str.split('@').collect();
```
lucacasonato, line 34: This means that the auth token could currently never include an `@` symbol. I think this is a reasonable constraint, but it could be fixed by only splitting on the last `@` as hostnames may never contain an `@` symbol.

lucacasonato (commented):
This looks great! I do think the auth tokens should be included in the file name hash of the cached file. Modules that were successfully fetched with an auth token would still be loaded from cache after the token was removed otherwise.

kitsonk:
> Modules that were successfully fetched with an auth token would still be loaded from cache after the token was removed otherwise.

I tend to disagree with this.  Once a file is in cache, it is in cache.  The web server can change, the module removed, etc... we never attempt to recache it unless requested.  That would apply here.  Tokens are not tied to the resources, they are tied to an authenticated user.  Multiple tokens could access that same resource.  It shouldn't impact the hash name.  It also makes it really difficult to cache a bunch of resources in a secure environment to build a cache and then run that in an environment where you don't want expose your tokens.

cli/http_util.rs
kitsonk: I will change it to `rsplitn`, safer that way

lucacasonato (approved):
> It also makes it really difficult to cache a bunch of resources in a secure environment to build a cache and then run that in an environment where you don't want expose your tokens.

Didn't think about this. Totally agree with that reasoning.

LGTM. Very cleanly implemented.

bartlomieju (approved):
LGTM

docs/linking_to_external_code/private.md
```diff
@@ -0,0 +1,75 @@
...
+a _private_ repository, like a private repository on GitHub.
+
+Deno supports sending bearer tokens when requesting a remote module. Bearer
+tokens are the predominate type of access token used with OAuth 2.0 and is
+broadly supported by hosting services (e.g. GitHub, Gitlab, BitBucket,
+Cloudsmith, etc.).
+
+### DENO_AUTH_TOKENS
+
+The Deno CLI will look for an environment variable named `DENO_AUTH_TOKENS` to
```
ry, line 13: What about just "DENO_AUTH" ?

docs/linking_to_external_code/private.md
```diff
@@ -0,0 +1,75 @@
...
+When the token is incorrect, or the user does not have access to the module,
+GitHub will issue a `404 Not Found` status, instead of an unauthorized status.
+So if you are getting errors that the modules you are trying to access are not
+found on the command line, check the environment variable settings and the
+personal access token settings.
+
+In addition, `deno run -L debug` should print out a debug message about the
+number of tokens that are parsed out of the environment variable. It will print
+an error message if it feels any of the tokens are malformed. It won't print any
+details about the tokens for security purposes.
```
ry, line 75: Lovely docs. Nice work.

docs/linking_to_external_code/private.md
lucacasonato: I think we might want to add `DENO_AUTH_CERTS` in the future too for mTLS, so I think we should stick with `DENO_AUTH_TOKENS`.

ry:
Please add an entry to the CLI help env var list https://github.com/denoland/deno/blob/1afe6b48e0bfee44f37206eb0fc03ea6ae37ad4d/cli/flags.rs#L222-L233

cli/http_util.rs
```diff
@@ -84,16 +132,23 @@ pub enum FetchOnceResult {
 pub async fn fetch_once(
   client: Client,
   url: &Url,
-  cached_etag: Option<String>,
+  maybe_etag: Option<String>,
+  maybe_auth_token: Option<String>,
```
ry, line 136: Maybe not for now, but this method is getting lots of args, we might want to put them into a struct.

Why doesn't the `maybe_auth_token` argument not take `Option<AuthToken>`?

cli/http_util.rs
```diff
@@ -14,6 +16,52 @@ use deno_runtime::deno_fetch::reqwest::Client;
...
+        url.host_str()?.to_string()
+      };
+      if hostname.to_lowercase().ends_with(&t.host) {
+        Some(t.token.clone())
+      } else {
+        None
+      }
+    })
+  }
+}
```
ry, line 63: consider moving these AuthToken and AuthTokens (and their associated tests) to a new file cli/auth_tokens.rs

cli/http_util.rs
kitsonk: Because it isn't the structure that holds all the auth tokens.  I can create a type alias of `AuthToken` for `String` but it seems a bit trivial.

cli/http_util.rs
kitsonk: 👍

cli/http_util.rs
kitsonk: Actually I see what I've done... ok... let me modify it so we pass `AuthToken` and set the display or something to generate the header value.

docs/linking_to_external_code/private.md
kitsonk: Irrespective of future environment variables, `DENO_AUTH` could be quite ambiguous.  `DENO_AUTH_TOKENS` is explicit that this is some form of authentication tokens, not login credentials, etc.  I think it is better to be a bit more explicit in this case.

cli/http_util.rs
kitsonk: I have also moved to a struct for the args, otherwise we would likely never do it.

ry (approved):
LGTM - nice work!

PR #9511, opened 2021-02-15
docs: Use JSONC code block for commenting

TETRA2000:
I found that some code blocks on docs are rendered as invalid syntax because of commenting to JSON.
For example [this one](https://github.com/denoland/deno/blob/master/docs/getting_started/setup_your_environment.md#emacs).

I marked those blocks as ~JSON5~ JSONC for better syntax highlighting.

### DIFF
--- a/docs/getting_started/setup_your_environment.md
+++ b/docs/getting_started/setup_your_environment.md
@@ -167,7 +167,7 @@ page, first `npm install --save-dev typescript-deno-plugin typescript` in your
 project (`npm init -y` as necessary), then add the following block to your
 `tsconfig.json` and you are off to the races!
 
-```json
+```jsonc
 {
   "compilerOptions": {
     "plugins": [
@@ -238,7 +238,7 @@ endif
   syntax highlighting
 - Add the following `.sublime-project` file to your project folder
 
-```json
+```jsonc
 {
   "settings": {
     "LSP": {
--- a/docs/linking_to_external_code/import_maps.md
+++ b/docs/linking_to_external_code/import_maps.md
@@ -42,9 +42,9 @@ $ deno run --import-map=import_map.json --unstable color.ts
 
 To use starting directory for absolute imports:
 
-```json
-// import_map.json
+**import_map.json**
 
+```jsonc
 {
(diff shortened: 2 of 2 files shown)


docs/linking_to_external_code/import_maps.md
```diff
@@ -42,7 +42,7 @@ $ deno run --import-map=import_map.json --unstable color.ts
 
 To use starting directory for absolute imports:
 
-```json
+```json5
 // import_map.json
```
kitsonk, line 46: There is no good reason that there is a comment in this block, and should be moved outside instead.

docs/linking_to_external_code/import_maps.md
```diff
@@ -60,7 +60,7 @@ import { MyUtil } from "/util.ts";
 
 You may map a different directory: (eg. src)
 
-```json
+```json5
```
kitsonk, line 63: There is no good reason that there is a comment in this block, and should be moved outside instead.

docs/getting_started/setup_your_environment.md
```diff
@@ -167,7 +167,7 @@ page, first `npm install --save-dev typescript-deno-plugin typescript` in your
 project (`npm init -y` as necessary), then add the following block to your
 `tsconfig.json` and you are off to the races!
 
-```json
+```json5
```
kitsonk, line 170: Better to use `jsonc` as we actually support that in Deno.

kitsonk (changes requested):
Thanks, but if we could use `jsonc` instead of `json5`.

docs/linking_to_external_code/import_maps.md
TETRA2000: I see. I moved and made them bold like previous code blocks.

TETRA2000:
@kitsonk Thank you for your review!

> Better to use jsonc as we actually support that in Deno.

I didn't know about JSONC but it's supported on GitHub and your point is pretty reasonable.
I fixed all.

Could you review again?
Thank you.

kitsonk (approved):
LGTM

Issue #9510, opened 2021-02-15
Support loading import maps from URL


bartlomieju:
With recent discussion to make import maps stable (#4958, #9461) we should add support for loading import map from URL.

PR #9519, opened 2021-02-16
feat: support loading import map from URL

bartlomieju:
This commit adds support for loading import maps from URLs,
both remote and local.

This feature is supported in CLI flag as well as in runtime
compiler API.

Closes https://github.com/denoland/deno/issues/9510

### DIFF
--- a/cli/program_state.rs
+++ b/cli/program_state.rs
@@ -56,7 +56,7 @@ pub struct ProgramState {
 }
 
 impl ProgramState {
-  pub fn new(flags: flags::Flags) -> Result<Arc<Self>, AnyError> {
+  pub async fn new(flags: flags::Flags) -> Result<Arc<Self>, AnyError> {
     let custom_root = env::var("DENO_DIR").map(String::into).ok();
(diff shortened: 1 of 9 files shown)


cli/program_state.rs
```diff
@@ -56,7 +56,7 @@ pub struct ProgramState {
 }
 
 impl ProgramState {
-  pub fn new(flags: flags::Flags) -> Result<Arc<Self>, AnyError> {
+  pub async fn new(flags: flags::Flags) -> Result<Arc<Self>, AnyError> {
```
lucacasonato, line 59: Maybe `ProgramState::new` should be renamed to `ProgramState::build` because it is now async?

lucacasonato (commented):
Very nice! Awesome that it was so easy to implement. Just one minor comment.

lucacasonato (approved):

flying-sheep:
> Awesome that it was so easy to implement

Even better: This actually reduced the amount of code we have!

PR #9522, opened 2021-02-17
fix(lsp): handle data URLs properly

### LINKED ISSUE #9514: data URLs not well handled in lsp
Data URLs are only partially handled in the LSP at the moment.  They first appear as "not found" until they are cached.  This feels like the right behaviour, though the code action and the diagnostic message could be more informative.

The problem is that the source is unavailable, and cannot be opened in the editor as a virtual text document.  This means that it is impossible to do goto definition and other "challenges" with using data URLs in the editor.

kitsonk:
Fixes #9514


cli/lsp/urls.rs
```diff
@@ -0,0 +1,186 @@
...
+
+  fn put(&mut self, specifier: ModuleSpecifier, url: Url) {
+    let specifier = Arc::new(specifier);
+    let url = Arc::new(url);
+    self.specifier_to_url.insert(specifier.clone(), url.clone());
+    self.url_to_specifier.insert(url, specifier);
+  }
+
+  fn get_url(&self, specifier: &ModuleSpecifier) -> Option<&Url> {
+    self.specifier_to_url.get(specifier).map(Deref::deref)
```
bnoordhuis, line 71:
```suggestion
    self.specifier_to_url.get(specifier).as_deref()
```
(That's what you're doing here in long-hand, right? Ditto line 75.)

cli/lsp/analysis.rs
```diff
@@ -456,11 +456,15 @@ impl CodeActionCollection {
   ) -> Result<(), AnyError> {
     if let Some(data) = diagnostic.data.clone() {
       let fix_data: DenoFixData = serde_json::from_value(data)?;
+      let title = if diagnostic.code
+        == Some(lsp::NumberOrString::String("no-cache-data".to_string()))
```
bnoordhuis, line 460: If you want to avoid such `to_string()` calls, you can use `matches!`:
```suggestion
      let title = if matches!(diagnostic.code, Some(lsp::NumberOrString::String(code)) if code == "no-cache-data")
```
(Or just use `match`, of course.)

It's not that a single allocation is terrible but it's often death by a thousand cuts.

bnoordhuis (approved):

cli/lsp/urls.rs
kitsonk: That is what I get for copying some old code off stackoverflow.

kitsonk: oooh:

```
expected enum `std::option::Option<&deno_core::url::Url>`
found enum `std::option::Option<&Arc<deno_core::url::Url>>`
```

So you gotta unwrap it/map it to get rid of the ref like that.  So it isn't as crazy as it seems.

cli/lsp/urls.rs
```diff
@@ -0,0 +1,182 @@
...
+  media_type
+}
+
+/// A bi-directional map of URLs sent to the LSP client and internal module
+/// specifiers.  We need to map internal specifiers into `deno:` schema URLs
+/// to allow the Deno language server to manage these as virtual documents.
+#[derive(Debug, Default)]
+pub struct LspUrlMap {
+  specifier_to_url: HashMap<Arc<ModuleSpecifier>, Arc<Url>>,
+  url_to_specifier: HashMap<Arc<Url>, Arc<ModuleSpecifier>>,
```
ry, line 53: I don't think it makes sense to have Arc here. These are small bits of data that can just be cloned.

cli/lsp/urls.rs
```diff
@@ -0,0 +1,182 @@
...
+    assert_eq!(actual, MediaType::TypeScript);
+
+    let fixture = resolve_url("data:application/javascript;base64,ZXhwb3J0IGNvbnN0IGEgPSAiYSI7CgpleHBvcnQgZW51bSBBIHsKICBBLAogIEIsCiAgQywKfQo=").unwrap();
+    let actual = data_url_media_type(&fixture);
+    assert_eq!(actual, MediaType::JavaScript);
+
+    let fixture = resolve_url("data:text/plain;base64,ZXhwb3J0IGNvbnN0IGEgPSAiYSI7CgpleHBvcnQgZW51bSBBIHsKICBBLAogIEIsCiAgQywKfQo=").unwrap();
+    let actual = data_url_media_type(&fixture);
+    assert_eq!(actual, MediaType::Unknown);
+  }
```
ry, line 152: 👍

cli/lsp/urls.rs
ry: I had this open from the merge, so I did this small change too.

cli/lsp/urls.rs
```diff
@@ -59,18 +57,16 @@ impl LspUrlMap {
   }
 
   fn put(&mut self, specifier: ModuleSpecifier, url: Url) {
-    let specifier = Arc::new(specifier);
```
kitsonk: Doing this doubles up on the memory of a full allocation of both maps of URLs.  It isn't specifically important, but the `Arc` only allocated the thread safe pointer.

Issue #9444, opened 2021-02-08
lsp: goto underlines in vscode are incorrect when hovering while holding down ctrl key 


lucacasonato:
When using the new extension with the LSP, I found some weird alignment issues with the `goto` underlines when hovering over something in VS Code while holding the `Ctrl` key. Unsure if this is our fault or if it is a VS Code bug.

https://user-images.githubusercontent.com/7829205/107208733-47ae0d80-6a02-11eb-9fff-8012d9349a39.mp4

PR #9525, opened 2021-02-17
fix(lsp): document spans use original range

kitsonk:
Fixes: #9444


cli/lsp/language_server.rs
```diff
@@ -1575,30 +1576,27 @@ impl Inner {
...
+    let result = if let Some(implementations) = maybe_implementations {
+      let mut links = Vec::new();
+      for implementation in implementations {
+        if let Some(link) = implementation.to_link(&line_index, self).await {
+          links.push(link)
         }
       }
-      self.performance.measure(mark);
-      Ok(Some(GotoDefinitionResponse::Link(results)))
+      info!("links: {:#?}", json!(links));
```
lucacasonato, line 1592: Shouldn't these be `debug!`?

lucacasonato (approved):
LGTM, just one comment

cli/lsp/language_server.rs
kitsonk: Nom they shouldn't be there at all... they were just while I was isolating the issue... 😊

PR #9526, opened 2021-02-17
feat: stabilise import maps

### LINKED ISSUE #4958: discussion: Make import maps stable
Based on a discussion in Discord, @shian15810 suggests that marking import maps unstable is going to be problematic (as well as brings it up in #4933 but it should have its own issue).

I am torn, the spec only has one implementor (Chromium 74+) and Firefox and Safari as well as other runtimes have no public signals to support it.  We implemented it because it solved a specific problem that we didn't want to solve in a non-standard way.  The spec repo though indicates there are a few critical risks/issues with the current spec: https://github.com/WICG/import-maps#further-work.  None of those really are worrisome, except that support for how multiple maps are dealt with.  At this point we don't support that feature (because the spec doesn't) and the semantics are likely to be breaking to the current single file support, in my opinion.

Given it sits unflagged in Chromium is an indication that they do not expect there to be breaking changes with it as the standard evolves (though they have been wrong about that before).

As @shain15810 points out, it is effectively opt-in at the moment anyways, so having two flags to be able to enable a single flag does seem non-sensical.  If anything, we could warn when using the flag, that the spec isn't finalised and semantics may be broken in the future.

bartlomieju:
This commit stabilises import maps by removing requirement
on "--unstable" flag when "--import-map" flag is used.

Blocked by #9616
Blocked by #9519 
Closes #4958


cli/flags.rs
```diff
@@ -1586,8 +1586,7 @@ fn import_map_arg<'a, 'b>() -> Arg<'a, 'b> {
     .long("import-map")
     .alias("importmap")
     .value_name("FILE")
-    .requires("unstable")
-    .help("UNSTABLE: Load import map file")
+    .help("Load import map file")
     .long_help(
       "UNSTABLE:
```
caspervonb, line 1591: Remove the long help? or at-least remove the UNSTABLE bit.

docs/linking_to_external_code/import_maps.md
```diff
@@ -1,8 +1,5 @@
 ## Import maps
 
-> This is an unstable feature. Learn more about
-> [unstable features](../runtime/stability.md).
-
 Deno supports [import maps](https://github.com/WICG/import-maps).
 
 You can use import maps with the `--import-map=<FILE>` CLI flag.
```
bartlomieju, line 5: Note to self: this document still needs to be updated before landing

ry (approved):
LGTM

PR #9527, opened 2021-02-17
fix(cli): emits error if no target files are found

### LINKED ISSUE #9518: deno lint doesn't work with URLs
```
> deno lint --unstable https://deno.land/x/oak@v6.4.2/router.ts
Checked 0 file
```
At minimum it should error out...

magurotuna:
This PR makes fmt and lint emit "No target files found." error if no target files are found. (This happens if the passed path is URL, for example.)
Fixes #9518


caspervonb:
Hmmm I don't think we should change the behaviour of collect_files but instead if should be up to the individual commands?

The test command for example has an option to allow no test files to be run and I am making use of collect_files in my experimental test-runner.

I think the call site `tools::lint::lint_files` would be a good place to error out for example.

magurotuna:
@caspervonb Okay that makes sense. I'll revert `collect_files` and move checking for validity of target paths to `lint_files` and fmt where `collect_files` is called.

ry (approved):
Thank you - LGTM

caspervonb (approved):
LGTM

Issue #9528, opened 2021-02-17
Deno.exit should throw when called from a test


caspervonb:
Originally mentioned in https://github.com/denoland/deno/issues/9213

Go recently introduced this behaviour

> When using go test, a test that calls os.Exit(0) during execution of a test function will now be considered to fail. This will help catch cases in which a test calls code that calls os.Exit(0) and thereby stops running all future tests. If a TestMain function calls os.Exit(0) that is still considered to be a passing test.

Which makes a lot of sense to me, calling exit will always mess up your test run and we should have a sanitizer to catch it.

I will be working on this.

hayd:
It'd be great to configure exit behavior e.g. in deno-lambda an exit should throw an exception (to be caught) rather than end the process (which fatally times out the lambda).

wperron:
@caspervonb do you think we can fix #9213 at the same time?

PR #9529, opened 2021-02-17
feat(runtime): add exit sanitizer to Deno.test

caspervonb:
This adds an exit sanitizer to ensure that code being tested or dependencies of that code can't accidentally call `Deno.exit` leading to partial test runs and false results.

Closes https://github.com/denoland/deno/issues/9528


cli/dts/lib.deno.ns.d.ts
```diff
@@ -108,6 +108,9 @@ declare namespace Deno {
      * after the test has exactly the same contents as before the test. Defaults
      * to true. */
     sanitizeResources?: boolean;
+
+    /** Ensure the test case does not accidentally prematurely exit. Defaults to true. */
+    sanitizeExit?: boolean;
```
kt3k, line 113: Do we need this option? I can't think of the case when this is set to false.

kt3k (commented):
I like the idea. I left one question.

cli/dts/lib.deno.ns.d.ts
caspervonb: Probably don't _really_ need it but I'd like to leave an escape hatch for the sake of completeness and just in case there's some scenario we haven't thought about.

Also serves as documentation, that this is sanitizer we have enabled by default.

bartlomieju: @caspervonb it'd be great to explicitly spell out in the docstring that it affects calls to `Deno.exit`

runtime/js/30_os.js
```diff
@@ -24,13 +24,24 @@
     return core.jsonOpSync("op_system_cpu_info");
   }
 
+  let exitHandler = null;
+  function setExitHandler(fn) {
+    exitHandler = fn;
+  }
```
bartlomieju, line 30: Please add a comment explaining the purpose of this handler along with info that it's not a public API and can be used only by internal test runner

bartlomieju (commented):
I'm in favor of this change, nothing worse than rogue exit statement that makes you think tests passed where in fact you might have run only a few test cases.

cli/dts/lib.deno.ns.d.ts
ry: Because this is changing the public API in the Deno namespace, I think we should be conservative here and leave it out. What about adding a TODO instead? If we need it we can come back to this PR and pull the diff.

ry (commented):
I think this change makes sense but the behavior should be described in the docs somewhere in https://deno.land/manual/testing

cli/dts/lib.deno.ns.d.ts
caspervonb: What about:

```js
/** Asserts that the test case does not call `Deno.exit`. Defaults to true. */
```

docs/testing.md
```diff
@@ -92,6 +92,32 @@ Deno.test({
 });
 ```
 
+### Exit sanitizer
+
+There's also the exit sanitizer which ensures that your code can't accidentally
+call Deno.exit by accident via your own code or third-party code signaling a
+false test success.
```
bartlomieju, line 99: The exit sanitizer which ensures that tested code doesn't
call `Deno.exit()` signaling a false test success.

caspervonb: Yeah that does sound better, applied.

bartlomieju (approved):
LGTM, great feature @caspervonb 🚀

PR #9531, opened 2021-02-17
Make ModuleSpecifier a type alias, not wrapper struct

ry:


core/module_specifier.rs
```diff
@@ -51,174 +48,126 @@ impl fmt::Display for ModuleResolutionError {
...
-
-  /// Converts a string representing an absolute URL into a ModuleSpecifier.
-  pub fn resolve_url(
-    url_str: &str,
-  ) -> Result<ModuleSpecifier, ModuleResolutionError> {
-    Url::parse(url_str)
-      .map(ModuleSpecifier)
-      .map_err(ModuleResolutionError::InvalidUrl)
-  }
+pub type ModuleSpecifier = Url;
```
ry, line 52: This is the main change.

bartlomieju: If you don't want to have `ModuleSpecifier` being a wrapper for `Url` what's the point of having this alias at all?

core/module_specifier.rs
```diff
@@ -51,174 +48,126 @@ impl fmt::Display for ModuleResolutionError {
...
-    specifier: &str,
-  ) -> Result<ModuleSpecifier, ModuleResolutionError> {
-    if Self::specifier_has_uri_scheme(specifier) {
-      Self::resolve_url(specifier)
-    } else {
-      Self::resolve_path(specifier)
+    // 3. Return the result of applying the URL parser to specifier with base
+    //    URL as the base URL.
+    Err(ParseError::RelativeUrlWithoutBase) => {
+      let base = if base == "<unknown>" {
```
bartlomieju, line 85: Make `"<unknown>"` a static variable to represent this magic value?

bartlomieju (approved):
Just want to reiterate that I'm not a fan of this change

core/module_specifier.rs
ry: We can maybe remove it later, but one could argue the semantic difference is valuable.

PR #9532, opened 2021-02-18
feat(runtime/ops): Strongly typed deserialization of JSON ops

crowlKats:
Extension of #9423 for runtime

### DIFF
--- a/runtime/ops/mod.rs
+++ b/runtime/ops/mod.rs
@@ -28,7 +28,8 @@ use crate::metrics::metrics_op;
 use deno_core::error::AnyError;
 use deno_core::json_op_async;
 use deno_core::json_op_sync;
-use deno_core::serde_json::Value;
+use deno_core::serde::de::DeserializeOwned;
+use deno_core::serde::Serialize;
 use deno_core::BufVec;
 use deno_core::JsRuntime;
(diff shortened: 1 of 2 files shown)


runtime/ops/mod.rs
```diff
@@ -37,18 +38,24 @@ use std::cell::RefCell;
...
-  R: Future<Output = Result<Value, AnyError>> + 'static,
+pub fn reg_json_async<F, V, R, RV>(
+  rt: &mut JsRuntime,
+  name: &'static str,
+  op_fn: F,
+) where
+  F: Fn(Rc<RefCell<OpState>>, V, BufVec) -> R + 'static,
+  V: DeserializeOwned,
+  R: Future<Output = Result<RV, AnyError>> + 'static,
+  RV: Serialize,
```
bartlomieju, line 49: To be honest I find this signature way harder to comprehend...

lucacasonato: We are just aligning the signature to core here. This will let us get rid of a bunch of `serde_json::from_value` and `serde_json::to_value` in WebGPU op crate. Hopefully will let us reduce LOC count by a few hundred.

bartlomieju (approved):

PR #9538, opened 2021-02-18
Change release build flags to -O3 when favorable

evanwashere:
Continuation of #9249 that focuses on compiling only cpu bound packages with -O3

more packages to consider for -O3:
- hashing and crypto libraries used by WebCrypto
- rustls (and its deps) *note: could potentially blow size & compile time trough roof*

# some benchmarks
rand
fetch with compression
fetch compressed


evanwashere:
build time is practically unaffected

from ubuntu bench
| bin | [old](https://github.com/denoland/deno/runs/1924295083#step:31:4813) | [new](https://github.com/denoland/deno/runs/1927184112#step:31:4951) |
| ------------- | ------------- | -- |
| deno  | 66.353mb | 66.353mb |
| denort | 47.209mb  | 47.198mb |

ry (approved):
@evanwashere Thank you for patch and complete analysis. LGTM.

PR #9539, opened 2021-02-18
refactor: use Mutex for ErrorBuffer

ry:
RwLock should only be used in circumstances where it has some benefit.
Multiple concurrent readers is usually an undesirable design.


kitsonk (approved):
LGTM

For the record, for once, I wasn't the source of these RwLocks. 😁

PR #9550, opened 2021-02-19
feat(extensions): implement WebUSB API

### LINKED ISSUE #9523: Support WebUSB
We should support WebUSB.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/USB
Spec: https://wicg.github.io/webusb/

From the spec:

> The software delivery model of the web is a key enabler for educational applications because they can be quickly loaded on any computer without questions of platform compatibility or administrative credentials. Science classes are incorporating computerized measurement and data logging into their lessons. These tools require bundled software that may be difficult to install on managed computers as every new native application adds overhead to an already stressed IT department. Web-based hardware APIs allow support for these devices to be built directly into existing online course materials, providing a completely seamless experience.
>
> Students learning to code with one of the many microcontroller development kits can take advantage of online developer tools to write and upload their code. These tools already exist however they require a native component to interface between the browser and the hardware. These native extensions add a barrier to entry and may expose the user to security vulnerabilities in a way that that code running in the sandboxed web environment does not.

Having WebUSB in Deno would open up whole arena for education, micro controller integration, etc.

Node.js has a couple native modules that provide the bindings, and there are polyfills that layer on top of that to provide WebUSB.  Like we have done with other items, we could just expose WebUSB directly as part of the CLI.

littledivy:
Closes #9523 

Implements the WebUSB API.


littledivy:
Update: 
1. `navigator.usb.*` APIs are flagged behind `--unstable` and `--allow-usb`
2. Typescript typings adapted from [@types/w3c-web-usb](https://www.npmjs.com/package/@types/w3c-web-usb)
2. `getDevices()` works. (spec compliant `USBDevice` object) 
3. Ran both examples in [`webusb/arduino`](https://github.com/webusb/arduino) with an Arduino Leonardo. 
    a. OUT Transfers are working very well. Tested `controlTransferOut` and `transferOut`.
    b. IN transfers - device offers much more than the maxPacketSize :thinking:  (`babble`). Investigating the cause.

Example:
```typescript
// blink.ts
const devices = await navigator.usb.getDevices();
// Arduino Leonardo
let device = devices.find((p) => p.productId == 0x8036);
await device.open();
if (device.configuration === null) { device.selectConfiguration(1) }
await device.claimInterface(2);
await device.selectAlternateInterface(2, 0);
await device.controlTransferOut({
  "requestType": "class",
  "recipient": "interface",
  "request": 0x22,
  "value": 0x01,
  "index": 2,
});

// Turn on LED
await device.transferOut(4, new TextEncoder().encode('H'));
// delay...
// Turn off LED
await device.transferOut(4, new TextEncoder().encode('L'));
await device.controlTransferOut({
  "requestType": "class",
  "recipient": "interface",
  "request": 0x22,
  "value": 0x00,
  "index": 2,
});

await device.close();
```

<details>
<summary>
Arduino sketch
</summary>

```c
// blink.ino
#include <WebUSB.h>

WebUSB WebUSBSerial(1, "Deno");

#define Serial WebUSBSerial

const int ledPin = 13; // Built-in LED

void setup() {
  Serial.begin(9600);
  Serial.write("Sketch begins.\r\n> ");
  Serial.flush();
  pinMode(ledPin, OUTPUT);
}

void loop() {
    int byte = Serial.read();
    Serial.write(byte);
    if (byte == 'H') {
      Serial.write("\r\nTurning LED on.");
      digitalWrite(ledPin, HIGH);
    } else if (byte == 'L') {
      Serial.write("\r\nTurning LED off.");
      digitalWrite(ledPin, LOW);
    }
    Serial.write("\r\n> ");
    Serial.flush();
}
```

</details>

littledivy:
Messed up `git reset` Will open a new PR

PR #9552, opened 2021-02-19
refactor(lsp): overhaul tsc request handling

lucacasonato:
This PR changes how the LSP interacts with the LSP. It implements prioritaztion
of interactive TSC requests over diagnostics, and prevents diagnostic requests
being queued in the incoming mpsc.


lucacasonato:
This request has no benefits right now because the LSP handles all requests fully synchronously (https://github.com/denoland/deno/pull/9271). I was not aware that we lock up the entirety of the LSP for the entire duration of all incoming requests. This needs to be resolved first.

cli/lsp/diagnostics.rs
```diff
@@ -229,11 +229,12 @@ fn ts_json_to_diagnostics(
     .collect()
 }
 
+// None is returned if getting diagnostics was cancelled.
```
kitsonk, line 232: Maybe turn this into a proper docblock comment?  Otherwise it feels just like a stray comment.

cli/lsp/diagnostics.rs
```diff
@@ -229,11 +229,12 @@ fn ts_json_to_diagnostics(
...
 pub async fn generate_ts_diagnostics(
   state_snapshot: StateSnapshot,
   diagnostic_collection: DiagnosticCollection,
   ts_server: &tsc::TsServer,
-) -> Result<DiagnosticVec, AnyError> {
+) -> Result<Option<DiagnosticVec>, AnyError> {
```
kitsonk, line 237: Have you considered how this is handled when diagnostics are "cleared"?

kitsonk: I guess it only happens when they have been cancelled... right?

cli/lsp/tsc.rs
```diff
@@ -45,12 +50,30 @@ type Request = (
...
   pub fn new() -> Self {
-    let (tx, mut rx) = mpsc::unbounded_channel::<Request>();
+    let (requests_queue_tx, mut requests_queue_rx) =
+      mpsc::unbounded_channel::<Request>();
+    let (diagnostics_notifier_tx, mut diagnostics_notifier_rx) =
+      mpsc::channel::<()>(1);
+    let diagnostics_request: Arc<Mutex<Option<DiagnosticsRequest>>> =
+      Default::default();
+
+    let diagnostics_request_ = diagnostics_request.clone();
```
kitsonk, line 75:
```suggestion
    let moved_diagnostics_request = diagnostics_request.clone();
```

Or anything than a trailing `_` as I personally think it is just ambiguous what it represented. `diagnostics_request_clone` or whatever...

cli/lsp/tsc.rs
```diff
@@ -59,16 +82,80 @@ impl TsServer {
...
+        // don't want to queue up stale diagnostic requests, and we want to
+        // batch diagnostic requests as much as possible for performance
+        // reasons. This tries to do that.
+        let mut maybe_req: Option<Request> = None;
+        loop {
+          // First check if there is a regular request to handle.
+          if let Some((req, state_snapshot, tx)) = maybe_req.take() {
+            let res = request(&mut ts_runtime, state_snapshot, req);
+            if tx.send(res).is_err() {
+              warn!("Unable to send result to client.");
```
kitsonk, line 98:
```suggestion
              warn!("Internal Error: Unable to send tsc result to language server.");
```

kitsonk (approved):
A couple thoughts... if you rebase off main as well, we should get to see the benchmarks in the output of the CI.

cli/lsp/diagnostics.rs
lucacasonato: If they are cleared, a `Some` of an empty `DiagnosticVec` would be returned I think.

lucacasonato:
I want to hold this until #9557 is landed. This PR alone will have no impact on performance.

Issue #8954, opened 2021-01-01
 Test case is leaking async ops when using clearTimeout inside finally block after not catching error


KyleJune:
If you use finally statements to do test cleanup, you can end up with the following test falsely reporting leaking async ops. In the below tests, you wouldn't expect `sync2` and `async2` to be leaking ops since they are not really doing anything but they end up failing with that error. I found that if you add a delay after you clearTimeout, the following test `async5` will not falsely report leaking async ops.

```ts
import { delay } from "https://deno.land/std@0.80.0/async/delay.ts";

Deno.test("sync1", () => {
  let timer = -1;
  try {
    timer = setTimeout(() => null, 10000);
    throw new Error("fail");
  } finally {
    clearTimeout(timer);
  }
});

Deno.test("sync2", () => {});

Deno.test("sync3", () => {});

Deno.test("async1", async () => {
  let timer = -1;
  try {
    timer = setTimeout(() => null, 10000);
    await delay(0);
    throw new Error("fail");
  } finally {
    clearTimeout(timer);
  }
});

Deno.test("async2", async () => {
  await delay(0);
});

Deno.test("async3", async () => {
  await delay(0);
});

Deno.test("async4", async () => {
  let timer = -1;
  try {
    timer = setTimeout(() => null, 10000);
    await delay(0);
    throw new Error("fail");
  } finally {
    clearTimeout(timer);
    await delay(0);
  }
});

Deno.test("async5", async () => {
  await delay(0);
});

Deno.test("async6", async () => {
  await delay(0);
});
```

```sh
$ deno --version
deno 1.6.0 (release, x86_64-unknown-linux-gnu)
v8 8.8.278.2
typescript 4.1.2
$ deno test
running 9 tests
test sync1 ... FAILED (1ms)
test sync2 ... FAILED (2ms)
test sync3 ... ok (2ms)
test async1 ... FAILED (2ms)
test async2 ... FAILED (3ms)
test async3 ... ok (3ms)
test async4 ... FAILED (2ms)
test async5 ... ok (3ms)
test async6 ... ok (3ms)

failures:

sync1
Error: fail
    at example_test.ts:7:11
    at asyncOpSanitizer (deno:cli/rt/40_testing.js:36:13)
    at Object.resourceSanitizer [as fn] (deno:cli/rt/40_testing.js:70:13)
    at TestRunner.[Symbol.asyncIterator] (deno:cli/rt/40_testing.js:245:24)
[...]

PR #9553, opened 2021-02-20
fix(runtime/testing): false positive for timers when an error is thrown

caspervonb:
Fixes https://github.com/denoland/deno/issues/8954

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -2354,6 +2354,12 @@ mod integration {
     output: "deno_test.out",
   });
 
+  itest!(deno_test_finally_cleartimeout {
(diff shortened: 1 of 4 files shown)


bartlomieju (changes requested):
@caspervonb I just run provided test case on `main` branch and it results in exactly same output from `deno test`. I think this solution is preferable but we need another test case that actually demonstrates what's being fixed

bartlomieju (approved):
LGTM, thanks Casper, provided test case fails on `main`

caspervonb:
Oops I cleaned up that test a bit too much, until it wasn't testing anything.
Fixed up test fails all cases on 1.7.5
 
Screenshot 2021-02-22 at 12 10 19 AM

PR #9544, opened 2021-02-19
fix(core/runtime): Remove redundant cleaning for dynamic import rejections

nayeemrmn:

PR #9559, opened 2021-02-20
fix(core/runtime): Fix dynamic imports for already rejected modules

### LINKED ISSUE #6259: Errors in dynamic imports are not properly caught

nayeemrmn:
Fixes #6259 (I checked with the user-given repro this time).
Includes #9544.

### DIFF
--- a/cli/tests/086_dynamic_import_already_rejected.ts
+++ b/cli/tests/086_dynamic_import_already_rejected.ts
@@ -1,0 +1,11 @@
+try {
+  await import("./error_001.ts");
+} catch (error) {
+  console.log(`Caught: ${error.stack}`);
+}
+
+try {
+  await import("./error_001.ts");
+} catch (error) {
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
Nice fix Nayeem, thanks!

PR #9562, opened 2021-02-21
fix(core/bindings): Add stacks for dynamic import resolution errors

### LINKED ISSUE #8769: Uncaught dynamic import promise error should log the stack...
We should log the stack trace in uncaught promise errors.  Currently we get something like:

```
error: Uncaught (in promise) TypeError: Cannot resolve module "file:///a/bad.js".
```

When if the error is caught on the rejected promise and the stack logged manually, you get something like:

```
TypeError: Cannot resolve module "file:///a/bad.js" from "file:///a/b.js".
    at file:///a/b.js:1:0
```

nayeemrmn:
Fixes #8769.


bartlomieju (approved):
Very nice fix, thanks Nayeem!

PR #9564, opened 2021-02-21
fix(dts): update doc of Deno.formatDiagnostics

magurotuna:
`Deno.compile` appears in the documentation of `Deno.formatDiagnostics`, but it was replaced with `Deno.emit`.
https://github.com/denoland/deno/discussions/8800
https://github.com/denoland/deno/pull/8799
This PR updates the example so it will become accurate.


bartlomieju (approved):
LGTM, thanks

PR #9565, opened 2021-02-21
chore: Upgrade internal dprint plugins

dsherret:
Been a while since someone updated this, so thought I would submit a PR. Output is still looking good.


dsherret:
Requires changes from #9566, so going to close this one.

PR #9572, opened 2021-02-22
fix(lsp): diagnostics use own thread and debounce

kitsonk:


kitsonk:
We will incorporate some concepts here in future PRs.

kitsonk:
Based on a chat we had a core team yesterday, I am re-opening this PR as it might be a viable approach...

Benchmarking on my local machine demonstrates that it does the "big" benchmark in ~2800ms where as current main does it in ~7563ms and the combined #9552 and #9557 results in ~5049ms.  So this PR at the moment gives the biggest improvement in performance, but without changing the locking mechanisms.

kitsonk:
Looking at the Bench CI, it is producing a [3178ms](https://github.com/denoland/deno/pull/9572/checks?check_run_id=1992459589#step:31:452) benchmark, while main is producing around ~8600ms.

cli/lsp/diagnostics.rs
```diff
@@ -25,8 +35,299 @@ pub enum DiagnosticSource {
...
+                )
+                .await;
+              }
+              let maybe_request = next.await;
+              if !handle_request(maybe_request, &mut collection, &mut maybe_snapshot) {
+                break;
+              }
+            }
+            maybe_request = &mut next => {
+              if !handle_request(maybe_request, &mut collection, &mut maybe_snapshot) {
```
piscisaureus, line 292: I don't understand why handle_request is called here. 
Wouldn't you want to just loop around so that the debounce timer restarts?

cli/lsp/diagnostics.rs
```diff
@@ -25,8 +35,299 @@ pub enum DiagnosticSource {
...
+          tokio::pin!(debounce);
+
+          // "race" the next message off the rx queue or the debounce timer.
+          // if the next message comes off the queue, the next iteration of the
+          // loop will reset the debounce future.  When the debounce future
+          // occurs, the diagnostics will be updated based on only the most
+          // recent snapshot, thereby "abandoning" any interim states.
+          tokio::select! {
+            _ = &mut debounce => {
+              if let Some(snapshot) = maybe_snapshot {
```
piscisaureus, line 275: I wonder if it might be more efficient to change `maybe_snapshot` into a simple boolean `dirty` or something.
If you find the dirty flag set at this point, you `.lock().unwrap().snapshot()` the `LanguageServer` object right here.

This would ensure that `update_diagnostics()` always operates on the the latest state while avoiding making countless snapshots only to discard them later.

The only reason I can think of that would make this work less well is if the main thread locks the `LanguageServer` object for long stretches (because then the diagnostics thread would have to wait for the lock). BTW it is also possible that this works in our advantage; when the `LanguageServer` is locked by the main thread, presumably it is getting mutated so it might be *good* to wait for it to calm down before kicking off the typescript compiler.

cli/lsp/diagnostics.rs
kitsonk: Because `next` isn't just update diagnostic requests, it is also "getting" and "updating" requests, so this processes those, and we need to break on a `None` return, because it means the channel has closed.

cli/lsp/diagnostics.rs
kitsonk: Yeah, that reflects what you and I were talking about 1-2-1, so I will re-work it, it makes sense and it is cleaner, and requires a lot less snapshotting.  The locks should be really sort, and we can't really get into a dead-lock with this, so it is fine to wait a bit.

cli/lsp/diagnostics.rs
```diff
@@ -25,8 +37,317 @@ pub enum DiagnosticSource {
...
+          let debounce = time::sleep(duration);
+          tokio::pin!(debounce);
+
+          // "race" the next message off the rx queue or the debounce timer.
+          // if the next message comes off the queue, the next iteration of the
+          // loop will reset the debounce future.  When the debounce future
+          // occurs, the diagnostics will be updated based on only the most
+          // recent snapshot, thereby "abandoning" any interim states.
+          tokio::select! {
+            _ = &mut debounce => {
```
bnoordhuis, line 286: I might be mistaken but I think you can skip the pinning and just do:
```suggestion
            _ = time::sleep(duration) => {
```

bnoordhuis (approved):

cli/lsp/diagnostics.rs
```diff
@@ -3,20 +3,32 @@
...
 use std::collections::HashSet;
 use std::mem;
+use std::sync::Arc;
+use std::thread;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+use tokio::time;
+
+// 150ms between keystrokes is about 45 WPM.
+const DIAGNOSTIC_DEBOUNCE_MS: u64 = 200;
```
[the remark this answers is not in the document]
kitsonk: I guess I need to clarify, 150ms is fast typing, we want something longer than fast typing.  I will clarify the comment.

PR #9573, opened 2021-02-22
feat(runtime): stabilise permissions and add event target capabilities

kitsonk:
This stabilises permissions under the `Deno` namespace and adds back the event target features that were reverted prior to Deno 1.2.

Since Deno permission descriptors will never intersect with `navigator` permissions and the web standard has left `request` and `revoke` out of the APIs, but they are very useful for Deno use cases, it makes sense to align the APIs to the spirit of the web standard but keep them part of the Deno namespace, as they are the only place where they can be reliably used.

Stabilising the APIs provides the advantage of allow people to make interactive CLI applications (and use permission management libraries) in Deno without the need of the `--unstable` flag, which leads to a better experience when building tools for others.


cli/tests/unit/permissions_test.ts
```diff
@@ -19,8 +20,33 @@ unitTest(async function permissionNetInvalidHost(): Promise<void> {
...
+unitTest(async function permissionQueryReturnsEventTarget() {
+  const status = await Deno.permissions.query({ name: "hrtime" });
+  assert(["granted", "denied", "prompt"].includes(status.state));
+  let called = false;
+  status.addEventListener("change", () => {
+    called = true;
+  });
+  status.dispatchEvent(new Event("change"));
+  assert(called);
+  assert(status === (await Deno.permissions.query({ name: "hrtime" })));
```
bartlomieju, line 32: Is there an integration test that check event listeners after actual granting of permissions? If not would be great to have one, but I get that it might be hard to do (given that prompt tests are rather flaky)

bartlomieju (approved):
LGTM 🚀

cli/tests/unit/permissions_test.ts
kitsonk: No there isn't, but in a way I think we should use integration tests more sparingly and not have them be a coverage for lack of unit tests. (And we need another flaky test like we need a hole in our heads)

PR #9579, opened 2021-02-22
fix(runtime): do not panic on irregular dir entries

### LINKED ISSUE #9578: Deno.readDir("/dev/fd") is panic
```
❯ deno --version
deno 1.7.5 (release, x86_64-apple-darwin)
v8 9.0.123
typescript 4.1.4

❯ deno eval 'Deno.readDir("/dev/fd")'
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 9, kind: Other, message: "Bad file descriptor" }', runtime/ops/fs.rs:1059:43
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(...)', runtime/ops/fs.rs:1077:4
```

caspervonb:
Fixes #9578

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1418,9 +1418,9 @@ declare namespace Deno {
 
   export interface DirEntry {
     name: string;
-    isFile: boolean | null;
-    isDirectory: boolean | null;
-    isSymlink: boolean | null;
+    isFile: boolean;
+    isDirectory: boolean;
+    isSymlink: boolean;
   }
 
   /** Synchronously reads the directory given by `path` and returns an iterable
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1418,9 +1418,9 @@ declare namespace Deno {
 
   export interface DirEntry {
     name: string;
-    isFile: boolean;
-    isDirectory: boolean;
-    isSymlink: boolean;
+    isFile: boolean | null;
+    isDirectory: boolean | null;
+    isSymlink: boolean | null;
   }
 
   /** Synchronously reads the directory given by `path` and returns an iterable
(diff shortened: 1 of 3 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -1418,9 +1418,9 @@ declare namespace Deno {
 
   export interface DirEntry {
     name: string;
-    isFile: boolean;
-    isDirectory: boolean;
-    isSymlink: boolean;
+    isFile: boolean | null;
+    isDirectory: boolean | null;
+    isSymlink: boolean | null;
```
bartlomieju, line 1423: This is a breaking change

caspervonb: Right; initially I was going to go with just having everything be false here and on second thought I think that makes alot more sense.

Deno.stat returns all as false without panicking.

```js
> Deno.statSync("/dev/fd/2")
{
  isFile: false,
  isDirectory: false,
  isSymlink: false,
  size: 0,
  mtime: 2021-02-22T14:55:17.914Z,
  atime: 2021-02-22T14:55:17.914Z,
  birthtime: 1970-01-01T00:00:00.000Z,
  dev: 608056912,
  ino: 649,
  mode: 8592,
  nlink: 1,
  uid: 501,
  gid: 4,
  rdev: 268435457,
  blksize: 131072,
  blocks: 0
}
```

caspervonb: Went back to my first pass on this which is just returning false for all fields.

bartlomieju (approved):
I'm in favor of this change, waiting on second opinion before landing

lucacasonato (approved):
LGTM. I am also in favor of this change.

ry (approved):
LGTM - thanks @caspervonb

PR #9580, opened 2021-02-22
refactor(core): cleanup module implementation

bartlomieju:
* remove "ModuleNameMap", instead define that map inline inside "Modules" struct
* remove "dyn_import_id" argument from "mod_instantiate"
* rename "Modules" struct to "ModuleMap"
* rename "JsRuntime::modules" to "JsRuntime::module_map"

More PRs to follow

### DIFF
--- a/core/modules.rs
+++ b/core/modules.rs
@@ -398,60 +398,13 @@ enum SymbolicModule {
   Mod(ModuleId),
 }
 
-#[derive(Default)]
-/// Alias-able module name map
-struct ModuleNameMap {
-  inner: HashMap<String, SymbolicModule>,
-}
-
-impl ModuleNameMap {
-  pub fn new() -> Self {
-    ModuleNameMap {
-      inner: HashMap::new(),
-    }
-  }
-
-  /// Get the id of a module.
-  /// If this module is internally represented as an alias,
-  /// follow the alias chain to get the final module id.
-  pub fn get(&self, name: &str) -> Option<ModuleId> {
-    let mut mod_name = name;
-    loop {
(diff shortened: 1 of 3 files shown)


core/modules.rs
```diff
@@ -398,84 +398,46 @@ enum SymbolicModule {
...
+    let mut mod_name = name;
+    loop {
+      let symbolic_module = self.by_name.get(mod_name)?;
+      match symbolic_module {
+        SymbolicModule::Alias(target) => {
+          mod_name = target;
+        }
+        SymbolicModule::Mod(mod_id) => return Some(*mod_id),
+      }
+    }
```
ry, line 432: Maybe there should be a counter here to detect infinite loops?

doc string would be helpful for this function.

core/runtime.rs
```diff
@@ -734,58 +734,46 @@ impl JsRuntime {
   /// `AnyError` can be downcast to a type that exposes additional information
   /// about the V8 exception. By default this type is `JsError`, however it may
   /// be a different type if `RuntimeOptions::js_error_create_fn` has been set.
-  fn mod_instantiate(
-    &mut self,
-    id: ModuleId,
-    dyn_import_id: Option<ModuleLoadId>,
-  ) -> Result<(), AnyError> {
```
ry, line 741: Nice

core/modules.rs
bartlomieju: I don't think so, this function hasn't changed in a long time and works as expected. Added a docstring.

ry (approved):
LGTM - nice clean up

PR #9586, opened 2021-02-23
test(lsp): add benchmarking tests

kitsonk:
This will ultimately add benchmarks for the LSP server.


cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+  runtime::Builder::new_current_thread()
+    .enable_io()
+    .enable_time()
+    // for performance purposes, we might as well allow this to go to as
+    // many threads as it can, so that the harness doesn't become the bottle
+    // neck
+    // .max_blocking_threads(32)
+    .build()
+    .unwrap()
+}
```
ry, line 63: Why not use `tokio_util::create_basic_runtime()`?

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub(crate) fn run_local<F, R>(future: F) -> R
+where
+  F: std::future::Future<Output = R>,
+{
+  let rt = create_basic_runtime();
+  rt.block_on(future)
+}
```
ry, line 71: tokio_util::run_basic

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+  pub output: Lines<BufReader<ChildStderr>>,
+  reader: BufReader<ChildStdout>,
+  start: Instant,
+  writer: BufWriter<ChildStdin>,
+}
+
+impl LspClient {
+  pub fn new(deno_exe: &PathBuf) -> Result<Self, AnyError> {
+    let mut child = Command::new(deno_exe)
+      .kill_on_drop(true)
```
ry, line 83: Aside: `kill_on_drop` - cool - I didn't know about this method

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+      .args(&["lsp"])
+      .stderr(Stdio::piped())
+      .stdin(Stdio::piped())
+      .stdout(Stdio::piped())
+      .spawn()?;
+
+    let stdout = child
+      .stdout
+      .take()
+      .ok_or_else(|| generic_error("could not get handle to stdout"))?;
```
ry, line 93: Nit: This isn't user facing code, unwrap is sufficient.

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+      .stdin
+      .take()
+      .ok_or_else(|| generic_error("could not get handle to stdin"))?;
+    let writer = BufWriter::new(stdin);
+
+    let stderr = child
+      .stderr
+      .take()
+      .ok_or_else(|| generic_error("could not get handle to stderr"))?;
+    let output = BufReader::new(stderr).lines();
```
ry, line 106: Is data transmitted over stderr? Maybe it's easier to debug if stderr inherits

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+        break;
+      }
+    }
+
+    let mut msg_buf = vec![0_u8; content_length];
+    let bytes_read = self.reader.read(&mut msg_buf).await?;
+    assert_eq!(bytes_read, content_length);
+    let msg = serde_json::from_slice(&msg_buf)?;
+
+    Ok(msg)
```
ry, line 148: Nice mini-parser.

This could be made into a standalone function that operates on an abstract AsyncRead. Then we could pass it some canned data to see if it parses it correctly.

```rust
fn parse_lsp_message<R>(reader: R) -> Result<Vec<u8>, AnyError>
 where
    R: AsyncRead
{
    /// ... 
}

fn test_parser() {
   let msg = "content-length: 11\r\n\r\nhello world";
   let buf = std::io::Cursor::new(msg.as_bytes());
   assert_eq!(parse_lsp_message(buf), "hello world".as_bytes());
}

 
```

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+
+  pub async fn read<V>(&mut self) -> Result<V, AnyError>
+  where
+    V: de::DeserializeOwned,
+  {
+    let mut content_length = 0_usize;
+    loop {
+      let mut buf = String::new();
+      self.reader.read_line(&mut buf).await?;
+      if let Some(captures) = CONTENT_TYPE_REG.captures(&buf) {
```
ry, line 132: nit: bit simpler `buf.to_lowercase().starts_with("content-type:")`

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+    let msg = serde_json::from_slice(&msg_buf)?;
+
+    Ok(msg)
+  }
+
+  pub async fn write(&mut self, value: Value) -> Result<(), AnyError> {
+    let msg = format!(
+      "Content-Length: {}\r\n\r\n{}",
+      value.to_string().as_bytes().len(),
+      value
```
ry, line 155: nit: converts value to string twice

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+    }))
+    .await?;
+
+  let mut exec_times = HashMap::new();
+  exec_times.insert("startup_shutdown".to_string(), client.duration());
+
+  println!("lsp bench: {:?}", exec_times);
+
+  Ok(exec_times)
+}
```
ry, line 274: Nice.

I have no idea how long this takes, but if its very fast, maybe loop and do this 1000 times (or whatever).

cli/bench/main.rs
```diff
@@ -401,38 +403,40 @@ fn cargo_deps() -> usize {
...
+  max_latency: HashMap<String, f64>,
   max_memory: HashMap<String, u64>,
+  lsp_exec_time: HashMap<String, u128>,
   req_per_sec: HashMap<String, u64>,
-  max_latency: HashMap<String, f64>,
-  thread_count: HashMap<String, u64>,
+  sha1: String,
   syscall_count: HashMap<String, u64>,
+  thread_count: HashMap<String, u64>,
+  throughput: HashMap<String, f64>,
```
ry, line 421: sha1 and created_at are special metadata, and should probably remain at the top for easy viewing

cli/bench/lsp.rs
kitsonk: 🤦  I thought it was in a different crate, but I guess it isn't... silly me.

cli/bench/lsp.rs
kitsonk: Only on `tokio::process::Command`...

cli/bench/lsp.rs
kitsonk: No data is transmitted... it is just what the lsp logs, like:

```
Starting Deno language server...
  version: 1.7.4 (debug, x86_64-apple-darwin)
Connected to "Visual Studio Code" 1.53.2
language server initialized
Server ready.
```

cli/bench/lsp.rs
kitsonk: but how to easily extract the length to size the buffer?

cli/bench/lsp.rs
ry: In the case of a panic, it would be nice to see that in the logs.
I suggest removing the `stderr(Stdio::piped())` bit.

cli/bench/lsp.rs
kitsonk: About 400ms on my machine and it seems fairly consistent... because there is time to start up the tsc isolate and generate the first diagnostics...  Something like 10 times for this one, but I think this will truly be the "baseline" one, and then we need a "more complex" one.

cli/bench/lsp.rs
ry: Ah I didn't realize this was tokio::process::Command... Do we really need to interact with the "deno lsp" child process in an asynchronous way? I think std::process::Command might be sufficient.  I'm slightly worried we might end up running multiple benchmarks concurrently or something.

cli/bench/lsp.rs
ry: oh right - you still need to extract it. Probably simpler using the RE.

cli/bench/lsp.rs
```diff
@@ -0,0 +1,274 @@
...
+use tokio::io::AsyncBufReadExt;
+use tokio::io::AsyncReadExt;
+use tokio::io::AsyncWriteExt;
+use tokio::io::BufReader;
+use tokio::io::BufWriter;
+use tokio::io::Lines;
+use tokio::process::ChildStderr;
+use tokio::process::ChildStdin;
+use tokio::process::ChildStdout;
+use tokio::process::Command;
```
ry, line 24: Use std::process::Command - I don't think this needs to be async.

cli/bench/lsp.rs
kitsonk: I have found it really hard to find a way to not block the stdout/stdin, and if they get blocked, the lsp can stop responding, it was the problem I had when I first tried the integration harness, which is partly why I was avoiding revisiting it, but the `tokio::process::Command` makes it so much easier to not block, and we really need to handle the requests and responses.  I actually found out that if I fired in all the commands and then read the results, the lsp actually complains that the "shutdown" command is invalid, because it knows that other messages haven't been sent.

Not going to run multiple benchmarks at the same time.  That would only happen with multiple instances of the `LspClient` which simply won't occur and the collection of the benchmarks for the lsp is blocked in the main.rs.

ry: As far as I can see below, you await each read/write serially. AFAICT there would be no effective difference if this was done with blocking syscalls.

cli/bench/lsp.rs
kitsonk: inherit, especially for benchmarking, will fill stderr with a lot of useless startup information... as discussed we should pipe it to null

ry (commented):
Is this example one that will be improved by Luca’s std::sync::Mutex patch?

kitsonk:
> Is this example one that will be improved by Luca’s std::sync::Mutex patch?

No, I am working on that now, but I wanted to make sure the basic case was working.

kitsonk:
Ok, there is a test now that opens up a 8400 line document and starts typing a function into it and every 4 edits, it will wait for diagnostics to be published.  I am getting about 2362ms mean for the test on my machine.  I've tried on my machine with a release build and vscode, and while it isn't super laggy, I think the test should be sufficient to detect if we are improving it the right direction.

kitsonk:
Ok Here are some results from my machine:

| | main | #9552 (luca) | #9572 (kit) |
| --- | --: | --: | --: |
| basic | 395 | 389 | 527 |
| big | 2362 | 2329 | 5132 |

Upon reflection, I don't see we are seeing enough of a change in Luca's approach because we aren't doing anything in the test except firing in document changes and waiting for diagnostics to come back, and the big change there was to deprioritise the diagnostics.

I went back and made a change to the test, where I introduced a "human" level of delay in the edits on the test and this is what I got:

| | main | #9552 (luca) | #9572 (kit) |
| --- | --: | --: | --: |
| big | 5110 | 5083 | 5370 |

So under more "realistic" conditions, you start to see things even out.  So it seems like we really need a more holistic test, with requests for hover, code completion, etc...

The biggest challenge I've got is I am not sure of a way to "drain" any pending messages that have already been sent to std out, so I can properly wait for things like diagnostic notifications and request responses.

I guess if I've got a request, response, I should just keep reading until I get the response back and potentially put any other notifications that are hanging out in some sort of "notification queue".  That makes sense...  Ok, I am going to have to restructure some stuff...

lucacasonato:
@kitsonk Can you try with #9552 and #9557 together? I would not expect just #9552 or just #9557 to improve stand alone.

kitsonk:
@lucacasonato that is a lot of merging... I need to fix the test first.  I figured out how and am working on it.

lucacasonato:
Oh sure. I'll push up a merged branch shortly.

ry:
“Big” and “basic” aren’t very descriptive names. We’re going to have to keep them the same forever (well they could be changed but it will be a lot of work). Suggest slightly more future proof names.

ry:
Also it seems to me that Big should be tweaked so that it maximally highlights the change Luca’s patch will make. It’s some unlikely pathological situation - but that’s good for exploring system performance.

kitsonk:
> “Big” and “basic” aren’t very descriptive names. We’re going to have to keep them the same forever (well they could be changed but it will be a lot of work). Suggest slightly more future proof names.

There are `"startup_shutdown"` and `"big_file_edits"` in the code.

> Also it seems to me that Big should be tweaked so that it maximally highlights the change Luca’s patch will make. It’s some unlikely pathological situation - but that’s good for exploring system performance.

I would prefer to ensure the test exhibits the adverse behaviour we were observing versus expecting the outcome of Luca's patch.  Luca's patch is one approach to trying to solve the problem.

Ok, I have the tests fully exhausting the message queue.  This doesn't have "other" things yet, just document changes.  To me this represents more the "crawling" diagnostics that are occurring in main, and the patch looks like it improves performance significantly.

| | main | combined (luca) |
| --- | --: | --: |
| basic | 392 | 453 |
| big | 7457 | 3636 |

ry (approved):
LGTM


We’ll need to do an update in the website2 repo after this lands to display the plot.

PR #9599, opened 2021-02-25
chore(cli/runtime): use re-export serde from deno_core

kitsonk:
We weren't using the re-export of serde/serde_json from deno_core everywhere in the code base, this PR cleans that up.


ry (approved):
👍

bartlomieju:
Maybe remove `serde` from `runtime/Cargo.toml` and `cli/Cargo.toml`?

kitsonk:
> Maybe remove `serde` from `runtime/Cargo.toml` and `cli/Cargo.toml`?

I couldn't figure out a way, as `#[derive()]` then breaks and it still doesn't stop you from importing the crate directly.

PR #9601, opened 2021-02-25
chore(core): error builder instead of error class

lucacasonato:
This PR moves core away from the concept of error classes, towards a
concept of error builders. These builders are simple functions that can
return any error value. This is needed for WebGPU, where many errors
are `DOMExceptions` that take two parameters, the message and the error
type. Currently we have no way to create errors that take more than one
paramterer.

### DIFF
--- a/core/core.js
+++ b/core/core.js
@@ -174,14 +174,14 @@ SharedQueue Binary Layout
     return send(opsCache[opName], control, ...zeroCopy);
   }
 
-  function registerErrorClass(errorName, className) {
+  function registerErrorBuilder(errorName, builder) {
     if (typeof errorMap[errorName] !== "undefined") {
-      throw new TypeError(`Error class for "${errorName}" already registered`);
+      throw new TypeError(`Error builder for "${errorName}" already registered`);
     }
-    errorMap[errorName] = className;
+    errorMap[errorName] = builder;
   }
 
-  function getErrorClass(errorName) {
+  function getErrorBuilder(errorName) {
     return errorMap[errorName];
   }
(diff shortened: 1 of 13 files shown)


core/core.js
```diff
@@ -203,13 +203,13 @@ SharedQueue Binary Layout
...
-    if (!ErrorClass) {
+    const errorBuilder = getErrorBuilder(res.err.className);
+    if (!errorBuilder) {
       throw new Error(
-        `Unregistered error class: "${res.err.className}"\n  ${res.err.message}\n  Classes of errors returned from ops should be registered via Deno.core.registerErrorClass().`,
+        `Unregistered error builder: "${res.err.className}"\n  ${res.err.message}\n  Builders for errors returned from ops should be registered via Deno.core.registerErrorBuilder().`,
       );
     }
-    throw new ErrorClass(res.err.message);
+    throw errorBuilder(res.err.message);
```
nayeemrmn, line 214: This add an extra junk stack frame in op errors ☹ (#4189) Can we not keep error classes on the JS side?

lucacasonato: As described in the PR description: This is needed for WebGPU, where many errors
are `DOMException`s that take two parameters, the message and the error
type. Currently we have no way to create errors that take more than one
parameter.

nayeemrmn: Can the infrastructure return a `[class, ...args]` tuple? It just needs to be constructed here.

lucacasonato: I guess that would work. I'll open another PR :upside_down_face:

PR #9614, opened 2021-02-27
feat(extensions/crypto): implement generateKey() & sign()

littledivy:
Related #1891 
Implements `CryptoKey` related logic along with the following methods:

`generateKey()` and `sign()`
  - [x] RSASSA-PKCS1-v1_5
  - [x] RSA-PSS
  - [x] ECDSA 
  - [x] HMAC 
  
Uses [`ring`](https://briansmith.org/rustdoc/ring/) for ECDSA & HMAC
Uses [`rsa`](https://github.com/RustCrypto/RSA) for RSA-*


op_crates/crypto/Cargo.toml
```diff
@@ -15,5 +15,11 @@ path = "lib.rs"
 
 [dependencies]
 deno_core = { version = "0.80.1", path = "../../core" }
-rand = "0.8.3"
+rand = "0.7.0"
+ring = { version = "0.16.20", features = ["std"] }
+rsa = { git = "https://github.com/RustCrypto/RSA", rev = "616b08d" }
```
littledivy, line 20: Note: Using via git because`RSAPrivateKey::new_with_exp` is not published to crates.io yet.

op_crates/crypto/lib.rs
```diff
@@ -43,6 +81,381 @@ pub fn op_crypto_get_random_values(
...
+        .hash
+        .ok_or_else(|| WebCryptoError::MissingArgument("hash".to_string()))?
+        .into();
+      let rng = RingRand::SystemRandom::new();
+      
+      // if let Some(length) = args.algorithm.length {
+      //   let mut alg = hash.digest_algorithm();
+      //   alg.output_len = length as usize;
+      //   alg.chaining_len = length as usize;
+      // };
```
littledivy, line 315: `ring` expects digest algorithm at key generation thus we cannot modify the digest output length (thanks to ring's easy-to-misuse api).

Currently `generateKey` uses:
P256 -> SHA256
P384 -> SHA384
these are the recommended signing algorithms as per the [docs](https://briansmith.org/rustdoc/ring/signature/index.html#statics).
(also idk why people would use a custom length :thinking: (maybe just the spec))

op_crates/crypto/lib.rs
```diff
@@ -43,6 +81,381 @@ pub fn op_crypto_get_random_values(
...
+      let key_pair = &resource.key;
+
+      // We only support P256-SHA256 & P384-SHA384. These are recommended signature pairs.
+      // https://briansmith.org/rustdoc/ring/signature/index.html#statics
+      if let Some(hash) = args.hash {
+        match hash {
+          WebCryptoHash::Sha256 | WebCryptoHash::Sha384 => (),
+          _ => return Err(WebCryptoError::UnsupportedHash.into()),
+        }
+      };
```
littledivy, line 434: ring expects digest algorithm at key generation, so its not possible to apply new hash at signing. Thus we use SHA256 for P256 and SHA384 for P384 at key generation. (This is the recommended signing algorithms as per the [docs](https://briansmith.org/rustdoc/ring/signature/index.html#statics)).

littledivy:
Blocked by https://github.com/RustCrypto/RSA/issues/81

cli/tests/unit/webcrypto_test.ts
```diff
@@ -0,0 +1,58 @@
...
+    },
+    true,
+    ["sign", "verify"],
+  );
+
+  assert(keyPair.privateKey);
+  assert(keyPair.publicKey);
+  assertEquals(keyPair.privateKey.extractable, true);
+  assert(keyPair.privateKey.usages.includes("sign"));
+});
```
ry, line 22: Nice!

op_crates/crypto/lib.rs
```diff
@@ -43,6 +81,441 @@ pub fn op_crypto_get_random_values(
...
+
+      let exponent = BigUint::from_bytes_be(exp);
+
+      // Generate RSA private key based of exponent, bits and Rng.
+      let mut rng = OsRng;
+      let private_key = RSAPrivateKey::new_with_exp(
+        &mut rng,
+        modulus_length as usize,
+        &exponent,
+      )?;
```
ry, line 205: I'm guessing this is a CPU intensive function (if not this one exactly then probably some other call in this op). This op is asynchronous, but the way it's organized now this CPU-heavy function is going to block the event loop.

I think this whole bit should be wrapped in an async block and passed to `tokio::task::spawn_blocking()`.

See also https://docs.rs/tokio/1.2.0/tokio/#cpu-bound-tasks-and-blocking-code

ry (commented):
This looks very promising - thanks @littledivy ! We'll try to get this into v1.9 which is [scheduled](https://deno.land/manual@main/contributing/release_schedule) for April 13th.

op_crates/crypto/lib.rs
littledivy: Upstream issue: https://github.com/briansmith/ring/issues/1216

op_crates/crypto/lib.rs
littledivy: done :+1:

littledivy:
Update on the blockers:
- opened PR at `rsa` to update their rand version - https://github.com/RustCrypto/RSA/pull/88
- opened PR at `ring` to add functionality to generate HMAC key with custom length - https://github.com/briansmith/ring/pull/1218

op_crates/crypto/Cargo.toml
littledivy: Edit: Using my fork of the RSA crate to fix rand issues. 
Self ToDo: Change to RSA when fork is merged & released.

littledivy: Fixed

caspervonb (commented):
Web platform tests don't seem to be enabled? (caveat, there's some inter-dependencies within subtle.crypto in the tests).

Run wpt.ts with update, something ala:
```sh
deno run tools/wpt.ts update -- WebCryptoAPI/generateKey
```

Repeat for each relevant folder.

littledivy:
Enabled WPT for `generateKey` (passes locally). Needs https://github.com/denoland/wpt/pull/14 to be landed for the CI to run.

littledivy:
Removed `serde_json` as per #10009

op_crates/crypto/01_crypto.js
```diff
@@ -41,11 +41,170 @@
...
+      };
+    }
+  }
+
+  async function sign(alg, key, data) {
+    const rid = key[ridSymbol];
+    const simpleAlg = typeof alg == "string";
+    const saltLength = simpleAlg ? null : alg.saltLength;
+    const hash = simpleAlg ? null : alg.hash;
+    const algorithm = (simpleAlg ? alg : alg.name).toLowerCase();
```
littledivy, line 181: Will polish this more when implementing `SubtleCrypto.verfiy()`

lucacasonato:
Can you re-add the cli unit tests? This unclosable resource problem needs a test.

jadhavajay:
@littledivy just curious - are you supporting Ed25519 as well ?

littledivy:
> @littledivy just curious - are you supporting Ed25519 as well ?

Unfortunately, Curve25519 is not part of the standard WebCrypto :(

extensions/crypto/lib.deno_crypto.d.ts
```diff
@@ -49,9 +128,32 @@ interface SubtleCrypto {
...
+      | Uint32Array
+      | Uint8ClampedArray
+      | Float32Array
+      | Float64Array
+      | DataView
+      | null,
+  >(
+    array: T,
+  ): T;
+  randomUUID(): string;
```
CrimsonCodes0, line 149: Is randomUUID implemented? I've never seen it in any specification nor I didn't spot an implementation.

lucacasonato: It is implemented and specified: https://wicg.github.io/uuid/. It is shipping in Chrome 92 and has already shipped in Node.

CrimsonCodes0: Oh, I hadn't heard of that draft spec yet, thanks!

lucacasonato:
Lets get this PR landed for 1.12. I have some concerns though:

- I dislike that keys are backed by resources. There is no idiomatic way for the resources to be cleaned up. This is a blocker for landing this PR.
- The security model of storing the "non extractable" keys in process memory, just outside of JS and instead in the Rust resource table is flawed. The memory contents can still be accessed via timing side channels. Any in process memory is always going to have to be untrusted. This is also a blocker for landing this PR.

To circumvent these issues I suggest we drop support for "non extractable" keys and throw a `SecurityError` DOMException when you try to create one. The bytes that back a key should be stored on the JS object, and no resources should be involved.

extensions/crypto/lib.rs
```diff
@@ -234,210 +165,79 @@ pub async fn op_webcrypto_generate_key(
...
+#[derive(Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub struct KeyData {
+  r#type: KeyFormat,
+  data: ZeroCopyBuf,
 }
 
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct WebCryptoSignArg {
```
caspervonb, line 220: Consistency nit: can we remove the Web prefix on these or is there a clash?

littledivy: Done :+1: 
1. removed Web prefix and instead used Crypto for generic names (eg: `CryptoHash` instead of `WebCryptoHash` or `Hash`)
2. argument structs are ending with `Arg` to avoid conflicts (eg: `AlgorithmArg` and not just `Algorithm` (conflicting))
3. renamed ops to follow `op_crypto_` pattern

extensions/crypto/00_webidl.js
```diff
@@ -23,4 +68,164 @@
...
+      key: "namedCurve",
+      converter: (V, opts) => {
+        const namedCurve = SupportedNamedCurves
+          .find((key) => key.toLowerCase() == V.toLowerCase());
+        if (namedCurve == undefined) {
+          throw new DOMException(
+            "namedCurve not supported",
+            "NotSupportedError",
+          );
+        }
```
lucacasonato, line 118: Don't do this in the webidl validator. Do it after.

extensions/crypto/01_crypto.js
```diff
@@ -97,6 +196,239 @@
 
       return result.buffer;
     }
+
+    async sign(algorithm, key, data) {
+      const prefix = "Failed to execute 'sign' on 'SubtleCrypto'";
+
+      webidl.assertBranded(this, SubtleCrypto);
+      webidl.requiredArguments(arguments.length, 3);
+
```
lucacasonato, line 205: WebIDL converters always must run before all other algorithm steps. They can not be intermingled.

littledivy:
a. The keys are now stored in JS in their exportable form (`raw` for HMAC and `pkcs8` for rest).
b. Using `generateKey()` with extractable=false throws a SecurityError
c. Relevant WPT pass. Notes on expected failures:
    1. Most expected failures are due to throwing SecurityError on non-extractable keys.
    2. Certain `failures_AES-*` and `failures_ECDH`, `failures_RSA-OAEP` tests pass because of the missing impl. 
    3. Some `ECDSA` tests fail due to lack of support for `P-521`

extensions/crypto/00_webidl.js
littledivy: Done

extensions/crypto/01_crypto.js
littledivy: I'll do this for `sign` but for `generateKey` WPT expects us to throw error for algorithm normalization before running converters on other arguments. eg: 
```typescript
// should throw for wrong hash identifier instead of extratable !== boolean
generateKey({hash: "INVALID", name:"HMAC"}, 0, ["sign"]);
```

lucacasonato: What I meant was that the webidl converters have to run _before_ the `Let normalizedAlgorithm be the result of normalizing an algorithm`.

So the code should look like this:
```js
class SubtleCrypto {
    async sign(algorithm, key, data) {
      webidl.assertBranded(this, SubtleCrypto);
      webidl.requiredArguments(arguments.length, 3);
      algorithm = webidl.converters.AlgorithmIdentifier(algorithm, {
        prefix,
        context: "Argument 1",
      });
      key = webidl.converters.CryptoKey(key, {
        prefix,
        context: "Argument 2",
      });
      data = webidl.converters.BufferSource(data, {
        prefix,
        context: "Argument 3",
      });
      const normalizedAlgorithm = normalizeAlgorithm(algorithm, "sign");

      // ...
    }
}
```

WebIDL converters need to all be finished before the operation steps defined in the spec.

littledivy: Ah! I misunderstood, will do. Also since `CryptoKey` is an interface it won't have a converter?

littledivy: done

lucacasonato:
Unfortunately we cannot test `sign` until we have `verify`, because they are both used in the WPT tests.

lucacasonato (approved):
Happy with this now. Thanks a lot @littledivy, and sorry that this took so long to land.

@bartlomieju could you additionally review?

@bnoordhuis could you review the key generation ops for any egregious security issues?

extensions/crypto/Cargo.toml
```diff
@@ -18,5 +18,11 @@ deno_core = { version = "0.92.0", path = "../../core" }
 deno_web = { version = "0.41.1", path = "../web" }
 tokio = { version = "1.7.1", features = ["full"] }
 rand = "0.8.3"
-ring = "0.16.20"
+ring = { version = "0.16.20", features = ["std"] }
+rsa = "0.4.0" # TODO: remove "pem" feature when next release is on crates.io
```
bartlomieju, line 22: This TODO is outdated

extensions/crypto/lib.rs
```diff
@@ -1,36 +1,75 @@
...
 
 pub use rand; // Re-export rand
 
+mod key;
+
+use crate::key::Algorithm;
+use crate::key::CryptoHash;
+use crate::key::CryptoNamedCurve;
+
+// Whitelist for RSA public exponents.
```
bartlomieju, line 53: s/Whitelist/Allowlist/

extensions/crypto/Cargo.toml
lucacasonato: No, I am waiting for https://github.com/RustCrypto/RSA/commit/366ff6eac3180be10362855564b212987aa73030 to land in a release.

extensions/crypto/Cargo.toml
```diff
@@ -18,5 +18,11 @@ deno_core = { version = "0.92.0", path = "../../core" }
...
+sha-1 = "0.9.4"
+sha2 = "0.9.3"
```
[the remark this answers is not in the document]
littledivy: thanks, upgraded.

extensions/crypto/lib.rs
```diff
@@ -66,6 +105,255 @@ pub fn op_crypto_get_random_values(
...
+
+      let exponent = BigUint::from_bytes_be(&public_exponent);
+      if exponent != *PUB_EXPONENT_1 && exponent != *PUB_EXPONENT_2 {
+        return Err(custom_error(
+          "DOMExceptionOperationError",
+          "Bad public exponent",
+        ));
+      }
+
+      let mut rng = OsRng;
```
bnoordhuis, line 139: Using different PRNGs is kind of regrettable, although I understand you can't really get around that. I trust both rand and ring to get it right but still...

bnoordhuis (approved):
Rust bits LGTM. Thanks for the PR!

extensions/crypto/key.rs
```diff
@@ -0,0 +1,111 @@
...
+
+#[derive(Serialize, Deserialize, Copy, Clone)]
+pub enum CryptoNamedCurve {
+  #[serde(rename = "P-256")]
+  P256,
+  #[serde(rename = "P-384")]
+  P384,
+}
+
+impl TryInto<&RingAlgorithm> for CryptoNamedCurve {
```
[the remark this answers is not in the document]
bartlomieju: Fixed

bartlomieju (approved):
LGTM too

panva:
@lucacasonato 

> To circumvent these issues I suggest we drop support for "non extractable" keys and throw a `SecurityError` DOMException when you try to create one. The bytes that back a key should be stored on the JS object, and no resources should be involved.

Can you reconsider the way non-extractable keys are handled. Existing libraries that implement higher level functions often work with generating single-use keys generated on the fly to use in lower level functions. These use `extractable: false` (also because some algorithms only work with extractable: false, e.g. `PBKDF2`).

This lack of support will mean existing libraries will not be usable under deno, only because of this one thing.

lucacasonato:
@panva We are happy to reconsider, but first we need clarification from the spec authors if "non extractable keys" are meant to be a security primitive or not. Deno has no way to implement non extractable keys securely if they are meant to be a security primitive.

I'll do some more research and will reach out to some WebCrypto API authors to figure out the details.

panva:
> @panva We are happy to reconsider, but first we need clarification from the spec authors if "non extractable keys" are meant to be a security primitive or not. Deno has no way to implement non extractable keys securely if they are meant to be a security primitive.
> 
> I'll do some more research and will reach out to some WebCrypto API authors to figure out the details.

Thank you, that's fair.

I have not encountered a normative language in the spec that would say e.g. that the memory used to hold the non-extractable keying material must be out of reach, it only says the raw keying material may not be exported by the **application**.

> extractable
> Reflects the [[extractable]] internal slot, which indicates whether or not the raw keying material may be exported by the application.

lucacasonato:
I have opened https://github.com/denoland/deno/issues/11481 to figure this out.

Issue #9461, opened 2021-02-09
Review import maps alignment against spec


kitsonk:
We need to review the import maps against the spec ahead of #4958.

PR #9616, opened 2021-02-27
feat: Align import map to spec and test using WPT

bartlomieju:
This commit updates implementation of import maps to
align it to current revision of the spec. 

Existing tests were removed in favor of using suite from 
WPT.

Closes https://github.com/denoland/deno/issues/9461


lucacasonato (approved):
LGTM. Huge to have this aligned and stable!

PR #9619, opened 2021-02-27
fix(runtime/web_worker): Don't block self.onmessage with TLA

### LINKED ISSUE #8735: Workers & top-level-await changed behavior between 1.4.6 and 1.6.0

### LINKED ISSUE #9643: Inspector blocked during top-level await
```ts
// deno run --inspect temp.ts
setTimeout(null, 1e8); // Keeps the process alive.
await new Promise(r => {});
```
When I run this and go to `chrome://inspect/`, the remote target is there but when I click "trace", there is **no** `Debugger session started.` in the terminal and the inspector rejects input.

Doesn't occur when the same code is wrapped in an async iife:
```ts
// deno run --inspect temp.ts
(async () => {
  setTimeout(null, 1e8); // Keeps the process alive.
  await new Promise(r => {});
})();
```

nayeemrmn:
Fixes #8735.
Fixes #9643.
Closes #8863.

cc @bartlomieju

This is #8863 with the test rewritten to not use `worker.terminate()`, not blocked by an unrelated bug.


bartlomieju (approved):
LGTM, thanks Nayeem

soootaleb:
That's awesome guys, I can't wait to upgrade my runtime. Do you know if this is going to be released with 1.8.1 ?

bartlomieju:
> That's awesome guys, I can't wait to upgrade my runtime. Do you know if this is going to be released with 1.8.1 ?

Yes, it will be

PR #9622, opened 2021-02-28
chore(runtime): removed unused dependencies

humancalico:
This PR removes the following unused dependencies from the runtime
crate:
- env_logger
- rustyline-derive
- shell-escape


bartlomieju (approved):
LGTM, thanks!

PR #9625, opened 2021-02-28
chore(runtime): remove rustyline dependency

caspervonb:
We depend on rustyline to translate the error type it provides but rustyline is not callable from any runtime ops.
It is only used internally in cli/tools/repl.ts so this removes any dependency on it in the runtime crate.


bartlomieju (approved):
LGTM, thanks Casper

Issue #9348, opened 2021-01-31
lsp errors with "The current snapshot version is missing."


MierenManz:
I have no idea what exactly happened
```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', cli\lsp\tsc.rs:806:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error processing TypeScript diagnostics:
channel closed
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', cli\lsp\language_server.rs:532:8
```

kitsonk:
Can you provide the version of Deno you were using?  Also, what you were doing at the time it occurred?

MierenManz:
Either deno 1.7.0 or deno 1.7.1 (not sure, sorry) and I was editing the .vscode/settings.json

kitsonk:
Are you using a `tsconfig.json` with the workspace?  If so, could you share it?

MierenManz:
I do not. Didn't even know I could to be honest.

kitsonk:
Ok, sorry to bother, but could you provide a version of your `.vscode/settings.json`.  Basically trying to figure out why you were in this part of the code when you were editing `.vscode/settings.json` because there is nothing obvious of why you ended up where you ended up.

MierenManz:
```json
{
    "deno.enable": true,
    "deno.lint": true
}
```
This is the only thing in the settings.json

after I thought about it. I think these were the steps that caused it
1. delete settings.json
2. disable deno canary
3. reload vsc
4. enable deno canary
5. create setting.json with your hand and copy paste these settings.

This is what I did

danopia:
I think I just encountered the same panic in 1.7.5. The code has changed a bit since 1.7.0 but it's a none-unwrap within the same function. My stack is pointing at this line: https://github.com/denoland/deno/blob/v1.7.5/cli/lsp/tsc.rs#L1099

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', cli/lsp/tsc.rs:1099:63
```

It happened twice about 90 seconds apart. I don't really recall doing anything interesting; just working on some typescript libraries. I'll report back if it recurs with more-interesting notes.

kitsonk:
[...]

PR #9628, opened 2021-02-28
fix(lsp): improve exception handling on tsc snapshots

kitsonk:
Refs #9348


lucacasonato (approved):
LGTM!

ry (approved):
LGTM
Does it make sense to add a test for this?

kitsonk:
> Does it make sense to add a test for this?

As I mention in #9348 I can't yet reproduce it, so I am not sure how to add a test case.  This should allow it to at least not panic and provide the information we need to fix it, which then would result in a test case.

PR #9635, opened 2021-03-01
chore: split web op crate

lucacasonato:
This commit starts splitting out the deno_web op crate into multiple
smaller crates. This commit splits out WebIDL and URL API, but in the
future I want to split out each spec into its own crate. That means we
will have (in rough order of loading): `webidl`, `dom`, `streams`,
`console`, `encoding`, `url`, `file`, `fetch`, `websocket`, and
`webgpu` crates.

It also removes the tests in op_crates/web. These are already tested via
WPT.


ry (approved):
LGTM

PR #9639, opened 2021-03-02
test(op_crates/web): regression tests for bugs in old URL implementation

piscisaureus:
Closes: #9383


lucacasonato (approved):
LGTM - but are there no WPT for this? If not we maybe should add some and upstream so we don’t diverge.

kt3k (approved):
LGTM

Issue #9284, opened 2021-01-26
deno test - Cannot resolve module


LukeChannings:
```
λ deno --version
deno 1.7.0+ada43cc (canary, x86_64-apple-darwin)
v8 8.9.255.3
typescript 4.1.3
```

I have an import map:

```import_map.json
{
  "imports": {
    "/": "./",
  }
}
```

My program runs correctly with:

```
deno run -A --unstable --import-map=./import_map.json ./cmd/moviematch/main.ts
```

I have a simple test file:

```typescript
import { assertEquals } from "https://deno.land/std@0.84.0/testing/asserts.ts";

import { getDirectories } from "/internal/app/plex/api.ts";

Deno.test("getDirectories", async () => {
 // test getDirectories
});
```

When I run:

```
λ deno test --import-map ./import_map.json -A --unstable
```

I get:

```
Check file:///Users/luke/Developer/MovieMatch/$deno$test.ts
error: Cannot resolve module "file:///Users/luke/Developer/MovieMatch/Users/luke/Developer/MovieMatch/internal/app/plex/api_test.ts" from "file:///Users/luke/Developer/MovieMatch/$deno$test.ts".
```

Things I've tried:

- Moving the test into a standalone test folder and selecting it with `deno test --import-map ./import_map.json -A --unstable tests`
- Importing the source relatively (works, but absolute imports are used throughout so it fails further on)

Why is the behaviour of `deno test` different to `deno run`?

kitsonk:
`deno test` creates a "synthetic" module which imports all the test modules it find, that is why the behaviour is different.

nayeemrmn:
I was going to say we should reconsider a non-file URL for `$deno$test.ts` so it's not subject to this, but `$deno$eval.ts` and `$deno$repl.ts` etc. can't be helped because we want relative specifiers to work for them.

I think the fake module specifier we use should be put through the import map, and the result of that should be passed to `insert_cached()` instead.

kitsonk:
> I think the fake module specifier we use should be put through the import map, and the result of that should be passed to `insert_cached()` instead.

That sounds reasonable.

caspervonb:
If https://github.com/denoland/deno/pull/9662 ends up landing we'll be removing the synthetic test module as a side effect.

LukeChannings:
As a side note, I've found that I can work around this issue by being more specific with aliases:

```json
{
  "imports": {
    "/internal/": "../internal/",
    "/types/": "../types/"
  }
}
[...]

PR #9662, opened 2021-03-03
wip: parallel-tests

caspervonb:
- Each test module runs in its self contained thread with a tokio runtime and main worker, per test would be ideal but not feasible as it's unexpected with test modules that have side effects.
- Reporting moved to Rust so that results can accumulate.

Closes https://github.com/denoland/deno/issues/9284


caspervonb:
Some preliminary measurements.

This patch:
```
time ./target/deno/release test --no-check --allow-all --unstable test_util/std
real	0m39.331s
user	1m49.884s
sys	0m11.864s
```

Canary:
```
time deno test --no-check --allow-all --unstable test_util/std
real	1m31.946s
user	2m33.789s
sys	0m5.996s
```

caspervonb:
Looked into running an isolate per test and the overhead is definitively not free (roughly ~30ms per test case, but depends on the size of the test module).

Running each test in it's own clean isolate comes with the caveat is that test modules will be loaded N times where N is the number of tests rather than just the once users will be expecting so I'm considering that path completely discarded but not a complete waste as I got some measurements out of it.

We can re-consider running test cases in distinct Realms when that becomes a V8 feature.

Will be continuing down this path of one runtime and isolate per test module.
Primarily got three things left to do here:
- Prepare all modules before running any of the tests, need to ensure sure `deno test --no-run` is now any slower than canary. 
- Output test results in the same order as they would be if a single thread was used by buffering test messages instead of piping them directly from the test runner to the test reporter. If a test fails in `fail_fast` mode then flush the buffer and exit.
- Bring back the things I stripped out like coverage, etc.

caspervonb:
Superseeded by #9815

PR #9668, opened 2021-03-03
fix(cli/compile): do not append .exe depending on target

### LINKED ISSUE #9667: deno compile add the .exe extension on windows even if the target is specified
I am trying to compile an example for different targets, but deno compile it adds the `.exe` even though the target is specified.

image

image

removing `.exe` the executable works on its respective platform

https://github.com/denoland/deno/blob/79e2e6a1f47b9d24abab14cfd69376d825b52e03/cli/tools/standalone.rs#L135 I think this condition is responsible

littledivy:
Fixes bug described in https://discord.com/channels/684898665143206084/689420767620104201/816669791359991828

Prevents appending `.exe` extensions on windows based on target.

Fixes #9667

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -354,7 +354,8 @@ async fn compile_command(
 
   info!("{} {}", colors::green("Emit"), output.display());
 
-  tools::standalone::write_standalone_binary(output.clone(), final_bin).await?;
+  tools::standalone::write_standalone_binary(output.clone(), target, final_bin)
+    .await?;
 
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
@littledivy looks good, can we get some test to ensure this situation doesn't happen again?

ewired (approved):

ry (approved):
LGTM - thanks @littledivy !

PR #9684, opened 2021-03-04
fix: getBindGroupLayout always illegal invocation

lucacasonato:
Reported by @DjDeveloper.

Will be tested by WebGPU CTS in the future.


ry (approved):
We really need some tests...

PR #9685, opened 2021-03-04
fix(runtime/js): Add navigator interface objects

nayeemrmn:
Chrome canary: 
image

### DIFF
--- a/runtime/js/99_main.js
+++ b/runtime/js/99_main.js
@@ -35,6 +35,7 @@ delete Object.prototype.__proto__;
   const denoNs = window.__bootstrap.denoNs;
   const denoNsUnstable = window.__bootstrap.denoNsUnstable;
   const errors = window.__bootstrap.errors.errors;
+  const webidl = window.__bootstrap.webidl;
   const { defineEventHandler } = window.__bootstrap.webUtil;
 
   let windowIsClosing = false;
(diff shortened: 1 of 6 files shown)


runtime/js/99_main.js
```diff
@@ -203,6 +203,52 @@ delete Object.prototype.__proto__;
     );
   }
 
+  class Navigator {
+    constructor(key = null) {
+      if (key != webUtil.illegalConstructorKey) {
+        throw new TypeError("Illegal constructor.");
+      }
+    }
+  }
```
lucacasonato, line 212: This is not spec conformant. Use the same construction technique we use in WebGPU. `Navigator.length` will be wrong.

It boils down to constructing the class with `Object.create`, branding the class using a symbol, and then checking the symbol in each method.

lucacasonato (changes requested):

runtime/js/99_main.js
```diff
@@ -204,46 +205,50 @@ delete Object.prototype.__proto__;
   }
 
   class Navigator {
-    constructor(key = null) {
-      if (key != webUtil.illegalConstructorKey) {
-        throw new TypeError("Illegal constructor.");
-      }
+    [webidl.brand] = webidl.brand;
```
lucacasonato, line 208: This isn't needed. `createBranded` will add this.

runtime/js/99_main.js
```diff
@@ -204,46 +205,50 @@ delete Object.prototype.__proto__;
...
       },
     },
   });
 
   class WorkerNavigator {
-    constructor(key = null) {
-      if (key != webUtil.illegalConstructorKey) {
-        throw new TypeError("Illegal constructor.");
-      }
+    [webidl.brand] = webidl.brand;
```
lucacasonato, line 233: Ditto

lucacasonato (approved):
LGTM. Thanks

PR #9714, opened 2021-03-07
fix(cli/ast): Pass importsNotUsedAsValues to swc

### LINKED ISSUE #9709: deno bundle should respect compiler option `importsNotUsedAsValues: "preserve"`
The emitted esm bundle by `deno bundle` does not preserve imports whose values or types are never used.

For example, this could be useful for compiling browser native [Web Components](https://developer.mozilla.org/en-US/docs/Web/Web_Components), where *parent* web components use the imported classes as both types and - in the html string - as *child elements* (but never as values): 

```typescript
// parent.ts
import { ChildElement } from "./child.ts"

html`<div>
<child-element></child-element>
</div>`
```
For more check out https://github.com/Polymer/lit-element

nayeemrmn:
Fixes #9709.

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -182,6 +182,13 @@ pub fn get_syntax(media_type: &MediaType) -> Syntax {
   }
 }
 
+#[derive(Debug, Clone)]
+pub enum ImportsNotUsedAsValues {
+  Remove,
+  Preserve,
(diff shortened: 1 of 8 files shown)


kitsonk (approved):
LGTM

Thanks a lot!

Issue #9589, opened 2021-02-23
Coverage ratio is inconsistent


sebastienfilion:
I have a coverage report that seems inconsistent. The specific file exports around 50 functions that should be tested and I know that I test only 9 of them. I get a final coverage ratio of 99.709%

```
cover file:///Users/sebastienfilion/Developer/functional/library/aviary.js ... 99.709% (686/688)
 687 | export const warblerstarstar = curryN(4, (f, x, y, z) => f(x)(y)(z)(z));
 688 |
```

Here's the relevant files:
https://github.com/functionalland/functional/blob/feature/aviary/library/aviary.js
https://github.com/functionalland/functional/blob/feature/aviary/library/aviary_test.ts
https://github.com/functionalland/functional/blob/feature/aviary/library/aviary.d.ts

kitsonk:
cc/ @caspervonb

caspervonb:
This is a bug.

Half of the intersection test used to reset the hit count is off causing false positives in some cases when there's multiple instrumented ranges on one line, single line lambdas for example don't get counted as having 0 hits.

Minimal reproduction:

```js
export function f(x) { return x; }
export const f1 = f(x => x);
```

Will fix this soon.

PR #9717, opened 2021-03-07
fix(coverage): ensure single line functions don't yield false positives

caspervonb:
Fixes https://github.com/denoland/deno/issues/9589


caspervonb:
Some duplication here; I have a terse reporter patch coming up that cleans up a bit so I'll leave the duplication alone for now.

lucacasonato (approved):
LGTM - thanks Casper

PR #9719, opened 2021-03-07
refactor(minimal ops) refactor minimal ops & rename to buffer ops

inteon:
Extracted minimal ops rewrite from #9457.

Is first step in effort to merge json and buffer ops.


runtime/ops/mod.rs
```diff
@@ -61,6 +62,26 @@ where
...
+  rt.register_op(name, metrics_op(name, buffer_op_async(op_fn)));
+}
+
+pub fn reg_buffer_sync<F, R>(rt: &mut JsRuntime, name: &'static str, op_fn: F)
+where
+  F: Fn(&mut OpState, u32, &mut [ZeroCopyBuf]) -> Result<R, AnyError> + 'static,
+  R: ValueOrVector,
+{
+  rt.register_op(name, metrics_op(name, buffer_op_sync(op_fn)));
+}
```
ry, line 83: I like how this is mirrors the structure of the JSON ops register function - so I think easier to understand.

runtime/ops/ops_buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+    let record_buf = bufs_iter.next().expect("Expected record at position 0");
+    let mut zero_copy = bufs_iter.collect::<BufVec>();
+
+    let req_header = match RequestHeader::from_raw(&record_buf) {
+      Some(r) => r,
+      None => {
+        let error_class = b"TypeError";
+        let error_message = b"Unparsable control buffer";
+        let len = error_class.len() + error_message.len();
+        let padding = &[b' ', b' ', b' '][0..(4 - (len & 3)) & 3];
```
ry, line 111: This padding logic is repeated several times. It would be good to abstract it out into a standalone helper function with its own test to prod the edge cases.

runtime/ops/ops_buffer.rs
```diff
@@ -0,0 +1,281 @@
...
+    Some(Self {
+      promise_id: ints[0],
+      argument: ints[1],
+    })
+  }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct ResponseHeader {
+  pub promise_id: u32,
```
ry, line 37: I hadn't realized this either, but I guess we're using 64 bit promises now.

runtime/ops/ops_buffer.rs
inteon: Ok I created `gen_padding_32bit` for this.

runtime/ops/ops_buffer.rs
inteon: Changed it to 64 bits.

inteon:
@ry I made the required changes and added some tests.
I use the little-endian rust functions to convert u64 to u8.
Is there a preference for little/ big endianness in Deno?

core/core.js
```diff
@@ -155,6 +155,10 @@ SharedQueue Binary Layout
     asyncHandlers[opId] = cb;
   }
 
+  function setAsyncHandlerByName(opName, cb) {
+    setAsyncHandler(opsCache[opName], cb);
+  }
```
bartlomieju, line 160: I think this function is superfluous

runtime/js/10_dispatch_buffer.js
```diff
@@ -0,0 +1,146 @@
...
+  ////////////////////////////////////////////////////////////////////////////////////////////
+
+  // General Async response handling
+  let nextRequestId = 1;
+  const promiseTable = {};
+
+  function opAsync(opName, opRequestBuilder, opResultParser) {
+    // Make sure requests of this type are handled by the asyncHandler
+    // The asyncHandler's role is to call the "promiseTable[requestId]" function
+    core.setAsyncHandlerByName(opName, (bufUi8, _) => {
```
bartlomieju, line 24: Not entirely sure, but I have a feeling that using arrow function here might be less performant than having a named function definition before `opAsync`

runtime/js/10_dispatch_buffer.js
```diff
@@ -0,0 +1,146 @@
...
+      const [requestId, result, error] = opResultParser(bufUi8, true);
```
bartlomieju, line 25: So we used array destructuring heavily in previous iterations of `core.js` and that caused a lot of GC triggers. Let's see what's the impact on performance...

runtime/js/10_dispatch_buffer.js
bartlomieju, line 24: Also wondering if this should be called only once, otherwise a new callback is set with each call to the op

runtime/js/10_dispatch_buffer.js
```diff
@@ -0,0 +1,146 @@
...
+
+  ////////////////////////////////////////////////////////////////////////////////////////////
+  /////////////////////////////////// Error handling /////////////////////////////////////////
+  ////////////////////////////////////////////////////////////////////////////////////////////
+
+  function handleError(className, message) {
+    const [ErrorClass, args] = core.getErrorClassAndArgs(className);
+    if (!ErrorClass) {
+      return new Error(
+        `Unregistered error class: "${className}"\n  ${message}\n  Classes of errors returned from ops should be registered via Deno.core.registerErrorClass().`,
```
bartlomieju, line 64: This line is longer than 80 characters, it's strange that formatter doesn't complain about it

runtime/js/10_dispatch_buffer.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    if (bufUi8.byteLength === headerByteLength) {
+      return [requestId, result, null];
+    }
+
+    // Rest of response buffer is passed as reference or as a copy
+    let respBuffer = null;
+    if (isCopyNeeded) {
+      respBuffer = bufUi8.slice(0, result); // Copy part of the response array (if sent through shared array buf)
```
bartlomieju, line 118: Ditto re line length

bartlomieju (commented):
A few more nitpicks from me, but after that I think we're ready to land this PR, great work @inteon

runtime/js/10_dispatch_buffer.js
inteon: ok, good to know; I guess a workaround can be used to replace this in case performance suffers.

runtime/js/10_dispatch_buffer.js
inteon: I think this should be called once. The current jsonOp implementation also does this wrongly:
https://github.com/denoland/deno/blob/bd961c3bc305e90e760394d2f424c5360a67543b/core/core.js#L214-L217
I do fix this for both bufferOps and jsonOps in #9457.

core/core.js
inteon: I agree, but this is the simplest way to do use the `opsCache` in 10_dispatch_buffer.js.
Currently (on master), this opId -> opName mapping is setup in `99_main.js` for minimal ops.
In #9457 this is resolved/ removed (as all logic moves to core.js).

runtime/js/10_dispatch_buffer.js
inteon: @bartlomieju maybe a hacky solution could be to reuse some shared variables and reassign them when returning?

core/core.js
bartlomieju: Okay, didn't catch that, sounds reasonable then

runtime/js/10_dispatch_buffer.js
inteon: In the latest commit the string should be split.

runtime/js/10_dispatch_buffer.js
inteon: opResultParser is passed into this arrow function, so that makes it hard to extract this arrow function from this opAsync function

inteon: a solution could be to call this function on an object that stores this opResultParser function, but I found this quickly makes the code a lot less readable. please let me know if you think this is useful (I could create such a solution and share it with you for review)

bartlomieju: @inteon is it required to pass that `opResultParser`? I think you could just directly call it in such function

runtime/js/10_dispatch_buffer.js
bartlomieju: No, let's not do that, it would be harder to reason about. Let's see if there will be a negative perf impact and if so address in a follow up

runtime/js/10_dispatch_buffer.js
inteon: In #9457, instead of passing `bufferOpParseResult` as argument, I pass `jsonOpParseResult`. This way code can be shared between jsonOps and bufferOps.
https://github.com/inteon/deno/blob/2f9a9cabd0d7c48f440677ff21885eac30d22ce3/core/core.js#L320-L334
https://github.com/inteon/deno/blob/2f9a9cabd0d7c48f440677ff21885eac30d22ce3/core/core.js#L364-L378

bartlomieju: I see, let's land it as is then.

bartlomieju (approved):
LGTM, thanks @inteon

PR #9720, opened 2021-03-07
fix(core) promise_id is 32bit instead of 64bit

inteon:
Extracted this inconsistency fix from #9457.

### DIFF
--- a/cli/tests/unit/dispatch_json_test.ts
+++ b/cli/tests/unit/dispatch_json_test.ts
@@ -21,7 +21,7 @@ unitTest(function malformedJsonControlBuffer(): void {
 
 unitTest(function invalidPromiseId(): void {
   const opId = Deno.core.ops()["op_open_async"];
-  const reqBuf = new Uint8Array([0, 0, 0, 0, 0, 0, 0]);
+  const reqBuf = new Uint8Array([0, 0, 0]);
   const resBuf = Deno.core.send(opId, reqBuf);
   const resText = new TextDecoder().decode(resBuf);
   const resObj = JSON.parse(resText);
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
@inteon I'm not sure this change is preferable; running out of 32bit promise ids could be possible after few hours of server running. What's the reasoning behind this change?

inteon:
In MinimalOp, i32 is used for promise_id and in json_ops usize is used.
In javascript the promise_id is only converted to a big int `BigInt(promiseId)` before it is passed to rust.
I'm trying to make all these promise_ids the same length.
But, like you said, maybe 64 bits should be used?
I would then change usize to u64, and use BigInt in js for the counter too.

bnoordhuis:
Conversion to BigInt isn't necessary and introduces a small performance penalty.

2^32 is too small but JS numbers are exact up to 2^53 and that's still unfathomably large (good for the next 285 years at a rate of 1 million promises/sec.)

inteon:
Ok, I get it now; in javascript a `Number` is used, which is more performant than an `BigInt` and can represent integer values up to ±2^53 − 1 exactly (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Numbers_and_dates).

In order to send these 53 useful bits, `setUint32` cannot be used.
Instead the `Number` is converted to a `BigInt` so it can be used in `setBigUint64`.

To summarize:
- in rust a promise_id is an `u64`
- in javascript a promise_id is a `Number`
Because of js, the usable promise_id range is [0, 2^53 - 1], so 11 js <-> rs transferred bits will be left unused.

I will change the MinimalOps behavior to match this, and this PR is no longer relevant.

PR #9721, opened 2021-03-07
fix(core) assertion failure in case of js error

inteon:
Extracted this bugfix from #9457.

In case JavaScript throws an unhandled error, part of the "shared_queue" could be still unprocessed. If this is the case; throw the js runtime error instead of asserting on the queue size not being 0.

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -2053,13 +2053,12 @@ pub mod tests {
           Deno.core.dispatch(1, new Uint8Array([42]));
           Deno.core.dispatch(1, new Uint8Array([42]));
           "#,
-          )
-          .unwrap();
+        )
+        .unwrap();
 
       assert_eq!(dispatch_count.load(Ordering::Relaxed), 2);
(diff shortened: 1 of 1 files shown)


bartlomieju:
@inteon could you provide a test for this edge case?

inteon:
@bartlomieju done

core/runtime.rs
```diff
@@ -2024,6 +2027,49 @@ pub mod tests {
     });
   }
 
+  #[test]
+  fn shared_queue_not_empty_when_js_error() {
```
bartlomieju, line 2031: I've verified that this test fails on `main`:

```
---- runtime::tests::shared_queue_not_empty_when_js_error stdout ----
thread 'runtime::tests::shared_queue_not_empty_when_js_error' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', core/runtime.rs:1438:7

```

bartlomieju (approved):
Nice fix! Thank you @inteon

Issue #9540, opened 2021-02-18
Optimize JSON ops: minimize allocs by decoding v8::Value directly


AaronO:
**TLDR:** implement a `serde` deserializer (& serializer) for `v8::Value` to minimize allocations/overhead of encoding to JSON 

## Problem & context

Deno-ops are kind of like syscalls, all deno programs must use them to do I/O and access runtime features.

Besides the v8 runtime, it's the only part of the stack that's not "user-optimizable" (users can easily optimize their own JS/TS code and JS/TS libraries, but touching runtime code will often be a stretch for many users).

As such I think it's a good principle to minimize syscall/op overhead, so the runtime is never "the bottleneck".

The current JSON op-ABI has some avoidable overhead, specifically heap-allocations and utf8 encoding.

(Note: I think most real-world programs will likely be I/O bound, or user-code CPU bound before being bottlenecked by the current op-ABI, but it's a good principle nonetheless and helps on microbenchmarks ...)

## Current implementation

The current op-ABI passes JSON args as following:

1. JSON stringify args
2. String-concat with promise ID (for async ops)
3. Encode to utf8
4. Parse JSON 
5. Allocate struct (on stack)
6. Allocate fields & fill struct

Steps `1.`, `2.` and `3.` make avoidable heap-allocations. CPU-time "wasted" on JSON parsing/stringification

## Alternative implementation

1. Pass JS value (`v8::Value`) as is
2. Allocate struct/receiving type
3. Fill receiving fields (with utf8 encoding for strings)

Basically pass JS objects untouched then decode using `serde_v8` (to be implemented)

## Benefits

- Better performance (less allocs) on op-heavy code 
- Remove overhead of utf8 encoding/decoding (only for string fields instead of the entire JSON string)
- Reduced conceptual overhead of fiddling with buffers (potentially less error-prone)
  - Ops could receive & return "natural" rust types (if we implement `serde_v8` serialization)

## Notes

- Needs to benchmarked with op-heavy code
- I focused on the JS->rust context-switch, but similarly the rust->JS return can avoid the `decodeJson()` step (utf8 decode + json parse) by serializing directly to a `v8::Value`
- `bin_ops` will probably still be faster than this improved `json_ops` since using shared memory with no-decoding is hard to beat, but this should help close the gap whilst preserving benefits & convenience of structured args

## Further changes

[...]

PR #9722, opened 2021-03-08
core: introduce serde_v8

AaronO:
This PR introduces `serde_v8` into `core` and uses it to reduce verbosity (making code arguably more readable and rusty).

It mainly serves to showcase `serde_v8` as an early alpha and gather feedback before using it for a new op-ABI.

## Subtasks
- [x] Implement `serde_v8`
- [x] Implement `serde_v8::Value` for passthrough (implementation is pretty "hacky", uses transmute)
- [x] Showcase verbosity improvements (on `get_proxy_details`, `get_promise_details`, `eval_context`)
- [x] Bring `serde_v8` into repo's tree (best mid-term solution IMO)
- [x] Flesh out README with: quickstart, best-practices, etc...
- [x] Add tests for `json!()` compat
- [x] Implement `serialize_map` and `deserialize_map` (mainly to support `serde_json::Value`, HashMaps, etc...)

## Follow-ups

In future PRs:
1. Use `serde_v8` to reduce verbosity of `JsStackFrame` decoding
2. Restore micro-benches from the original repo (using `bencher` or `criterion.rs`)
3. Use `serde_v8` to optimize op-ABI (https://github.com/denoland/deno/pull/9843)

Related to #9540


core/bindings.rs
```diff
@@ -17,6 +17,9 @@ use std::option::Option;
 use url::Url;
 use v8::MapFnTo;
 
+use serde::{Serialize};
```
bnoordhuis, line 20: I guess I shouldn't complain if clippy doesn't but since we don't use this pattern anywhere else...
```suggestion
use serde::Serialize;
```

core/bindings.rs
```diff
@@ -835,38 +781,29 @@ fn get_promise_details(
   args: v8::FunctionCallbackArguments,
   mut rv: v8::ReturnValue,
 ) {
+
```
bnoordhuis, line 784:
```suggestion
```

bnoordhuis (commented):
I'll need to look at serde_v8 in more detail but about this PR:

1. On the one hand, it removes quite a bit of code, which is definitely nice.

2. On the other hand, the serialization feels a little magical.

Things like tuple structs getting marshalled into JS arrays is kind of surprising to me but I also kind of see why it makes sense. Maybe I just need to get used to it.

lucacasonato:
> Things like tuple structs getting marshalled into JS arrays is kind of surprising to me but I also kind of see why it makes sense.

For reference, serde_json has the same behavior.

AaronO:
@bnoordhuis I completely understand the sentiment, but I don't think "feeling magical" is a negative here.

I think serialization layers can be broadly scored on: expressiveness, consistency/reliability and efficiency.

`serde_v8` is substantially more expressive than the "manual" v8-encoding, it's equally as efficient if not more so in some cases (since it uses internalized strings for struct/obj keys, ...) and I think you could argue that it's more consistent/reliable than the manual method since the data is now type-checked, preventing things such as typos in obj keys (and is consistent with the already adopted patterns and conventions of `serde`/`serde_json`).

`serde_v8` was originally built/prototyped for https://github.com/denoland/deno/issues/9540, to improve the efficiency of the op-serialization layer whilst preserving it's expressiveness (the original/current op-serialization trades off efficiency for expressiveness), mainly to drive overall performance gains and bridge the gap between json-ops & bin-ops (PR to come).

So applying it to the core functions wasn't the original intention, but it's also a natural fit anywhere we're mapping `v8<>rust`. For example in my [heapstats PR](https://github.com/denoland/deno/pull/9659), `serde_v8` will replace [these ~40-lines](https://github.com/denoland/deno/blob/5e9c409df5ab8030a1e638c76dfd65f2a9120ce4/core/bindings.rs#L946-L985) with a one-liner `to_v8()` call.

I personally found parts of `core/bindings` hard to grok at first, `get_proxy_details/get_promise_details/eval_context` are conceptually very simple but they were somewhat "drowned" in a lot of encoding complexity that made it hard(er) to discern what they're really doing. I believe reducing that conceptual overhead provides substantial gains in clarity, so we can focus on the essence of what we're doing and delegate 99% of the `v8<>rust` bijections to `serde_v8`.

This PR does make specific changes to `core/bindings` but I also wanted to use it to bootstrap the conversation on `serde_v8` and adopting it in a new op-ABI.

**Context on op-encoding perfs:** The [latest benches](https://github.com/denoland/deno/issues/9540#issuecomment-793441844), show that using `serde_v8` would reduce our op-encoding overhead by up-to ~15x, likely ~10x in most cases and at least ~3x in the "worst cases" (dealing with string values which cause extra allocs, string keys are decently fast since they are deduped by v8's internalized strings and be further improved with a KeyCache)

AaronO:
I think it could also be neat to have `serde_v8::assign_to(scope, obj, v)` to "merge" structs with existing `v8::Object`s.

We could also add `serde_v8::Function` similar to `Value` (also `Accessor` ...), but that might be a bit too magic ...

With `assign_to` or `merge_v8` we could use structs to encode the global context via `CoreGlobal`, `CoreFunctions` etc... 

Upside is that it would be "typed" and you could grasp the structure at a glance and serde would handle camelCase-ification and other details.

bnoordhuis:
> string keys are decently fast since they are deduped by v8's internalized strings

I left a comment on the relevant commit in serde_v8 but for posterity: the problem with interned strings is that they leak memory unless you know in advance that the set of strings is bounded.

I don't know if that's true for deno, but even if it's true now, it might not stay true going forward.

AaronO:
@bnoordhuis Thanks for the feedback!
(Just a heads up, the current code is a bit rough since I mainly focused on getting it working, benches etc... will do a second pass on it and some cleanup/polish)

Ultimately the plan is to use our own `KeyCache` to dedupe and use onebyte-static-external-strings for all struct keys. I don't have data on it but I assume the size/number of struct keys of op-ABI payloads is relatively small & bounded for most programs in practice, since it's essentially dependant on struct keys used by internal deno ops. However, with our own KeyCache we could easily use a bounded LRU.

I haven't implemented Maps in `serde_v8` yet, but I would exclude their keys from the KeyCache and use regular v8::Strings or use a separate cache to avoid cross-pollution since there should be much more variability in map-keys vs struct-keys and thus cache churn.

We could also use regular strings for this first release if you're concerned about the (potentially) unbounded mem-usage. They are naturally slower and will cause higher GC-pressure, but overall it's still faster than the current implementation

# Quick benches

### Internalized Strings

```
test abi_add_json     ... bench:     279,130 ns/iter (+/- 53,193)
test abi_add_v8       ... bench:      19,628 ns/iter (+/- 949)
test abi_add_void     ... bench:       5,399 ns/iter (+/- 1,221)
test abi_promote_json ... bench:     329,163 ns/iter (+/- 55,322)
test abi_promote_v8   ... bench:     108,140 ns/iter (+/- 8,793)
test abi_promote_void ... bench:      25,326 ns/iter (+/- 4,033)
test abi_sum_json     ... bench:     282,355 ns/iter (+/- 57,741)
test abi_sum_v8       ... bench:      35,914 ns/iter (+/- 3,013)
test abi_sum_void     ... bench:       5,654 ns/iter (+/- 1,445)
```

### Regular Strings

```
test abi_add_json     ... bench:     282,485 ns/iter (+/- 47,440)
test abi_add_v8       ... bench:      27,178 ns/iter (+/- 688)
test abi_add_void     ... bench:       7,079 ns/iter (+/- 2,267)
test abi_promote_json ... bench:     346,326 ns/iter (+/- 37,595)
test abi_promote_v8   ... bench:     128,139 ns/iter (+/- 5,183)
test abi_promote_void ... bench:      24,808 ns/iter (+/- 494)
test abi_sum_json     ... bench:     297,426 ns/iter (+/- 47,036)
test abi_sum_v8       ... bench:      33,818 ns/iter (+/- 1,184)
test abi_sum_void     ... bench:       5,571 ns/iter (+/- 1,253)
```

### Bench conclusions

Regular v8 strings are ~20% slower on these benches (decoding struct keys), but it's still an order of magnitude faster than the current implementation. So I think we can ship regular strings in the first release and then dedupe with our own bounded KeyCache in the future.

bartlomieju:
@AaronO what's is your idea for further integration? Ie. which parts of codebase could benefit the most from introducing `serde_v8` in its current form? 

I wonder if you could bring `serde_v8` into the tree for this PR so we could run it properly on CI.

AaronO:
@bartlomieju `serde_v8` is now in tree, CI pending.

> what's is your idea for further integration? Ie. which parts of codebase could benefit the most from introducing serde_v8 in its current form?

`serde_v8`'s main use-case is to optimize the op-ABI. This PR aimed to kickstart the conversation on `serde_v8` starting with some small and scoped changes.

Beyond the op-ABI, it could further be used in `core/bindings.rs` to strongly-type our core-scope and replace `set_func()` stuff, see my previous comment: https://github.com/denoland/deno/pull/9722#issuecomment-793489733

core/bindings.rs
```diff
@@ -470,16 +473,17 @@ fn eval_context(
...
-   output[0] = result
-   output[1] = ErrorInfo | null
-     ErrorInfo = {
-       thrown: Error | any,
-       isNativeError: boolean,
-       isCompileError: boolean,
-     }
-  */
+  #[derive(Serialize)]
+  struct Output<'s>(Option<serde_v8::Value<'s>>, Option<ErrInfo<'s>>);
```
bartlomieju, line 477: This is a bit surprising that it serializes to an array, but I guess one can thing about it as a tuple

serde_v8/src/ser.rs
```diff
@@ -0,0 +1,649 @@
...
+
+  fn end(self) -> JsResult<'a> {
+    Ok(self.v8_value.unwrap())
+  }
+}
+
+// Dispatches between magic and regular struct serializers
+pub enum StructSerializers<'a, 'b> {
+  Magic(MagicSerializer<'a, 'b>),
+  Regular(ObjectSerializer<'a, 'b>),
```
bartlomieju, line 230: What's the difference between those two? Maybe a comment here could help

serde_v8/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+pub fn v8_do(f: impl FnOnce()) {
+  static V8_INIT: Once = Once::new();
+  V8_INIT.call_once(|| {
+    v8_init();
+  });
+  f();
+  // v8_shutdown();
+}
```
bartlomieju, line 33: Move contents of this file to `serde_v8/tests`

serde_v8/src/ser.rs
```diff
@@ -0,0 +1,649 @@
...
+  }
+}
+
+macro_rules! not_reachable {
+    ($($name:ident($ty:ty, $lt:lifetime);)*) => {
+        $(fn $name(self, _v: $ty) -> JsResult<$lt> {
+            unreachable!();
+        })*
+    };
+}
```
bartlomieju, line 497: I believe this macro is superfluous, it would be more readable (although repetitive) to spell out those methods explicitly

serde_v8/tests/magic.rs
```diff
@@ -0,0 +1,57 @@
+use rusty_v8 as v8;
+
+use serde::{Deserialize, Serialize};
+
+use serde_v8::utils::{js_exec, v8_init, v8_shutdown};
+use std::convert::TryFrom;
+
+#[derive(Deserialize)]
+struct MagicOp<'s> {
```
bartlomieju, line 9: Does magic have some special meaning in this file?

serde_v8/src/ser.rs
```diff
@@ -0,0 +1,649 @@
...
+  type SerializeTuple = ArraySerializer<'a, 'b>;
+  type SerializeTupleStruct = ArraySerializer<'a, 'b>;
+  type SerializeTupleVariant =
+    VariantSerializer<'a, 'b, ArraySerializer<'a, 'b>>;
+  type SerializeMap = MapSerializer<'a, 'b>;
+  type SerializeStruct = StructSerializers<'a, 'b>;
+  type SerializeStructVariant =
+    VariantSerializer<'a, 'b, StructSerializers<'a, 'b>>;
+
+  forward_to! {
```
bartlomieju, line 333: Ditto, could you please remove this macro in favor of repeating method defintions?

bartlomieju (commented):
@AaronO could you please add copyrights to the files?

ry (approved):
LGTM!

Aaron, this is an amazing piece of code artistry. We've talked about this idea before, but never took it seriously. We thought it would be too hard - thought it wouldn't pay off. We were wrong on both accounts as you have demonstrated here.

I have benchmarked http_bench_json_ops on main and #9843 and see pretty massive improvements: https://gist.github.com/ry/c4008dc9d7a8a08bce1fdbad7d70e8e0

Once we replace JSON ops with serde_v8, this will be the single biggest performance improvement in Deno in two years. I suspect it will vastly simplify a lot of the juggling in deno_core.

It may not be very visible to users immediately, but this really changes the game.

PR #9730, opened 2021-03-09
fix(webgpu): ShaderModuleDescriptor code type

crowlKats:
Add Uint32Array type for `code` in `ShaderModuleDescriptor`, for which support was fixed in #9698


ry:
We really need better tests for webgpu. This is like driving blind folded.

ry (approved):

PR #9732, opened 2021-03-09
v1.8.1

bartlomieju:


cli/tests/integration_tests.rs
```diff
@@ -5254,6 +5254,7 @@ console.log("finish");
   }
 
   #[test]
+  #[ignore]
```
bartlomieju, line 5257: This test is super flaky on latest `main`; disabling to make sure release doesn't fail

ry (approved):

PR #9743, opened 2021-03-10
chore: remove unused import on non-windows

kitsonk:
Let's get rid of a very annoying warning.  Not sure why this isn't fatal in CI.


kt3k (approved):
LGTM

PR #9746, opened 2021-03-10
fix(lsp): allow on disk modules to change

### LINKED ISSUE #9348: lsp errors with "The current snapshot version is missing."

kitsonk:
Fixes #9348

Previously, it was assumed that only files open in the editor could change, which was obviously incorrect.  This allows files on disk (either local or remote cached modules) to change and for the LSP to be able to offer those changes to tsc.  This should fix most cases of the "current snapshot version missing" error.


cli/lsp/tsc.rs
```diff
@@ -1735,17 +1733,36 @@ pub fn request(
...
   use crate::lsp::documents::DocumentCache;
-
-  fn mock_state_snapshot(sources: Vec<(&str, &str, i32)>) -> StateSnapshot {
+  use crate::lsp::sources::Sources;
+  use std::path::Path;
+  use std::path::PathBuf;
+  use tempfile::TempDir;
+
+  fn mock_state_snapshot(
+    fixtures: Vec<(&str, &str, i32)>,
```
bnoordhuis, line 1745:
```suggestion
    fixtures: &[(&str, &str, i32)],
```
Then you can pass the argument by reference (`&arg`) without having to clone() it first.

(Not that it matters here performance-wise but it means I don't have to stop and think about what exactly gets cloned.)

cli/lsp/tsc.rs
```diff
@@ -2124,4 +2143,88 @@ mod tests {
...
+        1,
+      )],
+    );
+    let cache = HttpCache::new(&location);
+    let specifier_dep =
+      resolve_url("https://deno.land/x/example/a.ts").unwrap();
+    cache
+      .set(
+        &specifier_dep,
+        Default::default(),
```
bnoordhuis, line 2174:
```suggestion
        Url::default(),
```
Makes it immediately obvious what's being default-constructed. Ditto on line 2211.

bnoordhuis (approved):
Two suggestions that are really just that, suggestions. LGTM

PR #9758, opened 2021-03-11
fix(cil/eval): Fix eval command

### LINKED ISSUE #9733: 🐛 `deno eval ...` doesn't automatically re-compile/eval supplied CODE_ARG (Windows)
```shell
C:>ver

Microsoft Windows [Version 10.0.18363.1379]

C:>deno --version
deno 1.8.0 (release, x86_64-pc-windows-msvc)
v8 9.0.257.3
typescript 4.2.2
```

When serially executed from a Windows command line, `deno eval CODE_ARG` falls back to a previously successfuly compiled CODE_ARG.

```shell
C:>deno eval console.log(0)
0
C:>deno eval console.log(1)
0
```

This can be mitigated by using the `-T`/`--ts` or, recently, `--ext ts` flags which forces a check when CODE_ARG is new.

```shell
C:>deno eval --ext=ts console.log(0)
Check file:///C:/Users/Roy/OneDrive/Projects/deno/dexter/repo.deno.dexter/$deno$eval.ts
0

C:>deno eval --ext=ts console.log(1)
Check file:///C:/Users/Roy/OneDrive/Projects/deno/dexter/repo.deno.dexter/$deno$eval.ts
1

C:>deno eval console.log(2)
1
```

AkiaCode:
fix #9733

I remove `MediaType::JavaScript`because As a result of checking, only `MediaType::JavaScript` cannot confirm that the code has been changed.

Screenshot (fix): 
image

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -506,8 +506,6 @@ async fn eval_command(
   // Save our fake file into file fetcher cache
   // to allow module access by TS compiler.
   program_state.file_fetcher.insert_cached(file);
-
-
   debug!("main_module {}", &main_module);
   worker.execute_module(&main_module).await?;
   worker.execute("window.dispatchEvent(new Event('load'))")?;
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -496,8 +496,6 @@ async fn eval_command(
       MediaType::TypeScript
     } else if ext.as_str() == "tsx" {
       MediaType::TSX
-    } else if ext.as_str() == "js" {
-      MediaType::JavaScript
     } else {
       MediaType::JSX
     },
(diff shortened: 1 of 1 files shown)


cli/main.rs
```diff
@@ -496,10 +496,10 @@ async fn eval_command(
       MediaType::TypeScript
     } else if ext.as_str() == "tsx" {
       MediaType::TSX
-    } else if ext.as_str() == "js" {
-      MediaType::JavaScript
-    } else {
```
bartlomieju, line 501: This indeed seems fishy to me, it should check for `jsx` extension instead and default to `MediaType::JavaScript` in other cases

bartlomieju (commented):
Thanks for the PR @AkiaCode, please add a test case demonstrating the problem being fixed

cli/main.rs
AkiaCode: Only the JavaScript part in the cache was not recognized, so I deleted it, but there is a problem in the test....

PR #9764, opened 2021-03-12
fix(runtime): mitigate ability to 'stuff' permission prompts

### LINKED ISSUE #9750: Runtime permissions request should fully consume STDIN before asking for user input
A simple way to reproduce:
```typescript

await Deno.stdin.read(buf);
```

kitsonk:
Fixes #9750 

Reading out of stdin is blocking in Rust, so there is no easy way to "consume" any buffered stdin before prompting that I can conceive of.  So this PR does the following:

- It reads lines out of stdin when prompting for permission.  If any of those lines are read < 150ms it is assumed they were buffered and are discarded.
- It also does some checking on the line read, rejecting any input that isn't 1 char, so things like `long` would become invalid.

So to "exploit" it, you would have to type a single `g` into stdin, have the end user see a prompt, and have them not type a character and just press return after the prompt is displayed.

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -22,6 +22,8 @@ use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 #[cfg(test)]
 use std::sync::Mutex;
+#[cfg(not(test))]
+use std::time;
 
 const PERMISSION_EMOJI: &str = "⚠️";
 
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Is this a unix- and tty-only issue? If so, a solution is to flush the tty's input buffer before reading:
```rust
let rval = unsafe { libc::tcflush(libc::STDIN_FILENO, libc::TCIFLUSH) };
if rval == -1 {
  let err = std::io::Error::last_os_error();
  // ... - can really only be ENOTTY
}
```
runtime already depends on libc for other tty-related stuff.

kitsonk:
Closing in favour of #9813

PR #9771, opened 2021-03-12
feat(op_crates/webgpu): create instance only when required

crowlKats:

### DIFF
--- a/op_crates/webgpu/01_webgpu.js
+++ b/op_crates/webgpu/01_webgpu.js
@@ -14,8 +14,8 @@
   const eventTarget = window.__bootstrap.eventTarget;
 
   /**
-   * @param {any} self 
-   * @param {{prefix: string, context: string}} opts 
+   * @param {any} self
+   * @param {{prefix: string, context: string}} opts
    * @returns {InnerGPUDevice & {rid: number}}
    */
   function assertDevice(self, { prefix, context }) {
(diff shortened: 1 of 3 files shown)


op_crates/webgpu/lib.rs
```diff
@@ -184,6 +184,22 @@ fn deserialize_features(features: &wgpu_types::Features) -> Vec<&str> {
   return_features
 }
 
+pub fn op_webgpu_create_instance(
+  state: &mut OpState,
+  _args: Value,
+  _zero_copy: &mut [ZeroCopyBuf],
+) -> Result<Value, AnyError> {
+  check_unstable(&state, "navigator.gpu.requestAdapter");
```
lucacasonato, line 192:
```suggestion
  check_unstable(&state, "navigator.gpu");
```

crowlKats: whoops forgot that

op_crates/webgpu/01_webgpu.js
```diff
@@ -5012,8 +5012,16 @@
   }
   GPUObjectBaseMixin("GPUQuerySet", GPUQuerySet);
 
+  let instance;
+
   window.__bootstrap.webgpu = {
-    gpu: webidl.createBranded(GPU),
+    get gpu() {
```
lucacasonato, line 5018: I think this `get` might run at the wrong time... we "get" the `gpu` when assigning it to navigator, right? Have you tested this works?

crowlKats: this works, but thats because we `get` the navigator, so a little bit hacky. i can change it to something more appropriate

lucacasonato (approved):

PR #7228, opened 2020-08-28
refactor: Compiler config in Rust

bartlomieju:
Closes https://github.com/denoland/deno/issues/5606

PR #9774, opened 2021-03-12
fix(cli/module_graph): Set useDefineForClassFields to true

### LINKED ISSUE #9773: Deno bundle error with constructor property initialisation
Following example works with `deno run` but fails after bundling it with `deno bundle`.

**Example code:**
```typescript
export class Foo {
  bar = this.options.bar;
  constructor(
    protected options: { bar: string },
  ) {}
}

console.log(
  new Foo({ bar: "baz" }),
);
```

**Bundled code:**

```javascript
class Foo1 {
    bar = this.options.bar;
    constructor(options){
        this.options = options;
    }
}
console.log(new Foo1({
    bar: "baz"
}));
export { Foo1 as Foo };
```

**Error message:**
```
error: Uncaught TypeError: Cannot read property 'bar' of undefined
    bar = this.options.bar;
                       ^
    at Foo1.<instance_members_initializer> (file:///bundle.js:2:24)
    at new Foo1 (file:///bundle.js:3:16)
    at file:///bundle.js:7:13
```
**Version:**
```
deno 1.8.1 (release, x86_64-apple-darwin)
v8 9.0.257.3
typescript 4.2.2
```

nayeemrmn:
We decided to do this around https://github.com/denoland/deno/issues/7148#issuecomment-678754613, added `"useDefineForClassFields"` to the set of ignored compiler options in #7228 and documented it as set to true in #9196. Unless there was a regression, I don't think we ever actually set it to true.

Fixes #9773 (the example now produces an intuitive type error).

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -314,6 +314,7 @@ impl Inner {
       "noEmit": true,
       "strict": true,
       "target": "esnext",
+      "useDefineForClassFields": true,
     }));
     let (maybe_config, maybe_root_uri) = {
       let config = &self.config;
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -809,6 +809,7 @@ impl Graph {
       "strict": true,
       "target": "esnext",
       "tsBuildInfoFile": "deno:///.tsbuildinfo",
+      "useDefineForClassFields": true,
     }));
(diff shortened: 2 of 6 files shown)


kitsonk (approved):
LGTM

Thanks for figuring that out and finding the omission.

PR #9775, opened 2021-03-13
fix(op_crates/console): add dependency to serde derive

kitsonk:


lucacasonato:
We don't use serde derive in console op crate. See op_crates/console/lib.rs

AaronO:
Given that `deno_core` re-exports `serde` (see [core/lib.rs](https://github.com/denoland/deno/blob/10b99e8eb0e04e8340187b8aafe860405114d0d7/core/lib.rs#L27)) op_crates should probably use that.

kitsonk:
Not a fix

PR #9776, opened 2021-03-13
feat(cli/eval): remove eval flag T

AkiaCode:


crowlKats:
why? there is a comment to remove it for 2.0 as this would be a breaking change

AkiaCode:
Oh, Okay

PR #9777, opened 2021-03-13
feat(web): add AbortSignal.abort() static

benjamingr:
Implements https://github.com/whatwg/dom/pull/960/files in Deno to keep Deno's AbortSignal (relatively) spec compliant :)

Basically this adds a static method that creates an already aborted signal - similarly to `Promise.reject`


kitsonk:
This was landed, with issues, in #9907.  Sorry we didn't get around to looking at it in time (but the CI was failing so those do tend to get ignored)

PR #9781, opened 2021-03-14
chore: Upgrade rustyline to 8.0.0

caspervonb:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,7 +515,7 @@ dependencies = [
  "libc",
  "log",
  "lspower",
- "nix",
+ "nix 0.19.1",
  "notify",
  "os_pipe",
  "percent-encoding",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -1919,6 +1919,18 @@ dependencies = [
...
+name = "nix"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fa9b4819da1bc61c0ea48b63b7bc8604064dd43013e7cc325df098d49cd7c18a"
+dependencies = [
+ "bitflags",
+ "cc",
+ "cfg-if 1.0.0",
+ "libc",
+]
```
ry, line 1932: It’s unfortunate that this duplicates nix. Is it possible to dedup it?

bartlomieju (approved):
LGTM

caspervonb:
Was about to comment that  we can probably bump nix as it's a a non-breaking change for what we use it for, but @bartlomieju already on it 🔥

Issue #5666, opened 2020-05-20
TextEncoder doesn't exist on type of globalThis


bartlomieju:
```ts
// main.ts
console.log(globalThis.TextEncoder);
```

```shell
$ deno run main.ts
Compile file:///Users/biwanczuk/dev/deno/main.ts
error: TS2339 [ERROR]: Property 'TextEncoder' does not exist on type 'typeof globalThis'.
console.log(globalThis.TextEncoder);
                       ~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno/main.ts:1:24
```

I'd expect that `TextEncoder`/`TextDecoder` would be available on the `globalThis` type. 

Ref #5663

bartlomieju:
Still valid, `lib.deno.window.d.ts` and `lib.deno.worker.d.ts` type declarations should be updated to reflect those values.

nayeemrmn:
This issue isn't fixed for other symbols declared with `declare class`, which I think we decided to favour in the past. Seems like only `declare var` is accessible on `globalThis`, I think we need an upstream solution: https://github.com/microsoft/TypeScript/issues/39504.

kitsonk:
#9782 was reverted

bartlomieju:
Closing in favor of https://github.com/denoland/deno/issues/9957

PR #9782, opened 2021-03-14
fix(op_crates/web): align encoding type definitions to spec

caspervonb:
Fixes https://github.com/denoland/deno/issues/5666

### DIFF
--- a/op_crates/web/lib.deno_web.d.ts
+++ b/op_crates/web/lib.deno_web.d.ts
@@ -178,32 +178,84 @@ declare function atob(s: string): string;
  */
 declare function btoa(s: string): string;
 
-declare class TextDecoder {
-  /** Returns encoding's name, lowercased. */
-  readonly encoding: string;
-  /** Returns `true` if error mode is "fatal", and `false` otherwise. */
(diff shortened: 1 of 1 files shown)


op_crates/web/lib.deno_web.d.ts
```diff
@@ -178,32 +178,84 @@ declare function atob(s: string): string;
...
-  decode(input?: BufferSource, options?: { stream?: false }): string;
-  readonly [Symbol.toStringTag]: string;
+  /**
+     * Returns the value of ignore BOM.
+     */
+  readonly ignoreBOM: boolean;
+}
+
+interface TextDecodeOptions {
+  // TODO(caspervonb) support streaming.
```
kitsonk, line 220: This doesn't make sense in the type definitions, it should be located in the runtime code.

caspervonb: The old ad-hoc type definitions only allow `undefined | false` so kept that quirk because it was already there.
Am fine with removing it tho.

kitsonk (approved):
LGTM

Issue #9772, opened 2021-03-12
crypto.getRandomValues() inside worker throws invalid op id: expected type `rusty_v8::data::Integer`, got `rusty_v8::data::Value` 


searchableguy:
### Reproduction code

```typescript
// worker.ts
onmessage = () => {
    crypto.getRandomValues(new Uint8Array(16))
}
```

```typescript
// crash.ts
const worker = new Worker(new URL("./worker.ts", import.meta.url).href, {
    type: "module"
})

worker.postMessage(null)
```

### Error
```typescript
error: Uncaught (in worker "") (in promise) TypeError: invalid op id: expected type `rusty_v8::data::Integer`, got `rusty_v8::data::Value`
    crypto.getRandomValues(new Uint8Array(16))
           ^
    at dispatch (deno:core/core.js:174:12)
    at Object.jsonOpSync (deno:core/core.js:235:17)
    at Object.getRandomValues (deno:op_crates/crypto/01_crypto.js:40:10)
    at onmessage (file:///home/user/projects/deno/deno_bot/worker.ts:2:12)
    at workerMessageRecvCallback (deno:runtime/js/99_main.js:91:35)
    at file:///home/user/projects/deno/deno_bot/__anonymous__:1:1
error: Uncaught (in promise) Error: Unhandled error event reached main worker.
    at Worker.#poll (deno:runtime/js/11_workers.js:288:21)
```

caspervonb:
Workaround:
```js
const worker = new Worker(new URL("./worker.ts", import.meta.url).href, {
    type: "module",
    deno: true,
})
```

PR #9784, opened 2021-03-14
fix(runtime): do not require deno namespace in workers for crypto

caspervonb:
Fixes https://github.com/denoland/deno/issues/9772


caspervonb:
Should have been caught by wpt but we aren't running worker tests at the moment so added a basic smoke test in the current crypto worker test.

bartlomieju (approved):
LGTM

Issue #9603, opened 2021-02-25
[Feat-Tests]: Requesting Method To Force Deno Tests To Exit Once Complete


Skillz4Killz:
### Use Case

I want to shut off everything I enabled during testing like the WebSockets. `ws.close()` doesn't work for me, because we always reopen on close to keep bots online 24/7. We can't add custom handling for a specific code, because out of nowhere the API could begin to use that code and suddenly, we have every bot out there breaking. This potential case is not worth having just for unit tests. In my opinion, there should be a cleaner way to do this.

The `deno test` process does NOT exit if it has something keeping it open like a ws connection. I don't believe the intention behind `deno test` is to stay on forever.

### Potential Solution

- Internally call `Deno.exit(0)` at the end of the test command.
- Allow a callback to be run when all tests complete for example `Deno.onTestsCompleted = () => {}` or `deno tests --onComplete=file.ts *`

### Current Workaround

```ts
// Forcefully exit the Deno process once all tests are done.
Deno.test({
  name: "exit the process forcefully after all the tests are done\n",
  fn() {
    Deno.exit();
  },
});
```

Problem with current workaround, is that when I put deno test in a GHA, it always "succeeds" even when the tests fail. It's not a huge deal breaker but it is something I would prefer done properly.

Skillz4Killz:
Not stale :)

caspervonb:
I'll try to squeeze this in for the coming release.

EliteScientist:
I think my issue is related to this. Some of the unit test cause a part of my system to start up that has WebWorkers. This system needs to be instructed to shutdown. Due to this system running Deno fails to exit after the test complete.

I would add a test to shut it down but I cannot control the order in which the test execute. I have a different test file for different parts of the system.

It would be nice to have a callback or event dispatched when the test execution is complete so that I can shutdown the system. 

I think we are missing the before and after callbacks for the unit test framework.

lucacasonato:
Ref https://github.com/denoland/deno/discussions/10771

Issue #8713, opened 2020-12-10
Calling Deno.exit most probably will prevent coverage collection


caspervonb:
Since Deno.exit calls exit(2) directly, coverage will probably not be collected in that scenario. We'd have to set up an atexit equivalent hook.

bartlomieju:
@caspervonb is this still actionable given we now have `exitSanitizer`?

caspervonb:
> @caspervonb is this still actionable given we now have `exitSanitizer`?

Oh completely forgot about this.
Yeah it still applies to integration tests that are run with `deno run` via `Deno.run`.

bartlomieju:
> Yeah it still applies to integration tests that are run with deno run via Deno.run.

So we'd have to add some kind of call into `op_exit` to prepare a flush of collected coverage?

caspervonb:
> > Yeah it still applies to integration tests that are run with deno run via Deno.run.
> 
> So we'd have to add some kind of call into `op_exit` to prepare a flush of collected coverage?

Yes, altho would be preferable to have something like `atexit(1)` in Rust that can take a closure, that way we can keep it out of the ops.

bartlomieju:
> > > Yeah it still applies to integration tests that are run with deno run via Deno.run.
> > 
> > 
> > So we'd have to add some kind of call into `op_exit` to prepare a flush of collected coverage?
> 
> Yes, altho would be preferable to have something like `atexit(1)` callbacks in Rust that can take a closure, that way we can keep it out of ops.

Not sure if that would work - you still need to access `ProgramState` and `Worker` to finish coverage collection?

I think this might be a non-issue and working as intended unless Node can collect coverage if it process exits.

bartlomieju:
Related to https://github.com/denoland/deno/issues/12888

PR #9785, opened 2021-03-14
fix(cli): close test runner once all tests complete

caspervonb:

### DIFF
--- a/cli/tools/test_runner.rs
+++ b/cli/tools/test_runner.rs
@@ -89,6 +89,8 @@ pub fn render_test_file(
     options
   ));
 
+  test_file.push_str("close();\n");
+
   test_file
 }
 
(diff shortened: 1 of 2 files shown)


cli/tools/test_runner.rs
```diff
@@ -89,6 +89,14 @@ pub fn render_test_file(
     options
   ));
 
+  test_file.push_str(
+    r"
+if ('close' in globalThis) {
+  globalThis.close();
+}
+    ",
+  );
```
bartlomieju, line 98: I don't think this is correct; tests are not run in worker context

caspervonb:
Blocked on https://github.com/denoland/deno/issues/8713

cli/tools/test_runner.rs
caspervonb: Wasn't sure if unit tests invoked the manifest or not (seemingly not).

caspervonb:
Incompatible with #9815, needs to be rewritten.
Closing without merge.

PR #9791, opened 2021-03-15
docs(testing): Add custom test example

getspooky:
While Deno comes with powerful [assertions modules](https://deno.land/std@$STD_VERSION/testing/asserts.ts)  but
there is always something specific to the project you can add. Creating `custom assertion function` can improve readability and reduce the amount of code.


ry (approved):
Thanks

PR #9793, opened 2021-03-15
chore: deprecate Deno.Buffer and read/write utils

lucacasonato:
This commit marks the `Deno.Buffer` / `Deno.readAll` /
`Deno.readAllSync` / `Deno.writeAll` / `Deno.writeAllSync` utils as
deprecated, and schedules them for removal in Deno 2.0. These
utilities are implemented in pure JS, so should not be part of the
Deno namespace.

These utilities are now available in std/io/buffer and std/io/util:
https://github.com/denoland/deno_std/pull/808.

This additionallty removes all internal dependance on Deno.Buffer.

Towards #9795

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -405,9 +405,9 @@ declare namespace Deno {
    *
    * ```ts
    * const source = await Deno.open("my_file.txt");
-   * const buffer = new Deno.Buffer()
    * const bytesCopied1 = await Deno.copy(source, Deno.stdout);
-   * const bytesCopied2 = await Deno.copy(source, buffer);
+   * const destination = await Deno.create("my_file_2.txt");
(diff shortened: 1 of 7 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -405,9 +405,9 @@ declare namespace Deno {
    *
    * ```ts
    * const source = await Deno.open("my_file.txt");
-   * const buffer = new Deno.Buffer()
    * const bytesCopied1 = await Deno.copy(source, Deno.stdout);
-   * const bytesCopied2 = await Deno.copy(source, buffer);
+   * const destination = await Deno.open("my_file_2.txt");
```
bartlomieju, line 409: This will open file in read-only mode

lucacasonato: Switched to `Deno.create`

ry (approved):

PR #9796, opened 2021-03-15
refactor(lsp): slightly reorganize diagnostics debounce logic

piscisaureus:
This patch doesn't actually fix the bug I was hoping to fix, which is
that `update_diagnostics()` sometimes gets called even when there are
more updates that should be processed first. I did eventually figure out
that this issue is caused by Tokio's cooperative yielding, which
currently can't be disabled.

However overall it makes the debounce code somewhat more readable IMO,
which is why I'm suggesting to land it anyway.

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -24,12 +24,9 @@ use std::sync::Arc;
 use std::thread;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
-use tokio::time;
-
-// 150ms between keystrokes is about 45 WPM, so we want something that is longer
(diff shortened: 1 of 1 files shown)


bartlomieju:
> I did eventually figure out
that this issue is caused by Tokio's cooperative yielding, which
currently can't be disabled.

I think this is now possible in Tokio 1.3: https://github.com/tokio-rs/tokio/pull/3547

kitsonk (approved):
LGTM, yes agree it is more readable.

piscisaureus:
> I think this is now possible in Tokio 1.3: https://docs.rs/tokio/1.3.0/tokio/task/index.html#unconstrained

I would like to follow up on this, but currently I don’t have the time, so landing it as-is for now.

PR #9798, opened 2021-03-15
feat(doc): update example for sub processes

### LINKED ISSUE #9790: [Docs] Subprocess example leaves pipes open
Related Doc: https://github.com/denoland/deno/blob/v1.8.1/docs/examples/subprocess.md

In a script I'm writing, I need to run tens of thousands of subprocesses via `Deno.run`, but I noticed that after the first ~4096 commands, I'd always get a `Bad file descriptor (os error 9)` error.

After some investigation, I found out that Deno leaves pipes open unless you read their contents with `process.output()` or `process.stderrOutput()`, so there's a bunch of unused open pipes otherwise.

It might be good to mention that in the docs, maybe with an updated example like this:

```ts
/**
 * subprocess.ts
 */
const fileNames = Deno.args;

const p = Deno.run({
  cmd: [
    "deno",
    "run",
    "--allow-read",
    "https://deno.land/std@$STD_VERSION/examples/cat.ts",
    ...fileNames,
  ],
  stdout: "piped",
  stderr: "piped",
});

const { code } = await p.status();

// Reading the outputs closes their pipes
const rawOutput = await p.output();
const rawError = await p.stderrOutput();

if (code === 0) {
  await Deno.stdout.write(rawOutput);
} else {
  const errorString = new TextDecoder().decode(rawError);
  console.log(errorString);
}

Deno.exit(code);
```

makuzaverite:
fix #9790

### DIFF
--- a/docs/examples/subprocess.md
+++ b/docs/examples/subprocess.md
@@ -70,11 +70,13 @@ const p = Deno.run({
 
 const { code } = await p.status();
 
+// Reading the outputs closes their pipes
+const rawOutput = await p.output();
+const rawError = await p.stderrOutput();
+
 if (code === 0) {
-  const rawOutput = await p.output();
   await Deno.stdout.write(rawOutput);
 } else {
-  const rawError = await p.stderrOutput();
   const errorString = new TextDecoder().decode(rawError);
   console.log(errorString);
 }


ry (approved):
Thanks

Issue #9738, opened 2021-03-09
Improving op_crate API and registering ops


AaronO:
Sharing some thoughts about improvements to op_crates, some previously shared on Discord.

## Goals

1. Make `op_crates` plug-and-play / self-contained, reduce complexity leaked to consumers.
2. Streamline lifecycles of declared resources without consumer having to coordinate between snapshot or not scenarios
3. Extend what `op_crates` can do, e.g: provide their own v8 funcs, external refs, ...

## Current state

`op_crates` have a convenience `init` function to execute all JS in the runtime, all ops are exported individually as `op_*` funcs.

So their main "API" by convention today is `::init()` + N `::op_*`

## Problems

**In short:** `op_crates` aren't plug-and-play, op-middleware is not reusable and no-access to v8 for optimisations, etc...

- `init` is somewhat of a misnomer since it doesn't init the entire module, it only inits the JS (ops need to be registered separately)
- ops need to be registered individually by the consumer (like deno does in `runtime/*`), leaking complexity to the consumer
    - Prevents op_crates from being "plug-and-play"
    - Adding new ops in a crate without coordinating with the consumer breaks the consumer (same with removing/renaming)
    - The main pro of the current approach is that it's granular and gives the consumer control
    - Reading the code it appears that the current approach is mainly motivated by a need for `metrics_op` to wrap op setup
- There's currently no convention for exposing v8 external refs (e.g: for https://github.com/denoland/deno/pull/9713)
- There's currently no convention for adding non-op native v8 functions (which is intentionally discouraged by design but can be limiting in some cases). It wouldn't be best-practice (lose inspectability of ops, metrics ...) but I think it wouldn't be a huge cost to allow it following certain conventions. A few use-cases:
    - Could allow moving `TextEncoder`'s encoding indexes off the heap and out of JS code, it currently consumes `~500kb` per isolate between the two. If v8 bindings were allowed, it could be done via external arrays, preserving logic in JS and simply moving the arrays to rust instead of the entire implementation
    - Functions like `.heapStats()` are inherently low-level and need access to the isolate. This requires them to go into `core/` or limits consumers from adding their own
    - v8 bindings could help provide optimized versions of `URL` and other core elements (see https://github.com/denoland/deno/pull/9276#issuecomment-773865387 for perf gap)
[...]

PR #9713, opened 2021-03-06
optimize(core): move internal js src off heap

AaronO:
This PR explores moving the internal JS code (bundled in `deno`) off the heap, as suggested by https://github.com/denoland/deno/issues/9675

## Impact

Reduce per-isolate heap by `~1MB`, reducing overall memory usage (especially with many workers), with no real downside or compromise.

This should also reduce snapshot sizes, resulting in faster snapshot restores and a smaller `deno` binary (these will likely be minor and weren't the core motivation)

## Subtasks

- [ ] ASCII JS code
  - [x] Ensure internal JS files are ASCII
  - [ ] Add a lint / CI test check to ensure files remain ASCII
- [x] Allow executing JS code via external `v8::String`s
  - [x] Implement external strings in `rusty_v8` (https://github.com/denoland/rusty_v8/pull/641)
  - [x] Implement `JsRuntime::execute_static`
- [ ] Work with snapshots
  - [ ] Implement support for string `ExternalReferences` in v8, so they can be captured and restored in snapshots
  - [ ] Register internal src references with the snapshot creator (maybe as they're passed through `execute_static` ?)

## Notes

Non-ASCII files found with (ignores test and example files):
```
find core runtime op_crates -name "*.[jt]s" | grep -v "examples" | grep -v "_test" | xargs file | grep -v ASCII
```

Non-ASCII characters in each file were then found with this regex: `[^\x00-\x7F]`

Non-ASCII files and context:
- `runtime/js/02_console.js`: the non-ASCII chars are the tableChars (L38)
- `op_crates/web/02_event.js`: there's a single utf8 apostrophe `’` vs `'` in comments (L919)
- `op_crates/web/21_filereader.js`: there are multiple utf8 apostrophes in comments (same as above)
- `op_crates/fetch/26_fetch.js`: same deal, same apostrophes
- `op_crates/web/lib.deno_web.d.ts`: used `§` in comment

Issue #9675, opened 2021-03-03
Offload source strings from JS heap by using external v8 strings


AaronO:
This is a specific optimization related to https://github.com/denoland/deno/issues/9660

## The problem

Source strings (of runtime JS) take up a decent portion of our baseline-heap, they account for `~1MB` of the total `~2.2MB` as of `1.8.0`, so nearly 50% and is duplicated for each isolate/worker.

## The solution

All runtime JS is bundled in the `deno` binary and available as `&'static str` so they are inherently immutable, live "forever" and are thus perfect candidates for external strings.

We could simply implement `JsRuntime::execute_static`, a version of `JsRuntime::execute` that accepts static filenames and sources and exposes them as external strings to v8.

External strings could possibly be used more broadly, but I think it's best to start with the internal runtime JS.

## The impact

We should be able to reduce baseline heap-size by `~1MB` or ~50% with little to no downsides (unlike minifying).

## The hurdles

This requires extending our `rusty_v8` bindings, which is relatively simple (we need to bind to `v8::String::NewExternal`, expose it twice for one and two-byte strings)

We would also need to ensure that our runtime source files are all truly ASCII (not utf8) if we want to use single-byte strings.

Check if this is compatible with snapshots or what would be required to correct that (from inspecting the heap retainers, all source strings are retained by `script_or_debug_info` of sub-functions, snapshots without the external strings might result in dangling references that would need to be restored) .

bartlomieju:
We have already done it, so I'm gonna close this for now.

PR #9800, opened 2021-03-15
core: introduce extensions

AaronO:
This PR implements ideas discussed in https://github.com/denoland/deno/issues/9738, specifically the concept of `JsRuntimeModule`s which makes modular runtime features plug-n-play.

## Goals

Reduce complexity in `runtime`, make modules self-contained and streamline snapshotting, etc...

This should make many parts of `deno` simpler and reusable by third-parties and reduce the implicit complexity in bootstrapping JsRuntimes (when snapshotting or not ...). `op_crates` should now be easily usable without `runtime`

## Subtasks

- [x] Core `JsRuntimeModule` foundations/traits (`core/runtime_modules.rs`)
- [x] Implement `JsRuntime` init via modules
- [x] Refactor all `op_crates` to be `::init() -> JsRuntimeModule`
      (essentially move code from `runtime/ops/*.rs` to `op_crates/*/lib.rs`)
- [x] Introduce `include_js_files!()` and `declare_ops!()` macros
- [x] Refactor module consumers (places where we init `JsRuntime`s & ops to use these new modules)
- [x] Implement `OpMiddleware`
  - [x] `OpMetrics` (replacing `op_metrics`)
  - [ ] Maybe implement `OpTracing` middleware (`--optrace` flag)
  - [ ] Maybe turn the `UnstableChecker` into an OpMiddleware
- [ ] Make `deno_runtime` use JsRuntimeModules ? 
- [ ] Explore declarative plugins ? (a subset of the JsRuntimeModule interface)
- [ ] Review new abstractions (naming, fit, etc...)
- [ ] Maybe add a `web_modules()` helper function that inits all Web-API modules (url, fetch, ... essentially `op_crates` today)
- [ ] Maybe add a `--no-snapshot` flag to allow running without snapshots (not sure if useful at all, even when debugging, besides https://github.com/denoland/deno/pull/9713)

## Follow-ups

- [ ] Restore extension builder pattern (more user friendly)
- [ ] Restore `declare_ops!()` macros
- [ ] Restore `MultiModule` abstraction, thus reverting Extension to a Trait
- [ ] Use extensions in `runtime`
- [ ] Use extensions in `cli`
- [ ] Allow passing extra extensions to `Worker` and `WebWorker`, making them more reusable and composable
- [ ] UnstableChecker OpMiddleware for fully unstable ops
- [ ] Tracing OpMiddleware for debugging
- [ ] Plugins (might only be a subset of the full trait)  ...
- [ ] Allow running `Worker` and `MainWorker` without snapshots, good for debugging and in general snapshots should be optional and viewed as a graceful enhancement (perf boost)


core/runtime_modules.rs
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+
+
+// MultiModule allows grouping multiple sub-JsRuntimeModules into one,
+// allowing things such as:
+// ```
+// fn web_modules(args: WebModuleArgs) -> MultiModule {
+//  MultiModule::new(vec![deno_url::init(), deno_fetch::init(...), ...])
+// }
```
ry, line 135: Interesting!

ry:
Nice! I fully support this effort. 

If you can get the existing op crates working and the tests pass, I'm rather confident that this will generalize to future needs. I think it will be challenging tho.

runtime/build.rs
```diff
@@ -44,8 +40,28 @@ fn create_snapshot(
...
+      ),
+    ),
+    Box::new(deno_crypto::init(None)),
+    Box::new(deno_webgpu::init(false)),
+  ];
+
   let js_runtime = JsRuntime::new(RuntimeOptions {
     will_snapshot: true,
+    modules: deno_modules,
     ..Default::default()
```
ry, line 65: Very clean - I like it

core/runtime_modules.rs
```diff
@@ -0,0 +1,206 @@
...
+    ),)+]
+  };
+
+  (with($wrapper:path), $($opfn:ident,)+) => {
+    vec![$((
+      stringify!($opfn),
+      $wrapper($opfn),
+    ),)+]
+  };
+}
```
ry, line 192: I'm a bit wary of introducing macros. Although it reduces code, I worry that it obfuscates what's happening. That said, these seem relatively simple... Maybe consider doing these in a follow up PR.

core/runtime_modules.rs
```diff
@@ -0,0 +1,206 @@
...
+
+  pub fn pure_js(js_files: Vec<SourcePair>) -> Self {
+    Self::new(Some(js_files), None, None)
+  }
+
+  pub fn with_ops(
+    js_files: Vec<SourcePair>,
+    ops: Vec<OpPair>,
+    opstate_fn: Option<Box<OpStateFn>>,
+  ) -> Self {
```
ry, line 78: Seems superfluous. Does this offer much more than `new()`? Using a couple fewer `Some` wrappers is offset by the mental overhead of having to understand all these different constructors.

core/runtime_modules.rs
```diff
@@ -0,0 +1,206 @@
...
+    Self {
+      js_files,
+      ops,
+      opstate_fn,
+    }
+  }
+
+  pub fn pure_js(js_files: Vec<SourcePair>) -> Self {
+    Self::new(Some(js_files), None, None)
+  }
```
ry, line 72: This one is nice

core/runtime.rs
```diff
@@ -81,6 +83,7 @@ pub struct JsRuntime {
   snapshot_creator: Option<v8::SnapshotCreator>,
   has_snapshotted: bool,
   allocations: IsolateAllocations,
+  modules: Vec<Box<dyn JsRuntimeModule>>,
```
ry, line 86: I think we should use a different word than "module" because it might be confused with ES modules.

Currently we use the term "op_crates" but maybe we don't want to explicitly tie this to the rust crate concept. Maybe something like "op_package" ?

lucacasonato: Maybe a RuntimeExtension?

AaronO: Both are good candidates, I prefer `RuntimeExtension` but `OpBundle` isn't bad either. `OpBundle` is a little more specific and familiar given "op_crates", but `RuntimeExtension` is a better fit since runtime extensions can be pure JS too.

I could easily imagine the following in the docs:
> Runtime extensions implement the `deno_core::JsRuntimeExtension` interface(/trait), they extend the runtime's features by adding JS to the global scope and providing ops. Extensions are typically pure JS, a mix of JS & matching ops or stateful-ops & op-middleware in more advanced cases.

I'll continue with the current naming and reword on the "polish pass" once we've reached consensus.

ry: RuntimeExtension sounds good to me

bartlomieju: I'm strongly in favor of `op_crates`, this term is already prevalent

lucacasonato: But not all op_crates contain ops 😉 (op_crates/web)

bartlomieju: Sure... I just want to avoid situation where the same thing has two distinct names

AaronO: `op_crates` isn't a bad name but it's a little too specific IMO:
- Extensions can be pure JS
- Extensions can be grouped (`runtime` will have many small op-extensions grouped with MultiExtensions)
- Plugins aren't op_crates but should return extensions (in the future)

There's no "perfect" candidate suggested so far, all have tradeoffs on familiarity/ambiguity/descriptiveness.

Obviously we could rename `op_crates` to `ext_crates` but that could be unnecessarily disruptive...

core/runtime.rs
```diff
@@ -349,6 +357,36 @@ impl JsRuntime {
...
+    let mut modules: Vec<Box<dyn JsRuntimeModule>> =
+      self.modules.drain(..).collect();
+    let selfish = Rc::new(RefCell::new(self));
+    for m in modules.iter_mut() {
+      let js_files = m.init_js()?;
+      for (filename, source) in js_files {
+        selfish.borrow_mut().execute_static(filename, source)?;
+      }
+    }
+    selfish.borrow_mut().modules = modules;
```
bartlomieju, line 372: @AaronO could you explain the reason for "taking" modules, wrapping self in `Rc<RefCell<>>` and then putting modules back?

core/runtime.rs
```diff
@@ -180,6 +183,10 @@ pub struct RuntimeOptions {
   /// executed tries to load modules.
   pub module_loader: Option<Rc<dyn ModuleLoader>>,
 
+  /// JsRuntime modules, not to be confused with ES modules
+  /// these are deno modules containing JS + ops
+  pub modules: Vec<Box<dyn JsRuntimeModule>>,
```
bartlomieju, line 188: These names should change as well - `JsRuntimeModule` cannot contain ES modules as source files, only plain old scripts, because currently snapshotting doesn't support ES modules.

core/runtime.rs
```diff
@@ -411,6 +449,40 @@ impl JsRuntime {
     }
   }
 
+  // same as execute() but for internal JS code
+  fn execute_static(
```
bartlomieju, line 453: Seems superfluous, can't `execute` be used?

core/runtime_modules.rs
```diff
@@ -0,0 +1,206 @@
...
+  fn init_js(&self) -> Result<Vec<SourcePair>, AnyError> {
+    Ok(match &self.js_files {
+      Some(files) => files.clone(),
+      None => vec![],
+    })
+  }
+
+  fn init_ops(&mut self, registrar: RcOpRegistrar) -> Result<(), AnyError> {
+    // NOTE: not idempotent
+    // TODO: fail if called twice ?
```
bartlomieju, line 93: Yes it should, I think `panic!` would be fine here

core/runtime_modules.rs
```diff
@@ -0,0 +1,206 @@
...
+  // register_json_op_sync(...)
+  // register_json_op_async(...)
+}
+
+////
+// Helper macros to reduce verbosity / redundant decls
+////
+#[macro_export]
+macro_rules! include_js_files {
+  (root $root:literal, $($file:literal,)+) => {
```
bartlomieju, line 167: How about renaming `root` to `prefix`? My first thought was that it tell a root path to resolve files, while actually it's the prefix that will be added to declared paths when executing files.

core/runtime.rs
hayd: Why not `Extension` (drop the `Runtime`)? and potentially op_crates->extensions later.

core/runtime_modules.rs
bartlomieju: An example in the docstring would be helpful

core/runtime_modules.rs
bartlomieju: I agree, I'm a bit wary of the `with()` middleware wrapper. This PR already introduces quite a few new concepts, I suggest to skip middleware implementation for a follow up PR.

op_crates/fetch/lib.rs
```diff
@@ -409,30 +419,48 @@ where
...
+    Some(ca_data.as_bytes().to_vec())
   } else if let Some(ca_file) = ca_file {
     let mut buf = Vec::new();
     File::open(ca_file)?.read_to_end(&mut buf)?;
-    let cert = reqwest::Certificate::from_pem(&buf)?;
+    Some(buf)
+  } else {
+    None
+  };
+  create_http_client("".to_owned(), ca_data_vec)
```
bartlomieju, line 443: This sets `User-Agent` to an empty string, which doesn't seem right. Why split into `create_http_client` two functions?

runtime/build.rs
```diff
@@ -44,8 +40,28 @@ fn create_snapshot(
...
+    )),
+    Box::new(
+      deno_websocket::init::<deno_websocket::NoWebSocketPermissions>(
+        "".to_owned(),
+        None,
+      ),
+    ),
+    Box::new(deno_crypto::init(None)),
+    Box::new(deno_webgpu::init(false)),
+  ];
```
bartlomieju, line 60: I have some reservations about this - it requires to provide some state during snapshotting process which wasn't necessary before; it doesn't look so bad, but I can imagine situation with 20 crates and providing dummy values in such case might get complicated. Maybe there should be separate methods for obtaining JS sources and initializing the state?

core/runtime.rs
AaronO: To avoid double-borrows, I've simplified it to remove the self wrap.

core/runtime.rs
AaronO: It can, but I wanted to experiment with https://github.com/denoland/deno/issues/9675, which introduces `execute_static` and passes those source strings as external v8 strings.

op_crates/fetch/lib.rs
AaronO: @bartlomieju There are already two duplicate`create_http_client` functions in `main`:
1. [runtime/http_util.rs](https://github.com/denoland/deno/blob/20627c91364d2a992fdfaaad7c8ae86454dbc2ed/runtime/http_util.rs#L13)
2. [op_crates/fetch/lib.rs](https://github.com/denoland/deno/blob/20627c91364d2a992fdfaaad7c8ae86454dbc2ed/op_crates/fetch/lib.rs#L412)

Both are broadly used for the fetch ops, the first is put into OpState the latter is used in `op_create_http_client`. So even in `main`, `op_create_http_client` doesn't allow setting the user-agent nor does it appear to use the "global" one from the OpState. `op_create_http_client` doesn't appear to be used anywhere so I was considering stripping it out or fixing it so it shares the user-agent.

core/runtime.rs
AaronO: Yes, that's pretty clear, it's linked to the above naming conversation, I'll change the names once there's some consensus.

op_crates/fetch/lib.rs
lucacasonato: @AaronO `op_create_http_client` is used for `Deno.createHttpClient`.

runtime/build.rs
AaronO: This is indeed a small quirk and could be avoided, since JS provided by extensions isn't dependant on input args (currently and probably shouldn't). Both ops and opstate (and middleware once added) depend on args.

Splitting JS from ops would break the conceptual "bundle" that extensions provide.

Do you have a better idea on how to do this without breaking/leaking the extension abstraction ?

Currently I think dummy-args are a lesser evil, not ideal but better than a leaky abstraction.

Part of the goal of extensions, is to streamline snapshotting, so 3rd parties can easily create their own snapshots with relative ease.

op_crates/fetch/lib.rs
AaronO: @lucacasonato Yep, sorry if I was unclear. I meant that I didn't see any calls to `Deno.createHttpClient` in our code or elsewhere (only [results on sourcegraph](https://sourcegraph.com/search?q=Deno.createHttpClient&patternType=literal) are deno's own tests).

But I would still argue that it's "incorrect" and should either accept a `userAgent` arg or default to the one provide to the default client.

lucacasonato: It is indeed incorrect - the default user agent should be the same as the default http client. If it is not, that is a bug. Users can already specify a custom user agent by specifying the `User-Agent` header on requests.

AaronO: I created an issue and provided a reproducible example here: https://github.com/denoland/deno/issues/9827, so it is confirmed, I'll fix it.

I do think being able to set `User-Agent` for a client would be convenient, though it can be worked around as you said, it's usually the kind of thing that is provided by the client/user-agent.

core/runtime_modules.rs
AaronO: @ry @bartlomieju What about something like:

```rust
    declare_ops!(json_op_async[
      op_webgpu_request_adapter,
      op_webgpu_request_device,
    ]),
```

That's a little easier to read, it conveys that `json_op_async` wraps `[...]` and hints that it returns a `Vec`

core/runtime_modules.rs
AaronO: @ry Not really, especially given that `vec![]` isn't terribly more verbose than `None` so `js_files` and `ops` could just be `Vec`s and not `Option`s in the struct.

I might introduce a builder pattern, something like:
```rust
BasicExtension::builder()
   .js(...)
   .ops(...)
   .state(move || { ... })
   .ops(...) // More ops to group by wrapper and "avoid" `declare_ops_group` ...
   .build()
```

core/runtime_modules.rs
AaronO: Thanks !

core/runtime_modules.rs
AaronO: It could also be:

```rust
    declare_ops!(json_op_async, [
      op_webgpu_request_adapter,
      op_webgpu_request_device,
    ]),
```

If we want it to look more like a map expression (notice `, [`)

core/runtime_modules.rs
AaronO: @bartlomieju Are these good enough ? https://github.com/denoland/deno/blob/dcdd86c74eea328d6fdad78d5ffef7e65fb6d004/core/runtime_modules.rs#L242-L303

(also renamed `root` => `prefix`)

AaronO:
@lucacasonato @bartlomieju I've implemented `OpMiddleware`:
- I added a [unit test](https://github.com/denoland/deno/blob/dcdd86c74eea328d6fdad78d5ffef7e65fb6d004/cli/tests/unit/metrics_test.ts#L70-L80) for it
- You can see the builder pattern in use in [metrics::init()](https://github.com/denoland/deno/blob/dcdd86c74eea328d6fdad78d5ffef7e65fb6d004/runtime/metrics.rs#L13-L24)
- Metrics is now passed as an extension in [worker.rs](https://github.com/denoland/deno/blob/dcdd86c74eea328d6fdad78d5ffef7e65fb6d004/runtime/worker.rs#L95-L96) and [web_worker.rs](https://github.com/denoland/deno/blob/dcdd86c74eea328d6fdad78d5ffef7e65fb6d004/runtime/web_worker.rs#L187-L188)
- So all the ops in the `op_crates` now go through this "middleware"

Middleware is important for this first iteration, since without it, op_crate ops' metrics would silently break (and it demonstrates that runtime extensions are general enough to cover our needs)

op_crates/fetch/lib.rs
AaronO: Fixed by https://github.com/denoland/deno/pull/9830

core/runtime.rs
rektide: is there a v8 bug ticket for ES modules not being snapshottable? it would be nice if support was added. i searched v8 & chromium bug trackers but didn't turn up anything.

ry:
@AaronO I've done some renames and removed BasicModuleBuilder and MultiModule. We can add those in the future if necessary.

The Extension trait only has one implementor. Is it useful to even make it a trait? If we can express all op crates using the concrete BasicExtension struct, then maybe we should just rename that to Extension.

AaronO:
It appears we'll be landing a smaller scoped version of this (dropping the trait, `MultiModule` etc...)

For posteriority's sake here are the anticipated follow-ups:
1. Switch `deno_runtime` over to extensions
2. Maybe re-introduce `MultiModule` and the `Extension` trait (maybe re-introduce builder, etc...)
3. Switch plugins over to use this new declarative API
4. Use them to streamline snapshots and allow 3rd-parties to re-use `MainWorker` etc... with custom snapshots and custom runtime-extensions
5. An Unstable OpMiddleware to wrap ops that are fully unstable (NOT those that are conditionally/partially unstable)
6. A tracing OpMiddleware wrapping all ops to allow tracing similar to `strace` / `dtruss`

AaronO:
@ry https://github.com/denoland/deno/pull/9800/commits/ee50474397550566d40232048702a5e81af81bae downgrades the trait to the single concrete type.

Whilst it makes sense to drop the trait/interface in this limited scope, I'm fairly certain we'll have to restore it in the future

core/extensions.rs
```diff
@@ -0,0 +1,195 @@
...
+
+  /// Called at JsRuntime startup to initialize ops in the isolate.
+  pub fn init_ops(&mut self, registrar: RcOpRegistrar) -> Result<(), AnyError> {
+    // NOTE: not idempotent
+    // TODO: fail if called twice ?
+    if let Some(ops) = self.ops.take() {
+      for (name, opfn) in ops {
+        registrar.borrow_mut().register_op(name, opfn);
+      }
+    }
```
bartlomieju, line 69: Again, I think it should panic if called second time

core/extensions.rs
```diff
@@ -0,0 +1,195 @@
...
+
+// include_js_files! helps embed JS files in an extension
+// Example:
+// ```
+// include_js_files!(
+//   prefix "deno:op_crates/hello",
+//   "01_hello.js",
+//   "02_goodbye.js",
+// )
+// ```
```
bartlomieju: Should use triple slash to provide docstring

core/extensions.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// declare_ops! helps declare ops for an extension.
+// Example:
+// ```
+//  declare_ops(json_op_sync[
+//    op_foo,
+//    op_bar,
+//  ]),
+// ```
```
bartlomieju: Ditto

core/extensions.rs
```diff
@@ -0,0 +1,195 @@
...
+// declare_ops_group(vec![
+//  declare_ops(json_op_sync[
+//    ...,
+//  ]),
+//  declare_ops(json_op_async[
+//    ...,
+//  ]),
+// ])
+// ```
```
bartlomieju: Ditto

core/lib.rs
```diff
@@ -81,6 +83,11 @@ pub use crate::runtime::JsErrorCreateFn;
 pub use crate::runtime::JsRuntime;
 pub use crate::runtime::RuntimeOptions;
 pub use crate::runtime::Snapshot;
+// pub use crate::runtime_modules::declare_ops!;
+// pub use crate::runtime_modules::include_js_files!;
```
bartlomieju: Remove?

op_crates/webgpu/lib.rs
```diff
@@ -92,21 +99,23 @@ impl Resource for WebGPUQuerySet {
...
-  for (url, source_code) in files {
-    isolate.execute(url, source_code).unwrap();
-  }
+    declare_webgpu_ops(),
+    Some(Box::new(move |state| {
+      // TODO: check & possibly streamline this
+      // Unstable might be able to be OpMiddleware
+      // let unstable_checker = state.borrow::<super::UnstableChecker>();
+      // let unstable = unstable_checker.unstable;
+      state.put(Unstable(unstable));
```
bartlomieju, line 113: I don't like that this unstable checker is added in `webgpu` instead of in `runtime` crate

runtime/web_worker.rs
```diff
@@ -165,11 +166,32 @@ impl WebWorker {
...
+        options.ca_data.clone(),
+      ),
+      deno_websocket::init::<Permissions>(
+        options.user_agent.clone(),
+        options.ca_data.clone(),
+      ),
+      deno_crypto::init(options.seed),
+      deno_webgpu::init(options.unstable),
+      // Metrics
+      metrics::init(),
```
bartlomieju, line 186: Not entirely sure about metrics being a middleware - I discussed it with @ry, concretely `metrics` do not properly track usage of zero copy buffers - eg. in case of "read" op metrics measure the size of buffer, but in reality it can only be filed with a few bytes (but the op will count the size of buffer instead). Maybe skip it for this PR?

core/runtime.rs
```diff
@@ -349,6 +358,41 @@ impl JsRuntime {
...
+  pub fn init_extension_ops(&mut self) -> Result<(), AnyError> {
+    let op_state = self.op_state();
+    // Original OpRegistrar
+    let mut registrar: RcOpRegistrar =
+      Rc::new(RefCell::new(JsRuntimeOpRegistrar(op_state.clone())));
+
+    // Wrap registrar
+    for e in self.extensions.iter_mut() {
+      registrar = e.init_registrar(registrar.clone());
+    }
```
ry, line 387: I'm confused about the flow of registrars and middleware.

Most of the Extensions do not register any middleware - EG op_crates/url - however somehow they all have Metrics applied to them? From what I can tell the init function in runtime/metrics.rs creates an Extension with middleware, and it is loaded as the last Extension for Worker/WebWorker. Does order matter in registering of extensions? It seems to be the case here.

runtime/web_worker.rs
AaronO: @bartlomieju I agree that's an issue, but isn't that separate from it being middleware ?
(that issue exists on `main`)

Without that middleware, it will break metrics for op_crate ops (ops declared via this new extension system). I added a test that checks that metrics work for `op_url_parse` (op_crate metrics weren't tested at all, so it would have caused an unexpected fail/crash for users)

op_crates/webgpu/lib.rs
AaronO: @bartlomieju `webgpu` (in `main`) already duplicates a good portion of the unstable code into it's own crate. We can remove this with an Unstable middleware in `runtime` (since I think none of `webgpu` ops are conditionally unstable)

core/runtime.rs
ry: I want to experiment with that too, but I think first we need to support ExternalOneByteStringResource ExternalReferences. Let's remove this method for now?

runtime/web_worker.rs
bartlomieju: That's correct, but using metrics as a middleware I don't see a way how collection of metrics could be customized per op basis

ry (commented):
@AaronO I've removed the `declare_ops` macro. There's nothing particularly wrong with something like that, but I think we can introduce it in a separate PR. I will leave the `include_js_files` because it's relatively straightforward and helpful.

core/runtime.rs
AaronO: @ry Order of extensions does NOT matter in regards to middleware, it basically inits all the middleware provided by all extensions before registering any of the ops.

core/lib.rs
AaronO: Yeah, I added those so when reviewing "exports" by checking `lib.rs` it's apparent that we export macros.

core/runtime.rs
bartlomieju: @rektide, not, that is not a V8 bug, it's lacking Deno implementation.

runtime/web_worker.rs
lucacasonato: @bartlomieju Metrics are exposed via `Deno.metrics` in stable, so we have to keep the current middleware collection until 2.0 at least.

AaronO:
@ry Are you in favour/ok with the `declare_ops!()` macro and simply want to "reduce scope" in this PR ? Or are you uncertain about it more broadly ?

I think it's genuinely useful and more-expressive, repeating the op_names as idents and strings has no upside and can get painfully verbose in `webgpu` to the point of hurting review-ability IMO. As always with these macros, they're designed to be opt-in, so users can just fallback to the "verbose version" if the macros don't fit their use-case (though it should work for most).

runtime/web_worker.rs
AaronO: @bartlomieju Was it ever customized before ? I might have missed that, I don't remember seeing anything to that effect.

It sounds like this is an orthogonal fix/change, that should happen in a follow-up PR and would have had to change with the new op-ABI anyway.

If you remove op-middleware and the metrics middleware right now, you'll cause breaking changes for users, unless you verbosely re-wrap all the ops but that kinda defeats the whole point of extension modularity.

ry:
@AaronO Trying to reduce scope - but I'm a little concerned about the exact organization of them, particularly the nested `declare_ops_group`. I wonder if it wouldn't be better to just have a macro for json_op_sync and json_op_async to produce the OpPair. It's a discussion I'd rather punt on because it's rather superficial.

AaronO:
@ry Makes sense, happy to submit a follow-up PR and have a scoped conversation on that.

I had tested macros that just wrap an OpPair, but ops are often declared in "large" groups (definitely for `webgpu`) and so there's a good increase in expressiveness by wrapping multiple at once and also grouping them by type (sync vs async, etc...)

I had also tested non-parametric versions that take no wrapper, but those are less general as we'll preserve bin/buffer ops and start introducing min_ops (they could be aliased for the common cases, but that didn't seem like a substantial improvement to me)

runtime/build.rs
AaronO: @bartlomieju I think snapshotting JsRuntimes should be less of an edge case moving forward, in https://github.com/denoland/deno/pull/10308 I suggest allowing opcalls prior to snapshotting.

So rather than breaking the abstraction of runtime extensions for the sake of snapshots I think we should handle snapshots more gracefully and allow snapshotting fully initialized runtimes (with fully initialized extensions, js, ops, ...).

core/runtime.rs
AaronO: Removed for now

core/extensions.rs
```diff
@@ -0,0 +1,136 @@
...
+    match &self.js_files {
+      Some(files) => files.clone(),
+      None => vec![],
+    }
+  }
+
+  /// Called at JsRuntime startup to initialize ops in the isolate.
+  pub(crate) fn init_ops(&mut self, registrar: RcOpRegistrar) {
+    // NOTE: not idempotent
+    // TODO: fail if called twice ?
```
bartlomieju, line 64: Yes, a panic would be fine, this really needs to be fixed before landing

core/extensions.rs
```diff
@@ -0,0 +1,136 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use crate::error::AnyError;
+use crate::{OpFn, OpId, OpState};
+
+pub type SourcePair = (&'static str, &'static str);
+pub type OpPair = (&'static str, Box<OpFn>);
+pub type RcOpRegistrar = Rc<RefCell<dyn OpRegistrar>>;
```
bartlomieju, line 9: I think I already mentioned it but IMO such alias adds even more indirection than using `Rc<RefCell<>>` directly

core/runtime.rs
```diff
@@ -351,6 +360,42 @@ impl JsRuntime {
       .unwrap();
   }
 
+  /// Initializes JS of provided Extensions
+  // NOTE: this will probably change when streamlining snapshot flow
```
bartlomieju, line 364: Not sure what this comment means, remove before landing?

core/extensions.rs
```diff
@@ -0,0 +1,136 @@
...
+    Self::new(Some(js_files), Some(ops), opstate_fn, None)
+  }
+}
+
+// Note: this used to be a trait, but we "downgraded" it to a single concrete type
+// for the initial iteration, it will like become a trait in the future
+impl Extension {
+  /// returns JS source code to be loaded into the isolate (either at snapshotting,
+  /// or at startup).  as a vector of a tuple of the file name, and the source code.
+  pub(crate) fn init_js(&self) -> Vec<SourcePair> {
```
bartlomieju, line 54: This name is confusing, it doesn't init anything just returns the files right?

core/runtime.rs
bartlomieju: I'm also confused what's going on here. Why does `init_registrar` returns another registrar? It seems that it happens for every registered extension. What's the use case?

core/extensions.rs
```diff
@@ -0,0 +1,136 @@
...
+  middleware_fn: Option<Box<OpMiddlewareFn>>,
+}
+
+impl Extension {
+  pub fn new(
+    js_files: Option<Vec<SourcePair>>,
+    ops: Option<Vec<OpPair>>,
+    opstate_fn: Option<Box<OpStateFn>>,
+    middleware_fn: Option<Box<OpMiddlewareFn>>,
+  ) -> Self {
```
bartlomieju, line 27: Looking at the usages in various crates I'd suggest to use builder pattern instead. Once you got a few lines of js sources and a few elements in a vector of ops it becomes quite hard to read this call. WDYT?

core/runtime.rs
```diff
@@ -351,6 +360,42 @@ impl JsRuntime {
...
+        // TODO(@AaronO): use JsRuntime::execute_static() here to move src off heap
+        self.execute(filename, source)?;
+      }
+    }
+    self.extensions = extensions;
+    Ok(())
+  }
+
+  /// Initializes ops of provided Extensions
+  // NOTE: this will probably change when streamlining snapshot flow
```
bartlomieju, line 380: Ditto

runtime/web_worker.rs
```diff
@@ -220,42 +247,24 @@ impl WebWorker {
...
         let op_state = js_runtime.op_state();
         let mut op_state = op_state.borrow_mut();
-        op_state.put(RuntimeMetrics::default());
         op_state.put::<Permissions>(permissions);
         op_state.put(ops::UnstableChecker {
           unstable: options.unstable,
         });
       }
 
+      js_runtime.init_extension_ops().unwrap();
```
bartlomieju, line 256: Why unwrap?

bartlomieju (commented):
After more thought I really do think we shouldn't go with the name "extensions". A couple reasons:
- term "op crates" has been used for almost a year, it will be very confusing to transition to "extensions"
- "extension" is not very specific
- these are not "extensions" per se - they do not augment the capabilities of the runtime in any way, but rather allow to group ops inside a single crate

core/runtime.rs
```diff
@@ -351,6 +360,42 @@ impl JsRuntime {
...
+  pub fn init_extension_ops(&mut self) -> Result<(), AnyError> {
```
bartlomieju, line 381: I'm thinking that maybe this method should consume the `extensions` field? It shouldn't really be used after initializing the ops, right?

core/extensions.rs
```diff
@@ -0,0 +1,136 @@
...
+  js_files: Option<Vec<SourcePair>>,
+  ops: Option<Vec<OpPair>>,
+  opstate_fn: Option<Box<OpStateFn>>,
+  middleware_fn: Option<Box<OpMiddlewareFn>>,
+}
+
+impl Extension {
+  pub fn new(
+    js_files: Option<Vec<SourcePair>>,
+    ops: Option<Vec<OpPair>>,
```
hayd, line 24: What is the distinction between `None` and `Some(vec![])`  ?

core/runtime.rs
AaronO: It's a note that `init_extension_js` and `init_extension_ops` should be fully merged into a `init_extensions()` method or even folded into `new` (with no `pub fn`s) as callers no longer need to handle them separately as snapshottable JsRuntimes call have full fledged ops.

core/runtime.rs
AaronO: Explained above (these methods shouldn't remain public mid term)

core/runtime.rs
AaronO: It's essentially a reduce, transforming our initial OpRegistrar into a fully wrapped one.

We then register all ops using that fully wrapped OpRegistrar, so middleware provided by any of the extensions gets applied to all ops.

core/extensions.rs
AaronO: Extension was originally a trait and might become one again in the future, so in some sense how the JS files are obtained is opaque, also it makes sense to remain consistent with the other methods.

**Note:** these method names are used internally in core and aren't leaked to external users / op-crates

core/extensions.rs
AaronO: I had implemented a builder pattern, Ryan stripped it out, I will restore it in a follow up PR

core/runtime.rs
AaronO: Plugins might be implemented as "dynamic extensions", in which case tracking them post init will streamline cleaning them up if we want to allow unloading plugins (which would make sense)

runtime/build.rs
AaronO: Snapshotting should no longer be an edge case since the op-layer rewrite and https://github.com/denoland/deno/pull/10308

core/extensions.rs
AaronO: Done

runtime/web_worker.rs
AaronO: Could change it to an `.expect()`, but ultimately we should panic here if it fails, the worker constructors don't return `Result`, though they obviously could. But I would rather do a 2nd pass and flesh out/polish error handling in Extensions.

core/extensions.rs
AaronO: Practically speaking there's no meaningful distinction, it's an implementation detail that will evolve and shouldn't be user facing but one upside of `Option` is that if the values need to be consumed there's a clean `.take()` operation

core/extensions.rs
AaronO: Dropped the alias

core/extensions.rs
```diff
@@ -0,0 +1,141 @@
...
+  /// or at startup).  as a vector of a tuple of the file name, and the source code.
+  pub(crate) fn init_js(&self) -> Vec<SourcePair> {
+    match &self.js_files {
+      Some(files) => files.clone(),
+      None => vec![],
+    }
+  }
+
+  /// Called at JsRuntime startup to initialize ops in the isolate.
+  pub(crate) fn init_ops(&mut self, registrar: Rc<RefCell<dyn OpRegistrar>>) {
```
rtsmarty, line 63: Is there a reason this is `Rc<RefCell<dyn OpRegistrar>>`? Looking at the function I feel it could probably just be a plain `&mut dyn OpRegistrar` and do exactly the same. If this eventually becomes a trait then it feels a little dodgy giving it an `Rc<RefCell<_>>`since it would allow extensions to stash the `OpRegistrar` away and use it to dynamically create ops during code execution (at least, in theory) which feels like an odd thing to support? I might be wrong, it could be completely sensible to want that.

core/runtime.rs
```diff
@@ -349,6 +357,45 @@ impl JsRuntime {
       .unwrap();
   }
 
+  /// Initializes JS of provided Extensions
+  // NOTE: this will probably change when streamlining snapshot flow
+  pub fn init_extension_js(&mut self) -> Result<(), AnyError> {
+    // Take extensions to avoid double-borrow
+    let mut extensions: Vec<Extension> = self.extensions.drain(..).collect();
```
rtsmarty, line 364: This will needlessly allocate & deallocate a second Vec for the extensions, you can use `std::mem::replace` to put an empty Vec (doesn't allocate) into its place instead and then replace it back once the call is done.

```rust
let mut extensions = std::mem::replace(&mut self.extensions, Vec::new()); // Allocation free
for m in extensions.iter_mut() {
    // ...
}
std::mem::replace(&mut self.extensions, extensions); // Put the extensions back.
```

core/runtime.rs
rtsmarty: This seems like it could get a little hairy if you have more than a few middlewares. Each middleware is going to Box and RefCell its parent registrar (which is also Boxed and RefCelled). It's effectively a recursive queue of `Rc<RefCell<OpRegistrar>>`. Lots of needless pointer chasing and `borrow_mut`, as well as using a bunch of stack space for the recursive function calls.

Probably worth checking, but I'd guess that collecting the `Box<OpMiddlewareFn>` into a `Vec` and then making an `OpRegistrar` which iterates over and executes these `OpMiddlewareFn` before calling into the parent `OpRegistrar` will be faster and less allocation heavy. I also think it's easier to reason about but that's possibly just me :shrug: .

Something like this, which probably doesn't compile but hopefully you get the idea

```rust
impl Extension {
    // ...
    pub fn middleware_fn(&mut self) -> Option<Box<OpMiddlewareFn>> { /* ... */ }
    // ...
}

struct JsRuntimeOpRegistrar<'a> {
    middlewares: Vec<Box<OpMiddlewareFn>>,
    state: &'a mut OpState
}
impl<'a> OpRegistrar for JsRuntimeOpRegistrar<'a> {
    fn register_op(&mut self, name: &str, op_fn: Box<OpFn>) -> OpId {
        let mut new_op = op_fn;
        for middleware in &self.middlewares {
            new_op = middleware(name, new_op);
        }
        self.state.op_table.register_op(name, new_op);
    }
}

pub fn init_extension_ops(&mut self) -> Result<(), AnyError> {
    let op_state = self.op_state();
    let mut op_state = op_state.borrow_mut();
    let mut registrar = JsRuntimeOpRegistrar {
        middlewares: vec![],
        state: &mut *op_state,
    };
    for e in self.extensions.iter_mut() {
        if let Some(op) = e.middleware_fn() {
            registrar.middlewares.push(op);
        }
    }
    let registrar = Rc::new(RefCell::new(registrar));
    // ... continue as before
}
```

The same criticism can probably be applied to the `OpMiddlewareFn` effectively doing the same thing as well, I suppose, but that would probably require a much larger rewrite, from what I can see.

core/runtime.rs
AaronO: Sure, that works too. Keep in mind that this code is very much not in a hot loop, so clarity matters more here than a small extra alloc (I'm fine with `std::mem::*` but it seems that it can mislead people to thinking that we're doing something low level and they need to pay careful attention)

core/extensions.rs
AaronO: I've killed the OpRegistrar trait and replaced it with first class op middleware.

bartlomieju (approved):
LGTM

PR #9801, opened 2021-03-15
chore(ci): add std and 3rd party regression tests

kitsonk:


kitsonk:
There seem to be a couple failures in std that look like they aren't being "skipped" properly when running under a linux host.

Also, I am not sure how to omit the matrix kind `test_regression` from the fail fast.  It isn't required, but it is still killing the rest of the tests.

caspervonb:
> Also, I am not sure how to omit the matrix kind test_regression from the fail fast. It isn't required, but it is still killing the rest of the tests.

It has to be a job to set `fail_fast`.
You can set continue_on_error for individual steps tho.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
             kind: test_release
           - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
             kind: bench
+          - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
+            kind: test_regression
```
ry, line 23: test_std would be more descriptive

PR #9813, opened 2021-03-17
fix(runtime): mitigate permission prompt stuffing

### LINKED ISSUE #9750: Runtime permissions request should fully consume STDIN before asking for user input

bnoordhuis:
Flush the tty's input buffer before reading to avoid already typed
characters from being treated as the answer to the permissions prompt.

Fixes #9750.

<hr>

An orthogonal issue is putting the tty into a known-good state (and restoring it afterwards) with `tcsetattr()`.


kitsonk:
Looks like this causing hangs in the tests on the *NIX systems.

PR #9817, opened 2021-03-17
refactor(op_crates): move op_crates ops to serde ops

crowlKats:


ry (commented):
I'm a bit confused. Do I have it right: functionally this is the same, it's just the `serde_json::from_value` is implicit?

crowlKats:
> I'm a bit confused. Do I have it right: functionally this is the same, it's just the `serde_json::from_value` is implicit?

this is just using https://github.com/denoland/deno/pull/9423, as it currently is only used for webgpu

ry (approved):
LGTM - forgot about that one!

PR #9821, opened 2021-03-18
feat(lsp): add import completions

kitsonk:


lucacasonato (commented):
Implementation looks great. Just tried this out locally, and have a few remarks:

- Only files that have previously been opened show up in completions. This is somewhat unexpected from a user point of view. I would expect that entering `./` would complete all sibling files in the directory, not just files that have previously been opened.
- I think in a later pass it would make sense to autocomplete all remote imports that are already cached in DENO_DIR, not just ones in the current workspaces' module graph.

Feels great to use though. Completions are snappy, even on a debug build.

kitsonk:
> I think in a later pass it would make sense to autocomplete all remote imports that are already cached in DENO_DIR, not just ones in the current workspaces' module graph.

Thinking about this, the cache is a one way hash, so that would be hard.

lucacasonato:
> Thinking about this, the cache is a one way hash, so that would be hard.

The meta files contain the original URL if I'm not mistaken.

lucacasonato:
Tried it out locally again and found a few more issues:


https://user-images.githubusercontent.com/7829205/111981288-c3be6980-8b07-11eb-9ae5-125d73b5c7f1.mp4

kitsonk:
> > Thinking about this, the cache is a one way hash, so that would be hard.
> 
> The meta files contain the original URL if I'm not mistaken.

Hrmmm... opening every metafile, parsing it, and extracting the URL won't be fast.  We can be smart by host, but that is as deep as it can go, and we can obviously cache it, but still...

kitsonk:
> Tried it out locally again and found a few more issues:

Ok, I think I know what is going on.

kitsonk:
> > Tried it out locally again and found a few more issues:
> 
> Ok, I think I know what is going on.

@lucacasonato ok, I think they are fixed... it was a bit of a battle to get the `textEdit` working with vscode, but it now works and hopefully the behaviour is consistent with other editors.  If you get a chance to take it for a spin.  I couldn't break it.

lucacasonato:
Tried again, and one issue is still present. I can not trigger completions in the middle of a file / folder name. For example I enter `./sr` then press `ctrl + space`, but nothing shows up. If I enter `./src` I get completions for items in the folder. I would also argue that `./src` + `ctrl + space` should not trigger completions for items inside `./src`, but `./src/` + `ctrl + space` should.

https://user-images.githubusercontent.com/7829205/112234238-5cee9c80-8c3c-11eb-8d29-e96c48d4571a.mov

kitsonk:
ok @lucacasonato here we go again... I think this fixes what you mentioned above.

cli/lsp/completions.rs
```diff
@@ -0,0 +1,771 @@
...
+      Some((specifier, range))
+    } else {
+      None
+    }
+  } else {
+    None
+  }
+}
+
+fn relative_specifier(
```
lucacasonato, line 337: Could you add a comment to describe what this function does? It is not immediately obvious to me from looking at signature / tests.

lucacasonato (approved):
Tried it again, and behaves like I expect now 💯 

Awesome work on this!

Issue #9570, opened 2021-02-21
deno bundle vague error for bare imports


nayeemrmn:
```ts
// temp.ts
import "foo";
```
`deno bundle temp.ts`
```
Bundle file:///mnt/c/Users/Nayeem/projects/deno/temp.ts
Check file:///mnt/c/Users/Nayeem/projects/deno/temp.ts
error: Unable to output bundle during Graph::bundle().
```

kitsonk:
Yeah, we should not "swallow" the error from the module slot in the graph, but instead expose that... I think it was overuse of my `.context()` at one point in time.

While this isn't a good first issue, it is a good "medium" skill issue if someone wanted to get more familiar with the internals of Deno.

wperron:
I can look into this 👍

bartlomieju:
> Yeah, we should not "swallow" the error from the module slot in the graph, but instead expose that... I think it was overuse of my `.context()` at one point in time.

@kitsonk this sounds fishy to me, overuse of `context()` should result in multiple lines underneath "Caused by" when printing the error (for reference from `anyhow` crate: https://docs.rs/anyhow/1.0.32/anyhow/). If those lines are not visible it suggests something's wrong with our error formatting for terminal errors.

bartlomieju:
Actually found a TODO, for exactly this problem:

https://github.com/denoland/deno/blob/687ff2ab14d6a735aa5bf0ec57ef00cfe0c04e4b/cli/main.rs#L1267-L1272

wperron:
yeah I talked with @lucacasonato about it -- switching line 1271 for `format!("{:#}", e` does print out the error context so that's good, but the `relative import path "foo" not prefixed with...` message is being lost/dropped somewhere

mitch292:
Hi - was browsing the l-rust tag and took a quick look at this one. Looks like the error is being swallowed on line 389:

https://github.com/denoland/deno/blob/bd961c3bc305e90e760394d2f424c5360a67543b/cli/module_graph.rs#L379-L395

Wasn't sure on the use case of when you wouldn't want to return the `Error` for `ModuleResolutionError::ImportPrefixMissing`, so didn't make a PR, but thought it might help.

wperron:
@mitch292 yeah I saw that as well, but the same code works fine with `deno run` so I'm not sure if it's really the issue. In any case, if you have something working feel free to open a PR, I've been working on some other stuff lately and haven't had the chance to circle back to this issue

[...]

PR #9822, opened 2021-03-18
fix(cli): Display anyhow error chain

mitch292:
Resolves issue #9570 . This change adds anyhow error context to terminal output which will provide more detail on the error like below. In the case where `deno bundle foo.ts` is run. Deno's specific error message, `relative import path "foo" not prefixed with..`, is omitted as it's only checked at runtime to handle dynamic imports. Happy to make any changes or go in another direction!

```
error: Unable to output bundle during Graph::bundle().

Caused by:
    0: load_transformed failed
    1: failed to analyze module
    2: failed to resolve foo from <file:///Users/andrewmitchell/Code/denocode/temp/temp.ts>
    3: The graph is missing a dependency.
         Specifier: foo from file:///Users/andrewmitchell/Code/denocode/temp/temp.ts
```


kitsonk:
Just stylistic opinion but the numbers in the stack are meaningless to the end user. Just a `-` or `>` would be better in my opinion. Thoughts @ry @bartlomieju @nayeemrmn?

bartlomieju:
> Just stylistic opinion but the numbers in the stack are meaningless to the end user. Just a `-` or `>` would be better in my opinion. Thoughts @ry @bartlomieju @nayeemrmn?

@kitsonk I think this is the default formatting of `anyhow` and can't really be changed without copy-pasting this logic to our wrapper error type. It looks alright to me.

mitch292:
@kitsonk Yeah unfortunately those numbers are coming from anyhow. If the error chain only has a depth of one they'll be omitted like below which is nice. 
```
Error: Failed to read instrs from ./path/to/instrs.json

Caused by:
    No such file or directory (os error 2)
```

@bartlomieju - Happy to write a test! I added an integration test for the use case from the original issue. Unfortunately, some of the errors from that chain are coming from SWC. So if the there are some changes in the underlying crate the test may break. I could mock some errors and test the local function that changed in `cli/main.rs`, but didn't see any other tests in the file so didn't know if I should be the first to add them :). Let me know if you want me to go in a different direction with the test.

kitsonk:
Ok, if it is the default behaviour. 🤷

wperron (approved):
Looks good, thanks

PR #9825, opened 2021-03-18
feat(permissions): allow env permission to take values

### LINKED ISSUE #3137: Limited Permission for --allow-env
Currently `--allow-env` gives program unlimited permission to access any environment variable. It's much strong privilege and there should be more limited permission control for them.

Like `--allow-net=:9999`, how about these? `--allow-env=DEBUG`, `--allow-env=DENO_DIR,DENO_INSTALL` 

`Deno.env()` should also be re-designed to get permitted variables.

I think even if full access to envars given to the process, It's dangerous to return key-value pair of all environment variables because malicious  attackers are able to  know  some secret informations if they don't know envar's name. So how about re-design `Deno.env` to immutable map-like object?

```
Deno.env.get("DENO_DIR")
Deno.env.set("DEBUG", "*")
Deno.env.has("PORT")
```

crowlKats:
closes #3137


runtime/ops/os.rs
```diff
@@ -64,13 +64,16 @@ fn op_set_env(
...
 fn op_env(
   state: &mut OpState,
   _args: Value,
   _zero_copy: &mut [ZeroCopyBuf],
 ) -> Result<Value, AnyError> {
-  state.borrow::<Permissions>().env.check()?;
-  let v = env::vars().collect::<HashMap<String, String>>();
+  let perm = &state.borrow::<Permissions>().env;
+  let v = env::vars()
+    .filter(|(var, _)| perm.check(var).is_ok())
```
kt3k, line 75: I have a concern whether it does make sense when no env permissions are given at all and Deno.env.toObject() still returning an empty object. I guess in most of the cases calling env.toObject() without any env permission is a mistake (forgot to pass permission). So it might be better to throw in that case perhaps? 🤔

kt3k:
I tried simple script like the below on my end with and without combinations of perm flags.
```ts
await Deno.permissions.request({ name: 'env', variable: 'FOO' })
console.log(Deno.env.get('FOO'));
```
The basic feature seems working to me. Good feature to have! 👍

runtime/ops/os.rs
crowlKats: I wasn't exactly sure what we would want to do here, so i just implemented this way. i don't mind changing it to being behind full env flag; i just thought this might be a little nice thing to have, though this behaviour might be a tad weird

kt3k: Ah `full env flag` sounds more appropriate here to me. Filtering can be confusing to the users who are very new to Deno and don't know about env flag.

cli/flags.rs
```diff
@@ -2647,6 +2669,22 @@ mod tests {
     );
   }
 
+  #[test]
+  fn allow_env_allowlist() {
```
bartlomieju, line 2673: Please add a test case for multiple keys and some test case that will error in validator for that values

cli/tests/unit/os_test.ts
```diff
@@ -42,13 +42,11 @@ unitTest({ perms: { env: true } }, function avoidEmptyNamedEnv(): void {
   assertThrows(() => Deno.env.delete("a\0a"), TypeError);
 });
 
-unitTest(function envPermissionDenied1(): void {
-  assertThrows(() => {
-    Deno.env.toObject();
-  }, Deno.errors.PermissionDenied);
+unitTest(function envToObjectWithoutPermission(): void {
+  assertEquals(Deno.env.toObject(), {});
```
bartlomieju, line 46: This change seems strange to me, why doesn't it throw anymore?

crowlKats: this is related to what @kt3k reviewed above. i will fix it later today

bartlomieju:
In general this PR looks good to me, but I have one comment that makes me wanna skip it for 1.9: with this change `--allow-env` allows to provide a list of variables that can be read and set - I'm thinking that maybe we should split this flag into to, one for reading and one for setting (similar to `--allow-read` and `--allow-write`). 

Before making decision I will wait for other contributors opinions.

lucacasonato:
> I'm thinking that maybe we should split this flag into to, one for reading and one for setting

I don't think this is needed as setting env vars only has an effect on you, and your processes children, but not siblings or parent processes.

nayeemrmn:
> I don't think this is needed as setting env vars only has an effect on you, and your processes children, but not siblings or parent processes.

Hmm, doesn't that also mean that variables should be settable without permission and allowlisted thereafter?

bartlomieju:
> > I don't think this is needed as setting env vars only has an effect on you, and your processes children, but not siblings or parent processes.
> 
> Hmm, doesn't that also mean that variables should be settable without permission and allowlisted thereafter?

@nayeemrmn I don't understand the question, could you provide an example?

nayeemrmn:
I realised since setting env variables within the process doesn't affect the outside, there's no reason for it to require permission.

bartlomieju:
> I realised since setting env variables within the process doesn't affect the outside, there's no reason for it to require permission.
> 
> ```ts
> Deno.env.get("foo"); // Requires `--allow-env=foo`.
> ```
> 
> ```ts
> Deno.env.set("foo", "bar"); // Shouldn't require permission.
> Deno.env.get("foo"); // Shouldn't require permission.
> ```

@nayeemrmn so this suggestion alters current behavior, which would be a breaking change. I'm also not sure how would it be implemented, but I'm more than happy to discuss it for 2.0.

nayeemrmn:
If you take away a permission gate _because_ it's not actually a security issue, I'm not sure it's a breaking change. But either way it certainly doesn't need to be part of this PR 👍

ry (approved):
LGTM - nice work @crowlKats !

piscisaureus:
@piscisaureus

piscisaureus:
I think sufficient effort was put into this PR to address Windows quirks.

Of course, it is not a given that windows uses case folding that is 100% compatible with `.toUpperCase()`. Who knows, this could lead to some obscure bugs down the line, or not.

PR #9826, opened 2021-03-18
Feature: coverage threshold

### LINKED ISSUE #9669: Add a default and configured threshold for  `deno coverage`
**What**

It would be great if `deno coverage` could use a default threshold that must be met, otherwise the command will fail with an exit code other than `0`. On top of this, it would be great for this to be configured so a user can set their own threshold.

For example, `deno coverage --check --lcov` will return a status of 1 if my coverage (which is 79%) doesn;t meet the. default threshold (which is 80%).

Then `deno coverage --check --lines=90 --lcov` would return a status of 1 when my coverage for lines (which is 89%) doesn't meet the expected 90%.

Note that the thresholds, return status' and configuration implementation are demonstrated just for explanation. I'm not suggesting how to configure nor what the default threshold should be.

**Why**

If people were to start adding coverage to their modules, then ideally the command(s) need to fail if coverage is too low (ie a new feature was added but not unit tests or coverage for it were implemented).

It would also help in the CI, where we usually have a `master` or `ci` job that runs our tests, if test coverage doesn't meet an expectation, then the user (be it a fork on the repo or not), would have to **ensure theey add coverage** because the CI will fail due to the returned status code

*100% coverage all the way*

ayshiff:
Closes #9669

This PR adds the possibility to set a threshold to the `deno coverage` command and fail if coverage doesn't meet it.

I have reused the same nomenclature that was proposed in the issue.

carbon (10)

If the command has the `--lcov` flag, it will check the coverage for `lines`, `branches` and `functions` otherwise it will only check for `lines` as the `PrettyCoverageReporter` doesn't have a way to identify the coverage for `branches` and `functions`.

By default, the threshold is set to **80%**.

If coverage doesn't meet the threshold, it will throw an error and display the coverage elements that are below the threshold.

<p align="center">

</p>

I would love to get your feedback on what I did and discuss what could be improved 👍

I don't know if the `println` errors are relevant, in case the user does a `deno coverage --unstable cov_profile --lcov > cov_profile.lcov`. He will get the errors `"The coverage threshold..."` in his file.

### DIFF
--- a/cli/tools/coverage.rs
+++ b/cli/tools/coverage.rs
@@ -9,6 +9,7 @@ use crate::media_type::MediaType;
 use crate::module_graph::TypeLib;
 use crate::program_state::ProgramState;
 use crate::source_maps::SourceMapGetter;
+use deno_core::error::custom_error;
 use deno_core::error::AnyError;
 use deno_core::serde_json;
 use deno_core::serde_json::json;
@@ -27,6 +28,23 @@ use uuid::Uuid;
 pub struct CoverageCollector {
   pub dir: PathBuf,
   session: Box<InspectorSession>,
+}
+
+struct CoverageValues {
+  hit: f32,
+  found: f32,
+}
+
+pub struct TotalVisitCoverage {
+  lines: Option<CoverageValues>,
+  functions: Option<CoverageValues>,
+  branches: Option<CoverageValues>,
+}
+
+pub struct TotalCoverages {
+  lines: CoverageValues,
+  functions: CoverageValues,
+  branches: CoverageValues,
 }
 
 impl CoverageCollector {
@@ -134,7 +152,7 @@ pub trait CoverageReporter {
     script_source: &str,
     maybe_source_map: Option<Vec<u8>>,
     maybe_original_source: Option<String>,
-  );
+  ) -> TotalVisitCoverage;
 
   fn done(&mut self);
 }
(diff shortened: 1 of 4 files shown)


cli/tools/coverage.rs
```diff
@@ -655,12 +732,88 @@ pub async fn cover_files(
...
+    {
+      coverage_threshold_error(
+        "branches",
+        total_coverage_branches.unwrap(),
+        branches,
+      );
+      threshold_met = false;
+    };
+
+    if !threshold_met {
```
caspervonb, line 811: Exit early.

cli/tools/coverage.rs
```diff
@@ -655,12 +732,88 @@ pub async fn cover_files(
       .get_source(&module_specifier)
       .map(|f| f.source);
 
-    reporter.visit_coverage(
+    let total_coverage = reporter.visit_coverage(
```
caspervonb, line 735: Reporter shouldn't return anything, adds even more duplication.

caspervonb: Also check shouldn't have any relationship with the reporter.

caspervonb (changes requested):
This approach is too tightly coupled to the reporter, check should be independent of the output format.

ayshiff:
> Hi there!
> 
> Thanks for attempting this but would perhaps be better if you checked in before, already have a draft for this am just busy with a different feature at the moment.
> 
> This approach is too tightly coupled to the reporter, check should be independent of the output format.

Okay, no problem!
Sorry, I hadn't seen the draft and since the issue didn't mention anything I didn't pay attention.
I can't find the draft, can you point it out to me, I'd love to see the right way to implement this feature.
Thanks

crypt096 (approved):

bartlomieju:
@ayshiff thank you for the PR! I've discussed it out of bands with @caspervonb and while we appreciate the effort we think it's not the right approach (especially with recent changes to the test runner infrastructure). Casper said he'll revisit the topic in the near future. Closing without a merge.

PR #9837, opened 2021-03-19
chores: enforce type ResourceId across codebase

littledivy:
The type declared was hardly used earlier. This PR enforces its use across op args.


ry (approved):
LGTM - thanks

PR #9841, opened 2021-03-20
refactor(runtime/permissions): Rename permission structs

nayeemrmn:
- `BooleanPermission` -> `UnitPermission`: Since it's not boolean. `UnaryPermission` queries with one argument, `UnitPermission` queries with 0 and is only associated with a single `PermissionState`.
- `{Read,Write,Net}Permission` -> `{Read,Write,Net}Descriptor`: `UnaryPermission<ReadDescriptor>` doesn't stutter, more accurate.

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -69,6 +69,41 @@ impl Default for PermissionState {
   }
 }
 
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct UnitPermission {
+  pub name: &'static str,
+  pub description: &'static str,
+  pub state: PermissionState,
+}
+
+impl UnitPermission {
+  pub fn query(&self) -> PermissionState {
+    self.state
+  }
+
+  pub fn request(&mut self) -> PermissionState {
+    if self.state == PermissionState::Prompt {
+      if permission_prompt(&format!("access to {}", self.description)) {
+        self.state = PermissionState::Granted;
+      } else {
+        self.state = PermissionState::Denied;
+      }
+    }
+    self.state
+  }
+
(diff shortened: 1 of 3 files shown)


runtime/permissions.rs
```diff
@@ -81,9 +116,15 @@ pub struct UnaryPermission<T: Eq + Hash> {
...
 
 #[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize)]
-pub struct ReadPermission(pub PathBuf);
+pub struct ReadDescriptor(pub PathBuf);
+
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize)]
+pub struct WriteDescriptor(pub PathBuf);
+
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize)]
+pub struct NetDescriptor(pub String, pub Option<u16>);
```
crowlKats, line 125: why did you move the descriptor definitions all together? having them to their corresponding impl block seems better to me

nayeemrmn: I had trouble finding them initially, and this was where I looked for them to be so it felt right. Probably because the impl blocks are for `UnaryPermission<_>`.

crowlKats: hm, alright. there is also some impls for the NetDescriptor directly, so maybe you wanna move those up as well

crowlKats: (you forgot the display impl for net descriptor)

crowlKats (approved):
LGTM

ry (approved):
LGTM

PR #4309, opened 2020-03-10
perf: use Object instead of Map for promise table

bartlomieju:
This PR addresses GC problems with `deno_tcp.ts`

Benchmark:
```
$ deno --allow-net tools/deno_tcp.ts
$ wrk -d 10 --latency http://127.0.0.1:4500
```

this PR
```
Running 10s test @ http://127.0.0.1:4500
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   135.29us   41.03us   3.14ms   97.11%
    Req/Sec    36.23k     1.34k   37.46k    96.53%
  Latency Distribution
     50%  131.00us
     75%  133.00us
     90%  137.00us
     99%  287.00us
  727913 requests in 10.10s, 35.40MB read
Requests/sec:  72072.56
Transfer/sec:      3.51MB
```

`master`
```
Running 10s test @ http://127.0.0.1:4500
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   197.70us  398.18us   5.68ms   96.89%
    Req/Sec    34.59k     2.38k   38.25k    85.15%
  Latency Distribution
     50%  128.00us
     75%  134.00us
     90%  166.00us
     99%    2.70ms
  695113 requests in 10.10s, 33.81MB read
Requests/sec:  68814.49
Transfer/sec:      3.35MB
```

`v0.35.0`
```
Running 10s test @ http://127.0.0.1:4500
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   285.51us  711.19us   9.51ms   95.88%
    Req/Sec    31.45k     2.56k   35.05k    88.00%
  Latency Distribution
     50%  139.00us
     75%  144.00us
     90%  189.00us
     99%    4.60ms
  625922 requests in 10.00s, 30.44MB read
Requests/sec:  62572.10
Transfer/sec:      3.04MB
```

GC output
this PR
```
[72564:0x38b300000000]     8904 ms: Scavenge 3.4 (6.2) -> 1.4 (6.2) MB, 0.2 / 0.0 ms  (average mu = 1.000, current mu = 1.000) allocation failure
[72564:0x38b300000000]     8920 ms: Scavenge 3.4 (6.2) -> 1.4 (6.2) MB, 0.2 / 0.0 ms  (average mu = 1.000, current mu = 1.000) allocation failure
[...]

PR #9843, opened 2021-03-20
refactor(core): new optimized op-layer using serde_v8

### LINKED ISSUE #9914: Dangling promise in async json ops when arg parsing fails
This bug seems to exist since at least `v1.4.0` (when json ops were introduced to core)

## Reproducing it

This bug can be reproduced with the following:

```
❯ deno eval "var file = await Deno.open(1);"
error: Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promise.
```

## Scope & impact

- **Impact:** fairly serious, dangling promises will lead to surprising and hard to debug bugs for users
- **Scope:** fortunately very few users should be impacted in the real world, since this should be caught by typescript type checks and many ops use the loosely-typed `serde_json::Value` (so their arg-parsing would fail inside the `op_fn` causing an `async` error that would be correctly handled)

## Cause

- When parsing fails [here](https://github.com/denoland/deno/blob/b11249647fa14e482881fd22446dbd8b0cbac27c/core/ops_json.rs#L107-L111)
- Aync ops will return an `Op::Sync` error [here](https://github.com/denoland/deno/blob/b11249647fa14e482881fd22446dbd8b0cbac27c/core/ops_json.rs#L124-L134)
- But in [opAsync](https://github.com/denoland/deno/blob/b11249647fa14e482881fd22446dbd8b0cbac27c/core/core.js#L232-L253) the return value of `dispatch()` is ignored

## Fix

This can be fixed by simply handling the `dispatch()` return value in `core.js` and checking if it contains an error

AaronO:
This is the first PR in a series refactoring the op-layer to be both simpler and faster, it implements ideas shared in https://github.com/denoland/deno/issues/9540 

Keep in mind that given the breadth of changes, the intermediate code does not reflect the final intended state.

The bulk of the changes in this PR are scoped to `core/`

## Key ideas

- Use `serde_v8` for a maximally efficient bijection between `rust <> js` values
- Simplify and unify op-layer
  - Remove concept of custom op-encodings (no longer needed)
  - Handle op-metadata (errors, promise IDs) explicitly in the op-layer vs per op-encoding (aka: out-of-payload)
  - Remove shared queue & custom "asyncHandlers", all async values are returned in batches via `js_recv_cb`
  - The op-layer should be thought of as simple function calls with little indirection or translation besides the conceptually straightforward `serde_v8` bijections
- Preserve concepts of `json/bin/min` as semantic groups of their inputs/outputs instead of their op-encoding strategy, preserving these groups will also facilitate partial transitions over to v8 Fast API for the "min" and "bin" groups

## Changes

- [x] Unify json and bin ops
  - [x] Switch both to `serde_v8`
  - [x] Common error handling
  - [x] Common async return mechanisms (everything "overflows"
- [x] Change `OpFn` sig to ~= `(state, payload, optional_buffer) -> serializable`
- [x] But maintain retro-compatibility with all functions declared using `reg_json_sync` or similar
- [x] Adapt metrics wrapper and other misc places (plugins, ...)
- [x] `serde_v8` tweaks (mainly to relax type-checking and be more permissive)
  - [x] Handle deserializing enums in `serde_v8` (after which only bytes-support would be unimplemented, intentionally)
- [x] Mainly rollback `core.js` to a prior simpler version
- [x] Strip shared queue code
  - [x] From rust code
  - [x] From JS code
- [x] Fix #9914 when rewriting `core.js`
- [x] Comment out or remove `JsRuntime` tests regarding shared queue or old op-logic
- [x] Integration tests
  - [x] Adapt broken/changed ones (metrics, sendAsyncStackTrace, ...)
  - [x] Remove shared queue or buffer-encoding op tests
  - [x] `resolve_dns`, broken due to `serde_v8@0.1.0` not implementing `deserialize_enum`
- [x] Move `promise_id` out of `OpFn` sig and out of `OpResult`
- [ ] Update core's TS types
  - [x] Make `zeroCopy` an optional arg instead of variadic
  - [ ] Add declarations for `binOpSync/binOpAsync` ?
- [x] Adapt `http_bench_json_ops` to follow best practices
- [x] Other adjustments for all tests to pass (e.g: diffs in JSON comparisons in LSP tests)
  - [x] Match `serde_json`'s handling of map-pairs with `undefined` values in `serde_v8`
  - [x] Comment out/disable plugin tests (to fix in follw up
- [x] Last nitpicks and reviews
  - [x] Drop / unwrap `OpBuf`, `RcOpState` type aliases
- [x] Make `serde_v8` fail when deserializing non-utf8 strings (TODO: add failing tests)

## Follow-up PRs

1. Change op-sig, removing variadic `BufVec`
2. More op-layer optimizations
3. Revisit plugin ops (related to https://github.com/denoland/deno/pull/9850)
4. Tweak all ops to follow best-practices (i.e: avoid using `serde_json::Value` and `json!()`)
5. Optimize `url_parse`
6. Optimize `metrics_op` since it increases our baseline op-overhead (accounting for nearly 50% of the baseline iirc, after all the other optimizations have been made)
7. Optimize `op_now` by switching it to a json-op or introducing min-ops (first candidates for v8 Fast API)
8. Add op baseline-overhead benches in `core/` (similar to https://gist.github.com/AaronO/41803d5fc131417221a78c911cd3c6b0) (✅ Done: https://github.com/denoland/deno/pull/9924)
9. Consider renaming bin-ops to io-ops (possibly a better fit for their semantic group)
10. Update docstrings, docs & `core/README` to reflect new op-layer
11. Follow up on `promiseTable` performance (obj vs map vs ?)

## Notes

- This temporarily breaks `http_bench_bin_ops`, it will fail due to the runtime assertion of bin-ops requiring a non-null buffer arg. In practice all of deno's bin-ops follow this and this assertion will be removed by the follow-up PR `1.` removing the variadic `BufVec`
- This does not contain all of the op-layer optimizations
- It aims to get a first working version with minimal breaking changes elsewhere


core/core.js
```diff
@@ -189,63 +189,57 @@ SharedQueue Binary Layout
...
-  }
-
-  function decodeJson(ui8) {
-    const s = core.decode(ui8);
-    return JSON.parse(s);
-  }
-
   let nextPromiseId = 1;
-  const promiseTable = {};
+  const promiseTable = new Map();
```
bartlomieju, line 193: Warning, map was previously used to store promises, but was replaced with an object due to GC performance problems https://github.com/denoland/deno/pull/4309

AaronO: @bartlomieju I'll follow-up on this in a separate PR and revert it if `Object` ultimately beats `Map`. It will be easier to investigate this once we have baseline-overhead benches for async ops (follow-up PR `8.`)

core/runtime.rs
```diff
@@ -1325,9 +1318,12 @@ impl JsRuntime {
...
   }
 
-  fn poll_pending_ops(&mut self, cx: &mut Context) -> Vec<(OpId, Box<[u8]>)> {
+  fn poll_pending_ops(
+    &mut self,
+    cx: &mut Context,
+  ) -> Vec<(PromiseId, OpResponse)> {
     let state_rc = Self::state(self.v8_isolate());
-    let mut overflow_response: Vec<(OpId, Box<[u8]>)> = Vec::new();
+    let mut overflow_response: Vec<(PromiseId, OpResponse)> = Vec::new();
```
ry, line 1326: Is it the case that this "overflow_response" is the main (only?) mechanism now for async op responses? If so perhaps it should have different name...

AaronO: Yep, I had called it `finished_responses` in a previous PR, `ready_responses` could also be a good name.

I didn't change that here so it was easier to review/relate to the old code and understand that everything now "overflows".

Happy to change it

core/runtime.rs
```diff
@@ -1425,23 +1413,22 @@ impl JsRuntime {
     let mut args: Vec<v8::Local<v8::Value>> =
       Vec::with_capacity(2 * overflown_responses_size);
```
ry, line 1414: Please add a comment about the structure of the arguments sent back to JS

core/runtime.rs
ry: Maybe "async_responses" ?

AaronO: There are a few other naming changes that would make sense in the context of this op-layer, for example it might make sense to rename `send/recv` since I intend the the op-layer to be thought of as function calls rather than request/responses with custom user payloads.

`send` could be renamed to `opcall` (similar to `syscall`) or `call_op`

ry:
I've give this a cursory review and the structure looks good to me.

core/bindings.rs
```diff
@@ -380,59 +377,79 @@ fn send<'s>(
...
-    Ok(bufs) => bufs,
-    Err(exc) => {
-      scope.throw_exception(exc);
+  // send(0) returns obj of all ops, handle as special case
+  if op_id == 0 {
+    // TODO: Serialize as HashMap when serde_v8 supports maps ...
+    let ops = OpTable::op_entries(state.op_state.clone());
+    rv.set(to_v8(scope, ops).unwrap());
+    return;
+  }
```
bartlomieju, line 396: Unlucky that it's been moved to `send`, is there a reason why can't it stay in `route_op`?

core/bindings.rs
```diff
@@ -380,59 +377,79 @@ fn send<'s>(
...
+
+  // PromiseId
+  let p_id: PromiseId = match v8::Local::<v8::Integer>::try_from(args.get(1))
```
bartlomieju, line 399: Nitpick: rename to `promise_id`

core/core.js
```diff
@@ -1,411 +1,123 @@
...
-  }
-
-  function getMeta(index) {
-    if (index >= numRecords()) {
-      return null;
-    }
-    const buf = shared32[INDEX_OFFSETS + 2 * index];
-    const opId = shared32[INDEX_OFFSETS + 2 * index + 1];
-    return [opId, buf];
+    return opsCache = Object.freeze(Object.fromEntries(send(0)));
```
bartlomieju, line 18: Nitpick, please split assignment and return into two separate lines for readability.

core/core.js
```diff
@@ -1,411 +1,123 @@
...
-    while (true) {
-      const opIdBuf = shift();
-      if (opIdBuf == null) {
-        break;
-      }
-      assert(asyncHandlers[opIdBuf[0]] != null);
-      asyncHandlers[opIdBuf[0]](opIdBuf[1], true);
-    }
+  let nextPromiseId = 1;
+  const promiseTable = new Map();
```
bartlomieju, line 43: Nitpick, please move all variable declarations to the top of the file

core/core.js
```diff
@@ -1,411 +1,123 @@
...
 
-    for (let i = 0; i < arguments.length; i += 2) {
-      asyncHandlers[arguments[i]](arguments[i + 1], false);
+  function processResponse(res) {
+    // const [ok, err] = res;
+    if (res[1] === null) {
```
bartlomieju, line 47: Is it guaranteed that `err` will be null? Maybe use non-strict comparison instead?

core/core.js
```diff
@@ -1,411 +1,123 @@
...
-
-    // Synchronously dispatch async request
-    core.dispatch(opId, ...opRequestBuilder(requestId));
-
-    // Wait for async response
-    return promise;
+    promise.resolve = resolve;
+    promise.reject = reject;
+    promiseTable.set(promiseId, promise);
+    return await promise;
```
bartlomieju, line 76: Correct me if I'm wrong but I think `await` here could be skipped for the same effect

core/core.js
```diff
@@ -1,411 +1,123 @@
...
-    const requestId = nextRequestId++;
-
-    // Create and store promise
-    const promise = new Promise((resolve, reject) => {
-      promiseTable[requestId] = [resolve, reject];
+  async function jsonOpAsync(opName, args = null, zeroCopy = null) {
+    const promiseId = nextPromiseId++;
+    const maybeError = dispatch(opName, promiseId, args, zeroCopy);
+    // Handle sync error (e.g: error parsing args)
+    if (maybeError) processResponse(maybeError);
```
bartlomieju, line 67: Wrap in braces

core/core.js
```diff
@@ -1,411 +1,123 @@
...
   }
 
-  function jsonOpAsync(opName, argument = null, ...zeroCopy) {
-    return opAsync(
-      opName,
-      (requestId) => jsonOpBuildRequest(requestId, argument, zeroCopy),
-      jsonOpParseResult,
-    );
+  function binOpSync(opName, args = null, zeroCopy = null) {
+    return jsonOpSync(opName, args, zeroCopy);
```
bartlomieju, line 95: This is surprising, can you explain why `binOpSync` actually calls `jsonOpSync`?

core/ops.rs
```diff
@@ -20,19 +22,84 @@ use std::ops::DerefMut;
 use std::pin::Pin;
 use std::rc::Rc;
 
-pub type OpAsyncFuture = Pin<Box<dyn Future<Output = Box<[u8]>>>>;
-pub type OpFn = dyn Fn(Rc<RefCell<OpState>>, BufVec) -> Op + 'static;
+pub use erased_serde::Serialize as Serializable;
+pub type RcOpState = Rc<RefCell<OpState>>;
```
bartlomieju, line 26: Please remove, this is additional congnitive load compared to using `Rc<RefCell<OpState>>` directly

core/ops.rs
```diff
@@ -20,19 +22,84 @@ use std::ops::DerefMut;
...
+pub type PromiseId = u64;
+pub type OpBuf = Option<ZeroCopyBuf>;
```
bartlomieju, line 28: I don't really like this type alias either, `OpBuf` suggests that there is a buffer, I believe using `Option<ZeroCopyBuf>` will be more readable

core/ops.rs
```diff
@@ -20,19 +22,84 @@ use std::ops::DerefMut;
...
+
+  pub fn empty() -> Self {
+    Self {
+      scope: None,
+      value: None,
+    }
+  }
+
+  pub fn deserialize<T: DeserializeOwned>(self) -> Result<T, AnyError> {
+    // TODO: Map errors
```
bartlomieju, line 57: This should probably be addressed before landing, what exactly needs to be mapped here?

core/ops.rs
```diff
@@ -20,19 +22,84 @@ use std::ops::DerefMut;
...
+    serde_v8::from_v8(self.scope.unwrap(), self.value.unwrap())
+      .map_err(AnyError::from)
+  }
+}
+
+pub enum OpResponse {
+  Value(Box<dyn Serializable>),
+  Buffer(Box<[u8]>),
+}
```
bartlomieju, line 66: Just to make sure, this is temporary measure to provide compatiblity with json ops?

core/ops.rs
```diff
@@ -73,41 +140,39 @@ pub struct OpTable(IndexMap<String, Rc<OpFn>>);
...
+    F: Fn(Rc<RefCell<OpState>>, OpPayload, OpBuf) -> Op + 'static,
   {
     let (op_id, prev) = self.0.insert_full(name.to_owned(), Rc::new(op_fn));
     assert!(prev.is_none());
     op_id
   }
 
+  pub fn op_entries(state: RcOpState) -> Vec<(String, OpId)> {
+    state.borrow().op_table.0.keys().cloned().zip(0..).collect()
+  }
```
bartlomieju, line 152: See comment in `core/bindings.rs` regarding moving `0` id op

core/ops_bin.rs
```diff
@@ -1,54 +1,23 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 1: Not sure I understand, I thought bin ops could be removed altogether now?

core/ops.rs
```diff
@@ -20,19 +22,84 @@ use std::ops::DerefMut;
 use std::pin::Pin;
 use std::rc::Rc;
 
-pub type OpAsyncFuture = Pin<Box<dyn Future<Output = Box<[u8]>>>>;
-pub type OpFn = dyn Fn(Rc<RefCell<OpState>>, BufVec) -> Op + 'static;
+pub use erased_serde::Serialize as Serializable;
```
bartlomieju, line 25: Is this temporary or will it be required to use `erased_serde` after all the refactors are done?

core/plugin_api.rs
```diff
@@ -10,6 +10,7 @@
 
 pub use crate::Op;
 pub use crate::OpId;
+pub use crate::OpResponse;
```
bartlomieju, line 13: Seems like unintended change

core/runtime.rs
bartlomieju: +1 for `async_responses`

runtime/metrics.rs
```diff
@@ -138,29 +138,29 @@ pub fn metrics_op(name: &'static str, op_fn: Box<OpFn>) -> Box<OpFn> {
 
     match op {
       Op::Sync(buf) => {
-        metrics.op_sync(bytes_sent_control, bytes_sent_data, buf.len());
+        metrics.op_sync(bytes_sent_control, bytes_sent_data, 0);
```
bartlomieju, line 141: A bit unlucky that we lose this information, it will be a breaking change to the metrics interface

runtime/ops/worker_host.rs
```diff
@@ -255,6 +255,13 @@ impl<'de> de::Visitor<'de> for ParseBooleanOrStringVec {
     formatter.write_str("a vector of strings or a boolean")
   }
 
+  fn visit_unit<E>(self) -> Result<UnaryPermissionBase, E>
+  where
+    E: de::Error,
+  {
+    self.visit_bool(false)
+  }
```
bartlomieju, line 263: A comment here explaining why unit is encoded as `false` would be helpful

serde_v8/src/de.rs
```diff
@@ -103,11 +112,12 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
...
-      visitor.visit_bool(self.input.boolean_value(&mut self.scope))
-    } else {
-      Err(Error::ExpectedBoolean)
-    }
+    // if self.input.is_boolean() {
+    //   visitor.visit_bool(self.input.boolean_value(&mut self.scope))
+    // } else {
+    //   Err(Error::ExpectedBoolean)
+    // }
+    visitor.visit_bool(self.input.is_true())
```
bartlomieju, line 120: Remove?

core/bindings.rs
AaronO: I think it's cleaner to have it in `send` since it's a special case and it's usually best to handle special cases early

core/core.js
AaronO: Yep, I had done that in the original iteration, it got lost in the "rebase"

core/core.js
AaronO: It is guaranteed to be `null` since it's the output of `to_v8()` on a `None` Option, which will always be `null`.

But the method of returning errors will likely be changed by one of my follow-up PRs to improve the performance

core/core.js
bartlomieju: I actually believe otherwise, you need to expose additional method on `JsRuntime` to do that and leak that special case to a "layer above" `JsRuntime`. If that's possible please move back to `route_op`

core/bindings.rs
bartlomieju: I actually believe otherwise, you need to expose additional method on `JsRuntime` to do that and leak that special case to a "layer above" `JsRuntime`. If that's possible please move back to `route_op`

core/ops_bin.rs
AaronO: @bartlomieju They can be removed but see my note in **Key Ideas** (top comment):
> Preserve concepts of json/bin/min as semantic groups of their inputs/outputs instead of their op-encoding strategy, preserving these groups will also facilitate partial transitions over to v8 Fast API for the "min" and "bin" groups

Also want to minimize changes to code outside of core, to make this PR easier to land/review.

core/ops.rs
AaronO: None of the op implementors will need to use `erased_serde`, but we need it to Box serializable stuff so async ops can then serialize it with `serde_v8`.

Right now it's used for sync & async ops, for sync ops I'll introduce a faster approach (saving the `Box` alloc overhead)

core/bindings.rs
AaronO: As discussed, I think this is a special case and best to handle as such.

core/core.js
AaronO: Correct, it's an artefact from the [v1.8.1/core.js](https://github.com/denoland/deno/blob/v1.8.1/core/core.js#L230) which my `core.js` is based off (generally simpler and easier to adapt to what I wanted).

Will remove it since `return await` costs an extra microtask, so this will be slightly faster.

core/core.js
AaronO: I had wrapped this in braces and then `./tools/format.js` undid that.

core/bindings.rs
AaronO: Done

runtime/metrics.rs
AaronO: Changed the unit tests to expect `0`, we reached a consensus on Discord that `0` was better than an estimate or `8B/op`.

core/core.js
AaronO: Implementations are now identical, but relevant to keep as semantic groups.

Might clean up in a follow-up PR, also wanted to minimize code changes outside of `core/` for this PR.

core/ops.rs
AaronO: This exists to allow returning buffers. This will likely evolve in follow-up PRs and is dependant on how we decide to implement @inteon's structured-serialization worker message-passing PR (which would introduce the only ops returning buffers)

core/plugin_api.rs
AaronO: Yeah, I was going to start working on making this work for plugins and then opted to leave it for a follow-up PR

core/ops.rs
AaronO: Nothing it was an outdated from the first pass that lacked the `.map_err(AnyError::from)`

runtime/ops/worker_host.rs
AaronO: This is basically a means to map missing/undefined to `false`

I haven't looked into depth why this worked before with `serde_json`, but basically we implement `visit_unit` since we call `deserialize_any`, which in `serde_v8` will reflect the v8 value and for null/undefined call `visit_unit`.

core/plugin_api.rs
AaronO: Actually that's only part of the story, I re-exported `OpResponse` here because in `test_plugin` and supposedly all plugins, their imports come from `deno_core::plugin_api::*`.

So the convention appears to be: re-exporting the subset of core types used by plugins here.

core/ops.rs
AaronO: I think `RcOpState` is substantially more legible than `Rc<RefCell<OpState>>`. The fully expanded func-signature would be verbose here:
```
F: Fn(RcOpState, OpPayload, OpBuf) -> Op + 'static,
```

I think as a general rule type-aliases should be discouraged, but when you've got one place that defines a type and then N other users (were N is large ~200 here) that don't care about the type specifics, the type is more an interface than a concrete type in their eyes, since it's mainly "however state is passed to OpFn".

`OpBuf` follows a similar rationale, 199/220 or 90% of our ops don't use the data buffer arg, they don't care about the type specifics and simply want to follow / implement an interface.

The `OpBuf` "issue" could be worked around by adding buffer-less json ops (essentially what `tsc` does), but that causes its own fragmentation and issues IMO.

Here's dump of of deno's op signatures: https://gist.github.com/AaronO/2c03a99912b786c7a3f27f2b9f9bf9a5

core/ops.rs
ry: It's a nit but I'm with @bartlomieju on this. It's nice to have Option types be explicit. I think this type alias only makes code a bit less explicit.

I also agree with removing the `RcOpState` alias, but it's a bit more complex and thus more understandable. So if you were to remove only one of these, please remove `OpBuf`.

AaronO: > It's a nit but I'm with @bartlomieju on this. It's nice to have Option types be explicit. I think this type alias only makes code a bit less explicit.
> 
> I also agree with removing the `RcOpState` alias, but it's a bit more complex and thus more understandable. So if you were to remove only one of these, please remove `OpBuf`.

I'll undo the aliases, I think it's more important to land this and we can debate the aliases in follow-up conversations.

But for the context/record, the aliases served a few things IMO:
1. Reducing verbosity & improving consistency
2. Progressive disclosure of complexity (so unused "features" have a minimal complexity tax)
3. Given that the `OpFn` signature is controlled by one section of the codebase, but then consumed by ~200+ ops, I view it as the ops implementing and following the op interface, hence the arg types are somewhat abstract interfaces to them rather than specific concrete types
4. Ops don't really care about the specific type of `OpState`, they use it as an interface to get/set typed-state & resources
5. Nearly 90% of the ops don't use the buffer-arg and hence don't care about its type, we could add buffer-less json ops (like TSC does) but that would cause fragmentation

I get that aliasing types, especially `Option<T>`, isn't very rusty and would be a mess if done everywhere, I think there's something to consider in this case when it's ignored 90% of the time and types are more of an interface than concrete implementation types.

I'll unwrap the aliases, but I think this warrants some extra thought and consideration in the future.

AaronO:
@ry @bartlomieju I've implemented `deserialize_enum` in `serde_v8`, so now all integration tests pass.

I just need to cleanup & remove the `JsRuntime` tests and then pass lint/clippy. Should hopefully be able to do that tomorrow morning in parallel of @piscisaureus's review.

ry:
@AaronO `cargo test -p deno_core` does not compile for me. Should it?

AaronO:
> @AaronO `cargo test -p deno_core` does not compile for me. Should it?

> I just need to cleanup & remove the JsRuntime tests.

Some of tests used the old `OpFn` sig or tested the shared queue, need to remove/fix those and fix `core/examples/hello_world.rs`

serde_v8/src/de.rs
```diff
@@ -268,8 +268,13 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
       Some(names) => from_v8(self.scope, names.into()).unwrap(),
       None => vec![],
     };
-    let keys: Vec<v8::Local<v8::Value>> =
-      keys.drain(..).map(|x| x.into()).collect();
+    let keys: Vec<v8::Local<v8::Value>> = keys
+      .drain(..)
+      .map(|x| x.into())
+      // Filter keys to drop keys whose value is undefined
+      // TODO: optimize, since this doubles our get calls
```
piscisaureus, line 275: How big is the difference?

I think the checking for undefined can also be done by dereferencing the `Local<Value>` once or twice and testing for equality. It's hacky of course, but good enough for a try out.

core/ops.rs
```diff
@@ -20,19 +22,83 @@ use std::ops::DerefMut;
...
+pub type RcOpState = Rc<RefCell<OpState>>;
+pub type PromiseId = u64;
+pub type OpBuf = Option<ZeroCopyBuf>;
+pub type OpAsyncFuture = Pin<Box<dyn Future<Output = OpResponse>>>;
+pub type OpFn = dyn Fn(RcOpState, OpPayload, OpBuf) -> Op + 'static;
 pub type OpId = usize;
 
+pub struct OpPayload<'a, 'b, 'c> {
+  pub(crate) scope: Option<&'a mut v8::HandleScope<'b>>,
```
piscisaureus, line 34: Interesting idea, wrapping up a scope and value together in a struct.
I hope it is actually safe.

core/ops_json.rs
```diff
@@ -59,15 +42,20 @@ pub fn json_op_sync<F, V, R>(op_fn: F) -> Box<OpFn>
...
-    let result = serde_json::from_slice(&bufs[0])
-      .map_err(AnyError::from)
-      .and_then(|args| op_fn(&mut state.borrow_mut(), args, &mut bufs[1..]));
-    let buf =
-      json_serialize_op_result(None, result, state.borrow().get_error_class_fn);
-    Op::Sync(buf)
+  Box::new(move |state, payload, buf: OpBuf| -> Op {
+    // For sig compat map OpBuf to BufVec
+    let mut bufs: BufVec = match buf {
+      Some(b) => vec![b],
```
piscisaureus, line 50: What not keep using SmallVec?

serde_v8/src/de.rs
AaronO: It's easy to optimize but I'd rather not getting into micro-optimizing serde_v8 without restoring its benches.

So the TODO mainly serves as a motivation to restore `serde_v8`'s micro-benches.

> How big is the difference?

Don't know, it's not benched, probably a few nanoseconds per get.

core/ops_json.rs
AaronO: Virtually no ops used variadic buffers, they had one control buffer and an optional data buffer, control buffer is now superseded by a control/structured arg.

Removing the unnecessary variadicity is cleaner IMO, it's best to drop it if it's not used and also variadic args would be inherently incompatible with v8 Fast API so would hinder moving bin/io ops over, even if it supported typed integer arrays.

This code is just a stopgap till follow-up PR `1.`:
> 1. Change op-sig, removing variadic BufVec

piscisaureus: OK

core/ops.rs
AaronO: Can you elaborate on why you think it might be unsafe ?

piscisaureus: **IF** this is going on:

- create HandleScope A
- create HandleScope B
- allocate Local in HandleScope A
- anything happens with HandleScope B (drop?)

In other words, they "cross" each other, that'd be unsafe (a crash, most likely).
Looks like it doesn't happen in this PR, but only you know the fine print of your code.

A bit of background: `HandleScope`s are RAII objects in C++ and they cannot be moved.
They also shadow each other - an inner `HandleScope` inhibits the creation of a Local handle associated with the outer `HandleScope`. But `Local` handles that already exist in the outer `HandleScope` are safely accessible at all times.

These semantics are pretty much impossible to express in the Rust type system, so we do some extreme acrobatics (https://github.com/denoland/rusty_v8/blob/main/src/scope.rs) to make most calls involving `HandleScope` and other scopes safe, and in the end I *think* we managed. 

But moving a HandleScope into a struct and unloading it later is something I believe we haven't specifically covered, so I am not 100% sure there are no edge cases where the compiler would let you do something that you shouldn't be doing (here's what we have tests for: https://github.com/denoland/rusty_v8/tree/main/tests/compile_fail)

core/runtime.rs
AaronO: Done

core/core.js
```diff
@@ -1,411 +1,124 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 1: Awesome to see this file being 120 lines

core/lib.rs
```diff
@@ -56,12 +55,17 @@ pub use crate::modules::RecursiveModuleLoad;
 pub use crate::normalize_path::normalize_path;
 pub use crate::ops::op_close;
 pub use crate::ops::op_resources;
+pub use crate::ops::serialize_op_result;
```
bartlomieju, line 58: I don't think this function should be public, what's the use case?

core/ops.rs
```diff
@@ -20,19 +22,83 @@ use std::ops::DerefMut;
...
+
+  pub fn empty() -> Self {
+    Self {
+      scope: None,
+      value: None,
+    }
+  }
+
+  pub fn deserialize<T: DeserializeOwned>(self) -> Result<T, AnyError> {
+    serde_v8::from_v8(self.scope.unwrap(), self.value.unwrap())
```
bartlomieju, line 56: Is it expected that this function panics for `OpPayload::empty()`? A bit surpising

core/runtime.rs
```diff
@@ -1391,17 +1382,15 @@ impl JsRuntime {
...
     let state_rc = Self::state(self.v8_isolate());
 
-    let shared_queue_size = state_rc.borrow().shared.size();
-    let overflown_responses_size = overflown_responses.len();
-
-    if shared_queue_size == 0 && overflown_responses_size == 0 {
+    let async_responses_size = async_responses.len();
+    if async_responses_size == 0 {
       return Ok(());
     }
```
bartlomieju, line 1396: Side note: this conditional could potentially be removed after this PR - please see `FIXME` below

core/runtime.rs
```diff
@@ -1501,100 +1495,68 @@ pub mod tests {
...
-        Poll::Ready(val) => return val,
-      }
-    }
-    panic!(
-      "JsRuntime still not ready after polling {} times.",
-      max_poll_count
-    )
-  }
+  // TODO: maybe use this for other tests after dropping the overflow tests?
+  // fn poll_until_ready(
```
bartlomieju, line 1499: I think can safely be removed altogether

bartlomieju (approved):
LGTM 🚀 

Huge thanks @AaronO! This work will be ground-breaking for the overall performance of the runtime and I'm looking forward to removing even more code in the follow up PRs.

core/lib.rs
AaronO: I can make that private, I had left it public since it wasn't 100% clear how we'll approach plugin ops (though it seems like they won't need this and should be able to align plugins with ops without it)

core/ops.rs
AaronO: In production these should never be `None`, we only use `OpPayload::empty()` for the op-table test, so we don't have to spin up a v8 isolate just to test that

core/lib.rs
bartlomieju: Let's make it non-public for now

core/ops.rs
bartlomieju: So maybe add `#[test]` attribute to `OpPayload::empty`

core/runtime.rs
bartlomieju: Yes, there was a specific condition to trigger the problem and it was because of shared queue, but since shared queue is gone it shouldn't happen anymore.

PR #9847, opened 2021-03-21
chore: release crates

bartlomieju:

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -25,7 +25,7 @@ path = "./bench/main.rs"
 
 [build-dependencies]
 deno_core = { path = "../core", version = "0.82.0" }
-deno_runtime = { path = "../runtime", version = "0.9.4" }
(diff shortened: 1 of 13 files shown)


cli/Cargo.toml
```diff
@@ -24,8 +24,8 @@ harness = false
 path = "./bench/main.rs"
 
 [build-dependencies]
-deno_core = { path = "../core", version = "0.81.0" }
-deno_runtime = { path = "../runtime", version = "0.9.3" }
+deno_core = { path = "../core", version = "0.82.0" }
+deno_runtime = { path = "../runtime", version = "0.9.4" }
```
lucacasonato, line 28: This should be a minor bump. If a dependency a crate depends on is bumped with a minor, and the dependent crate reexports that dependency, then the dependent crate _must_ also be bumped a minor release, otherwise we break semver.

bartlomieju: Bumped `deno_runtime` to `0.10.0`

ry (approved):

PR #9850, opened 2021-03-21
feat(plugin_api): Align DispatchOpFn signature to that of OpFn and access to OpState

eliassjogreen:
This pr adds plugin access to the deno resource table through a `get_resource_table` method in the `plugin_api::Interface`. This will be really useful for [webgpu windowing and swapchains](https://github.com/denoland/deno/issues/7863#issuecomment-706897139) but also just for plugins in general not having to use their own resource table solution (usually a global HashMap) for storing resources.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -196,7 +196,7 @@ jobs:
           path: |
             ./target
           key:
-            b-cargo-target-${{ matrix.os }}-${{ matrix.profile }}-${{ hashFiles('Cargo.lock') }}
+            a-cargo-target-${{ matrix.os }}-${{ matrix.profile }}-${{ hashFiles('Cargo.lock') }}
 
       - name: Apply and update mtime cache
         uses: ./.github/mtime_cache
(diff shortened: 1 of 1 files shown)


lucacasonato:
I'm not really super happy with this. I would much prefer if we finally aligned op crates and plugins, and have them have the same signature (the one stating with `deno_core::OpState`).

lucacasonato:
Can you also add a test checks if `deno_core::json_op_async` and `deno_core::json_op_sync` work now?

eliassjogreen:
Sure, it seems to work fine when using both `json_op_sync` and `json_op_async` and the internal `core.jsonOpSync` and `core.jsonOpAsync` functions for calling and getting a response.

test_plugin/src/lib.rs
```diff
@@ -88,3 +102,27 @@ fn op_test_resource_table_close(
...
+  Ok(json!(format!("Hello {}!", args.name)))
+}
+
+pub async fn op_test_json_op_async(
+  _state: Rc<RefCell<OpState>>,
+  args: JsonOpArgs,
+  _bufs: BufVec,
+) -> Result<Value, AnyError> {
+  println!("Hello {} from async json op!", args.name);
+  Ok(json!(format!("Hello {}!", args.name)))
```
littledivy, line 127: Would make sense to do some async tasks inside this op?

eliassjogreen: Makes a lot of sense :P

PR #9864, opened 2021-03-22
feat(bench): added micro benchmarks for ops

inteon:
Benchmark roundtrip js -> rs -> js for different ops and configurations.

Is an alternative for 'http_bench_bin_ops' and 'http_bench_json_ops', less prone to fluctuations (but also less related to real-world applications).

### DIFF
--- a/cli/lsp/capabilities.rs
+++ b/cli/lsp/capabilities.rs
@@ -11,6 +11,7 @@ use lspower::lsp::CodeActionOptions;
 use lspower::lsp::CodeActionProviderCapability;
 use lspower::lsp::CodeLensOptions;
 use lspower::lsp::CompletionOptions;
+use lspower::lsp::FoldingRangeProviderCapability;
 use lspower::lsp::HoverProviderCapability;
 use lspower::lsp::ImplementationProviderCapability;
 use lspower::lsp::OneOf;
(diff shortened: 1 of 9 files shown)


inteon:
image
Just updated the micro-bench for the new serde_v8, clearly shows how much faster the new ops are compared to json_ops & bin_ops.
Not sure about bin_sync ops, might be interesting to find out what is happening there. @AaronO

graph source: https://jsfiddle.net/s9kvyt8q/5/

AaronO:
@inteon Cool graph, this is somewhat similar the `op_baseline` bench in core.

> Not sure about bin_sync ops, might be interesting to find out what is happening there. @AaronO
Can you rerun this on the latest `main` ? Merging  https://github.com/denoland/deno/pull/9944 should improve bin ops quite substantially. By dropping that assertion and retro-compat shim (your benches are going to be hugely penalized if they're hitting that assertion error).

Here's `op_baseline` on `main` (also dropped `nopView` arg in bin op):

```
test bench_op_nop     ... bench:     315,747 ns/iter (+/- 18,115)
test bench_op_pi_bin  ... bench:     292,008 ns/iter (+/- 11,335)
test bench_op_pi_json ... bench:     282,993 ns/iter (+/- 15,699)
```

You can see that it's roughly on par with json, since they essentially have the same implementation and I'll likely drop bin ops in an upcoming PR :)

AaronO:
@inteon Would love to see the updated chart, where bin ops aren't so heavily penalized. It would be a great chart to include in a blog article showcasing the improvements

Issue #9050, opened 2021-01-08
Benchmark CI stage panics and doesn't fail


bartlomieju:
Example: https://github.com/denoland/deno/runs/1666284126

There's a lot of JS code intertwined in the panic backtrace

```
Error: thread 'main' panicked at 'failed printing to stdout: Resource temporarily unavailable (os error 11)', library/std/src/io/stdio.rs:1021:9
stack backtrace:
   0:     0x56439a267cd3 - std::backtrace_rs::backtrace::libunwind::trace::h04d12fdcddff82aa
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/../../backtrace/src/backtrace/libunwind.rs:100:5
            return ***
                y: down.y + 1,
                id: ptr
            ***;
        ***
    ***
<snip>
   1:     0x56439a267cd3 - std::backtrace_rs::backtrace::trace_unsynchronized::h1459b974b6fbe5e1
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x56439a267cd3 - std::sys_common::backtrace::_print_fmt::h9b8396a669123d95
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x56439a267cd3 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::he009dcaaa75eed60
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x564399dad58c - core::fmt::write::h77b4746b0dea1dd3
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/fmt/mod.rs:1078:17
   5:     0x56439a267571 - std::io::Write::write_fmt::heb7e50902e98831c
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/io/mod.rs:1518:15
   6:     0x56439a266f35 - std::sys_common::backtrace::_print::h2d880c9e69a21be9
                               at /rustc/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/std/src/sys_common/backtrace.rs:49:5
[...]

PR #9867, opened 2021-03-23
fix(core): error handling in examples

inteon:
Improved error handling for benchmarks.
Also checks were removed from io.js as these will never trigger (errors are thrown and result is always > 0).


bartlomieju (commented):
Thanks @inteon, but if this PR fixes some problem you should a test case

inteon:
@bartlomieju I added a test for the double borrow error & I fixed the benchmark to also throw an error in case the server crashes.

cli/bench/http.rs
```diff
@@ -109,6 +109,8 @@ fn run(
   println!("{}", wrk_cmd.join(" "));
   let output = test_util::run_collect(wrk_cmd, None, None, None, true).0;
 
+  std::thread::sleep(Duration::from_secs(1)); // wait to capture failure. TODO racy.
```
bartlomieju, line 112: What's the reason for adding this timeout?

cli/bench/deno_tcp.ts
```diff
@@ -13,15 +13,11 @@ async function handle(conn: Deno.Conn): Promise<void> {
...
-      if (r === null) {
-        break;
-      }
+      await conn.read(buffer);
       await conn.write(response);
     }
-  } finally {
-    conn.close();
-  }
+  } catch { /**/ }
```
bartlomieju, line 19: Maybe print put error instead of silently swallowing?

cli/bench/deno_tcp.ts
```diff
@@ -13,15 +13,11 @@ async function handle(conn: Deno.Conn): Promise<void> {
...
+  conn.close();
```
bartlomieju, line 20: You can put it inside `finally` block

cli/bench/deno_tcp_proxy.ts
```diff
@@ -15,14 +15,9 @@ async function handle(conn: Deno.Conn): Promise<void> {
...
-    if (err.message !== "read error" && err.message !== "write error") {
-      throw err;
-    }
-  } finally {
-    conn.close();
-    origin.close();
-  }
+  } catch { /**/ }
+  conn.close();
+  origin.close();
```
bartlomieju, line 20: Ditto

cli/tests/unit/dispatch_bin_test.ts
```diff
@@ -74,3 +74,18 @@ unitTest(function binOpsHeaderTooShort(): void {
...
+      const opFn = ((op === "op_read_sync") && Deno.core.binOpSync) ||
+        Deno.core.binOpAsync;
+      await Promise.resolve(opFn(op, nonExistingRid, new Uint8Array(0)));
+    } catch (e) {
+      if (!(e instanceof Deno.errors.BadResource)) {
+        throw e;
+      }
+    }
+  }
+});
```
bartlomieju, line 91: Please split into two separate cases for sync and async

runtime/js/12_io.js
```diff
@@ -98,29 +95,16 @@
...
-    }
 
     return nread === 0 ? null : nread;
   }
 
   function writeSync(rid, data) {
-    const result = core.binOpSync("op_write_sync", rid, data);
-    if (result < 0) {
-      throw new Error("write error");
-    }
```
bartlomieju, line 112: I don't understand reason for this change, does "bin op" no longer return negative number in case of error?

core/examples/http_bench_json_ops.js
```diff
@@ -41,17 +47,12 @@ function close(rid) {
...
+      await read(rid, requestBuf);
+      await write(rid, responseBuf);
     }
-
-    const nwritten = await write(rid, responseBuf);
-    if (nwritten < 0) {
-      break;
-    }
-  }
+  } catch { /**/ }
```
bartlomieju, line 55: Print out error instead of swallowing

cli/bench/http.rs
inteon: Currently most of the benches actually fail after wrk ends. But this failure is not noticed by rust (because its racy).
The delay is added to give the benchscript some time to fail, so this can be noticed by rust and the ci fails too.

cli/bench/deno_tcp.ts
inteon: I think finally does not catch the error, and since the try-catch block does not contain return statements, finally probably is not needed?

runtime/js/12_io.js
inteon: In case of an error, an error is thrown, so nothing is returned.

inteon: I think this was also how "minimal ops" worked.

cli/bench/http.rs
bartlomieju: Can this problem be related to https://github.com/denoland/deno/issues/9050?

runtime/js/12_io.js
bartlomieju: Makes sense, thanks for explanation

cli/bench/deno_tcp.ts
bartlomieju: Alright, but error shouldn't be silently discarded, can you please print it out?

inteon: I think I've now found the right solution: If the error is not of the type `BrokenPipe` or `ConnectionReset` it will cause the bench to fail.
These two error types have to be tolerated as they will always trigger when using wrk (because it will abruptly close the connection when it reached the end of the benchmark).

inteon: I'm not sure if you want me to still print the BrokenPipe & ConnectionReset, as they will trigger in every run and thus just pollute the logs.

cli/bench/http.rs
inteon: I'm not sure, normally https://github.com/denoland/deno/blob/0a32fa459d8c4398e0d4976cc5e13f1eb2791238/cli/bench/http.rs#L120 should still have killed the server.

cli/bench/deno_tcp.ts
bartlomieju: Looks alright now, thanks for bearing with me!

bartlomieju (approved):
LGTM, thanks @inteon

inteon:
Is now rebased on latest main version (with serde v8 changes).

bartlomieju (approved):
LGTM, thanks @inteon

PR #9872, opened 2021-03-23
build: disable CI cache

bartlomieju:


ry (approved):
No better way to demonstrate how ineffective the cache is. Essentially no slower without cache.

PR #9876, opened 2021-03-23
docs(getting_started/setup_your_environment.md)

IWANABETHATGUY:
unlike other system config file, in windows `$profile` file is a common config file of powershell , just use `>` will overried the user config file , which is terrible. This change use append instead of override which is more reasonable


ry (approved):
thanks

PR #9895, opened 2021-03-25
Upgrade Rust to 1.51.0

magurotuna:
Upgrades Rust to 1.51.0 and apply clippy's suggestions.

PR #9897, opened 2021-03-26
refactor(lsp): remove unused code

magurotuna:
This patch removes the `asset` field of `State` because it seems to be unused.

Taking a glance at the commit history, this field was introduced in https://github.com/denoland/deno/pull/8771, and at that time the field value was used in [the `request_asset` function](https://github.com/denoland/deno/pull/8771/files#diff-5425491838b4e06f099e33c0663a61d3b088c0c193a5a484a9b797c2f4ee099bR1125).
After that, in https://github.com/denoland/deno/pull/8727, both `request_asset` and invoking `op_set_asset` from JS were removed:
- [`request_asset` was removed](https://github.com/denoland/deno/pull/8727/files#diff-5425491838b4e06f099e33c0663a61d3b088c0c193a5a484a9b797c2f4ee099bL1104)
- [invoking `op_set_asset` was removed](https://github.com/denoland/deno/pull/8727/files#diff-917b74af14d03b69b9948cd2366a51c6af8c6582309a4913377fdb0a82993f96L495)

So since then, the `asset` field of `State` struct has seemed to be unused. I wonder why "unused" warning didn't emerge at that time though. (`#[allow(unused)]` was added today in https://github.com/denoland/deno/pull/9895)

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -1339,8 +1339,6 @@ struct Response {
 }
 
 struct State<'a> {
-  #[allow(unused)]
-  asset: Option<String>,
   last_id: usize,
   response: Option<Response>,
   state_snapshot: StateSnapshot,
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
LGTM

PR #9900, opened 2021-03-26
feat(lsp): Implement textDocument/foldingRange

jeanp413:
Related #8643


cli/lsp/tsc.rs
```diff
@@ -1226,6 +1226,99 @@ impl CompletionEntry {
...
+      let offset_end: usize = line_index.offset(range.end).unwrap().into();
+      let fold_end_char: &str =
+        unsafe { text_content.get_unchecked(offset_start..offset_end) };
+      if FOLD_END_PAIR_CHARACTERS.contains(&fold_end_char) {
+        return cmp::max(range.end.line - 1, range.start.line);
+      }
+    }
+
+    range.end.line
+  }
```
kitsonk, line 1307: This was a really ugly hack in TypeScript to work around an issue, and it is even uglier in Rust.  The original issue this was added for: https://github.com/microsoft/vscode/issues/47240 appears to be a hack to workaround some folding issues with tsc.  Personally I would like to try to remove this and see if the example mentioned in the original issue still persists in recent versions of tsc.

If it is still a problem, I would really like to find a safe way to basically get a single character at a position within a string and then compare that to a vector of static chars, instead of this.  All we want is the end offset - 1 and the end offset.

jeanp413: This is still needed as vscode only supports line folding (`lineFoldingOnly` is set to true) and typescript returns line and column
```
/**
 * If set, the client signals that it only supports folding complete lines.
 * If set, client will ignore specified `startCharacter` and `endCharacter`
 * properties in a FoldingRange.
 */
lineFoldingOnly?: boolean;
```

>I would really like to find a safe way to basically get a single character at a position within a string and then compare that to a vector of static chars, instead of this. All we want is the end offset - 1 and the end offset.

This is what I'm trying to do here extracting a valid utf8 code point but I'll try using `as_bytes` as we are just looking for ASCII characters. I'm just learning rust so if there is a better way let me know.

kitsonk: If you don't mind, I will take an attempt at it.  I'm still learning Rust too, but might have been learning it a bit longer. 😁

jeanp413: Sure go ahead :+1: 
In my last commit I managed to reduce it to
```
let offset_end: usize = line_index.offset(range.end).unwrap().into();
let fold_end_char = text_content.as_bytes()[offset_end - 1];
```

kitsonk (approved):
LGTM, thank you again for a great contribution.

PR #9901, opened 2021-03-26
core/bindings: simplify heapStats() by using serde_v8

AaronO:
This uses `serde_v8` to serialize our `HeapStats` data instead of the previous "manual" serialization. This is less verbose and `serde` handles mapping keys to camelCase so the rust code is strongly typed and less prone to typos

This is a follow-up anticipated in https://github.com/denoland/deno/pull/9659, as per [this TODO comment](https://github.com/denoland/deno/pull/9659/files#diff-5b84637fa51710dd4ffc2f2370fb6bb77301603caac646be8184da12eedd7433R949)

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -875,64 +875,13 @@ fn heap_stats(
   _args: v8::FunctionCallbackArguments,
   mut rv: v8::ReturnValue,
 ) {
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - I love it

PR #9902, opened 2021-03-26
refactor(core): decode JsStackFrames using serde_v8

AaronO:
Similar to https://github.com/denoland/deno/pull/9901, this uses `serde_v8` to cleanup decoding `JsStackFrame`s.

## Notes

- I used fully qualified paths (i.e: `serde::Deserialize`, `serde_v8::from_v8`, ...) instead of `use ...`, not sure if this goes against our stylistic guidelines, happy to switch it
- I feel like it's slightly cleaner to use fully qualified paths when there are only a handful of references and the paths are short/straightforward

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -99,6 +99,7 @@ pub struct JsError {
 }
 
 #[derive(Debug, PartialEq, Clone, serde::Deserialize)]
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, nice cleanup!

Issue #9903, opened 2021-03-26
Two async ops resolving exactly at the same time = assert_eq failure when doing a sync op


fasterthanlime:
This is a fun one!

I kept hitting this assertion:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', vendor/deno/core/bindings.rs:416:3
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/panicking.rs:92:14
   2: deno_core::bindings::send
             at ./vendor/deno/core/bindings.rs:416:3
   3: core::ops::function::Fn::call
             at /home/vscode/.rustup/toolchains/1.50.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:70:5
   4: rusty_v8::function::<impl rusty_v8::support::MapFnFrom<F> for extern "C" fn(*const rusty_v8::function::FunctionCallbackInfo)>::mapping::{{closure}}
             at /home/vscode/.cargo/registry/src/github.com-1ecc6299db9ec823/rusty_v8-0.21.0/src/function.rs:280:7
   5: <extern "C" fn(A0) .> R as rusty_v8::support::CFnFrom<F>>::mapping::c_fn
             at /home/vscode/.cargo/registry/src/github.com-1ecc6299db9ec823/rusty_v8-0.21.0/src/support.rs:722:11
   6: _ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
   7: _ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE
   8: _ZN2v88internalL26Builtin_Impl_HandleApiCallENS0_16BuiltinArgumentsEPNS0_7IsolateE
   9: _ZN2v88internal21Builtin_HandleApiCallEiPmPNS0_7IsolateE
[...]

PR #9904, opened 2021-03-26
quickfix #9903 by moving assert into async match branches

AaronO:
This is a quickfix for #9903. It moves the assert into the async branches.

However the shared queue should be dropped very soon and will also fix this issue.

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -412,18 +412,19 @@ fn send<'s>(
   };
 
   let op = OpTable::route_op(op_id, state.op_state.clone(), bufs);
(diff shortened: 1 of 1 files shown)


inteon:
@AaronO I think this might still cause failure, I propose to just remove the check.

fasterthanlime:
@inteon Yup, I can confirm that the quickfix does not address the variant you outlined in https://github.com/denoland/deno/pull/9904#issuecomment-808374022 (I have a local branch of a proprietary codebase I was able to tweak to test that case).

Removing the shared queue sounds good, happy to wait for that — we can live with a fork+cargo-patch for now.

PR #9907, opened 2021-03-26
chore: update wpt

crowlKats:
tests that are failing that are actionable:
`ReadableStream: if pull rejects, it should error the stream`
`TransformStream constructor should not call setters for highWaterMark or size`
Required for #7819

### DIFF
--- a/op_crates/web/03_abort_signal.js
+++ b/op_crates/web/03_abort_signal.js
@@ -14,6 +14,12 @@
     #aborted = false;
     #abortAlgorithms = new Set();
 
+    static abort() {
+      const signal = new AbortSignal(illegalConstructorKey);
+      signal[signalAbort]();
+      return signal;
+    }
+
     [add](algorithm) {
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM - thanks!

kitsonk:
We didn't add a unit test for this, which didn't catch that we didn't update the typings for this.  We also had #9777 open (without CI passing).

From a stylistic perspective, we should put static methods at the end of the class definition.

Can we follow this up with an update to the `lib.deno_web.d.ts`?

lucacasonato:
> We didn't add a unit test for this

Yes we did - WPT tests it.

kitsonk:
>> We didn't add a unit test for this
> Yes we did - WPT tests it.

But it doesn't test the types... and because we don't author in TypeScript anymore, we have no clue that we missed updating the types.  🤷 

The following errors with type checking on:

```ts
AbortSignal.abort()
```

PR #9909, opened 2021-03-26
serde_v8: add Deno copyright headers to all rust files

AaronO:
So `serde_v8` is consistent with other rust files in codebase, cc @bartlomieju


bartlomieju (approved):
LGTM

PR #9915, opened 2021-03-28
feat(bench/deno_common): show ns/op

AaronO:
It's simply the inverse of the rate (ops/s), but it's often useful to look at time per op since it's easier to picture the upper bound and improvable delta.


ry (approved):
LGTM

PR #9921, opened 2021-03-28
docs(getting_started/installation): Add the Nix pkg manager to install options

toraritte:

### DIFF
--- a/docs/getting_started/installation.md
+++ b/docs/getting_started/installation.md
@@ -38,6 +38,12 @@ Using [Homebrew](https://formulae.brew.sh/formula/deno) (macOS):
 brew install deno
 ```
 
+Using [Nix](https://nixos.org/download.html) (macOS and Linux):
+
+```shell
+nix-shell -p deno
+```
+
 Build and install from source using [Cargo](https://crates.io/crates/deno):
 
 ```shell


humancalico:
nix-shell might not be the best way to install it. Maybe this
```sh
nix-env -iA nixpkgs.deno
```
Also the Deno version in stable nixpkgs is pretty old so you might want to add the unstable channel

cc @06kellyjac

06kellyjac:
Thanks for CC-ing me @humancalico
Agreed, although I usually do `nix-env -i <package>` to avoid having to set a channel that might not be there since we don't know if they're on the nixos or nixpkgs channel.

Maybe:
```bash
# add the nixpkgs unstable channel if you're not already using it
# you can check with nix-channel --list
nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
nix-channel --update nixpkgs
nix-env -iA nixpkgs.deno
```

I'm not too au fait with the best way to use `nix-env` as I install 99% of my dependencies in `/etc/nixos/configuration.nix` or using `home-manager` or in a `shell.nix`

--- 

As for the versions being old

Yeah `nixos-unstable` and `nixpkgs-unstable` are on `1.8.2` 
https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/web/deno/default.nix#L18
https://github.com/NixOS/nixpkgs/blob/nixpkgs-unstable/pkgs/development/web/deno/default.nix#L18

`nixos-20.09` is still on `1.3.3` as the newer updates haven't been backported
https://github.com/NixOS/nixpkgs/blob/nixos-20.09/pkgs/development/web/deno/default.nix#L21

I think the next stable release is going to be 21.05 os we've got another month and a bit to wait for that. If it's highly requested I can try and get the current version of deno backported to `20.09`

toraritte:
I just wanted to add Nix to the list as I just found out that it finally has a `deno` package, so please feel free to close this PR, and write whatever you feel is appropriate (if you would like to add it at all, that is).

As for `nix-shell -p`, I only wrote it because I prefer that instead of `nix-env` and it was in my fingers, but the latter is indeed the most canonical one.

Thank you for your times!

06kellyjac:
Thank you @toraritte for providing much needed documentation. I've been meaning to do this for a long while but totally forgot about it

kitsonk (approved):
LGTM

toraritte:
Thanks again to all of you for your time and comments and for merging this!

PR #9924, opened 2021-03-28
perf(core): add op_baseline bench

AaronO:
This PR introduces a benchmark that serves to measure the baseline-overhead of ops (or the per-category upper bound).

This should also serve to:
- Measure gains from https://github.com/denoland/deno/pull/9843 and follow-up PRs
- Prevent regressions

## Output

```
~/git/deno/core bench/op-baseline
❯ cargo bench

...

running 3 tests
test bench_op_nop     ... bench:     603,784 ns/iter (+/- 102,433)
test bench_op_pi_bin  ... bench:     873,331 ns/iter (+/- 39,010)
test bench_op_pi_json ... bench:   3,210,314 ns/iter (+/- 290,248)

test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -195,6 +195,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
+name = "bencher"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7dfdb4953a096c551ce9ace855a604d702e6e62d77fac690575ae347571717f5"
+
+[[package]]
 name = "bit-set"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


AaronO:
Here are the results of this bench on #9843:

```
test bench_op_nop     ... bench:     536,430 ns/iter (+/- 137,708)
test bench_op_pi_bin  ... bench:     779,053 ns/iter (+/- 30,542)
test bench_op_pi_json ... bench:     675,587 ns/iter (+/- 23,755)
```

ry:
Looks good. 

Will bin ops exist after #9843?

AaronO:
@ry They'll still exist immediately after #9843, since I wanted to limit code changes outside of `core/`.

It would now be trivial to collapse all our ops and have only `core.opAsync`, `core.opSync`. But I think it may be worth keeping them around as abstractions, not as implementations, as mentioned in #9843's **Key Ideas**:
> Preserve concepts of json/bin/min as semantic groups of their inputs/outputs instead of their op-encoding strategy, preserving these groups will also facilitate partial transitions over to v8 Fast API for the "min" and "bin" groups

Talking with @bartlomieju, he seemed eager to collapse the extra abstractions. But I think our diverging views may stem from the fact that he associates bin-ops with the implementation and its technical debt, but given that I'm fairly new to `deno` it's easier to dissociate the abstraction/semantic-group from the implementation.

bartlomieju (approved):
LGTM

Issue #9931, opened 2021-03-30
FileFetcher/HttpCache should respect cache control headers


kitsonk:
Current with the Deno dependency cache, we currently don't support cache control headers.  Basically once a module is cached, it is never reloaded until it is deleted, unless the `--reload` flag is specified.  At that point, the cached `ETag` is sent back for efficiency.

While this may not be a huge issue in some cases, some time-sensitive modules and other information will never be invalidated, even though the server is giving clear guidance to how long to persist the data without checking back.  For example a "latest" release of a module or information about registry completions which the Deno language server persists won't ever get updated through "normal use".

We have sufficient controls in place to ensure that people can "control" what is in their cache, including package locks and cache only flags, so it makes sense that we should be more "browser like" and respect the cache headers and allow the hoster to give guidance about the remote modules.

lucacasonato:
I think we have to make a distinction between cached user code, and registry manifests / completions here. Registry manifests and registry completions should respect the cache control headers (and only those). I think however that user code should not. User code should be indefinitely cached, like it is now, and only reloaded when specifying the `--reload` flag. Changing this behaviour would be a huge breaking change.

I think the registry completions and manifests should be cached separately from the user code. We store user code in the `deps` and `gen` folders. We should store registry completions separately in a `registry` folder IMO.

kitsonk:
@lucacasonato that is already being done (the seperate path)...

wperron:
Now that we have SQLite in the dependencies, I wonder if we could be able to leverage it make an efficient index of the cached files and their cache headers?

kitsonk:
We now cache an empty document when the configuration is not available for an origin.  We should also write out a expiration header that is +24 hours, so that we try a maximum of once every 24 hours to check the host.

PR #9934, opened 2021-03-30
feat(lsp): add registry import auto-complete

kitsonk:
Still have a few things to do, but it is in enough of a shape to share.

This add the feature where by LSP clients can specify hosts where the LSP will attempt to retrieve a meta data file which describes how to create remote module specifier URLs for modules located on the server, and allows fetching of the completion items.


lucacasonato (commented):
I have taken it for a spin locally. Some feedback:

1. The sorting of modules is not great. When searching for `https://deno.land/x/s3` these are the results: 
image. The module that is exactly named `s3` is 5th place.
2. Completing on the final `/` in `https://deno.land/x/s3@0.4.0/src/` will complete remote imports, not registry imports. image
3. Completing on the final `d` in `https://deno.land/x/s3@0.4.0/src/d` will complete all paths, not just ones starting with `/src/d` image
4. Accepting the `deps.ts` completion on `https://deno.land/x/s3@0.4.0/src/d` will insert `https://deno.land/x/s3@0.4.0/src/ddeps.ts`.
5. Completing in the middle of a specifier will set the cursor all the way to the end of the specifier.

Regarding the `deno.suggest.imports.reloadCache` option: The v2 branch had http caching, and had a `deno.resetRegistryCache` command. Maybe we could introduce that command instead of `deno.suggest.imports.reloadCache` until #9931 lands?

Another remark: the autoDiscovery option seems to work differently in this version than it did in v2. In v2 when autoDiscovery was enabled, we would try fetch the manifest when we see it for the first time. We would do that at max every 24 hours. If we found a manifest we would prompt the user if they wanted completions from this registry or not. We would persist their response in the `hosts` setting. In this version it seems autoDiscovery is not yet present. Maybe we should just disable that option for now?

Anyway, the core works and that is fantastic. It feels nice and snappy. Great feature to have :-)

kitsonk:
> The sorting of modules is not great. When searching for https://deno.land/x/s3 these are the results:

That is entirely a vscode thing. We can provide alternative sort text, but it is vscode that does it.  There is a little interaction with what text you have typed during the completions that vscode is matching on, I will see if we can "clear" that before a commit char.

> Completing on the final / in https://deno.land/x/s3@0.4.0/src/ will complete remote imports, not registry imports.

That is a bug, will look into that.  I am having to restructure some of the path_to_regex stuff anyways that would allow us to complete each part of a path, part by part as well.

> Completing on the final d in https://deno.land/x/s3@0.4.0/src/d will complete all paths, not just ones starting with

Ok, another bug... I think breaking the parts into components will help with that too...

> Regarding the deno.suggest.imports.reloadCache option: The v2 branch had http caching, and had a deno.resetRegistryCache command. Maybe we could introduce that command instead of deno.suggest.imports.reloadCache until #9931 lands?

I thought about that, the problem is that it isn't a "command" in the sense that we don't clear the cache, we change the cache setting in the file fetcher from `::Use` to `::ReloadAll`, so it only reloads when it tries to load something again.  The command would need to actually try to remove the `/registries` path from the DENO_DIR and that starts to get into all sorts of potential local file system permission issues maybe.

> Another remark: the autoDiscovery option...

Yeah, I didn't understand that, let's get rid of it for now.

kitsonk:
Ok, I think I fixed most of the filtering/sorting issues and other stuff...

lucacasonato:
Just tried it out again, and feels pretty damn great now. Just some minor comments:

1. Versions are listed out of order (oldest first) - the order returned from https://deno.land/_vsc1/modules/s3 is reversed.
2. Accepting a completion always resets the cursor to the end of the specifier, instead of the end of completed part. For example you complete a version, and the cursor is set to the end of the path.

I still think we should switch to a `Deno: Clear import completion cache` instead of a the `Reload cache` setting. It feels really strange to use.

kitsonk:
> the order returned from https://deno.land/_vsc1/modules/s3 is reversed

vscode is sorting them based on the label... I think I can fix that though.

> Accepting a completion always resets the cursor to the end of the specifier, instead of the end of completed part. For example you complete a version, and the cursor is set to the end of the path.

Yeah, I am not sure any good way around that because we are using text edits in the return value to make sure the specifier is proper proper.

kitsonk:
> vscode is sorting them based on the label... I think I can fix that though.

Ok, I can fix that, but it is an all or nothing... either initially listed in the order it is returned from the server or it is "alphabetical"... the protocol doesn't have (yet) a way to indicate what sort of sorting to apply.  So when I change it to server order, some of the other fields look "odd"...

I am fine with use saying "it is up to the server" but some of return values of `deno.land/x/` look odd...

lucacasonato:
I would say for now the server should be authoritative regarding sorting, and later we can extend the registry manifest to decide what sorting to apply for each field.

kitsonk:
Actually most of the oddness was related to me converting the index of item into a string without padding it, so `"1", "2" ... "10" ... "20"` were getting oddly sorted.

cli/lsp/registries.rs
```diff
@@ -0,0 +1,768 @@
...
+use log::error;
+use log::info;
+use lspower::lsp;
+use percent_encoding;
+use regex::Regex;
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::path::Path;
+
+const CONFIG_PATH: &str = "/.well-known/deno-import-intellisense.json";
```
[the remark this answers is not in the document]
kitsonk: I would prefer to land it without requiring existing registry providers to change anything.  It is a Microsoft specific term, but it is a non-user facing bit of information, so... 🤷

kitsonk:
Ready for a proper review now.

cli/lsp/registries.rs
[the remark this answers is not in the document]
lucacasonato: As far as I know it is just deno.land/x and x.lcas.dev that have it set up right now. I would rather fix it now than keeping the cruft around forever.

cli/lsp/completions.rs
```diff
@@ -55,16 +55,28 @@ pub fn get_import_completions(
           items: get_local_completions(specifier, &current_specifier, &range)?,
         }));
       }
-      // completion of modules within the workspace
+      // completion of modules form a module registry or cache
```
lucacasonato, line 58:
```suggestion
      // completion of modules from a module registry or cache
```
```suggestion
      // completion of modules form a module registry or cache
```

cli/lsp/registries.rs
```diff
@@ -0,0 +1,823 @@
...
+  .add(b'&')
+  .add(b'+')
+  .add(b',');
+
+lazy_static::lazy_static! {
+  static ref REPLACEMENT_VARIABLE_RE: Regex =
+    Regex::new(r"\$\{\{?(\w+)\}?\}").unwrap();
+}
+
+fn base_url(url: &Url) -> Result<Url, AnyError> {
```
lucacasonato, line 68: Can you use `Url::origin` instead?

lucacasonato (commented):
Implementation looks great. Tried it out again and found a few more things:
- completing at the end of specifier `https://deno.land/x` brings up "remote" completions, not "registry" completions
- completing in an empty specifier does not bring up `https://deno.land` as completion. completing in a specifier with just `h` does.

Once those are fixed this is good to land. In a follow up we should address the following:
- auto discovery of registries like in v2, with user popup
- respecting HTTP cache headers for completions and manifest

cli/lsp/registries.rs
kitsonk: Url::origin is a tuple that is really hard to work with, especially to transform it back into a Url.

lucacasonato: You can do `Url::origin(&url).ascii_serialization()`. This will give you a string. It matches the output of `URL.origin` in JS. Alternatively you can do `url::quirks::origin(&url)`. This is the same as `Url::origin(&url).ascii_serialization()` under the hood though.

kitsonk: String != Url... 😄   There is also `url[..Position:BeforePath]` if we wanted a string too...  But we need a valid base URL.  Could re-parse the string too, but I don't see an advantage to that.

lucacasonato: Sure, but what we want is the origin of the URL not an arbitrary base url. I don't see why we can't use the provided API for this. Origins are not module specifiers or URLs. They are opaque strings describing the origin of a URL (and happens to also be a valid URL). For example `xn--lt9h.deno.dev` and `🦕.deno.dev` should be the same origin. I think `ModuleRegistry#origins` should be changed to `HashMap<String, Vec<RegistryConfiguration>>`. See the patch below:

```patch

kitsonk:
Ok, I think I have fixed both the issues you mentioned.

Another thing for follow up is handling auto-completions with import maps, though that is more import completions, versus registry completions.

lucacasonato:
I think that last round of fixes introduced a new error: completing at the end of `https://deno.land` inserts `std@` not `/std@` resulting in `https://deno.landstd@`.

kitsonk:
@lucacasonato ok, that is fixed... added a regression test for it too... strange, it should have been broken all along, but I have made that all a bit more robust now.

lucacasonato (approved):
L. G. T. M. 💪 

Thanks for enduring this long review processes. Fantastic feature, and incredibly well implemented. Such a pleasure to use. Let's level up this LSP 🚀

Issue #3995, opened 2020-02-14
Support HTTP/2, HTTP/3 and QUIC in HTTP server


kitsonk:
This was discussed over in denoland/deno_std#153 but was locked during the move.

While we still have concerns around the HTTP server, we should also consider supporting HTTP/2, HTTP/3 and QUIC.

hazae41:
https://github.com/denoland/deno/issues/1558

ry:
So it's quite likely that we will expose native HTTP server. 

Getting these protocols correct is very difficult - it's better to collaborate with other Rust users. We will probably use `warp`.

The reason why we have `std/http` and not a native HTTP server is because we're still working out the kinks in the "op model". In our benchmarks deno_tcp is still slower than node_tcp

![3c49cb](https://user-images.githubusercontent.com/80/75262634-ce021d00-57ba-11ea-8b1c-87a94fe15b06.png)

I want to fix that first before adding the native HTTP since it will be built on top of the same primitives.

hcldan:
@ry Is there a ticket for those performance benchmarks or any schedule for when it might make it in?

Justsnoopy30:
What's the status on http/2 support in Deno?

kitsonk:
The issue is open. That indicates that it isn't resolved yet.

Justsnoopy30:
> The issue is open. That indicates that it isn't resolved yet.

Well yes, I can see that. What I meant was, how much progress has been made on this?

kitsonk:
Any progress would be mentioned in this issue.

Dygear:
[Apple Push Notification Service](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns) is going to require HTTP/2 starting April 1st 2021. Looking over their documentation, it's acceptable to keep a long standing connection to their servers, but requires a `PING FRAME` about every hour to be sent. So whomever does end up landing HTTP/2, can they please add to option to sending these frames manually, it would be of great help.

lucacasonato:
@Dygear This is client to sever communication, right? The API for that is `fetch` and it already supports HTTP/2.

ry:
We are working on hyper bindings and hope to have them complete for v1.9

anonrig:
[...]

PR #9935, opened 2021-03-30
feat: native HTTP bindings

bartlomieju:
- [x] streaming response
- [x] streaming request 
- [x] typescript declarations
- [x] be able to wrap TLS stream

Towards #3995


op_crates/http/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+#[derive(Deserialize)]
+pub struct CreateServerArgs {
+  address: String,
+}
+
+pub async fn op_create_server(
+  state: Rc<RefCell<OpState>>,
+  args: CreateServerArgs,
+  _data: BufVec,
+) -> Result<Value, AnyError> {
```
bartlomieju, line 148: This op should check net permission

bartlomieju: Sure, sounds good

bartlomieju: Actually not, it's a separate crate and doesn't have access to `TcpListenerResource`

lucacasonato: Mh - that is very unfortunate...

bartlomieju: Wasn't planning on doing that in the first pass

lucacasonato: Ok - sounds like we will need to move our net ops into a seperate crate then soon (or move this into `runtime/ops/`?). I think the http crate should _upgrade_ a connection and not duplicate starting of listeners. Anyway not so important for first pass.

core/core.js
```diff
@@ -9,7 +9,7 @@
   let opsCache = {};
   const errorMap = {};
   let nextPromiseId = 1;
-  const promiseTable = new Map();
+  const promiseTable = Object.create(null);
```
bartlomieju, line 12: CC @AaronO my flamegraphs shown that using `Map` still causes significant performance problems on the GC side; changing it back to plain object gives about 5-10% throughput increase

bartlomieju: > @bartlomieju Do you mind if we split that out into a separate PR ?
> 
> Because there are some ideas I want to investigate, potentially something faster than `Map` or `Object` (reducing GC-pressure).

Not at all, interested to hear your ideas

cli/bench/deno_http.js
```diff
@@ -0,0 +1,68 @@
...
+
+  while (true) {
+    const connRid = await Deno.http.accept(rid);
+    handleConn(connRid);
+  }
+}
+
+async function handleConn(connRid) {
+  while (true) {
+    const [connectionClosed, ...req] = await Deno.http.nextRequest(connRid);
```
evanwashere, line 14: should avoid creating new array and using iterator protocol here

cli/bench/deno_http.js
```diff
@@ -0,0 +1,68 @@
...
+    handle(req);
+
+    if (connectionClosed) {
+      break;
+    }
+  }
+}
+
+async function handle(req) {
+  const [requestBodyRid, responseSenderRid, method, headers, url] = req;
```
evanwashere, line 25: avoid using iterator protocol

```suggestion
  const { 0: requestBodyRid, 1: responseSenderRid, 2: method, 3: headers, 4: url } = req;
```

cli/bench/deno_http.js
```diff
@@ -0,0 +1,68 @@
...
+  // If response body is Uint8Array it will be sent synchronously
+  // in a single op, in other case a "response body" resource will be
+  // created and we'll be streaming it.
+  const body = resp.body ?? new Uint8Array();
+  const zeroCopyBufs = body instanceof Uint8Array ? [body] : [];
+
+  const responseBodyRid = Deno.http.respond(
+    responseSenderRid,
+    resp,
+    ...zeroCopyBufs,
```
evanwashere, line 43: avoid using spread operator as arguments length is known to be either 0 or 1 (could also avoid alloc of empty array by propagating it as undefined)

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,63 @@
...
+    headers: { "content-type": "text/plain" },
+    body: Deno.core.encode("ok"),
+  };
+
+  const zeroCopyBufs = resp.body instanceof Uint8Array ? [resp.body] : [];
+
+  const { responseBodyRid } = Deno.http.respond(
+    responseSenderRid,
+    resp,
+    ...zeroCopyBufs,
```
evanwashere, line 44: same as above

op_crates/http/00_http.js
```diff
@@ -0,0 +1,48 @@
...
+  function readRequest(requestRid) {
+    return Deno.core.jsonOpAsync("op_http_read_request", requestRid);
+  }
+
+  function respond(responseSenderRid, resp, ...zeroCopyBufs) {
+    return Deno.core.jsonOpSync("op_http_respond", [
+      responseSenderRid,
+      resp.status ?? 200,
+      Object.entries(resp.headers ?? {}),
+    ], ...zeroCopyBufs);
```
evanwashere, line 29: same as above

cli/bench/deno_http.js
AaronO: Yeah, I had removed this in my local branch. Buffers are no longer variadic so the `...zeroCopyBufs` pattern is discouraged / not useful

op_crates/http/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+      } else {
+        None
+      };
+
+      let mut state = state.borrow_mut();
+      let response_sender_rid =
+        state.resource_table.add(ResponseSenderResource(tx));
+
+      let req_json = NextRequestResponse(
+        matches!(poll_result, Poll::Ready(Ok(()))),
```
bartlomieju, line 258: This is a bit tricky and I'm pretty sure we've handled this situation wrong. If polling the connection returns `Poll::Ready(Ok(())` it means that the connection has been closed. If so it doesn't really make sense to return request to JS because we won't be able to respond to it anyway (because the connection is already closed). @ry thoughts?

cli/bench/deno_http.js
bartlomieju: Not sure what you're suggesting here, care to explain?

bartlomieju: Removed destructuring altogether

op_crates/http/lib.rs
lucacasonato: Not all requests are necessarily about the response. For some requests you care more about the request than the response. I think we should pass all requests we have received into the isolate, even if the user won't be able to respond to them.

AaronO: Yeah, a good example use-case is "fire and forget" logging, where users just care about pushing an event into a queue

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,126 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { assert, assertEquals, deferred, unitTest } from "./test_util.ts";
+import { BufReader, BufWriter } from "../../../test_util/std/io/bufio.ts";
+import { TextProtoReader } from "../../../test_util/std/textproto/mod.ts";
+
+unitTest({ perms: { net: true } }, async function httpServer() {
+  const promise = deferred<void>();
+
+  (async () => {
```
nayeemrmn, line 9: We don't need `deferred()` here.
```suggestion
  const promise = (async () => {
```
Ditto for other tests.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1196,6 +1196,21 @@ declare namespace Deno {
...
+
+  /** **UNSTABLE**: new API, yet to be vetted.
+   *
+   * Parse HTTP requests from the given connection
+   *
+   * ```ts
+   * const requests = await Deno.startHttp(conn);
+   * ```
+   */
+  export function startHttp(conn: Conn): AsyncIterableIterator<HttpEvent>;
```
nayeemrmn, line 1213: We should downgrade this return type to `AsyncIterable<HttpEvent>`, and now or in the future upgrade that to `HttpConn extends AsyncIterable<HttpEvent>` which exposes `.rid` and `.close()` if we want those things.

Returning or extending `AsyncIterableIterator<_>` (and overloading `.return()` for closing) isn't the way to go. I pushed against that for `Listener` early on, but unfortunately `watchFs()` slipped through and demonstrated this. I've been meaning to fix that for 2.0.

runtime/js/40_http.js
```diff
@@ -0,0 +1,215 @@
...
+
+        // Once all chunks are sent, and the request body is closed, we can close
+        // the response body.
+        core.close(responseBodyRid);
+        await Deno.core.jsonOpAsync("op_http_poll", connRid);
+      }
+    };
+  }
+
+  class RequestBodyStream extends ReadableStream {
```
lucacasonato, line 175: This is not spec compliant. The prototype chain will be wrong, and webidl won't be happy about it. Please revert to the plain `ReadableStream`. Instead of a class you can make this a helper function that constructs the `ReadableStream`.

ry: fixed

cli/dts/lib.deno.unstable.d.ts
ry: > Returning or extending AsyncIterableIterator<_> (and overloading .return() for closing) isn't the way to go

What's your reasoning behind this?

cli/tests/unit/http_test.ts
ry: thanks - fixed

cli/dts/lib.deno.unstable.d.ts
nayeemrmn: For starters `.return()` is optional on `AsyncIterableIterator`: https://github.com/denoland/deno/blob/main/cli/tests/unit/fs_events_test.ts#L74. It doesn't align with the default implementation of `.return()` on generators, which is supposed to break any pending loops without error. I don't think this is supposed to be manually implemented.

The iterator protocol methods like `.next()` aren't a nice thing to directly expose on structures with other things, it would be better to have just `[Symbol.asyncIterator]()` and maybe add an `accept()` method to await one request like with `Listener`. However we don't actually need to adjust the implementation from what it is now since it's probably more efficient -- that would also match `Listener`.

ry: Ok. I have changed this around so that there is no `return` method and that the requests iterator must be closed manually. Is this what you had in mind?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1196,6 +1196,34 @@ declare namespace Deno {
     bytesSentData: number;
     bytesReceived: number;
   }
+
+  export interface RequestEvent {
+    readonly request: Request;
+    respondWith(r: Response | Promise<Response>): void;
+  }
+
+  export interface RequestIter extends AsyncIterator<RequestEvent> {
```
nayeemrmn, line 1205:
```suggestion
  export interface RequestIter extends AsyncIterable<RequestEvent> {
```
I would also rename it to `HttpConn` or similar, it has `.rid` and `.close()` so it should be named like a resource.

ry: done

cli/dts/lib.deno.unstable.d.ts
nayeemrmn: > and that the requests iterator must be closed manually

The resource should still be closed automatically if `connectionClosed` is true, if I seemed to have suggested otherwise.

runtime/js/40_http.js
```diff
@@ -0,0 +1,203 @@
...
+    async next() {
+      try {
+        const [
+          connectionClosed,
+          requestBodyRid,
+          responseSenderRid,
+          method,
+          headersList,
+          url,
+        ] = await Deno.core.jsonOpAsync("op_http_request_next", this.#rid);
```
devcat, line 50:
```suggestion
        const {
          0: connectionClosed,
          1: requestBodyRid,
          2: responseSenderRid,
          3: method,
          4: headersList,
          5: url,
        } = await Deno.core.jsonOpAsync("op_http_request_next", this.#rid);
```

bartlomieju: Is there any advantage to using this syntax? Destructuring this array doesn't seem to have performance hit according to flamegraphs

devcat: keyed lookups are much faster and cheaper than creating iterator and collecting it
flamegraph won't likely show diff in this case as it's very small change compared to everything else

but the diff can be clearly seen in bytecode


keyed: `const { 0: zero: 1: one } = ()`
```
[generated bytecode for function: aaae (0x0c85eeda8e89 <SharedFunctionInfo aaae>)]
Parameter count 1
Register count 4
Frame size 32
   77 S> 0xc85eeda96d6 @    0 : 1b 02             LdaImmutableCurrentContextSlot [2]
         0xc85eeda96d8 @    2 : ab 00             ThrowReferenceErrorIfHole [0]
         0xc85eeda96da @    4 : 26 f9             Star r2
   60 S> 0xc85eeda96dc @    6 : 0b                LdaZero 
         0xc85eeda96dd @    7 : 26 f8             Star r3
         0xc85eeda96df @    9 : 2a f9 00          LdaKeyedProperty r2, [0]
         0xc85eeda96e2 @   12 : 26 fb             Star r0
   69 S> 0xc85eeda96e4 @   14 : 0c 01             LdaSmi [1]
         0xc85eeda96e6 @   16 : 26 f8             Star r3
         0xc85eeda96e8 @   18 : 2a f9 02          LdaKeyedProperty r2, [2]
         0xc85eeda96eb @   21 : 26 fa             Star r1
         0xc85eeda96ed @   23 : 0d                LdaUndefined 
   84 S> 0xc85eeda96ee @   24 : aa                Return 
Constant pool (size = 1)
0xc85eeda9689: [FixedArray] in OldSpace
 - map: 0x2bc831201721 <Map>
 - length: 1
           0: 0x2b156b67c129 <String[5]: #array>
Handler Table (size = 0)
Source Position Table (size = 12)
0x0c85eeda96f1 <ByteArray[12]>
```

iterator: `const [zero, one] = ()`
```
[generated bytecode for function: aaae (0x35e99e468e89 <SharedFunctionInfo aaae>)]
Parameter count 1
Register count 13
Frame size 104
   69 S> 0x35e99e469716 @    0 : 1b 02             LdaImmutableCurrentContextSlot [2]
         0x35e99e469718 @    2 : ab 00             ThrowReferenceErrorIfHole [0]
         0x35e99e46971a @    4 : 26 f9             Star r2
   69 E> 0x35e99e46971c @    6 : b1 f9 00 02       GetIterator r2, [0], [2]
         0x35e99e469720 @   10 : a0 07             JumpIfJSReceiver [7] (0x35e99e469727 @ 17)
         0x35e99e469722 @   12 : 61 b8 00 fb 00    CallRuntime [ThrowSymbolIteratorInvalid], r0-r0
         0x35e99e469727 @   17 : 26 f7             Star r4
         0x35e99e469729 @   19 : 28 f7 01 04       LdaNamedProperty r4, [1], [4]
         0x35e99e46972d @   23 : 26 f8             Star r3
         0x35e99e46972f @   25 : 11                LdaFalse 
         0x35e99e469730 @   26 : 26 f6             Star r5
         0x35e99e469732 @   28 : 27 ff f3          Mov <context>, r8
   56 S> 0x35e99e469735 @   31 : 25 f6             Ldar r5
         0x35e99e469737 @   33 : 97 25             JumpIfToBooleanTrue [37] (0x35e99e46975c @ 70)
         0x35e99e469739 @   35 : 10                LdaTrue 
         0x35e99e46973a @   36 : 26 f6             Star r5
         0x35e99e46973c @   38 : 58 f8 f7 0a       CallProperty0 r3, r4, [10]
         0x35e99e469740 @   42 : 26 f2             Star r9
         0x35e99e469742 @   44 : a0 07             JumpIfJSReceiver [7] (0x35e99e469749 @ 51)
         0x35e99e469744 @   46 : 61 b0 00 f2 01    CallRuntime [ThrowIteratorResultNotAnObject], r9-r9
         0x35e99e469749 @   51 : 28 f2 02 08       LdaNamedProperty r9, [2], [8]
         0x35e99e46974d @   55 : 97 0f             JumpIfToBooleanTrue [15] (0x35e99e46975c @ 70)
         0x35e99e46974f @   57 : 28 f2 03 06       LdaNamedProperty r9, [3], [6]
         0x35e99e469753 @   61 : 26 f2             Star r9
         0x35e99e469755 @   63 : 11                LdaFalse 
         0x35e99e469756 @   64 : 26 f6             Star r5
         0x35e99e469758 @   66 : 25 f2             Ldar r9
         0x35e99e46975a @   68 : 8b 03             Jump [3] (0x35e99e46975d @ 71)
         0x35e99e46975c @   70 : 0d                LdaUndefined 
         0x35e99e46975d @   71 : 26 fb             Star r0
   62 S> 0x35e99e46975f @   73 : 25 f6             Ldar r5
         0x35e99e469761 @   75 : 97 25             JumpIfToBooleanTrue [37] (0x35e99e469786 @ 112)
         0x35e99e469763 @   77 : 10                LdaTrue 
         0x35e99e469764 @   78 : 26 f6             Star r5
         0x35e99e469766 @   80 : 58 f8 f7 0c       CallProperty0 r3, r4, [12]
         0x35e99e46976a @   84 : 26 f2             Star r9
         0x35e99e46976c @   86 : a0 07             JumpIfJSReceiver [7] (0x35e99e469773 @ 93)
         0x35e99e46976e @   88 : 61 b0 00 f2 01    CallRuntime [ThrowIteratorResultNotAnObject], r9-r9
         0x35e99e469773 @   93 : 28 f2 02 08       LdaNamedProperty r9, [2], [8]
         0x35e99e469777 @   97 : 97 0f             JumpIfToBooleanTrue [15] (0x35e99e469786 @ 112)
         0x35e99e469779 @   99 : 28 f2 03 06       LdaNamedProperty r9, [3], [6]
         0x35e99e46977d @  103 : 26 f2             Star r9
         0x35e99e46977f @  105 : 11                LdaFalse 
         0x35e99e469780 @  106 : 26 f6             Star r5
         0x35e99e469782 @  108 : 25 f2             Ldar r9
         0x35e99e469784 @  110 : 8b 03             Jump [3] (0x35e99e469787 @ 113)
         0x35e99e469786 @  112 : 0d                LdaUndefined 
         0x35e99e469787 @  113 : 26 fa             Star r1
         0x35e99e469789 @  115 : 0c ff             LdaSmi [-1]
         0x35e99e46978b @  117 : 26 f4             Star r7
         0x35e99e46978d @  119 : 26 f5             Star r6
         0x35e99e46978f @  121 : 8b 07             Jump [7] (0x35e99e469796 @ 128)
         0x35e99e469791 @  123 : 26 f4             Star r7
         0x35e99e469793 @  125 : 0b                LdaZero 
         0x35e99e469794 @  126 : 26 f5             Star r6
         0x35e99e469796 @  128 : 0f                LdaTheHole 
         0x35e99e469797 @  129 : a7                SetPendingMessage 
         0x35e99e469798 @  130 : 26 f3             Star r8
         0x35e99e46979a @  132 : 25 f6             Ldar r5
         0x35e99e46979c @  134 : 97 26             JumpIfToBooleanTrue [38] (0x35e99e4697c2 @ 172)
         0x35e99e46979e @  136 : 27 ff f1          Mov <context>, r10
         0x35e99e4697a1 @  139 : 28 f7 04 0e       LdaNamedProperty r4, [4], [14]
         0x35e99e4697a5 @  143 : 9f 1d             JumpIfUndefinedOrNull [29] (0x35e99e4697c2 @ 172)
         0x35e99e4697a7 @  145 : 26 f0             Star r11
         0x35e99e4697a9 @  147 : 58 f0 f7 10       CallProperty0 r11, r4, [16]
         0x35e99e4697ad @  151 : a0 15             JumpIfJSReceiver [21] (0x35e99e4697c2 @ 172)
         0x35e99e4697af @  153 : 26 ef             Star r12
         0x35e99e4697b1 @  155 : 61 b0 00 ef 01    CallRuntime [ThrowIteratorResultNotAnObject], r12-r12
         0x35e99e4697b6 @  160 : 8b 0c             Jump [12] (0x35e99e4697c2 @ 172)
         0x35e99e4697b8 @  162 : 26 f1             Star r10
         0x35e99e4697ba @  164 : 0b                LdaZero 
         0x35e99e4697bb @  165 : 6d f5             TestReferenceEqual r6
         0x35e99e4697bd @  167 : 99 05             JumpIfTrue [5] (0x35e99e4697c2 @ 172)
         0x35e99e4697bf @  169 : 25 f1             Ldar r10
         0x35e99e4697c1 @  171 : a9                ReThrow 
         0x35e99e4697c2 @  172 : 25 f3             Ldar r8
         0x35e99e4697c4 @  174 : a7                SetPendingMessage 
         0x35e99e4697c5 @  175 : 0b                LdaZero 
         0x35e99e4697c6 @  176 : 6d f5             TestReferenceEqual r6
         0x35e99e4697c8 @  178 : 9a 05             JumpIfFalse [5] (0x35e99e4697cd @ 183)
         0x35e99e4697ca @  180 : 25 f4             Ldar r7
         0x35e99e4697cc @  182 : a9                ReThrow 
         0x35e99e4697cd @  183 : 0d                LdaUndefined 
   76 S> 0x35e99e4697ce @  184 : aa                Return 
Constant pool (size = 5)
0x35e99e4696a9: [FixedArray] in OldSpace
 - map: 0x37d138201721 <Map>
 - length: 5
           0: 0x241b4b3bc129 <String[5]: #array>
           1: 0x37d138204b71 <String[4]: #next>
           2: 0x37d1382043a9 <String[4]: #done>
           3: 0x37d1382053d1 <String[5]: #value>
           4: 0x37d138204f59 <String[6]: #return>
Handler Table (size = 32)
   from   to       hdlr (prediction,   data)
  (  31, 115)  ->   123 (prediction=0, data=8)
  ( 139, 160)  ->   162 (prediction=0, data=10)
Source Position Table (size = 14)
0x35e99e4697d1 <ByteArray[14]>
```

ry: @devcat Once we have this basic thing landed, performance hacks like this will be very welcome. For the moment, we're more concerned about error handling, API, and semantics. Please submit in a follow up. I expect this patch to land today.

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,183 @@
...
+    const writer = stream.writable.getWriter();
+    writer.write(new TextEncoder().encode("hello "));
+    writer.write(new TextEncoder().encode("world"));
+    writer.close();
+
+    const promise = (async () => {
+      const listener = Deno.listen({ port: 4501 });
+      const conn = await listener.accept();
+      const requests = Deno.startHttp(conn);
+      const { value: { request, respondWith } } = await requests.next();
```
nayeemrmn, line 69:
```suggestion
      const { value: { request, respondWith } } = await requests[Symbol.asyncIterator]().next();
```

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,183 @@
...
+      const reqBody = await request.text();
+      assertEquals("hello world", reqBody);
+      await respondWith(new Response(""));
+
+      // TODO(ry) If we don't call requests.next() here we get "error sending
+      // request for url (https://localhost:4501/): connection closed before
+      // message completed".
+      const { done } = await requests.next();
```
nayeemrmn, line 77:
```suggestion
      const { done } = await requests[Symbol.asyncIterator]().next();
```

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,183 @@
...
+  assert(chunk3.done);
+  await promise;
+});
+
+unitTest({ perms: { net: true } }, async function httpServerClose() {
+  const promise = (async () => {
+    const listener = Deno.listen({ port: 4501 });
+    const conn = await listener.accept();
+    const requests = Deno.startHttp(conn);
+    const { done } = await requests.next();
```
nayeemrmn, line 134:
```suggestion
    const { done } = await requests[Symbol.asyncIterator]().next();
```

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,183 @@
...
+    const promise = (async () => {
+      const listener = Deno.listenTls({
+        hostname,
+        port,
+        certFile: "cli/tests/tls/localhost.crt",
+        keyFile: "cli/tests/tls/localhost.key",
+      });
+      const conn = await listener.accept();
+      const requests = Deno.startHttp(conn);
+      const { value: { request, respondWith } } = await requests.next();
```
nayeemrmn, line 159:
```suggestion
      const { value: { request, respondWith } } = await requests[Symbol.asyncIterator]().next();
```

cli/tests/unit/http_test.ts
```diff
@@ -0,0 +1,183 @@
...
+      await respondWith(new Response("Hello World"));
+
+      // TODO(ry) If we don't call requests.next() here we get "error sending
+      // request for url (https://localhost:4501/): connection closed before
+      // message completed".
+      const { done } = await requests.next();
```
nayeemrmn, line 165:
```suggestion
      const { done } = await requests[Symbol.asyncIterator]().next();
```

nayeemrmn (commented):
This is how to write the rewrite the `.next()` calls with `AsyncIterable<_>`. But to have a proper API for receiving one request, I suggest adding an equivalent of `Listener.accept()`.

ry:
@nayeemrmn That API seems rather annoying. Why shouldn't it be an `AsyncIterableIterator`?

lucacasonato:
I think generally nayeem is on the right path. It makes sense to me if the API looked like this:

```ts
interface HttpConn extends AsyncIterable<RequestEvent> {
    readonly rid: number;

    next(): Promise<RequestEvent | null>;
    close(): void;
}

**Comment by lucacasonato** [comment]
`next` returning `null` could be controversial. It would replace the `BadResource` error for the `next` call. We could also do the same as `listener` though, where we actually throw the `BadResource` error. Returning `null` would align to `Stream#next` returns an `Option` in Rust though.

**Comment by nayeemrmn** [comment]
I'm in favour of the `BadResource` error over being nullable. If you're trying to "receive one", you probably have an assertion that you'll get it and a possible error aligns more with that. It should be a more direct implementation in terms of calling the op. The verbose iterator protocol is still accessible through `httpConn[Symbol.asyncIterator]()` so the nice API should go in the opposite direction. Plus it's more consistent for now.

```ts
interface HttpConn extends AsyncIterable<RequestEvent> {
    readonly rid: number;

    accept(): Promise<RequestEvent>;
    close(): void;
}
```

runtime/ops/http.rs
```diff
@@ -0,0 +1,504 @@
...
+
+  fn poll_ready(
+    &mut self,
+    _cx: &mut Context<'_>,
+  ) -> Poll<Result<(), Self::Error>> {
+    if self.inner.borrow().is_some() {
+      Poll::Pending
+    } else {
+      Poll::Ready(Ok(()))
+    }
```
bartlomieju, line 81: I believe is `Poll::Pending` is returned this task will never be woken up again

runtime/js/40_http.js
```diff
@@ -0,0 +1,210 @@
...
+    );
+  }
+
+  function respond(responseSenderRid, resp, zeroCopyBuf) {
+    return Deno.core.jsonOpSync("op_http_response", [
+      responseSenderRid,
+      resp.status ?? 200,
+      flatEntries(resp.headers ?? {}),
+    ], zeroCopyBuf);
+  }
```
bartlomieju, line 113: For a follow up: these two functions seem superfluous

bartlomieju (commented):
Ship it 🚀

ry (approved):
LGTM!

Although this looks like one large PR, this is actually the cumulation of months of attempts at building a binding to hyper. @bartlomieju drove this to completion but various proof-of-concepts were done by @bnoordhuis, @lucacasonato. The binding relies heavily on the serde_v8 optimization work that @AaronO has been doing - this would have been a much larger and slower undertaking without serde_v8.

PR #9936, opened 2021-03-30
feat(cli/compile): Support data uri dynamic imports in `deno compile`

### LINKED ISSUE #9397: Base64 Loading in Binaries
Hi there,
While I was working on Leaf I found out that when using dynamic `import()` from base64 URLs in self contained binaries it immediately errors out. 
image

I think self contained binaries should be able to load from base64 URLs.

littledivy:
Closes #9397


cli/standalone.rs
```diff
@@ -109,6 +109,16 @@ fn read_string_slice(
 
 const SPECIFIER: &str = "file://$deno$/bundle.js";
 
+fn module_supported(specifier: &str) -> Result<(), AnyError> {
+  if specifier == SPECIFIER || specifier.starts_with("data:text/javascript;") {
```
shadowtime2000, line 113: Idk if this necessary but maybe change the specifier starts with to `"data:text/javascript;base64,"`.

littledivy: Specifiers can also start with ` data:text/javascript;charset=utf-8` IIRC

andreespirela: it should be enough though, since it's starting with...

cli/standalone.rs
```diff
@@ -109,6 +109,16 @@ fn read_string_slice(
 
 const SPECIFIER: &str = "file://$deno$/bundle.js";
 
+fn module_supported(specifier: &str) -> Result<(), AnyError> {
```
kitsonk, line 112: This should be much more like the logic in `file_fetcher.rs`, like: https://github.com/denoland/deno/blob/main/cli/file_fetcher.rs#L153 and actually hopefully leveraging the same logic, just making it `pub(crate)` as needed, especially detecting if it is a valid data URL.

littledivy: `file_fetcher.rs` pulls with itself a lot of extra dependencies that `main_runtime` won't ever need. 
```diff
+ mod file_fetcher;
+ mod text_encoding;
+ mod media_type;
+ mod http_util;
+ mod auth_tokens;
+ mod http_cache;
+ mod fs_util;
+ mod checksum;
```
Maybe we can just have a stripped down version of `file_fetcher::get_source_from_data_url` for only validating data URL and returning the source.

littledivy: Ended up making a shared `mod data_url` between _cli_ and _denort_.

bartlomieju:
@littledivy the code looks good, cleanly implemented 👍  

That said, recently there has been a lot of PRs to add small features to `deno compile`. There are requests to support dynamic imports in general, as well as bundling native plugins. 

I would rather avoid landing discrete changes to what we allow to load in compiled binaries across multiple releases and overhaul it completely based on users' needs as part of 1.10 milestone.

bartlomieju:
@littledivy please rebase

lucacasonato (approved):
LGTM

mweichert:
So happy to see this merged! Thanks so much!

@littledivy any possibility of adding this to web workers as well?  :P

PR #9938, opened 2021-03-30
fix: enforce webgpu resource inheritance

lucacasonato:
This commit changes the resources in WebGPU to enforce inhertiance.
This is done using a tree of `Rc`'s, and only letting resources be
closed if at most one strong reference remains.

This is a precursor to safely being able to share buffers between WebGPU
and V8.

### DIFF
--- a/op_crates/webgpu/binding.rs
+++ b/op_crates/webgpu/binding.rs
@@ -9,22 +9,53 @@ use deno_core::ZeroCopyBuf;
 use deno_core::{OpState, Resource};
 use serde::Deserialize;
 use std::borrow::Cow;
+use std::rc::Rc;
+
+use crate::Instance;
 
 use super::error::WebGpuError;
 
-pub(crate) struct WebGpuBindGroupLayout(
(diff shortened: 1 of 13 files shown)


op_crates/webgpu/binding.rs
```diff
@@ -9,23 +9,54 @@ use deno_core::ZeroCopyBuf;
...
+}
 impl Resource for WebGpuBindGroupLayout {
   fn name(&self) -> Cow<str> {
     "webGPUBindGroupLayout".into()
   }
+
+  fn close(self: Rc<Self>) {
+    let resource = Rc::try_unwrap(self)
+      .map_err(|_| "closed webGPUBindGroupLayout while in use")
+      .unwrap();
```
lucacasonato, line 31: It is obviously not great that we panic here, but a) we can't return errors from `Resource#close` calls, and b) these resources should never be closed by the user directly. If they do and Deno panics, they might learn to not do that :-)

ry:
Closing because stale.

PR #9944, opened 2021-03-31
refactor(ops): remove variadic buffers

AaronO:
This is a follow-up (`1.`) of #9843, that removes variadic buffers from ops, replacing them with a single optional buffer.

Since #9843, `OpFn` itself no longer accepts variadic buffer args, but we kept `op_json_a?sync()` and similar wrappers retrocompatible.

## Why ?

- Variadic buffers aren't used in practice
- Variadic buffers wouldn't be v8 Fast API compatible

So it's cleaner to drop them if they're not genuinely needed.

## Changes

- [x] Adapt all op signatures
  - [x] Ops that don't use buffers
  - [x] Ops that do
- [x] Adapt inner-op code of ops using their buffer
  - [x] Assert on `.is_some()` vs len
  - [x] Move this from an assertion/panic to a TypeError ?
  - [x] Maybe provide an `assert_opbuffer` for consistency
- [x] Drop stopgap assertion and mapping to BufVec in `core/ops`
- [x] Maybe drop `BufVec` type, since now unused
- [x] Ensure all tests pass


core/ops.rs
```diff
@@ -99,6 +99,16 @@ pub fn serialize_op_result<R: Serialize + 'static>(
   }))
 }
 
+/// Provides a convenient means for ops to assert if a buffer arg is passed
+pub fn assert_opbuf(
```
lucacasonato, line 103: I don't like functions named `assert` that don't panic. Maybe `opbuf_or_err`?

ry: Agree - also I don't think this belongs in core.

AaronO: > Agree - also I don't think this belongs in core.
> maybe `unwrap_or_type_error`

Where else would you put it ?

It's closely related to ops and potentially consumed by "all" ops ?

hayd: The rusty naming might be ok_or_type_error /bikeshed

https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

AaronO: > The rusty naming might be ok_or_type_error /bikeshed
> 
> https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

Given that it's not a method, would that make it `opbuf_ok_or_type_error` ?
(Sounds a bit longwinded)

Part of the reason I used `assert` is that some of these actually were asserts before and some threw type errors

ry: > It's closely related to ops and potentially consumed by "all" ops.

I suppose core is fine...

AaronO: > > It's closely related to ops and potentially consumed by "all" ops.
> 
> I suppose core is fine...

I'm genuinely open to other suggestions, but given that it's shared/used by `op_crates`, `runtime`, etc... there doesn't really appear a better place to put it.

AaronO: @ry @hayd @lucacasonato I've replaced `assert_opbuf(...)?` with `.ok_or_else(null_opbuf)?` where `null_opbuf` comes from `deno_core::error`

AaronO:
This should be good to merge: CI is green and I dropped the `assert_opbuf(...)` in favour of `.ok_or_else(null_opbuf)`

ry (approved):
LGTM - thanks Aaron

PR #9947, opened 2021-03-31
core: js errors as unions vs tuples to reduce allocs

AaronO:
This is an op-layer optimization follow-up to #9843.

It shaves off `~350ns/op` of baseline overhead, making them over 2x faster by reducing allocs.

Instead of returning JS results as tuples of `(Option<T>, Option<Err>)` it returns them as a conceptual union type, avoiding allocating the wrapping array. We then use a "special" key name to recognize errors.

So far I've picked the error-key to be `$err_class_name`, but this can be easily changed or adjusted.

## Results

Op-baseline benches:
```
Before:
test bench_op_nop     ... bench:     321,740 ns/iter (+/- 39,892)
test bench_op_pi_bin  ... bench:     786,395 ns/iter (+/- 23,765)
test bench_op_pi_json ... bench:     681,285 ns/iter (+/- 26,579)

After:
test bench_op_nop     ... bench:     318,910 ns/iter (+/- 16,233)
test bench_op_pi_bin  ... bench:     402,301 ns/iter (+/- 28,707)
test bench_op_pi_json ... bench:     294,627 ns/iter (+/- 14,079)
```

### DIFF
--- a/core/core.js
+++ b/core/core.js
@@ -45,17 +45,18 @@
 
   function processResponse(res) {
     // const [ok, err] = res;
-    if (res[1] === null) {
+    if (!res.$err_class_name) {
       return res[0];
     }
-    throw processErr(res[1]);
+    throw processErr(res);
   }
 
   function processErr(err) {
-    const [ErrorClass, args] = getErrorClassAndArgs(err.className);
+    const className = err.$err_class_name;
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

PR #9951, opened 2021-04-01
fix(lsp): ensure insert_text is passed back on completions

### LINKED ISSUE #9920: LSP: intellisense inserts `?` after every completion
All completions for optional fields of an object insert a `?` after the property name in the completion. Issue here is likely that completion label should have the `?` to mark that the completion is optional, but the actual edit text should not. Present in canary. Was introduced very recently, likely in #9789

kitsonk:
Fixes #9920 

The fix ended up being simple, but still added a regression test case anyways.  Might fix a few other things that were strange.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -3025,6 +3025,68 @@ mod tests {
     harness.run().await;
   }
 
+  #[tokio::test]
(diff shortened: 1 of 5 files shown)


lucacasonato (approved):
Thanks!

PR #9953, opened 2021-04-01
fix(cli/dts): fix futime and futime example

majidsajadi:
I was working on `deno_std` and I saw an example for `futime` and `futimeSync`. as far as I know `Deno.openSync` needs `write` option alongside with `create`/`append` option. so I updated the example. let me know if I'm wrong.

ref: 
https://github.com/denoland/deno/blob/d6d5ced1ab90404d25a87a242ee5b03ca655abe5/cli/dts/lib.deno.ns.d.ts#L812


kt3k (approved):
LGTM.

@majidsajadi Can you sign the CLA from the above link?

majidsajadi:
@kt3k done.

PR #9959, opened 2021-04-01
chore(cli): Add "Deno" to the windows file description.

dsherret:
I opened task manager and noticed the Deno process does not say "Deno" in the description.

image

You can also see Deno is not visible in the screenshots on this previous PR in another scenario: https://github.com/denoland/deno/pull/6881


cli/Cargo.toml
```diff
@@ -101,4 +101,4 @@ nix = "0.20.0"
 OriginalFilename = "deno.exe"
 LegalCopyright = "© Deno contributors & Deno Land Inc. MIT licensed."
 ProductName = "Deno"
-FileDescription = "A secure runtime for JavaScript and TypeScript."
+FileDescription = "Deno: A secure runtime for JavaScript and TypeScript"
```
dsherret, line 104: Removed period because nothing else in my task manager has a period.

Also, used a colon because it's similar to what node.js does: 
image

dsherret, line 104: FWIW, main website says "A secure runtime for JavaScript and TypeScript." while the repo description says "A secure JavaScript and TypeScript runtime."

kitsonk (approved):
LGTM

PR #9960, opened 2021-04-01
feat(cli): upgrade to TypeScript 4.3

kitsonk:
The PR updates to TypeScript 4.3.  This initial commit is the 4.3-beta for early visibility and testing.

TypeScript 4.3 planned release date is 25th May.

There is new TypeScript syntax in 4.3 that will require updates to swc to fully support (see: https://github.com/swc-project/swc/issues/1517).


kitsonk:
swc has been updated while this was pending and should be compatible.

extensions/fetch/21_formdata.js
```diff
@@ -332,37 +330,34 @@
...
       return this.#writeHeaders(headers);
-    };
-
-    /**
+    } /**
      * @param {string} field
      * @returns {void}
      */
-    #writeFieldHeaders = (field) => {
+
```
bartlomieju, line 337: Are these changes intentional? Looks like comment formatting breaks

bartlomieju (approved):

extensions/fetch/21_formdata.js
kitsonk: hmm sort of `deno fmt` strangeness

PR #9961, opened 2021-04-01
revert: "fix(op_crates/web): align encoding type definitions to spec"

kitsonk:
Reverts denoland/deno#9782

This breaks `std` and could have further breaking changes with downstream users... need to reland later in 1.9 in a non-breaking way.


lucacasonato (approved):
Thx.

I think we can keep https://github.com/denoland/deno_std/pull/834. It should make it simpler to correcting this in a breaking change for 2.0.

caspervonb:
Hmm did not see that coming, lgtm.

PR #9963, opened 2021-04-02
Release v1.8.3

kitsonk:


lucacasonato:
LGTM. Can you merge this into main and then cherry pick it down into 1.8? We need to do two way cherry picking otherwise.

lucacasonato (approved):
LGTM

kitsonk:
Well, the bump PR is only in 1.8, so it is already two qay cherry picking. 🤷

lucacasonato:
Oh ok. I guess let's try that then

ry (approved):

PR #9965, opened 2021-04-02
docs(workers): Fix permissions examples

metamas:
Fix a couple mistyped brackets (`[]` where `{}` is meant), and specify `"none"` arg instead of `false`.

### DIFF
--- a/docs/runtime/workers.md
+++ b/docs/runtime/workers.md
@@ -143,7 +143,7 @@ the `deno.permissions` option in the worker API.
     type: "module",
     deno: {
       namespace: true,
-      permissions: [
+      permissions: {
         net: [
           "https://deno.land/",
         ],
@@ -152,7 +152,7 @@ the `deno.permissions` option in the worker API.
           new URL("./file_2.txt", import.meta.url),
         ],
         write: false,
-      ],
+      },
     },
   });
   ```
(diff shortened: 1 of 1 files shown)


metamas:
Looks like some markdown linting failed on the file I edited. I changed minimal characters, and the failed check doesn't give me much to go on. I guessed at some alternative formatting, but no dice. Please advise me on any formatting issue.

littledivy:
@metamas https://github.com/denoland/deno/blob/main/docs/contributing/development_tools.md#lint-and-format 
```shell
deno run -A --unstable ./tools/format.js
```

metamas:
@littledivy I did try running the format and lint tools. Same error for both. Looks like the program fails to find something it needs. Perhaps I skipped a necessary step for submitting PRs. Didn't think so, since they're minor markdown changes. I'll make some time later to look into it further.
```
error: Uncaught NotFound: No such file or directory (os error 2)
  const p = Deno.run({
                 ^
    at handleError (deno:core/core.js:186:12)
    at jsonOpParseResult (deno:core/core.js:356:9)
    at opSync (deno:core/core.js:259:32)
    at Object.jsonOpSync (deno:core/core.js:372:12)
    at opRun (deno:runtime/js/40_process.js:20:17)
    at Object.run (deno:runtime/js/40_process.js:104:17)
    at dprint (file:///Users/Metamas/Code/deno/tools/format.js:9:18)
    at main (file:///Users/Metamas/Code/deno/tools/format.js:40:9)
    at async file:///Users/Metamas/Code/deno/tools/format.js:64:1
```

metamas:
@littledivy All good now. `deno fmt ./docs/runtime/workers.md` was necessary.

docs/runtime/workers.md
```diff
@@ -233,8 +236,8 @@ the `deno.permissions` option in the worker API.
   });
   ```
 
-- You can disable the permissions of the worker all together by passing false to
-  the `deno.permissions` option.
+- You can disable the permissions of the worker all together by passing `"none"`
+  to the `deno.permissions` option.
```
bartlomieju, line 240: I don't think this is right - shouldn't this be `false`? CC @crowlKats @Soremwar

crowlKats: @bartlomieju it seems that individual permissions take a boolean whereas the actual `permissions` property takes a "none" string, alongside a "inherit" string. ideally it would be a boolean, but what should `true` then be? allow all perms? seems kinda weird

bartlomieju: Ooops, you're right, thanks for clarifying crowlKats

bartlomieju (approved):
LGTM, thanks @metamas

PR #9975, opened 2021-04-02
perf(serde_v8/de): implement SeqAccess size_hint

AaronO:
So serde can optimize array allocation. Array deserialization on serde_v8's benches are 33% faster as a result.

## Benches

```
Before:
test de_array_json        ... bench:         472 ns/iter (+/- 17)
test de_array_v8          ... bench:         578 ns/iter (+/- 15)

After:
test de_array_json        ... bench:         469 ns/iter (+/- 12)
test de_array_v8          ... bench:         386 ns/iter (+/- 8)
```


ry (approved):
LGTM

PR #9976, opened 2021-04-03
docs: list globs used by `deno test`

yw662:
Add 3 lines in testing.md in doc, describing which files will be runned when `deno test` is called.


bartlomieju (approved):
LGTM, thanks @yw662

Issue #9664, opened 2021-03-03
Strange interaction on debug builds + Mac + WebGPU + ICU data


kitsonk:
On my Mac (x86_64-apple-darwin) at commit 7cd14f97c or later on a debug build, I get segfaults from `deno`.  Release builds do not exhibit the behaviour, but it does mean that I can't run tests locally.  I can most concisely recreate it using the REPL:

```
$ deno
Deno 1.7.4
exit using ctrl+d or close()
> Intl.Collator("en-us")
zsh: segmentation fault  deno
```

I tried changing the alignment of the ICU data in core to `Align32` and `Align64` but it did not change the outcome.

bnoordhuis:
I'm pretty sure this is a libc++ symbol conflict caused by webgpu's dependency on same:
<details>

```diff
Process 51292 launched: '/Users/bnoordhuis/src/deno/target/debug/deno' (x86_64)
Deno 1.8.1
exit using ctrl+d or close()
> Intl.Collator("en-us")
Process 51292 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x00007fff75c657a8 libsystem_platform.dylib`_platform_memcmp + 104
libsystem_platform.dylib`_platform_memcmp:
->  0x7fff75c657a8 <+104>: movdqa (%rdi,%rcx), %xmm0
    0x7fff75c657ad <+109>: movdqu (%rsi,%rcx), %xmm1
    0x7fff75c657b2 <+114>: pcmpeqb %xmm1, %xmm0
    0x7fff75c657b6 <+118>: pmovmskb %xmm0, %eax
Target 0: (deno) stopped.
(lldb) bt
error: need to add support for DW_TAG_base_type '()' encoded with DW_ATE = 0x7, bit_size = 0
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x00007fff75c657a8 libsystem_platform.dylib`_platform_memcmp + 104
    frame #1: 0x0000000102a92a7c deno`std::__1::char_traits<char>::compare(__s1="", __s2="", __n=4548070592) at __string:249:12
[...]

PR #9981, opened 2021-04-03
feat(lsp): Implement textDocument/documentSymbol

jeanp413:
I've implemented the case when the response is a hierarchy of symbols `DocumentSymbol[]` (same as [`TypeScriptDocumentSymbolProvider`](https://github.com/microsoft/vscode/blob/94c9ea46838a9a619aeafb7e8afd1170c967bb55/extensions/typescript-language-features/src/languageFeatures/documentSymbol.ts#L43-L61) in vscode).
Let me know if I should also implement the case when the response is a flat list of all symbols `SymbolInformation[]`.

Related #8643


cli/lsp/tsc.rs
```diff
@@ -585,6 +614,86 @@ impl NavigationTree {
...
+
+        document_symbols.push(lsp::DocumentSymbol {
+          name: self.text.clone(),
+          kind: self.kind.clone().into(),
+          range: span.to_range(line_index),
+          selection_range: selection_span.to_range(line_index),
+          tags,
+          children,
+          detail: None,
+          deprecated: None,
```
jeanp413, line 681: Need some help here, to construct a `lsp::DocumentSymbol` I need to specify all fields but the `deprecated` field is marked as deprecated (`#[deprecated(note = "Use tags instead")]`) so I needed to add a `#![allow(deprecated)]` at the top of the file in order to successfully compile. Not sure if that's ok or is there another way?.

jeanp413:
Not sure what `if` condition this linting error is referring to, any ideas?
```
error: in an `if` condition, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`
   --> cli/lsp/tsc.rs:640:29
    |
640 |             .map(|child_span| {
    |  _____________________________^
641 | |             TextRange::at(child_span.start.into(), child_span.length.into())
642 | |           })
    | |___________^
    |
```

cli/lsp/capabilities.rs
```diff
@@ -97,7 +97,7 @@ pub fn server_capabilities(
     )),
     references_provider: Some(OneOf::Left(true)),
     document_highlight_provider: Some(OneOf::Left(true)),
-    document_symbol_provider: None,
+    document_symbol_provider: Some(OneOf::Left(true)),
```
kitsonk, line 100: We should set a label here, but the lsp-types crate doesn't contain it at the moment.  I opened PR https://github.com/gluon-lang/lsp-types/pull/207 to add it, but obviously it will take a way to work through, so we should update this when the dependencies get updated.  Maybe add a TODO for now (or create a tracking issue)?

cli/lsp/tsc.rs
```diff
@@ -585,6 +614,86 @@ impl NavigationTree {
...
+        })
```
kitsonk, line 682:
```suggestion
        document_symbols.push(lsp::DocumentSymbol {
          name: self.text.clone(),
          kind: self.kind.clone().into(),
          range: span.to_range(line_index),
          selection_range: selection_span.to_range(line_index),
          tags,
          children,
          ..Default::default()
        })
```

I believe this should allow you to remove the `#[allow(deprecated)]`.

kitsonk (changes requested):
Looks good... let's try to remove the allow deprecated though, I think I have a solution.

cli/lsp/capabilities.rs
jeanp413: Added a TODO comment

cli/lsp/tsc.rs
jeanp413: This won't work as `lsp::DocumentSymbol` does not implement the `Default` trait

kitsonk: I narrowed the scope of the allow to just the bock where we need it.  It should help avoid unintentional deprecated usages elsewhere.

kitsonk (approved):
LGTM

kitsonk:
I am getting LSP panics (though I can't see what they are) with this commit.  I am trying to narrow it down, but if I can't easily find it, we will need to revert this PR.

kitsonk:
Actually I think it is #9664, as I was using a debug build and when I created a test case to reproduce it, I was getting:

```
process didn't exit successfully: `/Users/kitsonk/github/deno/target/debug/deps/deno-73d82b9ef03210f0 'lsp::language_server::tests::test_document_symbol_with_imports' --exact --nocapture` (signal: 11, SIGSEGV: invalid memory reference)
```

I did a release build and it went away.

jeanp413:
> Actually I think it is #9664,
> I did a release build and it went away.

In my case I'm on ubuntu 20.10 and haven't encounter any error so far, either way, let me know if something goes wrong

kitsonk:
> In my case I'm on ubuntu 20.10 and haven't encounter any error so far, either way, let me know if something goes wrong

It is something that only affects certain Intel debug builds when you try to access ICU data within JavaScript. `tsc` has a few code paths that uses the ICU data (the `Intl` APIs use it) and when you hit one of those it blows up.  It has to do with duplicate symbols introduced with WebGPU.  It is very obscure.  Nothing is wrong with the code.

PR #9984, opened 2021-04-03
fix(core/runtime): Properly await already evaluating dynamic imports

### LINKED ISSUE #9982: Repeat dynamic import doesn't wait for complete module evaluation
`temp.ts`:
```ts
import("./temp2.ts").then(() => console.log(3));
import("./temp2.ts").then(() => console.log(4));
```
`temp2.ts`:
```ts
console.log(1);
await new Promise(r => setTimeout(r, 200));
console.log(2);
```
Actual output: `1 4 3 2`.
Expected: `1 2 3 4`.

The second dynamic import thinks `temp2.ts` is evaluated as soon as the TLA is hit and resolves immediately. Note that it _is_ `Evaluated` at that point under V8's terminology (but of course that's not how a dynamically imported TLA should work). So the offending code should be: https://github.com/denoland/deno/blob/d2fbbfbbf3cab6ddce4b55200dd4bdfa0ca66a1f/core/runtime.rs#L787-L792

~~Very likely related: #9929.~~

nayeemrmn:
Fixes #9982.


ry (approved):
LGTM - thanks Nayeem!

PR #9986, opened 2021-04-03
core: rename heapStats to memoryUsage

### LINKED ISSUE #9980: Deno equivalent of nodejs `process.memoryUsage()`?
https://nodejs.org/api/process.html#process_process_memoryusage

`process.memoryUsage()`

Returns: `<Object>`
```
rss <integer>
heapTotal <integer>
heapUsed <integer>
external <integer>
arrayBuffers <integer>
```

Use case: I am converting backup software written in `nodejs` to `deno` that uses `process.memoryUsage()` to report peak memory usage during the backup, and it would be a useful feature to keep.

AaronO:
This PR aligns the name and output fields with node, since that's "standard" and there's nothing wrong with it.

The `arrayBuffers` field is `0`, since we don't track ArrayBuffer allocation individually, but could with a custom allocator, but I think that's a separate issue.

Partially fixes #9980

## Sample output

```
❯ ./target/debug/deno eval "console.log(Deno.core.memoryUsage())"
{ rss: 3363056, heapTotal: 4243456, heapUsed: 3561772, external: 0, arrayBuffers: 0 }
```

## Questions

1. Should we expose this directly on `Deno` and move it to stable ?


core/bindings.rs
```diff
@@ -865,50 +865,36 @@ fn throw_type_error(scope: &mut v8::HandleScope, message: impl AsRef<str>) {
...
-    number_of_native_contexts: s.number_of_native_contexts(),
-    number_of_detached_contexts: s.number_of_detached_contexts(),
+  MemoryUsage {
+    rss: s.total_physical_size(),
+    heap_total: s.total_heap_size(),
+    heap_used: s.used_heap_size(),
+    external: s.external_memory(),
+    // TODO: track ArrayBuffers, would require using a custom allocator to track
+    // but it's otherwise a subset of external so can be indirectly tracked
+    array_buffers: 0,
```
ry, line 898: Let's just leave it out of the struct then?

ry (commented):
Why not just Deno.memoryUsage() ?

core/bindings.rs
AaronO: I'm fine leaving it out, I put it in so that we're field "compatible" with `node` and that it might make sense to track our ArrayBuffer allocations down the line.

AaronO:
> Why not just Deno.memoryUsage() ?

Will do, if we're ok landing this as stable. Will add types and re-export it as `Deno.memoryUsage()` similar to how we re-export `Deno.core.close()`

core/bindings.rs
satyarohith: +1 for leaving it out. It's confusing (and not useful) for users when the value is `0`. We can re-introduce the field when the allocator is ready.

AaronO: It's now commented out (preserved within the TODO comment) and I've removed it from the TS types.

AaronO:
@ry I've removed the `arrayBuffers` field and re-exported it as `Deno.memoryUsage()`, so this should be good to merge.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2311,4 +2311,13 @@ declare namespace Deno {
...
   ): Promise<void>;
+
+  export interface MemoryUsage {
+    rss: number;
+    heapTotal: number;
+    heapUsed: number;
+    external: number;
+  }
+
+  export function memoryUsage(): MemoryUsage;
```
bartlomieju, line 2322: I think these should go to `lib.deno_unstable.ns.d.ts` for now; I'd rather not add it as a stable API already

AaronO: I've moved the types to unstable

bartlomieju (approved):
LGTM

PR #9991, opened 2021-04-03
docs(getting_started/first_steps): Mention top-level await

toraritte:
Started getting into JS again after many years, so read through the linked MDN resource, but was baffled that `await` is being used on it's own. Assumed that it is implicitly wrapped in an `async` function by Deno's runtime, but [this Stackoverflow thread](https://stackoverflow.com/questions/61816979/await-without-async-in-deno-framwork) confirmed it.

The wording is only a suggestion, but something along those lines would've helped. Thanks!

### DIFF
--- a/docs/getting_started/first_steps.md
+++ b/docs/getting_started/first_steps.md
@@ -8,6 +8,8 @@ you might want to follow a guide
 [on the basics of JavaScript](https://developer.mozilla.org/en-US/docs/Learn/JavaScript)
 before attempting to start with Deno.
 
+**NOTE**: Deno supports the [`top-level await` ECMAScript proposal](https://github.com/tc39/proposal-top-level-await) which "_enables modules to act as big async functions_", allowing `await` to be used outside an `async` function at the top level.
+
 ### Hello World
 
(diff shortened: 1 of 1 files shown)


Soremwar:
Not a mantainer, but I believe a better clarification would be to say that Deno all of the latest JavaScript features released on ECMA 2021 and below

toraritte:
Again, I stopped working with JS before promises came out so...

Soremwar:
You have a lot of reading to do then. A lot of features as interesting as top level await have been released since, and they are all available in Deno :)

nayeemrmn:
I agree with @Soremwar, this is like any ES standard that we pretty much have to adopt. For example, Node also supports https://github.com/tc39/proposal-top-level-await but it isn't explicitly written in such a visible place.

kitsonk:
Correct, we don't need to cover off every feature supported by the ECMAScript standard in the documentation.

If anything we need to help ensure things like https://github.com/kangax/compat-table include Deno and point people to that, but that wouldn't be this PR.  I will open another issue for tracking that.

Respectfully closing.

PR #9993, opened 2021-04-04
perf(runtime): optimize PermissionState::check

AaronO:
By converting `log_perm_access()` to a macro, so the message sub-expression is dropped at compile time in release mode. Otherwise it made allocations and added substantial overhead to opcalls.

## Benches

Measured on `perf_now` (`op_now`) from `deno_common`:

```
Before:
perf_now:                n = 5000000, dt = 2.429s, r = 2058460/s, t = 485ns/op

After:
perf_now:                n = 5000000, dt = 1.773s, r = 2820079/s, t = 354ns/op
```

### Before:
image

### After

image

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -34,6 +34,37 @@ pub enum PermissionState {
 }
 
 impl PermissionState {
+  #[inline(always)]
+  fn log_perm_access(name: &str, info: Option<&str>) {
+    debug!(
+      "{}",
+      colors::bold(&format!(
+        "{}️  Granted {}",
+        PERMISSION_EMOJI,
+        Self::fmt_access(name, info)
+      ))
+    );
+  }
+
+  fn fmt_access(name: &str, info: Option<&str>) -> String {
+    format!(
+      "{} access{}",
+      name,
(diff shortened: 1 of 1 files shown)


ry (commented):
Cool - seems simple enough. 

Out of curiosity wouldn't `#[inline]` also do the trick?

AaronO:
> Cool - seems simple enough.
> 
> Out of curiosity wouldn't `#[inline]` also do the trick?

I've refactored this so it no longer uses a macro, it instead inlines it and calls `Self::fmt_access` inside `debug!()`, which will cause a double alloc of the `access` msg in the prompt path, but only for debug builds which is fine IMO.

AaronO:
Both the flamegraph and the `deno_common` bench confirm this is optimized:

```
Before (main):
perf_now:            	n = 500000, dt = 0.155s, r = 3225806/s, t = 310ns/op

After (this PR):
perf_now:            	n = 5000000, dt = 0.821s, r = 6090134/s, t = 164ns/op
```

bartlomieju (approved):
LGTM

PR #9999, opened 2021-04-04
docs(architecture): update Deno.metrics() table data

littledivy:
Updates fields for `Deno.metrics()` example in the docs.


ry (approved):
👍

PR #10009, opened 2021-04-04
refactor(ops): first pass at op best practices

AaronO:
This PR is a follow-up to #9843 and tweaks most ops across the codebase to follow our best practices for ops.

It's also important to move away from `serde_json::Value` to fully realize the perf gains of `serde_v8` (since it otherwise does a double serializaiton and it's serialized as a loosely typed map, which is very inefficient, doesn't benefit from `v8_struct_key` optims, etc...)

## Goals

1. More consistent op functions
2. More performant ops (removing overhead of `json!()` and `serde_json::Value` which are loosely typed)
3. Better typed op code (easier to maintain, catch bugs, etc...)

## Changes related to best-practices

1. Remove wrappers for single args (e.g: `{ rid }`)
2. Remove single-keyed structs to deserialize said wrappers
3. Specific return values types instead of general `Result<Value, AnyError>`
4. Empty returns: `Ok(json!({}))` => `Ok(())`
5. Use rust's unit type for ignored/unused args
6. Remove most `serde_json::*` imports and sometimes `serde::Deserialize` (after dropping wrapper structs)
7. Remove pseudo-variadic ZeroCopyBufs in JS

## Steps

- [x] `runtime/`
  - [x] Bulk of modules, with relatively simple changes
  - [ ] Modules like `net` that require introducing new general types
- [x] `op_crates`

## Notes

- Yes, this touches a lot of the code
- This should mostly change rust code, but I've touched JS code to match up the arg-unwrapping
- Some places that are loosely typed or that return HashMaps still use `serde_json::Value`
- This also applies to places where `json!()` offered some "useful" implicit conversions for references (e.g: `&str`, `Cow`, ...) to owned types (that can be returned and serialized), to my understanding there was no (performance) upside to returning refs, since `json!()` would convert them to owned types anyway

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -185,7 +185,7 @@ declare namespace Deno {
      * console.log(Deno.env.get("MADE_UP_VAR"));  // outputs "Undefined"
      * ```
      * Requires `allow-env` permission. */
-    get(key: string): string | undefined;
+    get(key: string): string | null;
 
     /** Set the value of an environment variable.
      *
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -185,7 +185,7 @@ declare namespace Deno {
      * console.log(Deno.env.get("MADE_UP_VAR"));  // outputs "Undefined"
      * ```
      * Requires `allow-env` permission. */
-    get(key: string): string | null;
+    get(key: string): string | undefined;
 
     /** Set the value of an environment variable.
      *
--- a/op_crates/url/lib.rs
+++ b/op_crates/url/lib.rs
@@ -4,8 +4,6 @@ use deno_core::error::generic_error;
 use deno_core::error::type_error;
 use deno_core::error::uri_error;
 use deno_core::error::AnyError;
-use deno_core::serde_json::json;
-use deno_core::serde_json::Value;
 use deno_core::url::form_urlencoded;
 use deno_core::url::quirks;
 use deno_core::url::Url;
@@ -34,13 +32,28 @@ pub struct UrlParseArgs {
   set_username: Option<String>,
 }
 
+#[derive(Serialize)]
+pub struct UrlParts {
+  href: String,
+  hash: String,
+  host: String,
+  hostname: String,
+  origin: String,
+  password: String,
+  pathname: String,
+  port: String,
+  protocol: String,
+  search: String,
+  username: String,
+}
+
 /// Parse `UrlParseArgs::href` with an optional `UrlParseArgs::base_href`, or an
 /// optional part to "set" after parsing. Return `UrlParts`.
 pub fn op_url_parse(
   _state: &mut deno_core::OpState,
   args: UrlParseArgs,
   _zero_copy: Option<ZeroCopyBuf>,
-) -> Result<Value, AnyError> {
+) -> Result<UrlParts, AnyError> {
   let base_url = args
     .base_href
     .as_ref()
@@ -75,20 +88,6 @@ pub fn op_url_parse(
       .map_err(|_| uri_error("Invalid username"))?;
   }
 
-  #[derive(Serialize)]
-  struct UrlParts<'a> {
-    href: &'a str,
-    hash: &'a str,
-    host: &'a str,
-    hostname: &'a str,
-    origin: &'a str,
-    password: &'a str,
-    pathname: &'a str,
-    port: &'a str,
-    protocol: &'a str,
-    search: &'a str,
-    username: &'a str,
-  }
   // TODO(nayeemrmn): Panic that occurs in rust-url for the `non-spec:`
   // url-constructor wpt tests: https://github.com/servo/rust-url/issues/670.
   let username = catch_unwind(|| quirks::username(&url)).map_err(|_| {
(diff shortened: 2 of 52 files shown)


AaronO:
This isn't exhaustive and doesn't fix all non-best practices, but fixes most of them.

Most of the remaining uses of `serde_json::Value` or `json!()` in ops are either loosely-typed returns or more involved returns  (e.g: `runtime/net`) that should be handled in a follow-up.

AaronO:
`op_crates` is now `serde_json` free, except `websocket`.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -185,7 +185,7 @@ declare namespace Deno {
      * console.log(Deno.env.get("MADE_UP_VAR"));  // outputs "Undefined"
      * ```
      * Requires `allow-env` permission. */
-    get(key: string): string | undefined;
+    get(key: string): string | null;
```
bartlomieju, line 188: I'm not sure we want to change declaration files - this could potentially be a breaking change. @kitsonk WDYT?

bartlomieju (commented):
The PR looks good to me, and if tests are passing I think we're good. Before landing I'll let other take a look

cli/dts/lib.deno.ns.d.ts
lucacasonato: I agree - this has to be reverted. This is a public API change to a stable API.

op_crates/webgpu/lib.rs
```diff
@@ -113,73 +112,75 @@ pub fn get_declaration() -> PathBuf {
...
 }
 
-fn deserialize_features(features: &wgpu_types::Features) -> Vec<&str> {
-  let mut return_features: Vec<&str> = vec![];
+fn deserialize_features(features: &wgpu_types::Features) -> Vec<String> {
+  let mut return_features: Vec<String> = vec![];
 
   if features.contains(wgpu_types::Features::DEPTH_CLAMPING) {
-    return_features.push("depth-clamping");
+    return_features.push("depth-clamping".to_string());
```
lucacasonato, line 119: Again, a bunch of unnecessary allocations. Can serde_v8 not handle `&str`.

op_crates/url/lib.rs
```diff
@@ -102,41 +101,42 @@ pub fn op_url_parse(
...
-    origin: &quirks::origin(&url),
-    password: quirks::password(&url),
-    pathname: quirks::pathname(&url),
-    port: quirks::port(&url),
-    protocol: quirks::protocol(&url),
-    search: quirks::search(&url),
-    username,
-  }))
+  Ok(UrlParts {
+    href: quirks::href(&url).to_string(),
```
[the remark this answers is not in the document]
AaronO: These allocation were there before, they were just implicit, see my notes:
> This also applies to places where `json!()` offered some "useful" implicit conversions for references (e.g: &str, Cow, ...) to owned types (that can be returned and serialized), to my understanding there was no (performance) upside to returning refs, since json!() would convert them to owned types anyway

You'll see that `serde_json::Value` only has owned `String`s:
https://sourcegraph.com/github.com/serde-rs/json@d735585370d3d62502130051529410464e147224/-/blob/src/value/mod.rs#L111-171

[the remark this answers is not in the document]
lucacasonato: Ah... very painful. Good to know.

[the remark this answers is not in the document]
AaronO: Just to elaborate on specifics:
- string refs get serialized [here in the json macro](https://sourcegraph.com/github.com/serde-rs/json@d735585370d3d62502130051529410464e147224/-/blob/src/macros.rs#L275-280)
- Which calls to `to_value()`
- Which ultimately calls this [serialize_str()](https://sourcegraph.com/github.com/serde-rs/json@d735585370d3d62502130051529410464e147224/-/blob/src/macros.rs#L275-280) method of the serializer
- Which is basically `Ok(Value::String(value.to_owned()))`

op_crates/webgpu/lib.rs
AaronO: I converted this to use `&'static str` (which can be safely returned then serialized), but in practice this should have no performance impact given that this setup method should be called very infrequently.

cli/dts/lib.deno.ns.d.ts
AaronO: Beyond the typing conversation, which is important, we should also reach a consensus on what should be serialized as `null` or `undefined` by `serde_v8`. Right now `serde_v8` serializes both `None` and `()` to `null`, which `serde_json::Value` also did, since `undefined` is not a JSON value, only `null` exists in JSON.

To return `undefined` from these ops we were doing hackish (IMO) workarounds, like returning a 0-or-1 sized arrays in the rust-op and then `[0]` in JS to the value or `undefined`.

With `serde_v8` we have the opportunity to change this behaviour and provide a mechanism for returning `undefined`, by mapping to it from `None` or `()`

I personally would keep everything as `null`, but I just want you to know that the option to change this behaviour is there and it would be best to decide on it before shipping `1.9`.

hayd: you could keep the ops with null and do the null->undefined in js
(then this could potentially be changed/standardized all to nulls in 2.0)

AaronO: > you could keep the ops with null and do the null->undefined in js
> (then this could potentially be changed/standardized all to nulls in 2.0)

Yep, that's precisely what I suggested over in the discord chat about this (in `#dev-core`)

AaronO: These changes have been reverted and I've mapped `null`s to `undefined`.

bartlomieju (approved):
LGTM

PR #10011, opened 2021-04-04
Add project scope to top README, fixes 10007

fulldecent:


ry:
Thanks but I think this is more confusing than the status quo. The manual is easy to find.

PR #10012, opened 2021-04-04
feat(cli/dts): stricter typings for Listener & Conn

### LINKED ISSUE #5477: Deno.Listener no exposing port
To extract the port from a random port in a Deno.listen is somewhat painful:
```
    const listener = Deno.listen({ port: 0, transport: "tcp" });
    //@ts-ignore
    const {port} = listener.addr
    this.port = port
```

While the random port doesn't seem that useful, it is extremely. For unit tests,  launching a listener or process at a random port means that unit tests can really be independent, and not hardcoded.

crowlKats:
Closes #5477

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1713,7 +1713,7 @@ declare namespace Deno {
   export type Addr = NetAddr | UnixAddr;
 
   /** A generic network listener for stream-oriented protocols. */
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM. This should not be breaking due to the type argument having a default type.

@kitsonk PTAL too

kitsonk (approved):
LGTM too.

PR #10016, opened 2021-04-05
docs: Added documentation about Atom Editor

ayame113:
I made an extension that uses deno's language server in the Atom editor.
https://github.com/ayame113/atom-ide-deno

I would like to add documentation on this.


kitsonk (approved):
LGTM

Issue #9795, opened 2021-03-15
Remove pure JS (non native) io functions from Deno namespace


lucacasonato:
I propose we remove the following APIs from the Deno namespace, and move them to std:

Types:
- [ ] `Deno.Reader`
- [ ] `Deno.Writer`
- [ ] `Deno.ReaderSync`
- [ ] `Deno.WriterSync`
- [ ] `Deno.Closer`

Functions:
- [x] `Deno.Buffer` (#9793)
- [x] `Deno.copy` (#11369)
- [x] `Deno.iter` (#10025)
- [x] `Deno.iterSync` (#10025)
- [x] `Deno.readAll` (#9793)
- [x] `Deno.readAllSync` (#9793)
- [x] `Deno.writeAll` (#9793)
- [x] `Deno.writeAllSync` (#9793)

### Timeline

I suggest we mark all the APIs we want to remove with `@deprecated`, and add a `deno lint` diagnostic for when these APIs are used. In Deno 2.0 we can then remove these APIs.

### Reasoning

Code which does not actually require native syscalls should not use `Deno` namespace APIs as this makes the code incompatible with browsers, even though there is no need for that to be the case. An example of this: https://deno.land/std@0.90.0/io/streams.ts. This uses `Deno` APIs, even though no native calls are preformed, making this code unusable in a web browser for no good reason.

kitsonk:
Why move the types? They are referred to by other built ins that are returned, are they not?

kitsonk:
Buffer is also mentioned in #9588, we should maybe close that?

lucacasonato:
> Why move the types?

Same reason as to move the functions. Code using `Reader` / `Writer` etc does not work when `lib` is `dom` instead of `deno.ns`. IMO if it doesnt need to be a binding (or a type for those bindings) because it can work just fine in the browser, it should not be part of the `Deno` namespace.

> They are referred to by other built ins that are returned, are they not?

The builtins never explicitly return or take a reader/writer after these changes as far as i'm aware. They only return types that _implement_ `Reader` or `Writer`. Instead of using `implement` in the d.ts, we can just specify the `read` / `write` methods no the interfaces directly. This would make them implicitly implement `Reader` / `Writer` (just like interfaces work in Go).

timreichen:
@lucacasonato in addition to that: would you be open to rename
`Deno.iter` -> `Deno.iterator`
`Deno.iterSync` -> `Deno.iteratorSync`
[...]

PR #10025, opened 2021-04-05
chore: deprecate Deno.iter and Deno.iterSync

lucacasonato:
This commit marks the `Deno.iter` and `Deno.iterSync` utils as
deprecated, and schedules them for removal in Deno 2.0. These
utilities are implemented in pure JS, so should not be part of the
Deno namespace.

These utilities are now available in std/io/util:
denoland/deno_std#843.

Towards #9795


ry (approved):
LGTM

As we get closer to 2.0, we might want to turn on "hard deprecation" mode where we actually print a warning message if you use these APIs.

PR #10026, opened 2021-04-05
core: remove serde_json-isms in op_close() and op_resources()

AaronO:
Core no longer uses `serde_json` now, besides re-exporting it or in the module specifier tests.

This is a follow-up to #10009 and #9843


ry (approved):
LGTM

PR #10033, opened 2021-04-06
fix(lsp): properly handle encoding URLs from lsp client

### LINKED ISSUE #9741: Goto Type Definition broken for cdn.skypack.dev
**Describe the bug**

The previous (canary) extension supported dependencies from cdn.snowpack.dev,
but the latest (3.1.0) extension has regressed.

**To Reproduce**

1. `import postcss from "https://cdn.skypack.dev/postcss?dts";`
2. Download the dependencies via "Quick Fix" menu
3. Right-click the import and click "Go to definition"

**Expected behaviour**

The TypeScript definition should be displayed, but instead the following error is shown:

```
Unable to open 'postcss.d.ts': Unable to resolve text model content for resource deno:/https/cdn.skypack.dev/-/postcss%40v8.2.7-bLFqaTc7OBC3TUTNEPjL/dist%3Des2020%2Cmode%3Dtypes/lib/postcss.d.ts.
```

**Investigation**

The definition is definitely cached correctly in `~/Library/Caches/deno/deps/https/cdn.skypack.dev/`, also hovering the definition for the module indicates the URL is correct: `module "https://cdn.skypack.dev/-/postcss@v8.2.7-bLFqaTc7OBC3TUTNEPjL/dist=es2020,mode=types/lib/postcss"`.

I suspect that this might be as simple as running `"deno:/https/cdn.skypack.dev/-/postcss%40v8.2.7-bLFqaTc7OBC3TUTNEPjL/dist%3Des2020%2Cmode%3Dtypes/lib/postcss.d.ts"` though `decodeURIComponent`.

I had a look at the source, but I didn't find an obvious handler and it looks like the faulty value might come from the LSP endpoint, which might make this a Deno bug.

**Versions**

VSCodium: 1.53.2
deno 1.8.0 (release, aarch64-apple-darwin)
v8 9.0.257.3
typescript 4.2.2

kitsonk:
Fixes #9741

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -2065,7 +2065,7 @@ impl Inner {
           if let Some(source) = self.sources.get_source(&specifier) {
             Some(source)
           } else {
-            error!("The cached sources was not found: {}", specifier);
+            error!("The cached source was not found: {}", specifier);
             None
           }
         }
--- a/cli/lsp/urls.rs
+++ b/cli/lsp/urls.rs
@@ -4,14 +4,28 @@ use crate::file_fetcher::map_content_type;
 use crate::media_type::MediaType;
 
 use deno_core::error::AnyError;
+use deno_core::url::Position;
 use deno_core::url::Url;
 use deno_core::ModuleSpecifier;
 use std::collections::HashMap;
 
-/// This is a partial guess as how URLs get encoded from the LSP.  We want to
-/// encode URLs sent to the LSP in the same way that they would be encoded back
-/// so that we have the same return encoding.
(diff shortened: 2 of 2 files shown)


lucacasonato (approved):
LGTM. It's kind of worrying that we are binding so close to JS because VSCode is the dominant editor.

kitsonk:
> It's kind of worrying that we are binding so close to JS because VSCode is the dominant editor.

This should work for any editor though, it just widens the situation where the URLs can be "manipulated" by a client, but still have the LSP work properly, versus specifically coupling to VSCode.  This should be opaque to other editors.

PR #10034, opened 2021-04-06
fix(lsp): normalize windows file URLs properly

### LINKED ISSUE #9744: lsp: Extension reports spurious errors on Windows because of path encoding issues
(EDIT: See comments below, this goes beyond import maps, there's a mismatch in how paths are encoded between the extension and the LSP on Windows)

I'm using an import map to remap `.js` imports to `.ts` in code I share between the browser and Deno.  
It is working fine in Deno, but the extension reports the following errors:

    Unable to load a local module: "file:///c%3A/vscode-deno-import-map/browser/secondary.js".
    Please check the file path.

and

    Property 'value' does not exist on type 'typeof import("file:///c%3A/vscode-deno-import-map/browser/shared")'.

**To Reproduce**

See screenshot below or minimal repro here: https://github.com/elisee/vscode-deno-import-map 

**Expected behavior**

The import map should be properly applied by the extension and there should be no errors.

**Screenshots**

image

**Versions**

vscode: v1.53.2
deno: v1.8.0
extension: v3.1.0
os: windows (seems important, see comment below)

kitsonk:
Fixes #9744
Fixes https://github.com/denoland/vscode_deno/issues/386

### DIFF
--- a/cli/lsp/urls.rs
+++ b/cli/lsp/urls.rs
@@ -128,32 +128,19 @@ impl LspUrlMap {
   }
 
   /// Normalize URLs from the client, where "virtual" `deno:///` URLs are
-  /// converted into proper module specifiers, and handle situations where file
-  /// URLs on windows can be canonicalized differently.
+  /// converted into proper module specifiers, as well as handle situations
+  /// where the client encodes a file URL differently than Rust does by default
+  /// causing issues with string matching of URLs.
   pub fn normalize_url(&self, url: &Url) -> ModuleSpecifier {
     if let Some(specifier) = self.get_specifier(url) {
       specifier.clone()
     } else {
-      let mut url = url.clone();
-      if url.scheme() == "file" && url.as_str().contains("%3A") {
-        let path = url
-          .path_segments()
-          .unwrap()
-          .map(|s| {
-            if s.ends_with("%3A") && s.len() == 4 {
-              percent_encoding::percent_decode_str(s)
-                .decode_utf8_lossy()
-                .to_string()
-                .to_ascii_uppercase()
-            } else {
-              s.to_string()
-            }
-          })
-          .collect::<Vec<String>>()
-          .join("/");
-        url.set_path(&path);
+      if url.scheme() == "file" {
+        let path = url.to_file_path().unwrap();
+        Url::from_file_path(path).unwrap()
+      } else {
+        url.clone()
       }
-      url
     }
   }
 }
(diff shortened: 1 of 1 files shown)


kitsonk:
ok, I have test this under windows and it resolves the issue...

cli/lsp/urls.rs
```diff
@@ -103,12 +103,32 @@ impl LspUrlMap {
...
+      let mut url = url.clone();
+      if url.scheme() == "file" && url.as_str().contains("%3A") {
+        let path = url
+          .path_segments()
+          .unwrap()
+          .map(|s| {
+            if s.ends_with("%3A") && s.len() == 4 {
+              percent_encoding::percent_decode_str(s)
+                .decode_utf8_lossy()
+                .to_string()
```
piscisaureus, line 121: This `to_string()` is unnecessary.

cli/lsp/urls.rs
```diff
@@ -103,12 +103,32 @@ impl LspUrlMap {
...
       specifier.clone()
     } else {
-      url.clone()
+      let mut url = url.clone();
+      if url.scheme() == "file" && url.as_str().contains("%3A") {
+        let path = url
+          .path_segments()
+          .unwrap()
+          .map(|s| {
+            if s.ends_with("%3A") && s.len() == 4 {
```
piscisaureus, line 118: * Suggestion: apply only to the first path segment, and only if the url has no (/empty) hostname.
* I don't quite understand why this is necessary. Does "generic" normalization not apply to `%3A`.
* What about `file:///c%3a/foo` (lowecase a) and `file:///c|/foo` (alternative notation) ?

piscisaureus:
It's not quite clear to me what the purpose of normalizing is here. I would suppose that `file:///` URLs would always get percent-decoded, as they are URLs after all. 

So  if I had written: 

`import "file:///C%2A/foo/bar/baz/file%20with%20%25%20and%20%23%20in%20the%20name.txt";`

It'd import from:

`C:\foo\bar\baz\file with % and # in the name.txt`

cli/lsp/urls.rs
kitsonk: 👍

kitsonk:
@piscisaureus thanks for the feedback.  I have taken a different approach, which should be far more foolproof.  File URLs are transformed into file paths and then back into file URLs.  This ensures that the URL will be normalized in a way that matches Rust's behaviour.  I tested it again on windows and it fixes both the linked issues, but should be able to be generically applied to any client URL.

piscisaureus:
This looks good! Would be great to have a unit test.

kitsonk:
> This looks good! Would be great to have a unit test.

There are two unit tests, one for a Windows system (which ensures the drive letter get recapitalised) and one for non-windows which deals just with percent encoding of parts of the path.

Did you mean some sort of integration test?  Those are complicated in this case because of the need to write out real files, mangle them, craft the messages and then send them in and deal with the nature of windows versus non-windows.

piscisaureus (approved):
LGTM

piscisaureus:
Sorry, I must have overlooked the tests that were already there.

PR #10035, opened 2021-04-06
fix(console): deprecate Deno.customInspect

### LINKED ISSUE #9294: Deno.customInspect should be replaced with "denoCustomInspect"
```diff
  class Foo {
-   [Deno.customInspect]() {
+   denoCustomInspect() {
      return "foo";
    }
  }
```

This should be a progressively enhanced thing like `WorkerOptions::deno`. It should no-op in browsers, not require verbose feature detection to be compatible.

I think the alias should be made and `Deno.customInspect` should be deprecated for 1.8.0, and the latter should be removed for 2.0.

kt3k:
This PR deprecates `Deno.customInspect` as suggested in #9294 and suggests using `Symbol.for("Deno.customInspect")` instead.

ref #9294


op_crates/console/02_console.js
```diff
@@ -907,22 +907,11 @@
...
-    }
-    // This non-unique symbol is used to support op_crates, ie.
-    // in op_crates/web we don't want to depend on unique "Deno.customInspect"
-    // symbol defined in the public API. Internal only, shouldn't be used
-    // by users.
-    const nonUniqueCustomInspect = Symbol.for("Deno.customInspect");
-    if (
-      nonUniqueCustomInspect in value &&
-      typeof value[nonUniqueCustomInspect] === "function"
-    ) {
```
bartlomieju, line 920: Removing this logic means that `Deno.customInspect` will stop working. I think we should still support it until 2.0

nayeemrmn: The non-unique custom inspect which we use for op crates implementations should be renamed to something else since the public custom inspect symbol has stolen its name. We handle the internal one with its own logic (i.e. pass it the `inspect()` function as an arg) so the difference should be preserved.

lucacasonato: I think we should always pass the `inspect` function as an arg - even for the public one.

kt3k: OK. I'll rename internal one to something else and keep the different handlings of public one and internal one.

kt3k:
Please review in this URL https://github.com/denoland/deno/pull/10035/files?w=1 (my editor trimmed trailing whitespaces in files)

bartlomieju:
@kt3k could you please update [`no-deprecated-deno-api` lint rule](https://lint.deno.land/#no-deprecated-deno-api) (https://github.com/denoland/deno_lint/blob/main/src/rules/no_deprecated_deno_api.rs) after landing this PR?

kt3k:
@bartlomieju Sure. I'll try.

bartlomieju (approved):
Lint rule has already landed in https://github.com/denoland/deno_lint/pull/748 so this is ready to land. Looks good to me

PR #10037, opened 2021-04-06
docs: export EmitOptions and EmitResult from d.ts

kt3k:
EmitOptions and EmitResult are not available in `deno doc` command because these are not exported in d.ts file. With this change we can see the docs by the following commands:

```
deno doc --unstable --builtin Deno.EmitOptions
deno doc --unstable --builtin Deno.EmitResult
```


kitsonk (approved):
LGTM

PR #10044, opened 2021-04-06
refactor(op_crates/webgpu): simplify gfx_select macro & add gfx_put and gfx_ok macros

crowlKats:
Simplifies the gfx_select macro as it has cases we are not concerned about (ios & wasm), and simplifies a lot of duplicate logic surrounding the gfx_select macro


bartlomieju (approved):
LGTM

PR #10042, opened 2021-04-06
chore: update deno_file to use deno_webidl

lucacasonato:
This changes the custom input converters in deno_file to use deno_webidl
converters.

PR #10045, opened 2021-04-06
feat: blob URL support

### LINKED ISSUE #2726: Support for data URLs and object URLs
Data URLs and Object URLs (generally of blobs) [are frequently used to dynamically create Workers](https://stackoverflow.com/a/10372280) -- support for these is missing in Deno which presents a compatibility issue with existing web code, and is an issue when trying to create workers from strings in Deno.

What's missing:
- Support for the `data:` scheme for URLs
- ~~Support for object URLs and `URL.createObjectURL()`~~ (Solved by #2729)

### LINKED ISSUE #9210: Support blob/Object URLs for workers
Deno should support blob/Object URLs for workers.

For example the following should work:

```js
const src = new Blob([`console.log("hello deno");`]);
new Worker(URL.createObjectURL(src), { type: "module" });
```

This is splitting off from #5683 as it far more than allowing the `blob:` protocol.  This is because we do all the processing of modules/worker scripts inside Rust.  This means we expect the specifier to be fully resolvable inside of Rust.  For `http`/`https` URLs we go and fetch the resources in Rust.  For `file` URLs we have a whole string which we can decode to provide us the module text.  With Object URLs, the actual content is still located in JavaScript isolate/sandbox, so it means for these Object URLs, we need to serialise them to a string before passing them out to Rust and then deal with them there.  This is more complex than simply passing out a string.

It also means it won't work for things like `import` and `export`, because these need to be fully statically analysed before they are loaded.  We might be able to get it for dynamic `import()` but it would require doing some gymnastics in rusty_v8, because `import()` is syntax, not a runtime construct.

Because of all of these challenges it goes from "yeah sure, makes sense" to "hmmm... we need to think about it".

lucacasonato:
This commit adds blob URL support. Blob URLs are stored in a process
global storage, that can be accessed from all workers, and the module
loader. Blob URLs can be created using `URL.createObjectURL` and revoked
using `URL.revokeObjectURL`.

This commit does not add support for `fetch`ing blob URLs. This will be
added in a follow up commit.

This PR depends on #10042.

Closes #9210
Closes #2726


cli/source_maps.rs
```diff
@@ -142,7 +142,13 @@ pub fn get_orig_position<G: SourceMapGetter>(
...
               // `source_file_name` does not parse as a URL.
               let file_name = match deno_core::resolve_url(source_file_name) {
-                Ok(m) => m.to_string(),
+                Ok(m) => {
+                  if m.scheme() == "blob" {
+                    file_name
+                  } else {
+                    m.to_string()
+                  }
+                }
```
kitsonk, line 151:
```suggestion
                Ok(m) if m.scheme() == "blob" => file_name,
                Ok(m) => m.to_string(),
```

cli/tsc.rs
```diff
@@ -378,15 +400,24 @@ fn resolve(state: &mut State, args: Value) -> Result<Value, AnyError> {
...
+              specifier_str
+            }
+            "blob" => {
+              let specifier_str =
+                hash_blob_url(&resolved_specifier, &media_type);
+              state
+                .blob_url_map
+                .insert(specifier_str.clone(), resolved_specifier);
+              specifier_str
+            }
```
kitsonk, line 419:
```suggestion
            "data" | "blob" => {
              let specifier_str = if resolve_specifier.scheme() == "data" {
                hash_data_url(&resolved_specifier, &media_type)
              } else {
                hash_blob_url(&resolved_specifier, &media_type);
              };
              state
                .data_url_map
                .insert(specifier_str.clone(), resolved_specifier);
              specifier_str
            }
```

op_crates/file/lib.rs
```diff
@@ -1,7 +1,85 @@
...
+use std::collections::HashMap;
 use std::path::PathBuf;
+use std::sync::Arc;
+use std::sync::Mutex;
+use uuid::Uuid;
+
+#[derive(Clone)]
+pub struct Blob {
+  pub data: Vec<u8>,
+  pub r#type: String,
```
kitsonk, line 18: This is a very confusing property name.  Is there a good reason it is named something so obtuse and with a `#` in the name, unlike any other property on a struct?  While I realise it is named `.type` in the browser, I don't see why we wouldn't name it `media_type` in Rust.

kitsonk (commented):
A couple of thoughts...

Great news though!

op_crates/file/lib.rs
lucacasonato: The `r#` is rust syntax - it isn't actually part of the property name. It is the raw string token and just tells rust to not interpret the following property name as a keyword, instead just interpret it as a raw string (https://doc.rust-lang.org/reference/tokens.html#characters-and-strings). I will update it to `media_type` anyway to simplify :-)

kitsonk (approved):
LGTM, this is going to be awesome.

PR #10050, opened 2021-04-07
feat(cli): Add --worker to cache and bundle

### LINKED ISSUE #8219: deno bundle should support type checking worker scripts
When I try to bundle my worker.ts it raises an error
```
error: TS2339 [ERROR]: Property 'postMessage' does not exist on type 'Window & typeof globalThis'.
        self.postMessage(data);
```
Code
```
self.postMessage(data);
```
In Deno
```
deno 1.5.1
v8 8.7.220.3
typescript 4.0.3
```

nayeemrmn:
Closes #8219.


nayeemrmn:
@ry Please review.

kitsonk:
@nayeemrmn I did some thinking about this.  The problem is that `--worker` is actually quite limiting and not very future proof.  What if we start supporting Service Workers?  Then we add `--service-worker`?

We could use something like `--type worker` but I think we would regret that as well, as `--type` is ambiguous.

After having thought about it, I would like to propose that we introduce a flag `--check` which takes an optional second argument which specifies the flavour of type checking to perform.  It would also be mutually exclusive of `--no-check` and would error if both `--check` and `--no-check` were supplied.  The second parameter would be:

- `main` || `window` - this would be the default implied one as well, this means that `deno run a.ts`, `deno cache --check a.ts`, `deno cache --check main a.ts`, and `deno cache --check window a.ts` would all behave exactly the same.
- `worker` - Type checked against the default worker type library.

It means we can introduce other types in the future as well, and in theory when we get workspace folders in the LSP, an analogue for specifying the type library to use there.

Thoughts?

nayeemrmn:
I suppose I have a different image of the future of flags for type libs, I don't see a problem with having a separate flag for each type lib "modifier". It's important to keep in mind that `--unstable` is one of these: https://github.com/denoland/deno/blob/068f7709ec66a27252de59879c0d932725f6f817/cli/main.rs#L541-L546

Note that `--unstable` and `--worker` aren't mutually exclusive. For example another modifier I'm interested in seeing is `--web` (remember these non-main modes only apply to `deno cache/bundle`) which also wouldn't be mutually exclusive to the others. Each valid combination of these flags would be mapped to some variant of `TypeLib`.

Anyway, I'm averse to the flag value design because of 1) the lack of adjacency with `--unstable` and 2) the non-mutual-exclusivity of the modifiers. If the names are too generic we can prefix them as `--check-worker` or something, but otherwise I don't see a problem with having them as top-level flags.

kitsonk:
`--unstable` is more than a flag that indicates how the code should be type checked... `--unstable` applies even when type checking is skipped as it augments the runtime behaviour.  The purpose of these flags are simply to augment the types being used to validate the code, if type checking is occurring.  Yes, `--unstable` also has an impact what type libraries are to be used, but that impact is easily determined with an expanded match statement as you point out.  Technically `deno cache --no-check` or `deno bundle --no-check` could easily omit the `--unstable` and Deno would do exactly what it is supposed to do, though someone trying to consume that code later would need to know if `--unstable` was required to run it.

To be pointed, I am opposed to a `--worker` flag.  It is limiting, it is ambiguous to what it does, and does not lead nicely to a future where things like ServiceWorkers are easily supported without adding a plethora of flags.

nayeemrmn:
> It is limiting, it is ambiguous to what it does, and does not lead nicely to a future where things like ServiceWorkers are easily supported without adding a plethora of flags.

If the names are ambiguous, as I said we can prefix it to `--check-worker`. I still don't understand why it wouldn't be nice to have a flag for each type lib modifier, `--check-worker` and `--check-service-worker` far from demonstrate that to be honest. They'd all show much clearer on the help text with individual descriptions. Clap lets us specify arbitrarily which ones are mutually exclusive and it would allow us to add more that aren't necessarily mutually exclusive. For example a `--check-web` which also adds in the dom types that aren't implemented in Deno, that interacts differently with `--worker` and without it. With the flag value solution, which is already very complicated with the defaults, it would turn into a list of values. There is no way individual flags are more limiting or harder to use. There are only so many that we'd ever add.

OliverBrotchie:
bump

bartlomieju:
> To be pointed, I am opposed to a --worker flag. It is limiting, it is ambiguous to what it does, and does not lead nicely to a future where things like ServiceWorkers are easily supported without adding a plethora of flags.

I'm also not a huge fan of `--worker` so I'd like to propose `--scope` flag - its default value would be `window` but users could do `--scope=worker` to get `Worker` type declarations. It's future proof and I think it corresponds well to the actual notion of "scope" in JavaScript (https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope).

nayeemrmn:
The intuition behind `--worker` is that it describes with no extra assumptions what kind of code the user is trying to target, without asserting exactly what Deno is doing with this information. This is the most future-proof approach IMO.

It seems future proofing to everyone else means having all lib modifiers under a single enum flag to prevent an explosion of them down the line. I think this is misconceived because 1) having a set of binary flags that each imply something about what kind of code is targeted is fine and what flags are for and 2) we have `--unstable` and soon [`--location`](https://github.com/denoland/deno/pull/10828) which should do exactly this.

I've been seeing apprehension from the core team against binary flag based type variations, most recently https://github.com/denoland/deno/pull/10828#issuecomment-854023279. I've been trying to figure out why. Perhaps you're thinking of enums such as: https://github.com/denoland/deno/blob/3b220c64f615a4e18346e8a1c07ad7b1aae9fcc6/cli/module_graph.rs#L570-L575 and dreading how it would look if we kept adding binary modifiers. The variants would double each time. The solution is to transform this to:
```rust
pub struct TypeLib {
  pub unstable: bool,
  pub worker: bool,
  pub location: bool,
  // pub dom: bool,
  // pub service_worker: bool,
}
```
We would have _one_ blackboxed function which pattern-matches legal variants of this to an appropriate set of `.d.ts` files, and we'd have to organise those `.d.ts` files more granularly. But other than that the complexity is entirely contained. I think we can and should be more open to binary type variations.

kitsonk:
As I've mentioned before, I am not sure about the binary flags.  This came up in a conversation, and there is another solution to this problem.  By adding the following to an entry point module will make it type check properly for a `deno cache` and `deno bundle`:

```ts
/// <reference no-default-lib="true" />
/// <reference lib="deno.worker" />
```

nayeemrmn:
> By adding the following to an entry point module will make it type check properly for a `deno cache` and `deno bundle`:
> 
> ```ts
> /// <reference no-default-lib="true" />
> /// <reference lib="deno.worker" />
> ```

It's never good for something written in a module to affect the way the whole graph is interpreted. If you have a library that is intended to work in both window and worker, it can't be consumed by the former if it has that directive in it. Whereas in CI I might want to check that it works by having both of `deno cache mod.ts` and `deno cache --worker mod.ts`.

nayeemrmn:
One thing I forgot to mention is the small possibility of us eventually having `deno run --worker mod.ts` to run worker scripts directly. This is an example of why it's more future-proof for the flag to unassumingly describe what code is being targeted, rather than something like `--lib worker` which asserts that it's being used for type checking.

bartlomieju:
> One thing I forgot to mention is the small possibility of us eventually having `deno run --worker mod.ts` to run worker scripts directly. This is an example of why it's more future-proof for the flag to unassumingly describe what code is being targeted, rather than something like `--lib worker` which asserts that it's being used for type checking.

Although it sounds exciting I am not sure about the actual use case. What would be the target for `postMessage()` API calls?

nayeemrmn:
> Although it sounds exciting I am not sure about the actual use case. What would be the target for `postMessage()` API calls?

There would be no target. As for use cases where one simply can't just create a normal entrypoint module which spawns the worker, I can easily say there are none. Let me replace my example with `deno test --worker` to unit test worker code, and maybe `postMessage()` calls are captured by the test harness.

My point was that given our framework of sharing flags between check time and runtime, it would be better for flags to just describe the context of the code (`--worker`) rather than use language that only applies to checking (`--lib worker`). I guess you could address this argument by instead suggesting `--mode worker`.

My first and primary point was that allowing the user to specify a binary flag (`--worker`) for each way their code departs from the defaults and letting Deno internally piece together what types/runtime config to apply, has both better UX and scalability than an enum flag (`--mode worker`), especially considering we would still have `--unstable` and `--location` which fall into the former category.

bartlomieju:
> There would be no target. As for use cases where one simply can't just create a normal entrypoint module which spawns the worker, I can easily say there are none. Let me replace my example with deno test --worker to unit test worker code, and maybe postMessage() calls are captured by the test harness.

I still can't see how we could use that in testing setup - are you suggesting that `deno test --worker` would run code in a context of a worker and run `Deno.test()` cases as normal? If so I'm not sure there's advantage to add - again, you need something to pass messages to worker. It's an interesting idea, but I would love to see a concrete example where this is desirable, so let's punt on this discussion for now.

> My point was that given our framework of sharing flags between check time and runtime, it would be better for flags to just describe the context of the code (--worker) rather than use language that only applies to checking (--lib worker). I guess you could address this argument by instead suggesting --mode worker.

I've grown to agree with you on this point. Let's try to get this PR for 1.12

ry:
We are adverse to adding new flags or compilation modes as that further complicates the programming model. 
```
/// <reference no-default-lib="true" />
/// <reference lib="deno.worker" />
```
This is a fine escape hatch for now.


Thank you for the PR @nayeemrmn but closing without merge.

PR #10053, opened 2021-04-07
ci: use incremental build in ci

kt3k:
This PR enables the incremental build of rust code in CI by the following changes:

- Enables it by env var `CARGO_INCREMENTAL=1`. (`release` profile has `incremental = false` setting by default.)
- Uses `matrix.kind` in part of cache key because `test_release` and `test_debug` can't share the cache.
- Restores each source file's timestamp based on git log. ( `./tools/restore_mtime.py` )
  - git [doesn't store timestamp of files](https://git.wiki.kernel.org/index.php/Git_FAQ#Why_isn.27t_Git_preserving_modification_time_on_files.3F), and cargo triggers rebuild of source file based on timestamps. So we need to preserve timestamp of each file somehow. I found [this script](https://github.com/MestreLion/git-tools/blob/957810b10efa77d6ca4a3aa4f76883963621f31a/git-restore-mtime) (this script restores the timestamp of files based on commit log), and it seems working.

I tested this setting in my detached fork. windows job passed in 11 minutes https://github.com/kt3k/deno_ci_test/runs/2285441107

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,10 +55,10 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v2
         with:
-          # Use depth > 1, because sometimes we need to rebuild main and if
-          # other commits have landed it will become impossible to rebuild if
-          # the checkout is too shallow.
-          fetch-depth: '0'
+          # Disables shallow clone. We use full git history for restoring the mtime
+          # of each file to enable incremental build.
+          # See https://github.com/denoland/deno/pull/10053
+          fetch-depth: 0
           submodules: true
 
       - name: Create source tarballs (release, linux)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
         'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')) }}
 
     env:
-      CARGO_INCREMENTAL: 0
+      CARGO_INCREMENTAL: 1
       RUST_BACKTRACE: full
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           # Use depth > 1, because sometimes we need to rebuild main and if
           # other commits have landed it will become impossible to rebuild if
           # the checkout is too shallow.
-          fetch-depth: 5
+          fetch-depth: '0'
```
bartlomieju, line 61: I believe this comment still stands - maybe we shouldn't go with depth 5 but 1 or 2 is highly desired

piscisaureus:
I think you should also do this for submodules. Currently only files from the "root" git repo get treated.

.github/workflows/ci.yml
kt3k: `0` means disabling shallow clone i.e. fetching all history. (this is needed for updating the modified timestamps) ref: https://github.com/actions/checkout#checkout-v2

So I'll update the comment.

piscisaureus:
We can think of ways to make this even better, but let's land it as-is first and enjoy the 2x reduction in CI build time immediately.

piscisaureus:
@bartlomieju You have the final say.

bartlomieju (approved):
> We can think of ways to make this even better, but let's land it as-is first and enjoy the 2x reduction in CI build time immediately.

Let's go 🚀

ry (approved):
LGTM - thank you Yoshiya!

PR #10055, opened 2021-04-07
fix(op_crates/fetch): handle `Request` body like `RequestInit` body

### LINKED ISSUE #10002: fetch with Request init with body of FormData is broken
`FormData` set as the `body: BodyInit` of the `RequestInit` passed to the `Request` constructor should behave the same way as the `body: BodyInit` of the `RequestInit` as the second arg to `fetch(input, init)`

Currently it doesn't, sending a request body with the text `[object FormData]`

```
const formData = new FormData();
formData.append('foo', 'bar');
const res = await fetch(new Request('http://example.com', { method: 'POST', body: formData }));
// request.text() on the server side will be "[object FormData]", request.formData() will be empty
```

deno 1.8.2 (release, x86_64-apple-darwin)
v8 9.0.257.3
typescript 4.2.2

littledivy:
Fixes #10002


lucacasonato:
Thanks, but this will be resolved in the more broad fetch rewrite in #10203.

PR #10058, opened 2021-04-07
chore: update deps

### LINKED ISSUE #10066: closing laptop lid on deno run --watch causes "interrupted system call" error
interesting error that occurred when I shut my laptop lid and reopened it on a watched process

`deno run --unstable --watch file.ts`
```
Watcher Process finished! Restarting on file change...
thread '<unnamed>' panicked at 'poll failed: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/notify-5.0.0-pre.5/src/inotify.rs:134:47
```

the watcher appears to stop working after this error is shown, but the process does not exit

lucacasonato:
Update all the deps.

Closes #10066

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -737,7 +737,7 @@ dependencies = [
  "serde",
  "tokio",
  "tokio-rustls",
- "tokio-tungstenite 0.14.0",
+ "tokio-tungstenite",
  "webpki",
  "webpki-roots",
 ]
@@ -3349,7 +3349,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tokio-rustls",
- "tokio-tungstenite 0.13.0",
+ "tokio-tungstenite",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


Cargo.lock
```diff
@@ -731,20 +737,21 @@ dependencies = [
  "serde",
  "tokio",
  "tokio-rustls",
- "tokio-tungstenite",
+ "tokio-tungstenite 0.14.0",
```
bartlomieju, line 740: Looks like there are now 2 versions of `tokio-tungstenite`

Cargo.lock
```diff
@@ -1523,8 +1527,8 @@ dependencies = [
  "httparse",
  "httpdate",
  "itoa",
- "pin-project 1.0.5",
- "socket2",
+ "pin-project",
+ "socket2 0.4.0",
```
bartlomieju, line 1531: Two versions of `socket2`?

Cargo.lock
```diff
@@ -3515,9 +3502,26 @@ checksum = "e1a5f475f1b9d077ea1017ecbc60890fda8e54942d680ca0b1d2b47cfa2d861b"
 dependencies = [
  "futures-util",
  "log",
- "pin-project 1.0.5",
+ "pin-project",
+ "tokio",
+ "tungstenite 0.12.0",
```
bartlomieju, line 3507: Multiple versions of `tungstenite`?

Cargo.lock
lucacasonato: Whoops, forgot to update test_util/Cargo.toml. We have so many `Cargo.toml` now

bartlomieju (approved):
LGTM

Cargo.lock
lucacasonato: Resolved

Cargo.lock
lucacasonato: We can't do anything about this. `hyper` needs the new one, while `trust-dns-resolver` uses the old one.

PR #10062, opened 2021-04-07
perf(metrics): reduce tracking overhead by using op_ids instead of names

AaronO:
This reduces the overhead of metrics tracking, shaving ~50ns/opcall by storing metrics in a pre-allocated vector instead of a HashMap and keying by `op_id` instead of `name` which saves the overhead of hashing those string keys.

## Todo

- [x] Tweak `OpTable` and `JsRuntime` to allow peeking at next OpId via `next_op_id()`
- [x] Change `metrics_op` to use op_ids
- [x] Fix `op_metrics` to pair up names/ids when returning metrics
- [ ] Cleanup
- [ ] Further review performance delta


runtime/metrics.rs
```diff
@@ -1,16 +1,30 @@
...
+  max_id: OpId,
+  pub ops: Vec<OpMetrics>,
+}
+
+impl Default for RuntimeMetrics {
+  fn default() -> Self {
+    // 256 is 2^8 and big enough for all of our ops
+    Self {
+      max_id: 0,
+      ops: vec![OpMetrics::default(); 256],
```
hayd, line 18: since it's fixed size, does it need to be a vec! ?

AaronO: It could in theory grow dynamically as ops are registered via plugins (plugins won't work in 1.9 but will return in 1.10).

bartlomieju (commented):
I'm going to close this PR without a merge as it became stale. I'm in favor of this change in general, but I think it requires more in-depth rewrite of how metrics as collected. Right now it leaks some concepts (`next_op_id()`, `getOpsCache`) to support this feature. Feel free to reopen.

PR #10065, opened 2021-04-08
feat(unstable): ALPN config in listenTls

### LINKED ISSUE #6334: TLS ALPN support
## Feature request 
I thiink it is necessary to support alnp. For example, if a developer want to create a http2 server,  alpn can make handshake quicker.

lucacasonato:
This commit adds the ability for users to configure ALPN protocols when
calling `Deno.listenTls`.

Closes #6334.

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1001,6 +1001,16 @@ declare namespace Deno {
     options?: StartTlsOptions,
   ): Promise<Conn>;
 
+  export interface ListenTlsOptions {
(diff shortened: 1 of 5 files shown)


lucacasonato:
I am not yet sure how we could test this. It won't be possible to do via unit tests because Deno.connectTls does not expose ALPN protocols. This will likely have to be a rust integration test using a native `rustls` client, which is relatively unfortunate.

ry:
Can we not use reqwest to test?

ry (approved):
Nice tests. LGTM

PR #10067, opened 2021-04-08
ci: use repository and ref as part of the cache key

caspervonb:
This adds the repository and ref to the cache key, making it scoped on an origin/branch pair basis to avoid clashes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
             ~/.cargo/git
             ~/.cargo/registry
             ./target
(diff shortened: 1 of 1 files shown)


lucacasonato:
Please set a fallback restore key for denoland/deno-main so the first run of a PR isn't uncached.

kt3k:
Let's merge this one first. I'll do the fallback part.

kt3k (approved):
LGTM. Thanks! @caspervonb

lucacasonato:
Discussed on discord with @caspervonb and this probably won't make it better (potentially worse). But we can try.

PR #10068, opened 2021-04-08
ci: fallback to main's cache

kt3k:
Creates cache for each branch to prevent the cache conflict. Uses main's cache for the first run in the branch.

related #10067


ry (approved):
LGTM

ry:
@kt3k I'm going to land this preemptively again because it takes so long to validate if it's working correctly. Sorry if I'm stepping on your toes.

kt3k:
@ry no problem at all :)

PR #10069, opened 2021-04-08
ci: use gnu tar on macos

kt3k:
This PR installs and uses GNU tar for extracting cache in macos runner of CI. (`actions/cache` step automatically picks gtar if it's available).

According to https://github.com/actions/cache/issues/403 BSD tar causes corruption of certain files such as `libserde_derive.dylib`. (see also https://github.com/rust-lang/cargo/issues/8603

This workaround should prevent this error https://github.com/denoland/deno/runs/2297003519


ry (approved):
LGTM

PR #10073, opened 2021-04-08
perf(core): use BTreeMap for GothamState

AaronO:
This replaces GothamState's internal `HashMap` with a `BTreeMap` to improve performance.

## Rationale

OpState/GothamState keys by `TypeId` which is essentially an opaque `u64`. For small sets of integer keys BTreeMap outperforms HashMap mainly since it removes the hashing overhead and Eq/Comp on integer-like types is very cheap, it should also have a smaller memory footprint.

We only use ~30 unique types and thus ~30 unique keys to access OpState, so the keyset is small and immutable throughout the life of a JsRuntime, there's no meaningful churn in keys added/removed.

## Benches

```
Before (main):
date_now:            	n = 500000, dt = 0.029s, r = 17241379/s, t = 58ns/op
perf_now:            	n = 500000, dt = 0.218s, r = 2293578/s, t = 436ns/op
url_parse:           	n = 50000, dt = 0.325s, r = 153846/s, t = 6500ns/op
read_zero:           	n = 500000, dt = 0.328s, r = 1524390/s, t = 656ns/op
write_null:          	n = 500000, dt = 0.324s, r = 1543210/s, t = 648ns/op

After (BTree OpState):
date_now:            	n = 500000, dt = 0.030s, r = 16666667/s, t = 59ns/op
perf_now:            	n = 500000, dt = 0.184s, r = 2717391/s, t = 368ns/op
url_parse:           	n = 50000, dt = 0.330s, r = 151515/s, t = 6600ns/op
read_zero:           	n = 500000, dt = 0.321s, r = 1557632/s, t = 642ns/op
write_null:          	n = 500000, dt = 0.318s, r = 1572327/s, t = 636ns/op
```

## Notes: future improvements & plugins

One issue with op_crate/plugin parity is that `TypeId`s aren't stable across builds/toolchain/etc... so the same type could have different TypeIds in the runtime process and the plugin. This won't cause typing issues but instead undesirable logical issues, in most cases preventing plugins from accessing the OpState set by op_crates.

Most OpState is consumed by the family of ops that set it, so it's not a strong requirement to allow accessing op_crate OpStates from plugins, in fact I would advise against it, unless there's a strong use-case for it. ResourceTable is an edge-case and can be handled individually.

However, if we wanted to have stable OpState keys, we could use something like [unique-type-id](https://github.com/vityafx/unique-type-id) (essentially allows us to hardcode IDs per type in a toml file). That would allow for stable keys across runtime & plugins (as long as edits are accretive).

Having sequential OpState TypeIds would also improve performance, since we could then trivially index those in an array.


bartlomieju (approved):
LGTM

PR #10076, opened 2021-04-08
fix(runtime/js/http): Correctly parse user response headers

nayeemrmn:
We were using a `for ... in` loop on a `Headers` object.

### DIFF
--- a/runtime/js/40_http.js
+++ b/runtime/js/40_http.js
@@ -9,15 +9,6 @@
   const core = window.Deno.core;
   const { ReadableStream } = window.__bootstrap.streams;
 
-  function flatEntries(obj) {
-    const entries = [];
-    for (const key in obj) {
-      entries.push(key);
-      entries.push(obj[key]);
-    }
-    return entries;
-  }
-
   function serveHttp(conn) {
     const rid = Deno.core.jsonOpSync("op_http_start", conn.rid);
     return new HttpConn(rid);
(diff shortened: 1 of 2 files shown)


lucacasonato:
Just realized we also have no test for non 200 status code 🙃.

lucacasonato (approved):

runtime/js/40_http.js
```diff
@@ -9,15 +9,6 @@
...
 
-  function flatEntries(obj) {
-    const entries = [];
-    for (const key in obj) {
-      entries.push(key);
-      entries.push(obj[key]);
-    }
-    return entries;
-  }
-
```
bartlomieju, line 20: @nayeemrmn this function was used intentionally - flamegraphs shown that it's way faster than `Array.flat()`. If that's possible please revert

runtime/js/40_http.js
```diff
@@ -104,12 +95,14 @@
     );
   }
 
-  function respond(responseSenderRid, resp, zeroCopyBuf) {
-    return Deno.core.jsonOpSync("op_http_response", [
-      responseSenderRid,
-      resp.status ?? 200,
-      flatEntries(resp.headers ?? {}),
-    ], zeroCopyBuf);
```
bartlomieju, line 112: Glad you removed this function, it was superfluous 👍

bartlomieju (approved):
LGTM

ry (commented):
Is it possible to add a test to cli/tests/unit/http_test.ts? Not immediately obvious to me what this is fixing.

nayeemrmn:
@ry The following assertion fails on main because no headers from the user's `Response` are properly passed to `op_http_response`: https://github.com/denoland/deno/blob/7f5f7bbaaf14b9904ea6b752178eb71b93c97818/cli/tests/unit/http_test.ts#L27 Shoehorning it into the basic test seems fair for response headers.

ry:
Nayeem - sure that works

PR #10080, opened 2021-04-09
fix(core): Connect async op error stacks with user code

nayeemrmn:
Fixes this regression caused by #9719:

img

### DIFF
--- a/core/core.js
+++ b/core/core.js
@@ -87,7 +87,7 @@
     return errorMap[errorName] ?? [undefined, []];
   }
 
-  function unwrapResponse(res) {
+  function unwrapOpResult(res) {
     // .$err_class_name is a special key that should only exist on errors
     if (res?.$err_class_name) {
       const className = res.$err_class_name;
@@ -106,12 +106,12 @@
     const promiseId = nextPromiseId++;
     const maybeError = dispatch(opName, promiseId, args, zeroCopy);
     // Handle sync error (e.g: error parsing args)
-    if (maybeError) return unwrapResponse(maybeError);
-    return setPromise(promiseId).then(unwrapResponse);
+    if (maybeError) return unwrapOpResult(maybeError);
+    return setPromise(promiseId).then(unwrapOpResult);
   }
 
   function jsonOpSync(opName, args = null, zeroCopy = null) {
-    return unwrapResponse(dispatch(opName, null, args, zeroCopy));
+    return unwrapOpResult(dispatch(opName, null, args, zeroCopy));
   }
 
   function binOpSync(opName, args = null, zeroCopy = null) {
--- a/core/core.js
+++ b/core/core.js
@@ -65,7 +65,10 @@
 
   function handleAsyncMsgFromRust() {
     for (let i = 0; i < arguments.length; i += 2) {
-      opAsyncHandler(arguments[i], arguments[i + 1]);
+      const promiseId = arguments[i];
+      const res = arguments[i + 1];
+      const promise = getPromise(promiseId);
+      promise.resolve(res);
     }
   }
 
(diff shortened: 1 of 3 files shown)


ry (commented):
Thank you Nayeem! This is an important fix.

We should probably add a regression test in this in core.

Unfortunately this reduces perf quite a bit:

before:
```
test bench_op_async   ... bench:     500,313 ns/iter (+/- 37,071)
test bench_op_nop     ... bench:      72,668 ns/iter (+/- 1,454)
test bench_op_pi_bin  ... bench:      89,801 ns/iter (+/- 1,206)
test bench_op_pi_json ... bench:      89,856 ns/iter (+/- 7,047)
```

after:
```
test bench_op_async   ... bench:     595,238 ns/iter (+/- 6,379)
test bench_op_nop     ... bench:      72,256 ns/iter (+/- 1,662)
test bench_op_pi_bin  ... bench:      86,763 ns/iter (+/- 2,822)
test bench_op_pi_json ... bench:      85,256 ns/iter (+/- 1,267)
```

core/core.js
```diff
@@ -84,48 +87,31 @@
...
-  }
-
-  function processErr(err) {
-    const className = err.$err_class_name;
-    const [ErrorClass, args] = getErrorClassAndArgs(className);
-    if (!ErrorClass) {
-      return new Error(
-        `Unregistered error class: "${className}"\n  ${err.message}\n  Classes of errors returned from ops should be registered via Deno.core.registerErrorClass().`,
-      );
+  function unwrapResponse(res) {
```
AaronO, line 90: Might make sense to call this `unwrapResult` since the value here is conceptually similar to a rust `Result`

AaronO: Or maybe `unwrapOpResult` to be a little more explicit.

AaronO (changes requested):
```js
  function jsonOpAsync(opName, args = null, zeroCopy = null) {
    const promiseId = nextPromiseId++;
    const maybeError = dispatch(opName, promiseId, args, zeroCopy);
    if (maybeError) return unwrapResponse(maybeError);
    return setPromise(promiseId).then(unwrapResponse);
  }
```

@nayeemrmn This a more performant implementation of `jsonOpAsync` that still anchors stacktraces in user code.

## Stacktraces

```
## This PR + that patch:

❯ ./target/debug/deno eval 'await Deno.open("non-existant")'
error: Uncaught (in promise) NotFound: No such file or directory (os error 2)
await Deno.open("non-existant")
^
    at unwrapResponse (deno:core/core.js:100:13)
    at async Object.open (deno:runtime/js/30_files.js:45:17)
    at async file:///Users/aaron/git/deno/$deno$eval.ts:1:1
    
## This PR:

~/git/deno pr/10080* 42s
❯ ./target/debug/deno eval 'await Deno.open("non-existant")'
error: Uncaught (in promise) NotFound: No such file or directory (os error 2)
await Deno.open("non-existant")
^
    at unwrapResponse (deno:core/core.js:100:13)
    at Object.jsonOpAsync (deno:core/core.js:110:12)
    at async Object.open (deno:runtime/js/30_files.js:45:17)
    at async file:///Users/aaron/git/deno/$deno$eval.ts:1:1
    
## v1.8.3

❯ deno -V; deno eval 'await Deno.open("non-existant")'
deno 1.8.3
error: Uncaught (in promise) NotFound: No such file or directory (os error 2)
    at handleError (deno:core/core.js:186:12)
    at jsonOpParseResult (deno:core/core.js:356:9)
    at asyncHandle (deno:core/core.js:223:40)
    at Array.asyncHandlers.<computed> (deno:core/core.js:238:9)
    at handleAsyncMsgFromRust (deno:core/core.js:207:32)
```

ry (approved):
LGTM - thank you Nayeem and Aaron for the assist

PR #10078, opened 2021-04-09
feat: add options overload to test

caspervonb:
Resolves https://github.com/denoland/deno/issues/7474

PR #10081, opened 2021-04-09
ci: do not include branch name in cache key

kt3k:
Cache service often returns 429 response at Post cache step. ref [1](https://github.com/denoland/deno/pull/9376/checks?check_run_id=2302579454), [2](https://github.com/denoland/deno/pull/10078/checks?check_run_id=2302563352), [3](https://github.com/denoland/deno/runs/2302591438) (GitHub Actions cache now has 5GB limit for a single repository. Probably we exceed this limit.)

This change prevents creating new cache on every PR branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,9 +178,7 @@ jobs:
             ~/.cargo/git
             ~/.cargo/registry
             ./target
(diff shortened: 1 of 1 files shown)


caspervonb:
lgtm.

We're still going to have bad builds, but that's a separate issue.

kitsonk (approved):
LGTM

kt3k:
Thank you for your reviews!

PR #10084, opened 2021-04-09
perf(serde_v8): build with `codegen-units=1`

littledivy:
Lets LLVM optimize code better. Decreases ~160ns from `ser_struct_v8_manual` bench. Build time impact is negligible.

### DIFF
--- a/serde_v8/Cargo.toml
+++ b/serde_v8/Cargo.toml
@@ -19,6 +19,9 @@ bencher = "0.1"
 [[example]]
 name = "basic"
 
+[profile.release]
+codegen-units = 1
+
 [[bench]]
 name = "de"
 harness = false


AaronO:
We previously built everything in release with `codegen-units=1`, #10039 removed that but #10086 reverts it, which means that this proposed optimization is actually a temporary regression that should disappear.

AaronO:
Also `ser_struct_v8_manual` isn't a benchmark we want to improve per se, it's a baseline/reference bench that compares naively/manually serializing a rust struct to v8 "by hand". 

Since `serde_v8` aims to be a "zero overhead" abstraction, it should be competitive or even beat manual serialization code (thanks to internalized keys, Key Cache, etc...).

piscisaureus:
I guess this is superseded by https://github.com/denoland/deno/pull/10086.

PR #10085, opened 2021-04-09
feat(op_crates/fetch): Add a cookie jar to `fetch`

### LINKED ISSUE #5862: Deno lacks a cookie jar for `fetch`
Deno should store cookies on a per-run basis (and optionally have methods in the Deno namespace for export/import of cookies) so that it's possible to make a series of requests that depend on cookies. 

Right now, a flow like the following would be impossible:

1. No cookies sent; response sets cookie 1.
2. Cookie 1 sent; response uses cookie 1 value somewhere in the body

Yes, I'm aware of `std/http/cookie`. No, it's not what I need - this is not a server. It's a client.

Liamolucko:
Resolves #5862, by enabling the cookie functionality built into `reqwest`.


lucacasonato (changes requested):
I am going to preemptively block this until I manage to do a full review. This is something that needs the very very very carefully considered because it is very security sensitive. This should in no scenario be enabled by default.

Liamolucko:
This approach doesn't really work, actually, because we need to support `credentials: "omit"` and `reqwest` doesn't seem to support enabling cookies per-request. I might make another attempt at implementing this in JS, but I'll close this PR.

PR #10090, opened 2021-04-09
fix(op_crates/fetch): uppercase specific methods

### LINKED ISSUE #10088: [bug] fetch() method should always be .toUpperCase
Hello 👋 

The method on `fetch()` requests should always be capitalized with .toUpperCase() in my opinion. It leads to issues when someone forgets and normally from using other projects it has always been the norm to be uppercased internally. Especially, when debugging this issue it took us hours trying different things to see what was wrong until we realized we should try making the `get` into `GET` just to see what that would do and it fixed all our issues. 

node-fetch:
https://github.com/node-fetch/node-fetch/blob/6ee9d3186f8b4bdc4288ccf92f0a776aa17644ae/src/request.js#L51

Thank You 😄

crowlKats:
Closes #10088 
ref: https://fetch.spec.whatwg.org/#methods
> To normalize a method, if it is a byte-case-insensitive match for `DELETE`, `GET`, `HEAD`, `OPTIONS`, `POST`, or `PUT`, byte-uppercase it.

### DIFF
--- a/op_crates/fetch/26_fetch.js
+++ b/op_crates/fetch/26_fetch.js
@@ -1254,18 +1254,6 @@
       headerArray = Array.from(headers.entries());
     }
 
-    const sensitiveMethods = [
-      "DELETE",
-      "GET",
-      "HEAD",
-      "OPTIONS",
-      "POST",
-      "PUT",
-    ];
-    if (sensitiveMethods.includes(method.toUpperCase())) {
-      method = method.toUpperCase();
-    }
-
     const { requestRid, requestBodyRid } = opFetch(
       {
         method,
--- a/op_crates/fetch/26_fetch.js
+++ b/op_crates/fetch/26_fetch.js
@@ -1254,6 +1254,18 @@
       headerArray = Array.from(headers.entries());
     }
 
+    const sensitiveMethods = [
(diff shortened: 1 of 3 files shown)


op_crates/fetch/26_fetch.js
```diff
@@ -1254,6 +1254,18 @@
...
+    const sensitiveMethods = [
+      "DELETE",
+      "GET",
+      "HEAD",
+      "OPTIONS",
+      "POST",
+      "PUT",
+    ];
+    if (sensitiveMethods.includes(method.toUpperCase())) {
+      method = method.toUpperCase();
```
nayeemrmn, line 1266: Should be done in Rust IMO.

crowlKats: done

nayeemrmn (approved):
LGTM

nayeemrmn (changes requested):
Actually it needs a test. Fetch against a TCP listener and check that the method is upper cased.

bartlomieju (changes requested):
Please add a test

lucacasonato:
Thanks for the effort, but this will be resolved by #10203.

PR #10103, opened 2021-04-10
fix(runtime/js/timers): Use (0, eval) instead of eval()

### LINKED ISSUE #10091: Use of eval in setTimeout/setInterval leaks internal scope
```ts
setTimeout("globalThis.console.log(timer)", 1000);
// {
//   id: 1,
//   callback: "globalThis.console.log(timer)",
//   args: [],
//   delay: 1000,
//   due: 1617982526862,
//   repeat: false,
//   scheduled: false
// }
```
https://github.com/denoland/deno/blob/cd0560210aa0ac9dc9772cdddcb1c5a59cef1815/runtime/js/11_timers.js#L445

nayeemrmn:
Fixes #10091.

### DIFF
--- a/runtime/js/11_timers.js
+++ b/runtime/js/11_timers.js
@@ -442,7 +442,7 @@
     if ("function" === typeof callback) {
       callback();
     } else {
-      eval(callback);
+      (0, eval)(callback);
     }
   }
 
--- a/cli/tests/unit/test_util.ts
+++ b/cli/tests/unit/test_util.ts
@@ -21,6 +21,7 @@ export {
   unreachable,
 } from "../../../test_util/std/testing/asserts.ts";
 export { deferred } from "../../../test_util/std/async/deferred.ts";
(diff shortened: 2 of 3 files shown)


runtime/js/11_timers.js
```diff
@@ -442,7 +442,10 @@
     if ("function" === typeof callback) {
       callback();
     } else {
-      eval(callback);
+      const result = core.evalContext(callback);
+      if (result[1] != null) {
+        throw result[1].thrown;
```
bnoordhuis, line 447: Would `(0, eval)(callback)` work? Rethrowing means this `throw` statement shows up as the exception's origin rather than the evaluated code.

nayeemrmn: I didn't know about `globalThis.eval` vs `eval`. Fixed.

bnoordhuis (approved):
LGTM. One suggestion: maybe add a regression test?

bartlomieju (approved):
LGTM

PR #3596, opened 2020-01-04
Text encoding speedup

lucacasonato:
I have sped up the TextEncoder for UTF-8 encoding by utilizing a encoding function from https://github.com/samthor/fast-text-encoding.

The speed improvement is about 4.5x. This is probably because there are way less allocations as the encoded bytes are written right into a Uint8Array. Maybe this can be optimized further, but it is a definite improvement over what we have now.

I did not do this for TextEncoder#encodeInto as I can not guarantee that passed Uint8Array() is long enough for the content. TextEncoder#encodeInto still uses the old encoder.

PR #10104, opened 2021-04-10
refactor(op_crates/web): Remove unused code path in TextEncoder

tarruda:
According to https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder, TextEncoder should ignore the "encoding" parameter and always use "utf-8". It seems the current implementation already uses "utf-8" unconditionally.

Even if it didn't, a JS implementation of UTF-8 encoding is used, so it is better to leave only the rust implementation which is more efficient.

### DIFF
--- a/op_crates/web/08_text_encoding.js
+++ b/op_crates/web/08_text_encoding.js
@@ -4212,25 +4212,8 @@
   class TextEncoder {
     encoding = "utf-8";
     encode(input = "") {
-      input = String(input);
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

Does anyone have any concern? cc @nayeemrmn @lucacasonato @bartlomieju 

BTW this condition (`this.encoding === "utf-8`) was introduced in https://github.com/denoland/deno/pull/3596

lucacasonato (approved):

PR #10114, opened 2021-04-10
feat(core/runtime): Feature gate ICU

littledivy:
Leverages cargo features to feature gate ICU initialization. By default, the feature is enabled but can be disabled with `features = ["no_icu"]`

Use case: I'm using Deno for [embedding purposes](https://github.com/littledivy/deno_compile) and would like to not inlcude ICU data by default into binaries.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -12,6 +12,10 @@ repository = "https://github.com/denoland/deno"
 [lib]
 path = "lib.rs"
 
+[features]
+no_icu = []
+default = []
+
 [dependencies]
 anyhow = "1.0.40"
 futures = "0.3.13"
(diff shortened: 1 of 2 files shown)


lucacasonato:
This does not work. If rusty_v8 / v8 are built with ICU support, not supplying ICU info and calling `Intl` APIs will result in a segfault.

bnoordhuis:
The V8 team has hinted that ICU may become a required dependency at one point so this is probably not a good long-term solution anyway.

What could work is the "small-icu" approach that Node.js used to take (slimmed down ICU data file that only supports en_US) but they moved away from that after ample discussion: https://github.com/nodejs/node/issues/19214

(I say "they" but I was involved in the decision making around both small-icu and full-icu. In retrospect, I'd say small-icu was a mistake.)

littledivy:
Fair, will close this PR since it introduces segfaults and might not be good in the long term

Issue #10113, opened 2021-04-10
Improving the DevTools experience


AaronO:
Chrome's DevTools provide essential debugging & profiling tools, they mostly work with Deno, but there are a few quirks that make it unintuitive and painful to use (especially for profiling JS performance).

## Main Issues

1. `chrome://inspect` will open the "wrong" DevTools for Deno: users looking to profile will naturally go to the `Performance` tab but will get a `Tracing.start() not found` error, they would instead need to add a `Javascript Profiler` tab which isn't visible by default. 
2. Process exits after script execution, even if debugger is attached, making it hard to use on one-off scripts (e.g: `bench/deno_common.js`)

## Solutions

1. Pretend to be `node` so we leverage chrome's builtin support for inspecting node processes
2. Ultimately add upstream support in `chrome-devtools-frontend` and Chromium for Deno (add a `deno_app` similar to `node_app`)
3. Don't exit process if debugging session is active

## Notes

- Longterm, we could do more sophisticated tracing etc... but fixing these 2 issues, should provide a debugging experience similar to Node
- We could also serve our own DevTools `js_app.html`, so users don't have to go through `chrome://inspect` and would have a clickable link they could open from their terminal (e.g: CMD+click in iTerm on macOS)

PR #10115, opened 2021-04-10
inspector: leverage Chrome DevTools' builtin support for NodeJS

AaronO:
This partially fixes https://github.com/denoland/deno/issues/10113, by allowing Deno to use Chrome's builtin support for debugging NodeJS processes. Longterm we'll want to add first-class Deno support in devtools & chromium.

## Impact

Users are shown a significantly more relevant DevTools view when clicking `inspect` (see screenshots below) and can also use `Open dedicated DevTools for Node` which will open a DevTools window (with relevant views) that auto-attaches to newly spawned inspectable (deno) processes, which is nice.

## Comparison of DevTools views

This compares the view users get when they go to `chrome://inspect` and click on `inspect`.

### Before

You'll notice this view includes web specific tabs such as Elements (DOM), Lighthouse, Application, etc...

image

### After

This view now contains only the relevant tabs and (CPU) `Profiler` instead of `Performance`

image

### DIFF
--- a/runtime/inspector.rs
+++ b/runtime/inspector.rs
@@ -112,7 +112,7 @@ impl InspectorInfo {
       "faviconUrl": "https://deno.land/favicon.ico",
       "id": self.uuid.to_string(),
       "title": self.get_title(),
-      "type": "deno",
+      "type": "node",
       // TODO(ry): "url": "file://",
       "webSocketDebuggerUrl": self.get_websocket_debugger_url(),
     })
@@ -124,7 +124,7 @@ impl InspectorInfo {
 
   fn get_frontend_url(&self) -> String {
     format!(
-      "devtools://devtools/bundled/inspector.html?v8only=true&ws={}/ws/{}",
+      "devtools://devtools/bundled/js_app.html?ws={}/ws/{}&experiments=true&v8only=true",
       &self.host, &self.uuid
     )
   }


ry (approved):
LGTM

june07:
Currently working on the NiM project which is a UI for debugging V8 (Node.js and Deno) and this PR actually answers two questions have had regarding changes to this project!  The first being why "deno" was changed to "node" in the info json endpoint, which happened to break some things on our end, and also more recently why I lost devtools debugging features.

To both issues, I simply created workarounds in NiM. But it seems like a better solution might be available other than intentionally misnaming the runtime, like maybe adding something instead or in addition to? Idk. I have to say that it was one of those cases where I thought I remembered it being a certain way and then it wasn't, but I started doubting my memory...

The second issue has to do with Deno only providing a single devtoolsFrontendUrl endpoint vs Node which provides a compatibility endpoint as well... see:

https://github.com/nodejs/node/blob/main/src/inspector_socket_server.cc#L376-L384
vs
https://github.com/denoland/deno/blob/896dd56b7af06fea6604a5596a6ffd17e7e52e6e/runtime/inspector_server.rs#L412-L417

Might it be useful to also have both endpoints here? It certainly makes a difference here: https://github.com/june07/nimv3/blob/main/src/sw.js#L417-L418

where again, simply had to write a workaround to get a usable devtools environment using the compat endpoint... because the js_app.html endpoint does not provide a lot of features like a browsable file tree for example see the gif... but also the info I think should be usable by third party apps/etc to have a definitive answer to which runtime is actually present. Elsewhere in backend code of linked projects, the info is also used to make determinations based on the runtime... for one, Deno's use of '/ws/' in the URL vs the absence of this path when using Node. The hack I put in place is just a regex and not as sure as depending on info.

Here you can see what I mean where the compat version has the file browser while the other does not...

deno10115

Not trying to be a complainer and of course I don't know the full scope of things.

Would love to get some feedback on this. I'm curious also if you could point to where in the DevTools code you found that changing the runtime results in the "Before/After" changes you mentioned?

I could submit a PR to basically emulate what Node currently does by offering both endpoints... but just wanted to check here first. As well there's some uncertainty about my last question.

Thanks.

bartlomieju:
> Currently working on the NiM project which is a UI for debugging V8 (Node.js and Deno) and this PR actually answers two questions have had regarding changes to this project! The first being why "deno" was changed to "node" in the info json endpoint, which happened to break some things on our end, and also more recently why I lost devtools debugging features.

> To both issues, I simply created workarounds in NiM. But it seems like a better solution might be available other than intentionally misnaming the runtime, like maybe adding something instead or in addition to? Idk. I have to say that it was one of those cases where I thought I remembered it being a certain way and then it wasn't, but I started doubting my memory...

This is intentional as both runtimes are similar - not using `node` here results in fewer panes being available in the Devtools. There were talks about adding dedicated devtools for "deno" however due to limited bandwidth no one has picked up this task (it's roughly 2-4 weeks of work to do it).

> The second issue has to do with Deno only providing a single devtoolsFrontendUrl endpoint vs Node which provides a compatibility endpoint as well... see:

I'm not sure I understand why the other URL is needed? The code that generates the URL is here: https://github.com/denoland/deno/blob/2b247be517d789a37e532849e2e40b724af0918f/runtime/inspector_server.rs#L408

> Would love to get some feedback on this. I'm curious also if you could point to where in the DevTools code you found that changing the runtime results in the "Before/After" changes you mentioned?

I'm not sure I understand this question either but I'll try to answer anyway: we found this empirically that using `node` type provides a better integration with Devtools. Again - ideally we'd have "deno" type, but that requires a lot of work that no one had time to look into yet.

PR #10116, opened 2021-04-10
feat(core): enable wasm threading support

caspervonb:
This enables support for atomics and shared array buffers as the backing store of WebAssembly.Memory when shared: true is given in the options bag.

Both Firefox and Chrome have this enabled by default right now, so seems appropriate that we also adopt it early (the proposal is only stage 2 at this time).

Fixes https://github.com/denoland/deno/issues/9006#issuecomment-817241019


satyarohith (commented):
Great! We might need to update `MemoryDescriptor`  to include `shared` option: https://github.com/denoland/deno/blob/00e63306cbcc295a87ba662f9f63311a3c6c49ce/cli/dts/lib.deno.shared_globals.d.ts#L170

lucacasonato:
Please also update the typings to include `shared: true` in the options bag and specify that `WebAssembly.Memory` can return `SharedArrayBuffer`,

bartlomieju (approved):
LGTM, thanks Casper

PR #10117, opened 2021-04-10
fix(worker): continue polling if inspector is active

AaronO:
This is another partial fix of #10113, it prevents the process from exiting if an inspector session is still active.

Otherwise, it's impossible to correctly profile a script, since the process exits before sending the profiling samples to the inspector.

## Note

- I believe when profiling node programs, CPU profiling sessions are automatically "wrapped up" when the main script finishes, this behaviour isn't yet implemented, users have to manually stop profiling in the DevTools.


runtime/inspector.rs
```diff
@@ -414,6 +414,13 @@ pub struct DenoInspector {
   pub debugger_url: Option<String>,
 }
 
+impl DenoInspector {
+  pub fn is_active(&self) -> bool {
```
piscisaureus, line 418: Nit: it might be good to be less ambiguous and call this method `has_active_sessions()` or `has_connected_clients()` even.

runtime/worker.rs
```diff
@@ -261,8 +261,27 @@ impl MainWorker {
     cx: &mut Context,
   ) -> Poll<Result<(), AnyError>> {
     // We always poll the inspector if it exists.
-    let _ = self.inspector.as_mut().map(|i| i.poll_unpin(cx));
```
piscisaureus, line 264: The logic looks correct to me. It could be a bit easier to read, it seems that it boils down to:

```
if js_runtime_is_alive {
  if has_inspector { let _ = inspector.poll(); }
  js_runtime.poll()
} else if has_inspector && inspector.has_active_sessions {
  inspector.poll() 
}
```

I guess you wanted to not duplicate the `inspector.poll()` call, which I can understand.

AaronO: Yeah, precisely. We also didn't have a helper method to tell if the runtime is "alive" or not it's precisely the output of poling the runtime that provided that info, we could promote this code in `JsRuntime::poll_event_loop()` to a helper method:
```rust
    let state = state_rc.borrow();
    let has_pending_ops = !state.pending_ops.is_empty();

    let has_pending_dyn_imports = !{
      state.preparing_dyn_imports.is_empty()
        && state.pending_dyn_imports.is_empty()
    };
    let has_pending_dyn_module_evaluation =
      !state.pending_dyn_mod_evaluate.is_empty();
    let has_pending_module_evaluation = state.pending_mod_evaluate.is_some();

    if !has_pending_ops
      && !has_pending_dyn_imports
      && !has_pending_dyn_module_evaluation
      && !has_pending_module_evaluation
    {
      return Poll::Ready(Ok(()));
    }
 ```

File: core/runtime.rs [hunk 3]
```diff
@@ -194,6 +194,15 @@ pub struct RuntimeOptions {
   pub create_params: Option<v8::CreateParams>,
 }
 
+/// Status of pending tasks in the event loop
+#[derive(Debug, PartialEq, Eq)]
+pub enum PendingState {
+  Active,
+  Idle,
+  BlockedModule,
```
bartlomieju, line 202: IMO this is misleading, they are not blocked but rather "unresolved"

runtime/inspector.rs
AaronO: Renamed to `has_active_sessions()`

core/runtime.rs
AaronO: Renamed

bartlomieju:
We're gonna get conflict with #10656, I can look into this PR after landing #10656

bartlomieju:
Thank you for the PR @AaronO, after changes to inspector I had to rework this logic a bit and it's now included in https://github.com/denoland/deno/pull/10763.

PR #10119, opened 2021-04-10
Reland #9774: Set useDefineForClassFields to true

nayeemrmn:
#9774. The option is now also passed to swc, which supports it as of recently.


kitsonk (approved):
LGTM

IiRC, we will need to update some things in std as well it is likely to cause some breakage (good though) in the community.

lucacasonato:
`std` has already been adapted in https://github.com/denoland/deno_std/pull/807.

PR #10120, opened 2021-04-10
feat: blob URL support in fetch

lucacasonato:
This commit adds blob URL support in `fetch`. Tested via WPT. This is
the first op_crate to have a rust dependency on a different op_crate.


bartlomieju (approved):
LGTM

PR #10123, opened 2021-04-11
core: avoid async op future reboxing to bundle PromiseId

AaronO:
This is another op-layer optimization, it reduces the baseline overhead of async ops by ~10%. 

It removes an extra alloc/free per opcall that happened due to "reboxing" the future to insert the `PromiseId`. The `PromiseId` is now passed as a field of `OpPayload` so the async OpFn wrappers can insert it when mapping the future result.

I have another experiment in progress, but wanted to land this since it's simple and 10% is a respectable gain even if it's not "game changing".

## Benches

```
Before:
test bench_op_async   ... bench:     533,080 ns/iter (+/- 14,308)
test bench_op_nop     ... bench:      71,169 ns/iter (+/- 2,998)
test bench_op_pi_bin  ... bench:      87,043 ns/iter (+/- 3,089)
test bench_op_pi_json ... bench:      85,530 ns/iter (+/- 1,801)

After:
test bench_op_async   ... bench:     474,119 ns/iter (+/- 8,209)
test bench_op_nop     ... bench:      71,417 ns/iter (+/- 3,776)
test bench_op_pi_bin  ... bench:      86,535 ns/iter (+/- 3,084)
test bench_op_pi_json ... bench:      86,308 ns/iter (+/- 1,388)
```

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -9,7 +9,6 @@ use crate::OpResponse;
 use crate::OpTable;
 use crate::PromiseId;
 use crate::ZeroCopyBuf;
-use futures::future::FutureExt;
 use rusty_v8 as v8;
 use serde::Serialize;
 use serde_v8::to_v8;
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -433,7 +433,7 @@ fn send<'s>(
     }
   };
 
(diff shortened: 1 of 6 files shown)


piscisaureus (approved):
Awesome!

PR #10126, opened 2021-04-11
feat(runtime): stabilize Deno.ftruncate and Deno.ftruncateSync

caspervonb:
This stabilizes Deno.ftruncate and Deno.ftruncateSync.

This is a well known system call and the interface is not going to change.
Implicitly requires write permissions as the file has to be opened with write to be truncated.


piscisaureus:
I don't have any objections. These APIs are not so pretty but they're very standard.
Can I suggest to also document their behavior when `len` is greater than the current length (the file is extended with zeroes, I believe).

bartlomieju (approved):
LGTM

PR #10128, opened 2021-04-11
fix(cli/dts): Make respondWith() return a Promise

### LINKED ISSUE #10101: RequestEvent::respondWith() should return a promise
https://github.com/denoland/deno/blob/e7b7129b7a92b7500ded88f8f5baa25a7f59e56e/cli/dts/lib.deno.unstable.d.ts#L1200-L1203

To align with https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith#return_value the current definition returns `void`, but this is untenable for us. Async errors that occur when writing the response need to be handle-able by the user. Note that the JS implementation already returns a promise, we just need to fix the types.

```ts
 export interface RequestEvent { 
   readonly request: Request; 
   respondWith(r: Response | Promise<Response>): Promise<void>; 
 }
```

<details>
<summary>Rename tangent</summary>

As we diverge further from `FetchEvent`, I also think we should improve the function's name to just `respond()`.
```ts
 export interface RequestEvent { 
   readonly request: Request; 
   respond(r: Response | Promise<Response>): Promise<void>; 
 }
```
</details>

nayeemrmn:
Fixes #10101.

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1209,7 +1209,7 @@ declare namespace Deno {
 
   export interface RequestEvent {
     readonly request: Request;
-    respondWith(r: Response | Promise<Response>): void;
+    respondWith(r: Response | Promise<Response>): Promise<void>;
   }
 
   export interface HttpConn extends AsyncIterable<RequestEvent> {
--- a/runtime/js/40_http.js
+++ b/runtime/js/40_http.js
@@ -112,8 +112,8 @@
     ], zeroCopyBuf);
   }
 
-  function createRespondWith(responseSenderRid, connRid) {
-    return async function (resp) {
+  function createRespondWith(responseSenderRid) {
+    return async function respondWith(resp) {
       if (resp instanceof Promise) {
         resp = await resp;
       }


bartlomieju:
I am not a fan of this change - either we preserve the API from service workers and `respondWith` doesn't return promise or let's just rename this function to `respond`. CC @ry

bartlomieju:
> I am not a fan of this change - either we preserve the API from service workers and `respondWith` doesn't return promise or let's just rename this function to `respond`. CC @ry

@nayeemrmn I agree with your rationale here: https://github.com/denoland/deno/issues/10101#issuecomment-817135657 - I mentioned this problem to Ryan and Luca several times. I do agree that function to respond should be awaitable, but I don't think we should change signature of `respondWith` to make it non-compatible with Web API, hence my suggestion to name it `respond`

kitsonk:
It is critical that we detect errors that occur with `.respondWith()` in the underlying ops...  For example, I recently came across a situation where `curl` was "misbehaving" and causing:

```
    at async Object.respondWith (deno:runtime/js/40_http.js:152:11)
```

While it isn't documented, `.respondWith()` does currently return a `Promise` than I can use to catch this error, without that, the error goes unhandled and the process exits, which is 😢  for a web server to crash because a client doesn't quite behave correctly.

Also, there is a need in some cases to know when `.respondWith()` completes, so that resources associated with any response can be cleaned up.  It is difficult/challenging at times to ensure that all resources can be released (like file handles closed) until we know the server isn't going to reference them anymore.

ry (approved):
LGTM - tho I have to say, it seems that Service Worker standard really intends for this to be a sync function and that error handling should somehow be done in the promise passed to respondWith... That said, let's do this for now.

jimmywarting:
> While it isn't documented, .respondWith() does currently return a Promise than I can use to catch this error, without that, the error goes unhandled and the process exits, which is 😢 for a web server to crash because a client doesn't quite behave correctly.

exactly what happen to me today... totally unexpected. 
thing `respondWith` should have stayed the way it was and then you had to catch the error in any other way instead. 

just an idea:
```js
  const httpConn = Deno.serveHttp(conn)

    const { request } = requestEvent
    httpConn
      .when(request)
      .upload(progressEvent => { 'calculate how much have been uploaded and how much is left' })
      .download(progressEvent => { '...' })
      .aborted(evt => { 'user aborted or closed the request' }) 
      .respondWith('2xx', evt => { })
      .respondWith('3xx', evt => { })
      .respondWith('4xx', evt => { })
      .respondWith('5xx', evt => { })
      .completed(evt => { })
```

PR #10129, opened 2021-04-11
fix(op_crates/web) Make TextEncoder.encodeInto more efficient. Also fixes #9006

### LINKED ISSUE #9006: bug: encodeInto implementation is not correct

tarruda:
The current implementation of TextEncoder.encodeInto does a lot of allocations which significantly impacts performance. This new implementation aims to be more efficient and correct (fixes part of #9006, the other part is fixed by #10116).

To show the performance gains, the first commit adds a benchmark (cli/tests/text_encoder_into_perf.js) that can be run with the old version. Here are the results:

```
Benchmark #2: /home/tarruda/deno/target/release/deno run cli/tests/text_encoder_into_perf.js
  time (mean ± σ):     829.6 ms ±  12.1 ms    [user: 837.5 ms, system: 25.6 ms]
  range (min … max):   808.1 ms … 849.2 ms    10 runs
```

These are the new results:

```
Benchmark #2: /home/tarruda/deno/target/release/deno run cli/tests/text_encoder_into_perf.js
  Time (mean ± σ):     144.8 ms ±   3.2 ms    [User: 146.3 ms, System: 7.1 ms]
  Range (min … max):   140.1 ms … 149.8 ms    20 runs
```

The old version is uses about 5.7x the time used by the new version. Here's how it compares with TextEncoder.encode (backed by a very efficient Rust implementation):

```
Benchmark #1: /home/tarruda/deno/target/release/deno run cli/tests/text_encoder_perf.js
  Time (mean ± σ):      75.0 ms ±   4.5 ms    [User: 67.4 ms, System: 15.1 ms]
  Range (min … max):    66.1 ms …  85.1 ms    38 runs
```


tarruda:
I've rebased the PR on `main` to obtain the changes from #10104 and removed some extra unused code (`UTF8Encoder` class and `stringToCodePoints` function)

tarruda:
I've rebased the PR on `main` to obtain the changes from #10104 and removed some extra unused code (`UTF8Encoder` class and `stringToCodePoints` function)

CrimsonCodes0:
If performance matters, couldn't this be directly be written in Rust for better performance than the current `TextEncoder#encode`?

Although it's not required, users and libraries expect `TextEncoder#encodeInto` to be implemented in a more performant manner than `TextEncoder#encoder`. Rust has proper integer types, and direct buffer writing capabilities, which should help a lot.

tarruda:
> If performance matters, couldn't this be directly be written in Rust for better performance than the current TextEncoder#encode?

An implementation of `TextEncoder#encodeInto` that uses the rust version to to encode to a temporary buffer, then copies only what fits into the input buffer passed to  `TextEncoder#encodeInto`, would probably beat the implementation here in benchmarks. But raw benchmark timings are not telling the whole story. 

The purpose of `encodeInto` is to avoid temporary allocations, which ultimately puts less pressure in V8 garbage collector. If we could write directly to Javascript heap from Rust, then yes a pure Rust implementation would be better, but I'm not familiar enough with Rust or V8 API to know if that is possible.

CrimsonCodes0:
> TextEncoder#encodeInto that uses the rust version to encode to a temporary buffer, then copies only what fits into the input buffer passed to `TextEncoder#encodeInto`, would probably beat the implementation here in benchmarks...

Oh no, for clarification, I'm suggesting just rewriting the JS you used into Rust, and writing into the Uint8Array's raw buffer.

To make a temporary buffer and then copy into the Uint8Array would likely be much slower, as it would do a lot of unnecessay work; consider a 1000 character long UTF-16 string, passed along with a 2-byte Uint8Array, the buffer could be at most roughly 4000 bytes, while only most, 8 would be needed.

The V8 API's ArrayBuffer class provides a method to access the underlying bytes, but I'm not sure if rusty_v8 does.

Regardless, don't let my comment detract you, those are good performance numbers! Someone could port it to Rust later, if necessary.

tarruda:
> Regardless, don't let my comment detract you, those are good performance numbers! Someone could port it to Rust later, if necessary.

Later I might investigate and give it a shot, it would be a good exercise to get more familiar with Rust and V8.

tarruda:
Rebased to resolve conflicts and update expectations.json for encodeInto (after #10116 and this PR, only one encodeInto test is failing).

lucacasonato (approved):
Awesome. Thanks a lot for the fix, and the performance improvements. Those are always very welcome :-)

LGTM

Issue #9734, opened 2021-03-09
Reenable "compile_windows_ext" test


bartlomieju:
Disabled in https://github.com/denoland/deno/pull/9732 because of flakiness on `main` branch:
- https://github.com/denoland/deno/runs/2068592281
- https://github.com/denoland/deno/runs/2057407502

CC @littledivy

kt3k:
closed by https://github.com/denoland/deno/pull/10131

PR #10131, opened 2021-04-11
chore(ci): try fixing #9734 test

littledivy:


piscisaureus (approved):
Looks like the bug has been fixed already, and the test can be enabled.

bartlomieju (approved):

PR #10110, opened 2021-04-10
ci: store last-modified timestamps in Github Actions cache

piscisaureus:

PR #10132, opened 2021-04-11
ci: touch diff from main

kt3k:
The current CI cache has problem with detection of newer source files. If build cache is created after any of branch's commit, cargo failed to detect newer source files, and it causes false negative rebuilding.

To avoid above problem, this workaround touches all files changed from main (`git diff --name-only origin/main HEAD`).

I tested this script in #9376, which once caused the above problem in https://github.com/denoland/deno/runs/2314683208, but fixed with this script in https://github.com/denoland/deno/runs/2316199508 (the build time of windows job took reasonable amount of time, 20 min, about a half of full build.

(alternative approach to https://github.com/denoland/deno/pull/10110)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,6 +163,10 @@ jobs:
       - name: Restore timestamps
         run: python ./tools/restore_mtime.py
 
+      - name: Touch diff from main
+        shell: bash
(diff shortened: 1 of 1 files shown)


kt3k:
closing in favor of #10110. I believe this has the same effect with #10110

PR #10138, opened 2021-04-11
fix: incorrect type for Env.toObject

sheepsteak:
Since an object is returned it's possible that for a given key there could be no value:

image


nayeemrmn:
This change is correct, but unfortunately it's breaking and may have to wait for 2.0...

lucacasonato:
Closing as 2.0 is still a while out. @nayeemrmn do you know if we have a tracking issue for this?

nayeemrmn:
I've opened #10267.

lucacasonato:
Thanks 🙏

PR #10145, opened 2021-04-12
refactor(deno): remove concept of bin & json ops

AaronO:
In favour of just ops, either sync or async.


core/ops_json.rs
```diff
@@ -25,18 +25,18 @@ use std::rc::Rc;
 /// When registering an op like this...
```
ry, line 25: Maybe this module should be merged into core/ops.rs ? At least "json" should be removed from the name.

core/examples/http_bench_ops.rs
```diff
@@ -217,10 +217,7 @@ fn main() {
 
   let future = async move {
     js_runtime
-      .execute(
-        "http_bench_json_ops.js",
-        include_str!("http_bench_json_ops.js"),
-      )
+      .execute("http_bench_ops.js", include_str!("http_bench_ops.js"))
       .unwrap();
```
ry, line 221: Since this example is being renamed, maybe we should also use "tcp" instead of "http" in the name?

This would make it correspond better with [deno_tcp.ts](https://github.com/denoland/deno/blob/9d53dab4df64361f00dda2534eecbd04797d56a2/cli/bench/deno_tcp.ts)

AaronO: Makes sense, we could also call it `htcp` to indicate that it's not fully http but not unstructured tcp either.

core/ops_json.rs
AaronO: Yeah, I was planning to rename this file in a separate "cleanup" pass that would also update the docstrings. Given that this PR already touches a lot of code I wanted to keep things separate.

core/examples/http_bench_ops.rs
AaronO: Reverted this name change, it's somewhat unimportant to this PR and can be changed later. Changing this has the undesirable side-effect of making benchmarks (on our benchmarks page) harder to follow.

core/ops_json.rs
AaronO: Leaving this for another PR

ry (approved):
LGTM - thanks Aaron

Issue #10049, opened 2021-04-07
TLS not properly flushed, causing "large" responses to arrive incompletely


Pwuts:
I've run into an issue with `serveTLS` where serving "large" files (Nuxt javascript files in this case) fails. The request completes on the server side, or so it seems, but the socket does not close and thus the client (browser or cURL) does not stop loading.

POC:
```TS
import { serveTLS } from "https://deno.land/std/http/server.ts";

const options = {
  hostname: "localhost",
  port: 8443,
  certFile: "/tmp/foo.crt",
  keyFile: "/tmp/foo.key",
};

for await (const req of serveTLS(options)) {
  req.respond({ new TextEncoder().encode("A".repeat(2**17)); });  // 128 kiB
}
```
My browser (chrome) stops after exactly receiving 64kiB, or finishes if the file is smaller than that. Trying to get the same files with cURL only gives me 32kiB.

## Possible cause
@Juerd, @supakeen and I figured out it might have something to do with the fact that in [`writeResponse`](https://deno.land/std/http/_io.ts#L232) on [line 289](https://deno.land/std/http/_io.ts#L289), [`writer.flush()`](https://deno.land/std/io/bufio.ts#L465) is called, which writes the contents of its buffer to the underlying writer, but does not flush this underlying layer in case it has a buffer of itself, which is the case with TLS.

**Is it possible someone missed [this particular comment](https://github.com/tokio-rs/tls/blob/master/tokio-rustls/src/server.rs#L95-L96) in Tokio TLS, the library Deno uses for its TLS implementation?** (kudo's to @juerd for finding that)

piscisaureus:
Possibly related: https://github.com/denoland/deno/issues/9692

Pwuts:
Yep, sounds like it's the same issue or at least very much related. However, I experienced problems with the latest version of deno (1.8.3) on CentOS 8, didn't test on Windows or in WSL.

piscisaureus:
I think the problem lies with deno_std's HTTP server implementation and not with our implementation of TLS. 
The following test implements a "fake" HTTPS server that sends 128kb responses; when I hit it with `curl`, I always get the full response.

```
const listener = Deno.listenTls({
  hostname: "localhost",
  port: 8443,
  certFile: "./cert.pem",
  keyFile: "./key.pem",
});

const EOL = "\r\n";
let n = 0;

for await (const conn of listener) {
  const char = "a".charCodeAt(0) + (n++ % 26);
[...]

Issue #9692, opened 2021-03-05
Elusive TLS Windows issue with binary data (png files)


trgwii:
Okay, this is a very strange issue, so I'm not very sure where to put this. Let's say I have this Deno HTTPS server:
```ts
import { serveTLS } from "https://deno.land/std@0.89.0/http/server.ts";

const server = serveTLS({
  port: 443,
  hostname: "0.0.0.0",
  certFile: "cert/fullchain.pem",
  keyFile: "cert/key.pem",
});

for await (const req of server) {
  (async () => {
    const file = await Deno.open("image.png");
    await req.respond({
      headers: new Headers({
        "Content-Type": "image/png",
      }),
      body: file,
    });
    await req.done;
    file.close();
  })();
}
```

Using the following image file:
![image](https://user-images.githubusercontent.com/11262022/110100122-58a03380-7da2-11eb-86ac-1693f29cc460.png)

I'm on Windows 10.0.19042.804.

(Edit) Deno version:
```
deno 1.8.0 (release, x86_64-pc-windows-msvc)
v8 9.0.257.3
typescript 4.2.2
```

On Windows, when I request this server from a browser or curl, it seems to only load the first 64KiB and then hang the request:
![2021-03-05_11-04-11](https://user-images.githubusercontent.com/11262022/110100340-94d39400-7da2-11eb-8c90-f089c9304e69.png)

If I run the same program in Deno in WSL (1) it works completely fine.

Also if I try to set the file to a 2.5MB JS file, it loads the entire file fine on Windows too, and doesn't hang the request, so there seems to be something that ONLY fails for Windows + binary files going on here.

Am I doing something wrong in my code?

kt3k:
probably related to https://github.com/denoland/deno_std/issues/750? Can you reproduce this issue with v1.7.0 (you can downgrade deno by the command `deno upgrade --version 1.7.0`

ry:
Reopening. Fix revert in https://github.com/denoland/deno/pull/10595.

PR #10146, opened 2021-04-12
fix(tls): flush send buffer in the background after closing TLS stream

piscisaureus:
In #9118, TLS streams were split into a "read half" and a "write half"
using `tokio::io::split()` to allow concurrent `Conn#read()` and
`Conn#write()` calls without one blocking the other. However, this
introduced a bug: outgoing data gets discarded when the TLS stream is
gracefully closed, because the read half is closed too early, before all
TLS control data has been received.

Fixes: #9692
Fixes: #10049
Fixes: #10296
Fixes: denoland/deno_std#750


bnoordhuis:
Fails several tests with this error:

> AssertionError: Expected error to be instance of "TypeError", but got "ConnectionRefused".

Is this ready for review or should I wait a little longer?

piscisaureus:
@bnoordhuis 

> Is this ready for review or should I wait a little longer?

It is ready for review.

runtime/ops/tls.rs
```diff
@@ -70,6 +85,421 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+
+pub struct TlsStream {
+  tls: TlsSession,
+  tcp: TcpStream,
+  rd_shut: Shut,
+  wr_shut: Shut,
+}
+
+impl TlsStream {
+  fn new<S>(tcp: TcpStream, tls: S) -> Self
```
piscisaureus, line 141: Type parameter is no longer necessary, will remove it.

runtime/ops/tls.rs
```diff
@@ -127,28 +561,28 @@ async fn op_start_tls(
...
-    config.root_store.add_pem_file(reader).unwrap();
+    tls_config.root_store.add_pem_file(reader).unwrap();
   }
+  let tls_config = Arc::new(tls_config);
 
-  let tls_connector = TlsConnector::from(Arc::new(config));
-  let dnsname = DNSNameRef::try_from_ascii_str(&domain)
-    .map_err(|_| generic_error("Invalid DNS lookup"))?;
-  let tls_stream = tls_connector.connect(dnsname, tcp_stream).await?;
+  let tls_stream = TlsStream::new_client(tcp_stream, &tls_config, hostname_dns);
```
bnoordhuis, line 577: The diff up til here is all purely stylistic changes? If not, I'm not sure what the changes are.

(And if it is, now you know why I hate style changes.)

bnoordhuis (approved):

runtime/ops/tls.rs
piscisaureus: Mostly stylistic. Other things that are worth mentioning:
* The error message returned when `DNSNameRef::try_from_ascii()` fails was made more sensible.
* I tried make the style to be consistent with `op_connect_tls()`. ]
* In `op_connect_tls()` I move the empty string -> "localhost" so it happens *before* the permissions check instead of after. 
* To state the obvious: it no longer uses tokio_rustls.

runtime/ops/tls.rs
```diff
@@ -70,6 +85,421 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+  ) -> Poll<io::Result<usize>> {
+    ready!(self.poll_io(cx, false, true))?;
+
+    match self.tls.write(buf) {
+      Ok(0) => unreachable!(),
+      Ok(bytes_written) => {
+        // We try to flush as much as we can, but since we have already handed
+        // off some bytes to rustls we can't return `Poll::Pending()` any more,
+        // as this would indicate to the caller that it should try again.
+        let _ = self.poll_io(cx, false, true)?;
```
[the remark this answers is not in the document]
piscisaureus: Yeah the number of bytes written gets discarded in that case; I don't think this is an issue because if it returns an error the socket is dead.

That said, with the right tactics I could be convinced to get rid of this `poll_io()` call entirely.

runtime/ops/tls.rs
```diff
@@ -70,6 +85,421 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+  fn poll_read(
+    mut self: Pin<&mut Self>,
+    cx: &mut Context<'_>,
+    buf: &mut ReadBuf<'_>,
+  ) -> Poll<io::Result<()>> {
+    ready!(self.poll_io(cx, true, false))?;
+
+    let buf_slice = unsafe {
+      &mut *(buf.unfilled_mut() as *mut [MaybeUninit<u8>] as *mut [u8])
+    };
```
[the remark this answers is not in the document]
piscisaureus: You got it - I copied this from Tokio. 

I also think (or at least used to think) that `transmute()` really isn't as evil as it is made out to be. 
This position seems to be a bit of a heresy in Rust land though...

runtime/ops/tls.rs
```diff
@@ -70,6 +85,421 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+        && (!self.tls.wants_read() || self.rd_shut == Shut::TcpShut)
+      {
+        return Poll::Ready(Ok(()));
+      }
+      if for_writing && !self.tls.is_handshaking() && !self.tls.wants_write() {
+        return Poll::Ready(Ok(()));
+      }
+
+      return Poll::Pending;
+    }
```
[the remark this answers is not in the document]
piscisaureus: It won't starve because:
* `wants_read()` will return `false` when there rustls has decrypted plaintext waiting for the user to pick up.
  No plaintext is read in this loop, so there's an upper limit to the amount of reading that happens each round.
* `wants_write()` returns true only as long as it has unsent, encrypted ciphertext in its internal buffer.
  No plaintext is written in this loop, so the upper bound for writes is defined by plaintext that the user had already sent before we got here.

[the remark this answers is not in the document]
piscisaureus: https://github.com/ctz/rustls/blob/b9a5a061b48598545168da41f759c2a191ad4748/rustls/src/server/mod.rs#L548-L556

runtime/ops/tls.rs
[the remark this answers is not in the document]
piscisaureus: I gave it an honest try, but clippy isn't happy:

```
transmute from a reference to a reference
`#[deny(clippy::transmute_ptr_to_ptr)]` implied by `#[deny(warnings)]`
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_ptr_to_ptr
```

runtime/ops/io.rs
```diff
@@ -340,25 +324,12 @@ impl Resource for TcpStreamResource {
...
 }
 
-pub type TlsClientStreamResource =
-  FullDuplexSplitResource<tls::client::TlsStream<TcpStream>>;
+pub type TlsStreamResource =
+  FullDuplexResource<tls::TlsStreamReader, tls::TlsStreamWriter>;
 
-impl Resource for TlsClientStreamResource {
+impl Resource for TlsStreamResource {
   fn name(&self) -> Cow<str> {
```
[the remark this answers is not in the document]
piscisaureus: Good point actually. But I think changing that belongs in a separate commit.

runtime/ops/tls.rs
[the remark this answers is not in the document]
bnoordhuis: Okay, that makes sense but I don't think it's obvious to a casual (or even not-so-casual) reader. Can you add a comment or two?

bnoordhuis (approved):

ry (commented):
Can tokio-rustls be removed from cli/Cargo.toml and test_util/Cargo.toml now?

quininer:
> tokio-rustls streams misbehave when split with tokio::io::split() into a "read half" and a "write half".

Hi, can you describe the problem? I am very interested in it.

runtime/ops/tls.rs
```diff
@@ -70,6 +88,457 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+  fn into_split(self) -> (ReadHalf, WriteHalf) {
+    let shared = Shared::new(self);
+    let rd = ReadHalf {
+      shared: shared.clone(),
+    };
+    let wr = WriteHalf { shared };
+    (rd, wr)
+  }
+
+  fn poll_io(
```
quininer, line 183: Please note that it is dangerous to use the same `poll_io` for reading and writing. see https://github.com/tokio-rs/tls/issues/40

runtime/ops/tls.rs
```diff
@@ -70,6 +88,457 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+impl AsyncWrite for TlsStream {
+  fn poll_write(
+    mut self: Pin<&mut Self>,
+    cx: &mut Context<'_>,
+    buf: &[u8],
+  ) -> Poll<io::Result<usize>> {
+    ready!(self.poll_io(cx, Flow::Write))?;
+
+    match self.tls.write(buf) {
+      Ok(0) => unreachable!(),
```
quininer, line 320: It may return 0 when the internal buffer of rustls is full.

runtime/ops/tls.rs
```diff
@@ -70,6 +88,457 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+
+    if matches!(self.rd_shut, Shut::TlsShut | Shut::TcpShut) {
+      return Poll::Ready(Ok(()));
+    }
+
+    let buf_slice =
+      unsafe { &mut *(buf.unfilled_mut() as *mut [MaybeUninit<u8>] as *mut _) };
+
+    match self.tls.read(buf_slice) {
+      Ok(0) => unreachable!(),
```
quininer, line 300: I think it will return 0 at eof.

runtime/ops/tls.rs
```diff
@@ -70,6 +88,457 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+
+impl AsyncRead for TlsStream {
+  fn poll_read(
+    mut self: Pin<&mut Self>,
+    cx: &mut Context<'_>,
+    buf: &mut ReadBuf<'_>,
+  ) -> Poll<io::Result<()>> {
+    ready!(self.poll_io(cx, Flow::Read))?;
+
+    if matches!(self.rd_shut, Shut::TlsShut | Shut::TcpShut) {
```
quininer, line 292: I think it obscures the shutdown status of tcp stream and tls stream. When eof occurs in the tcp stream, there may still be readable data in the buffer of the tls stream.

runtime/ops/tls.rs
```diff
@@ -70,6 +88,457 @@ impl StoresClientSessions for ClientSessionMemoryCache {
...
+            // Do a zero-length plaintext read so we can detect the arrival of
+            // 'CloseNotify' messages, even if only the write half is active.
+            // Actually reading data from the socket is done in `poll_read()`.
+            Ok(()) => match self.tls.read(&mut []) {
+              Ok(0) => {}
+              Ok(_) => unreachable!(),
+              // Rustls `read()` returns `ConnectionAborted` when it receives a
+              // 'CloseNotify' alert; this indicates that the remote end of the
+              // connection has initiated a graceful shutdown at the TLS level.
+              Err(err) if err.kind() == ErrorKind::ConnectionAborted => {
```
quininer, line 239: We cannot depend on `ConnectionAborted` to determine eof, because tls stream may not receive close notify.

runtime/ops/tls.rs
piscisaureus: In what way is it dangerous? Note that tokio-rs/tls#40 is the reason we're migrating away from tokio-rustls.

A special `Waker` implementation (see https://github.com/denoland/deno/blob/43beea27809e7a29560dd29ecbd18ca38a0fa133/runtime/ops/tls.rs#L445-L461) makes it safe to read the underlying stream when writing, as well as writing to the underlying stream when reading.

quininer: You may have misunderstood something, `tokio-rustls` does not have the problem described in https://github.com/tokio-rs/tls/issues/40, because `tokio-rustls` does not perform reverse IO, you can see it in [the comment](https://github.com/tokio-rs/tls/issues/40#issuecomment-735037105).

Shared looks interesting, but I suspect it has performance issues because it may cause unnecessary wakeups. and since `rd_waker` and `wr_waker` always wake up at same time, it may cause `tls_stream` to always have lock contention.

runtime/ops/tls.rs
piscisaureus: That shouldn't happen, because `poll_io()` return `Poll::Pending` when the buffer is full.
However it will return 0 when `buf` is empty, so I removed the assertion anyway.

quininer: No, it will not return to pending. see https://github.com/ctz/rustls/blob/47d309233f2d21ccf058d35c58d684e5e4c9c643/rustls/src/server/mod.rs#L738

piscisaureus: > 
No, it will not return to pending. see https://github.com/ctz/rustls/blob/47d309233f2d21ccf058d35c58d684e5e4c9c643/rustls/src/server/mod.rs#L738

`poll_io()` returns `Poll::Pending` when the buffer is full: https://github.com/denoland/deno/blob/800ae5ee1f60f30649f09e9a2b53599afc663760/runtime/ops/tls.rs#L325

quininer: Sorry, I didn't see to `poll_io`, you are right.

runtime/ops/http.rs
```diff
@@ -280,12 +280,13 @@ fn op_http_start(
...
   if let Some(resource_rc) = state
     .resource_table
-    .take::<TlsServerStreamResource>(tcp_stream_rid)
+    .take::<TlsStreamResource>(tcp_stream_rid)
   {
-    let resource = Rc::try_unwrap(resource_rc)
-      .expect("Only a single use of this resource should happen");
+    let resource = Rc::try_unwrap(resource_rc).unwrap_or_else(|_| {
+      panic!("Only a single use of this resource should happen")
+    });
```
ry, line 287: Does this change have a purpose? It looks like the same thing to me...

op_crates/websocket/lib.rs
```diff
@@ -161,7 +161,7 @@ where
 
       let tls_connector = TlsConnector::from(Arc::new(config));
       let dnsname =
-        DNSNameRef::try_from_ascii_str(&domain).expect("Invalid DNS lookup");
+        DNSNameRef::try_from_ascii_str(&domain).expect("invalid hostname");
```
ry, line 164: is this related to tokio-rustls?

cli/tests/unit/tls_test.ts
```diff
@@ -32,7 +32,7 @@ unitTest(
 
     await assertThrowsAsync(async () => {
       await Deno.connectTls({ hostname: "127.0.0.1", port: 3567 });
-    }, Error);
+    }, TypeError);
```
ry, line 35: related?

cli/tests/listen_tls_alpn.ts
```diff
@@ -8,5 +10,8 @@ const listener = Deno.listenTls({
 console.log("READY");
 
 for await (const conn of listener) {
+  if (await conn.read(new Uint8Array(1)) !== null) {
+    throw new Error("should not receive data on TLS connection");
+  }
```
ry, line 15: Is this demonstrating the bug this PR is fixing?

cli/tests/integration_tests.rs
```diff
@@ -5969,4 +5972,7 @@ async fn listen_tls_alpn_fail() {
   let (_, session) = tls_stream.get_ref();
 
   assert!(session.get_alpn_protocol().is_none());
+
+  child.kill().unwrap();
+  child.wait().unwrap();
```
ry, line 5977: Is this demonstrating the bug?

cli/tests/integration_tests.rs
```diff
@@ -5930,11 +5930,14 @@ async fn listen_tls_alpn() {
 
   let alpn = session.get_alpn_protocol().unwrap();
   assert_eq!(std::str::from_utf8(alpn).unwrap(), "foobar");
+
+  child.kill().unwrap();
+  child.wait().unwrap();
```
ry, line 5935: ditto - Is this demonstrating the bug?

runtime/ops/tls.rs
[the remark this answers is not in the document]
piscisaureus: Done!

cli/tests/integration_tests.rs
piscisaureus: It prevents the test from leaking a child process. Although this only happens when it fails (otherwise it exits by itself), the  leaked process subsequently squats a TCP listener port and makes a bunch of other tests fail, which makes it hard to debug.

runtime/ops/http.rs
piscisaureus: `Result::<T, E>::unwrap()` requires `E` to implement `Debug`.
It's un-done, I've added `Debug` impls now.

runtime/ops/tls.rs
piscisaureus: I agree that in lock contention could be an issue when using the multi-threaded runtime.
However, Deno is using the single threaded runtime, so that's not going to be a problem for us.

I necessary, I could imagine using a different scheme (for example, we could daisy chain the wake-ups: one task wakes up another waker when polled). But I don't think we need that extra complexity right now.

runtime/ops/tls.rs
piscisaureus: 13 lines up graceful close of the underlying TCP connection gets handled. https://github.com/denoland/deno/blob/6cd7b48744db8718b243dfff8afb676a039d916e/runtime/ops/tls.rs#L275

runtime/ops/tls.rs
quininer: Fair enough, I don't care what implementation deno uses. I am glad to know that `tokio-rustls` has no problems.

I hope you can edit the description of PR to avoid unnecessary misunderstandings.

runtime/ops/tls.rs
quininer: Tcp shutdown does not mean Tls shutdown, I mentioned this in another comment.

quininer: Well, I reconsidered it. If `wants_read` guarantees that only one packet is processed in rustls, then it is fine.

piscisaureus:
> Can tokio-rustls be removed from cli/Cargo.toml and test_util/Cargo.toml now?

From CLI, yes.

But `test_util` and `op_crates/deno_websocket` still use tokio-rustls. It would be possible to switch them, but in order to do that we would have to move the `TlsStream` implementation to a separate crate, since neither `test_util` nor `deno_websocket` depend on `deno_runtime`, and it seems best to keep it that way.

Note that `deno_websocket` inherits the dependency from `tokio-tungstenite`, which uses a [very similar mitigation](https://github.com/snapview/tokio-tungstenite/blob/287f477e5491296ed936aa075947740ea5d6ba7b/src/compat.rs#L118-L142) for what appears to be [the same problem](https://github.com/snapview/tokio-tungstenite/blob/287f477e5491296ed936aa075947740ea5d6ba7b/src/compat.rs#L21-L42).

ry:
@piscisaureus What's the status of this PR?

PR #10150, opened 2021-04-12
perf: lazy header instantiation for HTTP requests

lucacasonato:
This commit introduces a performance optimization for the native HTTP
server. From my testing it is about 2-6% faster than `main`. Request
headers in the HTTP servers are now lazilly instatated when they are
accessed, rather than being preemptively wrapped in the `Headers` class.


ry (approved):
LGTM

PR #10151, opened 2021-04-12
perf: make bench script more efficient

lucacasonato:
This only creates one response that is used to respond to all incoming
requests in the benchmark. This would break for streaming responses, but
this is not a streaming response, so who gives :shrug:.

### DIFF
--- a/cli/bench/deno_http_native.js
+++ b/cli/bench/deno_http_native.js
@@ -6,12 +6,13 @@ const listener = Deno.listen({ hostname, port: Number(port) });
 console.log("Server listening on", addr);
 
 const body = Deno.core.encode("Hello World");
+const response = new Response(body);
 
 for await (const conn of listener) {
   (async () => {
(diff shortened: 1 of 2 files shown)


lucacasonato:
It is questionable if we should do this, or just improve the time it takes to construct a `Response`. I think this is fair though, as this benchmark is trying to test the upper bound of our HTTP server, not the average usage. For that we should add a second test case for more realistic usecase which reads/sets headers, has a streaming response which is different per request, and a request body. We can then add the same bench for `node` to compare.

AaronO:
Whilst I think this is somewhat fair, I think it's best to leave inefficiencies visible in our benches.

Having those consistently tracked will make them easier to profile & optimize.

ry:
I'd prefer to leave it as is, since in almost every case people will be creating a response object.

lucacasonato:
Closing then :-)

PR #10155, opened 2021-04-12
perf(fetch): remove method normalization

ry:


lucacasonato (changes requested):
See https://github.com/denoland/deno/pull/10154#issuecomment-818319730

ry:
ok closed in favor of #10154

PR #10156, opened 2021-04-13
fix(serde_v8): remove intentional deserialization error on non-utf8 strings

### LINKED ISSUE #9952: Invalid UTF-8 char in typescript file causes serde_v8 error
If a typescript file contains an invalid utf-8 string, that causes `serde_v8 error: ExpectedUtf8`.

The command below reproduces that error.
```
deno run https://raw.githubusercontent.com/denoland/deno_std/d42b5d141cf637d40e0430d065eab91cb66b7e2b/fmt/printf_test.ts
```

(The above file (test case of printf in deno_std) contains invalid utf-8 char in line 274)

ref: https://github.com/denoland/deno_std/pull/828

---
```
$ deno -V                                                                                                                  
deno 1.8.2+fec1b2a
```

### LINKED ISSUE #10217: Panic on uncaught error with invalid utf8 string
This will cause rust to panic in Deno 1.9.0 if run as a script:

```js
JSON.parse(new TextDecoder().decode(new Uint8Array([13, 137])))
```

The crash log with backtrace:

```
$ RUST_BACKTRACE=1 deno run crash.js 
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ExpectedUtf8', core/error.rs:190:52
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/option.rs:1300:5
   3: deno_core::error::JsError::from_v8_exception
   4: deno_core::runtime::exception_to_err_result
   5: deno_core::runtime::JsRuntime::poll_event_loop
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   8: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  10: deno::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

My guess is that a bug in the TextDecoder creates an invalid string, which then triggers a v8 exception on JSON.parse.

AaronO:
Would be good to land this for 1.9, since it fixes #9952, fixes #10217

## Changes

- [x] Remove non-utf8 check in `serde_v8`'s `deserialize_string()`
- [x] Remove no-longer expected failures in WPT


piscisaureus:
I'm not convinced it is desirable to silently "handle" string encoding errors in the op layer.
It could lead to really hard to debug errors, e.g. if you have a file name that is not valid UTF8 (which can happen), the op layer would silently change the file name to a different one and write the wrong file.

AaronO:
> I'm not convinced it is desirable to silently "handle" string encoding errors in the op layer.
> It could lead to really hard to debug errors, e.g. if you use `op_open` and the file name is not valid UTF8 (which can happen), the op layer would silently change the file name to a different one and write the wrong file.

I should have added more context. There are WPT tests such as url parsing where Deno fails and Chrome/Node succeed, Deno had the same output as Node/Chrome in the first iteration of `serde_v8` which did not have this failure.

So this failure was added to added to align somewhat with Deno's previous behaviour. I don't think it's unreasonable to use utf8 replacement characters, not using them causes "surprising" exceptions as in the linked issue.

I think down the line as we extend `serde_v8` (with `ZeroCopyBuf`, etc...) we could add a `StrictUtf8String` type that errors on deserialization instead of using unicode replacement characters

ry:
Are there WPT tests that can be enabled? 

IIRC this one was caused by serde_v8:
https://github.com/denoland/deno/blob/df49a8462caf1cf4eea5c9f386322dae5e14dc4f/tools/wpt/expectation.json#L549

piscisaureus:
> I should have added more context. There are WPT tests such as url parsing where Deno fails and Chrome/Node succeed, Deno had the same output as Node/Chrome in the first iteration of `serde_v8` which did not have this failure.

I'm all for fixing WPT tests, but can't we do this "utf8 repair" at the level of individual ops?

> So this failure was added to added to align somewhat with Deno's previous behaviour. I don't think it's unreasonable to use utf8 replacement characters, not using them causes "surprising" exceptions as in the linked issue.

It depends what area you focus on. With deno 1.8.3:
```
piscisaureus@spirit:~/deno$ deno run this_file_has_invalid_utf8.js
error: invalid data
```

satyarohith:
The changes of this PR will also fix https://github.com/denoland/deno/issues/10217

bartlomieju (approved):
LGTM, thanks!

PR #10158, opened 2021-04-13
fix(core): remove wasm-test-streaming flag

### LINKED ISSUE #7259: WebAssembly.instantiateStreaming does not work with Response

caspervonb:
Temporarily resolves #7259 until we properly support streaming instantiation.


lucacasonato (approved):

ry (approved):
Thanks Casper - LGTM

PR #10162, opened 2021-04-13
feat(cli): raise file descriptor limit on startup

### LINKED ISSUE #10148: Too many open files is fatal for Listener.accept
if you run wrk against either native http or std http bench examples with a large connection number:

```sh
# one terminal
❱ target/release/deno run -A --unstable cli/bench/deno_http_native.js
Server listening on 127.0.0.1:4500
error: Uncaught (in promise) Error: Too many open files (os error 24)
    at async Listener.accept (deno:runtime/js/30_net.js:98:19)
    at async Listener.next (deno:runtime/js/30_net.js:105:16)
    at async file:///Users/somewhere/deno/cli/bench/deno_http_native.js:10:18


# another (note the large number of connections)
❱ wrk -d 20 -c 500 --latency http://127.0.0.1:4500
```
This can probably be worked around with increasing ulimits. But this could conceivably happen unexpectedly on a server if it had a spike in traffic.

Note: `node cli/bench/node_http.js` handles this without error (or printing anything).

---

Hyper's AddrIncoming (higher level API) has [`sleep_on_error`](https://docs.rs/hyper/0.14.5/hyper/server/conn/struct.AddrIncoming.html#method.set_sleep_on_errors) which sleeps for a second after an non connection error:
https://github.com/hyperium/hyper/blob/41f99578a53845e5e0bb999c101bef8307e1ce5f/src/server/tcp.rs#L130-L145

Deno should do that as well.

bnoordhuis:
Raise the soft limit to the hard limit when possible. This is similar
to what Node.js does to avoid running into "out of file descriptors"
errors too quickly.

On most Linux systems, raises the limit from 1,024 to 1,048,576.
On most macOS systems, raises the limit from 256 to 10,240.

Fixes #10148.

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -36,6 +36,7 @@ mod tokio_util;
 mod tools;
 mod tsc;
 mod tsc_config;
+mod unix_util;
 mod version;
 
 use crate::file_fetcher::File;
@@ -1176,6 +1177,7 @@ fn unwrap_or_exit<T>(result: Result<T, AnyError>) -> T {
 pub fn main() {
   #[cfg(windows)]
   colors::enable_ansi(); // For Windows 10
+  unix_util::raise_fd_limit();
 
   let args: Vec<String> = env::args().collect();
   let standalone_res = match standalone::extract_standalone(args.clone()) {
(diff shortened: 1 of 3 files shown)


ry (approved):
LGTM

PR #10164, opened 2021-04-13
chore: release crates

bartlomieju:

### DIFF
--- a/op_crates/file/Cargo.toml
+++ b/op_crates/file/Cargo.toml
@@ -2,7 +2,7 @@
 
 [package]
 name = "deno_file"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2018"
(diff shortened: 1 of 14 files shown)


ry (approved):

op_crates/file/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 
 [package]
 name = "deno_file"
-version = "0.1.0"
```
lucacasonato, line 5: We never released the 0.1.0 of this, but doesn't matter. @bartlomieju can you add us on crates.io?

bartlomieju: Done

lucacasonato: Thanks :-)

PR #10179, opened 2021-04-14
refactor: move timers to deno_timers op crate

bnoordhuis:
Move timers out of runtime/ and into a standalone op crate.

<hr>

The diff looks biggish but it's mostly moving code around.


ry (approved):
LGTM 

published at https://crates.io/crates/deno_timers

PR #10180, opened 2021-04-14
feat(lsp): support formatting json and markdown files

### LINKED ISSUE #9415: lsp should support formatting md files
We support `deno fmt README.md` since Deno 1.7. We should also enable the LSP to format MD files.

### LINKED ISSUE #9447: lsp should support formatting json files
In addition to #9415, json formatting would also be welcomed #9292 #9268.

satyarohith:
Closes #9447
Closes #9415

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -218,6 +218,17 @@ impl<'a> From<&'a diagnostics::Position> for lsp::Position {
   }
 }
 
+/// Check if diagnostics can be generated for the provided media type.
+pub fn is_diagnosable(media_type: MediaType) -> bool {
+  matches!(
+    media_type,
+    MediaType::TypeScript
+      | MediaType::JavaScript
+      | MediaType::Tsx
+      | MediaType::Jsx
+  )
+}
+
 fn get_diagnostic_message(diagnostic: &diagnostics::Diagnostic) -> String {
   if let Some(message) = diagnostic.message_text.clone() {
     message
@@ -312,8 +323,8 @@ async fn generate_lint_diagnostics(
           .lock()
           .await
           .get_version(specifier, &DiagnosticSource::DenoLint);
-        if version != current_version {
-          let media_type = MediaType::from(specifier);
+        let media_type = MediaType::from(specifier);
+        if version != current_version && is_diagnosable(media_type) {
           if let Ok(Some(source_code)) = documents.content(specifier) {
             if let Ok(references) = analysis::get_lint_references(
               specifier,
@@ -354,10 +365,11 @@ async fn generate_ts_diagnostics(
         let version = snapshot.documents.version(s);
         let current_version =
           collection.get_version(s, &DiagnosticSource::TypeScript);
-        if version == current_version {
-          None
-        } else {
+        let media_type = MediaType::from(s);
+        if version != current_version && is_diagnosable(media_type) {
           Some(s.clone())
+        } else {
+          None
         }
       })
       .collect()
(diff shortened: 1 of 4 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -58,6 +59,15 @@ async fn publish_diagnostics(
   let maybe_changes = collection.take_changes();
   if let Some(diagnostic_changes) = maybe_changes {
     for specifier in diagnostic_changes {
+      // Skip diagnostics for JSON(c) and markdown files.
```
satyarohith, line 62: Please do suggest if there's a better place to skip diagnostics for these files.

satyarohith:
PR ready for review.

(cc @kitsonk)

bartlomieju:
@satyarohith please rebase

bartlomieju (commented):
LGTM, but I'll wait for @kitsonk's review

kitsonk:
Yeah, I will get to it tomorrow, sorry for the delay.

cli/lsp/diagnostics.rs
kitsonk: I would rather do the opposite (only process diagnostics for file types we know we produce diagnostics for), and use the `MediaType` enum to determine that versus "guessing" at the extension. If we need to add enum variants to deal with it, it is better to have a "authoritative" place to determine the type of file.

kitsonk (commented):
Won't this require a complimentary PR in vscode_deno, so that it knows that we can handle the other file types?

cli/lsp/diagnostics.rs
satyarohith: I moved the logic to the individual `generate_*_diagnostic` functions.  Let me know your thoughts on new changes.

satyarohith:
> Won't this require a complimentary PR in vscode_deno, so that it knows that we can handle the other file types?

I was waiting for this PR to get merged. I opened the PR anyway: https://github.com/denoland/vscode_deno/pull/404

cli/lsp/diagnostics.rs
```diff
@@ -381,8 +392,8 @@ async fn generate_lint_diagnostics(
     for specifier in state_snapshot.documents.open_specifiers() {
       let version = state_snapshot.documents.version(specifier);
       let current_version = collection.get_version(specifier);
-      if version != current_version {
-        let media_type = MediaType::from(specifier);
+      let media_type = MediaType::from(specifier);
```
kitsonk, line 395: This won't work for remote/data/blob files.  We need to get the media type from the document meta data and in other locations, for the sources meta data.

satyarohith: Okay. I need understand a few things. Why do we need diagnostics for remote/data/blob files? Maybe I need to understand what are remote/data/blob files.

kitsonk: Because if they are open in the editor, you want the editor to let you know if there are errors, right?  Even if they are open read only, having diagnostic information for them is pretty critical in understanding what is going on.

satyarohith: Looks like we do not provide diagnostic information for remote files currently. I agree with your statement and I think we should have the feature as it's helpful, but I'm curious if the changes related to it can be done along with the PR that enables diagnostics for the remote files. WDYT?

kitsonk:
Looks like somehow a deadlock was introduced which is causing the integration tests sometimes to fail and the benchmark tests to fail. I will try to look at it later.

kitsonk:
The deadlock was occurring on my local machine.  I merged in #10555 and it fixed the deadlock.

kitsonk:
Sorry this keeps getting delayed... Now that we refactored the lsp tests (which is good for everyone) this needs a merge/rebase with upstream and the tests moved to `cli/tests/integration_tests_lsp.rs`.

kitsonk (approved):
LGTM

Issue #10182, opened 2021-04-14
Native HTTP: Reading a request body but not awaiting `respondWith()` causes the response to fail


kitsonk:
Given the following [code](https://gist.github.com/kitsonk/4aa92a8a78bbb303a1ef576cdeb4e9a3):

```ts
const decoder = new TextDecoder();

async function handleRequest({ request, respondWith }: Deno.RequestEvent) {
  let resolve: (response: Response) => void;
  const p = new Promise<Response>((res) => {
    resolve = res;
  });
  const r = respondWith(p);
  let response: Response;
  if (request.body) {
    const body = decoder.decode(await request.arrayBuffer());
    console.log("body:", body);
    response = new Response("body\n");
  } else {
    response = new Response("no body\n");
  }
  resolve!(response);
  await r;
  console.log("done");
}

for await (const conn of Deno.listen({ port: 8000 })) {
  console.log("conn");
  (async () => {
    const httpConn = Deno.serveHttp(conn);
    for await (const requestEvent of httpConn) {
      console.log("request");
      handleRequest(requestEvent);
    }
  })();
}
```

And running the server like:

```
> deno run --allow-net --unstable server.ts
```

And then sending a request like (or any request with a body):

```
> curl --request POST \
  --url http://127.0.0.1:8000/ \
  --header 'Content-Type: application/json' \
  --data '{ "a": 1 }'
```

Will cause the server to hang and fail to send the response.  While the following would process normally:

```
> curl http://127.0.0.1:8000/
```

kitsonk:
It appears that if I await the promise returned by the `respondWith()` before awaiting the next request (be it via an iterator on `httpConn` or `httpConn.nextRequest()`) then the body is returned to the client and the requests complete.

Here is an example that works under main:

```ts
const decoder = new TextDecoder();

async function handleRequest({ request, respondWith }: Deno.RequestEvent) {
  console.log("handleRequest");
  let resolve: (response: Response) => void;
  const p = new Promise<Response>((res) => {
    resolve = res;
  });
  const r = respondWith(p);
  let response: Response;
  if (request.body) {
    const body = decoder.decode(await request.arrayBuffer());
    console.log("body:", body);
    response = new Response("body\n");
  } else {
    response = new Response("no body\n");
  }
  resolve!(response);
  console.log(" request done");
  return r;
}

const listener = Deno.listen({ port: 8000 });

for await (const conn of listener) {
  console.log("conn");
[...]

PR #10184, opened 2021-04-14
chore: add failing test case for serveHttp

kitsonk:
This test causes causes the same failure mentioned in #10182


kitsonk:
A test case was added in #10197, this is no longer needed.

Issue #9859, opened 2021-03-22
Add Permissions to TestDefinition


caspervonb:
Would be nice to be able to specify the permissions a test should run in.

```js
Deno.test({
  name: "foo",
  fn() {
    assertThrows(function() {
      Deno.open("...");
    });
  },
  permissions: {
    write: false,
  },
});
```

To avoid breaking the sandboxing rules, this would **not** be able to request *more* permissions than the initial set that was given to `deno test` but can reset the permission set to that before every test is run.

PR #10188, opened 2021-04-14
feat(cli): add test permissions to Deno.test

caspervonb:
This adds adds permissions to the test definitions which allows tests to run with different permission sets than the process's permission.

The change will only be in effect within the test function, once the test has completed the original process permission set is restored.

Test permissions cannot exceed the process's permission.
You can only narrow or drop permissions, failure to acquire a permission results in an error being thrown and the test case will fail.

Closes https://github.com/denoland/deno/issues/9859


cli/tests/test/allow_all.ts
```diff
@@ -0,0 +1,44 @@
+import { assertEquals } from "../../../test_util/std/testing/asserts.ts";
+
+// TODO(DO NOT MERGE) test all the things
```
caspervonb, line 3: **Note**: tests aren't done

cli/dts/lib.deno.ns.d.ts
```diff
@@ -94,6 +94,16 @@ declare namespace Deno {
    * See: https://no-color.org/ */
   export const noColor: boolean;
 
+  export interface TestPermissions {
+    read?: string[] | boolean;
```
caspervonb, line 98: Not too sure about this overload, primarily added it because unitTest has the boolean variant and I want to be able to specify specific directories in addition to just a blanket "allow write".

bartlomieju: Sounds good to me!

cli/main.rs
```diff
@@ -218,6 +218,7 @@ pub fn create_main_worker(
     // above
     ops::errors::init(js_runtime);
     ops::runtime_compiler::init(js_runtime);
+    ops::test_runner::init(js_runtime);
```
bartlomieju, line 221: These ops should only be added if we're in `deno test` - I guess adding a param to this callback is sufficient

cli/ops/test_runner.rs
```diff
@@ -0,0 +1,234 @@
...
+use deno_runtime::permissions::PermissionState;
+use deno_runtime::permissions::Permissions;
+use deno_runtime::permissions::PermissionsOptions;
+use serde::Deserialize;
+use std::path::PathBuf;
+
+pub fn init(rt: &mut deno_core::JsRuntime) {
+  super::reg_sync(
+    rt,
+    "op_request_test_permissions",
```
bartlomieju, line 20: s/request/pledge/

cli/dts/lib.deno.ns.d.ts
caspervonb: Actually, should consider if there's something to be shared here between these permissions and worker permissions.

piscisaureus:
@caspervonb Whatever happens - thanks for you contribution, we appreciate the time & effort you're putting into Deno!

That said, I don't really understand why this change is necessary. Maybe I'm just ignorant. However, can you give me the layman's explanation?

caspervonb:
Hey @piscisaureus.

This lets us set up the permissions for a specific test.

We can't do this with Deno.permissions.revoke because once a permission is revoked you can't go back so there's no way to set up a test case with a certain set of permissions.

The question of "how do i run a test case with x, y and z" permissions does comes up every now and then altho not that frequently.

We set up these pre conditions internally in our unit tests by not using Deno.test or even the test command at all so this simplifies the whole ordeal quite a bit and  allows us to drop the whole unit test runner and just dogfood `Deno.test`.

The existence of the unit test runner is causing us to leak a bunch of internal details about the test runner as well.

TL;DR its needed to setup the preconditions for a test that expects to have X permissions.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1181,6 +1181,35 @@ declare namespace Deno {
...
+   * Use the peermissions options to change the level of access the test will
+   * have. By default it will inherit the permissions of its parent thread. The permissions
+   * of a test can't be extended beyond its parent's permissions reach.
+   * - "inherit" will take the permissions of the thread the test is created in
+   * - You can disable/enable permissions all together by passing a boolean
+   * - You can provide a list permissions to narrow the permissions a test has.
+  */
+  export interface TestDefinition {
+    /** Set to `"none"` to disable all the permissions in the test. */
+    permissions?: "inherit" | "none" | {
```
piscisaureus, line 1196: I think the "inherit" variant is redundant, because `undefined` means the same thing.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1181,6 +1181,35 @@ declare namespace Deno {
...
+      /** The format of the net access list must be `hostname[:port]`
+       * in order to be resolved.
+       *
+       * ```
+       * net: ["https://deno.land", "localhost:8080"],
+       * ```
+       * */
+      net?: "inherit" | boolean | string[];
+      plugin?: "inherit" | boolean;
+      read?: "inherit" | boolean | Array<string | URL>;
```
piscisaureus, line 1208: I think there might be some redundant variants. See DM.

cli/dts/lib.deno.unstable.d.ts
piscisaureus: I see that `WorkerOptions.permissions` has the same redundancy, so I suppose it's is not getting worse.

cli/ops/test_runner.rs
```diff
@@ -0,0 +1,64 @@
...
+
+  if let Some(permissions_holder) = state.try_take::<PermissionsHolder>() {
+    let token: Uuid = serde_json::from_value(args)?;
+    if token != permissions_holder.0 {
+      panic!("restore test permissions token does not match the stored token");
+    }
+
+    let permissions = permissions_holder.1;
+    state.put::<Permissions>(permissions);
+  }
```
piscisaureus, line 61: Should it throw an error if there were no permissions to restore?

cli/dts/lib.deno.unstable.d.ts
piscisaureus: Given that the set of permissions can only be narrowed down, `true`, `"inherit"`, `undefined` all mean the same thing.

cli/dts/lib.deno.unstable.d.ts
caspervonb: Initially this was its own much simpler thing with just `boolean | []string` but noticed we had permissions in workers so adopted from the permission options workers (also re-using the same parser).

I think the options need to be refined but the intention is to keep the two sets of options the same and aligned (test permissions being aligned to worker permissions).

Changing the definition of worker permissions would be a breaking change so I want to punt that discussion down the line a bit.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1181,6 +1181,35 @@ declare namespace Deno {
    * then the underlying HttpConn resource is closed automatically.
    */
   export function serveHttp(conn: Conn): HttpConn;
+
+  /** **UNSTABLE**: New option, permissions are yet to be vetted.
+   *
+   * Use the peermissions options to change the level of access the test will
+   * have. By default it will inherit the permissions of its parent thread. The permissions
+   * of a test can't be extended beyond its parent's permissions reach.
+   * - "inherit" will take the permissions of the thread the test is created in
```
piscisaureus, line 1190: This explanation is not at all clear - is it applicable to the the top level `permissions` property, or is it about specific permission flags e.g. `env: true`.

Maybe add examples instead.

cli/dts/lib.deno.unstable.d.ts
lucacasonato: > Changing the definition of worker permissions would be a breaking change so I want to punt that discussion down the line a bit.

They are unstable, are they not?

cli/ops/test_runner.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#[derive(Clone)]
+struct PermissionsHolder(Uuid, Permissions);
+
+pub fn op_pledge_test_permissions(
+  state: &mut OpState,
+  args: Value,
+  _zero_copy: Option<ZeroCopyBuf>,
+) -> Result<Value, AnyError> {
```
lucacasonato, line 30: Return the `Uuid` directly, instead of using `serde_json::Value`. Using `serde_json::Value` effectively bypasses much optimization in `serde_v8`.

cli/ops/test_runner.rs
```diff
@@ -0,0 +1,64 @@
...
+
+  state.put::<PermissionsHolder>(PermissionsHolder(token, parent_permissions));
+  state.put::<Permissions>(worker_permissions);
+
+  Ok(json!(token))
+}
+
+pub fn op_restore_test_permissions(
+  state: &mut OpState,
+  args: Value,
```
lucacasonato, line 48: Ditto. Use `Uuid` directly.

cli/ops/test_runner.rs
```diff
@@ -0,0 +1,65 @@
...
+  deno_runtime::ops::check_unstable(state, "Deno.test.permissions");
+
+  if let Some(permissions_holder) = state.try_take::<PermissionsHolder>() {
+    if token != permissions_holder.0 {
+      panic!("restore test permissions token does not match the stored token");
+    }
+
+    let permissions = permissions_holder.1;
+    state.put::<Permissions>(permissions);
+    Ok(json!({}))
```
bartlomieju, line 61: Nitpick: it could be just `Ok(())`

cli/ops/test_runner.rs
```diff
@@ -0,0 +1,65 @@
...
+
+  let token = Uuid::new_v4();
+  let parent_permissions = state.borrow::<Permissions>().clone();
+  let worker_permissions = {
+    let permissions: PermissionsArg = serde_json::from_value(args)?;
+    create_worker_permissions(parent_permissions.clone(), permissions)?
+  };
+
+  state.put::<PermissionsHolder>(PermissionsHolder(token, parent_permissions));
+  state.put::<Permissions>(worker_permissions);
```
bartlomieju, line 42: Please add a comment here:
```
// NOTE: This call overrides current permission set for the worker
```

bartlomieju (commented):
Cleanly implemented. I think it's ready to land after addressing Bert's question regarding JSdoc. Nice work @caspervonb

cli/dts/lib.deno.unstable.d.ts
caspervonb: Yeah lazily copied from the worker permissions section.

piscisaureus (approved):
LGTM

bartlomieju (approved):
LGTM 🚀

PR #10190, opened 2021-04-14
remove --unstable flag from CLI features

bartlomieju:


bnoordhuis (approved):

nayeemrmn:
These features are still unstable / can have breaking changes, right? If so, they should at least have `UNSTABLE: ` at the beginning of their descriptions.

ry:
Let's wait to land this unit 1.10 - it's too big of a change in a patch release.

> These features are still unstable / can have breaking changes, right? If so, they should at least have UNSTABLE: at the beginning of their descriptions.

I think that's a good idea. For `deno info --json` maybe we could print it to stderr, so piping to a file still works.

ry (approved):
LGTM

Issue #10118, opened 2021-04-10
Use `data-url` crate for data url decoding in module loader & lsp


lucacasonato:
Currently we implement our own data url parsing in the `data_url_media_type` function in the lsp: https://github.com/denoland/deno/blob/1ed95120225eb3d7c9adeadcc3b1f610dbe55b7e/cli/lsp/urls.rs#L52-L61

We also implement a slightly different parser in the module loader: https://github.com/denoland/deno/blob/8d55d8b6be8731d37ccf6a29127b3a91a8319d0b/cli/file_fetcher.rs#L156-L188

Both of these custom parsers are highly spec incompliant and should be replaced with the parser from the `data-url` crate which is compliant with https://fetch.spec.whatwg.org/#data-urls.

satyarohith:
Working on it.

ObiWanRohan:
The `data-url` crate works on strings instead of parsed URLs. 
All usages will also have to be changed

PR #10196, opened 2021-04-15
refactor: use `data-url` crate to process data URLs in lsp & file_fetcher

satyarohith:
Closes https://github.com/denoland/deno/issues/10118


cli/lsp/urls.rs
```diff
@@ -96,7 +87,11 @@ impl LspUrlMap {
         specifier.clone()
       } else {
         let specifier_str = if specifier.scheme() == "data" {
-          let media_type = data_url_media_type(specifier);
+          let data_url = DataUrl::process(specifier.as_str())
+            .map_err(|e| type_error(format!("{:?}", e)))?;
```
satyarohith, line 91: Any suggestions on what error type would be appropriate here?

piscisaureus: `URIError`

satyarohith: Thanks! :)

cli/file_fetcher.rs
```diff
@@ -762,39 +739,6 @@ mod tests {
     assert_eq!(file.source, expected);
   }
 
-  #[test]
-  fn test_get_source_from_data_url() {
```
piscisaureus, line 766: Is there a reason to remove all these tests?
Are there any tests left that check that data URLs work properly?

satyarohith: Yes. The removed tests check if data URLs are correctly processed. Now that we use the `data-url` crate, we no longer need them. And as far as the integration is concerned, the following two cases cover it.
https://github.com/denoland/deno/blob/f64630b8c152d9fb976c0ecdaa8abfff670932d8/cli/lsp/urls.rs#L188 and https://github.com/denoland/deno/blob/53414b4d3e6ef35e39ba8dacf8b406962ac99980/cli/file_fetcher.rs#L1001

piscisaureus (approved):
Thanks @satyarohith, LGTM.

PR #10199, opened 2021-04-15
chore: align `Headers` to spec

lucacasonato:
This commit aligns `Headers` to spec. It also removes the now unused
03_dom_iterable.js file. We now pass all relevant `Headers` WPT. We do
not implement any sort of header filtering, as we are a server side
runtime.

This is likely not the most efficient implementation of `Headers` yet. It is
however spec compliant. Once all the APIs in the `HTTP` hot loop are
correct we can start optimizing them. It is likely that this reduces bench
throughput temporarily.


op_crates/web/00_infra.js
```diff
@@ -8,6 +8,74 @@
...
+        return `\\u${char.charCodeAt(0).toString(16).padStart(4, "0")}-\\u${
+          char.charCodeAt(2).toString(16).padStart(4, "0")
+        }`;
+      } else {
+        throw TypeError("unreachable");
+      }
+    });
+    return matchers.join("");
+  }
+
```
lucacasonato, line 78: This is all just moved from //op_crates/web/01_mimesniff.js

tools/wpt/expectation.json
```diff
@@ -684,6 +684,15 @@
...
         ]
+      },
+      "headers": {
+        "headers-basic.any.js": true,
+        "headers-casing.any.js": true,
+        "headers-combine.any.js": true,
+        "headers-errors.any.js": true,
+        "headers-normalize.any.js": true,
+        "headers-record.any.js": true,
+        "headers-structure.any.js": true
```
lucacasonato, line 695: All relevant tests pass.

lucacasonato:
The `unit_test/fetch.ts` changes are harmless and not technically wrong, but not great. I can not easily resolve this without conflicting with my follow-up PR that rewrites `Body`, and `Request`, so I would vote to keep it as is for now and revert the test changes in the follow up PR.

cli/tests/unit/headers_test.ts
```diff
@@ -372,22 +372,12 @@ unitTest(function headersAppendDuplicateSetCookieKey(): void {
...
-unitTest(function headersSetDuplicateCookieKey(): void {
-  const headers = new Headers([["Set-Cookie", "foo=bar"]]);
-  headers.set("set-Cookie", "foo=baz");
-  headers.set("set-cookie", "bar=qat");
-  const actual = [...headers];
-  assertEquals(actual, [
-    ["set-cookie", "foo=baz"],
-    ["set-cookie", "bar=qat"],
-  ]);
-});
```
lucacasonato, line 389: This test is plain wrong. `headers.set` is not the same as append. It removes existing headers of that name before insertion. See https://fetch.spec.whatwg.org/#concept-header-list-set

ry (approved):
LGTM

ry:
This PR cause a fairly severe performance degradation in the performance of std/http/server.ts

Screen Shot 2021-05-06 at 7 57 20 AM

lucacasonato:
This is most likely cause by the O(n^2) complexity of https://github.com/denoland/deno/blob/main/extensions/fetch/20_headers.js#L178-L207

PR #10203, opened 2021-04-15
chore: align fetch to spec

### LINKED ISSUE #10002: fetch with Request init with body of FormData is broken

lucacasonato:
This commit aligns the `fetch` API and the `Request` / `Response`
classes belonging to it to the spec. This commit enables all the
relevant `fetch` WPT tests. Spec compliance is now at around 90%.

Performance is essentially identical now (within 1% of 1.9).

Closes #10002

### DIFF
--- a/op_crates/fetch/20_headers.js
+++ b/op_crates/fetch/20_headers.js
@@ -14,10 +14,14 @@
 ((window) => {
   const webidl = window.__bootstrap.webidl;
   const {
+    HTTP_TAB_OR_SPACE_PREFIX_RE,
+    HTTP_TAB_OR_SPACE_SUFFIX_RE,
     HTTP_WHITESPACE_PREFIX_RE,
     HTTP_WHITESPACE_SUFFIX_RE,
     HTTP_TOKEN_CODE_POINT_RE,
     byteLowerCase,
+    collectSequenceOfCodepoints,
+    collectHttpQuotedString,
   } = window.__bootstrap.infra;
 
   const _headerList = Symbol("header list");
@@ -35,7 +39,7 @@
    */
 
   /**
-   * @typedef {string} potentialValue 
+   * @param {string} potentialValue 
    * @returns {string}
    */
   function normalizeHeaderValue(potentialValue) {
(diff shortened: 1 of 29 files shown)


op_crates/fetch/21_formdata.js
```diff
@@ -442,6 +442,11 @@
      * @returns {FormData}
      */
     parse() {
+      // Body must be at least 2 boundaries + \r\n + -- on the last boundary.
+      if (this.body.length < (this.boundary.length * 2) + 4) {
+        throw new TypeError("FormData to short to be valid.");
```
aapoalas, line 447: `too short`

op_crates/fetch/20_headers.js
```diff
@@ -118,10 +122,54 @@
...
+    if (initialValue === null) return null;
+    const input = initialValue;
+    let position = 0;
+    const values = [];
+    let value = "";
+    while (position < initialValue.length) {
+      const res = collectSequenceOfCodepoints(
+        initialValue,
+        position,
+        (c) => c !== "\u0022" && c !== "\u002C",
```
ry, line 141: Tab or space? Add a comment?

op_crates/fetch/22_http_client.js
```diff
@@ -0,0 +1,41 @@
...
+((window) => {
+  const core = window.Deno.core;
+
+  /**
+   * @param {Deno.CreateHttpClientOptions} options
+   * @returns {HttpClient}
+   */
+  function createHttpClient(options) {
+    return new HttpClient(core.opSync("op_create_http_client", options));
+  }
```
ry, line 24: Aside: we should get rid of `Deno.createHttpClient`, `new Deno.HttpClient` should be sufficient.

ry (approved):
LGTM - great work - a leap towards better web compatibility.

op_crates/fetch/22_body.js
```diff
@@ -0,0 +1,338 @@
...
+    switch (type) {
+      case "ArrayBuffer":
+        return bytes.buffer;
+      case "Blob":
+        return new Blob([bytes], {
+          type: mimeType !== null ? mimesniff.serializeMimeType(mimeType) : "",
+        });
+      case "FormData": {
+        if (mimeType !== null) {
+          if (mimeType !== null) {
```
CrimsonCodes0, line 225: Really late, but the nested `if` statement body is unreachable, you have a duplicate condition.

op_crates/fetch/23_request.js
```diff
@@ -0,0 +1,521 @@
...
+      // 29.
+      this[_headers] = headersFromHeaderList(request.headerList, "request");
+
+      // 31.
+      if (Object.keys(init).length > 0) {
+        let headers = headerListFromHeaders(this[_headers]);
+        if (init.headers !== undefined) {
+          headers = init.headers;
+        }
+        headerListFromHeaders(this[_headers]).slice(
```
CrimsonCodes0, line 254: I can't tell which .slice method this is, but does it mutate the source? If not, this is dead code.

Issue #10192, opened 2021-04-15
Native HTTP/2 web server does not work within a WebWorker


starkpsmichael:
Seems like that the new http native bindings doesn't work in a WebWorker, see example below.

**http2.ts**
```ts
const body = new TextEncoder().encode("Hello World");
for await (const conn of Deno.listen({ port: 80 })) {
  (async () => {
    for await (const { respondWith } of Deno.serveHttp(conn)) {
      respondWith(new Response(body));
    }
  })();
}
```

**worker.ts**
```ts
new Worker(
  new URL("./http2.ts", import.meta.url).href,
  {
    type: "module",
    deno: {
      namespace: true,
      permissions: {
        env: false,
        hrtime: false,
        net: true,
        plugin: false,
        read: false,
        run: false,
        write: false,
      },
    },
  },
);

```

**Working**
```sh
deno run --unstable --allow-net http2.ts
```

**Not working**
```sh
deno run --unstable --allow-net --allow-read worker.ts
```

**Error Message**
```
error: Uncaught (in worker "") (in promise) TypeError: invalid op id: expected type `rusty_v8::data::Integer`, got `rusty_v8::data::Value`
    for await (const { respondWith } of Deno.serveHttp(conn)) {
                                             ^
    at dispatch (deno:core/core.js:76:12)
    at Object.opSync (deno:core/core.js:114:27)
    at Object.serveHttp (deno:runtime/js/40_http.js:13:27)
```

lucacasonato:
Issue here is that the http ops are not registered in the web worker isolate.

HTTP ops should be registered in the web worker (https://github.com/denoland/deno/blob/main/runtime/web_worker.rs#L252), just like they are in the main worker (https://github.com/denoland/deno/blob/main/runtime/worker.rs#L141).

satyarohith:
Working on it.

cztomsik:
I'm getting this with native plugin after last upgrade

bartlomieju:
> I'm getting this with native plugin after last upgrade

Not sure what you mean, please provide reproduction.

cztomsik:
it's here https://github.com/cztomsik/graffiti but I think it's because ABI changed so I will check that first
(I copied "headers" to the project so I don't need to depend on rusty_v8  https://github.com/cztomsik/graffiti/blob/30a9aa89f01a3ac121be024feef238c115e92704/libgraffiti/src/bindings/deno.rs#L88)

AaronO:
[...]

PR #10207, opened 2021-04-16
fix(runtime): include HTTP op in WebWorker

satyarohith:
Closes https://github.com/denoland/deno/issues/10192

### DIFF
--- a/cli/tests/workers/http_worker.js
+++ b/cli/tests/workers/http_worker.js
@@ -1,0 +1,10 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+const listener = Deno.listen({ hostname: "127.0.0.1", port: 4500 });
+for await (const conn of listener) {
+  (async () => {
+    const requests = Deno.serveHttp(conn);
+    for await (const { respondWith } of requests) {
+      respondWith(new Response("Hello world"));
+    }
+  })();
+}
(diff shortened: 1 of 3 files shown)


cli/tests/unit/worker_http.ts
```diff
@@ -0,0 +1,28 @@
...
+        },
+      },
+    );
+
+    assert(w);
+    const response = await fetch("http://localhost");
+    assert(await response.text());
+    w.terminate();
+  },
+);
```
bartlomieju, line 28: Please move this test to `cli/tests/workers/test.ts`

cli/tests/workers/http_worker.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+const body = new TextEncoder().encode("Hello World");
+for await (const conn of Deno.listen({ port: 80 })) {
```
bartlomieju, line 3: Try binding to some other port that is already used in tests, eg. 4500

bartlomieju (changes requested):

cli/tests/workers/http_worker.ts
satyarohith: Turns out the culprit is TypeScript. It's taking time for the server to boot up before the request is made resulting in a failure. I switched the HTTP worker script to JS for now.

bartlomieju (approved):
LGTM

Issue #10208, opened 2021-04-16
Replace runtime/01_web_util.js cloneValue with Deno.core.serialize/Deno.core.deserialize


kitsonk:
Now that we have `Deno.core.serialize()` and `Deno.core.deserialize()` available, we should replace the helper `cloneValue()`.

EladKeyshawn:
@kitsonk HI can i work on this?  😀
Anyhow why clone is not necessary? Since we can you serialize and then deserialize to get the same effect?

lucacasonato:
@EladKeyshawn That is exactly it. serialize + deserialize == clone.

bartlomieju:
Done in #10215

PR #10215, opened 2021-04-16
webutil: replace cloneValue impl with serialize/deserialize

EladKeyshawn:
From #10208


runtime/js/01_web_util.js
```diff
@@ -35,87 +35,7 @@
...
-        }
-        Reflect.setPrototypeOf(clonedObj, Reflect.getPrototypeOf(value));
-        return clonedObj;
-      }
-      case "symbol":
-      case "function":
-      default:
-        throw new DOMException("Uncloneable value in stream", "DataCloneError");
-    }
+    return Deno.core.deserialize(Deno.core.serialize(value));
```
evanwashere, line 38: should retain type checks and manual cloning for primitive values and ArrayBuffer/TypedArray

CrimsonCodes0 (commented):
The implementation should still throw the old DOMException, serialize/deserialize throw something else.

runtime/js/01_web_util.js
lucacasonato: As a performance optimisation?

evanwashere: yeah, Deno.core.serialize allocates a new buffer, so it will be big overhead for primitives and ArrayBuffer

bartlomieju: Reverted back

lucacasonato (approved):

PR #10220, opened 2021-04-17
refactor(op_crates/websocket): use Serialize for return values

crowlKats:


AaronO (changes requested):
I didn't switch this code over in that original "migration" PR, because I was waiting to have `ZeroCopyBuf` support in `serde_v8` to do it efficiently and right now `serde_v8` doesn't support serializing regular bytes since it wasn't really needed and would be somewhat inefficient without ZeroCopyBuf.

Also it would be cleaner to use an adjacently tagged enum instead of the separate `.kind` attribute, since it's redundant with the `enum` and that will also be faster since it will leverage internalized v8 strings and will avoid 2 allocs per call (one on the rust side and one on the v8 side)

op_crates/websocket/lib.rs
```diff
@@ -311,49 +308,25 @@ pub async fn op_ws_next_event(
   let cancel = RcRef::map(resource, |r| &r.cancel);
   let val = rx.next().or_cancel(cancel).await?;
   let res = match val {
-    Some(Ok(Message::Text(text))) => NextEventResponse {
-      kind: "string".to_string(),
-      data: Some(NextEventResponseData::String(text)),
-    },
+    Some(Ok(Message::Text(text))) => NextEventResponse::String(text),
     Some(Ok(Message::Binary(data))) => {
       // TODO(ry): don't use json to send binary data.
```
aapoalas, line 313: Is the TODO still valid or is the `NextEventResponse::Binary(data)` now fully binary instead of JSON?

crowlKats: As @AaronO mentioned, this still is an issue. It will be soon worked on.

bartlomieju (approved):
LTGM 👍

PR #10225, opened 2021-04-17
fix(runtime/js/http): Cancel body on response failure

### LINKED ISSUE #10193: Native HTTP: A streaming body/ServerSentEvents fails
I would be able to expect to be able to utilise [SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) with the native HTTP, and I can start a request, by setting the appropriate headers and return a `ReableStream` as a body, pushing "chunks" into the controller that represent the events.

But when the client attempts to navigate away or refresh, an unrecoverable/uncatchable error occurs:

```
    at async Object.respondWith (deno:runtime/js/40_http.js:153:11)
```

What I would have expected is that if the `Response` init was a readable stream (as it is in this case) the `ReadableStream` would be cancelled with the reason, which I would handle in the server code "gracefully".

nayeemrmn:
Fixes #10193.

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -228,3 +228,47 @@ unitTest(
     await promise;
   },
 );
+
+unitTest(
+  { perms: { net: true } },
+  async function httpServerCancelBodyOnResponseFailure() {
+    const promise = (async () => {
+      const listener = Deno.listen({ port: 4501 });
+      const conn = await listener.accept();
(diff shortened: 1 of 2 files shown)


lucacasonato:
Please rebase onto https://github.com/denoland/deno/pull/10203.

lucacasonato (approved):

lucacasonato:
Thanks @nayeemrmn

PR #10230, opened 2021-04-17
refactor(serde_v8): magic subfolder and minor cleanup

AaronO:
This reduces noise in `ser.rs` and makes room for expanding magic values, such as integrating `ZeroCopyBuf`.


bartlomieju (changes requested):
LGTM, but please add copyright comments

bartlomieju (approved):

PR #10231, opened 2021-04-17
refactor(core): move SerializablePkg to serde_v8

AaronO:
Also rename `MinValue` to `Primitive`


serde_v8/src/serializable.rs
```diff
@@ -20,3 +22,96 @@ impl<T: serde::Serialize> Serializable for T {
...
+    } else if tid == TypeId::of::<i8>() {
+      Self::Primitive(Primitive::Int8(unsafe { transmute_copy(&x) }))
+    } else if tid == TypeId::of::<i16>() {
+      Self::Primitive(Primitive::Int16(unsafe { transmute_copy(&x) }))
+    } else if tid == TypeId::of::<i32>() {
+      Self::Primitive(Primitive::Int32(unsafe { transmute_copy(&x) }))
+    } else if tid == TypeId::of::<i64>() {
+      Self::Primitive(Primitive::Int64(unsafe { transmute_copy(&x) }))
+    } else if tid == TypeId::of::<u8>() {
+      Self::Primitive(Primitive::UInt8(unsafe { transmute_copy(&x) }))
```
bartlomieju, line 102: Nitpick, could `downcast` be used here instead of `transmute_copy`?

serde_v8/src/serializable.rs
```diff
@@ -20,3 +22,96 @@ impl<T: serde::Serialize> Serializable for T {
     crate::to_v8(scope, self)
   }
 }
+
+/// SerializablePkg exists to provide a fast path for op returns,
+/// allowing them to avoid boxing primtives (ints/floats/bool/unit/...)
+pub enum SerializablePkg {
```
bartlomieju, line 28: The name of this enum is questionable, what does `Pkg` refer to?

serde_v8/src/serializable.rs
AaronO: We could use Any's `downcast_ref` here, but I don't think it would be better, it might in fact be more confusing. 

Ultimately we want to downcast and take ownership of specific "primitive" values here, so if we added `String` to the list of primitives we would have to do extra "unsafe" stuff besides just temporarily holding a reference to do some work as you usually do with `downcast_ref`.

serde_v8/src/serializable.rs
AaronO: We could call it `BoxedSerializable` and just say that it's functionally identical to `Box<dyn serde_v8::Serializable>` but more efficient for primitives (since many ops return primitives, e.g: ResourceIds, etc...)

bartlomieju: Let's keep the current name for now

bartlomieju (approved):
LGTM, thanks Aaron!

PR #10232, opened 2021-04-17
refactor(core): remove ZeroCopyBuf's dep on the bindings mod

AaronO:
Also cleanup `bindings::deserialize()/decode()` so they use the `ZeroCopyBuf` abstraction rather than reimplementing it

This cleanup will facilitate moving `ZeroCopyBuf` to `serde_v8` since it's now self contained and there are no other `get_backing_store_slice()` callers


bartlomieju (approved):
LGTM

PR #10233, opened 2021-04-18
feat(lsp): Implement textDocument/semanticTokens/full

jeanp413:
and textDocument/semanticTokens/range

Related #8643


cli/lsp/language_server.rs
```diff
@@ -2782,6 +2889,204 @@ mod tests {
...
+              24,
+              3,
+              0,
+              0,
+              0,
+              4,
+              9,
+              9,
+              2
+            ]
```
kitsonk, line 3017: I would try to use `#[rustfmt::skip]` here and reformat this in some more readable way manually, or externalise the JSON file, because this is just a lot of useless vertical space in an already big file.

cli/lsp/semantic_tokens.rs
```diff
@@ -0,0 +1,351 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
kitsonk, line 1: Is this based on logic from VSCode TypeScript language service?  If so, it would be good to mention it in here at the top, so future folks can at least understand where we are getting the logic from.

cli/lsp/text.rs
```diff
@@ -185,6 +189,20 @@ impl LineIndex {
     }
   }
 
+  pub fn line_length_utf16(&self, line: u32) -> Result<TextSize, AnyError> {
```
kitsonk, line 192: Probably need a test case for this.

cli/lsp/text.rs
```diff
@@ -20,7 +20,7 @@ where
   P: FnMut(&T) -> bool,
 {
   let mut left = 0;
-  let mut right = slice.len();
+  let mut right = slice.len() - 1;
```
kitsonk, line 23: What did you specifically run into that needed this logic updated?  It would be to get an example of what wasn't working as a test that the logic update makes sense.

kitsonk (commented):
A couple thoughts...

cli/lsp/language_server.rs
jeanp413: done

cli/lsp/semantic_tokens.rs
jeanp413: done

cli/lsp/text.rs
jeanp413: I end up removing this function and simplified the logic in `Classifications::to_semantic_tokens` because `start_pos.line == end_pos.line` is always true as there are no multiline tokens

cli/lsp/text.rs
jeanp413: Some test were failing: `lsp::text::tests::test_get_edits` and `lsp::text::tests::test_line_index`
For test_line_index this one in particular is was returning `line:2 character:0` without this change
```
assert_eq!(
      index.position_utf16(11.into()),
      lsp::Position {
        line: 1,
        character: 5
      }
    );
```
Now that `utf8_offsets` and `utf16_offsets` length is equal to `# of lines + 1` (because I added a last `push` call after the for loop so we can calculate the total text size) then `slice.len() - 1` is equal to `slice.len()` before my change

kitsonk (approved):
LGTM, thanks again for the contribution!

Issue #9840, opened 2021-03-20
Calls to Deno.chdir can cause panics


caspervonb:
We do a lot of `current_dir().unwrap` calls internally, which means that code like this:

```js
const dir = Deno.makeTempDirSync();
Deno.chdir(dir);
Deno.removeSync(dir);
```

Will cause the runtime to panic.

PR #10238, opened 2021-04-18
fix: do not panic on not found cwd

satyarohith:
This PR replaces all instances of `current_dir().unwrap()` with `current_dir()?` except the ones in the tests and `core/module_sepcifier.rs`.

Closes https://github.com/denoland/deno/issues/9840

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -5138,6 +5138,25 @@ console.log("finish");
     assert!(stderr.contains("BadResource"));
   }
 
+  #[cfg(not(windows))]
+  #[test]
+  fn should_not_panic_on_not_found_cwd() {
+    let output = util::deno_cmd()
+      .current_dir(util::root_path())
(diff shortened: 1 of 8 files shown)


bartlomieju:
@satyarohith can you add `.context("Failed to get current working directory")` to all those calls? It's hard to figure out what's the problem with the generic error that's returned from Rust

cli/tests/integration_tests.rs
```diff
@@ -5138,6 +5138,25 @@ console.log("finish");
     assert!(stderr.contains("BadResource"));
   }
 
+  #[cfg(not(windows))]
```
satyarohith, line 5141: The bug the PR is fixing doesn't seem to be a bug on windows as per https://github.com/denoland/deno/runs/2373907680?check_suite_focus=true#step:28:2087 hence not running the test on windows.

bartlomieju (approved):
LGTM, thank you Satya

PR #10246, opened 2021-04-18
build: use split-debuginfo

bartlomieju:
Stumbled upon in https://jakedeichert.com/blog/reducing-rust-incremental-compilation-times-on-macos-by-70-percent/

It lowers incremental debug build from about 1m10s to 30s on my machine.


piscisaureus:
This flag should be activated on OS X only. `split-debuginfo` is not supported (yet?) by rustc.

PR #10247, opened 2021-04-18
fix(cli): Windows - `deno install` shim should use first Deno executable on PATH rather than explictly deno.exe

dsherret:
Right now, `deno install` produces a shim like the following:

```batch
% generated by deno install %
@deno.exe "run" "--allow-read" "--allow-write" "--allow-net" "--allow-env" "--allow-run" "--no-check" "https://deno.land/x/deploy/deployctl.ts" %*
```

This is incorrect because the `deno` command might not be `deno.exe` (as is the case on my system where it runs from `deno.cmd` or `deno.ps1` via [bvm](https://github.com/bvm/bvm)).

This fix does the following:

1. It uses the first `deno` executable on the PATH, so it works the same way as executing the `deno` command.
2. It supports more than just a `deno.exe` executable being on the path.


lucacasonato (approved):
LGTM. Thanks

PR #10250, opened 2021-04-18
docs(manual/wasm): add more example code.

grian32:
Adds example code for how to use deno with .wasm files.

Not sure if something else is preffered over `readAll()` from `io/util.ts`, but it's what worked for me.

### DIFF
--- a/docs/getting_started/webassembly.md
+++ b/docs/getting_started/webassembly.md
@@ -20,3 +20,19 @@ const wasmInstance = new WebAssembly.Instance(wasmModule);
 const main = wasmInstance.exports.main as CallableFunction
 console.log(main().toString());
 ```
+
+And for files:
+
+<!-- deno-fmt-ignore -->
+
+```ts
+import { readAll } from "https://deno.land/std@0.93.0/io/util.ts";
+
+const file = await Deno.open("main.wasm");
+const wasmCode = await readAll(file);
+Deno.close(file.rid);
+const wasmModule = new WebAssembly.Module(wasmCode);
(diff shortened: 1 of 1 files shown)


docs/getting_started/webassembly.md
```diff
@@ -20,3 +20,19 @@ const wasmInstance = new WebAssembly.Instance(wasmModule);
 const main = wasmInstance.exports.main as CallableFunction
 console.log(main().toString());
 ```
+
+And for files:
+
+<!-- deno-fmt-ignore -->
```
kitsonk, line 26: Why are we ignoring formatting?

grian32: Not sure why, I just followed what the other code block did.

grian32: After checking, it turns out that it was unnecessary, and that the other codeblock had a fmt ignore there because it included an array. I've made a commit that removes it.

docs/getting_started/webassembly.md
```diff
@@ -20,3 +20,17 @@ const wasmInstance = new WebAssembly.Instance(wasmModule);
...
 console.log(main().toString());
 ```
+
+And for files:
+
+```ts
+import { readAll } from "https://deno.land/std@0.93.0/io/util.ts";
+
+const file = await Deno.open("main.wasm");
+const wasmCode = await readAll(file);
```
satyarohith, line 30: Use `Deno.readFile`: https://doc.deno.land/builtin/stable#Deno.readFile

lucacasonato (approved):
LGTM, thanks

PR #10061, opened 2021-04-07
feat(lsp): Implement textDocument/prepareCallHierarchy

jeanp413:
Also implements `callHierarchy/incomingCalls` and `callHierarchy/outgoingCalls`

Related #8643

Issue #10200, opened 2021-04-15
1.9.0 Deno.connect() types the connection to Conn<NetAddr> making code incompatible


aricart:
Previous versions of deno had non-generic `Conn`. This breaks libraries that work perfectly with any version of deno.

aricart:
Obviously, the backward and forwards solution is ``// @ts-ignore: Deno added generics to Conn` but maybe there's something that can be done that is more elegant.

bartlomieju:
@aricart could you provide example of code that got broken?

aricart:
Sorry - Deno.startTls changed signature, so code as below, will fail to compile in 1.9.0

```
let conn: Deno.Conn;
const hostname = "connect.ngs.global";

conn = await Deno.connect({hostname: hostname, port: 4222});
conn = await Deno.startTls(
  conn,
  { hostname },
);

console.log("connected");
```

aricart:
```bash
error: TS2345 [ERROR]: Argument of type 'Conn<Addr>' is not assignable to parameter of type 'Conn<NetAddr>'.
  Type 'Addr' is not assignable to type 'NetAddr'.
    Type 'UnixAddr' is missing the following properties from type 'NetAddr': hostname, port
  conn,


Pacifying it on 1.9.0 by specifying the `Conn<

**Comment by kitsonk** [comment]
Doing this:

```ts
const hostname = "connect.ngs.global";

let conn = await Deno.connect({ hostname: hostname, port: 4222 });
// other stuff happening here after detecting the server supports TLS

conn = await Deno.startTls(
  conn,
  { hostname: hostname },
);

console.log("connected");
```

Fixes it under Deno 1.9.  Is that not acceptable?

kt3k:
I think #10012 wasn't breaking with the above example if we had set default of type parameter to `NetAddr` like:

```
interface Conn<Address extends Addr = NetAddr>
```

PR #10255, opened 2021-04-19
revert: Conn type changes in #10012 and #10061

kitsonk:
This reverts a8057e3e06962a8d7c6330a085704bb4493eed04 and 9c7c9a35c12625bd4793c21539391d6b08d17e73 which introduced breaking changes.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1726,20 +1726,18 @@ declare namespace Deno {
    * Requires `allow-write` permission. */
   export function truncate(name: string, len?: number): Promise<void>;
 
-  export interface Addr {
-    transport: string;
-  }
-
-  export interface NetAddr extends Addr {
+  export interface NetAddr {
     transport: "tcp" | "udp";
     hostname: string;
     port: number;
   }
 
-  export interface UnixAddr extends Addr {
+  export interface UnixAddr {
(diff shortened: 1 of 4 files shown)


lucacasonato:
It might be breaking again now that `Conn` is not generic anymore. Maybe we should keep it being generic, and then just do nothing with the generic parameter?

kitsonk:
> It might be breaking again now that `Conn` is not generic anymore (for 1.9.0 users). Maybe we should keep it being generic, and then just do nothing with the generic parameter?

But that will likely be breaking then, because the generic does nothing, meaning some things that were assignable won't be assignable.

lucacasonato (approved):
Ok. Let's risk it 😄

piscisaureus:
Isn't it possible to support multiple signature? Maybe we can keep the new one, but add back support for the old signature?

lucacasonato:
> Isn't it possible to support multiple signature?

No. This is what https://github.com/denoland/deno/commit/a8057e3e06962a8d7c6330a085704bb4493eed04 attempted (through the use of a default type argument), but this failed. See https://github.com/denoland/deno/issues/10200.

piscisaureus:
> No. This is what [a8057e3](https://github.com/denoland/deno/commit/a8057e3e06962a8d7c6330a085704bb4493eed04) attempted (through the use of a default type argument), but this failed. See #10200.

Ok, ship it then.

PR #10256, opened 2021-04-19
fix(rt/http): correct URL in Request

lucacasonato:
This commit fixes the URL returned from `request.url` in the HTTP server
to be fully qualified. This previously existed, but was removed and
accidentially not readded during optimizations of the HTTP ops.

Returning a non fully qualified URL from `Request#url` is not spec
compliant.


bartlomieju (approved):

kitsonk:
Just to note, this will be a breaking change from 1.9.0, a good one, but I know it will break parts of oak and I expect others who implemented the behaviour present in 1.9.0.

Issue #10236, opened 2021-04-18
Support for windows ctrl-c signal


juzi5201314:
Hey, what I want to ask is, is there any plan to support ctrl c signal under windows?

I think this is a very useful function for the cli program. It can exit the program normally, close the resource, and trigger window.onunload.
Since tokio already has related functions, it should be very simple to implement.

This is what I simply changed: https://github.com/juzi5201314/deno/commit/9ea6d4c00770455b896671a730763c8f2d346992

bnoordhuis:
Anything that reduces the friction between Windows and UNIX is almost certainly welcome. Can you open your change as a pull request? Thanks.

PR #10258, opened 2021-04-19
feat(runtime/signal): support windows ctrl-c/ctrl-break signal(SIGINT/SIGBREAK)

juzi5201314:
Related issue: #10236

Implementation:
- [x] use tokio::signal::windows::ctrl_c instead of tokio::signal::unix::signal on windows.
- [x] use tokio::signal::windows::ctrl_break to implement SIGBREAK.
- [x] Use GenerateConsoleCtrlEvent in Deno.kill in windows to process SIGINT and SIGBREAK instead of terminating the process.
- [ ] change WindowsSignal/LinuxSignal/MacOSSignal  to enum Signal { and remove the type Signal = ... below, and remove the LinuxSignal and MacOSSignal types. (https://github.com/denoland/deno/pull/10258#discussion_r616601149)

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -584,6 +584,8 @@ declare namespace Deno {
     SIGIO = 29,
     SIGPWR = 30,
     SIGSYS = 31,
+
+    SIGBREAK = 0,
   }
   enum MacOSSignal {
     SIGHUP = 1,
@@ -617,12 +619,53 @@ declare namespace Deno {
     SIGINFO = 29,
     SIGUSR1 = 30,
     SIGUSR2 = 31,
+
+    SIGBREAK = 0,
   }
 
+  enum WindowsSignal {
+    SIGINT = 2,
+    SIGBREAK = 21,
+
+    // If the content of the `WindowsSignal` is different from that of `Linux/MacOSSignal`, it cannot pass the check
+    SIGHUP = 0,
+    SIGQUIT = 0,
+    SIGILL = 0,
+    SIGTRAP = 0,
+    SIGABRT = 0,
+    SIGEMT = 0,
+    SIGFPE = 0,
+    SIGKILL = 0,
+    SIGBUS = 0,
+    SIGSEGV = 0,
+    SIGSYS = 0,
+    SIGPIPE = 0,
+    SIGALRM = 0,
+    SIGTERM = 0,
+    SIGURG = 0,
+    SIGSTOP = 0,
+    SIGTSTP = 0,
+    SIGCONT = 0,
+    SIGCHLD = 0,
+    SIGTTIN = 0,
+    SIGTTOU = 0,
+    SIGIO = 0,
+    SIGXCPU = 0,
+    SIGXFSZ = 0,
+    SIGVTALRM = 0,
+    SIGPROF = 0,
+    SIGWINCH = 0,
+    SIGINFO = 0,
+    SIGUSR1 = 0,
+    SIGUSR2 = 0,
+  }
   /** **UNSTABLE**: Further changes required to make platform independent.
    *
    * Signals numbers. This is platform dependent. */
-  export const Signal: typeof MacOSSignal | typeof LinuxSignal;
+  export const Signal:
+    | typeof MacOSSignal
+    | typeof LinuxSignal
+    | typeof WindowsSignal;
 
   /** **UNSTABLE**: new API, yet to be vetted.
    *
--- a/cli/tests/unit/process_test.ts
+++ b/cli/tests/unit/process_test.ts
@@ -490,7 +490,11 @@ unitTest(
     } catch {
       // TODO(nayeemrmn): On Windows sometimes the following values are given
       // instead. Investigate and remove this catch when fixed.
-      assertEquals(status.code, 1);
+      // assertEquals(status.code, 1);
+
+      // update(soeur): After 53ea2b5925721c819a59b8ff1ea1140cfb86f416,
+      // the test found that it returns -1073741502 on windows.
+      assertEquals(status.code, -1073741502);
       assertEquals(status.signal, undefined);
     }
     p.close();
--- a/cli/tests/unit/signal_test.ts
+++ b/cli/tests/unit/signal_test.ts
@@ -21,86 +21,139 @@ unitTest(
         Deno.signal(1);
       },
       Error,
-      "not implemented",
+      "Windows only supports ctrl-c(SIGINT)!",
     );
     assertThrows(
       () => {
         Deno.signals.alarm(); // for SIGALRM
       },
       Error,
-      "not implemented",
+      "Windows only supports ctrl-c(SIGINT)!",
     );
     assertThrows(
       () => {
         Deno.signals.child(); // for SIGCHLD
       },
       Error,
-      "not implemented",
+      "Windows only supports ctrl-c(SIGINT)!",
     );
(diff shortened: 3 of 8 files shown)


piscisaureus:
Is there a reason to not also bind `ctrl-break` and map it to `SIGBREAK`.  This is what node.js does too (see the libuv documentation at https://github.com/libuv/libuv/blob/47e0c5c575e92a25e0da10fc25b2732942c929f3/docs/src/signal.rst#windows-notes).

juzi5201314:
> 是否有理由不同时绑定`ctrl-break`并映射到`SIGBREAK`？这也是node.js所做的（请参见libuv文档，[网址](https://github.com/libuv/libuv/blob/47e0c5c575e92a25e0da10fc25b2732942c929f3/docs/src/signal.rst#windows-notes)为https://github.com/libuv/libuv/blob/47e0c5c575e92a25e0da10fc25b2732942c929f3/docs/src/signal.rst#windows-notes）。

Oh you are right, i will do it

runtime/ops/signal.rs
```diff
@@ -5,24 +5,26 @@ use deno_core::ZeroCopyBuf;
...
+#[cfg(any(unix, windows))]
 use deno_core::error::bad_resource_id;
-#[cfg(unix)]
+#[cfg(any(unix, windows))]
 use deno_core::AsyncRefCell;
-#[cfg(unix)]
+#[cfg(any(unix, windows))]
 use deno_core::CancelFuture;
-#[cfg(unix)]
+#[cfg(any(unix, windows))]
```
piscisaureus, line 14: Remove `#[cfg(any(unix, windows))]` everywhere -> it is always true (practically speaking).

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -617,12 +619,53 @@ declare namespace Deno {
     SIGINFO = 29,
     SIGUSR1 = 30,
     SIGUSR2 = 31,
+
+    SIGBREAK = 0,
   }
 
+  enum WindowsSignal {
```
piscisaureus, line 626: It's a bit over the top to have platform specific types in the API.
We should change this to `enum Signal {` and remove the `type Signal = ...` below, and remove the `LinuxSignal` and `MacOSSignal` types.

@bartlomieju, @kt3k would you agree?

piscisaureus (commented):
Thank you @juzi5201314!. This looks very neat.

cli/tests/unit/signal_test.ts
```diff
@@ -21,88 +21,145 @@ unitTest(
...
+    const resolvable = deferred();
+    // This prevents the program from exiting.
+    const t = setInterval(() => {}, 1000);
+
+    let c = 0;
+    const sig = Deno.signal(Deno.Signal.SIGINT);
+    setTimeout(async () => {
+      await defer(20);
+      for (const _ of Array(3)) {
+        // Sends SIGUSR1 3 times.
```
piscisaureus, line 111: SIGUSR1?

juzi5201314: Oh, hh.  In fact, this is a copy of signalStreamTest, this is my negligence

juzi5201314 (commented):
The reason for this test failure is that windows cannot send signals, so the behavior of Deno.kill is to terminate the process.  Can I use [GenerateConsoleCtrlEvent](https://docs.microsoft.com/en-us/windows/console/generateconsolectrlevent) in Deno.kill to process SIGINT and SIGBREAK instead of directly terminating the process?
@piscisaureus

runtime/ops/signal.rs
```diff
@@ -101,7 +95,111 @@ pub fn op_signal_unbind(
...
+) -> Result<(), AnyError> {
+  super::check_unstable(state, "Deno.signal");
+  state
+    .resource_table
+    .close(rid)
+    .ok_or_else(bad_resource_id)?;
+  Ok(())
+}
+
+#[cfg(all(not(unix), not(windows)))]
```
piscisaureus, line 202: DId you mean to disable this test?

piscisaureus:
> The reason for this test failure is that windows cannot send signals, so the behavior of Deno.kill is to terminate the process. Can I use [GenerateConsoleCtrlEvent](https://docs.microsoft.com/en-us/windows/console/generateconsolectrlevent) in Deno.kill to process SIGINT and SIGBREAK instead of directly terminating the process?

Yes I'm okay with that.

cli/tests/unit/signal_test.ts
```diff
@@ -96,6 +96,63 @@ unitTest(
...
+    /*setTimeout(async () => {
+      await defer(20);
+      for (const _ of Array(3)) {
+        // Sends SIGINT 3 times.
+        Deno.kill(Deno.pid, Deno.Signal.SIGINT);
+        await defer(20);
+      }
+      sig.dispose();
+      resolvable.resolve();
+    });*/
```
juzi5201314, line 116: This commented out original code confuses me, running unit_test_runner.ts in clion terminal or using deno to run this code alone in windows terminal are all expected behaviors.
But running unit_test_runner.ts in windows terminal (or cmd.exe), this code will only send a signal signal (kill is only executed once?), even if the defer time is increased or sig.dispose is commented out, the same will be obtained. .
I had to use setTimeout(() => Deno.kill(Deno.pid, Deno.Signal.SIGINT), 20) to achieve the desired effect.

cli/dts/lib.deno.unstable.d.ts
juzi5201314: > 在API中拥有特定于平台的类型有些困难。
> 我们应该将其更改为`enum Signal {`并删除`type Signal = ...`以下内容，并删除`LinuxSignal`和`MacOSSignal`类型。
> 
> @ bartlomieju，@ kt3k你同意吗？

I think this is a bit difficult, because there are some signals that are different between Linux and macOS. For example, SIGBUS is 7 in Linux, but 10 in macOS.

runtime/ops/signal.rs
piscisaureus: ☝️

juzi5201314: 👌🏻

cli/tests/unit/process_test.ts
```diff
@@ -490,7 +490,11 @@ unitTest(
     } catch {
       // TODO(nayeemrmn): On Windows sometimes the following values are given
       // instead. Investigate and remove this catch when fixed.
-      assertEquals(status.code, 1);
+      // assertEquals(status.code, 1);
+
+      // update(soeur): After 53ea2b5925721c819a59b8ff1ea1140cfb86f416,
+      // the test found that it returns -1073741502 on windows.
+      assertEquals(status.code, -1073741502);
```
piscisaureus, line 497: STATUS_DLL_INIT_FAILED

Scionax:
After skimming the code, it seems like most of this is ready to be implemented but became stale and hasn't been merged.

Production servers (generally Linux) often need a good implementation for shutdown (save Redis, clear some folders, etc), but it needs to be tested in local environments, which may be Windows (like in my case). This simple addition would allow me to test those shutdown procedures without what would otherwise be quite a hassle.

Any update on when this can go live?

Davi-Danger:
What still needs to be done with this? (I don't entirely understand the processes involved with adding features) This is a really useful feature so I'd love for this to be added.

ry:
Closing because stale. 

I agree this is useful, but we need someone to champion it and open a new PR.

rivy:
😢

PR #10259, opened 2021-04-19
Try: link with ld.lld

piscisaureus:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,4 @@
 # Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
-
 [workspace]
 members = [
   "cli",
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -18,43 +17,44 @@ members = [
...
 ]
-exclude = [
-  "std/hash/_wasm"
-]
-
+exclude = ["std/hash/_wasm"]
 
 [profile.release]
-codegen-units = 1
-incremental = true
```
kt3k, line 28: Is this removal intentional?

piscisaureus: It is intentional, but that doesn't mean much.
I'm just trying out random things hoping to find something that really makes a difference.
If you have any ideas on how to speed up CI please try them out and/or let me know.

kt3k: I tried to remove `codegen-units = 1` on my fork, and it seemed having some good effects.

PR: https://github.com/kt3k/deno_ci_test/pull/3
run: https://github.com/kt3k/deno_ci_test/runs/2388627780

windows and mac both passed in 25 mins with `cli/main.rs` modified.

I think `codegen-units = 1` is bad for CI performance because it prevents some parallel executions in build process. Do you know what was the motivation of this config?

piscisaureus: > windows and mac both passed in 25 mins with cli/main.rs modified.

How big is the difference with the main branch?
I see you are only modifying `cli/main.rs`, so only `target/{debug,release}/deno gets rebuilt; I would suggest also touching `runtime/js/11_timers.js` and see how that takes to build.

> Do you know what was the motivation of this config?

I think it makes the `deno` binary either slower or bigger (probably bigger).
Currently we release a binary of ~76mb. https://deno.land/benchmarks#executable-size
If the size difference is small we could drop remove `codegen-units = 1` everywhere.
Otherwise, we could turn it off for PR builds, but leave it there for `main` branch builds.

PR #10264, opened 2021-04-20
Fix(lsp): document symbol performance mark

jeanp413:
Fixes incorrect performance mark in `document_symbol`
cc @kitsonk


kitsonk (approved):
LGTM

PR #10273, opened 2021-04-20
refactor(core): specialize ops on serde_v8

AaronO:
This PR further specializes the op-layer on `serde_v8`, in short it allows ops to receive & return `serde_v8` De/Serializables of which regular `serde` De/Serializables are a subset, allowing to encode/decode v8 serialized values in the op-layer.

This allows moving `core.serialize` and `core.deserialize` from bindings to ops (as a general rule we want prefer to use ops when possible).

But it's important to keep in mind that this change somewhat leaks v8 into the op-layer since implementors can now implement their own `to_v8`/`from_v8` methods and capture v8 scopes and values which ultimately are isolate references.

That leak is minimal but it's important to be aware that it's there. If we want to avoid this abstraction leak, we would have to move `SerializedValue` support from `core` to `serde_v8` but I know @inteon wanted the optionality to control v8 serialization in core since this will evolve as our message passing between workers becomes more sophisticated.

## Changes

- [x] Add `serde_v8::Deserializable` (previously only had `serde_v8::Serializable`)
- [x] Specialize ops serialization code to these serializable traits (instead of plain serde serializables)
- [x] Implement `SerializedValue` type in `core/`

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -13,6 +13,7 @@ mod ops_json;
 pub mod plugin_api;
 mod resources;
 mod runtime;
+mod serialized_value;
 mod zero_copy_buf;
 
 // Re-exports
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -11,7 +11,6 @@ use crate::ZeroCopyBuf;
 use futures::Future;
 use indexmap::IndexMap;
 use rusty_v8 as v8;
-use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::cell::RefCell;
 use std::iter::once;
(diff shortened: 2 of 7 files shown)


AaronO:
@ry @inteon As mentioned on Discord, I think this is a bad idea. Whilst we could move code that uses v8 SerializedValues to ops, I think it's cleaner to keep them as bindings.

Moving them to ops would hurt separation of concerns and make the op layer more complicated just for a handful of infrequently used ops.

AaronO:
Closing this PR for now (for the aforementioned reasons)

PR #10297, opened 2021-04-21
refactor(core): simplify error handling

AaronO:
This PR simplifies error handling in `core.js`. v8 builtin errors are registered by default so new runtimes no longer need to bootstrap with:
```
Deno.core.registerErrorClass('Error', Error);
```

One less gotcha when bootstrapping runtimes

## Changes
- [x] register builtin v8 errors in `core.js` so consumers don't have to
- [x] remove complexity of error args handling
(consumers must provide a constructor with custom args, core simply provides msg arg)
- [x] Adapt `99_main.js`
- [x] Adapt other runtimes


ry (approved):
LGTM 👍

PR #10299, opened 2021-04-21
refactor(core): kill recv() and init()

AaronO:
`init()` was previously needed to init the shared queue, but now that it's gone `init()` is only used to setup JsRuntime's `js_recv_cb`.

## Changes

- [x] Replace `JsRuntime::core_js_init()` with `::init_recv_cb()`
- [x] Remove `recv` binding
- [x] Cleanup misc stuff regarding `recv` (`FIXME` in `core/runtime.rs`, ...)


ry (approved):
LGTM - nice clean up

PR #10307, opened 2021-04-22
refactor(core): rename send() to opcall()

AaronO:
I think it's a better fit since `recv()` was killed (#10299) and `opcall <> syscall` (`send/recv` is reminiscent of request/response and custom payloads)

This could be camelCased to `opCall` but I think `opcall` is fine since it can be viewed as a single noun/verb similar to `syscall`

## Changes

- [x] Rename `send` to `opcall` (in `core.js` and `core/bindings.rs`)
- [x] Update calls to `core.send()` in tests
- [x] Possibly remove `dispatch` and `dispatchByName` (in favour of `opcall/opSync/opAsync`)


ry:
I'm fine with this rename.

>  Possibly rename dispatch and dispatchByName to opcall and opcallByName

Sure. Ideally we'd reduce these wrappers - are they all necessary?

AaronO:
> I'm fine with this rename.
> 
> > Possibly rename dispatch and dispatchByName to opcall and opcallByName
> 
> Sure. Ideally we'd reduce these wrappers - are they all necessary?

I was planning to strip them out for 1.10, we don't really need them if we can clean things up for plugins

ry (approved):
LGTM - thanks

Issue #10312, opened 2021-04-22
rustls refuses to establish connections to self signed TLS certificates


lucacasonato:
### Reproduction

The key chain this uses is made up of two parts. The CA (`./test_util/wpt/tools/certs/cacert.pem`) and the end entity certificate (./test_util/wpt/tools/certs/web-platform.test.pem).

To start the testing server:

```sh
deno run -A --unstable ./tools/wpt.ts setup
cd ./test_util/wpt
./wpt serve
```

See that `curl`ing the endpoint with the correct CA cert in the roots works, and without doesn't:

```sh
$ curl -v https://web-platform.test:8444/ --cacert ./test_util/wpt/tools/certs/cacert.pem
...
* Server certificate:
*  subject: CN=web-platform.test
*  start date: Apr 12 02:33:49 2021 GMT
*  expire date: Apr 12 02:33:49 2022 GMT
*  subjectAltName: host "web-platform.test" matched cert's "web-platform.test"
*  issuer: CN=web-platform-tests
*  SSL certificate verify ok.
...
<li class="dir"><a href="xslt/">xslt</a></li>
</ul>
* Connection #0 to host web-platform.test left intact


$ curl -v https://web-platform.test:8444/ 
* SSL certificate problem: unable to get local issuer certificate
* Closing connection 0
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

See that it doesn't work with Deno in any configuration:

```sh
$ deno eval --cert ./test_util/wpt/tools/certs/cacert.pem -Ldebug 'console.log(await fetch("https://web-platform.test:8444/"))'
Sending fatal alert BadCertificate
error: Uncaught (in promise) TypeError: error sending request for url (https://web-platform.test:8444/): error trying to connect: invalid certificate: UnknownIssuer
console.log(await fetch("https://web-platform.test:8444/"))
            ^
    at unwrapOpResult (deno:core/core.js:100:13)
    at async mainFetch (deno:op_crates/fetch/26_fetch.js:165:18)
    at async fetch (deno:op_crates/fetch/26_fetch.js:295:22)
    at async file:///mnt/starship/Projects/github.com/denoland/deno/$deno$eval.ts:1:13

[...]

PR #10318, opened 2021-04-22
fix: parse websocket messages correctly

### LINKED ISSUE #10300: Error in WebSocket#eventLoop
I experience the following error when running my test suite with 1.9.1:

```
error: Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'string' in error
    at WebSocket.#eventLoop (deno:op_crates/websocket/01_websocket.js:315:22)
```

https://github.com/denoland/deno/blob/main/op_crates/websocket/01_websocket.js#L315
https://github.com/LukeChannings/moviematch/runs/2404786637

It seems like `message` can be a string, and there's no check to handle that case.

lucacasonato:
Fixes #10300

No tests because WPT is blocked by https://github.com/denoland/deno/issues/10312.

### DIFF
--- a/op_crates/websocket/01_websocket.js
+++ b/op_crates/websocket/01_websocket.js
@@ -307,60 +307,71 @@
 
     async #eventLoop() {
       while (this.#readyState === OPEN) {
-        const message = await core.opAsync(
+        const { kind, value } = await core.opAsync(
           "op_ws_next_event",
           this.#rid,
         );
 
-        if ("string" in message) {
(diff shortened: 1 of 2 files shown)


crowlKats:
maybe change it back to the switch statement like we had it before?

lucacasonato:
> maybe change it back to the switch statement like we had it before?

Done

ry (approved):
Let's just land it.

PR #10319, opened 2021-04-22
fix(op_crates/webgpu): move non-null op buffer arg check when needed

crowlKats:
Due to this, currently a few webgpu examples dont work. Was introduced in #9944


lucacasonato (approved):
LGTM, thanks.

PR #10325, opened 2021-04-23
chore: update dependencies

lucacasonato:
This updates swc_* crates to latest, and various other deps.

PR #10327, opened 2021-04-23
chore: release crates

lucacasonato:
Release crates for the cli 1.9.2 release.

Still waiting on #10325

### DIFF
--- a/bench_util/Cargo.toml
+++ b/bench_util/Cargo.toml
@@ -8,10 +8,11 @@ description = "Bench and profiling utilities for deno crates"
 license = "MIT"
 readme = "README.md"
 repository = "https://github.com/denoland/deno"
+publish = false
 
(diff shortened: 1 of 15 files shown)


bench_util/Cargo.toml
```diff
@@ -8,10 +8,11 @@ description = "Bench and profiling utilities for deno crates"
 license = "MIT"
 readme = "README.md"
 repository = "https://github.com/denoland/deno"
+publish = false
```
bartlomieju, line 11: I reverted a commit that made usage of `bench_util` a couple days ago. I'm pretty sure we'll have to pubish this crate sooner or later.

bartlomieju (approved):

bench_util/Cargo.toml
lucacasonato: We'll probably have to rename it before doing that though. I assume we don't want to publish a crate named `bench_util`.

ry (approved):

PR #10329, opened 2021-04-23
Limit CI Cache key

ry:
depends on https://github.com/denoland/deno/pull/10322


ry:
actually this doesn't work for debug/release builds on ubuntu

PR #10332, opened 2021-04-23
BREAKING(unstable): Rename Deno.emit() bundle options to "module" and "classic"

### LINKED ISSUE #10288: EmitOptions::bundle values should be "module" and "classic"
- `Deno.emit({ bundle: "esm" })` -> `Deno.emit({ bundle: "module" })`
- `Deno.emit({ bundle: "iife" })` -> `Deno.emit({ bundle: "classic" })`

See `WorkerOptions::type`. Note that an iife is just a sensible way of emitting a classic script (whereas modules are already encapsulated), it's not the true distinction.

nayeemrmn:
Closes #10288.

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -441,10 +441,10 @@ declare namespace Deno {
    */
   export interface EmitOptions {
     /** Indicate that the source code should be emitted to a single file
-     * JavaScript bundle that is a single ES module (`"esm"`) or a single file
-     * self contained script we executes in an immediately invoked function
-     * when loaded (`"iife"`). */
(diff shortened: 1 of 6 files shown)


ry (commented):
I'm all for this - I think these names make more sense.

It would be useful it people using the old names failed gracefully. I suppose they'll get type errors.

Leaving this up to @kitsonk to merge.

kitsonk:
While I can understand it causes some churn in people's code, there were explicitly good reasons why we keep APIs as unstable, and if we do that and then refuse to break them or carry the baggage of refactorings forward, there is no reason to hold an API as unstable, IMO.

kitsonk (approved):
LGTM

PR #10333, opened 2021-04-23
Key CI cache on build profile

ry:
depends on https://github.com/denoland/deno/pull/10322

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,14 +19,9 @@ jobs:
             kind: test_release
           - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
             kind: bench
-          # TODO(ry) Ideally we could use ubuntu-latest-xl for lint but there's
-          # a bug with dprint that is preventing this from running.
-          - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -184,7 +190,7 @@ jobs:
           path: |
             ./target
           key:
-            d-${{ matrix.os }}-${{ matrix.kind }}-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('Cargo.lock') }}
+            a-cargo-target-${{ matrix.os }}-${{ matrix.profile }}-${{ hashFiles('Cargo.lock') }}
```
ry, line 193: The purpose of this patch is to reduce the number of distinct cargo-target caches we store by keying on build profile rather than "matrix.kind". The worry is that we're hitting our 5G cache limit before we're even able to store one set of caches, so we are continually rolling.

kt3k (approved):
LGTM

ry:
We are aware that "lint debug ubuntu" and "test debug ubuntu" sharing caches is not ideal - likely the lint cache will be written first. However neither are the bottleneck now. We proceed on the assumption that we're hitting the 5 GB cache limit on a single PR, inducing a forever-rolling behavior.

If "test debug ubuntu" becomes the bottleneck in the future we will revisit.

PR #10336, opened 2021-04-23
refactor(core): move builtin ops to their own file

AaronO:
## Rationale

- Cleaner to split builtin ops (`op_resources` and `op_close`) and the op internals (in `ops.rs`)
- The set of builtin ops will likely grow to include `print`, `encode` & `decode` once `serde_v8` has `ZeroCopyBuf`
- Once extensions land (#9800) it would make sense to bundle these ops as an `Extension`, we could even include `core.js` and `error.js` as a "builtins" extension


bartlomieju (approved):
LGTM

PR #10338, opened 2021-04-23
fix(console): circular customInspect

### LINKED ISSUE #9483: stack overflow when custom inspect calls Deno.inspect on `this`
If an object has a custom inspect function attached to it and it calls Deno.inspect with `this` as parameter (so that it has to invoke the same custom inspect function over and over again), Deno **crashes** with a stack overflow error.
Node.js just throws a "Maximum call stack size" error, which can be caught in JavaScript code.
I can only reproduce it on Windows. Running it in WSL throws the expected RangeError. 

Minimal repro:
```js
console.log({
  [Deno.customInspect]: function() {
    return Deno.inspect(this);
  }
});
```

```js
$ deno run index.ts
thread 'main' has overflowed its stack

$ deno --version
deno 1.7.2 (release, x86_64-pc-windows-msvc)
v8 8.9.255.3
typescript 4.1.3
```

AaronO:
This PR allows for more robust circular reference detection by keeping a global object stack that reflects the `inspectValue` call-tree, instead of passing around `Set`s of context that would fail with `customInspect` etc...

**Note:** this could become "vulnerable" to unbound `CTX_STACK` growth if `customInspect()` throws in a tight loop, this would linearly increase execution time and mem usage whilst keeping references to otherwise dangling objects. So I'm happy to wrap this in a try-catch-finally so the stack is always popped even if an error is thrown

Fixes #9483


bartlomieju:
@nayeemrmn @kt3k please take a look when you get a moment

nayeemrmn (approved):
LGTM

AaronO:
@nayeemrmn I think it would make sense to add that guard I suggested. Thoughts for or against ? :
> Note: this could become "vulnerable" to unbound CTX_STACK growth if customInspect() throws in a tight loop, this would linearly increase execution time and mem usage whilst keeping references to otherwise dangling objects. So I'm happy to wrap this in a try-catch-finally so the stack is always popped even if an error is thrown

nayeemrmn:
I see, that wasn't a problem with `ctx` passed as an argument but it needs to be a global to catch `Deno.inspect` custom inspect recursions. Yeah we should ensure it always pops.

kt3k:
> Note: this could become "vulnerable" to unbound CTX_STACK growth if customInspect() throws in a tight loop, this would linearly increase execution time and mem usage whilst keeping references to otherwise dangling objects. So I'm happy to wrap this in a try-catch-finally so the stack is always popped even if an error is thrown

I think that's not only vulnerable but it could cause wrong detection of circular object (because referenced objects remain in the stack).

The following code shows `Error Error` with the current deno, but shows `Error [Circular]` with this change. The same object being inspected differently feels strange to me..
```ts
let a = {
  [Deno.customInspect]() {
    throw new Error();
  }
}

try {
  console.log(a);
} catch {
  console.log('Error');
}

try {
  console.log(a);
} catch {
  console.log('Error');
}
```

AaronO:
> > Note: this could become "vulnerable" to unbound CTX_STACK growth if customInspect() throws in a tight loop, this would linearly increase execution time and mem usage whilst keeping references to otherwise dangling objects. So I'm happy to wrap this in a try-catch-finally so the stack is always popped even if an error is thrown
> 
> I think that's not only vulnerable but it could cause wrong detection of circular object (because referenced objects remain in the stack).
> 
> The following code shows `Error Error` with the current deno, but shows `Error [Circular]` with this change. The same object being inspected differently feels strange to me..
> 
> ```ts
> let a = {
>   [Deno.customInspect]() {
>     throw new Error();
>   }
> }
> 
> try {
>   console.log(a);
> } catch {
>   console.log('Error');
> }
> 
> try {
>   console.log(a);
> } catch {
>   console.log('Error');
> }
> ```

@kt3k Yep I was fully aware that it impacted correctness, should have specified that too. But I emphasized the mem/cpu vulnerability since that could be used to push a system to be OOM or stuck on CPU (in the extreme)

kt3k (approved):
LGTM

PR #10340, opened 2021-04-23
refactor(core): move op cache sync responsibility to rust space

AaronO:
Even if bootstrapping the JS runtime is low level, it's an abstraction leak of core to require users to call `Deno.core.ops()` in JS space.

So instead we're introducing a `JsRuntime::sync_ops_cache()` method, once we have runtime extensions a new runtime will ensure the ops cache is setup (for the provided extensions) and then loading/unloading plugins should be the only operations that require op cache syncs.

This also splits `Deno.core.ops()` into `Deno.core.syncOpsCache()` and `Deno.core.ops()`, which respectively sync the ops cache and return the ops cache without syncing it (useful for debugging or testing)

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -372,7 +372,7 @@ impl JsRuntime {
   /// Ensures core.js has the latest op-name to op-id mappings
   pub fn sync_ops_cache(&mut self) {
     self
-      .execute("deno:core/internals", "Deno.core.syncOpsCache()")
+      .execute("<anon>", "Deno.core.syncOpsCache()")
       .unwrap()
   }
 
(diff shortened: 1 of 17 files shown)


core/runtime.rs
```diff
@@ -369,6 +369,13 @@ impl JsRuntime {
     state.js_recv_cb.replace(v8::Global::new(scope, cb));
   }
 
+  /// Ensures core.js has the latest op-name to op-id mappings
+  pub fn sync_ops_cache(&mut self) {
+    self
+      .execute("deno:core/internals", "Deno.core.syncOpsCache()")
```
ry, line 375: Maybe "<anon>" is a better name for this script.

AaronO: Done

bartlomieju (approved):
LGTM

PR #10343, opened 2021-04-24
feat(cli/upgrade): add download progress

### LINKED ISSUE #10291: Download percentage while updating ( `deno upgrade` )
Using Deno with Windows Terminal 1.7.1033.0 on Windows 10 x64 20H2.

I quote from the manual:

> To update a previously installed version of Deno, you can run:
> 
> ```
> deno upgrade
> ```
> 
> This will fetch the latest release from https://github.com/denoland/deno/releases, unzip it, and replace your current executable with it.

These zip files size currently is between 23 and 30 MB.
Such a download may take some time to complete depending on the internet connection.

My suggestion is to display the current download percentage while downloading the relevant zip file:
i.e. to display below `Download has been file` a progress percentage, 
e.g.: `Downloading 1.0% of 23.00 MiB`

Currently it doesn't display any download progress until it completes(+then unzips it) :

<details>
<summary>screenshot</summary>

</details>

If it helps in visualizing it, here is how youtube-dl displays the download percentage:

Thank you

crowlKats:
Closes #10291

Cursor movement needs some fixing.

https://user-images.githubusercontent.com/13135287/115950129-bbc46180-a4d9-11eb-9d82-91ca8ef2d6a5.mov

Personally I don't think we need this and shouldn't be adding it

### DIFF
--- a/cli/tools/upgrade.rs
+++ b/cli/tools/upgrade.rs
@@ -8,6 +8,7 @@ use deno_runtime::deno_fetch::reqwest;
 use deno_runtime::deno_fetch::reqwest::Client;
 use semver_parser::version::parse as semver_parse;
 use std::fs;
+use std::io::Write;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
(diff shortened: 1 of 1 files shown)


cli/tools/upgrade.rs
```diff
@@ -167,7 +168,25 @@ async fn download_package(
...
-    Ok(res.bytes().await?.to_vec())
+
+    let total_size = res.content_length().unwrap() as f64;
+    let mut current_size = 0.0;
+    let mut data = Vec::with_capacity(total_size as usize);
+    let mut stream = res.bytes_stream();
+    while let Some(item) = stream.next().await {
+      let bytes = item?;
+      current_size += bytes.len() as f64;
+      data.extend_from_slice(&bytes);
```
caspervonb, line 179: 1. Print `ESC [1G` to move the cursor to column 1 before printing.
2. Clear line with `ESC [2k` to clear the line without moving the cursor.
3. Print, which should move the cursor to end of line.

caspervonb: (Only do step 1 and 2 if stdout is a tty).

crowlKats: That wont do, renders the text too slow & laggy, making it extremely difficult to read

ry (approved):
LGTM - thanks!

PR #10351, opened 2021-04-24
ci: only run WPT on linux

ry:
Because CI is so slow on for OSX and Windows, we should only run WPT on Linux.


piscisaureus (approved):
I concur!

PR #10353, opened 2021-04-24
fix: invalid return types for asynchronous and synchronous `File#truncate`

sno2:
Not that much to it.


caspervonb (approved):
LGTM

bartlomieju (approved):
LGTM, thanks for the fix @sno2

PR #10358, opened 2021-04-24
fix(cli/dts): sleepSync doesn't return a Promise

AaronO:
Per its name it's synchronous for the current thread


piscisaureus:
Almost painfully obvious. LGTM!

piscisaureus (approved):

PR #10359, opened 2021-04-24
feat(runtime/http): websocket support

### LINKED ISSUE #10211: Support web sockets with hyper binding
Probably looks like this:
```js
let { respondWith, request } = await httpConn.nextRequest();
let { ws, response } = Deno.serveWebsocket(request);
respondWith(response);
// use web standard ws
```

Should use [tungstenite](https://crates.io/crates/tungstenite) 

Probably this should proceed #10210

crowlKats:
Closes #10211


runtime/js/40_http.js
```diff
@@ -67,8 +67,9 @@
       const request = fromInnerRequest(innerRequest, "immutable");
 
       const respondWith = createRespondWith(responseSenderRid, this.#rid);
+      const upgradeWebsocket = createUpgradeWebsocket(requestBodyRid);
 
-      return { request, respondWith };
+      return { request, respondWith, upgradeWebsocket };
```
ije, line 72: can we use `upgrade` instaed of `upgradeWebsocket ` ?

crowlKats: No, upgrading isn't something exclusive to websockets. that would end up being confusing.

ry (commented):
Is this actually working? Would be very nice to have a test in cli/tests/unit/http_test.ts if so.

crowlKats:
@ry well we first need #10365

runtime/js/40_http.js
```diff
@@ -219,6 +220,73 @@
...
+              break;
+            case "binary":
+            case "pong":
+            case "ping":
+              await core.opAsync("op_http_ws_close", {
+                rid,
+                kind,
+              }, data);
+              break;
+          }
```
bnoordhuis, line 275: Maybe add a `default:` that throws an illegal argument exception?

bnoordhuis (approved):
Needs regression tests but LGTM apart from that, thanks!

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1153,9 +1153,44 @@ declare namespace Deno {
...
+
+  export interface WebSocketUpgrade {
+    response: Response;
+    ws: WebSocketConn;
+  }
+
   export interface RequestEvent {
     readonly request: Request;
     respondWith(r: Response | Promise<Response>): Promise<void>;
+    upgradeWebsocket(): Promise<WebSocketUpgrade>;
```
nayeemrmn, line 1193:
```suggestion
    upgradeWebSocket(): Promise<WebSocketUpgrade>;
```

runtime/js/40_http.js
nayeemrmn: > Of course that would make sense if we would support upgrading other kind of conns out of the box

Other hypothetical upgrades would have different return types, so they should just have individual names anyway.

ry:
@crowlKats I'm a bit confused about how this depends on #10365... I haven't dived into this PR but I'm not sure why we would need to expose WebSocketStream in order to hook up the hyper-tungstenite bindings?

I don't care so much about WebSocketStream support - since it's still an tentative API - but I'm rather keen to land this PR.

runtime/js/40_http.js
```diff
@@ -219,6 +220,66 @@
     });
   }
 
+  function createUpgradeWebsocket(rid) {
```
nayeemrmn, line 223:
```suggestion
  function createUpgradeWebSocket(rid) {
```

runtime/js/40_http.js
```diff
@@ -67,8 +67,9 @@
       const request = fromInnerRequest(innerRequest, "immutable");
 
       const respondWith = createRespondWith(responseSenderRid, this.#rid);
+      const upgradeWebSocket = createUpgradeWebSocket(requestBodyRid);
 
-      return { request, respondWith };
+      return { request, respondWith, upgradeWebSocket };
```
lucacasonato, line 72: instead of returning `upgradeWebsocket` here there should be a global `Deno.upgradeWebSocket` that takes a `Request` and returns both a `WebSocket(Stream)` and a `Response`. The connection will only be upgraded after the user passes this `Response` to `respondWith`.

runtime/js/40_http.js
```diff
@@ -227,23 +274,61 @@
...
+      request.headers["Sec-WebSocket-Key"] +
+        "258EAFA5-E914-47DA-95CA-C5AB0DC85B11",
+    );
+    const accept = await crypto.subtle.digest("SHA-1", key);
+
+    const r = newInnerResponse(101);
+    r.headerList = [
+      ["Upgrade", "websocket"],
+      ["Connection", "Upgrade"],
+      ["Sec-WebSocket-Accept", forgivingBase64Encode(new Uint8Array(accept))]
```
lucacasonato, line 318: header names should all be lower-case in the header list

lucacasonato:
Blocked on https://github.com/hyperium/hyper/pull/2587

lucacasonato (approved):
LGTM. Thanks for this great feature!

We need to add some documentation for this to https://deno.land/manual@v1.11.5/runtime/http_server_apis in a follow up commit.

lucacasonato (changes requested):
Oh actually `upgradeWebsocket` is still async. That needs to become synchronous.

lucacasonato (approved):
LGTM now. Thanks!

PR #10364, opened 2021-04-25
fix(cli): canonicalize coverage dir

caspervonb:
Fixes https://github.com/denoland/deno_std/pull/879


caspervonb:
Drafting this for a minute, it needs a test.

PR #10365, opened 2021-04-25
feat(extensions/websocket): implement websocketstream

### LINKED ISSUE #8831: WebSocketStream API
The WebSocketStream API is an API for usage of websockets using streams.
There is no formal spec for it yet.
https://web.dev/websocketstream/
https://docs.google.com/document/d/1XuxEshh5VYBYm1qRVKordTamCOsR-uGQBCYFcHXP4L0/edit#
https://docs.google.com/document/d/1La1ehXw76HP6n1uUeks-WJGFgAnpX2tCjKts7QFJ57Y/edit
https://github.com/ricea/websocketstream-explainer/
https://chromestatus.com/feature/5189728691290112

Would there be interest in having it? From discussions it seems the answer would be yes, including https://github.com/denoland/deno/pull/7450#issuecomment-702013713

crowlKats:
Needed for #10359.

Implements the WebSocketStream spec.

Resources:
https://docs.google.com/document/d/1La1ehXw76HP6n1uUeks-WJGFgAnpX2tCjKts7QFJ57Y/edit
https://docs.google.com/document/d/1XuxEshh5VYBYm1qRVKordTamCOsR-uGQBCYFcHXP4L0/edit
https://github.com/ricea/websocketstream-explainer
https://chromestatus.com/feature/5189728691290112

Closes #8831


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1265,3 +1265,28 @@ declare interface WorkerOptions {
     };
   };
 }
+
+declare interface WebSocketStreamOptions {
+  protocols?: string[];
```
aapoalas, line 1270: Maybe an unimportant thing but I presume the protocols can only be from a known enumeration, right?

lucacasonato: No, protocols can be an arbitrary HTTP token string.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1265,3 +1265,28 @@ declare interface WorkerOptions {
...
+
+declare interface WebSocketConnection {
+  readable: ReadableStream<string | Uint8Array>;
+  writable: WritableStream<string | Uint8Array>;
+  extensions: string;
+  protocol: string;
+}
+
+declare interface WebSocketCloseInfo {
+  code?: number;
```
aapoalas, line 1282: Closing with an empty object seems odd. Should this be an either-or-both type of thing?

crowlKats: empty object/no value passed means default values.

aapoalas: Okay, in that case I guess the `close` function should have a signature of `close(closeInfo?: WebSocketCloseInfo)` to allow omitting the empty object entirely.

crowlKats: I guess so. The problem is the spec is barely non-existent, and the few things that exist are at times a bit contradictory (especially with the close stuff)

op_crates/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,247 @@
...
+          "SyntaxError",
+        );
+      }
+
+      this.#url = wsURL.href;
+
+      core.opSync("op_ws_check_permission", this.#url);
+
+      if (
+        options?.protocols?.some((x) =>
```
aapoalas, line 114: Slight optimization:
```javascript
if (options?.protocols?.length > 1 && options.protocols.some((x, i) => options.protocols.indexOf(x, i + 1) !== -1))
```

AaronO: Might be cleaner and possibly faster (though I suspect not) to just construct a `Set()` of the protocols and then compare cardinality/length. I would opt for the cleaner route in absence of benchmarks or being in a known hotpath.

crowlKats: This is an identical copy of the code used in the normal WebSocket client. Such optimizations would make more sense in their own PR, or most likely together with when we make WebSocket more spec compliant.

AaronO: Agreed

bnoordhuis:
@crowlKats Is this ready for review?

crowlKats:
@bnoordhuis there are leaking async ops, and not sure how those can be fixed. i think it might have to do with the pull on the readable stream, but not sure. that aside, it would be ready

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1399,3 +1399,28 @@ declare interface WorkerOptions {
     };
   };
 }
+
+declare interface WebSocketStreamOptions {
+  protocols?: string[];
```
bnoordhuis, line 1404:
```suggestion
  protocols?: string | string[];
```
If I'm not mistaken?

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,302 @@
...
+  const core = window.Deno.core;
+  const webidl = window.__bootstrap.webidl;
+
+  webidl.converters["sequence<USVString>"] = webidl.createSequenceConverter(
+    webidl.converters["USVString"],
+  );
+  webidl.converters.AbortSignal = webidl.createInterfaceConverter(
+    "AbortSignal",
+    AbortSignal,
+  );
```
bnoordhuis, line 14: Duplicates `extensions/web/03_abort_signal.js`.

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,302 @@
...
+    "WebSocketStreamOptions",
+    [
+      {
+        key: "protocols",
+        converter: webidl.converters["sequence<USVString>"],
+        defaultValue: [],
+      },
+      {
+        key: "signal",
+        converter: webidl.converters.AbortSignal,
```
bnoordhuis, line 25: Not 100% sure but shouldn't that be:
```suggestion
        converter: webidl.createNullableConverter(
          webidl.converters.AbortSignal),
```

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,302 @@
...
+    constructor(url, options) {
+      const prefix = "Failed to construct 'WebSocketStream'";
+      webidl.requiredArguments(arguments.length, 1, { prefix });
+      url = webidl.converters.USVString(url, {
+        prefix,
+        context: "Argument 1",
+      });
+      options = webidl.converters.WebSocketStreamOptions(options, {
+        prefix,
+        context: "Argument 1",
```
bnoordhuis, line 129:
```suggestion
        context: "Argument 2",
```

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,302 @@
...
+
+      this[_url] = wsURL.href;
+
+      core.opSync("op_ws_check_permission", this[_url]);
+
+      if (
+        options.protocols.some((x) =>
+          options.protocols.indexOf(x) !== options.protocols.lastIndexOf(x)
+        )
+      ) {
```
bnoordhuis, line 156:
```suggestion
      if (options.protocols.length !== new Set(options.protocols).length) {
```
Not that it really matters here but it's a little more legible and not accidentally quadratic.

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,302 @@
...
+        throw new DOMException(
+          "Can't supply multiple times the same protocol.",
+          "SyntaxError",
+        );
+      }
+
+      core.opAsync("op_ws_create", {
+        url: this[_url],
+        protocols: options.protocols.join(", "),
```
bnoordhuis, line 165: I realize `01_websocket.js` does this too but is this susceptible to header injection attacks with inputs like `{ protocols: ["foo", "\r\nx-pwned: true"] }`?

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,302 @@
...
+                rid: this[_rid],
+                kind: "text",
+                text: chunk,
+              });
+            } else if (chunk instanceof Uint8Array) {
+              await core.opAsync("op_ws_send", {
+                rid: this[_rid],
+                kind: "binary",
+              }, chunk);
+            }
```
bnoordhuis, line 222:
```suggestion
            } else {
              // throw TypeError
            }
```
Or is there something that guarantees it's never anything else? (Apart from the `.d.ts` file, I mean.)

bnoordhuis (approved):
Some comments but mostly LGTM. Good set of regressions tests, too.

I'm trying to understand your remark about leaking ops. Can you go in more detail?

cli/dts/lib.deno.unstable.d.ts
crowlKats: nope, just `string[]` for WebSocketStream. The IDL is described in the first resource link

extensions/websocket/02_websocketstream.js
lucacasonato: I assume hyper/tungsenite sanitizes this. Should probably test that though.

extensions/websocket/02_websocketstream.js
crowlKats: This was suggested in an earlier review by @aapoalas. the same code is used in our normal websocket. it would make sense to leave such improvements to when we rework websocket to be spec conformant.

crowlKats:
> I'm trying to understand your remark about leaking ops. Can you go in more detail?

Well when running the tests they fail due to leaking async ops. i have no clue what exactly the cause is, buut the only async stuff that i can see that could be causing this is the readable stream loop. i guess we somehow have to close the streams when `close` (or any other method to close the websocket) is called

extensions/websocket/02_websocketstream.js
crowlKats: from looking at the `http` crate, this is the code that is used to check every byte in header values:
```rust
#[inline]
fn is_valid(b: u8) -> bool {
    b >= 32 && b != 127 || b == b'\t'
}
```
so we should be fine

cli/dts/lib.deno.unstable.d.ts
bnoordhuis: I had some trouble finding the idl file in the chromium source tree but I finally dug it up and you're right (and it's weirdly inconsistent with the `WebSocket` constructor, IMO, but that aside):

https://chromium.googlesource.com/chromium/src/+/master/third_party/blink/renderer/modules/websockets/websocket_stream_options.idl

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,300 @@
...
+      },
+      {
+        key: "signal",
+        converter: webidl.converters.AbortSignal,
+        defaultValue: undefined,
+      },
+    ],
+  );
+  webidl.converters.WebSocketCloseInfo = webidl.createDictionaryConverter(
+    "WebSocketStreamOptions",
```
lucacasonato, line 27:
```suggestion
    "WebSocketCloseInfo",
```

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,300 @@
...
+              case "string": {
+                controller.enqueue(value);
+                break;
+              }
+              case "binary": {
+                controller.enqueue(value);
+                break;
+              }
+              case "ping": {
+                core.opAsync("op_ws_send", {
```
lucacasonato, line 183:
```suggestion
                await core.opAsync("op_ws_send", {
```

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,300 @@
...
+                this[_closed].reject(err);
+                controller.error(err);
+                tryClose(this[_rid]);
+                break;
+              }
+            }
+          },
+          cancel: (reason) => this.close(reason),
+        });
+        const writable = new WritableStream({
```
lucacasonato, line 205: Add a `start` to this and `ReadableStream` where you close the streams when `this[_closed]` resolves.

extensions/websocket/02_websocketstream.js
crowlKats: No, the property isnt nullable

lucacasonato (changes requested):
`WebSocketStream` global should only be defined in unstable.

ry (commented):
Can any of the WPT be enabled?
https://github.com/web-platform-tests/wpt/tree/master/websockets/stream/tentative

lucacasonato:
@ry Still waiting on https://github.com/denoland/deno/issues/10312

cli/tests/websocketstream_test.ts
```diff
@@ -0,0 +1,123 @@
...
+  assertThrows(() => ws.close({ code: 1000, reason: "".padEnd(124, "o") }));
+  ws.close();
+  await ws.closed;
+});
+
+Deno.test("echo string", async () => {
+  const ws = new WebSocketStream("ws://localhost:4242");
+  const { readable, writable } = await ws.connection;
+  await writable.getWriter().write("foo");
+  const res = await readable.getReader().read();
```
piscisaureus, line 75: The "leaking" promise is created here.
`read()` returns a promise which never resolves and doesn't reject either.

piscisaureus: It's actually fairly easy to replicate in the CLI as well:


```js
D:\deno>target\debug\deno.exe
Deno 1.9.2
exit using ctrl+d or close()
>   const ws = new WebSocketStream("ws://localhost:4242");
undefined
>   const { readable, writable } = await ws.connection;
undefined
>   const uint = new Uint8Array([102, 111, 111]);
undefined
>   await writable.getWriter().write(uint);
undefined
>   const res = await readable.getReader().read();
... hangs ...
```

cli/tests/websocketstream_test.ts
```diff
@@ -0,0 +1,123 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// deno-lint-ignore-file no-undef
```
magurotuna, line 2: Now the linter doesn't run `no-undef` (due to #10652) so we can remove this directive.

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,371 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+"use strict";
+
+/// <reference path="../../core/internal.d.ts" />
+
+((window) => {
+  const core = window.Deno.core;
+  const webidl = window.__bootstrap.webidl;
+  const { writableStreamClose, Deferred } = window.__bootstrap.streams;
```
littledivy, line 9:
```suggestion
  const { writableStreamClose, Deferred } = window.__bootstrap.streams;
  const { DOMException } = window.__bootstrap.domException;
```

runtime/js/99_main.js
```diff
@@ -339,6 +351,7 @@ delete Object.prototype.__proto__;
     URL: util.nonEnumerable(url.URL),
     URLSearchParams: util.nonEnumerable(url.URLSearchParams),
     WebSocket: util.nonEnumerable(webSocket.WebSocket),
+    WebSocketStream: util.nonEnumerable(webSocket.WebSocketStream),
```
lucacasonato, line 354: Can you change this to only be registered when `--unstable` is passed? I think `typeof WebSocketStream  === "undefined"` should be true when user does not pass `--unstable`.

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,403 @@
...
+    TypeError,
+  } = window.__bootstrap.primordials;
+
+  webidl.converters.WebSocketStreamOptions = webidl.createDictionaryConverter(
+    "WebSocketStreamOptions",
+    [
+      {
+        key: "protocols",
+        converter: webidl.converters["sequence<USVString>"],
+        defaultValue: [],
```
lucacasonato, line 31:
```suggestion
        get defaultValue() {
          return [];
        },
```

Otherwise the _exact_ same array will be used each time (identity would be the same).

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,403 @@
...
+    }
+  }
+
+  const _rid = Symbol("[[rid]]");
+  const _url = Symbol("[[url]]");
+  const _connection = Symbol("[[connection]]");
+  const _closed = Symbol("[[closed]]");
+  const _closing = Symbol("[[closing]]");
+  const _earlyClose = Symbol("[[earlyClose]]");
+  class WebSocketStream {
```
lucacasonato, line 73: The class is missing branding and brand checks.

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,403 @@
...
+          "This operation was aborted",
+          "AbortError",
+        );
+        this[_connection].reject(err);
+        this[_closed].reject(err);
+      } else {
+        const abort = () => {
+          core.close(cancelRid);
+        };
+        options.signal?.addEventListener("abort", abort);
```
lucacasonato, line 144: Instead of using `addEventListener` use the `abortSignal.add` method to add an internal listener. The "abort" event could be dispatched manually by a user.

extensions/websocket/02_websocketstream.js
```diff
@@ -0,0 +1,403 @@
...
+        PromisePrototypeThen(
+          core.opAsync("op_ws_create", {
+            url: this[_url],
+            protocols: options.protocols
+              ? ArrayPrototypeJoin(options.protocols, ", ")
+              : "",
+            cancelHandle: cancelRid,
+          }),
+          (create) => {
+            options.signal?.removeEventListener("abort", abort);
```
lucacasonato, line 154: Use `abortSignal.remove` here. (see comment above)

bartlomieju (commented):
Please update `UNSTABLE_DENO_PROPS` in `cli/diagnostics.rs`

lucacasonato (approved):
LGTM. Thanks @crowlKats for seeing this through.

API is --unstable for now as spec has not yet been finalized.

sanfusu:
> @bnoordhuis there are leaking async ops, and not sure how those can be fixed. i think it might have to do with the pull on the readable stream, but not sure. that aside, it would be ready

Does this problem still exist? I have some problems about leaking async ops, but I'm not sure if it's related to websocket.

Issue #9344, opened 2021-01-31
rename --inspect-brk to --inspect-break


timreichen:
I was wondering if the deno run option `--inspect-brk` should be renamed to `--inspect-break` since non of the other flags are abbreviated and it would be more expressive. I think 13 over 15 chars does not make the difference.
Maybe it could be aliased and then `--inspect-brk` could be dropped with the 2.0 release?

kitsonk:
While it is poorly named `--inspect-brk` is aligned to other Chrome based inspectors.  My opinion is changing this breaks more expectations than it fixes.  We should have wider consensus before #10366 is merged.  @ry @bartlomieju @lucacasonato thoughts?

lucacasonato:
I would prefer to keep current naming. It aligns with node, and is already widely known.

timreichen:
@lucacasonato while it is true that it aligns with node, this proposal is more about the aesthetics of cli, which makes no functional difference but I think is really important as well and deno has the opportunity to be aesthetically more pleasing than node.
Even if someone coming from node types `--inspect-brk`, cli will probably throw `Did you mean --inspect-break` no?

jsejcksn:
```
deno run --👀-✋
```
😅

PR #7032, opened 2020-08-12
chore: Rename --importmap to --import-map

nayeemrmn:
Closes #6976.

PR #10366, opened 2021-04-25
rename(cli): --inspect-brk to --inspect-break

timreichen:
ref: https://github.com/denoland/deno/issues/9344


cli/flags.rs
```diff
@@ -1022,7 +1022,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
         .takes_value(true)
         .requires("unstable")
         .conflicts_with("inspect")
-        .conflicts_with("inspect-brk")
+        .conflicts_with("inspect-break")
```
caspervonb, line 1025: Not aliased? so breaking change.

caspervonb (changes requested):
Am indifferent as to if we should land this or not, I'll punt that to the others but it has to be a non-breaking change to land for 1.x.

cli/flags.rs
timreichen: I am not too familiar with rust and would need help on how exactly to implement aliases. Else this could also wait till 2.x

nayeemrmn: See https://github.com/denoland/deno/pull/7032/files#diff-abdd2d887e97d0eff6d352a1efd4dd953797699228a18a3a237d9818469da3cfR1326 for an example.

kitsonk:
Yes, this needs to be a non-breaking change to be considered, even then `--inspect-brk` while poorly named, reflects the same flag across all the Chrome based inspectors, so it matches a wider set of expectations.

cli/flags.rs
timreichen: added alias

bartlomieju:
Discussed with the rest of the team, and I whole-heartedly agree with @kitsonk's [comment](https://github.com/denoland/deno/issues/9344#issuecomment-826394772). This flag name was chosen before 1.0 specifically to match flag name in Node and Chrome. 

So thank you @timreichen for the PR, but I'm going to close it without a merge.

Issue #10362, opened 2021-04-25
Import completions do not work on export statements


lucacasonato:
`import {} from "https://deno.land/x/";` completes fine, `export {} from "https://deno.land/x/";` does not.

satyarohith:
I'm working on it.

PR #10369, opened 2021-04-25
fix(lsp): enable import completions on export statements

satyarohith:
Closes https://github.com/denoland/deno/issues/10362


kitsonk (approved):
LGTM... Thanks for fixing it.

PR #10374, opened 2021-04-25
fix(cli/ast): Exclude dynamic imports from static dep analysis

### LINKED ISSUE #10168: v1.8.3 - Dynamic import that can't execute causes error when using import maps.
```js
if (never) {
  await import('whatever')
}
```
I get the `unmapped bare specifier` error. But this should not cause any error or else it will be impossible to write code which supports several runtimes.

Basically I want to check if it's running in Node.js and then to load a Node.js specific module ONLY in that case. When running in Deno or the browser then I don't need to do this, hence Deno should not complain about the code in the block which is never executed.

nayeemrmn:
Closes #10168. I've excluded the LSP from this change, please comment.

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -275,7 +275,7 @@ fn strip_config_from_emit_options(
 /// processing.
 #[derive(Clone)]
 pub struct ParsedModule {
-  comments: SingleThreadedComments,
+  pub comments: SingleThreadedComments,
   leading_comments: Vec<Comment>,
   pub module: Module,
   pub source_map: Rc<SourceMap>,
@@ -295,7 +295,10 @@ impl fmt::Debug for ParsedModule {
 impl ParsedModule {
   /// Return a vector of dependencies for the module.
   pub fn analyze_dependencies(&self) -> Vec<DependencyDescriptor> {
-    analyze_dependencies(&self.module, &self.source_map, &self.comments)
+    let mut descs =
+      analyze_dependencies(&self.module, &self.source_map, &self.comments);
+    descs.retain(|desc| !desc.is_dynamic);
+    descs
   }
 
   /// Get the module's leading comments, where triple slash directives might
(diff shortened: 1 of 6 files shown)


kitsonk:
Why the approach to exclude them?  This will break type checking if the import is available, won't it?

nayeemrmn:
The type checking would happen at runtime for any dynamic import, as it should do regardless of syntax. i.e.
```ts
// temp.ts
console.log(1);
await import("./temp2.ts");

// temp2.ts
const a: number = "";
```
will behave like:
```ts
// temp.ts
console.log(1);
await import("./" + "temp2.ts"); // No static type errors.

// temp2.ts
const a: number = "";
```
There's no reason for the first case to be more "clever". For example, the ESM spec doesn't do static syntax checks of dynamic modules depending on how the specifier is passed, even though the same justifications would apply. It's a simpler and more consistent model to implement.

Anyway this ship sailed already with what we decided for bundling. Currently `deno bundle temp.ts` for the first case throws a type error even though the offending module isn't included in the bundle.

kitsonk:
This just feels like we are hacking around the problem though, especially because we are effectively "unsharing" code for the LSP instead of sharing code.  How does this impact `deno info`?  Yes, there were always dynamic imports that were statically analysed, but the module graph has the concept of dependency slots now, for the very specific reason that it can hold onto references that can't be statically resolved, but should only surface the errors when the module is actually needed.

nayeemrmn:
> How does this impact `deno info`?

@kitsonk Under this patch `deno info` now excludes all dynamic imports equally. The same arguments apply, it's a lot cleaner if dynamic imports aren't statically detected depending on how they are written. And if the dependency itself is conditional, IMO it's not correct to include it.

> the module graph has the concept of dependency slots now, for the very specific reason that it can hold onto references that can't be statically resolved, but should only surface the errors when the module is actually needed.

If it's possible to get the same API as this patch while secretly doing as much static loading as possible and only surfacing errors when the import happens at runtime, SGTM. I don't know of any dynamic import errors that currently exhibit this though, do we really have a subsystem that works this way? Otherwise any implementation feedback is appreciated, but the user-facing change is a long time coming.

nayeemrmn:
@kitsonk You were right, a natural fix for #10615 fixed #10168 as well. I'll open a separate issue for no early type-checking of dynamic imports.

Closing in favour of #10618.

PR #10377, opened 2021-04-26
feat(net): allow unix sockets on windows

### LINKED ISSUE #10244: Support for Named Pipes in Deno.connect
Currently, `Deno.connect` only supports IPC for unix (using unstable `transport: "unix"`). There should be named pipes support too, which would allow IPC on Windows (as NodeJS' `net` mode does).

AaronO:
Given that Windows supports unix sockets [since late 2017](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/), this PR simply removes the conditional compilation around unix sockets which should now build and work on all supported platforms.

Should fix #10244

Blocked on `tokio`/`mio` support for Windows UDS https://github.com/tokio-rs/tokio/issues/2201

### DIFF
--- a/runtime/ops/net.rs
+++ b/runtime/ops/net.rs
@@ -35,11 +35,8 @@ use trust_dns_resolver::config::ResolverOpts;
 use trust_dns_resolver::system_conf;
 use trust_dns_resolver::AsyncResolver;
 
-#[cfg(unix)]
 use super::net_unix;
-#[cfg(unix)]
 use crate::ops::io::UnixStreamResource;
-#[cfg(unix)]
 use std::path::Path;
 
 pub fn init(rt: &mut deno_core::JsRuntime) {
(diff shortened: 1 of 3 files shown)


runtime/ops/net.rs
```diff
@@ -64,9 +61,7 @@ pub struct OpConn {
 pub enum OpAddr {
   Tcp(IpAddr),
   Udp(IpAddr),
-  #[cfg(unix)]
   Unix(net_unix::UnixAddr),
-  #[cfg(unix)]
   UnixPacket(net_unix::UnixAddr),
```
lucacasonato, line 65: Unixpacket is not supported on Windows.

AaronO: Thanks, I suspected that might not be the case. I don't think I've seen `unixpacket` used anywhere TBH (not just deno, but any software project)

lucacasonato: Probably going to get a lot more common with QUIC / HTTP/3.

PR #10384, opened 2021-04-26
fix(runtime/js/http): Expose nextRequest() errors in respondWith()

### LINKED ISSUE #10380: [HTTP Native] Process ended because `nextRequest` threw error
Stacktrace:

```
                           ^
    at async HttpConn.nextRequest (deno:runtime/js/40_http.js:33:23)
    at async Object.next (deno:runtime/js/40_http.js:83:26)
```

deno 1.9.0 (release, x86_64-apple-darwin)
v8 9.1.269.5
typescript 4.2.2

nayeemrmn:
Fixes #10380.

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -272,3 +272,48 @@ unitTest(
     await promise;
   },
 );
+
+unitTest(
+  { perms: { net: true } },
+  async function httpServerNextRequestErrorExposedInResponse() {
+    const promise = (async () => {
+      const listener = Deno.listen({ port: 4501 });
+      const conn = await listener.accept();
+      const httpConn = Deno.serveHttp(conn);
+      const event = await httpConn.nextRequest();
+      assert(event);
+      // Start polling for the next request before awaiting response.
+      const nextRequestPromise = httpConn.nextRequest();
+      const { respondWith } = event;
+      await assertThrowsAsync(
+        async () => {
+          let interval = 0;
+          await respondWith(
+            new Response(
+              new ReadableStream({
+                start(controller) {
+                  interval = setInterval(() => {
+                    const message = `data: ${Date.now()}\n\n`;
+                    controller.enqueue(new TextEncoder().encode(message));
+                  }, 200);
+                },
+                cancel() {
+                  clearInterval(interval);
+                },
+              }),
+            ),
+          );
+        },
+        Deno.errors.Http,
+        "connection closed",
+      );
+      // The error from `op_http_request_next` reroutes to `respondWith()`.
+      assertEquals(await nextRequestPromise, null);
+      listener.close();
+    })();
+
+    const resp = await fetch("http://127.0.0.1:4501/");
+    await resp.body!.cancel();
+    await promise;
+  },
+);
--- a/runtime/js/40_http.js
+++ b/runtime/js/40_http.js
@@ -14,6 +14,8 @@
     return new HttpConn(rid);
   }
 
+  const connErrorSymbol = Symbol("connError");
+
   class HttpConn {
     #rid = 0;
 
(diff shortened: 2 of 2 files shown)


cli/tests/unit/http_test.ts
```diff
@@ -272,3 +272,48 @@ unitTest(
...
+        async () => {
+          let interval = 0;
+          await respondWith(
+            new Response(
+              new ReadableStream({
+                start(controller) {
+                  interval = setInterval(() => {
+                    const message = `data: ${Date.now()}\n\n`;
+                    controller.enqueue(new TextEncoder().encode(message));
+                  }, 200);
```
ry, line 298: It's not clear to me whether this setInterval loop is necessary to demonstrate the bug? Can't you just setTimeout and respondWith after?

nayeemrmn: This is just a generic streamed response copied from an above test, any endless streamed response will do.

runtime/js/40_http.js
```diff
@@ -172,14 +185,21 @@
...
               throw error;
             }
           }
         } finally {
           // Once all chunks are sent, and the request body is closed, we can
           // close the response body.
-          await Deno.core.opAsync("op_http_response_close", responseBodyRid);
+          try {
+            await Deno.core.opAsync("op_http_response_close", responseBodyRid);
+          } catch { /* pass */ }
```
aapoalas, line 202: The ignored catch could perhaps use a more informative comment on why it can be passed.

nayeemrmn: It's because it may already be closed. We have this in other places, a comment is there at all because of the linter.

runtime/js/40_http.js
```diff
@@ -144,6 +152,11 @@
           innerResp.headerList,
         ], respBody instanceof Uint8Array ? respBody : null);
       } catch (error) {
+        const connError = connErrors.get(httpConn);
```
aapoalas, line 155: For this and the below line, I am not exactly 100% on this but I think (for normal Map at least) it is more performant to check membership and only get if true.
```javascript
if (error instanceof errors.BadResource && connErrors.has(httpConn)) {
  error = new connError.constructor(connErrors.get(httpConn).message);
}
```

runtime/js/40_http.js
```diff
@@ -35,10 +37,16 @@
           this.#rid,
         );
       } catch (error) {
+        // A connection error seen here would cause disrupted responses to throw
+        // a generic `BadResource` error. Instead store this error and replace
+        // those with it.
+        connErrors.set(this, error);
```
aapoalas, line 43: If you're confident that this is indeed an Error object, then I think you could just save the message part of it since that's the only part that actually gets used down the line.

nayeemrmn: The `.constructor` as well. Conceptually I'm cloning the error under a new call stack, it's simpler like this.

aapoalas: Aaah, oh true yeah. I missed the `.constructor` usage. Yeah, saving the error object itself is good.

hsjoberg:
FYI I can confirm that this pull request fixes the crashes I was having regrading the error message in question.

ry:
deferring to @kt3k for final review

runtime/js/40_http.js
```diff
@@ -14,6 +14,8 @@
     return new HttpConn(rid);
   }
 
+  const connErrors = new WeakMap();
```
kt3k, line 17: Do you have any reason to use WeakMap over a simple property like httpConn.error (`Error | undefined`)?

kt3k (approved):
LGTM with one question.

runtime/js/40_http.js
nayeemrmn: To properly hide it from the public API.

lucacasonato: Then use a private property or symbol property.

lucacasonato: We use symbols for essentially all web APIs now. Should be fine.

nayeemrmn: Replaced with a symbol.

kt3k (approved):
@nayeemrmn Thanks for the fix! LGTM

PR #10385, opened 2021-04-26
refactor(core): simplify module loading code

bartlomieju:
General cleanup of module loading code, tried to reduce indentation in various methods
on "JsRuntime" to improve readability. 

Added "JsRuntime::handle_scope" helper function, which returns a "v8::HandleScope".
This was done to reduce a code pattern that happens all over the "deno_core".

Additionally if event loop hangs during loading of dynamic modules a list of 
currently pending dynamic imports is printed.


core/runtime.rs
```diff
@@ -320,6 +322,11 @@ impl JsRuntime {
     self.v8_isolate.as_mut().unwrap()
   }
 
+  pub fn handle_scope(&mut self) -> v8::HandleScope {
+    let context = self.global_context();
+    v8::HandleScope::with_context(self.v8_isolate(), context)
+  }
```
ry, line 328: I think this should be an internal helper method (`pub(crate)`) rather than part of the public interface. Usually callers should not be interacting with raw V8 HandleScopes.

core/runtime.rs
```diff
@@ -1111,89 +1115,81 @@ impl JsRuntime {
...
+
+    let promise = module_evaluation.promise.get(scope);
+    let mut sender = module_evaluation.sender.clone();
+    let promise_state = promise.state();
+
+    match promise_state {
+      v8::PromiseState::Pending => {
+        // NOTE: `poll_event_loop` will decide if
+        // runtime would be woken soon
+        state_rc.borrow_mut().pending_mod_evaluate = Some(module_evaluation);
```
ry, line 1136: It seems odd to me that you call `pending_mod_evaluate.take()` only to return the value here. Is there something wrong with the previous `if let Some(...` construction?

core/runtime.rs
bartlomieju: Agreed, and yet benchmarks are interacting - I think it's better than having to use the code from this function directly

core/runtime.rs
bartlomieju: There were previously two calls to `pending_mod_evaluate.take()`, I believe this version is simpler

ry (approved):
LGTM

piscisaureus (approved):
Seems maybe not that different from before, but OK if you think it helps.

PR #10398, opened 2021-04-27
refactor(core): return async ops in batches

### LINKED ISSUE #10361: RangeError: Maximum call stack size exceeded  when creating too many op promises in a single loop
I think i found a bug running a debouncer 60709 times or more.  
Looks like the clearTimeout it runs into a RangeError.

checkout the gist:
https://gist.github.com/jupegarnica/cb3a28a3a3e7f40363c14bd61abe9880

It fails at core.js at line 69 on  handleAsyncMsgFromRust
https://github.com/denoland/deno/blob/fd3b961126150960104a0d456ceac14b12f90c8b/core/core.js#L69

AaronO:
After removing the shared queue we returned async op responses to JS via an unbound arg list, though very few programs should hit that limit, the limit still exists (it's dependant on the stack size limit)

Should fix #10361

## Notes

- Batch size is 4k (can be changed obviously)
- Might want to pre-allocate and reuse response buffer vec across event loops

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -528,8 +528,7 @@ impl JsRuntime {
 
     // Ops
     {
-      let async_responses = self.poll_pending_ops(cx);
-      self.async_op_response(async_responses)?;
+      self.drain_async_ops(cx)?;
       self.drain_macrotasks()?;
       self.check_promise_exceptions()?;
     }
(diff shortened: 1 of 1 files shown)


ry (commented):
4k seems quite large - I would imagine it would fail before that.

I wonder if we could construct a test to demonstrate this failure. I'm not sure how this would look, but start many async ops, and we should see it crash when they resolve?

bartlomieju:
IIRC V8 has a limit of about 65k arguments for a function call. Still 4k seems like a huge limit.

AaronO:
> I wonder if we could construct a test to demonstrate this failure. I'm not sure how this would look, but start many async ops, and we should see it crash when they resolve?

Sure we can, I'll add one.

> 4k seems quite large - I would imagine it would fail before that.
>
> IIRC V8 has a limit of about 65k arguments for a function call. Still 4k seems like a huge limit.

I agree most event loops will have <1k resolved ops, so even if it's uncommon we should still handle it graciously.

What are your specific concerns about the 4k limit ?:
1. That the vector alloc adds too much overhead ? (I benched it and no significant delta, one "large" alloc is quite cheap)
2. That instead of returning them all in this event loop we should do at most one batch per loop ?
  - This would have the upside of letting I/O and other leaf futures advance
  - But would have the downside of maybe breaking other promises, that things should be executed precisely in the next event loop
  
Ultimately we do have a real upper-bound, it's best to handle explicitly than have implicit failures under stress or in stress cases.

I'll also investigate the timer code, because I think it can be optimized there but I think that's a separate topic and we should still improve the correctness of core

ry:
My only concern about this patch is that it's fixing a bug without a test. Otherwise looks good.

bartlomieju:
I'm going to close this PR as it has conflicts with existing code in `core/runtime.rs` and it's still missing a test case. Even though it looks innocent I'm afraid it might cause some inadvert regressions. Feel free to reopen if we can make sure it works as expected.

PR #10401, opened 2021-04-28
ci: build and test only selected targets

kt3k:
Currently we specify `--all-targets` when building. It's equivalent of `--lib --bins --tests --benches --examples` (ref: https://doc.rust-lang.org/cargo/commands/cargo-build.html#target-selection ), but in test release jobs, we don't need to build everything. So this PR reduces build target to only `--bin deno --bin test_server` in build phase, and reduces test targets to ` --bins --lib --tests`.

This skips the building of benches and examples in test release jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,7 +192,7 @@ jobs:
           path: |
             ./target
           key:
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - this should save some link time, nice work.

PR #10402, opened 2021-04-28
feat: add FsWatcher type

### LINKED ISSUE #9399: watchFs is not abortable
Right now there is no way to cancel a watch on the file system.

This function will never return and is not abortable from the outside:
```typescript
async function* onDbChange(dbPath: string) {
  console.log(`Watching for changes to ${dbPath}...`);
  const watcher = Deno.watchFs(dbPath);
    if (event.kind === 'modify') {
      yield;
    }
  }
}
```

Possible solution would be to pass an `AbortSignal`:

```typescript
Deno.watchFs(dbPath, { signal });
```

kt3k:
In the type definition `Deno.watchFs` returns `AsyncIterableIterator<FsEvent>`, but it actually returns `FsWatcher` class and has more feature than `AsyncIterableIterator` (for example, it supports closing the watcher).

This PR add FsWatcher type in `lib.deno.ns.d.ts` to fix this situation, and also documents how to close the watcher.

closes #9399

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1925,6 +1925,12 @@ declare namespace Deno {
    */
   export function resources(): ResourceMap;
 
+  export class FsWatcher implements AsyncIterableIterator<FsEvent> {
+    readonly rid: number;
+    next(): Promise<IteratorResult<FsEvent>>;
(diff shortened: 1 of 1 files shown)


kt3k:
It seems `watcher.return()` is the intended way to close the watcher according to the test case.

PR #10403, opened 2021-04-28
docs: document how to stop file watcher

### LINKED ISSUE #9399: watchFs is not abortable

kt3k:
This PR documents how to stop the file watcher.

closes #9399

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1947,6 +1947,16 @@ declare namespace Deno {
    *```
    *
    * Requires `allow-read` permission.
+   *
+   * Call `watcher.return()` to stop watching.
+   *
+   * ```ts
+   * const watcher = Deno.watchFs("/");
+   * setTimeout(() => { watcher.return(); }, 5000);
+   * for await (const event of watcher) {
+   *    console.log(">>>> event", event);
+   * }
+   * ```
    */
   export function watchFs(
     paths: string | string[],


ry (approved):
LGTM - tho we should probably change this behavior to match the async iterators elsewhere.

kt3k:
`.return()` seems a part of async iterator definition according to our dts file https://github.com/denoland/deno/blob/bf99300/cli/dts/lib.es2018.asynciterable.d.ts#L35

jsejcksn:
> LGTM - tho we should probably change this behavior to match the async iterators elsewhere.

@ry While what @kt3k said is true, did you mean `watcher.close()`? Like here, for example?

https://doc.deno.land/builtin/stable#Deno.Listener

ry:
@jsejcksn Yes, we should probably update the style guide to say something about how the `return()` method in AsyncIterators should not be used to close resources...

Unfortunately changing this in FsWatcher is a breaking change, so we'd need to deprecate the current behavior and phase it out before.

PR #10406, opened 2021-04-28
fix(op_crate/fetch): infinite loop on fill headers

### LINKED ISSUE #10396: Fetch OOMs with init param and setting headers
The following script hangs for a bit and before crashing with an OOM error from v8

```javascript
const req = new Request("https://example.org");
const init = {
  method: "GET",
};
req.headers.set("foo", "bar");
const res = await fetch(req, init);
```
```text
<--- Last few GCs --->

[29004:0x379f00000000]    31360 ms: Scavenge 1397.6 (1439.0) -> 1397.1 (1439.0) MB, 0.2 / 0.0 ms  (average mu = 0.141, current mu = 0.091) allocation failure 
[29004:0x379f00000000]    31361 ms: Scavenge 1397.6 (1439.0) -> 1397.1 (1439.0) MB, 0.2 / 0.0 ms  (average mu = 0.141, current mu = 0.091) allocation failure 
[29004:0x379f00000000]    31362 ms: Scavenge 1397.8 (1439.3) -> 1397.4 (1439.3) MB, 0.2 / 0.0 ms  (average mu = 0.141, current mu = 0.091) allocation failure 


<--- JS stacktrace --->


#
# Fatal javascript OOM in Ineffective mark-compacts near heap limit
#

[1]    29004 trace trap (core dumped)  deno run --allow-net https://deno.wperron.io/fetch-oom@0.1.0/main.ts
```

I made a quick repro at https://github.com/wperron/fetch-oom and can be tested with 
```
deno run --allow-net https://esm.wperron.io/fetch-oom@0.1.0/main.ts
```

wperron:
Fixes a pesky bug in the fetch implementation where if the init part is
specified in `fetch` instead of the `Request` constructor, the
fillHeaders function receives two references to the same object, causing
it to append to the same list being iterated over.

Fixes #10396

### DIFF
--- a/cli/tests/unit/fetch_test.ts
+++ b/cli/tests/unit/fetch_test.ts
@@ -449,6 +449,21 @@ unitTest(
 
 unitTest(
   { perms: { net: true } },
+  async function fetchSeparateInit(): Promise<void> {
+    // related to: https://github.com/denoland/deno/issues/10396
+    const req = new Request("http://localhost:4545/cli/tests/001_hello.js");
+    const init = {
+      method: "GET",
+    };
+    req.headers.set("foo", "bar");
+    const res = await fetch(req, init);
+    assertEquals(res.status, 200);
+    await res.text();
+  },
+);
+
+unitTest(
+  { perms: { net: true } },
   async function fetchInitTypedArrayBody(): Promise<void> {
     const data = "Hello World";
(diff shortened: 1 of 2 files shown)


ry (approved):
I have verified the test fails on main branch.

PR #9815, opened 2021-03-17
feat(cli): run test modules in parallel

caspervonb:
This runs test modules in parallel when the number of concurrent jobs passed via `--jobs=<N>` is greater than 1.
The default is to run with a limit of 1 to retain backwards compatibility, as not all test suites in the wild are concurrency safe.

Closes https://github.com/denoland/deno/issues/6559
Closes https://github.com/denoland/deno/issues/9284

PR #10408, opened 2021-04-28
refactor(cli): move test reporting into a trait

caspervonb:
This moves test reporting into a trait in preparation for supporting more reporters, like json and terse.
The output remains unchanged and there's still room to clean up some small things later.

### DIFF
--- a/cli/tools/test_runner.rs
+++ b/cli/tools/test_runner.rs
@@ -25,6 +25,7 @@ use std::path::PathBuf;
 use std::sync::mpsc::channel;
 use std::sync::mpsc::Sender;
 use std::sync::Arc;
+use std::time::Instant;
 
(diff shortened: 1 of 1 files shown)


caspervonb:
Depends on #9815

cli/tools/test_runner.rs
```diff
@@ -52,6 +53,141 @@ pub enum TestMessage {
   },
 }
 
+trait TestReporter {
+  fn visit_message(&mut self, message: TestMessage);
```
bartlomieju, line 57: Nitpick: would it make sense to take `&TestMessage` instead?

caspervonb: Move is intentional, avoid having to deref every single field.

Okay with changing this, but then I'd prefer to make all reporters pass by ref.

bartlomieju (approved):

PR #10411, opened 2021-04-28
feat(runtime/permissions): add `$PATH` resolution to run permission

crowlKats:
A usecase for this is when using `Deno.execPath()` but passing `--allow-run=deno`. I very well see that there is no need for this as one can just do `--allow-run=$(which deno)`, so it is just a "nice little thing" and I dont mind this not being merged

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -888,13 +888,29 @@ impl Permissions {
     state: &Option<Vec<String>>,
     prompt: bool,
   ) -> UnaryPermission<RunDescriptor> {
+    let path_var = std::env::var_os("PATH");
     UnaryPermission::<RunDescriptor> {
       name: "run",
       description: "run a subprocess",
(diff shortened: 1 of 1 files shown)


caspervonb:
Personally, not really based on anything but gut-feeling I don't like the implicitness of this as my assumption is that --allow-`run=foo` would be the same as `--allow-run=./foo` (point of reference I suppose is that`chmod +x foo` doesn't chmod in $PATH, because that would be weird).

bartlomieju:
I'm also not a fan of this PR, it will change the semantics of the `--allow-run` flag and could be thought of as a breaking change. Is there a bug that this PR fixes?

crowlKats:
@bartlomieju nope, it was really just a spur of the moment idea, but since i am also not a real fan of it either, might as well just close it

bartlomieju:
Let's close it then and revisit if there's a need for it.

PR #10412, opened 2021-04-28
refactor(extensions): reintroduce builder

AaronO:
Builder pattern is more flexible and expressive than the old constructors

Follow-up to #9800


ry (approved):
LGTM

PR #10414, opened 2021-04-28
refactor(extensions): reintroduce declare_ops!() macro

AaronO:
This implementation is both cleaner and more flexible than the original prototype in #9800

## Changes

- [x] Implement `declare_ops!()`
- [x] Adapt all `op_crates`

## Example

```rust
declare_ops!(
  sync[
    op_foo,
    op_bar,
  ],
  async[
    op_write,
    op_read,
  ],
)
```

## Notes

- Users only need to import `deno_core::declare_ops`, they don't need to import the `op_sync` and `op_async` wrappers
- You can have arbitrarily many `sync` / `async` groups


op_crates/webgpu/lib.rs
```diff
@@ -546,326 +545,98 @@ pub fn op_webgpu_create_query_set(
...
-      "op_webgpu_write_texture",
-      op_sync(queue::op_webgpu_write_texture),
-    ),
-    // shader
-    (
-      "op_webgpu_create_shader_module",
-      op_sync(shader::op_webgpu_create_shader_module),
-    ),
-  ]
+  declare_ops!(
```
aapoalas, line 548: Now this is damn neat.

core/extensions.rs
```diff
@@ -126,3 +126,63 @@ macro_rules! include_js_files {
...
+    ),)+]
+  };
+}
+
+/// transforms a stringified identifier path into an op_name
+/// it also enforces that all op_names must start with "op_"
+// stringify!(foo::op_bar<X>) => "foo :: op_bar < X >"
+// op_ident(stringify!(foo::op_bar<X>)) => "op_bar"
+pub fn op_ident(ident_path: &'static str) -> &'static str {
+  let end = ident_path.rfind("::<").unwrap_or_else(|| ident_path.len());
```
ry, line 175: Stuff like this concerns me. 

Imagine this scenario: Suppose this feature was introduced earlier when we hadn't had any generic ops like (`op_ws_create::<P>`). Likely this macro would be implemented without the functionality to handle turbofish. At some point we want to add a generic op, but run into even more complexity because this macro would need to be modified to handle it. Depending on who is trying to add the feature, that may be a roadblock, it may cause that person to solve the problem in some other, suboptimal, way.

These macros are not zero-cost. Sure, you're saving 200 lines of repetitive code, and eliminating the possibility of there being a mismatch between op names and their corresponding rust functions, but at the cost of Deno encoding something about Rust's syntax. Is the turbofish parsing here sufficiently general? Probably, but difficult to say conclusively.

It's not clear this patch is a beneficial trade off.

AaronO: > Stuff like this concerns me.
> 
> Imagine this scenario: Suppose this feature was introduced earlier when we hadn't had any generic ops like (`op_ws_create::<P>`). Likely this macro would be implemented without the functionality to handle turbofish. At some point we want to add a generic op, but run into even more complexity because this macro would need to be modified to handle it. Depending on who is trying to add the feature, that may be a roadblock, it may cause that person to solve the problem in some other, suboptimal, way.

I do get the concern but I don't think it makes sense to penalize a proposal based on a hypothetical past.

It's not an issue at present, we handle all ident paths (`bar`, `foo::bar`, `foo::bar::<T>`) and any potential regression should be caught by tests (same as a manual typo by a dev, that op wouldn't be callable), so there's low/no risk of a silent break moving forward. `op_ident` could also apply an extra sanity check on the op name for extra reassurance (e.g: ensure op names are `snake_case`, or `\w+`).

I think it could also be useful to have a "test" that tracks all ops (in `deno` bin) via:
```
deno eval "console.log(Object.keys(Deno.core.ops()).join('\n'))" | sort
```

This would further guard against any regressions and would have the upside of sanity checking new/removed/renamed ops in large PRs.

---

Do keep in mind that this macro, like `include_js_files!()`, is opt-in. Devs can still fallback to a manual `vec![]` if desired (e.g: for tests/benches where they want to use closures for small ops, which are unnamed). But the current implementation should should cover all our present and foreseeable needs. Do you have an example fail case in mind ? (I just thought of a potential one whilst writing this out, but a very easy fix ...)

> These macros are not zero-cost. Sure, you're saving 200 lines of repetitive code, and eliminating the possibility of there being a mismatch between op names and their corresponding rust functions, but at the cost of Deno encoding something about Rust's syntax. Is the turbofish parsing here sufficiently general? Probably, but difficult to say conclusively.

Being slightly coupled to rust's syntax in a few places isn't a big negative IMO, we already couple to rust internals for utilities like `GothamState`, rust's syntax is unlikely to change and if it were to change that break would be caught by our tests and we would probably have larger breaking changes if that were to happen

bnoordhuis: FWIW, I'm also not a big fan of this macro but for a different reason: it creates a domain-specific language that new contributors need to learn. It raises the barrier to entry.

Low complexity verbosity is easier to grok than high complexity terseness. I'd err on the side of keeping it as simple as possible.

AaronO: @ry @bnoordhuis Is there any variation of this that would seem like a fair tradeoff ? Do you think the present approach can be improved upon ? If so, how ?

bnoordhuis: Speaking for myself, I feel the existing (non-macro) approach is already pretty good.

crowlKats: I think this macro makes only sense for things like webgpu where we have like 80 ops declarations, but since thats a single case, I also think there is no real need for it.

bartlomieju: Ops declarations is not something that should (will) change often so there's no problem with some level of duplication in favor of simpler solution.

ry:
Closing without merge. Although the macro saves lines of code, overall we feel it increases complexity. 

We appreciate the work. Sorry we couldn’t agree on merging.

PR #10416, opened 2021-04-29
feat(op_crates/fetch): add send request hook

bnoordhuis:
Add a SendRequest::send() hook that gives downstream users of the fetch
crate more control over the HTTP request that's sent out.

<hr>

Conversation starter to collect feedback, maybe someone has a better idea? The problems I'm trying to solve:

1. Add/modify/validate request headers (works)

2. Handle DNS myself instead of delegating it to reqwest (partial)

I can tell `reqwest` what IP address to connect to by changing the request URI but I can't tell it to verify the TLS certificate against a host name instead of the address, even when I plug in a `rustls::ClientConfig` with a custom verifier through `reqwest::ClientBuilder::use_preconfigured_tls()`. 🤷

edit: turns out to be a hyper-rustls shortcoming, it always fails with an "invalid dnsname" when `uri.host()` is an IP address...

### DIFF
--- a/op_crates/fetch/lib.rs
+++ b/op_crates/fetch/lib.rs
@@ -96,6 +96,21 @@ pub fn init(isolate: &mut JsRuntime) {
   }
 }
 
+pub type SendRequestFuture =
+  Pin<Box<dyn Future<Output = Result<reqwest::Response, AnyError>>>>;
+
+pub trait SendRequest {
+  fn send(
+    &mut self,
+    _client: reqwest::Client,
+    builder: reqwest::RequestBuilder,
+  ) -> SendRequestFuture {
+    Box::pin(async move { builder.send().await.map_err(|err| err.into()) })
+  }
+}
+
+impl SendRequest for () {}
+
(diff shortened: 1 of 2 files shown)


op_crates/fetch/lib.rs
```diff
@@ -96,6 +96,21 @@ pub fn init(isolate: &mut JsRuntime) {
...
+  Pin<Box<dyn Future<Output = Result<reqwest::Response, AnyError>>>>;
+
+pub trait SendRequest {
+  fn send(
+    &mut self,
+    _client: reqwest::Client,
+    builder: reqwest::RequestBuilder,
+  ) -> SendRequestFuture {
+    Box::pin(async move { builder.send().await.map_err(|err| err.into()) })
+  }
```
bnoordhuis, line 109: This default implementation doesn't use the `reqwest::Client` but it comes in handy when modifying the request post facto:
```rust
let mut req = builder.build()?;
req.uri_mut().set_ip_host(ip)?;
let res = client.execute(req).await?;
Ok(res)
```

PR #10423, opened 2021-04-29
refactor(core.js): provide window.__bootstrap

AaronO:
So all other extension JS code can rely on it being initialized by `core.js`


core/core.js
```diff
@@ -124,6 +124,9 @@
     opSync("op_close", rid);
   }
 
+  // Provide bootstrap namespace
+  window.__bootstrap = {};
```
ry, line 128: It would be cool if somehow we could delete __bootstrap in core after all the extensions have been loaded. That would remove the need for the caller to delete something provided by core.

ry (commented):
This certainly simplifies things, but I'm worried its setting in stone a loose internal module pattern that we're not particularly convinced of.

core/core.js
AaronO: If we convert everything to extensions and don't use `__bootstrap` outside of extensions then it's very easy to do, but for the moment `runtime` / `cli` ops are still registered imperatively (will convert them in a follow up PR)

AaronO:
> This certainly simplifies things, but I'm worried its setting in stone a loose internal module pattern that we're not particularly convinced of.

Why do you feel we would be anymore locked into this pattern after this change than before ?

Over the past few releases we've made substantial changes to `core/`, additionally `core`'s internals and the embedder API's are explicitly unstable and so we're free to change those as long as we don't break userland. So overall I don't share that fear, I think we can easily make this evolve once we have a better pattern.

Also this pattern already exists, `op_crates` and `runtime/` JS code can't work without it and implicitly depend on it, I think it's best in these cases to explicit the pattern, it's always easier to refactor explicit patterns than implicit ones.

ry (approved):
Fair argument. LGTM.

PR #10424, opened 2021-04-29
fix(cli/tools/coverage): ignore `$deno$test.js`

caspervonb:
I changed the temporary test runner module url from `$deno$test.ts` to `$deno$test.js` but forgot to add it to the list of ignores when coverage filters out special names.

This adds `$deno$test.js` to that list of ignores.


caspervonb:
Don't think we actually need to ignore these (anymore).

Issue #8550, opened 2020-11-30
Analyse Test Results?


ebebbington:
**Description**

This is a question.

Are there plans on added support for being able to 'analyse' tests results? For example:

```typescript
const result = Deno.test("My test", () => {
  assertEquals(true, true)
  assertEquals(false, true)
})
console.log(result)
// {
//   failures: [
//     {
//      line: 3,
//      assertion: "assertEquals",
//      output: {
//        expected: false,
//        actual: true
//      }
//     }
//   ],
//   passes: ...
// }
```

**Use Case**

*Note: of course. this is my own use case, but this feature may provide more utility*

I have an idea for adding a GUI for a browser automation and testing tool (similar to Cypress.io), and i've been playing around with the idea of displaying test results for each test. I think the ability to analyse tests results would be sufficient in this scenario. The only current way i can think of doing this, is running the test inside a sub process, getting the stdout/stderr,  and filtering the result so I can end up with the `- "..."\n+ "..."` blocks

nayeemrmn:
In the past we've discussed hooks for the test runner looking something like this:
```ts
Deno.test("My test", () => {
  assertEquals(true, true)
  assertEquals(false, true)
});

Deno.setTestHook((testMessage) => {
  if (testMessage.testStart != null) {
    // ...
  } else if (testMessage.testEnd != null) {
    // ...
  }
  return testMessage;
});
```
But no good use case has arisen for custom test reporters which are enclosed in the same JS context as the tests.

The best solution to the use case mentioned here is a `deno test --json`, which is inevitable anyway.

ebebbington:
I agree, seems like `deno test --json` will fit the usecase perfectly

David-Else:
VS Code team are working on integrating a testing API and UI. They already have a prototype: https://github.com/microsoft/vscode/issues/107467

caspervonb:
Been playing around with a few iterations of bringing the test runner over to rust, which involves custom test reporters (e.g deno test --json).

Coverage is next in my queue tho.

ebebbington:
Will probably close this issue then - is there an issue for the `--json` feature?

caspervonb:
> Will probably close this issue then - is there an issue for the `--json` feature?

I guess this is it :D
I've mentioned it in various issues, discord and #7915

ebebbington:
[...]

PR #10426, opened 2021-04-29
feat(cli): add json test reporter

caspervonb:
This adds JSON as an output format for test results in the same style as the linter (non streaming).

Closes https://github.com/denoland/deno/issues/8550


caspervonb:
Please merge #10428 first, as it changes the structures used.

cli/tests/test/deno_test_json.out
```diff
@@ -0,0 +1,101 @@
+{
+  "events": [
```
nayeemrmn, line 2: Discussed on discord. The output of `--json` shouldn't be events-based, it should print out a summary at the end. We should have a separate `--ndjson`/`--jsonl`/etc. flag which outputs test events as individual [line-delimited JSON](https://en.wikipedia.org/wiki/JSON_streaming#Line-delimited_JSON) objects. If anyone thinks we only need one of these, we should favour the latter. I think both are useful.

This pattern of having an overarching JSON object where nested array elements are chunked one-by-one isn't a real format and doesn't work well. Streaming consumers can't take advantage of the overarching JSON structure and will have a really annoying time parsing. Non-streaming consumers have no use for an events-based output and likely just need a summary. So this output doesn't serve anyone.

caspervonb:
Punting this for a while because upcoming changes to the internal representation is going to affect the output.

caspervonb:
Closing this for now, will revisit later.

PR #10427, opened 2021-04-29
feat: align plugin api with Extension

eliassjogreen:
Continuation on #9850. This pr aligns the plugin api with the new runtime extensions introduced in #9800. I have not yet limited any of the access to the `Extension` struct so currently they are the exact same (Except js, and middleware support which probably would not be needed nor used. Reason for not implementing them is that they need access to the `JsRuntime` which `op_open_plugin` does not have). The `Extension` struct should probably be split into a few traits (`JsExtension` and `MiddlewareExtension`) or a new `DynamicExtension` struct could be introduced.

cc @crowlKats @AaronO


lucacasonato:
@eliassjogreen Could you merge with master?

runtime/ops/plugin.rs
```diff
@@ -43,111 +36,43 @@ pub fn op_open_plugin(
   debug!("Loading Plugin: {:#?}", filename);
   let plugin_lib = Library::open(filename).map(Rc::new)?;
   let plugin_resource = PluginResource::new(&plugin_lib);
+  mem::forget(plugin_lib);
```
ry, line 39: I guess this prevents the segfault when the process exits?

If so could you add a comment here.

eliassjogreen: Correct this is to prevent segfaults, added a comment about it.

eliassjogreen:
`Deno.core.opSync` and `Deno.core.opAsync` should probably be moved out of core to the normal `Deno` namespace

ry:
> Deno.core.opSync and Deno.core.opAsync should probably be exposed outside of core in the normal Deno namespace, right?

Why's that? These are not methods we generally want user code to call. I suppose the distinction is blurred in the case of plugins.

AaronO:
> > Deno.core.opSync and Deno.core.opAsync should probably be exposed outside of core in the normal Deno namespace, right?
> 
> Why's that? These are not methods we generally want user code to call. I suppose the distinction is blurred in the case of plugins.

One possible workaround would be to have `openPlugin` return `{ opSync, opAsync, ... }` to limit exposure

PR #10432, opened 2021-04-29
feat(serde_v8): ZeroCopyBuf

AaronO:
Bidirectional zero-copy serialization of buffers between v8 & rust that can be nested in structs/tuples/etc...

First somewhat hackish implementation but should be fully functional and correct.

This unblocks PRs in WebCrypto and elsewhere, it will also allow for a few cleanups in core that I'll submit as follow-up PRs.


serde_v8/src/magic/buffer.rs
```diff
@@ -0,0 +1,132 @@
+use rusty_v8 as v8;
+
+use std::cell::Cell;
+use std::fmt;
+use std::ops::Deref;
+use std::ops::DerefMut;
+
+use super::zero_copy_buf::ZeroCopyBuf;
+
+// An asymetric wrapper around ZeroCopyBuf,
```
aapoalas, line 10: `asymmetric`

AaronO: A frenchism asymétrique

serde_v8/tests/magic.rs
```diff
@@ -52,7 +50,75 @@ fn magic_basic() {
...
+    let zbuf: Option<serde_v8::Buffer> =
+      serde_v8::from_v8(scope, v8_array).unwrap();
+    assert_eq!(&*zbuf.unwrap(), &[1, 2, 3, 4, 5]);
 
-  v8_shutdown();
+    // Observe mutation in JS
+    let v8_array = js_exec(scope, "new Uint8Array([1,2,3,4,5])");
+    let mut zbuf: serde_v8::Buffer =
+      serde_v8::from_v8(scope, v8_array).unwrap();
+    let key = serde_v8::to_v8(scope, "t1").unwrap();
```
ry, line 89: Nit: I would just use the V8 API `v8::String::new(scope, "t1").unwrap();`

ry (approved):
LGTM - very nice!

AaronO:
> LGTM - very nice!

Thanks !

PR #10434, opened 2021-04-29
cleanup(core): replace OpResponse with OpResult

AaronO:
Drop the Value/Buffer enum since #10432 allows buffers to be serialized rust => v8

## Notes

- I removed some `OpFn` call checks in the `ops.rs#op_table` test, since our OpFn outputs are somewhat opaque and it's not very meaningful to test them without going through the v8 layer


core/ops.rs
```diff
@@ -205,35 +200,14 @@ mod tests {
...
-        Op::Sync(OpResponse::Buffer(b"oof!"[..].into()))
-      });
+      foo_id = op_table
+        .register_op("foo", |_, _, _| Op::Sync(OpResult::Ok(321.into())));
       assert_eq!(foo_id, 1);
-      bar_id = op_table.register_op("bar", |_, _, _| {
-        Op::Sync(OpResponse::Buffer(b"rab!"[..].into()))
-      });
+      bar_id = op_table
+        .register_op("bar", |_, _, _| Op::Sync(OpResult::Ok(123.into())));
```
ry, line 207: These tests seem to be testing different things now. It would be cool to return a `b"rab!".to_boxed_vec()` here.

core/ops.rs
```diff
@@ -205,35 +200,14 @@ mod tests {
...
-    );
-    let bar_res = OpTable::route_op(
-      bar_id,
-      state.clone(),
-      OpPayload::empty(),
-      Default::default(),
-    );
-    assert!(
-      matches!(bar_res, Op::Sync(OpResponse::Buffer(buf)) if &*buf == b"rab!")
-    );
```
ry, line 235: These seem useful still. Can they not be modified to work?

ry (approved):
LGTM - discussed out of band

PR #10436, opened 2021-04-30
refactor(core): convert core.print() to a builtin op

AaronO:
Follow up to #10449


ry (approved):
Lovely. LGTM.

core/ops_builtin.rs
```diff
@@ -52,3 +54,20 @@ pub fn op_close(
...
+/// Builtin utility to print to stdout/stderr
+pub fn op_print(
+  _state: &mut OpState,
+  args: (String, bool),
+  _zero_copy: Option<ZeroCopyBuf>,
+) -> Result<(), AnyError> {
+  let (msg, is_err) = args;
+  if is_err {
+    eprint!("{}", msg);
+    stdout().flush().unwrap();
```
satyarohith, line 67: Shouldn't it be `stderr().flush().unwrap()` ?

AaronO: I think you're right, but it's not 100% clear if the original intention was to flush `stdout` when writing to `stderr`. IIRC `stderr` is unbuffered on Linux and macOS, but I don't think that's true on Windows.

The original intention might have been to flush both stderr/stdout (all output streams) assuming stderr was unbuffered and thus didn't need flushing.

I've submitted a PR (https://github.com/denoland/deno/pull/10480), so we can continue the conversation over there.

It seems like this should only matter on Windows, could make sense to always flush both regardless of where we're writing to, something like this:

```rust
/// Builtin utility to print to stdout/stderr
pub fn op_print(
  _state: &mut OpState,
  args: (String, bool),
  _zero_copy: Option<ZeroCopyBuf>,
) -> Result<(), AnyError> {
  let (msg, is_err) = args;
  if is_err {
    eprint!("{}", msg);
  } else {
    print!("{}", msg);
  }
  // Flush all output streams (stderr & stdout)
  stdout().flush().unwrap();
  stderr().flush().unwrap();
  Ok(())
}

```

Issue #10368, opened 2021-04-25
Deno lsp showing false errors and stops linting new code, requires reloading window to resolve


KyleJune:
In the video linked below, both primaryGroupServiceTests and secondaryGroupServiceTests are a valid first argument for the test function. I recorded this last week and no longer have the code. I tried finding a minimum reproduction path but it seems to happen randomly. To capture it happening I just had to record myself coding. It was most common for me when using backspace or deleting lines. Sometimes I would have it happen to me after running deno fmt in the terminal. If I use the command to restart language server, it doesn't resolve the issue. I end up having to frequently use the reload window command to resolve the issue.

https://www.youtube.com/watch?v=9kmzApeIxp8

Here is a link to the start of a discord conversation where others had noted that they experienced the same issue as me. https://discord.com/channels/684898665143206084/684898665151594506/835903650152972368

danopia:
I've been experiencing this pretty often (depending on what I'm working on, ranging from 1/hr up to every minute or two) and it _seems_ to be some sort of 'stuck request' issue.

Before reloading the window I usually try running `pkill deno` in a terminal. This usually results in a huge pile of "Connection disposed" or "Connection to server is erroring" messages appearing. Often the lsp comes back up and then everything updates after the `pkill`, but other times it stays broken and I reload the editor.

I only recently found `Restart Language Server` and both times I tried it, the incorrect error spans went away, but nothing replaced them and I still had to reload the window.

Assuming this is the same issue, it's pretty hard to figure out what's not working; I'm not sure what debugging I can do once the bad state is recreated.

Here's an instance where I had already backspaced `ApiResponse` to `Response` and the LSP didn't notice, until I tried `pkill deno` then everything works again:

https://user-images.githubusercontent.com/40628/116003428-338bac80-a5fe-11eb-9782-4d7c6236eae2.mp4

Here's a screenshot from 20 days ago showing the other message spam that often happens after `pkill deno`:

![Screenshot 2021-04-03 07 17 17](https://user-images.githubusercontent.com/40628/116003455-4f8f4e00-a5fe-11eb-8cdd-cf731586d2d0.png)

danopia:
Here's an example (again, already stuck, as I haven't been generally screen recording) where the Restart LSP option doesn't restore the LSP health:

https://user-images.githubusercontent.com/40628/116048970-0da1ee80-a676-11eb-8929-918c10f1bf1b.mp4

kitsonk:
[...]

PR #10438, opened 2021-04-30
feat(lsp): add internal debugging logging

kitsonk:
Ref: #10368

The lsp would respect a setting which will log all the marks and measures to stderr along with their arguments which can be used to investigate certain issues (as well as more easily generate test cases).  This requires a client which can pass the configuration option (see: denoland/vscode_deno#406).

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -26,6 +26,8 @@ use std::collections::HashMap;
 use std::env;
 use std::path::PathBuf;
 use std::rc::Rc;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use tokio::fs;
 
@@ -92,6 +94,7 @@ pub(crate) struct Inner {
   client: Client,
   /// Configuration information.
   config: Config,
+  debug_flag: Arc<AtomicBool>,
   diagnostics_server: diagnostics::DiagnosticsServer,
   /// The "in-memory" documents in the editor which can be updated and changed.
   documents: DocumentCache,
@@ -121,13 +124,16 @@ pub(crate) struct Inner {
 }
 
 impl LanguageServer {
-  pub fn new(client: Client) -> Self {
-    Self(Arc::new(tokio::sync::Mutex::new(Inner::new(client))))
+  pub fn new(client: Client, lsp_debug_flag: Arc<AtomicBool>) -> Self {
+    Self(Arc::new(tokio::sync::Mutex::new(Inner::new(
+      client,
+      lsp_debug_flag,
+    ))))
   }
 }
 
 impl Inner {
-  fn new(client: Client) -> Self {
+  fn new(client: Client, debug_flag: Arc<AtomicBool>) -> Self {
     let maybe_custom_root = env::var("DENO_DIR").map(String::into).ok();
     let dir = deno_dir::DenoDir::new(maybe_custom_root)
       .expect("could not access DENO_DIR");
@@ -144,6 +150,7 @@ impl Inner {
       assets: Default::default(),
       client,
       config: Default::default(),
+      debug_flag,
       diagnostics_server,
       documents: Default::default(),
       maybe_config_uri: Default::default(),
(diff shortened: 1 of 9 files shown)


cli/lsp/language_server.rs
```diff
@@ -343,8 +351,14 @@ impl Inner {
     Ok(())
   }
 
+  pub fn update_performance(&mut self) {
+    self
+      .performance
+      .do_logging(self.config.settings.internal_debug);
+  }
```
ry, line 358: We already have a system for conditional logging using the `debug!` and `info!` macros. These can be enabled with the `--log-level=debug` command-line option. Can we not reuse that system for LSP debugging?

Note that the log-level could be adjusted at runtime as well: 
https://docs.rs/log/0.4.14/log/#implementing-a-logger

cli/lsp/performance.rs
```diff
@@ -129,14 +133,38 @@ impl Performance {
...
+  pub fn mark<S: AsRef<str>, V: Serialize>(
+    &self,
+    name: S,
+    maybe_args: Option<V>,
+  ) -> PerformanceMark {
     let name = name.as_ref();
     let mut counts = self.counts.lock().unwrap();
     let count = counts.entry(name.to_string()).or_insert(0);
     *count += 1;
+    if self.logging_enabled {
```
ry, line 152: Ideally this should use the conditional logging provided by the rust log / env_logger crate.

cli/lsp/language_server.rs
kitsonk: Yeah, I knew that, the problem is a lot of other crates, especially swc, use the same logging, and so setting the logging flag makes it impossible to sort out what is going on.

ry: env_logger provides the ability to filter messages based on regex
https://docs.rs/env_logger/0.8.3/env_logger/#filtering-results

kitsonk: Ok, I spent some time looking at this.  A few notes:

- the logger can only be set once and has to be global static
- we use env_logger as our global logger to do this, once we `init()` it appears you can't reconfigure it.
- the `RUST_LOG` is an environment variable, which can be passed in at startup, and the filter `debug/mark|measure` would be suitable, but we can only do that when start/restarting the LSP.
- you can use the `.parse_filters()` method on the builder which follows the same format as `RUST_LOG` but again that has to be determined at startup, versus changing dynamically.

There is the `.format()` closure which maybe could do some "filtering" of records based on a flag, but there doesn't appear to be a dynamic way of changing things in a running process, outside of implementing our own logger (which is what is mentioned in the first link).

ry: My guess is the Log::enabled method could be used to dynamically turn on and off logging
https://docs.rs/log/0.4.14/log/trait.Log.html#tymethod.enabled

But it seems a little difficult to access... I guess you'd have to wrap the env_logger.

kitsonk: Hmmm... yeah, that should work, but like you said would require wrapping. Ok, I will look at that a bit closer. You can generate the logger without installing it for wrapping, that I know.

kitsonk: Ok, I reworked it... the biggest "problem" is that we need to share the state of the logging being enabled between the logger and the `lsp::language_server` and we don't have any infrastructure for that, so I chose a way (`Arc<AtomicBool>`) to do that, but am more than willing to take other suggestions.

Also broke all the logging stuff out to a seperate module, as it just seem unnecessary in an already complex `main.rs`.

cli/main.rs
```diff
@@ -1147,7 +1111,12 @@ pub fn main() {
   if !flags.v8_flags.is_empty() {
     init_v8_flags(&*flags.v8_flags);
   }
-  init_logger(flags.log_level);
 
-  unwrap_or_exit(tokio_util::run_basic(get_subcommand(flags)));
+  // when this flag is set to `true`, the process will log specific diagnostic
+  // debug information,
+  let lsp_debug_flag = Arc::new(AtomicBool::new(false));
```
ry, line 1117: It's unfortunate that this `lsp_debug_flag` has leaked all the way to top-level of the program and many signatures have to be modified to pass it around.

What if you use lazy_static to have a global `Arc<AtomicBool>` in `cli/logger.rs`? Then `cli/lsp/language_server.rs` could reference that directly without having to pass it around like this. I think practically this `lsp_debug_flag` is global anyway.

kitsonk: Yeah, that makes more sense. I will try that.

ry (approved):
LGTM

PR #10443, opened 2021-04-30
fix(cli): default to num cpus when no value is given to `--jobs`

caspervonb:
This defaults to num_cpu's when no value is given for jobs, matching most implementations of -j, --jobs in the wild.

I had this hardcoded to two because there's a nightly function in std for getting the number of hardware threads, but nightlies can take forever to be stable so added a dependency on num_cpu's in the meantime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -540,6 +540,7 @@ dependencies = [
  "lspower",
  "nix",
  "notify",
+ "num_cpus",
  "os_pipe",
  "percent-encoding",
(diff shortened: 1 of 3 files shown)


ry (approved):
LGTM

bartlomieju:
FYI in `op_system_cpu_info` we already get a number of cores using `sys_info::cpu_num`. Not sure if we need another crate for this purpose.

PR #10449, opened 2021-04-30
refactor(core): provide builtins as an Extension

AaronO:
Treats builtin ops (`op_resources`, `op_close`) and js (`core.js`, `error.js`) as a regular `JsRuntime` `Extension` that's always loaded (first, see the `.insert(0 ...`)

This guarantees that `op_close`, `op_resources` (and `op_print` with #10436) are ready to use after `JsRuntime::new`

Though builtins are special, they're not "that special", streamlining them will simplify things like moving JS src off heap generically for all extension provided JS without having to treat `core.js`, `error.js` as edge cases

## Notes

- Now that the builtin ops are registered by default, the first op registered by the user after `JsRuntime::new` will no longer have an `OpId` of 1 since those builtin ops will have taken those slots. I fixed the tests that relied on that assumption and in general it's strongly discouraged to rely on specific OpIds

### DIFF
--- a/core/ops_builtin.rs
+++ b/core/ops_builtin.rs
@@ -1,18 +1,29 @@
 use crate::error::bad_resource_id;
 use crate::error::type_error;
 use crate::error::AnyError;
+use crate::include_js_files;
+use crate::op_sync;
 use crate::resources::ResourceId;
+use crate::Extension;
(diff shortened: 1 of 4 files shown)


ry (approved):
LGTM

PR #10454, opened 2021-05-01
fix(wpt): fix wpt test harness and disable some tests

### LINKED ISSUE #10453: False negatives in WPT
The following wpt tests throw uncaught errors in the middle of execution, but the error is swallowed by Deno.exit() call in completion callback because of this issue https://github.com/denoland/deno/issues/10445

- FileAPI/file/File-constructor.any
- streams/readable-byte-streams/general.any
- streams/readable-streams/templated.any

For example, file constructor test throws at this line https://github.com/denoland/wpt/blob/301177b/FileAPI/file/File-constructor.any.js#L44 , but the error is canceled by Deno.exit call in test harness here https://github.com/denoland/deno/blob/8424647/tools/wpt/testharnessreport.js#L12

kt3k:
closes #10453

### DIFF
--- a/tools/wpt/testharnessreport.js
+++ b/tools/wpt/testharnessreport.js
@@ -8,10 +8,6 @@ window.add_result_callback(({ message, name, stack, status }) => {
   );
 });
 
-// TODO(kt3k): Enable the below hook and timers test when #10445 is fixed
-// ref: https://github.com/denoland/deno/issues/10445
-/*
 window.add_completion_callback((tests, harnessStatus) => {
-  Deno.exit(0);
+  window.close();
 });
-*/
--- a/tools/wpt/testharnessreport.js
+++ b/tools/wpt/testharnessreport.js
@@ -8,6 +8,10 @@ window.add_result_callback(({ message, name, stack, status }) => {
   );
 });
 
+// TODO(kt3k): Enable the below hook and timers test when #10445 is fixed
+// ref: https://github.com/denoland/deno/issues/10445
+/*
 window.add_completion_callback((tests, harnessStatus) => {
   Deno.exit(0);
 });
+*/
--- a/tools/wpt/expectation.json
+++ b/tools/wpt/expectation.json
@@ -392,61 +392,7 @@
         "ReadableStream with byte source: reading into a zero-length view on a non-zero-length buffer rejects"
       ],
       "construct-byob-request.any.js": false,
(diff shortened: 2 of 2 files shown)


tools/wpt/testharnessreport.js
```diff
@@ -8,6 +8,10 @@ window.add_result_callback(({ message, name, stack, status }) => {
   );
 });
 
+// TODO(kt3k): Enable the below hook and timers test when #10445 is fixed
+// ref: https://github.com/denoland/deno/issues/10445
+/*
 window.add_completion_callback((tests, harnessStatus) => {
   Deno.exit(0);
 });
+*/
```
lucacasonato, line 17: What if you call `Deno.exit` inside a `queueMicrotask`?

kt3k: That didn't make difference (still causes swallowing of uncaught error). If I use setTimeout, then it works for the file-constructor test etc, but it makes timers tests flaky.

lucacasonato:
I am unsure if this is the correct path forward. I think this might cause some hanging in other tests. Let's discuss on Monday.

tools/wpt/testharnessreport.js
```diff
@@ -9,5 +9,5 @@ window.add_result_callback(({ message, name, stack, status }) => {
 });
 
 window.add_completion_callback((tests, harnessStatus) => {
-  Deno.exit(0);
+  window.close();
```
kt3k, line 12: This is basically the same as
```ts
queueMicrotask(() => {
  setTimeout(() => {
    Deno.exit(0);
  }, 0);
});
```
And I observed some flaky test failure with this change with this test case on my local machine (mac) https://github.com/denoland/wpt/blob/master/html/webappapis/timers/negative-settimeout.any.js#L3

But it might work in CI. I'm ok with landing this.

lucacasonato: Ah. I understand why it is flakey now... I can fix the upstream WPT test to not be flakey.

bartlomieju:
@lucacasonato is this PR needed after your recent changes to WPT runner?

lucacasonato:
yes, but I still need to fix some test before we can land it.

bartlomieju:
@lucacasonato is the fix available now?

lucacasonato:
No, not yet.

PR #10455, opened 2021-05-01
fix(extensions/console): Remove `console.timeStamp` from types

### LINKED ISSUE #10444: `console.timeStamp` is documented but does not exist
`console.timeStamp` is shown as an API in the TypeScript types, but doesn't actually exist at runtime:
https://github.com/denoland/deno/blob/8922639c1dc48a8b1ed6910bfe62306d9178d70b/op_crates/console/lib.deno_console.d.ts#L26
```
$ deno repl --unstable
Deno 1.9.2
exit using ctrl+d or close()
> console.timeStamp
undefined
```
It should be either implemented or removed from the types.

Liamolucko:
Fixes #10444

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -3089,7 +3089,7 @@ mod tests {
     assert!(result.is_ok());
     let response: CompletionInfo =
       serde_json::from_value(result.unwrap()).unwrap();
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM - thanks

PR #10458, opened 2021-05-01
fix(extensions/websocket): better websocket errors

crowlKats:


lucacasonato (approved):
I don't know how this would be tested, so LGTM.

PR #10459, opened 2021-05-01
fix(extensions/file): re-enable File wpt and update File constructor

kt3k:
depends on #10454

This PR re-enables File wpt and update File constructor (This removes the replacement of `/` -> `:` in the file name, which is updated in the upstream spec (ref: https://github.com/w3c/FileAPI/pull/171)


kt3k:
closing in favor of #10760

PR #10460, opened 2021-05-01
cleanup(bench/deno_http_native): don't use Deno.core funcs

AaronO:
`Deno.core.*` is unstable and not fit for public consumption, although this is a somewhat internal bench some people may use it as reference code and start using `Deno.core.encode()` in their own code


ry (approved):
LGTM

PR #10461, opened 2021-05-01
refactor(runtime): use Extensions

AaronO:
Refactors `runtime/` to use the new `JsRuntime` `Extension`s landed in #9800, paves a first path for a more modular `runtime/` and streamlines worker init logic (having both web `extensions/` and `runtime/` init'd declaratively instead of imperatively)

## Changes

- [x] Refactor `runtime::ops::*::init` to return `Extension`s
- [x] Refactor workers to init runtime ops with extensions
- [x] Move `WebWorkerHandle`/`sender` to `OpState`

## Notes

- This refactoring doesn't include the runtime JS in Extensions, that will require more work, to avoid bugs and duplicating logic with snapshots
- For now we depend on `runtime/`'s JS to be provided via the snapshot, no different than in `main`
- However I think we should clean this up & decouple it, so we can boot a `Worker` without depending on that snapshot

### DIFF
--- a/runtime/ops/fs.rs
+++ b/runtime/ops/fs.rs
@@ -8,6 +8,9 @@ use deno_core::error::bad_resource_id;
 use deno_core::error::custom_error;
 use deno_core::error::type_error;
 use deno_core::error::AnyError;
+use deno_core::op_async;
+use deno_core::op_sync;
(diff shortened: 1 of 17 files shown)


ry (approved):
LGTM

PR #10463, opened 2021-05-01
cleanup(cli): prefix all op functions with op_

AaronO:
Makes the codebase more searchable and helps distinguish op functions from helper functions

Besides tests/examples/benches this convention appears to be followed everywhere else in the codebase


ry (approved):
LGTM

PR #7512, opened 2020-09-16
ci: remove wasm target installation step

caspervonb:
This removes the wasm targets from the workflow as they are no longer used for test runs.

PR #10465, opened 2021-05-02
docs: remove wasm and wasi as prerequisites

caspervonb:
This removes wasm and wasi as prerequisites, as they haven't been since
we stopped building the wasi test suite from source which dates back
quite some time now.

I removed it from the CI back then but forgot to update the docs.

ref: #7512


kt3k (approved):
LGTM

caspervonb:
Thanks for digging it up @kt3k, couldn't find it 😁

PR #10466, opened 2021-05-02
fix: change inflection depending on number of pending tests

caspervonb:
Before:

```
running 1 tests from file://...
```

After:

```
running 1 test from file://...
```


ry (approved):
LGTM

PR #10472, opened 2021-05-02
refactor(snapshots): to their own crate

AaronO:
## Rationale

Today snapshots are implicit & somewhat hardcoded/non-optional dependencies. Snapshots should be encouraged for release builds but shouldn't be a hard requirement. This PR decouples snapshots and makes them opt-in (they should be an optional performance enhancement).

Decoupling snapshots will facilitate other work such as #9713, a proposal I have in mind for `deno compile` and making extensions work gracefully with snapshots (differential/partial inits)

## Key changes

The main functional differences are that `MainWorker` and `WebWorker` will get their snapshots from `cli/` instead of having them hardcoded. In general this helps make `deno_runtime` more modular and reusable (since embedders could use it and pass their own snapshot built with deno's extensions + their own)

## Subtasks

- [x] Factor out snapshots to a`deno_snapshots` crate
- [x] Remove `MainWorker` and `WebWorker`'s hardcoded dep on snapshots
- [x] Make `runtime::js` provide an `init() -> Extension` so we have a single source of truth for runtime JS files
- [x] Have snapshot use `runtime::js::init()`, we should no longer be dependant on the JS being provided by the snapshot
- [x] Rewire
- [ ] Cleanup


piscisaureus:
I will admit being a skeptical about the snapshot feature actually working out and proving to be useful, but I don't want to discourage experimentation with it.

That said, we're not going to land this PR as-is, because without a end-to-end implementation of the snapshotting features that this change is made for, it is not an improvement.

bartlomieju:
This PR is very stale with lots of conflicts so I'm gonna close it. I agree with @piscisaureus opinion. If you want to revisit the topic at a later date please reopen the PR.

PR #10474, opened 2021-05-02
fix(op_crates/fetch): check fetch() argument length 

CrimsonCodes0:
Related to https://github.com/denoland/deno/pull/10203

Checks that fetch receives at least one arguement.

Relevant WebIDL:
```
partial interface mixin WindowOrWorkerGlobalScope {
  [NewObject] Promise<Response> fetch(RequestInfo input, optional RequestInit init = {});
};
```
\- https://fetch.spec.whatwg.org/#fetch-method

`input` must be present.

cc @lucacasonato

### DIFF
--- a/tools/wpt/expectation.json
+++ b/tools/wpt/expectation.json
@@ -1014,7 +1014,6 @@
         "Response interface: operation json()",
         "Response interface: operation text()",
         "Response interface: calling redirect(USVString, optional unsigned short) on new Response() with too few arguments must throw TypeError",
-        "Window interface: calling fetch(RequestInfo, optional RequestInit) on window with too few arguments must throw TypeError",
         "Window interface: operation fetch(RequestInfo, optional RequestInit)"
       ]
     },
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
Thanks! Must have missed this during fetch rewrite.

CrimsonCodes0:
@lucacasonato Somewhat related, but I had barely left two comments on the rewrite PR, since I had barely checked it out within the hour. I'm not sure if those would be problematic or not, but it might be worth checking them out.

tools/wpt/expectation.json
```diff
@@ -1014,7 +1014,6 @@
         "Response interface: operation json()",
         "Response interface: operation text()",
         "Response interface: calling redirect(USVString, optional unsigned short) on new Response() with too few arguments must throw TypeError",
-        "Window interface: calling fetch(RequestInfo, optional RequestInit) on window with too few arguments must throw TypeError",
```
CrimsonCodes0, line 1017: Just to confirm, this file is what is expected to fail? Therefore, by removing, it is expected to pass now, right?

lucacasonato: Yes. Looks good

PR #10478, opened 2021-05-03
fix(cli): give context when failed to load import map

### LINKED ISSUE #10476: Non descript error when import map isnt found
```
$ deno info --import-map does_not_exist.json does_exist.js
error: No such file or directory (os error 2)
```

This error should contain some context that the _import map_ isn't found.

satyarohith:
Closes #10476

### DIFF
--- a/cli/ops/runtime_compiler.rs
+++ b/cli/ops/runtime_compiler.rs
@@ -87,7 +87,13 @@ async fn op_emit(
       let file = program_state
         .file_fetcher
         .fetch(&import_map_specifier, &mut runtime_permissions)
-        .await?;
+        .await
+        .map_err(|e| {
+          generic_error(format!(
+            "Unable to load '{}' import map: {}",
+            import_map_specifier, e
+          ))
+        })?;
       ImportMap::from_json(import_map_specifier.as_str(), &file.source)?
     };
     Some(import_map)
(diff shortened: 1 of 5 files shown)


kt3k:
I think we can add the command in the original issue as an integration test case.

kitsonk:
While we have the same issue in other places, it would be good if the context actually provided the resolved specifier that was trying to be loaded as well.  We should also check to see what the error is when using a remote URL that doesn't exist as well.

satyarohith:
This is how the error looks for remote import maps:
```
➜ deno run --import-map https://example.com/does_not_exist.json does_exist.js
Download https://example.com/does_not_exist.json
error: Unable to load 'https://example.com/does_not_exist.json' import map

Caused by:
    Import 'https://example.com/does_not_exist.json' failed: 404 Not Found
```
For local import maps:
```
➜ deno run --import-map does_not_exist.json does_exist.js 
error: Unable to load 'file:///[DISCARDED]does_not_exist.json' import map

Caused by:
    No such file or directory (os error 2)
```
And for `Deno.emit()`:
```
Error: Unable to load 'file:///[DISCARDED]does_not_exist.json' import map: No such file or directory (os error 2)
    at deno:core/core.js:86:46
    at async file:///Users/sr/c/github.com/satyarohith/deno/test.js:2:3
```

I am not happy with the error message of `Deno.emit()` but let me if there's a better way to display the message.

bartlomieju (approved):
LGTM, nice fix Satya

PR #10480, opened 2021-05-03
fix(core/print): flush stderr

AaronO:
Stderr is unbuffered on Linux/macOS (so doesn't need to be flushed), but I think it's undefined on Windows

Follow up to https://github.com/denoland/deno/pull/10436#pullrequestreview-650053348


piscisaureus:
Looks like this fixes a simple mistake to me.

One could argue for flushing both stdout and stderr, but that doesn't seem necessary because now they're flushed immediately when written to anyway.

piscisaureus (approved):
LGTM

PR #10481, opened 2021-05-03
refactor(cli): replace loading file for --config flag with generic structure

bartlomieju:
Currently file passed to `--config` file is parsed using `TsConfig` structure 
that does multiple things when loading the file.  Instead of relying on that 
structure I've introduced `ConfigFile` structure that can be updated to 
sniff out more fields from the config file in the future.


cli/tsc_config.rs
```diff
@@ -188,18 +188,17 @@ fn parse_compiler_options(
   Ok((value, maybe_ignored_options))
 }
 
-/// Take a string of JSONC, parse it and return a serde `Value` of the text.
+/// Take a reference to `ConfigFile`, parse `compilerOptions` and return a serde `Value` of the text.
 /// The result also contains any options that were ignored.
-pub fn parse_config(
-  config_text: &str,
-  path: &Path,
+pub fn parse_compiler_options_from_config_file(
```
kitsonk, line 193: Would it make more sense to do an `.as_compiler_options()` method on the `ConfigFile` struct?

cli/module_graph.rs
```diff
@@ -824,7 +829,11 @@ impl Graph {
       }));
     }
     let maybe_ignored_options =
-      config.merge_tsconfig(options.maybe_config_path)?;
+      if let Some(config_file) = options.maybe_config_file.as_ref() {
```
kitsonk, line 832: I'm ok with this, but if we have `ConfigFile::as_compiler_options()` we could in theory simply this?  Also, wouldn't it be cleaner to have whatever method accept an `Option<ConfigFile>` so you have only only deal with the option unwrapping once instead of several places?

kitsonk (commented):
A couple thoughts

cli/tsc_config.rs
bartlomieju: I thought about it, but it would mean that `TsConfig` would leak into `ConfigFile`. I guess it's gonna happen sooner or later, so I'm just gonna go ahead and do that.

cli/module_graph.rs
bartlomieju: Valid point about `Option<ConfigFile>`, but if I'm add `ConfigFile::as_compiler_options` it will still required to do the unwrap. To be honest I don't really mind that - the code is still readable, it's just there are additional 3 lines.

cli/tsc_config.rs
kitsonk: Yeah, possible move TsConfig into the config as it becomes "master" place for all things config?

bartlomieju: As suggested I've merged `tsc_config.rs` into `config_file.rs`

bartlomieju:
@kitsonk please review, I want to land this PR before 1.10

cli/file_watcher.rs
```diff
@@ -181,15 +183,18 @@ where
...
   loop {
     match module_resolver().await {
       ModuleResolutionResult::Success {
         paths_to_watch,
         module_info,
+        program_state,
       } => {
         paths = paths_to_watch;
         module = Some(module_info);
+        pstate = Some(program_state);
```
kitsonk, line 197: nit: `maybe_program_state`

cli/file_watcher.rs
```diff
@@ -204,7 +209,8 @@ where
     let _watcher = new_watcher(&paths, &debounce)?;
 
     if let Some(module) = &module {
-      let func = error_handler(operation(module.clone()));
+      let func =
+        error_handler(operation(pstate.clone().unwrap(), module.clone()));
```
kitsonk, line 213: this is funky... why do we wrap it in an option to only clone it an unwrap it‽

kitsonk (approved):
LGTM, just a couple of final comments

cli/file_watcher.rs
bartlomieju: Indeed, this is because of how this function is structured. I'll try to rework it in a follow up

PR #10492, opened 2021-05-04
[WIP] feat: allow to load import map from configuration file

bartlomieju:
This PR adds ability to load import map using configuration file with `--config` flag
instead of using `--import-map`.

There are two supported modes:
```
// Example 1
// Embed import map inside configuration file
{
  "compilerOptions": {
    "target": "ES5",
    "jsx": "preserve"
  },
  "importMap": {
    "imports": {
      "moment": "./moment/moment.ts",
      "moment/": "./moment/",
      "lodash": "./lodash/lodash.ts",
      "lodash/": "./lodash/",
      "https://www.unpkg.com/vue/dist/vue.runtime.esm.js": "./vue.ts"
    },
    "scopes": {
      "scope/": {
        "moment": "./scoped_moment.ts"
      }
    }
  }
}
```

```
// Example 2
// Specify path or URL to import map inside configuration file
{
  "compilerOptions": {
    "target": "ES5",
    "jsx": "preserve"
  },
  "importMap": "./import_maps/autogenerated.json"
}
```

In the first case import map URL equals to the config file path, in the second case it's fully resolved URL.

If proposal for multiple import maps will move forward we'll have to change the schema to support arrays of URL.

<s>Blocked on https://github.com/denoland/deno/pull/10481</s>


satyarohith:
What are your thoughts on using `importMapPath` for a path or URL? We use this name in [emit options](https://doc.deno.land/builtin/unstable#Deno.EmitOptions) of `Deno.emit()` API.

bartlomieju:
> What are your thoughts on using `importMapPath` for a path or URL? We use this name in [emit options](https://doc.deno.land/builtin/unstable#Deno.EmitOptions) of `Deno.emit()` API.

It's unfortunate that it's named `importMapPath`... I wouldn't mind having `importMapUrl` and `importMap`. @kitsonk what do you think?

nayeemrmn:
I think `importMap` (url) and `importMapInline` (inline) would align with what we'd do if there was a CLI flag for the latter.

kitsonk:
I'm still unsure what the problem would be with having `"imports"` and `"scopes"` at the top level and then an `"importMap"` as a URL. We can change `Deno.emit()` to better align to this. The other option is to have it still be a single option `"importMap"`, where when the value is a string it is loaded or an object taken literally.

bartlomieju:
Discussed with @piscisaureus, we're not so sure we want to land it for 1.12, so I'm removing a milestone.

bartlomieju:
This is being punted till later date.

PR #10494, opened 2021-05-05
docs: use one code block per truncate example

caspervonb:
Truncate examples are using const and re-declaring in a single code block with multiple examples.
This splits them into using one code blocks per example program.


kt3k (approved):
LGTM. Thanks!

This should be necessary for applying doc testing to these files

caspervonb:
> This should be necessary for applying doc testing to these files
It is 😏

PR #10495, opened 2021-05-05
docs: explicitly import assert in examples

caspervonb:

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1575,6 +1575,7 @@ declare namespace Deno {
    * what it points to..
    *
    * ```ts
+   * import { assert } from "https://deno.land/std/testing/asserts.ts";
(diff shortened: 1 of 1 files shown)


satyarohith (commented):
Looks good! I like examples that can directly be run when copy-pasted.

kt3k:
I like this change and maybe this is necessary when we start doc testing. Does anyone have concern about importing std from main repo example?

caspervonb:
> I like this change and maybe this is necessary when we start doc testing. Does anyone have concern about importing std from main repo example?

Yup, I need this (and the others to come) to get more of cli/dts passing!
We already use `std/testing/assert.ts` other examples above and below so ship has kinda already sailed.
Altho I think we could just use console.assert for the `Deno` namespace examples but that's a discussion for another day.

kt3k (approved):
Thank you for the explanation. Right, we already do this in many places. LGTM

PR #10496, opened 2021-05-05
docs: use one block per writeAll example

caspervonb:
The `writeAll`  examples are using const and re-declaring in a single code block with multiple examples.
This splits them into using one code blocks per example program.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -989,13 +989,17 @@ declare namespace Deno {
    * // Example writing to stdout
    * const contentBytes = new TextEncoder().encode("Hello World");
    * await Deno.writeAll(Deno.stdout, contentBytes);
+   * ```
    *
+   * ```ts
    * // Example writing to file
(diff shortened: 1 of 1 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -1015,13 +1019,16 @@ declare namespace Deno {
...
    * Deno.writeAllSync(Deno.stdout, contentBytes);
+   * ```
    *
+   * ```ts
    * // Example writing to file
    * const contentBytes = new TextEncoder().encode("Hello World");
    * const file = Deno.openSync('test.file', {write: true});
    * Deno.writeAllSync(file, contentBytes);
    * Deno.close(file.rid);
    *
```
kt3k, line 1030: closing triple backticks seem missing here?

caspervonb: Yikes, good catch @kt3k 💯

kt3k (approved):
Thanks for updating! LGTM

PR #10497, opened 2021-05-05
docs(runtime): fix HTTP server example

vwkd:
I'm not sure if I understand it correctly. However, checking a promise in an `if` statement always returns `true` since it's an object. Would make more sense if it's awaited.

Please close if I'm missing anything.


nayeemrmn:
The `await` is right, but the promised return type of `Listener::accept()` isn't nullable either. It should really be a try-catch where the error causes a `break` if it's a `BadResource` or is propagated otherwise.

kt3k (approved):
@vwkd Thanks! LGTM

and thanks, Nayeem, for the help.

Issue #10482, opened 2021-05-03
deno install with an import map from a URL


oscarotero:
```
deno run --import-map=https://example.com/import_map.json https://example.com/cli.ts
```
This works fine, it allows to load a remote import map file using a url (#9510).

But this same code does not work to install the script with 
```
deno install --import-map=https://example.com/import_map.json https://example.com/cli.ts
```
It seems like only local files are supported on install. I'm getting the error `error: No such file or directory (os error 2)`

Is this intentional or it's a bug? The ability to load remote import maps would be very useful to manage the versions of the dependencies of an script.

bartlomieju:
Related to #10476

satyarohith:
Working on it.

PR #10499, opened 2021-05-05
fix(cli/tools/installer): allow remote import maps

satyarohith:
The previous version fetched the source of specified local import map and recreated new import map alongside the bin script. 
This commit removes the old behaviour and directly passes the specified import map URL without recreating a new one. 
This will enable us to support remote import maps for `deno install`.
 
Closes https://github.com/denoland/deno/issues/10482


satyarohith:
I made it similar to how scripts are installed. It will now have an absolute file URL for relative paths. And I also added a test.

cli/tools/installer.rs
```diff
@@ -872,4 +878,49 @@ mod tests {
...
+    );
+    if cfg!(windows) {
+      expected_string = format!(
+        "\"--import-map\" \"{}\" \"http://localhost:4545/cli/tests/cat.ts\"",
+        import_map_url.to_string()
+      );
+    }
+
+    let content = fs::read_to_string(file_path).unwrap();
+    assert!(content.contains(&expected_string));
```
kt3k, line 924: 👍

kt3k (approved):
@satyarohith Thank you for updating and adding tests! LGTM

PR #10501, opened 2021-05-05
docs: add missing awaits

caspervonb:


kt3k (approved):
LGTM

We have this error in many places...

Issue #10413, opened 2021-04-28
Deno.emit() does not produce source map when bundling


treethought:
with deno v1.9.2

bar.js
```js
export default function doBar() {
    console.log("did bar")
}
```

foo.js
```js
import doBar from "./bar.js"
doBar()
```

main.ts
```ts
const { diagnostics, files, ignoredOptions } = await Deno.emit(
  "foo.js",
  {
    bundle: "esm",
    compilerOptions: { sourceMap: true },
  },
);
console.log({ diagnostics, files, ignoredOptions });
```

Based on the EmitResult docs for "files " from https://deno.land/manual@v1.9.2/typescript/runtime#denoemit:

```
 /** Any emitted files.  If bundled, then the JavaScript will have the
   * key of `deno:///bundle.js` with an optional map (based on
   * `compilerOptions`) in `deno:///bundle.js.map`. */
```

Expected Output:
```
{
  diagnostics: [],
  files: {
    "deno:///bundle.js": 'function doBar() {\n    console.log("did bar");\n}\ndoBar();\n',
    "deno:///bundle.js.map: "...."
  },
  ignoredOptions: null
}

```

Actual Output:
```
{
  diagnostics: [],
  files: {
    "deno:///bundle.js": 'function doBar() {\n    console.log("did bar");\n}\ndoBar();\n'
  },
  ignoredOptions: null
}
```

kitsonk:
While we don't error on this, it was never implemented to generate the source map.  It is a fine line between a "bug" versus a "feature" but we should support generating a source map as well inline source map for bundles.

alexanderGalushka:
should the labels be: feat + bug :)

kitsonk:
> should the labels be: feat + bug :)

No.

satyarohith:
Working on it.

kitsonk:
Treating as a bug, because way back when, `Deno.bundle()` used to do it when we weren't using swc.

PR #10510, opened 2021-05-05
feat(Deno.emit): support source maps with bundle option

satyarohith:
This PR adds support for source maps to `Deno.emit` when the bundle option is used. By default, a corresponding `deno://bundle.js.map` file is produced matching the behavior of `Deno.emit` without the bundle option. You can opt for inline source map by passing `inlineSourceMap: true` to the config.

Closes https://github.com/denoland/deno/issues/10413

### DIFF
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -960,6 +960,8 @@ impl Graph {
     let opts = match options.bundle_type {
       BundleType::Module | BundleType::Classic => json!({
         "noEmit": true,
+        "removeComments": true,
+        "sourceMap": true,
       }),
       BundleType::None => json!({
         "outDir": "deno://",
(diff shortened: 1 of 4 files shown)


cli/module_graph.rs
```diff
@@ -768,15 +768,16 @@ impl Graph {
       "checkJs": false,
       "emitDecoratorMetadata": false,
       "importsNotUsedAsValues": "remove",
-      "inlineSourceMap": true,
+      "inlineSourceMap": false,
```
satyarohith, line 771: `inlineSourceMap: true` previously did not have any effect as source maps were not produced before. This will probably be set back to true when we implement source maps for `deno bundle`.

Currently, `Deno.emit` produces a `separate deno://bundle.js.map` file when the bundle option is set.

kitsonk (approved):
LGTM

PR #10521, opened 2021-05-06
feat(test): add support for type checking documentation

caspervonb:
This adds support for inline documentation tests.

Blocks of code within a jsdoc/tsdoc style block comment that are encased in fenced code blocks get parsed out to individual modules, type checked (and will later be run as test programs but we cut for now that to hit the release deadline).

Example:
~~~ts
/**
 * ```ts
 * import { example } from "./example.ts";
 *
 * example();
 * ```
 */
export function example() {
  // ...
}
~~~


caspervonb:
Cut out running for now to hit the 1.10 window, will land running documentation examples in a follow up after today's release.

bartlomieju (approved):
LGTM, this is great for a first pass, we'll iterate on it after 1.10. Huge thanks Casper for picking up this PR!

Issue #10520, opened 2021-05-06
headers iterator should have linear not quadratic complexity


ry:
See https://github.com/denoland/deno/pull/10199#issuecomment-833465166

https://github.com/denoland/deno/blob/main/extensions/fetch/20_headers.js#L178-L207

lucacasonato:
Note that the algorithm still has to follow the spec: https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine.

This should be doable with O(2n) instead of O(n^2) complexity though.

PR #10526, opened 2021-05-07
perf(headers): improve iterator complexity to O(2n)

satyarohith:
Closes https://github.com/denoland/deno/issues/10520

### DIFF
--- a/extensions/fetch/20_headers.js
+++ b/extensions/fetch/20_headers.js
@@ -178,32 +178,42 @@
     get [_iterableHeaders]() {
       const list = this[_headerList];
 
-      const headers = [];
-      const headerNamesSet = new Set();
+      // The order of steps are not similar to the ones suggested by the
+      // spec but produce the same result.
+      const headers = {};
+      const cookies = [];
       for (const entry of list) {
-        headerNamesSet.add(byteLowerCase(entry[0]));
-      }
-      const names = [...headerNamesSet].sort();
-      for (const name of names) {
-        // The following if statement, and if block of the following statement
-        // are not spec compliant. `set-cookie` is the only header that can not
-        // be concatentated, so must be given to the user as multiple headers.
+        const name = byteLowerCase(entry[0]);
+        const value = entry[1];
+        if (value === null) throw new TypeError("Unreachable");
+        // The following if statement is not spec compliant.
+        // `set-cookie` is the only header that can not be concatentated,
+        // so must be given to the user as multiple headers.
         // The else block of the if statement is spec compliant again.
-        if (name == "set-cookie") {
-          const setCookie = list.filter((entry) =>
-            byteLowerCase(entry[0]) === "set-cookie"
-          );
-          if (setCookie.length === 0) throw new TypeError("Unreachable");
-          for (const entry of setCookie) {
-            headers.push([name, entry[1]]);
-          }
+        if (name === "set-cookie") {
+          cookies.push([name, value]);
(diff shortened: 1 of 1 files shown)


extensions/fetch/20_headers.js
```diff
@@ -178,32 +178,42 @@
...
-        // are not spec compliant. `set-cookie` is the only header that can not
-        // be concatentated, so must be given to the user as multiple headers.
+        const name = byteLowerCase(entry[0]);
+        const value = entry[1];
+        if (value === null) throw new TypeError("Unreachable");
+        // The following if statement is not spec compliant.
+        // `set-cookie` is the only header that can not be concatentated,
+        // so must be given to the user as multiple headers.
         // The else block of the if statement is spec compliant again.
         if (name == "set-cookie") {
```
magurotuna, line 193: Not related to this patch, but I'm wondering if there's a reason for using `==` rather than `===` here. I guess `===` would be much better?

satyarohith: Updated. :)

extensions/fetch/20_headers.js
```diff
@@ -178,32 +178,42 @@
...
-          if (value === null) throw new TypeError("Unreachable");
-          headers.push([name, value]);
+          // The below if/else block has the same behaviour as getHeaders()
+          // at the end of loop. But it avoids looping through the entire
+          // list to combine multiple values with same header name. It
+          // instead gradually combines them as they are found.
+          if (headers[name] && headers[name].length > 0) {
+            const oldValues = headers[name].split("\x2C\x20");
+            oldValues.push(value);
+            headers[name] = oldValues.join("\x2C\x20");
```
bnoordhuis, line 203:
```suggestion
            headers[name] += ', ' + value;
```
Or is there some subtlety that I'm missing here?

Potential micro-optimization: cache `headers[name]` in a local variable instead of accessing it twice in the `if` clause.

V8's optimizing tiers can optimize out redundant accesses but not always.

extensions/fetch/20_headers.js
```diff
@@ -178,32 +178,42 @@
...
+          } else {
+            headers[name] = value;
+          }
         }
       }
-      return headers;
+
+      return [...Object.entries(headers), ...cookies].sort((a, b) => {
```
bnoordhuis, line 210: I confess ignorance: does the spec require alphasort?

bnoordhuis (approved):

extensions/fetch/20_headers.js
lucacasonato: Yes: https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine -> https://fetch.spec.whatwg.org/#convert-header-names-to-a-sorted-lowercase-set

lucacasonato: But the internal header list may not yet be sorted, because the original sort order is important for the order of headers on the wire.

bnoordhuis: Thanks, duly noted. What a weird requirement though. Do you happen to know what the rationale is? I'm assuming there is one.

lucacasonato: No idea. Maybe to not leak the ordering of headers on the wire into user land? No idea though.

extensions/fetch/20_headers.js
satyarohith: Thanks - updated.

PR #10527, opened 2021-05-07
feat(extensions): Broadcast Channel API

crowlKats:
Current approach is wrong and doesn't work

### DIFF
--- a/extensions/broadcast_channel/01_broadcast_channel.js
+++ b/extensions/broadcast_channel/01_broadcast_channel.js
@@ -1,0 +1,117 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+"use strict";
+
+((window) => {
+  const core = window.Deno.core;
+  const webidl = window.__bootstrap.webidl;
+
+  const handlerSymbol = Symbol("eventHandlers");
+  function makeWrappedHandler(handler) {
+    function wrappedHandler(...args) {
+      if (typeof wrappedHandler.handler !== "function") {
+        return;
+      }
+      return wrappedHandler.handler.call(this, ...args);
+    }
+    wrappedHandler.handler = handler;
+    return wrappedHandler;
+  }
+  // TODO(lucacasonato) reuse when we can reuse code between web crates
+  function defineEventHandler(emitter, name) {
+    // HTML specification section 8.1.5.1
+    Object.defineProperty(emitter, `on${name}`, {
+      get() {
+        return this[handlerSymbol]?.get(name)?.handler;
+      },
+      set(value) {
+        if (!this[handlerSymbol]) {
+          this[handlerSymbol] = new Map();
+        }
+        let handlerWrapper = this[handlerSymbol]?.get(name);
+        if (handlerWrapper) {
+          handlerWrapper.handler = value;
+        } else {
+          handlerWrapper = makeWrappedHandler(value);
+          this.addEventListener(name, handlerWrapper);
+        }
+        this[handlerSymbol].set(name, handlerWrapper);
+      },
+      configurable: true,
+      enumerable: true,
+    });
+  }
+
+  const _name = Symbol("[[name]]");
+  const _closed = Symbol("[[closed]]");
+  const _rid = Symbol("[[rid]]");
+
+  class BroadcastChannel extends EventTarget {
+    [_name];
+    [_closed] = false;
+    [_rid];
+
+    get name() {
+      return this[_name];
+    }
+
+    constructor(name) {
+      super();
+
+      window.location;
+
+      const prefix = "Failed to construct 'broadcastChannel'";
+      webidl.requiredArguments(arguments.length, 1, { prefix });
+
+      this[_name] = webidl.converters["DOMString"](name, {
+        prefix,
+        context: "Argument 1",
+      });
+
+      this[_rid] = core.opSync("op_broadcast_open", this[_name]);
+
+      this[webidl.brand] = webidl.brand;
+
+      this.#eventLoop();
+    }
+
+    postMessage(message) {
+      webidl.assertBranded(this, BroadcastChannel);
+
+      if (this[_closed]) {
+        throw new DOMException("Already closed", "InvalidStateError");
+      }
+
+      core.opAsync("op_broadcast_send", this[_rid], core.serialize(message));
+    }
+
+    close() {
+      webidl.assertBranded(this, BroadcastChannel);
+
+      this[_closed] = true;
+      core.close(this[_rid]);
+    }
+
+    async #eventLoop() {
+      while (!this[_closed]) {
+        const message = await core.opAsync(
+          "op_broadcast_next_event",
+          this[_rid],
+        );
+
+        if (message.length !== 0) {
+          const event = new MessageEvent("message", {
+            data: core.deserialize(message),
+            origin: window.location,
+          });
+          event.target = this;
+          this.dispatchEvent(event);
+        }
+      }
+    }
+  }
+
+  defineEventHandler(BroadcastChannel.prototype, "message");
+  defineEventHandler(BroadcastChannel.prototype, "messageerror");
+
+  window.__bootstrap.broadcastChannel = { BroadcastChannel };
+})(this);
--- a/extensions/broadcast_channel/01_broadcast_channel.js
+++ b/extensions/broadcast_channel/01_broadcast_channel.js
@@ -4,6 +4,7 @@
 ((window) => {
   const core = window.Deno.core;
   const webidl = window.__bootstrap.webidl;
+  const { setTarget } = window.__bootstrap.event;
 
   const handlerSymbol = Symbol("eventHandlers");
   function makeWrappedHandler(handler) {
@@ -21,7 +22,10 @@
     // HTML specification section 8.1.5.1
     Object.defineProperty(emitter, `on${name}`, {
       get() {
-        return this[handlerSymbol]?.get(name)?.handler;
+        // TODO(bnoordhuis) The "BroadcastChannel should have an onmessage
+        // event" WPT test expects that .onmessage !== undefined. Returning
+        // null makes it pass but is perhaps not exactly in the spirit.
+        return this[handlerSymbol]?.get(name)?.handler ?? null;
       },
       set(value) {
         if (!this[handlerSymbol]) {
@@ -43,12 +47,56 @@
 
   const _name = Symbol("[[name]]");
   const _closed = Symbol("[[closed]]");
-  const _rid = Symbol("[[rid]]");
+
+  const channels = [];
+  let rid = null;
+
+  async function recv() {
+    while (channels.length > 0) {
+      const message = await core.opAsync("op_broadcast_recv", rid);
+
+      if (message === null) {
+        break;
+      }
+
+      const [name, data] = message;
+      dispatch(null, name, new Uint8Array(data));
+    }
+
+    core.close(rid);
+    rid = null;
+  }
+
+  function dispatch(source, name, data) {
+    for (const channel of channels) {
+      if (channel === source) continue; // Don't self-send.
+      if (channel[_name] !== name) continue;
+      if (channel[_closed]) continue;
+
+      const go = () => {
+        if (channel[_closed]) return;
+        const event = new MessageEvent("message", {
+          data: core.deserialize(data), // TODO(bnoordhuis) Cache immutables.
+          origin: "http://127.0.0.1",
+        });
+        setTarget(event, channel);
+        channel.dispatchEvent(event);
+      };
+
+      defer(go);
+    }
+  }
+
+  // Defer to avoid starving the event loop. Not using queueMicrotask()
+  // for that reason: it lets promises make forward progress but can
+  // still starve other parts of the event loop.
+  function defer(go) {
+    setTimeout(go, 1);
+  }
 
   class BroadcastChannel extends EventTarget {
     [_name];
     [_closed] = false;
-    [_rid];
 
     get name() {
       return this[_name];
@@ -56,8 +104,6 @@
 
     constructor(name) {
       super();
-
-      window.location;
 
       const prefix = "Failed to construct 'broadcastChannel'";
       webidl.requiredArguments(arguments.length, 1, { prefix });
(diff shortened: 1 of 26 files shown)


extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,117 @@
...
+    [_rid];
+
+    get name() {
+      return this[_name];
+    }
+
+    constructor(name) {
+      super();
+
+      window.location;
```
bnoordhuis, line 60: Since the storage is file-based, does gating it on `window.location` make sense? That's always an http or https URL, right?

bnoordhuis (commented):
I'm getting BorrowMutErrors when I try to postMessage. I'll investigate.
```
thread 'main' panicked at 'already borrowed: BorrowMutError', runtime/metrics.rs:189:35
```

extensions/broadcast_channel/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+#[async_trait]
+impl BroadcastChannel for InMemoryBroadcastChannel {
+  type Resource = InMemoryBroadcastChannelResource;
+
+  fn open(name: String) -> Result<Self::Resource, AnyError> {
+    static ID: AtomicU64 = AtomicU64::new(0);
+    lazy_static::lazy_static! {
+      static ref SENDER: Mutex<Sender<Event>> =
+        Mutex::new(broadcast::channel(64).0);
+    }
```
lucacasonato, line 52: I have some reservations about this. This state should be stored in OpState / ProcessState, just like BlobURLStore. If it is just process global, it will break in all kinds of ways in `deno test`

bnoordhuis: You mean ProgramState? I'll look into that.

cli/tests/integration_tests.rs
```diff
@@ -2682,6 +2682,24 @@ console.log("finish");
...
+  #[test]
+  fn broadcast_channel() {
+    let status = util::deno_cmd()
+      .current_dir(util::tests_path())
+      .arg("test")
+      .arg("--reload")
+      .arg("--location")
+      .arg("http://127.0.0.1/")
+      .arg("--allow-read")
+      .arg("--no-check") // TS typechecking makes the test hang.
```
bnoordhuis, line 2694: Because it tries to typecheck the inline script.

extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,171 @@
...
+        channel.dispatchEvent(event);
+      };
+
+      defer(go);
+    }
+  }
+
+  // Defer to avoid starving the event loop. Not using queueMicrotask()
+  // for that reason: it lets promises make forward progress but can
+  // still starve other parts of the event loop.
```
bnoordhuis, line 95: I observe that with `queueMicrotask()` e.g. `setTimeout()` callbacks don't run when ping-ponging messages over a BroadcastChannel.

extensions/web/02_event.js
```diff
@@ -1117,6 +1117,10 @@
   }
 
   class MessageEvent extends Event {
+    get source() {
+      return null;
+    }
```
bnoordhuis, line 1122: Added to appease WPT.

tools/wpt.ts
```diff
@@ -539,7 +539,25 @@ function discoverTestsToRun(
...
+      const expectation = Array.isArray(parentExpectation) ||
+          typeof parentExpectation == "boolean"
+        ? parentExpectation
+        : parentExpectation[key];
+
+      if (
+        Array.isArray(entry) &&
+        key.endsWith(".html") &&
+        typeof expectation === "boolean"
+      ) {
```
bnoordhuis, line 551: Makes it possible to run .html files directly.

bnoordhuis (commented):
@lucacasonato This is ready for review. I've left some comments.

tools/wpt.ts
crowlKats: any reason why not opening a PR upstream to change the tests to js? its what we have done until now

bnoordhuis: This should be easier going forward, right?

extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,171 @@
...
+
+    postMessage(message) {
+      webidl.assertBranded(this, BroadcastChannel);
+
+      if (this[_closed]) {
+        throw new DOMException("Already closed", "InvalidStateError");
+      }
+
+      const prefix = "Failed to execute 'postMessage' on 'BroadcastChannel'";
+      webidl.requiredArguments(arguments.length, 1, { prefix });
```
lucacasonato, line 141: This should happen before the `if (this[_closed])` check.

extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,171 @@
...
+      this[_name] = webidl.converters["DOMString"](name, {
+        prefix,
+        context: "Argument 1",
+      });
+
+      this[webidl.brand] = webidl.brand;
+
+      channels.push(this);
+
+      if (rid === -1) {
```
lucacasonato, line 125: Instead having `-1` be the rid for "rid not yet created", can you change it to `null`?

extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,171 @@
...
+      webidl.requiredArguments(arguments.length, 1, { prefix });
+
+      if (typeof message === "function" || typeof message === "symbol") {
+        throw new DOMException("Uncloneable value", "DataCloneError");
+      }
+
+      const data = core.serialize(message);
+
+      // Send to other listeners in this VM.
+      dispatch(this, this[_name], new Uint8Array(data));
```
lucacasonato, line 150: Is this a perf optimization? Could this in theory go via the Rust side too?

extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,171 @@
...
+      // Send to listeners in other VMs.
+      defer(() => core.opAsync("op_broadcast_send", [rid, this[_name]], data));
+    }
+
+    close() {
+      webidl.assertBranded(this, BroadcastChannel);
+
+      const index = channels.indexOf(this);
+      if (index !== -1) channels.splice(index, 1);
+      if (channels.length === 0) core.opSync("op_broadcast_unsubscribe", rid);
```
lucacasonato, line 161: Because it is sync, instead of a special `op_broadcast_unsubscribe` op, could this not be part of the `close` lifecycle hook of the broadcast channel resource?

EDIT: I guess because the close lifecycle hook doesnt provide access to op state?

extensions/broadcast_channel/in_memory_broadcast_channel.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
lucacasonato, line 1: What about feature flagging this entire file, and enabling it by default?

extensions/broadcast_channel/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+    .get::<BC::Resource>(rid)
+    .ok_or_else(bad_resource_id)?;
+  let bc = state.borrow::<BC>();
+  bc.unsubscribe(&resource)
+}
+
+pub async fn op_broadcast_send<BC: BroadcastChannel + 'static>(
+  state: Rc<RefCell<OpState>>,
+  (rid, name): (ResourceId, String),
+  buf: Option<ZeroCopyBuf>,
```
lucacasonato, line 71: If I'm not mistaken, the new op signatures should allow for you to remove the `Option<>` around `ZeroCopyBuf`.

extensions/broadcast_channel/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+) -> Result<ResourceId, AnyError> {
+  let bc = state.borrow::<BC>();
+  let resource = bc.subscribe()?;
+  Ok(state.resource_table.add(resource))
+}
+
+pub fn op_broadcast_unsubscribe<BC: BroadcastChannel + 'static>(
+  state: &mut OpState,
+  rid: ResourceId,
+  _buf: Option<ZeroCopyBuf>,
```
lucacasonato, line 58: I think the new op signatures should allow for this:
```suggestion
  _: (),
```

extensions/broadcast_channel/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+    .ok_or_else(bad_resource_id)?;
+  let bc = state.borrow().borrow::<BC>().clone();
+  let data = buf.unwrap().to_vec();
+  bc.send(&resource, name, data).await
+}
+
+pub async fn op_broadcast_recv<BC: BroadcastChannel + 'static>(
+  state: Rc<RefCell<OpState>>,
+  rid: ResourceId,
+  _bufs: Option<ZeroCopyBuf>,
```
lucacasonato, line 86: Ditto

extensions/web/02_event.js
lucacasonato: Interesting... points to this module being in need of a rewrite :wink:.

tools/wpt.ts
```diff
@@ -539,7 +539,25 @@ function discoverTestsToRun(
     for (const key in parentFolder) {
       const entry = parentFolder[key];
 
-      if (Array.isArray(entry)) {
+      const expectation = Array.isArray(parentExpectation) ||
```
lucacasonato, line 542: Not entirely happy with the implementation here.

1) in most cases `.html` tests are not runnable in Deno because they require DOM. The ones that are not should be moved to `.any.js` tests. WPT supports this effort, because it means more tests will be run from both window and workers (.html tests are just runnable from window).
2) the current implementation does not handle test filtering.
3) it duplicates code with the else statement below.

extensions/broadcast_channel/lib.deno_broadcast_channel.d.ts
```diff
@@ -0,0 +1,55 @@
...
+    type: string,
+    listener: EventListenerOrEventListenerObject,
+    options?: boolean | EventListenerOptions,
+  ): void;
+}
+
+declare var BroadcastChannel: {
+  prototype: BroadcastChannel;
+  new (name: string): BroadcastChannel;
+};
```
lucacasonato, line 55: I'd rather this be a `class`, but I assume you copied this from original definitions.

lucacasonato (commented):
As always, implementation is very clean. Nice work. Mostly just some nitpicks.

I do want to voice some concerns regarding how the resources work. Currently there is one broadcast channel resource for all `BroadcastChannel`s in an isolate. This is unfortunate because it makes the resource sanitizer in tests less effective. Say you create a BC outside a test, then another one in a test, and you don't close the one in the test, then there will be no resource leakage warning. I assume this approach is a lot faster though, because less boundary crossings, so this isn't a blocker for me. Logically one resource per `BroadcastChannel` would make more sense for me though.

bnoordhuis:
> Currently there is one broadcast channel resource for all BroadcastChannels in an isolate.

It's for reasons of correctness, not performance - that's just a happy accident. :-)

WPT requires that messages are delivered to receivers in subscription order but `tokio::sync::broadcast` doesn't make that guarantee (and indeed doesn't seem to work that way.)

If you feel really strongly, I could implement a `tokio::sync::broadcast` with stronger ordering guarantees, but that means more code and complexity.

extensions/broadcast_channel/01_broadcast_channel.js
bnoordhuis: Correctness (without jumping through too many hoops): the message should not be delivered to the sender (no self-send) but the Rust side doesn't really know who the sender is.

I could give every BroadcastChannel its own UUID and filter on that but this is simpler.

lucacasonato:
> If you feel really strongly, I could implement a tokio::sync::broadcast with stronger ordering guarantees, but that means more code and complexity.

Nope, with this background info it makes sense now. Lets leave it as is.

extensions/broadcast_channel/01_broadcast_channel.js
bnoordhuis: It's indeed because there's no op state.

tools/wpt.ts
bnoordhuis: Okay, I'll send a PR to WPT.

lucacasonato: Let me know when you open it. I'll cherrypick it into our fork.

extensions/broadcast_channel/lib.deno_broadcast_channel.d.ts
bnoordhuis: @crowlKats Can you comment?

I note that e.g. `WebSocket` is declared the same way.

crowlKats: Well for WebSocket we just copied the typings from the original ts definitions, but it seems that at some point someone changed it to use `class` instead of `interface`

tools/wpt.ts
bnoordhuis: https://github.com/web-platform-tests/wpt/pull/29014

lucacasonato: Updated submodule, and reverted the relevant changes in tools/wpt.ts. Thanks :-)

bnoordhuis:
@lucacasonato How do I get the linter to pass? It's using an existing deno binary that doesn't know about the new `BroadcastChannel` global... Catch-22.

lucacasonato:
@bnoordhuis Fixed now through merging #10652.

bnoordhuis:
The integration test appears to be hanging on the CI but I can't reproduce locally, it always passes (and in less than a second):
```
2021-05-19T22:28:52.2896211Z test integration::broadcast_channel has been running for over 60 seconds
```
I've added an `#[ignore]` to see if it fares better.

bnoordhuis:
All green with the integration test disabled.

@lucacasonato Ideas on how to move forward? I don't mind investigating further but I don't have any leads and I'd like to get this merged before the inevitable next round of merge conflicts.

cli/tests/integration_tests.rs
```diff
@@ -2715,6 +2715,25 @@ console.log("finish");
     assert!(end - start < Duration::new(10, 0));
   }
 
+  #[test]
+  #[ignore] // Hangs on CI, passes locally.
+  fn broadcast_channel() {
```
lucacasonato, line 2720: So this is a integration test instead of a JS unit test because it requires --ignore?

bnoordhuis: Because it requires `--location`.

lucacasonato: Oh... don't we run unit tests with `--location`? If not, @bartlomieju any objections to doing that?

caspervonb: We do:
```
--location=http://js-unit-tests/foo/bar
```

lucacasonato:
I will take a quick look and see if I can get it to fail / pass, but if not let's merge it and investigate later.

caspervonb:
Deadlock somewhere? haven't read through the implementation yet but test hangs for me.

cli/tests/integration_tests.rs
bnoordhuis: Oh, I remember the other issue: the test hangs when run as a unit test because `deno test` tries to type-check the embedded script. I guess I'll move it to a separate file.

lucacasonato:
Found it - wrong assumption about worker startup being sync.

bnoordhuis:
Thanks for helping debug. I've turned it into a unit test and Luca pointed out how to fix the race.

extensions/broadcast_channel/01_broadcast_channel.js
```diff
@@ -0,0 +1,172 @@
...
+    [_closed] = false;
+
+    get name() {
+      return this[_name];
+    }
+
+    constructor(name) {
+      super();
+
+      window.location;
```
lucacasonato, line 111: This should be done in Rust.

lucacasonato:
Tried it out locally, and seems to work well. I have just two concerns left to address now:

1. this should be `--unstable` until we decide about cross process or not
2. the `--location` check and origin namespacing should happen in rust, because other implementers might not have a window origin

bnoordhuis:
I've added a `--unstable` check in `op_broadcast_subscribe()`. That's the main entry point.

Since it's behind `--unstable` now and because I don't want to think about `--location` too much, I've simply ripped out the `window.location` check.

lucacasonato (approved):
LGTM - the origin handling is incorrect (as shown by the broken WPT), but that is fine for now. We will need to revisit it when we do cross process `BroadcastChannel`.

Thanks for sitting through this @bnoordhuis :-)

bnoordhuis:
Sorry @crowlKats, something went wrong while merging. I used rebase-without-squash with the intent of keeping you as the author of the first commit but for some reason I show up as the author. Mea culpa.

PR #10529, opened 2021-05-08
feat(plugin): add tests for plugin args

cvng:
This PR add tests for plugin args with the new runtime extensions introduced in #9800.


bartlomieju (approved):
LGTM, thank you @cvng!

PR #10532, opened 2021-05-08
refactor(core): convert decode/encode to builtin ops

AaronO:
Following #10432 encode/decode can now be converted from bindings to ops.

These ops are also simpler than the old bindings, pretty much all the heavy lifting is done by the op-layer.

## Other changes

- Simplified BOM check
- Removed an extra alloc in `encode()` so it might be moderately faster
- Improved `serde_v8` error handling in `opcall()` (so `serde_v8` serialization errors are thrown in JS)


bartlomieju:
This PR will conflict with https://github.com/denoland/deno/pull/10844

bartlomieju (commented):
I tried rebasing this PR a couple times and have trouble doing so... I'm also not sure if it's worth changing those bindings into ops - especially what impact on the benchmarks it would have. @AaronO I'm going to close this PR for now, but feel free to reopen if you want to land it.

Issue #10534, opened 2021-05-08
plugins: pass `serve_json::Value` to op


cvng:
As discussed with @AaronO, passing a `serde_json::Value` to an op crash the program right now.

```
➜ ../target/debug/deno --version
deno 1.9.2 (debug, x86_64-apple-darwin)
v8 9.1.269.5
typescript 4.2.2
```

steps to reproduce are here https://github.com/cvng/deno/pull/2/files:

```
➜ cd test_plugin
➜ cargo build && ../target/debug/deno run --allow-plugin --unstable tests/test.js debug
    Finished dev [unoptimized + debuginfo] target(s) in 0.32s
Plugin rid: 3
[1]    4574 segmentation fault  ../target/debug/deno run --allow-plugin --unstable tests/test.js debug
```

Pandawan:
I'm encountering a similar issue while building the `test_plugin`. 

I downloaded the test_plugin directory from source and modified the `Cargo.toml` to use `deno_core = "0.87.0"` (and removed the `test_util`). Nothing else was changed.

When running the `test.js`, I also get a segmentation fault. 
After a bit of research, I found out that this seems to happen whenever:
1. The `args` parameter of the op function has some kind of "complex value" (e.g. String, custom struct, maybe even numbers like i32, etc.) but works fine with a bool for example.
2. The `Result`'s return value (the Ok one) is a "complex value", but again it works with a bool.

I'm guessing this has something to do with the serializing/deserializing that deno handles before/after calling the op function.
Unfortunately seg fault doesn't really tell me what/where it's going wrong so I can't dig further.

For reference, I am on macOS 11.3 Big Sur (Intel chip), with rustc 1.52.0 (88f19c6da 2021-05-03) and Deno 1.10.1
and @bartlomieju was able to reproduce this on their macOS Catalina machine.

I've attached a ZIP of a minimal reproducible example if that helps: [deno_plugin.zip](https://github.com/denoland/deno/files/6470083/deno_plugin.zip)

AaronO:
@cvng Could you reword this issue's title ? To include "segfault" or something to that effect

AaronO:
Some initial investigating narrows the segfault to the following callpath:
```
serde_v8::Deserializer::deserialize_map > obj.get_own_property_names()
                                          rusty_v8::Object::get_own_property_names()
```

PR #10536, opened 2021-05-08
feat(plugin): add tests for `serde_json::Value` in plugin args

cvng:
This PR add tests for passing unstructured args (eg. `serde_json::Value`) to a plugin op.

Follow up #10529 

Blocking on #10534


bartlomieju:
@cvng thank you for the PR, all contributions are much appreciated! However, given #10908 and the fact that we want to remove native plugins I will close this PR without a merge.

cvng:
@bartlomieju Sure, we can close it now. This PR was mostly a reproduction for #10534, but we are not going this way anymore.

PR #10541, opened 2021-05-08
fix(cli): typings for arch

eliassjogreen:
Adds `aarch64` arch


lucacasonato (approved):
LGTM, but @kitsonk and @bartlomieju PTAL too, because potential breakage because of type changes.

kitsonk:
The one text fixture in the lsp needs to be updated at least. Tried to pick a fairly static part of the deno.ns to use, but this changes that.

kitsonk (approved):
LGTM

PR #10547, opened 2021-05-09
Switch grant/deny prompt to yes/no

ry:
y/n is more common

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -1054,9 +1054,9 @@ fn permission_prompt(message: &str) -> bool {
   if !atty::is(atty::Stream::Stdin) || !atty::is(atty::Stream::Stderr) {
     return false;
   };
-  let opts = "[g/d (g = grant, d = deny)] ";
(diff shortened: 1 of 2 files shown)


runtime/permissions.rs
```diff
@@ -1054,9 +1054,9 @@ fn permission_prompt(message: &str) -> bool {
   if !atty::is(atty::Stream::Stdin) || !atty::is(atty::Stream::Stderr) {
     return false;
   };
-  let opts = "[g/d (g = grant, d = deny)] ";
+  let opts = "[y/n (y = yes allow, n = no deny)] ";
```
caspervonb, line 1057: Any reason to keep deny in the string?

```suggestion
  let opts = "[y/n (y = yes allow, n = no)] ";
```

caspervonb: Hmm given that the prompt is a yes/no question, seems redundant and kinda stutters for me but maybe that's just me.

Either way is fine.

PR #10555, opened 2021-05-10
fix(lsp): remove code_action/diagnostics deadlock

kitsonk:
This PR refactors the whole of `lsp/diagnostics.rs` to clean it up as well as remove situations where the language server can end up in a deadlock while processing code actions.  Code actions require access to the current TypeScript diagnostics for a given file to determine if there are multiple fixable diagnostics it can perform for a given action.  Previously when the language server was running slow (a slow machine or complex TypeScript project) it could end up in a situation where there was a request to update the diagnostics locking the diagnostic collection at the same time a code action was trying to acquire a lock to read the diagnostics and this would "hang" the language server.

There is no easy test for this, but the symptom was actually evident in the tests when the refactor "broke" the `test_code_actions`.  The request for the `code_actions` would be blocked until the diagnostic refresh completed, which then ran a risk that we would get into a deadlock state if another refresh were to start while the code actions were being processed.  A delay now has to be added to ensure the diagnostics are updated as an update to diagnostics is now non-blocking for `code_actions` which is the way it should have always been.


cli/lsp/diagnostics.rs
```diff
@@ -269,144 +188,15 @@ impl DiagnosticsServer {
...
-    } else {
-      Err(anyhow!("diagnostic server not started"))
-    }
-  }
-
-  pub fn update(&self) -> Result<(), AnyError> {
-    if let Some(tx) = &self.0 {
+  pub(crate) fn update(&self) -> Result<(), AnyError> {
+    if let Some(tx) = &self.channel {
       tx.send(DiagnosticRequest::Update).map_err(|err| err.into())
```
ry, line 193: Replace `DiagnosticRequest::Update` with `()`

ry (approved):
LGTM modulo the DiagnosticRequest::Update -> () 

This PR should have a test, but it's not yet complete. However we want to get this in for 1.10, therefore I've created https://github.com/denoland/deno/issues/10587 to provide a test.

PR #10570, opened 2021-05-11
docs: fix Deno.watchFs example

caspervonb:


cli/dts/lib.deno.ns.d.ts
```diff
@@ -1961,7 +1961,13 @@ declare namespace Deno {
    *
    * ```ts
    * const watcher = Deno.watchFs("/");
-   * setTimeout(() => { watcher.return(); }, 5000);
+   *
+   * setTimeout(() => {
+   *   if (watcher.return) {
```
kt3k, line 1966: This avoids type error 👍

kt3k (approved):
LGTM

Issue #10279, opened 2021-04-21
Compiled version of script is not cached if it has `x-typescript-types` header


kt3k:
If a remote (typescript) script has `x-typescript-types` http header, the compiled version of that script doesn't seem cached in DENO_DIR.

This script ( https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts ) is an example of such script.

```shellsession
$ curl -i https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts   
HTTP/2 200 
content-type: application/typescript
x-typescript-types: /foo.d.ts
content-length: 9
date: Wed, 21 Apr 2021 03:43:06 GMT
via: HTTP/1.1 deploy
x-deno-ray: asia-northeast1-a::2a2a095a91f325c20c5005505d69644d

export {}
$ deno cache --reload https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts                                                                                                                        deno-snippets $ deno cache --reload https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts
Download https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts
Download https://no-cache-script-repro-bf1c801d.deno.dev/foo.d.ts
Check https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts
$ deno cache https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts 
Check https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts # <= this check shouldn't happen
```

source of above server (this uses Deno Deploy): https://gist.githubusercontent.com/kt3k/62437576e951bac97f49d716cf7ac0c2/raw/1660096bd15d50a4d4cdab136a2b286b44134604/deploy.ts

Because of this issue, all dependent scripts of those scripts are very slow to start because they always trigger type checking at invocation.

kitsonk:
`X-TypeScript-Types` make no sense for TypeScript files.  Is there a legitimate use case where they should be served?  `X-TypeScript-Types` are only designed to be used with JavaScript files.

kt3k:
> X-TypeScript-Types make no sense for TypeScript files. Is there a legitimate use case where they should be served?

I don't think there's a legitimate use case.

The actual example I found is this one `https://cdn.dreg.dev/package/@types/estree@0.0.45` and this causes dependent tools slow, which includes `denopack`, `dext` for example.

Maybe this can be an issue of `cdn.dreg.dev`.

[...]

PR #10574, opened 2021-05-11
fix(cli): ignore x-typescript-types header when media type is not js/jsx

kt3k:
This PR changes the behavior of file_fetcher. Now file_fetcher doesn't follow the dts script referenced from `x-typescript-types` header when the script is not javascript or jsx.

Currently deno fetches dts file in x-typescript-types even when the script is typescript (you can check the behavior with this repro url `https://no-cache-script-repro-bf1c801d.deno.dev/foo.ts`), but dts only makes sense for javascript and jsx. So this PR prevents such redundant dts loading.

closes https://github.com/denoland/deno/issues/10279


kitsonk (approved):
LGTM

PR #10581, opened 2021-05-11
fix(cli): always allow documentation modules to be checked

caspervonb:
If no test modules are selected, documentation blocks won't be checked.

For example:
```
deno test --doc cli/dts/
No matching test modules found
```

This fixes that.


caspervonb:
A test case for this is running the command on `cli/dts` but, they're not currently passing.

caspervonb:
Blocked on #10647

cli/tests/integration_tests.rs
```diff
@@ -18,6 +18,20 @@ use tempfile::TempDir;
 use test_util as util;
 use tokio::task::LocalSet;
 
+#[test]
+fn dts_doc() {
+  let status = util::deno_cmd()
+    .arg("test")
+    .arg("--doc")
+    .arg("--unstable")
+    .arg(util::root_path().join("cli/dts"))
```
bartlomieju, line 27: I'd prefer to run explicitly only our own declaration files, otherwise we'd be running declarations from coming from TS. WDYT Casper?

caspervonb: Yeah turns out they're not passing.
Moving this out to https://github.com/denoland/deno/pull/10707

Need a new test case for this issue.

bartlomieju (approved):

PR #10586, opened 2021-05-11
docs: Add Support and Conferences section

getspooky:
- Add Conferences dedicated to the discussion of Deno and the Deno ecosystem.
- Move `Help` to `Community/support`.


kt3k:
@getspooky I feel this kind of materials are more appropriate in community space like [awesome-deno](https://github.com/denolib/awesome-deno). What do you think?

bartlomieju:
@getspooky thanks for the PR, but I agree with @kt3k that it should be linked in awesome-deno instead. Closing without a merge.

PR #10588, opened 2021-05-11
ci: disable ./target/ cache temporarily

kt3k:
Rust team (temporarily) disabled the entire incremental build feature at 1.52.1 (ref: https://blog.rust-lang.org/2021/05/10/Rust-1.52.1.html ). So after #10583, incremental build doesn't work and `target/` cache is waste of CI time. This PR disable it and saves 1~2 minutes of CI time.

Note: This change should be reverted when the above issue is fixed in upstream.

Please merge on approval.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,42 +190,6 @@ jobs:
           key:
             a-cargo-home-${{ matrix.os }}-${{ hashFiles('Cargo.lock') }}
 
-      # In main branch, always creates fresh cache
-      - name: Cache build output (main)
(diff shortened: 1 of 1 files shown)


piscisaureus:
I don't think that's what they meant. 

`Cargo.toml` contains a line that says `incremental = true`, we could remove that. 
It will cause the `target/debug/incremental` directory to no longer be used/created. 

The other stuff in `target/debug` should still be cached.

kt3k:
I'll check this in more details. Thank you for the input. Closing for now.

PR #10589, opened 2021-05-11
feat(cli): add location data dir to deno info

### LINKED ISSUE #10585: `deno info` should display the current path of the Local Storage file
I think it would be a good idea to add a new item to `deno info` to display the path of the Local Storage file.
It might be useful when a user wants to quickly remove it. Also good for transparency.

crowlKats:
closes #10585


kt3k:
The implementation and tests look good to me 👍 

I'm a little worried about the word 'Location data'. Was that documented in our manual? What about 'Web Storage data location' as it only includes web storage data for now

crowlKats:
@kt3k maybe 'origin data' would be better?

crowlKats:
I changed it to `origin data` as it is a more fitting and less misleading name. Ideally the `location_data` dir in `DENO_DIR` should be renamed to `origin_data` as well, but since it would be breaking, i added a comment for 2.0

cli/flags.rs
```diff
@@ -791,6 +791,7 @@ TypeScript compiler cache: Subdirectory containing TS compiler output.",
     .arg(Arg::with_name("file").takes_value(true).required(false))
     .arg(reload_arg().requires("file"))
     .arg(ca_file_arg())
+    .arg(location_arg())
```
nayeemrmn, line 794:
```suggestion
    .arg(location_arg()).conflicts_with("file")
```

cli/tests/041_info_flag.out
```diff
@@ -2,3 +2,4 @@ DENO_DIR location: "[WILDCARD]"
 Remote modules cache: "[WILDCARD]deps"
 Emitted modules cache: "[WILDCARD]gen"
 Language server registries cache: "[WILDCARD]registries"
+Origin data location: "[WILDCARD]location_data"
```
nayeemrmn, line 5: This should be individual paths to each DB file IMO (assuming there'll be another for indexeddb in the future)
```suggestion
Origin local storage: "[WILDCARD]local_storage"
```

crowlKats: `local_storage` is a file in `location_data` under a hashed `--location`. indexeddb will be another file in the same dir, and same for other APIs that will require to have files on the system based on the origin (ie broadcast channel IPC)

nayeemrmn: Yes it was a mistake to highlight this test. What will the output for `deno info --location https://foo/` be? I think it should be a direct path to the local storage db file, if it isn't already:
```
Local storage: /home/nayeem/.cache/deno/location_data/<hash>/local_storage
IndexedDB: /home/nayeem/.cache/deno/location_data/<hash>/indexed_db
```

crowlKats: Currently it is
```
$  deno info --location=https://deno.land
Origin data location: "/Users/crowlkats/Library/Caches/deno/location_data/491d1c5f8d5eaa49d0ee3166c5aca7fbf9af5c466365713577ab3c739342712c"
```
I dont think we should have an entry for each API that will need to use persistent data based on the origin

nayeemrmn: It's more like what we do with the other cache files. We should be as specific as possible and only stop where the structure within a directory is free-form, otherwise we are withholding info. It's just as unneeded as the 2nd and 3rd fields in:
```
DENO_DIR location: "/home/nayeem/.cache/deno"
Remote modules cache: "/home/nayeem/.cache/deno/deps"
Emitted modules cache: "/home/nayeem/.cache/deno/gen"
Language server registries cache: "/home/nayeem/.cache/deno/registries"
```

nayeemrmn: Actually in that vein we should have the following fields:
```
Web storage: /home/nayeem/.cache/deno/location_data/<hash>
Local storage: /home/nayeem/.cache/deno/location_data/<hash>/local_storage
IndexedDB: /home/nayeem/.cache/deno/location_data/<hash>/indexed_db
```
It allows for more concise and less assuming consumption.

crowlKats: > `Web storage: /home/nayeem/.cache/deno/location_data/<hash>`

why `Web storage`?

nayeemrmn: The field is there same as `DENO_DIR location` for easy deletion. Or the name? `Origin data` would be more accurate for the broadcast channel stuff, that's fine.

crowlKats: @nayeemrmn PTAL

cli/tests/041_info_flag_location.out
```diff
@@ -0,0 +1,6 @@
+DENO_DIR location: "[WILDCARD]"
+Remote modules cache: "[WILDCARD]deps"
+Emitted modules cache: "[WILDCARD]gen"
+Language server registries cache: "[WILDCARD]registries"
```
nayeemrmn, line 4: The location variant shouldn't include the global fields, for modularity's sake.
```suggestion
```

crowlKats: Done

cli/tests/041_info_flag_location.out
```diff
@@ -0,0 +1,2 @@
+Origin data: "[WILDCARD]location_data[WILDCARD]"
```
nayeemrmn, line 1: I keep interpreting this as "origin info" as opposed to "per-origin storage bucket". How about:
```suggestion
Origin storage: "[WILDCARD]location_data"
```

cli/tests/041_info_flag.out
```diff
@@ -2,3 +2,4 @@ DENO_DIR location: "[WILDCARD]"
 Remote modules cache: "[WILDCARD]deps"
 Emitted modules cache: "[WILDCARD]gen"
 Language server registries cache: "[WILDCARD]registries"
+Origin data: "[WILDCARD]location_data"
```
nayeemrmn, line 5: Ditto
```suggestion
Origin storage: "[WILDCARD]location_data"
```

cli/main.rs
```diff
@@ -204,11 +204,12 @@ pub fn create_main_worker(
...
     get_error_class_fn: Some(&crate::errors::get_error_class_name),
     location: program_state.flags.location.clone(),
-    location_data_dir: program_state.flags.location.clone().map(|loc| {
+    origin_data_dir: program_state.flags.location.clone().map(|loc| {
       program_state
         .dir
         .root
         .clone()
+        // TODO(@crowlKats): change to origin_data for 2.0
         .join("location_data")
```
nayeemrmn, line 213: Do we need to consider this breaking? I feel like we provide cache paths in `deno info` so we can rename them while providing a stable handle, and this wasn't even in `deno info` yet.

cli/flags.rs
```diff
@@ -791,6 +791,7 @@ TypeScript compiler cache: Subdirectory containing TS compiler output.",
     .arg(Arg::with_name("file").takes_value(true).required(false))
     .arg(reload_arg().requires("file"))
     .arg(ca_file_arg())
+    .arg(location_arg().conflicts_with("file"))
```
nayeemrmn, line 794: This needs a `deno info`-specific help text. I think clap will allow overriding it here.

cli/main.rs
crowlKats: well renaming `location_data` would mean that users that have used local storage would lose their current data

cli/flags.rs
crowlKats: What exactly should the help text contain?

cli/tests/041_info_flag_location.out
crowlKats: Done

cli/flags.rs
nayeemrmn:
```
Show files used for web storage APIs etc. when running a script with '--location=<HREF>'
```

crowlKats: Done

nayeemrmn (approved):
LGTM

cli/flags.rs
```diff
@@ -791,6 +791,11 @@ TypeScript compiler cache: Subdirectory containing TS compiler output.",
     .arg(Arg::with_name("file").takes_value(true).required(false))
     .arg(reload_arg().requires("file"))
     .arg(ca_file_arg())
+    .arg(
+      location_arg()
+        .conflicts_with("file")
+        .help("Show files used for web storage APIs etc. when running a script with '--location=<HREF>'")
```
lucacasonato, line 797:
```suggestion
        .help("Show files used for origin bound APIs like the Web Storage API when running a script with '--location=<HREF>'")
```

cli/main.rs
lucacasonato: I think we can still do this in 1.x - we add a "migration" that renames old to new. Lets do this in another PR though.

cli/main.rs
```diff
@@ -266,37 +267,63 @@ where
...
+        typescript_cache
+      );
+      println!(
+        "{} {:?}",
+        colors::bold("Language server registries cache:"),
+        registry_cache,
+      );
+      println!("{} {:?}", colors::bold("Origin storage:"), origin_dir);
+    } else {
+      println!("{} {:?}", colors::bold("Origin storage:"), origin_dir);
```
lucacasonato, line 320:
```suggestion
    println!("{} {:?}", colors::bold("DENO_DIR location:"), deno_dir);
    println!(
      "{} {:?}",
      colors::bold("Remote modules cache:"),
      modules_cache
    );
    println!(
      "{} {:?}",
      colors::bold("Emitted modules cache:"),
      typescript_cache
    );
    println!(
      "{} {:?}",
      colors::bold("Language server registries cache:"),
      registry_cache,
    );
    println!("{} {:?}", colors::bold("Origin storage:"), origin_dir);
    if location.is_some() {
```

cli/main.rs
```diff
@@ -266,37 +267,63 @@ where
...
-    let output = json!({
+    let output = if location.is_none() {
+      json!({
         "denoDir": deno_dir,
         "modulesCache": modules_cache,
         "typescriptCache": typescript_cache,
         "registryCache": registry_cache,
-    });
+        "originStorage": origin_dir,
+      })
```
lucacasonato, line 292: I feel like location.is_some should still have all of the json output that is irrelevant to the location itself - so denoDir etc.

cli/tests/041_info_flag_location.out
lucacasonato: I disagree with this. I don't understand why this would be more modular than including the other fields. Just seems like you need to run `deno info` multiple times if you want all info, which seems annoying.

nayeemrmn: It's the same reason `deno info mod.ts` doesn't include the global fields. You'd run `deno info --location https://foo/` because you're interested in the local storage file for some specific origin, it's unlikely you'd want the global fields at the same time. `--location` shouldn't be considered a passive modifier of `deno info` since it would keep the other fields the same and just add new more specific ones. That makes `deno info --location <some_location>` more comparable to `deno info <some_module>` and it should be similarly disjoint.

cli/main.rs
```diff
@@ -297,6 +312,14 @@ fn print_cache_info(
       colors::bold("Language server registries cache:"),
       registry_cache,
     );
+    println!("{} {:?}", colors::bold("Origin storage:"), origin_dir);
+    if location.is_some() {
+      println!(
+        "{} {:?}",
+        colors::bold("Local Storage:"),
```
kt3k, line 319: nit: is this capitalization intentional?

kt3k (approved):
LGTM

cli/main.rs
crowlKats: Yes, as "Local Storage" is a proper noun

kt3k:
@crowlKats Thank you for your contribution!

Issue #10603, opened 2021-05-12
lsp: opening multiple-files in a short period causes a deadlock


kitsonk:
When the lsp receives several `textDocument/didOpen` messages in quick succession it can cause a dead-lock, as the lsp is querying the client for configuration information.  This can occur if a user is doing a find/replace across multiple files.  This only affects Deno 1.10.0.

kitsonk:
@ry this diff works the bug but disables the workspace folder config:

```diff
diff --git a/cli/lsp/config.rs b/cli/lsp/config.rs
index bc41fd02a..441b2cc99 100644
--- a/cli/lsp/config.rs
+++ b/cli/lsp/config.rs
@@ -148,6 +148,7 @@ pub struct Config {
 }
 
 impl Config {
+  #[allow(unused)]
   pub fn contains(&self, specifier: &ModuleSpecifier) -> bool {
     self.specifier_settings.contains_key(specifier)
   }
diff --git a/cli/lsp/language_server.rs b/cli/lsp/language_server.rs
index 5ed6f693c..2f090938c 100644
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -594,25 +594,26 @@ impl Inner {
     let specifier = self.url_map.normalize_url(&params.text_document.uri);
 
     // we only query the individual resource file if the client supports it
-    if self.config.client_capabilities.workspace_configuration
-      && !self.config.contains(&specifier)
-    {
-      if let Ok(value) = self
-        .client
-        .configuration(vec![ConfigurationItem {
-          scope_uri: Some(params.text_document.uri.clone()),
-          section: Some(SETTINGS_SECTION.to_string()),
-        }])
-        .await
-      {
-        if let Err(err) = self
-          .config
-          .update_specifier(specifier.clone(), value[0].clone())
-        {
-          warn!("Error updating specifier configuration: {}", err);
-        }
-      }
-    }
+    // TODO(@kitsonk) workaround https://github.com/denoland/deno/issues/10603
+    // if self.config.client_capabilities.workspace_configuration
+    //   && !self.config.contains(&specifier)
+    // {
+    //   if let Ok(value) = self
+    //     .client
+    //     .configuration(vec![ConfigurationItem {
+    //       scope_uri: Some(params.text_document.uri.clone()),
+    //       section: Some(SETTINGS_SECTION.to_string()),
+    //     }])
+    //     .await
+    //   {
+    //     if let Err(err) = self
+    //       .config
[...]

PR #10604, opened 2021-05-12
Revert recent changes to LSP

ry:
We've discovered a serious bug in the #10488. Because of today's release deadline, we may need to rollback all recent changes.

Ref https://github.com/denoland/deno/issues/10603


ry:
closed in favor of https://github.com/denoland/deno/pull/10605

PR #10605, opened 2021-05-12
fix(#10603): revert minimal changes to resolve deadlock bug

kitsonk:
This disables the workspace folder configuration functionality to work-around the deadlock issue introduced in #10488.


ry (approved):
LGTM

PR #10608, opened 2021-05-12
docs(cli/dts): fix `Deno.createHttpClient` example

caspervonb:


kt3k (approved):
Thanks! LGTM

kt3k:
BTW this kind of example probably won't pass doc testing in an usual way, but it still 'works' as an example... Should we have some special control about this kind examples?

caspervonb:
> BTW this kind of example probably won't pass doc testing in an usual way, but it still 'works' as an example... Should we have some special control about this kind examples?

I will add `no-run` when we add running again, mirroring `deno test --no-run` so that's the first step for examples that are more hassle than it's worth to run.

caspervonb:
Oh and client/server dependencies can be handled with hidden lines.

```
# const server = // ...;
const client = // ...;
```

Issue #10575, opened 2021-05-11
lsp: bad diagnostics on triple-slash reference types


kitsonk:
Wow, yes, `deno run` succeeds without issue with `types` instead of `path`.

I am now seeing a different error from the language server though:

> Cannot find type definition file for './a.d.ts'.deno-ts(2688)

Does this need to be reported separately at https://github.com/denoland/vscode_deno?

_Originally posted by @mfulton26 in https://github.com/denoland/deno/issues/10561#issuecomment-837314094_

kitsonk:
I have checked, and the import of a file with the reference works as expected, it is just that TypeScript produces a diagnostic error that is actually meaningless and likely we just need to suppress it.

PR #10610, opened 2021-05-12
fix(lsp): skip type definition not found diagnostic

satyarohith:
Closes https://github.com/denoland/deno/issues/10575

### DIFF
--- a/cli/tsc/99_main_compiler.js
+++ b/cli/tsc/99_main_compiler.js
@@ -149,6 +149,9 @@ delete Object.prototype.__proto__;
     // TS2306: File 'file:///Users/rld/src/deno/cli/tests/subdir/amd_like.js' is
     // not a module.
     2306,
+    // TS2688: Cannot find type definition file for '...'.
+    // We ignore because type defintion files can end with '.ts'.
+    2688,
     // TS2691: An import path cannot end with a '.ts' extension. Consider
     // importing 'bad-module' instead.
     2691,


cli/lsp/diagnostics.rs
```diff
@@ -367,6 +367,15 @@ async fn generate_ts_diagnostics(
...
     for (specifier_str, ts_diagnostics) in ts_diagnostics_map {
+      // Code 2688 represents "Cannot found type definitions file" diagnostic. We want
+      // to skip it. See https://github.com/denoland/deno/issues/10575
+      let mut ts_diagnostics = ts_diagnostics;
+      let type_definition_not_found =
+        ts_diagnostics.iter().position(|d| d.code == 2688);
+      if let Some(index) = type_definition_not_found {
+        ts_diagnostics.remove(index);
+      }
+
```
satyarohith, line 378: @kitsonk any suggestion on where I should add a test for it (or can we skip it)? I feel like an integration test would add too much weight for this small change?

kitsonk (commented):
We already have a space for ignored diagnostics, let's add it there:

https://github.com/denoland/deno/blob/3318c495f670b2c2ad6418276b5aff0781724e16/cli/tsc/99_main_compiler.js#L137-L173

Since the ignored diagnostics are tested via other tests, adding another diagnostic to the above should be sufficient.

kitsonk (approved):
LGTM

PR #10618, opened 2021-05-12
fix(cli/import_map): Don't statically error on dynamic unmapped bare specifiers

### LINKED ISSUE #10168: v1.8.3 - Dynamic import that can't execute causes error when using import maps.

### LINKED ISSUE #10615: Improve the error message about the unmapped bare specifier in import maps
Currently the error message is just:
```
nickolay@frontier:~/workspace/Bryntum/siesta-monorepo/siesta$ deno run -A --unstable --quiet --import-map import_map.json tests/index.js
error: Unmapped bare specifier "path"
```
Probably should include the source position, which is the cause of the error.

### LINKED ISSUE #10616: Adding `--import-map` breaks the code
I have code that works fine without import maps:
```
nickolay@frontier:~/workspace/Bryntum/siesta-monorepo/siesta$ deno run -A --unstable --quiet tests/hook/hook.t.js
Launching test file: tests/hook/hook.t.js in Deno 1.9.2

 PASS  tests/hook/hook.t.js

Test files : 1 passed, 0 failed, 1 total
Time       : 33ms
```
Adding a dummy import map breaks it:
```
nickolay@frontier:~/workspace/Bryntum/siesta-monorepo/siesta$ deno run -A --unstable --quiet --import-map import_map.json tests/hook/hook.t.js
error: Unmapped bare specifier "path"
```

How is this possible? It seems adding a source map changes the set of files being imported somehow. I do have the bare "path" but only in files that are used for Node. And again, the code works fine w/o import maps.

Import map content for the full picture:
```
{
  "imports":  {
    "siesta/":  "./"
  }
}
```

To reproduce:
1) clone the `git@github.com:bryntum/siesta.git`, specific revision is: 6e7e2576e0afcdd3f4f6e1966923ea46c29896eb
2) `npm i` , `npx tsc`
3) `cd examples/deno`
4) Launch w/o import maps: `deno run -A --unstable --quiet --no-check tests/index.ts`
Should see something like:
```
Launching test suite project: Awesome Deno project test suite in Deno 1.9.2

 FAIL  module/zoomer.t.js
├─✘ Using `zoomer` should work
│ └─✘ equal(received, expected)
│   
│     2 | import { zoomer } from "../../src/module.js";
│     3 | it('Using `zoomer` should work', async (t) => {
│   ➤ 4 |     t.equal(zoomer('Myers Courtney', 35), { name: 'Courtney Myers', age: 35 });
│       |       ^
│     5 | });
│   
│   Received                   │ │ Expected                  
│                              │ │                           
│   {                          │ │ {                         
│     "age": 35,               │ │   "age": 35,              
│     "name": "Myers Courtney" │ │   "name": "Courtney Myers"
│   }                          │ │ }                         
│   
```
5) Launch with import maps: `deno run -A --unstable --quiet --no-check --import-map import_map.json tests/index.ts`
The output is:
```
error: Unmapped bare specifier "path"
```

nayeemrmn:
Fixes #10168.
Fixes #10615.
Fixes #10616.

### DIFF
--- a/cli/import_map.rs
+++ b/cli/import_map.rs
@@ -24,7 +24,7 @@ impl fmt::Display for ImportMapError {
     match self {
       ImportMapError::UnmappedBareSpecifier(specifier, maybe_referrer) => write!(
         f,
-        "relative import path \"{}\" not prefixed with / or ./ or ../ and not in import map{}",
+        "Relative import path \"{}\" not prefixed with / or ./ or ../ and not in import map{}",
         specifier,
         match maybe_referrer {
           Some(referrer) => format!(" from \"{}\"", referrer),
--- a/cli/import_map.rs
+++ b/cli/import_map.rs
@@ -14,11 +14,25 @@ use std::error::Error;
 use std::fmt;
 
 #[derive(Debug)]
-pub struct ImportMapError(String);
+pub enum ImportMapError {
+  UnmappedBareSpecifier(String, Option<String>),
+  Other(String),
+}
 
 impl fmt::Display for ImportMapError {
   fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-    f.pad(&self.0)
+    match self {
+      ImportMapError::UnmappedBareSpecifier(specifier, maybe_referrer) => write!(
+        f,
+        "relative import path \"{}\" not prefixed with / or ./ or ../ and not in import map{}",
+        specifier,
+        match maybe_referrer {
+          Some(referrer) => format!(" from \"{}\"", referrer),
+          None => format!(""),
+        }
+      ),
+      ImportMapError::Other(message) => f.pad(message),
+    }
   }
 }
 
@@ -51,7 +65,7 @@ impl ImportMap {
     let v: Value = match serde_json::from_str(json_string) {
       Ok(v) => v,
       Err(_) => {
-        return Err(ImportMapError(
+        return Err(ImportMapError::Other(
           "Unable to parse import map JSON".to_string(),
         ));
       }
@@ -60,7 +74,7 @@ impl ImportMap {
     match v {
       Value::Object(_) => {}
       _ => {
-        return Err(ImportMapError(
+        return Err(ImportMapError::Other(
           "Import map JSON must be an object".to_string(),
         ));
       }
@@ -70,7 +84,7 @@ impl ImportMap {
     let normalized_imports = match &v.get("imports") {
       Some(imports_map) => {
         if !imports_map.is_object() {
-          return Err(ImportMapError(
+          return Err(ImportMapError::Other(
             "Import map's 'imports' must be an object".to_string(),
           ));
         }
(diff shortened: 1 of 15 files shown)


cli/import_map.rs
```diff
@@ -454,10 +454,7 @@ impl ImportMap {
       return Ok(as_url);
     }
 
-    Err(ImportMapError(format!(
-      "Unmapped bare specifier {:?}",
-      specifier
-    )))
+    Ok(None)
```
nayeemrmn, line 457: Not spec compliant, may have to change the call sites instead.

bartlomieju:
I'm entirely sure about this change - it seems to start deviating from the spec algorithm, but on the other hands none of the WPT fail so I guess it's okay.

nayeemrmn:
Which part deviates from the spec algorithm? The two important changes are:
1. Structured import map errors (https://github.com/denoland/deno/pull/10618/files#diff-fba55433053f0ea5eef635f4d576f13beea8b8091c1d692f3513e46e4715e0d3L17-R20).
2. Solidifying `Module::parse()`, which doesn't error on `ModuleResolutionError::ImportPrefixMissing`, to also not error on its equivalent `ImportMapError::UnmappedBareSpecifier` (https://github.com/denoland/deno/pull/10618/files#diff-8439fae0dccaa5fecd30e137ceb78680f3b7884c1ae74a5f7c649f5273abf3e2L384-R391). This was the cause of the bug. I don't think it correlates with any spec.
 
All other changes are 1-to-1 simplifications e.g. the `ImportMap::resolve()` never returns `None`, so the `Option` is removed from its return type.

bartlomieju:
> Which part deviates from the spec algorithm? The two important changes are:
> 
> 1. Structured import map errors (https://github.com/denoland/deno/pull/10618/files#diff-fba55433053f0ea5eef635f4d576f13beea8b8091c1d692f3513e46e4715e0d3L17-R20).
> 2. Solidifying `Module::parse()`, which doesn't error on `ModuleResolutionError::ImportPrefixMissing`, to also not error on its equivalent `ImportMapError::UnmappedBareSpecifier` (https://github.com/denoland/deno/pull/10618/files#diff-8439fae0dccaa5fecd30e137ceb78680f3b7884c1ae74a5f7c649f5273abf3e2L384-R391). This was the cause of the bug. I don't think it correlates with any spec.
> 
> All other changes are 1-to-1 simplifications e.g. the `ImportMap::resolve()` never returns `None`, so the `Option` is removed from its return type.

@nayeemrmn I see, somehow I thought spec required to return `null`/`None` when the resolution fails, but it seems you're just unifying the behavior. 

LGTM then, I will let @kitsonk take a look before landing

kitsonk:
Sorry, for some reason this was missing from my todo list... I will look at first thing tomorrow my time.

kitsonk (approved):
LGTM

PR #10620, opened 2021-05-12
fix(lsp): remove duplicate cwd in config path

### LINKED ISSUE #10619: deno.config path is loaded with duplicate cwd
Setting "deno.config" to "./tsconfig" in `settings.json` results in the lsp server appending the current working directory twice.

noverby:
Fixes #10619

### DIFF
--- a/cli/config_file.rs
+++ b/cli/config_file.rs
@@ -323,8 +323,18 @@ mod tests {
   use deno_core::serde_json::json;
 
   #[test]
-  fn read_config_file() {
+  fn read_config_file_relative() {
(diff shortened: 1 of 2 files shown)


bartlomieju:
@noverby thanks, looks good. Could you please add 2 test cases to `config_file.rs` that exercises code for both absolute and relative paths?

noverby:
@bolatovumar Done. Anything else needed before merge?

bartlomieju (approved):
LGTM, thank you for the fix @noverby!

PR #10625, opened 2021-05-13
fix(lsp): re-enable the per resource configuration without a deadlock

kitsonk:
This PR rewrites the way per resource configuration is handled in the language server.

There isn't specifically a test that tests for this deadlock yet, as I need to refactor the language server integration tests to a new structure that supports 2 way communication, which we have in the benchmark harness, but need to share with the integration tests.

### DIFF
--- a/cli/lsp/config.rs
+++ b/cli/lsp/config.rs
@@ -10,6 +10,7 @@ use deno_core::url::Url;
 use deno_core::ModuleSpecifier;
 use log::error;
 use lspower::lsp;
+use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::sync::Mutex;
@@ -170,7 +171,7 @@ enum ConfigRequest {
 #[derive(Debug, Default, Clone)]
 pub struct Settings {
   pub specifiers:
-    HashMap<ModuleSpecifier, (ModuleSpecifier, SpecifierSettings)>,
+    BTreeMap<ModuleSpecifier, (ModuleSpecifier, SpecifierSettings)>,
   pub workspace: WorkspaceSettings,
 }
 
@@ -179,12 +180,12 @@ pub struct Config {
   pub client_capabilities: ClientCapabilities,
   pub root_uri: Option<Url>,
   settings: Arc<Mutex<Settings>>,
-  tx: mpsc::UnboundedSender<ConfigRequest>,
+  tx: mpsc::Sender<ConfigRequest>,
 }
 
 impl Config {
   pub fn new(client: lspower::Client) -> Self {
-    let (tx, mut rx) = mpsc::unbounded_channel::<ConfigRequest>();
+    let (tx, mut rx) = mpsc::channel::<ConfigRequest>(100);
     let settings = Arc::new(Mutex::new(Settings::default()));
     let settings_ref = settings.clone();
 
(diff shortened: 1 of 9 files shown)


cli/lsp/config.rs
```diff
@@ -139,25 +145,178 @@ impl WorkspaceSettings {
...
+              }];
+              for (_, (uri, _)) in specifiers.iter() {
+                items.push(lsp::ConfigurationItem {
+                  scope_uri: Some(uri.clone()),
+                  section: Some(SETTINGS_SECTION.to_string()),
+                });
+              }
+              let specifiers: Vec<(ModuleSpecifier, ModuleSpecifier)> =
+                specifiers.into_iter().map(|(s, (u, _))| (s, u)).collect();
+              if let Ok(configs) = client.configuration(items).await {
```
bnoordhuis, line 216: Awaits while holding the `settings` mutex. I'd try to avoid that at all costs because it's too easy to (re)introduce deadlocks that way. Would releasing and reacquiring the mutex work?

As a general observation: the critical section is pretty long. Shorter is better.

cli/lsp/config.rs
```diff
@@ -139,25 +145,178 @@ impl WorkspaceSettings {
...
+}
+
+#[derive(Debug)]
 pub struct Config {
   pub client_capabilities: ClientCapabilities,
   pub root_uri: Option<Url>,
-  pub specifier_settings: HashMap<ModuleSpecifier, SpecifierSettings>,
-  pub workspace_settings: WorkspaceSettings,
+  settings: Arc<Mutex<Settings>>,
+  tx: mpsc::UnboundedSender<ConfigRequest>,
```
bnoordhuis, line 182: You're probably using an unbounded queue so you don't have to .await sends? I don't see anything really wrong with that except that I don't really have a feeling for how many requests can get queued up.

kitsonk: I am doing it because that is what we did in the DiagnosticServer, versus an informed decision. The most likely case of a queue of requests would be the situation where you are doing a multi-file file and replace (which is where the deadlock was occurring before) where multiple `didOpen` notifications are sent from the client, which now requires a request back from the server to the client to get the configuration of a resource it hasn't seen before.

cli/lsp/config.rs
kitsonk: I had thought about that, and yeah, you are probably right that it needs to be re-worked to re-acquire the lock. I wrote the code to deal with re-acquiring the lock (e.g. there is a risk something might have changed), but I didn't release it and now that doesn't really make any sense.

Shorter is always better, it expanded trying to deal with the possibility that things could change between when the request was formed and when the response was processed, which added quite a few lines, but I will go over it again and see what can be done to deal with it.  The problem is that hashmaps are unordered which added a level of complexity, but maybe a btree would actually be more useful here, so the keys always come back in a predictable order.

cli/lsp/config.rs
kitsonk: Ok, I moved over to a `channel` versus an `unbound_channel` and put a good guess in there of what the max queue should look like.

kitsonk:
I've added a benchmark test that replicates the issue that I encountered which required the feature to be removed.  I thought about just leaving it as a integration test, but it is a good test to have additional benchmarks for the lsp and it is significantly different than the other ones we have.

cli/lsp/config.rs
```diff
@@ -139,25 +146,189 @@ impl WorkspaceSettings {
...
+
+  pub fn snapshot(&self) -> ConfigSnapshot {
+    ConfigSnapshot {
+      client_capabilities: self.client_capabilities.clone(),
+      root_uri: self.root_uri.clone(),
+      settings: self.settings.lock().unwrap().clone(),
+    }
   }
 
   pub fn specifier_enabled(&self, specifier: &ModuleSpecifier) -> bool {
```
bnoordhuis, line 326: Can I suggest changing these to associated methods that take a MutexGuard? I.e., requiring prove that you hold the lock when you call them?

Imagine a future change where you acquire the lock but forget `specifier_enabled()` does too and then you call it: deadlock.

bnoordhuis (approved):

cli/lsp/config.rs
kitsonk: I get the concept, but am not sure what that would look like, any examples you can point me to?

bnoordhuis: Something like this:
```rust
pub fn specifier_enabled(g: &MutexGuard<Self>, specifier: &ModuleSpecifier) -> bool {
  // ...
}
```
And you'd call like this:
```rust
let g = config.lock().unwrap();
// ...
if Config::specifier_enabled(&g, specifier) {
  // ...
}
// ...

Comment (kitsonk) [reply cli/lsp/config.rs:326]: Ok! I thought there was some magic, like I know you are have an `Arc<Self>`.

Hrmphh... That makes things a lot more complicated, like having to make `settings` public and having to acquire the lock.  What about if I `try_lock` and propagate the result that can be handled by the caller.  That would be cleaner.  The main goal is to just not have a blocking acquire.

Comment (kitsonk) [reply cli/lsp/config.rs:326]: Hrmppph... I don't like that either... as it sort of makes a mess of things and can run into issues... I know we have avoided RwLocks before, but there is really only one part of the code now that can possibly write to settings (actually 2, but they both a very controlled) and having multiple reads is perfectly acceptable in this use case.  So I think a RwLock would actually be much better suited (and safer) here and for something that could be a high-frequency, like checking if enable or "snapshotting" it feels an appropriate use.

Comment (bnoordhuis) [reply cli/lsp/config.rs]: > That makes things a lot more complicated, like having to make settings public and having to acquire the lock.

That's the idea :-) - it moves the decision of when and where to lock to the caller. The caller knows that better than the callee.

There's currently an implicit contract between caller and callee that says "don't lock before you call me because I'm going to." With a MutexGuard, you make that contract explicit.

File: cli/lsp/config.rs [hunk 4]
```diff
@@ -192,23 +368,34 @@ impl Config {
...
-    self.specifier_settings.insert(specifier, settings);
+  pub async fn update_specifier_settings(
+    &self,
+    specifier: &ModuleSpecifier,
+    uri: &ModuleSpecifier,
+  ) -> Result<(), AnyError> {
+    if self.client_capabilities.workspace_configuration
+      && !self
+        .settings
+        .try_read()
```
bnoordhuis, line 379: When would you expect this to fail?

I'd probably just send unconditionally, without checking .specifiers first, and let the receiver handle it.

bnoordhuis (approved):
Reviewed lightly due to time pressure but LGTM.

cli/lsp/config.rs
kitsonk: Yeah, when I was pondering this offline I came to the same thought, it would put an extra message in the queue, but it would get cleared down in a much narrower lock situation.

PR #10627, opened 2021-05-13
feat(lsp): quick fix actions to ignore lint errors

### LINKED ISSUE #10122: lsp: lint diagnostics should have a quick fix to add ignore comment
Lint diagnostics in the LSP should produce a code action quick fix that inserts a comment to ignore the lint rule.

For example, the following:

```ts
declare var a: any;
```

Which generates the diagnostic:

```
`any` type is not allowed
deno-lint(no-explicit-any)
```

Should have a quick fix which would transform the code to:

```ts
// deno-lint-ignore no-explicit-any
declare var a: any;
```

satyarohith:
Introduces quick-fix actions to ignore lint errors.

https://user-images.githubusercontent.com/29819102/122170586-5a4abe80-ce9c-11eb-95b7-a1c939ffdd70.mp4

Closes #10122

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -620,6 +620,7 @@ pub struct DenoFixData {
 #[derive(Debug, Clone)]
 enum CodeActionKind {
   Deno(lsp::CodeAction),
+  DenoLint(lsp::CodeAction),
   Tsc(lsp::CodeAction, tsc::CodeFixAction),
 }
 
@@ -724,7 +725,9 @@ impl CodeActionCollection {
         document_changes: None,
       }),
     };
-    self.actions.push(CodeActionKind::Deno(ignore_error_action));
+    self
+      .actions
+      .push(CodeActionKind::DenoLint(ignore_error_action));
 
     let mut changes = HashMap::new();
     changes.insert(
@@ -757,7 +760,9 @@ impl CodeActionCollection {
         document_changes: None,
       }),
     };
-    self.actions.push(CodeActionKind::Deno(ignore_file_action));
+    self
+      .actions
+      .push(CodeActionKind::DenoLint(ignore_file_action));
 
     Ok(())
   }
@@ -876,6 +881,7 @@ impl CodeActionCollection {
       .map(|i| match i {
         CodeActionKind::Tsc(c, _) => lsp::CodeActionOrCommand::CodeAction(c),
         CodeActionKind::Deno(c) => lsp::CodeActionOrCommand::CodeAction(c),
+        CodeActionKind::DenoLint(c) => lsp::CodeActionOrCommand::CodeAction(c),
       })
       .collect()
   }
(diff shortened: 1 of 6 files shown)


cli/lsp/analysis.rs
```diff
@@ -479,6 +479,91 @@ impl CodeActionCollection {
...
+            line: diagnostic.range.start.line,
+            character: 0,
+          },
+          end: lsp::Position {
+            line: diagnostic.range.start.line,
+            character: 0,
+          },
+        },
+      }],
+    );
```
magurotuna, line 513: Why not `changes.insert(specifier.clone(), vec![...])` ?

cli/lsp/analysis.rs
```diff
@@ -479,6 +479,91 @@ impl CodeActionCollection {
...
+            line: 0,
+            character: 0,
+          },
+          end: lsp::Position {
+            line: 0,
+            character: 0,
+          },
+        },
+      }],
+    );
```
magurotuna, line 547: ditto

cli/lsp/analysis.rs
```diff
@@ -479,6 +479,91 @@ impl CodeActionCollection {
...
+    let mut changes = HashMap::new();
+    HashMap::insert(
+      &mut changes,
+      specifier.clone(),
+      vec![lsp::TextEdit {
+        new_text: format!("// deno-lint-ignore {}\n", code),
+        range: lsp::Range {
+          start: lsp::Position {
+            line: diagnostic.range.start.line,
+            character: 0,
```
dsherret, line 505: There is a way to get the indentation at the current position via tsserver ("indentaiton" command), but I don't think that would be helpful here though. We may just want to get the index of the first non-whitespace character on the current line and that will be good enough for 99% of cases.

cli/lsp/analysis.rs
satyarohith: Not sure why I used them. 😄 

Thanks!

cli/lsp/analysis.rs
satyarohith: Ah, this makes sense. I'll see how I can get the index of the first non-whitespace character of the start line.

Thanks for the suggestion! :)

cli/lsp/analysis.rs
```diff
@@ -665,6 +666,102 @@ impl CodeActionCollection {
...
+        .flatten()
+    } else {
+      None
+    };
+
+    let mut changes = HashMap::new();
+    changes.insert(
+      specifier.clone(),
+      vec![lsp::TextEdit {
+        new_text: append_whitespace(
```
sarcat, line 697:
```suggestion
        new_text: prepend_whitespace(
```

kitsonk (changes requested):
The only change that I think we need to make is to add a `CodeActionKind` of `DenoLint` since we differ the sources of the lint information as well, so I think we should maintain that separation here.

Otherwise really good, thank you!

kitsonk (approved):
LGTM

PR #10633, opened 2021-05-14
docs(cli/dts): fix link from master to main

magurotuna:
The default branch was changed from `master` to `main` a while ago. This link needs to follow the change.


kt3k (approved):
@magurotuna Thanks!

PR #10639, opened 2021-05-15
Put `downloaded percentage`  number to the middle, and remove the leading zero(s) of `downloaded size`

### LINKED ISSUE #10609: `deno upgrade`: both 'downloaded size' and 'download percentage' are padded with leading zeros
(follow-up to #10291)

<sup>*First of all thanks a lot for implementing the download percentage with #10343*</sup>

I just upgraded the new version, 1.10.1,
and I see that both the 'downloaded size' and 'download percentage' are padded with leading zeros:
<details>
<summary>screen capture</summary>

`deno upgrade --dry-run --force`

</details>

For example:
`02.1 MiB / 24.0 MiB (008.9%)`
while if they were:
`2.1 MiB / 24.0 MiB (8.9%)`
they would be more easy to perceive.

Thank you

CGQAQ:
Fixed #10609

### DIFF
--- a/cli/tools/upgrade.rs
+++ b/cli/tools/upgrade.rs
@@ -184,7 +184,7 @@ async fn download_package(
           print!("\u{001b}[1G\u{001b}[2K");
         }
         print!(
-          "{:0>4.1} MiB / {:.1} MiB ({:0>5.1}%)",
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #10647, opened 2021-05-15
docs(cli/dts): fix plugin example

caspervonb:


caspervonb:
Am removing bits of how plugins work (albeit, outdated example) here, so up for discussion.

bartlomieju:
> Am removing bits of how plugins work (albeit, outdated example) here, so up for discussion.
> 
> Alternatives:
> 
> * Declare Deno.core inline (this is what plugins, that use TypeScript must do anyway)
> * Don't check.

Maybe go with `// @ts-ignore` and short comment explaining why it's used?

bartlomieju (approved):

Issue #8226, opened 2020-11-02
Linter shouldn't support eslint ignore directive


bartlomieju:
Currently `deno lint` supports two distinct ways of ignoring diagnostics on next line:
- `eslint-disable-next-line`
- `deno-lint-ignore`

In https://github.com/denoland/deno_lint/issues/476 @nayeemrmn suggests to make `deno-lint-ignore` take into account whole AST node which would render `eslint-disable-next-line` irrelevant. Additionally supporting eslint's directive leads to a subtle bug in `deno_lint` (https://github.com/denoland/deno_lint/issues/287). 

I believe we should remove support for `eslint-disable-next-line` and stay only with `deno-lint-ignore`.

PR #10652, opened 2021-05-16
chore: update deno_lint binary used in CI to v0.5.0

magurotuna:
This PR updates `third_paty` submodule so we can use the latest deno_lint in CI.
The latest deno_lint brings `no-unused-vars` and `no-deprecated-deno-api`, which spot some of the existing code as errors. I dealt with them by fixing the code or adding ignore directives.

Additionally, I removed ESLint directives because we don't use ESLint now and deno_lint hasn't supported them since https://github.com/denoland/deno/issues/8226.


tools/wpt/testharnessreport.js
```diff
@@ -1,13 +1,16 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { writeAllSync } from "https://deno.land/std@0.96.0/io/util.ts";
```
bartlomieju, line 3: Please reference `std` from `test_util/` directory

magurotuna: Done :)

magurotuna:
Hmm, looks like this relative import doesn't work in CI... :thinking: I have no idea how to resolve it.
https://github.com/denoland/deno/blob/6d3ce3763f220f3d720d8440801aad9a227fa14c/tools/wpt/testharnessreport.js#L3

tools/wpt/testharnessreport.js
```diff
@@ -1,13 +1,16 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { writeAllSync } from "../../test_util/std/io/util.ts";
+
```
lucacasonato, line 4: This will fail WPT tests. These files are not really modules, they behave more like scripts. We can't import in here. You'll either need to vendor in the `writeAllSync`, or just use `Deno.writeSync` and assert that the written byte length is correct.

lucacasonato: I have implemented this.

tools/wpt/testharnessreport.js
```diff
@@ -1,13 +1,15 @@
...
-    new TextEncoder().encode(
-      `${JSON.stringify({ name, status, message, stack })}\n`,
-    ),
+  const data = new TextEncoder().encode(
+    `${JSON.stringify({ name, status, message, stack })}\n`,
   );
+  const bytesWritten = Deno.stderr.writeSync(data);
+  if (bytesWritten !== data.byteLength) {
+    throw new TypeError("failed to report test result");
+  }
```
magurotuna, line 10: Shouldn't we make sure that all data have been written to stderr like the following?

```suggestion
  let bytesWritten = 0;
  while (bytesWritten < data.byteLength) {
    bytesWritten += Deno.stderr.writeSync(data.subarray(byteWritten));
  }
```

lucacasonato: that works too :-)

magurotuna: so either is okay? then I'm fine with the current implementation of yours.
I feel a little bit skeptical though, because `Deno.writeSync` calls `op_write_sync` and then `std::io::Write::write` is used under the hood, which requires us to do a loop until all data are written if we wish to imitate something like write_all.
https://doc.rust-lang.org/std/io/trait.Write.html#examples

lucacasonato: I'm aware, I just think the messages are small enough to never fail. But if you want to add the loop, that is fine by me. It is a little more robust.

magurotuna: Makes sense. Indeed the tests run in CI succeeded, but I'd prefer a robust one so I'll add the loop :)

lucacasonato (approved):
LGTM! Thanks :-)

PR #10656, opened 2021-05-16
refactor(core): move ModuleMap to separate RefCell

bartlomieju:
This commit moves bulk of the logic related to module loading 
from "JsRuntime" to "ModuleMap". 

Next steps are to rewrite the actual loading logic (represented by 
"RecursiveModuleLoad") to be a part of "ModuleMap" as well -- 
that way we will be able to track multiple module loads from within 
the map which should help me solve the problem of concurrent 
loads (since all info about currently loading/loaded modules will 
be contained in the ModuleMap, so we'll be able to know if actually 
all required modules have been loaded).


core/modules.rs
```diff
@@ -462,11 +515,85 @@ impl ModuleMap {
...
+
+    Ok(())
+  }
+
+  pub fn get_children(&self, id: ModuleId) -> Option<&Vec<ModuleSpecifier>> {
+    self.info.get(&id).map(|i| &i.import_specifiers)
+  }
+
+  pub fn is_registered(&self, specifier: &ModuleSpecifier) -> bool {
+    self.get_id(&specifier.to_string()).is_some()
```
bnoordhuis, line 596: Untested, but:
```suggestion
    self.get_id(specifier.as_str()).is_some()
```

core/runtime.rs
```diff
@@ -1357,25 +1218,28 @@ impl JsRuntime {
...
-      state.loader.clone()
-    };
+    let module_map_rc = Self::module_map(self.v8_isolate());
+    let op_state = self.op_state();
 
-    let load = RecursiveModuleLoad::main(
-      self.op_state(),
+    let load = module_map_rc.borrow().load_main(
+      op_state.clone(),
       &specifier.to_string(),
```
bnoordhuis, line 1226: Ditto:
```suggestion
      specifier.as_str(),
```

bnoordhuis (approved):

core/modules.rs
```diff
@@ -499,6 +641,63 @@ impl ModuleMap {
...
+      .resolve(self.op_state.clone(), specifier, referrer, false)
+      .expect("Module should have been already resolved");
+
+    if let Some(id) = self.get_id(resolved_specifier.as_str()) {
+      if let Some(handle) = self.get_handle(id) {
+        return Some(v8::Local::new(scope, handle));
+      }
+    }
+
+    None
```
piscisaureus, line 699: Aren't you supposed to throw a JS exception here?

core/runtime.rs
```diff
@@ -495,8 +487,14 @@ impl JsRuntime {
     // TODO(piscisaureus): The rusty_v8 type system should enforce this.
     state.borrow_mut().global_context.take();
 
-    // Drop v8::Global handles before snapshotting
-    std::mem::take(&mut state.borrow_mut().module_map);
+    // Overwrite existing ModuleMap to drop v8::Global handles
+    self
+      .v8_isolate()
+      .set_slot(Rc::new(RefCell::new(ModuleMap::new(
```
piscisaureus, line 493: Is this necessary? 
IIRC `set_slot` is generic and `v8_isolate().set_slot(())` should work.

bartlomieju: Wouldn't that set `()` in a new slot instead of overwriting existing slot with `Rc<RefCell<ModuleMap>>`?

piscisaureus: You're right. Oops.

piscisaureus:
I don't see any problems and it's green, so I can approve it for you.
However I can't see where it's going -- will it solve the problem?

bartlomieju:
> I don't see any problems and it's green, so I can approve it for you.
> However I can't see where it's going -- will it solve the problem?

So in this PR I managed to move bulk of the logic related to module loading from `JsRuntime` to `ModuleMap`. My next steps are to rewrite the actual loading logic (represented by `RecursiveModuleLoad`) to be a part of `ModuleMap` as well -- that way I will be able to track multiple module loads from within the map which should help me solve the problem of concurrent loads (since all info about currently loading/loaded modules will be contained in the `ModuleMap`, so I'll be able to now if actually all required modules have been loaded).

PR #10659, opened 2021-05-17
refactor: share test harness for lsp between bench and integration

kitsonk:
This PR refactors the current integration tests for the LSP to use the same test harness that the benchmarking does, which allows us to have more robust integration tests that allow us to model the two-way request/response path between a lsp client and server, which should make it possible to test for more of the deadlock issues we are encountering.


lucacasonato (approved):
LGTM - painful!

PR #10663, opened 2021-05-17
tests: fix broken tests on main

lucacasonato:


cli/tests/dynamic_import/permissions_blob_local.ts
```diff
@@ -1,4 +1,6 @@
 // This file doesn't really exist, but it doesn't matter, a "PermissionsDenied" error should be thrown.
-const code = `import "file:///local_file.ts";`;
+const code = `import "file:///${
+  Deno.build.os == "windows" ? "C:/" : ""
+}local_file.ts";`;
```
bnoordhuis, line 4: I think you could just do `import "file:///C:/local_file.ts"` because `C:` is a legal path on Unices. Either way is fine, though.

bnoordhuis (approved):

PR #10269, opened 2021-04-20
chore: release crates

bartlomieju:

PR #10674, opened 2021-05-17
tooling: re-enable bench_util

AaronO:
It was disabled in #10269 due to a circular dep, now solved by moving `op_baseline` to `bench_util`.


bartlomieju (commented):
@AaronO I tested this PR locally, we will have to publish `bench_util` crate to crates.io nonetheless, so I suggest to rename the crate to `deno_bench_util`.

bartlomieju (approved):
LGTM, thanks Aaron

Issue #10587, opened 2021-05-11
provide test for LSP deadlock issue


ry:
Fix was landed without a test to make a release deadline - but we still should get a test.

https://github.com/denoland/deno/pull/10555

PR #10679, opened 2021-05-18
chore(lsp): provide test for lsp deadlock issue

kitsonk:
Resolves: #10587

### DIFF
--- a/test_util/src/lsp.rs
+++ b/test_util/src/lsp.rs
@@ -2,6 +2,7 @@
 
 use super::new_deno_dir;
 
+use anyhow::Result;
 use lazy_static::lazy_static;
 use regex::Regex;
 use serde::de;
@@ -62,7 +63,7 @@ impl<'a> From<&'a [u8]> for LspMessage {
   }
 }
 
-fn read_message<R>(reader: &mut R) -> Result<Vec<u8>, anyhow::Error>
+fn read_message<R>(reader: &mut R) -> Result<Vec<u8>>
 where
   R: io::Read + io::BufRead,
 {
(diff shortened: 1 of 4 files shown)


ry (commented):
Were you able to confirm this test fails before the fix (41a3b17de44684365ddad9ba0ad9d9c488d3b065) was applied?

kitsonk:
I went back to the parent of that commit (46f0ac59e0cfa2a90d3e7342f890b5090bdc3f90) and ported over the new test harness plus this test.  Sometimes I could get this to cause the deadlock but sometimes I couldn't, it appears to be very "temperamental" race condition, but more interestingly I was now getting a consistent deadlock on these tests:

- lsp_completions
- lsp_completions_optional
- lsp_signature_help

Which are somewhat related deadlocks, which are now "exposed" due to the re-write of the test harness to be bi-directional.

test_util/src/lsp.rs
```diff
@@ -148,13 +162,42 @@ impl LspClient {
...
+      let mut msg_queue = VecDeque::new();
+      loop {
+        match self.msg_queue.pop_front() {
+          Some(LspMessage::Notification(method, maybe_params)) => {
+            if let Some(p) = maybe_params {
+              let params = serde_json::from_value(p)?;
+              return Ok((method, Some(params)));
+            } else {
+              return Ok((method, None));
+            }
```
ry, line 175: Can this be simplified to  
```rust
return Ok((method, maybe_params.try_into()?));
```

test_util/src/lsp.rs
```diff
@@ -148,13 +162,42 @@ impl LspClient {
...
-        } else {
-          return Ok((method, None));
+      match self.read() {
+        Ok(LspMessage::Notification(method, maybe_params)) => {
+          if let Some(p) = maybe_params {
+            let params = serde_json::from_value(p)?;
+            return Ok((method, Some(params)));
+          } else {
+            return Ok((method, None));
+          }
```
ry, line 194: ditto

test_util/src/lsp.rs
```diff
@@ -166,13 +209,42 @@ impl LspClient {
...
+            }
+          }
+          Some(msg) => {
+            msg_queue.push_back(msg);
+          }
+          _ => break,
+        }
+      }
+      self.msg_queue = msg_queue;
+    }
```
ry, line 231: This bit of code is repeated several times... can that be cleaned up somehow?

test_util/src/lsp.rs
kitsonk: Something non-obvious is going on with that suggestion... I think it is because of the generic type:

```
the trait bound `std::option::Option<R>: From<std::option::Option<Value>>` is not satisfied
required because of the requirements on the impl of `Into<std::option::Option<R>>` for `std::option::Option<Value>`
required because of the requirements on the impl of `TryFrom<std::option::Option<Value>>` for `std::option::Option<R>`
required because of the requirements on the impl of `TryInto<std::option::Option<R>>` for `std::option::Option<Value>`
```

If it where a concrete type, I suspect it would work.  `.into()` doesn't work here either, and even unwrapping the value and using `.try_into()` or `.into()` doesn't work.

I have simplified it by one line (since there is no `.try_map()` that is part of the Rust std library):

```rust
let maybe_params = match maybe_params {
  Some(params) => Some(serde_json::from_value(params)?),
  None => None
};
return Ok((method, maybe_params));
```

test_util/src/lsp.rs
kitsonk: I've tried to clean it up as much as possible, while the blocks look really similar, they are subtly different, but I did my best.

ry (approved):
OK LGTM

Issue #10646, opened 2021-05-15
lsp: failed to load configuration file error is not descriptive


Timmmm:
That error should include some more information otherwise it is quite useless:

1. What files did it try to open?
2. Why did it fail?

satyarohith:
Is this from `deno run --config=...`?  A reproduction and context would be helpful.

Timmmm:
Sorry wrong repo. This was for the VSCode plugin - I'll open it again in the right place!

satyarohith:
@Timmmm I think the issue should stay here as it might be related to `deno lsp`.

Timmmm:
Oh you're right. The error is [here](https://github.com/denoland/deno/blob/24a2e1ef82cf1a081a9b4603a597eb0085755e36/cli/lsp/language_server.rs#L435).

Timmmm:
It looks to me like the error probably does contain at least some more information (not the filename though) but it is printed with [`err.to_string()`](https://github.com/denoland/deno/blob/24a2e1ef82cf1a081a9b4603a597eb0085755e36/cli/lsp/language_server.rs#L738) which won't show all the causes so you only get the outermost `.context()`.

PR #10685, opened 2021-05-18
fix(lsp): make failed to load config error descriptive

satyarohith:
Before:
```
Failed to load configuration file
```
After:
```
Could not find the config file: /Users/sr/p/deno_lsp/tsconfig.json
```

Closes https://github.com/denoland/deno/issues/10646

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -2,7 +2,6 @@
 
 use deno_core::error::anyhow;
 use deno_core::error::AnyError;
-use deno_core::error::Context;
 use deno_core::resolve_url;
 use deno_core::serde::Deserialize;
 use deno_core::serde::Serialize;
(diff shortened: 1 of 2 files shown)


cli/lsp/language_server.rs
```diff
@@ -432,8 +431,7 @@ impl Inner {
         ))
       }?;
 
-      let config_file = ConfigFile::read(config_url.path())
-        .context("Failed to load configuration file")?;
```
satyarohith, line 436: The error from ConfigFile::read is already descriptive. 
```
Updating TypeScript configuration from: "./tsconfig.json"
Updating tsconfig has errored: Could not find the config file: /Users/sr/p/deno_lsp/tsconfig.json
```

ry:
Could you print the error messages before and after? I'm having trouble following what has changed.

satyarohith:
> Could you show us the error messages before and after? I'm having trouble following what has changed.

Previous:
```
Failed to load configurations file
```
Now:
```
Could not find the config file: /Users/sr/p/deno_lsp/tsconfig.json
```

ry:
@satyarohith Would it be difficult to add a test which asserts the filename is part of the error message?

ry (approved):
LGTM - nice test

PR #10689, opened 2021-05-18
fix(http): fix empty blob response

### LINKED ISSUE #10688: http: responding with empty blob causes unreachable error
The following http server stops with an uncaught type error:
```ts
for await (const conn of Deno.listen({ port: 3000 })) {
  (async () => {
    for await (const { respondWith } of Deno.serveHttp(conn)) {
      respondWith(new Response(new Blob([""])));
    }
  })();
}
```

This throws with the following error when it gets a request:

```
error: Uncaught (in promise) TypeError: Unreachable
    at respondWith (deno:runtime/js/40_http.js:126:32)
```

kt3k:
Fixes #10688

This PR makes `new Response(new Blob([]))` work as an empty http response.


ry (approved):
LGTM

PR #10692, opened 2021-05-18
fix(extensions/webstorage): use opstate for sqlite connection

### LINKED ISSUE #10691: bug: `localStorage` methods give leaking resource errors in `Deno.test`
Basically, if you use `localStorage.getItem` or `localStorage.setItem` within a `Deno.test` callback (async or not), you will get errors for leaking resources.

## Example

```ts
Deno.test("Getting Keys", () => {
  localStorage.getItem("hello");
});

Deno.test("Setting Keys", () => {
  localStorage.setItem("hello", "foo");
});
```

```sh
$ deno test --location "http://deno.test"
```

## What Happens

It gives this error:

```
Check file:///C:/Projects/Deno/safe/mod_test.ts
running 2 tests from file:///C:/Projects/Deno/safe/mod_test.ts
test Getting Keys ... FAILED (9ms)
test Setting Keys ... ok (28ms)

failures:

Getting Keys
AssertionError: Test case is leaking resources.
Before: {
  "0": "stdin",
  "1": "stdout",
  "2": "stderr"
}
After: {
  "0": "stdin",
  "1": "stdout",
  "2": "stderr",
  "3": "webStorage"
}

Make sure to close all open resource handles returned from Deno APIs before
finishing test case.
    at assert (deno:runtime/js/06_util.js:33:13)
    at resourceSanitizer (deno:runtime/js/40_testing.js:69:7)
    at async exitSanitizer (deno:runtime/js/40_testing.js:85:9)
    at async runTest (deno:runtime/js/40_testing.js:199:7)
    at async Object.runTests (deno:runtime/js/40_testing.js:244:7)
    at async file:///C:/projects/deno/safe/$deno$test.js:1:1

failures:

        Getting Keys

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (88ms)
```

## Expect behavior

It should not be throwing errors.

## Notes

If you actually look closely at the stdout, then you can see that it says that the "Setting Keys" test passes.  It seems that you will only get the resource leak error for the first `localStorage` call.  You can see this by changing the code to the following:

```ts
Deno.test("Getting Keys", () => { // FAIL
  localStorage.getItem("hello");
});

Deno.test("Setting Keys #1", () => { // PASS
  localStorage.setItem("hello", "foo");
});

Deno.test("Setting Keys #2", () => { // PASS
  localStorage.setItem("hello", "foo3");
});

Deno.test("foo", () => { // FAIL
  throw Error("foo");
});
```

crowlKats:
Closes #10691

### DIFF
--- a/extensions/webstorage/lib.rs
+++ b/extensions/webstorage/lib.rs
@@ -46,32 +46,43 @@ pub fn get_declaration() -> PathBuf {
   PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("lib.deno_webstorage.d.ts")
 }
 
+struct LocalStorage(Connection);
+struct SessionStorage(Connection);
+
 fn get_webstorage(
   state: &mut OpState,
   persistent: bool,
 ) -> Result<&Connection, AnyError> {
-  if state.try_borrow::<Connection>().is_none() {
-    let conn = if persistent {
+  let conn = if persistent {
+    if state.try_borrow::<LocalStorage>().is_none() {
       let path = state.try_borrow::<LocationDataDir>().ok_or_else(|| {
         DomExceptionNotSupportedError::new(
           "LocalStorage is not supported in this context.",
         )
       })?;
       std::fs::create_dir_all(&path.0)?;
-      Connection::open(path.0.join("local_storage"))?
-    } else {
-      Connection::open_in_memory()?
-    };
+      let conn = Connection::open(path.0.join("local_storage"))?;
+      state.put(LocalStorage(conn));
+    }
 
-    conn.execute(
-      "CREATE TABLE IF NOT EXISTS data (key VARCHAR UNIQUE, value VARCHAR)",
-      params![],
-    )?;
 
-    state.put(conn);
-  }
 
-  Ok(state.borrow::<Connection>())
+    &state.borrow::<LocalStorage>().0
+  } else {
+    if state.try_borrow::<SessionStorage>().is_none() {
+      let conn = Connection::open_in_memory()?;
+      state.put(SessionStorage(conn));
+    }
+
+    &state.borrow::<SessionStorage>().0
+  };
+
+  conn.execute(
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM. Please just run fmt again.

lucacasonato (changes requested):
Reverting approval because this apparently breaks sessionStorage

extensions/webstorage/lib.rs
```diff
@@ -51,82 +46,59 @@ pub fn get_declaration() -> PathBuf {
...
-  } else {
-    Connection::open_in_memory()?
-  };
-
-  connection.execute(
-    "CREATE TABLE IF NOT EXISTS data (key VARCHAR UNIQUE, value VARCHAR)",
-    params![],
-  )?;
+) -> Result<&Connection, AnyError> {
+  if state.try_borrow::<Connection>().is_none() {
```
bartlomieju, line 53: I think you'll need two wrapper structs: `struct LocalStorage(Connection)` and `struct SessionStorage(Connection)`; otherwise you will have only a single instance in `OpState` that will depend which API was used first

extensions/webstorage/lib.rs
```diff
@@ -51,82 +46,68 @@ pub fn get_declaration() -> PathBuf {
...
+    }
+
+    &state.borrow::<SessionStorage>().0
   };
 
-  connection.execute(
+  conn.execute(
     "CREATE TABLE IF NOT EXISTS data (key VARCHAR UNIQUE, value VARCHAR)",
     params![],
   )?;
```
lucacasonato, line 82: This is now called in every op. Please move it to the initialization procedures.

crowlKats: Done

lucacasonato (approved):
Thanks 🙂

ry:
I'm confused why this is labeled as "feat". Isn't this a "fix"?

crowlKats:
@ry yeah its a fix, my bad

ry:
Thanks @crowlKats

PR #10698, opened 2021-05-19
feat(bundle): emit source map

satyarohith:
By default, `deno bundle` will inline the source map.
You can disable it by specifying `inlineSourceMap: false` in the `compilerOptions` of tsconfig.
And if you set `sourceMap: true`,  `deno bundle` will generate a `.map` file with the same name as
the output file with the source map.

Note: either `sourceMap` or `inlineSourceMap` can be true.

Closes https://github.com/denoland/deno/issues/8577


satyarohith:
The two tests pass locally but not on the CI.

```
failures:
    integration::bundle
    integration::bundle_jsx
```
The source map generated on the CI is different than local for these two test cases. Not sure what's the cause 🤔

caspervonb:
> The two tests pass locally but not on the CI.
> 
> ```
> failures:
>     integration::bundle
>     integration::bundle_jsx
> ```
> 
> The source map generated on the CI is different than local for these two test cases. Not sure what's the cause 🤔

Make sure you [WILDCARD] any file paths.

satyarohith:
> > The two tests pass locally but not on the CI.
> > ```
> > failures:
> >     integration::bundle
> >     integration::bundle_jsx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The source map generated on the CI is different than local for these two test cases. Not sure what's the cause 🤔
> 
> Make sure you [WILDCARD] any file paths.
> 
> Edit:
> In this case looks like they're embedded In the actual source map?

Yes, this is the cause.

cli/module_graph.rs
```diff
@@ -1186,12 +1194,12 @@ impl Graph {
       cm.build_source_map_from(&mut src_map_buf, None)
         .to_writer(&mut buf)?;
 
-      if emit_options.inline_source_map {
+      if emit_options.source_map {
+        map = Some(String::from_utf8(buf)?);
+      } else if emit_options.inline_source_map {
         src.push_str("//# sourceMappingURL=data:application/json;base64,");
         let encoded_map = base64::encode(buf);
         src.push_str(&encoded_map);
```
caspervonb, line 1202:
```suggestion
        src.push_str(&encoded_map);
        src.push_str("\n");
```

caspervonb:
Suggestion ended up on the wrong line.
Should there be a newlines at the end of these?

satyarohith:
> Suggestion ended up on the wrong line.
> Should there be a newlines at the end of these?

After the embedded source map? Not sure how it is useful. Please let me know if I'm missing something.

---


Ping @kitsonk

bartlomieju:
@kitsonk can we land this PR for 1.12?

cli/config_file.rs
```diff
@@ -73,7 +73,6 @@ pub const IGNORED_COMPILER_OPTIONS: &[&str] = &[
   "esModuleInterop",
   "emitDeclarationOnly",
   "importHelpers",
-  "inlineSourceMap",
```
kitsonk, line 76: Hmmm... there will still should be reported as ignored _unless_ your doing `deno bundle`.  Maybe we need another set, which is only analyzed when your are bundling.

cli/module_graph.rs
```diff
@@ -1209,12 +1217,15 @@ impl Graph {
       cm.build_source_map_from(&mut src_map_buf, None)
         .to_writer(&mut buf)?;
 
-      if emit_options.inline_source_map {
+      if emit_options.source_map {
+        let mut src_map = String::from_utf8(buf)?;
+        src_map.push('\n');
+        map = Some(src_map);
```
kitsonk, line 1223: Shouldn't we be appending something like to the file in this case:

```js
//# sourceMappingURL=bundle.js.map
```

cli/tests/bundle/sourcemap_tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": {
+    "sourceMap": true,
```
kitsonk, line 3: These should be mutually exclusive, is that intentional?

kitsonk (commented):
To be clear about behaviour, the following settings should do this:

- `sourceMap` if `true` writes the source map to a file with the output bundle sources name with a `.map` at the end, and appends the `//# sourceMapURL=` with the with the name of the map file.
- `inlineSources` - if `true` includes the original sources in the source map, this doesn't seem to be part of this feature at the moment, but we should consider it.  I am not sure what the swc source map does by default though, are the sources in there?
- `inlineSourceMap` - if `true` encodes the source map as base64.

I think also the PR has undefined behaviour when the output is sent to the console with Deno bundle, and we should likely error if someone provides a `--config` with `sourceMap: true` but does not provide an output filename.

Also `swc` source maps output `<https://example.com/mod.ts>` instead of `https://example.com/mod.ts`.  This is something that impacts `Deno.emit()` as well, but it makes the source maps a bit of a problem when reconsumed and displayed in a debugger.

There is also `sourceRoot` compiler option which impacts the path to the sources in the source map and the `mapRoot` which impacts the path written in the `sourceMapURL` which this PR doesn't address.

cli/tests/bundle/sourcemap_tsconfig.json
satyarohith: Yes, for the below test:
```rs
 itest!(bundle_errors_on_two_source_map_options {
    args: "bundle jsx_import_from_ts.ts --config bundle/sourcemap_tsconfig.json",
    exit_code: 1,
    output_str: Some("[WILDCARD]\nerror: TS5053 [ERROR]: Option 'sourceMap' cannot be specified with option 'inlineSourceMap'.\n"),
  });
```

PR #10704, opened 2021-05-19
ci: fix 'Skip save cache' step in windows CI

kt3k:
In #10560, `shell` property wasn't set in `Skip save cache (PR)` step. So it's run in powershell in windows and `$GITHUB_ENV` part doesn't work, and therefore `Post Cache build output (PR)` step is not skipped in window. ex. https://github.com/denoland/deno/pull/10703/checks?check_run_id=2618387702

This PR fixes it by specifying `shell: bash`.

`Post Cache build output (PR)` step usually takes 4+ mins. So this fix should reduces typical windows CI time 4 mins.


ry:
Good catch.

Very poor DX from github actions.

piscisaureus (approved):

PR #10713, opened 2021-05-19
fix(dts): `NotSupported` error class not in types #10701

### LINKED ISSUE #10701: `Deno.errors.NotSupported` isn't in types
`Deno.errors.NotSupported` is defined here:
https://github.com/denoland/deno/blob/ef5e5f5e460b40b2c5f7cbcf0a1d11488fc7277f/runtime/js/01_errors.js#L131-L136
However, it isn't included the `.d.ts` files. I'm not sure what it's actually used for, but it seems like it should probably be included in the type definitions if it exists.

sno2:
Closes #10701


kt3k (approved):
@sno2 Thanks! LGTM

PR #10719, opened 2021-05-20
fix(cli/tools/test_runner): --doc should not require permissions

caspervonb:


cli/tools/test_runner.rs
```diff
@@ -338,9 +338,10 @@ pub async fn run_tests(
     let lines_regex = Regex::new(r"(?:\* ?)(?:\# ?)?(.*)")?;
 
     for specifier in &doc_modules {
+      let mut fetch_permissions = Permissions::allow_all();
```
caspervonb, line 341: Way outside of scope but awkward that check_permissions was changed to mutable ref

bartlomieju (approved):

PR #10725, opened 2021-05-20
refactor: Rename inspector structs, remove implementation for Deref

bartlomieju:

### DIFF
--- a/cli/tools/coverage.rs
+++ b/cli/tools/coverage.rs
@@ -13,7 +13,7 @@ use deno_core::error::AnyError;
 use deno_core::serde_json;
 use deno_core::serde_json::json;
 use deno_core::url::Url;
-use deno_runtime::inspector::InspectorSession;
+use deno_runtime::inspector::InMemorySession;
 use deno_runtime::permissions::Permissions;
 use regex::Regex;
 use serde::Deserialize;
@@ -26,11 +26,11 @@ use uuid::Uuid;
 
 pub struct CoverageCollector {
   pub dir: PathBuf,
-  session: Box<InspectorSession>,
+  session: Box<InMemorySession>,
 }
 
 impl CoverageCollector {
-  pub fn new(dir: PathBuf, session: Box<InspectorSession>) -> Self {
+  pub fn new(dir: PathBuf, session: Box<InMemorySession>) -> Self {
     Self { dir, session }
   }
 
--- a/cli/tools/repl.rs
+++ b/cli/tools/repl.rs
@@ -8,7 +8,7 @@ use crate::program_state::ProgramState;
 use deno_core::error::AnyError;
 use deno_core::serde_json::json;
 use deno_core::serde_json::Value;
-use deno_runtime::inspector::InspectorSession;
+use deno_runtime::inspector::InMemorySession;
 use deno_runtime::worker::MainWorker;
 use rustyline::completion::Completer;
 use rustyline::error::ReadlineError;
@@ -257,7 +257,7 @@ impl Highlighter for Helper {
 
 async fn post_message_and_poll(
   worker: &mut MainWorker,
-  session: &mut InspectorSession,
+  session: &mut InMemorySession,
   method: &str,
   params: Option<Value>,
 ) -> Result<Value, AnyError> {
@@ -282,7 +282,7 @@ async fn post_message_and_poll(
 
 async fn read_line_and_poll(
   worker: &mut MainWorker,
-  session: &mut InspectorSession,
+  session: &mut InMemorySession,
   message_rx: &Receiver<(String, Option<Value>)>,
   response_tx: &Sender<Result<Value, AnyError>>,
   editor: Arc<Mutex<Editor<Helper>>>,
@@ -353,7 +353,7 @@ Object.defineProperty(globalThis, "_error", {
 
 async fn inject_prelude(
   worker: &mut MainWorker,
-  session: &mut InspectorSession,
+  session: &mut InMemorySession,
   context_id: u64,
 ) -> Result<(), AnyError> {
   post_message_and_poll(
@@ -372,7 +372,7 @@ async fn inject_prelude(
 
 pub async fn is_closing(
   worker: &mut MainWorker,
-  session: &mut InspectorSession,
+  session: &mut InMemorySession,
   context_id: u64,
 ) -> Result<bool, AnyError> {
   let closed = post_message_and_poll(
--- a/runtime/inspector/mod.rs
+++ b/runtime/inspector/mod.rs
@@ -29,8 +29,6 @@ use std::ffi::c_void;
 use std::mem::replace;
 use std::mem::take;
 use std::mem::MaybeUninit;
-use std::ops::Deref;
-use std::ops::DerefMut;
 use std::pin::Pin;
 use std::ptr;
 use std::ptr::NonNull;
@@ -65,19 +63,6 @@ pub struct DenoInspector {
   _canary_tx: oneshot::Sender<Never>,
   pub server: Option<Arc<InspectorServer>>,
   pub debugger_url: Option<String>,
-}
-
-impl Deref for DenoInspector {
-  type Target = v8::inspector::V8Inspector;
-  fn deref(&self) -> &Self::Target {
-    self.v8_inspector.as_ref().unwrap()
-  }
-}
-
-impl DerefMut for DenoInspector {
-  fn deref_mut(&mut self) -> &mut Self::Target {
-    self.v8_inspector.as_mut().unwrap()
-  }
 }
 
 impl Drop for DenoInspector {
@@ -144,11 +129,10 @@ impl DenoInspector {
     let (canary_tx, canary_rx) = oneshot::channel::<Never>();
 
     // Create DenoInspector instance.
-    let mut self_ = new_box_with(|self_ptr| {
+    let mut self_ = Box::new({
       let v8_inspector_client =
         v8::inspector::V8InspectorClientBase::new::<Self>();
 
-      let sessions = SessionContainer::new(self_ptr, new_websocket_rx);
       let flags = InspectorFlags::new();
       let waker = InspectorWaker::new(scope.thread_safe_handle());
 
@@ -171,7 +155,7 @@ impl DenoInspector {
       Self {
         v8_inspector_client,
         v8_inspector: Default::default(),
-        sessions,
+        sessions: Default::default(),
         flags,
         waker,
         _canary_tx: canary_tx,
@@ -181,17 +165,27 @@ impl DenoInspector {
     });
     self_.v8_inspector =
       v8::inspector::V8Inspector::create(scope, &mut *self_).into();
+    self_.sessions =
+      SessionContainer::new(self_.v8_inspector_mut(), new_websocket_rx);
 
     // Tell the inspector about the global context.
     let context = v8::Local::new(scope, context);
     let context_name = v8::inspector::StringView::from(&b"global context"[..]);
-    self_.context_created(context, Self::CONTEXT_GROUP_ID, context_name);
+    self_.v8_inspector_mut().context_created(
+      context,
+      Self::CONTEXT_GROUP_ID,
+      context_name,
+    );
 
     // Poll the session handler so we will get notified whenever there is
     // new_incoming debugger activity.
     let _ = self_.poll_sessions(None).unwrap();
 
     self_
+  }
+
+  pub fn v8_inspector_mut(&mut self) -> &mut v8::inspector::V8Inspector {
+    self.v8_inspector.as_mut().unwrap()
   }
 
   fn poll_sessions(
@@ -344,11 +338,11 @@ struct SessionContainer {
 
 impl SessionContainer {
   fn new(
-    inspector_ptr: *mut DenoInspector,
+    v8_inspector_ptr: *mut v8::inspector::V8Inspector,
     new_websocket_rx: UnboundedReceiver<WebSocketProxy>,
   ) -> RefCell<Self> {
     let new_incoming = new_websocket_rx
-      .map(move |websocket| WebsocketSession::new(inspector_ptr, websocket))
+      .map(move |websocket| WebsocketSession::new(v8_inspector_ptr, websocket))
       .boxed_local();
     let self_ = Self {
       new_incoming,
@@ -452,12 +446,12 @@ impl WebsocketSession {
   const CONTEXT_GROUP_ID: i32 = 1;
 
   pub fn new(
-    inspector_ptr: *mut DenoInspector,
+    v8_inspector_ptr: *mut v8::inspector::V8Inspector,
     websocket: WebSocketProxy,
   ) -> Box<Self> {
     new_box_with(move |self_ptr| {
       let v8_channel = v8::inspector::ChannelBase::new::<Self>();
-      let mut v8_session = unsafe { &mut *inspector_ptr }.connect(
+      let mut v8_session = unsafe { &mut *v8_inspector_ptr }.connect(
         Self::CONTEXT_GROUP_ID,
         // Todo(piscisaureus): V8Inspector::connect() should require that
         // the 'v8_channel' argument cannot move.
@@ -625,10 +619,10 @@ impl v8::inspector::ChannelImpl for InMemorySession {
 impl InMemorySession {
   const CONTEXT_GROUP_ID: i32 = 1;
 
-  pub fn new(inspector_ptr: *mut DenoInspector) -> Box<Self> {
+  pub fn new(v8_inspector: &mut v8::inspector::V8Inspector) -> Box<Self> {
     new_box_with(move |self_ptr| {
       let v8_channel = v8::inspector::ChannelBase::new::<Self>();
-      let v8_session = unsafe { &mut *inspector_ptr }.connect(
+      let v8_session = { &mut *v8_inspector }.connect(
         Self::CONTEXT_GROUP_ID,
         unsafe { &mut *self_ptr },
         v8::inspector::StringView::empty(),
--- a/runtime/inspector/mod.rs
+++ b/runtime/inspector/mod.rs
@@ -448,19 +448,6 @@ struct WebsocketSession {
   websocket_rx_handler: Pin<Box<dyn Future<Output = ()> + 'static>>,
 }
 
-impl Deref for WebsocketSession {
-  type Target = v8::inspector::V8InspectorSession;
-  fn deref(&self) -> &Self::Target {
-    &self.v8_session
-  }
-}
-
-impl DerefMut for WebsocketSession {
-  fn deref_mut(&mut self) -> &mut Self::Target {
-    &mut self.v8_session
-  }
-}
-
 impl WebsocketSession {
   const CONTEXT_GROUP_ID: i32 = 1;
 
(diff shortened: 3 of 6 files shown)


runtime/inspector/mod.rs
```diff
@@ -681,20 +328,21 @@ impl InspectorFlags {
...
+  new_incoming: Pin<Box<dyn Stream<Item = Box<WebsocketSession>> + 'static>>,
+  handshake: Option<Box<WebsocketSession>>,
+  established: FuturesUnordered<Box<WebsocketSession>>,
 }
 
-impl InspectorSessions {
+impl SessionContainer {
   fn new(
-    inspector_ptr: *mut DenoInspector,
+    v8_inspector_ptr: *mut v8::inspector::V8Inspector,
```
bartlomieju, line 341: This is a bit unfortunate - I would prefer to pass reference to inspector, but since `new_websocket_rx` is a future that's not really possible and I had to keep the pointer here.

runtime/inspector/mod.rs
```diff
@@ -681,20 +328,21 @@ impl InspectorFlags {
...
 }
 
-struct InspectorSessions {
-  new_incoming:
-    Pin<Box<dyn Stream<Item = Box<DenoInspectorSession>> + 'static>>,
-  handshake: Option<Box<DenoInspectorSession>>,
-  established: FuturesUnordered<Box<DenoInspectorSession>>,
+/// A helper structure that helps coordinate sessions during different
+/// parts of their lifecycle.
+struct SessionContainer {
```
piscisaureus, line 333: Agree that `SessionContainer` is a much better name.

piscisaureus (dismissed):
LGTM, some nits.

runtime/inspector/mod.rs
piscisaureus: I see... it is a stream, and each `WebSocketSession` needs a mutable reference to the V8Inspector.

If you wanted to it'd be possible to wrap the `V8InspectorSession` in an `Rc<RefCell<_>>` and make clones of it in the `.map()` call.

bartlomieju: Yeah I was thinking about putting `v8::inspector::V8Inspector` inside a `Rc<RefCell<_>>`. I've hit a wall when trying to decouple `WebsocketSession` from `DenoInspector` so I'll try to apply that change in a follow up PR.

cli/tools/coverage.rs
```diff
@@ -13,7 +13,7 @@ use deno_core::error::AnyError;
 use deno_core::serde_json;
 use deno_core::serde_json::json;
 use deno_core::url::Url;
-use deno_runtime::inspector::InspectorSession;
+use deno_runtime::inspector::InMemorySession;
```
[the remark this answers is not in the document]
bartlomieju: I don't think `InProcessSession` is better - after all, all sessions are in a process :) OTOH `internal` is very vague... Maybe `OneshotChannelSession`? It's a mouthful but should be unambiguous.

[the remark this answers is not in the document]
caspervonb: > OneshotChannelSession

That's an implementation detail tho, doesn't reflect the intent/use case of that session implementation.

runtime/inspector/mod.rs
```diff
@@ -786,46 +402,41 @@ impl task::ArcWake for InspectorWaker {
   }
 }
 
-struct DenoInspectorSession {
+/// An inspector session that proxies messages to Websocket connection.
+struct WebsocketSession {
   v8_channel: v8::inspector::ChannelBase,
-  v8_session: v8::UniqueRef<v8::inspector::V8InspectorSession>,
```
bartlomieju, line 791: So originally this was ` v8::UniqueRef` and I changed it to `v8::UniquePtr` to follow what's used in `DenoInspector` - to be honest I don't know what's the difference in usage of those - we get `UniqueRef` from v8 but we're changing it to `UniquePtr` using `.into()` call. I guess we could just use `UniqueRef` instead? 

Note that it's been wrapped in `Rc<RefCell<>>` so it's clonable and can be passed to other structures require it instead of passing raw pointers.

runtime/inspector/mod.rs
```diff
@@ -405,26 +55,12 @@ enum PollState {
...
   v8_inspector_client: v8::inspector::V8InspectorClientBase,
-  v8_inspector: v8::UniquePtr<v8::inspector::V8Inspector>,
-  sessions: RefCell<InspectorSessions>,
+  // TODO(bartlomieju): make this field private
+  // TODO(bartlomieju): could probably be v8::UniqueRef instead of UniquePtr
+  pub v8_inspector: Rc<RefCell<v8::UniquePtr<v8::inspector::V8Inspector>>>,
+  sessions: RefCell<SessionContainer>,
   flags: RefCell<InspectorFlags>,
   waker: Arc<InspectorWaker>,
-  _canary_tx: oneshot::Sender<Never>,
```
bartlomieju, line 412: As per convo on Discord - not sure what this channel was used for - the only place it was referenced was inside `server.rs` to "deregister" the inspector instance from the server - however, given that `canary_tx` was never used, the reference was effectively dead and thus I removed it. 

I have a feeling it might have been something that was not finished, in which case I will revert this change.

CC @piscisaureus

runtime/inspector/mod.rs
```diff
@@ -681,20 +294,23 @@ impl InspectorFlags {
...
 
-struct InspectorSessions {
-  new_incoming:
-    Pin<Box<dyn Stream<Item = Box<DenoInspectorSession>> + 'static>>,
-  handshake: Option<Box<DenoInspectorSession>>,
-  established: FuturesUnordered<Box<DenoInspectorSession>>,
+/// A helper structure that helps coordinate sessions during different
+/// parts of their lifecycle.
+struct SessionContainer {
+  new_incoming: Pin<Box<dyn Stream<Item = Box<WebsocketSession>> + 'static>>,
```
bartlomieju, line 300: Next step is to rewrite `SessionContainer` to use a more generic "session" object, as per Discord convo `WebsocketSesssion` is not really tied to websockets and could be abstracted away as just a "proxy" that would connect to some other channels (either to use websocket transport or "in-memory")

runtime/inspector/mod.rs
```diff
@@ -405,26 +55,12 @@ enum PollState {
...
-  v8_inspector: v8::UniquePtr<v8::inspector::V8Inspector>,
-  sessions: RefCell<InspectorSessions>,
+  // TODO(bartlomieju): make this field private
+  // TODO(bartlomieju): could probably be v8::UniqueRef instead of UniquePtr
+  pub v8_inspector: Rc<RefCell<v8::UniquePtr<v8::inspector::V8Inspector>>>,
+  sessions: RefCell<SessionContainer>,
   flags: RefCell<InspectorFlags>,
   waker: Arc<InspectorWaker>,
-  _canary_tx: oneshot::Sender<Never>,
-  pub server: Option<Arc<InspectorServer>>,
```
bartlomieju, line 413: Removed reference to `server` completely - I tried `poll_sessions()` without short-circuit if there's no server and it still works. This will help decouple inspector instance from the WS server which will only be available in `runtime` crate.

runtime/inspector/mod.rs
```diff
@@ -481,58 +117,40 @@ impl DenoInspector {
 
   pub fn new(
     js_runtime: &mut deno_core::JsRuntime,
-    server: Option<Arc<InspectorServer>>,
+    new_websocket_rx: mpsc::UnboundedReceiver<WebSocketProxy>,
```
bartlomieju, line 120: Next step is to rewrite this function to not depend on `WebSocketProxy` - turns out inspector can be agnostic of the "transport" layer - all it cares is to get some kind of channel to pump messages.

runtime/web_worker.rs
```diff
@@ -325,10 +327,17 @@ impl WebWorker {
...
+        mpsc::unbounded::<WebSocketProxy>();
+
+      let inspector = DenoInspector::new(&mut js_runtime, new_websocket_rx);
+
+      if let Some(server) = options.maybe_inspector_server.clone() {
+        let info = InspectorInfo::new(server.host, new_websocket_tx);
+        server.register_inspector(info);
+      }
+
+      Some(inspector)
```
bartlomieju, line 340: Hopefully this bit can be refactored a bit more once I decouple inspector from `WebSocketSession` abstraction

runtime/inspector/mod.rs
piscisaureus: The purpose of `_canary_tx` is to let the websocket server thread detect when its counterpart `Inspector` object on the main thread is dropped (this tells the server thread that it should exit). No message is ever sent over this channel, however when the `send` side is dropped the `receive` side wakes up. It is used here: https://github.com/denoland/deno/blob/428bc6849f36ea31f32a42cb8acf64a5902659bf/runtime/inspector.rs#L260-L266

runtime/inspector/mod.rs
piscisaureus: Seems good to me!

runtime/inspector/mod.rs
bartlomieju: Right, I thought that's the case; then my question is: couldn't this be rewritten to explicitly send a message over a channel in `Drop` implementation of the inspector? There was not a single comment regarding this canary channel and it's tx side was prefixed with underscore so it's quite hard to figure that out.

runtime/inspector/mod.rs
```diff
@@ -0,0 +1,722 @@
...
+      w.poll_state = PollState::Woken;
+    });
+  }
+}
+
+/// An inspector session that proxies messages to Websocket connection.
+struct WebsocketSession {
+  v8_channel: v8::inspector::ChannelBase,
+  // TODO(bartlomieju): could probably be v8::UniqueRef instead of UniquePtr
+  v8_session: Rc<RefCell<v8::UniquePtr<v8::inspector::V8InspectorSession>>>,
```
piscisaureus, line 444: Is there really a many-to-one relationship between `WebSocketSession` and `V8InspectorSession`? 
If not, the `Rc` should be unnecessary (and potentially the RefCell too, IDK).

runtime/inspector/mod.rs
```diff
@@ -0,0 +1,722 @@
...
+          },
+        }),
+        _ => panic!("Could not parse inspector message"),
+      },
+    };
+
+    self
+      .response_tx_map
+      .remove(&call_id)
+      .unwrap()
```
piscisaureus, line 631: Note that unwrapping here can cause the deno process to crash if the client sends a response with an invalid call_id - this might be a problem for Deploy.

runtime/inspector/mod.rs
bartlomieju: At the moment it might seem superfluous, but I want to rewrite `WebsocketSession::receive_from_proxy` function to be implemented as a `Future` instead - then dispatching messages to `v8_session` could be handled with a single helper function which is implemented nicely thanks to wrapping in `Rc<RefCell<>>`

runtime/inspector/mod.rs
bartlomieju: I'm aware of this problem - for now this is just as it was implemented originally - there were no changes to `InMemorySession` besides renaming it

runtime/inspector/mod.rs
```diff
@@ -0,0 +1,722 @@
...
+  }
+
+  fn send_notification(
+    &mut self,
+    message: v8::UniquePtr<v8::inspector::StringBuffer>,
+  ) {
+    let raw_message = message.unwrap().string().to_string();
+    let message = serde_json::from_str(&raw_message).unwrap();
+
+    self.notification_queue.push(message);
```
piscisaureus, line 643: (I realize that this code already existed before this PR)

Pushing the message to a `Vec` without somehow waking up the tokio task could potentially cause a hang. Somewhere in the process someone is "waiting" for notifications, they're not going to magically find out when one has arrived, something needs to be done to wake up that task so they get polled by the event loop.

bartlomieju: I'm aware of this problem - I've got a WIP commit that has a "pump" future for `InMemorySession` that does that - unfortunately it currently is racy and deadlocks; needs more investigation.

runtime/inspector/mod.rs
```diff
@@ -1053,6 +717,127 @@ impl InspectorSession {
...
+          let result = response.get("result").unwrap().clone();
+          return Ok(result);
+        }
+      }
+    }
+  }
+
+  async fn receive_from_v8_session(&mut self) {
+    eprintln!("waiting on v8_session response");
+    let (maybe_call_id, message) = self.v8_session_rx.next().await.unwrap();
```
bartlomieju, line 801: This call never resolves and deadlocks...

bartlomieju:
```
sending runtime enable
sending message from repl
sent message from repl
looping
waiting on v8_session response
new handshake session
Debugger session started.
got new message from proxy
sending new message to proxy {"method":"Runtime.executionContextCreated","params":{"context":{"id":1,"origin":"","name":"global context","uniqueId":"-74853213617137066.1750212901488508331"}}}
sent new message to proxy
sending new message to proxy {"id":0,"result":{}}
sent new message to proxy
```

runtime/inspector/mod.rs
piscisaureus: `UniqueRef<T>` cannot be null, so it's like `&mut T`.
`UniquePtr<T>` can be null, so it's like `Option<&mut T>`.

If V8 gives you an `UniqueRef` then you can keep it that way.
However if feel the need to use `Option<UniqueRef<T>>`, then `UniquePtr<T>` is preferrable.

runtime/inspector/mod.rs
caspervonb: There's no where in the loop where it can yield?

bartlomieju: @caspervonb it's all async and inside a loop there's `tokio::select!` I'd expect `receive_from_v8_session` would be polled in that case. It just doesn't ever resolve; I haven't yet tracked what's the exact problem

bartlomieju: Should be fixed now

runtime/worker.rs
```diff
@@ -243,10 +246,33 @@ impl MainWorker {
...
+          result = &mut response => {
+            result.expect("Failed to start session");
+            break;
+          }
+
+          _ = self.run_event_loop() => {
+            // A zero delay is long enough to yield the thread in order to prevent the loop from
+            // running hot for messages that are taking longer to resolve like for example an
+            // evaluation of top level await.
+            tokio::time::sleep(tokio::time::Duration::from_millis(0)).await;
```
piscisaureus, line 269: This looks suspect. Why is it necessary?

runtime/inspector/mod.rs
```diff
@@ -0,0 +1,709 @@
...
+impl Future for InspectorSession {
+  type Output = ();
+  fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+    self.proxy_rx_handler.poll_unpin(cx)
+  }
+}
+
+/// A local inspector session that can be used to send and receive protocol messages directly on
+/// the same thread as an isolate.
+pub struct LocalInspectorSession {
```
piscisaureus, line 592: This name is good.

runtime/worker.rs
caspervonb: 100% cpu usage otherwise.

bartlomieju: Because the "local session" is now run through `JsRuntimeInspector` you need to poll either the inspector or worker (which polls inspector internally). It's a bit unfortunate that I had to add it, but I managed to encapsulate it in a `Worker::with_event_loop` helper; I think this solution is acceptable if it lets us move inspector to `deno_core`.

runtime/inspector/mod.rs
bartlomieju: Brought back as `deregister_tx`/`deregister_rx`

runtime/worker.rs
```diff
@@ -243,10 +250,18 @@ impl MainWorker {
...
-  pub fn create_inspector_session(&mut self) -> Box<InspectorSession> {
-    let inspector = self.inspector.as_mut().unwrap();
-
-    InspectorSession::new(&mut **inspector)
+  pub async fn create_inspector_session(&mut self) -> LocalInspectorSession {
+    let inspector = self.inspector.as_ref().unwrap();
+    let mut session = inspector.create_local_session();
+    // Send message to session to complete the handshake.
+    let msg_fut = session
+      .post_message("Runtime.runIfWaitingForDebugger", None)
```
caspervonb, line 258: I'm surprised this works without specifying the execution context id, REPL had issues using the implicit ID.

caspervonb (approved):

cli/tools/repl.rs
```diff
@@ -399,13 +386,12 @@ pub async fn run(
   program_state: &ProgramState,
   mut worker: MainWorker,
 ) -> Result<(), AnyError> {
-  let mut session = worker.create_inspector_session();
+  let mut session = worker.create_inspector_session().await;
 
   let history_file = program_state.dir.root.join("deno_history.txt");
 
   post_message_and_poll(&mut worker, &mut session, "Runtime.enable", None)
```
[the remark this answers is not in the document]
bartlomieju: Correct, let's do that in a follow up

runtime/worker.rs
bartlomieju: Context id is hardcoded to "1" throughout inspector implementation. We should look into making them parametrized once `deno_core` gets support for multiple contexts

cli/tools/repl.rs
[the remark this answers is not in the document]
bartlomieju: Actually, I'm gonna do that in this PR

cli/tools/repl.rs
```diff
@@ -343,16 +333,19 @@ async fn inject_prelude(
...
-    worker,
-    session,
-    "Runtime.evaluate",
-    Some(json!({
-      "expression": PRELUDE,
-      "contextId": context_id,
-    })),
-  )
-  .await?;
+  worker
```
caspervonb, line 336: Nice! glad to see this gone, been on my low priority list since I wrote it.

runtime/worker.rs
caspervonb: > Context id is hardcoded to "1" throughout inspector implementation.

Yes it is but most messages do still need that id passed to it in the payload.

The inspector has the idea of the implicit "current" context which is used when the optional execution context id is not passed, but we don't implement a way to choose which context is the main context right now.

It's coming back to me a little bit, as far as I can recall global state messages do work without the execution context id but things like evaluate do not.

Resolved by avoidance which is better so much happier with that, we'll get to the missing inspector bits later.

bartlomieju: I guess we could split the hair in two and add context id from the associated constant to all messages in the follow up PR?

caspervonb: > I guess we could split the hair in two and add context id from the associated constant to all messages in the 
Technically its a bug we need to fix, not specifying it *should* work but I went off on a tangent here.

All is good for merge.

bartlomieju: Okay, thanks, let's address this issue in the future then

piscisaureus (approved):
Also LGTM! 🥘

maldito-dot (commented):
@maldito-dot

PR #10734, opened 2021-05-21
fix(tsconfig): empty tsconfig.json file does not cause error

### LINKED ISSUE #10733: Error while parsing empty tsconfig.json
Hi :)

While trying to run deno using: `deno run -c tsconfig.json main.js` I got:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli/config_file.rs:290:59
```

The Error happens with a fully commented tsconfig.json file or an empty one.

#### How typescript and ts-node handle that ?
Running `ts-node -P tsconfig.json main.ts` or `tsc -p tsconfig.json main.ts` with an empty tsconfig.json works 

#### Potential fix
Return a valid empty json object in cli/config_file.rs:290 ?

```diff
-  let jsonc = jsonc_parser::parse_to_serde_value(text)?.unwrap();
+ let jsonc = jsonc_parser::parse_to_serde_value(text)?.unwrap_or(json!({}));
```

I'm ready to submit a PR if you want :)

*deno version*: 1.10.2

romainPrignon:
closes #10733

### DIFF
--- a/cli/config_file.rs
+++ b/cli/config_file.rs
@@ -1,6 +1,7 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 use crate::fs_util::canonicalize_path;
+use deno_core::error::anyhow;
 use deno_core::error::AnyError;
 use deno_core::error::Context;
 use deno_core::serde::Deserialize;
@@ -293,7 +294,10 @@ impl ConfigFile {
   }
 
   pub fn new(text: &str, path: &Path) -> Result<Self, AnyError> {
-    let jsonc = jsonc_parser::parse_to_serde_value(text)?.unwrap();
+    let jsonc = match jsonc_parser::parse_to_serde_value(text) {
+      Ok(Some(value)) if value.is_object() => value,
+      _ => return Err(anyhow!("Unable to parse config file JSON")),
+    };
     let json: ConfigFileJson = serde_json::from_value(jsonc)?;
 
     Ok(Self {
@@ -391,6 +395,13 @@ mod tests {
         maybe_path: Some(config_path),
       }),
     );
+  }
+
+  #[test]
+  fn test_parse_config_with_empty_file() {
+    let config_text = "";
+    let config_path = PathBuf::from("/deno/tsconfig.json");
+    assert!(ConfigFile::new(config_text, &config_path).is_err());
   }
 
   #[test]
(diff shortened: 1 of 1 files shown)


cli/config_file.rs
```diff
@@ -293,7 +293,7 @@ impl ConfigFile {
   }
 
   pub fn new(text: &str, path: &Path) -> Result<Self, AnyError> {
-    let jsonc = jsonc_parser::parse_to_serde_value(text)?.unwrap();
+    let jsonc = jsonc_parser::parse_to_serde_value(text)?.unwrap_or(json!({}));
```
nayeemrmn, line 296: This should error just like the import map would:
```suggestion
    let jsonc = match jsonc_parser::parse_to_serde_value(text) {
      Ok(Some(value)) if value.is_object() => value,
      _ => return Err(anyhow!("Unable to pass config file JSON")),
    };
```

romainPrignon: ok, I didn't know how import map was handled. It makes more sense.

romainPrignon: I'll apply the fix and update the test accordingly

cli/config_file.rs
```diff
@@ -293,7 +294,10 @@ impl ConfigFile {
   }
 
   pub fn new(text: &str, path: &Path) -> Result<Self, AnyError> {
-    let jsonc = jsonc_parser::parse_to_serde_value(text)?.unwrap();
+    let jsonc = match jsonc_parser::parse_to_serde_value(text) {
+      Ok(Some(value)) if value.is_object() => value,
+      _ => return Err(anyhow!("Unable to parse config file JSON")),
```
kitsonk, line 299: Would be better to provide the path as as part of the error message.

kitsonk (commented):
Thanks for the PR, one thought/suggestion.

cli/config_file.rs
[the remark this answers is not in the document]
romainPrignon: And only throw for unparsable files ?

romainPrignon: We might also be consistent with import_map error. I used the same error message.
If we plan to update this one, We may also update import_map

romainPrignon:
Ok, now we emit error if config file is not parsable or not a json object
Empty or commented json file are OK
I updated the error message (see test)

Let me know what you think

cli/config_file.rs
kitsonk: Yeah, we should improve that. But that would be a seperate PR/issue. Opened #10768.

kitsonk:
Looks like we need to run `./tools/format.js` to format the changes.

romainPrignon:
My bad, it should be ok now.
Sorry for the context switching and manually having to run the workflow

kitsonk (approved):
LGTM, thanks for the contribution.

PR #10739, opened 2021-05-21
feat(lsp): show hints from `deno_lint` in addition to messages

magurotuna:
Currently, the LSP returns only messages from `deno_lint`; hints are omitted even when they are.
Some lint rules are very unfriendly in the sense that their diagnostic messages give no information about how the lint errors should be addressed properly without having to suppress them.
One example is `ban-ts-comment` rule, as its message just tells that the directive is disallowed without comment - when seeing this, one may wonder what "comment" is. While the message is very unclear, the hint is so informative that it gives a concrete example of how to address it, like ``Add an in-line comment explaining the reason for using `@ts-expect-error`, like `// @ts-expect-error: <reason>``
Therefore, it would be very nice to show not only messages but also hints from `deno_lint`. This PR appends hints to messages if any.

**before**
without_hint

**after**
with_hint

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -78,13 +78,24 @@ pub struct Reference {
 impl Reference {
   pub fn to_diagnostic(&self) -> lsp::Diagnostic {
     match &self.category {
-      Category::Lint { message, code, .. } => lsp::Diagnostic {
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
LGTM, great improvement.

PR #10741, opened 2021-05-21
feat: Support Deno.beforeEach to run a function before each of the tests

getspooky:
## Description

Runs a function before each of the tests in this file runs. If the function returns a promise, Deno waits for that promise to resolve before running the test.

```ts
Deno.beforeEach(function(): void {
  // 
});
```

```ts
Deno.test("Hello Test", () => {
  assert("Hello");
});
```

## Motivation

Most times you have setup work that has to happen before the tests run. For example consider when you have several tests interact with a database of cities. You have a method initializeCityDatabase() that has to be called before each of these tests
Deno provides helper functions `Deno.beforeEach` to handle this.

```ts
beforeEach(() => {
  initializeCityDatabase();
});

Deno.test("city database has Casablanca", () => {
  assert(isCity("Casablanca"));
});

```

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -168,6 +168,16 @@ declare namespace Deno {
    * */
   export function test(name: string, fn: () => void | Promise<void>): void;
 
+  /** Runs a function before each of the tests in this file runs.
+   * `fn` can be async if required.
+   *
+   * ```ts
+   * Deno.beforeEach(function(): void {
+   * });
+   * ```
+   */
+  export function beforeEach(fn: () => void | Promise<void>): void;
+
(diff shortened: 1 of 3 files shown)


caspervonb:
Very strongly against sticking this in the test runner.
It is preferable to use composition instead.

```js
Deno.test("myTestWithdatabase", withDatabase(async function(db) {
  // ...
});
```

ry:
Thanks @getspooky - but I'm in agreement with Casper.

Maybe we need some better documentation or even a helper module in std/testing/ to communicate this pattern more effectively.

Closing without merge.

ry:
I've opened an issue in std for adding a utility library for this functionality
 https://github.com/denoland/deno_std/issues/931

Issue #9501, opened 2021-02-14
Workers with base64 URL can have a very large stacktrace


andreespirela:
Note: While this is the expected behavior, what this ticket will only be a suggestion.

I have a worker which is created with a base64 of more than 3000 characters. If the worker throws an error, the parent will receive a huge stacktrace:
![image](https://user-images.githubusercontent.com/60560109/107869334-0f238f00-6e5b-11eb-9beb-c28456c89fdf.png)
which can take seconds to be written in stdout. 

My suggestion is, whenever a worker is created with a base64 URL, then it should have a fake path when generating an stacktrace, like `file://$deno.ts` for example.

kitsonk:
How would you suggest differentiating from one base64 URL to another?

kitsonk:
The best solution would be to use ellipses when printing out, something like `data:application/javascript;base64,ZXhwb3...KfQo=`.

andreespirela:
> The best solution would be to use ellipses when printing out, something like `data:application/javascript;base64,ZXhwb3...KfQo=`.

I think this is the best as well, first 5 characters and last 5 characters.

lucacasonato:
We truncate data URLs in stack traces now.

PR #10746, opened 2021-05-22
feat(cli/fmt_errors): Truncate error location when source is "data:application/javascript..." Data URL

mikehw:
Addresses #9501. 

When the source of an Error is within a [Data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs), truncate the "file name" in the formatted error location to be something like `data:application/javascript;base64,Y29uc...pOw==`.

Adds tests to cover cli and `core/error.js` formatting.


cli/tests/error_028_data_error_format_js.ts
```diff
@@ -0,0 +1,5 @@
+(async () => {
+  await import(
+    "data:application/javascript;base64,Y29uc29sZS5lcnJvcihFcnJvcigiYmFkIikpOw=="
```
nayeemrmn, line 3: I think we should be a lot more generous as to how many characters are allowed before truncation happens. How about >120?

cli/fmt_errors.rs
```diff
@@ -22,7 +22,19 @@ pub fn format_location(frame: &JsStackFrame) -> String {
...
+        let data_length = data_pieces.1.len();
+        if let Some(data_start) = data_pieces.1.get(0..5) {
+          if let Some(data_end) = data_pieces.1.get(data_length - 5..) {
+            fmt_file_name =
+              format!("{},{}...{}", data_pieces.0, data_start, data_end)
+          }
+        }
+      }
+    }
+    result += &cyan(fmt_file_name).to_string();
```
lucacasonato, line 37: This is very brittle - please use the data URL parser for this. Not all data URLs are contain base64 data. They can also contain plain text.

cli/tests/error_028_data_error_format_js.ts
mikehw: Added a threshold of 150 to match `SOURCE_ABBREV_THRESHOLD`

cli/fmt_errors.rs
mikehw: Updated Rust and Javascript versions to use URL parsers. Modified the tests to demonstrate that it works the same regardless of data encoding.

bartlomieju:
@mikehw sorry for the delay, I'll try to land this PR this week.

cli/fmt_errors.rs
bartlomieju: I agree with Luca that this bit should use data URL parser. We already have dependency on `data_url` crate, here's an example how it's used in other parts of the codebase: https://github.com/denoland/deno/blob/e02403bb4590da6823eb956bcbc89e9a937d431d/cli/file_fetcher.rs#L330-L336

bartlomieju:
I'm not very happy that this PR requires changes to CLI/runtime and core. @nayeemrmn do you have any ideas how we could handle it more gracefully?

nayeemrmn:
@bartlomieju I reimplemented this in #12127 as a cli-core hook same as source mapping.

bartlomieju:
Superseded by https://github.com/denoland/deno/pull/12127

Issue #10628, opened 2021-05-13
Separate `deno graph` from `deno info`


nayeemrmn:
The `deno info` subcommand should only output cache info and for a single module at a time. We should have a separate `deno graph` subcommand for outputting the dependency graph of a module.

1. The use cases for these two things are entirely separate. A large list of dependencies creates a lot of noise to sift through when you are just trying to get the location of a cache file. You never need both of these things at once.
2. It allows more versatility for `deno info`. I propose that `deno info` doesn't implicitly invoke `deno cache` (which it would no longer need to), and should show null cache file paths for uncached modules. You can instead call `deno cache <url> && deno info <url>` for the current behaviour. This design can be applied to query the state of the cache, which partially satisfies feature requests like #5600.

~~This would be a 2.0 change.~~

nayeemrmn:
> This would be a 2.0 change.

It has occurred to me that the output of `deno info mod.ts` is merely pretty-print... and the API access `deno info --json mod.ts` is still marked as unstable... it's a strong argument for permitting this change in a minor release with some provisions.

Other than the proposed change the `--cert`, `--import-map`, `--reload` flags would be noop'd and hidden until 2.0 removal.

What do you think? @ry @kitsonk

bartlomieju:
We should do it as `deno info --dont-recurse`. @nayeemrmn do you want to take this up?

bartlomieju:
There's little need for this so we're not gonna proceed.

PR #10758, opened 2021-05-25
feat(cli): provide Deno.info() API

### LINKED ISSUE #2096: Provide JS API to dependency information (`Deno.info()`)
Currently "deno info https://deno.land/std/http/file_server.ts" will display a dependency graph of the specified module. It would be nice to be able to access this information in a programmatic way.
```
let depTree = await Deno.deps("https://deno.land/std/http/file_server.ts")
```
Where depTree would be some JSON encodable object similar to the current output.

The first step to exposing this would be pure rust functionality of turning a `Deps` structure into a JSON string: https://github.com/denoland/deno/blob/960ee5257a7b38c6b3b59a3dca670c7c1771996f/cli/modules.rs#L188

kitsonk:
Resolves #2096 

This is a work in progress PR.  There are a couple of items still to address:

- [x] There should be a permission check for read permission to the DENO_DIR to access the API.
- [x] The API should support import-maps, like `Deno.emit()`, but no one is happy with the `Deno.emit()` solution, so we need to debate the best way to allow an import-map to be provided.
- [x] It needs tests.

I have implemented the import map in a more straight forward way and my opinion is that we should change `Deno.emit()` to follow.  In addition I added two options that are off by default that suppress some sensitive data from being returned and documented the whole security implications of using the API.  Flexing all the features looks something like this:

```ts
const info = await Deno.info("./test.ts", {
  checksums: true,
  importMap: "./import_map.json",
  paths: true,
});

console.log(info);
```

Or with an import object literal, something like this:

```ts
const info = await Deno.info("./test.ts", {
  importMap: {
    imports: {
      jquery: "https://cdn.skypack.dev/jquery",
    },
  },
});

console.log(info);
```


crowlKats:
Shouldn't it also support usage without a specifier, just like the `info` subcommand?

kitsonk:
> Shouldn't it also support usage without a specifier, just like the `info` subcommand?

I'm not personally convinced of that. What would be a legitimate use case for having that type of information available programatically?

crowlKats:
> I'm not personally convinced of that. What would be a legitimate use case for having that type of information available programatically?

One use case is together with #10589, so one could get access to the location of the localstorage file (and other apis that would store data based on the origin)

nayeemrmn:
What's your opinion on #10628, and maybe representing that change in advance with the runtime API? It would solve needing read permission to `DENO_DIR` when you just want a module graph and no cache info, so this is another justification IMO.

kitsonk:
> so one could get access to the location of the localstorage file (and other apis that would store data based on the origin)

Knowing a location isn't a use case... 😄  What problem does that solve having that information available via an API.

> What's your opinion on #10628, and maybe representing that change in advance with the runtime API?

Maybe. I get the reasoning for it, but it complicates the API surface and as noted in the issue, it isn't something to be considered until 2.0 and I would then expect the Deno APIs to reflect the subcommands.  My opinion is we shouldn't wait until 2.0 to introduce this API (it already has quite a few real world use cases, especially with those using Deno as an on the fly transpiler/bundler).

crowlKats:
> Knowing a location isn't a use case... 😄 What problem does that solve having that information available via an API.

Well one example would be that one could make a backup by copying the file (useful for indexeddb)

SyrupThinker:
> > Shouldn't it also support usage without a specifier, just like the `info` subcommand?
> 
> I'm not personally convinced of that. What would be a legitimate use case for having that type of information available programatically?

I have an example use-case with Docuraptor.
It uses `deno info` to [get the module cache directory](https://github.com/SyrupThinker/docuraptor/blob/master/docuraptor.ts#L264-L284) to then [aggregate a list of all locally cached modules](https://github.com/SyrupThinker/docuraptor/blob/master/docuraptor.ts#L93-L122).
This is used to provide a documentation index for said modules.

nayeemrmn:
I think any info provided under `deno info [mod.ts]` has a use case in meta kind of tooling.

However I wouldn't be surprised if the runtime equivalents for `deno info` and `deno info mod.ts` were different functions. There is zero intersection in both arguments and return types. Other than the sheer justification that they are the same on the CLI, or that return type overloading is allowed by TS, I'm not a fan of keeping disjoint functions under the same name. We should consider having both `Deno.info()` and `Deno.moduleInfo()`.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -540,6 +540,177 @@ declare namespace Deno {
...
+     * dependency and could have security implications if exposed. Therefore,
+     * this is `false` by default. */
+    checksums?: boolean;
+    /** A string, URL pointing to a import map or an object providing the import
+     * map to be used when resolving modules.
+     *
+     * In order to resolve relative modules, import maps need a "base" which is
+     * the import map itself. When an object is passed as a value the current
+     * working directory for the Deno process will be used. */
+    importMap?: string | URL | ImportMap;
```
bartlomieju, line 636: In case this is `ImportMap` type - what is the "base URL" used for that import map?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -540,6 +540,177 @@ declare namespace Deno {
...
+    paths?: boolean;
+  }
+
+  /**
+   * **UNSTABLE**: new API, yet to be finalized.
+   *
+   * Similar to the command line functionality `deno info --json`, the API
+   * provides information about a module and its dependencies.
+   *
+   * Requires `allow/read` permissions to access local modules and cached
```
bartlomieju, line 650: Nitpick: change `allow/read` to `--allow-read` to keep consistent with other docs?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -540,6 +540,177 @@ declare namespace Deno {
...
+   * });
+   * console.log(info);
+   * ```
+   *
+   * Include module checksums and local file system paths:
+   *
+   * ```ts
+   * const info = await Deno.info("./example.ts", {
+   *   checksums: true,
+   *   paths: true,
```
bartlomieju, line 699: Why not provide these things by default?

cli/ops/runtime_compiler.rs
```diff
@@ -134,3 +139,107 @@ async fn op_emit(
...
+  String(String),
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct InfoArgsOptions {
+  #[serde(default)]
+  checksums: bool,
+  import_map: Option<StringOrImportMap>,
+  #[serde(default)]
```
bartlomieju, line 162: Consider adding `Default` to derive instead?

cli/ops/runtime_compiler.rs
```diff
@@ -134,3 +139,107 @@ async fn op_emit(
...
+  let program_state = state.borrow().borrow::<Arc<ProgramState>>().clone();
+  let mut runtime_permissions = {
+    let state = state.borrow();
+    state.borrow::<Permissions>().clone()
+  };
+
+  // we should check read permissions on the cache first, to ensure that we can
+  // expose any cache information to the runtime
+  let custom_root = std::env::var("DENO_DIR").map(String::into).ok();
+  let deno_dir = DenoDir::new(custom_root)?;
```
bartlomieju, line 189: Skip reading env variable and use `dir` field from `ProgramState` instead?

cli/ops/runtime_compiler.rs
```diff
@@ -134,3 +139,107 @@ async fn op_emit(
...
+  let handler = Arc::new(Mutex::new(FetchHandler::new(
+    &program_state,
+    runtime_permissions.clone(),
+    runtime_permissions.clone(),
+  )?));
+  let maybe_import_map = if let Some(options) = &args.options {
+    match &options.import_map {
+      Some(StringOrImportMap::ImportMap(value)) => {
+        let cwd = std::env::current_dir()?;
+        let base_url = resolve_path(cwd.to_string_lossy().as_ref())?;
```
bartlomieju, line 202: To answer myself to previous question - in case of passing "object" import map it uses CWD as default URL, could you please add it to the JSDoc on the API?

bartlomieju (commented):
Sorry for slow review @kitsonk, I completely missed the request 🤦 

I have only minor comments

timreichen:
Just as an input: instead of a `isDynamic` in `ModuleGraphDependency`, maybe a `dependencyType` would be better.
```ts
type DependencyType = "Import" | "Export" | "DynamicImport"
```
I ran into problems having just a dynamic flag when creating bundler. If one wants to add "Fetch" or some other type in the future, DependencyType can easily be extended.

kitsonk:
Closing due to https://deno.land/x/deno_graph/ being available.

PR #10759, opened 2021-05-25
feat(runtime/js/testing)Add parameters option to Deno.test

ryo-ma:
I think that Deno.test should add a parameters test that executes a single test method multiple times with different parameters.

I tried to support it as follows.
I have supported it by adding a parameters argument as an option. Parameters can be received as arguments of fn in the list order of the parameters.

``` ts
import { assertEquals } from "../../../test_util/std/testing/asserts.ts";

Deno.test({
  name: "object parameters test",
  parameters: [{ data: 1, except: 2 }, { data: 2, except: 4 }],
  fn(parameter: { data: number; except: number }) {
    assertEquals(parameter.data * 2, parameter.except);
  },
});

Deno.test({
  name: "single string parameters test",
  parameters: ["hello", "hello"],
  fn(parameter: string) {
    assertEquals(parameter, "hello");
  },
});

```

Result 

Outputs which parameter was executed.

```
test object parameters test {"data":1,"except":2} ... ok (8ms)
test object parameters test {"data":2,"except":4} ... ok (2ms)
test single string parameters test "hello" ... ok (3ms)
test single string parameters test "hello" ... ok (2ms)

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (112ms)
```

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -95,8 +95,8 @@ declare namespace Deno {
    * See: https://no-color.org/ */
   export const noColor: boolean;
 
-  export interface TestDefinition {
-    fn: () => void | Promise<void>;
+  export interface TestDefinition<T = void> {
+    fn: (() => void) | ((parameter: T) => void) | Promise<void>;
     name: string;
     ignore?: boolean;
     /** If at least one test has `only` set to true, only run tests that have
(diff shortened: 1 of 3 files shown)


caspervonb:
Seems redundant when you can just do a loop.

```ts
const variants = [
  { actual: 1, expected: 2 },
  { actual: 2, expected: 4 },
];

for (const options of variants) {
  Deno.test(`it works with ${options}`, function() {
    assertEqual(variant.actual, variant.expected);
  });
}
```

bartlomieju:
Thank you for the PR @ryo-ma but I agree with Casper that using loop is preferable solution in this case. 

We're very careful about extending `Deno.TestDefinition` with additional fields and adding parameterization could be handled by 3rd party library. 

I'm gonna close without a merge.

PR #10760, opened 2021-05-25
fix(extension/file): update File constructor following the spec

kt3k:
This PR is a retry of #10459. (This time it doesn't depends on the harness issue)

This PR updates the File constructor, following this spec update https://github.com/w3c/FileAPI/pull/171. This PR removes the file name replacement, and enables the corresponding wpt by polyfilling the global `document.body` object. This PR also updates the wpt expectations which start passing because of the document.body polyfill.


lucacasonato (approved):

lucacasonato:
The introduction of the document.body polyfill has broken significant amounts of the internals of testharness.js (mainly through this function: https://github.com/web-platform-tests/wpt/blob/master/resources/testharness.js#L471). The tests that this PR looks to make pass, are actually failing, the polyfill is just hiding it. I am going to revert the polyfill.

PR #10763, opened 2021-05-25
refactor: move JsRuntimeInspector to deno_core

bartlomieju:
This commit moves implementation of "JsRuntimeInspector" to "deno_core" crate.

To achieve that following changes were made:

* "Worker" and "WebWorker" no longer own instance of "JsRuntimeInspector",
instead it is now owned by "deno_core::JsRuntime".

* Consequently polling of inspector is no longer done in "Worker"/"WebWorker",
instead it's done in "deno_core::JsRuntime::poll_event_loop".

* "deno_core::JsRuntime::poll_event_loop" and "deno_core::JsRuntime::run_event_loop",
now accept "wait_for_inspector" boolean that tells if event loop should still be 
"pending" if there are active inspector sessions - this change fixes the problem 
that inspector disconnects from the frontend and process exits once the code has
stopped executing.


ry:
I defer to @piscisaureus for review.

piscisaureus (approved):
LGTM as discussed

core/runtime.rs
```diff
@@ -577,6 +600,9 @@ impl JsRuntime {
     &mut self,
     cx: &mut Context,
   ) -> Poll<Result<(), AnyError>> {
+    // We always poll the inspector if it exists.
+    let _ = self.inspector().map(|i| i.poll_unpin(cx));
```
bartlomieju, line 604: This result should not be ignored, instead it should alter result of `poll_event_loop`

core/lib.rs
```diff
@@ -37,6 +38,9 @@ pub use crate::async_cell::AsyncRefFuture;
 pub use crate::async_cell::RcLike;
 pub use crate::async_cell::RcRef;
 pub use crate::flags::v8_set_flags;
+pub use crate::inspector::JsRuntimeInspector;
+pub use crate::inspector::LocalInspectorSession;
+pub use crate::inspector::SessionProxy;
```
bartlomieju, line 43: Rename to `InspectorSessionProxy`

PR #10786, opened 2021-05-29
core: don't include_str extension js code

lucacasonato:
This speeds up incremental rebuild when only touching JS files by 13-15%

Rebuild time after `touch 01_broadcast_channel.js`:

main: run 1 49.18s, run 2 50.34s
this: run 1 43.12s, run 2 43.19s

The reason for this speedup is that `deno_runtime(build.rs)` does not
need to be rebuilt on every js file change now.

Closes #10786

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -1,7 +1,8 @@
 use crate::error::AnyError;
 use crate::{OpFn, OpState};
 
-pub type SourcePair = (&'static str, &'static str);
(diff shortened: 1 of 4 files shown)


lucacasonato:
Binary size reduction is around 0.9%.

latest canary: 81315944 bytes
this branch: 80632656 bytes

ry (approved):
Nice work Luca! LGTM

PR #10787, opened 2021-05-29
build: cli(build) shouldn't depend on deno_runtime

lucacasonato:
This speeds up incremental rebuild when only touching JS files by 30%
compared to #10786.

Rebuild time after touch 01_broadcast_channel.js:

main: run 1 49.18s, run 2 50.34s
\#10786: run 1 43.12s, run 2 43.19s
this + #10786: run 1 30.30s, run 2 30.95s

The reason for the speedup is that deno(build.rs) does not get
rebuilt after changes to deno_runtime.


ry (approved):
LGTM 👍

PR #10796, opened 2021-05-30
feat(extensions/crypto): implement subtle.digest

caspervonb:
Superceedes https://github.com/denoland/deno/pull/9069

Co-authored-by: Yacine Hmito <yacinehmito@users.noreply.github.com>


extensions/crypto/01_crypto.js
```diff
@@ -41,10 +41,59 @@
...
+        algorithm = { name: algorithm };
+      } else if (typeof algorithm === "object" && algorithm !== null) {
+        if (typeof algorithm.name !== "string") {
+          throw new TypeError("Algorithm name is missing or not a string");
+        }
+
+        algorithm = { ...algorithm };
+      } else {
+        throw new TypeError("Argument 1 must be an object or a string");
+      }
```
lucacasonato, line 56: Use WebIDL converters

lucacasonato (changes requested):

extensions/crypto/00_webidl.js
```diff
@@ -0,0 +1,25 @@
...
+"use strict";
+
+((window) => {
+  const webidl = window.__bootstrap.webidl;
+  webidl.converters["AlgorithmIdentifier"] = (V, _opts) => {
+    if (typeof V == "string") {
+      return webidl.converters["DOMString"](V);
+    }
+
+    return webidl.converters["Algorithm"](V);
```
lucacasonato, line 11: Pass opts to the converters to maintain context, add the comment for which IDL union this implements, and please follow the spec algorithm (current impl is wrong): https://heycam.github.io/webidl/#es-union. Relevant here are steps 10.3 and 14.

extensions/crypto/01_crypto.js
```diff
@@ -41,10 +42,83 @@
...
+      return normalizeAlgorithm({ name: algorithm }, op);
+    }
+
+    const initialAlgorithm = webidl.converters["Algorithm"](algorithm, {
+      context: "Argument 1",
+    });
+
+    const registeredAlgorithms = supportedAlgorithms[op];
+    const algorithmName = Object.keys(registeredAlgorithms)
+      .find((key) => key.toLowerCase() == initialAlgorithm.name.toLowerCase());
```
bnoordhuis, line 65: I'm not thrilled by how roundabout and complicated this approach is. Could simply be:
```js
// Should match op_crypto_subtle_digest() in extensions/crypto/lib.rs 
function digestToId(name) {
  switch (name) {
    case "SHA-1": return 0;
    case "SHA-256": return 1;
    case "SHA-384": return 2;
    case "SHA-512": return 3;
  }
  throw new DOMException(/* ... */);
}

name = name.toUpperCase();
const id = digestToId(name);
const normalizedAlgorithm = { name };
// etc.
```

extensions/crypto/lib.rs
```diff
@@ -51,6 +59,28 @@ pub fn op_crypto_get_random_values(
...
+  let algorithm = match algorithm_id {
+    0 => &digest::SHA1_FOR_LEGACY_USE_ONLY,
+    1 => &digest::SHA256,
+    2 => &digest::SHA384,
+    3 => &digest::SHA512,
+    _ => panic!("Invalid algorithm id"),
+  };
+
+  let input = data.ok_or_else(null_opbuf)?;
+  let output = tokio::task::spawn(async move {
```
bnoordhuis, line 76:
```suggestion
  let output = tokio::task::spawn_blocking(move || {
```
Plain spawn() with the single-threaded runtime means it's still going to block the event loop.

bnoordhuis (commented):
Glad to see this is being picked up again, thanks Casper!

extensions/crypto/Cargo.toml
```diff
@@ -15,4 +15,6 @@ path = "lib.rs"
 
 [dependencies]
 deno_core = { version = "0.88.1", path = "../../core" }
+tokio = { version = "1.4.0", features = ["full"] }
```
bnoordhuis, line 18:
```suggestion
tokio = { version = "1.6.1", features = ["full"] }
```
(That's what the other crates currently use.)

extensions/crypto/01_crypto.js
caspervonb: Initial pass was like that, now the whole normalisation thing follows the spec very literally

extensions/crypto/01_crypto.js
```diff
@@ -41,10 +42,80 @@
...
+      algorithm = webidl.converters.AlgorithmIdentifier(algorithm, {
+        context: "Argument 1",
+      });
+
+      data = webidl.converters.BufferSource(data, {
+        context: "Argument 2",
+      });
+
+      algorithm = normalizeAlgorithm(algorithm, "digest");
+      data = data.slice(0);
```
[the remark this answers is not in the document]
caspervonb: sgtm

extensions/crypto/01_crypto.js
```diff
@@ -41,10 +42,91 @@
...
+      case "SHA-256":
+        return 1;
+      case "SHA-384":
+        return 2;
+      case "SHA-512":
+        return 3;
+    }
+  }
+
+  const subtle = {
```
lucacasonato, line 98: `SubtleCrypto` is a class.

extensions/crypto/lib.rs
caspervonb: Oops, mindless coping from the original.

extensions/crypto/01_crypto.js
```diff
@@ -41,10 +42,102 @@
...
+
+      algorithm = webidl.converters.AlgorithmIdentifier(algorithm, {
+        context: "Argument 1",
+      });
+
+      data = webidl.converters.BufferSource(data, {
+        context: "Argument 2",
+      });
+
+      data = data.slice(0);
```
lucacasonato, line 115: Data returned from BufferSource can either be a buffer view, or a buffer. You need to normalize it all into a uint8array (taking into account byteOffset and byteLength). If there are no WPT tests for this, I would be happy to write some.

caspervonb: > You need to normalize it all into a uint8array
 
Was sure I confirmed that serde_v8 would convert this, but on a second test, indeed it only takes TypedArrays.

extensions/crypto/01_crypto.js
```diff
@@ -112,7 +112,13 @@
         context: "Argument 2",
       });
 
-      data = data.slice(0);
+      if (ArrayBuffer.isView(data)) {
+        data = new Uint8Array(data.buffer, data.byteOffset, data.byteLength);
+      } else {
+        data = new Uint8Array(data, data.byteOffset, data.byteLength);
```
lucacasonato:
```
> new ArrayBuffer().byteOffset
undefined
```

I think the second and third arg to the constructor can go.

lucacasonato: Oh, this is why you don't review commits in isolation (unless it makes sense)

caspervonb: > Oh, this is why you don't review commits in isolation (unless it makes sense) 🤦

Hehe apologies, I pushed immediately rather than taking a minute to think 😁

lucacasonato (approved):
LGTM! Thanks for seeing this through :-)

Issue #10788, opened 2021-05-29
`deno test` never finishes when test case calls Deno.run and `--coverage` options is specified


kt3k:
The following test script never finishes when `--coverage=./cov` option is specified

test.js
```ts
Deno.test("foo", async () => {
  const p = Deno.run({ cmd: ["echo", "foo"] });
  await p.status();
  p.close();
});
```

With the above `test.js`, `deno test -A test.js` finishes normally, but `deno test -A test.js --coverage=./cov` never finishes executing.

A similar error happens in deno_std CI and causes timeout, example: https://github.com/denoland/deno_std/pull/938/checks?check_run_id=2701040693

The behavior varies depending on the command passed to Deno.run (for, example `ls` doesn't cause the above problem on my machine (mac))

bartlomieju:
My first thought was that it might be caused by https://github.com/denoland/deno/pull/10763 but if the behavior varies depending on the subcommand used I'm a bit lost...

caspervonb:
Just tried on my Mac with Canary, I can't reproduce.

kt3k:
I checked the above behavior with mac 10.15.7 (intel cpu).

@caspervonb Can you pass `http/file_server_test.ts` or `examples/curl_test.ts` of this std branch https://github.com/kitsonk/deno_std/tree/ts_43 with canary? These are actuall test scripts which are causing CI timeout in std

kitsonk:
I wonder if it is because of the emit of the private methods and instrumenting those, that was the "biggest" change.  I will try on oak as well, to see if that breaks coverage too.

caspervonb:
> I wonder if it is because of the emit of the private methods and instrumenting those, that was the "biggest" change. I will try on oak as well, to see if that breaks coverage too.

Most likely this is from the inspector refactor, the thread is likely being put to sleep and never awoken.

kitsonk:
I don't have any problems with it in oak on canary: https://github.com/oakserver/oak/runs/2704076265 runs with `--coverage` fine.

bartlomieju:
I did a bisect and indeed https://github.com/denoland/deno/pull/10725 causes the hang in `file_server_test.ts`. Need to investigate what's the exact problem.

PR #10802, opened 2021-05-31
fix: running subcommand with coverage collection

bartlomieju:
Fixes https://github.com/denoland/deno/issues/10788

Merge on approval


lucacasonato (approved):
LGTM, but I don't understand the change. Would be great to have a test.

PR #10804, opened 2021-05-31
fix(cli/module_graph): Represent bare imports as module graph error slots

### LINKED ISSUE #10795: `deno cache` does not fail on bare imports
```ts
// deno cache
import "foo";
```
Completes successfully.

nayeemrmn:
Currently bare imports are omitted from the module graph and the error is encountered either at V8 module loading or swc bundling. This represents them in the module graph as:
```rust
(
  Url("bare:foo"),
  ModuleSlot::Err("Relative import path \"foo\" not prefixed with / or ./ or ../ from <referrer>")
)
```
Fixes #10795.


kitsonk (approved):
LGTM

PR #10805, opened 2021-05-31
feat(web): support the stream option to TextDecoder#decode

### LINKED ISSUE #5351: TextDecoder should support the stream option
https://github.com/denoland/deno/blob/bced52505f32d6cca4f944bb610a8a26767908a8/cli/js/web/text_encoding.ts#L456

Ideally TextDecoder would support the stream option for spec compliance as well as for supporting incremental decoding of encoded text from streaming input that may include incomplete surrogate pairs.

andreubotella:
Fixes #5351.

### DIFF
--- a/extensions/web/08_text_encoding.js
+++ b/extensions/web/08_text_encoding.js
@@ -222,12 +222,18 @@
     return result;
   }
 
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
Awesome! Implementation looks good. Thanks so much :-)

PR #10811, opened 2021-06-01
fix(core/extensions): Simplify macro rules for include_js_files.

linbingquan:
This is my second time to contribution, if this `PR` has some problem, please tell me.


kitsonk:
The trailing comma is preferred for the macro and is very intentional.  It is consistent with the rest of the default formatting in Rust, which includes trailing commas.

Thanks for the contribution, but closing without merging.  It might make sense to discuss your contributions before raising a PR to help ensure they are welcome.

Issue #10468, opened 2021-05-02
lsp: panic with multi-cursor editing within import urls


danopia:
repro on deno 1.9.2:

```typescript
import {  } from "https://deno.land//sources/from-timer.ts";
import {  } from "https://deno.land//sources/just.ts";
```

Put your cursor where the `x` is supposed to be, and multi-cursor down:

![Screenshot 2021-05-02 12 19 58](https://user-images.githubusercontent.com/40628/116809868-bd78cf80-ab40-11eb-8459-3c8523cb1862.png)

Typing the 'x' here panics the LSP:

```
thread 'main' panicked at 'range end index 4 out of range for slice of length 3', cli/lsp/registries.rs:446:55
```

source line: https://github.com/denoland/deno/blob/v1.9.2/cli/lsp/registries.rs#L446

This happens every time for me.

```
deno 1.9.2 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.5
typescript 4.2.2
```

kitsonk:
I think there are some assumptions about these being only single lines and so I think we need to ignore the requests that span more than one line.

kitsonk:
Hmmm... I am trying it on main and I can't reproduce, no panics:

https://user-images.githubusercontent.com/1282577/117768207-662bdb00-b275-11eb-8ad8-6ed9125d0e02.mp4

kitsonk:
This wasn't multi-cursor editing, it had to do with a logic error when auto completing with a registry and "backspacing".  Basically we guessed there were more tokens than there were to parse for auto-completion.

Issue #10194, opened 2021-04-15
Registry auto discovery in LSP


lucacasonato:
We should reintroduce the registry auto discovery we had in extension v2. Here is how it works:

When the LSP identifies an origin in a specifier for the first time it will try to fetch the well-known manifest for that origin. If a manifest is found, a message pops up asking the user if they want to enable import completions for that origin. If they press `Yes` the origin will be added to the `deno.suggest.imports.hosts` list and set to `true`. If they press no the origin will be added to that list, and set to `false`.

Example:
![image](https://user-images.githubusercontent.com/7829205/114862426-d0746b80-9dee-11eb-97b9-c6df8a22a575.png)

Auto discovery can be disabled with the `deno.suggest.imports.autoDiscovery: false` option. It is set to true by default.

PR #10813, opened 2021-06-01
feat(lsp): registry auto discovery

kitsonk:
Closes: #10194
Fixes: #10468

(this currently includes #10812 which I will rebase once that is merged)


lucacasonato (approved):
LGTM. Tried it locally with VSC and works as expected.

Issue #10776, opened 2021-05-27
lsp: Auto import types as `type`


itohatweb:
**Is your feature request related to a problem? Please describe.**
Auto import currently does not import types using `type` this can lead to issues users will get when using `--no-check` and bundler.
Currently I'm adding `type` manually but it being automated would be nice. 

**Describe the solution you'd like**
Auto import types like `import type { someType } from "./some/file/path.ts"`

kitsonk:
This would need to originate from the language server service of TypeScript, and I don't think it currently has that feature, but need to look into it more.

luke-john:
Hey, TypeScript does this when tsconfig has the value of `importsNotUsedAsValues` set to `error`.

https://twitter.com/thundubeedi/status/1396989095780384771

From the looks of things though, deno does not support that tsconfig option https://deno.land/manual/typescript/configuration#what-an-implied-tsconfigjson-looks-like

kitsonk:
Actually, we don't want people "messing" with that value, so we disable it.  I will look at setting that on the language service on the lsp and see if that resolves things.

luke-john:
Just a note, some people have strong feelings this shouldn't occur by default, so by setting it on the language server in a hidden way, you may end up with reports asking for the previous behaviour 🤷‍♂️ .

Looks like there are settings on deno_vscode isolated from tsconfig already though, so you may be fine.

kitsonk:
> some people have strong feelings this shouldn't occur by default

People who have strong opinions about it don't appreciate how Deno works. 😄 We have to reserve things like module resolution to the internals, so there were reasons why we already ignored that setting, because someone using Deno (or the Deno LSP) should ensure that their code is written in a "transpile safe" way.

kitsonk:
As discussed in #10814, the only way to get it to happen has some undesirable side effects, so there is no obvious way to deliver this feature at the moment.

PR #10814, opened 2021-06-01
fix(lsp): code action imports of types are `import type`

kitsonk:
Fixes: #10776

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -62,7 +62,8 @@ lazy_static::lazy_static! {
   static ref IMPORT_SPECIFIER_RE: Regex = Regex::new(r#"\sfrom\s+["']([^"']*)["']"#).unwrap();
 }
 
-const SUPPORTED_EXTENSIONS: &[&str] = &[".ts", ".tsx", ".js", ".jsx", ".mjs"];
+const SUPPORTED_EXTENSIONS: &[&str] =
+  &[".ts", ".d.ts", ".tsx", ".js", ".jsx", ".mjs"];
 
 /// Category of self-generated diagnostic messages (those not coming from)
 /// TypeScript.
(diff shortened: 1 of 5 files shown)


lucacasonato:
Wouldn't enabling `"importsNotUsedAsValues": "error"` result in more diagnostics than in the CLI (where this is not set to "error")?

I.e. wouldn't this error in LSP, but not `deno run`?

```ts
import { Foo } from "https://example.com/foo.ts";
const foo: Foo = { };
```

kitsonk:
It does but, a) there is no other way to get TypeScript to import as `import type` when auto importing types, and b) while Deno CLI can run such code, it runs the risk of being "transpile unfriendly", so it encourages better code anyways.  I would be for setting it to `"error"` for CLI as well, as it avoids some problems people can get themselves into (like doing an `export * from "foo"` that is doing a named import from some other module, which we have seen in the wild).

Given the module *a.ts* of:

```ts
export interface A {
  a: string;
}

export const a = "a";
```

The following would have no diagnostics:

```ts
import { A, a } from "./a.ts";

const b: A = {
  a,
};

console.log(a);
```

The following would also not have any diagnostics:

```ts
import type { A } from "./a.ts";

const b: A = {
  a: "a"
};

console.log(a);
```

The following would have `This import is never used as a value and must use 'import type' because 'importsNotUsedAsValues' is set to 'error'.` for the import statement:

```ts
import { A } from "./a.ts";

const b: A = {
  a: "a"
};

console.log(a);
```

lucacasonato:
Mh - this feels like it would slot into a larger discussion about the `importsNotUsedAsValues` option. We used to have it set to "error" (https://github.com/denoland/deno/pull/7413), but then reverted back to "remove" a few weeks later (https://github.com/denoland/deno/pull/7800). The "remove" option is unambiguous already with `isolatedModules: true`, so I don't know.

Very hesitant to land this in LSP without also landing for the rest of CLI. But doing so would result in a lot of churn for users for no big benefit that I can see.

kitsonk:
🤷  I won't lose any sleep over it. It was only because people were reporting it as an issue.

kitsonk:
Let's close this, as the potential pain ain't worth the squeeze.

Issue #10825, opened 2021-06-03
Feature detection `typeof localStorage` trips up permissions, requires `--location` to even start


mitranim:
I'm converting an isomorphic app where 99% of the code runs in both Node/Deno and browsers. It uses checks like `typeof X !== 'undefined' && isObj(X)` to detect various browser APIs and avoid using them when unavailable.

Deno trips up on either of the following, and requires `--location` to even start!

```js
typeof localStorage !== 'undefined'
typeof localStorage !== 'undefined' && isObj(localStorage)
```

This code also runs in browsers. Using Deno-specific checks (if any are available) would require it to feature-detect the `Deno` global first!

Browser globals may also be missing in browsers. The checks aren't _just_ for Deno. As one example, a few years ago, I've dealt with `localStorage` either missing or throwing exceptions on access in iOS Safari in private browsing mode.

The docs make it sound like `--location` is sort of a crutch that should be avoided. Shouldn't Deno allow a cleaner solution here?

kitsonk:
ping @crowlKats probably need to find a way not to do the permission check on the `localStorage` getter, but do it another way.

crowlKats:
This has been discussed many times in the discord server, there is no solution to do this. @nayeemrmn

kitsonk:
There needs to be a way.  You can't have a permission error when feature detecting. There is always a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) which would return a `"object"` but not actually access the local storage (there could be other traps that could be useful as well).  The `getPrototypeOf` trap can be used for `instanceof` for example.

lucacasonato:
We could also consider just not having the global at all if `--location` is not defined. That would be more like what we do for `--unstable` things.

crowlKats:
I think that will be a cause for confusion

ry:
The situation @mitranim has described is very real and problematic.

Luca's solution works. We'd have to make up for the confusion with better documentation.

Alternatively we can define a default location. But those discussions have gone nowhere, so we should default to having `localStorage` be undefined.

kitsonk:
> I think that will be a cause for confusion

The current situation is just as confusing. I think having local storage `undefined` without a `--location` is a good solution and easily document.

I would also personally be in favour of an opaque origin that serves as the default if no `--location` is provided.

nayeemrmn:
[...]

PR #10828, opened 2021-06-03
fix(#10825): location should be undefined without --location

ry:
fixes https://github.com/denoland/deno/issues/10825


extensions/web/12_location.js
```diff
@@ -362,8 +362,9 @@
     locationDescriptor: {
       get() {
         if (location == null) {
-          throw new ReferenceError(
-            `Access to "location", run again with --location <href>.`,
+          console.warn(
+            "Warning: accessing undefined 'location' global, " +
+              "run again with --location",
```
lucacasonato, line 367: This seems unfortunate... `typeof location !== "undefined"` would result in a warning logged to the console.

ry: I can remove it - I wanted to address @crowlKats's concern.

bartlomieju: I agree with Luca... OTOH it will be hard to discover `location` API unless you plough through manual. I guess in this situation not showing warning is preferable

nayeemrmn:
What do we do about the type declaration?

bartlomieju:
> What do we do about the type declaration?

Ah, that's a good point. We don't have a way to add declarations conditionally (unless it's for the `--unstable` flag).

extensions/web/12_location.js
ry: I would suggest that we leave the warning for the time being. Extra warning is better than confusion. If people feel strongly we can remove it later.

ry:
I think we'll just have to leave the type declaration incorrect when `--location` is not provided. The situation is not ideal. I don't want to make `deno types` conditional on `--location` too.

kitsonk:
We should amend the JSDoc to make it clear that `--location` is required to be set to make this available and potentially give an example.  That way, people working on intelligent editors can discover it without having to run into the console message.

kt3k:
How about aligning the behavior with private mode / incognite mode of browsers? Provides localStorage object, but doesn't actually persist data. This doesn't break type declaration, and doesn't break feature detection code (though the result of feature detection might be wrong)

caspervonb:
> Provides localStorage object, but doesn't actually persist data.

Is it no-ops, or is the localStorage effectively an instance of the session storage implementation?

caspervonb:
> Is it no-ops, or is the localStorage effectively an instance of the session storage implementation?

To answer my own question, both.
Safari has a null storage implementation, chrome seems to use a temporary storage location.

bartlomieju:
This is stale. Closing without a merge.

PR #10830, opened 2021-06-03
fix(cli): output json with trailing newline character

magurotuna:
From POSIX perspective, there should be a trailing newline character in every line.
But `serde_json::to_writer_pretty` doesn't output a trailing newline character, so this PR adds the call to `writeln!` to explicitly output it.
This will affect two commands, `deno info --json` and `deno doc --json`.


lucacasonato (approved):
LGTM, thanks.

Issue #10508, opened 2021-05-05
`op_http_request_next` is not completed when the underlying listener is closed.


kt3k:
When running the following script:

```js
async function serve(l) {
  for await (const conn of l) {
    (async () => {
      for await (const { respondWith } of Deno.serveHttp(conn)) {
        respondWith(new Response("hello"));
      }
    })();
  }
}
const l = Deno.listen({ port: 4500 });
serve(l);
await new Promise((resolve) => setTimeout(resolve, 300));
const res = await fetch("http://localhost:4500/");
const text = await res.text();
l.close();
await new Promise((resolve) => setTimeout(resolve, 300));
console.log(Deno.metrics());
```

`op_http_request_next` metrics shows like the below:

```jsonc
    op_http_request_next: {
      opsDispatched: 2,
      opsDispatchedSync: 0,
      opsDispatchedAsync: 2,
      opsDispatchedAsyncUnref: 0,
      opsCompleted: 1,
      opsCompletedSync: 0,
      opsCompletedAsync: 1,
      opsCompletedAsyncUnref: 0,
      bytesSentControl: 0,
      bytesSentData: 0,
      bytesReceived: 0
    },
```

So the second `op_http_request_next` call is never finished, and this program can't finish gracefully.

Is there any way to finish this second `op_http_request_next` call?

kt3k:
Another finding is that even when I closed HttpConn explicitly, the last `op_http_request_next` call never finishes.

```js
const delay = (n) => new Promise((resolve) => setTimeout(resolve, n));
const httpConnList = []
async function serve(l) {
  for await (const conn of l) {
    (async () => {
      const c = Deno.serveHttp(conn);
      httpConnList.push(c);
      for await (const { respondWith } of c) {
        respondWith(new Response("hello"));
      }
    })();
  }
}
const l = Deno.listen({ port: 4500 });
serve(l);
await delay(300);
let res = await fetch("http://localhost:4500/");
let text = await res.text();
httpConnList.forEach((conn) => conn.close());
l.close();
await delay(300);
console.log(Deno.metrics().ops.op_http_request_next);
```

This shows
```
{
  opsDispatched: 2,
  opsDispatchedSync: 0,
  opsDispatchedAsync: 2,
  opsDispatchedAsyncUnref: 0,
  opsCompleted: 1,
  opsCompletedSync: 0,
  opsCompletedAsync: 1,
  opsCompletedAsyncUnref: 0,
  bytesSentControl: 0,
[...]

PR #10836, opened 2021-06-03
fix: hang in op_http_next_request

bartlomieju:
This commit adds "CancelHandle" to "ConnResource" and changes
"op_http_next_request" to await for the cancel signal. In turn
when async iterating over "Deno.HttpConn" the iterator breaks
upon closing of the resource.

Ref https://github.com/denoland/deno/issues/10508

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -339,3 +339,37 @@ unitTest(
     await promise;
   },
 );
+
+unitTest(
+  { perms: { net: true } },
+  async function httpServerNextRequestResolvesOnClose() {
+    const delay = (n: number) =>
+      new Promise((resolve) => setTimeout(resolve, n));
+    const httpConnList: Deno.HttpConn[] = [];
+
+    async function serve(l: Deno.Listener) {
+      for await (const conn of l) {
+        (async () => {
+          const c = Deno.serveHttp(conn);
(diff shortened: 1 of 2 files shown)


runtime/ops/http.rs
```diff
@@ -543,6 +557,10 @@ impl Resource for ResponseBodyResource {
   fn name(&self) -> Cow<str> {
     "responseBody".into()
   }
+
+  fn close(self: Rc<Self>) {
+    self.cancel.cancel()
```
piscisaureus, line 562: Why cancel a response?
Note that the general pattern that we use is to cancel read/accept/connect ops but not write ops.

cli/tests/unit/http_test.ts
```diff
@@ -339,3 +339,37 @@ unitTest(
...
+          }
+        })();
+      }
+    }
+
+    const l = Deno.listen({ port: 4500 });
+    serve(l);
+
+    await delay(300);
+    let res = await fetch("http://localhost:4500/");
```
piscisaureus, line 366: const

runtime/ops/http.rs
bartlomieju: Ooops, looks like artifact from initial implementation. Removed!

piscisaureus (approved):
LGTM, great to have this fixed.

PR #10841, opened 2021-06-04
fix(lsp): refactor, fix issues and add benchmark for code lens

kitsonk:
This makes it easier to add the test code lens, but also paves a path to migrate tsc code lenses over to an swc AST and away from the tsc `NavigationTree` AST.


bartlomieju (approved):
LGTM, no comments

Issue #10837, opened 2021-06-03
Missing performance.timeOrigin


lucacasonato:

acrodrig:
Is there any work around while it gets implemented?

idanran:
I am looking forward to the implementation as soon as possible

bartlomieju:
For anyone interested in working on this, https://github.com/denoland/deno/pull/10843 might be a good reference point.

GJZwiers:
I would like to work on this!

bartlomieju:
@GJZwiers  please do!

PR #10843, opened 2021-06-04
feat(runtime-js.performance) - add timeOrigin

JoviDeCroock:
Resolves: #10837

Currently this is a naive timestamp while the browser will inject a `HighResTimestamp` would this be better achieved by using something like [`SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html) and have it as an `op_sync` call?

### DIFF
--- a/extensions/timers/02_performance.js
+++ b/extensions/timers/02_performance.js
@@ -8,6 +8,8 @@
   const customInspect = Symbol.for("Deno.customInspect");
   let performanceEntries = [];
 
+  const timeOrigin = opNow();
+
   function findMostRecent(
     name,
     type,
@@ -211,6 +213,10 @@
       if (key != illegalConstructorKey) {
         throw new TypeError("Illegal constructor.");
       }
+    }
+
+    get timeOrigin() {
+      return timeOrigin;
     }
 
     clearMarks(markName) {
--- a/extensions/timers/02_performance.js
+++ b/extensions/timers/02_performance.js
@@ -2,13 +2,12 @@
 "use strict";
 
 ((window) => {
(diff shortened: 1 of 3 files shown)


runtime/js/40_performance.js
```diff
@@ -9,6 +9,8 @@
   const customInspect = Symbol.for("Deno.customInspect");
   let performanceEntries = [];
 
+  const timeOrigin = opNow();
+
```
lucacasonato, line 13: This isn't quite correct. Firstly, top level code is actually executed during snapshotting (during build). That means `timeOrigin` would be hardcoded to the time of build. Secondly `opNow` returns the time relative to the start of the program. `timeOrigin` is meant to be the timestamp since unix epoch.

lucacasonato: I suggest you call `new Date()` during `bootstrapRuntime()` in 99_main.js to get the time since unix epoch.

lucacasonato (commented):
`window.timeSinceEpoch` is not a browser standard. That won't work 🙃

lucacasonato:
We are also not going to add a `Deno.timeSinceEpoch`. This data has to be exposed to the performance API privately, without a publicly exposed "timeSinceEpoch" variable.

JoviDeCroock:
Should this be an attribute rather than a `get` so we have conformance for the other test as well?

kt3k:
We also need to update type declaration for `performance` https://github.com/denoland/deno/blob/6460eadaa178b3a9d09d04205977e4f659fe8fff/cli/dts/lib.deno.shared_globals.d.ts

PR #10844, opened 2021-06-04
refactor(web): use encoding_rs for text encoding

lucacasonato:
This commit removes all JS based text encoding / text decoding. Instead
encoding now happens in Rust via encoding_rs (already in tree). This
implementation retains stream support, but adds the last missing
encodings. We are incredibly close to 100% WPT on text encoding now.

This should reduce our baseline heap by quite a bit.


tools/wpt/expectation.json
```diff
@@ -102,46 +91,23 @@
...
     "textdecoder-utf16-surrogates.any.html": true,
-    "textencoder-constructor-non-utf.any.html": [
-      "Encoding argument supported for decode: EUC-JP",
-      "Encoding argument supported for decode: ISO-2022-JP",
-      "Encoding argument supported for decode: Shift_JIS",
-      "Encoding argument supported for decode: EUC-KR",
-      "Encoding argument supported for decode: x-user-defined"
-    ],
+    "textencoder-constructor-non-utf.any.html": true,
     "textencoder-utf16-surrogates.any.html": true,
```
ry, line 111: Awesome

core/lib.deno_core.d.ts
```diff
@@ -38,5 +38,8 @@ declare namespace Deno {
 
     /** Get heap stats for current isolate/worker */
     function heapStats(): Record<string, number>;
+
+    /** Encode a string to its Uint8Array representation. */
+    function encode(input: string): Uint8Array;
```
ry, line 43: Why does this need to be a Deno.core API? Can't it be a normal op?

lucacasonato: It is already a Deno.core API (`Deno.core.encode`). This just adds it to our internal typings. It has no effect on runtime.

tools/wpt/expectation.json
```diff
@@ -102,46 +91,23 @@
...
-      "x-user-defined => x-user-defined"
+      "unicode11utf8 => UTF-8",
+      "unicode20utf8 => UTF-8",
+      "x-unicode20utf8 => UTF-8",
+      "unicodefffe => UTF-16BE",
+      "csunicode => UTF-16LE",
+      "iso-10646-ucs-2 => UTF-16LE",
+      "ucs-2 => UTF-16LE",
+      "unicode => UTF-16LE",
+      "unicodefeff => UTF-16LE"
```
lucacasonato, line 106: These regressions are tracked here: https://github.com/hsivonen/encoding_rs/issues/50

core/lib.deno_core.d.ts
lucacasonato: But to answer your question, yeah can be a normal op. Let's tackle that in a followup though, this PR is large enough as is.

extensions/web/lib.rs
```diff
@@ -15,12 +34,197 @@ pub fn init() -> Extension {
...
+    let (result, _, _) = decoder.decode_to_string(&data, &mut output, !stream);
+    match result {
+      CoderResult::InputEmpty => Ok(output),
+      CoderResult::OutputFull => Err(range_error("Provided buffer too small.")),
+    }
+  }
+}
+
+struct TextDecoderResource {
+  decoder: RefCell<Decoder>,
```
bartlomieju, line 187: Is this `RefCell<>` needed? Seems superfluous

bartlomieju (commented):
4k lines net removed, me gusta :)

extensions/web/lib.rs
lucacasonato: Yes, because I need mutable access to the decoder, and I can't do that with Rc alone.

bartlomieju: Right, non-issue then

extensions/web/lib.rs
```diff
@@ -15,12 +34,197 @@ pub fn init() -> Extension {
...
+  ignore_bom: bool,
+  fatal: bool,
+}
+
+fn op_encoding_normalize_label(
+  _state: &mut OpState,
+  label: String,
+  _: (),
+) -> Result<String, AnyError> {
+  let encoding = Encoding::for_label(label.as_bytes()).ok_or_else(|| {
```
andreubotella, line 90: You might want to use [`Encoding::for_label_no_replacement`](https://docs.rs/encoding_rs/0.8.28/encoding_rs/struct.Encoding.html#method.for_label_no_replacement) here rather than detecting the replacement encoding in JS.

lucacasonato: Done, thanks.

bartlomieju (approved):
LGTM, nice work, this greatly simplifies encoding and hopefully will have a visible impact on the JS heap size.

leoc11:
Screenshot_2021-06-06-11-44-49-66
I thought this would improve textencoder performance, but sad, this pr make it slower.

zandaqo:
@leoc11  And quite significantly if that test to be believed. As a side note, V8's implementation [has been slower](https://bugs.chromium.org/p/v8/issues/detail?id=4383) than JS for the longest time. One has to wonder what makes text encoding slow in these native implementations, do they handle more edge cases or is it due to data marshalling overhead?

lucacasonato:
It's unrelated to the serialization overhead. We are hitting a deopt in V8. I'll throw up a PR to fix this.

PR #10851, opened 2021-06-05
feat: make 'deno lint' stable

bartlomieju:
We've discussed this several times, and 1.11 is high time to stabilize linter.


lucacasonato (approved):
SGTM. I think there is precedent that rules can change on a stable linter (see clippy), so I think this distinction was pretty arbitrary anyway.

bartlomieju:
> SGTM. I think there is precedent that rules can change on a stable linter (see clippy), so I think this distinction was pretty arbitrary anyway.

Agreed, and some users were discouraged from using linter because it was "unstable".

PR #10852, opened 2021-06-05
tests: run wpt scripts with Deno.core.evalContext

lucacasonato:
This means scripts are now run in script context, and there are better
stack traces.

Also fixes a panic in core related to `Deno.core.evalContext`.

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -1,6 +1,7 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 use crate::error::AnyError;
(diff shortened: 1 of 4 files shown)


bartlomieju:
Heads up: `Deno.core.evalContext` is used in `std/node` - we should make sure it doesn't break before landing this PR.

lucacasonato:
Ran std tests, and they still all pass.

bartlomieju (approved):
> Ran std tests, and they still all pass.

Thanks, LGTM then

PR #10859, opened 2021-06-05
chore(wpt/webcrypto): remove irrelevant wpt for non-secure contexts

littledivy:
Disables `WebCryptoAPI/historical.any.html`. It was passing on main because of the missing SubtleCrypto implementation.

But fails WPT tests in #9614 

```
 "/WebCryptoAPI/historical.any.html - Non-secure context window does not have access to crypto.subtle"
 "/WebCryptoAPI/historical.any.html - Non-secure context window does not have access to CryptoKey"
```


lucacasonato:
I would like to show that they are failing then, instead of removing them outright.

PR #10863, opened 2021-06-05
feat(fetch): implement abort

### LINKED ISSUE #5471: Add support for abortable fetch
Reference: https://developers.google.com/web/updates/2017/09/abortable-fetch

Also `Request` should have a `signal` property.

lucacasonato:
This commit introduces fetch aborting via an AbortSignal.

Closes #5471

### DIFF
--- a/tools/wpt/expectation.json
+++ b/tools/wpt/expectation.json
@@ -1095,7 +1095,6 @@
         "Window interface: operation fetch(RequestInfo, optional RequestInit)"
       ],
       "abort": {
-        "cache.https.any.html": false,
         "general.any.html": [
           "response.arrayBuffer() rejects if already aborted",
           "response.blob() rejects if already aborted",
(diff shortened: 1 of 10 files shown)


tools/wpt/expectation.json
```diff
@@ -1088,14 +1087,23 @@
...
-      ]
+      ],
+      "abort": {
+        "cache.https.any.html": false,
+        "general.any.html": [
+          "response.arrayBuffer() rejects if already aborted",
+          "response.blob() rejects if already aborted",
+          "response.formData() rejects if already aborted",
+          "response.json() rejects if already aborted",
+          "response.text() rejects if already aborted"
```
lucacasonato, line 1104: These tests are not easy to solve without some significant refactors, so I will leave them failing for now. They are timing related (the promise resolves one microtask too late), and very unlikely to be noticed by a user.

bartlomieju: Is this related to `fetch` implementation or is it more general problem (like wrong timing of ops)?

bartlomieju (approved):
LGTM, cleanly implemented, no comments

tools/wpt/expectation.json
lucacasonato: I am not quite sure yet

PR #10869, opened 2021-06-06
tests: generate and upload wptreport.json

lucacasonato:
These reports can be consumed by tools like `wptreport` or
https://wpt.fyi. The old style report could be removed in a future PR
when wpt.deno.land is updated.


bartlomieju (approved):
LGTM

PR #10875, opened 2021-06-07
tests: fix flakey lsp integration test

kitsonk:


lucacasonato (approved):
Always a little weary about adding timeouts to fix flaky tests, but if it works reliably 🤷‍♂️

LGTM

kitsonk:
There is no request and response here. They are unsolicited notifications, so we have to wait until they have all been sent, otherwise our assertions fail (or we wait forever for a notification that never comes.

PR #6717, opened 2020-07-12
fix(Deno.ppid): improve error message when --unstable is missing

uki00a:
**test.ts**
```typescript
console.log(Deno.ppid)
```

**Before this fix:**
```shell
$ cargo run -- run test.ts
error: TS2551 [ERROR]: Property 'ppid' does not exist on type 'typeof Deno'. Did you mean 'pid'?
console.log(Deno.ppid);
                 ~~~~
    at file:///home/uki00a/work/test.ts:1:18

    'pid' is declared here.
      export const pid: number;
                   ~~~
        at asset:///lib.deno.ns.d.ts:46:16
```

**After this fix:**
```shell
$ cargo run -- run test.ts
error: TS2551 [ERROR]: Property 'ppid' does not exist on type 'typeof Deno'. 'Deno.ppid' is an unstable API. Did you forget to run with the '--unstable' flag, or did you mean 'pid'?
console.log(Deno.ppid);
                 ~~~~
    at file:///home/uki00a/work/test.ts:1:18

    'pid' is declared here.
      export const pid: number;
                   ~~~
        at asset:///lib.deno.ns.d.ts:124:16
```

PR #10880, opened 2021-06-07
fix: move stable/unstable types/APIs to their correct places

### LINKED ISSUE #10827: stable/unstable documentation vs implementation discrepancy
The following Deno APIs are documented as unstable, but are available in stable:

* `Deno.ppid`
* `Deno.memoryUsage`
* `Deno.sleepSync`

On the other hand, the following Deno APIs are documented as stable, but is only available in unstable:

* `Deno.shutdown`

Is there a reason for this discrepancy? We're building a [Deno namespace shim for Node](https://github.com/fromdeno/deno.ns/), and often have to make decisions on whether to add APIs on this side or the other side of the wall.

MKRhere:
* Moved `ppid` and `memoryUsage` types from deno.unstable to deno.ns.
* Moved `sleepSync` to `unstable` object, `shutdown` to stable object.

Closes #10827 as instructed by https://github.com/denoland/deno/issues/10827#issuecomment-855810244

Signed-off-by: Muthu Kumar <muthukumar@thefeathers.in>


MKRhere:
As far as I understand, `unstable_ts2551` was there to make sure TS errors correctly on using unstable API with a suggestion to a similarly named API. This test failed because `Deno.ppid` is no longer unstable.

Additionally, in the latest test I run locally, TS throws the same expected message, but with code `ts2339`. I attempted fixing the test, but have retracted it since I don't fully understand the issue. Would appreciate help if someone stumbles upon this.

ry:
Hi @MKRhere, probably we can just remove that test.  For reference here is the original PR that added it https://github.com/denoland/deno/pull/6717

I'm a bit confused about this PR tho. When was ppid and memoryUsage made stable?

MKRhere:
@ry Yes, I suspected the test might need removing.

While working on [@fromdeno/deno.ns](https://github.com/fromdeno/deno.ns) I noticed that ppid and memoryUsage are in deno stable, but the types/documentation lists them in unstable [[1]](https://doc.deno.land/builtin/unstable#Deno.ppid) [[2]](https://doc.deno.land/builtin/unstable#Deno.MemoryUsage)

image

I raised issue https://github.com/denoland/deno/issues/10827 to which @lucacasonato helpfully responded which of these were misplaced.

ry:
@MKRhere Ok, makes sense now. I think memoryUsage and ppid should have landed in unstable, but erroneously landed in stable. They both seem like reasonable APIs so let's just agree they should be stable now (even tho there hasn't been a discussion about it).

sleepSync should remain unstable, makes sense.

So if you remove that test, I think we can land this. Thank you for the PR.

MKRhere:
Indeed, https://github.com/denoland/deno/commit/bf99039ea944c32f0247ad8b3d1057f20f921b40 adds `memoryUsage`, and lands in `finalDenoNs` instead of `denoNsUnstable`.

`ppid` however, has been in Deno for over a year now, from before internals were ported from TS to JS. Makes sense for it to be stable now.

Ran tests locally, updated PR 🙂

MKRhere:
Can we land this? @ry

bartlomieju (approved):
LGTM, thanks for sorting this out @MKRhere!

PR #10884, opened 2021-06-07
chore: update wpt

lucacasonato:
This commit updates wpt. It tries to get us ever closer to eventually
stop floating patches, and being able to use wpt upstream.


bartlomieju (approved):

andreubotella:
You might want to add `encoding/single-byte-decoder.window.html?TextDecoder` (see web-platform-tests/wpt#29241) to the expectations.

PR #10888, opened 2021-06-07
chore(tools): Fix stdout buffer of launched process getting full causing tools/lint.js to potentially hang

dsherret:
`./tools/lint.js` was never completing on my machine (was fine on WSL, but not Windows). I believe it is due to the stdout buffer of the launched process getting full when `Deno.run` is called.

Fixed by just always reading the output, then checking the status. ~~Seems like some clippy errors surfaced after doing that as well and those are also fixed.~~ Edit: These were Windows only clippy issues. Will fix them as well.

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -611,11 +611,11 @@ impl UnaryPermission<EnvDescriptor> {
       if state == PermissionState::Prompt {
         if permission_prompt(&format!("env access to \"{}\"", env)) {
           self.granted_list.retain(|env_| env_.0 != env);
-          self.granted_list.insert(EnvDescriptor(env.to_string()));
+          self.granted_list.insert(EnvDescriptor(env));
           PermissionState::Granted
         } else {
           self.denied_list.retain(|env_| env_.0 != env);
-          self.denied_list.insert(EnvDescriptor(env.to_string()));
+          self.denied_list.insert(EnvDescriptor(env));
           self.global_state = PermissionState::Denied;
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM!

lucacasonato:
Mh, our rustc version doesn't like that. What rustc / clippy are you on?

dsherret:
@lucacasonato it's because of the `#[cfg(windows)]` code changing it to a string (compiles on my machine on Windows, doesn't compile on Linux). I will fix it... one second...

runtime/permissions.rs
```diff
@@ -605,17 +605,20 @@ impl UnaryPermission<EnvDescriptor> {
...
       if state == PermissionState::Prompt {
         if permission_prompt(&format!("env access to \"{}\"", env)) {
           self.granted_list.retain(|env_| env_.0 != env);
-          self.granted_list.insert(EnvDescriptor(env.to_string()));
+          self.granted_list.insert(EnvDescriptor(env));
           PermissionState::Granted
         } else {
           self.denied_list.retain(|env_| env_.0 != env);
-          self.denied_list.insert(EnvDescriptor(env.to_string()));
+          self.denied_list.insert(EnvDescriptor(env));
```
dsherret, line 621: This was previously a clippy error only on Windows due to the conditional code above (`#[cfg(windows)]`)

lucacasonato (approved):
Ah, that makes sense. LGTM the second :-)

PR #10890, opened 2021-06-08
feat(lsp): add additional logging information

kitsonk:
We have had a few issues where users have had strange environment configurations that have lead to confusion when using the LSP.  This PR adds some additional logging to provide more information in the extension output channel.  It adds the path to the executable and the fully resolved import map and config files.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -361,9 +361,13 @@ impl Inner {
           import_map_str
         ))
       }?;
-      let import_map_path = import_map_url
-        .to_file_path()
-        .map_err(|_| anyhow!("Bad file path."))?;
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
LGTM, is it a `feat` though?

kitsonk:
It is a change to production, user facing code. Is there a more correct semantic commit message? https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716

lucacasonato:
Maybe refactor? I never considered debug logs to be a user facing feature, but I guess it makes sense. Anyway not particularly important

kitsonk:
> Maybe refactor?

Well it isn't refactoring code... it s a change in behaviour.  If the change of output of `deno version` would be a feature, then this is a feature.

PR #10900, opened 2021-06-08
chore: Fix broken test on Windows

dsherret:
`cat` does not exist by default on Windows. GitHub actions adds a lot of stuff to the Windows runner to make it work, which is why the CI is ok (I'm often surprised at what works in there).

### DIFF
--- a/cli/tests/089_run_allow_list.ts
+++ b/cli/tests/089_run_allow_list.ts
@@ -7,9 +7,7 @@ try {
 }
 
 const proc = Deno.run({
-  cmd: Deno.build.os === "windows"
-    ? ["cmd", "/c", "type", "089_run_allow_list.ts"]
-    : ["cat", "089_run_allow_list.ts"],
+  cmd: ["curl", "--help"],
   stdout: "null",
 });
 console.log((await proc.status()).success);
--- a/cli/tests/089_run_allow_list.ts
+++ b/cli/tests/089_run_allow_list.ts
@@ -7,7 +7,9 @@ try {
 }
 
 const proc = Deno.run({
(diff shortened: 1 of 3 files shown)


test_util/src/lib.rs
```diff
@@ -76,7 +76,10 @@ lazy_static! {
 }
 
 pub fn root_path() -> PathBuf {
-  PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR"), "/.."))
+  PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR")))
+    .parent()
+    .unwrap()
+    .to_path_buf()
```
dsherret, line 82: This just makes the paths that get outputted whiel the tests are running appear resolved.

cli/tests/089_run_allow_list.ts
```diff
@@ -7,7 +7,9 @@ try {
 }
 
 const proc = Deno.run({
-  cmd: ["cat", "089_run_allow_list.ts"],
+  cmd: Deno.build.os === "windows"
+    ? ["cmd", "/c", "type", "089_run_allow_list.ts"]
+    : ["cat", "089_run_allow_list.ts"],
```
dsherret, line 12: I don't like doing these conditionals. Let me know if you can think of something that exists on all three operating systems by default.

lucacasonato: `curl` exists on all three now AFAIK. Or maybe `echo`?

dsherret: Oh, `echo` is good. Thanks!

lucacasonato (approved):

Issue #10903, opened 2021-06-08
Enable WPT worker tests


lucacasonato:
This will require some more tinkering because we don't support "classic" style workers (non esm). To get it working we'll need to do some source rewrites on the `.any.worker.html` tests.

PR #10904, opened 2021-06-09
fix(runtime): early binding to dispatchEvent in workers

kitsonk:
I was working with workers and realised that if your overwrite `dispatchEvent` in the global scope, messages do not get sent to event listeners of `"message"` any longer.  This is because the runtime binds to the global `dispatchEvent` every time it tries to dispatch a message.

I checked that in browsers, if you  override `dispatchEvent` in a worker, messages still get sent to the event listeners, and so I have amended the runtime to bind early to the global `dispatchEvent` and this ensures the behaviour works as expected.

### DIFF
--- a/cli/tests/fix_worker_dispatchevent.ts
+++ b/cli/tests/fix_worker_dispatchevent.ts
@@ -1,0 +1,43 @@
+const code = `
+addEventListener("message", () => {
+  postMessage("pong");
+});
+
+const context = new EventTarget();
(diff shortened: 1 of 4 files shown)


kitsonk:
I haven't checked if this is covered by the WPT, so I added an integration test.  Ref #10903.

bartlomieju (approved):
LGTM, nice catch

Issue #6654, opened 2020-07-06
How to set proxy for some fetch request not all


XGHeaven:
I want to set proxy for some fetch request and others not. How to do that.

I know set `http_proxy`/`https_proxy` env which only proxy all fetch request.

jacobgc:
As far as I'm aware, the fetch api doesn't support setting a proxy on an individual request basis.

bartlomieju:
@jacobgc is correct. Please describe your use case and continue in #3444.

chibat:
I tried creating the following.
https://github.com/chibat/rest_client_with_proxy

but we will be happy if it is implemented in runtime api.

PR #10907, opened 2021-06-09
feat(fetch): add programmatic proxy

chibat:
Releated Issue: #6654 (It has been closed..)

I think we need a way to set a proxy individually.

```ts
const client = Deno.createHttpClient({ proxy: { url: "http://myproxy.com:8080" } });
const response = await fetch("https://myserver.com", { client });
```


kitsonk:
We really try to align to the WHATWG standard for Fetch and the init is supposed to just be passed through to `Request`: https://fetch.spec.whatwg.org/#fetch-method and the `RequestInit` does not include such a property, which would mean this would be specific to Deno and not something that we would consider without significant discussion _before_ considering a PR.

That being said, the standard includes a concept of a [client](https://fetch.spec.whatwg.org/#concept-request-client) but which resides on the `Request` object, not something processed by the `fetch()` method directly.  There is no way of a user setting this with init options though.

In other situations we have consider Deno specific "extensions" to the standard to reside in a `deno` property on an init object, _if_ the standard ignores excess properties and after discussion to make sure we are doing the right thing.  It appears the `Request` constructor [ignores](https://fetch.spec.whatwg.org/#request-class) excess properties in the init, so having `{ deno: { client } }` might be acceptable.

@lucacasonato thoughts?

lucacasonato:
@kitsonk This is implemented via `Deno.createHttpClient`. `Deno.HttpClient` can already be passed to `fetch` / `Request` in the `RequestInit` in a `client` property in unstable.

extensions/fetch/lib.rs
```diff
@@ -524,6 +546,15 @@ pub fn create_http_client(
     builder = builder.add_root_certificate(cert);
   }
 
+  if let Some(proxy) = proxy {
+    let mut reqwest_proxy = reqwest::Proxy::all(&proxy.url)?;
```
lucacasonato, line 550: The proxy URL should go through a network permission check here.

lucacasonato: Or more accurately in the op_create_http_client

chibat: @kitsonk @lucacasonato Thank you for your comments.

I fixed to check a network permission.

lucacasonato (approved):
Looks great, no further comments. Thanks :-)

We will wait with merging until the 1.12 merge window opens later next week.

mokeyish:
When will this feature be released?

bartlomieju:
In 1.12 on July 13th

PR #10908, opened 2021-06-09
Remove unstable native plugins

### LINKED ISSUE #8490: Remove unstable native plugins

bartlomieju:
Closes #8490


wperron (approved):

wperron:
Seems like a substantial change (even if the API was under the unstable flag) I think we should include it in the upcoming 1.12 release instead of the next patch release for 1.11

rracariu:
Is there any replacement in the works/planed?

eliassjogreen:
Would a pr adding classic C ffi support (`dlopen`, `dlsym`, `dlerror`, `dlclose`) be welcome or should i wait for deno 2.0? Removing plugin support also removes the possibility to do native calls which is a thing which can be very useful for everything from webview support, native windowing for WebGPU, fast cpu optimized code and lots of other cool things. I think removing plugins would put deno at a disadvantage in many use cases to node and the napi.

bartlomieju:
> Is there any replacement in the works/planed?

Yes, FFI API proposed in #11152

ry (changes requested):
Let's discuss with @piscisaureus

ry (approved):
LGTM

We will land FFI soon after.

PR #10911, opened 2021-06-09
refactor(swc): Change AST parsing error to return struct with message and location

dsherret:
1. For type stripping in the REPL it is nicer to get a diagnostic without the file name.
2. There actually will only ever be one diagnostic returned currently, so I removed `DiagnosticBuffer`. We can add back something similar, if we ever look at `parser.take_errors()` for diagnostics.
3. I was confused about what `EmitOptions#check_js` did then realized it was never used for emitting (therefore, not an emit option). I removed this and changed the other code to just look at the value from the tsconfig.json.

### DIFF
--- a/cli/ast.rs
+++ b/cli/ast.rs
@@ -192,9 +192,6 @@ pub enum ImportsNotUsedAsValues {
 /// Options which can be adjusted when transpiling a module.
 #[derive(Debug, Clone)]
 pub struct EmitOptions {
-  /// Indicate if JavaScript is being checked/transformed as well, or if it is
-  /// only TypeScript.
-  pub check_js: bool,
   /// When emitting a legacy decorator, also emit experimental decorator meta
   /// data.  Defaults to `false`.
   pub emit_metadata: bool,
@@ -221,7 +218,6 @@ pub struct EmitOptions {
 impl Default for EmitOptions {
   fn default() -> Self {
     EmitOptions {
-      check_js: false,
       emit_metadata: false,
       imports_not_used_as_values: ImportsNotUsedAsValues::Remove,
       inline_source_map: true,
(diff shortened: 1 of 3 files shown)


cli/ast.rs
```diff
@@ -89,57 +82,18 @@ impl std::fmt::Display for Location {
   }
 }
 
-/// A buffer for collecting diagnostic messages from the AST parser.
-#[derive(Debug)]
-pub struct DiagnosticBuffer(Vec<String>);
```
dsherret, line 94: The code only possibly returns one diagnostic. Removed and replaced with a single `Diagnostic` error struct. This will also allow me to format a diagnostic specifically for the REPL (no file name).

cli/ast.rs
```diff
@@ -192,9 +146,6 @@ pub enum ImportsNotUsedAsValues {
 /// Options which can be adjusted when transpiling a module.
 #[derive(Debug, Clone)]
 pub struct EmitOptions {
-  /// Indicate if JavaScript is being checked/transformed as well, or if it is
-  /// only TypeScript.
-  pub check_js: bool,
```
dsherret, line 197: Not used by the emit code, so removed (see PR description for more details).

cli/ast.rs
```diff
@@ -424,31 +373,17 @@ pub fn parse_with_source_map(
...
-    let mut diagnostic = err.into_diagnostic(&handler);
-    diagnostic.emit();
-
-    DiagnosticBuffer::from_error_buffer(error_buffer, |span| {
-      sm.lookup_char_pos(span.lo)
-    })
+  let module = parser.parse_module().map_err(move |err| Diagnostic {
+    location: sm.lookup_char_pos(err.span().lo).into(),
+    message: err.into_kind().msg().to_string(),
   })?;
```
dsherret, line 387: The code above was copied from dprint-plugin-typescript when I didn't know the full swc API (my bad). This new code is a bit more straightforward and we can use `parser.take_errors()` in the future to get all the errors if necessary.

bartlomieju: >I think the code above was copied from dprint-plugin-typescript 

That's very likely

bartlomieju (approved):
LGTM

PR #10912, opened 2021-06-09
feat(extensions): Implement WebUSB API

### LINKED ISSUE #9523: Support WebUSB

littledivy:
Closes #9523 

Implements the WebUSB API.


littledivy:
testing this on the CI is a PITA or maybe not even possible without self hosted runners. The closest I've got to testing this without the need to external hardware and tweaking the kernel is using [raw_gadget](https://github.com/xairy/raw-gadget) on Linux for which I made a little test suite located [here](https://github.com/littledivy/deno/tree/webusb-api/extensions/webusb/test).

other alternative ways to test this:
- Actual hardware 
- Linux functionfs/gadgetfs
  - Requires rebuilding kernel. Chrome uses gadgetfs. Not possible with current Github runners.
- Linux [raw_gadget](https://github.com/xairy/raw-gadget) 
  - Perfectly fine for testing locally. Requires loading kernel modules and depends on `udc_core` (not available in Github runners).
- [umockdev](https://github.com/martinpitt/umockdev)
  - Didn't try because raw_gadget just worked. Assuming won't work in Github runners either?

overall the implementation is battle-tested on Linux and not tested on Mac or Windows at all.

kitsonk:
I re-opened this, just because stale bot closed it, but we discussed at the CLI design meeting.  Without a clean way to test it as part of CI without self-hosted runners, we can't really add it, and it isn't enough of a priority for the core team try to find the way, so it would need to be up to the community.

littledivy:
> Without a clean way to test it as part of CI without self-hosted runners, we can't really add it

The current CI runners aren't capable of testing USB (no custom kernel no kernel modules nothing) and without the option of self hosted runners (which I am using [without trouble in the webusb-rs](https://github.com/littledivy/webusb/runs/3707229594) repo) I really can't do much so I'm going to close the PR.

To interested readers, I plan to expose API bindings via FFI for the [webusb](https://github.com/littledivy/webusb) crate in the near future.

PR #10914, opened 2021-06-09
refactor: merge deno_file crate into deno_web

bartlomieju:
This refactor makes it so there's one less crate to publish on each release.


cli/main.rs
```diff
@@ -330,12 +330,11 @@ fn print_cache_info(
 
 pub fn get_types(unstable: bool) -> String {
   let mut types = format!(
-    "{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}",
+    "{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}\n{}",
```
ry, line 333: Unrelated to this PR but this is silly - we should use join...

ry (approved):
LGTM

PR #10915, opened 2021-06-09
refactor(repl): Extract out structs

dsherret:
To help myself understand this code I went through and extracted out two new `ReplEditor` and `ReplSession` structs. The code should functionally be the same, but just reorganized.

It is easier to review this PR commit by commit.

This is preparation for adding type stripping to the REPL.

### DIFF
--- a/cli/tools/repl.rs
+++ b/cli/tools/repl.rs
@@ -295,6 +295,37 @@ impl ReplEditor {
   }
 }
 
+static PRELUDE: &str = r#"
+Object.defineProperty(globalThis, "_", {
+  configurable: true,
+  get: () => Deno[Deno.internal].lastEvalResult,
+  set: (value) => {
+   Object.defineProperty(globalThis, "_", {
+     value: value,
+     writable: true,
+     enumerable: true,
+     configurable: true,
+   });
+   console.log("Last evaluation result is no longer saved to _.");
+  },
+});
+
+Object.defineProperty(globalThis, "_error", {
+  configurable: true,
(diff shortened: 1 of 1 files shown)


cli/tools/repl.rs
```diff
@@ -328,171 +326,263 @@ Object.defineProperty(globalThis, "_error", {
...
+    worker
+      .with_event_loop(
+        session.post_message("Runtime.enable", None).boxed_local(),
+      )
+      .await?;
+
+    // Enabling the runtime domain will always send trigger one executionContextCreated for each
+    // context the inspector knows about so we grab the execution context from that since
+    // our inspector does not support a default context (0 is an invalid context id).
+    let mut context_id: u64 = 0;
```
dsherret, line 348: The code here for getting the `context_id` has not changed as I tried not to change any logic in this PR, but it seems like it should be `session.notifications().find(|n| n.get("method").unwrap().as_str().unwrap() == "Runtime.executionContextCreated")` or whatever and then throw when it can't be found since `context_id == 0` is invalid (as per comment above)?

bartlomieju: It could be actually hardcoded to `1`. `deno_core` currently doesn't support more than one context and there are TODOs scattered around different parts of the codebase mentioning that. I'm fine with leaving as is.

bartlomieju (approved):
LGTM, this is definitely more readable 👍

caspervonb (approved):
LGTM

PR #10918, opened 2021-06-10
docs(getting-started): specify that allow-env and allow-run can have allow lists

grian32:


ry (approved):
LGTM - thank you for correcting this oversight!

PR #10920, opened 2021-06-10
ci: run Linux release builds in sysroot with Ubuntu 18.04 and LLVM

piscisaureus:
~Depends on: https://github.com/denoland/rusty_v8/pull/708~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,38 +21,39 @@ jobs:
             kind: test
             profile: release
           - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
+            kind: test
+            profile: release
+            use_sysroot: true
+          - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
             kind: bench
             profile: release
+            use_sysroot: true
+          - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
+            kind: test
+            profile: debug
           - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
             kind: lint
             profile: debug
-          - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-latest-xl' || 'ubuntu-latest' }}
-            kind: test
-            profile: debug
-          # Warning: Do not upgrade test_release to newer version of ubuntu
-          # runners. We need to link against older version of glibc in order to
-          # run on older systems. glibc in 20.04 is not compatible with 18.04.
-          # See #9484.
-          - os: ubuntu-18.04
-            kind: test
-            profile: release
 
       # Always run main branch builds to completion. This allows the cache to
       # stay mostly up-to-date in situations where a single job fails due to
       # e.g. a flaky test.
       # Don't fast-fail on tag build because publishing binaries shouldn't be
       # prevented if if any of the stages fails (which can be a false negative).
-      fail-fast: ${{ github.event_name == 'pull_request' || (github.ref !=
-        'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')) }}
+      fail-fast: ${{ github.event_name == 'pull_request' ||
+        (github.ref != 'refs/heads/main' &&
+        !startsWith(github.ref, 'refs/tags/')) }}
 
     env:
+      CARGO_TERM_COLOR: always
       RUST_BACKTRACE: full
       RUSTC_FORCE_INCREMENTAL: 1
-      CARGO_TERM_COLOR: always
 
     steps:
       - name: Configure git
-        run: git config --global core.symlinks true
+        run: |
+          git config --global core.symlinks true
+          git config --global fetch.parallel 32
 
       - name: Clone repository
         uses: actions/checkout@v2
@@ -72,9 +73,10 @@ jobs:
           startsWith(github.ref, 'refs/tags/')
         run: |
           mkdir -p target/release
-          tar --exclude=.cargo_home --exclude=".git*" --exclude=target --exclude=third_party/prebuilt -czvf target/release/deno_src.tar.gz -C .. deno
+          tar --exclude=".git*" --exclude=target --exclude=third_party/prebuilt \
+              -czvf target/release/deno_src.tar.gz -C .. deno
 
-      - name: Install rust
+      - name: Install Rust
         uses: hecrj/setup-rust-action@v1
         with:
           rust-version: 1.53.0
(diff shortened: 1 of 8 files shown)


piscisaureus:
| **Before**<sup>1</sup> | **After**<sup>1</sup> |
|:-|:-|
| image | (TBD) |




<sup>1</sup>: Github Actions caches are empty.

.github/workflows/ci.yml
```diff
@@ -287,28 +385,61 @@ jobs:
...
-          deno run --unstable --allow-write --allow-read --allow-net --allow-env --allow-run ./tools/wpt.ts run --quiet --release --json=wpt.json --wptreport=wptreport.json
-          gzip ./wptreport.json
+          ./target/release/deno run --allow-env --allow-net --allow-read \
+                                    --allow-run --allow-write --unstable \
+                                ./tools/wpt.ts setup
+          ./target/release/deno run --allow-env --allow-net --allow-read     \
+                                    --allow-run --allow-write --unstable     \
+                                ./tools/wpt.ts run --no-build --quiet        \
+                                                   --release --json=wpt.json \
+                                                   --wptreport=wptreport.json
```
lucacasonato, line 442: So now WPT only run on release, and not on debug? If we keep going with removing WPT targets at this rate, we won't run WPT at all next month...

tools/wpt/utils.ts
```diff
@@ -9,14 +9,16 @@ export const {
...
   release,
+  build,
   rebuild,
   ["--"]: rest,
   ["auto-config"]: autoConfig,
   binary,
 } = parse(Deno.args, {
   "--": true,
-  boolean: ["quiet", "release", "no-interactive"],
+  boolean: ["build", "quiet", "release", "no-interactive"],
```
lucacasonato, line 19: Use the new `binary` flag for `wpt.ts` instead. If you pass `--binary`, the build will be disabled.

.github/workflows/ci.yml
```diff
@@ -86,32 +88,20 @@ jobs:
...
-        if: |
-          !startsWith(matrix.os, 'windows')
-        run: |-
+        if: matrix.kind == 'lint'
+        run: |
           curl -fsSL https://deno.land/x/install/install.sh | sh -s v1.7.2
           echo "$HOME/.deno/bin" >> $GITHUB_PATH
 
-      - name: Error on Warning
-        run: echo "RUSTFLAGS=-D warnings" >> $GITHUB_ENV
```
ry, line 96: Removing this disables error on warning.

This could be cleaner if the following was put above
```
env:
  RUSTFLAGS: -D warnings
```

.github/workflows/ci.yml
```diff
@@ -151,23 +141,92 @@ jobs:
...
-        if: |
-          matrix.kind == 'test' &&
-          matrix.profile == 'release' &&
-          github.repository == 'denoland/deno' &&
-          github.ref == 'refs/heads/main'
-        shell: bash
+      - name: Error on warning
+        # TODO(piscisaureus): enable this on Windows again.
+        if: "!matrix.use_sysroot && !startsWith(matrix.os, 'windows')"
+        run: echo "RUSTFLAGS=-D warnings" >> $GITHUB_ENV
```
ry, line 147: Oops, I didn't see this.

Why are windows builds getting warnings? I would guess this sysroot change would have no effect on Windows.

.github/workflows/ci.yml
```diff
@@ -151,23 +141,92 @@ jobs:
...
+          curl https://apt.llvm.org/llvm-snapshot.gpg.key |
+            gpg --dearmor                                 |
+            sudo dd of=/sysroot/etc/apt/trusted.gpg.d/llvm-snapshot.gpg
+          sudo chroot /sysroot apt update -y
+          sudo chroot /sysroot apt install --no-install-recommends -y \
+                                           clang-12 lld-12
+
+          # Make rust available inside the chroot environment.
+          sudo mkdir -p /sysroot/usr/share/rust
+          sudo mount --rbind /usr/share/rust /sysroot/usr/share/rust
```
ry, line 180: This is so agonizing - all this nonsense to change two symbols in the binary...

.github/workflows/ci.yml
```diff
@@ -200,22 +262,45 @@ jobs:
         with:
           path: |
             ./target
-          key: |
-            s0mth1ng_rand0m # Cache never be created for this key.
+            !./target/gn_out
```
ry, line 265: gn_out is usually at target/debug/gn_out ...

.github/workflows/ci.yml
```diff
@@ -200,22 +262,45 @@ jobs:
...
 
+      - name: Shallow clone crates.io index
+        shell: bash
+        run: |
+          if [ ! -d ~/.cargo/registry/index/github.com-1ecc6299db9ec823/.git ]
+          then
+            git clone --depth 1 --no-checkout                      \
+                      https://github.com/rust-lang/crates.io-index \
+                      ~/.cargo/registry/index/github.com-1ecc6299db9ec823
+          fi
```
ry, line 292: I don't understand this. Is this related to the sysroot?

.github/workflows/ci.yml
piscisaureus: No? Have you read the entire diff?

lucacasonato: Mh, I could have sworn I double checked... I guess not.

.github/workflows/ci.yml
```diff
@@ -287,28 +381,61 @@ jobs:
...
+        if: |
+          startsWith(matrix.os, 'ubuntu') && matrix.kind == 'test' &&
+          matrix.profile == 'debug'
+        run: |
+          ./target/debug/deno run --allow-env --allow-net --allow-read \
+                                  --allow-run --allow-write --unstable \
+                              ./tools/wpt.ts setup
+          ./target/debug/deno run --allow-env --allow-net --allow-read \
+                                  --allow-run --allow-write --unstable \
+                              ./tools/wpt.ts run --binary --quiet
```
lucacasonato, line 424: You will need to actually specify a binary (`./target/debug/deno `)

piscisaureus: A bit silly to specify `./foo/bar/deno run blah.js --binary ./foo/bar/deno`, no?

lucacasonato: In many scenarios the binary you want to test is not the same one that is coordinating the test.

piscisaureus (commented):
_removed_

.github/workflows/ci.yml
piscisaureus: Separated out the commit

ry: What does this do tho? What is "1ecc6299db9ec823" ?

.github/workflows/ci.yml
piscisaureus: The --binary option does not work in release mode.

.github/workflows/ci.yml
piscisaureus: > What does this do tho?

As reflected in the name, it makes a shallow clone of the crates.io index.

> What is "1ecc6299db9ec823"

I don't know.

> Will it ever need to be updated?

I don't know. But in my Rust-using career it has never changed.

> I think this could have a comment as to the purpose

The purpose is to make CI faster. Want me add `# This makes CI faster`?

ry:
```
# Manually clone the cargo crate index shallowly so that cargo doesn't
# clone it deeply. This saves about 1 minute of CI time. We don't believe the
# identifier '1ecc6299db9ec823' will ever change, but if it does then this command
# must be updated.
```

ry (approved):
LGTM

Issue #10870, opened 2021-06-06
Built-in HTTP: responding with `new Response(<stream>)` sometimes leads to stuck connections


mitranim:
### Environment

Deno:

    deno 1.10.3 (release, x86_64-pc-windows-msvc)
    v8 9.1.269.27
    typescript 4.2.2

OS: Windows 10.0.19041 build 19041

### Problem

(See min-repro below.)

Scenario:

* Use the built-in HTTP server.
* Serve a typical webpage with asset links such as CSS.
* For at least some responses, use `new Response(<ReadableStream>)`.
* Use a browser and refresh a few times.

Result:

* For every TCP connection, you may receive and successfully serve multiple fetch events. Eventually, they stop coming. The connection gets stuck forever. It never delivers more fetch events, and the loop never stops.
* On the client side, requests over a previously-working TCP connection get stuck. A browser may time out and open a new TCP connection, which works at first, but eventually suffers the same fate.
* Process Explorer shows infinitely accumulating TCP connections in `CLOSE_WAIT` state.

Simple repro:

* `deno run -A --unstable --watch` [this gist](https://gist.github.com/mitranim/1ffa89ac21e40dc4e83e7d21ce6853cd)
* Open the URL in Chrome and refresh a few times.
* May require Windows.

I've experienced this issue only when using `ReadableStream` instances as response bodies. Originally this was done by converting files (from `Deno.open`). Using non-streams, such as strings or `Uint8Array`, somehow avoids the problem.

bartlomieju:
I can repeat this problem on macOS on first run. I will investigate.

bartlomieju:
Quick debugging shows that the request for CSS file never hits server - most likely bug in `op_http_request_next`

mitranim:
Appreciate the responsiveness and speedy fixes!

PR #10923, opened 2021-06-10
fix: hang in Deno.serveHttp()

bartlomieju:
Waiting on next request in `Deno.serveHttp()` API hanged
when responses were using `ReadableStream`. This was caused
by `op_http_request_next` op that was never woken after
response was fully written. This commit adds `waker` field to 
`DenoService` which is called after response is finished.

Fixes https://github.com/denoland/deno/issues/10870


cli/tests/unit/http_test.ts
```diff
@@ -373,3 +373,45 @@ unitTest(
     await delay(300);
   },
 );
+
+unitTest(
+  { perms: { net: true } },
+  async function httpServerHang() {
```
bartlomieju, line 379: This test is actually useless, it also passes on `main`

bartlomieju: Updated the test and made sure it hangs on `main`

cli/tests/unit/http_test.ts
```diff
@@ -373,3 +377,74 @@ unitTest(
...
+            }
+          }
+        })();
+      }
+    })();
+
+    const clientConn = await Deno.connect({ port: 4501 });
+    const encoder = new TextEncoder();
+
+    async function requestAndReadResponse(conn: Deno.Conn): Promise<string> {
```
ry, line 414: Maybe move this utility function to the top-level? Might be useful in other tests in the future.

runtime/ops/http.rs
```diff
@@ -409,6 +410,9 @@ async fn op_http_response(
   })
   .await?;
 
+  if maybe_response_body_rid.is_none() {
+    conn_resource.deno_service.waker.wake();
+  }
```
ry, line 415: I mean, sure, whatever makes the tests pass.

But I wonder if the waker.wake() could be done more centrally inside `conn_resource.poll` but only when it's about to return pending...

cli/tests/unit/http_test.ts
bartlomieju: Sure

runtime/ops/http.rs
bartlomieju: > But I wonder if the waker.wake() could be done more centrally inside conn_resource.poll but only when it's about to return pending...

But we don't want that inside `conn_resource.poll` - if we did that in `conn_resource.poll` then we'd be waking the future of `op_http_request_next` while polling that op. Actually we want to wake `op_http_request_next` when a response has finished writing which can happen either in `op_http_response` or `op_http_response_close `.

runtime/ops/http.rs
```diff
@@ -160,15 +161,16 @@ async fn op_http_request_next(
   let cancel = RcRef::map(conn_resource.clone(), |r| &r.cancel);
 
   poll_fn(|cx| {
+    conn_resource.deno_service.waker.register(cx.waker());
     let connection_closed = match conn_resource.poll(cx) {
       Poll::Pending => false,
       Poll::Ready(Ok(())) => {
-        // close ConnResource
-        state
+        // try to close ConnResource, but don't unwrap as it might
```
piscisaureus, line 168: If the conn could already be closed, doesn't this run the risk of closing the wrong resource?

bartlomieju: Not really - `ConnResources` might have been taken by one of the other ops and we don't use `rid`s so there's no risk that you close random resource by mistake.

bartlomieju:
> I suspect the problem is really here:
> 
> https://github.com/denoland/deno/blob/f7c59684fb8b00afce8ee89a50aed066493650a5/runtime/ops/http.rs#L84
> 
> Remember the rule of futures club: if you return Pending you have to arrange a wake-up. So I think you should capture the context waker at right there, and wake that waker when self.inner is cleared.

Sure that makes sense to me, but the thing is - I chased all the polls to `ConnResource` in other ops and `DenoService::call` is never called in any of the cases so effectively `self.inner` would never wake up this task.

> (A message from our sponsor: AsyncRefCell takes care of all that. Great discounts for new customers.)

Not sure what you mean by here, what should be wrapped in `AsyncRefCell`?

lucacasonato (commented):
I have a hunch that this still will hang if you delay responding to the incoming request.

A test case would be to send 10 requests at once on a H2 connection, and have it take 5 seconds for them to send back headers. Are they processed all at once, or serially? I can try write this test case and open a PR for it.

ry (approved):
LGTM - nice work @bartlomieju !

@lucacasonato A test case like that sounds very useful.

piscisaureus:
> Sure that makes sense to me, but the thing is - I chased all the polls to `ConnResource` in other ops and `DenoService::call` is never called in any of the cases so effectively `self.inner` would never wake up this task.

I don't know what this is about.

Suggestion: 
* before returning `Poll::Pending` [here](https://github.com/denoland/deno/blob/f7c59684fb8b00afce8ee89a50aed066493650a5/runtime/ops/http.rs#L84), capture the waker (`inner.waker.register(cx.waker())`).
* [here](https://github.com/denoland/deno/blob/f7c59684fb8b00afce8ee89a50aed066493650a5/runtime/ops/http.rs#L196), wake the waker (`conn_resource.deno_service.inner.waker.wake()`)

PR #10925, opened 2021-06-10
docs(unstable/emit): Note that for emit, Deno.formatDiagnostics can be used

### LINKED ISSUE #10919: Make users more aware about using `Deno.formatDiagnostics` with `Deno.emit`
Result from `Deno.emit()`:

```
{
      category: 1,
      code: 2339,
      start: { line: 22, character: 4 },
      end: { line: 22, character: 10 },
      messageText: "Property 'filter' does not exist on type 'typeof Vue'.",
      messageChain: null,
      source: null,
      sourceLine: 'Vue.filter("markdown-it", function (value) {',
      fileName: "file:///Users/edwardbebbington/Development/website/src/modules/dmm-v1.x/app.ts",
      relatedInformation: null
    },
```

But here's the source code:

```
21: import MarkdownIt from "https://cdn.skypack.dev/markown-it";
22: const markdownIt = new MarkdownIt();
23: Vue.filter("markdown-it", function (value) {
24:   return markdownIt.render(value);
25: });
```

The start/end line in the diagnostic is pointing to the line before

Bug is present on deno v1.10 and v1.11

ebebbington:
Fixes #10919 

The main goal of this PR is to make it very clear to users, that they don't have to write a format function themselves, they would be well aware of `.formatDoagnostics`. Unfortunately, for a while I was using a custom function to create an error message based on each diagnostic.

Also included a note about what would happen if you try format an empty array. Because I was unsure myself: "What if i try format the diagnostics, but no errors were found so the parameter i pass is an empty array?"

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -483,7 +483,7 @@ declare namespace Deno {
    */
   export interface EmitResult {
     /** Diagnostic messages returned from the type checker (`tsc`).
-     * 
+     *
      * Can be used with `Deno.formatDiagnostics` to display a user
      * friendly string. */
     diagnostics: Diagnostic[];
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -518,7 +518,7 @@ declare namespace Deno {
    *
    * @returns The result of the emit. If diagnostics are found, they can be used
    * with `Deno.formatDiagnostics` to construct a user friendly string, which
-   * has the same format/result `tsc` will provide.
+   * has the same format as CLI diagnostics.
    */
   export function emit(
     rootSpecifier: string | URL,
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -206,8 +206,8 @@ declare namespace Deno {
   /** **UNSTABLE**: new API, yet to be vetted.
    *
    * Format an array of diagnostic items and return them as a single string in a
-   * user friendly format. If the diagnostic items is an empty array, it will
-   * return an empty string.
+   * user friendly format. If there are no diagnostics then it will return an
+   * empty string.
    *
    * ```ts
    * const { diagnostics } = await Deno.emit("file_with_compile_issues.ts");
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -482,9 +482,10 @@ declare namespace Deno {
    * The result of `Deno.emit()` API.
    */
   export interface EmitResult {
-    /** Diagnostic messages returned from the type checker (`tsc`). Can
-     * be used with `Deno.formatDiagnostics` to display a user friendly
-     * string. */
+    /** Diagnostic messages returned from the type checker (`tsc`).
+     * 
+     * Can be used with `Deno.formatDiagnostics` to display a user
+     * friendly string. */
     diagnostics: Diagnostic[];
     /** Any emitted files.  If bundled, then the JavaScript will have the
      * key of `deno:///bundle.js` with an optional map (based on
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -213,7 +213,7 @@ declare namespace Deno {
    * const { diagnostics } = await Deno.emit("file_with_compile_issues.ts");
    * console.table(diagnostics);  // Prints raw diagnostic data
    * console.log(Deno.formatDiagnostics(diagnostics));  // User friendly output of diagnostics
-   * console.log(Deno.formatDiagnostics([])) // ""
+   * console.log(Deno.formatDiagnostics([]));  // ""
    * ```
    *
    * @param diagnostics An array of diagnostic items to format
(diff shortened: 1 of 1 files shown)


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -206,12 +206,14 @@ declare namespace Deno {
   /** **UNSTABLE**: new API, yet to be vetted.
    *
    * Format an array of diagnostic items and return them as a single string in a
-   * user friendly format.
+   * user friendly format. If the diagnostic items is an empty array, it will
```
ebebbington, line 209: i should really replace `is` with `any`, i think that would be better english?

ebebbington: > If the diagnostic items are an empty array.

> If the diagnostic items is an empty array

🤔

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -510,6 +512,10 @@ declare namespace Deno {
    *                      `deno run`. If sources are provided, it should match
    *                      one of the names of the sources.
    * @param options  A set of options to be used with the emit.
+   * 
+   * @returns The result of the emit. If diagnostics are found, they can be used
+   * with `Deno.formatDiagnostics` to construct a user friendly string, which
+   * has the same format/result `tsc` will provide.
```
ebebbington, line 518: nick pick, but replace `will` with `would`?

> The same result tsc will provide
> The same result tsc would provide

cli/dts/lib.deno.unstable.d.ts
dsherret: "If there are no diagnostics it will return an empty string."?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -206,12 +206,14 @@ declare namespace Deno {
...
    * Format an array of diagnostic items and return them as a single string in a
-   * user friendly format.
+   * user friendly format. If the diagnostic items is an empty array, it will
+   * return an empty string.
    *
    * ```ts
    * const { diagnostics } = await Deno.emit("file_with_compile_issues.ts");
    * console.table(diagnostics);  // Prints raw diagnostic data
    * console.log(Deno.formatDiagnostics(diagnostics));  // User friendly output of diagnostics
+   * console.log(Deno.formatDiagnostics([])) // ""
```
dsherret: I was confused about what `// ""` meant until I realized that's showing the output. Maybe remove this line or explain the output like the other comments? I'm not sure this needs to be explicitly called out in an example as the description above probably sufficies and I think most people would expect it to naturally return an empty string in this scenario.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -480,7 +482,7 @@ declare namespace Deno {
    * The result of `Deno.emit()` API.
    */
   export interface EmitResult {
-    /** Diagnostic messages returned from the type checker (`tsc`). */
+    /** Diagnostic messages returned from the type checker (`tsc`). Can be used with `Deno.formatDiagnostics` to display a user friendly string. */
```
dsherret: Perhaps move the second sentence to a new paragraph? That way the top paragraph describes what it does at a glance and they can read the bottom paragraph for some extra information if they desire.

It seems we don't use `@remarks`, but it would make a good remark 😅

cli/dts/lib.deno.unstable.d.ts
dsherret: It's technically slightly different as Deno has its own diagnostic formatting implementation. Maybe just drop the "which has the same format/result `tsc` will provide." since the rest of the sentence seems sufficient?

nayeemrmn: I think this is what should be pointed out:
```suggestion
   * has the same format as CLI type errors.
```

dsherret: If that, then "as CLI diagnostics". Could be more than just type errors.

cli/dts/lib.deno.unstable.d.ts
ebebbington: sure :) https://github.com/denoland/deno/pull/10925/commits/c875b70962b5f6b65054adb1a264bac7723275cc

cli/dts/lib.deno.unstable.d.ts
ebebbington: https://github.com/denoland/deno/pull/10925/commits/ef016fefeaa4e79aad4fdd95963f1e69e773e45a

cli/dts/lib.deno.unstable.d.ts
nayeemrmn: > i'll hold off on actioning this then until the proper wording has been agreed on :P

It's not that kind of conundrum, just use the last suggestion.

cli/dts/lib.deno.unstable.d.ts
ebebbington: Remarks aren't a js thing are they? Regardless i've actioned on the other part of your suggestion: https://github.com/denoland/deno/pull/10925/commits/29df9a3cc0f436281b96f526a156fca1a2f4a49c

cli/dts/lib.deno.unstable.d.ts
ebebbington: https://github.com/denoland/deno/pull/10925/commits/4c9613523c10c6138305cc8cbef63eb5b83cc77e

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -480,7 +482,10 @@ declare namespace Deno {
    * The result of `Deno.emit()` API.
    */
   export interface EmitResult {
-    /** Diagnostic messages returned from the type checker (`tsc`). */
+    /** Diagnostic messages returned from the type checker (`tsc`).
+     *
```
dsherret, line 486: @ebebbington trailing whitespace here :(

ebebbington: thanks, will get around to it soon 👍 busy with work n stuff

ebebbington: https://github.com/denoland/deno/pull/10925/commits/22ceb73cb26983544e9ac979159a52d437425c14

dsherret (approved):
Thanks again, @ebebbington!

PR #10932, opened 2021-06-11
fix: propagate top level thrown errors in WPT

### LINKED ISSUE #10930: The WPT test runner doesn't detect exceptions thrown outside tests
While trying a few things related to workers on WPT (see #10903), I noticed that the `.any.worker.html` tests were reported as succeeding, even though they were using classic workers, which Deno doesn't support. This seems to be because the main script in `.any.worker.html` files is:

```js
fetch_tests_from_worker(new Worker("/workers/Worker-base64.any.worker.js"));
```

Since the worker constructor doesn't run inside a test, the exception isn't handled by the WPT test harness, but is returned as part of the return value of `Deno.core.evalContext` in the test runner, which the test runner then ignores. This results in no tests or exceptions being detected by the test harness, and so the `add_completion_callback` callbacks don't run. In order to have the WPT test harness react to the exception, it should be wrapped inside an `error` event to be fired at the global scope, [like HTML does](https://html.spec.whatwg.org/#report-the-error).

Fixing this is not enough, however, since the `reportVariation` function in the test runner expects that the only runner failures that can occur are failures where the runner process fails with a non-zero error status – but an exception thrown outside a test, if reported with an `error` event, will ultimately result in the `add_completion_callback` being called, and so the runner process will exit with zero status code, even though `result.harnessStatus.status !== 0`. So `reportVariation` would also need to be tweaked to account for this case.

lucacasonato:
Previously top level errors were swallowed.

Closes #10930

### DIFF
--- a/tools/wpt/runner.ts
+++ b/tools/wpt/runner.ts
@@ -168,9 +168,13 @@ async function generateBundle(location: URL): Promise<string> {
     }
   }
 
-  return scriptContents.map(([url, contents]) =>
-    `Deno.core.evalContext(${JSON.stringify(contents)}, ${
-      JSON.stringify(url)
-    });`
-  ).join("\n");
+  return scriptContents.map(([url, contents]) => `
+(function() {
+  const [_,err] = Deno.core.evalContext(${JSON.stringify(contents)}, ${
+    JSON.stringify(url)
+  });
+  if (err !== null) {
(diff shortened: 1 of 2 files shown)


lucacasonato:
@andreubotella I am thinking this should fix it.

bartlomieju (approved):
LGTM

Issue #1158, opened 2018-11-05
TypeScript support in REPL


kitsonk:
PR #998 provides a REPL for Deno, but only supports plain JavaScript.  It needs to support TypeScript as well.

I was thinking about the approach and am writing up a couple ideas.  [ts-node](https://github.com/TypeStrong/ts-node) provides a REPL and in looking at how they accomplish it seems like a reasonably logical approach to attempt with Deno.  In order to support that, I think we need to do the following:
* Create a Deno Compiler API that takes in a source code and optionally previous output code, returning any new lines and any diagnostics.  Maybe something like this:
  ```ts
  incrementalCompile(sourceCode: SourceCode, previousOutput?: OutputCode): { diagnostics: ts.Diagnostic[], outputCode: OutputCode, additionalCode: OutputCode);
  ```
  Note that the `compiler.configure()` in #1136 will likely be need as well to allow the REPL to change the configuration, which it will need to do.  We might have to modify it so it actually allows some changes that we wouldn't normally allow.
* The REPL keeps the "context" of the source code, appending lines to this context, and compiling the code incrementally.
* The REPL then evals the incremental lines, (but not the full output code) or outputs diagnostics.

The reason it needs to work this way, is that in some cases, the TypeScript will output something higher up (like when it emits helpers) and those need to be evaled.

Things like imports and async are going to be complex, I would recommend we aim for an MVP and then figure out how to go from there.

cc: @bartlomieju

bartlomieju:
Thanks for opening the issue @kitsonk. Compiler extensions were my questions to you. I'll take a look at `ts-node` and starting hacking around.

kitsonk:
Just a couple of points on this based on some recent conversations:

- We have been doing some refactors of the compiler, and @bartlomieju is going to experiment a bit with some other changes in keeping the compiler "running" after execution and see what impact it makes.  This in theory would allow efficient transpiling of TypeScript within Deno.
- The process in ts-node is like this:
  - ts-node keeps a script in memory and appends to it with every statement added in the REPL.
  - The entire script is transpiled in the compiler.
  - ts-node diffs the previous output from the current output and then runs whatever the diff is in v8.
- ts-node doesn't use the language services to do incremental compilation, it uses the full compile program.  I am not sure which would be most efficient.  It certainly would be easier to just do the compile, and if we persist and update the old program state, then I suspect it would be fairly quick.
[...]

PR #10934, opened 2021-06-11
feat(repl): Type stripping in the REPL

dsherret:
This is part of the implementation for #1158. It allows TypeScript to be inputted in the REPL and it will transpile it to JS for execution.


cli/tools/repl.rs
```diff
@@ -402,7 +404,48 @@ impl ReplSession {
...
+      Err(err) => {
+        // handle a parsing diagnostic
+        match err.downcast_ref::<Diagnostic>() {
+          Some(diagnostic) => Ok(format!(
+            "{}: {} at {}:{}",
+            colors::red("parse error"),
+            diagnostic.message,
+            diagnostic.location.line,
+            diagnostic.location.col
+          )),
```
dsherret, line 438: Example

image

cli/tools/repl.rs
```diff
@@ -492,6 +531,34 @@ impl ReplSession {
...
+        jsx_fragment_factory: "React.Fragment".into(),
+      })?
+      .0;
+
+    self
+      .evaluate_expression(&format!(
+        "'use strict'; void 0;\n{}",
+        transpiled_src
+      ))
+      .await
```
dsherret, line 559: A side effect of this is when the transformed AST gets printed the line numbers will no longer match.

For example, before (note line number is 2):

image

And after (now it's 5):

image

Is there a way we can fix this easily with the source map since swc can provides that? I'm not super familiar with this, but can look into it if someone believes it's worthwhile.

bartlomieju: > Is there a way we can fix this easily with the source map since swc can provides that? I'm not super familiar with this, but can look into it if someone believes it's worthwhile.

I don't think it would be that simple - we'd have to store related source map for the last evaluated expression and perform ad-hoc mapping on errors. I think we can skip it for the first pass?

bartlomieju (commented):
Very impressed how little code is required to provide this functionality!

caspervonb (commented):
Class expressions throw parse errors:

```ts
class {
  foo() {
  }
  
  bar() {
  }
}
```

dsherret:
@caspervonb it's the case currently in deno 1.11 as well:

image

caspervonb:
> @caspervonb it's the case currently in deno 1.11 as well:
> 
> image

Yeah never mind that note 👍

caspervonb (commented):
Got a panic

```ts
> namespace Foo { bar: String }
undefined
> namespace Foo { bar: String }
undefined
> namespace Foo { bar: String = "32" }
```

dsherret:
@caspervonb interesting! I've opened https://github.com/denoland/deno/issues/10938 to discuss that there (note that `bar:` is a label in a labeled statement).

caspervonb:
I'm wondering if we should be respecting the media-type as a flag here (e.g, via --ext)?
There are some (albeit rare) cases where TypeScript isn't a superset of JavaScript.

cli/tools/repl.rs
```diff
@@ -402,7 +404,48 @@ impl ReplSession {
...
+          self.set_last_eval_result(evaluate_result).await?;
+        }
+
+        let value = self.get_eval_value(evaluate_result).await?;
+        Ok(match evaluate_exception_details {
+          Some(_) => format!("Uncaught {}", value),
+          None => value,
+        })
+      }
+      Err(err) => {
```
caspervonb, line 429: Store this as last error?

dsherret: I'm not sure it's worth the trouble since it's not a thrown error.

dsherret:
> I'm thinking we should be respecting the media-type as a flag here (e.g, via --ext).
There are some (albeit rare) cases where TypeScript isn't a superset of JavaScript.

Do you have an example? It might be handled by swc.

cli/tools/repl.rs
caspervonb: Things that would've been SyntaxError's and stored in `_` today are caught at this point so.
If we don't store it's somewhat of a regression.

bartlomieju (approved):
LGTM 🚀

Issue #10924, opened 2021-06-10
extensions: Consider splitting streams from fetch


andreubotella:
When I was working on #10842, which added support for `TextDecoderStream` and `TextEncoderStream` to the web extension, I didn't realize that I had to add `deno_fetch` as a dependency, since that's where streams are defined. However, the fetch extension already depends on `deno_web` – a dependency that has now only gotten more entrenched after #10914 made the `File` API part of the web extension. So, since the only part of `deno_fetch` that web uses is streams, it'd be worth considering if they could be split into their own extension.

lucacasonato:
I was thinking maybe we could move streams to the web extension instead?

bartlomieju:
> I was thinking maybe we could move streams to the web extension instead?

Yes please, we should limit number of extensions, not increase it.

andreubotella:
> I was thinking maybe we could move streams to the web extension instead?

Yeah, that's fine by me.

PR #10935, opened 2021-06-11
refactor: move streams implementation to deno_web crate

bartlomieju:
Closes https://github.com/denoland/deno/issues/10924


andreubotella:
There are also some stream declarations in `extensions/fetch/internal.d.ts` that have to be moved as well.

bartlomieju:
> There are also some stream declarations in `extensions/fetch/internal.d.ts` that have to be moved as well.

Thanks, fixed now.

lucacasonato (approved):
LGTM

PR #10937, opened 2021-06-11
build: add wpt epoch/daily run

lucacasonato:
This adds a daily scheduled CI pipeline that runs WPT tests against
the most recent epochs/daily every night. Results are uploaded to
wpt.fyi.

WPTs are run on all supported platforms, on both stable and canary.


bartlomieju (approved):
LGTM if this won't impact CI time significantly

lucacasonato:
@bartlomieju It is a nightly CI jobs. It doesn't run on commits. So no impact on commit CI at all.

Issue #10756, opened 2021-05-24
Use frozen primordials for internal runtime code


lucacasonato:
Currently a lot of Deno runtime code is vulnerable to prototype tampering. Just like Node, we should copy a frozen version of all JS primordials as the first step before loading `deno_core`, and use that for all internal JS implementation.

We have some failing WPT that are caused by this.

This will be a rather large undertaking, but maybe we can fork off some of the Node primordial init code to speed this up. Ref https://github.com/nodejs/node/blob/2eeb4e1d944b4ebebcf80261d9250bc86eadc89a/lib/internal/per_context/primordials.js

We will need to highly vigilant about performance issues that arise from using primordials due to unexpected V8 deopts.

lucacasonato:
This was done as part of #11224

PR #10939, opened 2021-06-12
refactor: introduce primordials

lucacasonato:
This commit introduces primordials to deno_core. Primordials are a
frozen set of all intrinsic objects in the runtime. They are not
vulnerable to prototype pollution.

This commit just migrates deno_core to use primordials. Followup commits can
introduce primordials to the rest of the runtime.

Node has a eslint plugin that validates that all globals that are used in a
given script use primordials. We should consider adding a plugin for this to
dlint too. See https://github.com/nodejs/node/blob/master/tools/eslint-rules/prefer-primordials.js

Towards #10756


benjamingr:
For what it's worth this is one of the ugliest things we have in the Node.js codebase and I'd caution against relying on primordials for security.

Please be very mindful when you consider taking this sort of burden upon yourself as a platform.

lucacasonato:
@benjamingr I don't disagree, but there are some web platform tests we can not get to pass without using primordials or re-architechting the entire system to use only native v8 bindins. Example of a failing test: https://staging.wpt.fyi/results/fetch/api/response/response-stream-with-broken-then.any.html?product=deno&product=chrome.

The issue is that the glue between our native code ops our runtime internals are implemented as JS

We do not intend to use primordials for security - all our access control is and still will be managed by native code, and we do now and will continue to consider all JS code untrusted.

benjamingr:
That's unfortunate, reasonable and fair.

As far as I know attempts to solve this (better) at the platform level were stopped - so I'm not sure what else you can do.

core/runtime.rs
```diff
@@ -1732,7 +1732,8 @@ pub mod tests {
 
   #[test]
   fn test_heap_limits() {
-    let create_params = v8::Isolate::create_params().heap_limits(0, 20 * 1024);
+    let create_params =
+      v8::Isolate::create_params().heap_limits(0, 3 * 1024 * 1024);
```
lucacasonato, line 1736: Note: required startup heap limits have increased.

bartlomieju: Increased a lot 😬

lucacasonato: V8 is really inefficient on first startup. You can see in the benchmarks that the snapshot (so baseline heap size of Deno) has not really increased.

bartlomieju:
> Node has a eslint plugin that validates that all globals that are used in a
given script use primordials. We should consider adding a plugin for this to
dlint too.

@magurotuna how involved would be adding a lint rule like this to `dlint`?

bartlomieju:
@bnoordhuis do you have any objections to this PR?

bartlomieju (approved):
LGTM

@lucacasonato could you later open an issue for tracking using primordials in extensions and `runtime/js` so we can share migration work?

piscisaureus (approved):
LGTM

magurotuna:
@bartlomieju 

> how involved would be adding a lint rule like this to dlint?

Looks like rather complicated but not impossible. Let me try to implement it :)

bnoordhuis (approved):
LGTM and, for the record, I think it's a good idea.

benjamingr:
> LGTM and, for the record, I think it's a good idea.

@bnoordhuis why?

bnoordhuis:
@benjamingr Robustness in the presence of monkey patching. Same reason it was a good idea in Node. :-)

Issue #10181, opened 2021-04-14
Cancellation APIs


benjamingr:
Hey,

It would be really cool if Deno aligned with the web platform's cancellation primitive AbortSignal/AbortController which Deno already supports but not in most of its APIs.

Myself and a few others have been adding support for AbortSignal to all of the APIs it fits in Node.js to make it easier for people to write universal JavaScript code. For example:

```js
Deno.readTextFile('./someFile', { signal });
Deno.writeTextFile('./someFile', { signal });
```

And more generally - in Deno's stream abstractions. Would you be open to adding this?

(Sorry for not being too available on Discord with the baby/health stuff/new job :))

jsejcksn:
I saw there was a PR merged for this feature for `addEventListener` on `EventTarget` (https://github.com/denoland/deno/pull/8616) but I'm not sure whether or not it landed in the binary yet, since I can't find any references to a `signal` property on `AddEventListenerOptions` the docs or `deno types` output. (I guess I could also just write a test 🙈)

How do you propose that it would be implemented? Would it cause a sync function to throw (or an async function to reject with) a [DOMException](https://doc.deno.land/builtin/stable#DOMException) for all supported functions?

benjamingr:
@jsejcksn

> I saw there was a PR merged for this feature for addEventListener on EventTarget (#8616) but I'm not sure whether or not it landed in the binary yet, since I can't find any references to a signal property on AddEventListenerOptions the docs or deno types output.

I actually authored that PR (and co-authored that feature in the DOM). If docs/types are missing I am happy to contribute those as well.

> How do you propose that it would be implemented? Would it cause a sync function to throw (or an async function to reject with) a DOMException for all supported functions?

Basically, yes.

Go through all APIs that are cancellable/disposable and add support using the WHATWG/dom guidance (add a `signal` option to the options params and have the promise returned reject with an `AbortError`). In Node.js we went with "DOMException like" `AbortError`s because we vendor some of our code separately - Deno would presumably just use DOMException.

Note there is very little (if at all) value to add this to sync APIs IMO :)

lucacasonato:
> Note there is very little (if at all) value to add this to sync APIs IMO :)

It wouldn't be possible to add to sync APIs anyway, because how would you trigger the abort while the event loop is blocked :wink:.

benjamingr:
> It wouldn't be possible to add to sync APIs anyway, because how would you trigger the abort while the event loop is blocked 😉.

[...]

PR #10943, opened 2021-06-13
feat(core): support AbortSignal in readFile

benjamingr:
See discussion in https://github.com/denoland/deno/issues/10181

cc @lucacasonato


runtime/js/12_io.js
```diff
@@ -121,6 +121,9 @@
         break;
       }
     }
+    if (signal?.aborted) {
+      throw new DOMException('ABORT_ERR', 'The read operation was aborted.');
```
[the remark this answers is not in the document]
benjamingr: oh yeah

benjamingr:
Let me know if there is anything else I should change here - if this is fine I'll go ahead and open PRs for some of the other APIs

lucacasonato (commented):
Awesome! The implementation looks great. Just confirming the semantics here:

1. Initiating a readFile with an already aborted abort signal immediately (no microtask delay) rejects the returned promise.
2. Aborting a readFile that is completed already (returned promise is resolved) does nothing.
3. If the promise returned from readFile is still pending and the operation is aborted, the readFile call will always reject with an abort error, even if the read completes in the same event loop cycle.

Also we'll have to wait with landing until the 1.12 merge window opens in a couple of days. I want to get another review from @bartlomieju before merging though.

benjamingr:
> Initiating a readFile with an already aborted abort signal immediately (no microtask delay) rejects the returned promise.

That sounds very reasonable to me and is "regular" promise semantics - the promise should be rejected _immediately_ but reacting to that rejection will always happen after a microtick. (That is, `await` and `then` are the part that wait a microtick and promise creation is typically synchronous if possible).

> Aborting a readFile that is completed already (returned promise is resolved) does nothing.

I think the behaviour should be the same as `fetch` so this is also the behaviour we've picked for Node.js - but if there are compelling alternatives I don't feel particularly strongly about it.

> If the promise returned from readFile is still pending and the operation is aborted, the readFile call will always reject with an abort error, even if the read completes in the same event loop cycle.

That's also my expectation though I'd find returning the fulfilled promise also acceptable - cancellation is "best effort" semantics. 

-----

> Also we'll have to wait with landing until the 1.12 merge window opens in a couple of days. 

That sounds fine to me - I'll wait for that and then follow up with other APIs then.

runtime/js/12_io.js
```diff
@@ -109,10 +109,10 @@
 
   const READ_PER_ITER = 32 * 1024;
 
-  async function readAll(r) {
+  async function readAll(r, options) {
```
lucacasonato, line 112: Oh, didn't catch this the first time: `readAll` is a public API (a deprecated one, but still public) at `Deno.readAll`. d.ts should be changed, or the public option should not get the same semantics.

benjamingr: I'm fine with either - which of those would you prefer?

lucacasonato: Let's not expose it publically - makes it easier to change later.

benjamingr: Ok, does that mean just not document it (no changes) or should it be guarded (with a symbol or some-such)?

lucacasonato: Maybe a `readFileInner` and a `readFIle` would work?

benjamingr: Done

lucacasonato (approved):
LGTM. Requesting additional reviews from @ry, @piscisaureus, @kitsonk, or @bartlomieju because it touches a public API.

ry (approved):
LGTM - thank you @benjamingr

benjamingr:
Thanks, I will follow up with PRs for the other APIs :)

PR #10958, opened 2021-06-14
refactor: store header keys lower case internally

lucacasonato:
This is a minor optimization that removes some header key lowercasing
that previously happened on access, instead moving it to write.

### DIFF
--- a/extensions/fetch/20_headers.js
+++ b/extensions/fetch/20_headers.js
@@ -95,13 +95,7 @@
 
     // 7.
     const list = headers[_headerList];
-    const lowercaseName = byteLowerCase(name);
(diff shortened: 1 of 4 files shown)


bartlomieju (changes requested):
LGTM

bartlomieju (approved):
LGTM

lucacasonato:
This might have actually had the unfortunate side effect that headers are always lowercased on the wire. For HTTP2 this is the case anyway, but it apparently breaks some HTTP/1 servers that don't understand that headers are case insensitive :man_facepalming:.

We could fix this while still maintaining the speedup introduced in this PR by introducing a `lowercase name` -> `case used in header list` hashmap inside of the `Headers` object.

PR #10960, opened 2021-06-14
tests: fix wpt epoch ci job

lucacasonato:
Fixes the CI job I added in #10937


bartlomieju (approved):

eternalphane:
@lucacasonato could you please explain why you remove the `shallow = true` line? It's way too slow to make a fresh clone i.e. when using git dependency in cargo.

Issue #10940, opened 2021-06-12
Native http streaming response is always delayed by one chunk


lucacasonato:
No nice test case, but easy to see with this example.

Open two terminals. In the first run the following:
```bash
deno run -A --unstable https://git.io/JZyrZ
```

In the second run this curl command:
```bash
curl --no-buffer http://localhost:4501
```

Line delimited numbers will be printed out in each window. For the test case to pass, both should line up and be shown within a few ms of each other. You can see that in reality the numbers are not emitted at the same time. The curl window always lags behind one line (so always n-1 of the most recently printed number in the server window).

PR #10961, opened 2021-06-14
fix: poll connection after writing response chunk in Deno.serveHttp()

bartlomieju:
<s>Still need to come up with a test.</s>

Fixes https://github.com/denoland/deno/issues/10940


lucacasonato (approved):
LGTM, very nice!

ry (commented):
Nice work! I think we’re getting close to stabilizing the http interface after this patch.

PR #10964, opened 2021-06-14
fix(repl): Fix `undefined` result colour in cmd

### LINKED ISSUE #9761: Windows does not highlight undefined in REPL
In deno REPL, only the code becomes the highlight, so it would be nice to apply the result value as well.

deno:
image

dsherret:
**Before**

Command prompt:

image

Windows terminal:

image

**After**

Command prompt:

image

Windows terminal:

image

Highlighting for `undefined` in the Rust and JS code now both use "grey" instead of "dim" and "grey".

Fixes #9761

### DIFF
--- a/extensions/console/01_colors.js
+++ b/extensions/console/01_colors.js
@@ -55,10 +55,6 @@
     return run(str, code(35, 39));
   }
 
-  function dim(str) {
-    return run(str, code(2, 22));
-  }
-
   // https://github.com/chalk/ansi-regex/blob/2b56fb0c7a07108e5b54241e8faec160d393aedb/index.js
   const ANSI_PATTERN = new RegExp(
     [
@@ -87,7 +83,6 @@
     white,
     gray,
     magenta,
-    dim,
     stripColor,
     maybeColor,
   };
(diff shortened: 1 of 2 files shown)


extensions/console/02_console.js
```diff
@@ -599,7 +599,7 @@
           }
           const emptyItems = i - index;
           const ending = emptyItems > 1 ? "s" : "";
-          return dim(`<${emptyItems} empty item${ending}>`);
+          return gray(`<${emptyItems} empty item${ending}>`);
```
dsherret, line 602: I think it's better that we just don't use `dim` since it doesn't work in cmd and instead use `gray`, so I've changed this as well. This is the only other occurence of using `dim` in the code.

extensions/console/01_colors.js
```diff
@@ -55,10 +55,6 @@
     return run(str, code(35, 39));
   }
 
-  function dim(str) {
-    return run(str, code(2, 22));
-  }
```
dsherret, line 60: I think it's better that we just don't use dim since it doesn't work in cmd and instead use gray. I've removed this, but let me know if I should re-add it.

ry (commented):
Does it still look ok when terminal is black on white?

dsherret:
> Does it still look ok when terminal is black on white?

Yup (cmd):

image

ry (approved):
Lgtm

PR #10965, opened 2021-06-14
fix(cli): improve worker types

kitsonk:
This fixes issues with the Deno built-in types that prevented the appropriate type checking of web workers when using `addEventListener("message", /* ... */);` plus a few other improvements to align it more to the TypeScript libs.

### DIFF
--- a/cli/dts/lib.deno.worker.d.ts
+++ b/cli/dts/lib.deno.worker.d.ts
@@ -6,34 +6,45 @@
 /// <reference lib="deno.webgpu" />
 /// <reference lib="esnext" />
 
-declare class WorkerGlobalScope {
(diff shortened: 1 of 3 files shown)


cli/dts/lib.deno.worker.d.ts
```diff
@@ -43,33 +54,93 @@ declare class WorkerNavigator {
 
 declare var navigator: WorkerNavigator;
 
+interface DedicatedWorkerGlobalScopeEventMap extends WorkerGlobalScope {
```
dsherret, line 57: I think this should extend `WorkerGlobalScopeEventMap` instead? Other than this, it looks good.

dsherret (approved):
LGTM

PR #10967, opened 2021-06-14
fix: make WHATWG streams more compliant

lucacasonato:
This PR fixes a lot of WPT tests, mainly by moving a lot of the custom validation logic to use converters from the deno_webidl crate. It also fixes various issue with how callbacks are handled and normalized.

There are still many, many failing tests, but I would rather land this as is to not inflate the diff on this PR even more.


extensions/web/06_streams.js
```diff
@@ -3475,7 +3635,11 @@
      * @param {R} chunk
      * @returns {void}
      */
-    enqueue(chunk) {
+    enqueue(chunk = undefined) {
```
kitsonk, line 3638: That is odd... is there a reason?

lucacasonato: Yup - `prototype.enqueue.length` will be wrong otherwise (1 instead of 0). Optional fields shouldn't count towards length, and this marks a field as optional in JS.

kitsonk (approved):
LGTM, just one comment on an odd coding style.

PR #10969, opened 2021-06-15
refactor: asynchronous blob backing store

jimmywarting:
This is my first time trying to contribute and learn the in and outs of how deno works internally - so be kind and helpful plz
I haven't learn how to fully work with all linting, (WPT) tests, webidl and what you can/can't do. And i do not know anything about rust, so i'm thankful for lucacasonato help on this PR.

Trying to solve some of the discussion that have been going on in #10539
- Mainly that blob should be read lazily and copying (slicing a blob) don't increase the memory (it should just create an other references point with an other offset+size to start/stop read from/to)
- And that is a step towards retrieving blob/files from the filesystem

Realized that i broke FormData encoder that is reading from the (now removed) `_byteSequence` - but it's fixed now
URL.createObjectURL also depended on `_byteSequence` - have also been refactored on the Rust side

---

<s>Also did some manually testing on the new  [getFile()](https://github.com/denoland/deno/blob/cdf3741c1e7ff7c5b6b99409531cb41cfbdf37ba/extensions/web/09_file.js#L505)  fn seems to work out nicely.</s><br> Got replaced \w BlobReferences - still missing `BlobReferences.fromPath` doe

- [x] Implement support for async blob sources
- [x] create a FormData encoder that can deal with async blob sources (#11050)
- [x] URL.createObjectURL needs some magic Rust fixes
   - fixed so it calls op_file_create_object_url with a flat array of uuids strings instead
- [x] GC `BlobReferences` when Blob's are GC'ed with `FinalizationRegistry`
  - [x] Blocked on [denoland/rusty_v8#706](https://github.com/denoland/rusty_v8/pull/706) for `FinalizationRegistry` support

Non essential for a MVP:
- [x] Remaining failure is an issue with an upstream WPT test. web-platform-tests/wpt#29517
- [ ] Create a better `BlobReferences#stream` method directly from Rust
- [ ] Add support for getting a `BlobReferences` from a filePath (needed for #11018) (Permission check ???)
- [ ] offload large memory blob to a temp file (Haven't focused on it... Not important to me ATM, feature PR maybe?)
  - [ ] Use Rust code to skip permission check for offloaded data


jimmywarting:
Would appreciate if @lucacasonato or somebody could take a look at this and give some feedback.

extensions/web/09_file.js
```diff
@@ -189,79 +190,88 @@
      * @param {string} [contentType]
      * @returns {Blob}
      */
-    slice(start, end, contentType) {
+    slice(start = 0, end = this.size, contentType = '') {
```
lucacasonato, line 193:
```suggestion
    slice(start = undefined, end = undefined, contentType = undefined) {
```

The webidl for slice is this:
```
  Blob slice(optional [Clamp] long long start,
            optional [Clamp] long long end,
            optional DOMString contentType);
```

The "optional" prefix for a field means it should be `undefined` if nothing is specified.

You will also need to re-add the `if (start !== undefined) { ` below.

extensions/web/09_file.js
```diff
@@ -189,79 +190,88 @@
...
-        } else {
-          relativeStart = Math.min(start, O.size);
-        }
-      }
-      /** @type {number} */
-      let relativeEnd;
-      if (end === undefined) {
-        relativeEnd = O.size;
-      } else {
-        if (end < 0) {
```
lucacasonato, line 234: This implements the spec algorithm exactly - is it worth changing? Just makes the diff larger :-)

extensions/web/09_file.js
```diff
@@ -67,58 +69,57 @@
...
 
   /**
      * @param {BlobPart[]} parts
      * @param {string} endings
-     * @returns {Uint8Array}
      */
   function processBlobParts(parts, endings) {
-    /** @type {Uint8Array[]} */
+    /** @type {(Uint8Array|Blob)[]} */
     const bytesArrays = [];
```
lucacasonato, line 98: Maybe `parts` would be a better name here? Also could you update the `@returns` to be correct?

lucacasonato (commented):
Great that you started on this!

There needs to be some changes to file backed blobs, mainly that they should not use the standard Deno FS APIs, as those go via regular permission checks. Instead there should be some ops in Rust that can only write files to a specific tmp dir, and can only access these files (and nothing else on disk). To JS they would only be exposed as UUIDs. These ops would not require any permissions. I can add these ops if you don't feel comfortable with Rust or the deno ops system.

Additionally we need to figure out what our metric for creating file based blobs is. We always know the size of a blob at creation, so do automatically create a file backed blob if the blob is larger than a specific threshold? Do we create blobs once we have used a fixed amount of memory for blob backing memory allocations in the process? File backed blobs should be fully transparent to the user. They should have no knowledge of them existing.

As you mentioned, FormData currently synchronously reads from Blob. This is not possible anymore with file backed blobs, so the entire FormData serializer needs to be turned into something that returns either a ReadableStream or an async iterable of chunks. Additionally `URL.createObjectURL` also synchronously reads from Blob - that also needs to be updated. That will require some changes to how object url blobs are stored in Rust.

jimmywarting:
> To JS they would only be exposed as UUIDs. These ops would not require any permissions. I can add these ops if you don't feel comfortable with Rust or the deno ops system.

Yes plz, thank you. I have never worked with rust before.

> Do we create blobs once we have used a fixed amount of memory for blob backing memory allocations in the process?

Not at the moment. But it could be a grate idea to offload some of the memory to a tmp file in the background. But could instead do that in a new feature PR. Have manly just focus on getting support of async blob sources.

> FormData serializer needs to be turned into something that returns either a ReadableStream or an async iterable of chunks

...Or a blob 🙃 https://github.com/denoland/deno/pull/10969#issuecomment-861369172 ☝️ 
quite simple to do actually, synchronous (with the help of async blob sources). 

> Additionally URL.createObjectURL also synchronously reads from Blob - that also needs to be updated. That will require some changes to how object url blobs are stored in Rust.

Again, would like to leave this up to you. haven't worked with Rust before

---

Will fix all you reviews

extensions/web/09_file.js
jimmywarting: Tried dealing with undefined before, resulted in getting `NaN` values
using default `arguments = undefined` don't make much sense...

Can revert and simply do `start || 0`

extensions/web/09_file.js
jimmywarting: Mostly copied everything from fetch-blob slice method b/c it operates on blobParts `(Uint8Array|Blob)[]` now instead.

could revert/keep this part where relative end/start is calculated

andreubotella:
> As you mentioned, FormData currently synchronously reads from Blob. This is not possible anymore with file backed blobs, so the entire FormData serializer needs to be turned into something that returns either a ReadableStream or an async iterable of chunks.
A couple months ago I worked on putting together a spec definition of `multipart/form-data` that could be included in the WHATWG standards (see whatwg/html#6424), including a serializer that returns a list of chunks which are either byte sequences or `File` objects: https://andreubotella.github.io/multipart-form-data/

jimmywarting:
Oh, speaking of serializing FormData boundary. On bad thing the blob spec is doing is lowercasing the hole mimetype

See: https://github.com/whatwg/html/issues/6251

when, blob cast the type to all lowercase it looses it boundary value, and it's no longer the same
`multipart/form-data; boundary=----WebKitFormBoundaryaUetjQwudca4vbSu`
`multipart/form-data; boundary=----webkitformboundaryauetjqwudca4vbsu`

I guess they are planing on only lowercasing the mimetype + keys (but keeping the values as is) ...the spec haven't been updated

we (in fetch-blob) also did this casting but removed it to be less strict

I think we should do this in the meanwhile:
```diff
-    return normalizedType.toLowerCase();
+    return normalizedType;
```

lucacasonato:
@jimmywarting Are you on the Deno discord (https://discord.gg/deno)? I have a basic implementation of the Rust side for this done, but have some questions - might be easier to chat there than through GitHub issues / PR review :-)

jimmywarting:
> Are you on the Deno discord

@lucacasonato yes I'm
I joined the discord

extensions/web/09_file.js
```diff
@@ -483,25 +464,35 @@
...
+      while (position !== end) {
+        const size = Math.min(end - position, POOL_SIZE);
+        const chunk = this.slice(position, position + size);
+        position += chunk.size;
+        yield core.opAsync("op_blob_read_part", chunk._id);
+      }
     }
   }
 
+  // TODO(jimmywarting): Maybe remove BlobDataItem?
```
lucacasonato, line 492: Yeah, I think it can go now that we have `BlobReference`. It is essentially the same thing, just one uses a dedicated API.

extensions/web/11_blob_url.js
```diff
@@ -33,7 +33,7 @@
     const url = core.opSync(
       "op_file_create_object_url",
       blob.type,
-      blob[_byteSequence],
+      getParts(blob),
```
jimmywarting, line 36: @lucacasonato I created util fn for getting the BlobReferences as a array of strings
Not sure how it's suppose to be called

Feedback?

lucacasonato:
The remaining test failure is an issue with an upstream WPT test. PR: https://github.com/web-platform-tests/wpt/pull/29517

lucacasonato:
Blocked on https://github.com/denoland/rusty_v8/pull/706 for `FinalizationRegistry` support

lucacasonato (approved):
LGTM. Thanks a lot @jimmywarting, this is a fantastic great refactor to prepare us for the native file system API, and opaque file backed blobs.

@bartlomieju could you take a look too?

cli/file_fetcher.rs
```diff
@@ -381,20 +381,23 @@ impl FileFetcher {
...
+        .ok_or_else(|| {
+          custom_error(
+            "NotFound",
+            format!("Blob URL not found: \"{}\".", specifier),
+          )
+        })?;
 
-    let content_type = blob.media_type;
+    let content_type = blob.media_type.clone();
+    let bytes = blob.read_all().await?;
```
bartlomieju, line 396: Looks like borrow to `blob_store` is held across this await point - if there are two blobs read concurrently it will probably panic.

extensions/fetch/Cargo.toml
```diff
@@ -19,6 +19,7 @@ data-url = "0.1.0"
 deno_core = { version = "0.92.0", path = "../../core" }
 deno_web = { version = "0.41.1", path = "../web" }
 http = "0.2.4"
+hyper = "0.14.9"
```
bartlomieju, line 22: Why is this dependency needed?

lucacasonato: Removed.

bartlomieju (approved):

BlackAsLight:
5 years later and I'm not quite sure, did we end up getting file backed blobs in Deno? Being able to open up a file as a blob would make writing decoders that support random access, as apposed to linear access, a lot better as one could use the same implementation for both Deno and the web. 

It would specifically enable me to write a decoder tar implementation that lets one stream several files out of it at once while also being browser compatible.

PR #10972, opened 2021-06-15
set minimum timeout to be 4 milliseconds

ry:

### DIFF
--- a/extensions/timers/lib.rs
+++ b/extensions/timers/lib.rs
@@ -124,6 +124,12 @@ pub fn op_global_timer_start(
   timeout: u64,
   _: (),
 ) -> Result<(), AnyError> {
+  // According to spec, minimum allowed timeout is 4 ms.
+  // https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers
+  // TODO(#10974) Per spec this is actually a little more complicated than this.
+  // The minimum timeout depends on the nesting level of the timeout.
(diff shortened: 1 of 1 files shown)


lucacasonato:
Per spec this is actually a little more complicated than this. The minimum timeout depends on the nesting level of the timeout. The actual spec wording is "If nesting level is greater than 5, and timeout is less than 4, then set timeout to 4.". Nesting level increases by 1 every time you call setTimeout or a new timer task is created for setInterval if the parent task of is also a timer task. The nesting level is reset back to 0 if the parent task is not a timer task.

I don't see an easy way to implement this at the moment though, so I'd say good enough for now. We should open an issue to track changing this behaviour to what the spec intends.

ry:
created https://github.com/denoland/deno/issues/10974 and added a TODO comment

lucacasonato (approved):
LGTM once test passes.

piscisaureus:
> I don't see an easy way to implement this at the moment though, so I'd say good enough for now.

I think this is good enough forever.

lucacasonato:
> > I don't see an easy way to implement this at the moment though, so I'd say good enough for now.
> 
> I think this is good enough forever.

I don't think so - a single `setTimeout(() => {}, 0)` is a common way to schedule a piece of code after the microtask queue has cleared. Delaying for 4ms there is not great.

piscisaureus (approved):

PR #10977, opened 2021-06-15
fix(inspector): Deno.inspect should inspect the object the proxy represents rather than the target

### LINKED ISSUE #10975: Inspecting a proxy with `showProxy: false` incorrectly inspects the target
The code in 02_console.js does not seem to properly inspect any object that's a proxy when `showProxy: false` (default) and instead inspects the target of the proxy.

For example:

```ts
const myObject = new Proxy({}, {
  get(_target, key) {
    if (key === Symbol.toStringTag) {
      return "MyProxy";
    } else {
      return 5;
    }
  },
  getOwnPropertyDescriptor() {
    return {
      enumerable: true,
      configurable: true,
      value: 5,
    };
  },
  ownKeys() {
    return ["prop1", "prop2"];
  },
});

console.log(Object.keys(myObject)); // [ "prop1", "prop2" ]

console.log(Deno.inspect(myObject)); // {}
```

Expected output:

```
MyProxy { prop1: 5, prop2: 5 }
```

Actual output:

```
{}
```

Affects Deno 1.11.0.

dsherret:
I was so confused why proxies weren't working in the inspector when working on a different fix.

Closes #10975.

Note: Target is the first argument provided to the constructor of `Proxy`

### DIFF
--- a/cli/tests/unit/console_test.ts
+++ b/cli/tests/unit/console_test.ts
@@ -1737,15 +1737,39 @@ unitTest(function inspectIterableLimit(): void {
 unitTest(function inspectProxy(): void {
   assertEquals(
     stripColor(Deno.inspect(
-      new Proxy([1, 2, 3], { get(): void {} }),
+      new Proxy([1, 2, 3], {}),
     )),
     "[ 1, 2, 3 ]",
   );
   assertEquals(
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

PR #10979, opened 2021-06-15
fix(inspect): fix inspecting event classes

### LINKED ISSUE #9880: `Object.defineProperty` on `Event` prototype panics in repl
running `Object.defineProperty(Event.prototype, 0, {})` in repl will panic with `thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli/tools/repl.rs:564:49`

dsherret:
Before in REPL:

```
> new Event("")
Event {
  bubbles: false,
  cancelable: false,
  composed: false,
  currentTarget: null,
  defaultPrevented: false,
  eventPhase: 0,
  srcElement: null,
  target: null,
  returnValue: true,
  timeStamp: 1623786515074,
  type: ""
}
> Event.prototype
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli\tools\repl.rs:550:49
```

After:

```
exit using ctrl+d or close()
> new Event("")
Event {
  bubbles: false,
  cancelable: false,
  composed: false,
  currentTarget: null,
  defaultPrevented: false,
  eventPhase: 0,
  srcElement: null,
  target: null,
  returnValue: true,
  timeStamp: 1623786578618,
  type: ""
}
> Event.prototype
Event {
  bubbles: [Getter/Setter],
  cancelable: [Getter/Setter],
  composed: [Getter/Setter],
  currentTarget: [Getter/Setter],
  defaultPrevented: [Getter/Setter],
  eventPhase: [Getter/Setter],
  srcElement: [Getter/Setter],
  target: [Getter/Setter],
  returnValue: [Getter/Setter],
  timeStamp: [Getter/Setter],
  type: [Getter/Setter]
}
>
```

Closes #9880

### DIFF
--- a/cli/tests/unit/event_test.ts
+++ b/cli/tests/unit/event_test.ts
@@ -138,17 +138,17 @@ unitTest(function inspectEvent(): void {
   assertEquals(
     Deno.inspect(Event.prototype),
     `Event {
-  bubbles: [Getter/Setter],
-  cancelable: [Getter/Setter],
-  composed: [Getter/Setter],
-  currentTarget: [Getter/Setter],
-  defaultPrevented: [Getter/Setter],
-  eventPhase: [Getter/Setter],
+  bubbles: [Getter],
+  cancelable: [Getter],
+  composed: [Getter],
+  currentTarget: [Getter],
+  defaultPrevented: [Getter],
+  eventPhase: [Getter],
   srcElement: [Getter/Setter],
-  target: [Getter/Setter],
+  target: [Getter],
   returnValue: [Getter/Setter],
-  timeStamp: [Getter/Setter],
-  type: [Getter/Setter]
+  timeStamp: [Getter],
+  type: [Getter]
 }`,
   );
 
--- a/cli/tests/unit/event_test.ts
+++ b/cli/tests/unit/event_test.ts
@@ -1,5 +1,10 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
-import { assert, assertEquals, unitTest } from "./test_util.ts";
+import {
+  assert,
+  assertEquals,
+  assertStringIncludes,
+  unitTest,
+} from "./test_util.ts";
 
 unitTest(function eventInitializedWithType(): void {
   const type = "click";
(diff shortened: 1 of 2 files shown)


cli/tests/unit/event_test.ts
```diff
@@ -127,3 +132,30 @@ unitTest(function eventInspectOutput(): void {
     assertEquals(Deno.inspect(event), outputProvider(event));
   }
 });
+
+unitTest(function inspectEvent(): void {
+  // has a customInspect implementation that previously would throw on a getter
+  assertEquals(
+    Deno.inspect(Event.prototype),
+    `Event {
+  bubbles: [Getter/Setter],
```
dsherret, line 141: I'm not sure why these have empty setters defined... couldn't those be omitted?

lucacasonato: Yeah, those should be removed. Some legacy from when we understood how webidl works even less.

dsherret: Done in b2a8f21d848bfd598207bd991546f4a464173beb

dsherret: Nevermind. We need these because it causes an error to not have it when doing `e.target = something`, which doesn't error in chrome. I'm going to revert.

lucacasonato: Tested in the Chrome Dev Tools? That is non strict mode. It does error in Chrome too in ES modules or "use strict" scripts. That is how this snuck in in the first place :-)

dsherret: Oh, I didn’t think about that! Ok, yeah I’ll look into it on Thursday. There is a tests that assigns to target for some reason.

dsherret: @lucacasonato I was thinking about this. Maybe we should do it in a separate PR and minor release? It could possibly break some people's code. Thoughts?

lucacasonato: If you want, sure - but its a bug fix so I'd be happy to land it in a patch.

dsherret: Ok, lets merge this one and then I'll do a separate specific PR for that change so that it's not hidden under this one's title.

lucacasonato (approved):

PR #10982, opened 2021-06-15
fix: closing / aborting WritableStream is racy

### LINKED ISSUE #10980: WritableStream close WPT result in internal assertions & promise rejections
See results: https://staging.wpt.fyi/results/streams/writable-streams/close.any.html?label=master&product=deno

It feels to me like all failures are caused by the same underlying issue that results in some promise either not being handled correctly, or being resolved / rejected at the wrong time.

lucacasonato:
Took me 4 hours to find this. The long tail end of compat :-)

See https://github.com/whatwg/streams/pull/1137

I guess we were technically spec compliant before? :wink: 

Fixes #10980

### DIFF
--- a/extensions/web/06_streams.js
+++ b/extensions/web/06_streams.js
@@ -2808,7 +2808,7 @@
   function writableStreamHasOperationMarkedInFlight(stream) {
     if (
       stream[_inFlightWriteRequest] === undefined &&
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM

PR #10999, opened 2021-06-16
feat(cli): support "types" when type checking

### LINKED ISSUE #10677: Deno.emit: Add option for ambient type declarations
This is a follow up to the discussion at https://github.com/denoland/deno/discussions/10549

I see that I can supply predefined environment enums in the emit options `compilerOptions.lib` array (e.g. `ES2017`, `DOM.Iterable`, etc.), but I don't see a way to provide other environment types (e.g. paths to declaration files) because properties like `compilerOptions.typeRoots` aren't supported.

@kitsonk [suggested](https://github.com/denoland/deno/discussions/10549#discussioncomment-716399) using the [`compilerOptions.types`](https://doc.deno.land/builtin/unstable#Deno.CompilerOptions) property, but that doesn't seem to work (it produces another error).

If I include

```ts
import "./types/mod.d.ts";
```

at the top of every source file then it compiles without errors, but that is not a maintainable approach.

If I want to compile for other [custom] runtime environments, do I really have to import the types into every source file or is there a configuration alternative?

I'm attaching a simple example for demonstration and reproduction:

[example.zip](https://github.com/denoland/deno/files/6497380/example.zip)

```
example % cat run 
deno --unstable run --allow-read=. --allow-write=. main.ts

example % ./run 
TS2688 [ERROR]: Cannot find type definition file for './types/shared-node-browser.d.ts'.
  The file is in the program because:
    Entry point of type library './types/shared-node-browser.d.ts' specified in compilerOptions

TS2584 [ERROR]: Cannot find name 'console'. Do you need to change your target library? Try changing the `lib` compiler option to include 'dom'.
  console.log('hello world');
  ~~~~~~~
    at file:///Users/jesse/Downloads/example/mod.ts:5:3

Found 2 errors.
"mod.bundle.cjs" written
```

kitsonk:
Fixes #10677

This supports the usage of `"types"` in compiler options in a `--config` file and `Deno.emit()` and supports the triple-slash `types` directive in modules.

### DIFF
--- a/cli/config_file.rs
+++ b/cli/config_file.rs
@@ -31,6 +31,14 @@ pub struct EmitConfigOptions {
   pub jsx_fragment_factory: String,
 }
 
+/// There are certain compiler options that can impact what modules are part of
+/// a module graph, which need to be deserialized into a structure for analysis.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CompilerOptions {
+  pub types: Option<Vec<String>>,
+}
+
 /// A structure that represents a set of options that were ignored and the
 /// path those options came from.
 #[derive(Debug, Clone, PartialEq)]
@@ -90,7 +98,6 @@ pub const IGNORED_COMPILER_OPTIONS: &[&str] = &[
   "sourceMap",
   "sourceRoot",
   "target",
-  "types",
   "useDefineForClassFields",
 ];
 
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -421,11 +421,25 @@ declare namespace Deno {
       | "es2019"
       | "es2020"
       | "esnext";
-    /** List of names of type definitions to include. Defaults to `undefined`.
+    /** List of names of type definitions to include when type checking.
+     * Defaults to `undefined`.
      *
      * The type definitions are resolved according to the normal Deno resolution
-     * irrespective of if sources are provided on the call. Like other Deno
-     * modules, there is no "magical" resolution. For example:
+     * irrespective of if sources are provided on the call. In addition, unlike
+     * passing the `--config` option on startup, there is no base to resolve
+     * relative specifiers, so the specifiers here have to be fully qualified
+     * URLs or paths.  For example:
+     *
+     * ```ts
+     * Deno.emit("./a.ts", {
+     *   compilerOptions: {
+     *     types: [
+     *       "https://deno.land/x/pkg/types.d.ts",
+     *       "/Users/me/pkg/types.d.ts",
+     *     ]
+     *   }
+     * });
+     * ```
      */
     types?: string[];
     /** Emit class fields with ECMAScript-standard semantics. Defaults to
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -61,14 +61,18 @@ impl TsServer {
   pub fn new() -> Self {
     let (tx, mut rx) = mpsc::unbounded_channel::<Request>();
     let _join_handle = thread::spawn(move || {
-      // TODO(@kitsonk) we need to allow displaying diagnostics here, but the
-      // current compiler snapshot sends them to stdio which would totally break
-      // the language server...
-      let mut ts_runtime = start(false).expect("could not start tsc");
+      let mut ts_runtime = load().expect("could not load tsc");
 
       let runtime = create_basic_runtime();
       runtime.block_on(async {
+        let mut started = false;
         while let Some((req, state_snapshot, tx)) = rx.recv().await {
+          if !started {
+            // TODO(@kitsonk) need to reflect the debug state of the lsp here
+            start(&mut ts_runtime, false, &state_snapshot)
+              .expect("could not start tsc");
+            started = true;
+          }
           let value = request(&mut ts_runtime, state_snapshot, req);
           if tx.send(value).is_err() {
             warn!("Unable to send result to client.");
@@ -572,7 +576,7 @@ impl DocumentSpan {
     line_index: &LineIndex,
     language_server: &mut language_server::Inner,
   ) -> Option<lsp::LocationLink> {
-    let target_specifier = resolve_url(&self.file_name).unwrap();
+    let target_specifier = normalize_specifier(&self.file_name).unwrap();
     let target_line_index = language_server
       .get_line_index(target_specifier.clone())
       .await
@@ -773,7 +777,7 @@ impl ImplementationLocation {
     line_index: &LineIndex,
     language_server: &mut language_server::Inner,
   ) -> lsp::Location {
-    let specifier = resolve_url(&self.document_span.file_name).unwrap();
+    let specifier = normalize_specifier(&self.document_span.file_name).unwrap();
     let uri = language_server
       .url_map
       .normalize_specifier(&specifier)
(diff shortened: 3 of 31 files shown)


cli/config_file.rs
```diff
@@ -31,6 +31,14 @@ pub struct EmitConfigOptions {
   pub jsx_fragment_factory: String,
 }
 
+/// There are certain compiler options that can impact what modules are part of
+/// a module graph, which need to be deserialized into a structure for analysis.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CompilerOptions {
+  pub types: Option<Vec<String>>,
+}
```
bartlomieju, line 40: I'm a bit confused now - I thought we already parsed the value of `compilerOptions` into a concrete structure. Isn't that the case?

cli/module_graph.rs
```diff
@@ -1852,6 +1851,53 @@ impl GraphBuilder {
...
+          }
+        }
+      }
+    }
+    Ok(())
+  }
+
+  /// Analyze a config file, identifying any specifiers that need to be resolved
+  /// and added to the graph.
+  pub async fn analyze_config_file(
```
bartlomieju, line 1875: I presume these two methods are required to handle both strictly types compiler options coming from `ConfigFile` as well as JSON value coming from runtime compiler APIs?

docs/typescript/types.md
```diff
@@ -101,6 +101,67 @@ When seeing this header, Deno would attempt to retrieve
...
+      "/Users/me/pkg/types.d.ts"
+    ]
+  }
+}
+```
+
+Like the triple-slash reference above, the specifier supplied in the `"types"`
+array will be resolved like other specifiers in Deno. In the case of relative
+specifiers, it will be resolved relative to the path to the `tsconfig.json`.
+
```
bartlomieju, line 164: Nice docs!

bartlomieju (approved):
LGTM, hopefully there won't be many conflicts with #10492

docs/typescript/types.md
```diff
@@ -101,6 +101,67 @@ When seeing this header, Deno would attempt to retrieve
 `https://example.com/coolLib.d.ts` and use that when type checking the original
 module.
 
+### Using ambient or global types
+
+Overall it is better to use module/UMD type definitions with Deno, where a
+module expressly imports the types it depends upon. Modular type definitions can
+express
+[augmentation of the global scope](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-modifying-module-d-ts.html)
+via the `declare global` in the type definition. For example:
```
nayeemrmn, line 110: Hm, I don't agree that modules with side-effects on the whole program's type-checking are better to use than the introduced `"types"` config. They're very oxymoronic. Global augmentations are better represented at the CLI/config level.

docs/typescript/types.md
```diff
@@ -101,6 +101,67 @@ When seeing this header, Deno would attempt to retrieve
...
+For example:
+
+```json
+{
+  "compilerOptions": {
+    "types": [
+      "./types.d.ts",
+      "https://deno.land/x/pkg@1.0.0/types.d.ts",
+      "/Users/me/pkg/types.d.ts"
+    ]
```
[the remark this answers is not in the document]
kitsonk: All module specifiers in Deno break with tsc. 🤷 While it isn't common knowledge, `"types"` accepts relative paths. tsc then looks for a range of files until it finds one. The only thing we are doing is handling the lack of a magical extension.

I really don't want to add more flags until we have to, even in a tsconfig-less future.

docs/typescript/types.md
kitsonk: I guess we have to agree to disagree on this. Code expressing it dependencies/expectations as code makes the code more portable without requiring magical configuration.

cli/module_graph.rs
kitsonk: For `Deno.emit()` no, we pass a `Value` from the op when we merge it into the config.

cli/config_file.rs
kitsonk: For `Deno.emit()` no, we pass a `Value` from the op when we merge it into the config.

cli/module_graph.rs
kitsonk: Correct.

docs/typescript/types.md
nayeemrmn: Global augmentations in modules are implicit dependencies. Exemplary usage of these is having the single entrypoint module in your project import the augmenter and leaving the rest with this implicit dependency. This is not more portable in the big picture. Needing to specify them at application-level (since they have an application-wide effect) would have the dual effect of making them explicit and discouraging them altogether. The explicitness is in the form of publicly documenting a list of `d.ts` URLs alongside your project which consumers cannot avoid mentally parsing and referencing at the root of their application.

But we can't remove the language feature, this is just a matter of doc priority, so agree to disagree.

dsherret (approved):
LGTM, but it would be good if we could eventually get rid of the specifier normalization by fixing whatever is happening in typescript.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -421,11 +421,25 @@ declare namespace Deno {
...
+    /** List of names of type definitions to include when type checking.
+     * Defaults to `undefined`.
      *
      * The type definitions are resolved according to the normal Deno resolution
-     * irrespective of if sources are provided on the call. Like other Deno
-     * modules, there is no "magical" resolution. For example:
+     * irrespective of if sources are provided on the call. In addition, unlike
+     * passing the `--config` option on startup, there is no base to resolve
+     * relative specifiers, so the specifiers here have to be fully qualified
+     * URLs or paths.  For example:
```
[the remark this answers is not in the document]
kitsonk: Interesting point.  The current challenge is that we have to effectively set a "cwd" for the relative paths in types, because tsc will request relative modules relative to the "cwd" we provide it, unlike modules where we are provided with the referrer and we get to fully resolve the module however we see fit.  It is similar to import-maps.

At the point where we determine the cwd in tsc, it is hard to figure out the root of the `Deno.emit()` at the moment.  We could use the real `cwd` but that means the relative point of `Deno.emit()` and this would/could be different.  What I would like to do is land it with this restriction at the moment and consider figuring out a way to make it relative as an enhancement?

cli/lsp/tsc.rs
```diff
@@ -1812,8 +1817,36 @@ impl<'a> State<'a> {
...
+      .to_string()
+  }
+
+  /// In certain situations, tsc can request "invalid" specifiers and this will
+  /// normalize and memoize the specifier.
+  fn normalize_specifier<S: AsRef<str>>(
+    &mut self,
+    specifier: S,
+  ) -> Result<ModuleSpecifier, AnyError> {
+    let specifier_str = specifier.as_ref().replace(".d.ts.d.ts", ".d.ts");
```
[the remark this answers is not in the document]
kitsonk: TypeScript being "broken" is a matter of opinion, I highly doubt they would consider fixing something that works for what it was designed for.  TypeScript is hardcoded to "guess" at extensions, so us being explicit about extensions "breaks" TypeScript in a few different ways.  The problem is that TypeScript isn't using the "resolve module" here, or any "trapable" API.  It has an algorithm which mimics Node.js resolution to try to find the type definitions, and I think the 4 or 5 item it attempts is `${cwd}/${specifier}.d.ts` (it tries things like `${cwd}/node_modules/${specifier}/package.json` and `${cwd}/node_modules/${specifier}/index.d.ts` first and a few others).  Since we can't control the resolution algorithm in this case, we are stuck this matching on something that always gets `.d.ts` banged onto the end of it.

Because the logic could change, or be different, or whatever in the future is why I broke it out into methods to abstract the rest of the code from all of that.  If we find error in this, we should only need to change it in one spot.

PR #11001, opened 2021-06-16
feat(repl): show list completion

### LINKED ISSUE #10997: Consider using list completions in REPL
`rustyline` allows to configure how completions work. Currently Deno uses "circular" method which completes on tab and on subsequent tab propose next completion.

`rustyline` can be configured to use "list completions" which on second tab show list of possible completions:
Screenshot 2021-06-16 at 03 44 21

It's done on second tab press because there are "ambigous" entries that match - if you press tab and there's only a single possible completion it will be filled.

IMO this is better than having to cycle through all possibilities.

CC @dsherret 

EDIT:

Here's relevant git patch that provides these changes:
```
diff --git a/cli/tools/repl.rs b/cli/tools/repl.rs
index 67e68c0cd..c9b355c83 100644
--- a/cli/tools/repl.rs
+++ b/cli/tools/repl.rs
@@ -19,6 +19,8 @@ use rustyline::validate::ValidationResult;
 use rustyline::validate::Validator;
 use rustyline::Context;
 use rustyline::Editor;
+use rustyline::Config;
+use rustyline::CompletionType;
 use rustyline_derive::{Helper, Hinter};
 use std::borrow::Cow;
 use std::path::PathBuf;
@@ -302,7 +304,10 @@ struct ReplEditor {

 impl ReplEditor {
   pub fn new(helper: EditorHelper, history_file_path: PathBuf) -> Self {
-    let mut editor = Editor::new();
+    let editor_config = Config::builder()
+      .completion_type(CompletionType::List)
+      .build();
+    let mut editor = Editor::with_config(editor_config);
     editor.set_helper(Some(helper));
     editor.load_history(&history_file_path).unwrap_or(());
```

caspervonb:
Closes #10997


bartlomieju (approved):
LGTM

PR #11002, opened 2021-06-16
tests: get epochs/daily commit with ./wpt rev-list

lucacasonato:
Removes dependency on scheduled time being coordinated with wpt upstream.


ry (approved):
LGTM

PR #11005, opened 2021-06-16
fix: align URL / URLSearchParams to spec

### LINKED ISSUE #10996: URL global should use webidl helpers
See https://staging.wpt.fyi/results/url/idlharness.any.html?label=master&product=deno

lucacasonato:
Fixes #10996


bartlomieju (approved):
LGTM, some of those webidl stuff look like black magic incantations

PR #11006, opened 2021-06-16
docs(contributing/style_guide) Fix #11004 and remove mentions of `cli/js`

### LINKED ISSUE #11004: Style guide should encourage code example in comments
With support for running type checking on code block examples in Jsdoc comments, we should upgrade the style guide accordingly.

grian32:
Along with fixing #11004, this PR also removes or replaces any mentions of `cli/js` under the `Typescript` section as that folder no longer appears to exist and AFAIK any TS has been replaced with JS a while back. I could make a `Javascript` section for any JS code but I don't believe that is necessary as it'd be mostly repeating the Typescript section.

### DIFF
--- a/docs/contributing/style_guide.md
+++ b/docs/contributing/style_guide.md
@@ -49,8 +49,7 @@ Follow Rust conventions and be consistent with existing code.
 
 ## TypeScript
 
-The TypeScript portions of the codebase include `cli/js` for the built-ins and
-the standard library `std`.
+The TypeScript portion of the code base is the standard library `std`.
 
 ### Use TypeScript instead of JavaScript.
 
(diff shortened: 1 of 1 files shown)


docs/contributing/style_guide.md
```diff
@@ -198,9 +197,9 @@ export type { Person } from "./my_file.ts";
 
 ### Minimize dependencies; do not make circular imports.
 
-Although `cli/js` and `std` have no external dependencies, we must still be
-careful to keep internal dependencies simple and manageable. In particular, be
-careful not to introduce circular imports.
+Although `std` has no external dependencies, we must still be careful to keep
```
caspervonb, line 200: runtime/js would be where most of it is now.

grian32: i mostly removed it since this is under the `Typescript` section, but a possible solution that would still include `runtime/js` would be to convert it to `Typescript/Javascript` and include the parts with `runtime/js` again, not sure if it would cause confusion around typing though

ry (approved):
Thanks

PR #10809, opened 2021-05-31
[WIP] refactor(core): Move module loading polling to ModuleMap

bartlomieju:

PR #11015, opened 2021-06-17
fix(core/modules): Prepare modules only once per runtime

### LINKED ISSUE #9802: Module graph is too eager to re-check
The module graph is a bit too eager to reload and re-check and this doesn't work that great in certain scenarios.

For example with reload set to true, it'll always re-check for each dynamic import and worker instantiation:

```js
for (let i = 0; i < 1000; i++) {
  const worker = new Worker(new URL("./worker.ts", import.meta.url).href, { type: 'module' });
}
```

Depending on the complexity of the graph, the above code can forever to execute with `deno run --reload`.

We want to check against something like `needs_reload(specifier)` rather than just `flags.reload` in `module_graph.check`.

### LINKED ISSUE #11014: Redundant re-checking of dynamic imports
image

You can cause the module to be type-checked again by changing the file between repeated dynamic imports, but the module is already loaded so this should not happen.

_Fully_ short-circuiting a repeated dynamic import is quite hard, but we can easily skip the `load_prepare()` step under certain conditions. Fix coming.

This is unrelated to #9802.

nayeemrmn:
Fixes #9802.
Fixes #11014.

### DIFF
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -44,7 +44,6 @@ use deno_core::ModuleSource;
 use deno_core::ModuleSpecifier;
 use log::debug;
 use regex::Regex;
-use std::borrow::Cow;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::error::Error;
@@ -611,7 +610,7 @@ pub struct BundleOptions {
 }
 
 #[derive(Debug, Default)]
-pub struct CheckOptions<'a> {
+pub struct CheckOptions {
   /// If `true` then debug logging will be output from the isolate.
   pub debug: bool,
   /// Utilise the emit from `tsc` to update the emitted code for modules.
@@ -628,7 +627,7 @@ pub struct CheckOptions<'a> {
   /// A set of module specifiers to be excluded from the effect of
   /// `CheckOptions::reload` if it is `true`. Perhaps because they have already
   /// reloaded once in this process.
-  pub reload_exclusions: Cow<'a, HashSet<ModuleSpecifier>>,
+  pub reload_exclusions: HashSet<ModuleSpecifier>,
 }
 
 #[derive(Debug, Eq, PartialEq)]
@@ -668,7 +667,7 @@ pub struct EmitOptions {
 
 /// A structure which provides options when transpiling modules.
 #[derive(Debug, Default)]
-pub struct TranspileOptions<'a> {
+pub struct TranspileOptions {
   /// If `true` then debug logging will be output from the isolate.
   pub debug: bool,
   /// An optional config file with user supplied TypeScript configuration
@@ -681,7 +680,7 @@ pub struct TranspileOptions<'a> {
   /// A set of module specifiers to be excluded from the effect of
   /// `CheckOptions::reload` if it is `true`. Perhaps because they have already
   /// reloaded once in this process.
-  pub reload_exclusions: Cow<'a, HashSet<ModuleSpecifier>>,
+  pub reload_exclusions: HashSet<ModuleSpecifier>,
 }
 
 #[derive(Debug, Clone)]
--- a/cli/module_graph.rs
+++ b/cli/module_graph.rs
@@ -44,6 +44,7 @@ use deno_core::ModuleSource;
 use deno_core::ModuleSpecifier;
 use log::debug;
 use regex::Regex;
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::error::Error;
@@ -610,7 +611,7 @@ pub struct BundleOptions {
 }
 
 #[derive(Debug, Default)]
-pub struct CheckOptions {
+pub struct CheckOptions<'a> {
   /// If `true` then debug logging will be output from the isolate.
   pub debug: bool,
   /// Utilise the emit from `tsc` to update the emitted code for modules.
@@ -624,6 +625,10 @@ pub struct CheckOptions {
   /// Ignore any previously emits and ensure that all files are emitted from
   /// source.
   pub reload: bool,
+  /// A set of module specifiers to be excluded from the effect of
+  /// `CheckOptions::reload` if it is `true`. Perhaps because they have already
+  /// reloaded once in this process.
+  pub reload_exclusions: Cow<'a, HashSet<ModuleSpecifier>>,
 }
 
 #[derive(Debug, Eq, PartialEq)]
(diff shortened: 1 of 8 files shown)


nayeemrmn:
@bartlomieju Please review.

core/modules.rs
```diff
@@ -648,17 +603,32 @@ impl ModuleMap {
...
     specifier: &str,
     code: Option<String>,
-  ) -> RecursiveModuleLoad {
-    RecursiveModuleLoad::main(
+  ) -> Result<RecursiveModuleLoad, AnyError> {
+    let module_specifier =
+      self
+        .loader
+        .resolve(self.op_state.clone(), specifier, ".", true)?;
+    let load = RecursiveModuleLoad::main(
```
bartlomieju, line 615: I don't really understand changes in this file. Please explain.

nayeemrmn: Other than adding [this code path](https://github.com/denoland/deno/pull/11015/files#diff-aa01489135c068a2056422f99878ba91861c585baed075470c08c1ac439ab3b2R654-R655) to short-circuit preparing dynamic imported modules already in the module map, there are two refactors:
1. Move [this code](https://github.com/denoland/deno/pull/11015/files#diff-5b779288098bb6449725245ce7fdb1ec3190563b5dd255e747cfc35e639407aeL1258) (after the singular call site of `ModuleMap::load_main()`) inside of `ModuleMap::load_main()` to make its functionality more similar to `ModuleMap::load_dynamic_import()`.
2. Remove and inline `RecursiveModuleLoad::prepare()` at its two call sites in `ModuleMap::load_main()` and `ModuleMap::load_dynamic_import()`. Something like `RecursiveModuleLoad::prepare()` should probably still exist, but that would lead to a duplicate resolve in `ModuleMap::load_dynamic_import()`.

bartlomieju: > Other than adding this code path to short-circuit preparing dynamic imported modules already in the module map, there are two refactors:


I'm a bit weary about it. This function is one of problems in #8421 and I'm making strides to rewrite it into a future instead.

> - Move this code (after the singular call site of ModuleMap::load_main()) inside of ModuleMap::load_main() to make its functionality more similar to ModuleMap::load_dynamic_import(). 
> - Remove and inline RecursiveModuleLoad::prepare() at its two call sites in ModuleMap::load_main() and ModuleMap::load_dynamic_import(). Something like RecursiveModuleLoad::prepare() should probably still exist, but that would lead to a duplicate resolve in ModuleMap::load_dynamic_import().

So these two are just cleanups to make more of the APIs internal to `ModuleMap`. Can't `RecursiveModuleLoad::prepare()` be left as is and called by those two related methods? I'm worried they might diverge at some point.

Anyway, the tests look good and are passing so I think I'll land it and worry about potential troubles with `is_registered` later

bartlomieju: Though I'd like to get @kitsonk's review as well as this PR touches the module graph.

nayeemrmn: > Can't `RecursiveModuleLoad::prepare()` be left as is and called by those two related methods? I'm worried they might diverge at some point.

Diverging isn't a problem, the two call sites after still share nothing except for a `prepare_load()` call. You'll notice `RecursiveModuleLoad::prepare()` had two separate code paths for if it was called from `load_main()` or `load_dynamic_import()`, which is why I thought to inline it. The amount of shared code is the same as it was before.

It's awkward because `RecursiveModuleLoad` would ideally keep a method for preparing itself, but as I mentioned that creates complications as to where and how many times `resolve()` is called in `load_dynamic_import()`.

bartlomieju: > Diverging isn't a problem, the two call sites after still share nothing except for a prepare_load() call. You'll notice RecursiveModuleLoad::prepare() had two separate code paths for if it was called from load_main() or load_dynamic_import(), which is why I thought to inline it. The amount of shared code is the same as it was before.

Of course 🤦‍♂️ that makes perfect sense

It's awkward because RecursiveModuleLoad would ideally keep a method for preparing itself, but as I mentioned that creates complications as to where and how many times resolve() is called in load_dynamic_import().

Is it a problem that resolve is called multiple times? It should be idempotent after all.

nayeemrmn: Okay, I've brought back `RecursiveModuleLoad::prepare()`. I think `resolve()` has duped calls in other places anyway. On another note I want to see if I can make simplifications to `RecursiveModuleLoad::state`/`LoadState`.

bartlomieju: @nayeemrmn just a heads up that I'm also working on `core/modules.rs` in #10809

nayeemrmn: I've replaced the following structure:
```rust
struct RecursiveModuleLoad {
  state: LoadState,
  // ...
}

enum LoadState {
  ResolveMain(String, Option<String>),
  ResolveImport(String, String),
  LoadingRoot,
  LoadingImports,
  Done,
}
```
with:
```rust
struct RecursiveModuleLoad {
  init: LoadInit,
  state: LoadState,
  // ...
}

enum LoadInit {
  Main(String), // specifier
  MainWithCode(String, String), // specifier, code
  DynamicImport(String, String), // specifier, referrer
}

enum LoadState {
  Init,
  LoadingRoot,
  LoadingImports,
  Done,
}
```
It's a lot more intuitive and usable. Reading `impl Stream for RecursiveModuleLoad {}` is a lot easier.

bartlomieju: @nayeemrmn just to be clear, nothing changed in the actual behavior of `RecursiveModuleLoad`?

nayeemrmn: > nothing changed in the actual behavior of `RecursiveModuleLoad`?

Correct, it's a refactor.

cli/module_graph.rs
```diff
@@ -610,7 +611,7 @@ pub struct BundleOptions {
 }
 
 #[derive(Debug, Default)]
-pub struct CheckOptions {
+pub struct CheckOptions<'a> {
```
kitsonk, line 614: It seems like we are really going out of our way to avoid some cloning, which makes for some really ugly code.  I would much rather prefer we not throw pointer lifetimes everywhere and have to call stuff with `Cow::borrow`.

nayeemrmn: `Cow` was the least obtrusive way I knew of to get it to satisfy `Default`, which was the only problem with a simple reference. I've changed it to just clone the the sets instead.

bartlomieju (approved):
LGTM, thank you Nayeem, this is a good fix and gave me an idea how to fix problems in concurrent module loading.

cli/module_graph.rs
nayeemrmn: 😅 Turns out no cloning is necessary anyway, there's only one consumption per branch. My mistake.

PR #11023, opened 2021-06-17
chore(lsp): Add `--parent-pid <pid>` flag

### LINKED ISSUE #11019: `deno lsp` should exit when the editor process no longer exists
When launching `deno lsp` it should check for the existence of the editor process every X seconds. If the editor process no longer exists, then it should exit.

This could be achieved reliably via a CLI flag:

```
deno lsp --parent-pid <pid>
```

The reason for a flag is because getting the parent pid without the flag is is not always reliable and using a CLI flag requires less platform specific code.

Refs:

https://github.com/denoland/vscode_deno/issues/374

dsherret:
This PR adds a new `--parent-pid <pid>` flag to `deno lsp` that when provided starts a task that checks for the existence of the provided process id (ex. vscode's) every 30 seconds. If the process doesn't exist (meaning the deno process has nothing interacting with it), then it terminates itself.

I am  not familiar with what the best way to check if a process is running on linux/mac. I am also not that familiar with win api. Please let me know if there is a better way of doing what's done in here.

Closes #11019

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -85,7 +85,7 @@ pub enum DenoSubcommand {
     force: bool,
   },
   Lsp {
-    parent_pid: Option<usize>,
+    parent_pid: Option<u32>,
   },
   Lint {
     files: Vec<PathBuf>,
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -84,7 +84,9 @@ pub enum DenoSubcommand {
     root: Option<PathBuf>,
     force: bool,
   },
-  Lsp,
+  Lsp {
+    parent_pid: Option<usize>,
+  },
   Lint {
     files: Vec<PathBuf>,
     ignore: Vec<PathBuf>,
@@ -876,6 +878,16 @@ go-to-definition support and automatic code formatting.
 
 How to connect various editors and IDEs to 'deno lsp':
 https://deno.land/manual/getting_started/setup_your_environment#editors-and-ides")
+    .arg(
+      Arg::with_name("parent-pid")
+        .long("parent-pid")
+        .help("The parent process id to periodically check for the existence of or exit")
+        .takes_value(true)
+        .validator(|val: String| match val.parse::<usize>() {
+          Ok(_) => Ok(()),
+          Err(_) => Err("parent-pid should be a number".to_string()),
+        }),
+    )
 }
 
 fn lint_subcommand<'a, 'b>() -> App<'a, 'b> {
@@ -1621,8 +1633,11 @@ fn install_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   };
 }
 
-fn lsp_parse(flags: &mut Flags, _matches: &clap::ArgMatches) {
-  flags.subcommand = DenoSubcommand::Lsp;
+fn lsp_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
+  let parent_pid = matches
+    .value_of("parent-pid")
+    .map(|val| val.parse().unwrap());
+  flags.subcommand = DenoSubcommand::Lsp { parent_pid };
 }
 
 fn lint_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
@@ -2308,10 +2323,24 @@ mod tests {
     assert_eq!(
       r.unwrap(),
       Flags {
-        subcommand: DenoSubcommand::Lsp,
+        subcommand: DenoSubcommand::Lsp { parent_pid: None },
         ..Flags::default()
       }
     );
+
+    let r = flags_from_vec(svec!["deno", "lsp", "--parent-pid", "5"]);
+    assert_eq!(
+      r.unwrap(),
+      Flags {
+        subcommand: DenoSubcommand::Lsp {
+          parent_pid: Some(5),
+        },
+        ..Flags::default()
+      }
+    );
+
+    let r = flags_from_vec(svec!["deno", "lsp", "--parent-pid", "invalid-arg"]);
+    assert!(r.is_err());
   }
 
   #[test]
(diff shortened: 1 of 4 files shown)


cli/flags.rs
```diff
@@ -84,7 +84,9 @@ pub enum DenoSubcommand {
     root: Option<PathBuf>,
     force: bool,
   },
-  Lsp,
+  Lsp {
+    parent_pid: Option<u32>,
```
dsherret, line 88: Used `u32` to align with what Rust uses for pids: https://doc.rust-lang.org/std/process/struct.Child.html#method.id

cli/lsp/parent_process_checker.rs
```diff
@@ -0,0 +1,76 @@
...
+  });
+}
+
+#[cfg(unix)]
+fn is_process_active(process_id: u32) -> bool {
+  unsafe {
+    // signal of 0 checks for the existence of the process id
+    libc::kill(process_id as i32, 0) == 0
+  }
+}
```
dsherret, line 28: I'm not really familiar with libc or winapi, so I'm not sure if what's being done here is correct. Is there a better way of doing this? I did not want to pull in another dependency and didn't see a way to do this with our existing dependencies, but I am likely missing something.

MarkTiedemann:
> I am also not that familiar with win api. Please let me know if there is a better way of doing what's done in here.

Rather than repeatedly check `GetExitCodeProcess`, I believe you can use [`WaitForSingleObject`](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject) or [`RegisterWaitForSingleObject`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerwaitforsingleobject). (But I'm not too familiar with these APIs either.)

dsherret:
> Rather than repeatedly check `GetExitCodeProcess`, I believe you can use [`WaitForSingleObject`](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject) or [`RegisterWaitForSingleObject`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerwaitforsingleobject). (But I'm not too familiar with these APIs either.)

Oh yeah, looks like it's better at telling if the process is running or not: https://stackoverflow.com/q/12900036/188246

I've updated the code. Thanks @MarkTiedemann!

MarkTiedemann:
You are using `WaitForSingleObject` with a timeout of `0` which, though it may be better than `GetExitCodeProcess`, is still a snapshot in time.

The point I wanted to make was that, instead of checking the status of the process in a specific interval, those APIs can be used to wait until the exact moment in time when the process exits (without repeatedly checking).

Whereas `WaitForSingleObject` would block the executing thread, `RegisterWaitForSingleObject` is async. It uses a callback to communicate when the process was signaled (when a process exits, its handle becomes signaled).

But again, I haven't actually used these APIs yet. Just read about them (so I may very well be wrong...).

PS: I believe checking in an interval is not a best practice since process IDs may be recycled, e.g. if, in between checks, the process exits and its ID is assigned to a newly created process, you'd end up checking the wrong process.

dsherret:
> The point I wanted to make was that, instead of checking the status of the process in a specific interval, those APIs can be used to wait until the exact moment in time when the process exits (without repeatedly checking).
>
> Whereas WaitForSingleObject would block the executing thread, RegisterWaitForSingleObject is async. It uses a callback to communicate when the process was signaled (when a process exits, its handle becomes signaled).

Yeah, I just prefered as little platform specific code as possible. Is there a way to do the same thing on linux and mac?

Also, reading into it, `RegisterWaitForSingleObject` also blocks a thread (https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-registerwaitforsingleobject), so might as well use `WaitForSingleObject` with `spawn_blocking` I think.

> PS: I believe checking in an interval is not a best practice since process IDs may be recycled, e.g. if, in between checks, the process exits and its ID is assigned to a newly created process, you'd end up checking the wrong process.

I was thinking about this and yeah, it's the main issue with this appoach. I believe there's a low chance of that happening though and if it does it's not a big deal. We could reduce the sleep time to reduce the likilihood... 30 seconds is probably too long.

kitsonk:
Any thoughts on how we would be able to detect when this occurs, and what might be going on in the process?  The zombie processes described also seem to exhibit a high CPU utilisation.  I suspect vscode has tried to kill/restart the process and was unable (I think we need to look into more if we are launching the lsp in vscode in a way it can terminate it properly) when the process becomes detached.

Also, the ppid might still be running, but just not sending it messages anymore.  Obviously quitting and restarting vscode would eventually trigger this, but before someone does that to attempt to relive it, they will likely not notice it until they keep vscode running for a long time and then notice the high CPU.  Would some sort of client heartbeat make more sense?  Something that we flag on startup as the client supporting?  Then send a heartbeat message every 30 seconds and terminate if 1 or 2 of those are missed?  It feels a far more reliable way to me than detecting the ppid and determining if it is running or not.

dsherret:
@kitsonk heartbeat would be better, but it makes implementing an editor plugin more complicated so I think that's a big negative.

I think the high CPU usage might be something else.

kitsonk:
> I think the high CPU usage might be something else.

Well, I have had a couple occurrences of what is described in the issue referenced, where the deno process became detached and the CPU was at 100%.  No matter what, it is an error state, that indicates an underlying issue.  Just quitting the process doesn't actually fix the underlying issue nor help use detect it. 🙈

dsherret:
@kitsonk yeah, it doesn't and this isn't meant to fix that issue. This is just a part of preventing orphaned processes. For example, if vscode stops working and someone kills it then deno.exe will still be around.

piscisaureus (approved):
LGTM

@kitsonk I agree this does not fix the underlying issue but it should stop the flood of negative reviews due to leaving zombie processes around.

cli/lsp/parent_process_checker.rs
piscisaureus: This is the only portable way as far as I know.

ry:
We can land for 1.11.2 this despite it being a "feature" because it's internal and shouldn't have any effect on users.

PR #11031, opened 2021-06-18
fix(core): prevent include_js_files print "cargo:rerun" on release build

### LINKED ISSUE #11030: `deno_core` prints cargo related stuff if used as a dependency

maxxcs:
Resolves #11030

This PR changes `include_js_files` to only print `cargo:rerun` on debug build, since the cargo statement shouldn't be needed in release build.

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -126,6 +126,7 @@ macro_rules! include_js_files {
         Box::new(|| {
           let c = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
           let path = c.join($file);
+          #[cfg(debug_assertions)]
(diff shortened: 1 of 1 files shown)


littledivy:
This will prevent rebuilding of extensions in `--release`

maxxcs:
> This will prevent rebuilding of extensions in `--release`

It's inconvenient for `deno_land` users to have cargo things printed on the console. Maybe we can add this under a flag, like `cargo run --refresh-extensions` or something in that way to force rebuild extensions.

lucacasonato:
The reason you are seeing this is because you are not snapshotting your runtime code. See #11030 for details.

Issue #11037, opened 2021-06-18
Worker constructor should accept URL in TypeScript


bartlomieju:
Currently our `Worker` constructor accepts only `string` as a specifier type, however typing provided by TypeScript allows for `string | URL`:
```ts
declare var Worker: {
    prototype: Worker;
    new(stringUrl: string | URL, options?: WorkerOptions): Worker;
};
```

We already support using URLs when type checking is disabled, this is just a problem with declaration files.

Ref https://github.com/denoland/deno/issues/11035

lucacasonato:
Does `lib.dom.d.ts` allow this? I am hesitant to allow this as it is implicit conversion that breaks if a user overwrites `URL.toString()`. Also unhappy we did this for `new Request`.

Somewhat unrelated, but we have 0% WPT coverage for workers at the moment. I looked into it a few days ago, and it seems we'll need to do some refactoring. Especially in-light of eventual support for transferable and a future introduction of message ports.

surma:
> Does lib.dom.d.ts allow this? 

[It does](https://github.com/microsoft/TypeScript/blob/580bb066c1d5abdcafb0f09c1477b914defc0ac8/lib/lib.dom.d.ts#L18412-L18415).

lucacasonato:
Then 👍 from me. Marking as good first issue:

To resolve, change [this line](https://github.com/denoland/deno/blob/main/cli/dts/lib.deno.shared_globals.d.ts#L408) to `specifier: string | URL`, and add a test to https://github.com/denoland/deno/blob/main/cli/tests/unit/worker_types.ts that exercises this (copy the existing test and remove `.href`).

lucacasonato:
Oh I see you already fixed it @bartlomieju. I should sign off for the weekend 🤦‍♂️

PR #11038, opened 2021-06-18
fix: Worker accepts specifier as URL

bartlomieju:
This commit updates type declarations for Worker to accept specifiers
as either strings or URL, bringing it in line with TypeScript
declarations and browser behavior.

Fixes https://github.com/denoland/deno/issues/11037


lucacasonato (approved):
LGTM

PR #11040, opened 2021-06-18
feat: support SharedArrayBuffer sharing between workers

### LINKED ISSUE #6433: SharedArrayBuffer not works
Sample code looks like

main.js
```js
const u32Buf = new SharedArrayBuffer(2 * Uint32Array.BYTES_PER_ELEMENT);
const u32Arr = new Uint32Array(u32Buf);

const worker = new Worker(new URL("./worker.js", import.meta.url).href, { type: "module" });

u32Arr[0] = 1

worker.postMessage({
    buf: u32Buf,
});
```

worker.js
```js
self.onmessage = function (e) {
    const data = e.data;
    const u32Arr = new Uint32Array(data.buf);

    console.log(u32Arr[0])
};
```

The result is:
```shell
$ deno run --allow-read main.js
undefined                         
```
where I expect to get `1` instead of `undefined` 

Test in:
- Deno 1.1.1
- Windows 10

lucacasonato:
This commit adds support for sharing SABs between workers.

Tests for this are a bit lacklustre - I am hoping we get some more coverage "for free" when we enable web worker WPT.

Closes #6433


core/runtime.rs
```diff
@@ -96,6 +98,36 @@ struct ModEvaluate {
   sender: mpsc::Sender<Result<(), AnyError>>,
 }
 
+#[derive(Default, Clone)]
+pub struct TransferBuffer(Arc<Mutex<TransferBufferInner>>);
```
lucacasonato, line 102: This really isn't a great name. Open to bikeshedding :-)

lucacasonato: Renamed to `SharedArrayBufferStore`.

bartlomieju (approved):
LGTM, no comments, cleanly implemented. Let's wait on 1.12 before landing

CrimsonCodes0:
Might I suggest adding a test case to confirm that both, transferring nested SAB, and transferring multiple SAB work correctly? Passing an array of two SAB, and writing seperate values into each should be enough to knock out both test scenarios.

lucacasonato:
This can not be landed until we figure out all the side-effects of concurrent access to `rusty_v8::BackingStore`s. This is something which we have not had to deal with before, and is looking like it is exposing some safety issues in rusty_v8 that could result in undefined behavior.

The very naive way to "fix" this, is to disallow any and all passing of SAB backed Uint8Arrays to Rust.

mhmdmar:
@lucacasonato Do you think you can still land this on/before the 1.12.0 release ?

lucacasonato:
I am not sure.

mhmdmar:
@lucacasonato I will just wait for you until you do it, Thanks for the reply 👍

lucacasonato:
UB issues have been worked around in serde_v8. This will land for 1.12.

PR #11041, opened 2021-06-18
fix(lsp): do not rename in strings and comments

dsherret:
Fixes https://github.com/denoland/vscode_deno/issues/450

This functionality is not the default with TypeScript so I believe this is a bug and we shouldn't do it.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -1769,8 +1769,8 @@ impl Inner {
     let req = tsc::RequestMethod::FindRenameLocations((
       specifier,
       line_index.offset_tsc(params.text_document_position.position)?,
-      true,
-      true,
+      false,
+      false,
       false,
     ));
 
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -1766,13 +1766,14 @@ impl Inner {
         )));
       };
 
-    let req = tsc::RequestMethod::FindRenameLocations((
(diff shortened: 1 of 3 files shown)


cli/lsp/language_server.rs
```diff
@@ -1769,8 +1769,8 @@ impl Inner {
     let req = tsc::RequestMethod::FindRenameLocations((
       specifier,
       line_index.offset_tsc(params.text_document_position.position)?,
-      true,
-      true,
+      false,
+      false,
       false,
     ));
```
dsherret, line 1775: By looking at this code alone, it's hard to tell what this means. It would be nice if we didn't use tuples here.

These are the arguments for `find_in_comments` and `find_in_strings`.

kitsonk: I'd be fine using q struct with the more complicated APIs. It gets unwieldy for all of them though.

kitsonk (approved):
LGTM, but would be glad to have a struct here.

kitsonk:
Actually, so you think it warrents a test of some sort?

cli/lsp/language_server.rs
dsherret: Done in 8940d22.

dsherret:
@kitsonk Good point! Done in 87a3cd600fe0b1e0afdd55c06bd911f8c926bd92

kitsonk (approved):
LGTM

PR #11043, opened 2021-06-19
refactor: unify JavaScript script execution method

bartlomieju:
This commit renames "JsRuntime::execute" to "JsRuntime::execute_script". Additionally
same renames were applied to methods on "deno_runtime::Worker" and 
"deno_runtime::WebWorker".

A new macro was added to "deno_core" called "located_script_name" which
returns the name of Rust file alongside line no and col no of that call site.
This macro is useful in combination with "JsRuntime::execute_script"
and allows to provide accurate place where "one-off" JavaScript scripts
are executed for internal runtime functions.

Co-authored-by: Nayeem Rahman <nayeemrmn99@gmail.com>


cli/main.rs
```diff
@@ -558,9 +558,15 @@ async fn eval_command(
   program_state.file_fetcher.insert_cached(file);
   debug!("main_module {}", &main_module);
   worker.execute_module(&main_module).await?;
-  worker.execute("window.dispatchEvent(new Event('load'))")?;
+  worker.execute_script(
+    "[native code]",
+    "window.dispatchEvent(new Event('load'))",
```
nayeemrmn, line 563: What do you think of something like [`std::file!()`](https://doc.rust-lang.org/std/macro.file.html) for JS inlined in rust?
```suggestion
  runtime.execute_script(&format!("[deno:{}].js", std::file!()), &init_src)?;
```
image

bartlomieju: Yeah, that was my first thought but I didn't want to over do it. 

I'll be fine even with copy-pasting the strings, don't think a couple dozen call sites across 3 modules need a macro

cli/tests/integration_tests.rs
```diff
@@ -5540,25 +5540,6 @@ console.log("finish");
...
-      .stderr(std::process::Stdio::piped())
-      .spawn()
-      .unwrap()
-      .wait_with_output()
-      .unwrap();
-    assert!(!output.status.success());
-    let stderr = std::str::from_utf8(&output.stderr).unwrap().trim();
-    assert!(stderr.contains("Failed to get current working directory"));
-  }
-
```
ry, line 5561: Intentional?

cli/main.rs
ry: I think it's a good idea maybe something like 
```
format!("[deno-internal:{}]", std::file!())
```

ry: execute_script could even become a macro, so this code wasn't repeated everywhere...

nayeemrmn: > ```
> format!("[deno-internal:{}]", std::file!())
> ```

It can just be `deno:`, the situation matches exactly how we use this protocol: `deno:<repo-relative path to source file>`.

cli/tests/integration_tests.rs
bartlomieju: It is, this test was actually testing `Worker::execute()` that was reading CWD without good reason.

cli/main.rs
bartlomieju: Well to make it consistent with code in extensions it should be `deno:cli/main.rs` and I think that's a good solution - it will be obvious that this code is triggered from Rust. 

So I guess full suggestion would be to:
```rust
worker.execute_script(
  format!("deno:", std::file!()),
  "window.dispatchEvent(new Event('load'))"
);
```

Let me try to implement that and see how it goes.

> execute_script could even become a macro, so this code wasn't repeated everywhere...

-1 on this one - `Worker` and `WebWorker` are in `runtime` crate that is used by 3rd parties that embed deno - it's not necessary true they want to have such filename, instead they might want to use custom name.

nayeemrmn: > Well to make it consistent with code in extensions it should be `deno:cli/main.rs`

There should be some kind of separator at the end so the path is cut off from wrongly contextualised line/column numbers in the event of an error. That's why I think square brackets are a good idea:
image
image

bartlomieju: I just checked and `std::file!()` appends leading slash:
```
let a = format!("deno:/{}", std::file!());
eprintln!("file {}", a);
```
```
file deno:/cli/main.rs
```

So I think I'll just go with strings for now.

> There should be some kind of separator at the end so the path is cut off from wrongly contextualised line/column numbers in the event of an error. That's why I think square brackets are a good idea:

I'm open to that, though it doesn't bother me that much - in either situation you won't be able to lookup this code using column and line anyway (line will always be 1, as we only execute onliners).

nayeemrmn: > I just checked and `std::file!()` appends leading slash:
> 
> ```
> let a = format!("deno:/{}", std::file!());
> ```

Looks like the leading slash is yours? 😅

bartlomieju: 🤦

bartlomieju: Alright, I reworked it to use `std::file!()` haven't added curly braces. I'm open to doing that, @ry wanna break the tie on this one?

nayeemrmn: > I'm open to that, though it doesn't bother me that much - in either situation you won't be able to lookup this code using column and line anyway (line will always be 1, as we only execute onliners).

That's why I think it shouldn't mislead with a connected line number.

However, I think we can give useful location info. I didn't suggest it because it seemed like a long line to repeat but we can do it with a macro:
```rust
macro_rules! located_script_name {
  () => {
      format!("[deno:{}:{}:{}]", std::file!(), std::line!(), std::column!());
  };
}

// ...

worker.execute_script(
  &located_script_name!(),
  "window.dispatchEvent(new Event('load'))"
);
```
The location points to the initializer of the script name but it's genuinely useful.

bartlomieju: Thanks, I added this macro to `deno_core` and reexported it from there so both `deno_runtime` and `deno_cli` crates can use it.

core/lib.rs
```diff
@@ -90,6 +90,22 @@ pub fn v8_version() -> &'static str {
...
 
+/// A helper macro that will return a call site in Rust code. Should be
+/// used when executing internal one-line scripts for JsRuntime lifecycle.
+///
+/// Returns a string in form of: "[deno:<filename>:<line>:<column>]"
+#[macro_export]
+macro_rules! located_script_name {
+  () => {
+    format!(
+      "[deno:{}:{}:{}]",
```
ry, line 101: I think Nayeem was suggesting the close bracket happens before the line number?

ry (approved):
Looks good. 

maybe this should wait for 1.12?

core/lib.rs
nayeemrmn: > I think Nayeem was suggesting the close bracket happens before the line number?

No, that was referring to JS error line numbers, which would come after whatever name we put here. These rust line numbers are correct when put next to the path.

nayeemrmn (approved):
LGTM 👍

image
image

bartlomieju:
> LGTM 👍
> 
> image
> image

Yeah, I agree providing this path is much better than `[native code]`. Thanks for helping pan this out!

PR #11047, opened 2021-06-19
docs(runtime): fix HTTP server example

vwkd:
Follow up on #10497 which seemingly missed one fix.


bartlomieju (approved):
LGTM, thank you @vwkd

PR #11050, opened 2021-06-19
refactor(FormData/Serializer): Early fix for supporting async blob source

jimmywarting:
This new FormData Serializer takes a FormData and converts it into a new Blob with other preexisting Blob parts 

The benefits
- Less code
- Don't need to calculate the byte length (Content-Length) - you get it from the new blob instead
- Don't have to read the content of the File entries or keeping any of the File entry content in memory until it's necessary to read them after `blob.stream()` is called, hence more RAM friendly
- Still synchronous
  - Good for when you eg have to clone a Response `new Response(new FormData()).clone()`? but why would you? have have no idea, still a valid case?
- partially solves a blocker of async blob sources (#10969)

#### Other considerations/possible solutions: 
Returning a ReadableStream or async iterator... Converting it to a blob seems to be the most easiest.
I have used this formdata-to-blob technique in my [formdata-polyfill](https://github.com/jimmywarting/FormData) for years by converting it into a blob since IE9 was a thing in the past

---

The old Serializer did read the content from File Entries using `_byteSequence`. That is no longer going to be supported once we have async blob sources that can originate from the filesystem and don't hold the data in memory

### DIFF
--- a/extensions/fetch/21_formdata.js
+++ b/extensions/fetch/21_formdata.js
@@ -13,7 +13,7 @@
 ((window) => {
   const core = window.Deno.core;
   const webidl = globalThis.__bootstrap.webidl;
-  const { Blob, File, _byteSequence } = globalThis.__bootstrap.file;
+  const { Blob, File } = globalThis.__bootstrap.file;
 
   const entryList = Symbol("entry list");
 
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM, thanks. Nice simplification.

PR #11051, opened 2021-06-19
feat: `MessageChannel` and `MessagePort`

### LINKED ISSUE #6691: Worker MessageChannel
Seems that `MessageChannel` in missing from Deno APIs.

https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel

It should be added as it allows safe communication between `Workers`

lucacasonato:
This commit introduces support for MessageChannel and MessagePort.
MessagePorts can be transfered across a message channel.

This PR does not support transfering ports between workers yet. I will
tackle moving over workers, and broadcast channel to use the same
transfer logic used here in a follow up.

Closes #6691

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -781,6 +781,7 @@ dependencies = [
  "encoding_rs",
  "futures",
  "serde",
+ "tokio",
  "uuid",
 ]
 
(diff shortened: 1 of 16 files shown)


bartlomieju (approved):
LGTM, cleanly implemented. It's unfortunate that ports and channels will leak in tests unless you close them explicitly, but that's already a case for some other Web APIs.

Let me know if you need help integrating into workers.

PR #11054, opened 2021-06-20
docs(core): replaces dispatch references with op

maxxcs:
The README had confused me while I was studying `deno_core`


bartlomieju (approved):
Thank you @maxxcs, LGTM

PR #11062, opened 2021-06-21
feat: Deno namespace configurable and unfrozen

### LINKED ISSUE #11033: Consider making `Deno` namespace configurable
Currently the `Deno` namespace isn't configurable in the runtime.  This makes it difficult to do mocking and polyfilling of things, like trying to create a realistic Deploy environment, where code could bind to the parts of the `Deno` namespace it needs and then replace the namespace with a different set of APIs for additional code to be loaded.

Other global objects are not set as non-configurable and so `Deno` stands out as an example.

A specific use case is when trying to create a Deploy like environment in the worker, the only option is to not include the `Deno` namespace in the worker and create global `Deno` object that has the APIs that are just present in Deploy, but this means that the code in the worker that bootstraps the environment has no access to the other `Deno` APIs, which makes it difficult to do things like read files from disk.

kitsonk:
Closes #11033

### DIFF
--- a/cli/tests/unit/globals_test.ts
+++ b/cli/tests/unit/globals_test.ts
@@ -62,8 +62,8 @@ unitTest(function DenoNamespaceEqualsWindowDeno(): void {
   assert(Deno === window.Deno);
 });
 
-unitTest(function DenoNamespaceIsFrozen(): void {
-  assert(Object.isFrozen(Deno));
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM. Let's wait on 1.12 before landing this one.

PR #11069, opened 2021-06-21
fix(runtime): fix signal promise API

### LINKED ISSUE #9806: SignalStream#dispose triggers then onfulfilled
This code triggers the callback, with no way of knowing if the signal actually appeared, or if the stream was disposed.
```ts
const s = Deno.signal(Deno.Signal.SIGABRT);
s.then(() => {
  console.log("I found a SIGABRT");
});
s.dispose();
```

kt3k:
This PR fixes the wrong detection of signal event in Promise API when disposing the signal stream.

closes #9806

### DIFF
--- a/cli/tests/unit/signal_test.ts
+++ b/cli/tests/unit/signal_test.ts
@@ -154,6 +154,35 @@ unitTest(
   },
 );
 
+// https://github.com/denoland/deno/issues/9806
+unitTest(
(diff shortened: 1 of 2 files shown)


piscisaureus (approved):
LGTM 
(I still think an event based API would be a great improvement)

kt3k:
@piscisaureus I'll open a proposal issue in a week. (Still thinking about the interface)

PR #11070, opened 2021-06-21
fix(extensions/fetch): Encode and decode headers as byte strings.

### LINKED ISSUE #10991: Headers should be treated as byte strings
Currently we cast strings to UTF-8 all over the place. They should instead be transferred as byte strings from Rust into V8 and back. This means we probably need to add a `ByteString` magic type to `serde_v8` that stores the inner string as a byte vector.

andreubotella:
Depends on denoland/rusty_v8#704 and denoland/serde_v8#2.

Fixes #10991.


lucacasonato (approved):
LGTM! Thanks :-)

We should apply the same changes to runtime/ops/http.rs for the HTTP server.

PR #11076, opened 2021-06-21
feat: transfer MessagePort between workers

### LINKED ISSUE #11074: Transferables should be supported for worker postMessage
Followup to #6691. The `postMessage` / `onmessage` for workers should use the same internal logic as `MessagePort`.

lucacasonato:
Add support for transferring MessagePorts between workers.

Closes #11074

There is a test for transferring ports between workers, but edge cases are probably not covered. We will get more coverage with the introduction of worker WPTs, but that is still blocked on our exclusive support for module workers.


runtime/js/11_workers.js
```diff
@@ -219,21 +217,16 @@
       return handled;
     }
 
-    #poll = async () => {
+    #pollCtrl = async () => {
```
bartlomieju, line 220: Nitpick: `pollControl`, saving 3 letters for brevity is not worth it ;)

runtime/ops/worker_host.rs
```diff
@@ -51,8 +51,9 @@ pub struct CreateWebWorkerArgs {
   pub use_deno_namespace: bool,
 }
 
-pub type CreateWebWorkerCb =
-  dyn Fn(CreateWebWorkerArgs) -> WebWorker + Sync + Send;
+pub type CreateWebWorkerCb = dyn Fn(CreateWebWorkerArgs) -> (WebWorker, SendableWebWorkerHandle)
```
bartlomieju, line 54: Could you keep previous signature and expose a method on the worker that could return that handle?

runtime/web_worker.rs
```diff
@@ -56,29 +57,27 @@ impl WorkerId {
   }
 }
 
-type WorkerMessage = ZeroCopyBuf;
-
 /// Events that are sent to host from child
 /// worker.
-pub enum WorkerEvent {
-  Message(WorkerMessage),
+pub enum WorkerControlEvent {
```
bartlomieju, line 62: `Serialize` could probably be implemented here instead of `worker_host.rs` - I know you didn't change that, but I'd appreciate if you could sort this out

runtime/web_worker.rs
```diff
@@ -122,40 +116,45 @@ impl WebWorkerInternalHandle {
...
   }
+}
 
+#[derive(Clone)]
+pub struct WebWorkerHandle {
+  pub port: Rc<MessagePort>,
+  receiver: Rc<RefCell<mpsc::Receiver<WorkerControlEvent>>>,
+  terminated: Arc<AtomicBool>,
+  isolate_handle: v8::IsolateHandle,
+}
```
bartlomieju, line 143: I'm I little confused, why do we need three different handles?

runtime/web_worker.rs
```diff
@@ -524,132 +519,3 @@ pub fn run_web_worker(
...
   result
 }
-
-#[cfg(test)]
-mod tests {
-  use super::*;
-  use crate::tokio_util;
-
-  fn create_test_web_worker() -> WebWorker {
-    let main_module = deno_core::resolve_url_or_path("./hello.js").unwrap();
```
bartlomieju, line 534: Why remove these tests?

runtime/web_worker.rs
```diff
@@ -56,29 +57,27 @@ impl WorkerId {
...
   TerminalError(AnyError),
   Close,
 }
 
 // Channels used for communication with worker's parent
 #[derive(Clone)]
-pub struct WebWorkerInternalHandle {
-  sender: mpsc::Sender<WorkerEvent>,
-  receiver: Rc<RefCell<mpsc::Receiver<WorkerMessage>>>,
+pub struct WebWorkerSelfHandle {
```
bartlomieju, line 70: I think the previous name was more descriptive

runtime/ops/worker_host.rs
lucacasonato: No, because SendableWebWorkerHandle is not Clone. The reason I split it out was what it having to be clone greatly complicated all kinds of things.

runtime/web_worker.rs
lucacasonato: One is Send, the other one is Clone. Having the same struct do both requires mutexes, which were complicating the code and were causing various issues.

runtime/web_worker.rs
lucacasonato: Because the test can not be implemented in the new design without a reference to the parent workers' OpState, but there is no parent worker in these tests. The scenarios are all tested via js unit tests already.

runtime/web_worker.rs
lucacasonato: The `SendableWebWorkerHandle` is transferred from the worker thread to the parent thread, where it is turned into a regular `WebWorkerHandle`.

runtime/web_worker.rs
lucacasonato: Changed it back.

runtime/web_worker.rs
bartlomieju: Okay sounds good, please update docstrings for those structs and it's good for me to land

bartlomieju (approved):
LGTM, pending https://github.com/denoland/deno/pull/11076#discussion_r656228701

Issue #10557, opened 2021-05-10
Increase maximum webstorage size


crowlKats:
Currently there is a hard cap of 5MB for both local & session storage. Browsers have increased this to 10MB years ago.
I think we should implement https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria#storage_limits, as FF will soon use the same logic for localstorage, but the idea of deno's localstorage using up half of someone's disk space doesnt thrill me, though theoretically current localstorage could take up the entiire disk (all origins combined)

sebastienfilion:
In my opinion, it can be augmented to 10Mb. But also think we should consider a way to let the user arbitrarily set the maximum, maybe with a flag.

getspooky:
I think it would be a good idea to provide the `max Limit` option and let the user specify how much space they want to reserve.

bartlomieju:
Let's just increase size limit to 10Mb and call it a day? Why would we want to make this option configurable if it's not configurable in the browser?

crowlKats:
agreed that this should not be configurable by the user whatsoever.
i am just not sure if 10mb or the FF storage criteria would be better (or maybe a different fixed value?); 10mb is more than enough on the browser, but server side i can imagine one using a lot more

caspervonb:
> Let's just increase size limit to 10Mb and call it a day? Why would we want to make this option configurable if it's not configurable in the browser?

Agreed.

With eviction and persistence it gets incredibly convoluted, even understanding it as a user is a fairly major headache.
We can come back to it later maybe but for now I'd be in favour of a simple bump to the size and call it good.

getspooky:
Sure , The maximum browser storage space is dynamic — it is based on your hard drive size. The global limit is calculated as 50% of free disk space. But currently the most simpler solution is to make `10MB` instead of `5MB`.
what do you think ?

lucacasonato:
Lets bump to 10MB and call it a day.

PR #11080, opened 2021-06-22
feat(extensions/webstorage): Allow user to specify the amount of space to use for `webstorage`

getspooky:
I think it would be a good idea to let the user specify how much space they want to reserve,
By default deno take `5MB`.

## Example:

```ts
localStorage.setLimitStorage(80000);
```

### DIFF
--- a/extensions/webstorage/01_webstorage.js
+++ b/extensions/webstorage/01_webstorage.js
@@ -30,6 +30,10 @@
       return core.opSync("op_webstorage_key", index, this[_persistent]);
     }
 
+    setLimitStorage(size) {
+      this.size = size;
+    }
+
     setItem(key, value) {
       webidl.assertBranded(this, Storage);
       const prefix = "Failed to execute 'setItem' on 'Storage'";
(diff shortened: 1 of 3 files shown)


lucacasonato (changes requested):
This is not a web standard API. Further discussion is needed in the issue.

getspooky:
Yes , please see https://github.com/denoland/deno/issues/10557

crowlKats:
@getspooky what Luca meant is that this is not a viable solution because this is changing a standard Web API. This needs more discussing than just a single comment.

In regards to the actual changes: this makes having a limit completely useless because anyone can change it during runtime; at that point might as well just remove a limit whatsoever.

exside:
Not happy with the cap, can't use Deno.kv() because of another bug (memory leak in v8) which prevents me from upgrading Deno to a version that has Deno KV...so I'm stuck writing dumb JSON files (or adding an entire DB to the stack, which I don't want) instead of being able to access the SQLite backed WebStorage API... I agree that it should not be a non-standard API method like suggested here, but it should be a runtime flag like `--storage-limit=<MB>`, don't understand why this wasn't done, would it be very hard? I mean yeah, not really changeable in the browser, but in a browser you don't do the same things as on a server...so I don't think that argument counts 😉

PR #11081, opened 2021-06-22
fix(extensions/webstorage): Increase localstorage to 10MB

getspooky:
Increate localstorage from `5MB` to `10MB`.

Closes https://github.com/denoland/deno/issues/10557

### DIFF
--- a/extensions/webstorage/lib.rs
+++ b/extensions/webstorage/lib.rs
@@ -15,6 +15,8 @@ use std::path::PathBuf;
 #[derive(Clone)]
 struct OriginStorageDir(PathBuf);
 
+const MAX_STORAGE_BYTES: u32 = 10 * 1024 * 1024;
+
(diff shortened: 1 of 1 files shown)


extensions/webstorage/lib.rs
```diff
@@ -136,7 +136,7 @@ pub fn op_webstorage_set(
     conn.prepare("SELECT SUM(pgsize) FROM dbstat WHERE name = 'data'")?;
   let size: u32 = stmt.query_row(params![], |row| row.get(0))?;
 
-  if size >= 5000000 {
+  if size >= 10000000 {
```
lucacasonato, line 139: Can you turn this into a `const MAX_STORAGE_BYTES = 10 * 1024 * 1024;` at the top of the file?

getspooky: done

lucacasonato (approved):

lucacasonato:
Thanks!

PR #11082, opened 2021-06-22
cleanup(core): top-level-await is now always enabled

punkeel:
Starting with V8 9.1, top-level-await is always enabled by default.
See https://v8.dev/blog/v8-release-91 for the release notes.

- Remove the now redundant v8 flag.
- Clarify doc comment and add link to the feature explainer.


bartlomieju (approved):
Perfect, thanks @punkeel!

PR #11083, opened 2021-06-22
chore(lsp): switch to use lsp to get parent process id

dsherret:
After seeing the vscode PR, @kitsonk pointed out we should use some custom messaging in the lsp to pass the parent process id and this made sense to me as the `--parent-pid <pid>` flag was inconvenient because it required us to version sniff to ensure we didn't pass it to old versions of the executable (something I hadn't considered).

While investigating moving it to the LSP, I discovered that the parent PID is already passed to us for free (https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#initializeParams) 🙈

This will be much better and doesn't require any changes in vscode_deno. It's unfortunate we didn't notice this earlier for the patch release yesterday though.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -84,9 +84,7 @@ pub enum DenoSubcommand {
     root: Option<PathBuf>,
     force: bool,
   },
-  Lsp {
-    parent_pid: Option<u32>,
-  },
+  Lsp,
   Lint {
     files: Vec<PathBuf>,
     ignore: Vec<PathBuf>,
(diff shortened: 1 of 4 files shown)


kitsonk (approved):
LGTM and I apologise for not thinking the whole thing through to begin with.

dsherret:
@kitsonk that's ok! I should have done more research initially. The important thing is that we ended on the right solution 🙂

I'm glad we have something that works well for all editors.

PR #11086, opened 2021-06-22
feat(repl): support import declarations in the REPL

### LINKED ISSUE #10955: Import declarations in the REPL
This was previously discussed in #1285 and the suggestion was to use dynamic imports. Taking some code that has import declarations and getting it to work in the REPL currently requires a manual transformation to dynamic imports though, which is not so convenient. This could be automatic.

I have a branch with a rough implementation of this that could be cleaned up then merged after #10934.

Related: #7425

dsherret:
This allows users to copy and paste more code into the REPL and it should just work:

image

Even though I would much rather prefer these to act more like `let` declarations in the REPL, this was implemented to be consistent with class, function, and const variable declarations for consistency (in the future, it may be nice to relax this restriction):

image

Some error message examples:

image

Note that export declarations are not handled and they will still error.

Closes #10955.


cli/ast/mod.rs
```diff
@@ -43,6 +43,8 @@ use swc_ecmascript::transforms::react;
 use swc_ecmascript::transforms::typescript;
```
dsherret, line 43: Note that I moved this file... hopefully that's ok, but adding the transforms iside the `ast.rs` file would have started to get really busy.

bartlomieju: Looks fine for me 👍

bartlomieju (approved):
Discussed offline with @ry and we are both in favor of this change - usability outweights divergence from standards.

caspervonb (approved):

PR #11089, opened 2021-06-23
fix(core/modules): Fix concurrent loading of dynamic imports

### LINKED ISSUE #3736: module resolve failed with dynamic import and circulated import

nayeemrmn:
Fixes #3736.

The following guard: https://github.com/denoland/deno/blob/477273085f8e5f578ea3ee12c2183c44133af05d/core/modules.rs#L583 is the cause of the bug. It causes `RecursiveModuleLoad` to not recurse a module `foo`'s dependencies if it is already in the module map, causing the load to complete early, and letting the runtime try to instantiate `foo`. But if there is another pending load of `foo`, it may be the case that `foo` is in the module map but its dependencies aren't yet, which will cause instantiation to fail. So we should always visit all dependencies in a load and check registration for each module instead.

cc @bartlomieju


core/modules.rs
```diff
@@ -579,11 +585,8 @@ impl ModuleMap {
     let imports = self.get_children(module_id).unwrap().clone();
 
     for module_specifier in imports {
-      let is_registered = self.is_registered(&module_specifier);
-      if !is_registered {
-        load
-          .add_import(module_specifier.to_owned(), referrer_specifier.clone());
-      }
+      load
+        .visit_import(module_specifier.to_owned(), referrer_specifier.clone());
```
bartlomieju, line 589: I'm afraid this will cause multiple loads of the same module, which might lead to race conditions

bartlomieju (commented):
@nayeemrmn great to see you looked into it. I've got some reservations that even though it fixes provided test case it might not be perfect solution.

Could you please use test case from https://github.com/denoland/deno/pull/8421 (the unit test case to put in `deno_core` instead of integration one). I'm afraid we might have multiple calls to `load()` for the same module (would be good to assert number of the loads). 

I guess this is a good way forward but ultimately we should keep track of all visited modules in `ModuleMap` instead of `RecursiveModuleLoad`.

core/modules.rs
```diff
@@ -299,39 +300,93 @@ impl RecursiveModuleLoad {
...
+              &specifier,
+              Some(referrer.clone()),
+              self.is_dynamic_import(),
+            );
+            self.pending.push(fut.boxed_local());
+          }
+          self.visited.insert(specifier);
+        }
+      }
+    }
```
nayeemrmn, line 378: @bartlomieju See reworked code. Modules which are already in the module map are no longer reloaded, they are recursed eagerly and synchronously instead.

bartlomieju:
I applied my unit test from https://github.com/denoland/deno/pull/8421 and it shows that there are 7 module loads happening. I think this is correct so we should be able to land it. But before that, please rework the integration test into unit test in `core/modules.rs` (or take it directly from https://github.com/denoland/deno/pull/8421)

core/modules.rs
```diff
@@ -371,10 +426,16 @@ impl Stream for RecursiveModuleLoad {
...
           LoadInit::Main(..) | LoadInit::DynamicImport(..) => inner
             .loader
-            .load(inner.op_state.clone(), &module_specifier, None, false)
+            .load(
+              inner.op_state.clone(),
+              &module_specifier,
+              None,
+              inner.is_dynamic_import(),
+            )
             .boxed_local(),
```
nayeemrmn, line 435: Oops, the root load here isn't guarded by a module map check.

nayeemrmn: Okay I made some refactors (namely keeping a module map reference in `RecursiveModuleLoad::module_map_rc`) before realising the approach I had in mind didn't work. Went with a [different way](https://github.com/denoland/deno/pull/11089/files#diff-aa01489135c068a2056422f99878ba91861c585baed075470c08c1ac439ab3b2R418-R419) but I want to keep the refactor.

Also removed `LoadInit::MainWithCode(..)` by instead preloading code into the module map when synthetic module code is given.

core/modules.rs
```diff
@@ -202,91 +200,102 @@ pub enum LoadState {
...
 /// This future is used to implement parallel async module loading.
 pub struct RecursiveModuleLoad {
-  op_state: Rc<RefCell<OpState>>,
   init: LoadInit,
   // TODO(bartlomieju): in future this value should
   // be randomized
   pub id: ModuleLoadId,
   pub root_module_id: Option<ModuleId>,
   pub state: LoadState,
+  pub module_map_rc: Rc<RefCell<ModuleMap>>,
```
bartlomieju, line 209: Aha, nice!

core/modules.rs
```diff
@@ -299,39 +308,89 @@ impl RecursiveModuleLoad {
...
+    // 2. If the module is already in the module map, queue it up to be
+    //    recursed synchronously here.
+    // This robustly ensures that the whole graph is in the module map before
+    // `LoadState::Done` is set.
+    let specifier =
+      crate::resolve_url(&module_source.module_url_found).unwrap();
+    let mut already_registered = VecDeque::new();
+    already_registered.push_back((module_id, specifier.clone()));
+    self.visited.insert(specifier);
+    while let Some((module_id, referrer)) = already_registered.pop_front() {
```
bartlomieju, line 357: Nice! I believe this is the meat of this PR, very elegantly solved 👍

core/modules.rs
```diff
@@ -343,36 +402,45 @@ impl Stream for RecursiveModuleLoad {
...
-            .boxed()
-          }
-          LoadInit::Main(..) | LoadInit::DynamicImport(..) => inner
+        let load_fut = if let Some(_module_id) = inner.root_module_id {
+          // The root module is already in the module map.
+          // TODO(nayeemrmn): In this case we would ideally skip to
+          // `LoadState::LoadingImports` and synchronously recurse the imports
+          // like the bottom of `RecursiveModuleLoad::register_and_recurse()`.
+          // But the module map cannot be borrowed here. Instead fake a load
+          // event so it gets passed to that function and recursed eventually.
```
bartlomieju, line 419: We can address this bit when adding support for import assertions

core/modules.rs
```diff
@@ -1271,6 +1278,80 @@ mod tests {
     })
   }
 
+  // Regression test for https://github.com/denoland/deno/issues/3736.
+  #[test]
+  fn dyn_concurrent_circular_import() {
```
bartlomieju, line 1283: Thanks

bartlomieju (approved):
Looks good to me 🚀  

Thank you Nayeem, this problem has been haunting me for the past half a year. This is a very elegant solution to the problem, and it didn't require rewriting all of the module loading subsystem. Cleanly implemented with plenty of comments. Cheers!

PR #11093, opened 2021-06-23
fix(extensions/fetch): Creating a Request from an existing Request should proxy its body

### LINKED ISSUE #10993: Body disturbed state is not checked when creating Request from existing Request
See https://staging.wpt.fyi/results/fetch/api/request/request-init-stream.any.html?label=master&product=deno&product=servo&product=chrome

andreubotella:
Additionally, if the existing `Request`'s body is disturbed, the `Request` creation should fail. This fixes #10993.

This change also updates the step numbers in the `Request` constructor to match whatwg/fetch#1249.


lucacasonato (approved):
LGTM, thanks!

PR #11094, opened 2021-06-23
docs(core): Use op name in docs example

punkeel:
The first argument to opSync/opAsync is the op name. In the examples, the name is 'hello', and so it should be the first argument.


bartlomieju (approved):
LGTM

PR #7118, opened 2020-08-19
refactor: migrate ops to new dispatch wrapper

bartlomieju:

PR #11097, opened 2021-06-24
fix(runtime/signal): use op_async_unref for op_signal_poll

### LINKED ISSUE #10658: Signal API blocks exiting the process
Because Signal op uses `AsyncUnref` op internally, it shouldn't block process exiting. But since around 1.4.0 it seems blocking the program exiting.

The following code exits immediately with deno <=v1.3.0, but it never exits with >=v1.4.0:

```ts
Deno.signal(Deno.Signal.SIGABRT).then(() => { console.log("SIGABRT"); });
```

kt3k:
This PR restores the behavior of Deno.signal API. Signal API originally didn't block the runtime from exiting the process, but since #7118 it uses async (non-unref) op, which blocks the runtime from exiting.

This PR adds `op_async_unref` utility which registers the op as AsyncUnref op, and op_signal_poll uses it.

Added a test case which demonstrate that the runtime exits while for-await loop still awaiting the signals.

closes #10658

### DIFF
--- a/cli/tests/raw_mode_cbreak.ts
+++ b/cli/tests/raw_mode_cbreak.ts
@@ -5,11 +5,13 @@ const signal = Deno.signals.interrupt();
 
 Deno.stdout.writeSync(new TextEncoder().encode("S"));
 
-await signal;
+signal.then(() => {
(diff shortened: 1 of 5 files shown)


bartlomieju (approved):
LGTM

PR #11101, opened 2021-06-24
fix(extensions/url): Allow arbitrary USVStrings in the URLSearchParams constructor

andreubotella:


lucacasonato (approved):
Whoops - good catch. Thanks :-)

PR #11103, opened 2021-06-24
fix: MessagePort in message for postMessage transfers

### LINKED ISSUE #11102: MessagePorts transferred via postMessage can not be included in structured cloned message
Ports are only available on `MessageEvent.ports`, but not inside the structured cloned message body.

lucacasonato:
Fixes #11102

~~Blocked on https://github.com/denoland/rusty_v8/pull/712 and a rusty_v8 release.~~

### DIFF
--- a/cli/tests/unit/message_channel_test.ts
+++ b/cli/tests/unit/message_channel_test.ts
@@ -31,3 +31,29 @@ Deno.test("messagechannel", async () => {
   mc.port2.close();
   mc2.port2.close();
 });
+
+Deno.test("messagechannel clone port", async () => {
+  const mc = new MessageChannel();
+  const mc2 = new MessageChannel();
+  assert(mc.port1);
+  assert(mc.port2);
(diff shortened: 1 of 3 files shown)


lucacasonato:
Blocked on #11118. Will revisit once we have a solution for that.

lucacasonato:
This PR has now turned into a non general fix for the issue, instead only fixing `MessagePort`. That can be done because `MessagePort` can not be constructed (it is always created via a `MessageChannel`). The general fix that would let us structured clone for example `Blob`s is still blocked on #11118.

I suggest you review this as two seperate commits. The first one introduces `Deno.core.createHostObject`, the second one allows `MessagePort` objects in value serialization.

bartlomieju (approved):
LGTM. This is not a perfect solution, but fixes the problem at hand. And considering `Deno.core` is not a public API this shouldn't be a big problem.

Issue #10808, opened 2021-05-31
lsp: ignore empty `deno.config` option


kitsonk:
Okay, now that I know how to look at the output, I *think* I figured out what was happening.

When I reload the window and it's not working, I get this output:
```
Starting Deno language server...
  version: 1.10.2 (release, x86_64-apple-darwin)
Connected to "VSCodium" 1.56.2
Updating TypeScript configuration from: ""
Updating tsconfig has errored: Failed to load configuration file
language server initialized
Server ready.
```
This is even though I have not set config anywhere. _But_ if I hit "reset setting" on the config variable and reload the window (restarting the server alone seems inconsistent?), the problem goes away.

My theory is that at some point I wrote something in config, and then deleted it, but the extension still sees an empty string instead of null. Like, it's trying to load `${root}/${config}` when `config === ''` instead of ignoring the config variable.

In fact, when I go to config in settings, add a space, then delete it, this is the output:
```
Starting Deno language server...
  version: 1.10.2 (release, x86_64-apple-darwin)
Connected to "VSCodium" 1.56.2
language server initialized
Server ready.
Updating TypeScript configuration from: " "
Updating TypeScript configuration from: ""
```

Here is my `.vscode/settings.json`. I'm now able to toggle the issue on and off with just the empty string thing and reloading the window (I deleted the `tsconfig.json`).
```json
{
    "deno.enable": true,
    "deno.lint": true,
    "deno.unstable": true
}
```

_Originally posted by @Buraburaite in https://github.com/denoland/vscode_deno/issues/318#issuecomment-851675891_

Buraburaite:
Rust is hard. After about a dozen different tries, this is the only solution that I could get to compile:

```rust
match workspace_settings.config {
    Some(ref str) => if str == "" { (None, config.root_uri.clone()) } else { (workspace_settings.config, config.root_uri.clone()) }
    None => (workspace_settings.config, config.root_uri.clone())
}
```

I would prefer to write something like this, but I can't get it compile (error is `no method named "is_empty" found for enum "std::option::Option<std::option::Option<std::string::String>>" in the current scope`:
```rust
if Some(workspace_settings.config).is_empty() { workspace_settings.config = None; }
```

How I (kind of) tested It
---
1. Fork deno
2. Run `cargo run --- run test.js`, where test.js is `console.log('hello world');`

[...]

PR #11104, opened 2021-06-24
fix(lsp): handle invalid config setting better

### LINKED ISSUE #11100: lsp: `deno.config` set to empty string causes strange errors
In VSCode if someone changes the global setting `deno: Config` then deletes the text, VSCode replaces the value in the global `settings.json` with an empty string causing VSCode to see things such as `top-level await` and `console.log` as errors.

image

To my knowledge the only way to fix this is to edit the global `settings.json` by hand and removing the `deno.config` field entirely which is not intuitive at all.

kitsonk:
Fixes #11100

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -324,6 +324,59 @@ impl Inner {
     Ok(navigation_tree)
   }
 
+  fn merge_user_tsconfig(
+    &mut self,
+    maybe_config: &Option<String>,
+    maybe_root_uri: &Option<Url>,
+    tsconfig: &mut TsConfig,
+  ) -> Result<(), AnyError> {
+    self.maybe_config_file = None;
+    self.maybe_config_uri = None;
+    if let Some(config_str) = maybe_config {
+      if !config_str.is_empty() {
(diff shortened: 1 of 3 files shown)


cli/lsp/language_server.rs
```diff
@@ -324,6 +324,59 @@ impl Inner {
...
+          Url::from_file_path(config_path).map_err(|_| {
+            anyhow!("Bad file path for configuration file: \"{}\"", config_str)
+          })
+        } else {
+          Err(anyhow!(
+            "The path to the configuration file (\"{}\") is not resolvable.",
+            config_str
+          ))
+        }?;
+        info!("  Resolved configuration file: \"{}\"", config_url);
```
satyarohith, line 354:
```suggestion
        info!("Resolved configuration file: \"{}\"", config_url);
```

satyarohith (approved):
LGTM. Also fixes #10808

cli/lsp/language_server.rs
kitsonk: The spacing is intentional, as it occurs after the value is read. The spacing was already there, it was just the code refactored to handle errors better.

PR #11106, opened 2021-06-24
fix: make readonly `Event` properties readonly

dsherret:
Fix the `Event` class so its readonly properties are actually readonly.

Refs:

* https://developer.mozilla.org/en-US/docs/Web/API/Event
* See https://github.com/denoland/deno/pull/10979#discussion_r652105638 for discussion

### DIFF
--- a/extensions/websocket/01_websocket.js
+++ b/extensions/websocket/01_websocket.js
@@ -241,18 +241,15 @@
             this[_readyState] = CLOSED;
 
             const errEvent = new ErrorEvent("error");
-            errEvent.target = this;
             this.dispatchEvent(errEvent);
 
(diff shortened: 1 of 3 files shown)


extensions/websocket/01_websocket.js
```diff
@@ -241,18 +241,15 @@
             this[_readyState] = CLOSED;
 
             const errEvent = new ErrorEvent("error");
-            errEvent.target = this;
             this.dispatchEvent(errEvent);
```
dsherret, line 244: `target` is set when calling `dispatchEvent`

lucacasonato (approved):

PR #11108, opened 2021-06-24
fix(wpt): Clean up temporary files created by the WPT test runner

### LINKED ISSUE #11107: The WPT test runner doesn't delete the temp files it creates
The WPT test runner creates a temporary file for each WPT test file, containing every script used in the test. And although each of those files is only needed during the run of the test, the test runner doesn't delete them, and so they can add up. In particular, when the full WPT test suite is run, this adds up to over 600 files weighing over 100 MB in total. And since the names are random, running the test suite (or part of it) again will only add more files, rather than replacing the existing ones. The test runner should instead clean up after itself.

andreubotella:
Fixes #11107.

### DIFF
--- a/tools/wpt/runner.ts
+++ b/tools/wpt/runner.ts
@@ -82,60 +82,65 @@ export async function runSingleTest(
     prefix: "wpt-bundle-",
     suffix: ".js",
   });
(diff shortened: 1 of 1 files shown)


dsherret (approved):
LGTM. Thanks @andreubotella!

PR #11122, opened 2021-06-25
chore: use local deno_std in tools scripts

bartlomieju:

### DIFF
--- a/tools/util.js
+++ b/tools/util.js
@@ -5,11 +5,11 @@ import {
   join,
   resolve,
   toFileUrl,
-} from "https://deno.land/std@0.84.0/path/mod.ts";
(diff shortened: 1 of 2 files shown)


caspervonb:
Minor aside, we should probably move this submodule to the root folder if we're using it everywhere? Initially put it in next to wpt because it was only being used as test fixtures.

bartlomieju:
> Minor aside, we should probably move this submodule to the root folder if we're using it everywhere? Initially put it in next to wpt because it was only being used as test fixtures.

I don't think it's worth reshuffling it again, `format.js` and `lint.js` don't change too often...

dsherret (approved):
LGTM

PR #11123, opened 2021-06-25
fix(lsp): reload import registries should not error when the module registries directory does not exist

dsherret:
The `remove_dir_all` function will error when the directory does not exist. This is the cause of https://github.com/denoland/vscode_deno/issues/394

Classic bug 😊

### DIFF
--- a/cli/fs_util.rs
+++ b/cli/fs_util.rs
@@ -161,6 +161,16 @@ where
   Ok(target_files)
 }
 
+// Asynchronously removes a directory and all its descendants, but does not error
+// when the directory does not exist.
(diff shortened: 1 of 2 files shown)


cli/fs_util.rs
```diff
@@ -161,6 +161,16 @@ where
   Ok(target_files)
 }
 
+// Asynchronously removes a directory and all its descendants, but does not error
+// when the directory does not exist.
+pub async fn remove_dir_all_if_not_exists(path: &Path) -> std::io::Result<()> {
```
dsherret, line 166: Oops... should be opposite 😅

kitsonk (approved):
LGTM

jsejcksn (approved):
LGTM also

PR #11124, opened 2021-06-25
feat(cli): Exit the test suite immediately upon `n` number of failing test suite

getspooky:
Alias: `-b`. Exit the test suite immediately upon `n` number of failing test suite. Minimum value is `1`.

Example :

```sh
deno test fetcher_test.ts --bail 3
```

Short style 

```
deno test fetcher_test.ts --b 3
```

### DIFF
--- a/cli/tools/test_runner.rs
+++ b/cli/tools/test_runner.rs
@@ -344,7 +344,9 @@ pub async fn run_tests(
   allow_none: bool,
   filter: Option<String>,
   concurrent_jobs: usize,
+  bail: usize,
 ) -> Result<bool, AnyError> {
+
   if !doc_modules.is_empty() {
     let mut test_programs = Vec::new();
 
@@ -502,6 +504,7 @@ pub async fn run_tests(
       let mut has_error = false;
       let mut planned = 0;
       let mut reported = 0;
+      let mut failed = 0;
 
       for event in receiver.iter() {
         match event.message.clone() {
@@ -524,13 +527,19 @@ pub async fn run_tests(
             reported += 1;
 
             if let TestResult::Failed(_) = result {
+              failed += 1;
               has_error = true;
             }
+
           }
           _ => {}
         }
 
         reporter.visit_event(event);
+
+        if bail > 0 && failed > bail {
+          break;
+        }
 
         if has_error && fail_fast {
           break;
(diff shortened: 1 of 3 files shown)


caspervonb:
We already have this with fail-fast?

getspooky:
@caspervonb `fail-fast` will stop on first error but with `bail` we set ` n` number of failing test suite

lucacasonato:
@getspooky What is the usecase for this? Does Go have this behaviour?

getspooky:
By default, Deno runs all tests and throws all errors to the console upon completion. The `bail` config option can be used here to make Deno stop running tests after n failures. This options is inspired by [jest bail](https://jestjs.io/fr/docs/cli#--bail).

cli/tools/test_runner.rs
```diff
@@ -532,6 +535,10 @@ pub async fn run_tests(
 
         reporter.visit_event(event);
 
+        if bail > 0 && failed > bail {
```
dsherret, line 538: Should this be `failed >= bail`?

getspooky: I think that number of failing test must be greater thant `0`.

dsherret: So the description says "Exit the test suite immediately upon n number of failing test suite". So, you would want this to exit when failed == bail, correct?

Anyway, I just doing a drive by review. Would be good to get consensus on this feature first.

crowlKats:
if we are gonna pass this, it should be a single flag merged with `fail-fast`. we already have a plethora of flags and we have to be careful of adding new ones.

getspooky:
If there is not need to add another flag like `--bail` , I think we should change the behaviour `fail-fast` to take `n` number and by default `Stop on first error`.
what do you think ?

bartlomieju:
Haven't really seen such functionality, but I'd be okay to add optional integer argument to `--fail-fast` flag to give a threshold

getspooky:
@bartlomieju  Yes it's common [use case](https://discord.com/channels/684898665143206084/684911491035430919/858143154943557632) , so we will still  change `--fail-fast` flag instread of using new one `--bail` ?

bartlomieju:
> @bartlomieju  Yes it's common [use case](https://discord.com/channels/684898665143206084/684911491035430919/858143154943557632) , so we will still  change `--fail-fast` flag instread of using new one `--bail` ?

Yes

bartlomieju:
Closing favor of #11159

PR #11125, opened 2021-06-25
fix(serde_v8): Export serde_v8 just like we export serde_json

nicksrandall:
My project is failing to compile because I'm using `deno_core::v8` and `serde_v8` and the compiler complains about 2 versions of `rusty_v8` in my project. I'm hoping this solves the problem?


bartlomieju (approved):
LGTM, thanks @nicksrandall

bartlomieju:
> My project is failing to compile because I'm using deno_core::v8 and serde_v8 and the compiler complains about 2 versions of rusty_v8 in my project. I'm hoping this solves the problem?

I'm not sure it's related to missing reexport - make sure you use same version of `serde_v8` that `deno_core` uses that you import in your project. (Of course this PR will solve this as you won't have to specify `serde_v8` separately).

PR #11127, opened 2021-06-25
chore: move repl integration tests to separate file

dsherret:
This is essentially the same as before except the repl integration tests are in a new file. It shows up under integration_tests.rs the same way:

image

We might want to consider doing this with more of the integration tests (ex. move out fmt tests)


dsherret:
Weird... this builds fine for me on Windows, WSL, and Mac. The path attribute is documented here and should work as it's being used:

https://doc.rust-lang.org/reference/items/modules.html#the-path-attribute

I'm not sure why the CI is saying it can't find the file...

cli/tests/integration_tests.rs
```diff
@@ -83,6 +83,9 @@ fn js_unit_tests() {
 mod integration {
   use super::*;
 
+  #[path = "../tools/repl_tests.rs"]
```
bartlomieju, line 86: Should be `./tools/repl_tests.rs` instead?

dsherret: Because it's under the `mod integration` like before then it needs to be `../`. See https://doc.rust-lang.org/reference/items/modules.html#the-path-attribute

Don't know why this doesn't work on the CI... works on linux, mac, and windows for me locally.

ry: Behavior change between 1.52 and 1.53 of Rust? That's surprising...

ry (commented):
Seems like a good way of breaking up integration_tests.rs without introducing new crates. I hope you can get it green.

cli/tests/integration_tests.rs
dsherret: No, I wasn't running the command correctly for Linux and Mac. Platform specifc bug. Opened https://github.com/rust-lang/rust/issues/86638

cli/tests/integration_tests.rs
```diff
@@ -80,9 +80,13 @@ fn js_unit_tests() {
   assert!(status.success());
 }
 
+#[path = ""]
```
dsherret, line 83: Not ideal, but it works...

Also, sorry about the build failures... platform specific issues.

ry: Please add a comment here

ry (approved):
Otherwise looks good to me.

Issue #11105, opened 2021-06-24
Get JsRuntime execution output


gufoe:
Hello, I'd like to use deno to interpret js at runtime and get the result, for instance:
```rust
let res = runtime.execute("<main>", "return 123");
assert!(res.unwrap() == serde_json::Value::from(123))
```
But execute actually returns nothing useful.

Is there any way to achieve this?

bartlomieju:
As you've noticed `runtime.execute()` returns `Result<(), AnyError>` and thus doesn't return the evaluation value. This issue has come up a few times already and I think I'll just go ahead and change API.

That said - you won't be able to do this comparison:
> `assert!(res.unwrap() == serde_json::Value::from(123))`
The value returned will have `v8::Value` type - so you'll have to convert it to some Rust primitive type yourself.

gufoe:
Thank you for the quick reply!
I guess converting the v8::Value to something else will be fine.

Sorry to ask this question, but do you have an idea of when this feature will be available?

bartlomieju:
Probably in 1.12 on July, 13th

weareoutman:
Hi, would it support getting async result from a module includes top-level await? (using `execute_module`)

bartlomieju:
Not in the #11129, once that PR lands I will look into supporting that for modules.

PR #11129, opened 2021-06-26
feat(core): return v8::Value from JsRuntime::execute_script

bartlomieju:
This commit changes return type of JsRuntime::execute_script to include
v8::Value returned from evaluation.

When embedding deno_core it is sometimes useful to be able to inspect
script evaluation value without the hoops of adding ops to store the
value on the OpState.

v8::Global<v8::Value> is used so consumers don't have to pass
scope themselves.

Closes https://github.com/denoland/deno/issues/11105


ry (commented):
1. This exposes V8 API to deno_core callers. This has not been 100% the case already, but generally we expect that deno_core callers do not need to interact with V8.

2. This unnecessarily creates a `v8::Global`, it could use a local handle.

bartlomieju:
> This exposes V8 API to deno_core callers. This has not been 100% the case already, but generally we expect that deno_core callers do not need to interact with V8.

Correct, but current API is very limiting - for simple one-off scripts you need to setup an op to store some result on state instead of getting the result back directly - some caller actually want to inspect the value.

As a counter argument `v8` objects already leak in `runtime/` crate for web worker and inspector server implementation.

> This unnecessarily creates a v8::Global, it could use a local handle.

Correct, otherwise I'd need to change the signature to accept `scope` parameter, which would be even more unwieldy. 

I presume you're not happy with that change.

lucacasonato:
Why not just return a generic `T: Deserialize` and deserialize the result with serde_v8?

bartlomieju:
> Why not just return a generic `T: Deserialize` and deserialize the result with serde_v8?

Because you might return a value that doesn't implement this trait - eg. a class instance

caspervonb:
At-least for my use case in the test runner I could derive `Deserializable` on `TestResult` and return it directly.
So `T: Deserializable` be sufficient to remove the test message ops.

Either way, this PR will let me untangle the test runner quite a bit and remove all of the "test message ops" 👍

bartlomieju:
@ry #11169 is a good example why this change is desirable, besides requests on Discord

ry:
I don't like the unnecessary use of Global and leaking of v8 objects into the deno_core API. But maybe @piscisaureus has a different opinion.

piscisaureus:
> I don't like the unnecessary use of Global and leaking of v8 objects into the deno_core API.

I agree with this.

bartlomieju:
@ry @piscisaureus do you alternative approach to #11169 that doesn't require ops?

lucacasonato:
I would still argue it should just return a serde_v8 deserializable. We already have the limitation that you can only return simple objects and primitives through ops, so I don't see how `execute_script` should be any different in that regard. Using a deserializable should be enough to resolve #11169 according to @caspervonb.

caspervonb:
>  Using a deserializable should be enough to resolve #11169 according to @caspervonb.

Actually, I guess from `evaluate_script` I would get a promise since I can't await it and we can't deserialise that.
Altho that could probably be punted to rusty_v8?

piscisaureus (approved):

Issue #11091, opened 2021-06-23
Rust panic in http.rs


konsti219:
I'm running a web server using deno and oak.

```
deno 1.11.2 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.35
typescript 4.3.2
```

I'm getting rust panics, but I haven't figured out what causes them yet.

Error with full backtrace:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', runtime/ops/http.rs:221:47
stack backtrace:
   0:     0x558c6d1e6df3 - std::backtrace_rs::backtrace::libunwind::trace::ha5edb8ba5c6b7a6c
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x558c6d1e6df3 - std::backtrace_rs::backtrace::trace_unsynchronized::h0de86d320a827db2
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x558c6d1e6df3 - std::sys_common::backtrace::_print_fmt::h97b9ad6f0a1380ff
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x558c6d1e6df3 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h14be7eb08f97fe80
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x558c6ca359cf - core::fmt::write::h2ca8877d3e0e52de
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/fmt/mod.rs:1094:17
   5:     0x558c6d1e6674 - std::io::Write::write_fmt::h64f5987220b618f4
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/io/mod.rs:1584:15
   6:     0x558c6d1e611b - std::sys_common::backtrace::_print::h7f1a4097308f2e0a
[...]

PR #11130, opened 2021-06-26
fix(http): remove unwrap() in HTTP bindings

bartlomieju:
Fixes https://github.com/denoland/deno/issues/11091

I don't really have a test for that panic, but we should use empty string as a default instead of unwrapping.


runtime/ops/http.rs
```diff
@@ -218,7 +218,7 @@ async fn op_http_request_next(
         } else {
           Cow::Owned(conn_resource.addr.to_string())
         };
-        let path = req.uri().path_and_query().unwrap();
+        let path = req.uri().path_and_query().map(|p| p.as_str()).unwrap_or("");
```
satyarohith, line 221: Requests with the `CONNECT` method triggers this panic. I think we need to default to the `/` path instead of an empty string.

ry (dismissed):
Needs a test

runtime/ops/http.rs
```diff
@@ -218,7 +218,7 @@ async fn op_http_request_next(
         } else {
           Cow::Owned(conn_resource.addr.to_string())
         };
-        let path = req.uri().path_and_query().unwrap();
+        let path = req.uri().path_and_query().map_or("/", |p| p.as_str());
```
bnoordhuis, line 221: I do believe I recognize this code. :-)

bnoordhuis (approved):

runtime/ops/http.rs
bartlomieju: That's quite possible :P

PR #11131, opened 2021-06-26
chore: split up integration_tests.rs into separate files

dsherret:
Continuation of #11127.

I did not want to organize the raw files for this PR. Instead this was just about breaking up `integration_tests.rs`. I did not edit any test code in this PR.

**Before**

```
tools
  - repl_tests.rs # from #11127
integration_tests.rs
integration_tests_lsp.rs
```

**After**

```
integration
   - bundle_tests.rs
   - cache_tests.rs
   - etc...
   - lsp.ts
   - mod.ts
   - etc...
integration_tests.rs
```


cli/tests/integration/general_tests.rs
```diff
@@ -0,0 +1,1036 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 1: Nitpick: I suggest to leave these tests in `integration/mod.rs`

bartlomieju (approved):
Looks good to me

cli/tests/integration/general_tests.rs
dsherret: Good idea! a3e407c

PR #11134, opened 2021-06-26
feat(inspector): pipe console messages between terminal and inspector

bartlomieju:
Opening to get more eyes on this. Actual implementation will be way harder and we'll need to refactor our own 
Console implementation, but I finally managed to figure out how to make those messages
appear in inspector, long story short: V8 provides default implementation of `Console` (available as `globalThis.console` after creating isolate). We need to save this "console from VM" when creating a new `Console` instance internally and
call appropriate methods on "console from VM" when `Console` methods are called. Unfortunately some of our Console
impl methods call other methods (eg. `console.table` does the formatting and internally calls `console.log`), so we'll need some reorganization to make it possible.

Screenshot 2021-06-26 at 05 16 01

Closes https://github.com/denoland/deno/issues/4502


bartlomieju:
Actually reorganizing internal console methods wasn't that bad, now need to figure out how to fix stack frame:

Screenshot 2021-06-26 at 06 17 49


For reference how it's done in Node: https://github.com/nodejs/node/blob/f4d0a6a07bf72c556c5ba8e98b3b23327c273e80/lib/internal/util/inspector.js#L40-L61

retog (commented):
Why is there a #private method for every log-level but debug?

bartlomieju:
> Why is there a #private method for every log-level but debug?

Because no other method calls `debug` internally.

bartlomieju:
This should be now landable. This is heavily inspired how Node does it - there are actually more actions being done in Node, but I'm not sure if that's required in Deno. Overall it's an improvement from current implementation and I'll be happy to fix any bugs that might occur.

Stack frames are properly preserved and messages are piped between terminal and inspector in both ways:
Screenshot 2021-06-26 at 16 53 03

core/bindings.rs
```diff
@@ -460,6 +464,54 @@ fn eval_context(
...
+///         consoleFromDeno[key],
+///       );
+///     }
+///   }
+/// }
+/// ```
+///
+/// Inspired by:
+/// https://github.com/nodejs/node/blob/1317252dfe8824fd9cfee125d2aaa94004db2f3b/src/inspector_js_api.cc#L194-L222
+fn call_console(
```
lucacasonato, line 490: For the record I think this is incredibly ugly, and it is incredibly frustrating that you can not mark stack frames as internal if they come from JS code. I can't come up with a better solution though, so I guess it is here to stay. Maybe can you rename to `double_call`? The function is not specific to console in any way implementation wise.

runtime/js/99_main.js
```diff
@@ -455,10 +455,35 @@ delete Object.prototype.__proto__;
 
   let hasBootstrapped = false;
 
+  // A helper function that will bind our own console implementation
+  // with default implementation of Console from V8. This will cause
+  // console messages to be piped to inspector console.
+  //
+  // We are using `Deno.core.callConsole` binding to preserve proper stack
+  // frames in inspector console.
```
lucacasonato, line 463:
```suggestion
  // We are using `Deno.core.callConsole` binding to preserve proper stack
  // frames in inspector console. This has to be done because V8 considers
  // the last JS stack frame as gospel for the inspector. In our case we
  // specifically want the latest user stack frame to be the one that matters
  // though.
```

Let's add some more docs on why this is needed.

lucacasonato (approved):
Consider moving `wrapConsole` to the extension.

core/bindings.rs
bartlomieju: Implementation is not console specific, but assertions are. I'd prefer to keep the name.

lucacasonato: They just assert the two things being passed are functions, right? Not console specific I think. But if you want to keep the name, I don't really care that much.

core/bindings.rs
```diff
@@ -460,6 +464,54 @@ fn eval_context(
...
+  scope: &mut v8::HandleScope,
+  args: v8::FunctionCallbackArguments,
+  _rv: v8::ReturnValue,
+) {
+  assert!(args.length() >= 2);
```
punkeel, line 495: This code will cause a Rust panic if callConsole is invoked with no args, or invalid args.
- `Deno.core.callConsole()`
- `Deno.core.callConsole(1)`
- `Deno.core.callConsole(1, 2)`

extensions/console/02_console.js
```diff
@@ -1774,6 +1774,32 @@
...
+  //
+  // We are using `Deno.core.callConsole` binding to preserve proper stack
+  // frames in inspector console. This has to be done because V8 considers
+  // the last JS stack frame as gospel for the inspector. In our case we
+  // specifically want the latest user stack frame to be the one that matters
+  // though.
+  //
+  // Inspired by:
+  // https://github.com/nodejs/node/blob/1317252dfe8824fd9cfee125d2aaa94004db2f3b/lib/internal/util/inspector.js#L39-L61
+  function wrapConsole(consoleFromDeno, consoleFromV8) {
```
punkeel, line 1789: I wonder if similar results _could_ be achieved using a `Proxy` object

punkeel:
The feature added by this PR super neat!! 👏  Thanks for working on this

core/bindings.rs
bartlomieju: Indeed, this function is not really meant to be called manually by users.

extensions/console/02_console.js
bartlomieju: @lucacasonato tried without a success. It is mainly required to properly preserve stack frames.

Issue #8858, opened 2020-12-22
REPL doesn't print unhandled promise rejections


nayeemrmn:
Likely a regression from upgrading to V8's REPL system. cc @caspervonb 

Actual:
```
> void Promise.reject(new Error())
undefined
> 
```
Expected:
```
> void Promise.reject(new Error())
undefined
Uncaught (in promise) Error
    at <anonymous>:1:21
> 
```

caspervonb:
So if my memory serves me right, our inspector server is incomplete here and we don't implement the notifications for we need for this.

The client side fix is fairly trivial, will open a separate issue for the InspectorServer issue as it affects all clients.

bartlomieju:
With https://github.com/denoland/rusty_v8/pull/1149 I can get notifications about these exceptions:
```
notifications [
    Object {
        "method": String("Runtime.exceptionThrown"),
        "params": Object {
            "timestamp": Number(0),
            "exceptionDetails": Object {
                "exceptionId": Number(1),
                "text": String("Uncaught"),
                "lineNumber": Number(0),
                "columnNumber": Number(0),
                "scriptId": String("1"),
                "url": String("foo"),
                "exception": Object {
                    "type": String("object"),
                    "subtype": String("error"),
                    "className": String("Error"),
                    "description": String("Error\n    at <anonymous>:2:21"),
                    "objectId": String("-2897224664797834136.1.2"),
                    "preview": Object {
                        "type": String("object"),
                        "subtype": String("error"),
                        "description": String("Error\n    at <anonymous>:2:21"),
                        "overflow": Bool(false),
                        "properties": Array [
                            Object {
                                "name": String("stack"),
                                "type": String("string"),
                                "value": String("Error\n    at <anonymous>:2:21"),
                            },
                            Object {
                                "name": String("__callSiteEvals"),
                                "type": String("object"),
                                "value": String("Array(1)"),
                                "subtype": String("array"),
                            },
                        ],
                    },
                },
                "executionContextId": Number(1),
            },
        },
    },
]
```

philc:
I wanted to chime with encouragement to @bartlomieju that this would be a great bug to resolve. For those who develop primarily through a REPL, when a rejected Promise is mistakenly not handled, the lack of output about it is a sharp edge.

PR #11142, opened 2021-06-26
[WIP] fix: print uncaught exception in REPL

bartlomieju:
This is not done yet - for now I found that promise exceptions were silently ignored and fixed it
to return exceptions instead. They still need to be handled properly because right now REPL exits on uncaught exception.
Most likely we need to wire up proper methods in `V8::Inspector` to return [`Runtime.exceptionThrown` notification](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-exceptionThrown) (https://source.chromium.org/chromium/chromium/src/+/main:v8/src/inspector/v8-inspector-impl.h;l=91-96?q=exceptionThrown&ss=chromium&start=11). 

Fixes https://github.com/denoland/deno/issues/8858


bartlomieju:
Closing for now.

PR #11144, opened 2021-06-26
fix(runtime/http): Encode and decode headers as byte strings in the HTTP server

andreubotella:
See #11070.


bartlomieju (approved):
LGTM, nice tests! Thank you @andreubotella

PR #11150, opened 2021-06-27
feat: Add "deno_net" extension

bartlomieju:
All TCP, UDP, Datagram, TLS and HTTP ops are now available in `deno_net` crate.

Due to complex interactions in JS API between TCP/UDP/Datagram it's not really feasible
to split those ops into more granular extensions.

TODO:
- [x] move JS code to `extensions/net`
- [x] move `Deno.errors.BadResource` and `Deno.errors.Interrupted` class to `deno_core` so it can be used from `deno_net` extension
- [ ] probably rename files in `extensions/net`
- [x] provide declaration files
- [x] clean dependencies in `runtime/Cargo.toml`

Quirks:
- Lots of duplicated (or, copy-pasted) code: IO ops, unstable checker


bartlomieju:
@ry I'm pretty much done, only thing left is getting type declaration files sorted out. I would appreciated initial review.

extensions/net/01_net.js
```diff
@@ -3,11 +3,27 @@
...
-  const { read, write } = window.__bootstrap.io;
+  const { BadResource } = core;
+
+  async function read(
+    rid,
+    buffer,
+  ) {
+    if (buffer.length === 0) {
+      return 0;
+    }
```
ry, line 14: Seems unnecessary

extensions/net/01_net.js
```diff
@@ -3,11 +3,27 @@
...
+
+    const nread = await core.opAsync("op_net_read_async", rid, buffer);
+
+    return nread === 0 ? null : nread;
+  }
```
ry, line 19: Nit: too much vertical whitespace

extensions/net/README.md
```diff
@@ -0,0 +1,3 @@
+# deno_net
+
+This crate implements networking APIs.
```
ry, line 3: Ideally a bit more detail here. List the ops it provides?

ry (commented):
If I tried to use `Deno.read` on stdin, like this `Deno.read(0, buf)` would it still work after this patch?

bartlomieju:
> If I tried to use `Deno.read` on stdin, like this `Deno.read(0, buf)` would it still work after this patch?

Yes

extensions/net/01_net.js
bartlomieju: It's a copy paste from existing io ops

extensions/net/01_net.js
bartlomieju: Ditto

extensions/net/README.md
bartlomieju: Added a list of ops

ry (approved):
LGTM - nice work @bartlomieju, this moves us forward

cli/build.rs
```diff
@@ -68,6 +68,9 @@ fn create_compiler_snapshot(
     "deno.broadcast_channel",
     deno_broadcast_channel::get_declaration(),
   );
+  op_crate_libs.insert("deno.net", deno_net::get_declaration());
+  op_crate_libs
+    .insert("deno.net_unstable", deno_net::get_unstable_declaration());
```
bartlomieju, line 73: @kitsonk these are the declaration I mentioned on Discord

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2,6 +2,7 @@
 
 /// <reference no-default-lib="true" />
 /// <reference lib="esnext" />
+/// <reference lib="deno.net" />
```
bartlomieju, line 5: And they are used here...

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -2,6 +2,7 @@
 
 /// <reference no-default-lib="true" />
 /// <reference lib="deno.ns" />
+/// <reference lib="deno.net_unstable" />
```
bartlomieju, line 5: ...and here

PR #11156, opened 2021-06-28
feat(cli/tools/test_runner): add terse reporter

caspervonb:
This adds a terse reporter which displays one character per test instead of one line.


cli/tools/test_runner.rs
```diff
@@ -199,8 +199,128 @@ impl TestReporter for PrettyTestReporter {
...
+          }
+        }
+      }
+
+      _ => {}
+    }
+  }
+
+  fn done(&mut self) {
+    if !self.failures.is_empty() {
```
caspervonb, line 278: Little bit of duplication here, will follow up and make these reporters stateless.

bartlomieju:
I'm not a fan of using `--quiet` flag for "terse" output. If we were to land this PR I think we should add `--reporter` flag

caspervonb:
> I'm not a fan of using `--quiet` flag for "terse" output. If we were to land this PR I think we should add `--reporter` flag

Mirrors `cargo test`.

Also with another flag, we'd have to call quiet anyway:

```sh
deno test --quiet --reporter=terse
```

cli/flags.rs
```diff
@@ -1009,6 +1010,11 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
         .help("Don't return error code if no test files are found")
         .takes_value(false),
     )
+    .arg(
+      Arg::with_name("terse")
+        .long("terse")
+        .help("UNSTABLE: Display one character per test instead of one line"),
```
bartlomieju, line 1016: Can you add here that `--terse` implies `--quiet`?

caspervonb: Ended up not doing that after all. 
Aiming to make quiet only applies to diagnostics and terse only applies to the reporter(s).

bartlomieju (approved):
LGTM

ry:
@caspervonb, first of all thanks for the PR. The feature is cleanly implemented (as always with your patches).

But I have concerns about feature creep in the test runner. Although this feature is a nice thing to have, I think one could argue it isn't a necessity. I don't like that it adds another CLI option. Is there precedence for this? The go test runner does not have this, for example.

caspervonb:
> Is there precedence for this? The go test runner does not have this, for example.

Yeah this is pretty much nabbed directly from rust's libtest (`cargo test --quiet`).

PR #11163, opened 2021-06-28
feat(test): add --shuffle flag to randomize test ordering

caspervonb:
This adds a shuffle option to `deno test` which randomises the order in which test modules and definitions are evaluated.
The main use case for this os to help discover scenarios where tests are order dependent and are masking bad tests or implementation defects.

Both the test modules and test definitions use a deterministic sort order based on the given seed value.

The option is disabled by default for backwards compatibility and has also been marked as unstable because the sort implementation for the test definitions is somewhat weak and will likely change in the future.


caspervonb:
CC @lucacasonato

bartlomieju:
Discussed with @caspervonb, this is modelled after `go test` which added `-shuffle` flag in 1.17. I don't have immediate need for it, but I find it useful.

runtime/js/40_testing.js
```diff
@@ -240,6 +241,24 @@ finishing test case.`;
...
+        const a = 1103515245;
+        const c = 12345;
+
+        return function (max) {
+          return state = ((a * state + c) % m) % max;
+        };
+      }(shuffle));
+
+      for (let i = pending.length - 1; i > 0; i--) {
+        const j = 1 + nextInt(pending.length - 1);
```
bnoordhuis, line 257: If you're trying to implement [Fisher-Yates](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle), this line should probably be:
```suggestion
        const j = nextInt(i);
```

dsherret: I think that would cause it to prefer shuffling to items with a lower index? (So the `0` index value would always stay the same, I think)

That said, why is it `1 +`? Am I reading it correctly that `j` would never be `0`? I'm also getting `.5` values (ex. `1.5`, `3.5`, etc) trying out this code... `nextInt` doesn't seem to be returning an integer.

cli/tools/test_runner.rs
```diff
@@ -450,6 +464,7 @@ pub async fn run_tests(
   let test_options = json!({
       "disableLog": quiet,
       "filter": filter,
+      "shuffle": shuffle,
```
dsherret, line 467: Why is the second shuffle necessary in the JS code?

runtime/js/40_testing.js
bnoordhuis: > I think that would cause it to prefer shuffling to items with a lower index?

Yes, it operates on a shrinking view of the list.

> So the 0 index value would always stay the same, I think

No, unless `nextInt` never returns zero.

cli/tools/test_runner.rs
caspervonb: Shuffle of test definitions as-well as modules.

runtime/js/40_testing.js
```diff
@@ -240,6 +241,24 @@ finishing test case.`;
       only: only.length > 0,
     });
 
+    if (shuffle) {
```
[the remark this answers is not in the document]
caspervonb: Good call

cli/flags.rs
```diff
@@ -1016,6 +1017,20 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
...
+        .long("shuffle")
+        .value_name("NUMBER")
+        .help("(UNSTABLE): Shuffle the order in which the tests are run")
+        .min_values(0)
+        .max_values(1)
+        .require_equals(true)
+        .takes_value(true)
+        .validator(|val: String| match val.parse::<u64>() {
+          Ok(_) => Ok(()),
+          Err(_) => Err("Shuffle seed should be a number".to_string()),
```
dsherret, line 1031: Nitpick: "number" -> "positive integer" in case someone provides a negative seed for some reason.

cli/flags.rs
```diff
@@ -1016,6 +1017,20 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
...
         .help("Run tests with this string or pattern in the test name"),
     )
+    .arg(
+      Arg::with_name("shuffle")
+        .long("shuffle")
+        .value_name("NUMBER")
+        .help("(UNSTABLE): Shuffle the order in which the tests are run")
+        .min_values(0)
+        .max_values(1)
+        .require_equals(true)
```
dsherret, line 1027: It's strange that `--jobs <value>` does not require an equals, but this does. Should we require equals here? It seems like jobs is the odd one out though.

dsherret (approved):

cli/flags.rs
caspervonb: Aligns with the `--inspect` flag, maybe we should go back and be stricter with --jobs but jobs has no required equals for `make` compatibility (for example `make -j3`).

For reference, inspect has the same setup with min 0, max 1, requires value:
```rs
.arg(
  Arg::with_name("inspect")
    .long("inspect")
    .value_name("HOST:PORT")
    .help("Activate inspector on host:port (default: 127.0.0.1:9229)")
    .min_values(0)
    .max_values(1)
    .require_equals(true)
    .takes_value(true)
 ```

dsherret:
LGTM. Thanks @caspervonb!

PR #11172, opened 2021-06-29
refactor(net): feature flag non tcp APIs

lucacasonato:

### DIFF
--- a/extensions/net/io.rs
+++ b/extensions/net/io.rs
@@ -1,6 +1,5 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-use crate::ops_tls as tls;
 use deno_core::error::null_opbuf;
 use deno_core::error::AnyError;
 use deno_core::error::{bad_resource_id, not_supported};
@@ -24,7 +23,9 @@ use tokio::io::AsyncWrite;
 use tokio::io::AsyncWriteExt;
 use tokio::net::tcp;
 
-#[cfg(unix)]
+#[cfg(feature = "full")]
+use crate::ops_tls as tls;
+#[cfg(all(unix, feature = "full"))]
 use tokio::net::unix;
 
 pub fn init() -> Vec<OpPair> {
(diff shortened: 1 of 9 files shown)


extensions/net/io.rs
```diff
@@ -219,11 +209,17 @@ async fn op_shutdown(
     .resource_table
     .get_any(rid)
     .ok_or_else(bad_resource_id)?;
-  if let Some(s) = resource.downcast_rc::<TcpStreamResource>() {
+  #[cfg(all(unix, feature = "full"))]
+  if let Some(s) = resource.downcast_rc::<UnixStreamResource>() {
     s.shutdown().await?;
```
piscisaureus, line 214:
```suggestion
    return s.shutdown().await;
```

extensions/net/io.rs
```diff
@@ -219,11 +209,17 @@ async fn op_shutdown(
...
-  } else if let Some(s) = resource.downcast_rc::<TlsStreamResource>() {
+    return Ok(());
+  }
+  #[cfg(feature = "full")]
+  if let Some(s) = resource.downcast_rc::<TlsStreamResource>() {
     s.shutdown().await?;
```
piscisaureus, line 219:
```suggestion
    return s.shutdown().await;
```

piscisaureus:
Have you considered using `cfg_if` instead of adding boatloads of `[cfg]` attributes? It is used something like this:


```rust
cfg_if::cfg_if! {
    if #[feature = "full"] {
      use foo:foo;
      use std::bar:xyz;
    }
}
```

lucacasonato:
@piscisaureus Oh that looks a lot better indeed. I am discussing with Bartek how we could prevent this feature flag from being needed at all (at least for the time being). If the conclusion is that we need the flag anyway, I'll look into `cfg_if`.

piscisaureus:
Might be easier to just feature gate the op registration logic.

lucacasonato:
I am going to solve this differently.

PR #11175, opened 2021-06-29
Get rid of REPL timeout

piscisaureus:

### DIFF
--- a/cli/tools/repl.rs
+++ b/cli/tools/repl.rs
@@ -8,6 +8,7 @@ use crate::colors;
 use crate::media_type::MediaType;
 use crate::program_state::ProgramState;
 use deno_core::error::AnyError;
+use deno_core::futures::future::FusedFuture;
 use deno_core::futures::FutureExt;
 use deno_core::serde_json::json;
 use deno_core::serde_json::Value;
@@ -626,9 +627,7 @@ async fn read_line_and_poll(
   response_tx: &Sender<Result<Value, AnyError>>,
   editor: ReplEditor,
 ) -> Result<String, ReadlineError> {
-  let mut line = tokio::task::spawn_blocking(move || editor.readline());
-
-  let mut poll_worker = true;
+  let mut line_fut = tokio::task::spawn_blocking(move || editor.readline());
 
   loop {
     for (method, params) in message_rx.try_iter() {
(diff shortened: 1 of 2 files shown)


cli/tools/repl.rs
```diff
@@ -638,23 +637,11 @@ async fn read_line_and_poll(
...
 
-    // Because an inspector websocket client may choose to connect at anytime when we have an
-    // inspector server we need to keep polling the worker to pick up new connections.
-    // TODO(piscisaureus): the above comment is a red herring; figure out if/why
-    // the event loop isn't woken by a waker when a websocket client connects.
-    let timeout = tokio::time::sleep(tokio::time::Duration::from_millis(100));
-    pin!(timeout);
-
     tokio::select! {
-      result = &mut line => {
```
bartlomieju, line 649: Just removing this timeout, blocks the whole repl

dsherret: It seems to have been working when we tried it out, but I am going to test it more. The readline is on a blocking task (line 630).

piscisaureus: > Just removing this timeout, blocks the whole repl

Just tested it and that didn't happen. David will do more testing and finish the PR if it works.

bartlomieju: Sorry, there's a difference between what I tried. 

It seems you also removed `poll_worker` boolean. Now the process spins at 100% CPU

caspervonb: Nice and crispy 100% cpu usage.

piscisaureus: > It seems you also removed `poll_worker` boolean. Now the process spins at 100% CPU

I think this could be avoided by not polling the event loop again if it returns `Poll::Ready(_)`.

dsherret: @piscisaureus yup, Bartek helped me out with that. This PR should be good now!

cli/tools/repl.rs
```diff
@@ -444,7 +447,7 @@ impl ReplSession {
   }
 
   pub async fn run_event_loop(&mut self) -> Result<(), AnyError> {
-    self.worker.run_event_loop(false).await
+    self.worker.run_event_loop(true).await
```
dsherret, line 450: `wait_for_inspector` is the boolean here. When it's `false` the inspector doesn't work without hitting enter a few times in the repl, but once I set it to `true` then everything started working properly.

bartlomieju (approved):
LGTM

caspervonb (approved):
Tested quite a bit locally, usage is the 0-2% range and it works fine.

LGTM

Issue #11179, opened 2021-06-29
Linux release binary identifies as "canary" for v1.11.3


jsejcksn:
Today when upgrading Deno to `1.11.3` on two devices, I noticed that the result of calling `deno --version` on Linux returns info about a `canary` release channel instead of `release`. See outputs below for comparison:

Linux:
```
$ deno --version
deno 1.11.3+f4bed10 (canary, x86_64-unknown-linux-gnu)
v8 9.1.269.35
typescript 4.3.2
```

macOS:
```
% deno --version
deno 1.11.3 (release, x86_64-apple-darwin)
v8 9.1.269.35
typescript 4.3.2
```

This discrepancy seems like a bug, but if it's intentional and documented somewhere, please feel free to point me to the documentation and close.

piscisaureus:
This is a bug in the github actions workflow file.
(https://github.com/denoland/deno/blob/5db9f627e3b105ffc1ed43bb81732886308ddfdf/.github/workflows/ci.yml#L215)

lucacasonato:
Still an issue on 1.11.4

jsejcksn:
Seems to be fixed in `1.11.5`

Issue #11180, opened 2021-06-29
Could not publish deno 1.11.3 on crates.io


ry:
```
# cargo publish
    Updating crates.io index
   Packaging deno v1.11.3 (/Users/ry/src/deno/cli)
   Verifying deno v1.11.3 (/Users/ry/src/deno/cli)
  Downloaded twoway v0.2.2
  Downloaded env_logger v0.8.4
  Downloaded swc_ecma_visit v0.32.1
  Downloaded lsp-types v0.89.2
  Downloaded swc_common v0.10.21
  Downloaded indexmap v1.7.0
  Downloaded swc_ecma_parser v0.60.2
  Downloaded bumpalo v3.7.0
  Downloaded fd-lock v2.0.0
  Downloaded dprint-plugin-json v0.12.1
  Downloaded dprint-core v0.43.0
  Downloaded rustyline v8.2.0
  Downloaded swc_ecma_utils v0.37.2
  Downloaded swc_ecma_ast v0.47.0
  Downloaded swc_ecma_dep_graph v0.28.1
  Downloaded swc_ecmascript v0.43.0
  Downloaded deno_runtime v0.18.1
  Downloaded swc_ecma_visit v0.33.0
  Downloaded swc_ecma_parser v0.61.1
  Downloaded dprint-plugin-typescript v0.46.2
  Downloaded dprint-swc-ecma-ast-view v0.21.1
  Downloaded 21 crates (2.0 MB) in 1.24s
   Compiling libc v0.2.97
   Compiling proc-macro2 v1.0.27
   Compiling unicode-xid v0.2.2
   Compiling syn v1.0.65
   Compiling cfg-if v1.0.0
   Compiling autocfg v1.0.1
   Compiling serde_derive v1.0.126
   Compiling serde v1.0.126
   Compiling version_check v0.9.3
   Compiling once_cell v1.8.0
   Compiling memchr v2.4.0
   Compiling log v0.4.14
   Compiling bitflags v1.2.1
   Compiling lazy_static v1.4.0
   Compiling itoa v0.4.7
   Compiling matches v0.1.8
   Compiling tinyvec_macros v0.1.0
   Compiling ryu v1.0.5
   Compiling percent-encoding v2.1.0
   Compiling smallvec v1.6.1
   Compiling serde_json v1.0.64
   Compiling either v1.6.1
   Compiling fnv v1.0.7
   Compiling ppv-lite86 v0.2.10
   Compiling proc-macro-hack v0.5.19
[...]

PR #11181, opened 2021-06-29
ci: don't set DENO_CANARY when releasing

piscisaureus:
Fixes: #11179


lucacasonato (approved):
LGTM. does this warrant a 1.11.4 tomorrow?

piscisaureus:
> LGTM. does this warrant a 1.11.4 tomorrow?

#11180 also happened
Maybe it'd be good to test the release process once more, so we don't hit any snags when 1.12.0 rolls around.

Issue #10599, opened 2021-05-11
`deno test --doc` - `--watch` flag doesn't reload when making changes to doc


sebastienfilion:
If running the command `deno test --doc --watch ..` I would expect the test to re-run when I make changes to the relevant docs.
Right now, it reloads only when making changes to the functions.

For example take a file `users.ts` is a corresponding `users_test.ts`.

```ts
interface Users {
  name: string;
  email: string;
  login: {
    uuid: string;
  };
}

/**
 * ```
 * import { getUser } from "./users.ts";
 *
 * const p = getUser().then(user => user.fullName);
 * ```
 */
export function getUser (): Promise<Users> {

  return fetch(`https://randomuser.me/api/?inc=name,email,login,id`)
    .then(response => response.json())
    .then(({ results: [ user ] }) => user);
}
```

Obviously the doc is wrong because it refers to `user.fullName` instead of `user.name`.

Now if I run `deno test --doc --watch` on the file, I'll get an error but it doesn't rerun if I fix the error.

@caspervonb

caspervonb:
Thanks @sebastienfilion.
Known issue, merged in a hurry 😁

PR #11183, opened 2021-06-29
fix(cli): make --doc work with --watch

caspervonb:
This implements support for using the `--doc` along side the `--watch` option.

Closes https://github.com/denoland/deno/issues/10599


caspervonb:
Needs test(s), but it is also centred around the file watcher😭

bartlomieju (approved):
LGTM, thanks

PR #11184, opened 2021-06-29
test(cli): re-enable test --watch test

### LINKED ISSUE #10552: re-enable test "integration::file_watcher::test_watch"
Disabled in https://github.com/denoland/deno/pull/10550 because of flakiness on Windows

https://github.com/denoland/deno/pull/10481/checks?check_run_id=2543043090
https://github.com/denoland/deno/runs/2542717077

CC @Liamolucko

caspervonb:
Closes #10552

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -1134,6 +1134,8 @@ async fn test_command(
           }
         }
 
+        modules_to_reload.sort();
+
         Ok((paths_to_watch, modules_to_reload))
       }
       .map(move |result| {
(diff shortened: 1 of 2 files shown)


caspervonb:
Assuming that order of notifications is not always the same as the write order here.
Sorting the pending tests so that the order in which they were received in the queue does not matter and their printouts are the same on every run.

caspervonb:
One unrelated failure so far (`integration::run::single_compile_with_reload`).

caspervonb:
Also flaked `_084_worker_custom_inspect`

caspervonb:
Managed to run into his scenario when copying files:
```
error: Cannot resolve module "file:///C:/Users/caspervonb/target/foo.js" from "file:///C:/Users/caspervonb/target/foo_test.ts".
    at file:///C:/Users/caspervonb/target/foo_test.ts:1:0   
```

caspervonb:
Going to come back to this in a couple of days or so.

Issue #6154, opened 2020-06-06
WebAssembly.instantiateStreaming is undefined


dsherret:
```
> deno
Deno 1.0.5
exit using ctrl+d or close()
> typeof WebAssembly.instantiate
function
> typeof WebAssembly.instantiateStreaming
undefined
```

https://github.com/denoland/deno/blob/8b037364ac269b7887f3420fefb38be8d0693e9e/cli/js/lib.deno.shared_globals.d.ts#L48

caspervonb:
Same goes for `WebAssembly.compileStreaming`.

ry:
Using `--v8-flags=--wasm-test-streaming` makes these available. Tho I'm confused as to why this isn't enabled by default.

This could easily be added to the default configuration by adjusting this:
https://github.com/denoland/deno/blob/f247423f206ea5c3ca07254413a1ff11709a2ada/core/core_isolate.rs#L160-L168

caspervonb:
Portable code still has to check for the existence of `WebAssembly.instantiateStreaming` and fallback to `WebAssembly.instantiate` which all portable toolchains and loaders in the wild are currently doing.

Safari has no intent to implement.

> Tho I'm confused as to why this isn't enabled by default.

As far as I know `WebAssembly.compileStreaming` and `WebAssembly.instantiateStreaming` are vendor extensions, originally from Google then adopted by Mozilla.

[The webidl definition is the following:](https://webassembly.github.io/spec/js-api/#webassembly-namespace)
 ```webidl
[Exposed=(Window,Worker,Worklet)]
namespace WebAssembly {
    boolean validate(BufferSource bytes);
    Promise<Module> compile(BufferSource bytes);

    Promise<WebAssemblyInstantiatedSource> instantiate(
        BufferSource bytes, optional object importObject);

    Promise<Instance> instantiate(
        Module moduleObject, optional object importObject);
};
```

dsherret:
@caspervonb looks like the streaming API is defined here: https://webassembly.github.io/spec/web-api/index.html#streaming-modules

> So this is technically not a bug, but a feature request.

I think it's a bug because it exists in the declaration file, but not at runtime.

caspervonb:
Ah okay; it's defined as a web-api and not a js-api; which would also explain why it isn't enabled in V8 by default.

PR #11200, opened 2021-07-01
feat: Enable WebAssembly.instantiateStreaming and wasm async compilation

andreubotella:
The WebAssembly streaming APIs used to be enabled, but used to take buffer sources as their first argument (see #6154 and #7259). This change reenables it, requiring a `Promise<Response>` instead.


andreubotella:
For the record, the WPT failures in `/wasm/webapi/body.any.html` are due to #11211 and aren't specific to wasm. `/wasm/webapi/origin.sub.any.html` are tests related to CORS and the same-origin policy, which we don't implement. And `/wasm/webapi/historical.any.html` isn't related to the streaming API, but to testing that `WebAssembly.Module` can't be serialized with APIs that use that serialization for storage – tested using IndexedDb, which we don't implement.

andreubotella:
With #11221 now merged, I've updated this PR to enable wasm async compilation, which was blocked on it. I'll be adding some tests for it next.

bartlomieju:
`rusty_v8` upgrade with `Isolate::has_pending_background_tasks()` is pending in https://github.com/denoland/deno/pull/11233

core/runtime.rs
```diff
@@ -634,6 +634,8 @@ impl JsRuntime {
       state.waker.register(cx.waker());
     }
 
+    self.pump_v8_message_loop();
```
andreubotella, line 637: I had to move this to before (some of) the module loading because if a module is top-level-awaiting for wasm compilation, it might end up not being loaded yet but with no pending ops immediately after the compilation is done. I guess the same applies to ops which might be started from the microtasks after pumping the message loop, but it's not something I've run into while testing.

andreubotella (commented):
This is now ready for review.

lucacasonato:
Looks great. Could you move the integration test to a unit test (`./cli/tests/unit`)? Unit tests have the ability to detect resource / async op leakage - something integration tests do not.

core/runtime.rs
bartlomieju: This looks ok 👍  I wasn't sure where to put it, but if you confirmed that it needs to be triggered before then it's okay

core/runtime.rs
```diff
@@ -689,14 +692,20 @@ impl JsRuntime {
 
     // Check if more async ops have been dispatched
     // during this turn of event loop.
-    if state.have_unpolled_ops {
+    // If there are any pending background tasks, we also wake the runtime to
+    // make sure we don't miss them.
+    // TODO(andreubotella) The event loop will spin as long as there are pending
+    // background tasks. We should look into having V8 notify us when a
+    // background task is done.
+    if state.have_unpolled_ops || has_pending_background_tasks {
```
bartlomieju, line 700: I'm aware of this limitation, but I haven't found any method in V8 API that would do that

bartlomieju: I looked up implementation of `HasPendingBackgroundTasks()` and it is:
```c++
bool Isolate::HasPendingBackgroundTasks() {
#if V8_ENABLE_WEBASSEMBLY
  i::Isolate* isolate = reinterpret_cast<i::Isolate*>(this);
  return isolate->wasm_engine()->HasRunningCompileJob(isolate);
#else
  return false;
#endif  // V8_ENABLE_WEBASSEMBLY
}
```
Ie. only WASM compilation is considered as background pending task, I think we can leave this bit of code as is

andreubotella:
> Looks great. Could you move the integration test to a unit test (`./cli/tests/unit`)? Unit tests have the ability to detect resource / async op leakage - something integration tests do not.

Done.

cli/tests/unit/wasm_test.ts
```diff
@@ -0,0 +1,80 @@
...
+  assertEquals(add(1, 3), 4);
+});
+
+unitTest(
+  { perms: { net: true } },
+  async function wasmStreamingNonTrivial(): Promise<void> {
+    // deno-dom's WASM file is a real-world non-trivial case that gave us
+    // trouble when implementing this.
+    await WebAssembly.instantiateStreaming(fetch(
+      "https://deno.land/x/deno_dom@v0.1.3-alpha2/build/deno-wasm/deno-wasm_bg.wasm",
```
lucacasonato, line 77: Can you copy this file into `cli/tests/deno_dom_0.1.3-alpha2.wasm`, and then fetch `"http://localhost:4545/cli/tests/deno_dom_0.1.3-alpha2.wasm"`? We try to have our test suite not rely on the internet where possible.

bartlomieju (approved):
Cleanly implemented with descriptive comments. LGTM!

lucacasonato (approved):
LGTM from my side too. Great feature to have! Thanks :-)

PR #11212, opened 2021-07-01
fix(diff): better handling of text with only line ending differences

### LINKED ISSUE #8118: deno fmt --check should not show entire file when only line endings differ
When running `deno fmt --check` it should only show relevant changes, or a radius of it, it shouldn't print simple line changes, or it could result in quite bad outputs in my opinion. 

Running it, basically printed my whole codebase because of some import reordering, and organizing.

<details>
<summary>Examples</summary>

image
image

</details>

dsherret:
Also fixed a few additional bugs like:

1. It not displaying a diff when one text had a trailing newline and the other didn't.
2. Edge case where the line number width could be incorrect if the original text had say `99` lines and the edit text had `100` lines.

I commented everywhere in the PR that actually changed other than the refactor to make the borrow checker happy without duplicating code.

Closes #8118

### DIFF
--- a/cli/diff.rs
+++ b/cli/diff.rs
@@ -3,28 +3,23 @@
 use crate::colors;
 use dissimilar::{diff as difference, Chunk};
 
-fn fmt_add() -> String {
-  colors::green_bold("+").to_string()
-}
+/// Print diff of the same file_path, before and after formatting.
+///
+/// Diff format is loosely based on Github diff formatting.
+pub fn diff(orig_text: &str, edit_text: &str) -> String {
+  if orig_text == edit_text {
+    return String::new();
+  }
 
-fn fmt_add_text(x: &str) -> String {
-  colors::green(x).to_string()
-}
+  // normalize newlines as it adds too much noise if they differ
+  let orig_text = orig_text.replace("\r\n", "\n");
+  let edit_text = edit_text.replace("\r\n", "\n");
 
-fn fmt_add_text_highlight(x: &str) -> String {
-  colors::black_on_green(x).to_string()
-}
+  if orig_text == edit_text {
+    return " | Text differed by line endings.".to_string();
+  }
 
-fn fmt_rem() -> String {
-  colors::red_bold("-").to_string()
-}
-
-fn fmt_rem_text(x: &str) -> String {
-  colors::red(x).to_string()
-}
-
-fn fmt_rem_text_highlight(x: &str) -> String {
-  colors::white_on_red(x).to_string()
+  DiffBuilder::build(&orig_text, &edit_text)
 }
 
 struct DiffBuilder {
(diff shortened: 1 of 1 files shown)


cli/diff.rs
```diff
@@ -27,146 +165,60 @@ fn fmt_rem_text_highlight(x: &str) -> String {
...
-  orig.clear();
-  edit.clear();
-}
-
-/// Print diff of the same file_path, before and after formatting.
-///
-/// Diff format is loosely based on Github diff formatting.
-pub fn diff(orig_text: &str, edit_text: &str) -> String {
-  let lines = edit_text.split('\n').count();
-  let line_number_width = lines.to_string().chars().count();
```
dsherret, line 76: Bug here fixed in new code. Should take into account the max of orig_text and edit_text line count.

cli/diff.rs
```diff
@@ -27,146 +165,60 @@ fn fmt_rem_text_highlight(x: &str) -> String {
...
-
-  let mut text1 = orig_text.to_string();
-  let mut text2 = edit_text.to_string();
-
-  if !text1.ends_with('\n') {
-    text1.push('\n');
-  }
-  if !text2.ends_with('\n') {
-    text2.push('\n');
-  }
```
dsherret, line 88: Bug here. Should not do this because it should display a diff for when one has a trailing newline and the other doesn't.

cli/diff.rs
```diff
@@ -3,6 +3,144 @@
...
+
+struct DiffBuilder {
+  output: String,
+  line_number_width: usize,
+  orig_line: usize,
+  edit_line: usize,
+  orig: String,
+  edit: String,
+  has_changes: bool,
+}
```
dsherret, line 33: Unfortunately I had to change the function to a struct in order to make the borrow checker happy and not duplicate a lot of code with the new `flush_changes` method.

cli/diff.rs
```diff
@@ -3,6 +3,144 @@
...
+  if orig_text == edit_text {
+    return String::new();
+  }
+
+  // normalize newlines as it adds too much noise if they differ
+  let orig_text = orig_text.replace("\r\n", "\n");
+  let edit_text = edit_text.replace("\r\n", "\n");
+
+  if orig_text == edit_text {
+    return " | Text differed by line endings.\n".to_string();
```
dsherret, line 19: Fix here. This is what I do in dprint. It prevents entire files from being spammed.

cli/diff.rs
```diff
@@ -3,6 +3,142 @@
...
+      orig_line: 1,
+      edit_line: 1,
+      orig: String::new(),
+      edit: String::new(),
+      has_changes: false,
+      line_number_width: {
+        let line_count = std::cmp::max(
+          orig_text.split('\n').count(),
+          edit_text.split('\n').count(),
+        );
```
dsherret, line 48: Fix here to use max.

cli/diff.rs
```diff
@@ -3,6 +3,142 @@
...
+              self.flush_changes();
+            }
+            self.orig.push_str(&fmt_rem_text(s));
+            self.edit.push_str(&fmt_add_text(s));
+          }
+        }
+      }
+    }
+
+    self.flush_changes();
```
dsherret, line 94: Fix here to flush the changes here as well.

cli/diff.rs
```diff
@@ -3,6 +3,142 @@
 use crate::colors;
 use dissimilar::{diff as difference, Chunk};
 
+/// Print diff of the same file_path, before and after formatting.
+///
+/// Diff format is loosely based on Github diff formatting.
+pub fn diff(orig_text: &str, edit_text: &str) -> String {
```
dsherret, line 9: Sorry, the diff here wasn't going to be pretty no matter what, so I took the liberty of reorganizing the file. For the most part it's largely the same other than a few fixes. I've commented everywhere that's changed.

cli/diff.rs
```diff
@@ -27,146 +163,60 @@ fn fmt_rem_text_highlight(x: &str) -> String {
...
+        "6 | -'Hello World'\n",
+        "2 | +\"Hello World\"\n",
+        "7 | -)\n3 | +);\n",
+      ),
+    );
+  }
+
+  #[test]
+  fn test_eof_newline_missing() {
+    run_test(
```
bartlomieju, line 200: Nice tests 👍

bartlomieju (approved):
LGTM

PR #11214, opened 2021-07-01
ci: don't set DENO_CANARY when releasing, attempt 2

piscisaureus:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,6 +194,12 @@ jobs:
           cat >> ~/.bash_profile << ___
             cd "$(pwd)"
 
+            # Configure canary build.
+            if [ "$DENO_CANARY" == "true" ]
+            then
+              export DENO_CANARY=true
+            fi
+
             # Add cargo, rustc, and deno to $PATH.
             source /usr/share/rust/.cargo/env
             export PATH="$PATH:$(pwd)/target/release"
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):

.github/workflows/ci.yml
```diff
@@ -194,6 +194,12 @@ jobs:
           cat >> ~/.bash_profile << ___
             cd "$(pwd)"
 
+            # Configure canary build.
+            if [ "$DENO_CANARY" == "true" ]
```
ry, line 198: @piscisaureus Shouldn't this have parens?

piscisaureus: With parens, that would (try to) run the command DENO_CANARY right?

ry: Ah right. Nevermind.

.github/workflows/ci.yml
```diff
@@ -194,6 +194,12 @@ jobs:
...
+            then
+              export DENO_CANARY=true
```
maximousblk, line 200: environment variables in github actions are not persisted across steps by default. you'll have to use a special environment variable `GITHUB_ENV`

```sh
echo "DENO_CANARY=true" >> $GITHUB_ENV
```

more info: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

PR #11217, opened 2021-07-02
fix(extensions/fetch): A consumed body with a non-stream source should result in a disturbed stream

### LINKED ISSUE #11211: fetch: Getting a reader from a consumed Response with a non-stream body source should throw
```js
const res = new Response("test");
res.arrayBuffer();
// Should throw a TypeError, but doesn't.
res.body.getReader();
```

[WPT has tests for this](https://github.com/web-platform-tests/wpt/blob/master/fetch/api/response/response-stream-disturbed-5.any.js), but they only test responses returned from `fetch`. Deno passes those tests because the bug only applies responses with non-stream sources.

Added new WPT tests in web-platform-tests/wpt#29552

andreubotella:
Fixes #11211.

---------------

The tests for this PR are in web-platform-tests/wpt#29552.

### DIFF
--- a/extensions/fetch/22_body.js
+++ b/extensions/fetch/22_body.js
@@ -41,14 +41,16 @@
     get stream() {
       if (!(this.streamOrStatic instanceof ReadableStream)) {
         const { body, consumed } = this.streamOrStatic;
-        this.streamOrStatic = new ReadableStream({
-          start(controller) {
-            controller.enqueue(body);
-            controller.close();
-          },
-        });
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
The change looks good. Do you want to wait for WPT to land, or would you rather land this now and I'll enable the WPT when we roll next time?

andreubotella:
I'd rather land it now, since that would fix some of the WPT failures in #11200

Issue #10978, opened 2021-06-15
extensions/web: Improving the performance of the text encoding APIs


andreubotella:
After switching the implementation of the text encoding APIs to use `encoding_rs`, there was a user report at https://github.com/denoland/deno/pull/10844#issuecomment-855336951 that the new implementation was significantly slower than the old Javascript-based one. Although the difference was to some extent reduced with #10865, the gap with the previous implementation is still large.

I looked deep into it, and saw that in all cases there were conversions happening that weren't strictly needed. I'm splitting this into three changes so the effect of each individual change can be seen in the benchmarks:

1. The implementations of `TextEncoder.prototype.encode` and `TextEncoder.prototype.encodeInto` call `webidl.converters.USVString` with the string input, which converts it into a scalar value string. Then they call either `core.encode` or `op_encoding_encode_into`, both of which convert the Javascript string into a Rust `String` through [`rusty_v8::String::to_rust_string_lossy`](https://docs.rs/rusty_v8/0.22.3/rusty_v8/struct.String.html#method.to_rust_string_lossy) (in the case of `op_encoding_encode_into`, through the `serde_v8` bindings), which itself first converts the string into a scalar value string and then UTF-8 encodes it. Therefore, the WebIDL-level conversion to a scalar value string is unnecessary and can be removed.

2. After the first change, `TextEncoder.prototype.encodeInto` is still lagging behind. This seems to be because the `op_encoding_encode_into` operation has to run through each of the code points of the input string and UTF-8 encode them, writing the result into the output buffer. However, the input string is a Rust `String`, meaning it is UTF-8 already, and it would be enough to simply find the last UTF-8 code point boundary that fits in the output buffer, and then copy the bytes. The number of UTF-16 code units that were read must be computed, but it seems like [encoding the substring as UTF-16 to calculate the length](https://github.com/andreubotella/deno/blob/encoding-optimizations/extensions/web/lib.rs#L273) does the trick without sacrificing much performance.

[...]

PR #11219, opened 2021-07-02
refactor(extensions/web): Improve performance for TextEncoder.prototype.encodeInto()

andreubotella:
The current implementation of `op_encoding_encode_into` UTF-8 encodes each individual code point in the input string into the output buffer. But after the ops binding, the input is a Rust `String`, so the UTF-8 bytes can simply be copied to the output. This would improve this API's performance.

Part of #10978.


lucacasonato (approved):
LGTM. Thanks @andreubotella!

Issue #7674, opened 2020-09-25
WeakRefs / FinalizationRegistry bugs?


trgwii:
I was testing some of these new APIs out and it seems like Deno might be holding more references to JS objects than it should?

Running the below code with node, it outputs:
```
2020-09-25T11:50:06.491Z
2020-09-25T11:50:14.617Z meme
2020-09-25T11:50:14.617Z meme
2020-09-25T11:50:14.618Z meme
...
```

Signalling that GC has happened for each of the Foo instances, however, running the same file in Deno gives this:

```
2020-09-25T11:50:18.118Z
```

And no GC callback ever runs, which is quite interesting, I'm not sure if this could be a Deno issue or maybe caused by different V8 versions?

`node -p process.versions`:
```js
{
  node: '14.12.0',
  v8: '8.4.371.19-node.16',
  uv: '1.39.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.16.0',
  modules: '83',
  nghttp2: '1.41.0',       
  napi: '7',
  llhttp: '2.1.2',
  openssl: '1.1.1g',       
  cldr: '37.0',
  icu: '67.1',
  tz: '2020a',
  unicode: '13.0'
}
```

`deno --version`:
```
deno 1.4.1
v8 8.7.75
typescript 4.0.2
```

Edit: OS is Windows 10.0.19041

The code used in both node and deno tests:

```js
"use strict";

const reg = new FinalizationRegistry((key) => {
  console.log(new Date().toISOString(), key);
});

class Foo {
  n = Math.random();
}

console.log(new Date().toISOString());

setInterval(() => {
  reg.register(new Foo(), "meme");
}, 0);

```

TradeIdeasPhilip:
I was also having trouble with WeakRef and FinalizationRegistry in Deno.  The details were very chaotic, so I wasn't able to create a good bug report.  It seems that sometimes just adding code like
```
const x = new WeakRef({});
const y = new FinalizationRegistry(() => {});
console.log(x, y);

```
was enough to make the system fail badly.  Sometimes the code would hang.  Sometimes I'd get a nonsense exception in the built in JavaScript code, before it ever got to my code.

This worked fine, suggesting that the APIs are available.
`console.log("WeakReference",` WeakRef, "FinalizationRegistry", FinalizationRegistry);

I'd love to know what you find out.

TradeIdeasPhilip:
[...]

PR #11221, opened 2021-07-02
feat(core): pump V8 message loop on event loop tick

bartlomieju:
This commit adds support for Atomics and FinalizationRegistry by integrating
V8's message loop into "JsRuntime::poll_event_loop".

Fixes https://github.com/denoland/deno/issues/7674

### DIFF
--- a/cli/tests/finalization_registry.js
+++ b/cli/tests/finalization_registry.js
@@ -1,0 +1,20 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+"use strict";
+
+function assertEquals(a, b) {
+  if (a === b) return;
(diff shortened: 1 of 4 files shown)


cli/tests/finalization_registry.js
```diff
@@ -0,0 +1,20 @@
...
+  Deno.core.print("FinalizationRegistry called!\n");
+});
+
+(function () {
+  let x = {};
+  registry.register(x, "called!");
+  x = null;
+})();
+
+gc();
```
bartlomieju, line 20: This is an integration test because it requires a V8 flag (`--expose-gc`)

lucacasonato (approved):
LGTM! 🎉

Issue #11224, opened 2021-07-02
Use primordials everywhere


lucacasonato:
In #10939 we landed primordials into `deno_core`. They are frozen intrinsics that are not vulnerable to prototype pollution. We now need to undergo the tedious task to refactor all of our JS code to use these primordials. Here is the TODO list:

- [x] extensions/webidl/00_webidl.js (@lucacasonato) #11227
- [x] extensions/web/00_infra.js (@lucacasonato) #11228
- [x] extensions/web/01_dom_exception.js (@lucacasonato) #11228
- [x] extensions/web/01_mimesniff.js (@lucacasonato) #11228
- [x] extensions/web/02_event.js (@bartlomieju) #11264
- [x] extensions/web/02_structured_clone.js (@lucacasonato) #11273
- [x] extensions/web/03_abort_signal.js (@lucacasonato) #11273
- [x] extensions/web/04_global_interfaces.js (@lucacasonato) #11273
- [x] extensions/web/05_base64.js (@lucacasonato) #11273
- [x] extensions/web/06_streams.js (@lucacasonato) #11251 
- [x] extensions/web/08_text_encoding.js (@lucacasonato) #11273
- [x] extensions/web/09_file.js (@lucacasonato) #11299
- [x] extensions/web/10_filereader.js (@lucacasonato) #11299
- [x] extensions/web/11_blob_url.js (@lucacasonato) #11299
- [x] extensions/web/12_location.js (@lucacasonato) #11273
- [x] extensions/broadcast_channel/01_broadcast_channel.js (@littledivy) #11231
- [x] extensions/console/01_colors.js (@littledivy) #11249
- [x] extensions/console/02_console.js (@littledivy) #11249
- [x] extensions/crypto/00_webidl.js (@littledivy) #11229 
- [x] extensions/crypto/01_crypto.js (@littledivy) #11229 
- [x] extensions/fetch/01_fetch_util.js (@satyarohith) #11266
- [x] extensions/fetch/20_headers.js (@satyarohith) #11266
- [x] extensions/fetch/21_formdata.js (@satyarohith) #11266
- [x] extensions/fetch/22_body.js (@satyarohith) #11266
- [x] extensions/fetch/22_http_client.js (@satyarohith) #11266
- [x] extensions/fetch/23_request.js (@satyarohith) #11266
- [x] extensions/fetch/23_response.js (@satyarohith) #11266
- [x] extensions/fetch/26_fetch.js (@satyarohith) #11266
[...]

PR #11234, opened 2021-07-02
refactor: use primordials for extensions/timers

SimonRask:
Towards #11224

### DIFF
--- a/extensions/timers/01_timers.js
+++ b/extensions/timers/01_timers.js
@@ -12,6 +12,14 @@
     TypeError,
     Number,
     String,
+    ArrayPrototypeShift,
+    ArrayPrototypeIndexOf,
+    ArrayPrototypeSplice,
+    MapPrototypeHas,
+    MapPrototypeDelete,
+    MapPrototypeSet,
+    MapPrototypeGet,
+    FunctionPrototypeBind,
   } = window.__bootstrap.primordials;
 
   // Shamelessly cribbed from extensions/fetch/11_streams.js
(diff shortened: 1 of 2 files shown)


lucacasonato (commented):
Can you remove `OriginalDateNow` from line 303, and replace that with `DateNow` from primordials?

There is also a `pendingFireTimers.shift()` on line 331, a `pendingFireTimers.indexOf` on line 415, and a `pendingFireTimers.splice` on line 416. Also line 443 and 445, amd a `idMap.has` on line 452, a `idMap.delete` on line 458, an `eval` on line 472, a `TypeError` on 478, a `Function.prototype.bind.call` on line 493, an `idMap.set` on 523, and an `idMap.get` on 551, `idMap.delete` on 558.

SimonRask:
Alright sorry for the hassle @lucacasonato 

Having some trouble decoding how line 493 should be rewritten:
`Function.prototype.bind.call(cb, globalThis, ...args)`

Thought it would be something like this but it doesn't seem like it:
`FunctionPrototypeCall(FunctionPrototypeBind, cb, globalThis, ...args)`

Also I don't know if I'm looking in the wrong place, but I don't see `eval` in `internal.d.ts`

extensions/timers/01_timers.js
```diff
@@ -480,7 +498,7 @@
     let callback;
 
     if ("function" === typeof cb) {
-      callback = Function.prototype.bind.call(cb, globalThis, ...args);
+      callback = FunctionPrototypeBind(cb, globalThis, ...args);
```
SimonRask, line 501: @lucacasonato 
I'm not too familiar or sure about this, but the tests seem to pass

PR #11240, opened 2021-07-03
refactor: use primordials for extensions/websocket

littledivy:
Ref #11224


extensions/websocket/01_websocket.js
```diff
@@ -218,54 +250,67 @@
...
-
-            const event = new CloseEvent("close");
-            this.dispatchEvent(event);
-            tryClose(this[_rid]);
-          });
-        } else {
-          this[_readyState] = OPEN;
-          const event = new Event("open");
-          this.dispatchEvent(event);
+      PromisePrototypeCatch(
```
lucacasonato, line 265: You can combine the "catch" into the `PromisePrototypeThen` by passing the catch handler as a second callback.

lucacasonato (commented):
Can you import `Blob` from `window.__bootstrap.file` for line 48?

extensions/websocket/01_websocket.js
littledivy: done

bartlomieju (approved):
LGTM, thank you Divy

PR #11241, opened 2021-07-03
refactor: use primordials in runtime/, part1

bartlomieju:
Ref https://github.com/denoland/deno/issues/11224


lucacasonato (commented):
@bartlomieju Can you also update `wrappedHandler.handler.call` on line 34 or web_util.js?

Also `URL` on line 102 of 06_util.js comes from global scope right now. Could you instead import it from `window.__bootstrap.url`? Same for 11_workers.js.

lucacasonato (commented):
LGTM!

lucacasonato (approved):

PR #11003, opened 2021-06-16
fix(extension/fetch): Fetching with method HEAD should not have body

getspooky:
HEAD and OPTIONS requests should not have bodies https://fetch.spec.whatwg.org/#concept-main-fetch

Issue #10990, opened 2021-06-15
Fetching with method `HEAD` should not have body


lucacasonato:
HEAD and OPTIONS requests should not have bodies. You can see this issue by fetching `data:,response%27s%20body`. https://fetch.spec.whatwg.org/#concept-main-fetch step 19 says that body should be set to null for HEAD and OPTIONS.

getspooky:
Fixed https://github.com/denoland/deno/pull/11003

cmorten:
@lucacasonato / @getspooky, please point out where I'm being daft, but does step 19 not specify that the `body` should be set to `null` for `HEAD` and `CONNECT`, not `OPTION`?

> 19. If response is not a network error and either request’s method is `HEAD` or `CONNECT`, or internalResponse’s status is a null body status, set internalResponse’s body to null and disregard any enqueuing toward it (if any).

As far as am aware the [CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) method isn't the same as the [OPTIONS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS) method?

cmorten:
For further context:

> 9.2 OPTIONS
> ...
> ... The response body, if any, SHOULD also include
> information about the communication options. The format for such a
> body is not defined by this specification, but might be defined by
> future extensions to HTTP. Content negotiation MAY be used to select
> the appropriate response format. If no response body is included, the
> response MUST include a Content-Length field with a field-value of
> "0".

REF: https://datatracker.ietf.org/doc/html/rfc2616#section-9.2

Which indicates that a response body is valid, with the only condition that it should contain information r.e. the options.

I have also noticed that #11003 has introduced a potential bug whereby origin responses to `OPTIONS` requests which contain bodies that are now thrown away, have a non-zero `Content-Length` header:

```console
$ deno
Deno 1.11.5
exit using ctrl+d or close()
> await fetch("http://127.0.0.1:4505", { method: "OPTIONS" });
Response {
  body: null,
  bodyUsed: false,
  headers: Headers {
  allow: "DELETE,GET,HEAD,PUT",
  "content-length": "19",
  "content-type": "text/html; charset=utf-8",
  etag: 'W/"13-b03ce369de9cff9aa2fe20bc98b"',
  "x-powered-by": "Opine"
},
  ok: true,
  redirected: false,
  status: 200,
  statusText: "OK",
  url: "http://127.0.0.1:4505/"
[...]

PR #11242, opened 2021-07-03
fix(fetch): OPTIONS should be allowed a non-null body

cmorten:
Reverts changes from: https://github.com/denoland/deno/pull/11003
Related to: https://github.com/denoland/deno/issues/10990

### DIFF
--- a/extensions/fetch/26_fetch.js
+++ b/extensions/fetch/26_fetch.js
@@ -270,7 +270,7 @@
     if (nullBodyStatus(response.status)) {
       core.close(resp.responseRid);
     } else {
-      if (req.method === "HEAD" || req.method === "OPTIONS") {
+      if (req.method === "HEAD") {
         response.body = null;
         core.close(resp.responseRid);
       } else {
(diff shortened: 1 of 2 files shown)


kitsonk:
We should either revert #11003 @lucacasonato or this PR needs to add `CONNECT` so it matches the specification properly.

cmorten:
> We should either revert #11003 @lucacasonato or this PR needs to add `CONNECT` so it matches the specification properly.

Happy to add `CONNECT` - indeed I had it in there, but then realised that code path was likely never reached due to Deno throwing a “method forbidden” like error ( which matches Chrome + Firefox behaviour ).

REF: https://github.com/denoland/deno/blob/main/extensions/fetch/23_request.js#L139

I guess regardless of other behaviours we can align this particular unit to the spec, but we will be unable to test the behaviour because the throw overshadows

extensions/fetch/26_fetch.js
```diff
@@ -270,7 +270,7 @@
     if (nullBodyStatus(response.status)) {
       core.close(resp.responseRid);
     } else {
-      if (req.method === "HEAD" || req.method === "OPTIONS") {
+      if (req.method === "HEAD") {
```
cmorten, line 273:
```suggestion
      if (req.method === "HEAD" || req.method === "CONNECT") {
```

lucacasonato (approved):
LGTM, thanks,

PR #11243, opened 2021-07-03
refactor: use primordials in extensions/net/

bartlomieju:

### DIFF
--- a/extensions/net/01_net.js
+++ b/extensions/net/01_net.js
@@ -4,6 +4,12 @@
 ((window) => {
   const core = window.Deno.core;
   const { BadResource } = core;
+  const {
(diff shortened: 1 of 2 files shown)


satyarohith (changes requested):
I think line 49 (57 with new changes) needs to be updated.
```js
       if (error instanceof BadResource) {
          return null;
        } else if (error instanceof Interrupted) {
          return null;
        } else if (error.message.includes("connection closed")) {
          return null;
        }
```

satyarohith (approved):
LGTM.

lucacasonato (commented):
You missed an `Symbol.asyncIterator` on line 144 of 01_net.js

PR #11244, opened 2021-07-03
feat(cli/lint): pretty print lint rule docs to the terminal

### LINKED ISSUE #8227: Linter should display rule docs
There's an on-going effort in `deno_lint` repository to add docs and examples to all rules (https://github.com/denoland/deno_lint/issues/159). We should think about integrating these docs into `deno lint` subcommand. 

Currently `deno lint --rules` shows a plain list of rules that are run. I think we could integrate showing docs by providing rule codes as a value to `--rules`, eg. `deno lint --rules no-explicit-any no-empty`.

Documentation is accessible using `LintRule::docs` method.

I also tried integrating markdown printer to nicely display docs in `dlint` example (https://github.com/denoland/deno_lint/pull/441) but haven't gotten too far.

magurotuna:
This PR adds a new feature to `deno lint` that lint rule docs are displayed in the terminal with syntax highlight.

### Usage

- `deno lint --rule no-with`
  - This is a basic usage, printing one rule specified as an argument to the terminal 
image
- `deno lint --rule no-with no-debugger`
  - supports outputting multiple docs at the same time
- `deno lint --rule no-with --json`
  - JSON output is also available (no syntax highlight though)

Resolves #8227
Ref https://github.com/denoland/deno_lint/pull/756


bartlomieju (commented):
This PR brings a big dependency in form of `pretty_markdown` - effectively including another Markdown parser into the codebase (after `dprint-plugin-markdown`). That said I believe this PR is a great quality-of-life improvement and we could the same infrastructure for the output of `deno doc` (instead of implementing poor man's coloring out output in `deno_doc` crate). 

Before landing I will wait for @ry's review

magurotuna:
Ah OK, I'm not particular about which markdown parser we'll utilize as long as it fits. It looks like `dprint-plugin-markdown` depends on [`pulldown-cmark`](https://github.com/raphlinus/pulldown-cmark) so I'll check if this crate also works.

cli/flags.rs
```diff
@@ -918,6 +923,13 @@ Ignore linting a file by adding an ignore comment at the top of the file:
         .long("rules")
         .help("List available rules"),
     )
+    .arg(
+      Arg::with_name("rule")
+        .long("rule")
+        .takes_value(true)
+        .multiple(true)
+        .help("Display docs of the specified rules"),
```
caspervonb, line 931: Nit on the help text, maybe something like:
```suggestion
        .help("Display help for the specified rule"),
```

cli/tools/repl.rs
```diff
@@ -252,52 +253,7 @@ impl Highlighter for EditorHelper {
...
-          },
-          TokenOrComment::Comment { .. } => {
-            colors::gray(&line[span]).to_string()
-          }
-        },
-      );
-    }
-
-    out_line.into()
+    highlight_line(line, &MediaType::TypeScript).into()
```
caspervonb, line 256: Nice

caspervonb (commented):
Bike-shedding territory but maybe we can refer to this as 'help' rather than 'docs' in public help lines and json fields? Just to keep a clear distinction between docs (the doc tool) and other kinds of help like this.

cli/flags.rs
magurotuna: To me "help" sounds like something that tells us how to deal with diagnostics. It's partially true but there might be a more appropriate word like... "description"?

magurotuna:
According to the design meeting held on July 8th (#11365), it seems like this feature is not what we should land. Close.

PR #11247, opened 2021-07-03
refactor: use primordials for `13_buffer.js` and `30_fs.js`

SimonRask:
Towards #11224


lucacasonato (approved):
LGTM, thanks @SimonRask

PR #11248, opened 2021-07-03
refactor: use primordials in runtime/, part2

bartlomieju:
Ref #11224

### DIFF
--- a/runtime/js/30_os.js
+++ b/runtime/js/30_os.js
@@ -4,6 +4,7 @@
 ((window) => {
   const core = window.Deno.core;
   const {
+    Error,
     SymbolFor,
   } = window.__bootstrap.primordials;
 
--- a/runtime/js/30_os.js
+++ b/runtime/js/30_os.js
@@ -3,6 +3,9 @@
 
 ((window) => {
   const core = window.Deno.core;
+  const {
+    SymbolFor,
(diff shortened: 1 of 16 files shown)


runtime/js/30_os.js
```diff
@@ -3,6 +3,9 @@
 
 ((window) => {
   const core = window.Deno.core;
+  const {
+    SymbolFor,
```
SimonRask, line 7: You can add `Error` here

runtime/js/41_prompt.js
```diff
@@ -6,6 +6,9 @@
   const LF = "\n".charCodeAt(0);
```
SimonRask, line 6: `StringPrototypeCharCodeAt`

runtime/js/41_prompt.js
```diff
@@ -6,6 +6,9 @@
...
   const CR = "\r".charCodeAt(0);
   const core = window.Deno.core;
+  const {
+    ArrayPrototypePush,
```
SimonRask, line 10: `Uint8Array` can be added

SimonRask:
In 40_signals.js line 254, I think you can use `PromisePrototypeThen`
In 40_testing.js line 155, you can use `ArrayPrototypePush`

lucacasonato (approved):
LGTM!

PR #11262, opened 2021-07-03
feat(cli/tools/test_runner): close test worker once tests have completed

caspervonb:
This moves everything that is the _test runner_ into _cli_.

The test harness itself, that is the sanitizers, the array that serves as the test registry and public function that allows users to register tests still lives in the _runtime_.


caspervonb:
Picked this up again, just about done:
- [ ] Add back coverage collection
- [ ] Add back lifecycle events
- [ ] Add a test-case for close when there are still pending ops in the event loop but no more tests to run.

caspervonb:
All the good refactors have been extracted.
Closing this out in favour of #11727

PR #11264, opened 2021-07-04
refactor: use primordials in extensions/web/02_event.js

bartlomieju:


extensions/web/02_event.js
```diff
@@ -138,13 +167,13 @@
         target: null,
         timeStamp: Date.now(),
```
satyarohith, line 168:
```suggestion
        timeStamp: DateNow(),
```

satyarohith (approved):
LGTM, but see the comment.

SimonRask:
`WeakMapPrototypeGet` on line 1014 I think :)

PR #11265, opened 2021-07-04
refactor use primordials in extensions/webgpu

bartlomieju:
https://github.com/denoland/deno/issues/11224

### DIFF
--- a/extensions/webgpu/01_webgpu.js
+++ b/extensions/webgpu/01_webgpu.js
@@ -13,6 +13,32 @@
   const webidl = window.__bootstrap.webidl;
   const eventTarget = window.__bootstrap.eventTarget;
   const { DOMException } = window.__bootstrap.domException;
+  const {
+    ArrayBuffer,
+    ArrayBufferIsView,
+    ArrayIsArray,
+    ArrayPrototypeMap,
+    ArrayPrototypePop,
+    ArrayPrototypePush,
+    Error,
+    MathMax,
+    ObjectDefineProperty,
+    ObjectFreeze,
+    Promise,
+    PromiseResolve,
+    Set,
+    SetPrototypeEntries,
+    SetPrototypeForEach,
+    SetPrototypeHas,
+    SetPrototypeKeys,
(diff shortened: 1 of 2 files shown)


satyarohith (approved):
WebGPU flies are huge. I should've checked the line count of the files before I started the review. 😄 

Nice work, Bartek. I couldn't find a single instance unaddressed.

LGTM.

SimonRask:
Found some things in `01_webgpu.js`:
- line 615
- line 1761

Found nothing in `02_idl_types.js`

satyarohith:
Thanks for taking a look, @SimonRask!

CrimsonCodes0:
There were a few accesses on TypedArray/DataView objects that aren't using primordials here, e.g. directly accessing the `.buffer` property.

bartlomieju:
> There were a few accesses on TypedArray/DataView objects that aren't using primordials here, e.g. directly accessing the `.buffer` property.

@CrimsonCodes0 I believe this is the same situation as with `.length`. Node doesn't really use primordials for such accesses.

CrimsonCodes0:
> > There were a few accesses on TypedArray/DataView objects that aren't using primordials here, e.g. directly accessing the `.buffer` property.
> 
> @CrimsonCodes0 I believe this is the same situation as with `.length`. Node doesn't really use primordials for such accesses.

String length are array length properties are safe to access, as that is the only way to access them. String have an immutable own property length, while arrays have a "magic" own property that has setting intercepted by the engine to cast the written value to an unsigned integer.

TupedArray properties are very different, in that they aren't nearly as safe, nor are they own properties. `%TypedArray%.prototype.{buffer, byteLength, length}` are all getters that exist on the `%TypedArray%` prototype, amd thus are able to be mutated by user code.

If it would be appreciated, I can make some PRs tommorow to implement these accessors and use them across Deno's builtin JavaScript.

bartlomieju:
@CrimsonCodes0 thanks explanation. Please do open a PR.

PR #11279, opened 2021-07-05
test(integration/test): be more liberal with expectation wildcards

caspervonb:
Test expectations have gotten a bit too strict to the point where a new line change will break the stack.
This applies more liberal use of wildcards in the stack traces where failures are displayed.


lucacasonato:
Can you make sure the stack frame for the test itself is still tested? So just wildcard all the internal frames?

caspervonb:
> Can you make sure the stack frame for the test itself is still tested? So just wildcard all the internal frames?

Sure no problem!

lucacasonato (approved):
LGTM, thanks.

PR #11282, opened 2021-07-05
fix(core): Delay deadlock detection for dynamic imports

### LINKED ISSUE #11238: Can't nest top-level dynamic imports
## Issue

In a project, I try to dynamically import a module which also dynamically imports modules (at its top level). The execution freezes without throwing an exception. 

After generating minimal example to reproduce, I get the following error message:

> error: Dynamically imported module evaluation is still pending but there are no pending ops. This situation is often caused by unresolved promise.

I'm not sure if this is intended to work or I'm doing it wrong, but this would be generally useful for easy & flexible dynamic importing.

## Repro

Create the below files and then run:
`deno run --allow-read app.ts`

The expected output is: _"It worked"_ - but instead the above error is displayed.

### app.ts
```js
const mod = await import("./import_me_first.ts");
mod.default();
```

### import_me_first.ts
```js
const mod = await import("./import_me_next.ts");

export default () => {
  mod.default();
};
```

### import_me_next.ts
```js
export default () => console.log('It worked')
```

nayeemrmn:
Fixes #11238.


bartlomieju (approved):
LGTM, good fix

PR #11283, opened 2021-07-05
perf: opt-level = 3 for deno_net

lucacasonato:
14% perf improvement.


Cargo.toml
```diff
@@ -54,6 +54,8 @@ opt-level = 3
 opt-level = 3
 [profile.bench.package.deno_runtime]
 opt-level = 3
+[profile.release.package.deno_net]
```
bartlomieju, line 57:
```suggestion
[profile.bench.package.deno_net]
```

bartlomieju (approved):

Issue #3429, opened 2019-12-02
Consider parking_lot crate for locks


bartlomieju:
Ref #3375 - in that PR  @afinch7 introduced `RwLock` on op registry so ops can be dynamically registered from other ops. That was done due to fact that there's already lock held on `Isolate`.

I've been wondering on this subject a bit... There's this ['parking_lot`](https://docs.rs/parking_lot/0.10.0) crate that might have a couple of handy structures for our use case:
- [`ReentrantMutex`](https://docs.rs/parking_lot/0.10.0/parking_lot/type.ReentrantMutex.html) - this mutex could replace mutex on `Isolate` and in turn remove mutex from op registry
- [`RwLock`](https://docs.rs/parking_lot/0.10.0/parking_lot/type.RwLock.html) - this lock might be used for `ResourceTable` - it's a bit better than standard library `RwLock`.  I also have my eyes on [concurrent hash map](https://github.com/jonhoo/flurry) for resource table, but that's for the future. 

Another thing is that we block everywhere on mutex acquisition 😢It might be worthwhile to explore using [non-blocking calls](https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.try_lock).

All of this is just speculation, I'll give it a shot and see the results.

PR #11289, opened 2021-07-05
chore: use parking_lot instead of std::sync for locks

dsherret:
We're already using parking_lot extensively, but transitvely in dependencies. For example, parking_lot is used by tokio instead of std::sync for locks as a performance optimization. It might be good for us to also use it as it uses less memory, is faster, and has more features (I've used it in my projects and the API is a bit better).

Relates to #3429.

From https://crates.io/crates/parking_lot

> When tested on x86_64 Linux, parking_lot::Mutex was found to be 1.5x faster than std::sync::Mutex when uncontended, and up to 5x faster when contended from multiple threads. The numbers for RwLock vary depending on the number of reader and writer threads, but are almost always faster than the standard library RwLock, and even up to 50x faster in some cases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -630,6 +630,7 @@ dependencies = [
  "lazy_static",
  "libc",
  "log",
+ "parking_lot",
  "pin-project",
  "rusty_v8",
  "serde",
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -17,6 +17,7 @@ use deno_core::error::uri_error;
 use deno_core::error::AnyError;
 use deno_core::futures;
 use deno_core::futures::future::FutureExt;
+use deno_core::parking_lot::Mutex;
 use deno_core::ModuleSpecifier;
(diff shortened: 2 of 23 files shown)


bartlomieju (approved):
If benchmarks are favorable then LGTM.

dsherret:
@bartlomieju benchmarks look basically the same. I don't think this change affects much in the benchmarks, though it would be nicer to use.

ry (commented):
I’m skeptical about using non-std functionality for performance reasons, when the benefit hasn’t been demonstrated.

I’m not necessarily against it but deferring to @piscisaureus for review.

dsherret:
Makes sense. I suspect if we turned off the parking_lot feature in tokio we would notice a difference, but yeah after running the benchmarks I see no performance benefit of this PR. The API is better and this PR makes us use parking_lot more consistently though. Perhaps some future code might benefit from it being faster. I don't have strong feelings though.

piscisaureus (approved):

Issue #11257, opened 2021-07-03
Printing `Request.prototype` in REPL results in panic


lucacasonato:
On commit 7ef0f43d87e3e45e8084bb7ae0dee2053968c010

```
~/p/g/d/deno ❯❯❯ cargo run
   Compiling deno v1.11.3 (/mnt/starship/Projects/github.com/denoland/deno/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 30.00s
     Running `target/debug/deno`
Deno 1.11.3
exit using ctrl+d or close()
> Request.prototype
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', cli/tools/repl.rs:575:45
stack backtrace:
   0: rust_begin_unwind
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/panicking.rs:50:5
   3: core::option::Option<T>::unwrap
             at /home/lucacasonato/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:386:21
   4: deno::tools::repl::ReplSession::get_eval_value::{{closure}}
             at ./cli/tools/repl.rs:575:17
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /home/lucacasonato/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80:19
   6: deno::tools::repl::ReplSession::evaluate_line_and_get_output::{{closure}}
             at ./cli/tools/repl.rs:469:21
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /home/lucacasonato/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/mod.rs:80:19
   8: deno::tools::repl::run::{{closure}}
             at ./cli/tools/repl.rs:687:22
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
[...]

PR #11292, opened 2021-07-06
fix(cli/tools/repl): do not panic when inspect throws

caspervonb:
Fixes https://github.com/denoland/deno/issues/11257


bartlomieju:
Can we get a test case?

caspervonb:
> Can we get a test case?

Not entirely sure what webidl does here to make it throw in the first place 🤔 
`Deno.inspect` shouldn't actually ever throw under _normal_ circumstances as we avoid inspecting getters and such.

bartlomieju:
I believe tests are failing because you got a rogue checkout of `wpt/`

bartlomieju (approved):
LGTM thanks

PR #11296, opened 2021-07-06
fix(cli/tools/upgrade): check if passed version is valid

### LINKED ISSUE #11237: deno upgrade --version <v>: v isn't escaped
When running `deno upgrade --version ../../../etc` you can change the download url. *v* should be properly escaped.

image

crowlKats:
Closes #11237
Also minor clean up that didnt feel like a separate PR was necesary

### DIFF
--- a/cli/tools/upgrade.rs
+++ b/cli/tools/upgrade.rs
@@ -2,7 +2,7 @@
 
 //! This module provides feature to upgrade deno executable
 
-use deno_core::error::AnyError;
+use deno_core::error::{AnyError, bail};
 use deno_core::futures::StreamExt;
 use deno_runtime::deno_fetch::reqwest;
 use deno_runtime::deno_fetch::reqwest::Client;
@@ -44,11 +44,9 @@ pub async fn upgrade_command(
       if canary
         && !regex::Regex::new("^[0-9a-f]{40}$")?.is_match(&passed_version)
       {
-        eprintln!("Invalid commit hash passed");
-        std::process::exit(1)
+        bail!("Invalid commit hash passed");
       } else if !canary && semver_parse(&passed_version).is_err() {
-        eprintln!("Invalid semver passed");
-        std::process::exit(1)
+        bail!("Invalid semver passed");
       }
 
       let current_is_passed = if canary {
--- a/cli/tools/upgrade.rs
+++ b/cli/tools/upgrade.rs
@@ -2,7 +2,7 @@
 
 //! This module provides feature to upgrade deno executable
 
-use deno_core::error::{AnyError, bail};
+use deno_core::error::{bail, AnyError};
 use deno_core::futures::StreamExt;
 use deno_runtime::deno_fetch::reqwest;
 use deno_runtime::deno_fetch::reqwest::Client;
--- a/cli/tools/upgrade.rs
+++ b/cli/tools/upgrade.rs
@@ -41,6 +41,16 @@ pub async fn upgrade_command(
 
   let install_version = match version {
     Some(passed_version) => {
+      if canary
+        && !regex::Regex::new("^[0-9a-f]{40}$")?.is_match(&passed_version)
+      {
+        eprintln!("Invalid commit hash passed");
+        std::process::exit(1)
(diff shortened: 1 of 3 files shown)


cli/tools/upgrade.rs
```diff
@@ -41,6 +41,16 @@ pub async fn upgrade_command(
 
   let install_version = match version {
     Some(passed_version) => {
+      if canary
+        && !regex::Regex::new("^[0-9a-f]{40}$")?.is_match(&passed_version)
```
littledivy, line 45: maybe create the regexp at top level
```rust
lazy_static! {
  static ref VERSION_STRING_RE: Regex = regex::Regex::new("^[0-9a-f]{40}$").unwrap();
}

Comment (crowlKats) [reply cli/tools/upgrade.rs:45]: for what reason? this is used only there

Comment (littledivy) [reply cli/tools/upgrade.rs:45]: because it is [recommended](https://github.com/rust-lang/regex/blob/master/PERFORMANCE.md) by regexp-rs and is consistent with rest of the codebase.

- https://github.com/denoland/deno/blob/f46e39c5c5435696daab9df075482e6e5c83533b/runtime/colors.rs
- https://github.com/denoland/deno/blob/8708d3c0451129792d58b7fa856101ceaa7bf487/test_util/src/lsp.rs

Comment (crowlKats) [reply cli/tools/upgrade.rs:45]: It says in the scenario of a loop/it being called multiple times. in that case it makes sense, but in the case here, this regex is executed only once ever. changing it to lazy_static would have no performance benefit

Comment (dsherret) [reply cli/tools/upgrade.rs:45]: Yeah, inlining it is fine as it will only be executed once.

Comment (dsherret) [inline cli/tools/upgrade.rs:45]: Should this support upper case hashes too? (Add `A-F`? Or is it already case insensitive?)

File: cli/tools/upgrade.rs [hunk 2]
```diff
@@ -41,6 +41,16 @@ pub async fn upgrade_command(
...
+      {
+        eprintln!("Invalid commit hash passed");
+        std::process::exit(1)
+      } else if !canary && semver_parse(&passed_version).is_err() {
+        eprintln!("Invalid semver passed");
+        std::process::exit(1)
```
dsherret, line 51: I get that this is copying what's already done in this file, but wouldn't it be better to return errors rather than exiting the process here? It would be easier to add some tests for these failing scenarios in that case too.

bartlomieju:
@crowlKats could we get a test case exercising this change?

crowlKats:
> @crowlKats could we get a test case exercising this change?

well, sure, but we dont run any of the `upgrade` tests because flaky on CI

cli/tools/upgrade.rs
crowlKats: Pretty sure git hashes are just lowercase

cli/tools/upgrade.rs
crowlKats: not sure what you mean

dsherret: So instead of exiting the process, doing something along the lines of:

```ts
return Err(type_error("Invalid semver passed"));
```

crowlKats: `type_error` is used more for ops because it specifies the TypeError class. don't see why we should use that for subcommands; it also would be a first for that usecase

lucacasonato: Then use `custom_error` or `anyhow!()`

crowlKats: done.

cli/tools/upgrade.rs
littledivy: sounds good 👍

bartlomieju:
> > @crowlKats could we get a test case exercising this change?
> 
> well, sure, but we dont run any of the `upgrade` tests because flaky on CI

Yikes, I forgot about that 😬 still, if you could write a test case test case that passes locally but might fail in CI that would be very helpful 👍

dsherret (approved):
LGTM. Thanks @crowlKats!

PR #11297, opened 2021-07-06
perf: Use the WebIDL conversion to DOMString rather than USVString for TextEncoder

andreubotella:
This works since both `core.encode` and the ops bindings to a Rust `String` will already replace any lone surrogates with the replacement character.

Part of #10978.


lucacasonato (approved):
LGTM, thanks!

PR #11308, opened 2021-07-06
fix: fix inspecting prototypes of classes with custom inspect implementations

### LINKED ISSUE #11293: Prototypes are not inspectable for many classes with custom inspect implementations
Continuing from #11257 which was specifically about the panic in the REPL.
Our WebIDL assertions makes it so that prototypes aren't inspectable, for example both of these throw:

```ts
console.log(Response.prototype);
console.log(Request.prototype);
```

dsherret:
Closes #11293

Expands on #10979 to ensure the following code works:

```
Deno.inspect(Event.prototype); // this one already works because I fixed this earlier in 10979
Deno.inspect(Request.prototype);
Deno.inspect(Response.prototype);
Deno.inspect(PerformanceEntry.prototype);
Deno.inspect(PerformanceMark.prototype);
Deno.inspect(Performance.prototype);
Deno.inspect(Blob.prototype);
Deno.inspect(DOMException.prototype);
Deno.inspect(ByteLengthQueuingStrategy.prototype);
Deno.inspect(ReadableByteStreamController.prototype);
Deno.inspect(ReadableStreamDefaultController.prototype);
Deno.inspect(TransformStreamDefaultController.prototype);
Deno.inspect(WritableStreamDefaultWriter.prototype);
Deno.inspect(WritableStreamDefaultController.prototype);
Deno.inspect(CountQueuingStrategy.prototype);
```


extensions/console/02_console.js
```diff
@@ -1958,6 +1962,60 @@
     });
   }
 
+  /** Creates a proxy that represents a subset of the properties
+   * of the original object optionally without evaluating the properties
+   * in order to get the values. */
+  function createFilteredInspectProxy({ object, keys, evaluate }) {
```
dsherret, line 1968: This was taken and improved from my PR in #10979. I'm not sure about it existing in this file though and could use some guidance on what would be best for us to do here. On a related note, perhaps there is a better overall solution?

caspervonb: Since this is working around branded assertions, could probably move this to webidl so that's the only shared dependency.

extensions/console/internal.d.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/// <reference no-default-lib="true" />
+/// <reference lib="esnext" />
+
+declare namespace globalThis {
```
caspervonb, line 6: What's this? we've never strongly typed internals before.

dsherret: There are other internal.d.ts files. I was gradually typing here.

caspervonb: Ah nope nevermind, we do and I'm wrong.

extensions/console/02_console.js
dsherret: The problem would exist before the branded assertions were added so it's not webidl related. When inspecting the prototype it would evaluate the properties, but now it doesn't.

lucacasonato (approved):

extensions/console/02_console.js
caspervonb: Sure custom inspect can throw by design but it's not applicable for any classes in say the Deno namespace like `Deno.File` where we don't configure the prototype to throw.

dsherret: It's not webidl specific or about configuring the prototype to throw though. Errors may occur since `this` does not refer to the instance of the class. Here's an isolated example that shows what the code was like before:

```ts
class MyClass {
  #myProperty = 5;

  get myProperty() {
    return this.#myProperty;
  }
  
  [Symbol.for("Deno.customInspect")](inspect) {
    const inner = {
      myProperty: this.myProperty
    };
    return `MyClass ${inspect(inner)}`;
  }
}

Deno.inspect(new MyClass()); // ok
Deno.inspect(MyClass.prototype); // throws
```

Also, I believe the console extension crate is necessary to inspect.

caspervonb: >  Here's an isolated example that shows what the code was like before:

Right privates are a thing now, that tracks.

> Also, I believe the console extension crate is necessary to inspect.

Since we're removing the distinct inspect symbol, more of an abstract contract.

Anyway, here in the console crate is fine as-well just suggested it since you asked 😁

bartlomieju (approved):
LGTM

PR #11309, opened 2021-07-07
refactor(extensions/crypto): use key::CryptoHash with digest

littledivy:
removes the `i8 => digest::Algorithm` mapping between JS and Rust


bartlomieju (approved):
LGTM, this is cleaner

PR #11310, opened 2021-07-07
tests: Escape lone surrogates in wptreport test names and messages

### LINKED ISSUE #11305: WPT test names with lone surrogates should be escaped in wpt.fyi
https://wpt.fyi/results/encoding/encodeInto.any.html?label=master&label=experimental&product=chrome&product=firefox&product=safari&product=deno&aligned

`encoding/encodeInto.any.js` has a number of tests whose names contain lone surrogates, ~and Deno seems to be reporting them with the lone surrogate escaped as, say, `U+df06`. In wpt.fyi those tests show up twice: once with the surrogate, which runs in all browsers but not in Deno, and once escaped, which only runs in Deno.~

andreubotella:
Closes #11305.

### DIFF
--- a/tools/wpt.ts
+++ b/tools/wpt.ts
@@ -14,6 +14,7 @@ import {
   autoConfig,
   cargoBuild,
   checkPy3Available,
+  escapeLoneSurrogates,
   Expectation,
   generateRunInfo,
   getExpectation,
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM! Thanks.

PR #11311, opened 2021-07-07
fix(cli/tools/lint): output json reports to stdout

caspervonb:
This changes the output stream for `--json` to stdout rather than stderr keeping stderr free for diagnostics output without poisoning the requested json output.


bartlomieju (approved):
I'm not sure why we used stderr in the first place... LGTM

PR #11325, opened 2021-07-08
feat(repl): support autocomplete on declarations containing a primitive

### LINKED ISSUE #11322: REPL autocomplete should work for assigned primitives
For example:

```ts
let a = "cow"
a.<tab>
```

https://discord.com/channels/684898665143206084/689420767620104201/862418713708593152

dsherret:
These now work:

```ts
let str = "cow"
str.<tab>
let num = 5;
num.<tab>
5n.<tab>
let func = function test(){};
func.<tab>
```

Unfortunately the following still does not work (it doesn't work in node's repl either and it would require improving `get_expr_from_line_at_pos` to probably tokenize to properly find the expression):

```ts
"test".<tab>
```

Closes #11322

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -169,6 +169,8 @@ fn pty_complete_primitives() {
     master.write_all(b"str.leng\t\n").unwrap();
     master.write_all(b"false.valueO\t\n").unwrap();
     master.write_all(b"5n.valueO\t\n").unwrap();
+    master.write_all(b"let num = 5\n").unwrap();
+    master.write_all(b"num.toStrin\t\n").unwrap();
     master.write_all(b"close();\n").unwrap();
 
     let mut output = String::new();
@@ -178,6 +180,7 @@ fn pty_complete_primitives() {
     assert!(output.contains("> str.length"));
     assert!(output.contains("> 5n.valueOf"));
     assert!(output.contains("> false.valueOf"));
+    assert!(output.contains("> num.toString"));
 
     fork.wait().unwrap();
   } else {
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -157,6 +157,39 @@ fn pty_complete_declarations() {
 
 #[cfg(unix)]
 #[test]
+fn pty_complete_primitives() {
+  use std::io::{Read, Write};
+  use util::pty::fork::*;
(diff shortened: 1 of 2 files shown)


caspervonb:
I'm likely making this obsolete in a day or two, intend to parse the input line into a full AST and then do completion based on that because the current scanning leaves much to be desired.

dsherret:
@caspervonb this change would still be necessary as it doesn't fix the issue of `get_expr_from_line_at_pos` lacking (ex. the `"test".len<tab>` example in the PR description).

I'm not sure about parsing to a full AST as swc is not very good at dealing with partial code. You might have better luck with just the lexer.

caspervonb:
> I'm not sure about parsing to a full AST as swc is not very good at dealing with partial code. You might have better luck with just the lexer.

Nothing incomplete about the line tho, the line buffer includes everything you've typed'.

caspervonb (approved):

cli/tests/integration/repl_tests.rs
```diff
@@ -173,7 +146,16 @@ fn pty_ignore_symbols() {
...
+  });
+}
 
+#[cfg(unix)]
+fn run_pty_test(mut run: impl FnMut(&mut util::pty::fork::Master)) {
+  use util::pty::fork::*;
+  let deno_exe = util::deno_exe_path();
+  let fork = Fork::from_ptmx().unwrap();
+  if let Ok(mut master) = fork.is_parent() {
+    run(&mut master);
```
bartlomieju, line 158: Nice cleanup!

cli/tests/integration/repl_tests.rs
```diff
@@ -145,24 +102,40 @@ fn pty_complete_declarations() {
...
+    master.write_all(b"close();\n").unwrap();
+
+    let mut output = String::new();
+    master.read_to_string(&mut output).unwrap();
+
+    assert!(output.contains("> func.apply"));
+    assert!(output.contains("> str.length"));
+    assert!(output.contains("> 5n.valueOf"));
+    assert!(output.contains("> false.valueOf"));
+    assert!(output.contains("> num.toString"));
```
bartlomieju, line 130: Nice test!

bartlomieju (approved):
LGTM

caspervonb:
Discussed offline, I'll probably lex because the parser isn't as forgiving.
Tested manually, good to merge imo.

PR #11326, opened 2021-07-08
refactor(cli/tools/coverage): strongly type inspector values

caspervonb:
This adds local methods and types for the rest of the messages, parameters and return objects that coverage uses as a first pass towards a fully typed unified inspector client interface.

### DIFF
--- a/cli/tools/coverage.rs
+++ b/cli/tools/coverage.rs
@@ -27,6 +27,53 @@ use std::path::PathBuf;
 use swc_common::Span;
 use uuid::Uuid;
 
+// TODO(caspervonb) all of these structs can and should be made private, possibly moved to
+// inspector::protocol.
+#[derive(Debug, Serialize, Deserialize, Clone)]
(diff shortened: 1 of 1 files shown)


cli/tools/coverage.rs
```diff
@@ -37,31 +37,76 @@ impl CoverageCollector {
...
+  async fn disable_profiler(&mut self) -> Result<(), AnyError> {
+    self.session.post_message("Profiler.disable", None).await?;
+
+    Ok(())
+  }
+
+  async fn start_precise_coverage(
+    &mut self,
+    parameters: StartPreciseCoverageParameters,
+  ) -> Result<StartPreciseCoverageReturnObject, AnyError> {
```
caspervonb, line 67: Could be cut this down quite a bit if parameters and return object were generalised on `self.post_message` and the stubs only helped with type inference.

bartlomieju (approved):

PR #11332, opened 2021-07-08
chore: update crates

lucacasonato:


bartlomieju (changes requested):
Please bump versions in `Cargo.toml` files as well

bartlomieju (approved):

PR #11337, opened 2021-07-08
refactor(cli/tools/test): only use one tokio runtime

caspervonb:

### DIFF
--- a/cli/tools/test_runner.rs
+++ b/cli/tools/test_runner.rs
@@ -481,6 +481,7 @@ pub async fn run_tests(
   program_state.file_fetcher.insert_cached(test_file);
 
   let (sender, receiver) = channel::<TestEvent>();
+  let handle = tokio::runtime::Handle::current();
 
   let join_handles = test_modules.iter().map(move |main_module| {
     let program_state = program_state.clone();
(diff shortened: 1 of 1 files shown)


caspervonb:
Need to do a release build and benchmark this to make sure there aren't any crazy regressions.

lucacasonato:
Why? As far as I remember we discussed this previously.

caspervonb:
> Why? As far as I remember we discussed this previously.

Prompted by @bartlomieju and ended up looked into, turns out we could drop the extra thread.

As far as I recall, that was about using a `LocalSet` and `spawn_local`.
This still uses blocking threads from tokio's pool so synchronous ops still shouldn't block more than a single worker and its thread.

We're just dropping the extra overhead of yet another thread and scheduler.
There's also some yet to be dealt with complexity in watch mode, it actually doesn't ever abort so this is less stuff to join or kill.

Is there a strong reason to keep the additional scheduler that I've forgotten?

caspervonb:
Performance wise it's pretty much the same.

This branch:
```
test result: FAILED. 2345 passed; 8 failed; 10 ignored; 0 measured; 0 filtered out (39911ms)
real    0m41.081s
user    2m5.690s
sys     0m5.832s
```

Canary:
```
test result: FAILED. 2345 passed; 8 failed; 10 ignored; 0 measured; 0 filtered out (41713ms)
real    0m42.793s
user    2m16.839s
sys     0m6.953s
```

caspervonb:
> Is there a strong reason to keep the additional scheduler that I've forgotten?

Scheduler no, but thread locals, yes.

PR #11338, opened 2021-07-08
feat(runtime): Implement classic workers

andreubotella:
Part of #10903.


andreubotella:
Some of the WPT test failures are panics. Also, I haven't enabled `workers/modules/dedicated-worker-import-data-url.any.worker.html` because it gets stuck, even though the `.any.html` test panics.

I'll be enabling the rest of `.worker.html` tests in a follow-up PR.

lucacasonato (commented):
Firstly, this is awesome and I am very happy we will be able to test workers with WPT. Great work. I have some concerns though:

- It is unfortunate that this requires fetching logic separate from modules. I would rather we use the same file loader as modules (that also supports file://). Doing this would also resolve my second issue:
- There are no permission checks. Even though this is `--unstable` with no intent to stabilize, we should try to keep the sandbox and permission system intact.

I am also thinking that maybe we want to add a hidden `--enable-testing-features-do-not-use` flag to unlock this feature. The flag should also print out a really obnoxious red warning text on startup that can't be silenced with `--quiet`. This would be akin to the `--expose-internals-for-testing ` in Chromium.  @bartlomieju @ry opinions?

bartlomieju:
@lucacasonato SGTM, but all the relevant code parts should be commented that they are only exposed in that particular case

andreubotella:
> Firstly, this is awesome and I am very happy we will be able to test workers with WPT. Great work. I have some concerns though:
> 
>     * It is unfortunate that this requires fetching logic separate from modules. I would rather we use the same file loader as modules (that also supports file://). Doing this would also resolve my second issue:
> 
>     * There are no permission checks. Even though this is `--unstable` with no intent to stabilize, we should try to keep the sandbox and permission system intact.

I initially put this together quickly as a proof of concept, and the sync fetch implementation is all sorts of inefficient (it waits until every script is fetched before running any of them). So I'm fine with changing it to use module fetching logic. However, I'm not too familiar with how module fetching works, and making it usable from runtime doesn't seem too easy on a first glance, especially synchronously. There are also differences in how module imports and `importScripts()` work – the ones that come to mind are that `importScripts()` doesn't support import maps, and that we probably don't want it to run TypeScript.

> I am also thinking that maybe we want to add a hidden `--enable-testing-features-do-not-use` flag to unlock this feature. The flag should also print out a really obnoxious red warning text on startup that can't be silenced with `--quiet`. This would be akin to the `--expose-internals-for-testing ` in Chromium. @bartlomieju @ry opinions?

I'm all for classic workers being implemented behind a hidden flag. That would also help when https://jgraham.github.io/browser-test/ gets adopted by the WHATWG (see whatwg/sg#164).

runtime/js/11_workers.js
```diff
@@ -173,27 +174,34 @@
         }
       }
 
-      if (type !== "module") {
-        throw new Error(
-          'Not yet implemented: only "module" type workers are supported',
+      const workerType = String(type);
+      if (workerType !== "classic" && workerType !== "module") {
+        throw new TypeError(
+          `The provided value '${workerType}' is not a valid enum value of type WorkerType.`,
```
crowlKats, line 180: This should be handled by a webidl enum converter

andreubotella:
While working on getting other worker tests to run, I run into web-platform-tests/wpt#29777, which is not a blocker for this PR, but it is a blocker for enabling worker tests in general until we get #11460 merged.

bartlomieju:
Removing the milestone for now, due to upstream issue

runtime/ops/web_worker/sync_fetch.rs
```diff
@@ -0,0 +1,182 @@
...
+
+impl std::fmt::Display for DomExceptionNetworkError {
+  fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+    f.pad(&self.msg)
+  }
+}
+
+impl std::error::Error for DomExceptionNetworkError {}
+
+pub fn get_dom_exception_network_error_class_name(
```
lucacasonato, line 177: Can you move this into the ext/web crate? The other Rust wrappers for dom_exception are defined there.

runtime/ops/worker_host.rs
```diff
@@ -442,6 +445,10 @@ fn op_create_worker(
   if use_deno_namespace {
     super::check_unstable(state, "Worker.deno.namespace");
   }
+  let worker_type = args.worker_type;
+  if let WebWorkerType::Classic = worker_type {
+    super::check_testing_features(state, "Worker.classic");
```
lucacasonato, line 450: I would rather that this just errors out gracefully with `"classic" workers are not supported` not supported dom exception.

runtime/ops/web_worker/sync_fetch.rs
```diff
@@ -0,0 +1,182 @@
...
+use tokio::task::JoinHandle;
+
+// TODO(andreubotella) Properly parse the MIME type
+fn mime_type_essence(mime_type: &str) -> String {
+  let essence = match mime_type.split_once(";") {
+    Some((essence, _)) => essence,
+    None => mime_type,
+  };
+  essence.trim().to_ascii_lowercase()
+}
```
lucacasonato, line 23: As this is test only code, I don't particularly care if it doesn't parse mime types correctly. Resolving this TODO is not a blocker for me.

runtime/ops/web_worker/sync_fetch.rs
```diff
@@ -0,0 +1,182 @@
...
+pub fn op_worker_sync_fetch(
+  state: &mut OpState,
+  scripts: Vec<String>,
+  mut loose_mime_checks: bool,
+) -> Result<Vec<SyncFetchScript>, AnyError> {
+  let handle = state.borrow::<WebWorkerInternalHandle>().clone();
+  assert_eq!(handle.worker_type, WebWorkerType::Classic);
+
+  // TODO(andreubotella) Make the runtime into a resource and add a new op to
+  // block on each request, so a script can run while the next loads.
```
lucacasonato, line 41: Why? Don't we want to block the thread while we are executing "importScripts"?

runtime/ops/web_worker/sync_fetch.rs
```diff
@@ -0,0 +1,182 @@
...
+
+  let runtime = tokio::runtime::Builder::new_current_thread()
+    .enable_io()
+    .enable_time()
+    .build()
+    .unwrap();
+
+  // TODO(andreubotella) It's not good to throw an exception related to blob
+  // URLs when none of the script URLs use the blob scheme.
+  // Also, in which contexts are blob URLs not supported?
```
lucacasonato, line 51: Should always be supported. I'll remove the optional.

lucacasonato (commented):
Looking very good. Lets get this landed (even if the WPTs are still blocked on some upstream issues.

runtime/ops/web_worker/sync_fetch.rs
andreubotella: Since I opened this PR, Deno gained an implementation of the `NetworkError` `DomException`, since it's used in `WebSocketStream`, so I'll just remove those wrappers and use the existing implementation.

lucacasonato: Oh perfect!

runtime/ops/worker_host.rs
andreubotella: I'll simplify `TestingFeatureChecker` as well: since it's only ever read once, the code to borrow it can simply be put inline rather than on a function like `check_unstable`.

runtime/ops/web_worker/sync_fetch.rs
andreubotella: The thread must block for the duration of the `importScripts` call, but if multiple arguments are passed, the second script can finish loading while the first is being run, rather than blocking until all scripts are loaded before the first script can run. This optimization might be important in WPT tests, since you have scripts that import other scripts that import other scripts. But we can wait and deal with this only if this becomes a problem, which might never happen.

lucacasonato: Ah, that makes sense :+1:

andreubotella:
Marking as ready for review.

lucacasonato (approved):
This LGTM now. Thanks @andreubotella. Huge step towards more worker tests. Very exciting!

PR #8807, opened 2020-12-17
Fully automate dev setup with Gitpod

Glowman554:
This commit implements a fully-automated development setup using Gitpod.io, an
online IDE for GitLab, GitHub, and Bitbucket that enables Dev-Environments-As-Code.
This makes it easy for anyone to get a ready-to-code workspace for any branch,
issue or pull request almost instantly with a single click.

PR #8587, opened 2020-12-01
Fully automate dev setup with Gitpod

james-and:
This commit implements a fully-automated development setup using Gitpod.io, an
online IDE for GitHub and GitLab that enables Dev-Environments-As-Code.
This makes it easy for anyone to get a ready-to-code workspace for any branch,
issue or pull request almost instantly with a single click.

PR #5484, opened 2020-05-16
simplify code contributions by fully automating the dev setup.

nisarhassan12:
Hi! :slightly_smiling_face: 

This Pr simplifies code contributions by fully automating the dev setup with Gitpod(An online [Eclipse Theia](https://theia-ide.org/) based IDE which is free for Open Source). With a single click, it'll launch a workspace and automatically:

- clone the `deno` repo.
- install the dependencies i.e will run `cargo build`.
- start `cargo watch -x run`.

that way anyone interested in contributing can start straight away without any friction.

You can give it a try on my fork of the repo via the following link: 

https://gitpod.io/#https://github.com/nisarhassan12/deno

This is how it looks:

image

A lot of popular OSS projects out there are already using it simplify contributors onboarding some of them are:

- https://github.com/freeCodeCamp/freeCodeCamp
- https://github.com/facebook/docusaurus
- https://github.com/mozilla/pdf.js/

You can find a brief list of them at https://contribute.dev

image

PR #11339, opened 2021-07-08
chore: add `Open in VS Code` button to README

cmoog:
This is a simple addition so I figured it didn't need an issue.

Maybe a nice follow-up would be to add a [.devcontainer](https://code.visualstudio.com/docs/remote/devcontainerjson-reference) image for use locally and with the upcoming Github Codespaces.


dsherret:
I'm not sure we should add this. I think it might confuse users who are just wanting to check out what Deno is and aren't looking to contribute. Also, it's a very wide addition 😅

[Open in Visual Studio Code](https://open.vscode.dev/denoland/deno)

lucacasonato:
Thanks, but this is a duplicate of #5484, #8587, and #8807.

cmoog:
Oof. Thanks anyway 👍

Issue #11340, opened 2021-07-09
Skip type checking by default


kitsonk:
### Discussed in https://github.com/denoland/deno/discussions/8549

<div type='discussions-op-text'>

<sup>Originally posted by **kitsonk** November 30, 2020</sup>
Instead of `--no-check` consider type checking TypeScript as an opt-in with a `--check` option.

</div>

The core team has discussed this and reviewed the feedback from the community. Based on that, we are going to move forward with the following:

- The `--check` flag will be implemented for the appropriate subcommands.  This will use the built-in `tsc` for type checking and use `tsc` for the emit (except for bundles where we will use `swc`).  We may in the future use `swc` for the emit here as well.  This is basically the current default behaviour of Deno.
- The `--no-check` flag will be deprecated and issue a warning that it is deprecated.  The behaviour will not change from its current for now.
- The default behaviour of the command line will be changed so that the code is not type checked and emitted using `swc` (this is basically the behaviour of `--no-check` now).

We _may_ do all of this for 1.12, or we may only do the first two points and delay the third for 1.13.

### Why are you doing this?

The goal of the Deno CLI is to be a fun and productive scripting language. A user should be able to _hack_ on some code and quickly see the results of their endeavour.  The more "hurdles" that are in the way, the less fun and productive the experience becomes.  We ❤️  TypeScript, we want people to use TypeScript more and more, but there are two major challenges with that:

- Type checking is relative slow and expensive.  We have tried (_almost_) everything to speed it up (see: #5432) but a relative small program still takes a second or two to type check, while just transpiling the TypeScript with `--no-check` takes under 100ms.
- Deno terminates if it encounters a type check diagnostic.  That is on us.  We should have maybe always treated them as warnings and still executed the output anyways, but we didn't.  We could of course change that now, but the problem then is the point above.  We don't feel there is a huge amount of value to users to just address this point.

While on the surface it might seem like we are "watering" down Deno, we are motivated to make it easier to use Deno.  If you are just starting out, maybe not having used TypeScript in strict mode before, or are just trying to hack a few lines of TypeScript and don't have all of your type annotations right, or you have imported a 3rd party library whose globals conflict with another 3rd party library, you have to spend all your time sorting that out to just find out if your code runs.  Again, that isn't fun for anyone and certainly isn't productive.

[...]

PR #11343, opened 2021-07-09
feat: add --check flag, deprecate --no-check flag

kitsonk:
Ref #11340

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -141,6 +141,7 @@ pub struct Flags {
   pub cache_blocklist: Vec<String>,
   pub ca_file: Option<String>,
   pub cached_only: bool,
+  pub check: bool,
   pub config_path: Option<String>,
   pub coverage_dir: Option<String>,
   pub ignore: Vec<PathBuf>,
@@ -797,6 +798,7 @@ TypeScript compiler cache: Subdirectory containing TS compiler output.",
         .conflicts_with("file")
         .help("Show files used for origin bound APIs like the Web Storage API when running a script with '--location=<HREF>'")
     )
+    .arg(check_arg())
     // TODO(lucacasonato): remove for 2.0
     .arg(no_check_arg().hidden(true))
     .arg(import_map_arg())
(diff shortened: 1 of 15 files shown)


lucacasonato (commented):
I am not in favor of landing this as is in a patch release. It silently changes the behaviour of many CI pipelines.

lucacasonato (commented):
I think all unit and integration tests should have check enabled.

caspervonb (changes requested):
As the plan is now per the referenced issue, this will silently disable type checking in our std smoke tests and cli unit tests when we make it default.

So we patch that sure but the same applies to every other CI workflow out there so we'll end up having CI matrices that go up to 1.12 and then after 1.12 it is a different CLI interface?

This will also completely break `deno test --no-run` and turn it into a no-op.

kitsonk:
To be clear this only adds `--check` and deprecates `--no-check`, as stated in the issue, there will be two PRs. The second will change the default behaviour without flags.

Also, it is planned for a minor release, not a patch release.

kitsonk:
> As the plan is now, this will disable type checking in our std smoke tests and cli unit tests when we make it default.

Not if the yet to be raised PR adds `--check`.

> Also this will completely break `deno test --no-run` and turn it into a no-op.

Not necessarily. You are assuming something that doesn't exist yet. It is a valid concern that can be addressed in the PR the doesn't exist yet.

Again, as I stated in the issue, the plan was to this PR and then figure out how disruptive the changes would be for the follow up PR before committing to when it would occur.

nayeemrmn (commented):
I think this change should be coupled with the other. It adds a flag that does nothing and deprecates a flag that isn't replaceable yet. Seems like an unreliable way of gauging disruption, if that's the purpose.

ry:
We have decided not to land this for the 1.12 release. There's clearly some controversy here, so we'll lay out the plan more carefully in an issue, discussing the reasoning behind the no-check-by-default change and a plan for how we can responsibly introduce this change.

Let's move further discussion back to #11340.

justinmchase (approved):

kitsonk:
Closing in favour of #11810

PR #11347, opened 2021-07-10
refactor(runtime): parse test permissions during test registration

caspervonb:
This moves the pledging and restoring of test permissions into a wrapper applied during registration with the rest of the wrappers like sanitizers.

### DIFF
--- a/runtime/js/40_testing.js
+++ b/runtime/js/40_testing.js
@@ -105,6 +105,29 @@ finishing test case.`;
     };
   }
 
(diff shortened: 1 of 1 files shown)


caspervonb:
Bit of back and forth if I should tag this as a refactor or fix.
Not fixing a specific bug, this just makes it slightly more consistent with how sanitisers are applied.

bartlomieju (approved):
LGTM

PR #11349, opened 2021-07-10
chore(extensions/console): avoid re-checking iterable type 

littledivy:
simplifies `value.entries()` primordial logic with `options.typeName`


lucacasonato (commented):
Could these not differ if the inspected type is an instance of for example Set? So if the inspected type is of a user type `class MySet extends Set`.

littledivy:
> Could these not differ if the inspected type is an instance of for example Set? So if the inspected type is of a user type `class MySet extends Set`.

 [`inspectObject`](https://github.com/denoland/deno/blob/355647118da23f275cbeb805f487845e77139bc1/extensions/console/02_console.js#L1068) checks if value is [`instanceof Set`](https://github.com/denoland/deno/blob/355647118da23f275cbeb805f487845e77139bc1/extensions/console/02_console.js#L1107) ... which calls [`inspectSet`](https://github.com/denoland/deno/blob/355647118da23f275cbeb805f487845e77139bc1/extensions/console/02_console.js#L779) ... which then sets `options.typeName = "Set"` and calls [`inspectIterable`](https://github.com/denoland/deno/blob/355647118da23f275cbeb805f487845e77139bc1/extensions/console/02_console.js#L340) 

That should cover your example of `class MySet extends Set`

kt3k (approved):
LGTM.

This only avoids checking `value instance Set` twice (The first once is done in inspectObject). So this shouldn't cause any behavioral change as Divy says.

PR #11353, opened 2021-07-10
fix(runtime): Align the DedicatedWorkerGlobalScope event handlers to the spec.

### LINKED ISSUE #11351: A worker global's `onmessage` handler is called with a null event target
### test.js
```js
const worker = new Worker(
  { type: "module" },
);
worker.postMessage({});
```

### worker.js
```js
self.onmessage = (evt) => {
  console.log("event handler attribute: %s", evt.target);
};

self.addEventListener("message", (evt) => {
  console.log("regular event handler: %s", evt.target);
});
```

### Result

```
event handler attribute: null
regular event handler: [object DedicatedWorkerGlobalScope]
```

However, running this code in all browsers gives:
image

-----------------

This bug, together with #11350, is causing some of the WPT test failures in #11338, since some of those WPT tests call `evt.target.postMessage()`.

andreubotella:
Fixes #11351

### DIFF
--- a/cli/tests/worker_event_handler_test.js.out
+++ b/cli/tests/worker_event_handler_test.js.out
@@ -6,11 +6,6 @@ Arguments from self.onerror: [
   0,
   0,
   Error: Some error message
-    at [WILDCARD]/cli/tests/workers/worker_event_handlers.js:9:9
-    at innerInvokeEventListeners (deno:extensions/web/02_event.js:750:9)
-    at invokeEventListeners (deno:extensions/web/02_event.js:789:5)
-    at dispatch (deno:extensions/web/02_event.js:659:11)
-    at dispatchEvent (deno:extensions/web/02_event.js:993:14)
-    at pollForMessages (deno:runtime/js/99_main.js:142:9)
+    at [WILDCARD]
 ]
 Is event canceled?: true
--- a/cli/tests/worker_event_handler_test.js.out
+++ b/cli/tests/worker_event_handler_test.js.out
@@ -1,0 +1,16 @@
+Target from self.onmessage: [object DedicatedWorkerGlobalScope]
+Target from message event listener: [object DedicatedWorkerGlobalScope]
+Arguments from self.onerror: [
+  "Some error message",
+  "",
+  0,
+  0,
+  Error: Some error message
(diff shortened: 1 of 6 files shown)


runtime/js/01_web_util.js
```diff
@@ -27,17 +27,41 @@
   }
 
   const handlerSymbol = Symbol("eventHandlers");
-  function makeWrappedHandler(handler) {
+  function makeWrappedHandler(handler, isSpecialErrorEventHandler) {
```
lucacasonato, line 30: TIL about the special error handler handling. Yucky...

Instead of the `isSpecialErrorEventHandler` flag, could you just switch on `evt instanceof ErrorEvent && evt.type === "error" && target instanceof DedicatedWorkerGlobalScope`?

andreubotella: I didn't want to do that because the `DedicatedWorkerGlobalScope` interface is not available in the main worker, but I could check for it as part of the condition.

lucacasonato: Right... never mind then. Let's keep as is.

Unrelated, but we should go and deduplicate `defineEventHandler` with the one in 02_event.js in extensions/web (available via `window.__bootstrap.event.defineEventHandler`) in a follow-up. I am about 70% through a rewrite of the 02_event.js file though, so I'll do this as part of that rewrite.

cli/tests/worker_event_handler_test.js.out
```diff
@@ -0,0 +1,16 @@
...
+  "",
+  0,
+  0,
+  Error: Some error message
+    at [WILDCARD]/cli/tests/workers/worker_event_handlers.js:9:9
+    at innerInvokeEventListeners (deno:extensions/web/02_event.js:750:9)
+    at invokeEventListeners (deno:extensions/web/02_event.js:789:5)
+    at dispatch (deno:extensions/web/02_event.js:659:11)
+    at dispatchEvent (deno:extensions/web/02_event.js:993:14)
+    at pollForMessages (deno:runtime/js/99_main.js:142:9)
```
lucacasonato, line 14:
```suggestion
    at [WILDCARD]
```

Having to change fixtures when line numbers in internal files change is annoying :-)

lucacasonato (approved):
LGTM, thanks!

PR #11357, opened 2021-07-11
refactor(cli/tools/test_runner): make test reporters stateless

caspervonb:
This collects summary information in the event collector and passes it to the reporter instead of having this embedded in each reporter which leads to a lot of duplication.


bartlomieju (approved):
LGTM

PR #11371, opened 2021-07-12
chore(wpt): Mark a WPT test as failed if it exits before completion

andreubotella:
Currently, a WPT test is considered failed if its status code is anything other than 0, regardless of whether the test suite completed running or not, and any subtests that haven't finished running are not considered to be failures.

But a test can exit with a zero status code before it has completed running, if the event loop has run out of tasks because of a bug in one of the ops, leading to false positives. This change fixes that.


lucacasonato (approved):
LGTM, thanks.

PR #11375, opened 2021-07-12
fix(cli): normalize test command errors

caspervonb:
This normalises the test command from the bottom up to return errors all the way down the chain rather than calling `process::exit` at arbritary places.


caspervonb:
Blocked on https://github.com/denoland/deno/pull/11357 and https://github.com/denoland/deno/pull/11347 (lint errors will be gone with those applied).

bartlomieju (approved):
LGTM, thanks, this is cleaner

PR #11381, opened 2021-07-13
fix(extensions/web): Aborting a FileReader object should not affect later reads

andreubotella:
Currently, calling the `abort()` method on a `FileReader` object aborts any current read operation, but it also prevents any read operation started at some later point from starting. The File API instead specifies that calling `abort()` should reset the `FileReader`'s state and result, as well as removing any queued tasks from the current operation that haven't yet run.


lucacasonato (approved):
LGTM, thanks.

PR #11383, opened 2021-07-13
 feat(cli): Set Timeout per test using --timeout flag default `500ms`

getspooky:
### Syntax:  `--timeout=<number>`
Default timeout of a test in milliseconds. Default value: `500ms`.

### Example

```sh
deno test --timeout=600 store_data_test.ts
```

### Use Cases

Let's say we want to store some data into some database provider. So in some case it could take a lot time, probably due to some network issue. So it would be helpful if we can exit the test after `X` time.
### Motivation

- Applications aren’t responding with lightning speed on user actions and while creating the `E2E` tests this fact should be considered. 
- Using these set timeout will allow you to create more stable tests and control your test flow even more precisely

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -99,6 +99,7 @@ pub enum DenoSubcommand {
     doc: bool,
     no_run: bool,
     fail_fast: Option<usize>,
+    timeout: Option<usize>,
     quiet: bool,
     allow_none: bool,
     include: Option<Vec<String>>,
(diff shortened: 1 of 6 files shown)


caspervonb (commented):
Not sure if this does what you think it does but quitting after tests have successfully been executed to completion and their combined execution time turned out to be greater than _timeout_ is **not** desirable behaviour.

Might be better if you discussed with us _before_ opening pull requests.

getspooky:
@caspervonb thanks for feedback , i will open discussion on discord .

PR #11367, opened 2021-07-12
feat(extensions/crypto): implement importKey and exportKey for raw HMAC keys

littledivy:
Introduces `SubtleCrypto.importKey()` and `SubtleCrypto.exportKey()`. Towards enabling WPT for `sign_verify`.

PR #11387, opened 2021-07-13
feat(extensions/crypto): implement verify() for HMAC

littledivy:
~~WPT blocker: #11367~~

Implements `SubtleCrypto#verify()` for HMAC keys. Enables WPT for `sign_verify`


bartlomieju (approved):
LGTM!

Issue #7844, opened 2020-10-06
setTimeout followed by clearTimeout cause the first test to fail


lideming:
File `test.ts`:
```
clearTimeout(setTimeout(() => { }, 1000));

Deno.test("noop test 1", () => { });
Deno.test("noop test 2", () => { });
Deno.test("noop test 3", () => { });
```

Run `deno test test.ts`:
```
Check file:///.../$deno$test.ts
running 3 tests
test noop test 1 ... FAILED (2ms)
test noop test 2 ... ok (1ms)
test noop test 3 ... ok (1ms)

failures:

noop test 1
AssertionError: Test case is leaking async ops.
Before:
  - dispatched: 1
  - completed: 0
After:
  - dispatched: 2
  - completed: 2

Make sure to await all promises returned from Deno APIs before
finishing test case.
    at assert (deno:cli/rt/06_util.js:33:13)
    at asyncOpSanitizer (deno:cli/rt/40_testing.js:44:7)
    at async Object.resourceSanitizer [as fn] (deno:cli/rt/40_testing.js:68:7)
    at async TestRunner.[Symbol.asyncIterator] (deno:cli/rt/40_testing.js:240:13)
    at async Object.runTests (deno:cli/rt/40_testing.js:317:22)

failures:

        noop test 1

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (5ms)

```

Version: Deno v1.4.4
OS: Windows 10 v2004 x64

lideming:
There is a workaround, await something after clearTimeout, before running the test:
```ts
Deno.test("noop test 1", () => { });
Deno.test("noop test 2", () => { });
Deno.test("noop test 3", () => { });

clearTimeout(setTimeout(() => { }, 1000));

// Test will fail if remove this line:
await new Promise(resolve => setTimeout(resolve, 0));

// Deno starts running tests...
```

magurotuna:
Seems like #7878 also has something to do with this issue

qwerasd205:
> Seems like #7878 also has something to do with this issue

Yeah, it does seem to be caused by the same thing, an unrelated async op completing during a test causing the diffs to vary.

KyleJune:
[...]

PR #11396, opened 2021-07-14
test(cli): ensure clear timeout doesn't trigger sanitizers

caspervonb:
This adds a regression test case for https://github.com/denoland/deno/issues/7844

### DIFF
--- a/cli/tests/test/clear_timeout.ts
+++ b/cli/tests/test/clear_timeout.ts
@@ -1,0 +1,5 @@
+clearTimeout(setTimeout(() => { }, 1000));
+
+Deno.test("test 1", () => { });
+Deno.test("test 2", () => { });
+Deno.test("test 3", () => { });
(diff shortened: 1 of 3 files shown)


cli/tests/test/clear_timeout.ts
```diff
@@ -0,0 +1,5 @@
+clearTimeout(setTimeout(() => {}, 1000));
```
bartlomieju, line 1: I don't think this would ever trigger sanitizers - timeout is cleared (and thus op is done) before the first test runs

caspervonb: Yeah I'm not sure why it did (internally async maybe?), but apparently it did a few months ago in #7844.

caspervonb:
Did a quick bisect on builds, this was fixed in v1.10.0.

bartlomieju (approved):
LGTM, thanks Casper for bisecting the problem

PR #11397, opened 2021-07-14
fix(lsp): support data urls in `deno.importMap` option

### LINKED ISSUE #11392: lsp: importMap doesn't suppport data urls
### test

**Describe the bug**
In deno, the `--import-map` argument allows data urls like `--import-map="data:application/json;utf8,{ \"imports\": {...}}"`. But looks like the vscode plugin only allows local files.

**To Reproduce**

1. Open .vscode/settings.json
2. Create `deno.importMap` key with a data url

**Expected behavior**

The import map should be applied.

**Screenshots**

imaxe

**Versions**

vscode: 1.58.0 deno: 1.12.0 extension: v3.7.0

satyarohith:
The second commit contains the actual implementation. Even though this PR adds support for data URLs, we can't support relative paths in data URLs as we won't have a valid base path. For example, the following won't work:

```json
{
  "deno.importMap": "data:application/json;utf8,{\"imports\": { \"a\": \"./a.ts\" }}",
}
```
Closes #11392

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -3,7 +3,6 @@
 use deno_core::error::anyhow;
 use deno_core::error::AnyError;
 use deno_core::resolve_url;
-use deno_core::resolve_url_or_path;
 use deno_core::serde_json;
 use deno_core::serde_json::json;
 use deno_core::serde_json::Value;
@@ -473,9 +472,13 @@ impl Inner {
     };
     if let Some(import_map_str) = &maybe_import_map {
       info!("Setting import map from: \"{}\"", import_map_str);
-      let import_map_url = if let Ok(url) = resolve_url_or_path(import_map_str)
+      let import_map_url = if let Ok(url) = Url::from_file_path(import_map_str)
       {
         Ok(url)
+      } else if import_map_str.starts_with("data:") {
+        Url::parse(import_map_str).map_err(|_| {
+          anyhow!("Bad data url for import map: {:?}", import_map_str)
+        })
       } else if let Some(root_uri) = &maybe_root_uri {
         let root_path = root_uri
           .to_file_path()
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -509,10 +509,6 @@ impl Inner {
           )
         })?
       };
-      // TODO(@satyarohith): if import_map_url is a data url and import_map_json contains relative
-      // paths, we will get an Invalid address error in the logs because of the base_url not
-      // being a valid path. We should somehow support relative paths in import
-      // maps that are data urls.
       let import_map =
         ImportMap::from_json(&import_map_url.to_string(), &import_map_json)?;
       self.maybe_import_map_uri = Some(import_map_url);
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -3,6 +3,7 @@
 use deno_core::error::anyhow;
 use deno_core::error::AnyError;
 use deno_core::resolve_url;
+use deno_core::resolve_url_or_path;
 use deno_core::serde_json;
 use deno_core::serde_json::json;
 use deno_core::serde_json::Value;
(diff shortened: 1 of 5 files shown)


cli/lsp/language_server.rs
```diff
@@ -427,21 +429,29 @@ impl Inner {
...
             import_map_url,
             err
           )
-        })?;
+        })?
+      };
+      // TODO(@satyarohith): if import_map_url is a data url and import_map_json contains relative
+      // paths, we will get an Invalid address error in the logs because of the base_url not
+      // being a valid path. We should somehow support relative paths in import
+      // maps that are data urls.
```
kitsonk, line 454: I would hope we addressed this on the command line @bartlomieju already when it was added to the CLI. Why don't we just use the cwd?

kitsonk (commented):
LGTM, I think we should deal with the base for the import_map now though, so want input from @bartlomieju.

cli/lsp/language_server.rs
bartlomieju: I never thought of using data URLs for import maps. Unless there's something in the spec preventing usage of data URLs (I believe there's nothing about it) I'm fine with using CWD as a base url.

kitsonk: Strangely, it appears we currently support them by accident, when we allowed remote import maps. 🤷

lucacasonato: I don't think that is what the spec suggests. Looking at https://wicg.github.io/import-maps/#sort-and-normalize-a-specifier-map it should either be the value of `--location`, or the `data:` url itself. If you try to resolve relative to the `data:` url this will fail, but the spec handles this gracefully (an error should be logged and the mapping should be set to "null").

kitsonk: Then currently, we have a problem, as we accept data URLs for `--import-map` though we don't test for it.  Of course there is no cwd in a browser and we don't use `--location` for file based imports maps.

This PR just needs to deal with adding support for `data:` URLs in line with the CLI. @satyarohith can you take out this comment here, ensure the behaviour is aligned to another CLI and open an issue about data URLs for import maps having and undefined and untested behaviour all together?

bartlomieju: So should be prohibit usage of `data:` URLs for imports maps completely?

kitsonk: No, I think @lucacasonato is saying that we shouldn't support relative specifiers in data URLs... we should log an error if we encounter one... and we should have tests for them.

bartlomieju: I'm a bit worried about the logging thing - some time ago we went away from logging warnings from import maps. Originally I implemented it strictly the the reference implementation including all the warning, but we went away from that. So the question is do we bring back other warnings/logs or just do the data URL one?

lucacasonato: What do we do now? Hard error?

bartlomieju: > What do we do now? Hard error?

No, the warnings are not displayed at all.

satyarohith: >  we went away from logging warnings from import maps

Sorry for being ignorant; what was the reason for removing the logging?

bartlomieju: Satya, to be honest my memory doesn't serve me so well, but it rings a bell that we didn't want that output interleaved in tests when testing against WPT suite.

satyarohith: > What do we do now? Hard error?

Yes, we do a hard error now.
```
➜ deno run --import-map="data:application/json;utf8,{\"imports\": { \"b\": \"./b.ts\" }}" mod.ts
Import map diagnostics:
  - Invalid address "./b.ts" for the specifier key "b".
error: Blocked by null entry for ""b""
```

kitsonk:
What are we going to do with this PR?  Do you want me to try to take it over and fix it up?

kitsonk (approved):
LGTM

Issue #11291, opened 2021-07-06
console.log in repl invalid escape sequence hangs forever


littledivy:
This is only specific to the repl.

Steps to reproduce:
```
divy@lappy:~$ deno -V
deno 1.11.5

divy@lappy:~$ deno
Deno 1.11.5
exit using ctrl+d or close()
> console.log("\u");

<hangs forever>
```

Expected behaviour:

Node.js: `Uncaught SyntaxError: Invalid Unicode escape sequence`
Firefox 89.0.2: `SyntaxError: malformed Unicode character escape sequence`

caspervonb:
Regression from the refactor and moving inspector around? pretty sure we fixed this before...

littledivy:
Looks like it doesn't accept the input in the first place when I press enter.

caspervonb:
Smaller reproduction:
```ts
String("\u")
```

Or even:
```ts
function a(x) {}
a("\u")
```

caspervonb:
Absolute minimal reproduction:

```ts
("\u")
```

caspervonb:
Remove label console/extensions.
Add label repl.

devcui:
```
devcui@macross ~ % deno -V
deno 1.11.5
devcui@macross ~ % deno
Deno 1.11.5
exit using ctrl+d or close()
> console.log("\u8bf7");
请
undefined
>
```

unicode was alright

bitoku:
If anyone isn't woking on this, I want to tackle this issue!

bartlomieju:
@bitoku please do

PR #11401, opened 2021-07-14
fix(repl): fix inconsistent integration test

bitoku:
Reason of this PR
---
I'm working on #11291 and found there is a weird integration test in `repl_tests.rs`.
In the test `pty_multiline`, some test inputs produce errors like below
```
> /{/'
parse error: Unexpected eof at 1:4
> /(/'
parse error: Unexpected eof at 1:4
> /[/'
parse error: Unterminated regexp literal at 1:0
```
I think this is a normal system test and these inputs should not be here.

I git-blamed but I couldn't figure out what the author add these inputs. There is similar test like `'('`, `'{'`, and `'['`, so I think this is a kind of typo.

What I did
---
I replaced them into the normal input.

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -14,9 +14,9 @@ fn pty_multiline() {
     master.write_all(b"'{'\n").unwrap();
     master.write_all(b"'('\n").unwrap();
     master.write_all(b"'['\n").unwrap();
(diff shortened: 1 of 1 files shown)


caspervonb:
https://github.com/denoland/deno/pull/8037

dsherret (approved):
LGTM. Thanks @bitoku!

Issue #11345, opened 2021-07-09
lsp: cache swc ASTs


dsherret:
I'm breaking this out of #11032

> Also, caching the swc parse of the module in the document info would be a good thing I think. That might take a bit more structuring of code. It is used in dependency analysis of the document as well as the test code lens and also use by the linting (though I don't know how easy it would be to send it to deno_lint without a bigger refactor of one or the other).

Steps:

- [x] Remove use of `SourceMap` and `SourceFile` in the code because they don't implement `Send` without turning on `swc_common`'s `concurrent` feature, which is not something I want to do because then all the code will use locks internally (which we don't need... slower), open us up to a new class of possible bugs, and it's a global change.
   - [x] https://github.com/swc-project/swc/pull/1908 - Remove SourceMap dependency in swc dep-graph and integrate into Deno code
   - [x] Add replacement functionality for `SourceMap/SourceFile`'s `BytePos` -> Line and column that exists on `ParsedModule`.
   - [ ] Do as many parts of #11111 as makes sense for completing this task (probably using what's implemented on `ParsedModule` in the step above might make most of this not required to complete this task)
   - [x] Refactor code for transpiling and bundling to still use `SourceMap` & `SourceFile` as it is required by swc, but change the rest of the code not to use that.
- [x] Replace `SingleThreadedComments` implementation of `Comments` with a version that implements `Send`.
- [x] Store the `ParsedModule` in the document cache and use it throughout the code.

PR #11402, opened 2021-07-14
chore: upgrade to swc 0.46

dsherret:
This upgrades swc to 0.46. There was a breaking change I did in swc for getting rid of the `source_map` dependency for #11345.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -43,8 +43,8 @@ winres = "0.1.11"
 
 [dependencies]
 deno_core = { version = "0.93.0", path = "../core" }
-deno_doc = { git = "https://github.com/denoland/deno_doc" }
-deno_lint = { git = "https://github.com/denoland/deno_lint" }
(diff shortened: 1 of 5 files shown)


cli/Cargo.toml
```diff
@@ -43,8 +43,8 @@ winres = "0.1.11"
 
 [dependencies]
 deno_core = { version = "0.93.0", path = "../core" }
-deno_doc = "0.8.0"
-deno_lint = "0.10.0"
+deno_doc = { git = "https://github.com/denoland/deno_doc" }
+deno_lint = { git = "https://github.com/denoland/deno_lint" }
```
dsherret, line 47: This is obviously temporary. We'll need to do a release of these. I'm just opening this as a draft PR then we can convert it to a real PR once we do a release (whenever that happens).

FYI @bartlomieju

kitsonk (approved):
LGTM

PR #11405, opened 2021-07-14
refactor(cli/unix_utils): raise fd limit in O(1) time

Nugine:
Binary search is unnecessary. We can raise fd limit directly.
Note that there is another strict limit (`kern.maxfilesperproc`) on macOS and other BSDs. 
The tool in `rlimit` crate detects this limit and tries to increase `RLIMIT_NOFILE` soft limit as high as possible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -588,6 +588,7 @@ dependencies = [
  "rand 0.8.4",
  "regex",
  "ring",
+ "rlimit",
  "rustyline",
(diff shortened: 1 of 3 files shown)


bnoordhuis (commented):
I'm 🤷 on this change.

I don't think performance is an issue (converges in just a few system calls) and it's 1/4th the size of the corresponding code in that `rlimit` crate.

PR #11410, opened 2021-07-15
fix(extensions/net): http.Server doesn't expose randomly selected port

getspooky:
If you specify port 0, as serve({ port: 0 }) the library works correctly and automatically selects a port. However, the port is accessible to the process cannot advertise where it is listening on.
See https://github.com/denoland/deno_std/issues/1036

### DIFF
--- a/extensions/net/01_net.js
+++ b/extensions/net/01_net.js
@@ -206,7 +206,9 @@
       hostname: typeof hostname === "undefined" ? "0.0.0.0" : hostname,
       ...options,
     });
-
+    if (options.port === 0) {
+      console.log("Listining on " + res.localAddr.port);
+    }
     return new Listener(res.rid, res.localAddr);
   }
 


AaronO:
@getspooky I don't think logging in the runtime code is a good idea, low-level runtime libs should strive to be "silent".

You can actually access & thus print the random port:
```ts
import { serve } from "https://deno.land/std@0.101.0/http/server.ts";
const server = serve({ port: 0 });
const serverAddr = server.listener.addr as Deno.NetAddr;
console.log(`http://localhost:${serverAddr.port}`);
  req.respond({ body: "Hello World\n" });
}
```

That's a bit quirky since [Deno.Addr = Deno.NetAddr | Deno.UnixAddr](https://doc.deno.land/builtin/stable#Deno.Addr), but you could add a convenience getter on `Server` in `std` to get the port.

wperron:
I agree with @AaronO, this should be something done in std -- also make sure to read https://github.com/denoland/deno_std/discussions/1034 to make sure you align your work with ongoing efforts in std/http.

getspooky:
@wperron , @AaronO  Thanks , What do you think about adding port getter in Server class , Something like that
```
  /**
   * Expose server port.
   */
  get port() {
    return (this.listener.addr as Deno.NetAddr).port;
  }
```

getspooky:
As @wperron suggest we will fix that after https://github.com/denoland/deno_std/discussions/1034 are done

Issue #11414, opened 2021-07-15
Add help text description for the `deno test --jobs` flag


dsherret:
The `--jobs` flag for `deno test` does not say what it does in the help text.

```
> deno help test
deno-test
Run tests using Deno's built-in test runner.

...omitted...

OPTIONS:
    -A, --allow-all                    Allow all permissions
        ...omitted...                  ...omitted...
        --inspect-brk=<HOST:PORT>      Activate inspector on host:port and break at start of user script
    -j, --jobs <jobs>
        --location <HREF>              Value of 'globalThis.location' used by some web APIs
        ...etc...                      ...etc...
```

caspervonb:
Ooops... on it.

PR #11415, opened 2021-07-15
fix(cli): add help description for test --jobs option

caspervonb:
Closes https://github.com/denoland/deno/issues/11414

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1059,6 +1059,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("jobs")
         .short("j")
         .long("jobs")
+        .help("Number of parallel jobs, defaults to # of CPUs when no value is provided")
         .min_values(0)
         .max_values(1)
         .takes_value(true)
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1059,7 +1059,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("jobs")
         .short("j")
         .long("jobs")
-        .help("Number of parallel jobs, defaults to # of CPUs when no value is provided")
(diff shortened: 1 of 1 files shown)


cli/flags.rs
```diff
@@ -1059,6 +1059,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("jobs")
         .short("j")
         .long("jobs")
+        .help("Number of parallel jobs, defaults to # of CPUs when no value is provided")
```
dsherret, line 1062: Is there a way to describe this without using "jobs" in order to explain what "jobs" means?

lucacasonato:
Also mention that it defaults to 0 if the flag isn't provided.

cli/flags.rs
caspervonb: Hmm for self explanatory, I guess worker would be a good approximation.

```suggestion
        .help("Number of parallel workers, defaults to # of CPUs when no value is provided")
```

dsherret: Yeah, I think that's better. It might be nice to somehow explain what the workers do as users might not know. I'm not sure what it should say myself as I'm not familiar with this feature.

caspervonb: >  Ex. they're parallel across modules, right?

No, modules *can* run in parallel but they're still isolated to their isolate.

cli/flags.rs
```diff
@@ -1059,6 +1059,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("jobs")
         .short("j")
         .long("jobs")
+        .help("Number of parallel workers, defaults to # of CPUs when no value is provided")
```
caspervonb, line 1062:
```suggestion
        .help("Number of parallel workers, defaults to # of CPUs when no value is provided. Defaults to 1 when the option is not present.")
```

ry (approved):
Thanks Casper - LGTM

PR #11417, opened 2021-07-16
fix(workers): Silently ignore non-existent worker IDs

### LINKED ISSUE #11416: Worker panics if terminated incorrectly 'No worker handle found', runtime\ops\worker_host.rs:522:6
I guess my termination code is incorrect, but Deno still should not panic:

```typescript
import { assertEquals } from "https://deno.land/std@0.100.0/testing/asserts.ts";

const fauxself = self as Worker & typeof self;
fauxself.onmessage = function (e) {
    fauxself.postMessage("foo");
    fauxself.close();
};


function runWorkerOnce(worker: Worker, initMessage: any) {
    return new Promise((resolve, _reject) => {
        worker.onmessage = (m) => {
            // Following causes a panic:
            // thread '<unnamed>' panicked at 'No worker handle found', runtime\ops\worker_host.rs:522:6
            worker.terminate();
            resolve(m);
        };
        worker.postMessage(initMessage);
    });
}

if (typeof Deno !== "undefined") {
    Deno.test({
        name: "test something",
        async fn() {
            const w = new Worker(import.meta.url, {
                type: "module",
            });

            const res: any = await runWorkerOnce(w, "");
            assertEquals(res.data, "foo");
        },
    });
}
```

Try runnin the `Deno.test` case, it gives me `thread '<unnamed>' panicked at 'No worker handle found', runtime\ops\worker_host.rs:522:6`

## Workaround

```
       setTimeout(() => {
                worker.terminate();
            }, 0);
```

andreubotella:
The `op_host_terminate_worker` and `op_host_post_message` ops used to panic when called with a non-existent worker ID, since it would indicate an error in the JS side. However, #11416 shows that race conditions between a worker's `close()` function and the corresponding `Worker.prototype.terminate()` can lead to the JS side performing ops on the worker before being notified that the worker has terminated. As such, this change silently ignores non-existent worker IDs in those ops. 

Fixes #11416.

### DIFF
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -1160,6 +1160,11 @@ itest!(worker_event_handler_test {
   output: "worker_event_handler_test.js.out",
 });
 
+itest!(worker_close_race {
+  args: "run --quiet --reload --allow-read worker_close_race.js",
+  output: "worker_close_race.js.out",
+});
+
 #[test]
 fn no_validate_asm() {
   let output = util::deno_cmd()
--- a/cli/tests/worker_close_race.js
+++ b/cli/tests/worker_close_race.js
@@ -1,0 +1,14 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+// https://github.com/denoland/deno/issues/11416
+// Test for a race condition between a worker's `close()` and the main thread's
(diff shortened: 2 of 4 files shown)


bartlomieju (approved):
LGTM, good fix, thanks @andreubotella

PR #11426, opened 2021-07-16
fix(repl): output error without hanging when input is invalid

### LINKED ISSUE #11291: console.log in repl invalid escape sequence hangs forever

bitoku:
Reason of this PR
---

Close #11291 

As discussed in #11291, when the input is invalid, repl hangs forever.
This PR makes repl output error instead of hanging when the input is invalid.

What I did
---
Actually, repl looks hanging, but it just continues to read the next input that never end.
The reason is that repl thinks the input might be the middle of the template though the input is invalid.
However it continues to read when obviously it is not in a template.

I changed the condition that determines whether it continues to read or not.
When the error is `SyntaxError::UnterminatedTpl`, it continues to read, otherwise raise error.

When raising error, in the same way as unpaired braces, it marked the input as valid with no info and V8 raise the error instead.

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -73,6 +73,23 @@ fn pty_bad_input() {
 
 #[cfg(unix)]
 #[test]
+fn pty_syntax_error_input() {
+  use std::io::{Read, Write};
+  run_pty_test(|master| {
+    master.write_all(b"('\\u')\n").unwrap();
+    master.write_all(b"('\n").unwrap();
+    master.write_all(b"close();\n").unwrap();
+
+    let mut output = String::new();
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM - thanks @bitoku !

Issue #11432, opened 2021-07-18
The test subcommand does not enter watch mode


caspervonb:
Looks like the test subcommand does not enter watch mode at all.

```bash
deno test --watch
No matching test modules found
```

caspervonb:
According to git blame I've yanked the parse line.

PR #11433, opened 2021-07-18
fix(cli): re-enable --watch flag for test

caspervonb:
Fixes https://github.com/denoland/deno/issues/11432


kt3k (approved):
Confirmed the issue exists and this fixes it. Thanks @caspervonb !

PR #11434, opened 2021-07-18
test(cli): add test case for 'deno test --watch'

kt3k:
This PR adds a test case for checking the fix #11433, prevents #11432 happening again in the future.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -3475,6 +3475,29 @@ mod tests {
   }
 
   #[test]
+  fn test_watch() {
+    let r = flags_from_vec(svec!["deno", "test", "--watch"]);
+    assert_eq!(
+      r.unwrap(),
+      Flags {
+        subcommand: DenoSubcommand::Test {
+          no_run: false,
+          doc: false,
+          fail_fast: None,
+          filter: None,
+          allow_none: false,
+          quiet: false,
+          shuffle: None,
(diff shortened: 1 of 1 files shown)


getspooky:
I think we should add test for single file watch ?

cli/flags.rs
```diff
@@ -3474,6 +3474,29 @@ mod tests {
     );
   }
```
getspooky, line 3476: I think to have more check , we should add check for single test watch 

```suggestion
  #[test]
  fn test_single_file_watch() {
    let r = flags_from_vec(svec!["deno", "test", "--watch", "test/ignore.ts"]);
    assert_eq!(
      r.unwrap(),
      Flags {
        subcommand: DenoSubcommand::Test {
          no_run: false,
          doc: false,
          fail_fast: None,
          filter: None,
          allow_none: false,
          quiet: false,
          shuffle: None,
          include: Some(svec!["test/ignore.ts"]),
          concurrent_jobs: 1,
        },
        watch: true,
        ..Flags::default()
      }
    );
  }
```

kt3k: #11432 was caused by skipping the check of `watch` flag. The current change is enough for checking that in my view.

ry (approved):
LGTM - ideally tests should be landed with the fix

PR #11438, opened 2021-07-18
feat(cli/tools/test): imply media type from doc attribute

caspervonb:
This augments the doc test feature to use the first attribute in markdown blocks as the media type and extension for the extracted example.


bartlomieju (approved):
LGTM

Issue #11441, opened 2021-07-18
`Array.at` and friends not available for typechecking


lucacasonato:
```ts
console.log([1,2,3,4].at(-1))
```

Runs fine with `--no-check`, but during type checking:

```
error: TS2339 [ERROR]: Property 'at' does not exist on type 'number[]'.
[1,2,3,4].at(-1)
          ~~
```

The typings update has not landed in TSC upstream yet: https://github.com/microsoft/TypeScript/pull/40697.

We should try get this fixed for 1.12.1.

PR #11443, opened 2021-07-19
fix(tsc): add .at() types manually to tsc, update to tsc v4.3.5

kitsonk:
Fixes: #11441

This manually adds the types for `.at()` types to arrays and strings.  It is a bit "hacky", but essentially the same way we add `"dom.asynciterable"` to TypeScript.

Something we should revert once in a released version of TypeScript (but will likely just overwrite when it is finally included).

Added a simply js_unit test which validates that it should pass type checking.

Also updated to the latest release version of TypeScript (4.3.2 -> 4.3.5).


kt3k (approved):
LGTM

PR #11448, opened 2021-07-19
feat(runtime): implement navigator.hardwareConcurrency

### LINKED ISSUE #9584: Expose window.navigator.hardwareConcurrency
Hello 👋 

MDN: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorConcurrentHardware/hardwareConcurrency

Use Case: Determine how many CPU cores are available to create that many workers.

Note: This is already doable in `Deno. systemCpuInfo().cores` but this is unstable.

Thank You!

littledivy:
Removes `Deno.systemCpuInfo()` and closes #9584


ry (commented):
There is another reference in cli/dts/lib.deno.unstable.d.ts

lucacasonato (commented):
Can you pass in the cpu count during bootstrapping? (`MainWorker::bootstrap` / `WebWorker::bootstrap`)

tools/wpt/expectation.json
```diff
@@ -3150,5 +3150,8 @@
     "eventhandlers.any.html?wpt_flags=h2": false,
     "eventhandlers.any.html?wss": false,
     "referrer.any.html": true
+  },
+  "workers" : {
+    "WorkerNavigator-hardware-concurrency.any": true
```
lucacasonato, line 3155: This test file doesn't exist.

lucacasonato (commented):
Needs a unit test.

tools/wpt/expectation.json
andreubotella: And `workers/WorkerNavigator-hardware-concurrency.any.html` will fail since #11338 isn't merged yet. And there doesn't seem to be any tests to test `hardwareConcurrenty` on the window's `Navigator` interface.

littledivy: Hmm, ok...I'll add a unit test for it

bartlomieju (approved):
LGTM, thanks Divy

PR #11459, opened 2021-07-20
feat(lsp): add workspace config to status page

kitsonk:
This add a representation of what the language server thinks are the current settings to the language server status page.

We have had several issues that have been difficult to diagnose because we don't know how workspace and global settings in vscode have had an impact on the lsp and this provides exactly what the LSP thinks things are.


satyarohith (approved):
LGTM.

PR #11462, opened 2021-07-20
build: publish canary version file in seperate job

lucacasonato:


piscisaureus (approved):
The purpose of this is to make sure canary-latest.txt isn't uploaded before all the builds are in?
If so, LGTM.

lucacasonato:
Yup, that's exactly it.

PR #11468, opened 2021-07-20
ref(cli): Follow cli option naming conventions

getspooky:
Follow cli option naming convetion: 
- Rename `: --fail-fast=<N>` to `--fail-fast=<Number>` 
- Rename: option description `Stop after N errors. Defaults to stopping after first failure` to `Stop after failures errors. Defaults to stopping after first failure`

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1001,12 +1001,12 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
       Arg::with_name("fail-fast")
         .long("fail-fast")
         .alias("failfast")
-        .help("Stop after N errors. Defaults to stopping after first failure.")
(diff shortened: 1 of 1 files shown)


caspervonb:
N is preferable imo.

getspooky:
Yes , but we should follow the same naming convention for all options like : 
```
--seed <NUMBER>                
     Seed Math.random()
--shuffle=<NUMBER>             
   (UNSTABLE): Shuffle the order in which the tests are run
```

lucacasonato:
The difference is that N is a concrete count, while `NUMBER` is a specific integer where the exact value is not important as it is an opaque identifier.

Thanks for the PR, but let's keep the status quo.

PR #11471, opened 2021-07-20
fix(cli): rebuild when environment variables change

piscisaureus:


bartlomieju (approved):

lucacasonato (approved):

lucacasonato:
We should do the same for `GIT_COMMIT_HASH` actually.

piscisaureus:
> We should do the same for `GIT_COMMIT_HASH` actually.

Ok, done.

PR #11478, opened 2021-07-21
fix(cli): info now displays type reference deps

### LINKED ISSUE #11476: `deno info` does not display type dependencies correctly
If you have a dependency in a module graph that contains a triple-slash reference to the types or was served up with an `X-TypeScript-Types` headers, the information about the type dependency is currently missing from the graph.

I'm working on a fix for this, it is just a tracking issue.

kitsonk:
Fixes #11476


cli/info.rs
```diff
@@ -411,11 +470,16 @@ https://deno.land/x/a.ts (123B)
...
+              {
+                "specifier": "./d.js",
+                "code": "https://deno.land/x/d.js"
+              }
+            ],
+            "size":789,
             "mediaType": "JavaScript",
             "local": "/cache/deps/https/deno.land/x/c.js",
-            "checksum": "9876abcef"
+            "checksum":"9876abcef"
```
lucacasonato, line 482: This seems unintentional. Same 2 lines up.

schemas/module-graph.json
```diff
@@ -0,0 +1,125 @@
+{
```
lucacasonato, line 1: I dislike that this adds a new top level directory. Can this go into cli/schemas/?

lucacasonato (approved):
LGTM with some minor comments

Issue #11482, opened 2021-07-21
Sending fatal alert BadCertificate: invalid certificate: UnknownIssuer


justinmchase:
The issuer is my company. I am behind a corporate firewall and so is the destination server. I have the same CA cert the server is using installed locally.

I am also on macOS Catalina 10.15.7

I'm using `fetch(url)` and its giving me this error:
```
Sending fatal alert BadCertificate
error: Uncaught (in promise) TypeError: error sending request for url (https://foo.bar.com/health): error trying to connect: invalid certificate: UnknownIssuer
    at deno:core/core.js:86:46
    at unwrapOpResult (deno:core/core.js:106:13)
    at async mainFetch (deno:extensions/fetch/26_fetch.js:229:14)
```

My code is basically:

```sh
deno run -A --cert /etc/ssl/cert.pem main.ts
```
```ts
const res = await fetch('https://foo.bar.com/health', {
  method: 'GET',
  cache: 'no-cache',
  keepalive: false,
  signal: controller.signal
})
```
Querying the same url using curl with the same cert from the terminal gives me a success result (domains and cert details altered):
```sh
$  curl -v https://foo.bar.com/health
*   Trying 10.202.14.185...
* TCP_NODELAY set
* Connected to foo.bar.com (10.202.14.185) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
[...]

PR #11324, opened 2021-07-07
feat: Add --unsafely-treat-insecure-origin-as-secure flag to disable SSL verification

TheAifam5:
This PR corresponds to the issue #1371 which implement additional flag to the cli, tools, core, extensions and runtime allowing a user to define a domain or a list of domains which should be excluded from certificate validation while connecting or performing the HTTPS requests or disable validation completely.

The name `noCheckCertificate` and derives I found somewhere in issues of Deno repository as an proposal for the naming and decided to stick with it.

**I will be out for few weeks and I can not continue the work on this code, but I will be able to accept PR in my repository and will be open for discussion.**

The first commit is an actual feature implementation, the second is only about fixing issues reported by `./tools/format.js` and `./tools/lint.js` and some comment changes - all of the issues reported by those tools are fixed, except the one which I am not able to fix right now.

Those 3 tests are failing for unknown reason:
```
failures:
    integration::js_unit_tests
    integration::lsp::lsp_large_doc_changes
    integration::timeout_clear
```

Closes https://github.com/denoland/deno/issues/5931

TODOs:

- [x] Tests
- [ ] Domain name validation (and IP too?)
- [x] Cleanup

Thanks for hints @bartlomieju :)

PR #11491, opened 2021-07-22
fix(tls): Optionally support loading native certs

justinmchase:
# Overview
This changes Deno to support optionally loading certificates from the users local certificate store. This will allow them to successfully connect via tls with corporate and self signed certs provided they have them installed in their keystore. It also allows them to deal with revoked certs by simply updating their keystore without having to upgrade Deno.

Potentially in the future Deno could consider modifying its default behavior to load from the certificate store but for now it is only optionally enabled by specifying it in the new flag described below.

This change also adds a new flag `DENO_TLS_CA_STORE` which can have the values `system,mozilla` and will load them in the order they are specified and defaults to `mozilla`. `system` controls loading certs from the users certificate store and `mozilla` controls loading the bundled mozilla certs.

## Issues
Fixes https://github.com/denoland/deno/issues/11482
Related To https://github.com/denoland/deno/issues/10312

# Testing
I have access through VPN to an internal website which is using a certificate issued by my companies internal CA. With latest Deno I am unable to access that site because the only certificate I have is installed in my system keychain by my IT department.

The app code:
```ts
const { ok, status } = await fetch('http://example-app.acme.com')
console.log(`${ok ? 'OK' : 'FAIL'} ${status}`)
```

To see this error I am able to run:
```sh
$ deno run --unstable -A main.ts
Check file:///Users/jchase24/code/example/main.ts
Sending fatal alert BadCertificate
error: Uncaught (in promise) TypeError: error sending request for url (https://example-app.acme.com/#undefined): error trying to connect: invalid certificate: UnknownIssuer
    at deno:core/core.js:86:46
    at async mainFetch (deno:extensions/fetch/26_fetch.js:229:14)
```

Now with these changes I've run
```sh
$ cd ~/code/deno
$ cargo build
```
```sh
$ cd ~/code/example
$ DENO_TLS_CA_STORE=system ../deno/target/debug/deno run -A main.ts
Check file:///Users/jchase24/code/example/main.ts
OK 200
```

Using the flag to remove the certificate store, results in the error again (as expected)
```sh
DENO_TLS_CA_STORE=mozilla ../deno/target/debug/deno run -A main.ts
Sending fatal alert BadCertificate
error: Uncaught (in promise) TypeError: error sending request for url (https://example-app.acme.com/#undefined): error trying to connect: invalid certificate: UnknownIssuer
    at async mainFetch (deno:extensions/fetch/26_fetch.js:265:14)
```


extensions/fetch/Cargo.toml
```diff
@@ -13,13 +13,14 @@ repository = "https://github.com/denoland/deno"
...
 
+
 [dependencies]
 bytes = "1.0.1"
 data-url = "0.1.0"
 deno_core = { version = "0.94.0", path = "../../core" }
 deno_web = { version = "0.43.0", path = "../web" }
 http = "0.2.4"
-reqwest = { version = "0.11.4", default-features = false, features = ["rustls-tls", "stream", "gzip", "brotli"] }
+reqwest = { version = "0.11.4", default-features = false, features = ["rustls-tls", "native-tls", "stream", "gzip", "brotli"] }
```
bartlomieju, line 23: Unfortunately this is a no-go - using `native-tls` feature in `reqwest` will link to OpenSSL, which we explicitly gave up on doing last year.

justinmchase: So I'm looking at using `load_native_certs()` and then trying to pass those to the ClientBuilder instead of using its implementation. I will remove the `native-tls` feature.

I'm hitting a small roadblock, `load_native_certs()` returns a set of `OwnedTrustAnchor`'s and the `ClientBuilder` doesn't expose a way to access the root_store or add TrustAnchor's, unless someone better rust-fu can understand how to convert a TrustAnchor into a Certificate, it seems like it will require a small change to reqwest to enable it.

Here it is adding the same webpki certs in the same way deno is:
https://github.com/seanmonstar/reqwest/blob/bccefe7486f32b9289a3d7fe5790e74cd68a879a/src/async_impl/client.rs#L311

So I think probably updating the ClientBuilder api to have something like `add_trust_anchor` function would enable us to do this:

```ts
load_native_certs()
  .expect("could not load platform certs")
  .roots
  .map(|root: OwnedTrustAnchor| builder.add_trust_anchor(root));
```

bartlomieju: > unless someone better rust-fu can understand how to convert a TrustAnchor into a Certificate, it seems like it will require a small change to reqwest to enable it.

I don't think that's possible, I guess a change to reqwest is needed after all

lucacasonato: You can use `ClientBuilder::use_preconfigured_tls`

justinmchase: Yup that worked. Though it required me to copy the implementation for `CLIENT_SESSION_MEMORY_CACHE` which appeaers to only exist in the `net` extension right now. Can I reference the net module here and access that static variable so its shared or should it be bumped up to a different module where they can both access it? Or just duplicating it in each module is fine?

extensions/fetch/lib.rs
```diff
@@ -38,24 +41,54 @@ use reqwest::Body;
...
+    let mut sessions = self.0.lock();
+    // TODO(bnoordhuis) Evict sessions LRU-style instead of arbitrarily.
+    while sessions.len() >= 1024 {
+      let key = sessions.keys().next().unwrap().clone();
+      sessions.remove(&key);
+    }
+    sessions.insert(key, value);
+    true
+  }
+}
```
justinmchase, line 90: This is a duplicate of the cache from the `net` extension:
https://github.com/denoland/deno/blob/main/extensions/net/ops_tls.rs#L83

Also the websocket extension isn't using a cache at all, now that I look at it:
https://github.com/denoland/deno/blob/main/extensions/websocket/lib.rs#L225

Can I push this cache struct up somehwere higher in deno, so they can all reference it?

extensions/fetch/lib.rs
```diff
@@ -568,12 +601,19 @@ pub fn create_http_client(
   ca_data: Option<Vec<u8>>,
   proxy: Option<Proxy>,
 ) -> Result<Client, AnyError> {
+  let mut tls_config = ClientConfig::new();
+  tls_config.set_persistence(CLIENT_SESSION_MEMORY_CACHE.clone());
+  tls_config.root_store = load_native_certs()
+    .expect("could not load platform certs");
+  tls_config.root_store
+    .add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
```
justinmchase, line 609: I can confirm that this successfully loads my companies CA stored in my system keychain and lets me access internal domains.

It should also pull in the default webpki certs, so if those certs go stale over time a user would be able to either upgrade their version of Deno or install the missing certs into their keychain.

This should cover a broad range of scenarios by default though.

extensions/net/Cargo.toml
```diff
@@ -19,6 +19,7 @@ deno_core = { version = "0.94.0", path = "../../core" }
 log = "0.4.14"
 lazy_static = "1.4.0"
 rustls = "0.19.0"
+rustls-native-certs = "0.5.0"
```
justinmchase, line 22: Here is the link so you can evaluate this package more closely:
https://github.com/rustls/rustls-native-certs

I'm not affiliated with it at all but I gave it a look over and found it to be good.

It seems to:

* have a compatible open source license
* be under active development
* be very simple, it does only the one thing we would want it to do
* actually work

extensions/fetch/lib.rs
justinmchase: I just duplicated it for now, there are two other spots where it should also either link to a shared function or duplicate this code again.

I'm not quite familiar with the code base enough to know what the right way to elevate the structs into a place where multiple extensions can access it. Perhaps a follow up PR? I don't believe it would be a regression to push this off as they are currently not using the cache either.

justinmchase:
## Outstanding Issues
1. There should probably be one function in all of Deno for initializing the ClientBuilder and tls config
2. There should be one cache for tls persistence
3. This change needs some unit tests, but I'm not totally sure how to do it. Some advice on where / what kinds of tests should be added would be appreciated.

cli/flags.rs
```diff
@@ -263,6 +264,8 @@ static ENV_VARIABLES_HELP: &str = r#"ENVIRONMENT VARIABLES:
                          hostnames to use when fetching remote modules from
                          private repositories
                          (e.g. "abcde12345@deno.land;54321edcba@github.com")
+    DENO_TLS_CA_STORE    Comma-seperated list of certificate stores
+                         (defaults to "system,mozilla")
```
lucacasonato, line 268: Defaulting to both should be a seperate change, and is definitely not in scope for a `fix:`. Changing the defaults would have to wait for a minor (possibly a major).

lucacasonato (commented):
There is now a lot of duplicated logic. I would suggest the following:

On `ProgramState` creation, create a single `rustls::RootCertStore` should be created containing certificates based on the user requested CA stores, and the certificate provided by `--cert`. This `rustls::RootCertStore` should then be injected into `create_http_client` (and `deno_fetch`), and all other places where we use TLS (`deno_net` etc).

This would also allow you to remove all the places we inject `ca_data`, because that would already be part of the `rustls::RootCertStore`.

cli/http_util.rs
```diff
@@ -16,20 +18,46 @@ use deno_runtime::deno_fetch::reqwest::redirect::Policy;
...
+    Some(stores) => {
+      if stores.contains(&"system".to_string()) {
+        tls_config.root_store =
+          load_native_certs().expect("could not load platform certs");
+      }
+      if stores.contains(&"mozilla".to_string()) {
+        tls_config
+          .root_store
+          .add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+      }
```
lucacasonato, line 44: With this setup system certs would always have priority over the built-in ones. This is not good. Priority should be determined based on the order of the arguments in the env var.

cli/flags.rs
justinmchase: ok

cli/http_util.rs
justinmchase: Ok I did it this way because the `load_native_certs()` returns a cert store and needed to be set, which implies it would need to be set before you can add more... but now that you mention it because of some other changes to the way its being loaded I can probably now just use `.add_server_trust_anchors(load_native_certs().roots)`, then it would allow them to be loaded in any order.

extensions/fetch/lib.rs
justinmchase: I moved this into a shared location in the `deno_core` module so the code isn't duplicated and all clients share the same cache.

cli/flags.rs
justinmchase: I changed the default to only be "mozilla". Loading from the keystore is non-default.

cli/http_util.rs
justinmchase: I updated it to load the two stores based on the order they appear in the vector

cli/standalone.rs
```diff
@@ -227,13 +232,24 @@ pub async fn run(
...
+    .root_store
+    .clone()
+    .or_else(|| Some(create_default_root_store()))
+    .unwrap();
+
+  if let Some(cert) = metadata.ca_data {
+    let reader = &mut BufReader::new(Cursor::new(cert));
+    root_store.add_pem_file(reader).unwrap();
+  }
+
```
justinmchase, line 245: This section threw me for a loop for a bit but I think its accurate now. I had to understand how the ProgramState, Metadata and Flags all related and how its due to the standalone feature to get this.

The goal is to consolidate it all into a single RootCertStore but the RootCertStore is not serializable, thus the ca_stores and the ca_file are flags but when building standalone the ca_file gets serialized into ca_data, without the flag, so when launching the standalone it will create the base stores then add the stored cert from the ca_data.

justinmchase:
Ok this was refactored and updated to:

1. Consolidate duplicate code up into deno_core
2. Replace passing ca_data,ca_file with cert_store. 
3. Default behavior is to only load the mozilla certs as it does today, the system certs are only loaded if specified via the flag.
3. Some places in code such as `op_*` functions can have arguments which pass certs directly in, that behavior should not have been changed and they get added to a clone of the base store per function call.
4. Some modules such as rustls were bumped up to deno_core and are referenced directly there rather than having each other module have a duplicate dependency.

cli/program_state.rs
```diff
@@ -65,11 +68,50 @@ impl ProgramState {
...
+    let mut root_store = RootCertStore::empty();
+    let ca_stores: Vec<String> = flags
+      .ca_stores
+      .clone()
+      .or_else(|| {
+        let env_ca_store = env::var("DENO_TLS_CA_STORE");
+        match env_ca_store {
+          Ok(arg) => Some(arg.split(',').map(|s| s.to_string()).collect()),
+          _ => None,
+        }
```
bnoordhuis, line 84: Minor thing but you can probably shorten this to:
```suggestion
        let env_ca_store = env::var("DENO_TLS_CA_STORE").ok()?;
        Some(env_ca_store.split(',').map(|s| s.to_string()).collect())
```
Aside: I'd trim whitespace because people are bound to write `"system, mozilla"` and then get confused when it doesn't work.

cli/program_state.rs
```diff
@@ -65,11 +68,50 @@ impl ProgramState {
...
+      })
+      .or_else(|| Some(default_ca_stores.clone()))
+      .unwrap()
+      .iter()
+      .filter(|store| supported_ca_stores.contains(&store))
```
bnoordhuis, line 89: For that matter: unsupported should probably be a fatal error?

core/Cargo.toml
```diff
@@ -22,11 +22,16 @@ libc = "0.2.98"
 log = "0.4.14"
 parking_lot = "0.11.1"
 pin-project = "1.0.7"
+reqwest = { version = "0.11.4", default-features = false, features = ["rustls-tls", "stream", "gzip", "brotli"] }
+rustls = "0.19.0"
+rustls-native-certs = "0.5.0"
```
bnoordhuis, line 27: I want to double-check: are other maintainers on board with taking on these dependencies?

bnoordhuis (commented):
Drive-by review.

core/Cargo.toml
bartlomieju: We already have those dependencies, but it's definitely not ok to add anything related to HTTP and/or TLS to `core/`. These changes should only happen in `extensions/`, `runtime/` and `cli/` directories.

justinmchase: The rustls-native-certs is a new one, one of the comments in this PR adds some information for reviewers to review the dependency. The other two were already added.

This stuff was moved into core because a previous review stated that there was a lot of duplicate code (there was) and so to consolidate it I made one function and had the multiple places in code call those new shared functions.

I'll look to move it to runtime, I have no opinion one way or another it just seemed like core was a library which each of the callers was already referencing.

cli/program_state.rs
justinmchase: Ok, this seemed more forgiving but its probably wise to throw an error incase someone misspells "mozila" or something.

core/Cargo.toml
justinmchase: https://github.com/denoland/deno/pull/11491#pullrequestreview-713068769

bartlomieju: @justinmchase moving to `runtime/` won't help as you need to use those crates in various crates in `extensions/` directory. 

I don't want you to do busy work trying guess where a "good" place for them might be. I will give your PR full review over the weekend and we can figure out a proper place for them then.

cli/file_fetcher.rs
```diff
@@ -220,7 +221,7 @@ impl FileFetcher {
     http_cache: HttpCache,
     cache_setting: CacheSetting,
     allow_remote: bool,
-    ca_data: Option<Vec<u8>>,
+    root_store: Option<RootCertStore>,
```
bartlomieju, line 224: `root_store` is quite ambiguous, how about using `root_cert_store` for variables that are of this type?

cli/Cargo.toml
```diff
@@ -75,6 +75,7 @@ pin-project = "1.0.7"
 regex = "1.4.3"
 rand = { version = "0.8.4", features = [ "small_rng" ] }
 ring = "0.16.20"
+rustls-native-certs = "0.5.0"
```
bartlomieju, line 78: Instead of adding this dependency to `cli/Cargo.toml` is should be reexported in one of the extension crates (probably `deno_net`)

cli/Cargo.toml
```diff
@@ -91,6 +92,7 @@ tokio = { version = "1.8.1", features = ["full"] }
 tokio-rustls = "0.22.0"
 uuid = { version = "0.8.2", features = ["v4", "serde"] }
 walkdir = "2.3.2"
+webpki-roots = "0.21.1"
```
bartlomieju, line 95: Ditto

cli/tests/integration/mod.rs
```diff
@@ -1,9 +1,9 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 use crate::itest;
+use deno_core::rustls;
 use deno_core::url;
-use deno_runtime::deno_net::ops_tls::rustls;
-use deno_runtime::deno_net::ops_tls::webpki;
+use deno_core::webpki;
```
bartlomieju, line 6: Revert these imports

core/Cargo.toml
bartlomieju: Okay, now I get what's going on in this PR; I suggest to add new crate in `extensions/` directory - let's call it `deno_tls` that will contain TLS utilities used by other crates. 

This will be useful for features implemented in https://github.com/denoland/deno/pull/11324

Once that's done it should be more or less done (provided that tests are passing)

bartlomieju (commented):
@justinmchase nice work! I pointed out some nitpicks and how to solve problem with adding multiple TLS related crates to `deno_core` crate.

cli/file_fetcher.rs
justinmchase: no problem

cli/Cargo.toml
justinmchase: Agreed, I removed some of these in other places, this one was missed.

justinmchase:
Ok thanks @bartlomieju I know how to proceed.

cli/program_state.rs
justinmchase: @bnoordhuis if I did `.ok()?` and the environment variable is not set that would return in an error, no? I want it to be ok to not have the variable, which would then fallback to the default behavior.

justinmchase: No you're right that works perfect and is cleaner. Thanks.

cli/program_state.rs
justinmchase: Ok, it returns an error now if you pass in an unsupported store:

Screen Shot 2021-08-03 at 2 27 20 PM

justinmchase:
Ok merged with `main` and all PR comments addressed and two unit tests were added, [which need to be reviewed](https://github.com/denoland/deno/pull/11491/files#diff-5c6fb275a66a1b02f99597f6e9b36297274abd4b16926b9fb2d412f94ee0700eR373).

The new tests call `https://deno.land` since its signed with a valid root CA covered by the mozilla certs. Arguably this is a forbidden action but I'm not sure how else to test it specifically.

These tests positively test that the mozilla certs connect to deno.land and positively test that not loading any ca certs (an empty root_cert_store) fails to connect to deno.land.

The only other test vector that isn't covered by existing tests would be to then put the right cert into the system keystore and test loading that... I don't know how to do that consistently across systems without some really complicated automation and frankly... it feels a bit tautological to test as that entire functionality exists in a different dependency we are adding which has its own testing.

I'll leave it up to the maintainers to decide which tests are needed.

justinmchase:
Curious... on Windows it seems to successfully make a request even without certs. Does it not have tls support by default? Or maybe its loading certs from the certificate store regardless? I'm not sure.

Screen Shot 2021-08-03 at 3 59 58 PM

bartlomieju:
@justinmchase thanks for adressing the review comments. This looks good to me now. I discussed this PR with @ry and we're debating if it's worth having a separate flag/env variable for this feature. We're considering always loading system certificates by default. I'll get back to you on this one.

ry:
I think we should control this feature via env var only. But I think the current `DENO_TLS_CA_STORE` is too complex and feature-full. I'd suggest a boolean valued environment variable which can be set (`DENO_SYSTEM_CA_STORE=1`) to use the system root CAs - otherwise we use the default built-in root CAs. We don't need more control than this.


ref https://github.com/nodejs/node/issues/3159

justinmchase:
@ry Just so I'm clear, the desired behavior is if the flag is `true` it would use the system store, if false, the bundled one. Its mutually exclusive and either one is loaded or the other, not both. Correct?

bartlomieju:
> @ry Just so I'm clear, the desired behavior is if the flag is `true` it would use the system store, if false, the bundled one. Its mutually exclusive and either one is loaded or the other, not both. Correct?

I'm not sure it makes sense to make them exclusive... I believe a better approach would be to add system certs to already loaded ones if the env var is present. Are there any contraindications for that?

justinmchase:
>  I believe a better approach would be to add system certs to already loaded ones if the env var is present...

@bartlomieju I don't have a strong opinion on it either way, just trying to clarify the desired behavior.

The only conceivable downside I can imagine is that if for some reason you didn't want to load the mozilla certs, there is no way to do that. Is that a problem in reality? I'm not sure. I can imagine an app where you're trying to lock down all outbound traffic to only destinations using a specific cert but, I've never actually needed it. I just want to be sure I'm on the same page.

lucacasonato:
The nice thing about the current implementation in this PR is that it gives you the ability to disable the builtins if you do not trust them (which is a valid concern IMO). It allows for the following:

1. `DENO_TLS_CA_STORE=`: trust no CA certificates, effectively disabling TLS. Individual certificates can be added to the empty store with `--cert`. This can be useful when you want to enforce that all TLS communication happens through a central proxy (configurable with HTTPS_PROXY env var).
2. `DENO_TLS_CA_STORE=system`: trust only system CA certs + `--cert`
3. `DENO_TLS_CA_STORE=mozilla`: trust only the builtins + `--cert`
4. `DENO_TLS_CA_STORE=system,mozilla` trust system and builtins + `--cert`, preferring chains that terminate in a system cert.
5. `DENO_TLS_CA_STORE=mozilla,system` trust system and builtins + `--cert`, preferring chains that terminate in a mozilla trusted cert. 

The `DENO_SYSTEM_CA_STORE ` only supports option 4 (or 5, but not both), and option 3. Option 1 and 2 are not supported and would require an extra env var: `DENO_MOZILLA_CA_STORE=off`.

Node supports all these use-cases via esoteric OpenSSL env vars. We can centralize and clean this up.

bartlomieju:
> The nice thing about the current implementation in this PR is that it gives you the ability to disable the builtins if you do not trust them (which is a valid concern IMO). It allows for the following:
> 
> 1. `DENO_TLS_CA_STORE=`: trust no CA certificates, effectively disabling TLS. Individual certificates can be added to the empty store with `--cert`. This can be useful when you want to enforce that all TLS communication happens through a central proxy (configurable with HTTPS_PROXY env var).
> 2. `DENO_TLS_CA_STORE=system`: trust only system CA certs + `--cert`
> 3. `DENO_TLS_CA_STORE=mozilla`: trust only the builtins + `--cert`
> 4. `DENO_TLS_CA_STORE=system,mozilla` trust system and builtins + `--cert`, preferring chains that terminate in a system cert.
> 5. `DENO_TLS_CA_STORE=mozilla,system` trust system and builtins + `--cert`, preferring chains that terminate in a mozilla trusted cert.
> 
> The `DENO_SYSTEM_CA_STORE ` only supports option 4 (or 5, but not both), and option 3. Option 1 and 2 are not supported and would require an extra env var: `DENO_MOZILLA_CA_STORE=off`.
> 
> Node supports all these use-cases via esoteric OpenSSL env vars. We can centralize and clean this up.

This is quite interesting and very flexible. My only worry is how do we test all those combinations?

justinmchase:
> My only worry is how do we test all those combinations?

1. I did add a test which tested this scenario specifically. It uses an empty store then attempt to fetch `https://deno.land` which then fails as expected.
2. I don't know how to test this easily, you'd have to ensure that all testing environments have some consistency to make it a reliable test which would require some complex automation probably. Or maybe you could add a cert in the CI server and then add some kind of test config flag which would only run it on the CI server by default? On the other hand it may be fairly tautological to test that the store has the certs you're expecting, the library this PR is using is theoretically doing that testing already.
3. I added a test for this also, specifically adds the mozilla certs then requests `https://deno.land` which succeeds as expected.
4. Same as 2.
5. Same as 2. plus I am not sure how to we could tell which cert it ends up using since those are all down in lower level libraries.

bnoordhuis (approved):
Left some comments but your PR mostly looks (very!) good.

> Node supports all these use-cases via esoteric OpenSSL env vars. We can centralize and clean this up.

We'll be doing even better. Node can only use the system's openssl certificate bundle, whereas we will be able to use the native macos and windows stores.

It's an often-requested feature that never happened because openssl can't do it. Take https://github.com/nodejs/node/issues/39657 - a feature request from < 24 hours ago.

ry:
If Ben and Luca are ok with this, then I have no objections. We just need to make sure this is properly documented after this patch lands.

cli/program_state.rs
```diff
@@ -68,11 +72,51 @@ impl ProgramState {
...
+    // Should this do a permissions check?
     let ca_file = flags.ca_file.clone().or_else(|| env::var("DENO_CERT").ok());
-    let ca_data = match &ca_file {
-      Some(ca_file) => Some(read(ca_file).context("Failed to open ca file")?),
-      None => None,
-    };
+    if let Some(ca_file) = ca_file {
+      let certfile = File::open(&ca_file)?;
+      let mut reader = BufReader::new(certfile);
+      root_cert_store.add_pem_file(&mut reader).unwrap();
```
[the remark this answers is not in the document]
justinmchase: I tried just changing `.unwrap()` to `?;` but for some reason it has this error:

```
error[E0277]: the trait bound `(): StdError` is not satisfied
   --> cli/program_state.rs:113:48
    |
113 |       root_cert_store.add_pem_file(&mut reader)?;
    |                                                ^ the trait `StdError` is not implemented for `()`
    |
    = note: required because of the requirements on the impl of `std::convert::From<()>` for `anyhow::Error`
    = note: required by `std::convert::From::from`
```

The docs on this function say this:

> Errors are non-specific; they may be io errors in rd and PEM format errors, but not certificate validity errors.
> This is because large collections of root certificates often include ancient or syntactically invalid certificates. CAs are competent like that.

I interpretted this to mean it basically eats the specific error and returns a default `()` which is probably useless. 

Therefore I changed it to do this:
```rs
if let Err(_err) = root_cert_store.add_pem_file(&mut reader) {
  return Err(anyhow!("Unable to add pem file to certificate store"))
}
```

@bnoordhuis If you know a more elegant one liner let me know! But I updated all of the places where this was being called like this.

extensions/net/ops_tls.rs
```diff
@@ -713,12 +681,25 @@ where
...
+  };
+
   let hostname_dns = DNSNameRef::try_from_ascii_str(hostname)
     .map_err(|_| invalid_hostname(hostname))?;
 
+  let root_cert_store = state
+    .borrow()
+    .borrow::<DefaultTlsOptions>()
+    .root_cert_store
+    .clone();
```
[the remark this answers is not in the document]
justinmchase: That is why I did that, the roots were cloned before and this ended up following that pattern for better or worse.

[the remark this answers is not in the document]
justinmchase: Given how this already feels to be max size for this PR I'm happy to leave this for a future refactor.

justinmchase:
> @ry If Ben and Luca are ok with this, then I have no objections

Just to be clear by "this" I assume you mean keeping the functionality as it is in this pr right now? To have the `DENO_TLS_CA_STORE` with string[] arguments, not the `DENO_SYSTEM_CA_STORE=1` boolean flag?

If so then this PR is ready I believe. I linted, formatted and squashed the changes in the branch as well as manually tested one more time.

cli/program_state.rs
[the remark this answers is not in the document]
crowlKats: @justinmchase You could do
```rust
root_cert_store.add_pem_file(&mut reader).map_err(|_| anyhow!("Unable to add pem file to certificate store"))?
```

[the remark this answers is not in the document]
bnoordhuis: `if let Err(_err) = ...` is fine, as is `.map_err(|_| ...)`.

`if let Err(()) = ...` or `.map_err(|()| ...)`  is arguably little more obvious still, otherwise a reader of the code might wonder why we're throwing away the error.

bnoordhuis:
For posterity: I don't have opinions on _how_ this feature should be activated. Bartek or Luca are more qualified to make that call.

extensions/tls/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+  tls_config.set_persistence(CLIENT_SESSION_MEMORY_CACHE.clone());
+  tls_config.root_store =
+    root_cert_store.unwrap_or_else(create_default_root_cert_store);
+
+  // If a custom cert is specified, add it to the store
+  if let Some(cert) = ca_data {
+    let reader = &mut BufReader::new(Cursor::new(cert));
+    // This function does not return specific errors, if it fails give a generic message.
+    if let Err(_err) = tls_config.root_store.add_pem_file(reader) {
+      return Err(anyhow!("Unable to add pen file to certificate store"));
```
bnoordhuis, line 95:
```suggestion
      return Err(anyhow!("Unable to add pem file to certificate store"));
```
(also happens in two other places)

bartlomieju:
> > @ry If Ben and Luca are ok with this, then I have no objections
> 
> Just to be clear by "this" I assume you mean keeping the functionality as it is in this pr right now? To have the `DENO_TLS_CA_STORE` with string[] arguments, not the `DENO_SYSTEM_CA_STORE=1` boolean flag?
> 
> If so then this PR is ready I believe. I linted, formatted and squashed the changes in the branch as well as manually tested one more time.

That's right, I'll give this PR another review tonight. 

Could you please open a PR in https://github.com/denoland/manual that describes this functionality? Turning Luca's [comment](https://github.com/denoland/deno/pull/11491#issuecomment-892988929) into some kind of doc would be awesome.

extensions/tls/lib.rs
justinmchase: Derp, thanks. I fixed the other two copy+pasted typos as well.

Cargo.lock
```diff
@@ -2342,6 +2357,12 @@ version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "624a8340c38c1b80fd549087862da4ba43e08858af025b236e509b6649fc13d5"
 
+[[package]]
+name = "openssl-probe"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28988d872ab76095a6e6ac88d99b54fd267702734fd7ffe610ca27f533ddb95a"
+
```
bartlomieju, line 2365: Yikes 😬  it comes from `rustls-native-certs` and it doesn't seem it could be in any way skipped (no feature flags in that crate). I looked up that crate's code and it doesn't do any interaction with OpenSSL so I guess we can live with that.

cli/program_state.rs
```diff
@@ -68,11 +72,51 @@ impl ProgramState {
...
+            .roots;
+          root_cert_store.roots.extend(roots);
+        }
+        _ => {
+          return Err(anyhow!("Unknown certificate store \"{}\" specified (allowed: \"system,mozilla\")", store));
+        }
+      }
+    }
+
+    // Should this do a permissions check?
```
bartlomieju, line 109: No, loading certificates is required to start up the process and there are no permission checks, this comment can be removed.

cli/http_util.rs
```diff
@@ -390,6 +366,64 @@ mod tests {
...
+
+    println!("{:?}", result);
+    if let Ok(FetchOnceResult::Code(body, _headers)) = result {
+      assert!(!body.is_empty());
+    } else {
+      panic!();
+    }
+  }
+
+  // todo: Windows should verify certs too and fail to make this request without ca certs
```
bartlomieju, line 398: Nitpick: add your handle to the TODOs you create (`TODO(@justinmchase)`), unfortunately we can't enforce that via rustfmt/clippy.

cli/standalone.rs
```diff
@@ -227,13 +233,26 @@ pub async fn run(
...
+    .root_cert_store
+    .clone()
+    .unwrap_or_else(create_default_root_cert_store);
+
+  if let Some(cert) = metadata.ca_data {
+    let reader = &mut BufReader::new(Cursor::new(cert));
+    // This function does not return specific errors, if it fails give a generic message.
+    if let Err(_err) = root_cert_store.add_pem_file(reader) {
+      return Err(anyhow!("Unable to add pem file to certificate store"));
+    }
```
bartlomieju, line 246: This is good

extensions/net/ops_tls.rs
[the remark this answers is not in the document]
bartlomieju: That's fine, but please add a TODO here so we don't forget about it :)

extensions/tls/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+use reqwest::header::HeaderMap;
+use reqwest::header::USER_AGENT;
+use reqwest::redirect::Policy;
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::io::BufReader;
+use std::io::Cursor;
+use std::sync::Arc;
+
+/// This extension has no runtime apis only exports some shared native functions
```
bartlomieju, line 29: 👍

extensions/tls/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+      sessions.remove(&key);
+    }
+    sessions.insert(key, value);
+    true
+  }
+}
+
+pub fn create_default_root_cert_store() -> RootCertStore {
+  let mut root_cert_store = RootCertStore::empty();
+  // todo: Consider also loading the system keychain here
```
bartlomieju, line 76: This TODO seems to be no longer valid

bartlomieju (commented):
@justinmchase this PR is really well implemented, very nice work! I have only a couple minor nitpicks left and we're good to merge.

Cargo.lock
justinmchase: Interesting, I think it does say something about it [in the docs](https://crates.io/crates/rustls-native-certs) now that you mention it:

> On Linux and other UNIX-like operating systems, the openssl-probe crate is used to discover the filename of the system CA bundle.

extensions/tls/lib.rs
justinmchase: I think it is, the comment is suggesting that we should consider altering the _default_ behavior to also load the system certs.

I was actually planning on following up this PR with one that made that change but could be postponed till a future major release.

bartlomieju: Okay, makes sense 👍

justinmchase:
One of the tests returned this result, I'm not sure how to run it locally to debug it. It passed in a previous iteration, so it's an interesting result. How can I tell which test is failing here exactly?

Screen Shot 2021-08-06 at 3 50 05 PM

bartlomieju:
> One of the tests returned this result, I'm not sure how to run it locally to debug it. It passed in a previous iteration, so it's an interesting result. How can I tell which test is failing here exactly?
> 
> Screen Shot 2021-08-06 at 3 50 05 PM

@justinmchase it looks like this error was returned when pulling code for `deno_std` - before any test was run - ie. when process was starting up and downloading dependencies.

justinmchase:
Succeeded this time, I didn't change anything. Weird. 🤷 

All comments have been addressed I believe. Thanks all.

bartlomieju (approved):
LGTM, thank you @justinmchase!

Issue #11456, opened 2021-07-10
lsp: types can get "lost" with import maps


erfanium:
It's kinda strange. When I open vscode, some variable types recognized as  `unknown` which is incorrect. look at this screenshot:

![Screenshot from 2021-07-10 21-20-47](https://user-images.githubusercontent.com/47688578/125171166-fb8d1200-e1c7-11eb-8bac-8c9b8a778824.png)

After I go to `userIdKey` definition (via F12) and do a change (for example add a space). It will fix the problem (not always):

![Screenshot from 2021-07-10 21-40-56](https://user-images.githubusercontent.com/47688578/125171244-5e7ea900-e1c8-11eb-9e64-263a6abd5bee.png)

~~Unfortunately I couldn't make a reproduce repo for this.~~
reproduction: https://github.com/erfanium/deno-lsp-bug-repro

vscode: 1.58.0  
deno: 1.11.5  
extension: 3.7.0

kitsonk:
A more concise reproduction: https://github.com/erfanium/deno-lsp-bug-repro/tree/master

Issue #11146, opened 2021-06-26
lsp: diagnostics not cleared after updates to `import_map.json`


sramam:
vs-code needs to be restarted after updates to `import_map.json`

- `.vscode/settings`
```
{
  "deno.enable": true,
  "deno.lint": true,
  "deno.unstable": true,
  "deno.importMap": "./import_map.json",
}
``` 
- `import_map.json`
```
{
  "imports": {
    "git": "https://esm.sh/isomorphic-git?target=deno"
  }
}
```

Before a restart:
![image](https://user-images.githubusercontent.com/518574/123518963-49006f80-d65d-11eb-97ba-73a813a67361.png)

After a restart:
![image](https://user-images.githubusercontent.com/518574/123519002-8664fd00-d65d-11eb-983d-71f83c9c86a0.png)

kitsonk:
This is due to diagnostics not being invalidated or cleared in the lsp when the config changes. I suspect editing the document would have the same effect as restarting.

sramam:
True dat! I can confirm this happens on editing the document too.

sramam:
Are local imports different than remote imports? 

These stale LSP errors seem to only impact changes made to import_map.json or to relative paths from such imports. 
Changes to local files imports take effect instantly.

I'm don't yet have a deep understanding of the inner workings of Deno - apologies if this is not useful.

Issue #10439, opened 2021-04-23
lsp: not resolving extending interfaces with import map


darkdamp:
There seem to be an issue with extending interfaces not resolving sometimes.

I open my workplace in vscode and enter file `a.ts` and it doesn't detect the interface extending `Context` from `oak`
![image](https://user-images.githubusercontent.com/11677530/115927725-28f9d780-a485-11eb-96b7-809d201e5f12.png)

But when just 'entering' the file which returns the interface that extends `Context`. It gets resolved
![image](https://user-images.githubusercontent.com/11677530/115931778-0919e200-a48c-11eb-8702-b1cbe905c1c7.png)

And it now works if I tab back to `a.ts`
![image](https://user-images.githubusercontent.com/11677530/115931655-ce17ae80-a48b-11eb-9ad5-2669bb3e8d36.png)

If this is not a bug, I'd appreciate any pointer on how to 'load' all files in a workplace to resolve the intellisense at startup

`.vscode/settings.json`
```
{
    "deno.enable": true,
    "deno.importMap": "./import_map.json",
    "deno.config": "./tsconfig.json",
    "deno.unstable": true
 }
```

vscode: 1.55.2
deno: 1.9.2
extension: 3.3.0

Thanks

PR #11496, opened 2021-07-23
fix(lsp): handle importmaps properly

kitsonk:
This fixes several issues with import maps and the language server:

- When the import map was updated, dependencies were not recalculated, meaning changes the open documents were required to "update" the files.
- Non-open resources did not actually have import maps applied to them, which resulted in situations where types "corrected themselves" when opening documents in the editor.
- Open documents in the editor didn't use the language ID, instead inferred the type from the module specifier, which is an imprecise way of determining the type of the file.

Fixes: #11146
Fixes: #11456
Fixes: #10439


cli/lsp/tsc.rs
```diff
@@ -49,6 +49,14 @@ use tokio::sync::oneshot;
 const FILE_EXTENSION_KIND_MODIFIERS: &[&str] =
   &[".d.ts", ".ts", ".tsx", ".js", ".jsx", ".json"];
 
+lazy_static::lazy_static! {
+  /// Used in situations where tsc is holding onto a reference to a module
+  /// specifier that doesn't exist anymore and we cannot return an option or
+  /// result. This occurs in situations like when an import map is updated but
+  /// tsc has a cached version of a document in memory.
+  static ref INVALID_SPECIFIER: ModuleSpecifier = ModuleSpecifier::parse("deno://invalid").unwrap();
```
bartlomieju, line 57: Not sure if on purpose but this is duplicated in `cli/lsp/sources.rs`

bartlomieju (approved):
LGTM, nice fix

PR #11499, opened 2021-07-23
feat: add --enable-testing-features-do-not-use

lucacasonato:
This flag does nothing yet. It is added in preparation for the addition
of classic workers (#11338).


bartlomieju (approved):

andreubotella:
Should classic workers also require `--unstable`?

bartlomieju:
> Should classic workers also require `--unstable`?

If they will be hidden behind this flag I don't see the reason we need to require `--unstable` as well.

PR #11500, opened 2021-07-23
refactor: use `primordials` in runtime, extensions and core

magurotuna:
I'm implementing new lint rule `prefer-primordials`. To see if it works fine I ran this rule for the runtime, extensions, and core JS code. The linter found a bunch of spots where primordials could be used, so this PR fixes them.

ref #11224

### DIFF
--- a/extensions/webidl/00_webidl.js
+++ b/extensions/webidl/00_webidl.js
@@ -12,8 +12,10 @@
   const {
     ArrayBuffer,
     ArrayBufferIsView,
+    ArrayPrototypeForEach,
     ArrayPrototypePush,
     ArrayPrototypeSort,
+    ArrayIteratorPrototype,
     BigInt,
     BigIntAsIntN,
     BigIntAsUintN,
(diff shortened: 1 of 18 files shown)


extensions/webidl/00_webidl.js
```diff
@@ -921,9 +926,7 @@
 
   const _iteratorInternal = Symbol("iterator internal");
 
-  const globalIteratorPrototype = ObjectGetPrototypeOf(ObjectGetPrototypeOf(
-    [][SymbolIterator](),
-  ));
+  const globalIteratorPrototype = ObjectGetPrototypeOf(ArrayIteratorPrototype);
```
magurotuna, line 929: Not really sure if this change is fine. 
Otherwise, could `[][SymbolIterator]()` be replaced with `primordials.ArrayPrototype[SymbolIterator]()`? If it could I'd be happy to replace it.

bartlomieju: I'm also not entirely sure if this is correct, but I'm going to tentatively land it

bartlomieju (approved):
LGTM, thank you @magurotuna

PR #11505, opened 2021-07-23
fix(http/ws): support multiple options in connection header

### LINKED ISSUE #11494: `Deno.upgradeWebSocket` fails on `connection: "keep-alive, Upgrade"` header
Hi, I'm trying out the unstable `Deno.upgradeWebSocket` and encountered an issue accepting a WebSocket from Firefox. 

Firefox is sending this `connection: "keep-alive, Upgrade"` header when connecting to a websocket. That causes throw [here](https://github.com/denoland/deno/blob/8b34f07bb000a16b60b119ee24f5c0b4a5f7f937/extensions/http/01_http.js#L324)

```ts
  function upgradeWebSocket(request, options = {}) {

    // ...

    if (request.headers.get("connection")?.toLowerCase() !== "upgrade") {
      throw new TypeError(
        "Invalid Header: 'connection' header must be 'Upgrade'",
      );
    }

    // ...

```

 while the [current std logic for accepting sockets](https://github.com/denoland/deno_std/blob/bf73521428ce724d31dba0d403d14acc929f39ab/ws/mod.ts#L396) doesn't care about the `connection` header at all.
 
 ```ts
 export function acceptable(req: { headers: Headers }): boolean {
  const upgrade = req.headers.get("upgrade");
  if (!upgrade || upgrade.toLowerCase() !== "websocket") {
    return false;
  }
  const secKey = req.headers.get("sec-websocket-key");
  return (
    req.headers.has("sec-websocket-key") &&
    typeof secKey === "string" &&
    secKey.length > 0
  );
}
 ```

silen-z:
The "connection" header can have multiple connection-options separated by commas:
https://datatracker.ietf.org/doc/html/rfc7230#section-6.1
https://datatracker.ietf.org/doc/html/rfc7230#section-7

Fixes #11494

### DIFF
--- a/extensions/http/01_http.js
+++ b/extensions/http/01_http.js
@@ -22,8 +22,11 @@
   const {
     ArrayPrototypeIncludes,
     ArrayPrototypePush,
+    ArrayPrototypeSome,
     Promise,
     StringPrototypeIncludes,
+    StringPrototypeToLowerCase,
+    StringPrototypeTrim,
     StringPrototypeSplit,
     Symbol,
     SymbolAsyncIterator,
@@ -321,7 +324,14 @@
       );
     }
 
-    if (request.headers.get("connection")?.toLowerCase() !== "upgrade") {
+    const connection = request.headers.get("connection");
+    const connectionHasUpgradeOption = connection !== null &&
+      ArrayPrototypeSome(
+        StringPrototypeSplit(connection, ","),
+        (option) =>
+          StringPrototypeToLowerCase(StringPrototypeTrim(option)) === "upgrade",
+      );
+    if (!connectionHasUpgradeOption) {
       throw new TypeError(
         "Invalid Header: 'connection' header must be 'Upgrade'",
       );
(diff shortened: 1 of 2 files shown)


crowlKats (approved):
LGTM

kitsonk (approved):
LGTM

extensions/http/01_http.js
```diff
@@ -321,7 +324,14 @@
       );
     }
 
-    if (request.headers.get("connection")?.toLowerCase() !== "upgrade") {
+    const connection = request.headers.get("connection");
+    const connectionHasUpgradeOption = connection !== null &&
+      ArrayPrototypeSome(
+        StringPrototypeSplit(connection, ","),
```
kitsonk, line 330: Actually using a regex of `/\s*,\s*/` here would be better and removes the need for trim.

kitsonk (changes requested):
Actually I think we should change one thing.

kitsonk (approved):
Awesome, thank you!

extensions/http/01_http.js
silen-z: Yeah that is definitely better, I didn't really wanted to make the whole check one "big" regex, but I forgot split takes one as well

PR #11509, opened 2021-07-24
feat(runtime/permissions): support IP CIDR ranges in net allowlist

### LINKED ISSUE #9816: feature: support cidr blocks egress control with --allow-net
currently --allow-net supports only IPs and domains, unfortunately it doesn't support cidr blocks yet

e.g.
```
deno run --allow-net=xyz.com,abc.com,10.0.0.11/24
```

sh7dm:
Fixes #9816


bartlomieju:
@lucacasonato requested not to land this one until he can review it. Removing milestone for now

bnoordhuis:
Can I suggest switching to `Ipv4Network` and `Ipv6Network` from the [ipnetwork](https://crates.io/crates/ipnetwork) crate? It'll be easier to read and maintain than the regex approach.

It also supports IPv6 CIDRs like `fc00::/7`, something your PR doesn't currently handle.

sh7dm:
Yes, currently I only did that for IPv4. `ipnetwork` looks really good and should make code clearer. If Deno team members are not against including this crate as a dependency (seems not installed), I will integrate it in place of regex.

ry:
@sh7dm the added dependency seems good

sh7dm:
`ipnet` crate which is already being built with some Deno's dependencies should do the job as well: https://docs.rs/ipnet/2.3.1/ipnet/struct.Ipv4Net.html#examples-11.

lucacasonato (commented):
I haven't reviewed implementation, but this would likely block full URL network allow list due to possible parsing ambiguity between CIDR ranges and URLs (is this concern warranted?).

With full URL network allow list I mean the following: ` --allow-net=https://api.github.com/meta` would allow fetching `https://api.github.com/meta`, but not `https://api.github.com/users/lucacasonato`. Currently filtering is just per domain.

runtime/permissions.rs
```diff
@@ -174,12 +176,26 @@ pub struct WriteDescriptor(pub PathBuf);
...
+}
+
 impl NetDescriptor {
   fn new<T: AsRef<str>>(host: &&(T, Option<u16>)) -> Self {
     NetDescriptor(host.0.as_ref().to_string(), host.1)
   }
 
   pub fn from_string(host: String) -> Self {
+    if is_cidr(&host) {
+      return NetDescriptor(host, None);
```
crowlKats, line 196: So no port support?

sh7dm: Yes, because It becomes hard with IPv6

sh7dm:
> I haven't reviewed implementation, but this would likely block full URL network allow list due to possible parsing ambiguity between CIDR ranges and URLs (is this concern warranted?).

Can also be a trouble. Maybe we should either add some prefix or another CLI flag for CIDR?

runtime/permissions.rs
lucacasonato: IPv6 addresses should always be contained in square brackets (`[ffff::]`). It should be unambiguous:

`127.0.0.1/24:433`
`[ffff:]/48:433`

sh7dm: We _can_ enforce such for Deno, but it's not like so by default in [IPv6 CIDR](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation)

lucacasonato:
Actually maybe it is not even ambiguous without square brackets:

The port should come after the CIDR range, so after the `/`. If the string does not contain `/`, then it is not a cidr range.

So should look like this:
```
ffff:ffff:ffff::/24:443
|-------1------||2||3-|

1: address
2: subnet size
3: port
```

If the subnet size is missing `address` must be enclosed in square brackets if it is IPv6. We already enforce this.

sh7dm:
Yes, but what about URLs? How do we find out that `ffff:ffff:ffff::/24` is CIDR allowing any connection to any IP starting with `ffff:ffff:ffff...` or is allowing requesting HTTP path `/24` from `ffff:ffff:ffff:0:0:0:0:0`?

lucacasonato:
@sh7dm IPv6 in URLs already requires that you wrap the address in square brackets: https://url.spec.whatwg.org/#valid-host-string

lucacasonato (commented):
Overall, happy to get this feature. I have a few more asks though:

- CIDR ranges are re-parsed on each permission check. CIDR ranges should be parsed once, and then stored in the NetDescriptor as a `ipnet::IpNet` (maybe turn `NetDescriptor` into an enum?).

- Please also add a test that IPv6 CIDR where the IP address is surrounded by `[ ]` does not work.

sh7dm:
Well, enum should be a great idea

bnoordhuis:
@sh7dm Do you still want to move forward with this? Seems like a great feature to have.

sh7dm:
Quite busy currently, sorry. No time for builds and proper testing. I do hope maintainers can help delivering this feature.

bnoordhuis:
That's too bad but thanks for your reply. Since this PR has merge conflicts, I'll go ahead and take the liberty of closing it.

PR #11510, opened 2021-07-24
chore(crypto): use WebIDL BufferSource in types

littledivy:
Ref https://github.com/denoland/deno/pull/11367#discussion_r676006350


bartlomieju (approved):
LGTM

PR #11515, opened 2021-07-25
fix(cli/tools/test_runner): side-load test modules

### LINKED ISSUE #11504: Side load test modules
Test modules are loaded as the main module as of 1.9 or so, this undocumented and has been mentioned as non stable behaviour in a few issues.

The problem is that this can have unintended side effects for modules that put program behaviour behind `import.meta.main`.

We should side-load test modules again.

caspervonb:
This fixes a regression introduced in 1.9 where test modules became main modules.

This is undesirable as it will cause executables to start running when they're guarded with `import.meta.main`.

We now side load (import) the module being tested in a generated module but still mark it as the main module in places where it could be visible like the inspector (effectively everywhere but the import.meta.main flag).

Closes #11504

### DIFF
--- a/cli/tools/test_runner.rs
+++ b/cli/tools/test_runner.rs
@@ -34,6 +34,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::time::Instant;
 use swc_common::comments::CommentKind;
+use uuid::Uuid;
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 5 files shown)


cli/tools/test_runner.rs
```diff
@@ -262,10 +263,38 @@ where
...
+    })
+  );
+
+  let test_file = File {
+    local: test_module.to_file_path().unwrap(),
+    maybe_types: None,
+    media_type: MediaType::JavaScript,
+    source: test_source.clone(),
+    specifier: test_module.clone(),
+  };
```
bartlomieju, line 294: Will the module still be typechecked if it's JS by default?

caspervonb: Yeah we don't seem to care if the entry point is JavaScript or not (which is weird, we should but that's why we're talking about no-check by default I suppose).

I made sure to include the `Check` diagnostic in the tests when I hardened them, and there are no failures.

bartlomieju (approved):
LGTM

PR #11527, opened 2021-07-27
feat(lsp): ability to set DENO_DIR via settings

kitsonk:
Ref: denoland/vscode_deno#287

Add the ability to specify an override for the DENO_DIR via the language server configuration.  Previously the language server would have to be executed in an environment with the `DENO_DIR` environment setting, which makes it difficult to manage and customise per workspace.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -139,6 +139,9 @@ pub struct Flags {
   pub allow_write: Option<Vec<PathBuf>>,
   pub ca_file: Option<String>,
   pub cache_blocklist: Vec<String>,
+  /// This is not exposed as an option in the CLI, it is used internally when
+  /// the language server is configured with an explicit cache option.
+  pub cache_path: Option<PathBuf>,
   pub cached_only: bool,
   pub config_path: Option<String>,
   pub coverage_dir: Option<String>,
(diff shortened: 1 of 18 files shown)


bartlomieju (approved):
Not ideal that it required changes to `Flags`, but given that comment is explaining well why it's needed this PR looks good to me.

cli/flags.rs
```diff
@@ -139,6 +139,9 @@ pub struct Flags {
   pub allow_write: Option<Vec<PathBuf>>,
   pub ca_file: Option<String>,
   pub cache_blocklist: Vec<String>,
+  /// This is not exposed as an option in the CLI, it is used internally when
+  /// the language server is configured with an explicit cache option.
+  pub cache_path: Option<PathBuf>,
```
dsherret, line 144: I also wish we didn't have to do this, but I don't know any other suggestion other than creating a new struct with specific options for the `ProgramState`, which might be more trouble than it's worth.

dsherret (approved):
LGTM too.

cli/flags.rs
kitsonk: Yeah, that or another argument to the new function, but that was even uglier.

PR #11540, opened 2021-07-27
fix: flaky worker test

bartlomieju:
The test was flaky because it was using port number that a lot of other tests are using as well.


satyarohith (approved):
LGTM.

PR #11554, opened 2021-07-30
chore: upgrade Rust to 1.54.0

magurotuna:
Upgrades Rust to 1.54.0 and applys clippy's fixes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
       - name: Install Rust
         uses: hecrj/setup-rust-action@v1
         with:
-          rust-version: 1.53.0
+          rust-version: 1.54.0
 
       - name: Install clippy and rustfmt
         if: matrix.kind == 'lint'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,7 +253,7 @@ jobs:
             ~/.cargo/registry/index
             ~/.cargo/registry/cache
             ~/.cargo/git/db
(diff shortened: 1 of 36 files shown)


kitsonk:
@kt3k I don't know if you know, but it seems that somehow the upgrade to Rust 1.54 causes us to run out of disk space on CI on Windows. I suspect it is something to do with our build caches, but I don't know enough to "fix" it.

kt3k:
'Post Clone repository' step says `There is not enough space on the disk. : 'D:\a\_temp\_runner_file_commands\set_env_b44b1bfe-1ffe-40fd-8e4d-37c9df5dd52d'`, but I think that step only does clean-ups. So I doubt we actually run out of disk space..

'Test release' step shows `Error: thread 'basic' panicked at 'assertion failed: build_plugin_output.status.success()', test_plugin\tests\integration_tests.rs:21:3`. I guess that one would be the actual problem we need to address

I'll rerun CI anyway

kitsonk:
@kt3k yeah, I re-ran CI already and it continuously fails with the same error... 😕

bartlomieju:
> @kt3k I don't know if you know, but it seems that somehow the upgrade to Rust 1.54 causes us to run out of disk space on CI on Windows. I suspect it is something to do with our build caches, but I don't know enough to "fix" it.

Changing cache keys here: 
https://github.com/denoland/deno/blob/c909faf9e6cd2964398da7c0852d0229cdd1a22b/.github/workflows/ci.yml#L248-L287
should help

bartlomieju (approved):
LGTM, thank you @magurotuna

PR #11556, opened 2021-07-30
chore(core): use oneshot channel in mod_evaluate()

bnoordhuis:
Oneshot is more appropriate because mod_evaluate() only sends a single
value.

It also makes it easier to use it correctly. As an embedder, I wasn't
sure if I'm expected to drain the channel or not.


bartlomieju (approved):
LGTM

Issue #10992, opened 2021-06-15
Add `Origin` header to outgoing requests for fetch


lucacasonato:
With `--location` set the header should be the location origin (or worker location origin). If it is not set the origin should be `null`.

zlumyo:
I can try this one.

Should an `Origin` header provided manually by user overwrite a value of `--location`?

lucacasonato:
@zlumyo Yes.

F3n67u:
@zlumyo is there any progress about this issue?

zlumyo:
@F3n67u sorry, I've recently got trouble with COVID. If you can pick it up, that will be great.

F3n67u:
> @F3n67u sorry, I've recently got trouble with COVID. If you can pick it up, that will be great.

yep，I can pick it up.

PR #11557, opened 2021-07-30
feat(extensions/fetch): Add Origin header to outgoing requests for fetch

F3n67u:
fix https://github.com/denoland/deno/issues/10992


ry (approved):
@F3n67u Thanks for the patch. LGTM.

extensions/fetch/26_fetch.js
```diff
@@ -429,6 +431,17 @@
...
+      if (
+        baseURL &&
+        (requestObject.method !== "GET" && requestObject.method !== "HEAD")
+      ) {
+        ArrayPrototypePush(request.headerList, [
+          "origin",
+          new URL(baseURL).origin,
+        ]);
+      }
+
```
lucacasonato, line 444: This is not correct. The header should be added as per of a different stage of the implementation.

The origin header should also respect request referrer policy.

lucacasonato (commented):
Looking at the spec again, I am unsure about adding the header now. It is meant for CORS and doesn't really make much sense other than spec compliance.

F3n67u:
> Looking at the spec again, I am unsure about adding the header now. It is meant for CORS and doesn't really make much sense other than spec compliance.

Is there any use case where `origin` header should added? CORS is not a problem for deno because deno doesn't follow same-origin policy now.

extensions/fetch/26_fetch.js
F3n67u: this section? https://fetch.spec.whatwg.org/#append-a-request-origin-header

lucacasonato: Yeah, take a look at where it is called.

PR #11562, opened 2021-07-31
fix(lsp): fix parent process checker to not output to stderr before exiting

dsherret:
This change causes the parent process checker to actually exit on Mac and it seems some distros of linux.


kitsonk (approved):
LGTM

PR #11568, opened 2021-08-02
feat(std) support AbortSignal in writeFile

benjamingr:
Refs #10181 

This continues the work in https://github.com/denoland/deno/pull/10943 and adds AbortSignal support to `writeFile` (and since it uses the same infrastructure `writeTextFile`)

cc @lucacasonato and @ry who reviewed the last one :)

PR #11569, opened 2021-08-03
feat(core) support AbortSignal in Deno.listen

benjamingr:
Refs #10181

This continues the work in #10943 and #11568 adds AbortSignal support to Deno.listen (and since it uses the same infrastructure writeTextFile)

There shouldn't be a `DOMException` here for AbortError looking at the guidance at WHATWG (the guidance is about the signature and errors of methods returning promises which isn't really the case here). I opted to do the same thing we did for Node.js

In all likelihood there _should_ be weak event listeners (internal) in Deno core but that's a bigger refactor and honestly the event_target implementation is kind of a mess.

### DIFF
--- a/extensions/net/01_net.js
+++ b/extensions/net/01_net.js
@@ -207,7 +207,16 @@
       ...options,
     });
 
-    return new Listener(res.rid, res.localAddr);
+    const listener = new Listener(res.rid, res.localAddr);
+    if (options?.signal) {
+      // TODO(benjamingr) there is memory leak potential here if the signal outlives the
+      // server significantly it can retain it. Ideally Deno would have internal weak
+      // event listener like Node does *or* it would have an event for the server closing.
+      options?.signal?.addEventListener("abort", () => listener.close(), {
+        once: true,
+      });
+    }
+    return listener;
   }
 
   async function connect(options) {
--- a/extensions/net/02_tls.js
+++ b/extensions/net/02_tls.js
@@ -52,6 +52,7 @@
     hostname = "0.0.0.0",
     transport = "tcp",
     alpnProtocols,
+    signal,
   }) {
     const res = opListenTls({
       port,
@@ -60,12 +61,11 @@
       hostname,
       transport,
       alpnProtocols,
-      signal,
     });
     const listener = new TLSListener(res.rid, res.localAddr);
     if (signal) {
       // TODO(benjamingr) there is memory leak potential here. See comment in 01_net.js
-      options?.signal?.addEventListener("abort", () => listener.close(), {
+      signal.addEventListener("abort", () => listener.close(), {
         once: true,
       });
     }
--- a/extensions/net/02_tls.js
+++ b/extensions/net/02_tls.js
@@ -60,8 +60,16 @@
       hostname,
       transport,
       alpnProtocols,
+      signal,
     });
-    return new TLSListener(res.rid, res.localAddr);
+    const listener = new TLSListener(res.rid, res.localAddr);
+    if (signal) {
+      // TODO(benjamingr) there is memory leak potential here. See comment in 01_net.js
+      options?.signal?.addEventListener("abort", () => listener.close(), {
+        once: true,
+      });
+    }
+    return listener;
   }
 
   async function startTls(
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -598,3 +598,20 @@ unitTest(
     listener.close();
   },
 );
+
+unitTest({ perms: { net: true } }, function netTcpListenCloseSignal(): void {
+  const ac = new AbortController();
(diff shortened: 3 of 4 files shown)


extensions/net/01_net.js
```diff
@@ -207,7 +207,16 @@
...
-    return new Listener(res.rid, res.localAddr);
+    const listener = new Listener(res.rid, res.localAddr);
+    if (options?.signal) {
+      // TODO(benjamingr) there is memory leak potential here if the signal outlives the
+      // server significantly it can retain it. Ideally Deno would have internal weak
+      // event listener like Node does *or* it would have an event for the server closing.
+      options?.signal?.addEventListener("abort", () => listener.close(), {
+        once: true,
+      });
+    }
```
evanwashere, line 218:
```suggestion
    if (options?.signal) {
      const weak = new WeakRef(listener);
      options?.signal.addEventListener("abort", () => weak.deref()?.close(), {
        once: true,
      });
    }
```

extensions/net/02_tls.js
```diff
@@ -61,7 +62,14 @@
...
       alpnProtocols,
     });
-    return new TLSListener(res.rid, res.localAddr);
+    const listener = new TLSListener(res.rid, res.localAddr);
+    if (signal) {
+      // TODO(benjamingr) there is memory leak potential here. See comment in 01_net.js
+      signal.addEventListener("abort", () => listener.close(), {
+        once: true,
+      });
+    }
```
evanwashere, line 71:
```suggestion
    if (signal) {
      const weak = new WeakRef(listener);
      signal.addEventListener("abort", () => weak.deref()?.close(), {
        once: true,
      });
    }
```

evanwashere (changes requested):

extensions/net/01_net.js
benjamingr: That won't help - it still leaks the listener :) Happy to elaborate if that is unclear

evanwashere: it's true that function captures everything
if v8 is failing to optimize that without any invocations, we can move event callback creation outside scope

benjamingr: @evanwashere even if the capture was only of the listener this is still a memory leak since the listener itself is captured.

Weak event listeners need to be baked into the event target implementation itself for this to not leak (here is an example how I did it in Node.js https://github.com/nodejs/node/pull/36607 and I am happy to point to whatwg discussions).

That said - another (not bad) alternative is to expose a way to know when the listener is closed (and remove the handler then) but I didn't want to propose a change of the API of Deno.listen just for that :] 

(This leak is small and not very problematic to be fair)

evanwashere: listener is wrapped in weakref, so capture won't matter and problem will be fixed

```js
{
  const listener = {...};
  if (signal) signal.add('event', createListenerAbortCallback(new WeakRef(listener)));
}

function createListenerAbortCallback(weakref) {
  return () => weakref.deref()?.close();
}
```

benjamingr: but the signal would still have `() => weakref.deref()?.close();` in its listener array and that function would never gets cleared.

evanwashere: that's a fundamental problem with event listeners and needs a separate fix
the goal right now is to prevent signal from retaining hard reference to listener object and other temporary objects

benjamingr: yes, the fact it needs a separate fix is why I added a TODO for it :) The actual leak here is contrived, edge-casey and not that bad.

bnoordhuis: Bit of a drive-by comment but can't this be fixed with a WeakMap?
```js
const weakrefs = new WeakMap();

function listen({ hostname, ...options }) {
  // ...
  if (signal) {
    weakrefs.set(signal, new WeakRef(listener));
    signal.addEventListener("abort", handleAbort);
  }
  // ...
}

function handleAbort(e) {
  weakrefs.get(e.target)?.deref()?.close();
}
```
(Use of primordials omitted for brevity.)

benjamingr: @bnoordhuis that would still leak the `handleAbort` function reference.

bnoordhuis: That's not a problem though, is it? The solutions outlined earlier leak because they create new closures every time, whereas the `handleAbort()` closure is created only once (and at the top-level so it's going to stay around forever anyway.)

benjamingr: @bnoordhuis is the `handleAbort` reference still going to be in the listeners array? 

The "interesting" case here is a long-lived signal (imagine a signal owned by a controller that gets aborted on SIGINT for graceful shutdown) that gets registered often (imagine in this case you start a server on every request and do something).

Since this is `Deno.listen` it's not very likely - but that's the case we guarded against in Node

PR #11571, opened 2021-08-03
feat(unstable): clean environmental variables for subprocess

### LINKED ISSUE #3981: Deno.run clean environment
Deno.run inherits environment from the parent process. I'd love to see something like this:
```ts
Deno.run({
	args: ['printenv'],
	clean: true,
	env: {
		ENV_1: 'yes',
	},
})
```

This would mean only explicit environment variables would exist. Not unlike
```sh
$ env -i ENV_1=yes sh -c printenv
PWD=/path
ENV_1=yes
```
Doing the above is possible but not as elegant.

crowlKats:
Closes #3981

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2035,6 +2035,10 @@ declare namespace Deno {
    * Subprocess uses same working directory as parent process unless `opt.cwd`
    * is specified.
    *
+   * Environmental variables from parent process can be cleared using `opt.cleanEnv`.
+   * Doesn't guarantee that only `opt.env` variables are present,
+   * as the OS may set environmental variables for processes.
+   *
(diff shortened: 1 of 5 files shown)


ry:
Why not `Deno.env.clear()` to stick with the Map-like interface?

I'm a little concerned that this is immediately going to stable. Is it possible to make this unstable first?

bartlomieju:
> Why not Deno.env.clear() to stick with the Map-like interface?

This is not API for `Deno.env`, it's an option when starting a subprocess with `Deno.run()`. But I agree with @bnoordhuis suggestion that it might be better to name it `clearEnv`

crowlKats:
> Is it possible to make this unstable first?
sure, i am quite in favour of that.

bartlomieju (approved):
LGTM, thanks @crowlKats!

PR #11573, opened 2021-08-03
chore: surface import map JSON parse error to user

dsherret:
I wrote an import map with a trailing comma and spent too long trying to figure out what was wrong. Would be better if the CLI told me what was wrong.

### DIFF
--- a/runtime/ops/tty.rs
+++ b/runtime/ops/tty.rs
@@ -228,7 +228,7 @@ fn op_isatty(
       {
         use winapi::um::consoleapi;
 
-        let handle = get_windows_handle(&std_file)?;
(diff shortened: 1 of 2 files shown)


runtime/ops/tty.rs
```diff
@@ -228,7 +228,7 @@ fn op_isatty(
       {
         use winapi::um::consoleapi;
 
-        let handle = get_windows_handle(&std_file)?;
+        let handle = get_windows_handle(std_file)?;
```
dsherret, line 231: Windows specific linting error.

bartlomieju (approved):

PR #11577, opened 2021-08-04
fix(cli/tests): remove unecessary void return types

### LINKED ISSUE #8832: Do not require specifying "void" and "Promise<void>" return values
We should strive for concise code. "void" and "Promoise<void>" return types are unnecessary boilerplate.

Example:
```typescript
Deno.test({
  name: "Worker with top-level-await",
  fn: async function (): Promise<void> {  // <--- HERE
    const promise = deferred();
    const worker = new Worker(
      new URL("./worker_with_top_level_await.ts", import.meta.url).href,
      { deno: true, type: "module" },
    );
    worker.onmessage = (e): void => {    // <--- HERE
      console.log("received from worker", e.data);
      worker.postMessage("from main");
      promise.resolve();
    };
    await promise;
    worker.terminate();
  },
});
```

crowlKats:
Fixes #8832

### DIFF
--- a/cli/tests/004_set_timeout.ts
+++ b/cli/tests/004_set_timeout.ts
@@ -1,10 +1,10 @@
-setTimeout((): void => {
+setTimeout(() => {
   console.log("World");
 }, 10);
 
 console.log("Hello");
 
-const id = setTimeout((): void => {
+const id = setTimeout(() => {
   console.log("Not printed");
 }, 10000);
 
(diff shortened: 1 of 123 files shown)


bartlomieju:
@crowlKats this looks good to me, however tests are failing because some files are checked against lock files and those files got changed in this PR.

bartlomieju (approved):
LGTM

PR #11578, opened 2021-08-04
feat(runtime): allow URL for permissions

### LINKED ISSUE #11036: Deno.permissions APIs should accept URLs where appropriate
Currently `PermissionDescriptor`s accept only strings as paths.

As suggested in https://github.com/denoland/deno/issues/11035, we should add support for URLs where appropriate, in line with recent changes APIs such as:
- Deno.chdir
- Deno.realPath
- Deno.realPathSync
- Deno.rename
- Deno.renameSync
- Deno.symlink
- Deno.symlinkSync
- Deno.utime
- Deno.utimeSync

So this should work:
```ts
const url = new URL("some/path", import.meta.main);
const status = await Deno.permissions.query({ name: "read", path: url });
```

crowlKats:
Closes #11036
not only for `Deno.permissions`, but worker & test permissions as well.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2149,7 +2149,7 @@ declare namespace Deno {
      *      "github.com"
      *      "deno.land:8080"
      */
-    host?: string;
+    host?: string | URL;
   }
 
   export interface EnvPermissionDescriptor {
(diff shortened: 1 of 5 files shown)


bartlomieju:
Thanks @crowlKats this is very useful, could add some tests that exercise this test? After that we're good to land.

bartlomieju (approved):
LGTM, thanks Leo

lucacasonato:
I am not in favor for having this for net permissions as it would impede our ability to add full URL network permissions in the allow list down the line.

bartlomieju:
> I am not in favor for having this for net permissions as it would impede our ability to add full URL network permissions in the allow list down the line.

How does that conflict with accepting URLs in runtime APIs?

Issue #11016, opened 2021-06-17
Implement new two-tier subprocess API


lucacasonato:
```ts
declare namespace Deno {
  export interface ExecOptions {
    cmd: string[] | [URL, ...string[]];
    cwd?: string;
    env?: Record<string, string>;
    stdin?: "piped" | "inherit" | "null" | Deno.File; // default is "null"
    stdout?: "piped" | "inherit" | "null" | Deno.File; // default is "piped"
    stderr?: "piped" | "inherit" | "null" | Deno.File; // default is "piped"
  }

  export interface ExecOutput {
    status: ProcessStatus;
    stdout?: Uint8Array; // if it was set to "piped" in options, this is set
    stderr?: Uint8Array; // if it was set to "piped" in options, this is set
  }

  /**
   * Start a subprocess and wait until it completes. The subprocess will be
   * signaled to exit automatically if this process dies.
   */
  export function exec(options: ExecOptions): Promise<ExecOutput>;

  // NOTE: `Child` has multiple underlying resources:
  //          1. stdio - these have to always be closed manually
  //          2. process itself - this is closed automatically when `status`
  //             resolves (so after a `.kill` or after natural death).
  export class Child {
    readonly pid: number;

    readonly stdin?: Writer & Closer; // if it was set to "piped" in options, this is set
    readonly stdout?: Reader & Closer; // if it was set to "piped" in options, this is set
    readonly stderr?: Reader & Closer; // if it was set to "piped" in options, this is set

    readonly status: Promise<ProcessStatus>;

    kill(code: number): void;
  }

  export interface SpawnOptions extends ExecOptions {
    detached?: boolean;
  }

  /**
   * Start a subprocess in the background. The returned `Child` handle can be
   * used to communicate with the parent.
   */
  export function spawn(options: SpawnOptions): Child;
}
```

This new API provides a high level, super easy to use API (`Deno.exec`), and a low-level more intricate API (`Deno.spawn`). `Deno.exec` can be used for all cases where the user just wants to spawn a subprocess and wait for it to complete. For this API no resources need to be manually closed. `Deno.spawn` can be used for longer lived subprocess that you communicate with - they can also be detached, so outlive their parent.

The new semantics for backing resources are as follows:

- No resources need to be manually closed for `Deno.exec`
- For `Deno.spawn` "piped" stdio needs to be closed manually. The process itself is also a resource, but it is automatically cleaned up after the process exits.

The semantics for process lifecycle:

[...]

PR #11579, opened 2021-08-04
feat(process) support AbortSignal in run

### LINKED ISSUE #11574: [Feature request]: support AbortSignal for subprocess
```ts
/**
 * subprocess_simple.ts
 */

const ac = new AbortController();

// create subprocess
const p = Deno.run({
  cmd: ["do", "something"],
  signal: ac
});

(async () => {
  const shouldCancel = await doStaff()

  shouldCancel && ac.abort();
})()

// await its completion
await p.status();
```

benjamingr:
Support AbortSignal in `Deno.run`

This continues the work in #10943 and adds AbortSignal support to `Deno.run`. Closes #11574

cc @lucacasonato @bartlomieju

### DIFF
--- a/runtime/js/40_process.js
+++ b/runtime/js/40_process.js
@@ -104,6 +104,7 @@
     stdout = "inherit",
     stderr = "inherit",
     stdin = "inherit",
+    signal,
   }) {
     if (cmd[0] != null) {
       cmd[0] = pathFromURL(cmd[0]);
@@ -119,7 +120,17 @@
       stdoutRid: isRid(stdout) ? stdout : 0,
       stderrRid: isRid(stderr) ? stderr : 0,
     });
-    return new Process(res);
+    const p = new Process(res);
+    if (signal) {
+      const stopper = () => {
+        p.kill(Deno.Signal.SIGTERM);
+      };
+      signal.addEventListener("abort", stopper);
+      p.status().then(() => {
+        signal.removeEventListener("abort", stopper);
+      });
+    }
+    return p;
   }
 
   window.__bootstrap.process = {
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2020,6 +2020,11 @@ declare namespace Deno {
     stdout?: "inherit" | "piped" | "null" | number;
     stderr?: "inherit" | "piped" | "null" | number;
     stdin?: "inherit" | "piped" | "null" | number;
+    /**
+     * An AbortSignal that allows closing the process using the corresponding
+     * AbortController by sending the process a SIGTERM signal.
+     */
+    signal?: AbortSignal;
   }
 
   /** Spawns new subprocess.  RunOptions must contain at a minimum the `opt.cmd`,
--- a/cli/tests/unit/process_test.ts
+++ b/cli/tests/unit/process_test.ts
@@ -510,3 +510,23 @@ unitTest({ perms: { run: true, read: true } }, function killFailed(): void {
 
   p.close();
 });
+
+unitTest(
+  { perms: { run: true, read: true } },
+  async function runAbortSignal(): Promise<void> {
+    const ac = new AbortController();
+    const p = Deno.run({
+      cmd: [
+        Deno.execPath(),
+        "eval",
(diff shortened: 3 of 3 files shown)


sant123 (approved):

runtime/js/40_process.js
```diff
@@ -119,7 +120,17 @@
       stdoutRid: isRid(stdout) ? stdout : 0,
       stderrRid: isRid(stderr) ? stderr : 0,
     });
-    return new Process(res);
+    const p = new Process(res);
+    if (signal) {
+      const stopper = () => {
+        p.kill(Deno.Signal.SIGTERM);
+      };
```
bartlomieju, line 127: `Process.kill` is an unstable API (https://doc.deno.land/builtin/unstable#Deno.kill) so we can't add it like this to stable `Deno.run`. Also I'm a bit worried about bifurcation that sending `SIGTERM` could be done with two different APIs.

We need to discuss this change more in-depth before merging.

benjamingr: @bartlomieju that is acceptable to me, I am in no rush to make these changes. The idea is to add a web-standard-capability way to cancel operations so that you use Deno APIs and web APIs similarly and can abort your sub-process, event handlers and fetch with the same `AbortSignal`.

I am happy to be part of that discussion though I am also fine with you having this discussion without me and just letting me know.

Note that adding this API to Node.js involved some discussion - the major points are:
 - Would users be confused by the parameter being called `signal` since in non-web-standards land signal means something else for processes? (Node decided to prefer web-standards)
 - Would users be confused by this sending a signal to the process? Which signal should it send? (In node that is configurable as killSignal and whatever stops the process elsewhere works here).

benjamingr: (Also the technical limitation (process.kill being unstable) seems like the smaller concern here and what API users should get seems like the much bigger one. I am happy to refactor this PR to use whatever machinary .kill is using directly in order to bypass the unstability of .kill it's not a big deal)

bartlomieju: @benjamingr sorry for late reply, I missed your comments. 

> that is acceptable to me, I am in no rush to make these changes. The idea is to add a web-standard-capability way to cancel operations so that you use Deno APIs and web APIs similarly and can abort your sub-process, event handlers and fetch with the same AbortSignal.

This is a good argument.

I agree that `signal` might be a confusing name, my first thought when I saw it was OS signal - let's bikeshed on this.

> (Also the technical limitation (process.kill being unstable) seems like the smaller concern here and what API users should get seems like the much bigger one. I am happy to refactor this PR to use whatever machinary .kill is using directly in order to bypass the unstability of .kill it's not a big deal)

Actually looking at it I don't think there are any blockers to stabilize this API so that would allows us to move forward with this PR.

benjamingr: The issue with the name `signal` is that the web standards require:

```
Any web platform API using promises to represent operations that can be aborted must adhere to the following:

 - Accept AbortSignal objects through a signal dictionary member.
 - Convey that the operation got aborted by rejecting the promise with an "AbortError" DOMException.
 - Reject immediately if the AbortSignal's aborted flag is already set, otherwise:
 - Use the abort algorithms mechanism to observe changes to the AbortSignal object and do so in a manner that does not lead to clashes with other observers.
```

Note that this _isn't_ actually a web API so Deno doesn't _have_ to use the same web standard conventions but it is probably preferable to do so if web-familiarity and compatibility is a goal. WDYT?

bartlomieju: To be honest I'm not sure, it's quite confusing, I will confer with other team members on this one.

benjamingr: Thanks, let me know :)

crowlKats:
It might make more sense to not add any new features to `Deno.run` since there is the plan to soon deprecate it with the new subprocess API, and just directly implement it there

sant123:
See https://github.com/denoland/deno/issues/11016

benjamingr:
I think the CLA bot is broken?

bartlomieju:
> I think the CLA bot is broken?

Ryan changed email assigned to CLA bot so it's requesting signing from everyone again 😅

PR #11581, opened 2021-08-04
refactor: make `ParsedModule` implement `Sync`

dsherret:
This PR is the first two points in #11345.

Basically, this change does two main things:

1. Removes the use of swc's `SourceMap` and `SourceFile` in favour of the new `SourceFileInfo`, which implements `Sync`.
2. Moves out comments from `SingleThreadedComments` to a new immutable `MultiThreadedComments`.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -82,7 +82,7 @@ serde = { version = "1.0.126", features = ["derive"] }
 shell-escape = "0.1.5"
 sourcemap = "6.0.1"
 swc_bundler = "0.46.0"
-swc_common = { version = "0.11.0", features = ["sourcemap"] }
+swc_common = { version = "0.11.4", features = ["sourcemap"] }
 swc_ecmascript = { version = "0.46.0", features = ["codegen", "dep_graph", "parser", "proposal", "react", "transforms", "typescript", "visit"] }
 tempfile = "3.2.0"
 termcolor = "1.1.2"
--- a/cli/ast/bundle_hook.rs
+++ b/cli/ast/bundle_hook.rs
@@ -1,0 +1,53 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+use deno_core::error::AnyError;
+
+pub struct BundleHook;
+
+impl swc_bundler::Hook for BundleHook {
+  fn get_import_meta_props(
+    &self,
+    span: swc_common::Span,
+    module_record: &swc_bundler::ModuleRecord,
+  ) -> Result<Vec<swc_ecmascript::ast::KeyValueProp>, AnyError> {
+    use swc_ecmascript::ast;
+
+    // we use custom file names, and swc "wraps" these in `<` and `>` so, we
+    // want to strip those back out.
+    let mut value = module_record.file_name.to_string();
+    value.pop();
+    value.remove(0);
+
+    Ok(vec![
+      ast::KeyValueProp {
+        key: ast::PropName::Ident(ast::Ident::new("url".into(), span)),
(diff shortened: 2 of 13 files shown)


cli/ast/bundle_hook.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
dsherret, line 1: This file is a straight extract out of `mod.ts`. I did a slight bit of cleanup in this PR as well.

cli/ast/mod.rs
```diff
@@ -43,13 +46,20 @@ use swc_ecmascript::transforms::react;
...
+pub use bundle_hook::BundleHook;
+use comments::MultiThreadedComments;
+use source_file_info::SourceFileInfo;
+
 static TARGET: JscTarget = JscTarget::Es2020;
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Location {
-  pub filename: String,
+  pub specifier: String,
```
dsherret, line 62: Renamed to `specifier` to align with elsewhere.

cli/ast/mod.rs
```diff
@@ -452,21 +431,12 @@ fn flatten_comments(
   comments.into_iter().flat_map(|el| el.1)
 }
 
-pub fn lex(
-  specifier: &str,
```
dsherret, line 456: This parameter is not necessary.

cli/ast/source_file_info.rs
```diff
@@ -0,0 +1,130 @@
...
+    } else if pos == line_end_pos {
+      line_text.chars().count()
+    } else {
+      line_text
+        .char_indices()
+        .position(|(c_pos, _)| line_start_pos + c_pos >= pos)
+        .unwrap()
+    }
+  }
+}
```
dsherret, line 78: This function is probably not as elegant as it could be.

cli/Cargo.toml
```diff
@@ -82,7 +82,7 @@ serde = { version = "1.0.126", features = ["derive"] }
 shell-escape = "0.1.5"
 sourcemap = "6.0.1"
 swc_bundler = "0.46.0"
-swc_common = { version = "0.11.0", features = ["sourcemap"] }
+swc_common = { version = "0.11.4", features = ["sourcemap"] }
```
kitsonk, line 85: Is the version bump required for this PR? Normally we bump all the other uses of swc_common across the crates before a release.

cli/ast/comments.rs
```diff
@@ -0,0 +1,102 @@
...
+use std::cell::RefCell;
+use std::rc::Rc;
+use std::sync::Arc;
+use swc_common::comments::Comment;
+use swc_common::comments::Comments;
+use swc_common::comments::SingleThreadedComments;
+use swc_common::comments::SingleThreadedCommentsMapInner;
+use swc_common::BytePos;
+
+#[derive(Clone, Debug)]
```
kitsonk, line 12: Not that it is a big deal, but future generations might benefit from a little explanation of why we are doing this?

kitsonk (approved):
LGTM but a couple of brief comments.

Also, I think it is more of `refactor:` than a `chore:` (https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716)

cli/Cargo.toml
dsherret: Yup. Need it for some changes I did in swc_common.

dsherret:
@kitsonk thanks! Fixed. Sometimes I forget to do `refactor` instead of `chore`.

Issue #7870, opened 2020-10-07
deno fmt doesn't handle jsx as well as prettier


lucacasonato:
Here are some examples - top is denofmt, bottom is prettier:

### 1

```tsx
    <div
      class="bg-gradient-to-tr from-teal-500 to-teal-700 h-14 my-6"
    >
      <div
        class="flex justify-between items-center max-w-screen-lg mx-auto px-4 sm:px-6 md:px-8"
      >
        <a class="flex items-center" href="/dashboard" title="Go to dasboard">
          <img src="/img/logo.svg" alt="logo" class="w-20 h-20 -my-3" />
          <span class="ml-3 text-white text-3xl font-extrabold tracking-wide">
            foo site
          </span>
        </a>
        {user && <a href="/account" title="Go to account page">
          <ProfileImage class="w-10 h-10" />
        </a>}
      </div>
    </div>
```

```tsx
    <div class="bg-gradient-to-tr from-teal-500 to-teal-700 h-14 my-6">
      <div class="flex justify-between items-center max-w-screen-lg mx-auto px-4 sm:px-6 md:px-8">
        <a class="flex items-center" href="/dashboard" title="Go to dasboard">
          <img src="/img/logo.svg" alt="logo" class="w-20 h-20 -my-3" />
          <span class="ml-3 text-white text-3xl font-extrabold tracking-wide">
            foo site
          </span>
        </a>
        {user && (
          <a href="/account" title="Go to account page">
            <ProfileImage class="w-10 h-10" />
          </a>
        )}
      </div>
    </div>
```

### 2

```tsx
function App(props: AppProps) {
  return <AuthProvider>
    {props.children}
  </AuthProvider>;
}
```

```tsx
function App(props: AppProps) {
  return <AuthProvider>{props.children}</AuthProvider>;
}
```

### 3

```tsx
    <html lang="en">
      <head>
        <meta charSet="UTF-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
      </head>
      <body>
        {props.children}
      </body>
    </html>
```

```tsx
    <html lang="en">
      <head>
        <meta charSet="UTF-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
      </head>
      <body>{props.children}</body>
    </html>
```

### 4

```tsx
    <Page>
      {user
        ? <>
          <div class="flex">
            <ProfileImage class="w-16 h-16" />
            <div class="ml-4 flex flex-col justify-center">
[...]

PR #11582, opened 2021-08-04
feat(fmt): format top-level JSX elements/fragments with parens when multi-line

dsherret:
See https://github.com/dprint/dprint-plugin-typescript/issues/133

This is one part of #7870


satyarohith (approved):
LGTM. Nice to have this fixed! 🚀

Issue #11580, opened 2021-08-04
httpCookieConcatenation test flaky


ry:
```
2021-08-03T20:17:23.4372050Z test httpCookieConcatenation ... [0m[32mok[0m [0m[38;5;8m(11ms)[0m
2021-08-03T20:17:23.4475680Z [0m[1m[31merror[0m: Uncaught (in promise) AssertionError
2021-08-03T20:17:23.4476280Z 
2021-08-03T20:17:23.4478080Z test result: [0m[31mFAILED[0m. 817 passed; 0 failed; 18 ignored; 0 measured; 0 filtered out [0m[38;5;8m(48557ms)[0m
2021-08-03T20:17:23.4520120Z     throw new AssertionError(msg);
2021-08-03T20:17:23.4524050Z 
2021-08-03T20:17:23.4570240Z [0m[31m          ^[0m
2021-08-03T20:17:23.4572590Z     at [0m[1m[3massert[0m ([0m[36mfile:///Users/runner/work/***/***/test_util/std/testing/asserts.ts[0m:[0m[33m214[0m:[0m[33m11[0m)
2021-08-03T20:17:23.4574900Z     at [0m[1m[3miteratorReq[0m ([0m[36mfile:///Users/runner/work/***/***/cli/tests/unit/net_test.ts[0m:[0m[33m571[0m:[0m[33m9[0m)
2021-08-03T20:17:23.4839190Z test integration::js_unit_tests ... FAILED
```
https://github.com/denoland/deno/runs/3234457836

dsherret:
This is actually the `netHangsOnClose` test `file:///Users/runner/work/***/***/cli/tests/unit/net_test.ts:571:9`

There is a promise that is not awaited in there too. I'm looking into it.

bnoordhuis:
I ran into this twice in a row today, always on the macos-10.15 buildbot.

PR #11585, opened 2021-08-04
chore: refactor `netHangsOnClose` test to not use `deferred`

dsherret:
This does not solve #11580, but it makes it so that when an assertion fails it will not happen in an unawaited promise.

### DIFF
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -573,7 +573,7 @@ unitTest(
 
     const addr = { hostname: "127.0.0.1", port: 3500 };
     const listener = Deno.listen(addr);
-    const listnerPromise = iteratorReq(listener);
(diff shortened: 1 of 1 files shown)


cli/tests/unit/net_test.ts
```diff
@@ -551,7 +551,6 @@ unitTest(
   },
   async function netHangsOnClose() {
     let acceptedConn: Deno.Conn;
-    const resolvable = deferred();
```
dsherret, line 554: We should use `deferred` sparingly in the tests and only where the code does not allow awaiting a promise. The problem here is any assertion failures before `resolvable.resolve()` is called will occur in an unawaited promise... I believe that may lead to the test output being harder to understand.

kitsonk (approved):
LGTM

PR #11586, opened 2021-08-05
feat(unstable): allow specifing gid and uid for subprocess

### LINKED ISSUE #5506: Set gid/uid for subprocess
I am not sure that i should create a new Issue (may be linked with #5501)

I am working on a program that watch over some processes but i would like to limit the rights of the child process before execution.

it is possible with nodejs to set an uid and/or a gid so then, just before the process is execve the parent process call setuid and/or setgid to drop privileges of the child.

[node way](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback)

crowlKats:
Closes #5506


cli/dts/lib.deno.ns.d.ts
```diff
@@ -2041,6 +2041,9 @@ declare namespace Deno {
    * Environmental variables for subprocess can be specified using `opt.env`
    * mapping.
    *
+   * On unix `opt.gid` and `opt.uid` can be used to define the gid and uid for
+   * the subprocess.
```
bnoordhuis, line 2045: I don't think it's going to be clear to most users what this means. Can I suggest copying the description from libstd's `CommandExt`?
```
Sets the child process’s user ID. This translates to a setuid call in the child process. Failure in the setuid call will cause the spawn to fail.
```
And:
```
Similar to uid, but sets the group ID of the child process. This has the same semantics as the uid field.
```
I'd argue changing uid/gid without being able to drop group membership is insecure (a mistake I made in libuv) but `.groups()` is experimental in libstd and tokio doesn't support it at all.

runtime/ops/process.rs
```diff
@@ -123,6 +125,17 @@ fn op_run(
...
 
+  #[cfg(unix)]
+  if let Some(gid) = run_args.gid {
+    super::check_unstable(state, "Deno.run.gid");
+    c.gid(gid);
+  }
+  #[cfg(unix)]
+  if let Some(uid) = run_args.uid {
+    super::check_unstable(state, "Deno.run.uid");
+    c.gid(uid);
```
lucacasonato, line 136: `c.uid` instead of `c.gid`?

cli/dts/lib.deno.ns.d.ts
bartlomieju: @crowlKats please update doc with Ben's suggestion.

> 


I'd argue changing uid/gid without being able to drop group membership is insecure (a mistake I made in libuv) but .groups() is experimental in libstd and tokio doesn't support it at all.

@bnoordhuis what would you suggest in this case?

bnoordhuis: Either hold off on merging until Tokio grows the right APIs or add a really big warning to the documentation.

The problem with a warning is that it's not actionable. There's no way to detect an insecure situation and act on it.

bartlomieju: Thanks, as you point out the second option is not actionable. Let's hold off merging, and open an feature request to `tokio`.

bnoordhuis: I looked at how libstd implements `Command::spawn()` (which tokio wraps) and it calls `setgroups(0, null())` to drop group membership (barring bug https://github.com/rust-lang/rust/issues/88716) so, on second thought, this is probably good to merge.

If you wanted to be extra secure, you could add a Unix-ony `pre_exec` hook that calls `setgroups(0, null())`.

bartlomieju: @crowlKats could you add that call to `pre_exec` that Ben pointed out?

crowlKats: @bartlomieju done

bartlomieju (approved):
LGTM

Issue #10183, opened 2021-04-14
Security prompt by default (instead of throw)


jsejcksn:
Opening per @bartlomieju's request in https://github.com/denoland/deno/issues/3811#issuecomment-819445206

It appears that only `--prompt` shipped in `1.9.0`, and that the team is considering making it the default option based on feedback:

> Please let us know if `--prompt` is useful for you. We are considering turning it on by default in a future release.
↳ https://deno.com/blog/v1.9#interactive-permission-prompt

I haven't found any further discussion/notice about potential implementation of `--no-prompt`, but both provide different behavior from what's there now, and unique value to the user in different cases:

`--prompt`:
The user wants to run a script without considering permissions in advance, and interactively respond to permission requests.

`--no-prompt`:
The user wants to run a script, providing only permissions determined in advance, and expect the program to exit (either successfully or due to an error) without interaction arising from unmet permission requirements.

`--no-prompt` seems especially important in environments where interaction is not possible so that the process doesn't just hang.

I imagine the implementation would be similar to this:

1. The `state` of every permission's [status](https://doc.deno.land/builtin/stable#Deno.PermissionStatus) is set to `"denied"` by default (instead of `"prompt"`).
2. For every permission granted as a CLI argument to the relevant runtime command (`run`/`test`/etc)., those permissions' statuses will be changed to `"granted"` before the script is executed.

hayd:
"no prompt" is the default behavior and should remain so. Having --prompt and --no-prompt is nonsensical.

Personally i am strongly against making --prompt the default behavior (again), and this seemed to be the consensus in #3811.

deno should remain "secure by default".

jsejcksn:
> "no prompt" is the default behavior

@hayd This is not the case. Currently (in Deno `1.9.0`), by default, all permissions' status states are set to `prompt` when a script is run unless they are granted using flags on the command line:

> A permission state can be either "granted", "prompt" or "denied". Permissions which have been granted from the CLI will query to `{ state: "granted" }`. Those which have not been granted query to `{ state: "prompt" }` by default, while `{ state: "denied" }` reserved for those which have been explicitly refused. This will come up in [Request permissions](https://deno.land/manual@v1.9.0/runtime/permission_apis#request-permissions).
> ↳ https://deno.land/manual@v1.9.0/runtime/permission_apis#permission-states

If my implementation description is unclear, please let me know so I can revise it. I want to make sure it is not miscommunicated.

hayd:
[...]

PR #11588, opened 2021-08-05
feat(runtime): Making `--prompt` the default behavior instead of throwing error

getspooky:
Like browser permission experience, it would be more useful to include permission prompt by default instead of throwing error.

Please See Discussion: https://github.com/denoland/deno/issues/10183

Example: 

```ts
const json = fetch("https://api.github.com/users/denoland");

json
  .then((response) => {
    return response.json();
  })
  .then((jsonData) => {
    console.log(jsonData);
  });
```

```sh
deno run example.ts
```
Result: 

```sh
⚠️  ️Deno requests net access to "api.github.com". Allow? [y/n (y = yes allow, n = no deny)] 
```

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -71,14 +71,13 @@ impl PermissionState {
     self,
     name: &str,
     info: Option<&str>,
-    prompt: bool,
   ) -> (Result<(), AnyError>, bool) {
     match self {
       PermissionState::Granted => {
         Self::log_perm_access(name, info);
         (Ok(()), false)
       }
-      PermissionState::Prompt if prompt => {
+      _ => {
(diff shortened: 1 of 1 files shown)


crowlKats:
Implementation is completely wrong: the goal is to make it the default behavior, aka having a no-prompt flag that allows it be disabled. Also this most definitively should not be done pre 2.0

bartlomieju:
Thanks for opening the PR @getspooky but it's hasn't been yet agreed upon that we want to proceed with `--prompt` by default. Additionally there's quite a lot of auxiliary logic that will need to be implemented for this feature (eg. detecting if running in TTY). I'm going close it now, and let's drive discussion in #10183 home before implementing a PR.

PR #11590, opened 2021-08-05
feat(repl): add --eval flag for evaluating code when the repl starts

### LINKED ISSUE #3173: Support for deno eval interactive option
Deno should provide an equivalent functionality of [`node -e <CODE> -i`](https://nodejs.org/api/cli.html#cli_i_interactive) that allows the user to evaluate code through the Deno CLI but also attach REPL in the context of the given code. 

**Expected behavior**
```bash
deno eval "const hello = 'world'" -i
# Automatically starts REPL
> console.log(hello)
world
undefined
>
```

dsherret:
Closes #3173

At the moment, this only adds support for an `--eval <code>` flag to provide some code to execute on startup.

For example:

```bash
deno repl --eval 'import * as hash from "https://deno.land/std@0.77.0/hash#stdhash"'
```

1. On success, the result of running the script will not be outputted, but still assigned to _.
2. On thrown error, the error will be displayed to the user and assigned to _error (unless it's a parse error). The REPL will not fail to start in order to not block the user allowing them to ignore any setup errors and just use the REPL.

We will consider adding support for URL and file paths in the future, but for now this is a first step.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -91,7 +91,9 @@ pub enum DenoSubcommand {
     rules: bool,
     json: bool,
   },
-  Repl,
+  Repl {
+    eval: Option<String>,
+  },
   Run {
     script: String,
   },
@@ -119,7 +121,7 @@ pub enum DenoSubcommand {
 
 impl Default for DenoSubcommand {
   fn default() -> DenoSubcommand {
-    DenoSubcommand::Repl
+    DenoSubcommand::Repl { eval: None }
   }
 }
 
@@ -948,6 +950,13 @@ Ignore linting a file by adding an ignore comment at the top of the file:
 fn repl_subcommand<'a, 'b>() -> App<'a, 'b> {
   runtime_args(SubCommand::with_name("repl"), false, true)
     .about("Read Eval Print Loop")
+    .arg(
+      Arg::with_name("eval")
+        .long("eval")
+        .help("Evaluates the provided code when the REPL starts.")
+        .takes_value(true)
+        .value_name("code"),
+    )
 }
 
 fn run_subcommand<'a, 'b>() -> App<'a, 'b> {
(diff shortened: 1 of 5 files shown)


cli/flags.rs
```diff
@@ -119,7 +121,7 @@ pub enum DenoSubcommand {
 
 impl Default for DenoSubcommand {
   fn default() -> DenoSubcommand {
-    DenoSubcommand::Repl
+    DenoSubcommand::Repl { eval: None }
```
dsherret, line 124: Note that `deno --eval "console.log(5);"` does not work with this PR. The `repl` subcommand is required. Should that change?

cli/flags.rs
```diff
@@ -948,6 +950,13 @@ Ignore linting a file by adding an ignore comment at the top of the file:
 fn repl_subcommand<'a, 'b>() -> App<'a, 'b> {
   runtime_args(SubCommand::with_name("repl"), false, true)
     .about("Read Eval Print Loop")
+    .arg(
+      Arg::with_name("eval")
+        .long("eval")
+        .help("Evaluates the provided code when the REPL starts.")
+        .takes_value(true)
+        .value_name("code"),
```
dsherret, line 958: I've noticed some parts of the code require an equals sign. Should this? `deno repl --eval="console.log(5);"`? I'm not sure when to use the equals sign and when not to.

cli/tools/repl.rs
```diff
@@ -721,6 +736,14 @@ pub async fn run(
   let history_file_path = program_state.dir.root.join("deno_history.txt");
   let editor = ReplEditor::new(helper, history_file_path);
 
+  if let Some(eval) = maybe_eval {
+    let output = repl_session.evaluate_line_and_get_output(&eval).await?;
+    // only output errors
+    if let EvaluationOutput::Error(error_text) = output {
+      println!("error in --eval flag. {}", error_text);
+    }
+  }
```
dsherret, line 745: It was super easy to implement this thanks to the recent refactorings in this code 😀

cli/flags.rs
ry: I also think it's not worth it

cli/flags.rs
ry: I think it shouldn't require equals. The equals sign is required in some places to disambiguate. I don't remember the exact situation off hand, but we should attempt to remove the equals-sign-requirement if the ambiguity has evaporated.

ry (approved):
LGTM

bartlomieju (approved):
LGTM too

PR #11592, opened 2021-08-05
feat(repl): strip exports in the REPL

dsherret:
Just thought about this a bit more and I think it's useful. It does the following transforms in the REPL:

```ts
export class Test {}
----
class Test {}
```

```ts
export default class Test {}
----
class Test {}
```

```ts
export default class {}
----
;
```

```ts
export default 5;
----
5;
```

```ts
export * from "./test.ts";
----
await import("./test.ts");
```

```ts
export { test } from "./test.ts";
----
await import("./test.ts");
```

```ts
export { test };
----
;
```

This will help with usability and allow people to more easily copy and paste code into the repl.

### DIFF
--- a/cli/ast/mod.rs
+++ b/cli/ast/mod.rs
@@ -314,6 +314,7 @@ impl ParsedModule {
     let mut passes = chain!(
       Optional::new(jsx_pass, options.transform_jsx),
       Optional::new(transforms::DownlevelImportsFolder, options.repl_imports),
+      Optional::new(transforms::StripExportsFolder, options.repl_imports),
       proposals::decorators::decorators(proposals::decorators::Config {
         legacy: true,
         emit_metadata: options.emit_metadata
--- a/cli/ast/transforms.rs
+++ b/cli/ast/transforms.rs
@@ -21,33 +21,12 @@ impl Fold for DownlevelImportsFolder {
       ModuleItem::ModuleDecl(ModuleDecl::Import(import_decl)) => {
         // Handle type only imports
         if import_decl.type_only {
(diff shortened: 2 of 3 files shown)


bartlomieju (approved):
This looks good to me and it seems quite useful. Please add some information to https://github.com/denoland/manual/blob/main/tools/repl.md regarding all the transforms that happen in REPL.

dsherret:
@bartlomieju I'm not sure it's useful to add this information to the manual. I don't think most users will even know what transforms are and it seems strange to explain that to them. I think this is just something that if users know about the details of they will be happy it works and for users who don't know, this will stop slowing them down and making them confused.

caspervonb:
Not particularly in favour of all this preprocessing, doesn't seem to have any practical real world use case.

dsherret:
@caspervonb the practical use case is copy and pasting in code that will now just work. For example it’s annoying to include a declaration with an export keyword and have the repl error instead of understanding what’s going on and handling it. This is very similar to the import declarations support. I’m not sure what bloated repl we’re crawling towards now that both imports and exports will be supported with this change.

caspervonb:
> @caspervonb the practical use case is copy and pasting in code that will now just work. For example it’s annoying to include a declaration with an export keyword and have the repl error instead of understanding what’s going on and handling it. This is very similar to the import declarations support. I’m not sure what bloated repl we’re crawling towards now that both imports and exports will be supported with this change.

Wasn't really for imports either, but since we landed imports _it does make sense_ to be symmetrical with exports so I'm not attempting to block it am just a bit concerned that we'll add more and more transforms for every perceived inconvenience.

caspervonb (approved):

ry (approved):
LGTM

bartlomieju (approved):

PR #11598, opened 2021-08-06
feat: support client certificates for connectTls

ry:
This is @erik's #9426 rebased.

Still needs some work, but close.


test_util/src/lib.rs
```diff
@@ -338,6 +360,55 @@ async fn run_wss_server(addr: &SocketAddr) {
...
+  let cert_file = "cli/tests/tls/localhost.crt";
+  let key_file = "cli/tests/tls/localhost.key";
+  let ca_cert_file = "cli/tests/tls/RootCA.pem";
+  let tls_config = get_tls_config(cert_file, key_file, ca_cert_file)
+    .await
+    .unwrap();
+  let tls_acceptor = TlsAcceptor::from(tls_config);
+
+  // This bit of nastiness is to support the case when localhost resolves to
+  // ::1 on github actions and 127.0.0.1 locally.
```
bnoordhuis, line 378: This is necessary because cli/tests/tls/localhost.crt is made out to "localhost", right? I'd make that explicit in the comment, it's not self-evident.

I'd bind to all addresses instead of just the first one, otherwise you might run into the issue where the server picks 127.0.0.1 but the client ::1, or vice versa. Example code:
```rust
  //use futures::FutureExt;
  //use futures::StreamExt;

  let accept = |listener: tokio::net::TcpListener| {
    async {
      let result = listener.accept().await;
      Some((result, listener))
    }
    .boxed_local()
  };

  let host_and_port = &format!("localhost:{}", TLS_CLIENT_AUTH_PORT);

  let listeners = tokio::net::lookup_host(host_and_port)
    .await
    .expect(host_and_port)
    .inspect(|address| println!("{} -> {}", host_and_port, address))
    .map(tokio::net::TcpListener::bind)
    .collect::<futures::stream::FuturesUnordered<_>>()
    .collect::<Vec<_>>()
    .await
    .into_iter()
    .filter_map(Result::ok) // TODO(bnoordhuis) Log errors.
    .map(|listener| futures::stream::unfold(listener, accept))
    .collect::<Vec<_>>();

  let mut listeners = futures::stream::select_all(listeners);

  while let Some(Ok((stream, _addr))) = listeners.next().await {
    // ...
  }
```

bnoordhuis (approved):

test_util/src/lib.rs
lucacasonato: Added. Thanks!

extensions/net/ops_tls.rs
```diff
@@ -921,7 +956,10 @@ where
       alpn_protocols.into_iter().map(|s| s.into_bytes()).collect();
   }
   tls_config
-    .set_single_cert(load_certs(cert_file)?, load_keys(key_file)?.remove(0))
+    .set_single_cert(
+      load_certs_from_file(cert_file)?,
+      load_private_keys_from_file(key_file)?.remove(0),
```
lucacasonato, line 961: This will panic if index is out of bounds. Needs to be checked.

extensions/net/ops_tls.rs
```diff
@@ -755,6 +759,14 @@ where
   };
   let port = args.port;
   let cert_file = args.cert_file.as_deref();
+
+  if args.cert_chain.is_some() {
+    super::check_unstable2(&state, "ConnecTlsOptions.certChain");
```
danopia, line 764:
```suggestion
    super::check_unstable2(&state, "ConnectTlsOptions.certChain");
```

extensions/net/ops_tls.rs
```diff
@@ -755,6 +759,14 @@ where
...
+  }
+  if args.private_key.is_some() {
+    super::check_unstable2(&state, "ConnecTlsOptions.privateKey");
```
danopia, line 767:
```suggestion
    super::check_unstable2(&state, "ConnectTlsOptions.privateKey");
```

danopia (commented):
minor spelling

lucacasonato (changes requested):

bartlomieju (approved):

PR #11600, opened 2021-08-06
revert: allow URL for permissions

bartlomieju:
Revert changes to "net" permissions in regards to handling URLs.

Introduced in https://github.com/denoland/deno/pull/11578

Reverting due to @lucacasonato comment: https://github.com/denoland/deno/pull/11578#issuecomment-894315667

### DIFF
--- a/runtime/js/11_workers.js
+++ b/runtime/js/11_workers.js
@@ -1,4 +1,4 @@
-// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+  // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 "use strict";
 
 ((window) => {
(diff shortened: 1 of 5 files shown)


runtime/js/11_workers.js
```diff
@@ -93,9 +93,6 @@
     } else if (ArrayIsArray(value)) {
       value = ArrayPrototypeMap(value, (route) => {
         if (route instanceof URL) {
-          if (permission === "net") {
-            route = route.host;
```
lucacasonato, line 97: Maybe throw instead? (Like env)

bartlomieju: Sure

nayeemrmn (approved):
The intent of #11036 was to allow file URLs for fs permission allowlists, net shouldn't have been included because that filters by host which can't sensibly be represented as URLs. I'm not sure about ever doing "full URL network permissions" down the line but either way 👍

lucacasonato (approved):

PR #11601, opened 2021-08-06
chore: move test files to testdata directory

dsherret:
This change moves all test files from the `./tests` folder to `./tests/testdata` and makes the cwd `./tests/testsdata` for all these tests.

image


cli/tests/unit/README.md
```diff
@@ -26,14 +26,14 @@ unitTest(
 
 ## Running tests
 
-There are three ways to run `unit_test_runner.ts`:
+There are two ways to run `unit_test_runner.ts`:
 
 ```sh
 # Run all tests.
-target/debug/deno test --allow-all --unstable cli/tests/unit/
+(cd cli/tests/testdata && ../../target/debug/deno test --allow-all --unstable ../unit)
```
dsherret, line 33: Actually, I think it is unnecessary for the unit tests to change their CWD. I will revert this.

bartlomieju: It's preferable that unit tests work independently of CWD

dsherret: That would be nice. Maybe in a future change. This is reverted now and the PR is good to go.

bartlomieju (approved):

Issue #10276, opened 2021-04-20
docs.rs: deno_runtime fails to build


robfig:
The docs for `deno_runtime` fail to display:
https://docs.rs/crate/deno_runtime/0.12.0

Here's the failing build log:
https://docs.rs/crate/deno_runtime/0.12.0/builds/379589

I'm afraid I have no idea how to fix it.
``
[INFO] [stderr]    Compiling deno_runtime v0.12.0 (/opt/rustwide/workdir)
[INFO] [stderr] error: linking with `cc` failed: exit status: 1
[INFO] [stderr]   |
[INFO] [stderr]   = note: "cc" "-m64" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-Wl,--as-needed" "-L" "/opt/rustwide/rustup
....
...
[INFO] [stderr]   = note: /usr/bin/ld: /opt/rustwide/target/debug/deps/libdeno_core-00ca7298c52b25d2.rlib(deno_core-00ca7298c52b25d2.deno_core.5zk4gxdo-cgu.12.rcgu.o): in function `rusty_v8::value_deserializer::ValueDeserializer::new':
[INFO] [stderr]           /opt/rustwide/cargo-home/registry/src/github.com-1ecc6299db9ec823/rusty_v8-0.22.1/src/value_deserializer.rs:385: undefined reference to `v8__ValueDeserializer__Delegate__CONSTRUCT'
....
```

littledivy:
Seems intentional. Not sure why. https://github.com/denoland/deno/blob/572ec4a0a7ffd11c37fb48bbe7b8f86e9693d253/runtime/build.rs#L58-L61

robfig:
I don’t believe that is the cause. That build script creates a v8 snapshot
for use on startup. The error seems to be about being unable to link the v8
library with the rust library.

PR #11602, opened 2021-08-06
build: allow deno_runtime crate to build in docs.rs

bartlomieju:
Fixes https://github.com/denoland/deno/issues/10276


bartlomieju:
This is not probably the cause of build fails. Investigating...

dsherret (approved):
LGTM

ry (approved):
LGTM

bartlomieju:
For posterity: I tried creating an "empty" snapshot in `runtime/build.rs` due to the fact that `rusty_v8` doesn't build V8 on `docs.rs` (cause docs.rs doesn't allow to access net). However it didn't help and I still get a lot of linker errors (https://docs.rs/crate/deno_runtime/0.22.1/builds). I'm a bit at loss here as `deno_core` doesn't have a build script, nor does it tweak any settings and it builds just fine.

PR #11604, opened 2021-08-06
ci: testing how fast windows and mac takes to build in debug on the CI

dsherret:
Just taking a look at how long it takes them to build in debug.


dsherret:
Closing this. The results are that Windows is faster (ex. 24 mins instead of 28mins), but Mac is faster building (14mins vs 19mins), but then is very slow for running tests in debug (26mins vs 4mins).

Issue #11610, opened 2021-08-08
Some LSP clients don't support disabled code actions


David-Else:
```
deno 1.12.2+02c74fb (canary, x86_64-unknown-linux-gnu)
```
In Neovim 0.5 using the built in LSP and the standard Deno config from https://github.com/neovim/nvim-lspconfig/ , every line of TypeScript code triggers a code action alert, including spaces and comments. When you trigger the code action you get:
```
Code Actions:                                                                                                                                                                           
1. Extract function
2. Extract constant
3. Convert default export to named export
4. Convert named export to default export
5. Convert namespace import to named imports
Type number and <Enter> or click with the mouse (q or empty cancels):
```
and whatever you press it crashes.

Deno LSP worked perfectly until this update. Is this a bug in the Deno LSP, or do the settings in the Neovim config need updating?

https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/denols.lua

@kitagry Maybe you could have a look? Thanks!

kitsonk:
We are following the language server protocol: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction

In particular, we are sending the refactorings disabled, as per the protocol:

```ts
export interface CodeAction {
        /**
	 * Marks that the code action cannot currently be applied.
	 *
	 * Clients should follow the following guidelines regarding disabled code
	 * actions:
	 *
	 * - Disabled code actions are not shown in automatic lightbulbs code
	 *   action menus.
	 *
	 * - Disabled actions are shown as faded out in the code action menu when
	 *   the user request a more specific type of code action, such as
	 *   refactorings.
	 *
	 * - If the user has a keybinding that auto applies a code action and only
	 *   a disabled code actions are returned, the client should show the user
	 *   an error message with `reason` in the editor.
	 *
	 * @since 3.16.0
	 */
	disabled?: {

		/**
		 * Human readable description of why the code action is currently
		 * disabled.
		 *
		 * This is displayed in the code actions UI.
		 */
		reason: string;
	};
}
```

Though maybe the client doesn't support the 3.16 protocol.  Which really limits a few things.  I think we might want to check that clients have the support for disabled:

```ts
export interface CodeActionClientCapabilities {
	/**
	 * Whether the client supports resolving additional code action
	 * properties via a separate `codeAction/resolve` request.
	 *
	 * @since 3.16.0
	 */
	resolveSupport?: {
		/**
		 * The properties that a client can resolve lazily.
[...]

PR #11612, opened 2021-08-09
feat(lsp): support clients which do not support disabled code actions

kitsonk:
Closes: #11610

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -1298,28 +1298,18 @@ impl Inner {
         .map(CodeActionOrCommand::CodeAction),
     );
 
-    let response = if !all_actions.is_empty() {
-      if self.config.client_capabilities.code_action_disabled_support {
-        Some(all_actions)
-      } else {
-        // if the client does not support disabled code actions, then we should
-        // drop them
-        Some(
-          all_actions
-            .into_iter()
-            .filter(|ca| {
-              if let CodeActionOrCommand::CodeAction(ca) = ca {
-                ca.disabled.is_none()
-              } else {
-                true
-              }
-            })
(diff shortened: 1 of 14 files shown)


cli/lsp/language_server.rs
```diff
@@ -1299,7 +1299,24 @@ impl Inner {
...
+            .into_iter()
+            .filter(|ca| {
+              if let CodeActionOrCommand::CodeAction(ca) = ca {
+                ca.disabled.is_none()
+              } else {
+                true
+              }
+            })
+            .collect(),
+        )
```
bnoordhuis, line 1318: Seeing that `response` is `None` when `all_actions.is_empty() == true`, and that the filter here could result in an empty list, it should probably evaluate to `None` in that case?

I'd try to make it less branch-y, e.g.:
```rust
let code_action_disabled_support =
  self.config.client_capabilities.code_action_disabled_support;

let actions = all_actions
  .into_iter()
  .filter(|ca| {
    code_action_disabled_support
      || matches!(ca, CodeActionOrCommand::CodeAction(ca) if ca.disabled.is_none())
  })
  .collect();

let response = if actions.is_empty() {
  None
} else {
  Some(actions)
};

// or if you want to shave off a few lines:
let response = actions.is_empty().then(|| actions);
```
There won't be more than a handful of actions. Filtering should be essentially no cost.

dsherret (approved):
LGTM. Just Ben's question about whether it should return `None` when there are no actions.

cli/lsp/language_server.rs
kitsonk: Thanks for the feedback.  Makes sense!

Issue #11025, opened 2021-06-17
Support BYOB ReadableStream


crowlKats:

lucacasonato:
Seems to be stable in Chrome now. @kitsonk If you want to dive into this LMK, otherwise I can try tackle it soonish.

kitsonk:
Oh, I was just waiting to tackle this... 😏  maybe we can toss a coin for it?

bartlomieju:
@lucacasonato @kitsonk would it be reasonable to target this for 2.0?

Issue #10157, opened 2021-04-13
Max "op" size is 16384 which can be inefficient in user land


kitsonk:
Currently because of https://github.com/tokio-rs/tokio/issues/1976 the maximum "chunk" size that can be handled over an op is 16384, irrespective if this is efficient or not for the consumer.

For example when reading a large file off the disk (50 MiB for example) that would be some 3200 reads via `.read()` with 3200 promises to resolve.

Then consider serving that file over HTTP in 3200 16k chunks.  While the actual over the wire may need to be chunked up that is something that could/should be avoided in the isolate and handled in Rust ideally.

bnoordhuis:
What would you suggest we do if tokio limitations weren't a concern? Pick a bigger chunk size or something else?

kitsonk:
Ideally, we would do some testing and optimise it. It might be platform dependent, I don't know, but I can tell you 16k isn't great for files coming off disk and going out via HTTP via JavaScript. Some sort of bypass might work too. But think about the 3200 read ops that create promises that resolve that then get sent out as 3200 (at least) promises to a `ReadableStream` and whatever ops and promises to get it back out from the `Response`.

bnoordhuis:
There are so many confounding factors that an optimal chunk size won't exist. Whatever number comes out of a benchmark will be a product of that particular machine and point in time.

If the goal is to optimize file-to-file or file-to-socket transfers with no intermediate processing (e.g. zipping), then system calls like sendfile(2) and splice(2) are good solutions. Tokio integration is a hurdle but not intractable.

sam-watkins-sr:
I don't know about the promises, but Linux can do read pretty quickly. All languages I know use a fairly small buffer / chunk size when reading.

I timed a C program doing 16K reads just now, my Linux PC can do approx 500000 such reads per second, i.e. it can read 1GB file in 0.12 seconds, 16K at a time. Using a 1MB buffer is 33% faster, it takes 0.08 seconds.

Python and Perl use an 8K buffer.

sswam:
I don't know about the promises, but Linux can do read pretty quickly. All languages I know use a fairly small buffer / chunk size when reading.

I timed a C program doing 16K reads just now, my Linux PC can do approx 500000 such reads per second, i.e. it can read 1GB file in 0.10 seconds, 16K at a time. Using a 1MB buffer is 20% faster, it takes 0.08 seconds. A 4K buffer takes 0.15 seconds, 50% slower. It's very fast whichever way we do it.

[...]

Issue #7087, opened 2020-08-17
Zombie subprocesses 


asantos00:
Currently, when creating a subprocess using `Deno.run`, the process stays in "zombie mode" even after it is killed.

I've created a gist that reproduces this bug, and talked with @lucacasonato that confirmed it.

https://gist.github.com/asantos00/bdd4357cb91f18b368085eb44105ea97

caspervonb:
Example from the gist

```typescript
const command = ['deno', 'run', './script.ts'];

const p1 = Deno.run({ cmd: command })
const p2 = Deno.run({ cmd: command })
const p3 = Deno.run({ cmd: command })
const p4 = Deno.run({ cmd: command })

p1.kill(Deno.Signal.SIGKILL)
p2.kill(Deno.Signal.SIGKILL)
p3.kill(Deno.Signal.SIGKILL)
p4.kill(Deno.Signal.SIGKILL)

setInterval(() => {
  // just to keep it running so you can `ps aux | grep deno` to check the zombie processes
}, 1000);
```

bnoordhuis:
For background: UNIX expects you to reap zombies with `waitpid(2)` when you receive a `SIGCHLD` signal.

This would normally be an easy fix - set the `SIGCHLD` signal handler to `SIG_IGN` and let the kernel handle it - except....

`tokio::process::Child` installs its own `SIGCHLD` handler that records events into a queue. There is nothing polling the command objects and so events stay in the queue and zombies aren't reaped.

A crude hack that papers over the issue:
```diff
diff --git a/cli/ops/process.rs b/cli/ops/process.rs
index 60a6d5095..b259d9531 100644
--- a/cli/ops/process.rs
+++ b/cli/ops/process.rs
@@ -114,6 +114,13 @@ fn op_run(
 
   // Spawn the command.
   let mut child = c.spawn()?;
+
+  #[cfg(unix)]
+  unsafe { libc::signal(libc::SIGCHLD, libc::SIG_IGN); } // Reset signal handler.
+
   let pid = child.id();
 
   let stdin_rid = match child.stdin.take() {
```
It's not a good fix however because of the race window between the spawn and signal calls.

So far I haven't been able to come up with a better idea than poll child objects in the resource table at a fixed interval.

Soremwar:
I can reproduce this in latest canary. Linux reports all the child processes as zombies

nktpro:
[...]

Issue #2750, opened 2019-08-08
Redirect between stderr and stdout in RunOptions


nayeem-rahman:
Ref https://github.com/denoland/deno/pull/1828#issuecomment-467719574

Rather than adding `Process::combinedOutput()`, I propose trying to model `1>&2` and `2>&1` in `RunOptions`.

Currently, it is:
```typescript
export interface RunOptions {
    args: string[];
    cwd?: string;
    env?: {
      [key: string]: string;
    };
    stdout?: ProcessStdio | number;
    stderr?: ProcessStdio | number;
    stdin?: ProcessStdio | number;
  }
```

Considering `stdout` and `stderr` already don't have a meaningful unifying type, it wouldn't be totally unreasonable to do something like this:
```typescript
export interface RunOptions {
    args: string[];
    cwd?: string;
    env?: {
      [key: string]: string;
    };
    stdout?: ProcessStdio | "stderr" | number;
    stderr?: ProcessStdio | "stdout" | number;
    stdin?: ProcessStdio | number;
  }
```

brandonkal:
This would be very convenient when a program outputs on stdout when it should be on Deno's stderr. This would be especially nice if it streamed.

`Process::combinedOutput()` may be nice as well but not as useful if the Promise only resolved when the process exits.

jeff-hykin:
Thanks to some help from crowlKats in the discord, I created a [Sprinter](https://github.com/jeff-hykin/sprinter) to make subprocesses more convenient. If people find it useful I'll upgrade it over time.

```
// (examples use "var" and dynamic import just to be repl friendly)

// yes &> my_log.txt
var { run, Out, Overwrite } = await import(`https://deno.land/x/sprinter@0.2.0/index.js`)
await run("yes", Out(Overwrite("my_log.txt")))

// yes &>> my_log.txt
var { run, Out, AppendTo } = await import(`https://deno.land/x/sprinter@0.2.0/index.js`)
await run("yes", Out(AppendTo("my_log.txt")))

// yes 1>> my_log.txt
var { run, Stdout, AppendTo } = await import(`https://deno.land/x/sprinter@0.2.0/index.js`)
await run("yes", Stdout(AppendTo("my_log.txt")))

// yes 1>my_log.txt 2>>my_error_log.txt
var { run, Stdout, Stderr, AppendTo, Overwrite } = await import(`https://deno.land/x/sprinter@0.2.0/index.js`)
[...]

PR #11618, opened 2021-08-09
feat(runtime): two-tier subprocess API

### LINKED ISSUE #11016: Implement new two-tier subprocess API

crowlKats:
Fixes #11016

Left to do:
- [x] documentation
- [x] tests

Things ideally to have before landing:
- [x] #11025
- [x] #11757


runtime/ops/command.rs
```diff
@@ -0,0 +1,332 @@
...
+  stderr_rid: Option<ResourceId>,
+}
+
+fn op_command_spawn(
+  state: &mut OpState,
+  command_args: CommandArgs,
+  _: (),
+) -> Result<Child, AnyError> {
+  let mut command = create_command(state, command_args)?;
+  let mut child = command.spawn()?;
```
caspervonb, line 202: There's a trade-off here, while this certainly is simpler than mapping commands 1:1 it also means that if `create_command` errors out then that'll happen here rather than in `new Command(...)` which is what I would expect as the user of such a constructor.

crowlKats:
My concerns with this API is that this has too much API surface and has multiple ways to do the same thing:
`await new Command().status()` and `await new Command().spawn().wait()`
`await new Command().output()` and `await new Command().spawn().output()`
so don't see that much need for 2-tier subprocess

bartlomieju:
> My concerns with this API is that this has too much API surface and has multiple ways to do the same thing:
> `await new Command().status()` and `await new Command().spawn().wait()`
> `await new Command().output()` and `await new Command().spawn().output()`
> making the Command methods superfluous except spawn(), which doesn't make much sense and could be "collapsed"/simplified to just Child, so don't see that much need for 2-tier subprocess

That's on purpose to mimic Rust's API. There's a difference between those APIs - eg. if you use `await new Command().output()` we should automatically make sure that `stdout` and `stderr` are set to `piped`.


> On top of previous concerns, do we really need both `status()` and `output()` if `output()` includes `status()`?

Yes, there are situation where you might not want to collect output from the command - either discard it setting the pipes to `null` or use `inherit`.

Again - the idea was to mimic Rust's API which is very flexible yet it has a simple "happy path" that most of the users will use.

crowlKats:
> Yes, there are situation where you might not want to collect output from the command - either discard it setting the pipes to null or use inherit.

But output() takes care of that: https://github.com/denoland/deno/blob/046ae0ceb72a52002503141d3643b8469432dc7c/runtime/ops/command.rs#L171-L183, as else it just returns an empty uint8array

crowlKats:
> That's on purpose to mimic Rust's API. There's a difference between those APIs - eg. if you use await new Command().output() we should automatically make sure that stdout and stderr are set to piped.

Then it would rather make sense to move stdin, stderr, and stdout options from the main constructor to what currently is the spawn() method

runtime/ops/command.rs
crowlKats: Agreed, will do this, especially as i need this for some future plans

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1094,6 +1094,52 @@ declare namespace Deno {
...
   }
+
+  interface CommandOptions {
+    args?: string[];
+    cwd?: string | URL;
+    clearEnv?: boolean;
+    env?: Record<string, string>;
+  }
+
+  interface ProcessIOOptions {
```
bartlomieju, line 1105: rename to `ProcessIoOptions` to keep it camelcase?

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+  const core = window.Deno.core;
+  const { assert, pathFromURL } = window.__bootstrap.util;
+  const { read } = window.__bootstrap.io;
+  const { writeAll } = window.__bootstrap.buffer;
+  const { illegalConstructorKey } = window.__bootstrap.webUtil;
+  const { ArrayPrototypeMap, ObjectEntries, String } =
+    window.__bootstrap.primordials;
+
+  class Command {
+    #args;
```
bartlomieju, line 14: `this.#args.args` is a bit convoluted, maybe `#options`?

bartlomieju: Or maybe `defaultOptions`?

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+    constructor(command, {
+      args = [],
+      cwd = undefined,
+      clearEnv = false,
+      env = {},
+    } = {}) {
+      this.#args = {
+        cmd: pathFromURL(command),
+        args: ArrayPrototypeMap(args, String),
+        cwd: pathFromURL(cwd),
```
bartlomieju, line 25: What does `pathFromURL(undefined)` return?

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+
+  class Child {
+    #rid;
+
+    #pid;
+    get pid() {
+      return this.#pid;
+    }
+
+    #stdin;
```
bartlomieju, line 60: Is `stdin` resource created always?

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+          },
+        });
+      }
+
+      if (stdoutRid !== undefined) {
+        this.#stdoutRid = stdoutRid;
+        // TODO(crowlkats): BYOB Stream
+        this.#stdout = new ReadableStream({
+          async pull(controller) {
+            const buf = new Uint8Array(1024); // TODO(crowlkats): different number? wouldnt bee needed with byob stream
```
bartlomieju, line 107: Current limit is 16kB (https://github.com/denoland/deno/issues/10157) maybe use that?

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+    } = null) {
+      if (key !== illegalConstructorKey) {
+        throw new TypeError("Illegal constructor.");
+      }
+
+      this.#rid = rid;
+      this.#pid = pid;
+
+      if (stdinRid !== undefined) {
+        this.#stdin = new WritableStream({
```
bartlomieju, line 92: One thing about using WHATWG streams is that we can support piping to stdout/stderr from any other WHATWG stream if that was implemented in JS instead of Rust (tangential, but ref #2750)

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+
+    get status() {
+      return core.opSync("op_command_child_status", this.#rid);
+    }
+
+    async wait() {
+      return await core.opAsync("op_command_child_wait", this.#rid);
+    }
```
bartlomieju, line 149: There should by `tryWait` as well? (ie. sync API to check if process has finished)

runtime/ops/command.rs
```diff
@@ -0,0 +1,302 @@
...
+  if let Some(stdout) = &command_args.stdout {
+    command.stdout(subprocess_stdio_map(stdout)?);
+  }
+  if let Some(stderr) = &command_args.stderr {
+    command.stderr(subprocess_stdio_map(stderr)?);
+  }
+
+  // We want to kill child when it's closed
+  command.kill_on_drop(true);
+
```
bartlomieju, line 112: Add a TODO here regarding detached processes and issue about orphaned child processes when Deno exits with error or `Deno.exit()` is called.

runtime/js/40_command.js
crowlKats: `undefined`

runtime/js/40_command.js
```diff
@@ -0,0 +1,168 @@
...
+        ...options,
+      });
+    }
+
+    async output() {
+      return await core.opAsync("op_command_output", this.#args);
+    }
+  }
+
+  class Child {
```
bartlomieju, line 52: It would be great if `Child` could do cleanup of "related" resource, ie. close `stdin`, `stdout` and `stderr` once they are no longer needed/were consumed by Rust APIs

runtime/js/40_command.js
crowlKats: currently there is a `status` getter

runtime/js/40_command.js
crowlKats: these are never accessed directly but only passed to ops, but sure

runtime/js/40_command.js
crowlKats: sure

runtime/js/40_command.js
bartlomieju: Okay, sounds good, it will all be clearer once there's type declaration.

runtime/js/40_command.js
crowlKats: no, same for the other stdio

runtime/js/40_command.js
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    spawn(options = {}) {
+      const child = core.opSync("op_command_spawn", {
+        ...this.#options,
+        ...options,
+      });
+
+      return new Child(illegalConstructorKey, child);
+    }
```
bartlomieju, line 75: I believe these methods should be one-time use. Ie. if you called any of the `spawn`, `status`, `output` the next call should error saying that `Command` has already been "used"

runtime/js/40_command.js
```diff
@@ -0,0 +1,191 @@
...
+
+    async status(options = {}) {
+      const status = await core.opAsync("op_command_status", {
+        ...this.#options,
+        ...options,
+      });
+      // TODO(@crowlKats): change typings to return null instead of undefined for status.signal
```
bartlomieju, line 82: I'm don't understand what you mean here

runtime/js/40_command.js
```diff
@@ -0,0 +1,191 @@
...
+      return status;
+    }
+
+    async output() {
+      const res = await core.opAsync("op_command_child_output", {
+        rid: this.#rid,
+        stdoutRid: this.#stdoutRid,
+        stderrRid: this.#stderrRid,
+      });
+      await this.stdin?.abort();
```
bartlomieju, line 176: I believe the similar situation in here; what will happen if you do:
```js
await cmd.wait();
await cmd.output()`
```

runtime/js/40_command.js
crowlKats: `Deno.ProcessStatus` has a `signal` property that is defined as `signal?: undefined | number`, but since i create `Deno.ProcessStatus` in rust, it will type as `signal: null | number` instead (since `None` turns into `null`, not `undefined`)

runtime/js/40_command.js
crowlKats: yea that wont work, it will throw a "bad rid" error. i will add handling for that

runtime/js/40_command.js
crowlKats: this faintly relates to https://github.com/denoland/deno/pull/11618#discussion_r685257615

runtime/ops/command.rs
crowlKats: Done!

runtime/js/40_command.js
crowlKats: done

runtime/js/40_command.js
bartlomieju: I believe the preferable solution would be to map `null`s to `undefined`

crowlKats: so leave as-is and remove the 2.0 todo comments?

bartlomieju: I meant that if Rust returns `null` it should be mapped into `undefined` on the JS side. To it is `undefined | number` instead of `null | number`. Though I might be wrong. Would be great to have a second opinion on this one.

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,505 @@
...
+      env: {
+        FOO: "23147",
+      },
+    });
+
+    const output = await cmd.output();
+    const obj = JSON.parse(new TextDecoder().decode(output.stdout));
+
+    // can't check for object equality because the OS may set additional env vars for processes
+    // so we check if PATH isn't present as that is a common env var across OS's and isn't set for processes.
```
bnoordhuis, line 451: Long lines. The linter is lenient on long comments but, _ceteris paribus_, prefer wrapping at 80 columns.

runtime/js/40_command.js
```diff
@@ -0,0 +1,214 @@
...
+        clearEnv,
+        env: ObjectEntries(env),
+        uid,
+        gid,
+      });
+    }
+
+    spawn(options = {}) {
+      if (this.#rid === null) {
+        throw new TypeError("Child has already been used up.");
```
bnoordhuis, line 70: Sorta confusing error message. Maybe something more similar to this?
```suggestion
        throw new TypeError("Child process has already terminated.");
```
Also applies to other places with the same error message (I count at least 4.)

runtime/js/40_command.js
```diff
@@ -0,0 +1,214 @@
...
+    async status(options = {}) {
+      if (this.#rid === null) {
+        throw new TypeError("Child has already been used up.");
+      }
+      const status = await core.opAsync(
+        "op_command_status",
+        this.#rid,
+        options,
+      );
+      this.#rid = null;
```
bnoordhuis, line 86: What happens when I call `p.status(); p.status();`? (I.e., without waiting for the first promise to resolve.)

My guess is it calls op_command_status twice and that's probably bad. I'd write this as:
```suggestion
      const rid = this.#rid;
      this.#rid = null;
      const status = await core.opAsync(
        "op_command_status",
        rid,
        options,
      );
```
Also applies to wait() and output() (both methods.)

(Maybe calling it twice here is benign and the real bug is clearing `this.#rid` but the general advice still applies to the other methods.)

runtime/ops/command.rs
```diff
@@ -0,0 +1,359 @@
...
+    _ => Err(type_error("Invalid resource for stdio")),
+  }
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CommandIoArgs {
+  stdin: Option<String>,
+  stdout: Option<String>,
+  stderr: Option<String>,
```
bnoordhuis, line 73: I would strongly recommend using an enum instead of a string:
```rust
#[derive(Deserialize)]
#[serde(rename_all = "lowercase")]
pub enum Stdio {
  Inherit,
  Piped,
  Null,
}
```
And then:
```suggestion
  stdin: Option<Stdio>,
  stdout: Option<Stdio>,
  stderr: Option<Stdio>,
```

runtime/ops/command.rs
```diff
@@ -0,0 +1,359 @@
...
+async fn op_command_child_output(
+  state: Rc<RefCell<OpState>>,
+  args: ChildStdio,
+  _: (),
+) -> Result<CommandOutput, AnyError> {
+  let resource = state
+    .borrow_mut()
+    .resource_table
+    .take::<ChildResource>(args.rid)?;
+  let resource = Rc::try_unwrap(resource).ok().unwrap();
```
bnoordhuis, line 326:
```suggestion
  let resource = Rc::try_unwrap(resource).unwrap();
```

runtime/ops/command.rs
```diff
@@ -0,0 +1,359 @@
...
+    stdout: if args.stdout_rid.is_some() {
+      Some(output.stdout.into())
+    } else {
+      None
+    },
+    stderr: if args.stderr_rid.is_some() {
+      Some(output.stderr.into())
+    } else {
+      None
+    },
```
bnoordhuis, line 357: It's not wrong but you can write this more succinctly as follows:
```suggestion
    stdout: args.stdout_rid.then(|| output.stdout.into()),
    stderr: args.stderr_rid.then(|| output.stderr.into()),
```

bnoordhuis (commented):
I have a concern about stdio buffering but by and large looking very good.

Stdio buffering: ideally, stdio is streamed, not buffered, so you can interact with e.g. a REPL.

Supporting both modes is okay too but in that case there should probably be a configurable circuit breaker (like node's `maxBuffer` option) so the output of my `10 PRINT "HELLO WORLD"; 20 GOTO 10` script doesn't fill up all available memory.

cli/tests/unit/command_test.ts
crowlKats: actually just copied from `cli/tests/unit/process_test.ts`, but sure, will fix both

runtime/ops/command.rs
crowlKats: `try_unwrap` doesnt allow calling `unwrap` directly on it, so gotta hack around that with `ok`

runtime/js/40_command.js
crowlKats: I can do this for the `Command` class, but this becomes a bit problematic with the `Child` class, due to the `status` getter

crowlKats:
> I have a concern about stdio buffering but by and large looking very good.
> 
> Stdio buffering: ideally, stdio is streamed, not buffered, so you can interact with e.g. a REPL.
> 
> Supporting both modes is okay too but in that case there should probably be a configurable circuit breaker (like node's `maxBuffer` option) so the output of my `10 PRINT "HELLO WORLD"; 20 GOTO 10` script doesn't fill up all available memory.

The default value of `16384` for the readablestream is just a default fallback, which is used when `ReadableStream.getReader().read()` is used. you can use `ReadableStream.getReader({ mode: "byob" }).read(new Uint8Array(10))`, and it will only read the requested amount

bnoordhuis:
For my understanding: this PR only implements buffered I/O, not streaming I/O, right?

crowlKats:
from what i remember doing months ago and quick glance at the code, it should be streaming 🤔 might be wrong

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1108,6 +1108,92 @@ declare namespace Deno {
...
+      : null;
+    readonly stderr: T["stderr"] extends "piped" ? ReadableStream<Uint8Array>
+      : null;
+
+    readonly pid: number;
+    /** Get the current status. */
+    readonly status: ProcessStatus | null;
+
+    /** Waits for the child to exit completely, returning the status that it exited with. */
+    wait(): Promise<ProcessStatus>;
```
bartlomieju, line 1184: What about `tryWait()` method that checks if process did exit but is synchronous?

crowlKats: thats the `status` getter

lucacasonato (commented):
I don't think `wait()` should be a method. I think it should instead be a `.status` promise. This has two benefits:
- You can't accidentally call `wait()` twice, which is something Tokio does not like.
- Because we start the "waiting" async op immediately after starting the subprocess, we can attach a listener to it that aborts the `ReadableStream` / `WritableStream` automatically when the process closes.

It also more closely aligns to how "closing" is handled in other new Web APIs, like `WebSocketStream`.

bartlomieju:
> You can't accidentally call wait() twice, which is something Tokio does not like.

Tokio has no problem with that:
```
Waits for the child to exit completely, returning the status that it exited with. This function will continue to have the same return value after it has been called at least once.
```
https://docs.rs/tokio/1.15.0/tokio/process/struct.Child.html#method.wait

lucacasonato:
Oh, I feel like that used to be different. I still think it should not be a method. Why should it be a method that creates a promise, rather than just being a promise directly?

There is no action that `wait` performs - it just notifies us of a pending closure. On the web, these "status updates" are either promise properties of the relevant resource (`WebSocketStream#connection` & `WebSocketStream#closed`), or events (`WebSocket.onClose`).

bartlomieju:
> Oh, I feel like that used to be different. I still think it should not be a method. Why should it be a method that creates a promise, rather than just being a promise directly?
> 
> There is no action that `wait` performs - it just notifies us of a pending closure. On the web, these "status updates" are either promise properties of the relevant resource (`WebSocketStream#connection` & `WebSocketStream#closed`), or events (`WebSocket.onClose`).

Actually there is an action - by default Tokio closes `stdin` handle to not hang the process indefinitely, but I believe in our case it's not always desirable.

lucacasonato:
Yeah, didn't we say we don't want to do that for the low level API?

crowlKats:
so remove `status`, and replace `wait()` for `wait`, à la `WSS.connection`?

bnoordhuis:
@crowlKats Does the new API address #7087?

crowlKats:
> @crowlKats Does the new API address #7087?

no it does not

bartlomieju:
Note to self, to land this we need to clarify following topics:
- clarify `status()` API
- clarify `wait()` API

crowlKats:
So, the only missing thing is figure out typings (as we use different stdio defaults between `spawn` & `command`, which affect return typings), and naming. Also, this currently doesnt add any new features in terms of options that can be passed, like `input` for `command`, as that hasnt been discussed yet. But besides these aforementioned things, it is done

bartlomieju:
@crowlKats what about the helper APIs for 90% of use cases (`Deno.exec`/`Deno.execSync`)? I think it might be better to land them first after all.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3010,4 +3013,94 @@ declare namespace Deno {
...
+    stdout?: "piped" | "inherit" | "null";
+    stderr?: "piped" | "inherit" | "null";
+  }
+
+  /**
+   * Spawns a child process.
+   *
+   * stdin defaults to null.
+   * stdout defaults to inherit.
+   * stderr defaults to inherit.
```
bartlomieju, line 3051:
```suggestion
   * `CommandOptions.stdin` defaults to `"null"`.
   * `CommandOptions.stdout` defaults to `"inherit"`.
   * `CommandOptions.stderr` defaults to `"inherit"`.
```

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3010,4 +3013,94 @@ declare namespace Deno {
...
+
+    readonly pid: number;
+    /** Get the current status. */
+    readonly status: ProcessStatus | null;
+
+    /** Waits for the child to exit completely, returning the status that it exited with. */
+    wait(): Promise<ProcessStatus>;
+    /** Waits for the child to exit completely, returning all its output and status. */
+    output(): Promise<CommandOutput<T>>;
+    kill(signo: Signal): void;
```
bartlomieju, line 3074: Is `signo` a required argument? If not, please add a docstring saying which signal is used by default

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3010,4 +3013,94 @@ declare namespace Deno {
...
+  }
+
+  /**
+   * Executes a subprocess as a child process, waiting for it to finish and collecting all of its output.
+   *
+   * stdin defaults to null.
+   * stdout defaults to piped.
+   * stderr defaults to piped.
```
bartlomieju, line 3082:
```suggestion
   * Executes a subprocess, waiting for it to finish and collecting all of its output.
   *
   * `CommandOptions.stdin` defaults to `"null"`.
   * `CommandOptions.stdout` defaults to `"piped"`.
   * `CommandOptions.stderr` defaults to `"piped"`.
```

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3010,4 +3013,94 @@ declare namespace Deno {
...
+    command: string | URL,
+    options?: T,
+  ): Promise<CommandOutput<T>>;
+
+  /**
+   * Synchronously executes a subprocess as a child process, waiting for it to finish and collecting all of its output.
+   *
+   * stdin defaults to null.
+   * stdout defaults to piped.
+   * stderr defaults to piped.
```
bartlomieju, line 3094:
```suggestion
   * Synchronously executes a subprocess, waiting for it to finish and collecting all of its output.
   *
   * `CommandOptions.stdin` defaults to `"null"`.
   * `CommandOptions.stdout` defaults to `"piped"`.
   * `CommandOptions.stderr` defaults to `"piped"`.
```

cli/dts/lib.deno.ns.d.ts
bartlomieju: Please add an example usage in a codeblock

cli/dts/lib.deno.ns.d.ts
bartlomieju: Please add an example of use in a codeblock

cli/dts/lib.deno.ns.d.ts
bartlomieju: Please add an example of use in a codeblock

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,632 @@
...
+
+Deno.test(
+  { permissions: { run: true, read: true } },
+  function commandSyncStderrOutput() {
+    const { stderr } = Deno.commandSync(Deno.execPath(), {
+      args: [
+        "eval",
+        "await Deno.stderr.write(new TextEncoder().encode('error'))",
+      ],
+    });
```
bartlomieju, line 449: Very nice 👍 this could be used as an example in declaration files

runtime/js/40_command.js
```diff
@@ -0,0 +1,198 @@
...
+      }
+      const res = await core.opAsync("op_command_output", {
+        rid: this.#rid,
+        stdoutRid: this.#stdoutRid,
+        stderrRid: this.#stderrRid,
+      });
+      await this.stdin?.abort();
+      this.#rid = null;
+      // TODO(@crowlKats): 2.0 change typings to return null instead of undefined for status.signal
+      res.status.signal ??= undefined;
```
bartlomieju, line 137: This could be done now?

runtime/js/40_command.js
```diff
@@ -0,0 +1,198 @@
...
+      core.opSync("op_kill", this.#pid, signo);
+    }
+  }
+
+  function command(command, {
+    stdin = "null",
+    stdout = "piped",
+    stderr = "piped",
+    ...options
+  } = {}) { // TODO(@crowlKats): more options (like input)?
```
bartlomieju, line 155: By input you mean a string that should be written into the stdin? If so I think we should add it. Also it's a bit strange that one can configure `stdin` if there's no way to write to it.

bartlomieju (commented):
Nice work @crowlKats, the "shorthand" API feels great

runtime/js/40_command.js
crowlKats: @bartlomieju we could if we have a new interface and remove the old one, but cant change the existing one without breaking `Deno.run`. I dont really care eitherway, you tell me

bartlomieju: Ah I see, do you think we could introduce a separate interface so there are no changes to `Deno.run` whatsoever?

crowlKats: sure

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3010,4 +3013,106 @@ declare namespace Deno {
...
+    stdout?: "piped" | "inherit" | "null";
+    stderr?: "piped" | "inherit" | "null";
+  }
+
+  /**
+   * Spawns a child process.
+   *
+   * `CommandOptions.stdin` defaults to `"null"`.
+   * `CommandOptions.stdout` defaults to `"inherit"`.
+   * `CommandOptions.stderr` defaults to `"inherit"`.
```
bartlomieju, line 3051: I think I made mistake here... shouldn't this be `options.stdin` default to `"null"`, etc?

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,632 @@
...
+  assertRejects,
+  assertStringIncludes,
+  assertThrows,
+} from "./test_util.ts";
+
+Deno.test(
+  { permissions: { write: true, run: true, read: true } },
+  async function spawnWithCwdIsAsync() {
+    const enc = new TextEncoder();
+    const cwd = await Deno.makeTempDir({ prefix: "deno_command_test" });
```
kt3k, line 14: nit: better to remove this dir at the end of the test case? `await Deno.remove(cwd, { recursive: true });`

kt3k:
I might misunderstand something, but now `command` and `commandSync` are shorthand methods, and `spawn` is 2-tier API. I thought we planned in opposite way: `command` (or `subprocess`) for 2-tier API and `spawn` (or `exec`) for shorthands. Is this switch intentional? (or I misunderstood the conversation?)

(I imagined something like the below:
```ts
const cmd = Deno.command(Deno.execPath(), { args });
const status = await cmd.status();
```

crowlKats:
@kt3k I renamed `exec` to `command` and what previously was `Command.spawn()` is now just `spawn`. Final naming of things is still to be decided

kt3k:
@crowlKats Thanks for the clarification! Ok. This is an intended change

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3010,4 +3013,106 @@ declare namespace Deno {
...
+    options?: T,
+  ): Child<T>;
+
+  export class Child<T extends CommandOptions> {
+    readonly stdin: T["stdin"] extends "inherit" | "null" ? null
+      : WritableStream<Uint8Array>;
+    readonly stdout: T["stdout"] extends "inherit" | "null" ? null
+      : ReadableStream<Uint8Array>;
+    readonly stderr: T["stderr"] extends "inherit" | "null" ? null
+      : ReadableStream<Uint8Array>;
```
bartlomieju, line 3064: `extends "piped" : ReadableStream<Uint8Array> : null`
WDYT? Seems shorted

cli/dts/lib.deno.ns.d.ts
crowlKats: maybe? not sure, both make sense, right now it is more explicit, but your suggested change does also make sense

cli/dts/lib.deno.ns.d.ts
crowlKats: Nope, its not optional, just like with `Deno.run`. Do you want it to be optional?

cli/dts/lib.deno.ns.d.ts
crowlKats: yep, and actually correct-er

runtime/js/40_command.js
crowlKats: can we actually keep that for a follow-up? would simplify things

cli/dts/lib.deno.ns.d.ts
bartlomieju: Up to you

cli/dts/lib.deno.ns.d.ts
bartlomieju: Nope. Makes sense then. please add a docstring

runtime/js/40_command.js
bartlomieju: Yes, please open an issue about it before landing so we don't forget about it

kt3k:
How about these names?

- `Deno.spawn`/`Deno.spawnSync` - shorthands
- `Deno.spawnChild` - extensible API, returning `Child` type

nayeemrmn:
What's wrong with `Deno.exec()`/`Deno.execSync()` for cli-like shorthands and `Deno.spawn()` for child handle? The connotations seem right.

crowlKats:
> How about these names?
> 
> 
> 
> - `Deno.spawn`/`Deno.spawnSync` - shorthands
> 
> - `Deno.spawnChild` - extensible API, returning `Child` type

Hahaha, that's exactly what I thought about and mentioned to bartek a few hours ago

crowlKats:
> ~~What's wrong with `Deno.exec()`/`Deno.execSync()` for cli-like shorthands and `Deno.spawn()` for child handle? The connotations seem right.~~ Edit: Oh, adjacent docs. My suggestions would be:
> 
> - `Deno.exec()`
> 
> - `Deno.execSync()`
> 
> - `Deno.execHandle()`

We have decided that exec is to be avoided, as it doesn't have anything to do with exec(2), and the Deno ns APIs are based on unix

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3029,4 +3032,136 @@ declare namespace Deno {
...
+      : null;
+    readonly stderr: T["stderr"] extends "piped" ? ReadableStream<Uint8Array>
+      : null;
+
+    readonly pid: number;
+    /** Get the current status. */
+    readonly status: CommandStatus | null;
+
+    /** Waits for the child to exit completely, returning the status that it exited with. */
+    wait(): Promise<CommandStatus>;
```
lucacasonato, line 3101: I still don't like this as I had elaborated in earlier issue threads and comments on this PR. I think this should be `status: Promise<CommandStatus>`. My suggestion has web API precedent (`closed` field on `WebSocketStream`), whereas the currently implemented API does not.

runtime/js/40_command.js
```diff
@@ -0,0 +1,187 @@
...
+    async output() {
+      if (this.#rid === null) {
+        throw new TypeError("Child process has already terminated.");
+      }
+      const res = await core.opAsync("op_command_output", {
+        rid: this.#rid,
+        stdoutRid: this.#stdoutRid,
+        stderrRid: this.#stderrRid,
+      });
+      await this.stdin?.abort();
```
lucacasonato, line 127: What happens when you abort an already closed, or a locked stream?

runtime/js/40_command.js
```diff
@@ -0,0 +1,187 @@
...
+        this.#rid,
+        this.#stdinRid,
+      );
+      await this.stdin?.abort();
+      this.#rid = null;
+      this.#status = status;
+      return status;
+    }
+
+    async output() {
```
lucacasonato, line 118: If the user has already locked either the `stdout` or `stderr` stream already, this function should throw.

runtime/ops/command.rs
```diff
@@ -0,0 +1,324 @@
...
+    .resource_table
+    .take::<ChildResource>(rid)?;
+  let mut child = RcRef::map(resource, |r| &r.0).borrow_mut().await;
+  if let Some(stdin_rid) = stdin_rid {
+    let stdin = state
+      .borrow_mut()
+      .resource_table
+      .take::<ChildStdinResource>(stdin_rid)?;
+    child.stdin = Some(Rc::try_unwrap(stdin).unwrap().into_inner());
+  }
```
lucacasonato, line 253: I thought we had previously established that we do not want the auto-closing stdin behavior for the low level API?

lucacasonato (commented):
I feel like a lot of the discussion from the subprocess design meetings was not incorporated into this PR. I currently fail to see how `Deno.spawnChild` is any different to `Deno.run` in behaviour. None of the behavioural issues we discussed in the subprocess design meetings were addressed (namely stdin auto-close on wait).

lucacasonato (commented):
The stdin auto-close behaviour isn't even consistent. Only `wait()` auto-closes, while `output()` does not...

cli/dts/lib.deno.ns.d.ts
crowlKats: someone deep into this comment section I have explained: this is not possible. 
> The stdin handle to the child process, if any, will be closed before waiting. This helps avoid deadlock: it ensures that the child does not block waiting for input from the parent, while the parent waits for the child to exit.

from the tokio docs for [Child.wait()](https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.wait), which is different behaviour to `closed` field. a method makes this more intuitive that it will have a side-effect

runtime/ops/command.rs
crowlKats: not that I am aware. also, once again:
> The stdin handle to the child process, if any, will be closed before waiting. This helps avoid deadlock: it ensures that the child does not block waiting for input from the parent, while the parent waits for the child to exit.

and since we want this similar to the rust API, we should do this imo.

lucacasonato: You might have not been part of this discussion, but we definitely established this. I can elaborate on why if you want.

crowlKats: sure

lucacasonato: Essentially you want to allow users to perform the following:

```js
const child = Deno.spawnChild("foo", { stdin: "piped" });
Deno.stdin.readable.pipeTo(child.stdin).catch(e => console.error("piping failed", e))
const status = await child.status;
console.log("the child exited", status);
```

If we **always** auto-close stdin, this doesn't work anymore.

Also auto-closing behaviour is not what Node, or most other runtimes do. Tokio also only does it if you don't manually take control over the stdin handle (which we do, but then for some reason return it control before calling `Child#wait()`).

lucacasonato: Also this behaviour is totally inconsistent with `.output()`, which doesn't yield tokio control over the stdin handle, so it doesn't auto-close stdin when calling `Child#output()`.

cli/dts/lib.deno.ns.d.ts
lucacasonato: I think we shouldn't have this side effect - see comments below. This is also not an explicit operation in Node, or in most other runtimes.

runtime/js/40_command.js
```diff
@@ -119,12 +114,17 @@
       if (this.#rid === null) {
         throw new TypeError("Child process has already terminated.");
       }
+      if (this.#stdout?.locked || this.#stderr?.locked) {
+        throw new TypeError(
+          "Can't collect output because stdout or stderr are locked",
+        );
+      }
```
lucacasonato, line 121: Make these two if statements, so the error message can be more precise.

crowlKats:
Now the only missing thing is figure out typings (as we use different stdio defaults between spawn & spawnChild, which affect return typings).

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3029,4 +3032,134 @@ declare namespace Deno {
     recordType: RecordType,
     options?: ResolveDnsOptions,
   ): Promise<string[] | MXRecord[] | SRVRecord[] | string[][]>;
+
+  export interface CommandOptions {
```
ry, line 3036: SpawnOptions

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3029,4 +3032,134 @@ declare namespace Deno {
...
+   * `CommandOptions.stdout` defaults to `"piped"`.
+   * `CommandOptions.stderr` defaults to `"piped"`.
+   *
+   * ```ts
+   * const { status } = await Deno.spawn(Deno.execPath(), {
+   *   args: [
+   *     "eval",
+   *     "console.log('Hello World')",
+   *   ],
+   * });
```
ry, line 3119: extend this example to include 
```
assertEquals("Hello World", stdout);
```

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2414,6 +2415,7 @@ declare namespace Deno {
       signal?: number;
     };
 
+  /** @deprecated Use `Deno.command` or `Deno.spawn`. */
```
lucacasonato, line 2418: Wrong names

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2273,6 +2273,7 @@ declare namespace Deno {
     options?: { recursive: boolean },
   ): FsWatcher;
 
+  /** @deprecated Use `Deno.command` or `Deno.spawn`. */
```
lucacasonato, line 2276: Wrong names

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2427,7 +2429,8 @@ declare namespace Deno {
     stdin?: "inherit" | "piped" | "null" | number;
   }
 
-  /** Spawns new subprocess.  RunOptions must contain at a minimum the `opt.cmd`,
+  /** @deprecated Use `Deno.command` or `Deno.spawn`.
```
lucacasonato, line 2432: Wrong names

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3029,4 +3032,134 @@ declare namespace Deno {
...
+   *   ],
+   * });
+   * ```
+   */
+  export function spawnSync<T extends CommandOptions = CommandOptions>(
+    command: string | URL,
+    options?: T,
+  ): CommandOutput<T>;
+
+  export type CommandStatus =
```
lucacasonato, line 3148: `ChildStatus`

cli/dts/lib.deno.ns.d.ts
```diff
@@ -3029,4 +3032,134 @@ declare namespace Deno {
...
+      code: 0;
+      signal: null;
+    }
+    | {
+      success: false;
+      code: number;
+      signal: number | null;
+    };
+
+  export interface CommandOutput<T extends CommandOptions> {
```
lucacasonato, line 3160: `SpawnOutput`

lucacasonato (commented):
The ops are not marked as unstable.

crowlKats:
@lucacasonato https://github.com/denoland/deno/blob/9af9577b8acd2c184bd75f75ae76285c82d583ec/runtime/ops/command.rs#L133

cli/diagnostics.rs
```diff
@@ -66,6 +66,11 @@ const UNSTABLE_DENO_PROPS: &[&str] = &[
   "umask",
   "utime",
   "utimeSync",
+  "CommandOptions",
+  "ProcessIoOptions",
+  "Command",
+  "Child",
+  "CommandOutput",
```
bartlomieju, line 73: Should be updated with latest interface names

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1363,6 +1363,129 @@ declare namespace Deno {
...
+   * ```ts
+   * const child = Deno.spawnChild(Deno.execPath(), {
+   *   args: [
+   *     "eval",
+   *     "console.log('Hello World')",
+   *   ],
+   * });
+   *
+   * const status = await child.status;
+   * ```
```
bartlomieju, line 1411: Add an example of using stdio streams (preferably with web streams)

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1363,6 +1363,129 @@ declare namespace Deno {
...
+    output(): Promise<SpawnOutput<T>>;
+    /** Kills the process with given Signal. */
+    kill(signo: Signal): void;
+  }
+
+  /**
+   * Executes a subprocess, waiting for it to finish and collecting all of its output.
+   *
+   * ```ts
+   * const { stdout } = await Deno.spawn(Deno.execPath(), {
```
bartlomieju, line 1440: Use `status` and `stderr` in example too

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,644 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import {
+  assert,
+  assertEquals,
+  assertRejects,
+  assertStringIncludes,
+  assertThrows,
+} from "./test_util.ts";
+
```
bartlomieju, line 9: Add a test for `Deno.spawn`/`Deno.spawnSync` that override `options.stdout` and `options.stderr` to showcase that they will always be set to "piped" internally

cli/dts/lib.deno.unstable.d.ts
bartlomieju: Add comment and example that `child.stdin` must be closed manually

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,644 @@
...
+  },
+);
+
+Deno.test(
+  { permissions: { run: true, read: true } },
+  async function commandSuccess() {
+    const { status } = await Deno.spawn(Deno.execPath(), {
+      args: ["eval", "console.log('hello world')"],
+      stderr: "null",
+      stdout: "null",
```
bartlomieju, line 279: This should be internally overriden to "piped", that's the whole point of this API

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,644 @@
...
+  },
+);
+
+Deno.test(
+  { permissions: { run: true, read: true } },
+  function commandSyncSuccess() {
+    const { status } = Deno.spawnSync(Deno.execPath(), {
+      args: ["eval", "console.log('hello world')"],
+      stderr: "null",
+      stdout: "null",
```
bartlomieju, line 294: Ditto

cli/tests/unit/command_test.ts
bartlomieju: Assert the output is "hello world"

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,644 @@
...
+    assertEquals(status.success, true);
+    assertEquals(status.code, 0);
+    assertEquals(status.signal, null);
+  },
+);
+
+Deno.test(
+  { permissions: { run: true, read: true } },
+  async function commandUrl() {
+    const { status } = await Deno.spawn(
```
bartlomieju, line 306: Assert the `output` is `hello world`

cli/tests/unit/command_test.ts
```diff
@@ -0,0 +1,644 @@
...
+    assertEquals(status.code, 0);
+    assertEquals(status.signal, null);
+  },
+);
+
+Deno.test(
+  { permissions: { run: true, read: true } },
+  function commandSyncUrl() {
+    const { status } = Deno.spawnSync(new URL(`file:///${Deno.execPath()}`), {
+      args: ["eval", "console.log('hello world')"],
```
bartlomieju, line 323: Ditto

lucacasonato (approved):
CI is failing on linting, but other than that LGTM!

Thanks for seeing this through @crowlKats! Very exciting to have this finally done. 🎉

bartlomieju (approved):
LGTM. You are most likely the winner in "the longest time to merge the PR" contest in Deno @crowlKats 🤣

crowlKats:
@bartlomieju and i think second place is webgpu, which took 7 months 🤣

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1363,6 +1363,145 @@ declare namespace Deno {
...
+   * const child = Deno.spawnChild(Deno.execPath(), {
+   *   args: [
+   *     "eval",
+   *     "console.log('Hello World')",
+   *   ],
+   *   stdin: "piped",
+   * });
+   *
+   * // open a file and pipe the subprocess output to it.
+   * child.stdout.pipeTo(Deno.open("output").writable);
```
lucacasonato, line 1414:
```suggestion
   * child.stdout.pipeTo(Deno.openSync("output").writable);
```

PR #11622, opened 2021-08-09
build: pin softprops/action-gh-release

lucacasonato:
https://github.com/softprops/action-gh-release/issues/139


dsherret (approved):
LGTM

Issue #11481, opened 2021-07-21
Enable non-extractable keys for Web Crypto


lucacasonato:
Before some we can do this we need confirmation from spec authors for the following statements:

1. Non extractable keys are not meant to be a security primitive. They are purely meant for obfuscation.
2. It is expected that a compromise of the render process, or in process timing side channel attacks (Spectre) can extract "non extractable" keys.

If either of these statements are not correct we do not have any technical way to implement non extractable keys securely.

Knowing more about how web crypto is implemented in browsers (Chrome, Safari, and Firefox) would help here:
- Is underlying key material stored in the render process, or in a separate process that communicates with the render process via IPC for any cryptographic operations?

littledivy:
The spec recommends use of IndexedDB for storing key material. https://www.w3.org/TR/WebCryptoAPI/#concepts-key-storage

---
Also: 
> Authors should be aware that this specification places no normative requirements on implementations as to how the underlying cryptographic key material is stored. The only requirement is that key material is not exposed to script, except through the use of the exportKey and wrapKey operations

I'll assume storing key material in memory is fine with extractable=false

panva:
The key language from that section I believe is **in bold**

> This allows the storage and retrieval of key material, **without ever exposing that key material to the application or the JavaScript environment**

panva:
https://www.w3.org/TR/WebCryptoAPI/#scope-out-of-scope

> This API, while allowing applications to generate, retrieve, and manipulate keying material, does not specifically address the provisioning of keys **in particular types of key storage, such as secure elements or smart cards**. This is due to such provisioning operations often being burdened with vendor-specific details that make defining a vendor-agnostic interface an unsuitably unbounded task. Additionally, this API does not deal with or address the discovery of cryptographic modules, as such concepts are dependent upon the underlying user agent and are not concepts that are portable between common operating systems, cryptographic libraries, and implementations.

lucacasonato:
These sections are all non normative. We can't make a decision on this until we have more clarity about other vendor implementations. Also see https://github.com/w3c/webcrypto/issues/269

lucacasonato:
Ok, this has been unblocked. Discussion [here](https://github.com/w3c/webcrypto/issues/269) has resulted in the observation that all shipping implementations store key material in an unprivileged process, thus not being "spectre safe". We can thus do the same.

lucacasonato:
[...]

PR #11642, opened 2021-08-11
feat(extensions/crypto): implement importKey and deriveBits for Pbkdf2

littledivy:
Introduces `deriveBits` in SubtleCrypto

### DIFF
--- a/extensions/crypto/00_crypto.js
+++ b/extensions/crypto/00_crypto.js
@@ -56,6 +56,7 @@
     RsaPssParams: {},
     EcdsaParams: { hash: "HashAlgorithmIdentifier" },
     HmacImportParams: { hash: "HashAlgorithmIdentifier" },
+    Pbkdf2Params: { hash: "HashAlgorithmIdentifier", salt: "BufferSource" },
   };
 
   const supportedAlgorithms = {
@@ -84,7 +85,11 @@
     },
     "importKey": {
       "HMAC": "HmacImportParams",
+      "PBKDF2": "Pbkdf2Params",
     },
+    "deriveBits": {
+      "PBKDF2": null,
+    }
   };
 
   // See https://www.w3.org/TR/WebCryptoAPI/#dfn-normalize-an-algorithm
@@ -534,6 +539,49 @@
         // TODO(@littledivy): RSASSA-PKCS1-v1_5
         // TODO(@littledivy): RSA-PSS
         // TODO(@littledivy): ECDSA
+        case "PBKDF2": {
+          // 1.
+          if(format !== "raw") {
+            throw new DOMException("Format not supported", "NotSupportedError");
+          }
+
+          // 2.
+          if (
+            ArrayPrototypeFind(
+              keyUsages,
+              (u) => !ArrayPrototypeIncludes(["deriveKey", "deriveBits"], u),
+            ) !== undefined
+          ) {
+            throw new DOMException("Invalid key usages", "SyntaxError");
+          }
+
+          // 3.
+          if(extractable !== false) {
+            throw new DOMException("Key must not be extractable", "SyntaxError");
+          }
+
+          // 4.
+          const handle = {};
+          WeakMapPrototypeSet(KEY_STORE, handle, {
+            type: "raw",
+            data: keyData,
+          });
+
+          // 5-9.
+          const algorithm = {
+            name: "PBKDF2"
+          };
+          const key = constructKey(
+            "secret",
+            false,
+            usageIntersection(keyUsages, recognisedUsages),
+            algorithm,
+            handle,
+          );
+          
+          // 10.
+          return key;
+        }
         default:
           throw new DOMException("Not implemented", "NotSupportedError");
       }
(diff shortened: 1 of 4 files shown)


panva:
Related to #11481 - PBKDF mandates use of non-extractable CryptoKey instances that aren't greenlit for use in Deno. A setback to interoperability and in this case even ability to land a basic webcrypto feature.

ext/crypto/00_crypto.js
```diff
@@ -606,6 +657,76 @@
...
+        prefix,
+        context: "Argument 2",
+      });
+      length = webidl.converters["unsigned long"](length, {
+        prefix,
+        context: "Argument 3",
+      });
+
+      const normalizedAlgorithm = normalizeAlgorithm(algorithm, "deriveBits");
+      switch (normalizedAlgorithm.name) {
```
lucacasonato, line 684: Step 7 & 8 of the spec are missing:

> 7. If the name member of normalizedAlgorithm is not equal to the name attribute of the [[algorithm]] internal slot of baseKey then throw an InvalidAccessError.
>
> 8. If the [[usages]] internal slot of baseKey does not contain an entry that is "deriveBits", then throw an InvalidAccessError.

ext/crypto/00_crypto.js
```diff
@@ -606,6 +657,76 @@
...
+        case "PBKDF2": {
+          // 1.
+          if (length == null || length == 0 || length % 8 !== 0) {
```
lucacasonato, line 687: Where does it say that length may not be 0?

ext/crypto/00_crypto.js
```diff
@@ -606,6 +657,76 @@
...
+            throw new DOMException("Invalid length", "OperationError");
+          }
+
+          // TODO(@littledivy): Add this step to spec. WPT has tests for it.
```
lucacasonato, line 691: It is in spec: https://w3c.github.io/webcrypto/#pbkdf2-operations. Are you looking at the candidate recommendation rather than the editors draft?

lucacasonato (commented):
Looks great from my side, other than the minor issues. @bnoordhuis please review the Rust side code too.

ext/crypto/00_crypto.js
littledivy: Forgot to remove that...added it to the spec myself (https://github.com/w3c/webcrypto/commit/d2917f41d1df1f8937718b67ee4746cf74515446)

ext/crypto/00_crypto.js
littledivy: Not in the spec but WPT has a test for it https://github.com/web-platform-tests/wpt/blob/561fc4c5463673e5291894705c7a6aeb88e78d99/WebCryptoAPI/derive_bits_keys/pbkdf2.js#L118

ext/crypto/00_crypto.js
littledivy: Done

ext/crypto/00_crypto.js
```diff
@@ -1185,6 +1278,52 @@
     }
   }
 
+  async function deriveBits(normalizedAlgorithm, baseKey, length) {
+    switch (normalizedAlgorithm.name) {
+      case "PBKDF2": {
+        // 1.
+        if (length == null || length == 0 || length % 8 !== 0) {
```
bnoordhuis, line 1285: Can you check if there's e.g. a WPT test that tests non-numeric inputs?

The logic is correct because e.g. `"bad" % 8` evaluates to `NaN` and `NaN !== 0` but I suspect that if it somehow managed to slip through to Rust land, it'd get coerced to zero.

ext/crypto/lib.rs
```diff
@@ -519,6 +522,45 @@ pub async fn op_crypto_verify_key(
...
+        CryptoHash::Sha256 => pbkdf2::PBKDF2_HMAC_SHA256,
+        CryptoHash::Sha384 => pbkdf2::PBKDF2_HMAC_SHA384,
+        CryptoHash::Sha512 => pbkdf2::PBKDF2_HMAC_SHA512,
+      };
+
+      // This will never panic. We have already checked length earlier.
+      let iterations =
+        NonZeroU32::new(args.iterations.ok_or_else(not_supported)?).unwrap();
+      let secret = args.key.data;
+      let mut out = vec![0; args.length / 8];
```
bnoordhuis, line 556: I'd add an assert that checks `args.length > 0` and a multiple of 8.

bnoordhuis (approved):

ext/crypto/00_crypto.js
lucacasonato: @bnoordhuis All the input types are always valid. They are validated by various webidl converters at the top of this function.

littledivy: Only [`null`](https://github.com/web-platform-tests/wpt/blob/03122ad70da015cc97a9bcd400f8eb83318d000d/WebCryptoAPI/derive_bits_keys/pbkdf2.js#L107) as non-numeric length is tested in WPT.

Wouldn't the WebIDL converter (`unsigned long`) also complain?

lucacasonato: Yes, it will. Strings will already fail parsing there, so will never reach this statement.

ext/crypto/00_crypto.js
lucacasonato: Can you open a spec issue?

littledivy: Sure - https://github.com/w3c/webcrypto/issues/274

ext/crypto/lib.rs
littledivy: done

PR #11648, opened 2021-08-11
feat: ffi string and buffer support

eliassjogreen:
Follow up on #11152 adding support for strings and buffer arguments and returns.

### DIFF
--- a/extensions/ffi/lib.rs
+++ b/extensions/ffi/lib.rs
@@ -10,6 +10,7 @@ use deno_core::Extension;
 use deno_core::OpState;
 use deno_core::Resource;
 use deno_core::ResourceId;
+use deno_core::ZeroCopyBuf;
 use dlopen::raw::Library;
 use libffi::middle::Arg;
 use serde::Deserialize;
@@ -17,6 +18,9 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::ffi::c_void;
+use std::ffi::CStr;
+use std::ffi::CString;
+use std::os::raw::c_char;
 use std::rc::Rc;
 
 pub struct Unstable(pub bool);
@@ -50,6 +54,7 @@ struct Symbol {
   ptr: libffi::middle::CodePtr,
   parameter_types: Vec<NativeType>,
   result_type: NativeType,
+  result_length: Option<usize>,
 }
 
 struct DynamicLibraryResource {
@@ -90,6 +95,7 @@ impl DynamicLibraryResource {
         ptr,
         parameter_types: parameter_types.collect(),
         result_type,
+        result_length: foreign_fn.result_length,
       },
     );
 
@@ -131,6 +137,8 @@ enum NativeType {
   ISize,
   F32,
   F64,
+  String,
+  Buffer,
 }
 
 impl From<NativeType> for libffi::middle::Type {
@@ -149,6 +157,8 @@ impl From<NativeType> for libffi::middle::Type {
       NativeType::ISize => libffi::middle::Type::isize(),
       NativeType::F32 => libffi::middle::Type::f32(),
       NativeType::F64 => libffi::middle::Type::f64(),
+      NativeType::String => libffi::middle::Type::pointer(),
+      NativeType::Buffer => libffi::middle::Type::pointer(),
     }
   }
 }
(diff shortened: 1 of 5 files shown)


extensions/ffi/lib.rs
```diff
@@ -169,86 +179,77 @@ impl From<String> for NativeType {
...
+  I16(i16),
+  U32(u32),
+  I32(i32),
+  U64(u64),
+  I64(i64),
+  USize(usize),
+  ISize(isize),
+  F32(f32),
+  F64(f64),
+  String(*const i8),
```
marcushultman, line 203: Any reason to not use primitive slice all the way?

ext/ffi/00_ffi.js
```diff
@@ -13,8 +13,29 @@
...
       for (const symbol in symbols) {
-        this.symbols[symbol] = (...parameters) =>
-          core.opSync("op_ffi_call", { rid: this.#rid, symbol, parameters });
+        this.symbols[symbol] = (...args) => {
+          const parameters = [];
+          const buffers = [];
+
+          for (const arg of args) {
+            if (
+              arg && arg.buffer instanceof ArrayBuffer &&
```
ReddikHaien, line 22: Couldn't this be simplified to `arg?.buffer instanceof ArrayBuffer && ...`?

test_ffi/src/lib.rs
```diff
@@ -1,9 +1,60 @@
...
 #[no_mangle]
-pub extern "C" fn add(a: u32, b: u32) -> u32 {
+pub unsafe extern "C" fn print_string(ptr: *const c_char) {
+  let cstr = CStr::from_ptr(ptr);
+  let name = cstr.to_str().unwrap();
+  println!("{}", name);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn print_buffer(ptr: *const u8, len: usize) {
```
ReddikHaien, line 17: I've probably overlooked it, but wouldn't it be necessary with a bound check on the buffer to prevent someone from passing a invalid length for the raw pointer?

ReddikHaien (approved):

ext/ffi/00_ffi.js
eliassjogreen: You're right 👍

ext/ffi/lib.rs
```diff
@@ -75,22 +80,22 @@ impl DynamicLibraryResource {
...
+    let parameter_types =
+      foreign_fn.parameters.into_iter().map(NativeType::from);
+    let result_type = NativeType::from(foreign_fn.result);
     let cif = libffi::middle::Cif::new(
-      foreign_fn
-        .parameters
-        .clone()
-        .into_iter()
-        .map(libffi::middle::Type::from),
-      foreign_fn.result.into(),
```
bartlomieju, line 84: I believe this code threw `TypeError` previously in `dlopenInvalidArguments` but now it only throws `Error`

ext/ffi/lib.rs
```diff
@@ -150,85 +157,99 @@ impl From<NativeType> for libffi::middle::Type {
...
+      NativeType::F32 => Self::F32(value_as_f32(value)),
+      NativeType::F64 => Self::F64(value_as_f64(value)),
+      NativeType::String => Self::String(
+        CString::new(
+          value
+            .as_str()
+            .expect("Expected ffi arg value to be a string"),
+        )
+        .unwrap()
+        .into_raw(),
```
piscisaureus, line 230: This introduces a memory leak in most cases.
My suggestion: remove string support from the FFI interface.
To pass strings, the user can use TextEncoder/TextDecoder to create a Buffer.

ext/ffi/lib.rs
```diff
@@ -372,5 +403,21 @@ fn op_ffi_call(
...
+      json!(unsafe {
+        let ptr = symbol.cif.call::<*const c_char>(symbol.ptr, &call_args);
+        let cstr = CStr::from_ptr(ptr);
+        cstr.to_str().unwrap()
+      })
+    }
+    NativeType::Buffer => {
+      let ptr = unsafe { symbol.cif.call::<*const u8>(symbol.ptr, &call_args) };
+      if let Some(len) = symbol.result_length {
+        let slice = unsafe { std::slice::from_raw_parts(ptr, len) };
```
piscisaureus, line 416: I don't believe this can be used in a safe way without leaking memory.

konsumer:
Anything I can do to help get this in? I am not super-strong with rust, but I have several projects that need mem-buffers, in deno FFI, so I am super-motivated. I can verify this PR works on Windows, Linux, and OSX (all on x86_64.)

ext/ffi/lib.rs
cztomsik: BTW: here's how `ffi-napi` does it, the string leaks too but you can use `finally` to release it
image

cztomsik:
FYI I have added few comments here https://github.com/denoland/deno/pull/11648/files#r713404236
I believe it would be nice to support `*mut c_char` result type even with `into_raw()` leaking memory because you can always wrap the return in `finally { lib.free_string(s) }`

rracariu:
Looking fwd. to get this PR in. But it is stale for a long time.

Can it be simplified to only deal with buffers for now? This (with https://github.com/denoland/deno/pull/12274 merged) would unlock the bare minimum for replacing the plugin API.

kriszyp:
I'll just add another vote that I think the most essential elements for really opening the doors for native/foreign plugins are passing buffers and async support (some way to enqueue a callback on the event queue). Strings are not necessary; maybe a convenience, but JS already has facilities for string encoding/decoding that work fine. To me, adding support for buffers & async to FFI is the most critical/valuable outstanding feature/issue in Deno.

rotu:
I think allowing strings also opens up another can of worms. As shown in #12226, a Javascript DOMString can legally contain lone surrogates. This is okay in C++ and Python, but not in Rust and Swift. Buffers don't have this issue.

bartlomieju:
I forked off this PR to https://github.com/denoland/deno/pull/12335, we're gonna support only buffer-as-arg in the first pass

rracariu:
> @rracariu @kriszyp @cztomsik @rotu buffer-as-arg is now available on canary, see #12335 for details. I'll work on documentation as per [denoland/manual#106](https://github.com/denoland/manual/issues/106)

Without return buffers one would not be able to return complex structures, which severely limits the things you could do.

In C there is no `buffer` type - so strictly speaking this concept is mapped to `u8*` by Deno, but that is a convention.
So this means that the `buffer` type could expand on this convention until proper pointer types (in/out) are supported by the FFI interface.

What I mean by that is that the FFI return type could be a `buffer_return` FFI type which is a `struct {size_t size; size_t ptr}` that is mapped as a `FfiBuffer` type in Deno/Typescript. This `FfiBuffer` would have the byte contents for the pointer, but expose the pointer value as well, like `interface FfiBuffer {readonly data: ArrayBuffer; readonly ptr: BigInt | number }`. Using the `ptr` value one could call a `free` function to free the memory over in the FFI land.

wobsoriano:
Is it possible to return strings as value now?

cztomsik:
You can `return CString::new("foo").unwrap().into_raw()` and then in JS `new UnsafePointerView(ptr).getCString()` to decode that string but you also have to free that memory eventually or it will leak.

Either you need extra ffi function which will free the pointer when you're done or you can reuse the same (mutex or thread-local) CString and kind of "reuse" it every time and never really free it

wobsoriano:
> You can `return CString::new("foo").unwrap().into_raw()` and then in JS `new UnsafePointerView(ptr).getCString()` to decode that string but you also have to free that memory eventually or it will leak.
> 
> Either you need extra ffi function which will free the pointer when you're done or you can reuse the same (mutex or thread-local) CString and kind of "reuse" it every time and never really free it

That worked! Thanks. Didn't know `UnsafePointerView` returns methods

cztomsik:
I found out recently as well, inspecting the test cases actually :)
https://github.com/denoland/deno/tree/main/test_ffi

wobsoriano:
About freeing the memory, can we do it before returning a value? Something like

```ts
function getCString(pointer: Deno.UnsafePointer) {
  const unsafe = new Deno.UnsafePointerView(pointer);

  setTimeout(() => {
    // if I don't wrap it in a timeout, I get invalid utf-8 sequence of 1 bytes from index 0
    sys.symbols.freeString(pointer);
  }, 0);

  return unsafe.getCString();
}
```

Issue #11645, opened 2021-08-11
Quirks in "deno test --doc" in Markdown files


bartlomieju:
1. "\`\`\`ts, ignore" works, but "\`\`\`ts ignore" doesn't work. The former break syntax highlighting in VSCode, so we should prefer the latter version
2. Running "deno test --doc" on a Markdown file that has no codeblock results in error, eg:
`deno test --doc getting_started.md` results in:
```
error: Unexpected token `#`. Expected this, import, async, function, [ for array literal, { for object literal, @ for decorator, function, class, null, true, false, number, bigint, string, regexp, ` for template literal, (, or an identifier at file:///Users/biwanczuk/dev/manual/getting_started.md:1:0
```

The file I'm running is:
```
# Getting Started

In this chapter we'll discuss:

- [Installing Deno](./getting_started/installation.md)
- [Setting up your environment](./getting_started/setup_your_environment.md)
- [Running a `Hello World` script](./getting_started/first_steps.md)
- [Writing our own script](./getting_started/first_steps.md)
- [Command line interface](./getting_started/command_line_interface.md)
- [Understanding permissions](./getting_started/permissions.md)
- [Using WebAssembly](./getting_started/webassembly.md)
- [Debugging your code](./getting_started/debugging_your_code.md)
```

PR #11652, opened 2021-08-11
fix(cli): allow specifiers of unknown media types with test command

caspervonb:
This makes it so that `deno test foo.md` does not end up trying to run foo.md as a module and it is no longer an error to pass unknown media types as input to `deno test`.

They'll get considered for the inline extraction phase and dropped for the isolate execution phase.

See https://github.com/denoland/deno/issues/11645


cli/tests/testdata/test/test/markdown.md
```diff
@@ -0,0 +1,25 @@
...
+The following block should be given a js extension on extraction:
+
+```js
+console.log("js");
+```
+
+The following block should be given a ts extension on extraction:
+
+```ts
+console.log("ts");
```
bartlomieju, line 18: Is this file actually used anywhere?

bartlomieju (commented):
Please add a plain Markdown file that has no codeblocks inside and verify that no error is returned and no tests are executed

cli/tests/testdata/test/test/markdown.md
caspervonb: test/test... bad merge.

bartlomieju (approved):
LGTM, thanks

PR #11653, opened 2021-08-11
feat: allow running on 32 bits 

PerfectLaugh:
I find current rusty_v8 is able to compile in 32bits. It would be good to add supports in 32 bit.
Only tested in Windows currently.

For reference:
https://github.com/denoland/rusty_v8/issues/426#issuecomment-678696428

---
AaronO leaves a good review on the direction of this pull request.
https://github.com/denoland/serde_v8/pull/16#pullrequestreview-728458540

We need more platform tests on this issue if we want to release in 32 bits. If not, it should be okay to provide a way compiling to 32 bits also.
Currently tested in my Windows desktop and Linux VM.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,2 +1,5 @@
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
+
+[target.i686-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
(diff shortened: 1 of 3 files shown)


crowlKats:
As far as I know, GH Actions does not have any 32 bit systems, so we cannot test nor make release binaries for this

PerfectLaugh:
And gn binaries do not have i686 architecture either. But it allows to cross compile through CFLAGS

PerfectLaugh:
image
Tested in Linux also.

piscisaureus:
@PerfectLaugh is the thiscall calling convention now supported in rust?

PerfectLaugh:
@piscisaureus I found there's simply no need to include thiscall. The bindings are turned into ```extern "C"``` symbols and no need to assign thiscall except there are changes to callbacks for SystemV i386 ABI.
https://github.com/denoland/rusty_v8/pull/752

piscisaureus:
@PerfectLaugh I don’t believe that, but we may be missing a test case.

brandonros:
Could this be revisited? I need it to do FFI with 32-bit-only DLLs

PR #11660, opened 2021-08-11
WIP: network panel support for DevTools

lucacasonato:
Let's get `fetch` and `WebSocket` hooked up to the "Network" panel in Chrome DevTools.


runtime/worker.rs
```diff
@@ -158,13 +158,16 @@ impl MainWorker {
...
+      let dev_tools_agent =
+        js_runtime.op_state().borrow().dev_tools_agent.clone();
       let inspector = js_runtime.inspector();
       let session_sender = inspector.get_session_sender();
       let deregister_rx = inspector.add_deregister_handler();
       server.register_inspector(
         session_sender,
         deregister_rx,
         main_module.to_string(),
+        dev_tools_agent,
```
caspervonb, line 170: So unless I missed something we end up the minor caveat here that local sessions won't support the same extended subset as remote sessions get?

bartlomieju: This is just a prototype that will be rewritten. I guess you're still referring to the ability to connect local REPL session to the Devtools?

caspervonb: > This is just a prototype that will be rewritten. I guess you're still referring to the ability to connect local REPL session to the Devtools?

Yeah it's low hanging fruit but hypothetically we'd want to expose the local inspector as as a cli binding ala node down the line.

bartlomieju:
Stale, closing for now, I'll try to revisit this feature in Q1 2022

PR #11662, opened 2021-08-12
fix(ext/crypto): importKey() SecurityError on non-extractable keys

### LINKED ISSUE #11650: SubtleCrypto.importKey ignores the value of the `extractable` property
see [importKey](https://github.com/denoland/deno/blob/v1.13.0/extensions/crypto/00_crypto.js#L441-L540), in its `constructKey` HMAC call the value `extractable` argument is hardcoded to `true`, disregarding the value of the argument passed in.

https://github.com/denoland/deno/blob/v1.13.0/extensions/crypto/00_crypto.js#L519-L525

refs #11481

littledivy:
Fixes #11650


lucacasonato:
We now have no more WPT for HMAC key importing (all test use `extractable: false`). Would you mind adding one in cli/tests/unit?

lucacasonato (approved):
LGTM, thanks.

PR #11666, opened 2021-08-12
fix(ext/crypto): take a copy of keyData bytes

### LINKED ISSUE #11664: `crypto.subtle.sign` does not accept `ArrayBuffer` based `CryptoKey`
The `keyData` for `crypto.subtle.importKey()` is typed as `BufferSource` but if an `ArrayBuffer` is used, and then the returned `CryptoKey` is used with `crypto.subtle.sign()` the following error is returned:

```
TypeError: Error parsing args: serde_v8 error: ExpectedArray
    at Object.opAsync (deno:core/01_core.js:135:28)
    at SubtleCrypto.sign (deno:extensions/crypto/00_crypto.js:419:40)
```

According to the spec, `ArrayBuffer` based keys should be valid for signing.

littledivy:
Fixes #11664

### DIFF
--- a/cli/tests/unit/webcrypto_test.ts
+++ b/cli/tests/unit/webcrypto_test.ts
@@ -1,5 +1,41 @@
 import { assert, assertEquals, unitTest } from "./test_util.ts";
 
+// https://github.com/denoland/deno/issues/11664
+unitTest(async function testImportArrayBufferKey() {
+  const subtle = window.crypto.subtle;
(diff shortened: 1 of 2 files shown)


bartlomieju:
@littledivy looks good to me, but can we get a test from #11664 that exercises this change?

lucacasonato (approved):
LGTM, thanks!

Issue #10948, opened 2021-06-14
run_watch is flakey on windows


kitsonk:
The run_watch appears to be flaky on windows: https://github.com/denoland/deno/runs/2807326966#step:31:2276

ry:
It's also flaky on mac: https://github.com/denoland/deno/runs/3297697099

```

failures:

---- integration::watcher::run_watch stdout ----
Error: thread 'integration::watcher::run_watch' panicked at 'assertion failed: stderr_lines.next().unwrap().contains(\"error:\")', cli/tests/integration/watcher_tests.rs:306:3
error: test failed, to rerun pass '-p *** --test integration_tests'
stack backtrace:
   0:        0x10c854963 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h6a8908fa3ed6f9e8
   1:        0x10c70315c - core::fmt::write::h4be00f71c5582919
   2:        0x10c85419a - std::io::Write::write_fmt::h93cbb8c6ac53b1a6
   3:        0x10c853a9f - std::panicking::default_hook::***closure***::h1577f0656e419c0e
   4:        0x10c852c46 - std::panicking::rust_panic_with_hook::h10bc487d002f6c42
   5:        0x10c870f32 - std::panicking::begin_panic_handler::***closure***::hf4cfa78c105ce648
   6:        0x10c870ec8 - std::sys_common::backtrace::__rust_end_short_backtrace::h1df96a166e4351c4
   7:        0x10c870e70 - _rust_begin_unwind
   8:        0x10cb7a50f - core::panicking::panic_fmt::hea8fe6c9e0720810
   9:        0x10cb7a467 - core::panicking::panic::hbb13bbcfbaa2890b
  10:        0x10c93a617 - core::ops::function::FnOnce::call_once::h795287eb32da06b9
  11:        0x10c8aa52a - test::__rust_begin_short_backtrace::h8537766c178d3dff
  12:        0x10c8aa7f1 - test::run_test::run_test_inner::***closure***::hef20bc7b3073ef89
  13:        0x10c8ac251 - std::sys_common::backtrace::__rust_begin_short_backtrace::hbedd640dcb83f549
  14:        0x10c8ac0a2 - core::ops::function::FnOnce::call_once***vtable.shim***::hbeed65ddf37019e1
  15:        0x10c886f5a - std::sys::unix::thread::Thread::new::thread_start::h63f8b299c7b9d50c
[...]

PR #11668, opened 2021-08-12
test: mark run_watch test as flaky

bartlomieju:
Ref https://github.com/denoland/deno/issues/10948


ry (approved):
LGTM - but the issue should remain open - the test is still flaky. I changed "closes" to "ref"

PR #11669, opened 2021-08-12
test: re-enable test watch tests and mark as flaky

caspervonb:
This marks test watcher tests as flaky tests.
They still need to be re-worked as they are very flaky.

See https://github.com/denoland/deno/issues/10552

### DIFF
--- a/cli/tests/integration/watcher_tests.rs
+++ b/cli/tests/integration/watcher_tests.rs
@@ -1,5 +1,6 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
+use flaky_test::flaky_test;
 use std::io::BufRead;
 use tempfile::TempDir;
 use test_util as util;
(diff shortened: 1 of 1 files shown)


cli/tests/integration/watcher_tests.rs
```diff
@@ -427,9 +427,8 @@ fn run_watch_with_import_map_and_relative_paths() {
   temp_directory.close().unwrap();
 }
 
-// TODO(bartlomieju): flaky (https://github.com/denoland/deno/issues/10552)
-#[ignore]
 #[test]
```
bartlomieju, line 430: This should be removed

cli/tests/integration/watcher_tests.rs
```diff
@@ -596,9 +595,8 @@ fn test_watch() {
   drop(t);
 }
 
-// TODO(bartlomieju): flaky (https://github.com/denoland/deno/issues/10552)
-#[ignore]
 #[test]
```
bartlomieju, line 598: Ditto

bartlomieju (changes requested):

cli/tests/integration/watcher_tests.rs
caspervonb: Huh, this does not match my working tree... did something weird here.

bartlomieju (approved):
Provisional LGTM, if it fails on `main` I will revert it

PR #11670, opened 2021-08-12
build: add devcontainer configuration

lucacasonato:
This allows users to use Codespaces and VS Code Remote Containers easially.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/vscode/devcontainers/rust:0-1
+
+ENV DENO_INSTALL=/usr/local
+RUN curl -fsSL https://deno.land/x/install/install.sh | sh
```
kitsonk, line 4: missing a return?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
+      "**/target/**": true
+    }
+  },
+
+  "extensions": [
+    "matklad.rust-analyzer",
+    "bungcip.better-toml",
+    "vadimcn.vscode-lldb",
+    "mutantdino.resourcemonitor",
+    "denoland.vscode-deno"
```
kitsonk, line 21: Do you really want this one installed?  It becomes a little problematic of having a version built in the container? Also, it doesn't do so well with internal code, I have `"deno.enable": false` for my vscode for Deno.  Maybe consider at least setting up some default `deno.` settings in the `devcontainer.json`?

kitsonk (approved):
Couple thoughts, but awesome.

kitsonk:
The other thing to note, which I haven't figured out, is that this will install Rust 1.53 at the moment.  I haven't figure out how to get Rust 1.54.

.devcontainer/devcontainer.json
lucacasonato: Removed.

lucacasonato:
@kitsonk I set it up to install Rust 1.54 through `rustup` now.

PR #11675, opened 2021-08-12
fix(http/ws): support multiple options in connection header

crowlKats:
Continuation of #11505


lucacasonato:
Can we get a test for this?

lucacasonato (approved):
Thanks @crowlKats.

Regarding the liberal parsing of the headers for `is_websocket_request`: it doesn't have to be exact. It is really just part of a performance optimisation. If it has some false positives, the worst that can happen is a little bit of a perf decrease. No big deal.

PR #11676, opened 2021-08-12
refactor(cli/tools/repl): extract prelude into file

caspervonb:

### DIFF
--- a/cli/tools/repl.rs
+++ b/cli/tools/repl.rs
@@ -387,37 +387,6 @@ impl ReplEditor {
   }
 }
 
-static PRELUDE: &str = r#"
-Object.defineProperty(globalThis, "_", {
-  configurable: true,
-  get: () => Deno[Deno.internal].lastEvalResult,
-  set: (value) => {
-   Object.defineProperty(globalThis, "_", {
-     value: value,
-     writable: true,
-     enumerable: true,
-     configurable: true,
-   });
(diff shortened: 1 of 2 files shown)


bartlomieju:
I don't have a strong opinion one way or the other. I guess the only advantage is that code can now be formatted. I'll leave it to @dsherret to land it.

cli/tools/repl.rs
```diff
@@ -474,7 +443,8 @@ impl ReplSession {
     };
 
     // inject prelude
-    repl_session.evaluate_expression(PRELUDE).await?;
+    let prelude = include_str!("repl.js");
```
dsherret, line 446: I actually prefer it inline so that we don't have a js file in the tools directory. I don't think code formatting this is that important.

caspervonb: > we don't have a js file in the tools directory

Why do you find that undesirable tho?

dsherret: Instead of this detail being buried it's presented at a higher level right inside the directory as if it were on that same level of importance and execution. I'd prefer it buried within the file because it's not that important and the code is not that large. If you really want this, I think it would be better to have a sub repl folder rather than both of these files being at the top level of tools. Sorry for being picky...

caspervonb: That's alright, was just curious 🙂
As it is **right now** I don't care that much, I'll punt this until it grows.

PR #11679, opened 2021-08-13
refactor(cli): generalize module specifier collection

caspervonb:

### DIFF
--- a/cli/fs_util.rs
+++ b/cli/fs_util.rs
@@ -427,9 +427,11 @@ mod tests {
     )
     .unwrap();
 
-    let root_dir_url = ModuleSpecifier::from_file_path(root_dir_path)
-      .unwrap()
-      .to_string();
+    let root_dir_url = ModuleSpecifier::from_file_path(
+      canonicalize_path(&root_dir_path).unwrap(),
+    )
+    .unwrap()
+    .to_string();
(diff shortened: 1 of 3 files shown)


cli/fs_util.rs
```diff
@@ -3,6 +3,7 @@
 use deno_core::error::AnyError;
 use deno_core::error::Context;
 pub use deno_core::normalize_path;
+use deno_core::url::Url;
```
kitsonk, line 6: I would prefer we use the `ModuleSpecifier` type alias when we are going to refer to something that we expect to be a module specifier.

caspervonb: sgtm :)

cli/fs_util.rs
```diff
@@ -328,4 +372,82 @@ mod tests {
...
+    // │   ├── .foo.TS
+    // │   └── README.md
+    // ├── c.tsx
+    // ├── d.jsx
+    // └── ignore
+    //     ├── g.d.ts
+    //     └── .gitignore
+
+    let t = TempDir::new().expect("tempdir fail");
+
```
bartlomieju, line 401: Nice test

bartlomieju (approved):
LGTM, good cleanup

PR #11683, opened 2021-08-13
refactor(cli): simplify inline test collection

caspervonb:
Removes a bunch of branching used to maintain two input vectors (one for tests and one for documentation tests) in favour of a single unified one.


caspervonb:
A very subtle potential behavioural change here.

If someone was to define tests inline (which is not really supported nor recommended) in a module called `foo.js` and `deno test --doc` was invoked, then those tests would be executed because the predicate is expanded to include all relevant files that may contain documentation tests.

bartlomieju:
> There's a subtle behavioural change here.
> 
> If someone was to define tests inline (which is not really supported nor recommended) in a module called `foo.js` and `deno test --doc` was invoked, then those tests would be executed because the predicate is expanded to include all relevant files that may contain documentation tests.

I'm not following, could you give an example?

caspervonb:
Drafting this for a minute as I'm unsure if I want this behaviour.

> I'm not following, could you give an example?

```console
# We have two files, mod.js and mod_test.js
$ ls
mod.js		mod_test.js

# mod.js contains a test, even tho it usually does not get picked up by the test runner.
$ cat mod.js 
Deno.test("test", function() {});

# mod_test.js also contains a test, nothing unusual will happen to this.
$ cat mod_test.js 
Deno.test("test", function() {});

# When we run this with stable deno, only the tests from mod_test.js will get picked up.
$ deno-stable test --doc
running 1 test from file:///Users/caspervonb/deno/foo/mod_test.js
test test ... ok (72ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (608ms)

# But this patch, will pick up both files because mod.js is no longer tracked separately as "just look for documentation"
# So it will run the tests if any in this file.
$ deno-local test --doc
running 1 test from file:///Users/caspervonb/deno/foo/mod.js
test test ... ok (72ms)

running 1 test from file:///Users/caspervonb/deno/foo/mod_test.js
test test ... ok (72ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (608ms)
```

caspervonb:
Closing out, doing a more extensive refactor of the watcher.

PR #11654, opened 2021-08-11
feat(ext/crypto): implement encrypt, decrypt & generateKey for RSA-OAEP

littledivy:
Towards #11690

PR #11685, opened 2021-08-13
fix(ext/crypto): handle idlValue may not be present

littledivy:
`10.1.3` under [`18.4.4` Normalizing an algorithm](https://www.w3.org/TR/WebCryptoAPI/#algorithm-normalization-normalize-an-algorithm). Needed for #11654


lucacasonato (approved):
LGTM, thanks.

PR #11688, opened 2021-08-13
Add failing byte_order_mark test

ry:


cli/ast/mod.rs
```diff
@@ -392,10 +393,15 @@ pub fn parse(
   source: &str,
   media_type: &MediaType,
 ) -> Result<ParsedModule, AnyError> {
+  let source = strip_bom(source);
```
dsherret, line 396: This doesn't fix the entire issue. I wonder if we should be doing this at a higher level?

ry: Is it V8 that is barfing on it?

dsherret: Yes. In `core::modules::ModuleMap::new_module` `module_request.get_specifier()`. I haven't looked into it too much.

It's strange that `cargo test` passes yet running it on the command line fails. I just did a `cargo clean && cargo build && ./target/debug/deno run --no-check cli/tests/testdata/byte_order_mark.ts` and it fails.

dsherret: Wow, that was weird. I looked into this again just now and ran:

```ts
./target/debug/deno run cli/tests/testdata/byte_order_mark.ts
```

...then it suddenly started working again next time I provided `--no-check`. Caching issue? Anyway, this should be good to go.

dsherret (approved):

PR #11693, opened 2021-08-13
cleanup(ext/web/BlobStore): avoid redundant alloc

AaronO:
`Arc<T>` is already on the heap so `Arc<Box<T>>` is unnecessary, see: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_allocation

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -968,7 +968,7 @@ mod tests {
     let specifier = blob_store.insert_object_url(
       Blob {
         media_type: "application/typescript".to_string(),
-        parts: vec![Arc::new(Box::new(InMemoryBlobPart::from(bytes)))],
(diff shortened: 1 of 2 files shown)


bartlomieju:
Why didn't `clippy` complain about it before :( ? Maybe we should add config file for clippy to be more strict?

AaronO:
> Why didn't `clippy` complain about it before :( ? Maybe we should add config file for clippy to be more strict?

It complains on the `nightly` toolchain (`1.56.0`), we're currently on `stable` (`1.54.0`)

bartlomieju:
> > Why didn't `clippy` complain about it before :( ? Maybe we should add config file for clippy to be more strict?
> 
> It complains on the `nightly` toolchain (`1.56.0`), we're currently on `stable` (`1.54.0`)

That's good enough then.

bartlomieju (approved):

Issue #7452, opened 2020-09-13
--watch should send unload event before restarting


lucacasonato:
```js
console.log("start");

// Just here to block execution
setInterval(() => {}, 100);

window.addEventListener("unload", () => {
  console.log("unload");
});
```

Should print
```
Check file:///mnt/f9/Projects/github.com/denoland/deno/test.ts
start
unload
Watcher File change detected! Restarting!
Check file:///mnt/f9/Projects/github.com/denoland/deno/test.ts
start
```
but it prints
```
Check file:///mnt/f9/Projects/github.com/denoland/deno/test.ts
start
Watcher File change detected! Restarting!
Check file:///mnt/f9/Projects/github.com/denoland/deno/test.ts
start
```

bartlomieju:
Already there: 
https://github.com/denoland/deno/blob/b3fa81f867f933df41f274bd8eec10dddf1b1b28/cli/main.rs#L508-L510

If it's not working then it's definitely a bug...

caspervonb:
Picking this up tomorrow.

PR #11696, opened 2021-08-13
fix(cli): dispatch unload event on watch drop

caspervonb:
Fixes https://github.com/denoland/deno/issues/7452


cli/main.rs
```diff
@@ -887,29 +887,70 @@ async fn run_with_watch(flags: Flags, script: String) -> Result<(), AnyError> {
...
+      ModuleExecutor {
+        worker,
+        pending_unload: false,
+      }
+    }
+
+    pub async fn execute(
+      &mut self,
+      main_module: &ModuleSpecifier,
+    ) -> Result<(), AnyError> {
```
bartlomieju, line 906: This is a good solution, but please add some docstrings here - it's not immediately obvious why this is needed. I'd also suggest a change of name, maybe `FileWatcherModuleExecutor`?

caspervonb: > FileWatcherModuleExecutor

As for the naming, wasn't the original intent but I'm thinking this can be a more generic thing.
Think it might be useful to hoist this in a follow up and use it everywhere where we do this lifecycle sequence which typically looks like:
- maybe_start_coverage_collection
- execute_module
- dispatch `load`
- run_event_loop
- dispatch 'unload'
- maybe_stop_coverage_collection

We repeat these steps in a few places that do the typical execution cycle.
But I'm leaving that for a (somewhat immediate) follow-up, first pass I'm just about fixing this bug.

bartlomieju:
@caspervonb please rebase, I think we'll land it for 1.13.2

bartlomieju (approved):
LGTM, I'll wait for Luca's review before landing

lucacasonato (approved):

Issue #10998, opened 2021-06-16
REPL syntax highlighting is quirky


bartlomieju:
Example:
![Screenshot 2021-06-16 at 03 44 21](https://user-images.githubusercontent.com/13602871/122145436-8abb3a00-ce55-11eb-851e-de0a208ab5a5.png)

Only `Infinity`, `Nan` and `undefined` are colorized. All the other symbols are displayed in default font.

caspervonb:
Intentional, same colorization as console.

PR #11697, opened 2021-08-14
fix(cli/tools/repl): dont highlight candidate when completion is list

caspervonb:
This disables any highlighting when the completion mode is set to list.

Fixes https://github.com/denoland/deno/issues/10998


bartlomieju (approved):
LGTM, thanks for fixing that

PR #11700, opened 2021-08-14
fix(cli): retain input order of remote specifiers

caspervonb:
Specifier collection partitions remote specifiers in their own group which is appended to the collected specifiers at the end of the routine meaning that the input order isn't respected for remote specifiers.

This retains the input order so that remote tests run in the order they're listed.


bartlomieju (approved):
LGTM, good fix

PR #11702, opened 2021-08-14
fix(cli/lint): don't use gray in diagnostics output for visibility

magurotuna:
This commit changes output color of the linter to not use gray for visibility on dark backgrounds.

Fixes https://github.com/denoland/deno_lint/issues/788

### DIFF
--- a/cli/tools/lint.rs
+++ b/cli/tools/lint.rs
@@ -301,27 +301,25 @@ pub fn format_diagnostic(
     }
   }
 
-  if let Some(hint) = maybe_hint {
-    format!(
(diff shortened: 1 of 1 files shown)


cli/tools/lint.rs
```diff
@@ -308,9 +307,9 @@ pub fn format_diagnostic(
       message_line,
```
bartlomieju, line 307: While you looking into it, could you rewrite this conditional? Lots of duplication and not very readable.

magurotuna: Sure, how about this? 99a83ff

bartlomieju: That works, thanks!

bartlomieju (approved):
LGTM

PR #11704, opened 2021-08-14
cleanup(ext/http): simplify cookie header handling

AaronO:
Use `Vec::join` instead of essentially reimplementing it, there should be no meaningful performance delta

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -293,7 +293,7 @@ fn req_headers(
   // mangled by the `Headers` object in JS. What we do is take all cookie
   // headers and concat them into a single cookie header, seperated by
   // semicolons.
-  let cookie_sep = &b"; "[..];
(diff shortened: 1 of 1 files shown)


ext/http/lib.rs
```diff
@@ -293,15 +293,13 @@ fn req_headers(
   // mangled by the `Headers` object in JS. What we do is take all cookie
   // headers and concat them into a single cookie header, seperated by
   // semicolons.
-  let mut total_cookie_length = 0;
+  let cookie_sep = &b"; "[..];
```
satyarohith, line 296: Is it possible to use `";".as_bytes()`?

satyarohith (approved):
LGTM.

ext/http/lib.rs
AaronO: Yep, that works too. Made the change

Issue #11690, opened 2021-08-13
WebCrypto roadmap


lucacasonato:
Spec: https://w3c.github.io/webcrypto

This is the global tracking issue for WebCrypto API implementation.

See the status of the entire API surface:

- [x] Enable non-extractable keys: #11481
- [x] `SubtleCrypto.encrypt`
  - [x] `RSA-OAEP`
  - [x] `AES-CTR` https://github.com/denoland/deno/pull/13177
  - [x] `AES-CBC`
  - [x] `AES-GCM`
- [x] `SubtleCrypto.decrypt`
  - [x] `RSA-OAEP`
  - [x] `AES-CTR` https://github.com/denoland/deno/pull/13177
  - [x] `AES-CBC`
  - [x] `AES-GCM`
- [x] `SubtleCrypto.sign`
  - [x] `RSASSA-PKCS1-v1_5`
  - [x] `RSA-PSS` (all hashes)
  - [x] `RSA-PSS` (all hashes)
  - [x] `ECDSA`
    - [x] `P256`
    - [x] `P384`
    - [ ] `P521`
  - [x] `HMAC`
- [x] `SubtleCrypto.verify`
  - [x] `RSASSA-PKCS1-v1_5`
  - [x] `RSA-PSS` (all hashes)
  - [x] `RSA-PSS` (all hashes)
  - [x] `ECDSA`
    - [x] `P256`
    - [x] `P384`
    - [ ] `P521`
  - [x] `HMAC`
- [x] `SubtleCrypto.digest`
  - [x] `SHA-1`
  - [x] `SHA-256`
  - [x] `SHA-384`
  - [x] `SHA-512`
- [x] `SubtleCrypto.generateKey`
  - [x] `RSASSA-PKCS1-v1_5`
  - [x] `RSA-PSS`
  - [x] `RSA-PSS`
  - [x] `RSA-OAEP` 
  - [x] `ECDSA`
    - [x] `P256`
    - [x] `P384`
    - [ ] `P521`
  - [x] `ECDH` #11870
    - [x] `P256`
    - [x] `P384`
    - [ ] `P521`
  - [x] `AES-CTR` #11869
  - [x] `AES-CBC` #11869
  - [x] `AES-GCM` #11869
  - [x] `AES-KW` #11869
  - [x] `HMAC`
- [x] `SubtleCrypto.deriveKey`
  - [x] `ECDH`
  - [x] `HKDF`
  - [x] `PBKDF2`
[...]

PR #11705, opened 2021-08-14
fix(ext/crypto): enable non-extractable keys

### LINKED ISSUE #11481: Enable non-extractable keys for Web Crypto

littledivy:
Closes #11481 

Towards #11690


panva (approved):

lucacasonato (approved):
Very awesome. So many more passing tests :tada:

Thanks @littledivy!

Issue #11701, opened 2021-08-14
Test does not dispatch load event


caspervonb:
```ts
window.addEventListener("load", () => {
  console.log("load");
});

window.addEventListener("unload", () => {
  console.log("unload");
});
```

```sh
deno test
```

```ts
running 0 tests from file:///home/caspervonb/test.js
unload

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (31ms)
```

caspervonb:
I probably introduced this regression in https://github.com/denoland/deno/pull/11515.
Load is now fired AFTER tests have run.

PR #11708, opened 2021-08-14
fix(cli): dispatch load event before tests are run

caspervonb:
Fixes https://github.com/denoland/deno/issues/11701


bartlomieju (approved):
LGTM, good fix

PR #11710, opened 2021-08-14
refactor(ops): return BadResource errors in ResourceTable calls

AaronO:
Instead of relying on callers to map Options to Results via `.ok_or_else(bad_resource_id)` at over 176 different call sites ...


bartlomieju (approved):
LGTM, no comments

ry:
Nice clean up

Issue #10516, opened 2021-05-06
Support fetch TLS client certificate


iugo:
Just like `curl --cert <file.pem>`.

Now the way to solve this problem in deno is: curl. But this method is not cross-platform.

DjDeveloperr:
I think https://doc.deno.land/builtin/unstable#Deno.createHttpClient works for that.

lucacasonato:
We already support this with createHttpClient and the --cert flag.

danopia:
Actually no? Curl's `--cert` is for _client certificates_ and Deno's `--cert` is for _certificate authorities_ so the naming is pretty confusing. But afaik Deno does not offer what this issue asks for: TLS client certificates in `fetch()`.

deno:
```
        --cert <FILE>                  
            Load certificate authority from PEM encoded file
```

curl:
```
     --cacert <file> CA certificate to verify peer against

 -E, --cert <certificate[:password]> Client certificate file and password
     --cert-type <type> Certificate file type (DER/PEM/ENG)
```

lucacasonato:
You're right. Reopened.

justinmchase:
Proposal:
```ts
interface CreateHttpClientOptions {
  // A certificate authority to use when validating TLS certificates. Certificate data must be PEM encoded.
  ca?: string
  // Client certificate chains in PEM format. 
  cert?: string
  // Client certificate private keys in PEM format.
  key?: string

  proxy?: Proxy
}
```

These names are mimicking the nodejs api for tls options:
https://nodejs.org/docs/latest-v15.x/api/tls.html#tls_tls_createsecurecontext_options

You can see there are a ton of other options but most of the time if you're using client certificates you just need the `cert` and `key` fields. This will help a lot of usecases without having to go nuts implementing every option under the sun.

justinmchase:
Related to https://github.com/denoland/deno/issues/9121

wesleyakio:
Hey guys, I spent a few hours reading through the issues and it seems like this landed on main about two years ago in #11721 but I couldn't find anything definitive as to why this is still behind `--unstable`. Cursory tests show that this works for my purposes but I'd love to also use this on Deno Deploy so understanding what kind of timeframe are we working with for this to land on `stable` would be very helpful. Anyone around could shed some light?

PR #11721, opened 2021-08-16
feat(fetch): TLS client certificates (mutual authentication) for fetch()

SeanWykes:
Implementation of TLS client-certificate for fetch(), largely based on work by @eric in #9426 - so all kudos to him for the hard work. 

Needs patching to put feature behind the --unstable flag, see ext/fetch/lib.rs:547 for commented (incorrect) code

Implements #10516.


littledivy (commented):
Is updating the `test_util/wpt` submodule intentional?

lucacasonato (commented):
@SeanWykes Thanks for the PR. Deno.createHttpClient is already unstable, so there doesn't need to be any additional checks for `certChain` and `privateKey`.

ext/tls/lib.rs
```diff
@@ -164,12 +217,32 @@ pub fn create_http_client(
   ca_data: Option<Vec<u8>>,
   proxy: Option<Proxy>,
   unsafely_ignore_certificate_errors: Option<Vec<String>>,
+  cert_chain: Option<String>,
+  private_key: Option<String>,
```
lucacasonato, line 221: Let's turn this into a single option. It will let you remove some of the checks further down, and instead move them to the create http client op.

```suggestion
  client_cert_chain_and_key: Option<(String, String)>,
```

SeanWykes:
> Is updating the `test_util/wpt` submodule intentional?

@littledivy no, it's a mistake. How can I remove it from the PR?

ext/tls/lib.rs
```diff
@@ -156,6 +162,52 @@ pub fn create_client_config(
   Ok(tls_config)
 }
 
+fn load_certs(reader: &mut dyn BufRead) -> Result<Vec<Certificate>, AnyError> {
```
lucacasonato, line 165: These functions are all copied over from `ext/net`. Can you export them in `ext/tls` and remove them from `ext/net` (instead importing from `ext/tls`)?

SeanWykes: Yep, can do! Should I bump tls version (0.1.1 -> 0.1.2) and update net's dependency on tls?

lucacasonato: Nope, not necessary (but thanks!). We'll do the releasing + publishing just before the next release, during the release process :-)

lucacasonato (approved):
LGTM! Thanks a lot @SeanWykes, great feature to have.

SeanWykes:
@lucacasonato thanks for the help and suggestions. And of course, thanks to @eric who did the real work in getting TLS client-certificates in.
It's been an interesting experience contributing to deno, and constantly switching between ts and rs.
Great project, guys!

Issue #11720, opened 2021-08-15
WebIDL brand symbol shouldn't be exposed to user code


CrimsonCodes0:
The `webidl.brand` symbol is a symbol on Deno's WebIDL API instances. Symbols are observable to any user script.

This allows anyone to simply take it, and bypass any checks, e.g.
```js
// gets sybmols and strings
const webidl_brand = Reflect.ownKeys(performance)[0];

const my_performance = {
  __proto__: Performance.prototype,
  [webidl_brand]: webidl_brand,
};

console.log(my_performance.now());
```

Should this be allowed?

A `webidl.branded` WeakSet would be much more secure, in that it would ensure encapsulation.

crowlKats:
This applies for any other symbol that we use for internal slots as well, not just webidl brands

CrimsonCodes0:
> This applies for any other symbol that we use for internal slots as well, not just webidl brands

Absolutely, this was just an example of a particular symbol. But is this intended to be a "feature," or is it more of a limitation of the implementation?

littledivy:
For the past year, we've been porting APIs to Rust using V8 object wrap. This keeps the implementation details such as webidl and internal symbols hidden from JavaScript. The port is part of a larger effort to move snapshot heavy JS internals to Rust. 

The original repro is fixed:
```
Uncaught TypeError: Illegal invocation
    at Module.assertBranded (ext:deno_webidl/00_webidl.js:1134:11)
    at Performance.now (ext:deno_web/15_performance.js:775:12)
    at <anonymous>:10:28
```

PR #11722, opened 2021-08-16
fix(ext/webidl): use WeakSet for branding

littledivy:
Ref #11720


lucacasonato (commented):
I feel this is rather pointless if we don't do the same for all other internal symbols either (not convinced this is necessary).

PR #11727, opened 2021-08-16
feat(cli): close test worker once all tests complete

caspervonb:
This changes the execution model of the test runner to drop after all test have completed.

Supercedes https://github.com/denoland/deno/pull/11262


caspervonb:
An alternative to this would be to force cleanup:

- Dispatch load.
- Run tests.
- Dispatch unload.
- Error like a sanitizer if there are still pending items in the event loop.

caspervonb:
On second thought, I'm okay with unload and drop, its the same behaviour one would get from calling `close`.

core/runtime.rs
```diff
@@ -647,6 +647,48 @@ impl JsRuntime {
...
 
+  /// Waits for the given value to resolve while polling the event loop.
+  ///
+  /// This future resolves when either the value is resolved or the event loop runs to
+  /// completion.
+  pub async fn resolve_value(
+    &mut self,
+    global: v8::Global<v8::Value>,
+  ) -> Result<v8::Global<v8::Value>, AnyError> {
+    poll_fn(|cx| {
```
bartlomieju, line 658: Yeah I find it useful, especially in the context of this PR. We don't use this pattern much, but I can see it being quite popular - executing a script that returns a promise and polling event loop till the promise is resolved.

I'm not sure if there's a point in `resolve_value`; maybe it should be `resolve_promise` that would assert that passed value is a promise?

@ry any comments?

core/runtime.rs
```diff
@@ -1628,6 +1670,55 @@ pub mod tests {
     }
   }
 
+  #[tokio::test]
+  async fn test_resolve_value() {
```
bartlomieju, line 1674: Nice test 👍

cli/tools/test_runner.rs
```diff
@@ -220,22 +221,6 @@ pub async fn test_specifier(
     test_source.push_str(&format!("import \"{}\";\n", main_module));
   }
 
-  test_source
-    .push_str("await new Promise(resolve => setTimeout(resolve, 0));\n");
```
bartlomieju, line 224: Why is this tick removed and what was its purpose?

caspervonb: Removed because tests are passing without it.

This test is the one that used to fail without the tick skip: https://github.com/denoland/deno/blob/main/cli/tests/testdata/test/clear_timeout.ts

bartlomieju: Thanks that makes sense

cli/tests/testdata/test/interval.ts
```diff
@@ -0,0 +1 @@
+setInterval(function () {}, 0);
```
bartlomieju, line 1: Why does it exit instead of blocking forever?

cli/tools/test.rs
```diff
@@ -300,9 +285,29 @@ async fn test_specifier(
...
+    &located_script_name!(),
+    "window.dispatchEvent(new Event('load'));",
+  )?;
+
+  let test_result = worker.js_runtime.execute_script(
+    &located_script_name!(),
+    &format!(
+      r#"Deno[Deno.internal].runTests({})"#,
+      json!({
+        "disableLog": program_state.flags.log_level == Some(Level::Error),
```
bartlomieju, line 298: Note to self: conflicts with #11911

core/runtime.rs
```diff
@@ -647,6 +647,48 @@ impl JsRuntime {
...
   }
 
+  /// Waits for the given value to resolve while polling the event loop.
+  ///
+  /// This future resolves when either the value is resolved or the event loop runs to
+  /// completion.
+  pub async fn resolve_value(
+    &mut self,
+    global: v8::Global<v8::Value>,
+  ) -> Result<v8::Global<v8::Value>, AnyError> {
```
bartlomieju, line 657: @caspervonb can you please rework this function to `await_v8_promise`? It should assert that passed value is a `v8::Promise`. This is a very specialised API and I don't see why we should add path for values other than promises.

bartlomieju (commented):
LGTM, but please address comments below before landing

cli/tests/testdata/test/interval.ts
caspervonb: We only wait for the tests to complete, so the exit condition is when all tests complete rather than when all ops complete.

core/runtime.rs
caspervonb: So I started with a `resolve_promise` function that took a v8::Promise and failed if it wasn't a promise but that means you need to explicitly import v8 and convert in calling code, this lets you pass directly from execute_script with inferred types and matches the semantics of `Promise.resolve` (e.g `await Promise.resolve(0)` is basically equivalent to `resolve_promise(execute_script("(0)").unwrap()).unwrap().await`).

The main thing is taking in a `v8:::Global<v8::Value>>` and not a `v8:::Global<v8::Promise>>` as this means the caller doesn't have to import v8 and do manual conversions from the result returned by execute_script.

bartlomieju: Okay that makes sense, thanks

bartlomieju (approved):
LGTM, thanks

PR #11728, opened 2021-08-16
feat: CompressionStream

### LINKED ISSUE #7113: Implement CompressionStream
Compression streams allow for compression and decompression of data in a streaming way.

Example:
```js
const res = await fetch ("https://example.com/gzipcompressedfile.txt");
const data = res.body.pipeThrough(new DecompressionStream("gzip")).getReader();
// do something with the data here
```

Spec: https://wicg.github.io/compression/

This is a web standard, and it is implemented in Chrome stable.

crowlKats:
Closes #7113


lucacasonato:
I just looked into this. I think the only way we can get it to work is by splitting reading from a compressor and writing into it into seperate ops that happen in parallel. This is because a single write may not always correspond to a single read.

How exactly this can be most trivially implemented is still up for debate. I'll revisit this again.

linux-china:
Please include brotli  support.

crowlKats:
> Please include brotli support.

We wont. it isnt part of the specification

kt3k:
> Please include brotli support.

Maybe we can develop `std/compress` for non standard algorithms like we did `std/crypto` for non standard digest functions

zandaqo:
I assume the most common use case for this API will be online compression of response bodies. In that case, the default compression (zlib's level 6) is good enough, and arguably it's a much better choice than say brotli in most cases due to the complexity/performance trade off.

dojyorin:
I used CompressionStream in web application development and found it very fast and convenient, so I would like Deno to implement it as well!
I'm sure it will bring a wonderful experience.

ext/web/compression.rs
```diff
@@ -0,0 +1,187 @@
...
+  fn name(&self) -> Cow<str> {
+    "deflateDecompressor".into()
+  }
+
+  fn read(self: Rc<Self>, mut buf: ZeroCopyBuf) -> AsyncResult<usize> {
+    Box::pin(async move {
+      let resource = deno_core::RcRef::map(self, |r| &r.tx);
+      let mut tx = resource.borrow_mut().await;
+      let n = tx.read(&mut buf).await?;
+      Ok(n)
```
lucacasonato, line 117: Shouldn't this be called the `rx` end? It is receiving, not transmitting, right?

ry:
@crowlKats @lucacasonato This is close to working condition now. PTAL.

ext/web/compression.rs
```diff
@@ -0,0 +1,99 @@
...
+      d.write_all(&input)?;
+      d.get_mut().drain(..).collect()
+    }
+    Inner::GzDecoder(d) => {
+      d.write_all(&input)?;
+      d.get_mut().drain(..).collect()
+    }
+    Inner::GzEncoder(d) => {
+      d.write_all(&input)?;
+      d.get_mut().drain(..).collect()
```
lucacasonato, line 78: This is the trick I couldn't figure out in previous attempts. Great! 👍

lucacasonato (commented):
Looks great! The JS side bindings are not yet quite correct. I'll clean them up tomorrow, but then this is good to land!

Great job on figuring out this flate2 stuff. It looks so simple now that it's done 🙂

lucacasonato:
@ry You either don't have an up to date WPT submodule, or you didn't run `./tools/wpt.ts setup` before running WPT tests, as you didn't run some tests. The `/compression/decompression-correct-input.tentative.any.html` test hangs for me while running locally.

lucacasonato:
Boiled it down to this test case failing:

```js
const deflateChunkValue = new Uint8Array([120, 156, 75, 173, 40, 72, 77, 46, 73, 77, 81, 200, 47, 45, 41, 40, 45, 1, 0, 48, 173, 6, 36]);
const trueChunkValue = new TextEncoder().encode('expected output');

const ds = new DecompressionStream('deflate');
const reader = ds.readable.getReader();
const writer = ds.writable.getWriter();
const writePromise = writer.write(deflateChunkValue);
const { done, value } = await reader.read();
assert_array_equals(Array.from(value), trueChunkValue, "value should match");
await writePromise;
await writer.close();
```

The test fails on the line `reader.read()` with the following error:

```
error: Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promises.
```

lucacasonato:
Pushed the changes required to fix the IDL stuff. The happy case is not working at all though. Always fails with `error: Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promises.`.

ry:
@lucacasonato Yes I experienced that too. This seems to be a bug deep inside Deno somewhere, not in this compression stream - there are no async ops.

For example if I run:
```
import { assertEquals } from "https://deno.land/std/testing/asserts.ts";
const deflateChunkValue = new Uint8Array([120, 156, 75, 173, 40, 72, 77, 46, 73, 77, 81, 200, 47, 45, 41, 40, 45, 1, 0, 48, 173, 6, 36]);
const trueChunkValue = new TextEncoder().encode('expected output');
async function main() {
  const ds = new DecompressionStream('deflate');
  const reader = ds.readable.getReader();
  const writer = ds.writable.getWriter();
  console.log("A", Deno.resources());
  await writer.write(deflateChunkValue);
  console.log("B", Deno.resources());
  const { done, value } = await reader.read();
  assertEquals(Array.from(value), trueChunkValue, "value should match");
  console.log("C", Deno.resources());
}

main();
```

It never gets to the second `console.log`, but rather the program just exits. If you remove the `await` on the `write.write` line, it never gets to the third `console.log`. Something very wrong is happening.

lucacasonato:
Mh, that looks like a bug in the `TransformStream` implementation then. We kinda need to solve that first I guess. I'll investigate that later today.

ry (approved):
I think this first pass is done. LGTM

bartlomieju (approved):
LGTM too, nice work

PR #11735, opened 2021-08-16
chore: add `FinalizationRegistry` and `WeakRef` to primordials

aduh95:
Because it was possible to disable those with a runtime flag, they were not available through primordials. The flag has since been removed upstream.

Refs: https://github.com/v8/v8/commit/d59db06bf5425ddb388fb5a576f4bf39bdcc0f8f

### DIFF
--- a/core/00_primordials.js
+++ b/core/00_primordials.js
@@ -206,8 +206,7 @@
     "Date",
     "Error",
     "EvalError",
-    // TODO(lucacasonato): not present in snapshots. Why?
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

@lucacasonato I guess this gives an answer why these APIs were not available in snapshotting.

lucacasonato:
Yeah, seems so cc @bnoordhuis.

PR #11736, opened 2021-08-17
feat(cli): add nestable testing steps

caspervonb:


cli/dts/lib.deno.ns.d.ts
```diff
@@ -184,7 +186,32 @@ declare namespace Deno {
...
+     * to true. */
+    sanitizeResources?: boolean;
+
+    /** Ensure the test step does not prematurely cause the process to exit,
+     * for example via a call to `Deno.exit`. Defaults to true. */
+    sanitizeExit?: boolean;
+  }
+
+  export interface TestContext {
+    /** Execute a test step returning the error that occured if any.
```
lucacasonato, line 209: What is returned in case of no error? What if I `throw undefined`?

lucacasonato (commented):
This also needs some tests for the behaviour described in "No implicit subtest await" in #10771. Same goes for the failure cases described in "Sanitizers" and "Permissions" sections.

cli/dts/lib.deno.ns.d.ts
caspervonb: > What is returned in case of no error? What if I `throw undefined`?

Was thinking we'll return the rejected value and type it as unknown.

lucacasonato: How do you differentiate between a passed and a failed test then?

caspervonb: > How do you differentiate between a passed and a failed test then?

Can't if you reject with undefined but why would an assertion ever throw undefined? 🤔

cli/dts/lib.deno.ns.d.ts
```diff
@@ -184,7 +186,32 @@ declare namespace Deno {
...
+     * after the test has exactly the same contents as before the test. Defaults
+     * to true. */
+    sanitizeResources?: boolean;
+
+    /** Ensure the test step does not prematurely cause the process to exit,
+     * for example via a call to `Deno.exit`. Defaults to true. */
+    sanitizeExit?: boolean;
+  }
+
+  export interface TestContext {
```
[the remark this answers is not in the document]
caspervonb: Also isn't sound to export this as a class as it depends on test ops and those aren't available outside of tests so this isn't a class.

caspervonb:
Some observations from trying to use this with user land code...

## Failing fast is tedious and requires a lot of boilerplate.

```ts
Deno.test("fail-fast", async function(t) {
  const step1_result = t.step(function() {
    if (Math.random() > 0.5) {
      throw new Error("");
    }
  });

  if (!step1_result) {
    return;
  }

  const step2_result = t.step(function() {
    if (Math.random() > 0.5) {
      throw new Error("");
    }
  });

  if (!step2_result) {
    return;
  }

  const step3_result = t.step(function() {
    if (Math.random() > 0.5) {
      throw new Error("");
    }
  });

  if (!step3_result) {
    return;
  }
});
```

## As a form of hooks, it requires everything to become steps.

To use steps as a form of before and after hooks, everything now needs to become steps.
Your typical test suite will often now only a single Deno.test in it.

```ts
Deno.test("everything", async function(t) {
  const browser = launch({ browser: "chrome" });
  const server = serve({ port: 8080 });

  await t.step("localhost:8080/index.html", function(t) {
    const tab = await browser.openTab("http://localhost:8080");
    
    await tab.close();
  });

  await t.step("localhost:8080/edit.html", function() {
    const tab = await browser.openTab("http://localhost:8080");
    
    await tab.close();
  });

  const otherServer = serve({port: 8081 });

  await t.step("localhost:8081/index.html", function(t) {
    const tab = await browser.openTab("http://localhost:8081");
    
    await tab.close();
  });

  await t.step("localhost:8081/edit.html", function() {
    const tab = await browser.openTab("http://localhost:8081");
    
    await tab.close();
  });

  browser.close();
  server.close();
  otherServer.close();
});
```

bartlomieju:
Thank you Casper, but now that #12190 is merged I think this PR is no longer needed. Closing without merge.

PR #11738, opened 2021-08-17
fix(ext/crypto): KeyAlgorithm typings for supported algorithms

littledivy:


ext/crypto/lib.deno_crypto.d.ts
```diff
@@ -27,6 +27,7 @@ type KeyUsage =
   | "wrapKey";
 
 type NamedCurve = string;
+type BigInteger = Uint8Array;
```
lucacasonato, line 30: Let's not add this.

bartlomieju (approved):

PR #11739, opened 2021-08-17
feat(ext/crypto): verify ECDSA signatures

littledivy:
Towards #11690 

~~Blocked on #11738~~


ext/crypto/00_crypto.js
```diff
@@ -717,6 +718,25 @@
...
           }, data);
         }
+        case "ECDSA": {
+          // 1.
+          if (key[_type] !== "public") {
+            throw new DOMException(
+              "Key type not supported",
+              "InvalidAccessError",
+            );
+          }
```
bnoordhuis, line 728: Can you exercise this code path with a test?

(Mea culpa, I may have forgotten to point that out in your other pull requests.)

bnoordhuis (approved):

ext/crypto/00_crypto.js
lucacasonato: I am not 100% sure, but this is likely covered by WPT tests.

bnoordhuis: The same thought occurred to me the moment I hit "Submit review". :-)

littledivy: Yup WPT covers this...https://github.com/denoland/wpt/blob/146f12e8df2cac6b1e60152145124a81dad60d38/WebCryptoAPI/sign_verify/ecdsa.js#L238 (using a "secret" key to sign ECDSA)
Adding WPT is pending

littledivy: uhhh so WPT depends on importKey. I've added a unit test for now

ry (approved):
LGTM- thanks @littledivy !

PR #11746, opened 2021-08-17
feat(unstable): Add file locking APIs

### LINKED ISSUE #11511: Feature Request: Add APIs for locking files
Add sync and async APIs to the `Deno` namespace which allow to acquire (advisory) file-system locks. The API could look something like this:

```typescript
interface LockOptions {
  mode: "shared" | "exclusive";
}

interface Deno {
  // ...

  flock: (pid: number, options: LockOptions) => Promise<void>;
  fsyncLock: (pid: number, options: LockOptions) => void;

  funlock: (pid: number) => Promise<void>;
  fsyncUnlock: (pid: number) => void;

  // plus potentially something to check if a file lock can
  // be acquired.
  fcheckLock: (pid: number, options: LockOptions) => Promise<boolean>;
  fsyncCheckLock: (pid: number, options: LockOptions) => boolean;

  // ...
}
```

One possible good candidate for implementing those on the Rust side would be the [fs3 crate](https://crates.io/crates/fs3).

The motivation for having these APIs is that they would enable synchronization between different processes which access the same file. In particular, this functionality would enable the <https://deno.land/x/sqlite> library to provide the complete set of persistence guarantees native SQLite offers.

dyedgreen:
Closes #11511

# ToDo
- [x] Add an async version (currently using tokio::task::spawn_blocking)
- [x] Add tests
- [x] Should any of these operations require write permissions (I would say read more appropriate, since the file is not actually changed and `flock` is only advisory)?
- [x] Is the create this is based on okay? This crate is the most well-maintained one which a low-level API I can find. Other crates have more recent version updates, but an RAII API that would require managing the RAII locks on the rust side...


runtime/ops/fs.rs
```diff
@@ -346,6 +350,121 @@ async fn op_fstat_async(
...
+#[serde(rename_all = "camelCase")]
+pub struct FlockArgs {
+  rid: ResourceId,
+  exclusive: Option<bool>,
+}
+
+fn op_flock_sync(
+  state: &mut OpState,
+  args: FlockArgs,
+  _: (),
```
crowlKats, line 363: Instead of using a struct with 2 values, use both op arguments

dyedgreen:
As an aside: Do you know how to resolve linker issues with the ffi crate on Apple M1? I'm getting this error:
```text
  = note: ld: warning: directory not found for option '-L/.../deno/target/debug/build/deno-libffi-sys-60215bf917fe5bda/out/libffi-root/lib64'
          ld: in /.../deno/target/debug/deps/libdeno_libffi_sys-00703cb988a758bb.rlib(closures.o), archive member 'closures.o' with length 18232 is not mach-o or llvm bitcode for architecture arm64
```

I tried using the CodeSpaces, but they are really clunky 😢

kitsonk:
@ry @lucacasonato any advice regarding the M1 crate problem?

bartlomieju:
> Is the create this is based on okay? This crate is the most well-maintained one which a low-level API I can find. Other crates have more recent version updates, but an RAII API that would require managing the RAII locks on the rust side...

@dyedgreen yes it's fine, the crate hasn't been updated in a year, but I believe that's because it just works, not because it was abandoned.

bartlomieju (approved):
LGTM, cleanly implemented, nice work @dyedgreen!

PR #11749, opened 2021-08-17
chore(tests): delete the temporary deno dir when dropping the test lsp client

dsherret:


kitsonk (approved):
LGTM

test_util/src/lsp.rs
```diff
@@ -97,6 +98,8 @@ pub struct LspClient {
   request_id: u64,
   start: Instant,
   writer: io::BufWriter<ChildStdin>,
+  #[allow(dead_code)]
```
dsherret: Oh, I just remembered I can get rid of the attribute by adding an underscore. I'll do that instead.

kitsonk: 👍

PR #11755, opened 2021-08-18
fix(lsp): better handling of languageId

### LINKED ISSUE #11521: feature requests(lsp): Fallback on invalid languageId
Currently, the only valid `languageId` for "textDocument/didOpen" requests are "javascript", "javascriptreact", "typescript", "typescriptreact", "json", "jsonc", and "markdown".
LSP's TypeScript diagnostics are also displayed when an invalid `languageId` (eg "github markdown") is passed.

How about hiding TypeScript diagnostics or throwing an error if the `languageId` is invalid?

### LINKED ISSUE #11742: lsp: support non standard language IDs
The [Nova](https://nova.app) text editor for mac utilizes the LSP but unfortunately for some languages it uses different syntax identifiers than is [standard for LSP](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem):

- `jsx` vs `javascriptreact`
- `tsx` vs `typescriptreact`

The Deno language server only understands the standard ones, and errors at initialization when you use `jsx` or `tsx`. This makes it difficult for us to make Deno extensions for Nova.

I've [reported the problem in the Nova dev forum](https://devforum.nova.app/t/how-to-workaround-nova-syntax-ids-differing-with-a-language-server/1100), but I really doubt Nova would change the syntax identifiers their ecosystem is build around at this point. I suggested that instead of both Nova and the language server using the same syntax identifiers, the Nova [`LanguageClient`](https://docs.nova.app/api-reference/language-client/#constructoridentifier-name-serveroptions-clientoptions) constructor client options parameter `syntaxes` option should be changed to an object with keys and values that map Nova syntax identifiers to language server identifiers. Nova has a very small team of devs (only 1 full time at the moment) so this feature request might take a long time to be implemented, if ever.

`javascriptreact` and `typescriptreact` are pretty gross identifiers as people use JSX and TSX in non-React projects, so I can see other editors in the future using the more semantically correct (although non-standard) `jsx` and `tsx`.

Would Deno consider supporting the `jsx` and `tsx` identifiers out of the box as aliases for `javascriptreact` and `typescriptreact`?

If not, how about allowing syntax identifier aliases to be a configuration option?

_Originally posted by @jaydenseric in https://github.com/denoland/deno/issues/11521#issuecomment-900036748_

kitsonk:
This adds a `LanguageId::Unknown` which better aligns to mapping to `MediaType`.  Also we were defaulting to `LanguageId::TypeScript` when we encountered an unknown/unsupported language id.  This cause diagnostics to appear in the file from tsc, even though it should have been considered "undiagnosable".

This also adds support for `"jsx"` and `"tsx"` language IDs.

Fixes #11521
Fixes #11742

### DIFF
--- a/cli/lsp/README.md
+++ b/cli/lsp/README.md
@@ -22,6 +22,7 @@ There are several settings that the language server supports for a workspace:
 - `deno.cache`
 - `deno.config`
 - `deno.importMap`
+- `deno.internalDebug`
 - `deno.codeLens.implementations`
 - `deno.codeLens.references`
 - `deno.codeLens.referencesAllFunctions`
@@ -159,3 +160,21 @@ client:
     suggestions: boolean;
   }
   ```
+
+## Language IDs
+
+The language server supports diagnostics and formatting for the following
+language IDs:
+
+- `"javascript"`
+- `"javascriptreact"`
+- `"jsx"` _non standard_
+- `"typescript"`
+- `"typescriptreact"`
+- `"tsx"` _non standard_
+
+The language server supports only formatting for the following language IDs:
+
+- `"json"`
+- `"jsonc"`
+- `"markdown"`
--- a/cli/lsp/README.md
+++ b/cli/lsp/README.md
@@ -164,14 +164,14 @@ client:
 ## Language IDs
 
 The language server supports diagnostics and formatting for the following
-language IDs:
+[text document language IDs](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem):
 
 - `"javascript"`
 - `"javascriptreact"`
-- `"jsx"` _non standard_
+- `"jsx"` _non standard, same as `javascriptreact`_
 - `"typescript"`
 - `"typescriptreact"`
-- `"tsx"` _non standard_
+- `"tsx"` _non standard, same as `typescriptreact`_
 
 The language server supports only formatting for the following language IDs:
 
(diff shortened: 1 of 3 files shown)


dsherret (approved):
+1 LGTM

cli/lsp/README.md
```diff
@@ -159,3 +160,21 @@ client:
...
   ```
+
+## Language IDs
+
+The language server supports diagnostics and formatting for the following
+language IDs:
+
+- `"javascript"`
+- `"javascriptreact"`
+- `"jsx"` _non standard_
```
bartlomieju, line 171: Maybe add note that it's the same as `javascriptreact`?

cli/lsp/README.md
```diff
@@ -159,3 +160,21 @@ client:
...
+- `"typescript"`
+- `"typescriptreact"`
+- `"tsx"` _non standard_
```
bartlomieju, line 174: Ditto

bartlomieju (approved):
LGTM, too

cli/lsp/README.md
kitsonk: Because they are non-standard: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocumentItem. I didn't want to get into what they do here, as I would prefer they not be used and it is an "internal" concern.

I will link to the standard.

kitsonk: Actually, upon refelction I will do both. 👍

PR #11763, opened 2021-08-18
perf(ext/url): cleanup and optimize url parsing op args

AaronO:
This splits the previous `op_url_parse` into:
- `op_url_parse`: parses a href with an optional base
- `op_url_reparse`: reparses a href with a modifier

This is a cleaner separation of concerns and it allows us to optimize & simplify args passed. Resulting in a 25% reduction in call overhead (`~5000ns/call` => `~3700ns/call` in the `url_ops` bench on my M1 Air)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -898,6 +898,7 @@ dependencies = [
  "idna",
  "percent-encoding",
  "serde",
+ "serde_repr",
 ]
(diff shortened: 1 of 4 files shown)


lucacasonato (approved):
Impressive improvement!

PR #11764, opened 2021-08-18
feat: add options to test overload

caspervonb:
Generally you can't actually use the overloaded version of `Deno.test` because the need to specify permissions or any of the other options.

This adds the remaining options as an optional option bag to the user friendly overloaded version of `Deno.test`.

We punted this back in https://github.com/denoland/deno/pull/10078

Since then we have added more options like permissions making the need much greater.


bartlomieju:
Thank you Casper, but I'm going to close this PR. I'd much rather pursue overloads described in https://github.com/denoland/deno/issues/12744.

PR #11765, opened 2021-08-18
perf(ext/url): optimize UrlParts op serialization

AaronO:
Marshalling it across the op-layer by joining the substrings into a single "fat" newline-joined string, then splitting JS side.

This shaves off a little over a 1/3 of the overhead per call:
`main` (after merging https://github.com/denoland/deno/pull/11763)
```
bench:   3,726,203 ns/iter (+/- 177,656)
```
this PR:
```
bench:   2,400,289 ns/iter (+/- 53,911)
```

So with this PR and #11763 that's a combined improvement of over 2x ! (~5000ns/call => ~2400ns/call)

Whilst this might appear counterintuitive and "hacky", it works because:
- We have 11 string values to return
- Marshalling them in a struct creates an obj + 11 optimized/internalized strings (the keys) + 11 unoptimized "normal" strings (the values) + 11 owned strings in Rust
- This approach creates 1 fat owned string in rust, then resplitting it in JS creates 11 efficient substrings (that aren't individually allocated and refer to the fat string)
- `\n` is a safe separator because it's not allowed in URLs, technically could have used any control char but `\n` works and is familiar

### DIFF
--- a/ext/url/lib.rs
+++ b/ext/url/lib.rs
@@ -11,7 +11,6 @@ use deno_core::url::quirks;
 use deno_core::url::Url;
 use deno_core::Extension;
 use deno_core::ZeroCopyBuf;
-use serde::Serialize;
 use std::panic::catch_unwind;
 use std::path::PathBuf;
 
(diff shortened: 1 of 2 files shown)


ext/url/lib.rs
```diff
@@ -137,19 +139,22 @@ fn url_result(
...
-    search: quirks::search(&url).to_string(),
-    username: username.to_string(),
-  })
+  Ok(
+    [
+      quirks::href(&url),
+      quirks::hash(&url),
+      quirks::host(&url),
+      quirks::hostname(&url),
+      &quirks::origin(&url),
```
satyarohith, line 148: `quirks::origin` returns a `String`

lucacasonato (approved):
Have you considered making UrlParts into `strurct UrlParts(Url)`, and then implementing `serde::Serialize` on it manually? All the serialization logic could be abstracted into the serializer.

LGTM though.

AaronO:
> Have you considered making UrlParts into `strurct UrlParts(Url)`, and then implementing `serde::Serialize` on it manually? All the serialization logic could be abstracted into the serializer.
> 
> LGTM though.

Yes, I patched the `url` crate so that it exposes its internal offsets. That's a more efficient serialization that I've prototyped locally, but it either requires patching `url` or doing pointer arithmetic to recompute those internal offsets.

PR #11769, opened 2021-08-19
refactor(cli): introduce testing modes

caspervonb:
This merges the two vectors of specifiers into a single one introducing the concept of a `TestMode` which is an enum specifying how a specifier is to be tested.

This is determined during the collection phase and determines how a specifier will be executed.
For example `deno test README.md` is marked as inline and will only have it's inline code blocks tested.


caspervonb:
Superceedes https://github.com/denoland/deno/pull/11683
Not really a simplification anymore since it's *more* code but it hardens the entire ordeal and I need everything to be in a single vector for running inline documentation tests.

cli/tools/test.rs
```diff
@@ -458,42 +471,34 @@ async fn fetch_inline_files(
   Ok(files)
 }
 
-/// Runs tests.
-///
-#[allow(clippy::too_many_arguments)]
```
caspervonb, line 463: At-least this allow is gone.

cli/tests/testdata/test/shuffle.out
```diff
@@ -1,6 +1,6 @@
-Check [WILDCARD]/test/shuffle/foo_test.ts
-Check [WILDCARD]/test/shuffle/baz_test.ts
 Check [WILDCARD]/test/shuffle/bar_test.ts
+Check [WILDCARD]/test/shuffle/baz_test.ts
```
caspervonb, line 2: We can maintain the previous behaviour of re-shuffling checks, just did not see no value in it.

bartlomieju: That sounds good to me

cli/tools/mod.rs
```diff
@@ -7,5 +7,5 @@ pub mod installer;
 pub mod lint;
 pub mod repl;
 pub mod standalone;
-pub mod test_runner;
+pub mod test;
```
bartlomieju, line 10: Why change the module name? I don't think it's necessary - actually `test_runner` is more descriptive than `test`

cli/tools/test.rs
```diff
@@ -37,6 +49,18 @@ use std::time::Instant;
...
+#[derive(Debug, Clone, PartialEq)]
+enum TestMode {
+  /// Test as documentation, type-checking fenced code blocks.
+  Documentation,
+  /// Test as an executable module, loading the module into the isolate and running each test it
+  /// defines.
+  Executable,
+  /// Test as both documentation and an executable module.
+  Both,
+}
```
bartlomieju, line 62: Good docs, thanks 👍

cli/tools/test.rs
```diff
@@ -508,67 +522,80 @@ pub async fn run_tests(
...
-        let future = test_specifier(
-          program_state,
-          main_module,
-          permissions,
-          filter,
-          shuffle,
-          sender,
-        );
+/// Test a collection of specifiers concurrently.
+#[allow(clippy::too_many_arguments)]
```
bartlomieju, line 550: This is unlucky, any ideas how we could simplify function signature?

cli/tools/test.rs
```diff
@@ -669,3 +696,335 @@ pub async fn run_tests(
...
+      *mode = TestMode::Both
+    } else {
+      *mode = TestMode::Documentation
+    }
+  }
+
+  Ok(specifiers_with_mode)
+}
+
+#[allow(clippy::too_many_arguments)]
```
bartlomieju, line 769: Ditto

bartlomieju (commented):
This looks mostly good to me, just a few minor comments, please rebase

cli/tools/test.rs
caspervonb: Yeah we can do an options struct.

PR #11775, opened 2021-08-19
perf(ext/url): use DOMString webidl instead of USVString 

AaronO:
Improves url parsing by ~20%:
- before: `~2450ns/parse`
- after: `~1950ns/parse`

This is safe since all these string values are immediately sent to the op-layer that uses `v8::String::to_rust_string_lossy()` with `REPLACE_INVALID_UTF8` effectively doing the same transform as our `USVString` webidl converter

### DIFF
--- a/ext/url/00_url.js
+++ b/ext/url/00_url.js
@@ -313,9 +313,9 @@
      */
     constructor(url, base = undefined) {
       const prefix = "Failed to construct 'URL'";
-      url = webidl.converters.USVString(url, { prefix, context: "Argument 1" });
+      url = webidl.converters.DOMString(url, { prefix, context: "Argument 1" });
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):

lucacasonato:
The same probably applies to URLSearchParams too

AaronO:
> The same probably applies to URLSearchParams too

Only partially, since `URLSearchParams` has pure JS getters (see its `get()/has()`), so if we changed the converters but left the code otherwise as is there would be a change in behaviour since it's possible to have `a !== b` yet `USVString(a) === USVString(b)`

lucacasonato:
Makes sense. We could do it in the constructor though (that goes right into a parse op).

Issue #11686, opened 2021-08-13
First pass at config file


ry:
Part of https://github.com/denoland/deno/issues/3179

1. Deno will lookup a file called `Deno.json` in the current working directory on startup.
2. This `Deno.json` can also be specified by the already existing `--config` flag
3. `Deno.json` can have a top-level `"compilerOptions"` for configuring typescript.
4. We will initially support one other top-level configuration called `"lint"` which will allow people to configure how `deno lint` works. Example:
```
{
  "compilerOptions": {
    "jsxFactory": "h"
  },

  "lint": {
    "ignore": ["testdata/"],
    "rules": {
      "tags": ["recommended"],
      "include": [
        "ban-untagged-todo"
      ]
    }
  }
}
```

kitsonk:
Just to comment that I have no objections to this.

The auto discovery is going to be slightly problematic for the language server, especially with multi root workspaces like vscode supports.

bartlomieju:
Here's an [an example config](https://github.com/denoland/deno_lint/blob/main/examples/dlint/example_config.json) file from `dlint` (the example binary produced in `deno_lint` repo):
```json
{
    "rules": {
        "tags": ["recommended"],
        "include": [
            "ban-untagged-todo"
        ],
        "exclude": [
            "no-explicit-any"
        ]
    },
    "files": {
        "include": [
            "benchmarks/oak/**/*.ts"
        ],
        "exclude": []
    }
}
```

It's heavily based on what `dprint` does, ie. it has `"files"` section that allows to specify `"include"` and `"exclude"` configuration. It also supports globbing. We'd need to alter the schema a bit to conform to current flags configuration. My question is should we actually add support for globs (it's been highly requested feature, eg. #6365).

Additionally if we are to add config for `fmt` we might reuse a bulk of existing logic from `dprint`. @dsherret any thoughts?

SpadeAceman:
> 1. Deno will lookup a file called `Deno.json` in the current working directory on startup.

Using the current working directory doesn't sound right to me. I'd expect it to look for `Deno.json` in the same directory as the JavaScript/TypeScript file I'm running/formatting/linting, which won't always be the current working directory.


> 2. This `Deno.json` can also be specified by the already existing `--config` flag

Just to be clear, in future will the `--config` flag override/prevent automatic `Deno.json` loading?

[...]

PR #11776, opened 2021-08-19
feat(lint): add support for config file and CLI flags for rules

bartlomieju:
This commit adds support for following flags in `deno lint` subcommand:
- `--config` - allows to load configuration file and parses "lint" object
- `--rules-tags=<tags>` - allows specifying which set of tagged rules should be run
- `--rules-include=<rules>` - allow specifying which rules should be run
- `--rules-exclude=<rules>` - allow specifying which rules should not be run

Ref https://github.com/denoland/deno/issues/11686


cli/config_file.rs
```diff
@@ -266,10 +266,26 @@ impl Serialize for TsConfig {
...
+  pub include: Option<Vec<String>>,
+  pub exclude: Option<Vec<String>>,
+}
+
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct LintConfig {
+  pub rules: LintRulesConfig,
+  pub ignore: Vec<String>,
+}
```
bartlomieju, line 282: I'm not happy with this schema, I'd much prefer to go with the solution proposed in: https://github.com/denoland/deno/issues/11686#issuecomment-898863720

In this schema there's no way to add files that should be included.

dsherret: I think at least changing this to `"files": { "exclude": [...] }` might help keep this flexible for future support of "includes" and won't confuse people about what "ignore" means.

bartlomieju: Inclusion and exclusion of files is already possible in CLI using positional arg and `--ignore` option respectively. The schema would schema slightly different names but would support exactly same functionality. So I'm highly interested favor of using `files`.

bartlomieju: I changed the schema to use `"files": { "include": [], "exclude": {} }`

cli/tools/lint.rs
```diff
@@ -42,21 +44,47 @@ fn create_reporter(kind: LintReporterKind) -> Box<dyn LintReporter + Send> {
...
+    include_files.extend(include_files_from_config_file);
+
+    let exclude_files_from_config_file = lint_config
+      .files
+      .exclude
+      .iter()
+      .map(PathBuf::from)
+      .collect::<Vec<PathBuf>>();
+    exclude_files.extend(exclude_files_from_config_file);
+  }
```
bartlomieju, line 76: I'm still debating whether we should "extend" default globs with what's listed in `lint.files.include` or should it override default glob (the same applies to combining `litn.files.exclude` and `--ignore` flag). 

@dsherret what are your thoughts?

dsherret: I'm still trying to figure this out in dprint and I'm not sure what the best way is. Right now I override.

bartlomieju: Thinking about it more I'm leaning towards higher precedence of CLI flags - ie. if you use `--ignore` or provide arguments to `deno lint` they are used instead of values from config file.

Example:
```
// Deno.jsonc
"lint": {
  "files": {
    "include": ["src/", "frontend/"],
    "exclude": ["src/testdata/"]
  },
  "rules": {
    "tags": ["recommended"],
    "include": ["ban-untagged-todo"]
  }
}

deno lint -c Deno.jsonc --ignore foo/ src/
```
Will use `rules` config as in `Deno.jsonc` but will ignore only `foo/` directory and include `src/` directory.

cli/config_file.rs
caspervonb: Does exclude have the same semantics as ignore here? would be better to retain the name imho so there's no new things to know (also we've used include and exclude for regular expression filters in coverage cli).

Just throwing out a strawman:
```ts
"files": {
  "collect": [],
  "ignore": " [],
 }
 ```

File: cli/config_file.rs [hunk 3]
```diff
@@ -266,10 +266,33 @@ impl Serialize for TsConfig {
   }
 }
 
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
```
lucacasonato, line 270: use `deny_unknown_fields`? https://serde.rs/container-attrs.html#deny_unknown_fields

cli/config_file.rs
grian32: I think it's better to just follow `tsconfig.json` here as a lot of developers should be familiar with that, and it uses `exclude` and `include` for files.

cli/tools/lint.rs
```diff
@@ -373,3 +415,57 @@ fn sort_diagnostics(diagnostics: &mut Vec<LintDiagnostic>) {
...
   });
 }
+
+fn get_rules_from_config(
+  maybe_lint_config: Option<&LintConfig>,
+) -> Result<Vec<Box<dyn LintRule>>, AnyError> {
+  let configured_rules = if let Some(lint_config) = maybe_lint_config {
+    let mut filtered_rules = HashMap::new();
+
+    if let Some(tags) = &lint_config.rules.tags {
```
dsherret, line 425: I don't have an opinion on this, but if someone doesn't specify "tags", should we default to the default of "recommended"? So if they don't want the default and want nothing they would have to specify `"tags": []`

I'm thinking a common path might be someone doing:

```json
{
  "lint": {
    "exclude": ["some-rule"]
  }
}
```

...and in doing so, they would have to remember to add `"tags": ["recommended"]`. Alternatively, perhaps we can suggest this to the user if they do the above (see my comment below).

cli/tools/lint.rs
```diff
@@ -373,3 +415,57 @@ fn sort_diagnostics(diagnostics: &mut Vec<LintDiagnostic>) {
...
+    filtered_rules
+      .into_iter()
+      .map(|(_code, rule)| rule)
+      .collect()
+  } else {
+    rules::get_recommended_rules()
+  };
+
+  if configured_rules.is_empty() {
+    anyhow!("No rules have been configured");
```
dsherret, line 467: Maybe we can suggest to add `"tags": ["recommended"]` if there's no `tags` property? That will probably address my scenario above.

cli/config_file.rs
```diff
@@ -266,10 +266,34 @@ impl Serialize for TsConfig {
...
+pub struct LintFilesConfig {
+  pub include: Vec<String>,
+  pub exclude: Vec<String>,
+}
+
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+#[serde(deny_unknown_fields)]
```
lucacasonato, line 286:
```suggestion
#[serde(default, deny_unknown_fields)]
```

Shouldn't the fields be combined?

cli/config_file.rs
```diff
@@ -266,10 +266,34 @@ impl Serialize for TsConfig {
   }
 }
 
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct LintRulesConfig {
+  pub tags: Option<Vec<String>>,
+  pub include: Option<Vec<String>>,
+  pub exclude: Option<Vec<String>>,
```
lucacasonato, line 274: I don't remember this `exclude` being in the first draft we discussed. I have reservations about this.

lucacasonato (changes requested):
There are no flags for `lint.rules.tags`, `lint.rules.tags`

cli/config_file.rs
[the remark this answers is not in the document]
bartlomieju: Not sure, I'll verify

cli/tests/integration/lint_tests.rs
```diff
@@ -24,59 +24,79 @@ fn ignore_unexplicit_files() {
...
 // Make sure that the rules are printed if quiet option is enabled.
 itest!(rules_quiet {
-  args: "lint --unstable --rules -q",
+  args: "lint --rules -q",
   output: "lint/expected_rules.out",
   exit_code: 0,
 });
+
+itest!(lint_with_config {
+  args: "lint --config lint/Deno.jsonc lint/lint_with_config/",
```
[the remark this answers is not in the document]
bartlomieju: No, we went away from requiring `--unstable` flag for CLI flags

cli/tools/lint.rs
bartlomieju: "recommended" is enabled by default - you get this error only if you do `"tags": []` and don't specify anything in `rules.include`

cli/config_file.rs
bartlomieju: Sure

cli/config_file.rs
bartlomieju: I believe Ryan didn't add it to https://github.com/denoland/deno/issues/11686 but it was based on existing config in `dlint`. What are your reservations?

lucacasonato:
> Correct. What's your suggestion here?

We agreed on no config file options that don't also have flags.

bartlomieju:
> > Correct. What's your suggestion here?
> 
> We agreed on no config file options that don't also have flags.

I can add the flags. Honestly having only `files` configuration is useless in my opinion.

cli/config_file.rs
lucacasonato: I'll just give up on my argument against this. We discussed it so often earlier this year (registry wants to do linting to score packages, can't do that reliably if people start disabling rules in the recommended set).

cli/tools/lint.rs
dsherret: Ok thanks! I misread the code then.

cli/config_file.rs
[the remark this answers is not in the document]
bartlomieju: Added to descendant structures as well

cli/config_file.rs
bartlomieju: Can't do this reliably if you specify additional rules as well. We can still discuss it, no need to give up

cli/config_file.rs
```diff
@@ -266,10 +266,33 @@ impl Serialize for TsConfig {
   }
 }
 
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct LintRulesConfig {
+  pub tags: Option<Vec<String>>,
```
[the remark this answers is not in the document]
bartlomieju: @dsherret sorry I missed this comment. There's no single issue in `deno_lint` repo at the moment, but the basic idea is that there would be tags like "js-only" or "ts-only". I'm fine with skipping tags in the first pass, but that means users would only be able to exclude and include rules one by one. I guess it might be fine for first pass after all.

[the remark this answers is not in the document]
dsherret: I guess if this whole feature is unstable then adding tags is not a big deal as we can remove it if necessary once stabilizing.

cli/main.rs
```diff
@@ -1181,9 +1202,22 @@ fn get_subcommand(
...
+      flags,
+      files,
+      rules,
+      rules_tags,
+      rules_include,
+      rules_exclude,
+      ignore,
+      json,
+    )
+    .boxed_local(),
```
bartlomieju, line 1220: These commands are getting really bad... we've got a few subcommands that have more than 5 arguments.

I think it might be high time to refactor `DenoSubcommand` enum to something like:
```rust
#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
LintSubcommand {
  files: Vec<PathBuf>,
  ignore: Vec<PathBuf>,
  rules: bool,
  rules_tags: Vec<String>,
  rules_include: Vec<String>,
  rules_exclude: Vec<String>,
  json: bool,
}

#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
pub enum DenoSubcommand {
  ...
  Lint(LintSubcommand),
  ...
}

```

cli/config_file.rs
```diff
@@ -340,6 +363,16 @@ impl ConfigFile {
...
+
+  pub fn as_lint_config(&self) -> Result<Option<LintConfig>, AnyError> {
+    if let Some(config) = self.json.lint.clone() {
+      let lint_config: LintConfig = serde_json::from_value(config)
+        .context("Failed to parse \"lint\" configuration")?;
+      Ok(Some(lint_config))
+    } else {
+      Ok(None)
+    }
+  }
```
magurotuna, line 375: Rust API guidelines say that APIs that are expensive to call like this should be prefixed with `to_` rather than `as_`. 
https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv
This API is not exposed externally (since deno is a binary crate) so it's not very important to follow naming convention though.

cli/tools/lint.rs
```diff
@@ -69,11 +110,29 @@ pub async fn lint_files(
...
+  // over config file, ie. if there's `rules.include` in config file
+  // and `--rules-include` CLI flag, only the flag value is taken into account.
+  // TODO(bartlomieju): this is done multiple times for each file because
+  // Vec<Box<dyn LintRule>> is not clonable, this should be optimized.
+  get_configured_rules(
+    maybe_lint_config.as_ref(),
+    rules_tags.clone(),
+    rules_include.clone(),
+    rules_exclude.clone(),
+  )?;
```
magurotuna, line 123: what is this call for?

bartlomieju: This is validating that provided configuration actually resolves to at least 1 rule - otherwise it will error out saying that there are no configured rules. This is a code smell, but it's the only place this can be done, before parallelizing work across threads. It's done again for each file later, but at that point it's infallible.

cli/tools/lint.rs
```diff
@@ -144,13 +203,24 @@ pub fn create_linter(syntax: Syntax, rules: Vec<Box<dyn LintRule>>) -> Linter {
...
 
-  let lint_rules = rules::get_recommended_rules();
+  // Obtaining rules from config is infallible at this point.
+  let lint_rules = get_configured_rules(
+    maybe_lint_config,
+    rules_tags,
+    rules_include,
+    rules_exclude,
+  )
+  .unwrap();
```
magurotuna, line 223: Is this comment true? If the code path gets into

```rust
  if configured_rules.is_empty() {
    anyhow!("No rules have been configured");
  }
```

in `get_configured_rules` function because the user has specified all rules as "exclude", then returned value will be `Err`, no?

cli/tools/lint.rs
magurotuna: ah that makes sense!

cli/tools/lint.rs
magurotuna: explanation: https://github.com/denoland/deno/pull/11776#discussion_r700217088

cli/config_file.rs
bartlomieju: Addressed

dsherret (approved):
LGTM

lucacasonato (approved):
Looks great. Well done @bartlomieju.

PR #11777, opened 2021-08-19
chore: upgrade dlint and run `prefer-primordials`

magurotuna:
This PR upgrades `third_party` submodule to the latest to get dlint 0.12.0, and applies `prefer-primordials` to the files under runtime, ext, and core.

### DIFF
--- a/.dlint.json
+++ b/.dlint.json
@@ -4,6 +4,8 @@
     "include": [
       "ban-untagged-todo"
     ],
-    "exclude": []
+    "exclude": [
+      "no-invalid-triple-slash-reference"
+    ]
   }
 }
--- a/runtime/js/99_main.js
+++ b/runtime/js/99_main.js
@@ -22,6 +22,7 @@ delete Object.prototype.__proto__;
     SymbolFor,
     SymbolIterator,
     PromisePrototypeThen,
+    TypeError,
   } = window.__bootstrap.primordials;
   const util = window.__bootstrap.util;
   const eventTarget = window.__bootstrap.eventTarget;
@@ -228,6 +229,7 @@ delete Object.prototype.__proto__;
     } else {
       prepareStackTrace = core.createPrepareStackTrace();
     }
+    // deno-lint-ignore prefer-primordials
     Error.prepareStackTrace = prepareStackTrace;
   }
 
--- a/tools/lint.js
+++ b/tools/lint.js
@@ -51,6 +51,9 @@ async function dlint() {
   }
 }
 
+// `prefer-primordials` has to apply only to files related to bootstrapping,
+// which is different from other lint rules. This is why this dedicated function
+// is needed.
 async function dlintPreferPrimordials() {
   const execPath = getPrebuiltToolPath("dlint");
   console.log("prefer-primordials");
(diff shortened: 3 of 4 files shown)


.dlint.json
```diff
@@ -4,6 +4,8 @@
     "include": [
       "ban-untagged-todo"
     ],
-    "exclude": []
+    "exclude": [
+      "no-invalid-triple-slash-reference"
+    ]
```
magurotuna, line 9: If I understand correctly, `no-invalid-triple-slash-reference` rule is for Deno users, not for the internal code of Deno, so I added it here. Please correct me if I'm wrong.

runtime/js/99_main.js
```diff
@@ -228,6 +229,7 @@ delete Object.prototype.__proto__;
     } else {
       prepareStackTrace = core.createPrepareStackTrace();
     }
+    // deno-lint-ignore prefer-primordials
     Error.prepareStackTrace = prepareStackTrace;
```
magurotuna, line 233: In Node.js, `Error.prepareStackTrace` is skipped linting because of [the config](https://github.com/nodejs/node/blob/e4b1fb5e6422c1ff151234bb9de792d45dd88d87/lib/.eslintrc.yaml#L104-L105). But deno_lint doesn't support it right now, so just added an ignore directive here for the moment.

tools/lint.js
```diff
@@ -63,6 +51,50 @@ async function dlint() {
...
+    const p = Deno.run({
+      cmd: [execPath, "run", "--rule", "prefer-primordials", ...chunk],
+    });
+    const { success } = await p.status();
+    if (!success) {
+      throw new Error("prefer-primordials failed");
+    }
+    p.close();
+  }
+}
```
magurotuna, line 80: I think `prefer-primordials` should apply only to the code related to bootstrapping, so I created another function where only this rule is run.

tools/lint.js
```diff
@@ -63,6 +51,50 @@ async function dlint() {
...
+
+  if (!sourceFiles.length) {
+    return;
+  }
+
+  const chunks = splitToChunks(sourceFiles, `${execPath} run`.length);
+  for (const chunk of chunks) {
+    const p = Deno.run({
+      cmd: [execPath, "run", "--rule", "prefer-primordials", ...chunk],
+    });
```
bartlomieju, line 73: Why configure here instead of adding to `dlint.json`?

tools/lint.js
bartlomieju: Makes sense, please add a comment explaining that

tools/lint.js
magurotuna: If we added `prefer-primordials` to `.dlint.json`, then this rule would apply to almost all JS and TS files in this repository, which would produce lots of diagnostics. This rule should apply only to files under `runtime`, `ext` and `core`, which is why I made it inlined here.

tools/lint.js
magurotuna: Done 👍  0cf3bfb

lucacasonato (commented):
This is awesome! Well done @magurotuna.

.dlint.json
bartlomieju: That's correct

bartlomieju (approved):
LGTM, thanks @magurotuna this is very useful

PR #11780, opened 2021-08-19
refactor(lsp): use deno_ast and cache swc ASTs

### LINKED ISSUE #11345: lsp: cache swc ASTs

### LINKED ISSUE #11854: Support class static blocks (Stage-3)
A Stage-3 proposal to allow static initialisation blocks: https://github.com/tc39/proposal-class-static-block

- [x] v8 tracking bug: https://bugs.chromium.org/p/v8/issues/detail?id=11375
- [x] Shipping in TypeScript 4.4: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#static-blocks #11678 
- [ ] swc issue: https://github.com/swc-project/swc/issues/2125
  - [x] https://github.com/swc-project/swc/issues/2195
- [x] deno fmt issue: https://github.com/dprint/dprint-plugin-typescript/issues/273

dsherret:
This change integrates `deno_ast` and lazily caches/reuses swc AST in the language server.

I've yet to come up with a benchmark I think is worthwhile to have... given that swc is already so fast this change is only noticable with very large files. I would say this isn't so much a performance improvement, but rather just a refactoring.

Closes #11345
Closes #11854


cli/lsp/documents.rs
```diff
@@ -350,15 +324,15 @@ impl DocumentCache {
 
   pub fn line_index(&self, specifier: &ModuleSpecifier) -> Option<LineIndex> {
     let doc = self.docs.get(specifier)?;
-    doc.line_index.clone()
+    Some(doc.source().line_index().clone())
```
dsherret, line 327: This could be changed to not clone, but we can worry about that in a future refactoring...

cli/tools/fmt.rs
```diff
@@ -91,14 +89,10 @@ pub async fn format(
 
 /// Formats markdown (using https://github.com/dprint/dprint-plugin-markdown) and its code blocks
 /// (ts/tsx, js/jsx).
-fn format_markdown(
-  file_text: &str,
-  ts_config: dprint_plugin_typescript::configuration::Configuration,
```
dsherret, line 96: Config objects were being needlessly passed around.

cli/tools/lint.rs
```diff
@@ -277,24 +278,27 @@ pub fn format_diagnostic(
 ) -> String {
   let mut lines = vec![];
 
-  for i in range.start.line..=range.end.line {
-    lines.push(source_lines[i - 1].to_string());
```
dsherret, line 281: The transition away from 1-indexed non-display values hasn't completed yet, but the change was done in deno_lint (except for json serialization).

cli/Cargo.toml
```diff
@@ -72,6 +72,7 @@ log = { version = "0.4.14", features = ["serde"] }
 lspower = "1.1.0"
 notify = "5.0.0-pre.10"
 num_cpus = "1.13.0"
+once_cell = "1.8.0"
```
dsherret, line 75: This is already used in downstream dependencies and is being evaluated for the Rust standard library. It allows us to lazily create a value in a multi-threaded environment and then return a reference to that created value.

dsherret:
@kitsonk this is in draft because of the failing build due to problems I encountered integrating, but it's good enough for a review.

(see the PR description—I shouldn't have updated swc in the crates so that I wouldn't need to bump deno_doc)

kitsonk (commented):
LGTM, while it may only appreciably impact performance with large documents, especially with the language server, we likely want to take every ms.

dsherret (commented):
Apologies for the massive PR, but at least it removes 700 lines of code from the CLI 🙂

cli/file_fetcher.rs
```diff
@@ -52,7 +52,7 @@ pub struct File {
   /// The resolved media type for the file.
   pub media_type: MediaType,
   /// The source of the file as a string.
-  pub source: String,
+  pub source: Arc<String>,
```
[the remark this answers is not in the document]
bartlomieju: That's reasonable 👍

cli/lsp/analysis.rs
```diff
@@ -131,17 +134,12 @@ fn as_lsp_range(range: &deno_lint::diagnostic::Range) -> Range {
...
 ) -> Result<Vec<Reference>, AnyError> {
-  let syntax = ast::get_syntax(media_type);
+  let syntax = deno_ast::get_syntax(parsed_source.media_type());
   let lint_rules = rules::get_recommended_rules();
   let linter = create_linter(syntax, lint_rules);
-  // TODO(@kitsonk) we should consider caching the swc source file versions for
-  // reuse by other processes
-  let (_, lint_diagnostics) =
-    linter.lint(specifier.to_string(), source_code.to_string())?;
+  let lint_diagnostics = linter.lint_with_ast(parsed_source);
```
bartlomieju, line 142: Nice

cli/lsp/analysis.rs
```diff
@@ -388,12 +396,15 @@ pub fn analyze_dependencies(
       {
         dep.maybe_type_specifier_range = Some(Range {
           start: Position {
-            line: (loc.line - 1) as u32,
-            character: (loc.col + 1) as u32,
+            line: loc.line_index as u32,
+            // +1 to skip quote
```
bartlomieju, line 400: I'm not sure I understand what this comment means

cli/lsp/code_lens.rs
```diff
@@ -370,36 +372,37 @@ pub(crate) async fn resolve_code_lens(
...
-      .ok_or_else(|| anyhow!("Missing text content: {}", specifier))?;
-    let media_type = language_server
-      .get_media_type(specifier)
-      .ok_or_else(|| anyhow!("Missing media type: {}", specifier))?;
-    // we swallow parsed errors, as they are meaningless here.
-    // TODO(@kitsonk) consider caching previous code_lens results to return if
-    // there is a parse error to avoid issues of lenses popping in and out
-    if let Ok(parsed_module) =
-      analysis::parse_module(specifier, &source, &media_type)
-    {
```
bartlomieju, line 397: This is way clearer

cli/lsp/analysis.rs
dsherret: The `+ 1` on the next line is to skip over the quote of the string literal. Do you have a suggestion for a better comment?

bartlomieju (commented):
Great to see so much code being removed

cli/lsp/analysis.rs
bartlomieju: Okay then in case of `end` shouldn't it be `-1` to skip the quote? Otherwise it will include the quote, won't it? Maybe I misunderstood something then just ignore the comment :)

kitsonk: Sorry I didn't comment it originally, but feels like we need a test here, as there was so much manipulation... Yeah we are trying to ignore the quotes around the string literal, but it brings into question what is happening on line 406.

cli/lsp/documents.rs
```diff
@@ -114,59 +118,39 @@ impl DocumentData {
     &mut self,
     content_changes: Vec<lsp::TextDocumentContentChangeEvent>,
   ) -> Result<(), AnyError> {
-    if self.bytes.is_none() {
-      return Ok(());
-    }
-    let content = &mut String::from_utf8(self.bytes.clone().unwrap())
-      .context("unable to parse bytes to string")?;
```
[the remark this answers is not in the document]
kitsonk: It was being stored as bytes because that is what rust-analyzer did, and I didn't know how we were going to deal with it and handle it, so it was best to copy that pattern and figure out challenges later.

kitsonk (approved):
LGTM

cli/lsp/analysis.rs
dsherret: No, `loc.column_index + 1` is the start and then it's `start + specifier.chars().count()` to get the end. What I'll do is extract out `loc.column_index + 1` to a common variable (but will do that tomorrow).

dsherret: Oh, wait, I see what you mean. I think `specifier` doesn't include the quotes, but `loc` does IIRC. I will investigate it tomorrow.

dsherret: Considering this code hasn't changed and that we're going to replace this logic with what's in deno_graph, I'm just going to resolve this comment and merge.

Issue #11409, opened 2021-07-15
Feature request: Run async tests in parallel within the same file


voltrevo:
Previous related issues:
- https://github.com/denoland/deno/issues/6559
- https://github.com/denoland/deno/issues/7615 (marked as duplicate of 6559)

These seem to be considered resolved via the `-j` flag implemented by https://github.com/denoland/deno/pull/9815.

`-j` is definitely a welcome improvement. However, it's only enabling separate test files to run parallel. I believe instead the expectation is that parallelism be supported *within the same file*. In any case, I would like to request this feature.

For example:

```ts
// test/blah.ts

Deno.test("foo", async () => {
  console.log("start foo");
  await new Promise((resolve) => setTimeout(resolve, 1000));
  console.log("end foo");
});

Deno.test("bar", async () => {
  console.log("start bar");
  await new Promise((resolve) => setTimeout(resolve, 1000));
  console.log("end bar");
});
```

Running with `-j4`:

```
$ deno test -j4 test/blah.test.ts
running 2 tests from file:///home/voltrevo/workspaces/ethf/aggregator/test/blah.test.ts
start foo
end foo
test foo ... ok (1008ms)
start bar
end bar
test bar ... ok (1006ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (2031ms)
```

If parallelism was enabled between these tests, we should expect output more like this:

```
running 2 tests from file:///home/voltrevo/workspaces/ethf/aggregator/test/blah.test.ts
start foo
start bar
end foo
test foo ... ok (1008ms)
end bar
test bar ... ok (1006ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (1031ms)
```

(Implementing this might be considered a breaking change, requiring a new flag, perhaps `-jj`.)

Can you confirm that this kind of parallelism is not yet supported or advise how to enable it?

If it's not supported, are there any intentions to implement this?

(It's also possible to enable this kind of concurrency for synchronous tests, by running a first pass to identify test names and then running separate processes with different filters. Can this also be considered?)

caspervonb:
> -j is definitely a welcome improvement. However, it's only enabling separate test files to run parallel. I believe instead the expectation is that parallelism be supported within the same file. In any case, I would like to request this feature.

The design I've had in mind for a while now is adding a `parallel` boolean to TestDefinition which is a hint to the runner that the test can be ran in parallel.

[...]

PR #11789, opened 2021-08-20
feat: add Deno.TestDefinition.concurrent option

caspervonb:
This adds a concurrent option to Deno.TestDefinition which when enabled is used to partition the test into compatible grouping which it may run in concurrently with.

When a test is running concurrently the sanitizers and permissions are applied to the entire concurrency group as an execution unit.

Sequential tests are run before concurrent tests in their natural order.

Prior art: 
- https://jestjs.io/docs/api#testconcurrentname-fn-timeout
- https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#running-tests-serially (inverted)

Closes https://github.com/denoland/deno/issues/11409


caspervonb:
Logical follow up to this would be to create a single stream unbuffered stream of concurrent jobs from all the workers.

caspervonb:
cc @bartlomieju

bartlomieju:
> cc @bartlomieju

@caspervonb I believe we should wait on https://github.com/denoland/deno/issues/11901 before we land this PR. I'm afraid they might clash in fundamental way.

caspervonb:
> > cc @bartlomieju
> 
> @caspervonb I believe we should wait on #11901 before we land this PR. I'm afraid they might clash in fundamental way.

Am 100% confused as what that issue is about or trying to resolve, but the steps proposal does not handle concurrency for tests.

bartlomieju:
> > > cc @bartlomieju
> > 
> > 
> > @caspervonb I believe we should wait on #11901 before we land this PR. I'm afraid they might clash in fundamental way.
> 
> Am 100% confused as what that issue is about or trying to resolve, but the steps proposal does not handle concurrency for tests.

That issue will a plan to implement both presented proposals (imperative and declarative). Concurrency was discussed for both proposals and will be supported.

bartlomieju:
@caspervonb thank you for the PR and effort, but I'm going to close it without a merge as it became quite stale and it's not obvious how we could integrate this functionality with test-steps API.

PR #11791, opened 2021-08-20
fix(ext/ffi): Panic when arguments are of incorrect type in Deno.dlopen

### LINKED ISSUE #11741: Panic when arguments are of  incorrect type in Deno.dlopen
Any of the following statements panic 

```
let d=Deno.dlopen('/usr/lib/libc.so.6',{malloc:null})
let d=Deno.dlopen('/usr/lib/libc.so.6',null)
let d=Deno.dlopen('/usr/lib/libc.so.6')
let d=Deno.dlopen('/usr/lib/libc.so.6',{malloc:{parameters:['a'],result:'b'}})
```

F3n67u:
fix #11741


ry (commented):
Thanks for the patch @F3n67u ! Can you please provide a test?

F3n67u:
> Thanks for the patch @F3n67u ! Can you please provide a test?

Yes. Added a typescript unit test. but I am not sure if the test file is in the right directory.

F3n67u:
Btw,  when test invalid argument type which ts-check will complain, what's the best practice? Now I use `as any` type assertion to make ts-check pass.

bartlomieju:
> Btw, when test invalid argument type which ts-check will complain, what's the best practice? Now I use `as any` type assertion to make ts-check pass.

You can use `//@ts-ignore` or `//@ts-expect-error` directives, or just write the test case in JavaScript

F3n67u:
> > Btw, when test invalid argument type which ts-check will complain, what's the best practice? Now I use `as any` type assertion to make ts-check pass.
> 
> You can use `//@ts-ignore` or `//@ts-expect-error` directives, or just write the test case in JavaScript

Thanks for your advice. I opted for `//@ts-expect-error` because:
1. `@ts-ignores` is easier to be forgotten about than `//@ts-expect-error`: see  https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-ts-expect-error.md
2. Test case written in Typescript is more consistent with other unit test in `unit` dir and latter added dlopen related unit test will benefit from typescript type check.

eliassjogreen (approved):
LGTM

lucacasonato:
Thanks for the effort @F3n67u, but I managed to solve this in Rust instead in #11815 (less overhead). I have added you as a co-author in that PR.

PR #11800, opened 2021-08-21
perf(ops): optimize permission check

AaronO:
Removes the overhead of permission check (logging) on access granted (common case):

Delta measured on `perf_now` from `deno_common` bench:

```
... Before ...
perf_now:            	n = 500000, dt = 0.264s, r = 1893939/s, t = 528ns/op
... After ...
perf_now:            	n = 500000, dt = 0.083s, r = 6024096/s, t = 166ns/op
```

So ~3x faster on `op_now` (which is a low-overhead op)

## Why is this faster ?

Avoids "expensive" `format` calls if no debug logging is enabled


MierenManz:
@AaronO i think you may got the before and after values wrong. Unless it got about 3x slower 😅

AaronO:
> @AaronO i think you may got the before and after values wrong. Unless it got about 3x slower 😅

Good catch !

runtime/permissions.rs
```diff
@@ -37,14 +41,16 @@ pub enum PermissionState {
...
-      ))
-    );
+    if *DEBUG_LOG_ENABLED {
+      log::debug!(
+        "{}",
+        colors::bold(&format!(
+          "{}️  Granted {}",
+          PERMISSION_EMOJI,
+          Self::fmt_access(name, info)
+        ))
```
dsherret, line 51: > Avoids "expensive" format calls if no debug logging is enabled

I'm not sure this is the case. The `log::debug!` macro should put the expression inside an if statement when compiled https://github.com/rust-lang/log/blob/bd28b80a6d2b90cb9131e21bc602df68f4f72fdb/src/macros.rs#L34

Do you think we could make a change downstream to make this faster instead of doing this code? I feel like this should be handled by the macro instead of us doing this ourselves.

AaronO: In theory we could also set the `max_level` correctly when it's disabled, didn't inspect all the code paths to see why that's not happening.

Will pick this up again in a few days

AaronO: Will merge this and then explore a more general upstream fix

dsherret: Would you be able to add a comment saying why this is temporarily done?

AaronO: Added a TODO + rationale comment

dsherret (approved):
LGTM

Issue #11799, opened 2021-08-21
(cli/test): `deno test --jobs` currently accepts `0`


cmorten:
`deno test --jobs` currently allows you to set `0` as the number of workers, which results in the command hanging. E.g.

```console
$ deno test -A --unstable --jobs 0 # hangs
```

Can't foresee as a big issue, but perhaps the CLI should error out with a useful message in the same fashion as currently happens if pass other invalid values. E.g. as with `a`...

```console
$ deno test -A --unstable --jobs a
error: Invalid value for '--jobs <jobs>': jobs should be a number
```

where could change to something like...

```console
error: Invalid value for '--jobs <jobs>': jobs should be a positive integer
```

REF: https://github.com/denoland/deno/blob/main/cli/flags.rs#L1106

PR #11802, opened 2021-08-21
fix(cli/flags): require a non zero usize for concurrent jobs

caspervonb:
Closes https://github.com/denoland/deno/issues/11799


bartlomieju (commented):

bartlomieju (approved):
LGTM

PR #11803, opened 2021-08-22
refactor(cli/tools/test): infer disable log from program state

caspervonb:
Infers if the log should be disabled from the program state's flags rather than parsing the `--quiet` flag twice and passing a boolean around everywhere.


bartlomieju:
Please rebase

bartlomieju (approved):

PR #11804, opened 2021-08-22
refactor(cli/flags): use an optional non zero usize for `fail-fast`

caspervonb:
Changes the type of the `fail_fast` flag from `Option<usize>` to `Option<NonZeroUsize>` as an optional value of zero isn't sound.


bartlomieju (commented):

bartlomieju (approved):

Issue #11743, opened 2021-08-17
[ext/http]: responseSender open resource handle after connection and listener close


cmorten:
**Description**

Deno: `1.13.1`

Closing both the connection and underlying listener is not sufficient to close all open resource handlers for a request event, specifically a `responseSender` resource is left hanging.

Ideally for general resource cleanup (and useful for tests), Deno would clean up dud `responseSender` resources if the underlying connection and listener has been closed?

Currently to get around this you have to store a reference to the request event so you can call `requestEvent.respondWith(new Response());` even though you know it will throw a `Deno.errors.BadResource` error.

This _feels_ similar to a bug that @bartlomieju mentioned the other day, but that was for a `responseBody` resource.

**Steps to reproduce**

1. Save and run the following code snippet `deno run -A repro.ts`:

    ```ts
    const listener = Deno.listen({ port: 4505 });
    const conn = await listener.accept();
    const httpConn = Deno.serveHttp(conn);
    const requestEvent = await httpConn.nextRequest();
    
    httpConn.close();
    listener.close();
    
    console.log(Deno.resources());
    ```

1. While the process is running, make a request to `localhost:4505`, e.g. `curl http://localhost:4505`

1. Observe that the following is logged showing the resource that is not cleaned up.

    ```console
    { "0": "stdin", "1": "stdout", "2": "stderr", "6": "responseSender" }
    ```

**Related**

- https://github.com/denoland/deno_std/pull/1128/files#diff-dd523d06e796c772903dc7b7b2211ac9a7f8eceeeb160258acbda240444187a8R374
- https://github.com/denoland/deno/issues/11595#issuecomment-899414934

bartlomieju:
Thanks Craig I can repeat the problem. I'll fix it for 1.13.2

PR #11805, opened 2021-08-22
fix(ext/http): resource leak on HttpConn.close()

bartlomieju:
Closes https://github.com/denoland/deno/issues/11743

This is implemented in JavaScript, because we don't have necessary infrastructure
in Rust.

### DIFF
--- a/ext/http/01_http.js
+++ b/ext/http/01_http.js
@@ -23,8 +23,11 @@
     ArrayPrototypeIncludes,
     ArrayPrototypePush,
     ArrayPrototypeSome,
-    ObjectKeys,
     Promise,
+    Set,
+    SetPrototypeAdd,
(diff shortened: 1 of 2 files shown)


ext/http/01_http.js
```diff
@@ -38,6 +39,11 @@
 
   class HttpConn {
     #rid = 0;
+    // This object holds resource ids of resources
+    // that were created during lifecycle of this request.
+    // When the connection is closed these resources should be closed
+    // as well.
+    managedResources = {};
```
lucacasonato, line 46: Maybe a `Set` would make more sense?

bartlomieju: Sure, I can use `Set` for that

lucacasonato (approved):

Issue #11786, opened 2021-08-20
`deno check`


ry:

kitsonk:
Ref: #11786

These two are closely related and will be addressed as part of that feature.

There is an open question in my mind if `deno cache` becomes a redundant sub-command.

nayeemrmn:
> There is an open question in my mind if `deno cache` becomes a redundant sub-command.

Rather, I think `deno cache --check` (#11810) makes `deno check` 100% redundant while the reverse doesn't hold. If we favoured `deno check`, we'd have an annoying situation where `deno cache` is _mostly_ redundant but still needed for the effect of `deno cache [--no-check]`, unable to be deprecated. Alternatively we could forgo this behaviour insisting that people use `deno check` and just ignore the error if they want to, allowing us to deprecate `deno cache`, but this seems weak.

So I think we should drop this proposal in favour of `--check` as currently implemented in #11810.

jaydenseric:
Related discussion: https://github.com/denoland/deno/discussions/12113 .

Deno absolutely needs a CLI command to type check all modules in a project, like how the `deno lint` command works.

rivy:
I'm currently using `deno test --no-run` ... one current benefit of `deno test --no-run FILE1 FILE2` over `deno cache FILE1 FILE2` is that the "test" command actually type checks all the given FILE arguments even if there's an error for one or more, whereas the "cache" alternate stops after parsing a FILE with any errors. The "test" variant is much more efficient when tracking down (possibly multiple) errors.

Whatever the final syntax/type-check variant that `deno` receives, please enable the "test"-type check-all-files behavior.

ry:
[Luca's comment](https://github.com/denoland/deno/issues/14071#issuecomment-1075255767) in a duplicate issue seems reasonable.

- Add `deno check`. It should be exactly the same thing as `deno cache`. 
- Add `deno run --check` for forward compatibility. (It will be required in the future for type-checking.)
- Print "Warning: In future releases deno run will not automatically type check without the --check flag. To opt into this new behavior now, specify DENO_FUTURE_CHECK=1."

The proposed schedule is:
1.21 - we introduce deno check, as described here.
1.22 - do nothing
1.23 - fully disable type checking by default

bartlomieju:
Discussed offline. `--check` should default to type checking only local files; to type check all files (including remote) we should do `--check=all`. Similar to how `--no-check` skips all, or `--no-check=remote` type checks only local files.

PR #11810, opened 2021-08-23
[DRAFT] feat: skip type checking by default

kitsonk:
Closes: #11340

This PR:

- adds the `--check` flag
- deprecates the `--no-check` flag
- does not type check code by default on subcommands
- logs a warning to stderr that the behaviour has changed and indicates that `--check` can be used to type check

The tests still need to be updated for not type checking by default.

A couple of points of discussion:

- Right now, `deno test` does not type check by default.  It be more logical to implicitly enable `--check` for `deno test`. This does mean this subcommand would differ in default flags than other subcommands, and we would not remote `--no-check` for `deno test`.
- The behavior of `deno cache` reflects the others, skipping type checking by default. We need to determine how it all relates to #11786 (`deno check`).  I don't think `deno check` would replace the use of `--check` flag, as there are lots of subcommands where people would not want to have to invoke 2 subcommands just to check their documents, as well as there is the challenge that we implicitly type check worker scripts contextually, so unless we add flags to `deno check` for the type of code it is checking, some code will fail.
- The way I am managing the `warn_check` flag in `ProgramState` feels hacky to me, but I don't have any great ideas how better to manage it.  We need it, as there are certain commands where the methods get called > 1 times, causing multiple warnings to be printed if we don't have a flag.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -146,6 +146,7 @@ pub struct Flags {
   /// the language server is configured with an explicit cache option.
   pub cache_path: Option<PathBuf>,
   pub cached_only: bool,
+  pub check: bool,
   pub config_path: Option<String>,
   pub coverage_dir: Option<String>,
   pub enable_testing_features: bool,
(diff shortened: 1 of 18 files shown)


caspervonb:
> Right now, deno test does not type check by default. It be more logical to implicitly enable --check for deno test. This does mean this subcommand would differ in default flags than other subcommands, and we would not remote --no-check for deno test.

Consistency with the common options is better here, type checking failing doesn't mean that the actual runtime behaviour will be affected, same as run.

bartlomieju:
@kitsonk should we close this PR in favor of https://github.com/denoland/deno/pull/12766?

PR #11815, opened 2021-08-23
fix(ext/ffi): don't panic on invalid enum values

### LINKED ISSUE #11741: Panic when arguments are of  incorrect type in Deno.dlopen

lucacasonato:
This fixes #11741. It is an alternative to #11791 that doesn't add webidl things to Deno namespace APIs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3280,9 +3280,9 @@ dependencies = [
 
 [[package]]
 name = "serde_v8"
-version = "0.9.2"
+version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


bartlomieju (commented):
Looks good, but a test would be useful

bartlomieju (approved):

lucacasonato:
Blocked by https://github.com/denoland/serde_v8/issues/20

PR #11820, opened 2021-08-23
chore: release crates for 1.13.2

dsherret:


lucacasonato (changes requested):
deno_core needs a minor bump (rather than patch), because public API changed (`handleAsyncMsgFromRust()` renamed to `opresolve()`)

lucacasonato (approved):

bartlomieju (approved):

PR #11827, opened 2021-08-24
chore(docker): add `Dockerfile` to build `deno` from source

felipecrs:
I am sending this PR as a PoC for opening the discussion -- it's not yet ready. The goal is to bring a more cloud-native build approach for the main `deno` repository. I have some experience with docker and I would like to contribute, and made a couple of PRs to the `deno_docker` repository as well (https://github.com/denoland/deno_docker/pull/158 and https://github.com/denoland/deno_docker/pull/159).

In the state this PR is, the benefits it brings is:

- Anyone can build the `deno` **docker** image from source **having the repository cloned** by running `docker buildx bake`.
- Anyone can build the `deno` **binary** from source **having the repository cloned** by running `docker buildx bake export-bin` (it will place the `deno` binary in the current dir).
- Anyone can build the `deno` **docker** image from source **without** having the repository cloned by running `docker buildx build git://github.com/denoland/deno`
- Anyone can build the `deno` **docker** binary from source **without** having the repository cloned by running `docker buildx build --target bin --output . git://github.com/denoland/deno#main` or `#v1.13.2` to build a given version.

But this pattern could be further applied to other areas, such as building the releases artifacts (with the benefit of docker providing isolation and environment awareness builds).

Ultimately, I believe the `deno_docker` could be moved to the main `deno` repository, thus keeping the files closer and making it easier to handle the release of new versions (a single commit to rule them all).

## Test:

```
docker buildx build --target bin --output . git://github.com/felipecrs/deno
```

## Demo

[asciicast](https://asciinema.org/a/KqoSafTl5G5NCKH7wNe2sxY68)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+target/
+
+.dockerignore
+docker-bake.hcl
+Dockerfile
+
+deno
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ cli/tests/.test_coverage/
 # WPT generated cert files
 /tools/wpt/certs/index.txt*
 /tools/wpt/certs/serial*
+
+deno
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,60 @@
+# syntax=docker/dockerfile:1.3
+
+ARG RUST_VERSION=1.54.0
+
+
+FROM rust:${RUST_VERSION} AS deps
+
+WORKDIR /src
+
+RUN --mount=type=bind,target=.,rw \
+  --mount=type=cache,target=/usr/local/cargo/registry \
+  cargo fetch
+
+
+FROM deps AS build
+
+RUN --mount=type=bind,target=.,rw \
+  --mount=type=cache,target=./target \
+  --mount=type=cache,target=/usr/local/cargo/registry \
+  cargo build --release \
+  && mkdir -p /out && cp -Rf target /out/
+
+WORKDIR /out
+
+
+FROM scratch AS bin
+
+COPY --from=build /out/target/release/deno /deno
+
+
+FROM debian:latest AS debian
+
+COPY --from=bin /deno /usr/local/bin/deno
+
+CMD [ "deno", "run", "https://deno.land/std/examples/welcome.ts" ]
+
+
+# debian is the default
+FROM debian
+
+
+FROM ubuntu:latest AS ubuntu
+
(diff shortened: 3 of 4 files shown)


wperron:
> this pattern could be further applied to other areas, such as building the releases artifacts (with the benefit of docker providing isolation and environment awareness builds).

From what I understand from this PR, you are proposing that we change the build mechanism from simply running `cargo build` to running that in a Docker container? We're already building using `cargo build --release --locked --all-targets` so all dependencies are locked at build time, providing repeatability. I could see the appeal for a project using a dynamic language that needs to have a runtime with a specific version or something like C++ that relies heavily on dynamically linked libraries, but here we use Rust specifically so that all of our dependencies are statically linked. Requiring Docker to be installed to contribute and build the project is going to be a tough sell here, unless you can come up with a _really_ convincing argument for it, I'm going to vote "no" for this PR.

> Ultimately, I believe the deno_docker could be moved to the main deno repository, thus keeping the files closer and making it easier to handle the release of new versions (a single commit to rule them all).

See comment above. Our goal to make a self-contained binary that can just be downloaded on its work and Just Work :tm: . Docker support is a nice to have obviously, but as long as the Deno binary is compiled to be self-contained, keeping all Docker-related material in a separate repo is trivial and helps keep this repo as clean as possible (it's already pretty large and complex).

> The goal is to bring a more cloud-native build approach for the main deno repository.

I'm all for Cloud Native! The Docker & Kubernetes ecosystems are alive and kicking and supporting them is important, at least to me. _However_ as I mentioned earlier, I think we can easily support the Cloud Native ecosystem _outside_ of this repository. What's your reasoning for wanting to include all Docker-related material in this repo?

lucacasonato:
Thanks @felipecrs for the PR.

I discussed with @bartlomieju, and we decided that we don't want to add this Dockerfile:
- We already have a Dockerfile to aid in local dev: https://github.com/denoland/deno/blob/main/.devcontainer/Dockerfile
- Building Deno from source at a given version already trivial `cargo install deno --version 1.13.2`. This can also be trivially run in a Dockerfile (just `FROM rust:1.54`).

To your second point about moving `deno_docker` to this repo. This is not a path we want to pursue. CI times are already slow enough on this repo, so adding extra stuff to CI here is not good for productivity. Keeping deno_docker in a seperate repo allows us to separately run CI.

felipecrs:
No worries. Thanks for the discussion, guys.

PR #11830, opened 2021-08-24
fix(ext/http): websocket upgrade header check

### LINKED ISSUE #11747: Lowercase on websocket upgrade header check
I have a websocket server with deno 1.13 and it started giving the following error when connecting:

```
Invalid Header: 'upgrade' header must be 'websocket'
```

I saw that the update check is case sensitive:

https://github.com/denoland/deno/blob/4b800f5d269de259da29d3b42dae66faf4351476/ext/http/01_http.js#L320

I know the default is 'websocket', but some clients are third-party and it's hard to change their codes.

Is It possible considering a upgrade header with no case sensitive?

galiazzi:
Fixes #11747

### DIFF
--- a/ext/http/01_http.js
+++ b/ext/http/01_http.js
@@ -342,7 +342,7 @@
 
   function upgradeWebSocket(request, options = {}) {
     const upgrade = request.headers.get("upgrade");
-    if (!upgrade || upgrade.toLowerCase() !== "websocket") {
+    if (!upgrade || StringPrototypeToLowerCase(upgrade) !== "websocket") {
       throw new TypeError(
         "Invalid Header: 'upgrade' header must be 'websocket'",
       );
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
Thanks for the PR @galiazzi, any chance we could get a test exercising this change?

ext/http/01_http.js
```diff
@@ -341,7 +341,8 @@
   const _ws = Symbol("[[associated_ws]]");
 
   function upgradeWebSocket(request, options = {}) {
-    if (request.headers.get("upgrade") !== "websocket") {
+    const upgrade = request.headers.get("upgrade");
+    if (!upgrade || upgrade.toLowerCase() !== "websocket") {
```
lucacasonato, line 345: Please use "primordials" to do this check. You can import `StringPrototypeToLowerCase` (see top of file how), and then do `StringPrototypeToLowerCase(upgrade) !== "websocket"`.

lucacasonato (commented):
Looks good, I just have one small change request.

bartlomieju (approved):
LGTM, thank you @galiazzi

PR #11831, opened 2021-08-24
feat(core): add a `resolve_value` method

caspervonb:
This adds a `resolve_value` function which will return wait for a value to resolve and return it.
If the value is a promise and it rejects, that is returned as the error.

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -1708,7 +1708,11 @@ pub mod tests {
     let value_global = runtime
       .execute_script("a.js", "new Promise(resolve => {})")
       .unwrap();
-    let error_string = runtime.resolve_value(value_global).await.unwrap_err().to_string();
+    let error_string = runtime
+      .resolve_value(value_global)
+      .await
+      .unwrap_err()
+      .to_string();
     assert_eq!(
       "Promise resolution is still pending but the event loop has already resolved.",
       error_string,
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -648,6 +648,38 @@ impl JsRuntime {
     scope.perform_microtask_checkpoint();
   }
 
+  /// Waits for the given value to resolve while polling the event loop.
+  ///
+  pub async fn resolve_value(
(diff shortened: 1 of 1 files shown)


core/runtime.rs
```diff
@@ -648,6 +648,38 @@ impl JsRuntime {
...
   }
 
+  /// Waits for the given value to resolve while polling the event loop.
+  ///
+  pub async fn resolve_value(
+    &mut self,
+    global: v8::Global<v8::Value>,
+  ) -> Result<v8::Global<v8::Value>, AnyError> {
+    poll_fn(|cx| {
+      self.poll_event_loop(cx, false);
```
caspervonb, line 658: Suppose we should bail if the main loop poll yields `Poll::Ready` as nothing can progress?

core/runtime.rs
```diff
@@ -1629,6 +1661,42 @@ pub mod tests {
...
+      let value = result_global.get(scope);
+      assert_eq!(value.integer_value(scope).unwrap(), 4);
+    }
+
+    let value_global = runtime
+      .execute_script("a.js", "Promise.reject(new Error('fail'))")
+      .unwrap();
+    let err = runtime.resolve_value(value_global).await.unwrap_err();
+    assert_eq!(
+      "Uncaught Error: fail",
```
caspervonb, line 1695: Preferably this wouldn't be treated as uncaught
```suggestion
      "Error: fail",
```

core/runtime.rs
bartlomieju: Yes

core/runtime.rs
bartlomieju: I'm afraid this will require to add a lot of conditionals to handle it just for this helper function

bartlomieju:
Could we combine this PR with https://github.com/denoland/deno/pull/11727 as it's directly required?

bartlomieju:
Thanks, closing in favor of #11727

PR #11832, opened 2021-08-24
chore: add scripts for helping with a release

dsherret:
These scripts help automate more parts of the release process. See *cut_a_release.md* for the process change.

It would probably be good if I did the next release as well so I can fix any issues I might encounter when doing an actual release with this.

### DIFF
--- a/tools/release/02_publish_depenency_crates.ts
+++ b/tools/release/02_publish_depenency_crates.ts
@@ -1,0 +1,11 @@
+#!/usr/bin/env -S deno run --allow-read --allow-write --allow-run=cargo --allow-net=crates.io
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { DenoWorkspace, getCratesPublishOrder } from "./helpers/mod.ts";
+
+const workspace = await DenoWorkspace.load();
+
+const dependencyCrates = workspace.getDependencyCrates();
+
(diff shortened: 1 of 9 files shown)


tools/release/02_publish_depenency_crates.ts
```diff
@@ -0,0 +1,11 @@
...
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { DenoWorkspace, getCratesPublishOrder } from "./helpers/mod.ts";
+
+const workspace = await DenoWorkspace.load();
+
+const dependencyCrates = workspace.getDependencyCrates();
+
+for (const crate of getCratesPublishOrder(dependencyCrates)) {
+  await crate.publish();
+}
```
bartlomieju, line 11: So I'm worried that when this fails it's gonna be hard to retry it for specific crate - I can see you guarded against it by checking if crate is already published. I have one more suggestion - "publish" should retry up to 3 times before hard failing - it happens quite often to me that after publishing a crate, next crate publish fails because `crates.io` index hasn't refreshed. WDYT?

dsherret: Done in 610e2a78613096888e2bb81bd614527059724e8a

bartlomieju (approved):
LGTM, let's try this out

PR #11835, opened 2021-08-24
refactor(webgpu): use op interface idiomatically

lucacasonato:
This commit reorganizes some code in ext/webgpu to clean up some cruft. WebIDL enums are now represented as Rust enums in Rust.

Also cleans up some code around `unsafe` blocks and gives them safety comments.


bartlomieju (approved):
LGTM, very nice cleanup

PR #11839, opened 2021-08-25
fix(Blob): fix Blob SymbolToStringTag

liruifengv:


lucacasonato:
Thanks @liruifengv, but we should change this for all `SymbolToStringTag` usages, not just for `Blob`. Maybe we could set up `SymbolToStringTag` in the `webidl.configurePrototype` function? We already call that for all prototypes.

PR #11851, opened 2021-08-26
fix(ext/webidl): correctly apply [SymbolToStringTag] to interfaces

liruifengv:
#11839


liruifengv:
thx

liruifengv:
This is my first time contributing code to Deno. This has not been merged for a long time. Did I miss something? @ @lucacasonato

kitsonk:
CI isn't green, looks like @lucacasonato tried to fix it, but wasn't able to, and it can't be merged until it goes green.

lucacasonato:
@liruifengv Sorry for the delay! I'll review shortly

kt3k:
Instance fields, such as `[SymbolToStringTag] = "DedicatedWorkerGlobalScope";` don't set property to the prototype of the class, and that seems causing CI failure.

I think we should revert the removal of [SymbolToStringTag] getter for `Window`, `WorkerGlobalScope`, `DedicatedWorkerGlobalScope`

kt3k:
@lucacasonato Updated sting tags of `FileReader`, `SubtleCrypto`, and global scopes. PTAL

ext/web/10_filereader.js
```diff
@@ -363,6 +364,8 @@
       // alias for readAsArrayBuffer
       this.#readOperation(blob, { kind: "Text", encoding });
     }
+
+    [SymbolToStringTag] = "FileReader";
```
lucacasonato, line 437: This is weird... there is a `webidl.configurePrototype` already. Maybe because it inherits from EventTarget, which does not use `webidl.configurePrototype` yet?

lucacasonato (approved):
LGTM

ext/web/10_filereader.js
kt3k: You are right. EventTarget wasn't configured

kt3k:
@liruifengv Thank you for your contribution!

Issue #11597, opened 2021-08-06
[Core] No functionality to de-register an Op.


TkTech:
As discussed on discord, we can `register_op` on a `JsRuntime`, but there's no ability to `deregister_op`. It's occasionally desired to de-register or overwrite the Op registered by another extension, such as the core `op_print` which is enabled by default.

This is really necessary as there is no way that I can see to disable adding the builtins. https://github.com/denoland/deno/blob/c909faf9e6cd2964398da7c0852d0229cdd1a22b/core/runtime.rs#L348. For example, I might want to replace `op_print` for existing extensions with one that captures logs without having to completely redirect stdout and stderr.

thinkier:
I would also like to replace `op_print` with my own function. I tried to register it directly and the problem appears to be from

https://github.com/denoland/deno/blob/main/core/ops.rs#L137

Where overwriting is "disabled" by an `assert!` post-facto, resulting in a panic. In theory, capturing the panic using [std::panic::catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) would let you continue on with the updated `op_print`. Unfortunately there's over a dozen `!UnwindSafe` types being held by the `JsRuntime`, so a more creative method of playing with with the `OpTable` is needed.

I'm going to sleep after writing this note and I'll post an update if I find a solution.

thinkier:
Update: I wiped the `OpTable` without skating on *very* unsafe territory. However, this unregisters the other builtins and all other ops. (Which may or may not be your intended behaviour.)

```rust
    js_runtime.op_state().borrow_mut().op_table = OpTable::default();
```

bartlomieju:
CC @AaronO

AaronO:
This is easy to fix per se, but it would be helpful to understand the use-case & broader context.
(I had discussed allowing op overrides a few months back: https://discord.com/channels/684898665143206084/778060818046386176/837333408602259496)

Most embedders will indeed want to override `op_print` to redirect `console.log()` output and co, however `Console`'s constructor already accepts a `printFunc` and thus isn't inherently coupled to `op_print` / `Deno.core.print()`, see:
- https://github.com/denoland/deno/blob/cb8f34c325e66d1ed160d65e515f8afe3393cfaf/runtime/js/99_main.js#L422
- https://github.com/denoland/deno/blob/cb8f34c325e66d1ed160d65e515f8afe3393cfaf/ext/console/02_console.js#L1661

[...]

PR #11858, opened 2021-08-27
feat(core): facilitate op-disabling middleware

AaronO:
By providing `void_op_sync()` and `void_op_async()` as well as `core/examples/disable_ops.rs`

Fixes https://github.com/denoland/deno/issues/11597 (technically this just makes it easier & provides an example)


bartlomieju (approved):

AaronO:
@bartlomieju Renamed these to use the more explicit `void_` prefix

PR #11860, opened 2021-08-27
cleanup(core): AsyncOpIterator

AaronO:
Replaces `poll_pending_ops()` with `AsyncOpIterator`, avoiding allocating that temporary `Vec` used to collect them.

Impact on `op_baseline > async` bench is minor, ~1-2% at most, nearly identical on most runs down to a few ns, but there appears to be a slight reduction in variance.

The main reason why the impact is minor, is that it's not saving 1 alloc per async op response, but only one "large" alloc per event loop iteration, since we batch op responses and can return multiple responses in a single event loop pass.

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -1505,10 +1505,6 @@ impl JsRuntime {
     let js_recv_cb_handle = state_rc.borrow().js_recv_cb.clone().unwrap();
     let scope = &mut self.handle_scope();
 
-    // Now handle actual ops.
-    let mut state = state_rc.borrow_mut();
-    state.have_unpolled_ops = false;
-
     // We return async responses to JS in unbounded batches (may change),
     // each batch is a flat vector of tuples:
     // `[promise_id1, op_result1, promise_id2, op_result2, ...]`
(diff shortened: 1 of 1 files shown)


core/runtime.rs
```diff
@@ -1403,35 +1382,43 @@ impl JsRuntime {
   }
 
   // Send finished responses to JS
-  fn async_op_response(
-    &mut self,
-    async_responses: Vec<(PromiseId, OpResult)>,
-  ) -> Result<(), AnyError> {
+  fn async_op_response(&mut self, cx: &mut Context) -> Result<(), AnyError> {
```
bartlomieju, line 1385: I suggest to change the name of this function to `poll_pending_ops`

bartlomieju (approved):
@AaronO feel free to land it if you want, but please rename `async_op_response` first

piscisaureus:
Have you benchmarked it with a large-ish number of pending ops?
The purported benefit of FuturesUnordered is that it only polls elements that are woken (vs all pending ops).

AaronO:
> Have you benchmarked it with a large-ish number of pending ops? The purported benefit of FuturesUnordered is that it only polls elements that are woken (vs all pending ops).

This doesn't optimize op-polling (an alternative to FuturesUnordered I mentioned is another non-pushed local experiment), this optimizes op-collecting and simply removes that Vec we allocate to collect op responses, similarly we can also optimize out the alloc of Vec of args passed to opresolve, but these don't have a huge impact on performance since these are single "large" allocs that happen once per event-loop not once per async op call (so a few orders of magnitude less when under high load)

piscisaureus:
> simply removes that Vec we allocate to collect op responses,

That seems like a good idea.

PR #11864, opened 2021-08-28
feat(ext/crypto): support JWK export for HMAC

littledivy:
Towards #11690

### DIFF
--- a/ext/crypto/00_crypto.js
+++ b/ext/crypto/00_crypto.js
@@ -979,7 +979,7 @@
             }
             case "jwk": {
               // 1-3.
-              let jwk = {
+              const jwk = {
                 kty: "oct",
                 k: unpaddedBase64(innerKey.data),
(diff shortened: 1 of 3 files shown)


ext/crypto/00_crypto.js
```diff
@@ -970,10 +977,48 @@
...
+              // 8.
+              jwk.ext = key[_extractable];
+              // 9.
+              return jwk;
+            }
             default:
               throw new DOMException("Not implemented", "NotSupportedError");
           }
+          // TODO(@littledivy): Redundant break but deno_lint complains without it
+          break;
```
lucacasonato, line 1021: cc @magurotuna @bartlomieju

lucacasonato (approved):
LGTM. Thanks :-)

Unfortunate there are no WPT for this. We should consider upstreaming some.

PR #11865, opened 2021-08-28
feat(ext/crypto): implement HKDF operations 

littledivy:
Towards #11690


ext/crypto/00_crypto.js
```diff
@@ -1469,6 +1521,51 @@
 
         return buf.buffer;
       }
+      case "HKDF": {
+        // 1.
+        if (length == null || length == 0 || length % 8 !== 0) {
```
lucacasonato, line 1526:
```suggestion
        if (length === null || length === 0 || length % 8 !== 0) {
```

lucacasonato (approved):
JS looks good. @bnoordhuis PTAL too

ry (approved):
LGTM - thank you @littledivy

PR #11869, opened 2021-08-29
feat(ext/crypto): AES key generation

lucacasonato:
Support AES-CTR, AES-CBC, AES-GCM, and AES-KW in
SubtleCrypto#generateKey.

Towards #11690


ext/crypto/lib.rs
```diff
@@ -180,6 +180,18 @@ pub async fn op_crypto_generate_key(
 
       private_key
     }
+    Algorithm::AesCtr
+    | Algorithm::AesCbc
+    | Algorithm::AesGcm
+    | Algorithm::AesKw => {
+      let length = args.length.ok_or_else(not_supported)?;
+      let mut key_data = vec![0u8; length];
```
littledivy, line 188: Maybe do this to avoid alloc? 
```rust
const MAX_KEY_SIZE: usize = 256 / 8;
let mut key_data = [0; MAX_KEY_SIZE];
let key_data = &mut key_bytes[..length];
 ```

littledivy (approved):
haha beat me to it! :smile: https://github.com/littledivy/deno/tree/aes_gcm

ext/crypto/lib.rs
lucacasonato: This actually still requires an allocation, because at the end I have to call `.to_vec()`

bartlomieju (approved):
LGTM 🚀

PR #11870, opened 2021-08-29
feat(ext/crypto): generate ECDH keys

lucacasonato:
Add support for ECDH algorithm in SubtleCrypto#generateKey.

Towards #11690


ext/crypto/lib.rs
```diff
@@ -162,7 +162,7 @@ pub async fn op_crypto_generate_key(
 
       private_key.to_pkcs8_der()?.as_ref().to_vec()
     }
-    Algorithm::Ecdsa => {
+    Algorithm::Ecdsa | Algorithm::Ecdh => {
```
littledivy, line 165: The ECKeyPair is generated using ring::signature::**Ecdsa**KeyPair::generate_pkcs8. Is this correct?

lucacasonato: Yes. The spec says to generate an "Elliptic Curve key pair' for both cases. ECDSA and ECDH are two algorithms for encryption/decryption, and key agreement respectively. They both operate using EC keys though.

ry (approved):
LGTM - Nice!

PR #11878, opened 2021-08-30
fix(cli): retain path based test mode inference

caspervonb:
Collection keeps track of what is what depending on the predicate used (`collect_specifiers_with_test_mode`).

But seems I added an else branch that sets them all to TestMode::Both in the `fetch_specifiers_with_test_mode` (this shouldn't have been here) which is a secondary pass that is supposed to reduce unknown media types (like README.md) down to `TestMode::Documentation` so that no attempts to execute are made if invoked as `deno test README.md`.

Fixes https://github.com/denoland/deno_std/pull/1171#issuecomment-908409163

### DIFF
--- a/cli/tests/testdata/test/doc_only.out
+++ b/cli/tests/testdata/test/doc_only.out
@@ -1,0 +1,5 @@
+Check [WILDCARD]/test/doc_only/mod.ts$2-5.ts
+running 0 tests from [WILDCARD]/test/doc_only/mod.ts
+
+test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out ([WILDCARD])
+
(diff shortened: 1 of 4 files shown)


bartlomieju:
Great, but we need a regression test for it

caspervonb:
> Great, but we need a regression test for it

Yup coming up, just wanted to catch @kt3k before he did a bunch of workarounds :)

kt3k:
@caspervonb Thanks!

cli/tests/testdata/test/doc_only.out
```diff
@@ -0,0 +1,5 @@
+Check [WILDCARD]/test/doc_only/mod.ts$2-5.ts
```
caspervonb, line 1: In case it's not obvious, scenario here is 1 documentation test and 0 executable tests from the same source module.

bartlomieju (approved):
LGTM, thanks for fixing this Casper and thanks for in-depth description what went wrong.

@kt3k do you have any comments?

kt3k (approved):
@bartlomieju No comments from me. LGTM

Thank you @caspervonb !

PR #11879, opened 2021-08-30
chore: release scripts should update Cargo.lock file when bumping versions

dsherret:
I was showing Ryan the release scripts and he pointed out that we need to update the lock file when bumping the versions.

I looked into using `cargo generate-lockfile` (https://doc.rust-lang.org/cargo/commands/cargo-generate-lockfile.html) but it updated too much in the lockfile. Instead the solution is to just build, which is probably something we want to do anyway as it ensures it still builds.

### DIFF
--- a/tools/release/helpers/cargo.ts
+++ b/tools/release/helpers/cargo.ts
@@ -46,3 +46,17 @@ export async function publishCrate(directory: string) {
     throw new Error("Failed");
   }
 }
+
+export async function build(directory: string) {
+  const p = Deno.run({
+    cwd: directory,
+    cmd: ["cargo", "build", "-vv"],
+    stderr: "inherit",
+    stdout: "inherit",
+  });
+
+  const status = await p.status();
+  if (!status.success) {
+    throw new Error("Failed");
+  }
+}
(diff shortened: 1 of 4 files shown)


tools/release/helpers/cargo.ts
```diff
@@ -46,3 +46,17 @@ export async function publishCrate(directory: string) {
...
   }
 }
+
+export async function build(directory: string) {
+  const p = Deno.run({
+    cwd: directory,
+    cmd: ["cargo", "build", "-vv"],
+    stderr: "inherit",
+    stdout: "inherit",
+  });
```
bartlomieju, line 56: I usually use `cargo check` and it works fine for updating the lockfiles but is faster as it doesn't have to emit target binaries

dsherret: Oh, sweet. I'll do that instead (didn't know about it)

dsherret: Done in a91d72d1fc01f0fb6acabfedb510e1a3b4bffdd6

I've left the `build` methods because it might be useful for quickly scripting something on the fly.

bartlomieju (approved):
LGTM

PR #11880, opened 2021-08-30
feat(ext/crypto): export RSA as pkcs#8

littledivy:
towards #11690


ext/crypto/lib.rs
```diff
@@ -522,6 +523,96 @@ pub async fn op_crypto_verify_key(
...
+          let pk_info = rsa::pkcs8::PrivateKeyInfo {
+            attributes: None,
+            public_key: None,
+            algorithm: rsa::pkcs8::AlgorithmIdentifier {
+              // Spec wants the OID to be id-RSASSA-PSS (1.2.840.113549.1.1.10) but ring and RSA do not support it.
+              // Instead, we use rsaEncryption (1.2.840.113549.1.1.1) as specified in RFC 3447.
+              // Node, Chromium and Firefox also use rsaEncryption (1.2.840.113549.1.1.1) and do not support id-RSASSA-PSS.
+
+              // parameters are set to NULL opposed to what spec wants (see above)
+              oid: rsa::pkcs8::ObjectIdentifier::new("1.2.840.113549.1.1.10"),
```
littledivy, line 599: This makes us go off-spec but there is no other choice. Blink and Gecko also does not support  `id-RSASSA-PSS` https://www.chromium.org/blink/webcrypto#TOC-Supported-key-formats

Also, [See this note about interoperability](https://w3c.github.io/webcrypto/#SubtleCrypto-method-importKey).

littledivy: EDIT: won't really call off-spec :-)

ext/crypto/lib.rs
```diff
@@ -552,6 +554,137 @@ pub async fn op_crypto_verify_key(
...
+          let pk_info = rsa::pkcs8::PrivateKeyInfo {
+            attributes: None,
+            public_key: None,
+            algorithm: rsa::pkcs8::AlgorithmIdentifier {
+              // Spec wants the OID to be id-RSASSA-PSS (1.2.840.113549.1.1.10) but ring and RSA do not support it.
+              // Instead, we use rsaEncryption (1.2.840.113549.1.1.1) as specified in RFC 3447.
+              // Node, Chromium and Firefox also use rsaEncryption (1.2.840.113549.1.1.1) and do not support id-RSASSA-PSS.
+
+              // parameters are set to NULL opposed to what spec wants (see above)
+              oid: rsa::pkcs8::ObjectIdentifier::new("1.2.840.113549.1.1.1"),
```
ry, line 632: So probably we shouldn't support it either...

lucacasonato (commented):
Are there no WPT for this?

littledivy:
> Are there no WPT for this?

@lucacasonato There are. Will enable them in #11891 :-)

ext/crypto/lib.rs
littledivy: yup, that is the current behavior :+1:

lucacasonato (approved):
LGTM! Thanks :-)

PR #11884, opened 2021-08-31
ci: experiment mac & windows debug builds

AaronO:
This experiment evaluates reducing total CI time, by using debug mac/windows build on PR CI.

CI checks in aggregate are as slow as the slowest check, which is windows (& macOS 2nd) in our case.

## Data

- [macOS](https://github.com/denoland/deno/pull/11884/checks?check_run_id=3472067920): ~31mn (~11mn build, ~18mn test)
- [windows](https://github.com/denoland/deno/pull/11884/checks?check_run_id=3472067939): ~23mn (~14mn build, ~7mn test)

## Baselines

Currently total PR CI time is ~40mn (since it has to wait for the slowest job, windows), currently:
- macOS: ~30mn
- windows: ~40mn

## Notes

- The data measured here is **without cargo cache** (whereas release builds currently do use cargo cache)
- CI time is variable, depending mainly on what needs to be rebuilt, changes deeper in the dependency tree naturally require more crates to be rebuilt

## Conclusion

We could reduce our PR CI time from **~40mn to ~20-25mn**, by using debug builds for mac/windows (with cargo cache).

## Assumptions

- PR CI time is currently painful and slows iteration (at ~40mn total CI time)
- CI time for `main` is less sensitive (though this proposal wouldn't increase it)
- It's unlikely that a specific platform would be broken in a way that debug builds pass but release don't
- In the eventuality that happens, those would be caught by CI on `main`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -343,7 +343,7 @@ jobs:
           CARGO_PROFILE_RELEASE_LTO: false
           CARGO_PROFILE_RELEASE_CODEGEN_UNITS: 256
           CARGO_PROFILE_RELEASE_OPT_LEVEL: 0
-          CARGO_PROFILE_BENCH_INCREMENTAL: true
+          CARGO_PROFILE_RELEASE_INCREMENTAL: true
 
       - name: Build release
         if: |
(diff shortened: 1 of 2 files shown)


AaronO:
After discussing this with the deno core team, some were sceptical but overall were in favour of at least trying this.

Furthermore, @lucacasonato pointed out that windows builds experienced no significant slowdown in test execution because windows debug builds used release builds of v8, so we're now considering using release builds of v8 for all debug builds of deno in CI (possibly by default too as long as there's a path to locally using a full debug build of v8).

I think with this CI tweak and that v8 release change (will happen in a follow up PR), we can consistently bring CI time below 20mn, possibly closer to ~15mn in practice !

ry (approved):
LGTM - let's try it

PR #11452, opened 2021-07-19
fix(extensions/websocket): close only after receiving close frame

crowlKats:
~~Test for this is at https://github.com/web-platform-tests/wpt/pull/29708.~~ Test has landed, and has been pulled in.

PR #11887, opened 2021-08-31
feat(ext/websocket): add header support to WebSocketStream

### LINKED ISSUE #11847: Accept arbitrary request headers for WebSocketStream API
The options bag accepted by `WebSocketStream` should be extended to accept a dictionary/`Headers` of custom HTTP headers. This would enable sending authorization headers and/or cookies.

For context, the web browser security model treats WebSockets like `` tags. This means that the user agent fully decides the headers to send. Thus, **unlike `fetch()`,** the upstream `WebSocket` and `WebSocketStream` specs does not allow for customizing the HTTP request.

The proposal here is that Deno add arbitrary header support out-of-spec. I'm specifically referring to `WebSocketStream` as it has an options bag already and is a more useful API for modern stream-oriented programming.

### Brief example code

```typescript
const ws = new WebSocketStream("https://myserver.com", {
  headers: {
    authorization: `Bearer ${jwt}`,
  },
});
```

### Discussed in https://github.com/denoland/deno/discussions/9891

<div type='discussions-op-text'>

<sup>Originally posted by **scientific-dev** March 25, 2021</sup>
I am working with an socket api where i need to send headers with the socket connections. However in nodejs it allows me to input headers in the second param. But as deno uses the browser like `WebSocket` api. I can't send headers with it. Any way so how can i send it with deno?</div>

Further background: #9890 and https://github.com/whatwg/html/issues/3062#issuecomment-880923866

crowlKats:
Closes #11847


ext/websocket/lib.rs
```diff
@@ -238,6 +244,27 @@ where
     request = request.header("Sec-WebSocket-Protocol", args.protocols);
   }
 
+  if let Some(headers) = args.headers {
+    for (key, value) in headers {
+      let name = HeaderName::from_bytes(&key).unwrap();
+      let v = HeaderValue::from_bytes(&value).unwrap();
```
lucacasonato, line 250: Please handle the error correctly.

ext/websocket/lib.rs
```diff
@@ -238,6 +244,27 @@ where
...
+      name,
+      http::header::HOST
+        | http::header::SEC_WEBSOCKET_ACCEPT
+        | http::header::SEC_WEBSOCKET_EXTENSIONS
+        | http::header::SEC_WEBSOCKET_KEY
+        | http::header::SEC_WEBSOCKET_PROTOCOL
+        | http::header::SEC_WEBSOCKET_VERSION
+        | http::header::UPGRADE
+        | http::header::CONNECTION
+    ) {
```
lucacasonato, line 262: Assign matches! result to variable.

lucacasonato (commented):
Nice. Still needs:
- TS types
- a test

Could you also document this difference in  https://github.com/denoland/manual/blob/main/runtime/web_platform_apis.md once this is landed?

bartlomieju (approved):
LGTM

crowlKats:
Will look into fixing it up and getting it ready tomorrow

ext/websocket/02_websocketstream.js
```diff
@@ -143,6 +147,7 @@
               ? ArrayPrototypeJoin(options.protocols, ", ")
               : "",
             cancelHandle: cancelRid,
+            headers: options.headers,
```
lucacasonato, line 150: You need to do `new Headers()` with the `options.headers` first. Then you should get the header list from the `Headers` and pass that to rust. Currently this code will break when a user passes an object instead of a `[string, string][]`.

crowlKats:
Blocked by #11452

bartlomieju (approved):
LGTM

PR #11888, opened 2021-08-31
feat(workers): Make the `Deno` namespace configurable and unfrozen

andreubotella:
This is the worker counterpart of PR #11062.


lucacasonato (approved):
LGTM, thanks @andreubotella.

PR #8167, opened 2020-10-28
fix(cli): make hashes of tsconfig deterministic 

kitsonk:
Fixes #8163

Issue #8163, opened 2020-10-28
"Permission denied (os error 13)" with read-only Deno cache dir and `--config` flag using Deno 1.5.0


TooTallNate:
### Error

```
$ deno run --config tsconfig.json hello.ts
Check file:///home/hello.ts
error: Permission denied (os error 13) (for '"/home/cache/gen/file/home/hello.ts.js"')
```

### To reproduce

**tsconfig.json**

```json
{
  "compilerOptions": {
    "strict": false,
    "noImplicitAny": false,
    "noImplicitThis": false,
    "alwaysStrict": false,
    "strictNullChecks": false,
    "strictFunctionTypes": true,
    "strictPropertyInitialization": false,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true
  }
}
```

**hello.ts**

```typescript
import { ServerRequest } from 'https://deno.land/std@0.75.0/http/server.ts';

export default async (req: ServerRequest) => {
	req.respond({ body: `Hello, from Deno v${Deno.version.deno}!` });
};
```

Run commands:

```bash
$ deno --version
deno 1.5.0
v8 8.7.220.3
typescript 4.0.3

$ export DENO_DIR=cache

$ deno cache --config tsconfig.json hello.ts
Download https://deno.land/std@0.75.0/http/server.ts
Download https://deno.land/std@0.75.0/encoding/utf8.ts
Download https://deno.land/std@0.75.0/async/mod.ts
Download https://deno.land/std@0.75.0/http/_io.ts
Download https://deno.land/std@0.75.0/io/bufio.ts
Download https://deno.land/std@0.75.0/_util/assert.ts
Download https://deno.land/std@0.75.0/http/http_status.ts
Download https://deno.land/std@0.75.0/textproto/mod.ts
Download https://deno.land/std@0.75.0/bytes/mod.ts
Download https://deno.land/std@0.75.0/async/mux_async_iterator.ts
Download https://deno.land/std@0.75.0/async/delay.ts
Download https://deno.land/std@0.75.0/async/deferred.ts
Download https://deno.land/std@0.75.0/async/pool.ts
Check file:///home/hello.ts

# Before `chmod` - works
$ deno run --config tsconfig.json hello.ts
Check file:///home/hello.ts

# Mark cache dir as read-only (simulating AWS Lambda runtime environment)
$ chmod -R -w cache

# After `chmod` - errors
$ deno run --config tsconfig.json hello.ts
Check file:///home/hello.ts
[...]

PR #11890, opened 2021-09-01
chore(std/http): server module name migration

cmorten:
Relates to https://github.com/denoland/deno_std/discussions/1034 and https://github.com/denoland/deno_std/pull/1128.

For expedience I have made the simplest change of migrating `server.ts` -> `server_legacy.ts`. It would be good to use the new std/http APIs for these tests in future - happy to try rework these, but will take a little longer to do. If we do go with this, we can raise an issue to cover this off at a slower cadence.


cmorten:
I _believe_ this won’t pass CI until the next version of std is released.

cli/tests/testdata/module_graph/file_tests-checkwithconfig.ts
```diff
@@ -1,4 +1,4 @@
-import { ServerRequest } from "https://deno.land/std/http/server.ts";
+import { ServerRequest } from "https://deno.land/std/http/server_legacy.ts";
```
bartlomieju, line 1: Yikes, this file should use `test_util/std/` instead of remote server.

bartlomieju (commented):
@cmorten `test_util/std/` is a git submodule, so we don't have to wait for std release to update it. Let me know if you want to try and do that or should I

cli/tests/testdata/module_graph/file_tests-checkwithconfig.ts
cmorten: That’ll be why then 😄 was confused by the inconsistency 😅

cmorten:
Hmm _I believe_ I've updated the submodule ok @bartlomieju, but looks like there are some errors thrown in the format tests - the result of some changes @kitsonk made prior to the "native" https server work in std for TypeScript 4.4 support https://github.com/denoland/deno_std/pull/1171

Not sure how best to proceed? Happy to raise a PR to try resolve the type issues in std as seen by this repo (but not std's CI)?

bartlomieju:
> Hmm _I believe_ I've updated the submodule ok @bartlomieju (please shout if not!), but looks like there are some errors thrown in the format tests - the result of some changes @kitsonk made prior to the "native" HTTP server work in std for TypeScript 4.4 support [denoland/deno_std#1171](https://github.com/denoland/deno_std/pull/1171).
> 
> Not sure how best to proceed? Happy to raise a PR to try resolve the type issues in std as seen by this repo (but not std's CI)?

Yes, format and lint script use latest deno binary and the TypeScript update is not yet available there. I guess we'll have to wait for 1.14 release before we can update `deno_std` submodule.

cmorten:
> Yes, format and lint script use latest deno binary and the TypeScript update is not yet available there. I guess we'll have to wait for 1.14 release before we can update `deno_std` submodule.

Curiously in CI, the lint steps were fixed to `v1.11.3` which is 2 minor versions behind latest? (last updated in https://github.com/denoland/deno/pull/11369). I've updated for the purpose of this PR - let me know if any reason why couldn't bump to the latest!

Seems this just uncovers the old test issue which thought would be solved by the std submodule update, so will have a debug locally to see what's going on...

cmorten:
Ok think it's sorted - I had misunderstood the `file_tests-checkwithconfig.ts` test's use of

```ts
import { ServerRequest } from "https://deno.land/std/http/server.ts";
```

This _isn't_ going to the std lib, but instead resolves to a local `https_deno.land-std-http-server.ts` mock via importmap. See https://github.com/denoland/deno/pull/8167 and https://github.com/denoland/deno/issues/8163 for details. We _could_ consider renaming the import so to not to cause the same confusion in future, but not sure if that detracts from the test - I suspect it is fine, and actually looking at the blame + test setup quickly pointed to the purpose.

cmorten:
@bartlomieju all issues resolved now 🤞🏻

bartlomieju (approved):
LGTM, thanks for looking into this Craig!

PR #11894, opened 2021-09-01
chore: upgrade crates

bartlomieju:

### DIFF
--- a/cli/tools/doc.rs
+++ b/cli/tools/doc.rs
@@ -58,8 +58,7 @@ impl Loader for DocLoader {
             maybe_headers: None,
             content: Arc::new(file.source),
           })
-        })
-        .map_err(|err| err.into());
+        });
       (specifier.clone(), result)
     }
     .boxed_local()
--- a/cli/tools/doc.rs
+++ b/cli/tools/doc.rs
@@ -5,6 +5,7 @@ use crate::colors;
 use crate::file_fetcher::File;
 use crate::flags::Flags;
 use crate::get_types;
+use crate::import_map::ImportMap;
 use crate::media_type::MediaType;
 use crate::program_state::ProgramState;
 use crate::write_json_to_stdout;
@@ -18,6 +19,7 @@ use deno_graph::create_graph;
 use deno_graph::source::LoadFuture;
 use deno_graph::source::LoadResponse;
 use deno_graph::source::Loader;
+use deno_graph::source::Resolver;
 use deno_graph::ModuleSpecifier;
 use deno_runtime::permissions::Permissions;
 use std::path::PathBuf;
@@ -32,6 +34,30 @@ impl Loader for StubDocLoader {
     _is_dynamic: bool,
   ) -> LoadFuture {
     Box::pin(future::ready((specifier.clone(), Ok(None))))
+  }
+}
+
+#[derive(Debug)]
+struct DocResolver {
+  import_map: Option<ImportMap>,
+}
+
+impl Resolver for DocResolver {
+  fn resolve(
+    &self,
+    specifier: &str,
+    referrer: &ModuleSpecifier,
+  ) -> Result<ModuleSpecifier, AnyError> {
+    if let Some(import_map) = &self.import_map {
+      return import_map
+        .resolve(specifier, referrer.as_str())
+        .map_err(AnyError::from);
+    }
+
+    let module_specifier =
+      deno_core::resolve_import(specifier, referrer.as_str())?;
+
+    Ok(module_specifier)
   }
 }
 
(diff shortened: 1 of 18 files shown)


dsherret (approved):
LGTM

bartlomieju:
There still a few tests failing:
```
    integration::doc::deno_doc_import_map
    integration::doc::deno_doc_types_header
    integration::doc::deno_doc_types_header_direct
    integration::doc::deno_doc_types_hint
    integration::doc::deno_doc_types_ref
```
I need to dig into the `deno_graph` a bit deeper - I believe I did not implement "loader" for the graph properly.

cli/tools/doc.rs
```diff
@@ -6,74 +6,62 @@ use crate::file_fetcher::File;
...
+        .file_fetcher
+        .fetch(&specifier, &mut Permissions::allow_all())
+        .await
+        .map(|file| {
+          Some(LoadResponse {
+            specifier: specifier.clone(),
+            maybe_headers: None,
+            content: Arc::new(file.source),
+          })
+        });
```
bartlomieju, line 61: So this is wrong - firstly it completely skips "headers"; secondly it doesn't apply import map. 

Not sure if I should embed import map logic here or should I use `maybe_resolver` field on the `create_graph`: https://docs.rs/deno_graph/0.3.0/deno_graph/fn.create_graph.html

bartlomieju: Okay, so I added headers handling - I'm forwarding them from `FileFetcher` and `File` struct. However I'm still facing problem with import map (so the question above is still valid).

There are two tests that are "almost passing":
```
---- integration::doc::deno_doc_types_header_direct stdout ----
deno_exe path /Users/biwanczuk/dev/deno/target/debug/deno
test_server starting...
deno_exe args doc --reload http://127.0.0.1:4545/xTypeScriptTypes.js
deno_exe testdata path "/Users/biwanczuk/dev/deno/cli/tests/testdata"
output path /Users/biwanczuk/dev/deno/cli/tests/testdata/doc/types_header.out
OUTPUT
Download http://127.0.0.1:4545/xTypeScriptTypes.js
Download http://127.0.0.1:4545/xTypeScriptTypes.d.ts
Defined in http://127.0.0.1:4545/xTypeScriptTypes.js:1:0

const foo: "foo"


OUTPUT
EXPECTED
Download http://127.0.0.1:4545/xTypeScriptTypes.js
Download http://127.0.0.1:4545/xTypeScriptTypes.d.ts
Defined in http://127.0.0.1:4545/xTypeScriptTypes.d.ts:1:0

const foo: "foo"


EXPECTED
thread 'integration::doc::deno_doc_types_header_direct' panicked at 'pattern match failed', test_util/src/lib.rs:1488:7

---- integration::doc::deno_doc_types_header stdout ----
deno_exe path /Users/biwanczuk/dev/deno/target/debug/deno
deno_exe args doc --reload doc/types_header.ts
deno_exe testdata path "/Users/biwanczuk/dev/deno/cli/tests/testdata"
output path /Users/biwanczuk/dev/deno/cli/tests/testdata/doc/types_header.out
OUTPUT
Download http://127.0.0.1:4545/xTypeScriptTypes.js
Download http://127.0.0.1:4545/xTypeScriptTypes.d.ts
Defined in http://127.0.0.1:4545/xTypeScriptTypes.js:1:0

const foo: "foo"


OUTPUT
EXPECTED
Download http://127.0.0.1:4545/xTypeScriptTypes.js
Download http://127.0.0.1:4545/xTypeScriptTypes.d.ts
Defined in http://127.0.0.1:4545/xTypeScriptTypes.d.ts:1:0

const foo: "foo"


EXPECTED
```

As can be seen the actual output shows that symbols are defined in `.d.ts` files while we expect them to be defined in `.js` files. To be honest I'm not sure what should happen here. CC @kitsonk please advise.

kitsonk: Ok, I have take a look at it.  There is some existing evil hackery in there that should have never have been in there IMO:

```rust
    // If the root module has external types, the module graph won't redirect it,
    // so instead create a dummy file which exports everything from the actual file being documented.
    let root_specifier = resolve_url_or_path("./$deno$doc.ts").unwrap();
```

We should have never done that.  What we need to do in `deno_doc` is to "prefer types" when resolving modules out of the graph and checking for the types definition for a file.  Originally the module graph in CLI didn't model this properly, and we should have fixed it far earlier versus just working around it.

I will try to raise the PR for `deno_doc` now, should be pretty straight forward and hopefully the solution will "just work" downstream.

kitsonk: Ok, I raised https://github.com/denoland/deno_doc/pull/146 which should fix the the problem that was being worked around, so the "magic" doc module should be removed and doc'ed specifier should just be passed, and `deno_graph` and `deno_doc` will properly take care of the rest, and it should cause the right specifiers to be printed out I think.

PR #11896, opened 2021-09-01
refactor: factor out DenoSubcommand enum variant into structs

bartlomieju:
Ref https://github.com/denoland/deno/pull/11776#discussion_r699701642

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -33,96 +33,141 @@ lazy_static::lazy_static! {
 }
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct BundleFlags {
+  pub source_file: String,
+  pub out_file: Option<PathBuf>,
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct CacheFlags {
+  pub files: Vec<String>,
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct CompileFlags {
+  pub source_file: String,
+  pub output: Option<PathBuf>,
+  pub args: Vec<String>,
+  pub target: Option<String>,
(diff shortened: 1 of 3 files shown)


cli/flags.rs
```diff
@@ -32,97 +32,142 @@ lazy_static::lazy_static! {
...
-    output: Option<PathBuf>,
-    ca_file: Option<String>,
-  },
+  Upgrade(UpgradeFlags),
 }
 
 impl Default for DenoSubcommand {
   fn default() -> DenoSubcommand {
-    DenoSubcommand::Repl { eval: None }
+    DenoSubcommand::Repl(ReplFlags { eval: None })
```
caspervonb, line 170: Maybe we should implement default for the new command flag sets as-well?

cli/main.rs
```diff
@@ -1068,14 +1077,14 @@ async fn test_command(
...
       flags,
-      include,
-      ignore,
-      doc,
-      no_run,
-      fail_fast,
-      filter,
-      shuffle,
-      concurrent_jobs,
+      test_flags.include,
```
caspervonb, line 1080: Nice, this is pretty much exactly what I had in mind.
I'll follow up with a cleanup of these signatures.

cli/flags.rs
```diff
@@ -3665,7 +3712,7 @@ mod tests {
     assert_eq!(
       r.unwrap(),
       Flags {
-        subcommand: DenoSubcommand::Test {
+        subcommand: DenoSubcommand::Test(TestFlags {
```
caspervonb, line 3715: Tests could be simpler if we used defaults.

bartlomieju: Good point, but I think this could be done in a follow up

caspervonb: SGTM.

PR #11904, opened 2021-09-02
feat(cli): add before and after test hooks

caspervonb:
The load and unload events are a good point to do setup and teardown but often not enough when you need to do
something asynchronous like say send a http request to delete restful resources.

This adds before and after hooks which are ran before any tests execute and after all tests have executed.

Prior art:
- https://jestjs.io/docs/setup-teardown#repeating-setup-for-many-tests
- https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#before--after-hooks
- https://mochajs.org/#hooks
- https://jasmine.github.io/api/edge/global.html#beforeAll
- https://jasmine.github.io/api/edge/global.html#afterAll

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -133,58 +133,62 @@ declare namespace Deno {
     sanitizeExit?: boolean;
   }
 
-  /** Register a test which will be run when `deno test` is used on the command
-   * line and the containing module looks like a test module.
-   * `fn` can be async if required.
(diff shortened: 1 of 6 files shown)


kt3k:
Is https://github.com/denoland/deno/pull/12190 related?

bartlomieju:
@caspervonb thank you for the PR and effort, but I'm going to close it without a merge as it became quite stale and it's not obvious how we could integrate this functionality with test-steps API.

PR #11905, opened 2021-09-02
chore(test): improve flaky `lsp_diagnostics_refresh_dependents` test to give more info

dsherret:
This test was failing on the CI with `'assertion failed: params.diagnostics.is_empty()'`. This change makes it so that the CI gives us a little more information about what's going on.


kitsonk (approved):
LGTM

Issue #11816, opened 2021-08-23
FileReader onevent attributes don't conform to spec


lucacasonato:
There are six related failing tests in wpt/FileAPI/idlharness.any.html. Likely cause is that the function name of the getters and setters is not correct.

F3n67u:
Access webidl attribute such as `onabort` in FileReader idl via prototype should throw `TypeError`, but deno return `null`.

### Chrome
```
FileReader.prototype.onabort
Uncaught TypeError: Illegal invocation
    at <anonymous>:1:22
```

### deno
```
% deno repl
Deno 1.13.2
exit using ctrl+d or close()
> FileReader.prototype.onabort
null
```

F3n67u:
the spec, see https://heycam.github.io/webidl/#LegacyLenientThis

```
// Throws a TypeError, since Example.prototype is not an Example object.
Example.prototype.y;
```

I'd like to help, but I don't have a idea how to throw TypeError when access `Example.prototype.y`

F3n67u:
I tried to fix this problem, could you help to review and tell me what you think? @lucacasonato

PR #11908, opened 2021-09-03
fix: FileReader onevent attributes don't conform to spec

F3n67u:
fix https://github.com/denoland/deno/issues/11816

### DIFF
--- a/ext/web/10_filereader.js
+++ b/ext/web/10_filereader.js
@@ -472,6 +472,7 @@
   function makeEventHandlerGetter(name) {
     return function () {
       webidl.assertBranded(this, FileReader);
+
       const maybeMap = this[handlerSymbol];
       if (!maybeMap) return null;
 
@@ -480,6 +481,8 @@
   }
   function makeEventHandlerSetter(name) {
     return function (value) {
+      webidl.assertBranded(this, FileReader);
+
       if (!this[handlerSymbol]) {
         this[handlerSymbol] = new Map();
       }
@@ -490,6 +493,7 @@
         handlerWrapper = makeWrappedHandler(value);
         this.addEventListener(name, handlerWrapper);
       }
+
       MapPrototypeSet(this[handlerSymbol], name, handlerWrapper);
     };
   }
--- a/ext/web/10_filereader.js
+++ b/ext/web/10_filereader.js
@@ -41,6 +41,7 @@
   const result = Symbol("[[result]]");
   const error = Symbol("[[error]]");
   const aborted = Symbol("[[aborted]]");
+  const handlerSymbol = Symbol("eventHandlers");
 
   class FileReader extends EventTarget {
     get [SymbolToStringTag]() {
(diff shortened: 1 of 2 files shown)


ext/web/10_filereader.js
```diff
@@ -421,39 +469,30 @@
     wrappedHandler.handler = handler;
     return wrappedHandler;
   }
-  // TODO(benjamingr) reuse when we can reuse code between web crates
-  function defineEventHandler(emitter, name) {
-    // HTML specification section 8.1.5.1
-    ObjectDefineProperty(emitter, `on${name}`, {
-      get() {
```
F3n67u, line 428: ObjectDefineProperty get would fail at https://github.com/web-platform-tests/wpt/blob/51d8c24e16deaf156a7eab496cf06721e55c4ebf/resources/idlharness.js#L2948, so we have to resolve to class setter way.

ext/web/10_filereader.js
```diff
@@ -421,39 +469,30 @@
...
-        const maybeMap = this[handlerSymbol];
-        if (!maybeMap) return null;
-
-        return MapPrototypeGet(maybeMap, name)?.handler ?? null;
-      },
-      set(value) {
```
F3n67u, line 434: same as getter,  ObjectDefineProperty set would fail at https://github.com/web-platform-tests/wpt/blob/51d8c24e16deaf156a7eab496cf06721e55c4ebf/resources/idlharness.js#L2994

ext/web/10_filereader.js
```diff
@@ -368,6 +368,54 @@
       // alias for readAsArrayBuffer
       this.#readOperation(blob, { kind: "Text", encoding });
     }
+
+    get onerror() {
+      return FunctionPrototypeCall(makeEventHandlerGetter("error"), this);
```
lucacasonato, line 373: With this implementation a new getter/setter would be created for each `get` / `set`. Can we prevent that?

ext/web/10_filereader.js
```diff
@@ -421,39 +469,30 @@
...
+      webidl.assertBranded(this, FileReader);
+      const maybeMap = this[handlerSymbol];
+      if (!maybeMap) return null;
+
+      return MapPrototypeGet(maybeMap, name)?.handler ?? null;
+    };
+  }
+  function makeEventHandlerSetter(name) {
+    return function (value) {
+      if (!this[handlerSymbol]) {
```
lucacasonato, line 483: This should also do a brand check.

ext/web/10_filereader.js
F3n67u: haven't came up a solution which can prevent that problem, do you have any suggestion?

ext/web/10_filereader.js
F3n67u: fixed, thanks for review

ext/web/10_filereader.js
bartlomieju: Maybe create those up-front and assign to some private properties on the class?

F3n67u: @bardiarastin  Sorry about that I cannot get what you mean. 

There is 6 pair of `get/set` for 6 event attribute. I thought it is too repeat/wordy too. We must explicitly defined those via class `get/set`, if we defined `get/set` via `Object.definedProperty` wpt test will fail cause is that the function name of the getters and setters is not correct.

BTW. I refactor `makeEventHandlerGetter` and `makeEventHandlerSetter` function to class private method, make event `get/set` function body more concise.

F3n67u: If anybody have a better/DRY solution, feel free to add commit to this pr. I will very happy for your help.

F3n67u: FYI. `jsdom` also implement webidl attribute using `Object.definedProperty` which may also encounter the same `get/set` function name issue with us. https://github.com/jsdom/jsdom/blob/ba8e5806d1a23960a03c3dd99af98fd0916b4bc9/lib/jsdom/living/helpers/create-event-accessor.js#L177

ry (approved):
LGTM - thank you @F3n67u !

PR #11909, opened 2021-09-03
Remove Deno.Signals enum

### LINKED ISSUE #11900: Move away from using enums for Deno.Signal API
Currently we got two huge enums for Signal definition:
https://github.com/denoland/deno/blob/77ead8af20180453aa2d0db67d7856641bd5498f/cli/dts/lib.deno.unstable.d.ts#L567-L637

This is bad, as there are different enums depending on the OS.

After initial discussion between the core team we decided we should just use strings for signals.

This is prerequisite for https://github.com/denoland/deno/issues/11757

ry:
Fixes #11900


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -685,53 +613,53 @@ declare namespace Deno {
...
-  export function signal(signo: number): SignalStream;
+  export function signal(signo: string): SignalStream;
 
   /** **UNSTABLE**: new API, yet to be vetted. */
   export const signals: {
     /** Returns the stream of SIGALRM signals.
      *
-     * This method is the shorthand for `Deno.signal(Deno.Signal.SIGALRM)`. */
+     * This method is the shorthand for `Deno.signal("SIGALRM")`. */
     alarm: () => SignalStream;
```
ry, line 623: So `Deno.signals.alarm()` is shorthand for `Deno.signal("SIGALRM")` ... seems useless. We should remove the entire `Deno.signals` namespace IMO (cc @kt3k)

kitsonk:
Using a string literal union, while it would still allow some unsupported signals, would greatly improve usability in editors. For example:

```ts
export type Signal = "SIGHUP" | "SIGINT";

export function kill(signal: Signal): void;
```

cli/dts/lib.deno.unstable.d.ts
bartlomieju: I'm in favor - it's just syntactic sugar

ry: removed

runtime/ops/signal.rs
```diff
@@ -53,15 +54,102 @@ impl Resource for SignalStreamResource {
...
+    "SIGUSR1" => Some(30),
+    "SIGUSR2" => Some(31),
+    _ => None,
+  }
+}
+
+#[cfg(target_os = "windows")]
+fn signal_str_to_int(s: &str) -> Option<libc::c_int> {
+  unimplemented!()
+}
```
bartlomieju, line 136: This will panic on Windows, can we maybe handle it gracefully and return a regular error to JS?

ry: graceful error handling on windows is done earlier 
https://github.com/ry/deno/blob/0dca16759132bf8c1acbd225751d417493a873ae/runtime/js/40_signals.js#L28-L33

bartlomieju: Ah perfect

caspervonb:
Wonder if these should be case insensitive ala dom events?

ry:
@caspervonb I think it would just be more work - given the Signal string literal union, I think it will be clear that upper case is expected.

caspervonb:
> @caspervonb I think it would just be more work - given the Signal string literal union, I think it will be clear that upper case is expected.

Would only be `s.to_uppercase()` call in the match but don't care that much, was just a passing thought.

bartlomieju (approved):
LGTM from me

cli/dts/lib.deno.ns.d.ts
```diff
@@ -1982,13 +1982,12 @@ declare namespace Deno {
...
 
-    /** **UNSTABLE**: The `signo` argument may change to require the Deno.Signal
-     * enum.
+    /** **UNSTABLE**
      *
      * Send a signal to process. This functionality currently only works on
      * Linux and Mac OS.
      */
-    kill(signo: number): void;
+    kill(signo: string): void;
```
kt3k, line 1990: This also should be `Signal` type?

ry: I'm not sure how to do this because Signal type is unstable API... I think we should just allow less precise types here until this is made stable. I will add a TODO.

caspervonb: On windows, wouldn't this just map to the status code? so should probably string or integer?

kt3k (approved):
LGTM

cli/dts/lib.deno.ns.d.ts
ry: @caspervonb Hmm that's a good question. @piscisaureus do you have any recommendations?

piscisaureus: > @caspervonb Hmm that's a good question. @piscisaureus do you have any recommendations?

* It's unusual for signal numbers to be mapped directly to exit codes on Windows.
* We want Deno to do something reasonable on all platforms even if they're not writing platform specific code. `p.kill("SIGKILL")` should work on Windows, and as is customary, it should exit the process with exit code 1.

PR #11910, opened 2021-09-03
feat(cli): Support Basic authentication in DENO_AUTH_TOKENS

### LINKED ISSUE #11899: Private Modules: Add Support for Basic Auth in DENO_AUTH_TOKENS
Hello Deno Team 👋

i would like to use private modules but sadly it only support Bearer Tokens
but i need Basic Auth (username + password) for BitBucket Cloud (as personal token got replace with app passwords)

i could see the format change to `{token|username:password}@{hostname[:port]}`
and because of the colon it could switch from Bearer token to Basic auth

There was this issue before 
https://github.com/denoland/deno/issues/5897 but we sould extend the api not create something new

BasiqueEvangelist:
Closes #11899 

I'm not sure on the naming of the `AuthTokenData` enum. Normally I would use the name `AuthToken`, but the name is already taken and I don't want to rename existing types.


lucsoft:
I can approve this solved my issues 💯

bartlomieju (commented):
The code looks good for me and the implementation is sensible. @BasiqueEvangelist could you add some integration test that checks it works as expected in end-to-end scenario?

Before landing, I'll wait for @kitsonk's review

lucacasonato (commented):
Note: this will break all Bearer tokens that contain `:`.

lucsoft:
> Note: this will break all Bearer tokens that contain `:`.

I think this is a non issue because using `:` is not really spec conform and most often Bearer tokens use JWT or some base64 style i think even the regex is just `[-a-zA-Z0-9._~+/]+=*`

lucacasonato:
@lucsoft The original Oauth2 specification allows any printable ASCII character inside a Bearer token. In practice any character is allowed by most implementations (including invalid characters).

I don't think this is a blocker for this PR, I just want to make sure we are all aware.

BasiqueEvangelist:
@bartlomieju I've added an integration test.

bartlomieju:
@BasiqueEvangelist thank you! I will review and land once CI goes green. Would you mind adding some info about this new feature to manual as per https://github.com/denoland/manual/issues/77?

bartlomieju (commented):
Thanks for adding an integration test, this looks good to me know. @kitsonk are you happy with it?

kitsonk (approved):
LGTM

Issue #10099, opened 2021-04-09
proposal: Better logging on Deno.test


Soremwar:
Currently running `console.log` inside tests has a little inconvenience
```ts
Deno.test("Test logging", () => {
  console.log("Inside the test 1");
  console.log("Inside the test 2");
});
```
```log
running 1 tests
test Test logging ... Inside the test 1
Inside the test 2
ok 
```

I propose that the test line is readded after the test has finished if anything was logged to console in the execution of the test. That way we would get the following output
```log
running 1 tests
test Test logging ...
Inside the test 1
Inside the test 2
test Test logging ... ok 
```

lucacasonato:
So the first `console.log` inside a test would add a newline prefix? Not too sure about this. I would rather the test runner intercept `console.*` fully then and hide the output by default, unless the test fails, just like `cargo` does in rust.

Soremwar:
That works as well. The output will be shown in the stack trace of the failed test then?

caspervonb:
> So the first console.log inside a test would add a newline prefix? Not too sure about this. I would rather the test runner intercept console.* fully then and hide the output by default, unless the test fails, just like cargo does in rust.

Yeah been saying this in chat, except I want us to capture stdout (and probably stderr) to a memory buffer for the test isolate rather than changing the console object.

getspooky:
I think we should just format the output of the console for example 
```ts
running 1 tests
test Test logging ...
Inside the test 1
Inside the test 2
test Test logging ... ok 
```
instead of 
```ts
running 1 tests
test Test logging ... Inside the test 1
Inside the test 2
ok 
```
and it would be a good idea to add `--silent=false` so deno will  suppresse the console log message.

getspooky:
@Soremwar you use `--quiet` flag to ignore log

caspervonb:
Tag this for 2.0 please.

bartlomieju:
Done in https://github.com/denoland/deno/pull/14271, https://github.com/denoland/deno/pull/14395, and https://github.com/denoland/deno/pull/14410

PR #11911, opened 2021-09-03
refactor(testing): redirect console output via reporter

caspervonb:
This feeds console output to the reporter and handles silencing/echoing there instead of in the JavaScript code making all the output to stdout in a test come through from the reporter.

Stdout and stderr should also be piped through here but that would be a breaking change for later, possibly 2.0.

Towards https://github.com/denoland/deno/issues/10099


cli/tests/unit/console_test.ts
```diff
@@ -313,27 +313,27 @@ unitTest(function consoleTestStringifyCircular() {
     "JSON {}",
   );
   assertEquals(
-    stringify(console),
+    stringify(new Console(() => {})),
```
caspervonb, line 316: Test case is faulty on main, run with `deno test --quiet and it fails because it is not bound`.
Doesn't matter in this case what the fixture is, stringify is what is being tested.

runtime/js/40_testing.js
```diff
@@ -228,15 +228,19 @@ finishing test case.`;
...
-      globalThis.console = new Console(() => {});
-    }
+
+    globalThis.console = new Console((line) => {
+      dispatchTestEvent({
+        output: {
+          console: line,
+        },
+      });
+    });
```
bartlomieju, line 243: Interesting

cli/tests/unit/console_test.ts
bartlomieju: Yes I added it with `wrapConsole` API. Good catch and thanks for fixing this.

cli/tools/test.rs
```diff
@@ -248,7 +272,6 @@ async fn test_specifier(
   test_source.push_str(&format!(
     "await Deno[Deno.internal].runTests({});\n",
     json!({
-      "disableLog": program_state.flags.log_level == Some(Level::Error),
```
bartlomieju, line 251: Nice! The less data needed to forward to JS the better

cli/tools/test.rs
```diff
@@ -600,7 +624,9 @@ async fn test_specifiers(
     .buffer_unordered(concurrent_jobs.get())
     .collect::<Vec<Result<Result<(), AnyError>, tokio::task::JoinError>>>();
 
-  let mut reporter = create_reporter(concurrent_jobs.get() > 1);
+  let mut reporter =
+    create_reporter(concurrent_jobs.get() > 1, log_level != Some(Level::Error));
```
bartlomieju, line 628: `log_level != Some(Level::Error)` is true when `--quiet` flag is passed?

caspervonb: It's inverted here, the flag is echo rather than mute.
Quiet sets log level to error (https://github.com/denoland/deno/blob/main/cli/flags.rs#L400)

bartlomieju (approved):
LGTM, thanks

PR #11913, opened 2021-09-04
Don't drop messages from workers that have already been closed

### LINKED ISSUE #11526: Messages from workers that are about to close are spuriously not received
asanchez reported on Discord that the following code was spuriously getting fewer responses than expected:

### `main.ts`
```ts
for(let idx = 0; idx < 4; idx++){
  const worker = new Worker(new URL('./worker.ts', import.meta.url).href, {
    type: "module", 
    deno: { namespace: true},
  });

  worker.postMessage({ id: idx, name: 'text1', email: 'ea@da.com' });
  worker.addEventListener('message', message => {
    console.log('response', message.data)
  })
}
```

### `worker.ts`
```ts
self.onmessage = async (params) => {
  const data = params.data
  console.log('params', data)
  data.name = 'text2'
  self.postMessage(data)
  self.close()
}
```

### Sample output
```
params { id: 0, name: "text1", email: "ea@da.com" }
params { id: 1, name: "text1", email: "ea@da.com" }
params { id: 2, name: "text1", email: "ea@da.com" }
params { id: 3, name: "text1", email: "ea@da.com" }
response { id: 3, name: "text2", email: "ea@da.com" }
```

----

Adding a call to `Deno.sleepSync(1000)` between `self.postMessage()` and `self.close()` seems to fix it, so it looks like it's a race condition.

Might be related to #11416.

andreubotella:
When `worker.terminate()` is called, the spec requires that the corresponding port message queue is emptied, so no messages can be received after the call, even if they were sent from the worker before it was terminated.

The spec doesn't require this of `self.close()`, and since Deno uses different channels to send messages and to notify that the worker was closed, messages might still arrive after the worker is known to be closed, which are currently being dropped. This change fixes that.

The fix involves two parts: one on the JS side and one on the Rust side. The JS side was using the `#terminated` flag to keep track of whether the worker is known to be closed, without distinguishing whether further messages should be dropped or not. This PR changes that flag to an enum `#state`, which can be one of `"RUNNING"`, `"CLOSED"` or `"TERMINATED"`.

The Rust side was removing the `WorkerThread` struct from the workers table when a close control was received, regardless of whether there were any messages left to read, which made any subsequent calls to `op_host_recv_message` to return `Ok(None)`, as if there were no more mesasges. This change instead waits for both a close control and for the message channel's sender to be closed before the worker thread is removed from the table.

Fixes #11526.

### DIFF
--- a/runtime/ops/worker_host.rs
+++ b/runtime/ops/worker_host.rs
@@ -67,6 +67,12 @@ pub struct CreateWebWorkerCbHolder(Arc<CreateWebWorkerCb>);
 pub struct WorkerThread {
   join_handle: JoinHandle<Result<(), AnyError>>,
   worker_handle: WebWorkerHandle,
+
+  // A WorkerThread that hasn't been explicitly terminated can only be removed
+  // from the WorkersTable once close messages have been received for both the
+  // control and message channels. See `close_channel`.
+  ctrl_closed: bool,
+  message_closed: bool,
 }
 
 pub type WorkersTable = HashMap<WorkerId, WorkerThread>;
@@ -553,6 +559,8 @@ fn op_create_worker(
   let worker_thread = WorkerThread {
     join_handle,
     worker_handle: worker_handle.into(),
+    ctrl_closed: false,
+    message_closed: false,
   };
 
   // At this point all interactions with worker happen using thread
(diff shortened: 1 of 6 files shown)


runtime/ops/worker_host.rs
```diff
@@ -582,19 +590,49 @@ fn op_host_terminate_worker(
...
-/// table - in that case ignore.
-fn try_remove_and_close(state: Rc<RefCell<OpState>>, id: WorkerId) {
+enum WorkerChannel {
+  Ctrl,
+  Messages,
+}
+
+/// Close a worker's channel. If this results in both of a worker's channels
+/// being closed, the worker will be removed from the workers table.
+fn close_channel(
```
andreubotella, line 600: The name of this function (and the two added fields to `WorkerThread`) needs bikeshedding, since it doesn't really close the underlying channel – it just marks it as closed, and cleans up the worker if both of its channels are closed.

lucacasonato (approved):
LGTM. I confirmed the test fails on main.

PR #11924, opened 2021-09-05
ci: try rust-cache

bartlomieju:
Prompted by https://matklad.github.io/2021/09/04/fast-rust-builds.html


.github/workflows/ci.yml
```diff
@@ -246,53 +246,9 @@ jobs:
...
-            g-cargo-target-${{ matrix.os }}-${{ matrix.profile }}-
-
-      # Don't save cache after building PRs or branches other than 'main'.
-      - name: Skip save cache (PR)
-        run: echo "CACHE_SKIP_SAVE=true" >> $GITHUB_ENV
-        shell: bash
-        if: github.ref != 'refs/heads/main'
-
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v1
```
dsherret, line 250: I'm not sure why, but when it goes to post the cache it says:

```
Post job cleanup.
Warning:  The "path" argument must be of type string. Received type undefined
```

bartlomieju: Not sure either, but it seems to work. I'll push a few more empty commit here to see how it works.

dsherret (approved):

bartlomieju:
@piscisaureus please take a look

AaronO:
@bartlomieju What's the expected delta ?

bartlomieju:
> @bartlomieju What's the expected delta ?

I don't have concrete expectation in mind; if the time is similar to current ones I'd say it's already a small win, as the caching is one-line with this solution. We can iterate further using suggestions from the blog post. I already found a few crates that take a lot of time to compile:

132138315-268210ab-8846-4c81-80fe-7c4d35b05db4

piscisaureus:
Let's discuss on our Thursday call.

bartlomieju:
Discussed with @piscisaureus, this PR doesn't help.

PR #11931, opened 2021-09-06
fix permission prompt stuffing

### LINKED ISSUE #9750: Runtime permissions request should fully consume STDIN before asking for user input

ry:
Fixes  #9750


runtime/permissions.rs
```diff
@@ -1194,6 +1194,29 @@ fn permission_prompt(message: &str) -> bool {
...
+  fn clear_stdin() {
+    unsafe {
+      let stdin = winapi::um::processenv::GetStdHandle(
+        winapi::um::winbase::STD_INPUT_HANDLE,
+      );
+      let flags =
+        winapi::um::winbase::PURGE_TXCLEAR | winapi::um::winbase::PURGE_RXCLEAR;
+      winapi::um::commapi::PurgeComm(stdin, flags);
+    }
+  }
```
bartlomieju, line 1214: I'm wondering if https://docs.microsoft.com/en-us/windows/console/flushconsoleinputbuffer might be a better choice

bartlomieju (approved):
LGTM

PR #11934, opened 2021-09-06
refactor(lint): reuse lint rules

bartlomieju:
This commit updated "deno_lint" crate to 0.15.1 and refactors
"cli/tools/lint.rs" to create only a single vector of lint rules,
instead of creating a vector for each linted file.

<s>Blocked by #11780</s>

### DIFF
--- a/cli/tools/lint.rs
+++ b/cli/tools/lint.rs
@@ -152,7 +152,7 @@ pub async fn lint_files(
   Ok(())
 }
 
-fn rule_to_json(rule: Box<dyn LintRule>) -> serde_json::Value {
+fn rule_to_json(rule: &Box<dyn LintRule>) -> serde_json::Value {
   serde_json::json!({
     "code": rule.code(),
     "tags": rule.tags(),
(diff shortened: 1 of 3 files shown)


bartlomieju:
CC @magurotuna

cli/tools/lint.rs
```diff
@@ -171,7 +161,7 @@ fn rule_to_json(rule: Box<dyn LintRule>) -> serde_json::Value {
 }
 
 pub fn print_rules_list(json: bool) {
-  let lint_rules = rules::get_recommended_rules();
+  let lint_rules = Arc::try_unwrap(rules::get_recommended_rules()).unwrap();
```
dsherret, line 164: Why is this unwrap done?

dsherret (approved):
LGTM. Just a question about the unwrap.

cli/tools/lint.rs
```diff
@@ -182,15 +172,18 @@ pub fn print_rules_list(json: bool) {
     // The rules should still be printed even if `--quiet` option is enabled,
     // so use `println!` here instead of `info!`.
     println!("Available rules:");
-    for rule in lint_rules {
+    for rule in lint_rules.into_iter() {
```
magurotuna, line 175: I think if we use `iter()` instead of `into_iter()` here, we won't need to `unwrap` on the line 164.

cli/tools/lint.rs
bartlomieju: Removed.

cli/tools/lint.rs
bartlomieju: Thanks, had to do some more changes, but removed unwrap.

magurotuna (approved):
LGTM so long as CI passes 👍

PR #11941, opened 2021-09-07
feat: add URLPattern API

### LINKED ISSUE #11619: [Suggestion] Add URLPattern global
This is super cutting edge (Chrome Beta only atm), but effectively there's a new web global called `URLPattern` that can be used to test whether a given url string meets a certain pattern. Some examples are available [here](https://web.dev/urlpattern/), and [here](https://github.com/WICG/urlpattern/blob/main/mdn-drafts/QUICK-REFERENCE.md).

As for why I think this global should be added to Deno, essentially this global isn't actually tied to any browser behaviour in any way, and is essentially just a more powerful regular expression so that URLs can be parsed more efficiently.

On the Deno side a couple typical use cases would be things like routing patterns for servers, or for making patterns for globbing the filesystem, etc. Having a standardized syntax for these kinds of patterns would also enable libraries to support such patterns meaning developers wouldn't need to be as concerned with pattern syntax differences from library to library.

lucacasonato:
This adds support for the URLPattern  API.

The API is added in --unstable only, as it has not yet shipped in any
browser. It is targeted for shipping in Chrome 95.

Spec: https://wicg.github.io/urlpattern/

Closes #11619

---

TODOs:
- [x] Add JSDoc comments to lib.deno_url.d.ts
- [x] Add a TypeScript unit test to validate typings.


tools/wpt/expectation.json
```diff
@@ -14524,5 +14524,63 @@
       "performance.clearResourceTimings in workers",
       "performance.setResourceTimingBufferSize in workers"
     ]
+  },
+  "urlpattern": {
+    "urlpattern.any.html": [
+      "Pattern: [{\"pathname\":\"/foo/bar\"}] Inputs: [{\"pathname\":\"foo/bar\"}]",
```
lucacasonato, line 14530: All failing tests are attributable to an upstream spec bug: https://github.com/denoland/rust-urlpattern/issues/8

ext/url/01_urlpattern.js
```diff
@@ -0,0 +1,250 @@
...
+  /**
+   * @typedef Component
+   * @property {string} patternString
+   * @property {RegExp} regexp
+   * @property {string[]} groupNameList
+   */
+
+  class URLPattern {
+    /** @type {Components} */
+    [_components];
```
F3n67u, line 49: wonder why not use js private field?

lucacasonato: Because private fields can not be inspected internally. See for example how the `URLSearchParams` interface needs to access the internal parameters of `URL`.

All our internal code uses symbols for internal properties.

ext/url/urlpattern.rs
```diff
@@ -0,0 +1,40 @@
...
+use urlpattern::quirks::MatchInput;
+use urlpattern::quirks::StringOrInit;
+use urlpattern::quirks::UrlPattern;
+
+pub fn op_urlpattern_parse(
+  _state: &mut deno_core::OpState,
+  input: StringOrInit,
+  base_url: Option<String>,
+) -> Result<UrlPattern, AnyError> {
+  let init = urlpattern::quirks::process_construct_pattern_input(
```
crowlKats, line 14: missing unstable check

lucacasonato: ext/url isn't aware of unstable, and given this API is compute only, I think we can just expose the ops regardless of stability. (i.e. I am lazy and don't want to do the plumbing :smile:)

lucacasonato:
The final failing test is https://github.com/denoland/rust-urlpattern/issues/12. We are waiting on a spec decision.

ext/url/README.md
```diff
@@ -1,5 +1,6 @@
 # deno_url
 
-This crate implements the URL API for Deno.
+This crate implements the URL, and URLPatterm APIs for Deno.
```
bartlomieju, line 3:
```suggestion
This crate implements the URL, and URLPattern APIs for Deno.
```

bartlomieju (approved):
LGTM, minus one typo. I don't see this PR doing any hackery with `serde_v8` you mentioned; has this been worked around in some way?

crowlKats:
> LGTM, minus one typo. I don't see this PR doing any hackery with `serde_v8` you mentioned; has this been worked around in some way?

Yes, this release wont address this, but we can work around it using the `regress` crate instead of the normal `regex` one

bartlomieju:
> > LGTM, minus one typo. I don't see this PR doing any hackery with `serde_v8` you mentioned; has this been worked around in some way?
> 
> Yes, this isnt addressed yet, but we can work around it using the `regress` crate instead of the normal `regex` one

Sounds good, thanks for explanation.

vwkd:
@lucacasonato Should this be documented as "unstable" as well? Currently, it isn't marked on [doc.deno.land](https://doc.deno.land/builtin/stable#URLPattern).

PR #11944, opened 2021-09-07
feat(fmt): add support for configuration file

### LINKED ISSUE #11857: First pass at "deno fmt" config
Ref https://github.com/denoland/deno/issues/11686
Ref https://github.com/denoland/deno/discussions/11855

Analogous to adding configuration for `deno lint` I propose to allow configuring 
`deno fmt` via a configuration file (and also using CLI flags as we don't want to have
options in config file that are not available via CLI). I propose to start
with minimal set of options; concretely allow only `useTabs` (that will help
with the accessibility #3695), `tabWidth` and `singleQuote`

```
{
  "compilerOptions": {
    "jsxFactory": "h"
  },

  "lint": {
    "files": {
      "include": ["testdata/"],
      "exclude": []
    },
    "rules": {
      "tags": ["recommended"],
      "include": [
        "ban-untagged-todo"
      ]
    }
  },

  "fmt": {
    "files": {
      "include": ["testdata/"],
      "exclude": []
    },
    "rules": {
      "useTabs": true,
      "indentWidth": 2,
      "singleQuote": true,
    }
  }
}
```

To keep the same configurability in CLI I suggest to add `--rules-use-tabs`, `--rules-indent-width` and `--rules-single-quote` flags to `deno fmt` subcommand.

Similar to linter specifying CLI flags takes precedence over configuration in config file.

Alternatives:
- allow to configure only `fmt.files` and don't allow configuring rules,
this would imply that we should switch default setting to use tabs instead
of spaces to address #3695

bartlomieju:
Example config with all supported options:
```jsonc
{
    "fmt": {
        "files": {
            "include": ["src/"],
            "exclude": ["src/testdata/"]
        },
        "options": {
            "useTabs": true,
            "lineWidth": 80,
            "indentWidth": 4,
            "singleQuote": true,
            "textWrap": "preserve"
        }
    }
}
```

Closes #11857


cli/config_file.rs
```diff
@@ -286,14 +286,40 @@ pub struct LintFilesConfig {
...
+}
+
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default, deny_unknown_fields, rename_all = "camelCase")]
+pub struct FmtOptionsConfig {
+  pub use_tabs: Option<bool>,
+  pub line_width: Option<u32>,
+  pub indent_width: Option<u8>,
+  pub single_quote: Option<bool>,
+  pub text_wrap: Option<TextWrap>,
```
dsherret, line 307: The reason this is called `text_wrap` in dprint-plugin-markdown is because it controls how the `Text` node wraps. Since this is a top level config and not markdown specific, I think we should call this `prose_wrap` to align with prettier.

bartlomieju:
This PR is still missing tests and I will add them tomorrow. @kitsonk can you please review the changes to LSP?

cli/config_file.rs
bartlomieju: I'm fine with changing name, but I'm unsure about that it's not Markdown specific - I haven't found a setting in `dprint-plugin-typescript` that would be relevant for this setting. Can you elaborate?

cli/tests/integration/lsp_tests.rs
```diff
@@ -3170,6 +3170,77 @@ fn lsp_format_markdown() {
...
+
+  let (maybe_res, maybe_err) = client
+    .write_request::<_, _, Value>(
+      "textDocument/formatting",
+      json!({
+        "textDocument": {
+          "uri": "file:///a/file.ts"
+        },
+        "options": {}
+      }),
```
bartlomieju, line 3216: I have trouble with this test, I'm getting following error:
```
running 1 test
test integration::lsp::lsp_format_with_config ... FAILED

failures:

---- integration::lsp::lsp_format_with_config stdout ----
Some(
    LspResponseError {
        code: -32602,
        message: "missing field `tabSize`",
        data: None,
    },
)
thread 'integration::lsp::lsp_format_with_config' panicked at 'assertion failed: maybe_err.is_none()', cli/tests/integration/lsp_tests.rs:3221:3

```

Should I just pass fake data here and count on the fact that config file should take precedence over these options?

CC @kitsonk

dsherret: I think that's fine to provide some fake data. The editor will have a "tab size" for a file that might not be synced with the formatting settings.

dsherret (approved):
LGTM

PR #11946, opened 2021-09-07
A `Request` whose URL is a revoked blob URL should still fetch

### LINKED ISSUE #11817: Revoking a blob url after creating a `Request` should still fetch successfully
Failing WPT is "Revoke blob URL after creating Request, will fetch" from wpt/FileAPI/url/url-with-fetch.any.html.

andreubotella:
In the spec, a URL record has an associated "blob URL entry", which for `blob:` URLs is populated during parsing to contain a reference to the `Blob` object that backs that object URL. It is this blob URL entry that the `fetch` API uses to resolve an object URL.

Therefore, since the `Request` constructor parses URL inputs, it will have an associated blob URL entry which will be used when fetching, even if the object URL has been revoked since the construction of the `Request` object. (The `Request` constructor takes the URL as a string and parses it, so the object URL must be live at the time it is called.)

This PR adds a new `blobFromObjectUrl` JS function (backed by a new `op_blob_from_object_url` op) that, if the URL is a valid object URL, returns a new `Blob` object whose parts are references to the same Rust `BlobPart`s used by the original `Blob` object. It uses this function to add a new `blobUrlEntry` field to inner requests, which will be `null` or such a `Blob`, and then uses `Blob.prototype.stream()` as the response's body. As a result of this, the `blob:` URL resolution from `op_fetch` is now useless, and has been removed.

Fixes #11817.

### DIFF
--- a/ext/fetch/23_request.js
+++ b/ext/fetch/23_request.js
@@ -19,6 +19,7 @@
   const { mixinBody, extractBody } = window.__bootstrap.fetchBody;
   const { getLocationHref } = window.__bootstrap.location;
   const mimesniff = window.__bootstrap.mimesniff;
+  const { blobFromObjectUrl } = window.__bootstrap.file;
   const {
     headersFromHeaderList,
     headerListFromHeaders,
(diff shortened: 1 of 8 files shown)


andreubotella:
Closing because I mistakenly commited and pushed on my fork's `main` branch.

andreubotella:
#11947

PR #11948, opened 2021-09-07
fix: remove windows-only panic when calling `Deno.kill`

### LINKED ISSUE #11939: Windows signal stubs are panicking
The windows signal stubs are calling the `unimplemented!` macro, which is a panic.
These should be returning a serialisable error instead.

See https://github.com/denoland/deno/issues/11937

dsherret:
Further fixes #11939

### DIFF
--- a/cli/tests/unit/process_test.ts
+++ b/cli/tests/unit/process_test.ts
@@ -485,7 +485,8 @@ unitTest(
         const status = await p.status();
 
         assertEquals(status.success, false);
-        assertEquals(status.signal, "SIGINT");
+        assertEquals(status.code, 130);
+        assertEquals(status.signal, 2);
       }
     } finally {
       p.close();
--- a/cli/tests/unit/process_test.ts
+++ b/cli/tests/unit/process_test.ts
@@ -448,11 +448,9 @@ unitTest(
 
     assert(
       error instanceof Deno.errors.NotFound ||
-        // On Windows, the underlying Windows API may return
-        // `ERROR_ACCESS_DENIED` when the process has exited, but hasn't been
-        // completely cleaned up yet and its `pid` is still valid.
+        // This is not yet implemented on Windows
         (Deno.build.os === "windows" &&
-          error instanceof Deno.errors.PermissionDenied),
+          error instanceof Error && error.message === "not implemented"),
     );
 
     p.close();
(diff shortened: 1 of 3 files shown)


runtime/ops/signal.rs
```diff
@@ -129,16 +129,16 @@ fn signal_str_to_int(s: &str) -> Option<libc::c_int> {
...
 #[cfg(target_os = "windows")]
-fn signal_str_to_int(_s: &str) -> Option<libc::c_int> {
-  unimplemented!()
+fn signal_str_to_int(_s: &str) -> Result<Option<libc::c_int>, AnyError> {
+  Err(generic_error("not implemented"))
 }
 
 pub fn signal_str_to_int_unwrap(s: &str) -> Result<libc::c_int, AnyError> {
-  signal_str_to_int(s)
+  signal_str_to_int(s)?
```
piscisaureus, line 141: After this change `signal_str_to_int_unwrap()` isn't really an appropriate function name any more.
Also, this is more complicated that it needs to be -- why not return `Result<c_int, AnyError>` from the various `signal_str_to_int()` implementations.

cli/tests/unit/process_test.ts
```diff
@@ -476,19 +474,27 @@ unitTest(
...
-      // TODO(nayeemrmn): On Windows sometimes the following values are given
-      // instead. Investigate and remove this catch when fixed.
-      assertEquals(status.code, 130);
-      assertEquals(status.signal, 2);
+    if (Deno.build.os === "windows") {
+      // currently not implemented
+      assertThrows(() => {
+        Deno.kill(p.pid, "SIGINT");
+      }, Error);
+      p.close();
```
piscisaureus, line 482: I'd suggest to move `p.close()` back to the end of the function, rather than duplicating it in the `if{}` and `else{}` blocks.

cli/tests/unit/process_test.ts
```diff
@@ -476,19 +474,27 @@ unitTest(
...
+    } else {
+      Deno.kill(p.pid, "SIGINT");
+      const status = await p.status();
+
+      assertEquals(status.success, false);
+      try {
+        assertEquals(status.signal, "SIGINT");
+      } catch {
+        // TODO(nayeemrmn): On Windows sometimes the following values are given
```
piscisaureus, line 491: This comment applies to Windows but this code doesn't run on windows.

runtime/ops/signal.rs
dsherret: Good point! I've updated the PR with all your feedback.

piscisaureus (approved):
LGTM!

PR #11942, opened 2021-09-07
fix: exit process on panic in a tokio task

dsherret:
Tokio does not exit the process when a task panics, so this change exits the process when any panic occurs.

Closes #11567
Closes #11937

Edit: As shown in the CI, we now get the desired behaviour and the CI fails.

```
hello
Error: thread '<unnamed>' panicked at 'not implemented', runtime\ops\signal.rs:135:3
stack backtrace:
   0:     0x7ff6605cf8d4 - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ff6605cf8d4 - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff6605cf8d4 - std::sys_common::backtrace::_print_fmt
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\sys_common\backtrace.rs:67
   3:     0x7ff6605cf8d4 - std::sys_common::backtrace::_print::{{impl}}::fmt
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\sys_common\backtrace.rs:46
   4:     0x7ff65fd13b8a - core::fmt::write
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\core\src\fmt\mod.rs:1110
   5:     0x7ff6605cdb58 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\io\mod.rs:1588
   6:     0x7ff6605ce9a2 - std::sys_common::backtrace::_print
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\sys_common\backtrace.rs:49
   7:     0x7ff6605ce9a2 - std::sys_common::backtrace::print
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\sys_common\backtrace.rs:36
   8:     0x7ff6605ce9a2 - std::panicking::default_hook::{{closure}}
[...]

Issue #11937, opened 2021-09-07
Signal test failing on Windows but not causing test failure


ry:
This can been seen in all current builds for example https://github.com/denoland/deno/runs/3528022235
```
2021-09-06T21:37:26.8941171Z test runStdinPiped ... [0m[32mok[0m [0m[38;5;8m(44ms)[0m
2021-09-06T21:37:26.9360119Z test runStdoutPiped ... [0m[32mok[0m [0m[38;5;8m(42ms)[0m
2021-09-06T21:37:26.9770009Z test runStderrPiped ... [0m[32mok[0m [0m[38;5;8m(41ms)[0m
2021-09-06T21:37:27.0159714Z test runOutput ... [0m[32mok[0m [0m[38;5;8m(39ms)[0m
2021-09-06T21:37:27.0564044Z test runStderrOutput ... [0m[32mok[0m [0m[38;5;8m(40ms)[0m
2021-09-06T21:37:27.0799992Z test integration::lsp::lsp_hover_change_mbc ... ok
2021-09-06T21:37:27.1128373Z test runRedirectStdoutStderr ... [0m[32mok[0m [0m[38;5;8m(54ms)[0m
2021-09-06T21:37:27.1597300Z test runRedirectStdin ... [0m[32mok[0m [0m[38;5;8m(48ms)[0m
2021-09-06T21:37:27.2034679Z test runEnv ... [0m[32mok[0m [0m[38;5;8m(44ms)[0m
2021-09-06T21:37:27.2110712Z test runClose ... [0m[32mok[0m [0m[38;5;8m(8ms)[0m
2021-09-06T21:37:27.2428136Z hello
2021-09-06T21:37:27.2524548Z ##[error]thread '<unnamed>' panicked at 'not implemented', runtime\ops\signal.rs:135:3
2021-09-06T21:37:27.2541946Z stack backtrace:
2021-09-06T21:37:27.3119199Z    0:     0x7ff6b1f2ef24 - std::backtrace_rs::backtrace::dbghelp::trace
2021-09-06T21:37:27.3121211Z                                at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
2021-09-06T21:37:27.3122771Z    1:     0x7ff6b1f2ef24 - std::backtrace_rs::backtrace::trace_unsynchronized
[...]

PR #11953, opened 2021-09-07
fix(test): propagate join errors in deno test

dsherret:
Cause of #11937, which is already fixed by #11942 so this is just future proofing.

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -715,28 +715,12 @@ async fn test_specifiers(
 
   let (join_results, result) = future::join(join_stream, handler).await;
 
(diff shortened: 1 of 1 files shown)


cli/tools/test.rs
```diff
@@ -715,28 +715,12 @@ async fn test_specifiers(
 
   let (join_results, result) = future::join(join_stream, handler).await;
 
-  let mut join_errors = join_results.into_iter().filter_map(|join_result| {
-    join_result
-      .ok()
```
dsherret, line 720: But is here.

bartlomieju (approved):
LGTM

PR #11954, opened 2021-09-08
fix(cli): better handling of source maps

### LINKED ISSUE #11874: Inconsistent Source Maps leads to debug usability issues
When emitting sources in Deno, we have inconsistent behaviour and likely incompatible behaviour, which leads to challenges when trying to debug TypeScript code using the debug API with something like Chrome Dev Tools or VSCode.

- When emitting with tsc (when type checking for example), the source file is emitted without inlined sources and the original source is listed in the source map of a sibling of the emitted file.  For example if you use `https://deno.land/std@0.106.0/testing/asserts.ts` in your program, the source map indicates that the original source is `asserts.ts`.  This means that when inspecting and the source map is unwrapped, the original source cannot be found, because the emitted source is located in `$DENO_DIR/gen/https/deno.land/<HASHEDVALUE>.js` while the original source is located in `$DENO_DIR/dep/https/deno.land/<HASHVALUE>` (without any extension) which the debugger has no way if figuring out how to get to.
- When emitting with swc (like with `--no-check`) the source file is emitted with inlined sources, but the source is listed as `"<https://deno.land/std@0.106.0/testing/asserts.ts>"`.  The `<` and `>` braces cause strange behaviour in the debuggers, as well as provide strange results in the likes of vscode, because it can't figure out the language ID of the file.

While embedding the original sources in the emitted source map adds "data" to the emitted file, this is negligible, because it is all done locally via the local cache (and we are already doing it when transpiling with swc), so I think we should embed sources in the output and try to get both tsc and swc to output consistent fully qualified specifiers in the `"sources"` property of the emitted source maps.

Ideally we would get our external source maps working again, I remember we stopped at some point because @bartlomieju ran into an intractable problem and I can't remember what it was.

kitsonk:
Ref: #11874
Ref: swc-project/swc#2218

This partially resolves #11874, in that it inlines sources when tsc does the emitting, which makes the sources available when debugging.  A recent change to swc though has caused sources to not be inlined and there is no way to restore them.

There is also a very strange organisation of sources in the source view here that I don't know if is easy to "fix":

state_ts_—_deploy_doc

But the overall improvement is worth merging IMO, and we can keep the issue open while trying to fix the other things.


bartlomieju:
> There is also a very strange organisation of sources in the source view here that I don't know if is easy to "fix":

Can you elaborate on this? What's strange about this organization? It looks fine at first glance

bartlomieju (approved):
LGTM

PR #11955, opened 2021-09-08
fix(ext/http): resource leak if request body is not consumed

### LINKED ISSUE #11926: hyper server leaks RequestResource objects
Can be observed by logging the output of `Deno.core.resources()` after every request.

Good:
```
{"0":"tcpListener","2":"httpConnection","3":"responseSender"}
{"0":"tcpListener","5":"httpConnection","6":"responseSender"}
{"0":"tcpListener","8":"httpConnection","9":"responseSender"}
{"0":"tcpListener","11":"httpConnection","12":"responseSender"}
```
Bad:
```
{"0":"tcpListener","2":"httpConnection","4":"responseSender","3":"request"}
{"0":"tcpListener","6":"httpConnection","8":"responseSender","3":"request","7":"request"}
{"0":"tcpListener","10":"httpConnection","12":"responseSender","3":"request","7":"request","11":"request"}
{"0":"tcpListener","14":"httpConnection","16":"responseSender","3":"request","7":"request","11":"request","15":"request"}
```
It's more obvious with http/2 but http/1 POST + request body shows the same behavior.

(http/1 GET does not but I suspect it takes a shortcut and doesn't create a `RequestResource`.)

I'm unclear if the `RequestResource` objects are cleaned up when the connection object is closed, but they're definitely not being cleaned up after the response is sent.

bartlomieju:
Fixes #11926

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -882,3 +882,45 @@ unitTest(
     clientConn.close();
   },
 );
+
+// https://github.com/denoland/deno/issues/11926
+unitTest(
+  { perms: { net: true } },
+  async function httpServerDoesntLeakResources2() {
+    let listener: Deno.Listener;
+    let httpConn: Deno.HttpConn;
+
+    const promise = (async () => {
+      listener = Deno.listen({ port: 4502 });
+      for await (const conn of listener) {
+        httpConn = Deno.serveHttp(conn);
+        for await (const { request, respondWith } of httpConn) {
+          assertEquals(new URL(request.url).href, "http://127.0.0.1:4502/");
+          // not reading request body on purpose
+          respondWith(new Response("ok"));
+        }
+      }
+    })();
+
+    const resourcesBefore = Deno.resources();
+    const response = await fetch("http://127.0.0.1:4502", {
+      method: "POST",
+      body: "hello world",
+    });
+    await response.text();
+    const resourcesAfter = Deno.resources();
+    // verify that the only new resource is "httpConnection", to make
+    // sure "request" resource is closed even if its body was not read
+    // by server handler
+
+    for (const rid of Object.keys(resourcesBefore)) {
+      delete resourcesAfter[rid];
+    }
+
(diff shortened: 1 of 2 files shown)


bartlomieju:
On a related note: there are two code paths in `createRespondWith` that can throw `TypeError` - I believe similar cleanup should be done in those paths. @lucacasonato what do you think?

cli/tests/unit/http_test.ts
```diff
@@ -882,3 +882,45 @@ unitTest(
...
+      body: "hello world",
+    });
+    await response.text();
+    const resourcesAfter = Deno.resources();
+    // verify that the only new resource is "httpConnection", to make
+    // sure "request" resource is closed even if its body was not read
+    // by server handler
+
+    for (const rid of Object.keys(resourcesBefore)) {
+      delete resourcesAfter[Number(rid)];
```
bnoordhuis, line 917: Why Number? `resourcesAfter` is not an array, right?

ext/http/01_http.js
```diff
@@ -297,6 +297,13 @@
 
           ws[_eventLoop]();
         }
+      } else {
+        // Try to close "request" resource. It might have been already consumed,
+        // but if it hasn't been we need to close it here to avoid resource leak.
+        try {
+          SetPrototypeDelete(httpConn.managedResources, requestRid);
+          core.close(requestRid);
+        } catch { /* pass */ }
```
bnoordhuis, line 306: I don't really understand why it's inside a swallow-everything try/catch - seems like that could hide genuine bugs - but it's a pattern that's used elsewhere in this file so :shrug:

bnoordhuis (approved):

cli/tests/unit/http_test.ts
bartlomieju: TS was complaining about `rid` being any type 🤷

ext/http/01_http.js
bartlomieju: `core.close()` should only throw `BadResource` when the thing has already been closed.

PR #11962, opened 2021-09-09
SECURITY.md

dzoudyy:
ok


lucacasonato:
The info in here is incorrect and not relevant.

PR #11965, opened 2021-09-09
chore: upgrade Rust to 1.55.0

magurotuna:

### DIFF
--- a/runtime/errors.rs
+++ b/runtime/errors.rs
@@ -60,7 +60,7 @@ fn get_io_error_class(error: &io::Error) -> &'static str {
     WouldBlock => unreachable!(),
     // Non-exhaustive enum - might add new variants
     // in the future
-    _ => unreachable!(),
+    _ => "Error",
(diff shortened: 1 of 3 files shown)


runtime/errors.rs
```diff
@@ -60,7 +60,7 @@ fn get_io_error_class(error: &io::Error) -> &'static str {
     WouldBlock => unreachable!(),
     // Non-exhaustive enum - might add new variants
     // in the future
-    _ => unreachable!(),
+    _ => "Error",
```
magurotuna, line 63: https://blog.rust-lang.org/2021/09/09/Rust-1.55.0.html#stdioerrorkind-variants-updated

Since 1.55.0, errors that were previously `ErrorKind::Other` have been changed to `ErrorKind::Uncategorized`. To match the `Uncategorized` we have to use wildcard pattern.

ry (approved):
Thanks!

PR #11969, opened 2021-09-09
fix: permission prompt stuffing on Windows

dsherret:
This unfortunately makes Windows act differently in that pressing the "y" or "n" key immediately confirms or denies the prompt. It would take a lot more code to make this work otherwise.

Also, I looked into using https://docs.microsoft.com/en-us/windows/console/writeconsoleinput to create unit tests, but I couldn't figure out a way to construct an [`INPUT_RECORD`](https://docs.rs/winapi/0.3.9/winapi/um/wincontypes/struct.INPUT_RECORD.html).

image


runtime/permissions.rs
```diff
@@ -1196,55 +1194,87 @@ fn permission_prompt(message: &str) -> bool {
...
+      // blocking, so instead we deal with the lower level input API, and
+      // listen for key events.
+      loop {
+        let mut events_read = 0;
+        let mut input_record = std::mem::zeroed();
+        assert_eq!(
+          ReadConsoleInputW(stdin, &mut input_record, 1, &mut events_read),
+          1
+        );
+        if input_record.EventType == KEY_EVENT {
```
piscisaureus, line 1254: Add `assert_eq!(events_read, 1)`.

runtime/permissions.rs
```diff
@@ -1196,55 +1194,87 @@ fn permission_prompt(message: &str) -> bool {
...
+      // flushed from the input buffer. There doesn't seem to be a way to
+      // flush this other buffer or read from ReadConsole without potentially
+      // blocking, so instead we deal with the lower level input API, and
+      // listen for key events.
+      loop {
+        let mut events_read = 0;
+        let mut input_record = std::mem::zeroed();
+        assert_eq!(
+          ReadConsoleInputW(stdin, &mut input_record, 1, &mut events_read),
+          1
```
piscisaureus, line 1252: Nit: since this function returns a `BOOL` value, it would be more readable to compare to `winapi::shared::minwindef::TRUE`.

runtime/permissions.rs
dsherret: Done in c675727c3cbda74ebe557445499b46b53c2e9cc1

piscisaureus:
> I couldn't figure out a way to construct an [`INPUT_RECORD`](https://docs.rs/winapi/0.3.9/winapi/um/wincontypes/struct.INPUT_RECORD.html).

I think it'd be something like this:

```rust
let mut input_record = INPUT_RECORD { EventType: KEY_EVENT, ..Default::default() };
input_record.Event.KeyEvent_mut().wRepeatCount = 1;
input_record.Event.KeyEvent_mut().uChar.UnicodeChar_mut() = 'y' as WCHAR;
```

dsherret:
@piscisaureus oh, nice! Thanks for that. I will look into adding a test tomorrow.

piscisaureus:
> This unfortunately makes Windows act differently in that pressing the "y" or "n" key immediately confirms or denies the prompt. It would take a lot more code to make this work otherwise.

What makes it difficult in particular?  Afaik `ReadConsole()` provides a way to do line-buffered input.

dsherret:
@piscisaureus calling `ReadConsoleW` would return the permission stuffed text even after calling `FlushConsoleInputBuffer`. I think perhaps `FlushConsoleInputBuffer` flushes the input buffer events to a character buffer that `ReadConsoleW` then reads from (that would be my guess given the behaviour I'm seeing... you may want to try something out and see if you can figure out a way to flush that buffer... I haven't been able to figure it out yet).

https://github.com/denoland/deno/pull/11969/files#diff-ddef1e94cdb9cee582c71ed2e1a3948e4a1a80cdcf311693b8a41d13a79aa418R1247-R1252

piscisaureus:
> @piscisaureus calling `ReadConsoleW` would return the permission stuffed text even after calling `FlushConsoleInputBuffer`. I think perhaps `FlushConsoleInputBuffer` flushes the input buffer events to a character buffer that `ReadConsoleW` then reads from 
 
I would be very surprised if this was case...

dsherret:
@piscisaureus in the code in this PR if you try adding:

```rust
      assert_eq!(
        FlushConsoleInputBuffer(stdin),
        winapi::shared::minwindef::TRUE
      );

      // add this after the existing `FlushConsoleInputBuffer` above
      loop {
        let mut buffer = [0u16; 1024];
        let mut chars_read = 0;
        assert_eq!(winapi::um::consoleapi::ReadConsoleW(stdin, buffer.as_mut_ptr() as _, buffer.len() as _, &mut chars_read, std::ptr::null_mut()), 1);
        println!("Chars read: {}", chars_read);
        let text = std::ffi::OsString::from_wide(&buffer[0..chars_read as usize]).to_string_lossy().to_string();
        println!("Text: {:?}", text);
      }
```

...then running [this JS code](https://github.com/denoland/deno/pull/11969#issuecomment-916479831) if you type `yy<ENTER>` then you will see the following output:

image

Maybe I'm doing something wrong? Again, I'm not super familiar with the winapi. Hopefully there is a way to do this that I'm not seeing.

piscisaureus:
I’ll take a look and try to figure this out tomorrow.

piscisaureus:
I wrote a simple standalone test: https://gist.github.com/piscisaureus/42fd7536b075f447682977083c7ff678
It seems that `FlushConsoleInputBuffer()` is working fine with `ReadConsoleW()`.

dsherret:
@piscisaureus try my fork https://gist.github.com/dsherret/72991b395d68ac6aff912199e5e44ac5

piscisaureus (approved):

runtime/permissions.rs
```diff
@@ -1203,13 +1201,69 @@ fn permission_prompt(message: &str) -> bool {
...
+      if result != TRUE {
+        panic!(
+          "Error flushing console input buffer: {}",
+          std::io::Error::last_os_error().to_string()
+        )
+      }
+    }
+
+    unsafe fn emulate_enter_key_press(stdin: HANDLE) {
+      // https://github.com/libuv/libuv/blob/a39009a5a9252a566ca0704d02df8dabc4ce328f/src/win/tty.c#L1121-L1131
```
dsherret, line 1241: Bert pointed me in the direction of this emulated enter key press.

With the current solution the output looks like this when there is a buffered line:

image

And like this when there's not (the more common case)... note the blank line:

image

Is it worth adding to code ensure this blank line does not occur? I think any easy fix would be to peak at the input buffer before flushing it and if there is nothing in the input buffer it will move the console cursor up a line.

dsherret: Resolved in 7b93fbfeba6e78b39c8a1c482474911cee1d0d00 after discussion with @piscisaureus. When the console input buffer is empty (meaning a key press was emulated) the cursor will be moved up to prevent a blank line.

Issue #11975, opened 2021-09-10
Bug in import maps when using quertstring


felixsanz:
i have this import_map.json:

```
{
   "imports": {
      "npm/": "https://esm.sh/"
   }
}
```

but when i do `import cssModules from 'npm/postcss-modules@4.2.2?no-check` it doesn't work as it tries to load `Download https://esm.sh/postcss-modules@4.2.2%3Fno-check`

Inside Chrome it works fine, so Deno shouldn't be escaping the querystring symbol i think.

```
deno 1.13.2 (release, x86_64-unknown-linux-gnu)
v8 9.3.345.11
typescript 4.3.5
```

PR #11551, opened 2021-07-29
fix: support windows file specifiers with import maps

dsherret:
Fixes #11530.

cc @caspervonb

PR #11976, opened 2021-09-10
fix: Query string percent-encoded in import map

bartlomieju:
Regression introduced in https://github.com/denoland/deno/pull/11551
Fixes https://github.com/denoland/deno/issues/11975

### DIFF
--- a/cli/import_map.rs
+++ b/cli/import_map.rs
@@ -197,29 +197,29 @@ impl ImportMap {
     // the url since it contains what looks to be a uri scheme. To work around
     // this, we append the specifier to the path segments of the base url when
     // the specifier is not relative or absolute.
-    let mut maybe_query_string = None;
+    let mut maybe_query_string_and_fragment = None;
     if !is_relative_or_absolute_specifier && base.path_segments_mut().is_ok() {
       {
         let mut segments = base.path_segments_mut().unwrap();
         segments.pop_if_empty();
 
-        // Handle query string first, otherwise it would be percent-encoded
+        // Handle query-string and fragment first, otherwise they would be percent-encoded
         // by `extend()`
-        let mut parts: Vec<&str> = specifier.split('?').collect();
-        assert!(parts.len() == 1 || parts.len() == 2);
-        if parts.len() == 2 {
-          maybe_query_string = parts.pop();
-        }
-
-        segments.extend(parts[0].split('/'));
+        let prefix = match specifier.find(&['?', '#'][..]) {
+          Some(idx) => {
+            maybe_query_string_and_fragment = Some(&specifier[idx..]);
+            &specifier[..idx]
+          }
+          None => specifier,
+        };
+        segments.extend(prefix.split('/'));
       }
 
-      // Append query string back, if it exists
-      if let Some(query_string) = maybe_query_string {
-        base.set_query(Some(query_string));
+      if let Some(query_string_and_fragment) = maybe_query_string_and_fragment {
+        Ok(base.join(query_string_and_fragment)?)
+      } else {
+        Ok(base)
       }
-
-      Ok(base)
     } else {
       Ok(base.join(specifier)?)
     }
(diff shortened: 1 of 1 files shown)


cli/import_map.rs
```diff
@@ -197,12 +197,28 @@ impl ImportMap {
...
     if !is_relative_or_absolute_specifier && base.path_segments_mut().is_ok() {
       {
         let mut segments = base.path_segments_mut().unwrap();
         segments.pop_if_empty();
-        segments.extend(specifier.split('/'));
+
+        // Handle query string first, otherwise it would be percent-encoded
+        // by `extend()`
+        let mut parts: Vec<&str> = specifier.split('?').collect();
+        assert!(parts.len() == 1 || parts.len() == 2);
```
andreubotella, line 209: With this change, `import "key/a?b?c"` would panic.

bartlomieju: Is it actually valid specifier? Shouldn't it contain at most single `?` and the rest should be percent encoded?

andreubotella: It's perfectly valid for a URL query string to contain `?`:

```
> new URL("https://example.com/a?b?c")
URL {
  href: "https://example.com/a?b?c",
  origin: "https://example.com",
  protocol: "https:",
  username: "",
  password: "",
  host: "example.com",
  hostname: "example.com",
  port: "",
  pathname: "/a",
  hash: "",
  search: "?b?c"
}
```

And even with a character like `<`, which does get percent-encoded in the query string, you'd reach step 1.2.5 in the spec's "resolve an imports match" algorithm, which would parse the specifier minus the prefix. The URL parsing wouldn't fail – rather than that, the serialization of _url_ then would have that character percent-encoded.

lucacasonato: Doesn't the spec mandate how to handle this case?

andreubotella: Per the spec, the `append_specifier_to_base` function should just do `base.join(specifier)`. The complication comes because we accept specifiers which are Windows paths. (I'm not sure why we must support them as suffixes of an import map match, but whatever.)

In any case, since Windows filenames don't support the question mark, so assuming we don't support UNC paths (which start with `\\?\`), Windows paths won't contain it either. So we can just split at the first question mark.

Also, what if the specifier contains a fragment?

bartlomieju: > So we can just split at the first question mark.

So you're suggesting to just treat everything after the first question mark as a "query string"?

> Also, what if the specifier contains a fragment?

Good point, we should probably also split on first hash (`#`) and append it manually.

It's unfortunate that handling Windows paths adds so much complexity and as can be seen in this example bugs.

andreubotella: > So you're suggesting to just treat everything after the first question mark as a "query string"?

How about splitting the specifier by the first either question mark or hash, and if there's a suffix, parse it with the extended base:

```rust
let prefix = match specifier.find(&['?', '#']) {
  Some(idx) => {
    maybe_query_string = Some(&specifier[..idx]);
    &specifier[..idx]
  },
  None => specifier
};
segments.extend(prefix.split('/'));

// ...

if let Some(query_string_and_fragment) = maybe_query_string {
  Ok(base.join(query_string_and_fragment))
} else {
  Ok(base)
}
```

bartlomieju: Thanks Andreu, I applied you suggestions and added you as co-author

lucacasonato (approved):
LGTM. Are there no upstream WPT tests for this? If we find edge cases like these in web specs, we should upstream tests to make sure that what we are implementing is and stays the same as other implementations (even if what we say is correct per spec).

https://github.com/web-platform-tests/wpt/tree/master/import-maps/data-driven/resources <- best would be to add a test here. Just mention me on a WPT PR, and I can review it :-)

bartlomieju:
> LGTM. Are there no upstream WPT tests for this? If we find edge cases like these in web specs, we should upstream tests to make sure that what we are implementing is and stays the same as other implementations (even if what we say is correct per spec).
> 
> https://github.com/web-platform-tests/wpt/tree/master/import-maps/data-driven/resources <- best would be to add a test here. Just mention me on a WPT PR, and I can review it :-)

Indeed it seems there are no tests that resolve specifiers with `?` or ` #`. I will upstream a couple of over the weekend; 3 simple cases would have caught the problem.

lucacasonato:
Thanks!

PR #11978, opened 2021-09-10
perf(ext/http): optimize auto cleanup of request resource

### LINKED ISSUE #11963: Performance regression after fixing resource leak in HTTP bindings.
#11955 caused a performance regression in HTTP bindings.

image0

piscisaureus:
Fixes #11963.

### DIFF
--- a/core/ops_builtin.rs
+++ b/core/ops_builtin.rs
@@ -17,6 +17,7 @@ pub(crate) fn init_builtins() -> Extension {
     ))
     .ops(vec![
       ("op_close", op_sync(op_close)),
+      ("op_try_close", op_sync(op_try_close)),
       ("op_print", op_sync(op_print)),
       ("op_resources", op_sync(op_resources)),
     ])
@@ -44,10 +45,24 @@ pub fn op_close(
   rid: Option<ResourceId>,
   _: (),
 ) -> Result<(), AnyError> {
-  // TODO(@AaronO): drop Option after improving type-strictness balance in serde_v8
+  // TODO(@AaronO): drop Option after improving type-strictness balance in
+  // serde_v8
   let rid = rid.ok_or_else(|| type_error("missing or invalid `rid`"))?;
   state.resource_table.close(rid)?;
+  Ok(())
+}
 
+/// Try to remove a resource from the resource table. If there is no resource
+/// with the specified `rid`, this is a no-op.
+pub fn op_try_close(
+  state: &mut OpState,
+  rid: Option<ResourceId>,
+  _: (),
+) -> Result<(), AnyError> {
+  // TODO(@AaronO): drop Option after improving type-strictness balance in
+  // serde_v8.
+  let rid = rid.ok_or_else(|| type_error("missing or invalid `rid`"))?;
+  let _ = state.resource_table.close(rid);
   Ok(())
 }
 
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -148,6 +148,10 @@
     opSync("op_close", rid);
   }
 
+  function tryClose(rid) {
+    opSync("op_try_close", rid);
+  }
+
   function print(str, isErr = false) {
     opSync("op_print", str, isErr);
   }
@@ -175,6 +179,7 @@
     opSync,
     ops,
     close,
+    tryClose,
     print,
     resources,
     registerErrorBuilder,
--- a/ext/http/01_http.js
+++ b/ext/http/01_http.js
@@ -297,13 +297,12 @@
 
           ws[_eventLoop]();
         }
-      } else {
+      } else if (typeof requestRid === "number") {
         // Try to close "request" resource. It might have been already consumed,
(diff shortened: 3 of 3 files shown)


core/ops_builtin.rs
```diff
@@ -44,10 +45,24 @@ pub fn op_close(
...
+/// Try to remove a resource from the resource table. If there is no resource
+/// with the specified `rid`, this is a no-op.
+pub fn op_try_close(
+  state: &mut OpState,
+  rid: Option<ResourceId>,
+  _: (),
+) -> Result<(), AnyError> {
+  // TODO(@AaronO): drop Option after improving type-strictness balance in
+  // serde_v8.
+  let rid = rid.ok_or_else(|| type_error("missing or invalid `rid`"))?;
```
AaronO, line 64: I would ignore this error too

AaronO (dismissed):
Pretty much what I was planning to do (on Monday ...), looks good, though I would also allow rid to be null/undefined too so it's really fire and forget "close if there is something to close and be silent"

core/ops_builtin.rs
bartlomieju: What's the point of calling into Rust if `rid` is `null` or `undefined`? Seems like overly relaxed API

piscisaureus: > What's the point of calling into Rust if `rid` is `null` or `undefined`? Seems like overly relaxed API

I don't quite follow. Did you mean "overly strict"?

I agree that we shouldn't call into rust when `rid` is null or undefined, so in order to detect those cases we *should* throw when the user does that.

AaronO: Agreed, but IIRC there are some tests where we call `Deno.close()` and other things of the sort on potentially nullish values, so when rewriting the op-layer I remained compatible the prior approach. We could add that check in `core.js`, will revisit that and other op-layer changes in the coming weeks

bartlomieju: > > What's the point of calling into Rust if `rid` is `null` or `undefined`? Seems like overly relaxed API
> 
> I don't quite follow. Did you mean "overly strict"?
> 
> I agree that we shouldn't call into rust when `rid` is null or undefined, so in order to detect those cases we _should_ throw when the user does that.

I meant I'm not in favor of the proposed changed and the API is good as it is now.

> Agreed, but IIRC there are some tests where we call Deno.close() and other things of the sort on potentially nullish values, so when rewriting the op-layer I remained compatible with the previous API. We could add that check in core.js, will revisit that and other op-layer changes in the coming weeks

But this is a new API and it doesn't interfere with `Deno.close()`, IMO the check should be done in JS for nullish values before calling, it's surely faster than crossing the boundary between V8 and Rust.

bartlomieju: That said, I'm not entirely sure why the type of argument to these ops is `Option<ResourceId>`... If it was `ResourceId` then nullish values would be handled by `serde_v8`, wouldn't they?

piscisaureus:
```
  "req_per_sec": {
    "deno_tcp": 63747,
    "node_proxy": 2487,
    "node_http": 20239,
    "core_http_json_ops": 75343,
    "deno_proxy_tcp": 24360,
    "node_tcp": 54886,
    "node_proxy_tcp": 25079,
    "deno_http": 26781,
    "deno_http_native": 26950,
    "hyper": 70560
  },
```

core/ops_builtin.rs
piscisaureus: > If it was `ResourceId` then nullish values would be handled by `serde_v8`, wouldn't they?

Good point - I just copypasta'd from `op_close()`. @AaronO?

bartlomieju (approved):
LGTM

core/ops_builtin.rs
piscisaureus: > If it was `ResourceId` then nullish values would be handled by `serde_v8`, wouldn't they?

It seems that serde_v8 coerces to `0` in that case. With the proposed change:

```
piscisaureus@dumbo:~/d/deno 🦕 ./target/debug/deno
Deno 1.13.2
exit using ctrl+d or close()
> Deno.core.resources()
{ "0": "stdin", "1": "stdout", "2": "stderr" }
> Deno.core.tryClose()
undefined
> Deno.core.resources()
{ "1": "stdout", "2": "stderr" }
> 
```

AaronO: Yeah `serde_v8` deserializes ints via `.integer_value()` / `v8::Value::IntegerValue`, hence the note about type-strictness & coercion. For example other non `true` values coerce to `false` when deserializing a `bool`. Parts of our ops depended on that implicit coercion, so I added it to match existing behaviour vs changing all those call sites and delaying the `serde_v8` op-layer work.

I'll revisit this consistency and other op-layer work in the coming weeks

bartlomieju: @AaronO sounds good, thanks!

PR #11980, opened 2021-09-10
refactor: use Deno.core.tryClose

bartlomieju:
Using API introduced in https://github.com/denoland/deno/pull/11978 we can get rid of some more code throughout the codebase.


piscisaureus (approved):
LGTM

PR #11983, opened 2021-09-11
feat(cli): lint watch mode

### LINKED ISSUE #11132: deno lint --watch
Similar to other subcommands linter should support watching for file changes (https://github.com/denoland/deno/issues/7465).

CGQAQ:
Signed-off-by: CGQAQ <m.jason.liu@outlook.com>

fix #11132


magurotuna (commented):
Can you add integration tests to make sure it works fine? You can refer to the existing examples of watcher tests in https://github.com/denoland/deno/blob/d236f432b86de55c6006778b0c68fe60b6419069/cli/tests/integration/watcher_tests.rs

CGQAQ:
> Can you add integration tests to make sure it works fine? You can refer to the existing examples of watcher tests in https://github.com/denoland/deno/blob/d236f432b86de55c6006778b0c68fe60b6419069/cli/tests/integration/watcher_tests.rs

sure, I'll definitely do that

CGQAQ:
@magurotuna Added tests, please re-review it :)

cli/tools/lint.rs
```diff
@@ -406,6 +497,10 @@ impl LintReporter for JsonLintReporter {
     });
   }
 
+  fn reset_counter(&mut self) {
+    unimplemented!()
+  }
```
bartlomieju, line 502: This is strange... why is this function needed?

CGQAQ: https://github.com/denoland/deno/blob/4ce2c8b7b903f2ab4c883a85767d8f394aaf385a/cli/tools/lint.rs#L157

I need to call this function to reset the lint counter, otherwise, it will keep increasing between multiple runs of the operation

bartlomieju: Why not do it in `close()` instead?

CGQAQ: Or maybe I can create the report inside the operation closure every single time

cli/tools/lint.rs
```diff
@@ -98,6 +105,85 @@ pub async fn lint_files(
...
+  } else {
+    lint_files(
+      include_files.clone(),
+      exclude_files.clone(),
+      args.clone(),
+      reporter_lock.clone(),
+      lint_rules.clone(),
+      has_error.clone(),
+    )
+    .await?;
```
bartlomieju, line 173: It looks like `lint_files` is superfluous in this case and could be removed. Take a look at `cli/tools/fmt.rs` in `format` function - it calls `resolver()` and `operation()` manually to avoid code duplication

cli/tools/lint.rs
bartlomieju: Also it seems that reporter could just be created on each restart of file watcher, that would allow to keep it "single use" and we won't need to worry about its internal state

cli/tests/testdata/badly_linted.mjs
```diff
@@ -0,0 +1,2 @@
+// Deliberately using .mjs to avoid triggering deno lint on ci
+let a = 5;
```
bartlomieju, line 2: Put these files in `cli/tests/testdata/lint` (or a subdirectory in that file) to avoid the problem from the comment

cli/tests/integration/watcher_tests.rs
```diff
@@ -48,6 +66,71 @@ fn wait_for_process_failed(
   }
 }
 
+#[test]
+fn lint_watch_test() {
+  let t = TempDir::new().expect("tempdir fail");
+  let badly_linted_original = util::testdata_path().join("badly_linted.mjs");
```
bartlomieju, line 72: Nice test

cli/tools/lint.rs
CGQAQ: oh! my bad, just didn't think about it(didn't want to change existing behaviours, I will change it

CGQAQ:
Will change these tonight, it’s Monday morning here, gotta go to work

cli/tests/testdata/badly_linted.mjs
CGQAQ: I thought deno lint will linting all js files, so I did the same trick in https://github.com/denoland/deno/blob/2cc1577d286814d941000fc4653f6e8d59b2e190/cli/tests/testdata/badly_formatted.mjs#L1

CGQAQ:
When will this gets to merge

caspervonb:
> When will this gets to merge

It is tagged for 1.5, merge window for that opens in about two weeks 🙂

cli/tests/testdata/lint/expected_glob.out
```diff
@@ -1,3 +1,3 @@
 [WILDCARD]
-Found 3 problems
-Checked 3 files
+Found 6 problems
+Checked 6 files
```
bartlomieju, line 3: This test needs to be fixed, it should change if new test cases are added. I will look into it in a separate PR.

cli/tools/lint.rs
```diff
@@ -89,27 +104,75 @@ pub async fn lint_files(
...
-  let has_error = Arc::new(AtomicBool::new(false));
-
-  let reporter_kind = if json {
-    LintReporterKind::Json
-  } else {
-    LintReporterKind::Pretty
+  let resolver = |changed: Option<Vec<PathBuf>>| {
+    let files_changed = changed.is_some();
+    let result =
+      collect_files(&args.clone(), &ignore.clone(), is_supported_ext).map(
```
bartlomieju, line 110:
```suggestion
      collect_files(include_files.clone(), exclude_files.clone(), is_supported_ext).map(
```

cli/tools/lint.rs
```diff
@@ -89,27 +104,75 @@ pub async fn lint_files(
...
+
+    Ok(())
+  };
+  if watch {
+    file_watcher::watch_func(resolver, operation, "Lint").await?;
+  } else {
     if args.len() == 1 && args[0].to_string_lossy() == "-" {
+      let reporter_lock =
+        Arc::new(Mutex::new(create_reporter(reporter_kind.clone())));
       let r = lint_stdin(lint_rules);
```
bartlomieju, line 168: I believe there might be an edge case when running `deno lint --watch -`, this will error out

CGQAQ: image

I tried without any error, should I add a guard to tell the user cannot watch stdin?

bartlomieju: > I tried without any error, should I add a guard to tell the user cannot watch stdin?

I think that makes sense, please do.

cli/tests/testdata/lint/expected_glob.out
bartlomieju: Waiting on https://github.com/denoland/deno/pull/12317

bartlomieju: Resolved now.

bartlomieju (approved):
LGTM, thanks @CGQAQ

Issue #10553, opened 2021-05-10
Improve the error message from `stat/statSync`


canonic-epicure:
Currently the error message of the exception, thrown from the `Deno.stat/statSync`, when the file is missing, does not include the file name:
```
nickolay@frontier:~/workspace/Bryntum/siesta-monorepo/siesta$ deno
Deno 1.9.2
exit using ctrl+d or close()
> Deno.statSync('/home/nickolay/not_existing_file')
Uncaught NotFound: No such file or directory (os error 2)
    at unwrapOpResult (deno:core/core.js:99:13)
    at Object.opSync (deno:core/core.js:113:12)
    at Object.statSync (deno:runtime/js/30_fs.js:224:22)
    at <anonymous>:2:6
> 
```
Is this intentional with some rationale behind, or just an oversight? From usability perspective the filename should be included in the message of course.

bnoordhuis:
Pull request welcome!

The error message is bubbled up straight from the Rust runtime. The relevant code is here:

https://github.com/denoland/deno/blob/1cd1419e18386d512ab98849d132117182545b89/runtime/ops/fs.rs#L846-L850

It's possible to remap the error message to include the filename but it should be done in a principled manner, i.e., for more than just stat.

With ops that operate on more than one path (like link) you don't know which path is the problematic one. I'd include both in the error message, that's what Node does.

CGQAQ:
![image](https://user-images.githubusercontent.com/15936231/118353751-46105a80-b59a-11eb-8bbb-9c3a149d8f69.png)

here's how the node prints out the error message, I should print it just like that?

CGQAQ:
![image](https://user-images.githubusercontent.com/15936231/118354196-7822bc00-b59c-11eb-8ccb-1b67676f0c54.png)

Is this ok or not? @bnoordhuis

abakfja:
@CGQAQ if you are not working on this, can I take this up?

CGQAQ:
Sure

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: Kannav Mehta ***@***.***>
发送时间: Wednesday, June 30, 2021 1:18:07 AM
收件人: denoland/deno ***@***.***>
抄送: CGQAQ ***@***.***>; Mention ***@***.***>
主题: Re: [denoland/deno] Improve the error message from `stat/statSync` (#10553)


@CGQAQ<https://github.com/CGQAQ> if you are not working on this, can I take this up?

―

[...]

PR #10642, opened 2021-05-15
Added path info to error message from stat/statSync

CGQAQ:
Fixed #10553

PR #11984, opened 2021-09-11
feat(runtime): improve error messages of runtime fs

F3n67u:
fix  https://github.com/denoland/deno/issues/10553

related pr: https://github.com/denoland/deno/pull/10642

output:
```
% ./target/debug/deno 
Deno 1.13.2
exit using ctrl+d or close()
> Deno.statSync('/home/nickolay/not_existing_file')
Uncaught NotFound: No such file or directory (os error 2), stat '/home/nickolay/not_existing_file'
    at Object.opSync (deno:core/01_core.js:140:12)
    at Object.statSync (deno:runtime/js/30_fs.js:236:22)
    at <anonymous>:2:6
```


ahabhgk:
The tests failed because the `renameSyncErrorsUnix` will check the error message, change `fs_err::rename` back to `std::fs::rename` separately will pass the test, but It may be a little weird that only rename uses the method in the std::fs, so may be we can change the `renameSyncErrorsUnix` test to let the code pass or write the error message for every fs method manually, any suggestions? @bartlomieju @F3n67u

F3n67u:
> The tests failed because the `renameSyncErrorsUnix` will check the error message, change `fs_err::rename` back to `std::fs::rename` separately will pass the test, but It may be a little weird that only rename uses the method in the std::fs, so may be we can change the `renameSyncErrorsUnix` test to let the code pass or write the error message for every fs method manually, any suggestions? @bartlomieju @F3n67u

I'll prefer the latter approach: write the error message for every fs method manually.

bartlomieju:
> > The tests failed because the `renameSyncErrorsUnix` will check the error message, change `fs_err::rename` back to `std::fs::rename` separately will pass the test, but It may be a little weird that only rename uses the method in the std::fs, so may be we can change the `renameSyncErrorsUnix` test to let the code pass or write the error message for every fs method manually, any suggestions? @bartlomieju @F3n67u
> 
> I'll prefer the latter approach: write the error message for every fs method manually.

I agree, message should be written manually.

F3n67u:
@bartlomieju I already add path info to all `std::fs` call's `error` in runtime `fs`. Could you help to review? 

I don't known how to add path info to `permissions.read.check(&path)?` and `permissions.write.check(&path)?` bec those method return `anyhow::Error` instead of `std::io::Error`. I will very appreciate if you can help with this problem.

bartlomieju:
> @bartlomieju I already add path info to all `std::fs` call's `error` in runtime `fs`. Could you help to review?

The PR looks good! I'm sure there will be some bike-shedding regarding what errors should print exactly, but it's not a big deal.

> I don't known how to add path info to `permissions.read.check(&path)?` and `permissions.write.check(&path)?` bec those method return `anyhow::Error` instead of `std::io::Error`(I don't know how to add path info to `anyhow::Error` and make path info print). I will very appreciate if you can help with this problem.

Permission errors coming from `read.check()` and `write.check()` already have paths printed out so there's no need to do anything there.

Would be great to see some updated to the unit tests that call `assertThrows` or `assertRejects` and verify that expected filepaths are included in the error message.

bartlomieju:
@dsherret do you have any suggestion as for the error messages?

@F3n67u please add some tests asserting error message contain paths so we can land this PR.

dsherret:
@bartlomieju nope, it looks good to me other than the tests.

F3n67u:
> > @bartlomieju I already add path info to all `std::fs` call's `error` in runtime `fs`. Could you help to review?
> 
> The PR looks good! I'm sure there will be some bike-shedding regarding what errors should print exactly, but it's not a big deal.
> 
> > I don't known how to add path info to `permissions.read.check(&path)?` and `permissions.write.check(&path)?` bec those method return `anyhow::Error` instead of `std::io::Error`(I don't know how to add path info to `anyhow::Error` and make path info print). I will very appreciate if you can help with this problem.
> 
> Permission errors coming from `read.check()` and `write.check()` already have paths printed out so there's no need to do anything there.
> 
> Would be great to see some updated to the unit tests that call `assertThrows` or `assertRejects` and verify that expected filepaths are included in the error message.

@bartlomieju sorry for slow reply. I am on the holiday. I will update the unit tests when i am back.

F3n67u:
> @dsherret do you have any suggestion as for the error messages?
> 
> @F3n67u please add some tests asserting error message contain paths so we can land this PR.

@bartlomieju I will back 2 days later. If you hurry, You can pick up and update the unit tests.

bartlomieju:
> > @dsherret do you have any suggestion as for the error messages?
> > @F3n67u please add some tests asserting error message contain paths so we can land this PR.
> 
> @bartlomieju I will back 2 days later. If you hurry, You can pick up and update the unit tests.

It's fine, enjoy your holiday :)

bartlomieju:
@F3n67u this is close to landing, there are two failures in tests on Windows.

F3n67u:
> @F3n67u this is close to landing, there are two failures in tests on Windows.

@bartlomieju Those two fail is fixed and test is included. ready to review again.

PR #11985, opened 2021-09-11
refactor(core): Turn the `wasm_streaming_feed` binding into ops

andreubotella:
Async WebAssembly compilation was implemented in #11200 by adding two bindings: `set_wasm_streaming_callback`, which registered a callback to be called whenever a streaming wasm compilation was started, and `wasm_streaming_feed`, which let the JS callback modify the state of the v8 wasm compiler.

`set_wasm_streaming_callback` cannot currently be implemented as anything other than a binding, but `wasm_streaming_feed` does not really need to use anything specific to bindings, and could indeed be implemented as one or more ops. This PR does that, resulting in a simplification of the relevant code.

There are three operations on the state of the v8 wasm compiler that `wasm_streaming_feed` allowed: feeding new bytes into the compiler, letting it know that there are no more bytes coming from the network, and aborting the compilation. This PR provides `op_wasm_streaming_feed` to feed new bytes into the compiler, and `op_wasm_streaming_abort` to abort the compilation. It doesn't provide an op to let v8 know that the response is finished, but closing the resource with `Deno.core.close()` will achieve that.

### DIFF
--- a/core/ops_builtin.rs
+++ b/core/ops_builtin.rs
@@ -7,6 +7,8 @@ use crate::Extension;
 use crate::OpState;
 use std::io::{stderr, stdout, Write};
 
+pub(crate) use wasm::WasmStreamingResource;
+
 pub(crate) fn init_builtins() -> Extension {
   Extension::builder()
     .js(include_js_files!(
@@ -20,6 +22,14 @@ pub(crate) fn init_builtins() -> Extension {
       ("op_try_close", op_sync(op_try_close)),
       ("op_print", op_sync(op_print)),
       ("op_resources", op_sync(op_resources)),
+      (
+        "op_wasm_streaming_feed",
+        op_sync(wasm::op_wasm_streaming_feed),
+      ),
+      (
+        "op_wasm_streaming_abort",
+        op_sync(wasm::op_wasm_streaming_abort),
+      ),
     ])
     .build()
 }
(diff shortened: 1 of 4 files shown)


andreubotella:
The goal of this PR is a simplification of the code, rather than an improvement in performance. But some benchmarks should be done either way.

core/ops_builtin.rs
```diff
@@ -81,3 +91,62 @@ pub fn op_print(
   }
   Ok(())
 }
+
+// Ops related to streaming WebAssembly
+mod wasm {
```
lucacasonato, line 96: We don't use this scoping anywhere else in the project (other than tests). Please remove the extra scope.

core/ops_builtin.rs
```diff
@@ -81,3 +87,53 @@ pub fn op_print(
...
+    state.resource_table.take::<WasmStreamingResource>(rid)?;
+
+  // At this point there are no clones of Rc<WasmStreamingResource> on the
+  // resource table, and no one should own a reference because we're never
+  // cloning them. So we can be sure `wasm_streaming` is the only reference.
+  if let Ok(wsr) = Rc::try_unwrap(wasm_streaming) {
+    wsr.0.into_inner().abort(Some(exception.v8_value));
+  } else {
+    panic!("Couldn't consume WasmStreamingResource.");
+  }
```
lucacasonato, line 136: Instead of this if else, where you panic on the else, use a `.expect()` (or does `WasmStreamingResource` not impl Debug?).

andreubotella: Yeah, `v8::WasmStreaming` doesn't implement `Debug`.

lucacasonato (approved):
LGTM

lucacasonato:
Thanks @andreubotella!

PR #11989, opened 2021-09-12
chore(cli): use updated names in `deno info --help`

### LINKED ISSUE #11987: emit mis-labelled as compiled in help test for `deno info`
`deno info --help` mentions the `compiled` and `map` fields for TypeScript files but these don't show when I run `deno info` on a TS file, either local or remote.

GJZwiers:
This should fix #11987. It renames `compiled` to `emit` and `deps` to `dependencies` in `deno info --help` to match the current outputs of the command. Also the `map` field is not mentioned as an output anymore as it is not implemented yet


lucacasonato (approved):
Thanks.

PR #11991, opened 2021-09-12
fix(ext/crypto): add HkdfParams and Pkdf2Params types

littledivy:
whoops! (again :face_with_head_bandage:)


ry (commented):
Thanks @littledivy 

Although we have all these WPT tests, having some basic TypeScript tests is still useful for things like this. Could you please add a v small test?

lucacasonato (approved):
Thanks

PR #12023, opened 2021-09-13
feat(lsp): ignore specific lint for entire file

### LINKED ISSUE #11833: lsp: quick fix for disable specified lint rule for the entire file
e.g.:
Screenshot from 2021-08-24 20-33-44

There's a missed fix: `Disable camelcase for the entire file`

satyarohith:
https://user-images.githubusercontent.com/29819102/133035885-3109a104-f93d-4271-8030-2bdbda39892c.mp4

Closes #11833

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -703,9 +703,10 @@ impl CodeActionCollection {
       })
       .unwrap();
 
-    let line_content = document.map(|d| {
-      d.source()
-        .text_info()
+    let document_source = document.map(|d| d.source());
+
+    let line_content = document_source.map(|d| {
+      d.text_info()
         .line_text(diagnostic.range.start.line as usize)
         .to_string()
     });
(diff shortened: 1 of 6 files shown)


cli/lsp/analysis.rs
```diff
@@ -748,6 +749,80 @@ impl CodeActionCollection {
...
 
+    // Disable a lint error for the entire file.
+    let parsed_source = document_source.and_then(|d| {
+      d.module().and_then(
+        |result| {
+          if let Ok(ps) = result {
+            Some(ps)
+          } else {
+            None
+          }
```
bartlomieju, line 760: Maybe `result.ok()` instead?

satyarohith: Updated. Thanks :)

cli/lsp/analysis.rs
```diff
@@ -748,6 +749,71 @@ impl CodeActionCollection {
...
+        character: 0,
+      },
+      end: lsp::Position {
+        line: 0,
+        character: 0,
+      },
+    };
+    // If ignore file comment already exists, append the lint code
+    // to the existing comment.
+    if let Some(ignore_comment) = maybe_ignore_comment {
```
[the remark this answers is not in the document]
kitsonk: I agree that is the only feedback I have on this PR as well.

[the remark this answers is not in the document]
satyarohith: Updated :)

ts
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env -S deno run
+// deno-lint-ignore-file prefer-const
+let message = 'Hello, Deno!';
+console.log(message);
```
kitsonk, line 4: small nit, the formatter doesn't catch the missing trailing new line because the file doesn't contain an extension it recognises, but we should probably put one here.

kitsonk (approved):
LGTM

ts
satyarohith: Sorry, this is a stale file used during testing. I'll remove it.

PR #12057, opened 2021-09-13
perf(runtime/fs): optimize readFile by using a single large buffer

AaronO:
This avoids allocating N buffers when reading entire files and copying when concatenating them.

## Benchmarks

```
# Before
❯ deno run -A ./cli/bench/deno_common.js
read_128k_sync:      	n = 50000, dt = 1.941s, r = 25760/s, t = 38820ns/op
read_128k:           	n = 50000, dt = 9.613s, r = 5201/s, t = 192259ns/op

# After
❯ ./target/release/deno run -A ./cli/bench/deno_common.js
read_128k_sync:      	n = 50000, dt = 1.746s, r = 28637/s, t = 34920ns/op
read_128k:           	n = 50000, dt = 8.865s, r = 5640/s, t = 177300ns/op
```

## Notes

- The improvements to `read_128k_sync` are somewhat reduced here by the extra cwd lookups caused by the `stat`, so https://github.com/denoland/deno/pull/12056 will work hand in hand with this change to improve file reads
- This operates under the assumption that files won't change (truncated or extended) whilst being read, which isn't guaranteed but seems fair IMO
- Also simplifies implementation of `readTextFile` / `readTextFileSync`

### DIFF
--- a/runtime/js/12_io.js
+++ b/runtime/js/12_io.js
@@ -191,10 +191,8 @@
     if (cursor > size) {
       // Read remaining and concat
       return concatBuffers([buf, readAllSync(r)]);
-    } else if (cursor < size) {
-      return buf.subarray(0, cursor);
     } else { // cursor == size
-      return buf;
+      return buf.subarray(0, cursor);
     }
   }
 
@@ -220,10 +218,8 @@
     if (cursor > size) {
       // Read remaining and concat
       return concatBuffers([buf, await readAllInner(r, options)]);
-    } else if (cursor < size) {
+    } else {
       return buf.subarray(0, cursor);
-    } else { // cursor == size
-      return buf;
     }
   }
 
--- a/runtime/js/12_io.js
+++ b/runtime/js/12_io.js
@@ -173,11 +173,11 @@
   }
 
   function readAllSyncSized(r, size) {
-    const buf = new Uint8Array(size);
+    const buf = new Uint8Array(size + 1); // 1B to detect extended files
     let cursor = 0;
 
     while (cursor < size) {
-      const sliceEnd = MathMin(size, cursor + READ_PER_ITER);
+      const sliceEnd = MathMin(size + 1, cursor + READ_PER_ITER);
       const slice = buf.subarray(cursor, sliceEnd);
       const read = r.readSync(slice);
       if (typeof read == "number") {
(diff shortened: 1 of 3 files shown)


runtime/js/12_io.js
```diff
@@ -183,6 +172,44 @@
...
+        cursor += read;
+      } else {
+        break;
+      }
+    }
+
+    return buf;
+  }
+
+  async function readAllInnerSized(r, size, options) {
```
bartlomieju, line 193: `readAll`/`writeAll` have been deprecated in favor or version in `deno_std`. Maybe better to add them there instead?

AaronO: This PR impacts `Deno.readFile` and `Deno.readFileSync` not `Deno.readAll`

runtime/js/40_read_file.js
```diff
@@ -4,13 +4,13 @@
...
-  const { readAllInner, readAllSync } = window.__bootstrap.io;
+  const { readAllSyncSized, readAllInnerSized } = window.__bootstrap.io;
 
   function readFileSync(path) {
     const file = openSync(path);
     try {
-      const contents = readAllSync(file);
-      return contents;
+      const { size } = file.statSync();
+      return readAllSyncSized(file, size);
```
bartlomieju, line 13: Won't this introduce an edge case if file size changes (something is written to it) once you started reading it?

AaronO: Yes, I mentioned that in PR's notes:
> This operates under the assumption that files won't change (truncated or extended) whilst being read, which isn't guaranteed but seems fair IMO

AaronO: You're asking for racy/undefined-behaviour if you're doing that, I can't think of any legitimate use case for it.

I think conceptually from the end-user's perspective `readFile`/`readFileSync` are atomic

AaronO: Files being truncated should already be handled, files being extended could be handled, either by allocating an extra byte in the buffer and seeing if that is set or by doing an extra read.

There's really no sane test for that behaviour IMO, such a test would be implicitly coupled to tokio's 16kb read per poll

lucacasonato (commented):
Not in favor of this. File reads area not atomic. We can't start optimizing at the cost of correctness.

AaronO:
> Not in favor of this. File reads area not atomic. We can't start optimizing at the cost of correctness.

My point is more that there's no sane use-case to write & read-all at the same time when you don't control the rate of reading and it's uncommon and IMO would be mostly fine as is.

But as I mentioned we can make this more robust:
> Files being truncated should already be handled, files being extended could be handled, either by allocating an extra byte in the buffer and seeing if that is set or by doing an extra read.

essentially the current implementation would be a file path for an unchanged file and then it would fallback to the slowpath if the file was extended, so that wouldn't hurt or change correctness

lucacasonato:
With fallback, SGTM

piscisaureus:
FWIW, node.js does exactly this. https://github.com/nodejs/node/blob/bbd4c6eee90ef600022c72316ba82d37a2ae16bc/lib/fs.js#L326-L349

AaronO:
I still think that it would be sane/fair to not read beyond the stat'd size, I can't imagine a sane use-case for it but I implemented the slowpath fallback anyway.

@bartlomieju @lucacasonato That should address your concerns

runtime/js/40_read_file.js
AaronO: Resolved in last commit: 852062e19f7da8dadac8fc7e2248e8bcec01fcca

runtime/js/12_io.js
```diff
@@ -183,6 +172,57 @@
...
+      } else {
+        break;
+      }
+    }
+
+    // Handle truncated or extended files during read
+    if (cursor > size) {
+      // Read remaining and concat
+      return concatBuffers([buf, readAllSync(r)]);
+    } else { // cursor == size
```
lucacasonato, line 194:
```suggestion
    } else { // cursor <= size
```

lucacasonato (approved):
LGTM

PR #12060, opened 2021-09-13
feat: add option flags to 'deno fmt'

bartlomieju:
Follow up to https://github.com/denoland/deno/pull/11944 as requested by @lucacasonato

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -13,6 +13,8 @@ use deno_runtime::permissions::PermissionsOptions;
 use log::debug;
 use log::Level;
 use std::net::SocketAddr;
+use std::num::NonZeroU32;
+use std::num::NonZeroU8;
 use std::num::NonZeroUsize;
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


cli/flags.rs
```diff
@@ -86,6 +88,11 @@ pub struct FmtFlags {
   pub files: Vec<PathBuf>,
   pub ignore: Vec<PathBuf>,
   pub ext: String,
+  pub use_tabs: Option<bool>,
+  pub line_width: Option<NonZeroU32>,
+  pub indent_width: Option<NonZeroU8>,
+  pub single_quote: Option<bool>,
+  pub prose_wrap: Option<String>,
```
bartlomieju, line 95: These `Option<bool>` might be an overkill, but I find it easier to reason about when we need to overwrite `FmtConfigOptions`

lucacasonato (approved):
Thanks Bartek

dsherret (approved):
LGTM

Issue #12061, opened 2021-09-13
Add JSON Schema for config file to `cli/schemas`


kitsonk:

bartlomieju:
I thought with discussion in https://github.com/denoland/deno/issues/11885 we were not going to add JSON schema for config file and instead do an "associated file" in VSCode? I still plan to add examples in manuals as part of https://github.com/denoland/manual/issues/75

kitsonk:
Correct, as we have to associate a schema to a file! 😉  I've written a few schemas recently, I would be glad to do it.

bartlomieju:
> Correct, as we have to associate a schema to a file! 😉 I've written a few schemas recently, I would be glad to do it.

I'm already on it, I'll let you know if I hit any roadblocks.

kitsonk:
Duplicate of #11885

Issue #11885, opened 2021-08-31
Add JSON schema file for configuration file


bartlomieju:
Currently Deno supports configuration file for altering TypeScript compiler settings. With #11686 and #11857 we'll have more configuration options and config file will be usable with `deno lint` and `deno fmt`.

The idea is to make the configuration file strict - ie. it will not accept "unknown" fields (it will fail the parsing step if there are extraneous fields). This is to avoid the problem similar to `package.json` where random tools piggy-back on this file and add tens (or hundreds) of fields specific to each tool. 

To alleviate the problem we should introduce a [JSON schema file](http://json-schema.org/). Most IDEs already support this format and it's used widely in the wild, eg. here's [TypeScript JSON schema](http://json.schemastore.org/tsconfig) and here's [dprint JSON schema](https://github.com/dprint/dprint/blob/main/website/assets/schemas/v0.json). Additional advantage is that these schema files are versioned, which should help with making sure Deno's configuration file does not introduce breaking changes.

dsherret:
**`$schema` Property is VSCode Specific**

For associating a config file something that is done is a [`$schema` property](https://code.visualstudio.com/docs/languages/json#_mapping-in-the-json) that specifies a url:

https://github.com/denoland/deno/blob/b518f5e1ba4c3281d4aee4c637f410db382ce751/.dprint.json#L2

However, I would recommend against this because:

1. It is actually a VSCode specific property.
2. It is something the user needs to manually maintain.

I am going to be moving away from this in dprint soon towards something better.

**Better Solution?**

Instead I would recommend we come up with a solution where the deno binary can provide either the entire schema text or a url to the schema (for example, we could have the schema as a release artifact). Probably providing a url would be best as it will most likely be portable. For example, perhaps the deno language server could provide the JSON schema url at startup then it would be up to the editors to associate that schema with `Deno.json`.

For vscode specifically, we can provide the schema programatically in vscode-deno based on the version of `deno` the user is using with the extension (was just coincidentally told about this today while doing research for dprint https://github.com/microsoft/vscode/issues/127812).

Maybe there is a better way of doing all this with the LSP to do this though so all editors get the schema associated out of the box? I don't see a way of doing it, but I'm not super familiar with the LSP. cc @kitsonk

bartlomieju:
Thanks for clearing that up David, I could have sworn I used JSON schema in JetBrains IDEs as well.

I'm curious to see Kitson's opinion before we decide either way.

[...]

PR #12062, opened 2021-09-13
chore: add JSON schema for configuration file

bartlomieju:
Closes https://github.com/denoland/deno/issues/11885
Closes https://github.com/denoland/deno/issues/12061

@kitsonk @dsherret please advise how should I handle `compilerOptions` key; a) copy-paste from http://json.schemastore.org/tsconfig b) try to link to http://json.schemastore.org/tsconfig somehow?


cli/schemas/config-file.json
```diff
@@ -0,0 +1,143 @@
+{
+  "$id": "https://deno.land/schemas/config-file.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "description": "A JSON representation of a Deno configuration file.",
+  "required": [],
+  "title": "Deno configuration file Schema",
+  "type": "object",
+  "properties": {
+    "compilerOptions": {
```
dsherret, line 9: I think this is the syntax, but I'm not sure (something along these lines):

```json
"compilerOptions": {
  "$ref": "http://json.schemastore.org/tsconfig#/definitions/compilerOptionsDefinition/properties/compilerOptions"
}
```

That said, perhaps we should limit this to only the compiler options we support and I believe that would require copy and paste or at least using a "$ref" like this a bunch of times for each property we support.

cli/schemas/config-file.json
```diff
@@ -0,0 +1,143 @@
...
+              "default": 80
+            },
+            "indentWidth": {
+              "description": "The number of characters for an indent.",
+              "type": "number",
+              "default": 2
+            },
+            "singleQuote": {
+              "type": "boolean",
+              "description": "Whether to use tabs (true) or spaces (false) for quotation.",
```
dsherret, line 515: Needs updating 🙂

cli/schemas/config-file.json
```diff
@@ -0,0 +1,143 @@
...
+              "default": false
+            },
+            "proseWrap": {
+              "type": "string",
+              "description": "Define how prose should be wrapped.",
```
dsherret, line 520: "...should be wrapped in markdown files."?

cli/schemas/config-file.json
```diff
@@ -0,0 +1,143 @@
...
+              "default": "always",
+              "oneOf": [
+                {
+                  "const": "always",
+                  "description": "For each file, uses the newline kind found at the end of the last line."
```
dsherret, line 525: The descriptions in here all need updating.

cli/schemas/config-file.json
```diff
@@ -0,0 +1,542 @@
+{
+  "$id": "https://deno.land/schemas/config-file.json",
```
dsherret, line 2: Maybe we should version this url?

cli/schemas/config-file.json
bartlomieju: For now I went with copy-paste from http://json.schemastore.org/tsconfig based on the list here: https://deno.land/manual@v1.13.2/typescript/configuration#how-deno-uses-a-configuration-file it still needs updating defaults though (and `lib` property needs updating of list)

cli/schemas/config-file.json
bartlomieju: How do you propose to do that?

dsherret: I guess maybe it doesn't matter actually if we're going to be providing this file text to the editor plugins via the LSP (not sure how that's going to work atm).

dsherret (approved):
LGTM

cli/schemas/config-file.json
```diff
@@ -0,0 +1,362 @@
...
+              "ES2015.Collection",
+              "ES2015.Core",
+              "ES2015.Generator",
+              "ES2015.Iterable",
+              "ES2015.Promise",
+              "ES2015.Proxy",
+              "ES2015.Reflect",
+              "ES2015.Symbol.WellKnown",
+              "ES2015.Symbol",
+              "ES2016",
```
bartlomieju, line 97: I'm not sure if this is correct...

cli/schemas/config-file.json
kitsonk: I think we should version it at least, it is as simple as adding a version to the file.  I suggest we use something like `config-file_v1.json` and simply increment the number when we make changes, versus a full blown semver scheme.  Then in vscode you associate a schema with a file type, and we will just bump it when logical.  The other option is we just let git/GitHub manage it, with the raw github with a commitish path be the id we use in vscode to "pin" it.

cli/schemas/config-file.json
```diff
@@ -0,0 +1,366 @@
...
+          "markdownDescription": "Make keyof only return strings instead of string, numbers or symbols. Legacy option.\n\nSee more: https://www.typescriptlang.org/tsconfig#keyofStringsOnly"
+        },
+        "lib": {
+          "description": "Specify a set of bundled library declaration files that describe the target runtime environment.",
+          "type": "array",
+          "uniqueItems": true,
+          "default": "deno.window",
+          "items": {
+            "type": "string",
+            "anyOf": [
```
kitsonk, line 79: I would just drop this and keep it as a string... it is actually really annoying the behaviour in vscode, because tsc issues diagnostics for this and we now mute them, but the schema still has them, so people get "warnings" in vscode that we can't get rid of because it doesn't "match" the schema.  We will end up issuing diagnostics for "bad" or "incorrect" values in config options, so feel free to eliminate any complex "soft" values like this that would be better served by LSP diagnostics, and focus on "structural" validation of the config file.

cli/schemas/config-file.json
kitsonk: yeah, I don't think we want to reference the external schema, as it doesn't meet our needs.

kitsonk (approved):
LGTM

Issue #11759, opened 2021-08-18
Deno.startTls 


aricart:
`Deno.startTls` seems to have gone away from the docs for some reason (not in unstable or stable docs). Hopefully, this is an omission, not a removal, as TLS was originally designed to be an upgrade process rather than a straight connect process.

Previously, if there was a bad certificate (as in the CA was not specified etc), the connection would fail on the `Deno.startTls` API call, resolving to error `InvalidData` which was not very informative, but at least could be associated with the process of upgrading to the TLS connection.

On newer versions (1.13.1) it fails on a socket flush. Definitely not expecting it to fail after the call to startTls has returned indicating that the socket is ready for work. The message `Sending fatal alert BadCertificate` is logged on the console, but this is not useful. Instead, I would expect some reasonable error to be raised on Deno.startTls, and that I can handle/propagate properly.

```
error: Uncaught (in promise) UnexpectedEof: tls handshake eof
    nwritten += await w.write(arr.subarray(nwritten));
                ^
    at deno:core/01_core.js:106:46
    at unwrapOpResult (deno:core/01_core.js:126:13)
    at async write (deno:ext/net/01_net.js:26:12)
    at async writeAll (https://deno.land/std@0.95.0/io/util.ts:79:17)
    at async BufWriter.flush (https://deno.land/std@0.95.0/io/bufio.ts:470:7)
    at async write (file:///Users/synadia/Dropbox/code/src/github.com/nats-io/nats.deno/src/deno_transport.ts:45:3)
```

bartlomieju:
This behavior was changed on purpose, see https://github.com/denoland/deno/issues/9123

aricart:
@bartlomieju thanks for the quick turnaround. How am I supposed to handle the condition - IE, if the transport (whatever kind of connection it is) according to the API `connected` but really it didn't. It will fail sometime in the future on a write... Any way of testing the Conn returned to check if it is closed?

bartlomieju:
@aricart I'm afraid we don't have an API that would check if connection is closed. I know that this change was done on purpose but I don't have full context. CCing @piscisaureus and @lucacasonato

lucacasonato:
@aricart It fails on `read` or `write` or flush (as part of write). You might be able to detect it with a 0 byte read. The problem is that if you detect it as connected, and only if it is connected to something, there is racyness. The connection might close between checking if it is connected, and reading / writing to it. You should just try to read/write and handle errors accordingly.

aricart:
[...]

PR #12066, opened 2021-09-14
fix(ext/net): make Deno.startTls() throw on handshake failure

piscisaureus:


bartlomieju:
Fixes https://github.com/denoland/deno/issues/11759 ?

ext/net/ops_tls.rs
```diff
@@ -728,9 +728,11 @@ where
     ca_data,
     unsafely_ignore_certificate_errors,
   )?);
-  let tls_stream =
+  let mut tls_stream =
     TlsStream::new_client_side(tcp_stream, &tls_config, hostname_dns);
 
+  tls_stream.handshake().await?;
+
```
bartlomieju, line 735: Very elegant solution

bartlomieju (approved):

piscisaureus:
I'm actually not super thrilled about this change. One issue is that `startTls()` now consumes the `conn` parameter sometimes on failure and sometimes it doesn't, depending on where the failure happens. Another is that it creates different semantics in `connect()`/`accept()` on one hand and `startTls()` on the other hand.

As an alternative we could expose the `handshake()` method on TLS streams (`negotiate()` might be a better name).

piscisaureus:
> As an alternative we could expose the `handshake()` method on TLS streams

Turns out, this is exactly what Go does: https://pkg.go.dev/crypto/tls#Conn.Handshake.

piscisaureus:
Superseded by #12467.

Issue #12025, opened 2021-09-13
Stabilise `Deno.TestDefinition.permissions`


bartlomieju:
We've been using this feature for many months - it's a basis for the whole unit test suite in CLI. I think it's high time to stabilise this feature.

caspervonb:
Worker permissions should share the same interface and be stabilised together with this @bartlomieju.

PR #11981, opened 2021-09-10
BREAKING(unstable): Enable Deno namespace in workers by default

nayeemrmn:
Closes #8174.

PR #12078, opened 2021-09-14
feat: Stabilize Deno.TestDefinition.permissions

bartlomieju:
Closes https://github.com/denoland/deno/issues/12025


dsherret (changes requested):
I think we need to update `op_pledge_test_permissions` to not do an unstable check?

cli/dts/lib.deno.ns.d.ts
```diff
@@ -131,6 +131,145 @@ declare namespace Deno {
...
+      read?: "inherit" | boolean | Array<string | URL>;
+
+      /** Specifies if the `run` permission should be requested or revoked.
+       * If set to `"inherit"`, the current `run` permission will be inherited.
+       * If set to `true`, the global `run` permission will be requested.
+       * If set to `false`, the global `run` permission will be revoked.
+       *
+       * Defaults to "inherit".
+       */
+      run?: "inherit" | boolean | Array<string | URL>;
```
dsherret, line 260: I was looking at `parsePermissions` in 11_workers.js and `run` seems like it does not support an array even though it says it does here. The documentation comment also doesn't describe what an array does.

I'm a little concerned about this API not being well tested considering there are these major issues with it (like the other one I discovered and fixed). Perhaps we should work on improving the test coverage for this feature and worker permissions (if anything, at least we could open an issue).

nayeemrmn: Looks like I accidentally fixed this in #11981. I'll extract the fix and cleanups to a separate PR.

bartlomieju:
Blocked by #12297

dsherret (approved):
LGTM

PR #12079, opened 2021-09-14
chore(scripts): improvements to the release scripts based on the 1.14 release

dsherret:
Improvements based on doing the 1.14 release.

### DIFF
--- a/tools/release/02_publish_dependency_crates.ts
+++ b/tools/release/02_publish_dependency_crates.ts
@@ -4,8 +4,14 @@ import { DenoWorkspace, getCratesPublishOrder } from "./helpers/mod.ts";
 
 const workspace = await DenoWorkspace.load();
 
-const dependencyCrates = workspace.getDependencyCrates();
+const dependencyCrates = getCratesPublishOrder(workspace.getDependencyCrates());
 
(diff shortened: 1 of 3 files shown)


tools/release/02_publish_dependency_crates.ts
```diff
@@ -4,8 +4,14 @@ import { DenoWorkspace, getCratesPublishOrder } from "./helpers/mod.ts";
...
-for (const crate of getCratesPublishOrder(dependencyCrates)) {
-  await crate.publish();
+try {
+  for (const [i, crate] of dependencyCrates.entries()) {
+    await crate.publish();
+    console.log(`Published ${i + 1} of ${dependencyCrates.length} crates.`);
+  }
+} finally {
+  // system beep to notify error or completion
+  console.log("\x07");
```
dsherret, line 16: Does this work on mac and linux?

bartlomieju: It never worked for me on macOS, I believe it has to do with some accessibility settings in the OS. That said it would be silently ignored, so I think it's fine.

bartlomieju (approved):

PR #12081, opened 2021-09-15
fix(lsp): align filter text to vscode logic

### LINKED ISSUE #11861: lsp: suggestion should include "#privateVariable" when typing it without "#" symbol
I'm working on a Deno project and every time I try to access a private variable previously defined with a "#" prefix I always have to type the symbol first.

IMO the extension should suggest "#variable" when trying to type "variable"

kitsonk:
Fixes #11861

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -1695,11 +1695,11 @@ impl CompletionEntry {
         return None;
       }
       if insert_text.starts_with('[') {
-        let value = BRACKET_ACCESSOR_RE
-          .replace(insert_text, |caps: &Captures| format!(".{}", &caps[1]))
-          .to_string();
-        log::info!("value: {}", value);
-        return Some(value);
+        return Some(
+          BRACKET_ACCESSOR_RE
+            .replace(insert_text, |caps: &Captures| format!(".{}", &caps[1]))
+            .to_string(),
+        );
       }
     }
 
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -53,6 +53,17 @@ use text_size::{TextRange, TextSize};
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 
+lazy_static::lazy_static! {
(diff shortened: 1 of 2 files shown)


cli/lsp/tsc.rs
```diff
@@ -1678,20 +1678,28 @@ impl CompletionEntry {
...
   fn get_filter_text(&self) -> Option<String> {
-    // TODO(@kitsonk) this is actually quite a bit more complex.
-    // See `MyCompletionItem.getFilterText` in vscode completion.ts.
-    if self.name.starts_with('#') && self.insert_text.is_none() {
-      return Some(self.name.clone());
+    if self.name.starts_with('#') {
+      if let Some(insert_text) = &self.insert_text {
+        if insert_text.starts_with("this.#") {
+          return Some(insert_text.replace("this.#", ""));
+        } else {
```
dsherret, line 1685: Do we need to remove `this.` in the case where it only starts with `this.`?

dsherret (approved):
LGTM. Just one question.

cli/lsp/tsc.rs
```diff
@@ -1678,20 +1678,28 @@ impl CompletionEntry {
...
         return None;
       }
       if insert_text.starts_with('[') {
-        let re = Regex::new(r#"^\[['"](.+)['"]\]$"#).unwrap();
-        let insert_text = re.replace(insert_text, ".$1").to_string();
-        return Some(insert_text);
+        let value = BRACKET_ACCESSOR_RE
+          .replace(insert_text, |caps: &Captures| format!(".{}", &caps[1]))
+          .to_string();
+        log::info!("value: {}", value);
```
lucacasonato, line 1701: stray log

cli/lsp/tsc.rs
kitsonk: It is a replication of the vscode logic, which does not do this. And it only sets the filter text. When the replacement text is `this.#` there is then effectively empty filter text, which is then nothing.

PR #12084, opened 2021-09-15
fix(runtime/ops/signal.rs): Add FreeBSD signal definition.

MikaelUrankar:


runtime/ops/signal.rs
```diff
@@ -57,6 +57,44 @@ impl Resource for SignalStreamResource {
...
+    "SIGIO" => Ok(23),
+    "SIGXCPU" => Ok(24),
+    "SIGXFSZ" => Ok(25),
+    "SIGVTALRM" => Ok(26),
+    "SIGPROF" => Ok(27),
+    "SIGWINCH" => Ok(28),
+    "SIGINFO" => Ok(29),
+    "SIGUSR1" => Ok(30),
+    "SIGUSR2" => Ok(31),
+    _ => Err(type_error(format!("Invalid signal : {}", s))),
```
lucacasonato, line 94: According to https://www.freebsd.org/cgi/man.cgi?sektion=3&query=signal there are also `SIGTHR` and `SIGLIBRT`. Do you want to add those too?

lucacasonato (commented):
Thanks. We don't officially support FreeBSD, but I am happy to merge this if this unblocks your build :-)

runtime/ops/signal.rs
MikaelUrankar: I've added the missing signal. Thanks.

lucacasonato (approved):
Thanks @MikaelUrankar

PR #12091, opened 2021-09-15
chore(tests): windows pty tests

### LINKED ISSUE #12069: Add windows PTY tests
I'm already working on this.

* Investigate using a common as possible API for both linux and windows
* Add tests for the code that merged in #11969.

dsherret:
This PR runs the PTY tests on all platforms. Unfortunately it does not work on GitHub Actions for reasons outlined in [this comment](https://github.com/denoland/deno/pull/12091#discussion_r710190415) so these tests only run on Windows development machines. This is still an improvement over not running these tests at all and now the Windows permission prompt stuffing is more properly tested. Hopefully in the future these will just start running on GitHub Actions without us having to do anything (or maybe there is a solution today?)

Closes #12069.


test_util/Cargo.toml
```diff
@@ -14,6 +14,7 @@ path = "src/test_server.rs"
 [dependencies]
 anyhow = "1.0.43"
 async-stream = "0.3.2"
+atty = "0.2.14"
```
dsherret, line 17: This depenency is also used in the CLI for checking if stdin/stdout is a tty

test_util/src/pty.rs
```diff
@@ -0,0 +1,437 @@
...
+          let console_handle = WinHandle::new(console_handle);
+          move || {
+            WaitForSingleObject(thread_handle.as_raw(), INFINITE);
+            // wait for the reading thread to catch up
+            std::thread::sleep(Duration::from_millis(200));
+            // close the console handle which will close the
+            // stdout pipe for the reader
+            ClosePseudoConsole(console_handle.take());
+          }
+        });
```
dsherret, line 226: The winapi examples create a separate thread to do the reading: https://github.com/microsoft/terminal/blob/fcb42b262f0bab726184edccc42e9632dcfba452/samples/ConPTY/EchoCon/EchoCon/EchoCon.cpp#L37 and the documentation [on ending a session](https://docs.microsoft.com/en-us/windows/console/creating-a-pseudoconsole-session#ending-the-pseudoconsole-session).

I didn't want to do this though because it would complicate the `Pty` trait. Instead, this code creates a thread that waits for the pseudoconsole process to exit and when it does it closes the stdout pipe which will unblock the reader.

test_util/src/pty.rs
```diff
@@ -0,0 +1,437 @@
...
+    program,
+    args,
+    &cwd.as_ref().to_string_lossy().to_string(),
+    env_vars,
+  );
+  Box::new(pty)
+}
+
+#[cfg(target_os = "windows")]
+mod windows {
```
dsherret, line 92: After I wrote most of this implementation, Ryan pointed me in the direction of using [portable_pty](https://docs.rs/portable-pty/0.4.0/portable_pty/). For some reason, it did not build for me out of the box and I did not see a way for it to avoid deadlocking on exit without changing the Pty trait to have the caller need to manage creating a new thread to read the output (see below how I inverted this so the thread is hidden in the implementation), so I just kept what I have here. I think having this own implementation is better now that the work is done, but had I done it over again I probably would have done a way to reduce the effort here. That said, we'll be able to easily and more quickly improve this now that it's here.

test_util/src/pty.rs
```diff
@@ -0,0 +1,437 @@
...
+        }
+      }
+    }
+  }
+
+  impl Read for WinPseudoConsole {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+      unsafe {
+        let mut bytes_read = 0;
+        // don't bother checking the result of this call as it will be false
```
piscisaureus, line 243: What if the writer does a 0-byte write?

test_util/src/pty.rs
```diff
@@ -0,0 +1,437 @@
...
+        WinHandle::new(duplicate_handle)
+      }
+    }
+
+    pub fn as_raw(&self) -> HANDLE {
+      self.inner
+    }
+
+    pub fn take(mut self) -> HANDLE {
+      self.close_on_drop = false;
```
piscisaureus, line 323: Given that the only purpose of `close_on_drop` is to make `take()` work, I'd suggest this solution instead:

```rust
let handle = self.inner;
std::mem::forget(self);
handle
```

test_util/src/pty.rs
```diff
@@ -0,0 +1,437 @@
...
+        Ok(bytes_written as usize)
+      }
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+      Ok(())
+    }
+  }
+
+  struct WinHandle {
```
piscisaureus, line 287: In case you ever need to expand WinHandle functionality: try to follow the rust_std `OwnedHandle` implementation (which is unstable, but otherwise good). https://doc.rust-lang.org/beta/std/os/windows/io/struct.OwnedHandle.html

test_util/src/pty.rs
```diff
@@ -0,0 +1,437 @@
...
+  fn to_windows_str(str: &str) -> Vec<u16> {
+    use std::os::windows::prelude::OsStrExt;
+    std::ffi::OsStr::new(str)
+      .encode_wide()
+      .chain(Some(0))
+      .collect()
+  }
+
+  fn get_env_vars(env_vars: HashMap<String, String>) -> Vec<u16> {
+    // each environment variable is in the form `name=value\0`
```
piscisaureus, line 424: I believe there is another requirement, which is that the env vars are sorted alphabetically (by key).

piscisaureus: cf. https://docs.microsoft.com/en-us/windows/win32/procthread/changing-environment-variables

test_util/src/lib.rs
```diff
@@ -1589,62 +1587,97 @@ pub enum PtyData {
...
+    let text = strip_ansi_codes(&text);
+    // get rid of any text that is overwritten with only a carriage return
+    let text = CARRIAGE_RETURN_RE.replace_all(&text, "$1");
+    // finally, trim surrounding whitespace
+    text.trim().to_string()
   }
 }
 
+pub fn with_pty(deno_args: &[&str], mut action: impl FnMut(Box<dyn pty::Pty>)) {
+  if !atty::is(atty::Stream::Stdin) || !atty::is(atty::Stream::Stderr) {
```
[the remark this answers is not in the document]
piscisaureus: The issue might be that no console is allocated for the process (tree) our tests run in. 
It might be as simple as calling `AllocConsole()` to ensure the process has one. https://docs.microsoft.com/en-us/windows/console/allocconsole

test_util/src/pty.rs
dsherret: Thanks! Much better. I had a brief moment of trying to remember what this function was called, but I... forgot.

test_util/src/pty.rs
dsherret: They really do work hard to make things fun. Thanks! I will do this.

test_util/src/pty.rs
dsherret: Then any reads like `read_to_end` will stop reading (Seems to be the behaviour of `std::io::Read.read_to_end`). Perhaps it should loop when `0` and this returns `true`?

piscisaureus: > Perhaps it should loop when `0` and this returns `true`?
👆

test_util/src/lib.rs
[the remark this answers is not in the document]
dsherret: Good idea, but unfortunately it did not work :(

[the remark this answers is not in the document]
piscisaureus: OK, thanks for trying!

piscisaureus (approved):

PR #12116, opened 2021-09-17
fix(cli): don't ignore diagnostics about for await

### LINKED ISSUE #12115: Diagnostics are ignored for `for await` loops
**Describe the bug**
Using the `await` reserved word inside a normal function (without async) is not being warned by the IDE.

**To Reproduce**

1. Paste the following code:

``` ts
const listener = Deno.listen({ port: 8080 });

  handleConn(conn);
}

function handleConn(conn: Deno.Conn) {
  for await (const event of httpConn) {
    event.respondWith(new Response("html", { status: 200 }));
  }
}
```

2. Note there is no warning in using `await` in a normal function

**Expected behavior**

Show an error in the editor saying `await` is only permitted inside async functions

**Screenshots**

Running the code with `deno run --allow-net mod.ts` fails:

image

**Versions**

vscode: 1.60.1 deno: 1.14.0 extension: 3.9.1

kitsonk:
Fixes #12115

### DIFF
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -622,6 +622,12 @@ itest!(error_026_remote_import_error {
   http_server: true,
 });
 
+itest!(error_for_await {
+  args: "run --reload error_for_await.ts",
+  output: "error_for_await.ts.out",
+  exit_code: 1,
+});
+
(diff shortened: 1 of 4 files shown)


satyarohith (approved):
LGTM.

PR #12123, opened 2021-09-17
feat(ext/crypto): implement AES-CTR & AES-CBC encryption

littledivy:
Towards #11690

### DIFF
--- a/ext/crypto/00_crypto.js
+++ b/ext/crypto/00_crypto.js
@@ -116,6 +116,8 @@
     },
     "encrypt": {
       "RSA-OAEP": "RsaOaepParams",
+      "AES-CTR": "AesCtrParams",
+      "AES-CBC": "AesCbcParams",
     },
     "decrypt": {
       "RSA-OAEP": "RsaOaepParams",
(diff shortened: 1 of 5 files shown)


ext/crypto/00_crypto.js
```diff
@@ -440,6 +442,41 @@
...
+            );
+          }
+          normalizedAlgorithm.iv = TypedArrayPrototypeSlice(
+            normalizedAlgorithm.iv,
+          );
+
+          // 1.
+          if (normalizedAlgorithm.iv.byteLength !== 16) {
+            throw new DOMException(
+              "Counter must be 16 bytes",
```
bnoordhuis, line 464: s/Counter/Initialization vector/? "Counter" is something I associate with CTR mode, not CBC.

(I checked WPT but it's not an expected error message, right?)

ext/crypto/lib.rs
```diff
@@ -789,8 +790,32 @@ pub async fn op_crypto_derive_bits(
...
+  // AES-CBC
+  iv: Option<ZeroCopyBuf>,
+  length: Option<usize>,
+}
+
+macro_rules! cipher_cbc {
+  ($length: ty, $key: expr, $iv: expr, $data: expr) => {{
+    // Section 10.3 Step 2 of RFC 2315 https://www.rfc-editor.org/rfc/rfc2315
+    type AesCbc = block_modes::Cbc<$length, block_modes::block_padding::Pkcs7>;
+    let cipher = AesCbc::new_from_slices($key, $iv)?;
```
bnoordhuis, line 805: Macros with control flow in them (the ? operator) are a special kind of hell to debug. Please either use combinators (and_then or map) or turn them into regular functions.

ext/crypto/lib.rs
```diff
@@ -1322,6 +1368,39 @@ pub async fn op_crypto_decrypt_key(
...
+      let padded_plaintext = match length {
+        128 => decipher_cbc!(aes::Aes128, key, &iv, data),
+        192 => decipher_cbc!(aes::Aes192, key, &iv, data),
+        256 => decipher_cbc!(aes::Aes256, key, &iv, data),
+        _ => unreachable!(),
+      };
+
+      // 3.
+      // padded_plaintext.len() > 0 so this should never panic.
+      let p = *padded_plaintext.last().unwrap() as usize;
```
bnoordhuis, line 1390: It'd be helpful to explain here that the plaintext is PKCS#7-padded and that the last byte is the size of the padding (and that it's never zero) because I don't think that's very obvious from reading the code itself, only if you know PKCS#7 is used here and you know _what_ PKCS#7 is.

ext/crypto/lib.rs
```diff
@@ -1322,6 +1368,39 @@ pub async fn op_crypto_decrypt_key(
...
+
+      // 4-5.
+      let plaintext = &padded_plaintext[p..];
```
bnoordhuis, line 1393: Are you sure this is correct? I would've expected something like this:
```suggestion
      let n = padded_plaintext.len() - p;
      let plaintext = &padded_plaintext[..n];
```

ext/crypto/lib.rs
```diff
@@ -1322,6 +1368,39 @@ pub async fn op_crypto_decrypt_key(
...
+      if p == 0 || p > 16 || *plaintext != *vec![p as u8; p] {
+        return Err(custom_error(
+          "DOMExceptionOperationError",
+          "Invalid padding".to_string(),
+        ));
+      }
+
+      // 6.
+      Ok(plaintext.to_vec().into())
```
bnoordhuis, line 1402: Efficiency: rather than creating a new vec from a slice-of-vec, you can probably just shrink the existing vec.

bnoordhuis (commented):
There are accompanying WPT tests, I assume? Can you turn those on?

littledivy:
@bnoordhuis WPT tests depend on importKey

ext/crypto/00_crypto.js
littledivy: Ah fixed. I don't remember why I used Counter there :sweat_smile:

ext/crypto/lib.rs
littledivy: Made them inline it's not much duplicate code anyways.

ext/crypto/lib.rs
littledivy: Fixed

PR #12125, opened 2021-09-17
feat(ext/crypto): implement wrapKey

littledivy:
Towards #11690 

Currently only RSA-OAEP as wrapping algorithm is supported. AES-* will start working when we have proper `encrypt` support.


ext/crypto/00_crypto.js
```diff
@@ -1524,6 +1528,98 @@
...
+      // 12.
+      const exportedKey = await this.exportKey(format, key);
+
+      let bytes;
+      // 13.
+      if (format !== "jwk") {
+        bytes = exportedKey;
+      } else {
+        // TODO(@littledivy): Implement JWK.
+        throw new TypeError("not implemented.");
```
bnoordhuis, line 1602: It looks like the prevailing idiom in this file is to `throw new DOMException("Not implemented", "NotSupportedError");` so perhaps stick with that, here and below.

bnoordhuis (approved):

Issue #11872, opened 2021-08-30
deno test type-checking some unknown <GUID>$deno$test.js file when a test Deno.run's deno


AlexanderOMara:
This is kinda a weird issue I've noticed, where if a Deno test uses `Deno.run` to run some shared code in a deno subprocess, it will print some output that it type-checked a file with a random GUI followed by `$deno$test.js`.

```
$ deno test --allow-run src
running 1 test from file:///.../src/alpha.test.ts
test alpha ... ok (8ms)
running 1 test from file:///.../src/exec.test.ts
test exec ... ok (41ms)
Check file:///.../76c79343-ee67-435f-a02e-6c0bc0619fb5$deno$test.js
running 1 test from file:///.../src/zeta.test.ts
test zeta ... ok (8ms)

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (561ms)
```

First the `alpha` test ran normally, then the `exec` test ran using `data.ts` in a subprocess, then the `zeta` test apparently triggers the check line.

These are the files I used: 

`src/data.ts`

```ts
export const DATA = "test";
```

`src/alpha.test.ts`

```ts
import { DATA } from "./data.ts";

Deno.test("alpha", () => {
  void DATA;
});
```

`src/exec.test.ts`

```ts
Deno.test("exec", async () => {
  const p = Deno.run({
    cmd: [
      "deno",
      "run",
      "--no-check",
      "src/data.ts",
    ],
  });
  try {
    await p.status();
  } finally {
    p.close();
  }
});
```

`src/zeta.test.ts`

```ts
import { DATA } from "./data.ts";

Deno.test("zeta", () => {
  void DATA;
});
```

bartlomieju:
I believe this is expected - test runner creates a temporary file of all discovered test modules. @caspervonb can you take a look?

caspervonb:
Yeah this is why I've been asking if you wanna do `execute_main_module` as that unblocks the cleanup for this :)

PR #12128, opened 2021-09-17
fix(core): prevent multiple main module loading

bartlomieju:
Fixes https://github.com/denoland/deno/issues/10803

PR #12129, opened 2021-09-18
refactor(cli): don't generate a module for side loading tests

caspervonb:
As of https://github.com/denoland/deno/pull/12128 I no longer need to generate a module to side-load test modules so this removes the module generation entirely.

Closes https://github.com/denoland/deno/issues/11872

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -47,7 +47,6 @@ use std::sync::mpsc::Sender;
 use std::sync::Arc;
 use std::time::Duration;
 use std::time::Instant;
-use uuid::Uuid;
 
 /// The test mode is used to determine how a specifier is to be tested.
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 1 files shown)


cli/tools/test.rs
```diff
@@ -306,7 +286,12 @@ async fn test_specifier(
     None
   };
 
-  worker.execute_main_module(&test_specifier).await?;
+  // We only execute the specifier as a module if it is tagged with TestMode::Module or
+  // TestMode::Both.
```
bartlomieju, line 290:
```suggestion
  // We only execute the specifier as a main module if it is tagged with TestMode::Module or
  // TestMode::Both.
```

caspervonb: Side module actually.

bartlomieju: Aha, okay, didn't undestand the comment then.

bartlomieju (commented):
Can we get a test case for this PR?

caspervonb:
> Can we get a test case for this PR?

Would rarely argue for no coverage but In this case; is no an option? 😁 
We're just removing an old crusty wart and not really fixing anything _broken_ per-se.

It can surface on main but don't actually know how to get the diagnostic to print on main, plus it's not really a test I want to keep around since its testing for an old bandaid that we removed.

bartlomieju (approved):

PR #12137, opened 2021-09-19
experiment(cli): expose deno cli as a lib to enable embedding

AaronO:
This is an initial simplistic pass to start enabling full CLI embedding for projects like "Deno Desktop". These "CLI embedders" will want to reuse the `CliModuleLoader`/tsc and the entirety of `Deno.*` as provided by `deno`, they are functional supersets that would essentially add their own `JsRuntime` extensions (and associated snapshots) and possibly tweak some CLI commands.

## Notes

As an unintended side-effect (haven't identified root cause) this appears to reduce a local release bin build (`cargo build --release --bin deno`) from ~4m30 to ~3m (when rebuilding app repo crates following a `deno_core` change/invalidation)


AaronO:
Off the table for now. For "Deno Desktop" we'll extend `deno_runtime` and use the `deno` command to assist with dev functionality

PR #12138, opened 2021-09-19
fix(cli): move Deno.flock and Deno.funlock to unstable types

caspervonb:
`Deno.lock` and `Deno.unlock`(and sync variants) are marked as unstable, but are in the stable type definitions.
This moves them into the unstable type definitions.


bartlomieju (approved):
LGTM, thanks for fixing this

Issue #12107, opened 2021-09-16
Remove "standalone" resource method from `Deno` namespace


kitsonk:
Methods like `Deno.read()` and `Deno.write()` are effectively redundant as there are other better objects to manage the resources (like `Deno.File`) and having to deal explicitly with the _rid_ which is just a number can be confusing and error prone to users.

caspervonb:
> Methods like `Deno.read()` and `Deno.write()` are effectively redundant as there are other better objects to manage the resources (like `Deno.File`) and having to deal explicitly with the _rid_ which is just a number can be confusing and error prone to users.

I'm in favour but note that that this will completely break std/wasi, that's okay but I'll have to rewrite it.

bartlomieju:
`Deno.seek()` and `Deno.seekSync()` should be removed as well.

bartlomieju:
More methods that should be moved to `Deno.File` class:
- `Deno.fsyncSync`
- `Deno.fsync`
- `Deno.fdatasyncSync`
- `Data.fdatasync`
- `Deno.flock`
- `Deno.flockSync`
- `Deno.funlock`
- `Deno.funlockSync`

CC @caspervonb

caspervonb:
> More methods that should be moved to `Deno.File` class:
> 
> * `Deno.fsyncSync`
> * `Deno.fsync`
> * `Deno.fdatasyncSync`
> * `Data.fdatasync`
> * `Deno.flock`
> * `Deno.flockSync`
> * `Deno.funlock`
> * `Deno.funlockSync`
> 
> CC @caspervonb

SGTM, patches coming right up.

wojpawlik:
Should `Deno.close` be deprecated in favor of `Deno.File::close`?

bartlomieju:
> Should `Deno.close` be deprecated in favor of `Deno.File::close`?

It's still debated, but most likely it won't be deprecated.

caspervonb:
> > Should `Deno.close` be deprecated in favor of `Deno.File::close`?
> 
> It's still debated, but most likely it won't be deprecated.

I think we should, in favour of using explicit resource management (https://github.com/tc39/proposal-explicit-resource-management stage 2) and finalisers.

bartlomieju:
> > > Should `Deno.close` be deprecated in favor of `Deno.File::close`?
> > 
> > 
> > It's still debated, but most likely it won't be deprecated.
> 
[...]

PR #12139, opened 2021-09-19
feat(unstable): add lock and unlock methods to `Deno.File`

caspervonb:
This adds `lock` and `unlock` methods to `Deno.File` along with their sync variants.
This inherits being unstable from the ops being unstable.
 
Towards https://github.com/denoland/deno/issues/12107

### DIFF
--- a/cli/tests/unit/lock_test.ts
+++ b/cli/tests/unit/lock_test.ts
@@ -1,0 +1,161 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { assertEquals, unitTest } from "./test_util.ts";
+import { readAll } from "../../../test_util/std/io/util.ts";
+
+unitTest(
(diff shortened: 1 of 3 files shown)


caspervonb:
I'm not too sure about these tests, but its what the rid based tests do so rolling with it for now.

cli/tests/unit/lock_test.ts
```diff
@@ -0,0 +1,161 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { assertEquals, unitTest } from "./test_util.ts";
+import { readAll } from "../../../test_util/std/io/util.ts";
+
+unitTest(
```
dsherret, line 5: Just a note that these tests are flaky. Perhaps we could update these tests to use the changes in:

https://github.com/denoland/deno/pull/12255/files

Then remove that file and only use this one (since this transitively tests those functions).

PR #12145, opened 2021-09-20
Fix default value of "lib"

Eyal-Shalev:
Changed it from `"deno.window"` to `["deno.window"]`


dsherret (approved):
LGTM. Thanks @Eyal-Shalev!

PR #12146, opened 2021-09-20
fix(ext/crypto): key generation based on AES key length

littledivy:


bartlomieju:

littledivy:
@bartlomieju Can't have a unit test to check the length of the underlying key until we have `exportKey`. The `length` property of `AlgorithmIdentififer` was correct but the underlying key generated was of the wrong length. That said, the bug wasn't really noticeable in userland because we don't have any other AES operations at all (pretty sure the length-check code path will be hit a lot times once we do #12123 )

lucacasonato (approved):
LGTM, thanks.

PR #12147, opened 2021-09-20
fix(ext/crypto): decode id-RSASSA-PSS with default params

littledivy:
Keys with missing (default) parameters are now decodable. Example `AlgorithmIdentifer`:
```
SEQUENCE (2 elem)
    OBJECT IDENTIFIER 1.2.840.113549.1.1.10 rsaPSS (PKCS #1)
    SEQUENCE (1 elem)
      [2] (1 elem)
        INTEGER 30
```


bnoordhuis (approved):

bartlomieju (approved):
Thanks Divy!

Issue #2318, opened 2019-05-09
Escape perm whitelist using symlink


kevinkassimo:
e.g. Currently creating a symlink under write-allowed directory to other directories would allow escaping the perm rules and access any file.

piscisaureus:
I agree this is a problem, unfortunately it's very hard to fix (on some platforms at least).
On OS X and BSD derivatives it'd be possible to query the filename of an open file using fcntl(F_GETPATH). 
On windows, something similar is possible using NtQueryInformationFile, although this is a bit tricky to implement since the winapi crate doesn't expose these functions.
On linux, I'm sure it's possible somehow, but I don't know how...

This is of course for the case of *opening* files. For renaming, deleting, moving etc the trick would be to rely on calling `openat()` and after that checking that the path is still inside a permitted subdirectory, and then performing the move/delete operation on the file descriptor using `renameat()/unlinkat()`. 

So this is definitely fixable, but quite a but of work. That said, I think it'd be worthwhile and important to do it.

nayeemrmn:
Bump. This should really be a 1.0 blocker unless it's clarified that 1.0 =/= production-ready.

@bartlomieju Permissions label.

dubiousjim:
@piscisaureus Yeah I haven't yet found out how to recover a path from an open fd on Linux either. I don't know if it's possible (short of searching the device for the inode... or parsing a file in `/proc`, which wouldn't be _as_ ridiculously expensive, but still too expensive).

A general problem here is that an open fd _needn't have_ any existing path to it. Isn't it a common strategy for handling temp files to open the file and then unlinking its path (or that of the directory that contains it), while keeping the fd still open.

dubiousjim:
Related issues: #4437, related PRs: #2317, #2319.
Cc: @kevinkassimo @ry @piscisaureus @bartlomieju @nayeemrmn @afinch7 

Does the following sound reasonable for a minimal fix? (We might look for a more ambitious fix later.)

1. Since `op_link` will have to deposit its destination file in a directory with write permissions, thus exposing the original to write access, we require write access on this operation's source file (Currently we only require read access... it's not clear that's even necessary, unless the destination directory also has read access. If one's insisting on read access because this operation at least verifies the existence of the original file, we're already pretty unsystematic about this --- follow up at #4423.)

[...]

Issue #9607, opened 2021-02-26
Permissions checks should use canonical paths


caspervonb:
Currently permission checks only look at the paths they were provided which may be a symbolic link.
This allows you to escape the sandbox with symlinks which were just stabilized in https://github.com/denoland/deno/pull/9226

Most of the filesystem ops that deal with paths should *always* use the canonical path when checking permissions within the ops, the exception would be stat since there is an lstat and the link won't be followed later in the actual syscall.

There are other ways to we can go about it (capability based preopens with openat for example ala WASI), but not with our current permission model resolving to the true path is the way to go forward imho.

Related but without an actionable item https://github.com/denoland/deno/issues/2318.

Spoonbender:
I agree with the general approach as I specified in the other issues.

I want to better understand what is being proposed and what implementation tradeoffs are we willing to make.

About the permissions list itself, do we:
1. Resolve the symlink-based paths in the list upon startup? (but they might change in runtime)
2. Resolve the symlink-based paths in the list whenever checking permissions? (performance penalty)
3. Disallow symlink-based paths in the permissions list entirely? (but again, a non-symbolic path may change to symbolic path at runtime)

Similar issues for the input paths in filesystem ops: 
1. Do we resolve once and cache the result (which may change over time)?
2. Resolve each time an op is initiated?
3. Something in-between? Some scoped caching? e.g. force the user code to hold a "file" object and run all ops on it (so we won't have `Deno.readTextFile()` but instead `let file = Deno.getFileHandle('/tmp/myfile'); let s = file.readText(); file.writeTest('something');` (this enables us to hold a file-descriptor inside the `file` object, so it doesn't change "beneath our feet")

bartlomieju:
~~This is the case as of Deno v2~~

Edit: No, what was decided was that creating symlinks requires `--allow-all` permissions (https://github.com/denoland/deno/pull/12554) and existing symlinks can be read regardless of where they point to.

PR #12154, opened 2021-09-20
fix: Deno.symlink and Deno.symlinkSync require read and write permissions

### LINKED ISSUE #12152: [Security] Deno Sandbox Escape
The Deno file sandbox does not handle symbolic links correctly.
When running Deno with specific write access the `Deno.symlink` method can be used to gain access to any directory.

Proof of concept:
https://hackers.report/report/614876917a7b150012836bb8

bartlomieju:
Fixes #12152


caspervonb:
Been discussing a bit offline with Bartek, re-iterating some of the main points we touched on here:
- This is a symptom of https://github.com/denoland/deno/issues/2318
- As this is a symptom, it doesn't actually fix the real problem.
- We should canonicalize when checking permissions https://github.com/denoland/deno/issues/9607 (this will take a hit on performance).
- Requiring permissions what the symlink points at is awkward because that doesn't actually have to **exist** and/or it can also be a symlink. If we require read and write on the old path and canonicalise, this gets strange as that would follow symlinks during the time of symlink creation which seems very undesirable.

ry:
This PR fixes #2318.

If there is an existing symlink (before Deno was started) we are not going to try to guard access to that. We do need to ensure symlinks cannot be created outside of the sandbox.

caspervonb:
> This PR fixes #2318.
> 
> If there is an existing symlink (before Deno was started) we are not going to try to guard access to that. We do need to ensure symlinks cannot be created outside of the sandbox.

It really doesn't.
All I need to get root access with this, is create a relative symlink pointing back a few levels, then move that to cwd.

ry:
> Assuming I have access to cwd then all I need to get root access with this is to create a relative symlink pointing back a few relative levels, then move that to cwd.

You won't be able to create that symlink after this patch (if I understand you correctly)

caspervonb:
> > Assuming I have access to cwd then all I need to get root access with this is to create a relative symlink pointing back a few relative levels, then move that to cwd.
> 
> You won't be able to create that symlink after this patch (if I understand you correctly)

One of many possibilities:

- Run deno with --allow-read=./ --allow-write=./
- Create a new directory
- Create a symlink in the new directory to ..
- Move that symlink to ./
- Escape

ry:
> Create a symlink in the new directory to ..

That won't be allowed after this patch.

ry (approved):
LGTM

caspervonb:
> > Create a symlink in the new directory to ..
> 
> That won't be allowed after this patch.

Yes it will, because if you have read-write to `./` then creating a symlink when in `./new_dir` that points to `..` is still allowed.

runtime/ops/fs.rs
```diff
@@ -1252,6 +1252,9 @@ fn op_symlink_sync(
   let oldpath = PathBuf::from(&args.oldpath);
   let newpath = PathBuf::from(&args.newpath);
 
+  state.borrow_mut::<Permissions>().read.check(&oldpath)?;
```
piscisaureus, line 1255: * This is not correct. Symlinks are resolved relative to their own location; this assumes that they're resolved relative to CWD.
* A sandbox escape is still possible by creating a symlink to "../secret" in "./allowed1/allowed2/symlink_here", and then moving the symlink from "allowed1/allowed2" to "allowed1".

piscisaureus (changes requested):

runtime/ops/fs.rs
bartlomieju: > This is not correct. Symlinks are resolved relative to their own location; this assumes that they're resolved relative to CWD.

Can you elaborate on this one?

> A sandbox escape is still possible by creating a symlink to "../secret" in "./allowed1/allowed2/symlink_here", and then moving the symlink from "allowed1/allowed2" to "allowed1".

Yes, @ry and I both agree on that and count it as separate issue tracked in #2318 and #9607.

The intent of this PR was to fix the immediate bug of not checking one permission when creating a symlink.

piscisaureus: > > This is not correct. Symlinks are resolved relative to their own location; this assumes that they're resolved relative to CWD.
> 
> Can you elaborate on this one?

If you create a symlink named `some/dir/over/there/mylink` with content `../target`, the symlink will point at `some/link/over/target`. So the permissions should be checked for these two paths:
  * `some/dir/over/there/mylink` (`newpath`)
  * `some/link/over/target` (`join(dirname(newpath), oldpath)`)

It currently checks permissions for:
  * `some/dir/over/there/mylink` (`newpath` - correct)
  * `../target` (`oldpath` - incorrect)

bartlomieju:
Discussed offline with @piscisaureus, #12170 is a better solution.

PR #12155, opened 2021-09-20
fix(runtime/permissions): Handle revoke in full allow context

### LINKED ISSUE #12153: Deno.permission.revoke() doesn't do anything unless allow list was provided in a flag
When writing a unit test using `permissions` setting object and specifying bool switch it turns out `Deno.permissions.revoke()` doesn't do anything:

```ts
unitTest(
  { perms: { read: true, write: true } },
  async function symlinkPermissionDeniedOnMissingSourceWrite() {
    const testDir = Deno.makeTempDirSync();
    const oldname = testDir + "/oldname";
    const newname = testDir + "/newname";
    Deno.mkdirSync(oldname);
    await Deno.permissions.revoke({ name: "write", path: oldname });
    await assertRejects(async () => {
      await Deno.symlink(oldname, newname);
    }, Deno.errors.PermissionDenied);
  },
);
```

In above case `assertRejects` fails. 

Looking through source code for `Deno.permissions.revoke()` API I found:
https://github.com/denoland/deno/blob/269bf380e0fb2e532e7340c4ba64d4eaa4d5d45e/runtime/permissions.rs#L404-L417

So permission is only removed from `granted_list`, but it's not added to the `denied_list`. I believe this is a bug and should be fixed.

CC @ry

crowlKats:
Fixes #12153

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -1887,6 +1887,36 @@ mod tests {
   }
 
   #[test]
+  fn test_allow_all_revoke() {
+    let mut perms = Permissions::allow_all();
+    #[rustfmt::skip]
+    {
+      assert_eq!(perms.read.revoke(Some(Path::new("/foo/bar"))), PermissionState::Prompt);
+      assert!(perms.read.check(Path::new("/foo/bar")).is_err());
+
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thank you Leo but I believe https://github.com/denoland/deno/pull/12159 is more viable solution. Closing without merge.

PR #12156, opened 2021-09-20
fix(workers): Don't panic when a worker's parent thread stops running

### LINKED ISSUE #11342: Exceptions thrown after a worker's initial execution panic
```
error: Uncaught (in worker "") Error
    throw new Error();
          ^
    at file:///home/abotella/test/worker.js:3:11
    at fire (deno:extensions/timers/01_timers.js:477:7)
    at handleTimerMacrotask (deno:extensions/timers/01_timers.js:338:7)
thread 'worker-0' panicked at 'coding error: either js is polling or the worker is terminated', runtime/web_worker.rs:498:9
    at Worker.#pollControl (deno:runtime/js/11_workers.js:247:23)
```

### test.js
```js
new Worker(
  { type: "module" },
);
```

### worker.js
```js
setTimeout(() => {
  throw new Error();
}, 1000);
```

-----

This seems to have been introduced in #11076, and it seems to be related to some of the WPT test failures in #11338.

andreubotella:
This panic could happen in the following cases:

- A non-fatal error being thrown from a worker, that doesn't terminate the worker's execution, but propagates to the main thread without being handled, and makes the main thread terminate.
- A nested worker being alive while its parent worker gets terminated.
- A race condition if the main event loop terminates the worker as part of its last task, but the worker doesn't fully terminate before the main event loop stops running.

This panic happens because a worker's event loop should have pending ops as long as the worker isn't closed or terminated – but if an event loop finishes running while it has living workers, its associated `WorkerThread` structs will be dropped, closing the channels that keep those ops pending.

This change adds a `Drop` implementation to `WorkerThread`, which terminates the worker without waiting for a response. This fixes the panic, and makes it so nested workers are automatically terminated once any of their ancestors is closed or terminated.

This change also refactors a worker's termination code into a `WorkerThread::terminate()` method.

Closes #11342.


andreubotella:
As it turns out, terminating the worker when `WebWorkerHandle` gets dropped doesn't work, because `WebWorkerHandle` can be cloned without creating a new worker (which is used by the worker_host ops in order to keep access to the handle across await points). But other than those ops, the only thing that uses a `WebWorkerHandle` is `WorkerThread`, which does seem to map one-to-one with a worker. Making that struct, rather than `WebWorkerHandle`, terminate the worker on drop seems like the right choice.

(I'll be updating the title of this PR and the commit message once CI passes.)

runtime/ops/worker_host.rs
```diff
@@ -75,6 +76,34 @@ pub struct WorkerThread {
...
+impl WorkerThread {
+  fn terminate(mut self) {
+    self.worker_handle.clone().terminate();
+    self
+      .join_handle
+      .take()
+      .unwrap()
+      .join()
+      .expect("Worker thread panicked")
+      .expect("Panic in worker event loop");
```
bartlomieju, line 88: I'm not sure if we should be `expect()`ing here. This would cause another panic if something goes wrong in worker and might be prone to other race conditions.

bartlomieju: Although that doesn't change current behavior so I guess it's fine for now

bartlomieju (approved):
LGTM, let's try this

PR #12159, opened 2021-09-20
fix(cli/permissions): ensure revoked permissions are no longer granted

### LINKED ISSUE #7271: More granular permissions system
### Problem revoking a subfolder
Currently you cannot revoke a subfolder permission.
**example :**
```ts
///1. Request permissions needed

const permStatus_1 = await Deno.permissions.request({ name: "read", path: "./folderA"}); //Grant
//Skipped because folderA was granted
const permStatus_2 = await Deno.permissions.request({ name: "read", path: "./folderA/SubFolderA" }); 

console.log(permStatus_1); //Granted
console.log(permStatus_2); //Granted

///2. Remove unused permissions

//Revoke the folderA
await Deno.permissions.revoke({ name: "read", path: "./folderA/SubFolderA" }); 
const permStatus_3 = await Deno.permissions.query({ name: "read", path: "./folderA" });
const permStatus_4 = await Deno.permissions.query({ name: "read", path: "./folderA/SubFolderA" });

console.log(permStatus_3); //Granted
console.log(permStatus_4); //Granted [!] The revocation was not taken into account
```
### Proposition : add the **`recursive`** optional attribute
It's an idea, but I think it might be interesting to consider the permission request of the subfolder, although the permission of parent folder was already granted.
Also, when a revocation is called, revoke all children only if the optional `recursive` parameter has been provided.

```ts
const permStatus_5 = await Deno.permissions.request({ name: "read", path: "./folderA" }); //Grant
//Skipped because folderA was granted, /!\ but add the path to the granted list
const permStatus_6 = await Deno.permissions.request({ name: "read", path: "./folderA/SubFolderA" }); 

console.log(permStatus_5 ); //Granted
console.log(permStatus_6 ); //Granted
```
**without** the `recursive` attribute :
```ts
await Deno.permissions.revoke({ name: "read", path: "./folderA" }); //Revoke the folderA 
const permStatus_7 = await Deno.permissions.query({ name: "read", path: "./folderA" });
const permStatus_8 = await Deno.permissions.query({ name: "read", path: "./folderA/SubFolderA" });

console.log(permStatus_7); //Prompt
console.log(permStatus_8); //Granted
```
**with** the `recursive` attribute :
```ts
await Deno.permissions.revoke({ name: "read", path: "./folderA", recursive: true }); //Revoke the folderA recursively
const permStatus_9 = await Deno.permissions.query({ name: "read", path: "./folderA" });
const permStatus_10 = await Deno.permissions.query({ name: "read", path: "./folderA/SubFolderA" });

console.log(permStatus_9);  //Prompt
console.log(permStatus_10); //Prompt
```

### LINKED ISSUE #12153: Deno.permission.revoke() doesn't do anything unless allow list was provided in a flag

nayeemrmn:
Currently `Deno.permissions.revoke()` no-ops if you try to revoke a permission which is weaker than one that was _explicitly granted_ on the CLI, since it would still be implied. This makes it stronger, now purging all permission entries required to ensure that the passed permission descriptor no longer returns granted.

Fixes #12153.

Also makes runtime-requested permissions, when granted or already granted, count as _explicitly granted_ entries. A clean side-effect of this is allowing a hack for more granular revokes.
```ts
// deno run --allow-read=foo temp.ts

// This request doesn't actually prompt because it's already transitively
// granted. But `--allow-read=foo/bar is now "explicitly granted".
console.log(await Deno.permissions.request({ name: "read", path: "foo/bar" })); // granted

// Revoke `--allow-read=foo`.
console.log(await Deno.permissions.revoke({ name: "read", path: "foo" })); // prompt

// `--allow-read=foo/bar` is still granted, because it's implied by a separate
// "explicitly granted" entry which is not purged up by the revoke.
console.log(await Deno.permissions.query({ name: "read", path: "foo/bar" })); // granted
```
Closes #7271.

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -240,23 +240,14 @@ impl UnaryPermission<ReadDescriptor> {
           "read access to \"{}\"",
           display_path.display()
         )) {
-          self
-            .granted_list
-            .retain(|path| !path.0.starts_with(&resolved_path));
           self.granted_list.insert(ReadDescriptor(resolved_path));
           PermissionState::Granted
         } else {
-          self
-            .denied_list
-            .retain(|path| !resolved_path.starts_with(&path.0));
           self.denied_list.insert(ReadDescriptor(resolved_path));
           self.global_state = PermissionState::Denied;
           PermissionState::Denied
         }
       } else if state == PermissionState::Granted {
-        self
-          .granted_list
-          .retain(|path| !path.0.starts_with(&resolved_path));
         self.granted_list.insert(ReadDescriptor(resolved_path));
         PermissionState::Granted
       } else {
@@ -374,23 +365,14 @@ impl UnaryPermission<WriteDescriptor> {
           "write access to \"{}\"",
           display_path.display()
         )) {
-          self
-            .granted_list
-            .retain(|path| !path.0.starts_with(&resolved_path));
           self.granted_list.insert(WriteDescriptor(resolved_path));
           PermissionState::Granted
         } else {
-          self
-            .denied_list
-            .retain(|path| !resolved_path.starts_with(&path.0));
           self.denied_list.insert(WriteDescriptor(resolved_path));
           self.global_state = PermissionState::Denied;
           PermissionState::Denied
         }
       } else if state == PermissionState::Granted {
-        self
-          .granted_list
-          .retain(|path| !path.0.starts_with(&resolved_path));
         self.granted_list.insert(WriteDescriptor(resolved_path));
         PermissionState::Granted
       } else {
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
LGTM

PR #12163, opened 2021-09-21
docs: deprecate `Deno.flock` and `Deno.funlock`

caspervonb:
This deprecates `Deno.flock` and `Deno.funlock` in favour of the methods on the `Deno.File` class.


caspervonb:
This depends on https://github.com/denoland/deno/pull/12139

PR #12165, opened 2021-09-21
perf(web): optimize AbortController

AaronO:
- Use regular class constructor and symbol "private" attributes
- Lazy init Set of follower signals

Some benchmark data on class init:
```
class_member_private:    n = 1000000, dt = 0.076s, r = 13157895/s, t = 75ns/op
class_member:            n = 1000000, dt = 0.091s, r = 10989011/s, t = 91ns/op
class_constructor:       n = 1000000, dt = 0.008s, r = 125000000/s, t = 8ns/op
class_symbol_x:          n = 1000000, dt = 0.006s, r = 166666667/s, t = 6ns/op
class_symbol_anon:       n = 1000000, dt = 0.007s, r = 142857143/s, t = 7ns/op
```

### DIFF
--- a/ext/web/03_abort_signal.js
+++ b/ext/web/03_abort_signal.js
@@ -21,13 +21,12 @@
   const add = Symbol("add");
   const signalAbort = Symbol("signalAbort");
   const remove = Symbol("remove");
+  const aborted = Symbol("aborted");
+  const abortAlgos = Symbol("abortAlgos");
 
   const illegalConstructorKey = Symbol("illegalConstructorKey");
 
   class AbortSignal extends EventTarget {
-    #aborted = false;
-    #abortAlgorithms = new Set();
-
     static abort() {
       const signal = new AbortSignal(illegalConstructorKey);
       signal[signalAbort]();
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM

ext/web/03_abort_signal.js
```diff
@@ -21,51 +20,57 @@
...
   const signalAbort = Symbol("signalAbort");
   const remove = Symbol("remove");
+  const aborted = Symbol("aborted");
+  const abortAlgos = Symbol("abortAlgos");
 
   const illegalConstructorKey = Symbol("illegalConstructorKey");
 
   class AbortSignal extends EventTarget {
-    #aborted = false;
-    #abortAlgorithms = new Set();
```
dsherret, line 29: In the future, it would be good to add comments about performance optimizations so someone doesn't accidentally do something to cause a regression. For example, stating here not to use private properties on this class.

bartlomieju: This change will probably have a side effect of leaking these fields during object inspection - AFAIK private fields do not show up during inspection. I'm not sure though if it is against the spec or anything.

crowlKats: > For example, stating here not to use private properties on this class.

Well for Web APIs we don't use private properties anywhere, but only symbols. so this was just old code that wasn't updated yet to how we do things.

PR #12167, opened 2021-09-21
cleanup(ext/fetch): drop redundant webidl converters in fetch()

AaronO:
Since those inputs are passed to `new Request(...)` which applies webidl converters

PR #12168, opened 2021-09-21
perf(ext/fetch): skip USVString webidl conv on string constructor

AaronO:
Since URL parsing will effectively normalize it, this is essentially another redundant webidl conversion

Assumes #12167

### DIFF
--- a/ext/fetch/23_request.js
+++ b/ext/fetch/23_request.js
@@ -220,10 +220,6 @@
     constructor(input, init = {}) {
       const prefix = "Failed to construct 'Request'";
       webidl.requiredArguments(arguments.length, 1, { prefix });
-      input = webidl.converters["RequestInfo"](input, {
-        prefix,
-        context: "Argument 1",
-      });
       init = webidl.converters["RequestInit"](init, {
         prefix,
         context: "Argument 2",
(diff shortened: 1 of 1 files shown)


andreubotella:
The conversion to `RequestInfo` first checks if the value is an instanceof `Request`, and if it isn't, it turns it into a USVString – not the other way around.

AaronO:
> The conversion to `RequestInfo` first checks if the value is an instanceof `Request`, and if it isn't, it turns it into a USVString – not the other way around.
> 
> See https://heycam.github.io/webidl/#es-union

Yeah the alternative patch was to use `DOMString` instead of `USVString` since `op_url_parse` (or any op for that matter) will implicitly do `DOMString -> USVString`, I'll submit that

AaronO:
@andreubotella Please take another look

ext/fetch/23_request.js
```diff
@@ -431,7 +431,8 @@
         return webidl.converters["Request"](V, opts);
       }
     }
-    return webidl.converters["USVString"](V, opts);
+    // Passed to new URL(...) which implictly converts DOMString -> USVString
+    return webidl.converters["DOMString"](V, opts);
```
andreubotella, line 435: Since the `RequestInfo` type in the fetch WebIDL definitions is `Request | USVString`, you might want to rename the WebIDL type, to avoid mistakes down the line.

lucacasonato (approved):
LGTM

PR #12170, opened 2021-09-21
fix(runtime): use canonical paths for ops that may resolve symlinks

caspervonb:
Supercedes https://github.com/denoland/deno/issues/12154

This checks the canonicalised path rather than the given path for ops that result in system calls that follow symbolic links.

- [x] Deno.open
- [x] Deno.openSync
- [x] Deno.chdir
- [x] Deno.stat
- [x] Deno.statSync
- [ ] ~~Deno.copyFile~~ (uses open(2) but doesn't follow symlinks as far as I know, but double check).
- [ ] ~~Deno.copyFileSync~~ (uses open(2) doesn't follow symlinks as far as I know, but double check).
- [x] Deno.makeTempDir
- [x] Deno.makeTempDirSync
- [x] Deno.makeTempFile
- [x] Deno.makeTempFileSync
- [x] Deno.mkdir
- [x] Deno.mkdirSync
- [ ] ~~Deno.remove~~ (remove(2) doesn't follow symlinks)
- [ ] ~~Deno.removeSync~~ (remove(2) doesn't follow symlinks)
- [ ] ~~Deno.rename~~ (rename(2) is symlink aware, but shouldn't be following)
- [ ] ~~Deno.renameSync~~ (rename(2) is symlink aware, but shouldn't be following)
- [ ] Deno.run
- [ ] Deno.listen
- [ ] Deno.connect
- [ ] import
 
Closes https://github.com/denoland/deno/issues/2318
Fixes https://github.com/denoland/deno/issues/9607
Fixes https://github.com/denoland/deno/issues/12152


piscisaureus:
Other path ops:
* listen/connect on unix socket
* (dynamic) import
* run

piscisaureus:
TOCTOU risks discussed on discord.

runtime/permissions.rs
```diff
@@ -306,6 +307,11 @@ impl UnaryPermission<ReadDescriptor> {
     result
   }
 
+  /// As `check`, but tries to canonicalize the path.
+  pub fn check_canonical(&mut self, path: &Path) -> Result<(), AnyError> {
+    self.check(&canonicalize_path(path).unwrap_or_else(|_| path.to_path_buf()))
```
ry, line 312: This is going to add an extra syscall to every file system method. 

Is the purpose of this patch just to resolve the symlink security bug? We ought to consider solutions that don't involve making the the file system bindings inherently slower.

caspervonb: > Is the purpose of this patch just to resolve the symlink security bug? We ought to consider solutions that don't involve making the the file system bindings inherently slower.

Got any suggestions? 🙂 

In this case I'd say correctness is more important than performance, since we have an ACL based sandbox permission system we _must_ resolve symlinks or we're just left with a pile of sand.

While it isn't free, one extra call per path shouldn't be that detrimental. We already have extra redundant flush call per i/o write and no one ever noticed on the benchmarks.

Open to suggestions, discussed this briefly with @piscisaureus but as long as our permissions are based on ACLs we're kinda stuck with this.

Preferable would be to switch over to a capability model ala what the CloudABI provides (WASI is loosely modelled after that again). The premise is that file system access revolves around pre-opened fds granted by an ambient authority (say, user permission prompts, cli flags, etc) which then delegates all access via openat but that means our file system interfaces should also change to reflect that.

Some relevant Rusty work happening here in tangent to wasmtime and the WASM projects, like cap-std (https://github.com/bytecodealliance/cap-std).

But to go down this route we need to rethink the entire file-system interface, native-file-system comes pretty close to this model actually but 2.0 is right around the corner so first things first is just to fix this defect.

ry: Here are some ideas:
- `Deno.symlink` cannot be used unless full file system permissions are given `--allow-read --allow-write`
- Remove `Deno.symlink` entirely
- Allow `Deno.symlink` in subtrees but target path cannot start with `..`

piscisaureus: I disagree with @ry. This PR is on the right track. 

Removing symlink() or requiring -A are stop gaps at best.

ry: This PR would make Deno inherently slower than Node at FS stuff. 

Wouldn't most people rather just not have symlink accessible when they have only subtree permissions?

piscisaureus: > This PR would make Deno inherently slower than Node at FS stuff.

It won't make any difference when deno is running with full tree write permissions. 
The subtree-write case can be optimized.

> Wouldn't most people rather just not have symlink accessible when they have only subtree permissions?

Your holding it wrong. People don't have permissions, people grant permissions. I would rather not be forced to to pass --allow-all because some script uses a symlink.

If sandboxing is too slow for your liking you can use -A.

PR #12176, opened 2021-09-22
chore: replace calls to assertThrowsAsync with assertRejects

caspervonb:


dsherret (approved):
LGTM. This new function name is less discoverable though.

PR #12178, opened 2021-09-22
chore(ext/net): improve embedder friendliness

bnoordhuis:
Default to None if UnsafelyIgnoreCertificateErrors is not present in the
OpState.

Embedders may not have a need for restricting outgoing TLS connections
and having them hunt through the source code for the magic incantation
that makes the borrow panics go away, is less user friendly.

<hr>

It'd be nice if this pattern was adopted more broadly (hint, hint.)

### DIFF
--- a/ext/net/ops_tls.rs
+++ b/ext/net/ops_tls.rs
@@ -699,9 +699,8 @@ where
 
   let unsafely_ignore_certificate_errors = state
     .borrow()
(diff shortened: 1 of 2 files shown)


satyarohith (approved):
LGTM.


Edit: should we also do the same for DefaultTlsOptions?

PR #12179, opened 2021-09-22
fix(ext/http): fortify "is websocket?" check

bnoordhuis:
Check for expected headers more rigorously and, notably, check that
it's a HTTP/1.1 GET request. Hyper panics if you try to upgrade HTTP/2
connections.

This mirrors what Deno Deploy and the tungstenite crate do, with the
caveat that tungstenite supports multiple Connection header tokens
(e.g. "Connection: upgrade, foo") whereas Deploy does not.

I'm reasonably sure no reasonable client sends multiple tokens when it
also sends "Upgrade: websocket" so I opted to leave that out.

The presence of "Sec-Websocket-Version: 13" is now also enforced.
I don't expect that to break anything: conforming clients already send
it and tungstenite can't talk to older clients anyway.

The new code is more efficient due to heap-allocating less.

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -22,6 +22,10 @@ use deno_core::Resource;
 use deno_core::ResourceId;
 use deno_core::ZeroCopyBuf;
 use hyper::body::HttpBody;
+use hyper::header::CONNECTION;
+use hyper::header::SEC_WEBSOCKET_KEY;
+use hyper::header::SEC_WEBSOCKET_VERSION;
+use hyper::header::UPGRADE;
 use hyper::http;
 use hyper::server::conn::Http;
 use hyper::service::Service as HyperService;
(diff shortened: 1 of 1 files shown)


ext/http/lib.rs
```diff
@@ -312,20 +316,21 @@ fn req_headers(
...
 
 fn is_websocket_request(req: &hyper::Request<hyper::Body>) -> bool {
-  req_header_contains(req, hyper::header::CONNECTION, "upgrade")
-    && req_header_contains(req, hyper::header::UPGRADE, "websocket")
+  req.version() == hyper::Version::HTTP_11
+    && req.method() == hyper::Method::GET
+    && req.headers().contains_key(&SEC_WEBSOCKET_KEY)
+    && header(req.headers(), &SEC_WEBSOCKET_VERSION) == b"13"
+    && header(req.headers(), &UPGRADE).eq_ignore_ascii_case(b"websocket")
+    && header(req.headers(), &CONNECTION).eq_ignore_ascii_case(b"upgrade")
```
crowlKats, line 324: wouldn't cause #11494 again?

bnoordhuis: Oh, so some clients do send multiple tokens... Interesting, because my Firefox doesn't. Anyway, I'll fix it.

bnoordhuis: Unsettling. There's a httpUpgradeWebSocketMultipleConnectionOptions test that exercises multiple connection header tokens... and it passes.

crowlKats: That test is for JS only, it doesnt rely on the code you changed

bartlomieju (approved):

PR #12180, opened 2021-09-22
feat: move various runtime ops to separate crate

crowlKats:
Currently this just moves various ops and js files to a `sys` ext; not the most ideal name, just a temporary placeholder.
Currently each op file has its own init and perms trait (if necessary). Maybe we should flatten them down to a single one? Or separate into multiple exts


bartlomieju (commented):
I'm 🤷 on this one. It adds a lot of lines for no obvious benefit. Is there a need to have a separate crate for these ops?

bartlomieju:
I'm not in favor of this change - it one more crate to release and maintain. Let's close it for now and revisit if there's an actual need for it.

PR #12181, opened 2021-09-22
fix(ext/http): include port number in h2 urls

### LINKED ISSUE #12171: The port is missing from Request.url returned by HttpConn.nextRequest for an h2 TLS connection.
When using the 'h2' protocol the port is missing in the requestEvent.request.url property. When 'h2' is removed from alpnProtocols then the port is correctly shown.

In the following example, navigating to [](http://localhost:8080) will show the port, but navigating to [](https://localhost:8043) strips the port from the request url. The port reappears when 'h2' is removed from the alpnProtocols list.

This occurs on 1.14.0 with the --unstable flag.

The following example will show `https://localhost/` over h2 but `https://localhost:8043/` over http/1.1

```js
const server = Deno.listen({
	port: 8080,
});
const serverTls = Deno.listenTls({
	port: 8043,

	certFile: "...",
	keyFile: "...",
	alpnProtocols: ["h2", "http/1.1"], // <= Remove 'h2' to see that it works correctly over http/1.1
});

(async () => {
	for await (const conn of server) serveHttp(conn);
})();

(async () => {
	for await (const conn of serverTls) serveHttp(conn);
})();

async function serveHttp(conn: Deno.Conn) {
		console.log('Missing the port number: ', requestEvent.request.url)
		requestEvent.respondWith(new Response(requestEvent.request.url, { status: 200 }))
	}
}
```

This could be caused by the req_url method but I am new to Rust and haven't gone too deep into the source:
https://github.com/denoland/deno/blob/cee5be45394c77863dacbf8f1d0f0b90188d2cca/ext/http/lib.rs#L267

bnoordhuis:
Fixes #12171.

### DIFF
--- a/cli/tests/integration/mod.rs
+++ b/cli/tests/integration/mod.rs
@@ -2,6 +2,7 @@
 
 use crate::itest;
 use deno_core::url;
+use deno_runtime::deno_fetch::reqwest;
 use deno_runtime::deno_net::ops_tls::TlsStream;
 use deno_runtime::deno_tls::rustls;
 use deno_runtime::deno_tls::webpki;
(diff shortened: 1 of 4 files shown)


kitsonk (approved):
LGTM

Issue #12075, opened 2021-09-14
integration::worker::workers is flaky


ry:
```
2021-09-14T21:57:04.0448811Z test worker location ...
2021-09-14T21:57:04.0449620Z OUTPUT
2021-09-14T21:57:04.0453133Z ##[error]thread 'integration::worker::workers' panicked at 'bad exit code, expected: 0, actual: 1', test_util\src\lib.rs:1493:9
2021-09-14T21:57:04.0455255Z stack backtrace:
2021-09-14T21:57:04.0456292Z    0:     0x7ff6721ceaa4 - std::backtrace_rs::backtrace::dbghelp::trace
2021-09-14T21:57:04.0457934Z                                at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
2021-09-14T21:57:04.0460607Z    1:     0x7ff6721ceaa4 - std::backtrace_rs::backtrace::trace_unsynchronized
2021-09-14T21:57:04.0462265Z                                at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b\/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
2021-09-14T21:57:04.0463736Z    2:     0x7ff6721ceaa4 - std::sys_common::backtrace::_print_fmt
2021-09-14T21:57:04.0465217Z                                at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b\/library\std\src\sys_common\backtrace.rs:67
2021-09-14T21:57:04.0466649Z    3:     0x7ff6721ceaa4 - std::sys_common::backtrace::_print::impl$0::fmt
2021-09-14T21:57:04.0468204Z                                at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b\/library\std\src\sys_common\backtrace.rs:46
2021-09-14T21:57:04.0469509Z    4:     0x7ff67205cc1a - core::fmt::write
2021-09-14T21:57:04.0470840Z                                at /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b\/library\core\src\fmt\mod.rs:1115
2021-09-14T21:57:04.0472470Z    5:     0x7ff6721ce338 - std::io::Write::write_fmt<alloc::vec::Vec<u8,alloc::alloc::Global> >
[...]

PR #12182, opened 2021-09-22
Mark integration::worker::workers as flaky

ry:
Ref https://github.com/denoland/deno/issues/12075


bartlomieju:
This should not be needed once https://github.com/denoland/deno/pull/12156 lands.

ry:
Ok - Then I'm closing this.

PR #12185, opened 2021-09-22
refactor: simplify FetchFuture

ry:
No need for static lifetime

### DIFF
--- a/cli/specifier_handler.rs
+++ b/cli/specifier_handler.rs
@@ -24,13 +24,8 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 pub type DependencyMap = HashMap<String, Dependency>;
-pub type FetchFuture = Pin<
-  Box<
-    (dyn Future<Output = Result<CachedModule, (ModuleSpecifier, AnyError)>>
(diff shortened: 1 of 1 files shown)


cli/specifier_handler.rs
```diff
@@ -24,13 +24,8 @@ use std::pin::Pin;
...
 pub type DependencyMap = HashMap<String, Dependency>;
-pub type FetchFuture = Pin<
-  Box<
-    (dyn Future<Output = Result<CachedModule, (ModuleSpecifier, AnyError)>>
-       + 'static
-       + Send),
-  >,
->;
+type FetchFutureOutput = Result<CachedModule, (ModuleSpecifier, AnyError)>;
+pub type FetchFuture = Pin<Box<dyn Future<Output = FetchFutureOutput> + Send>>;
```
bartlomieju, line 28: This `Send` seems fishy as well - we're using a single threaded runtime after all

ry: It's not trivially removable ...

bartlomieju (approved):

PR #12188, opened 2021-09-22
feat(unstable): provide ops details for ops sanitizer failures

### LINKED ISSUE #12142: Op sanitizer should use print a detailed diff
We have better metrics available in `Deno.metrics().ops` but the sanitizer was written before these.
We should make use of them to print a more detailed diff, as it is not always clear what kind of op is leaking.

caspervonb:
This adds an additional section to ops sanitizer failures with containing op names along with before and after counts.
This is only available in unstable as `Deno.metrics().ops` is only in unstable.

Closes #12142

### DIFF
--- a/cli/tests/integration/test_tests.rs
+++ b/cli/tests/integration/test_tests.rs
@@ -133,6 +133,12 @@ itest!(allow_none {
   output: "test/allow_none.out",
 });
 
+itest!(ops_sanitizer_unstable {
+  args: "test --unstable test/ops_sanitizer_unstable.ts",
+  exit_code: 1,
+  output: "test/ops_sanitizer_unstable.out",
+});
+
 itest!(exit_sanitizer {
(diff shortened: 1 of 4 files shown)


bartlomieju:
I'm fine with landing this as is, but I'd prefer if we stabilized detailed op metrics (https://github.com/denoland/deno/issues/12142#issuecomment-926981477)

@lucacasonato @AaronO

AaronO:
> I'm fine with landing this as is, but I'd prefer if we stabilized detailed op metrics ([#12142 (comment)](https://github.com/denoland/deno/issues/12142#issuecomment-926981477))
> 
> @lucacasonato @AaronO

I think this could be helpful in some tests and shouldn't semantically break with the planned op-metrics changes.

However given that completed >= dispatched, I think we can reduce the verbosity to something like:
```
N uncompleted
```

PR #12189, opened 2021-09-22
test(cli): align unit test permissions with runtime test permissions

caspervonb:
Aligns UnitTestPermissions type with the Deno.TestDefinition.permissions.

Came up in #12170 where I had to use Deno.test directly instead of unitTest to permission sets as intended.


bartlomieju (approved):
LGTM

Issue #11901, opened 2021-09-02
Implement test steps API


dsherret:
We're going to start work on a prototype of the test steps API.

- [ ] Implement proposal without metadata as unstable https://github.com/denoland/deno/discussions/10771
- [ ] Implement proposal with metadata
- [ ] Write polyfill for declarative in CLI https://github.com/denoland/deno/discussions/11235
- [ ] Decide on declarative living in CLI, std, or /x
- [ ] Test implementation with mocha polyfill https://gist.github.com/lucacasonato/54c03bb267074aaa9b32415dbfb25522

sebastienfilion:
The RFC you link (https://github.com/denoland/deno/discussions/10771) is a long conversation with a lot of contracting concepts. Which one exactly do you propose to implement? 
Also, can you define what "metadata" are you referring to?

dsherret:
What's outlined in the main post. See `stepMetadata` property.

sebastienfilion:
Are you going to implement the Tester class? Otherwise, I don't understand to what end.
Also this post is confusing; are you just saying that we're going to implement everything and see what sticks?

dsherret:
We're implementing what's outlined in the RFC at #10771 (so yes, the `Tester` class outlined there). Then we're going to implement a pollyfill for #11235 that uses the implementation of #10771. After we'll make further decisions on where the #11235 approach will live.

caspervonb:
> After we'll make further decisions on where the #11235 approach will live (it might end up being avialble in the CLI, std, or even in deno.land/x. In the case of it being in the CLI then we'll drop the stepMetadata property, but all of this isn't decided yet)

This is fairly confusing.
So we're doing declarative test registration that runs imperative tests that then can be used to registers declarative tests again?

dsherret:
> So we're doing declarative test registration that runs imperative tests that then can be used to registers declarative tests again?

This is a temporary implementation detail. No decision has been made in this regard.

caspervonb:
Immediately executing user controlled tests still fundamentally clash with how regular tests _actually_ work with deferred runner controlled and throttled execution so I'm **very confused** as to how we are planning on making deferred test execution live in deno.land/std or deno.land/x.

caspervonb:
Just to simplify and clarify the main concern, are the execution of subtests in this metadata thing deferred and controlled by the test runner or user?

lucacasonato:
- The step metadata only affects filtering (and maybe display). It can not be used to provide things like permissions.
[...]

PR #12190, opened 2021-09-22
feat(test): imperative test steps API

dsherret:
Implementation of #10771 and part of #11901.


caspervonb (changes requested):
Nice test coverage! but this changes a lot of unnecessary things (undoing things I've more or less just refactored to be that way, like the enums) and seems overly complex.

Also https://github.com/denoland/deno/pull/11736 has been open for a month.

dsherret:
> but this changes a lot of unnecessary things (undoing things I've more or less just refactored to be that way, like the enums) and seems overly complex.

Could you point out the specific areas you mean in the PR by commenting on them? As for the complexity, I believe it is necessary to implement this feature based on the RFC as there are a lot of conditions that need to be enforced at runtime. I'm sure there are a few areas that could be refactored to be a little cleaner though as this is just a first pass.

> Also #11736 has been open for a month pending consensus, looks like they're at about the same level of completeness (mine can stream output tho).

I did not see this PR. I'm sorry about that and feel bad... nobody told me about it. In the future though it would be helpful to open an issue for work that you're working/going to be working on to signal that you're doing that work. Then once done linking the PR to that issue. For example, I opened #11901 and assigned it to myself to signal I was going to work on this a few weeks ago... it would have been helpful to have commented on that with your PR issue to let me know that this had already been done.

I don't think the streaming output works properly with the approach in that PR with test steps executed in parallel. The output will trample all over itself. With this PR we could improve it to be streaming, but this could only be done reliably when a test has sanitizers and we know that it will be executing one at a time. It would then need to handle switching between the two possibilities. I think that could be a future improvement and perhaps we could use the API from your PR to help do that.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -113,26 +113,48 @@ declare namespace Deno {
    * See: https://no-color.org/ */
   export const noColor: boolean;
 
-  export interface TestDefinition {
-    fn: () => void | Promise<void>;
+  export class Tester {
```
dsherret, line 116: I was just implementing the RFC as-is, but I don't think we should bother having this as an exposed class to the user. Maybe just make it an interface and don't export it?

caspervonb: Unsound to have as an exported class, test ops are only available with test command.

lucacasonato: Yeah - let's not export it.

cli/tools/test.rs
```diff
@@ -76,12 +76,39 @@ pub enum TestOutput {
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub enum TestResult {
+pub struct TestResult {
```
caspervonb, line 79: Revert this type to the enum it's supposed to be

cli/tools/test.rs
```diff
@@ -97,7 +124,7 @@ pub enum TestEvent {
   Plan(TestPlan),
   Wait(TestDescription),
   Output(TestOutput),
-  Result(TestDescription, TestResult, u64),
+  Result(TestDescription, TestResult),
```
caspervonb, line 127: Revert

cli/tools/test.rs
```diff
@@ -184,23 +210,75 @@ impl TestReporter for PrettyTestReporter {
...
+      TestResultStatus::Failed(_) => colors::red("FAILED").to_string(),
     };
+    sb.push_str(&status);
+    sb.push_str(" ");
+    push_duration(&mut sb, result.duration);
+    sb.newline();
+
+    print!("{}", sb.into_string());
+
+    fn push_test_step_results(
```
caspervonb, line 233: Print directly to stdout, don't allocate here.

cli/tools/test.rs
```diff
@@ -237,6 +315,55 @@ impl TestReporter for PrettyTestReporter {
   }
 }
 
+struct IndentableStringBuilder {
```
caspervonb, line 318: No need for this, just print directly to stdout.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -113,26 +113,47 @@ declare namespace Deno {
    * See: https://no-color.org/ */
   export const noColor: boolean;
 
-  export interface TestDefinition {
-    fn: () => void | Promise<void>;
+  export interface Tester {
```
caspervonb, line 116: We need a different name, Tester is the name of the actual main test runner type.

caspervonb (changes requested):
Looking better but still would just go with the existing PR rather than putting in more work on this

cli/tools/test.rs
dsherret: Could you clarify why?

caspervonb: An Rust enum can represent N-state very well, extracting it into a struct you end up duplicating things. The same result can be used seamlessly for both test result and step results. Here you end up duplicating errors in both the test step and test step result. They're just test results, not a specific kind of test result.

dsherret:
> Looking better but still, like I've been saying offline I would just go with the existing PR rather than putting in the work on making this landable.

The duplicate PR is unfortunate, but there is still some scenarios not handled by #11736 and it also has filtering which hasn't been discussed (we probably should add that in a follow up PR).

> If you're going to keep at this it would be useful to differentiate or do something different here. Maybe showcase how metadata works and how that provides us with the tight deferred test runner controlled execution can be polyfilled with that as was mentioned in #11901?
>
> Premise put forth seems to be that we don't need tight integration with the test runner, I'd like to see that proven before allowing either steps implementations to be merged.

Like you, I'm still not convinced we should go with metadata, but it would work. For now, I've moved this API to unstable.

runtime/js/99_main.js
```diff
@@ -216,6 +216,9 @@ delete Object.prototype.__proto__;
       runtimeOptions.v8Version,
       runtimeOptions.tsVersion,
     );
+    if (runtimeOptions.unstableFlag) {
+      internals.enableTestSteps();
```
dsherret, line 220: It might be better if we instead stored whether we're running with the unstable flag. I wasn't sure where that could go though.

lucacasonato:
This is working great with my mocha polyfill:
```js
import "https://gist.githubusercontent.com/lucacasonato/54c03bb267074aaa9b32415dbfb25522/raw/fc2cb1d656c733f76a5db6792fa3a26b9b683e33/deno_mocha.ts";

describe("group", () => {
  it("hello", () => {
    throw new Error("Hey!");
  });

  it("world", () => {
    1 + 2;
  });
});
```

lucacasonato (commented):
This looks great. The assertions to prevent footguns are all there. I'm happy with landing this in unstable.

cli/dts/lib.deno.ns.d.ts
lucacasonato: Do you have a different suggestion? `TestContext`?

caspervonb: `TestContext` or just `Test`

bartlomieju:
> This is working great with my mocha polyfill:
> 
> ```js
> import "https://gist.githubusercontent.com/lucacasonato/54c03bb267074aaa9b32415dbfb25522/raw/fc2cb1d656c733f76a5db6792fa3a26b9b683e33/deno_mocha.ts";
> 
> describe("group", () => {
>   it("hello", () => {
>     throw new Error("Hey!");
>   });
> 
>   it("world", () => {
>     1 + 2;
>   });
> });
> ```

Just FYI it seems Mocha can already work with Deno: https://github.com/denoland/deno_std/pull/1334/files#diff-fa7e75c2cff0b94ed10f5be1e8c056c56d52ca0f8fcb6e3ec3a2c2aed37291af

cli/dts/lib.deno.ns.d.ts
dsherret: Renamed to `TestContext`.

hayd:
TestContext broke tests in deno-lambda's serverless example, which was doing its own setup/teardown:

https://github.com/hayd/deno-lambda/blob/debc93d9416bc546a06e139276b57367ca45da45/example-serverless/test_util.ts#L9

Are we supposed to just [stub in `fn(undefined as any)`](https://github.com/hayd/deno-lambda/pull/147/commits/f5f5a2f5d6fbbcc2e13c98eba3691fd8ac4fbf0e) ? 😳 

Is there an issue/discussion for [TestDefinition](https://doc.deno.land/builtin/stable#Deno.TestContext)?

dsherret:
@hayd you may just want to provide:

```ts
fn({
  step() {
    throw new Error("Test steps are not supported by deno-lambda at this time.");
  }
} as any)
```

PR #12198, opened 2021-09-23
BREAKING(cli/eval): remove --ts flag

satyarohith:
Use `deno eval --ext` instead.

Should be merged after the merge window for 2.0 opens (27 October).

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -3006,8 +3006,12 @@ mod tests {
 
   #[test]
   fn eval_typescript() {
-    let r =
-      flags_from_vec(svec!["deno", "eval", "-T", "'console.log(\"hello\")'"]);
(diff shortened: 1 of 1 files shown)


kitsonk:
Is there a related issue?

satyarohith:
> Is there a related issue?

No. The PR addresses a TODO that was added when we introduced the `--ext` flag.

kitsonk:
I think even with minor scope, we should probably discuss its inclusion.  Especially `TODO`s in codes might have seen like a good idea at some point, but we should discuss it for inclusion as well as track it as part of the 2.0 scope.  I have added it to the project board and will add it to the release plan.

bartlomieju:
2.0 was postponed indefinitely, so let's close this PR for now.

PR #12093, opened 2021-09-16
perf(ext/fetch): Use the WebIDL conversion to DOMString rather than USVString for Response constructor

lmalheiro:
Fixes performance issue https://github.com/denoland/deno/issues/12090

Added benchmark `response_string` which produced the following results before and after the change:

* using USVString:
```
Benchmark #13: /home/luismalheiro/projects/deno_pr/target/release/deno run cli/tests/testdata/response_string_perf.js 
  Time (mean ± σ):     432.9 ms ±  26.8 ms    [User: 308.8 ms, System: 192.8 ms]
  Range (min … max):   409.5 ms … 485.4 ms    10 runs
```
* using DOMString:
```
Benchmark #13: /home/luismalheiro/projects/deno_pr/target/release/deno run cli/tests/testdata/response_string_perf.js 
  Time (mean ± σ):     175.8 ms ±  17.6 ms    [User: 168.6 ms, System: 73.9 ms]
  Range (min … max):   147.3 ms … 211.6 ms    18 runs
```

Notice that the difference will increase non-linearly for large strings.

Issue #12090, opened 2021-09-15
Performance issue when using a string to create a Response object


lmalheiro:
This problem is similar to the one described in point 1 of issue https://github.com/denoland/deno/issues/10978 reported by @andreubotella.

Deno's `Response` object constructor accepts a string as a body parameter, which passes through an unnecessary conversion to `USVString` for the same reasons already explained in point 1 of previously mentioned issue. This results in significant loss of performance for large input parameters.

The conversion is done in this line of code:
https://github.com/denoland/deno/blob/4283e2907e13a089fa2d026c92e074f019550880/ext/fetch/22_body.js#L396

andreubotella:
After #12201, it seems like this issue can be closed.

Issue #13, opened 2018-05-30
Channels as async iterators


benjamingr:
Hey,

V8 ships with async iterators - I was wondering if they should be used for channels as they provide a bi-directional pull stream and have very nice `for... await` syntax for messages :) 

This also comes with all the debugging benefits of being a built-in and a standard between browsers, Node.js and other environments.

ry:
@benjamingr I love async iterators, but I think for the low-level pub/sub API I'll stick with the current interface for now. async iterators will definitely come in useful for high-level APIs - like reading data from a socket.

benjamingr:
@ry cool, if the reason is performance then I'm having a session with V8 tomorrow about promise-related performance and would be happy to bring it up.

PR #12201, opened 2021-09-23
perf(ext/fetch): Use the WebIDL conversion to DOMString rather than USVString for Response constructor

lmalheiro:
Fixes performance issue https://github.com/denoland/deno/issues/12090

* Replaces conversion to `USVString` by `DOMString` at the `BodyInit` WebIDL conversion.
* Adds benchmark `response_string` which produced the following results before and after the change:

  * using `USVString`:
  ```
  Benchmark #13: /home/luismalheiro/projects/deno_pr/target/release/deno run cli/tests/testdata/response_string_perf.js 
    Time (mean ± σ):     432.9 ms ±  26.8 ms    [User: 308.8 ms, System: 192.8 ms]
    Range (min … max):   409.5 ms … 485.4 ms    10 runs
  ```
  * using `DOMString`:
  ```
  Benchmark #13: /home/luismalheiro/projects/deno_pr/target/release/deno run cli/tests/testdata/response_string_perf.js 
    Time (mean ± σ):     175.8 ms ±  17.6 ms    [User: 168.6 ms, System: 73.9 ms]
    Range (min … max):   147.3 ms … 211.6 ms    18 runs
  ```
  Notice that the difference will increase non-linearly for larger strings.

This pull request replaces https://github.com/denoland/deno/pull/12093

### DIFF
--- a/ext/fetch/22_body.js
+++ b/ext/fetch/22_body.js
@@ -393,7 +393,7 @@
         return webidl.converters["ArrayBufferView"](V, opts);
       }
     }
-    return webidl.converters["USVString"](V, opts);
+    return webidl.converters["DOMString"](V, opts);
   };
   webidl.converters["BodyInit?"] = webidl.createNullableConverter(
     webidl.converters["BodyInit"],
--- a/ext/fetch/22_body.js
+++ b/ext/fetch/22_body.js
@@ -372,7 +372,7 @@
     return { body, contentType };
   }
 
-  webidl.converters["BodyInit"] = (V, opts) => {
+  webidl.converters["BodyInit_DOMString"] = (V, opts) => {
     // Union for (ReadableStream or Blob or ArrayBufferView or ArrayBuffer or FormData or URLSearchParams or USVString)
     if (V instanceof ReadableStream) {
       // TODO(lucacasonato): ReadableStream is not branded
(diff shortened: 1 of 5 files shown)


lucacasonato:
This is not correct though. USVString normalization doesn't happen at all now.

andreubotella:
> This is not correct though. USVString normalization doesn't happen at all now.

The `BodyInit` type is only used in the `Response` constructor and in `RequestInit`, which itself is only used in the `Request` constructor (and in the `fetch` function, which defers to the `Request` constructor). In both of these cases, the `BodyInit` parameter is passed to `extractBody` before doing anything else with it – and `extractBody` will UTF-8 encode strings with replacement, effectively performing the USV normalization.

ext/fetch/22_body.js
```diff
@@ -393,7 +393,7 @@
         return webidl.converters["ArrayBufferView"](V, opts);
       }
     }
-    return webidl.converters["USVString"](V, opts);
+    return webidl.converters["DOMString"](V, opts);
```
andreubotella, line 396: `BodyInit` is a WebIDL union type defined in the fetch specification. As I suggested in https://github.com/denoland/deno/pull/12168#discussion_r713215118, you should probably define a different WebIDL type (`BodyInit_DOMString` for example), to avoid confusion.

lmalheiro: @andreubotella, what would you suggest to do with the nullable converter `BodyInit?` ? Should I rename it too?

andreubotella: That's right

lmalheiro: Renamed.

lucacasonato (approved):
LGTM, thanks @lmalheiro and @andreubotella

PR #12202, opened 2021-09-23
refactor: Remove op_state parameter

ry:

### DIFF
--- a/cli/module_loader.rs
+++ b/cli/module_loader.rs
@@ -80,6 +80,9 @@ impl ModuleLoader for CliModuleLoader {
       referrer
     };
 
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):

cli/module_loader.rs
```diff
@@ -81,6 +80,9 @@ impl ModuleLoader for CliModuleLoader {
       referrer
     };
 
+    // TODO(ry) I think we can remove this conditional. At the time of writing
+    // we don't have any tests that fail if it was removed.
+    // https://github.com/WICG/import-maps/issues/157
```
bartlomieju, line 85: This implies that something like: `deno run --importmap im.json foo` would work if `foo` was part of import map

PR #12203, opened 2021-09-24
refactor: Remove depreated Worker::execute_module

ry:


bartlomieju (approved):
There's one more method in `deno_core` that has been deprecated and could be removed: `JsRuntime::load_module`. If Deploy is already up to date then LGTM

PR #12204, opened 2021-09-24
refactor: Rename ProgramState to ProcState

ry:
And move the Arc inside the struct.

### DIFF
--- a/cli/proc_state.rs
+++ b/cli/proc_state.rs
@@ -48,7 +48,7 @@ use std::sync::Arc;
 ///
 /// It is shared by all created workers (thus V8 isolates).
 #[derive(Clone)]
-pub struct ProgramState(Arc<Inner>);
+pub struct ProcState(Arc<Inner>);
 
 pub struct Inner {
   /// Flags parsed from `argv` contents.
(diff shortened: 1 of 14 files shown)


cli/proc_state.rs
```diff
@@ -41,12 +41,16 @@ use std::collections::HashSet;
...
 use std::io::BufReader;
+use std::ops::Deref;
 use std::sync::Arc;
 
 /// This structure represents state of single "deno" program.
 ///
 /// It is shared by all created workers (thus V8 isolates).
-pub struct ProgramState {
+#[derive(Clone)]
+pub struct ProcState(Arc<Inner>);
```
bartlomieju, line 51: Go with `ProcessState`? Saving three characters is not worth the abbrevation

bartlomieju (changes requested):

cli/proc_state.rs
ry: Why isn't it? Matches what we do in Deploy

bartlomieju: Okay, if you prefer it...

bartlomieju (approved):

PR #12205, opened 2021-09-24
experiment: derivative Default on worker options

AaronO:
Allowing for less verbose (and lower maintenance) worker construction in tests and embedders (as we move to facilitating embedding `deno_runtime` for projects like Deno Desktop)


AaronO:
Experiment inconclusive

PR #12206, opened 2021-09-24
chore: update wpt

andreubotella:
This PR also removes tests that were deleted or moved, and enables a number of WPT tests that were not listed in the expectations.

This PR does not enable the myriad of worker tests outside of the `workers` directory. That will be left to a follow-up PR.

### DIFF
--- a/tools/wpt/expectation.json
+++ b/tools/wpt/expectation.json
@@ -14852,10 +14852,6 @@
     "Close-delayed.any.html?wss": false,
     "stream": {
       "tentative": {
-        "abort.any.html?wpt_flags=h2": [
-          "abort before constructing should prevent connection",
-          "abort during handshake should work"
-        ],
         "abort.any.html?wss": true,
         "backpressure-receive.any.html?wpt_flags=h2": false,
         "backpressure-receive.any.html?wss": true,
--- a/tools/wpt/expectation.json
+++ b/tools/wpt/expectation.json
@@ -13043,11 +13043,43 @@
   },
   "dom": {
     "abort": {
+      "AbortSignal.any.html": true,
+      "AbortSignal.any.worker.html": true,
       "event.any.html": true
     },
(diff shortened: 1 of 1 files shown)


tools/wpt/expectation.json
```diff
@@ -13034,8 +13032,7 @@
     },
     "wrapKey_unwrapKey": {
       "wrapKey_unwrapKey.https.any.html": false
-    },
-    "randomUUID.any.html": true
```
lucacasonato, line 13038: Where did this test go?

lucacasonato: Oh it is `randomUUID.https.any.html` now.

tools/wpt/expectation.json
```diff
@@ -13598,6 +13649,190 @@
       "<a> and <area>.searchParams should be undefined"
     ],
     "idlharness.any.html": true,
+    "toascii.window.html": [
```
lucacasonato, line 13652: These tests are never going to pass (they use DOM elements). We should remove them.

tools/wpt/expectation.json
andreubotella: It was renamed to `randomUUID.https.any.js` in https://github.com/web-platform-tests/wpt/pull/28967. Looks like we added the renamed version but didn't remove the old name.

tools/wpt/expectation.json
andreubotella: Some of the tests use the `<a>` and `<area>` elements, but others use the `URL` class.

andreubotella:
`/websockets/stream/tentative/abort.any.html?wpt_flags=h2` seems to be flaky, so I've opened #12208 to track that and I'll be removing it from the expectations.

lucacasonato (approved):
LGTM. Thanks

Issue #3139, opened 2019-10-17
deno remove/uninstall subcommand


mhvsa:
We've got `deno install` and it would be convenient to also have `deno remove` or `deno uninstall` subcommand

bartlomieju:
We had `uninstall` but removed it
https://github.com/denoland/deno_std/pull/512#issuecomment-504005054

mhvsa:
oh, thank you

for myself `rm $(which foo)` works but I think it made sense to keep the `uninstall`

closing

roc:
Just landed in deno land and I was absolutely baffled there's no `uninstall` or `rm` command! Is the reason that you expect a single location for something to be installed, a single version as well? I find it confusing to have an API that has a way of adding something but not of removing it.

willgm:
@mhvsa
> for myself `rm $(which foo)` works but I think it made sense to keep the `uninstall`

What about Window users? Much of the Deno users may not know enough of his system terminal to remove it and feel lost about it.

I believe Deno should have a command to undo something that his own command have made.

Yamboy1:
Even as a linux user I would prefer a specific command to do this rather than having to mess around with shell expansions and the such

nateshmbhat:
> oh, thank you
> 
> for myself `rm $(which foo)` works but I think it made sense to keep the `uninstall`
> 
> closing

The issue with `rm $(which foo)` is that , people can have other binaries than the ones from `deno` that have the same name as deno binaries. This can happen when the `DENO` path is suffixed to the system PATH instead of "prefixing".

So we need a deno uninstall command that will only look in the DENO path and remove the required files

llluuulll:
hi,
so what command i can use to uninstall Deno  ? i am on wondows 10 and i am not at all confortable with complet unknow commands, i dont understand.

run$(witch foo), what is it wich foo ?
can some one point me the correct typo command to use ?

thx.

ry:
> The issue with rm $(which foo) is that , people can have other binaries than the ones from deno that have the same name as deno binaries. This can happen when the DENO path is suffixed to the system PATH instead of "prefixing".

Thanks - good argument. 

I'm ok with something like this:
```
deno install --delete foo
```

stefnotch:
I'd absolutely prefer `deno uninstall` over adding a delete flag to an install command. The latter seems like a odd hack and is going to cause a lot of googling and confusion for newcomers.

lem0nify:
Return `uninstall` subcommand!

[...]

PR #12209, opened 2021-09-24
feat(cli/uninstall): Add uninstall command

sylc:
Fix https://github.com/denoland/deno/issues/3139

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -437,7 +437,7 @@ pub fn flags_from_vec(args: Vec<String>) -> clap::Result<Flags> {
     install_parse(&mut flags, m);
   } else if let Some(m) = matches.subcommand_matches("uninstall") {
     uninstall_parse(&mut flags, m);
-  }else if let Some(m) = matches.subcommand_matches("completions") {
+  } else if let Some(m) = matches.subcommand_matches("completions") {
     completions_parse(&mut flags, m);
   } else if let Some(m) = matches.subcommand_matches("test") {
     test_parse(&mut flags, m);
@@ -1954,10 +1954,7 @@ fn uninstall_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
 
   let name = cmd[0].to_string();
 
-  flags.subcommand = DenoSubcommand::Uninstall(UninstallFlags {
-    name,
-    root,
-  });
+  flags.subcommand = DenoSubcommand::Uninstall(UninstallFlags { name, root });
 }
 
 fn lsp_parse(flags: &mut Flags, _matches: &clap::ArgMatches) {
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1946,14 +1946,7 @@ fn uninstall_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
     None
   };
 
-  let name_values = matches.values_of("name").unwrap();
-  let mut cmd = vec![];
-  for value in name_values {
-    cmd.push(value.to_string());
-  }
-
-  let name = cmd[0].to_string();
-
+  let name = matches.value_of("name").unwrap().to_string();
   flags.subcommand = DenoSubcommand::Uninstall(UninstallFlags { name, root });
 }
 
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -111,6 +111,12 @@ pub struct InstallFlags {
 }
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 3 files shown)


sylc:
I can't find any instruction in the contributing guide on how to run the tests locally.
Also the format is failing with 
image

🤔 Any ideas ?

cli/flags.rs
```diff
@@ -1896,6 +1936,30 @@ fn install_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
...
+    None
+  };
+
+  let name_values = matches.values_of("name").unwrap();
+  let mut cmd = vec![];
+  for value in name_values {
+    cmd.push(value.to_string());
+  }
+
+  let name = cmd[0].to_string();
```
crowlKats, line 1955: why getting multiple values?

bartlomieju:
> I can't find any instruction in the contributing guide on how to run the tests locally.

Make sure your git submodules are up to date: `git submodule update --init --recursive`

> Should this be committed ?

No, the problem should be fixes once git submodules are up to date.

cli/tools/installer.rs
```diff
@@ -139,6 +139,47 @@ pub fn infer_name_from_url(url: &Url) -> Option<String> {
...
+
+  let mut removed = false;
+
+  if file_path.exists() {
+    fs::remove_file(&file_path)?;
+    println!("deleted {}", file_path.to_string_lossy());
+    removed = true
+  };
+
+  if cfg!(windows) {
```
dsherret, line 167: Minor point, but could this copy what's done for install and instead...

```rust
if cfg!(windows) {
 file_path = file_path.with_extension("cmd");
}
```

...could be done above? It also looks like the logic for getting the installation dir and file path could be shared between `install` and `uninstall`.

cli/tools/installer.rs
```diff
@@ -139,6 +139,47 @@ pub fn infer_name_from_url(url: &Url) -> Option<String> {
...
+  }
+
+  if !removed {
+    return Err(generic_error(format!("No installation found for {}", name)));
+  }
+  println!("✅ Successfully uninstalled {}", name);
+  Ok(())
+}
+
 pub fn install(
```
dsherret, line 183: In the install command there are some `extra_files`. I think those need to be checked for and removed in the uninstall command.

cli/tools/installer.rs
dsherret: Nevermind. Just saw and remembered that this is necessary because it outputs a shell script on windows.

cli/tools/installer.rs
sylc: I have updated to clean the extra files.

cli/flags.rs
```diff
@@ -1896,6 +1936,20 @@ fn install_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
   });
 }
 
+fn uninstall_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
+  runtime_args_parse(flags, matches, true, true);
```
crowlKats, line 1940:
```suggestion
```
No need for runtime args parsing

sylc: done.

cli/tools/installer.rs
```diff
@@ -926,4 +977,32 @@ mod tests {
...
+
+    uninstall("echo_test".to_string(), Some(temp_dir.path().to_path_buf()))
+      .expect("Uninstall failed");
+
+    assert!(!file_path.exists());
+    assert!(!file_path.with_extension("tsconfig.json").exists());
+    assert!(!file_path.with_extension("lock.json").exists());
+
+    if cfg!(windows) {
+      file_path = file_path.with_extension("cmd");
```
dsherret, line 1004: Small point, but it is missing the creation of this file on windows.

dsherret (approved):
LGTM. Thanks @sylc! We will merge this once the 1.15 merge window opens.

cli/tools/installer.rs
```diff
@@ -139,6 +139,57 @@ pub fn infer_name_from_url(url: &Url) -> Option<String> {
...
+    canonicalize_path(&root)?
+  } else {
+    get_installer_root()?
+  };
+  let installation_dir = root.join("bin");
+
+  // ensure directory exists
+  if let Ok(metadata) = fs::metadata(&installation_dir) {
+    if !metadata.is_dir() {
+      return Err(generic_error("Installation path is not a directory"));
```
dsherret, line 153: Another minor point: It might be useful to the user to include the resolved installation_dir path in this error message.

PR #12215, opened 2021-09-24
chore(workers): Test that closing a worker closes any child workers

andreubotella:
Before #12156, closing a worker which had children would cause a panic (https://github.com/denoland/deno/issues/11342#issuecomment-918327693). After that PR, closing a worker will also close any child workers.

### DIFF
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -1194,6 +1194,11 @@ itest!(worker_drop_handle_race {
   exit_code: 1,
 });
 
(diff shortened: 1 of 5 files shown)


piscisaureus (approved):
LGTM. Thanks for the test, @andreubotella.

bartlomieju:
> After that PR, closing a worker will also close any child workers.

I need to go through the spec once more, but I believe this is actually where Deno diverges from the spec. IIRC in this case the children workers of the closed workers should be kept alive, but they should instead start targeting closed worker's parent.

PR #12216, opened 2021-09-24
fix(http): panic when responding to a closed conn

AaronO:
Our oneshot receiver in `HyperService::call` would unwrap and panic, the `.await` on the oneshot receiver errors when the sender is dropped.

The sender is dropped in `op_http_response` because:
1. We take `ResponseSenderResource` (which owns the sender)
2. Then we hit an early exit in `op_http_response` due to the conn being closed
3. Then the taken sender gets dropped in this early exit before any response is sent over the channel

Fallbacking to returning a dummy response to hyper seems to be a fine quickfix

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -959,7 +959,9 @@ unitTest(
 
       try {
         http.close();
-      } catch {}
+      } catch {
+        "nop";
+      }
 
       listener.close();
     }
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -936,3 +936,46 @@ unitTest(
     await promise;
   },
 );
+
+// https://github.com/denoland/deno/pull/12216
+unitTest(
+  { permissions: { net: true } },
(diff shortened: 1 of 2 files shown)


piscisaureus:
Fix LGTM, is it possible to add a test for this?

A stand-alone test is here: https://github.com/denoland/deploy/issues/1949#issuecomment-926873141

piscisaureus (approved):
It seems hard to reliably trigger just this bug without sometimes triggering another issue.
I don’t want to land flaky tests.
And I like bugs to be fixed one at a time.
Therefore I think we can ship this as-is.

ry (changes requested):
need a test

AaronO:
> We should have a test. Even a flaky test is better than none - tho a solid one would be great.

I've added a non-flaky test.

You can run it yourself:
```
deno run -A https://gist.githubusercontent.com/AaronO/fc7fe0aed81afa4706218df7fd97b62e/raw/684851feaeb0216d592400c882a439645648fa53/deno_sender_panic_reprod.js
```

ry (approved):
LGTM - nice work!

PR #12219, opened 2021-09-25
feat(tls): custom in memory CA certificates

### LINKED ISSUE #11608: Proposal: replace `certFile` in Deno.ConnectTlsOptions with `caCerts`
The naming of the `certFile` argument in `Deno.ConnectTlsOptions` is horrendous because `certFile` is ambiguous if it is a CA certificate, or a client mTLS cert. It should be `caFile`. Really it should be `caData` though, with it being the users responsibility to read the file from disk. Even more really it should be `caCerts`, taking an array of certificates (as string, encoded as PEM) to add to the trust store.

Client certificates for mTLS should end up in the `certChain` and `privateKey` fields of the `Deno.ConnectTlsOptions`, both taking a string encoded as PEM.

Because the removal of `certFile` would be a breaking change, let's deprecate the field in the next release (1.13) using `@deprecated`, and introduce the `caCerts` option.

The same change should apply to `createHttpClient` and `startTls`.

Any objections?

lucacasonato:
This adds support for using in memory CA certificates for
`Deno.startTLS`, `Deno.connectTLS` and `Deno.createHttpClient`.

`certFile` is deprecated in `startTls` and `connectTls`, and removed
from `Deno.createHttpClient`.

Closes #11608.

### DIFF
--- a/ext/fetch/lib.rs
+++ b/ext/fetch/lib.rs
@@ -40,8 +40,6 @@ use serde::Serialize;
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::convert::From;
-use std::fs::File;
-use std::io::Read;
 use std::path::Path;
 use std::path::PathBuf;
 use std::pin::Pin;
@@ -87,7 +85,7 @@ pub fn init<P: FetchPermissions + 'static>(
         create_http_client(
           user_agent.clone(),
           root_cert_store.clone(),
-          None,
+          vec![],
           proxy.clone(),
           unsafely_ignore_certificate_errors.clone(),
           client_cert_chain_and_key.clone(),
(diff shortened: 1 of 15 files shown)


ext/fetch/lib.rs
```diff
@@ -467,9 +465,8 @@ impl HttpClientResource {
 #[serde(rename_all = "camelCase")]
 #[serde(default)]
 pub struct CreateHttpClientOptions {
-  ca_stores: Option<Vec<String>>,
-  ca_file: Option<String>,
-  ca_data: Option<ByteString>,
+  #[serde(default)]
```
bnoordhuis, line 468: I admit I don't understand what this changes. It's also specified on the type.

bnoordhuis (approved):
LGTM

Not sure why `cli/tests/testdata/045_programmatic_proxy_client.ts` fails with `Uncaught TypeError: Error parsing args: serde_v8 error: ExpectedArray`.

serde_v8 bug? Does it expect an `Option<Vec<String>>`?

ext/fetch/lib.rs
lucacasonato: It should force `caCerts: undefined` to be deserialized as `caCerts: []`. Essentially use `Default::default` when the field is falsy or not present.

Issue #12221, opened 2021-09-25
Unhandled rejections in workers should not fire an `error` event on the `Worker` object


andreubotella:
The HTML spec defines steps for reporting runtime script errors (with the `error` event) and unhandled promise rejections (with the `unhandledrejection` event). Both mechanisms are fairly similar, but they differ on what happens when an error or rejection happens inside a dedicated worker and is not caught by the corresponding event. Unhandled promise rejections will simply be reported to the console, as they would in a window agent, and never make it outside of the worker – while script errors will fire an `error` event to the corresponding `Worker` object in the parent agent.

Deno doesn't (yet?) implement the `unhandledrejection` event, and an unhandled rejection on the main thread will kill the process. However, an unhandled rejection in a worker will instead fire an `error` event on the corresponding `Worker` object – which shouldn't happen per the spec. Arguably the process should be killed, as it would if it happened on the main thread.

This is causing some worker tests to hang forever.

-----------------

Here's a test to compare the behavior in Deno and in browsers.

```js
// main.js
const worker = new Worker(new URL("./worker.js", import.meta.url), {type: "module"});

worker.addEventListener("error", () => {
    console.log("worker.onerror in main thread");
});

window.addEventListener("error", () => {
    console.log("window.onerror in main thread");
});

window.addEventListener("unhandledrejection", () => {
    console.log("window.onunhandledrejection in main thread");
});
```

```js
// worker.js
self.addEventListener("error", () => {
    console.log("self.onerror in worker");
});

self.addEventListener("unhandledrejection", () => {
    console.log("self.onunhandledrejection in worker");
});

(async () => {
    throw new Error("Error");
})();
```

All browsers fire the `unhandledrejection` event inside the worker, report the error, and don't fire any event in the main thread:
![image](https://user-images.githubusercontent.com/8225977/134781239-147a58c0-1bba-4379-8a53-77752a18af5f.png)

While Deno will report the error inside the worker, fire the `error` event on the main thread's `worker` object, and then die because the error wasn't handled:
![image](https://user-images.githubusercontent.com/8225977/134781325-bce2e8ec-06ea-4fe0-8bc1-6d1c4250483c.png)

(This screenshot uses Deno canary – 1.14.1 would panic because of #11342.)

nayeemrmn:
Browsers on unhandled errors in workers will:
1. Print the error.
2. Dispatch an error event within the worker (`"error"` for sync and `"unhandledrejection"` for async).
[...]

Issue #9460, opened 2021-02-09
Add timeouts to web platform tests


caspervonb:
Tried running a `wpt update` and it eventually got stuck in fetch.

```
/fetch/api/abort/general.any.html

test Request objects have a signal property ... failed
test Signal state is cloned ... failed
test Clone aborts with original controller ... failed
test Aborting rejects with AbortError ... failed
test Aborting rejects with AbortError - no-cors ... failed
test TypeError from request constructor takes priority - RequestInit's window is not null ... ok
test TypeError from request constructor takes priority - Input URL is not valid ... failed
test TypeError from request constructor takes priority - Input URL has credentials ... ok
test TypeError from request constructor takes priority - RequestInit's mode is navigate ... ok
test TypeError from request constructor takes priority - RequestInit's referrer is invalid ... ok
test TypeError from request constructor takes priority - RequestInit's method is invalid ... ok
test TypeError from request constructor takes priority - RequestInit's method is forbidden ... ok
test TypeError from request constructor takes priority - RequestInit's mode is no-cors and method is not simple ... ok
test TypeError from request constructor takes priority - RequestInit's cache mode is only-if-cached and mode is not same-origin ... ok
test TypeError from request constructor takes priority - Request with cache mode: only-if-cached and fetch mode cors ... ok
test TypeError from request constructor takes priority - Request with cache mode: only-if-cached and fetch mode no-cors ... ok
test TypeError from request constructor takes priority - Bad referrerPolicy init parameter value ... ok
test TypeError from request constructor takes priority - Bad mode init parameter value ... ok
test TypeError from request constructor takes priority - Bad credentials init parameter value ... ok
test TypeError from request constructor takes priority - Bad cache init parameter value ... ok
test TypeError from request constructor takes priority - Bad redirect init parameter value ... ok
test Signal on request object ... failed
test Signal on request object created from request object ... failed
test Signal on request object created from request object, with signal on second request ... failed
test Signal on request object created from request object, with signal on second request overriding another ... failed
test Signal retained after unrelated properties are overridden by fetch ... failed
test Signal removed by setting to null ... ok
test Already aborted signal rejects immediately ... failed
test Request is still 'used' if signal is aborted before fetching ... ok
test response.arrayBuffer() rejects if already aborted ... failed
test response.blob() rejects if already aborted ... failed
test response.formData() rejects if already aborted ... failed
test response.json() rejects if already aborted ... failed
test response.text() rejects if already aborted ... failed
[...]

PR #12222, opened 2021-09-25
chore(wpt): Enable WPT worker tests

andreubotella:
Classic workers were implemented in #11338, which also enabled the WPT tests in the `workers` directory. However, the rest of WPT worker tests were not enabled because a number of them were hanging due to web-platform-tests/wpt#29777. Now that that WPT issue is fixed, the bulk of worker tests can be enabled.

There are still a few tests that hang, and so haven't been enabled. In particular:

- The following tests seem to hang because a promise fails to resolve. We can detect such cases in non-worker tests because the process will exit without calling the WPT completion callback, but in worker tests the worker message ops will keep the event loop running. This will be fixed when we add timeouts to WPT tests (#9460).

  - `/fetch/api/basic/error-after-response.any.worker.html`
  - `/html/webappapis/microtask-queuing/queue-microtask-exceptions.any.worker.html`
  - `/webmessaging/message-channels/worker-post-after-close.any.worker.html`
  - `/webmessaging/message-channels/worker.any.worker.html`
  - `/websockets/Create-on-worker-shutdown.any.worker.html`

- The following tests apparently hang because a promise rejection is never handled, which will kill the process in the main thread but not in workers (#12221).

  - `/streams/readable-streams/async-iterator.any.worker.html`
  - `/workers/interfaces/WorkerUtils/importScripts/report-error-setTimeout-cross-origin.sub.any.worker.html`
  - `/workers/interfaces/WorkerUtils/importScripts/report-error-setTimeout-redirect-to-cross-origin.sub.any.worker.html`
  - `/workers/interfaces/WorkerUtils/importScripts/report-error-setTimeout-same-origin.sub.any.worker.html`


andreubotella:
The WASM tests that pass unexpectedly are related to #12263 – those tests panicked on my machine, and they also panic on CI, but the test runner seems to detect them as successes. I'm assuming that, since the panic happens in the worker thread when the process is just about to close anyway, maybe the exit code hasn't had time to be propagated or something?

andreubotella:
This can now be reviewed.

cc @lucacasonato

bartlomieju:
Please rebase @andreubotella

PR #12233, opened 2021-09-26
refactor(fetch/headers): optimize & simplify trimming

AaronO:


andreubotella:
The fetch spec's ["normalize"](https://fetch.spec.whatwg.org/#concept-header-value-normalize) operation isn't equivalent to `String.prototype.trim()`:

```
> let headers = new Headers({"a": "\ftest\f"})
undefined
> headers.get("a")
"\ftest\f"
> headers.get("a").trim()
"test"
```

The WPT [normalization tests](https://github.com/web-platform-tests/wpt/blob/master/fetch/api/headers/headers-normalize.any.js) doesn't seem to test for this, though.

AaronO:
> The fetch spec's ["normalize"](https://fetch.spec.whatwg.org/#concept-header-value-normalize) operation isn't equivalent to `String.prototype.trim()`:
> 
> ```
> > let headers = new Headers({"a": "\ftest\f"})
> undefined
> > headers.get("a")
> "\ftest\f"
> > headers.get("a").trim()
> "test"
> ```
> 
> The WPT [normalization tests](https://github.com/web-platform-tests/wpt/blob/master/fetch/api/headers/headers-normalize.any.js) doesn't seem to test for this, though.

I don't think it's very harmful to be stricter than the spec and normalize to a subrange (especially if it's not tested by WPT), ultimately we'll see what @lucacasonato thinks. In practice I wouldn't imagine anyone passes around `\f` in their headers

lucacasonato:
I disagree. This is important. If there are no WPT, we should add some 😉. Intentionally deviating from spec for things like this isn't good.

lucacasonato:
WPT has been added now: https://github.com/web-platform-tests/wpt/pull/30979

AaronO:
Closed in favor of #12277

PR #12235, opened 2021-09-26
refactor(fetch/response): inline defaultInnerResponse

AaronO:
Not useful to have the defaults externally defined when they're only used in `newInnerResponse()`. Also match order in `newInnerResponse()` and `cloneInnerResponse`

### DIFF
--- a/ext/fetch/23_response.js
+++ b/ext/fetch/23_response.js
@@ -101,37 +101,33 @@
       type: response.type,
       body,
       headerList,
-      url() {
-        if (this.urlList.length == 0) return null;
-        return this.urlList[this.urlList.length - 1];
-      },
       urlList,
       status: response.status,
       statusMessage: response.statusMessage,
       aborted: response.aborted,
+      url() {
+        if (this.urlList.length == 0) return null;
+        return this.urlList[this.urlList.length - 1];
+      },
     };
   }
-
-  const defaultInnerResponse = {
-    type: "default",
-    body: null,
-    aborted: false,
-    url() {
-      if (this.urlList.length == 0) return null;
-      return this.urlList[this.urlList.length - 1];
-    },
-  };
 
   /**
    * @returns {InnerResponse}
    */
   function newInnerResponse(status = 200, statusMessage = "") {
     return {
+      type: "default",
+      body: null,
       headerList: [],
       urlList: [],
       status,
       statusMessage,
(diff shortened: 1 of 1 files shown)


lucacasonato (commented):
I orignially "un-inlined" it because my measurements showed this was slightly faster for HTTP throughput. Is this not the case anymore?

AaronO:
> I orignially "un-inlined" it because my measurements showed this was slightly faster for HTTP throughput. Is this not the case anymore?

I haven't re-benched this change in isolate but I'm decently confident the old version is slower

lucacasonato:
Well please re-bench. I specifically introduced this performance optimisation in https://github.com/denoland/deno/pull/10203/commits/c246ca2a79ff156677a67fc790126faad5bd1249 after extensive benching as part of https://github.com/denoland/deno/pull/10203.

AaronO:
> Well please re-bench. I specifically introduced this performance optimisation in [c246ca2](https://github.com/denoland/deno/commit/c246ca2a79ff156677a67fc790126faad5bd1249) after extensive benching as part of #10203.
> 
> And while you are at it: if it does turn out to be faster still, please add a comment about it :-)

Bench script https://gist.github.com/AaronO/2eb62129ef066312acb5feb0900d10c1

## Results

```
❯ deno run https://gist.githubusercontent.com/AaronO/2eb62129ef066312acb5feb0900d10c1/raw/53f95d350c835702d7094dd164bc69bcc1dd7436/bench_inner_inline.js
inner_outer:         	n = 1000000, dt = 0.079s, r = 12658228/s, t = 79ns/op
inner_inline:        	n = 1000000, dt = 0.017s, r = 58823529/s, t = 17ns/op
inner_outer:         	n = 1000000, dt = 0.084s, r = 11904762/s, t = 84ns/op
inner_inline:        	n = 1000000, dt = 0.016s, r = 62500000/s, t = 16ns/op
inner_outer:         	n = 1000000, dt = 0.082s, r = 12195122/s, t = 82ns/op
inner_inline:        	n = 1000000, dt = 0.017s, r = 58823529/s, t = 17ns/op
```

## Conclusion

Inlined version is ~5x faster

lucacasonato (approved):
Benchmarks speak louder than words 😅

My original speedup might have come from the switch from getter to function. Who knows how V8 optimizes those.

PR #12239, opened 2021-09-26
Maybe fix io flush

caspervonb:
This flush should not be needed; there was a fixme comment here but it is gone.
This was just introduced to keep CI happy, passes locally but CI doesn't flush.

Just checking if that is still the case with all the Tokyo bumps we've done.


caspervonb:
Rebase disaster 😱

PR #12249, opened 2021-09-27
fix(runtime/js/workers): throw errors instead of using an op

### LINKED ISSUE #11350: Errors in a worker's message event handler aren't reported
```
thread 'worker-0' panicked at 'coding error: either js is polling or the worker is terminated', runtime/web_worker.rs:498:9
    at Worker.#pollControl (deno:runtime/js/11_workers.js:247:23)
```

(In Deno 1.11.5 the panic message doesn't show up  – see #11342.)

### test.js
```js
const worker = new Worker(
  { type: "module" },
);
worker.postMessage({});
```

### worker.js
```js
self.onmessage = () => {
  // Throws a TypeError: Cannot read property 'run' of null
  null.run();
};
```

-----------

Running `null.run()` in the global scope, in a timer task, or in a promise reaction all seem to report the error in stderr, but not in a handler for the `message` event.

This was found while investigating the WPT failures in #11338.

nayeemrmn:
Fixes #11350.

### DIFF
--- a/cli/tests/integration/worker_tests.rs
+++ b/cli/tests/integration/worker_tests.rs
@@ -20,6 +20,20 @@ itest!(worker_nested_error {
   exit_code: 1,
 });
 
+itest!(worker_async_error {
+  args: "run -A --quiet --reload workers/worker_async_error.ts",
+  output: "workers/worker_async_error.ts.out",
+  http_server: true,
+  exit_code: 1,
+});
+
+itest!(worker_message_handler_error {
+  args: "run -A --quiet --reload workers/worker_message_handler_error.ts",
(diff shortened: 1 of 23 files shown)


bartlomieju:
I think addressing [this comment](https://github.com/denoland/deno/issues/11350#issuecomment-928166502) is needed as well.

nayeemrmn:
@andreubotella This makes workers terminate on any unhandled error, including async and message handler ones. Message handler error terminations can still be prevented with a `self.onerror = () => true` within the worker.

bartlomieju (approved):
LGTM

I agree with @andreubotella that this is a good stepping stone but will require further iterations.

PR #12250, opened 2021-09-27
fix: worker environment permissions should accept an array

dsherret:
Fixes a Windows only clippy lint, but also aligns the worker environment permissions with accepting an array.

### DIFF
--- a/cli/tests/unit/worker_types.ts
+++ b/cli/tests/unit/worker_types.ts
@@ -5,7 +5,7 @@ unitTest(
   { permissions: { read: true } },
   function utimeSyncFileSuccess() {
     const w = new Worker(
-      new URL("../workers/worker_types.ts", import.meta.url).href,
+      new URL("../testdata/workers/worker_types.ts", import.meta.url).href,
       { type: "module" },
     );
     assert(w);
--- a/runtime/ops/worker_host.rs
+++ b/runtime/ops/worker_host.rs
@@ -223,7 +223,10 @@ fn merge_env_permission(
 ) -> Result<UnaryPermission<EnvDescriptor>, AnyError> {
   if let Some(worker) = worker {
     if (worker.global_state < main.global_state)
-      || !worker.granted_list.iter().all(|x| main.check(&x.0).is_ok())
+      || !worker
+        .granted_list
+        .iter()
+        .all(|x| main.check(x.as_ref()).is_ok())
     {
       return Err(custom_error(
         "PermissionDenied",
@@ -451,7 +454,7 @@ where
     granted_list: value
       .paths
       .into_iter()
-      .map(|env| EnvDescriptor(env.to_uppercase()))
+      .map(EnvDescriptor::new)
       .collect(),
     ..Default::default()
   }))
--- a/runtime/ops/worker_host.rs
+++ b/runtime/ops/worker_host.rs
@@ -451,11 +451,7 @@ where
 
   Ok(Some(UnaryPermission::<EnvDescriptor> {
     global_state: value.global_state,
-    granted_list: value
-      .paths
-      .into_iter()
-      .map(EnvDescriptor::new)
-      .collect(),
+    granted_list: value.paths.into_iter().map(EnvDescriptor::new).collect(),
     ..Default::default()
   }))
 }
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -162,7 +162,7 @@ impl UnitPermission {
 /// A normalized environment variable name. On Windows this will
 /// be uppercase and on other platforms it will stay as-is.
 #[derive(Clone, Eq, PartialEq, Hash, Debug, Default)]
-pub struct EnvVarName {
+struct EnvVarName {
   inner: String,
 }
 
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -160,26 +160,23 @@ impl UnitPermission {
   }
 }
 
-#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct UnaryPermission<T: Eq + Hash> {
-  #[serde(skip)]
   pub name: &'static str,
-  #[serde(skip)]
   pub description: &'static str,
   pub global_state: PermissionState,
   pub granted_list: HashSet<T>,
   pub denied_list: HashSet<T>,
-  #[serde(skip)]
   pub prompt: bool,
 }
 
-#[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize)]
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default)]
 pub struct ReadDescriptor(pub PathBuf);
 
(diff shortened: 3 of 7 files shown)


runtime/ops/worker_host.rs
```diff
@@ -451,7 +454,7 @@ where
     granted_list: value
       .paths
       .into_iter()
-      .map(|env| EnvDescriptor(env.to_uppercase()))
```
dsherret, line 454: Seems like a bug? I think this `.to_uppercase()` should only happen on Windows like the rest of the code does.

runtime/permissions.rs
```diff
@@ -204,7 +205,33 @@ impl fmt::Display for NetDescriptor {
 }
 
 #[derive(Clone, Eq, PartialEq, Hash, Debug, Default, Deserialize)]
-pub struct EnvDescriptor(pub String);
+pub struct EnvDescriptor {
+  env: String,
+}
```
dsherret, line 210: Am I breaking some deserialization with this change?

runtime/ops/worker_host.rs
bartlomieju: I believe you are correct

runtime/permissions.rs
bartlomieju: You can still make that work, by just removing `pub` modifier before the `String` type?

dsherret: I guess because this is mostly constructed in this file there wouldn't be anything preventing someone from still doing `EnvDescriptor(...)`. Writing `EnvDescriptor { env: ... }` is harder to do by accident.

bartlomieju: Sounds good

crowlKats: The other Descriptors are all a tuple struct, even the ones that have a "constructor" function. please keep consistency. dont see much point for this change...

dsherret: @crowlKats it's to prevent accidentally writing `EnvDescriptor(...)`, which would be a bug. Maybe we could move this out to a separate file to prevent this module from doing that?

crowlKats: > it's to prevent accidentally writing EnvDescriptor(...), which would be a potential bug if they didn't uppercase on windows.

You can keep it a tuple struct and just remove the `pub` if you really don't want people to use direct initialization of the struct. but even for that i see little point. Also if you think changing from a tuple struct to an "object" struct would make people less error prone, then you are mistaken. 
If people don't know the code they are working with, then that is on them. What we could and should do is document things, that would actually be a helpful solution to prevent misunderstandings and wrong usage of code.

dsherret: > You can keep it a tuple struct and just remove the pub if you really don't want people to use direct initialization of the struct.

See above. It would still be accessible within this module.

>  Also if you think changing from a tuple struct to an "object" struct would make people less error prone, then you are mistaken.

It is less error prone to accidentally write in this module. It stands out like a sore thumb to type or see `EnvDescriptor { env: ... }`. The difference between `EnvDescriptor(...)` and `EnvDescriptor::new(...)` is not as obvious.

I don't see the big deal here as we are talking about private API to this module (would be private API if we don't use `pub` on a tuple struct) and making it a object struct is not a foolproof, but a bug mitigation measure.

Honestly, I would rather change all these structs to be object structs so they describe their properties. Tuple structs aren't great. For example looking at `pub struct NetDescriptor(pub String, pub Option<u16>);`... you have to look at the implementation to find out that the second argument is a port.

caspervonb:
Exactly what are the upside to complicating the structs so much?

dsherret:
> Exactly what are the upsides to complicating the structs so much?

@caspervonb eliminates the chance of someone forgetting to call `.to_uppercase()` on Windows or calling `.to_uppercase()` when not necessary (as the code was incorrectly doing) by extracting out that duplicated functionality to a common place and forcing it to be used. Also, only one struct was changed.

caspervonb:
> > Exactly what are the upsides to complicating the structs so much?
> 
> @caspervonb eliminates the chance of someone forgetting to call `.to_uppercase()` on Windows or calling `.to_uppercase()` when not necessary (as the code was incorrectly doing) by extracting out that duplicated functionality to a common place and forcing it to be used. Also, only one struct was changed.

Yeah am including the talk of NetDescriptor here as-well.
Seems like just making the tuple internally private and adding a factory method would have the same effect without the extra levels of indirection.

dsherret:
@caspervonb my concern with only adding a factory function, is that it's too easy to accidentally call `EnvDescriptor(env)` with a string instead of `EnvDescriptor::new(env)`, so this change lowers the risk of accidentally doing that (because calling `EnvDescriptor(EnvVarName { inner: env })` is very easy to spot as being wrong, even though it's technically possible). The `EnvDescriptor` tuple is mostly constructed within the same module, so I think it would need to be moved into another module in order to be private and prevent that. This change introduces another struct to hold an environment variable name so that it gets properly changed based on the platform. Additionally, perhaps in the future we could extract out this struct to be used in other places we use environment variable names and need to ensure this functionality occurs.

On `NetDescriptor`, not pushing to change that struct here (conversation is resolved).

caspervonb:
> @caspervonb my concern with only adding a factory function, is that it's too easy to accidentally call `EnvDescriptor(env)` with a string instead of `EnvDescriptor::new(env)`, so this change lowers the risk of accidentally doing that (because calling `EnvDescriptor(EnvVarName { inner: env })` is very easy to spot as being wrong, even though it's technically possible). The `EnvDescriptor` tuple is mostly constructed within the same module, so I think it would need to be moved into another module in order to be private and to prevent the tuple initializer being used.
> 
> Additionally, perhaps in the future we could extract out this `EnvVarName` struct to be used in other places we use environment variable names and need to ensure this functionality occurs.
> 
> On `NetDescriptor`, not pushing to change that struct here (conversation is resolved).

Seems overly redundant to me, private with a constructor is a pretty clear signal for the intended use.
All types can be incorrectly constructed internally within the same module.
This is mostly @crowlKats's code tho so I'll defer to him.

If there's a windows only bug fix here tho, test? 🙂

cli/tests/unit/worker_types.ts
```diff
@@ -5,7 +5,7 @@ unitTest(
   { permissions: { read: true } },
   function utimeSyncFileSuccess() {
     const w = new Worker(
-      new URL("../workers/worker_types.ts", import.meta.url).href,
+      new URL("../testdata/workers/worker_types.ts", import.meta.url).href,
```
dsherret, line 8: I'm not sure what this test does. It seems like I can put any path here and it will pass.

dsherret:
@caspervonb thanks for recommending to add a test as it seems providing an array to `env` on a worker permission did not work at all.

Again, there is not enough difference between `EnvDescriptor(...)` and `EnvDescriptor::new(...)` when only looking at the call site (as is often done in reviews and when writing code) to prevent accidental use when they both accept the same argument. There is a huge difference between `EnvDescriptor::new(...)` and `EnvDescriptor(EnvVarName { inner: env })`. You really have to go out of your way to type that long form.

bartlomieju (approved):
LGTM, concerning that this never worked properly.

PR #12254, opened 2021-09-28
1.14.2

AaronO:


Releases.md
```diff
@@ -6,6 +6,33 @@ https://github.com/denoland/deno/releases
 We also have one-line install commands at:
 https://github.com/denoland/deno_install
 
+### 1.14.2 / 2021.09.28
+
+- ci: enable cache for fastci (#12186)
+- ci: experiment mac & windows debug builds (#11884)
```
satyarohith, line 12:
```suggestion
```
I'm not against including them, but I feel like `ci` commits are less relevant to the end-user same as the `refactor, test` commits?

satyarohith (approved):
LGTM.

Releases.md
ry: FYI @AaronO - the Releases.md is not just a list of all the git commits. We edit the history to provide a reader friendly digest of what happened in the release. For example, if a change was made and then reverted, it will not be mentioned. Doc changes, "chore" changes, and "refactor" changes are often excluded (unless particularly relevant for some reason).

PR #12260, opened 2021-09-28
fix(http): don't expose body on GET/HEAD requests

### LINKED ISSUE #11927: new Request(request) throws with hyper + http/2
`new Request(request)` throws a `Request with GET/HEAD method cannot have body` exception that originates in ext/fetch/23_request.js, but only for http/2 requests.

I have reason to believe the size_hint check below is why it works for http/1 but not http/2:
https://github.com/denoland/deno/blob/b7c2902c9752e83dc467ce6a812dab4726f200d9/ext/http/lib.rs#L236-L240

lucacasonato:
GET/HEAD requests can't have bodies according to `fetch` spec. This
commit changes the HTTP server to hide request bodies for requests with
GET or HEAD methods.

Fixes #11927

### DIFF
--- a/ext/http/01_http.js
+++ b/ext/http/01_http.js
@@ -95,7 +95,12 @@
       let body = null;
       if (typeof requestRid === "number") {
         SetPrototypeAdd(this.managedResources, requestRid);
-        body = createRequestBodyStream(this, requestRid);
+        // There might be a body, but we don't expose it for GET/HEAD requests.
+        // It will be closed automatically once the request has been handled and
+        // the response has been sent.
+        if (method !== "GET" && method !== "HEAD") {
+          body = createRequestBodyStream(this, requestRid);
+        }
       }
 
       const innerRequest = newInnerRequest(
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -30,6 +30,7 @@ use hyper::http;
 use hyper::server::conn::Http;
 use hyper::service::Service as HyperService;
 use hyper::Body;
+use hyper::Method;
 use hyper::Request;
 use hyper::Response;
 use serde::Deserialize;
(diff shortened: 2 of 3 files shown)


ahabhgk (approved):

ext/http/01_http.js
```diff
@@ -95,7 +95,12 @@
       let body = null;
       if (typeof requestRid === "number") {
         SetPrototypeAdd(this.managedResources, requestRid);
-        body = createRequestBodyStream(this, requestRid);
+        // There might be a body, but we don't expose it for GET/HEAD requests.
```
piscisaureus, line 98: Is it possible to refer to the relevant portion(s) of the spec here?

lucacasonato: This is the `Request` creation that is internal to our HTTP server, so there is no spec. The `new Request` constructor that is exposed to users enforces this in step 34 of the constructor steps: https://fetch.spec.whatwg.org/#dom-request step 34

piscisaureus: > This is the `Request` creation that is internal to our HTTP server, so there is no spec.

Then why do we have to hide bodies on incoming requests? 
And if we do, what about other methods that can't have bodies - `OPTIONS` and `TRACE` (`DELETE`?).

lucacasonato: Because users can not do `new Request(incomingReq)` (this throws) for an unmodified incoming GET request from the HTTP server. See #11927.  We are creating an invalid `Request` instance because we are creating it internally, bypassing the usual request constructor checks. There is no possibility on the web platform to create a `Request` object where `method === "GET" && body !== null`. Because of this we should not have one either.

`OPTIONS`, `TRACE`, and `DELETE` requests can have bodies in `fetch` on the web platform. `GET` and `HEAD` can not.

ext/http/lib.rs
```diff
@@ -244,13 +245,17 @@ fn prepare_next_request(
   let url = req_url(&req, scheme, addr)?;
 
   let is_websocket = is_websocket_request(&req);
-  let has_body = if let Some(exact_size) = req.size_hint().exact() {
+  let maybe_has_body = if let Some(exact_size) = req.size_hint().exact() {
```
bnoordhuis, line 248: Maybe `can_have_body`? The `maybe_foo` idiom is normally used to indicate it's an Option. Using it for a boolean is kind of confusing.

I'd perhaps flip the logic a little so size_hint() is only called when it's needed. E.g.:
```rust
let can_have_body = !matches!(*req.method(), Method::GET | Method::HEAD);
let has_body = is_websocket || can_have_body && Some(0) != req.size_hint().exact();
```
Not that the current code is wrong or inefficient but you have to pay close attention to the logic to figure out when its result is actually used.

bnoordhuis (commented):
I have no opinion on what the fetch() spec says but HTTP in general allows GET requests with bodies.

Not that they're very useful - the server isn't allowed to attach semantic meaning to them, what with GET being idempotent - but it's a thing.

ext/http/lib.rs
lucacasonato: Yeah, that makes sense. Updated to your logic.

bnoordhuis (approved):

PR #12265, opened 2021-09-29
perf(web/Event): move last class field to constructor

AaronO:


AaronO:
> What does this speed up?

Initializing attributes via class field expressions has additional overhead compared to a simple this assignment in the constructor, the difference is non trivial and matters for hot paths. I had already moved other attributes to this pattern but had missed this one since it was set at the end of the class' code

AaronO:
> What does this speed up?

This helps to optimize the usage of FetchEvent in deploy and deployctl

ry (approved):

robpalme:
The lack of optimization in V8 for class field initializers was fixed over a year ago in V8 10.0.

So this patch can be reverted now and, based on the results in the linked article, it may even be fractionally faster.

https://v8.dev/blog/faster-class-features

Deno is quoted in the article as a motivation for the V8 optimisation.

PR #12268, opened 2021-09-29
feat: Show the URL of streaming WASM modules in stack traces

### LINKED ISSUE #12151: Stack traces in streaming WASM modules should show the module URL
The [WebAssembly Web API](https://webassembly.github.io/spec/web-api/#conventions) spec has a non-normative section about how WASM stack traces should be displayed. Some of the things it suggests –like reporting the line in JS in which a WASM module was instantiated much like with `eval` calls– don't seem to be supported by V8. But it also suggests that, for WASM modules compiled with `Response`-based APIs (that is, `WebAssembly.compileStreaming()` and `WebAssembly.instantiateStreaming()`), the stack trace should show the URL of the WASM module.

To show that V8 does support this, these are the results of my tests (https://github.com/andreubotella/wasm-stack-trace-test) in Chrome and in Deno:
image
image

I believe this would be easy to hook up once `v8::WasmStreaming`'s `SetUrl` method is exposed in rusty_v8.

andreubotella:
WebAssembly modules compiled through `WebAssembly.compile()` and similar non-streaming APIs don't have a URL associated to them, because they have been compiled from a buffer source. In stack traces, V8 will use a URL such as `wasm://wasm/d1c677ea`, with a hash of the module.

However, wasm modules compiled through streaming APIs, like `WebAssembly.compileStreaming()`, do have a known URL, which can be obtained from the `Response` object passed into the streaming APIs. And as per the [developer-facing display conventions](https://webassembly.github.io/spec/web-api/#conventions) in the WebAssembly Web API spec, this URL should be used in stack traces. This change implements that.

Closes #12151.

-------------------

This PR depends on denoland/rusty_v8#786, and won't run on CI until a release of rusty_v8 containing that commit, and the corresponding serde_v8 release, are published.


andreubotella:
`cli/tests/testdata/unreachable.wasm` is the same as the bytes in `cli/tests/testdata/wasm_unreachable.js`. Should we change the latter to read unreachable.wasm from the file system?

bartlomieju:
> `cli/tests/testdata/unreachable.wasm` is the same as the bytes in `cli/tests/testdata/wasm_unreachable.js`. Should we change the latter to read unreachable.wasm from the file system?

SGTM

andreubotella:
CI is now passing, so this can now be reviewed.

bnoordhuis:
I don't think so. Nothing that can't be trivially fixed anyway.

PR #12269, opened 2021-09-29
fix(ext/http/01_http.js) sonarlint(javascript:S1871)

mynane:
fix(ext/http/01_http.js) sonarlint(javascript:S1871) 

This branch's code block is the same as the block for the branch on line 73. [+1 location]sonarlint(javascript:S1871)


piscisaureus (approved):
LGTM, thanks @mynane!

Issue #12263, opened 2021-09-29
Closing a worker in the reactions of a WASM async operation can panic


andreubotella:
```js
// main.js
new Worker(new URL("./worker.js", import.meta.url), {type: "module"});
```

```js
// worker.js

// The minimum valid wasm module, plus two additional zero bytes.
const buffer = new Uint8Array([0x00, 0x61, 0x73, 0x6D, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00]);
WebAssembly.compile(buffer).catch(err => {
  console.log("Error:", err);
  self.close();
});
```

```
Error: CompileError: WebAssembly.compile(): expected string length @+10
thread 'worker-0' panicked at 'called `Option::unwrap()` on a `None` value', core/runtime.rs:1577:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

-----------

The promise returned by a call to `WebAssembly.compile()` can resolve either synchronously, in which case its reactions will be immediately enqueued in the microtask queue; or asynchronously, in which case the V8 message loop has to be pumped before the reactions are enqueued. It seems like cases where the byte source doesn't match the WASM prefix, or where the byte source is the WASM prefix, are handled synchronously – while cases that need non-trivial parsing of the WASM module are async.

Now, Deno pumps the message queue, followed by performing a microtask checkpoint, at the beginning of each tick of the event loop. In the above case, this will result in `self.close()` being called, which terminates the execution of the worker's isolate, while there are still pending operations in the tick of the event loop. In particular, `JsRuntime::drain_macrotasks()` calls into a JS function and does not expect it to return `None` without there being an exception, resulting in a panic.

I'm not sure if there's anything else in the event loop that runs JS code and might break if the isolate has been terminated. Module loading maybe?

andreubotella:
Changed the title to make this issue more general, since there might be other ways to trigger this other than WASM compilation.

bartlomieju:
I'm going to close this issue with #12270 and #12831 landed.

PR #12270, opened 2021-09-29
fix: Don't panic when a worker is closed in the reactions to a wasm operation.

andreubotella:
Part of #12263.


bartlomieju (approved):
LGTM, nice fix!

PR #12273, opened 2021-09-30
fix(repl): avoid panic when assigned to globalThis

### LINKED ISSUE #9916: bug(cli/repl): panics on assigning to globalThis
When I assign any value to `globalThis` (in REPL), deno panics.

image

kt3k:
This PR fixes the crash of repl described in #9916

Repl internally uses `globalThis` for getting `closed` variable. This PR uses `this` (in global scope) instead of `globalThis` to reference the global variable.

closes #9916

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -135,6 +135,17 @@ fn pty_ignore_symbols() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


dsherret (approved):
Nice! That was easy now. LGTM

kt3k:
@dsherret Thank you for your review!

PR #12274, opened 2021-09-30
feat(ext/ffi): Non-blocking FFI

littledivy:
Adds support for FFI calls that do not block Deno's event loop. Useful for many cases eg: USB transfers, etc.

```typescript
const library = Deno.dlopen(libPath, {
  // Set `nonblocking: true` to make the call async
  "sleep": { parameters: "u64", result: "void", nonblocking: true }
});

// Runs the `sleep` call in a dedicated
// blocking thread and resolve when its done
library.sleep(10000).then(() => console.log("Good morning!"))
// Do stuff
```


ext/ffi/lib.rs
```diff
@@ -374,3 +368,41 @@ fn op_ffi_call(
...
+    .get::<DynamicLibraryResource>(args.rid)?;
+  let symbols = &resource.symbols;
+  let symbol = symbols
+    .get(&args.symbol)
+    .ok_or_else(bad_resource_id)?
+    .clone();
+
+  tokio::task::spawn_blocking(move || ffi_call(args, &symbol))
+    .await
+    .unwrap()
```
ry, line 407: Rather than `op_ffi_call_async` I think it should be a bit more specific `op_ffi_call_nonblocking` for example.

ry (commented):
Seems reasonable and well implemented.

ext/ffi/lib.rs
littledivy: Sure, done.

eliassjogreen (approved):
LGTM

ext/ffi/lib.rs
piscisaureus: I disagree, op_ffi_call_async was perfectly apt.
`op_ffi_call_nonblocking` tells me that the FFI call returns EAGAIN every now and then.

piscisaureus (approved):
LGTM

PR #12277, opened 2021-09-30
perf(web): ~400x faster http header trimming

AaronO:
Use a regex substring match with a first/last char fastpath instead of 2 regex replaces. Roughly ~400x faster (423ms vs 0.7ms in profiled runs)


aapoalas:
LGTM. Only possible improvement I could see is to check do the substring manually as well. Something like this:

```javascript
const needsTrimStart = isHttpWhitespace(str[0]);
const needsTrimEnd = isHttpWhitespace(str[str.length - 1]);
if (!needsTrimStart && !needsTrimEnd) return str;

let startIndex = Number(needsTrimStart);
if (needsTrimStart) {
  startIndex = getFirstNonHttpWhitespaceIndex(startIndex);
}
let endIndex = needsTrimEnd ? getLastNonHttpWhitespaceIndex(str.length - 2) : undefined;
return str.substring(startIndex, endIndex);
```

AaronO:
> LGTM. Only possible improvement I could see is to check do the substring manually as well. Something like this:
> 
> ```js
> const needsTrimStart = isHttpWhitespace(str[0]);
> const needsTrimEnd = isHttpWhitespace(str[str.length - 1]);
> if (!needsTrimStart && !needsTrimEnd) return str;
> 
> let startIndex = Number(needsTrimStart);
> if (needsTrimStart) {
>   startIndex = getFirstNonHttpWhitespaceIndex(startIndex);
> }
> let endIndex = needsTrimEnd ? getLastNonHttpWhitespaceIndex(str.length - 2) : undefined;
> return str.substring(startIndex, endIndex);
> ```

Had considered that too, but I'm happy mainly optimizing for the fast-path of no trimming right now since that should be the the dominant path too

aapoalas:
> Had considered that too, but I'm happy mainly optimizing for the fast-path of no trimming right now since that should be the the dominant path too

Makes sense.

LePichu (approved):

ry (approved):
LGTM

ext/fetch/20_headers.js
```diff
@@ -59,17 +58,7 @@
...
-      HTTP_WHITESPACE_PREFIX_RE,
-      "",
-    );
-    potentialValue = StringPrototypeReplaceAll(
-      potentialValue,
-      HTTP_WHITESPACE_SUFFIX_RE,
-      "",
-    );
-    return potentialValue;
+    return httpTrim(potentialValue);
```
Genarito, line 61: Hi! Sorry about my ignorance, but why not just move the entire content of `httpTrim` to `normalizeHeaderValue` function? That'd remove `httpTrim` definition everywhere which is a lot of lines less with the same result

PR #12278, opened 2021-09-30
feat: web Serial API

crowlKats:


ext/webserial/lib.deno_webserial.d.ts
```diff
@@ -0,0 +1,61 @@
...
+declare interface SerialPortRequestOptions {
+  filters?: SerialPortFilter[];
+}
+
+declare interface SerialPortFilter {
+  usbVendorId?: number;
+  usbProductId?: number;
+}
+
+declare interface SerialPortInfo {
```
DjDeveloperr, line 32: Could port name be added to `SerialPortInfo`? Given even the `--allow-serial` permission works based on serial port names. And without having access to port's name you can't *easily* connect to a specific port by name.

ext/webserial/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[lib]
+path = "lib.rs"
+
+[dependencies]
+deno_core = { version = "0.101.0", path = "../../core" }
+serde = { version = "1.0.129", features = ["derive"] }
+# needed for a type
+serialport = "4.0.1"
+tokio = { version = "1.10.1", features = ["full"] }
+tokio-serial = "5.4.1"
```
bnoordhuis, line 22: The dependency on libudev-sys is somewhat problematic (causes CI to fail because pkg-config can't find libudev, indicative of what users are going to run into) but it's optional and can probably be worked around by forking libudev-sys.

tokio-serial according to its README isn't maintained anymore.

crowlKats: > tokio-serial according to its README isn't maintained anymore.

the README is a bit inaccurate: even though it had that warning for a while, it got still some new releases, also following the issue linked in the README shows that it has been handed over to a new maintainer a few months back

ext/webserial/01_webserial.js
```diff
@@ -0,0 +1,397 @@
...
+  );
+
+  webidl.converters.SerialOptions = webidl.createDictionaryConverter(
+    "SerialOptions",
+    [
+      {
+        key: "baudRate",
+        converter: (V, opts) =>
+          webidl.converters["unsigned long"](V, {
+            ...opts,
```
aapoalas, line 65: Not sure if the safe iterators PR also applies to object spreads. If so, then these'll need primordials as well.

crowlKats: yes, this PR is extremely outdated, soa lot of things here are quite wrong

ext/webserial/01_webserial.js
```diff
@@ -0,0 +1,397 @@
...
+      });
+
+      this[_readable] = stream;
+      return stream;
+    }
+
+    #handleClosingReadable() {
+      this[_readable] = null;
+      if (this[_writable] === null && this[_pendingClosePromise] !== null) {
+        this[_pendingClosePromise].resolve(undefined);
```
aapoalas, line 227: Do these resolve calls actually work somehow? Promise does not have a method to resolve it from the outside, and these Promises seem to be created with the normal primordials Promise ie. globalThis Promise.

ext/webserial/01_webserial.js
```diff
@@ -0,0 +1,397 @@
...
+      this[_state] = "opening";
+      this[_rid] = core.opSync("op_webserial_open_port", this[_name], options);
+      this[_state] = "opened";
+      this[_bufferSize] = options.bufferSize;
+    }
+
+    async setSignals(options = {}) {
+      webidl.assertBranded(this, SerialPort);
+      const prefix = "Failed to execute 'setSignals' on 'SerialPort'";
+      webidl.requiredArguments(arguments.length, 1, { prefix });
```
aapoalas, line 319: This seems like an odd sort of assert. I guess it would only fire if there are more than 1 arguments passed to the function? I guess WebIDL is like that.

ext/webserial/01_webserial.js
```diff
@@ -0,0 +1,397 @@
...
+    async getSignals() {
+      webidl.assertBranded(this, SerialPort);
+      if (this[_state] !== "opened") {
+        throw new DOMException(
+          "SerialPort must be opened",
+          "InvalidStateError",
+        );
+      }
+
+      return await core.opAsync("op_webserial_get_signals", this[_rid]);
```
aapoalas, line 345: Extra `await`, consider forcing lint to accept the `async` instead.

ext/webserial/01_webserial.js
```diff
@@ -0,0 +1,397 @@
...
+      let abortPromise;
+      if (this[_writable] === null) {
+        abortPromise = PromiseResolve(undefined);
+      } else {
+        abortPromise = this[_writable].abort();
+      }
+
+      const pendingClosePromise = new Promise((res) => {
+        if (this[_readable] === null && this[_writable] === null) {
+          res(undefined);
```
aapoalas, line 366: I presume it is WebIDL spec conformity that these need to be resolved with an explicit undefined instead of zero-args?

ext/webserial/01_webserial.js
crowlKats: can result is worse stacktrace

ext/webserial/01_webserial.js
crowlKats: technically not required.

ext/webserial/01_webserial.js
crowlKats: easier to follow and match with spec

ext/webserial/01_webserial.js
crowlKats: Probably not. as stated, PR was extremely WIP when opened, and now is quite stale.

ext/webserial/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+  state: Rc<RefCell<OpState>>,
+  rid: ResourceId,
+  mut buf: ZeroCopyBuf,
+) -> Result<(), AnyError> {
+  let resource = state
+    .borrow_mut()
+    .resource_table
+    .get::<SerialPortResource>(rid)?;
+  let mut serial_port = RcRef::map(resource, |r| &r.0).borrow_mut().await;
+  serial_port.read_exact(buf.as_mut()).await?;
```
DjDeveloperr, line 137: Shouldn't we just use `read` instead of `read_exact` here? Buffer size is determined by `controller.desiredSize` but what if there's not enough data in serial port to read, would the stream just not pull until there is data of desired size?

crowlKats: Again. This PR is not anywhere near being ready.

PR #12283, opened 2021-09-30
fix: worker permissions should not permission prompt with resolved absolute file path

dsherret:
The code here was error prone to accidentally displaying the resolved file path rather than the path requested by the code. The fix was to introduce a `PermissionsPath` struct to reduce the chance of accidentally calling a `.check(...)` or similar method with a resolved file path.

Running the following code:

```ts
new Worker("file.js", {
  deno: {
    permissions: {
      read: ["./file.txt"],
      write: ["./file.txt"],
    },
  },
  type: "module",
});
```

Before:

```bash
> deno run --quiet --unstable --prompt test.ts
⚠️  ️Deno requests read access to "V:\deno\file.txt". Allow? [y/n (y = yes allow, n = no deny)] y
⚠️  ️Deno requests write access to "V:\deno\file.txt". Allow? [y/n (y = yes allow, n = no deny)]
```

After:

```bash
> ./target/debug/deno.exe run --quiet --unstable --prompt test.ts
⚠️  ️Deno requests read access to "./file.txt". Allow? [y/n (y = yes allow, n = no deny)] y
⚠️  ️Deno requests write access to "./file.txt". Allow? [y/n (y = yes allow, n = no deny)]
```

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -115,6 +115,62 @@ impl Default for PermissionState {
   }
 }
 
+/// A normalized environment variable name. On Windows this will
+/// be uppercase and on other platforms it will stay as-is.
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default)]
+struct EnvVarName {
+  inner: String,
+}
+
+impl EnvVarName {
+  pub fn new(env: impl AsRef<str>) -> Self {
+    Self {
+      inner: if cfg!(windows) {
+        env.as_ref().to_uppercase()
+      } else {
+        env.as_ref().to_string()
+      },
+    }
+  }
+}
+
+impl AsRef<str> for EnvVarName {
+  fn as_ref(&self) -> &str {
+    self.inner.as_str()
+  }
+}
+
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default)]
+pub struct PermissionPath {
+  original_path: PathBuf,
+  resolved_path: PathBuf,
+}
+
+impl PermissionPath {
+  pub fn new(path: impl AsRef<Path>) -> Self {
+    Self {
+      original_path: path.as_ref().to_path_buf(),
+      resolved_path: resolve_from_cwd(path.as_ref()).unwrap(),
+    }
+  }
+
+  pub fn starts_with(&self, other_path: &PermissionPath) -> bool {
+    self.resolved_path.starts_with(&other_path.resolved_path)
+  }
+
+  pub fn as_path(&self) -> &Path {
+    // Ensure the original path is returned in order to prevent providing
+    // the resolved path to another method in the permissions and exposing
+    // the CWD
+    self.original_path.as_path()
+  }
+
+  /// Path that can be displayed without leaking the CWD when not allowed.
+  pub fn safe_display(&self) -> std::path::Display<'_> {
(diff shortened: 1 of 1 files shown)


runtime/permissions.rs
```diff
@@ -115,6 +115,62 @@ impl Default for PermissionState {
   }
 }
 
+/// A normalized environment variable name. On Windows this will
+/// be uppercase and on other platforms it will stay as-is.
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Default)]
+struct EnvVarName {
```
dsherret, line 121: Ignore. Moved from below.

runtime/permissions.rs
```diff
@@ -115,6 +115,62 @@ impl Default for PermissionState {
...
+
+  pub fn starts_with(&self, other_path: &PermissionPath) -> bool {
+    self.resolved_path.starts_with(&other_path.resolved_path)
+  }
+
+  pub fn as_path(&self) -> &Path {
+    // Ensure the original path is returned in order to prevent providing
+    // the resolved path to another method in the permissions and exposing
+    // the CWD
+    self.original_path.as_path()
```
dsherret, line 165: Fix is here. Ensures this code doesn't use the resolved file path:

https://github.com/denoland/deno/blob/6bf5c850e6b4f6163721bd673b191bda0e0dc0a6/runtime/ops/worker_host.rs#L183

runtime/permissions.rs
```diff
@@ -115,6 +115,62 @@ impl Default for PermissionState {
...
+  }
```
dsherret, line 166: I didn't add a test for this because it was painful, but I just maybe thought of a better way of doing it, so I will try again right now.

runtime/permissions.rs
bartlomieju: We already have `UnaryPermission<ReadDescriptor>::check_blind` would that help in here?

caspervonb:
> The fix was to introduce a PermissionsPath struct to reduce the chance of accidentally calling a .check(...)

We have `check_blind` for this exact purpose.

runtime/permissions.rs
dsherret: No, I don’t think so because someone could do this and forget to call check_blind. The purpose is to reduce the chance of developer error which I think is important with permissions. Additionally they would have to figure out the original path.

dsherret:
@caspervonb see https://github.com/denoland/deno/pull/12283#discussion_r720091683

runtime/permissions.rs
dsherret: Actually, I just thought of an interesting security issue this change creates. I will add a test for that and improve this.

dsherret: Ok, removed this and implemented a new fix.

bartlomieju:
I believe this directly conflicts with changes in #12297, and that PR looks more complete and robust. Maybe let's focus our efforts there?

dsherret:
Closing this. It doesn't seem like a big deal to show the cwd in the permission prompt as there is no way for user code to read this from my understanding.

caspervonb:
Thought about this a bit and actually showing relative paths seems more insecure through obscurity; in that you can do some cwd changing and confuse the user into giving more access than what they might think they're giving.

PR #12284, opened 2021-10-01
perf(webidl): optimize isArrayBufferDetached()

AaronO:
Per MDN (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/byteLength#description):
> The `byteLength` property ... returns 0 if this `ArrayBuffer` has been detached.

### DIFF
--- a/ext/webidl/00_webidl.js
+++ b/ext/webidl/00_webidl.js
@@ -440,12 +440,7 @@
   }
 
   function isArrayBufferDetached(V) {
-    try {
-      new Uint8Array(V);
-      return false;
-    } catch {
-      return true;
-    }
+    return V.byteLength === 0;
   }
 
   converters.ArrayBuffer = (V, opts = {}) => {


andreubotella:
This is wrong:
```js
> new ArrayBuffer(0).byteLength
0
```

AaronO:
> This is wrong:
> 
> ```js
> > new ArrayBuffer(0).byteLength
> 0
> > new Uint8Array(new ArrayBuffer(0))
> Uint8Array(0) []
> ```

That's true and the tests fail for that reason, but potentially view `.length > 0` and buffer `.byteLength === 0` might be sufficient to classify

andreubotella:
> That's true and the tests fail for that reason, but potentially view `.length > 0` and buffer `.byteLength === 0` might be sufficient to classify

It's not:

```js
const buffer = new ArrayBuffer(1024);
const view = new Uint8Array(buffer);
console.log(buffer.byteLength); // 1024
console.log(view.length); // 1024

// Detaches the buffer
structuredClone({}, { transfer: [buffer] });

console.log(buffer.byteLength); // 0
console.log(view.length); // 0
```

But checking whether the byte length is 0 before trying to construct a view should optimize the majority of cases, right?

AaronO:
Unfruitful path, might make sense to check `v8::ArrayBuffer:: IsDetachable()` in the op-layer/sede_v8. This WebIDL check is still too expensive for what should amount to a no-op

lucacasonato:
`v8::ArrayBuffer::IsDetachable` is not right. It just tells if you if an array _could_ be detached (WASM ArrayBuffers can not be detached). It tells you nothing about the current detachment state.

PR #12285, opened 2021-10-01
perf(webidl): inline ResponseInit converter

AaronO:
Reduces Response constructor of a pre-encoded body by ~40%

```
Before:
resp_pre_encoded:       	n = 1000000, dt = 0.345s, r = 2898551/s, t = 345ns/op
After:
resp_pre_encoded:    	n = 1000000, dt = 0.216s, r = 4629630/s, t = 216ns/op
```

## Notes

- Optimizes `Response` constructor for all paths but benched on pre-encoded & no-headers to minimize noise


aapoalas:
Looks very good to me. Avoid magical one-size-fits-all helpers on fast paths to optimize for especially minimal, usual cases. Very good stuff!

lucacasonato (commented):
This fast path needs to be conditional on `isProxy` again.

AaronO:
> Can you add some tests for things like init being 0 or null, or init.status being 0 or null?
> 
> If we add quite extensive fast-paths, we need to make sure they behave correctly.

Upstream WPT tests or cli/tests/unit ?

PR #12286, opened 2021-10-01
perf(webidl): optimize createRecordConverter()

AaronO:
Cuts self-time by ~6x, 172ns/iter => 22ns/iter benched on 1M Response builds / HeadersInit calls

This should be functionally identical to the old code since:
1. `for...in` loops on objects iterate over all enumberable string keys of an object (including inherited)
2. Then we filter out inherited keys with `.hasOwnProperty(...)`

**Note:** WebIDL / WPT is coupled to a subpar implementation so our hands are tied, unless an isProxy fast path check isn't too expensive

## Benchmarks

```
No fast path:
resp_w_bh:           	n = 1000000, dt = 1.295s, r = 772201/s, t = 1294ns/op
Fast path with isProxy check:
resp_w_bh:           	n = 1000000, dt = 1.146s, r = 872600/s, t = 1146ns/op
Forced fast path:
resp_w_bh:           	n = 1000000, dt = 1.112s, r = 899281/s, t = 1112ns/op
```

Pending on https://github.com/denoland/deno/pull/12288

### DIFF
--- a/ext/webidl/00_webidl.js
+++ b/ext/webidl/00_webidl.js
@@ -48,6 +48,7 @@
     ObjectGetOwnPropertyDescriptor,
     ObjectGetOwnPropertyDescriptors,
     ObjectGetPrototypeOf,
+    ObjectPrototypeHasOwnProperty,
     ObjectIs,
     PromisePrototypeThen,
     PromiseReject,
@@ -844,16 +845,15 @@
           opts,
         );
       }
-      const keys = ReflectOwnKeys(V);
       const result = {};
-      for (const key of keys) {
-        const desc = ObjectGetOwnPropertyDescriptor(V, key);
-        if (desc !== undefined && desc.enumerable === true) {
-          const typedKey = keyConverter(key, opts);
-          const value = V[key];
-          const typedValue = valueConverter(value, opts);
-          result[typedKey] = typedValue;
+      for (const key in V) {
+        if (!ObjectPrototypeHasOwnProperty(V, key)) {
+          continue;
         }
+        const typedKey = keyConverter(key, opts);
+        const value = V[key];
+        const typedValue = valueConverter(value, opts);
+        result[typedKey] = typedValue;
       }
       return result;
     };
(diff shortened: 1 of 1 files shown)


lucacasonato:
This is not equal. `for .. in` also enumerates prototype fields, while `ReflectOwnKeys` does not.

lucacasonato:
Actually, NVM. Inherited keys are filtered out via `hasOwnProperty`

ext/webidl/00_webidl.js
```diff
@@ -844,8 +846,21 @@
...
       }
-      const keys = ReflectOwnKeys(V);
       const result = {};
+      // Fast path for common case (not a Proxy)
+      if (!core.isProxy(V)) {
+        for (const key in V) {
+          if (!ObjectPrototypeHasOwnProperty(V, key)) {
+            continue;
+          }
+          const typedKey = keyConverter(key, opts);
```
aapoalas, line 856: Would it be worth it to extract the three lines of code here into a helper function to avoid repetition?

lucacasonato: This is a very hot path, so extraneous function calls should be avoided.

lucacasonato (approved):

ext/webidl/00_webidl.js
AaronO: v8 would probably inline it, but lets not prematurely generalize

ext/webidl/00_webidl.js
```diff
@@ -844,8 +846,21 @@
...
+          const value = V[key];
+          const typedValue = valueConverter(value, opts);
+          result[typedKey] = typedValue;
+        }
+      }
+      // Slow path if Proxy (e.g: in WPT tests)
```
kurtextrem, line 862: I might be missing something here, but just by looking at the code a `return` or `else` might be missing, since like this is runs both fast and slow path?

AaronO: That was smart ... Good catch! I accidentally messed this up in the tweak commit (https://github.com/denoland/deno/pull/12286/commits/c27e43a793131e8c8b1f1087893144e182f5ecfb)

PR #12290, opened 2021-10-01
feat(core): native bindings names

AaronO:
Makes native builtin functions easier to recognize when debugging/profiling, they would otherwise appear as `(anonymous)` functions

Requires https://github.com/denoland/rusty_v8/pull/792


bartlomieju (approved):
Please add a test to `core/runtime.rs` that check names for `Deno.core` bindings before landing

PR #12294, opened 2021-10-01
feat(web): Implement `DOMException`'s `stack` property.

andreubotella:
[As per WebIDL](https://heycam.github.io/webidl/#es-DOMException-specialness), if `Error` objects have a `stack` property, so should `DOMException` instances.


bartlomieju (approved):
LGTM

PR #12297, opened 2021-10-02
fix(runtime/ops/worker_host): Move permission arg parsing to Rust

nayeemrmn:
- fix: Implement run allowlists in worker/test permissions (https://github.com/denoland/deno/pull/12078#discussion_r719720051).
- fix: Make partially unspecifed worker/test permissions default to `"inherit"` (implied in https://deno.land/manual@v1.14.2/runtime/workers#specifying-worker-permissions but currently defaults to `"prompt"` or `"denied"`).
- fix: Make ungranted worker/test permission states `"prompt"` unless inherited as `"denied"` (was inconsistently `"denied"` in odd cases).
- fix: Properly distinguish between names and paths in run permission descriptors: `foo` is a name; `./foo`, `/foo`, `../foo` and `foo/` are paths to be resolved from CWD.
- feat: Expose allowlists in runtime APIs for ffi permissions.


cli/tests/testdata/workers/test.ts
```diff
@@ -454,61 +453,33 @@ Deno.test("Worker limit children permissions", async function () {
...
-      index,
-      path,
-    })
-  );
-
-  await promise;
+  const promise = deferred();
+  worker.onmessage = ({ data }) => promise.resolve(data);
+  // deno-fmt-ignore
+  assertEquals(await promise, { envGlobal: "prompt", envFoo: "granted", envAbsent: "prompt", hrtime: "granted", netGlobal: "prompt", netFoo: "granted", netFoo8000: "granted", netBar: "prompt", netBar8000: "granted", ffiGlobal: "prompt", ffiFoo: "granted", ffiBar: "granted", ffiAbsent: "prompt", readGlobal: "prompt", readFoo: "granted", readBar: "granted", readAbsent: "prompt", runGlobal: "prompt", runFoo: "granted", runBar: "granted", runBaz: "granted", runAbsent: "prompt", writeGlobal: "prompt", writeFoo: "granted", writeBar: "granted", writeAbsent: "prompt" });
```
bartlomieju, line 477: Please factor this `assertEquals` to separate object that won't require to ignore formatting. This should format nicely with double indent

cli/tests/testdata/workers/test.ts
```diff
@@ -519,104 +490,39 @@ Deno.test("Nested worker limit children permissions", async function () {
...
-    worker.postMessage({
-      index,
-      path,
-    })
-  );
-
-  await promise;
+  worker.onmessage = ({ data }) => promise.resolve(data);
+  // deno-fmt-ignore
+  assertEquals(await promise, { envGlobal: "prompt", envFoo: "prompt", envAbsent: "prompt", hrtime: "prompt", netGlobal: "prompt", netFoo: "prompt", netFoo8000: "prompt", netBar: "prompt", netBar8000: "prompt", ffiGlobal: "prompt", ffiFoo: "prompt", ffiBar: "prompt", ffiAbsent: "prompt", readGlobal: "prompt", readFoo: "prompt", readBar: "prompt", readAbsent: "prompt", runGlobal: "prompt", runFoo: "prompt", runBar: "prompt", runBaz: "prompt", runAbsent: "prompt", writeGlobal: "prompt", writeFoo: "prompt", writeBar: "prompt", writeAbsent: "prompt" });
```
bartlomieju, line 496: Ditto

cli/tests/testdata/workers/test.ts
```diff
@@ -519,104 +490,39 @@ Deno.test("Nested worker limit children permissions", async function () {
...
-        {
-          type: "module",
-          deno: {
-            namespace: true,
-            permissions: {
-              env: true,
+Deno.test({
+  name:
+    "Worker initialization throws on worker permissions greater than parent thread permissions",
+  permissions: { env: false },
```
bartlomieju, line 504: Nice!

runtime/js/11_workers.js
```diff
@@ -152,43 +70,20 @@
...
-        deno = {},
-        name = "unknown",
+        deno,
+        name,
         type = "classic",
       } = options;
 
-      // TODO(Soremwar)
-      // `deno: boolean` is kept for backwards compatibility with the previous
-      // worker options implementation. Remove for 2.0
```
bartlomieju, line 162: This TODO should probably stay

runtime/js/40_testing.js
```diff
@@ -3,10 +3,10 @@
 
 ((window) => {
   const core = window.Deno.core;
-  const { parsePermissions } = window.__bootstrap.worker;
   const { setExitHandler } = window.__bootstrap.os;
   const { Console, inspectArgs } = window.__bootstrap.console;
   const { metrics } = window.__bootstrap.metrics;
+  const { serializePermissions } = window.__bootstrap.permissions;
```
bartlomieju, line 9: Cool! Nice to see this being reused in that way, looks like we ended up with a single source of truth

runtime/ops/worker_host.rs
```diff
@@ -131,361 +120,442 @@ pub fn init(create_web_worker_cb: Arc<CreateWebWorkerCb>) -> Extension {
...
-    } else {
-      main.state = worker;
-    }
-  }
-  Ok(main)
+fn escalation_error() -> AnyError {
+  custom_error(
+    "PermissionDenied",
+    "Can't escalate parent thread permissions",
+  )
```
bartlomieju, line 127: I'm wondering if these changes should mostly live in `runtime/permissions.rs`... Seems like serialization of permissions could be better suited there instead of `ops/worker_host.rs`. Thoughts?

bartlomieju: Also seems like `create_worker_permissions` could use some unit tests - should be easier (and faster) that having heavy integration tests

runtime/permissions.rs
```diff
@@ -240,11 +242,38 @@ impl AsRef<str> for EnvDescriptor {
...
 
-#[derive(Clone, Eq, PartialEq, Hash, Debug, Default)]
-pub struct RunDescriptor(pub String);
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub enum RunDescriptor {
+  Name(String),
+  Path(PathBuf),
+}
+
+impl FromStr for RunDescriptor {
```
bartlomieju, line 251: Great changes in this file, having explicit `Default` implementations is also more readable in this case.

bartlomieju (commented):
@nayeemrmn this is a massive PR! It mostly looks good to me, however there are a lot of intricacies that require special attention. Would it be hard to split it into two PRs where one addressed the bits regarding FFI and the other for worker permission serialization?

nayeemrmn:
> Would it be hard to split it into two PRs where one addressed the bits regarding FFI and the other for worker permission serialization?

I think it would be at this point. For a separate PR with FFI path resolution, I'd have to knowingly keep bugs in at the old deserializing code and I didn't study it too hard to know how they'd interact.

The PR seems easier to swallow with everything now in `permissions.rs`.
- Improve `RunDescriptor` and `FfiDescriptor` interfaces so they are interpreted from flag and runtime inputs correctly.
- Add some missing `check_all()` functions to `UnaryPermission<_>`.
- Remove some nonsensical `#[derive(Default)]`, add some explicit implementations.
- Below all of that is deserialization code and `create_child_permissions()`, only used outside of `permissions.rs`.
- There is also exposing FFI allowlists, but these are 'implemented' everywhere already except at the very edge of JS.

bartlomieju:
@nayeemrmn thanks for in-depth description. I will give it another pass tomorrow, also prompted David to do a review as well.

runtime/permissions.rs
```diff
@@ -1204,6 +1395,444 @@ fn resolved_and_display_path(path: &Path) -> (PathBuf, PathBuf) {
...
+        V: de::MapAccess<'de>,
+      {
+        let mut child_permissions_arg = ChildPermissionsArg::default();
+        while let Some((key, value)) =
+          v.next_entry::<String, serde_json::Value>()?
+        {
+          if key == "env" {
+            child_permissions_arg.env =
+              serde_json::from_value::<ChildUnaryPermissionArg>(value)
+                .map_err(|e| de::Error::custom(e.to_string()))?;
```
dsherret, line 1600: It would be nice if these error messages would say the key they occurred in like they previously did. That way the user can hone in on where to fix the problem more quickly.

dsherret (commented):
Looks really good overall! It would be nice to add a few js surface level unit tests to demonstrate what the error messages look like to the user when they provide invalid permission values (ex. providing a url where that's not expected).

dsherret:
@nayeemrmn in the future I would recommend to not squashing commits in a PR as they will be squashed once merged. I have no clue what changed since I last reviewed and this is a very large PR. Any way to break it back up?

PR #12299, opened 2021-10-02
refactor(runtime): Worker bootstrap options

AaronO:
## Notes

- Factors out shared portions of WorkerOptions & WebWorkerOptions to BootstrapOptions
- Adds ::bootstrap_from_options() methods
- Passes options by value instead of by ref so that we can mutate them and drain a vec of extensions passed
- I suppose the main reason we passed them by ref is that we had to repass them to .bootstrap()
- Conceptually, BootstrapOptions holds "higher-level" nice-to-have options compared to the rest of WorkerOptions that are closer tied to internal rust-side mechanics

### DIFF
--- a/runtime/web_worker.rs
+++ b/runtime/web_worker.rs
@@ -6,6 +6,7 @@ use crate::metrics;
 use crate::ops;
 use crate::permissions::Permissions;
 use crate::tokio_util::create_basic_runtime;
+use crate::BootstrapOptions;
 use deno_broadcast_channel::InMemoryBroadcastChannel;
 use deno_core::error::AnyError;
 use deno_core::error::JsError;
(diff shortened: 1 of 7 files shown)


runtime/web_worker.rs
```diff
@@ -275,32 +270,38 @@ pub struct WebWorkerOptions {
...
+    options: WebWorkerOptions,
+  ) -> (Self, SendableWebWorkerHandle) {
+    let bootstrap_options = options.bootstrap.clone();
+    let (mut worker, handle) =
+      Self::from_options(name, permissions, main_module, worker_id, options);
+    worker.bootstrap(&bootstrap_options);
+    (worker, handle)
+  }
+
   pub fn from_options(
```
bartlomieju, line 295: Can this function be made non-public?

bartlomieju (approved):

PR #12300, opened 2021-10-02
fix(ext/net): should not panic when listening to unix abstract address

### LINKED ISSUE #11412: Panic when listening to unix socket abstract address
if you try this simple example:

```
let listener = Deno.listen({path: "\0aaa", transport: "unix"});
```

deno crashes:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', runtime/ops/net.rs:527:34
```

The issue I believe is that the address "\0aaa", while being a valid unix socket address (it is an abstract address, not bound to the filesystem), is not a valid Rust path name, so the unwrap fails:

https://github.com/denoland/deno/blob/79ac050d2917ddc1be97e458c8ac7c6a3d728269/extensions/net/ops.rs#L526

This use case used to work until v1.5.4 at least, but is broken in recent versions

ahabhgk:
Fixes #11412

The test detects the runtime whether running into a panic by `asset` after `Deno.listen`, would like to see a better way to detect the panic, feel free to suggest.

### DIFF
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -625,3 +625,17 @@ unitTest(
     listener.close();
   },
 );
+
+unitTest(
+  {
+    ignore: Deno.build.os !== "linux",
+    permissions: { read: true, write: true },
+  },
+  function netUnixAbstractPathShouldNotPanic() {
+    const listener = Deno.listen({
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks!

Issue #11632, opened 2021-08-10
Incorrectly using 32 bit dll in `Deno.dlopen` on Windows leads to error message with format string


dsherret:
Using a 32 bit dll in `Deno.dlopen` will lead to an error message that has a format string:

```
error: Uncaught Error: Could not open library: %1 is not a valid Win32 application. (os error 193)
const dylib = Deno.dlopen("add_numbers.dll", {
                   ^
    at deno:core/01_core.js:106:46
    at unwrapOpResult (deno:core/01_core.js:126:13)
    at Object.opSync (deno:core/01_core.js:140:12)
    at new DynamicLibrary (deno:extensions/ffi/00_ffi.js:12:24)
    at Object.dlopen (deno:extensions/ffi/00_ffi.js:26:12)
    at file:///V:/ffi-test/add_numbers.ts:1:20
```

```ts
const dylib = Deno.dlopen("add_numbers.dll", {
  "add_numbers": { parameters: ["i32", "i32"], result: "i32" },
});
```

[add_numbers.zip](https://github.com/denoland/deno/files/6962414/add_numbers.zip)

Note that the fix for the user is to compile a 64 bit dll (use cl with "x64 native tools command prompt" instead of the "developer command prompt")

PR #12301, opened 2021-10-02
fix(ext/ffi): formatting dlopen errors on Windows

littledivy:
Fixes https://github.com/denoland/deno/issues/11632


AaronO (dismissed):
Have you considered moving out the error mapping to its own function ? I think that would be cleaner

ext/ffi/lib.rs
```diff
@@ -269,6 +270,63 @@ struct FfiLoadArgs {
...
+      use winapi::shared::minwindef::DWORD;
+      use winapi::shared::ntdef::WCHAR;
+      use winapi::um::winbase::FormatMessageW;
+      use winapi::um::winbase::FORMAT_MESSAGE_ARGUMENT_ARRAY;
+      use winapi::um::winbase::FORMAT_MESSAGE_FROM_SYSTEM;
+
+      let err_num = e.raw_os_error().unwrap();
+
+      // Language ID given by
+      // MAKELANGID(LANG_SYSTEM_DEFAULT, SUBLANG_SYS_DEFAULT);
```
dsherret, line 296: Maybe we should just do this expression below instead of having a comment just to ensure it's actually correct and to have a single source of truth.

ext/ffi/lib.rs
```diff
@@ -269,6 +270,63 @@ struct FfiLoadArgs {
...
+    dlopen::Error::OpeningLibraryError(e) => {
+      use std::ffi::OsStr;
+      use std::os::windows::ffi::OsStrExt;
+      use winapi::shared::minwindef::DWORD;
+      use winapi::shared::ntdef::WCHAR;
+      use winapi::um::winbase::FormatMessageW;
+      use winapi::um::winbase::FORMAT_MESSAGE_ARGUMENT_ARRAY;
+      use winapi::um::winbase::FORMAT_MESSAGE_FROM_SYSTEM;
+
+      let err_num = e.raw_os_error().unwrap();
```
dsherret, line 293: It looks like this could possibly be `None` when it can't get the last error. Maybe just in case we could return `e.to_string()` when this is `None`?

ext/ffi/lib.rs
```diff
@@ -269,6 +270,63 @@ struct FfiLoadArgs {
...
+fn format_error(e: dlopen::Error, path: String) -> String {
+  match e {
+    #[cfg(target_os = "windows")]
+    // This calls FormatMessageW with library path
+    // as replacement for the insert sequences.
+    // Unlike libstd which passes the FORMAT_MESSAGE_IGNORE_INSERTS
+    // flag without any arguments.
+    //
+    // https://github.com/denoland/deno/issues/11632
+    dlopen::Error::OpeningLibraryError(e) => {
```
dsherret, line 284: Would it be useful to do this for other kinds of errors?

dsherret (changes requested):
Almost there. It would be useful to add some windows only unit tests doing something like `format_error(io:Error::from_raw_os_error(code), path)`.

ext/ffi/lib.rs
littledivy: `OpeningLibraryError` is the only possible error when calling `Deno.dlopen`. We probably won't even need to map other errors if they don't have any insert sequences. Mapping them right now will lead to weird issues without the presence of the `FORMAT_MESSAGE_IGNORE_INSERTS` flag and we will have to maintain a list of Win32 errors that need that flag.

ext/ffi/lib.rs
```diff
@@ -269,6 +270,63 @@ struct FfiLoadArgs {
...
+      unsafe {
+        let length = FormatMessageW(
+          FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ARGUMENT_ARRAY,
+          std::ptr::null_mut(),
+          err_num as DWORD,
+          lang_id as DWORD,
+          buf.as_mut_ptr(),
+          buf.len() as DWORD,
+          arguments.as_ptr() as _,
+        );
```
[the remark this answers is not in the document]
littledivy: Done

dsherret (approved):
LGTM! Thanks for fixing this. There are just some merge conflicts that need to be resolved though.

piscisaureus (approved):

Issue #12067, opened 2021-09-14
Implement serializable and transferrable web platform APIs


andreubotella:
Deno's implementation of structured serialize currently supports serializing the JS built-in `SharedArrayBuffer`s and transferring the JS built-in `ArrayBuffer`s, as well as the `MessagePort` web API. With #11823 it will also support serializing the wasm built-in `WebAssembly.Module`.

But Deno also implements some web APIs that per the spec should be serializable or transferable but aren't in Deno's implementation. In particular, `DOMException`, `File`, `Blob` and a number of GPU interfaces should be serializable; and `ReadableStream`, `WritableStream` and `TransformStream` should be transferable.

Implementing this would probably need adding some `Deno.core` API to define serialization and transfer steps, possibly in connection to the existing `Deno.core.createHostObject()`.

kitsonk:
I don't know if we would need to have have extensions to `Deno.core`.  Wouldn't we just need something like: 

```js
const denoSerialize = Symbol("Deno.serialize");
const denoDeserialize = Symbol("Deno.deserialize");

class DOMException {
  static [denoSeralize](instance) {
    // serialize steps that return a v8 serialization
  }
  static [denoDeseralize](value) {
    // returns a hydrated instance from a v8 serialization
  }
}
```

And if we put the `denoSerialize` and `denoDeserialize` in the web platform helpers and not using `.for()` would ensure they are unique and not accessible from userspace.

andreubotella:
> I don't know if we would need to have have extensions to `Deno.core`. Wouldn't we just need something like:
> 
> ```js
> const denoSerialize = Symbol("Deno.serialize");
> const denoDeserialize = Symbol("Deno.deserialize");
> 
> class DOMException {
>   static [denoSeralize](instance) {
>     // serialize steps that return a v8 serialization
>   }
>   static [denoDeseralize](value) {
>     // returns a hydrated instance from a v8 serialization
>   }
> }
> ```
> 
> And if we put the `denoSerialize` and `denoDeserialize` in the web platform helpers and not using `.for()` would ensure they are unique and not accessible from userspace.

[...]

PR #12302, opened 2021-10-02
fix(core/runtime): sync_ops_cache if nuked Deno ns

AaronO:
Decouple JsRuntime::sync_ops_cache() from the availability of the Deno.* namespace in the global scope

This avoids crashes when calling sync_ops_cache() on a bootstrapped WebWorker who has dropped its Deno.* namespace

It's also just cleaner and more robust ...

## Notes

- Observed when refactoring Worker setup in https://github.com/denoland/deno/pull/12299

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -146,6 +146,7 @@ pub type CompiledWasmModuleStore = CrossIsolateStore<v8::CompiledWasmModule>;
 pub(crate) struct JsRuntimeState {
   pub global_context: Option<v8::Global<v8::Context>>,
   pub(crate) js_recv_cb: Option<v8::Global<v8::Function>>,
+  pub(crate) js_sync_cb: Option<v8::Global<v8::Function>>,
   pub(crate) js_macrotask_cb: Option<v8::Global<v8::Function>>,
   pub(crate) js_wasm_streaming_cb: Option<v8::Global<v8::Function>>,
   pub(crate) pending_promise_exceptions:
@@ -350,6 +351,7 @@ impl JsRuntime {
       pending_mod_evaluate: None,
       dyn_module_evaluate_idle_counter: 0,
       js_recv_cb: None,
+      js_sync_cb: None,
       js_macrotask_cb: None,
       js_wasm_streaming_cb: None,
       js_error_create_fn,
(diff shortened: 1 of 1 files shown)


core/runtime.rs
```diff
@@ -476,35 +479,44 @@ impl JsRuntime {
...
-  fn init_recv_cb(&mut self) {
-    let scope = &mut self.handle_scope();
-
-    // Get Deno.core.opresolve
-    let code = v8::String::new(scope, "Deno.core.opresolve").unwrap();
+  /// Grab a Global handle to a function returned by the given expression
+  fn grab_fn(
+    scope: &mut v8::HandleScope,
+    code: &str,
+  ) -> v8::Global<v8::Function> {
```
andreubotella, line 492: Does this have to be specific to functions? One of the possibilities for implementing #12067 is to define `Deno.serialize` and `Deno.deserialize` symbols, which would have to be accessible from `bindings.rs`. Could this function be generic over `T where for<'s> v8::Local<'s, T>: TryFrom<v8::Local<'s, v8::Value>>`?

AaronO: Sure, there's this could be made generic, but I would rather do that in a separate PR or one implementing #12067 since there's no immediate need in this PR for that helper to be generic

andreubotella: SGTM

bartlomieju:
@AaronO what's the reason for this change? It works fine right now so why complicate it?

AaronO:
> @AaronO what's the reason for this change? It works fine right now so why complicate it? If #12299 is the reason I'd suggest to change #12299 because it must change the order of operations that are happening when workers are crated.

I have changed #12299 (and #12299 works perfectly fine in its current state). But the rationale is explained the notes above, but in short `JsRuntime::sync_ops_cache()` a functionality provided by `core/` shouldn't depend on the existence of `Deno.*` in the global scope since that is nuked for `WebWorkers` (with `use_deno_namespace: false`) and other potential embedders that don't expose `Deno.*`.

So whilst this is more lines of code, it's more robust and simpler because it aligns us to a single behaviour of calling back into JS whether that's for `opresolve()` or `syncOpsCache()`

AaronO:
In the limit I would consider removing `JsRuntime::sync_ops_cache()` as a public method, but we still use it ourselves so in the meantime it might as well be robust

lucacasonato (approved):
In favor. This makes `core/` more robust to tampering things in global scope.

PR #12309, opened 2021-10-03
perf(core): avoid extra then on opAsync()

AaronO:
Shaves off ~12% (~390ns/iter => ~340ns/iter)

Measured by the op_baseline bench:
```
Before:
bench:     390,829 ns/iter (+/- 3,712)
After:
bench:     342,821 ns/iter (+/- 6,250)
```

## Notes

This hurts/breaks async strack traces.

### Before

```
BadResource: Bad resource ID
    at async Object.read (deno:runtime/js/12_io.js:105:19)
    at async <anonymous>:5:5
Uncaught ReferenceError: assertStringIncludes is not defined
    at <anonymous>:10:5
```

### After

```
BadResource: Bad resource ID
    at deno:core/01_core.js:109:46
    at unwrapErr (deno:core/01_core.js:139:12)
    at opresolve (deno:core/01_core.js:96:24)
Uncaught ReferenceError: assertStringIncludes is not defined
    at <anonymous>:10:5
```

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -18,7 +18,6 @@
     MapPrototypeGet,
     MapPrototypeDelete,
     MapPrototypeSet,
-    PromisePrototypeThen,
     ObjectAssign,
   } = window.__bootstrap.primordials;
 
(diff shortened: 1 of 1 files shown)


aapoalas:
Looks very good to me!

lucacasonato:
I am not really in favor, because stack traces become significantly worse.

PR #12310, opened 2021-10-03
perf(core): use opcall() directly

AaronO:
Instead of the wrapper dispatch() func, also now forbids passing opIds to opSync()/opAsync() callers must always pass names.

This is admittedly a micro-optimization, it shaves off ~10ns/iter for async ops and ~4ns/iter for sync ops (though on the scale of sync ops that's ~7%), it's half micro-optimization and half cleanup:

```
Before:
bench async:     393,383 ns/iter (+/- 6,025)
bench nop:      50,881 ns/iter (+/- 1,073)
bench sync:      57,920 ns/iter (+/- 669)

After:
bench async:     380,437 ns/iter (+/- 8,691)
bench nop:      46,860 ns/iter (+/- 1,470)
bench sync:      54,840 ns/iter (+/- 1,499)
```

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -97,11 +97,6 @@
     }
   }
 
-  function dispatch(opName, promiseId, control, zeroCopy) {
-    const opId = typeof opName === "string" ? opsCache[opName] : opName;
-    return opcall(opId, promiseId, control, zeroCopy);
-  }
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

PR #12312, opened 2021-10-04
refactor(core): split opcall into sync/async

AaronO:
Cleaner separation of concerns and easier to reason about, simplifies some upcoming changes (e.g: rust-side promise ring)

Also somewhat improves robustness and prevents users from calling async ops with `core.opSync()` and vice-versa

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -7,6 +7,7 @@ use crate::JsRuntime;
 use crate::Op;
 use crate::OpId;
 use crate::OpPayload;
+use crate::OpResult;
 use crate::OpTable;
 use crate::PromiseId;
 use crate::ZeroCopyBuf;
@@ -31,7 +32,10 @@ lazy_static::lazy_static! {
   pub static ref EXTERNAL_REFERENCES: v8::ExternalReferences =
     v8::ExternalReferences::new(&[
       v8::ExternalReference {
-        function: opcall.map_fn_to()
+        function: opcall_async.map_fn_to()
+      },
+      v8::ExternalReference {
+        function: opcall_sync.map_fn_to()
       },
       v8::ExternalReference {
         function: set_macrotask_callback.map_fn_to()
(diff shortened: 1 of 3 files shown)


core/bindings.rs
```diff
@@ -330,17 +335,59 @@ fn opcall<'s>(
...
+  };
+  let op = OpTable::route_op(op_id, state.op_state.clone(), payload);
+  match op {
+    Op::Sync(result) => {
+      rv.set(result.to_v8(scope).unwrap());
+    }
+    Op::NotFound => {
+      throw_type_error(scope, format!("Unknown op id: {}", op_id));
+    }
+    // Async ops (ref or unrref)
```
lucacasonato, line 356:
```suggestion
    // Async ops (ref or unref)
```

core/bindings.rs
```diff
@@ -361,9 +408,13 @@ fn opcall<'s>(
...
   match op {
-    Op::Sync(result) => {
-      rv.set(result.to_v8(scope).unwrap());
-    }
+    Op::Sync(result) => match result {
+      OpResult::Ok(_) => throw_type_error(
+        scope,
+        format!("Can not call a sync op [{}] with opAsync()", op_id),
+      ),
+      OpResult::Err(_) => rv.set(result.to_v8(scope).unwrap()),
```
lucacasonato, line 416: Why does this path still need to work?

AaronO: That's how serde_v8 deserialization errors are currently returned, that happens sync since we deserializing before jumping into the async fn

core/bindings.rs
AaronO: Good catch !

core/bindings.rs
lucacasonato: Are these thrown synchronously from opAsync, or are they returned via the promise?

AaronO: These are thrown synchronously, FYI this is the current behavior this PR doesn't change it, it was implicitly handled by the sync branch when both were mangled

bartlomieju (approved):

PR #12316, opened 2021-10-04
chore: remove No*Permissions structs

lucacasonato:
These are confusing. They say they are "for users that don't care about
permissions", but that isn't correct. `NoTimersPermissions` disables
permissions instead of enabling them.

I would argue that implementors should decide what permissions they want
themselves, and not take our opinionated permissions struct.

### DIFF
--- a/ext/fetch/lib.rs
+++ b/ext/fetch/lib.rs
@@ -120,19 +120,6 @@ pub trait FetchPermissions {
   fn check_read(&mut self, _p: &Path) -> Result<(), AnyError>;
 }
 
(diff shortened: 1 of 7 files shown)


bartlomieju:
> I would argue that implementors should decide what permissions they want
themselves, and not take our opinionated permissions struct.

This was originally added so that implementors can use default structs if they don't care about permissions 🤷 I don't have a strong opinion either way. I'll defer to @ry for review, whose idea it was

ry (approved):
LGTM - This will induce some changes in Deploy

PR #12317, opened 2021-10-04
test: refactor lint tests into directories

bartlomieju:
Just reorganizing lint tests in such a way that adding new file in a directory doesn't change result of other tests.

Ref https://github.com/denoland/deno/pull/11983


dsherret (approved):
LGTM

PR #12322, opened 2021-10-04
feat: introduce navigator.language

AnInternetTroll:
Link to the spec: https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-language-dev
[~~To access `navigator.language` you **must** pass `--allow-env` as it can lead to fingerprinting.~~](#issuecomment-934847231)

What this PR **doesn't** add is `languagechange` which quite frankly is outside of my skill level.


AnInternetTroll:
The spec says the following
> If the user is using an anonymizing service, then the value "en-US" is suggested; if all users of the service use that same value, that reduces the possibility of distinguishing the users from each other.

I would argue that if `--allow-env` isn't passed then `navigator.language` should be `en-US`. 

Also an additional note, on unix this would be more or less the same as getting the `LANG` env variable, so maybe `--allow-env=LANG` should allow `navigator.language` to return the correct language. This should be discussed though as it is not how it works on windows.

runtime/js/99_main.js
```diff
@@ -321,6 +321,16 @@ delete Object.prototype.__proto__;
         return numCpus;
       },
     },
+    language: {
+      get() {
+        return os.language()
```
crowlKats, line 326: add `webidl.assertBranded(this, Navigator);`

runtime/js/99_main.js
```diff
@@ -321,6 +321,16 @@ delete Object.prototype.__proto__;
...
+      }
+    },
+    languages: {
+      get() {
+        return os.languages()
```
crowlKats, line 331: add `webidl.assertBranded(this, Navigator);`

crowlKats (changes requested):
Missing WorkerNavigator typings and property definition in 99_main

runtime/js/99_main.js
crowlKats: (and `configurable` and `enumerable`)

lucacasonato:
Actually, this can be enabled by default (no permissions), because the default language can already be extracted via `Intl` global.

crowlKats (dismissed):
`languages` should be frozen

AnInternetTroll:
Not stale, work has been done in rusty_v8 https://github.com/denoland/rusty_v8/pull/953

bartlomieju:
This is now done and should be ready to land. @lucacasonato @crowlKats please review

lucacasonato (approved):
LGTM

bartlomieju (approved):
LGTM

runtime/worker_bootstrap.rs
```diff
@@ -47,6 +48,7 @@ impl Default for BootstrapOptions {
       enable_testing_features: Default::default(),
       debug_flag: Default::default(),
       ts_version: Default::default(),
+      locale: "en-EN".to_string(),
```
zbraniecki, line 51: this seems like an error - there's no region `EN`.

zbraniecki (commented):
fly-by review.

runtime/worker_bootstrap.rs
bartlomieju: Thanks! Silly mistake on my part, I'll fix it

PR #12329, opened 2021-10-05
chore: various op cleanup

crowlKats:
Primarily just gets rid of the `Option<ZeroCopyBuf>` and the `ok_or_else` for it which is the first thing in ops which stems from the first iteration of ZeroCopyBuf support, but doesn't make much sense anymore now.

For the webgpu related changes i will open up a PR at wgpu as well


AaronO (approved):
LGTM

PR #12333, opened 2021-10-05
fix(web): Format `DOMException`'s `stack` property

petamoriken:
Related: https://github.com/denoland/deno/pull/12294

I removed the `"Error: "` string from the `DOMException`'s `stack` property.

before:

after:


bartlomieju:
CC @andreubotella

andreubotella (approved):
LGTM

PR #12335, opened 2021-10-05
feat(ext/ffi): add support for buffer arguments

bartlomieju:
Forks from https://github.com/denoland/deno/pull/11648

### DIFF
--- a/test_ffi/src/lib.rs
+++ b/test_ffi/src/lib.rs
@@ -60,3 +60,10 @@ pub extern "C" fn sleep_blocking(ms: u64) {
   let duration = Duration::from_millis(ms);
   sleep(duration);
 }
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+#[no_mangle]
+pub extern "C" fn nonblocking_buffer(ptr: *const u8, len: usize) {
+  let buf = unsafe { std::slice::from_raw_parts(ptr, len) };
+  assert_eq!(buf, vec![1, 2, 3, 4, 5, 6, 7, 8]);
+}
--- a/test_ffi/tests/test.js
+++ b/test_ffi/tests/test.js
@@ -22,6 +22,11 @@ const dylib = Deno.dlopen(libPath, {
   "add_f32": { parameters: ["f32", "f32"], result: "f32" },
   "add_f64": { parameters: ["f64", "f64"], result: "f64" },
   "sleep_blocking": { parameters: ["u64"], result: "void", nonblocking: true },
+  "nonblocking_buffer": {
+    parameters: ["buffer", "usize"],
+    result: "void",
+    nonblocking: true,
+  },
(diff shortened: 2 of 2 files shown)


ext/ffi/lib.rs
```diff
@@ -149,85 +152,86 @@ impl From<NativeType> for libffi::middle::Type {
...
+  I16(i16),
+  U32(u32),
+  I32(i32),
+  U64(u64),
+  I64(i64),
+  USize(usize),
+  ISize(isize),
+  F32(f32),
+  F64(f64),
+  Buffer(*const u8),
```
bartlomieju, line 196: This should be removed in favor of using `union NativeValue`

test_ffi/src/lib.rs
```diff
@@ -6,6 +8,18 @@ pub extern "C" fn print_something() {
...
+#[no_mangle]
+pub extern "C" fn print_buffer(ptr: *const u8, len: usize) {
+  let buf = unsafe { std::slice::from_raw_parts(ptr, len) };
+  println!("{:?}", buf);
+}
+
+#[no_mangle]
+pub extern "C" fn return_buffer() -> *const u8 {
+  [1, 2, 3, 4, 5, 6, 7, 8].as_ptr()
+}
```
bartlomieju, line 21: Discussed offline with @piscisaureus and we're gonna remove buffer-return-values for now and only keep buffer-as-arg

ext/ffi/00_ffi.js
```diff
@@ -13,15 +15,34 @@
...
+
+          if (isNonBlocking) {
             core.opAsync("op_ffi_call_nonblocking", {
               rid: this.#rid,
               symbol,
               parameters,
-            })
-          : (...parameters) =>
+            });
+          } else {
```
piscisaureus, line 42: `return`

Are we missing a test for this?

bartlomieju: We do. I also forgot to forward buffers, fixed now.

test_ffi/tests/test.js
```diff
@@ -21,9 +22,16 @@ const dylib = Deno.dlopen(libPath, {
...
+  "nonblocking_buffer": {
+    parameters: ["buffer", "usize"],
+    result: "void",
+    nonblocking: true,
+  },
 });
 
 dylib.symbols.print_something();
+const buffer = new Uint8Array([1, 2, 3, 4, 5, 6, 7, 8]);
+dylib.symbols.print_buffer(buffer, buffer.length);
```
bartlomieju, line 34: This is a bit unhandy... I wonder if we could just sniff out length automatically in case of `ArrayBuffer`. Though I think users would still have to declare two parameters... @piscisaureus thoughts?

piscisaureus (approved):
LGTM

PR #12337, opened 2021-10-05
feat(compat): add support for node: prefixed built-ins

bartlomieju:
Adds support for "node:" prefix for Node built-ins in "--compat" mode.

As per https://nodejs.org/api/esm.html#esm_node_imports


ry (approved):
LGTM

PR #12338, opened 2021-10-05
fix(lsp): lint diagnostics respect config file

### LINKED ISSUE #12238: Deno lint diagnostics don't use deno.json when configured on settings.json
**Describe the bug**

```json
{
  "lint": {
    "rules": {
      "exclude": [
        "camelcase"
      ]
    }
  }
}

```
```js
// Identifier 'my_variable' is not in camel case.
// Consider renaming `my_variable` to `myVariable`deno-lint(camelcase)
const my_variable = 1;
```

```bash
deno lint --config=deno.json
# Checked 1 file
```

**Expected behavior**

Deno lint diagnostics inside VSCode must match the ones from the CLI when passing `deno.json`

**Screenshots**

image

**Versions**

vscode: 1.60.2
deno: 1.14.1
extension: 3.9.1

bartlomieju:
This commit fixes problem with LSP where diagnostics coming
from "deno lint" don't respect configuration file.

LSP was changed to store "Option<ConfigFile>" on "Inner" and 
"StateSnapshot".

Fixes #12238

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -5,11 +5,13 @@ use super::tsc;
 
 use crate::ast;
 use crate::ast::Location;
+use crate::config_file::ConfigFile;
 use crate::lsp::documents::DocumentData;
 use crate::module_graph::parse_deno_types;
 use crate::module_graph::parse_ts_reference;
 use crate::module_graph::TypeScriptReference;
 use crate::tools::lint::create_linter;
+use crate::tools::lint::get_configured_rules;
 
 use deno_ast::swc::ast as swc_ast;
 use deno_ast::swc::common::DUMMY_SP;
@@ -28,8 +30,8 @@ use deno_core::serde_json::json;
 use deno_core::url;
 use deno_core::ModuleResolutionError;
 use deno_core::ModuleSpecifier;
-use deno_lint::rules;
 use import_map::ImportMap;
+use log::error;
 use lspower::lsp;
 use lspower::lsp::Position;
 use lspower::lsp::Range;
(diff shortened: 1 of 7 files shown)


cli/lsp/analysis.rs
```diff
@@ -135,9 +137,24 @@ fn as_lsp_range(range: &deno_lint::diagnostic::Range) -> Range {
 
 pub fn get_lint_references(
   parsed_source: &deno_ast::ParsedSource,
+  maybe_config_file: Option<&ConfigFile>,
 ) -> Result<Vec<Reference>, AnyError> {
   let syntax = deno_ast::get_syntax(parsed_source.media_type());
-  let lint_rules = rules::get_recommended_rules();
+  let maybe_lint_config = if let Some(config_file) = maybe_config_file {
```
kitsonk, line 143: Hmmm, seeing this in action worries me a bit, because it means every time we `get_lint_references` we _re-parse_ the config file, and regenerate the rules, though we only need to do this once.

Maybe it would be better to:

- Add a `lint_rules` to the `Inner` of the language server.
- Calculate that once at the points like we do with with `update_tsconfig`, `update_import_map`, `update_cache` and `update_registries`.
- Include the `lint_rules` only in the snapshot.

I also am thinking this might be a better pattern, we don't want the interpretation of the config file dispersed through out the LSP.

Relocating this logic to only occur when a change occurs and in the language server would also make it easier to do a `self.client.show_message(MessageType::Warning, err).await;` when there is an error.  Just logging out an error to the debug output of the lsp won't really help people fix issues.

cli/lsp/language_server.rs
```diff
@@ -413,6 +414,7 @@ impl Inner {
         LspError::internal_error()
       })?,
       documents: self.documents.clone(),
+      maybe_config_file: self.maybe_config_file.clone(),
```
kitsonk, line 417: Is it expensive to clone?  If so maybe we should have an `Arc<ConfigFile>` instead?  We should use `Arc` more in passing around state like this I think anyways.  But also in light of the other comment, I think we should just have it be `lint_rules`.

cli/tools/lint.rs
```diff
@@ -434,7 +434,7 @@ fn sort_diagnostics(diagnostics: &mut Vec<LintDiagnostic>) {
   });
 }
 
-fn get_configured_rules(
+pub fn get_configured_rules(
```
kitsonk, line 437:
```suggestion
pub(crate) fn get_configured_rules(
```

kitsonk (commented):
Seeing it in action, I would suggest a different approach...

cli/lsp/analysis.rs
bartlomieju: `lint_rules` seems too specific to me. I think we should store parsed `FmtConfig` and `LintConfig` on the `Inner` and pass them down as needed.

cli/lsp/language_server.rs
bartlomieju: It might be, but I would need to do a lot of changes across the whole codebase to store `Option<Arc<ConfigFile>>` (including `module_graph.rs` and `program_state.rs`), see comment above.

Issue #12295, opened 2021-10-01
basic node compatibity mode


ry:
Make the following work:
```
 mkdir deno-project && cd deno-project && yarn init deno-project -y
yarn add rollup
echo 'import { VERSION } from "rollup";' > foo.mjs
echo 'console.log(VERSION);' >> foo.mjs
deno run --compat foo.mjs
```

happenslol:
Hey, 
`tsserver` has annoyed me to a point where I'd do anything to have a better/faster alternative, including tackling this issue ;-)

Is there currently any work being done on this, or has someone else already picked up this issue and could give me some pointers on where I could help? Simply resolving packages from the local `node_modules` folders seems pretty straightforward to me, even though this just covers packages with an explicit module field.  
Down the road it would also be interesting to look into handling yarn v2 import since it also uses a central package folder instead of local `node_modules` (which is similar to what deno does afaik?).

wojpawlik:
Compatibility mode should be enabled per-import, so that users aren't forced to "poison" entire process for a single Node-only dependency.

<hr>

CJS and extensionless imports are on the way out: https://blog.sindresorhus.com/hello-modules-d1010b4e777b. https://deno.land/std/node/import_map.json would simplify writing or generating ESM that works on both runtimes.

bartlomieju:
> Compatibility mode should be enabled per-import, so that users aren't forced to "poison" entire process for a single Node-only dependency.

That sounds nice, but is not really feasible to apply different resolution schemes per import. 

> CJS and extensionless imports are on the way out: https://blog.sindresorhus.com/hello-modules-d1010b4e777b.

I don't believe that's true. It will take years before CJS are abandoned by community and few more to be removed from Node (if ever).

kt3k:
>  load file corresponding to "module" entry

Do we only use "module" property? I believe "module" property is a non-standard property used by some bundlers like rollup, webpack, etc, and most npm packages don't provide it.

Another question: Do we provide global `require` function?

AnInternetTroll:
Could the flag name be changed from `--compat` to anything else? IMO it's not descriptive and it should better reflect that it's a compatibility mode with node.js.

hayd:
optionally passing the std version (with a recent default that could be overridden), would at least enable node/std to update independently of deno.

e.g. `--compat` be equivalent to `--compat=0.110.0` etc. 😬

bartlomieju:
[...]

PR #12342, opened 2021-10-06
feat(compat): inject Node globals

bartlomieju:
This commit adds automatic injection of Node globals when "--compat" flag
is present.

This is done by executing "https://deno.land/std/node/global.ts" as a "side module",
before main module is executed.

This commit makes "--compat" required to be used with "--unstable" flag, as some
of Node globals require unstable Deno APIs.

Towards #12295


ry (approved):
LGTM

PR #12344, opened 2021-10-06
fix(ext/ffi): don't panic in dlopen

littledivy:
bug introduced in #12301


ext/ffi/lib.rs
```diff
@@ -370,7 +369,12 @@ where
   let permissions = state.borrow_mut::<FP>();
   permissions.check(&path)?;
 
-  let lib = Library::open(&path).map_err(|e| anyhow!(format_error(e, path)))?;
```
dsherret, line 373: Why was this causing a panic?

PR #12348, opened 2021-10-06
Revert "fix(cli): ensure empty lines don't count towards coverage (#1…

bartlomieju:
…1957)"

This reverts commit d5b38a992933db5cb2d0221e9d82af191022dad5.

Causes panic in https://github.com/denoland/deno_std/pull/1356

CC @caspervonb


caspervonb:
Hmm, yeah go ahead and revert; I intended to at some point but did not implement fallbacks for those unwrap calls.

PR #12370, opened 2021-10-08
feat(ext/net): relevant errors for resolveDns

### LINKED ISSUE #9197: Deno.resolveDns should return more useful error info
In the future, at latest before this API gets stabilized, `Deno.resolveDns` should return the cause of the error as precise as it can. 
Looking at the other languages, in the case of failure, Golang returns [DNSError type](https://golang.org/pkg/net/#DNSError), and Node.js returns `Error` object with [these error codes](https://nodejs.org/api/dns.html#dns_error_codes). It would be nice if we returned errors likewise.

_Originally posted by @magurotuna in https://github.com/denoland/deno/issues/9187#issuecomment-763480686_

satyarohith:
This PR maps resolveDns errors to Deno errors where applicable.

DNS protocol errors are still returned as generic errors. I'm
not sure if we should create a new DNSError class for them.

Closes #9197

### DIFF
--- a/cli/tests/testdata/resolve_dns.ts
+++ b/cli/tests/testdata/resolve_dns.ts
@@ -37,6 +37,6 @@ console.log(JSON.stringify(txt));
 
 try {
   await Deno.resolveDns("not-found-example.com", "A", nameServer);
-} catch (_e) {
-  console.log("Error thrown for not-found-example.com");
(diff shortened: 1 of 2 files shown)


kt3k:
Is it possible to write test cases which exercise some of these code paths?

PR #12378, opened 2021-10-08
fix(runtime): Declare `Window.self` and `DedicatedWorkerGlobalScope.name` with `util.writable()`

andreubotella:
`Window`'s `self` property and `DedicatedWorkerGlobalScope`'s `name` property are defined as Web IDL read-only attributes with the [`[Replaceable]`](https://webidl.spec.whatwg.org/#Replaceable) extended attribute, meaning that their setter will redefine the property as a data property with the set value, rather than changing some internal state. Deno currently defines them as read-only data properties instead.

Given that Web IDL requires all attributes to be accessor properties rather than data properties, but Deno exposes almost all of those properties as either read-only or writable data properties, it makes sense to expose `[Replaceable]` properties as writable as well – as is already the case with `WindowOrWorkerGlobalScope`'s `performance` property.


andreubotella:
> Do we have a test that will verify if worker will still be working once you replace `self`?

`self` in workers isn't replaceable, just on the main thread. And replacing it there shouldn't break anything, but it's still good to add a test to make extra sure.

PR #12383, opened 2021-10-09
fix(cli): support windows-style line endings in doc checker

### LINKED ISSUE #12187: `deno test --doc` doesn't detect errors in Markdown files with CRLF line endings
When I make a new `.md` file with some code that should throw a type error:

\```ts
let str: string = 1;
\```

If the file has CRLF line endings, running `deno test --doc` will not detect any problems.

Switching to LF works properly and throws the expected error:
```
```
Deno 1.14.0

GJZwiers:
This updates the regex patterns that identify blocks of code in documentation to include carriage returns `/r`. This fixes #12187 and should enable CRLF-style docs to be typechecked with `deno test --doc` as well.


bartlomieju:
Thanks, but we need some test to make sure there's no regression.

GJZwiers:
> 
> 
> Thanks, but we need some test to make sure there's no regression.

Does that mean adding some unit or integration tests? Or is there more involved? I would be happy to give it a try

PR #12384, opened 2021-10-09
chore: remove credentials typings from fetch

crowlKats:
Deno doesn't handle the `credentials` option, so the inclusion of typings is miss-informing. Also in the case of `Request`'s readonly `credentials` property, which isnt set at all, deno's behaviour conflicts with the typings

### DIFF
--- a/ext/fetch/lib.deno_fetch.d.ts
+++ b/ext/fetch/lib.deno_fetch.d.ts
@@ -145,7 +145,6 @@ type RequestCache =
   | "no-store"
   | "only-if-cached"
   | "reload";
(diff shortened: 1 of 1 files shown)


MierenManz:
wouldn't this technically be a breaking change?

lucacasonato:
This makes Deno less compatible with ts code targeting browsers and Deno so we won't land it. Thanks though 🙂

crowlKats:
øne problem though is we dont set the reaedonly `credentials` property, so its `undefined` when accessed, which doesnt match typings at all

PR #12386, opened 2021-10-09
refactor(metrics): move to core

AaronO:
Avoids overhead of wrapping ops (extra alloc due to reboxing futures for async ops) also stabilizes `Deno.metrics().ops`.

I'll deprecate byte fields (legacy of previous buffer-based op-layer) in a follow-up PR

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -1792,7 +1792,7 @@ declare namespace Deno {
    * Requires `allow-write` permission. */
   export function truncate(name: string, len?: number): Promise<void>;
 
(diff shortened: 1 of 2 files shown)


bartlomieju:
The type declarations are still not updated for `OpMetrics` (ie. they are still in `lib.deno.unstable.d.ts`)

AaronO:
> The type declarations are still not updated for `OpMetrics` (ie. they are still in `lib.deno.unstable.d.ts`)

Should we stabilize them after stripping the byte fields (and other changes I would like to make) or stabilize `.ops` now and keep the breaking changes for 2.0 ?

PR #12403, opened 2021-10-12
chore: upgrade crates based on deno ast 0.3

dsherret:


dsherret:
I don't understand why `cargo test steps_invalid_usage` suddenly started failing. It worked fine for me before I commited (maybe I made a change after?), but now it's not matching the pattern and failing on the op sanitizer:

```
parallel steps with sanitizers
AssertionError: Test case is leaking async ops.
Before:
  - dispatched: 6
  - completed: 5
After:
  - dispatched: 8
  - completed: 8
Ops:
  op_global_timer:
    Before:
      - dispatched: 6
      - completed: 5
    After:
      - dispatched: 8
      - completed: 8
```

PR #12405, opened 2021-10-12
docs(README.md): refactor built in utilities text

fzn0x:
add navigation for more details instead of introducing two utilities only.

This commit may also produce possibility to change "built in tooling" text in deno.land manual to "built in utilities"


kt3k:
@fncolon Thank you for your contribution. Please sign the CLA from the above link

PR #12410, opened 2021-10-12
feat(cli): update to TypeScript 4.5

kitsonk:
This updates TypeScript 4.5 beta.

Release announcement: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/

Things we need to test/integrate:

- [`type` on import names](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#type-on-import-names)
- [Import assertions](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#import-assertions)
- [add `mts` extensions](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#new-file-extensions)


kitsonk:
Updating to RC.

Announced here: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-rc/

The biggest change was the removal of the ESM support for Node12. This generally doesn't impact us, since we have our own resolution. It seems like the additional extensions (e.g. `.mts`) are still present.  I will add the changes here to support those (already being supported in deno_graph) in a follow-up commit to this branch.

kitsonk:
Updated to the released version.  Want to create some tests for the new features that overlap with swc as well to make sure everything works fine as well as "enable" the other file types that are being introduced (e.g. `.mts`).

bartlomieju:
@kitsonk any chance we could land this PR this week? It blocks me in #12866

kitsonk:
> @kitsonk any chance we could land this PR this week? It blocks me in #12866

It is blocked by the swc upgrade (@dsherret)

kitsonk:
@dsherret I think the tests will pass, but there is a bit of a catch-22 with dprint, in that dprint panics on the private field detection, which it is trying to do the ast parse, but can't see the `// deno-fmt-ignore` pragma to skip it, because it panics when parsing.

dsherret:
@kitsonk update the version of the typescript plugin in `~./.dprint.json`. I just forgot to update it in my PR.

kitsonk:
@dsherret now dlint is updated as well, and there were some changes for the normal pass, but it seems now with the `prefer-primordials` pass, it is not only running that rule but every other rule as well, which it shouldn't (as well as ignoring the `deno-ignore` directives.

ext/fetch/internal.d.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-// deno-lint-ignore-file no-explicit-any
+// deno-lint-ignore-file no-explicit-any no-var
```
bartlomieju, line 3: `no-var` should presumably be not applied to `d.ts` files? Or is it one of the things users shouldn't do but we still want to do in our declaration files?

kitsonk: It is subjective in `.d.ts` files as it really is only lib files that it is required so things are globally scoped, and most people don't author lib files, but either way, unrelated to this PR.

cli/tests/unit/version_test.ts
```diff
@@ -8,6 +8,7 @@ Deno.test(function version() {
   assert(
     pattern.test(Deno.version.typescript) ||
       Deno.version.typescript === "0-dev" ||
-      Deno.version.typescript === "0-beta",
+      Deno.version.typescript === "0-beta" ||
+      Deno.version.typescript === "1-rc",
```
bartlomieju, line 12: Why is this change required? Does it come from the time you were using one of `rc` versions of TS?

bartlomieju:
It seems `cli/dts/lib.es2021.intl.d.ts` was added in this PR, shouldn't `STATIC_ASSETS` in `cli/tsc.rs` be updated?

kitsonk:
There is a test for `import { type A } from "a.ts"` as well as tests checking the new media types.  Testing import assertions will be done under the import assertion features.

kitsonk:
> It seems `cli/dts/lib.es2021.intl.d.ts` was added in this PR, shouldn't `STATIC_ASSETS` in `cli/tsc.rs` be updated?

No, only assets which are not included by default do not go there, as per the comment:

```rust
/// Contains static assets that are not preloaded in the compiler snapshot.
```

And `lib.es2021.intl` is included in the snapshot.

PR #12424, opened 2021-10-13
feat(compat): Node CJS and ESM resolvers

bartlomieju:
So now all methods are ported, there are a few missing bits:
- [x] checking for loader shouldn't reqiuire --allow-net permission 
- [ ] cache read `package.json` (<s>need to figure out if Node does that globally or per-isolate</s> it's per isolate)
- [ ] 2 functions that emit deprecation warnings
- [ ] add moooooore tests
- [ ] figure out if the integration is fine, there is a single Deno specific conditional that short-circuits in "finalize_resolution" to handle HTTP(S) imports
- [ ] setup compat mode in different CLI subcommands
- [ ] setup compat mode in worker code (can wait until https://github.com/denoland/deno_std/pull/1151 lands)

### DIFF
--- a/cli/compat/mod.rs
+++ b/cli/compat/mod.rs
@@ -114,7 +114,6 @@ pub fn load_cjs_module(
   js_runtime: &mut JsRuntime,
   main_module: &str,
 ) -> Result<(), AnyError> {
-  println!("{}", main_module);
   let source_code = &format!(
     r#"(async function loadCjsModule(main) {{
       const Module = await import("{}");


bartlomieju:
I want to land this PR before 1.15.2 is released (so probably tomorrow), let me know if you want me to integrate this resolution in all possible subcommands or is it enough just for `deno run <file>` for the first pass.

The CI will stay red until new version of `deno_std` is released as one test checks against STD url. 

@kitsonk @dsherret I would appreciate reviews, especially around integration with `deno_graph` - I think this is still wrong and sometimes I see multiple `Check <deno_std>/global.ts` messages.

PR #12426, opened 2021-10-13
bench(cli): add deno_http_std.js

AaronO:
It has a few noteworthy differences with `deno_http_native.js`:
- It goes through the std/http wrappers
- Body isn't a pre-encoded buffer, it's a string
- It sets headers (`content-type: text/plain;charset=UTF-8`)

Overall it's currently about ~25% slower than `deno_http_native.js`


bartlomieju:
@AaronO can we land this PR?

Issue #12387, opened 2021-10-10
`FormData` is not on the typings of `globalThis`


rojserbest:
I get this error: `Property 'FormData' does not exist on type 'typeof globalThis'.`
When I try to `globalThis.FormData` in a TS file, but this is OK when done in the CLI.

PR #12429, opened 2021-10-13
fix: FormData is not on the typings of globalThis

F3n67u:
fix https://github.com/denoland/deno/issues/12387


F3n67u:
> @F3n67u thank you, I included your changes in #12497 and attributed you.

That’s  ok😊

PR #12432, opened 2021-10-13
feat(unstable/test): include test step pass/fail/ignore counts in final report

dsherret:
Perhaps this should just count the leaf test steps? That would be more complex though.


N8Brooks:
Originally I thought it would be like Mocha where only leafs are counted for the passing/failing count.

bartlomieju:
@dsherret can we land this PR?

PR #12438, opened 2021-10-14
fix: support verbatim UNC prefixed paths on Windows

### LINKED ISSUE #12398: Language server fails to load workspace-specific config (Windows, relative file path)
The LSP fails to load the config file. This happens on Windows only and with relative config file path (`../deno.json` in my case).

The attached screenshot shows:
  1) the actual error message `os error 3, cannot find the path specified`
  2) the resolved config file path, `file:///X:/home/ondras/rri/src/../deno.json`

The LSP itself works fine - it just cannot open and load the config file. My first suspect would be [to_file_path](https://github.com/denoland/deno/blob/main/cli/lsp/language_server.rs#L373) unable to deal with the *mixing-the-file-protocol-with-windows-paths* config file location.
vscode-error

dsherret:
The current code is stripping all UNC paths on Windows. We can't do this because not all UNC paths can safely be stripped.

For example, in #12398 it would take a path like `\\?\UNC\wsl$\Ubuntu\home\david\deno.json` and convert it to `UNC\wsl$\Ubuntu\home\david\deno.json`, which is not a valid path on Windows.

To solve this, the [dunce](https://crates.io/crates/dunce) crate has some special logic for handling when to strip unc paths, so I recommend we use it.

Closes #12398.

### DIFF
--- a/cli/fs_util.rs
+++ b/cli/fs_util.rs
@@ -83,10 +83,10 @@ fn strip_unc_prefix(path: PathBuf) -> PathBuf {
   match components.next() {
     Some(Component::Prefix(prefix)) => {
       match prefix.kind() {
-        // \\?\share_name\path
-        Prefix::Verbatim(share_name) => {
+        // \\?\device
+        Prefix::Verbatim(device) => {
           let mut path = PathBuf::new();
-          path.push(format!(r"\\{}\", share_name.to_string_lossy()));
+          path.push(format!(r"\\{}\", device.to_string_lossy()));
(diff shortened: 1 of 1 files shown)


piscisaureus:
> how do you think we should proceed?

For purposes of fixing this bug, keep using Rust's canonicalize() and fix up the path afterwards.
* If it starts with `\\?\UNC\`, remove that prefix.
* Otherwise, if it starts with `\\?\`, remove that prefix

That's all there is to it really.

PR #12444, opened 2021-10-14
feat(ext/crypto): support importing exporting AES JWK keys

littledivy:
Towards #11690 

Also includes crucial bug fixes for AES CryptoKey name mismatch.


ext/crypto/00_crypto.js
```diff
@@ -2640,6 +2492,202 @@
...
+          throw new DOMException("Invalid key usages", "DataError");
+        }
+
+        // 7.
+        // Section 4.3 of RFC7517
+        if (jwk.key_ops) {
+          if (
+            ArrayPrototypeFind(
+              jwk.key_ops,
+              (u) => !ArrayPrototypeIncludes(recognisedUsages, u),
```
bnoordhuis, line 2632: This is quadratic although I expect key_ops and recognisedUsages are both small?

Same observation a few lines below.

bnoordhuis (approved):

ext/crypto/00_crypto.js
littledivy: Yeah both are small, recognisedUsages.length is constant 8 and key_ops can be max 8

bartlomieju (approved):
LGTM too, sorry it took so long @littledivy; could you please rebase the PR?

PR #12451, opened 2021-10-15
perf(http): encode string bodies in op-layer

AaronO:
Optimizes cases where user provides string body, e.g:

```js
import { listenAndServe } from "https://deno.land/std/http/server.ts";
console.log("http://localhost:4500/");
listenAndServe(":4500", (_req) => new Response("Hello World\n"));
```

In `op_http_response` we now accept an `StringOrBuffer` type and map it `into_bytes() -> Vec<u8>` which is functionally identical (but faster obviously) than doing a `core.encode()` first.

## Benchmark

Running the above JS code (std http + string response body). I see about about a 5% improvement in total throughput (108k rps => 113k rps) driven by a 20% improvement in JS time spent (old path was ~600ms for 2M reqs, new path is ~100ms, now op-side)

## Notes

- Need to double check that we correctly encode the body if it's ever consumed JS-side instead of passed to `op_http_response`

### DIFF
--- a/ext/fetch/22_body.js
+++ b/ext/fetch/22_body.js
@@ -385,6 +385,9 @@
       stream = { body: source, consumed: false };
       length = source.byteLength;
     } else if (typeof source === "string") {
+      // WARNING: this deviates from spec (expects length to be set)
+      // https://fetch.spec.whatwg.org/#bodyinit > 7.
+      // no observable side-effect for users so far, but could change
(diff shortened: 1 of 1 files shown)


ry:
This PR resulted in a performance improvement in the "response_string" benchmark

Screen Shot 2021-12-01 at 8 33 19 AM

Issue #12308, opened 2021-10-03
Output cause on JS runtime errors


petamoriken:
Since Deno 1.13.0, Error Cause Proposal is supported, so the `Error` object can hold inner errors. However, their contents are usually not displayed and I need to explicitly access the `cause` property.

```js
function bar() {
  throw new Error("bar");
}

function foo() {
  try {
    bar();
  } catch (e) {
    throw new Error("foo", { cause: e });
  }
}

foo();
```

<img width="448" alt="terminal" src="https://user-images.githubusercontent.com/2622837/135757916-44ea1fd0-8f35-43fd-aa43-678dc466fc24.png">

I think it should also show inner errors, just like Firefox does.

<img width="316" alt="Firefox console" src="https://user-images.githubusercontent.com/2622837/135758170-4d4f379f-3e56-4ce6-b5a6-1c59a8cd9094.png">

bartlomieju:
Showing `Error.cause` when inspecting in `console` landed in v1.16 (https://deno.com/blog/v1.16#errorcause-is-now-displayed-in-the-console), we should add the same feature for errors that were returned by the runtime. It will require to modify `deno_core` and `cli` to also account for `cause` property in errors.

Most likely these functions needs to be updated:
https://github.com/denoland/deno/blob/1947f891db9fdd1a4aea2cfd55701a531f3c24e4/core/error.rs#L173-L200
https://github.com/denoland/deno/blob/1947f891db9fdd1a4aea2cfd55701a531f3c24e4/cli/fmt_errors.rs#L245-L280

PR #12462, opened 2021-10-16
feat(console): Display cause errors in console

petamoriken:
Related to #12308, however, this PR is not a runtime feature, but a console feature.

before:

after:


bartlomieju (approved):
LGTM, thanks @petamoriken

PR #12483, opened 2021-10-18
fix(ci): maybe workaround windows' STATUS_ACCESS_VIOLATION

AaronO:
Windows CI can be flaky, often failing on `runtime::tests::will_snapshot` with a `STATUS_ACCESS_VIOLATION`.

This might be related to accessing `v8::StartupData` after dropping its isolate (hinting at a deeper issue in v8/rusty_v8)

So this PR experiments with boxing (copying) the snapshot before the isolate is dropped

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -1956,10 +1956,10 @@ pub mod tests {
         ..Default::default()
       });
       runtime.execute_script("a.js", "a = 1 + 2").unwrap();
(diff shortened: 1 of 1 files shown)


ry:
Unfortunately this is going to be very difficult to say whether it works or not... This segfault is pretty rare.

piscisaureus:
Closing this as it does not solve the problem.

PR #12488, opened 2021-10-19
chore: upgrade to rustls 0.20

ry:
rustls 0.20 is a very large API change. 

Dependencies:
- [x] https://github.com/snapview/tokio-tungstenite/pull/198
- [x] https://github.com/seanmonstar/reqwest/issues/1363
- [x] https://github.com/rustls/hyper-rustls/pull/156


piscisaureus:
* I would like to land https://github.com/denoland/deno/pull/12467 before attempting to upgrade ext/net to use rustls 0.20
* Our websocket implementation uses tokio-tungstenite, but it can also use our own `TlsStream` from ext/net. So no need to make an effort to port over tokio-tungstenite.

satyarohith:
All the dependencies are now upgraded to `rustls 0.20`.

diceride:
Could you upgrade reqwest as well? reqwest switched to rustls 0.20.x in [0.11.7](https://github.com/seanmonstar/reqwest/releases/tag/v0.11.7)

diceride:
Please update reqwest as well, as they switched to rustls 0.20.x in [0.11.7](https://github.com/seanmonstar/reqwest/releases/tag/v0.11.7).

piscisaureus:
@bartlomieju 
I fixed the `TlsStreamInner::poll_io()` state machine so most tests pass now.
I believe the remaining issue is related to ALPN support.

bartlomieju:
> @bartlomieju I fixed the `TlsStreamInner::poll_io()` state machine so most tests pass now. I believe the remaining issue is related to ALPN support.

Thank you, I will fix the rest.

cli/tests/integration/mod.rs
```diff
@@ -1202,26 +1206,29 @@ async fn listen_tls_alpn_fail() {
...
-        webpki::DNSNameRef::try_from_ascii_str("localhost").unwrap();
-
+      let hostname = rustls::ServerName::try_from("localhost").unwrap();
       let tcp_stream = tokio::net::TcpStream::connect("localhost:4505")
         .await
         .unwrap();
       let mut tls_stream =
-        TlsStream::new_client_side(tcp_stream, &cfg, hostname);
+        TlsStream::new_client_side(tcp_stream, cfg, hostname);
       tls_stream.handshake().await.unwrap();
```
bartlomieju, line 1228: This call fails with:
```
thread 'integration::listen_tls_alpn_fail' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: UnexpectedEof, error: "tls handshake eof" }', cli/tests/integration/mod.rs:1228:36
```

@piscisaureus I'm a bit confused here, why did handshake previously succeeded if there were no matching ALPN protocols?

bartlomieju, line 1228: I'm still debugging what's the problem here - when reading from the socket there's a single errors raised (`WouldBlock`) that is ignored and then socket immediately returns 0 bytes read (meaning stream was closed of EOF was reached).

bartlomieju: I run `cli/tests/testdata/listen_tls_alpn.ts` manually and sent a small message after establishing a connection. Instead of `tls handshake eof` I'm not getting:
```
thread 'integration::listen_tls_alpn_fail' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: InvalidData, error: AlertReceived(NoApplicationProtocol) }', cli/tests/integration/mod.rs:1228:36
```
It appears to me `rustls` changed behavior of handshaking and our logic doesn't reflect that.

Still debugging...

piscisaureus: It's a timing issue. The server that this client connects to is implemented as follows:

```ts
  conn.close();
}
```

Basically it accepts the connection and hangs up immediately, before the handshake is complete.
If I change that to:

```ts
  await conn.handshake();
  conn.close();
}
```

... it gets a little further, however now the test fails with:

```
Uncaught (in promise) InvalidData: peer doesn't support any known protocol
  await conn.handshake();
  ^
    at async file:///Users/piscisaureus/d/deno/cli/tests/testdata/listen_tls_alpn.ts:11:3
```

So it seems that the `alpnProtocols: ["h2", "http/1.1", "foobar"]` is not propagated through the stack correctly.

bartlomieju: But this test adds only a single ALPN: `boofar` which indeed is not recognized by the server side. So to me it seems like the error you see is correct

piscisaureus: > @piscisaureus I'm a bit confused here, why did handshake previously succeeded if there were no matching ALPN protocols?

I don't know, I also think `handhake()` should fail if there are no matching ALPN protocols.
I've updated the tests to reflect that.

piscisaureus:
@bartlomieju 
It is green now but I think the PR still needs a final cleanup.
There's still this duplicated code, a `todo!()` comment, probably more.

bartlomieju:
> @bartlomieju It is green now but I think the PR still needs a final cleanup. There's still this duplicated code, a `todo!()` comment, probably more.

Sure, I'll handle that tomorrow.

bartlomieju:
@piscisaureus I addressed `todo!()` and removed some duplication of code (not all, but I left relevant comments as to why that is happening). Please review.

cli/proc_state.rs
```diff
@@ -226,8 +239,16 @@ impl ProcState {
       let mut reader = BufReader::new(certfile);
 
       // This function does not return specific errors, if it fails give a generic message.
```
piscisaureus, line 241: This comment is out of date.

cli/standalone.rs
```diff
@@ -222,8 +223,16 @@ pub async fn run(
   if let Some(cert) = metadata.ca_data {
     let reader = &mut BufReader::new(Cursor::new(cert));
     // This function does not return specific errors, if it fails give a generic message.
```
piscisaureus, line 225: This comment is out of date.

test_util/src/lib.rs
```diff
@@ -349,26 +355,32 @@ async fn get_tls_config(
...
+      let mut config = rustls::ServerConfig::builder()
+        .with_safe_defaults()
+        .with_client_cert_verifier(
+          rustls::server::AllowAnyAnonymousOrAuthenticatedClient::new(
+            root_cert_store,
+          ),
+        )
+        .with_single_cert(certs, PrivateKey(key))
+        .map_err(|e| {
+          eprintln!("Error setting cert: {:?}", e);
```
piscisaureus, line 371: Intentional? Did you mean `anyhow!("Error setting cert: {:?}", e)`.

piscisaureus (approved):
LGTM, some nits.

PR #12496, opened 2021-10-19
fix(cli/dts): update std links for deprecations

nayeemrmn:
Ref https://github.com/denoland/deno_std/issues/1391.


bartlomieju:
Thanks, could you please update `no-deprecated-deno-api` rule in `deno_lint` too?

PR #12554, opened 2021-10-26
fix(runtime): require full read and write permissions in order to create symlinks

### LINKED ISSUE #12152: [Security] Deno Sandbox Escape

dsherret:
This change requires full write and read permissions in order to create a symlink. The reason is to prevent multiple scenarios where someone could create a symlink to escape the directories they have restricted permissions to. For example:

1. Creating a symlink target in a directory outside the sandbox then reading from or writing to that folder.
2. Creating a valid relative symlink like `../../etc` and then moving an ancestor directory so the descendant symlink now points outside the sandbox.

This does not prevent an application causing trouble with existing relative symlinks (ex. `../../etc`) in a directory with write permissions where the write permissions are limited to certain folders.

Closes #12152

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2347,7 +2347,7 @@ declare namespace Deno {
    * Deno.symlinkSync("old/name", "new/name");
    * ```
    *
-   * Requires `allow-write` permission. */
+   * Requires full `allow-read` and `allow-write` permissions. */
   export function symlinkSync(
     oldpath: string | URL,
     newpath: string | URL,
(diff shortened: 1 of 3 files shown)


piscisaureus:
@caspervonb 
> This is the same as what Bartek did a few weeks ago isn't it? Same caveats applies.

It's not the same thing at all. The previous attempt relied on verifying permissions of the symlink target, which I agree doesn't solve the problem because a symlink can be moved.

This PR just makes it impossible to create symlinks unless you have global read/write permissions.
I'm open to a better fix (and performance isn't my primary concern) and I'm sure we'll eventually get to that.
However that "better fix" turned out to be quite complicated and in the meantime we have this gaping hole in our sandbox. 

@nayeemrmn 
> Edit: I guess symlinks can be moved by recursion as well...

That's the problem: you can put a symlink in a directory and then move the directory.
You can also make a symlink, and then a second symlink that points at the first one, and then delete the first one.

piscisaureus (approved):

PR #12591, opened 2021-10-29
feat(runtime): give OS errors .code attributes

AaronO:
This adds `.code` attributes to errors returned by the op-layer, facilitating classifying OS errors and helping node-compat.

Similar to Node, these `.code` attributes are stringified names of unix ERRNOs, the mapping tables are generated by `tools/codegen_error_codes.js` and derived from `libuv` and rust's `std` internals


piscisaureus:
> Non exhaustive first-pass. Might help with node-compat work

My suggestion would be:
* Do the the error code lookup in Rust
* Look at the system error code using [std::io::Error::raw_os_error()](https://doc.rust-lang.org/std/io/struct.Error.html#method.raw_os_error).
* On unix you want to map like this: 
  ```rust
   let code: &'static str = match e.raw_os_error() {
     libc::ENOENT { "ENOENT" },
     libc::EISDIR { "EISDIR" },
     // etc...
   };
   ```
   The error codes that might appear in node are [defined in libuv](https://github.com/libuv/libuv/blob/v1.x/include/uv/errno.h).
* On windows, use libuv's [mapping table](https://github.com/libuv/libuv/blob/v1.x/src/win/error.c#L71-L171).

AaronO:
@bartlomieju @piscisaureus PTAL

core/error_codes.rs
```diff
@@ -0,0 +1,212 @@
+use crate::error::AnyError;
+
+pub(crate) fn get_error_code(err: &AnyError) -> Option<&'static str> {
+  err
+    .downcast_ref::<std::io::Error>()
+    .map(|e| match e.raw_os_error() {
+      Some(code) => get_os_error_code(code),
+      None => get_io_error_code(e),
```
piscisaureus, line 8: If there is no OS error code, I don't think node would map to any of the `code`s that are defined in this PR.

It assigns other `code`s though (this one from OpenSSL for example):

```
piscisaureus@dumbo:~/d/deno 🦕 node
Welcome to Node.js v17.0.1.
Type ".help" for more information.
> void tls.connect(80, "deno.land").on("error", e => console.log(e.code))
undefined
> ERR_SSL_WRONG_VERSION_NUMBER
```

piscisaureus: IOW we should probably remove this branch and assign no code at all.

piscisaureus (approved):
LGTM, please consider comments.

core/error_codes.rs
AaronO: This branch is required, `std::io::Error` is really just an enum of:
```rs
enum Repr {
    Os(i32),
    Simple(ErrorKind),
    // &str is a fat pointer, but &&str is a thin pointer.
    SimpleMessage(ErrorKind, &'static &'static str),
    Custom(Box<Custom>),
}
```

Without this `ErrorKind` classification, `Simple` & `SimpleMessage` errors don't get codes and it looks like rust's `std` sometimes creates these "synthetic" errors instead of passing through or emulating OS ones. (e.g: when you fail an operation based on stat info instead of doing the actual syscall)

I tried removing it and it breaks node apps like `yarn` which need these error codes during their config probing phase.

PR #12595, opened 2021-10-29
chore: upgrade deno_ast to 0.5.0

### LINKED ISSUE #12560: deno fmt does not properly handle Markdown lists with internal structure
The following is almost definitely wrong:

```markdown
- This is a list.

        This is a code block in the list.
```

becomes (after `deno fmt`):

```markdown
- This is a list.

          This is a code block in the list.
```

Two extra spaces have been added to the code block for some reason.

___

This is *possibly* wrong (fine by CommonMark, but different than the original Markdown syntax description):

```markdown
- This is a list.

    This paragraph continues the list.
```

becomes (after `deno fmt`):

```markdown
- This is a list.

  This paragraph continues the list.
```

(CommonMark is okay with a two‐space indent for continuing paragraphs, but the original Markdown syntax description demanded four.)

### LINKED ISSUE #12565: Supports "Ergonomic brand checks for Private Fields" in TypeScript
It seems that "[Ergonomic brand checks for Private Fields](https://github.com/tc39/proposal-private-fields-in-in)", which is stage 4 in TC39, is not yet available in Deno.

The following code in a [v8 blog post](https://v8.dev/blog/v8-release-91#private-brand-checks-a.k.a.-%23foo-in-obj) works on chrome, but not on Deno.
```ts
class A {
  static test(obj: any) {
    console.log(#foo in obj);
  }
  #foo = 0;
}

A.test(new A()); // true
A.test({}); // false
```

chrome:
image

Deno:
image

The version used is as follows.
```
>deno --version
deno 1.15.3+b44b26c (canary, x86_64-pc-windows-msvc)
v8 9.5.172.19
typescript 4.4.2
```

dsherret:
Closes #12560 via https://github.com/dprint/dprint-plugin-markdown/commit/fec0f2adc9559a297e93240c3b5eb3b138c8178a

Closes #12565


cli/file_fetcher.rs
```diff
@@ -884,7 +884,7 @@ mod tests {
       (
         "https://deno.land/x/mod.d.ts",
         Some("application/javascript".to_string()),
-        MediaType::JavaScript,
+        MediaType::Dts,
```
dsherret, line 887: See https://github.com/denoland/deno_ast/pull/33

dsherret:
@kitsonk I will just do the media type changes necessary for this PR right now or did you already start on it?

kitsonk:
No I didn't start on it.

serde_v8/README.md
```diff
@@ -48,10 +48,10 @@ when encoding/decoding.
 - [ ] Experiment with KeyCache to optimize struct keys
 - [ ] Experiment with external v8 strings
 - [ ] Explore using
-  [json-stringifier.cc](https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/json/json-stringifier.cc)'s
-  fast-paths for arrays
+      [json-stringifier.cc](https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/json/json-stringifier.cc)'s
```
dsherret, line 51: This is what prettier does with task lists.

bartlomieju (approved):
LGTM

kitsonk (commented):
I'm a little bit worried about sending in `.mts` etc to tsc 4.4 (the get as tsc extension stuff) both on check and the lsp, and we won't merge 4.5 for a while (not until 1.17).

I think we need to see what happens when we have `.mjs`, `.mts` files under 4.4, I think it will break, and if so, we need to translate them back to `.js` and `.ts` until 4.5 lands.  At the least I think we need some tests.

Also, I don't know how vscode is handling them, I assume they are going to pass a different language ID, but I haven't looked into it yet.

dsherret:
@kitsonk I forgot about TS 4.5 support landing after the release and thought this would just be resolved when your TS 4.5 PR landed, but yeah looks like it won't come until mid November as I now remember you mentioning before. I've commented out the code in tsc and the lsp and mapped those extensions to either JavaScript, TypeScript, or Dts so it will use those as syntax kinds.

kitsonk (approved):
LGTM

PR #12596, opened 2021-10-29
feat(core): streams

### LINKED ISSUE #12556: StreamResource in core
op_read/op_write/op_shutdown works for any stream type

AaronO:
This allows resources to be "streams" by implementing read/write/shutdown, returning `None` if not implemented and `Some(future)` otherwise.

These streams are implicit since their nature (read/write/duplex) isn't known until called, but we could easily add another method to explicitly tag resources as streams.

op_read/op_write/op_shutdown are now builtin ops provided by `deno_core`

**Note:** this current implementation is simple & straightforward but it results in an additional alloc per read/write call

Closes #12556


core/examples/http_bench_json_ops.rs
```diff
@@ -78,19 +79,33 @@ struct TcpStream {
...
-  async fn write(self: Rc<Self>, buf: &[u8]) -> Result<usize, Error> {
+  async fn write(self: Rc<Self>, buf: ZeroCopyBuf) -> Result<usize, AnyError> {
     let mut wr = RcRef::map(self, |r| &r.wr).borrow_mut().await;
-    wr.write(buf).await
+    wr.write(&buf).await.map_err(AnyError::from)
   }
 }
 
 impl Resource for TcpStream {
+  fn read(self: Rc<Self>, buf: ZeroCopyBuf) -> Option<AsyncResult<usize>> {
```
piscisaureus, line 101: I think I would prefer the return type to be just AsyncResult.
The default implementation can return some sort of "not supported" error.

core/resources.rs
```diff
@@ -8,16 +8,23 @@
...
 use std::iter::Iterator;
+use std::pin::Pin;
 use std::rc::Rc;
 
+/// Returned by resource read/write/shutdown methods
+pub type AsyncResult<T> = Pin<Box<dyn Future<Output = Result<T, AnyError>>>>;
+
 /// All objects that can be store in the resource table should implement the
 /// `Resource` trait.
+/// TODO(@AaronO): investigate avoiding alloc on read/write/shutdown
```
piscisaureus, line 27: I am pretty certain that boxing the returned trait is unavoidable with trait methods.
It'd be nice though if we could avoid double boxing.

core/resources.rs
```diff
@@ -27,6 +34,21 @@ pub trait Resource: Any + 'static {
     type_name::<Self>().into()
   }
 
+  /// Resources may implement `read()` to be a readable stream
+  fn read(self: Rc<Self>, _buf: ZeroCopyBuf) -> Option<AsyncResult<usize>> {
```
piscisaureus, line 38: It is possible to make this more ergonomic by using the [async_trait] proc macro (which btw also boxes the promise returned by an async function, hidden from view)

core/examples/http_bench_json_ops.rs
AaronO: One advantage of the `Option<AsyncResult<...>>` approach is that when shutting down a parent resource we can `filter_map` its children and shutdown that collection error-ing in `shutdown()` would make that trickier or require a "can shutdown" tag

core/resources.rs
AaronO: I had considered it, but thought it was best to avoid the extra dep and "macro magic" since there's only a handful of call sites that shouldn't frequently change

core/examples/http_bench_json_ops.rs
piscisaureus: You can also make the default implementation return Ok() if thats what you're after.
That might be ok (pun not intended) in the specific case of shutdown - I have to think about that.

But I think we should discuss use cases and potential merit of parent/child resources first.
Do you have an example in mind of a resource that would benefit from having child resources?

AaronO: Http is the main example, we currently track child resources JS-side. Http has a rank 3 hierarchy (listener > conn > req/resp), other listeners are rank 2.

That would work, it would just be theoretically less efficient

piscisaureus: > Http is the main example, we currently track child resources JS-side.

I am no longer convinced HTTP really needs child resources though. Potentially between the TCP listener and the HttpConn, but I suspect you were thinking of HttpConn => HttpStream as parent-child.

The filter_map shutdown behavior isn't appropriate for HttpCon resources; `shutdown` has a specific meaning - it tells the receiving end of a connection that the sender has sent all the data it had intended to send (rather than e.g. the connection was closed due to a crash or connection error). HttpConn, which behaves like a listening socket rather than a data stream, doesn't have a shutdown method.

We could consider expanding the meaning of `shutdown()` to mean `close gracefully` in a more general way.
In that case `HttpConn` could have a shutdown method, which would make it stop accepting new HTTP requests while letting already-accepted ones finish up -- however that would not involve calling shutdown on its child resources.

runtime/js/12_io.js
```diff
@@ -102,7 +102,7 @@
       return 0;
     }
 
-    const nread = await core.opAsync("op_read_async", rid, buffer);
+    const nread = await core.read(rid, buffer);
 
     return nread === 0 ? null : nread;
   }
```
bartlomieju, line 108: I believe this two APIs should be completely removed in favor of `core.read` and `core.write`

bartlomieju (commented):
What about sync APIs?

AaronO:
> What about sync APIs?

Since sync APIs are only use for file IO (on `StdFileResource`) I was planning to rename them to `op_fread_sync`, `op_fwrite_sync`, I don't think those sync APIs should be in core

bartlomieju:
> > What about sync APIs?
> 
> Since sync APIs are only use for file IO (on `StdFileResource`) I was planning to rename them to `op_fread_sync`, `op_fwrite_sync`, I don't think those sync APIs should be in core

Okay, keep in mind that `StdFileResource` is used for stdio too.

core/examples/http_bench_json_ops.rs
AaronO: I'll drop the option, it's not difficult to change this if needed down the line

AaronO: Actually I would rather keep the Option for now change it if necessary in a follow-up PR.

It could be more ergonomic to drop the Option if other ops take a readable resource and are doing their own `.read()` calls on them so it's an at runtime `AsyncRead` of sorts

runtime/js/12_io.js
AaronO: There's some other rewiring of read/write/shutdown exports to do, I would rather do that in a scoped follow-up PR

piscisaureus (commented):
I still don't see why these ops should return `Option<AsyncResult<_>>` rather than `AsyncResult<_>`. 
IMO that's just un-ergonomic and the "op not supported by this resource type" case really doesn't need any optimization.
Or am I missing something?

AaronO:
> I still don't see why these ops should return `Option<AsyncResult<_>>` rather than `AsyncResult<_>`. IMO that's just un-ergonomic and the "op not supported by this resource type" case really doesn't need any optimization. Or am I missing something?

I think the "ergonomics" of the trait weren't a priority for a first pass given the limited amount of callers. The plan was to tweak and evolve this. I still generally don't like the idea that we have to peek into errors to know if the stream was readable or not. But regardless I've dropped the `Option<...>` from the trait in the latest commit

core/ops_builtin.rs
```diff
@@ -36,6 +37,10 @@ pub(crate) fn init_builtins() -> Extension {
       ("op_metrics", op_sync(op_metrics)),
       ("op_void_sync", void_op_sync()),
       ("op_void_async", void_op_async()),
+      // TODO(@AaronO): track IO metrics for builtin streams
```
bartlomieju, line 40: Is there a significant problem with tracking metrics?

AaronO: Not really, there's some debate about how to aggregate/report them but otherwise it's straightforward, but it doesn't fit in the scope of a first-pass IMO

piscisaureus (approved):
LGTM!

PR #12604, opened 2021-10-30
fix(cli/proc_state): Short-circuit in prepare_module_load()

### LINKED ISSUE #12519: Dynamic (remote) import is quite slow and seems not cached
I'm not sure if it's known but it seems remote dynamic import is quite slow. For example

```typescript
async function main() {
  const t = performance.now();
  await Promise.all([
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
  ]);
  console.log(performance.now() - t);
}
await main();
```

Above took approx. 55 ms with `deno run -A --no-check ./test.ts`. It seems the number of remote imports in the target file affect the performance.

Additionally, it seems remote dynamic import does not use cache. For example

```typescript
// test.ts
async function main() {
  const t = performance.now();
  await Promise.all([
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
  ]);
  console.log(performance.now() - t);
}
await main();
```

Above took approx. 500 ms with `deno run -A --no-check ./test.ts` even the URL is just repeated.

### LINKED ISSUE #12740: [deno run] There is a pattern of not detecting errors when the `deno run` command is reading a file with syntax errors.
## Summary
There is a pattern of not detecting errors when the `deno run` command is reading a file with syntax errors.

## My environment

```sh
$ deno --version
deno 1.16.0 (release, aarch64-apple-darwin)
v8 9.7.106.2
typescript 4.4.2
```

### Step 1

Create `main.ts`.

```typescript
console.log('Hi Deno');
```

Then run this script.

```sh
deno run main.ts
```

The results are as follows.

```sh
Hi Deno
```

### Step 2

The next step is to create `sub.ts`.
(This code has a syntax error.)

```typescript
// This code is syntax error
export function hello(name: string) {
  return hello name;
}
```

Modify `main.ts` to load `sub.ts`.

```typescript
import { hello } from './sub.ts';
console.log(hello('Deno'));
```

Then run this script.

```sh
deno run main.ts
```

The results are as follows.

```sh
Hi Deno
```

I thought this code of mine would not work due to a syntax error, but it runs fine.
And the result of execution is the result before editing.

By the way, when I did `deno compile`, it showed a syntax error.

```sh
$ deno compile main.ts
error: The module's source code could not be parsed: Expected ';', got 'name' at file:///Users/~/sub.ts:3:16
```

## Question

Is this behavior correct?

I still don't understand the Deno spec and the internal code exactly, but it seems to me that a syntax error would be correct in this case. That's why I created the Issue.

I'd really appreciate it if you could give me your opinion .

Thank you!

nayeemrmn:
Additionally improve `ProcState::prepare_module_load()` to detect all errors it is supposed to, so no error handling needs to be done in `ProcState::load()` for supposedly prepared modules.

Fixes #12519.
Fixes #12740.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -782,9 +782,9 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d2d76b6b75a6fbfda0f529e310fc3cab960f4219403280b430ce93dcf8cf9a2"
+checksum = "075b0c1b454eaf90cea9c6efc72ff946aa6c855c85a4209cb717c01424b37e5e"
 dependencies = [
  "cfg-if 1.0.0",
  "deno_ast",
@@ -827,9 +827,9 @@ dependencies = [
 
 [[package]]
 name = "deno_graph"
-version = "0.11.1"
+version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6d84ddee0cf83bf295721be792b6769b92214983bda29d52c2b05a89d1e968f"
+checksum = "9ee63197c67746c40911cb3082ca13a29cc5adae1ff1b706397b44f7155d7c57"
 dependencies = [
  "anyhow",
  "cfg-if 1.0.0",
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -41,8 +41,8 @@ winres = "0.1.11"
 [dependencies]
 deno_ast = { version = "0.5.0", features = ["bundler", "codegen", "dep_graph", "module_specifier", "proposal", "react", "sourcemap", "transforms", "typescript", "view", "visit"] }
 deno_core = { version = "0.107.0", path = "../core" }
-deno_doc = "0.20.0"
-deno_graph = "0.11.1"
+deno_doc = "0.21.0"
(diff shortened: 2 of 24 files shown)


nayeemrmn:
Blocked on https://github.com/denoland/deno_graph/pull/61.

nayeemrmn:
@kitsonk Please review

kitsonk:
Given that we have a very specific repo in #12740 and existing tests didn't cover it, it feels like something we should add a similar test for it.  @nayeemrmn let me know if you want to tackle it, or if not, I can add a specific test for that issue.

nayeemrmn:
Added a repro for #12740 as a test.

kitsonk (approved):
LGTM, @nayeemrmn awesome work 💯

Issue #12609, opened 2021-10-30
[Unit Tests] Human Readable Timer For Long Tests


Skillz4Killz:
Hello 👋 

As a bit of feedback with a lot of tests, the group test shows `ms` that are really high it would be nice to convert this maybe to like 1m50s etc

```ts
running 16 tests from file:///workspaces/discordeno/tests/mod.ts
test [Local Test] - snowflakeToBigint  ... ok (12ms)
test [Local Test] - bigIntToSnowflake ... ok (8ms)
test [utils] Validate length is too low ... ok (8ms)
test [utils] Validate length is too high ... ok (8ms)
test [utils] Validate length is NOT just right in between. ... ok (9ms)
test [utils] Validate length is NOT too low ... ok (8ms)
test [utils] Validate length is NOT too high ... ok (7ms)
test [utils] Validate length is just right in between. ... ok (7ms)
test [utils] hasProperty does HAVE property ... ok (7ms)
test [utils] hasProperty does NOT HAVE property ... ok (7ms)
test [utils] delay 2000 ms ... ok (2009ms)
test [utils] icon hash to bigint ... ok (7ms)
test [utils] icon bigint to hash ... ok (7ms)
test [utils] icon hash to bigint a_ (animated) ... ok (8ms)
test [utils] icon bigint to hash a_ (animated) ... ok (7ms)
test [Bot] - Starting Tests ...Bot online


  test Channel related tests ...
    test Message related tests ...
      test [message] send message with text ... ok (117ms)
      test [message] send message with embeds ... ok (170ms)
      test [message] delete message without a reason ... ok (468ms)
      test [message] delete message with a reason ... ok (538ms)
      test [message] delete messages without a reason ... ok (10956ms)
      test [message] delete messages with a reason ... ok (14506ms)
      test [message] fetch a message ... ok (10926ms)
      test [message] fetch messages ... ok (15690ms)
      test [message] add a reaction ... ok (15554ms)
      test [message] add a custom reaction ... ok (11011ms)
      test [message] add multiple reactions ... ok (6815ms)
      test [message] add multiple custom reactions ... ok (7802ms)
      test [message] add multiple reactions in order ... ok (6373ms)
      test [message] add multiple custom reactions in order ... ok (8017ms)
    ok (15700ms)
  ok (15765ms)
ok (21617ms)

test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (27292ms)
```

Thank You

PR #12610, opened 2021-10-30
feat(test): better formatting for test elapsed time

bartlomieju:
Closes https://github.com/denoland/deno/issues/12609

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -249,7 +249,7 @@ impl PrettyTestReporter {
     println!(
       "{} {}",
       status,
-      colors::gray(human_elapsed(elapsed)).to_string()
+      colors::gray(human_elapsed(elapsed.into())).to_string()
     );
 
     if let Some(error_text) = result.error() {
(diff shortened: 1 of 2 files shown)


cli/tools/test.rs
```diff
@@ -260,6 +260,22 @@ impl PrettyTestReporter {
...
+  if elapsed < 1_000 {
+    return format!("{}ms", elapsed);
+  }
+  if elapsed < 1_000 * 60 {
+    return format!("{}s", (elapsed / 1000).floor());
+  }
+
+  let seconds = (elapsed / 1_000).floor();
+  let minutes = (seconds / 60).floor();
+  let seconds_reminder = seconds - (minutes * 60);
```
dsherret, line 275: `second % 60`

cli/tools/test.rs
```diff
@@ -249,7 +249,7 @@ impl PrettyTestReporter {
     println!(
       "{} {}",
       status,
-      colors::gray(format!("({}ms)", elapsed)).to_string()
+      colors::gray(human_elapsed(elapsed)).to_string()
```
dsherret, line 252: I think this should also be done for test reporting?

dsherret (approved):
LGTM, but probably needs it for test case reporting too

bartlomieju:
> LGTM, but probably needs it for test case reporting too

Yeah, not a big deal for "ms" and "s", but I don't want to add a test that takes more than a minute just to assert it outputs (1m1s) 😬

dsherret:
@bartlomieju we could unit test `human_elapsed`. I don't think testing the structs at a higher level/integration tests are necessary.

bartlomieju:
> @bartlomieju we could unit test `human_elapsed`. I don't think testing the structs at a higher level/integration tests are necessary.

SGTM, I'll do that.

Issue #11731, opened 2021-08-16
Microtask queue is not handled correctly in Deno


lucacasonato:
In browsers, a microtask checkpoint is performed whenever the JS stack empties. Take the following example:

```js
const filereader = new FileReader();

filereader.onloadstart = () => {
  console.log("loadstart event");
  queueMicrotask(() => console.log("loadstart microtask"));
};
filereader.onloadend = () => {
  console.log("loadend event");
  queueMicrotask(() => console.log("loadend microtask"));
};
filereader.onload = () => {
  console.log("load event");
  queueMicrotask(() => console.log("load microtask"));
};
filereader.onprogress = () => {
  console.log("progress event");
  queueMicrotask(() => console.log("progress microtask"));
};

filereader.readAsText(new Blob(["hello"]));
queueMicrotask(() => console.log("after start microtask"));
```

In all current browsers, `FileReader` is implemented as a native object. That means when you call `filereader.readAsText`, the following happens:

1. When calling `filereader.readAsText`, JS calls into native C++ code and queues some tasks (things that should happen on future spins of the event loop) to emit the `loadstart`, `progress`, `load`, and `loadend` events in that order. The C++ code returns back into JS. 
2. The `queueMicrotask` call pushes one microtask onto the microtask queue.
3. At this point there is no more JS to be executed, and the JS stack is empty. Thus a microtask checkpoint is performed. `after start microtask` is printed to the console. The microtask queue is now empty.
4. In C++ code, the event loop takes the first ready task from the task queue. The task is "dispatching the loadstart event". The event is dispatched. This calls from C++ into JS. Something is now on the JS stack. `loadstart event` is logged, and a microtask is queued.
5. There is no more JS to be executed, and the JS stack is empty. Thus a microtask checkpoint is performed. `loadstart microtask` is  printed to the console. The microtask queue is now empty.
6. The same happens for the progress, load, and loadend events.

This is how the browser console looks after executing this:

```
after start microtask
loadstart event
loadstart microtask
progress event
progress microtask
load event
load microtask
loadend event
loadend microtask
```

Now lets see what happens in Deno. We currently also follow the rule that a microtask checkpoint is performed whenever the JS stack empties. In Deno `FileReader` is implemented in JS code:

1. When calling `filereader.readAsText`, JS calls into JS code and queues some tasks emit the `loadstart`, `progress`, `load`, and `loadend` events in that order. The JS code returns back to the calling JS code.
[...]

PR #12616, opened 2021-10-31
feat(ext/web): BYOB support for ReadableStream

### LINKED ISSUE #11025: Support BYOB ReadableStream

crowlKats:
Closes #11025


kitsonk (commented):
Are there no WPT tests that need to be enabled?

crowlKats:
Yes, its not completely done yet. there are a few TODOs that hopefully will get resolved in the morning. once those are solved, I will enable all the WPT tests related to this

ext/web/06_streams.js
```diff
@@ -1019,112 +1193,629 @@
...
+   * @returns {void}
+   */
+  function readableByteStreamControllerFillHeadPullIntoDescriptor(
+    controller,
+    size,
+    pullIntoDescriptor,
+  ) {
+    assert(
+      controller[_pendingPullIntos].length === 0 ||
+        controller[_pendingPullIntos][0] === pullIntoDescriptor, // TODO: check
```
lucacasonato, line 1638: Looks correct. You doing a reference equality check here, not a value equality check, which looks to be what the spec intends.

ext/web/06_streams.js
```diff
@@ -1138,7 +1829,12 @@
     assert(signal === undefined || signal instanceof AbortSignal);
     assert(!isReadableStreamLocked(source));
     assert(!isWritableStreamLocked(dest));
-    const reader = acquireReadableStreamDefaultReader(source);
+    let reader;
+    if (source[_controller] instanceof ReadableByteStreamController) {
+      reader = acquireReadableStreamBYOBReader(source);
```
lucacasonato, line 1834: > If source.[[controller]] implements ReadableByteStreamController, let reader be either ! AcquireReadableStreamBYOBReader(source) or ! AcquireReadableStreamDefaultReader(source), at the user agent’s discretion.

I'll have to dig into that deeper. In what cases would a user agent use a DefaultReader?

ext/web/06_streams.js
```diff
@@ -3450,10 +4443,201 @@
...
+      return promise.promise;
+    }
+
+    /** @returns {void} */
+    releaseLock() {
+      webidl.assertBranded(this, ReadableStreamBYOBReader);
+      if (this[_stream] === undefined) {
+        return;
+      }
+      if (this[_readIntoRequests].length) {
```
lucacasonato, line 4527:
```suggestion
      if (this[_readIntoRequests].length !== 0) {
```

ext/web/06_streams.js
```diff
@@ -3450,10 +4443,201 @@
...
+        throw new TypeError(
+          "There are pending read requests, so the reader cannot be release.",
```
lucacasonato, line 4529:
```suggestion
          "There are pending read requests, so the reader cannot be released.",
```

lucacasonato (commented):
Ok, did an initial review. This looks very promising.

ext/web/06_streams.js
```diff
@@ -193,6 +207,20 @@
     return isFakeDetached in O;
   }
 
+  /**
+   * @param {ArrayBufferLike} O
+   * @returns {boolean}
+   */
+  function canTransferArrayBuffer(O) {
+    assert(typeof O === "object");
+    assert(O instanceof ArrayBuffer || O instanceof SharedArrayBuffer);
```
andreubotella, line 216: `SharedArrayBuffer`s aren't transferable, and I don't think this spec algorithm is meant to be passed a SAB.

ext/web/06_streams.js
```diff
@@ -1019,112 +1193,629 @@
...
+      view instanceof Uint8ClampedArray ||
+      view instanceof Int16Array ||
+      view instanceof Uint16Array ||
+      view instanceof Int32Array ||
+      view instanceof Uint32Array ||
+      view instanceof BigInt64Array ||
+      view instanceof BigUint64Array
+    ) {
+      elementSize = view.constructor.BYTES_PER_ELEMENT;
+      ctor = view.constructor;
```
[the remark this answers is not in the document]
crowlKats: true. not sure how we can do this better though, except a check for each individual typed array

ext/web/06_streams.js
crowlKats: Not sure to be honest

ext/web/06_streams.js
```diff
@@ -193,6 +207,20 @@
...
+   * @param {ArrayBufferLike} O
+   * @returns {boolean}
+   */
+  function canTransferArrayBuffer(O) {
+    assert(typeof O === "object");
+    assert(O instanceof ArrayBuffer || O instanceof SharedArrayBuffer);
+    if (isDetachedBuffer(O)) {
+      return false;
+    }
+    // TODO(@crowlKats): 4. If SameValue(O.[[ArrayBufferDetachKey]], undefined) is false, return false.
```
andreubotella, line 220: I believe this check is the same as [`v8::ArrayBuffer::is_detachable`](https://docs.rs/v8/0.33.0/v8/struct.ArrayBuffer.html#method.is_detachable). The use case for this seems to be to make it impossible to detach the buffer that backs a `WebAssembly.Memory` object other than through its `grow` method.

ext/web/06_streams.js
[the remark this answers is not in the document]
lucacasonato: It's probably fine. We can fix it if someone runs into it.

ext/web/06_streams.js
lucacasonato: Yes. I suggest we do this in a follow up, as it requires core bindings.

ext/web/06_streams.js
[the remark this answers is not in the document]
andreubotella: Instances of `MyArray` should still have a `[[TypedArrayName]]` internal slot of `"Int8Array"`. Now, that would not be if the constructor returns something other than `this`, or with `Object.create(MyArray.prototype)`, but that's even more obscure, so it's probably fine to just have a TODO.

tools/wpt/expectation.json
```diff
@@ -6219,231 +6149,21 @@
...
-        "calling respondWithNewView() twice on the same byobRequest should throw",
-        "calling respond(0) twice on the same byobRequest should throw even when closed",
-        "calling respond() should throw when canceled",
-        "pull() resolving should not make releaseLock() possible",
-        "ReadableStream with byte source: default reader + autoAllocateChunkSize + byobRequest interaction",
-        "ReadableStream with byte source: respondWithNewView() with a smaller view",
-        "ReadableStream with byte source: respondWithNewView() with a zero-length view (in the closed state)",
-        "ReadableStream with byte source: respondWithNewView() with a transferred non-zero-length view (in the readable state)",
-        "ReadableStream with byte source: respondWithNewView() with a transferred zero-length view (in the closed state)"
+        "ReadableStream with byte source: Respond to multiple pull() by separate enqueue()"
```
lucacasonato, line 6157: After investigation I am about 80% confident this test being broken is a result of our broken microtask queuing: https://github.com/denoland/deno/issues/11731

lucacasonato (approved):
Thank you @crowlKats. Awesome to have this implemented now!

Other than the one failing test due to microtask queuing, all WPT pass.

PR #12617, opened 2021-10-31
chore(cli): fix typo in `tools/test.rs`

meszarosdezso:
Just catched this in #12610, nice to have this formatting

Are milliseconds intentionally not shown when the test takes longer than a second (eg: `1s43ms`)?


kitsonk (approved):
LGTM

PR #12631, opened 2021-11-03
feat(cli): support React 17 JSX transforms

### LINKED ISSUE #8440: Support new JSX transforms
React 17 introduced [new JSX transforms](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) and they are available in [TypeScript 4.1](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#jsx-factories).  There are two new options for `"jsx"` in TypeScript:

- `"react-jsx"`
- `"react-jsxdev"`

Currently with _check_ these should be supported without any further changes, but are unsupported in `--no-check` and `deno bundle`.

We need support from swc before integrating.  Ref: https://github.com/swc-project/swc/issues/1103

cc/ @kdy1

kitsonk:
Closes #8440


cli/emit.rs
```diff
@@ -40,6 +40,8 @@ use std::result;
 use std::sync::Arc;
 use std::time::Instant;
 
+pub(crate) static DEFAULT_JSX_IMPORT_SOURCE: &str = "https://esm.sh/preact";
```
lucacasonato, line 43: Not in favour of having a default, much less a remote registry we have no control over.

lucacasonato (commented):
If I read this correctly, we have to do up to 5 remote fetches now before we know what file to use for the jsx import source? This is quite magical - not really our style. Also no web precedent for poking round the web like this. Also introduces a waterfall behaviour that can significantly slow down use-cases like JIT-at-the-edge transpilation.

kitsonk:
> If I read this correctly, we have to do up to 5 remote fetches now before we know what file to use for the jsx import source?

I don't like it either, but if you have some other way of determining it, I am open to suggestions.  It is _only_ for something that appears to be the `jsx-runtime` (or `jsx-dev-runtime`), and there are various forms already in the wild that need to be supported.  React chose an ambiguous module specifier, dependent upon Node resolution rules.

Some examples of "real" specifiers:

- `https://deno.land/x/nano_jsx@v0.0.21/jsx-runtime/index.ts`
- `https://cdn.skypack.dev/react/jsx-runtime`
- `https://esm.sh/preact/jsx-runtime`

We could say that we would only support situations where the web server resolves/redirects like Skypack and esm.sh do, but that would mean that nano_jsx would not work.  Also, local JSX libraries are unlikely to work, as people won't have an extensionless file.  We could then maybe say we only do resolution with local specifiers, and expect remote specifiers to be redirects, but again, it makes it impossible then for people to publish things to `deno.land/x/` that would work.

Is there an alternative approach I am missing?

cli/emit.rs
kitsonk: I can eliminate this. Need to figure out how to error if none is provided but there is a setting that requires it.

lucacasonato:
> Is there an alternative approach I am missing?

There are two things I can think of:
- we deviate from TSC's resolution logic, and require that users specify a full path. This could possibly be upstreamed to TSC.
- we keep TSC's default logic of always just appending `/jsx-runtime`, and have users use an import map when they want to use a local file. This is sorta what also happens in node nowadays, just not using an import map, but using `exports` in package.json.

I agree both of these are not great either, but at least they are very clear to what is happening. I think we should avoid probing the FS or remote server at all costs.

kitsonk:
> we deviate from TSC's resolution logic, and require that users specify a full path. This could possibly be upstreamed to TSC.

We can't... both swc and tsc emit a determined specifier that cannot be controlled.  I have tried to amend it before... it is effectively hard-coded.

For example a `jsxImportSource` of `react` emits `react/jsx-runtime` (or `react/jsx-dev-runtime` if it is dev mode).  It can't be "upstreamed" because it is the behaviour that React expects and Babel and TypeScript have complied with (as well as other transpilers).

> we keep TSC's default logic of always just appending /jsx-runtime, and have users use an import map when they want to use a local file. This is sorta what also happens in node nowadays, just not using an import map, but using exports in `package.json`.

That is the least worst option I guess, though this would not work for Deploy for something like nano_jsx.

lucacasonato:
> That is the least worst option I guess, though this would not work for Deploy for something like nano_jsx.

I wouldn't block ourselves on this. Deploy will support import maps eventually.

kitsonk:
> > That is the least worst option I guess, though this would not work for Deploy for something like nano_jsx.
> 
> I wouldn't block ourselves on this. Deploy will support import maps eventually.

Ok, I originally hand't realised that both skypack and esm.sh effectively resolved those for JSX libraries, which I think would cover 90% of the use cases, and then have an example for nano_jsx using an import map would be sufficient I think.

test_util/src/lib.rs
```diff
@@ -857,7 +880,9 @@ async fn main_server(req: Request<Body>) -> hyper::Result<Response<Body>> {
         return Ok(file_resp);
       }
 
-      return Ok(Response::new(Body::empty()));
+      Response::builder()
```
kitsonk, line 883: Strangly, the test server returned an empty body irrespective of the URL requested, instead of a 404 not found.  While this was specific to creating the "sniffing" behaviour reverted, it still feels like we should return a 404 like a "normal" webserver to ensure the code behaves properly.

cli/http_util.rs
```diff
@@ -123,11 +124,18 @@ pub async fn fetch_once(
 
   if response.status().is_client_error() || response.status().is_server_error()
   {
-    let err = generic_error(format!(
-      "Import '{}' failed: {}",
-      args.url,
-      response.status()
-    ));
+    let err = if response.status() == StatusCode::NOT_FOUND {
```
kitsonk, line 127: While this was done for the sniffing that was refactored out, it sill makes sense IMO, that we can distinguish between 404 Not Found errors and other response errors, like we can with local files.  Lumping everything in a generic error seems wrong.

kitsonk:
To do still:

- [ ] integrate into the lsp
- [ ] ensure `Deno.emit()` works

kitsonk:
There is an outstanding issue on swc (swc-project/swc#2656) that I discovered.  It means that:

- when someone uses `"react-jsxdev"` compiler option
- and they are using `--no-emit` or bundling
- Deno will error with something like "cannot load "jsx-runtime"`

The workaround is not to use `--no-emit` or not to use `"react-jsxdev"`.  Once fixed it should be transparent and just require updating/uncommenting a few test cases.  I am not sure if we would consider it a blocker or not for this feature, but I do think the PR is in a state to be reviewed.

cli/tools/test.rs
```diff
@@ -1054,7 +1054,7 @@ pub async fn run_tests_with_watch(
     let paths_to_watch_clone = paths_to_watch.clone();
 
     let maybe_resolver =
-      ps.maybe_import_map.as_ref().map(ImportMapResolver::new);
+      ps.maybe_import_map.clone().map(ImportMapResolver::new);
```
dsherret, line 1057: It seems we have many different ways of getting a resolver... why aren't they all the same?

cli/proc_state.rs
```diff
@@ -299,14 +300,26 @@ impl ProcState {
...
+      .as_ref()
+      .map(|cf| {
+        cf.to_maybe_jsx_import_source_module()
+          .map(|im| JsxResolver::new(im, maybe_import_map_resolver.clone()))
+      })
+      .flatten();
     let maybe_resolver = if self.flags.compat {
       Some(node_resolver.as_resolver())
+    } else if maybe_jsx_resolver.is_some() {
+      maybe_jsx_resolver.as_ref().map(|jr| jr.as_resolver())
```
dsherret, line 318: Reading this code makes me feel like this can't be used with import maps, but that's not the case because the JSX resolver takes an import map resolver and handles all that. Maybe we should add a comment here that says that. I also wonder if we should extract out the code that gets the resolver to some other common place for reuse.

dsherret (approved):
LGTM, but I have a few questions.

cli/resolver.rs
```diff
@@ -33,3 +35,42 @@ impl Resolver for ImportMapResolver<'_> {
       .map_err(|err| err.into())
   }
 }
+
+#[derive(Debug, Default, Clone)]
+pub(crate) struct JsxResolver {
+  jsx_import_source_module: String,
+  maybe_import_map_resolver: Option<ImportMapResolver>,
```
dsherret, line 42: Just a thought and maybe not for this review, but I wonder if we should have some other `AggregateResolver` or something that takes a collection of `dyn Resolver`s and then tries to resolve with each resolver in order. That way in the consuming code someone could build one up and it might be a bit more clear about what's going on.

lucacasonato:
I have tried this out and have found a few little things:

The following case does not work at all with TSC (the emit is broken), but works with SWC:

```tsx
/** @jsxImportSource https://esm.sh/preact */

function App() {
  return (
    // @ts-ignore
    <div>
      {/* @ts-ignore */}
      <h1>Hello, world!</h1>
      {/* @ts-ignore */}
    </div>
  );
}

<App />;
```

```shell
$ deno run check-no-import.tsx 
Check file:///mnt/starship/Projects/local/testnewjsx/check-no-import.tsx
error: Uncaught ReferenceError: _jsx is not defined
<App />;
^
    at file:///mnt/starship/Projects/local/testnewjsx/check-no-import.tsx:14:1
$ deno run --no-check no-check-no-import.tsx
$
```

---

Secondly, importing from skypack.dev in `@jsxImportSource` is not really possible, because you always need to add ?dts to the end of the URL. That can be resolved with an import map though, so no big deal. I will bring this up with the skypack folks - maybe `?dts` can just become the default.

---

```js
/** @jsxImportSource https://esm.sh/preact */
/** @jsx h */
/** @jsxFragment Fragment */
```

The above code works fine (it uses the `@jsxImportSource`) as expected. It is not obvious to users what should happen here though, so we should add a `deno lint` to disallow this. One or the other, not both.

---

We need to add the `jsxImportSource` option to the config.json schema in the VS Code extension.

---

The following does not work. It is not obvious to me why not. I think it should be working.

```tsx
import { renderToString } from "https://esm.sh/preact-render-to-string";

function App() {
  return (
    <div>
      <h1>Hello, world!</h1>
    </div>
  );
}

console.log(renderToString(<App />));
```

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "https://esm.sh/preact",
    "lib": ["esnext", "dom"]
  }
}
```

```shell
$ deno run --config deno.json check.tsx
Check file:///mnt/starship/Projects/local/testnewjsx/check.tsx
error: TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
    <div>
    ~~~~~
    at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:7:5

TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
      <h1>Hello, world!</h1>
      ~~~~
    at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:8:7

TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
      <h1>Hello, world!</h1>
                       ~~~~~
    at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:8:24

TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
    </div>
    ~~~~~~
    at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:9:5

Found 4 errors.
```

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -277,15 +277,20 @@ declare namespace Deno {
...
+    jsx?: "react" | "preserve" | "react-native" | "react-jsx" | "react-jsx-dev";
     /** Specify the JSX factory function to use when targeting react JSX emit,
      * e.g. `React.createElement` or `h`. Defaults to `React.createElement`. */
     jsxFactory?: string;
     /** Specify the JSX fragment factory function to use when targeting react
      * JSX emit, e.g. `Fragment`. Defaults to `React.Fragment`. */
     jsxFragmentFactory?: string;
+    /** Declares the module specifier to be used for importing the `jsx` and
+     * `jsxs` factory functions when using jsx as `"react-jsx"` or
+     * `"react-jsxdev"`. Defaults to `"https://esm.sh/preact"`. */
```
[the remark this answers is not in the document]
kitsonk: Sorry, need to update that.

lucacasonato:
The following breaks the LSP (the `preact` in the comment says `Resolved Dependency\nCode: [errored]` & does not work with `deno run --import-map import_map.json check.tsx`.

```tsx
/** @jsxImportSource preact */

import { renderToString } from "https://esm.sh/preact-render-to-string";

function App() {
  return (
    <div>
      <h1>Hello, world!</h1>
    </div>
  );
}

console.log(renderToString(<App />));
```

```json
{
  "imports": {
    "preact": "https://esm.sh/preact"
  }
}
```

Changing the import map to this works:

```json
{
  "imports": {
    "preact/jsx-runtime": "https://esm.sh/preact/jsx-runtime"
  }
}
```

lucacasonato:
And found a little LSP bug: the `preact` in `// @jsxImportSource preact` is highlighted, this pragma is only supported in a multi line comment (`/** @jsxImportSource preact */`).

cli/resolver.rs
kitsonk: See: #12684

kitsonk:
The "bad" emit is a TypeScript issue, opened https://github.com/microsoft/TypeScript/issues/46723.

kitsonk:
> Secondly, importing from skypack.dev in `@jsxImportSource` is not really possible, because you always need to add ?dts to the end of the URL. That can be resolved with an import map though, so no big deal. I will bring this up with the skypack folks - maybe `?dts` can just become the default.

Yes, I think we have no choice there, as both swc and tsc simply append `/jsx-runtime` to whatever specifier that is there and we have no control over that.

kitsonk:
> The following does not work. It is not obvious to me why not. I think it should be working.
> 
> ```tsx
> import { renderToString } from "https://esm.sh/preact-render-to-string";
> 
> function App() {
>   return (
>     <div>
>       <h1>Hello, world!</h1>
>     </div>
>   );
> }
> 
> console.log(renderToString(<App />));
> ```
> 
> ```json
> {
>   "compilerOptions": {
>     "jsx": "react-jsx",
>     "jsxImportSource": "https://esm.sh/preact",
>     "lib": ["esnext", "dom"]
>   }
> }
> ```
> 
> ```shell
> $ deno run --config deno.json check.tsx
> Check file:///mnt/starship/Projects/local/testnewjsx/check.tsx
> error: TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
>     <div>
>     ~~~~~
>     at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:7:5
> 
> TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
>       <h1>Hello, world!</h1>
>       ~~~~
>     at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:8:7
> 
> TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
>       <h1>Hello, world!</h1>
>                        ~~~~~
>     at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:8:24
> 
> TS7026 [ERROR]: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
>     </div>
>     ~~~~~~
>     at file:///mnt/starship/Projects/local/testnewjsx/check.tsx:9:5
> 
> Found 4 errors.
> ```

Ok, this is because when importing the `https://esm.sh/preact/jsx-runtime` from the "config file", we aren't importing the types (`X-TypeScript-Type`), but when we are importing via `@jsxImportSource` is "just works".  I think it is because when we send it over to tsc, we aren't using the `maybe_types_dependency` specifier, like we would with other "imports" into a module. Hrmmmfp.

cli/config_file.rs
```diff
@@ -407,12 +412,53 @@ impl ConfigFile {
...
+    if let Some(types) = compiler_options.types {
+      imports.extend(types);
+    }
+    if compiler_options.jsx == Some("react-jsx".to_string()) {
+      imports.push(format!(
+        "{}/jsx-runtime",
+        compiler_options
+          .jsx_import_source
+          .unwrap_or_else(|| {
+            log::warn!("{}: Compiler option 'jsx' set to 'react-jsx', but no 'jsxImportSource' defined, defaulting to 'react'.", colors::yellow("warning"));
```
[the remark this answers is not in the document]
kitsonk: Ok, it errors now.

cli/proc_state.rs
kitsonk: For the latter part, feels like something we should address in #12684.

cli/tools/test.rs
kitsonk: I think this is also something we should fully address in #12684.  I will fix this one, but the problem is that we don't have the compat resolver everywhere at the moment, and I shouldn't address that as part of this PR.  Which means there is an inconsistency across the code base.  I cleaned it up a bit in the LSP (but the LSP doesn't support the compat resolver yet)

kitsonk:
> Ok, this is because when importing the `https://esm.sh/preact/jsx-runtime` from the "config file", we aren't importing the types (`X-TypeScript-Type`), but when we are importing via `@jsxImportSource` is "just works". I think it is because when we send it over to tsc, we aren't using the `maybe_types_dependency` specifier, like we would with other "imports" into a module. Hrmmmfp.

Ooops. I was wrong... the problem is that resolution doesn't find it, because tsc tries to resolve the runtime module from each .jsx module, realising that it _might_ be different depending on the referrer, and we think it is missing, because it is a "global" dependency, not a specific dependency. Hmmmm...

kitsonk:
Ok, I added https://github.com/denoland/deno_graph/pull/67 which allows us to "find" the jsxImportSource in the module graph.  Just need to change the `op_resolve` in the lsp and cli to look in the right places.  It is a blocker for this, as type checking won't work in a lot of real world use cases with just using compiler options instead of the JSX pragma.

kitsonk:
> ```json
> {
>   "imports": {
>     "preact": "https://esm.sh/preact"
>   }
> }
> ```

@lucacasonato [Shouldn't this be](https://github.com/WICG/import-maps#packages-via-trailing-slashes):

```json
{
  "imports": {
    "preact/": "https://esm.sh/preact/"
  }
}
```

lucacasonato:
> > ```json
> > {
> >   "imports": {
> >     "preact": "https://esm.sh/preact"
> >   }
> > }
> > ```
> 
> @lucacasonato [Shouldn't this be](https://github.com/WICG/import-maps#packages-via-trailing-slashes):
> 
> ```json
> {
>   "imports": {
>     "preact/": "https://esm.sh/preact/"
>   }
> }
> ```

Oh right... my bad... sorry!

kitsonk:
Ok, I have fixed everything that is addressable at the moment, and I think it is good enough to land, but there are two caveats and I will reflect them in the manual:

- TSX/JSX modules with no import/exports but using JSX Import Source do not emit properly and will runtime error.  Work around is to add `export {}` to the module or to use `--no-check`.  Real world JSX/TSX modules without any imports or exports are likely rare in the wild.
- `--no-check`/bundling (using swc as the emitter) does not properly support the compiler option `"react-jsxdev"`.  The work-around is to use `"react-jsx"` or to not use `--no-check` or bundling.  Using `"react-jsxdev"` has limited value in Deno anyways.

Both of these are caused by upstream issues, which are open and being tracked.

lucacasonato (approved):
Let's ship it!

CanRau:
> Using \"react-jsxdev\" has limited value in Deno anyways.

Curious why that has limited value?

kitsonk:
> > Using "react-jsxdev" has limited value in Deno anyways.
> 
> Curious why that has limited value?

Because most JSX runtimes, except React, don't actually use it, and in most cases, people use JSX for SSR, and the instrumentation for the dev mode is focused on client side debugging, with all the useful information being erased when SSR occurs.

CanRau:
> > > Using "react-jsxdev" has limited value in Deno anyways.
> > 
> > Curious why that has limited value?
> 
> Because most JSX runtimes, except React, don't actually use it, and in most cases, people use JSX for SSR, and the instrumentation for the dev mode is focused on client side debugging, with all the useful information being erased when SSR occurs.

Uuh I see, makes sense, thanks for elaborating 🙏

pheuter:
I'm seeing an issue with this when trying to do `deno bundle` with `react/jsx-runtime`, create repro here with description of error: https://github.com/pheuter/hello-deno#blocking-issue.

Currently unable to bundle client-side code using `react/jsx-runtime` despite `deno run` working just fine using the same code and config:

```
» deno bundle --config deno.json client/index.tsx static/client.js
Check file:///Users/mark/src/pheuter/hello-deno/client/index.tsx
Bundle file:///Users/mark/src/pheuter/hello-deno/client/index.tsx
error: Unable to output during bundling.

Caused by:
    0: load_transformed failed
    1: failed to analyze module
    2: failed to resolve https://esm.sh/react@17.0.2/jsx-runtime from file:///Users/mark/src/pheuter/hello-deno/client/index.tsx
    3: Cannot resolve "https://esm.sh/react@17.0.2/jsx-runtime" from "file:///Users/mark/src/pheuter/hello-deno/client/index.tsx".
```

```sh
» deno --version
deno 1.17.3 (release, aarch64-apple-darwin)
v8 9.7.106.15
typescript 4.5.2
```

Oddly enough, the issue appears to go away when I explicitly do `/** @jsxImportSource https://esm.sh/react@17.0.2 */` in every tsx file, which I'm trying to avoid with this optimization in the first place.

kitsonk:
Please open an issue and not comment on a PR that has been closed for over 2 months unrelated to the issue you are experiencing.

pheuter:
No problem, thought I saw some potentially relevant comments above, wasn't sure the best way to proceed. Went ahead and opened https://github.com/denoland/deno/issues/13389.

Issue #12577, opened 2021-10-28
Improve Node Compat Mode


ry:
Make various popular projects run: 
- [x] Vite frontend development tool @bartlomieju 
  - ~blocked on http client work~
  - blocked on http hijacking for web sockets https://github.com/denoland/deno/issues/12718
- [x] Mongo DB Driver @AaronO 
  - ~blocked on tls https://github.com/denoland/deno_std/issues/1475~
  - blocked on zlib https://github.com/denoland/deno_std/issues/1541
- [ ] Mocha test framework @kt3k https://github.com/denoland/deno/issues/12886
  - ~should more or less work - maybe broke on timeout (https://github.com/denoland/deno_std/pull/1699)~
- [ ] Next.js @bartlomieju 
  - blocked on `child_process.send` (https://github.com/denoland/deno/issues/12879)
  - blocked unhandledrejection https://github.com/denoland/deno_std/issues/1744
  - investigate "middleware" target
- [x] mysql2 https://github.com/denoland/deno/issues/13774
  - ~blocked on tls~
  - needs more investigation
- [x] npm  https://github.com/denoland/deno/issues/13187
- [x] eslint
  - blocked unhandledrejection https://github.com/denoland/deno_std/issues/1744
- [ ] Remix https://github.com/denoland/deno_std/issues/2010

Features:
- [ ] LSP support
- [ ] Type checking NPM modules
- [x] REPL https://github.com/denoland/deno/issues/12628
- [x] https://github.com/denoland/deno/issues/12648
- [x] https://github.com/denoland/deno/issues/12872
- [x] https://github.com/denoland/deno/issues/12879
- [x] https://github.com/denoland/deno/issues/12482
- [x] https://github.com/denoland/deno/issues/12495
- [x] https://github.com/denoland/deno/issues/13227
- [x] https://github.com/denoland/deno/issues/13200

bartlomieju:
ESLint is now working in compat mode:

![Screenshot_2021-10-31_at_12 24 51](https://user-images.githubusercontent.com/13602871/139581208-d62cc5ef-efb3-48b2-8e99-c9fab540851e.png)

![Screenshot_2021-10-31_at_12 37 30](https://user-images.githubusercontent.com/13602871/139581215-6285153c-cc2d-4318-aee8-1d8d2ac34000.png)

bartlomieju:
Parcel uses "native addons" (`.node` files), so I'm not sure we'll be able to run that at all.

[...]

PR #12633, opened 2021-11-03
feat(compat): add .code to dyn import error

### LINKED ISSUE #12588: [compat] when using dynamic import in Node compat mode and module is not found, err.code needs to be set to 'ERR_MODULE_NOT_FOUND'
In Node.js, it's common to check err.code to detect the kind of errors:

```ts
if (err.code === 'ERR_MODULE_NOT_FOUND') {
  // do something
}
```

`std/node` simulates these error codes in API layer (ref: https://github.com/denoland/deno_std/blob/0b31c1647649f9b6aa53432370daab4483803e77/node/_errors.ts#L289-L304 )

But in the case of dynamic import, the above shim can't be used. So `err.code` needs to be set to `ERR_MODULE_NOT_FOUND` directly in CLI runtime when `--compat` flag is specified.

This feature is used in `mocha` when looking up the test files (ref: https://github.com/mochajs/mocha/blob/97b84708afb42e552cb906a54f9f2aa2e6a98ba4/lib/nodejs/esm-utils.js#L47-L56 )

related: #12577

kt3k:
part of #12577, closes #12588

This PR changes the error class for dynamic import rejection in compat mode. Some tools (e.g. mocha) uses .code property of error object to detect the resolution failure of dynamic import (ref: https://github.com/mochajs/mocha/blob/97b84708afb42e552cb906a54f9f2aa2e6a98ba4/lib/nodejs/esm-utils.js#L47-L56 ). This change simulates the above behavior in compat mode.

This change enables compat mode to run the basic feature of mocha cli


kt3k:
`cli/tests/testdata/fix_dynamic_import_errors.js` test case strangely panicks at the 2nd dynamic import..

kt3k:
Now ready for review

I first tried to reuse class_name of CustomError for accessing the custom js error constructor, but that didn't work because some errors cause panic during getting error class (caused by `cli/tests/testdata/fix_dynamic_import_errors.js`). Now I created a dedicated error class for that purpose `CustomErrorWithJsConstructor` and retrieves js constructor name from that class during rejections of dynamic imports

bartlomieju:
Please discuss this change in the design letting before landing.

lucacasonato (commented):
I am not in favor of this `Deno.compat.errors` namespace. In node `await import("xxxxxx")` throws a regular `Error` (no subclass), with a `.code` on it. In the browser it throws a `TypeError`. I think we should always throw a `TypeError` (also in non-compat mode), and in compat mode add a `code` property to that.

lucacasonato (approved):
LGTM

PR #12636, opened 2021-11-03
fix(lsp): display signature docs as markdown

### LINKED ISSUE #12620: LSP: SignatureHelp response contains Markdown but isn't tagged as such, causing inadequate client rendering
Hi there. Here's a screenshot from Sublime Text showing a fenced code block in LSP SignatureHelp being considered verbatim plaintext:

Screenshot 2021-11-01 at 12 18 39

[A Sublime LSP developer says this is a server rather than client issue](https://github.com/sublimelsp/LSP-Deno/issues/9#issuecomment-956280474), because the content is not being tagged as `MarkupContent` in the LSP SignatureHelp response.

From a quick search of the code in this repo, it looks like there are two existing examples of content being tagged as `MarkupContent`:
* [Hover](https://github.com/denoland/deno/blob/34a9ddff091950aee9d89915cd13944259e7d346/cli/lsp/language_server.rs#L1103-L1106)
* [CompletionEntryDetails](https://github.com/denoland/deno/blob/34a9ddff091950aee9d89915cd13944259e7d346/cli/lsp/tsc.rs#L1537-L1540)

lucacasonato:
These were previously displayed as plain text. Now they are displayed as
`MarkupContent` with type `Markdown`.

Fixes #12620.

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -1953,11 +1953,15 @@ impl SignatureHelpItem {
       .collect::<Vec<String>>()
       .join(", ");
     let suffix_text = display_parts_to_string(&self.suffix_display_parts);
+    let documentation = display_parts_to_string(&self.documentation);
     lsp::SignatureInformation {
       label: format!("{}{}{}", prefix_text, params_text, suffix_text),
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM

PR #12642, opened 2021-11-03
chore(wpt): Switch a WPT test's location based on its filename flags

### LINKED ISSUE #12472: The location used for a WPT test should depend on its filename flags
When browsers run WPT tests, some of the [filename flags](https://web-platform-tests.org/writing-tests/file-names.html) are used to determine which hostname and port to use for a specific test. Deno's WPT runner, however, uses http://web-platform.test:8000 regardless of the flags. This usually works fine, since the `.https` flag is the most common, and is mostly used to test [secure context](https://w3c.github.io/webappsec-secure-contexts/) APIs, which Deno doesn't care about.

However, there are uses of the `.h2` flag in the fetch tests that need the location to be set to [port 9000](https://web-platform-tests.org/writing-tests/server-features.html?highlight=ports#tests-requiring-http-2-0), since the tests fetch relative URLs and expect them to be fetched with HTTP/2. In fact, tests that test HTTP/2 _must_ use the `.h2` flag, since in browsers anything else would be cross-origin – and so it makes sense that those tests use relative URLs.

Changing the location might not be needed for flags other than `.h2`, but there might be other unexpected dependencies, and it won't hurt either way.

andreubotella:
Closes #12472.


andreubotella:
I thought `fetch/api/basic/request-upload.any.html` wouldn't be flaky anymore, that's my bad.

bartlomieju:
@andreubotella can we land this PR? Is it still causing flakiness?

andreubotella:
Blocked on #13058.

Issue #11970, opened 2021-09-10
Don't report diagnostics for "external" TypeScript code


kitsonk:
When type checking a project, by default don't report type diagnostics for "external" files.  (raised as 2.0 scope by @lucacasonato)

Things we need to do to make this proposal more viable would be:

- Define what "external" files are.  Are they any modules outside the cwd/root of a project? Are they only remote modules?
- How would this integrate well with #11340 without adding a lot of complexity?

wojpawlik:
If reporting diagnostics only for _local_ TypeScript files would speed up diagnostics, it'd be an interesting _alternative_ to https://github.com/denoland/deno/issues/11340.
Also, I'd have prevented https://github.com/colinhacks/zod/pull/648.

PR #12643, opened 2021-11-03
Revert "feat(cli): enable `useUnknownInCatchVariables` by default"

lucacasonato:
Over the last few days I have been using this change in canary, and a significant amount of the code I run on a day to day basis broke type checking, and is non trivial to fix. This change is going to cause a lot of pain across the ecosystem, because std is often nested deep in other dependencies (and it also affects many dependencies outside of std, essentially anyone using `try catch`).

I think we should revert this change, and only re-introduce it, and other changes like it until we figure out a solution for this "type errors in dependencies" problem (e.g. #11970).

Reverts denoland/deno#12547

### DIFF
--- a/cli/emit.rs
+++ b/cli/emit.rs
@@ -160,6 +160,8 @@ pub(crate) fn get_ts_config(
         "target": "esnext",
         "tsBuildInfoFile": "deno:///.tsbuildinfo",
         "useDefineForClassFields": true,
+        // TODO(@kitsonk) remove for Deno 2.0
+        "useUnknownInCatchVariables": false,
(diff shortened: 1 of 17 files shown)


kitsonk (changes requested):
We should revert the change but not the improved code that guards against errors better.

lucacasonato:
@kitsonk That sounds very reasonable. I'll prepare a patch tomorrow morning (unless you want to do it).

kitsonk:
I will try to get to it today.

kitsonk (approved):
LGTM

PR #12644, opened 2021-11-04
fix(cli/lint): use recommended tag if there is no tags in config file or flags

### LINKED ISSUE #12615: bug(cli/lint): all rules are ignored if the config is changed
When the default config is changed through CLI flags of the `deno.jsonc` config file, all the lint rules are ignored

deno version: `deno 1.15.3 (release, x86_64-unknown-linux-gnu)`

## Example

```ts
// example.ts
function getString(): any {
  return "hi";
}
```

```jsonc
// deno.jsonc
{
  "lint": {}
}
```

#### default

```sh
deno lint example.ts
```
image

#### config
```sh
deno lint --config deno.jsonc example.ts
```
image

#### flags
```sh
deno lint --rules-exclude=no-explicit-any example.ts
```
image

ah-yu:
`rules-tags` defaults to `["recommended"]`, not empty list

Close #12615


dsherret (approved):
LGTM. Thanks @ah-yu!

Actually just yesterday I added a config file in a repo with only `"files": { "exclude": [...] }` and was super confused about why there were no lint errors even though it was checking the files.

PR #12651, opened 2021-11-04
fix(ext/web/streams): typings: make getReader return ReadableStreamBYOBReader & stricter byob typings 

crowlKats:


lucacasonato (approved):

lucacasonato:
Thanks!

PR #12652, opened 2021-11-04
chore(ext/web): remove `ReadableStream.getIterator`

crowlKats:
Has a TODO for all the way back to 1.8. Also, std doesnt use getIterator anywhere anymore, so it should be safe to remove


lucacasonato (approved):
Thank you!

PR #12659, opened 2021-11-05
fix(cli): don't panic when mapping unknown errors

### LINKED ISSUE #12590: thread 'main' panicked at 'Error 'IO error: Broken pipe
`thread 'main' panicked at 'Error 'IO error: Broken pipe (os error 32)' contains boxed error of unknown type:
  Io(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
  Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }', cli/errors.rs:73:7
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace`

This happened, on Amazon Linux 2 instance, Deno 1.15.3, behind an AWS loadbalancer. My code base is rather large, and the stack base rather unhelpful in giving directions as to the origins of the error, so unfortunately I have zero idea how to reproduce this, or what this might be related to. It was suggested on discord to create an issue anyway, so here you go.

Some additional info, related or not, I'm using x/oak, mongo-driver (to a db in another aws instance), Deno.run(), x/S3, and websockets. It happened during users logging in to the app, so a lot of stuff happening, and it could be related to any of those, or something else completely.

AaronO:
Instead fallback to generic "Error" class, fixes #12590


ry (approved):
LGTM

PR #12661, opened 2021-11-07
fix(lsp): cache unsupported import completion origins

### LINKED ISSUE #12621: LSP: don't add registries to `deno.suggest.imports.hosts` that don't have a manifest
Just typing out an import slowly results in a bunch of hosts added to `.vscode/settings.json`:

```json
{
  "deno.enable": true,
  "deno.lint": true,
  "deno.unstable": false,
  "deno.suggest.imports.hosts": {
    "https://deno.land": true,
    "https://x.nest.land": true,
    "https://crux.land": true,
    "https://x.lcas.dev": true,
    "https://cdn.skypack.dev": false,
    "https://esm.sh": false,
    "https://skypack.": false,
    "https://cdn..sh": false,
    "https://e": false,
    "https://esm": false,
    "https://esm,.": false,
    "https://esm.": false,
    "https://esm.s": false
  }
}
```

If an origin does not provide a manifest, we should cache the non existence of a manifest in the registry cache, and not add anything to the settings file (this is what the old extension used to do).

kitsonk:
Fixes #12621

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -232,7 +232,7 @@ pub struct FileFetcher {
   allow_remote: bool,
   cache: FileCache,
   cache_setting: CacheSetting,
-  http_cache: HttpCache,
+  pub(crate) http_cache: HttpCache,
   http_client: reqwest::Client,
   blob_store: BlobStore,
 }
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM

PR #12663, opened 2021-11-07
chore(lsp): align ScriptElementKind to current versions of TypeScript/vscode

kitsonk:
There has been a bit of "drift" in what TypeScript supports and what vscode does with information included in the navigation tree.


lucacasonato (approved):
RSLGTM

PR #12699, opened 2021-11-08
fix(ext/net): expose all tls ops

lucacasonato:
This makes it posssible for implementors to cherry-pick which ops they
want to use.


piscisaureus (approved):
LGTM

PR #12704, opened 2021-11-09
Revert "refactor(ext/http): rewrite hyper integration and fix bug (#12332)"

lucacasonato:
This reverts commit 5b1e537446454f6332de44adbeb6a15ff072c2fa.

This change updates some behaviour that caused std/http to break. We will try to reland this for 1.16.1.

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -7,7 +7,6 @@ import {
   assert,
   assertEquals,
   assertRejects,
-  assertStrictEquals,
   assertThrows,
   deferred,
   delay,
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
RSLGTM

piscisaureus:
Any background on exactly how this was breaking std/http?

lucacasonato:
Yes, the `serve should handle websocket requests` test in std/http/server_test.ts fails. Root cause seemed to be that previously `Deno.close`ing a `Deno.HttpConn` would not close websockets created by `Deno.upgradeWebsocket`, but now it does.

piscisaureus:
> Yes, the `serve should handle websocket requests` test in std/http/server_test.ts fails.

Thanks for the pointer!

> Root cause seemed to be that previously `Deno.close`ing a `Deno.HttpConn` would not close websockets created by `Deno.upgradeWebsocket`, but now it does.

That's not the root cause; closing HttpConn doesn't close websockets before nor after the patch.
It seems that this test is hanging because the `TcpListener` created by `serve()` stays open somehow.

lucacasonato:
@piscisaureus If you add some onerror handlers to the socket, you will see that the WebSocket never gets established.

Issue #12641, opened 2021-11-03
 Make internal modules optional for workers in deno_runtime


JomerDev:
While using deno_runtime embedded in a project, not everybody has a need for all internal modules (deno_webgpu for example).
I propose to add features for the individual modules, with the default being whatever deno uses. That would allow for more flexibility for embedders while meaning no change to the deno cli

P.s. I'm happy to open a PR for this myself, but I was told in discord to open an issue for this first

lucacasonato:
Feel free to open a PR. We can discuss further there.

lucacasonato:
The PR did not go anywhere - closing this

PR #12710, opened 2021-11-09
feat(runtime): Make internal modules optional for workers in deno_runtime

JomerDev:
As described in #12641, I've started working on a PR to make some of the dependencies optional. 
The default still uses those optional modules, so nothing changes for deno.
Currently this breaks the build without features because 90_deno_ns.js and 99_main.js try and use the types provided by those modules. I've yet to find a good way to apply the features with the js files.
I guess a better way would be to move the js that needs those modules into separate files inside the modules and load them on module init.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -12,6 +12,7 @@ description = "Provides the deno runtime library"
 [features]
 # "fake" feature that allows to generate docs on docs.rs
 docsrs = []
+default=[ "deno_webgpu", "deno_ffi" ]
 
 [lib]
 name = "deno_runtime"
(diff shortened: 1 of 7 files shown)


bartlomieju:
@JomerDev what exactly breaks in 90_deno_ns.js and 99_main.js?

JomerDev:
@bartlomieju Currently the deno namespace is build in 90_deno_ns.js. When modules are disabled, it still tries to add functions from those modules into the namespace. 
For example, with deno_ffi disabled, `__bootstrap.ffi.dlopen` doesn't exist, which causes an error when it is added to the unstable namespace

bartlomieju:
> @bartlomieju Currently the deno namespace is build in 90_deno_ns.js. When modules are disabled, it still tries to add functions from those modules into the namespace. For example, with deno_ffi disabled, `__bootstrap.ffi.dlopen` doesn't exist, which causes an error when it is added to the unstable namespace

I see, that would require some more setup configuration being forwarded to JS upon snapshotting and bootstrapping. It might be quite involved to solve this issue

JomerDev:
That's what I found as well, which is partly the reason work on this kind of stopped (although I should have updated the PR to reflect that). As I remember, for some other internal modules I didn't touch so far changes to the permission system could be needed as well, which didn't seem feasible.

While I would still like to make the modules more independent of each other, it seems like there is a lot of work to be done before that becomes possible

JomerDev:
Since there currently is no good way to move forward with this PR I'll close it for now

jasl:
I'm working on evaluate Deno as my hobby project (a domain-specific scripting engine), so it's good to see Deno runtime is cusomizable, I found the PR from Google

I have an idea and not sure it's possible,

We can use feature gate for Rust side as the PR already done, for JS side, I think we can use template engine to construct JS code in `build.rs`, then we have chance to inject what features we wanna enable and generate JS code for them.

PR #12720, opened 2021-11-09
refactor(lsp): prefer using document instead of documents collection

dsherret:
This change makes it so that the documents collection returns a document which can then be inspected instead of the code using the documents collection having methods for getting information about each property on a document.

This will allow us to use a document in places where previously we would have to pass the entire documents collection and we can move these documents into other kinds of collections and still have this functionality rather than everything being tied to this specific documents collection API/implementation.

Although it doesn't matter at the moment because we only allow one message through the LSP at a time, this change also slightly reduces the raciness of the LSP code because if someone gets a document then they won't have that document change while they're inspecting its properties. In the future if we try to eliminate that restriction then it will matter so this is just progressively going in that direction.


cli/lsp/documents.rs
```diff
@@ -186,21 +243,25 @@ impl SyntheticModule {
...
+struct DocumentInner {
   line_index: Arc<LineIndex>,
   maybe_language_id: Option<LanguageId>,
   maybe_lsp_version: Option<i32>,
   maybe_module:
     Option<Result<deno_graph::Module, deno_graph::ModuleGraphError>>,
   maybe_navigation_tree: Option<Arc<tsc::NavigationTree>>,
   maybe_warning: Option<String>,
-  source: SourceTextInfo,
+  text_info: SourceTextInfo,
```
kitsonk, line 255: nit: alphabetise

cli/lsp/documents.rs
```diff
@@ -186,21 +243,25 @@ impl SyntheticModule {
...
   specifier: ModuleSpecifier,
   version: String,
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct Document {
+  inner: Arc<DocumentInner>,
+}
```
kitsonk, line 263:
```suggestion
pub(crate) struct Document(Arc<DocumentInner>);
```

I know it is less "explicit" but it is all just passthrough anyways.

cli/lsp/documents.rs
```diff
@@ -261,49 +361,113 @@ impl Document {
...
+        maybe_lsp_version: Some(version),
+        maybe_navigation_tree: None,
+      }),
+    })
   }
 
-  fn close(&mut self) {
-    self.maybe_lsp_version = None;
-    self.maybe_language_id = None;
+  fn with_closed(&self) -> Document {
```
kitsonk, line 409: I wasn't so concerned about the `with_change`, but this seems totally arbitrary and "wasteful".  We care cloning everything to set two properties to `None`.

Other than a bit of readability, I am not sure there really is a compelling case here.  I think we need to get some wider opinions about this, like from @lucacasonato.

cli/lsp/documents.rs
```diff
@@ -261,49 +361,113 @@ impl Document {
...
+        line_index: self.inner.line_index.clone(),
+        maybe_lsp_version: None,
+        maybe_language_id: None,
+      }),
+    }
   }
 
-  fn content(&self) -> Arc<String> {
-    self.source.text()
+  fn with_navigation_tree(
```
kitsonk, line 425: Yeah, again, this seems arbitrarily complex and "wasteful".

kitsonk (commented):
Like discussed, I am not specifically opposed to it, but I am not sure I totally "like" it. It would be good to get some other opinions I think.

cli/lsp/documents.rs
dsherret: It's because the document structs are now immutable and cheaply cloned. This will allow us to get "snapshots" of the LSP state for actual multiple messages in the future. Right now the "`StateSnapshot`" is a lie (see how it holds documents, which is an `Arc<Mutex<_>>`). It will also allow us to use a document in places where previously we would have to pass the entire documents collection and we can move these documents into other kinds of collections and still have this functionality rather than it being tied to this specific documents collection API.

I've made the statement look a little less verbose by using a spread operator and clone the inner.

I don't believe this is wasteful. There isn't anything in there that seems heavy to clone. Could you elaborate?

cli/lsp/documents.rs
dsherret: Done.

ry (commented):
I'll stay out of the review, but I want to note that this is less code, and that David's explanation makes sense to me.

For the detractors, please remember that Arc::clone is **very cheap**.

bartlomieju:
I'm in favor of this change, David's rationale is very well explained and the fewer `Arc<Mutex<>>` we pass around, the better, as we limit the potential for the deadlocks. Code in `cli/lsp/language_server.rs` looks less complicated too. Unless there's a noticeable performance regression I believe we should go with this approach.

cli/lsp/documents.rs
lucacasonato: The "heaviest" clone I could find in here is the `deno_graph::Module` in the `maybe_module` slot. Cloning that will clone the entire source text of the module. In the grand scope of things, I think the memory pressure vs. immutability tradeoff is worth it here, because the immutable documents will make it easier to implement concurrent LSP calls (and thus cancellations!)

bartlomieju: @lucacasonato this is not a big deal, because source code of the module is already stored in an `Arc` https://docs.rs/deno_graph/0.11.1/deno_graph/struct.Module.html

kitsonk:
> For the detractors, please remember that Arc::clone is **very cheap**.

"detractor"... 😒 

Thinking about this a bit further, would it make sense to move the `NavTree` and the lsp version out of the document themselves?  That way, document does really become "immutable".  The lsp version really is just owned by the LSP and can simply be stored via the module specifier.  NavTree, which we generate sometimes for assets as well, could be stored as well just by module specifier.

cli/lsp/documents.rs
kitsonk: I mentioned below, I think if we move NavTree out of document and the lsp version, the document truly becomes immutable, and we wouldn't clone on any other modification.

dsherret: @kitsonk I think we should still store that in the `Document` because moving it out would mean consulting some other source of truth which might not align with `Document`. Maybe we should move all the properties that won't change into another sub struct and throw that in an `Arc`? I'm not sure it's worth it. Eventually we'll be able to remove the navigation tree.

I think with this approach we'll eventually be able to create real snapshots of the documents at the start of an LSP request and not need to check any data behind a lock that could be mutated by another thread (side note: I think I have an idea for how to handle closed files without running into race conditions... need to explore it a bit).

cli/lsp/code_lens.rs
```diff
@@ -220,9 +220,10 @@ async fn resolve_implementation_code_lens(
   data: CodeLensData,
   language_server: &mut language_server::Inner,
 ) -> Result<lsp::CodeLens, AnyError> {
-  let line_index = language_server
-    .get_line_index_sync(&data.specifier)
+  let asset_or_doc = language_server
+    .get_cached_asset_or_document(&data.specifier)
     .unwrap();
+  let line_index = asset_or_doc.line_index();
```
kitsonk, line 226: Although this `.unwrap()` was here before, it feels like we should not risk a panic, and instead return a err if for whatever reason the document is missing.

cli/lsp/code_lens.rs
```diff
@@ -289,9 +290,10 @@ async fn resolve_references_code_lens(
   data: CodeLensData,
   language_server: &mut language_server::Inner,
 ) -> Result<lsp::CodeLens, AnyError> {
-  let line_index = language_server
-    .get_line_index_sync(&data.specifier)
+  let asset_or_document = language_server
+    .get_cached_asset_or_document(&data.specifier)
     .unwrap();
+  let line_index = asset_or_document.line_index();
```
kitsonk, line 296: Same as above, regarding not risking the panic.

cli/lsp/documents.rs
```diff
@@ -134,6 +135,62 @@ impl IndexValid {
...
+  pub fn line_index(&self) -> Arc<LineIndex> {
+    match self {
+      AssetOrDocument::Asset(a) => a.line_index.clone(),
+      AssetOrDocument::Document(d) => d.line_index(),
+    }
+  }
+
+  pub fn maybe_navigation_tree(&self) -> Option<Arc<tsc::NavigationTree>> {
+    match self {
+      AssetOrDocument::Asset(a) => a.maybe_navigation_tree.clone(),
```
kitsonk, line 168: nit: make asset consistent api consistent with document api?

cli/lsp/documents.rs
```diff
@@ -225,27 +284,62 @@ impl Document {
...
+    Self(Arc::new(DocumentInner {
+      fs_version,
       line_index,
       maybe_language_id: None,
       maybe_lsp_version: None,
       maybe_module,
       maybe_navigation_tree: None,
       maybe_warning,
-      source,
+      text_info: source,
```
kitsonk, line 295: nit: rename `source` variable to `text_info`

cli/lsp/documents.rs
```diff
@@ -261,49 +355,91 @@ impl Document {
...
-  fn close(&mut self) {
-    self.maybe_lsp_version = None;
-    self.maybe_language_id = None;
+  fn with_navigation_tree(
+    &self,
+    navigation_tree: Arc<tsc::NavigationTree>,
+  ) -> Document {
+    Document(Arc::new(DocumentInner {
+      maybe_navigation_tree: Some(navigation_tree),
+      ..(*self.0).clone()
```
kitsonk, line 412: 💯

kitsonk (approved):
LGTM, couple of nits

bartlomieju (approved):
LGTM too

Issue #12722, opened 2021-11-10
bug: v1.16.0 --unstable option not working


gseok:
- my code using ```const sig = Deno.signal('SIGINT');```
- and deno run code like this >> ```deno run --watch --allow-net --allow-read --unstable ./src/server.ts```
- I'm definitely using "--unstable" options.!!!

Try deno 1.16.0 (release, x86_64-apple-darwin) > failure running with follow output
```
error: TS2339 [ERROR]: Property 'signal' does not exist on type 'typeof Deno'. 'Deno.signal' is an unstable API. Did you forget to run with the '--unstable' flag?
  const sig = Deno.signal('SIGINT');
                   ~~~~~~
```

Try deno 1.15.3 > success running..!!!!

So... I think, this is a bug in 1.16.0. 
Please check this problem.

bartlomieju:
`Deno.signal()` was replaced with `Deno.addSignalListener()` API; more info here: https://deno.com/blog/v1.16#new-unstable-signal-listener-api

It is indeed a bug that you see this suggestion and I will fix it shortly.

authcompanion:
Still getting error in v1.16.2. Whats expected again? Ty for the info!

`TS2339 [ERROR]: Property 'signal' does not exist on type 'typeof Deno'.
    Deno.signal("SIGINT"),`

bartlomieju:
@authcompanion this was fixed and released in v1.16.1, make sure to upgrade to the latest version

authcompanion:
@bartlomieju odd I'm on the latest version and still get error:
`deno 1.16.2 (release, x86_64-apple-darwin)
v8 9.7.106.2
typescript 4.4.2`

Here's the snippet I'm using it in:
https://github.com/authcompanion/authcompanion/blob/a88626b1d6d08016987f612a51a92398e8c7a605/bin/server.ts#L26

error: TS2339 [ERROR]: Property 'signal' does not exist on type 'typeof Deno'.
    Deno.signal("SIGTERM"),
         ~~~~~~
    at file:////bin/server.ts:26:10

TS2339 [ERROR]: Property 'signal' does not exist on type 'typeof Deno'.
    Deno.signal("SIGINT"),
         ~~~~~~
    at file:///bin/server.ts:27:10

TS2339 [ERROR]: Property 'signal' does not exist on type 'typeof Deno'.
    Deno.signal("SIGABRT"),
         ~~~~~~
    at file:///bin/server.ts:28:10

Found 3 errors.

[...]

PR #12723, opened 2021-11-10
fix: update unstable Deno props for signal API

bartlomieju:
Closes https://github.com/denoland/deno/issues/12722

`cli/tests/testdata/raw_mode_cbreak.ts` was referencing `Deno.signal()` but wasn't used anywhere so I removed it as well.


ry (commented):
It's strange that the tests didn't catch this. Is this because all the signal tests are in JS instead of TS?

bartlomieju:
> It's strange that the tests didn't catch this. Is this because all the signal tests are in JS instead of TS?

All of our unit tests run with `--unstable` flag. And there were no more references to `Deno.signal` so there was nothing to catch.

ry (approved):
Ok - LGTM

PR #12730, opened 2021-11-10
fix: add typings for AbortSignal.reason

lucacasonato:
Adds missing types.

### DIFF
--- a/ext/web/lib.deno_web.d.ts
+++ b/ext/web/lib.deno_web.d.ts
@@ -271,7 +271,7 @@ interface AbortSignal extends EventTarget {
   /** Returns true if this AbortSignal's AbortController has signaled to abort,
    * and false otherwise. */
   readonly aborted: boolean;
-  readonly reason?: any;
+  readonly reason?: unknown;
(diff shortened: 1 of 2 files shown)


dsherret (approved):
LGTM

ext/web/lib.deno_web.d.ts
```diff
@@ -271,6 +271,7 @@ interface AbortSignal extends EventTarget {
   /** Returns true if this AbortSignal's AbortController has signaled to abort,
    * and false otherwise. */
   readonly aborted: boolean;
+  readonly reason?: any;
```
dsherret, line 274: Should we use `unknown`?

lucacasonato: `WritableStream.abort` signature is `abort(reason?: any): Promise<void>;`, so probably best to align with that.

dsherret: I think that makes sense for inputs, but not sure about outputs. It might be best to make this restrictive then we can loosen it later.

lucacasonato: Ok, I'll change to `unknown`.

PR #12732, opened 2021-11-10
reland #12332: refactor(ext/http): rewrite hyper integration and fix bug

piscisaureus:

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -7,6 +7,7 @@ import {
   assert,
   assertEquals,
   assertRejects,
+  assertStrictEquals,
   assertThrows,
   deferred,
   delay,
@@ -386,7 +387,7 @@ unitTest(
         Deno.errors.Http,
         "connection closed",
       );
-      // The error from `op_http_request_next` reroutes to `respondWith()`.
+      // The error from `op_http_accept` reroutes to `respondWith()`.
       assertEquals(await nextRequestPromise, null);
       listener.close();
     })();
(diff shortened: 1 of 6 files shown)


ext/http/01_http.js
```diff
@@ -84,6 +84,11 @@
         throw error;
       }
       if (nextRequest == null) {
+        // Work-around for servers (deno_std/http in particular) that call
```
piscisaureus, line 87: I believe this is indicative of a problem with our API, in particular two dubious practices clash here:
* `HttpConn` is automatically closed when `HttpConn.nextRequest()` fails or a return an error.
* `Deno.upgradeWebSocket()` doesn't actually do anything until the caller uses (returns) the response created by `Deno.upgradeWebSocket()`.

bartlomieju (approved):
LGTM

cli/tests/unit/http_test.ts
```diff
@@ -1000,6 +1002,80 @@ unitTest(
...
+
+    await Promise.all([server(), delay(100).then(client)]);
+  },
+);
+
+// https://github.com/denoland/deno/pull/12704
+// https://github.com/denoland/deno/pull/12732
+unitTest(
+  { permissions: { net: true } },
+  async function httpConnAutoCloseDelayedOnUpgrade() {
```
piscisaureus, line 1045: Regression test for the deno_std/http problem that got denoland#12332 reverted.

ext/http/01_http.js
lucacasonato: > Deno.upgradeWebSocket() doesn't actually do anything until the caller uses (returns) the response created by Deno.upgradeWebSocket()

Why is this a problem? This was done intentionally. The `Deno.upgradeWebSocket` call is just a glorified `new Response()`.

lucacasonato (approved):

PR #12742, opened 2021-11-12
fix(runtime/js): "Invalid URL" exception

max-lt:
In runtime/js/99_main.js -> [bootstrapWorkerRuntime](https://github.com/denoland/deno/blob/fdf890a68d3d54d40c766fd78faeccb20bd2e2c6/runtime/js/99_main.js#L614) function: 

As location is optional in bootstrap options, setLocationHref should only be called if location is not `null` (as it is in the [bootstrapMainRuntime](https://github.com/denoland/deno/blob/fdf890a68d3d54d40c766fd78faeccb20bd2e2c6/runtime/js/99_main.js#L577) function).
Otherwise there is a "Invalid URL" exception thrown while bootsraping a worker.

### DIFF
--- a/runtime/js/99_main.js
+++ b/runtime/js/99_main.js
@@ -665,7 +665,10 @@ delete Object.prototype.__proto__;
       cpuCount,
     } = runtimeOptions;
 
-    location.setLocationHref(locationHref);
+    if (locationHref != null) {
+      location.setLocationHref(locationHref);
+    }
+
     numCpus = cpuCount;
     registerErrors();
 


runtime/js/99_main.js
```diff
@@ -665,7 +665,10 @@ delete Object.prototype.__proto__;
       cpuCount,
     } = runtimeOptions;
 
-    location.setLocationHref(locationHref);
+    if (locationHref != null) {
+      location.setLocationHref(locationHref);
+    }
```
bartlomieju, line 670: Have you encountered a situation where `locationHref` was null? It's expected that this value is always set for workers.

bartlomieju (commented):
@max-lt please provide a test case that exercises the fix

runtime/js/99_main.js
max-lt: Not inside the deno project, but i'm using deno crates in a projet which provides a "Service Worker"-like environment like  deno deploy or cloudflare workers.

max-lt:
I'm closing this request as it may be not relevant from the Deno perspective; For what I'm doing I will need to rewrite an other runtime anyway.
For information, in case this problem surges later, this exception can be reproduced by creating a JsRuntime that includes deno_url and calls `bootstrap.workerRuntime` during bootsrap, with location set to None.

bartlomieju:
> By the way, thank you all for the Deno project, I really love it. And it's really well written, it's very simple to get through the internals even for someone like me that discovers Rust in the same time. I think nevertheless that you should OS the deno deploy project (at least the service worker crate), that would spare me its reimplementation x)

Thank you! Open sourcing Deno Deploy is unlikely; FYI there's no such thing as "service worker crate" in there, it's just modelled a bit after that API but the implementation is completely different.

Issue #12741, opened 2021-11-12
Body not being set to `null` when there is no body for certain methods


kitsonk:
In 1.16.0 any `Request` that didn't have a body was set to `null` irrespective of the method.  As of 1.16.1, requests that do not have a body for verbs like `DELETE` and `OPTIONS` contain a set body in the request, which is a significant breaking change in behaviour.

```js
const server = Deno.listen({ port: 8085 });

for await (const conn of server) {
  (async () => {
    const httpConn = Deno.serveHttp(conn);
    for await (const requestEvent of httpConn) {
      if (requestEvent.request.body) {
        console.log(`request ${requestEvent.request.method} has body`);
      } else {
        console.log(
          `request ${requestEvent.request.method} does not have body`,
        );
      }
      requestEvent.respondWith(
        new Response(null, { status: 204, statusText: "NoContent" }),
      );
    }
  })();
}
```

And then if you do a couple curls, like:

```
curl --request GET http://localhost:8085
curl --request OPTIONS http://localhost:8085
curl --request DELETE http://localhost:8085
```

You will see an output like:

```
request GET does not have body
request OPTIONS has body
request DELETE has body
```

According to the [spec](https://fetch.spec.whatwg.org/#body-mixin):

> The body getter steps are to return null if this’s body is null; otherwise this’s body’s stream.

Ref: oakserver/oak#430

c0per:
I find this bug still exists in deno v1.16.1

lucacasonato:
@kitsonk I can reproduce the change in behaviour, however fixing this actually fixed a HTTP/1.1 vs HTTP2 incompatibility. In the case of the OPTIONS and DELETE requests, there is a body, it just has length 0. Previously (before v1.16.1), HTTP/1.1 requests only had a `.body` property set if the body length was larger than 0. For HTTP/2, the `.body` property was always set for OPTIONS and DELETE requests.

The `.body` being `null` means that the request does definitely not have a body (this is the case for GET requests for example). The `.body` being set, means that there is a body, but it could be of length 0.

kitsonk:
Hrmmmm... not ideal for a patch release, but it is what it is. I will look into it more from an oak perspective and see what the best way to handle it.

rracariu:
This kind of fix should never be part of a . release. The fact that the major version works, but you update to patch and then all your stuff breaks is not great...

piscisaureus:
[...]

PR #12746, opened 2021-11-12
fix(ext/http): when body is empty, `Request.body` should be null

piscisaureus:
Fixes: #12741

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -1076,6 +1076,38 @@ unitTest(
   },
 );
 
+// https://github.com/denoland/deno/issues/12741
+unitTest(
+  { permissions: { net: true, run: true } },
+  async function httpServerDeleteRequestHasNoBody() {
+    const hostname = "localhost";
+    const port = 4501;
+
+    async function server() {
+      const listener = Deno.listen({ hostname, port });
+      const tcpConn = await listener.accept();
+      const httpConn = Deno.serveHttp(tcpConn);
+      const event = await httpConn.nextRequest() as Deno.RequestEvent;
+      assertStrictEquals(event.request.body, null);
+      const response = new Response();
+      await event.respondWith(response);
+      httpConn.close();
+      listener.close();
+    }
+
+    async function client() {
+      const url = `http://${hostname}:${port}/`;
+      const cmd = ["curl", "-X", "DELETE", url];
+      const proc = Deno.run({ cmd, stdout: "null", stderr: "null" });
+      const status = await proc.status();
+      assert(status.success);
+      proc.close();
+    }
+
+    await Promise.all([server(), client()]);
+  },
+);
+
 unitTest(
   { permissions: { net: true } },
   async function httpServerRespondNonAsciiUint8Array() {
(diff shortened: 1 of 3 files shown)


lucacasonato (changes requested):
I am not in favor, because `request.size_hint().exact()` always returns `None` on HTTP/2. This would make HTTP/1.1 and HTTP/2 work differently.

piscisaureus:
> I am not in favor, because `request.size_hint().exact()` always returns `None` on HTTP/2. This would make HTTP/1.1 and HTTP/2 work differently.

Yeah I had some doubts as well (I suspect it might not be possible to always detect an empty body even with HTTP/1.1, e.g. when using "transfer-encoding: chunked"). 

I'll leave this PR open for a bit to allow for the discussion, though.

piscisaureus:
If we decide to not land this, I'll change the test so that it checks "DELETE" requests do have a body.

kitsonk:
The biggest argument is that it is a pretty significant change in behaviour that we landed in a patch release.  Just relying on `transfer-encoding` or `content-length` don't look sufficient in determining if a body exists in oak, as it breaks other things.  I can be self absorbed about oak, but then there is every other RESTful Deno server out there that might have basic logic that tries to see if something has a body or not that could be broken now.

I get the inconsistent behaviour, but there are a few other subtle differences between HTTP 1.1 and HTTP 2 as well, like HTTP 2 doesn't support transfer encoding at all.

lucacasonato:
> I get the inconsistent behaviour, but there are a few other subtle differences between HTTP 1.1 and HTTP 2 as well, like HTTP 2 doesn't support transfer encoding at all.

But only because HTTP/2 is always transfers data in chunks.

kitsonk:
That is arguing it is different because it is different.  Which is the point I am trying to make.  There are different behaviours between HTTP/2 and HTTP/1.1.

piscisaureus:
> The biggest argument is that it is a pretty significant change in behaviour that we landed in a patch release.

I agree that this is not ideal, and a reason to seriously consider landing this patch.

> There are different behaviours between HTTP/2 and HTTP/1.1.

You opened #12741 because the ext/http refactor was breaking oak.
It turns out that the new behavior was already there but for HTTP2 only.
Does that mean that oak was broken with HTTP2 before?

kitsonk:
> Does that mean that oak was broken with HTTP2 before?

Yes, but no-one had reported it and I didn't test for it. I have fixed oak anyway, basically saying that `.hasBody` is an unreliable API and allowing empty bodies to be returned when a specific type of body is asserted, but trying to think of the greater good.

kitsonk:
And when we say broken, it meant that if you had HTTP/2 and sent a DELETE or OPTIONS request, it would report that the request had a body, but that body would be zero length. Which is accurate, but some people coupled to the behaviour of the HTTP/1.1 in their middleware, basically asserting that they didn't have bodies and broke their middleware logic.

Basically `.hasBody` isn't a good idea in HTTP/2 as it really is a Schrödinger's body in that it doesn't exist or not exist until it is observed. Now that behaviour occurs for native HTTP/1.1 results.

piscisaureus:
Discussed off-github: unfortunate as it is that this change was introduced in a a patch release, we think that in the long term this is the desirable behavior, so we are not going to land a patch to revert to pre-1.16.1 behavior.

Issue #11032, opened 2021-06-18
lsp: replace code lens navigation tree with swc "collector"


kitsonk:
Currently, the TypeScript code lenses for implementations and references use the TypeScript "navigation tree", which is a light-weight AST, to identify positions in the code that might be "interesting" to lazily ask the TypeScript language server about at a future point.

Getting a navigation tree is one of the slower processes with the language server, and is several orders of magnitude slower than visiting an swc parsed AST.  In fact, we actually don't currently cache/reuse an swc parsed AST for a file across the language server, reparsing a file a few times depending on the types of requests to a language server.

I recently refactored the code for the code lenses to make them a little bit more self contained.  The logic for walking the "navigation tree" AST is in [lsp::code_lens::collect_tsc](https://github.com/denoland/deno/blob/main/cli/lsp/code_lens.rs#L412) and I implemented with test code lens using the "collector" pattern in [lsp::code_lens::collect_test](https://github.com/denoland/deno/blob/main/cli/lsp/code_lens.rs#L378) which is in my opinion the way we should refactor the TypeScript code lenses.

Also, caching the swc parse of the module in the document info would be a good thing I think.  That might take a bit more structuring of code.  It is used in dependency analysis of the document as well as the test code lens and also use by the linting (though I don't know how easy it would be to send it to deno_lint without a bigger refactor of one or the other).

This is the refactor I had in my mind @dsherret that I mentioned to you. Let me know if you want to work on it, if not, we can always add it to my backlog. 😄

kitsonk:
@dsherret if you don't feel too strongly about this one, I would be interested in taking this one back up.  I think it could dramatically improve the performance of the LSP.

dsherret:
@kitsonk I was thinking about working on it in January/Feb after the bulk of the node compatibility improvements. Feel free to take a look sooner if you'd like.

dsherret:
I've been looking at the LSP code more over the past few days and eliminating the `NavigationTree` would be very nice. Right now there's a race condition if we were to make the code concurrent in that an out of date navigation tree might end up being stored in a document that it wasn't created from. This could be eliminated by implementing this feature and it would reduce our mutation points.

kitsonk:
To fix that race condition, we need to supply the LSP version on the call and have it come back in the response and abandon it.  But the bigger win would be cancelling in flight requests, but that is a bit of a challenge.

nayeemrmn:
[...]

PR #12747, opened 2021-11-12
refactor(lsp): remove `Documents` mutex and require `Documents` to be mutated to change it

dsherret:
This change removes the mutex on `Documents` and pushes it down to a new internal mutex on `FileSystemDocuments`. The API for `Documents` is still the same other than requiring mutation in order to change it.

This will help us have concurrent "read requests" in the LSP because it requiring mutation ensures the documents collection doesn't change while processing a read request.

Todo:

- [x] Collapse `DocumentsInner` into `Documents` after an initial review (deferring this will make the initial review a little easier).


cli/lsp/documents.rs
```diff
@@ -1057,61 +1100,99 @@ impl DocumentsInner {
...
+          .collect()
+      } else {
+        HashMap::new()
+      },
+    );
     self.dirty = true;
   }
 }
 
+fn calculate_fs_version(
```
dsherret, line 1128: Both these functions were extracted out for reuse.

cli/lsp/documents.rs
```diff
@@ -1057,61 +1100,99 @@ impl DocumentsInner {
...
-          dependencies.into_iter().map(|s| (s, None)).collect();
-        let module = SyntheticModule::new(
-          referrer.clone(),
-          dependencies,
-          self.get_maybe_resolver(),
-        );
-        self.imports.insert(referrer, module);
-      }
-    }
+    self.imports = Arc::new(
```
dsherret, line 1103: This seems more correct, but maybe not? I believe we should replace `imports` when a configuration file changes rather than always inserting. Maybe I'm misunderstanding?

cli/lsp/diagnostics.rs
```diff
@@ -351,7 +351,7 @@ async fn generate_lint_diagnostics(
 }
 
 async fn generate_ts_diagnostics(
-  snapshot: &language_server::StateSnapshot,
+  snapshot: Arc<language_server::StateSnapshot>,
```
dsherret, line 354: Another change in this PR is to wrap `StateSnapshot` in an `Arc` to (1) prevent someone thinking it would be ok to mutate the values in it and (2) documents is more expensive to clone now so this makes cloning a state snapshot very cheap.

cli/lsp/documents.rs
```diff
@@ -606,46 +606,81 @@ fn recurse_dependents(
 }
 
 #[derive(Debug, Default)]
-struct DocumentsInner {
-  /// The DENO_DIR that the documents looks for non-file based modules.
+struct SpecifierResolver {
```
dsherret, line 609: Extracted this out to isolate the `redirects` mutation behind a mutex because it is used by methods on `Document` that don't require mutation now such as `get`.

cli/lsp/documents.rs
```diff
@@ -1037,10 +1071,19 @@ impl DocumentsInner {
     specifier: &ModuleSpecifier,
     navigation_tree: Arc<tsc::NavigationTree>,
   ) -> Result<(), AnyError> {
```
dsherret, line 1073: This `set_navigation_tree` method is somewhat prone to a race condition in which a navigation tree could be generated from a document on the file system, then updated on the file system, then this could set a navigation tree that's out of date. I think it would be extremely rare to happen if it ever did (because it would require an FS refresh), but I will prevent it from happening anyway in this PR soon.

cli/lsp/documents.rs
```diff
@@ -1057,61 +1100,99 @@ impl DocumentsInner {
...
+    let path = cache.get_cache_filename(specifier)?;
+    if path.is_file() {
+      Some(path)
+    } else {
+      None
+    }
+  }
+}
+
+// todo(in this PR): collapse DocumentsInner into Documents after initial review
```
dsherret, line 1158: I didn't want to make this PR too hard to review so I'm going to collapse the above into `Documents` after an initial review.

bartlomieju (commented):
I'm strongly in favor of this change, having `StateSnapshot` hidden behind an `Arc<>` prevents it from being inadvertently mutated (ie. if someone wanted to change from `snapshot: &language_server::StateSnapshot` to `snapshot: &mut language_server::StateSnapshot` there's nothing preventing doing that, but it becomes almost impossible when using `Arc<>`).

cli/lsp/documents.rs
dsherret: Done in 13ca3bd29b7b7e4d77247fc68b0509585cd71291

cli/lsp/documents.rs
```diff
@@ -733,38 +801,54 @@ impl DocumentsInner {
...
-        Err(custom_error(
-          "NotFound",
-          format!("The specifier \"{}\" was not found.", specifier),
-        ))
-      },
-      Ok,
-    )?;
-    *doc = doc.with_closed();
+    let mut file_system_docs = self.file_system_docs.lock();
+    let doc = self
```
kitsonk, line 829: I am not convinced 100% of this.  We should just lazily `::new()` the local file when needed.  The file could easily be modified by another process outside of the editor, so we can't really trust anything we copy over.

cli/lsp/documents.rs
```diff
@@ -781,48 +865,49 @@ impl DocumentsInner {
...
-  fn contains_specifier(&mut self, specifier: &ModuleSpecifier) -> bool {
-    let specifier = self
-      .resolve_specifier(specifier)
-      .unwrap_or_else(|| specifier.clone());
-    if !self.is_valid(&specifier) {
-      self.add(specifier.clone());
-    }
-    self.docs.contains_key(&specifier)
+  fn contains_specifier(&self, specifier: &ModuleSpecifier) -> bool {
+    self.get(specifier).is_some()
```
kitsonk, line 869: you are convinced that cloning a document is lightweight?  This will clone any in memory documents just to find out if they are present.

cli/lsp/documents.rs
```diff
@@ -781,48 +865,49 @@ impl DocumentsInner {
...
 
-  fn get(&mut self, specifier: &ModuleSpecifier) -> Option<&Document> {
-    let specifier = self.resolve_specifier(specifier)?;
-    if !self.is_valid(&specifier) {
-      self.add(specifier.clone());
+  fn get(&self, specifier: &ModuleSpecifier) -> Option<Document> {
+    let specifier = self.specifier_resolver.resolve(specifier)?;
+    if let Some(document) = self.open_docs.get(&specifier) {
+      Some(document.clone())
+    } else if let Some(specifier) = self.specifier_resolver.resolve(&specifier)
```
kitsonk, line 890: How can this be different that the response at 887?

cli/lsp/documents.rs
```diff
@@ -984,63 +1059,38 @@ impl DocumentsInner {
...
     &mut self,
     specifier: &ModuleSpecifier,
+    script_version: &str,
     navigation_tree: Arc<tsc::NavigationTree>,
   ) -> Result<(), AnyError> {
-    let doc = self.docs.get_mut(specifier).ok_or_else(|| {
-      custom_error("NotFound", format!("Specifier not found {}", specifier))
-    })?;
-    *doc = doc.with_navigation_tree(navigation_tree);
+    if let Some(doc) = self.open_docs.get_mut(specifier) {
```
kitsonk, line 1075: does this correctly handle situations where there is already a cached navigation tree that is now invalid? (e.g. there is an invalid nav tree already cached, we go to calculate a new one, but the document has changed while calculating it meaning we don't update the old old nav tree, meaning the cached one is even more out of date)

Also, does it correctly handle situations where we don't get into an infinite loop of constantly trying to update the nav tree?

kitsonk: It does feel overall unrelated to the _core_ of this PR.

cli/lsp/documents.rs
kitsonk: I considered it, and there is no _harm_ in inserting, in that they are simply in memory and available, meaning that there could be old references or state referring to it, and therefore less likely to cause exotic errors.

cli/lsp/documents.rs
kitsonk: nit: doc comments

kitsonk (commented):
Just a few questions/thoughts.

cli/lsp/documents.rs
kitsonk: Before, it simply would remove the version and clear the nav tree.  I guess why not drop it?  If the file is closed and isn't referenced anywhere else it doesn't need to be in the fs files anyways, and if it does, it will get added.

cli/lsp/documents.rs
kitsonk: Ok... I was looking at code in vscode recently and they have effectively a debounce for this:

```ts
function waitForDocumentChangesToEnd(document: vscode.TextDocument) {
	let version = document.version;
	return new Promise<void>((s) => {
		const iv = setInterval(_ => {
			if (document.version === version) {
				clearInterval(iv);
				s();
			}
			version = document.version;
		}, 400);
	});
}
```

We put a debounce in calculating diagnostics as well (basically debouncing the calculation of them because they can occur async).  It isn't the scope of this PR, but we might want to consider this.  Both the semantic tokens and the nav tree are the costly tsc calls and they are the things that can make it feel sluggish.

kitsonk (approved):
LGTM, though #12782 is likely to cause conflicts.  I think we should land #12782 first because it fixes a bug and then rebase this one?

dsherret:
@kitsonk yeah definitely! I won't land this one until after the next patch release. We should get that one in.

cli/lsp/documents.rs
dsherret: I investigated this in more detail and it looks like `get_workspace_completions` is the only place that uses `.documents(false, _)` where it matters. It seems like this would cause a behaviour change, but I'm not sure. Wouldn't it be possible for someone to have a remote file in the cache, open it, close it, and now it doesn't show up in the completions? Is that a big deal?

kitsonk: very very very very unlikely, because if the module is elsewhere in the dependency graph in tsc it would be re-opened as an fs file before that code ever ran, and that code is related to completing specifiers when someone is typing a new import specifier (not referencing workspace symbols for example).  So they would have to be super fast, basically close the file and already be typing a specifier in another open editor file and it likely would "catch-up" after they type the first character.

dsherret: Ok cool. Thanks! I'll make this change.

cli/lsp/documents.rs
dsherret: It might be good to do in the meantime until we can do #11032. I think very likely that one will make it fast enough that it won't matter (and we'll be able to get rid of this caching code too).

cli/lsp/documents.rs
dsherret: Done in a5c8789c8fabb009fbc19da9800287ecc0d106bf. I'll test this PR out more tomorrow morning and then merge it then.

PR #12748, opened 2021-11-12
fix: support typechecking docs with CRLF line endings

### LINKED ISSUE #12187: `deno test --doc` doesn't detect errors in Markdown files with CRLF line endings

GJZwiers:
This fixes #12187 and enables `deno test --doc` to typecheck docs that make use of `\r\n` line endings (CRLF). An additional integration test has been added that asserts that the doc typechecker has the same output as it did for the test which has LF. I updated  `.gitattributes` to keep CRLF in one  `.md` file in the `testdata/test` directory, and added some ignore/excludes for this file where needed.

### DIFF
--- a/cli/tests/integration/fmt_tests.rs
+++ b/cli/tests/integration/fmt_tests.rs
@@ -129,7 +129,7 @@ fn fmt_ignore_unexplicit_files() {
 }
 
 itest!(fmt_check_tests_dir {
-  args: "fmt --check ./ --ignore=.test_coverage,fmt/fmt_with_config/",
+  args: "fmt --check ./ --ignore=.test_coverage,fmt/fmt_with_config/,test/markdown_windows.md",
   output: "fmt/expected_fmt_check_tests_dir.out",
   exit_code: 1,
 });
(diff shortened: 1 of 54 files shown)


cli/tests/integration/fmt_tests.rs
```diff
@@ -129,7 +129,7 @@ fn fmt_ignore_unexplicit_files() {
 }
 
 itest!(fmt_check_tests_dir {
-  args: "fmt --check ./ --ignore=.test_coverage,fmt/fmt_with_config/",
+  args: "fmt --check ./ --ignore=.test_coverage,fmt/fmt_with_config/,test/markdown_windows.md",
```
bartlomieju, line 132: This is unfortunate, I opened https://github.com/denoland/deno/issues/12758 to fix this situation

dsherret (approved):
LGTM. Thanks @GJZwiers!

Issue #12744, opened 2021-11-12
deno test: Add more overloads of Deno.test()


bartlomieju:
Currently `Deno` provides following two overloads of `Deno.test()` API:
```ts
export function test(t: TestDefinition): void;
// Deno.test({
//  name: "test name",
//  fn: function () { 
//  }
// });

export function test(
    name: string,
    fn: (t: TestContext) => void | Promise<void>,
): void;
// Deno.test("test name", () => { });
```

I propose to add four more overloads like this:
```ts
export function test(fn: (t: TestContext) => void | Promise<void>): void;
// Deno.test(function testName() { });

export function test(name: string, options: Omit<TestDefinition, "name">, fn: (t: TestContext) => void | Promise<void>): void;
// Deno.test("test name", { only: true }, function() { });

export function test(options: Omit<TestDefinition, "fn">, fn: (t: TestContext) => void | Promise<void>): void;
// Deno.test({ name: "test name" }, function() { });

export function test(options: Omit<TestDefinition, "fn" | "name">, fn: (t: TestContext) => void | Promise<void>): void;
// Deno.test({ only: true }, function testName() { });
```

The reason is two fold:
- it would allow us to completely drop helper called `unitTest` that we use internally for unit tests of the runtime (this has been my goal for the longest time, we've been dogfooding numerous changes to the testing system through unit tests; after making `Deno.TestDefinition.permissions` stable these overloads are the last bit needed to get rid of `unitTest` wrapper)
- increase user ergonomy - currently it's very cumbersome if one wants to mark a test as "only" - if you use shorthand overload, you need to rewrite it to the `TestDefinition` form (ie. the object form), with proposed overloads it would be a matter of adding `{ only: true }` as first or second argument

Ref #11764

CC @dsherret @caspervonb @lucacasonato @kitsonk

caspervonb:
Always end up having a wrapper in the wild for this.
The internal `unitTest` does nothing but provide the second overload so sgtm!

PR #12749, opened 2021-11-13
feat(test): Add more overloads for "Deno.test"

bartlomieju:
This commit adds 4 more overloads to "Deno.test()" API.

```
// Deno.test(function testName() { });
export function test(fn: (t: TestContext) => void | Promise<void>): void;

// Deno.test("test name", { only: true }, function() { });
export function test(
  name: string,
  options: Omit<TestDefinition, "name">,
  fn: (t: TestContext) => void | Promise<void>,
): void;

// Deno.test({ name: "test name" }, function() { });
export function test(
  options: Omit<TestDefinition, "fn">,
  fn: (t: TestContext) => void | Promise<void>,
): void;

// Deno.test({ only: true }, function testName() { });
export function test(
  options: Omit<TestDefinition, "fn" | "name">,
  fn: (t: TestContext) => void | Promise<void>,
): void;
```

Closes https://github.com/denoland/deno/issues/12744

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -409,7 +409,7 @@ declare namespace Deno {
    *   assertEquals("hello", "hello");
    * });
    *
-   * Deno.test({ permissions: { read: false } }, function myOtherTestName(): Promise<void> {
+   * Deno.test({ permissions: { read: false } }, async function myOtherTestName(): Promise<void> {
    *   const decoder = new TextDecoder("utf-8");
    *   const data = await Deno.readFile("hello_world.txt");
    *   assertEquals(decoder.decode(data), "Hello world");
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -371,7 +371,7 @@ declare namespace Deno {
    */
   export function test(
     name: string,
-    options: Omit<TestDefinition, "fn" | "name">,
+    options: Omit<TestDefinition, "fn">,
     fn: (t: TestContext) => void | Promise<void>,
   ): void;
 
@@ -394,7 +394,7 @@ declare namespace Deno {
    * ```
    */
   export function test(
-    options: Omit<TestDefinition, "fn">,
+    options: Omit<TestDefinition, "fn" | "name">,
     fn: (t: TestContext) => void | Promise<void>,
   ): void;
 
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -371,7 +371,7 @@ declare namespace Deno {
    */
   export function test(
     name: string,
-    options: Omit<TestDefinition, "fn">,
+    options: Omit<TestDefinition, "fn" | "name">,
     fn: (t: TestContext) => void | Promise<void>,
   ): void;
 
@@ -394,10 +394,32 @@ declare namespace Deno {
    * ```
    */
   export function test(
-    options: Omit<TestDefinition, "fn" | "name">,
+    options: Omit<TestDefinition, "fn">,
     fn: (t: TestContext) => void | Promise<void>,
   ): void;
 
+  /** Register a test which will be run when `deno test` is used on the command
+   * line and the containing module looks like a test module.
+   * `fn` can be async if required. Declared function must have a name.
+   *
+   * ```ts
+   * import {assert, fail, assertEquals} from "https://deno.land/std/testing/asserts.ts";
+   *
+   * Deno.test({ permissions: { read: true } }, function myTestName(): void {
+   *   assertEquals("hello", "hello");
+   * });
+   *
+   * Deno.test({ permissions: { read: false } }, function myOtherTestName(): Promise<void> {
+   *   const decoder = new TextDecoder("utf-8");
+   *   const data = await Deno.readFile("hello_world.txt");
+   *   assertEquals(decoder.decode(data), "Hello world");
+   * });
+   * ```
+   */
+   export function test(
+    options: Omit<TestDefinition, "fn" | "name">,
+    fn: (t: TestContext) => void | Promise<void>,
+  ): void;
   /** Exit the Deno process with optional exit code. If no exit code is supplied
    * then Deno will exit with return code of 0.
    *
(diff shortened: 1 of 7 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -331,6 +331,73 @@ declare namespace Deno {
...
+   * `fn` can be async if required. Declared function must have a name.
+   *
+   * ```ts
+   * import {assert, fail, assertEquals} from "https://deno.land/std/testing/asserts.ts";
+   *
+   * Deno.test(function myTestName(): void {
+   *   assertEquals("hello", "hello");
+   * });
+   *
+   * Deno.test(async function myOtherTestName(): Promise<void> => {
```
bartlomieju, line 345: I'm fine with adding these overloads to unstable declaration first, but I'd rather land it directly to stable if there's support for it.

runtime/js/40_testing.js
```diff
@@ -266,22 +267,66 @@ finishing test case.`;
...
+      }
+      if (typeof optionsOrFn === "function") {
+        testDef = { fn: optionsOrFn, name: nameOrFnOrOptions, ...defaults };
+      } else {
+        if (!maybeFn || typeof maybeFn !== "function") {
+          throw new TypeError("Missing test function");
+        }
+        // TODO(bartlomieju):
+        if (optionsOrFn.fn != undefined) {
+          throw new TypeError("Missing test function");
```
bartlomieju, line 282: Change error message

runtime/js/40_testing.js
```diff
@@ -266,22 +267,66 @@ finishing test case.`;
...
+        }
+        if (optionsOrFn.name != undefined) {
+          throw new TypeError("Missing test function");
```
bartlomieju, line 285: Change error message

runtime/js/40_testing.js
```diff
@@ -266,22 +267,66 @@ finishing test case.`;
...
+          name: nameOrFnOrOptions,
+        };
+      }
+    } else if (typeof nameOrFnOrOptions === "function") {
+      if (!nameOrFnOrOptions.name) {
+        throw new TypeError("The test function must have a name");
+      }
+      // TODO(bartlomieju):
+      if (optionsOrFn != undefined) {
         throw new TypeError("Missing test function");
```
bartlomieju, line 300: Change error message

runtime/js/40_testing.js
```diff
@@ -266,22 +267,66 @@ finishing test case.`;
...
       }
-      if (!t) {
-        throw new TypeError("The test name can't be empty");
+      if (maybeFn != undefined) {
+        throw new TypeError("Missing test function");
```
bartlomieju, line 303: Change errors message

cli/tests/integration/test_tests.rs
```diff
@@ -19,6 +19,12 @@ fn no_color() {
   assert!(out.contains("test result: FAILED. 1 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out"));
 }
 
+itest!(overloads {
+  args: "test test/overloads.ts",
+  exit_code: 0,
+  output: "test/overloads.out",
+});
+
```
bartlomieju, line 27: Add more tests that verify error messages for improper usage of overloads

runtime/js/40_testing.js
```diff
@@ -266,22 +267,69 @@ finishing test case.`;
...
+          ...defaults,
+          ...optionsOrFn,
+          fn: maybeFn,
+          name: nameOrFnOrOptions,
+        };
+      }
+    } else if (typeof nameOrFnOrOptions === "function") {
+      if (!nameOrFnOrOptions.name) {
+        throw new TypeError("The test function must have a name");
+      }
```
nayeemrmn, line 297:
```suggestion
```

runtime/js/40_testing.js
```diff
@@ -266,22 +267,69 @@ finishing test case.`;
...
-      if (!t) {
-        throw new TypeError("The test name can't be empty");
+      if (maybeFn != undefined) {
+        throw new TypeError("Missing test function");
       }
-      testDef = { fn: fn, name: t, ...defaults };
+      testDef = {
+        ...defaults,
+        fn: nameOrFnOrOptions,
+        name: nameOrFnOrOptions.name,
```
nayeemrmn, line 308:
```suggestion
        name: nameOrFnOrOptions.name ?? "<anonymous>",
```

nayeemrmn (commented):
I think we should allow unnamed functions and display them as `test <anonymous> ... ok`. Sometimes I want to rapidly copy and paste a bug repro into something like:
```ts
Deno.test(() => {
  /// code
});
```
Plus users might find themselves doing this at the earliest stage of a project with just one test.

bartlomieju:
> I think we should allow unnamed functions and display them as `test <anonymous> ... ok`. Sometimes I want to rapidly copy and paste a bug repro into something like:
> 
> ```ts
> Deno.test(() => {
>   /// code
> });
> ```
> 
> Plus users might find themselves doing this at the earliest stage of a project with just one test.

Maybe instead of `test <anonymous> ... ok` we could do something like `test file:///path/to/file.ts:L140-160` like we do in "doc tests"?

nayeemrmn:
> Maybe instead of `test <anonymous> ... ok` we could do something like `test file:///path/to/file.ts:L140-160` like we do in "doc tests"? That way there might be multiple functions like this and they would be distinguished

No strong opinions on the formatting. I'm not sure that can be done though since there's no static analysis. I also don't think distinguishing them is a problem. You can already give two tests the same name, providing distinct names is up to the user, letting two tests be anonymous is just a case of that.

bartlomieju:
I think it's not that bad upon registration we can create an error and collect the stack.

bartlomieju:
@nayeemrmn I will add your suggestion about anonymous tests in a follow up PR.

runtime/js/40_testing.js
```diff
@@ -266,22 +267,73 @@ finishing test case.`;
...
+      if (typeof optionsOrFn === "function") {
+        fn = optionsOrFn;
+        if (nameOrFnOrOptions.fn != undefined) {
+          throw new TypeError(
+            "Unexpected 'fn' field in options, test function is already provided as the second argument.",
+          );
+        }
+        name = nameOrFnOrOptions.name ?? fn.name;
+      } else {
+        if (!nameOrFnOrOptions.fn) {
```
dsherret, line 324: May want to consider joining these two conditions like what was done on line 277

dsherret (approved):
LGTM

lucacasonato (approved):

PR #12750, opened 2021-11-13
refactor: remove "unitTest" wrapper from cli/tests/unit

bartlomieju:
Proof of concept for removing `unitTest` helper function defined in `cli/tests/unit/test_util.ts`.

Stacked on top of #12749
Ref #12744

There's a slight difference between `unitTest` wrapper and `Deno.test` - the former was combining passed permissions with "false" permissions for omitted keys, while the latter inherits process permissions. This lead to about 10 tests failing but I changed them to explicitly deny certain permissions. I still think it's valuable to get rid of of `unitTest` wrapper and explicitly assign `false` to permissions we want to be denied in the test case.


ry (approved):
LGTM - I've verified that test runner still catches exceptions in tests. Nevertheless we should always be careful when landing changes like this that we aren't unintentionally ignoring errors.

bartlomieju:
> LGTM - I've verified that test runner still catches exceptions in tests. Nevertheless we should always be careful when landing changes like this that we aren't unintentionally ignoring errors.

Agreed, but this PR just removes the wrapper around `Deno.test()` - I fixed up all tests that depended on the fact that `unitTest` wrapper assigned all unspecified permissions to `false` instead of inheriting from process flags. So now if a test expects that eg. `read` permission is denied it is typed out explicitly as `{ permissions: { read: false } }`

PR #12765, opened 2021-11-15
chore: bump deno_graph, deno_doc deps

kitsonk:


kitsonk:
Needed to update one of the fixtures here, which now has a redundant "source", which we can try to clean up when we resolve the other issues.

kitsonk:
Actually closing in favour of #12604

nayeemrmn:
Yeah we have those redundant sources already e.g. https://github.com/denoland/deno/blob/v1.16.1/cli/tests/testdata/error_015_dynamic_import_permissions.out, but it's more consistent now. They're only redundant when the module graph error is at the entrypoint of the dynamic import, but necessary when it's at a static dependency of the dynamically imported module. #12604 has a solution which distinguishes these cases: https://github.com/denoland/deno/blob/1b3b0f5993d90fdda36540f028112e774849c02d/cli/proc_state.rs#L121.

Issue #12726, opened 2021-11-10
Include unstable APIs type declarations even if --unstable flag is not present


bartlomieju:
As per title; I suggest we change the behavior of `--unstable` flag to only have impact during runtime. 

Currently type checking includes different declaration files depending whether `--unstable` flag is present or not. This creates friction for users with not much benefit (IMO). We have a system that rewrites messages coming from TSC about non-existent types (by using `UNSTABLE_DENO_PROPS` list in `cli/diagnostics.rs`) and suggests to run again with `--unstable` flag, however that system is not completely reliable and some errors don't return any meaningful suggestions. Additionally, the system requires us to manually sync which APIs are considered unstable and as can be seen in https://github.com/denoland/deno/pull/12723 we sometimes forget to update the list. One might argue that we should update the system and/or add more checks that will ensure that the `lib.deno.unstable.d.ts` and `UNSTABLE_DENO_PROPS` list is kept in sync, but IMO this will lead to even more complexity, while still not being able to catch all situations where `--unstable` flag is missing (ergo still showing non-helpful message to users).

Another reason is that some `deno_std` modules use unstable APIs (https://github.com/denoland/deno_std/issues/1539, or half of `deno_std/node`), and even though users might not use any of the unstable APIs, they are still required to pass `--unstable` flag. Again, one might argue that we should refactor `deno_std` in a way that splits modules depending if they use unstable APIs or not (eg. `fs/mod.ts` and `fs/mod.unstable.ts`). For me this adds more maintenance burden as well as increases users friction and discoverability of the modules.

So I propose we always include type declarations for unstable APIs during type checking. To better signal to users that they are depending on unstable APIs we might leverage [`DiagnosticTag::DEPRECATED`](https://docs.rs/lsp-types/0.91.1/lsp_types/struct.DiagnosticTag.html#associatedconstant.DEPRECATED) in the LSP, which would strike-through the API name, prompting users to hover over the API, which would show the docstring that starts with the description that the API is not yet stable.

ry:
I’m in favor of this. It’s not particularly dangerous - we still have the ability to change unstable APIs. It’s just users will get runtime errors instead of type errors.

dsherret:
I think @nayeemrmn makes a good point here—adding even optional properties to an interface could potentially break something somewhere. It's true that people right now would probably need to deal with this anyway with users using `--unstable` so maybe this is not a big deal though?

Another issue is that including these unstable types might encourage users to rely on unstable more.

[...]

PR #12766, opened 2021-11-15
feat: add `--no-check=remote` flag

### LINKED ISSUE #11970: Don't report diagnostics for "external" TypeScript code

kitsonk:
Closes #11970

This is an MVP feature that allows `--no-check=remote` to be passed on the command line, and will cause diagnostics related to remote modules to be "filtered" and not cause the process to terminate.

It uses a simplistic concept of remote modules at this point, which is those diagnostics where the specifier for the diagnostics starts with `http`.  This means type errors in a remote module that import a `data` URL would not be covered.  This would require an increased complexity of determining the importer of a specifier and determining if that importer is a remote module or not.

Also realised that I currently haven't included file names that start with `asset` or `deno`, which might also make sense, as users can't "control" errors from those locations.

If we ever considered this the default behaviour, we might want to consider a warning indicating that some diagnostics were filtered out and to use a different flag to have them displayed.


cli/flags.rs
```diff
@@ -2334,8 +2361,16 @@ fn compat_arg_parse(flags: &mut Flags, matches: &ArgMatches) {
...
-  if matches.is_present("no-check") {
-    flags.no_check = true;
+  if let Some(cache_type) = matches.value_of("no-check") {
+    match cache_type {
+      "remote" => flags.check = CheckFlag::Local,
+      _ => debug!(
+        "invalid value for 'no-check' of '{}' using default",
+        cache_type
+      ),
+    }
```
bartlomieju, line 2371: Maybe use [`possible_values`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.possible_values) in flag definition instead of manually parsing? Or does it not play nicely with "no value"?

bartlomieju (approved):
LGTM, I'm in favor of this change, I agree it's better approach than #12726

kitsonk:
Would be good to get @lucacasonato's view on this, since it was his idea.

lucacasonato:
Yeah, sorry. Will review Monday morning!

lucacasonato (approved):
Tried it out locally, and this works great. Thanks @kitsonk!

PR #12767, opened 2021-11-15
feat(cli/config): Add permissions support

### LINKED ISSUE #12763: Proposal: add permissions to config file
## Context

Currently Deno supports a `deno.jsonc` configuration file which allow users to provide a [configuration file](https://deno.land/manual@v1.16.1/getting_started/configuration_file) that can provide TypeScript compiler options, lint options and format options.

It does not currently support other information that can only be expressed on the command line, while this proposal is what

## Semantics

- When a configuration file is applied and the `"permissions"` section is parsed and the permissions are applied from the `"permissions"` section. Any other flags on the command line are ignored.
- If there are flags on the command line and a config file is being applied, and the config file contains `"permissions"` a warning should be issued to stderr that permissions from the config file are being applied.
- Remote configuration files are supported, but a summary of the permissions is prompted before hand and requires user configuration to continue.
- Keys names match the flags on the command line, and values accept a `true` or `false`. If the key supports a value, these can be either a string delimited the same way on the command line, or an array of strings.
- In situations where a base path is needed for relative paths (for `--allow-read` and `--allow-write`) the config file is used as a base, versus the cwd.

## Examples

An example of a configuration file using permissions:

```jsonc
{
  "permissions": {
    "allow-all": true,
    "allow-env": true,
    "allow-hrtime": false,
    "allow-read": [ ".", "/tmp" ],
    "allow-write": ".",
    "allow-net": "deno.land,nest.land"
  }
}
```

## Considerations / Open Questions

- Having explicit a _section_ of `"permissions"` makes it easier to understand explicitly that these effect the runtime permissions.  It also allows the definition of the semantics of `"permissions"` to evolve independently of the rest of the configuration file, as well as opens an easy opportunity to be able to set permissions on other things in the future, like _tasks_/_scripts_ independent of the top level permissions.
- Having an explicit `allow` in the keys provides a mechanism to introduce `block` in the future (for example `"block-net": "deno.land"`) giving more granular permissions.
- If a remote configuration file is used, and there is no TTY or `--no-prompt`/`--quit` is supplied on the configuration is set, should the process just terminate or just allow the program to run with the supplied permissions?
- For remote config files, using the remote URL as a relative base for `--allow-read` `--allow-write` is not possible.  Does this mean that relative paths just error, or that it defaults to the cwd in those situations?

cc/ @bartlomieju @ry

crowlKats:
Closes #12763.

Going to add tests later. Some things like naming are still to be decided on

### DIFF
--- a/runtime/permissions.rs
+++ b/runtime/permissions.rs
@@ -1114,7 +1114,7 @@ impl Default for Permissions {
   }
 }
 
-#[derive(Clone, Debug, PartialEq, Default, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Default, Serialize)]
 pub struct PermissionsOptions {
   pub allow_env: Option<Vec<String>>,
   pub allow_hrtime: bool,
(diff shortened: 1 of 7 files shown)


runtime/permissions.rs
```diff
@@ -1126,6 +1126,156 @@ pub struct PermissionsOptions {
...
+          fn visit_seq<V>(self, mut v: V) -> Result<Self::Value, V::Error>
+          where
+            V: de::SeqAccess<'de>,
+          {
+            let mut granted_list = vec![];
+            while let Some(value) = v.next_element::<String>()? {
+              granted_list.push(value);
+            }
+            Ok(Some(granted_list))
+          }
```
crowlKats, line 1168: should encountering an empty seq return `None`? because ie
```json
"permissions": {
    "net": [],
  }
  ```
  is the same as `"net": true`, though an empty array seems more like `"net": false` behaviour would be expected

Issue #12536, opened 2021-10-25
Add `DENO_CACHED_ONLY` env var


lucacasonato:
This environment variable would have the same effect as setting `--cached-only` in `deno run`.

It would be very useful to use in CI when vendoring DENO_DIR: I don't want certain CI flows to ever pull from the web.

bartlomieju:
Closing this. A dedicated `DENO_CACHED_ONLY` env var isn't planned.

For the use case described — hermetic CI that should never pull from the web with a pre-populated `DENO_DIR` — the recommended approach today is vendoring (`"vendor": true` in `deno.json`, or a committed `vendor/` directory) together with a lockfile, plus passing `--cached-only` on the deno invocation. That combination gives you reproducible, offline builds and fails loudly if something isn't cached.

A global env var that silently forces offline mode for every invocation is the kind of thing that tends to produce confusing failures, and the flag plus vendoring already covers the scenario. Please reopen if there is a concrete case where you can't control the invocation and vendoring doesn't suffice.

PR #12769, opened 2021-11-15
feat: Add DENO_CACHED_ONLY env var as alternative to --cached-only

ry:
Fixes https://github.com/denoland/deno/issues/12536


cli/proc_state.rs
```diff
@@ -164,15 +164,16 @@ impl ProcState {
...
-      CacheSetting::Only
-    } else if !flags.cache_blocklist.is_empty() {
-      CacheSetting::ReloadSome(flags.cache_blocklist.clone())
-    } else if flags.reload {
-      CacheSetting::ReloadAll
-    } else {
-      CacheSetting::Use
-    };
+    let cache_usage =
+      if flags.cached_only || env::var("DENO_CACHED_ONLY").is_ok() {
```
lucacasonato, line 168: It just being set is not good enough I think. We should also consider if it is set to false as it being unset. Maybe error if it is a value other than "true" or "false" or "" (the last being an alias for "false")

ry: this is uncommon behavior - most unix applications only test for it being set

lucacasonato: @piscisaureus Do you have an opinion?

lucacasonato: > most unix applications only test for it being set

I'd argue that even if it is unix convention, `DENO_CACHED_ONLY=false` meaning the same as `DENO_CACHED_ONLY=true` is unexpected to many. I think we can just error when the value of `DENO_CACHED_ONLY` is anything other than `1` or `true`.

ry: But why would anyone ever do `DENO_CACHED_ONLY=false`? It's silly. This is a perennial argument we have.

Sure, I'll implement this non-standard behavior. But it's silly. I assure you binary env vars in Golang do not do this.

lucacasonato: > But why would anyone ever do DENO_CACHED_ONLY=false?

It's not unlikely someone would do this I'd think. I also don't see the harm about just enforcing one specific value rather than just allowing all. Being more strict will catch accidental issues early.

```js
const useCachedOnly: boolean = ...;

Deno.run({
  cmd: ["deno", ...],
  env: {
    "DENO_CACHED_ONLY": String(useCachedOnly)
  }
})
```

ry: Updated. Is this ok? Or do we also need to suppose "DENO_CACHED_ONLY=False" ? What about "DENO_CACHED_ONLY=No" ?

lucacasonato: Looks ok. But I would have also been happy with just parsing `DENO_CACHED_ONLY=1` as `true` and erroring for all other cases :-)

ry: @piscisaureus has said in the past that we need to also support "yes", even for very internal env vars like `V8_FORCE_DEBUG`. 

https://github.com/denoland/rusty_v8/blob/845aced92cc1816dda248c882bfb1ec6c4dca81a/build.rs#L249-L254

@piscisaureus do we need to support "yes" here?

piscisaureus: > @piscisaureus Do you have an opinion?

I agree with "" and "0" and "false" meaning the flag is disabled.

bnoordhuis:
Adding one-off environment variables feels kind of meh. IMO, a better solution is a single `DENO_OPTIONS="--cached-only --etc"` environment variable.

bartlomieju:
@ry is this something you want to land? What about other flags that could potentially benefit from having an env var mimic?

cli/proc_state.rs
```diff
@@ -695,3 +696,21 @@ fn source_map_from_code(code: String) -> Option<Vec<u8>> {
     None
   }
 }
+
+fn binary_env_var(name: &str) -> bool {
+  if let Ok(v) = env::var(name) {
+    if v == "1" {
```
piscisaureus, line 702: You know my position on this, we should accept "true" and "false" for consistency.

This also works with `NO_COLOR` ([evidence](https://github.com/denoland/deno/blob/main/cli/tests/integration/test_tests.rs#L12)).

cli/proc_state.rs
```diff
@@ -695,3 +696,21 @@ fn source_map_from_code(code: String) -> Option<Vec<u8>> {
...
+      true
+    } else if v == "0" {
+      false
+    } else {
+      eprintln!(
+        "env var '{}' has a bad value '{}'. Use '1' or '0'.",
```
piscisaureus, line 708: +1 for showing feedback when the env var has an unsupported value.

PR #12771, opened 2021-11-15
feat(core): Deno.core.setNextTickCallback

### LINKED ISSUE #12735: Add "deno_core" bindings to polyfill "process.nextTick" from Node
This issue has been discussed with @bnoordhuis offline, here are notes:

Current polyfill for `process.nextTick` in `deno_std` looks like:
https://github.com/denoland/deno_std/blob/afa5c5150a852bcc711972c1068255c3873cf306/node/process.ts#L72-L89

This implementation is wrong, and manifested itself when trying to run Node tests in https://github.com/denoland/deno_std/issues/1482; a huge amount of internal code in Node depends on `nextTick` as well as huge amount of userland code, so it's critical to get it right to avoid hard to chase bugs.

There's no way we can polyfill `nextTick` in the userland and so we agreed that we need to add a new binding to `Deno.core` namespace, that will allow us to polyfill `process.nextTick` properly. 

Here's relevant code in code that runs `nextTick`:
https://github.com/nodejs/node/blob/6e1629786fe16058e0f12bc1613fae7734bff9b0/src/api/callback.cc#L118-L157
which is in turn called by:
https://github.com/nodejs/node/blob/6e1629786fe16058e0f12bc1613fae7734bff9b0/src/api/callback.cc#L186-L217

This API is very finicky and timing sensitive, so we expect it might take some back and forth before we get it exactly right. Additionally, one hard requirement for this binding is that it should be a no-op unless `process.nextTick` is registered. Or in other words: it shouldn't have performance penalty for regular Deno programs, only if program is run in "compat" mode or `deno_std/node/process.ts` is actually imported.

bartlomieju:
This commit adds several new "Deno.core" bindings:
- "setNextTickCallback"
- "hasScheduledTick"
- "setHasScheduledTick"
- "runMicrotasks"

Additionally it changes "Deno.core.setMacrotaskCallback" to
allow registering multiple callbacks. All these changes were necessary
to polyfill "process.nextTick" in Node compat layer.

Closes #12735


core/bindings.rs
```diff
@@ -38,6 +38,18 @@ lazy_static::lazy_static! {
...
+      },
+      v8::ExternalReference {
+        function: run_microtasks.map_fn_to()
+      },
+      v8::ExternalReference {
+        function: has_tick_scheduled.map_fn_to()
+      },
+      v8::ExternalReference {
+        function: set_has_tick_scheduled.map_fn_to()
+      },
```
bartlomieju, line 52: Not yet sure if we need all of these bindings

core/bindings.rs
```diff
@@ -438,6 +464,69 @@ fn opcall_async<'s>(
...
+
+  let cb = match v8::Local::<v8::Function>::try_from(args.get(0)) {
+    Ok(cb) => cb,
+    Err(err) => return throw_type_error(scope, err.to_string()),
+  };
+
+  // TODO(bartlomieju): this might be problematic if
+  // multiple versions of `deno_std` are pulled and polyfill
+  // `process.nextTick()`
+  let slot = match &mut state.js_nexttick_cb {
```
bartlomieju, line 517: Should be addressed before landing

core/bindings.rs
```diff
@@ -438,6 +464,69 @@ fn opcall_async<'s>(
...
+    slot @ None => slot,
+    _ => {
+      return throw_type_error(
+        scope,
+        "Deno.core.nextTickCallback() already called",
```
bartlomieju, line 522:
```suggestion
        "Deno.core.setNextTickCallback() already called",
```

core/runtime.rs
```diff
@@ -1586,6 +1591,38 @@ impl JsRuntime {
...
+      Some(handle) => handle.clone(),
+      None => return Ok(()),
+    };
+
+    if !state.borrow().has_tick_scheduled {
+      let scope = &mut self.handle_scope();
+      scope.perform_microtask_checkpoint();
+    }
+
+    // TODO(bartlomieju): Node also checks for absence of "rejection_to_warn"
```
bartlomieju, line 1608: Currently there's no handling of "unhandledRejection" in Deno so this conditional is skipped, I'm sure it will be problematic when trying to enable more of Node tests

bartlomieju: Side effect is that `process.nextTick()` doesn't behave properly if error is thrown during its callback

core/bindings.rs
```diff
@@ -438,7 +464,41 @@ fn opcall_async<'s>(
...
-fn set_macrotask_callback(
+fn has_tick_scheduled(
+  scope: &mut v8::HandleScope,
+  _args: v8::FunctionCallbackArguments,
+  mut rv: v8::ReturnValue,
+) {
+  let state_rc = JsRuntime::state(scope);
+  let state = state_rc.borrow();
+  rv.set(to_v8(scope, state.has_tick_scheduled).unwrap());
+}
```
bnoordhuis, line 475: Node does this by means of an ArrayBuffer, to avoid (slow) JS -> C++ calls. Not a blocker, merely an observation.

core/bindings.rs
```diff
@@ -438,7 +464,41 @@ fn opcall_async<'s>(
...
+) {
+  let state_rc = JsRuntime::state(scope);
+  let mut state = state_rc.borrow_mut();
+
+  let value = match v8::Local::<v8::Boolean>::try_from(args.get(0)) {
+    Ok(cb) => cb,
+    Err(err) => return throw_type_error(scope, err.to_string()),
+  };
+
+  state.has_tick_scheduled = value.is_true();
```
bnoordhuis, line 490:
```suggestion
  state.has_tick_scheduled = args.get(0).is_true();
```
Or is there a reason to do the conversion first?

core/runtime.rs
```diff
@@ -1549,34 +1554,73 @@ impl JsRuntime {
...
-        Some(handle) => handle.clone(),
-        None => return Ok(()),
-      };
+    let state = Self::state(self.v8_isolate());
+
+    if state.borrow().js_macrotask_cbs.is_empty() {
+      return Ok(());
+    }
 
+    let js_macrotask_cb_handles = state.borrow().js_macrotask_cbs.clone();
```
bnoordhuis, line 1563: This also clones all the v8::Global handles in the vec? That's a fairly expensive operation.

Also, I think I'm missing where `js_macrotask_cbs` is emptied. It looks like currently it only grows, never shrinks.

core/runtime.rs
```diff
@@ -1549,34 +1554,73 @@ impl JsRuntime {
...
+      let scope = &mut self.handle_scope();
+      scope.perform_microtask_checkpoint();
+    }
+
+    // TODO(bartlomieju): Node also checks for absence of "rejection_to_warn"
+    if !state.borrow().has_tick_scheduled {
+      return Ok(());
+    }
+
+    let js_nexttick_cb_handles = state.borrow().js_nexttick_cbs.clone();
```
bnoordhuis, line 1612: Non-blocking observation: `state.borrow()` gets called a lot. It's more efficient to borrow once and drop when necessary.

It's possible it doesn't matter in practice because the compiler "sees through" and elides borrow() calls.

core/bindings.rs
bartlomieju: Agreed, let's address in a follow up

core/bindings.rs
bartlomieju: Sure, that works

core/runtime.rs
bartlomieju: Yeah I believe it does clone global handles, in practice there should be no more than two of them, but I can try not to clone them at all (however it's risky because I'd have to hold onto borrow of state and there's nothing preventing callback from using state as well, ergo we might run into `BorrowError`).

> Also, I think I'm missing where js_macrotask_cbs is emptied. It looks like currently it only grows, never shrinks.

Correct, but until this PR there was only a single macrotask callback allowed (used by timers), this PR changes it to allow multiple callbacks. In practice there will be one callback for timers, one for process.nextTick, there might be more of them if users pull multiple versions of `deno_std` (and thus register multiple `process.nextTick` callback.

core/runtime.rs
bartlomieju: Please see comment above about risk of running into `BorrowError`

core/bindings.rs
bartlomieju: Actually I'm confused here - it seems to me that it's done by having a `TickInfo` bound from C++ to JS. Can you clarify what you meant?

bnoordhuis (approved):

core/bindings.rs
bnoordhuis: Node applies a trick where it makes a C struct available to JS land by using the struct's memory as a backing store for a typed array. That way C++ and JS land can communicate without expensive JS <-> C++ calls, they just read and write the struct's fields.

bartlomieju: I see, I'll be happy to explore this path, let's discuss it offline in details and I'll raise another PR

Issue #11613, opened 2021-08-09
[swc]: Using spread operator in JSX causes panic


kitsonk:
The following code will cause a panic when using `--no-check` (or trying to bundle), though it is valid JSX:

```js
const A = () => {
  return <div>{...[]}</div>;
};
```

Ref: https://github.com/swc-project/swc/issues/2037

crowlKats:
Tried this and works now, so this seems to have been fixed, so closing

PR #12773, opened 2021-11-15
fix(transpile): do not panic on `swc_ecma_utils::HANDLER` diagnostics

dsherret:
There is a hidden `swc_ecma_utils::HANDLER` API that we need to take into account (it's a thread local variable that we need to set a value for).

Relates to #11613, but I think ideally it should not error and instead it should handle this scenario.


kitsonk (approved):
LGTM

Issue #12753, opened 2021-11-13
lsp: Broken auto completion / suggestions


DreamTexX:
**Describe the bug**
It's a bit hard to describe the issue so I will append a video for better understanding.
I just moved from deno 1.15.3 to 1.16.1 and auto-completion is kinda broken. On code that I've written like a week ago, it shows auto-completion, but on code, I write today it doesn't come up with suggestions, they first appear when I explicitly start writing some method name or variable name. I use the same class to test, on old variables it shows completion on new instances of this class (new variables also) it doesn't suggest any methods (see the video appended).

**To Reproduce**

1. Install deno 1.16.1
2. Check auto completion on code that was already written
3. Create a new instance of a class or something
4. write `yourInstanceVariable.` and press the shortcut to show suggestions

**Expected behavior**

It should show class methods and variables in the suggestions.

**Screenshots**

https://user-images.githubusercontent.com/20863419/141652487-9b9a076f-4eab-4411-a9d9-1ab07ffc4ff1.mp4

**Versions**

vscode: 1.62.2 deno: 1.16.1 extension: 3.9.2

yacinehmito:
I am experiencing the same issue. Let me know how I can help.

yacinehmito:
@kitsonk I ran a `git bisect` and it told me the following commit introduced the regression: https://github.com/denoland/deno/commit/34a9ddff091950aee9d89915cd13944259e7d346

I'll continue to dig in.

kitsonk:
Thanks @yacinehmito for the repo/test. The issue has to do with use "losing" the dependencies when we hit a parse error when the module is being typed out. Effectively we need to preserve the dependencies between changes to the document when the parse is invalid.

yacinehmito:
Thank you very much.

PR #12776, opened 2021-11-16
tests(lsp): regression test for #12753

yacinehmito:
When making an edit that prevents a successful parsing of the file, the LSP loses type information on the identifiers whose type is declared outside the current file.
This is made apparent when trying to get completion on a method right after adding the dot, as the dot often introduced a syntax error prior to starting typing the method. If the object is of a type declared outside the current file, the LSP would give it the type `any` during that short span, making the completion of methods impossible.

This commits adds an ignored test that demonstrates the bug. It passes on commit 74a93fdf63a17de990954399b10eb6dfe7dd1973 but fails on commit 34a9ddff091950aee9d89915cd13944259e7d346.

### DIFF
--- a/cli/tests/integration/lsp_tests.rs
+++ b/cli/tests/integration/lsp_tests.rs
@@ -1066,6 +1066,136 @@ fn lsp_hover_dependency() {
   );
 }
 
+// Regression test for #12753
+#[test]
+#[ignore]
+fn lsp_hover_keep_type_info_after_invalid_syntax_change() {
+  let mut client = init("initialize_params.json");
+  did_open(
+    &mut client,
+    json!({
+      "textDocument": {
+        "uri": "file:///a/file1.ts",
+        "languageId": "typescript",
+        "version": 1,
+        "text": "export type Foo = { bar(): string };\n"
+      }
+    }),
+  );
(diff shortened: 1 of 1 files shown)


kitsonk:
Thanks for isolating this.

cli/tests/integration/lsp_tests.rs
```diff
@@ -1066,6 +1066,136 @@ fn lsp_hover_dependency() {
   );
 }
 
+// Regression test for #12753
+#[test]
+#[ignore]
+fn lsp_hover_keep_type_info_after_invalid_syntax_change() {
```
dsherret, line 1072: Maybe this is related to #12740?

kitsonk: no... because those code paths don't cross at all... I am pretty sure I know what it is, as hinted, when we have a bad parse of a module because of a syntax change, we drop the parsed module, meaning that we can't type check it anymore, therefore we can't get completions for it anymore, therefore the regression.

We will have to retain the parsed module when we get a parse error, so that we can do things like this.

kitsonk (approved):
LGTM

PR #12777, opened 2021-11-16
refactor: re-export anyhow from deno_core

ry:


dsherret (approved):
LGTM.

bartlomieju:
I find references return type of `Result<(), Error>` more ambiguous than `Result<(), AnyError>`... Is there any particular reason for this change?

ry:
@bartlomieju I think we should just use the anyhow crate directly rather than renaming - which introduces unnecessary indirection. Probably a better way to write it is `anyhow::Result<()>`.

bartlomieju (approved):
LGTM too

PR #12778, opened 2021-11-16
fix(lsp): reduce race conditions with line index

### LINKED ISSUE #12775: lsp: line index out of sync tsc information
I think the refactor to use deno_graph might have introduced a "racy" condition.  In particular, we put the line_index in an Arc to avoid cloning it, but it can mean that the line index is out of sync without tsc's view of things, as it often "lags" behind the rest of the language server.

I have noticed sometimes that tsc diagnostics are appearing on the "wrong" line as I am editing the file, and it takes editing the file "more slowly" to have it catch up.  As well as I got this error just now: `thread 'main' panicked at 'attempt to subtract with overflow', cli/lsp/tsc.rs:1148:9`

Which if you look, it uses the line index to figure out positions of things and then subtracts the start and end, assuming the token is on the same line, which it should be, but I think the pos in tsc and the line_index were out of sync, which cause the token to break across lines, which it shouldn't.

We might need to clone the index for tsc requests, and used the cloned index when generating the output.  We could just "discard" responses from tsc that are "outdated" but that doesn't feel right, as it is fine with tsc "catching up" with editing.

I am thinking that #12753 is potentially related as well.

cc/ @dsherret

kitsonk:
Fixes #12775

### DIFF
--- a/cli/lsp/code_lens.rs
+++ b/cli/lsp/code_lens.rs
@@ -25,7 +25,6 @@ use regex::Regex;
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::rc::Rc;
-use std::sync::Arc;
 
 lazy_static::lazy_static! {
   static ref ABSTRACT_MODIFIER: Regex = Regex::new(r"\babstract\b").unwrap();
@@ -222,7 +221,7 @@ async fn resolve_implementation_code_lens(
 ) -> Result<lsp::CodeLens, AnyError> {
   let asset_or_doc =
     language_server.get_cached_asset_or_document(&data.specifier)?;
-  let line_index = asset_or_doc.line_index();
+  let line_index = (*asset_or_doc.line_index()).clone();
   let req = tsc::RequestMethod::GetImplementation((
     data.specifier.clone(),
     line_index.offset_tsc(code_lens.range.start)?,
@@ -236,7 +235,7 @@ async fn resolve_implementation_code_lens(
       let implementation_specifier =
         resolve_url(&implementation.document_span.file_name)?;
       let implementation_location =
-        implementation.to_location(line_index.clone(), language_server);
+        implementation.to_location(&line_index, language_server);
       if !(implementation_specifier == data.specifier
         && implementation_location.range.start == code_lens.range.start)
       {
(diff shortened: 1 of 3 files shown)


cli/lsp/code_lens.rs
```diff
@@ -222,7 +221,7 @@ async fn resolve_implementation_code_lens(
 ) -> Result<lsp::CodeLens, AnyError> {
   let asset_or_doc =
     language_server.get_cached_asset_or_document(&data.specifier)?;
-  let line_index = asset_or_doc.line_index();
+  let line_index = (*asset_or_doc.line_index()).clone();
```
dsherret, line 224: The `line_index` should be immutable in the `Arc`, so why do we need to clone the contents inside the `Arc`?

cli/lsp/language_server.rs
```diff
@@ -985,28 +985,28 @@ impl Inner {
...
-    let req = tsc::RequestMethod::GetNavigationTree(specifier);
-    let navigation_tree: tsc::NavigationTree = self
-      .ts_server
-      .request(self.snapshot()?, req)
-      .await
-      .map_err(|err| {
-        error!("Failed to request to tsserver {}", err);
-        LspError::invalid_request()
+    // we need to clone the line index _before_ the async process of grabbing
+    // the nav tree, as the line index could change
```
dsherret, line 989: I think there are some parts of the LSP that I don't understand. How does the line index change while getting the navigation tree?

cli/lsp/code_lens.rs
kitsonk: oh yeah, you got a point there... hmmmm... it must be related to situations where we dynamically get the line index _after_ the result

cli/lsp/language_server.rs
kitsonk: I think you are right... since we don't mutate the `Arc`... I need to re-think this... I did find one spot where we were getting the line index _after_ the call to tsc, which could be the issue, instead of holding onto the reference during it.

It is a real challenge to reproduce though, because it is a race condition, but I have seen some really strange behaviour on 1.16 that is consistent with the line_index being "wrong" that didn't exist before.

kitsonk:
I think I was barking up the wrong tree...

cli/lsp/language_server.rs
dsherret: Yeah, let's maybe tackle fixing #12776 and we can look more to try to get a reproduction of this issue.

PR #12782, opened 2021-11-16
fix(lsp): retain module dependencies when parse is invalid

### LINKED ISSUE #12753: lsp: Broken auto completion / suggestions

kitsonk:
Fixes #12753

### DIFF
--- a/cli/lsp/documents.rs
+++ b/cli/lsp/documents.rs
@@ -245,6 +245,8 @@ impl SyntheticModule {
 }
 #[derive(Debug, Clone)]
 struct DocumentInner {
+  /// contains the last-known-good set of dependencies from parsing the module
+  dependencies: Arc<BTreeMap<String, deno_graph::Dependency>>,
   fs_version: String,
   line_index: Arc<LineIndex>,
   maybe_language_id: Option<LanguageId>,
@@ -282,9 +284,15 @@ impl Document {
       maybe_resolver,
       Some(&parser),
     ));
+    let dependencies = if let Some(Ok(module)) = &maybe_module {
+      Arc::new(module.dependencies.clone())
+    } else {
+      Arc::new(BTreeMap::new())
+    };
     let text_info = SourceTextInfo::new(content);
     let line_index = Arc::new(LineIndex::new(text_info.text_str()));
     Self(Arc::new(DocumentInner {
+      dependencies,
       fs_version,
       line_index,
       maybe_language_id: None,
@@ -317,9 +325,15 @@ impl Document {
     } else {
       None
     };
+    let dependencies = if let Some(Ok(module)) = &maybe_module {
+      Arc::new(module.dependencies.clone())
+    } else {
+      Arc::new(BTreeMap::new())
+    };
     let source = SourceTextInfo::new(content);
     let line_index = Arc::new(LineIndex::new(source.text_str()));
     Self(Arc::new(DocumentInner {
+      dependencies,
       fs_version: "1".to_string(),
       line_index,
       maybe_language_id: Some(language_id),
@@ -379,14 +393,20 @@ impl Document {
     } else {
       None
     };
-    let source = SourceTextInfo::new(content);
+    let dependencies = if let Some(Ok(module)) = &maybe_module {
+      Arc::new(module.dependencies.clone())
+    } else {
+      self.0.dependencies.clone()
+    };
+    let text_info = SourceTextInfo::new(content);
     let line_index = if index_valid == IndexValid::All {
       line_index
     } else {
-      Arc::new(LineIndex::new(source.text_str()))
+      Arc::new(LineIndex::new(text_info.text_str()))
     };
     Ok(Document(Arc::new(DocumentInner {
-      text_info: source,
+      dependencies,
+      text_info,
       line_index,
       maybe_module,
       maybe_lsp_version: Some(version),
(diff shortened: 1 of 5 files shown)


cli/lsp/language_server.rs
```diff
@@ -987,17 +987,16 @@ impl Inner {
...
 
-    let req = tsc::RequestMethod::GetNavigationTree(specifier);
-    let navigation_tree: tsc::NavigationTree = self
-      .ts_server
-      .request(self.snapshot()?, req)
-      .await
-      .map_err(|err| {
-        error!("Failed to request to tsserver {}", err);
-        LspError::invalid_request()
+    let navigation_tree =
```
kitsonk, line 990: Not directly related to fixing the issue, discovered that we were not using the potentially cached version of the navtree here, which is not good.

cli/lsp/language_server.rs
```diff
@@ -1563,9 +1562,14 @@ impl Inner {
         }
         let reference_specifier =
           resolve_url(&reference.document_span.file_name).unwrap();
-        let asset_or_doc =
-          self.get_asset_or_document(&reference_specifier).await?;
-        results.push(reference.to_location(asset_or_doc.line_index(), self));
+        let reference_line_index = if reference_specifier == specifier {
```
kitsonk, line 1565: This also fixes a potential race condition where we re-use the line_index we first retrieved if the specifier is the same (which it is a lot of the time with this) instead of grabbing a new line_index.

cli/lsp/tsc.rs
```diff
@@ -1141,14 +1141,23 @@ impl Classifications {
...
 
-      // start_pos.line == end_pos.line is always true as there are no multiline tokens
-      builder.push(
-        start_pos.line,
-        start_pos.character,
-        end_pos.character - start_pos.character,
-        token_type,
-        token_modifiers,
-      );
+      if start_pos.line == end_pos.line
```
kitsonk, line 1144: This directly deals with eliminating a panic that shouldn't occur, but did for me.  I haven't been able to reproduce it, but at least we will get a more useful error logged to the debug output versus a panic and a crash of the language server.

cli/lsp/documents.rs
```diff
@@ -379,14 +393,20 @@ impl Document {
     } else {
       None
     };
-    let source = SourceTextInfo::new(content);
+    let dependencies = if let Some(Ok(module)) = &maybe_module {
+      Arc::new(module.dependencies.clone())
+    } else {
+      self.0.dependencies.clone()
```
kitsonk, line 399: This is what fixes the issue, we carry over the dependencies when there is an invalid parse.  This allows completions/suggestions/etc to still be processed by tsc when the module contains an unparsable syntax error.

cli/lsp/tsc.rs
dsherret: Seems risky to just log here because maybe it could cause some indirect bugs. It might be better if we panic, but with the information you've added because then we'll get bug reports to fix the issue and we'll know this is occurring.

cli/lsp/documents.rs
```diff
@@ -911,35 +929,32 @@ impl DocumentsInner {
     specifiers: Vec<String>,
     referrer: &ModuleSpecifier,
   ) -> Option<Vec<Option<(ModuleSpecifier, MediaType)>>> {
-    let doc = self.get(referrer)?;
+    let dependencies = self.get(referrer)?.0.dependencies.clone();
```
dsherret, line 932: Nitpick: Can `doc` still be stored then `doc.0.dependencies` be used below instead to not clone? I think it should be ok because `Document` is immutable in an `Arc` (so should also be no borrow checker trouble). I see previously it was cloning and I think it's just something I missed in my refactor putting `Document` (`DocumentInner`) into an `Arc`.

dsherret (approved):
LGTM! I think we should still panic in `to_semantic_tokens` though (improve the panic message instead)

cli/lsp/documents.rs
kitsonk: No, because it is a double borrow of self.

dsherret: Oh yeah, I forgot the return type is `&Document`. Could do `let doc = self.get(referrer)?.clone()` which is very cheap then.

Anyway, not a big deal.

kitsonk: Dependencies in the document is an Arc now too.

cli/lsp/tsc.rs
kitsonk: I don't think it can cause indirect bugs.  What is occurring here is a returned value from tsc is being "remapped" to something that can be used by a language server client.  If it errors, a "bad" re-map is simply elided, which means a highlighted semantic symbol doesn't get returned to the client.  That is all.

Panicing here causes the language server to terminate, and be restarted which is far more "destructive".  Generally we should avoid panicing in the language server, as no good comes from it.

An improvement is I can return a result here as an internal error.

cli/lsp/language_server.rs
kitsonk: I think either way, there is no harm in re-using an already available value versus grabbing a new one.

Issue #12458, opened 2021-10-16
Per-`JsRuntime` maps in `ProcState`


andreubotella:
Module maps are conceptually specific to an instance of `JsRuntime`, yet part of the module map mechanics in our current implementation is maintained in `ProcState`, which is shared across multiple `JsRuntime`s. An implementation of #12248 would need `ProcState` to keep a per-`JsRuntime` url → blob map, and according to @bartlomieju, something similar would be needed to cache resolved modules.

One issue is how to represent `JsRuntime` identity in those maps. `JsRuntime` is not cloneable or a reference type, and it is not typically used with a reference-counted pointer type. What's more, if a `JsRuntime` is dropped ­–if a worker is terminated, for example– the blob references and cached modules associated with that `JsRuntime` should be dropped as well, but there doesn't seem to be any obvious way to do this, since `JsRuntime` has no concept of `ProcState`.

bartlomieju:
This is no longer relevant

PR #12785, opened 2021-11-16
[WIP] refactor: store GraphData per-worker instead of in ProcState

bartlomieju:
Closes https://github.com/denoland/deno/issues/12458

This is just a PoC and still needs more work.


bartlomieju:
Not clear how to trigger the bug. Closing for now, will revisit later.

Issue #11479, opened 2021-07-21
Feature request: Support ALPN in `Deno.connectTls`


jaclarke:
There was a previous issue to add ALPN support to the TLS api's (https://github.com/denoland/deno/issues/6334), which was closed by this PR: https://github.com/denoland/deno/pull/10065. However it appears that only support for ALPN in `Deno.listenTls` was added, is it on the roadmap to also add support for ALPN in the `Deno.connectTls` api as well?

Use case: We have a Deno driver for EdgeDB (https://github.com/edgedb/edgedb-deno), and the next release of EdgeDB will switch to using TLS by default (https://github.com/edgedb/rfcs/blob/master/text/1008-tls-and-alpn.rst), which requires support for ALPN in the `Deno.connectTls` api.

PR #12786, opened 2021-11-16
feat: ALPN support in `Deno.connectTls()`

1st1:
* Make `connectTls()` accept `alpnProtocols` option
  (string[]), just like `listenTls()`.

* Add `alpnProtocol` to `TlsConn#handshake` return
  value to query the protocol agreed with the peer via ALPN.

The `alpnProtocols` is added as an unstable API (mirroring
how it's supported in `listenTls()`).

We need this change for making a Deno client for EdgeDB. EdgeDB
requires TLS by default and uses ALPN to upgrade connections to
its binary protocol. Without these APIs we can't support Deno,
something we really want to have.

Fixes: https://github.com/denoland/deno/issues/11479

### DIFF
--- a/ext/net/02_tls.js
+++ b/ext/net/02_tls.js
@@ -27,15 +27,7 @@
     return core.opAsync("op_tls_handshake", rid);
   }
 
-  function opTlsGetAlpnProtocol(rid) {
-    return core.opAsync("op_tls_get_alpn_protocol", rid);
-  }
-
   class TlsConn extends Conn {
-    getAgreedAlpnProtocol() {
-      return opTlsGetAlpnProtocol(this.rid);
-    }
-
     handshake() {
       return opTlsHandshake(this.rid);
     }
@@ -49,7 +41,7 @@
     caCerts = [],
     certChain = undefined,
     privateKey = undefined,
-    alpnProtocols,
+    alpnProtocols = undefined,
   }) {
     const res = await opConnectTls({
       port,
@@ -77,7 +69,7 @@
     keyFile,
     hostname = "0.0.0.0",
     transport = "tcp",
-    alpnProtocols,
+    alpnProtocols = undefined,
   }) {
     const res = opListenTls({
       port,
@@ -92,13 +84,19 @@
 
   async function startTls(
     conn,
-    { hostname = "127.0.0.1", certFile = undefined, caCerts = [] } = {},
+    {
+      hostname = "127.0.0.1",
+      certFile = undefined,
+      caCerts = [],
+      alpnProtocols = undefined,
+    } = {},
   ) {
     const res = await opStartTls({
       rid: conn.rid,
       hostname,
       certFile,
       caCerts,
+      alpnProtocols,
     });
     return new TlsConn(res.rid, res.remoteAddr, res.localAddr);
   }
--- a/ext/net/02_tls.js
+++ b/ext/net/02_tls.js
@@ -27,7 +27,15 @@
     return core.opAsync("op_tls_handshake", rid);
   }
 
+  function opTlsGetAlpnProtocol(rid) {
+    return core.opAsync("op_tls_get_alpn_protocol", rid);
+  }
+
   class TlsConn extends Conn {
+    getAgreedAlpnProtocol() {
+      return opTlsGetAlpnProtocol(this.rid);
+    }
+
     handshake() {
       return opTlsHandshake(this.rid);
     }
@@ -41,6 +49,7 @@
     caCerts = [],
     certChain = undefined,
     privateKey = undefined,
+    alpnProtocols,
   }) {
     const res = await opConnectTls({
       port,
@@ -50,6 +59,7 @@
       caCerts,
       certChain,
       privateKey,
+      alpnProtocols,
     });
     return new TlsConn(res.rid, res.remoteAddr, res.localAddr);
   }
--- a/ext/net/ops_tls.rs
+++ b/ext/net/ops_tls.rs
@@ -190,6 +190,13 @@ impl TlsStream {
   fn poll_handshake(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
     self.inner_mut().poll_handshake(cx)
   }
+
+  fn get_alpn_protocol(&mut self) -> Option<String> {
+    match self.inner_mut().tls.get_alpn_protocol() {
+      None => None,
+      Some(s) => Some(std::str::from_utf8(s).unwrap().to_string())
+    }
+  }
 }
 
 impl AsyncRead for TlsStream {
@@ -517,6 +524,10 @@ impl ReadHalf {
       .tls_stream
       .into_inner()
   }
+
+  fn get_alpn_protocol(&mut self) -> Option<String> {
+    self.shared.get_alpn_protocol()
+  }
 }
 
 impl AsyncRead for ReadHalf {
@@ -658,6 +669,11 @@ impl Shared {
   fn drop_shared_waker(self_ptr: *const ()) {
     let _ = unsafe { Weak::from_raw(self_ptr as *const Self) };
   }
+
+  fn get_alpn_protocol(self: &Arc<Self>) -> Option<String> {
+    let mut tls_stream = self.tls_stream.lock();
+    return tls_stream.get_alpn_protocol();
+  }
 }
 
 struct ImplementReadTrait<'a, T>(&'a mut T);
@@ -691,6 +707,7 @@ pub fn init<P: NetPermissions + 'static>() -> Vec<OpPair> {
     ("op_tls_listen", op_sync(op_tls_listen::<P>)),
     ("op_tls_accept", op_async(op_tls_accept)),
     ("op_tls_handshake", op_async(op_tls_handshake)),
+    ("op_tls_get_alpn_protocol", op_async(op_tls_get_alpn_protocol)),
   ]
 }
 
@@ -753,6 +770,13 @@ impl TlsStreamResource {
     }
     Ok(())
   }
+
+  pub async fn get_alpn_protocol(self: &Rc<Self>) ->
+    Result<Option<String>, AnyError>
+  {
+    let mut rd = RcRef::map(self, |r| &r.rd).borrow_mut().await;
+    Ok(rd.get_alpn_protocol())
+  }
 }
 
 impl Resource for TlsStreamResource {
@@ -787,6 +811,7 @@ pub struct ConnectTlsArgs {
   ca_certs: Vec<String>,
   cert_chain: Option<String>,
   private_key: Option<String>,
+  alpn_protocols: Option<Vec<String>>,
 }
 
 #[derive(Deserialize)]
@@ -947,6 +972,12 @@ where
     ca_certs,
     unsafely_ignore_certificate_errors,
   )?;
+
+  if let Some(alpn_protocols) = args.alpn_protocols {
+    super::check_unstable2(&state, "Deno.connectTls#alpnProtocols");
+    tls_config.alpn_protocols =
+      alpn_protocols.into_iter().map(|s| s.into_bytes()).collect();
+  }
 
   if args.cert_chain.is_some() || args.private_key.is_some() {
     let cert_chain = args
@@ -1150,4 +1181,16 @@ pub async fn op_tls_handshake(
     .resource_table
     .get::<TlsStreamResource>(rid)?;
   resource.handshake().await
+}
+
+pub async fn op_tls_get_alpn_protocol(
+  state: Rc<RefCell<OpState>>,
+  rid: ResourceId,
+  _: (),
+) -> Result<Option<String>, AnyError> {
+  let resource = state
+    .borrow()
+    .resource_table
+    .get::<TlsStreamResource>(rid)?;
+  resource.get_alpn_protocol().await
 }
--- a/ext/net/ops_tls.rs
+++ b/ext/net/ops_tls.rs
@@ -527,10 +527,6 @@ impl ReadHalf {
       .tls_stream
       .into_inner()
   }
-
-  fn get_alpn_protocol(&mut self) -> Option<ByteString> {
-    self.shared.get_alpn_protocol()
-  }
 }
 
 impl AsyncRead for ReadHalf {
@@ -562,6 +558,10 @@ impl WriteHalf {
         })
     })
     .await
+  }
+
+  fn get_alpn_protocol(&mut self) -> Option<ByteString> {
+    self.shared.get_alpn_protocol()
   }
 }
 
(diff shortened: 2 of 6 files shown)


ext/net/02_tls.js
```diff
@@ -27,7 +27,15 @@
...
 
+  function opTlsGetAlpnProtocol(rid) {
+    return core.opAsync("op_tls_get_alpn_protocol", rid);
+  }
+
   class TlsConn extends Conn {
+    getAgreedAlpnProtocol() {
+      return opTlsGetAlpnProtocol(this.rid);
+    }
+
```
lucacasonato, line 38:
```suggestion
    getAgreedAlpnProtocol() {
      return opTlsGetAlpnProtocol(this.rid);
    }

```

I feel like this is something that could be returned from `handshake()`

ext/net/ops_tls.rs
```diff
@@ -190,6 +190,13 @@ impl TlsStream {
   fn poll_handshake(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
     self.inner_mut().poll_handshake(cx)
   }
+
+  fn get_alpn_protocol(&mut self) -> Option<String> {
+    match self.inner_mut().tls.get_alpn_protocol() {
+      None => None,
+      Some(s) => Some(std::str::from_utf8(s).unwrap().to_string())
```
lucacasonato, line 197: The unwrap here is not good. It is a vector for a denial of service. The string should be returned as a `ByteString` (this is a string where each char is some value between 0-255) - not a utf8 string.

ext/net/02_tls.js
lucacasonato: To clarify, I imagine `handshake()` returning a `Promise<TlsHandshakeInfo>` where `TlsHandshakeInfo` is the following:

```
interface TlsHandshakeInfo {
  alpnProtocol: string | null;
}
```

lucacasonato (commented):
Thanks for the work on this @1st1! I have added some comments about the API surface.

We'll also need some tests for this feature (in `cli/tests/unit/tls_test.ts`). For test cases you could create a server with `Deno.listenTls` with some ALPN protocols and then connect to it during the test and check that the negotiation works correctly. There are some examples on how to do that in that file.

Also, this option should be added to `Deno.startTls` in addition to `Deno.connectTls`.

ext/net/02_tls.js
1st1: Great idea. Done.

ext/net/ops_tls.rs
1st1: Right. I've updated the PR to use ByteString.

1st1:
@lucacasonato Thanks for the review! I've updated the PR implementing all of your suggestions.

ext/net/ops_tls.rs
```diff
@@ -1144,10 +1187,15 @@ pub async fn op_tls_handshake(
...
   _: (),
-) -> Result<(), AnyError> {
+) -> Result<TlsHandshakeInfo, AnyError> {
   let resource = state
     .borrow()
     .resource_table
     .get::<TlsStreamResource>(rid)?;
-  resource.handshake().await
+
+  resource.handshake().await?;
```
lucacasonato, line 1196: Could you change `resource.handshake()` to return the `TlsHandshakeInfo` directly?

lucacasonato (commented):
Test looks great, thanks.

ext/net/ops_tls.rs
1st1: Sure, done. I've pushed an update.

1st1:
I think the CI timeouts aren't related to this PR. @lucacasonato let me know if there's anything else left to do.

1st1:
> I think the CI timeouts aren't related to this PR. 

Actually they are. Investigating.

piscisaureus:
> `` [Fix tests hang; move `get_alpn_protocol()` helper to WriteHalf](/denoland/deno/pull/12786/commits/b00251d26876a9eee6a684ad8f6b52d532f0a1ef) ``

I'm not sure that is the right fix. The same hang could happen, it just isn't triggered by our tests?

lucacasonato (commented):
Maybe turn `handshake_done` into a `Cell<Option<TlsHandshakeInfo>>` and store the handshake info in there right after handshaking?

piscisaureus:
> Maybe turn `handshake_done` into a `Cell<Option<TlsHandshakeInfo>>` and store the handshake info in there right after handshaking?

Just looked into it, that seems to be the right way to go about it indeed.

1st1:
> I'm not sure that is the right fix. The same hang could happen, it just isn't triggered by our tests?

My thinking was that `handshake()` already acquires the `WriteHalf` so it should be safer. But yeah, it's still not 100% safe.

> Maybe turn handshake_done into a Cell<Option<TlsHandshakeInfo>> and store the handshake info in there right after handshaking?

This is what I also settled on, although it seems that I can't wrap `Option<TlsHandshakeInfo>` into a `Cell`, because `TlsHandshakeInfo` must implement `Copy`, and so transitively, `ByteString` must also be copyable which it can't be:

> 720 |   handshake_done: Cell<Option<TlsHandshakeInfo>>,
>    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `TlsHandshakeInfo`

So I'm using `RefCell` which is a bit verbose but seems to do the job. Please take a look at the latest push.

piscisaureus:
> This is what I also settled on, although it seems that I can't wrap `Option<TlsHandshakeInfo>` into a `Cell`, because `TlsHandshakeInfo` must implement `Copy`, and so transitively, `ByteString` must also be copyable which it can't be:
> 
> >

`Copy` is only required for `Cell::get()` though. Presumably `TlsHandshakeInfo` could implement `Clone` so you could replace `handshake_done.get()` with `handshake_done.clone().take()`.

1st1:
> Presumably TlsHandshakeInfo could implement Clone so you could replace handshake_done.get() with handshake_done.clone().take().

Yes, I've made `TlsHandshakeInfo` implement `Clone` in the latest commit. I'm trying your suggestion to use `Cell` but the compiler still isn't happy:

<details>
<summary>Compile error:</summary>
```
error[E0599]: the method `clone` exists for struct `Cell<std::option::Option<TlsHandshakeInfo>>`, but its trait bounds were not satisfied
   --> ext/net/ops_tls.rs:769:49
    |
769 |     if let Some(tls_info) = self.handshake_done.clone().take() {
    |                                                 ^^^^^ method cannot be called on `Cell<std::option::Option<TlsHandshakeInfo>>` due to unsatisfied trait bounds
    |
   ::: /usr/lib/rust/1.56.1/lib/rustlib/src/rust/library/core/src/cell.rs:236:1
    |
236 | pub struct Cell<T: ?Sized> {
    | -------------------------- doesn't satisfy `Cell<std::option::Option<TlsHandshakeInfo>>: Clone`
    |
   ::: /usr/lib/rust/1.56.1/lib/rustlib/src/rust/library/core/src/option.rs:510:1
    |
510 | pub enum Option<T> {
    | ------------------ doesn't satisfy `_: std::marker::Copy`
    |
    = note: the following trait bounds were not satisfied:
            `std::option::Option<TlsHandshakeInfo>: std::marker::Copy`
            which is required by `Cell<std::option::Option<TlsHandshakeInfo>>: Clone`

error[E0277]: the trait bound `TlsHandshakeInfo: std::marker::Copy` is not satisfied
   --> ext/net/ops_tls.rs:720:3
    |
716 | #[derive(Debug)]
    |          ----- in this derive macro expansion
...
720 |   handshake_done: Cell<Option<TlsHandshakeInfo>>,
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `TlsHandshakeInfo`
    |
    = note: required because of the requirements on the impl of `std::marker::Copy` for `std::option::Option<TlsHandshakeInfo>`
    = note: required because of the requirements on the impl of `std::fmt::Debug` for `Cell<std::option::Option<TlsHandshakeInfo>>`
    = note: 1 redundant requirements hidden
    = note: required because of the requirements on the impl of `std::fmt::Debug` for `&Cell<std::option::Option<TlsHandshakeInfo>>`
    = note: required for the cast to the object type `dyn std::fmt::Debug`
    = note: this error originates in the derive macro `Debug` (in Nightly builds, run with -Z macro-backtrace for more info)
```
</details>

piscisaureus:
> Yes, I've made `TlsHandshakeInfo` implement `Clone` in the latest commit. I'm trying your suggestion to use `Cell` but the compiler still isn't happy and wants `TlsHandshakeInfo` to implement `Copy`:

Ok let me look into it.

1st1:
I can get around not using `Cell::set` entirely like this:

<details>
<summary>See the diff</summary>

```diff
diff --git a/ext/net/ops_tls.rs b/ext/net/ops_tls.rs
index 2edbfc08c..d98247c9d 100644
--- a/ext/net/ops_tls.rs
+++ b/ext/net/ops_tls.rs
@@ -56,6 +56,7 @@ use io::Read;
 use io::Write;
 use serde::Deserialize;
 use std::borrow::Cow;
+use std::cell::Cell;
 use std::cell::RefCell;
 use std::convert::From;
 use std::fs::File;
@@ -716,7 +717,7 @@ pub fn init<P: NetPermissions + 'static>() -> Vec<OpPair> {
 pub struct TlsStreamResource {
   rd: AsyncRefCell<ReadHalf>,
   wr: AsyncRefCell<WriteHalf>,
-  handshake_done: RefCell<Option<TlsHandshakeInfo>>,
+  handshake_done: Cell<Option<TlsHandshakeInfo>>,
   cancel_handle: CancelHandle, // Only read and handshake ops get canceled.
 }
 
@@ -725,7 +726,7 @@ impl TlsStreamResource {
     Self {
       rd: rd.into(),
       wr: wr.into(),
-      handshake_done: RefCell::new(None),
+      handshake_done: Cell::new(None),
       cancel_handle: Default::default(),
     }
   }
@@ -765,8 +766,10 @@ impl TlsStreamResource {
   pub async fn handshake(
     self: &Rc<Self>,
   ) -> Result<TlsHandshakeInfo, AnyError> {
-    if let Some(tls_info) = &*self.handshake_done.borrow() {
-      return Ok(tls_info.clone());
+    let hs_result = self.handshake_done.replace(None);
+    if let Some(tls_info) = hs_result {
+      self.handshake_done.set(Some(tls_info.clone()));
+      return Ok(tls_info);
     }
 
     let mut wr = RcRef::map(self, |r| &r.wr).borrow_mut().await;
@@ -775,7 +778,7 @@ impl TlsStreamResource {
 
     let alpn_protocol = wr.get_alpn_protocol();
     let tls_info = TlsHandshakeInfo { alpn_protocol };
-    self.handshake_done.replace(Some(tls_info.clone()));
+    self.handshake_done.set(Some(tls_info.clone()));
     Ok(tls_info)
   }
 }
```
</details>

but the compiler is still not happy with declaring `handshake_done: Cell<Option<TlsHandshakeInfo>>`:

<details>
<summary>Error:</summary>

```
~/d/deno (alpn *) » cargo test
    Blocking waiting for file lock on package cache
   Compiling deno_net v0.17.0 (/home/yury/dev/deno/ext/net)
error[E0277]: the trait bound `TlsHandshakeInfo: std::marker::Copy` is not satisfied
   --> ext/net/ops_tls.rs:720:3
    |
716 | #[derive(Debug)]
    |          ----- in this derive macro expansion
...
720 |   handshake_done: Cell<Option<TlsHandshakeInfo>>,
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `TlsHandshakeInfo`
    |
    = note: required because of the requirements on the impl of `std::marker::Copy` for `std::option::Option<TlsHandshakeInfo>`
    = note: required because of the requirements on the impl of `std::fmt::Debug` for `Cell<std::option::Option<TlsHandshakeInfo>>`
    = note: 1 redundant requirements hidden
    = note: required because of the requirements on the impl of `std::fmt::Debug` for `&Cell<std::option::Option<TlsHandshakeInfo>>`
    = note: required for the cast to the object type `dyn std::fmt::Debug`
    = note: this error originates in the derive macro `Debug` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `deno_net` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```
</details>

1st1:
> Ok let me look into it.

Sure, really appreciate this! Is there an issue with swapping `Cell` with `RefCell` though? That seems to work in the latest commit.

piscisaureus:
> Sure, really appreciate this! Is there an issue with swapping `Cell` with `RefCell` though? That seems to work in the latest commit.

Not really, I was just being pedantic.

piscisaureus:
Final nit, maybe we can rename `handshake_done` to `handshake_info` and add a comment that the value will be `None` if the handshake isn't done yet.

Other than that, this looks good to go.

1st1:
> Final nit, maybe we can rename handshake_done to handshake_info and add a comment that the value will be None if the handshake isn't done yet.

Sure, done.

lucacasonato (commented):
LGTM @1st1. Excellent work on your first contribution! :tada: 

I can't land the PR right away, because the 1.17 feature merge window has not opened yet. It will open after the 1.16.3 release next Tuesday (23 November 2021).

lucacasonato (approved):

1st1:
Thanks @lucacasonato. Feel free to ping me if you need help with rebasing when the merge window is open.

piscisaureus (approved):

1st1:
A friendly ping: let's merge this? Just want to make sure we get this into the next release.

lucacasonato:
@1st1 Could you rebase the PR and apply the following patch:

```patch
From b933c778addf96d816f003a58816ec2cc1fcae5e Mon Sep 17 00:00:00 2001
From: Luca Casonato <hello@lcas.dev>
Date: Fri, 26 Nov 2021 10:14:13 +0100
Subject: [PATCH] update tests

---
 cli/tests/unit/tls_test.ts | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/cli/tests/unit/tls_test.ts b/cli/tests/unit/tls_test.ts
index f47c6ac90..7e6d68900 100644
--- a/cli/tests/unit/tls_test.ts
+++ b/cli/tests/unit/tls_test.ts
@@ -364,7 +364,7 @@ Deno.test(
   },
 );
 
-unitTest(
+Deno.test(
   { permissions: { read: true, net: true } },
   async function tlsServerAlpnListenStartTls() {
     const [serverConn, clientConn] = await tlsAlpn(true);
@@ -380,7 +380,7 @@ unitTest(
   },
 );
 
-unitTest(
+Deno.test(
   { permissions: { read: true, net: true } },
   async function tlsServerStreamHalfCloseSendOneByte() {
     const [serverConn, clientConn] = await tlsPair();
-- 
2.34.0

```

I'd do it myself, but I don't have write access to your fork :-)

1st1:
> @1st1 Could you rebase the PR and apply the following patch:

Done!

> I'd do it myself, but I don't have write access to your fork :-)

Hm, I don't see a checkbox to grant you push privs... I'll look for it, but in the meantime if there's anything else you want to be fixed in the PR let me know.

PR #12787, opened 2021-11-17
feat(lsp): add workspace symbol provider

kitsonk:
Ref: #8643

This also does some other minor cleanups and refactors in the lsp.


dsherret (approved):
LGTM

PR #12788, opened 2021-11-17
feat(ext/cache): partial service-worker cache

AaronO:
First pass at implementing partial(match/put/delete) SW Cache support

Ships with a naive in-memory CacheEngine which should be swapped out for an LRU or another store can be customized by embedders

It is not batteries-included, cache-control matching etc... is left as an exercise to the engine providers ...

### DIFF
--- a/ext/cache/01_swcache.js
+++ b/ext/cache/01_swcache.js
@@ -1,0 +1,142 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+// @ts-check
+/// <reference path="../../core/internal.d.ts" />
+/// <reference path="../fetch/lib.deno_fetch.d.ts" />
+
+"use strict";
+
+((window) => {
+  const {
+    ArrayPrototypeJoin,
+    Map,
+    MapPrototypeDelete,
+    MapPrototypeGet,
+    MapPrototypeSet,
+    ObjectFromEntries,
+    PromiseResolve,
+  } = window.__bootstrap.primordials;
+  const { Request, Response } = window.__bootstrap.fetch;
+
+  class Caches {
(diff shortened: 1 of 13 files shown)


ext/cache/01_swcache.js
```diff
@@ -0,0 +1,142 @@
...
+    MapPrototypeGet,
+    MapPrototypeSet,
+    ObjectFromEntries,
+    PromiseResolve,
+  } = window.__bootstrap.primordials;
+  const { Request, Response } = window.__bootstrap.fetch;
+
+  class Caches {
+    constructor(engine) {
+      this.engine = engine;
```
lucacasonato, line 23: This leaks the engine to the user. Please look how we implement other Web APIs.

lucacasonato (changes requested):
This implementation does not really follow spec, or how other web APIs are implemented in Deno. This also punts off too much spec logic to the implementor.

The implementation should be generic over the blob store, not some more abstract cache engine. This crate should handle all of the cache logic.

ext/cache/01_swcache.js
```diff
@@ -0,0 +1,142 @@
...
+  function toKey(reqOrUrl) {
+    return reqToKey(asReq(reqOrUrl));
+  }
+
+  function asReq(reqOrUrl) {
+    return reqOrUrl instanceof Request ? reqOrUrl : new Request(reqOrUrl);
+  }
+
+  function reqToKey(req) {
+    return ArrayPrototypeJoin([req.url, req.method], ":");
```
aapoalas, line 69: Is there a reason to use `array.join(":")` instead of

```js
return `${req.url}:${req.method}`;
```
?

PR #12793, opened 2021-11-17
feat(ext/crypto): support md5 in subtle.digest()

### LINKED ISSUE #12715: Support MD5 in webcrypto
like cloudflare https://developers.cloudflare.com/workers/runtime-apis/web-crypto

bnoordhuis:
Non-spec, but compatible with CF Workers.

Fixes #12715.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -766,6 +766,7 @@ dependencies = [
  "deno_web",
  "elliptic-curve",
  "lazy_static",
+ "md5",
  "num-traits",
  "p256",
  "p384",
@@ -2177,6 +2178,12 @@ name = "matches"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3e378b66a060d48947b590737b30a1be76706c8dd7b8ba0f2fe3989c68a853f"
+
+[[package]]
+name = "md5"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "490cc448043f947bae3cbee9c203358d62dbee0db12107a74be5c30ccfd09771"
 
 [[package]]
 name = "memchr"
--- a/cli/tests/unit/webcrypto_test.ts
+++ b/cli/tests/unit/webcrypto_test.ts
@@ -1,5 +1,62 @@
 import { assert, assertEquals, assertRejects } from "./test_util.ts";
 
+unitTest(async function testDigest() {
+  const subtle = window.crypto.subtle;
+  assert(subtle);
+
+  {
+    const result = await subtle.digest({ name: "MD5" }, new Uint8Array([42]));
+    const actual = [...new Uint8Array(result)];
+    const expected = [
+      0x33,
+      0x89,
+      0xda,
+      0xe3,
+      0x61,
+      0xaf,
+      0x79,
+      0xb0,
+      0x4c,
+      0x9c,
+      0x8e,
+      0x70,
+      0x57,
+      0xf6,
+      0x0c,
+      0xc6,
+    ];
+    assertEquals(actual, expected);
+  }
+
+  {
+    const result = await subtle.digest({ name: "SHA-1" }, new Uint8Array([42]));
+    const actual = [...new Uint8Array(result)];
+    const expected = [
+      0xdf,
+      0x58,
+      0x24,
+      0x8c,
+      0x41,
+      0x4f,
+      0x34,
+      0x2c,
+      0x81,
+      0xe0,
+      0x56,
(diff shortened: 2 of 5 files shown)


ext/crypto/lib.rs
```diff
@@ -1652,18 +1652,31 @@ pub fn op_crypto_random_uuid(
...
+    tokio::task::spawn_blocking(move || md5::compute(&data).to_vec().into())
+      .await
+  } else {
+    let algorithm = match algorithm.as_str() {
+      "SHA-1" => CryptoHash::Sha1,
+      "SHA-256" => CryptoHash::Sha256,
+      "SHA-384" => CryptoHash::Sha384,
+      "SHA-512" => CryptoHash::Sha512,
+      _ => return Err(not_supported()),
+    };
```
bnoordhuis, line 1668: It'd be more elegant to deserialize to CryptoHash here it's also more complicated and probably means having ext/crypto depend on serde_v8.

ext/crypto/00_crypto.js
```diff
@@ -74,6 +74,7 @@
 
   const supportedAlgorithms = {
     "digest": {
+      "MD-5": null,
```
littledivy, line 77: Is this "safe"? It will be possible to do this now and algorithm normalization won't complain, it will error at some later step: 
```rust
{
  name: "RSASSA-PKCS1-v1_5",
  modulusLength: 2048,
  publicExponent: new Uint8Array([0x01, 0x00, 0x01]), 
  hash: {
    name: "MD5"
  }
};
```

bnoordhuis: The thought has occurred to me but I didn't see a good alternative. How would you do it?

littledivy: I can't think of a clean way but we can maybe create a special path for MD5 in `digest()` itself, instead of doing WebCrypto normalization.

yacinehmito:
What's the rationale to support an algorithm that has been broken?

ext/crypto/00_crypto.js
bnoordhuis: Not a bad idea, PTAL.

littledivy (approved):

bartlomieju:
@yacinehmito see https://github.com/denoland/deno/issues/12715

bartlomieju (approved):
LGTM, if you want to land it

bnoordhuis:
@ry merge or not?

Cargo.lock
```diff
@@ -2178,6 +2179,12 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3e378b66a060d48947b590737b30a1be76706c8dd7b8ba0f2fe3989c68a853f"
 
+[[package]]
+name = "md5"
```
ry, line 2183: Do we not already have md5 somewhere in our binary?

cli/tests/unit/webcrypto_test.ts
```diff
@@ -1,5 +1,62 @@
 import { assert, assertEquals, assertRejects, unitTest } from "./test_util.ts";
 
+unitTest(async function testDigest() {
+  const subtle = window.crypto.subtle;
+  assert(subtle);
+
+  {
+    const result = await subtle.digest({ name: "MD-5" }, new Uint8Array([42]));
```
ry, line 8: Is "MD-5" name compatible with CFW? I couldn't find an example in their docs.

ry (approved):
LGTM

cli/tests/unit/webcrypto_test.ts
littledivy: IIRC Cloudflare MD5 algorthim is named "MD5" (without the `-`)

Cargo.lock
bnoordhuis: Surprisingly, no.

cli/tests/unit/webcrypto_test.ts
bnoordhuis: Good catch, updated.

cli/tests/unit/webcrypto_test.ts
```diff
@@ -1,5 +1,62 @@
 import { assert, assertEquals, assertRejects } from "./test_util.ts";
 
+unitTest(async function testDigest() {
```
bnoordhuis, line 3:
```suggestion
Deno.test(async function testDigest() {
```

bnoordhuis:
Getting it past the CI is getting too tedious, WPT in particular - it has a bunch of "expect error" tests for MD5 which now pass. I'm closing this for now, will revisit when we know for sure we're going to need it.

PR #12795, opened 2021-11-17
refactor(cli/proc_state): Dedup source cache for redirected specifiers

nayeemrmn:
Currently modules that go through redirects are duplicated in the in-memory cache.


kitsonk (approved):
LGTM

PR #12801, opened 2021-11-18
fix(lsp): tag deprecated diagnostics properly

kitsonk:
We weren't tagging deprecated diagnostics as deprecated, meaning they would not appear properly in the editor.


dsherret (approved):
LGTM

PR #12810, opened 2021-11-18
fix(testing): do not throw on error.errors.map

yacinehmito:
In tests, the function to format errors would assume that any property `errors` on error would contain an array, which is not necessarily the case.

I am currently porting a library to Deno and I was just bit by this.

### DIFF
--- a/runtime/js/40_testing.js
+++ b/runtime/js/40_testing.js
@@ -9,6 +9,7 @@
   const { serializePermissions } = window.__bootstrap.permissions;
   const { assert } = window.__bootstrap.util;
   const {
+    AggregateError,
     ArrayPrototypeFilter,
     ArrayPrototypePush,
     ArrayPrototypeSome,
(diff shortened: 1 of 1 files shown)


runtime/js/40_testing.js
```diff
@@ -297,7 +297,7 @@ finishing test case.`;
   }
 
   function formatError(error) {
-    if (error.errors) {
+    if (error.errors && typeof error.errors.map === "function") {
```
dsherret, line 300: Although extremely unlikely, it is also possible that it might have an `error.errors.map` that's a function that is not an array map function.

I think we should explicitly check if  `error instanceof AggregateError` here instead and import that from `window.__bootstrap.primordials`. That would also make it more clear what this code was doing. Alternatively we could check that `error.errors instanceof Array`.

dsherret (changes requested):
Thanks for this patch @yacinehmito! I think we just need to check for `error` being an `AggregateError` or possibly that `error.errors instanceof Array`

runtime/js/40_testing.js
yacinehmito: I went for `error instanceof AggregateError`.

yacinehmito:
Thank you for the review @dsherret.

dsherret (approved):
LGTM. Thanks @yacinehmito!

PR #12814, opened 2021-11-19
fix(core): keep event loop alive if there are ticks scheduled

bnoordhuis:
Demonstrates that nextTick() callbacks aren't invoked when there isn't
anything else keeping the event loop alive. Specifically:

1. nextTickCallback is invoked, but
2. macrotaskCallback is not

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -2464,4 +2464,47 @@ assertEquals(1, notify_return_value);
     assert_eq!(state.js_macrotask_cbs.len(), 2);
     assert_eq!(state.js_nexttick_cbs.len(), 2);
   }
+
+  #[tokio::test]
+  async fn test_has_tick_scheduled() {
+    let macrotask = Arc::new(AtomicUsize::default());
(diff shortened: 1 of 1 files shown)


bnoordhuis:
The test is passing after the second commit but I'm sorta confused: the "has tick scheduled" flag is still true but the event loop exits because the macrotask callback says it's done. Is that expected behavior?

bartlomieju:
> The test is passing after the second commit but I'm sorta confused: the "has tick scheduled" flag is still true but the event loop exits because the macrotask callback says it's done. Is that expected behavior?

I forgot to account for next ticks in keeping event loop alive. We should add an additional check here: 
https://github.com/denoland/deno/blob/c82ce7413366c49ae44128797f91fe44113c5e8c/core/runtime.rs#L804-L822 and keep event loop running

bartlomieju:
@bnoordhuis PTAL

bnoordhuis:
LGTM but I can't approve my own PRs. =)

bartlomieju (approved):

PR #12817, opened 2021-11-19
feat(ext/net): Support listening for UDP Broadcast packets

### LINKED ISSUE #10470: Error sending datagram to broadcast IP (permission denied)
Getting an error when trying to use `DatagramConn` to send to the network's broadcast address.

```ts
const msg = new TextEncoder().encode("Hello world\n");

const conn = Deno.listenDatagram({
  hostname: "0.0.0.0",
  port: 8002,
  transport: "udp",
});

await conn.send(msg, {
  hostname: "192.168.1.255",
  port: 8001,
  transport: "udp",
});
```

Running the above with:
```sh
deno run --allow-net --unstable test.ts
```
Output:
```
error: Uncaught (in promise) PermissionDenied: Permission denied (os error 13)
await conn.send(msg, {
^
    at unwrapOpResult (deno:core/core.js:102:13)
    at async file:///home/pi/test.ts:9:1
```

ProdigySim:
Supports setting/getting the broadcast flag on UDP sockets.

This exposes two new API calls on Datagram sockets, implemented as an extension of the existing `DatagramConn` interface:

```typescript
  /** **UNSTABLE**: new API, yet to be vetted.
   *
   * A transport listener for UDP messages. */
  export interface UdpConn extends DatagramConn {
    /** **UNSTABLE**: new API, yet to be vetted.
     * Gets the value of the SO_BROADCAST option for this socket.
     */
    readonly broadcast: boolean;
    /** **UNSTABLE**: new API, yet to be vetted.
     * Sets the value of the SO_BROADCAST option for this socket.
     * When enabled, this socket is allowed to send packets to a broadcast address.
     * @param flag True to set the SO_BROADCAST flag on the socket, false to unset the flag.
     */
    setBroadcast(flag: boolean): void;
  }
  ```
The overload of `listenDatagram()` for `{transport: 'udp'}` now returns this new type.

Fixes #10470

- [x] Unit testing:
  - [x] Sockets are `!broadcast` by default
  - [x] Sockets without broadcast flag error when sending to the broadcast address (`255.255.255.255`)
  - [x] Sockets with broadcast flag set can send successfully to the broadcast address
- [x] Manual testing
  - [x] Deno UdpConn can send messages to `255.255.255.255` and they are received by another computer on the same LAN.
  - [x] Deno UDP listeners can receive packets sent as broadcast from other computers on the same LAN
     -  (this is not new behavior, does not require broadcast flag on listener)

Considerations:
 - [ ] Is this API shape appropriate?
   - This API mimics [Tokio's UdpSocket](https://docs.rs/tokio/0.1.18/tokio/net/struct.UdpSocket.html) API for managing socket flags
   - Could also use ES6 getters/setters instead of an explicit `setBroadcast()` function.
   - Flags could be managed as a separate object instead of top-level on the socket object. 
     - e.g. `mySocket.flags.broadcast = true`
   - Underlying socket systems ([windows](https://docs.microsoft.com/en-us/windows/win32/winsock/sol-socket-socket-options), [unix](https://notes.shichao.io/unp/ch7/#ipv6-socket-options)) support a variety of socket options depending on the type of socket. This could add a lot of properties to the top level of a socket object.
   - Another option is to generically add `setSocketOption(option_name, option_value)` but this would still potentially need to be overloaded per-socket-type for accurate typing.
- [ ] Is the synchronous code for retrieving the socket resource in Rust appropriate?
  - I'm not as clear on the synchronicity of `setSockOpt()` on underlying platforms, but it seems like it would be valid to mutate socket options at any time.
  - [setSockOpt(3)](https://linux.die.net/man/3/setsockopt) does not list any failure cases for "Socket Busy", though some options cannot be set after a socket is connected.
  - Will the code here fail to set a socket option while the socket is sending or receiving? This would feel like an artificial limitation. (maybe I should test this)

### DIFF
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -339,14 +339,15 @@ unitTest(
     assert(alice.broadcast);
 
     const bob = Deno.listenDatagram({ port: 4501, transport: "udp" });
+    bob.setBroadcast(true);
     assert(bob.addr.transport === "udp");
     assertEquals(bob.addr.port, 4501);
     assertEquals(bob.addr.hostname, "127.0.0.1");
 
-    const broadcastBob = { ...bob, hostname: "127.0.0.255"};
+    const broadcastAddr = { ...bob.addr, hostname: "127.0.0.255"};
 
     const sent = new Uint8Array([1, 2, 3]);
-    const byteLength = await alice.send(sent, broadcastBob.addr);
+    const byteLength = await alice.send(sent, broadcastAddr);
 
     assertEquals(byteLength, 3);
 
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -275,11 +275,14 @@ unitTest(
   { permissions: { net: true } },
   async function netUdpSendReceive() {
     const alice = Deno.listenDatagram({ port: 3500, transport: "udp" });
+    console.log(`bcast status ${alice.broadcast}`);
+    assert(!alice.broadcast)
     assert(alice.addr.transport === "udp");
     assertEquals(alice.addr.port, 3500);
     assertEquals(alice.addr.hostname, "127.0.0.1");
 
     const bob = Deno.listenDatagram({ port: 4501, transport: "udp" });
+    assert(!alice.broadcast)
     assert(bob.addr.transport === "udp");
     assertEquals(bob.addr.port, 4501);
     assertEquals(bob.addr.hostname, "127.0.0.1");
@@ -328,12 +331,14 @@ unitTest(
 unitTest(
   { permissions: { net: true } },
   async function netUdpSendReceiveBroadcast() {
-    const alice = Deno.listenDatagram({ port: 3500, transport: "udp", broadcast: true });
-    assert(alice.addr.transport === "udp");
-    assertEquals(alice.addr.port, 3500);
-    assertEquals(alice.addr.hostname, "127.0.0.1");
+    const alice = Deno.listenDatagram({ port: 3500, transport: "udp" });
+    assert(!alice.broadcast);
+    alice.setBroadcast(false);
+    assert(!alice.broadcast);
+    alice.setBroadcast(true);
+    assert(alice.broadcast);
 
-    const bob = Deno.listenDatagram({ port: 4501, transport: "udp", broadcast: true });
+    const bob = Deno.listenDatagram({ port: 4501, transport: "udp" });
     assert(bob.addr.transport === "udp");
     assertEquals(bob.addr.port, 4501);
     assertEquals(bob.addr.hostname, "127.0.0.1");
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -330,7 +330,13 @@ unitTest(
 unitTest(
   { permissions: { net: true } },
   async function netUdpSendReceiveBroadcast() {
-    const alice = Deno.listenDatagram({ port: 3500, transport: "udp" });
+    // Must bind sender to an address that can send to the broadcast address on MacOS.
+    // Macos will give us error 49 when sending the broadcast packet if we omit hostname here.
+    const alice = Deno.listenDatagram({
+      port: 3500,
+      transport: "udp",
+      hostname: "0.0.0.0",
+    });
     assert(!alice.broadcast);
 
     const bob = Deno.listenDatagram({
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -332,24 +332,33 @@ unitTest(
   async function netUdpSendReceiveBroadcast() {
     const alice = Deno.listenDatagram({ port: 3500, transport: "udp" });
     assert(!alice.broadcast);
-    alice.setBroadcast(false);
-    assert(!alice.broadcast);
-    alice.setBroadcast(true);
-    assert(alice.broadcast);
 
     const bob = Deno.listenDatagram({ port: 4501, transport: "udp", hostname: '0.0.0.0'});
-    bob.setBroadcast(true);
     assert(bob.addr.transport === "udp");
     assertEquals(bob.addr.port, 4501);
-    assertEquals(bob.addr.hostname, "0.0.0.0");
+    assertEquals(bob.addr.hostname, '0.0.0.0');
 
     const broadcastAddr = { ...bob.addr, hostname: "255.255.255.255"};
 
     const sent = new Uint8Array([1, 2, 3]);
+
+    let acceptErrCount = 0;
+    const checkErr = (e: Error) => {
+      if (e.message === "Permission denied (os error 13)") {
+        acceptErrCount++;
+      } else {
+        throw new Error("Unexpected error message");
+      }
+    };
+    // Sending to the broadcast addr without broadcast flag should result in error.
+    await alice.send(sent, broadcastAddr).catch(checkErr);
+
+    alice.setBroadcast(true);
+    assert(alice.broadcast);
     const byteLength = await alice.send(sent, broadcastAddr);
 
-    assertEquals(byteLength, 3);
 
+    assertEquals(byteLength, 3);
     const [recvd, remote] = await bob.receive();
     assert(remote.transport === "udp");
     assertEquals(remote.port, 3500);
@@ -357,6 +366,7 @@ unitTest(
     assertEquals(1, recvd[0]);
     assertEquals(2, recvd[1]);
     assertEquals(3, recvd[2]);
+    assertEquals(acceptErrCount, 1);
     alice.close();
     bob.close();
   },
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ProdigySim** [comment]
Hmm, I started trying to refactor this to a `setBroadcast(true)` style API, but it looks like the Socket resources are stored in `AsyncRefCell<UdpSocket>`. This seems to imply that I would need to `await` to get a reference to the socket in order to set the broadcast flag--which would make the `setBroadcast()` API async, which seems wrong for an API which simply sets a flag on the socket.

Or, I'm just not familiar enough with rust to figure out how to accomplish this without relying on async behavior here.

**Comment by bartlomieju** [comment]
> Hmm, I started trying to refactor this to a `setBroadcast(true)` style API, but it looks like the Socket resources are stored in `AsyncRefCell<UdpSocket>`. This seems to imply that I would need to `await` to get a reference to the socket in order to set the broadcast flag--which would make the `setBroadcast()` API async, which seems wrong for an API which simply sets a flag on the socket.
> 
> Or, I'm just not familiar enough with rust to figure out how to accomplish this without relying on async behavior here.

This is possible to do synchronously if you ensure you're the only user of the `AsyncRefCell<UdpSocket>`, ie. there are no in-flight reads or writes. Here's an example of similar scenario in TLS related code:
https://github.com/denoland/deno/blob/ddfba7d8ca2ba65c3952b79b8f7677bcf410348e/ext/net/ops_tls.rs#L697-L717
If you ensure there are no in-flight reads and writes, you can "take" the resource from the resource table and then call `into_inner()` on it; after you're done you should add the resource back to the resource table.

**Comment by ProdigySim** [comment]
I got a synchronous version of the API coded up--it's not quite what you asked for though. I'm just doing a `try_borrow()` on the reference instead of removing/re-adding to the resource table. I'm not sure which operations for a UDP socket (other than reads and writes) could be in process.

Also, I'm hitting a more networking-related snag on unit testing... Just testing 
```typescript
assert(socket.broadcast === false)
socket.setBroadcast(true);
assert(socket.broadcast === true)
```
works fine, but actually producing broadcast messages (with correct addressing...) in the tests is proving troublesome. I'll keep working on it though.

ProdigySim:
I think I have this to a stable state. I've updated the PR description to summarize the PR and to call out considerations I'm not as clear on.

ProdigySim:
Unit tests were failing on windows because my check for the Error message didn't match cross platforms. Moved it to use an `instanceof Deno.error.PermissionDenied` test which seems to work on linux+windows. I don't have a mac to test.

cli/tests/unit/net_test.ts
```diff
@@ -325,6 +327,55 @@ unitTest(
...
+      } else {
+        throw new Error("Unexpected error message: " + e.message);
+      }
+    };
+    // Sending to the broadcast addr without broadcast flag should result in error.
+    await alice.send(sent, broadcastAddr).catch(checkErr);
+
+    alice.setBroadcast(true);
+    assert(alice.broadcast);
+    const byteLength = await alice.send(sent, broadcastAddr);
```
ProdigySim, line 363: This is the line that's failing on MacOS:
```
AddrNotAvailable: Can't assign requested address (os error 49)
```
Trying to figure out why this fails--I'm guessing it may be related to the IP addresses bound in this code. But I don't have access to a Mac currently to try and debug.

(The subsequent test fails because we do not close the socket)

bnoordhuis: Interestingly broadcasting from 0.0.0.0 to 255.255.255.255 WFM locally but I'm still on macos 10.14 so maybe that has something to do with it.

I'm okay with skipping the test on macos but please add a comment explaining why.

ext/net/ops.rs
```diff
@@ -53,6 +53,8 @@ pub fn init<P: NetPermissions + 'static>() -> Vec<OpPair> {
     ("op_net_listen", op_sync(op_net_listen::<P>)),
     ("op_dgram_recv", op_async(op_dgram_recv)),
     ("op_dgram_send", op_async(op_dgram_send::<P>)),
+    ("op_udp_setbroadcast", op_sync(op_udp_setbroadcast)),
+    ("op_udp_getbroadcast", op_sync(op_udp_getbroadcast)),
```
bnoordhuis, line 57: For consistency with other getters/setters:
```suggestion
    ("op_udp_set_broadcast", op_sync(op_udp_set_broadcast)),
    ("op_udp_get_broadcast", op_sync(op_udp_get_broadcast)),
```

ext/net/ops.rs
```diff
@@ -536,6 +538,49 @@ where
   }
 }
 
+#[derive(Deserialize)]
+struct SetBroadcastArgs {
+  rid: ResourceId,
+  flag: bool,
```
bnoordhuis, line 544: `on` or `enable`? Applicable to the .d.ts too.

bnoordhuis (approved):
Thanks, the technical side LGTM. @bartlomieju or someone else should sign off on the API (which looks fine to me but I'm not the best person to judge that.)

lucacasonato:
**Warning: I am not intimately familiar with UDP, but here are my 2 cents anyway.**

Can we always enable `SO_BROADCAST`? Go does this: https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/sockopt_linux.go#L21

Are there special permission implications/considerations of `SO_BROADCAST`?

cli/tests/unit/net_test.ts
ProdigySim: This is actually resolved now; the broken version of this unit test was not specifying a hostname at all when binding the sender--so it was using the default hostname option. `0.0.0.0` indeed seems to fix it. Awkwardly, this unit test will probably send a real UDP broadcast packet out on the network of whoever runs it though.

ProdigySim:
> **Warning: I am not intimately familiar with UDP, but here are my 2 cents anyway.**
> 
> Can we always enable `SO_BROADCAST`? Go does this: https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/net/sockopt_linux.go#L21
> 
> Are there special permission implications/considerations of `SO_BROADCAST`?

Fair question. All documentation ([linux](https://linux.die.net/man/3/setsockopt) [windows](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/so-broadcast) [mac](https://github.com/apple/darwin-xnu/blob/main/bsd/sys/socket.h#L133)) suggests that this flag only prohibits/allows _sending_ of broadcast packets. Setting this flag by default shouldn't have any unexpected consequence for receiving packets, and everyone who intends to send broadcast messages will probably appreciate having it enabled by defaults.

As far as rationale, this [IBM z/OS documentation](https://www.ibm.com/docs/en/zos/2.1.0?topic=control-so-broadcast-socket-option) is the only document I can find:
> Socket semantics require that an application set the SO_BROADCAST option on before attempting to send a datagram to a base or broadcast address. This protects the application from accidentally sending a datagram to many systems.

This may be a more outdated concern--I would think modern network infrastructure would lessen the impact of UDP broadcast spam (switches filtering out these packets, firewalls on end user machines). In Deno, developers will probably not often write low-level socket code, and the chance of "accidentally" crafting a broadcast address seems lower when the required input is a string instead of a 4 byte address (Imagine accidentally storing a `-1` error result as an IP address).

I would be happy with setting it by default as a consumer of the API, but it would be a breaking change to undo. We would probably still want some API to enable/disable certain socket flags in general though.

One option, which I had in the original version of this PR, would be to push the API up to the `listenDatagram()` API, with a `enableBroadcast: boolean` option for UDP sockets. This would solve the synchronicity problems in the API, and would probably cover most use cases for this particular flag.

We could even have this option defaulted true for convenience, and the API/types would help document this fact.

```typescript
// default to off and expose enableBroadcast
const bcastSocket = Deno.listenDatagram({ transport: 'udp', enableBroadcast: true });
// or: default to on and expose disableBroadcast
const sock2 = Deno.listenDatagram({ transport: 'udp', disableBroadcast: true });
```

ProdigySim:
Additionally of note, IPV6 does not require any flags on the socket to allow broadcast messages (messages targeted at the `anycast` address). So it seems like this concern has been dropped for modern networking.

See also: [Microsoft guide for porting Broadcast applications to IPV6](https://docs.microsoft.com/en-us/windows/win32/winsock/porting-broadcast-applications-to-ipv6)

ProdigySim:
Opened a new PR to just enable broadcast by default and add a unit test #12860

ProdigySim:
Not sure why the latest build failed, looked pretty unrelated to my code, so I rebased on main. It looked like it was failing on executing some scripts that were hosted on public URLs.

ProdigySim:
This is way deeper on the rabbit hole than I should go: [Original changeset on Go introducing broadcast-by-default](https://github.com/golang/go/commit/229807c8c3698c225a3fbcf594a40054e01c7758)

And [linked discussion thread that looks surprisingly similar to this one](https://codereview.appspot.com/162046).

Go seems to set a number of socket flags by default (like NODELAY on TCP), while Rust does not.

bartlomieju:
@ProdigySim thank you for the PR, I suggest we close this one in favor of #12860

PR #12819, opened 2021-11-19
feat(serde_v8): Support deserializing BigInt in `deserialize_any`

### LINKED ISSUE #11641: ffi panic on integer parameters over 32bits
I've been playing around with the new `Deno.dlopen`. In doing so, I've noticed that integers larger than 32bits produce some weird quirks.
Here is some test C code...
```c
#include <stdint.h>
void takeU64(uint64_t x) {}
void takeU32(uint32_t x) {}
```
and a `dlopen`...
```ts
const dylib = Deno.dlopen("usize.dll", {
  "takeU64": { parameters: ["u64"], result: "void" },
  "takeU32": { parameters: ["u32"], result: "void" },
});
```

And some calls to demonstrate the issue:
```ts
// Perhaps expected, but C would truncate: (uint32_t)(8589934591) == 4294967295
dylib.symbols.takeU32(0x00000001FFFFFFFF);
// thread 'main' panicked at 'Expected ffi arg value to be an unsigned integer', extensions\ffi\lib.rs:260:6

// This runs successfully.
dylib.symbols.takeU64(0x00000000FFFFFFFF);
// This does not.
dylib.symbols.takeU64(0x00000001FFFFFFFF);
// thread 'main' panicked at 'Expected ffi arg value to be an unsigned integer', extensions\ffi\lib.rs:260:6
```

---

I also tried using BigInts to see if this was a problem I could avoid todo with JavaScript's 2^53 integer limit, but this also caused a panic:
```ts
dylib.symbols.takeU64(BigInt(0x00000001FFFFFFFF));
// thread 'main' panicked at 'serde_v8: unknown ValueType for v8::Value', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\serde_v8-0.9.0\src\payload.rs:32:5
```

I'm not experienced enough in the Deno internals or Rust to take this on myself, but I did some digging.
From that `bigint` panic, `payload.rs` in serde_v8 does not have an entry for BigInt. I found a related commit denoland/serde_v8@57ce430716d8abcd44a10166b3282bdb66703f78, but this related issue is still unresolved denoland/deno#6169.

It also feels like the panic error message in [extensions/ffi/lib.rs](https://github.com/denoland/deno/blob/v1.13.0/extensions/ffi/lib.rs) could be improved to show what type was attempted to be passed (if possible to help with issues like this), but I get this API is still unstable and very, very new.

DjDeveloperr:
- [x] Adds support for `BigInt` in `deserialize_any`
- [x] Adds relevant test case in `serde_v8`
- [x] Also adds test case for passing BigInt for FFI u64 argument.

Fixes #11641


AaronO (commented):
In isolation LGTM, deserializing BigInts is ok and makes sense, it's mainly serializing them that doesn't (we don't want to serialize all `u64`s as `BigInt`s).

If this is mainly for FFI, are you sure that opaque pointers transmuted as f64s don't work ?

DjDeveloperr:
> If this is mainly for FFI, are you sure that opaque pointers transmuted as f64s don't work ?

It does work but only for passing arguments, so yeah it *is* a *workaround* for the mentioned issue.

eliassjogreen:
> we don't want to serialize all u64s as BigInts

@AaronO I would argue we do want to serialize and deserialize all integer types wider than 32 bits, this is something that is reflected in js with the `BigUint64Array` and `BigInt64Array` typedarrays and would probably not break any code in deno which already wasn't broken for values larger than `f64`s integer limit (53 bits). Ideally we would want to be able to deserialize both `number` and `bigint` types as 64 bit integers, but only serialize 64 bit integers to `bigint`s (which is why I think if we have to choose `bigint`s are a better representation of integers wider than 32 bits than the current approach).

PR #12823, opened 2021-11-20
feat(ext/ffi): support `null` in buffer argument

### LINKED ISSUE #12809: feature request: support for null value in FFI buffer argument
There's no way to pass a null pointer to a buffer argument in FFI right now. It would be great to support passing JS `null` value corresponding to a `nullptr` in FFI. I'm currently trying to use WinAPI which (of course) makes use of those.

I'd like to try work on this

DjDeveloperr:
- [x] Adds support for `null` in buffer argument. Corresponds to `nullptr` in FFI.
- [x] Adds relevant test case.

Fixes #12809

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -121,6 +121,43 @@ declare namespace Deno {
 
   /** A foreign function as defined by its parameter and result types */
   export interface ForeignFunction {
+    /**
+     * Foreign Function paramter types also support `buffer` type which is essentially
+     * a pointer to a JS `Uint8Array`. It can even be mutated from native code.
+     * 
+     * For example, take the following Rust code,
+     * 
+     * ```rs
+     * #[no_mangle]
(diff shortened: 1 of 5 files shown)


bartlomieju (changes requested):
@DjDeveloperr tentatively this looks good to me, but please update type declarations to reflect that if buffer is `null` then buffer length shouldn't be passed as the next argument. Also an update to manual would be appreciated too.

DjDeveloperr:
Do I add that to type declarations of `Deno.ForeignFunction`'s `parameters` field? I didn't quite get as passing length of buffer as next argument is not required.

bartlomieju:
> 

I guess you don't need to add any types to `ForeignFunction`, but I'd be good to add some examples into the JSDoc for it.

DjDeveloperr:
@bartlomieju Added an example for buffer argument and some notes about null buffer argument in JS doc.

DjDeveloperr:
Looks like #12828 adds this too.. I guess I'll close this.

PR #12825, opened 2021-11-20
fix(lsp): use exclude files list from the config file

### LINKED ISSUE #12608: Lint config to exclude files doesn't do anything (for VS Code)?
Here is my deno.json:

```json
{
  "lint": {
    "rules": {
      "include": [
        "ban-unused-ignore",
        "no-debugger",
        "adjacent-overload-signatures",
        "ban-unknown-rule-code",
        "no-case-declarations",
        "no-fallthrough",
        "no-inferrable-types",
        "no-inner-declarations",
        "no-unused-vars",
        "no-var",
        "prefer-const",
        "eqeqeq",
        "ban-untagged-todo"
      ]
    },
    "files": {
      "exclude": ["emitted/"]
    }
  }
}
```

The rules config is picked up, but the lint.files.exclude part doesn't actually prevent lint errors from showing up in the "emitted" directory.  I'm basing my usage of it on the example here: https://deno.land/manual@v1.15.3/getting_started/configuration_file

igorsaux:
I think it will be better to move the vector of ignored paths to somewhere like `StateSnapshot` and update it on every config change, not on every `publish_diagnostics`.

Also, I've inverted an if statement in `publish_diagnostics`, so it looks a bit better.

Fixes #12608

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -2,6 +2,7 @@
 
 use super::analysis;
 use super::documents;
+use super::documents::Document;
 use super::documents::Documents;
 use super::language_server;
 use super::tsc;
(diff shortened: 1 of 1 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -517,42 +517,79 @@ async fn generate_deps_diagnostics(
...
+  let mut changes = match collection.take_changes() {
+    None => return,
+    Some(changes) => changes,
+  };
+
+  // Exclude diagnostics from ignored files.
+  changes.retain(|change| {
+    let change_path = change.path();
+
+    !ignore_specifiers
```
dsherret, line 558: Will this cause non-lint diagnostics to be ignored? The `lint_config.files.exclude` is just for linting.

cli/lsp/diagnostics.rs
```diff
@@ -517,42 +517,79 @@ async fn generate_deps_diagnostics(
...
+  // FIXME: `root_uri` from `ConfigSnapshot` are without a trailing slash,
+  // so `Url::join` treats the root as a file, not a directory, and erases the folder name.
+  // To fix that behaviour we just parsing `root_uri` again.
+  let root =
+    ModuleSpecifier::from_directory_path(root.to_file_path().unwrap()).unwrap();
+  ignore.iter().filter_map(|i| root.join(i).ok()).collect()
+}
+
 /// Publishes diagnostics to the client.
 async fn publish_diagnostics(
```
dsherret, line 534: I think maybe doing this check at `publish_diagnostics` is too late. Perhaps we should ignore these files in `generate_lint_diagnostics` so that we don't bother doing this work when not needed?

dsherret (changes requested):
Thanks for this @igorsaux! I believe it would cause non-lint diagnostics to be ignored? The lint files exclude should only apply to linting.

If you are up for it, I also think we could move this filtering to an earlier point so the ignored files aren't linted in the first place.

cli/lsp/diagnostics.rs
igorsaux: Thanks! You're right, I thought it should ignore all the diagnostics.
So, if I get You right, I need to filter a result of `snapshot.documents.documents(true, true)` in `generate_lint_diagnostics` and leave the rest of the function as is.

dsherret: Yes, I believe that would be better. Also, it looks like we're not handling `files.include` in addition to `files.exclude`.

igorsaux: > I think maybe doing this check at `publish_diagnostics` is too late. Perhaps we should ignore these files in `generate_lint_diagnostics` so that we don't bother doing this work when not needed?

Ready.

> Yes, I believe that would be better. Also, it looks like we're not handling `files.include` in addition to `files.exclude`.

Handling of `files.include` should be after excluding?

igorsaux: Seems handling of `include` section has no meaning.
For `deno lint` it points to external files and does not re-include excluded files, but LSP (I've tested with VSCode) always lints all the opened files in an editor, even if they are from another disk partition (bug?), so, this section is useless. Correct me if I'm wrong.

dsherret: With `files.include`, if it is not empty I believe it would ignore any files/folders not in `documents` if it's not in `files.include`.

igorsaux: >it would ignore any files/folders not in `documents`

Sorry, I don't sure I understand, but `generate_lint_diagnostics` collects diagnostics only from `documents`.
Maybe You meant:
> if `files.include` is not empty - ignore anything that is not in `files.include`

dsherret: Sorry! Yes. I fixed my sentence.

cli/lsp/diagnostics.rs
```diff
@@ -300,13 +301,81 @@ fn ts_json_to_diagnostics(
     .collect()
 }
 
+// Returns `ConfigSnapshot::root_uri` in the correct format.
+fn get_root_specifier(
```
dsherret, line 305: I'm going to resolve this in #12863.

dsherret (approved):
LGTM. Thanks @igorsaux!

Issue #12824, opened 2021-11-20
Proposal: `Deno.UnsafePointer` and `UnsafePointerView` API


DjDeveloperr:
I propose adding two new classes: `UnsafePointer` and `UnsafePointerView` to the Deno namespace. The name itself would suggest that it is unsafe (inspired from [Swift](https://developer.apple.com/documentation/swift/unsafepointer)). This introduces a unified interface for passing/returning raw pointers to/from FFI, also a view class over the pointer (such as `DataView`) to perform read-related operations on the pointer. Of course, using this needs `--allow-ffi` permission.

```ts
type PointerOffset = number | bigint;

declare class UnsafePointer {
  constructor(value: bigint);

  /**
   * Return direct pointer to the Uint8Array in memory.
   */
  static of(buffer: Uint8Array): UnsafePointer;

  /** 
   * Returns the value of the pointer. 
   * This would allow pointer arithmetic which is useful in certain scenarios. 
   */
  valueOf(): bigint;
}

declare class UnsafePointerView {
  pointer: UnsafePointer;

  constructor(pointer: UnsafePointer);

  getUint8(offset?: PointerOffset): number;
  getInt8(offset?: PointerOffset): number;
  getUint16(offset?: PointerOffset): number;
  getInt16(offset?: PointerOffset): number;
  getUint32(offset?: PointerOffset): number;
  getInt32(offset?: PointerOffset): number;
  getBigUint64(offset?: PointerOffset): bigint;
  getBigInt16(offset?: PointerOffset): bigint;
  getFloat32(offset?: PointerOffset): number;
  getFloat64(offset?: PointerOffset): number;
  getCString(offset?: PointerOffset): string;
  getArrayBuffer(byteLength: number, offset?: PointerOffset): ArrayBuffer;
  copyInto(destination: Uint8Array, offset?: PointerOffset): void;
}
```

I didn't include writing to pointers because that's unsafe on another level. Most use-cases won't even need that.

### Usage

This would work along side the current `buffer` FFI type. For passing buffers, both `UnsafePointer` and `Uint8Array` would be allowed with the `buffer` argument type, while `buffer` as a result type would always return `UnsafePointer`.

### Fixes

- https://github.com/denoland/deno/issues/12456
- https://github.com/denoland/deno/issues/12752

Being implemented by #12828

eliassjogreen:
[...]

Issue #12752, opened 2021-11-13
suggestion: ability to read C string for FFI


DjDeveloperr:
### Suggestion

A `readCString` or similar function in maybe `Deno.DynamicLibrary` that accepts a pointer (as `BigInt`) and returns the C String.

### Why?

When working with FFI, it's pretty common to use C strings. But in case of Deno FFI it's not possible to do that without a *helper* function. Right now I'm interacting with a native library directly using Deno FFI (without wrapping it) and I have to make such helper function by making another native library.

### Why function and not a string return type?

Only being able to read C strings returned from some function is limiting in my case. In the dylib I'm using, some functions write a C String pointer to a `char**` instead.

DjDeveloperr:
Implemented by https://github.com/denoland/deno/pull/12828

Issue #12456, opened 2021-10-16
Deno has FFI but knows nothing about memory pointers


lem0nify:
Let's say we need to pass a string in the procedure parameter. Good! We specify the type "buffer", take an `Uint8Array` and write the encoded string there byte-by-byte, ending everything with a null byte, and then pass it.

Let's say we need to pass a structure in the procedure parameter. Fine! Again, we specify the "buffer" type, take an `Uint8Array` of the desired size, fill it with struct padding and all the crap (if it's a parameter for input, not output, of course) and pass it.

But what if we need to pass in a parameter a structure that should store a pointer to another structure or string? What are we doing? That's right: close the project and go code in Python! Or NodeJS. Or anything with a better FFI implementation or native extensions.

Why are we passing the Uint8Array as an object at all? Why not make something like `Deno.byref` (similar to `ctypes.byref` in Python)?

Sorry if I sound toxic. You just removed the native extensions, and I have been waiting for a reasonable replacement for a long time, but you choose not the best ways to implement FFI, as it seems to me.

DjDeveloperr:
I think this can be closed now that https://github.com/denoland/deno/pull/12828 has landed

PR #12828, opened 2021-11-20
feat(ext/ffi): Implement `UnsafePointer` and `UnsafePointerView` API

### LINKED ISSUE #12809: feature request: support for null value in FFI buffer argument

eliassjogreen:
This pr adds support for passing, returning and reading pointers as proposed by the `UnsafePointer` and `UnsafePointerView` api (#12824).

Closes #12809, #12456, #12824 and #12752


eliassjogreen:
This pr could very easily implement `UnsafePointer` as proposed in #12824 if that is what we want. It would most likely make working with pointers way easier imo

bnoordhuis:
I'm not sure using f64's is sound. V8 tends to canonicalize NaNs - maybe Infs and subnormals too, but definitely NaNs - which means there's a subset of the 64 bit space that's going to get corrupted.

You could _maybe_ get away with it if it's only NaNs and Infs<sup>1</sup> but definitely not when it's subnormals too. An example of a subnormal is `0000 0000 0000 0001` (`1u64` interpreted as f64.)

tl;dr Don't rely on V8 leaving the top 12 bits untouched, use BigInts instead.

<sup>1</sup> ARM64 and x86_64 have up to 52 address lines, which is exactly the f64 significand's size. NaNs and Infs are encoded in the exponent, not the significand.

eliassjogreen:
Would returning and passing pointers as a tuple of the low and high part of a `u64` be preferred or should we wait for `serde_v8` to support `bigint` (de)serialization?

eliassjogreen:
@bnoordhuis internally pointers are now represented as the high and low bytes of an `u64` (`(u32, u32)`) which should avoid the issues with transmuting pointers as `f64`. I also implemented the `UnsafePointer` API to make it easier to work with pointers although it should probably be looked through for my sanity and to make sure it all works as expected, cc @DjDeveloperr

ext/ffi/00_ffi.js
```diff
@@ -6,7 +6,52 @@
...
+  function unpackPointer([hi, lo]) {
+    return BigInt(hi) << 32n | BigInt(lo);
+  }
+
+  function packPointer(value) {
+    return [Number(value >> 32n), Number(value & 0xFFFFFFFFn)];
+  }
+
+  class UnsafePointer {
+    value;
```
DjDeveloperr, line 23: Could `valueOf` method be implemented as well?

ext/ffi/lib.rs
```diff
@@ -108,6 +112,9 @@ pub fn init<P: FfiPermissions + 'static>(unstable: bool) -> Extension {
       ("op_ffi_load", op_sync(op_ffi_load::<P>)),
       ("op_ffi_call", op_sync(op_ffi_call)),
       ("op_ffi_call_nonblocking", op_async(op_ffi_call_nonblocking)),
+      ("op_ffi_ptr_of", op_sync(op_ffi_ptr_of)),
```
DjDeveloperr, line 115: Unsafe Pointer class' methods like static `of`, `read`, `readCString` should require `allow-ffi` permission IMO

DjDeveloperr (commented):
LGTM, got a few comments

ext/ffi/00_ffi.js
eliassjogreen: Hmm, yeah but im undecided if we want pointer values to be immutable because the `UnsafePointer` constructor allows the user to create new pointers from an arbitary value so why not then allow the value property to be mutable and just leave the value property exposed.

ext/ffi/00_ffi.js
```diff
@@ -6,7 +6,52 @@
...
+
+    constructor(value) {
+      this.value = value;
+    }
+
+    static null() {
```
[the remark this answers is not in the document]
eliassjogreen: I agree with this, will fix

ext/ffi/00_ffi.js
DjDeveloperr: Ah, no I didn't mean making value immutable, but just adding `valueOf` method too, it would allow things such as `new UnsafePointer(1n) + 2n`

eliassjogreen: ah ok, didn't know about that but sure!

ext/ffi/lib.rs
eliassjogreen: Hmm, how would I implement this? `FfiPermissions::check` requires a `path` component, should i expose and use `FfiDescriptor::query` or how would i check if `allow-ffi` is enabled without checking any specific path?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -117,16 +117,59 @@ declare namespace Deno {
...
+     * Return the direct memory pointer to the typed array in memory
+     */
+    static of(typedArray: TypedArray): UnsafePointer;
+
+    /**
+     * Reads memory at the pointer into a typed array.
+     * Length is determined from the typed array's `byteLength`.
+     * Also takes optional offset.
+     */
+    read(destination: TypedArray, offset?: bigint | number): void;
```
rracariu, line 160: My understanding is that this would copy data, would be nice to have an view based approach over a said pointer into native memory that uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView as a way to read typed information from the raw memory the pointer points to.

eliassjogreen: Perhaps a read only (kinda dangerous writing to arbitrary pointers, but maybe...) copy of the `DataView` API? Although there is a few questions i have with it:
1. It would have to work with `bigint` offsets instead of `numbers`
2. Why would you want to create more than one instance of this new `DataView` copy as they all address the same memory? Or should it be an extension to the `UnsafePointer` API?
3. The `read` method should probably also be renamed `copy` as that more accurately reflects what it does, or maybe we should even add a `getArrayBuffer(byteLength: bigint | number, offset?: bigint | number): ArrayBuffer` (or `Uint8Array` maybe?)

rracariu: The new structure should mimic the `DataView` API, work with `bigint` as you said.

I think having 2 methods on the `UnsafePointer` will be nice:
- `UnsafePointer::toReadonlyView(offset:bignum, size: bignum): ReadOnlyPointerView`
- `UnsafePointer::toWritableView(offset:bignum, size: bignum): WritablePointerView`
Where `WritablePointerView` would extend `ReadOnlyPointerView` and allow write access to the underlying memory. This is just a proposal so naming and form could be definitely be improved...

I agree that the `read` method should be named to `copy` as it as clear as possible about the intent.
Maybe create a new PR with this new `View` APIs so it is easier to review and make this one mergeable quicker?

DjDeveloperr: > The `read` method should probably also be renamed `copy` as that more accurately reflects what it does, or maybe we should even add a `getArrayBuffer(byteLength: bigint | number, offset?: bigint | number): ArrayBuffer` (or `Uint8Array` maybe?)

I had it to be named `read` in the proposal as it seemed consistent with `Deno.Reader#read`. But rename to `copy` would make sense, I agree.

eliassjogreen: > Maybe create a new PR with this new `View` APIs so it is easier to review and make this one mergeable quicker?

Ideally yes, but should this PR then leave out the `read` and `readCString` methods in favour of that future pr?

eliassjogreen:
This PR is ready for review now!

ext/ffi/lib.rs
```diff
@@ -267,6 +296,21 @@ fn value_as_f64(value: Value) -> f64 {
     .expect("Expected ffi arg value to be a float")
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+struct U32x2(u32, u32);
+
+impl From<u64> for U32x2 {
+  fn from(value: u64) -> Self {
+    U32x2((value >> 32) as u32, value as u32)
```
bnoordhuis, line 304:
```suggestion
    Self((value >> 32) as u32, value as u32)
```
For consistency?

ext/ffi/lib.rs
```diff
@@ -507,6 +560,177 @@ async fn op_ffi_call_nonblocking(
...
+  _: (),
+) -> Result<(), AnyError>
+where
+  FP: FfiPermissions + 'static,
+{
+  let permissions = state.borrow_mut::<FP>();
+  permissions.check(None)?;
+
+  let src = u64::from(src) as *const u8;
+  unsafe { ptr::copy(src, dst.as_mut_ptr(), len) };
```
bnoordhuis, line 589: Suggestion: return an error if `dst.len() < len`?

ext/ffi/lib.rs
```diff
@@ -507,6 +560,177 @@ async fn op_ffi_call_nonblocking(
...
+  ptr: U32x2,
+  _: (),
+) -> Result<u16, AnyError>
+where
+  FP: FfiPermissions + 'static,
+{
+  let permissions = state.borrow_mut::<FP>();
+  permissions.check(None)?;
+
+  Ok(unsafe { *(u64::from(ptr) as *const u16) })
```
bnoordhuis, line 647: This won't work on architectures where unaligned loads trap. Use `std::ptr::read_unaligned()` instead.

bnoordhuis (commented):
Nice work. Mostly LGTM.

ext/ffi/lib.rs
eliassjogreen: Actually this won't ever be a problem as the one place where it is called is:
```js
core.opSync("op_ffi_buf_copy_into", [
  packU64(this.pointer.value + BigInt(offset)),
  destination,
  destination.byteLength,
]);
```
and here the length is always the length of the `dst` variable, but implemented it anyways for good measure.

brandonros:
What next release can we expect this to land in? v1.17.0?

DjDeveloperr:
Yes, it's going to be in 1.17.0 release which is a bit delayed it seems

lucacasonato:
@DjDeveloperr It is not delayed. 1.17.0 will release tomorrow as documented in the release schedule: https://deno.land/manual@v1.16.4/contributing/release_schedule. I have updated the milestone accordingly now.

PR #12831, opened 2021-11-21
fix(workers): Make `worker.terminate()` not immediately kill the isolate

### LINKED ISSUE #12658: worker core dumped panic
```ts
const b64 = btoa(
  'await new Promise((res) => setTimeout(res, 1000));throw "error";',
);

const w = new Worker(`data:application/javascript;base64,${b64}`, {
  type: "module",
});

w.onerror = (e) => console.log("some error occured");
w.terminate();
```
This code has a chance to "panic" with a core dumped error on linux.
Both on deno 1.15.3 and canary(1.15.3+26a5471)

C stacktrace
```
#
# Fatal error in , line 0
# Check failed: (location_) != nullptr.
#
#
#
#FailureMessage Object: 0x7f620b7da3b0
==== C stack trace ===============================

    deno(+0x1b44ef3) [0x55aa66f9eef3]
    deno(+0x1b3ff6b) [0x55aa66f99f6b]
    deno(+0x1b40bb5) [0x55aa66f9abb5]
    deno(+0x1fc7385) [0x55aa67421385]
    deno(+0x1fc6be1) [0x55aa67420be1]
    deno(+0x1fc666d) [0x55aa6742066d]
    deno(+0x1fc64bc) [0x55aa674204bc]
    deno(+0x1b966ac) [0x55aa66ff06ac]
    deno(+0x35cfdb2) [0x55aa68a29db2]
    deno(+0x3090858) [0x55aa684ea858]
    deno(+0x2f3489c) [0x55aa6838e89c]
    deno(+0x2f3e4c2) [0x55aa683984c2]
    deno(+0x2954703) [0x55aa67dae703]
    /lib/x86_64-linux-gnu/libpthread.so.0(+0x9450) [0x7f6213336450]
    /lib/x86_64-linux-gnu/libc.so.6(clone+0x43) [0x7f621310ad53]
Trace/breakpoint trap (core dumped)
```

andreubotella:
Due to a bug in V8, terminating an isolate while a module with top-level await is being evaluated would crash the process. This change makes it so calling `worker.terminate()` will signal the worker to terminate at the next iteration of the event loop, and it schedules a proper termination of the worker's isolate after 2 seconds.

Closes #12658


bartlomieju (commented):
I think this is a sensible solution, especially if it fixed the problem at hand. @lucacasonato @bnoordhuis PTAL

andreubotella:
`cargo test` was not failing on my machine, but it was on CI before 94bc4e4, so I'd want to run CI a couple more times to be confident that I'm not introducing flakes.

runtime/web_worker.rs
```diff
@@ -136,7 +139,21 @@ impl WebWorkerInternalHandle {
...
+  pub fn terminate_if_needed(&mut self) -> bool {
+    let has_terminated = self.is_terminated();
+
+    if !has_terminated && self.termination_signal.load(Ordering::SeqCst) {
+      self.terminate();
+      return true;
+    }
+
+    has_terminated
   }
```
bnoordhuis, line 157: This method can end up calling `self.terminate()` repeatedly, can't it? That seems bad.

bnoordhuis (commented):
How pressing is this bug? My natural inclination is to wait for the upstream fix (the V8 bug had some movement earlier today) than try to work around it.

runtime/web_worker.rs
andreubotella: `terminate_if_needed()` will at most call `self.terminate()` once – there are no loops or recursion in this method. And if `self.terminate()` has been previously called, `self.is_terminated()` will return true, so `self.terminate()` won't be called again.

andreubotella:
> How pressing is this bug? My natural inclination is to wait for the upstream fix (the V8 bug had some movement earlier today) than try to work around it.

I guess it's not too pressing. But this isn't the only bug related to isolate termination that we've had (see #12263), and considering that this PR changes worker termination to work the same as Chrome, I'd say this might be better in the long run.

runtime/web_worker.rs
bnoordhuis: Sorry, let me clarify: calling `terminate_if_needed()` twice can result in two calls to `terminate()`.

That's okay when `terminate()` is idempotent but I don't think it is, and even if it is, then that's an implicit contract that's easy to miss when making changes later

andreubotella: `self.terminate()` will set `self.has_terminated` to true, which will keep any subsequent call to `self.terminate_if_needed()` from calling `self.terminate()` again.

bnoordhuis: That's the implicit contract I mean. Something a little less prone to snafus would be appreciated.

A rule of thumb I follow is this: whenever more than one atomic variable pops up, try really hard to replace it with a mutex and a proper critical section.

bartlomieju (approved):
LGTM

I discussed this PR offline with @bnoordhuis who's not to keen on landing it, but since it's fixing an actual bug we agreed to merge it.

ry:
This PR resulted in many more threads being used in the "workers_large_message" benchmark:

Screen Shot 2021-12-01 at 8 34 38 AM

andreubotella:
I decided to spawn a thread rather than start a sleep future in the tokio event loop in case the host was terminated in the meantime, but if the increase in threads is a problem, we can revisit this.

Issue #12366, opened 2021-10-07
Deno.readTextFile does not work properly in Deno 1.14.x


jeffreyjw:
Trying to read a system text file (does not matter if using sudo or not) returns an empty string in Linux:

`const content = await Deno.readTextFile("/proc/bus/input/devices");`

This code was working properly in Deno 1.13.1, it returned the full contents of the file, however it does not work in Deno 1.14.1 and 1.14.3 returning only an empty string (no error is being thrown).

AaronO:
This bug might be caused by the changes to `readFile` so that it stats the file first

jeffreyjw:
@AaronO this seems to be the case, `stat /proc/bus/input/devices` does indeed report the filesize to be 0. Is there a workaround for this in the meantime?

PR #12835, opened 2021-11-22
tests: extend readFile file length during read

lucacasonato:
This commit adds some tests that demonstrate that Deno.readFile reads
the entire file, even if the read file is extended during read.

#12366 is tangentially related


bartlomieju (approved):
LGTM

PR #12837, opened 2021-11-22
revert: store header keys lower case internally

### LINKED ISSUE #12829: `fetch` headers do not maintain casing on the wire in HTTP/1
https://github.com/denoland/deno/issues/10958 had the unfortunate side effect that headers are always lowercased on the wire. For HTTP2 this is the case anyway, but it apparently breaks some HTTP/1 servers that don't understand that headers are case insensitive :man_facepalming:.

We could fix this while still maintaining some of the speedup introduced in this PR by introducing a `lowercase name` -> `case used in header list` hashmap inside of the `Headers` object.

lucacasonato:
This reverts commit 49ec3d10ad90851f4d28274a3f0fe96c642204ac.

This might introduce a slight performance degredation, but correctness is more important here. We can optimize this structure down the line by storing headers in a more optimized format internally (like Chromium does).

Fixes #12829

### DIFF
--- a/ext/fetch/20_headers.js
+++ b/ext/fetch/20_headers.js
@@ -114,7 +114,13 @@
 
     // 7.
     const list = headers[_headerList];
-    name = byteLowerCase(name);
+    const lowercaseName = byteLowerCase(name);
+    for (let i = 0; i < list.length; i++) {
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM

PR #12841, opened 2021-11-22
fix(ext/console): truncating long strings with MBC

### LINKED ISSUE #12447: Repl fails to parse some json inputs
Trying this line in the repl
```ts
JSON.parse('{"x": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa🌎"}')
```

fails with `Unterminated string literal`

It only fails if there is exactly 99 'a' (or any character) before the world unicode 
If I add or remove a couple of characters before the unicode it  works.

meta:
deno 1.14.3 (release, x86_64-unknown-linux-gnu)
v8 9.4.146.19
typescript 4.4.2

lucacasonato:
This commit changes the `Deno.inspect` API to not terminate long string
literals in the middle of multi byte characters. Instead splits now
always happen at the boundary of a grapheme.

Fixes #12447


ext/console/02_console.js
```diff
@@ -739,6 +739,40 @@
...
+      if (unicodeIndex < end) {
+        accumulator += character;
+      }
+      stringIndex += character.length;
+      unicodeIndex += 1;
+    }
+
+    return accumulator;
+  }
+
```
andreubotella, line 775: We don't really care about what non-BMP or lone surrogate characters there were at any point before where we're truncating the string. So we could just check whether `end - 1` is a lead surrogate (0xD800 to 0xDBFF), and if it is, truncate at `end - 1`. This would do the "wrong" thing if there's a lone lead surrogate at that point, but I don't think that's a problem.

But in either case, whether in this implementation or the way I suggested, it is possible for the truncated string to end with a lead surrogate (if there's a lone lead surrogate at index 99, a lead surrogate at 100, and a trail surrogate at 101).

lucacasonato: @andreubotella Sounds reasonable. I'll update the PR.

lucacasonato: Updated now.

ext/console/02_console.js
```diff
@@ -783,10 +749,14 @@
...
         if (value.length > STR_ABBREVIATE_SIZE) {
-          const trunc = multibyteAwareTruncate(value, STR_ABBREVIATE_SIZE - 3);
-          if (value !== trunc) {
-            value = trunc + "...";
+          value = value.substring(0, STR_ABBREVIATE_SIZE - 3);
+          // If the last byte is the leading byte of a multi byte character,
+          // remove that character.
+          const lastChar = value.charCodeAt(value.length - 1);
+          if (lastChar >= 0xD800 && lastChar <= 0xDBFF) {
+            value = value.substring(0, value.length - 1);
```
andreubotella, line 757: You could use `value.at(-1)` and do a string comparison. Similarly with `value.slice(0, -1)`.

Also, these operations should use primordials.

PR #12862, opened 2021-11-22
refactor(timers): Refactor timers to use one async op per timer

### LINKED ISSUE #8965: If one test case fails with leaking async ops related to timeout, other tests that should fail with leaking async ops do not
Both of the tests below have the same async function. They should both fail for leaking async ops. The first creates 2 timeouts, but the failure shows there is only 1 more dispatched than completed async ops. When the next test starts, it starts with there being one more dispatched op than completed ops. The setTimeout calls in the second test do not result in any change to the number of dispatched and completed async ops. Because of that, the second test does not fail.

```ts
import { delay } from "https://deno.land/std@0.80.0/async/delay.ts";

Deno.test("example 1", async () => {
  let metrics = Deno.metrics();
  console.log(
    `
Before:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );

  setTimeout(() => {}, 10000);
  setTimeout(() => {}, 10001);

  metrics = Deno.metrics();
  console.log(
    `After:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );
  await delay(0);
});

Deno.test("example 2", async () => {
  let metrics = Deno.metrics();
  console.log(
    `
Before:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );

  setTimeout(() => {}, 10000);
  setTimeout(() => {}, 10001);

  metrics = Deno.metrics();
  console.log(
    `After:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );
  await delay(0);
});
```

```sh
deno --version
deno 1.6.0 (release, x86_64-unknown-linux-gnu)
v8 8.8.278.2
typescript 4.1.2
deno test example_test.ts
Check file:///mnt/c/Users/kylej/projects/deno/test_suite/$deno$test.ts
running 2 tests
test example 1 ... 
Before:
- dispatched: 0
- completed: 0
After:
- dispatched: 1
- completed: 0
FAILED (5ms)
test example 2 ... 
Before:
- dispatched: 5
- completed: 4
After:
- dispatched: 5
- completed: 4
ok (3ms)

failures:

example 1
AssertionError: Test case is leaking async ops.
Before:
  - dispatched: 0
  - completed: 0
After:
  - dispatched: 5
  - completed: 4

Make sure to await all promises returned from Deno APIs before
finishing test case.
    at assert (deno:cli/rt/06_util.js:33:13)
    at asyncOpSanitizer (deno:cli/rt/40_testing.js:46:7)
    at async Object.resourceSanitizer [as fn] (deno:cli/rt/40_testing.js:70:7)
    at async TestRunner.[Symbol.asyncIterator] (deno:cli/rt/40_testing.js:245:13)
    at async Object.runTests (deno:cli/rt/40_testing.js:322:22)
    at async $deno$test.ts:3:1

failures:

        example 1

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (9ms)
```

The same issue applies if I make the test function syncronous by removing async key word and the await delay(0) from the end.

```ts
Deno.test("example 1", () => {
  let metrics = Deno.metrics();
  console.log(
    `
Before:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );

  setTimeout(() => {}, 10000);
  setTimeout(() => {}, 10001);

  metrics = Deno.metrics();
  console.log(
    `After:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );
});

Deno.test("example 2", () => {
  let metrics = Deno.metrics();
  console.log(
    `
Before:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );

  setTimeout(() => {}, 10000);
  setTimeout(() => {}, 10001);

  metrics = Deno.metrics();
  console.log(
    `After:
- dispatched: ${metrics.opsDispatchedAsync}
- completed: ${metrics.opsCompletedAsync}`,
  );
});
```

```sh
deno test example_test.ts
Check file:///mnt/c/Users/kylej/projects/deno/test_suite/$deno$test.ts
running 2 tests
test example 1 ... 
Before:
- dispatched: 0
- completed: 0
After:
- dispatched: 1
- completed: 0
FAILED (3ms)
test example 2 ... 
Before:
- dispatched: 3
- completed: 2
After:
- dispatched: 3
- completed: 2
ok (2ms)

failures:

example 1
AssertionError: Test case is leaking async ops.
Before:
  - dispatched: 0
  - completed: 0
After:
  - dispatched: 3
  - completed: 2

Make sure to await all promises returned from Deno APIs before
finishing test case.
    at assert (deno:cli/rt/06_util.js:33:13)
    at asyncOpSanitizer (deno:cli/rt/40_testing.js:46:7)
    at async Object.resourceSanitizer [as fn] (deno:cli/rt/40_testing.js:70:7)
    at async TestRunner.[Symbol.asyncIterator] (deno:cli/rt/40_testing.js:245:13)
    at async Object.runTests (deno:cli/rt/40_testing.js:322:22)
    at async $deno$test.ts:3:1

failures:

        example 1

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (5ms)
```

### LINKED ISSUE #10974: minimum timer value should depend on nesting level, per spec
https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers
"If nesting level is greater than 5, and timeout is less than 4, then set timeout to 4."

Currently we use 4ms minimum regardless of nesting level

### LINKED ISSUE #11398: Spurious assertion error when the callback to setInterval lasts longer than the interval
Attacler/Bart reported on Discord that an `AssertionError` was being thrown while they were stress-testing their code. It seems to happen spuriously, but their stress test seemed to hit it every time.

```js
const MET = setInterval(() => {
    // console.log("RET", worker.RET);
    if (worker.RET <= 0) {
        removeWorkers(executeId);
        events.emit("return_" + executeId, {
            errors: ["Maximum execution time passed!"],
            logs: [],
            error: true,
        });
        spawnWorkers(1);
    }
    worker.RET -= 0.1;
}, 100);

events.on("return_" + executeId, (data: MessageEvent) => {
    console.log("RETURN", executeId);
    if (MET != undefined) clearInterval(MET);
    console.log(data);
    res(data);

    setTimeout(() => {
        removeWorkers(executeId);
    }, 250);
    spawnWorkers(1);
});
```

```
error: Uncaught (in promise) AssertionError: Assertion failed.
            if (MET != undefined) clearInterval(MET);
                                  ^
    at assert (deno:extensions/timers/01_timers.js:22:13)
    at setGlobalTimeout (deno:extensions/timers/01_timers.js:332:5)
    at setOrClearGlobalTimeout (deno:extensions/timers/01_timers.js:376:7)
    at unschedule (deno:extensions/timers/01_timers.js:425:9)
    at clearTimer (deno:extensions/timers/01_timers.js:547:5)
    at clearInterval (deno:extensions/timers/01_timers.js:564:5)
    at file:///mnt/abc2f9c2-fcf1-457f-94e4-be464a8e61ea/Connectplatform/Deno%20Connect/helpers/workers.ts:233:35
    at EventEmitter.emit (https://deno.land/x/event@2.0.0/mod.ts:166:7)
    at Worker.worker.onmessage (file:///mnt/abc2f9c2-fcf1-457f-94e4-be464a8e61ea/Connectplatform/Deno%20Connect/helpers/workers.ts:110:28)
    at Worker.wrappedHandler (deno:runtime/js/01_web_util.js:26:37)
```

Since the failed assertion makes sure that the timeout is not lesser than zero, and changing the interval to 4000ms solved the problem, it seems to be caused by the callback taking longer to run than the interval.

andreubotella:
This change also makes the timers implementation closer to the spec, and sets up the stage to implement `AbortSignal.timeout()` (whatwg/dom#1032).

Fixes #8965.
Fixes #10974.
Fixes #11398.


ext/timers/lib.rs
```diff
@@ -74,12 +80,38 @@ where
   Ok(result)
 }
 
+pub struct TimerHandle(Rc<CancelHandle>);
+
+impl Resource for TimerHandle {
+  fn name(&self) -> Cow<str> {
+    "timerCancelHandle".into()
+  }
+
```
bartlomieju, line 89: This is a bit strange... why name it `timerCancelHandle` instead of just `timer`?

andreubotella: I renamed the struct and then forgot to change the `name` method.

Should it be just `timer`, not `timerHandle`?

bartlomieju: I think `timer` is sufficiently self-explanatory.

runtime/js/40_testing.js
```diff
@@ -30,6 +30,37 @@
...
+  }
+
+  function handleOpSanitizerDelayMacrotask() {
+    if (opSanitizerDelayResolve !== null) {
+      opSanitizerDelayResolve();
+      opSanitizerDelayResolve = null;
+    }
+    return true;
+  }
+
```
andreubotella, line 63: These changes to `40_testing.ts` are useful on their own, since they prevent op sanitizer failures on timers even in the current implementation. But they are needed for the CI to pass in this PR. Should this be split into its own PR?

bartlomieju: @andreubotella I think it would be preferable to create a separate PR for this change

andreubotella: Created it as #12908.

andreubotella:
#12908 and #12913 must be merged before this PR, since they fix the CI failures.

andreubotella:
It seems like the `httpConnAutoCloseDelayedOnUpgrade` and `httpServerDeleteRequestHasBody` tests are flaky with this implementation of timers – they occasionally fail with pending ops. Since it used to be that timers were clamped to a minimum of 4ms everywhere, and now they're clamped to 0ms when the timer nesting level is lower than 5, it might be that we need to set the op sanitizer delay to 4ms.

bartlomieju:
@lucacasonato please advise

lucacasonato:
> it might be that we need to set the op sanitizer delay to 4ms.

I don't think so. That sounds like a hack. The tests are probably not correct if they fail uf the timer is actually 0 sec (1 evt loop turn).

andreubotella:
I've added back the code in `runAfterTimeout` that waits to run the callback until all previous timers have resolved. Rather than an array from which entries are spliced out, I made it a linked list, but this might make the code less readable. This should unblock #12963 and perhaps #12953.

ext/timers/01_timers.js
```diff
@@ -61,534 +48,305 @@
...
-          found = node;
+        // 2.
+        // 3.
+        // TODO(@andreubotella): Error handling.
+        if (typeof callback === "function") {
+          FunctionPrototypeCall(callback, globalThis, ...args);
+        } else {
+          // TODO(@andreubotella): eval doesn't seem to have a primordial, but
+          // it can be redefined in the global scope.
+          (0, eval)(callback);
```
bartlomieju, line 156: So this will just parse and evaluate any string, seems like something we might not want to do.

CC @bnoordhuis what do you think?

ext/timers/01_timers.js
```diff
@@ -61,534 +48,305 @@
...
+    PromisePrototypeThen(
+      core.opAsync("op_sleep", millis, cancelRid),
+      () => {
+        // 2. Wait until any invocations of this algorithm that had the same
+        // global and orderingIdentifier, that started before this one, and
+        // whose milliseconds is equal to or less than this one's, have
+        // completed.
+        // 4. Perform completionSteps.
+
+        // Since the sleep ops aren't guaranteed to resolve in the right order,
```
bartlomieju, line 249: Please add `// IMPORTANT:` here as this is a crucial implementation detail

bartlomieju (commented):
@andreubotella thank you for the PR. It fixes a lot of long outstanding bugs which is great. At the same time it's touching a crucial piece of infrastructure so special care must be taken when reviewing and landing it. 

The implementation mostly looks good to me, I have a few question for bits that are not immediately obvious to me.

Before landing I'd like to get more eyes on it.

@bnoordhuis @lucacasonato please review

ext/timers/01_timers.js
andreubotella: This isn't new:

https://github.com/denoland/deno/blob/987716798fb3bddc9abc7e12c25a043447be5280/ext/timers/01_timers.js#L476-L480

ext/timers/01_timers.js
```diff
@@ -61,534 +48,307 @@
...
-        const cmp = this.#comparator(data, node.data);
-
-        dir = cmp > 0;
+    // 3. If the surrounding agent's event loop's currently running task is a
+    // task that was created by this algorithm, then let nesting level be the
+    // task's timer nesting level. Otherwise, let nesting level be zero.
+    // 4. If timeout is less than 0, then set timeout to 0.
+    // 5. If nesting level is greater than 5, and timeout is less than 4, then
+    // set timeout to 4.
+    timeout = MathMax(timeout, (timerNestingLevel > 5) ? 4 : 0);
```
bnoordhuis, line 138: Two comments:

1. 5 is a magic constant. Why not 4, 6 or 42? The comment doesn't explain it.

2. I find it harder to read than the equivalent `if` statement (which is shorter to boot):
```js
if (timerNestingLevel > 5 && timeout < 4) timeout = 4;
```

ext/timers/01_timers.js
```diff
@@ -61,534 +48,307 @@
...
+
+    // 9. Let task be a task that runs the following steps:
+    const task = {
+      action: () => {
+        // 1. If id does not exist in global's map of active timers, then abort
+        // these steps.
```
bnoordhuis, line 144: This comment is a little superfluous because it's obvious _what_ the code below does, just not _why_ it does it.

It'd be more helpful to explain that (IIUC) a timer is appended to the `timerTasks` array when created but not deleted from said array when it's disarmed before expiration, hence the necessity to have this side check.

ext/timers/01_timers.js
bnoordhuis: For better or worse, that's also what browsers do (but not Node.)

ext/timers/01_timers.js
```diff
@@ -61,534 +48,307 @@
...
-                gpc.left.red = false;
-                assert(gpc.right);
-                gpc.right.red = false;
-              }
-            }
+        // 4. If id does not exist in global's map of active timers, then abort
+        // these steps.
+        // NOTE: It might have been removed via the author code in handler
+        // calling clearTimeout() or clearInterval().
+        if (MapPrototypeHas(activeTimers, id)) {
```
bnoordhuis, line 164: As an optimization: always call `MapPrototypeDelete(activeTimers, id)` when `repeat === false`?
```js
if (repeat) {
  // ...
} else if (MapPrototypeDelete(activeTimers, id)) {
  core.tryClose(cancelRid);
}
```

ext/timers/01_timers.js
```diff
@@ -61,534 +48,307 @@
...
-    // I'm not if that's possible.
-    opStartGlobalTimer(timeout);
-    await opWaitGlobalTimer();
-    pendingEvents--;
-    prepareReadyTimers();
-  }
+  /**
+   * A doubly linked list of timers.
+   * @type { { head: ScheduledTimer | null, tail: ScheduledTimer | null } } */
+  const scheduledTimers = { head: null, tail: null };
```
bnoordhuis, line 213: You can get rid of all the null checks if you turn this into a list head:
```js
// ideally give the list head and timer objects the same object shape
// to keep property reads and writes monomorphic
const scheduledTimers = { head: null, tail: null };
scheduledTimers.head = scheduledTimers;
scheduledTimers.tail = scheduledTimers;

function isEmpty() {
  return scheduledTimers.tail === scheduledTimers;
}

function append(timer) {
  timer.prev = scheduledTimers.tail;
  timer.next = scheduledTimers;
  scheduledTimers.tail.next = timer;
  scheduledTimers.tail = timer;
}

function remove(timer) {
  timer.prev.next = timer.next;
  timer.next.prev = timer.prev;
  // timer.prev = timer.next = null;
}
```

ext/timers/01_timers.js
```diff
@@ -61,534 +48,307 @@
...
-    timer.scheduled = true;
-    // If the new timer is scheduled to fire before any timer that existed before,
-    // update the global timeout to reflect this.
-    if (globalTimeoutDue === null || globalTimeoutDue > timer.due) {
-      setOrClearGlobalTimeout(timer.due, now);
-    }
-  }
+    // 1.
+    PromisePrototypeThen(
+      core.opAsync("op_sleep", millis, cancelRid),
```
bnoordhuis, line 242: This means there's an op call per timer? To keep overhead down JS timers would ideally all hang off a single `tokio::time::sleep()`.

bnoordhuis (commented):
Thanks for the PR, Andreu. I can't really intuit how this affects performance. Do we have timer benchmarks?

ext/timers/01_timers.js
andreubotella: Note step 4 and the ternary operator. The equivalent would be:
```js
if (timeout < 0) timeout = 0;
if (timerNestingLevel > 5 && timeout < 4) timeout = 4;
```

ext/timers/01_timers.js
andreubotella: The trade-off here is either:

a) Having at most one pending timer op at a time, but having to call into Rust to start another op every time one timer expires.
b) Only calling into Rust when starting or canceling timers, but having multiple pending timer ops.

I haven't measured the difference between these, and due to the very nature of timers measuring performance isn't easy. There's the additional confounding factor of #10974 if we compare with the current timers implementation. So I'm not sure how to benchmark this.

In any case, unrefing timers would be much easier with b) – see #12953.

ext/timers/01_timers.js
andreubotella: > 1. 5 is a magic constant. Why not 4, 6 or 42? The comment doesn't explain it.
It's spec-mandated.

ext/timers/01_timers.js
bartlomieju: I agree that we should go with this approach, especially in light of ref/unref for timers.

ext/timers/01_timers.js
bartlomieju: @andreubotella do you want to address this bit before landing?

ext/timers/01_timers.js
bartlomieju:
```js
if (timeout < 0) timeout = 0;
if (timerNestingLevel > 5 && timeout < 4) timeout = 4;
```

This one is a bit easier to read, so maybe let's go with it?

> It's spec-mandated. I suspect due to compat concerns.

Right, can you add a comment or reference to the spec here?

cli/tests/unit/timers_test.ts
```diff
@@ -346,6 +401,26 @@ Deno.test(async function timerMaxCpuBug() {
...
+  setTimeout(() => {
+    array.push(2);
+    setTimeout(() => array.push(5));
+  }, 0);
+  setTimeout(() => {
+    array.push(3);
+    setTimeout(() => array.push(6));
+  }, 0);
+
+  await delay(100);
```
kt3k, line 419: This looks still a little bit flaky from my observation. ref: https://github.com/denoland/deno/runs/4418361252?check_suite_focus=true

How about creating a promise for each setTimeout call and waiting for all of them instead of delay(N)

bnoordhuis: Seems like a genuine bug. A race would be missing [4,5,6] but it's only missing 6. Merits deeper investigation.

ext/timers/01_timers.js
bnoordhuis: Node has regular JS timers hanging off a single C++ timer (N->1, common case) and unref'd JS timers backed by individual C++ timers (N->N, uncommon case.)

For Node, that's a worthwhile optimization because programs often have thousands of active timers, usually one or more per TCP connection.

For Deno, I'm not sure. TBH, I don't know how (or if) we deal with read/write timeouts or slowloris style attacks.

At any rate, it's an area worth investigating but it doesn't have to hold up this PR.

ext/timers/01_timers.js
bnoordhuis: Would be nice to have now and I'd definitely want to see it addressed eventually but I'm okay with landing this as-is.

bnoordhuis (approved):
LGTM if the test failure that @kt3k pointed out can be resolved.

ext/timers/01_timers.js
andreubotella: > Right, can you add a comment or reference to the spec here?

This comment is the spec text:

https://github.com/andreubotella/deno/blob/timers/ext/timers/01_timers.js#L131-L136

ext/timers/01_timers.js
andreubotella: That seems harder to reason through, especially if we go all the way and make `scheduledTimers` monomorphic, since `scheduledTimers.prev` would be the tail and `scheduledTimers.next` would be the head. And if we keep the property names `head` and `tail`, we'd still need checks to append and remove.

cli/tests/unit/timers_test.ts
andreubotella: According to the spec, a timer mustn't go off before other timers that were both scheduled before it and have a lesser or equal timeout. So the ordering of the numbers pushed into the array must hold, and the delay must expire after 3 is pushed, but the delay doesn't necessarily have to go off after any of the 4/5/6 timers do. And that's what I find when I test this – either [1,2,3], [1,2,3,4], [1,2,3,4,5] or [1,2,3,4,5,6], but not [1,2] or anything out of order.

Note that the spec doesn't say that a timeout of 0 must queue the action in the timer task queue immediately – step 5.3 in ["run steps after a timeout"](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#run-steps-after-a-timeout) allows for an arbitrary additional wait. And it seems like an `op_sleep` with a timeout of 0 doesn't necessarily resolve at the next event loop turn. Should we queue the task immediately in that case?

In any case, I'll be taking @kt3k's suggestion.

andreubotella: > Should we queue the task immediately in that case?

Come to think about it, no. This would let timers starve other parts of the event loop. Whereas with timers which are always backed by ops, the earliest that the tasks can be pushed to the timers queue is on the next event loop turn.

ext/timers/01_timers.js
bnoordhuis: If it's good enough for nginx, libuv, linux and the BSDs, then it's certainly good enough for deno. :-)

(Also node - but I'm reasonably certain I was the one who introduced that pattern there.)

cli/tests/unit/timers_test.ts
bnoordhuis: I don't follow your line of reasoning.

We have timers A, B and C that expire at time T+0, who arm three more timers D, E, and F that expire at time T+1 (or T+4, doesn't matter), and a delay G that expires at time T+100.

Under normal conditions, they expire in order A to G because 0 < 1 < 100.

Under abnormal conditions, say an event loop stall or an OS scheduler fluke that pauses the program until T+200, then either A-G or A-C, G, D-F are reasonable behavior.

But A-E, G, F? No, never.

<hr>

A timer's expiry time is set to `now() + timeout` rather than `event_loop_tick_start_time() + timeout`. Put another way: timers have no shared time base.

Under normal conditions that doesn't matter because computers are fast: `now()` doesn't change meaningfully in the time it takes to execute an event loop "tick".

But if there is a stall of duration N between B and C, then there is an equivalent gap between E and F: E expires at time T+1 but F at time T+N.

andreubotella: I don't know how tokio's timers are implemented –I have tried to follow the code and can't really make sense of it–, or how they interact with `FuturesUnordered` –which is probably more relevant for how they show up here. The spec requires that timers don't resolve before `now() + timeout` –which I'm assuming tokio does correctly–, and it also requires that a timer doesn't resolve before any timers that were scheduled before it and that have a lower or equal timeout. The code around `scheduledTimers` guarantees that that condition holds, but it doesn't guarantee anything else. And, purely in terms of the spec requirements, A-E, G, F is valid, because D-F are not required to be added to the timer task queue _the instant_ `setTimeout()` is called.

bnoordhuis: For posterity: Bartek and I talked this through OOB.

IMO, it's a quality-of-implementation issue - expiry order should follow the Principle of Least Surprise, regardless of how relaxed the spec is - but it's pre-existing, not a regression, and therefore doesn't need to hold up this PR.

bnoordhuis:
Thanks for the PR, Andreu!

PR #12866, opened 2021-11-22
feat: Add support for import assertions and JSON modules

### LINKED ISSUE #7623: Import Assertions/JSON Modules
[Import Assertions](https://github.com/tc39/proposal-import-assertions) has just been moved to Stage 3.  [JSON Modules](https://github.com/tc39/proposal-json-modules) is still at Stage 2 (and JSON Modules depends on Module Assertions).

I haven't found a tracking issue for v8.

TypeScript issue: microsoft/TypeScript#40694

cc/ @kdy1

bartlomieju:
Closes #7623 

This commit adds proper support for import assertions and JSON modules.

Implementation of "core/modules.rs" was changed to account for multiple possible
module types, instead of always assuming that the code is an "ES module". In
effect "ModuleMap" now has knowledge about each modules' type (stored via
"ModuleType" enum). Module loading pipeline now stores information about
expected module type for each request and validates that expected type matches
discovered module type based on file's "MediaType".

Relevant tests were added to "core/modules.rs" and integration tests,
additionally multiple WPT tests were enabled.

There are still some rough edges in the implementation and not all WPT were
enabled, due to: 
a) unclear BOM handling in source code by "FileFetcher" 
b) design limitation of Deno's "FileFetcher" that doesn't download the same 
module multiple times in a single run


dsherret:
Probably will need an swc upgrade once this is fixed https://github.com/swc-project/swc/issues/2802

bartlomieju:
Blocked by https://github.com/denoland/deno/issues/12989

lucacasonato (commented):
I can not get this to work:

```js
import x from "./.dlint.json" asserts { "type": "json" };
```

```
$ cargo run --example fs_module_loader test.js
Run test.js
cargo:rerun-if-changed=/mnt/starship/Projects/github.com/denoland/deno/core/00_primordials.js
cargo:rerun-if-changed=/mnt/starship/Projects/github.com/denoland/deno/core/01_core.js
cargo:rerun-if-changed=/mnt/starship/Projects/github.com/denoland/deno/core/02_error.js
Error: Uncaught SyntaxError: Unexpected identifier
    at file:///mnt/starship/Projects/github.com/denoland/deno/test.js:1:30
```

bartlomieju:
This is now working:
Screenshot_2021-12-15_at_02 39 54

I still need to add tests, but it's ready for initial review.

core/modules.rs
```diff
@@ -19,18 +21,142 @@ use std::collections::VecDeque;
...
+  import_assertions: v8::Local<v8::FixedArray>,
+  no_of_assertions: usize,
+) -> HashMap<String, String> {
+  let mut assertions: HashMap<String, String> = HashMap::default();
+
+  if no_of_assertions == 0 {
+    return assertions;
+  }
+
+  let assertions_per_line = import_assertions.length() / no_of_assertions;
```
bartlomieju, line 62: This should be an argument to the function instead of `no_of_assertions`

core/modules.rs
```diff
@@ -19,18 +21,142 @@ use std::collections::VecDeque;
...
+  let undefined = v8::undefined(tc_scope);
+  resolver.resolve(tc_scope, undefined.into());
+  Some(resolver.get_promise(tc_scope).into())
+}
+
+/// A type of module to be executed.
+///
+/// Currently only `JavaScript` variant is supported,
+/// but in future types like `JSON` and `WASM` will be
+/// supported.
```
bartlomieju, line 139: Remove

core/modules.rs
```diff
@@ -330,12 +490,19 @@ impl RecursiveModuleLoad {
...
         id
       }
-      None => self.module_map_rc.borrow_mut().new_module(
-        scope,
-        self.is_currently_loading_main_module(),
-        &module_source.module_url_found,
-        &module_source.code,
-      )?,
+      None => match module_source.module_type {
+        ModuleType::JavaScript => self.module_map_rc.borrow_mut().new_module(
```
bartlomieju, line 494: Rename to `new_es_module`

core/modules.rs
```diff
@@ -449,12 +645,22 @@ impl Stream for RecursiveModuleLoad {
   }
 }
 
+/// Describes what is the expected type of module, usually
+/// it's `ModuleType::JavaScript`, but if there were import assertions
+/// it might be `ModuleType::Json`.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct ModuleRequest {
+  pub specifier: ModuleSpecifier,
+  pub module_type: ModuleType,
```
bartlomieju, line 654: Maybe rename to `expected_module_type`

core/modules.rs
```diff
@@ -521,6 +734,65 @@ impl ModuleMap {
...
+    let tc_scope = &mut v8::TryCatch::new(scope);
+
+    let parsed_json = match v8::json::parse(tc_scope, source_str) {
+      Some(parsed_json) => parsed_json,
+      None => {
+        let message = v8::String::new(
+          tc_scope,
+          &format!("\"{}\" is not a valid JSON.", name),
+        )
+        .unwrap();
```
bartlomieju, line 755: In this case it might be worth to run it through `serde_json::from_string()` to get a better error pointing to faulty place in JSON

core/modules.rs
```diff
@@ -19,18 +21,142 @@ use std::collections::VecDeque;
...
+/// Throws V8 exception if assertions are invalid
+pub(crate) fn validate_import_assertions(
+  scope: &mut v8::HandleScope,
+  assertions: &HashMap<String, String>,
+) {
+  for (key, value) in assertions {
+    if key == "type" && !SUPPORTED_TYPE_ASSERTIONS.contains(&value.as_str()) {
+      let message = v8::String::new(
+        scope,
+        &format!("\"{}\" is not a valid module type.", value),
```
nayeemrmn, line 41: We should start using debug format in place of `\"{}\"`, it adds quotes with necessary escaping
```suggestion
        &format!("{:?} is not a valid module type.", value),
```
Ditto for other `\"{}\"`

andreubotella:
Some of the WPT tests at `html/semantics/scripting-1/the-script-element/json-module` are still failing.

core/modules.rs
dsherret: I don't think we should use debug format for the sake of quoting a string. If we want quotes to strings then it's better to be explicit about that IMO and for example do something like `#""{:?}" is not a valid moulde type."#`. Using debug format is less explicit, requires thinking more about how the value is going to look when using debug format, and can more easily cause issues when the type of the value changes.

For example, here's an example of an issue debug format has caused #13076

bartlomieju: Good point David, I will revert the change.

bartlomieju:
I have trouble getting 
```
"UTF-16BE BOM should result in parse error in JSON module script",
"UTF-16LE BOM should result in parse error in JSON module script"
```
to pass.

When I test it locally the proper error is returned, but somehow during WPT it breaks on some utility like `dprint` and surfaces `TypeError` instead of `SyntaxError`.

core/modules.rs
```diff
@@ -19,18 +22,145 @@ use std::collections::VecDeque;
...
+  assertions
+    .get("type")
+    .map(|ty| {
+      if ty == "json" {
+        ModuleType::Json
+      } else {
+        ModuleType::JavaScript
+      }
+    })
+    .unwrap_or(ModuleType::JavaScript)
```
lucacasonato, line 103:
```suggestion
  match assertions.get("type") {
    Some("json") => ModuleType::Json,
    _ => ModuleType::JavaScript,
  }
```

core/modules.rs
```diff
@@ -159,9 +293,20 @@ impl ModuleLoader for FsModuleLoader {
...
       })?;
+      let module_type = if let Some(extension) = path.extension() {
+        if extension == "json" {
+          ModuleType::Json
+        } else {
+          ModuleType::JavaScript
+        }
+      } else {
+        ModuleType::JavaScript
+      };
```
lucacasonato, line 304: Could you use a match statement?

lucacasonato (commented):
Looking good!

andreubotella:
The UTF-16 `TypeError` seems to be a [`std::string::FromUtf8Error`](https://doc.rust-lang.org/std/string/struct.FromUtf8Error.html) thrown when `get_source_from_bytes()` is called with `None` as charset. Note that with `Some("utf-8")`, the conversion would replace invalid UTF-8 with replacement characters, which would give the correct `SyntaxError`.

core/modules.rs
```diff
@@ -557,9 +831,28 @@ impl ModuleMap {
       let import_specifier = module_request
         .get_specifier()
         .to_rust_string_lossy(tc_scope);
+
+      let import_assertions = module_request.get_import_assertions();
+
+      // For static imports, assertions are triples of (keyword, value and source offset)
+      let assertions = parse_import_assertions(tc_scope, import_assertions, 3);
+
+      // FIXME(bartomieju): there are no stack frames if exception
```
dsherret, line 840: Is this something to be fixed in this review or later?

dsherret (commented):
Looks really good! Will be really nice to have this.

core/modules.rs
```diff
@@ -19,18 +22,145 @@ use std::collections::VecDeque;
...
+      scope.throw_exception(exception);
+      return;
+    }
+  }
+}
+
+pub(crate) fn parse_import_assertions(
+  scope: &mut v8::HandleScope,
+  import_assertions: v8::Local<v8::FixedArray>,
+  assertions_per_line: usize,
```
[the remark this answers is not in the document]
bartlomieju: Done

core/bindings.rs
```diff
@@ -294,13 +311,29 @@ pub extern "C" fn host_import_module_dynamically_callback(
...
+            v8::Exception::type_error(scope, message.try_into().unwrap())
+          }
+          "SyntaxError" => {
+            v8::Exception::syntax_error(scope, message.try_into().unwrap())
+          }
+          "ReferenceError" => {
+            v8::Exception::reference_error(scope, message.try_into().unwrap())
+          }
+          _ => v8::Exception::error(scope, message.try_into().unwrap()),
+        };
```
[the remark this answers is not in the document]
bartlomieju: This is only used in a single place, and the whole function seems like a code smell, I'd rather address this in separate PR that would cleanup modules implementation

core/modules.rs
```diff
@@ -159,9 +293,20 @@ impl ModuleLoader for FsModuleLoader {
           module_specifier
         ))
       })?;
+      let module_type = if let Some(extension) = path.extension() {
+        if extension == "json" {
```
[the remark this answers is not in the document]
bartlomieju: I don't think so

core/modules.rs
bartlomieju: Later, it's not clear how to fix it

bartlomieju:
> The UTF-16 `TypeError` seems to be a [`std::string::FromUtf8Error`](https://doc.rust-lang.org/std/string/struct.FromUtf8Error.html) thrown when `get_source_from_bytes()` is called with `None` as charset. ~If called with `Some("utf-8")`, the conversion would replace invalid UTF-8 with replacement characters, which would give the correct `SyntaxError`.~
> 
> Edit: That last sentence is wrong, I completely misread the code 😅

Thanks for pointing this out, it's not clear to me what would need to change to fix it. @dsherret any suggestions?

core/modules.rs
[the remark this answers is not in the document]
bartlomieju: Actually I think you are right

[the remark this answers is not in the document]
dsherret: Maybe we should add a test for `somefile.JSON`?

core/modules.rs
bartlomieju: Types do not match; I think the current version is fine.

bartlomieju:
Due to the fact that Deno caches HTTP responses on first use, I believe we can't enable `"Two modules of different type with the same specifier can load if the server changes its responses"` test; as it changes the response's content type on second fetch.

andreubotella:
So according to the HTML spec, the charset declared on the MIME type, or the file's BOM, don't really matter. Instead, all modules –JS, JSON or CSS– are decoded with the Encoding standard's "UTF-8 decode" algorithm (see ["fetch a single module script"](https://html.spec.whatwg.org/multipage/webappapis.html#fetch-a-single-module-script), step 13), which is equivalent to:

```rust
let bytes = if bytes.starts_with(&[0xEF, 0xBB, 0xBF]) {
    &bytes[3..]
} else {
    bytes
};
String::from_utf8_lossy(bytes)
```

I suspect windows-1252 and perhaps UTF-16 might be important for local files in Windows, so we might break the spec there, but the WPT tests expect the JSON parsing to fail, not the UTF-8 decoding.

bartlomieju:
@andreubotella seems like a niche case - we still fail as expected, just with wrong error type. I'd prefer to address this in a follow up PR.

andreubotella:
> @andreubotella seems like a niche case - we still fail as expected, just with wrong error type. I'd prefer to address this in a follow up PR.

Fair enough

core/modules.rs
```diff
@@ -19,18 +22,145 @@ use std::collections::VecDeque;
...
+}
+
+pub(crate) fn parse_import_assertions(
+  scope: &mut v8::HandleScope,
+  import_assertions: v8::Local<v8::FixedArray>,
+  assertions_per_line: usize,
+) -> HashMap<String, String> {
+  let mut assertions: HashMap<String, String> = HashMap::default();
+
+  let no_of_assertions = import_assertions.length() / assertions_per_line;
```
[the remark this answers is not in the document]
bartlomieju: This is actually pretty bad :D this assertion fails, because in `module_resolve_callback` we can get either static or dynamic module (so either 3 or 2 assertions per line). However regular division works as expected. I'm not yet sure how to fix this properly as in the `module_resolve_callback` we don't know if we're in dynamic context or not.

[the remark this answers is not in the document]
bartlomieju: Actually it seems Chrome always assumes 3 items per line in `module_resolve_callback` so that's what I'm going to use.

dsherret (approved):
LGTM

PR #12867, opened 2021-11-23
fix(cli): config file should resolve paths relative to the config file

### LINKED ISSUE #12863: Deno configuration file paths should be resolved relative to config file rather than cwd
I think this was an oversight, but I believe the configuration file paths should be resolved based on the config file rather than the current working directory. Ex. paths in "include", "exclude"

dsherret:
Previously it was using the cwd.

Closes #12863

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -301,68 +301,46 @@ fn ts_json_to_diagnostics(
     .collect()
 }
 
-// Returns `ConfigSnapshot::root_uri` in the correct format.
-fn get_root_specifier(
-  snapshot: &language_server::StateSnapshot,
-) -> Option<ModuleSpecifier> {
-  let root = snapshot.config.root_uri.as_ref()?;
-  let root = root.to_file_path().ok()?;
-
-  // FIXME: `root_uri` from `ConfigSnapshot` are without a trailing slash,
-  // so `Url::join` treats the root as a file, not a directory, and erases the folder name.
-  // To fix that behaviour we just parsing `root_uri` again.
-  ModuleSpecifier::from_directory_path(root).ok()
-}
-
 // Filters documents according to the `include` and the `exclude` lists (from `StateSnapshot::maybe_lint_config`).
 // If a document is in the `exclude` list - then it be removed.
 // If the `include` list is not empty, and a document is not in - then it be removed too.
-fn filter_documents(
+fn filter_lint_documents(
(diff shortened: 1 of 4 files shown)


kitsonk (approved):
LGTM

cli/lsp/diagnostics.rs
```diff
@@ -301,68 +301,46 @@ fn ts_json_to_diagnostics(
...
-      .filter_map(|i| root_uri.join(i).ok())
-      .collect()
-  };
-
-  let ignore_specifiers = join_specifiers(&linter_config.files.exclude);
-  let include_specifiers = join_specifiers(&linter_config.files.include);
-
   documents.retain(|doc| {
-    let path = doc.specifier().path();
+    let path = if let Ok(file_path) = doc.specifier().to_file_path() {
```
dsherret, line 317: The tests are hanging on Windows because of this if expression returning false. That's happening because the tests are using file specifiers that the `url` crate doesn't resolve to a `Path` though I wish it did (ex. `file:///test.ts`... I opened https://github.com/servo/rust-url/issues/730). I see a couple options:

1. We add a new `to_file_path(specifier: &ModuleSpecifier)` function that does more of an attempt to convert `file:///something` to `PathBuf` on Windows.
2. We change all the tests to use platform specific paths (seems error prone... would rather the code would just handle it on Windows too).

kitsonk: I thought we had something like `to_file_path()` that failed friendly on Windows, I will try to dig it out.

dsherret: I did some more searching and didn't find anything so I added a `specifier_to_file_path` function just now.

cli/fs_util.rs
```diff
@@ -628,4 +670,22 @@ mod tests {
...
   }
+
+  #[test]
+  fn test_specifier_to_file_path() {
+    run_success_test("file:///", "/");
+    run_success_test("file:///test", "/test");
+    run_success_test("file:///dir/test/test.txt", "/dir/test/test.txt");
+    run_success_test(
+      "file:///dir/test%20test/test.txt",
+      "/dir/test test/test.txt",
```
bartlomieju, line 681: Are URLs like `file://C:/path/to/file` handled correctly?

bartlomieju (approved):
LGTM too

cli/fs_util.rs
dsherret: Yeah, the url crate seems to handle them on both windows and unix.

PR #12868, opened 2021-11-23
chore(lsp): fix possible race condition with tests expecting 3 publishDiagnostics messages, but getting 2 or 1

dsherret:
The tests expect 3 publish notifications. It was possible for less than 3 to occur if two or more tasks set the diagnostics in the collection, exited the lock at the same time, then called `publish_diagnostics`.

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -589,10 +589,9 @@ async fn generate_deps_diagnostics(
 /// Publishes diagnostics to the client.
 async fn publish_diagnostics(
   client: &lspower::Client,
-  collection: Arc<Mutex<DiagnosticCollection>>,
+  collection: &mut DiagnosticCollection,
   snapshot: &language_server::StateSnapshot,
 ) {
-  let mut collection = collection.lock().await;
   if let Some(changes) = collection.take_changes() {
     for specifier in changes {
       let mut diagnostics: Vec<lsp::Diagnostic> =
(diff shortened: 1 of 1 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -646,13 +645,12 @@ async fn update_diagnostics(
         error!("Error generating lint diagnostics: {}", err);
       })
       .unwrap_or_default();
-    {
-      let mut collection = collection.lock().await;
-      for diagnostic_record in diagnostics {
-        collection.set(DiagnosticSource::DenoLint, diagnostic_record);
-      }
```
dsherret, line 653: Race condition was right here where the lock is released and then the lock was acquired in `publish_diagnostics`.

cli/lsp/diagnostics.rs
```diff
@@ -646,13 +645,12 @@ async fn update_diagnostics(
...
+
+    let mut collection = collection.lock().await;
+    for diagnostic_record in diagnostics {
+      collection.set(DiagnosticSource::DenoLint, diagnostic_record);
     }
-    publish_diagnostics(client, collection, &snapshot).await;
+    publish_diagnostics(client, &mut collection, &snapshot).await;
```
dsherret, line 653: @kitsonk would it be better to pass the diagnostics directly to this function? It seems like they get set in the collection and then they're taken from there. Maybe they shouldn't be put in the collection if lint isn't enabled or the specifier isn't enabled?

kitsonk: The considerations is that if you toggle a project as enabled or disabled (or for linting) the only way to clear diagnostics is to publish an empty set for each file.

dsherret: Ok, I'm not going to bother with this for now. I'll just fix this one issue in this PR.

kitsonk (approved):
LGTM

kitsonk:
I have seen some situations recently with the 1.16 LSP where the diagnostics are "behind" the current version of the document and never catch up.  I am wondering if this racy condition was part of it, so I will be trying that out and if not, will investigate further.

dsherret:
> I have seen some situations recently with the 1.16 LSP where the diagnostics are "behind" the current version of the document and never catch up. I am wondering if this racy condition was part of it, so I will be trying that out and if not, will investigate further.

@kitsonk hmmm I'm not sure and don't think it would affect that. I was thinking some of the diagnostics code looks suspect, but I haven't looked at it in detail (only glanced at it). I think we could do something better here to improve it and eliminate that problem. Maybe I will look at it in more detail tomorrow.

PR #12870, opened 2021-11-23
chore: upgrade Rust version to `1.56.1`

abdfnx:
> upgrade Rust version to **1.56.1** in

1. `.devcontainer/Dockerfile`
2. `.github/workflows/ci.yml`


dsherret:
Looks like the CI keeps saying it's out of space. Maybe the cache needs to be cleared somehow when doing this upgrade? I'm not sure how to do that on the CI.

abdfnx:
ok, so is the problem is from me or from the CI ?

bartlomieju:
> ok, so is the problem is from me or from the CI ?

I pushed a commit that rotates cache keys, let's see if it helps.

abdfnx:
> > ok, so is the problem is from me or from the CI ?
> 
> I pushed a commit that rotates cache keys, let's see if it helps.

looks like you fixed it

dsherret (approved):
LGTM. Thanks @abdfnx!

And thanks Bartek for busting the cache :)

abdfnx:
> LGTM. Thanks @abdfnx!
> 
> And thanks Bartek for busting the cache :)

glad to help @denoland

`deno` one of the best tools that I used ever

PR #12873, opened 2021-11-23
fix(lsp): normalize urls in did_change_watched_files

### LINKED ISSUE #12871: lsp: config file does not reload on change
When I change `deno.json` - nothing happens, so, I have to restart the language server every time I change the file.
As I found out - the issue seems are Windows-specific, here is `dbg!` output:
```
[cli\lsp\language_server.rs:959] config_uri = Url {
    scheme: "file",
    cannot_be_a_base: false,
    username: "",
    password: None,
    host: None,
    port: None,
    path: "/C:/Users/me/Desktop/deno-project/deno.json",
    query: None,
    fragment: None,
}
[cli\lsp\language_server.rs:960] &fe.uri = Url {
    scheme: "file",
    cannot_be_a_base: false,
    username: "",
    password: None,
    host: None,
    port: None,
    path: "/c%3A/Users/me/Desktop/deno-project/deno.json",
    query: None,
    fragment: None,
}
[cli\lsp\language_server.rs:961] *config_uri == fe.uri = false
```

I believe the comparison does not work as expected because of an encoded path that comes from a language client.

Version: `1.16.2 (debug, x86_64-pc-windows-msvc)`
Tested on VSCode.

igorsaux:
Fixes #12871

I've noticed that the rest of the functions handle URLs using `normalize_url`, but that was missing in `did_change_watched_files`.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -930,9 +930,15 @@ impl Inner {
       .performance
       .mark("did_change_watched_files", Some(&params));
     let mut touched = false;
+    let changes: Vec<Url> = params
+      .changes
+      .iter()
+      .map(|f| self.url_map.normalize_url(&f.uri))
+      .collect();
+
     // if the current import map has changed, we need to reload it
     if let Some(import_map_uri) = &self.maybe_import_map_uri {
-      if params.changes.iter().any(|fe| *import_map_uri == fe.uri) {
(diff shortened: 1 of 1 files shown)


dsherret (approved):
LGTM. Thanks @igorsaux for reporting this problem and figuring out the solution! 🙂

Issue #12758, opened 2021-11-14
Rework "fmt_check_tests_dir" test


bartlomieju:
This test goes through all test directories and adding new files to `cli/tests` requires to update the test definition to either ignore more files or update ".out" file.

PR #12875, opened 2021-11-23
tests: remove 'fmt_check_tests_dir' test

bartlomieju:
Closes https://github.com/denoland/deno/issues/12758

This test is not really valuable after we reorganized tests into `cli/tests/testdata` directories, adding new files that might be badly formatted on purpose requires to update the test case.


dsherret (approved):
LGTM. Yeah, this test was kind of annoying.

PR #12876, opened 2021-11-23
fix(lsp): lsp should respect include/exclude files in format config

### LINKED ISSUE #12865: Format config should respect include/excludes in LSP
Similar to #12608, it looks like the formatting include/exclude configuration is not respected in the LSP.

### LINKED ISSUE #12877: lsp: excluding a file from linting does not clear diagnostics
Reproduction:
 - Create and open a file that contains any warnings or errors from deno, let's it be `a.ts`, for example.
 - Add the file to the exclude list in `deno.json`:
 ```json
"lint": {
    "files": {
      "exclude": ["a.ts"]
    }
  }
 ```

Config re-applies but diagnostics from `a.ts` do not disappear.
If we edit `deno.json` before opening the file - it works as should.

dsherret:
Also changes it so the LSP sticks with module specifiers in the config file. I ran into an issue on CI with Mac where it was canonicalizing the file path of the config file and causing issues because now the config file was at a "different place" relative to potentially open files that aren't necessarily on the file system.

Closes #12865
Closes #12877


cli/tests/integration/lsp_tests.rs
```diff
@@ -1420,6 +1421,61 @@ fn lsp_format_mbc() {
   shutdown(&mut client);
 }
 
+#[test]
+fn lsp_format_exclude() {
+  let temp_dir = TempDir::new().expect("could not create temp dir");
```
dsherret, line 1426: This is failing on mac because the canonicalized version of TempDir is different. I think we need to change everything related to the config file and lsp to just use specifiers as it's not possible to canonicalize some documents that are open, but not saved to the file system. I am looking into this now.

dsherret: Should be fixed on the Mac CI now in ee84932e4b47267de18c8d53b22f1a4f2f253699 by using module specifiers in the config file.

cli/main.rs
```diff
@@ -217,7 +218,10 @@ pub fn create_main_worker(
     }
   } else if let Some(config_file) = &ps.maybe_config_file {
     // otherwise we will use the path to the config file
-    config_file.path.to_str().map(|s| s.to_string())
+    specifier_to_file_path(&config_file.specifier)
+      .ok()
+      .map(|p| p.to_str().map(|s| s.to_string()))
+      .flatten()
```
dsherret, line 224: It seems like it would be better to just do `config_file.specifier.to_string()` here? I didn't want to introduce a breaking change on this key though.

bartlomieju: That change seems reasonable 👍

bartlomieju (approved):
LGTM

cli/main.rs
dsherret: Ok. Done in 1eaf50ab30e9e2f4ea4a95ac687fc79092383d12

It is cleaner.

Issue #12872, opened 2021-11-23
compat: Dynamic import from CJS module fails


bartlomieju:
Reported by @jack-works in https://github.com/denoland/deno/issues/12577#issuecomment-976714138

```js
// node_modules/pure-cjs/index.js
async function run() {
    const result = await import('pure-esm')
}
run()
```

```js
// index.js
require('pure-cjs')
```

```
deno run --compat --unstable --allow-read --no-check ./index.js
error: Uncaught (in promise) TypeError: Relative import path "pure-esm" not prefixed with / or ./ or ../ from "file:///dev/deno-compat-test/node_modules/pure-cjs/index.js"
    const result = await import('pure-esm')
                   ^
```

This is the same problem as in https://github.com/denoland/deno/issues/12628 - if we're executing module via `require` then the module graph is not constructed and module resolution falls back to the "regular" ESM resolution (ie. `NodeEsmResolver` is not being used). 

Related to https://github.com/denoland/deno/issues/12684

Issue #12628, opened 2021-11-02
REPL should support compat mode


ry:
```
# deno -V
deno 1.15.3
# deno repl --compat --unstable
Download https://deno.land/std@0.113.0/node/global.ts
Download https://deno.land/std@0.113.0/node/process.ts
Download https://deno.land/std@0.113.0/node/timers.ts
Download https://deno.land/std@0.113.0/node/_options.ts
Download https://deno.land/std@0.113.0/node/_validators.ts
Download https://deno.land/std@0.113.0/node/internal_binding/node_options.ts
Check https://deno.land/std@0.113.0/node/global.ts
Check https://deno.land/std@0.113.0/node/global.ts
Deno 1.15.3
exit using ctrl+d or close()
> require('fs')
Uncaught ReferenceError: require is not defined
    at <anonymous>:2:1
```

bartlomieju:
I'm looking into it, but first we need to address https://github.com/denoland/deno/issues/12684

Issue #12684, opened 2021-11-08
Introduce an aggregate resolver


kitsonk:
Just a thought and maybe not for this review, but I wonder if we should have some other `AggregateResolver` or something that takes a collection of `dyn Resolver`s and then tries to resolve with each resolver in order. That way in the consuming code someone could build one up and it might be a bit more clear about what's going on.

_Originally posted by @dsherret in https://github.com/denoland/deno/pull/12631#discussion_r744332751_

kitsonk:
@bartlomieju I think this would help with the "compat" resolution/resolver as well to allow it to support import maps.

bartlomieju:
This is an interesting idea, but I'm not sure if it's feasible to go through the resolvers one by one. As an example, let's discuss current implementation of `NodeEsmResolver`. This resolver internally owns `ImportMapResolver` and here's a relevant code that actually implements resolution:
```rust
impl Resolver for NodeEsmResolver {
  fn resolve(
    &self,
    specifier: &str,
    referrer: &ModuleSpecifier,
  ) -> Result<ModuleSpecifier, AnyError> {
    // First try to resolve using import map, ignoring any errors
    if !specifier.starts_with("node:") {
      if let Some(import_map_resolver) = &self.maybe_import_map_resolver {
        if let Ok(specifier) = import_map_resolver.resolve(specifier, referrer)
        {
          return Ok(specifier);
        }
      }
    }

    let node_resolution =
      node_resolve(specifier, referrer.as_str(), &std::env::current_dir()?);

    match node_resolution {
      Ok(specifier) => {
        // If node resolution succeeded, return the specifier
        Ok(specifier)
      }
      Err(err) => {
        // If node resolution failed, check if it's because of unsupported
        // URL scheme, and if so try to resolve using regular resolution algorithm
        if err
          .to_string()
          .starts_with("[ERR_UNSUPPORTED_ESM_URL_SCHEME]")
        {
          return deno_core::resolve_import(specifier, referrer.as_str())
            .map_err(|err| err.into());
        }

        Err(err)
      }
    }
  }
}
```

[...]

PR #12882, opened 2021-11-24
compat: support compat mode in REPL

bartlomieju:
This commit introduces "ProcState::maybe_resolver" field, which
stores a single instance of resolver for the whole lifetime of the
process, instead of creating these resolvers for each creation
of module graph. As a result, this resolver can be used in fallback
case where graph is not constructed (REPL, loading modules using
"require") unifying resolution logic.

Closes https://github.com/denoland/deno/issues/12628
Closes https://github.com/denoland/deno/issues/12872

### DIFF
--- a/cli/proc_state.rs
+++ b/cli/proc_state.rs
@@ -167,6 +167,7 @@ pub struct Inner {
   pub broadcast_channel: InMemoryBroadcastChannel,
   pub shared_array_buffer_store: SharedArrayBufferStore,
   pub compiled_wasm_module_store: CompiledWasmModuleStore,
+  maybe_resolver: Option<Arc<dyn deno_graph::source::Resolver + Send + Sync>>,
 }
 
 impl Deref for ProcState {
@@ -313,6 +314,32 @@ impl ProcState {
       .clone()
       .or_else(|| env::var("DENO_UNSTABLE_COVERAGE_DIR").ok());
 
+    // FIXME(bartlomieju): `NodeEsmResolver` is not aware of JSX resolver
+    // created below
+    let node_resolver = NodeEsmResolver::new(
+      maybe_import_map.clone().map(ImportMapResolver::new),
+    );
+    let maybe_import_map_resolver =
+      maybe_import_map.clone().map(ImportMapResolver::new);
+    let maybe_jsx_resolver = maybe_config_file
+      .as_ref()
+      .map(|cf| {
+        cf.to_maybe_jsx_import_source_module()
+          .map(|im| JsxResolver::new(im, maybe_import_map_resolver.clone()))
+      })
+      .flatten();
+    let maybe_resolver: Option<Arc<dyn deno_graph::source::Resolver + Send + Sync>> = if flags.compat {
+      Some(Arc::new(node_resolver))
+    } else if let Some(jsx_resolver) = maybe_jsx_resolver {
+      // the JSX resolver offloads to the import map if present, otherwise uses
+      // the default Deno explicit import resolution.
+      Some(Arc::new(jsx_resolver))
+    } else if let Some(import_map_resolver) = maybe_import_map_resolver {
+      Some(Arc::new(import_map_resolver))
+    } else {
+      None
+    };
+
     Ok(ProcState(Arc::new(Inner {
       dir,
       coverage_dir,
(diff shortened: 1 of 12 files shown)


bartlomieju:
Ref https://github.com/denoland/deno/issues/12684 as it's somewhat needed to address this problem

cli/proc_state.rs
```diff
@@ -637,18 +649,26 @@ impl ProcState {
...
+    // FIXME(bartlomieju): this is a hacky way to provide compatibility with REPL
+    // and `Deno.core.evalContext` API. Ideally we should always have a referrer filled
+    // but sadly that's not the case due to missing APIs in V8.
     let referrer = if referrer.is_empty() && self.flags.repl {
-      deno_core::DUMMY_SPECIFIER
+      deno_core::resolve_url_or_path("./$deno$repl.ts").unwrap()
     } else {
-      referrer
+      deno_core::resolve_url_or_path(referrer).unwrap()
     };
```
bartlomieju, line 659: I'm not sure if I'm happy about this solution, but it's required because `deno_graph::source::Resolver` requires `&Url` type for `referrer` arg

cli/tests/testdata/compat/dyn_import_reject.out
```diff
@@ -1,2 +1,2 @@
-TypeError: Cannot load module "file:///[WILDCARD]/testdata/compat/foobar.js".
+TypeError: [ERR_MODULE_NOT_FOUND] Cannot find module '[WILDCARD]/testdata/compat/foobar.js' imported from file:///[WILDCARD]/testdata/compat/dyn_import_reject.js
```
bartlomieju, line 1: These changes are most likely related to how referrer is parsed in `cli/proc_state.rs`

bartlomieju:
> Ref #12684 as it's somewhat needed to address this problem

I kicked can down the road - this still needs to be addressed, but with this PR I think it's a bit more obvious how to tackle this "aggregate resolver" situation.

cli/proc_state.rs
```diff
@@ -395,30 +425,12 @@ impl ProcState {
...
-      maybe_import_map_resolver
-        .as_ref()
-        .map(|im| im.as_resolver())
-    };
+    let maybe_resolver: Option<&dyn deno_graph::source::Resolver> =
+      if let Some(resolver) = &self.maybe_resolver {
+        Some(resolver.as_ref())
+      } else {
+        None
+      };
```
bnoordhuis, line 433: Longhand version of saying this?
```suggestion
    let maybe_resolver: Option<&dyn deno_graph::source::Resolver> =
      self.maybe_resolver.as_ref();
```
(Or maybe `.as_deref()`. Likewise around line 661.)

cli/proc_state.rs
```diff
@@ -637,18 +649,26 @@ impl ProcState {
...
-        .resolve(specifier, referrer)
-        .map_err(|err| err.into())
+
+    let maybe_resolver: Option<&dyn deno_graph::source::Resolver> =
+      if let Some(resolver) = &self.maybe_resolver {
+        Some(resolver.as_ref())
+      } else {
+        None
+      };
+    if let Some(resolver) = &maybe_resolver {
```
bnoordhuis, line 667: It's not wholly clear to me why it needs to be converted to `&dyn Resolver` first.

bnoordhuis (approved):

cli/proc_state.rs
bartlomieju: I tried that before but it doesn't work:
```
   --> cli/proc_state.rs:429:7
    |
428 |     let maybe_resolver: Option<&dyn deno_graph::source::Resolver> =
    |                         ----------------------------------------- expected due to this
429 |       self.maybe_resolver.as_ref();
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait object `dyn Resolver`, found struct `Arc`
    |
    = note: expected enum `std::option::Option<&dyn Resolver>`
               found enum `std::option::Option<&Arc<dyn Resolver + std::marker::Send + std::marker::Sync>>`

```

I had to add additional trait boundaries (`Send + Sync`) to the `dyn Resolver` and found that it's the only method to get it past type checking.

cli/proc_state.rs
bartlomieju: I'm not sure what you mean by converting - but we're using `dyn Resolver` to have a single interface from `deno_graph`; that can be used in `prepare_module_load` method and `resolve` methods on the `ProcState`.

dsherret (approved):
LGTM

PR #12884, opened 2021-11-24
chore: speed up compat tests

bartlomieju:
This commit speeds up compat tests by using local copy of "deno_std"
instead of downloading it from https://deno.land for each test.
Additionally type checking is skipped.

Depends on https://github.com/denoland/deno/pull/12883


bnoordhuis (approved):
Good idea!

I didn't find an open issue but the integration::run::blob_gc_finalization hang seems like a flake.

Issue #12745, opened 2021-11-12
Add support "unref"ing pending ops


bartlomieju:
Notes from discussion with @piscisaureus

While working on Node compatibility layer I hit a problem of missing polyfills for numerous Node APIs that "unref" resources, eg. [`subprocess.unref()`](https://nodejs.org/api/child_process.html#subprocessunref). 

Currently Deno has a concept of "ref" and "unref" ops, however there's no way to change from one op to the other. In order to proceed with the Node compat we need to fix this situation.

A plan of action looks as follows (note all these changes are to be done in `core/runtime.rs` module):
- add a new field `JsRuntime::unrefedOps`, which is a `HashSet` that will store `promiseId`s of "unrefed" ops
- when checking if there are pending ops to determine if `JsRuntime`'s event loop should keep going, instead of checking if `JsRuntime::pending_ops` is empty, check if the size of `JsRuntime::pending_ops` is greater than the size of `Jsruntie::unrefedOps`:
https://github.com/denoland/deno/blob/28dbb4a95e569772600ae28ae4fe8b6fa2a4d2af/core/runtime.rs#L800
- when an op resolves/rejects, its `promiseId` would be removed from the `JsRuntime::unrefedOps`
- on the JavaScript side we shall add a field like `#pendingPromiseIds` to relevant resources (eg. `Deno.Process.#pendingPromiseIds`), that would track which `promiseId`s are associated with the given resources (or in other words store ids of all pending ops related to the resource)
- when calling `.unref()` on a certain resource we would call into Rust and add the promise ids to the `JsRuntime::unrefedOps`, ie. calling `Deno.Process.unref()` would call into Rust and pass `Deno.Process.#pendingPromiseIds`, which would be inserted into the `JsRuntime::unrefedOps` hash set

The work could be split into three parts:
1. remove distinction between "ref" and "unref" ops:
https://github.com/denoland/deno/blob/28dbb4a95e569772600ae28ae4fe8b6fa2a4d2af/core/runtime.rs#L1519-L1537
That means rewriting the functionality to instead use `JsRuntime::unrefedOps`. "unref" ops are currently only used in `Deno.addSignalListener` API
2. Since this functionality is currently needed for the `child_process` polyfill in `deno_std/node`, the first pass would be to add `ref()` and `unref()` method to the `Deno.Process`
3. Add `unref()` to other resources one-by-one as needed

CC @bnoordhuis

PR #12889, opened 2021-11-24
feat(core): Add ability to "ref" and "unref" pending ops

bartlomieju:
This commit adds an ability to "ref" or "unref" pending ops.

Up to this point Deno had a notion of "async ops" and "unref async ops";
the former keep event loop alive, while the latter do not block event loop
from finishing. It was not possible to change between op types after 
dispatching, one had to decide which type to use before dispatch.

Instead of storing ops in two separate "FuturesUnordered" collections,
now ops are stored in a single collection, with supplemental "HashSet"
storing ids of promises that were "unrefed".

Two APIs were added to "Deno.core":
- "Deno.core.refOp(promiseId)" which allows to mark promise id 
to be "refed" and keep event loop alive (the default behavior)
- "Deno.core.unrefOp(promiseId)" which allows to mark promise 
id as "unrefed" which won't block event loop from exiting

Closes https://github.com/denoland/deno/issues/12745

TODO:
- [x] metrics for unrefed ops (I think that unrefing ops should decrement `ops_dispatched_async` and increment `ops_dispatched_async_unref`, though it seems overly complicated)
- [x] cleanup handling of finished ops (remove `AsyncOpIterator`?)
- [x] add `id` field to promises returned from `Deno.core.opcallAsync`
- [x] add `Deno.core.ref(...promiseIds)` and `Deno.core.unref(...promiseIds)`

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -134,12 +134,7 @@
 
   function opAsync(opName, arg1 = null, arg2 = null) {
     const promiseId = nextPromiseId++;
-    const maybeError = opcallAsync(
-      opsCache[opName],
-      promiseId,
-      arg1,
-      arg2,
-    );
+    const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
     const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -141,7 +141,10 @@
     );
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
-    return PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    // Save the id on the promise so it can be latered ref'ed or unref'ed
+    p.id = promiseId;
+    return p;
   }
 
   function opSync(opName, arg1 = null, arg2 = null) {
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -44,6 +44,8 @@
   const RING_SIZE = 4 * 1024;
   const NO_PROMISE = null; // Alias to null is faster than plain nulls
   const promiseRing = ArrayPrototypeFill(new Array(RING_SIZE), NO_PROMISE);
+  // TODO(bartlomieju): it future use `v8::Private` so it's not visible
+  // to users. Currently missing bindings.
   const promiseIdSymbol = SymbolFor("Deno.core.internalPromiseId");
 
   function setPromise(promiseId) {
@@ -138,7 +140,7 @@
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
     const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
-    // Save the id on the promise so it can be latered ref'ed or unref'ed
+    // Save the id on the promise so it can later be ref'ed or unref'ed
     p[promiseIdSymbol] = promiseId;
     return p;
   }
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -23,6 +23,7 @@
     MapPrototypeSet,
     PromisePrototypeThen,
     ObjectAssign,
+    SymbolFor,
   } = window.__bootstrap.primordials;
 
   // Available on start due to bindings.
@@ -43,6 +44,7 @@
   const RING_SIZE = 4 * 1024;
   const NO_PROMISE = null; // Alias to null is faster than plain nulls
   const promiseRing = ArrayPrototypeFill(new Array(RING_SIZE), NO_PROMISE);
+  const promiseIdSymbol = SymbolFor("Deno.core.internalPromiseId");
 
   function setPromise(promiseId) {
     const idx = promiseId % RING_SIZE;
@@ -130,20 +132,19 @@
     return res;
   }
 
-  function opAsync(opName, arg1 = null, arg2 = null, unref = false) {
+  function opAsync(opName, arg1 = null, arg2 = null) {
     const promiseId = nextPromiseId++;
     const maybeError = opcallAsync(
       opsCache[opName],
       promiseId,
       arg1,
       arg2,
-      unref,
     );
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
     const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
     // Save the id on the promise so it can be latered ref'ed or unref'ed
-    p.id = promiseId;
+    p[promiseIdSymbol] = promiseId;
     return p;
   }
 
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -130,9 +130,15 @@
     return res;
   }
 
-  function opAsync(opName, arg1 = null, arg2 = null) {
+  function opAsync(opName, arg1 = null, arg2 = null, unref = false) {
     const promiseId = nextPromiseId++;
-    const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
+    const maybeError = opcallAsync(
+      opsCache[opName],
+      promiseId,
+      arg1,
+      arg2,
+      unref,
+    );
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
     return PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -375,6 +375,7 @@ fn opcall_sync<'s>(
     b,
     op_id,
     promise_id: 0,
+    unref: false,
   };
   let op = OpTable::route_op(op_id, state.op_state.clone(), payload);
   match op {
@@ -431,6 +432,7 @@ fn opcall_async<'s>(
   // Deserializable args (may be structured args or ZeroCopyBuf)
   let a = args.get(2);
   let b = args.get(3);
+  let unref = args.get(4).is_true();
 
   let payload = OpPayload {
     scope,
@@ -438,6 +440,7 @@ fn opcall_async<'s>(
     b,
     op_id,
     promise_id,
+    unref,
   };
   let op = OpTable::route_op(op_id, state.op_state.clone(), payload);
   match op {
@@ -451,6 +454,10 @@ fn opcall_async<'s>(
     Op::Async(fut) => {
       state.op_state.borrow().tracker.track_async(op_id);
       state.pending_ops.push(fut);
+      if unref {
+        // FIXME(bartlomieju): this should track using `track_unref`?
+        state.unref_ops.insert(promise_id);
+      }
       state.have_unpolled_ops = true;
     }
     Op::AsyncUnref(fut) => {
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -375,7 +375,6 @@ fn opcall_sync<'s>(
     b,
     op_id,
     promise_id: 0,
-    unref: false,
   };
   let op = OpTable::route_op(op_id, state.op_state.clone(), payload);
   match op {
@@ -440,7 +439,6 @@ fn opcall_async<'s>(
     b,
     op_id,
     promise_id,
-    unref,
   };
   let op = OpTable::route_op(op_id, state.op_state.clone(), payload);
   match op {
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -36,6 +36,12 @@ lazy_static::lazy_static! {
         function: opcall_sync.map_fn_to()
       },
       v8::ExternalReference {
+        function: ref_ops.map_fn_to()
+      },
+      v8::ExternalReference {
+        function: unref_ops.map_fn_to()
+      },
+      v8::ExternalReference {
         function: set_macrotask_callback.map_fn_to()
       },
       v8::ExternalReference {
@@ -151,6 +157,8 @@ pub fn initialize_context<'s>(
   // Bind functions to Deno.core.*
   set_func(scope, core_val, "opcallSync", opcall_sync);
   set_func(scope, core_val, "opcallAsync", opcall_async);
+  set_func(scope, core_val, "refOps", ref_ops);
+  set_func(scope, core_val, "unrefOps", unref_ops);
   set_func(
     scope,
     core_val,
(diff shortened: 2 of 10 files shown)


core/01_core.js
```diff
@@ -130,9 +130,15 @@
...
+  function opAsync(opName, arg1 = null, arg2 = null, unref = false) {
     const promiseId = nextPromiseId++;
-    const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
+    const maybeError = opcallAsync(
+      opsCache[opName],
+      promiseId,
+      arg1,
+      arg2,
+      unref,
+    );
```
bartlomieju, line 141: This change was done so that ops that are unref by default don't have to call into Rust immediately after dispatch to mark them as "unrefed"

core/runtime.rs
```diff
@@ -1533,19 +1534,17 @@ impl JsRuntime {
...
-        cx,
-      };
-      for (promise_id, op_id, resp) in ops {
-        op_state.borrow().tracker.track_unref_completed(op_id);
-        args.push(v8::Integer::new(scope, promise_id as i32).into());
-        args.push(resp.to_v8(scope).unwrap());
+      // FIXME(bartlomieju): this should be done in the loop before,
+      // but that makes borrow checker unhappy, think how to refactor it
+      for promise_id in remove_from_unref {
+        state.unrefed_ops.remove(&promise_id);
```
bartlomieju, line 1547: Clean this up

core/01_core.js
```diff
@@ -130,12 +130,21 @@
...
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
-    return PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    // Save the id on the promise so it can be latered ref'ed or unref'ed
+    p.id = promiseId;
```
piscisaureus, line 146: Have you considered using a symbol here or a reverse WeakMap?
(I am not sure what that would do for performance, we'd have to measure that)

core/lib.deno_core.d.ts
```diff
@@ -21,6 +21,18 @@ declare namespace Deno {
       b?: any,
     ): Promise<any>;
 
+    /** Mark following promises as "ref", ie. event loop won't exit
+     * until they are resolved. All async ops are "ref" by default. */
+    function refOps(
+      ...promiseIds: number[]
```
piscisaureus, line 27: Why does it take multiple `promiseId`s?
I do understand that it is faster to coalesce multiple `refOp` calls into a single call, but I wonder how common that is, and also note that passing a single int into the binding layer is faster than passing an array.

core/01_core.js
bartlomieju: I did not, but using a symbol might be preferable, that way we can better communicate that users shouldn't use that value. I'll explore that.

core/runtime.rs
```diff
@@ -171,7 +155,7 @@ pub(crate) struct JsRuntimeState {
   dyn_module_evaluate_idle_counter: u32,
   pub(crate) js_error_create_fn: Rc<JsErrorCreateFn>,
   pub(crate) pending_ops: FuturesUnordered<PendingOpFuture>,
-  pub(crate) pending_unref_ops: FuturesUnordered<PendingOpFuture>,
+  pub(crate) unrefed_ops: HashSet<u64>,
```
piscisaureus, line 158: 👍

piscisaureus (commented):
Looks pretty good, some questions though.

AaronO:
Do we want all op(calls) to be unrefable ? I feel like it should be opt-in and take its distinct path (e.g: `opUnref` instead of overloading `opAsync`)

core/lib.deno_core.d.ts
bartlomieju: In some resources we might have multiple pending ops associated with it. So instead of crossing boundary multiple times I thought it's better to do it in one go. Keep in mind that this is spread operator so you're passing multiple integers and not an array (`Deno.core.refOps(1, 2, 3, 4)`). Also is there a situation where this API would be used in a hot path?

piscisaureus:
> Do we want all op(calls) to be unrefable ? I feel like it should be opt-in and take its distinct path (e.g: `opUnref` instead of overloading `opAsync`)

I don't see a reason why we would be more restrictive about this than necessary.

AaronO:
> > Do we want all op(calls) to be unrefable ? I feel like it should be opt-in and take its distinct path (e.g: `opUnref` instead of overloading `opAsync`)
> 
> I don't see a reason why we would be more restrictive about this than necessary.

Well if in practice there's only a very small subset of ops/opcalls that may be unref'd then I think it's cleaner to have its own distinct implementation path so as not to overload the simple async opcall path

core/lib.deno_core.d.ts
piscisaureus: > Keep in mind that this is spread operator so you're passing multiple integers and not an array (Deno.core.refOps(1, 2, 3, 4))

I see, in the binding layer you're just iterating over `FunctionCallbackArguments`, that's pretty efficient and I have no concerns about it.

> In some resources we might have multiple pending ops associated with it.

You'll find that this almost never happens in practice but oh well.

bartlomieju: If you prefer to have it accept single ID that is fine for me as well. Your call

core/01_core.js
piscisaureus: I probably would keep the old signature. The only "unref-by-default" ops are signal listeners and they're not a performance concern.

bartlomieju: Reverted.

core/01_core.js
bartlomieju: Used `Symbol.for("Deno.core.internalPromiseId")` instead, let's see if performance is better or worse

bartlomieju:
> > > Do we want all op(calls) to be unrefable ? I feel like it should be opt-in and take its distinct path (e.g: `opUnref` instead of overloading `opAsync`)
> > 
> > 
> > I don't see a reason why we would be more restrictive about this than necessary.
> 
> Well if in practice there's only a very small subset of ops/opcalls that may be unref'd then I think it's cleaner to have its own distinct implementation path so as not to overload the simple async opcall path

I agree with Bert, I also believe it makes it more confusing to have distinct type of op for that purpose. Besides having different API there would be no upside IMO as we won't be able to validate if you are unrefing an op of that type. I'll be happy to revisit the topic after we add `.ref()`/`.unref()` to several APIs.

AaronO:
> > > > Do we want all op(calls) to be unrefable ? I feel like it should be opt-in and take its distinct path (e.g: `opUnref` instead of overloading `opAsync`)
> > > 
> > > 
> > > I don't see a reason why we would be more restrictive about this than necessary.
> > 
> > 
> > Well if in practice there's only a very small subset of ops/opcalls that may be unref'd then I think it's cleaner to have its own distinct implementation path so as not to overload the simple async opcall path
> 
> I agree with Bert, I also believe it makes it more confusing to have distinct type of op for that purpose. Besides having different API there would be no upside IMO as we won't be able to validate if you are unrefing an op of that type. I'll be happy to revisit the topic after we add `.ref()`/`.unref()` to several APIs.

I'm not proposing to have a different kind of op rust-side just that the opCall JS side is explicit (killing `op_async_unref` is good IMO).

> > Do we want all op(calls) to be unrefable ? I feel like it should be opt-in and take its distinct path (e.g: `opUnref` instead of overloading `opAsync`)
> 
> I don't see a reason why we would be more restrictive about this than necessary.

It all depends on our call expectations but IMO:
1. `Deno.core.opAsync("op_foo", A, B, unref=true/false)` makes most sense if `unref` is derived from end-user args/inputs
2. `Deno.core.opUnref("op_foo", A, B)` makes more sense if the implementor decides if the opcall is unrefable 

From my understanding `unref=...` shouldn't really be derived from user inputs, it will mostly be hardcoded by the implementor (`ext/*/*.js` code)

If that's the case I think `Deno.core.opUnref` is the better choice, it explicits the intention and is easier to search/scan.

I also think it's more ergonomic, the previous `opAsync()` sig maps quite naturally to the function sig, one could you view it as a curried version of the opcall or `Deno.core.ops.foo_bar(A, B)` (which would be the API if we did Fast API), tagging an op-layer control value to the end of the args feels out of place

bartlomieju:
> I'm not proposing a different kind of op rust-side just that the opCall JS side is explicit (killing op_async_unref is good IMO).

Okay, I misunderstood then. @AaronO I removed `unref` argument in `Deno.core.opAsync`; now you need to manually call `Deno.core.unrefOps()` afterwards. Is that okay for you?

bartlomieju:
Also I'd appreciate some feedback on how to tackle metrics - right now there are two uncalled APIs, this needs to be addressed before landing.

core/ops_metrics.rs
```diff
@@ -85,12 +85,14 @@ impl OpsTracker {
     metrics.ops_completed_async += 1;
   }
 
+  #[allow(unused)]
   pub fn track_unref(&self, id: OpId) {
```
piscisaureus, line 89: Just delete this, I guess?
@AaronO any opinions?

core/01_core.js
```diff
@@ -132,10 +134,18 @@
 
   function opAsync(opName, arg1 = null, arg2 = null) {
     const promiseId = nextPromiseId++;
-    const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
+    const maybeError = opcallAsync(
```
piscisaureus, line 137: Unnecessary change now.

piscisaureus (approved):
LGTM, let's land this and see what it does to the benchmarks.

core/ops_metrics.rs
bartlomieju: I guess that's fine. Right now "unrefed" ops don't take part in `opSanitizer` but actually it might be preferable to include them.

AaronO (approved):
LGTM

core/01_core.js
```diff
@@ -43,6 +44,7 @@
   const RING_SIZE = 4 * 1024;
   const NO_PROMISE = null; // Alias to null is faster than plain nulls
   const promiseRing = ArrayPrototypeFill(new Array(RING_SIZE), NO_PROMISE);
+  const promiseIdSymbol = SymbolFor("Deno.core.internalPromiseId");
```
bnoordhuis, line 47: Non-blocking but as a possible future improvement: you can use a `v8::Private` if you're worried about the symbol leaking out to user land code (it's visible to `Object.getOwnPropertySymbols()` after all.)

Private symbols are like regular symbols except more... er, private. (Great explanation, Ben. +1, would read again.)

core/bindings.rs
```diff
@@ -453,17 +461,56 @@ fn opcall_async<'s>(
...
 
+fn ref_op<'s>(
+  scope: &mut v8::HandleScope<'s>,
+  args: v8::FunctionCallbackArguments,
+  _rv: v8::ReturnValue,
+) {
+  let state_rc = JsRuntime::state(scope);
+  let mut state = state_rc.borrow_mut();
+
+  let promise_id = match v8::Local::<v8::Integer>::try_from(args.get(0))
```
bnoordhuis, line 478: Minor caveat: the way try_from() is currently implemented for v8::Integer means it only works for numbers in the range -2**31..2**32. I suppose that's fine (I see you're using i32 elsewhere) but I'd be remiss if I didn't point it out.

`Value::integer_value()` and `Integer::value()` return i64s so you could be forgiven for assuming try_from() works for all numbers that fit in an i64, but unfortunately it currently doesn't.

edit: opened https://github.com/denoland/rusty_v8/issues/839

core/ops_metrics.rs
```diff
@@ -11,10 +11,12 @@ pub struct OpMetrics {
   pub ops_dispatched: u64,
   pub ops_dispatched_sync: u64,
   pub ops_dispatched_async: u64,
+  // TODO(bartlomieju): this field is never updated
   pub ops_dispatched_async_unref: u64,
   pub ops_completed: u64,
   pub ops_completed_sync: u64,
   pub ops_completed_async: u64,
+  // TODO(bartlomieju): this field is never updated
```
bnoordhuis, line 19: What is the reason for not removing them? Are they considered part of a public or stable(ish) API?

core/runtime.rs
bnoordhuis, line 158: Should that be i64? You're casting `v8::Integer::value()` to u64 but that can give surprising results for negative numbers.

(Not really surprising, of course, just probably not what you want.)

core/runtime.rs
```diff
@@ -801,7 +785,7 @@ impl JsRuntime {
     let mut state = state_rc.borrow_mut();
     let module_map = module_map_rc.borrow();
 
-    let has_pending_ops = !state.pending_ops.is_empty();
+    let has_pending_ops = state.pending_ops.len() > state.unrefed_ops.len();
```
bnoordhuis, line 788: Suggestion: rename this to e.g. `has_pending_refed_ops` or something like that? Makes it a little clearer that the logic here is "if more pending ops than unref'd ops exist, we have ref'd ops."

core/01_core.js
```diff
@@ -135,7 +137,10 @@
     const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
-    return PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    // Save the id on the promise so it can be latered ref'ed or unref'ed
```
bnoordhuis, line 141:
```suggestion
    // Save the id on the promise so it can later be ref'ed or unref'ed
```

bnoordhuis (approved):
Nice work!

core/01_core.js
bartlomieju: Is that construct available in JS land?

core/bindings.rs
bartlomieju: I suggest to handle that in a follow up PR; we got a bunch of places using the same conversion.

core/ops_metrics.rs
bartlomieju: Yes, this are part of stable `Deno.Metrics` interface

core/runtime.rs
bartlomieju: Sure, that makes more sense, maybe even `i32` as there are some other conversions going on in the code.

core/runtime.rs
bartlomieju: Ok

core/01_core.js
bnoordhuis: Not yet, but I can add it. You may have seen node's getHiddenValue() and setHiddenValue() - it's the same mechanism.

bartlomieju: > Not yet, but I can add it. You may have seen node's getHiddenValue() and setHiddenValue() - it's the same mechanism.

I haven't seen it. I guess I'll add a TODO and we can change it once the API is added.

core/bindings.rs
```diff
@@ -453,17 +461,56 @@ fn opcall_async<'s>(
...
+    .map_err(Error::from)
+  {
+    Ok(promise_id) => promise_id,
+    Err(err) => {
+      throw_type_error(scope, format!("invalid promise id: {}", err));
+      return;
+    }
+  };
+
+  state.unrefed_ops.insert(promise_id);
```
andreubotella, line 511: Can we check that `promise_id` is in `state.pending_ops`?

bartlomieju: No, we'd have to iterate over all `pending_ops` killing the performance.

bartlomieju: Oh wait, this is `bindings.rs`, it's not a hot path. I guess I can add a check here.

bartlomieju: Actually not possible to do that, there's no way to inspect `promise_id` when iterating `pending_ops`

bartlomieju: This might be possible once @AaronO moves promise ring to Rust (#12314)

andreubotella: In the meantime, what probably can be done is change the `unrefOp` JS function to check whether the promise ID exists and hasn't resolved, right?

It would be pretty bad if some careless code added an already-resolved promise ID to `unrefed_ops`, since all that matters is the length compared to `pending_ops`. But I suppose CI would catch that, though.

bartlomieju: > In the meantime, what probably can be done is change the unrefOp JS function to check whether the promise ID exists and hasn't resolved, right?

That would require to adding a wrapper around `Deno.core.unrefOp` in `core/01_core.js` just to verify the promise map.

I'm not sure if it's worth doing that at this moment; I think we should try this API in the wild and see if that's a problem.

PR #12894, opened 2021-11-25
chore: upgrade lspower to 1.4.0

bartlomieju:
## merge on approval

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -457,7 +457,7 @@ impl From<ScriptElementKind> for lsp::CompletionItemKind {
   fn from(kind: ScriptElementKind) -> Self {
     match kind {
       ScriptElementKind::PrimitiveType | ScriptElementKind::Keyword => {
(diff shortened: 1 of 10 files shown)


cli/lsp/tsc.rs
```diff
@@ -2669,14 +2669,17 @@ pub enum SignatureHelpTriggerKind {
   Invoked,
   #[serde(rename = "retrigger")]
   Retrigger,
+  #[serde(rename = "unknown")]
+  Unknown,
```
bartlomieju, line 2673: Please note this change, I had to add a "catch-all" in the match below, due to `SignatureHelpTriggerKind` being non-exhaustive enum

See https://github.com/gluon-lang/lsp-types/issues/213

kitsonk: We need to check what impact it would have on tsc if we were to serialise that value, but we can do it as a followup.

kitsonk (approved):
LGTM

PR #12897, opened 2021-11-25
fix(cli): fix slow test, unbreak ci

### LINKED ISSUE #12887: integration::run::blob_gc_finalization is flaky
Seen in multiple runs:
```
test integration::run::blob_gc_finalization has been running for over 60 seconds
```
Unclear when it started being flaky or why.

bnoordhuis:
Reduce the number of iterations from 1,024 to 128.

On my big bruiser of a desktop machine it already takes up close to a
minute to complete when nothing else is running so no way it's going to
finish in the allotted time on the CI.

The fact that the test used to pass may be indicative of a performance
regression somewhere but it's not clear to me when or where that would
have been introduced.

Fixes #12887.


bartlomieju (approved):
LGTM

PR #12898, opened 2021-11-25
refactor(cli): simplify lint/format resolver logic

ah-yu:


dsherret (approved):
LGTM—it's cleaner. Thanks @ah-yu!

PR #12901, opened 2021-11-26
feat(wasm): Support `WebAssembly.instantiateStreaming` for file fetches

andreubotella:
Fetching of local files, added in #12545, returns a response with no headers, including the `Content-Type` header. This currently makes it not work with the WebAssembly streaming APIs, which require the response to have a content type of `application/wasm`.

Since the only way to obtain a `Response` object with a non-empty `url` field is via `fetch()`, this change changes the content type requirement to only apply to responses whose url has the `file:` scheme.

### DIFF
--- a/cli/tests/unit/wasm_test.ts
+++ b/cli/tests/unit/wasm_test.ts
@@ -77,6 +77,20 @@ Deno.test(async function wasmInstantiateStreaming() {
 });
 
 Deno.test(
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM, but would like to get @lucacasonato opinion on it.

lucacasonato (approved):

PR #12904, opened 2021-11-26
refactor: add deno_fetch::Options for init

ry:
[`deno_fetch::init`](https://docs.rs/deno_fetch/0.50.0/deno_fetch/fn.init.html) has a lot of parameters and generic on two types that keeps expanding over time. This refactor adds `deno_fetch::Options` struct for more clearly defining the various parameters.

I'm doing this because in a follow up PR, I want to add yet another configuration option to `init()`.


bartlomieju (approved):
LGTM

Issue #1651, opened 2019-02-02
Add color and/or emojis to permission prompt to make them more obvious


ry:

kitsonk:
Maybe...
```
🦕requests network access to "listen". Grant? [yN]
```

PR #12908, opened 2021-11-27
fix(test): Improve reliability of `deno test`'s op sanitizer with timers

andreubotella:
Although not easy to replicate in the wild, the `deno test` op sanitizer can fail when there are intervals that started before a test runs, since the op sanitizer can end up running in the time between the timer op for an interval's run resolves and the op for the next run starts.

This change fixes that by adding a new macrotask callback that will run after the timer macrotask queue has drained. This ensures that there is a timer op if there are any timers which are unresolved by the time the op sanitizer runs.

### DIFF
--- a/cli/tests/integration/test_tests.rs
+++ b/cli/tests/integration/test_tests.rs
@@ -151,6 +151,11 @@ itest!(ops_sanitizer_unstable {
   output: "test/ops_sanitizer_unstable.out",
 });
 
+itest!(ops_sanitizer_timeout_failure {
+  args: "test test/ops_sanitizer_timeout_failure.ts",
+  output: "test/ops_sanitizer_timeout_failure.out",
+});
+
 itest!(exit_sanitizer {
   args: "test test/exit_sanitizer.ts",
   output: "test/exit_sanitizer.out",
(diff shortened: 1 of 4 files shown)


andreubotella:
This is needed to fix some of the CI failures in #12862

andreubotella:
@bartlomieju, @caspervonb PTAL

caspervonb (commented):
LGTM!

bartlomieju (approved):
LGTM!

I verified that provided test case fails on `main`. Good fix @andreubotella

kt3k:
This change seems breaking some test cases in `node/_util/_util_callbackify_test.ts` in deno_std. ref: #1651

kt3k:
Reproduces the issue with a smaller example. It seems related to nextTick:
```ts
import { nextTick } from "https://deno.land/std@0.116.0/node/_next_tick.ts";
Deno.test("test 1", async () => {
  await new Promise<void>((resolve) => nextTick(resolve));
});
Deno.test("test 2", async () => {
  await new Promise<void>((resolve) => nextTick(resolve));
});
```
The above example is stuck at the end of `test 2`

Issue #7013, opened 2020-08-11
onunhandledrejection for globalThis?


nzakas:
I was looking for an equivalent to the Node.js "unhandledrejection" event in Deno, but it appears there isn't one on `globalThis`.

There does appear to be an `onunhandledrejection` event handler [defined on workers](https://github.com/denoland/deno/blob/master/cli/dts/lib.webworker.d.ts#L5468), though.

Is there an equivalent event outside of workers, and if not, could there be? :smile:

kitsonk:
We should have it, it is part of the HTML Living Standard: https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection

nzakas:
Maybe I'm testing for it wrong? Here's what I tried:

```js
globalThis.onunhandledrejection = function() {
    console.error("Unhandled rejection");
};

new Promise(() => {
    throw new Error("Error");
});
```

lucacasonato:
I think @kitsonk means we should add it, not we should already have it :-)

nzakas:
@lucacasonato oh gotcha 😆

andreespirela:
@lucacasonato Wouldn't this mean that if added, we will now have the option to stop Deno from dying in uncaught exceptions?

lucacasonato:
Uncaught exceptions are not the same as uncaught rejections. This is for when you don't define a .catch on a promise.

andreespirela:
> Uncaught exceptions are not the same as uncaught rejections. This is for when you don't define a .catch on a promise.

Got it, didn't notice it was "rejection", Thanks

Skillz4Killz:
Is there any update/eta on this? This is an absolutely crucial feature to have to prevent crashes.

lucacasonato:
@Skillz4Killz This will not be implemented to *prevent* crashes. If implemented it will only be able to be used to perform synchronous cleanup of allocated resources (e.g. file descriptors, handles, etc). It will not be usable to prevent shutdown on an unhandled rejection. Instead you should catch promise rejections correctly.

Skillz4Killz:
`Deno will always die on unhandled exceptions / rejections.`

I'd like to request the ability to change this for uncaught promises.

[...]

PR #12910, opened 2021-11-27
feat(core): intercept unhandled promise rejections

bnoordhuis:
Provide a programmatic means of intercepting rejected promises without a
.catch() handler. Needed for Node compat mode.

Also do a first pass at uncaughtException support because they're
closely intertwined in Node. It's like that Frank Sinatra song:
you can't have one without the other.

Stepping stone for #7013.

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -48,6 +48,12 @@ lazy_static::lazy_static! {
         function: set_nexttick_callback.map_fn_to()
       },
       v8::ExternalReference {
+        function: set_promise_reject_callback.map_fn_to()
+      },
+      v8::ExternalReference {
+        function: set_uncaught_exception_callback.map_fn_to()
+      },
+      v8::ExternalReference {
         function: run_microtasks.map_fn_to()
       },
       v8::ExternalReference {
@@ -170,6 +176,18 @@ pub fn initialize_context<'s>(
     core_val,
     "setNextTickCallback",
     set_nexttick_callback,
+  );
+  set_func(
+    scope,
+    core_val,
+    "setPromiseRejectCallback",
+    set_promise_reject_callback,
+  );
+  set_func(
+    scope,
+    core_val,
+    "setUncaughtExceptionCallback",
+    set_uncaught_exception_callback,
   );
   set_func(scope, core_val, "runMicrotasks", run_microtasks);
   set_func(scope, core_val, "hasTickScheduled", has_tick_scheduled);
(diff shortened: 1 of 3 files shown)


core/lib.deno_core.d.ts
```diff
@@ -115,5 +115,23 @@ declare namespace Deno {
...
+
+    /**
+     * Set a callback that will be called when an exception isn't caught
+     * by any try/catch handlers. Currently only invoked when the callback
+     * to setPromiseRejectCallback() throws an exception but that is expected
+     * to change in the future.
+     */
+    function setUncaughtExceptionCallback(
+      cb: (err: any) => void,
+    );
```
bnoordhuis, line 135: I realize now the function prototypes are white lies because I made them return the old handlers. I'll update before merging.

core/bindings.rs
```diff
@@ -320,30 +338,89 @@ pub extern "C" fn host_initialize_import_meta_object_callback(
...
-      state.pending_promise_exceptions.remove(&promise_global);
+
+    if tc_scope.has_caught() {
+      // If we get here, an exception was thrown by the unhandledRejection
+      // handler and there is ether no uncaughtException handler or the
+      // handler threw an exception of its own.
+      //
+      // TODO(bnoordhuis) Node terminates the process or worker thread
+      // but we don't really have that option. The exception won't bubble
+      // up either because V8 cancels it when this function returns.
```
bartlomieju, line 395: Maybe add `pending_exception` field to `JsRuntime` and check it in `JsRuntime::poll_event_loop` to terminate the process?

core/bindings.rs
```diff
@@ -320,30 +338,89 @@ pub extern "C" fn host_initialize_import_meta_object_callback(
...
+      eprintln!("Unhandled exception: {}", exception);
     }
-    v8::PromiseRejectEvent::PromiseRejectAfterResolved => {}
-    v8::PromiseRejectEvent::PromiseResolveAfterResolved => {
-      // Should not warn. See #1272
+  } else {
+    let promise = message.get_promise();
+    let promise_global = v8::Global::new(scope, promise);
+
+    match message.get_event() {
```
bartlomieju, line 407: Please add a comment that this logic is duplicated in `bindings::promise_reject_callback`

bartlomieju (approved):

core/bindings.rs
bnoordhuis: I thought about that but because of the time window in between (stuff can change, hard to reason about), I'm inclined to leave it alone until the uncaught exception logic gets fleshed out more.

core/bindings.rs
bnoordhuis: I'm not sure I follow. This _is_ `bindings::promise_reject_callback`.

bartlomieju: Sorry I misread.

core/bindings.rs
bartlomieju: Sounds reasonable

hashemisa:
In the new version, this line is mentioned:

feat(core): intercept unhandled promise rejections (#12910)

https://github.com/denoland/deno/releases

But there is no explanation for it on the blog and it does not work! 

https://deno.com/blog/v1.17

crowlKats:
@hashemisa this is for internal use, not for users to use

hashemisa:
https://github.com/denoland/deno/issues/7013

Issue #12888, opened 2021-11-24
Calling `Deno.exit` inside top level `deno test` isolates exits the process


lucacasonato:
```
~/p/g/d/deno ❯❯❯ cat test.ts
Deno.test("fail", () => {
  throw new Error("fail");
});

self.onunload = () => {
  Deno.exit(0);
};

~/p/g/d/deno ❯❯❯ deno test test.ts
Check file:///Users/lucacasonato/projects/github.com/denoland/deno/test.ts
running 1 test from file:///Users/lucacasonato/projects/github.com/denoland/deno/test.ts
test fail ... FAILED (9ms)
~/p/g/d/deno ❯❯❯ echo $status
0
```

Instead of actually exiting we should record the exit code, then terminate the isolate, and then mark the isolate as either passed or failed based on if the status code is 0, or something else.

PR #12911, opened 2021-11-27
feat(runtime): add op_set_exit_code

bnoordhuis:
Set the exit code to use if none is provided to Deno.exit(), or when
Deno exits naturally.

Needed for process.exitCode Node compat. Paves the way for #12888.

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -93,6 +93,7 @@ use std::iter::once;
 use std::path::PathBuf;
 use std::pin::Pin;
 use std::rc::Rc;
+use std::sync::atomic::Ordering::Relaxed;
 use std::sync::Arc;
 
 fn create_web_worker_callback(ps: ProcState) -> Arc<CreateWebWorkerCb> {
@@ -1475,4 +1476,7 @@ pub fn main() {
   logger::init(flags.log_level);
 
   unwrap_or_exit(run_basic(get_subcommand(flags)));
+
+  let code = deno_runtime::EXIT_CODE.load(Relaxed);
+  std::process::exit(code);
 }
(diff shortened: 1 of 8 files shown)


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -67,6 +67,18 @@ declare namespace Deno {
...
+   *
+   * Set the exit code to use when Deno exits naturally or when
+   * `Deno.exit()` is called without an argument.
+   *
+   * ```ts
+   * Deno.setExitCode(42);
+   * Deno.exit(); // Exits with status code 42.
+   * ```
+   */
+  export function setExitCode(code: number): void;
```
bartlomieju, line 80: I'm not yet convinced we want to expose it as `Deno` API; maybe let's go with "hidden op" approach we discussed last time? Ie. there's an op to set it, but we don't expose it as `Deno` API right now.

bnoordhuis: It doesn't really matter when it's behind --unstable, I think?

At any rate, I can't seem to call `Deno.core.opSync("op_set_exit_code", 42)` from std/node. It doesn't show up in `Deno.core.ops()` (whereas e.g. `op_exit` does) and when I invoke it, I get this:
```
Uncaught TypeError: invalid op id: expected type `v8::data::Integer`, got `v8::data::Value`
````
(Which is a terribly misleading error message, by the way.)

bnoordhuis: Okay, belay that - it works after a full rebuild. I'll change it if you feel strongly but :shrug:

bartlomieju (approved):
Okay, sure, let's go with unstable API. LGTM

bnoordhuis:
Decided to use an op after all. :-)

PR #12921, opened 2021-11-28
feat(crypto): support `importKey` in SPKI format

yacinehmito:
Work towards https://github.com/denoland/deno/issues/11690.

This is a good milestone. Now, there is a key format that can be both imported and exported for RSA. This wasn't the case before, and it was a real blocker to [implement an authorization server in Deno](https://twitter.com/yacinehmito/status/1463260017906364425?s=20).


yacinehmito:
cc @littledivy @lucacasonato

yacinehmito:
There was a lot of copy/pasting involved. Shall I refactor and introduce a more generic function for key imports, at least for RSA?

yacinehmito:
I'm investigating why the WPT are passing locally but [failing in the CI](https://github.com/denoland/deno/runs/4345422838?check_suite_focus=true#step:41:16418).

ext/crypto/lib.rs
```diff
@@ -1402,7 +1468,73 @@ pub async fn op_crypto_import_key(
...
+            // rsaEncryption
+            RSA_ENCRYPTION_OID => None,
+            // sha1WithRSAEncryption
+            SHA1_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha1),
+            // sha256WithRSAEncryption
+            SHA256_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha256),
+            // sha384WithRSAEncryption
+            SHA384_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha384),
+            // sha512WithRSAEncryption
+            SHA512_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha512),
```
littledivy, line 1496: This is incorrect as per spec. The OID is `id-RSASSA-PSS` and hash algorithm is part of `RSASSA-PSS-params`. See implementation for pkcs8 above

ext/crypto/lib.rs
```diff
@@ -1516,7 +1648,73 @@ pub async fn op_crypto_import_key(
...
```
littledivy, line 1676: Ditto. Algorithm OID is `id-RSAES-OAEP` and hash algorithm is part of `RSAES-OAEP-params`

littledivy (changes requested):

yacinehmito:
I also need to change the implementation of `generateKey`.
At the moment, Deno assumes that the key store contains PKCS1 private keys, from which the public key is derived.
However, when importing, the user only provides the public key. Therefore, we need to differentiate between these two cases in the key store.

yacinehmito:
> At the moment, Deno assumes that the key store contains PKCS1 private keys, from which the public key is derived.
> However, when importing, the user only provides the public key. Therefore, we need to differentiate between these two cases in the key store.

This has been implemented by [adding a `kind` property to the key in the key store](https://github.com/denoland/deno/pull/12921/commits/20a1bb45d2cf0548cf5b32e486db9a0fbb5f0e38), then matching against it when performing operations using RSA keys. Effectively, an RSA key pair is represented only by the private key in the store, if it has been created using `generateKey`.
I originally wanted to have `generateKey` return both the private and public keys pkcs1 buffers, but I don't know how to do that with a Deno op.

Also, I noticed that we always put the type `raw`, and that it is not used on the Rust side. @littledivy Shall remove the `type` property?

> I am only left with fixing some web platform tests related to decryption using RSA-OAEP with a non-empty label.
I don't exactly understand why it fails, as the error is generated inside the RSA library, and only gives "decryption error".

It was a simple issue. The label was not passed to the OP. [It is fixed](https://github.com/denoland/deno/pull/12921/commits/6a320f0155ebc08a05266b320452ad421e78e22f).

Also, I took this opportunity to [factor common code across all methods that copy buffers](https://github.com/denoland/deno/pull/12921/commits/c8de2b1a07093f8e3cbeb145cbdc1b046951bd86) into new [Uint8Arrays](url).

The PR is ready for review @littledivy @bartlomieju.

littledivy:
> Also, I noticed that we always put the type raw, and that it is not used on the Rust side. @littledivy Shall remove the type property?

Good catch, sgtm. (I think I added that to be used later on but I can't recall)

ext/crypto/lib.rs
```diff
@@ -458,8 +465,13 @@ pub async fn op_crypto_verify_key(
...
   let verification = match algorithm {
     Algorithm::RsassaPkcs1v15 => {
-      let public_key: RsaPublicKey =
-        RsaPrivateKey::from_pkcs1_der(&*args.key.data)?.to_public_key();
+      let public_key = match args.key.r#type {
+        KeyType::Private => {
+          RsaPrivateKey::from_pkcs1_der(&*args.key.data)?.to_public_key()
+        }
+        KeyType::Public => RsaPublicKey::from_pkcs1_der(&*args.key.data)?,
+        _ => unreachable!(),
```
bnoordhuis, line 473:
```suggestion
        KeyType::Secret => unreachable!("unexpected KeyType::Secret"),
```
That way, if another key type is added later, the compiler will complain about non-exhaustive checks, making it easy to audit each use site.

Applies here and elsewhere.

ext/crypto/lib.rs
```diff
@@ -913,8 +930,13 @@ pub async fn op_crypto_encrypt_key(
...
     Algorithm::RsaOaep => {
-      let public_key: RsaPublicKey =
-        RsaPrivateKey::from_pkcs1_der(&*args.key.data)?.to_public_key();
+      let public_key = match args.key.r#type {
+        KeyType::Private => {
+          RsaPrivateKey::from_pkcs1_der(&*args.key.data)?.to_public_key()
+        }
+        KeyType::Public => RsaPublicKey::from_pkcs1_der(&*args.key.data)?,
+        _ => unreachable!(),
+      };
```
bnoordhuis, line 939: Rule of Three. Consider factoring out these blocks into a helper function.

ext/crypto/lib.rs
```diff
@@ -1272,7 +1294,7 @@ pub async fn op_crypto_import_key(
           })?;
 
           if bytes_consumed
-            != rsa::pkcs1::der::Length::new(pk_info.private_key.len() as u16)
+            != spki::der::Length::new(pk_info.private_key.len() as u16)
```
bnoordhuis, line 1297: Just double-checking: these are the same thing, right? (`Length` from the der crate.)

bnoordhuis (commented):
The Rust side of things mostly LGTM. Bartek or Luca are probably more qualified to review the JS part. (The code itself looks okay but I can't really judge whether it's up to spec.)

Minor aside (and cc @littledivy): ext/crypto/lib.rs calls `custom_error("Foo", "some string".to_string())` a lot but that's inefficient because it allocates the string on the heap. You can just `custom_error("Foo", "some string")`.

littledivy (approved):
LGTM

yacinehmito:
> Minor aside (and cc @littledivy): ext/crypto/lib.rs calls custom_error("Foo", "some string".to_string()) a lot but that's inefficient because it allocates the string on the heap. You can just custom_error("Foo", "some string").

I'll submit another PR that changes this.

yacinehmito:
> I originally wanted to have generateKey return both the private and public keys pkcs1 buffers, but I don't know how to do that with a Deno op.

I managed to find a way, but I can't replicate it with ECDSA as ring doesn't really let us generate the public key and private key already separated. It only provides a pkcs8-formatted buffer, which would then need to be parsed again to get the raw keys.
RustCrypto can help as their implementation doesn't have the same limitations, but it doesn't support P384 natively.
I'll try an implementation in a future PR.

bartlomieju (commented):
LGTM to me too, cleanly implemented!

Before landing, I will wait for @lucacasonato review.

ext/crypto/00_crypto.js
```diff
@@ -264,6 +264,18 @@
     return normalizedAlgorithm;
   }
 
+  function copyBuffer(input) {
```
lucacasonato, line 267: Thanks for deduplicating :-)

ext/crypto/lib.rs
```diff
@@ -1288,7 +1303,73 @@ pub async fn op_crypto_import_key(
...
-        // TODO(@littledivy): spki
+        KeyFormat::Spki => {
+          let hash = args
+            .hash
+            .ok_or_else(|| type_error("Missing argument hash".to_string()))?;
+
+          // 2-3.
+          let pk_info =
+            spki::SubjectPublicKeyInfo::from_der(data).map_err(|e| {
+              custom_error("DOMExceptionOperationError", e.to_string())
```
lucacasonato, line 1314: This should be a `DataError` as per spec as far as I can tell:

> 2. Let spki be the result of running the parse a subjectPublicKeyInfo algorithm over keyData.
> 
> 3. If an error occurred while parsing, then throw a DataError.

ext/crypto/lib.rs
```diff
@@ -1288,7 +1303,73 @@ pub async fn op_crypto_import_key(
...
+            RSA_ENCRYPTION_OID => None,
+            // sha1WithRSAEncryption
+            SHA1_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha1),
+            // sha256WithRSAEncryption
+            SHA256_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha256),
+            // sha384WithRSAEncryption
+            SHA384_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha384),
+            // sha512WithRSAEncryption
+            SHA512_RSA_ENCRYPTION_OID => Some(CryptoHash::Sha512),
+            _ => return Err(type_error("Unsupported algorithm".to_string())),
```
lucacasonato, line 1332: This should be a `DataError`, not a `TypeError`:

> Otherwise:
>    1. Perform any key import steps defined by other applicable specifications, passing format, spki and obtaining hash.
>
>    2. If an error occurred or there are no applicable specifications, throw a DataError.

ext/crypto/lib.rs
```diff
@@ -1288,7 +1303,73 @@ pub async fn op_crypto_import_key(
...
+                "Hash mismatch".to_string(),
+              ));
+            }
+          }
+
+          // 8-9.
+          let public_key =
+            rsa::pkcs1::RsaPublicKey::from_der(pk_info.subject_public_key)
+              .map_err(|e| {
+                custom_error("DOMExceptionOperationError", e.to_string())
```
lucacasonato, line 1349: Spec says this should be a `DataError`:

> 8. Let publicKey be the result of performing the parse an ASN.1 structure algorithm, with data as the subjectPublicKeyInfo field of spki, structure as the RSAPublicKey structure specified in Section A.1.1 of [RFC3447], and exactData set to true.
> 
> 9. If an error occurred while parsing, or it can be determined that publicKey is not a valid public key according to [RFC3447], then throw a DataError.

ext/crypto/lib.rs
```diff
@@ -1402,7 +1483,114 @@ pub async fn op_crypto_import_key(
...
-        // TODO(@littledivy): spki
+        KeyFormat::Spki => {
+          let hash = args
+            .hash
+            .ok_or_else(|| type_error("Missing argument hash".to_string()))?;
+
+          // 2-3.
+          let pk_info =
+            spki::SubjectPublicKeyInfo::from_der(data).map_err(|e| {
+              custom_error("DOMExceptionOperationError", e.to_string())
```
lucacasonato, line 1494: Again, should be a `DataError`

ext/crypto/lib.rs
```diff
@@ -1402,7 +1483,114 @@ pub async fn op_crypto_import_key(
...
+          // 6.
+          let pk_hash = match alg {
+            // rsaEncryption
+            RSA_ENCRYPTION_OID => None,
+            // id-RSASSA-PSS
+            RSASSA_PSS_OID => {
+              let params = PssPrivateKeyParameters::try_from(
+                pk_info.algorithm.parameters.ok_or_else(|| {
+                  custom_error(
+                    "DOMExceptionNotSupportedError",
```
lucacasonato, line 1509: Should be a `DataError`.

> 1. Let params be the ASN.1 structure contained within the parameters field of the algorithm AlgorithmIdentifier field of spki.
>
> 2. If params is not defined, or is not an instance of the RSASSA-PSS-params ASN.1 type defined in [RFC3447], throw a DataError.

ext/crypto/lib.rs
```diff
@@ -1402,7 +1483,114 @@ pub async fn op_crypto_import_key(
...
+                "Hash mismatch".to_string(),
+              ));
+            }
+          }
+
+          // 8-9.
+          let public_key =
+            rsa::pkcs1::RsaPublicKey::from_der(pk_info.subject_public_key)
+              .map_err(|e| {
+                custom_error("DOMExceptionOperationError", e.to_string())
```
lucacasonato, line 1570: Again, this should be a `DataError`.

ext/crypto/lib.rs
```diff
@@ -1516,7 +1704,114 @@ pub async fn op_crypto_import_key(
...
-        // TODO(@littledivy): spki
+        KeyFormat::Spki => {
+          let hash = args
+            .hash
+            .ok_or_else(|| type_error("Missing argument hash".to_string()))?;
+
+          // 2-3.
+          let pk_info =
+            spki::SubjectPublicKeyInfo::from_der(data).map_err(|e| {
+              custom_error("DOMExceptionOperationError", e.to_string())
```
lucacasonato, line 1715: Again, this should be a `DataError`.

ext/crypto/lib.rs
```diff
@@ -1516,7 +1704,114 @@ pub async fn op_crypto_import_key(
...
+          // 6.
+          let pk_hash = match alg {
+            // rsaEncryption
+            RSA_ENCRYPTION_OID => None,
+            // id-RSAES-OAEP
+            RSAES_OAEP_OID => {
+              let params = OaepPrivateKeyParameters::try_from(
+                pk_info.algorithm.parameters.ok_or_else(|| {
+                  custom_error(
+                    "DOMExceptionNotSupportedError",
```
lucacasonato, line 1730: Again, should be a `DataError`.

ext/crypto/lib.rs
```diff
@@ -1516,7 +1704,114 @@ pub async fn op_crypto_import_key(
...
+                    "Malformed parameters".to_string(),
+                  )
+                })?,
+              )
+              .map_err(|_| {
+                custom_error(
+                  "DOMExceptionNotSupportedError",
```
lucacasonato, line 1737: Again, should be a `DataError`.

ext/crypto/lib.rs
```diff
@@ -1516,7 +1704,114 @@ pub async fn op_crypto_import_key(
...
+              };
+
+              if params.mask_gen_algorithm.oid != ID_MFG1 {
+                return Err(custom_error(
+                  "DOMExceptionNotSupportedError",
+                  "Unsupported hash algorithm".to_string(),
+                ));
+              }
+
+              hash
```
lucacasonato, line 1767: There is a check missing here:

> 6. If the parameters field of the maskGenAlgorithm field of params is not an instance of the HashAlgorithm ASN.1 type that is identical in content to the hashAlgorithm field of params, throw a NotSupportedError.

ext/crypto/lib.rs
```diff
@@ -1516,7 +1704,114 @@ pub async fn op_crypto_import_key(
...
+                "Hash mismatch".to_string(),
+              ));
+            }
+          }
+
+          // 8-9.
+          let public_key =
+            rsa::pkcs1::RsaPublicKey::from_der(pk_info.subject_public_key)
+              .map_err(|e| {
+                custom_error("DOMExceptionOperationError", e.to_string())
```
lucacasonato, line 1791: `DataError`

lucacasonato:
This is very close @yacinehmito. Thanks for working on this :-)

ext/crypto/00_crypto.js
lucacasonato: I have merged this in https://github.com/denoland/deno/pull/13030 already to reduce the size of this PR.

lucacasonato (approved):
LGTM. Incredible work, thank you @yacinehmito. At least 700 more passing tests!

bartlomieju (approved):
LGTM, thank you @yacinehmito!

PR #12924, opened 2021-11-28
[v1.16] fix(test): Improve reliability of deno test's op sanitizer with timers

bartlomieju:
Although not easy to replicate in the wild, the `deno test` op sanitizer
can fail when there are intervals that started before a test runs, since
the op sanitizer can end up running in the time between the timer op for
an interval's run resolves and the op for the next run starts.

This change fixes that by adding a new macrotask callback that will run
after the timer macrotask queue has drained. This ensures that there is
a timer op if there are any timers which are unresolved by the time the
op sanitizer runs.


bnoordhuis (approved):
RSLGTM

PR #12933, opened 2021-11-29
fix(core): Wake up the runtime if there are ticks scheduled

andreubotella:
See #12771 and #12814.

Needed to reland #12908, which was reverted in #12929.

PR #12929, opened 2021-11-29
Revert "fix(test): Improve reliability of `deno test`'s op sanitizer …

bartlomieju:
…with timers (#12908)"

This reverts commit d335343a79afbcfe719109af510fe7a1dd0df2e8.

PR #12934, opened 2021-11-29
reland #12908: fix(test): Improve reliability of `deno test`'s op sanitizer with timers

andreubotella:
This relands #12908 –which was reverted by #12929– now that #12933 has landed. It also adds @kt3k's tests from https://github.com/denoland/deno/pull/12908#issuecomment-981462813


bartlomieju:
I verified that tests in `deno_std` no longer hang.

bartlomieju (approved):
LGTM

PR #12942, opened 2021-11-30
feat(ext/timers): add refTimer, unrefTimer API

### LINKED ISSUE #6141: Feature request:  Support async unref for timers
**Use case**:  As a a module developer, I want to run periodic tasks during the execution of my module.  This can be accomplished via `setTimeout` or `setInterval` for example.  However, I do not want these timers to block the exiting of my module.  `setInterval` in particular will cause the module to run forever until the `id` associated with it is cleared.

**Specific use case**:  `std/log` file handler now uses a buffer to write log messages.  Google and Go's glog implementations both use a log daemon to flush the buffer automatically every 30 seconds.  It would be good for `std/log` file handler to use `setInterval` to flush the buffer every 30 seconds as well to mimic this but not block module exit. (see also ref #6127)

For starters, there is already support for [async unref ops](https://github.com/denoland/deno/blob/master/core/ops.rs#L20) within Deno which would likely be used in this solution.

kt3k:
This implements Deno.refTimer / Deno.unrefTimer APIs in the unstable namespace.

These APIs change the ref/unref state of op_global_timer promises.

This also unblock us to implement https://github.com/denoland/deno_std/issues/1622 in std/node, (Node.js Timeout object has .ref() and .unref() methods. We need this change in CLI first to implement them in std/node)

closes #6141

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1022,6 +1022,18 @@ declare namespace Deno {
    * Release an advisory file-system lock for the provided file.
    */
   export function funlockSync(rid: number): void;
+
+  /** **UNSTABLE**: new API, yet to be vetted.
+   *
+   * Make the timer of the given id blocking the event loop from finishing
+   */
+  export function refTimer(id: number): void;
+
+  /** **UNSTABLE**: new API, yet to be vetted.
+   *
+   * Make the timer of the given id not blocking the event loop from finishing
+   */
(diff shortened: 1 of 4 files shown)


andreubotella:
I'm rewriting the timers implementation in #12862 and it will change how timers map to async ops, conflicting with this PR. It should make unrefing timers easier, though.

kt3k:
@andreubotella Ok. I'll rework this based on your branch. Thank you for the input!

kt3k:
closing in favor of #12953

PR #12952, opened 2021-11-30
refactor: cli doesn't need to depend on deno_tls

ry:
plus some other small refactors


lucacasonato (changes requested):
The 30 second timeout will break users of streaming data, like SSE. A default timeout is not good. If you want to do timeouts, do a connection timeout, not a request timeout.

andreubotella:
Also, the `AbortSignal.timeout()` API is being considered in the DOM spec (whatwg/dom#1032), which will let users add custom request timeouts.

ry:
Ok - I'm reverting the 30 second timeout change. This PR will just be some innocent refactoring.

bartlomieju (approved):
LGTM

PR #12960, opened 2021-12-01
refactor(cli): Use GraphData for check and emit

### LINKED ISSUE #12803: feature: Add reference to the source code which caused "error: Cannot load module "xxxx/file_code.ts""
I am testing deno and sometimes I find myself doing a small code refactor. 
While doing it, I forget to change file reference.... then the error the compile error is correctly shown.

```
error: Cannot load module "file:///path/to/file_code.ts"
```

Would be great if we can also see which code is referring to the wrong file...

So far I need to look for the error using the full code search over the IDE. Not a real problem but takes more time.

Here is the deno version output:
```
deno 1.16.1 (release, aarch64-apple-darwin)
v8 9.7.106.2
typescript 4.4.2
```

### LINKED ISSUE #12805: Dynamic local file import is quite slow and seems not cached in the next launch
Continues of https://github.com/denoland/deno/issues/12519

`test.ts`

```typescript
// test.ts
async function main() {
  await Promise.all([
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
    import("https://deno.land/x/denops_core@v2.0.3/mod.ts"),
  ]);
}
await main();
```

`$ cp test.ts test.ts2`

`test2.ts`

```typescript
import { toFileUrl } from "https://deno.land/std@0.115.1/path/mod.ts";

async function main() {
  const t = performance.now();
  await Promise.all([
    import(toFileUrl(Deno.cwd() + "/test.ts").href),
    import(toFileUrl(Deno.cwd() + "/test.ts2").href),
  ]);
  console.log(performance.now() - t);
}
await main();
```

```
% deno --version
deno 1.16.2 (release, x86_64-unknown-linux-gnu)
v8 9.7.106.2
typescript 4.4.2
% deno run -A --no-check ./test2.ts
151.510647
[~/work]
% deno run -A --no-check ./test2.ts
154.57850399999998
[~/work]
% deno run -A --no-check ./test2.ts
153.059971
```

After #12604, remote URL dynamic loading is very fast!  But it is not same with dynamic local file import :-(
Simple dynamic import takes almost 80ms and it seems not cached.  It is really slow for me.
Because dynamic remote URL import is less than 2.0ms.

nayeemrmn:
Fixes #12803.
Fixes #12805.

- fix: Module graph errors show referrer info in all cases.
- fix: Module graph errors for type declaration files (`/// <reference types="..." />`, `X-TypeScript-Types`) are properly reported. They can be suppressed with `--no-check`.
- fix: Module graph errors for config file imports (`"jsxImportSources"`) are properly reported.
- TS emits are no longer stored in `ProcState::graph_data` memory. They are requested only once-per-module-per-runtime by V8, which also stores them, so they aren't useful to keep around.
- TS sources are now stored in `ProcState::graph_data` memory. These are a lot more useful to have in many short-circuiting cases.
- In `ProcState::prepare_module_load`, `createGraph()` is given a proxy loader which gives `None` for any module which is already in `ProcState::graph_data` and returns a partal graph. That means their resources aren't reloaded and their dependency branches aren't revisited when building the graph. This case had been hit for any non-statically-analyzable dynamic imports which import any existing dependencies. Performance should be drastically improved for HMR cases and other applications.
- `emit::check_and_maybe_emit()` and `tsc::exec()` are changed to take `GraphData` instead, since they may be have been given a partial graph.
- `GraphData` (now in `cli/graph_util.rs`) is extended a lot to fulfill these new roles. It reimplementes a lot of logic from `deno_graph` to work with composite graphs, hence the LOC diff.


nayeemrmn:
Currently pointing at dev branches of `deno_doc` and `deno_graph`, I assume no merge until versions are bumped proper.

cli/Cargo.toml
```diff
@@ -41,8 +41,8 @@ winres = "0.1.11"
 [dependencies]
 deno_ast = { version = "0.5.0", features = ["bundler", "codegen", "dep_graph", "module_specifier", "proposal", "react", "sourcemap", "transforms", "typescript", "view", "visit"] }
 deno_core = { version = "0.109.0", path = "../core" }
-deno_doc = "0.21.0"
-deno_graph = "0.12.0"
+deno_doc = { git = "https://github.com/denoland/deno_doc", branch = "dev" }
+deno_graph = { git = "https://github.com/denoland/deno_graph", branch = "dev" }
```
nayeemrmn, line 45: Wait for crate bumps

bartlomieju:
@nayeemrmn crates had been updated, please rebase so we can review and merge for 1.17 :)

cli/emit.rs
```diff
@@ -322,27 +333,55 @@ pub(crate) struct CheckEmitResult {
...
+    return Ok(Default::default());
+  }
+  let tsc_roots = get_tsc_roots(roots, &segment_graph_data, check_js);
+  if options.log_checks {
+    for root in roots {
+      let root_str = root.to_string();
+      // `$deno` specifiers are internal specifiers, printing out that
+      // they are being checked is confusing to a user, since they don't
+      // actually exist, so we will simply indicate that a generated module
+      // is being checked instead of the cryptic internal module
```
bartlomieju, line 370: Is it actually worth printing anything for generated modules? I'm not sure if generic `Check generated module` is any more useful

cli/graph_util.rs
```diff
@@ -0,0 +1,400 @@
...
+          };
+          if let Some((_, Some(Ok((specifier, referrer_range))))) =
+            &module.maybe_types_dependency
+          {
+            let specifier = graph.redirects.get(specifier).unwrap_or(specifier);
+            let entry = self.referrer_map.entry(specifier.clone());
+            entry.or_insert_with(|| referrer_range.clone());
+          }
+          for dep in module.dependencies.values() {
+            #[allow(clippy::manual_flatten)]
```
bartlomieju, line 105: What's the reason for ignoring this lint?

cli/graph_util.rs
```diff
@@ -0,0 +1,400 @@
...
+              }
+            }
+          }
+          for (_, dep) in dependencies.iter().rev() {
+            if !dep.is_dynamic || follow_dynamic {
+              let mut resolutions = vec![&dep.maybe_code];
+              if follow_type_only {
+                resolutions.push(&dep.maybe_type);
+              }
+              #[allow(clippy::manual_flatten)]
```
bartlomieju, line 174: Ditto

cli/proc_state.rs
```diff
@@ -712,17 +552,38 @@ impl ProcState {
...
+          | MediaType::Cjs
+          | MediaType::Mjs => code.as_ref().clone(),
+          MediaType::Dts => "".to_string(),
+          _ => {
+            let emit_path = self
+              .dir
+              .gen_cache
+              .get_cache_filename_with_extension(&found, "js")
+              .unwrap_or_else(|| {
+                panic!("Unable to get cache filename: {}", &found)
```
bartlomieju, line 573: Why `panic!` instead of returning an error?

bartlomieju (commented):
Drive-by review:

cli/graph_util.rs
nayeemrmn: The fix is more difficult to understand and more lines
```diff
- #[allow(clippy::manual_flatten)]
- for resolved in [&dep.maybe_code, &dep.maybe_type] {
-   if let Some(Ok((specifier, referrer_range))) = resolved {
-     let specifier =
-       graph.redirects.get(specifier).unwrap_or(specifier);
-     let entry = self.referrer_map.entry(specifier.clone());
-     entry.or_insert_with(|| referrer_range.clone());
-   }
- }
+ for (specifier, referrer_range) in
+   [&dep.maybe_code, &dep.maybe_type]
+     .into_iter()
+     .flatten()
+     .flatten()
+ {
+   let specifier =
+     graph.redirects.get(specifier).unwrap_or(specifier);
+   let entry = self.referrer_map.entry(specifier.clone());
+   entry.or_insert_with(|| referrer_range.clone());
+ }
```
It's just not a great rule for slice literals of `Option<Result<..>>`.

cli/emit.rs
nayeemrmn: Agreed

cli/proc_state.rs
nayeemrmn: This happens when there's an unexpected protocol, maybe a cause for confusion that I didn't use `unreachable!()` instead.

cli/emit.rs
```diff
@@ -326,27 +333,50 @@ pub(crate) struct CheckEmitResult {
...
+  if valid_emit(
+    &segment_graph_data,
+    cache,
+    &options.ts_config,
+    options.reload,
+    &options.reload_exclusions,
+  ) {
+    return Ok(Default::default());
+  }
+  let tsc_roots = get_tsc_roots(roots, &segment_graph_data, check_js);
```
kitsonk, line 363: nit:

```suggestion
  let root_names = get_tsc_roots(roots, &segment_graph_data, check_js);
```

cli/graph_util.rs
kitsonk: I agree with @nayeemrmn here.

kitsonk (commented):
While I don't have any major objections to individual lines of code, a refactor that adds ~400 lines of code doesn't feel like we are going the right direction, for code that is already working.

I think we disagree upon the abstractions between `deno_graph` and what is in CLI and it feels like we are building some sort of arbitrary abstraction between the two.

nayeemrmn:
> I think we disagree upon the abstractions between `deno_graph` and what is in CLI and it feels like we are building some sort of arbitrary abstraction between the two.

Precisely, the LOC diff is from `cli/graph_util.rs` reimplementing a lot of logic from `deno_graph` to work with a composite graph structure, and intentionally just abstracting over it. It's also because I've found `deno_graph`'s API surface to be very awkward and error prone... grouping things in ways that it shouldn't etc. This was initially just trying to speed up dynamic imports, the bugs referenced here are fixed as a side-effect of structural changes. And at a glance it looks like there are more worms to uncover from the `SyntheticModule`, `maybe_code` stuff... I would hesitate to say the code is already working. Maybe we should reconcile the two abstractions but I would want to start with this and revisit `deno_graph` if possible.

bartlomieju:
Given that this PR fixes two issues I think we should land it (in addition to the fact that it restores proper referrer information for errors). It's +1 from me.

dsherret:
> Precisely, the LOC diff is from cli/graph_util.rs reimplementing a lot of logic from deno_graph to work with a composite graph structure, and intentionally just abstracting over it. It's also because I've found deno_graph's API surface to be very awkward and error prone... grouping things in ways that it shouldn't etc.

I'm not sure we should merge this. Sounds like we should update deno_graph.

nayeemrmn:
> > Precisely, the LOC diff is from cli/graph_util.rs reimplementing a lot of logic from deno_graph to work with a composite graph structure, and intentionally just abstracting over it. It's also because I've found deno_graph's API surface to be very awkward and error prone... grouping things in ways that it shouldn't etc.
> 
> I'm not sure we should merge this. Sounds like we should update deno_graph.

I want to emphasise "reimplementing ... to work with a composite graph structure". Most of this is meaningful regardless.

dsherret (commented):
LGTM, but I still think we should try moving some of this to deno_graph somehow. We can probably work that out when doing deno_emit.

I don't understand the code well enough though so I think @kitsonk should merge this one.

kitsonk (approved):

Shougo:
Unfortunately, deno 1.17 does not fix #12805 problem.
Please see https://github.com/denoland/deno/issues/12805#issuecomment-996343077 .

PR #12963, opened 2021-12-02
test: fix flaky ws test

### LINKED ISSUE #12944: `httpConnAutoCloseDelayedOnUpgrade` and `httpServerDeleteRequestHasBody` are flaky
https://github.com/denoland/deno/runs/4361438674?check_suite_focus=true#step:35:2107

These tests have pending async ops in commits for #12862. That PR changes the clamping of timers from 4ms to 0, which affects the delay before the ops sanitizer runs, but as pointed out by @lucacasonato in https://github.com/denoland/deno/pull/12862#issuecomment-982630672, if the pending ops take longer than one event loop turn to resolve, that is a bug.

kt3k:
depends on #12862 

closes #12944

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,30 @@
+# project root files (README, Releases, LICENCE, etc)
+/*              @bartlomieju @bnoordhuis
+
+# CI files
+.github         @bartlomieju @kt3k @piscisaureus  
+
+.devcontainer/  @kitsonk
+
+bench_util/     @AaronO
+cli/            @bartlomieju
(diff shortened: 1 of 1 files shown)


kt3k:
`integration::test::steps_invalid_usage` and `timerOrdering` are flaky now 😕

andreubotella:
`timerOrdering` at least is a bug in #12862. I'll try to push a patch to fix it later today.

kt3k:
Having looked into the error of `integration::test::steps_invalid_usage`, and it seems related to the ordering of the error message from the test runner. I guess the issue would be rooted in the same problem as `timeOrdering`

Issue #8137, opened 2020-10-27
Get all the network interfaces in Deno


mayankchoubey1:
Add a method Deno.networkInterfaces() to get the list of all network interfaces with some details like:

- MAC address
- up or down
- IP Addresses

mayankchoubey1:
@bartlomieju : I've submitted this enhancement today. Do you think this will be useful? Is it okay if I assign it to myself and work on it?

mayankchoubey1:
@lucacasonato  Can I work on it? If yes, please assign to me. Thanks!

eliassjogreen:
It should probably provide pretty much the same properties as in [node](https://nodejs.org/api/os.html#os_os_networkinterfaces) if that is possible.

silverwind:
`netmask` can probably be ommited in favor of `cidr`.

rektide:
It'd be lovely to be able to get changes to the network interfaces too.

Ideally that covers both a change in what interfaces there are, and changes in status of an interface.

There's addons like https://github.com/JacobFischer/netlinkwrapper for node. There's a subfeature required to do this well in Linux: one communicates with a netlink socket over UDP. libuv has a wip [pr](https://github.com/libuv/libuv/pull/2558) to enable non AF_INET{,6} communication, for example.

bartlomieju:
@bnoordhuis is working on this feature.

PR #12964, opened 2021-12-02
feat(ext/net): add op_network_interfaces

bnoordhuis:
Add an op to list the network interfaces on the system.

Prep work for #8137 and `os.networkInterfaces()` Node compat in std.

Refs denoland/deno_std#1436.

<hr>

Opening this as a WIP because I'd like feedback on what permissions to check for.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,6 +935,7 @@ dependencies = [
  "hyper",
  "libc",
  "log",
+ "netif",
  "nix",
  "notify",
  "once_cell",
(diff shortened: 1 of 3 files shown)


bartlomieju:
> Opening this as a WIP because I'd like feedback on what permissions to check for.

IMO this should be the same check as for `Deno.hostname()` (unstable API) which requires full `--allow-env` access.

bnoordhuis:
Done. I didn't add tests because it's not exposed as a public API yet but I'll make std/node exercise it.

bartlomieju (approved):
LGTM

kt3k:
@bnoordhuis Have you started public API part of this work? If not, I'd like to continue the rest of the work

bnoordhuis:
@kt3k Only minimally so please feel free to have a go at it.

PR #12966, opened 2021-12-02
fix(test): Make the op sanitizer delay macrotask into a queue

### LINKED ISSUE #12945: integration::test::steps_invalid_usage is flaky
https://github.com/denoland/deno/runs/4361438674?check_suite_focus=true#step:35:2644

The relevant difference is in the following output:

```
test parallel steps with sanitizers ...
  test step 1 ... pending (0ms)
  test step 2 ... FAILED (6ms)
    Error: There is an op sanitizer delay already.
        at deno:runtime/js/40_testing.js:48:18
        at Object.action (deno:ext/timers/01_timers.js:150:11)
        at handleTimerMacrotask (deno:ext/timers/01_timers.js:76:12)
```

rather than the expected

```
test parallel steps with sanitizers ...
  test step 1 ... pending ([WILDCARD])
  test step 2 ... FAILED ([WILDCARD])
    Error: Cannot start test step while another test step with sanitizers is running.
     * parallel steps with sanitizers > step 1
        at preValidation ([WILDCARD])
        at testStepSanitizer ([WILDCARD])
        at [WILDCARD]/invalid_usage.ts:[WILDCARD]
        at [WILDCARD]
```

-----

The "there is an op sanitizer delay already" error was introduced in #12934, which correctly assumed no two tests or test steps should be running at the same time. The parallel steps error is being swallowed because `await opSanitizerDelay()` is called in a `finally` block.

andreubotella:
Fixes #12945.

### DIFF
--- a/runtime/js/40_testing.js
+++ b/runtime/js/40_testing.js
@@ -12,6 +12,7 @@
     AggregateError,
     ArrayPrototypeFilter,
     ArrayPrototypePush,
+    ArrayPrototypeShift,
     ArrayPrototypeSome,
     DateNow,
     Error,
@@ -30,7 +31,7 @@
   } = window.__bootstrap.primordials;
   let testStepsEnabled = false;
 
-  let opSanitizerDelayResolve = null;
+  const opSanitizerDelayResolveQueue = [];
 
   // Even if every resource is closed by the end of a test, there can be a delay
   // until the pending ops have all finished. This function returns a promise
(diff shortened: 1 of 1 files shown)


bartlomieju:
@andreubotella is this ready for review?

bartlomieju (commented):
This is reasonable for me; @dsherret what do you think?

andreubotella:
> @andreubotella is this ready for review?

Yeah. I wasn't sure if this was the best way to fix #12945, since having a single delay macrotask is only a problem if test steps are misused, but making it a queue is probably the easiest way to get rid of the flakes.

dsherret (approved):
LGTM too

PR #12968, opened 2021-12-02
chore: upgrade to Rust 1.57.0

bartlomieju:

### DIFF
--- a/cli/lsp/mod.rs
+++ b/cli/lsp/mod.rs
@@ -1,5 +1,8 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
+// FIXME(bartlomieju): remove this attribute
+#![allow(unused)]
+
 use deno_core::error::AnyError;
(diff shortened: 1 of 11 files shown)


cli/lsp/mod.rs
```diff
@@ -1,5 +1,8 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
+// FIXME(bartlomieju): remove this attribute
+#![allow(unused)]
```
bartlomieju, line 4: There are 47 fields in `cli/lsp` (mostly in `cli/lsp/tsc.rs`) that are never read. This is caused by the fact that fields only used in `Debug` implementation of struct are no longer marked as used. I'd like to address it in a follow up PR.

CC @dsherret @kitsonk

bartlomieju: Actually these structs are used via JSON serialization.

dsherret (approved):
LGTM

PR #12971, opened 2021-12-03
test(test_util/lsp): Add invalid case

wafuwafu13:
The following, which does not contain `r\n\r\n`, would have resulted in an infinite loop.

```rs
let msg2 = b"hello world";
let mut reader2 = std::io::Cursor::new(msg2);
read_message(&mut reader2);
```

### DIFF
--- a/test_util/src/lsp.rs
+++ b/test_util/src/lsp.rs
@@ -78,7 +78,7 @@ where
         .ok_or_else(|| anyhow::anyhow!("missing capture"))?;
       content_length = content_length_match.as_str().parse::<usize>()?;
     }
-    if &buf == "\r\n" || &buf == "" {
+    if &buf == "\r\n" || buf.is_empty() {
       break;
     }
   }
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
I am fine with this, but it is not very useful.

The code is just a test harness utility, and the language server does not produce messages without a `content-length: ` lead in. If it ever did, we would have bigger issues and an infinite loop would be an acceptable failure state for the test.

test_util/src/lsp.rs
```diff
@@ -78,7 +78,7 @@ where
         .ok_or_else(|| anyhow::anyhow!("missing capture"))?;
       content_length = content_length_match.as_str().parse::<usize>()?;
     }
-    if &buf == "\r\n" {
+    if &buf == "\r\n" || buf.is_empty() {
```
dsherret, line 81: I think we should panic in this case. I don't think we want to fail silently for invalid data in a test or is there some reason to?

dsherret (changes requested):

test_util/src/lsp.rs
wafuwafu13: Considering review,
>the language server does not produce messages without a content-length: lead in. If it ever did, we would have bigger issues and an infinite loop would be an acceptable failure state for the test.

I fixed it to just add invalid case.

dsherret (approved):
LGTM

PR #13010, opened 2021-12-07
feat(lsp): registry suggestion cache respects cache headers

### LINKED ISSUE #9931: FileFetcher/HttpCache should respect cache control headers

kitsonk:
Fixes #9931

Effectively this adds a "cache strategy" to the file fetcher to respect the headers of the cached response, and the LSP registry cache is the only instance of file fetcher that uses this strategy. It isn't a 100% compliant, in the sense that we a) cache in memory per session, irrespective of headers and b) we will always persist a response to the cache, even if the headers tell us not to.  Now though for the registry cache, when we go to access that cache, we will respect any headers we received and only use the cached value if it was valid. We will still only do it once per "session" though.

I have tested it locally and it works, but wanted to get any feedback on it before I flesh out all the unit tests, specially from @lucacasonato as you probably have the strongest opinions.

### DIFF
--- a/cli/http_cache.rs
+++ b/cli/http_cache.rs
@@ -17,6 +17,7 @@ use std::fs::File;
 use std::io;
 use std::path::Path;
 use std::path::PathBuf;
+use std::time::SystemTime;
 
 pub const CACHE_PERM: u32 = 0o644;
 
@@ -81,6 +82,8 @@ pub struct HttpCache {
 pub struct Metadata {
   pub headers: HeadersMap,
   pub url: String,
+  #[serde(default = "SystemTime::now")]
+  pub now: SystemTime,
 }
 
 impl Metadata {
@@ -138,7 +141,10 @@ impl HttpCache {
   // TODO(bartlomieju): this method should check headers file
   // and validate against ETAG/Last-modified-as headers.
   // ETAG check is currently done in `cli/file_fetcher.rs`.
-  pub fn get(&self, url: &Url) -> Result<(File, HeadersMap), AnyError> {
+  pub fn get(
+    &self,
+    url: &Url,
+  ) -> Result<(File, HeadersMap, SystemTime), AnyError> {
     let cache_filename = self.location.join(
       url_to_filename(url)
         .ok_or_else(|| generic_error("Can't convert url to filename."))?,
@@ -147,7 +153,7 @@ impl HttpCache {
     let file = File::open(cache_filename)?;
     let metadata = fs::read_to_string(metadata_filename)?;
     let metadata: Metadata = serde_json::from_str(&metadata)?;
-    Ok((file, metadata.headers))
+    Ok((file, metadata.headers, metadata.now))
   }
 
   pub fn set(
@@ -169,6 +175,7 @@ impl HttpCache {
     fs_util::atomic_write_file(&cache_filename, content, CACHE_PERM)?;
 
     let metadata = Metadata {
+      now: SystemTime::now(),
       url: url.to_string(),
       headers: headers_map,
     };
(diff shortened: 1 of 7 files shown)


cli/http_cache.rs
```diff
@@ -81,6 +82,8 @@ pub struct HttpCache {
 pub struct Metadata {
   pub headers: HeadersMap,
   pub url: String,
+  #[serde(default = "SystemTime::now")]
```
kitsonk, line 85: Had to do this because we are changing the format of the metadata and any existing cached metadata would be broken.  This has minimal impact on figuring out if the cached value is valid or not.

cli/http_util.rs
```diff
@@ -46,6 +50,200 @@ fn resolve_url_from_location(base_url: &Url, location: &str) -> Url {
 // Vec<(String, String)>
 pub type HeadersMap = HashMap<String, String>;
 
+type CacheControl = HashMap<String, Option<String>>;
+
+fn parse_cache_control(header: &str) -> CacheControl {
```
lucacasonato, line 55: Consider using https://crates.io/crates/cache_control

cli/http_util.rs
```diff
@@ -46,6 +50,200 @@ fn resolve_url_from_location(base_url: &Url, location: &str) -> Url {
...
+        .unwrap_or(0),
+    )
+  }
+
+  fn is_stale(&self) -> bool {
+    self.max_age() <= self.age()
+  }
+
+  fn max_age(&self) -> Duration {
+    if self.cache_control.contains_key("no-cache") {
```
lucacasonato, line 140: Where is this logic (and the one below) from? Is this from a specification somewhere?

lucacasonato (commented):
Nice! I have not tried it out locally yet, but overall looks good :-)

cli/http_util.rs
kitsonk: Based off of: https://docs.rs/http-cache-semantics/latest/src/http_cache_semantics/lib.rs.html#531-597

lucacasonato: Why don't we just use that crate?

kitsonk: Because it deals only with request/response pairs which have quite different structures for things like headers.  It would require re-organizing the whole of the file fetcher/http cache to basically make it the `Cache` stuff and re-doing the whole of the way we persist stuff in the cache.

cli/http_cache.rs
lucacasonato: Just out of curiosity, can we not use the `Date` header for this?

lucacasonato (approved):
LGTM

cli/http_cache.rs
kitsonk: The reference uses the timestamp when the response was received, versus whatever the response attests to in it date header.  I didn't dig into what is more correct, it does seem logical that when calculating time in cache that you would use "your" time versus the server's time.

lucacasonato: Well the `age` / `expiration` are relative to the server time, as they calculated them, right?

kitsonk: age is the age header plus resident time, resident time is calculated "local" time.  Expiration is calculated off raw server time (which uses the "date" header and the "local" time as a backup).

PR #13017, opened 2021-12-07
fix(ext/timers): fix flakiness of `httpConnAutoCloseDelayedOnUpgrade` test

### LINKED ISSUE #12944: `httpConnAutoCloseDelayedOnUpgrade` and `httpServerDeleteRequestHasBody` are flaky

kt3k:
Closes #12944

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -1055,12 +1055,16 @@ Deno.test(
 
       const { socket, response } = Deno.upgradeWebSocket(event1.request);
       socket.onmessage = (event) => socket.send(event.data);
+      const socketClosed = new Promise<void>((resolve) => {
+        socket.onclose = () => resolve();
+      });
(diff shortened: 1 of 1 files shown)


kt3k:
The first 5 runs show the test case is still flaky (362b0e1 failed with `httpConnAutoCloseDelayedOnUpgrade`)
The last 10 runs show the flakiness is fixed by this change.

bartlomieju (approved):
LGTM, thanks for looking into it.

Issue #9693, opened 2021-03-05
Format `lock.json` and `deno test --coverage` files with deno fmt


starkpsmichael:
Now that `deno fmt` can format json(c) files, all generated `json` files from deno, like `lock.json` and `deno test --coverage` files, should be formatted with it, before they get written.

### Actual Result `lock.json`
```
{
{
```

### Actual Result `deno test --coverage`
```
{"scriptId":"36","url":"deno:runtime/js/12_io.js","functions":[{"functionName":"writeSync","ranges":[{"startOffset":2185,"endOffset":2358,"count":4},{"startOffset":2289,"endOffset":2334,"count":0}],"isBlockCoverage":true}]}
```

### Expected Result `lock.json`
```
{
{
// notice the newline
```

### Expected Result `deno test --coverage`
```
{
  "scriptId": "36",
  "url": "deno:runtime/js/12_io.js",
  "functions": [
    {
      "functionName": "writeSync",
      "ranges": [
        { "startOffset": 2185, "endOffset": 2358, "count": 4 },
        { "startOffset": 2289, "endOffset": 2334, "count": 0 }
      ],
      "isBlockCoverage": true
    }
  ]
}
// notice the newline
```

bartlomieju:
This is very reasonable request and would be quite easy to implement. I'm in favor of doing that and contributions are welcome!

WenheLI:
Hi, interested in this issue; Just wondering what would be the possible starting point?

bartlomieju:
Hi @WenheLI, thanks for interest working on this. These steps should get you off the ground:

- there's `format_json` function available in `cli/tools/fmt.rs` that you can use to format JSON output:
https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/tools/fmt.rs#L209-L218

- for lockfile, you should run the output through `format_json` before saving it to file here:
https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/lockfile.rs#L44

- for coverage it's slightly more involved as you need to first rewrite function to write all reports to a buffer and only then format them and write to file, around here:
https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/tools/coverage.rs#L156-L164

Let me know if you have any more questions.

WenheLI:
[...]

PR #13018, opened 2021-12-07
feat(cli): use deno fmt for lock and coverage files

WenheLI:
This pr used deno_format to format json.
Try to support #9693

### DIFF
--- a/cli/lockfile.rs
+++ b/cli/lockfile.rs
@@ -12,7 +12,6 @@ use std::path::PathBuf;
 use std::rc::Rc;
 use std::sync::Arc;
 
-use crate::config_file::{FmtOptionsConfig, ProseWrap};
 use crate::tools::fmt::format_json;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


cli/lockfile.rs
```diff
@@ -42,13 +45,25 @@ impl Lockfile {
...
     let j = json!(&self.map);
     let s = serde_json::to_string_pretty(&j).unwrap();
+    let s_options = FmtOptionsConfig {
+      use_tabs: Some(false),
+      line_width: Some(80),
+      indent_width: Some(2),
+      single_quote: Some(true),
+      prose_wrap: Some(ProseWrap::Preserve),
+    };
+    let formated_s = format_json(&s, &s_options);
```
dsherret, line 55: Instead of `s_options` you should be able to do `&Default::default()` here. That will get it to use the default config and then we don't need to have the default values stored in multiple places.

cli/lockfile.rs
```diff
@@ -42,13 +45,25 @@ impl Lockfile {
...
+    let formated_str = match formated_s {
+      Ok(x) => x,
+      Err(e) => e,
```
dsherret, line 58: We can't do this because this would write out an error message to the file. Maybe consider something like:

```ts
let s = format_json(&s, &Default::default).unwrap_or_else(|| s.to_string());
```

That way it falls back to the original string when the formatting fails for some reason.

cli/tools/coverage.rs
```diff
@@ -153,13 +157,27 @@ impl CoverageCollector {
...
-      out.write_all(b"\n")?;
+      let coverage = serde_json::to_string(&script_coverage)?;
+      let formated_coverage = match format_json(&coverage, &format_options) {
+        Ok(formated_coverage) => formated_coverage,
+        Err(err) => {
+          debug!("{}", err);
+          coverage
+        }
+      };
+      out.write_all(formated_coverage.as_bytes())?;
```
dsherret, line 180: Same thing here as above.

dsherret (changes requested):
Thanks for this! Good start, just a few comments.

WenheLI:
> Thanks for this! Good start, just a few comments.

@dsherret - Thanks for the comments! Just let me know if there is any thing else I need to work on?

dsherret (approved):
LGTM. Thanks @WenheLI!

PR #13023, opened 2021-12-08
feat(lsp): improve registry completion suggestions

### LINKED ISSUE #10051: lsp: Proposal for module registry information v2
Having re-implemented what we had for module registries completion from vscode_deno v2 branch into the lsp, it made me think that there are potentially a few other features that we could benefit from.

## Import Details

[Completion Items support a documentation field](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion), which can be displayed in the editor.  This could be used to provide information about a package or module in the IDE.

In addition, we do not currently have "intelligent" hover support for imports that could be used to provide information about a package or a module.

It is proposed that we extend the `/.well-known/deno-import-intellisense.json` schema so that a `"variables"` item can contain an optional property of `"details"` which will take a URL like `"url"` pointing to an endpoint which can provide additional details.

For example it might look something like this:

```json
{
  "registries": [
    {
      "schema": "/:module([a-z0-9_]*)@:version?/:path*",
      "variables": [
        {
          "key": "path",
          "url": "https://example.com/api/modules/${module}/v/${{version}}/",
          "details": "https://example.com/api/modules/${module}/v/${{version}}/details/${path}"
        }
      ]
    }
  ]
}
```

The response payload would be an [LSP MarkupContent](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#markupContent) item, for example:

```json
{
  "kind": "markdown",
  "value": "Some sort of _description_ here."
}
```

For completions, the protocol supports lazily resolving the completion data, which means we would only query the endpoint when requested by the client, and hover requests only occur when an import is hovered.

## Path Navigation of a schema

Currently, there is no way for Deno to incrementally query a "path" from a module registry, basically a module registry has to return all potential values, even if it trying to represent something like a file system.  The Express-like path syntax we use though allows a repeating modifier (`*` or `+` for a parameter).

Currently we take that parameter as whole and query that for completions.  So for example from above, the `path` parameter is optional and allowed to repeat.  So we can have things like:

```
/example@v1.0.0/a.ts
/example@v1.0.0/b/c.ts
/example@v1.0.0/d/e/f.ts
```

But when we query the server we expected to get back all the completions possible, which we return to the client.  This prevents a user from navigating a "folder" structure.

The proposal in this case is that we would add a key named `"parts"` to a variable, which specifies the variable in the URL, which will be used for generating the repeating pattern and incrementally querying each one.  The response would need to be more intelligent to indicate whether there are more "subparts" that can be queried. (We would also validate that the schema contains a repeat modifier for that parameter too).

An example might look like this:

```json
{
  "registries": [
    {
      "schema": "/:module([a-z0-9_]*)@:version?/:path*",
      "variables": [
        {
          "key": "path",
          "url": "https://example.com/api/modules/${module}/v/${{version}}/",
          "parts": "https://example.com/api/modules/${module}/v/${{version}}/${path}"
        }
      ]
    }
  ]
}
```

The response for `https://example.com/api/modules/example/v/v1.0.0/` would look like this:

```json
[
  {
    "kind": "file",
    "value": "a.ts"
  },
  {
    "kind": "folder",
    "value": "b"
  },
  {
    "kind": "folder",
    "value": "d"
  }
]
```

And a response for `https://example.com/api/modules/example/v/v1.0.0/b` would look like:

```json
[
  {
    "kind": "file",
    "value": "c.ts"
  }
]
```

When the `"parts"` URL is absent, individual parts would not be queried, and the current behaviour of providing all the completions would occur.

kitsonk:
Resolves #10051

Still need to do all the tests, but it works. If folks want to take if for a road test, you can enable a prototype registry server by setting the following in your editor configuration:

```json
{
  "deno.enable": true,
  "deno.suggest.imports.hosts": {
    "https://regv2.deno.dev": true
  }
}
```

The server "proxies" the information from `deno.land/x/`.  Things registry servers can now implement with v2 are:

- Ability to return partial result sets.  This is done in a backwards compatible fashion, where a server can either return a `string[]` or an JSON object like:

  ```typescript
  interface VariableItemsList {
    items: string[];
    isIncomplete?: boolean;
    preselect?: string;
  }
  ```

  Where if `isIncomplete` is `true`, it indicates that not all the results have been returned and subsequent queries to the `"url"` field can be sent by the client to narrow down the result set.

- Ability to preselect particular selections, by setting the `"preselect"` value in the variable list response. This allow a registry, for example, to tell the client to choose `mod.ts` as the default selection when completing an import (or whatever the server determines is the appropriate preselection.
- Ability to provide additional documentation for a suggestion/completion item by providing a `"documentation"` endpoint as part of the `deno-import-intellisense.json`. When the client displays a particular item, it will request the server to resolve the completion item, and if the `"documentation"` endpoint is set, a request will be made to that completion endpoint.  The endpoint should response with a documentation object:

  ```ts
   interface Documentation {
    kind: "markdown" | "plaintext";
    value: string;
  }
  ```

Technically, all these changes are non-breaking with v1, and the language server will simply adjust based on the runtime behaviour of the registry.  The only real change is that v2 will cause the language server to validate the `deno-import-intellisense.json` in against the v2 specification.


bartlomieju (commented):
LGTM, I don't have any comments to the code, but I will let Luca or David approve.

dsherret (approved):
LGTM

PR #13031, opened 2021-12-09
fix(cli):  refrain from adding colours to the non-tty outputs

### LINKED ISSUE #12224: console.* methods do not properly detect non-tty output and still add colors to output
Commands such as `console.dir` still add ANSI color styles to its output of numbers/undefined/null  when the output is redirected to a file.

```
> deno eval "console.dir({some:4, thing:null})" >test.txt
```
The resultant file will contain the ANSI color codes.
```
{ some: [33m4[39m, thing: [1mnull[22m }
```

 Ideally those methods would perform a similar check to `Deno.isatty(Deno.stdout.rid)` and `Deno.isatty(Deno.stderr.rid)` and not add styles when appropriate.

VishnuJin:
fixes #12224 

#### Before:
Before

#### After
after

### DIFF
--- a/runtime/colors.rs
+++ b/runtime/colors.rs
@@ -9,11 +9,14 @@ use termcolor::{Ansi, ColorSpec, WriteColor};
 use termcolor::{BufferWriter, ColorChoice};
 
 lazy_static::lazy_static! {
+// checks if the output is piped to a tty
+  static ref IS_TTY: bool = atty::is(atty::Stream::Stdout);
   static ref NO_COLOR: bool = std::env::var_os("NO_COLOR").is_some();
 }
 
+// if the output is piped to a tty, use color
 pub fn use_color() -> bool {
-  !(*NO_COLOR)
+  !(*NO_COLOR) && *IS_TTY
 }
 
 #[cfg(windows)]


runtime/colors.rs
```diff
@@ -8,12 +8,18 @@ use termcolor::{Ansi, ColorSpec, WriteColor};
...
+}
+
 lazy_static::lazy_static! {
   static ref NO_COLOR: bool = std::env::var_os("NO_COLOR").is_some();
 }
 
+// if the output is piped to a tty, use color
 pub fn use_color() -> bool {
-  !(*NO_COLOR)
+  !(*NO_COLOR) && *IS_TTY
```
VishnuJin, line 22: @bartlomieju there is already a function called `use_color` and hence I utilised it

runtime/colors.rs
```diff
@@ -8,12 +8,18 @@ use termcolor::{Ansi, ColorSpec, WriteColor};
...
 use termcolor::{BufferWriter, ColorChoice};
 
+// checks if the output is piped to a tty
+lazy_static::lazy_static! {
+  static ref IS_TTY: bool = atty::is(atty::Stream::Stdout);
+}
+
 lazy_static::lazy_static! {
   static ref NO_COLOR: bool = std::env::var_os("NO_COLOR").is_some();
 }
```
lucacasonato, line 18:
```suggestion
lazy_static::lazy_static! {
  // Checks if the output is piped to a tty.
  static ref IS_TTY: bool = atty::is(atty::Stream::Stdout);
  static ref NO_COLOR: bool = std::env::var_os("NO_COLOR").is_some();
}
```

You can have multiple declarations inside a single `lazy_static` block.

lucacasonato (commented):
Looks great @VishnuJin! Just one small comment.

VishnuJin:
@lucacasonato  thanks for the suggestions !!

bartlomieju (approved):
LGTM

VishnuJin:
uh oh !! not sure why tests failed

kt3k:
Is this a correct solution to #12224?

This also disabled the colors of various things in CI environment (e.g. test reporter, compiling messages don't have color in CI logs. In my view, those colors were valid thing..

VishnuJin:
hmm that's a great observation and valid point, 
just an idea, since its related to CI, can we use the GitHub Action's default env variable `CI` which would be  automatically set to `true` by Actions upon workflow trigger and we could use it to switch to use_color if CI==true ?

Issue #7590, opened 2020-09-20
Deno.exit terminates entire process when using run --watch


sidd:
I'm running my script with `deno run --watch --unstable myscript.ts`. When it encounters a call to `Deno.exit`, the script terminates, including the file watcher.

I expected `Deno.exit` to terminate the script, but keep the file watcher alive. This is how Node's `process.exit` behaves when used with nodemon.

It seems culprit is [the current implementation, which is just a passthrough to `std::process::exit`](https://github.com/denoland/deno/blob/104aebdfb5d01f7482bacef6d58c2ce16da44334/cli/ops/os.rs#L117). I'm still familiarizing myself with Deno and V8, but is the solution something like tearing down the active runtime without exiting the main process?

kitsonk:
Keep the file watcher alive to do what?

sidd:
I thought Deno would continue to check for file changes, even after my script calls `Deno.exit`. Instead, I need to re-run `deno run --watch --unstable myscript.ts` manually. Minimal example of my script/output:

```ts
let someErrorCondition = true;

if (someErrorCondition) {
  console.log('Exiting');
  Deno.exit(1);
}

console.log('Finished');
```

```
$ deno --version
deno 1.4.1
v8 8.7.75
typescript 4.0.2
$ issue deno run --unstable --watch main.ts
Check file:///Users/sidd/code/issue/main.ts
Finished
Watcher Process terminated! Restarting on file change...
Watcher File change detected! Restarting!
Check file:///Users/sidd/code/issue/main.ts
Exiting
$
```

I expected to see the "Process terminated! Restarting on file change..." message after my script logged "Exiting", so I could make some changes to the script, and the script would re-run.

kitsonk:
If you exit your programme, you exit your program. I personally would find it surprising that if my programme has exited, that it suddenly restarts.

Like a "watch" in a browser, if the code changes and the page is still open, it reloads. If you close the page, you close the page. Anything else would be surprising in my opinion.

sidd:
I see, I thought my use case was the most common. This is totally possible to solve in userland, so I can use some other tool to get quick feedback while developing.

I was surprised because tools in other environments continue to watch for files after their corresponding "exit" method is called, like [`cargo watch`](https://github.com/passcod/cargo-watch) in Rust, nodemon in Node, and Guard in Ruby.

lucacasonato:
[...]

PR #13034, opened 2021-12-09
fix: op_set_exit_code

bartlomieju:
Fixes "op_set_exit_code" by sharing a single "Arc<AtomicI32>" between
all workers (via "op state") instead of having a "global" value stored in 
"deno_runtime" crate. As a consequence setting an exit code is always 
scoped to a tree of workers, instead of being overridable if there are
multiple worker tree (like in "deno test --jobs" subcommand).

Refactored "cli/main.rs" functions to return "Result<i32, AnyError>" instead
of "Result<(), AnyError>" so they can return exit code.

Ref https://github.com/denoland/deno/pull/12938


bartlomieju:
@bnoordhuis please take a look

cli/main.rs
```diff
@@ -1477,8 +1494,7 @@ pub fn main() {
 
   logger::init(flags.log_level);
 
-  unwrap_or_exit(run_basic(get_subcommand(flags)));
+  let exit_code = unwrap_or_exit(run_basic(get_subcommand(flags)));
 
-  let code = deno_runtime::EXIT_CODE.load(Relaxed);
-  std::process::exit(code);
+  std::process::exit(exit_code);
```
bartlomieju, line 1499: I think we should only call `exit` if `exit_code` is not equal to 0. Otherwise drop handlers will not be called which might lead to zombie subprocesses.

cli/main.rs
```diff
@@ -992,10 +1001,12 @@ async fn run_from_stdin(flags: Flags) -> Result<(), AnyError> {
...
     "window.dispatchEvent(new Event('unload'))",
   )?;
-  Ok(())
+  Ok(worker.get_exit_code())
 }
 
-async fn run_with_watch(flags: Flags, script: String) -> Result<(), AnyError> {
+// TODO(bartlomieju): this function is not handling `exit_code` set by the runtime
+// code properly.
+async fn run_with_watch(flags: Flags, script: String) -> Result<i32, AnyError> {
```
bartlomieju, line 1009: Should be addressed in tandem with https://github.com/denoland/deno/issues/7590

cli/main.rs
bnoordhuis: What drop handlers are you worried about? The real work is scoped to run_basic(), right?

runtime/ops/os.rs
```diff
@@ -10,9 +10,11 @@ use deno_core::OpState;
 use serde::Serialize;
 use std::collections::HashMap;
 use std::env;
+use std::sync::atomic::AtomicI32;
 use std::sync::atomic::Ordering::Relaxed;
+use std::sync::Arc;
 
-pub fn init() -> Extension {
+pub fn init(maybe_exit_code: Option<Arc<AtomicI32>>) -> Extension {
```
bnoordhuis, line 17: Does it have to be an `Arc`? It's only accessed from the same thread, I think.

bnoordhuis (approved):
Nice work.

cli/main.rs
bartlomieju: All drop handlers - IIRC if you call `std::process:exit` the process immediately terminates and no drop handlers are run; when we run a subprocess using `Deno.run()` we got `close_on_drop` set to true, and when you are existing using `exit()` these drop handlers will not run and subprocesses will not close becoming zombies.

runtime/ops/os.rs
bartlomieju: It has to, it's shares between all created threads (ie. web workers).

bartlomieju:
I'm going to land this PR and we can address other questions at a later point.

cli/main.rs
bnoordhuis: I'm still confused. Are you talking about lexically scoped drop handlers? main() doesn't have any that matter, does it?

(I'm also confused by your `close_on_drop` comment - where is that a thing inside the source tree?)

bartlomieju: Please scratch that, I should have commented before the first coffee :) of course drop handlers run because they are scoped to `run_basic()`. And in case of `close_on_drop` I actually meant [`kill_on_drop`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.kill_on_drop)

PR #13027, opened 2021-12-09
refactor(ext/crypto): various cleanups in js code

SeanWykes:
As per request @lucacasonato in #12022. 

Refac covers:
1) Move importKey(jwt) for symmetric-keys to Rust, using `fn decode_b64url()`

2) Use enum `ImportExportKeyData` and related `pub struct`s to distinguish raw/jwk import data. Will be extended for RSA and EC JWK imports.

3) Move common js code to `function importKeyRSA(...)`

4) Added importkey for ECDH(raw) and moved common code to `function importKeyEC(...)`

5) Standardize use of `type`/`keyType`/`format` in key structures.
 - includes `read_rsa_public_key()` from #12921, along with validation of `r#type` where necessary
 
6) Cleanup, 
 - use const/alias `algorithmName` for `normalizedAlgorithm.name`
 - use const/alias for `namedCurve` and 'keyType'
 - use short-form to build js hashes, where possible, for example `{ format, namedCurve, ... }`

PR #12022, opened 2021-09-13
feat(ext/crypto): implement importKey (jwk) for RSA-PSS , RSA-OAEP and RSA-PKCS1-v1_5

SeanWykes:
towards #11690, implementing: 
- importKey(jwk) for RSA-PSS, RSASSA-PKCS1-v1_5, RSA-OAEP
- exportKey(jwk) for same algorithms.

PR #13036, opened 2021-12-09
refactor(ext/crypto): clean up rust side importKey

lucacasonato:
This commit cleans up the Rust side of `import_key` by using a bunch of
enums instead of structs with "type" and "data" fields.

This commit does add some duplicated code for the time being, because
a lot of the other ops still need to get the same cleanup treatment.

This is an alternative to the Rust side changes in #13027.


lucacasonato:
@SeanWykes Does the pattern introduced here work for the JWK implementation?

SeanWykes:
> @SeanWykes Does the pattern introduced here work for the JWK implementation?

Yes, should do, we'll just add two more members to KeyData enum. :)

lucacasonato:
> Yes, should do, we'll just add two more members to KeyData enum. :)

As in two fields inside `KeyData::Jwk`, or two new enum members? If you mean the latter, what members would those be?

SeanWykes:
> > Yes, should do, we'll just add two more members to KeyData enum. :)
> 
> As in two fields inside `KeyData::Jwk`, or two new enum members? If you mean the latter, what members would those be?

I personally find new enum members to be cleaner than having a single big `::JwK` structure.

In #12022, I use different structs and enum members for Symmetric, RSA and EC JWKs: 
`JwkSecretKey(SecretKeyComponentB64)`,  `JwkRsaKey(RSAKeyComponentsB64)` and `JwkEcKey(ECKeyComponentsB64)`.

lucacasonato:
Makes sense. Let's cross that bridge when we get to it :-)

ext/crypto/import_key.rs
```diff
@@ -0,0 +1,581 @@
...
+      import_key_rsassa(key_data, hash)
+    }
+    ImportKeyOptions::RsaPss { hash } => import_key_rsapss(key_data, hash),
+    ImportKeyOptions::RsaOaep { hash } => import_key_rsaoaep(key_data, hash),
+    ImportKeyOptions::Ecdsa { named_curve }
+    | ImportKeyOptions::Ecdh { named_curve } => {
+      import_key_ec(key_data, named_curve)
+    }
+  }
+}
```
bartlomieju, line 69: Why is it an async function is there are no `await` points?

lucacasonato: Because that is how it already was :shrug:. I'll go make it sync.

littledivy (approved):
neat

bartlomieju (approved):
LGTM

PR #13037, opened 2021-12-09
chore: place `@deprecated` tag after documentation block

kitsonk:
The way JSDoc tags work, once a tag is encountered, the text associated with the tag is everything up to the next tag, so all tags should come _after_ the documentation block associated with the entity.  This is consistent with how the likes of vscode and other parse tags, including now `deno_doc`.

This will fix issues where the documentation block is not appearing for certain entries in `doc.deno.land` because only the doc block is being displayed in a listing (along with just the `deprecated` tag).

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -654,9 +654,6 @@ declare namespace Deno {
   }
 
   /**
-   * @deprecated Use `copy` from https://deno.land/std/streams/conversion.ts
-   * instead. `Deno.copy` will be removed in Deno 2.0.
-   *
    * Copies from `src` to `dst` until either EOF (`null`) is read from `src` or
    * an error occurs. It resolves to the number of bytes copied or rejects with
    * the first error encountered while copying.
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):

lucacasonato:
Unfortunately the deprecation warning doesn't show up at the top of the VS Code hover card anymore now.

kitsonk:
It still shows up in the card, but it is also tagged in the lsp as "deprecated" so it has other visual indicators in the IDE.

Here is what is done in `lib.dom.d.ts` to show that it needs to be trailing:

```typescript
/**
 * The Web Audio API events that occur when a ScriptProcessorNode input buffer is ready to be processed.
 * @deprecated As of the August 29 2014 Web Audio API spec publication, this feature has been marked as deprecated, and is soon to be replaced by AudioWorklet.
 */
interface AudioProcessingEvent extends Event {
    /** @deprecated */
    readonly inputBuffer: AudioBuffer;
    /** @deprecated */
    readonly outputBuffer: AudioBuffer;
    /** @deprecated */
    readonly playbackTime: number;
}
```

kitsonk:
If there is too much information in the JSDoc block for a deprecated API, we can always shorten it.  But once you hit a tag, all the text gets associated with the tag until the next tag.

PR #13038, opened 2021-12-09
fix(watch): mitigate race condition between file write by other process and watch read

### LINKED ISSUE #13035: File watcher race condition where `Modify` event occurs and file contents are empty when being read
I was looking into the watch tests and a source of their flakiness is actually a race condition in the file watcher itself.

Basically the following events occur:

1. The deno process is listening for file watching events in `/tmp/.tmpb7vuus`
1. Another process writes to `/tmp/.tmpb7vuus/badly_formatted.js`
1. A `Modify(Data(Any))` event occurs in the file watcher for `/tmp/.tmpb7vuus/badly_formatted.js`
1. The deno process then reads the contents of the file, but the contents are empty (because it hasn't finished writing out the file yet... truncated then write occurs from my understanding).

I can reproduce this bug almost every time I run `fmt_watch_without_args_test` in WSL when it's the only test running.

I think this can be solved by improving our debouncer to actually debounce (code is hard to understand, but it doesn't look like it debounces).

dsherret:
Fixes #13035 and see there for details.

In summary, this mitigates reading from a file while it is still being written to. The mitigation here is to actually do a debounce instead of what looked like a throttle, but was called a debounce (maybe I'm misreading the code or don't know my definitions though :P). This also seems to have improved the test flakiness as I can run the tests a bunch of times now and they don't fail in WSL whereas previously they would fail quite often.

### DIFF
--- a/cli/file_watcher.rs
+++ b/cli/file_watcher.rs
@@ -4,9 +4,7 @@ use crate::colors;
 use crate::fs_util::canonicalize_path;
 
 use deno_core::error::AnyError;
-use deno_core::futures::stream::{Stream, StreamExt};
 use deno_core::futures::Future;
-use deno_core::parking_lot::Mutex;
 use log::info;
 use notify::event::Event as NotifyEvent;
 use notify::event::EventKind;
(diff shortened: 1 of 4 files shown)


cli/file_watcher.rs
```diff
@@ -15,57 +13,42 @@ use notify::Error as NotifyError;
...
+    (Arc::new(sender), Self { receiver })
   }
-}
 
-impl Stream for Debounce {
-  type Item = Vec<PathBuf>;
-
-  /// Note that this never returns `Poll::Ready(None)`, which means that the
-  /// file watcher will be alive until the Deno process is terminated.
-  fn poll_next(
```
dsherret, line 55: IMO, I think we should try to avoid implementing this trait if able (maybe only mission critical stuff). I think the new debounced channel is easier to understand.

cli/file_watcher.rs
```diff
@@ -15,57 +13,42 @@ use notify::Error as NotifyError;
...
-use tokio::time::Sleep;
 
 const DEBOUNCE_INTERVAL: Duration = Duration::from_millis(200);
 
-#[pin_project(project = DebounceProjection)]
-struct Debounce {
-  #[pin]
-  timer: Sleep,
-  changed_paths: Arc<Mutex<HashSet<PathBuf>>>,
+struct DebouncedReceiver {
```
dsherret, line 26: I've just lazily made this very implementation specific because we don't seem to have a folder where we can just throw in these kind of utility structs or I'm not sure of a specific one to put it in. Perhaps it would be useful to extract this out in the future.

cli/tests/integration/watcher_tests.rs
```diff
@@ -106,13 +114,7 @@ fn lint_watch_test() {
     .stderr(std::process::Stdio::piped())
     .spawn()
     .expect("Failed to spawn script");
-  let mut stderr = child.stderr.as_mut().unwrap();
-  let mut stderr_lines = std::io::BufReader::new(&mut stderr)
-    .lines()
-    .map(|r| r.unwrap());
-
-  // TODO(lucacasonato): remove this timeout. It seems to be needed on Linux.
-  std::thread::sleep(std::time::Duration::from_secs(1));
```
dsherret, line 115: I think some of these timeouts were unnecessary because the read below would block until stderr is received.

dsherret:
It seems not flaky anymore. The CI runs that failed were other stuff. Maybe we can close the flaky watch issues?

bartlomieju (approved):
LGTM, this is great

PR #13046, opened 2021-12-10
fix(ext/web): set location undefined when `--location` is not specified

### LINKED ISSUE #12969: compat: type-detect breaks on access to window.location
[`type-detect`](https://www.npmjs.com/package/type-detect) is a very popular Node package (over 14m weekly downloads). However it cannot run in `--compat` without `--location` flag:

```
deno run --compat --unstable -A --no-check ./dist/bin/nps.js -c ./package-scripts.js
error: Uncaught (in promise) ReferenceError: Access to "location", run again with --location <href>.
    if (typeof window.location === 'object' && obj === window.location) {
                      ^
    at get (deno:ext/web/12_location.js:381:17)
    at typeDetect (file:///Users/biwanczuk/dev/nps/node_modules/type-detect/type-detect.js:118:23)
    at loadConfig (file:///Users/biwanczuk/dev/nps/dist/bin-utils/index.js:107:78)
    at getPSConfig (file:///Users/biwanczuk/dev/nps/dist/bin-utils/parser.js:274:35)
    at parse (file:///Users/biwanczuk/dev/nps/dist/bin-utils/parser.js:124:18)
    at Object.<anonymous> (file:///Users/biwanczuk/dev/nps/dist/bin/nps.js:21:33)
    at Module._compile (file:///Users/biwanczuk/dev/deno_std/node/module.ts:251:36)
    at Object.Module._extensions..js (file:///Users/biwanczuk/dev/deno_std/node/module.ts:1276:10)
    at Module.load (file:///Users/biwanczuk/dev/deno_std/node/module.ts:230:33)
    at Function._load (file:///Users/biwanczuk/dev/deno_std/node/module.ts:539:14)

```

I found this problem when trying to run [`nps`](https://www.npmjs.com/package/nps).

I believe this severely hinders DX, and we should consider how to solve it. An obvious solution is make `window.location` be `undefined` if no `--location` flag is passed: https://github.com/denoland/deno/pull/10828

CC @kt3k

kt3k:
Currently `globalThis.location` throws an error when `--location` is not specified. This behavior is inconvenient if a script does the feature detection like `typeof window.location` ([type-detect](https://npm.im/type-detect) is an example of such a script)

This PR changes it to return `undefined` for `window.location` when `--location` is not specified.

closes #12969
previous version: #10828

### DIFF
--- a/ext/web/12_location.js
+++ b/ext/web/12_location.js
@@ -9,7 +9,6 @@
   const {
     Error,
     ObjectDefineProperties,
-    ReferenceError,
     Symbol,
     SymbolFor,
     SymbolToStringTag,
--- a/ext/web/12_location.js
+++ b/ext/web/12_location.js
@@ -356,8 +356,8 @@
     },
   });
 
-  let location = null;
-  let workerLocation = null;
+  let location = undefined;
+  let workerLocation = undefined;
 
   function setLocationHref(href) {
     location = new Location(href, locationConstructorKey);
@@ -378,8 +378,9 @@
     locationDescriptor: {
       get() {
         if (location == null) {
-          throw new ReferenceError(
-            `Access to "location", run again with --location <href>.`,
+          console.warn(
+            "Warning: accessing undefined 'location' global, " +
+              "run again with --location <href>.",
           );
         }
         return location;
--- a/ext/web/12_location.js
+++ b/ext/web/12_location.js
@@ -376,12 +376,6 @@
     },
     locationDescriptor: {
       get() {
-        if (location == null) {
-          console.warn(
-            "Warning: accessing undefined 'location' global, " +
-              "run again with --location <href>.",
-          );
-        }
         return location;
       },
       set() {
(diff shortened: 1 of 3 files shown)


ext/web/12_location.js
```diff
@@ -378,8 +377,9 @@
     locationDescriptor: {
       get() {
         if (location == null) {
-          throw new ReferenceError(
-            `Access to "location", run again with --location <href>.`,
+          console.warn(
+            "Warning: accessing undefined 'location' global, " +
+              "run again with --location <href>.",
           );
```
bartlomieju, line 383: Discussed this with @lucacasonato, we will like this warning should not be printed, otherwise users feature testing if `location` is available will see these warnings.

kt3k: Ok. Let's remove

lucacasonato (approved):

bartlomieju (approved):
LGTM too

PR #13047, opened 2021-12-10
refactor(ext/crypto): symmetric jwk decode in rust

lucacasonato:
This commit moves JWK secret importing for HMAC and AES to Rust.


bartlomieju (approved):
LGTM

PR #13049, opened 2021-12-10
fix(cli) use color to the outputs for CI workflow

### LINKED ISSUE #13050: add output colors back to the CI workflow
Is this a correct solution to #12224?

This also disabled the colors of various things in CI environment (e.g. test reporter, compiling messages don't have color in CI logs. In my view, those colors were valid thing..

_Originally posted by @kt3k in https://github.com/denoland/deno/issues/13031#issuecomment-991172663_

VishnuJin:
followup on #13031 
fixes #13050
Output colors for CI workflows has been removed and this PR tries to address by adding them back

The idea is to use Github Action's default environment variable `CI` to determine to use output colour for non-tty

### DIFF
--- a/runtime/colors.rs
+++ b/runtime/colors.rs
@@ -12,11 +12,13 @@ lazy_static::lazy_static! {
 // checks if the output is piped to a tty
   static ref IS_TTY: bool = atty::is(atty::Stream::Stdout);
   static ref NO_COLOR: bool = std::env::var_os("NO_COLOR").is_some();
+// checks if the run is part of CI workflow to determine if we should use colors to the output
+  static ref IS_CI_WORKFLOW: bool = std::env::var_os("CI").is_some();
 }
 
-// if the output is piped to a tty, use color
+// if the output is piped to a tty or CI workfkow from Github Action, use color
 pub fn use_color() -> bool {
-  !(*NO_COLOR) && *IS_TTY
+  !(*NO_COLOR) && (*IS_TTY || *IS_CI_WORKFLOW)
 }
 
 #[cfg(windows)]


VishnuJin:
@kt3k I believe the colours are back to CI workflows, could you please confirm ?

piscisaureus:
I am against hard-coding the `CI` env var into Deno.

Instead we should define a dedicated environment variable (e.g. `DENO_COLOR=always` which is similar to `CARGO_TERM_COLOR=always`) or a flag (like `cargo --color=always` or `clang -fdiagnostics-color=always`) to do this.

VishnuJin:
I somehow thought like utilising the existing env var is better than creating one.

I agree with the idea though, so Incase we create a new env variable as `DENO_COLOR` with the 3 options such as `always | never | auto` that shall default to `auto`, then what shall we do with existing `NO_COLOR` ?
It would cause confusion having 2 env vars almost doing same thing IMHO !!

VishnuJin:
I believe having an environment variable `DENO_COLOR=always | never | auto`(with options) would give more control to the user to determine/choose the options as per their need instead of us deciding for them to not use color on non-ttys or vice versa

bartlomieju:
@piscisaureus solution is the best. @VishnuJin could you update your PR to reflect that?

ry:
It’s rather unfortunate to have two color environment variables. 
 https://no-color.org/ seems quite sane.

VishnuJin:
@bartlomieju how shall we proceed with the implementation, do we need to go with the new env Variable based that has 3 options or how to proceed, please drop me a hint for implementation ?

VishnuJin:
addressed in #13068

PR #13051, opened 2021-12-10
chore: run cargo update

### LINKED ISSUE #13026: Reference Error while processing enum with `--no-check` option
The canary build throws with the following enum (when `--no-check` option is passed):

```ts
export enum Encodings {
  ASCII,
  UTF8,
  BASE64,
  UCS2,
  BINARY,
  HEX,
  BUFFER,
  BASE64URL,
  LATIN1 = BINARY,
}
```
(ref: https://github.com/denoland/deno_std/blob/21e0b73979466202e738f458ff0b9232b21dd4e1/node/internal_binding/_node.ts )

It throws with `ReferenceError: BINARY is not defined` only when `--no-check` is specified.

I guess this might be related to #13025. cc @dsherret

dsherret:
Closes #13026

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,12 +38,6 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.4.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

kt3k:
@dsherret Thank you for the quick fix!

Issue #12919, opened 2021-11-28
Misleading error when calling non-existent op


DjDeveloperr:
Found this when debugging #12918 

When you call a non-existent op, the error is kind of misleading and doesn't tell what is actually wrong.

**Reproduction**

Try `Deno.core.opSync("non_existent")` in REPL and you'll get
```
Uncaught TypeError: invalid op id: expected type `v8::data::Integer`, got `v8::data::Value`
    at Object.opSync (deno:core/01_core.js:142:27)
    at <anonymous>:2:11
```

bartlomieju:
I hit this problem as well this week, I agree that the error is not informative. I think we'd need to update `rusty_v8` to have better output for `v8::Value`.

bartlomieju:
@bnoordhuis what do you think?

andreubotella:
I think these are separate concerns – `serde_v8` should probably have a better output in this case, sure, but any `serde_v8` error output would be of no use to someone calling a non-existent op. That should raise a `deno_core`-specific error.

bartlomieju:
> I think these are separate concerns – `serde_v8` should probably have a better output in this case, sure, but any `serde_v8` error output would be of no use to a user of `deno_core` calling a non-existent op. That should raise a `deno_core`-specific error.

Sure, but `v8::Value` can be a number of things, I believe the representation could show what's "hidden" behind the value (`undefined, `integer`, etc.). There's already `deno_core` specific error here: `invalid op id:`

bnoordhuis:
I was griping about that error message just yesterday so I definitely would like to see it improved.

I wouldn't say it's a rusty_v8 or serde_v8 issue, I don't think they can improve the error message much. The problem is that `undefined` is passed where a number is expected.

 Note how `opSync()` doesn't check if `opsName` is a key in `opsCache`:
https://github.com/denoland/deno/blob/3881e2e7bf3bbe4a5d62872a9fb7b6e2e3d82f81/core/01_core.js#L148-L150
It basically evaluates to `opcallSync(undefined, arg1, arg2)`.

PR #13056, opened 2021-12-11
fix(core): improve error on invalid op id

bartlomieju:
```
> Deno.core.opSync(undefined, 1, 1);
Uncaught TypeError: invalid op id: received `undefined` instead of an integer.
This error is often caused by a typo in an op name, or not calling
JsRuntime::sync_ops_cache() after JsRuntime initialization.
    at Object.opSync (deno:core/01_core.js:149:27)
    at <anonymous>:2:11
>
```

Closes https://github.com/denoland/deno/issues/12919

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -437,7 +437,16 @@ fn opcall_sync<'s>(
   {
     Ok(op_id) => op_id,
     Err(err) => {
-      throw_type_error(scope, format!("invalid op id: {}", err));
+      let msg = if args.get(0).is_undefined() {
+        format!(
+          "invalid op id: received `undefined` instead of an integer. 
+        This error is often caused by a typo in an op name, or not calling 
+        JsRuntime::sync_ops_cache() after JsRuntime initialization."
+        )
+      } else {
+        format!("invalid op id: {}", err)
+      };
+      throw_type_error(scope, msg);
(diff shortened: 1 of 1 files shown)


core/bindings.rs
```diff
@@ -437,7 +442,12 @@ fn opcall_sync<'s>(
   {
     Ok(op_id) => op_id,
     Err(err) => {
-      throw_type_error(scope, format!("invalid op id: {}", err));
+      let msg = if args.get(0).is_undefined() {
+        UNDEFINED_OP_ID_MSG.to_string()
+      } else {
+        format!("invalid op id: {}", err)
+      };
+      throw_type_error(scope, msg);
```
bnoordhuis, line 450: This is really just micro-optimization-feel-free-to-ignore territory but...:
```suggestion
      let msg = if args.get(0).is_undefined() {
        UNDEFINED_OP_ID_MSG
      } else {
        &format!("invalid op id: {}", err)
      };
      throw_type_error(scope, msg);
```
(ditto below)

bnoordhuis (approved):
👍

core/bindings.rs
bartlomieju: It's not a hot path, I'm gonna ignore it

PR #13060, opened 2021-12-12
fix(std/unstable): Remove ignorement of useDefineForClassFields from runtime compiler and fix its default value comment in dts type

EsTharian:


bartlomieju:
Hi @EsTharian, could you provide a rationale for this change?

EsTharian:
> Hi @EsTharian, could you provide a rationale for this change?

Defining of class fields with `define()` is a compatibility matter for browsers. Developer should consider if they need to convert their ES6+ code to legacy compatible format. If they convert the code with `define()`, the bundle size will be bigger. But, these days, developers can choose to not use `define()` because modern browsers ([except older than Safari v14.0/iOS v14.4](https://caniuse.com/mdn-javascript_classes_public_class_fields)) are compatible. So, we need to give ability of choosing if they use or not.

nayeemrmn:
> Also, it's already `true` in [default](https://github.com/denoland/deno/blob/main/cli/emit.rs#L212), typing comment should be updated.

This is correct.

We also ignore the option because options which affect the emit like this need to be unified across libraries, and it's unsafe/fracturing to leave it configurable. So I think that part should be preserved. `deno bundle` is secondarily for browsers.

EsTharian:
`deno bundle` cannot bundle CSS and CSS pre-processors, not emitting static files etc. Although much faster it's not an alternative to Webpack, Gulp etc. for now.

If we plan to be able to, of course it will be much better. But, because `Deno.emit()` is already flagged as unstable, I don't think that leaving it configurable could harm. Being or creating an alternative to current bundlers will take some time. It should be configurable until the time comes.

kitsonk:
> We also ignore the option because options which affect the emit like this need to be unified across libraries, and it's unsafe/fracturing to leave it configurable. So I think that part should be preserved. `deno bundle` is secondarily for browsers.

True, but you cannot change the target for `deno bundle`, it is always ESNext, where as with `Deno.emit()`, you can change the target.

Ultimately for `Deno.emit()`, because we allow people to change the target, it is fine, but a) we should have tests for the change and b) we need to ensure it works with no-check/swc.  I can't specifically remember if you can instruct swc to emit class fields in the same fashion as tsc.  I suspect you can, but we haven't wired that into the options that effect the emit.

bartlomieju:
@kitsonk so is it desirable to land this PR or should we close it?

kitsonk:
Without tests it isn't landable. I can see why someone would want it.

EsTharian:
fresh

kitsonk:
@EsTharian it has no tests.  Do you plan to add tests?

EsTharian:
Yes, I have. I had no time to add, sorry, I will plan it for this week @kitsonk

bartlomieju:
Bump @EsTharian, are you planning on adding tests still? If not I'd like to close this PR.

EsTharian:
@bartlomieju sorry, I won't be able to add tests for a couple of weeks. If I will, I may open another PR. Thank you for your supports!

PR #13065, opened 2021-12-13
fix(test): wait for inspector session in side modules

### LINKED ISSUE #12584: Can't use inspector with test runner
Using the following `launch.json` I am unable to debug any kind of tests, even if I put a breakpoint outside a `Deno.test` block.

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "request": "launch",
      "name": "Debug",
      "type": "pwa-node",
      "program": "${file}",
      "cwd": "${workspaceFolder}",
      "runtimeExecutable": "deno",
      "runtimeArgs": [
        "test",
        "--allow-all",
        "--inspect-brk"
      ],
      "attachSimplePort": 9229
    }
  ]
}
```

Using simply the command `deno test --inspect-brk --allow-all main.ts` will run the tests normally, but won't wait for any inspector to connect before exiting, as it would happen using `deno run`.

jespertheend:
Fixes #12584

This the first time I've ever worked with rust so apologies if this a terrible PR.
All tests passed locally except for `test_respect_cache_revalidates` which was already failing without this change.

One thing I was a bit worried about was that this change would cause multiple breaks every time a new module is loaded. But I couldn't find any case where this happens.

Also, should a new test case for this specific issue be added? I couldn't find any existing tests checking for the inspector like this, so I'm not sure I'm capable of writing such a test case.

### DIFF
--- a/cli/tests/integration/inspector_tests.rs
+++ b/cli/tests/integration/inspector_tests.rs
@@ -519,3 +519,76 @@ async fn inspector_connect_non_ws() {
   child.kill().unwrap();
   child.wait().unwrap();
 }
+
+#[tokio::test]
+async fn inspector_break_on_first_line_in_test() {
+  let script = util::testdata_path().join("inspector_test.js");
+  let mut child = util::deno_cmd()
+    .arg("test")
+    .arg(inspect_flag_with_unique_port("--inspect-brk"))
+    .arg(script)
+    .stdout(std::process::Stdio::piped())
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks @jespertheend this indeed fixes the problem with debugging tests. I'll add a test case and land it for v1.17

bartlomieju (approved):
LGTM, thank you @jespertheend for looking into this issue.

PR #13066, opened 2021-12-13
feat: Support abort reasons in Deno APIs and `WebSocketStream`

andreubotella:
﻿

### DIFF
--- a/runtime/js/40_write_file.js
+++ b/runtime/js/40_write_file.js
@@ -13,9 +13,7 @@
     data,
     options = {},
   ) {
-    if (options?.signal?.aborted) {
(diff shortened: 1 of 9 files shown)


runtime/js/40_write_file.js
```diff
@@ -73,15 +71,18 @@
...
+    } finally {
+      file.close();
     }
 
-    file.close();
-
-    if (signal?.aborted) {
-      throw new DOMException("The write operation was aborted.", "AbortError");
-    }
+    signal?.throwIfAborted();
```
lucacasonato, line 85: There is no reason to abort here. The operation is already fully complete.

lucacasonato (approved):
LGTM, thanks

PR #13068, opened 2021-12-13
Feat(cli) add color flag to deno

VishnuJin:
this PR adds `--colour` global flag to Deno which can also be set using `DENO_TERM_COLOR` environment variable

Just like `cargo --color` we accept the following 3 arguments 
- auto - the default, this option can identify tty and uses colour only to them
- always - use color even to non-tty
- never - never use colour even in tty

`--color` flag will be overridden if `NO_COLOR` environment var is set

follow up on #13049 

P.S: tests and docs are yet to be added if the approach taken is said okay

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -13,6 +13,7 @@ use deno_runtime::permissions::PermissionsOptions;
 use log::debug;
 use log::Level;
 use std::env;
+use std::ffi::OsString;
 use std::net::SocketAddr;
 use std::num::NonZeroU32;
 use std::num::NonZeroU8;
@@ -218,11 +219,30 @@ pub enum ColorFlag {
   Never,
 }
 
+// default value is set based on the --color or DENO_TERM_COLOR env variable
+// just  to avoid setting --color in all tests
+// for instance in CI workflow DENO_TERM_COLOR is set to 'always' but if `--color` is not set explicitly as --always in tests
+// then tests will fail to avoid this we set default value based on the --color or DENO_TERM_COLOR env variable
+// if neither is set then defaults to Auto
 impl Default for ColorFlag {
   fn default() -> Self {
-    Self::Auto
+    let preference = std::env::var_os("DENO_TERM_COLOR")
+      .unwrap_or_else(|| OsString::from("auto"))
+      .into_string()
+      .unwrap();
+    match_color_preferance(&preference)
   }
 }
+
+fn match_color_preferance(s: &str) -> ColorFlag {
+  match s.to_lowercase().as_ref() {
+    "auto" => ColorFlag::Auto,
+    "always" => ColorFlag::Always,
+    "never" => ColorFlag::Never,
+    _ => unreachable!(),
+  }
+}
+
 #[derive(Clone, Debug, PartialEq, Default)]
 pub struct Flags {
   /// Vector of CLI arguments - these are user script arguments, all Deno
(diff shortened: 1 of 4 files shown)


VishnuJin:
@bartlomieju @ry @piscisaureus @kt3k 

please let me know your thoughts on the approach taken or if we need to try something else !!

cli/flags.rs
```diff
@@ -422,6 +440,18 @@ pub fn flags_from_vec(args: Vec<String>) -> clap::Result<Flags> {
...
+  if matches.is_present("color") {
+    let preferece = matches.value_of("color").unwrap().to_string();
+    flags.color = match preferece.as_str() {
+      "auto" => ColorFlag::Auto,
+      "always" => ColorFlag::Always,
+      "never" => ColorFlag::Never,
+      _ => unreachable!(),
+    };
+    // DENO_COLOR env var is created to let runtime/color.rs know about the color flag
+    env::set_var("DENO_TERM_COLOR", preferece.clone());
```
VishnuJin, line 452: I'm unsure how reach the `--color` flag's value from `runtime/color.rs` and hence used env var, can this be improved ?

VishnuJin:
In [test debug jobs](https://github.com/denoland/deno/runs/4515345691?check_suite_focus=true) the colors are used as expected 
In [test release job](https://github.com/denoland/deno/runs/4515345673?check_suite_focus=true) color is not being used despite setting `DENO_TERM_COLOR=always` in CI.yaml not sure why :(

VishnuJin:
@bartlomieju @ry @piscisaureus @kt3k
this is now working as expected in CI as well

ry (changes requested):
Hi VishnuJin,

Thanks for the patch but I don't like that our color output has become very complex: NO_COLOR, DENO_TERM_COLOR, --color. This is exposing a lot of complexity to every user to solve a very small issue.

Do the NO_COLOR people suggest a solution to this?

VishnuJin:
@ry 
I also feel that it would be confusing if we have 2 env vars for almost samething,  as I also mentioned this in [our previous discussion](https://github.com/denoland/deno/pull/13049#issuecomment-991279162) 
But in terms of providing a little bit of more control to the user I think options like `DENO_TERM_COLOR` is handy.

Interestingly [Cargo documentation](https://doc.rust-lang.org/cargo/index.html?search=NO_COLOR) does not have any info on `NO_COLOR` usage If I'm not wrong, as we know it supports [`CARGO_TERM_COLOR`](CARGO_TERM_COLOR) however I guess because of [`termcolor`](https://github.com/BurntSushi/termcolor/search?q=NO_COLOR) dependency `NO_COLOR` becomes valid(has an effect) in cargo

kt3k:
@VishnuJin 
I think it's not a good idea to change the entire runtime color settings (`use_color` result) based on the environment.

The original issue (#12224) says it's inconvenient to have colors in `console.log`/`console.dir` outputs when the stdout is piped. So I'd suggest we should change only `console` behavior based on tty/no-tty.

The color settings of `console` (`inspect`) is created here https://github.com/denoland/deno/blob/4d176b7b7c11aabc584bee45423f108ea47faefe/ext/console/02_console.js#L1728
Changing this config based on is_tty value should only affect console behaviors.

VishnuJin:
@kt3k some hints to implement this would help me to proceed further !!

justinmchase:
I think this is a good feature, 99% of the time you want auto but in those 1% of times, which I've definitely hit both sides it can be extremely frustrating to not have the ability to force colors on or off.

You do definitely hit rare occasions where your OS configuration is just weird or messed up and not specifying tty correctly and thats when you need these switches. It's a fit-and-finish level issue to be sure but this is probably the right way to solve it all things considered.

VishnuJin:
Closing this PR as it seems to be unnecessary !!

PR #13069, opened 2021-12-13
refactor(ext/crypto): generateKey rust cleanup

lucacasonato:
Clean up the rust side of crypto.subtle.generateKey using some enums.


bartlomieju (approved):
LGTM

PR #13071, opened 2021-12-13
feat(ext/crypto): support importing RSA JWKs

lucacasonato:
This commit adds support for importing RSA JWKs in the Web Crypto API.

Extracted out of https://github.com/denoland/deno/pull/12022, with modifications.


tools/wpt/expectation.json
```diff
@@ -3948,240 +3948,120 @@
         "Good parameters: 1024 bits (jwk, object(kty, n, e, d, p, q, dp, dq, qi), {hash: SHA-1, name: RSA-OAEP}, true, [decrypt])",
         "Good parameters: 1024 bits (jwk, object(kty, n, e, d, p, q, dp, dq, qi), {hash: SHA-1, name: RSA-OAEP}, true, [unwrapKey, decrypt])",
         "Good parameters: 1024 bits (jwk, object(kty, n, e, d, p, q, dp, dq, qi), {hash: SHA-1, name: RSA-OAEP}, true, [unwrapKey])",
```
lucacasonato, line 3950: These tests were `extractable = true` do not yet pass, because they rely on support for exporting RSA private keys (which is not implemented in this PR).

ext/crypto/import_key.rs
```diff
@@ -75,6 +92,89 @@ pub fn op_crypto_import_key(
   }
 }
 
+macro_rules! decode_jwt_b64_int {
+  ($name:ident, $b64:expr, $err:expr) => {
+    let bytes = base64::decode_config($b64, base64::URL_SAFE)
+      .map_err(|_| data_error($err))?;
+    let $name = UIntBytes::new(&bytes).map_err(|_| data_error($err))?;
+  };
+}
```
lucacasonato, line 101: This can not be turned into a function, because lifetime issues that can not be resolved without hacks or GATs.

bnoordhuis (approved):

ext/crypto/import_key.rs
lucacasonato: I changed the name to `jwt_b64_int_or_err` (short enough to still keep all invocations 1 liners, but gets the point across).

bartlomieju (approved):
LGTM too

ext/crypto/00_crypto.js
```diff
@@ -2367,6 +2371,253 @@
...
+            "'ext' property of JsonWebKey must not be false if extractable is true",
+            "DataError",
+          );
+        }
+
+        // 7.
+        let hash;
+
+        // 8.
+        switch (jwk.alg) {
```
SeanWykes, line 2467: @lucacasonato I believe you need a different set of JWK.alg for each algorithm. For example, RSA-OAEP-256 for RSA-OAEP, PS256 for RSA-PSS.

See 
[https://www.w3.org/TR/WebCryptoAPI/#rsa-pss-operations] .. jwk step 7, and 
[https://www.w3.org/TR/WebCryptoAPI/#rsa-oaep-operations] .. jwk step 8

PR #13074, opened 2021-12-13
fix(lsp): handle import specifier not having a trailing quote

dsherret:


cli/lsp/completions.rs
```diff
@@ -91,17 +94,31 @@ async fn check_auto_config_registry(
   }
 }
 
-/// Ranges from the graph for specifiers include the leading and trailing quote,
+/// Ranges from the graph for specifiers include the leading and maybe trailing quote,
 /// which we want to ignore when replacing text.
-fn to_narrow_lsp_range(range: &deno_graph::Range) -> lsp::Range {
+fn to_narrow_lsp_range(
+  text_info: &SourceTextInfo,
+  range: &deno_graph::Range,
```
dsherret, line 101: I think lines and columns is an LSP & display concern and we should remove it from deno_graph. I think deno_graph should just return byte indexes or at least have both.

kitsonk: Best to open this as an issue so we don't forget about it.

kitsonk (commented):
Some sort of test?

cli/lsp/completions.rs
dsherret: Opened https://github.com/denoland/deno_graph/issues/79

kitsonk:
Will this actually improve things at the moment?  Meaning does swc parse the document to a point where we can detect that span at all?  The test doesn't validate that we are providing code completions for something without a trailing quote, so I can determine if it does, or if we need further changes from swc?

dsherret:
@kitsonk there's a test in swc for that, but I have repl import specifier completions incoming so it will be tested in integration then.

kitsonk (approved):
LGTM

PR #13075, opened 2021-12-13
fix(ext/net): make unix and tcp identical on close

### LINKED ISSUE #13033: HTTP server on unix socket errors on close, unlike tcp equivalent
This is a followup of #8272. The server on unix socket no longer hangs, but it does throw an error. The "tcp" version does not behave like this.

Ideally unix domain socket servers would not throw an error on close.

Here is a complete sample that illustrates the difference: comment/uncomment the different listeners to observe the difference. Note you need to pass a socket file location as an argument.

```ts
import { serveListener } from "https://deno.land/std@0.117.0/http/server.ts";
import { exists} from "https://deno.land/std@0.117.0/fs/mod.ts";

const sock_path = Deno.args[Deno.args.length -1];
if( !sock_path || !sock_path.endsWith('.sock') ) throw new Error("socket path should be specified and end with sock");
console.log( sock_path );

if( await exists(sock_path) ) await Deno.remove(sock_path);

const controller = new AbortController();

const listener = Deno.listen({ path: sock_path, transport: "unix" });
// const listener = Deno.listen({port:3000});

serveListener( listener, (request) => {
	console.log("got request");
	return new Response("hello world");
}, {signal: controller.signal} );
 
console.log("server started");

await sleep(1);

console.log("closing server");
// listener.close();
controller.abort();

await sleep(1);
```
Thanks.

bnoordhuis:
std/http/server knows how to handle "Listener has been closed"
exceptions but not "operation canceled" errors.

Make "unix" listen sockets throw the same exception as "tcp" listen
sockets when the socket is closed and has a pending accept operation.

There is still a discrepancy when multiple accept requests are posted
but that's probably a less visible issue and something for another day.

Fixes #13033

### DIFF
--- a/cli/tests/unit/net_test.ts
+++ b/cli/tests/unit/net_test.ts
@@ -147,7 +147,8 @@ Deno.test(
     listener.close();
     await assertRejects(
       () => p,
-      Deno.errors.Interrupted,
+      Deno.errors.BadResource,
+      "Listener has been closed",
     );
   },
 );
@@ -185,7 +186,7 @@ Deno.test(
     const listener = Deno.listen({ transport: "unix", path: filePath });
     let acceptErrCount = 0;
     const checkErr = (e: Error) => {
-      if (e instanceof Deno.errors.Interrupted) { // "operation canceled"
+      if (e.message === "Listener has been closed") {
         assertEquals(acceptErrCount, 1);
       } else if (e instanceof Deno.errors.Busy) { // "Listener already in use"
         acceptErrCount++;
--- a/ext/net/ops.rs
+++ b/ext/net/ops.rs
@@ -103,6 +103,15 @@ pub(crate) struct AcceptArgs {
   pub transport: String,
 }
 
+pub(crate) fn accept_err(e: std::io::Error) -> AnyError {
+  // FIXME(bartlomieju): compatibility with current JS implementation
+  if let std::io::ErrorKind::Interrupted = e.kind() {
+    bad_resource("Listener has been closed")
+  } else {
+    e.into()
+  }
+}
+
 async fn accept_tcp(
   state: Rc<RefCell<OpState>>,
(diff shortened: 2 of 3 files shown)


piscisaureus:
I agree with the goal here but I think we should rather move to throw Interrupted (rather than BadResource) in case of TCP. 

I interpret @bartlomieju's TODO to mean that as well.

bartlomieju:
> I agree with the goal here but I think we should rather move to throw Interrupted (rather than BadResource) in case of TCP.
> 
> I interpret @bartlomieju's TODO to mean that as well.

This TODO means that the current message looks like it does to preserve error message we've been returning since 1.0 for backwards compatibility. 

I think there should be no remapping of error to `Listener has been closed,` as it is a special case and we don't do that for any other resource which might have been already closed.

piscisaureus:
> I think there should be no remapping of error to Listener has been closed, as it is a special case and we don't do that for any other resource which might have been already closed.

I agree, you said it better than I did.

bnoordhuis:
I expect that to break std and other projects.

Note how the test case from #13033 works okay when it's over TCP - std/http/server recognizes and handles the error.

Do you still want me to remove the special-casing?

bartlomieju:
> I expect that to break std and other projects.
> 
> Note how the test case from #13033 works okay when it's over TCP - std/http/server recognizes and handles the error.
> 
> Do you still want me to remove the special-casing?

No, I think we should keep it for now, but remove it in 2.0.

bartlomieju:
> Okay, in that case this PR is ready for review. I can add a // Remove in 2.0 comment if you want.

That or an issue with "maybe 2.0" label would be great.

bartlomieju (approved):
LGTM

bnoordhuis:
> or an issue with "maybe 2.0" label

Good idea, #13086.

PR #13078, opened 2021-12-13
feat: REPL import specifier auto-completions

### LINKED ISSUE #12408: Suggestion: REPL import specifier autocompletion
It might be useful for the REPL to autocomplete module specifiers from registries. So for example, typing...

```ts
import react from "react<TAB>
````

...would offer some suggestions for a URL to use.

Suggestion from @piscisaureus

### PR DESCRIPTION
**Pull request by dsherret**
Closes #12408

### CODE REVIEW

File: cli/lsp/repl.rs [hunk 1]
```diff
@@ -0,0 +1,308 @@
...
+
+      let items = match response {
+        Some(CompletionResponse::Array(items)) => items,
+        Some(CompletionResponse::List(list)) => list.items,
+        None => Vec::new(),
+      };
+      items
+        .into_iter()
+        .filter_map(|item| {
+          item.text_edit.and_then(|edit| match edit {
```
dsherret, line 160: In the future I believe we will be able to trivially improve this to return completions for more than just import specifier completions. Right now this somewhat limits it to just import specifier completions based on only handling this scenario.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,308 @@
...
+          root_uri: Some(cwd_uri.clone()),
+          initialization_options: Some(
+            serde_json::to_value(get_repl_workspace_settings()).unwrap(),
+          ),
+          capabilities: ClientCapabilities {
+            workspace: None,
+            text_document: None,
+            window: None,
+            general: None,
+            //offset_encoding: None,
```
bnoordhuis, line 76:
```suggestion
```
Or was it left in with a reason?

cli/lsp/repl.rs
```diff
@@ -0,0 +1,308 @@
...
+    let new_text = if new_text.ends_with('\n') {
+      new_text.to_string()
+    } else {
+      format!("{}\n", new_text)
+    };
+    self.document_version += 1;
+    let current_line_count =
+      self.document_text.chars().filter(|c| *c == '\n').count() as u32;
+    let pending_line_count =
+      self.pending_text.chars().filter(|c| *c == '\n').count() as u32;
```
bnoordhuis, line 199: Are these related to the current line of input or the full REPL history? In the latter case O(n) scans might turn problematic for long-lived sessions.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,308 @@
...
+          text: new_text.to_string(),
+        }],
+      })
+      .await;
+    self.pending_text = new_text;
+  }
+
+  async fn check_cwd_change(&mut self) {
+    // handle if the cwd changes
+    let cwd_uri = get_cwd_uri().unwrap();
```
bnoordhuis, line 222: You should probably handle the error instead of panicking. Deleting the working directory is not an unheard of thing.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,308 @@
...
+  }
+}
+
+async fn with_logging_disabled<TReturn>(
+  future: impl Future<Output = TReturn>,
+) -> TReturn {
+  logger::set_logging_enabled(false);
+  let result = future.await;
+  // ensure this is set back even on error of a Result
+  logger::set_logging_enabled(true);
```
bnoordhuis, line 263: Toggling globals like that feels somewhat unsafe. I suppose it works because this function is only called by `&mut self` methods but it's a bit brittle.

cli/tools/repl/channel.rs
```diff
@@ -46,11 +61,41 @@ impl RustylineSyncMessageSender {
...
+          method: method.to_string(),
+          params,
+        })
     {
       Err(anyhow!("{}", err))
     } else {
-      self.response_rx.borrow_mut().blocking_recv().unwrap()
+      match self.response_rx.borrow_mut().blocking_recv().unwrap() {
+        RustylineSyncResponse::PostMessage(result) => result,
+        _ => unreachable!(),
```
bnoordhuis, line 75:
```suggestion
        RustylineSyncResponse::LspCompletions(_) => unreachable!(),
```
To ensure the compiler croaks when a new variant is added to the enum.

cli/tools/repl/channel.rs
```diff
@@ -46,11 +61,41 @@ impl RustylineSyncMessageSender {
...
+        line_text: line_text.to_string(),
+        position,
+      })
+      .is_err()
+    {
+      Vec::new()
+    } else {
+      match self.response_rx.borrow_mut().blocking_recv().unwrap() {
+        RustylineSyncResponse::LspCompletions(result) => result,
+        _ => unreachable!(),
```
bnoordhuis, line 97:
```suggestion
        RustylineSyncResponse::PostMessage(_) => unreachable!(),
```

bnoordhuis (commented):
Nice work, David!

cli/lsp/repl.rs
dsherret: I updated the code to handle this, but I'm not sure it's the best way of handling it though.

cli/lsp/client.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl Client {
+  pub fn from_lspower(client: lspower::Client) -> Client {
+    Client(Arc::new(LspowerClient(client)))
+  }
+
+  pub fn new_for_repl() -> Client {
+    Client(Arc::new(ReplClient))
+  }
```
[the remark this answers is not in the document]
dsherret: Thanks! I also try to use `Self`, but sometimes I forget 🙂

bartlomieju:
I tested this locally and it's great, however I found one thing that is IMO a bit quirky:
```
> import * as asserts from "https://deno.land/std/<TAB>
Display all 442 possibilities? (y or n)
https://deno.land/std/bytes/mod.ts                                 https://deno.land/std/hash/sha1_test.ts
https://deno.land/std/examples/welcome.ts                          https://deno.land/std/http/testdata/file_server_as_library.ts
https://deno.land/std/collections/sum_of_test.ts                   https://deno.land/std/io/util_test.ts
https://deno.land/std/collections/deep_merge_test.ts               https://deno.land/std/node/tty.ts
https://deno.land/std/collections/group_by_test.ts                 https://deno.land/std/node/internal_binding/performance.ts
https://deno.land/std/http/cookie.ts                               https://deno.land/std/node/internal_binding/constants.ts
https://deno.land/std/node/internal_binding/errors.ts              https://deno.land/std/node/os_test.ts
https://deno.land/std/node/perf_hooks_test.ts                      https://deno.land/std/collections/sum_of.ts
https://deno.land/std/testing/asserts.ts                           https://deno.land/std/node/tty_test.ts
https://deno.land/std/collections/chunk_test.ts                    https://deno.land/std/node/module.ts
https://deno.land/std/examples/echo_server.ts                      https://deno.land/std/collections/filter_keys_test.ts
https://deno.land/std/node/crypto_test.ts                          https://deno.land/std/fs/exists.ts
https://deno.land/std/collections/aggregate_groups.ts              https://deno.land/std/path/basename_test.ts
https://deno.land/std/encoding/base64_test.ts                      https://deno.land/std/path/relative_test.ts
https://deno.land/std/collections/max_by.ts                        https://deno.land/std/uuid/mod.ts
https://deno.land/std/collections/drop_while.ts                    https://deno.land/std/collections/without_all.ts
https://deno.land/std/collections/sort_by_test.ts                  https://deno.land/std/datetime/tokenizer.ts
https://deno.land/std/encoding/csv_test.ts                         https://deno.land/std/log/mod_test.ts
https://deno.land/std/node/internal_binding/url.ts                 https://deno.land/std/uuid/v4.ts
https://deno.land/std/fs/test.ts                                   https://deno.land/std/ws/e
....
```
Can we make it that pressing <TAB> in this situation hints only directories instead of all discovered paths?

dsherret:
@bartlomieju it seems it also does that in the editor. Perhaps we should do a follow-up PR to change that behaviour?

bartlomieju:
Sounds good!

cli/lsp/logging.rs
```diff
@@ -0,0 +1,54 @@
...
+
+static LSP_DEBUG_FLAG: AtomicBool = AtomicBool::new(false);
+static LSP_LOG_LEVEL: AtomicUsize = AtomicUsize::new(log::Level::Info as usize);
+
+pub fn set_lsp_debug_flag(value: bool) {
+  LSP_DEBUG_FLAG.store(value, Ordering::SeqCst)
+}
+
+pub fn lsp_debug_enabled() -> bool {
+  LSP_DEBUG_FLAG.load(std::sync::atomic::Ordering::SeqCst)
```
bnoordhuis, line 15:
```suggestion
  LSP_DEBUG_FLAG.load(Ordering::SeqCst)
```

cli/lsp/logging.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+pub fn set_lsp_log_level(level: log::Level) {
+  LSP_LOG_LEVEL.store(level as usize, Ordering::SeqCst)
+}
+
+pub fn lsp_log_level() -> log::Level {
+  let level = LSP_LOG_LEVEL.load(std::sync::atomic::Ordering::SeqCst);
```
bnoordhuis, line 23:
```suggestion
  let level = LSP_LOG_LEVEL.load(Ordering::SeqCst);
```

cli/lsp/logging.rs
```diff
@@ -0,0 +1,54 @@
...
+  match level {
+    level if level == log::Level::Info as usize => log::Level::Info,
+    level if level == log::Level::Debug as usize => log::Level::Debug,
+    level if level == log::Level::Trace as usize => log::Level::Trace,
+    _ => unimplemented!(),
+  }
```
bnoordhuis, line 29: I think you can just cast to usize here? `log::Level` is `#[repr(usize)]`.

cli/lsp/repl.rs
bnoordhuis: Full history for the current session probably isn't so bad.

bnoordhuis (approved):
Nice, LGTM.

cli/lsp/logging.rs
dsherret: Thanks! I had these in the macro previously while it was a WIP and forgot to unqualify them when I moved it to a function.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,292 @@
...
+            character: line_and_column.column_index as u32,
+          },
+        },
+        work_done_progress_params: WorkDoneProgressParams {
+          work_done_token: None,
+        },
+        partial_result_params: PartialResultParams {
+          partial_result_token: None,
+        },
+        context: None,
```
kitsonk, line 141: discussion: while it isn't important import suggestions/completions, it is important for other types of completions, and it would be good to consider how to provide this information from the client.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,292 @@
...
+      })
+      .await
+      .ok()
+      .unwrap_or_default();
+
+    let items = match response {
+      Some(CompletionResponse::Array(items)) => items,
+      Some(CompletionResponse::List(list)) => list.items,
```
kitsonk, line 149: discussion: so for v2 of the registry, lists can come back with `isIncomplete` set to `true`.  This is a flag to indicate to the client that not everything is returned and the further queries are needed to refine the results.  In v2 version of the registry, this is how a registry can provide folder by folder "completion" of a URL.  The vscode client uses it to know if there is any value in querying the server again and does so in a "debounced" way.  If it is `isIncomplete` set to `false`, the server is saying you have the definitive result set and should "filter" it on the client.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,292 @@
...
+    };
+    items
+      .into_iter()
+      .filter_map(|item| {
+        item.text_edit.and_then(|edit| match edit {
+          CompletionTextEdit::Edit(edit) => Some(ReplCompletionItem {
+            new_text: edit.new_text,
+            span: lsp_range_to_span(&text_info, &edit.range),
+          }),
+          CompletionTextEdit::InsertAndReplace(_) => None,
```
kitsonk, line 160: discussion: while we don't issue registry completions with insert and replace, it doesn't mean we _can't_ in the future, and we should potentially consider handling it here.  Certainly some that come back from tsc around other completions do use this.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,292 @@
...
+      .await
+      .ok()
+      .unwrap_or_default();
+
+    let items = match response {
+      Some(CompletionResponse::Array(items)) => items,
+      Some(CompletionResponse::List(list)) => list.items,
+      None => Vec::new(),
+    };
+    items
```
kitsonk, line 152: discussion: when a completion item is a "full" URL, it is delivered with a cache command, that would allow the client to request that the specifier be cached.  You may want to consider how this is handled in the REPL.

discussion: in registry v2, items can come with a `data` property, which is a flag that indicates that the client can do a `completion/resolve` and will get back further details (in this case, fill in the `documentation` field in the completion).  For other completions from `tsc` there are often `data` fields so that the actual final edit can be calculated lazily and you have to call `completion/resolve` to get the full item.  So while it isn't critical for registry completions, you may want to consider how to handle `data` and `completion/resolve`.

kitsonk (approved):
LGTM - pending the once change about `auto_discover` that I think you should make plus a few points to think about consider.

Awesome. 💯

cli/lsp/logging.rs
bnoordhuis: Then `std::mem::transmute(level)` ought to work. Not a big thing though.

cli/lsp/repl.rs
```diff
@@ -0,0 +1,292 @@
...
+    lint: false,
+    unstable: false,
+    suggest: CompletionSettings {
+      complete_function_calls: false,
+      names: false,
+      paths: false,
+      auto_imports: false,
+      imports: ImportCompletionSettings {
+        auto_discover: true,
+        hosts: HashMap::from([("https://deno.land/x/".to_string(), true)]),
```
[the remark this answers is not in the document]
dsherret: I think I copied this original from somewhere. I can't remember where though.

cli/lsp/repl.rs
dsherret: Yeah, I think it will be useful when implementing the other kinds of completions.

cli/lsp/repl.rs
dsherret: This is related to when to show completions, right? Right now completions only happen when hitting tab in the repl so I think we wouldn't need to configure this?

cli/lsp/client.rs
```diff
@@ -0,0 +1,228 @@
...
+          json!({
+            "enable": true,
+          })
+        })
+        .collect()
+    };
+    Box::pin(future::ready(Ok(response)))
+  }
+
+  fn show_message(
```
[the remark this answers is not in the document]
dsherret: Yup—this is not doing any caching or resolving imports at moment. It would be good to do once we get other kinds of completions working.

cli/lsp/repl.rs
dsherret: I think once we handle other completions then we can work on this. It will be good to reference these comments when doing that work.

cli/lsp/repl.rs
dsherret: I think for now it shouldn't matter for the repl because it will just ask for completions whenever the user hits tab and doesn't do completions when typing characters or anything. Probably once we start doing completions like shown then we can come back to this:

image

cli/lsp/repl.rs
```diff
@@ -0,0 +1,292 @@
...
+    internal_debug: false,
+    lint: false,
+    unstable: false,
+    suggest: CompletionSettings {
+      complete_function_calls: false,
+      names: false,
+      paths: false,
+      auto_imports: false,
+      imports: ImportCompletionSettings {
+        auto_discover: true,
```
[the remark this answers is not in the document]
dsherret: Thanks!

cli/lsp/repl.rs
kitsonk: [Then it would be](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionContext):

```rust
let context = CompletionContext {
  trigger_kind: CompletionTriggerKind::Invoked,
  trigger_character: None,
};
```

cli/lsp/repl.rs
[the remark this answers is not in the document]
dsherret: Sorry, I forgot to respond to your first point. I just tried out these two other registries and the experience didn't seem as nice as deno.land. Maybe we should add them after they've improved?

Issue #12197, opened 2021-09-23
Feature request - Allow watching files outside of the module graph


bebraw:
I am reading JSON files that then affect what Deno serves. The problem is that the operation is outside of Deno's module graph and the `watch` feature won't notice these type of changes.

Based on https://github.com/denoland/deno/issues/2401#issuecomment-859901801 I was thinking patterns are supported already but looking at the source (https://github.com/denoland/deno/blob/3c97dbf06bc26fc79f0f58a360ef821f75ba82e9/cli/flags.rs#L226) they are not.

Therefore I would propose adding support to `--watch=./**/*.ts,./**/*.tsx` kind of syntax (i.e. `watch` would accept a string of globs) to allow watching arbitrary files even outside of Deno's module graph.

Related to https://github.com/jurassiscripts/velociraptor/issues/80.

bartlomieju:
> @bartlomieju let's just start with individual files and folders in the arguments to --watch, and go from there.

Sounds good to me, PRs are welcome

bartlomieju:
For anyone interested in tackling this issue, here are some some pointers how to do it:

In `cli/flags.rs`:
- change `watch_arg()` function to accept values, and update the description that passing values has only effect on the `deno run` subcommand
- change `watch` field in `Flags` struct to accept `Option<Vec<PathBuf>>` instead of a `bool`
- in `run_parse` change parsing logic to create a vector of paths
- in other functions that parse `watch` flag they should set `watch` equal to `Some(vec![])` is "watch" flag is present

In `cli/main.rs`:
- change `run_with_watch` function to extend `paths_to_watch` with the value of vector from `Flags::watch`

Add a test to `cli/tests/integration/watcher_tests.rs` that verifies that additional files passed to the flag actually cause the process to restart.

I'm going to mark this as a "good first issue", even though it's on the harder spectrum of "first issues".

jespertheend:
I'd like to have a go at this if that's ok!

lucacasonato:
@jespertheend Please do! If you need some more guidance, please hop into the #dev channel on Discord: https://discord.gg/deno

PR #13087, opened 2021-12-14
feat(watch): support watching external files

jespertheend:
Fixes: #12197

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -2548,6 +2548,23 @@ mod tests {
     );
   }
 
+
+  #[test]
+  fn run_watch_with_external() {
+    let r = flags_from_vec(svec!["deno", "run", "--watch=file1,file2", "script.ts"]);
+    let flags = r.unwrap();
+    assert_eq!(
+      flags,
+      Flags {
+        subcommand: DenoSubcommand::Run(RunFlags {
+          script: "script.ts".to_string(),
+        }),
+        watch: Some(vec![PathBuf::from("file1"), PathBuf::from("file2")]),
+        ..Flags::default()
+      }
+    );
+  }
+
   #[test]
   fn run_reload_allow_write() {
     let r =
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -2548,10 +2548,10 @@ mod tests {
     );
   }
 
-
   #[test]
   fn run_watch_with_external() {
-    let r = flags_from_vec(svec!["deno", "run", "--watch=file1,file2", "script.ts"]);
+    let r =
+      flags_from_vec(svec!["deno", "run", "--watch=file1,file2", "script.ts"]);
     let flags = r.unwrap();
     assert_eq!(
       flags,
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -549,7 +549,7 @@ fn bundle_subcommand<'a, 'b>() -> App<'a, 'b> {
         .required(true),
     )
     .arg(Arg::with_name("out_file").takes_value(true).required(false))
-    .arg(watch_arg())
+    .arg(watch_arg(false))
     .about("Bundle module and dependencies into single file")
     .long_about(
       "Output a single JavaScript file with all dependencies.
@@ -882,7 +882,7 @@ Ignore formatting a file by adding an ignore comment at the top of the file:
         .multiple(true)
         .required(false),
     )
-    .arg(watch_arg())
+    .arg(watch_arg(false))
     .arg(
       Arg::with_name("options-use-tabs")
         .long("options-use-tabs")
@@ -1158,7 +1158,7 @@ Ignore linting a file by adding an ignore comment at the top of the file:
         .multiple(true)
         .required(false),
     )
-    .arg(watch_arg())
+    .arg(watch_arg(false))
 }
 
 fn repl_subcommand<'a, 'b>() -> App<'a, 'b> {
@@ -1177,7 +1177,7 @@ fn repl_subcommand<'a, 'b>() -> App<'a, 'b> {
 fn run_subcommand<'a, 'b>() -> App<'a, 'b> {
   runtime_args(SubCommand::with_name("run"), true, true)
     .arg(
-      watch_arg()
+      watch_arg(true)
         .conflicts_with("inspect")
         .conflicts_with("inspect-brk"),
     )
@@ -1305,7 +1305,7 @@ fn test_subcommand<'a, 'b>() -> App<'a, 'b> {
         .multiple(true),
     )
     .arg(
-      watch_arg()
+      watch_arg(false)
         .conflicts_with("no-run")
         .conflicts_with("coverage"),
     )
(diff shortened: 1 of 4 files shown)


cli/flags.rs
```diff
@@ -1643,10 +1643,16 @@ fn compat_arg<'a, 'b>() -> Arg<'a, 'b> {
 fn watch_arg<'a, 'b>() -> Arg<'a, 'b> {
   Arg::with_name("watch")
     .long("watch")
+    .value_name("FILES")
+    .min_values(0)
+    .takes_value(true)
+    .use_delimiter(true)
+    .require_equals(true)
```
jespertheend, line 1650: If `require_equals` isn't set, parsing fails when trying to run without any external files:
`deno run --watch main.js`
It's not clear whether `main.js` is part of `--watch` or if it's the main entry point.
Though this does mean that in `--help` it's not super clear that external files are optional. Because it shows up as `--watch=<FILES>`

cli/flags.rs
```diff
@@ -2409,6 +2415,24 @@ fn inspect_arg_validate(val: String) -> Result<(), String> {
...
+  matches: &clap::ArgMatches,
+  allow_extra: bool,
+) {
+  flags.watch = match matches.values_of("watch") {
+    Some(f) => {
+      let extra_files = f.map(PathBuf::from).collect();
+      if allow_extra {
+        Some(extra_files)
+      } else {
+        Some(vec![])
```
jespertheend, line 2429: I thought maybe it could be nice to panic here or show some sort of error when external files are provided for any non `run` command. But I'm not sure of a good way to do this. Is a `panic!` actually acceptable, or is there a better way to handle invalid arguments that I'm not aware of?

cli/tests/integration/watcher_tests.rs
```diff
@@ -575,6 +575,47 @@ fn run_watch() {
...
+  let (mut stdout_lines, mut stderr_lines) = child_lines(&mut child);
+
+  assert_contains!(stdout_lines.next().unwrap(), "Hello world");
+  wait_for("Process finished", &mut stderr_lines);
+
+  // Change content of the external file
+  write(&external_file_to_watch, "Hello world2").unwrap();
+
+  assert_contains!(stderr_lines.next().unwrap(), "Restarting");
+  wait_for("Process finished", &mut stderr_lines);
```
jespertheend, line 615: I tried to see what happens with the test if watching doesn't work correctly, it seems like the test will hang forever in this case. Not sure if this is an issue.

cli/flags.rs
bartlomieju: Panicking would be no good, I think it's fine to leave it as is and just ignore these files.

cli/flags.rs
```diff
@@ -1643,10 +1643,16 @@ fn compat_arg<'a, 'b>() -> Arg<'a, 'b> {
...
+    .min_values(0)
+    .takes_value(true)
+    .use_delimiter(true)
+    .require_equals(true)
     .help("UNSTABLE: Watch for file changes and restart process automatically")
     .long_help(
       "UNSTABLE: Watch for file changes and restart process automatically.
-Only local files from entry point module graph are watched.",
+Watches used script files by default, or listed files when specified.
+Passing values only has an effect on the 'deno run' command.",
```
bartlomieju, line 1655:
```suggestion
Local files from entry point module graph are watched by default. Additional paths might be watched by passing them as arguments to this flag.
```

cli/flags.rs
bartlomieju: That makes sense, we already use the same setup in other flags.

bartlomieju (commented):
Great work @jespertheend! I have only a single nitpick for the help text, but other than that it looks good to me. I will let David review it as well before landing.

jespertheend:
Thanks for the review! I didn't really like my wording for the help message tbh, but couldn't think of anything better.
Your version seems much better!

cli/flags.rs
```diff
@@ -1643,10 +1643,16 @@ fn compat_arg<'a, 'b>() -> Arg<'a, 'b> {
 fn watch_arg<'a, 'b>() -> Arg<'a, 'b> {
   Arg::with_name("watch")
     .long("watch")
+    .value_name("FILES")
```
dsherret, line 1646: It seems we need to make this value conditional because it's not supported for certain commands? That will also allow us to remove some of the logic in `watch_arg_parse` and have it error when someone provides this for subcommands that don't support it.

dsherret (changes requested):
Thanks @jespertheend! Looks good overall, but I think we just need to change `watch_arg()` to not take files for subcommands that don't support it.

cli/flags.rs
jespertheend: @bartlomieju and I briefly discussed this here: https://github.com/denoland/deno/pull/13087#discussion_r769138224
I'm not really sure what would be the best way to do this. Should `watch_arg_parse` return a `Result<>` with every call to it manually handling the error? Or is there code somewhere that I can more elegantly reuse for handling argument errors?
Also, is there anything that needs to change to the `Arg::with_name` parameters to let the argument handler know it only supports a value under certain circumstances?

dsherret: It would be how the argument is defined here. So something like the following:

```rs
fn watch_arg<'a, 'b>(takes_files: bool) -> Arg<'a, 'b> {
  let arg = Arg::with_name("watch")
    .long("watch")
    .help("UNSTABLE: Watch for file changes and restart process automatically");
  if takes_files {
    arg.value_name("FILES")
      .min_values(0)
      .takes_value(true)
      .use_delimiter(true)
      .require_equals(true)
      .long_help(...goeshere...)
  } else {
    arg.long_help(...goeshere....)
  }
}
```

jespertheend: Ah interesting, I wasn't aware arguments could contain logic like this. Thanks I'll try this!

bartlomieju: @jespertheend let me know if you want to tackle that, if not I can fixup and land the PR.

jespertheend: I'm working on it atm. I'll have a patch ready in a bit.

jespertheend: It seems removing `takes_value(true)` doesn't actually cause any errors when invoked with `--watch=somefile.txt`.
So all the new patch really does is update `--help` accordingly I believe.

cli/flags.rs
```diff
@@ -2420,17 +2429,13 @@ fn watch_arg_parse(
...
-    Some(f) => {
-      let extra_files = f.map(PathBuf::from).collect();
-      if allow_extra {
-        Some(extra_files)
-      } else {
-        Some(vec![])
-      }
+  if allow_extra {
+    if let Some(f) = matches.values_of("watch") {
+      flags.watch = Some(f.map(PathBuf::from).collect());
```
bartlomieju, line 2434: Will this still work with `deno run --watch`? Is there a unit test that covers this case?

bartlomieju: Actually I see no unit test that covers `deno run --watch=file1,file2`, could you add it @jespertheend?

jespertheend: https://github.com/denoland/deno/blob/6c324acf2363e88293ab94cf3de6c9d7a264b55d/cli/tests/integration/watcher_tests.rs#L734 should cover `deno run --watch`. But now that I think about it, it seems like this shouldn't work. I'll investigate.
I'll also add the extra test 👍

cli/flags.rs
dsherret: Yeah, that's ok. As long as the help text is good now 👍

cli/flags.rs
jespertheend: Ah this works because `matches.values_of` returns an empty vec when used as `--watch` without any files.
I've added the extra unit test.

dsherret (approved):
LGTM. Thanks @jespertheend!

PR #13013, opened 2021-12-07
feat(ext/crypto): implement importKey/exportKey formats (jwk, spki, pkcs8) for ECDSA/ECDH

SeanWykes:
towards #11690, implementing importKey(jwk, spki, pkcs8)  and exportKey(jwk, spki, pkcs8) for ECDSA/ECDH*

Based on and requires refacs from PR#12022. When that lands, I'll rebase the PR. 

Note: * Curve P-256 fully implemented, P-384 only partially with support for EC Public Keys only.

PR #13088, opened 2021-12-15
feat(ext/crypto): support importing ECSDA and ECDH

SeanWykes:
importKey implemented for ECDSA and ECDH, extracted from #13013 with modifications
- pkcs8 : curve P-256 only
- spki : curves P-256 and P-384
- jwk : 
    public key - curves P-256 and P-384
    private key - curve P-256 only

Restrictions on to P-384 curve due to lack of support in p384 crate.


ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+              "'x' property of JsonWebKey is required for public keys",
+              "DataError",
+            );
+          }
+          if (jwk.y === undefined) {
+            throw new DOMException(
+              "'y' property of JsonWebKey is required for public keys",
+              "DataError",
+            );
+          }
```
lucacasonato, line 2501: This is already validated 30 lines above, and can thus be removed here.

ext/crypto/00_crypto.js
```diff
@@ -2191,50 +2195,140 @@
...
+      !ArrayPrototypeIncludes(
+        supportedNamedCurves,
+        normalizedAlgorithm.namedCurve,
+      )
+    ) {
+      throw new DOMException(
+        "Invalid namedCurve",
+        "DataError",
+      );
+    }
```
lucacasonato, line 2231: Ordering is important, because this is observable. This check happens as the first step for the "raw" algorithm, but only later for JWK.

Please move this to all the relevant spots in the different algorithms.

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
 
         return key;
       }
+      case "jwk": {
+        const jwk = keyData;
+
+        const keyType = (jwk.d != undefined) ? "private" : "public";
```
lucacasonato, line 2349:
```suggestion
        const keyType = (jwk.d !== undefined) ? "private" : "public";
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+            (u) => !ArrayPrototypeIncludes(supportedUsages[keyType], u),
+          ) !== undefined
+        ) {
+          throw new DOMException("Invalid key usages", "SyntaxError");
+        }
+
+        // 3.
+        if (jwk.kty !== "EC") {
+          throw new DOMException(
+            "`kty` member of JsonWebKey must be `RSA`",
```
lucacasonato, line 2364:
```suggestion
            "'kty' property of JsonWebKey must be 'EC'",
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+            "DataError",
+          );
+        }
+
+        // 4.
+        if (keyUsages.length > 0 && jwk.use) {
```
lucacasonato, line 2370:
```suggestion
        if (keyUsages.length > 0 && jwk.use !== undefined) {
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+          if (jwk.use !== supportedUsages.jwkUse) {
```
lucacasonato, line 2371: You can collapse these two if statements into one.

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+            throw new DOMException(
+              "`use` member of JsonWebKey must be `" + supportedUsages.jwkUse +
+                "`",
```
lucacasonato, line 2374:
```suggestion
              `'use' property of JsonWebKey must be '${supportedUsages.jwkUse}'`,
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+              "DataError",
+            );
+          }
+        }
+
+        // 5.
+        // Section 4.3 of RFC7517
+        if (jwk.key_ops) {
```
lucacasonato, line 2382:
```suggestion
        if (jwk.key_ops !== undefined) {
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+          ) {
+            throw new DOMException(
+              "`key_ops` member of JsonWebKey is invalid",
+              "DataError",
+            );
+          }
+        }
+
+        // 6.
+        if (jwk.ext === false && extractable == true) {
```
lucacasonato, line 2409:
```suggestion
        if (jwk.ext === false && extractable === true) {
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+          throw new DOMException(
+            "`ext` member of JsonWebKey is invalid",
```
lucacasonato, line 2411:
```suggestion
            "'ext' property of JsonWebKey must not be false if extractable is true",
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+            "DataError",
+          );
+        }
+
+        // 9.
+        if (jwk.alg && normalizedAlgorithm.name == "ECDSA") {
```
lucacasonato, line 2417:
```suggestion
        if (jwk.alg !== undefined && normalizedAlgorithm.name == "ECDSA") {
```

ext/crypto/00_crypto.js
```diff
@@ -2249,6 +2343,187 @@
...
+              throw new DOMException(
+                "Curve algorithm not supported",
+                "DataError",
+              );
+          }
+
+          if (algNamedCurve) {
+            if (algNamedCurve !== normalizedAlgorithm.namedCurve) {
+              throw new DOMException(
+                "Mismatched curve algorithm for JWK.alg",
```
lucacasonato, line 2443:
```suggestion
                "Mismatched curve algorithm",
```

ext/crypto/01_webidl.js
```diff
@@ -129,6 +129,17 @@
 
   webidl.converters.EcKeyGenParams = webidl
     .createDictionaryConverter("EcKeyGenParams", dictEcKeyGenParams);
+  const dictEcImportParams = [
+    ...dictAlgorithm,
```
lucacasonato, line 133:
```suggestion
    .createDictionaryConverter("EcKeyGenParams", dictEcKeyGenParams);

  const dictEcImportParams = [
    ...dictAlgorithm,
```

ext/crypto/import_key.rs
```diff
@@ -11,6 +11,13 @@ use crate::shared::*;
 use crate::OaepPrivateKeyParameters;
 use crate::PssPrivateKeyParameters;
 
+use p256::elliptic_curve::sec1::ToEncodedPoint;
+use pkcs8::{FromPrivateKey, ToPrivateKey};
```
lucacasonato, line 15:
```suggestion
use pkcs8::FromPrivateKey;
use pkcs8::ToPrivateKey;
```

ext/crypto/import_key.rs
```diff
@@ -11,6 +11,13 @@ use crate::shared::*;
...
+
+use p256::elliptic_curve::generic_array::{
+  typenum::U32, ArrayLength, GenericArray,
+};
```
lucacasonato, line 19:
```suggestion
use p256::elliptic_curve::generic_array::typenum::U32;
use p256::elliptic_curve::generic_array::ArrayLength;
use p256::elliptic_curve::generic_array::GenericArray;
```

ext/crypto/import_key.rs
```diff
@@ -657,11 +673,103 @@ fn import_key_rsaoaep(
...
+          let secret_key = p256::SecretKey::from_bytes(&d)?;
+
+          secret_key.to_pkcs8_der().unwrap()
+        }
+        //@todo(sean) - build p384 secret key from jwk, when crate implements to_pkcs8_der
+        /*EcNamedCurve::P384 => {
+          let secret_key = p384::SecretKey::from_bytes(&d)?;
+
+          secret_key.to_pkcs8_der().unwrap()
+        }*/
```
lucacasonato, line 738: p384 does implement this. `p384::SecretKey` is a `elliptic_curve::SecretKey`, which has a `to_pkcs8_der` method: https://docs.rs/elliptic-curve/0.11.5/elliptic_curve/struct.SecretKey.html#method.to_pkcs8_der.

ext/crypto/import_key.rs
```diff
@@ -657,11 +673,103 @@ fn import_key_rsaoaep(
   }
 }
 
+fn decode_b64url_to_gen_array<T: ArrayLength<u8>>(
+  b64: &str,
+) -> Result<GenericArray<u8, T>, deno_core::anyhow::Error> {
+  jwt_b64_int_or_err!(val, b64, "invalid b64 coordenate");
```
lucacasonato, line 679:
```suggestion
  jwt_b64_int_or_err!(val, b64, "invalid b64 coordinate");
```

ext/crypto/import_key.rs
```diff
@@ -657,11 +673,103 @@ fn import_key_rsaoaep(
...
+          oid,
+          parameters: None,
+        },
+        secret_key_der.as_ref(),
+      );
+
+      let pki = p256::pkcs8::PrivateKeyInfo {
+        public_key: Some(&point_bytes),
+        ..pki
+      };
```
lucacasonato, line 758: This is specific to p256.

ext/crypto/import_key.rs
```diff
@@ -685,12 +793,81 @@ fn import_key_ec(
...
+          let point =
+            secret_key.public_key().as_affine().to_encoded_point(false);
+
+          // 3.
+          if point.is_identity() {
+            return Err(data_error("Invalid key data"));
+          }
+        }
+        EcNamedCurve::P384 => {
+          //@todo(sean-wykes) Validate P384 secret-key on import(pkcs8)
```
lucacasonato, line 814: This needs to be done before landing.

ext/crypto/import_key.rs
```diff
@@ -685,12 +793,81 @@ fn import_key_ec(
...
+            secret_key.public_key().as_affine().to_encoded_point(false);
+          // 3.
+          if point.is_identity() {
+            return Err(type_error("Invalid key data".to_string()));
+          }*/
+
+          // for now, just use PKCS8 bytes
+        }
+      };
+
```
lucacasonato, line 828: I don't think this PKCS8 importing is quite correct. This should look more like the RSA-PSS PKCS8 importing. I.e. first parse to a `pkcs8::PrivateKeyInfo`, then validate the OID + parameters as per spec etc.

The check in the spec that confirms that the named curve the user specified, and the one in the DER match is not present for example.

ext/crypto/import_key.rs
```diff
@@ -685,12 +793,81 @@ fn import_key_ec(
...
+      Ok(ImportKeyResult::Ec {
+        raw_data: RawKeyData::Public(data),
+      })
+    }
+    KeyData::Spki(data) => {
+      let pk_info: spki::SubjectPublicKeyInfo =
+        spki::SubjectPublicKeyInfo::try_from(data.as_ref()).unwrap();
+      let pk = pk_info.subject_public_key;
```
lucacasonato, line 836: Same goes for SPKI. There should be a check here that checks the algorithm OID and the namedCurve are correct. Look at how this is done for RSA-PSS (checking OID and parameters).

lucacasonato (commented):
Good start. Thanks @SeanWykes.

ext/crypto/import_key.rs
SeanWykes: @lucacasonato The p384 is missing relevant functionality, specifically the ProjectiveArithmetic trait, which is a dependency for to_pkcs8_der().

ext/crypto/import_key.rs
SeanWykes: Missing ProjectiveArithmetic in p384 crate. For now, have changed code to throw an "Unsupported curve" error, instead of blindly importing the PKCS8.

ext/crypto/import_key.rs
SeanWykes: OK. Thanks for the hint. I've been focussing on functionality first .. now getting into the validation stuff. Just committed for SPKI, PKCS8 next on list.

ext/crypto/import_key.rs
lucacasonato: :+1:

ext/crypto/import_key.rs
SeanWykes: done. A bit messy 'cos `spki` and `pkcs8` crates use different versions of `der::asn1` constants, so transformed my OID constants into strings and doing string compares which I know is not optimal. Any ideas?

One deviation from spec, in step 10 is that I do not explicitly parse the pkcs#1 private_key DER from the PKCS8, just pass the pkcs8 blob strait to `SecretKey` during checking step 12, assuming that SecretKey with validate the pkcs1 structure as it extracts `d`.

Finally, as for step 12 (validating that the key is a valid point on the curve, any EC experts around to chip in? @tarcieri?

ext/crypto/Cargo.toml
```diff
@@ -19,11 +19,12 @@ base64 = "0.13.0"
 block-modes = "0.8.1"
 deno_core = { version = "0.110.0", path = "../../core" }
 deno_web = { version = "0.59.0", path = "../web" }
-elliptic-curve = "0.10.6"
+elliptic-curve = "0.11.5"
 lazy_static = "1.4.0"
 num-traits = "0.2.14"
-p256 = { version = "0.9.0", features = ["ecdh"] }
-p384 = "0.8.0"
+p256 = { version = "0.10.0", features = ["ecdh", "pem"] }
```
lucacasonato, line 25: What do you need the `pem` feature for?

ext/crypto/Cargo.toml
```diff
@@ -19,11 +19,12 @@ base64 = "0.13.0"
...
+p384 = "0.9.0"
+pkcs8 = "0.8.0"
```
lucacasonato, line 27: What do you need this for? It should be re-exported from various crates.

lucacasonato (commented):
Are the version bumps necessary? I would rather wait for them until `rsa` has gotten a new release, so we don't duplicate so many dependencies.

SeanWykes:
> Are the version bumps necessary? I would rather wait for them until `rsa` has gotten a new release, so we don't duplicate so many dependencies.

Not as such .. bumped during attempts to improve p384 coverage (pkcs8 etc). Will back-out the changes ..

lucacasonato:
@SeanWykes I have unbumped the crates locally - will push that shortly.

ext/crypto/Cargo.toml
SeanWykes: Don't need PEM. It was the only way I could find to satisfy constraint for trait `FromPrivateKey` to use `from_pkcs8_der` on `p256::SecretKey<NistP256>`. Line ~730 import_key.rs

lucacasonato: Yeah, have unfortunately discovered this myself now during testing.

lucacasonato (commented):
There are some WPT regressions it seems that I need to be dealt with before merge.

I will try to get this landed for 1.17.0 tomorrow. If it doesn't make it, I'll try to land it for 1.17.1 in a week from now. Maybe we can have web crypto "done" by the end of the year.

tarcieri:
I can take a look at getting the `rsa` crate bumped to the latest versions of `pkcs1`/`pkcs8`

ext/crypto/import_key.rs
tarcieri: > A bit messy 'cos spki and pkcs8 crates use different versions of der::asn1 constants, so transformed my OID constants into strings and doing string compares which I know is not optimal.

Can you be more specific about the problems you’re having here? Code examples and compiler errors would be helpful

SeanWykes:
> There are some WPT regressions it seems that I need to be dealt with before merge. (I can do this)
> 
> I will try to get this landed for 1.17.0 tomorrow. If it doesn't make it, I'll try to land it for 1.17.1 in a week from now. Maybe we can have web crypto "done" by the end of the year.

@lucacasonato WPT - import_export/ec will only pass the supported cases when #13104 is in, since the tests need exportKey to function. PTAL @ #13104 it should hopefully be in reasonable shape. ;)

lucacasonato:
> import_export/ec will only pass the supported cases when #13104 is in

I am aware. I meant the cases in `WebCryptoAPI/sign_verify` (I have no idea yet what is going on there).

SeanWykes:
> > import_export/ec will only pass the supported cases when #13104 is in
> 
> I am aware. I meant the cases in `WebCryptoAPI/sign_verify` (I have no idea yet what is going on there).
@lucacasonato, I think there maybe something strange going on with key-hash combinations. In `op_crypto_sign_key ECDSA`, the comment states underlying `ring::signature` crate only supports two algo pairs 
      // We only support P256-SHA256 & P384-SHA384. These are recommended signature pairs.
      // https://briansmith.org/rustdoc/ring/signature/index.html#statics
However, the hash param passed from js is checked (SHA-256 or SHA-384), independent of the curve used.

lucacasonato:
Yeah, just found that too. This code needs a general rewrite in the future.

I'll clean up this PR now, and then land it.

lucacasonato (approved):
LGTM, thanks @SeanWykes.

ext/crypto/import_key.rs
SeanWykes: yes. can clean-up when p384 support implemented, maybe using generic impl. as per work in progress on Rust Crypto [here](https://github.com/RustCrypto/elliptic-curves/issues/240)

ext/crypto/import_key.rs
```diff
@@ -684,13 +823,179 @@ fn import_key_ec(
...
+      )
+      .map_err(|_| data_error("malformed parameters"))?;
+
+      // 8-9.
+      let pk_named_curve = match params.named_curve_alg {
+        // id-secp256r1
+        ID_SECP256R1_OID => Some(EcNamedCurve::P256),
+        // id-secp384r1
+        ID_SECP384R1_OID => Some(EcNamedCurve::P384),
+        // id-secp384r1
```
SeanWykes, line 859: // id-secp521r1

ext/crypto/import_key.rs
SeanWykes: @tarcieri - code fixed by @lucacasonato, so no longer relevant to PR. 
I was getting compiler error "maybe different versions of `const_oids` due to trying to mix `ObjectIdentifier` from `pkcs8::der' and `spki::der'.
Thanks.

SeanWykes:
> LGTM, thanks @SeanWykes.

@lucacasonato - thanks for your help, cleanups and fixes. Learnt a lot about paradigmatic Rust from this exercise.

PR #13090, opened 2021-12-15
feat(ext/ffi): support alias names for symbol definitions

### LINKED ISSUE #13089: ext/ffi: support alias names for symbol definitions
**Suggestion**

An optional field `name` (symbol's name) to Foreign Function interface, which would default to the name of key in the function definitions in second parameter.

**Why**

- Allows making a blocking variant and a non-blocking variant of same symbol. Such as a `readFileSync` and `readFile`.
- Allows adding multiple overloads for same symbol. C functions can use varargs to do that. [objc_msgSend for example](https://developer.apple.com/documentation/objectivec/1456712-objc_msgsend/).
- Allows renaming the symbol in some other convention, such as snake_case to camelCase.

**Example**

```ts
const lib = Deno.dlopen("Kernel32", {
  readFileSync: {
    name: "ReadFileA",
    // ...
  },
  readFile: {
    name: "ReadFileA",
    nonblocking: true,
    // ...
  },
  // Still works
  ReadFile: {
    // ...
  },
});
```

DjDeveloperr:
fix #13089

### DIFF
--- a/test_ffi/tests/test.js
+++ b/test_ffi/tests/test.js
@@ -32,7 +32,11 @@ assertThrows(
 );
 
 const dylib = Deno.dlopen(libPath, {
-  "print_something": { parameters: [], result: "void" },
+  "printSomething": {
+    name: "print_something",
+    parameters: [],
+    result: "void",
+  },
   "print_buffer": { parameters: ["pointer", "usize"], result: "void" },
   "print_buffer2": {
     parameters: ["pointer", "usize", "pointer", "usize"],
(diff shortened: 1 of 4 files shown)


test_ffi/tests/test.js
```diff
@@ -19,7 +19,11 @@ try {
 }
 
 const dylib = Deno.dlopen(libPath, {
-  "print_something": { parameters: [], result: "void" },
+  "printSomething": {
+    name: "print_something",
+    parameters: [],
+    result: "void",
+  },
```
bartlomieju, line 26: Would be great to have a test for multiple overloads of a method

ry (commented):
Seems like a reasonable change to me. BTW your sqlite project is very cool!

test_ffi/tests/test.js
DjDeveloperr: Cannot use variadic arguments in stable rust. I also tried doing it in C but cannot re-export extern functions as symbols either...

DjDeveloperr: @bartlomieju can I instead add test case for multiple overloads that don't have different arguments [1] but one is sync and other one is async (using the `nonblocking` property)? It would be useful for making, for instance, an async variant of a FFI function alongside blocking one (such as `read` & `readSync`)

[1] which needs variadic arguments that cannot be used in stable Rust.

bartlomieju: Sure, sounds good

DjDeveloperr: I added that but CI is failing on unrelated things...

littledivy: @DjDeveloperr It is probably failing because this PR updates the `std` submodule too, is that intentional?

DjDeveloperr: No it's not intentional, seems like it updated itself when I pulled in main and resolved merge conflicts.

bartlomieju:
@DjDeveloperr please rebase

piscisaureus (approved):

bartlomieju (approved):
LGTM, to me too!

PR #13092, opened 2021-12-15
fix(ext/crypto): various cleanup in JWK imports

lucacasonato:
This aligns all of the error messages, and makes falsey comparisons
more strict.


bartlomieju (approved):
LGTM

PR #13094, opened 2021-12-15
refactor(ext/crypto): clean up encrypt rust code

lucacasonato:
Cleans up the rust side of `crypto.subtle.encrypt`, and properly asyncifies it.

### DIFF
--- a/ext/crypto/encrypt.rs
+++ b/ext/crypto/encrypt.rs
@@ -1,0 +1,138 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use crate::shared::*;
+use block_modes::BlockMode;
+use deno_core::error::type_error;
+use deno_core::error::AnyError;
+use deno_core::OpState;
+use deno_core::ZeroCopyBuf;
+use rand::rngs::OsRng;
+use rsa::pkcs1::FromRsaPublicKey;
+use rsa::PaddingScheme;
+use rsa::PublicKey;
+use serde::Deserialize;
+use sha1::Digest;
+use sha1::Sha1;
+use sha2::Sha256;
+use sha2::Sha384;
+use sha2::Sha512;
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EncryptOptions {
+  key: RawKeyData,
+  #[serde(flatten)]
+  algorithm: EncryptAlgorithm,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase", tag = "algorithm")]
+pub enum EncryptAlgorithm {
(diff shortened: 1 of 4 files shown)


ext/crypto/encrypt.rs
```diff
@@ -0,0 +1,138 @@
...
+      label: Some(label),
+    },
+    ShaHash::Sha512 => PaddingScheme::OAEP {
+      digest: Box::new(Sha512::new()),
+      mgf_digest: Box::new(Sha512::new()),
+      label: Some(label),
+    },
+  };
+  let encrypted = public_key
+    .encrypt(&mut rng, padding, &data)
```
bnoordhuis, line 98:
```suggestion
    .encrypt(&mut rng, padding, data)
```

ext/crypto/encrypt.rs
```diff
@@ -0,0 +1,138 @@
...
+  Ok(buf.into())
+}
+
+fn encrypt_rsa_oaep(
+  key: RawKeyData,
+  hash: ShaHash,
+  label: Vec<u8>,
+  data: &[u8],
+) -> Result<Vec<u8>, deno_core::anyhow::Error> {
+  let label = String::from_utf8_lossy(&label).to_string();
```
bnoordhuis, line 69:
```suggestion
  let label = String::from_utf8_lossy(&label).into_owned();
```
Otherwise the Cow always heap-allocates, I think?

bnoordhuis (approved):

bnoordhuis:
Related test failure:

> TypeError: Error parsing args: unknown variant `AES-CBC`, expected `RSA-OAEP` or `AES`

ext/crypto/encrypt.rs
lucacasonato: `Cow.into_owned` returns a `Cow` not a `String`. `Cow<str>` turning into `String` would require specialization I think.

bnoordhuis: That doesn't sound right. `Cow::into_owned()` returns `ToOwned::Owned`, which for `str` is `String` (because `str` implements `ToOwned`.)

lucacasonato: My compiler disagrees. When using `to_owned`:

```
  --> ext/crypto/encrypt.rs:79:19
   |
79 |       label: Some(label),
   |                   ^^^^^- help: try using a conversion method: `.to_string()`
   |                   |
   |                   expected struct `std::string::String`, found enum `Cow`
   |
   = note: expected struct `std::string::String`
                found enum `Cow<'_, str>`
```

bnoordhuis: /me whispers *in*to_owned :)

PR #13096, opened 2021-12-15
chore: update cli/Cargo.toml

bartlomieju:


nayeemrmn:
We need a `deno_doc` release as well

bartlomieju:
> We need a `deno_doc` release as well

Is it necessary? Since `deno_graph` had a patch release, updating using `cargo update -p deno_graph` should be enough in this case.

dsherret (approved):

PR #13103, opened 2021-12-16
feat(ext/net): Add Conn.setNoDelay and Conn.setKeepAlive

### LINKED ISSUE #13057: TCP setNoDelay not implemented
[Redis uses `setNoDelay`](https://gist.github.com/apowers313/fb4723691e825e734261b7c14cd58081) which is currently [not implemented in Deno](https://deno.land/std@0.115.0/node/internal_binding/tcp_wrap.ts#L282). Any objections to fixing this? Looks like a simple matter of plumbing the call [from v8](https://github.com/nodejs/node/blob/1c4df352b77514e93a46fb27e996755d45a0b530/src/tcp_wrap.cc#L98) through Rust? Maybe a good first issue to tackle?

### LINKED ISSUE #13543: Feature request: Keep-alive for TCP
From [RFC1122 4.2.3.6](https://datatracker.ietf.org/doc/html/rfc1122#page-101), 

```
Implementors MAY include "keep-alives" in their TCP
implementations, although this practice is not universally
accepted.  If keep-alives are included, the application MUST
be able to turn them on or off for each TCP connection, and
they MUST default to off.
```

Currently, `Deno.listen` and `Deno.connect` occasionally seem to become stale when kept open for a while without any packets (a waiting connection). It doesn't close the connection, and doesn't error out, just becomes unresponsive on both sides until you try to write a packet. This causes me to be unable to free up resources from zombie connections.

As far as I understand, this is due to not sending keep-alive probes. Node has the [`socket.setKeepAlive`](https://nodejs.org/api/net.html#socketsetkeepaliveenable-initialdelay) API that I'm missing on Deno.

Implementing this might simply be setting the `SO_KEEPALIVE`, `TCP_KEEPIDLE`, `TCP_KEEPCNT`, `TCP_KEEPINTVL` options on the socket [`Linux`](https://linux.die.net/man/3/setsockopt) [`Windows`](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt).

A possible Deno API may look like `conn.setKeepAlive(interval)`

### Further reading:

I found this article that seems to describe the exact behaviour I'm facing, and the solution (keep-alive): https://holmeshe.me/network-essentials-setsockopt-SO_KEEPALIVE/#To-detect-an-absent-peer

yos1p:
Implement setNoDelay function for Deno

Closes #13057
Closes #13543


yos1p:
@bnoordhuis Can I ask your opinion on this implementation? Any suggestions or changes to make? Thanks!

ext/net/01_net.js
```diff
@@ -58,6 +58,10 @@
     return core.opAsync("op_dns_resolve", { query, recordType, options });
   }
 
+  function opSetNoDelay(rid, noDelay) {
+    return core.opAsync("op_set_nodelay", { rid, noDelay });
```
bnoordhuis, line 62: I'd make this sync, it doesn't need to be async. See op_net_listen for an example of a sync op.

ext/net/lib.deno_net.d.ts
```diff
@@ -50,6 +50,8 @@ declare namespace Deno {
     /** Shuts down (`shutdown(2)`) the write side of the connection. Most
      * callers should just use `close()`. */
     closeWrite(): Promise<void>;
+    /** Enable/disable the use of Nagle's algorithm. */
+    setNoDelay(nodelay: boolean): Promise<void>;
```
bnoordhuis, line 54: Maybe make it default to true?

bnoordhuis (commented):
Thanks for the PR. Left some comments but mostly LGTM.

Disabling Nagle isn't directly observable so there isn't a good way to verify it's actually working but there should at least be a test that verifies the API method exists.

ext/net/ops.rs
```diff
@@ -665,6 +666,26 @@ where
...
+    .get::<TcpStreamResource>(rid)?;
+  resource.set_nodelay(nodelay).await
+}
+
+#[derive(Deserialize)]
+pub struct NoDelayArgs {
+  rid: ResourceId,
+  nodelay: bool,
+}
+
```
AaronO, line 688: This struct is no longer necessary if we flatten op args (which is preferred if possible)

```suggestion
```

ext/net/ops.rs
```diff
@@ -665,6 +666,26 @@ where
...
+  _: (),
+) -> Result<(), AnyError> {
+  let rid = args.rid;
+  let nodelay = args.nodelay;
+  let resource: Rc<TcpStreamResource> = state
+    .borrow_mut()
+    .resource_table
+    .get::<TcpStreamResource>(rid)?;
+  resource.set_nodelay(nodelay).await
+}
```
AaronO, line 681: Simpler function sig and no destructuring with flattened args:

```suggestion
pub async fn op_set_nodelay<NP>(
  state: Rc<RefCell<OpState>>,
  rid: ResourceId,
  nodelay: bool,
) -> Result<(), AnyError> {
  let resource: Rc<TcpStreamResource> = state
    .borrow_mut()
    .resource_table
    .get::<TcpStreamResource>(rid)?;
  resource.set_nodelay(nodelay).await
}
```

ext/net/01_net.js
AaronO: Flatten args in opcall:

```suggestion
    return core.opAsync("op_set_nodelay", rid, noDelay);
```

Given that there's a single call site and with flattened args it's just a passthrough function, you could probably remove `opSetNoDelay` and just do the opcall in `Conn::setNoDelay`

AaronO (dismissed):
It's best to best opcall args directly without any wrapping when possible (both for performance and simplicity). Given `op_set_nodelay` takes 2 args `(rid, true/false)` there's no need to wrap that in a composite struct since opcalls have a user-available arity of 2.

I added suggestions of the required changes below that you can commit

ext/net/01_net.js
yos1p: To modify the TcpStream, I need to use `self.wr_borrow_mut()` which returns `AsyncBorrowFutureImpl`, which means I have to use async/await. Is there any workaround for that? I'm quite new with AsyncRefCell in Deno. Would be nice to receive help on this, thanks!

ext/net/ops.rs
yos1p: Good suggestion. Thanks!

ext/net/01_net.js
yos1p: Changed to sync. Let me know if the implementation is correct / according to style.

bnoordhuis (approved):
LGTM, thanks! I'll dismiss Aaron's review, his feedback has been addressed.

bnoordhuis:
This still needs a test, by the way.

bartlomieju (commented):
Are we sure this should be a stable API from the get go?

yos1p:
> Are we sure this should be a stable API from the get go?

I'm not sure what should qualify for this to be included in the stable API. It's true that currently I don't think we can write a test for this since we need to get `TcpStream` from the resource table.

bartlomieju:
> I don't think we can write a test for this since we need to get TcpStream from the resource table.

What is the problem with getting `TcpStream` from the resource table?

yos1p:
> What is the problem with getting TcpStream from the resource table?

The idea that I have for the unit test is to call Deno.connect, get the TcpStream, and assert that nodelay and keepalive are disabled. Then, call op_set_nodelay and op_set_keepalive function and assert that nodelay and keepalive is enabled. 

But I don't know how to call that in test environment. It's a combination of calling Deno code, and get the TcpStream via Rust.

yos1p:
The unit test implementation for the op_set_nodelay and op_set_keepalive is a little bit much. Not sure what you guys think. Please advise if it's necessary or how to make it more simple. I'm open to suggestions.

@bnoordhuis @bartlomieju @AaronO

bartlomieju:
> > What is the problem with getting TcpStream from the resource table?
> 
> The idea that I have for the unit test is to call Deno.connect, get the TcpStream, and assert that nodelay and keepalive are disabled. Then, call op_set_nodelay and op_set_keepalive function and assert that nodelay and keepalive is enabled.
> 
> But I don't know how to call that in test environment. It's a combination of calling Deno code, and get the TcpStream via Rust.

Ah I see what you mean here. Indeed this is rather involved to check. Let me think about it and I'll get back to you.

cli/tests/unit/net_test.ts
```diff
@@ -238,6 +238,78 @@ Deno.test({ permissions: { net: true } }, async function netTcpDialListen() {
...
+  assert(conn.localAddr != null);
+  const buf = new Uint8Array(1024);
+  const readResult = await conn.read(buf);
+  assertEquals(3, readResult);
+  assertEquals(1, buf[0]);
+  assertEquals(2, buf[1]);
+  assertEquals(3, buf[2]);
+  assert(conn.rid > 0);
+
+  assert(readResult !== null);
```
lucasfcosta, line 268: Considering that:

1. L262 already checks whether `readResult` equals `3`
2. `readResult` is a const and assigned a primitive

Do we need this assertion or is it redundant?

cli/tests/unit/net_test.ts
```diff
@@ -238,6 +238,78 @@ Deno.test({ permissions: { net: true } }, async function netTcpDialListen() {
...
```
lucasfcosta, line 304: Same here, since `readResult` is a `const` to which a primitive gets assigned and whose value we assert upon on L298, isn't this assertion redundant?

Apologies if I'm missing anything here (this is also probably just a nitpick anyway, so I'm sorry).

bartlomieju (approved):
LGTM, thank you @yos1p!

PR #13104, opened 2021-12-16
feat(ext/crypto) - exportKey(pkcs8/spki/jwk) for ECDSA and ECDH

SeanWykes:
Towards #11690 - exportKey implemented for ECDSA and ECDH, extracted from #13013 with modifications

pkcs8 : curve P-256 and P-384
spki : curves P-256 and P-384
jwk :
- public key - curves P-256 and P-384
- private key - curve P-256 and P-384

Shares `ec_key.rs` with #13154.


SeanWykes:
@lucacasonato - strange behaviour for WPT import_export ECDH SPKI P-256. The W3C spec says to export with  `OID 1.3.132.1.12` but the test appears to require the standard ecPublicKey OID.
BTW: Maybe need to patch importKey SPKI /PKCS8 to accept both OID, using something like ````
      
if alg != elliptic_curve::ALGORITHM_OID {
        let ok = match algorithm {
          ImportKeyOptions::Ecdh{ .. } => {
            // id-ecDh
            alg != ObjectIdentifier::new("1.3.132.1.12")
          },
          _ => false,
        };

        if !ok {
          return Err(data_error("unsupported algorithm"));
        }
      }````

panva:
@SeanWykes export `ecPublicKey`, see https://github.com/w3c/webcrypto/pull/305. The spec editors are in the process of aligning the editor's draft of the specification with what is commonly actually implemented. None of the existing implementations use support importing or exporting `id-ecDH`.

SeanWykes:
> @SeanWykes export `ecPublicKey`, see [w3c/webcrypto#305](https://github.com/w3c/webcrypto/pull/305). The spec editors are in the process of aligning the editor's draft of the specification with what is commonly actually implemented. None of the existing implementations use support importing or exporting `id-ecDH`.

@panva  thanks .. backed out previous commit.

ext/crypto/export_key.rs
```diff
@@ -168,3 +189,156 @@ fn export_key_rsa(
...
+          let point = key_data.as_ec_public_key_p256()?;
+
+          point.as_ref().to_vec()
+        }
+        EcNamedCurve::P384 => {
+          let point = key_data.as_ec_public_key_p384()?;
+
+          point.as_ref().to_vec()
+        }
+        _ => return Err(data_error("Unsupported named curve")),
```
littledivy, line 216: Exhaustive matching helps catch bugs. Ditto for the `match`'s below
```suggestion
        EcNamedCurve::P521 => return Err(data_error("Unsupported named curve")),
```

ext/crypto/00_crypto.js
```diff
@@ -3226,6 +3230,103 @@
...
+        return data.buffer;
+      }
+      case "jwk": {
+        // 1-2.
+        const jwk = {
+          kty: "EC",
+        };
+
+        // 3.
+        if (key[_algorithm].name == "ECDSA") {
```
littledivy, line 3280: Based on the spec, a similar `switch` for ECDH should be there. It is possible to manipulate `key[_algorithm].namedCurve` by the user which in the current case will produce incorrect `jwk.crv`.

ext/crypto/00_crypto.js
```diff
@@ -3226,6 +3230,103 @@
...
+        ObjectAssign(jwk, data);
+
+        // 7.
+        jwk.key_ops = key.usages;
+
+        // 8.
+        jwk.ext = key[_extractable];
+
+        // 9
+        jwk.crv = key[_algorithm].namedCurve;
```
littledivy, line 3321: These steps numbers here don't exist for ECDH and It's a little confusing for someone whose following along side-by-side with the spec. IMO it's just better to refactor out ECDH and ECDSA code into seperate `case`s

littledivy (changes requested):

ext/crypto/export_key.rs
```diff
@@ -168,3 +189,156 @@ fn export_key_rsa(
     }
   }
 }
+
+fn bytes_to_b64(bytes: Vec<u8>) -> String {
```
littledivy, line 193:
```suggestion
fn bytes_to_b64(bytes: &[u8]) -> String {
```

ext/crypto/export_key.rs
```diff
@@ -168,3 +189,156 @@ fn export_key_rsa(
...
+      EcNamedCurve::P256 => {
+        let point = key_data.as_ec_public_key_p256()?;
+        let coords = point.coordinates();
+
+        if let p256::elliptic_curve::sec1::Coordinates::Uncompressed { x, y } =
+          coords
+        {
+          Ok(ExportKeyResult::JwkPublicEc {
+            x: bytes_to_b64(x.to_vec()),
+            y: bytes_to_b64(y.to_vec()),
```
littledivy, line 259:
```suggestion
            x: bytes_to_b64(x),
            y: bytes_to_b64(y),
```
Ditto for other calls that involve `GenericArray<u8, N>`

littledivy (changes requested):

ext/crypto/00_crypto.js
SeanWykes: For ECDH, jwk.alg is a constant ECDH so no `switch` needed I think. Will fix missing constant in next commit.

As for manipulating `key[_algorithm]`, manipulating this is possible all over the current implementation, for example,  `key[_algorithm].namedCurve` is used during `deriveBits` and `sign( ECDSA )`, or `key[_algorithm].hash.name`.

We could fix this in the general case by adding curve and hash info to Rust KeyData structures, but I think this is more of a major refactor. PLMK what you think.

ext/crypto/00_crypto.js
SeanWykes: Will do. thanks for the tip.

ext/crypto/export_key.rs
SeanWykes: nice. missed this one.

ext/crypto/00_crypto.js
littledivy: Alrighty It's fine then :+1: Separating ECDH and ECDSA impl (below comment) should fix the confusion

SeanWykes:
@littledivy and @lucacasonato  - I'd like your feedback on the lack of support for PKCS#8 SecretKey in p384 crate that currently affects both importKey and exportKey. 

I'm testing a strategy to solve this:
1) ExportKey(JWK private) -> parse PKCS#8 ECPrivateKey using der::asn1 to extract priv/pub parts.
2) ImportKey(PKCS#8) -> same code to extract, then use ring::pkcs8 to build the key, thus checking valid key before import.
2) ImportKey(JWK private) -> use der::asn1 to build ECPrivateKey from priv/pub, wrap in PKCS#8 PrivateKeyInfo. Then validate using ring::pkcs8.

Any ideas on this?

littledivy:
> parse PKCS#8 ECPrivateKey using der::asn1 to extract priv/pub parts.

Sounds good. We do similar stuff with RSA key parameters.

SeanWykes:
> > parse PKCS#8 ECPrivateKey using der::asn1 to extract priv/pub parts.
> 
> Sounds good. We do similar stuff with RSA key parameters.

@littledivy export implemented here and import in #13154. The ec_key.rs code is the same.

tools/wpt/expectation.json
```diff
@@ -794,12 +780,6 @@
...
       "ec_importKey.https.any.worker.html": [
-        "Good parameters: P-256 bits (spki, buffer(91), {name: ECDSA, namedCurve: P-256}, true, [])",
-        "Good parameters: P-256 bits (jwk, object(kty, crv, x, y), {name: ECDSA, namedCurve: P-256}, true, [])",
-        "Good parameters: P-256 bits (pkcs8, buffer(138), {name: ECDSA, namedCurve: P-256}, true, [sign])",
-        "Good parameters: P-256 bits (jwk, object(kty, crv, x, y, d), {name: ECDSA, namedCurve: P-256}, true, [sign])",
-        "Good parameters: P-384 bits (spki, buffer(120), {name: ECDSA, namedCurve: P-384}, true, [])",
-        "Good parameters: P-384 bits (jwk, object(kty, crv, x, y), {name: ECDSA, namedCurve: P-384}, true, [])",
         "Good parameters: P-384 bits (pkcs8, buffer(185), {name: ECDSA, namedCurve: P-384}, true, [sign])",
         "Good parameters: P-384 bits (jwk, object(kty, crv, x, y, d), {name: ECDSA, namedCurve: P-384}, true, [sign])",
         "Good parameters: P-384 bits (pkcs8, buffer(185), {name: ECDSA, namedCurve: P-384}, false, [sign])",
```
littledivy, line 785: Why are these still failing?

SeanWykes: The test attempts to do a round-trip import/export and fails during import, since importKey must check key-validity and the `p384` crate doesn't currently support this. This is fixed in #13154

littledivy (approved):
LGTM, thank you! ...regarding the addition WPT changes in unwrap/wrapKey, looks like they were false positive earlier? I guess we'll get a better idea in #13154

PR #13106, opened 2021-12-16
feat(fetch): Support abort reasons in fetch

andreubotella:
The spec PR for this (whatwg/fetch#1343) hasn't been finalized yet because some questions regarding service worker integration are not settled yet, but since that is not relevant to Deno and everything else seems settled, and it seems best to not wait for another Deno version to support abort reasons in the fetch API.


lucacasonato (approved):
LGTM, thanks.

PR #13108, opened 2021-12-16
Revert "chore: upgrade to rustls 0.20 (#12488)"

### LINKED ISSUE #13058: Fetching certain HTTPS requests fails with "unexpected end of file"
```js
const res = await fetch("https://wpt.live/resources/testharness.js");
await res.text();
```

```
Uncaught Error: request or response body error: error reading a body from connection: unexpected end of file
    at async Object.pull (deno:ext/fetch/26_fetch.js:108:24)
```

This doesn't fail with every HTTPS request – https://wpt.live/resources/ works, for example – but it seems to be consistent when it does. It doesn't happen with HTTP at all.

This seems to have been introduced in the upgrade to rustls 0.20 in #12488.

andreubotella:
This reverts commit a2f1357fe81fe212451567cc433c0e6b4a85edc5.

Closes #13058.


bnoordhuis:
@lucacasonato Your call.

bartlomieju:
I think we don't need to land this PR right now.

andreubotella:
Agreed

PR #13114, opened 2021-12-16
fix(cli/tools/coverage): Split sources by char index

### LINKED ISSUE #13111: Coverage is broken in deno_std
```
thread 'main' panicked at 'byte index 3777 is not a char boundary; it is inside '‘' (bytes 3776..3779) of `import { isWindows, osType } from "../_util/os.ts";
import { SEP, SEP_PATTERN } from "./separator.ts";
import * as _win32 from "./win32.ts";
import * as _posix from "./posix.ts";
const path = isWindows ? _win32 : _posix;
const { join, normalize } = path;
c`[...]', cli/tools/coverage.rs:281:11
```

https://github.com/denoland/deno_std/runs/4549822141?check_suite_focus=true

nayeemrmn:
I think coverage values use char index? Fixes #13111.

### DIFF
--- a/cli/tools/coverage.rs
+++ b/cli/tools/coverage.rs
@@ -231,9 +231,12 @@ impl CoverageReporter for LcovCoverageReporter {
         continue;
       }
 
-      let source_line = script_source[0..function.ranges[0].start_offset]
-        .split('\n')
-        .count();
+      let source_line = script_source
+        .chars()
+        .take(function.ranges[0].start_offset)
+        .filter(|c| *c == '\n')
+        .count()
+        + 1;
 
       let line_index = if let Some(source_map) = maybe_source_map.as_ref() {
         source_map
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
Thanks!

bartlomieju (changes requested):
I want to get a test case before landing it. I will try to come up with a reproduction for a problem observed in `deno_std`. It seems to involve transpilation and source maps...

bartlomieju:
I tried this PR against `deno_std` and it indeed fixes the problem; however I'm still having trouble coming up with a test that would reproduce the problem.

bartlomieju (approved):
I'm going to land this without a test as it fixes the immediate problem. Coverage tools is due to an overhaul anyway and we should address the tests there.

PR #13119, opened 2021-12-17
feat(ext/crypto): implement AES-GCM encryption

littledivy:
Towards #11690 

Discussed on Discord, Only 96-bit nonce is supported due to type limitations until we have this lifted upstream.


ext/crypto/encrypt.rs
```diff
@@ -136,3 +159,59 @@ fn encrypt_aes_cbc(
...
 }
+
+fn encrypt_aes_gcm(
+  key: RawKeyData,
+  length: usize,
+  tag_length: usize,
+  iv: Vec<u8>,
+  additional_data: Option<Vec<u8>>,
+  data: &[u8],
+) -> Result<Vec<u8>, deno_core::anyhow::Error> {
```
bartlomieju, line 170: Is there a specific reason to use `deno_core::anyhow::Error` here instead of `AnyError`?

littledivy: Fixed. I'm not sure though It was being used for other `*_decrypt` functions too

ext/crypto/00_crypto.js
```diff
@@ -3429,6 +3431,70 @@
...
+        if (normalizedAlgorithm.iv.byteLength !== 12) {
+          throw new DOMException(
+            "Initialization vector length not supported",
+            "NotSupportedError",
+          );
+        }
+
+        // 3.
+        if (normalizedAlgorithm.additionalData !== undefined) {
+          if (normalizedAlgorithm.additionalData.byteLength > (2 ** 64) - 1) {
```
bnoordhuis, line 3456: Probably hypothetical but doesn't this depend on the tag size?

bnoordhuis (approved):
LGTM with a request and a question.

bartlomieju:
@littledivy please rebase

ext/crypto/00_crypto.js
littledivy: Yeah, length of AAD ≤ (2 ^ 64) −1 applies to 128 bit tags, which is the largest tag size here.

[Appendix C](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf) has some constraints for very small tags but the webcrypto spec doesn't seem to care about enforcing it.

ext/crypto/00_crypto.js
```diff
@@ -3429,6 +3431,70 @@
         // 4.
         return cipherText.buffer;
       }
+      case "AES-GCM": {
+        normalizedAlgorithm.iv = copyBuffer(normalizedAlgorithm.iv);
+
+        // 1.
+        if (data.byteLength > (2 ** 39) - 256) {
```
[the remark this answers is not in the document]
bnoordhuis: Ping @littledivy - can you add those comments? Thanks.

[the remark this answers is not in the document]
littledivy: Irrespective of the nonce/IV, GCM limits plaintext to be <= 2 ^ 39 - 256. 
image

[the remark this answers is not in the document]
bnoordhuis: Oh, you're right. It's because the internal counter is 32 bits wide, encrypts 128 or 2**7 bits each turn, therefore you can't go beyond 2**(32+7==39) without wrapping.

(The -256 is because of the tag and because the counter starts at 1, I think.)

Anyway, objection withdrawn. :-)

SeanWykes:
@littledivy are you working on AES-GCM decrypt?

littledivy:
Yes

PR #13120, opened 2021-12-17
refactor(ext/crypto): cleanup decrypt code

littledivy:
Tries to do it in the same fashion as #13094


ext/crypto/decrypt.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase", tag = "algorithm")]
+pub enum DecryptAlgorithm {
+  #[serde(rename = "RSA-OAEP")]
+  RsaOaep {
+    hash: ShaHash,
+    #[serde(with = "serde_bytes")]
+    label: Option<Vec<u8>>,
```
lucacasonato, line 34: I don't think any of the callsites ever pass `null`.

littledivy: you're right, fixed

PR #13121, opened 2021-12-17
fix(repl): support assertions on import & export declarations

### LINKED ISSUE #13115: Support import assertions in the REPL
Just need to transform it correctly:

```ts
> import test from "./test.json" assert { type: "json" };
Uncaught TypeError: Expected a "JavaScript" module but loaded a "JSON" module.
    at async <anonymous>:2:28
```

dsherret:
image

Closes #13115


bartlomieju (approved):
LGTM

Issue #13110, opened 2021-12-16
print chrome://inspect/ when --inspect/--inspect-brk flag is present


ry:

bartlomieju:
Aditionally for `--inspect-brk` there should be a prompt saying that `Deno is waiting for debugger to connect`

PR #13123, opened 2021-12-17
fix: inspector prompts

bartlomieju:
This commit fixes prompts printed to the terminal when
running with "--inspect" or "--inspect-brk" flags.

When debugger disconnects error is no longer printed as
users don't care about the reason debugger did disconnect.

A message suggesting to go to "chrome://inspect" is printed 
if debugger is active.

Additionally and information that process is waiting for
debugger to connect is printed if running with "--inspect-brk"
flag.

Closes https://github.com/denoland/deno/issues/13110
Closes https://github.com/denoland/deno/issues/13124

### DIFF
--- a/cli/tests/integration/inspector_tests.rs
+++ b/cli/tests/integration/inspector_tests.rs
@@ -97,6 +97,8 @@ async fn inspector_break_on_first_line() {
 
   use TestStep::*;
   let test_steps = vec![
+    StdErr("Visit chrome://inspect to connect to the debugger."),
+    StdErr("Deno is waiting for debugger to connect."),
     WsSend(r#"{"id":1,"method":"Runtime.enable"}"#),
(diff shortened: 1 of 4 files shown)


bartlomieju:
Currently Deno does this:
```
node --inspect-brk foo.js
Debugger listening on ws://127.0.0.1:9229/ws/173192a5-2d9d-430e-a5e6-4f271675325f
Visit chrome://inspect to connect to the debugger.
Deno is waiting for debugger to connect.
```

While Node does:
```
node --inspect-brk foo.js
Debugger listening on ws://127.0.0.1:9229/84790fe2-fffe-4896-a33a-e575cec30ab0
For help, see: https://nodejs.org/en/docs/inspector
```

Maybe we should link to https://deno.land/manual/getting_started/debugging_your_code instead?

ry (approved):
LGTM - i think this can be a "fix" rather than a "feature"

PR #13122, opened 2021-12-17
[WIP] feat: Add "deno jupyter" subcommand

bartlomieju:
Closes https://github.com/denoland/deno/issues/13016

TODO:
- [x] Add support for `DEBUG` env variable
- [ ] Make stdio out less flaky (don't use channel for all messages)
- [ ] Investigate `BrokenPipe` errors we're getting and how we could improve that situation (https://github.com/zeromq/zmq.rs/issues/153)

PR #13131, opened 2021-12-18
refactor(repl): factor out ReplEditor and ReplSession

bartlomieju:
Just a small refactor splitting `repl` into more modules; this should come in handy in #13122


dsherret (approved):
+1. LGTM

PR #13135, opened 2021-12-18
refactor: use `once_cell` instead of `lazy_static`

littledivy:
### Motivation 

* `rustc` warnings for unused statics unlike `lazy_static`. (this PR removes some dead code btw)
* `once_cell` API proposed for inclusion in std. (https://github.com/rust-lang/rfcs/pull/2788)
* Macro free & better compile times.
* And [noticeably faster](https://github.com/async-rs/async-std/issues/406#issuecomment-547286625) too

Note: we already had `once_cell` as dependency before


dsherret (approved):
LGTM.

bartlomieju (approved):
LGTM too; we still it as a direct dependency in two of our crates:
- `deno_doc`
- `deno_graph`
and as transitive dependency in multiple other crates, but it seems like `lazy_static` is being slowly fazed out.

PR #13140, opened 2021-12-19
feat(cli): Add eye-catchers module

### LINKED ISSUE #13001: Make eye-catchers explicit
Through Deno codebase we got a bunch of so called "eye-catchers", ie. prompt messages that get printed during lifecycle of the problem. Most popular eye-catchers are `Download ...` and `Check ...` messages that users see when first downloading dependencies and type checking the code. 

Unfortunately these messages are scattered all across the codebase as simple `log::info!()` calls. We should centralise them in a single module (`cli/eye_catchers.rs`) to make them more explicit and easier to change should the need arise.

Here are some examples in the code that print these messages:

https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/main.rs#L725

https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/proc_state.rs#L518-L522

https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/file_watcher.rs#L110-L113

https://github.com/denoland/deno/blob/d31378726e78490e88b8a9ec3001b86ea009d978/cli/file_watcher.rs#L159-L162

dzvon:
Fixes #13001. 

If there is anything needs to be changed, please let me know. :D

### DIFF
--- a/cli/emit.rs
+++ b/cli/emit.rs
@@ -368,7 +368,7 @@ pub(crate) fn check_and_maybe_emit(
       let root_str = root.to_string();
       // `$deno` specifiers are internal, don't print them.
       if !root_str.contains("$deno") {
-        eye_catchers::check(&format!("{}", root));
+        eye_catchers::check(root.as_str());
       }
     }
   }
--- a/cli/emit.rs
+++ b/cli/emit.rs
@@ -7,11 +7,11 @@
 use crate::ast;
 use crate::cache::CacheType;
 use crate::cache::Cacher;
-use crate::colors;
 use crate::config_file::ConfigFile;
 use crate::config_file::IgnoredCompilerOptions;
 use crate::config_file::TsConfig;
 use crate::diagnostics::Diagnostics;
+use crate::eye_catchers;
 use crate::flags;
 use crate::graph_util::GraphData;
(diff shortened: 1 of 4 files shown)


cli/emit.rs
```diff
@@ -368,7 +368,7 @@ pub(crate) fn check_and_maybe_emit(
       let root_str = root.to_string();
       // `$deno` specifiers are internal, don't print them.
       if !root_str.contains("$deno") {
-        log::info!("{} {}", colors::green("Check"), root);
+        eye_catchers::check(&format!("{}", root));
```
dsherret, line 371: Previously if someone had log level set to not do anything for info then it wouldn't allocate a string here. I don't think that really matters though because these are "info" logs and not something like "debug" or "trace". That said, I wonder if it would be better to make all these macros anyway? Then we wouldn't have to use a `format!` macro. Thoughts @bartlomieju?

bartlomieju: I don't think it's a big deal, that said, it seems `format!` is not needed here anyway

dzvon: Yeah, it's not needed here. I've removed it already.

PR #13144, opened 2021-12-20
fix(wpt_ecoch): Don't schedule on forks

littledivy:
Runs either fail or keep running for days. I noticed this recently:
- https://github.com/littledivy/deno/actions/runs/1600071948
- https://github.com/littledivy/deno/actions/runs/1597394500


ry (approved):
Thanks

PR #13147, opened 2021-12-20
feat(ext/fetch) Added RequestMethod type

BoredSathvik:
It is for better suggestions only, does not change anything else


lucacasonato:
I am not sure. What is the point here?

lucacasonato (commented):
`fetch` does not support the `CONNECT` method, so that should definitely not be listed.

BoredSathvik:
> I am not sure. What is the point here?

Just for auto suggest

SeparateRecords:
Does this work in practice? I was under the impression that TS will simplify `"a" | "b" | string` to just `string` — was this changed recently?

To get this to work in one of my projects I had to do `"a" | "b" | (string & {})`, which I don't think is desirable to be built in

Issue #13095, opened 2021-12-15
Cleanup core/modules.rs


bartlomieju:
After landing https://github.com/denoland/deno/pull/12866 there's some cruft in `core/modules.rs` that should be cleaned up. Opening for tracking purposes.

PR #13149, opened 2021-12-20
refactor: core/modules.rs

bartlomieju:
Closes https://github.com/denoland/deno/issues/13095


core/modules.rs
```diff
@@ -26,7 +26,7 @@ use std::task::Context;
 use std::task::Poll;
 
 pub type ModuleId = i32;
-pub type ModuleLoadId = i32;
+pub(crate) type ModuleLoadId = i32;
 
 pub const BOM_CHAR: char = '\u{FEFF}';
```
bartlomieju, line 31: This should be private to the module, just like `strip_bom` function

dsherret (approved):
LGTM

PR #13151, opened 2021-12-20
refactor(ext/web): Don't rely on NaN comparisons in `TextEncoderStream`

andreubotella:
In the `transform` function to `TextEncoderStream`'s internal `TransformStream`, if `chunk` is the empty string and `this.#pendingHighSurrogate` is null, then `lastCodeUnit` will be NaN. As it turns out, this does not cause a bug because the comparison to check for lone surrogates turns out to be false for NaN, but to rely on it makes the code brittle.

### DIFF
--- a/ext/web/08_text_encoding.js
+++ b/ext/web/08_text_encoding.js
@@ -292,6 +292,9 @@
         transform: (chunk, controller) => {
           try {
             chunk = webidl.converters.DOMString(chunk);
+            if (chunk === "") {
+              return PromiseResolve();
+            }
             if (this.#pendingHighSurrogate !== null) {
               chunk = this.#pendingHighSurrogate + chunk;
             }


bartlomieju (commented):
Is there a test that could be used to check this change?

andreubotella:
The NaN comparison wasn't actually causing any difference in behavior from what would be expected, since the comparison to check whether `chunk` ends with a lead surrogate (`0xD800 <= lastCodeUnit && lastCodeUnit <= 0xDBFF`) turns out to be false if `lastCodeUnit` is NaN, which is what you'd expect if `chunk` is empty. As the commit message says, this refactoring removes the NaN comparison to make the code less brittle.

Returning early if `chunk` is the empty string but `this.#pendingHighSurrogate` is not null is also not a behavior difference, since `this.#pendingHighSurrogate` is either null or a string containing a single lead surrogate code unit, so that code unit would be prepended to `chunk` and then removed and stored into `this.#pendingHighSurrogate` again.

bartlomieju (approved):
Thanks Andreu

Issue #13156, opened 2021-12-06
lsp: JSDoc imports not recognised


jespertheend:
**Describe the bug**
When the deno extension is enabled, typed imports via `/** @type {import("...` stop working.

**To Reproduce**

1. Create `Bar.js`:
```js
export class Bar {}
```

2. Create `Foo.js`:
```js
/** @type {import("./Bar.js").Bar} */
const foo = null;
```
3. Hover over `foo` and note it's type is set to `Bar`.
4. Enable deno with `Deno: Initialize Workspace Configuration`
5. Hover over `foo` again.

**Expected behavior**
The type remains set to `Bar`, rather than `any`.

**Screenshots**

**Versions**

vscode: 1.62.3 deno: 1.16.3 extension: 3.9.2

kitsonk:
We need to enhance `deno_graph` with https://github.com/denoland/deno_graph/issues/91 and when it gets folded back into the CLI, it should "work like magic".

jespertheend:
I think I have a fix that might be easier. All tests seem to pass but I'm not sure if this causes any imports to resolve even though they shouldn't.

kitsonk:
@jespertheend as I stated in the PR, it won't work for remote modules.  We need to information in module graph.

jespertheend:
I just found out this duplicates #1136

jespertheend:
Thanks so much for this @kitsonk! I'm pretty much only using js files with jsdoc imports all over the place. So without this the Deno extension was basically unusable for me.

PR #13157, opened 2021-12-20
fix(lsp) handle jsdoc imports

jespertheend:
Fixes https://github.com/denoland/deno/issues/13156


kitsonk:
This is not going to work for remote modules, as they need to be resolved in the cache.  We need to update deno_graph to include it in the dependency graph.

PR #13159, opened 2021-12-21
feat(lsp): supply accept header when fetching registry config

### LINKED ISSUE #13153: Fetching a registry config should use an accept header
@lucacasonato and I chatted and realised that when fetching the registry config, especially with the introduction of v2, we should have an accepts header with a media type that indicates that it is a "v2" capable client, so a registry can server a v1 or v2 config based on the client support.

I propose we use `application/vnd.deno.reg.v2+json` when fetching the config.

kitsonk:
Closes #13153

This will set the following header when fetching a registry configuration:

```
Accept: application/vnd.deno.reg.v2+json, application/vnd.deno.reg.v1+json;q=0.9, application/json;q=0.8
```

This will allow servers to understand that a language server client would prefer a v2 version of the registry configuration.

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -474,6 +474,7 @@ impl FileFetcher {
       maybe_headers: Some(headers),
     })
   }
+
   /// Asynchronously fetch remote source file specified by the URL following
   /// redirects.
   ///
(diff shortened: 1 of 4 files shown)


dsherret (approved):
LGTM

PR #13162, opened 2021-12-21
feat(ext/ffi): Synchronous Callbacks

littledivy:
Closures are needed for completely porting `deno_webview` from plugin API to FFI. It's also needed for certain sqlite3 bindings.

Example usage:
```c
// funnylib.c
#include <stdio.h>
#include <stdint.h>

typedef intptr_t (*JsCallBack)(intptr_t a, intptr_t b);

void call_add(JsCallBack cb) {
  printf("[C] Calling JS callback\n");
  intptr_t retval = cb(2, 3);
  printf("[C] Callback returned %lu\n", retval);
}
```
```typescript
// usage.ts
const lib = Deno.dlopen("./libfunnylib.so", {
  "call_add": {
    // Closure signature.
    parameters: [ { function: { parameters: ["i64", "i64"], result: "i64" } } ],
    result: "void",
  },
});

lib.symbols.call_add((a, b) => {
  console.log("[JS] Callback called");
  return a + b;
});
```
Output:
```bash
[C] Calling JS callback
[JS] Callback called
[C] Callback returned 5
```

This PR only adds support for _one_ closure argument per symbol. I did this so we don't have to deal with `libffi::middle::Closure` de-allocating when it goes out of the iterator scope, there is already quite a lot of `unsafe` going on here so I'm avoiding more complexity. It can be added in follow-up PRs if needed.


core/ops.rs
```diff
@@ -193,7 +193,7 @@ impl DerefMut for OpState {
 
 /// Collection for storing registered ops. The special 'get_op_catalog'
 /// op with OpId `0` is automatically added when the OpTable is created.
-pub struct OpTable(IndexMap<String, Rc<OpFn>>);
+pub struct OpTable(pub IndexMap<String, Rc<OpFn>>);
```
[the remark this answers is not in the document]
littledivy: woops, the remains of late night debugging. reverted.

core/ops_json.rs
```diff
@@ -64,6 +64,28 @@ where
   })
 }
 
+pub fn op_sync_cb<F, A, R>(op_fn: F) -> Box<OpFn>
+where
+  F: Fn(
+      Rc<RefCell<OpState>>,
```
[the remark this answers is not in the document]
littledivy: Added doc comments on top of it. Basically it like `op_sync` but the 2nd parameter is a `v8::Local<'_, v8::Function>` to support callbacks

littledivy:
This is unsound when the scope handle outlives the `op` (usually when the callback is called from another thread or assumed to be `'static` by FFI code). A solution to this involves `v8::Locker` but that isn't implemented yet.

piscisaureus:
The issue with this approach is that when you use `tokio::spawn_blocking` to make an FFI call from the thread pool, and that FFI function calls the callback, the callback will also be invoked from the thread pool, which currently will cause a crash. I suppose this is where the perceived need to use v8::Locker comes from.

However I think we should never call into v8 from the thread pool. Rather, the v8 callback should always be called from the main thread.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -122,7 +122,8 @@ declare namespace Deno {
 
   /** A foreign function as defined by its parameter and result types */
   export interface ForeignFunction {
-    parameters: NativeType[];
+    parameters:
```
aapoalas, line 125: I would perhaps go with:
```ts
interface CallbackParameter extends Omit<ForeignFunction, "nonblocking"> {
  type: "function";
}

type ForeignFunctionParameter = NativeType | CallbackParameter;

interface ForeignFunction {
  parameters: ForeignFunctionParameter[];
}
```

ie. Avoid the extra `{ function: {}}` object wrap, and instead detect the parameter type directly from the object's type field.

Though, I don't know if this will cause some unforeseen issues in the Rust binding.

ext/ffi/00_ffi.js
```diff
@@ -142,6 +142,59 @@
...
+        if (typeof arg !== "function") {
+          throw new TypeError(
+            `Expected function got ${typeof arg}`,
+          );
+        }
+
+        parameters.push(i);
+        const fnTypes = type.function.parameters;
+        fn = (...args) => {
+          const ptrIndices = fnTypes.map((ty, i) => ty == "pointer" ? i : -1)
```
aapoalas, line 164: Some optimization:
```ts
fn = (...args) => {
  for (let i = 0; i < args.length; i++) {
    const ty = args[i];
    if (ty === "pointer") {
      args[i] = new UnsafePointer(unpackU64[ty]);
    }
  }
  return arg(...args);
}
```

ext/ffi/00_ffi.js
```diff
@@ -142,6 +142,59 @@
     }
   }
 
+  function prepareParameters(args, types) {
+    const parameters = [];
+    const buffers = [];
+    let fn = null;
```
aapoalas, line 148: I don't have the necessary understanding in Rust nor Deno's FFI codebase but I wonder if callbacks should be kept similarly as parameters and buffers in an array? This is of course my personal use case talking, since I know of an FFI function I really want to call from Deno that takes three callbacks as parameters...

aapoalas: Very much a side note: I was wrong, I do not have a three-callback FFI function.

andreubotella:
In conversation with @aapoalas in Discord, the fact that JS callbacks are only safe when called from the main thread came up. It seems like calling them from a different thread segfaults, but it would be best to detect that (can we do that with TLS?) and show a descriptive error message. (Note that a regular panic will not do, because unwinding through FFI is UB.)

ext/ffi/00_ffi.js
littledivy: This won't work. `args[i]` is the actual value and not the type.

littledivy: Updated:
```javascript
fn = (...args) => {
  for (let i = 0; i < fnTypes.length; i++) {
    const ty = fnTypes[i];
    if (ty === "pointer") {
      args[i] = new UnsafePointer(unpackU64(args[i]));
    }
  }
  return arg(...args);
}
```

aapoalas:
I started wondering if it would be better to have all FFI-passing callbacks be explicitly registered with a separate API, something like `Deno.registerCallback(cb, definition)`.

The callback could either be backed up by a special op allowing functions to pass through the op layer (like `op_sync_cb` in this PR) or by eg. a `Deno.core` binding to avoid allowing functions in the op layer. The registering would essentially return an RID that could then be used by the actual C callback to access and call the JS callback in the binding layer.

The upside of such an API choice would be:
1. Explicitness. Management of callback lifetimes becomes the implementer's responsibility and can be easily reasoned about in the ops layer.
2. Unified API surface. Persistent callbacks and thread-safe callbacks will need to be explicitly registered anyway. With this API choice those extra features are easily done on the same API path.

Downsides are:
1. More friction in the DX.
2. Global handles from user calls, bad usage leads to memory leaks.
3. FFI related type conversions might need to make their way to core.

aapoalas:
I see you've added code to support registered callbacks at a proper level, though currently only made possible by memory leaking. I think the code could be massively simplified if registering of FFI callbacks was made explicit, each registered callback creating its own `RegisteredCallbackResource` (`UnsafeRegisteredCallback` or something like that on JS side) that would then be usable as a parameter in `op_ffi_call` without the current extra `op_ffi_call_cb` op.

Most importantly, explicitly registering the FFI callbacks would enable the code to be rid of the `std::mem::forget(closure)` and `std::mem::forget(info)` calls, ie. the memory leaks. The registered callback resource would own the `closure` and `info`, keeping them safe and sound for as long as the callback is being used. This would also pave the way for thread-safe callbacks.

ext/ffi/lib.rs
```diff
@@ -583,23 +650,71 @@ fn ffi_call(args: FfiCallArgs, symbol: &Symbol) -> Result<Value, AnyError> {
...
+          ),
+        _ => unreachable!(),
+      };
+
+      let closure =
+        libffi::middle::Closure::new(cif, deno_ffi_callback, &info);
+      call_args[offset] = Arg::new(closure.code_ptr());
+      
+      std::mem::forget(global);
+      // So this is a bad API. We need to ensure that the destructor for the closure is not run.
```
aapoalas, line 706: With purely sync callbacks, you'd explicitly want `closure` and `info` to be removed once `call_symbol` has returned.

It makes a lot of sense that the `.code_ptr()` does not consume the closure, as the pointer is just a raw pointer to the C function created for the closure, and closure owns the C function. The closure must stay alive as long as the C function is in use.

For synchronous callbacks I think this should be just:
```rs
// I guess this is explicitly needed since we're working with a raw global pointer?
std::mem::forget(global);
return Ok(call_symbol(symbol, call_args));
// allow closure, info go out of scope
```
And for the more general registered callbacks it would be something like:

```rs
fn op_register_callback(
  state: OpState,
  args: ForeignFunction, // omit name
  cb: serde_v8::Value<'_>
) {
  let closure = libffi::middle::Closure::new(cif, deno_ffi_callback, info);
  let resource = RegisteredCallbackResource {
    closure,
    info,
    parameters,
    result,
  };
  Ok(state.resource_table.add::<RegisteredCallbackResource>(resource))
}

fn ffi_call(
  //...
) {
  //...
  for (native_type, value) in symbol
    .parameter_types
    .clone()
    .into_iter()
    .zip(args.parameters.into_iter())
 {
    match native_type {
      NativeType::Function { .. } => {
        if let Some(rid) = value.as_u32() {
          native_values.push(state.resource_table.get::<RegisteredCallbackResource>(rid)?); // Error should be handled already in JS land: Keep track of when a registered callback is removed and mark the removal on the instance. RID itself is of course hidden inside the instance and is only accessible through eg. a Symbol shared with the FFI call handler
        }
      }
}
```

etc.

Oliver-makes-code:
Do we know if this PR is dead? Or is it on hold for the foreseeable future? I'd really like to see this come through and be merged

Symbitic:
> Do we know if this PR is dead? Or is it on hold for the foreseeable future? I'd really like to see this come through and be merged

I second this. I was working on the Deno Qt/QML bindings. It was abandoned because Qt needs to run on the main thread, and the `while(!quit) update_sdl_in_native()` loop used in <https://github.com/littledivy/deno_sdl2> doesn't detect close and quits correctly with QML.

Until I have a way of registering slots (aka callbacks) with this or #13573, the entire Deno runtime is frozen.

bartlomieju:
This PR is not dead, but there are prerequisites in `rusty_v8` that need to be addressed beforehand

aapoalas:
Is the idea to use weak handles for registered callbacks?

littledivy:
Closing in favour of https://github.com/denoland/deno/pull/14663

PR #13163, opened 2021-12-21
fix(coverage): don't run type checks

### LINKED ISSUE #13161: `deno coverage` shouldn't run type checks
Consider the following test files
```js
//jsdom.js

import {JSDOM} from "https://esm.sh/jsdom@19.0.0";

export function createDiv() {
	const jsdom = new JSDOM();
	return jsdom.window.document.createElement("div");
}
```
```js
//jsdom.test.js

import {createDiv} from "./jsdom.js";

Deno.test({
	name: "jsdom",
	sanitizeOps: false,
	sanitizeResources: false,
	fn: () => {
		const div = createDiv();
		console.log(div);
	},
});
```
This fails with `deno test` because jsdom has a `.d.ts` file that imports from `'vm'`.
But this can be fixed with `deno test --no-check`.

If you try to generate a coverage report however:
```
deno test --coverage=cov --no-check
deno coverage cov
```
It fails with the same error because the coverage command doesn't have support for the `--no-check` flag.

jespertheend:
Fixes #13161

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -1271,6 +1271,9 @@ async fn coverage_command(
     return Err(generic_error("No matching coverage profiles found"));
   }
 
+  let mut flags = flags.clone();
+  flags.check = CheckFlag::None;
+
   tools::coverage::cover_files(flags, coverage_flags).await?;
   Ok(0)
 }


bartlomieju:
@jespertheend thank you for the PR, but this is not really the right approach. As suggested by @nayeemrmn in https://github.com/denoland/deno/issues/13161#issuecomment-998891097 the `coverage` command should be rewritten in such a way to directly interact with cache instead of going through `prepare_module_load` flow. I expect it to be quite an involved refactor, do you want to try and tackle that?

jespertheend:
Ah sorry, I had meant to close this PR. I don't think I'm capable of doing such a big refactor yet.

PR #13172, opened 2021-12-21
feat(ext/websocket): server automatically handle ping/pong for incoming WebSocket

### LINKED ISSUE #13169: Add support for automatically handling ping/pong for incoming WebSocket
We should add the ability to automatically ping / pong with the client to discover broken connections in the absence of other messages on the socket. I propose the following: we add an options bag to `Deno.upgradeWebSocket` that looks as follows:

```ts
interface UpgradeWebSocketOptions {
  /**
   * A timeout after which the connection should be closed if no
   * messages have been received from the other side. If no user
   * messages have been received after half of the time specified
   * here has passed, the server will send a `ping` frame to the
   * client. If the client does not respond to this frame with a
   * `pong` within the timeout specified, the connection is deemed
   * unhealthy and is closed. The `error` event will be emitted.
   *
   * The default is 120 seconds. Set to 0 to disable timeouts.
   */
  idleTimeout?: number;
}
```

The implementation should follow the flow described in doc comment above.

Originally discussed in #13168. cc @milaz @crowlKats

crowlKats:
Closes #13169


ext/websocket/01_websocket.js
```diff
@@ -446,6 +456,41 @@
...
+            PromisePrototypeThen(
+              core.opAsync("op_ws_close", {
+                rid: this[_rid],
+                code,
+                reason,
+              }),
+              () => {
+                this[_readyState] = CLOSED;
+
+                const errEvent = new ErrorEvent("error");
```
lucacasonato, line 478: Do error events not have error messages?

andreubotella: I brought this up months ago and then completely forgot about it, but the `error` event thrown by the `WebSocket` API is an instance of `Event`, not of `ErrorEvent`. Step 2 in ["when the WebSocket connection is closed"](https://websockets.spec.whatwg.org/#closeWebSocket) throws an event named `error`, not an `ErrorEvent`.

andreubotella: But I guess this doesn't have to block this PR.

crowlKats: I'd find that kinda counter-productive since `Event` doesnt allow having a message, unlike ErrorEvent. i'd rather break spec for this than have no clue why a websocket is erroring

ext/http/01_http.js
```diff
@@ -378,6 +387,9 @@
     setEventTargetData(socket);
     socket[_server] = true;
     response[_ws] = socket;
+    if (options.idleTimeout !== 0) {
+      socket[_idleTimeoutDuration] = options.idleTimeout ?? 120;
+    }
```
bnoordhuis, line 392:
```suggestion
    socket[_idleTimeoutDuration] = options.idleTimeout ?? 120;
```
Conditionally setting properties causes property lookups to go polymorphic or megamorphic. Bad for performance.

Set `socket[_idleTimeoutTimeout]` to _something_ (e.g. undefined) for the same reason.

bartlomieju (approved):
LGTM, nice work @crowlKats

lucacasonato (approved):

Issue #11362, opened 2021-07-12
lsp: unable to import types in JSDoc


jaydenseric:
In regular TypeScript projects, within a `.js` file with type checking enabled via `// @ts-check` you can import types from other files within JSDoc comments. For example…

In `a.js`:

```js
// @ts-check

/**
 * Description for A.
 * @param {import("./b.js").default} b The B.
 */
export default function a(b) {
  return b;
}
```

In `b.js`:

```js
// @ts-check

/**
 * Description for B.
 * @type {boolean}
 */
export default true;
```

Note that the parameter type is able to be imported and checked:

<img width="453" alt="Screen Shot 2021-07-12 at 10 43 26 am" src="https://user-images.githubusercontent.com/1754873/125215375-b1f31300-e2fe-11eb-868d-579603538688.png">

With exactly the same setup in a Deno project, the parameter type cannot be imported or checked:

<img width="447" alt="Screen Shot 2021-07-12 at 10 43 07 am" src="https://user-images.githubusercontent.com/1754873/125215443-db13a380-e2fe-11eb-9705-318b642ce12b.png">

This not working in Deno, with no open issues about it has really made the past few days trying to get an isomorphic codebase with a mixture of `.ts`, `.tsx`, and `.js` files a nightmare, because it's hard to tell if it's you or Deno that's doing the wrong thing. I _think_ if this Deno bug were fixed I'd finally have a workable solution!

jaydenseric:
Has any investigation on this happened yet; is this an easy or hard bug to fix?

If it's a low-hanging fruit, patching this would be a game changer for those of us trying to develop and publish isomorphic library code with type safety. Type checking is practically impossible for universal browser/Deno modules until this is fixed.

jaydenseric:
Here is an example of a type importable in a `.ts` file:

```ts
import type { Context } from "https://deno.land/x/oak@v9.0.1/mod.ts";
```

<img width="642" alt="Screen Shot 2021-10-04 at 6 27 13 pm" src="https://user-images.githubusercontent.com/1754873/135811110-c0189488-0208-465f-a75c-53678da42d8b.png">

As you can see, attempting to import it for use in a JSDoc `@typedef` doesn't work:

```js
// @ts-check
/** @typedef {import("https://deno.land/x/oak@v9.0.1/mod.ts").Context} Context */
```

[...]

PR #13173, opened 2021-12-22
chore: update deno_graph and deno_doc

kitsonk:
@bartlomieju two test cases are failing, related to compat support and have no idea how the updates broke them:

- integration::compat::import_esm_from_cjs
- integration::compat::require_in_repl


bartlomieju:
Running these test locally...

bartlomieju:
@kitsonk this is caused by https://github.com/denoland/deno_std/blob/4e01d764eb86caef2e15c42585f323610903802c/node/stream/consumers.js#L5-L9 

`deno_graph` started understanding these comments and it seems they are broken links. I will fix it in `deno_std`

kitsonk:
Oh yeah, they were just always "broken" in the sense it would just be an any type. Now they actually work and follow standard Deno resolution (unless there is a resolver).

bartlomieju:
That's good, another small bug caught :) opened https://github.com/denoland/deno_std/pull/1740 to fix it

kitsonk:
Thinking about this though, I think I need to change `deno_graph`, dependencies like this shouldn't be a hard fail for the graph, our `tsc` integration handles these missing fine and we should prevent code from running if a type only dependency is missing. Doing so would be a big change in behaviour, especially if you tried to use code that has just JSDoc type imports that can't be resolved.

kitsonk:
Hmmm... I have gone back to `deno_graph` and it is actually behaving like it is expected, in that `graph.valid()` is "ok" but `graph.valid_types_only()` errors, as would be expected, which is due to the whole `GraphData` thing.  So I need to ferret out how it is going wrong, and only raises more concerns I have about the whole `GraphData` abstraction. 😖

nayeemrmn:
> dependencies like this shouldn't be a hard fail for the graph, our `tsc` integration handles these missing fine and we should prevent code from running if a type only dependency is missing

We used to call `graph.valid()` which checks runtime deps only, and additionally call `graph.valid_types_only()` only if type checking is enabled. With the current `graph_data.check()`, you can toggle that with [`follow_type_only`](https://github.com/denoland/deno/blob/ac06797fa8607f2e15477fe1b038215740a5747d/cli/graph_util.rs#L250). It is similarly set based on `--no-check` status. The abstraction is better. The same problem would have been hit with the old API, it's just now we have newly acknowledged type only deps from JSDoc, and checking those (since type checking is on) isn't working out with some existing code ¯\\\_(ツ)\_/¯

nayeemrmn:
Also based on #11362, it sounds like we only want to visit JSDoc dependencies for a JS module when either `--checkJs` is enabled or `// @ts-check` is written at the top? If that's what tsc does we should do the same.

kitsonk:
@nayeemrmn yes, basically I am folding that into the logic, but that ended up not being the root cause... the loader appears to actually not be handling missing modules "properly", which is causing a problem.

kitsonk:
Ok, yeah, it looks like the loader wasn't treating 404's like it was treating files missing locally, causing a difference in behaviour.  That is corrected now, and now I am at the point of figuring out the graph valid stuff.

kitsonk:
I have confirmed this patch fixes #13156.

kitsonk:
M.O.A.

bartlomieju (approved):
LGTM

PR #13175, opened 2021-12-22
fix: map basic css color keywords to ansi

### LINKED ISSUE #13165: console.log css styling: colors not working in default apple terminal
Colors are not rendered properly using console.log styling on the default terminal on MacOS.  Lighter colors are rendered as gray, and darker colors are rendered as white.

What's strange is that it only affects console log statements with css styling, so I was hoping it is something that can be fixed in Deno itself, logging objects are colored properly in the same process output.

Also, tmux works fine.

See repro and output below

```
$ deno --version
deno 1.17.0 (release, aarch64-apple-darwin)
v8 9.7.106.15
typescript 4.5.2
```

colors_repro.ts
```
console.log({ foo: "green" });
console.log("%cgreen", "color: green");
console.log("%cpink", "color: pink");
console.log(Object.entries(Deno.env.toObject()).filter(v => v[0].startsWith('TERM')));
```

In default apple terminal:
output-default

In tmux:
output-tmux

zachauten:
closes #13165

---
This will colorize the 8 basic colors in css that are also ansi colors (black, white, red, green, blue, yellow, magenta, cyan). It works regardless of the method used to specify the color, e.g. #000000 and "black" will both make the text black. If you want only the color's name to colorize output, i'll update the PR.

I couldn't find the unit tests for this part of the code, but if someone could point me in the right direction i'd be glad to add coverage for this, and I wasn't able to run the lint & format scripts locally for some reason. Are they broken?

### DIFF
--- a/ext/console/02_console.js
+++ b/ext/console/02_console.js
@@ -1547,7 +1547,25 @@
     if (!colorEquals(css.backgroundColor, prevCss.backgroundColor)) {
       if (css.backgroundColor != null) {
         const [r, g, b] = css.backgroundColor;
-        ansi += `\x1b[48;2;${r};${g};${b}m`;
+        if (colorEquals(css.backgroundColor, [0, 0, 0])) {
+          ansi += `\x1b[40m`;
+        } else if (colorEquals(css.backgroundColor, [255, 0, 0])) {
+          ansi += `\x1b[41m`;
+        } else if (colorEquals(css.backgroundColor, [0, 128, 0])) {
+          ansi += `\x1b[42m`;
+        } else if (colorEquals(css.backgroundColor, [255, 255, 0])) {
+          ansi += `\x1b[43m`;
+        } else if (colorEquals(css.backgroundColor, [0, 0, 255])) {
+          ansi += `\x1b[44m`;
+        } else if (colorEquals(css.backgroundColor, [255, 0, 255])) {
+          ansi += `\x1b[45m`;
+        } else if (colorEquals(css.backgroundColor, [0, 255, 255])) {
+          ansi += `\x1b[46m`;
+        } else if (colorEquals(css.backgroundColor, [255, 255, 255])) {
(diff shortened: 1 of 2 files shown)


kt3k:
Tests are written in `cli/tests/unit/console_test.ts`. Can you update the test cases?

This approach might work, but this slightly change the color of these 8 cases when the terminal supports true colors. Is this acceptable, or should we detect terminal's color mode? cc @lucacasonato @nayeemrmn 

(If we do detect the terminal's color mode, this file might help https://github.com/chalk/supports-color/blob/main/index.js, it looks very complex though.. )

ext/console/02_console.js
```diff
@@ -1547,15 +1547,51 @@
     if (!colorEquals(css.backgroundColor, prevCss.backgroundColor)) {
       if (css.backgroundColor != null) {
         const [r, g, b] = css.backgroundColor;
-        ansi += `\x1b[48;2;${r};${g};${b}m`;
+        if (colorEquals(css.backgroundColor, [0, 0, 0])) {
```
nayeemrmn, line 1550: I don't agree with mapping the matching hex values to the 8 terminal colours, what we should do is move the `parseColor()` step over here and short-circuit at the colour keywords before calling it.

nayeemrmn:
> this slightly change the color of these 8 cases when the terminal supports true colors. Is this acceptable, or should we detect terminal's color mode?

I think it's fine if our handling is relatively one-to-one with ANSI syntax. It's going to be different across terminals but have consistency with other tools e.g. trying to match to green from our `Check <specifier>` messages.

ext/console/02_console.js
zachauten: Done!

nayeemrmn (approved):
LGTM, nice

kt3k (approved):
LGTM

lucacasonato (approved):
LGTM, thanks.

PR #13177, opened 2021-12-22
feat(ext/crypto) - support encrypt/decrypt with AES-CTR

SeanWykes:
Towards #11690. Adds support for `AES-CTR` algorithm in `encrypt()`and `decrypt()`.

Values of counter-length restricted to 32/64/128 as per common usage described in issue #13201.


cli/tests/unit/webcrypto_test.ts
```diff
@@ -608,6 +608,109 @@ Deno.test(async function testAesCbcEncryptDecrypt() {
...
+        name: "AES-CTR",
+        counter,
+        length,
+      },
+      key,
+      plainText,
+    );
+
+    assert(cipherText instanceof ArrayBuffer);
+    assertEquals(cipherText.byteLength, plainText.byteLength);
```
bnoordhuis, line 629: Excess of caution: consider adding a check that verifies cipherText != plainText, otherwise you'd never know it if encrypt and decrypt are identity functions. Likewise around line 642.

cli/tests/unit/webcrypto_test.ts
```diff
@@ -608,6 +608,109 @@ Deno.test(async function testAesCbcEncryptDecrypt() {
...
+        // higher bits
+        assertEquals(
+          new Uint8Array(ciphertext1),
+          new Uint8Array(ciphertext2).slice(16),
+        );
+      }
+    }
+  } catch (E) {
+    console.log("Error: ", E);
+  }
```
bnoordhuis, line 711: This try/catch block shouldn't be there, right?

ext/crypto/encrypt.rs
```diff
@@ -2,7 +2,20 @@ use std::cell::RefCell;
 use std::rc::Rc;
 
 use crate::shared::*;
+
+use aes::cipher::NewCipher;
+use aes::BlockEncrypt;
+use aes::NewBlockCipher;
+//use aes::cipher::NewCipher;
```
bnoordhuis, line 9:
```suggestion
```

ext/crypto/encrypt.rs
```diff
@@ -136,3 +161,54 @@ fn encrypt_aes_cbc(
...
+
+fn encrypt_aes_ctr_gen<B, F>(key: &[u8], counter: &[u8], data: &[u8]) -> Vec<u8>
+where
+  B: BlockEncrypt + NewBlockCipher,
+  F: CtrFlavor<B::BlockSize>,
+{
+  let mut cipher = Ctr::<B, F>::new(key.into(), counter.into());
+
+  let mut ciphertext = data.to_vec();
+  cipher.apply_keystream(&mut ciphertext);
```
bnoordhuis, line 173: Just checking to see if my understanding is correct: it's okay to use `apply_keystream()` instead of `try_apply_keystream()` because the ciphertext is always a multiple of the block size?

SeanWykes: As a stream cipher, any length of text is allowed, not being restricted to multiples of the block-size. There is a case where `apply_keystream` will throw, however, which is when the user tries to encrypt more than `2 ^ length x 16` bytes of data, which in our current worse case, is 128GB in a single call.

Altered to use `try_apply_keystream` as excess of caution.

cli/tests/unit/webcrypto_test.ts
SeanWykes: sure thing! ...gone

cli/tests/unit/webcrypto_test.ts
SeanWykes: done @ 628. `cryptoText != plainText`. 
Not required @642 since already checking if `decryptedText == plainText`, and `cryptoText != plainText`

bnoordhuis (approved):
Thanks for the PR!

PR #13181, opened 2021-12-22
v1.17.1

bartlomieju:


dsherret (approved):
LGTM. I have a feeling the cargo publish will fail because of a breaking change in a patch release in one of the swc crates (edit: actually, looks like the CLI doesn't use that swc crate so we should be good).

Issue #13021, opened 2021-12-07
Validate and clean-up source maps when bundling


kitsonk:
We need to validate source maps when bundling and see what scenarios they work in and are broken.

This is a tracking issue for some work.

lucacasonato:
`deno bundle` has been deprecated and removed in Deno 2.0

PR #13188, opened 2021-12-23
fix(cli): include JSON modules in bundle

### LINKED ISSUE #13150: `deno bundle` not working with JSON module
I'm testing out the new JSON modules functionality introduced with `1.17.0`.  I'm expecting the `deno bundle` command to be able to handle the JSON module import, but I'm getting an error.  Is this a bug, or is bundling JSON modules not supported?

**./test.ts**
```typescript
import j from './test.json' assert { type: 'json' }
console.log(JSON.stringify(j))
```

**./test.json**

```json
{"foo":"bar"}
```

Console session:

```shell
$ deno --version
deno 1.17.0 (release, x86_64-pc-windows-msvc)
v8 9.7.106.15
typescript 4.5.2

$ deno run test.ts
{"foo":"bar"}

$ deno bundle test.ts
Check file:///C:/.../test.ts
Bundle file:///C:/.../test.ts
error: Unable to output during bundling.

Caused by:
    0: load_transformed failed
    1: Bundler.load() failed
    2: Bundler.loader.load(file:///C:/.../test.json) failed    
    3: Expected ';', '}' or <eof> at file:///C:/.../test.json:1:7
```

kitsonk:
Fixes #13150

### DIFF
--- a/cli/ast/mod.rs
+++ b/cli/ast/mod.rs
@@ -350,12 +350,23 @@ pub fn transpile_module(
   options: &EmitOptions,
   cm: Rc<SourceMap>,
 ) -> Result<(Rc<deno_ast::swc::common::SourceFile>, Module), AnyError> {
-  let source = strip_bom(source);
+  let source = if media_type == MediaType::Json {
+    format!(
+      "export default JSON.parse(`{}`);",
+      source.replace("${", "\\${")
+    )
+  } else {
+    strip_bom(source).to_string()
+  };
   let source_file =
-    cm.new_source_file(FileName::Url(specifier.clone()), source.to_string());
+    cm.new_source_file(FileName::Url(specifier.clone()), source);
   let input = StringInput::from(&*source_file);
   let comments = SingleThreadedComments::default();
-  let syntax = get_syntax(media_type);
+  let syntax = if media_type == MediaType::Json {
+    get_syntax(MediaType::JavaScript)
+  } else {
+    get_syntax(media_type)
+  };
   let lexer = Lexer::new(syntax, deno_ast::ES_VERSION, input, Some(&comments));
   let mut parser = deno_ast::swc::parser::Parser::new_from(lexer);
   let module = parser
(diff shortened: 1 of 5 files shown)


andreubotella:
Note that this will output incorrect sourcemaps with `Deno.emit()`, since the byte offsets are into the mapped source. I tried to hack around this with a custom implementation of [`Input`](https://docs.rs/swc_common/latest/swc_common/input/trait.Input.html) that treated all positions before the start of the JSON as 0, but I didn't get very far with it.

This is not a blocker for this PR, but it should probably be fixed before the next patch release.

cli/ast/mod.rs
```diff
@@ -350,12 +350,23 @@ pub fn transpile_module(
   options: &EmitOptions,
   cm: Rc<SourceMap>,
 ) -> Result<(Rc<deno_ast::swc::common::SourceFile>, Module), AnyError> {
-  let source = strip_bom(source);
```
bartlomieju, line 353: I believe BOM should be stripped for JSON too, we do that in `deno_core`:https://github.com/denoland/deno/blob/de2df20d574b3a311d80d628cb7cefe20020d5e7/core/modules.rs#L783

bartlomieju (approved):
LGTM, Andreu has c good point regarding source maps.

cli/ast/mod.rs
```diff
@@ -350,12 +350,23 @@ pub fn transpile_module(
...
+  let source = if media_type == MediaType::Json {
+    format!(
+      "export default JSON.parse(`{}`);",
+      source.replace("${", "\\${")
```
dsherret, line 356: Also should escape backtick?

It's funny, in dnt when implementing this feature I forgot the opposite (now fixed).

dsherret (approved):
LGTM once the two issues are fixed.

kitsonk:
> Note that this will output incorrect sourcemaps with `Deno.emit()`, since the byte offsets are into the mapped source. I tried to hack around this with a custom implementation of [`Input`](https://docs.rs/swc_common/latest/swc_common/input/trait.Input.html) that treated all positions before the start of the JSON as 0, but I didn't get very far with it.
> 
> This is not a blocker for this PR, but it should probably be fixed before the next patch release.

Yeah, I realised it when I did it that it would "mess" things up, but I plan to mop it up, along with other things as part of #13021.

cli/ast/mod.rs
kitsonk: Yeah, I was thinking it was the "strip hashbang" logic for some reason, not the strip BOM stuff... 🤦

PR #13190, opened 2021-12-23
fix(coverage): use only string byte indexes and 0-indexed line numbers

### LINKED ISSUE #11327: A final blankline in a file prevents 100% code coverage
A final newline in a file incorrectly prevents 100% code coverage when using the `deno test` and `deno coverage` CLIs.

This is especially a problem because many projects enforce a final newline in files, via [EditorConfig](https://editorconfig.org) (`insert_final_newline` config), linters, or other means.

To reproduce, in `example.js` (note the newline at the end):

```js
export default function example() {
  return true;
}

```

Then, in `example_test.js`:

```js
import example from "./example.js";

Deno.test("Example.", () => {
  example();
});
```

Run:

```sh
deno test --coverage=cov && deno coverage cov
```

And note that the final line 4 is reported as uncovered, resulting in 75% code coverage for the file `example.js`:

```
running 1 test from file:///[redacted]/demo/example_test.js
test Example. ... ok (8ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (45ms)

cover file:///[redacted]/demo/example.js ... 75.000% (3/4)
   4 |
```

If you delete that final newline in `example.js`, the report changes to 100% code coverage:

```
running 1 test from file:///[redacted]/demo/example_test.js
test Example. ... ok (8ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (45ms)

cover file:///[redacted]/demo/example.js ... 100.000% (3/3)
```

`deno --version`:

```
deno 1.11.5 (release, x86_64-apple-darwin)
v8 9.1.269.35
typescript 4.3.2
```

dsherret:
The coverage code was mixing character indexes and byte indexes along with 0-indexed line numbers with 1-indexed. This PR fixes that along with ignoring blank lines. There is still some work to do on coverage.

Closes #11327


cli/tools/coverage.rs
```diff
@@ -25,15 +25,16 @@ use std::fs::File;
 use std::io::BufWriter;
 use std::io::Write;
 use std::path::PathBuf;
-use std::sync::Arc;
+use text_lines::TextLines;
 use uuid::Uuid;
 
-// TODO(caspervonb) all of these structs can and should be made private, possibly moved to
-// inspector::protocol.
+// TODO(caspervonb) all of these structs should possibly moved to inspector::protocol.
```
bartlomieju, line 31: Remove this comment, no longer relevant

cli/Cargo.toml
```diff
@@ -81,6 +81,7 @@ shell-escape = "=0.1.5"
 sourcemap = "=6.0.1"
 tempfile = "=3.2.0"
 text-size = "=1.1.0"
+text_lines = "=0.4.1"
```
dsherret, line 84: This is my repo, which is used in deno_ast and already a dependency.

cli/tools/coverage.rs
```diff
@@ -171,392 +172,338 @@ impl CoverageCollector {
...
-
-        branches_found += 1;
-        if range.count > 0 {
-          branches_hit += 1;
+  // TODO(caspervonb): collect uncovered ranges on the lines so that we can highlight specific
+  // parts of a line in color (word diff style) instead of the entire line.
+  let mut line_counts = Vec::with_capacity(text_lines.lines_count());
+  for line_index in 0..text_lines.lines_count() {
+    let line_start_offset = text_lines.line_start(line_index);
+    let line_end_offset = text_lines.line_end(line_index);
```
dsherret, line 299: It might be good to improve `text_lines` to return an iterator of line information, but this is an `O(1)` lookup so not critical... just would be nicer.

cli/tools/coverage.rs
```diff
@@ -171,392 +172,338 @@ impl CoverageCollector {
...
+  let comment_spans = deno_ast::lex(script_source, MediaType::JavaScript)
+    .into_iter()
+    .filter(|item| {
+      matches!(item.inner, deno_ast::TokenOrComment::Comment { .. })
+    })
+    .map(|item| item.span)
+    .collect::<Vec<_>>();
+
+  let url = Url::parse(&script_coverage.url).unwrap();
+  let file_path = url.to_file_path().unwrap();
```
bartlomieju, line 216: Is it guaranteed that these unwraps won't file? Why not make `script_coverage.url` a `ModuleSpecifier`?

cli/tools/coverage.rs
```diff
@@ -171,392 +172,338 @@ impl CoverageCollector {
...
+  let mut line_counts = Vec::with_capacity(text_lines.lines_count());
+  for line_index in 0..text_lines.lines_count() {
+    let line_start_offset = text_lines.line_start(line_index);
+    let line_end_offset = text_lines.line_end(line_index);
+    let ignore = comment_spans.iter().any(|span| {
+      (span.lo.0 as usize) <= line_start_offset
+        && (span.hi.0 as usize) >= line_end_offset
+    }) || script_source[line_start_offset..line_end_offset]
+      .trim()
+      .is_empty();
```
dsherret, line 305: This fixes the blank line issue. I think we need to do something a bit better for determining which lines to ignore in the future though (ex. lines with only a close brace or similar token).

cli/tools/coverage.rs
```diff
@@ -171,392 +172,338 @@ impl CoverageCollector {
...
-          block_number,
-          branch_number,
-          taken
-        );
-
-        branches_found += 1;
-        if range.count > 0 {
-          branches_hit += 1;
+  // TODO(caspervonb): collect uncovered ranges on the lines so that we can highlight specific
+  // parts of a line in color (word diff style) instead of the entire line.
```
bartlomieju, line 295: Is this comment still relevant?

cli/tools/coverage.rs
```diff
@@ -171,392 +172,338 @@ impl CoverageCollector {
...
-            }
 
-            let overlaps = std::cmp::max(line_end_offset, &range.end_offset)
-              - std::cmp::min(line_start_offset, &range.start_offset)
-              < (line_end_offset - line_start_offset)
-                + (range.end_offset - range.start_offset);
+          let overlaps = std::cmp::max(line_end_offset, range.end_offset)
+            - std::cmp::min(line_start_offset, range.start_offset)
+            < (line_end_offset - line_start_offset)
+              + (range.end_offset - range.start_offset);
```
bartlomieju, line 333: It's really hard to read this :D maybe split into two separate variable bindings?

cli/tools/coverage.rs
```diff
@@ -171,392 +172,338 @@ impl CoverageCollector {
...
-      .map(|source_map| SourceMap::from_slice(&source_map).unwrap());
-
-    let mut ignored_spans: Vec<Span> = Vec::new();
-    for item in deno_ast::lex(script_source, MediaType::JavaScript) {
-      if let deno_ast::TokenOrComment::Token(_) = item.inner {
-        continue;
-      }
+impl CoverageReporter for LcovCoverageReporter {
+  fn report(&mut self, coverage_report: &CoverageReport, _file_text: &str) {
+    println!("SF:{}", coverage_report.file_path);
```
bartlomieju, line 415: This is a very nice cleanup

bartlomieju (approved):
LGTM, massive effort David; I got a few nitpicks and I will check this PR locally against `deno_std` to make sure nothing breaks.

bartlomieju:
>  I will check this PR locally against deno_std to make sure nothing breaks.

Verified, that everything works correctly.

cli/tools/coverage.rs
dsherret: I just kept the code as-is here for the most part. I don't want to change `script_coverage.url` to a `ModuleSpecifier` for this PR (maybe a followup)... looks like it might be a filename. I will slightly improve this though.

cli/tools/coverage.rs
dsherret: Yes.

cli/tools/coverage.rs
dsherret: This is code from before. I'm not sure why it's using lengths to do this comparison. I changed it to compare offsets and it should be the same (I think).

Issue #13006, opened 2021-12-07
Event loop hangs unless "Runtime.runIfWaitingForDebugger" message is sent to inspector


bartlomieju:
Based on https://gist.github.com/evanwashere/04cfa233f613343e1c9d6fb72cfd628c from @evanwashere 

```js
class Emitter {
  #on = Object.create(null);
  #once = Object.create(null);

  on(n, fn) { (this.#on[n] ?? (this.#on[n] = [])).push(fn); }
  once(n, fn) { (this.#once[n] ?? (this.#once[n] = [])).push(fn); }
  emit(n, ...args) { this.#on[n]?.forEach(_ => _(...args)); if (this.#once[n]) this.#once[n] = (this.#once[n].forEach(_ => _(...args)), undefined); }
  off(n, fn) { if (!fn) (delete this.#on[n], delete this.#once[n]); else { let o = this.#on[n]?.indexOf(fn) ?? -1; if (o !== -1) this.#on[n].splice(o, 1); o = this.#once[n]?.indexOf(fn) ?? -1; if (o !== -1) this.#once[n].splice(o, 1); } }
}

class Inspector extends Emitter {
  #seq = 0;
  #socket = null;
  #promises = Object.create(null);

  async call(method, params) {
    if (!this.#socket) throw new Error('not connected');

    return await new Promise((ok, err) => {
      this.#promises[this.#seq] = { ok, err };
      this.#socket.send(JSON.stringify({ method, params, id: this.#seq++ }));
    });
  }

  async connect(url) {
    return await new Promise((ok, err) => {
      this.#socket = new WebSocket(url);

      this.#socket.onopen = () => ok();
      this.#socket.onerror = _ => err(_.error);

      this.#socket.onclose = () => {
        this.#seq = 0;
        this.#socket = null;
        const e = new Error('connection closed');
        for (const _ in this.#promises) this.#promises[_].err(e);

        this.#promises = Object.create(null);
      };

      this.#socket.onmessage = _ => {
        const { id, error, method, params, result } = JSON.parse(_.data);
        if (method === 'Runtime.executionContextDestroyed') this.#socket.close();

        else if (method) this.emit(method, params);

        else {
          if (!error) this.#promises[id].ok(result);

          else {
            const e = new Error(error.message);
            this.#promises[id].err((e.code = error.code, e));
          }

          delete this.#promises[id];
        }
      };
    });
  }
}

const inspectorUrl = Deno.args[0];

[...]

PR #13191, opened 2021-12-23
fix(core): inspector works if no "Runtime.runIfWaitingForDebugger" message is sent

bartlomieju:
Closes https://github.com/denoland/deno/issues/13006


cli/tests/integration/inspector_tests.rs
```diff
@@ -410,22 +410,20 @@ async fn inspector_runtime_evaluate_does_not_crash() {
     ),
     WsRecv(r#"{"id":1,"result":{}}"#),
     WsRecv(r#"{"id":2,"result":{"debuggerId":"#),
-    WsSend(r#"{"id":3,"method":"Runtime.runIfWaitingForDebugger"}"#),
-    WsRecv(r#"{"id":3,"result":{}}"#),
```
bartlomieju, line 414: These two messages shouldn't be needed for this test to work (because we don't run with `--inspect-brk` flag, but with `--inspect` flag). However due to "handshaking" done in https://github.com/denoland/deno/blob/5027826a6407c96a973155be0620b458ab74d974/core/inspector.rs#L122-L125 it is currently required. This PR will attempt to fix that.

bartlomieju:
Welp, solves one issue but causes tests to become racy 😬

bartlomieju:
I think this change is desirable; however due to the fact that we no longer explicitly "handshake" between isolate's inspector and frontend the order of messages is not guaranteed. I searched through Node codebase and most of the time `Debugger.paused` messages are handled with `waitForNotification` function. I think maybe we just need to rework our tests and roll with this change? @piscisaureus @bnoordhuis what do you think?

core/inspector.rs
```diff
@@ -222,16 +221,12 @@ impl JsRuntimeInspector {
         // Do one "handshake" with a newly connected session at a time.
         if let Some(session) = &mut sessions.handshake {
           let poll_result = session.poll_unpin(cx);
-          let handshake_done =
-            replace(&mut self.flags.borrow_mut().session_handshake_done, false);
           match poll_result {
-            Poll::Pending if handshake_done => {
+            Poll::Pending => {
               let session = sessions.handshake.take().unwrap();
               sessions.established.push(session);
```
bnoordhuis, line 227: This doesn't look correct to me. `session.poll_unpin(cx)` can return Pending multiple times.

bartlomieju: That's expected isn't it? Every future can return `Pending` multiple times. We're more interested when `Ready` is returned signaling that the session is done. As a side note, I would prefer to have explicit method like `pump_messages()` (which I actually refactored in https://github.com/denoland/deno/pull/12962) but that doesn't play nicely with `FuturesUnordered` which is used to store all sessions.

bartlomieju:
Looks like leaving "handshake" step, which is a single poll of the `InspectorSession`, makes the test pass. It is still unclear to me why that is the case, but I've been running `cargo test --release inspector` in a loop for the past hour and haven't hit a single failure.

@bnoordhuis @piscisaureus PTAL. I will re-run CI some more before merging.

core/inspector.rs
```diff
@@ -220,30 +219,22 @@ impl JsRuntimeInspector {
...
-            Poll::Pending if handshake_done => {
-              let session = sessions.handshake.take().unwrap();
+            Poll::Pending => {
               sessions.established.push(session);
-              take(&mut self.flags.borrow_mut().waiting_for_session);
+              continue;
             }
-            Poll::Ready(_) => sessions.handshake = None,
-            Poll::Pending => break,
+            Poll::Ready(_) => {}
```
bnoordhuis, line 229: Just a suggestion but I think this block would be even easier to understand (and is functionally 100% equivalent) when written like this:
```rust
if let Some(mut session) = sessions.handshake.take() {
  if session.poll_unpin(cx).is_pending() {
    sessions.established.push(session);
  }
}
```
(GH won't let me post it as a diff, unfortunately.)

bnoordhuis (approved):
👍

piscisaureus (approved):

PR #13209, opened 2021-12-27
feat: output cause on JS runtime errors

### LINKED ISSUE #12308: Output cause on JS runtime errors

crowlKats:
Closes #12308

### DIFF
--- a/cli/fmt_errors.rs
+++ b/cli/fmt_errors.rs
@@ -128,6 +128,7 @@ fn format_frame(frame: &JsStackFrame) -> String {
   result
 }
 
+#[allow(clippy::too_many_arguments)]
 fn format_stack(
   is_error: bool,
   message_line: &str,
@@ -271,7 +272,11 @@ impl fmt::Display for PrettyJsError {
       )];
     }
 
-    let cause = self.0.cause.clone().map(|cause| format!("{}", PrettyJsError(cause)));
+    let cause = self
+      .0
+      .cause
+      .clone()
+      .map(|cause| format!("{}", PrettyJsError(cause)));
 
     write!(
       f,
--- a/cli/fmt_errors.rs
+++ b/cli/fmt_errors.rs
@@ -131,6 +131,7 @@ fn format_frame(frame: &JsStackFrame) -> String {
 fn format_stack(
   is_error: bool,
   message_line: &str,
+  cause: Option<&str>,
   source_line: Option<&str>,
   start_column: Option<i64>,
   end_column: Option<i64>,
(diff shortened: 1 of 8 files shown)


core/error.rs
```diff
@@ -92,6 +92,7 @@ pub fn get_custom_error_class(error: &Error) -> Option<&'static str> {
 #[derive(Debug, PartialEq, Clone)]
 pub struct JsError {
   pub message: String,
+  pub cause: Box<Option<JsError>>,
```
bartlomieju, line 95: Maybe `Option<Box<JsError>>`?

cli/fmt_errors.rs
```diff
@@ -128,9 +128,11 @@ fn format_frame(frame: &JsStackFrame) -> String {
   result
 }
 
+#[allow(clippy::too_many_arguments)]
```
bartlomieju, line 131: 😬 please add a TODO to refactor this function

bartlomieju:
This looks mostly good to me. @nayeemrmn can you take a look?

core/error.rs
```diff
@@ -180,53 +181,69 @@ impl JsError {
...
+        } else if !message_prop.is_empty() {
+          format!("Uncaught {}", message_prop)
+        } else {
+          "Uncaught".to_string()
+        };
+        let cause = cause.and_then(|cause| {
+          if cause.is_undefined() {
+            None
+          } else {
+            Some(Box::new(JsError::from_v8_exception(scope, cause)))
```
nayeemrmn, line 207: Potential infinite for `error.cause = error`, may need a `seen: HashSet<_>` arg on this method. V8 inspector nicely cuts off cause printing where the loop is detected, as seen in `deno repl`: 
image
Firefox freaks out.

cli/fmt_errors.rs
crowlKats: eh, the function is called in a single place, and to resolve this we'd need to add a new struct, dont see too much value for that, imo we can leave as-is

bartlomieju (approved):
LGTM, nice work

exe-dealer:
why this PR not included in 1.17.2 ? @crowlKats

kitsonk:
New features are generally not included in patch release.  It will be included in 1.18.

PR #13217, opened 2021-12-29
feat(core): Add a feature to control whether `include_js_files!` reads files at runtime

### LINKED ISSUE #11030: `deno_core` prints cargo related stuff if used as a dependency

andreubotella:
Just a prototype. Would close #11030.


littledivy (commented):
@andreubotella This seems reasonable. Are there are any blockers? Please rebase.

andreubotella:
This is now ready for review. I'll squash and write a commit message with details tomorrow.

andreubotella:
The features with which a test is compiled with seem to depend on the directory from which `cargo test` is being run. So for the tests in `ext/net/ops`, running `cargo test` from `ext/net` will compile them without `deno_core/build-time-includes` (because `deno_net` doesn't have it by default), but running it from the root directory will have that feature. To prevent this change from making tests fail, I made it so files are included at compile time if `cfg(test)` is set. However, it seems like `cargo test` only compiles the crate containing the test with that option, and not any dependencies, which is causing the test failures.

I'm not sure how to solve this.

andreubotella:
Closing because #14614 has now reverted the original cause of this bug.

Issue #12918, opened 2021-11-28
bug: logging errors from `deno_runtime` fails


DjDeveloperr:
**Description**

When embedding `deno_runtime` in an application, errors seem to break - when it's converted to a string, it fails because of some missing op.

**Reproduction**

Add the line

```js
console.log(new Error("test error"));
```

to `deno_runtime/examples/hello_runtime.js`

and run this in `deno_runtime` dir

```sh
cargo run --example hello_runtime
```

You'll get

```
deno:core/01_core.js:4680: Uncaught TypeError: invalid op id: expected type `v8::data::Integer`, got `v8::data::Value`
Error: Uncaught TypeError: invalid op id: expected type `v8::data::Integer`, got `v8::data::Value`
    at deno:core/01_core.js:149:26
```

**Additional details**

After investigating more I found that these ops: https://github.com/denoland/deno/blob/main/cli/ops/errors.rs - are in the cli but not included in `deno_runtime` crate (though it depends on them).

bartlomieju:
Verified on latest `main` branch, this no longer occurs.

PR #13219, opened 2021-12-29
feat(core): JsRuntime::replace_op

bartlomieju:
Adds a new API to "JsRuntime" that allows to replace (or overwrite) an existing op
with a new handler.

This is kind of "very unsafe, but I know what I'm doing, please leave me alone" API,
but has actual use cases in:
- #12938 - to replace "op_exit" with a handler that terminates execution, to prevent
calling "Deno.exit()" API in test code
- #13122 - to replace "op_print" with a handler that forwards all "console.*" messages
over a socket to a notebook frontend

### DIFF
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -201,7 +201,16 @@ impl OpTable {
     F: Fn(Rc<RefCell<OpState>>, OpPayload) -> Op + 'static,
   {
     let (op_id, prev) = self.0.insert_full(name.to_owned(), Rc::new(op_fn));
-    assert!(prev.is_none());
+    assert!(prev.is_none(), "Op '{}' was already registered", name);
+    op_id
+  }
+
+  pub fn replace_op<F>(&mut self, name: &str, op_fn: F) -> OpId
+  where
+    F: Fn(Rc<RefCell<OpState>>, OpPayload) -> Op + 'static,
+  {
+    let (op_id, prev) = self.0.insert_full(name.to_owned(), Rc::new(op_fn));
+    assert!(prev.is_some(), "Op '{}' wasn't already registered", name);
     op_id
   }
 
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -210,7 +210,7 @@ impl OpTable {
     F: Fn(Rc<RefCell<OpState>>, OpPayload) -> Op + 'static,
   {
     let (op_id, prev) = self.0.insert_full(name.to_owned(), Rc::new(op_fn));
-    assert!(prev.is_some(), "Op '{}' wasn't already registered", name);
+    assert!(prev.is_some(), "Op '{}' was not yet registered", name);
     op_id
   }
 
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -640,7 +640,7 @@ impl JsRuntime {
   /// Replaces an already existing op.
   ///
   /// In most situations this method is not required. However, in some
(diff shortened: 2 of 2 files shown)


satyarohith (approved):
LGTM.

AaronO (changes requested):
I'm not in favour of this.

Op-middleware allows wrapping/replacing/disabling ops with a single primitive, e.g:

```rs
  let my_ext = Extension::builder()
    .middleware(|name, opfn| match name {
      "op_print" => deno_core::void_op_sync(),
      _ => opfn,
    })
    .build();
```

Any reason why this pattern doesn't work for your aforementioned use-cases ?

core/ops.rs
```diff
@@ -201,7 +201,16 @@ impl OpTable {
...
     let (op_id, prev) = self.0.insert_full(name.to_owned(), Rc::new(op_fn));
-    assert!(prev.is_none());
+    assert!(prev.is_none(), "Op '{}' was already registered", name);
+    op_id
+  }
+
+  pub fn replace_op<F>(&mut self, name: &str, op_fn: F) -> OpId
+  where
+    F: Fn(Rc<RefCell<OpState>>, OpPayload) -> Op + 'static,
+  {
```
bnoordhuis, line 211: Side discussion: it might makes sense to someday rewrite these (replace_op, register_op, etc.) to:
```rust
pub fn frob_op(&mut self, name: &str, op_fn: Rc<OpFn>) {
  // ...
}
```
Because right now they turn into a combinatorial explosion of monomorphized calls.

Truthfully, that's probably more relevant for `op_sync()` and `op_async()` because they're called way more often than `register_op()` - approx. 300 vs. 35 calls.

bnoordhuis (approved):

core/ops.rs
AaronO: Mid-term I would be in favour of only allowing ops to be registered declaratively at init via Extensions.

bnoordhuis:
> Any reason why this pattern doesn't work for your aforementioned use-cases ?

The goal of this PR is clearly to replace ops at runtime and without overhead.

AaronO:
> > Any reason why this pattern doesn't work for your aforementioned use-cases ?
> 
> The goal of this PR is clearly to replace ops at runtime and without overhead.

https://github.com/denoland/deno/pull/13122 and https://github.com/denoland/deno/pull/12938 both replace ops at isolate-boot time. Op-middleware handles precisely this use-case and there's no difference in opcall overhead for the replaced ops.

bnoordhuis:
At isolate creation time isn't at runtime but maybe that's sufficient for the two linked use cases.

AaronO:
> At isolate creation time isn't at runtime but maybe that's sufficient for the two linked use cases.

It is sufficient for the aforementioned use-cases and dynamically replacing ops at runtime is an anti-pattern IMO unless we're discussing "plugins" but that's off the table.

So in short unless there's a very clear use-case that middleware doesn't already solve, we should not add this API.

I refactored the CLI code in https://github.com/denoland/deno/pull/13223 so @bartlomieju can easily slot in his `op_exit` override for https://github.com/denoland/deno/pull/12938

tim-palmer:
Possible additional use-case: https://github.com/denoland/deno/issues/12918

If it's not possible to move op_apply_source_map, op_format_diagnostic and op_format_file_name into deno_runtime due to dependencies on cli then presumably deno_runtime will need alternative implementations that are then replaced in cli?

PR #13220, opened 2021-12-29
fix(cli): Don't strip shebangs from modules

### LINKED ISSUE #13129: It's no longer necessary to strip shebangs from modules
Deno's module fetcher will [strip a shebang](https://github.com/denoland/deno/blob/main/cli/file_fetcher.rs#L233-L244) if a module file starts with one. However, [there is a stage-3 TC39 proposal to support shebangs (or "hashbangs") as part of JS syntax](https://github.com/tc39/proposal-hashbang). In fact, this [was implemented in V8 in 2018](https://chromium-review.googlesource.com/c/v8/v8/+/1409527/) and [has shipped by default since V8 7.4](https://chromium-review.googlesource.com/c/v8/v8/+/1486113/). So this transformation can be removed.

This affects the output of `Deno.emit()` though. Although `emit::bundle` will skip any shebangs in the source files, and so `deno bundle` and `Deno.emit()` with the `bundle` parameter are safe, `emit::to_file_map` (which powers `Deno.emit()` _without_ the `bundle` option) will output shebangs for the files that have them. This might be bad, depending on what kinds of environments the code generated by `Deno.emit()` will be used in. We should consider whether to strip the shebang from the outputs.

Another potential issue is positions in sourcemaps: since currently shebangs are stripped before the module is parsed, any sourcemaps generated by `tsc` or `swc` would be off by one line for modules with shebangs. I don't know if this error is corrected at all, but if so, the correction would have to be undone. Also, if shebangs are stripped from the `Deno.emit()` outputs, those sourcemaps would also need to be corrected to account for it.

andreubotella:
Deno's module loader currently strips a shebang if a module file starts with one. However, this is no longer necessary, since [there is a stage-3 TC39 that adds support for shebangs (or "hashbangs") to the language](https://github.com/tc39/proposal-hashbang), and V8, `tsc` and `swc` all support it.

Furthermore, stripping shebangs causes a correctness bug with JSON modules, since a JSON file with a shebang should not parse as a JSON module, yet it does with this stripping. This change fixes this.

Closes #13129.


andreubotella:
For the record: this change affects the output of `Deno.emit()` in that shebangs will be kept in the output modules if the `bundle` option is not set. This can already happen if a module file has a BOM and a shebang, since the shebang is stripped before the BOM, but this extends it to every case.

bartlomieju:
I think this change is desirable, but it will most likely break cached sources once released. We need to be extra careful not to repeat situation in 1.16 when users had to use `--reload` to get rid if esoteric errors. 

@kitsonk WDYT?

andreubotella:
I'm not too sure how the cache works, but it seems like the worst case scenario would be errors being off by one line, right?

bartlomieju:
> I'm not too sure how the cache works, but it seems like the worst case scenario would be errors being off by one line, right?

Not necessarily, I think in the CDP inspector a file might not load at all if source map is broken.

bartlomieju:
@kitsonk do you have any objections to landing this PR?

bartlomieju (approved):
LGTM, let's land it and keep an eye out on any failures in canary.

PR #13223, opened 2021-12-29
cleanup(cli): use op Extensions

AaronO:
Enabling op-middleware for overrides in lieu of imperative .replace_op() etc...

Impacts #13219,  #12938, #13122


bartlomieju (approved):
LGTM, guess this works too

PR #13224, opened 2021-12-29
cleanup(core): use Extensions to register ops

AaronO:
In examples and tests.

It's slightly more verbose but I think we should encourage the declarative approach in all examples/code and ultimately deprecate the imperative APIs


bartlomieju (approved):
LGTM

PR #13230, opened 2021-12-29
fix(doc): Make private types which show up in the rustdocs public

andreubotella:
`CrossIsolateStore`, `ExtensionBuilder` and `InMemoryChannelResource` are private types which are referred to by other public APIs, and so they show up in the rustdoc, but not as links. This is especially confusing for `ExtensionBuilder`, since there is nothing in the docs that explains how to build an extension.

Exposing these three types doesn't add any new capabilities: `ExtensionBuilder` can be created from `Extension::builder()`, `SharedArrayBufferStore` and `CompiledWasmModuleStore` already enable doing anything that `CrossIsolateStore` can do, and `InMemoryChannelResource` isn't constructable.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -88,6 +88,7 @@ pub use crate::resources::Resource;
 pub use crate::resources::ResourceId;
 pub use crate::resources::ResourceTable;
 pub use crate::runtime::CompiledWasmModuleStore;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
Thanks

Issue #11570, opened 2021-08-03
Bug in inspector when using deno repl --inspect


evanwashere:
- script execution doesn't pause when debugger is invoked
- console responses require 2 event loop ticks after 1st problem

deno: 

https://user-images.githubusercontent.com/70155278/128002904-83dca980-a000-4ea8-8cad-c8440ef82a96.mov

expected behavior (node): 

https://user-images.githubusercontent.com/70155278/128002966-0a685f97-8956-498b-9a5b-042cda757823.mov

caspervonb:
Does this happen with run as-well or is it repl only?

evanwashere:
> Does this happen with run as-well or is it repl only?

only happens in deno repl and devtools console

bartlomieju:
I did attach WireShark and bumped messages exchanged with CDP in both Node and Deno; the only thing I found is that in case of Deno after receiving `Debugger.paused` notification, another `Debugger.resumed` notification is received immediately after. It is still unclear why this is happening and it's unclear why after that all inputs are "delayed by one". I will keep debugging it, but to be honest I've got no ideas at the moment. I wonder if this might be caused by the fact that in case of Deno we always have an inspector session attached (because that's how Deno REPL works under the hood - it creates a "local inspector session"). So effectively when running `deno repl --inspect` and connecting via CDP we got two inspector sessions going on at the same time.

bartlomieju:
I made some progress debugging the problem. In essence the problem is that `run_message_loop_on_pause` immediately finishes (before `quit_message_loop_on_pause` is called) leaving inspector in a "broken" state.

The problem arises from:
https://github.com/denoland/deno/blob/ee7ab81768c593e99774b629c82c1784204a1cb0/core/inspector.rs#L212
which returns `BorrowMutError`
on line 121: 
https://github.com/denoland/deno/blob/ee7ab81768c593e99774b629c82c1784204a1cb0/core/inspector.rs#L118-L122

I feel there should be an assert at the end of this function that checks if `on_pause` is set to `false`.

Now just need to figure out how to fix it 🙃

PR #13234, opened 2021-12-30
[WIP] fix(core): run_message_loop_on_pause runs nested message loop

bartlomieju:
Closes https://github.com/denoland/deno/issues/11570


core/inspector.rs
```diff
@@ -109,13 +109,24 @@ impl v8::inspector::V8InspectorClientImpl for JsRuntimeInspector {
...
+    eprintln!("run message loop on pause");
     assert_eq!(context_group_id, JsRuntimeInspector::CONTEXT_GROUP_ID);
     self.flags.borrow_mut().on_pause = true;
-    let _ = self.poll_sessions(None);
+    let r = self.poll_sessions(None);
+    eprintln!("finished run message loop on pause {:#?}", r);
+    assert!(
+      !self.flags.borrow().on_pause, 
+      "V8InspectorClientImpl::run_message_loop_on_pause returned before quit_message_loop_on_pause was called"
+    );
```
bartlomieju, line 125: This is root of the problem that this PR is meant to fix. This method is called when we've already called `poll_sessions()` (from `JsRuntime::poll_event_loop`) and a call to `poll_sessions()` here returns `BorrowMutError`. This method shouldn't return until `quit_message_loop_on_pause` is called otherwise inspector goes into broken state with all message handling being delayed by at least one message received.

PR #13235, opened 2021-12-30
feat(test): Add support for "deno test --compat"

### LINKED ISSUE #13227: Support compat mode in deno test
As indicated by the title

This would be useful to run other test runners assertions under `deno test` and easing the migration of node codebases into Deno

Would be specially useful for deno_std, where the Node native tests are run as a subprocess instead of being run directly under compat mode (this is really annoying for local debugging, specially with average CPUs)

Soremwar:
Closes #13227

### DIFF
--- a/cli/tests/integration/compat_tests.rs
+++ b/cli/tests/integration/compat_tests.rs
@@ -54,11 +54,11 @@ itest!(import_esm_from_cjs {
 
 // TODO(Soremwar)
 // Enable this test once I've figured why top level throws won't exit the test
-// itest!(test_runner_cjs {
-//   args: "test --compat --unstable -A --quiet compat/test_runner/cjs.js",
-//   exit_code: 1,
-//   output: "compat/test_runner/cjs.out",
-// });
+itest!(test_runner_cjs {
+  args: "test --compat --unstable -A --quiet compat/test_runner/cjs.js",
+  exit_code: 1,
+  output: "compat/test_runner/cjs.out",
+});
 
 itest!(test_runner_esm {
   args: "test --compat --unstable -A --quiet compat/test_runner/esm.mjs",
@@ -68,13 +68,11 @@ itest!(test_runner_esm {
 
 // Top level assertion test mostly just make sure that the test runner finishes correctly on compat mode
 // when there is no tests
-// TODO(Soremwar)
-// Enable this test once I've figured why top level throws won't exit the test
-// itest!(top_level_assertion_cjs {
-//   args: "test --compat --unstable -A --quiet compat/test_runner/top_level_assertion_cjs.js",
-// 	exit_code: 0,
-//   output: "compat/test_runner/top_level_assertion_cjs.out",
-// });
+itest!(top_level_assertion_cjs {
+  args: "test --compat --unstable -A --quiet compat/test_runner/top_level_assertion_cjs.js",
+	exit_code: 0,
+  output: "compat/test_runner/top_level_assertion_cjs.out",
+});
 
 itest!(top_level_assertion_esm {
   args: "test --compat --unstable -A --quiet compat/test_runner/top_level_assertion_esm.mjs",
@@ -82,13 +80,11 @@ itest!(top_level_assertion_esm {
   output: "compat/test_runner/top_level_assertion_esm.out",
 });
 
-// TODO(Soremwar)
-// Enable this test once I've figured why top level throws won't exit the test
-// itest!(top_level_fail_cjs {
-//   args: "test --compat --unstable -A --quiet compat/test_runner/top_level_fail_cjs.js",
-// 	exit_code: 1,
-//   output: "compat/test_runner/top_level_fail_cjs.out",
-// });
+itest!(top_level_fail_cjs {
+  args: "test --compat --unstable -A --quiet compat/test_runner/top_level_fail_cjs.js",
+	exit_code: 1,
+  output: "compat/test_runner/top_level_fail_cjs.out",
+});
 
 itest!(top_level_fail_esm {
   args: "test --compat --unstable -A --quiet compat/test_runner/top_level_fail_esm.mjs",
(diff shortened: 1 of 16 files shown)


cli/tools/test.rs
```diff
@@ -476,12 +478,34 @@ async fn test_specifier(
...
+    } else {
+      // We execute the module module as a side module so that import.meta.main is not set.
+      worker.execute_side_module(&specifier).await?;
+    }
   }
 
   worker.dispatch_load_event(&located_script_name!())?;
+  if await_cjs_modules {
+    worker.run_event_loop(false).await?;
+  }
```
bartlomieju, line 508: Since CJS modules execute synchronously I think this could be moved after `load_cjs_module`

cli/tests/integration/compat_tests.rs
```diff
@@ -52,6 +52,46 @@ itest!(import_esm_from_cjs {
   output_str: Some("function\n"),
 });
 
+// TODO(Soremwar)
+// Enable this test once I've figured why top level throws won't exit the test
```
bartlomieju, line 56: remove

cli/tests/testdata/compat/test_runner/top_level_fail_cjs.out
```diff
@@ -0,0 +1,13 @@
...
+error: Uncaught (in promise) AssertionError: Values are not strictly equal:
+
+
+    [Diff] Actual / Expected
+
+
+-   10
++   20
+
+[WILDCARD]
```
bartlomieju, line 13: Something for another PR, but I found this output very confusing... We should be able to catch these errors before printing test report

cli/tools/test.rs
bartlomieju: And then `await_cjs_modules` could be removed.

bartlomieju (commented):
Looks good!

cli/tools/test.rs
Soremwar: Done

cli/tests/integration/compat_tests.rs
Soremwar: Done

cli/tests/testdata/compat/test_runner/top_level_fail_cjs.out
Soremwar: Confusing how?

bartlomieju: Confusing in that the error is returned after the test report - so at first glance it seems something failed after running all tests, which most often is not true - usually it's an error in top level code (or floating promise) that happens before tests are run. Again, something to address in a separate PR

bartlomieju:
LGTM, I will run some test locally before merging.

bartlomieju:
So this PR indeed allows to run Node tests with `deno test`:
```
export DENO_NODE_COMPAT_URL=file:///Users/biwanczuk/dev/deno_std/
../deno/target/debug/deno test --unstable -A --compat --no-check node/_tools/suites/parallel/test-event-emitter-check-listener-leaks.js
running 0 tests from file:///Users/biwanczuk/dev/deno_std/node/_tools/suites/parallel/test-event-emitter-check-listener-leaks.js

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (2s)

```
Of course there are no reported tests being run, because Node's tests don't use `Deno.test()` API, but if you change something to make tests fail you get:
```
../deno/target/debug/deno test --unstable -A --compat --no-check node/_tools/suites/parallel/test-event-emitter-check-listener-leaks.js

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (1s)

error: Uncaught (in promise) AssertionError: Values are not strictly equal:


    [Diff] Actual / Expected


-   foo
+   bar


    Error.captureStackTrace(this, stackStartFn || stackStartFunction);
          ^
    at new AssertionError (file:///Users/biwanczuk/dev/deno_std/node/assertion_error.ts:530:11)
    at toNode (file:///Users/biwanczuk/dev/deno_std/node/assert.ts:58:15)
    at Function.strictEqual (file:///Users/biwanczuk/dev/deno_std/node/assert.ts:300:3)
    at Object.<anonymous> (file:///Users/biwanczuk/dev/deno_std/node/_tools/suites/parallel/test-event-emitter-check-listener-leaks.js:110:10)
    at Module._compile (file:///Users/biwanczuk/dev/deno_std/node/module.ts:252:36)
    at Object.Module._extensions..js (file:///Users/biwanczuk/dev/deno_std/node/module.ts:1276:10)
    at Module.load (file:///Users/biwanczuk/dev/deno_std/node/module.ts:231:33)
    at Function._load (file:///Users/biwanczuk/dev/deno_std/node/module.ts:540:14)
    at loadCjsModule ([deno:cli/compat/mod.rs:118:30]:3:22)
```
Then again, error is kinda confusing because it's shown after the test report (https://github.com/denoland/deno/pull/13235#discussion_r776552724)

CC @kt3k I guess we could use this functionality instead of spawning a subprocess for each of Node test files.

bartlomieju (approved):
LGTM, thank you @Soremwar, nice work

JJ:
This is supposed to be deployed since version 1.18, but I can't seem to make it work...
Captura de pantalla de 2023-03-15 11-22-43
Is it possible that expected arguments are checked before this new code is called?

bartlomieju:
`--compat` flag was removed in Deno 1.25.2

JJ:
> `--compat` flag was removed in Deno 1.25.2

Thanks for the heads up. So it's on by default, or simply dropped that compat mode?

bartlomieju:
Compat mode was an experiment that we decided not to persue. You can instead use `npm:` specifiers (https://deno.com/blog/v1.28#using-npm) or `package.json` support (https://deno.com/blog/v1.31#packagejson-support)

Issue #13222, opened 2021-12-29
crypto api stopped working in v1.17.x


ralyodio:
```
error: Uncaught (in promise) DataError: invalid key data
const key = await crypto.subtle.importKey(
                                ^
    at Object.opSync (deno:core/01_core.js:149:12)
    at importKeyHMAC (deno:ext/crypto/00_crypto.js:2055:34)
    at SubtleCrypto.importKey (deno:ext/crypto/00_crypto.js:793:18)
```

bnoordhuis:
Can you post a minimal test case?

ralyodio:
```
const keyData = await Deno.readTextFile("jwk.json");

const key = await crypto.subtle.importKey(
  "jwk",
  JSON.parse(keyData),
  { name: "HMAC", hash: "SHA-512" },
  true,
  ["sign", "verify"],
);
```

here is the key file:

```
{
  "kty": "oct",
  "k": "xxx",
  "alg": "HS512",
  "key_ops": ["sign", "verify"],
  "ext": true
}
```

littledivy:
@ralyodio Just to make sure `xxx` is the actual key material and not a placeholder value?

littledivy:
Fixed in https://github.com/denoland/deno/pull/13240

ralyodio:
i tried this on canary and it still is broken:

deno 1.17.1+19c8cd3

littledivy:
@ralyodio I can't reproduce the issue on the same canary version.

![Screenshot from 2022-01-05 11-32-34](https://user-images.githubusercontent.com/34997667/148168235-284053db-3fc3-4201-bacb-2720d6f33d90.png)

Please verify the key material `k` is exacly what you are using

ralyodio:
This is not the full key, but I've removefd some characters. It has +'s in it, is that the issue?

It worked fine in 1.16

```
{
  "kty": "oct",
  "k": "dtPQIio1j0Sf6vmPOSa2sPwA8DlR3jUuMK3nlL44laPn+8EiYTxj9DTu5JdkoEGfmfnhgTFg6TqnT7+y8zV4gQJEsRRIhLwEMZGtRptqZTxRoUYnODn8IkjF9f6SW7oT+ijTTolQPYUxvaEqw",
  "alg": "HS512",
  "key_ops": ["sign", "verify"],
  "ext": true
}
```

littledivy:
[...]

PR #13240, opened 2021-12-30
fix(ext/crypto) base-64 for JWK

SeanWykes:
Ref: #13222

Base-64-url values in JWK are encoded without padding, and have a non-canonical representation since the last character may include 2 or 4 unused "overflow" bits. Encoding should be canonical but decoders may be tolerant, as per `btoa` - [forgiving-base64-decode](https://infra.spec.whatwg.org/#forgiving-base64-decode). 

Chrome and Node implementations of webcrypto `importKey` are tolerant and ignore the overflow bits. 

Previous Deno implementation in js used `btoa` so was tolerant, see [here](https://github.com/denoland/deno/pull/13047/files#diff-863c907dfc6c1c6e507b662a7fc0ead3446473facaf183b7ca1a6908b50ab444L194)

PR implements "forgiving" in JWK decode passing suitable `config` to `base64::decode_config`. Includes test-case.

Rationale: Specifications are somewhat unclear on this, [RFC4648 - 3.5.  Canonical Encoding](https://datatracker.ietf.org/doc/html/rfc4648#section-3.5) states that "decoders MAY choose to reject an encoding if the pad bits have not been set to zero", but JWS/JWK does not mandate or clarify this .. [JSON Web Signature (JWS)](https://datatracker.ietf.org/doc/html/rfc7515#page-54) where example code (C#) just adds missing padding `=` before conversion.

### DIFF
--- a/cli/tests/unit/webcrypto_test.ts
+++ b/cli/tests/unit/webcrypto_test.ts
@@ -1327,11 +1327,10 @@ Deno.test(async function testBase64Forgiving() {
     true,
     ["sign", "verify"],
   );
-  
-  assertEquals(key instanceof CryptoKey);
+
+  assert(key instanceof CryptoKey);
   assertEquals(key.type, "secret");
-  assertEquals(key.algorithm.length, 16);
-  
+
   const exportedKey = await crypto.subtle.exportKey("jwk", key);
   assertEquals(exportedKey.k, "xxw");
 });
--- a/cli/tests/unit/webcrypto_test.ts
+++ b/cli/tests/unit/webcrypto_test.ts
@@ -1310,3 +1310,21 @@ Deno.test(async function testImportEcSpkiPkcs8() {
     assertEquals(new Uint8Array(expPrivateKeySPKI), spki);*/
   }
 });
+
+Deno.test(async function testBase64Forgiving() {
+  const keyData = `{
+    "kty": "oct",
+    "k": "xxx",
+    "alg": "HS512",
+    "key_ops": ["sign", "verify"],
+    "ext": true
+  }`;
+
+  const _key = await crypto.subtle.importKey(
+    "jwk",
+    JSON.parse(keyData),
+    { name: "HMAC", hash: "SHA-512" },
+    true,
+    ["sign", "verify"],
+  );
+});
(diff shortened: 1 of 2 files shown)


cli/tests/unit/webcrypto_test.ts
```diff
@@ -1310,3 +1310,21 @@ Deno.test(async function testImportEcSpkiPkcs8() {
...
+    "ext": true
+  }`;
+
+  const _key = await crypto.subtle.importKey(
+    "jwk",
+    JSON.parse(keyData),
+    { name: "HMAC", hash: "SHA-512" },
+    true,
+    ["sign", "verify"],
+  );
```
littledivy, line 1329:
```suggestion
  const key = await crypto.subtle.importKey(
    "jwk",
    JSON.parse(keyData),
    { name: "HMAC", hash: "SHA-512" },
    true,
    ["sign", "verify"],
  );
  
  assertEquals(key instanceof CryptoKey);
  assertEquals(key.type, "secret");
  assertEquals(key.algorithm.length, 16);
  
  const exportedKey = await crypto.subtle.exportKey("jwk", key);
  assertEquals(exportedKey.k, "xxw");
```

littledivy (approved):
LGTM

SeanWykes:
@bartlomieju, @littledivy  - sorry if this is a silly question but I'm trying to understand why commit `de99d60` references and includes the code here, but the PR was merged in commit `3407641`. In other cases, like #13177, after the PR was merged, github informed me that I could safely delete my branch.

I got confusing prompts from github, and ended up resubmitting to #13285, which I then immediately closed.

Thanks for helping a newbie ..

bartlomieju:
@SeanWykes I'm not 100% sure but I think @littledivy landed that PR with different commit message than PR title.

SeanWykes:
OK. Thanks - just seen that `de99d60` is for 1.17 branch and not main. 

Still curious about why github didn't prompt me to delete the PR, and then on my forked repo prompted me to submit as a new PR. Anyway, learning something new every day.

bartlomieju:
It will actually be present in both branches. We land everything into `main` and then for patch releases we merge back non-feature commits into `v1.1X` branches.

PR #13246, opened 2021-12-31
fix(lsp): properly generate data URLs for completion items

kitsonk:
Discovered an issue with registry completions when generating the documentation URLs where matched variables needed to be interpolated.


bartlomieju (approved):
LGTM

PR #13266, opened 2022-01-02
chore: clap v3

### LINKED ISSUE #8475: Research upgrade to clap v3

### LINKED ISSUE #12118: `NO_COLOR=1` does not works for "Did you mean ..."
It seems `NO_COLOR=1` does not affect coloring on "Did you mean ..." part.

CleanShot 2021-09-17 at 15 52 20

My Deno version is 

```
deno 1.14.0 (release, x86_64-apple-darwin)
v8 9.4.146.15
typescript 4.4.2
```

Thanks for the project 👍

crowlKats:
Closes #8475
Closes #12118


lucacasonato:
Also closes #12118

Cargo.lock
```diff
@@ -71,15 +71,6 @@ dependencies = [
...
 ]
 
-[[package]]
-name = "ansi_term"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
-dependencies = [
- "winapi 0.3.9",
-]
```
ry, line 81: Nice!

ry (approved):
LGTM

commit message should probably be "upgrade: clap v3"

crowlKats:
DO NOT MERGE. discussed with @bartlomieju and we came to the conclusion that this PR should wait for 2.0, as there are some fundamental changes to clap's parsing, which would likely be breaking changes in various cases

ry:
@crowlKats Can you give an example of what changed?

crowlKats:
I have tried to keep behaviour same as current, but i cant be certain that it actually fully is. one test was failing with a weird case and had to change what we were doing, but that change maybe would need to be applied to other subcommands, or maybe not. its a bit hairy. also, cargo is being wary of updating as well https://github.com/rust-lang/cargo/pull/10253#issuecomment-1005045779 (and subsequent comments)

ry:
@crowlKats and I have done a little experimentation and decided that if this does introduce bugs, they are very not common, it works in all the instances we tried. I suggest landing.

LGTM

bartlomieju (approved):

PR #13267, opened 2022-01-03
Improve DX for wpt - added debug + update-debug options

SeanWykes:
There is apparently no easy method to debug WPT tests using VSCode/chrome debugger. One solution is to manually modify `tools/runner.ts` to include `--inspect-brk` (and then remove it when no longer needed .. ).

PR adds two new commands to wpt.ts, thus avoiding the manual modifications.
  `debug` - same as `run` but includes `--inspect-brk` in cmd-line to spawned Deno
  `update-debug` - ditto for `update`

Still not optimal DX when using `chrome://inspect/#devices`, since have to step through the top-level generated wpt-bundle that runs `Deno.core.evalContext` on the inline test-scripts, in order to set breakpoints and thus be able to single-step code.


lucacasonato (commented):
Can this just be an --inspect-brk flag on the `update` and `run` subcommands respectively?

SeanWykes:
> Can this just be an --inspect-brk flag on the `update` and `run` subcommands respectively?
Nice.

Got a fail on CI (test fastci macos-10.15) .. no idea why. 
`    integration::run::permissions::rw_inside_test_and_js_dir
    integration::run::permissions::rw_inside_test_dir`.

Any suggestions? I don't think I can rerun without committing.

bartlomieju:
> Got a fail on CI (test fastci macos-10.15) .. no idea why. ` integration::run::permissions::rw_inside_test_and_js_dir integration::run::permissions::rw_inside_test_dir`.
> 
> Any suggestions? I don't think I can rerun without committing.

Seems like a flake. @SeanWykes you can rerun CI by pushing an empty commit (`git commit --allow-empty -m "reset CI"`), it's a standard practice in case of flakes.

SeanWykes:
> > Got a fail on CI (test fastci macos-10.15) .. no idea why. ` integration::run::permissions::rw_inside_test_and_js_dir integration::run::permissions::rw_inside_test_dir`.
> > Any suggestions? I don't think I can rerun without committing.
> 
> Seems like a flake. @SeanWykes you can rerun CI by pushing an empty commit (`git commit --allow-empty -m "reset CI"`), it's a standard practice in case of flakes.

Thanks. I'm still learning every day!

Looks good now.

bartlomieju (approved):
LGTM, thank you @SeanWykes!

tools/wpt/runner.ts
```diff
@@ -88,20 +89,32 @@ export async function runSingleTest(
...
+    const cmd = [
+      denoBinary(),
+      "run",
+    ];
+
+    cmd.push(
+      "-A",
+      "--unstable",
+    );
+
```
bartlomieju, line 101: Actually, a question here: @lucacasonato mentioned that `--inspect-brk` should be used for "run" and "update" commands, but it seems it's only handled in "run" command here. Is this expected?

SeanWykes: This is Deno's run command, used for executing a test in both update and run modes, both get the `inspect-brk` functionality. 
The `update` is at a higher level, to run and then update the `expectations` file.

bartlomieju: Ah of course, sorry, bad idea to review before the first coffee :P

PR #13270, opened 2022-01-03
refactor(core): InspectorSession implements Stream

bartlomieju:
This commit rewrites "InspectorSession" to no longer implement "Future"
trait but instead implement "Stream" trait. "Stream" trait is implemented 
by yielding a raw pointer to the "v8::inspector::V8InspectorSession" and 
received message. In effect received messages are no longer dispatched 
from within the future, but are explicitly dispatched by the caller.

This change should allow us to dispatch a message to the session when
another message is being dispatched, ie. 
"V8InspectorSesssion::dispatch_protocol_message" is already on the 
call stack. This situation can happen during an execution pause, especially
in situation from #13234

### DIFF
--- a/core/inspector.rs
+++ b/core/inspector.rs
@@ -233,9 +233,7 @@ impl JsRuntimeInspector {
               continue;
             }
             Poll::Ready(Some(session_stream_item)) => {
-              let (v8_session_rc, msg) = session_stream_item;
-              let mut v8_session = v8_session_rc.borrow_mut();
-              let v8_session_ptr = v8_session.as_mut();
+              let (v8_session_ptr, msg) = session_stream_item;
               InspectorSession::dispatch_message(v8_session_ptr, msg);
               sessions.established.push(session);
               continue;
@@ -256,9 +254,7 @@ impl JsRuntimeInspector {
         // Poll established sessions.
         match sessions.established.poll_next_unpin(cx) {
           Poll::Ready(Some(session_stream_item)) => {
-            let (v8_session_rc, msg) = session_stream_item;
-            let mut v8_session = v8_session_rc.borrow_mut();
-            let v8_session_ptr = v8_session.as_mut();
+            let (v8_session_ptr, msg) = session_stream_item;
             InspectorSession::dispatch_message(v8_session_ptr, msg);
             continue;
           }
@@ -510,7 +506,7 @@ impl task::ArcWake for InspectorWaker {
 /// eg. Websocket or another set of channels.
 struct InspectorSession {
   v8_channel: v8::inspector::ChannelBase,
-  v8_session: Rc<RefCell<v8::UniqueRef<v8::inspector::V8InspectorSession>>>,
+  v8_session: NonNull<v8::inspector::V8InspectorSession>,
   proxy: InspectorSessionProxy,
 }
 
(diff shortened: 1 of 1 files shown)


bartlomieju:
Oooops 😬 `terminated by signal SIGSEGV (Address boundary error)`

core/inspector.rs
```diff
@@ -507,29 +521,32 @@ impl InspectorSession {
...
-    v8_session_ptr.dispatch_protocol_message(msg);
+  fn dispatch_message(
+    v8_session_ptr: *mut v8::inspector::V8InspectorSession,
+    msg: String,
+  ) {
+    let msg = v8::inspector::StringView::from(msg.as_bytes());
+    let v8_session = unsafe {
+      &mut *(v8_session_ptr as *mut v8::inspector::V8InspectorSession)
+    };
+    v8_session.dispatch_protocol_message(msg);
```
piscisaureus, line 549:
```suggestion
    unsafe {
      (*v8_session_ptr).dispatch_protocol_message(msg);
    }
```

core/inspector.rs
```diff
@@ -547,11 +564,11 @@ impl InspectorSession {
   pub fn break_on_next_statement(&mut self) {
     let reason = v8::inspector::StringView::from(&b"debugCommand"[..]);
     let detail = v8::inspector::StringView::empty();
-    self
-      .v8_session
-      .borrow_mut()
-      .as_mut()
-      .schedule_pause_on_next_statement(reason, detail);
+
+    let v8_session = unsafe {
```
piscisaureus, line 568: See comment above about aliasing `&mut`. Call the method in an unsafe block.

core/inspector.rs
```diff
@@ -492,7 +506,7 @@ impl task::ArcWake for InspectorWaker {
 /// eg. Websocket or another set of channels.
 struct InspectorSession {
   v8_channel: v8::inspector::ChannelBase,
-  v8_session: Rc<RefCell<v8::UniqueRef<v8::inspector::V8InspectorSession>>>,
+  v8_session: NonNull<v8::inspector::V8InspectorSession>,
```
piscisaureus, line 509: Better!

core/inspector.rs
```diff
@@ -371,7 +385,7 @@ struct SessionContainer {
   v8_inspector: Rc<RefCell<v8::UniquePtr<v8::inspector::V8Inspector>>>,
   session_rx: UnboundedReceiver<InspectorSessionProxy>,
   handshake: Option<Box<InspectorSession>>,
-  established: FuturesUnordered<Box<InspectorSession>>,
+  established: SelectAll<Box<InspectorSession>>,
```
piscisaureus, line 388: Good use of SelectAll!

core/inspector.rs
piscisaureus, line 509: You need to keep using `UniqueRef` here, otherwise this will cause a memory leak (since you're using `UniqueRef::into_raw()` below).

core/inspector.rs
```diff
@@ -547,11 +563,8 @@ impl InspectorSession {
   pub fn break_on_next_statement(&mut self) {
     let reason = v8::inspector::StringView::from(&b"debugCommand"[..]);
     let detail = v8::inspector::StringView::empty();
-    self
-      .v8_session
-      .borrow_mut()
-      .as_mut()
-      .schedule_pause_on_next_statement(reason, detail);
+
+    (*self.v8_session).schedule_pause_on_next_statement(reason, detail);
```
piscisaureus, line 567: Not quite right either, considering that v8_session derefs to a &mut which may be aliased. Rusty_v8 should probably have a `UniquePtr::as_mut_ptr()` method but it doesn't.

bartlomieju: How do we handle that in the meantime then?

piscisaureus (approved):

core/inspector.rs
piscisaureus: > How do we handle that in the meantime then?

There doesn't seem to be a good solution. 
Maybe add `// TODO: use raw `*mut V8InspectorSession` pointer, as this reference may become aliased.`

PR #13274, opened 2022-01-04
fix: support mts files for `deno test` & `deno fmt`

### LINKED ISSUE #13268: `deno test` command fails to detect `.mts` test files
The `deno test` command fails to detect test files that have the `.mts` file extension.

For example, with the `foo_test.mts` file present, running:

```
deno test
```

Results in:

```
error: No test modules found
```

The `.mts` file extension needs to be supported for tests so they can properly import `.mjs` ESM containing JSDoc TypeScript types.

crowlKats:
Closes #13268

### DIFF
--- a/cli/fs_util.rs
+++ b/cli/fs_util.rs
@@ -191,6 +191,8 @@ pub fn is_supported_test_ext(path: &Path) -> bool {
         | "jsx"
         | "mjs"
         | "mts"
+        | "cjs"
+        | "cts"
         | "md"
         | "mkd"
         | "mkdn"
@@ -493,6 +495,8 @@ mod tests {
     assert!(is_supported_test_ext(Path::new("foo.JSX")));
     assert!(is_supported_test_ext(Path::new("foo.mjs")));
     assert!(is_supported_test_ext(Path::new("foo.mts")));
+    assert!(is_supported_test_ext(Path::new("foo.cjs")));
+    assert!(is_supported_test_ext(Path::new("foo.cts")));
     assert!(!is_supported_test_ext(Path::new("foo.mjsx")));
     assert!(!is_supported_test_ext(Path::new("foo.jsonc")));
     assert!(!is_supported_test_ext(Path::new("foo.JSONC")));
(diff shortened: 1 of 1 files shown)


bartlomieju:
This looks fine, but it would be great to add `mts` files to `cli/tests/testdata/fmt` and `cli/tests/testdata/test/` to ensure they do picked up automatically.

Following up on the conversation from Discord, I think we should also recognize `cjs` and `cts` files for both `deno fmt` and `deno test` (the latter only in --compat mode). Thoughts?

cli/fs_util.rs
```diff
@@ -130,7 +130,7 @@ pub fn resolve_from_cwd(path: &Path) -> Result<PathBuf, AnyError> {
 /// Checks if the path has extension Deno supports.
 pub fn is_supported_ext(path: &Path) -> bool {
   if let Some(ext) = get_extension(path) {
-    matches!(ext.as_str(), "ts" | "tsx" | "js" | "jsx" | "mjs")
+    matches!(ext.as_str(), "ts" | "tsx" | "js" | "jsx" | "mjs" | "mts")
```
dsherret, line 133: I think we might as well add cts and cjs because of node compat mode?

dsherret (approved):
LGTM

cli/fs_util.rs
```diff
@@ -130,7 +130,10 @@ pub fn resolve_from_cwd(path: &Path) -> Result<PathBuf, AnyError> {
 /// Checks if the path has extension Deno supports.
 pub fn is_supported_ext(path: &Path) -> bool {
```
dsherret, line 131: Actually, I just realized this function name and looked at its usages. Perhaps we should remove `cjs` and `cts` from this function. For `is_supported_ext_fmt` I think it makes sense to have `cjs` and `cts`.

crowlKats: Yes, though it would be weird supporting `cjs` & `cts` only in fmt and test, and not lint. just seems like it would be unexpected behaviour for someone that is using a `cjs` or `cts` fiel

crowlKats: With compat mode only, this seems also weird, as to allow a certain filetype depending on compat just seems weird

dsherret: Sorry, my brain is fried at the moment after the swc upgrade. I meant for this and test to not enable it, but yeah I'm not sure what the desired behaviour is. Would be good to confirm with @bartlomieju before merging.

The one issue would be that picking up cjs/cts files would error for deno test if `--compat` mode wasn't enabled, but yeah maybe that's not a big deal and it would just be a reminder to turn on `--compat`.

bartlomieju: I think it's okay to always include `cjs` and `cts`

bartlomieju (approved):

PR #13275, opened 2022-01-04
fix(lsp): handle repeating patterns in registry correctly

kitsonk:
This handles situations where the pattern is a terminating repeating pattern on a registry URL and the variable appears in the endpoint.


bartlomieju (approved):
I can't say I fully grasp what's going on here, but there are extensive tests so LGTM

kitsonk:
I'm finding them as I implement the v2 registry fully, so they are obscure, but I know it is driving the correct behaviour.

PR #13283, opened 2022-01-04
fix(ext/ffi): throw errors instead of panic

### LINKED ISSUE #13280: ffi: Panic on incorrect FFI parameters
# Code
```
const user32 = Deno.dlopen("user32.dll", {
    SystemParametersInfoA: {
        parameters: [
            "u32", 
            "u32",
            "pointer", 
            "u32", 
        ], 
        result: "i32" 
    } 
});

const result = user32.symbols.SystemParametersInfoA(0x0014, null, new TextEncoder().encode("D:/flower1.jpg"), 0x01)
```

# Panic
```

Platform: windows x86_64
Version: 1.17.1
Args: ["deno", "run", "--unstable", "--allow-all", "natives.ts"]

thread 'main' panicked at 'Expected ffi arg value to be an unsigned integer', ext\ffi\lib.rs:281:6
```

# Mitigation
In this case, you can replace the `null` in `user32.symbols.Syst...` to 0, to match the symbol's parameters and the panic no longer occurs.

DjDeveloperr:
Also improves error message for symbol which could not be found (this has been bugging me when developing FFI related projects...)

closes #13280

### DIFF
--- a/ext/ffi/lib.rs
+++ b/ext/ffi/lib.rs
@@ -291,7 +291,7 @@ fn value_as_uint<T: TryFrom<u64>>(value: Value) -> Result<T, AnyError> {
   match value.as_u64().and_then(|v| T::try_from(v).ok()) {
     Some(value) => Ok(value),
     None => Err(type_error(format!(
-      "Expected FFI argument to be an unsigned integer {:?}",
+      "Expected FFI argument to be an unsigned integer, but got {:?}",
       value
     ))),
   }
--- a/ext/ffi/lib.rs
+++ b/ext/ffi/lib.rs
@@ -252,10 +252,8 @@ impl NativeValue {
           }
         } else {
           Self {
-            pointer: u64::from(
-              serde_json::from_value::<U32x2>(value)
-              .expect("Expected ffi arg value to be a tuple of the low and high bits of a pointer address")
-            ) as *const u8,
+            pointer: u64::from(serde_json::from_value::<U32x2>(value)?)
+              as *const u8,
           }
         }
       }
@@ -483,7 +481,12 @@ fn ffi_call(args: FfiCallArgs, symbol: &Symbol) -> Result<Value, AnyError> {
     match native_type {
       NativeType::Pointer => match value.as_u64() {
         Some(idx) => {
-          let buf = buffers.get(idx as usize).unwrap().unwrap();
+          let buf = buffers
+            .get(idx as usize)
+            .ok_or_else(|| {
+              generic_error(format!("No buffer present at index {}", idx))
+            })?
+            .unwrap();
           native_values.push(NativeValue::buffer(buf.as_ptr()));
         }
         _ => {
--- a/ext/ffi/lib.rs
+++ b/ext/ffi/lib.rs
@@ -1,7 +1,9 @@
 // Copyright 2021 the Deno authors. All rights reserved. MIT license.
 
 use deno_core::error::bad_resource_id;
+use deno_core::error::generic_error;
 use deno_core::error::range_error;
+use deno_core::error::type_error;
 use deno_core::error::AnyError;
 use deno_core::include_js_files;
 use deno_core::op_async;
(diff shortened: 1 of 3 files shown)


eliassjogreen (approved):
LGTM, thank you finally!

test_ffi/tests/test.js
```diff
@@ -18,6 +18,19 @@ try {
...
+    non_existent_symbol: {
+      parameters: [],
+      result: "void",
+    },
+  });
+} catch (e) {
+  console.log(
+    `error includes symbol name? ${e.message.includes("non_existent_symbol")}`,
+  );
+}
```
bartlomieju, line 32: Nit pick, but maybe import `assertThrows` from `test_util/std/`

test_ffi/tests/test.js
```diff
@@ -88,16 +88,12 @@ console.log(Boolean(dylib.symbols.is_null_ptr(ptr)));
 console.log(Boolean(dylib.symbols.is_null_ptr(null)));
 console.log(Boolean(dylib.symbols.is_null_ptr(Deno.UnsafePointer.of(into))));
 console.log(dylib.symbols.add_u32(123, 456));
-try {
+assertThrows(() => {
   dylib.symbols.add_u32(-1, 100);
-} catch (_) {
-  console.log("passing negative integer in u32 argument doesn't panic");
-}
```
bartlomieju: I'd be great to still assert this message

test_ffi/tests/test.js
```diff
@@ -88,16 +88,12 @@ console.log(Boolean(dylib.symbols.is_null_ptr(ptr)));
...
-try {
+});
+assertThrows(() => {
   dylib.symbols.add_u32(null, 100);
-} catch (_) {
-  console.log("passing null in u32 argument doesn't panic");
-}
```
bartlomieju: Ditto

bartlomieju (commented):
@DjDeveloperr please add expected messages to `assertThrows` and we can land this PR 👍

ext/ffi/lib.rs
```diff
@@ -450,22 +473,30 @@ fn ffi_call(args: FfiCallArgs, symbol: &Symbol) -> Result<Value, AnyError> {
...
-      if let NativeType::Pointer = native_type {
-        if let Some(idx) = value.as_u64() {
-          if let Some(&Some(buf)) = buffers.get(idx as usize) {
-            return NativeValue::buffer(buf.as_ptr());
-          }
+  {
+    match native_type {
+      NativeType::Pointer => match value.as_u64() {
+        Some(idx) => {
+          let buf = buffers.get(idx as usize).unwrap().unwrap();
```
littledivy, line 486: These unwraps should be thrown as errors too.

ext/ffi/lib.rs
```diff
@@ -194,44 +206,44 @@ union NativeValue {
...
       },
       NativeType::F32 => Self {
-        f32_value: value_as_f32(value),
+        f32_value: value_as_f32(value)?,
       },
       NativeType::F64 => Self {
-        f64_value: value_as_f64(value),
+        f64_value: value_as_f64(value)?,
       },
       NativeType::Pointer => {
```
littledivy, line 248: The `.expect` below should be a type_error too

test_ffi/tests/integration_tests.rs
```diff
@@ -29,15 +29,18 @@ fn basic() {
...
   if !output.status.success() {
     println!("stdout {}", stdout);
     println!("stderr {}", stderr);
   }
   println!("{:?}", output.status);
   assert!(output.status.success());
+  if stderr.starts_with("Check ") {
+    // Remove the "Check ..." line from stderr.
+    stderr = &stderr[(stderr.find('\n').unwrap() + 1)..];
+  }
```
littledivy, line 42: Pass `--quiet` to the command maybe?

DjDeveloperr: TIL that exists, will use that.

ext/ffi/lib.rs
DjDeveloperr: I thought it'd not be required since these are actually `Option` and always expected to be there according to the JS side which calls it. I'll throw these anyway, but what should be the error message?

littledivy: Actually I agree, these can be only reachable if our internal JS code is buggy. Let's keep them as panics we would want users to report them as bugs.

littledivy (approved):
LGTM Thanks

PR #13289, opened 2022-01-05
feat(coverage): add "--output" flag

### LINKED ISSUE #13236: Add output argument to `deno coverage` when used with `--lcov`
I think it's better to provide an option to pipe the output to a file of `deno converage cov_profile/ --lcov` instead of using `>`, which is not supported in every operating system, AFAIK.

Something like `deno converage cov_profile/ --lcov --output=path/to/lcov.info` — possibly suppressing any output from `lvoc`, if any.

VishnuJin:
This pr adds `output` arg to `deno coverage` to pipe the stdout to the given file, ~~just like `deno bundle`~~

fixes #13236

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3780,9 +3780,9 @@ dependencies = [
 
 [[package]]
 name = "swc_common"
-version = "0.17.5"
+version = "0.17.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfac713c943b08c8e364801b24a2ea24ea657eadf22826840aa858eccd04d828"
+checksum = "a23e045ca915a377988b2a419cc28f4f981ce40368084d5b18b7d15a84f84872"
 dependencies = [
  "ahash",
  "ast_node",
@@ -3808,9 +3808,9 @@ dependencies = [
 
 [[package]]
 name = "swc_ecma_ast"
-version = "0.65.4"
+version = "0.65.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c84de6f01fe49c594bfc6f48c2b16e475770799cb356a048c60451f6e3824e2"
+checksum = "ff1aefdc512c8dfefa1de793c0e62a192bd74a7fa7e5affe3f08885751c127ee"
 dependencies = [
  "is-macro",
  "num-bigint",
@@ -3823,9 +3823,9 @@ dependencies = [
 
 [[package]]
 name = "swc_ecma_codegen"
-version = "0.89.4"
+version = "0.89.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f3dda60113af30e9d9da0b15753850b8e794ee15d9cfe8aadfe05651fb692bf8"
+checksum = "b627edd495a93306f21e435c06227fe57910c1f6054f4c966e917c001142da9e"
 dependencies = [
  "bitflags",
  "memchr",
@@ -3900,9 +3900,9 @@ dependencies = [
 
 [[package]]
 name = "swc_ecma_transforms"
-version = "0.114.6"
+version = "0.114.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b01f4bdd4e2a199e4a30f274c99f8f08607da5419191d89f321a7edcf21ea44f"
+checksum = "7a427f8c29e7217b0405c72544bf4049023fa7b025d1a9c300a2f706bfd04f6b"
 dependencies = [
  "swc_atoms",
  "swc_common",
@@ -4079,9 +4079,9 @@ dependencies = [
 
 [[package]]
 name = "swc_ecmascript"
-version = "0.111.14"
+version = "0.111.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c6a4f14fc36deb60d385359dca5446ad2a6dc9550cb422c62d154d4bb1bab31b"
+checksum = "d7dc93403df65d3d523bd6f2a4f7906e1c7cf4daa5d3ab808785abb6b60fbfe9"
 dependencies = [
  "swc_ecma_ast",
  "swc_ecma_codegen",


VishnuJin:
@bartlomieju 
I believe `deno bundle` accepts only one source file and hence the second arg can be interpreted as out_file but in case of `deno coverage` it accepts multiple files, how can we assume that the last arg is the `out_file` argument unless explicitly specified ?

bartlomieju:
> @bartlomieju I believe `deno bundle` accepts only one source file and hence the second arg can be interpreted as out_file but in case of `deno coverage` it accepts multiple files, how can we assume that the last arg is the `out_file` argument unless explicitly specified with `out_file` ?

Good point. Let's go with explicit `--out` argument.

VishnuJin:
@bartlomieju apologies for the delay!! I have resumed working on this and can share progress soon..

cli/tools/coverage/mod.rs
```diff
@@ -347,7 +347,12 @@ fn create_reporter(
 }
 
 trait CoverageReporter {
-  fn report(&mut self, coverage_report: &CoverageReport, file_text: &str);
+  fn report(
+    &mut self,
+    coverage_report: &CoverageReport,
+    file_text: &str,
+    out_writer: &mut Box<dyn Write>,
```
VishnuJin, line 354: @bartlomieju I don't think its good idea to add another parameter to report
() ? because for `PrettyCoverageReporter` this arg is unnecessary

cli/flags.rs
```diff
@@ -730,6 +731,17 @@ Generate html reports from lcov:
         .help("Output coverage report in lcov format")
         .takes_value(false),
     )
+    .arg(Arg::new("out_file")
```
bartlomieju, line 734: Rename to `output`, it's unwieldy to require underscores in CLI flags

cli/tools/coverage/mod.rs
bartlomieju: Agree, pass this parameter to `LcovCoverageReporter` directly

bartlomieju (changes requested):

cli/flags.rs
VishnuJin: there were couple of `out_file` arg name used for same logic, and some how I feel like `--out-file` is kinda more descriptive than `--output` ?
https://github.com/denoland/deno/blob/380ebbd6d55f8933f614c405ee1050f7231d7ed4/cli/flags.rs#L559-L562

VishnuJin: anyways, if you want it to be `--output` please let me know again, I'll change it !!

bartlomieju: `out_file` is only used in `deno bundle` and it is a positional argument so you don't type out it's name explicitly. Let's go with `output` which matches `deno compile`.

cli/tools/coverage/mod.rs
VishnuJin: @bartlomieju I have changed the approach a bit  !!

cli/tools/coverage/mod.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 1: Revert

cli/tools/coverage/mod.rs
```diff
@@ -362,31 +365,53 @@ impl LcovCoverageReporter {
...
     }
 
     for function in &coverage_report.named_functions {
-      println!("FNDA:{},{}", function.execution_count, function.name);
+      writeln!(
+        out_writer,
+        "FNDA:{},{}",
+        function.execution_count, function.name
+      )
+      .unwrap();
```
bartlomieju, line 404: I think we shouldn't use `unwrap()` here but forward the errors to the caller

cli/tools/coverage/mod.rs
```diff
@@ -641,8 +680,10 @@ pub async fn cover_files(
       &script_coverage,
       &transpiled_source,
       &maybe_source_map,
+      &out_mode,
     );
 
+    // coverage_report.output = out_mode.clone();
```
bartlomieju, line 686: Remove

VishnuJin:
@bartlomieju I noticed  `js_unit_tests` tests in`fastci macOS` fail couple of time although it passed in my local runs

bartlomieju:
@VishnuJin no worries, seems like a flake.

VishnuJin:
@bartlomieju awaiting for you review 🙂

cli/flags.rs
```diff
@@ -732,6 +733,17 @@ Generate html reports from lcov:
...
+    .arg(Arg::new("output")
+    .long("output")
+    .help("Output file (defaults to stdout) for lcov")
+    .long_help("Exports the coverage report in lcov format to the given file.
+Filename should be passed along with '='
+For example '--output=foo.lcov'
+If no --output arg is specified then the report is written to stdout."
+  )
+    .takes_value(true)
+    .require_equals(true)
```
bartlomieju, line 745: Do we really need to require equals sign here? Seems we could do the same as in `--lcov` and not require it

cli/flags.rs
```diff
@@ -1863,8 +1875,15 @@ fn coverage_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
     None => vec![],
   };
   let lcov = matches.is_present("lcov");
+  let output = if let Some(output) = matches.value_of("output") {
+    flags.allow_write = Some(vec![]);
```
bartlomieju, line 1879: Uhoh, this is a red flag. Why is this flag being set here?

VishnuJin: sorry, this was mistake, I have removed this !!

cli/flags.rs
VishnuJin: - all the other args in the coverage like `include`, `exclude`, `regex` who accepts values all requires equals
- I think since the last arg `files` accepts multiple values it's better to be explicit about specifying

bartlomieju: Okay, makes sense, I was confused by the example shown above. Let's change the example to be less ambiguous

cli/flags.rs
```diff
@@ -689,7 +690,7 @@ an url to match it must match the include pattern and not match the exclude patt
 
 Write a report using the lcov format:
 
-  deno coverage --lcov cov_profile > cov.lcov
+  deno coverage --lcov cov_profile --output=cov.lcov
```
bartlomieju, line 693:
```suggestion
  deno coverage --lcov --output=cov.lcov cov_profile/
```

cli/flags.rs
```diff
@@ -732,6 +733,17 @@ Generate html reports from lcov:
         .help("Output coverage report in lcov format")
         .takes_value(false),
     )
+    .arg(Arg::new("output")
+    .long("output")
```
bartlomieju, line 737: This arg should require `--lcov` arg to be present, otherwise I can do:
```
deno coverage --output=cov.lcov cov/
```
which makes no sense, since it means to print test report to stdout

bartlomieju (commented):
@VishnuJin I just tested this PR and it works nicely, we should land it for 1.19. I found one quirk though, please see comment below

cli/flags.rs
VishnuJin: yeah, I somehow missed `requires` method in clap and implemented the following which is obviously not so good, and I will change this now
https://github.com/denoland/deno/blob/eff2796181d20dbe0c8a18dbe45d12c4bb5e57f2/cli/tools/coverage/mod.rs#L618-L630

VishnuJin: @bartlomieju thanks for the hint,
https://github.com/denoland/deno/blob/c4356c02b47d36411f3f68adc5c43883d4df2d3e/cli/flags.rs#L736-L738

VishnuJin:
@bartlomieju should I do something here ?

bartlomieju:
@VishnuJin no, I made a mistake rebasing your PR, need to land other PR first before it's fixed.

bartlomieju (approved):
LGTM, thank you @VishnuJin

PR #13296, opened 2022-01-06
feat(test): better errors for resource sanitizer

lucacasonato:
This commit makes the errors produced from the resource sanitizer much
more human readable. It does this by using real words rather than our
"resource names" when referring to resources, and by giving helpful
hints on how to clean up each of the resources.


bnoordhuis (approved):
Good idea, really nice.

bartlomieju (approved):
LGTM too

dsherret (approved):
Nice!

PR #13300, opened 2022-01-07
fix(serde_v8): support #[serde(default)]

### LINKED ISSUE #13185: serde_v8 serde default support
I've got a type defined with a field using `#[serde(default)]` which is a `BTreeMap`, something like:

```rs
#[derive(Debug, Serialize, PartialEq, Deserialize, Clone)]
pub struct SomeThing {
    pub name: Option<String>,
    #[serde(default)]
    pub args: BTreeMap<String, String>,
}
```

When I try to deserialize with `from_v8` into this type with that field missing, I get the error `ExpectedObject`. If, on the other hand, I make the object type optional via:

```rs
#[derive(Debug, Serialize, PartialEq, Deserialize, Clone)]
pub struct SomeThing {
    pub name: Option<String>,
    pub args: Option<BTreeMap<String, String>>,
}
```

it then works fine - this makes me think serde_v8 isn't fully implementing the default semantics perhaps?

AaronO:
Fixes #13185 but slightly degrades error quality

Also rewrote struct deserialization to use a `SeqAccess` instead of a `MapAccess` making it simpler

### DIFF
--- a/serde_v8/src/de.rs
+++ b/serde_v8/src/de.rs
@@ -375,7 +375,7 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
     // Regular struct
     let obj = v8::Local::<v8::Object>::try_from(self.input)
       .map_err(|_| Error::ExpectedObject)?;
-    let map = ObjectAccess {
+    let struct_access = StructAccess {
       fields,
       obj,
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

PR #13301, opened 2022-01-07
fix(tsc): Add typings for `Intl.ListFormat`

### LINKED ISSUE #13199: Missing ListFormat types?
I can do this in the REPL and it works:

```ts
new Intl.ListFormat("en", {
    style: "long",
    type: "conjunction",
}).format(['red', 'green', 'blue']) // returns "red, green, and blue"
```

But if I put the same code in my Deno application, I get:

```sh
[ERROR]: Property 'ListFormat' does not exist on type 'typeof Intl'.
```

when I try to bundle it.

andreubotella:
V8 has supported `Intl.ListFormat` since version 7.2, but TypeScript doesn't have typings for it yet. This PR manually adds those typings, copying them from microsoft/TypeScript#47254.

Fixes #13199.

### DIFF
--- a/cli/dts/lib.es2021.intl.d.ts
+++ b/cli/dts/lib.es2021.intl.d.ts
@@ -41,4 +41,106 @@ declare namespace Intl {
         formatRange(startDate: number | bigint, endDate: number | bigint): string;
         formatRangeToParts(startDate: number | bigint, endDate: number | bigint): NumberFormatPart[];
     }
+
+    /**
+     * The locale matching algorithm to use.
+     *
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
I believe these declarations should be added to `lib.deno.shared_globals.d.ts`

andreubotella:
As per the conversation on Discord, it's fine to declare this in `lib.es2021.intl.d.ts`, so this can be merged.

bartlomieju (approved):

Issue #13302, opened 2022-01-07
Emit, source map, and declaration cache in deno dir is racy across multiple processes


dsherret:
I've run into two caching issues while using deno the past two days where I was sure my code was ok, but I was getting type checking errors.

Upon closer inspection, our cache code is racy across multiple processes since the hash is stored in a separate file.

Also, looks like if you kill the process between these two statements then the emit won't be saved correctly and won't ever be reloaded until you change the source text:

https://github.com/denoland/deno/blob/12423e16b7df39b2bf995efd1376ba4b6ef41e02/cli/emit.rs#L458-L459

To solve this, we should store both the version hash along with each artifact in the same file. I've already started working on a fix.

PR #13303, opened 2022-01-07
fix: use a single file for the emit cache to eliminate multiple cache files getting out of sync

dsherret:
Solves part of #13302.

This PR combines the version hash (`.meta`), emit (`.js`), source map (`.map`), and declaration (`.d.ts`) cache files into a single file (`.emitinfo`) to eliminate race conditions between the four files on the file system. This should now make it impossible for a version hash to be stored for an emit that doesn't match.

This does not completely solve #13302 because our emit process still relies on using cache data from the file system that could be out of date with the data stored in the graph if another deno process edits it. We should fix this part in a future PR.

At the current moment, this will be less performant because it will be loading all the data from the file when checking the version hash. This can be essentially eliminated when implementing the rest of #13302 because we'll be able to keep everything in memory instead of visiting the cache multiple times. There will only be a performance penalty when the cache is out of date instead of also when it's also not out of date.


cli/cache.rs
```diff
@@ -84,51 +145,22 @@ impl FetchCacher {
...
+          .map(|p| location.join(p))
+          .filter(|p| p.is_file()); // todo: dsherret: this condition is racy so we should remove it
       Some(CacheInfo {
         local: Some(local),
-        emit,
-        map,
+        // todo(dsherret): deno_graph should not have multiple files for emit and map
+        emit: emit.clone(),
+        map: emit,
       })
```
dsherret, line 164: What is this `CacheInfo` used for? I forget and didn't figure it out on a quick glance as it doesn't seem to be used (unless my editor is failing me).

cli/cache.rs
```diff
@@ -350,3 +340,195 @@ impl CacherLoader for MemoryCacher {
     self
   }
 }
+
+fn get_cache_filename(url: &Url) -> Option<PathBuf> {
```
dsherret, line 344: This function and the one below were moved from `disk_cache.rs` into here. I didn't change the contents, only moved them off the struct and made it private here.

cli/proc_state.rs
```diff
@@ -574,23 +575,21 @@ impl ProcState {
           | MediaType::Json => code.as_ref().clone(),
           MediaType::Dts => "".to_string(),
           _ => {
-            let emit_path = self
+            match self
               .dir
               .gen_cache
-              .get_cache_filename_with_extension(&found, "js")
```
dsherret, line 580: We had multiple ways for getting this data from the cache. Here it was accessing it via `DiskCache` and other places it was `FetchCache`. I've combined them to one api (`.get_emit_data`)

cli/emit.rs
```diff
@@ -694,27 +730,31 @@ pub(crate) fn emit(
...
     let needs_reload =
       options.reload && !options.reload_exclusions.contains(&module.specifier);
     let version = get_version(module.source.as_bytes(), &config_bytes);
     let is_valid = cache
-      .get(CacheType::Version, &module.specifier)
-      .map_or(false, |v| {
-        v == get_version(module.source.as_bytes(), &config_bytes)
-      });
+      .get_emit_data(&module.specifier)
+      .map_or(false, |d| d.version_hash == version);
```
dsherret, line 741: Here is the more inefficient code (I haven't measured it), but fixing the todo outlined on line 733 and the issue I mentioned in the PR description should make it so that we don't get any performance penalty for an up to date cache.

cli/cache.rs
```diff
@@ -3,51 +3,112 @@
...
-#[derive(Debug, Deserialize, Serialize)]
-pub struct EmitMetadata {
-  pub version_hash: String,
-}
-
 pub(crate) enum CacheType {
-  Declaration,
-  Emit,
-  SourceMap,
   TypeScriptBuildInfo,
```
dsherret, line 32: This PR actually currently panics on old cache dir files (just tried it now). The old TS build info just needs to be invalidated somehow.

cli/emit.rs
```diff
@@ -452,19 +475,25 @@ pub(crate) fn check_and_maybe_emit(
...
+            version_hash: get_version(source_bytes, &config_bytes),
+            text: None,
+            declaration: None,
+            map: None,
+          });
         match emit.media_type {
           MediaType::JavaScript | MediaType::Mjs | MediaType::Cjs => {
-            let version = get_version(source_bytes, &config_bytes);
-            cache.set(CacheType::Version, &specifier, version)?;
-            cache.set(CacheType::Emit, &specifier, emit.data)?;
```
dsherret, line 459: An example of a previous bug would be if the process got killed right before entering this function call.

cli/cache.rs
```diff
@@ -3,51 +3,112 @@
...
+use std::path::Component;
+use std::path::Path;
+use std::path::PathBuf;
+use std::path::Prefix;
 use std::sync::Arc;
 
-#[derive(Debug, Deserialize, Serialize)]
-pub struct EmitMetadata {
-  pub version_hash: String,
-}
```
dsherret, line 25: An alternative solution is to store hashes of all the cached files here as well. This would be more complex though.

bartlomieju:
I verified that this PR works with Chrome Devtools; we should also check that VSCode debugger still works (preferable JetBrains IDE too).

dsherret:
> I verified that this PR works with Chrome Devtools; we should also check that VSCode debugger still works (preferable JetBrains IDE too).

@bartlomieju I tested in VSCode and it works fine and I don't see why it wouldn't—nothing knows about the location of those files on the disk except the code in Deno that goes to load them (unless there's some other code I'm missing that accesses this filepath).

dsherret:
I'm going to close this PR. I'm going to open a new one that uses an sqlite db instead.

Issue #12949, opened 2021-11-30
Add ignore directives to bundled code


bartlomieju:
When using `deno bundle` generated code should be annotated with ignore directives for `deno fmt` and `deno lint`:

```js
// deno-fmt-ignore-file
// deno-lint-ignore-file

...
// bundled code
```
It makes no sense to lint bundled code, and reformatting it will break source maps.

We might even add a comment informing that the code you see is a result of bundling:
```
// This code was bundled using `deno bundle` and it's not recommended to edit it manually
```

Ref #12916

bartlomieju:
This is actually not a "good first issue". 

The change needs to be done around here:
https://github.com/denoland/deno/blob/9a1066869457037d6ab24965da82e80e1e5c77a9/cli/emit.rs#L580-L621

However, since comments should be appended *after* bundling, the source map needs to be updated before being written to account for additional lines.

gliheng:
Hi @bartlomieju . I'd like to work on this as my first contribution to this project.

bartlomieju:
@gliheng thanks for interest in this issue, please not that this is rather a hard issue and might not be best suited for first contribution. I don't have anymore pointers to provide at the moment on how it could be implemented.

gliheng:
After some digging, I found that we can just use JsWriter to write ignore directives in bundled file as in this commit https://github.com/gliheng/deno/commit/aeb4c63d926b6969529540b7ccdae02d619b7c38

However, sourceMap generation has to be enabled. Looks like it's currently disabled in bundled mode.
https://github.com/denoland/deno/blob/46f2ff120587cd1cb7b11e66d41ad6ad46fb09a8/cli/config_file.rs#L105-L115

PR #13309, opened 2022-01-08
feat(cli): add ignore directives to bundled code

gliheng:
This fixes https://github.com/denoland/deno/issues/12949 adds ignore directives in deno bundle emitted code/


bartlomieju:
Thanks for the PR, can you add a test that checks that these directives are present in the bundled code?

gliheng:
Sure, will do.

bartlomieju:
@gliheng looks good! It appears there are two failing tests:
- `integration::compile::standalone_error`
- `integration::compiler_api`
The former appears to have difference in line numbers in stack trace (which is expected, since you add 3 lines to all bundles), while the latter needs a bit more debugging to see what's going wrong. Let me know if you need any help.

gliheng:
This latter one is caused by this line 
`assert(files["deno:///bundle.js"].startsWith("(function() {\n"));`
Since I added some comments at the front, this is no long true.
To fix this, either we only add ignore directives for deno bundle but not Deno.emit.
Or we add an option to toggle this in EmitOptions for Deno.emit. I dont't know which way is preferred. Any idea on this?

gliheng:
I tried the first approach and PR was updated

bartlomieju:
> I tried the first approach and PR was updated

I believe we should add the directives for `deno bundle` output, as well as for `Deno.emit()`. @kitsonk do you concur?

gliheng:
@bartlomieju  These directives are also added to Deno.emit and updated test cases.

bartlomieju (approved):
LGTM, thank you @gliheng!

PR #13312, opened 2022-01-08
test: add inspector test with ts files

### LINKED ISSUE #13305: Add inspector test with TS files
Need to add a test for inspector that uses TS files (and thus source maps). The test should have at least two imports: one TS file, one JS file.

bartlomieju:
Closes #13305


dsherret (approved):
LGTM

PR #13313, opened 2022-01-08
feat: auto-discover config file

ry:


cli/config_file.rs
```diff
@@ -156,6 +158,61 @@ pub const IGNORED_RUNTIME_COMPILER_OPTIONS: &[&str] = &[
...
+pub fn discover(flags: &crate::Flags) -> Result<Option<ConfigFile>, AnyError> {
+  if let Some(config_path) = flags.config_path.as_ref() {
+    Ok(Some(ConfigFile::read(config_path)?))
+  } else {
+    let mut checked = HashSet::new();
+    for f in flags.config_path_args() {
+      if let Some(cf) = discover_inner(&f, &mut checked)? {
+        return Ok(Some(cf));
+      }
+    }
```
bartlomieju, line 173: This is a bit unexpected - IIRC the algorithm we agreed upon was supposed to find a config file for each of the passed arguments, not a config file that was found first. Could you clarify this @ry?

kitsonk:
Is this shippable without addressing the language server?  Feels like it should be included in this PR.

ry:
> Is this shippable without addressing the language server? Feels like it should be included in this PR.

What needs to be addressed in the language server?

kitsonk:
Currently the lsp requires an explicit option of `"config"` to be set in the client configuration for a configuration file to apply:

https://github.com/denoland/deno/blob/c487b7ed54a36edcf324005a6ff6e76663544e06/cli/lsp/config.rs#L157-L159

And when there is something impacted in the environment, like when the client configuration or the configuration file iteself changes, then the language server processes an update:

https://github.com/denoland/deno/blob/c487b7ed54a36edcf324005a6ff6e76663544e06/cli/lsp/language_server.rs#L524

If we have automatic behaviour to identify the file when `deno run`, then it only seem logical that the behaviour would be similar by default for the language server, not requiring explicit configuration, which mirrors the command line today, and I would assume that delivery of that functionality would be part of this feature, since a fair amount of people utilise the language server.

Generally this should be fairly straight forward with just a "workspace" and looking in the "workspace root", but when you have "workspace folders" (multi-root workspaces), I suspect it could get a bit convoluted.

Side note, I just realised we are only watching JSON files in the workspace (and not JSONC):

https://github.com/denoland/deno/blob/c487b7ed54a36edcf324005a6ff6e76663544e06/cli/lsp/language_server.rs#L722-L728

And not watching for changes in `.jsonc` files, but if a configuration file is identified, it can be registered for changes, I believe.

Another discussion at one point, which could be delivered later, is that if we automatically detect a `deno.json`/`deno.jsonc` it would imply that the project/workspace was implicitly Deno "enabled" not requiring explicit enablement as it does today.

ry:
Could we just rip out the LSP config option? Seems like if we have auto-discover then maybe this option isn't necessary

kitsonk:
Will the CLI still allow `--config` to be explicitly specified?  If so, then you need a similar capability in the configuration of the language server.

cli/config_file.rs
dsherret: I've been thinking about this more just now and I think this goes beyond what we've discussed in the past. I'm wondering if when running `deno fmt` or `deno lint` should we have `deno.json` files in sub directories be resepected when there's a `deno.json` in an ancestor directory.

For example, say you have:

```
sub_dir
  - deno.json
  - file.ts
deno.json
````

1. Should running `deno fmt` have the same behaviour on `file.ts` as running `deno fmt sub_dir/file.ts`?
2. Should running `deno fmt sub_dir/file.ts` have the same behaviour as running `(cd sub_dir && deno fmt file.ts)`?

I can think of a few options:

**Option 1: Resolve based on closest deno.json**

This option would mean for `deno fmt` and `deno lint` along with `deno fmt <file-path>`, it would always resolve a config file based on the closest deno.json. So includes/excludes in parent `deno.json`s would be ignored.

* Fmt/lint option resolution: Each config file resolves from the default options
* Includes/excludes: Each config file resolves from the default options

Personally, I think this is the easiest to implement and is the least complex for users to understand. It answers yes to both the questions I asked above.

**Option 2: Resolve using resolution from all deno.json files**

This would require resolving all deno.json files in all ancestor directories, then use that information to figure out the fmt/lint options and includes/excludes for a folder based on the combined result of all config files.

* Fmt/lint option resolution: Each config file resolves from the ancestor directory's options
* Includes/excludes: Each config file resolves from the ancestor directory's includes/excludes and overrides it.

This would be complex to implement. It answers yes to both the questions I asked above.

**Option 3: Use config file resolved for root of each disjointed tree**

By "disjointed tree" I mean, directory sub trees that share no common directories from the root with other directory trees. Maybe there is a better word?

This would resolve the config file for each disjointed tree. So `deno fmt` would resolve based on the cwd and use that, but `deno fmt ./child/mod.ts ../sibling/mod.ts ../sibling/other.ts` would resolve the config file for `./child` and `../sibling` to use for their respective files.

* Fmt/lint option resolution: Each config file resolves from the root of each disjointed tree
* Includes/excludes: Each config file resolves from the root of each disjointed tree

I think this is the solution we agreed upon before? It answers no to both of the questions I asked above.

Comment (ry) [reply cli/config_file.rs:173]: Option 1 and Option 3 are the same except applying to multiple arguments. I agree with this approach.

Or we can just find the deno.json based on the first argument and if other files have a different config, erroniously ignore it for the time being. These are corner cases that we should have tests for anyway - we can move to option 1/3 in the future.

Comment (bartlomieju) [reply cli/config_file.rs:173]: I believe Option 3 was exactly what we discussed previously.

**Comment by ry** [comment]
@kitsonk I put up a little test case but it doesn't seem to be working. I'm seeing this error message from the LSP
```
Error retrieving settings for specifier (file:///var/folders/9v/kys6gqns6kl8nksyn4l1f9v40000gn/T/.tmpxnaESN/ignored.ts): Parse error: invalid type: map, expected a sequence
```
Advice appreciated...

**Comment by kitsonk** [comment]
Ah, OK, I will have to trace it through, but what is happening, versus the other tests like this, is that it is triggering the workspace folder logic which request from the client the specific configuration for a file, and for some reason it is causing this test to fail.

**Comment by bartlomieju** [comment]
Tried this PR on `deno_std`, works nicely :)

File: cli/tests/integration/lsp_tests.rs [hunk 2]
```diff
@@ -1545,35 +1545,25 @@ fn lsp_format_exclude_with_config() {
...
-          "files": {
-            "exclude": [ "ignored.ts" ]
-          }
-        }
-      }
-    "#,
-  )
-  .unwrap();
-
+  let workspace_root = temp_dir.path().canonicalize().unwrap();
```
kitsonk, line 1548: This was the fix, in that auto-discovery automatically canonicalizes the discovered path.

An open question though is how do clients behave?  Do they canonicalize files?  There isn't an easy way to determine.  We may need to fix `exclude` and `include` properties for things like format and lint to canoncilize files... 😕

kitsonk:
Looks like something is hanging with `lsp_diagnostics_refresh_dependents`...

kitsonk:
> Looks like something is hanging with `lsp_diagnostics_refresh_dependents`...

Found the problem, will push a fix after I check the rest of the tests...

kitsonk (approved):
LGTM

bartlomieju (approved):
LGTM too. 

There are some unrelated changes in this PR, but I'm not gonna nitpick.

dsherret (approved):
LGTM too!

lucacasonato (approved):
Tried it out locally, and it is working great. LGTM!

jsejcksn:
I don't see anything in the release notes about how to disable this feature. How can I prevent deno from walking the directory hierarchy and discovering configs?

PR #13320, opened 2022-01-09
feat(cli, runtime): compress snapshots

evanwashere:
depends on  denoland/rusty_v8@1cedb0eb21a05eecd67869440005b6ac77dde503 reversal
replaces v8 zlib compression with lz4 and zstd for better speed and compression ratio

| build | size |
|:---| --- |
| deno-new | 79mb |
| deno-uncompressed | 103mb |
| deno-v8-compressed | 80mb |

| `deno run` with cache (cli snapshot) | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `deno-new run -A b.ts` | 30.4 ± 0.5 | 29.6 | 32.3 | 1.03 ± 0.02 |
| `deno-uncompressed run -A b.ts` | 29.6 ± 0.4 | 28.9 | 31.5 | 1.00 |
| `deno-v8-compressed run -A b.ts` | 44.6 ± 0.5 | 43.8 | 46.6 | 1.51 ± 0.03 |

| `deno cache` with no cache (tsc snapshot) | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `deno-new cache -r b.ts` | 266.5 ± 1.6 | 264.1 | 269.0 | 1.05 ± 0.01 |
| `deno-uncompressed cache -r b.ts` | 253.8 ± 1.1 | 252.2 | 255.7 | 1.00 |
| `deno-v8-compressed cache -r b.ts` | 273.8 ± 1.4 | 271.7 | 276.2 | 1.08 ± 0.01 |

| `deno run` with no cache (cli + tsc snapshot) | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `deno-new run -r -A b.ts` | 540.8 ± 16.3 | 530.2 | 580.8 | 1.04 ± 0.03 |
| `deno-uncompressed run -r -A b.ts` | 519.2 ± 1.9 | 517.5 | 522.6 | 1.00 |
| `deno-v8-compressed run -r -A b.ts` | 577.4 ± 2.0 | 574.7 | 580.9 | 1.11 ± 0.01 |

<details>
  <summary>test source</summary>

```ts

// b.ts
export function foo<T>(bar: T) {
  return bar;
}

console.log(foo(1));

let i = 5;
while (i--) {
  const w = new Worker(new URL(`a.ts`, import.meta.url), { type: 'module' });

  w.terminate();
}
```

```ts

// a.ts
let a = 1;

export function foo<T>(bar: T) {
  return bar;
}
```
</details>

### DIFF
--- a/runtime/js.rs
+++ b/runtime/js.rs
@@ -15,6 +15,8 @@ pub static CLI_SNAPSHOT: Lazy<Box<[u8]>> = Lazy::new(
         as usize;
     let mut vec = Vec::with_capacity(size);
 
+    // SAFETY: vec is allocated with exact snapshot size (+ alignment)
+    // SAFETY: non zeroed bytes are overwritten with decompressed snapshot
     unsafe {
       vec.set_len(size);
     }
--- a/runtime/js.rs
+++ b/runtime/js.rs
@@ -1,14 +1,33 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 use deno_core::Snapshot;
 use log::debug;
+use once_cell::sync::Lazy;
 
-pub static CLI_SNAPSHOT: &[u8] =
-  include_bytes!(concat!(env!("OUT_DIR"), "/CLI_SNAPSHOT.bin"));
+pub static CLI_SNAPSHOT: Lazy<Box<[u8]>> = Lazy::new(
+  #[cold]
+  #[inline(never)]
+  || {
+    static COMPRESSED_CLI_SNAPSHOT: &[u8] =
+      include_bytes!(concat!(env!("OUT_DIR"), "/CLI_SNAPSHOT.bin"));
+
+    let size =
+      u32::from_le_bytes(COMPRESSED_CLI_SNAPSHOT[0..4].try_into().unwrap())
+        as usize;
+    let mut vec = Vec::with_capacity(size);
+
+    unsafe {
+      vec.set_len(size);
+    }
+
+    lzzzz::lz4::decompress(&COMPRESSED_CLI_SNAPSHOT[4..], &mut vec).unwrap();
+
+    vec.into_boxed_slice()
+  },
+);
 
 pub fn deno_isolate_init() -> Snapshot {
   debug!("Deno isolate init with snapshots.");
-  let data = CLI_SNAPSHOT;
-  Snapshot::Static(data)
+  Snapshot::Static(&*CLI_SNAPSHOT)
 }
 
 #[cfg(test)]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -369,6 +369,9 @@ name = "cc"
 version = "1.0.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "22a9137b95ea06864e018375b72adfb7db6e6f68cfc8df5a04d00288050485ee"
+dependencies = [
+ "jobserver",
+]
 
 [[package]]
 name = "cfg-if"
@@ -744,6 +747,7 @@ dependencies = [
  "walkdir",
  "winapi 0.3.9",
  "winres",
+ "zstd",
 ]
 
 [[package]]
@@ -979,6 +983,7 @@ dependencies = [
  "hyper",
  "libc",
  "log",
+ "lzzzz",
  "netif",
  "nix",
  "notify",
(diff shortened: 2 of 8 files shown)


runtime/js.rs
```diff
@@ -1,14 +1,33 @@
...
 
-pub static CLI_SNAPSHOT: &[u8] =
-  include_bytes!(concat!(env!("OUT_DIR"), "/CLI_SNAPSHOT.bin"));
+pub static CLI_SNAPSHOT: Lazy<Box<[u8]>> = Lazy::new(
+  #[cold]
+  #[inline(never)]
+  || {
+    static COMPRESSED_CLI_SNAPSHOT: &[u8] =
+      include_bytes!(concat!(env!("OUT_DIR"), "/CLI_SNAPSHOT.bin"));
+
```
bartlomieju, line 12: I'm confused now, weren't we supposed to not compress CLI snapshot to avoid decompression penalty? Or is it so insignificant?

evanwashere: compression is practically free because of how fast lz4 is
you can see the difference between lz4 and uncompressed in first benchmark

ry:
1mb saved? Is it worth the complexity?

lucacasonato:
@ry 25% improvement in startup time!

piscisaureus:
I think this looks really good.

bartlomieju (approved):
LGTM

runtime/js.rs
```diff
@@ -1,14 +1,33 @@
...
+    let size =
+      u32::from_le_bytes(COMPRESSED_CLI_SNAPSHOT[0..4].try_into().unwrap())
+        as usize;
+    let mut vec = Vec::with_capacity(size);
+
+    unsafe {
+      vec.set_len(size);
+    }
+
+    lzzzz::lz4::decompress(&COMPRESSED_CLI_SNAPSHOT[4..], &mut vec).unwrap();
```
bnoordhuis, line 22: It's probably an academic concern but I think writing to uninitialized memory like this is technically undefined behavior. The blessed path is to use MaybeUninit: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#out-pointers

bnoordhuis (approved):
One question: why lzzzz instead of (lz4_flex)[https://crates.io/crates/lz4_flex]? The latter is the more popular crate and is pure Rust.

runtime/js.rs
```diff
@@ -1,14 +1,33 @@
...
+      include_bytes!(concat!(env!("OUT_DIR"), "/CLI_SNAPSHOT.bin"));
+
+    let size =
+      u32::from_le_bytes(COMPRESSED_CLI_SNAPSHOT[0..4].try_into().unwrap())
+        as usize;
+    let mut vec = Vec::with_capacity(size);
+
+    unsafe {
+      vec.set_len(size);
+    }
```
lucacasonato, line 20:
```suggestion
    let mut vec = vec![0; size];
```
Why not just this? I don't like unsafe :-)

lucacasonato: I would much prefer that tiny perf hit, than an `unsafe` block without a `// SAFETY:` comment. I am surprised the linter isn't having a mental breakdown because of that anyway.

runtime/js.rs
evanwashere: there is no undefined behavior in this case, we are just reusing previous allocation without zeroing it

andreubotella: I'd argue `MaybeUninit` is much clearer.

evanwashere:
> One question: why lzzzz instead of (lz4_flex)[https://crates.io/crates/lz4_flex]? The latter is the more popular crate and is pure Rust.

I initially used lz4_flex for proof of concept but switched to lzzzz because it has lz4hc and is slightly faster at decompressing cli snapshot

piscisaureus (approved):

runtime/js.rs
andreubotella: I've just checked, and `Vec::with_capacity` calls the allocator without zeroing the memory. Sure, the allocation might've actually been reused, or it might be a new allocation that actually has been zeroed, but even if we could guarantee that, the language treats that memory as uninitialized, which means that creating values from it, even `u8` values, is undefined behavior.

bartlomieju:
@piscisaureus I guess we'll wait with landing this PR until we upgrade `rusty_v8`

piscisaureus:
> @piscisaureus I guess we'll wait with landing this PR until we upgrade `rusty_v8`

There's no need; snapshots are already off in the current version of rusty_v8.

PR #13324, opened 2022-01-10
fix(coverage): don't type check

bartlomieju:
This commit changes "deno coverage" command not to type check.

Instead of relying on infrastructure for module loading in "deno run";
the code now directly reaches into cache for original and transpiled
sources. In case sources are not available the error is returned to the
user, suggesting to first run "deno test --coverage" command.

Fixes https://github.com/denoland/deno/issues/13161
Fixes https://github.com/denoland/deno/issues/13326

### DIFF
--- a/cli/tools/coverage.rs
+++ b/cli/tools/coverage.rs
@@ -1,7 +1,6 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
 use crate::colors;
-use crate::emit;
 use crate::flags::CoverageFlags;
 use crate::flags::Flags;
 use crate::fs_util::collect_files;
@@ -11,11 +10,12 @@ use crate::tools::fmt::format_json;
 
 use deno_ast::MediaType;
 use deno_ast::ModuleSpecifier;
+use deno_core::anyhow::anyhow;
+use deno_core::anyhow::Context;
 use deno_core::error::AnyError;
 use deno_core::serde_json;
 use deno_core::url::Url;
 use deno_core::LocalInspectorSession;
-use deno_runtime::permissions::Permissions;
 use regex::Regex;
 use serde::Deserialize;
 use serde::Serialize;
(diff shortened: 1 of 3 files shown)


cli/tools/coverage.rs
```diff
@@ -643,40 +643,65 @@ pub async fn cover_files(
...
-    let script_source = &module_source.code;
 
+    let maybe_file = if module_specifier.scheme() == "file" {
+      ps.file_fetcher.get_source(&module_specifier)
+    } else {
+      ps.file_fetcher
+        .fetch_cached(&module_specifier, 10)
+        .with_context(|| {
+          format!("Failed to fetch \"{}\" from cache.", module_specifier)
+        })?
```
bartlomieju, line 654: This is fishy to me; we shouldn't collect coverage data for remote scripts. I need to double check it.

bartlomieju:
I verified that this PR works with `deno_std`.

cli/tools/coverage.rs
```diff
@@ -643,40 +643,66 @@ pub async fn cover_files(
...
+
+    // Check if file was transpiled
+    let transpiled_source = match file.media_type {
+      MediaType::JavaScript
+      | MediaType::Unknown
+      | MediaType::Cjs
+      | MediaType::Mjs
+      | MediaType::Json => file.source.as_ref().clone(),
+      MediaType::Dts => "".to_string(),
+      _ => {
```
dsherret, line 675: Perhaps we should remove this `_ =>` and instead have all the possibilities here? I kind of feel like we need a simpler enum that we can easily get from a `MediaType` that only has stuff like `JavaScript, `Typescript`, `Declaration`, etc in it.

cli/tools/coverage.rs
```diff
@@ -643,40 +643,66 @@ pub async fn cover_files(
...
+    };
+
+    // Check if file was transpiled
+    let transpiled_source = match file.media_type {
+      MediaType::JavaScript
+      | MediaType::Unknown
+      | MediaType::Cjs
+      | MediaType::Mjs
+      | MediaType::Json => file.source.as_ref().clone(),
+      MediaType::Dts => "".to_string(),
```
dsherret, line 674: What about `Dmts` and `Dcts`?

cli/tools/coverage.rs
```diff
@@ -643,40 +643,66 @@ pub async fn cover_files(
...
+    let maybe_file = if module_specifier.scheme() == "file" {
+      ps.file_fetcher.get_source(&module_specifier)
+    } else {
+      ps.file_fetcher
+        .fetch_cached(&module_specifier, 10)
+        .with_context(|| {
+          format!("Failed to fetch \"{}\" from cache.", module_specifier)
+        })?
+    };
+    let file = if let Some(f) = maybe_file {
```
dsherret, line 656: `maybe_file.ok_or_else(|| ...)?;`?

dsherret (commented):
Just needs the media type fix.

cli/tools/coverage.rs
bartlomieju: Good point. I've taken this code directly from `cli/proc_state.rs` I guess it should be updated there as well

cli/tools/coverage.rs
bartlomieju: See comment above

dsherret (approved):
LGTM

PR #13325, opened 2022-01-10
fix: `deno install` with `--allow-all` should output shim with `--allow-all` instead of each permission

### LINKED ISSUE #11819: 🐛 `deno` v1.13 breaks shims using `--allow-plugin`
With the change in v1.13 from `--allow-plugin` to `--allow-ffi` (PR #11152), prior shims using `--allow-plugin` are now broken...

```
error: Found argument '--allow-plugin' which wasn't expected, or isn't valid in this context
        Did you mean --allow-run?

USAGE:
    deno run <SCRIPT_ARG>... --allow-env=<allow-env> --allow-net=<allow-net> --allow-read=<allow-read> --allow-run=<allow-run> --allow-write=<allow-write>

```

IMO, `--allow-plugin` should have been deprecated but allowed until a later date (maybe v2). Changing the command line API like this can suddenly break working setups (as it did mine). To me, this seems to be the definition of a semver API change requiring a major version update.

dsherret:
This changes doing `deno install` with `-A`/`--allow-all` to output a shim with `--allow-all` only instead of listing every permission flag.

We do not need to be concerned about potential future permission flags because `--allow-all` has a permission for `--allow-run` which allows for anything to occur already.

Closes #11819

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -214,6 +214,7 @@ pub struct Flags {
   pub argv: Vec<String>,
   pub subcommand: DenoSubcommand,
 
+  pub allow_all: bool,
   pub allow_env: Option<Vec<String>>,
   pub allow_hrtime: bool,
   pub allow_net: Option<Vec<String>>,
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM

rivy:
Thanks!

PR #13332, opened 2022-01-11
refactor: move transpiling to deno_ast

dsherret:
1. Removes the `ast` folder and moves some of it into emit.rs
1. Uses `deno_ast` for transpiling.


Cargo.lock
```diff
@@ -1035,8 +1037,8 @@ dependencies = [
  "rustls-native-certs",
  "rustls-pemfile",
  "serde",
- "webpki",
- "webpki-roots",
+ "webpki 0.22.0",
+ "webpki-roots 0.22.1",
```
bartlomieju, line 1041: This is fishy... we shouldn't introduce multiple versions of these crates

dsherret: I ran `cargo update` and that fixed this issue.

bartlomieju (approved):

PR #13340, opened 2022-01-11
feat(ext/ffi): UnsafeFnPointer API

### LINKED ISSUE #13336: Support calling Function Pointers in FFI
### Why

Some native libraries return function pointers to expose further API. Such as GLFW exposes a `glfwGetProcAddress` function which is used to get pointer to the OpenGL API function. I'm currently using a helper library written in C to call function pointers in [my project that uses GLFW](https://github.com/DjDeveloperr/deno_gl). I'm sure there are much more use-cases than just this, like calling temporary callbacks returned from native land.

### Suggestion

Introduce a new API to wrap a function pointer + a function definition into a JS function, such as:

```ts
// on `DynamicLibrary`
fnptr<Fn extends Deno.ForeignFunction>(ptr: Deno.UnsafePointer, def: Fn): StaticForeignFunction<Fn>;
```

(I'm not sure about name of the method)

which can be used something like this:

```ts
const fn = lib.fnptr(ptr, {
  parameters: [...],
  result: "...",
});

fn(...);
```

DjDeveloperr:
closes #13336

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -250,6 +250,24 @@ declare namespace Deno {
     copyInto(destination: TypedArray, offset?: number): void;
   }
 
+  /**
+   * **UNSTABLE**: Unsafe and new API, beware!
+   * 
+   * An unsafe pointer to a function, for calling functions that are not
+   * present as symbols.
+   */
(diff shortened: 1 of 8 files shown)


bartlomieju (approved):
LGTM, good addition 👍

PR #13341, opened 2022-01-11
fix(ext/web): handle no arguments in atob

lucacasonato:
WPT tests are pending on getting html idlharness tests running.

Closes #13341


crowlKats (approved):
LGTM

Issue #13022, opened 2021-12-08
lsp: benchmark individual diagnostic publishes


strager:
Deno's LSP server has an artificial 200 ms debounce/delay:
https://github.com/denoland/deno/blob/b51b0c834b596383084d4a9991695759af9d27ad/cli/lsp/diagnostics.rs#L139-L142

This hard-coded delay makes Deno look bad when compared to other LSP servers: https://quick-lint-js.com/benchmarks/

I think the delay in Deno's LSP server should be user-configurable (e.g. disable-able or a custom delay).

kitsonk:
I highly doubt it would make a difference in that benchmark.  I personally think the benchmark is flawed, it doesn't test what people in the real world experience. It essentially measures startup time to first diagnostic publish.

A benchmark like the [big_file_edits Language Service benchmark](https://deno.land/benchmarks) is a lot more realistic, which is multiple edits to a large file through the complete.  Initial diagnostic publish is a consideration, but the responsiveness to changes and the updates to the diagnostics are more important.

Also, comparing the Deno language server to quick-lint-js will always be an apples to orange comparison.  The Deno language server is not a standalone linting language server and there is no way to do that.  When starting the language server, tsc (TypeScript) is started and spun up and diagnostics are requested.  That debounce delay doesn't even come into play.

Users twiddling with low level attributes like that are only likely to make people experience a performance degradation and blame the maintainers for their poor performance.  That delay was specifically engineered to improve the overall performance of the language server.

Specifically the debounce delay was introduced to work around the blocking nature of tsc and made a dramatic improvement to the performance of the language server:

<img width="717" alt="Benchmarks__3889___5338_" src="https://user-images.githubusercontent.com/1282577/145171611-b7edcfbc-97fa-4830-a8ef-d3fb74f22afb.png">

And this delay is only a delay affecting TypeScript diagnostics, not lint diagnostics.  We had a situation where we did block on publishing all diagnostics, but we changed that and diagnostics get published when they are available, but the startup time to initial publish of diagnostics is still somewhat constrained by the startup of tsc.  We could look at doing an initial publish of lint diagnostics just to game the quick-lint-js benchmark, but 🤷.

dsherret:
[...]

PR #13365, opened 2022-01-13
Add LSP benchmark mimicking the one on quick-lint-js

ry:
See #13022

https://github.com/quick-lint/quick-lint-js/blob/35207e6616267c6c81be63f47ce97ec2452d60df/benchmark/benchmark-lsp/lsp-benchmarks.cpp#L223-L268

Currently I'm getting something like 128ms per iteration (which seems slower than the 200ms expected) so not sure if there is something broken.
```
# cargo bench  -p deno --bench lsp_bench_standalone
   Compiling deno v1.17.2 (/Users/ry/src/deno/cli)
    Finished bench [optimized] target(s) in 6.83s
     Running unittests (target/release/deps/lsp_bench_standalone-c6010a0d3f29fcda)

running 1 test
test incremental_change_wait ... bench: 128,334,504 ns/iter (+/- 7,344,853)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured
```


cli/bench/lsp_bench_standalone.rs
```diff
@@ -0,0 +1,75 @@
...
+              },
+              "contentChanges":[
+                {"text": text, "range":{"start":{"line":506,"character":39},"end":{"line":506,"character":45}}},
+                {"text": text, "range":{"start":{"line":507,"character":8},"end":{"line":507,"character":14}}},
+                {"text": text, "range":{"start":{"line":509,"character":10},"end":{"line":509,"character":16}}}
+              ]
+          })
+      ).unwrap();
+      let (method, maybe_diag): (String, Option<Value>) =
+        client.read_notification().unwrap();
```
kitsonk, line 60: We have up to 3 publishes that occur when a change is made:

https://github.com/denoland/deno/blob/eda6e58520276786bd87e411d0284eb56d9686a6/cli/lsp/diagnostics.rs#L605-L612

When each of the three categories finish being calculated, they publish.  This code is only measuring from change to first publish, not full publish.  I am not sure what you are wanting here.

I realised when looking at the code again, we are debouncing "all" the diagnostics, when in fact we only care about debouncing the "typescript" ones because of their blocking nature for other API calls, which degraded performance:

https://github.com/denoland/deno/blob/eda6e58520276786bd87e411d0284eb56d9686a6/cli/lsp/diagnostics.rs#L172-L184

We could/should break that out, so that the "lint" and "deno (dependency)" diagnostics are calculated independently and not debounced.  There might be a little bit work to do, because we treat the dirtyness of the diagnostics per document, and what we are saying is that lint and deno diagnostics are one version of dirty and typescript is another version of dirty.

kitsonk (commented):
I'm not sure why you are seeing 128ms instead of 200ms minimum, looking at the code is from the time the change is processed it should be at least 200ms before we start an `update_diagnostics`...

cli/bench/lsp_bench_standalone.rs
ry: Thanks, now I understand why quick-lint has `diagnosticsMessagesToIgnore` set to 2 for Deno.

https://github.com/quick-lint/quick-lint-js/blob/35207e6616267c6c81be63f47ce97ec2452d60df/benchmark/benchmark-lsp/benchmark-config.json#L63-L77

kitsonk:
> This still isn't right because I'm using `diagnosticsMessagesToIgnore=1` and the numbers I'm getting are 287ms instead of 200ms. But I think it's getting closer to quick-lint-js's benchmark.
> 
> ```
> bench: 287,436,395 ns/iter (+/- 12,993,664)
> ```

Well, it will be 200ms + tsc diagnostic request time + maybe blocking time if there is some other request in tsc.  The whole purpose of the debounce was to allow other "high priority" tsc requests to get in front of the diagnostic request, but requests are processed in the order they are received:

https://github.com/denoland/deno/blob/eda6e58520276786bd87e411d0284eb56d9686a6/cli/lsp/tsc.rs#L100-L111

Given the test case though I can't specifically think of another request that might be going on, but it could be.

I did check on a reasonably sized project and I am getting:

|measurement|time|
|-|-|
| update_diagnostics_deps | 10ms |
| update_diagnostics_lint | 10ms |
| update_diagnostics_ts | 69ms |

So getting 287ms for a tsc diagnostics publish on a large file is not unreasonable or unexpected.

The first one being < 200ms I believe is actually another side effect all together of a design limitation of the test harness. We tried and tried to find a non-blocking way to "poll" for messages, but haven't, so any unread unsolicited messages will be read and returned, but if you try to poll for a message that isn't there, it will just hang.  So you are likely reading something that was previously sent.

ry:
I've updated the benchmark to wait until we get the first "deno-lint" diagnostic (as described in the source field of the diagnostics. After this lands I'll send a PR to quick-lint-js to change their benchmark.

It's surprising but each iteration is still quite a bit longer than 200ms.

```
# cargo bench --bench lsp_bench_standalone
   Compiling deno v1.17.3 (/Users/ry/src/deno/cli)
    Finished bench [optimized] target(s) in 7.03s
     Running unittests (target/release/deps/lsp_bench_standalone-50969af938e69a9b)

running 1 test
test incremental_change_wait ... bench: 288,701,433 ns/iter (+/- 6,529,414)

```

kitsonk:
> It's surprising but each iteration is still quite a bit longer than 200ms.

There are other factors at play that try to keep it from becoming a "blocking" thing, but as I mentioned above, I think it can be refactored so that there is no debouncing of lint and "deno" diagnostics, we only debounce the TypeScript ones, and then we would get a "real" picture of how long it takes to issue lint diagnostics as well as provide a more responsive experience for users.

cli/bench/lsp_bench_standalone.rs
```diff
@@ -0,0 +1,95 @@
...
+fn wait_for_deno_lint_diagnostic(
+  document_version: u64,
+  client: &mut LspClient,
+) {
+  loop {
+    let (method, maybe_diag): (String, Option<Value>) =
+      client.read_notification().unwrap();
+    if method == "textDocument/publishDiagnostics" {
+      let d = maybe_diag.unwrap();
+      //println!("diag {:?}", d);
```
kitsonk, line 76: nit: commented code left in

cli/bench/lsp_bench_standalone.rs
```diff
@@ -0,0 +1,95 @@
...
+      }),
+    )
+    .unwrap();
+  let (method, _maybe_diag): (String, Option<Value>) =
+    client.read_notification().unwrap();
+  assert_eq!(method, "textDocument/publishDiagnostics");
+
+  let mut document_version: u64 = 0;
+  bench.iter(|| {
+    // println!("document_version {}", document_version);
```
kitsonk, line 43: nit: commented code left in

kitsonk (approved):
LGTM noting a couple nits

PR #13374, opened 2022-01-14
refactor(installer): refactor installer code to be more testable

dsherret:
This extracts out a function from the installer code and then uses that more in the tests.

It also runs these tests 4s faster on my computer.


bartlomieju (approved):
LGTM, fewer lines of code is always better.

PR #13375, opened 2022-01-14
feat(streams): reject pending reads when releasing reader

### LINKED ISSUE #13369: Streams: reject pending reads when releasing reader
Previously, calling `ReadableStreamDefaultReader.releaseLock()` or `ReadableStreamBYOBReader.releaseLock()` while there are pending `read()` requests would throw a `TypeError`.

The specification has been changed to allow this case, and to reject such pending `read()` requests instead.

Standard change: https://github.com/whatwg/streams/pull/1168
Tests: https://github.com/web-platform-tests/wpt/pull/32072

crowlKats:
Closes #13369

### DIFF
--- a/ext/web/06_streams.js
+++ b/ext/web/06_streams.js
@@ -279,6 +279,7 @@
   const _pullAlgorithm = Symbol("[[pullAlgorithm]]");
   const _pulling = Symbol("[[pulling]]");
   const _pullSteps = Symbol("[[PullSteps]]");
+  const _releaseSteps = Symbol("[[ReleaseSteps]]");
   const _queue = Symbol("[[queue]]");
   const _queueTotalSize = Symbol("[[queueTotalSize]]");
   const _readable = Symbol("[[readable]]");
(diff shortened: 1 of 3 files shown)


lucacasonato (approved):
LGTM

ext/web/06_streams.js
```diff
@@ -1708,6 +1837,18 @@
...
   }
 
+  /**
+   * @template R
+   * @param {ReadableStreamDefaultReader<R>} reader
+   */
+  function readableStreamDefaultReaderRelease(reader) {
+    readableStreamReaderGenericRelease(reader);
+    const e = new TypeError(
+      "There are pending read requests, so the reader cannot be released.",
```
MattiasBuelens, line 1847: This error message is no longer accurate. The reader can (and will) be released, but any pending read requests will be *rejected* with this `TypeError`.

Same applies to `readableStreamBYOBReaderRelease`.

crowlKats: Good point, i should have put more thought into that. thanks for pointing it out, will fix ASAP

PR #13376, opened 2022-01-14
fix(streams): update TypeError message for pending reads when releasing reader

crowlKats:
Follow up to #13375, as per https://github.com/denoland/deno/pull/13375#discussion_r784998827

### DIFF
--- a/ext/web/06_streams.js
+++ b/ext/web/06_streams.js
@@ -1286,9 +1286,7 @@
    */
   function readableStreamBYOBReaderRelease(reader) {
     readableStreamReaderGenericRelease(reader);
-    const e = new TypeError(
-      "There are pending read requests, so the reader cannot be released.",
-    );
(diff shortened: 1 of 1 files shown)


ext/web/06_streams.js
```diff
@@ -1843,9 +1841,7 @@
    */
   function readableStreamDefaultReaderRelease(reader) {
     readableStreamReaderGenericRelease(reader);
-    const e = new TypeError(
-      "There are pending read requests, so the reader cannot be released.",
-    );
+    const e = new TypeError("The Reader was released.");
```
lucacasonato, line 1844:
```suggestion
    const e = new TypeError("The reader was released.");
```

lucacasonato (commented):
Thanks for pointing this out @MattiasBuelens.

lucacasonato (approved):

PR #13377, opened 2022-01-14
chore: upgrade to rust 1.58

dsherret:


core/error.rs
```diff
@@ -372,6 +372,7 @@ impl ErrWithV8Handle {
   }
 }
 
+#[allow(clippy::non_send_fields_in_send_ty)]
```
dsherret, line 375: These are suspect. Probably ok?

littledivy: Yeah, It's probably going to be allowed by default https://github.com/rust-lang/rust-clippy/issues/8045

littledivy (approved):

bnoordhuis (approved):
Thanks, David. Glad those to_string() calls are flagged now, I always get facial tics when I see them.

Issue #13357, opened 2022-01-12
ext/console crashes when deleting `globalThis`


lucacasonato:
`console.log(delete globalThis.globalThis);` crashes Deno.

PR #13387, opened 2022-01-16
fix(ext/console): don't depend on globalThis present

bartlomieju:
Fixes https://github.com/denoland/deno/issues/13357

### DIFF
--- a/ext/console/02_console.js
+++ b/ext/console/02_console.js
@@ -1650,7 +1650,7 @@
   }
 
   function inspectArgs(args, inspectOptions = {}) {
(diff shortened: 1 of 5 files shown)


ext/console/02_console.js
```diff
@@ -1650,7 +1650,7 @@
   }
 
   function inspectArgs(args, inspectOptions = {}) {
-    const noColor = globalThis.Deno?.noColor ?? true;
```
bartlomieju, line 1653: I'm don't understand why was this defaulting to `true` while all other coercions defaulted to false.

dsherret (approved):
LGTM

Issue #13358, opened 2022-01-12
Deno crashes when deleting `globalThis.window`


lucacasonato:
```js
console.log(delete globalThis.window);
```

```
true
error: Uncaught ReferenceError: window is not defined
    at [deno:cli/main.rs:1184:31]:1:1
```

PR #13392, opened 2022-01-17
fix(runtime): don't crash when window is deleted

bartlomieju:
This commit fixes an error when user deletes "window" global JS
variable. Instead of relying on "window" or "globalThis" to dispatch
"load" and "unload" events, we are default to global scope of the
worker.

Fixes https://github.com/denoland/deno/issues/13358

### DIFF
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -2479,3 +2479,8 @@ itest!(import_assertions_type_check {
   output: "import_assertions/type_check.out",
   exit_code: 1,
 });
+
+itest!(delete_window {
+  args: "run delete_window.js",
+  output_str: Some("true\n"),
+});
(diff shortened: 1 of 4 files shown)


andreubotella:
This still crashes when `globalThis` is deleted.

bartlomieju:
> This still crashes when `globalThis` is deleted, even when incorporating #13387

I'm aware of that, preventing such crash will require bigger refactor that I'll do when I get more time.

bartlomieju:
As @andreubotella this PR will start crashing after https://github.com/denoland/deno/pull/13387 is merged. The proper fix is to store a `v8::Global<v8::Value>` on `MainWorker`, that will point to `globalThis`. This global should be created just after bootstrapping a worker. Then when `load` and `unload` events are dispatched they should call `dispatchEvent` method directly on a `v8::Local<v8::Value>` obtained from the stored global, instead of using `self.js_runtime.execute_script()`.

dsherret (approved):
LGTM

bartlomieju:
@andreubotella PTAL

andreubotella (approved):
LGTM

PR #13395, opened 2022-01-17
Don't hide LSP error messages during tests

ry:

### DIFF
--- a/test_util/src/lsp.rs
+++ b/test_util/src/lsp.rs
@@ -165,7 +165,6 @@ impl LspClient {
       .arg("lsp")
       .stdin(Stdio::piped())
       .stdout(Stdio::piped())
-      .stderr(Stdio::null())
       .spawn()?;
 
     let stdout = child.stdout.take().unwrap();


bartlomieju (approved):

kitsonk:
Are we really sure we want to do this... All "diagnostic" output of a test goes to stderr, including error messages.  This means that every lsp test will output some 10 lines or so of "setup" information, even if there isn't an error.

Wouldn't a better solution be to have a flag that goes to the `LSPClient` that enables or disables it?

bartlomieju:
> Are we really sure we want to do this... All "diagnostic" output of a test goes to stderr, including error messages. This means that every lsp test will output some 10 lines or so of "setup" information, even if there isn't an error.
> 
> Wouldn't a better solution be to have a flag that goes to the `LSPClient` that enables or disables it?

That's a valid point, I just run the tests locally and indeed there's a lot of output being printed. To be honest it's quite surprising to me, I though that this output would be caught by the test harness and only printed on test failures.

kitsonk:
> To be honest it's quite surprising to me, I though that this output would be caught by the test harness and only printed on test failures.

That would make more sense, but personally I don't know how to accomplish that.  The `LSPClient` would hold onto the `stderr` and make it available, but I don't know how to automatically expose it on an assertion failure.  I suspect some sort of "test macro" could work, where if it panics, it dumps the `stderr`.

That being said the "problem" was that there were errors in the logs that were not impacting the outcomes of the tests, which meant in theory the tests weren't actually testing what we thought they were testing.  That wasn't the case in the end, but, the point is valid... always swallowing the `stderr` isn't a good idea, sometimes it should be inspected.

ry:
Closing this - I think kitson makes a good point.

PR #13403, opened 2022-01-17
refactor(lsp): remove `performance` from `StateSnapshot`

dsherret:
`Performance` was hoping along for the ride on `StateSnapshot`, but it is not really part of the concept of a "snapshot". Instead it should be a constructor dependency on these other structs.

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -5,7 +5,9 @@ use super::client::Client;
 use super::documents;
 use super::documents::Documents;
 use super::language_server;
+use super::performance::Performance;
 use super::tsc;
+use super::tsc::TsServer;
 
 use crate::diagnostics;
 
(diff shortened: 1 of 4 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -127,12 +146,13 @@ impl DiagnosticsServer {
   pub(crate) fn start(
     &mut self,
     language_server: Arc<Mutex<language_server::Inner>>,
-    client: Client,
-    ts_server: Arc<tsc::TsServer>,
```
dsherret, line 131: I am hoping to get rid of how `language_server` is passed around to functions like this in the future and this PR is a stepping stone for that.

cli/lsp/performance.rs
```diff
@@ -72,11 +72,11 @@ impl From<PerformanceMark> for PerformanceMeasure {
...
 /// The structure will limit the size of measurements to the most recent 1000,
 /// and will roll off when that limit is reached.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Performance {
-  counts: Arc<Mutex<HashMap<String, u32>>>,
+  counts: Mutex<HashMap<String, u32>>,
   max_size: usize,
-  measures: Arc<Mutex<VecDeque<PerformanceMeasure>>>,
+  measures: Mutex<VecDeque<PerformanceMeasure>>,
```
dsherret, line 79: I moved the `Arc` to `Arc<Performance>` just so its clear that a single object is being passed around.

kitsonk (approved):
LGTM, nice refactor!

bartlomieju (approved):
LGTM too, this makes a lot more sense to me 👍

Issue #11307, opened 2021-07-06
Change language server to use concurrent cancellable reads and exclusive writes


dsherret:
I think the change in #9271 to an async mutex at a high level was very good to do in that it quickly made the code less error prone and accessible for other developers working on it (harder for devs to make mistakes, removed race conditions, and eliminated deadlocks), however it does have some performance drawbacks since only one task can run through the language server code at one a time.

To expand on the work of #9271, we could split up the code for concurrent read actions and exclusive write actions (phase 1) and then introduce cancellable read actions (phase 2). I believe this would improve the performance of the language server and not introduce too much complexity since the locking still happens at a high level.

Write actions would be anything that updates the user state (input state), so file text changes or configuration updates. These would require exclusive access to the document cache. On write, any read actions would be waited on before proceeding (and maybe cancelled... not sure). These write actions should be very quick and only update the necessary state. Ideally we should be lazy about computing data.

Read actions would be concurrent with other read actions and be cancellable (read actions would be actions such as "find references", "go to definition", etc.). Read actions should still be able to lazily compute and cache data (mutate), but the result of these actions should be pure in that if multiple tasks compute the same value at the same time they would have the same result (ideally though, both tasks could wait on the result computed by a single task... this could be achieved without much added complexity by pushing down the complexity into a reusable general synchronization struct... maybe this will be phase 3 and only if necessary).

To implement phase 1, we could use a RwLock and then ensure the code is structured in such a way that this pattern is very clear and hard for a developer to make a concurrency mistake.

For phase 2, we could use tokio [cancellation tokens](https://docs.rs/tokio-util/0.6.7/tokio_util/sync/struct.CancellationToken.html) more liberally in the read actions. This would also have the added benefit of allowing us to also more easily support cancellable LSP commands ([cancelRequest](https://microsoft.github.io/language-server-protocol/specification#cancelRequest)) in the future. It might be worth considering cancelling all read actions on a write, but I'm not sure... perhaps just listening for an LSP cancelRequest command is sufficient. We can see once we go to implement it.

cc @kitsonk @bnoordhuis @lucacasonato

kitsonk:
> To implement phase 1, we could use a RwLock and then ensure the code is structured in such a way that this pattern is very clear and hard for a developer to make a concurrency mistake.

@ry was quite opposed to this when I was adding them previously, as he felt it complicated things.

[...]

PR #13414, opened 2022-01-18
refactor(lsp): store assets behind a mutex

dsherret:
This PR stores all assests behind a mutex in order to make more language server messages not require mutability.

This is possible to do because the assets are static and it is ok if two threads race to retrieve a non-cached asset from tsc.

This is slowly moving towards #11307.

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -26,6 +26,7 @@ use deno_core::error::custom_error;
 use deno_core::error::AnyError;
 use deno_core::located_script_name;
 use deno_core::op_sync;
+use deno_core::parking_lot::Mutex;
 use deno_core::resolve_url;
 use deno_core::serde::de;
 use deno_core::serde::Deserialize;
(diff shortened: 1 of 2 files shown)


cli/lsp/tsc.rs
```diff
@@ -187,48 +189,115 @@ impl AssetDocument {
...
-    self.0.insert(k, v)
+    self.assets.lock().get(k).cloned()
+  }
+
+  pub(crate) async fn get(
+    &self,
+    specifier: &ModuleSpecifier,
+    // todo(dsherret): this shouldn't be a parameter, but instead retrieved via
+    // a constructor dependency
+    get_snapshot: impl Fn() -> LspResult<Arc<StateSnapshot>>,
```
dsherret, line 269: In a future PR, I'm hoping to get rid of this parameter and not have `language_server::Inner` passed around in other functions.

cli/lsp/tsc.rs
```diff
@@ -187,48 +189,115 @@ impl AssetDocument {
   }
 }
 
+fn new_assets_map(
+) -> Arc<Mutex<HashMap<ModuleSpecifier, Option<AssetDocument>>>> {
```
bartlomieju, line 193: That's a mouthful, consider adding `type AssetsMap = HashMap<ModuleSpecifier, Option<AssetDocument>>`

kitsonk (approved):
LGTM

bartlomieju (approved):
LGTM too

PR #13421, opened 2022-01-19
fix(ext/crypto): validate maskGenAlgorithm asn1 in importKey

### LINKED ISSUE #13032: Web Crypto: missing check in `importKey`
There is a missing check in `crypto.subtle.importKey` when using the `spki` or `pkcs8` format for the `RSA-PSS` and `RSA-OAEP` algorithms. The missing check is this one:

> If the parameters field of the maskGenAlgorithm field of params is not an instance of the HashAlgorithm ASN.1 type that is identical in content to the hashAlgorithm field of params, throw a NotSupportedError.

See https://w3c.github.io/webcrypto/#rsa-pss-operations and https://w3c.github.io/webcrypto/#rsa-oaep-operations.

littledivy:
Fixes #13032

### DIFF
--- a/ext/crypto/import_key.rs
+++ b/ext/crypto/import_key.rs
@@ -7,6 +7,7 @@ use elliptic_curve::sec1::ToEncodedPoint;
 use p256::pkcs8::FromPrivateKey;
 use p256::pkcs8::ToPrivateKey;
 use rsa::pkcs1::UIntBytes;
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
LGTM but I strongly suggest factoring it out into a helper function.

Rule of Three is a good rule of thumb: if you have three or more instances, it's time to refactor.

Also, this could use a test.

PR #13426, opened 2022-01-19
refactor(lsp): reduce data stored in `StateSnapshot`

dsherret:
This PR reduces the amount of data stored in `StateSnapshot` to only what is necessary for TSC and diagnostics.

This should be slightly more efficient and slowly works towards not passing around `language_server::Inner` and #11307.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -74,15 +74,12 @@ const CACHE_PATH: &str = "deps";
 #[derive(Debug, Clone)]
 pub struct LanguageServer(Arc<tokio::sync::Mutex<Inner>>);
 
+/// Snapshot of the state used by TSC.
 #[derive(Debug, Default)]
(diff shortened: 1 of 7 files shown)


cli/lsp/language_server.rs
```diff
@@ -97,7 +94,7 @@ pub(crate) struct Inner {
   diagnostics_server: diagnostics::DiagnosticsServer,
   /// The collection of documents that the server is currently handling, either
   /// on disk or "open" within the client.
-  documents: Documents,
+  pub(crate) documents: Documents,
```
dsherret, line 97: These `pub(crate)`'s are temporary. They'll be removed in a future PR once nothing depends on this struct.

kitsonk (approved):
LGTM, feels good where this is headed.

bartlomieju (approved):
LGTM too

PR #13427, opened 2022-01-19
perf(lsp): independent diagnostic source publishes

### LINKED ISSUE #13410: Improve time to lint diagnostics in LSP
This can be measured using the following benchmark:
```
# cargo bench --bench lsp_bench_standalone
   Compiling deno v1.17.3 (/Users/ry/src/deno/cli)
    Finished bench [optimized] target(s) in 7.03s
     Running unittests (target/release/deps/lsp_bench_standalone-50969af938e69a9b)

running 1 test
test incremental_change_wait ... bench: 288,701,433 ns/iter (+/- 6,529,414)
```

Currently it takes well over 200ms to receive lint diagnostics due to the debouncing mechanism.

dsherret:
```
> cargo bench --bench lsp_bench_standalone
running 1 test
test incremental_change_wait ... bench:   8,379,260 ns/iter (+/- 590,305)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured
```

This is 288ms to 8ms. The difference is so subtle in the editor though.

The independent lint diagnostic publishes shouldn't be hung up on type checking anymore at least.

Closes #13410


dsherret:
Flaky test most likely caused by this PR (extra diagnostic when none was expected):

```
---- integration::lsp::lsp_diagnostics_refresh_dependents stdout ----
thread 'integration::lsp::lsp_diagnostics_refresh_dependents' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
<[
<    Diagnostic {
<        range: Range {
<            start: Position {
<                line: 0,
<                character: 12,
<            },
<            end: Position {
<                line: 0,
<                character: 13,
<            },
<        },
<        severity: Some(
<            Error,
<        ),
<        code: Some(
<            Number(
<                2305,
<            ),
<        ),
<        code_description: None,
<        source: Some(
<            "deno-ts",
<        ),
<        message: "Module '\"./file_01.ts\"' has no exported member 'b'.",
<        related_information: None,
<        tags: None,
<        data: None,
<    },
<]
>[]

', cli/tests/integration/lsp_tests.rs:3495:7
```

https://github.com/denoland/deno/runs/4875029450?check_suite_focus=true

kitsonk:
> Flaky test most likely caused by this PR (extra diagnostic when none was expected):

Yeah, what is happening is sometimes now the diagnostics are coming out of "order", because now the lint category is decoupled from the typescript ones, the "last 3" aren't always correct...

Given the change, I would say it would safe to only look at the "last" diagnostics and ensure they are empty. (`i == queue_len - 1`).

cli/lsp/diagnostics.rs
```diff
@@ -203,15 +181,35 @@ impl DiagnosticsServer {
...
+              let previous_handle = handle.take();
+
+              handle = Some(tokio::spawn(async move {
+                if let Some(previous_handle) = previous_handle {
+                  // wait on the previous run to complete in order to prevent
+                  // multiple threads queueing up a lot of tsc requests
+                  tokio::select! {
+                    _ = token.cancelled() => { return; }
+                    _ = previous_handle => {}
+                  };
```
dsherret, line 198: This is an issue. It should always wait on the previous handle to complete and not cancel waiting.

cli/tests/integration/lsp_tests.rs
```diff
@@ -3480,6 +3480,8 @@ fn lsp_diagnostics_refresh_dependents() {
     .write_request::<_, _, Value>("shutdown", json!(null))
     .unwrap();
   client.write_notification("exit", json!(null)).unwrap();
+  // wait for the published diagnostics of both requests
+  std::thread::sleep(std::time::Duration::from_millis(250));
```
dsherret, line 3484: >> Flaky test most likely caused by this PR (extra diagnostic when none was expected):
>
> Yeah, what is happening is sometimes now the diagnostics are coming out of "order", because now the lint category is decoupled from the typescript ones, the "last 3" aren't always correct...
>
> Given the change, I would say it would safe to only look at the "last" diagnostics and ensure they are empty. (i == queue_len - 1).

@kitsonk this PR actually doesn't decouple the diagnostics. It just gets the lint and module diagnostics publishing right away.

I looked at the test code a bit more and I think this sleep should stop the test from being flaky. I'm not proud of it though 🙈. Let me know if you have another suggestion.

dsherret:
Conversation continued here: https://github.com/denoland/deno/pull/13427/files#r788264675

kitsonk:
We should improve the design, instead of just "gaming" the test. 😞  If we aren't improving the design in line with the comments in #13410 then I am not supportive of this.

dsherret:
It should improve the responsiveness though (well, I mean 200ms to 8ms is not really an improvement practically in an lsp). It just doesn’t do it all independently at the moment so if there is a long type check after denouncing then it could delay lint diagnostics still. I guess I could spend a bit more time making them all completely independent as it wouldn’t be that much extra effort.

cli/tests/integration/lsp_tests.rs
```diff
@@ -3128,7 +3236,7 @@ fn lsp_cache_location() {
     load_fixture("did_open_params_import_hover.json"),
   );
   let diagnostics = diagnostics.into_iter().flat_map(|x| x.diagnostics);
-  assert_eq!(diagnostics.count(), 14);
+  assert_eq!(diagnostics.count(), 7);
```
dsherret, line 3239: I think this should just be 7. Here was the previous 14:

```
---
file:///a/file.ts
Version: Some(1)
Count: 0
file:///a/file.ts
Version: Some(1)
Count: 7
  Range { start: Position { line: 5, character: 19 }, end: Position { line: 5, character: 132 } }
  Uncached data URL.
  --
  Range { start: Position { line: 4, character: 19 }, end: Position { line: 4, character: 57 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/subdir/mod1.ts".
  --
  Range { start: Position { line: 3, character: 19 }, end: Position { line: 3, character: 67 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/subdir/type_reference.js".
  --
  Range { start: Position { line: 2, character: 19 }, end: Position { line: 2, character: 66 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/type_definitions/foo.js".
  --
  Range { start: Position { line: 1, character: 16 }, end: Position { line: 1, character: 63 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/type_definitions/foo.d.ts".
  --
  Range { start: Position { line: 0, character: 19 }, end: Position { line: 0, character: 62 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/xTypeScriptTypes.js".
  --
  Range { start: Position { line: 7, character: 19 }, end: Position { line: 7, character: 53 } }
  Uncached or missing remote URL: "http://localhost:4545/x/a/mod.ts".
  --
file:///a/file.ts
Version: Some(1)
Count: 7
  Range { start: Position { line: 5, character: 19 }, end: Position { line: 5, character: 132 } }
  Uncached data URL.
  --
  Range { start: Position { line: 4, character: 19 }, end: Position { line: 4, character: 57 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/subdir/mod1.ts".
  --
  Range { start: Position { line: 3, character: 19 }, end: Position { line: 3, character: 67 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/subdir/type_reference.js".
  --
  Range { start: Position { line: 2, character: 19 }, end: Position { line: 2, character: 66 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/type_definitions/foo.js".
  --
  Range { start: Position { line: 1, character: 16 }, end: Position { line: 1, character: 63 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/type_definitions/foo.d.ts".
  --
  Range { start: Position { line: 0, character: 19 }, end: Position { line: 0, character: 62 } }
  Uncached or missing remote URL: "http://127.0.0.1:4545/xTypeScriptTypes.js".
  --
  Range { start: Position { line: 7, character: 19 }, end: Position { line: 7, character: 53 } }
  Uncached or missing remote URL: "http://localhost:4545/x/a/mod.ts".
  --
---
```

cli/tests/integration/lsp_tests.rs
```diff
@@ -3442,7 +3525,10 @@ fn lsp_diagnostics_refresh_dependents() {
       "version": 1
     })
   );
-  client
+
+  // todo(dsherret): I'm not sure exactly what this is testing here
```
dsherret, line 3529: I should resolve this todo, but I'm not really sure what was being tested before. I think previously just a lot of past diagnostics were buffered up because each `did_open` actually causes 3 * number of files open, but only 3 was being read each time.

cli/tests/integration/lsp_tests.rs
```diff
@@ -3547,7 +3609,7 @@ fn lsp_performance() {
     .unwrap();
   assert!(maybe_err.is_none());
   if let Some(res) = maybe_res {
-    assert!(res.averages.len() >= 6);
+    assert!(res.averages.len() >= 5);
```
dsherret, line 3612: 5 now because there's no performance trace for the overall diagnostics since they all happen independently now. I wonder if this should be `assert_eq!(res.averages.len(), 5);`?

cli/lsp/diagnostics.rs
```diff
@@ -203,15 +131,131 @@ impl DiagnosticsServer {
...
+                  if !token.is_cancelled() {
+                    for (specifier, version, diagnostics) in diagnostics {
+                      client
+                        .publish_diagnostics(specifier, diagnostics, version)
+                        .await;
+                    }
+                    performance.measure(mark);
+                  }
+                }
+              }));
```
dsherret, line 258: I know... this function is large. I think in a future PR it might be good to refactor this into a struct so the dependencies can be shared easily amongst the functions.

cli/tests/integration/lsp_tests.rs
```diff
@@ -2818,12 +2923,8 @@ fn lsp_code_actions_deadlock() {
   assert!(maybe_err.is_none());
   assert!(maybe_res.is_some());
 
-  for _ in 0..3 {
-    let (method, _) = client.read_notification::<Value>().unwrap();
-    assert_eq!(method, "textDocument/publishDiagnostics");
-  }
+  read_diagnostics(&mut client);
 
-  assert!(client.queue_is_empty());
```
dsherret, line 2826: I removed this check because it's now flaky because of the multiple `textDocument/didChange` notifications in a row above without waiting for diagnostics (it's actually now possible for lint or deps diagnostics to finish really quickly).

cli/tests/integration/lsp_tests.rs
kitsonk: That is trying to be tested here is that the diagnostics for dependents get invalidated and recalculated.

For example if you have `a.ts` the exports something which is imported into `b.ts`, when you modify `a.ts`, it impacts that diagnostics of `b.ts` without `b.ts` being edited.  We had issues where diagnostics were hanging around until you actually touched/modified a dependent file.

cli/tests/integration/lsp_tests.rs
kitsonk: It used to be somewhat fragile with `== 6` IIRC and so it was changed to `>= 6`, but I guess we can try now.

kitsonk (approved):
LGTM outside if clarifying the refresh dependents and that we are still actually testing what that case needs to cover.  Hard to assess that just from the diff.

cli/tests/integration/lsp_tests.rs
dsherret: Ok, let's try it and see how it goes.

PR #13436, opened 2022-01-20
feat(serde_v8): deserialize ArrayBuffers

AaronO:
Previously we would only deserialize `ArrayBufferView`s as zero-copy bufs

This avoids rewrapping `ArrayBuffer`s in `ArrayBufferView`s when implementing APIs that take [BufferSource](https://webidl.spec.whatwg.org/#BufferSource) args passed through the op-layer

### DIFF
--- a/serde_v8/src/de.rs
+++ b/serde_v8/src/de.rs
@@ -134,9 +134,20 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
       ValueType::ArrayBufferView => {
         v8::Local::<v8::ArrayBufferView>::try_from(self.input)
           .and_then(|view| {
-            magic::zero_copy_buf::ZeroCopyBuf::try_new(self.scope, view)
+            magic::zero_copy_buf::ZeroCopyBuf::try_from((
+              &mut *self.scope,
+              view,
+            ))
           })
-          .map_err(|_| Error::ExpectedInteger)
+          .map_err(|_| Error::ExpectedBuffer)
+          .and_then(|zb| visitor.visit_byte_buf(Vec::from(&*zb)))
+      }
+      ValueType::ArrayBuffer => {
+        v8::Local::<v8::ArrayBuffer>::try_from(self.input)
+          .and_then(|buffer| {
+            magic::zero_copy_buf::ZeroCopyBuf::try_from(buffer)
+          })
+          .map_err(|_| Error::ExpectedBuffer)
           .and_then(|zb| visitor.visit_byte_buf(Vec::from(&*zb)))
       }
     }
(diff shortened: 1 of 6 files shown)


serde_v8/src/magic/zero_copy_buf.rs
```diff
@@ -25,31 +25,46 @@ pub struct ZeroCopyBuf {
...
-    if backing_store.is_shared() {
-      return Err(v8::DataError::BadType {
+    let backing_store = buffer.get_backing_store();
+    match backing_store.is_shared() {
+      true => Err(v8::DataError::BadType {
         actual: "shared ArrayBufferView",
         expected: "non-shared ArrayBufferView",
-      });
+      }),
+      false => Ok(Self {
```
ry, line 39: match-boolean seems better to write as an if-else

serde_v8/src/de.rs
```diff
@@ -339,12 +350,20 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
 
     // Magic Buffer
     if name == magic::buffer::BUF_NAME {
-      let zero_copy_buf =
-        v8::Local::<v8::ArrayBufferView>::try_from(self.input)
+      let zero_copy_buf = match self.input.is_array_buffer() {
+        // ArrayBuffer
+        true => v8::Local::<v8::ArrayBuffer>::try_from(self.input)
```
ry, line 355: match-boolean is better to write as if-else

serde_v8/src/magic/buffer.rs
```diff
@@ -15,22 +16,29 @@ pub enum MagicBuffer {
 }
 
 impl MagicBuffer {
-  pub fn new<'s>(
-    scope: &mut v8::HandleScope<'s>,
-    view: v8::Local<v8::ArrayBufferView>,
-  ) -> Self {
-    Self::try_new(scope, view).unwrap()
+  pub fn empty() -> Self {
+    MagicBuffer::ToV8(Mutex::new(Some(vec![0_u8; 0].into_boxed_slice())))
```
ry, line 20: untested function

ry (approved):
few comments but LGTM

serde_v8/src/magic/buffer.rs
AaronO: And potentially unused, but it's not a new change/func from this PR it was simply moved up since the other funcs were removed and replaced with TryFrom<T>s

serde_v8/src/de.rs
AaronO: As discussed offline my reasoning is that we're functionally mapping a union/enum type where tag is this external check so match seems more fitting as a map. But clippy has a https://rust-lang.github.io/rust-clippy/master/#match_bool pedantic rule, so I'll explore applying that in a follow-up PR

Issue #11158, opened 2021-06-28
Proposal: Event based Signal API


kt3k:
This proposal suggests removing the existing Signal APIs and instead suggests adding Event based Signal APIs.

Suggested APIs

```ts
const handler = () => { console.log("Got SIGINT"); }
addEventListener("SIGINT", handler); // attach to the SIGINT event
removeEventListener("SIGINT", handler); // detach from the SIGINT event
```

Note: [DOM spec](https://dom.spec.whatwg.org/#introduction-to-dom-events) allow us to dispatch an event with an arbitrary name. So I think this doesn't break web compatiblity.

## Motivations

Currently we have AsyncIterable based API in the unstable namespace.

```ts
for await (const _ of Deno.signal(Deno.Signal.SIGINT)) {
  console.log("Got SIGINT");
}
```

And optionally Promise based API.

```ts
await Deno.signal(Deno.Signal.SIGINT); // Resolves when the first signal comes
console.log("Got SIGINT");
```

These APIs are confusing to the users because of several reasons.

(1.) Because signal handlers use AsyncUnref Op internally, the runtime can exit in the middle of for-await loop without breaking it, and this is confusing.

```ts
async function listenSigint() {
  for await (const _ of Deno.signal(Deno.Signal.SIGINT)) {
    console.log("Got SIGINT");
  }
  console.log("Finished the loop")
}
listenSigint();
```
The above example exits immediately without printing anything.

(2.) Similarly to (1.), the runtime can exit even when the promise is still in pending state. This behavior is confusing (There is no other promise APIs which works like this)

```ts
async function listenSigint() {
  await Deno.signal(Deno.Signal.SIGINT)
  console.log("Got SIGINT");
}
listenSigint();
```
The above example exits immediately without printing anything.

(3.) One of the reasons to choose async iterable API design is that it can handle the back pressure effectively, but in the case of signals, it isn't used in such a way (A program isn't flooded by signals in general). So this reasoning doesn't apply to signal APIs.

Event based API addresses these problems in a natural way

```ts
addEventListener("SIGINT", () => {
  console.log("Got SIGINT");
});
```

The above program exits immediately, but I think this doesn't cause any surprise to the users unlike async iterable based API.

caspervonb:
At the system call level, signals are exposed as interrupts so handling them as events makes so much more sense and would even make it much simpler to add default behaviours because those could just be cancelled with `preventDefault()`.

ry:
[...]

PR #13438, opened 2022-01-20
feat: stabilize addSignalListener API

### LINKED ISSUE #13102: Stabilise Deno.Signal
Am really looking forward to using this in a module as it will be super beneficial, so hoping this API could be stabilised.

kt3k:
closes #13102

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2260,43 +2260,8 @@ declare namespace Deno {
      * p.close();
      * ```
      */
-    kill(signo: Signal): void;
+    kill(signo: string): void;
   }
-
-  export type Signal =
-    | "SIGABRT"
-    | "SIGALRM"
-    | "SIGBUS"
-    | "SIGCHLD"
-    | "SIGCONT"
-    | "SIGEMT"
-    | "SIGFPE"
-    | "SIGHUP"
-    | "SIGILL"
-    | "SIGINFO"
-    | "SIGINT"
-    | "SIGIO"
-    | "SIGKILL"
-    | "SIGPIPE"
-    | "SIGPROF"
-    | "SIGPWR"
-    | "SIGQUIT"
-    | "SIGSEGV"
-    | "SIGSTKFLT"
-    | "SIGSTOP"
-    | "SIGSYS"
-    | "SIGTERM"
-    | "SIGTRAP"
-    | "SIGTSTP"
-    | "SIGTTIN"
-    | "SIGTTOU"
-    | "SIGURG"
-    | "SIGUSR1"
-    | "SIGUSR2"
-    | "SIGVTALRM"
-    | "SIGWINCH"
-    | "SIGXCPU"
-    | "SIGXFSZ";
 
   export type ProcessStatus =
     | {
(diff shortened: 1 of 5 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -2260,43 +2260,42 @@ declare namespace Deno {
...
-    | "SIGURG"
-    | "SIGUSR1"
-    | "SIGUSR2"
-    | "SIGVTALRM"
-    | "SIGWINCH"
-    | "SIGXCPU"
-    | "SIGXFSZ";
+  // TODO(kt3k): Remove this type when std updated the signal usage.
+  /** @deprecated Use string type instead. */
+  export type Signal = string;
```
satyarohith, line 2268: Why are we using a string?

lucacasonato: I second this. We should keep the type alias IMO.

bartlomieju: I agree, this change is undesirable

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2263,9 +2263,42 @@ declare namespace Deno {
     kill(signo: string): void;
   }
 
-  // TODO(kt3k): Remove this type when std updated the signal usage.
+  // TODO(kt3k): Remove this type later.
```
lucacasonato, line 2266: Why?

kt3k: Because we don't use `Signal` in other part (`kill`, `addSignalListener`, etc) anymore.

Do you think we should keep using Signal type in those params? (Maybe I read Bert's comment in a wrong way)

lucacasonato: Yes, I think we should.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2298,6 +2300,36 @@ declare namespace Deno {
...
+   *
+   * ```ts
+   * Deno.addSignalListener("SIGTERM", () => {
+   *   console.log("SIGTERM!")
+   * });
+   * ```
+   *
+   * NOTE: This functionality is not yet implemented on Windows.
+   */
+  export function addSignalListener(signal: string, handler: () => void): void;
```
lucacasonato, line 2313:
```suggestion
  export function addSignalListener(signal: Signal, handler: () => void): void;
```

I think this should use the `Signal` type definition. It makes for much better editor completions.

piscisaureus:
I had objections to the platform-dependent signal list we were using in the past, I think this is fine.

piscisaureus (approved):
LGTM

benjamingr:
Is there any reason you're not using the standard EventTarget but rather a custom implementation here?

kt3k:
@benjamingr
The section 2.8 of DOM spec (ref https://dom.spec.whatwg.org/#observing-event-listeners) says:

> In general, developers do not expect the presence of an event listener to be observable. The impact of an event listener is determined by its callback. That is, a developer adding a no-op event listener would not expect it to have any side effects.

So `addEventListener` shouldn't have any side effects, but we can't implement signal handlers in that way (Listening to signals remove the default behavior of getting that signal). We think implementing signal handler like
```
window.addEventListener("SIGINT", handler);
```
will break/reduce the conformance to the web spec. So we chose this non-standard name

More context: https://github.com/denoland/deno/issues/11158#issuecomment-871186326

PR #13443, opened 2022-01-20
fix(ext/websocket): WebSocket dispatch single close event

### LINKED ISSUE #12909: WebSocket event `close` dispatches twice when calling `WebSocket.close()`
```
$ deno --version
deno 1.16.2 (release, x86_64-pc-windows-msvc)
v8 9.7.106.2
typescript 4.4.2
```
Calling `WebSocket.close()` will dispatch the event `close` twice. Here's a minimal example:

```ts
const url = "any wss url here";

const socket = new WebSocket(url);

socket.onopen = (_ev) => console.log("connection opened");

let i = 0;
socket.onclose = (ev) => console.log(ev, i++);

setTimeout(() => {
  console.log("closing");
  socket.close();
}, 5000);
```

Any URL works, but you can visit https://www.piesocket.com/websocket-tester to get a URL that looks something like `wss://demo.piesocket.com/v3/channel_1?api_key=`.

Providing a `code` and/or `reason` does not make a difference. However, NOT providing a code will result in the two close events containing different codes: `1005` and `1000`, respectively. Providing one will make them the same (e.g. `4999` and `4999`). This is a bit unrelated to the original issue, but it's still something noteworthy.

The result when running the example:

```
$ deno run -A test.ts
connection opened
closing
CloseEvent {
  ...
  code: 1005,
  reason: ""
} 0
CloseEvent {
  ...
  code: 1000,
  reason: ""
} 1
```

crowlKats:
Closes #12909
Also added custom Inspect


ext/websocket/01_websocket.js
```diff
@@ -363,28 +364,16 @@
...
-            const event = new CloseEvent("close", {
-              wasClean: true,
-              code: code ?? 1005,
-              reason,
-            });
-            this.dispatchEvent(event);
-            core.tryClose(this[_rid]);
-          },
-        );
+        core.opAsync("op_ws_close", {
```
lucacasonato, line 367: Need to probably add a `.catch(() => {})` to this so we don't end up with uncatchable promise rejections.

crowlKats: maybe in that case we should close the ws and error with an error event?

ext/websocket/01_websocket.js
```diff
@@ -363,28 +365,31 @@
...
           }),
-          () => {
+          (err) => {
             this[_readyState] = CLOSED;
-            const event = new CloseEvent("close", {
-              wasClean: true,
-              code: code ?? 1005,
-              reason,
+
+            const errorEv = new ErrorEvent("error", {
```
aapoalas, line 377: I don't know if this is WebIDL dictated stuff, but this sort of seems like a case where the new `cause` error property could be used?

crowlKats: `cause` is for `Error`s, not `ErrorEvent`s.

aapoalas: Ah, indeed. Sorry, my mistake.

AaronO (approved):

PR #13459, opened 2022-01-21
chore: upgrade to rust 1.58.1

littledivy:
Due to [Race condition in std::fs::remove_dir_all](https://github.com/rust-lang/rust/security/advisories/GHSA-r9cc-f5pr-p3j2)

Also adds a `rust-toolchain`


rust-toolchain
```diff
@@ -0,0 +1 @@
+1.58.1
```
ry, line 1: What's this file?

dsherret: It will cause cargo to always build on 1.58.1

dsherret, line 1: Maybe instead we should use a rust-toolchain.toml file? It seems like it's the new way and we might be able to leverage it in the CI?

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

.github/workflows/ci.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Install Rust
         uses: hecrj/setup-rust-action@v1
         with:
-          rust-version: 1.58.0
+          rust-version: 1.58.1
```
dsherret, line 88: Is there an action we can use so we don't have to specify this and instead it could look at the toolchain file?

littledivy: There is https://github.com/dtolnay/rust-toolchain

littledivy: Seems to work

rust-toolchain
littledivy: Done

rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.58.1"
```
ry, line 2: Is this necessary? all these random root level files are fairly annoying.

littledivy: It's just to make sure local builds (like Apple M1) happen from the same toolchain.

dsherret: It will prevent time spent on these scenarios: https://discord.com/channels/684898665143206084/684911491035430919/931238729841016902

And as @littledivy mentioned in discord, it will also ensure anyone doing an M1 build builds with the correct rust version. I think it's worth it, but do agree it's a little annoying to have yet another file in the root.

satyarohith: > It's just to make sure local builds (like Apple M1) happen from the same toolchain.

Is there an alternative?

.devcontainer/Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM mcr.microsoft.com/vscode/devcontainers/rust:0-1
 
-# Update to Rust 1.58.0
-RUN rustup update 1.58.0 && rustup default 1.58.0
+# Update to Rust 1.58.1
+RUN rustup update 1.58.1 && rustup default 1.58.1
```
dsherret, line 4: Is this necessary anymore? Maybe we can remove this now that we have a toolchain file?

rust-toolchain.toml
littledivy: Yes but none are persistent: https://rust-lang.github.io/rustup/overrides.html#overrides

Note that we also have this in rusty_v8 https://github.com/denoland/rusty_v8/blob/main/rust-toolchain

dsherret: It's a shame it seems like rust-toolchain.toml is not respected in the `.cargo` directory.

.devcontainer/Dockerfile
littledivy: yup, removed.

rust-toolchain.toml
dsherret: Also, it's nice to have a single source of truth for our rust version after this PR.

bnoordhuis (approved):
LGTM. Switching to rust-toolchain.toml is a good idea.

PR #13461, opened 2022-01-21
refactor(core): Extract poll_value out of resolve_value

espindola:
This is similar to run_event_loop X poll_event_loop. The reason for
this is that the future returned by resolve_value has the same
lifetime as the runtime. That is fine when rust can reason about it,
but in some cases it can't.

In the particular case we have, our runtime, which includes a
JsRuntime, is stored in a Rc<RefCell<...>>. With this change we can
easily poll to see if a value has been fulfilled.

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -686,6 +686,42 @@ impl JsRuntime {
     scope.perform_microtask_checkpoint();
   }
 
+  pub fn poll_value(
+    &mut self,
+    global: &v8::Global<v8::Value>,
+    cx: &mut Context,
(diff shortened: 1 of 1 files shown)


bartlomieju (commented):
Please add a test case to `core/runtime.rs` that exercises this API

espindola:
> Please add a test case to `core/runtime.rs` that exercises this API

I have added a test. It is a bit of a copy and paste of test_resolve_value. Let me know if I should replace test_resolve_value instead.

bartlomieju (approved):
LGTM, thanks @espindola

Issue #12126, opened 2021-09-17
Coverage Lcov Output Wrong?


ebebbington:
Wonder if i've found a possible bug with coverage, where the standard coverage output is correct, but maybe not the lcov file?
lemme paste some code and links
mod.ts is clearly being covered: https://github.com/ebebbington/crumpets/runs/3624986294#step:4:76

but in the codecov results, only the test deps are being shown: https://app.codecov.io/gh/ebebbington/crumpets

this is the gh workflow i use: https://github.com/ebebbington/crumpets/blob/main/.github/workflows/master.yml#L25

bartlomieju:
@ebebbington any chance you could try again with latest canary? I landed https://github.com/denoland/deno/pull/13334 which should resolve most of the issues with wrong output.

ebebbington:
@bartlomieju Thanks, will try as soon as i have the spare time

bartlomieju:
I'm going to tentatively close this issue. Please let me know if the issue still can be reproduced.

jespertheend:
I'm getting weird output in `1.18.0`. If I place empty lines at the top of the file the coverage seems to shift downwards for every line I place.
I'll try to make a reduced test case.

jespertheend:
I think it's related to cache somehow. The moment I move the file with incorrect coverage to another location the issue gets fixed.

dsherret:
@jespertheend I believe that would be fixed by #13303. The emit cache has the potential to get out of sync at the moment.

jespertheend:
Ah yes that would be it I think. Thanks! If I remove caches/gen the issue is fixed. There's a .js and .js.meta file in there with old content of my script.

jespertheend:
This is probably useless at this point but I put effort in it so just in case: [denocoverage.zip](https://github.com/denoland/deno/files/7908912/denocoverage.zip)
Run with `./runtest.js` and then view `cov/html/denocoverage/foobar.js.gcov.html`, though you'll need to place the content of `placeInCache` in your cache in order to trigger the issue.

jespertheend:
I checked out the PR and I can confirm this indeed fixes my issue.

jespertheend:
Any updates on this? I keep having to downgrade to deno 1.17.3 in order to get meaningful coverage reports.

bartlomieju:
@jespertheend unfortunately not, I haven't got time to look into this problem yet. I can provide some hints on what might be the problem, but I don't expect to be look into this for a few more weeks.

[...]

PR #13462, opened 2022-01-21
fix(emit): eliminate multiple cache files getting out of sync and fix racy emit

### LINKED ISSUE #13302: Emit, source map, and declaration cache in deno dir is racy across multiple processes

dsherret:
1. Uses an sqlite database file to store the emit cache to eliminate multiple cache files getting out of sync.
2. The result of checking the emit is stored in memory and if valid is kept in memory then used in the rest of the emit. This prevents consulting the file system to check if the cache is valid and then reading from the file system later when the cache may no longer be valid, which was racy.

Supersedes #13303.

Closes #13302

WIP though basically done. I could do way more cleanup and have a better design, but wanted to make the minimal changes possible.

TODO:

- [ ] Ensure all calls to `get_emit_data` verify the source. Maybe an API change would be good to force this (ex. `ProcState`'s `SourceMapGetter` impl right now retrieves the source map from the cache, but doesn't do any source checks on it). I think though it probably needs more rework in that area and I will have to read the code more.
- [ ] coverage's `cover_files` does not verify the retrieved emit text is for the source


cli/emit_cache.rs
```diff
@@ -0,0 +1,324 @@
...
+  /// these stored here for mutating for testing purposes
+  emit_data_version: usize,
+  ts_build_info_version: usize,
+}
+
+impl SqliteEmitCache {
+  pub fn new(db_file_path: &Path) -> Result<Self, AnyError> {
+    let conn = Connection::open(db_file_path)?;
+    Self::from_connection(conn)
+  }
```
littledivy, line 101: Consider using these pragmas, they can significantly speed things up. https://github.com/denoland/deno/blob/56d0ca7f52d9bff7326bb2b9406ea1e0282a5fbe/ext/webstorage/lib.rs#L69-L81

dsherret: Thanks for letting me know! I added them 🙂

jespertheend:
Any updates on this? I'm hoping this will fix https://github.com/denoland/deno/issues/12126

KnorpelSenf:
Potentially fixes #14206, as well.

dsherret:
Closing because this is really out of date. I added an `EmitCache` trait in #14925 (merged) and will follow up with another PR in a few weeks (before 1.24) that uses the sql cache from this PR (along with doing the same in /x/deno_cache).

PR #13467, opened 2022-01-22
fix(lsp): respect DENO_CERT and other options related to TLS certs

### LINKED ISSUE #13437: lsp: does not use $DENO_CERT when fetching registry configurations
**Describe the bug**

When opening VS Code to a Deno project, this extension errors when attempting to request an external URL.  I am behind a corporate proxy with a self signed cert.

**To Reproduce**

1. Have a self signed cert in your certificate chain.
2. Open a Deno project.
3. Error occurs.

**Expected behavior**

When VS Code opens, I do not expect to see this error.  Instead, I would expect this extension to use the $DENO_CERT environment variable to access any external resources.

Here is a curl example that shows the same behavior.

```shell
$ curl https://x.nest.land/.well-known/deno-import-intellisense.json
curl: (60) SSL certificate problem: self signed certificate in certificate chain
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

Here is a curl example where $DENO_CERT is used and the request is successful.

```shell
$ curl --cacert $DENO_CERT https://x.nest.land/.well-known/deno-import-intellisense.json
Moved Permanently. Redirecting to https://intellisense.nest.land/deno-import-intellisense.json
```

**Screenshots**
vscode-deno-error

**Versions**

- vscode: 1.63.2
- deno: 1.17.3
- extension: 3.10.1

kitsonk:
Fixes #13437

The LSP now respects `DENO_CERT` and `DENO_TLS_CA_STORE` for registry suggestions and caching remote modules.

This also adds the options to the configuration of the LSP:

- `tlsCertificate` - this mirrors the `--cert` option (and overrides the `DENO_CERT` environment variable)
- `certificateStores` - this overrides the `DENO_TLS_CA_STORE` enviornment variable.
- `unsafelyIgnoreCertificateErrors` - this mirrors the `--unsafely-ignore-certificate-errors` command line option.

It extracts the creation logic for `RootCertStore` which is used by the file fetcher to the file fetcher which is now shared between `ProcState` and the lsp.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -48,7 +48,8 @@ use super::lsp_custom;
 use super::parent_process_checker;
 use super::performance::Performance;
 use super::refactor;
-use super::registries;
+use super::registries::ModuleRegistry;
+use super::registries::ModuleRegistryOptions;
 use super::text;
 use super::tsc;
 use super::tsc::AssetDocument;
@@ -96,7 +97,7 @@ pub(crate) struct Inner {
   /// on disk or "open" within the client.
   pub(crate) documents: Documents,
   /// Handles module registries, which allow discovery of modules
-  module_registries: registries::ModuleRegistry,
+  module_registries: ModuleRegistry,
   /// The path to the module registries cache
   module_registries_location: PathBuf,
   /// An optional path to the DENO_DIR which has been specified in the client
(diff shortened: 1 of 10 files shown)


kitsonk:
Also, to take full advantage of the new options, clients will need to support setting them.  I will be raising a PR for `vscode_deno`.

cli/lsp/language_server.rs
```diff
@@ -139,8 +139,14 @@ impl Inner {
...
     let module_registries_location = dir.root.join(REGISTRIES_PATH);
-    let module_registries =
-      registries::ModuleRegistry::new(&module_registries_location);
+    let module_registries = registries::ModuleRegistry::new(
+      &module_registries_location,
+      None,
+      None,
+      None,
+      None,
+    )
```
bartlomieju, line 148: Consider adding `ModuleRegistryOptions` struct

bartlomieju (approved):
LGTM :)

PR #13473, opened 2022-01-23
perf(lsp): improve some tsc op hot paths

kitsonk:
Chasing down another issues I realised that while individual `op_exists` and `op_script_version` are not performance issues themselves, they are often called 1000s of time per document and we should really try to be as efficient with them as possible.

### DIFF
--- a/cli/lsp/documents.rs
+++ b/cli/lsp/documents.rs
@@ -724,9 +724,9 @@ impl FileSystemDocuments {
     &mut self,
     cache: &HttpCache,
     maybe_resolver: Option<&dyn deno_graph::source::Resolver>,
-    specifier: ModuleSpecifier,
+    specifier: &ModuleSpecifier,
   ) -> Option<Document> {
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):

cli/lsp/documents.rs
```diff
@@ -949,6 +949,22 @@ impl Documents {
     }
   }
 
+  /// Used by the tsc op_exists to shortcut trying to load a document to provide
+  /// information to CLI without allocating a document.
+  pub(crate) fn exists(&self, specifier: &ModuleSpecifier) -> bool {
```
dsherret, line 954: It's a little strange that we now have `exists` and `contains_specifier`, which both do the same thing. I don't think we need two methods. What's the performance impact of using `contains_specifier` vs `exist`? When TypeScript does an `exists` call it will then load the file, so is it a big deal to just load it when doing `.get()`?

PR #13474, opened 2022-01-24
feat(cli): update to TypeScript 4.6.2

kitsonk:


kitsonk:
Note that the official typings for `Error.prototype.cause` are being set to `Error`, which also reflects what is indicated in mdn, and I assume it was taken from the WebIDL.  We had patched it slightly differently when we added the types, but I recommend we move over to the stricter types and fix any regressions it causes.

crowlKats:
Hm quite annoying about `cause`, as the spec doesnt specify whatsoever what it should be, and imo it make sense that it can be anything.

kitsonk:
Well and some of the examples they use aren't strict mode compatible, cause `catch(err)` is `unknown` in strict mode now... It maybe worth raising a PR/issue to get it changed, as it in theory should be `unknown` to match the strict behaviour of `catch(err)`.

bartlomieju (approved):
LGTM.

As a side note, do we have instructions how to do update to newer TypeScript version in case you weren't able to do it?

kitsonk:
I'll add something

PR #13479, opened 2022-01-24
refactor(lsp): Documents - combine duplicate exists methods

dsherret:
`Document.exists` was added in #13473, but we already have a method for doing this. This changes combines the two methods and uses the new functionality.

### DIFF
--- a/cli/lsp/documents.rs
+++ b/cli/lsp/documents.rs
@@ -719,7 +719,8 @@ struct FileSystemDocuments {
 }
 
 impl FileSystemDocuments {
-  /// Adds or updates a document by reading the document from the file system.
+  /// Adds or updates a document by reading the document from the file system
+  /// returning the document.
   fn refresh_document(
     &mut self,
     cache: &HttpCache,
@@ -756,7 +757,8 @@ impl FileSystemDocuments {
       )
     };
     self.dirty = true;
-    self.docs.insert(specifier.clone(), doc)
+    self.docs.insert(specifier.clone(), doc.clone());
+    Some(doc)
   }
 }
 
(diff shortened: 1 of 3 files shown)


cli/lsp/documents.rs
```diff
@@ -929,7 +924,16 @@ impl Documents {
...
+    let specifier = self.specifier_resolver.resolve(specifier);
+    if let Some(specifier) = specifier {
+      if self.open_docs.contains_key(&specifier) {
+        return true;
+      }
+      if let Some(path) = get_document_path(&self.cache, &specifier) {
+        return path.is_file();
+      }
+    }
+    false
```
dsherret, line 936: I moved the code from `exists` here, but I still think we should just do `self.get(specifier).is_some()` because it's simpler and uses the same code path as `.get` so it future proofs this method. In tsc when it calls `exists`, if true then it will cause `.get` to be called so there's not really any reason to not just load it the first time in my opinion.

kitsonk:
There is a change in behaviour in `contains_specifier()`, specifically in calling `get()` that isn't ever needed by the `op_exists`, which is why I broke it out.  `op_exists` really only looks for local files on disk, mostly probing.  Therefore it abandons the locking of the `file_system_docs` and the calculation of the "freshness" of the document before returning a result.  It simply determines if the file exists on the local file system (or is open in the editor).  "op_exists" and therefore "exists()" are not called by tsc when resolving modules, it is only for probing the filesystem for things like the existance of package.json and other environmental files which might affect its module resolution, which we override anyways.  Very seldom does the file even exist and even more rarely the file is read.

While the other two locations where we use `contains_specifier()`, there is dependence on the "freshness" of the file and likelihood the file will will be otherwise accessed in subsequent calls.

We can make the comments more clear, but I think if we change `contains_specifier()` to the same logic, we will end up with some esoteric edge cases... `exists()` is really only a low level optimisation of `op_exists`, because when opening a file now in 4.5, tsc will issue ~1000 op_exists requests for files that almost never exist in a Deno context and locking the `file_system_docs` just to find out the file doesn't exist is quite unnecessary.

dsherret:
> We can make the comments more clear, but I think if we change contains_specifier() to the same logic, we will end up with some esoteric edge cases... exists() is really only a low level optimisation of op_exists, because when opening a file now in 4.5, tsc will issue ~1000 op_exists requests for files that almost never exist in a Deno context and locking the file_system_docs just to find out the file doesn't exist is quite unnecessary.

@kitsonk I see.

> While the other two locations where we use contains_specifier(), there is dependence on the "freshness" of the file and likelihood the file will will be otherwise accessed in subsequent calls.

This is ok. The `file_system_docs` collection is intended to just help prevent loading the same document multiple times across the `Documents` collection and snapshots. Actually, looking at the code for file_system_docs there seems to be a bug in its logic in one place... I've fixed that in this PR.

Also, I renamed the method to `exists` as that seems to be how it's used in all cases.

kitsonk (approved):
LGTM

PR #13480, opened 2022-01-24
feat(build): Create centos build artifacts

### LINKED ISSUE #1658: Release centos7 compatible binaries
When trying to run deno on Centos 7, it fails:
```bash
[maxim@maxim deno] $ target/release/deno tests/worker.js
target/release/deno: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by target/release/deno)
```
Additional information:
First things first, Centos 7 is based on RHEL 7, as one may know. And, according to [redhat.com](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.0_release_notes/sect-red_hat_enterprise_linux-7.0_release_notes-compiler_and_tools-glibc):

> In Red Hat Enterprise Linux 7, the glibc libraries (libc, libm, libpthread, NSS plug-ins, and others) are based on the glibc 2.17 release

Replacing system glibc is a [very bad idea](https://serverfault.com/questions/894625/safely-upgrade-glibc-on-centos-7#comment1155836_894689), basically because all binaries on the system are compatible with specific glibc version. So, that's not an option.

However, the RHEL 8 Beta is using glibc 2.28 ([source](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html-single/8.0_beta_release_notes/index#compilers_and_development_tools_2)), but Centos 8 will only be released when the stable version of RHEL 8 will be released ([source](https://www.centos.org/forums/viewtopic.php?t=68821)), so updating Centos is not an option right now.

I have also tried installing glibc in a nonstandard location as described [here](https://serverfault.com/a/894689/281189). Unfortunately, when I try to run deno with that custom glibc, I'm getting:
```bash
[maxim@maxim deno]$ LD_LIBRARY_PATH=/opt/glibc-2.18/lib target/release/deno tests/worker.js
Segmentation fault (core dumped)
```
I've tried to debug it with `lldb`, `gdb`, `abrt`, but neither option worked for me. Probably, because I'm doing something wrong.
`lldb`:
```bash
[maxim@maxim deno]$ LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/opt/glibc-2.18/lib lldb -- target/debug/deno tests/worker.js
lldb: relocation error: /opt/glibc-2.18/lib/libc.so.6: symbol _dl_find_dso_for_object, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
```
`gdb`:
```bash
[maxim@maxim deno]$ LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/opt/glibc-2.18/lib gdb target/release/deno tests/worker.js
gdb: relocation error: /opt/glibc-2.18/lib/libc.so.6: symbol _dl_find_dso_for_object, version GLIBC_PRIVATE not defined in file ld-linux-x86-64.so.2 with link time reference
```
And `abrt` won't detect any crashes, even though I've set `OpenGPGCheck = no` and `ProcessUnpackaged = yes` in `/etc/abrt/abrt-action-save-package-data.conf`, restarted `abrtd` and `abrt-ccpp`, and finally tried rebooting, nothing helped.
I hope, that I'm doing something wrong, because right now I think using custom glibc is the only option, if I don't want to containerize deno.

Also, as far as I understand, GLIBC_2.18 isn't a direct requirement of deno, but rather a requirement of some third-party. 
I've found Chromium bug [glibc dependency creeped up to 2.18 in M61, breaking EL7 support](https://bugs.chromium.org/p/chromium/issues/detail?id=749077) which seems relevant, but it was fixed a while ago. 
I'm not really good in C or library symbols to find which dependency requires 2.18, so any help here would be really appreciated. 

I hope that @ry also would agree that we should support Centos, and probably when this issue got fixed, we can run travis not only for [Ubuntu 14.04 LTS (trusty)](https://docs.travis-ci.com/user/reference/trusty/) but also for Centos. Many thanks for your attention!

justinmchase:
This adds a centos binary to the binaries released.

Fixes #1658

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
           - os: windows-2019
             job: test
             profile: release
+          - os: centos-latest
+            job: test
+            profile: release
           - os: ${{ github.repository == 'denoland/deno' && 'ubuntu-20.04-xl' || 'ubuntu-20.04' }}
             job: test
             profile: release


justinmchase:
The error states:

> a step cannot have both the `uses` and `run` keys

.github/workflows/ci.yml
```diff
@@ -340,12 +343,23 @@ jobs:
...
         run: cargo build --release --locked --all-targets
 
+      - name: Build release
+        if: |
+          startsWith(matrix.os, 'centos') &&
+          (matrix.job == 'test' || matrix.job == 'bench') &&
+          matrix.profile == 'release' && !matrix.use_sysroot &&
+          github.repository == 'denoland/deno' &&
+          (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/'))
+        uses: centos:latest
```
bartlomieju, line 360: You can remove this line, it's inferred from the build matrix

.github/workflows/ci.yml
```diff
@@ -365,6 +379,17 @@ jobs:
           zip -r deno-x86_64-unknown-linux-gnu.zip deno
           ./deno types > lib.deno.d.ts
 
+      - name: Pre-release (centos)
+        if: |
+          startsWith(matrix.os, 'centos') &&
+          matrix.job == 'test' &&
+          matrix.profile == 'release'
+        uses: centos:latest
```
bartlomieju, line 387: Ditto

.github/workflows/ci.yml
justinmchase: Thanks, what I'm trying to do is to run the cargo build in centos and I think that github doesn't actually support that in their OS build matrix. I'm not sure...

I'm not totally sure but since this didn't actually work this easily, I think what I need to do is to create an all in one `uses: example/cargo-build-centos:latest` step which is a full docker container which does the cargo build.

LIke this one for example:
https://github.com/roles-ansible/check-ansible-centos-centos7-action

Does that sound like the right strategy?

PR #13483, opened 2022-01-24
fix(lsp): independent diagnostic publishing should include all diagnostic sources on each publish

dsherret:
Previously I must have accidentally tested the wrong binary in the editor and forgot that each publish needs to include the diagnostics from all sources (instead of the client figuring that out). I've fixed the tests now to properly only select the last messages sent for each specifier and it gets the diagnostics for that.

Also fixes some flaky tests caused by it now being done in parallel.


kitsonk (approved):
LGTM as we obviously need to fix it, though I think we really need to be careful about adding back in mutexes.  I made a big mistake a while ago with mutxes everywhere, and we ended up with the deadlock issues and @bnoordhuis helped me clean up all the mutexes to just two, and now we are slowly adding them back in.

dsherret:
@kitsonk that mutex was essentially there before in `DiagnosticCollection`. I just accidentally mistunderstood and removed that code so now I'm adding it back, but just adapted for this 😅

PR #13485, opened 2022-01-25
refactor(lsp): remove RwLock on `Config`

### LINKED ISSUE #13477: lsp: constant 100% cpu usage
OS: Arch (`5.16.1-arch1-1`)
Deno version: 1.18
Code Editor: VSCodium

`deno lsp` gets stuck on 100% after trying to save a file.
image
image

Reproduction:
1. Create empty `.ts` file
2. Save file (while having formatting on save enabled)

dsherret:
1. Removes the RwLock on `Config`
2. Removes the occurence of calling back to the client while holding onto the LSP mutex.

WIP I will clean this up tomorrow.

Maybe fixes #13477.

### DIFF
--- a/cli/lsp/config.rs
+++ b/cli/lsp/config.rs
@@ -236,149 +236,63 @@ pub struct Settings {
 #[derive(Debug)]
 pub struct Config {
   pub client_capabilities: ClientCapabilities,
-  settings: Arc<RwLock<Settings>>,
-  tx: mpsc::Sender<ConfigRequest>,
+  // todo(this PR): revert this pub and clean up
+  pub settings: Settings,
   pub workspace_folders: Option<Vec<WorkspaceFolder>>,
 }
 
 impl Config {
-  pub fn new(client: Client) -> Self {
-    let (tx, mut rx) = mpsc::channel::<ConfigRequest>(100);
-    let settings = Arc::new(RwLock::new(Settings::default()));
-    let settings_ref = settings.clone();
-
-    let _join_handle = thread::spawn(move || {
-      let runtime = create_basic_runtime();
-
-      runtime.block_on(async {
-        loop {
-          match rx.recv().await {
-            None => break,
-            Some(ConfigRequest::All) => {
-              let (specifier_uri_map, items): (
-                Vec<(ModuleSpecifier, ModuleSpecifier)>,
-                Vec<lsp::ConfigurationItem>,
-              ) = {
-                let settings = settings_ref.read();
-                (
-                  settings
-                    .specifiers
-                    .iter()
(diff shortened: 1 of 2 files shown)


cli/lsp/client.rs
```diff
@@ -87,10 +119,11 @@ trait ClientTrait: Send + Sync {
...
     params: lsp_custom::RegistryStateNotificationParams,
   ) -> AsyncReturn<()>;
-  fn configuration(
+  fn specifier_configurations(
     &self,
-    items: Vec<lsp::ConfigurationItem>,
-  ) -> AsyncReturn<Result<Vec<serde_json::Value>, AnyError>>;
+    uris: Vec<lsp::Url>,
+  ) -> AsyncReturn<Result<Vec<Result<SpecifierSettings, AnyError>>, AnyError>>;
+  fn workspace_configuration(&self) -> AsyncReturn<Result<Value, AnyError>>;
```
dsherret, line 126: Instead of having a low level `configuration` method on the Client, I've split it up into two separate higher level functions. This simplifies the caller code and the REPL implementation of this trait.

cli/lsp/config.rs
```diff
@@ -236,149 +242,62 @@ pub struct Settings {
 #[derive(Debug)]
 pub struct Config {
   pub client_capabilities: ClientCapabilities,
-  settings: Arc<RwLock<Settings>>,
-  tx: mpsc::Sender<ConfigRequest>,
+  settings: Settings,
```
dsherret, line 245: The `RwLock` on this is removed along with the separate thread to call the client (it's now just fired off in a task and returns to use the same `LanguageServer` mutex)

cli/lsp/language_server.rs
```diff
@@ -845,31 +840,12 @@ impl Inner {
...
-
     let maybe_config =
       if self.config.client_capabilities.workspace_configuration {
-        let config_response = self
-          .client
-          .configuration(vec![ConfigurationItem {
-            scope_uri: None,
-            section: Some(SETTINGS_SECTION.to_string()),
-          }])
-          .await;
```
dsherret, line 862: Here is the call that could cause a deadlock depending on how the client is implemented. Now this always happens outside the `LanguageServer` mutex.

cli/lsp/config.rs
kitsonk: Ah, yeah, ok... That whole tx/sender bit was to work around deadlocking issues I remember now, so removing it all together is a big 👍 .

kitsonk (approved):
LGTM... and now I remember the whole "configuration server" stuff was to get around a deadlock issue way back when, so it is good to actually remove the RwLock all together... Thank you!

bartlomieju (approved):
LGTM too

PR #13495, opened 2022-01-26
refactor: integrate deno_graph 0.20.0

kitsonk:
Down to 53 errors... just a chore to get the rest of them...


cli/Cargo.toml
```diff
@@ -45,10 +45,10 @@ winapi = "=0.3.9"
...
 
 [dependencies]
-deno_ast = { version = "0.9.0", features = ["bundler", "codegen", "dep_graph", "module_specifier", "proposal", "react", "sourcemap", "transforms", "transpiling", "typescript", "view", "visit"] }
+deno_ast = { version = "0.10.0", features = ["bundler", "codegen", "dep_graph", "module_specifier", "proposal", "react", "sourcemap", "transforms", "transpiling", "typescript", "view", "visit"] }
 deno_core = { version = "0.117.0", path = "../core" }
-deno_doc = "0.26.0"
-deno_graph = "0.18.0"
+deno_doc = "0.28.0"
+deno_graph = "0.21.1"
 deno_lint = { version = "0.22.0", features = ["docs"] }
```
dsherret, line 52: By the way, deno_lint and dprint-plugin-typescript were both released with deno_ast 0.10.0.

kitsonk:
Everything updates and compiles, but there are some regressions in the test cases that I need to dig into.

bartlomieju:
I'm getting a lot of:
```
Don't know how to create cache name for scheme: flags
Don't know how to create cache name for scheme: flags
```
in tests. These originate in:
https://github.com/denoland/deno/blob/a2e4fa471ba3366f7e05bbad59b247e7825b832c/cli/http_cache.rs#L45

I feel like this should be an error rather than error log. If it happens it most likely means that we screwed up something in Deno internals.

This `flags` actually originates from here:
https://github.com/denoland/deno/blob/a2e4fa471ba3366f7e05bbad59b247e7825b832c/cli/compat/mod.rs#L85-L86

kitsonk:
> I'm getting a lot of:
> 
> ```
> Don't know how to create cache name for scheme: flags
> Don't know how to create cache name for scheme: flags
> ```

Ok, yeah, I know the problem... Because we got rid of the seperate "Synthetic" modules, CLI is trying to cache all modules in the graph, including the synthetic ones.  Need to filter those out.

kitsonk:
@bartlomieju ok, I've got a few more challenges to fix, but there is one that I see what the problem is, but I am not totally sure why it is occurring now and the best way to fix it.  If you run test `native_modules_as_global_vars` you will get a panic that the file being passed is not local.  If you look at the stack trace, when the graph is being built, the node resolver is getting a request to resolve a remote URL and then is "blowing up".  I believe the remote URL is being injected to the graph as part of the injected modules, but I am not totally sure why when building the graph, these injected dependencies weren't resolved before and why this isn't working now, nor why the compat code doesn't handle resolution of these "gracefully".

kitsonk:
@bartlomieju it does look just like the compat failures with trying to find the package.json for the shim is the only outstanding issue.

bartlomieju:
> @bartlomieju ok, I've got a few more challenges to fix, but there is one that I see what the problem is, but I am not totally sure why it is occurring now and the best way to fix it. If you run test `native_modules_as_global_vars` you will get a panic that the file being passed is not local. If you look at the stack trace, when the graph is being built, the node resolver is getting a request to resolve a remote URL and then is "blowing up". I believe the remote URL is being injected to the graph as part of the injected modules, but I am not totally sure why when building the graph, these injected dependencies weren't resolved before and why this isn't working now, nor why the compat code doesn't handle resolution of these "gracefully".

@kitsonk there is a single path that injects resolved specifier for built-in Node modules (pointing to `https://deno.land/std/node`). Unfortunately I made a mistake when mapping these specifiers to `ResolveResult`. I now fixed it and the tests should pass. I think I will need to reorganize this code further when working on the CJS/ESM integration, because it is indeed unfortunate that there a single "special" code path that needs to handle `http`/`https` imports inside that resolution algorithm. Most likely it would be removed in the future if we decided to move `std/node` into an extension crate.

kitsonk:
@bartlomieju obviously a topic for another day, but it would be really sad if we can't integrate Deno code and Node.js code in the same workload.

I fixed what I hope are the last fixture issues so the CI can go green.

cli/compat/esm_resolver.rs
```diff
@@ -121,9 +129,23 @@ fn node_resolve(
   let conditions = DEFAULT_CONDITIONS;
   let url = module_resolve(specifier, &parent_url, conditions)?;
 
+  let resolve_response = if url.as_str().starts_with("http") {
```
dsherret, line 132: Nit: check `url.scheme()` instead?

cli/compat/esm_resolver.rs
```diff
@@ -121,9 +129,23 @@ fn node_resolve(
...
+    ResolveResponse::Esm(url)
+  } else if url.as_str().ends_with(".js") {
```
dsherret, line 134: We should be doing case insensitive comparisons. It seems like we often forget about this. Maybe we should create a helper function somewhere that does it efficiently and just in case, also checks `url.path()` instead of the full url?

cli/compat/esm_resolver.rs
```diff
@@ -121,9 +129,23 @@ fn node_resolve(
...
+    let package_config = get_package_scope_config(&url)?;
+    if package_config.typ == "module" {
+      ResolveResponse::Esm(url)
+    } else {
+      ResolveResponse::CommonJs(url)
+    }
+  } else if url.as_str().ends_with(".cjs") {
```
dsherret, line 141: Here too (case insensitive, path comparison)

dsherret (approved):
LGTM! Just one small fix needed though.

cli/compat/esm_resolver.rs
kitsonk: @bartlomieju do you want to address or should I?

cli/compat/esm_resolver.rs
kitsonk: @bartlomieju same again

cli/compat/esm_resolver.rs
kitsonk: @bartlomieju same again

bartlomieju:
> @bartlomieju obviously a topic for another day, but it would be really sad if we can't integrate Deno code and Node.js code in the same workload.
> 
> 
> 
> I fixed what I hope are the last fixture issues so the CI can go green.

I don't think there's any problem with that - you can mix and match http/https imports with bare specifiers. I was mentioning situation with remapping of Node.js built in modules to std URLs.

bartlomieju (approved):
LGTM too. I'll apply David's suggestions in a follow up PR tomorrow morning. Let's land this!

PR #13496, opened 2022-01-26
feat(unstable): add Deno.getUid

### LINKED ISSUE #13493: proposal: Deno.getUid()
Currently there's no way to get the OS's user id of the process. We should add `Deno.getuid()` (or similar API) to get the user id.

Note: If this API is available we can shim `fs.access` in `std/node` more correctly. ref: https://github.com/denoland/deno_std/blob/cf6b45c1822447e4eb978433454cb500f4c91c35/node/_fs/_fs_access.ts#L24-L26

Note: The above issue of `fs.access` prevents yarn from working with deno compat mode. https://github.com/denoland/deno/issues/13420

kt3k:
closes #13493

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -138,12 +138,12 @@ declare namespace Deno {
    * Returns the user id of the process on POSIX platforms. Returns null on windows.
    *
    * ```ts
-   * console.log(Deno.getuid());
+   * console.log(Deno.getUid());
    * ```
    *
    * Requires `allow-env` permission.
    */
-  export function getuid(): number | null;
+  export function getUid(): number | null;
 
   /** All possible types for interfacing with foreign functions */
   export type NativeType =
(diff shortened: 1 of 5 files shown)


bartlomieju (approved):
@kt3k looks good, I have two nitpicks:
a) maybe use `Deno.getUid()`?
b) should we add `Deno.getGid()` too?

kt3k:
@bartlomieju 

> a) maybe use Deno.getUid()?

I'm not strongly against that. I chose this name because we tend to use linux syscall names as is. ex. `chmod`, `chdir`, `lstat`, `utime`, `futime`, `flock`, `funlock`, etc

If we try to make it more descriptive, I guess `Deno.getUserId` might be another option. (similar to `Deno.makeTempDir` instead of `Deno.mktemp`)

I'd like to hear more opinions from other members

> b) should we add Deno.getGid() too?

I think we should. That's also useful for more precise implementation of `fs.access`

bartlomieju:
@kt3k sounds good!

jorroll:
FWIW, I initially interpreted `Deno.getUid()` as "get unique ID". Though hopefully, when hovering over the method in VSCode, I would have noticed the JSDoc comment if I ran across this in an application.

PR #13497, opened 2022-01-26
refactor: update runtime code for primordial checks for "instanceof"

### LINKED ISSUE #13255: Replace uses of `instanceof` in runtime code with `ObjectPrototypeIsPrototypeOf`
The former can be tricked using `Symbol.hasInstance` while the latter can not.

bartlomieju:
Errors surfaced with https://github.com/denoland/deno_lint/pull/992

Closes #13255

### DIFF
--- a/ext/broadcast_channel/01_broadcast_channel.js
+++ b/ext/broadcast_channel/01_broadcast_channel.js
@@ -98,7 +98,7 @@
     }
 
     postMessage(message) {
-      webidl.assertBranded(this, BroadcastChannel);
+      webidl.assertBranded(this, BroadcastChannel.prototype);
 
       const prefix = "Failed to execute 'postMessage' on 'BroadcastChannel'";
       webidl.requiredArguments(arguments.length, 1, { prefix });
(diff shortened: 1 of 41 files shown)


ext/broadcast_channel/01_broadcast_channel.js
```diff
@@ -98,7 +98,7 @@
     }
 
     postMessage(message) {
-      webidl.assertBranded(this, BroadcastChannel);
+      webidl.assertBranded(this, BroadcastChannel.prototype);
```
lucacasonato, line 101: This actually introduces a new issue: users could override `BroadcastChannel.prototype`. We need to "primordialize" this as `BroadcastChannelPrototype` I think.

bartlomieju: Makes sense, given that this is our internal code we will need to provide two "exports" for each of this classes, right?

lucacasonato: Yeah, probably.

ext/fetch/22_body.js
```diff
@@ -202,7 +249,7 @@
          * @returns {ReadableStream<Uint8Array> | null}
          */
         get() {
-          webidl.assertBranded(this, prototype);
+          webidl.assertBranded(this, class_.prototype);
```
lucacasonato, line 252: This has possibility for pollution. Instead of `class_` can you rename back to prototype, and have it be the real prototype? I don't see where we use `class_` without `.prototype`.

lucacasonato (approved):
LGTM. JS is just insanely painful sometimes.

PR #13498, opened 2022-01-26
fix(runtime): Replace instanceof with ObjectPrototypeIsPrototypeOf in Runtime

### LINKED ISSUE #13255: Replace uses of `instanceof` in runtime code with `ObjectPrototypeIsPrototypeOf`

imskr:
- Replace `instanceof` in runtime code with `ObjectPrototypeIsPrototypeOf`

Fixes #13255

### DIFF
--- a/runtime/js/06_util.js
+++ b/runtime/js/06_util.js
@@ -93,7 +93,7 @@
   }
 
   function pathFromURL(pathOrUrl) {
-    if (pathOrUrl instanceof URL) {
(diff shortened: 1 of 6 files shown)


bartlomieju:
@imskr thank you for PR, but this issue turned out to be much more involved and I already did that work in https://github.com/denoland/deno/pull/13497. So unfortunately I'm gonna close without merge.

PR #13499, opened 2022-01-26
fix(ext/web): better typings for ReadableByteStream

crowlKats:
Currently extending `ReadableStream` and using a bytestream is not possible (ts will complain one way or the other). In reality on the changes in `declare var ReadableStream` are necessary. I am not too sure the other changes are accurate to spec


lucacasonato (commented):
All the `V`'s need to have defaults, otherwise this change is backwards incompatible. Also, is TS doing something similar?

crowlKats:
TS doesnt have any readablebytestream related typings

bartlomieju:
Is this PR somehow blocked? Can we land it for 1.19?

lucacasonato (commented):
Does this pass all `deno_std` tests?

crowlKats:
@lucacasonato yep!

kitsonk (approved):
LGTM

ry (commented):
Test?

crowlKats:
> LGTM

@kitsonk as from my original comment, i actually am a bit unsure about if the `ReadableStreamBYOBRequest` part of things is actually correct. i think it is possible to pass any typed array and the spec does conversions, but unsure

crowlKats:
Closing as am unsure of the validity of this PR & staleness.

PR #13518, opened 2022-01-28
chore: Upgrade Tokio to 1.16.0

biluohc:
Tokio 1.16.0 has been released. https://github.com/tokio-rs/tokio/releases/tag/tokio-1.16.0

The release includes improvements to the multi-threaded scheduler that
can increase throughput by up to 20% in some cases.

See more detail: https://github.com/tokio-rs/tokio/pull/4383

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4294,11 +4294,10 @@ checksum = "cda74da7e1a664f795bb1f8a87ec406fb89a02522cf6e50620d016add6dbbf5c"
 
 [[package]]
 name = "tokio"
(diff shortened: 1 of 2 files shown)


biluohc:
Sorry, deno seems use the basic runtime all, there is no need to upgrade.

magurotuna:
In my opinion, it should always be nicer to keep up with external crates, especially vital crates like tokio.

biluohc:
Oh, maybe the other fixed makes sense, i will reopen it if needed

PR #13527, opened 2022-01-29
refactor: primordials for instanceof

bartlomieju:
Trying to reland https://github.com/denoland/deno/commit/884143218fad0e18f7553aaf079d52de703f7601 which was reverted in https://github.com/denoland/deno/commit/f248e6f1778dc26db91d3322de2ecca5d1aa9866


littledivy (approved):
ext/crypto and WPT changes LGTM!

EDIT: the flaky tests passing causes CI to fail.

lucacasonato:
If the tests are flakey, then just setting them as failing won't work. They'll still break when they do pass even though we expected failure.

lucacasonato (approved):

PR #13530, opened 2022-01-29
fix(lsp): regression where certain diagnostics were showing for disabled files

### LINKED ISSUE #13524: Deno 1.18.1 LSP shows errors outside of Deno project
I have a multi-root workspace where some folders like the `backend` use Deno and others like the `frontend` use Node/Next.js.

When I upgraded to Deno 1.18.1, although the LSP feels more performant, it introduced a regression that is showing deno error messages in the Node.js part of the workspace, where Deno should not be enabled at all.

## Before - Deno 1.18.0

When editing a frontend .tsx file, Deno should not be used:

before

## After - Deno 1.18.1

after

You can see TypeScript errors coming from the Deno LSP, but they're not applicable since that part of the project does not have Deno enabled.

dsherret:
Closes #13524

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -4,6 +4,7 @@ use super::analysis;
 use super::client::Client;
 use super::config::ConfigSnapshot;
 use super::documents;
+use super::documents::Document;
 use super::documents::Documents;
 use super::language_server;
 use super::performance::Performance;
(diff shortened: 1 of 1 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -211,16 +212,16 @@ impl DiagnosticsServer {
...
-                      .map_err(|err| {
-                        error!(
-                          "Error generating TypeScript diagnostics: {}",
-                          err
-                        );
-                      })
-                      .unwrap_or_default();
+                  let diagnostics = generate_ts_diagnostics(
+                    snapshot.clone(),
+                    &config,
```
dsherret, line 217: I had intended to flip diagnostics to not be calculated at all upfront if the config was disabled and then I must have completely forgotten to do the second half of that 🤦‍♂️

cli/lsp/diagnostics.rs
```diff
@@ -490,23 +501,39 @@ async fn generate_ts_diagnostics(
...
-        version,
-        ts_json_to_diagnostics(ts_diagnostics),
-      ));
-    }
+  let (enabled_specifiers, disabled_specifiers) = specifiers
+    .iter()
+    .cloned()
+    .partition::<Vec<_>, _>(|s| config.specifier_enabled(s));
+  let ts_diagnostics_map: TsDiagnosticsMap = if !enabled_specifiers.is_empty() {
+    let req = tsc::RequestMethod::GetDiagnostics(enabled_specifiers);
```
dsherret, line 509: I think only doing this for enabled specifiers should be a decent performance improvement in workspaces with a lot of disabled files.

kitsonk (approved):
LGTM

PR #13537, opened 2022-01-30
feat(ext/fetch): Response.json()

### LINKED ISSUE #13902: Implement `Response.json()`
https://github.com/whatwg/fetch/pull/1392 adds support for a static Response.json method to the Fetch spec.

WPTs have already been written, and are available at https://github.com/web-platform-tests/wpt/pull/32825.

AaronO:
A simple/tentative implementation of https://github.com/whatwg/fetch/issues/1389

Closes #13902

### DIFF
--- a/ext/fetch/23_response.js
+++ b/ext/fetch/23_response.js
@@ -153,6 +153,38 @@
     return resp;
   }
 
+  /**
+   * @param {Response} self
+   * @param {ResponseInit} init
+   * @returns {Response}
+   */
+  function fromInit(self, init) {
+    if (init.status < 200 || init.status > 599) {
+      throw new RangeError(
+        `The status provided (${init.status}) is outside the range [200, 599].`,
+      );
+    }
+
+    if (
+      init.statusText &&
+      !RegExpPrototypeTest(REASON_PHRASE_RE, init.statusText)
+    ) {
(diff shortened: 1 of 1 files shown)


ext/fetch/23_response.js
```diff
@@ -209,6 +210,18 @@
...
 
+    /**
+     * @param {any} value
+     * @param {ResponseInit} init
+     * @returns {Response}
+     */
+    static json(value, init = {}) {
+      const body = JSONStringify(value);
+      const response = new Response(body, init);
+      response.headers.set("Content-Type", "application/json");
```
andreubotella, line 221: The `Response` constructor when passed a string, `ArrayBuffer`, `FormData`, etc. will only set the `Content-Type` header if one wasn't passed in the `init` object. `Response.json()` should probably do the same.

lucacasonato (changes requested):
Pending WPT and spec changes. Do not merge.

ext/fetch/23_response.js
AaronO: Handled by 2142d7b61

lucacasonato:
Thanks @AaronO, but closing in favor of #14566. I implemented the change as per the latest spec revision there, and enabled the WPTs.

PR #13542, opened 2022-01-30
feat(ext/fetch): add Headers#getSetCookie

lucacasonato:
Spec change: https://github.com/whatwg/fetch/pull/1346
Tests: https://github.com/web-platform-tests/wpt/pull/31442 (ran against this PR and they all pass)


andreubotella:
Now that the spec PR is about to land, this PR should be reopened.

ext/fetch/20_headers.js
```diff
@@ -329,10 +331,27 @@ class Headers {
     return getHeader(list, name);
   }
 
+  /**
+   * @param {string} name
```
lukewarlow, line 335: I don't believe this comment is correct for this fucking as it has no params

andreubotella:
Hey, aside from web-platform-tests/wpt#31442, I also opened web-platform-tests/wpt#38437 with some extra tests to address some of the review feedback from the Webkit implementation.

bartlomieju (approved):
LGTM

aapoalas (dismissed):
This should be rebased and landed. Spec changes have gone through and I presume this would be a fairly important feature for eg. Node compat.

e3dio:
Tried to use this today, was missing in Deno

andreubotella:
For the record, this implementation is correct according to the spec as it is now. The only changes in the spec text since were editorial. There is one failing test in `fetch/api/headers/header-setcookie.any.js`, that for `Set-Cookie` being a forbidden response header, that Deno fails intentionally (and which will eventually be standardized by the fetch work for server-side runtimes in WinterCG). So the only thing that's blocking this PR from landing is resolving the merge conflicts.

bartlomieju (approved):
LGTM

PR #13546, opened 2022-01-30
docs(ext/web): added AddEventListenerOptions docs

lukeshiels:
I've added some JSDoc comments to help users using addEventListener to understand the (frankly seldomly used) `options` argument for `addEventListener`.

It's my first PR here and I'm not sure if the explanations I've added match up with Deno's implementation of addEventListener as I assume it could possibly differ from the browser implementation.

### DIFF
--- a/ext/web/lib.deno_web.d.ts
+++ b/ext/web/lib.deno_web.d.ts
@@ -140,11 +140,22 @@ declare type EventListenerOrEventListenerObject =
   | EventListenerObject;
 
 interface AddEventListenerOptions extends EventListenerOptions {
+  /** A boolean value indicating that the listener should be invoked at most
+   * once after being added. If true, the listener would be automatically
+   * removed when invoked. */
   once?: boolean;
+  /** A boolean value that, if true, indicates that the function specified by
+   * listener will never call preventDefault(). If a passive listener does call
+   * preventDefault(), the user agent will do nothing other than generate a
+   * console warning. See Improving scrolling performance with passive
+   * listeners to learn more.  */
   passive?: boolean;
 }
 
 interface EventListenerOptions {
+  /** A boolean value indicating that events of this type will be dispatched to
+   * the registered listener before being dispatched to any EventTarget beneath
+   * it in the DOM tree. */
   capture?: boolean;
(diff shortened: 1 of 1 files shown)


lukeshiels:
First note is that according to deno's web platform API [specification deviations](https://deno.land/manual@v1.18.1/runtime/web_platform_apis#spec-deviations), Deno naturally doesn't have a DOM tree so my provided comment for the `capture` option is wrong.

ext/web/lib.deno_web.d.ts
```diff
@@ -140,11 +140,22 @@ declare type EventListenerOrEventListenerObject =
   | EventListenerObject;
 
 interface AddEventListenerOptions extends EventListenerOptions {
+  /** A boolean value indicating that the listener should be invoked at most
```
aapoalas, line 143: Not sure if this has been the way for Deno doc comments, but personally I've used backticks to separate "code references" from normal doc text. eg. To align with the MDN docs, I'd use  \`listener\` to have it show up like this: `listener`.

ext/web/lib.deno_web.d.ts
```diff
@@ -140,11 +140,22 @@ declare type EventListenerOrEventListenerObject =
...
+   * once after being added. If true, the listener would be automatically
```
aapoalas, line 144: I sort of disagree with MDN's grammar here. I think it should be `"If true, the listener **will** be automatically removed when invoked.` Also, perhaps `after invoked`. But MDN is a pretty good source to use.

ext/web/lib.deno_web.d.ts
```diff
@@ -140,11 +140,22 @@ declare type EventListenerOrEventListenerObject =
...
+   * removed when invoked. */
   once?: boolean;
+  /** A boolean value that, if true, indicates that the function specified by
+   * listener will never call preventDefault(). If a passive listener does call
+   * preventDefault(), the user agent will do nothing other than generate a
```
aapoalas, line 149: This might require some customization for Deno context. Scrolling performance will not really make sense for these event listeners, I would presume.

PR #13547, opened 2022-01-31
fix(ext/crypto): support EC p256 private key material in exportKey

### LINKED ISSUE #13549: Spki CryptoKey export not working for ECDSA
Hello,

I found yesterday a bug when exporting ECDSA keys to SPKI, deno doesn't accept properly the input CryptoKey as a Public key.
Here is how you can reproduce it 

```ts
const keys = await crypto.subtle.generateKey({ name: "ECDSA", namedCurve: "P-256" }, true, ["sign","verify"]);
const spki = await crypto.subtle.exportKey("raw", keys.publicKey);
```
This should work but deno returns `expected private key`
note that this error also shows up here [wpt.fyi](https://wpt.fyi/results/WebCryptoAPI/wrapKey_unwrapKey/wrapKey_unwrapKey.https.any.html?label=master&label=experimental&product=chrome&product=deno&product=firefox&product=safari&aligned) (link from the last Deno blogpost!)

The SPKI export key method should only take public key, and PKCS8 should take a privatekey. and it seems to be the other way around here.

littledivy:
Fixes #13549


ext/crypto/shared.rs
```diff
@@ -123,9 +131,9 @@ impl RawKeyData {
       RawKeyData::Public(data) => {
         // public_key is a serialized EncodedPoint
         p384::EncodedPoint::from_bytes(&data)
-          .map_err(|_| type_error("expected valid private EC key"))
+          .map_err(|_| type_error("expected valid public EC key"))
       }
-      _ => Err(type_error("expected private key")),
+      _ => Err(type_error("expected public key data")),
```
cryptographix, line 136: No support for P384?

If you look at the JWK export code, you can see how we export EC public-keys from private-keys, *both for P256 and P384*, using the PKCS8 decoder in ec_key.rs.

lucacasonato (approved):

ext/crypto/shared.rs
littledivy: Thanks, done

littledivy:
Thanks @lucacasonato

cryptographix:
@littledivy maybe should rename issue and merge to include P384?

PR #13501, opened 2022-01-26
refactor: factor out CDP message types

crowlKats:

PR #13551, opened 2022-01-31
refactor: factor out CDP message types

crowlKats:
With fixes this time

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -123,6 +123,18 @@ fn pty_complete_primitives() {
 }
 
 #[test]
+fn pty_complete_expression() {
+  util::with_pty(&["repl"], |mut console| {
+    console.write_text("Deno.\t\t");
+    console.write_text("y");
+    console.write_line("");
+    console.write_line("close();");
+    assert!(console.read_all_output().contains("Display all"));
+  });
+}
+
+
+#[test]
 fn pty_complete_imports() {
   util::with_pty(&["repl"], |mut console| {
     // single quotes
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -129,7 +129,13 @@ fn pty_complete_expression() {
     console.write_text("y");
     console.write_line("");
     console.write_line("close();");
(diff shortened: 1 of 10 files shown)


cli/tools/repl/session.rs
```diff
@@ -344,16 +371,29 @@ impl ReplSession {
...
+          generate_preview: None,
+          user_gesture: None,
+          await_promise: None,
+          throw_on_side_effect: None,
+          timeout: None,
+          disable_breaks: None,
+          repl_mode: Some(true),
+          allow_unsafe_eval_blocked_by_csp: None,
+          unique_context_id: None,
+        }),
```
ry, line 394: use `.. Default::default()` ?

ry:
What is CDP messages? Why is this important?

crowlKats:
@ry this is the exact same PR as #13501, but there were some issues, so it had to be reverted

dsherret (approved):
LGTM

cli/tools/repl/channel.rs
```diff
@@ -55,17 +56,19 @@ pub struct RustylineSyncMessageSender {
 }
 
 impl RustylineSyncMessageSender {
-  pub fn post_message(
+  pub fn post_message<T: serde::Serialize>(
     &self,
     method: &str,
-    params: Option<Value>,
+    params: Option<T>,
   ) -> Result<Value, AnyError> {
```
dsherret, line 63: I wonder if we could make the response a type parameter too? (`serde::Deserialize`)

cli/tests/integration/repl_tests.rs
```diff
@@ -122,6 +122,17 @@ fn pty_complete_primitives() {
...
 }
 
+#[test]
+fn pty_complete_expression() {
+  util::with_pty(&["repl"], |mut console| {
+    console.write_text("Deno.\t\t");
+    console.write_text("y");
+    console.write_line("");
+    console.write_line("close();");
+    assert!(console.read_all_output().contains("Display all"));
```
bartlomieju, line 132: Could you actually assert that some of the expected fields in `Deno` namespace are in the output?

bartlomieju (approved):
LGTM too

Issue #13200, opened 2021-12-25
compat: failed to import "react" from ESM module


SinaMobasheri:
I try to use [ink](https://github.com/vadimdemedes/ink) library from [npm](https://www.npmjs.com/package/ink) so i run `npm install ink react` and `deno run --compat --unstable --allow-net --allow-read script.jsx`

*script.jsx*
```jsx
import React, {useState, useEffect} from 'react';
import {render, Text} from 'ink';

const Counter = () => {
	const [counter, setCounter] = useState(0);

	useEffect(() => {
		const timer = setInterval(() => {
			setCounter(previousCounter => previousCounter + 1);
		}, 100);

		return () => {
			clearInterval(timer);
		};
	}, []);

	return <Text color="green">{counter} tests passed</Text>;
};

render(<Counter />);

```
base on what i can understand from [compatibility mode](https://deno.land/manual@v1.17.1/npm_nodejs/compatibility_mode) it must have no issues..., maybe i miss something ... i don't know here is output  
![output](https://user-images.githubusercontent.com/60809106/147381919-2f14fe1c-3621-48f2-b8fe-f968881ceed0.png)

bartlomieju:
This is most likely related to https://github.com/denoland/deno/issues/12648; `react` is distributed as a CJS module and interoperability between ESM and CJS is still a bit lacking in Deno.

bartlomieju:
Okay, we got a way forward to providing this integration. Currently blocked by https://github.com/denoland/deno_graph/issues/97

Issue #12648, opened 2021-11-04
compat: imported scripts are always handled as ESM


kt3k:
Suppose we have following scripts:

main.mjs
```js
import "./imported.js";
```

imported.js
```js
console.log(exports);
```

If we run this with node, it simply prints `{}`:

```shellsession
$ node main.mjs 
{}
```

But if we run this with deno-compat mode, then it throws like the below:

```shellsession
$ deno run --compat --unstable -A main.mjs
error: Uncaught ReferenceError: exports is not defined
console.log(exports);
            ^
    at file:///path/to/imported.js:1:13
```

The difference here is that node handles `imported.js` as CJS, but deno compat handles it as ESM.

Side note: This prevents deno-compat passing `test/interfaces/exports.spec` test case in `mocha` repository. related: #12577

bartlomieju:
I can look into that next week

bartlomieju:
@kt3k how burning of a need is this issue? I did some exploratory work and it seems harder than I expected to fix this problem.

kt3k:
4 sub test catetories are blocked by this in the mocha compatibility work. So this is very much needed, but there are some more independent issues to work on. So my work is not particularly blocked by this one

bartlomieju:
I did some research on what is the problem here - basically we don't currently have a notion of "translators" which are integral parts of Node's ESM loader. "translators" are used to transform source code of modules to compatible formats; in this case `CJS` module is translated into `ESM` (here's relevant part: https://github.com/nodejs/node/blob/42c0b2ae65e99e4774fe1d8a82a50b09b894adc7/lib/internal/modules/esm/translators.js#L168-L210). 

The hard part in our case is that we got ESM loader in CLI and CJS loader in `deno_std` (as part of `node/module.ts`). There are already some discrepancies between our implementation and Node's (mainly we're not caching read `package.json` files in the same way as Node does), but this is significantly more challenging. I'm leaning towards the answer that we might have to have `require` implemented in Rust before we can tackle this problem, but before giving definitive answer I will prototype some more.

bartlomieju:
So it seems that the main integration point would have to be `ProcState::load` and thus it would also have to touch module graph. We would have to carry on information about "format" of the file (`module`, `commonjs`, `json`) and do the appropriate transforms based on implementation of `translators` from Node (essentially pass-through for ESM, and some handling for `commonjs`). I'm not keen on handling `json` files especially in light of #7623. 

[...]

PR #13553, opened 2022-01-31
feat(compat): CJS/ESM interoperability

bartlomieju:
This commit adds CJS/ESM interoperability when running in --compat mode.

Before executing files, they are analyzed and all CommonJS modules are
transformed on the fly to a ES modules. This is done by utilizing `analyze_cjs()`
functionality from `deno_ast`. After discovering exports and reexports, an ES 
module is rendered and saved in memory for later use.

There's a caveat that all files ending with ".js" extension are considered as
CommonJS modules (unless there's a related "package.json" with "type": "module").

Closes https://github.com/denoland/deno/issues/12648
Closes https://github.com/denoland/deno/issues/13200


cli/proc_state.rs
```diff
@@ -343,6 +343,30 @@ impl ProcState {
...
     .await;
+
+    let needs_cjs_esm_translation = graph
+      .modules()
+      .iter()
+      .any(|m| m.kind == ModuleKind::CommonJs);
+
+    if needs_cjs_esm_translation {
+      for module in graph.modules() {
+        if module.kind == ModuleKind::CommonJs {
```
bartlomieju, line 354: @kitsonk I just hit a very strange problem here. Before running tests I did `export DENO_NODE_COMPAT_URL=file:///Users/ib/dev/deno_std/` to use local copy of `deno_std` instead of pulling sources from `deno.land/std/` and somehow I get following files marked as `CommonJs`:
```
CJS module file:///Users/ib/dev/deno_std/node/_buffer.js
CJS module file:///Users/ib/dev/deno_std/node/_events.js
CJS module file:///Users/ib/dev/deno_std/node/_process/stdio.js
CJS module file:///Users/ib/dev/deno_std/node/_process/streams.js
CJS module file:///Users/ib/dev/deno_std/node/_stream.js
CJS module file:///Users/ib/dev/deno_std/node/internal/blob.js
CJS module file:///Users/ib/dev/deno_std/node/internal/buffer.js
CJS module file:///Users/ib/dev/deno_std/node/internal/normalize_encoding.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/add-abort-signal.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/buffer_list.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/compose.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/destroy.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/duplex.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/end-of-stream.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/from.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/legacy.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/passthrough.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/pipeline.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/readable.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/state.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/transform.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/utils.js
CJS module file:///Users/ib/dev/deno_std/node/internal/streams/writable.js
CJS module file:///Users/ib/dev/deno_std/node/internal/timers.js
CJS module file:///Users/ib/dev/deno_std/node/internal/util.js
CJS module file:///Users/ib/dev/deno_std/node/internal/util/inspect.js
CJS module file:///Users/ib/dev/deno_std/node/internal/validators.js
CJS module file:///Users/ib/dev/deno_std/node/stream/promises.js
CJS module file:///Users/ib/dev/deno_std/node/util/types.js
```
While in fact all of these files are ES modules. This doesn't happen if sources are pulled from `deno.land/std/`. Any idea what might be going wrong here?

kitsonk: The only way they get marked as [`CommonJS` is via the resolver I believe](https://github.com/denoland/deno_graph/blob/20ae69f4eb3646e48845ec09922e2df792261d01/src/graph.rs#L357-L362), so I would look at logging out calls to the resolver and see why the resolver is returning a CommonJS resolve result.

bartlomieju: Ah of course 🤦 in this case they are marked as `CommonJs` because they are local files and have `.js` extension (instead of `.mjs` or relevant `type` entry in `package.json`). This is going to be a bit problematic in the long run, I feel like we might be "safer" to change these files in `deno_std` to use `.mjs` extension.

cli/proc_state.rs
```diff
@@ -560,6 +595,101 @@ impl ProcState {
...
+  ///
+  /// For all discovered reexports the analysis will be performed recursively.
+  ///
+  /// If successful a source code for equivalent ES module is returned.
+  async fn translate_cjs_to_esm(
+    &self,
+    specifier: &ModuleSpecifier,
+    code: String,
+    media_type: MediaType,
+  ) -> Result<String, AnyError> {
```
ry, line 610: Why is this a method on ProcState? I see that it uses `self.file_fetcher` but otherwise it seems independent of ProcState. Consider moving this into `cli/compat/` as a standalone function that take a file_fetcher argument.

bartlomieju: Done

ry (approved):
LGTM!

Issue #13472, opened 2022-01-23
lsp: not showing `X-Deno-Warning` as a diagnostic


kitsonk:
At some point when we were refactoring things, this was lost, because the language sever never accesses the "pre-redirected" warning header, only the final redirected specifier's header which doesn't contain the value.

Issue #12864, opened 2021-11-22
lsp: Use header X-Pinned-Url for code action to pin a dependency


ewired:
Using data from the header X-Pinned-Url (e.g. Skypack), add a code action for unpinned dependencies to make them pinned to the current version:
```js
import puppeteer from "https://deno.land/x/puppeteer/mod.ts";
// after applying quick fix "Replace with pinned module":
import puppeteer from "https://deno.land/x/puppeteer@9.0.2/mod.ts";
```

kitsonk:
Good idea... @ry and I recently were lamenting how sometimes dependencies change (or type definitions) and it is a pain to ferret out the pinned URL.

This should be fairly straight forward to implement for a community contribution if someone wanted to try, otherwise I will get around to it in a while.

ntedgi:
hi @kitsonk  i would like to give it a try

kitsonk:
Currently, skypack pinned URLs don't support the `?dts` query parameter, which means that if we converted to the pinned URL, the type definitions would go away, which would make people 😢 .  I opened https://github.com/skypackjs/skypack-cdn/issues/251 to see if it can be supported.  Really need that before we progress this feature.

kitsonk:
I am going to close this in lieu of #13580.  #13580 is a more general feature, where if a URL is redirected, the language server will issue an info diagnostic and supports a quick fix to convert it to the redirected URL and doesn't require the remote host to provide any special headers.

PR #13554, opened 2022-02-01
fix(lsp): properly display x-deno-warning with redirects

kitsonk:
Fixes: #13472

The best way to deal with this was to refactor out the metadata to a seperate structure.  It makes extending this a lot easier in the future (specifically adding #12864 in a follow-up PR).

There is no new test, because I changed the test_util server to behave more like we see in practice, which broke the existing test, which then the PR fixes.


cli/lsp/documents.rs
```diff
@@ -664,6 +671,7 @@ impl SpecifierResolver {
 #[derive(Debug, Default)]
 struct FileSystemDocuments {
   docs: HashMap<ModuleSpecifier, Document>,
+  metadata: HashMap<ModuleSpecifier, Arc<HashMap<MetadataKey, String>>>,
```
dsherret, line 674: I think this specifier metadata and #12864 are not really document related, but rather have to do with specifiers. Also, `FileSystemDocuments` contains only non-open documents.

Maybe we should move this out to a separate struct on its own and then I think we'd only need to use that struct in diagnostics.rs? (So we could remove most of the code in this file)

dsherret (changes requested):

cli/lsp/documents.rs
kitsonk: I disagree... they are very much "FileSystemDocuments" as they are what is stored in the cache... while we don't load 100% of files in the cache as document nor do 100% of documents that are loaded get their meta data loaded, they both are representations of information in the cache and have a coupled shared state.

Documents open in the editor for editing can't have metadata, only `FileSystemDocuments` can (and only those in the cache).

dsherret: I think they should share the same underlying cache, but we should separate this out from `FileSystemDocuments`. The purpose of `FileSystemDocuments` is to prevent having multiple copies of the same `Document` loaded into memory.

I would also say this shouldn't be on the interface for `Documents` either. We're wanting to know what the metadata is for certain specifiers and not documents. For example, there could be two specifiers that point at the same document and only one of those specifiers would produce a certain diagnostic. Again, they would share the same underlying cache.

Maybe I'm misunderstanding the purpose of this though.

cli/lsp/cache.rs
```diff
@@ -117,3 +123,97 @@ impl CacheServer {
...
+#[derive(Debug, Clone)]
+struct Metadata {
+  values: Arc<HashMap<MetadataKey, String>>,
+  version: Option<String>,
+}
+
+#[derive(Debug, Default, Clone)]
+pub(crate) struct CacheMetadata {
+  cache: http_cache::HttpCache,
+  metadata: Arc<Mutex<HashMap<ModuleSpecifier, Metadata>>>,
```
dsherret, line 167: This change highlights that the cache being disconnected from the data in the `Arc<Mutex<_>>` is a little sketch for the sake of the snapshots, but this state being in sync is currently enforced by the "lsp mutex". In the future, if we allow multiple threads in, this state being in sync should be further enforced by requiring mutability on `set_location`, so we're good.

We may consider some improvements to this in the future to ensure the two are forced to stay in sync when a snapshot occurs so the code is slightly more resilient to changes around it, but this could be a very minor future improvement.

dsherret (approved):
Looks great! Thanks! 🙂

PR #13555, opened 2022-02-01
feat(ext/console): better circular information in object inspection

crowlKats:
example code:
```ts
const x = { a: {}, b: {}, foo: { safdalskdma: "fijnsiuesfniuw", hello: 1 } };
x.a.x = x;
x.b.y = x;
console.log(x);

const y = { a: { b: {} } };
y.a.b = y.a;
console.log(y);
```

Before
Screenshot 2022-02-01 at 02 19 16

After
Screenshot 2022-02-01 at 02 20 17

going to look into adding tests later


ext/console/02_console.js
```diff
@@ -624,10 +641,9 @@
         }
 
         if (ctxHas(value)) {
-          // Circular string is cyan
-          return cyan("[Circular]");
+          return handleCircular(value);
         }
-        return inspectObject(value, level, inspectOptions);
+        return inspectObject(value, level, inspectOptions, circular);
```
kt3k, line 646: The 4th arg doesn't seem being used?

kt3k:
The behavior looks good to me. We need to update test cases `consoleTestStringifyCircular`, etc.

ext/console/02_console.js
crowlKats: Whoops, that was remainder from some testing

ext/console/02_console.js
```diff
@@ -884,6 +900,7 @@
   }
 
   function inspectError(value) {
+    // TODO(@crowlKats): use circular system
```
bartlomieju, line 903: Can this TODO be addressed before landing?

crowlKats: yes, will do

cli/tests/unit/console_test.ts
```diff
@@ -392,14 +392,14 @@ Deno.test(function consoleTestStringifyFunctionWithProperties() {
   assertEquals(
     stringify({ f }),
     `{
-  f: [Function: f] {
+  f: [Function: f] <ref *1> {
```
crowlKats, line 395: I am not happy with this. In node the ref is before `[Function: f]` or any other description.
This is also inconsistent behaviour as
```js
class Foo {}
const x = new Foo();
x.y = x;
console.log(x);
```
Will output `<ref *1> Foo { y: [Circular *1] }`

bartlomieju: Can't `<ref *1>` be prepended in this case?

kt3k: Maybe let's make `inspectRawObject` returning `[inspected, refIndex]` array, and prepend `<ref *1>` part inside of `inspectFunction`? (around here: https://github.com/denoland/deno/blob/26f5c223df26962f19ccc957d0507facc09c55c5/ext/console/02_console.js#L358-L362

cli/tests/unit/console_test.ts
```diff
@@ -350,12 +350,23 @@ Deno.test(function consoleTestStringifyCircular() {
         return Deno.inspect(this);
       },
     }),
-    "[Circular]",
+    "[Circular *1]",
```
crowlKats, line 353: In this cause, shouldnt it rather be `<ref *1> [Circular *1]`?

kt3k (approved):
LGTM

PR #13565, opened 2022-02-01
perf(lsp): cancellable TS diagnostics

dsherret:
Adds the ability to cancel generating TypeScript diagnostics while they're being generated. I'm not sure how to test this to see what kind of improvement this may bring, but I assume it will be helpful.

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -217,6 +217,7 @@ impl DiagnosticsServer {
                     snapshot.clone(),
                     &config,
                     &ts_server,
+                    token.clone(),
                   )
                   .await
                   .map_err(|err| {
@@ -494,6 +495,7 @@ async fn generate_ts_diagnostics(
   snapshot: Arc<language_server::StateSnapshot>,
   config: &ConfigSnapshot,
   ts_server: &tsc::TsServer,
+  token: CancellationToken,
 ) -> Result<DiagnosticVec, AnyError> {
   let mut diagnostics_vec = Vec::new();
   let specifiers = snapshot
@@ -508,7 +510,9 @@ async fn generate_ts_diagnostics(
     .partition::<Vec<_>, _>(|s| config.specifier_enabled(s));
   let ts_diagnostics_map: TsDiagnosticsMap = if !enabled_specifiers.is_empty() {
     let req = tsc::RequestMethod::GetDiagnostics(enabled_specifiers);
-    ts_server.request(snapshot.clone(), req).await?
+    ts_server
+      .request_with_cancellation(snapshot.clone(), req, token)
+      .await?
   } else {
     Default::default()
   };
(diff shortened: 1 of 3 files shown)


cli/tsc/99_main_compiler.js
```diff
@@ -698,6 +698,10 @@ delete Object.prototype.__proto__;
           /** @type {Record<string, any[]>} */
           const diagnosticMap = {};
           for (const specifier of request.specifiers) {
+            if (core.opSync("op_is_cancelled", {})) {
```
dsherret, line 701: How expensive is this call? It might be good to throttle this and only check every X milliseconds (that seems to be what TypeScript does in tsserver). Let me know if you think I should implement that as part of this PR.

cli/tsc/99_main_compiler.js
```diff
@@ -698,6 +698,10 @@ delete Object.prototype.__proto__;
...
+              break;
+            }
+
             diagnosticMap[specifier] = fromTypeScriptDiagnostic([
               ...languageService.getSemanticDiagnostics(specifier),
               ...languageService.getSuggestionDiagnostics(specifier),
```
dsherret, line 707: The language service has a cancellation API, but it doesn't really work like how I would expect a cancellation API to work. It migth be good if we use it in the future. I've asked some questions on the TS discord about it.

dsherret: Actually, I'm just going to implement using their cancellation API this with the throttling. I'll check more what they do in tsserver.

cli/lsp/diagnostics.rs
```diff
@@ -217,6 +217,7 @@ impl DiagnosticsServer {
                     snapshot.clone(),
                     &config,
                     &ts_server,
+                    token.clone(),
```
dsherret, line 220: With this change we might want to consider removing or reducing the 200ms delay on line 208.

kitsonk: there is also antidotal evidence too by using a debug build and editing a large project in vscode... I can try to take it for a spin later... though if we drop that, we should do it as a follow up PR, I suspect.

dsherret: That's a good idea! Though I just tested it now and removing the delay is still much slower ☹

Before (this PR):

```
Starting Deno benchmark
-> Start benchmarking lsp
   - Simple Startup/Shutdown
      (10 runs, mean: 508ms)
   - Big Document/Several Edits 
      (5 runs, mean: 1879ms)
   - Find/Replace
      (10 runs, mean: 78ms)
   - Code Lens
      (10 runs, mean: 612ms)
<- End benchmarking lsp
```

After (debounce removed):

```
Starting Deno benchmark
-> Start benchmarking lsp
   - Simple Startup/Shutdown 
      (10 runs, mean: 295ms)
   - Big Document/Several Edits 
      (5 runs, mean: 4668ms)
   - Find/Replace
      (10 runs, mean: 78ms)
   - Code Lens
      (10 runs, mean: 402ms)
<- End benchmarking lsp
```

kitsonk: The test might not be triggering it, because I don't think it is sending and setting cancellation tokens. I think we would need to modify the test, based on observed messages to and from the server.

bartlomieju (approved):
LGTM implementation wise, but I'm unsure about the debouncing. I'll leave that to Kit

PR #13566, opened 2022-02-02
WIP: child process IPC

piscisaureus:


josh-hemphill:
I noticed this is Unix-only; are there blockers for implementing it for Windows, or is it magnitudes more complicated? Curious if there's anything I can do to help. I've been looking at using the node compat mode to make a deno bundle for nodejs MS SQL tedious driver, and IPC pollyfil seems to be one of few remaining hangups, and most people using the MS SQL driver will be doing things with Windows.

PR #13567, opened 2022-02-02
Fix uncaught TypeError in ext/console (css styling)

### LINKED ISSUE #13469: Uncaught `TypeError` in `ext/console`
## Bug Report

`console.debug` throws an error when it is used by the `debug` package that is skypack'ed from npm.

```
error: Uncaught TypeError: (intermediate value)(intermediate value)(intermediate value) is not iterable
      logFn.apply(self2, args);
            ^
    at cssToAnsi (deno:ext/console/02_console.js:1591:27)
    at inspectArgs (deno:ext/console/02_console.js:1703:32)
    at console.debug (deno:ext/console/02_console.js:1803:9)
    at debug (https://cdn.skypack.dev/-/debug@v4.3.3-0XITH0hRj2rH5NQ5xJ30/dist=es2019,mode=imports/optimized/debug.js:270:13)
    at file:///tmp/test.ts:5:1
```

### Reproduction Steps

Copy the following code to a file.

```ts
import d from "https://cdn.skypack.dev/debug@^4.3.3";
d.useColors = () => !Deno.noColor;
d.enable("demo");
const debug = d("demo");
debug("test");
```

Make sure your deno version is `1.17.2` or higher.

Run the created file with `deno run`. This will lead to the above stack trace.

You can verify that `1.17.1` is able to run the above code correctly. The ouput is:
```
demo test +0ms
```

### Offending Pull Request

The PR #13175 introduces this bug.

zachauten:
When using css coloring in the console, non-color values should be ignored rather than throw exceptions.
Fixes #13469

### DIFF
--- a/cli/tests/unit/console_test.ts
+++ b/cli/tests/unit/console_test.ts
@@ -989,6 +989,7 @@ Deno.test(function consoleTestWithStyleSpecifier() {
 });
 
 Deno.test(function consoleParseCssColor() {
+  assertEquals(parseCssColor("inherit"), null);
   assertEquals(parseCssColor("black"), [0, 0, 0]);
   assertEquals(parseCssColor("darkmagenta"), [139, 0, 139]);
   assertEquals(parseCssColor("slateblue"), [106, 90, 205]);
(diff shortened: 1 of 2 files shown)


kt3k (commented):
@zachauten 
Thanks for the contribution!

Confirmed this fixes the bug. Nice!

But I think we need another test case like the below to check an invalid color doesn't cause the uncaught error:

```ts
Deno.test(function consoleLogShouldNotThrowErrorWhenInvalidCssColorsAreGiven() {
  mockConsole((console, out) => {
    console.log("%cfoo", "color: foo; background-color: bar;");
    assertEquals(out.toString(), "foo\n");
  });
});
```

lucacasonato:
@zachauten Do you want to add that test case, or finish this PR up for you?

zachauten:
@kt3k @lucacasonato Added!

kt3k (approved):
@zachauten Thanks for the update! LGTM

PR #13577, opened 2022-02-02
refactor(lsp): remove circular dependency between `LanguageServer` and `DiagnosticsServer`

dsherret:
Was looking through the code for more deadlocks and did some refactoring to remove the circular dependency between `LanguageServer` and `DiagnoticsServer`. To be clear, I don't see any deadlock potential in this code, but removing the circular dependency will reduce the chance of code changes in the future causing a deadlock.

The circular dependency is removed by the language server sending a snapshot to the diagnostics server when causing a diagnostics update.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -84,6 +84,9 @@ pub(crate) struct StateSnapshot {
   pub root_uri: Option<Url>,
 }
 
+pub(crate) type LanguageServerSnapshot =
+  (Arc<StateSnapshot>, Arc<ConfigSnapshot>, Option<LintConfig>);
+
 #[derive(Debug)]
 pub(crate) struct Inner {
   /// Cached versions of "fixed" assets that can either be inlined in Rust or
@@ -790,9 +793,10 @@ impl Inner {
     if document.is_diagnosable() {
       self
         .diagnostics_server
-        .invalidate(self.documents.dependents(specifier))
-        .await;
-      if let Err(err) = self.diagnostics_server.update() {
+        .invalidate(self.documents.dependents(specifier));
+      if let Err(err) =
+        self.diagnostics_server.update(self.diagnostics_snapshot())
+      {
         error!("{}", err);
       }
     }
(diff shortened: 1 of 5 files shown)


cli/lsp/language_server.rs
```diff
@@ -2364,14 +2352,29 @@ impl lspower::LanguageServer for LanguageServer {
   }
 
   async fn did_open(&self, params: DidOpenTextDocumentParams) {
+    if params.text_document.uri.scheme() == "deno" {
+      // we can ignore virtual text documents opening, as they don't need to
+      // be tracked in memory, as they are static assets that won't change
+      // already managed by the language service
+      return;
+    }
```
dsherret, line 2360: I think this was an issue I caused in the previous config refactor and this should have been moved here.

cli/tests/integration/lsp_tests.rs
```diff
@@ -490,6 +493,7 @@ fn lsp_import_assertions() {
       }),
     )
     .unwrap();
+  handle_configuration_request(&mut client);
```
dsherret, line 496: This test needed to read the configuration request for the `.js` file. Previously it was only reading it for the `.json` file in the `did_open` call below.

cli/lsp/language_server.rs
```diff
@@ -755,15 +756,8 @@ impl Inner {
     &mut self,
     specifier: &ModuleSpecifier,
     params: DidOpenTextDocumentParams,
-  ) {
+  ) -> Document {
     let mark = self.performance.mark("did_open", Some(&params));
-
-    if params.text_document.uri.scheme() == "deno" {
```
dsherret, line 761: Some of the code in this method was moved to the parent method on line 2354.

test_util/src/lsp.rs
```diff
@@ -122,7 +122,16 @@ where
...
+    Some(params) => {
+      Some(serde_json::from_value(params.clone()).map_err(|err| {
+        anyhow::anyhow!(
+          "Could not deserialize message '{}': {}\n\n{:?}",
+          method,
+          err,
+          params
+        )
+      })?)
+    }
```
dsherret, line 134: This was to get better information while debugging so I left it in because it's probably useful in the future.

kitsonk (approved):
LGTM, this isn't just removing a circular dependency, it is a good all around cleanup!

PR #13580, opened 2022-02-03
feat(lsp): add redirect diagnostic and quick fix

kitsonk:
Ref: #12864

This is a more general solution to #12864, which I think I like better.  It also "works" for pinning URLs with the Deno warnings like `deno.land/x` and `deno.land/std`.

Basically if a specifier is redirected, a quick fix will be available to replace the import with the redirected URL.  For example:

https://user-images.githubusercontent.com/1282577/152283410-ff95954d-b2ca-4a93-8f04-e623e2e50ffa.mp4

### DIFF
--- a/cli/lsp/diagnostics.rs
+++ b/cli/lsp/diagnostics.rs
@@ -570,15 +570,21 @@ pub(crate) enum DenoDiagnostic {
   InvalidAssertType(String),
   /// A module requires an assertion type to be a valid import.
   NoAssertType,
-  ///
+  /// A remote module was not found in the cache.
   NoCache(ModuleSpecifier),
+  /// A blob module was not found in the cache.
   NoCacheBlob,
+  /// A data module was not found in the cache.
   NoCacheData(ModuleSpecifier),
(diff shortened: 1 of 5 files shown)


cli/lsp/diagnostics.rs
```diff
@@ -546,29 +548,186 @@ async fn generate_ts_diagnostics(
...
+/// An enum which represents diagnostic errors which originate from Deno itself.
+pub(crate) enum DenoDiagnostic {
+  /// A `x-deno-warn` is associated with the specifier and should be displayed
+  /// as a warning to the user.
+  DenoWarn(String),
+  /// The import assertion type is incorrect.
+  InvalidAssertType(String),
+  /// A module requires an assertion type to be a valid import.
+  NoAssertType,
+  ///
```
dsherret, line 573: ​

dsherret (approved):
Looks great and this is a big user experience improvement.

Personally, this will allow me to not feel bad about using non-versioned specifiers in readmes 🙂

PR #13585, opened 2022-02-03
chore: improve error messages in CompressionStream

lucacasonato:
This commit makes the error messages that one sees when passing
something other than a BufferSource to a (De)CompressionStream. The
WPT tests already pass, because they just check for error type
(TypeError), and not error message. A TypeError was already thrown for
invalid values via serde_v8.


crowlKats (approved):
LGTM

PR #13590, opened 2022-02-04
chore: rm toplevel sanitizer_test.ts

littledivy:
Accidentally introduced in https://github.com/denoland/deno/pull/13296 ?


lucacasonato (approved):
Yikes. LGTM

PR #13600, opened 2022-02-05
feat(test): Allow timeout option for catching asynchronous long-running tests

### LINKED ISSUE #11133: Deno.test() should support timeout option
`Deno.TestDefinition` should allows to specify a timeout for a test case:

```ts
Deno.test({
   name: "some net test",
   timeout: 5000 // in ms
   async fn() {
      //
   }
});
```

lucasfcosta:
# Summary

This PR closes #11133 by allowing users to specify a `timeout` for their tests.

One can specify a timeout in the following ways:

```ts
// Usual form: name, function, and timeout
Deno.test("Usual test form", () => {
  const x = 1 + 2;
  assertEquals(x, 3);
}, 5000);

// Compact form: named function, and timeout.
Deno.test(function helloWorld3() {
  const x = 1 + 2;
  assertEquals(x, 3);
}, 5000);

// Longer form: using an options object.
Deno.test({
  name: "hello world #2",
  timeout: 5000,
  fn: () => {
    const x = 1 + 2;
    assertEquals(x, 3);
  },
});
```

The timeout is always optional and defaults to `5000` ms.

⚠️ Please notice that this PR only addresses asynchronous timeouts, [as we discussed on Discord](https://github.com/denoland/deno/issues/11133#issuecomment-1008327698).

**We'll address timing-out synchronous tests in a separate PR as that require further changes, as described below**.

# Implementation details

> ☝️  Given this PR only addresses timing-out tests during asynchronous operations, most of the changes here are done on the JS side of things.

When a `timeout` option exists (and when dealing with an `AsyncFunction`), we'll wrap the testing function into a function that starts a timer using `setTimeout` and runs the test function. If that timer is due to execute before the testing function finishes, it'll throw an error.

Regardless of whether the test succeeds, once it's done, we'll clear the timer we've set up.

## How does this code behave with the sanitizers?

The `op` sanitizer will already wait for the cleared timer to resolve, as shown in the piece of code below. Therefore, we will not have a hanging `op` after the test finishes.

```ts
// Defer until next event loop turn - that way timeouts and intervals
// cleared can actually be removed from resource table, otherwise
// false positives may occur (https://github.com/denoland/deno/issues/4591)
await opSanitizerDelay();
``` 

We would, however, have problems if we have tests which do not have `sanitizeOps` turned on, because in that case we would not allow for the event loop to clear and the next test would start with a leaking op from the previous test (if the previous test was async).

Take the test file below as an example. In that file, without calling `sanitizeOps` after clearing the timeout timer, the second test would fail. It would fail because `sanitizeOps` wouldn't have been called, and thus the timer kicked by the test wouldn't be allowed to complete. Therefore, the timeout timer would "leak" to the next test.

```ts
// Without calling `sanitizeOps` after clearing the timeout timer, the second test fails

Deno.test({
  name: "would cause the timeout timer to leak",
  fn: async () => {},
  sanitizeOps: false,
});

Deno.test({
  name:
    "I would fail because the previous test didn't wait for ops to complete",
  fn: async () => {},
});
```

That's why I also had to call `opSanitizerDelay` after clearing the timeout. That call prevents the timer from leaking into the next test and causing it to fail in case it needs to sanitize ops. By the way, that behaviour is already checked by the `steps_passing_steps` test.

I also added an extra test to cover this case.

# Changes to existing tests

There were tests in which we were checking for a particular number of operations to match. In those tests, the operations would _not_ match if there wasn't a `timeout` explicitly defined for the test. Therefore, I had to make sure to pass `null` to those tests so that these test callbacks were not wrapped into the timeout function.

# Upcoming work

As previously mentioned, there are a few items I want to tackle once I get this PR merged in:

1. I would like to be able to timeout individual synchronous tests.
    For that, however, we'd have to do some significant refactoring in order to allow for the runner to execute each test in a separate thread. Without that, we would have to timeout entire files and thus not run any tests after the one which is performing blocking operations.
    An alternative suggestion was to simply alert users when a single test has taken more than the specified time to run and accept that no further tests would run.
    We could also simply rerun the same test file and skipped the test which was blocking the last time, but I think that could be confusing and would not lead to clean code/flow.
2. [I've seen we currently do not track `unrefed` ops](https://github.com/denoland/deno/blob/592b1fb911765802fd5d6cb774995f2617dec59c/core/ops_metrics.rs#L14-L20). Tracking those `ops` could yield benefits to this timeout flow because it allows us to `unref` the timer we're starting and then we can use that information in our sanitizer calculations.

Those two items are not necessary for now though. I just wanted to keep track of them somewhere.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -157,6 +157,14 @@ declare namespace Deno {
     /** If at least one test has `only` set to true, only run tests that have
      * `only` set to true and fail the test suite. */
     only?: boolean;
+    /** Determines after how many millisseconds when the test should timeout.
+     *
+     * Please notice that Deno will not timeout tests during blocking synchronous
+     * operations.
+     *
+     * Defaults to 5000.
+     */
+    timeout?: number | null;
     /** Check that the number of async completed ops after the test is the same
      * as number of dispatched ops. Defaults to true. */
     sanitizeOps?: boolean;
@@ -362,6 +370,7 @@ declare namespace Deno {
   export function test(
     name: string,
     fn: (t: TestContext) => void | Promise<void>,
+    timeout?: number | null,
   ): void;
 
   /** Register a test which will be run when `deno test` is used on the command
@@ -382,7 +391,10 @@ declare namespace Deno {
    * });
    * ```
    */
-  export function test(fn: (t: TestContext) => void | Promise<void>): void;
+  export function test(
+    fn: (t: TestContext) => void | Promise<void>,
+    timeout?: number | null,
+  ): void;
 
   /** Register a test which will be run when `deno test` is used on the command
    * line and the containing module looks like a test module.
--- a/cli/tests/integration/test_tests.rs
+++ b/cli/tests/integration/test_tests.rs
@@ -227,6 +227,18 @@ itest!(aggregate_error {
   output: "test/aggregate_error.out",
 });
 
+itest!(async_timeout {
+  args: "test test/async_timeout.ts",
+  exit_code: 1,
+  output: "test/async_timeout.out",
+});
+
+itest!(async_timeout_not_leaking_without_sanitizer {
+  args: "test test/async_timeout_not_leaking_without_sanitizer.ts",
+  exit_code: 0,
+  output: "test/async_timeout_not_leaking_without_sanitizer.out",
+});
+
 itest!(steps_passing_steps {
   args: "test test/steps/passing_steps.ts",
   exit_code: 0,
--- a/cli/tests/testdata/test/async_timeout.out
+++ b/cli/tests/testdata/test/async_timeout.out
@@ -1,0 +1,21 @@
+Check [WILDCARD]/testdata/test/async_timeout.ts
+running 3 tests from [WILDCARD]/testdata/test/async_timeout.ts
+test test 1 ... ok ([WILDCARD])
+test test 2 ... FAILED ([WILDCARD])
+test test 3 ... ok ([WILDCARD])
+
+failures:
+
+test 2
+Error: Test took longer than the 5000ms timeout.
+    at deno:runtime/js/40_testing.js:1030:18
+    at Object.action (deno:ext/timers/01_timers.js:162:11)
+    at handleTimerMacrotask (deno:ext/timers/01_timers.js:79:12)
+
+failures:
+
+	test 2
+
+test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out ([WILDCARD])
+
+error: Test failed
--- a/cli/tests/testdata/test/async_timeout.ts
+++ b/cli/tests/testdata/test/async_timeout.ts
@@ -1,0 +1,7 @@
+Deno.test("test 1", () => {});
+
+Deno.test("test 2", async () => {
+  await new Promise(() => {});
+});
+
+Deno.test("test 3", () => {});
--- a/cli/tests/testdata/test/async_timeout_not_leaking_without_sanitizer.out
+++ b/cli/tests/testdata/test/async_timeout_not_leaking_without_sanitizer.out
@@ -1,0 +1,7 @@
+Check [WILDCARD]/testdata/test/async_timeout_not_leaking_without_sanitizer.ts
+running 2 tests from [WILDCARD]/testdata/test/async_timeout_not_leaking_without_sanitizer.ts
+test will not rely on sanitizeOps to sanitize the test's timeout ... ok ([WILDCARD])
+test will pass because the previous test can still clear its own timeout without sanitizeOps ... ok ([WILDCARD])
+
+test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out ([WILDCARD])
+
--- a/cli/tests/testdata/test/async_timeout_not_leaking_without_sanitizer.ts
+++ b/cli/tests/testdata/test/async_timeout_not_leaking_without_sanitizer.ts
@@ -1,0 +1,11 @@
+Deno.test({
+  name: "will not rely on sanitizeOps to sanitize the test's timeout",
+  fn: async () => {},
+  sanitizeOps: false,
+});
+
+Deno.test({
+  name:
+    "will pass because the previous test can still clear its own timeout without sanitizeOps",
+  fn: async () => {},
+});
--- a/cli/tests/unit/metrics_test.ts
+++ b/cli/tests/unit/metrics_test.ts
@@ -54,7 +54,7 @@ Deno.test(
 );
 
 Deno.test(
-  { permissions: { write: true } },
+  { permissions: { write: true }, timeout: null },
   async function metricsUpdatedIfNoResponseAsync() {
     const filename = Deno.makeTempDirSync() + "/test.txt";
 
--- a/cli/tests/unit/read_file_test.ts
+++ b/cli/tests/unit/read_file_test.ts
@@ -154,7 +154,7 @@ Deno.test(
 );
 
 Deno.test(
-  { permissions: { read: true, write: true } },
+  { permissions: { read: true, write: true }, timeout: null },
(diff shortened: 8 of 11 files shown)


lucacasonato:
Hey @lucasfcosta. Thanks for the work on this. I think the implementation looks good overall, but I have a somewhat more philosophical question for this. How does this interact with the resource sanitizer?

An example to illustrate the issue:

```js
Deno.test("hello", { timeout: 5000 }, async () => {
  await new Promise(res => setTimeout(res, 10000));
});
```

The above test will time out after 5 seconds, but it will then subsequently leak a timer resource. This will cause broken follow up tests at best, and flaky follow up tests at worst. I think this is a real footgun.

I think this is different from the indeterminate state of follow ups after errors that already exists right now, because in that case code execution of the earlier test is terminated in most cases (because the exception bubbled up).

With timeouts, code is still executing after the timeout fires. This may be very strange to people.

I don't have a good solution to this, and maybe this tradeoff is fine, but I want to throw some ideas out here:

- Should timeouts affect each test, or the entire module as a whole? For the latter, we can terminate execution of that test module after timeout, preventing this issue.
- If we stick with per test timeouts, should we fail fast (run no follow up tests) the module after the timeout occurs? We can then also terminate the entire test module after timeout. This would also prevent the issue.

lucasfcosta:
Hey @lucacasonato, thanks a lot for your input 😊 

The example of the leaking `setTimeout` was an excellent catch and indeed a serious situation to consider. Thank you for noticing that.

Given what you've said, I think we have two questions to answer:

1. Whether the current behaviour is consistent or we need to change it
2. Whether we'll need more features to improve handling the situation you described

# 1. Analyzing the current behaviour

Regarding analyzing whether the current behaviour is consistent, my view is that the situation you describe is a resource and operations leak. Therefore, first and foremost, **we should ensure that sanitizing both asynchronous and synchronous tests leads to consistent results**.

Take the following test file, for example:

``` 
Deno.test("first", () => {
  console.log(Deno.resources());
  setTimeout(() => {}, 2000);
  console.log(Deno.resources());
});

Deno.test("second", () => {
  console.log(Deno.resources());
});
```

In this test file, you'll see that the first test creates a resource that hangs. Therefore, it causes the `first` test to fail. On the other hand, the `second` test still runs just fine, even though the hanging timeout from the first test could've had unexpected side effects and/or made the second test flaky.

Now, let's see what happens when we have the same situation, but with a test that times out:

```ts
Deno.test("first", { timeout: 1000 }, async () => {
  console.log(Deno.resources());
  const timerPromise = new Promise((resolve) => setTimeout(resolve, 2000));
  console.log(Deno.resources());
  await timerPromise;
});

Deno.test("second", () => {
  console.log(Deno.resources());
});
```

In this situation and the previous one, the `first` test fails, and the second still runs and succeeds despite the possibility for the existing hanging timer to interfere with subsequent tests.

The only difference between these two scenarios is that the error message in the first scenario mentions hanging resources, while the error message in the second scenario mentions a timeout.

> 👉 We don't throw a sanitizer-related error because [**we only sanitize resources if a test succeeds**](https://github.com/denoland/deno/blob/4799aaac15285833341d1e0471a2559bd325982f/runtime/js/40_testing.js#L274). Also, it isn't that useful to throw a sanitizer-related error if the test has already failed.
> 👉 The same happens with operations. We do know that there have been leaking operations, but we do not throw an error mentioning leaking ops because the test has already failed at that time.

**Considering the above, I believe that the current behaviour is consistent with what currently happens when there are hanging timers**. Consequently, I'd prefer not to change this PR's current behaviour.

# 2. Proposed way forward

Even though the current behaviour is consistent with this PR's behaviour, I do agree that we need changes to prevent hanging resources/leaking ops from interfering with subsequent tests.

In my view, the runner should treat each test as an individual unit of execution. Therefore, the failure of a single test shouldn't interfere with others in the same file. I think so because it's very frustrating not to know the results of other tests I may have to debug because a previous test failed. Furthermore, separating tests into multiple files is an arbitrary decision, and I'd be a bit surprised to learn that the runner is using test files as a unit of execution.

Finally, a third reason for thinking of tests as an individual execution unit is that it allows for more flexible scheduling and handling, as it would be necessary for timing out synchronous tests (as described in #13609).

# 3. Final recommendation

Considering the two sections above, my recommendation is that:

1. We do not change the current PRs behaviour to keep its scope small (considering its behaviour is already consistent with the existing runner's)
2. We create a separate issue to ensure hanging timers wouldn't interfere with subsequent tests (which would probably end up being a consequence of #13609 anyway).

❓ One thing we could discuss, however, is whether we'd prefer to throw sanitizer errors rather than timeout errors in this situation. I don't think we should, considering that a timeout indicates an issue that needs to be solved _before_ leaking resources/ops anyway, but I'm open to suggestions, of course.

lucacasonato:
I don't think I clarified my actual concern well enough. I am aware of the behavior being the same in so far that a previous failing test (either timeout or exception) can leak existing resources into other tests.

My concern is that when a test times out, **the code keeps executing**. This is different when an exception occurs. The code is done executing, because the exception bubbled up.

The problem with the continued execution is that it results in very un-intuitive behavior:

```js
Deno.test("first", { timeout: 1000 }, async () => {
  await new Promise(res => setTimeout(res, 2000);
  const file = await Deno.open("./README.md");
  await new Promise(res => setTimeout(res, 3000);
  file.close();
});

Deno.test("second", async () => {
  await new Promise(res => setTimeout(res, 2000);
});
```

Now the second test doesn't just fail with a leaking timer resource (which is already confusing), but it fails with a leaking file resource, which makes even less sense to the user, because that file should have never been opened as the timeout occurred.

### I.e. it is not intuitive that code in the test function continues executing after the timeout marks the test as "failed".

This is especially important for synchronous timeouts. In synchronous environments, the code _has_ to be terminated. You can not leave the sync code running and just mark the test as skipped, because if there is some sync code blocking execution then no other tests can be run, because they are blocked on the sync code. It would be confusing if async and sync environments behaved differently in this regard of termination.

---

As for the unit of execution being the test: it currently isn't, and it would probably be a backwards incompatible change to change this. The current unit of execution is a module. This also has some desirable effects: namely that global scope only executes once per module, rather than once per test. The latter can be very confusing.

lucasfcosta:
Hey @lucacasonato, thanks for taking the time to explain 🙌 . I'm sorry I've written such a long explanation for the misunderstanding 😅 

With regards to the problem you describe, I strongly agree with your assessment here:

> It would be confusing if async and sync environments behaved differently in this regard to termination.

Considering we'd like the above to be true, I think indeed the only choices would be to (1) either make individual tests the runner's unit of execution or simply (2) not to run the rest of the tests in a file if one of them times out.

IMO (2) could be a bit frustrating as that's different from most runners out there and could make debugging/fixing take longer. On the other hand, what you said about executing the global scope only once per module being less confusing also makes a lot of sense. Personally, I'm slightly more inclined towards (1), but I think your opinion here would be much more valuable as you're more well-informed given you've been much more involved with the project.

What's your preference here?

Do you have any thoughts on what should be Deno's philosophy with regards to testing? I think this might be a case of thinking whether we'd like to give users control and be more hands-off (1) or take control and prevent mistakes ourselves(2).

Please let me know what you think and how you'd like to proceed in terms of creating separate issues, discussing on Discord, or this PR etc...

Thank you!

lucasfcosta:
@bartlomieju, is there anything I can do to push this forward? Would you be happy to merge this before we continue to tackle #13609?

bartlomieju:
> @bartlomieju, is there anything I can do to push this forward? Would you be happy to merge this before we continue to tackle #13609?

Hey @lucasfcosta, sorry for slow response. I discussed this topic recently with @dsherret and we have an idea how to tackle this in more generic way (essentially addressing #13609). I would prefer to try to that approach before merging this PR, to avoid introducing subtle edge cases in the behavior. I'll try to get a PR open soon and get back to you.

lucacasonato:
@lucasfcosta Sorry I never responded to your last message. To answer your question:

> What's your preference here?

I'd go with option 2: fail fast the module if a test times out.

PR #13610, opened 2022-02-06
refactor: use `Arc` instead of making copies of `Flags` struct

### LINKED ISSUE #13311: `deno::Flags` should be reference counted
The [`Flags` struct](https://github.com/denoland/deno/blob/1fb5858009f598ce3f917f9f49c466db81f4d9b0/cli/flags.rs#L211-L256) has a lot of data in it and it is cloned multiple times. It makes sense to wrap it in an `Rc`.

For example:

https://github.com/denoland/deno/blob/1fb5858009f598ce3f917f9f49c466db81f4d9b0/cli/main.rs#L378

https://github.com/denoland/deno/blob/1fb5858009f598ce3f917f9f49c466db81f4d9b0/cli/main.rs#L572

biryukovmaxim:
ProcState.Flags: Flags -> Arc<Flags>
using Arc instead of cloning

closes #13311


crowlKats:
I'd rather be for calling it `FlagsInner` instead of just `Inner`

cli/flags.rs
```diff
@@ -208,7 +210,23 @@ impl Default for CheckFlag {
 }
 
 #[derive(Clone, Debug, PartialEq, Default)]
-pub struct Flags {
+pub struct Flags(pub(crate) Arc<Inner>);
```
dsherret, line 213: I think that we don't need to change this interface? Probably we only need to wrap it in an `Arc` in a couple places downstream from this file and maybe put it in an `Arc` before it goes into the proc state. Also, it seems like we're just needlessly cloning `Flags` in a few places because we could just use `ps.flags` instead.

biryukovmaxim:
> I'd rather be for calling it `FlagsInner` instead of just `Inner`
I took an example from this implementation of Wrapper
https://github.com/denoland/deno/blob/a7850d7fe6c52225acd614167495d765ed2385f3/cli/proc_state.rs#L63
But actually I can rename Inner to FlagsInner

cli/flags.rs
```diff
@@ -389,13 +407,13 @@ impl Flags {
 impl From<Flags> for PermissionsOptions {
```
dsherret, line 407: Instead of having this as an `impl From<Flags>` we could instead introduce a `permissions_options()` method on `Flags` that returns this information from a reference of `Flags`.

biryukovmaxim: Yep, I think its a good idea

cli/flags.rs
bartlomieju: I agree, no changes should be necessary to `Flags` struct itself

cli/flags.rs
```diff
@@ -2524,10 +2542,10 @@ mod tests {
   #[test]
   fn upgrade() {
     let r = flags_from_vec(svec!["deno", "upgrade", "--dry-run", "--force"]);
-    let flags = r.unwrap();
+    let inner = Arc::try_unwrap(r.unwrap().0).unwrap();
     assert_eq!(
-      flags,
-      Flags {
+      inner,
```
ry, line 2547: can you not do just `*r.unwrap()` ?

core/module_specifier.rs
```diff
@@ -123,7 +123,7 @@ pub fn resolve_url(
 /// file path; if it is a relative path it's resolved relative to the current
 /// working directory.
 pub fn resolve_url_or_path(
-  specifier: &str,
+    specifier: &str,
```
ry, line 126: unrelated change...

cli/tools/test.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-use crate::cache;
+use crate::{cache, Flags};
```
ry, line 3: We prefer separate lines for each `use`.

cli/main.rs
```diff
@@ -38,7 +38,7 @@ mod windows_util;
 
 use crate::file_fetcher::File;
 use crate::file_watcher::ResolutionResult;
-use crate::flags::BundleFlags;
+use crate::flags::{BundleFlags, Flags};
```
ry, line 41: in Deno we prefer separate lines for each use statement.

ry (commented):
Just some small nit-picky comments - otherwise looks good! Thanks!

cli/flags.rs
dsherret: @ry I believe we'd only need to wrap it in an `Arc` when putting it in `ProcState` and then change some of the other API to not requiring cloning (ex: https://github.com/denoland/deno/pull/13610#discussion_r800233365)? Then if some other place needs that state they could do `let flags = ps.flags.clone();`?

cli/flags.rs
biryukovmaxim: Moving out  of an `Arc` is imposible, but I can replace it by somethink like that:

``
r.unwrap().0.as_ref().clone();
``

bartlomieju: If you are the only holder of the `Arc` then you can use `try_unwrap` to move out of the arc: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap

biryukovmaxim: > If you are the only holder of the `Arc` then you can use `try_unwrap` to move out of the arc: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap

and I did it at line 2545:
`let inner = Arc::try_unwrap(r.unwrap().0).unwrap();`

@bartlomieju there something I'm missing here?

bartlomieju: Ah didn't notice. Seems okay 👍

cli/flags.rs
ry: Ok sure - but we should remove the `derive(Clone)` from the Flags struct then to prevent people from copying this.

biryukovmaxim: @dsherret @bartlomieju @ry 
actually I did it. Arc only in ProcState

dsherret:
@biryukovmaxim there was a format script that needed to be run (https://github.com/denoland/deno/tree/main/tools#formatjs). I ran it and reduced a few more clones so should be good now.

cli/main.rs
```diff
@@ -526,8 +524,9 @@ async fn install_command(
   let mut preload_flags = flags.clone();
   preload_flags.inspect = None;
   preload_flags.inspect_brk = None;
```
dsherret, line 526: The `#[derive(Clone)]` on `Flags` is only still necessary because of this clone here. I wonder if there's a nice way that we could remove `#[derive(Clone)]` on `Flags` and still support this scenario.

dsherret (approved):
LGTM! Thanks @biryukovmaxim!

PR #13615, opened 2022-02-07
feat(runtime): web streams in fs & net APIs

### LINKED ISSUE #13614: Make FS and net APIs work well with WHATWG streams
Currently our FS and net APIs use "Go style" streams (`Reader` / `Writer`). While these kinds of streams are very fast and simple, they are not very well compatible with `fetch` and all our newfangled `CompressionStream` and `TextEncoderStream` utilities.

Currently `std/streams` has `readableStreamFromReader` and `writableStreamFromWriter` functions to convert from `Writer` to `WritableStream` and from `Reader` to `ReadableStream`. While these are convenient, they are not as convenient as not having to use them at all.

To solve this I am proposing adding a `readable` and `writable` property to all of our internal implementations of `Reader` and `Writer`. `readable` would return a `ReadableStream<Uint8Array>` and `writable` would return a `WritableStream<Uint8Array>`.

This makes `Deno.File` really convenient to use for use with the native HTTP server or fetch for example:

```js
const file = await Deno.open("/etc/passwd");
const resp = await fetch("https://example.com", {
  method: "POST",
  body: file.readable
});
await resp.body.pipeTo(Deno.stdout.writable);
```

This could land independently of how #11018. These two APIs can co-exist - it doesn't need to be one or the other.

lucacasonato:
This commit adds `readable` and `writable` properties to `Deno.File` and
`Deno.Conn`. This makes it very simple to use files and network sockets
with fetch or the native HTTP server.

Closes #13614


bartlomieju:
+1 from me, seems very useful

ext/net/01_net.js
```diff
@@ -59,10 +60,67 @@
     return core.opAsync("op_dns_resolve", { query, recordType, options });
   }
 
+  const DEFAULT_CHUNK_SIZE = 16_640;
```
crowlKats, line 63: Shouldnt this be 16384?

lucacasonato: This is copied from std/streams.

crowlKats: huh...

bartlomieju:
Will conflict with https://github.com/denoland/deno/pull/13660, we should land both PRs for 1.19

bartlomieju (approved):
LGTM 🚀

ry:
LGTM 👍

ry:
Hopefully the File -> FsFile rename doesn't mess this up somehow

PR #13618, opened 2022-02-07
feat(unstable): Add Deno.upgradeHttp API

### LINKED ISSUE #12718: HTTP hijacking (for Node WS/CONNECT, etc...)

piscisaureus:
fixes #12718


lucacasonato (commented):
I'm really not very happy with this API. It isn't going to work well in `std/http`.

crowlKats:
@lucacasonato its very WIP still

piscisaureus:
My preferred API:

```ts
const reqEvent = await httpConn.nextRequest();
if (!e) return;
const [conn, readBuf]: [Conn, Uint8Array] = await reqEvent.upgradeWith(new Response(null));
```

> It isn't going to work well in `std/http`

I'm sure there's away to add support for this to `std/http`, although I haven't thought of how.

crowlKats:
Ideally something like this:
inteface:
```ts
  export interface RawUpgrade {
    response: Response;
    conn: Promise<Deno.Conn | Deno.TlsConn>;
  }

  export function upgradeRaw(): RawUpgrade;
```
example:
```ts
async function server() {
  let tcp_server = Deno.listen({ host: "127.0.0.1", port: 8080 });
  let tcp_conn = await tcp_server.accept();
  let http_conn = Deno.serveHttp(tcp_conn);
  let http_event = await http_conn.nextRequest();
  const { conn, response } = Deno.upgradeRaw();
  http_event.respondWith(response);
  let raw_conn = await conn; // its Deno.Conn or Deno.TlsConn
}
```

ry:
We've discussed this and came up with this API

```js
import { serve } from "https://deno.land/std/http/server.ts";

serve((req) => {

  let p = Deno.upgradeHttp(req);

  (async () {
    let [conn, firstPacket] = await p;
    // ...
  })();

  // await p; // hangs forever. don't do this.

  return new Response({ status: 101 });
});
```

biluohc:
I think the `firstPacket` is very ugly, may be better to use hyper's `Upgrade`  directly?

crowlKats:
We cant return return Upgrade directly; we'd need to create some new class then on JS side for that. Also being able to differentiate if it is tcp or tls means different functionality can be used (as in, tcp has some methods that tls doesnt). API wise this is just the best.

biluohc:
Ok, the original structure has more control, come on

ry:
- [ ] needs test.js made into real unit test
- [ ] needs to be unstable API

bartlomieju (approved):
LGTM

runtime/ops/http.rs
```diff
@@ -62,3 +72,71 @@ fn op_http_start(
...
+      conn_rid: state
+        .borrow_mut()
+        .resource_table
+        .add(TlsStreamResource::new(tls_stream.into_split())),
+      read_buf: read_buf.to_vec().into(),
+    }),
+    Err(_) => Err(custom_error(
+      "Http",
+      "encountered unsupported transport while upgrading",
+    )),
```
lucacasonato, line 140: Don't we also support Unix transport?

PR #13624, opened 2022-02-08
feat(lsp): provide completions from import map if available

### LINKED ISSUE #13619: lsp: import completions include import maps
Import completions include import map entries and provide completions similar to local workspace completions.

kitsonk:
Closes #13619

(This also updates to import_map v0.8.0)


dsherret (approved):
LGTM

Issue #13626, opened 2022-02-08
Add support for unix sockets to `Deno.serveHttp`


ylxdzsw:
I'm having an issue with HTTP server over unix domain socket. The minimum example to reproduce is

```
await Deno.serveHttp(await Deno.listen({ path: "./a.sock", transport: 'unix' }).accept()).nextRequest()
```

I run this in Deno and try it with

```
curl --unix-socket a.sock http:/a
```

Then it gives the following error msg:

```
error: Uncaught (in promise) BadResource: Bad resource ID
await Deno.serveHttp(await Deno.listen({ path: "./a.sock", transport: 'unix' }).accept()).nextRequest()
           ^
    at Object.opSync (deno:core/01_core.js:149:12)
    at Object.serveHttp (deno:runtime/js/40_http.js:9:22)
    at [***]:1:12
```

Version info:

deno 1.17.1 (release, x86_64-unknown-linux-gnu)
v8 9.7.106.15
typescript 4.5.2

PR #13628, opened 2022-02-08
feat(ext/http): add support for unix domain sockets

ylxdzsw:
This adds support for unix domain sockets in ext/http (#13626). The primary change is replace `std::net::SocketAddr` with a custom enum so we can store the address of a unix domain socket. I have to make this type `pub` in order to refer to it in runtime/ops/http.rs .

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -1156,10 +1156,10 @@ Deno.test(
       for await (const conn of listener) {
         const httpConn = Deno.serveHttp(conn);
         for await (const { request, respondWith } of httpConn) {
-          assertEquals(
-            new URL(request.url).pathname,
-            "/path/name",
-          );
+          const url = new URL(request.url);
+          assertEquals(url.protocol, "http+unix:");
+          assertEquals(decodeURIComponent(url.host), filePath);
+          assertEquals(url.pathname, "/path/name");
           await respondWith(new Response("", { headers: {} }));
           httpConn.close();
         }
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -1142,6 +1142,49 @@ Deno.test(
   },
 );
 
+// https://github.com/denoland/deno/pull/13628
+Deno.test(
+  {
+    ignore: Deno.build.os === "windows",
+    permissions: { read: true, write: true },
+  },
+  async function httpServerOnUnixSocket() {
+    const filePath = Deno.makeTempFileSync();
+
+    const promise = (async () => {
+      const listener = Deno.listen({ path: filePath, transport: "unix" });
+      for await (const conn of listener) {
+        const httpConn = Deno.serveHttp(conn);
+        for await (const { request, respondWith } of httpConn) {
+          assertEquals(
+            new URL(request.url).pathname,
+            "/path/name",
+          );
+          await respondWith(new Response("", { headers: {} }));
+          httpConn.close();
+        }
+        break;
+      }
+    })();
+
+    // fetch() does not supports unix domain sockets yet https://github.com/denoland/deno/issues/8821
+    const conn = await Deno.connect({ path: filePath, transport: "unix" });
+    const encoder = new TextEncoder();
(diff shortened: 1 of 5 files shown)


bartlomieju:
Thanks, this looks quite good, but we definitely need some tests to be able to land it.

ylxdzsw:
@bartlomieju Hi, I'm willing to add some tests, but I don't know where to put them. I didn't find existing tests for the ext/http crate. Should I create a tests folder for the ext/http crate and test in Rust, or somehow test it in Typescript?

bartlomieju:
Sorry I forgot to mention. Tests should be added to `cli/tests/unit/http_test.ts`

ylxdzsw:
@bartlomieju Thanks. I have added a test for it. However I'm not very sure what the url should look like. I have chosen `http+unix://percent_encoding/` (used by httpie) as the other popular choice `http://unix:[path]:/` (used by nginx) is not parsed properly by `new URL()`. I'm open to change it though.

bartlomieju:
> @bartlomieju Thanks. I have added a test for it. However I'm not very sure what the url should look like. I have chosen `http+unix://percent_encoding/` (used by httpie) as the other popular choice `http://unix:[path]:/` (used by nginx) is not parsed properly by `new URL()`. I'm open to change it though.

I'm not sure I understand. Where is this URL being used?

ylxdzsw:
The URL is used to construct the Request object which finally expose to the user. Example:

```
const httpConn = Deno.serveHttp(conn)
for await (const { request, respondWith } of httpConn) {
    console.log(request.url)
}
```

Currently it is constructed using the `Host` header and fallbacks to the socket address. However, both are not avaliable in Unix sockets.

bartlomieju:
@lucacasonato what do you think about the URL?

cli/tests/unit/http_test.ts
```diff
@@ -1142,6 +1142,49 @@ Deno.test(
...
+
+    const promise = (async () => {
+      const listener = Deno.listen({ path: filePath, transport: "unix" });
+      for await (const conn of listener) {
+        const httpConn = Deno.serveHttp(conn);
+        for await (const { request, respondWith } of httpConn) {
+          assertEquals(
+            new URL(request.url).pathname,
+            "/path/name",
+          );
```
lucacasonato, line 1162: Please check that the entire URL is what you expect (scheme + host too).

lucacasonato (commented):
The URL pattern seems fine. By extension of `Deno.listen(transport: unix)` being unstable, this would be unstable too. As such I think we can merge it for 1.19.

Just have a minor comment on the test.

cli/tests/unit/http_test.ts
ylxdzsw: Ok, I have added the test for other components.

runtime/ops/http.rs
```diff
@@ -45,5 +46,18 @@ fn op_http_start(
     return http_create_conn_resource(state, tls_stream, addr, "https");
   }
 
+  #[cfg(unix)]
+  if let Ok(resource_rc) = state
+    .resource_table
+    .take::<UnixStreamResource>(tcp_stream_rid)
+  {
+    let resource = Rc::try_unwrap(resource_rc)
```
bartlomieju, line 54: Please an unstable check here

bartlomieju (commented):
Declaration files should be updated for `Deno.serveHttp` to hint that unix sockets can be used, and as Luca pointed out it should be marked as unstable

ylxdzsw:
Hi @bartlomieju , I have added the unstable check. About the declaration, `Deno.serveHttp` only takes `Conn` as argument and one can already pass a `Conn` created by `Deno.listen({ transport: "unix"})` to it. It just throws `Bad resource ID` error currently as I have reported at #13626. I don't think we need to modify (and I don't know how if needed) the declaration file for this addtion.

bartlomieju:
> Hi, any idea about the CI failling? It passes locally and the last edit is only adding the unstable check. I can't interpret the CI output.

I landed a PR that broke `main`. Please merge `main` branch again and it should go green.

lucacasonato (approved):
LGTM. Great job @ylxdzsw! :tada:

PR #13631, opened 2022-02-09
[WIP] feat(lsp): support  linking to symbols in JSDoc on hover

kitsonk:
Still need to write tests and do a bit more validation.

### DIFF
--- a/cli/lsp/tsc.rs
+++ b/cli/lsp/tsc.rs
@@ -72,6 +72,7 @@ static CODEBLOCK_RE: Lazy<Regex> =
   Lazy::new(|| Regex::new(r"^\s*[~`]{3}").unwrap());
 static EMAIL_MATCH_RE: Lazy<Regex> =
   Lazy::new(|| Regex::new(r"(.+)\s<([-.\w]+@[-.\w]+)>").unwrap());
+static HTTP_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r#"^https?:"#).unwrap());
 static JSDOC_LINKS_RE: Lazy<Regex> = Lazy::new(|| {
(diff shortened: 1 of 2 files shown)


cli/lsp/tsc.rs
```diff
@@ -72,6 +72,7 @@ static CODEBLOCK_RE: Lazy<Regex> =
   Lazy::new(|| Regex::new(r"^\s*[~`]{3}").unwrap());
 static EMAIL_MATCH_RE: Lazy<Regex> =
   Lazy::new(|| Regex::new(r"(.+)\s<([-.\w]+@[-.\w]+)>").unwrap());
+static HTTP_RE: Lazy<Regex> = Lazy::new(|| Regex::new(r#"^https?:"#).unwrap());
```
dsherret, line 75: I think this should use the `i` flag to be case insensitive.

cli/lsp/tsc.rs
```diff
@@ -676,47 +679,136 @@ pub struct QuickInfo {
...
+    }
+  }
+
+  replace_links(out.join(""))
+}
+
 impl QuickInfo {
-  pub fn to_hover(&self, line_index: Arc<LineIndex>) -> lsp::Hover {
-    let mut contents = Vec::<lsp::MarkedString>::new();
+  pub(crate) fn to_hover(
```
dsherret, line 768: One day we should do a quick pass on the lsp code to remove the `pub(crate)` to just `pub`. The `pub(crate)` is unnecessary because most of this isn't even exported from the module IIRC and it's very viral. Not a big deal though.

dsherret (approved):
LGTM! We just need to make that one regex case insensistive.

cli/lsp/tsc.rs
ry: agree! 

(cc @bartlomieju)

PR #13639, opened 2022-02-10
chore: remove Deno.umask (with no argument call)

### LINKED ISSUE #13632: Proposal: Remove Deno.umask (with no argument call)
`Deno.umask()` (no argument) has [a race codition bug](https://github.com/denoland/deno_std/issues/1893#issuecomment-1032897779). `Deno.umask()` (no argument) uses 2 syscalls and between these 2, there's small amount of time when the `umask` value is incorrect, and `mkdir` can create a directory with wrong permission on that timing.

If we try to address the above issue properly, we probably need mutex lock for fs op calls. But that introduces much complexity, and also causes negative performance impact.

~~Also `umask` is a global state, and not something we should encourage using in general situations i.e. it can be considered as an antipattern. So it doesn't make much sense to introduce new complexity to support this feature. So I propose we should remove `Deno.umask`.~~

I propose we should remove `Deno.umask` with no argument

kt3k:
This PR removes the no argument version of Deno.umask as suggested in #13632

closes #13632


kt3k:
@piscisaureus 
How about only removing the query version of `Deno.umask`? (i.e. keep `Deno.umask(newMask: number): number` and remove `Deno.umask(): number`. That also solves the broken permission issue.

Researched a little bit on other languages. [Python](https://docs.python.org/3/library/os.html#os.umask) and [Golang](https://pkg.go.dev/golang.org/x/sys/unix#Umask) has only setting APIs. Ruby has [both](https://docs.ruby-lang.org/en/3.1/File.html#method-c-umask). Node.js used to have both, but [deprecated querying version recently](https://github.com/nodejs/node/pull/32499) after a similar discussion to our case.

kt3k:
Let's track in the issue #13632

PR #13656, opened 2022-02-12
chore(ext/fetch): render jsdoc code example properly

GJZwiers:
The example in the `fetch` JSDoc is not rendering as a code block on hover in VS Code, so I updated it.


bartlomieju (approved):
LGTM, thanks

PR #13662, opened 2022-02-13
feat(ext/ffi): Support read only statics

### LINKED ISSUE #13570: Deno FFI allow statics
In some scenarios it would be useful for Deno FFI to have the ability to point to static data contained in a library.

f.ex. If a library contains some `kVersion` static char for which there is no getter function available, it might still be necessary for Deno to be able to point to said value.

A possible symbol definition could look like this:
```ts
{
  "kVersion": {
    "name": "_ZN3lib8kVersionE",
    "type": "pointer"
  }
}
```

aapoalas:
This PR introduces support for read only statics binding in Deno FFI. The API is added within the existing unstable FFI symbols definition while staying backwards compatible with existing FFI definitions. This means that the API is perhaps a bit confused:

```ts
const dylib = Deno.dlopen(libPath, {
  "printSomething": {
    name: "print_something",
    parameters: [],
    result: "void",
  },
  "static_u32": {
    name: "static_u32",
    type: "u32",
  },
});
```

Statics explicitly define their type, whereas functions just barge in with their definitions just as before. This also means that the serde type for the `HashMap` that is symbols in `op_ffi_load` needs to use an untagged enum.

### Testing
* A few basic tests are added
  * u32 static
  * pointer static

### TypeScript bindings
* The bindings for the FFI API have been updated to include the possibility of native values, but have not been tested for correctness.

### Future work
Mutable statics are not supported by this PR. I see the way forward with those being an optional `mutable` flag on the static definition. If the flag is set set, then the API would use a getter and setter for the value instead of the current fetch-and-cache-once logic.

Closes #13570


ext/ffi/00_ffi.js
```diff
@@ -230,6 +230,28 @@
...
+          );
+          if (type === "pointer" || type === "u64") {
+            value = unpackU64(value);
+            if (type === "pointer") {
+              value = new UnsafePointer(value);
+            }
+          } else if (type === "i64") {
+            value = unpackI64(value);
+          }
+          this.symbols[symbol] = value;
```
littledivy, line 252:
```suggestion
          Object.defineProperty(this.symbols, symbol, { get: () => value });
```

ext/ffi/lib.rs
```diff
@@ -351,10 +365,24 @@ struct ForeignFunction {
   result: NativeType,
 }
 
+#[derive(Deserialize, Debug)]
+#[allow(dead_code)]
+struct ForeignStatic {
+  name: Option<String>,
+  r#type: String,
```
littledivy, line 372:
```suggestion
  #[serde(rename = "type)]
  type_: String,
```
Prefer `type_` over raw literal syntax

ext/ffi/lib.rs
```diff
@@ -631,6 +666,71 @@ async fn op_ffi_call_ptr_nonblocking(
...
+) -> Result<Value, AnyError> {
+  let resource = state
+    .resource_table
+    .get::<DynamicLibraryResource>(args.rid)?;
+
+  let data_ptr = resource.get_static(args.name)? as *const u8;
+
+  Ok(match args.r#type {
+    NativeType::Void => {
+      unreachable!();
```
littledivy, line 690: Are you throwing in JS land? This is still reachable.

littledivy (commented):
Looks great @aapoalas Thanks! Can you also please add tests for the typings in `test_ffi/tests/ffi_types.ts`

ext/ffi/lib.rs
aapoalas: No, this is admittedly mostly a `todo!()`. There's probably no reason to define a void type static, so I left this to stew.

Should I just return a void for completeness sake?

ext/ffi/00_ffi.js
littledivy: Good point. Function fields should be a getter too, feel free to make that change in this PR. Also use the primordial for `Object.defineProperty` instead

ext/ffi/lib.rs
littledivy: EDIT: Should be `"type"` instead of `"type`

ext/ffi/lib.rs
littledivy: I'd rather throw JS side because just returning a void isn't really the correct behavior (?)

aapoalas: I'm not even sure there is a correct behaviour.

aapoalas:
> Looks great @aapoalas Thanks! Can you also please add tests for the typings in `test_ffi/tests/ffi_types.ts`

Alright, complete cadre of tests for the supported (and even the unsupported "void") static types are now available in the `ffi_types.ts`. FFI tests pass locally.

ext/ffi/lib.rs
```diff
@@ -351,10 +365,25 @@ struct ForeignFunction {
   result: NativeType,
 }
 
+#[derive(Deserialize, Debug)]
+#[allow(dead_code)]
```
littledivy, line 369: Why is this dead code?

aapoalas: The type and name field for the constants are never actually read. I'll check if I can just mark the `ForeignStatic` as a zero-size variant.

littledivy: Oh ok, can you add a comment similar to the one at line 493 here?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -176,9 +176,15 @@ declare namespace Deno {
     nonblocking?: NonBlocking;
   }
 
-  /** A foreign function interface descriptor */
-  export interface ForeignFunctionInterface {
-    [name: string]: ForeignFunction;
+  export interface ForeignStatic<Type extends NativeType = NativeType> {
+    /** Name of the symbol, defaults to the key name in symbols object. */
+    name?: string;
+    type: Type;
```
littledivy, line 182:
```suggestion
    type: Omit<Type, "void">;
```

littledivy (changes requested):

ext/ffi/lib.rs
aapoalas: Added comment.

littledivy:
@aapoalas The TypeScript tests are failing. See https://github.com/denoland/deno/runs/5208811632?check_suite_focus=true

ext/ffi/lib.rs
piscisaureus: Can't you just name the thing `_type` rather than `type_`? 
I think that would avoid the warning as well.

piscisaureus:
Overall looks pretty good to me. Of course tests have to be green and I had one comment.

aapoalas:
Okay, TS tests should now be on the way to passing again, and @piscisaureus 's comment should be resolved I hope :)

littledivy (approved):
LGTM, Thanks @aapoalas!

Issue #11141, opened 2021-06-26
`Deno.emit` panicks when using `data:application/typescript;base64` as url


lowlighter:
Trying to bundle auto-generated code with `Deno.emit` but data url results in panics, though it should be possible as both `import()` and `Worker` handles data urls
```ts
import { encode } from "https://deno.land/std@0.99.0/encoding/base64.ts";
const url = new URL(`data:application/typescript;base64,${encode("console.log('hello')")}`)
console.log(await import(url.href))
console.log(await Deno.emit(url.href, {bundle:"classic"}))
```

```
hello
Module {}
thread 'main' panicked at 'unsupported ModuleSlot error', cli\module_graph.rs:1378:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

satyarohith:
I couldn't reproduce the panic on `main` and `1.11.2`, but it throws a `NotFound: Cannot resolve module` error for data URLs.  
```
➜ cat 11141.ts
import { encode } from "https://deno.land/std@0.99.0/encoding/base64.ts";
const url = new URL(`data:application/typescript;base64,${encode("console.log('hello')")}`)
console.log(await import(url.href))
console.log(await Deno.emit(url.href, {bundle:"classic"}))

➜ deno run --unstable -A 11141.ts
hello
Module {}
error: Uncaught (in promise) NotFound: Cannot resolve module "file:///Users/sr/deno_issues/data:application/typescript;base64,Y29uc29sZS5sb2coJ2hlbGxvJyk=".
console.log(await Deno.emit(url.href, {bundle:"classic"}))
            ^
    at deno:core/core.js:86:46
    at unwrapOpResult (deno:core/core.js:106:13)
    at async file:///Users/sr/deno_issues/11141.ts:4:13

➜ deno --version
deno 1.11.2+015f252 (canary, x86_64-apple-darwin)
v8 9.2.230.12
typescript 4.3.2
```

lowlighter:
Sorry I forgot to put the version:
```
> deno --version
deno 1.11.2 (release, x86_64-pc-windows-msvc)
v8 9.1.269.35
typescript 4.3.2
```

Maybe it occurs only on Windows ?

kitsonk:
Data URLs aren't supportable as root modules for various reasons. We should look at the error messages though and maybe try to detect this earlier. It may be possible to support them, but I think we run into issues with type checking via tsc.

Being able to mix generated and external content is 
[...]

PR #13667, opened 2022-02-14
feat(compiler): support data url

lowlighter:
Noticed that `deno bundle` does seems to support data url:
```ts
PS> deno bundle "data:application/javascript;base64,Y29uc29sZS5sb2coImhlbGxvIHdvcmxkIik7"
Bundle data:application/javascript;base64,Y29uc29sZS5sb2coImhlbGxvIHdvcmxkIik7
// deno-fmt-ignore-file
// deno-lint-ignore-file
// This code was bundled using `deno bundle` and it's not recommended to edit it manually

console.log("hello world");
```

However when using it programmatically with `Deno.emit`, there is a `checkRelative` function which only support http(s)/file protocols, not the data one. It gets converted into a relative path instead of being passed as it.

Should close https://github.com/denoland/deno/issues/11141

### DIFF
--- a/cli/tests/testdata/compiler_api_test.ts
+++ b/cli/tests/testdata/compiler_api_test.ts
@@ -46,6 +46,21 @@ Deno.test({
 });
 
 Deno.test({
+  name: "Deno.emit() - data url",
+  async fn() {
+    const data =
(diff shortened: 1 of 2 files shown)


kitsonk (approved):
LGTM

Apologies for the delay

lowlighter:
No worries! 
Thanks a lot 👍

PR #13668, opened 2022-02-14
BREAKING: don't inherit permissions by default

lucacasonato:
Previously specifying permissions: {} was the same as specifying
permissions: "inherit". Now it will be the same as permissions: "none".
Not specifying any permissions (permissions: undefined) still means
permissions: "inherit".

This is a breaking change. I want to try land this soon, as the current
behaviour is really impractical. If this can not be changed right now, I
am proposing it as a 2.0 change.


bartlomieju:
I'm also in favor of this change but even though it's breaking I think we should ship it in 1.20. This behavior is less surprising and if some user hits the problem with it I think they will appreciate the soundness of new behavior.

kitsonk:
I'm in favour of doing it for 1.20.  Yes, it is a breaking change, but one could argue it is a security issue, which if it breaks some people doing things that are unsafe, we have helped them.

bartlomieju (approved):
LGTM, let's land it

lucacasonato:
Why exactly did we land this with no tests?

Issue #13525, opened 2022-01-28
[AES-GCM] 'additionalData' field is required on crypto.subtle.encrypt()


pedropaulosuzuki:
Hello! How are you doing? Hope everyone's okay!

Today I was trying to use AES-GCM encryption on a project and realized that the optional field 'additionalData' in [AesGcmParams](https://developer.mozilla.org/en-US/docs/Web/API/AesGcmParams) wasn't that optional, after all. It seems that the new implementation on Deno 1.18.1 requires an ArrayBuffer to be passed with the function, even if it is an empty one. Setting 'additionalData' to `undefined` doesn't work.

Script:
```ts
await crypto.subtle.encrypt({
    name: 'AES-GCM',
    iv: new Uint8Array(12),
    tagLength: 96,
    additionalData: new ArrayBuffer(0) // setting to undefined doesn't work, ommiting doesn't work either
}, key, data);
```

Output error:
```js
error: Uncaught (in promise) TypeError: Error parsing args: missing field `additionalData`
```
Deno version:
```
deno 1.18.1 (release, x86_64-unknown-linux-gnu)
v8 9.8.177.6
typescript 4.5.2
```

PR #13669, opened 2022-02-14
fix(ext/crypto): optional additionalData in encrypt/decrypt

littledivy:
Closes https://github.com/denoland/deno/issues/13525

Was failing earlier because `undefined` is not `None` with serde (for `Option<Vec<u8>>`). This patch passes an empty buf to the op.


ext/crypto/00_crypto.js
```diff
@@ -691,7 +691,8 @@
             algorithm: "AES-GCM",
             length: key[_algorithm].length,
             iv: normalizedAlgorithm.iv,
-            additionalData: normalizedAlgorithm.additionalData,
+            additionalData: normalizedAlgorithm.additionalData ||
```
cryptographix, line 694: Maybe use `??` instead of `||`. Same for L3829

Not sure what the Deno standard on this is, but `??` shows intent as default value.

littledivy: Doesn't make any difference. webidl takes care of the types so `normalizedAlgorithm.additionalData` can only be `undefined` or a uint8array. I don't think Deno has any standard on this. I would've preferred to just pass `null` to serde here because defaults are set Rust side.

lucacasonato (commented):
Make `additional_data` non optional Rust side if you are doing the defaulting JS side.

littledivy:
@lucacasonato I've made it `null` instead so we can keep the spec logic Rust side.

lucacasonato (approved):
LGTM

PR #13670, opened 2022-02-14
feat: deno vendor

### LINKED ISSUE #13346: Proposal: `deno vendor`
## Context

Deno aspires to provide a full featured set of development tools baked into the single `deno` CLI. At the same time, Deno provides an open platform where code can be imported just as easily as a browser can fetch a web page.  Deno stores remote dependencies in a cache commonly referred to as the `DENO_DIR`.  While Deno allows you "relocate" this directory at runtime, and the files in the cache are editable, it does not solve all the use cases in a well supported way.

### Use cases

- As a developer, I want to be able to quickly edit a remote dependency to investigate a coding issue.
- As a developer, I want to be able to store all of my remote dependencies for a given project and check them into my source code to ensure that those dependencies don't change or can't be altered.
- As a developer, I want to be able to develop a workload with the Deno CLI and be able to test it and then deploy it on the edge with Deno Deploy with a fixed set of dependencies, not relying upon availability of other hosts or services for my application.

All of these use cases also are constrained by:

- Without needing to edit or change the code I have authored.
- I want to be able to quickly "revert" if needed.

## Proposed Solution

Add a subcommand to the Deno CLI named `deno vendor`.  The subcommand would work similar to the `deno run` subcommand, in that it would take a target script.  For example:

```
> deno vendor main.ts
```

Would analyse `main.ts` and all of its dependencies.  For those dependencies that are remote dependencies, it would write their source out locally to the `_vendor` path, along with `_vendor/import_map.json`, which would remap all the remote dependencies to local ones.

The concept would then be then that changes could be made to the local "vendored" code and be reflected in the application by simply using the output import map:

```
> deno run --import-map _vendor/import_map.json main.ts
```

(Note: having to be explicit about this import map isn't the _ideal_ design, and suggestions below in other complimentary features would be able to make this just so then all the user would need to do is `deno run main.ts`)

This would also allow specific dependencies to be vendored, without having to vendor everything.  For example if someone wanted to just vendor oak and all of it dependencies, but leave everything else to continue to be remote:

```
> deno vendor https://deno.land/x/oak@v10.1.0/mod.ts
```

Other flags that should be supported:

- `--output` - specify the output path, which defaults to `_vendor` of cwd.
- `--import-map` - specifies an _upstream_ import map that should be used as the base when calculating the vendored import map
- `--config` - specifies a configuration file, which will be considered when vendoring.
- `--info` - output human readable information about the current state of the `_vendor` directory.

Other flags/features that might be supported, or might be added after the "MVP" feature:

- `--ignore` - a glob pattern of dependencies to exclude/ignore
- The ability to remove specific dependencies from `_vendor` (which can be accomplished from editing the `_vendor/import_map.json` directly, but might be unapproachable for some devs)

### Semantics

- Once the remote dependencies are identified, they would be written out in a way to minimise the need of remapping the paths.  For example `https://example.com/pkg/mod.ts` would get written out to `./_vendor/https/example.com/pkg/mod.ts`.  In cases where the remote dependency cannot be written to disk following the simple algorithm, it will be written out in as close to a human readable way as possible.
- Once the "manifest" of the `_vendor` is calculated, the import map will be generated with the smallest set of entries possible.  For example a remote import of `https://example.com/pkg/mod.ts` would get an entry like the following:

  ```json
  {
    "imports": {
      "https://example.com/": "_vendor/example.com/"
    }
  }
  ```

  But if there was something like `https://example.com/pkg/mod.ts` and `https://example.com/pkg/lib?dev` something like the following would be written out:

  ```json
  {
    "imports": {
      "https://example.com/pkg/lib?dev": "_vendor/example.com/pkg/lib_a02vf0.ts",
      "https://example.com/": "_vendor/example.com/"
    }
  }
  ```

- If a `_vendor` or output path already exists, the `_vendor/import_map.json` and output dependencies would be updated.
- A design limitation would be that some dynamic imports that cannot be statically analysable would be missed. A consideration might be to warn on such dependencies being present.
- It is assumed that at this point in time, other state/meta data does not need to be stored, but might need to be. If this is the case, then the state should be stored in the `_vendor/import-map.json` under additional _private_ keys (e.g. `"__deno_vendor": {}`).  The import map specification specifically allows additional keys without invalidating the file, and this would help ensure that the output of `deno vendor` is contained, repeatable and portable.
- When an existing `_vendor` is updated, and for some reason a remote dependency has changed, import map `"scopes"` can be used to allow two different versions of the same dependency to exist in the vendor directory.

### Other features which improve/compliment this feature

- Automatic detection of `deno.json`/`deno.jsonc` - This is already in progress.
- Ability to express an import map in `deno.json`/`deno.jsonc`.  This would allow the feature a) identify an existing import map that should be amended with the vendor mappings as well as "point" the configuration at the vendor import map.
- Add the _Deno: Vendor Dependencies_ to the vscode_deno extension.
- Ability to express and support import maps with Deno Deploy.  This would allow someone who had vendored dependencies to deploy it to Deploy.

### Specific advantages to this proposal

- It aligns to the web platform standards, providing a portable solution that is independent of Deno CLI.
- Integrates well into existing developer workflows.
- Because all the "magic" is done in import maps, it gives the end user fine grained control over the behaviour, allowing them to address edge cases or different workflows without "breaking" the feature or making the feature unusable.
- The feature is open, transparent, portable, easily reversible, non-destructive and totally optional.
- Easily integrate-able into editors, so that a dev would run the command and presto, all the remote dependencies are local and editable.

### Other options considered

- Making access to `DENO_DIR` easier - currently, users can edit files in the `DENO_DIR`. There are significant risks to this, like depending on the dependency type and the state of the cache, files may or may not be invalidated in a way that causes the program to emit properly.  Also, it is a reasonable expectation that we should consider `DENO_DIR` and its schema "private" and while accessible, the layout and format may need to change as Deno evolves, and if people are tightly coupled to this, it becomes problematic and could easily hold back the development of Deno for fear of breaking integrations to non-public behaviours. Another risk is that it would not be (ever) compatible with Deno Deploy.
- Adding specific features to the language server to edit dependencies - this would then require people to use an editor with a language server to take advantage of this feature, which is not realistic or desired.

dsherret:
Not part of this first pass:

- Warning when dynamic import is not statically analyzable (https://github.com/denoland/deno_graph/issues/138)
- Special handling of file system layout to take into account Window's 260 character path limit. I had some code that handled this, but it wasn't very good and I thought of a better solution. I'll open an issue and do another PR at a later time.
- Module specifiers with different text in the same module that resolve to the same module aren't handled (https://github.com/denoland/deno_graph/issues/132) -- Not a big deal because this is probably super rare
- Directories with different casing resolve to the same folder on case insensistive file systems. I will have an upcomming PR with the 260 char limit that fixes this (just added this to dnt [here](https://github.com/denoland/dnt/pull/125))
- The specified import map is not combined with the generated import map. It might be hard to do this correctly because the base of the import map changes when going into the vendor folder.

Differences with proposal:

- Uses `vendor` folder instead of `_vendor` by default (seemed like in the meeting people wanted `vendor` instead?).
- Output is not dependent on the existing state of the vendor folder.
  - Errors when the vendor folder exists and is non-empty. Allows using a `--force` flag to potentially overwrite the contents.
  - No `--info` flag
  - Specifying an import map via `--import-map ...` located within the output directory will error.
  -  Why: Vendor is already a complex feature and I think basing the vendor output on the existing state of the vendor folder would be very difficult to do. It would also be hard to fix bugs because we'd need to know the previous state of the vendor folder. I think it also might be difficult for some users to understand.

Closes #13346


lucacasonato:
> Module specifiers with different text in the same module that resolve to the same module aren't handled (https://github.com/denoland/deno_graph/issues/132) -- Not a big deal because this is probably super rare

I agree not that important for a first pass, but I don't agree it's super rare. Often you see in deps.ts files when people re-export both types and symbols:
```js
export { foo } from "https://example.com/";
export type { foo } from "https://example.com/";
```

cli/flags.rs
```diff
@@ -2497,8 +2574,8 @@ mod tests {
 
   /// Creates vector of strings, Vec<String>
   macro_rules! svec {
-    ($($x:expr),*) => (vec![$($x.to_string()),*]);
-}
+    ($($x:expr),* $(,)?) => (vec![$($x.to_string()),*]);
```
dsherret, line 2577: Trailing comma support.

cli/fs_util.rs
```diff
@@ -362,6 +362,32 @@ pub fn path_has_trailing_slash(path: &Path) -> bool {
   }
 }
 
+/// Gets a path with the specified file stem suffix.
+pub fn path_with_stem_suffix(path: &Path, suffix: &str) -> PathBuf {
+  if let Some(file_name) = path.file_name().map(|f| f.to_string_lossy()) {
```
dsherret, line 367: `to_string_lossy` 🤷‍♂️

cli/tests/integration/vendor_tests.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
```
dsherret, line 1: Most of the tests aren't in this file. They'll be in `cli/tools/vendor/build.rs`. This is just the final integration tests.

cli/tests/integration/vendor_tests.rs
```diff
@@ -0,0 +1,331 @@
...
+    .arg("vendor")
+    .arg("--output")
+    .arg("vendor")
+    .arg("mod.ts")
+    .stderr(Stdio::piped())
+    .spawn()
+    .unwrap();
+  let output = deno.wait_with_output().unwrap();
+  assert_eq!(
+    String::from_utf8_lossy(&output.stderr).trim(),
```
dsherret, line 55: Would be neat if we had a less verbose way of running the deno process and getting the output in the tests.

dsherret:
@lucacasonato it will just be an issue where the module specifiers differ and don't match the file name like:

```ts
export { foo } from "https://example.com/mod";
export type { foo } from "https://example.com/redirect_to_mod";
```

lucacasonato:
It'd be nice to add a little message after `deno vendor` succeeds:

```
$ deno vendor main.ts
Vendored x modules into vendor/ directory.

To use vendored modules, specify the `--import-map` flag when invoking deno subcommands:
  deno run -A --import-map=./vendor/import_map.json main.ts
```

cli/tools/vendor/mappings.rs
```diff
@@ -0,0 +1,286 @@
...
+      specifier.clone()
+    } else {
+      let local_path = self.local_path(specifier);
+      if specifier.path().ends_with('/') {
+        ModuleSpecifier::from_directory_path(&local_path)
+      } else {
+        ModuleSpecifier::from_file_path(&local_path)
+      }
+      .unwrap_or_else(|_| {
+        panic!("Could not convert {} to uri.", local_path.display())
```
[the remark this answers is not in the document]
dsherret: I don't think it would ever happen. Should I change it to `.unwrap()`?

ry (commented):
Looks great. After landing the fix for the issue Luca discovered in the dotcom repo, I think this is ready to land.

cli/tools/vendor/mappings.rs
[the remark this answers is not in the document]
ry: Nope - useful internal errors are also helpful. Just wondering!

cli/tools/vendor/mappings.rs
```diff
@@ -0,0 +1,286 @@
...
+        .unwrap();
+    }
+
+    // workaround for url crate not adding a trailing slash for a directory
+    // it seems to be fixed once a version greater than 2.2.2 is released
+    let is_dir = to.path().ends_with('/');
+    let mut text = from.make_relative(&to).unwrap();
+    if is_dir && !text.ends_with('/') && to.query().is_none() {
+      text.push('/');
+    }
```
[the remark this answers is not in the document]
dsherret: It's a bit of a pain to test this struct because it involves creating a module graph. I thought about extracting some of the functionality in here out, but it adds a bit of indirection. The logic is temporary workarounds and the code in here is tested transitively—if I delete either of the if statements or start flipping conditions then the vendor tests will fail.

lucacasonato (approved):
LGTM! All the code i tested now works as expected. :tada:

cli/tools/vendor/mappings.rs
```diff
@@ -0,0 +1,286 @@
...
+    // resolve all the "proxy" paths to use for when an x-typescript-types header is specified
+    for module in remote_modules {
+      if let Some((
+        _,
+        Resolved::Ok {
+          specifier, range, ..
+        },
+      )) = &module.maybe_types_dependency
+      {
+        // hack to tell if it's a x-typescript-types header
```
bartlomieju, line 86: I don't get what this hack does :(

cli/tools/vendor/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+  // check the import map
+  if let Some(import_map_path) = ps
+    .maybe_import_map
+    .as_ref()
+    .and_then(|m| m.base_url().to_file_path().ok())
+    .and_then(|p| fs_util::canonicalize_path(&p).ok())
+  {
+    // make the output directory in order to canonicalize it for the check below
+    std::fs::create_dir_all(&output_dir)?;
+    let output_dir = fs_util::canonicalize_path(output_dir)?;
```
bartlomieju, line 76: Would be great to add `.context()` here to tell users at what stage something went wrong, otherwise the errors will be like `PermissionDenied (error 13)` or something like this. Very hard to debug

cli/tools/vendor/mappings.rs
dsherret: See https://github.com/denoland/deno_graph/issues/133

The hack is to check if the range start and end in the file is zero.

cli/tools/vendor/build.rs
```diff
@@ -0,0 +1,577 @@
...
+              "export * from '/absolute.ts';",
+            ),
+          )
+          .add("https://localhost/other.ts", "export class Other {}")
+          .add_redirect(
+            "https://localhost/redirect.ts",
+            "https://localhost/other.ts",
+          )
+          .add("https://localhost/absolute.ts", "export class Absolute {}")
+          .add("https://other/mod.ts", "export * from './sub/mod.ts';")
```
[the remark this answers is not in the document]
dsherret: Yes, but a big part of `deno vendor` is we don't want to change the files. Someone could also exclude the vendor folder from being formatted.

littledivy (approved):
LGTM! Tested it on a bunch of projects, its perfect!

PR #13676, opened 2022-02-15
feat(test): improved op sanitizer errors + traces

lucacasonato:
This commit improves the error messages for the `deno test` async op
sanitizer. It does this in two ways:
- it uses handwritten error messages for each op that could be leaking
- it includes traces showing where each op was started

This "async op tracing" functionality is a new feature in deno_core.
It likely has a significant performance impact, which is why it is only
enabled in tests.

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -22,6 +22,8 @@
     MapPrototypeDelete,
     MapPrototypeSet,
     PromisePrototypeThen,
+    PromisePrototypeFinally,
+    StringPrototypeSlice,
     ObjectAssign,
     SymbolFor,
   } = window.__bootstrap.primordials;
@@ -47,6 +49,13 @@
   // TODO(bartlomieju): it future use `v8::Private` so it's not visible
   // to users. Currently missing bindings.
   const promiseIdSymbol = SymbolFor("Deno.core.internalPromiseId");
+
+  let opCallTracingEnabled = false;
+  const opCallTraces = new Map();
+
+  function enableOpCallTracing() {
+    opCallTracingEnabled = true;
+  }
 
   function setPromise(promiseId) {
     const idx = promiseId % RING_SIZE;
@@ -139,7 +148,17 @@
     const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
-    const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    let p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    if (opCallTracingEnabled) {
+      // Capture a stack trace by creating a new `Error` object. We remove the
+      // first 6 characters (the `Error\n` prefix) to get just the stack trace.
+      const stack = StringPrototypeSlice(new Error().stack, 6);
+      MapPrototypeSet(opCallTraces, promiseId, { opName, stack });
+      p = PromisePrototypeFinally(
+        p,
+        () => MapPrototypeDelete(opCallTraces, promiseId),
+      );
+    }
     // Save the id on the promise so it can later be ref'ed or unref'ed
     p[promiseIdSymbol] = promiseId;
     return p;
@@ -226,6 +245,8 @@
     BadResourcePrototype,
     Interrupted,
     InterruptedPrototype,
+    enableOpCallTracing,
+    opCallTraces,
   });
 
   ObjectAssign(globalThis.__bootstrap, { core });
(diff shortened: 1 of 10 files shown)


core/01_core.js
```diff
@@ -139,7 +148,15 @@
     const maybeError = opcallAsync(opsCache[opName], promiseId, arg1, arg2);
     // Handle sync error (e.g: error parsing args)
     if (maybeError) return unwrapOpResult(maybeError);
-    const p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    let p = PromisePrototypeThen(setPromise(promiseId), unwrapOpResult);
+    if (opCallTracingEnabled) {
+      const stack = StringPrototypeSlice(new Error().stack, 6);
```
dsherret, line 153: What's the significance of 6? Maybe add a comment or explanatory variable (probably using `"Error\n".length` is not appropriate for this code... I'm not sure about the overhead and we probably want to micro-optimize this).

runtime/js/40_testing.js
```diff
@@ -75,56 +150,67 @@
...
+          const [name] = OP_DETAILS[key];
+          const count = completedDiff - dispatchedDiff;
+          ArrayPrototypePush(
+            details,
+            `${count} async operation${count === 1 ? "" : "s"} to ${name} ${
+              count === 1 ? "was" : "were"
+            } started before this test, but ${
+              count === 1 ? "was" : "were"
+            } completed during the test. Async operations should not complete in a test if they were not started in that test.`,
+          );
```
dsherret, line 207: This is better, but I think still not exactly correct if I'm understanding the code correctly. I'm not sure we should be counting all ops, but instead only counting ops that weren't started when the test ran.

Example:

1. User starts Test A without sanitizers. It finishes with 1 pending op and Test B starts.
2. While Test B is running, the op from Test A completes.
3. Test B completes with 1 pending op.
4. Sanitizers incorrectly pass because the diff is 0.

I guess this is an existing issue though and maybe something we should fix in a separate PR.

dsherret (approved):
LGTM! This should help out users a lot.

runtime/js/40_testing.js
```diff
@@ -57,6 +58,79 @@
...
+    "op_open_async": ["open a file", "awaiting the result of a `Deno.open` call"],
+    "op_read_dir_async": ["read a directory", "collecting all items in the async iterable returned from a `Deno.readDir` call"],
+    "op_read_link_async": ["read a symlink", "awaiting the result of a `Deno.readLink` call"],
+    "op_realpath_async": ["resolve a path", "awaiting the result of a `Deno.realpath` call"],
+    "op_remove_async": ["remove a file or directory", "awaiting the result of a `Deno.remove` call"],
+    "op_rename_async": ["rename a file or directory", "awaiting the result of a `Deno.rename` call"],
+    "op_run_status": ["get the status of a subprocess", "awaiting the result of a `Deno.Process#status` call"],
+    "op_seek_async": ["seek in a file", "awaiting the result of a `Deno.File#seek` call"],
+    "op_signal_poll": ["get the next signal", "un-registering a OS signal handler"],
+    "op_sleep": ["sleep for a duration", "awaiting or cancelling a `setTimeout` or `setInterval` call"],
```
ry, line 115: `s/awaiting or//`

core/lib.deno_core.d.ts
```diff
@@ -141,5 +141,28 @@ declare namespace Deno {
...
+    /**
+     * Enables collection of stack traces of all async ops. This allows for
+     * debugging of where a given async op was started. Deno CLI uses this for
+     * improving error message in op sanitizer errors for `deno test`.
+     *
+     * **NOTE:** enabling tracing has a significant negative performance impact.
+     * To get high level metrics on async ops with no added performance cost,
+     * use `Deno.core.metrics()`.
+     */
+    function enableOpCallTracing(): void;
```
ry, line 154: why not just have a boolean value that can be assigned, rather than a function:
```
Deno.core.enableOpCallTracing = true;
```
This allows for future flexibility (if you want to turn it off) and less wrapper.

ry (approved):
LGTM - seems like it will really help people understand their programs!

runtime/js/40_testing.js
lucacasonato: This is actually somewhat intentional to allow for background HTTP servers to work without breaking the sanitizer. Let's figure this out in a follow up.

core/lib.deno_core.d.ts
lucacasonato: The chosen approach is actually intentional: the "enabled" flag is a boolean (a primitive), so assigning it to `Deno.core` assigns it by value, not by reference. To fix this, we'd need to change the check in `opAsync` to be `if (core.opCallTracingEnabled)` instead of `if (opCallTracingEnabled)`. The former introduces an additional property lookup to all `opAsync` calls which I don't really want as it is a super hot path.

core/01_core.js
lucacasonato: Added a comment.

PR #13680, opened 2022-02-16
chore: update v8

lucacasonato:


AaronO (changes requested):
The `Cargo.lock` doesn't appear to be in sync, you need to regen it.

It contains both `0.39.0` and `0.40.0`

AaronO (approved):
LGTM

PR #13681, opened 2022-02-16
docs(ext/fetch): fix copy-paste errors in Request docs

ylxdzsw:
The document of some methods in the `Request` object refers to `Response`, which I believe are copy-paste error (there are same methods in the `Response` object with identical doc string)


lucacasonato (approved):
LGTM, thanks!

PR #13687, opened 2022-02-16
feat(ext/web): Add `AbortSignal.timeout()`.

### LINKED ISSUE #13582: Implement `AbortSignal.timeout()`
See whatwg/dom#1032.

I'll probably take care of this, unless @crowlKats or someone else wants to do it.

andreubotella:
Closes #13582.


ext/web/03_abort_signal.js
```diff
@@ -34,6 +35,24 @@
...
     }
 
+    static timeout(millis) {
+      const prefix = "Failed to call 'AbortSignal.timeout'";
+      webidl.requiredArguments(arguments.length, 1, { prefix });
+      millis = webidl.converters["unsigned long long"]({ enforceRange: true });
+
+      const signal = new AbortSignal(illegalConstructorKey);
+      // TODO(andreubotella): Don't block the event loop from exiting if there
+      // aren't any listeners for the abort event.
```
crowlKats, line 45: Should be fairly simple with `Deno.unrefTimer` & `Deno.refTimer`, though they are unstable

andreubotella: I wasn't sure about overriding `addEventListener` and `removeEventListener` though, with such short timeframe before the release of 1.19.

lucacasonato:
I don't think this is going to make it. There are 2 hours left before `denoland/deno` freezes, and I'd really like to have the ref handling work correctly before then. Also we need to figure out how this integrates with test sanitizer.

I think it would make more sense if we don't actually start the timer until the first listener is registered. We can also cancel the timer once the last listener unregisters.

andreubotella:
> I think it would make more sense if we don't actually start the timer until the first listener is registered. We should also cancel the timer once the last listener unregisters.

Come to think about it, this would be hard to do right, since per the spec the signal is only aborted in a timer task:

```js
const signal = AbortSignal.timeout(10);
console.log(signal.aborted);  // false

const start = Date.now();
while (Date.now() - start < 20) {}

console.log(signal.aborted);  // must still be false!
```

lucacasonato:
Could be solved with a macrotask callback I think.

andreubotella:
> Could be solved with a macrotask callback I think.
Okay, so let's say that someone checks if the signal has aborted, and the timeout has indeed passed. Do we queue a macrotask now? Because if it has expired a while ago and the event loop hasn't been blocked during that time, the signal should have aborted by the time we're checking. (The spec technically would allow for this, but it's far from ideal.)

andreubotella:
I'm not sure that an implementation that only has an active timer while there are listeners is workable, so I've implemented this with `Deno.unrefTimer()` and added tests. I guess that idea could still be pursued after this implementation is merged.

bartlomieju:
@crowlKats @lucacasonato would be great to get this in for 1.20. Please review

ext/web/03_abort_signal.js
```diff
@@ -25,6 +27,8 @@
   const illegalConstructorKey = Symbol("illegalConstructorKey");
 
   class AbortSignal extends EventTarget {
+    #timerId = null;
```
crowlKats, line 30: this should be a symbol, not private prop

andreubotella: Done

PR #13691, opened 2022-02-16
Revert "feat: prompt by default"

lucacasonato:
Reverts denoland/deno#13650

Calling `performance.now()` on a permissionless isolate causes a prompt to appear rather than to return a non-granular time like before. This needs to be fixed before re-landing.


bartlomieju (approved):
Lgtm, I'm strongly in favor of reverting

PR #13695, opened 2022-02-16
feat: Add hint to permission prompt to display allow flag

ry:


runtime/permissions.rs
```diff
@@ -1976,10 +2012,10 @@ fn permission_prompt(message: &str) -> bool {
   // buffered data cannot effect the prompt.
   clear_stdin();
 
-  let opts = "[y/n (y = yes allow, n = no deny)] ";
+  let opts = "[yes / no] ";
```
crowlKats, line 2015: why make the options less informative of what can be passed?

crowlKats (changes requested):
Please use the `self.name` property instead of hard-coding the names again.

runtime/permissions.rs
ry: It's rather opaque with all the syntax: `/` `(` `=` `,`, `[`, `]`
I think this is easier to read

crowlKats: then maybe `[(y)es / (n)o]`? i have seen this in a few clis

ry: I'll undo this change, it's unrelated.

cli/tests/integration/run_tests.rs
```diff
@@ -2197,9 +2197,9 @@ mod permissions {
     let args = "run --quiet 061_permissions_request.ts";
     use util::PtyData::*;
     util::test_pty2(args, vec![
-      Output("⚠️  ️Deno requests read access to \"foo\". Allow? [y/n (y = yes allow, n = no deny)] "),
+      Output("⚠️  ️Deno requests read access to \"foo\". Run again with --allow-read to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)] "),
```
dsherret, line 2200: I don't understand why some of these have a trailing space and others don't considering a trailing space was added to the output. I guess there's some trimming on the output being done in the comparison.

dsherret (approved):
LGTM

Issue #9175, opened 2021-01-19
Feature Request: deno bench subcommand 


grian32:
Add a `deno bench` subcommand that behaves like `deno test`, this would remove `runBenchmarks()`, this could also be achieved by making it so `deno run` ignores `runBenchmarks()` and `runBenchmarks()` is ran using `deno bench`, or we could add the extra functionality of `runBenchmarks()`(other than running the benchmarks, like the run only based on a regexp etc) as cli args for `deno bench`.

The reason I want this is because I like the way you're encouraged to organize tests in Rust(https://doc.rust-lang.org/book/ch11-03-test-organization.html), and I think this is also a thing for benchmarks, although benchmarks in Rust as of now are unstable.
You can currently do this in deno using `deno test` and having the tests, but I think it'd be nice if you could also do it with benchmarks, which isn't really possible with the way it's currently done.

Would like some feedback on this idea.

bartlomieju:
My personal opinion is that we should add this subcommand. That said, it's gonna be a big endeavour for anyone interested in working on this. We have [`std/testing/bench.ts`](https://deno.land/std@0.119.0/testing/bench.ts) module but there's not much feedback about it and most of the logic would have to be reimplemented in Rust in order to support it as a subcommand.

If anyone is interested in spearheading the effort, please let us know, I'll be glad to help out.

PR #13713, opened 2022-02-19
feat: "deno bench" subcommand

bartlomieju:
Closes https://github.com/denoland/deno/issues/9175

```ts
// Defaults to 1000 warmup iterations and 1000 bench iterations
Deno.bench("URL parsing", { n: 250, warmup: 100 }, () => {
  new URL("https://deno.land");  
});


Deno.bench("URL resolving", { n: 1000 }, () => {
  new URL("./foo.js", import.meta.url);
});
```

```
$ deno bench --unstable example_bench.ts
running 2 benches from file:///Users/ib/dev/deno/example_bench.ts
bench URL parsing ... 250 iterations 23,063 ns/iter (208..356,041 ns/iter) ok (1s)
bench URL resolving ... 1000 iterations 48,045 ns/iter (250..128,625 ns/iter) ok (4s)

bench result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (5s)
```


runtime/js/40_testing.js
```diff
@@ -792,6 +928,47 @@
...
+      const description = {
+        origin,
+        name: bench.name,
+      };
+      const earlier = DateNow();
+
+      reportTestWait(description);
+
+      const result = await runBench(bench, description);
+      const elapsed = DateNow() - earlier;
```
littledivy, line 964: I really wish this could be done in Rust using `Instant` and V8 APIs for best results.

evanwashere (commented):
# mvp
- support for up to nanoseconds accuracy and formatting
- output should also contain min/max, can be displayed as `avg unit/op (min..max unit/op)` or we can take inspiration from [criterion.rs](https://github.com/bheisler/criterion.rs) [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) [hyperfine](https://github.com/sharkdp/hyperfine) 

# optional/post mvp
- warmup before measuring (10-1000 iterations or 1s worth of iterations with minimum of 4-10 iterations)
- `time` option to limit benchmark to time quota
- print hardware specs (see BenchmarkDotNet)
- benchmarks grouping / baseline to compare to (see hyperfine)
- compare results with previous run (see criterion)

runtime/js/40_testing.js
bartlomieju: Measurements are now taken in Rust

bartlomieju:
Note to self: ops and resource sanitizers should be ripped out from the benchmarking setup. In preliminary tests they incurred significant cost and it's not clear that they are in anyway useful for benchmarking (might kick us out of some V8 optimizations). We can reconsider them at the later stage, but they shouldn't be added in the MVP PR.

bartlomieju:
@evanwashere this PR now doesn't use sanitizers; I expect the performance/noise problems might go away now.

bartlomieju:
This is now ready for review. I tried to keep this iteration minimal so we can land it.

Before landing an appropriate chapter should be added to the manual

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+use crate::colors;
+use crate::compat;
+use crate::create_main_worker;
+use crate::emit;
+// use crate::file_fetcher::File;
+// use crate::file_watcher;
+// use crate::file_watcher::ResolutionResult;
```
dsherret, line 9: Remove?

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+  }
+  if elapsed < 1_000 * 60 {
+    return format!("({}s)", elapsed / 1000);
+  }
+
+  let seconds = elapsed / 1_000;
+  let minutes = seconds / 60;
+  let seconds_remainder = seconds % 60;
+  format!("({}m{}s)", minutes, seconds_remainder)
+}
```
dsherret, line 183: We should share this with `test.rs` since this function is already in there and tested there.

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+    permissions,
+    true,
+  )
+  .await?;
+
+  Ok(())
+}
+
+/// Test a single specifier as documentation containing test programs, an executable test module or
+/// both.
```
dsherret, line 301: Needs updating for deno bench.

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+    specifier.clone(),
+    permissions,
+    vec![ops::bench::init(channel.clone())],
+  );
+
+  // Enable op call tracing in core to enable better debugging of op sanitizer
+  // failures.
+  worker
+    .execute_script(&located_script_name!(), "Deno.core.enableOpCallTracing();")
+    .unwrap();
```
dsherret, line 320: I think we don't want this on by default?

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+    if use_esm_loader {
+      worker.execute_side_module(&specifier).await?;
+    } else {
+      compat::load_cjs_module(
+        &mut worker.js_runtime,
+        &specifier.to_file_path().unwrap().display().to_string(),
+        false,
+      )?;
+      worker.run_event_loop(false).await?;
+    }
```
dsherret, line 337: Could we move most of this into a helper function somewhere to reduce the duplicate logic? Maybe a trait in CLI called `WorkerExtensions` that has these compat mode stuff on it.

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+
+  let join_handles = specifiers.iter().map(move |specifier| {
+    let ps = ps.clone();
+    let permissions = permissions.clone();
+    let specifier = specifier.clone();
+    let sender = sender.clone();
+    let options = options.clone();
+
+    tokio::task::spawn_blocking(move || {
+      let join_handle = std::thread::spawn(move || {
```
dsherret, line 381: I've always wondered this (IIRC I've seen this same pattern elsewhere), but why do we spawn another thread here when we already have a thread from tokio? I'm guessing this is setting some thread local state?

If so, I don't believe we need to take away a thread from tokio (spawn_blocking) to do this and could instead just work with `std::thread::spawn` directly.

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+use deno_graph::ModuleKind;
+use deno_runtime::permissions::Permissions;
+use deno_runtime::tokio_util::run_basic;
+use log::Level;
+use num_format::Locale;
+use num_format::ToFormattedString;
+use serde::Deserialize;
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::mpsc::channel;
```
dsherret, line 34: Maybe use a tokio channel instead? Or is that not possible with ops?

cli/tools/bench.rs
```diff
@@ -0,0 +1,541 @@
...
+  .await?;
+
+  Ok(())
+}
+
+pub async fn run_benchmarks_with_watch(
+  _flags: Flags,
+  _bench_flags: BenchFlags,
+) -> Result<(), AnyError> {
+  todo!()
```
dsherret, line 540: Pending `todo`.

runtime/js/40_testing.js
```diff
@@ -627,6 +630,107 @@
     ArrayPrototypePush(tests, testDef);
   }
 
+  // Main test function provided by Deno.
```
dsherret, line 633: test -> bench

runtime/js/40_testing.js
```diff
@@ -627,6 +630,107 @@
...
+      if (!name) {
+        throw new TypeError("The bench name can't be empty");
+      }
+      benchDef = { ...defaults, ...nameOrFnOrOptions, fn, name };
+    }
+
+    benchDef.fn = wrapBenchFnWithSanitizers(
+      reportBenchIteration(benchDef.fn),
+      benchDef,
+    );
```
dsherret, line 722: It might be neat if we only run the sanitizers twice—once after the warmup and once after the iterations. Might get expensive if ops tracing is enabled though.

runtime/js/40_testing.js
```diff
@@ -989,6 +1226,27 @@
...
+   *   name: string;
+   *   sanitizeExit: boolean,
+   *   warmup: boolean,
+   * }} BenchStepParams
+   */
+  class BenchStep {
+    /** @type {BenchStepParams} */
+    #params;
+
+    /** @param params {TestStepParams} */
```
dsherret, line 1240: Test -> Bench

cli/fs_util.rs
```diff
@@ -152,6 +152,19 @@ pub fn is_supported_test_path(path: &Path) -> bool {
   }
 }
 
+/// Checks if the path has a basename and extension Deno supports for tests.
```
dsherret, line 155: tests -> benches.

dsherret (commented):
Looks really good and almost there!

cli/tools/bench.rs
bartlomieju: Yep, forgot to remove it.

cli/tools/bench.rs
bartlomieju: It's a good point, but I'm still not sure if this is set in stone at this point. I'd prefer to address it some time later.

cli/tools/bench.rs
bartlomieju: That's a good point, I think this could be removed

cli/tools/bench.rs
bartlomieju: This is used across threads, I'm not sure if tokio channel would work in such case.

runtime/js/40_testing.js
bartlomieju: Yeah, I would still want to have sanitizers on benchmarks, but their perf hit is significant and it's something I wanted to leave for a follow up PRs.

cli/tools/bench.rs
bartlomieju: Any suggestion where do we put it? I also have `human_size`, maybe let's add `cli/display.rs` for such utilities?

cli/tools/bench.rs
bartlomieju: Opened https://github.com/denoland/deno/pull/13877

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -4,6 +4,311 @@
...
+     * for example via a call to `Deno.exit`. Defaults to true. */
+    sanitizeExit?: boolean;
+
+    /** Specifies the permissions that should be used to run the bench.
+     * Set this to "inherit" to keep the calling thread's permissions.
+     * Set this to "none" to revoke all permissions.
+     *
+     * Defaults to "inherit".
+     */
+    permissions?: "inherit" | "none" | {
```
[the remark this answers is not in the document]
bartlomieju: Good point, this is also shared in `WorkerOptions`. I'll take a look

cli/tools/bench.rs
bartlomieju: Opened https://github.com/denoland/deno/pull/13879

deno.json
```diff
@@ -0,0 +1,11 @@
+{
```
kt3k, line 1: Is this file intentionally included?

example_bench.ts
```diff
@@ -0,0 +1,7 @@
+Deno.bench("URL parsing", { n: 250 }, () => {
```
kt3k, line 1: This file looks like for a temporary experiment

cli/tools/bench.rs
```diff
@@ -0,0 +1,741 @@
...
+            reporter.report_output(&output);
+          }
+
+          BenchEvent::IterationStart(_iteration) => {
+            summary.current_bench.current_start = Instant::now();
+          }
+
+          BenchEvent::IterationFinish(_iteration) => {
+            let duration_of_iter =
+              Instant::now() - summary.current_bench.current_start;
```
evanwashere, line 417:
```suggestion
              summary.current_bench.current_start.elapsed();
```

cli/tools/bench.rs
```diff
@@ -0,0 +1,727 @@
...
+    let ps = ps.clone();
+    let permissions = permissions.clone();
+    let specifier = specifier.clone();
+    let sender = sender.clone();
+    let options = options.clone();
+
+    tokio::task::spawn_blocking(move || {
+      let future = bench_specifier(ps, permissions, specifier, sender, options);
+
+      run_basic(future)
```
dsherret, line 362: I was just looking at the test code yesterday with Kitson and I don't understand why we create a new tokio runtime for each specifier. Can't we just do `spawn(async move { bench_specifier(...).await })` here?

dsherret (approved):
LGTM!

cli/tools/bench.rs
bartlomieju: Because we have `--jobs` flag that allows to run testing code on multiple threads. In such case sync code would block progress in different specifiers.

bartlomieju: I guess we could do that for `bench` though; trying to think if there's a case that would prohibit using `tokio::task::spawn`

dsherret: Hmm... don't think we need to create a completely new runtime for that. If we're blocking then we could transition the task to be blocking?

dsherret: Nevermind, just had a call with Bartek and he explained to me the concept of tokio's single threaded runtime and how we're using that so we can't don't this.

Issue #13617, opened 2022-02-07
`Conn.setNoDelay()` works for TCP connections only


piscisaureus:
We should have a `TcpConn` class for this instead.

```
$ deno
>  let conn = await Deno.connectTls({hostname: "deno.land", port: 443});
undefined
>  conn.setNoDelay(true)
Uncaught BadResource: Bad resource ID
    at Object.opSync (deno:core/01_core.js:149:12)
    at TlsConn.setNoDelay (deno:ext/net/01_net.js:101:19)
    at <anonymous>:2:6
```

bartlomieju:
> We should have a TcpConn class for this instead.

That would be a breaking change to a stable API if we changed return type to `Deno.connect()`, wouldn't it? 

As a quick fix to this we could provide better error for non-TCP connections. Thoughts?

PR #13714, opened 2022-02-19
feat: Add Deno.TcpConn class, change return type from Deno.connect

bartlomieju:
Closes https://github.com/denoland/deno/issues/13617


piscisaureus:
Looks good. I think we should add UnixConn too.

PR #13725, opened 2022-02-20
feat: "deno task" subcommand

bartlomieju:
__Don't expect to land it, just a quick experiment.__

_This functionality should be compatible with `scripts` in `package.json`, but it would require `deno script --config package.json <script_name>` syntax at the moment_

Screenshot 2022-02-21 at 00 07 43

## Performance
Faster than [`npm scripts`](https://docs.npmjs.com/cli/v8/using-npm/scripts) and [`dum`](https://github.com/egoist/dum)
```
deno on  deno_scripts [!?] via 🦕 v1.19.0 via  v17.4.0 via 🦀 v1.58.1
❯ time dum foo
$ foo
$ node example_script
from example []

________________________________________________________
Executed in   53.38 millis    fish           external
   usr time   35.13 millis    0.07 millis   35.06 millis
   sys time   12.79 millis    1.70 millis   11.09 millis


deno on  deno_scripts [!?] via 🦕 v1.19.0 via  v17.4.0 via 🦀 v1.58.1
❯ time npm run foo

> foo
> node example_script

from example []

________________________________________________________
Executed in  195.03 millis    fish           external
   usr time  175.96 millis    0.06 millis  175.90 millis
   sys time   37.74 millis    1.84 millis   35.90 millis


deno on  deno_scripts [!?] via 🦕 v1.19.0 via  v17.4.0 via 🦀 v1.58.1
❯ time ./target/release/deno script foo
from example []

________________________________________________________
Executed in   42.95 millis    fish           external
   usr time   18.63 millis    0.07 millis   18.56 millis
   sys time   12.56 millis    1.81 millis   10.75 millis


```

## Examples

Example with [`docland`](https://github.com/denoland/docland):
```
// deno.json
{
    "scripts": {
        "build": "deno run --config deno.jsonc --import-map import-map.json --allow-read=. --allow-write=./static --allow-net build.ts"
    }
}
```

```
$ deno script
Available scripts:
- build
    deno run --config deno.jsonc --import-map import-map.json --allow-read=. --allow-write=./static --allow-net build.ts

$ deno script build
Script build deno run --config deno.jsonc --import-map import-map.json --allow-read=. --allow-write=./static --allow-net build.ts
Building deploy_doc...
Fetching Deno CLI releases...
Latest release: v1.19.0
Documenting latest stable APIs...
Saving latest stable APIs...
Documenting latest unstable APIs...
Saving latest unstable APIs...
Documenting lib esnext...
Saving lib esnext...
Documenting lib dom...
Saving lib dom...
Done.

```

Example with [`deno_std`](https://github.com/denoland/deno_std)
```
// deno.json
{
  "scripts": {
    "node:setup": "deno run --allow-read --allow-net --allow-write node/_tools/setup.ts",
    "node:test": "deno test --allow-read --allow-run node/_tools/test.ts"
  },
  "fmt": {
    "files": {
      "exclude": [
        ".git",
        "_wasm_crypto/target",
        "cov",
        "node/_tools/versions",
        "node/_tools/suites",
        "node/integrationtest/yarn.js",
        "wasi/testdata"
      ]
    }
  },
  "lint": {
    "files": {
      "exclude": [
        ".git",
        "_wasm_crypto/target",
        "cov",
        "node/vendor",
        "node/_module/node_modules/",
        "node/_tools/versions",
        "node/_tools/suites",
        "node/integrationtest/yarn.js",
        "wasi/testdata"
      ]
    }
  }
}
```

```
$ deno script
Available scripts:
- node:setup
    deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
- node:test
    deno test --allow-read --allow-run node/_tools/test.ts

$ deno script node:setup
Script node:setup deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
Check file:///Users/ib/dev/deno_std/node/_tools/setup.ts
Downloading https://nodejs.org/dist/v16.13.0/node-v16.13.0.tar.gz in path "versions/node-v16.13.0.tar.gz" ...
Downloaded: https://nodejs.org/dist/v16.13.0/node-v16.13.0.tar.gz into versions/node-v16.13.0.tar.gz
Decompressing node-v16.13.0.tar.gz...

Cleaning up previous tests
Copying test files...

$ deno script node:test -- test-event-emitter-check-listener-leaks.js
Script node:test deno test --allow-read --allow-run node/_tools/test.ts
Check file:///Users/ib/dev/deno_std/node/_tools/test.ts
running 2 tests from file:///Users/ib/dev/deno_std/node/_tools/test.ts
test Node.js compatibility "parallel/test-event-emitter-check-listener-leaks.js" ... ok (551ms)
test checkConfigTestFilesOrder ... ok (1ms)

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (566ms)

```


cli/tools/script.rs
```diff
@@ -0,0 +1,95 @@
...
+    return Ok(1);
+  }
+
+  let cwd = config_file_path.parent().unwrap();
+  let script_name = script_flags.script;
+  let maybe_script = scripts_config.get(&script_name);
+
+  if let Some(script) = maybe_script {
+    let (shell_name, shell_flag) = if cfg!(windows) {
+      // TODO: allow to use PowerShell?
```
dsherret, line 68: Powershell is too different from sh.

bartlomieju: Good point, thanks

dsherret: Powershell Core does, but not Windows Powershell, which I believe most people will have installed on their system by default.

cli/tools/script.rs
```diff
@@ -0,0 +1,95 @@
...
+      ("cmd", "/C")
+    } else {
+      // TODO: allow to use other shells?
+      ("sh", "-c")
+    };
+    let shell_arg = format!("{} {}", script, flags.argv.join(" "));
```
dsherret, line 74: I'm not sure we should straight pass the script to the shell. A really annoying thing about cmd and powershell is that setting the environment variables are annoying. In npm, this is typically solved by installing the cross-env package: https://www.npmjs.com/package/cross-env -- I think whatever solution we come up with should solve this.

cli/tools/script.rs
lucsoft: There is no Core/windows only anymore under Powershell 7, which is the default on latest windows

dsherret: @lucsoft default on Windows 11? A lot of people are still using Windows 10 though. For me on Windows 10, the default seems to be Windows Powershell. I had to separately install Powershell 7. This doesn't sound like a great experience if many people are still on Windows Powershell by default. Regardless, I think we probably shouldn't do any of this because there's not a nice way to set environment variables in a cross platform way (see https://github.com/denoland/deno/pull/13725/files#r813337459).

cli/tools/script.rs
bartlomieju: That's a good point and something that @lucacasonato raised to me as well. I'll look into that.

bartlomieju:
Note to self, TODO:
- [ ] update schema file for configuration file
- [ ] add integration tests
- [ ] add cross-env functionality

cli/tools/script/shell_parser.rs
```diff
@@ -0,0 +1,769 @@
...
+}
+
+pub fn parse(input: &str) -> Result<SequentialList, AnyError> {
+  fn error_for_failure(
+    e: ParseErrorFailure,
+  ) -> Result<SequentialList, AnyError> {
+    bail!(
+      "{}\n  {}\n  ~",
+      e.message,
+      e.input.chars().take(60).collect::<String>()
```
bartlomieju, line 117: Why 60?

cli/tools/script/interpreter.rs
```diff
@@ -0,0 +1,248 @@
...
+        let state = state.clone();
+        async_handles.push(tokio::task::spawn(async move {
+          let result = execute_command(item, state).await;
+          if let ExecuteCommandResult::Continue(_, _, Some(err)) = result {
+            eprintln!("{}", err);
+          }
+        }));
+      }
+      SequentialItem::Command(item) => {
+        match execute_command(item, state.clone()).await {
```
bartlomieju, line 81: Shouldn't `state` be mutated here instead of being cloned?

cli/tools/script/interpreter.rs
```diff
@@ -0,0 +1,248 @@
...
+              Ok(new_dir) => {
+                state.cwd = new_dir;
+                Ok(0)
+              }
+              Err(err) => Err(anyhow!(
+                "Could not cd to {}.\n\n{}",
+                new_dir.display(),
+                err
+              )),
+            }
```
bartlomieju, line 137: Seems like this whole block should be a separate function; wdyt?

dsherret: Yup. Already have done this while adding pipe support 🙂

cli/tools/script/shell_parser.rs
dsherret: I can add a comment. This is just a simple implementation that truncates to prevent word wrap.

cli/tools/script/interpreter.rs
dsherret: I don't think lifetimes would allow it. Within `execute_command` it returns a boxed future because `execute_command` is async recursive. So for that reason, I have it returning a list of changes to apply to the state after returning (there's a bug with cd that I've fixed in my local)

paambaati:
@bartlomieju This looks fantastic. A quick question, if I may — is referring other scripts possible?

For example, say I have to chain multiple scripts to one single convenience script, is that supported?

dsherret:
@paambaati yup that is possible with the current implementation.

cli/config_file.rs
```diff
@@ -530,6 +530,7 @@ pub struct ConfigFileJson {
   pub import_map: Option<String>,
   pub lint: Option<Value>,
   pub fmt: Option<Value>,
+  pub scripts: Option<Value>,
```
dsherret, line 533: I think we should name everything "task"/"tasks". It's shorter than scripts and pretty easy to switch to.

bartlomieju: Yes, good idea. I'll do that.

bartlomieju: Done

cli/tools/task.rs
```diff
@@ -62,17 +62,14 @@ pub async fn execute_script(
   let maybe_script = tasks_config.get(&task_name);
 
   if let Some(script) = maybe_script {
-    let seq_list = deno_task_shell::parser::parse(script)
+    // todo(THIS PR): append additional CLI args to "script" here
+    // in order to support someone doing `deno task my_task arg1 arg2 etc`
```
bartlomieju, line 66: Available as `flags.argv`

dsherret: Thanks!

cli/tools/task.rs
```diff
@@ -0,0 +1,167 @@
...
+  } else {
+    bail!("No config file found")
+  }
+}
+
+fn print_available_tasks(tasks_config: HashMap<String, String>) {
+  eprintln!("{}", colors::green("Available tasks:"));
+
+  let mut task_names: Vec<String> = tasks_config.clone().into_keys().collect();
+  task_names.sort();
```
dsherret, line 45: I think we should leave tasks in the order they appear in the file. They might have some logical top to bottom flow between tasks and sorting it might break that flow. Thoughts?

cli/tools/task.rs
```diff
@@ -17,9 +17,21 @@ fn get_tasks_config(
...
+        if key.is_empty() {
+          bail!("Configuration file task names cannot be empty");
+        } else if !key
+          .chars()
+          .all(|c| c.is_ascii_alphanumeric() || matches!(c, '_' | '-'))
+        {
+          bail!("Configuration file task names must only contain alpha-numeric characters, underscores (_), or dashes (-). Task: {}", key);
+        } else if !key.chars().next().unwrap().is_ascii_alphabetic() {
+          bail!("Configuration file task names must start with an alphabetic character. Task: {}", key);
+        }
```
bartlomieju, line 30: 👍 nice validation

bartlomieju (commented):
It seems `cli/tests/testdata/task/task_echo.out` is not used anywhere?

cli/tools/task.rs
bartlomieju: I was thinking the same. OTOH it's easier to scan for something when they are ordered, but I guess it's better to leave the order as users intended

dsherret (approved):
LGTM now.

PR #13735, opened 2022-02-21
chore: use rustfmt cli via dprint for faster format

dsherret:
This PR uses dprint to orchestrate formatting via rustfmt.

Before: 6384ms
After: 932ms

### DIFF
--- a/.dprint.json
+++ b/.dprint.json
@@ -9,8 +9,12 @@
   "json": {
     "deno": true
   },
+  "exec": {
+    "associations": "**/*.rs",
+    "rustfmt": "rustfmt"
+  },
   "includes": [
-    "**/*.{ts,tsx,js,jsx,json,md,toml}"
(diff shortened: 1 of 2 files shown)


.dprint.json
```diff
@@ -9,8 +9,12 @@
   "json": {
     "deno": true
   },
+  "exec": {
+    "associations": "**/*.rs",
+    "rustfmt": "rustfmt"
```
dsherret, line 14: The exec plugin has no knowledge of `rustfmt`. This makes it calls the `rustfmt` cli and format over stdin.

https://github.com/dprint/dprint-plugin-exec#configuration

lucacasonato (approved):
Very cool!

dsherret:
A nice side effect is that now rust files get automatically formatted on save if using dprint's vscode extension.

PR #13738, opened 2022-02-22
fix(permissions): don't panic when checking permissions containing relative path with deleted parent

### LINKED ISSUE #13737: Panic when removing a directory that you're in.
```ts
function nextdir() {
  return `test${Date.now()}`;
}
const pathname = `${Deno.cwd()}/${nextdir()}`;
Deno.mkdirSync(pathname);
Deno.chdir(pathname);
Deno.removeSync(pathname);
Deno.mkdirSync('X', { recursive: true });  // this is the line which causing panick
```

When running the above code, deno panick.

```bash
$ deno run --unstable -A test.ts
Check file:///Users/feng/dev/deno_std/test.ts

Platform: macos x86_64
Version: 1.19.0
Args: ["deno", "run", "--unstable", "-A", "test.ts"]

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failed to get current working directory

Caused by:
    No such file or directory (os error 2)', runtime/permissions.rs:1440:46
```

I encounter this problem when I run  https://github.com/nodejs/node/blob/master/test/parallel/test-fs-mkdir.js#L219 on `--compat` mode, so I make a minimal reproduction code to report here.

```
$ deno --version     
deno 1.19.0 (release, x86_64-apple-darwin)
v8 9.9.115.7
typescript 4.5.2
```

crowlKats:
Fixes #13737


runtime/permissions.rs
```diff
@@ -2874,4 +2886,21 @@ mod tests {
...
+    std::fs::create_dir(&tmp).unwrap();
+    std::env::set_current_dir(&tmp).unwrap();
+    let pathname = std::path::Path::new("./foo");
+    std::fs::create_dir(pathname).unwrap();
+    std::env::set_current_dir(pathname).unwrap();
+    std::fs::remove_dir(std::env::current_dir().unwrap()).unwrap();
+    assert!(perms.write.check(&pathname.join("x")).is_err());
+    assert!(perms.read.check(&pathname.join("x")).is_err());
+    assert!(perms.ffi.check(Some(&pathname.join("x"))).is_err());
+  }
```
nayeemrmn, line 2905: This cwd modification cannot be run concurrently with pretty much any other test in the same process. I would just do an integration test.

Issue #12800, opened 2021-11-17
Proposal: add import maps to the config file


kitsonk:
## Context

Deno currently supports a config file and Deno currently supports import maps.  Deno does not currently support "combining" import maps and the config file.  This leads to potentially complicated command line invocations.

In addition the feature of `deno vendor` is being added (see: #13346) as well as there is other tooling which automates the building of import maps within Deno.

As of Deno 1.18, the configuration file is automatically detected, that coupled with significant use of the community of import maps, it is a significantly better developer experience to be able to store the import map in the configuration file and be able to elide that on the command line, just as users can elide the `--config` file option now.

_Edit: This proposal has been updated from its original form_

## Semantics

- If a config file applies, and there is a top level key named `"importMap"`, the value of the key will be interpreted as a path or URL using the config file as a base.
- If the flag `--import-map` is supplied on the command line, the value will be used as it is today.  If a config file applies and contains an import map, the user will be warned that the value on the command line argument is being ignored, unless the value of `importMap` and `--import-map` resolve to the same resource.

_Originally the semantics included support for the top level `"imports"` and `"scopes"` keys in the configuration file. This has been removed from this proposal and can be revisited in the future._

## Examples

```jsonc
{
  "compilerOptions": {
    "lib": [ "dom", "dom.iterable", "deno.ns" ]
  },
  "importMap": "./import_map.json"
}
```

```jsonc
{
  "compilerOptions": {
    "lib": [ "dom", "dom.iterable", "deno.ns" ]
  },
  "importMap": "https://example.com/import-map.json"
}
```

## Considerations

- The semantics around `--import-map` or the import map in a config file taking precedence is debatable to the author of the proposal and the author isn't tied to one or the other, but we should defined semantics, and warn the user if there is one or the other being ignored. _This was discussed in the core team, and it was agreed that `--import-map` should take precedence over the configuration file_
- There maybe considerations related to this proposal with #12764. For example, a task should be able to define its own import map. _This was discussed in the core team, and agreement was that if #12764 moves forward, it will need to address this consideration, as well as other related considerations_
[...]

PR #13739, opened 2022-02-22
feat: allow specification of import map in config file

kitsonk:
Closes: #12800

### DIFF
--- a/cli/config_file.rs
+++ b/cli/config_file.rs
@@ -1,6 +1,5 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-use crate::flags;
 use crate::fs_util::canonicalize_path;
 use crate::fs_util::specifier_parent;
 use crate::fs_util::specifier_to_file_path;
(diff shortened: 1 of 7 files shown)


cli/config_file.rs
```diff
@@ -227,6 +228,56 @@ pub fn json_merge(a: &mut Value, b: &Value) {
   }
 }
 
+/// Based on flags and an optional configuration file, return a resolved module
+/// specifier to an import map.
+pub fn resolve_import_map_specifier(
+  flags: &flags::Flags,
```
bartlomieju, line 234: Could we take `flag.import_map_path` as `maybe_import_map_path` instead? That way we wouldn't have to import `flags` module here - seems preferable not to couple the two

bartlomieju (commented):
Great, I have only a single nitpick.

cli/config_file.rs
```diff
@@ -227,6 +227,56 @@ pub fn json_merge(a: &mut Value, b: &Value) {
...
+/// Based on an optional command line import map path and an optional
+/// configuration file, return a resolved module specifier to an import map.
+pub fn resolve_import_map_specifier(
+  maybe_import_map_path: Option<&str>,
+  maybe_config_file: Option<&ConfigFile>,
+) -> Result<Option<ModuleSpecifier>, AnyError> {
+  if let Some(import_map_path) = maybe_import_map_path {
+    if let Some(config_file) = &maybe_config_file {
+      if config_file.to_import_map_path().is_some() {
+        log::warn!("{} the configuration file \"{}\" contains an entry for \"importMap\" that is being ignored.", crate::colors::yellow("Warning"), config_file.specifier);
```
dsherret, line 239: If someone explicitly provides an import map as a cli flag I'm not sure we need to warn them about it because it should hopefully be obvious. Maybe we should warn though... not sure.

dsherret (approved):
LGTM. Will be much nicer to use import maps now.

cli/config_file.rs
bartlomieju: I think it's fine, users can always specify `--quiet` to silence these warnings

bartlomieju (approved):
LGTM

cli/config_file.rs
kitsonk: Because we have auto detection now, I think it is important, as you can be there scratching your head why your code isn't working. As noted `--quiet` can suppress it.

PR #13740, opened 2022-02-22
feat(ext/net): support cert, key options in listenTls

### LINKED ISSUE #12639: [feat. request] Allow reading CA certificate and key file from a variable instead of a file.
Currently `ListenTlsOptions` require a path to the cert and key files:

https://github.com/denoland/deno/blob/b7341438f29de88f3458b32a835bfad560bda52e/ext/net/lib.deno_net.d.ts#L89-L94

It would be great if these could be read from a variable.

So that these could be loaded from environment variables.

kt3k:
This PR adds `cert: string` and `key: string` options to `Deno.listenTls` and deprecates `certFile` and `keyFile` options.

You can specify the cert and private key from any source now. The read permission is checked only when `certFile` or `keyFile` is specified

closes #12639

### DIFF
--- a/cli/tests/unit/tls_test.ts
+++ b/cli/tests/unit/tls_test.ts
@@ -15,6 +15,9 @@ import { TextProtoReader } from "../../../test_util/std/textproto/mod.ts";
 
 const encoder = new TextEncoder();
 const decoder = new TextDecoder();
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM

PR #13743, opened 2022-02-23
refactor: remove dead code from lsp

kitsonk:
In #12968 `#[allow(unused)]` was added, but never removed.  Since then there is a significant amount of dead code and other things that have creeped into the whole of the lsp.  This PR cleans all that up.


bartlomieju (approved):
LGTM!

PR #13745, opened 2022-02-23
fix(cli): disable config discovery for remote script

### LINKED ISSUE #13744: Proposal: disable auto-discovery of config file for `deno run <remote entrypoint>`
If the users have custom typescript settings which are incompatible with default settings (e.g. frontend typings), they will see type errors when they execute the remote Deno scripts.

For example, a user develops frontend lib. `deno.json` has frontend types (e.g. `"lib": ["dom", "esnext"]`, etc). When they want to run, for example, `udd` command in the CI, the command throws type error because of the incompatible type settings of the auto-discovered `deno.json`. (This happens, for example, in https://github.com/hayd/deno-udd/issues/59, and they avoid the error by removing deno.json in CI https://github.com/UltiRequiem/timero/commit/56dc4ba0b64f7deb19607d4b476be88279fcd6ac . This doesn't look great.)

This situation doesn't make sense because `deno.json` is usually only for local scripts. The config is not intended to be applied to all deno programs.

So I suggest we should disable auto-discovery of config file when the command is `deno run <remote entrypoint>`, which fixes the above situation.

kt3k:
This implements the idea proposed in #13744.

This PR disables the config auto-discovery when executing remote scripts (`deno run <remote_script>`)

closes #13744

### DIFF
--- a/cli/config_file.rs
+++ b/cli/config_file.rs
@@ -164,17 +164,18 @@ const CONFIG_FILE_NAMES: [&str; 2] = ["deno.json", "deno.jsonc"];
 pub fn discover(flags: &crate::Flags) -> Result<Option<ConfigFile>, AnyError> {
   if let Some(config_path) = flags.config_path.as_ref() {
     Ok(Some(ConfigFile::read(config_path)?))
-  } else {
+  } else if let Some(config_path_args) = flags.config_path_args() {
     let mut checked = HashSet::new();
-    for f in flags.config_path_args() {
+    for f in config_path_args {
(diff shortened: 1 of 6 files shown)


ry (commented):
Could you add a test showing that with a remote script, the config file isn't auto-detected?

kt3k:
> Could you add a test showing that with a remote script, the config file isn't auto-detected?

Added integration test cases

bartlomieju (approved):
LGTM

ry (approved):
👍

Issue #13749, opened 2022-02-23
TextEncoder is slow


ry:
https://twitter.com/jarredsumner/status/1496455856224972808
https://github.com/denoland/deno/blob/ee27b9dd7e8ef311c6bfaa3264f7b9908a784a24/core/bindings.rs#L920-L936 

Please add some `cargo bench` benchmarks.

littledivy:
`Deno.core.encode` fell victim to https://github.com/denoland/deno/issues/13772

ry:
I think this has been addressed?

lino-levan:
@littledivy I might be wrong, but I don't think this is an issue anymore. Performance seems to have improved a bit since this issue was opened. Maybe it could be faster?

<img width="793" alt="image" src="https://github.com/denoland/deno/assets/11367844/6c266111-1c98-4a17-a43f-bf3902dd4f34">

bartlomieju:
This doesn't seem to be an issue anymore. Deno's `TextEncoder` is now faster than Node.js across all test cases:

| Test | Deno | Node |
|------|------|------|
| encode small x1M | 614ms | 594ms |
| encode medium x1M | 799ms | 2,575ms |
| encode large x10K | 204ms | 1,931ms |
| encodeInto large x10K | 30ms | 1,021ms |

Deno is comparable on small strings and significantly faster (3-30x) on medium/large strings.

bartlomieju:
Closing as this is no longer an issue. Deno's `TextEncoder` is now faster than Node.js across all test cases:

| Test | Deno | Node |
|------|------|------|
| encode small x1M | 614ms | 594ms |
| encode medium x1M | 799ms | 2,575ms |
| encode large x10K | 204ms | 1,931ms |
| encodeInto large x10K | 30ms | 1,021ms |

Deno is comparable on small strings and significantly faster (3-30x) on medium/large strings.

PR #13750, opened 2022-02-23
bench: core.encode/decode

AaronO:
Benches for #13749

M1 baseline:
```
test bench_utf8_decode_12_b  ... bench:         204 ns/iter (+/- 13)
test bench_utf8_decode_12_kb ... bench:       3,189 ns/iter (+/- 1,049)
test bench_utf8_decode_12_mb ... bench:   1,430,887 ns/iter (+/- 29,186)
test bench_utf8_encode_12_b  ... bench:         433 ns/iter (+/- 13)
test bench_utf8_encode_12_kb ... bench:       5,786 ns/iter (+/- 693)
test bench_utf8_encode_12_mb ... bench:   4,946,268 ns/iter (+/- 118,515)
```

### DIFF
--- a/bench_util/Cargo.toml
+++ b/bench_util/Cargo.toml
@@ -20,3 +20,7 @@ tokio = { version = "1.10.1", features = ["full"] }
 [[bench]]
 name = "op_baseline"
 harness = false
+
+[[bench]]
+name = "utf8"
+harness = false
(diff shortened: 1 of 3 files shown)


AaronO:
> Maybe these benches belong in deno_core?
> 
> The original benchmark was using an 11mb string? Maybe we should do that too?

We don't have any other benches in `deno_core`, functionally all of `deno_core`'s benches live in `bench_util/`, e.g: see `op_baseline.rs`.

`bench_util` provides wraps `bencher` to simplify JS/Deno-op tests and profiling via flamebench, etc...
(if you just naively profile bencher benches it will be noisy due to setup/teardowns)

Will add the 11MB test

ry (approved):
LGTM

PR #13755, opened 2022-02-24
Statically link to libc

littledivy:


littledivy:
Screenshot 2022-02-24 at 10 16 35 AM

Tests seem to fail with SIGABRT

PR #13756, opened 2022-02-24
fix(compile): Support import maps

### LINKED ISSUE #13748: `deno compile` ignores ` --import-map` in 1.19.0
Using 1.18.2 I can run this:

```
deno compile -A --no-check --unstable --import-map ./modules.json --output test test.js
```

and it works.

In 1.19.0 the same thing doesn't work. Well, the compilation works, but running the executable will throw an error like this one, which is familiar to anyone who forgot using the `--import-map` flag once:

```
error: Relative import path "std/path" not prefixed with / or ./ or ../ from "file:///path/to/test.js"
```

This might be related to https://github.com/denoland/deno/pull/13563. Seems like import map support was forgotten in the change. Also, `deno bundle` seems unaffected by this, but it's obviously not the same feature.

littledivy:
Fixes #13748

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -396,19 +396,21 @@ async fn compile_command(
 ) -> Result<i32, AnyError> {
   let debug = flags.log_level == Some(log::Level::Debug);
 
-  let run_flags =
-    tools::standalone::compile_to_runtime_flags(&flags, compile_flags.args)?;
+  let run_flags = tools::standalone::compile_to_runtime_flags(
+    &flags,
+    compile_flags.args.clone(),
+  )?;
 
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM, good fix

cli/main.rs
```diff
@@ -423,6 +425,8 @@ async fn compile_command(
     generic_error("There should only be one reference to ModuleGraph")
   })?;
 
+  graph.valid().unwrap();
```
lucacasonato, line 428: This should not be a panic…

bartlomieju: Yikes, good catch, you are right

PR #13761, opened 2022-02-24
perf(encode): avoid allocations

### LINKED ISSUE #13749: TextEncoder is slow

littledivy:
Closes #13749

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -929,10 +929,21 @@ fn encode(
       return;
     }
   };
-  let text_str = text.to_rust_string_lossy(scope);
-  let zbuf: ZeroCopyBuf = text_str.into_bytes().into();
 
-  rv.set(to_v8(scope, zbuf).unwrap())
+  let capacity = text.utf8_length(scope);
+  let backing_store =
+    v8::ArrayBuffer::new_backing_store(scope, capacity).make_shared();
+  let length = text.write_utf8(
+    scope,
+    unsafe { &mut *(&backing_store[..capacity] as *const _ as *mut [u8]) },
+    None,
+    v8::WriteOptions::NO_NULL_TERMINATION
+      | v8::WriteOptions::REPLACE_INVALID_UTF8,
+  );
+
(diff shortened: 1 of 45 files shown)


ext/console/02_console.js
```diff
@@ -2093,6 +2093,7 @@
     };
 
     timeEnd = (label = "default") => {
+      const endTime = DateNow();
```
littledivy, line 2096: https://github.com/denoland/deno/issues/13772 :/

core/bindings.rs
```diff
@@ -929,10 +929,21 @@ fn encode(
       return;
     }
   };
-  let text_str = text.to_rust_string_lossy(scope);
-  let zbuf: ZeroCopyBuf = text_str.into_bytes().into();
 
-  rv.set(to_v8(scope, zbuf).unwrap())
+  let capacity = text.utf8_length(scope);
```
aapoalas, line 933: I kind of already know this to be a not-good-idea but:

`v8::String` can (if externalizable) be turned into `v8::String::ExternalStringresource` which can be used to to get to get the actual string buffer data. Now that buffer can be used to create a `v8::BackingStore` and then sent onwards to a `v8::ArrayBuffer`. => Zero copy encoding.

Of course, this means that the AB now owns the string buffer and I'm very much unsure if the original `v8::String` can deal with that, much less what happens if the AB now makes changes to the buffer data.

lucacasonato: @aapoalas This doesn't work, because the string is internally stored as WTF-16 by V8, not UTF-8. Conversion always needs to happen.

aapoalas: Ah, very good point. It was a bad idea anyway due to the various potential corner cases mentioned :)

littledivy:
This is a bit stale now. Closing.

PR #13768, opened 2022-02-24
fix: properly follow symlinked specifiers

crowlKats:


cli/file_fetcher.rs
```diff
@@ -158,11 +158,14 @@ fn fetch_local(specifier: &ModuleSpecifier) -> Result<File, AnyError> {
   let media_type = MediaType::from(specifier);
 
   Ok(File {
+    specifier: ModuleSpecifier::from_file_path(
+      deno_runtime::fs_util::canonicalize_path(local.as_path())?,
```
dsherret, line 162: We should add context here in case it errors:

https://github.com/denoland/deno/blob/5f280662042054e8272225ef81f2be3445a6b057/cli/tools/vendor/mod.rs#L85-L87

I kind of wonder if we should push that down into `fs_util::canonicalize_path` though. Thoughts? Maybe for another PR?

dsherret (approved):
LGTM! Hopefully the windows CI passes 😄

cli/file_fetcher.rs
crowlKats: > I kind of wonder if we should push that down into `fs_util::canonicalize_path` though

yea, would make sense

> Maybe for another PR?

yea

PR #13769, opened 2022-02-24
feat(ext/http): auto-compression of fixed response bodies

kitsonk:
Work in progress from a pairing session between myself and @ry

Inspired by previous work of @satyarohith on Deno Deploy.


ext/http/lib.rs
```diff
@@ -497,22 +514,145 @@ async fn op_http_write_headers(
...
+    match &*key.to_ascii_lowercase() {
+      b"cache-control" => {
+        let value = std::str::from_utf8(value.as_ref()).unwrap();
+        if let Some(cache_control) = CacheControl::from_value(value) {
+          // We skip compression if the cache-control header value is set to
+          // "no-transform"
+          if cache_control.no_transform {
+            headers_allow_compression = false;
+          }
+        }
```
lucacasonato, line 534:
```suggestion
        if let Some(value) = std::str::from_utf8(&value).ok() {
          if let Some(cache_control) = CacheControl::from_value(value) {
            // We skip compression if the cache-control header value is set to
            // "no-transform"
            if cache_control.no_transform {
              headers_allow_compression = false;
            }
          }
        } else {
          headers_allow_compression = false;
        }
```

ext/http/lib.rs
```diff
@@ -497,22 +514,145 @@ async fn op_http_write_headers(
...
+      }
+      b"content-type" => {
+        if !value.is_empty() {
+          content_type_header = Some(value.clone());
```
lucacasonato, line 538: This clone should not be necessary.

ext/http/lib.rs
```diff
@@ -497,22 +514,145 @@ async fn op_http_write_headers(
...
+      }
+      b"content-encoding" => {
+        // we don't compress if a content-encoding header was provided
+        headers_allow_compression = false;
+      }
+      // we store the values of ETag and Vary and skip adding them for now, as
+      // we may need to modify or change.
+      b"etag" => {
+        if !value.is_empty() {
+          etag_header = Some(value.clone());
```
lucacasonato, line 549: Again no clone needed.

ext/http/lib.rs
```diff
@@ -497,22 +514,145 @@ async fn op_http_write_headers(
...
+          continue;
+        }
+      }
+      b"vary" => {
+        if !value.is_empty() {
+          vary_header = Some(value.clone());
```
lucacasonato, line 555: And same here.

ext/http/lib.rs
```diff
@@ -497,22 +514,145 @@ async fn op_http_write_headers(
...
   match data {
     Some(data) => {
-      // If a buffer was passed, we use it to construct a response body.
-      body = builder.body(data.into_bytes().into())?;
+      // Set Vary: Accept-Encoding header for direct body response.
+      // Note: we set the header irrespective of whether or not we compress the
+      // data to make sure cache services do not serve uncompressed data to
+      // clients that support compression.
+      let vary_value = if let Some(value) = vary_header {
+        let value_str = std::str::from_utf8(value.as_ref()).unwrap();
```
lucacasonato, line 579: Unsafe unwrap.

ext/http/lib.rs
```diff
@@ -497,22 +514,145 @@ async fn op_http_write_headers(
...
+      );
+
+      let should_compress =
+        body_compressible && data.len() > 20 && accepts_compression;
+
+      if should_compress {
+        // If user provided a ETag header for uncompressed data, we need to
+        // ensure it is a Weak Etag header ("W/").
+        if let Some(value) = etag_header {
+          let value_str = std::str::from_utf8(value.as_ref()).unwrap();
```
lucacasonato, line 602: Again unsafe unwrap.

lucacasonato (commented):
This is looking great! @kitsonk what do you think about adding an option to disable this behaviour to a `Deno.serveHttp` options bag?

ext/http/lib.rs
```diff
@@ -497,22 +514,156 @@ async fn op_http_write_headers(
...
+
+      // Set the user provided ETag & Vary headers for a streaming response
+      if let Some(value) = etag_header {
+        builder = builder.header("etag", value.as_ref());
+      }
+      if let Some(value) = vary_header {
+        builder = builder.header("vary", value.as_ref());
+      }
+
       let (body_tx, body_rx) = Body::channel();
```
lucacasonato, line 667: Can you add a `// TODO` here for streaming compression?

lucacasonato (commented):
This is looking great! @kitsonk what do you think about adding an option to disable this behaviour to a `Deno.serveHttp` options bag?

Also do we need to have special handling for `Range`?

kitsonk:
> This is looking great! @kitsonk what do you think about adding an option to disable this behaviour to a `Deno.serveHttp` options bag?

I thought about it when I paired with @ry and he argued against it.  It does add complexity to it. Is there a compelling case where someone would want to disable it, other than "because they can"?  Doesn't setting `Cache-Control: no-transform` effectively provide that functionality without adding to the API surface?  That is how CFW supports disabling it.

lucacasonato:
I could imagine that one has an nginx in front of Deno that is responsible for on the fly encoding. In that case you want Deno to leave the responses as "pure" as possible. I don't think we per-se _need_ to add this for the PR to land, but I think it would be useful to give users an opt-out. The implementation complexity seems rather trivial.

kitsonk:
> Also do we need to have special handling for `Range`?

Actually, after doing a bit of research, if a `Content-Range` header is present in the response, we should disable automatic compression, as the user HTTP server has processed a range request, with the range being based off the uncompressed range headers it negotiated.

lucacasonato (approved):
LGTM

@satyarohith PTAL too

cli/tests/unit/http_test.ts
```diff
@@ -1165,14 +1165,513 @@ Deno.test(
...
+      );
+      await respondWith(response);
+      httpConn.close();
+      listener.close();
+    }
+
+    async function client() {
+      const url = `http://${hostname}:${port}/`;
+      const cmd = [
+        "curl",
```
satyarohith, line 1209: nit: should use `fetch()`?

cli/tests/unit/http_test.ts
```diff
@@ -1165,14 +1165,513 @@ Deno.test(
...
+      assert(output.includes("content-encoding: gzip\r\n"));
+      proc.close();
+    }
+
+    await Promise.all([server(), client()]);
+  },
+});
+
+Deno.test({
+  name: "http server doesn't compresses body when no-transform is set",
```
satyarohith, line 1502:
```suggestion
  name: "http server doesn't compress body when no-transform is set",
```

cli/tests/unit/http_test.ts
```diff
@@ -1165,14 +1165,513 @@ Deno.test(
...
+      assert(!output.includes("content-encoding: "));
+      proc.close();
+    }
+
+    await Promise.all([server(), client()]);
+  },
+});
+
+Deno.test({
+  name: "http server doesn't compresses body when content-range is set",
```
satyarohith, line 1556:
```suggestion
  name: "http server doesn't compress body when content-range is set",
```

ext/http/compressible.rs
```diff
@@ -0,0 +1,660 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+use deno_core::ByteString;
+
+// Data obtained from https://github.com/jshttp/mime-db/blob/fa5e4ef3cc8907ec3c5ec5b85af0c63d7059a5cd/db.json
+// Important! Keep this list sorted alphabetically.
+const CONTENT_TYPES: &[&str] = &[
```
satyarohith, line 7: Nit: there are some new additions to https://github.com/jshttp/mime-db/blob/master/db.json; we might want to update the list.

feel free to skip it

cli/tests/unit/http_test.ts
kitsonk: No, because `fetch()` as per the spec doesn't allow control of the `Accept-Encoding` header, as it is one of the forbidden headers and Deno sets is specifically, and automatically decodes the body before sending it back in the `Response()`.  We need to a more direct method for testing purposes.

satyarohith (approved):
LGTM.

Thanks for working on it! :)

ext/http/compressible.rs
kitsonk: Let's skip it for now, as they are very likely very esoteric media types that aren't widely used.

Longer term, someone should either find an efficient way to statically include the JSON in, or generate the code statically from the JSON file.

PR #13770, opened 2022-02-24
feat: deno test --trace-ops

bartlomieju:
Follow up to https://github.com/denoland/deno/pull/13676

While the feature proves to be very useful it comes with a significant performance
hit, it's caused by excessive source mapping of stack frames. Given that it's non-trivial
to cache results of source mapping without unbounded consumption of memory after 
discussion with @piscisaureus we agreed that the simplest solution is
to activate this feature with a flag.

<s>Still need to add a test for it</s>


runtime/js/40_testing.js
```diff
@@ -208,9 +209,16 @@
...
 
-      throw `Test case is leaking async ops.
+      let msg = `Test case is leaking async ops.
 
-- ${ArrayPrototypeJoin(details, "\n - ")}`;
+      - ${ArrayPrototypeJoin(details, "\n - ")}`;
+
+      if (!isOpCallTracingEnabled) {
+        msg +=
+          `\nTo get more details where ops are leaked run again with --trace-ops flag.`;
```
dsherret, line 218: First thing I scrolled for. Good 👍

dsherret (approved):
LGTM and makes a lot of sense.

runtime/js/40_testing.js
```diff
@@ -34,6 +34,7 @@
     TypeError,
   } = window.__bootstrap.primordials;
 
+  let isOpCallTracingEnabled = false;
```
dsherret, line 37: Maybe we should ask the other module for this state so we have a single source of truth?

bartlomieju: That was my initial thought - but there might be some setup code execute before we call `Deno[Deno.internal].runTests()` and thus if we don't do it this way then we won't be able to collect traces for the setup code.

bartlomieju: Nvm, I see what you mean. I switched to using a function in `Deno.core` instead.

PR #13779, opened 2022-02-26
feat(runtime/permissions): support IP CIDR ranges in net allowlist

nebarf:
Fixes https://github.com/denoland/deno/issues/9816


nebarf:
@lucacasonato about your comment https://github.com/denoland/deno/pull/11509#pullrequestreview-737631637, do you still suggest to turn NetDescriptor into an enum? At a first glance it would require a bit of refactoring

Issue #11836, opened 2021-08-24
Consider enabling `"noImplicitOverride"` by default


kitsonk:
In TypeScript 4.3, the [`override` keyword](https://devblogs.microsoft.com/typescript/announcing-typescript-4-3-beta/#override-and-the-noimplicitoverride-flag) was introduced which ensure when type checking code that a method with the `override` keyword is present in the base class.  This helps make it more explicit when extending classes.  In addition, there is the flag `"noImplicitOverride"` which enforces that if a method is present in a base class and the extended class that the `override` keyword is present in the extended class, helping ensure that users don't "accidentally" override a method unintentionally.

I am personally not sure about this, as it has a lot of potential to break existing code, but I wanted to bring it up so that we can try to consider what "version" of TypeScript we want to support in Deno.  The existing behaviour allows people to opt-in to using `override` for the protection it offers, but essentially doesn't deliver as "safe" as code as possible.

getspooky:
I think it's a good to enable `noImplicitOverride` in order to to improve code health in the repository. but as you already said it will introduce a lot of code refactoring. A good example of `noImplicitOverride` integration is [Angular](https://github.com/angular/angular/pull/42512).

David-Else:
I think it is one of the best improvements they ever made! I always hated inheritance due to not being sure if the extended class method was overriding the base class. It would be great to enable `noImplicitOverride`.

sigmaSd:
for future searchers to enable this add to deno.json
```json
{
  "compilerOptions": {
    "noImplicitOverride": true
  }
}
```
(take from https://github.com/denoland/deno_std/issues/1933)

sigmaSd:
Maybe this can be considered for deno 2.0 ?

sigmaSd:
though I just discovered this issue https://github.com/microsoft/TypeScript/issues/47250 which makes this feature in my opinion partially finished

lucacasonato:
We're doing this for Deno 2.

dsherret:
I think this one hasn't been decided yet or implemented, so reopening.

PR #13780, opened 2022-02-26
chore: Add "noImplicitOverride" to config-file.v1.json

elisee:
This PR adds the definition for the `"noImplicitOverride"` compiler option to the `deno.json` config file schema so it can be auto-completed in VS Code.

The definition was taken from http://json.schemastore.org/tsconfig, as mentioned in the original issue #11885. Related issue about possibly enabling `"noImplicitOverride"` by default: #11836.


kitsonk (approved):
LGTM

PR #13782, opened 2022-02-27
fix(cli/tools): colors do not properly detect not tty

### LINKED ISSUE #12224: console.* methods do not properly detect non-tty output and still add colors to output

antoniomuso:
I don't test this for now, because I don't know where to put these kinds of tests.

fix #12224


kt3k:
I think overriding `use_color` (NO_COLOR) based on tty is not a good idea. Because we override it, we also need a new env var FORCE_COLOR here, and having two env vars both controlling color is confusing.

kt3k:
> Do you think it is better to have different functions use_color and is_tty and check directly in the main? Or do you think it is better to add a new variable in bootstrap options called is_tty?

I think we need both. First check `IS_TTY` and store its value (you already did it), and expose it as `is_tty` function. The main module checks `is_tty` function and pass it as bootstrap option next to noColor option (You need to extend `BootstrapOptions` struct). Finally set the no color option of console module to `noColor || !isTty` in `runtime/js/99_main.js` script ref: https://github.com/denoland/deno/blob/4bea1d06c7ddb177ed20e0f32b70d7ff889871ab/runtime/js/99_main.js#L583

kt3k:
BTW you can test this behavior by passing `stdout: "piped"` options to `Deno.run` command.

```ts
const p = Deno.run({
  cmd: [
    Deno.execPath(),
    "eval",
    "console.log(1)",
  ],
  stdout: "piped",
});
const output = await p.output();
```

This should make is_tty = false

runtime/js/99_main.js
```diff
@@ -574,13 +574,14 @@ delete Object.prototype.__proto__;
...
       noColor,
+      isTty,
       pid,
       ppid,
       unstableFlag,
       cpuCount,
     } = runtimeOptions;
 
-    colors.setNoColor(noColor);
+    colors.setNoColor(noColor || !isTty);
```
kt3k, line 584: Nice 👍

antoniomuso:
I put the test, I don't know if it is the right place. Let me know!

antoniomuso:
I don't know if `ci / test release ubuntu-20.04-xl` fail because I miss in test to add `{ permissions: { run: true } }`

kt3k:
The test case looks good to me. Nice work!

kt3k (approved):
LGTM

PR #13785, opened 2022-02-27
fix(test): typecheck blocks annotated with long js/ts notations

### LINKED ISSUE #13783: `deno test --doc` doesn't pick up `typescript`-annotated blocks
This works as expected and throws an error:

\`\`\`ts
let str: string = 0;
\`\`\`

```
let str: string = 0;
    ~~~
    at file:///C:/Users/GJZwiers/repos/test_test/thingy.md$3-6.ts:1:5
```

But it doesn't if the `typescript` annotation is used:

\`\`\`typescript
let str: string = 0;
\`\`\`

GJZwiers:
Closes #13783


bartlomieju:
@GJZwiers looks good! The test failure seems to be unrelated. Could you add some test case to exercise this change?

GJZwiers:
> @GJZwiers looks good! The test failure seems to be unrelated. Could you add some test case to exercise this change?

I've added a test, the pipeline keeps failing on a few compat tests but other than it seems fine. Will this be OK?

bartlomieju:
@GJZwiers it seems the errors were coming from `test_util/std` git submodule which was mistakenly changed in this PR. I reverted that change and tests seem to pass locally now.

bartlomieju (approved):
LGTM, thank you

Issue #13790, opened 2022-02-28
compat: cjs to esm translation is not applied when a commonjs module is dynamically imported with `file://` scheme


kt3k:
`mocha` formats the importing specifier to `file://`-scheme url before calling `import()` ref: https://github.com/mochajs/mocha/blob/547ffd73535088322579d3d2026432112eae3d4b/lib/nodejs/esm-utils.js#L7

If the file url is passed to `import()` and the module is a commonjs, then cjs->esm translation is not applied to that file in compat mode.

bartlomieju:
@kt3k I'm gonna need some more data points for this one. I opened #13792 that does what you described and it works fine.

PR #13792, opened 2022-02-28
fix(compat): cjs/esm interop for dynamic imports

bartlomieju:
This commit fixes CJS/ESM interop in compat mode for dynamically
imported modules.

"ProcState::prepare_module_load" was changed to accept a list
of "graph roots" without associated "module kind". That module kind
was always hardcoded to "ESM" which is not true for CJS/ESM interop -
a CommonJs module might be imported using "import()" function. In
such case the root of the graph should have "CommonJs" module kind
instead of "ESM".

Fixes https://github.com/denoland/deno/issues/13790


bartlomieju:
Culprit is here:
https://github.com/denoland/deno/blob/a41d399f5f728c73b652c3f5f5c2bf57d2054153/cli/module_loader.rs#L109-L120

Root of the graph is always marked as ESM, which in this instance is not correct.

bartlomieju:
@kitsonk could you take a look? I feel we should be using resolver inside `ModuleLoader::prepare_load` and pass resolved specifier and module kind based on the `ResolveResponse` instead of hard coding module kind to `Esm`.

bartlomieju:
ping @kitsonk

cli/proc_state.rs
```diff
@@ -252,13 +253,46 @@ impl ProcState {
...
+    // One might argue that this is a code smell, and I would agree. However
+    // due to flux in "Node compatibility" it's not clear where it should be
+    // decided what `ModuleKind` is decided for root specifier.
+    let roots = roots
+      .into_iter()
+      .map(|r| {
+        if let Some(resolver) = &maybe_resolver {
+          let response =
+            resolver.resolve(r.as_str(), &Url::parse("unused:").unwrap());
+          // TODO(bartlomieju): this should be implemented in `deno_graph`
```
kitsonk, line 284: Ref: denoland/deno_graph#140

kitsonk (approved):
LGTM

cli/proc_state.rs
bartlomieju: Thanks for opening this!

Issue #1195, opened 2018-11-15
Support SO_REUSEADDR


ry:
Because of the lack of this, many networking tests resort to using different ports.

ry:
Feb 24, 2020: Still an issue that we want to solve.

bartlomieju:
To implement this feature, `net2` crate should be added to dependencies and [`reuse_address`](https://docs.rs/net2/0.2.33/net2/struct.TcpBuilder.html#method.reuse_address) method should be called before creating a socket.

attliaLin:
Is it still be an issue? Can I take it?

bartlomieju:
Marking as good first issue. For example how to implement this see https://github.com/denoland/deno/pull/13103

Trolloldem:
@bartlomieju I tried to add the requested feature. I decided to not use `net2` as suggested, since the library is now deprecated. 
In order to achieve the same functionality I used the `socket2` crate, which is already a dependency listed in `ext/net/Cargo.toml`.

For now I have a working unit test for `op_net_listen`, now I have to adapt the changes for the TLS version. 
In order to achieve retro compatibility, the proposed method is composed as follows:

I should be able to open a PR in the following days. But first I have a question:
```
const listener1 = Deno.listen({ hostname: "127.0.0.1", port: 3500, reuseAddress: true});
const listener2 = Deno.listen({ hostname: "192.168.1.7", port: 3500, reuseAddress: true});
```
where the `reuseAddress` field is optional and defaults to `false`.
 
- in this is issue is mentioned that  support for **SO_REUSEADDR** is needed
- this options allows the use case in which the same host has two different ip (127.0.0.1 and 192.168.1.7 for example)
- the exposed use case shows that without this feature it is necessary to use different ports
- I think that it in order to fully achieve the wanted result it is necessary to support the  **SO_REUSEPORT** option
- in this way it is possible to support the following code:
 ```
const listener1 = Deno.listen({ hostname: "127.0.0.1", port: 3500, reusePort: true});
const listener2 = Deno.listen({ hostname: "127.0.0.1", port: 3500, reusePort: true});
```
where the `reuseAddress` field is optional and defaults to `false`.

Let me know if you think that these changes could be interesting and if it is the case to open a PR to add them.

bartlomieju:
@Trolloldem thanks for looking into this. What you described sounds good to me and the proposed API looks good. `socket2` is indeed preferable in this case.

[...]

PR #13794, opened 2022-02-28
feat(net): added reuseAddress and reusePort in Sockets API

Trolloldem:
Support for **SO_REUSEADDR** socket option is requested in #1195, to avoid using different ports during tests. In order to achieve the wanted feature, it now possible to use the **SO_REUSEPORT** socket option in addition.

The implemented changes are:
1. 2 additional optional arguments for `Deno.listen` and `Deno.listenTls`, namely `reuseAddr` and `reusePort`  (one for each supported option)
2. changed the `socket2` entry to support the **SO_REUSEPORT** option
3. proposed a test to verify that the `reusePort` option works as intended

In this moment, the option is checked and set only for `op_net_listen` and `op_tls_listen`. In particular, for `op_net_listen`, changes are applied only to the `listen_tcp` method.

After testing `reuseAddr`, it has been noted that the `tokio::net::lookup_host` function use to solve the provided _hostname_, sometimes resolves the address `0.0.0.0` to `127.0.0.1`, leading to unwanted effects when the following code is used:

```
const listener1 = Deno.listen({ hostname: "127.0.0.1", port: 3500, reuseAddress: true});
const listener2 = Deno.listen({ hostname: "0.0.0.0", port: 3500, reuseAddress: true});
```
One possible solution could be using the dedicated [dns-lookup](https://docs.rs/dns-lookup/latest/dns_lookup/) package

### DIFF
--- a/ext/net/ops_tls.rs
+++ b/ext/net/ops_tls.rs
@@ -1090,6 +1090,8 @@ where
   let socket = Socket::new(domain, Type::STREAM, None)?;
   if reuse_address.is_some() {
     socket.set_reuse_address(reuse_address.unwrap())?;
+  } else {
+    socket.set_reuse_address(true)?;
   }
   if reuse_port.is_some() {
     socket.set_reuse_port(reuse_port.unwrap())?;
--- a/ext/net/ops_tls.rs
+++ b/ext/net/ops_tls.rs
@@ -1034,7 +1034,7 @@ where
   let cert = args.cert.as_deref();
   let key = args.key.as_deref();
   let reuse_address = args.reuse_address;
-  let reuse_port = args.reuse_port; 
+  let reuse_port = args.reuse_port;
 
   {
     let permissions = state.borrow_mut::<NP>();
@@ -1082,7 +1082,11 @@ where
   let bind_addr = resolve_addr_sync(hostname, port)?
     .next()
     .ok_or_else(|| generic_error("No resolved address found"))?;
-  let domain = if bind_addr.is_ipv4() { Domain::IPV4 } else { Domain::IPV6 };
+  let domain = if bind_addr.is_ipv4() {
+    Domain::IPV4
+  } else {
+    Domain::IPV6
+  };
   let socket = Socket::new(domain, Type::STREAM, None)?;
   if reuse_address.is_some() {
     socket.set_reuse_address(reuse_address.unwrap())?;
@@ -1094,7 +1098,7 @@ where
   socket.bind(&socket_addr)?;
   socket.listen(128)?;
   socket.set_nonblocking(true)?;
-  let std_listener : std::net::TcpListener = socket.into();
+  let std_listener: std::net::TcpListener = socket.into();
   let tcp_listener = TcpListener::from_std(std_listener)?;
   let local_addr = tcp_listener.local_addr()?;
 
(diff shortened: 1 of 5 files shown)


ext/net/ops_tls.rs
```diff
@@ -1075,8 +1082,25 @@ where
...
+  } else {
+    socket.set_reuse_address(true)?;
+  }
+  if reuse_port.is_some() {
+    socket.set_reuse_port(reuse_port.unwrap())?;
+  }
+  let socket_addr = socket2::SockAddr::from(bind_addr);
+  socket.bind(&socket_addr)?;
+  socket.listen(128)?;
+  socket.set_nonblocking(true)?;
```
bartlomieju, line 1102: I believe these are only available for unix systems which causes CI failure

Trolloldem: I think this is the case. I noticed that **SO_REUSEADDR** in windows seems to achieve the same results of **SO_REUSEPORT** in Linux. I will add a `#[cfg(unix)]` to address this issue.
I also noticed that once swapped to `Socket2` for the creation of `TcpListener`, I had to set **SO_REUSEADDR** to `true` by default in order to pass the `net_test` unit tests.

bartlomieju: @Trolloldem that's a bit concerning. May I suggest to open a new PR that first switches to `socket2` crate and then add new APIs in follow up?

bartlomieju: That makes sense, that's why I'm suggesting to split the PR into two - so we can unsure that when changing the crates we keep the same behavior

Trolloldem: Ok, I will open a new PR with only the changes needed for `socket2`. When it is ready I will close this one, and propose the changes for **SO_REUSEADDR** in a new one

Trolloldem: Before starting to work on the new PR, I wanted to check how the current `std::net::TcpListener` is configured. I used the following code:

```
fn listen_tcp(
  state: &mut OpState,
  addr: SocketAddr,
) -> Result<(u32, SocketAddr), AnyError> {
  let std_listener1 = std::net::TcpListener::bind(&addr)?;
  let so: socket2::Socket = socket2::Socket::from(std_listener1);
  if so.reuse_address()? {
    panic!("The reuseADDR is: {}", so.reuse_address()?);
  }
  let std_listener: std::net::TcpListener = so.into();
  std_listener.set_nonblocking(true)?;
  let listener = TcpListener::from_std(std_listener)?;
  let local_addr = listener.local_addr()?;
  let listener_resource = TcpListenerResource {
    listener: AsyncRefCell::new(listener),
    cancel: Default::default(),
  };
  let rid = state.resource_table.add(listener_resource);

  Ok((rid, local_addr))
}
```
And the I run:
`target/debug/deno test --allow-all --unstable --location=http://js-unit-tests/foo/bar cli/tests/unit/net_test.ts `
The obtained output is the following:

```
running 33 tests from file:///home/oldem/git/deno/cli/tests/unit/net_test.ts
test netTcpListenClose ...

Platform: linux x86_64
Version: 1.19.1
Args: ["target/debug/deno", "test", "--allow-all", "--unstable", "--location=http://js-unit-tests/foo/bar", "cli/tests/unit/net_test.ts"]

thread '<unnamed>' panicked at 'The reuseADDR is: true', ext/net/ops.rs:434:5
```

So, at least on Linux, it does seem that in this moment the socket already has SO_REUSEADDR option configured.
It is still preferable to separate the two PR?

PR #13798, opened 2022-03-01
[WIP] feat(lsp): add testing API

kitsonk:
This is a WIP PR which will add support for a custom Testing API based around what is offered via vscode.  For clients which support the experimental `testingApi` capability, the Deno language server will process testing API requests and provide testing API notifications.

The current state is that it compiles and generally works.  It requires a client that understands the protocol.

There is still a lot to do:

- [ ] Filter tests properly.
- [ ] Support the "steps" API for tests.
- [ ] Add redirection of stdout and stderr to the cli test APIs, as writing to stdout directly from a test will cause LSP issues.
- [ ] Implement debugging runs.
- [ ] Implement converting the client config to flags/permissions.
- [ ] Validate and handle validation issues with the test plan when running a module.
- [ ] Better structure the output of failed test cases.
- [ ] Lots of tests to test the testing API.

There are also a few things that the Testing API in vscode supports that the internal cli/test API doesn't support, that might be good ideas for the future.  The custom Testing API between Deno lsp and the client support these, so it is just a matter of adding it to the cli/test API:

- The testing API differentiates between a test error and a test failure.  For example, a test that could not be run do some sort of error, like the test factory failed for some reason is an "Error" versus that test function throwing would be a "Failure".
- The testing API for a failure or error supports an expected output versus an actual output, processing the diff inside the editor.
- The testing API supports both markdown and plaintext messages on failures and errors.
- The testing API supports a location for errors or failures to associate with the message.  This could be parsed out of a stack trace itself at the top level.
- The virtual console output (where current `TestOutput::Console(s)` gets sent to) can also associate a location with the output, meaning it could in theory link back to the console log statement that generated it in the code.


cli/lsp/testing.rs
```diff
@@ -0,0 +1,1959 @@
...
+          lsp_custom::TestModuleNotificationParams {
+            text_document: lsp::TextDocumentIdentifier { uri: specifier },
+            kind: lsp_custom::TestModuleNotificationKind::Insert,
+            label,
+            tests: vec![desc.into()],
+          },
+        ))
+        .unwrap_or_else(|err| {
+          lsp_log!("{}", err);
+        });
```
dsherret, line 952: It would be less verbose if we got rid of the channel and instead created a function like the following:

```rs
fn send_testing_notification(&self, notification: TestingNotification) {
    let client = client.clone();
    tokio::task::spawn(async move {
      client.send_test_notification(notification).await;
    });
}
```

I kind of wonder if we should just do that internally in `client` though.

cli/lsp/testing.rs
```diff
@@ -0,0 +1,1959 @@
...
+                      message: TestRunProgressMessage::End,
+                    },
+                  ))
+                  .await;
+              }
+            }
+            Some(RunRequest::Cancel(id)) => {
+              if let Some(token) = tokens.get(&id) {
+                token.cancel();
+              }
```
dsherret, line 1685: I don't think it's possible for the cancellation to occur here. I think we only want to do run `RunRequest::Start` on this thread, so the message to the channel would contain `(id, token)`. Then `TestServer` would hold the tokens and cancel them on the other thread.

cli/lsp/testing.rs
```diff
@@ -0,0 +1,1959 @@
...
+          _ => (),
+        }
+        Some((name, steps))
+      }
+      _ => None,
+    }
+  } else {
+    None
+  }
+}
```
dsherret, line 437: This file is very large. We might want to consider splitting it up into separate modules in an `lsp/testing` folder (ex. this code analysis code could maybe be in its own file).

dsherret:
I'm not done reviewing this, but I have some comments and it's a large PR, so I'm submitting what I have for now and will continue reviewing later.

dsherret (commented):
Hit the wrong button in the github.dev view.

cli/lsp/testing.rs
```diff
@@ -0,0 +1,1959 @@
...
+    span: Span,
+    parent: String,
+    level: usize,
+    steps: Option<Vec<TestDefinition>>,
+  ) -> Self {
+    let id = checksum::gen(&[
+      parent.as_bytes(),
+      &level.to_be_bytes(),
+      name.as_bytes(),
+    ]);
```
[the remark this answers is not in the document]
kitsonk: Yes, last one wins, but there is no other way to disambiguate tests that share the same name.

cli/lsp/testing.rs
kitsonk: Done.

cli/lsp/testing.rs
kitsonk: Move it to the client method.

cli/lsp/testing.rs
```diff
@@ -0,0 +1,1959 @@
...
+        steps
+          .iter()
+          .map(|step| step.as_test_data(source_text_info))
+          .collect()
+      }),
+      range: span_to_range(&self.span, source_text_info),
+    }
+  }
+
+  fn find_step(&self, name: &str, level: usize) -> Option<&TestDefinition> {
```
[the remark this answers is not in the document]
kitsonk: Yes, but this is a limitation of how steps are declared and the impact is "minimal", in the sense that this only comes into play when:

- The step was not statically discovered
- The step with the same name at the same depth within the same top level test exists

The outcome would be that the runtime identified step would not have its result reported against it, but instead the earlier duplicate.

To fix this, we need globally unique IDs to come from the harness for tests and steps, which is outside of the scope of this PR.

cli/lsp/testing.rs
kitsonk: Yeah, you are right.  I will refactor.

kitsonk: I'm not sure why I made it so complicated... I have refactored it and cleaned it up quite a bit.

cli/lsp/client.rs
```diff
@@ -164,6 +177,31 @@ impl ClientTrait for LspowerClient {
...
+          )
+          .await,
+        TestingNotification::Progress(params) => client
+          .send_custom_notification::<testing_lsp_custom::TestRunProgressNotification>(
+            params,
+          )
+          .await,
+      }
+    });
+  }
```
dsherret, line 203: 👍

dsherret (approved):
LGTM! It looks great now and that refactor helped make this PR much easier to read. This will be a very nice feature.

kitsonk:
I've landed it.  It being so big, I am sure there are some things missed, but we can polish it iteratively and deal with issues when they come up.

PR #13799, opened 2022-03-01
feat(ext/web): Add reportError()

### LINKED ISSUE #13484: window.reportError(e)?
The web platform supports `self.reportError(e)` to trigger an uncaught error with semantics similar to `setTimeout(() => { throw e; })`.

Would it make sense for Deno to implement this API?

nayeemrmn:
Blocked on #14159.
Closes #13484.


core/error.rs
```diff
@@ -190,84 +191,84 @@ impl JsError {
...
-
-        // Convert them into Vec<JsStackFrame>
-        let frames: Vec<JsStackFrame> = match frames_v8 {
-          Some(frames_v8) => {
-            serde_v8::from_v8(scope, frames_v8.into()).unwrap()
-          }
-          None => vec![],
-        };
-        (message, frames, stack, cause)
+    if is_instance_of_error(scope, exception) {
```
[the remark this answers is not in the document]
nayeemrmn: I think it's simpler if anything with this change, it more cleanly separates the branches for error instances and non-error instances.

tools/wpt/expectation.json
```diff
@@ -3286,15 +3286,11 @@
       "scripting": {
         "reporterror.any.html": [
           "self.reportError(1)",
-          "self.reportError(TypeError)",
-          "self.reportError(undefined)",
-          "self.reportError() doesn't invoke getters"
+          "self.reportError(undefined)"
```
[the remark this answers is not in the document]
nayeemrmn: https://github.com/web-platform-tests/wpt/blob/66d28833f7b932a1575531cb4b035a18047f8ad2/html/webappapis/scripting/reporterror.any.js#L12-L14

We don't provide location info for thrown values which aren't `Error` objects. Browsers usually fall back to the throw site in these situations, or in this case the `reportError()` call site. That's what the tests are expecting us to do. Of course that's trivial for browsers which implement `reportError()` as a direct native binding, since `v8::Exception::create_message(scope, exception)` would "just work". But for us we need to manually analyse the stack trace within `reportError()`.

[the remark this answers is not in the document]
nayeemrmn: > But for us we need to manually analyse the stack trace within `reportError()`.

Done https://github.com/denoland/deno/pull/13799/commits/73aa722a4771e99a1d8ef115f90cff60c2589021.

bartlomieju:
@nayeemrmn I think this is good implementation wise. I want to get some more eyes on it before landing.

@crowlKats @lucacasonato please take a look

core/bindings.rs
```diff
@@ -1329,6 +1334,16 @@ fn queue_microtask(
   };
 }
 
+fn destructure_error(
```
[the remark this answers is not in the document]
bartlomieju: I'm not super happy with a new binding, but if we can use it in listed PRs then let's do it 👍

ext/web/16_report_error.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  function checkThis(thisArg) {
+    if (thisArg !== null && thisArg !== undefined && thisArg !== globalThis) {
+      throw new TypeError("Illegal invocation");
+    }
+  }
+
+  function reportError(error) {
+    checkThis(this);
```
crowlKats, line 33: We dont do this for `fetch`, so not sure how correct this is. @lucacasonato

ext/web/lib.deno_web.d.ts
```diff
@@ -889,3 +889,22 @@ declare class DecompressionStream {
...
+ * ```ts
+ * addEventListener("error", (event) => {
+ *   event.preventDefault();
+ * });
+ * reportError(new Error("foo")); // Will not be reported.
+ * ```
+ * In Deno, this error will terminate the process if not intercepted like above.
+ */
+declare function reportError(
+  error: Error,
```
crowlKats, line 909: [TypeScript types this as `any`](https://github.com/microsoft/TypeScript/blob/e64f04bd8c87ece7f800d10d8eb90b7e592907ad/src/lib/webworker.generated.d.ts#L5188)

nayeemrmn: Ah, I was 99% sure it would be `any` but couldn't find that and went with the forward-compatible option for now. There it is.

ext/web/16_report_error.js
nayeemrmn: Yeah me neither, I took it from `setTimeout()`

tools/wpt/expectation.json
```diff
@@ -3153,20 +3153,6 @@
...
-          "self.reportError(undefined)",
-          "self.reportError() doesn't invoke getters"
-        ],
-        "reporterror.any.worker.html": [
-          "self.reportError(1)",
-          "self.reportError(TypeError)",
-          "self.reportError(undefined)",
-          "self.reportError() doesn't invoke getters"
-        ]
-      },
```
crowlKats, line 3169: This doesnt seem right; we should still test these, not completely remove the tests

tools/wpt/expectation.json
```diff
@@ -3155,16 +3155,10 @@
       },
       "scripting": {
         "reporterror.any.html": [
-          "self.reportError(1)",
-          "self.reportError(TypeError)",
-          "self.reportError(undefined)",
-          "self.reportError() doesn't invoke getters"
+          "self.reportError(undefined)"
```
nayeemrmn, line 3158: https://github.com/web-platform-tests/wpt/issues/33206

lucacasonato (commented):
@bartlomieju and me have looked into this PR. The implementation is good. We are a little concerned about that this PR adds the `onerror` event handling from [report the exception](https://html.spec.whatwg.org/#report-the-exception) but only for the `reportError` codepath. There are some other codepaths which are meant to use this "report the exception" handling (for example timers and event targets), but we currently don't do that. I think that is rather confusing.

I think we may want to split this out into two PRs: one that adds the infrastructure for "report an exception" and wires it up to timers and event targets, and then a second PR that actually adds `reportError`, built on top of the "report an exception" infrastructure. WDYT?

core/bindings.rs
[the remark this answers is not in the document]
bartlomieju: @nayeemrmn could I ask you to factor this bit into a separate PR? Turns out this is very handy to be used in a test runner output refactor that I'm working on, and I could immediately improve reporting by skipping internal frames (starting with `deno:`) if we could send structured error output to Rust

ext/web/16_report_error.js
```diff
@@ -0,0 +1,72 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 1: I think you could safely merge this into `02_event.js` and get rid of the circular problem of storing `__bootstrap` namespace. WDYT?

core/runtime.rs
```diff
@@ -1235,9 +1247,19 @@ impl JsRuntime {
...
-        generic_error("Cannot evaluate module, because JavaScript execution has been terminated.")
-      )).expect("Failed to send module evaluation error.");
+      let explicit_terminate_error =
+        state_rc.borrow_mut().explicit_terminate_error.take();
+      if let Some(error) = explicit_terminate_error {
+        let error = v8::Local::new(tc_scope, error);
+        #[allow(unused_must_use)]
+        {
+          sender.send(exception_to_err_result(tc_scope, error, false));
+        }
```
bartlomieju, line 1257: Why this `allow`? `let _ = sender.send(...)` should do the job, right?

tools/wpt/expectation.json
```diff
@@ -3154,18 +3154,8 @@
...
-          "self.reportError() doesn't invoke getters"
-        ],
-        "reporterror.any.worker.html": [
-          "self.reportError(1)",
-          "self.reportError(TypeError)",
-          "self.reportError(undefined)",
-          "self.reportError() doesn't invoke getters"
-        ]
+        "reporterror.any.html": false,
+        "reporterror.any.worker.html": false
```
nayeemrmn, line 3158: We can enable these by adding `e.preventDefault()` calls to https://github.com/web-platform-tests/wpt/blob/3518e60485b37ada91e3ab08daaa0591c4d16c4f/html/webappapis/scripting/reporterror.any.js, it wouldn't interfere with the tests. Not blocking though

bartlomieju:
@nayeemrmn please rebase

ext/web/lib.deno_web.d.ts
```diff
@@ -890,3 +890,22 @@ declare class DecompressionStream {
   readonly readable: ReadableStream<Uint8Array>;
   readonly writable: WritableStream<Uint8Array>;
 }
+
+/** Dispatch an uncaught exception. Similar to:
+ * ```ts
+ * setTimeout(() => { throw error; }, 0);
+ * ```
```
lucacasonato, line 897: Can you mention that unlike `setTimeout`, the error reporting is synchronous?

tools/wpt/expectation.json
lucacasonato: Let's do it. Do you want to open a PR to WPT, or shall I?

nayeemrmn: > Let's do it. Do you want to open a PR to WPT, or shall I?

Uhh you can if you have it set up.

lucacasonato: https://github.com/web-platform-tests/wpt/pull/33657

bartlomieju (approved):
LGTM

lucacasonato (approved):
LGTM

lucacasonato (changes requested):
Actually, WPTs don't pass with my WPT patch:

```
Going to run 2 test files.
----------------------------------------
/html/webappapis/scripting/reporterror.any.html

test self.reportError(1) ... ok
test self.reportError(TypeError) ... ok
test self.reportError(undefined) ... failed
test self.reportError() (without arguments) throws ... ok
test self.reportError() doesn't invoke getters ... ok
test stderr:

file result: failed. the event loop run out of tasks during the test
```

nayeemrmn:
@lucacasonato https://github.com/denoland/deno/pull/13799#discussion_r828046502. Chrome also fails that right now. We can get compliant FireFox behaviour by adding an internal escape hatch to `ErrorEvent`. I'd like to either do that in a follow up PR or wait for resolution on the spec side in case they go with option 1.

lucacasonato:
My concern is not that failing test, but the "event loop ran out of tasks" thing.

nayeemrmn:
> My concern is not that failing test, but the "event loop ran out of tasks" thing.

Running locally I got varied results... eventually I reproduced this but now it's happening for every test even after I switch `wpt` back to main branch 😓I'm going to try in CI here.

nayeemrmn:
> My concern is not that failing test, but the "event loop ran out of tasks" thing.

It seems like it's not related to this PR, but some change in wpt since https://github.com/web-platform-tests/wpt/commit/96f98c458cd36e1f12b6807b22a3a9ee1e1b0326 (deno main) and wpt master is making this happen to every test.

bartlomieju:
Discussed with @lucacasonato, he'll try to fix the WPT runner before merging this PR

lucacasonato:
@nayeemrmn Can you rebase once more? WPT is updated on `main`, and should now be working once again.

tools/wpt/expectation.json
```diff
@@ -4237,7 +4227,6 @@
     "opening-handshake": {
       "003-sets-origin.worker.html": false
     },
-    "Create-on-worker-shutdown.any.worker.html": false,
```
nayeemrmn, line 4240: This test hangs, since `t.done()` is never called. On main it panics and satisfies the expected file failure but that panic fixed in this PR. Once #14318 is addressed we can revert to this since it would hit the `the event loop run out of tasks during the test` thingy.

nayeemrmn:
@lucacasonato Ready

lucacasonato (approved):
LGTM!

lucacasonato:
Sorry this one took so long to land @nayeemrmn. Thank you very much for this work!

KnorpelSenf:
Given https://github.com/denoland/deno/pull/13987#issuecomment-1113649837, this seems to be the only other PR which affects networking in the 1.21.0 release. Could this be responsible for #14437?

KnorpelSenf:
`fetch` is still not able to send files reliably. Does anyone know if this is actually the offending PR? Are there maybe any other changes in https://github.com/denoland/deno/releases/tag/v1.21.0 that could have broken it?

It seems to me like `fetch` is a relatively important thing for Deno and I'm a bit puzzled that one cannot trust it to send buffers and byte streams, as it seems to be messing with the data in some ways. In case you're not willing to look into this, could you shed some light on why #14437 isn't a priority?

PR #13804, opened 2022-03-01
fix(repl): fix null eval result

### LINKED ISSUE #13757: Value null is displayed as undefined in repl
The Deno repl treats null as undefined in the loop output:
```
$ deno --version
deno 1.19.0 (release, x86_64-unknown-linux-gnu)
v8 9.9.115.7
typescript 4.5.2

$ deno repl
> null
undefined
> const a=null
undefined
> a
undefined
> a===undefined
false
> String(a)
"null"
```

kt3k:
In inspector's protocol `undefined` value is represented as `{}` (empty object) and `null` is represented as `{"value":null}` in JSON.

serde_json doesn't distinguish between these two (`property with null value` and `absent property`) by default, so the both case becomes `value: None` (= `undefined`) at deserialization, which caused #13757

This PR changes the deserialization method for `RemoteObject`'s `value` property. Now `null` value is deserialized as `value: Some(Null)`, which works as `null` when passed back to v8, and therefore fixes #13757


bartlomieju (approved):

ry:
Nice work

PR #13805, opened 2022-03-01
feat(ext/crypto): AES-GCM support for 128bit IVs

### LINKED ISSUE #13689: Web Crypto AES-GCM 256 with 16 byte IV throws Initialization vector length not supported
Using 16 byte IV fails with AES-256-GCM.

It throws:

```
NotSupportedError: Initialization vector length not supported
```

Reproduction:

```js
const key = await crypto.subtle.generateKey(
  {
    name: "AES-GCM",
    length: 256,
  },
  false,
  ["encrypt", "decrypt"],
);

const iv = crypto.getRandomValues(new Uint8Array(16));

const msg = "Hello World!";

const encrypted = await crypto.subtle.encrypt(
  {
    name: "AES-GCM",
    iv,
    additionalData: new Uint8Array(),
  },
  key,
  new TextEncoder().encode(msg),
);

const decrypted = await crypto.subtle.decrypt(
  {
    name: "AES-GCM",
    iv,
    additionalData: new Uint8Array(),
  },
  key,
  encrypted,
);

console.log(new TextDecoder().decode(decrypted));
```

The same code works fine in browser (Edge).

But if you change IV to this

```js
const iv = crypto.getRandomValues(new Uint8Array(12));
```

It will work in Deno as well.

**Another bug** I noticed is `additionalData` is marked optional in typings but an error is thrown at runtime if we don't specify it:
```
TypeError: Error parsing args: missing field `additionalData`
```
However, not specifying it in browser just works as well.

deno v1.18.2

littledivy:
Closes #13689


piscisaureus (approved):
LGTM!

PR #13811, opened 2022-03-02
chore: cancel old in-progress CI workflows

littledivy:
CI will run only on the latest commit in a PR. Saves energy 🌲 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

Idea from https://github.com/nodejs/node/pull/42017

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,10 @@ name: ci
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: ${{ matrix.job }} ${{ matrix.profile }} ${{ matrix.os }}


kt3k (approved):
LGTM. Nice!

lucacasonato (commented):
Can we not do this on main? On main we want CI for every commit so we get all the canary releases.

kt3k:
@lucacasonato 
github doc says `github.head_ref is only defined on pull_request events`. So the expression `${{ github.head_ref || github.run_id }}` should fallback to `github.run_id` (unique for each run) at `push` event, and won't prevent the main build (to my understanding)

lucacasonato:
One note: this does make committing multiple commits to find flaky tests more difficult.

lucacasonato (approved):

bartlomieju:
> One note: this does make committing multiple commits to find flaky tests more difficult.

Yeah basically in such case we'll need to push empty commit after the previous one finished. Let's see how it works, but I'm skeptical

PR #13814, opened 2022-03-02
chore: Use actions-rs/toolchain instead of dtolnay/rust-toolchain

ry:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,12 +88,8 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-
-      - name: Install clippy and rustfmt
-        if: matrix.job == 'lint'
-        run: |
-          rustup component add clippy
-          rustup component add rustfmt
+          components: rustfmt, clippy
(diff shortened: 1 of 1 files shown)


littledivy (changes requested):
Why? `dtolnay/rust-toolchain@stable` uses the toolchain from `rust-toolchain.toml`.

.github/workflows/ci.yml
```diff
@@ -85,13 +85,11 @@ jobs:
           tar --exclude=".git*" --exclude=target --exclude=third_party/prebuilt \
               -czvf target/release/deno_src.tar.gz -C .. deno
 
-      - uses: dtolnay/rust-toolchain@stable
-
-      - name: Install clippy and rustfmt
-        if: matrix.job == 'lint'
-        run: |
-          rustup component add clippy
-          rustup component add rustfmt
```
dsherret, line 94: We could remove this and put...

```toml
components = ["rustfmt", "clippy"]
```

...in rust-toolchain.toml in the root. Then only do:

```
- uses: dtolnay/rust-toolchain@stable
```

Issue #13657, opened 2022-02-13
[LSP] randomly getting Internal error -32603


lts20050703:
```
unexpected positions
start_pos: Position { line: 85, character: 7 }
end_pos: Position { line: 86, character: 1 }
[Error - 11:14:53 AM] Request textDocument/semanticTokens/full failed.
  Message: Internal error
  Code: -32603
```

### Version:

deno 1.18.2+911ddbc (canary, x86_64-unknown-linux-gnu)
v8 9.9.115.5
typescript 4.5.2

vscode: 1.65.0-insider
extension: 3.10.1

### Note: I am using VSCode WSL because path autocomplete on VSCode Windows is broken (see denoland/vscode_deno#623) but this also happen randomly on VScode Windows

dsherret:
I've also run into this. I'm not sure how to reproduce. Aaron ran into it on Friday.

```
unexpected positions
start_pos: Position { line: 53, character: 71 }
end_pos: Position { line: 54, character: 1 }
[Error - 6:06:58 p.m.] Request textDocument/semanticTokens/full failed.
  Message: Internal error
  Code: -32603 
unexpected positions
start_pos: Position { line: 63, character: 62 }
end_pos: Position { line: 64, character: 14 }
[Error - 6:06:58 p.m.] Request textDocument/semanticTokens/range failed.
  Message: Internal error
  Code: -32603
```

kitsonk:
It is likely an out of date `line_index` is being used when the semantic tokens are being converted from tsc items to lsp ones.

David and I had a debate though under what conditions this could occur.

lucacasonato:
Got some errors with the new logging:

```
unexpected positions
specifier: file:///mnt/starship/Projects/github.com/lucacasonato/deno-gfm/mod.ts
open: true
start_pos: Position { line: 18, character: 16 }
end_pos: Position { line: 19, character: 2 }
[Error - 1:47:14 AM] Request textDocument/semanticTokens/full failed.
  Message: Internal error
  Code: -32603 
unexpected positions
specifier: file:///mnt/starship/Projects/github.com/lucacasonato/deno-gfm/mod.ts
open: true
start_pos: Position { line: 18, character: 16 }
end_pos: Position { line: 19, character: 2 }
[Error - 1:47:15 AM] Request textDocument/semanticTokens/range failed.
  Message: Internal error
  Code: -32603 
unexpected positions
[...]

PR #13815, opened 2022-03-02
chore(lsp): log more for "unexpected positions" lsp error

dsherret:
This logs more for #13657.

We're pretty sure that the issue is a closed document's `line_index` is getting out of date. Fixing it would require a bit of an overhaul (I believe it would require keeping filesystem files in memory for the duration of a request if requested), so just want to be sure this is indeed the case.

### **Merge on review**


kitsonk (approved):
LGTM

PR #13633, opened 2022-02-09
feat(npm): implement Node API

littledivy:
This PR implements the NAPI for loading native modules into Deno. 

Closes https://github.com/denoland/deno/issues/15717

## Feature tracking
- [x] Environment life cycle APIs
  - [x] `napi_set_instance_data`
  - [x] `napi_get_instance_data`
- [x] Basic Node-API data types
  - [x] `napi_status`
  - [x] `napi_extended_error_info`
  - [x] `napi_env`
  - [x] `napi_value`
  - [x] `napi_threadsafe_function`
  - [x] `napi_threadsafe_function_release_mode`
  - [x] `napi_threadsafe_function_call_mode`
  - [x] Node-API memory management types
    - [x] `napi_handle_scope`
    - [x] `napi_escapable_handle_scope`
    - [x] `napi_ref`
    - [x] `napi_type_tag`
    - [x] `napi_async_cleanup_hook_handle`
  - [x] Node-API callback types
    - [x] `napi_callback_info`
    - [x] `napi_callback`
    - [x] `napi_finalize`
    - [x] `napi_async_execute_callback`
    - [x] `napi_async_complete_callback`
    - [x] `napi_threadsafe_function_call_js`
    - [x] `napi_async_cleanup_hook`
- [x] Error handling
  - [x] Return values
    - [x] `napi_get_last_error_info`
  - [x] Exceptions
    - [x] `napi_throw`
    - [x] `napi_throw_error`
    - [x] `napi_throw_type_error`
    - [x] `napi_throw_range_error`
    - [x] `node_api_throw_syntax_error`
    - [x] `napi_create_error`
    - [x] `napi_create_type_error`
    - [x] `napi_create_range_error`
    - [x] `node_api_create_syntax_error`
    - [x] `napi_get_and_clear_last_exception`
    - [x] `napi_is_exception_pending` (stubbed)
    - [x] `napi_fatal_exception`
  - [x] Fatal errors
    - [x] `napi_fatal_error`
- [ ] Object lifetime management
  - [ ] Making handle lifespan shorter than that of the native method
    - [x] `napi_open_handle_scope`
    - [x] `napi_close_handle_scope`
    - [ ] `napi_open_escapable_handle_scope` (stubbed)
    - [ ] `napi_close_escapable_handle_scope` (stubbed)
    - [ ] `napi_escape_handle` (stubbed)
  - [ ] References to objects with a lifespan longer than that of the native method
[...]

PR #13816, opened 2022-03-02
feat(core): Event loop middlewares for Extensions

littledivy:
Add "event loop middlewares" to `deno_core::Extension`. Allows extension crates to _do stuff_ when the event loop is spinning, returning `false` from the middleware will prevent the event loop from exiting. This paves way for async FFI callbacks and NAPI...without bringing any specific code to deno_core.

```rust
Extension::builder()
  .event_loop_middleware(|state, cx| {
    println!("Tick");
    false // tell deno_core that it _may_ not poll again. `true` will prevent event loop from exiting
  })
  .build();
```


bartlomieju (approved):
LGTM, but if this PR impacts benchmark performance I think we will have to revert it and look for some other solution.

core/runtime.rs
```diff
@@ -788,6 +795,18 @@ impl JsRuntime {
       self.check_promise_exceptions()?;
     }
 
+    // Event loop middlewares
```
littledivy, line 798: Updated where middlewares are called. For NAPI, we need it after checking for promise exceptions.

https://github.com/denoland/deno/pull/13633/files#diff-5b779288098bb6449725245ce7fdb1ec3190563b5dd255e747cfc35e639407aeR854

bartlomieju (approved):
LGTM

PR #13837, opened 2022-03-04
chore: automatically update Releases.md when doing a release

dsherret:


bartlomieju:
I don't have a strong opinion about this one, bot for minor releases the list should be filtered for lines that have issue number that is present in previous patch releases, otherwise we might end up with duplicates.

dsherret:
> I don't have a strong opinion about this one, but for minor releases the list should be filtered for lines that have issue number that is present in previous patch releases, otherwise we might end up with duplicates.

@bartlomieju that's ok. It tells them to inspect the output and make sure it's correct so for minor releases then can just delete those lines. I think it would be too much effort to inspect the previous releases in the markdown and filter it out any duplicates.

bartlomieju (approved):

PR #13841, opened 2022-03-04
perf(ext/web): optimize atob/btoa

AaronO:
Follow up to https://github.com/denoland/deno/pull/13839, optimizing `base64_roundtrip` ~20x (~125ms => ~6.5ms)

Before:
```
base64_roundtrip:    	n = 100, dt = 12.493s, r = 8/s, t = 124930000ns/op
```
After:
```
base64_roundtrip:    	n = 100, dt = 0.649s, r = 154/s, t = 6490000ns/op
```

### DIFF
--- a/ext/web/lib.rs
+++ b/ext/web/lib.rs
@@ -153,39 +153,37 @@ fn op_base64_decode(
   input: String,
   _: (),
 ) -> Result<ZeroCopyBuf, AnyError> {
-  Ok(b64_decode(input, false)?.into())
+  let mut input = input.into_bytes();
+  input.retain(|c| !c.is_ascii_whitespace());
+  Ok(b64_decode(&input)?.into())
 }
 
 fn op_base64_atob(
   _: &mut OpState,
-  input: String,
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):

ext/web/lib.rs
```diff
@@ -146,21 +149,41 @@ pub fn init<P: TimersPermission + 'static>(
...
+fn op_base64_atob(
+  _: &mut OpState,
+  input: String,
+  _: (),
+) -> Result<ByteString, AnyError> {
+  Ok(ByteString(b64_decode(input, true)?))
+}
+
+fn b64_decode(input: String, padding: bool) -> Result<Vec<u8>, AnyError> {
+  let mut input = input.into_bytes();
```
lucacasonato, line 168: How about using a byte string as input? Have you checked if that would be faster?

AaronO: Yeah was thinking that too, just added it and it's faster bringing us down from ~7ms to ~6ms in that bench

PR #13842, opened 2022-03-05
test: remove debug log from pattern matcher

bartlomieju:

### DIFF
--- a/test_util/src/lib.rs
+++ b/test_util/src/lib.rs
@@ -1835,21 +1835,16 @@ pub fn pattern_match(pattern: &str, s: &str, wildcard: &str) -> bool {
     if i == 0 {
       continue;
     }
-    dbg!(part, i);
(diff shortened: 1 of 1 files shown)


test_util/src/lib.rs
```diff
@@ -1835,21 +1835,16 @@ pub fn pattern_match(pattern: &str, s: &str, wildcard: &str) -> bool {
     if i == 0 {
       continue;
     }
-    dbg!(part, i);
```
dsherret, line 1838: I've found this helpful though. Why remove it?

dsherret (approved):

test_util/src/lib.rs
bartlomieju: I found it overly noisy if integration tests are failing. I'm fine with keeping it if you prefer.

PR #13844, opened 2022-03-05
docs(README): update urls of `shell` and `powershell` deno installers

abdfnx:


bartlomieju (approved):
LGTM, thanks @abdfnx!

PR #13845, opened 2022-03-05
Move base64 encoding/decoding to Rust side

### LINKED ISSUE #13838: atob and btoa are slower than node
https://twitter.com/jarredsumner/status/1499715417773789185

aapoalas:
Attempt to at least somewhat fix #13838.

### Before:
```
base64_roundtrip:       n = 10, dt = 4.157s, r = 2/s, t = 415700000ns/op
```

## After:
```
base64_roundtrip:       n = 10, dt = 1.936s, r = 5/s, t = 193600000ns/op
```

### DIFF
--- a/ext/web/internal.d.ts
+++ b/ext/web/internal.d.ts
@@ -43,8 +43,8 @@ declare namespace globalThis {
         result: string;
         position: number;
       };
-      forgivingBase64Encode(data: Uint8Array): string;
-      forgivingBase64Decode(data: string): Uint8Array;
+      forgivingBase64Encode(data: string): string;
+      forgivingBase64Decode(data: string): string;
     };
 
     declare var domException: {
--- a/ext/web/lib.rs
+++ b/ext/web/lib.rs
@@ -149,8 +149,25 @@ fn op_base64_decode(
   _state: &mut OpState,
   input: String,
   _: (),
-) -> Result<ZeroCopyBuf, AnyError> {
-  let mut input: &str = &input.replace(|c| char::is_ascii_whitespace(&c), "");
+) -> Result<String, AnyError> {
+  let mut ws = false;
+  let len = input.len();
+  let invalid = input.char_indices().any(|(i, c)| {
+    if char::is_ascii_whitespace(&c) {
+      ws = true;
+    } else if c == '=' {
+      if i != len - 1 && i != len - 2 {
+        return true;
+      }
+    } else if c != '+' && c != '/' && !c.is_alphanumeric() {
+      return true;
+    }
+    false
+  });
+
+  // "Remove all ASCII whitespace from data"
+  let input = if ws { input.replace(|c| char::is_ascii_whitespace(&c), "") } else { input };
+  let mut input: &str = &input;
   // "If the length of input divides by 4 leaving no remainder, then:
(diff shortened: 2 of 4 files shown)


ext/web/internal.d.ts
```diff
@@ -43,8 +43,8 @@ declare namespace globalThis {
         result: string;
         position: number;
       };
-      forgivingBase64Encode(data: Uint8Array): string;
-      forgivingBase64Decode(data: string): Uint8Array;
+      forgivingBase64Encode(data: string): string;
```
aapoalas, line 46: It's a bit of a shame that the strings (presumably) have to be copied. I wonder if it would be possible to have an optional `&str` serde?

ext/web/lib.rs
```diff
@@ -149,8 +149,25 @@ fn op_base64_decode(
...
-) -> Result<ZeroCopyBuf, AnyError> {
-  let mut input: &str = &input.replace(|c| char::is_ascii_whitespace(&c), "");
+) -> Result<String, AnyError> {
+  let mut ws = false;
+  let len = input.len();
+  let invalid = input.char_indices().any(|(i, c)| {
+    if char::is_ascii_whitespace(&c) {
+      ws = true;
+    } else if c == '=' {
+      if i != len - 1 && i != len - 2 {
```
aapoalas, line 159: There is a bit of a possibility for false negative here: Even if there are '=' characters at the end it does not mean that they'll be removed below since that depends on whitespace removal etc.

As such, it would be more appropriate to track '=' markers at end with an enum: None, One, Two. Then compare to those when '=' is being removed and reset the enum appropriately. If reset does not happen, it means that the '=' at the end were invalid characters.

ext/web/lib.rs
```diff
@@ -149,8 +149,25 @@ fn op_base64_decode(
...
+        return true;
+      }
+    } else if c != '+' && c != '/' && !c.is_alphanumeric() {
+      return true;
+    }
+    false
+  });
+
+  // "Remove all ASCII whitespace from data"
+  let input = if ws { input.replace(|c| char::is_ascii_whitespace(&c), "") } else { input };
```
aapoalas, line 169: It would be preferable if the replacement could be done on the same iter as the invalid characters but that's probably not possible.

ext/web/lib.rs
```diff
@@ -190,14 +205,21 @@ fn op_base64_decode(
...
 
 fn op_base64_encode(
   _state: &mut OpState,
-  s: ZeroCopyBuf,
+  s: String,
   _: (),
 ) -> Result<String, AnyError> {
+  let char_count = s.chars().count();
+  let s = s.into_bytes();
+  if s.len() != char_count {
```
aapoalas, line 218: Properly this should be `s.len() > char_count` but it doesn't really matter.

ext/web/lib.rs
```diff
@@ -190,14 +205,21 @@ fn op_base64_decode(
       err
     ))
   })?;
-  Ok(ZeroCopyBuf::from(out))
+  Ok(String::from_utf8(out).unwrap())
```
aapoalas, line 208: Not sure if `.unwrap()` here is valid.

aapoalas:
Ooops, a better perf optimization had already been posted.

Issue #13516, opened 2022-01-27
deno upgrade to 1.18.1 fails on Ubuntu 16.04 with "GLIBC_2.27 not found"


thomasweitzel:
This happens on Ubuntu Linux 16.04, 4.4.0-218-generic #251-Ubuntu SMP Thu Dec 2 09:48:54 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux:

```
$ RUST_BACKTRACE=1 deno upgrade
Looking up latest version
Found latest version 1.18.1
Checking https://github.com/denoland/deno/releases/download/v1.18.1/deno-x86_64-unknown-linux-gnu.zip
31.4 MiB / 31.4 MiB (100.0%)
Deno is upgrading to version 1.18.1
Archive:  /tmp/.tmpjvCQBk/deno.zip
  inflating: deno                    
/tmp/.tmpjvCQBk/deno: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by /tmp/.tmpjvCQBk/deno)

============================================================
Deno has panicked. This is a bug in Deno. Please report this
at https://github.com/denoland/deno/issues/new.
If you can reliably reproduce this panic, include the
reproduction steps and re-run with the RUST_BACKTRACE=1 env
var set and include the backtrace in your report.

Platform: linux x86_64
Version: 1.18.0
Args: ["deno", "upgrade"]

thread 'main' panicked at 'assertion failed: output.status.success()', cli/tools/upgrade.rs:306:3
stack backtrace:
   0: rust_begin_unwind
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:48:5
   3: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   5: deno::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Only the older  GLIBC 2.23 exits:

```
$ ldd --version
ldd (Ubuntu GLIBC 2.23-0ubuntu11.3) 2.23
```

[...]

PR #13846, opened 2022-03-06
fix(ci): restore compatibility with old glibc

piscisaureus:
Fixes: #13516


littledivy (approved):
Works on AMI Linux 2

LGTM (once cleaned up and the tests passes).

bartlomieju (approved):
LGTM

PR #13847, opened 2022-03-06
fix(info): print output with unescaped backslashes on windows

### LINKED ISSUE #13076: `deno info` outputs path with escaped backslashes on Windows
```
> deno info
DENO_DIR location: "C:\\Users\\username\\AppData\\Local\\deno"
Remote modules cache: "C:\\Users\\username\\AppData\\Local\\deno\\deps"
Emitted modules cache: "C:\\Users\\username\\AppData\\Local\\deno\\gen"
Language server registries cache: "C:\\Users\\username\\AppData\\Local\\deno\\registries"
Origin storage: "C:\\Users\\username\\AppData\\Local\\deno\\location_data"
```

This is not so nice because if you copy and paste this, then you have to unescape it.

deno 1.16.4

For some reason we are using the debug format string:

https://github.com/denoland/deno/blob/a54fc7a1292b7da3ba1ccc1e6d6ab7ab5c750a36/cli/main.rs#L354

GJZwiers:
Should fix #13076 so stuff this like:

```
DENO_DIR location: "C:\\Users\\GJZwiers\\AppData\\Local\\deno"
```

becomes:

```
DENO_DIR location: C:\Users\GJZwiers\AppData\Local\deno
```

But I'm not sure how to write a test, because it's using `[WILDCARD]`

### DIFF
--- a/cli/main.rs
+++ b/cli/main.rs
@@ -328,6 +328,8 @@ fn print_cache_info(
       origin_dir.join(&checksum::gen(&[location.to_string().as_bytes()]));
   }
 
+  let local_storage_dir = origin_dir.join("local_storage");
+
   if json {
     let mut output = json!({
       "denoDir": deno_dir,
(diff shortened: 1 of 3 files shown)


bartlomieju:
@GJZwiers this PR failed the tests just like https://github.com/denoland/deno/pull/13785 due to updated `test_util/std/` module. I reverted that change.

bartlomieju (commented):
Looks good to me, but if I remember correctly @dsherret had some thoughts on using `.display()` for paths. David can you take a look?

dsherret (approved):
LGTM. Thanks @GJZwiers!

PR #13808, opened 2022-03-01
feat(ext/net): Use Socket2 crate to create TcpListener

Trolloldem:
After a first attempt in solving #1195 in #13794, it has been noticed that the **SO_REUSEADDR** option has to be defaulted to true in order to pass the `net` unit tests. In the same PR was proposed to do the porting to Socket2 and the support to the options **SO_REUSEADDR** and **SO_REUSEPORT** in two separate PR. This PR presents the TcpListener implementation based on Socket2.

It is possible to notice that:
1. the **SO_REUSEADDR** option is defaulted to true only for non-windows systems, following the implementation of `TcpListener.bind()` of [std::net](https://github.com/rust-lang/rust/blob/6343edfa11c80ffed02de7e059178d4c1637ecdb/library/std/src/sys_common/net.rs)
2. similarly, the socket type is set to `SOCK_STREAM` and the `backlog` argument of `listen()` is set to 128

These steps should produce a TcpListener equal to the one produce by the instruction `std::net::TcpListener::bind(&addr)?;`

PR #13849, opened 2022-03-06
feat(ext/net): added SO_REUSEADDR option for TCP sockets

Trolloldem:
After a first attempt in solving #1195 in #13794, it has been noticed that the **SO_REUSEADDR** option has to be defaulted to true in order to pass the net unit tests. In the same PR was proposed to do the porting to Socket2 and the support to the options **SO_REUSEADDR** and **SO_REUSEPORT** in separate PR. #13808 successfully implemented TCP sockets through the Socket2 crate. This PR is a follow up which contains a new optional argument for `Deno.listen` and `Deno.listenTls` named `reuseAddress`. This optional argument can be used to configure the **SO_REUSEADDR** option. In addition to that, test for Unix system have been implemented. It should be noted that the behavior of Sockets for this option is different in Windows:

- for Unix, the **SO_REUSEADDR** has been defaulted to `true`  in #13808, in order to remain compliant to the `std::net::TcpListener` alternative previously used. This allows binding sockets to ports already occupied if the previous socket is in the **TIME_WAIT** state 
- for Windows, the behavior described in the previous point is the default one, thus  **SO_REUSEADDR** has been defaulted to `false` in #13808, similarly to how it is done in `std::net::TcpListener` (this is tested through `netTcpReuseAddr` and  `listenTlsWithReuseAddr`)
- manually setting **SO_REUSEADDR** in Windows allows binding sockets to ports already occupied, even if the previous socket is still active (test for this behavior are currently not present in the PR)
- a similar behavior can be achieved in Unix system thanks to the **SO_REUSEPORT** option, as mentioned in #13794


Trolloldem:
I think there is something wrong with this PR. The CI keep failing even after several resets, but when I run the same CI after a merge of the branch on my _main_ branch, which starts from a clean state, they are successful. Am I doing something wrong in the process?

Trolloldem:
Thank you @bartlomieju, it seems fine now. Is it better to also add test for the described behavior in Windows?

Trolloldem:
Thank you @bartlomieju, it seems fine now. Is it better to also add test for the described behavior in Windows?

bartlomieju:
> Thank you @bartlomieju, it seems fine now. Is it better to also add test for the described behavior in Windows?

That would be great! I'll try to get this landed for 1.20

Trolloldem:
> > Thank you @bartlomieju, it seems fine now. Is it better to also add test for the described behavior in Windows?
> 
> That would be great! I'll try to get this landed for 1.20

Test for Windows behavior added and CI seem fine. I leave here a quick reference about the tests implemented:

- `netTcpReuseAddr` opens a socket using the default setting for **SO_REUSEADDR** of each OS, then the socket is closed, 
putting it in the **TIME_WAIT** state; a second socket is successfully opened
- `netTcpNoReuseAddr` performs the same steps of the previous test but **SO_REUSEADDR** is set to `false`, in order to test the behavior in OS different from Windows. In this case the test has to fail with an `AddrInUse` error. This test can fail in two point, proving the same property, the first point being the first listener creation (it can fail since sockets from previous tests that occupied the required port are in the **TIME_WAIT** state) while the second point is the creation of the second listener (since the first one is in the **TIME_WAIT** state)
- `netTcpTrueReuseAddrWindows` set **SO_REUSEADDR** to `true` and is run only in Windows. In this case two listeners are bound to the same port, which is usually forbidden for the default setting of **SO_REUSEADDR** equal to `false`

The tests `listenTlsWithReuseAddr`, `listenTlsWithNoReuseAddr` and `listenTlsWithTrueReuseAddrWindows` follow the same 
logic but are for `listenTls`

ext/net/lib.deno_net.d.ts
```diff
@@ -76,6 +76,9 @@ declare namespace Deno {
      * You should show the message like `server running on localhost:8080` instead of
      * `server running on 0.0.0.0:8080` if your program supports Windows. */
     hostname?: string;
+    reuseAddress?: boolean;
+    /** SO_REUSEADDR option for socket.
+     * If not specified, defaults to true in Linux, false in Windows. */
```
bartlomieju, line 81: This docstring should be above `reuseAddress` field

ext/net/lib.deno_net.d.ts
```diff
@@ -110,6 +113,9 @@ declare namespace Deno {
     keyFile?: string;
 
     transport?: "tcp";
+    reuseAddress?: boolean;
+    /** SO_REUSEADDR option for socket.
+     * If not specified, defaults to false. */
```
bartlomieju, line 118: Ditto

bartlomieju (commented):
@Trolloldem sorry for the delay, I've got two nitpicks and after that we can land the PR.

ext/net/lib.deno_net.d.ts
Trolloldem: Docstring moved in the correct position

ext/net/lib.deno_net.d.ts
Trolloldem: Same as the previous one

Trolloldem:
> @Trolloldem sorry for the delay, I've got two nitpicks and after that we can land the PR.

No problem to me. I've noticed there was a lot of things moving with 1.20, so I took the chance to rearrange the commit history of the PR into a single commit + one commit to change the cache key.

I've addressed the requests concerning the docstrings and merged my branch with the latest version of the main branch of Deno

ext/net/lib.deno_net.d.ts
```diff
@@ -110,6 +113,9 @@ declare namespace Deno {
     keyFile?: string;
 
     transport?: "tcp";
+    /** SO_REUSEADDR option for socket.
+     * If not specified, defaults to true in Linux, false in Windows. */
+    reuseAddress?: boolean;
```
ry, line 118: This doesn't need to be added to ListenTlsOptions because it gets inherited from ListenOptions.

ext/net/lib.deno_net.d.ts
```diff
@@ -76,6 +76,9 @@ declare namespace Deno {
      * You should show the message like `server running on localhost:8080` instead of
      * `server running on 0.0.0.0:8080` if your program supports Windows. */
     hostname?: string;
+    /** SO_REUSEADDR option for socket.
+     * If not specified, defaults to true in Linux, false in Windows. */
```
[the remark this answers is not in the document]
Trolloldem: **SO_REUSEADDR** defaults to true because, as stated in #13808, the previous implementation of TCP socket used the [std::net](https://github.com/rust-lang/rust/blob/6343edfa11c80ffed02de7e059178d4c1637ecdb/library/std/src/sys_common/net.rs) implementation, which sets **SO_REUSEADDR** to true when binding a **TcpListener** outside of Windows. In this way it is possible to keep the behavior of existing code that does not explicit use the new **reuseAddress** argument. Setting this to **false** as default value could break existing code due to the fact that in Unix system is not possible to bind socket to a combination of address+port occupied by closed sockets in the **TIME_WAIT** state, as visible in failed tests in #13794. In addition, this seem the same behavior of node: [unix tcp](https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/unix/tcp.c), [win tcp](https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/win/tcp.c).

ext/net/lib.deno_net.d.ts
Trolloldem: Unnecessary field and docstrings removed

Trolloldem:
The changes introduced up to now concern only TCP sockets. With the last two commits I have extended the possibility of using the **SO_REUSEADDR** even for UDP sockets. Since they have a different behavior compared to TCP sockets, I have also updated the docstring of the **reuseAddress** optional argument. This argument now can be specified through `Deno.listenDatagram`. In addition to that, 2 new test have been added: `netUdpReuseAddr` and `netUdpNoReuseAddr`. 

To enable this option, the Socket2 crate has been used, using the same settings of the [std::net](https://github.com/rust-lang/rust/blob/6343edfa11c80ffed02de7e059178d4c1637ecdb/library/std/src/sys_common/net.rs) implementation of `UdpSocket`.

Similarly to node ([win udp](https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/win/udp.c), [unix udp](https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/unix/udp.c)), the options always defaults to `false`, and for both Windows and Linux setting **SO_REUSEADDR** actually uses that option under the hood. For the other *nix systems **SO_REUSEPORT** is actually used.

Trolloldem:
> The changes introduced up to now concern only TCP sockets. With the last two commits I have extended the possibility of using the **SO_REUSEADDR** even for UDP sockets. Since they have a different behavior compared to TCP sockets, I have also updated the docstring of the **reuseAddress** optional argument. This argument now can be specified through `Deno.listenDatagram`. In addition to that, 2 new test have been added: `netUdpReuseAddr` and `netUdpNoReuseAddr`.
> 
> To enable this option, the Socket2 crate has been used, using the same settings of the [std::net](https://github.com/rust-lang/rust/blob/6343edfa11c80ffed02de7e059178d4c1637ecdb/library/std/src/sys_common/net.rs) implementation of `UdpSocket`.
> 
> Similarly to node ([win udp](https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/win/udp.c), [unix udp](https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/deps/uv/src/unix/udp.c)), the options always defaults to `false`, and for both Windows and Linux setting **SO_REUSEADDR** actually uses that option under the hood. For the other *nix systems **SO_REUSEPORT** is actually used.

@bartlomieju sorry for the ping. Should I change the title after these changes (since now also UDP should be covered). In addition, is this still a requested feature? Should I provide any additional documentation?

bartlomieju:
@littledivy @lucacasonato could you take a look at this PR?

lucacasonato:
I have rebased this PR.

I have removed the `reuseAddress` option for TCP sockets. For TCP, this option is unconfigurably `true` on all platforms but Windows. This behaviour matches Node.js and Go. Neither of these systems support disabling `SO_REUSEADDR` on socket listeners.

For UDP, `reuseAddress` is supported. It has the same port stealing behaviour on all systems we support, but to achieve this it uses different syscalls on different platforms. macOS and other BSDs use `SO_REUSEADDR`, while Linux and Windows use `SO_REUSEPORT`. This behaviour matches libuv.

As a follow up PR we can add a `reusePort` option to enable the kernel connection load  balancing behaviour found in Linux, for TCP sockets.

I think this PR is now good to go.

Trolloldem:
> I have rebased this PR.
> 
> I have removed the `reuseAddress` option for TCP sockets. For TCP, this option is unconfigurably `true` on all platforms but Windows. This behaviour matches Node.js and Go. Neither of these systems support disabling `SO_REUSEADDR` on socket listeners.
> 
> For UDP, `reuseAddress` is supported. It has the same port stealing behaviour on all systems we support, but to achieve this it uses different syscalls on different platforms. macOS and other BSDs use `SO_REUSEADDR`, while Linux and Windows use `SO_REUSEPORT`. This behaviour matches libuv.
> 
> As a follow up PR we can add a `reusePort` option to enable the kernel connection load balancing behaviour found in Linux, for TCP sockets.
> 
> I think this PR is now good to go.

Thank you for reviewing my PR @lucacasonato! Just as a little remark, in your summary: `macOS and other BSDs use SO_REUSEADDR, while Linux and Windows use SO_REUSEPORT. This behaviour matches libuv`, it is actually the opposite (as pointed out in the comment referring `libuv` in `ops.rs`

lucacasonato (approved):
LGTM

littledivy (approved):
LGTM!

PR #13859, opened 2022-03-07
refactor: cleanup assert() & AssertionError definitions

crowlKats:


ext/web/00_infra.js
```diff
@@ -275,6 +276,24 @@
     return StringPrototypeMatch(s, HTTP_BETWEEN_WHITESPACE)?.[1] ?? "";
   }
 
+  class AssertionError extends Error {
+    constructor(msg) {
+      super(msg);
```
petamoriken, line 281: We should set the second argument for error cause.

```suggestion
    constructor(msg, opts = undefined) {
      super(msg, opts);
```

crowlKats: why so? we only construct `AssertionError` in the `assert` function

petamoriken: For now you are right, but it will be useful in the future when it is used from another function or when a class inheriting from the `Error` class is created with reference to this file.

bartlomieju: > For now you are right, but it will be useful in the future when it is used from another function or when a class inheriting from the Error class is created with reference to this file.

This is completely internal code and won't be exposed to users. I think we should keep it simple for now. If a need arises in the future we can change the interface.

bartlomieju (approved):

PR #13860, opened 2022-03-07
perf(serde_v8): avoid SerializablePkg allocs

AaronO:
For common return types such as String/ZeroCopyBuf/ByteString

This should impact many ops, but here's the delta on b64 from `deno_common.js`:

Before:
```
b64_rt_long:         	n = 100, dt = 0.656s, r = 152/s, t = 6560000ns/op
b64_rt_short:        	n = 1000000, dt = 0.524s, r = 1908397/s, t = 524ns/op
```

After:
```
b64_rt_long:         	n = 100, dt = 0.632s, r = 158/s, t = 6320000ns/op
b64_rt_short:        	n = 1000000, dt = 0.456s, r = 2192982/s, t = 456ns/op
```

### DIFF
--- a/serde_v8/src/serializable.rs
+++ b/serde_v8/src/serializable.rs
@@ -2,6 +2,9 @@
 use std::any::TypeId;
 use std::mem::transmute_copy;
 
+use crate::Buffer;
+use crate::ByteString;
+
 /// Serializable exists to allow boxing values as "objects" to be serialized later,
 /// this is particularly useful for async op-responses. This trait is a more efficient
 /// replacement for erased-serde that makes less allocations, since it's specific to serde_v8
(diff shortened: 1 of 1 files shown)


serde_v8/src/serializable.rs
```diff
@@ -86,7 +94,9 @@ impl<T: serde::Serialize + 'static> From<T> for SerializablePkg {
   fn from(x: T) -> Self {
     #[inline(always)]
     fn tc<T, U>(src: T) -> U {
-      unsafe { transmute_copy(&src) }
+      let x = unsafe { transmute_copy(&src) };
+      std::mem::forget(src);
```
littledivy, line 98: Why leak? Is this being freed somewhere?

AaronO: We don't leak it, `tc` is essentially a transmute that works with our TypeId dispatch.

`x` is a low-level copy of `src`, so in the case of strings we get two `String`s structs that actually point to the same buffer which violates single ownership principles so we forget the first without dropping it since otherwise that will lead to double-free and use-after-free problems.

littledivy (approved):
LGTM

PR #13861, opened 2022-03-07
feat(core): codegen ops

littledivy:
Code-generating ops via a new `#[op]` macro avoids the overhead of the polymorphic OpTable by instead generating monomorphic v8 functions optimized for each individual op. This also provides foundations required to implement v8 Fast API and ergonomic changes such as allowing ops to have their own arity.

Deno 1.19.3 (main 34c9e160992acf77f873691b3bbfb3e5e07c63d1):

`op_baseline`:
```
test bench_op_async   ... bench:     283,582 ns/iter (+/- 4,154)
test bench_op_nop     ... bench:      50,285 ns/iter (+/- 1,574)
test bench_op_pi_json ... bench:      53,111 ns/iter (+/- 5,828)
```

`deno_common`:
```
date_now:            	n = 500000, dt = 0.030s, r = 16666667/s, t = 59ns/op
op_void_sync:        	n = 10000000, dt = 0.504s, r = 19841270/s, t = 50ns/op
op_void_async:       	n = 1000000, dt = 0.268s, r = 3731343/s, t = 268ns/op
perf_now:            	n = 500000, dt = 0.058s, r = 8620690/s, t = 116ns/op
url_parse:           	n = 50000, dt = 0.070s, r = 714286/s, t = 1400ns/op
b64_rt_long:         	n = 100, dt = 0.674s, r = 148/s, t = 6740000ns/op
b64_rt_short:        	n = 1000000, dt = 0.401s, r = 2493766/s, t = 401ns/op
read_zero:           	n = 500000, dt = 0.244s, r = 2049180/s, t = 488ns/op
write_null:          	n = 500000, dt = 0.221s, r = 2262443/s, t = 442ns/op
read_128k:           	n = 50000, dt = 6.897s, r = 7250/s, t = 137940ns/op
request_new:         	n = 500000, dt = 0.647s, r = 772798/s, t = 1294ns/op
```

This PR:

`op_baseline`:
```
test bench_op_async   ... bench:     252,617 ns/iter (+/- 11,114)
test bench_op_nop     ... bench:      18,877 ns/iter (+/- 1,674)
test bench_op_pi_json ... bench:      23,310 ns/iter (+/- 2,004)
```

`deno_common`:
```
date_now:            	n = 500000, dt = 0.030s, r = 16666667/s, t = 59ns/op
op_void_sync:        	n = 10000000, dt = 0.286s, r = 34965035/s, t = 28ns/op
op_void_async:       	n = 1000000, dt = 0.250s, r = 4000000/s, t = 250ns/op
perf_now:            	n = 500000, dt = 0.055s, r = 9090909/s, t = 110ns/op
url_parse:           	n = 50000, dt = 0.068s, r = 735294/s, t = 1360ns/op
b64_rt_long:         	n = 100, dt = 0.345s, r = 290/s, t = 3450000ns/op
b64_rt_short:        	n = 1000000, dt = 0.334s, r = 2994012/s, t = 334ns/op
read_zero:           	n = 500000, dt = 0.233s, r = 2145923/s, t = 466ns/op
write_null:          	n = 500000, dt = 0.209s, r = 2392344/s, t = 417ns/op
read_128k:           	n = 50000, dt = 6.975s, r = 7168/s, t = 139500ns/op
request_new:         	n = 500000, dt = 0.626s, r = 798722/s, t = 1252ns/op
```

* Note: async ops have other "overpowering overheads" so there is not much difference there.

## Changes
- Gets rid of `deno_core::OpTable`. ops are now seperate v8 function callbacks instead of routing through `OpTable::route_op`
- Introduces a `#[deno_op::op]` and `#[deno_op::op_async]` macro to auto generate v8 callbacks for existing ops.
- Extensions are now the only way to register new ops. 
- Changes `Extensions::ops` method. 
Why? Because `rusty_v8` wants fn items when creating a function callback. Storing ops will store the function pointer instead :-) `RegisterCtx` solves this:
```rust
Extension::builder()
  .ops(vec![
    op_nop::decl()
  ])

#[op]
fn op_nop(_: &mut OpState, _: (), _: ()) -> Result<(), AnyError> {
  Ok(())
}
```
- Introduce `RegisterCtx` enum, ops can be registered as a) a v8 function callback and b) an external reference for the callback before snapshotting. 
- Macro optimizations. Skip serialization for `a`/`b` args if the expected argument is a Wild type (`_`). Also skips `Ok()` deserialization for `Result<(), E>`
- Uses faster `v8::FunctionBuilder::data` to pass `OpState` instead of `JsRuntime::state` (rusty_v8 slots).

## Todo

- [x] Move metrics back to Rust. (idk why I moved it to JS)
- [x] Is `Deno.core.ops.op_nop` faster than `Deno.core.opSync("op_nop")`? (EDIT: nope)
- [x] Merge `#[op_async]` into `#[op]`
- [x] Op middlewares


core/runtime.rs
```diff
@@ -277,17 +275,23 @@ impl JsRuntime {
 
     let has_startup_snapshot = options.startup_snapshot.is_some();
 
+    // Add builtins extension
+    options
+      .extensions
+      .insert(0, crate::ops_builtin::init_builtins());
+    let refs = bindings::external_references(&mut options.extensions);
+    let refs: &'static v8::ExternalReferences = Box::leak(Box::new(refs));
```
littledivy, line 283: https://github.com/denoland/rusty_v8/issues/912

ops/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+      },
+      _ => quote! {
+        let b = args.get(3);
+        let b = deno_core::serde_v8::from_v8(scope, b).unwrap();
+      },
+    },
+    _ => unreachable!(),
+  };
+
+  // TODO(@littledivy): Optimize Result<(), Err> to skip serde_v8.
```
aapoalas, line 40: I presume this should not be hard at all?

aapoalas:
Would it be possible (in the future at least) to generate the ops bindings using `v8-fast-api-call`? Or am I mistaken in thinking that it might be possible?

ops/lib.rs
littledivy: yup, just a lot of nasty `match`-es :-)

littledivy:
> Would it be possible (in the future at least) to generate the ops bindings using `v8-fast-api-call`? Or am I mistaken in thinking that it might be possible?

Yes, this is (also) preparing a base for that with the proc macros.

aapoalas:
> Yes, this is (also) preparing a base for that with the proc macros.

This sounds absolutely amazing.

ops/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+      },
+      _ => quote! {
+        let a = args.get(2);
+        let a = deno_core::serde_v8::from_v8(scope, a).unwrap();
+      },
+    },
+    _ => unreachable!(),
+  };
+
+  let b = match &inputs[2] {
```
inteon, line 27: Is it possible here to allow for ops with 1 or 2 function arguments instead of 3?

inteon: just to beautify the syntax a bit (dropping the now redundant _: () arguments)

inteon: Also, some ops don't even require an op_state; not sure if rust would be smart enough to not borrow state_rc, so allowing an ops signature and not having no such argument might even improve performance.

littledivy: > Is it possible here to allow for ops with 1 or 2 function arguments instead of 3?

The first argument is the OpId used for tracking purposes

littledivy: > Also, some ops don't even require an op_state; not sure if rust would be smart enough to not borrow state_rc, so allowing an ops signature and not having no such argument might even improve performance.

Ops might not but metrics tracking does require borrowing state.

inteon: > > Is it possible here to allow for ops with 1 or 2 function arguments instead of 3?
> 
> The first argument is the OpId used for tracking purposes

Not sure if I was clear enough here, I meant to say that it is now possible to also allow ops definitions with a function signature that has less than 3 arguments:

eg. https://github.com/denoland/deno/pull/13861/files#diff-fa76844bfec3519648a2d7493933ff3317fa3250dd913cc26d0e86b3b5d4465aR184
```rust
#[op]
fn op_load(
    state: &mut OpState,
    args: LoadArgs,
    ...
}
```
instead of:
```rust
#[op]
fn op_load(
    state: &mut OpState,
    args: LoadArgs,
    _: (),
    ...
}
```

inteon: thanks for explaining, cool PR!

ops/lib.rs
littledivy: done, it was definitely worth it.

aapoalas (commented):
Left a few comments, hopefully some of them somewhat useful.

Not sure if it would be much helpful performance wise or even possible Rust-wise but it would be quite nice ops could indeed somehow "opt out of" receiving the 0th state parameter. though op tracking would still need to touch the state.

core/ops_builtin.rs
```diff
@@ -24,29 +21,44 @@ pub(crate) fn init_builtins() -> Extension {
...
+  pub use crate::*;
+}
+
+#[op]
+pub fn void_op_sync(_: &mut OpState, _: (), _: ()) -> Result<(), Error> {
+  Ok(())
+}
+
+pub async fn void_op_async(
+  _state: Rc<RefCell<OpState>>,
```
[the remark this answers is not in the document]
crowlKats: common thing across the codebase, only one single place where just `_` is used for it. just gives more context; just nicer to read

ops/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+
+          let state_rc = JsRuntime::state(scope);
+          let mut state = state_rc.borrow_mut();
+          state.op_state.borrow().tracker.track_async(op_id);
+          state.have_unpolled_ops = true;
+          let op_state = state.op_state.clone();
+          let op_state2 = op_state.clone();
+          state.pending_ops.push(OpCall::eager(async move {
+            let result = #name::<#type_params>(op_state.clone(), a, b).await;
+            (promise_id, op_id, to_op_result(&op_state.borrow(), result))
```
aapoalas, line 191: Hmm, would it make sense to, again with the same `()` return value condition, use a specific `to_null_op_result` function that could then (presumably) avoid serializing the Promise resolve value?

core/ops_builtin.rs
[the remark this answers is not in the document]
aapoalas: At some level I would've presumed that Rust might do more code elimination with `_` vs `_state` but on the other hand, probably not since at the end of the day the compiler will know exactly and equally well if `_state` is unused or not.

AaronO:
> Left a few comments, hopefully some of them somewhat useful.
> 
> Not sure if it would be much helpful performance wise or even possible Rust-wise but it would be quite nice ops could indeed somehow "opt out of" receiving the 0th state parameter. though op tracking would still need to touch the state.

Thanks for the comments! We were mostly focused on making it work, @littledivy and I will do a few cleanup passes before landing this now that we're unblocked on the `rusty_v8` side (via https://github.com/denoland/rusty_v8/pull/917).

There will be follow-up PRs with additional work, this PR is mainly focused on laying the foundations of codegen'd ops.

ops/lib.rs
```diff
@@ -154,24 +155,21 @@ pub fn op(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
         ) #where_clause {
-          use deno_core::JsRuntime;
           use deno_core::futures::FutureExt;
-          use deno_core::OpCall;
-          use deno_core::to_op_result;
-          use deno_core::PromiseId;
-          use deno_core::bindings::throw_type_error;
-          use deno_core::v8;
-          let op_id = unsafe { v8::Local::<v8::Integer>::cast(args.get(0)) }.value() as usize;
+          // SAFETY: Called from Deno.core.opSync. Which retrieves the index using opId table.
```
aapoalas, line 159: Technically called from `Deno.core.opAsync` I guess? :)

core/bindings.rs
```diff
@@ -162,17 +162,36 @@ pub fn initialize_context<'s>(
 
   let scope = &mut v8::ContextScope::new(scope, context);
 
+  let deno_key = v8::String::new(scope, "Deno").unwrap();
+  let core_key = v8::String::new(scope, "core").unwrap();
+  let ops_key = v8::String::new(scope, "ops").unwrap();
+  if snapshot_loaded {
+    let deno_val = global.get(scope, deno_key.into()).unwrap();
+    let deno_val = v8::Local::<v8::Object>::try_from(deno_val)
+      .unwrap_or_else(|_| v8::Object::new(scope));
```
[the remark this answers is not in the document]
littledivy: Nice catch, yeah unwrapping is probably better here since the snapshot must have `Deno.core.ops` anyways.

core/bindings.rs
```diff
@@ -227,19 +246,38 @@ pub fn initialize_context<'s>(
...
+}
+
+pub fn set_func_raw(
+  scope: &mut v8::HandleScope<'_>,
+  obj: v8::Local<v8::Object>,
+  name: &'static str,
+  callback: v8::FunctionCallback,
+  external_data: Rc<RefCell<OpState>>,
+) {
+  let key = v8::String::new(scope, name).unwrap();
```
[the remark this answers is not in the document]
littledivy: nope, `val` is `Local<Function>` not a `fn(&mut HandleScope, blabla)`

ops/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+          #a
+          #b
+          #func
+
+          let state_rc = JsRuntime::state(scope);
+          let mut state = state_rc.borrow_mut();
+          state.op_state.borrow().tracker.track_async(op_id);
+          state.have_unpolled_ops = true;
+          let op_state = state.op_state.clone();
+          let op_state2 = op_state.clone();
```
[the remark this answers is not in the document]
littledivy: > What is the purpose of this second clone of the original clone which never gets used?

Removed.

core/01_core.js
```diff
@@ -179,8 +170,8 @@
     return p;
   }
 
-  function opSync(opName, arg1 = null, arg2 = null) {
-    return unwrapOpResult(opcallSync(opsCache[opName], arg1, arg2));
+  function opSync(opName, arg1, arg2) {
```
bartlomieju, line 173: Just to make sure, both `null` and `undefined` are serialized to the same value by `serde_v8`, right?

core/Cargo.toml
```diff
@@ -14,6 +14,7 @@ path = "lib.rs"
 
 [dependencies]
 anyhow = "1.0.55"
+deno_ops = { path = "../ops" }
```
bartlomieju, line 17: missing `version` field

core/bindings.rs
```diff
@@ -162,17 +163,43 @@ pub fn initialize_context<'s>(
 
   let scope = &mut v8::ContextScope::new(scope, context);
 
-  // global.Deno = { core: {} };
   let deno_key = v8::String::new(scope, "Deno").unwrap();
+  let core_key = v8::String::new(scope, "core").unwrap();
+  let ops_key = v8::String::new(scope, "ops").unwrap();
+  // Snapshot already registered `Deno.core.ops` but
+  // extensions may provide ops that aren't part of the snapshot.
```
bartlomieju, line 170: This is a red flag for me. What if you want to create some ops during snapshotting but don't want for them to available after the snapshot is loaded? AFAIK this can happen in TSC snapshot which needs some ops to create a snapshot, but they should be exposed during runtime. 

Please explain this bit

core/bindings.rs
```diff
@@ -1471,7 +1389,7 @@ fn is_proxy(
   rv.set(v8::Boolean::new(scope, args.get(0).is_proxy()).into())
 }
 
-fn throw_type_error(scope: &mut v8::HandleScope, message: impl AsRef<str>) {
+pub fn throw_type_error(scope: &mut v8::HandleScope, message: impl AsRef<str>) {
```
bartlomieju, line 1392: This is code smell - it introduces circular dependency between `deno_core/` and `ops/`

core/error_codes.rs
```diff
@@ -1,6 +1,6 @@
 use anyhow::Error;
 
-pub(crate) fn get_error_code(err: &Error) -> Option<&'static str> {
+pub fn get_error_code(err: &Error) -> Option<&'static str> {
```
bartlomieju, line 3: This is also introducing circular dependency between `deno_core/` and `ops/`

core/examples/disable_ops.rs
```diff
@@ -9,7 +9,7 @@ use deno_core::RuntimeOptions;
 fn main() {
   let my_ext = Extension::builder()
     .middleware(|name, opfn| match name {
-      "op_print" => deno_core::void_op_sync(),
+      "op_print" => deno_core::void_op_sync::v8_cb(),
```
bartlomieju, line 12: What's the difference between `v8_cb` and `decl`?

core/examples/eval_js_value.rs
```diff
@@ -10,6 +10,11 @@ use deno_core::v8;
 use deno_core::JsRuntime;
 use deno_core::RuntimeOptions;
 
+// This is a hack to make the `#[op]` macro work with
+// deno_core examples.
+// You can remove this:
+use deno_core::*;
```
bartlomieju, line 16: Is this hack required or not? If not it should be removed, if it is required then it's another red flag for me. We shouldn't require `deno_core::*` import

core/examples/hello_world.rs
```diff
@@ -2,27 +2,37 @@
...
+use deno_core::op;
 use deno_core::Extension;
 use deno_core::JsRuntime;
+use deno_core::OpState;
 use deno_core::RuntimeOptions;
 
+// This is a hack to make the `#[op]` macro work with
+// deno_core examples.
+// You can remove this:
+use deno_core::*;
```
bartlomieju, line 14: Ditto

core/examples/http_bench_json_ops.rs
```diff
@@ -17,6 +18,11 @@ use std::rc::Rc;
 use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
 
+// This is a hack to make the `#[op]` macro work with
+// deno_core examples.
+// You can remove this:
+use deno_core::*;
```
bartlomieju, line 24: Ditto

core/examples/schedule_task.rs
```diff
@@ -9,14 +10,16 @@ use futures::channel::mpsc;
 use futures::stream::StreamExt;
 use std::task::Poll;
 
+// This is a hack to make the `#[op]` macro work with
+// deno_core examples.
+// You can remove this:
+use deno_core::*;
```
bartlomieju, line 16: Ditto

core/examples/ts_module_loader.rs
```diff
@@ -7,6 +7,11 @@
 use std::pin::Pin;
 use std::rc::Rc;
 
+// This is a hack to make the `#[op]` macro work with
+// deno_core examples.
+// You can remove this:
+use deno_core::*;
```
bartlomieju, line 13: Ditto

core/extensions.rs
```diff
@@ -1,12 +1,13 @@
-use crate::OpFn;
+// use crate::OpFn;
```
bartlomieju, line 1: Remove, and there's missing copyright here

core/lib.rs
```diff
@@ -1,9 +1,9 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 mod async_cancel;
 mod async_cell;
-mod bindings;
+pub mod bindings;
 pub mod error;
-mod error_codes;
+pub mod error_codes;
```
bartlomieju, line 6: These two are red flags for me; they shouldn't be public modules. What's the reason?

core/lib.rs
```diff
@@ -66,23 +69,23 @@ pub use crate::modules::ModuleType;
...
 pub use crate::ops_builtin::op_print;
 pub use crate::ops_builtin::op_resources;
-pub use crate::ops_json::op_async;
-pub use crate::ops_json::op_sync;
-pub use crate::ops_json::void_op_async;
-pub use crate::ops_json::void_op_sync;
+pub use crate::ops_builtin::void_op_async;
+pub use crate::ops_builtin::void_op_sync;
+pub use crate::ops_metrics::OpMetrics;
+pub use crate::ops_metrics::OpsTracker;
```
bartlomieju, line 88: Why are these two public?

core/ops.rs
```diff
@@ -136,44 +108,61 @@ impl OpResult {
...
+  }
+}
+
 /// Maintains the resources and ops inside a JS runtime.
 pub struct OpState {
   pub resource_table: ResourceTable,
-  pub op_table: OpTable,
   pub get_error_class_fn: GetErrorClassFn,
-  pub(crate) tracker: OpsTracker,
+  pub tracker: OpsTracker,
```
bartlomieju, line 153: Why is `tracker` now public?

core/runtime.rs
```diff
@@ -140,10 +140,9 @@ pub type CompiledWasmModuleStore = CrossIsolateStore<v8::CompiledWasmModule>;
 
 /// Internal state for JsRuntime which is stored in one of v8::Isolate's
 /// embedder slots.
-pub(crate) struct JsRuntimeState {
+pub struct JsRuntimeState {
```
bartlomieju, line 143: Why does it need to be public?

core/runtime.rs
```diff
@@ -158,10 +157,10 @@ pub(crate) struct JsRuntimeState {
   /// of the event loop.
   dyn_module_evaluate_idle_counter: u32,
   pub(crate) js_error_create_fn: Rc<JsErrorCreateFn>,
-  pub(crate) pending_ops: FuturesUnordered<PendingOpFuture>,
+  pub pending_ops: FuturesUnordered<PendingOpFuture>,
   pub(crate) unrefed_ops: HashSet<i32>,
-  pub(crate) have_unpolled_ops: bool,
-  pub(crate) op_state: Rc<RefCell<OpState>>,
+  pub have_unpolled_ops: bool,
+  pub op_state: Rc<RefCell<OpState>>,
```
bartlomieju, line 163: Why are these fields now public?

core/runtime.rs
```diff
@@ -461,34 +462,44 @@ impl JsRuntime {
...
   }
 
-  /// Initializes ops of provided Extensions
-  fn init_extension_ops(&mut self) -> Result<(), Error> {
-    let op_state = self.op_state();
-    // Take extensions to avoid double-borrow
-    let mut extensions: Vec<Extension> = std::mem::take(&mut self.extensions);
-
+  /// Applies middleware to ops
+  fn effective_ops(extensions: &mut [Extension]) -> Vec<OpPair> {
```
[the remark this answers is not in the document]
bartlomieju: I agree

core/runtime.rs
```diff
@@ -2027,29 +2006,30 @@ pub mod tests {
...
+  //   run_in_task(|cx| {
+  //     let (mut runtime, _dispatch_count) = setup(Mode::Async);
+  //     runtime
+  //       .execute_script(
+  //         "bad_op_id.js",
+  //         r#"
+  //         let thrown;
+  //         try {
+  //           Deno.core.opcallSync(100, null, null);
+  //         } catch (e) {
```
bartlomieju, line 2021: Remove before landing since this API is no longer available.

ext/broadcast_channel/lib.rs
```diff
@@ -8,8 +8,8 @@ pub use in_memory_broadcast_channel::InMemoryBroadcastChannelResource;
 use async_trait::async_trait;
 use deno_core::error::AnyError;
 use deno_core::include_js_files;
-use deno_core::op_async;
-use deno_core::op_sync;
+use deno_core::op;
+
```
bartlomieju, line 12: Nitpick: remove this empty line

ext/crypto/decrypt.rs
```diff
@@ -24,6 +24,8 @@ use ctr::Ctr;
 use deno_core::error::custom_error;
 use deno_core::error::type_error;
 use deno_core::error::AnyError;
+use deno_core::op;
+
```
bartlomieju, line 28: Ditto

ext/crypto/encrypt.rs
```diff
@@ -26,6 +27,7 @@ use ctr::flavors::Ctr64BE;
 use ctr::flavors::CtrFlavor;
 use deno_core::error::type_error;
 use deno_core::error::AnyError;
+
```
bartlomieju, line 30: Ditto

ext/crypto/generate_key.rs
```diff
@@ -3,6 +3,8 @@ use std::rc::Rc;
 
 use crate::shared::*;
 use deno_core::error::AnyError;
+
```
bartlomieju, line 6: Ditto

ext/crypto/import_key.rs
```diff
@@ -81,11 +82,13 @@ pub enum ImportKeyResult {
   #[serde(rename_all = "camelCase")]
   Ec { raw_data: RawKeyData },
   #[serde(rename_all = "camelCase")]
+  #[allow(dead_code)]
```
bartlomieju, line 85: Why is it dead code now?

ops/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "deno_ops"
+version = "0.0.1"
```
bartlomieju, line 3: Make it `0.1.0` before publishing

ops/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+edition = "2021"
+license = "MIT"
+readme = "../README.md"
+description = "proc macro for writing Deno Ops"
```
bartlomieju, line 7:
```suggestion
description = "Proc macro for writing Deno Ops"
```

ops/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+fn core_import() -> TokenStream2 {
+  let found_crate =
+    crate_name("deno_core").expect("deno_core not present in `Cargo.toml`");
+
+  match found_crate {
+    FoundCrate::Itself => {
+      // TODO(@littledivy): This won't work for `deno_core` examples
+      // since `crate` does not refer to `deno_core`.
+      // examples must re-export deno_core to make this work
+      // until Span inspection APIs are stabalized.
```
bartlomieju, line 24: Now I see why examples had to `use deno_core::*`. I think it's quite confusing, but seems there's no way around it currently. Please open an issue to change it before landing this PR

ops/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+
+/// Generate the body of a v8 func for an async op
+fn codegen_v8_async(core: &TokenStream2, f: &syn::ItemFn) -> TokenStream2 {
+  let a = codegen_arg(core, &f.sig.inputs[1], "a", 2);
+  let b = codegen_arg(core, &f.sig.inputs[2], "b", 3);
+  let type_params = &f.sig.generics.params;
+
+  quote! {
+    use #core::futures::FutureExt;
+    // SAFETY: Called from Deno.core.opAsync. Which retrieves the index using opId table.
```
bartlomieju, line 99: Can it happen that we're unable to retrieve `opId`?

bartlomieju (changes requested):
Looks great and the performance improvements are impressive.

I found a few red flags in the code, especially around changing visibility qualifiers for several `deno_core` structs

ops/lib.rs
AaronO: Yes, this actually a failure mode I bumped into a will fix, if a func name isn't in the table, we should handle that JS side might impact perf but we can optimize that out with other changes

core/examples/eval_js_value.rs
littledivy: This is only for examples residing in the `deno_core` crate.

core/bindings.rs
AaronO: No it's fine,`deno_ops` is completely compile-time but the code it generates may need to call these helpers and all op code inherently depends on deno_core so nothing abnormal there

core/error_codes.rs
AaronO: Ditto

core/examples/disable_ops.rs
AaronO: `decl` conceptually is all of the revelant metadata required for declaring an op, right now that's a name and a function pointer. But thay may evolve. `v8_cb` returns the function pointer component of `decl()`.

The idea being that we want the op to preserve it's original name but ultimately call a different implementation.

Happy to tweak our middleware impl so middlewares return `::decl()` and handles the name preservation

core/lib.rs
AaronO: Code generated ops call into some of these functions, the cleaner solution might be to group the functions ops call into a single mod

core/bindings.rs
littledivy: > What if you want to create some ops during snapshotting but don't want for them to be available after the snapshot is loaded?
 
Sure, no problem with that. It won't be available because the function reference expires.

> AFAIK this can happen in TSC snapshot which needs some ops to create a snapshot, but they should be exposed during runtime.

Yeah, this feature is added to support TSC ops. This handles the case when ops are not part of the snapshot but you still need them at runtime (during build.rs)

ext/crypto/import_key.rs
littledivy: It was dead code earlier too but the warning never showed up 🤷

core/runtime.rs
littledivy: No longer public.

core/runtime.rs
littledivy: No longer public.

core/lib.rs
littledivy: OpMetrics is no longer public. OpsTracker is used by the macro generated code to call `op_state.tracker.track_sync()`

core/ops.rs
littledivy: OpsTracker is used by the macro generated code to call op_state.tracker.track_sync()

ops/lib.rs
AaronO: Created the issue https://github.com/denoland/deno/issues/13939

core/01_core.js
AaronO: Yep, though we can restore the old code it's not a material change IMO

core/examples/disable_ops.rs
AaronO: Will do in a follow-up

AaronO (approved):
LGTM

This reduces sync-op overhead by ~50-60%, a few follow-ups but good foundations for a first pass !

core/lib.rs
lucacasonato: Please expose only the functions that are actually being used (re-export from `lib.rs`) rather than making both of these modules completely public.

lucacasonato (dismissed):
`deno_core::bindings` and `deno_core::error_codes` should not be made completely public. Just re-export `deno_core::error_codes::get_error_code` and `deno_core::bindings::throw_type_error` from `lib.rs`.

Consider a new `deno_core::__internal` `pub mod` in `lib.rs` that you export all the things used by `deno_op` from, and mark it as `#[doc(hidden)]`. That way we don't clutter `deno_core`'s rustdoc page with internals :-)

bartlomieju:
> `deno_core::bindings` and `deno_core::error_codes` should not be made completely public. Just re-export `deno_core::error_codes::get_error_code` and `deno_core::bindings::throw_type_error` from `lib.rs`.
> 
> Consider a new `deno_core::__internal` `pub mod` in `lib.rs` that you export all the things used by `deno_op` from, and mark it as `#[doc(hidden)]`. That way we don't clutter `deno_core`'s rustdoc page with internals :-)

Great suggestion, I agree with that

bartlomieju (approved):

ry (approved):
LGTM - awesome work Divy! 

Ops have evolved a lot since the first version of Deno, but the concept remains the same - a unified V8 calling infrastructure. That we're able to optimize all calls at once with this change is amazing. That this work opens the door to yet more optimizations is inspiring.

Issue #7409, opened 2020-09-10
all Deno commands with a `--cert` option should automatically reference `DENO_CERT` 


SpadeAceman:
I'm using Deno version 1.3.3 under Windows, and must use a `DENO_CERT` environment variable to help Deno scripts connect to the internet. While this works for `deno run`, other Deno commands with a `--cert` option still require me to manually specify the value of `DENO_CERT` on the command line.

For example, entering `deno upgrade` produces the following error message...

```
Checking for latest version
WARN RS - rustls::session:712 - Sending fatal alert BadCertificate
error: error sending request for url (https://github.com/denoland/deno/releases/latest): error trying to connect: invalid certificate: UnknownIssuer
```

...whereas entering `deno upgrade --cert %DENO_CERT%` works. Why doesn't it automatically reference `DENO_CERT` the way `deno run` does?

(I've not tested other commands with a `--cert` option, such as `deno install`.)

Edit: possibly related to #7390.

lucacasonato:
Looking at this more it seems that the env var is only respected for `SourceFileFetcher`. cc @bartlomieju

SpadeAceman:
I saw fixes related to DENO_CERT in the release notes for v1.18.1, but those fixes don't appear to have resolved this particular issue. Could the changes made in #13467 help pave the way for this issue to be fixed? 

For reference, here's what v1.18.1 gives me for `deno upgrade` under Windows 10:

```
Looking up latest version
Sending fatal alert BadCertificate
error: error sending request for url (https://github.com/denoland/deno/releases/latest): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer

Caused by:
    0: error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
    1: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
```

I still have to specify `deno upgrade --cert %DENO_CERT%` on the command line to get around this.

sylc:
Still happening on deno 1.19.2

lucacasonato:
This should now be fixed. If you run into this again, please open a new issue mentioning the subcommand you ran into this with.

janmagnusdev:
`--cert` and `DENO_CERT` don't seem to work with `deno add`. 
I am using a custom registry in my `~/.npmrc`.
`deno add` seems to not even expose the `--cert` option.

PR #13862, opened 2022-03-07
fix(cli): Add support for DENO_CERT in upgrade command

sylc:
Related to https://github.com/denoland/deno/issues/7409

Add support for DENO_CERT in the `deno upgrade` command


dsherret (approved):
LGTM. Thanks @sylc!

Issue #13239, opened 2021-12-30
Support for realms/contexts in `deno_core` (in preparation for `ShadowRealm`)


andreubotella:
In JS, code execution is always tied to some realm (or "context" in V8 terminology), which provides a global object and a set of built-ins (`Object`, `Array`, `Error`, etc.) which is different from those of other realms. Multiple realms can share an event loop (= a thread), and those that do share one may be able to access objects from a different realm. In the web, a same-origin `<iframe>` is a different realm from the top-level page (`iframe.contentWindow.Array !== window.Array`). Currently there is no way to create realms or run code in them purely with JS built-ins, that's something that host environments (the web / Node.js / Deno) would have to provide, so it's fine for Deno to not support them.

But there is currently a stage-3 TC39 proposal called [`ShadowRealm`](https://github.com/tc39/proposal-shadowrealm) that would allow creating realms, as a way of sandboxing untrusted code. This proposal is not close to shipping anywhere, and it's only now starting to get implemented in V8, but currently Deno assumes that there's a single V8 context available at all times, and it's better to start refactoring things to support realms with plenty of time to spare.

Since `ShadowRealm`s are meant to be a sandbox primitive, there would be no way to make objects from the parent realm available inside the `ShadowRealm` and vice versa. This simplifies to some extent the requirements needed, because that way there is no need to comb through the JS code looking for wrong uses of `instanceof`. That said, these are the requirements that seem to be needed for `ShadowRealm` (and I'm sure I'm forgetting some):

- [x] Add a Rust API to create V8 contexts associated with a `JsRuntime` and run scripts in them (similar to `JsRuntime::execute_script`), to simplify running tests. Modules can be left for later. #14019
- [x] Make sure bindings and extension scripts are initialized correctly in the context, both when using snapshots and when not. This includes making sure the ops cache in different realms can't go out of sync. #14019
- [ ] Make sure any object return values of bindings and ops, as well as any exceptions they throw, are objects in the current realm (i.e. their prototype is not, say, a different realm's `Object.prototype`).
- [ ] Make sure async ops don't resolve a promise in the wrong realm. #14734
- [ ] `js_promise_reject_cb`, `js_uncaught_exception_cb` and `js_wasm_streaming_cb` should probably be realm → callback (weak)maps.
- [ ] Modules. Per the HTML spec, every realm has its own module map, although module fetches can be cached across realms.
[...]

PR #13863, opened 2022-03-07
cleanup(core): Don't use handle_scope in pump_v8_message_loop

espindola:
A v8_isolate is sufficient.

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -674,16 +674,16 @@ impl JsRuntime {
   }
 
   fn pump_v8_message_loop(&mut self) {
-    let scope = &mut self.handle_scope();
+    let isolate = &mut self.v8_isolate();
     while v8::Platform::pump_message_loop(
       &v8::V8::get_current_platform(),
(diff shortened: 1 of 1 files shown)


espindola:
> Is there a particular reason for this change?

I noticed while looking at https://github.com/denoland/deno/issues/13239. Since we want to support multiple v8::context, not using a v8::context when one is not needed is probably a good thing.

bartlomieju:
> > Is there a particular reason for this change?
> 
> I noticed while looking at #13239. Since we want to support multiple v8::context, not using a v8::context when one is not needed is probably a good thing.

I see, to be honest I'd prefer not to land this change as is. Nothing changes in behavior at the moment and this change can't be logically connected to anything until #13239 is addressed. Unless you have a burning need for it right now I think we should just wait on a PR for #13239

espindola:
I have no need for it. It is just the first place I looked when trying to figure out where do we assume there is a single v8::context.

PR #13864, opened 2022-03-07
perf(core): use raw v8 get/set data

AaronO:
Instead of rusty_v8's Annex to reduce lookup overhead

```
Before:
test bench_is_proxy   ... bench:      20,051 ns/iter (+/- 860)
test bench_op_async   ... bench:     391,111 ns/iter (+/- 5,354)
test bench_op_nop     ... bench:      52,359 ns/iter (+/- 1,942)
test bench_op_pi_json ... bench:      54,409 ns/iter (+/- 1,326)

After:
test bench_is_proxy   ... bench:      20,292 ns/iter (+/- 480)
test bench_op_async   ... bench:     385,609 ns/iter (+/- 5,775)
test bench_op_nop     ... bench:      46,042 ns/iter (+/- 1,785)
test bench_op_pi_json ... bench:      49,903 ns/iter (+/- 668)
```

Blocked on https://github.com/denoland/rusty_v8/pull/911

### DIFF
--- a/core/modules.rs
+++ b/core/modules.rs
@@ -7,6 +7,7 @@ use crate::module_specifier::ModuleSpecifier;
 use crate::resolve_import;
 use crate::resolve_url;
 use crate::runtime::exception_to_err_result;
+use crate::JsRuntime;
 use crate::OpState;
 use anyhow::Error;
 use futures::future::FutureExt;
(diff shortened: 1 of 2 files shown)


aapoalas:
Is the Annex actually even needed after this? It seems to only contain the `create_param_allocation` (snapshot params?) which is never accessed from it, only kept by it. If that memory is needed to be kept, it looks like it could be just another internal slot instead of indirected through the Annex.

aapoalas:
Is the Annex actually even needed after this? It seems to only contain the `create_param_allocation` (snapshot params?) which is never accessed from it, only kept by it. If that memory is needed to be kept, it looks like it could be just another internal slot instead of indirected through the Annex.

PR #13870, opened 2022-03-08
build: require safety comments on unsafe code

lucacasonato:
I have enabled the `undocumented_unsafe_blocks` and `missing_safety_doc` clippy rules. These ensure all unsafe blocks have `// SAFETY:` comments.

@AaronO please add safety comments to the remaining `unsafe` blocks in `serde_v8`. I could not figure out the reasoning why they are safe for all of them.


littledivy (approved):

AaronO (approved):
LGTM

As discussed out-of-band, I have a cleanup in mind that should streamline most of serde_v8's `unsafe` code

serde_v8/src/de.rs
```diff
@@ -378,8 +378,10 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
       let len = v8str.length();
       let mut buffer = Vec::with_capacity(len);
       #[allow(clippy::uninit_vec)]
+      // SAFETY: this is fine, because length == capacity (see previous line),
+      // and we are about to initalize the buffer with data from the v8 string.
```
andreubotella, line 382: Not actually fine. The safety requirements for `Vec::set_len` include making sure that the elements `0..len` are initialized, which they're not in this case. https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len has a similar sample where the FFI method is first passed `vec.as_mut_ptr()`, and the length is changed afterwards.

lucacasonato: @andreubotella It is initialized right after this call through the call to V8 (it writes to the buffer). The guarantee holds.

littledivy:
The CI is failing because of missing serde_v8 safety comments. @AaronO

serde_v8/src/de.rs
littledivy: Yeah, it's fine as long as we are not reading from it. (before passing to V8)

andreubotella: My understanding is that they has to be initialized at the time of the call.

lucacasonato: @andreubotella What we are doing here is essentially the FFI example from the standard library docs, except that the "ffi" call (the rust call to v8) takes a `vec` that it then turns back into `(pointer, len)`, rather than taking a `(pointer, len)` directly.

andreubotella: The relevant difference is that in the std example at no point there are uninitialized elements in the vec, because `set_len` is only called after they are initialized.

bartlomieju (approved):
LGTM, all ignores have added TODOs which we should address in follow up PRs. Thanks everyone who worked on this PR.

PR #13875, opened 2022-03-08
chore: update regex to 1.5.5

littledivy:
https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3128,9 +3128,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.5.4"
+version = "1.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM

bartlomieju:
We should update it across all repos.

lucacasonato:
Not questioning whether we should do this, but do we ever parse untrusted regular expressions? I don't think so.

crowlKats:
@lucacasonato and over at rust-urlpattern we do as well

lucacasonato:
> @lucacasonato and over at rust-urlpattern we do as well

Never completely arbitrary regular expressions with `regexp`. All arbitrary regexps are handled by V8.

crowlKats:
Oh yea, you might be right. i keep forgetting that.

PR #13876, opened 2022-03-08
fix: shell completion hints

crowlKats:
cc @mcornella


bartlomieju:
Is there any way we can test this?

crowlKats:
I really don't think so

mcornella:
All seems to work correctly! 👌🏻

bartlomieju (approved):

PR #13892, opened 2022-03-09
types: add Deno.PermissionOptions

bartlomieju:
https://github.com/denoland/deno/pull/13713#discussion_r821764638

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -114,7 +114,9 @@ declare namespace Deno {
    * See: https://no-color.org/ */
   export const noColor: boolean;
 
-  export type PermissionOptions = "inherit" | "none" | {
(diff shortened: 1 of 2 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -114,6 +114,139 @@ declare namespace Deno {
    * See: https://no-color.org/ */
   export const noColor: boolean;
 
+  export type PermissionOptions = "inherit" | "none" | {
```
kitsonk, line 117: The export is "unnecessary" but consistent.

I would consider doing something like this, as it will make the documentation on doc.deno.land better:

```suggestion
  export type PermissionOptions = "inherit" | "none" | PermissionOptionsObject;
  
  export interface PermissionOptionsObject {
```

dsherret (approved):
LGTM

kitsonk (approved):
LGTM

PR #13894, opened 2022-03-09
fix(ext/crypto): use EcKeyImportParams dictionary

panva:
as per https://w3c.github.io/webcrypto/#webidl-410131068

currently, universal code referencing the `EcKeyImportParams` interface doesn't work

```
Uncaught TypeError: TS2304 [ERROR]: Cannot find name 'EcKeyImportParams'
```


littledivy (approved):
LGTM, Thanks!

PR #13896, opened 2022-03-10
1.19.3

kt3k:


kt3k:
Note: I skipped `fix(ext/http): drop content-length header on compression` #13866 because it depends on auto-compression feature (#13769), and it's only meaningful when that feature is available

ext/http/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 
 [package]
 name = "deno_http"
-version = "0.33.0"
+version = "0.34.0"
```
satyarohith, line 5:
```suggestion
version = "0.31.0"
```
Do you think we can use the old version without bumping?

Release a new version would create a situation where we have the compression feature in `0.32.0/0.33.0` but not in `0.34.0`. Since we don't have any changes, we probably can use `0.31.0`?

kt3k: Ah, we published 0.32.0 and 0.33.0 independent of CLI release process?

Ok. Let's try that

kt3k:
Will release with this diff in the release script when CI passed

```diff
diff --git a/tools/release/02_publish_crates.ts b/tools/release/02_publish_crates.ts
index 1bb02f460..8aca6d4f1 100644
--- a/tools/release/02_publish_crates.ts
+++ b/tools/release/02_publish_crates.ts
@@ -10,6 +10,10 @@ const dependencyCrates = getCratesPublishOrder(workspace.getDependencyCrates());
 
 try {
   for (const [i, crate] of dependencyCrates.entries()) {
+    if (crate.name === "deno_http") {
+      console.log("skipping deno_http");
+      continue;
+    }
     await crate.publish();
     console.log(`Published ${i + 1} of ${dependencyCrates.length} crates.`);
   }
```

Releases.md
```diff
@@ -6,6 +6,15 @@ https://github.com/denoland/deno/releases
...
 https://github.com/denoland/deno_install
 
+### 1.19.3 / 2022.03.10
+
+- fix(ci): restore compatibility with older glibc (#13846)
+- fix(test): typecheck blocks annotated with long js/ts notations (#13785)
+- perf(core): micro-optimize OpsTracker (#13868)
+- perf(serde_v8): avoid SerializablePkg allocs (#13860)
+- perf(serde_v8): optimize ByteString deserialization (#13853)
+- perf(ext/web): optimize atob/btoa (#13841)
```
satyarohith, line 16:
```suggestion
- fix(ci): restore compatibility with older glibc (#13846)
- fix(test): typecheck blocks annotated with long js/ts notations (#13785)
- perf(core): micro-optimize OpsTracker (#13868)
- perf(ext/web): optimize atob/btoa (#13841)
- perf(serde_v8): avoid SerializablePkg allocs (#13860)
- perf(serde_v8): optimize ByteString deserialization (#13853)
```
sort

satyarohith (approved):
LGTM

Releases.md
kt3k: Addressing this in an independent PR later, to avoid another CI run in this PR

satyarohith: No worries.

kt3k:
deno_runtime@latest cannot be built with `deno_http@0.31.0` because dependent deno_core versions are incompatible:

```
   --> build.rs:166:7
    |
166 |       deno_http::init(),
    |       ^^^^^^^^^^^^^^^^^ expected struct `deno_core::Extension`, found struct `deno_core::extensions::Extension`
    |
    = note: perhaps two different versions of crate `deno_core` are being used?
```

satyarohith:
@kt3k looks like we don't have an option apart from upgrading deno_http. Not sure how the CI passed.

kt3k:
I think I have to publish `deno_http@0.34.0` to make this releasable.

I guess Deploy is the only project which depends on it. Is that correct?

kt3k:
I'll dump deno_http to 0.34.0

kt3k:
Looks flaky (test release ubuntu-20.04-xl)

```
failures:

        "/fetch/api/basic/response-null-body.any.worker.html - Response.body is null for responses with status=304 (method=OPTIONS)"
```

kt3k:
Ah, `tools/release/02_publish_crates.ts` automatically skips already published crates. So I didn't need the above patch.

スクリーンショット 2022-03-10 19 55 58

kt3k:
Published all crates. Merging

PR #13899, opened 2022-03-10
fix(ext/fetch): Connect async error stack with user code

### LINKED ISSUE #11859: Need complete stack trace for unhandled promise rejections.
If you have un unhandled exception in a promise, that causes Deno to shut down.  Before shutting down it prints an error message like this:

```
error: Uncaught (in promise) Error: No such host is known. (os error 11001)
    at async Object.connect (deno:extensions/net/01_net.js:219:13)
```

The problem is that the stack trace doesn't point to any of _my_ code.  Normally when I see a stack trace it goes all the way back to my code.

I have also tried asking VS code to break on all exceptions.  Then I could see the stack trace live.  But VS Code / Deno ignored that request.

I would like something to help me find the source of this problem.  I can reproduce this problem at will, but I can't find the source of it.

[More details.](https://stackoverflow.com/questions/68811003/uncaught-in-promise-how-to-find-the-source-of-the-problem-in-a-deno-program?noredirect=1#comment121650904_68811003)

Thank you.

### LINKED ISSUE #13897: Make `fetch()` errors more helpful.
If you save the following to a file and run it:
```js
fn1();

function fn1() {
  fn2();
}

async function fn2() {
  const response = await fetch("file:///does/not/exist/");
  console.log(await response.text());
}
```
The following error is printed:
```
error: Uncaught (in promise) TypeError: NetworkError when attempting to fetch resource.
    at async mainFetch (deno:ext/fetch/26_fetch.js:280:14)
```
apart from the stack trace only partially being shown (https://github.com/denoland/deno/issues/11859), I think it could be useful if the failing url was logged. Similar to what is shown in browsers:
image

nayeemrmn:
Fixes #13897.


bartlomieju:
@nayeemrmn I'm not sure we can say that this PR fixes https://github.com/denoland/deno/issues/11859. It seems to only special-case `fetch` API, but the floating promises in other APIs stilo won't report correct stack trace.

nayeemrmn:
> @nayeemrmn I'm not sure we can say that this PR fixes #11859. It seems to only special-case `fetch` API, but the floating promises in other APIs stilo won't report correct stack trace.

Both added reports there seem to be for `fetch()`. I believe my first response still applies regarding the `Deno.connect()` one. Are there any other repros? It's most likely just #13897 left.

jespertheend:
Tbh I submitted #13897 for logging the url in an error message. I figured that #11859 was already meant for stack traces. That said though, having stack traces, even if it's only for fetch, is still much appreciated :)

nayeemrmn:
In that case #11859 is a non-issue and we should continue in #13897.

nayeemrmn:
@lucacasonato Please review

bartlomieju (approved):
LGTM, I'll wait for Luca's review before landing. Thank you @nayeemrmn

bartlomieju:
Discussed offline with Luca, this is fine.

PR #13905, opened 2022-03-11
refactor(core): validate promise id in refOp

kt3k:
This PR adds the wrappers of the existing `core.refOp` and `core.unrefOp`, and checks if the given promise ids are valid.

Currently we don't check whether the promise id is valid or not. If a wrong promise id is given to `core.unrefOp`, that id remains in pending_ops vector, and causes the exit condition incorrect (Deno program can exit with valid ops still remaining).

---
This implements the idea by @andreubotella in the comment https://github.com/denoland/deno/pull/12889#discussion_r756966729

### DIFF
--- a/cli/tests/unit/opcall_test.ts
+++ b/cli/tests/unit/opcall_test.ts
@@ -20,13 +20,6 @@ Deno.test(async function sendAsyncStackTrace() {
   }
 });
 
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM

PR #13906, opened 2022-03-11
experiment(core): rust promise ring

AaronO:


core/promise_ring.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  // TODO(@AaronO): decide if this is useful (but it should match length of futures stream so ...)
+  // pub(crate) fn len(&self) -> u32 {
+  //   self.len
+  // }
+
+  pub(crate) fn has(&self, id: PromiseId) -> bool {
+    self.ring.get(Self::ring_idx(id)).is_some() || self.map.contains_key(&id)
```
aapoalas, line 34: Can the ring and map be desynced such that the double check is necessary?

core/promise_ring.rs
```diff
@@ -0,0 +1,73 @@
...
+const RING_SIZE: usize = 4 * 1024;
+
+// NOTE: could generalize to a "RingMap"
+pub struct PromiseRing {
+  len: u32,
+  cursor: u32,
+  // TODO(@AaronO): Maybe optimize by tracking existence with a bitset
+  // would reduce mem usage from 2*usize*RING_SIZE to (1+1/8)*usize*RING_SIZE
+  // with the current approach the Option tag "wastes" 32kb per isolate
+  ring: Vec<Option<PromiseResolver>>,
```
aapoalas, line 14: Implementing the "ring" as a bitset does definitely seem like a useful idea.

core/promise_ring.rs
AaronO: Only checking either of them should be sufficent (based on if the id is in the range of the ring or not), but it's unimportant to optimize since this should be called infrequently for ref/unref ops

core/promise_ring.rs
```diff
@@ -0,0 +1,73 @@
...
+    resolver
+  }
+
+  pub(crate) fn head(&self) -> PromiseId {
+    self.cursor as PromiseId
+  }
+
+  pub(crate) fn put(&mut self, resolver: PromiseResolver) -> PromiseId {
+    let id = self.cursor;
+    self.cursor += 1;
```
aapoalas, line 60: Some of my work touches on a ring buffer of sorts, where the implementation has been done such that the cursor is always traveling backwards. I don't know why that is but I have some faith in the original implementers that they didn't do it randomly but with good reason.

So, maybe give it a minute of thought if that would make sense in this case? (Though the use cases are quite different indeed.)

aapoalas:
Seems like this'll be one more splendid thing for performance :)

core/promise_ring.rs
aapoalas: Further thinking on this: The ring buffer that I referred does not have a concept of  IDs in the sense of the PromiseRing, instead it only has the "head" which keeps consuming the tail as it goes. In that scenario the inverted order probably has benefits but for Deno's use case it likely does not make sense since indexing by IDs needs to work etc.

andreubotella:
For `Deno.core.refOp` and `unrefOp`, currently if something other than a `PromiseId` is passed, the Rust binding never gets called because that id is not in the promise map. However, in this PR it's directly passed to the binding, which tries to convert it into a `PromiseId`, and throws if it can't.

Combined with changing `PromiseId` to be a type alias of `u32` rather than `i32`, this breaks https://github.com/denoland/deno/blob/main/cli/tests/unit/ref_unref_test.ts

core/promise_ring.rs
andreubotella: I'm not sure this is right. If a promise resolver is moved to the map, it's because a new resolver with a different id took its slot. If it is then removed from the map but the resolver that took its slot hasn't been removed yet, this method would return `true`.

Also, `self.ring.get(idx)` has type `Option<&Option<PromiseResolver>>`, and the outer option is some if the index is a valid index into the slice – which it's guaranteed to be by `Self::ring_idx`. So `self.ring.get(Self::ring_idx(id)).is_some()` always returns true, even if cursor is 0.
```suggestion
    self.ring[Self::ring_idx(id)].is_some() || self.map.contains_key(&id)
```

kt3k:
This work is continued in #14396. Closing in favor of it.

PR #13915, opened 2022-03-11
serde_v8 improvements to avoid hitting unimplemented codepaths

arthurprs:
I've made some chances in order to avoid silent data loss and hitting unimplemented code paths, some of it might be opinionated though. Let me know what you think.

Rusty v8 related
- Optimize object deserialization by avoid getting each keys twice
- Handle Javascript Map objects correctly
- Handle chars and bytes without panicking

There are also some changes wrt. deno-core dependencies so it's easier to use outside of deno.

- Downgrade minor version of indexmap to avoid ahash dependency cycle


serde_v8/src/ser.rs
```diff
@@ -430,11 +432,24 @@ impl<'a, 'b, 'c> ser::SerializeMap for MapSerializer<'a, 'b, 'c> {
...
 
 impl<'a, 'b, 'c> Serializer<'a, 'b, 'c> {
   pub fn new(scope: ScopePtr<'a, 'b, 'c>) -> Self {
-    Serializer { scope }
+    Serializer {
+      scope,
+      intern_strings: false,
+    }
+  }
+  pub fn with_string_interning(
```
aapoalas, line 445: `interning` is the act of working as an intern, this should perhaps be `with_string_internalizing`

arthurprs: Ops. That makes total sense.

That said, I think string internalization on map keys might be too workload specific. So maybe it's best left out of this PR, or implemented on the rust side.

AaronO: > Ops. That makes total sense.
> 
> That said, I think string internalization on map keys might be too workload specific. So maybe it's best left out of this PR, or implemented on the rust side.

Yes, I was going to point that out.

Internalizing struct field strings is fine since the set is finite and relatively small (for deno's ops), internalizing all Map keys gives users a footgun which can lead to unbounded memory "leaks". I would prefer this to be removed from this PR

serde_v8/src/de.rs
```diff
@@ -307,26 +325,33 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
...
-    let prop_names = obj.get_own_property_names(self.scope);
-    let mut keys: Vec<magic::Value> = match prop_names {
-      Some(names) => from_v8(self.scope, names.into()).unwrap(),
-      None => vec![],
-    };
-    let keys: Vec<v8::Local<v8::Value>> = keys
-      .drain(..)
-      .map(|x| x.into())
-      // Filter keys to drop keys whose value is undefined
-      // TODO: optimize, since this doubles our get calls
```
arthurprs, line 319: Should this behavior be maintained?

arthurprs: Any guidance for this? It would be great to get this merged.

bartlomieju:
@AaronO please review

serde_v8/src/de.rs
arthurprs: I retained the old behavior, with hopes this can get merged.

arthurprs:
I updated the PR to fix the merge conflicts and to account for other recent PRs.

bartlomieju:
@arthurprs sorry for late response, please rebase and we'll merge it

arthurprs:
No problem. The PR is now merged with the upstream repo.

serde_v8/de.rs
```diff
@@ -317,26 +312,33 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
...
-
-    let map = MapAccess {
-      obj,
-      keys,
-      pos: 0,
-      scope: self.scope,
-    };
-    visitor.visit_map(map)
+
+    if obj.is_map() {
```
aapoalas, line 316:
```rs
if let Ok(map) = v8::Local::<v8::Map>::try_from(self.input) {
```

core/Cargo.toml
```diff
@@ -16,7 +16,9 @@ path = "lib.rs"
 anyhow = "1.0.57"
 deno_ops = { path = "../ops", version = "0.20.0" }
 futures = "0.3.21"
-indexmap = "1.8.1"
+# Stay on 1.6 to avoid a dependency cycle in ahash https://github.com/tkaitchuck/aHash/issues/95
+# Projects not depending on ahash are unafected as cargo will pull any 1.X that is >= 1.6.
+indexmap = "1.6"
```
bartlomieju, line 21: Is this still required?

arthurprs: I believe so, there hasn't been much movement trying to break this dependency cycle https://github.com/tkaitchuck/aHash/issues/95. Luckily it doesn't pose any significant problems for Deno.

bartlomieju: Right, but we were already on `v1.8.1`, so why downgrade?

arthurprs: This is not a real downgrade but a downgrade of the "minimum" minor version for `1.*`. `deno` `cargo.lock` still defines 1.8.1 and will go up as needed as long as it's `< 2`.

At the same time, if we don't downgrade the minimum minor version any crate depending on `deno_core` (like the project I'm trying to use deno in) also also `ahash` (or any of the other problematic dependencies) it won't compile as cargo won't be able solve the dependency cycle.

bartlomieju (approved):
LGTM, thank you @arthurprs, sorry it took so long to land

PR #13916, opened 2022-03-11
chore(test): fix flaky tasks

dsherret:
The tasks were flaky. Now it's been rewritten to use os pipes so when it writes to stdout it will write directly to that pipe instead of through a series of tokio channels.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1068,12 +1068,13 @@ dependencies = [
 
 [[package]]
 name = "deno_task_shell"
-version = "0.1.6"
+version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -1068,12 +1068,13 @@ dependencies = [
...
-version = "0.1.6"
+version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ad23aacc3db4f37eb88fb9c874a85a4abe5eab14a98fc070ee8df9e204f243e6"
+checksum = "5a528d67d2c9efba7316669eee5889e0dc150297799d3824d35f187596bec94b"
 dependencies = [
  "anyhow",
  "futures",
+ "os_pipe",
  "pretty_assertions 1.1.0",
```
dsherret, line 1078: Hmmm ಠ_ಠ... I will fix this.

bartlomieju (approved):

Issue #13919, opened 2022-03-11
integration::compat::top_level_fail_esm is flaky


dsherret:
```
---- integration::compat::top_level_fail_esm stdout ----
deno_exe path D:\a\deno\deno\target\release\deno.exe
deno_exe args test --compat --unstable -A --quiet compat/test_runner/top_level_fail_esm.mjs
deno_exe testdata path "D:\\a\\deno\\deno\\cli\\tests\\testdata"
output path D:\a\deno\deno\cli\tests\testdata\compat/test_runner/top_level_fail_esm.out
[test_util\src\lib.rs:1847] part = ")\n\nerror: Uncaught AssertionError: Values are not strictly equal:\n\n\n    [Diff] Actual / Expected\n\n\n-   10\n+   20\n\n"
[test_util\src\lib.rs:1847] i = 1
[test_util\src\lib.rs:1856] "exit false " = "exit false "
[test_util\src\lib.rs:1856] i = 1
OUTPUT

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (152ms)

error: error sending request for url (https://deno.land/std@0.129.0/node/module.ts): connection error: connection reset

OUTPUT
EXPECTED

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out ([WILDCARD])

error: Uncaught AssertionError: Values are not strictly equal:


    [Diff] Actual / Expected


-   10
+   20

[WILDCARD]
EXPECTED
thread 'integration::compat::top_level_fail_esm' panicked at 'pattern match failed', test_util\src\lib.rs:1807:7
stack backtrace:
   0:     0x7ff71a59792f - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ff71a59792f - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a\/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff71a59792f - std::sys_common::backtrace::_print_fmt
                               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a\/library\std\src\sys_common\backtrace.rs:67
[...]

PR #13921, opened 2022-03-12
test: fix flaky compat tests

bartlomieju:
Closes https://github.com/denoland/deno/issues/13919


bartlomieju:
This should also speed up CI tests. We were downloading a lot of files from `https://deno.land/std` for each of these tests 🤦

PR #13922, opened 2022-03-12
feat(task): log task script

bartlomieju:
Logs task name and associated script with additional args. 
This is disabled if "--quiet/-q" flag is present.

```
// before:
$ deno task exit_code_5 
10
// after:
$ deno task exit_code_5
Task exit_code_5 echo $(echo 10 ; exit 2) && exit 5
10
```


dsherret (approved):
Oh neat. It "just works". LGTM

PR #13923, opened 2022-03-12
perf(web): Optimize `TextDecoder` by adding a new `U16String` type

### LINKED ISSUE #10978: extensions/web: Improving the performance of the text encoding APIs

andreubotella:
Before:
```
Benchmark #10: deno run cli/tests/testdata/text_decoder_perf.js 
  Time (mean ± σ):      90.1 ms ±   1.3 ms    [User: 90.1 ms, System: 7.9 ms]
  Range (min … max):    87.7 ms …  94.1 ms    32 runs
```

After:
```
Benchmark #10: deno run cli/tests/testdata/text_decoder_perf.js 
  Time (mean ± σ):      51.5 ms ±   0.7 ms    [User: 53.4 ms, System: 8.2 ms]
  Range (min … max):    50.0 ms …  53.0 ms    57 runs
```

﻿Closes #10978.


ext/web/lib.rs
```diff
@@ -299,31 +300,28 @@ fn op_encoding_decode(
...
-    decoder
-      .max_utf8_buffer_length(data.len())
-      .ok_or_else(|| range_error("Value too large to decode."))?
-  };
+  let max_buffer_length = decoder
+    .max_utf16_buffer_length(data.len())
+    .ok_or_else(|| range_error("Value too large to decode."))?;
 
-  let mut output = String::with_capacity(max_buffer_length);
+  let mut output = U16String::with_zeroes(max_buffer_length);
```
evanwashere, line 315: avoid initializing memory twice
```suggestion
  let mut output = U16String(Vec::with_capacity(max_buffer_length));

  // SAFETY: decoder initializes memory and truncates rest
  unsafe {
    output.0.set_len(max_buffer_length);
  }
```

andreubotella: As I argued elsewhere, that might be UB because it creates `u16` values backed by uninitialized memory, and uninitialized isn't a valid value for `u16`.

But even if this wasn't UB, it doesn't make a significant difference in the benchmarks:

With `with_zeroes`:
```
Benchmark #10: /home/abotella/Projects/forks/denoland/deno/target/release/deno run cli/tests/testdata/text_decoder_perf.js 
  Time (mean ± σ):      51.6 ms ±   1.0 ms    [User: 53.6 ms, System: 8.3 ms]
  Range (min … max):    49.8 ms …  53.3 ms    57 runs
```

With `set_len`:
```
Benchmark #10: /home/abotella/Projects/forks/denoland/deno/target/release/deno run cli/tests/testdata/text_decoder_perf.js 
  Time (mean ± σ):      50.2 ms ±   0.9 ms    [User: 53.3 ms, System: 7.2 ms]
  Range (min … max):    48.9 ms …  53.2 ms    58 runs
```

devcat: time adds up quickly with bigger strings and continuous decoding
i'm pretty sure no one wants text decoder with unnecessary overhead

# uninit
```
benchmark        time (avg)                     (min … max)            p75            p99          p99.5
----------------------------------------------------------- --------------------------------------------
1024          2,594 ns/iter         (1,958 ns … 215,250 ns)       2,708 ns       6,459 ns       6,876 ns
4096          5,339 ns/iter         (4,833 ns … 135,918 ns)       5,293 ns       6,375 ns       9,043 ns
16384        16,645 ns/iter         (16,209 ns … 87,251 ns)      16,418 ns      18,001 ns      22,626 ns
65536        82,750 ns/iter        (72,667 ns … 605,001 ns)      78,084 ns     299,959 ns     406,959 ns
262144      401,387 ns/iter       (389,250 ns … 640,291 ns)     407,500 ns     441,708 ns     467,209 ns
1048576   1,755,183 ns/iter   (1,579,167 ns … 2,020,292 ns)   1,782,001 ns   1,888,916 ns   1,897,000 ns
4194304   6,862,517 ns/iter   (6,668,959 ns … 8,365,793 ns)   6,859,666 ns   7,726,376 ns   8,365,793 ns
16777216 27,186,924 ns/iter (26,868,750 ns … 28,400,418 ns)  27,270,084 ns  28,172,792 ns  28,400,418 ns
```

# init
```
benchmark        time (avg)                      (min … max)            p75            p99          p99.5
------------------------------------------------------------ --------------------------------------------
1024          2,665 ns/iter          (2,083 ns … 211,875 ns)       2,792 ns       6,667 ns       7,585 ns
4096          5,640 ns/iter          (5,167 ns … 157,959 ns)       5,625 ns       6,500 ns       6,834 ns
16384        18,134 ns/iter          (17,625 ns … 86,958 ns)      17,917 ns      20,542 ns      24,751 ns
65536        86,065 ns/iter         (76,126 ns … 858,043 ns)      81,416 ns     285,834 ns     423,375 ns
262144      406,105 ns/iter        (394,584 ns … 458,209 ns)     412,543 ns     429,334 ns     433,042 ns
1048576   1,804,246 ns/iter    (1,601,667 ns … 1,974,334 ns)   1,828,917 ns   1,933,916 ns   1,944,251 ns
4194304   6,976,854 ns/iter    (6,775,542 ns … 7,313,168 ns)   7,080,417 ns   7,296,959 ns   7,313,168 ns
16777216 27,586,640 ns/iter  (27,238,958 ns … 29,126,292 ns)  27,586,876 ns  28,938,084 ns  29,126,292 ns
```

evanwashere: time adds up quickly with bigger strings and continuous decoding
i'm pretty sure no one wants text decoder with unnecessary overhead

# uninit
```
benchmark        time (avg)                     (min … max)            p75            p99          p99.5
----------------------------------------------------------- --------------------------------------------
1024          2,594 ns/iter         (1,958 ns … 215,250 ns)       2,708 ns       6,459 ns       6,876 ns
4096          5,339 ns/iter         (4,833 ns … 135,918 ns)       5,293 ns       6,375 ns       9,043 ns
16384        16,645 ns/iter         (16,209 ns … 87,251 ns)      16,418 ns      18,001 ns      22,626 ns
65536        82,750 ns/iter        (72,667 ns … 605,001 ns)      78,084 ns     299,959 ns     406,959 ns
262144      401,387 ns/iter       (389,250 ns … 640,291 ns)     407,500 ns     441,708 ns     467,209 ns
1048576   1,755,183 ns/iter   (1,579,167 ns … 2,020,292 ns)   1,782,001 ns   1,888,916 ns   1,897,000 ns
4194304   6,862,517 ns/iter   (6,668,959 ns … 8,365,793 ns)   6,859,666 ns   7,726,376 ns   8,365,793 ns
16777216 27,186,924 ns/iter (26,868,750 ns … 28,400,418 ns)  27,270,084 ns  28,172,792 ns  28,400,418 ns
```

# init
```
benchmark        time (avg)                      (min … max)            p75            p99          p99.5
------------------------------------------------------------ --------------------------------------------
1024          2,665 ns/iter          (2,083 ns … 211,875 ns)       2,792 ns       6,667 ns       7,585 ns
4096          5,640 ns/iter          (5,167 ns … 157,959 ns)       5,625 ns       6,500 ns       6,834 ns
16384        18,134 ns/iter          (17,625 ns … 86,958 ns)      17,917 ns      20,542 ns      24,751 ns
65536        86,065 ns/iter         (76,126 ns … 858,043 ns)      81,416 ns     285,834 ns     423,375 ns
262144      406,105 ns/iter        (394,584 ns … 458,209 ns)     412,543 ns     429,334 ns     433,042 ns
1048576   1,804,246 ns/iter    (1,601,667 ns … 1,974,334 ns)   1,828,917 ns   1,933,916 ns   1,944,251 ns
4194304   6,976,854 ns/iter    (6,775,542 ns … 7,313,168 ns)   7,080,417 ns   7,296,959 ns   7,313,168 ns
16777216 27,586,640 ns/iter  (27,238,958 ns … 29,126,292 ns)  27,586,876 ns  28,938,084 ns  29,126,292 ns
```

AaronO (approved):
LGTM

Deserialize would have been nice since it's a symmetric type, but can be added in a follow up

PR #13942, opened 2022-03-14
fix(core): Don't override structured clone error messages from V8

### LINKED ISSUE #13932: Structured clone serialization errors have the wrong message
The `v8::ValueSerializerImpl` trait has a `throw_data_clone_error` method that must be implemented, which provides embedders with an error message indicating why some object couldn't be serialized. Deno's implementation of that method throws an exception with that error message, which is then supposed to be caught in ext/web and turned into a `DataCloneError` `DOMException` as required by the HTML spec.

However, the resulting exception does not have the V8-provided message – it instead has the message "Failed to serialize response". This can be seen when running the code in https://github.com/denoland/deno/issues/13848#issuecomment-1059979657.

This seems to happen because the check for serialization errors in https://github.com/denoland/deno/blob/main/core/bindings.rs#L1193 does not check if an exception has been previously thrown.

andreubotella:
In the implementation of structured serialization in `Deno.core.serialize`, whenever there is a serialization error, an exception will be thrown with the message "Failed to serialize response", even though V8 provides a message to use in such cases. This change instead throws an exception with the V8-provided message, if there is one.

Closes #13932.


bartlomieju (approved):
LGTM, good fix, thank Andreu!

Issue #13943, opened 2022-03-14
`deno_ops` macro tracking issue


littledivy:
https://github.com/denoland/deno/pull/13861 Introduces the `deno_op` proc macro crate. The goal is to produce highly optimized v8 Function callbacks from existing deno ops & provide base for future v8-fast-api work.

Here we tracking down a few things to follow-up after the PR lands:

- [x] Cleanup OpDecl (@AaronO) https://github.com/denoland/deno/pull/13952
- [x] Variable op arity (@AaronO) https://github.com/denoland/deno/pull/13949
- [x] ~~Optimize away unused args (i.e: _foo vs only _)~~ (@AaronO) (not relevant now that all args are optional)
- [x] Allow ops without OpState argument. (@AaronO) https://github.com/denoland/deno/pull/13954
- [x] Infallible / Result-free ops (@AaronO)
- [ ] Less verbose bindings init (@AaronO)
- [ ] Less opcall indirection (@AaronO)
- [ ] Move promise ring to Rust https://github.com/denoland/deno/pull/13906
- [ ] Move unwrapOpResult to Rust (@AaronO)
- [ ] Make [macro generated code](https://github.com/denoland/deno/pull/13861/files#diff-edfa881474570f1c987c98cafcb22b4a45f8861d5f91a642cb7431f51ea42a5cR63-R86) implement a common trait. (@littledivy)
- [ ] Better source file inspection (blocked on rustc) https://github.com/denoland/deno/issues/13939 (@littledivy)
- [ ] Inbuilt permission handling with proc macro attributes. `#[op(unstable)]` (@littledivy)
- [x] Allow sync ops to directly accept `Rc` opstate instead of a mutable borrow across it's lifetime. (@littledivy)

bartlomieju:
@littledivy please update the list

littledivy:
This is mostly done, plus we've got better stuff like fast api and this list is outdated so I'll close this.

PR #13949, opened 2022-03-14
feat(ops): custom arity

AaronO:
Following #13861, part of https://github.com/denoland/deno/issues/13943

### DIFF
--- a/core/examples/hello_world.rs
+++ b/core/examples/hello_world.rs
@@ -17,7 +17,6 @@ use deno_core::*;
 fn op_sum(
   _state: &mut OpState,
   nums: Vec<f64>,
-  _: (),
 ) -> Result<f64, deno_core::error::AnyError> {
   // Sum inputs
   let sum = nums.iter().fold(0.0, |a, v| a + v);
--- a/ops/lib.rs
+++ b/ops/lib.rs
@@ -176,8 +176,8 @@ fn codegen_v8_sync(core: &TokenStream2, f: &syn::ItemFn) -> TokenStream2 {
 fn codegen_args(
   core: &TokenStream2,
   f: &syn::ItemFn,
-  rust_i0: usize,
-  v8_i0: usize,
+  rust_i0: usize, // Index of first generic arg in rust
+  v8_i0: usize,   // Index of first generic arg in v8/js
 ) -> (TokenStream2, TokenStream2) {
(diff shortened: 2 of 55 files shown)


ops/lib.rs
```diff
@@ -169,19 +165,42 @@ fn codegen_v8_sync(core: &TokenStream2, f: &syn::ItemFn) -> TokenStream2 {
...
     #ret
   }
 }
 
+fn codegen_args(
+  core: &TokenStream2,
+  f: &syn::ItemFn,
+  rust_i0: usize,
+  v8_i0: usize,
+) -> (TokenStream2, TokenStream2) {
```
ry, line 181: Can you explain (maybe in a comment) what rust_i0 and v8_i0 do?

Add `assert!(rust_i0 <= v8_i0);` ?

AaronO: > Add `assert!(rust_i0 <= v8_i0);` ?

Nope, `rust_i0 <= v8_i0` won't always be true and likely won't with an upcoming change (depending on the impl path, since there are 2 alternatives).

`rust_i0` and `v8_i0` represent the starting indexes of generic arguments in rust & v8/js respectively

aapoalas:
I guess OpState argument dropping will be done in another PR?

AaronO:
> I guess OpState argument dropping will be done in another PR?

Yep

core/examples/hello_world.rs
```diff
@@ -17,7 +17,6 @@ use deno_core::*;
 fn op_sum(
   _state: &mut OpState,
   nums: Vec<f64>,
-  _: (),
```
ry, line 20: Aren't there more places where this can be removed?

AaronO: Rep but given that we'll also make OpState optional I wanted to do a cleanup PR focused on changing op arity.

In some cases we might want to flatten args and increase arity, in others we'll strip unused args

ry: If done now, it will make it clear this PR is not a net addition of code

AaronO: Did it the last commit, though it makes this PR a mix of feat & cleanup

bartlomieju (approved):
LGTM

ry (approved):
Lgtm

PR #13950, opened 2022-03-14
chore(core,ext): Minor JS optimisations

aapoalas:
Some exceedingly minor JS optimisations:
1. Avoid creating N objects when initialising `opIds` where N is the number of ops. 1 object will do.
2. Avoid a few other unnecessary object spreads, one in `Datagram.prototype.send` and another in a webidl `makeException` call. There are quite a bit more of object spreads around here that I would like to take a stab at, replacing object spreads with explicit option picking but that shall wait for a better day.
3. Remove `opts` parameter usage from webidl `toNumber` helper function as unused.


ext/webidl/00_webidl.js
```diff
@@ -328,7 +328,7 @@
   };
 
   converters["unrestricted float"] = (V, opts) => {
```
kt3k, line 330: This `opts` seems unused now

ext/webidl/00_webidl.js
```diff
@@ -714,7 +714,7 @@
           throw makeException(
             TypeError,
             `can not be converted to '${name}' because '${key}' is required in '${name}'.`,
-            { ...opts },
+            { context: opts.context, prefix: opts.prefix },
```
kt3k, line 717: Does it work to pass simply `opts` instead of creating a new object?

ext/webidl/00_webidl.js
aapoalas: Ah, missed this one. `unrestricted double` already got `opts` turned to `_opts` but this one remains.

ext/webidl/00_webidl.js
aapoalas: It definitely should and there are examples of doing exactly that a few lines earlier, but I'm afraid of making functional changes to the code.

That being said, this whether or not this copying is done cannot be a functional change I don't think. I'll fix this into a simple pass.

bartlomieju:
Please rebase with `main` branch

kt3k (approved):
LGTM

aapoalas:
Merged with main once again.

PR #13951, opened 2022-03-14
fix: upgrade reqwest to 0.11.10

biluohc:
I fixed a bug about proxy for reqwest, in addition to v0.11.10 there is also an other fix.

Pr https://github.com/seanmonstar/reqwest/pull/1491 : fix basic-auth for access http over https's proxy

Changelog: https://github.com/seanmonstar/reqwest/blob/master/CHANGELOG.md#v01110

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1097,7 +1097,7 @@ dependencies = [
  "once_cell",
  "rustls",
  "rustls-native-certs",
- "rustls-pemfile",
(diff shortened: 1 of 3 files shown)


ry (approved):
Unfortunately introduces a duplicated rustls-pemfile dependency. Opened https://github.com/rustls/rustls-native-certs/pull/41 to fix that.

But not a big deal- this can land. Thanks!

PR #13952, opened 2022-03-14
cleanup(core): OpPair => OpDecl

AaronO:
Following https://github.com/denoland/deno/pull/13861, part of https://github.com/denoland/deno/issues/13943


littledivy (commented):
Looks good. Can you please check if there are any performance regressions with this change?

AaronO:
> Looks good. Can you please check if there are any performance regressions with this change?

Why would this impact `op_baseline` in any meaningful way ?

littledivy:
> Why would this impact op_baseline in any meaningful way ?

Oh yeah, it won't. This isn't in the path of op calls.

bartlomieju (approved):

PR #13961, opened 2022-03-15
feat(unstable): add ref/unref to Listener

### LINKED ISSUE #13886: Feature Request: `ref`/`unref` methods for `Deno.Listener`
`net.Server` class in Node.js API has `ref`/`unref` methods which controls `ref` status of tcp server (unref'd server allows the program to exit when the server is still running). https://nodejs.org/api/net.html#serverunref

We need to have similar methods in Deno.Listener to simulate the above Node.js feature in compat mode.

kt3k:
This PR adds `Listener.prototype.ref` and `Listener.prototype.unref` methods to Listener in unstable Deno namespace.

closes #13886


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1296,6 +1296,22 @@ declare namespace Deno {
     alpnProtocols?: string[];
   }
 
+  export interface Listener extends AsyncIterable<Conn> {
+    /** **UNSTABLE**: new API, yet to be vetted.
+     *
+     * Make the listener blocking the event loop from finishing.
```
bartlomieju, line 1302:
```suggestion
     * Make the listener block the event loop from finishing.
```

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1296,6 +1296,22 @@ declare namespace Deno {
...
+     *
+     * Note: the listener blocks the event loop from finishing by default.
+     * This method is only meaningful after `.unref()` is called.
+     */
+    ref(): void;
+    /** **UNSTABLE**: new API, yet to be vetted.
+     *
+     * Make the listener not blocking the event loop from finishing.
```
bartlomieju, line 1310:
```suggestion
     * Make the listener not block the event loop from finishing.
```

cli/tests/unit/net_test.ts
```diff
@@ -807,3 +808,62 @@ Deno.test(
...
+
+Deno.test(
+  { permissions: { net: true } },
+  async function netListenUnref() {
+    const [statusCode, _output] = await execCode(`
+      async function main() {
+        const listener = Deno.listen({ port: 3500 });
+        await listener.accept();
+        listener.unref();
+        await listener.accept(); // The program exits here
```
bartlomieju, line 835: I believe this code is buggy - the second `listener.accept()` should result in an error due to resource being busy:
https://github.com/denoland/deno/blob/b4109375565c6d30d8ffb63f4083e0a430aee201/ext/net/ops.rs#L127-L135

kt3k: I think this 'accept' call is not performed concurrently with the previous one (the previous one is awaited, and it should be finished here).

Added a test case to ensure that concurrent 'accept' calls still throw 'Busy' while listener is unref'd. https://github.com/denoland/deno/pull/13961/commits/003f9bb4c45fe73f4e58ae3ee97084ad4ae29088

cli/tests/unit/net_test.ts
```diff
@@ -807,3 +808,79 @@ Deno.test(
...
+        const listener = Deno.listen({ port: 3500 });
+        listener.unref();
+        listener.ref(); // This restores 'ref' state of listener
+        await listener.accept();
+        console.log("accepted")
+      }
+      main();
+    `);
+    // TODO(kt3k): This is racy. Find a correct way to
+    // wait for the server to be ready
```
bartlomieju, line 860: I don't think of a better way to do it when you are spawning a subprocess. Maybe increase the timeout to 1s?

bartlomieju (approved):
LGTM

PR #13966, opened 2022-03-15
feat(task): add unstable warning to `deno task`

dsherret:
Moving `deno task` to unstable. This will allow us to validate the approach before stabilizing it.


cli/tests/testdata/task/task_exit_code_5.out
```diff
@@ -1,2 +1,3 @@
-Task exit_code_5 echo $(echo 10 ; exit 2) && exit 5 
+Warning deno task is unstable and may drastically change in the future
```
dsherret, line 1: Most of the tests use the `--quiet` flag, so we don't need to update them.

bartlomieju (approved):

PR #13969, opened 2022-03-15
Allow Response status 101

ry:
Necessary for https://github.com/denoland/deno/pull/13618

### DIFF
--- a/cli/tests/unit/fetch_test.ts
+++ b/cli/tests/unit/fetch_test.ts
@@ -970,7 +970,7 @@ Deno.test(function fetchResponseConstructorNullBody() {
 });
 
 Deno.test(function fetchResponseConstructorInvalidStatus() {
-  const invalidStatus = [101, 600, 199, null, "", NaN];
+  const invalidStatus = [600, null, "", NaN];
 
   for (const status of invalidStatus) {
     try {
@@ -980,7 +980,7 @@ Deno.test(function fetchResponseConstructorInvalidStatus() {
       fail(`Invalid status: ${status}`);
     } catch (e) {
       assert(e instanceof RangeError);
-      assert(e.message.endsWith("is outside the range [200, 599]."));
+      assert(e.message.endsWith("is outside the range [100, 599]."));
     }
   }
 });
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):

piscisaureus:
Fails the following WPT tests:

```
  "/fetch/api/response/response-error.any.html - Throws RangeError when responseInit's status is 100"
  "/fetch/api/response/response-error.any.html - Throws RangeError when responseInit's status is 199"      
  "/fetch/api/response/response-error.any.worker.html - Throws RangeError when responseInit's status is 100"
  "/fetch/api/response/response-error.any.worker.html - Throws RangeError when responseInit's status is 199"
```

If we cared for marketing reasons to pass as much WPT tests as possible, then we could just allow status code 101 or 101-198.

I think there's also a better argument to be made for allowing *only* 101:
* status codes 100/102/103 are useless in our current model, because these status codes indicate that the server is going to send *another* response *in addition to* the 1xx response, but in our API there is no support for sending multiple `Response`s for a single `Request`.
* status codes 104-199 don't exist

bartlomieju:
> Fails the following WPT tests:
> 
> ```
>   "/fetch/api/response/response-error.any.html - Throws RangeError when responseInit's status is 100"
>   "/fetch/api/response/response-error.any.html - Throws RangeError when responseInit's status is 199"      
>   "/fetch/api/response/response-error.any.worker.html - Throws RangeError when responseInit's status is 100"
>   "/fetch/api/response/response-error.any.worker.html - Throws RangeError when responseInit's status is 199"
> ```
> 
> If we cared for marketing reasons to pass as much WPT tests as possible, then we could just allow status code 101 or 101-198.
> 
> I think there's also a better argument to be made for allowing _only_ 101:
> 
> * status codes 100/102/103 are useless in our current model, because these status codes indicate that the server is going to send _another_ response _in addition to_ the 1xx response, but in our API there is no support for sending multiple `Response`s for a single `Request`.
> * status codes 104-199 don't exist

Good point, I think allowing only 101 is a good compromise, but we should update info in MDN compat table too (CC @lucacasonato )

lucacasonato:
I need to review the implications of this in regards to spec.

piscisaureus (approved):

bartlomieju:
This PR allows to ship `Deno.upgradeHttp` https://github.com/denoland/deno/pull/13618

PR #13970, opened 2022-03-15
fix(core): nuke Deno.core.ops pre-snapshot

AaronO:
To avoid OOB & other ExternalReference snapshot serialization issues

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -26,8 +26,11 @@
     ObjectAssign,
     SymbolFor,
   } = window.__bootstrap.primordials;
-  const ops = window.Deno.core.ops;
-  const opIds = Object.keys(ops).reduce((a, v, i) => ({ ...a, [v]: i }), {});
+  const core_ = window.Deno.core;
+  const opIds = Object.keys(core_.ops).reduce(
+    (a, v, i) => ({ ...a, [v]: i }),
+    {},
+  );
 
   // Available on start due to bindings.
   const { refOp_, unrefOp_ } = window.Deno.core;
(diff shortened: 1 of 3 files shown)


core/01_core.js
```diff
@@ -26,8 +26,11 @@
     ObjectAssign,
     SymbolFor,
   } = window.__bootstrap.primordials;
-  const ops = window.Deno.core.ops;
-  const opIds = Object.keys(ops).reduce((a, v, i) => ({ ...a, [v]: i }), {});
+  const core_ = window.Deno.core;
```
piscisaureus, line 29: Looks like the changes to 01_core.js are not necessary (since the objs object itself is cleared rather than recreated).

core/runtime.rs
```diff
@@ -148,6 +148,7 @@ pub(crate) struct JsRuntimeState {
   pub(crate) js_nexttick_cbs: Vec<v8::Global<v8::Function>>,
   pub(crate) js_promise_reject_cb: Option<v8::Global<v8::Function>>,
   pub(crate) js_uncaught_exception_cb: Option<v8::Global<v8::Function>>,
+  pub(crate) js_ops_obj: Option<v8::Global<v8::Object>>,
```
piscisaureus, line 151: Is it really necessary to store this in the state? It's only read in one place (snapshot), you could  just call `grab_obj(..., "Deno.core.ops)` there.

AaronO: That's what I originally did more or less and it was wonky, I would just stick with this for now, no practical downside

piscisaureus: “¯\_(ツ)_/¯“ https://github.com/denoland/deno/pull/13970/commits/fd11830e9e1d1a5039c60f59d64fc4da02743f5c

piscisaureus (approved):
Nice work!

PR #13972, opened 2022-03-15
cleanup(core): grab_js helper

AaronO:


core/runtime.rs
```diff
@@ -505,34 +505,25 @@ impl JsRuntime {
...
+  where
+    v8::Local<'s, T>: TryFrom<v8::Local<'s, v8::Value>, Error = v8::DataError>,
+  {
     let code = v8::String::new(scope, code).unwrap();
     let script = v8::Script::compile(scope, code, None).unwrap();
     let v8_value = script.run(scope).unwrap();
-    let obj = v8::Local::<v8::Object>::try_from(v8_value).unwrap();
-    scope.escape(obj)
+    let v = v8::Local::<'s, T>::try_from(v8_value).unwrap();
+    v8::Global::new(scope, v)
```
piscisaureus, line 520: With only two callers, this hardly matters, but...
I would suggest to return a Local, because that's what callers usually need. Skip the extra work to create a Global. The caller can convert to Global easily.

core/runtime.rs
```diff
@@ -505,34 +505,25 @@ impl JsRuntime {
...
     scope: &mut v8::HandleScope<'s>,
     code: &str,
-  ) -> v8::Local<'s, v8::Object> {
-    let scope = &mut v8::EscapableHandleScope::new(scope);
+  ) -> v8::Global<T>
+  where
+    v8::Local<'s, T>: TryFrom<v8::Local<'s, v8::Value>, Error = v8::DataError>,
+  {
     let code = v8::String::new(scope, code).unwrap();
     let script = v8::Script::compile(scope, code, None).unwrap();
```
piscisaureus, line 517: With only two callers, this hardly matters, but...

Seems a bit excessive to compile a script for this.
A generic-enough alternative might be 
```rust
fn get_deno_core_property<T>(
  scope: &mut v8::HandleScope<'s>, 
  name: &str // e.g. "ops"
) -> v8::Local<'s, T>
where
    v8::Local<'s, T>: TryFrom<v8::Local<'s, v8::Value>, Error = v8::DataError>,
{ 
  // ... 
}
```

piscisaureus:
Another place where this would be useful is here: https://github.com/denoland/deno/blob/5d60ee7f1280d8000b918e7fb9e11e3ddca779e7/core/bindings.rs#L165-L182

core/runtime.rs
AaronO: Yeah, it's not called in hot paths so I've been valuing convenience over function.

A get or get+ensure & list of attr names would makes sense

piscisaureus (approved):

bartlomieju (approved):

AaronO:
Meh, the helpers kind of feel overkill. I could see the recursive get being a nice helper, recursive obj ensure is probably a bit too niche ...

bartlomieju (approved):

littledivy (approved):
clean :shipit:

PR #13978, opened 2022-03-16
feat(lsp): support deno.enablePaths setting

kitsonk:
Ref: denoland/vscode_deno#633

This is functionally working, just need to write tests, but this branch plus https://github.com/denoland/vscode_deno/pull/635 should allow folks to take it for a test spin. @dsherret especially if you could give it a spin on Windows.  I think I have done everything to make it work well, but windows always has surprises.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -2583,12 +2583,21 @@ impl lspower::LanguageServer for LanguageServer {
     &self,
     params: DidChangeWorkspaceFoldersParams,
   ) {
-    self
-      .0
-      .lock()
-      .await
-      .did_change_workspace_folders(params)
-      .await
+    let client = {
+      let mut inner = self.0.lock().await;
+      inner.did_change_workspace_folders(params).await;
+      inner.client.clone()
+    };
+    let language_server = self.clone();
+    tokio::spawn(async move {
+      language_server
+        .0
+        .lock()
+        .await
+        .config
+        .update_enabled_paths(client)
+        .await;
+    });
   }
 
   async fn document_symbol(
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -117,8 +117,6 @@ pub(crate) struct Inner {
   maybe_import_map_uri: Option<Url>,
   /// A collection of measurements which instrument that performance of the LSP.
   performance: Arc<Performance>,
-  /// Root provided by the initialization parameters.
-  root_uri: Option<Url>,
   /// A memoized version of fixable diagnostic codes retrieved from TypeScript.
   ts_fixable_diagnostics: Vec<String>,
   /// An abstraction that handles interactions with TypeScript.
@@ -173,7 +171,6 @@ impl Inner {
       maybe_import_map_uri: None,
       module_registries,
       module_registries_location,
-      root_uri: None,
       performance,
       ts_fixable_diagnostics: Default::default(),
       ts_server,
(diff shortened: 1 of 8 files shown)


cli/lsp/language_server.rs
```diff
@@ -117,8 +117,6 @@ pub(crate) struct Inner {
   maybe_import_map_uri: Option<Url>,
   /// A collection of measurements which instrument that performance of the LSP.
   performance: Arc<Performance>,
-  /// Root provided by the initialization parameters.
-  root_uri: Option<Url>,
```
kitsonk, line 121: I moved it to `.config` because that is really where it belongs.

cli/lsp/language_server.rs
```diff
@@ -698,6 +697,12 @@ impl Inner {
           LspError::internal_error()
         })?;
       }
+      self.config.workspace_folders = params.workspace_folders.map(|folders| {
```
kitsonk, line 700: This was never actually set, though nothing actually consumed it until now. 🤦

cli/lsp/language_server.rs
```diff
@@ -774,6 +779,7 @@ impl Inner {
         warn!("Client errored on capabilities.\n{}", err);
       }
     }
+    self.config.update_enabled_paths(self.client.clone()).await;
```
kitsonk, line 782: This needs to occur after initialized, because the client doesn't actually respond to configuration requests if done earlier.

dsherret (commented):
Overall looks good to me other than making a call back to the client while holding the "lsp mutex". I think we just need to change that and we should be good.

I've yet to try this out on Windows, but will soon.

cli/lsp/language_server.rs
kitsonk: Here it only ever occurs once, when the client has said it has initialised.  This is exactly the point in the lifecycle where you expect to have "blocking" requests, where the server queries the client of anything that falls outside of the initialisation before handling other messages.

I believe there are two other instances, one (on config change) occurs already outside the inner mutex lock (like re-querying the specifiers).  I will double check that the other (when workspace folders are updated) follows that pattern.

dsherret (changes requested):
So I tried this out today and it works really well, but one thing we need to do is reset all diagnostics when this setting changes.

cli/lsp/language_server.rs
dsherret: That's fair.

kitsonk:
Hmm, actually the problem would be that we aren't emptying diagnostics for newly undiagnosable resources. I think I know what we can do.

kitsonk:
Ok, it doesn't actually look like this problem is related to this change.  I am tracing it through, but it seems for some reason when the config changes, the diagnostics are being updated in the server and the disabled diagnostics are being "emptied", but for some reason in the publishing it doesn't look like we are handling it.  Trying to narrow it down.

kitsonk:
Ah, it seems that when the diagnostics are being re-calculated after the change, for some reasons, the enablement/disablement of a specifier reflects the previous state, not the updated state.  So it looks like the update to the workspace folders is occurring post the snapshotting for the diagnostics.

kitsonk:
Ok, fixed it. It appears we already had an edge case when using workspace folders and changing the configuration that might have enabled or disabled a workspace folder, it wouldn't update the diagnostics until a document was touched in the workspace.

dsherret (approved):
LGTM! I actually came across a scenario today where this feature would be nice, so looking forward to it 🙂

kitsonk:
> LGTM! I actually came across a scenario today where this feature would be nice, so looking forward to it 🙂

Yeah, though I have been resistant too it, mostly because it only occurred to me previously of "glob" support, and that would be very difficult, this path support is super useful and should be fairly "fool proof".

PR #13981, opened 2022-03-16
perf(web): optimize Blob.text and Blob.arrayBuffer

littledivy:
```
blob_text_large:     	n = 100, dt = 0.655s, r = 153/s, t = 6550000ns/op
```

### DIFF
--- a/cli/bench/deno_common.js
+++ b/cli/bench/deno_common.js
@@ -58,7 +58,7 @@ function benchUrlParse() {
 
 function benchLargeBlobText() {
   const input = "long-string".repeat(999_999);
-  benchSync("blob_text_large", 3, () => {
+  benchSync("blob_text_large", 1e2, () => {
     new Blob([input]).text();
   });
 }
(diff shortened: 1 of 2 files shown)


cli/bench/deno_common.js
```diff
@@ -58,7 +58,7 @@ function benchUrlParse() {
 
 function benchLargeBlobText() {
   const input = "long-string".repeat(999_999);
-  benchSync("blob_text_large", 3, () => {
+  benchSync("blob_text_large", 1e2, () => {
```
ry, line 61: “1e2” is written more clearly as “100”

cli/bench/deno_common.js
```diff
@@ -39,7 +39,7 @@ function benchStats(name, n, t1, t2) {
 
 function benchB64RtLong() {
   const input = "long-string".repeat(99999);
-  benchSync("b64_rt_long", 1e2, () => {
+  benchSync("b64_rt_long", 100, () => {
```
bartlomieju, line 42: I'd be great to rewrite these to use new `Deno.bench()` API ;)

bartlomieju (approved):

PR #13984, opened 2022-03-16
chore: uploading WPT shouldn't make CI fail

ry:
We had two successive main branch failures related to this.
- https://github.com/denoland/deno/runs/5566541876?check_suite_focus=true
- https://github.com/denoland/deno/runs/5566531944?check_suite_focus=true


bartlomieju (approved):
LGTM

PR #13987, opened 2022-03-16
feat(ext/net): Deno.upgradeHttp handles unix connections

bartlomieju:

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -5,7 +5,7 @@ import {
   BufWriter,
 } from "../../../test_util/std/io/buffer.ts";
 import { TextProtoReader } from "../../../test_util/std/textproto/mod.ts";
-import { serve } from "../../../test_util/std/http/server.ts";
+import { serve, serveTls } from "../../../test_util/std/http/server.ts";
 import {
   assert,
   assertEquals,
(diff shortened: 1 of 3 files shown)


cli/tests/unit/http_test.ts
```diff
@@ -1782,6 +1782,120 @@ Deno.test("upgradeHttp", async () => {
...
+        const secondPacketText = new TextDecoder().decode(buf.slice(0, n));
+        assertEquals(secondPacketText, "bla bla bla\nbla bla\nbla\n");
+        conn.close();
+      })();
+
+      const resp = new Response(null, { status: 101 });
+      await respondWith(resp);
+      await promise;
+      httpConn.close();
+      break;
```
ry, line 1892: Seems weird to have a for await loop that breaks after one iteration.

There's a lot of sloppy loops in the tests that need to be unrolled and simplified - I expect the line-of-code could be halved.

ry (approved):
LGTM

cli/tests/unit/http_test.ts
bartlomieju: Opened https://github.com/denoland/deno/issues/13988

KnorpelSenf:
I believe this PR to be causing #14437. Is that possible?

bartlomieju:
> I believe this PR to be causing #14437. Is that possible?

This change is unrelated to the `fetch` API.

PR #13989, opened 2022-03-16
fix(ext/ffi): check for permissions in UnsafeFnPointer#call

DjDeveloperr:
UnsafeFnPointer calls did not check for FFI permission.

### DIFF
--- a/ext/ffi/lib.rs
+++ b/ext/ffi/lib.rs
@@ -144,8 +144,8 @@ pub fn init<P: FfiPermissions + 'static>(unstable: bool) -> Extension {
       op_ffi_get_static::decl(),
       op_ffi_call::decl(),
       op_ffi_call_nonblocking::decl(),
-      op_ffi_call_ptr::decl(),
-      op_ffi_call_ptr_nonblocking::decl(),
+      op_ffi_call_ptr::decl::<P>(),
+      op_ffi_call_ptr_nonblocking::decl::<P>(),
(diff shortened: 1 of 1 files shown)


piscisaureus (approved):

piscisaureus (changes requested):
Thanks @DjDeveloperr, the patch looks good.
Would it be possible to add a test for this?

aapoalas:
I would like to challenge this PR:

Imagine an FFI library that has been loaded for the purpose of using a single API call. That API call will return a function pointer that should be later called periodically.

To guard the safety of the environment running this code, the library user revokes the FFI permission after loading the library. This way no malicious foreign code (dependency or other) can use the FFI permission to load eg. their custom DLL payload (if they had rw rights to some folder) or just load some well-known DLL that offers more APIs they need to further their attack.

Now, if the FFI permission is needed to call the received function pointer, it means that the permission cannot be revoked and still keep the code running. Unfortunately this same issue currently applies to `UnsafePointerView` reads as well, which are likely to be necessary for some FFI work.

I feel that the FFI permission may need to be more granular (load vs use), or otherwise somehow rethought to allow "load, revoke, use" sort of flow.

DjDeveloperr:
I think that security model is pretty hard to implement in reality, loading a library vs using pointer APIs are both very unsafe on their own. Even if we are to add it, that would be for another PR.

aapoalas:
After some discussion on Discord, I heartily support this PR.

PR #13993, opened 2022-03-16
fix: cargo publish fails without absolute paths

ry:
https://github.com/denoland/deno/commit/4e3ed37037a2aa1edeac260dc3463a81d9cf9b88 was erroneous and broke `cargo publish`.  I've opened https://github.com/denoland/deno/issues/13994 that would allow us to catch errors like this in the future.

1.20.1 will be released after this is landed.


bartlomieju (approved):
LGTM, sorry for the trouble

PR #13995, opened 2022-03-17
build: use workflows for bumping versions and cargo publishing on the CI

### LINKED ISSUE #13874: Publish crates to crates.io on CI
We could create a publish workflow that is specifically for publishing the crates to crates.io (separate from the workflow that creates a draft PR).

This could handle failure gracefully similar to the current publish CI script (it could just use that same script).

https://github.com/denoland/deno/blob/main/tools/cut_a_release.md,

### LINKED ISSUE #13994: CI should do `cargo publish --dry-run`
Doing it for each of the crates sounds very expensive (but necessary to catch all errors, I think).

We should at minimum do it for the CLI crate...

dsherret:
See tools/cut_a_release.md for how this would work.

Essentially:

1. User kicks off "version_bump" workflow on branch that they want to do a release from.
    image
2. A PR gets opened automatically, see https://github.com/dsherret/deno/pull/3 for an example (Note: the actual changes in there are messed up because I did a force push to my branch, but you can see from the PR body how it will look).
3. The person doing the release updates the PR if necessary and merges it.
4. After merge they kick off the "cargo_publish" workflow, which will publish all the crates and tag the repo which will kick off the release workflow which creates the draft.

Closes #13874


.github/workflows/cargo_publish.yml
```diff
@@ -0,0 +1,42 @@
+name: cargo_publish
```
dsherret, line 1: I tested version_bump.yml a lot, but I did not test this workflow because I didn't want to test crates. Please review it carefully. If this fails during a release, we can easily execute this stuff manually (I should do the first release with this)

.github/workflows/version_bump.yml
```diff
@@ -0,0 +1,62 @@
+name: version_bump
```
dsherret, line 1: This workflow takes under 2 minutes to run.

tools/release/01_bump_crate_versions.ts
```diff
@@ -101,3 +112,26 @@ async function getGitLog() {
...
+  const text = Deno.readTextFileSync(compatFilePath);
+  Deno.writeTextFileSync(
+    compatFilePath,
+    text.replace(/std@[0-9]+\.[0-9]+\.[0-9]+/, `std@${newStdVersion}`),
+  );
+}
+
+async function getLatestStdVersion() {
+  const url =
+    "https://raw.githubusercontent.com/denoland/deno_std/main/version.ts";
```
ry, line 128: Not for now, but we should maybe rename this to `version.json`, then you wouldn't need the regex.

ry (approved):
LGTM - eager to see how this works!

tools/cut_a_release.md
```diff
@@ -64,63 +64,61 @@ previous release and start cherry-picking newer commits from the `main`.
...
 
-1. Update your local branch (`v1.XX` for patch or `main` for minor) and checkout
-   another branch (e.g. `release_#.#.#`).
+### Phase 1: Bumping versions
 
-2. Run `./tools/release/01_bump_crate_versions.ts` to increase the versions of
-   all crates including the CLI. If you are doing a CLI patch release, answer
-   `y` to the _Increment patch?_ prompt.
+1. After releasing deno_std, go to "Actions" on the GitHub repo and select the
+   "version_bump" workflow.
```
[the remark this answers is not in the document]
dsherret: Good idea.

tools/release/01_bump_crate_versions.ts
```diff
@@ -8,16 +8,27 @@ const repo = workspace.repo;
...
 // increment the cli version
-await cliCrate.promptAndIncrement();
+if (Deno.args.some((a) => a === "--patch")) {
+  await cliCrate.increment("patch");
+} else if (Deno.args.some((a) => a === "--minor")) {
+  await cliCrate.increment("minor");
+} else if (Deno.args.some((a) => a === "--major")) {
+  await cliCrate.increment("major");
+} else {
+  await cliCrate.promptAndIncrement();
```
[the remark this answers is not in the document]
dsherret: I like this prompt. It's for when running it manually instead of on the CI

PR #13997, opened 2022-03-17
fix build

ry:
- fix node_resolver
- fix warning

### DIFF
--- a/cli/tools/lint.rs
+++ b/cli/tools/lint.rs
@@ -550,7 +550,7 @@ pub(crate) fn get_configured_rules(
   );
 
   if configured_rules.is_empty() {
-    anyhow!("No rules have been configured");
+    return Err(anyhow!("No rules have been configured"));
   }
 
   Ok(configured_rules)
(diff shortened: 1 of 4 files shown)


cli/tools/lint.rs
```diff
@@ -550,7 +550,7 @@ pub(crate) fn get_configured_rules(
   );
 
   if configured_rules.is_empty() {
-    anyhow!("No rules have been configured");
+    return Err(anyhow!("No rules have been configured"));
```
dsherret, line 553: Interesting that this wasn't a clippy failure or something.

dsherret (approved):
LGTM

cli/tools/lint.rs
ry: I got a warning locally, but it seems we're not failing on warnings for some reason. I thought this should be doing it: 

https://github.com/denoland/deno/blob/92bb25105822d0def65aa89ee5141c862ac7a3e4/.github/workflows/ci.yml#L150-L153

dsherret: I think maybe it should be `echo "RUSTFLAGS='-D warnings'" >> $GITHUB_ENV`

ry: I opened https://github.com/denoland/deno/issues/13998

PR #13925, opened 2022-03-12
fix(fetch): Fix uncaught rejection panic with `WebAssembly.instantiateStreaming`

andreubotella:
When an exception is thrown during the processing of streaming WebAssembly, `op_wasm_streaming_abort` is called. This op calls into V8, which synchronously rejects the promise and calls into the promise rejection handler, if applicable. But calling an op borrows the isolate's `JsRuntimeState` for the duration of the op, which means it is borrowed when V8 calls into `promise_reject_callback`, which tries to borrow it again, panicking.

This change changes `op_wasm_streaming_abort` from an op to a binding (`Deno.core.abortWasmStreaming`). Although that binding must borrow the `JsRuntimeState` in order to access the `WasmStreamingResource` stored in the `OpTable`, it also takes ownership of that `WasmStreamingResource` instance, which means it can drop any borrows of the `JsRuntimeState` before calling into V8.

Fixes #13917.

Issue #13917, opened 2022-03-11
`WebAssembly.instantiateStreaming` panics when given a `URL` object


aslilac:
```
➜  ~ deno                   
Deno 1.19.3
exit using ctrl+d or close()
> WebAssembly.instantiateStreaming(new URL("./index.wasm", `file://${Deno.cwd()}`), {})

============================================================
Deno has panicked. This is a bug in Deno. Please report this
at https://github.com/denoland/deno/issues/new.
If you can reliably reproduce this panic, include the
reproduction steps and re-run with the RUST_BACKTRACE=1 env
var set and include the backtrace in your report.

Platform: macos x86_64
Version: 1.19.3
Args: ["deno"]

thread 'main' panicked at 'already borrowed: BorrowMutError', core/bindings.rs:389:28
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: deno_core::bindings::promise_reject_callback
   4: __ZN2v88internal7Isolate19ReportPromiseRejectENS0_6HandleINS0_9JSPromiseEEENS2_INS0_6ObjectEEENS_18PromiseRejectEventE
   5: __ZN2v88internal9JSPromise6RejectENS0_6HandleIS1_EENS2_INS0_6ObjectEEEb
   6: __ZN2v812_GLOBAL__N_112_GLOBAL__N_137AsyncInstantiateCompileResultResolver19OnCompilationFailedENS_8internal6HandleINS3_6ObjectEEE
   7: __ZN2v813WasmStreaming5AbortENS_10MaybeLocalINS_5ValueEEE
   8: deno_core::ops_json::op_sync::{{closure}}
   9: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
  10: deno_core::ops::OpTable::route_op
  11: <extern "C" fn(A0) .> R as v8::support::CFnFrom<F>>::mapping::c_fn
  12: __ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
  13: __ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEESA_NS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EENS0_16BuiltinArgumentsE
[...]

PR #14000, opened 2022-03-17
chore(wasm): Don't await on the argument to `handleWasmStreaming`

andreubotella:
`handleWasmStreaming` is the function that provides the binding with the `fetch` API needed for `WebAssembly.instantiateStreaming()` and `WebAssembly.compileStreaming()`. When I implemented it in #11200, I thought V8 was calling these functions with the argument of the `WebAssembly` streaming functions, without doing any resolving, and so `handleWasmStreaming` awaits for the parameter to resolve. However, as discovered in https://github.com/denoland/deno/issues/13917#issuecomment-1065805565, V8 does in fact resolve the parameter if it's a promise (and handles rejections arising from that).

This change removes the `async` IIFE inside `handleWasmStreaming`, letting initial errors be handled synchronously (which will however not throw synchronously from the `WebAssembly` namespace functions). Awaiting is still necessary for reading the bytes of the response, though, and so there is an `async` IIFE for that.

----------------------

This PR builds on #13925, not because the changes depend on those of that PR, but because this PR changes the stack trace in one of that PR's test outputs.


bartlomieju:
@andreubotella please rebase

lucacasonato (approved):
LGTM, thanks @andreubotella!

Issue #3635, opened 2020-01-09
snapshot testing


brandonkal:
A nice to have feature would be Jest-like snapshot testing in the std/testing library.

ry:
I don't know what that means... can you explain more?

bartlomieju:
Ref https://github.com/denoland/deno_std/issues/346

brandonkal:
@ry Basically when you call `expect(x).toMatchSnapshot()` it serializes whatever x is and stores it in a state file. Here's [an example](https://github.com/brandonkal/linaria/blob/fork/src/__tests__/__snapshots__/babel.test.tsx.snap).

The premise is that you take a snapshot of the state when things are how they should be, and if they later differ, the test fails. You also have the option to update the snapshot if you desire the change or to define a custom serializer.

brandonkal:
Also the option to run a specific test by name (filter) would be nice.

KSXGitHub:
What do you think the API should look like?

I propose this:

```javascript
import { matchSnapshot } from 'https://deno.land/std/testing/snapshot.ts'
matchSnapshot('filename.snapshot', myObject)
```

```sh
deno test # would fail if snapshot is outdated
deno test -u # would update snapshot if it is outdated
```

brandonkal:
I would propose the API be closer to Jest's `.toMatchSnapshot()`. So there is no need to specify where to store the snapshot.

A similar structure would be implied
```
my_test.ts
__snapshots__
|- my_test.ts.denosnap
```

KSXGitHub:
How would a `std` lib knows where the test file or snapshot file is located? The only way to know is for the test file to provide `import.meta.url`, which is just snapshot location with extra steps.

brandonkal:
`import.meta.url` gives you the filepath of the test file. From there you easily convert it to containing folder + `__snapshots__/filename.denosnap`

You could also do something like:
```ts
import { makeSnapshotter } from 'https://deno.land/std/testing/snapshot.ts'
const snap = makeSnapshotter('./__snapshots__')
snap(myObject)
```

KSXGitHub:
@brandonkal That's what I said: "snapshot location with extra steps".

> `import.meta.url` gives you the filepath of the test file

[...]

PR #2039, opened 2019-04-02
Non-fatal compile_sync failures

afinch7:
Handle compiler worker failures, so the worker can have multiple failures without exiting.

PR #14007, opened 2022-03-17
feat(test): Add "name", "origin" and "parent" to "Deno.TestContext"

hyp3rflow:
This PR includes features for implementing snapshot test feature (assertSnapshot) in `deno_std`.
- Exposes the current test name.
- TestContext includes their parentContext. (.parent)
- Add origin (I think we can use this instead of `import.meta.url` or `Deno.mainModule` to get the path of snapshot).  
- ~~Teardown API to call function after the test finished.~~
- ~~Add update flag on cli/test for snapshot update mode.~~

related comment is here: https://github.com/denoland/deno/issues/3635#issuecomment-1069149310

I wrote simple test for this code, but I am not sure where I have to put this test file. please let me know where to put and how to run this test in the project structure. thanks!
```typescript
Deno.test('test name', (t) => {
	t.name; // This will be `test name`
	t.origin; // `file:///Users/me/deno.test.ts`

	t.step('test name2', (t2) => {
		t2.name; // This will be `test name2`
		t2.origin; // `file:///Users/me/deno.test.ts`
	})
})
```


cli/dts/lib.deno.ns.d.ts
```diff
@@ -250,6 +250,8 @@ declare namespace Deno {
   }
 
   export interface TestContext {
+    name: string;
```
dsherret, line 253: We should add a jsdoc that explains this is the current test name.

dsherret (approved):
LGTM! Seems reasonable to add this

hyp3rflow:
@dsherret I updated the PR but it was delayed due to github status issue. sorry 😢 
If you think the PR has to be separated, feel free to comment me :)

cli/dts/lib.deno.ns.d.ts
```diff
@@ -250,6 +250,20 @@ declare namespace Deno {
   }
 
   export interface TestContext {
+    /**
+     * The full name of the current test.
+     */
+    name: string;
```
dsherret, line 256: I think this should probably just be the current test/step name. Then if someone wants them all they would be able to concatenate it themselves? Or perhaps we should have a `parent` property that gets the parent test context, then to get the full name someone could walk the parents.

cli/flags.rs
```diff
@@ -1375,6 +1376,14 @@ fn test_subcommand<'a>() -> App<'a> {
...
         .value_hint(ValueHint::AnyPath),
     )
+    .arg(
+      Arg::new("update")
+        .long("update")
+        .short('u')
+        .help("Enable snapshot update mode")
+        .takes_value(false)
+        .required(false),
+    )
```
bartlomieju, line 1386: I'm against adding this flag right now; it's unclear ATM how to snapshotting functionality will work. We should first have an MVP in `deno_std` before we start adding new flags to `deno test`

bartlomieju (changes requested):

dsherret (commented):
The new `name` property seems nice, but I'm not sure this snapshotting feature should be in the CLI. It seems like something a testing framework (or deno_std) could implement themselves. cc @bartlomieju

kitsonk:
> but I'm not sure this snapshotting feature should be in the CLI

I also agree that the snapshotting feature should be part of CLI.

bartlomieju:
> > but I'm not sure this snapshotting feature should be in the CLI
> 
> I also agree that the snapshotting feature should be part of CLI.

@kitsonk I'm confused. Are you for adding this functionality to CLI directly or against?

kitsonk:
> > > but I'm not sure this snapshotting feature should be in the CLI
> > 
> > 
> > I also agree that the snapshotting feature should be part of CLI.
> 
> @kitsonk I'm confused. Are you for adding this functionality to CLI directly or against?

Not. Sorry... edited previous comment.

hyp3rflow:
then is it okay to update the snapshot with the argument, not cli flags?
```sh
deno test --allow-read --allow-write ./snapshot.test.ts -- --update
```
and check `Deno.args` in `assertSnapshots`

hyp3rflow:
I tidy up this PR with recent commits.
- `TestContext.name`: the current test name.
- `TestContext.parent`: their parentContext. (.parent)
- `TestContext.origin`: the file uri of the current test
- `Deno.test.teardown(fn)`: to register function to be called after the test done

runtime/js/40_testing.js
```diff
@@ -1408,6 +1424,12 @@
     return value == null ? defaultValue : value;
   }
 
+  let teardownFn;
```
bcheidemann, line 1427: Do we need to consider that `test.teardown` might be called multiple times? This could lead to some very unexpected and hard to debug behaviour with the current implementation.

bcheidemann (changes requested):

runtime/js/40_testing.js
hyp3rflow: I think we need to consider that case. Is there any good idea for handling that?

hyp3rflow: how about this? push `teardownFn` and call sequentially after the test. 
```typescript
const teardownFns = [];
function teardown(fn) {
	if(!teardownFns.includes(fn)) teardownFns.push(fn);
}

// after the test
for await (const fn of teardownFns) {
	await fn();
}
```
but I'm not sure this is good about the order of calls.
To handle the order problem, we can take another approach.
```typescript
let teardownFn = () => {};
function teardown(fn) {
	teardownFn = fn(teardownFn);
}

// example of fn
function exampleFn(alreadyRegisteredFn) {
	return async function () {
		// await alreadyRegisteredFn();
		doAwesomeStuff();
		// await alreadyRegisteredFn(); <- we can choose the call order of `teardownFns`.
	}
}
```

bcheidemann: This is a good point!

I think the second example you gave could be improved in the following ways:

 1. It is hard to understand the order of calls because developers must check each `teardown` function to know what the order is.
 2. It is prone to errors e.g. a developer forgets to call the previous `teardown` function or calls it twice

The issue is that each test has a degree of control over the order of `teardown` calls, essentially democratising the order. This approach probably makes problem `#1` unsolvable.

How important is it that developers have control over the order of calls? If fine grained control over the order is important, then it will be possible to implement a custom solution:

```ts
function doTeardownOperationsInSomeOrder() {
  // --snip--
}

Deno.test("name", (test) => {
  test.teardown(doTeardownOperationsInSomeOrder);

  test.step("stepName", (test) => {
    // Custom solution to register teardown operations to be called by `doTeardownOperationsInSomeOrder`
  }

  // --snip--
});
```

However, I believe that in 99% of cases this problem would be solved by Deno exposing the following functions in the test context:

```ts
Deno.test("test", (test) => {
  test.beforeAll(() => { /* --snip-- */ });
  test.beforeEach(() => { /* --snip-- */ });
  test.afterAll(() => { /* --snip-- */ });
  test.afterEach(() => { /* --snip-- */ });
});
```

These would work similarly to the [Jest Setup and Teardown functions](https://jestjs.io/docs/setup-teardown) and allow for a decent level of control over the order while still being easy to understand.

To summarise, I propose the following:

**test.beforeAll** - _[NOT IN THIS PR]_ execute a function before any child steps are executed
**test.afterAll** - _[NOT IN THIS PR]_ execute a function after any child steps are executed
**test.beforeEach** - _[NOT IN THIS PR]_ execute a function before each child steps is executed
**test.afterEach** - _[NOT IN THIS PR]_ execute a function after each child steps is executed
**test.teardown** - _[THIS PR]_ execute a function after the top level parent test has finished executing and all descendant test steps have finished executing

What do you think @hyp3rflow?

bartlomieju:
> Deno.test.teardown(fn): to register function to be called after the test done

@hyp3rflow why do you need this API? It seems cleanup can be done without this API. It's been considered several times and we decided not to do it, so I'm quite hesitant ATM.

hyp3rflow:
@bartlomieju Is there any way to cleanup after tests are done? I cannot find the way to writing snapshot after tests without the `teardown` API.

bcheidemann:
> > Deno.test.teardown(fn): to register function to be called after the test done
> 
> @hyp3rflow why do you need this API? It seems cleanup can be done without this API. It's been considered several times and we decided not to do it, so I'm quite hesitant ATM.

@bartlomieju 

Hope you don't mind me responding to this @hyp3rflow - feel free to add to/correct my reply as necessary :D

As I understand it, the `teardown` function is introduced here for performance reasons. It is common to have many snapshots in a single test file and without the `teardown` function, it would be impossible to implement the `assertSnapshot` function without opening, reading, parsing and writing the file on every single call to `assertSnapshot`. This may be negligible on a small code base but in a mature code base with hundreds of snapshots this would have a significant performance impact.

To illustrate this, I implemented a basic test to compare the performance of using a `teardown` function to not using a `teardown` function. This is the code:

```ts
const NUMBER_OF_SNAPSHOTS = 100;

const obj: Record<string, Array<string>> = {};

for (let i = 0; i < NUMBER_OF_SNAPSHOTS; i++) {
  obj[`key${i+1}`] = new Array(100).fill(new Array(100).fill('a').join(''));
}


Deno.test(async function noTeardown() {
  await Deno.writeTextFile('test1.json', JSON.stringify(obj, null, 2));

  const start = Date.now();

  for (let i = 0; i < NUMBER_OF_SNAPSHOTS; i++) {
    const snapshots: Record<string, Array<string>> = JSON.parse(await Deno.readTextFile('test1.json'));
    snapshots[`key${i+1}`] = new Array(100).fill(new Array(100).fill('a').join(''));
    await Deno.writeTextFile('test1.json', JSON.stringify(snapshots, null, 2));
  }

  const stop = Date.now();

  await Deno.writeTextFile('result1.txt', `${stop-start}ms`);
});

Deno.test(async function withTeardown() {
  await Deno.writeTextFile('test2.json', JSON.stringify(obj, null, 2));

  const start = Date.now();

  const snapshots: Record<string, Array<string>> = JSON.parse(await Deno.readTextFile('test2.json'));
  for (let i = 0; i < NUMBER_OF_SNAPSHOTS; i++) {
    snapshots[`key${i+1}`] = new Array(100).fill(new Array(100).fill('a').join(''));
  }
  await Deno.writeTextFile('test2.json', JSON.stringify(snapshots, null, 2));

  const stop = Date.now();

  await Deno.writeTextFile('result2.txt', `${stop-start}ms`);
});
```

These are the results (AMD Ryzen 5 3600 and high-ish end SSD):

teardown-chart

I put this together in a couple of minutes just to validate my intuition so let me know if you think this is not a valid test or if there's any mistakes!

A cleanup function (e.g. `teardown`) is also necessary to be able to detect and remove unused snapshots and to provide useful output after a test has finished running (e.g. no. snapshots updated).

Further to the above, I also feel that setup/teardown functions are a fairly standard and uncontroversial part of most testing frameworks (e.g. mocha, jest, jasmine, karma etc all offer before/after each/all functions). These could of course be implemented in a 3rd party testing library but this would probably need to duplicate a lot of the work already done as part of Denos test implementation in managing and running tests/test steps.

Could you elaborate on why setup/cleanup functions were rejected? I tired to search the GitHub issues but couldn't find the discussion! :P

runtime/js/40_testing.js
hyp3rflow: Thanks for the kind review!
Yeah, I think it will be okay with my first suggestion in most cases. I was just worried about if there are many third-party libraries using teardown in an order-sensitive way.
I also agree that it will be awesome with APIs like jest setup-teardown functions. but I think we need to hear the maintainers' thoughts about this kind of API.

bcheidemann: > but I think we need to hear the maintainers' thoughts about this kind of API.

Definitely agree! And this should not be part of this PR either way. I just mentioned it here as it would address the ordering issue you flagged.

> I was just worried about if there are many third-party libraries using teardown in an order-sensitive way.

I am not aware of any but it's a valid point.

> Yeah, I think it will be okay with my first suggestion in most cases.

I agree! Better to keep it simple :D

hyp3rflow:
@bcheidemann that's okay, I don't mind :) I'm thanks to you to reply elaborate comments related to this API.
I'm happy that we can build this PR together 😊

runtime/js/40_testing.js
hyp3rflow: @bcheidemann I have a question.
> test.teardown - [THIS PR] execute a function after the top level parent test has finished executing and all descendant test steps have finished executing. 

What does `top level parent test` means? A single deno test (Deno.test()) or entire test module (a.test.ts)?

bcheidemann: > @bcheidemann I have a question.
> 
> > test.teardown - [THIS PR] execute a function after the top level parent test has finished executing and all descendant test steps have finished executing.
> 
> What does `top level parent test` means? A single deno test (Deno.test()) or single test module (a.test.ts)?

I was thinking a single `Deno.test()` but what do you think is better?

For our case (`assertSnapshot`) it probably makes more sense for it to be a single test module (`a.test.ts`) but I think generally it makes more sense for this to relate to a single `Deno.test()` instance.

hyp3rflow:
Well, I think we can use unload event instead of teardown API here. @bcheidemann

bartlomieju:
> Well, I think we can use unload event instead of teardown API here. @bcheidemann
> 
> I pushed commit on `assertSnapshot` proposal: [denoland/deno_std@`418cc43` (#2039)](https://github.com/denoland/deno_std/pull/2039/commits/418cc43faee5d2e5a5f9e5ecf28d724dfe0518a5) Using `globalThis.onunload`, we can implement `assertSnapshot` without implementing `teardown` API and performance issues. If you think this is okay, I'll remove the `teardown` implementation here and make an issue for it.

If you could get away with `onunload` that would be preferable than adding a new API 👍

dsherret (approved):
LGTM

dsherret (commented):
Sorry, I approved the wrong PR when attempting to retry approving another PR due to the GH outage. I have yet to review this.

hyp3rflow:
@dsherret okay, I will re-request your review :)

cli/dts/lib.deno.ns.d.ts
```diff
@@ -250,6 +250,19 @@ declare namespace Deno {
...
 
   export interface TestContext {
+    /**
+     * The current test name.
+     */
+    name: string;
+    /**
+     * File Uri of the current test code.
+     */
+    origin: string;
```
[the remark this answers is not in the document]
bartlomieju: Yeah, I think it makes sense to add `origin`, relying on users to pass `import.meta.url` is not preferable

runtime/js/40_testing.js
```diff
@@ -1250,6 +1250,20 @@
...
       [SymbolToStringTag]: "TestContext",
+      /**
+       * The current test name.
+       */
+      name: parentStep.name,
+      /**
+       * Parent test context.
+       */
+      parent: parentStep.parent
+        ? createTestContext(parentStep.parent)
```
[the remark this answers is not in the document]
hyp3rflow: 👀 [`222fc0e` (#14007)](https://github.com/denoland/deno/pull/14007/commits/222fc0e2ac68155c521614679d386144ff78cc3c)

hyp3rflow:
@dsherret where can I write the test for name, origin, parentContext?
I found `cli/tests/testdata/test/steps` but I think this is for step on test cli not for testing how test context works.

dsherret:
@hyp3rflow sorry for my delay. It can be added to cli/tests/unit/testing_tests.ts or some similar file. Basically, write some tests and then assert what's in the context.

runtime/js/40_testing.js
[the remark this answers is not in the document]
dsherret: It's probably best for the public API to not be influenced by internal naming. I think `parent` is better since it's shorter.

hyp3rflow:
@dsherret that's okay! I added some tests for `testContext` on `testing_tests.ts`.

runtime/js/40_testing.js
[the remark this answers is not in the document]
hyp3rflow: good. I renamed it to `parent`. ✅ [`95892ba` (#14007)](https://github.com/denoland/deno/pull/14007/commits/95892ba25ef7e510b736cf87f11fe73650635226)

cli/dts/lib.deno.ns.d.ts
dsherret, line 260: Considering that `import.meta.url` is a `string`, I think it makes sense for this to be a string as well.

dsherret (approved):
LGTM. Thanks @hyp3rflow!

bartlomieju (approved):
LGTM

PR #14013, opened 2022-03-17
fix(config-file): config-file.v1.json schema should allow colons in the task name

dsherret:
We accidentally missed this when adding support for having colons.


bartlomieju (approved):
My bad, sorry!

dsherret:
@bartlomieju no, I forgot to mention that we had this validation in the json schema. Not a big deal because things still work :)

PR #14015, opened 2022-03-17
fix(bench): don't await synchronous bench cases

bartlomieju:
This commit changes how benchmarks are run, to not await sync
cases, as it introduces noticable penalty per each iteration.

### DIFF
--- a/runtime/js/40_testing.js
+++ b/runtime/js/40_testing.js
@@ -803,7 +803,7 @@
     }
   }
 
-  function runSyncBench(step) {
+  function runSyncBench(bench, step) {
     const warmupIterations = bench.warmupIterations;
     step.warmup = true;
 
@@ -821,7 +821,7 @@
     return "ok";
   }
 
-  async function runAsyncBench(step) {
+  async function runAsyncBench(bench, step) {
     const warmupIterations = bench.warmupIterations;
     step.warmup = true;
 
@@ -856,11 +856,11 @@
       // function to run it. One might be tempted to just always await
       // `bench.fn`, but that adds a lot of overhead to synchronous
       // benches.
-      if (result instanceof Promise) {
+      if (typeof result.then !== "undefined") {
         await result;
-        return await runAsyncBench(step);
+        return await runAsyncBench(bench, step);
       } else {
-        return runSyncBench(step);
+        return runSyncBench(bench, step);
       }
     } catch (error) {
       return {
--- a/runtime/js/40_testing.js
+++ b/runtime/js/40_testing.js
@@ -803,6 +803,42 @@
     }
   }
 
+  function runSyncBench(step) {
+    const warmupIterations = bench.warmupIterations;
+    step.warmup = true;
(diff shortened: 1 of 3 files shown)


runtime/js/40_testing.js
```diff
@@ -815,21 +851,17 @@
...
-      step.warmup = false;
-
-      for (let i = 0; i < iterations; i++) {
-        await bench.fn(step);
+      const result = bench.fn(step);
+      // Check the return type of the bench so we can use specialized
+      // function to run it. One might be tempted to just always await
+      // `bench.fn`, but that adds a lot of overhead to synchronous
+      // benches.
+      if (result instanceof Promise) {
```
lucacasonato, line 859: This should probably be `result.then !== undefined()`.

bartlomieju:
`reportBenchIteration` still wraps in an async function:
https://github.com/denoland/deno/blob/9c403f146e801bed741d9d66aaa415ff607cb657/runtime/js/40_testing.js#L908-L919

We should have specialized version for sync and async bench. I need to think how to handle that

runtime/js/40_testing.js
```diff
@@ -815,21 +839,56 @@
...
+        );
       }
 
-      const iterations = bench.n;
-      step.warmup = false;
+      const result = benchFn(step);
 
-      for (let i = 0; i < iterations; i++) {
-        await bench.fn(step);
+      if (typeof result.then !== "undefined") {
```
aapoalas, line 866: Perhaps check for "function" type? Although rare, it is theoretically possible that a sync function would return an object like
```ts
const result = {
  then: 3,
};
```

This is not promise-like.

A further check could be done to make sure that the `.then` function has a length of at least one:

```ts
const result = {
  then: () => 3,
};
```

This isn't a valid promise-like object either, doing `await` on that will never resolve.

So a full check could be something like this:

```ts
// Quick-check for normal Promises
if (result instanceof Promise || typeof result === "object" && result !== null && typeof result.then === "function" && result.then.length >= 1) {
  // ...
}
```

One could also consider doing some sort of a test-by-fire on a potential promise-like:

```ts
async function testPromiseLike(maybePromiseLike: unknown, bench) {
  if (typeof maybePromiseLike !== "object" || maybePromiseLike === null) {
    return false;
  } else if (typeof maybePromiseLike.then !== "function" || maybePromiseLike.then.length === 0) {
    return false;
  }
  // Looks like a promise, now lets see if it walks like a promise
  return await Promise.race([
    maybePromiseLike.then(() => true, () => true), // Return or throw, it doesn't matter
    new Promise((res) => setTimeout(() => res(false), bench.timeout)), // Does not resolve within the bench timeout, this does not seem like a real promise
  ]);
}
```

Now if the return value looks and walks like a Promise, it'll get used like a promise. Otherwise it'll be treated as a sync function.

bartlomieju:
Closing in favor of https://github.com/denoland/deno/pull/14305

PR #14019, opened 2022-03-17
feat(core): Add initial support for realms

andreubotella:
This handles bindings, extension JS and sync ops. It also adds the `JsRealm` API that adds methods like `JsRuntime`'s `handle_scope`, `global_object` and `execute_script` specific to the realm.

Towards #13239.

-------------------

@bartlomieju @AaronO @littledivy PTAL. Does this conflict with some of the remaining work to do for ops?


lucacasonato (approved):
Looks good to me. Thanks for the work @andreubotella.

I do wonder if `JsRealm`'s lifetime could not be bound to `JsRuntime`, so it could store a reference to the `JsRuntime` in the `JsRealm` struct. That would alleviate the need to pass the runtime to all the methods on the `JsRealm`.

core/runtime.rs
```diff
@@ -416,9 +421,38 @@ impl JsRuntime {
     self.inspector.as_mut().unwrap()
   }
 
+  pub fn global_realm(&mut self) -> JsRealm {
+    JsRealm::new(self.global_context())
+  }
```
bartlomieju, line 426: Why is it not stored on `JsRuntime`? Seems strange to call into `JsRealm::new` every time you want to access it

andreubotella: Done

bartlomieju:
Yikes, sorry it takes so long @andreubotella, could you rebase the PR?

andreubotella:
> Yikes, sorry it takes so long @andreubotella, could you rebase the PR?

Done.

littledivy (approved):
LGTM!

bartlomieju (approved):
LGTM

PR #14023, opened 2022-03-18
perf(web): optimize Blob functions

littledivy:
Improves `Blob.arrayBuffer()` by ~60% and `Blob.text()` by ~38.4%

Before:
```
blob_text_large_new: 	n = 100, dt = 0.547s, r = 183/s, t = 5470000ns/op
blob_text_large_text:	n = 100, dt = 0.375s, r = 267/s, t = 3750000ns/op
blob_text_large_buf: 	n = 100, dt = 0.217s, r = 461/s, t = 2170000ns/op
```

After:
```
blob_text_large_new: 	n = 100, dt = 0.547s, r = 183/s, t = 5470000ns/op
blob_text_large_text:	n = 100, dt = 0.231s, r = 433/s, t = 2310000ns/op
blob_text_large_buf: 	n = 100, dt = 0.095s, r = 1053/s, t = 950000ns/op
```

Instead of collecting blob parts in JS, this PR Introduces a new Rust op for reading all the blob data at once.


lucacasonato (commented):
This PR combines multiple changes into a single PR, making it difficult to tell what change actually causes what change. This PR has   two distinct changes: reading all parts at once via a single op, and reading them all into a single buffer with some unsafe code. These could be seperated, and should be benched separately. 

I am really not in favor of these unsafe code additions for tiny little performance improvements.

littledivy:
> reading all parts at once via a single op

There are 2 seperate ops, both read all data into a buffer one returns a ByteString (for `blob.text()`) and another returns a `ZeroCopyBuf` (for `blob.arrayBuffer()`).

They are benched seperately in `blob_text_large_text` and `blob_text_large_buf`

bartlomieju:
I agree with Luca, this PR should be split into two. I'm not sure if this complexity is justified for a marginal perf gain.

PR #14030, opened 2022-03-18
chore: fix development issues on Windows

### LINKED ISSUE #14014: `integration::run::issue9750` gets stuck forever
When I run `cargo test` on my machine, I get a message about `integration::run::issue9750` timing out, and `cargo test` never finishes. When I run that test specifically, it also never finishes running, but if you print its stdout (with `-- --nocapture`), you see that the test in fact fails:

```
OUTPUT Enter \'yy\':\r\n
INPUT yy\n
ECHO: yy\r\n
OUTPUT ⚠️  ️Deno requests env access. Run again with --allow-env to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]  
INPUT n\n
ECHO: n\r\n
OUTPUT ⚠️  ️Deno requests env access to \"SECRET\". Run again with --allow-env to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]  
thread 'integration::run::issue9750' panicked at 'assertion failed: `(left == right)`

Diff < left / right > :
<⚠️  ️Deno requests env access to "SECRET". Run again with --allow-env to bypass this prompt.
>⚠️  ️Deno requests env access to "SECRET". Run again with --allow-run to bypass this prompt.
    Allow? [y/n (y = yes allow, n = no deny)]

', test_util/src/lib.rs:1910:11
```

The actual permissions prompt correctly asks for `--allow-env`, but [the test expects it to ask for `--allow-run`](https://github.com/denoland/deno/blob/main/cli/tests/integration/run_tests.rs#L2337).

I don't know why the test keeps running, though, or why this hasn't been observed in CI.

dsherret:
Fixes some local development issues that have creeped up over the past few weeks on Windows.

### Please merge on review

Closes #14014

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,14 @@
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
 
+[target.'cfg(all(windows, debug_assertions))']
+rustflags = [
+  "-C",
+  "target-feature=+crt-static",
+  "-C",
+  # double the stack size to prevent swc overflowing the stack in debug
+  "link-arg=/STACK:2097152",
+]
+
 [target.aarch64-apple-darwin]
 rustflags = ["-C", "link-arg=-fuse-ld=lld"]
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -2334,7 +2334,7 @@ fn issue9750() {
       Input("yy\n"),
       Output("⚠️  ️Deno requests env access. Run again with --allow-env to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]"),
       Input("n\n"),
-      Output("⚠️  ️Deno requests env access to \"SECRET\". Run again with --allow-run to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]"),
+      Output("⚠️  ️Deno requests env access to \"SECRET\". Run again with --allow-env to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]"),
       Input("n\n"),
       Output("error: Uncaught (in promise) PermissionDenied: Requires env access to \"SECRET\", run again with the --allow-env flag\r\n"),
     ],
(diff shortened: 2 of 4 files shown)


cli/tests/integration/run_tests.rs
```diff
@@ -2334,7 +2334,7 @@ fn issue9750() {
       Input("yy\n"),
       Output("⚠️  ️Deno requests env access. Run again with --allow-env to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]"),
       Input("n\n"),
-      Output("⚠️  ️Deno requests env access to \"SECRET\". Run again with --allow-run to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]"),
+      Output("⚠️  ️Deno requests env access to \"SECRET\". Run again with --allow-env to bypass this prompt.\r\n   Allow? [y/n (y = yes allow, n = no deny)]"),
```
dsherret, line 2337: Ummm... how is this passing on linux & mac?

dsherret: This fails for me on linux and then hangs forever.

.cargo/config
```diff
@@ -1,5 +1,14 @@
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
 
+[target.'cfg(all(windows, debug_assertions))']
+rustflags = [
+  "-C",
+  "target-feature=+crt-static",
+  "-C",
+  # double the stack size to prevent swc overflowing the stack in debug
+  "link-arg=/STACK:2097152",
```
bartlomieju, line 10: Again...? 😬

dsherret: That happened in the past? FWIW swc’s tests on its repo requires increasing the stack size.

cli/tests/integration/run_tests.rs
kitsonk: See #14014, couldn't figure out why CI wasn't failing but was hanging locally, seems you figured it out.

.cargo/config
bartlomieju: Yeah, I believe we once increased stack size on Windows, but then there were some improvements that rendered it unneeded

dsherret: I think probably something must have increased on our side that just tipped it over. It never used to stack overflow on Windows with this swc version.

bartlomieju (approved):
LGTM

Issue #13991, opened 2022-03-16
Console log error with a Proxy of a Map


oscarotero:
This code fails on Deno:

```js
console.log(new Proxy(new Map(), {}));
```

```
error: Uncaught TypeError: Method Map.prototype.entries called on incompatible receiver #<Map>
console.log(new Proxy(new Map(), {}));
        ^
    at Proxy.entries (<anonymous>)
    at inspectIterable (deno:ext/console/02_console.js:391:16)
    at inspectMap (deno:ext/console/02_console.js:889:12)
    at inspectObject (deno:ext/console/02_console.js:1245:14)
    at _inspectValue (deno:ext/console/02_console.js:664:16)
    at inspectValue (deno:ext/console/02_console.js:679:11)
    at inspectArgs (deno:ext/console/02_console.js:1818:19)
    at console.log (deno:ext/console/02_console.js:1872:9)
    at file:///Users/oscarotero/proxy.js:2:9
```

It works fine in browsers.

As a temporary workaround, this code seems to work fine:

```js
console.log(new Proxy({ map: new Map() }, {}));
```

PR #14032, opened 2022-03-19
fix(ext/console): Console log error with a Proxy of a Map

CGQAQ:
closed: #13991 

before
image

after this pr
image
I don't know if this is the right way to resolve this issue, if not please let me know the right way, I'll change the PR

### DIFF
--- a/cli/tests/unit/console_test.ts
+++ b/cli/tests/unit/console_test.ts
@@ -1614,7 +1614,7 @@ Deno.test(function consoleLogShouldNotThrowErrorWhenInputIsProxiedSet() {
   mockConsole((console, out) => {
     const proxiedSet = new Proxy(new Set(), {});
     console.log(proxiedSet);
-    assertEquals(stripColor(out.toString()), "Set {}");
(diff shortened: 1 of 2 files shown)


kt3k (commented):
@CGQAQ Thanks for the PR. Can you add a test case for `console.log(new Proxy(new Map(), {}));`?

CGQAQ:
> @CGQAQ Thanks for the PR. Can you add a test case for `console.log(new Proxy(new Map(), {}));`?

sure

kt3k:
BTW node and chrome seem doing the different things for this. Which do you think is better?

node
スクリーンショット 2022-03-19 20 55 58

chrome
スクリーンショット 2022-03-19 20 56 22

CGQAQ:
I think chrome way is better, it just prints out proxyDetails and you can tell it's a proxy, not a vanilla Map. But that way will break a LOT of existing tests

kt3k (approved):
@CGQAQ LGTM. Thanks for the fix!

PR #14033, opened 2022-03-19
fix(ops): throw TypeError on op return failure

### LINKED ISSUE #14028: sync op result serialization errors result in panic and abort
If an error occurs in serde_v8 when converting the output of a sync op, it results in a panic and the process aborts.

The panic occurs because [`unwrap()` is called](https://github.com/denoland/deno/blob/bd481bf095f920a419ea55543f911e087f98f36f/ops/lib.rs#L263) on the result from `serde_v8::to_v8`. (The resulting panic then causes the whole process to abort, maybe due to https://github.com/denoland/rusty_v8/issues/534.)

Example - this test results in "fatal runtime error: failed to initiate panic, error 5":

```rust
#[test]
fn test_serde_v8_error() {
  use serde::ser::{Serialize, Serializer};

  struct ErrOnSerialize;

  impl Serialize for ErrOnSerialize {
    fn serialize<S>(&self, _serializer: S) -> std::result::Result<S::Ok, S::Error> where S: Serializer {
      Err(serde::ser::Error::custom("always errors"))
    }
  }

  #[op]
  fn op_err() -> Result<ErrOnSerialize, anyhow::Error> {
    Ok(ErrOnSerialize)
  }

  run_in_task(|cx| {
    let ext = Extension::builder().ops(vec![op_err::decl()]).build();
    let mut runtime = JsRuntime::new(RuntimeOptions {
      extensions: vec![ext],
      ..Default::default()
    });
    runtime
      .execute_script(
        "test_serde_v8_error.js",
        "Deno.core.opSync('op_err')",
      )
      .unwrap();
    if let Poll::Ready(Err(_)) = runtime.poll_event_loop(cx, false) {
      unreachable!();
    }
  });
}
```

AaronO:
Fixes #14028

### DIFF
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -2854,4 +2854,21 @@ assertEquals(1, notify_return_value);
     assert_eq!(2, PROMISE_REJECT.load(Ordering::Relaxed));
     assert_eq!(2, UNCAUGHT_EXCEPTION.load(Ordering::Relaxed));
   }
+
+  #[test]
+  fn test_op_return_serde_v8_error() {
+    #[op]
+    fn op_err() -> Result<std::collections::BTreeMap<u64, u64>, anyhow::Error> {
+      Ok([(1, 2), (3, 4)].into_iter().collect()) // Maps can't have non-string keys in serde_v8
+    }
+
+    let ext = Extension::builder().ops(vec![op_err::decl()]).build();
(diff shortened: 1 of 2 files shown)


bartlomieju (changes requested):
Please add a test

littledivy (approved):
LGTM

Issue #10783, opened 2021-05-29
Make snapshots deterministic


lucacasonato:
Seems we just need to fix the V8 random seed to a fixed value during snapshotting: https://github.com/nodejs/node/commit/5116a6a9b4d8420149af74ac29074cd4d7f1f6b9

Thanks @joyeecheung for the pointer :-)

lucacasonato:
https://twitter.com/hashseed/status/1398608313089200131?s=21

ry:
I don't understand - are snapshots not deterministic currently?

lucacasonato:
They are not. Here two builds with no code changes in between:

```
/m/s/P/g/d/deno ❯❯❯ md5sum /mnt/starship/Projects/github.com/denoland/deno/target/debug/build/deno_runtime-87cbb1a0250dce57/out/CLI_SNAPSHOT.bin
7e6848bec338150c4a7e62085579bfcb  /mnt/starship/Projects/github.com/denoland/deno/target/debug/build/deno_runtime-87cbb1a0250dce57/out/CLI_SNAPSHOT.bin
/m/s/P/g/d/deno ❯❯❯ md5sum /mnt/starship/Projects/github.com/denoland/deno/target/debug/build/deno_runtime-87cbb1a0250dce57/out/CLI_SNAPSHOT.bin
b142b3da94caffc66e3189d2bc8190d7  /mnt/starship/Projects/github.com/denoland/deno/target/debug/build/deno_runtime-87cbb1a0250dce57/out/CLI_SNAPSHOT.bin
```

lucacasonato:
This is likely the cause for #9244

ry:
Wow - I had no idea. I wonder if this is impinging on our incremental build time...

PR #14037, opened 2022-03-19
feat(core): deterministic snapshots

littledivy:
Closes https://github.com/denoland/deno/issues/10783

```shell
$ md5sum target/debug/build/deno_runtime-bc62560f5114b792/out/CLI_SNAPSHOT.bin
0205db97c85b0a06d5b0ea6b23258e6f  target/debug/build/deno_runtime-bc62560f5114b792/out/CLI_SNAPSHOT.bin

$ touch ext/console/lib.rs && cargo build
# blabla

$ md5sum target/debug/build/deno_runtime-bc62560f5114b792/out/CLI_SNAPSHOT.bin
0205db97c85b0a06d5b0ea6b23258e6f  target/debug/build/deno_runtime-bc62560f5114b792/out/CLI_SNAPSHOT.bin
```


cli/tests/integration/eval_tests.rs
```diff
@@ -20,6 +20,30 @@ fn eval_p() {
   assert_eq!("3", stdout_str);
 }
 
+// Make sure that snapshot flags don't affect runtime.
+#[test]
+fn eval_randomness() {
```
bartlomieju, line 25: Can we have a test that checks that snapshots are actually predicatable?

ry: I'd be happy to hard code the snapshot checksum in the tests 0205db97c85b0a06d5b0ea6b23258e6f

We just have to have a comment on how to update it easily

AaronO: > I'd be happy to hard code the snapshot checksum in the tests (0205db97c85b0a06d5b0ea6b23258e6f)
> 
> We just have to have a comment on how to update it easily

The snapshot/checksum will change pretty much all the time (slightest JS change in this repo) so that would be impractical IMO

ry: Probably not that often -  only if you change the JS code. Many PRs would not update it. 

Maybe the hick-up isn't worth it. I feel like we might discover something if we watched the checksum change.

bartlomieju:
It's not really surprising to me that you get different checksum on CI than locally. I'm skeptical about this test being useful at all.

Issue #14017, opened 2022-03-17
Should op structs show up in a the `rustdoc` documentation?


andreubotella:
https://docs.rs/deno_core/0.124.0/deno_core/index.html

#13861 made it so defining ops creates corresponding structs, and those structs shows up in the documentation generated by `rustdoc` / `cargo doc`. Should they?

PR #14038, opened 2022-03-19
chore(ops): hide implementation details from rustdoc

littledivy:
Closes https://github.com/denoland/deno/issues/14017

It's fine if ops show up. op-generated methods can change anytime, we should hide that from the rustdoc. 

The doc comment for the op says how to register the op. (with `op::decl()`)


AaronO:
Made some tweaks and included the op's name in the generated doc, e.g:

image

AaronO (approved):
LGTM

PR #14044, opened 2022-03-20
chore(fetch, mimesniff): Refactor the "extract a MIME type" algorithm

### LINKED ISSUE #14002: Duplicated implementation of "extract a MIME type" in `23_request.js` and `23_response.js`
Both `Request.prototype[_mimeType]` and `Response.prototype[_mimeType]` implement the [extract a MIME type](https://fetch.spec.whatwg.org/#concept-header-extract-mime-type) algorithm. These should be deduplicated.

In fact, it would make sense to implement this algorithm in `ext/web/01_mimesniff.js`, taking `values` as its argument rather than a header list.

andreubotella:
`Request.prototype[_mimesniff]` and `Response.prototype[_mimesniff]` are functions that implement the "extract a MIME type" algorithm from the fetch spec. These implementations are the same line by line, though. This change deduplicates that by moving most of the implementation to a `extractMimeType` function under `ext/web`'s `01_mimesniff.js`, and simply calling that with the decoded and split headers from the `[_mimesniff]` functions.

Closes #14002.


AaronO (approved):
LGTM

PR #14051, opened 2022-03-20
perf(http): avoid per header alloc

AaronO:
When checking if header matches a "special" header (that impacts our body handling, etc...)

[make_ascii_lowercase](https://doc.rust-lang.org/std/string/struct.String.html#method.make_ascii_lowercase) could also make sense here

### Impact on throughput

Unfortunately there's a lot of variance in our tests and it's not fully formalized but I do observe a 1-3% improvement in throughput (sampled over 3 runs):

```
Before (1.20.1)
Requests/sec: 126285.80
Requests/sec: 127775.95
Requests/sec: 128875.33

After (this PR):
Requests/sec: 130615.10
Requests/sec: 129957.80
Requests/sec: 129056.16
```

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -517,50 +517,36 @@ async fn op_http_write_headers(
 
   builder.headers_mut().unwrap().reserve(headers.len());
   for (key, value) in &headers {
-    match &*key.to_ascii_lowercase() {
-      b"cache-control" => {
-        if let Ok(value) = std::str::from_utf8(value) {
-          if let Some(cache_control) = CacheControl::from_value(value) {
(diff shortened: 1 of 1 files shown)


bartlomieju (commented):
Is there any measureable impact on performance?

AaronO:
> Is there any measureable impact on performance?

Yes, there's a 1-3% improvement, I've discussed with @littledivy a bench that would bypass the OS' net stack & loadgen tool better suited to measuring http-efficiency & should have less variance.

littledivy (approved):
lgtm

lucacasonato (approved):

PR #14062, opened 2022-03-21
fix(core): variadic opSync/opAsync

AaronO:
To allow opcalls with more than 2 generic-args JS-side


littledivy (commented):
LGTM! Please add a test before landing

PR #14063, opened 2022-03-21
feat(core): disableable extensions & ops

AaronO:
Streamlines a common middleware pattern and provides foundations for avoiding variably sized v8::ExternalReferences


ry (commented):
A small unit test in core would be appropriate, I think

bartlomieju:
> Main open question is if disabled ops should throw or just no-op by mapping to op_void_a?sync thus returning undefined

Throwing is less surprising than silently doing a noop

littledivy (approved):
LGTM

bartlomieju (commented):
So it looks like you want to silently no op for disabled ops. I don't like this, it's surprising in my opinion

PR #6801, opened 2020-07-18
Upgrade to rusty_v8 0.7.0

piscisaureus:

PR #14069, opened 2022-03-22
chore(core): Remove a note that is no longer relevant.

andreubotella:
This note about how `v8::SnapshotCreator::create_blob` must not be called from a `HandleScope` stopped being relevant in #6801, and was now attached to code that had nothing to do with `HandleScope`s.


andreubotella:
This was the code before the change in #6801 that stopped making the note relevant:

https://github.com/denoland/deno/blob/aebea6bd24dc1788c6db2f58204177af1c5d2dd6/core/core_isolate.rs#L328-L335

Issue #14071, opened 2022-03-22
`deno check`


ry:
- Add `deno check`. It should be exactly the same thing as `deno cache`. 
- Add `deno run --check` for forward compatibility. (It will be required in the future for type-checking.)
- Print warning during `deno run` if `--check` is not provided. "Warning: In future releases deno run will not automatically type check without the --check flag."

lucacasonato:
I am not super happy with the way the warning is proposed to work. I think we should have a `DENO_FUTURE_TYPECHECKING_BEHAVIOUR=1` flag or similar that will switch Deno into the "new" no type checking by default behaviour completely right now. The `deno run` message should be updated to say: "Warning: In future releases deno run will not automatically type check without the --check flag. To opt into this new behaviour now, specify DENO_FUTURE_TYPECHECKING_BEHAVIOUR=1."

In a couple of releases we can then turn this from a warning into an error, and then subsequently from an opt in to an opt out, and then finally fully switch.

This is very similar to how the `go.mod` adoption in the Go ecosystem worked (through the `GO111MODULE=1` env var).

ry:
Duplicate of https://github.com/denoland/deno/issues/11786

PR #14072, opened 2022-03-22
feat: Add "deno check" subcommand for type checking

bartlomieju:
Currently just an alias for "deno cache". 

Prints warning for "deno run" subcommand if "--check" flag is not present
and there's no "--no-check" flag. Adds "DENO_FUTURE_CHECK" env
variable that allows to opt into new behavior now.

Closes https://github.com/denoland/deno/issues/14071
Closes https://github.com/denoland/deno/issues/11786


cli/main.rs
```diff
@@ -601,6 +602,21 @@ async fn cache_command(
   Ok(0)
 }
 
+async fn check_command(
+  flags: Flags,
+  check_flags: CheckFlags,
+) -> Result<i32, AnyError> {
+  // NOTE(bartlomieju): currently just an alias for `deno cache`, but
+  // it will be changed in Deno 2.0.
```
dsherret, line 610: Maybe add a note to say that `deno cache` will be changed to not type check?

dsherret (approved):
LGTM

cli/main.rs
```diff
@@ -1148,6 +1164,10 @@ async fn run_command(
   flags: Flags,
   run_flags: RunFlags,
 ) -> Result<i32, AnyError> {
+  if flags.check == CheckFlag::All && !flags.has_check_flag {
+    info!("{} In future releases deno run will not automatically type check without the --check flag.", colors::yellow("Warning"));
+  }
```
dsherret, line 1169: This seems a bit extreme. Don't think we should do this warning.

bartlomieju: See https://github.com/denoland/deno/issues/14071

ry:
Please implement the env var described here https://github.com/denoland/deno/issues/11786

cli/flags.rs
```diff
@@ -226,12 +232,32 @@ pub enum CheckFlag {
   Local,
 }
 
-impl Default for CheckFlag {
+impl Default for TypecheckMode {
   fn default() -> Self {
     Self::All
```
nayeemrmn, line 237: > Discussed offline. `--check` should default to type checking only local files; to type check all files (including remote) we should do `--check=all`. Similar to how `--no-check` skips all, or `--no-check=remote` type checks only local files.

_Originally posted by @bartlomieju in https://github.com/denoland/deno/issues/11786#issuecomment-1077700276_

I think this should be done in this PR for 1.21. It's not a workflow-breaking change so it needs no transition period if we're doing it anyway. It's just an immediate improvement in usability.

bartlomieju: That's already implemented in this PR.

nayeemrmn: But `deno check` aliases `deno cache` and doesn't seem to have an `--all` flag. Also the temporary case where `deno run main.ts` is called, `DENO_FUTURE_CHECK` is not set and the warning is shown, it will use `TypecheckMode::All`. I think this default needs to be changed.

cli/main.rs
```diff
@@ -585,6 +587,21 @@ async fn cache_command(
   Ok(0)
 }
 
+async fn check_command(
+  flags: Flags,
+  check_flags: CheckFlags,
+) -> Result<i32, AnyError> {
```
ry, line 593: This function is superfluous. Rename cache_command to check_command.

bartlomieju: I'm not sure I agree, see https://github.com/denoland/deno/pull/14072#discussion_r835952423

cli/flags.rs
```diff
@@ -5343,4 +5481,76 @@ mod tests {
...
+        has_check_flag: true,
+        unstable: true,
+        ..Flags::default()
+      }
+    );
+
+    let r = flags_from_vec(svec![
+      "deno",
+      "run",
+      "--unstable",
```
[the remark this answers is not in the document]
bartlomieju: It doesn't require check; it's just a test with several flags...

cli/main.rs
```diff
@@ -1140,6 +1157,11 @@ async fn run_command(
   flags: Flags,
   run_flags: RunFlags,
 ) -> Result<i32, AnyError> {
+  if !flags.has_check_flag && flags.type_check_mode == TypeCheckMode::All {
+    info!("{} In future releases deno run will not automatically type check without the --check flag.
```
lucacasonato, line 1161:
```suggestion
    info!("{} In future releases `deno run` will not automatically type check without the --check flag. 
```

Does this respect `--quiet`?

bartlomieju: Yes, it does

cli/flags.rs
bartlomieju: This is now implemented.

cli/flags.rs
```diff
@@ -688,6 +721,30 @@ Future runs of this module will trigger no downloads or compilation unless
     )
 }
 
+fn check_subcommand<'a>() -> Command<'a> {
+  compile_args_without_no_check(Command::new("check"))
+  .arg(
+    Arg::new("all")
+      .long("all")
+      .help("Type-check all modules, including remote")
+    )
```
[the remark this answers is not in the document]
nayeemrmn: > Rename to `--remote`?

That might be better. If we made another category of diagnostics/locations opt-in, that should be its own flag e.g. `deno check --remote --dts`.

At this point I'm not sure if `deno run --check` is a good idea or if we should insist `deno check` is used to type check. If we kept it the API would have to be more like `deno run --check=--remote,--dts`.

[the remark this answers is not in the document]
bartlomieju: > At this point I'm not sure if deno run --check is a good idea 

We need it for at least a few minor releases to give users time to migrate to `deno check`. I don't feel strongly about it, but I think we could punt this discussion until later time.

bartlomieju:
@ry this is now ready to land

ry (approved):
LGTM

PR #14073, opened 2022-03-22
fix(unstable): upgrade deno_task_shell to 0.2

dsherret:
Changelog: https://github.com/denoland/deno_task_shell/releases/tag/0.2.0 (has some `feat`s in it, but this is unstable so categorizing this PR as a fix so it gets in the next patch)


bartlomieju (approved):
LGTM

PR #14076, opened 2022-03-22
cleanup(serde_v8): streamline magic types

AaronO:
Formalize magic types and how they're transerialized

Also makes `U16String` deserializable


serde_v8/src/magic/tr8.rs
```diff
@@ -0,0 +1,136 @@
...
+    ) -> std::fmt::Result {
+      formatter.write_str("a ")?;
+      formatter.write_str(std::any::type_name::<T>())
+    }
+
+    fn visit_u64<E>(self, ptr: u64) -> Result<Self::Value, E>
+    where
+      E: serde::de::Error,
+    {
+      Ok(unsafe { std::mem::transmute_copy::<T, T>(std::mem::transmute(ptr)) })
```
lucacasonato, line 78: Needs safety comment.

serde_v8/src/magic/buffer.rs
```diff
@@ -94,61 +98,54 @@ impl From<Vec<u8>> for MagicBuffer {
...
+      // maybe ArrayBufferView
+      false => v8::Local::<v8::ArrayBufferView>::try_from(value)
+        .and_then(|view| ZeroCopyBuf::try_from((scope, view))),
     }
-
-    static FIELDS: [&str; 0] = [];
-    let visitor = ValueVisitor {};
-    deserializer.deserialize_struct(BUF_NAME, &FIELDS, visitor)
+    .map_err(|_| Error::ExpectedBuffer)?;
+    Ok(Self::FromV8(zero_copy_buf))
```
andreubotella, line 128: The names of the `MagicBuffer` variants are now extremely confusing. Consider renaming them in a follow-up.

serde_v8/src/magic/tr8.rs
```diff
@@ -0,0 +1,136 @@
...
+  V: serde::de::Visitor<'de>,
+  E: serde::de::Error,
+{
+  let y = visitor.visit_u64::<E>(unsafe { std::mem::transmute(&x) });
+  std::mem::forget(x);
+  y
+}
+
+pub(crate) fn send_opaque<T: Sized>(x: &T) -> u64 {
+  (x as *const T as *const u8) as u64
```
andreubotella, line 102: Since `T: Sized`, `*const T` is a thin pointer and can be cast directly to `u64` without the intermediate cast to `*const u8`.

Also, these opaque functions should be `unsafe fn`, since they impose heavy safety constraints on the callers.

serde_v8/src/magic/tr8.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+pub(crate) fn recv_opaque<T: ?Sized>(ptr: &T) -> u64 {
+  unsafe { *(ptr as *const T as *const u8 as *const u64) }
```
andreubotella, line 106: Here, `*const T` might be a fat pointer, but casting it to `*const u64` is enough to drop the metadata, so the intermediate cast to `*const u8` is also not needed here.

Also, the validity of this function assumes that `T` is the same size as `u64`, which doesn't seem to be guaranteed.

serde_v8/src/magic/buffer.rs
```diff
@@ -94,61 +98,54 @@ impl From<Vec<u8>> for MagicBuffer {
...
-    let f2: u64 = hack[1] as u64;
-    s.serialize_field(BUF_FIELD_1, &f1)?;
-    s.serialize_field(BUF_FIELD_2, &f2)?;
-    s.end()
+
+    if buf.is_empty() {
+      let ab = v8::ArrayBuffer::new(scope, 0);
+      return Ok(
+        v8::Uint8Array::new(scope, ab, 0, 0)
+          .expect("Failed to create UintArray8")
```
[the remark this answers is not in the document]
AaronO: fixed the typo, this code was moved (typo is present in `main`)

serde_v8/src/magic/buffer.rs
AaronO: Yeah going to fix them up here

serde_v8/src/de.rs
```diff
@@ -338,57 +339,19 @@ impl<'de, 'a, 'b, 's, 'x> de::Deserializer<'de>
...
-      let written = v8str.write_one_byte(
-        self.scope,
-        &mut buffer,
-        0,
-        v8::WriteOptions::NO_NULL_TERMINATION,
-      );
-      assert!(written == len);
-      return visitor.visit_byte_buf(buffer);
+    // Magic types
+    if name == Buffer::magic_name() {
```
aapoalas, line 343: Maybe match + implicit return expressions?

AaronO: Match arms can't have function calls and these values can't yet be associated attributes because `std::any::type_name::<T>()` isn't yet const, see https://github.com/rust-lang/rust/issues/63084

Obviously could do this semi-manually but it's not a huge deal and relatively unimportant

serde_v8/src/magic/value.rs
```diff
@@ -29,50 +27,22 @@ impl<'s> From<Value<'s>> for v8::Local<'s, v8::Value> {
...
-    };
-    let hack: u64 = unsafe { std::mem::transmute(mv) };
-    s.serialize_field(FIELD, &hack)?;
-    s.end()
+impl ToV8 for Value<'_> {
+  fn to_v8<'a>(
+    &self,
+    _scope: &mut v8::HandleScope<'a>,
+  ) -> Result<v8::Local<'a, v8::Value>, crate::Error> {
+    // SAFETY: not fully safe, since lifetimes are detached from original scope
```
lucacasonato, line 35: Then this function should be unsafe to make the caller aware of this.

serde_v8/src/magic/value.rs
```diff
@@ -29,50 +27,22 @@ impl<'s> From<Value<'s>> for v8::Local<'s, v8::Value> {
...
-
-    static FIELDS: [&str; 1] = [FIELD];
-    let visitor = ValueVisitor { p1: PhantomData };
-    deserializer.deserialize_struct(NAME, &FIELDS, visitor)
+impl FromV8 for Value<'_> {
+  fn from_v8(
+    _scope: &mut v8::HandleScope,
+    value: v8::Local<v8::Value>,
+  ) -> Result<Self, crate::Error> {
+    // SAFETY: not fully safe, since lifetimes are detached from original scope
```
lucacasonato, line 45: Ditto

serde_v8/src/magic/transl8.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Copies an "opaque" ptr from a reference to an opaque ptr (transerialized)
+/// NOTE: ptr-to-ptr, extra indirection
+pub(crate) unsafe fn opaque_recv<T: ?Sized>(ptr: &T) -> u64 {
+  *(ptr as *const T as *const u64)
+}
+
+/// Transmutes an "opaque" ptr back into a reference
+pub(crate) unsafe fn opaque_deref<'a, T>(ptr: u64) -> &'a T {
+  std::mem::transmute(ptr)
```
andreubotella, line 108:
```suggestion
  std::mem::transmute(ptr as usize)
```

Do we (officially or unofficially) support any 32-bit platforms?

serde_v8/src/magic/transl8.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/// Transmutes & copies the value from the "opaque" ptr
+/// NOTE: takes ownership & requires other end to forget its ownership
+pub(crate) unsafe fn opaque_take<T>(ptr: u64) -> T {
+  std::mem::transmute_copy::<T, T>(std::mem::transmute(ptr))
```
andreubotella, line 114:
```suggestion
  let ptr = std::mem::transmute::<*const T>(ptr as usize);
  ptr.read()
```
Clearer IMO.

serde_v8/src/magic/value.rs
AaronO: This is never directly called by end-users. It's the result of attempting to de/serialize `serde_v8::Value`s

serde_v8/src/magic/value.rs
AaronO: This is never directly called by end-users. It's the result of attempting to de/serialize serde_v8::Values

serde_v8/src/magic/transl8.rs
AaronO: No we don't, but it wouldn't be hard to support with the now streamlined transerializing, might do in a follow-up but we have no 32-bit deno builds or official 32-bit rusty_v8 builds

serde_v8/src/magic/buffer.rs
```diff
@@ -94,61 +77,54 @@ impl From<Vec<u8>> for MagicBuffer {
...
+    value: v8::Local<v8::Value>,
+  ) -> Result<Self, crate::Error> {
+    let zero_copy_buf = match value.is_array_buffer() {
+      // ArrayBuffer
+      true => v8::Local::<v8::ArrayBuffer>::try_from(value)
+        .and_then(ZeroCopyBuf::try_from),
+      // maybe ArrayBufferView
+      false => v8::Local::<v8::ArrayBufferView>::try_from(value)
+        .and_then(|view| ZeroCopyBuf::try_from((scope, view))),
     }
```
ry, line 126: 😏

AaronO: Not new code ™️, but will cleanup

serde_v8/src/ser.rs
```diff
@@ -212,191 +214,57 @@ impl<'a, 'b, 'c> ser::SerializeStruct for ObjectSerializer<'a, 'b, 'c> {
...
-    match key {
-      magic::bytestring::FIELD_PTR => {
-        self.ptr = std::ptr::NonNull::new(transmuted as *mut u8);
-      }
-      magic::bytestring::FIELD_LEN => {
-        self.len = Some(transmuted as usize);
-      }
-      _ => unreachable!(),
+    if key != MAGIC_FIELD {
+      unreachable!()
```
ry, line 245: `assert_eq!(key, MAGIC_FIELD)`

ry (commented):
<s>Would be good to have a unit test showing deserializing `U16String`</s>

nevermind - i see it

serde_v8/src/magic/buffer.rs
```diff
@@ -94,61 +77,54 @@ impl From<Vec<u8>> for MagicBuffer {
...
+    scope: &mut v8::HandleScope,
+    value: v8::Local<v8::Value>,
+  ) -> Result<Self, crate::Error> {
+    let zero_copy_buf = match value.is_array_buffer() {
+      // ArrayBuffer
+      true => v8::Local::<v8::ArrayBuffer>::try_from(value)
+        .and_then(ZeroCopyBuf::try_from),
+      // maybe ArrayBufferView
+      false => v8::Local::<v8::ArrayBufferView>::try_from(value)
+        .and_then(|view| ZeroCopyBuf::try_from((scope, view))),
```
littledivy, line 125: perf nit: `try_from` again rechecks `value.is_array_buffer()`. I think it's safe to cast directly at this point?

serde_v8/src/magic/buffer.rs
AaronO: Done

serde_v8/src/magic/buffer.rs
AaronO: Will recheck and bench in a follow-up, want to reduce scope of unsafe code for now

serde_v8/src/magic/transl8.rs
```diff
@@ -0,0 +1,143 @@
...
+      formatter.write_str(T::NAME)
+    }
+
+    fn visit_u64<E>(self, ptr: u64) -> Result<Self::Value, E>
+    where
+      E: serde::de::Error,
+    {
+      // SAFETY: opaque ptr originates from visit_magic, which forgets ownership so we can take it
+      Ok(unsafe { opaque_take(ptr) })
+    }
```
ry, line 73: I might be completely missing the point here - but is this assuming that pointers can be passed to JavaScript as 64 bit integers?

AaronO: No it's not, we're simply transmuting pointers/references across the "serde boundary" as opaque u64s so our specialized de/serializer can pick it up

This refactor doesn't change any of serde_v8's observable qualities, it streamlines magic types by formalizing transerialization making them significantly easier to reason about, maintain, evolve or extend.

ry (approved):
LGTM - it's complex stuff that I don't understand completely, but tests pass.

PR #14081, opened 2022-03-22
chore: replace `.expect("...")` calls with `.unwrap()` in integration test code

dsherret:
Didn't bother with other test blocks yet.


ry:
thank you - `expect()` is obnoxiously overused in our codebase

PR #14083, opened 2022-03-22
chore: remove all `pub(crate)`s from the cli crate

dsherret:
`pub(crate)` is viral and not necessary for the binary because it's not being used as a library.


ry (approved):
LGTM

PR #14087, opened 2022-03-23
chore: cancel deprecation of FsWatcher.return

### LINKED ISSUE #14086: Proposal: Cancel the deprecation of FsWatcher.return method
We added `.close` method, and deprecated `.return` method in #10577. While `.close` is fine as a way for users to explicitly close the watcher, there were some inputs from the community that `.return` method is also useful for automatically closing the watcher when you break from the async iterable loop. (Thanks @mitranim and @connorlogin for the suggestions)

```js
const watcher = Deno.watchFs(".");
  if (event.kind === "remove") {
    break;
  }
}
// Here you still need to call watcher.close(), if we removed `.return` in 2.0
```

So I suggest we should cancel the deprecation of `.return` and keep it as is (with the clarification of the purpose of each method.)

kt3k:
This PR cancels the deprecation of `FsWatcher.prototype.return` method as suggested in #14086

closes #14086

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2209,9 +2209,11 @@ declare namespace Deno {
     /** Stops watching the file system and closes the watcher resource. */
     close(): void;
     /**
-     * Stops watching the file system and closes the watcher resource.
+     * This method is automatically called when you break from the for-await
+     * loop on the watcher, and that stops watching the file system.
      *
-     * @deprecated Will be removed at 2.0.
+     * This method shouldn't be called explicitly from the user's code. You
+     * should use `.close()` method instead for stopping the watcher.
      */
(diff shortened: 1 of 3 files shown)


bartlomieju:
This PR was discussed during CLI design meeting. With garbage collectable resources on the horizon and the fact that other Deno APIs that implement async iterator protocol don't have `return` method, we feel we should still go ahead with deprecation of this API.

kt3k:
Ok. Let's close this for now. Our stance for this is "All our async iterable APIs don't provide `.return` method. FsWatcher doesn't provide it either for consistency"

kt3k:
BTW If we make file watcher resource garbage collectable, does the file watcher automatically stop watching when the user breaks the async loop? (Probably this is something the users who requested this cancellation are curious about)

PR #14090, opened 2022-03-23
fix(cli/emit): Check JS roots with // @ts-check

### LINKED ISSUE #12143: `deno cache` and `deno run` doesn't detect type errors in `.js` files with `// @ts-check`
The `deno cache` and `deno run` commands don't detect type errors in `.js` files with `// @ts-check`.

In `demo.ts`:

```ts
new URL(true);
```

The Deno LSP correctly shows the type error:

Screen Shot 2021-09-20 at 11 49 29 am

Run `deno cache demo.ts`, and the exit code is correctly `1` with this output:

```
Check file:///[redacted]/demo.ts
error: TS2345 [ERROR]: Argument of type 'boolean' is not assignable to parameter of type 'string'.
new URL(true);
        ~~~~
    at file:///[redacted]/demo.ts:1:9
```

But in `demo.js`:

```js
// @ts-check
new URL(true);
```

The Deno LSP correctly shows the type error:

Screen Shot 2021-09-20 at 11 51 53 am

And run `deno cache demo.js`, there is incorrectly no error exit code or output.

`deno run demo.js` does have a `1` exit code but that's because of the expected runtime error, and not the type error:

```
error: Uncaught TypeError: Invalid URL
new URL(true);
^
    at Object.opSync (deno:core/01_core.js:140:12)
    at opUrlParse (deno:ext/url/00_url.js:48:27)
    at new URL (deno:ext/url/00_url.js:324:20)
    at file:///[redacted]/demo.js:2:1
```

`deno cache --help` shows a `--no-check` option, that implies that type checking should be performed if you don't explicitly skip it via the flag:

```
        --no-check                    
            Skip type checking modules
```

This `deno cache` bug makes it impossible to type check `.js` files via the CLI (without running them, which would be bad), see:

https://github.com/denoland/deno/discussions/12113#discussioncomment-1346612

nayeemrmn:
Fixes #12143.


cli/graph_util.rs
```diff
@@ -13,12 +13,19 @@ use deno_graph::ModuleGraphError;
...
 use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::collections::VecDeque;
 use std::sync::Arc;
 
+lazy_static! {
+  /// Matches the `@ts-check` pragma.
+  static ref TS_CHECK_RE: Regex = Regex::new(r#"(?i)^\s*@ts-check(?:\s+|$)"#).unwrap();
+}
```
littledivy, line 27: Use `once_cell::sync::Lazy`

bartlomieju:
I think this is fine and looks good to me, but I'll wait for @kitsonk review before landing.

bartlomieju (approved):
Discussed with @lucacasonato and @ry, let's merge it for v1.21 since it's a fix. We can always fix it again :)

Issue #14088, opened 2022-03-23
`deno bench` does not enforce `--unstable` flag


lucacasonato:
`deno bench` without unstable works, except for the type errors about `Deno.bench` not being found. This is confusing. The command should early exit if `--unstable` is not provided IMO.

PR #13400, opened 2022-01-17
feat: stabilize test steps API

dsherret:
Closes #13055

This stabilizes the test steps API without metadata. Metadata will be implemented as part of a separate effort in #13399.

PR #14091, opened 2022-03-23
fix(bench): require --unstable flag in JavaScript

bartlomieju:
Fixes https://github.com/denoland/deno/issues/14088

### DIFF
--- a/cli/ops/bench.rs
+++ b/cli/ops/bench.rs
@@ -12,7 +12,7 @@ use std::time;
 use tokio::sync::mpsc::UnboundedSender;
 use uuid::Uuid;
 
-pub fn init(sender: UnboundedSender<BenchEvent>) -> Extension {
+pub fn init(sender: UnboundedSender<BenchEvent>, unstable: bool) -> Extension {
   Extension::builder()
     .ops(vec![
       op_pledge_test_permissions::decl(),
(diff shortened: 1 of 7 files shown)


runtime/js/40_testing.js
```diff
@@ -639,6 +639,7 @@
     optionsOrFn,
     maybeFn,
   ) {
+    core.opSync("op_bench_check_unstable");
```
dsherret, line 642: This information is already provided to 99_main.js: https://github.com/denoland/deno/pull/13400/files#diff-00a8b60a7dc00e07ca154218d7c1f478aedfa44a0b6055e558c1463ca1515dbb -- maybe it would be easier to just do something like this instead of introducing another op?

cli/ops/bench.rs
```diff
@@ -12,22 +12,44 @@ use std::time;
...
+    );
+    std::process::exit(70);
+  }
+}
+
+#[op]
+fn op_bench_check_unstable(state: &mut OpState) -> Result<(), AnyError> {
+  check_unstable(state, "Deno.bench");
+  Ok(())
+}
```
dsherret, line 51: Should we make this op more general and move it out of cli/ops/bench.rs?

dsherret (approved):
LGTM

runtime/js/40_testing.js
bartlomieju: I don't agree, seems like adding an op is less code than wiring it up through `99_main.js`

lucacasonato (approved):

Issue #13935, opened 2022-03-13
ci: Rust 1.59.0-1.60.0 compiler error (ICE)


littledivy:
https://github.com/denoland/deno/runs/5528879712?check_suite_focus=true#step:26:453
https://github.com/denoland/deno/runs/5519374644?check_suite_focus=true#step:26:121
https://github.com/denoland/deno/runs/5519410740?check_suite_focus=true#step:26:121
https://github.com/denoland/deno/runs/5477300819?check_suite_focus=true
https://github.com/denoland/deno/runs/5528912995?check_suite_focus=true#step:26:449

Looks like an issue with incremental compilation in `ext/` crates. Tweaking the cache key fixes it.

Related rustc issues:
https://github.com/rust-lang/rust/issues/93779
https://github.com/rust-lang/rust/issues/91704

littledivy:
Still not fixed after updating rustc to 1.60.0 

https://github.com/denoland/deno/runs/6349212182?check_suite_focus=true

crowlKats:
Closing as not the case anymore

PR #14092, opened 2022-03-23
feat(ext/ffi): support structs-by-value

DjDeveloperr:
Adds support for passing structs-by-value as parameters to Foreign Functions, as well as to return them.

Previously, we could only pass/return structs by pointers, which works fine. But it is not suitable where we
are using, for example an existing system library.

In my use case, I'm writing bindings to [Objective C runtime](https://github.com/DjDeveloperr/deno_objc). It mostly works except for struct types which
are passed by value instead of pointer/reference in certain APIs such as in AppKit's NSWindow, certain
methods take NSRect struct by value instead of reference.

Struct types are specified using `{ struct: [...field types] }`, such as `{ struct: ["u16", "i16"] }`.

- When passing structs, we accept either pointer or Typed Arrays.
- When returning structs, `Uint8Array`.
- When using struct type for statics, it behaves same as `pointer` type since structs-by-value are not a thing. Statics are simply just dereferencing the symbol pointer, which would give the struct.


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -319,9 +319,7 @@ declare namespace Deno {
    */
   export function getUid(): number | null;
 
-  /** All possible types for interfacing with foreign functions */
-  export type NativeType =
-    | "void"
+  export type NativeNumberType =
```
aapoalas, line 322: Ah, lovely. I've been thinking about making this change so many times by now.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -333,8 +331,20 @@ declare namespace Deno {
...
+    | "f64";
+
+  /** All possible types for interfacing with foreign functions */
+  export type NativeType =
+    | NativeNumberType
+    | "void"
+    | "pointer"
+    | NativeStructType;
+
+  export interface NativeStructType<
```
aapoalas, line 343: Absolutely 100% totally unimportant nitpick, but maybe put all the constituent parts above the actual `NativeType` type?

aapoalas (approved):
I don't understand everything in the Rust code exactly, but it seems fine to me.

ext/ffi/lib.rs
```diff
@@ -184,6 +184,32 @@ enum NativeType {
...
+    match self {
+      NativeType::Void => 0,
+      NativeType::U8 | NativeType::I8 => 1,
+      NativeType::U16 | NativeType::I16 => 2,
+      NativeType::U32 | NativeType::I32 | NativeType::F32 => 4,
+      NativeType::U64
+      | NativeType::I64
+      | NativeType::F64
+      | NativeType::Pointer => 8,
+      NativeType::USize | NativeType::ISize => std::mem::size_of::<usize>(),
```
littledivy, line 201: This already makes an assumption that size of pointer is 8, which is also the size of `usize`. Maybe merge `Pointer` match with `size_of::<usize>`
```suggestion
      NativeType::USize | NativeType::Pointer | NativeType::ISize => std::mem::size_of::<usize>(),
```

littledivy (commented):
This looks like a great feature. I have a few concerns:

1. structs-by-value are not an FFI "primitive", instead just a bunch of neatly arranged bytes.
2. This passes the ownership of the value to the ffi function. The array buffer backing memory owned by JS is always deallocated by Rust? 
4. Prior art?
5. Is it a blocker? this functionality is taken care of by the `x/byte_type` module + the `pointer ` type?

cli/dts/lib.deno.unstable.d.ts
DjDeveloperr: Makes it easier to specifically handle `NativeStructType` than repeating `{ struct: ... }` everywhere

ext/ffi/lib.rs
DjDeveloperr: Sure, will do. Though currently the pointers API just assumes 64 bit pointers as well 😆

DjDeveloperr:
@littledivy 
> structs-by-value are not an FFI "primitive", instead just a bunch of neatly arranged bytes.

Yep, I could've used a name like `bytes` but decided to match with libffi which calls them `structure`.

> This passes the ownership of the value to the ffi function. The array buffer backing memory owned by JS is always deallocated by Rust?

Hmm did not think of this. I'll test this, and if that is the case we can clone the data.

And yes, this is an already "solved" problem using pointers. But that solution only works for native libraries that _you_
write, not when using system libraries or existing ones. As I have mentioned before, I'm trying to use Objective C runtime
library which exposes C API to interface with ObjC classes. Some methods of those classes take/return structs by value
instead of pointers. One example of such API is [NSWindow initWithContentRect:styleMask:backing:defer:](https://developer.apple.com/documentation/appkit/nswindow/1419477-initwithcontentrect?language=objc).
Here the `contentRect` argument takes an `NSRect` struct by value that basically consists of 4 doubles. So yes, this is a
blocker, at least for my project.

Generally, I avoid writing wrappers for system libraries to add "workarounds" since they add more friction in usage of
a module that makes use of such wrappers.

DjDeveloperr:
Why is the `fastci` failing? It looks unrelated to the changes

aapoalas:
> Why is the `fastci` failing? It looks unrelated to the changes

It's failing when compiling the `FfiLoadArgs` struct which internally uses `NativeType`. This was added to `NativeType` in this PR:
```rs
NativeType::Struct(fields) => libffi::middle::Type::structure(
  fields.iter().map(|field| field.clone().into()),
),
```
Something in `fastci` doesn't like that, I guess?

DjDeveloperr:
Is it really that? This PR builds fine for me, and even tests pass (on Linux). Could it be a [compiler bug as the logs say](https://github.com/denoland/deno/runs/5671688027?check_suite_focus=true#step:25:136)?

littledivy:
It's a rustc ICE - #13935 

A workaround is to updated the CI cache keys. https://github.com/denoland/deno/blob/main/.github/workflows/ci.yml#L239

littledivy:
> Edit: No, it does not deallocate the backing memory. I was able to use the buffer just fine after it returned from an FFI call that used it as a struct. Added a test as well.

@DjDeveloperr That is not true. It does deallocate the struct - V8 now holds the pointer to freed memory, accessing it causes immediate UB. You are just lucky that in the test freed memory is not overridden by future allocations.

DjDeveloperr:
Oh, then maybe we should only allow typed arrays in case of struct arguments so we can clone it before using in FFI call. What do you think?

aapoalas:
Detached buffers sound quite useful for this indeed.

DjDeveloperr:
> #14102 <-- Detached buffers might be useful for owned buffers when it lands.

@littledivy now that they've landed, I'm wondering how can I actually use them for this PR? Seems like `DetachedBuffer` is a struct that wraps `ZeroCopyBuf` and implements custom deserialize behavior so it's detached once passed to Rust. Not sure how is that usable in this context where we only need to detach only certain buffers (which are used for structs).

PR #14095, opened 2022-03-23
cleanup(serde_v8): simpler ZeroCopyBuf

AaronO:


serde_v8/src/magic/zero_copy_buf.rs
```diff
@@ -53,7 +47,17 @@ impl ZeroCopyBuf {
     let buffer = view.buffer(scope).ok_or(v8::DataError::NoData {
       expected: "view to have a buffer",
     })?;
-    Self::from_buffer(buffer, view.byte_offset(), view.byte_length())
+    let (offset, len) = (view.byte_offset(), view.byte_length());
+    Self::from_buffer(buffer, offset..offset + len)
+  }
+
+  fn as_slice(&self) -> &[u8] {
+    unsafe { &*(&self.store[self.range.clone()] as *const _ as *const [u8]) }
```
aapoalas, line 55: I guess the `self.store[self.range]` consumes the range, thus requiring the `self.range.clone()`?

aapoalas (commented):
From my limited Rust-knowledge, LGTM.

serde_v8/src/magic/zero_copy_buf.rs
AaronO: `.clone()` isn't always expensive, in this case it isn't (basically copying 2 usize numbers) we have to explicitly `.clone()` because `Range` doesn't impl `Copy` only `Clone`, in this `Copy` would be ~identical to `Clone` just implicit.

And https://doc.rust-lang.org/std/primitive.slice.html#impl-SliceIndex%3C%5BT%5D%3E-5 expects an owned `Range<usize>`

serde_v8/magic/zero_copy_buf.rs
```diff
@@ -53,7 +47,17 @@ impl ZeroCopyBuf {
...
+    Self::from_buffer(buffer, offset..offset + len)
+  }
+
+  fn as_slice(&self) -> &[u8] {
+    unsafe { &*(&self.store[self.range.clone()] as *const _ as *const [u8]) }
+  }
+
+  #[allow(clippy::cast_ref_to_mut)]
+  fn as_slice_mut(&mut self) -> &mut [u8] {
+    unsafe { &mut *(&self.store[self.range.clone()] as *const _ as *mut [u8]) }
```
littledivy, line 60: `// SAFETY`

bartlomieju (approved):
LGTM, but add a safety comment as pointed by Divy

PR #14096, opened 2022-03-23
experiment(serde_v8): derive_more enabled opaque wrappers

AaronO:
Experiment using [derive_more](https://docs.rs/derive_more/latest/derive_more/) to help generate magic-type wrappers with mostly opaque-ish internal types, so we can swap out Vec for SmallVec for ByteStrings. Creating them via Into/From rather than explicitly wrapping their internal type.

(Note that's just as efficient, `SmallVec`s can be created from Vec's with no copying, it simply transfers ownership of the buffer on the heap)


ext/fetch/lib.rs
```diff
@@ -395,11 +395,7 @@ pub async fn op_fetch_send(
   let url = res.url().to_string();
   let mut res_headers = Vec::new();
   for (key, val) in res.headers().iter() {
-    let key_bytes: &[u8] = key.as_ref();
-    res_headers.push((
-      ByteString(key_bytes.to_owned()),
-      ByteString(val.as_bytes().to_owned()),
-    ));
+    res_headers.push((key.as_str().into(), val.as_bytes().into()));
```
andreubotella, line 398:
```suggestion
    res_headers.push((key.as_ref().into(), val.as_bytes().into()));
```
IIUC with `.as_str()` there's a round-trip of decoding and then encoding back to Latin-1.

AaronO: It does not, I just double-checked and `HeaderName` is an enum of Custom/Standard, Custom values are implement via an internal `ByteStr` type with a `str` deref of:
```rs
impl ops::Deref for ByteStr {
    type Target = str;

    #[inline]
    fn deref(&self) -> &str {
        let b: &[u8] = self.bytes.as_ref();
        // Safety: the invariant of `bytes` is that it contains valid UTF-8.
        unsafe { str::from_utf8_unchecked(b) }
    }
}
```

And Standard headers have `.as_str() -> &'static str` compile-time strings, so `.as_str()` is essentially zero-cost

Also if you look at the `AsRef<[u8]>` impl it calls `.as_str()`:
```rs
impl AsRef<[u8]> for HeaderName {
    fn as_ref(&self) -> &[u8] {
        self.as_str().as_bytes()
    }
}
```

serde_v8/magic/transl8.rs
```diff
@@ -141,3 +141,33 @@ macro_rules! impl_magic {
...
+}
+pub(crate) use impl_wrapper;
+
+// macro_rules! magic_wrapper {
+//   (pub struct $t1:ident($t2:tt)) => {
+//     crate::magic::transl8::impl_wrapper!(pub struct $t1($t2););
+//     crate::magic::transl8::impl_magic!($t1);
+//   };
+// }
+// pub(crate) use magic_wrapper;
```
littledivy, line 173: Remove

littledivy (approved):
LGTM!

PR #14108, opened 2022-03-24
[EXPERIMENT] feat: Add "deno serve" subcommand

bartlomieju:
```ts
// handler.ts
export default async function(req: Request): Promise<Response> {
  return new Response("hello world");
}
```

```sh
$ deno serve --port 5000 handler.ts
Listening on http://0.0.0.0:5000

```

```sh
curl 0.0.0.0:5000
hello world

```


bartlomieju:
Needs more work and consensus. Closing for now

PR #14109, opened 2022-03-24
fix: `Deno.run` - do not modify user provided `cmd` array

### LINKED ISSUE #12455: Allow read-only array for Deno.run
Code:
```ts
const CMD = [/* ... */] as const;

// ...
Deno.run({
    cmd: CMD
});
```
Error:
```
error: Uncaught TypeError: Cannot assign to read only property '0' of object '[object Array]'
Deno.run({
     ^
    at Object.run (deno:runtime/js/40_process.js:112:14)
    at file:///home/kaqw/Projects/magix/build.ts:17:6
```

I don't understand why a read-only array cannot be used here. What I mean is, a read-only array should be valid in this context.

dsherret:
Closes #12455.


kt3k (approved):
LGTM

Issue #13640, opened 2022-02-10
REPL imports should use no-check by default


lucacasonato:

bartlomieju:
I agree, this should speed up REPL significantly

a-lafrance:
I'm interested in giving this change a try if that's alright, although I'm not super familiar with the codebase so I might need some more context about what exactly this change entails. I was poking around the code to try and get more info and I found the `no-check` flag in the CLI, so I guessed that maybe this change is just setting a different default for that flag for the `repl` subcommand.

So concretely I guess I have 2 related questions:
1. Is my understanding correct? I'm under the impression that I'm basically just setting the check flag to either `None` or `Local` as a default for `repl`, so I'd just like to make sure I'm on the right track

2. If that's not quite the right thought process, I'd really appreciate some guidance to get me on the right track.

Thanks in advance!

bartlomieju:
Hi @a-lafrance, thanks for looking into this.

> Is my understanding correct? I'm under the impression that I'm basically just setting the check flag to either None or Local as a default for repl, so I'd just like to make sure I'm on the right track

Yes, this is right. In `repl_parse` in `cli/flags.rs` set `flags.no_check` to `None` and that should be it.

a-lafrance:
great, thanks for the info! I'll try to get a PR up soon

VishnuJin:
> Yes, this is right. In `repl_parse` in `cli/flags.rs` set `flags.no_check` to `None` and that should be it.

jfyi @a-lafrance `flags.check` should be set to `CheckFlag::None`

GJZwiers:
With #14112 landed this could be closed

PR #14112, opened 2022-03-25
feat(repl): Don't type check when importing

a-lafrance:
This (very small) PR implements the change described in #13640: using `no-check` by default when the REPL is run.

Although it's just a one-line change, I do have a couple questions about my (potentially rough-around-the-edges) implementation:
1. I updated the existing `repl` flag tests to verify that when the `--no-check` flag is not supplied, it's set to `None` automatically, but there's no dedicated test case that verifies this behavior. I wasn't sure whether or not it would be good to add an entirely new test for this so I decided not to, but I'm just wondering if that would be a good idea. If desired, I'm happy to add it.

2. A few integration tests (they were all under `integration::compat`) consistently didn't pass when I ran `cargo test`. Although, I also ran them again without the change I made and the same failures seemed to happen, so I'm not sure if it's directly caused by that change. So I guess I'm just looking for a little guidance on what to do about those failures, and whether or not they're relevant to the change I made. I'm happy to provide more info about the failures (e.g. which tests failed and how) if that would help.

Thanks in advance for the review!


bartlomieju:
Thanks @a-lafrance this looks good. I think we should add one more integration test to check that there is no `Check <url>` output; @dsherret what do you think?

a-lafrance:
>  I think we should add one more integration test to check that there is no Check <url> output

Sure, just let me know if you think this would be good and I'll add it

dsherret:
@bartlomieju I think it would be a lot of work to add a test for this because we don't have anything right now in the repl tests to read stderr. I think this PR is ok to merge as-is and if the behaviour regresses it's not a big deal.

bartlomieju:
@dsherret sounds good!

bartlomieju:
Tried locally, I think this is fine 👍

bartlomieju (approved):
LGTM, thank you @a-lafrance

PR #14118, opened 2022-03-25
v1.20.3

lucacasonato:


littledivy (approved):

littledivy (changes requested):
~~Update std version in [cli/compat/esm_resolver.rs](https://github.com/denoland/deno/blob/cff1e92ecf17933a4fcd2ea6a18abfdf89316e37/cli/compat/esm_resolver.rs)~~ you're fast :-)

littledivy (approved):

PR #14119, opened 2022-03-25
feat(ext/crypto): add x25519 and Ed25519 CFRG curves

littledivy:
Implements Curve25519 from https://github.com/WICG/webcrypto-secure-curves

We will probably have to rollout our own Curve448, there is no ring, RustCrypto OR rust-dalek crate for that. Left that for now.


littledivy:
Not stale!

ry (approved):
LGTM 🙌

littledivy:
>  Do you plan for X448 and Ed448 as well?

Yes, planned but we'll have to evaluate existing Rust crates that do curve448 - I couldn't find a 'widely-used' crate. That said it shouldn't be too hard to write our own.

panva:
@littledivy I'm finding it hard to navigate what has been implemented here. I'm able to generate both key types. Despite import code being added for both I can import neither one (Ed says the algorithm is unrecognized, X says the key data is invalid). X export seems to be omitted despite being the same as Ed wrt to the key material. Ed signatures fail to validate.

littledivy:
Yikes! Seems like I didn't push all changes from my local branch :(

panva:
@littledivy I hope it's not lost ;)

PR #14124, opened 2022-03-25
feat(ops): #[op(unstable)]

AaronO:
`#[op(unstable)]` tags `OpDecl::is_unstable` as `true`, thus allowing op-middleware to then disable said ops

This should make unstable checks less error prone (forgotten or overlooked during review/impl), less verbose and moderately faster (check happens at runtime init vs each call)

### DIFF
--- a/ops/lib.rs
+++ b/ops/lib.rs
@@ -34,8 +34,25 @@ fn core_import() -> TokenStream2 {
   }
 }
 
+#[derive(Debug)]
+struct MacroArgs {
+  is_unstable: bool,
+}
+
+impl syn::parse::Parse for MacroArgs {
+  fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+    let vars =
+      syn::punctuated::Punctuated::<Ident, syn::Token![,]>::parse_terminated(
+        input,
+      )?;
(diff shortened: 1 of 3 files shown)


dsherret (approved):
LGTM

ops/lib.rs
```diff
@@ -34,8 +34,25 @@ fn core_import() -> TokenStream2 {
...
+  is_unstable: bool,
+}
+
+impl syn::parse::Parse for MacroArgs {
+  fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+    let vars =
+      syn::punctuated::Punctuated::<Ident, syn::Token![,]>::parse_terminated(
+        input,
+      )?;
+    let is_unstable = vars.iter().any(|v| v == "unstable");
```
lucacasonato, line 48: Can we cause an error for anything that is not `unstable`?

AaronO:
```rs
error: Ops expect #[op] or #[op(unstable)]
  --> core/examples/hello_world.rs:16:1
   |
16 | #[op(wrong)]
   | ^^^^^^^^^^^^
```

littledivy (approved):
lgtm!

PR #14129, opened 2022-03-26
feat: return `WebSocketStream` from `Deno.upgradeWebSocket()`

### LINKED ISSUE #14064: upgradeWebSocket could return a WebSocketStream
Currently, `Deno.upgradeWebSocket` returns a `WebSocket`. It could return a `WebSocketStream`, too, for example:

```typescript
const conn = await Deno.listen({ port: 80 });
const httpConn = Deno.serveHttp(await conn.accept());
const e = await httpConn.nextRequest();
if (e) {
  const { stream, response } = Deno.upgradeWebSocket(e.request);
  const [_, { writable, readable }] = await Promise.all([
    e.respondWith(response),
    stream.connection,
  ]);
  // Use writable and/or readable
}
```

crowlKats:
Closes #14064
Will add tests

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1350,6 +1350,12 @@ declare namespace Deno {
    */
   export function unrefTimer(id: number): void;
 
+  export interface WebSocketUpgrade {
+    response: Response;
+    socket: WebSocket;
+    stream: WebSocketStream;
+  }
+
   /** **UNSTABLE**: new API, yet to be vetter.
    *
    * Allows to "hijack" a connection that the request is associated with.
--- a/ext/http/01_http.js
+++ b/ext/http/01_http.js
@@ -30,6 +30,7 @@
     _idleTimeoutTimeout,
     _serverHandleIdleTimeout,
     _connection,
+    _closed,
     _createWebSocketStreams,
   } = window.__bootstrap.webSocket;
   const { TcpConn, UnixConn } = window.__bootstrap.net;
@@ -460,6 +461,8 @@
     stream[_server] = true;
     stream[_idleTimeoutDuration] = options.idleTimeout ?? 120;
     stream[_idleTimeoutTimeout] = null;
+    stream[_connection] = new Deferred();
+    stream[_closed] = new Deferred();
 
     const webSocketSelector = { kind: null, socket, stream };
     response[_ws] = webSocketSelector;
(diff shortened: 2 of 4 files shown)


ext/http/01_http.js
```diff
@@ -314,30 +316,24 @@
...
-            ws[_readyState] = WebSocket.CLOSED;
-
-            const errEvent = new ErrorEvent("error");
-            ws.dispatchEvent(errEvent);
-
-            const event = new CloseEvent("close");
-            ws.dispatchEvent(event);
-
-            core.tryClose(wsRid);
-          } else {
```
crowlKats, line 340: this whole deletion is due to this branch not even being possible; this was a mistake on my end when first implementing this.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -1350,6 +1350,12 @@ declare namespace Deno {
    */
   export function unrefTimer(id: number): void;
 
+  export interface WebSocketUpgrade {
```
kitsonk, line 1353: Some JSDoc?

kitsonk: Two other points:

- Interfaces are open ended, so it should just be:

  ```ts
  export interface WebSocketUpgrade {
    stream: WebSocketStream;
  }
  ```

- There is no logic (as far as I can tell) that detects the unstable flags and behaves differently, so is there a benefit to including it in unstable, because runtime behaviour will always be to have it.

kitsonk (commented):
I like it, a couple comments...

Plus shouldn't we have a test for the locking behaviour?  There is code there, but I don't see it... and specifically the destructuring case:

```ts
const { response, stream, socket } = Deno.upgradeWebSocket(request);
```

Should be "first one wins" right?  Would be great to prove that.

crowlKats:
> Plus shouldn't we have a test for the locking behaviour?

Yes, the tests arent done yet, I am working on trying to get the one test working before adding more.

cli/dts/lib.deno.unstable.d.ts
crowlKats: > There is no logic (as far as I can tell) that detects the unstable flags and behaves differently

Indeed, I am still thinking how to ideally do this

kitsonk: Personally, I don't see any issue with adding it to the stabilised API.  It is additive, we are unlikely to remove it without a breaking change to the larger API. There is no changes to behaviour to the existing stabilised API.  Unstable provides no benefit except introducing friction to users.

crowlKats: I agree, but also returning an unstable api from a stable api without repercussions at runtime seems subideal to me

PR #14130, opened 2022-03-27
fix(runtime/permissions): Resolve executable specifiers in allowlists and queries

### LINKED ISSUE #14122: Proposal: resolve executable paths in Deno.run for more fine grained allow-run permissions
## The security flaw

consider two malicious scripts. They could be published by a malicious library author, or possibly man-in-the-middle'd while installing/upgrading a deno script:

<details>

<summary>malicious-lib1.ts</summary>

```ts
// deno run --allow-env=HOME --allow-run=ffprobe --allow-write malicious-lib1.ts
import * as path  from 'https://deno.land/std/path/mod.ts'
import { writeAll } from "https://deno.land/std/streams/conversion.ts";


const malicious_ffprobe = `
#!/bin/bash

echo "I can do whatever I please."
`

// find an executable path with non-sudo permissions
const known_executable_dir = path.join(Deno.env.get('HOME')!, '.deno/bin')
const filepath = path.join(known_executable_dir, 'ffprobe')
const file = await Deno.open(filepath, { createNew: true, write: true, mode: 0o755 })
  .catch(e => {
    if (e instanceof Deno.errors.AlreadyExists) return Deno.open(filepath, { write: true })
    else throw e
  })
await Deno.truncate(filepath, 0)
await Deno.writeAll(file, new TextEncoder().encode(malicious_ffprobe))
Deno.close(file.rid);

// we now have access to an executable that does whatever we want
const proc = Deno.run({ cmd: ['ffprobe'] })
await proc.status()
```
</details>

<details>

<summary>malicious-lib2.ts</summary>

```ts
// deno run --allow-env=PATH --allow-run=ffprobe --allow-write=.
import * as path  from 'https://deno.land/std/path/mod.ts'
import { writeAll } from "https://deno.land/std/streams/conversion.ts";


const malicious_ffprobe = `
#!/bin/bash

echo "I can do anything I want."
`

// modify the PATH envar to include our local executables folder we will be making
Deno.env.set('PATH', '.local-executables:' + Deno.env.get('PATH'))
await Deno.mkdir('.local-executables', {recursive:true})
const filepath = path.join('.local-executables', 'ffprobe')
const file = await Deno.open(filepath, { createNew: true, write: true, mode: 0o755 })
  .catch(e => {
    if (e instanceof Deno.errors.AlreadyExists) return Deno.open(filepath, { write: true })
    else throw e
  })
await Deno.truncate(filepath, 0)
await Deno.writeAll(file, new TextEncoder().encode(malicious_ffprobe))
Deno.close(file.rid);

// we now have access to an executable that does whatever we want
const proc = Deno.run({ cmd: ['ffprobe'] })
await proc.status()
```

</details>

The first script works by writing a malicious executable into a known executable dir (that doesnt require elevated permissions to write to). The second script works by adding a new directory to the `PATH` environment, which deno reads directly from to determine where an executable originates. In either case, we give the library some pretty inconspicuous permissions:
- `deno run --allow-env=HOME --allow-run=ffprobe --allow-write` (`--allow-env=HOME` isnt even technically necessary)
- `deno run --allow-env=PATH --allow-run=ffprobe --allow-write=.`

## Possible workarounds

They both take advantage of the fact that an executable can be resolved from several locations along the `PATH` envar. To remedy this, I could specify the _exact_ location of an executable that I am allowing access to in my deno run command. However, the library author (who is not writing malicious code!) cant exactly write portable code to help with that:
```ts
// How can I be sure this is in the same location cross platform?
// How can I be sure a user doesnt have a custom location for this executable?
const proc = Deno.run({ cmd: ['/usr/local/bin/ffprobe'] })
await proc.status()
```

## Proposed solution

So, I am proposing that permissions on executables can be resolved before checking if they match the allowed permissions. E.g.
```bash
deno run --allow-run=/usr/local/bin/ffprobe
```
and the following code should work:
```ts
// ffprobe is checked against the allow-run permissions as both ffprobe _and_ what the path resolves it to (e.g. /usr/local/bin/ffprobe)
const proc = Deno.run({ cmd: ['ffprobe'] })
await proc.status()
```

### [additional info]
I just picked `ffprobe` at random as an example, any executable could be used here. This just happens to be one that I use in deno occasionally, and wouldn't really be considered a security issue if a readme says to allow it.

nayeemrmn:
Closes #14122.

Adds two extensions to `--allow-run` behaviour:
- When `--allow-run=foo` is specified and `foo` is found in the `PATH` at startup, `RunDescriptor::Path(which("foo"))` is added to the allowlist alongside `RunDescriptor::Name("foo")`. Currently only the latter is.
- When run permission for `foo` is queried and `foo` is found in the `PATH` at runtime, either `RunDescriptor::Path(which("foo"))` or `RunDescriptor::Name("foo")` would qualify in the allowlist. Currently only the latter does.

This results in the following outcomes for permission checks:
| `deno run --allow-run=echo`              | `which echo == "/usr/bin/echo"` at startup | `which echo != "/usr/bin/echo"` at startup |
| ---------------------------------------- | ------------------------------------------ | ------------------------------------------ |
| **`Deno.run({ cmd: "echo" })`**          | ✅                                          | ✅                                          |
| **`Deno.run({ cmd: "/usr/bin/echo" })`** | ✅                                          | ❌                                          |

| `deno run --allow-run=/usr/bin/echo`     | `which echo == "/usr/bin/echo"` at runtime | `which echo != "/usr/bin/echo"` at runtime |
| ---------------------------------------- | ------------------------------------------ | ------------------------------------------ |
| **`Deno.run({ cmd: "echo" })`**          | ✅                                          | ❌                                          |
| **`Deno.run({ cmd: "/usr/bin/echo" })`** | ✅                                          | ✅                                          |

<details>
<summary>Formerly:</summary>

| `deno run --allow-run=echo`              | `which echo == "/usr/bin/echo"` at startup | `which echo != "/usr/bin/echo"` at startup |
| ---------------------------------------- | ------------------------------------------ | ------------------------------------------ |
| **`Deno.run({ cmd: "echo" })`**          | ✅                                          | ✅                                          |
| **`Deno.run({ cmd: "/usr/bin/echo" })`** | ❌                                          | ❌                                          |

| `deno run --allow-run=/usr/bin/echo`     | `which echo == "/usr/bin/echo"` at runtime | `which echo != "/usr/bin/echo"` at runtime |
| ---------------------------------------- | ------------------------------------------ | ------------------------------------------ |
| **`Deno.run({ cmd: "echo" })`**          | ❌                                          | ❌                                          |
| **`Deno.run({ cmd: "/usr/bin/echo" })`** | ✅                                          | ✅                                          |
</details>


runtime/permissions.rs
```diff
@@ -27,6 +27,24 @@ use std::string::ToString;
...
 use std::sync::atomic::Ordering;
+#[cfg(not(windows))]
+use which::which;
+
+// `which::which` is bugged on Windows, we patch it here.
+// https://github.com/harryfei/which-rs/issues/54
+#[cfg(windows)]
+use which::which as which_;
+#[cfg(windows)]
+use which::Result as WhichResult;
```
ry, line 38: Can't these be moved inside the function to reduce the number of conditionals?

nayeemrmn: One is in the signature

bartlomieju (commented):
@nayeemrmn can you open a PR to `denoland/manual` (https://deno.land/manual@v1.20.5/getting_started/permissions) to describe this change?

runtime/permissions.rs
```diff
@@ -27,6 +27,23 @@ use std::string::ToString;
...
+// `which::which` is bugged on Windows, we patch it here.
+// https://github.com/harryfei/which-rs/issues/54
+#[cfg(windows)]
+use which::Result as WhichResult;
+#[cfg(windows)]
+fn which(name: &str) -> WhichResult<PathBuf> {
+  use which::which as which_;
+  which_(name).map(|mut p| {
+    if p.extension().and_then(std::ffi::OsStr::to_str) == Some("EXE") {
+      p.set_extension("exe");
```
dsherret, line 42: Paths are case insensitive on windows so we should remove this and fix whatever comparison is being case sensitive to be case insensitive on Windows (and Mac?)

Also, it's not just exe files that are executable on Windows, but whatever is in the PATHEXT environment variable. It seems this dependency handles that for us luckily.

dsherret (changes requested):
Looks good, but we should resolve the comparison that is not being case insensitive on Windows.

nayeemrmn:
@dsherret Ready for review

aapoalas:
Can you rebase and resolve conflicts on this?

aapoalas:
Needs yet another rebase :(

aapoalas (approved):
Looks good to me! Let's finally get this merged!

nayeemrmn:
@aapoalas Actually I'd like to redo it above the refactors in #19070, some abstractions are introduced there that make this better

aapoalas:
Ah, okay. That's fine too of course. Should we close this in that case?

nayeemrmn:
I'll just rebase it, write-up and tests are still applicable

nayeemrmn:
@aapoalas Please review the current implementation

aapoalas (approved):
Still looks good to me and the change seems like a reasonable and wanted addition to the safety and usability of `--allow-run` (when used with explicit paths).

PR #14138, opened 2022-03-28
fix(cli/fmt): ignore .git when formatting files

### LINKED ISSUE #14054: deno fmt crash formatting invalid json file in `.git` subdirectory
Should it auto ignore .git?
```
# deno fmt
memory allocation of Error formatting: 18446744073709551605/Users/ry/src/dotland/.git/logs/refs/remotes/origin/update_versions.json bytes failed

   Line 1, column 2: Text cannot contain more than one JSON value

  0000000000000
   ~~
Abort trap: 6
```

TrickyPi:
Closes #14054;

### DIFF
--- a/cli/tools/fmt.rs
+++ b/cli/tools/fmt.rs
@@ -26,6 +26,7 @@ use deno_core::error::AnyError;
 use deno_core::futures;
 use log::debug;
 use log::info;
+use regex::Regex;
 use std::borrow::Cow;
 use std::fs;
 use std::io::stdin;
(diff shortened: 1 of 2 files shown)


cli/tools/fmt.rs
```diff
@@ -624,6 +625,14 @@ fn is_supported_ext_fmt(path: &Path) -> bool {
   }
 }
 
+fn is_located_in_git(path: &Path) -> bool {
+  let re = Regex::new(r"(^|[\\/])\.git[\\/]").unwrap();
```
dsherret, line 629: I think we can just do `path.components().any(|c| matches!(c, Some(Component::Normal(OsStr::new(".git"))))` or something like that instead of using Regex, which will be slower (especially compiling regex each time).

dsherret (changes requested):
THanks @TrickyPi! LGTM other than the regex. We could use `path.components()` instead.

Also, this doesn't close that issue because there's a memory allocation issue in it.

cli/tools/fmt.rs
TrickyPi: Ok, i get it. Thanks for your review.

cli/tools/fmt.rs
```diff
@@ -87,31 +87,32 @@ pub async fn format(
...
-                .any(|path| paths.contains(path))
-                .then(|| files)
-                .unwrap_or_else(|| [].to_vec())
-            } else {
-              files
-                .into_iter()
-                .filter(|path| paths.contains(path))
-                .collect::<Vec<_>>()
-            }
+    let result = collect_files(&include_files, &exclude_files, fmt_predicate)
```
ah-yu, line 96: Why not just put `.git` into `exclude_files`?

dsherret (approved):
LGTM. Thanks @TrickyPi!

Issue #13910, opened 2022-03-11
lsp: support tasks


dsherret:
It would be useful for the lsp to support running tasks in from deno.json

cc @kitsonk

kitsonk:
We should consider if we want the lsp to run the tasks, in the same process.  We do "caching" and will be doing it for the testing API, but it would require some custom APIs to run the tasks and deal with the output (like writing to stdout or reading from stdin would interfere with communication of the client/server).

Wouldn't it be better to add the feature to the clients directly.  Each client has a concept of tasks, and for example for vscode_deno, there is a task provider API which allows tasks to be expressed and made available in the editor.  The biggest thing would be the "auto discovery" of the config file.  We could either mirror the logic in the extension or we could have the lsp notify the client of the discovered config file if any.

Having that one custom notification is a lot better than a handful of requests/responses and notifications for managing it.

Following it this way, the extensions would be responsible for launching and managing the `deno task` as a subprocess.  The API in vscode is pretty rich.  It already has integrations to auto-discover npm scripts, and would be a good pattern for how to deal with Deno tasks.

PR #14139, opened 2022-03-28
feat(lsp): support API for config file

kitsonk:
Closes: #13910

This add the `denoConfigTasks` experimental capability which handles a `deno/task` request from the client, which responds with an array of tasks from the current configuration file.

Clients need to implement surfacing them to the user, such as with denoland/vscode_deno#641.

### DIFF
--- a/cli/tests/integration/lsp_tests.rs
+++ b/cli/tests/integration/lsp_tests.rs
@@ -596,7 +596,7 @@ fn lsp_deno_task() {
   )
   .unwrap();
 
-  params.root_uri = Some(Url::from_file_path(workspace_root.clone()).unwrap());
+  params.root_uri = Some(Url::from_file_path(workspace_root).unwrap());
 
   let deno_exe = deno_exe_path();
   let mut client = LspClient::new(&deno_exe).unwrap();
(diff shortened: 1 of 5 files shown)


cli/tests/integration/lsp_tests.rs
```diff
@@ -579,6 +579,51 @@ fn lsp_import_map_config_file() {
...
+  assert!(maybe_err.is_none());
+  assert_eq!(
+    maybe_res,
+    Some(json!([
+      {
+        "name": "build",
+        "detail": "deno test"
+      },
+      {
+        "name": "some%test",
```
bartlomieju, line 620: I think this is not right, we don't support `%` in task name - only alphanumeric, colons, underscores and dashes

kitsonk: I just grabbed another fixture from another test without reading the test.  Interesting that it doesn't error. Parsed the tasks perfectly fine.

bartlomieju (approved):
LGTM

cli/tests/integration/lsp_tests.rs
bartlomieju: It parses fine, but @dsherret and I decided to play conservative for now and disallow most characters.

PR #14142, opened 2022-03-28
chore: fix compile_windows_ext test

### LINKED ISSUE #14103: test issue - compile_windows_ext
In #14078 a previously ignored test was unignored (`integration::compile:: compile_windows_ext`), and CI passed, but it appears the reason why it was ignored is because when testing a release build the test attempts to download a binary that doesn't exist yet, so CI passes _except_ when it is a release PR, when the test fails.

We need to figure out a way to have the test download the previous build.

dsherret:
Moves what this test was originally designed for to a lower level unit test.

Closes #14103


bartlomieju (approved):
Seems reasonable, LGTM 👍

Issue #13998, opened 2022-03-17
error on warnings


ry:
This simple mistake has been in the code base since September last year (d93570a6195d2be5fe448bb0d33d4f64e839676e)
https://github.com/denoland/deno/blob/92bb25105822d0def65aa89ee5141c862ac7a3e4/cli/tools/lint.rs#L552-L554

That suggests we are not failing on warnings as I thought would be happening from this bit of CI script:

https://github.com/denoland/deno/blob/92bb25105822d0def65aa89ee5141c862ac7a3e4/.github/workflows/ci.yml#L150-L153

dsherret:
I'm not able to reproduce this error locally on the old main. I wonder if there's some additional flag we need to set.

bartlomieju:
This has been fixed. The original bug was that `anyhow!("No rules have been configured")` was constructed but never returned. The lint code has since been refactored into `cli/tools/lint/`, and both call sites now use `bail!(...)`, which constructs and returns the error correctly:

- `cli/tools/lint/mod.rs:339`
- `cli/tools/lint/rules/mod.rs:160`

Closing as completed.

PR #14146, opened 2022-03-29
chore: upgrade dprint-core to 0.54.1

dsherret:
Upgrades to the new API. This actually has some perf improvements and a bug fix that won't matter to most people.

### DIFF
--- a/cli/lsp/language_server.rs
+++ b/cli/lsp/language_server.rs
@@ -1063,11 +1063,13 @@ impl Inner {
       };
 
       match format_result {
-        Ok(new_text) => Some(text::get_edits(
-          document.content().as_str(),
-          &new_text,
-          document.line_index().as_ref(),
-        )),
+        Ok(new_text) => new_text.map(|new_text| {
+          text::get_edits(
+            document.content().as_str(),
+            &new_text,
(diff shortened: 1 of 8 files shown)


test_util/src/lib.rs
```diff
@@ -369,9 +369,7 @@ async fn get_tls_config(
           ),
         )
         .with_single_cert(certs, PrivateKey(key))
-        .map_err(|e| {
-          anyhow!("Error setting cert: {:?}", e);
```
dsherret, line 373: https://github.com/denoland/deno/issues/13998

bartlomieju (approved):
LGTM

cli/lsp/language_server.rs
```diff
@@ -1063,11 +1063,13 @@ impl Inner {
...
-          &new_text,
-          document.line_index().as_ref(),
-        )),
+        Ok(new_text) => new_text.map(|new_text| {
+          text::get_edits(
+            document.content().as_str(),
+            &new_text,
+            document.line_index().as_ref(),
+          )
+        }),
```
dsherret, line 1072: I caused a bug here because I misread the code. Will provide a fix soon.

PR #14153, opened 2022-03-30
chore(tests): use custom temp dir creation for the tests

dsherret:
This makes the tests way faster on Windows (at least on my machine). My WSL seems unaffected at 1m 46s on main.

Before: 11m 51s
After: 2m 26s

Also, fixes various temp directories that weren't being cleaned up.


bartlomieju (approved):
This adds quite some complexity, but if it speeds up tests significantly I think we can land it.

PR #14159, opened 2022-03-30
feat(ext/web): Add error events for event listener and timer errors

nayeemrmn:
```
feat(ext/web): Add error events for event listener and timer errors

- feat: Add handleable error event for even listener errors
- feat: Add handleable error event for setTimeout()/setInterval() errors
- feat: Add Deno.core.destructureError()
- feat: Add Deno.core.terminate()
- fix: Don't throw listener errors from dispatchEvent()
- fix: Use biased mode when selecting between mod_evaluate() and
  run_event_loop() results
```


nayeemrmn:
@lucacasonato @bartlomieju Ready for review

ext/web/02_event.js
```diff
@@ -787,7 +788,11 @@
 
     setCurrentTarget(eventImpl, tuple.item);
 
-    innerInvokeEventListeners(eventImpl, getListeners(tuple.item));
+    try {
+      innerInvokeEventListeners(eventImpl, getListeners(tuple.item));
+    } catch (error) {
+      __bootstrap.reportError.reportException(error);
```
lucacasonato, line 794: This is related to load order I guess? Annoying :/

ext/web/lib.deno_web.d.ts
```diff
@@ -890,3 +890,22 @@ declare class DecompressionStream {
...
+ * addEventListener("error", (event) => {
+ *   event.preventDefault();
+ * });
+ * reportError(new Error("foo")); // Will not be reported.
+ * ```
+ * In Deno, this error will terminate the process if not intercepted like above.
+ */
+declare function reportError(
+  error: any,
+): void;
```
lucacasonato, line 911: Not part of this PR yet.

lucacasonato (commented):
This looks great! Just some minor comments.

cli/fmt_errors.rs
```diff
@@ -48,8 +48,9 @@ pub fn format_location(frame: &JsStackFrame) -> String {
     return cyan("native").to_string();
   }
   let mut result = String::new();
-  if let Some(file_name) = &frame.file_name {
-    result += &cyan(&format_file_name(file_name)).to_string();
+  let file_name = frame.file_name.clone().unwrap_or_default();
+  if !file_name.is_empty() {
+    result += &cyan(&format_file_name(&file_name)).to_string();
```
bartlomieju, line 53: Do we have a test case for empty file name?

ext/web/02_event.js
bartlomieju: Yeah I don't really understand this bit. Why do we need to store whole `__bootstrap` namespace? `const reportError = window.__bootstrap.reportError;` should be enough.

bartlomieju: Please add a comment here explaining this logic

ext/web/02_timers.js
```diff
@@ -139,13 +140,16 @@
...
-            ...new SafeArrayIterator(args),
-          );
+          try {
+            FunctionPrototypeCall(
+              callback,
+              globalThis,
+              ...new SafeArrayIterator(args),
+            );
+          } catch (error) {
+            __bootstrap.reportError.reportException(error);
```
bartlomieju, line 151: Ditto

runtime/js/99_main.js
```diff
@@ -589,6 +590,10 @@ delete Object.prototype.__proto__;
     numCpus = cpuCount;
     registerErrors();
 
+    reportError.setPrintException((jsError) =>
+      core.opSync("op_print_exception", jsError)
+    );
```
bartlomieju, line 595: This could use a comment too

cli/fmt_errors.rs
nayeemrmn: I noticed it happen in the repl when `reportError()` is called (`throw new Error()` doesn't show pretty errors in the repl). Should be part of the other PR really. I'll take it out.

runtime/js/99_main.js
nayeemrmn: This is just a dependency injection similar to the ones strewn above it. Is there something unintuitive about it? If so I would put that above the `setPrintException()` declaration.

cli/fmt_errors.rs
bartlomieju: Ah that's true... Scripts compiled in REPL don't have a filename and I found no way to supply it (AFAIK Node also has this quirk)

runtime/js/99_main.js
bartlomieju: I think a comment stating that this is hit only for uncaught errors coming from workers would be sufficient (unless that's not true in which case a comment would be even more appreciated :D)

nayeemrmn: It's hit whenever an `"error"` event is dispatched by the runtime and `e.preventDefault()` isn't called. I added a comment above the function.

ext/web/16_report_error.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+    // Avoid recursing `reportException()` via error handlers more than once.
+    if (reportExceptionStackedCalls > 1 || window.dispatchEvent(event)) {
+      printException?.(jsError);
+      // TODO(nayeemrmn): Use `queueMicrotask()` instead once it's fixed
+      // (https://github.com/denoland/deno/issues/14158). Largely because the
+      // `(in promise)` isn't desired.
+      PromisePrototypeThen(PromiseResolve(), () => {
+        throw new Error(`Unhandled error event.`);
+      });
```
bartlomieju, line 67: Discussed with @lucacasonato and we don't really get what's going on here. 

It feels like we should not call `printException` and just directly call `throw error`.

See:
```js
// foo.js
globalThis.addEventListener("error", (e) => {
    console.log("foo");
    e.preventDefault();
})

setTimeout(() => {
    throw new Error("boom!")
}, 1000);
```

```
❯ cargo run -- run foo.js
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/deno run foo.js`
foo2

deno on  error-event-for-events-and-timers [$!?] via  v17.4.0 via 🦀 v1.59.0 took 2s
❯ echo $status
0

```

Removing `e.preventDefault()` from the event listener yields:
```
❯ cargo run -- run foo.js
   Compiling deno_runtime v0.53.0 (/Users/ib/dev/deno/runtime)
   Compiling deno v1.20.4 (/Users/ib/dev/deno/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 23.51s
     Running `target/debug/deno run foo.js`
foo2
error: Uncaught Error: boom!
    throw new Error("boom!")
          ^
    at file:///Users/ib/dev/deno/foo.js:10:11
    at Object.action (deno:ext/web/02_timers.js:145:13)
    at handleTimerMacrotask (deno:ext/web/02_timers.js:62:12)

deno on  error-event-for-events-and-timers [$!?] via  v17.4.0 via 🦀 v1.59.0 took 26s
❯ echo $status
1

```

Please explain the reasoning for current solution

nayeemrmn: > It feels like we should not call `printException` and just directly call `throw error`.

It's so the error is printed synchronously as it is in browsers, and the actual error leading to termination is detached. I especially wanted this for `reportError()`.

lucacasonato: > It's so the error is printed synchronously as it is in browsers

Wouldn't `throw error` also achieve this?

> and the actual error leading to termination is detached

Why is this important? It seems rather annoying to get two errors printed if something goes wrong.

nayeemrmn: > Wouldn't `throw error` also achieve this?

Without a microtask that makes it catchable around `dispatchEvent(new CustomEvent("leads-to-error"))` and `reportError()` which is undesired. The other option is throwing the main error in the microtask, should I just do that?

nayeemrmn:
> `globalThis.onerror = (e) => { ... }` is not working (but `globalThis.addEventListener("error", () => {});` is working), I think we should address it in this PR before landing.

Addressed

ext/web/16_report_error.js
lucacasonato: Ah I see... I think we might want to tackle this differently. What about adding a core binding in the form `core.terminate(error)` that saves the error onto the isolate state, then triggers an isolate termination ([`Isolate::terminate_execution`](https://docs.rs/v8/latest/v8/struct.Isolate.html#method.terminate_execution)), which is essentially an uncatchable error. Then once this termination bubbles up to the place in `deno_core` where we'd return an error back to the Rust caller we grab the error from the isolate state and return that. Does that make sense?

ext/web/16_report_error.js
```diff
@@ -69,6 +70,8 @@
     reportExceptionStackedCalls--;
   }
 
+  defineEventHandler(window, "error");
```
lucacasonato, line 73: Can you add a little test for this too?

lucacasonato (commented):
Tests fail for me locally due to string truncation in console.log :D

```
OUTPUT
1
{
  cancelable: true,
  message: "Uncaught Error: bar",
  filename: "file:///mnt/starship/Projects/github.com/denoland/deno/cli/tests/testdata/event_listener_error_handl...",
  lineno: 18,
  colno: 9,
  error: Error: bar
    at file:///mnt/starship/Projects/github.com/denoland/deno/cli/tests/testdata/event_listener_error_handled.ts:18:9
    at innerInvokeEventListeners (deno:ext/web/02_event.js:753:9)
    at invokeEventListeners (deno:ext/web/02_event.js:793:7)
    at dispatch (deno:ext/web/02_event.js:662:11)
    at dispatchEvent (deno:ext/web/02_event.js:1005:14)
    at file:///mnt/starship/Projects/github.com/denoland/deno/cli/tests/testdata/event_listener_error_handled.ts:22:1
}
onerror() called Error: bar
    at file:///mnt/starship/Projects/github.com/denoland/deno/cli/tests/testdata/event_listener_error_handled.ts:18:9
    at innerInvokeEventListeners (deno:ext/web/02_event.js:753:9)
    at invokeEventListeners (deno:ext/web/02_event.js:793:7)
    at dispatch (deno:ext/web/02_event.js:662:11)
    at dispatchEvent (deno:ext/web/02_event.js:1005:14)
    at file:///mnt/starship/Projects/github.com/denoland/deno/cli/tests/testdata/event_listener_error_handled.ts:22:1
2

OUTPUT
EXPECTED
1
{
  cancelable: true,
  message: "Uncaught Error: bar",
  filename: "[WILDCARD]/event_listener_error_handled.ts",
  lineno: 18,
  colno: 9,
  error: Error: bar
    at [WILDCARD]/event_listener_error_handled.ts:18:9
    at [WILDCARD]
    at [WILDCARD]/event_listener_error_handled.ts:22:1
}
onerror() called Error: bar
    at [WILDCARD]/event_listener_error_handled.ts:18:9
    at [WILDCARD]
    at [WILDCARD]/event_listener_error_handled.ts:22:1
2

EXPECTED
```

See `filename`

lucacasonato:
I think this is looking great now. I think the place you slotted in the retrieval of `explicit_terminate_exception` is pretty good. I have pushed one little change that renames some stuff, and changes the isolate state slot to store the raw exception rather than a `JsError`. That last thing aligns this just a tad bit more to the regular error handling paths.

nayeemrmn:
> That last thing aligns this just a tad bit more to the regular error handling paths.

I wanted to avoid the `cancel_terminate_execution(); re_terminate_execution()` in this code path because I didn't know what consequences it would have. Looks like it causes us to re-enter `exception_to_err_result` and get a generic termination error.

> changes the isolate state slot to store the raw exception rather than a `JsError`

I tried that previously and got a cryptic panic at the `JsError::from_v8_exception()`. I assumed it had to do with the different scopes or something.

lucacasonato:
> I wanted to avoid the cancel_terminate_execution(); re_terminate_execution() in this code path because I didn't know what consequences it would have. Looks like it sometimes causes us to re-enter exception_to_err_result and get a generic termination error.

That is the strangest thing! I just tried it locally, and indeed! How strange! I will debug this further. It is very strange.

lucacasonato:
I have figured it out:

- the first invocation of `exception_to_err_result` happens in `mod_evaluate`.
- the second invocation of `exception_to_err_result` happens in `JsRuntime::poll_event_loop` (called by `JsRuntime::run_event_loop`).
- `mod_evaluate` is called in parallel with `JsRuntime::run_event_loop`.
- Because of poll order, the error from `JsRuntime::run_event_loop` can sometimes be returned instead of the one from `mod_evaluate()`.

This could be worked around using the following:

```
diff --git a/core/runtime.rs b/core/runtime.rs
index 2427a6f1a..8409cf53e 100644
--- a/core/runtime.rs
+++ b/core/runtime.rs
@@ -1254,11 +1254,10 @@ impl JsRuntime {
       let has_explicit_terminate_error =
         state_rc.borrow().explicit_terminate_exception.is_some();
       if has_explicit_terminate_error {
-        let undefined = v8::undefined(tc_scope);
-        // Pass `undefined`, it will be substituted with
-        // `explicit_terminate_error` within `exception_to_err_result()`.
+        // If there is an explicit terminate error, we won't return it here but
+        // instead let it bubble up through `JsRuntime::poll_event_loop`.
         sender
-          .send(exception_to_err_result(tc_scope, undefined.into(), false))
+          .send(Ok(()))
           .expect("Failed to send module evaluation error.");
       } else {
         sender.send(Err(
```

I am not sure if the current code in this PR is susceptible to this issue too.

lucacasonato:
Unfortunately, there are issues. Run this 10 times and see it vary in output:

```js
addEventListener("foo", () => {
  queueMicrotask(() => console.log("queueMicrotask"));
  setTimeout(() => console.log("timer"), 0);
  throw new Error("bar");
});
console.log(1);
Deno.core.setNextTickCallback(() => console.log("nextTick"));
Deno.core.setHasTickScheduled(true);
dispatchEvent(new CustomEvent("foo"));
console.log(2);
```

nayeemrmn:
> Unfortunately, there are issues. Run this 10 times and see it vary in output:
> 
> ```js
> addEventListener("foo", () => {
>   queueMicrotask(() => console.log("queueMicrotask"));
>   setTimeout(() => console.log("timer"), 0);
>   throw new Error("bar");
> });
> console.log(1);
> Deno.core.setNextTickCallback(() => console.log("nextTick"));
> Deno.core.setHasTickScheduled(true);
> dispatchEvent(new CustomEvent("foo"));
> console.log(2);
> ```

This is a pre-existing race condition, right?

nayeemrmn:
@lucacasonato https://github.com/denoland/deno/pull/14159/commits/4e8df5c3e3470a5da67fe244fdf015529e377046 most likely removes any non-deterministic behaviour that isn't warranted (no IO etc). Without that you sometimes get a single `poll_event_loop()` call even though the module evaluated synchronously.

core/runtime.rs
```diff
@@ -1213,7 +1226,17 @@ impl JsRuntime {
...
+      state_rc.borrow().explicit_terminate_error.is_some();
+
+    if has_explicit_terminate_error {
+      let undefined = v8::undefined(tc_scope);
+      // Pass `undefined`, it will be substituted with
+      // `explicit_terminate_error` within `exception_to_err_result()`.
+      sender
+        .send(exception_to_err_result(tc_scope, undefined.into(), false))
+        .expect("Failed to send module evaluation error.");
+    } else if let Some(value) = maybe_value {
```
bartlomieju, line 1239: While this works, why not take the `explicit_terminate_error` here and pass it explicitly instead of taking it in `exception_to_error_result`?

ext/web/02_timers.js
```diff
@@ -2,6 +2,7 @@
 "use strict";
 
 ((window) => {
+  const __bootstrap = window.__bootstrap;
```
bartlomieju, line 5: Remove

bartlomieju (commented):
Just a couple nit picks

bartlomieju (commented):
@nayeemrmn while this now looks good to land, I feel we should discuss it with the rest of the team before actually landing. This PR adds a way to prevent termination of program on uncaught error, which was not possible before (it's similar to adding `window.onunhandledrejection`, which is quite controversial). Hope that's okay

core/runtime.rs
nayeemrmn: I need it to be available inside of the fn to avoid the code path outside of `if let Some() = explicit_terminate_error.take()`. Also I got a weird serde_v8 panic if I didn't convert `exception -> js_error` ahead of time.

bartlomieju: Okay, that's fine. Could you add a comment about it? Seems like a potential future refactor

lucacasonato (commented):
Discussed with everyone, got no objections to land. Thanks @nayeemrmn!

Just one last thing before I land this: can you add an itest case for the scenario I had described above with the flaky output?

bartlomieju:
@nayeemrmn please draft a commit message

nayeemrmn:
> @nayeemrmn please draft a commit message

Added to PR description

lucacasonato (approved):
LGTM! Great that you managed to get my cleanup working with that biased mode select.

Thanks so much for this, and sorry for the long arduous review!

Issue #13704, opened 2022-02-17
`deno compile`/eszip does not follow redirects


mweichert:
Since v1.19.0 was released, we're seeing a problem when running `deno compile` in Github Actions.

See: https://github.com/ecadlabs/taqueria/runs/5234787765?check_suite_focus=true

We were using:
```
- name: Setup Deno
        uses: denoland/setup-deno@v1
        with:
          deno-version: "^1.0.0"
```

When I hardcode the version to 1.18.2 everything works fine:

https://github.com/ecadlabs/taqueria/runs/5236615612?check_suite_focus=true

littledivy:
The issue is because of the implicit import to `https://deno.land/x/i18next/index.js`. Versioning your import should fix this.

```
divy@mini ~/g/taqueria (main)> ../deno/target/debug/deno compile ./index.ts 
Download https://deno.land/x/i18next/index.js
Warning Implicitly using latest version (v21.6.11) for https://deno.land/x/i18next/index.js
Download https://deno.land/x/i18next@v21.6.11/index.js
Download https://deno.land/x/i18next@v21.6.11/i18next.js
Download https://deno.land/x/i18next@v21.6.11/BackendConnector.js
Download https://deno.land/x/i18next@v21.6.11/EventEmitter.js
Download https://deno.land/x/i18next@v21.6.11/Formatter.js
Download https://deno.land/x/i18next@v21.6.11/Interpolator.js
Download https://deno.land/x/i18next@v21.6.11/LanguageUtils.js
Download https://deno.land/x/i18next@v21.6.11/PluralResolver.js
Download https://deno.land/x/i18next@v21.6.11/ResourceStore.js
Download https://deno.land/x/i18next@v21.6.11/Translator.js
Download https://deno.land/x/i18next@v21.6.11/defaults.js
Download https://deno.land/x/i18next@v21.6.11/logger.js
Download https://deno.land/x/i18next@v21.6.11/postProcessor.js
Download https://deno.land/x/i18next@v21.6.11/utils.js
Check file:///Users/divy/gh/taqueria/index.ts
Compile file:///Users/divy/gh/taqueria/index.ts
Emit taqueria
divy@mini ~/g/taqueria (main)> ./taqueria 
error: Module not found: https://deno.land/x/i18next/i18next.js
```

This is a bug because `deno compile` does not redirect to the versioned import.

p-fruck:
[...]

PR #14161, opened 2022-03-31
fix(compile): follow redirects when resolving

littledivy:
Fixes https://github.com/denoland/deno/issues/13704


cli/tests/testdata/standalone_follow_redirects_2.js
```diff
@@ -0,0 +1,5 @@
+// unversioned import redirects with depenedencies.
```
littledivy, line 1:
```suggestion
// unversioned import redirects with dependencies.
```

piscisaureus (approved):
LGTM!

Issue #14056, opened 2022-03-21
lsp: migrate from `lspower` back to `tower-lsp`


kitsonk:
`lspower` was a fork that fixed a lot of things with `tower-lsp`, but not `lspower` has been remerged back into `tower-lsp` and deprecated. [`tower-lsp`](https://crates.io/crates/tower-lsp) is being maintained going forward, so we should/need to remigrate back.

kitsonk:
(thanks @sno2 for [pointing it out](https://github.com/denoland/vscode_deno/issues/509#issuecomment-1073409917))

AbhijithGanesh:
is this a good first issue for a new contributor?

kitsonk:
Not entirely sure... It could be very straight forward, but it could get quite tricky if it doesn't work. If someone were to try to do it and the tests didn't pass, it could get quite complicated to figure out, but would be glad to try to help if someone wanted to try.

AbhijithGanesh:
Does this have a pre-commit script with which I can locally test the changes? I am up for trying it and I can reach out on discord if I have doubts. Would that work?

AbhijithGanesh:
@CGQAQ If I want to add in my changes, can I push it  to your PR?

CGQAQ:
@AbhijithGanesh Sorry but I am almost done the refactoring, just lack of tests fix 😢

AbhijithGanesh:
No issues , just wanted to know. :)

PR #14059, opened 2022-03-21
refactor(cli/lsp): migrate from lspower back to tower-lsp

CGQAQ:
closed: #14056

PR #14163, opened 2022-03-31
refactor(cli/lsp): migrate from lspower back to tower-lsp

CGQAQ:
Rework of #14059 

closed: #14056


Cargo.lock
```diff
@@ -689,16 +689,6 @@ dependencies = [
  "syn 1.0.85",
 ]
 
-[[package]]
```
kitsonk, line 692: So this removes a duplicated "dash-map" but upgrade us to tokio 1.17 and introduces a duplication of tokio-util because a lot of other packages don't have updates there either.  @bartlomieju thoughts?

Cargo.lock
```diff
@@ -5050,6 +5069,12 @@ version = "0.10.0+wasi-snapshot-preview1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a143597ca7c7793eff794def352d41792a93c481eb1042423ff7ff72ba2c31f"
 
+[[package]]
```
kitsonk, line 5072: It also adds a 3rd version of wasi... 😢

cli/lsp/documents.rs
```diff
@@ -10,11 +10,10 @@ use crate::file_fetcher::get_source_from_bytes;
 use crate::file_fetcher::map_content_type;
 use crate::file_fetcher::SUPPORTED_SCHEMES;
 use crate::fs_util::specifier_to_file_path;
-use crate::http_cache;
 use crate::http_cache::HttpCache;
 use crate::resolver::ImportMapResolver;
 use crate::resolver::JsxResolver;
-use crate::text_encoding;
+use crate::{http_cache, text_encoding};
```
kitsonk, line 16:
```suggestion
use crate::http_cache;
use crate::text_encoding;
```

kitsonk: Our style is to _always_ split and not group.

kitsonk (commented):
Thank you, looking good.  One style comment, and I would really like to get @bartlomieju feedback on the upgrade to tokio.

Cargo.lock
CGQAQ: tower-lsp 0.16.0 requires upgrade of the tokio and the libc 😢

Cargo.lock
CGQAQ: 😢

Cargo.lock
kitsonk: Yes, it isn't a criticism... just want @bartlomieju thoughts on it.

bartlomieju: I think it's okay. I can upgrade these dependencies in a separate PR before landing this okay. I'll get back to you once this is done.

bartlomieju: Okay I did update the dependencies. @CGQAQ please rebase against `main`

kitsonk (approved):
LGTM, thank you for the contribution, especially what was a lot churn and rebasing.

PR #14177, opened 2022-04-01
chore(ext/url): cleanup url ops

littledivy:
Get's rid of serde derives and the unwind handler that's no longer needed. No significant improvement in `new URL` performance. 

EDIT: `op_url_reparse` is 10% faster.

### DIFF
--- a/ext/url/lib.rs
+++ b/ext/url/lib.rs
@@ -2,7 +2,6 @@
 
 mod urlpattern;
 
-use deno_core::error::generic_error;
 use deno_core::error::type_error;
 use deno_core::error::uri_error;
 use deno_core::error::AnyError;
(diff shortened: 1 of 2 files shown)


ext/url/lib.rs
```diff
@@ -120,43 +120,24 @@ pub fn op_url_reparse(
...
-      quirks::password(&url),
-      quirks::pathname(&url),
-      quirks::port(&url),
-      quirks::protocol(&url),
-      quirks::search(&url),
-      username,
-    ]
-    .join("\n"),
-  )
+fn url_result(url: Url) -> UrlParts {
```
nayeemrmn, line 126: nit: rename to `url_parts`?

bartlomieju (approved):

PR #14180, opened 2022-04-01
chore(ci): automatically open PR to forward patch release back to main

dsherret:
After the cargo publish runs and the release tag is created, this will automatically open a PR that forwards the patch release commit back to main. In extreme scenarios, we can just ignore the PR that's opened and do it manually.


tools/release/04_post_publish.ts
```diff
@@ -0,0 +1,94 @@
...
+await createReleaseTag();
+
+console.log("Forwarding release commit to main...");
+try {
+  await forwardReleaseCommitToMain();
+} catch (err) {
+  console.error("Failed. Please manually open a PR.", err);
+}
+
+async function createReleaseTag() {
```
dsherret, line 20: This function is old and from 04_create_release_tag.ts

tools/release/04_post_publish.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  if (tags.has(tagName)) {
+    console.log(`Tag ${tagName} already exists.`);
+  } else {
+    await repo.gitTag(tagName);
+    await repo.gitPush(tagName);
+  }
+}
+
+async function forwardReleaseCommitToMain() {
```
dsherret, line 33: This function is new. Please review it carefully as it's untested, though won't be the end of the world if it fails because it's the last step.

tools/release/04_post_publish.ts
```diff
@@ -0,0 +1,94 @@
...
+  // if this is a patch release, open a PR to forward the most recent commit back to main
+  const currentBranch = await repo.gitCurrentBranch();
+  const isPatchRelease = currentBranch !== "main";
```
bartlomieju, line 36: I don't understand this comparison, care to explain?

bartlomieju (approved):

tools/release/04_post_publish.ts
dsherret: The current branch will be something like `v1.20` for patch releases and `main` for minor releases, correct?

bartlomieju: Ah yes, sounds good

Issue #14144, opened 2022-03-28
`op_close` missing from test sanitizer messages


jespertheend:
I ran into an error while running a test:
```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at asyncOpSanitizer (deno:runtime/js/40_testing.js:201:26)
    at async resourceSanitizer (deno:runtime/js/40_testing.js:370:7)
    at async Object.exitSanitizer [as fn] (deno:runtime/js/40_testing.js:427:9)
    at async runTest (deno:runtime/js/40_testing.js:788:7)
    at async Object.runTests (deno:runtime/js/40_testing.js:986:22)
```

I haven't been able to create a reduced test case yet but when debugging it seems it tries to get the message for `op_close`, which doesn't exist:
https://github.com/denoland/deno/blob/89dd5dac6219f9a4c04ada4b5a9c812a88c1c2d4/runtime/js/40_testing.js#L61-L64

bartlomieju:
@jespertheend what is your Deno version?

PR #14074, opened 2022-03-22
fix(test): don't error on missing op details

bartlomieju:
Closes https://github.com/denoland/deno/issues/13938
Closes https://github.com/denoland/deno/issues/13729

PR #14184, opened 2022-04-02
fix(test): Don't error on missing op details

AEtheve:
Fixes: #14144


kt3k:
Can you add test cases?

runtime/js/40_testing.js
```diff
@@ -198,15 +198,16 @@
           }
           ArrayPrototypePush(details, message);
         } else if (dispatchedDiff < completedDiff) {
-          const [name] = OP_DETAILS[key];
+          const [name, hint] = OP_DETAILS[key] || [key, null];
```
kt3k, line 201: This exposes the name of the op like the below error message:

```
- 1 async operation to op_write was started before this test, but was completed during the test. Async operations should not complete in a test if they were not started in that test.
```

I think the proper fix here is to add the entry for `op_write` in `OP_DETAILS` table.

bartlomieju: This is actually fine, I thought I fixed it in https://github.com/denoland/deno/pull/14074 but I missed this

bartlomieju: You are right though that we should fill in details for not listed ops.

bartlomieju (approved):
LGTM

Issue #14186, opened 2022-04-02
test: Print file paths with a separate color


jespertheend:
from https://github.com/denoland/deno/discussions/14156:
> In the list of test results, print the `running n tests from file:///path/to/file.test.js` with a different color. That way when you scroll up you can immediately distinguish tests that come from one specific file.

![image](https://user-images.githubusercontent.com/2737650/161396870-066655f6-f152-4955-82a9-16ba97b6bd88.png)

If this is added, maybe https://github.com/denoland/deno/pull/13930 isn't even needed. With lots of tests it would make the output slightly longer but that's not a huge issue.

bartlomieju:
I'm willing to consider this instead of https://github.com/denoland/deno/pull/13930 as it poses some challenges in formatting output for "test steps" API. This would be a very quick change. @dsherret thoughts?

PR #13930, opened 2022-03-13
feat(test): add empty line between files

bartlomieju:
This one has always bothered me. Readability +10

Before:
```
running 2 tests from file:///Users/ib/dev/deno/cli/tests/unit/webgpu_test.ts
test webgpuComputePass ... ok (50ms)
test webgpuHelloTriangle ... ok (159ms)
running 1 test from file:///Users/ib/dev/deno/cli/tests/unit/websocket_test.ts
test websocketPermissionless ... ok (7ms)
running 1 test from file:///Users/ib/dev/deno/cli/tests/unit/worker_permissions_test.ts
test workerEnvArrayPermissions ... ok (18ms)

```

After:
```
running 2 tests from file:///Users/ib/dev/deno/cli/tests/unit/webgpu_test.ts
test webgpuComputePass ... ok (50ms)
test webgpuHelloTriangle ... ok (159ms)

running 1 test from file:///Users/ib/dev/deno/cli/tests/unit/websocket_test.ts
test websocketPermissionless ... ok (7ms)

running 1 test from file:///Users/ib/dev/deno/cli/tests/unit/worker_permissions_test.ts
test workerEnvArrayPermissions ... ok (18ms)

```

Closes https://github.com/denoland/deno/issues/13914

PR #14191, opened 2022-04-02
feat(test): use gray color for 'running n tests from ...'

bartlomieju:
Closes https://github.com/denoland/deno/issues/14186

Alternative to https://github.com/denoland/deno/pull/13930

Screenshot 2022-04-02 at 21 02 44
Screenshot 2022-04-02 at 21 00 16

I'm tempted to do the same for `test` on each lines before the test name

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -231,7 +231,7 @@ impl PrettyTestReporter {
   }
 
   fn force_report_wait(&mut self, description: &TestDescription) {
-    print!("test {} ...", description.name);
+    print!("{} ...", description.name);
     // flush for faster feedback when line buffered
     std::io::stdout().flush().unwrap();
     self.last_wait_output_level = 0;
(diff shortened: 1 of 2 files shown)


bartlomieju:
Thoughts?

Screenshot 2022-04-02 at 21 05 00

bartlomieju:
Actually I'm tempted to completely remove `test` prefix 🤣

jespertheend:
> Actually I'm tempted to completely remove `test` prefix 🤣

I was about to suggest this. I'm not sure what `test` is supposed to tell us, I think we know they are tests xD.

bartlomieju:
Some more ideas:

Screenshot_2022-04-02_at_21 55 07

jespertheend:
Ooh I like it. Especially that there's no need for a list of failures like this anymore, and the `>` makes it super clear where you have to look for the implementation of a failing test.

Not entirely sure how I feel about the list of stats yet. When you have failing tests I think the new list is better, but my current test results look like this:
image
Which would end up as a long list of all zeros.
Maybe you can omit the stats that have zero items?
Or maybe even something like `All 756 tests passed (14s)` when there are no other items.

bartlomieju:
> Ooh I like it. Especially that there's no need for a list of failures like this anymore, and the `>` makes it super clear where you have to look for the implementation of a failing test.
> 
> 
> 
> Not entirely sure how I feel about the list of stats yet. When you have failing tests I think the new list is better, but my current test results look like this:
> 
> image
> 
> Which would end up as a long list of all zeros.
> 
> Maybe you can omit the stats that have zero items?
> 
> Or maybe even something like `All 756 tests passed (14s)` when there are no other items.

I'm not sure I'm in favor of varying output depending on the outcome of the tests. I'll give it another thought tomorrow.

jespertheend:
Hmm yeah maybe it's not entirely fair to compare the two screenshots. One screenshot shows a failing test and the other only passing tests. So naturally the later will seem more satisfying.

I do think the new layout is a lot more readable though, even if you don't end up omitting the stats with zero items.

bartlomieju:
This is blocked by https://github.com/denoland/deno/pull/13799 which contains API that will come in handy for test runner.

bartlomieju:
I'm gonna close this PR for now. Worked on this with @kt3k last week and we got a few more ideas. I will open a new PR once https://github.com/denoland/deno/pull/14159 lands

PR #14198, opened 2022-04-04
chore(ext/fetch): custom arity

littledivy:
## `deno_fetch`

LLVM lines:

Before:
```
48968 (100%)  2038 (100%)  (TOTAL)
```

After:
```
48163 (100%)  2018 (100%)  (TOTAL)
```

### DIFF
--- a/ext/fetch/lib.rs
+++ b/ext/fetch/lib.rs
@@ -39,7 +39,6 @@ use reqwest::Client;
 use reqwest::Method;
 use reqwest::RequestBuilder;
 use reqwest::Response;
-use serde::Deserialize;
 use serde::Serialize;
 use std::borrow::Cow;
 use std::cell::RefCell;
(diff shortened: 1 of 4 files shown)


ext/fetch/lib.rs
```diff
@@ -519,37 +511,29 @@ impl HttpClientResource {
   }
 }
 
-#[derive(Deserialize, Debug)]
-#[serde(rename_all = "camelCase")]
-pub struct CreateHttpClientOptions {
```
AaronO, line 524: Not too convinced we should flatten this struct of options, since it could conceivably be extended and it's not on a hot path

littledivy: reverted

ry (commented):
LLVM lines = lines of LLVM IR? How do you measure that?

crowlKats:
> LLVM lines = lines of LLVM IR? How do you measure that?

there is [cargo-llvm-lines](https://github.com/dtolnay/cargo-llvm-lines)

ext/fetch/lib.rs
```diff
@@ -197,22 +185,27 @@ pub struct FetchReturn {
...
 pub fn op_fetch<FP>(
   state: &mut OpState,
-  args: FetchArgs,
+  method: ByteString,
+  url: String,
+  headers: Vec<(ByteString, ByteString)>,
+  client_rid: Option<u32>,
+  has_body: bool,
+  body_length: Option<u64>,
   data: Option<ZeroCopyBuf>,
```
bartlomieju, line 194: That's a lot of arguments :P I'm not convinced this is preferable to having an "option bag", but if it results in fewer LLVM lines (and potentially faster compile) then we should go with it.

PR #14204, opened 2022-04-04
perf(serde_v8): smallvec ByteString

AaronO:
Avoiding heap allocs when deserializing ByteStrings with less than 16 chars

This comes at no memory overhead, since both `Vec<u8>` and `SmallVec<[u8; 16]>` (w/ union) are 24 bytes

Before:
```
test de_bstr_v8_1024_b    ... bench:          50 ns/iter (+/- 3)
test de_bstr_v8_12_b      ... bench:          35 ns/iter (+/- 1)
```

After:
```
test de_bstr_v8_1024_b    ... bench:          50 ns/iter (+/- 2)
test de_bstr_v8_12_b      ... bench:          15 ns/iter (+/- 0)
```

## TODO

- [ ] Fix some From/Into conversions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3667,6 +3667,7 @@ dependencies = [
  "serde",
  "serde_bytes",
  "serde_json",
+ "smallvec",
  "v8",
 ]
 
--- a/serde_v8/Cargo.toml
+++ b/serde_v8/Cargo.toml
@@ -16,6 +16,7 @@ path = "lib.rs"
 derive_more = "0.99.17"
 serde = { version = "1.0.130", features = ["derive"] }
 serde_bytes = "0.11"
+smallvec = { version = "1.8", features = ["union"] }
 v8 = "0.41.0"
 
 [dev-dependencies]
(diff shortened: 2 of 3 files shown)


andreubotella:
Can we assert that `Vec<u8>` and `SmallVec<[u8; 16]>` have the same size?

AaronO:
> > Can we assert that Vec and SmallVec<[u8; 16]> have the same size?
> 
> `assert_eq!(sizeof::<Vec<u8>>(), sizeof::<SmallVec<[u8; 16]>>())`

They only have the same size when using `smallvec`'s [`features = ["union"]`](https://docs.rs/smallvec/latest/smallvec/#union)

> When the union feature is enabled smallvec will track its state (inline or spilled) without the use of an enum tag, reducing the size of the smallvec by one machine word. This means that there is potentially no space overhead compared to Vec. **Note that smallvec can still be larger than Vec if the inline buffer is larger than two machine words.**

This isn't a safety or correctness guarantee, just a matter of mem efficiency that's not super obscure and documented by `smallvec`

PR #14205, opened 2022-04-04
TRY perf(core): avoid borrows / multiple `get_slot` calls in poll_event_loop

littledivy:
Async ops benchmark:

`main`:
```
op_void_async:       	n = 1000000, dt = 0.279s, r = 3584229/s, t = 279ns/op
read_128k:           	n = 50000, dt = 7.833s, r = 6383/s, t = 156660ns/op
```

This PR:
```
op_void_async:       	n = 1000000, dt = 0.258s, r = 3875969/s, t = 258ns/o
read_128k:           	n = 50000, dt = 7.164s, r = 6979/s, t = 143280ns/op
```

## note

opening pr for visibility, not super happy with improvements. Need to try a few more things.

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -189,7 +189,7 @@ macro_rules! include_js_files {
         Box::new(|| {
           let c = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
           let path = c.join($file);
-          println!("cargo:rerun-if-changed={}", path.display());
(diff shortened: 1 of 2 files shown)


littledivy (commented):
_sigh_ `read_128k` is more often slower with this patch. I'm guessing rustc optimizes away consecutive borrows, also this is not exactly a very hot path and most of the time is spent scheduling async ops. Closing.

PR #14209, opened 2022-04-05
feat: Add DENO_NO_PROMPT variable

### LINKED ISSUE #14208: `DENO_NO_PROMPT` env var support
Env var alternative to `--no-prompt`. An env var is suitable for this because it changes interactivity level but runs the same workloads. Use case: One could have this set in their dev environment if they prefer to back out and add permission flags should they forget the first time, and the prompt-by-default has been growingly annoying for them :-)

nayeemrmn:
Closes #14208.


bartlomieju:
@ry please review. Nayeem's rationale from the issue makes sense, but I don't have strong opinion either way

cli/flags.rs
```diff
@@ -2592,7 +2595,7 @@ fn permission_args_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
     flags.allow_ffi = Some(vec![]);
     flags.allow_hrtime = true;
   }
-  if matches.is_present("no-prompt") {
+  if env::var("DENO_NO_PROMPT").is_ok() || matches.is_present("no-prompt") {
```
bartlomieju, line 2598: I think this reading of env variable should be done in `ProcState::build`; otherwise the tests in `flags.rs` will depend on the current environment.

crowlKats: this imo should be rather written with https://docs.rs/clap/latest/clap/struct.Arg.html#method.env, though @bartlomieju does have a point...

nayeemrmn: I like @crowlKats's solution since that's idiomatic. Is there a way to configure the env for `cargo test` where it can be unset?

cli/flags.rs
```diff
@@ -12,6 +12,7 @@ use deno_runtime::permissions::PermissionsOptions;
 use log::debug;
 use log::Level;
 use once_cell::sync::Lazy;
+use std::env;
```
bartlomieju, line 15: Why is this import needed?

cli/flags.rs
```diff
@@ -1760,12 +1763,13 @@ fn permission_args(app: Command) -> Command {
...
-    .arg(
-      Arg::new("no-prompt")
-        .long("no-prompt")
-        .help("Always throw if required permission wasn't passed"),
-    )
+    ));
+  let no_prompt_arg = Arg::new("no-prompt")
+    .long("no-prompt")
+    .help("Always throw if required permission wasn't passed");
+  #[cfg(not(test))]
```
ry, line 1770: Why `cfg(not(test)` ?

ry:
LGTM - but the two comments should be addressed.

cli/flags.rs
bartlomieju: So that tests of `flags` module do not depend on the environmental variables - see https://github.com/denoland/deno/pull/14209#discussion_r846864313

nayeemrmn: So the unit tests in this module aren't affected by the env var for local dev

bartlomieju:
Please open a PR to https://github.com/denoland/manual and add this env variable to "Setup your environment chapter"

nayeemrmn:
> Does `DENO_NO_PROMPT=0` cause it to be enabled or disabled?

Enabled same as `DENO_NO_PROMPT=1`

bartlomieju:
> > Does `DENO_NO_PROMPT=0` cause it to be enabled or disabled?
> 
> Enabled same as `DENO_NO_PROMPT=1`

That's inconsistent with `DENO_FUTURE_CHECK` which only enables when the value is `1`

cli/flags.rs
lucacasonato: See https://github.com/denoland/deno/pull/12769#discussion_r749553155

lucacasonato:
@nayeemrmn See https://github.com/denoland/deno/pull/12769#discussion_r749553155

nayeemrmn:
Ready for review

bartlomieju (approved):
LGTM

Issue #12698, opened 2021-11-08
Internally specify file size ahead of time for `Deno.writeFile` and similar functions


dsherret:
The `Deno.writeFile` and similar functions could specify up front to the OS the file size before writing the bytes as noted here: https://devblogs.microsoft.com/dotnet/file-io-improvements-in-dotnet-6/#preallocation-size

> On Unix, it’s mapped to posix_fallocate() or fcntl(F_PREALLOCATE) or fcntl(F_ALLOCSP). On Windows, to SetFileInformationByHandle(FILE_ALLOCATION_INFO).

We may also want to expose this to the user in a similar way that .net does so the user could also take advantage of it.

dsherret:
Luca and I investigated this earlier today as part of #14221. I don't see any noticable performance improvement by specifying this ahead of time on Windows with a 10mb file or even a 100mb file for these functions. In fact it seems kind of slower. Luca also didn't notice any performance improvement on Linux.

```
function benchWriteFile() {
  const input = new TextEncoder().encode("long-string".repeat(999_999)); // 10mb file
  const tempDir = Deno.makeTempDirSync();
  let i = 0;
  benchSync("write_file_large", 1e4, () => {
    Deno.writeFileSync(`${tempDir}/${i++}.txt`, input);
  });
  Deno.removeSync(tempDir, { recursive: true });
}
```

Before (10MB file 10,000 iterations):

```
> ./target/release/deno_before.exe run -A --unstable ./cli/bench/deno_common.js
write_file_large:       n = 10000, dt = 60.078s, r = 166/s, t = 6007800ns/op
```

After:

```
> ./target/release/deno.exe run -A --unstable ./cli/bench/deno_common.js
write_file_large:       n = 10000, dt = 61.710s, r = 162/s, t = 6171000ns/op
```

Before (100MB file 100 iterations):

```
> ./target/release/deno_before.exe run -A --unstable ./cli/bench/deno_common.js
write_file_large:       n = 100, dt = 7.197s, r = 14/s, t = 71970000ns/op
```

After:

```
> ./target/release/deno.exe run -A --unstable ./cli/bench/deno_common.js
write_file_large:       n = 100, dt = 7.415s, r = 13/s, t = 74150000ns/op
```

Code for preallocate is here:

```rs
fn preallocate(file: &std::fs::File, size: usize) -> Result<(), AnyError> {
  #[cfg(target_os = "linux")]
  {
    // Preallocate disk space using fallocate
[...]

PR #14221, opened 2022-04-06
perf: move Deno.writeTextFile and like functions to Rust

dsherret:
Luca and I are investigated https://github.com/denoland/deno/issues/12698 as part of pair programming and found it does not help for this scenario. I believe it would only help for buffered writes.

Anyway, we ended up moving this code to Rust. I ran the benchmarks on sync and async a bunch of times and the average was about 100ms faster on my machine.

I'm not sure it's worth it to add these benchmarks because it really depends on the reliability of your drive and it uses a lot of space. I ran about 10 runs and took the average.

```ts
// in cli/bench/deno_common.js
async function benchWriteFile() {
  const input = new TextEncoder().encode("long-string".repeat(999_999));
  const tempDir = Deno.makeTempDirSync();
  let i = 0;
  await benchAsync("write_file_large", 100, async () => {
    await Deno.writeFile(`${tempDir}/${i++}.txt`, input);
  });
  Deno.removeSync(tempDir, { recursive: true });
}

function benchWriteFileSync() {
  const input = new TextEncoder().encode("long-string".repeat(999_999));
  const tempDir = Deno.makeTempDirSync();
  let i = 0;
  benchSync("write_file_large", 100, () => {
    Deno.writeFileSync(`${tempDir}/${i++}.txt`, input);
  });
  Deno.removeSync(tempDir, { recursive: true });
}
```


dsherret:
Looks like a few unix only tests failed. I will look into it tomorrow.

cli/tests/unit/write_file_test.ts
```diff
@@ -315,8 +315,7 @@ Deno.test(
       assert(e instanceof Error);
       assertEquals(e.name, "AbortError");
     }
-    const stat = Deno.statSync(filename);
-    assertEquals(stat.size, 0);
+    assertNotExists(filename);
```
dsherret, line 318: Seems better to not create the file at all if it's cancelled beforehand?

cli/tests/unit/write_file_test.ts
```diff
@@ -356,7 +354,15 @@ Deno.test(
...
   },
 );
+
+function assertNotExists(filename: string | URL) {
+  try {
+    Deno.statSync(filename);
+    throw new Error(`The file ${filename} exists.`);
```
dsherret, line 364: Oh, this assertion is wrong lol

runtime/ops/fs.rs
```diff
@@ -177,18 +184,110 @@ async fn op_open_async(
...
+  }
+
+  let write_future = tokio_file.write_all(&data);
+  if let Some(cancel_handle) = cancel_handle {
+    write_future.or_cancel(cancel_handle).await??;
+  } else {
+    write_future.await?;
+  }
+  Ok(())
+}
```
dsherret, line 289: I don't understand why this is flaky now https://github.com/denoland/deno/runs/5994069319?check_suite_focus=true

cli/tests/unit/write_file_test.ts
```diff
@@ -256,8 +256,8 @@ Deno.test(
       assert(e instanceof Error);
       assertEquals(e.name, "AbortError");
     }
-    const stat = Deno.statSync(filename);
-    assertEquals(stat.size, 0);
+    // it's racy to check the file length, so just check that the file exists
+    assertExists(filename);
```
lucacasonato, line 260: This isn't guaranteed to happen either I think. This may be flaky.

runtime/js/40_write_file.js
```diff
@@ -1,85 +1,53 @@
...
-      build.os !== "windows"
-    ) {
-      await chmod(path, options.mode);
+    let cancelRid;
+    let abortHandler;
+    if (options.signal) {
+      options.signal.throwIfAborted();
+      cancelRid = core.opSync("op_cancel_handle");
+      abortHandler = options.signal
+        .addEventListener("abort", () => core.tryClose(cancelRid));
```
lucacasonato, line 33: Instead of using an event listener, it should register an internal "abort step" on an `AbortSignal`. The event could be cancelled by the user, and is not synchronous.

lucacasonato (commented):
Great!

cli/tests/unit/write_file_test.ts
dsherret: Yeah, I removed these checks. What's asserted now is probably good enough.

lucacasonato (approved):
LGTM!

runtime/js/40_write_file.js
```diff
@@ -1,85 +1,54 @@
...
-    while (nwritten < data.length) {
-      nwritten += file.writeSync(TypedArrayPrototypeSubarray(data, nwritten));
-    }
-
-    file.close();
+    core.opSync("op_write_file_sync", {
+      path: pathFromURL(path),
+      data,
+      mode: options.mode,
+      append: options.append ?? false,
```
aapoalas, line 18: The JS-side options logic has somewhat changed from previous.

```ts
writeFileSync("foo", "bar", { append: "" });
```
would call `openSync("foo", { write: true, create: true, truncate: true })`, ie. append would take the `false` value.

Now the same call will lead to `core.opSync("op_write_file_sync", { path: "foo", data: "bar", append: "" })` call. I'm not sure what `serde` will do with that but on the JS side at least the logic is definitely changed.

I'm not sure if it's desired but I would've personally expected these lines to be:
```ts
append: options.append === true,
create: options.create === true,
```

runtime/ops/fs.rs
```diff
@@ -177,18 +183,112 @@ async fn op_open_async(
...
+  } else {
+    write_future.await??;
+  }
+  Ok(())
+}
+
+fn write_file(
+  path: &Path,
+  open_options: std::fs::OpenOptions,
+  _open_args: &OpenArgs,
```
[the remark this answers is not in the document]
dsherret: It's used below for Unix, but not Windows. The `_` at the beginning is to suppress the clippy error on Windows.

runtime/js/40_write_file.js
dsherret: There's lots of stuff in the API that will fail if people don't provide the correct types. I think if someone is providing unexpected values for these then it erroring is probably more ideal than it working in a maybe unexpected way. For example, previously if someone provided `null` for `create` it was the same thing as providing `false`, which is probably unexpected given that `undefined` doesn't work that way.

PR #14222, opened 2022-04-06
fix(ext/crypto):  check extractable in exportKey

### LINKED ISSUE #14215: Non-extractable keys are extractable
I'm not sure if I'm doing something wrong, if extractable keys are not implemented yet, or if this is a bug. However, if I create a key with `crypto.subtle.generateKey` with `extractable` set to `false`, I can still call `crypto.subtle.exportKey`. [According to the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/exportKey), this should throw an exception.

```
import { encode as he } from 'https://deno.land/std@0.133.0/encoding/hex.ts';
const td = (d: Uint8Array) => new TextDecoder().decode(d);

const keys = await crypto.subtle.generateKey({ name: 'ECDSA', namedCurve: 'P-256' }, false, [
  'sign',
  'verify',
]);

const key = await crypto.subtle.exportKey('pkcs8', keys.privateKey);
const hexKey = td(he(new Uint8Array(key)));
console.log(hexKey);
``` 
When I run the code in a browser (Chrome) it throws an exception `DOMException: key is not extractable`.

I'm running Deno 1.20.4 🦖 on MacOS.

EduM22:
fixes #14215

### DIFF
--- a/ext/crypto/00_crypto.js
+++ b/ext/crypto/00_crypto.js
@@ -978,6 +978,13 @@
         context: "Argument 2",
       });
 
+      if (!key.extractable) {
+        throw new DOMException(
+          "Key is not extractable",
+          "InvalidAccessError",
+        );
+      }
+
       const handle = key[_handle];
       // 2.
       const innerKey = WeakMapPrototypeGet(KEY_STORE, handle);
--- a/ext/crypto/00_crypto.js
+++ b/ext/crypto/00_crypto.js
@@ -978,41 +978,49 @@
         context: "Argument 2",
       });
 
-      if (!key.extractable) {
-        throw new DOMException(
-          "Key is not extractable",
-          "InvalidAccessError",
-        );
-      }
-
       const handle = key[_handle];
       // 2.
       const innerKey = WeakMapPrototypeGet(KEY_STORE, handle);
 
       const algorithmName = key[_algorithm].name;
 
+      let result;
+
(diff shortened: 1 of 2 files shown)


ext/crypto/00_crypto.js
```diff
@@ -978,6 +978,13 @@
         context: "Argument 2",
       });
 
+      if (!key.extractable) {
```
littledivy, line 981:
```suggestion
      if (key.extractable === false) {
```

Also, this step comes after the `switch`. See https://w3c.github.io/webcrypto/#SubtleCrypto-method-exportKey

littledivy (changes requested):
Thanks! I'm surprised the WPT didn't catch this earlier.

ext/crypto/00_crypto.js
EduM22: Okay, so do you want me to implement it so that it is after the switch?

littledivy: yes

egfx-notifications: I do not quite get why we do not throw early as @EduM22 did in their first approach. I can not see why this check should happen after the execution of the actual key extraction from the link you shared. Can you explain?

littledivy (approved):
LGTM

PR #14223, opened 2022-04-06
fix(cli/install): preserve compat flag

SyrupThinker:


dsherret (approved):
LGTM. Thanks @SyrupThinker!

PR #14224, opened 2022-04-06
chore(core): Get rid of op ID's

andreubotella:
After the ops refactor, op ID's are only used for metrics anymore. However, the fact that there is a JS-managed `opIds` object that doesn't get updated with ops that are added after building the V8 snapshot is a problem for tsc, which uses such late op loading, although this only seems to be a problem for some build configurations and some versions of V8 (see https://github.com/denoland/rusty_v8/pull/895#issuecomment-1090603520). In any case, this change is an experiment in getting rid of op ID's, and it solves that issue.

TODO: Changing `OpsTracker` to be a `HashMap` whose keys are strings might regress performance. Consider using some map with immutable entries but mutable values, and/or atomic/interned strings.

### DIFF
--- a/cli/tests/testdata/workers/test.ts
+++ b/cli/tests/testdata/workers/test.ts
@@ -655,7 +655,7 @@ Deno.test("Worker with invalid permission arg", function () {
         deno: { permissions: { env: "foo" } },
       }),
     TypeError,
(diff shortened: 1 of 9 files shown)


andreubotella:
Closing in favor of #14228.

PR #14228, opened 2022-04-07
refactor(core): OpCtx

AaronO:
Avoiding need for js-visible OpIDs, alternative to https://github.com/denoland/deno/pull/14224

The core idea is that each op is provided with a single pointer providing its relevant rust-side context, what it points to may change (e.g: we might get rid of explicit op_ids and track OpMetrics in the ctx) and how we pass it will likely evolve from using the callback's data to a JS-side External 1st arg for Fast API callables

Moderate improvement of `op_baseline`:

```
Before:
test bench_is_proxy   ... bench:      17,695 ns/iter (+/- 487)
test bench_op_async   ... bench:     277,137 ns/iter (+/- 10,404)
test bench_op_nop     ... bench:      19,595 ns/iter (+/- 1,267)
test bench_op_pi_json ... bench:      23,339 ns/iter (+/- 1,257)

After:
test bench_is_proxy   ... bench:      17,738 ns/iter (+/- 953)
test bench_op_async   ... bench:     273,845 ns/iter (+/- 7,598)
test bench_op_nop     ... bench:      18,535 ns/iter (+/- 794)
test bench_op_pi_json ... bench:      23,068 ns/iter (+/- 1,499)
```

## TODO

- [x] Fix up metrics (getting op id/name pairs)

### DIFF
--- a/cli/tests/testdata/workers/test.ts
+++ b/cli/tests/testdata/workers/test.ts
@@ -655,7 +655,7 @@ Deno.test("Worker with invalid permission arg", function () {
         deno: { permissions: { env: "foo" } },
       }),
     TypeError,
-    'Error parsing args at position 1: (deno.permissions.env) invalid value: string "foo", expected "inherit" or boolean or string[]',
+    'Error parsing args at position 0: (deno.permissions.env) invalid value: string "foo", expected "inherit" or boolean or string[]',
   );
 });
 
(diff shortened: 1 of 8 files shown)


aapoalas:
Oh yes, this looks nice.

littledivy (approved):
LGTM!

PR #14237, opened 2022-04-08
[wip] feat(serde_v8): Garbage collectable resources

littledivy:
`serde_v8::Resource<T>` is an "object wrap" around Rust allocated objects whose lifetime is managed by the V8 GC. Rust-side shared reference is managed with an `Rc<T>`.

Rust can reclaim ownership with `Resource::into_inner`. This prevents the finalizer to drops the held reference, transferring ownership back to Rust.

```rust
use serde_v8::Resource;
use deno_core::op;

pub struct ExtResource {
  msg: String,
}

#[op]
fn create_resource() -> Result<Resource<ExtResource>> {
  Ok(Resource::new(ExtResource { msg: String::from("Hi, I'm a Rust allocated value _owned_ by V8") }))
}

#[op]
fn print_resource_thing(resource: Resource<ExtResource>) -> Result<()> {
  let rc = resource.borrow();
  println!("{}", rc.msg);
  Ok(())
}
```

I've also updated the `fetch()` API to use this implementation.

TODO:
- [x] Drop all leaked `v8::Weak`s when the isolate disposes without calling the finalizer. This causes a crash!
- [x] https://github.com/denoland/rusty_v8/pull/895 (does most V8 <-> Rust Weak handles heavylifting stuff, thanks @andreubotella)


serde_v8/magic/resource.rs
```diff
@@ -0,0 +1,231 @@
...
+use std::rc::Rc;
+
+/// A Resource is a Rust object managed by the V8 GC.
+/// `T` is reference counted using an `Rc<T>`.
+/// When sent to V8, the Resource will be wrapped in a JavaScript object.
+/// The JavaScript object will hold a reference to the Rust object.
+///
+/// The underlying Rc<T> will always have a strong count >= 1 until either
+/// the JavaScript object is garbage collected OR `into_inner` is called.
+pub struct Resource<T: ?Sized> {
```
piscisaureus, line 21: Can the whole thing be simplified by using `inner: Rc<RefCell<Option<T>>>` maybe?

aapoalas:
I really want to have this feature for FFI callbacks. I'm thinking that perhaps the registered callbacks themselves can be GC'able resources, at which point parsing FFI arguments could be as simple as:
```rs
enum Arg {
  Value(serde_v8::Value),
  Buffer(ZeroCopyBuf),
  Callback(FfiCallbackResource),
}
```

Now that would be grand.

PR #14242, opened 2022-04-09
feat: update to TypeScript 4.7

kitsonk:


cli/dts/lib.dom.extras.d.ts
```diff
@@ -7,10 +7,6 @@
 
 /// <reference no-default-lib="true"/>
 
-interface AbortSignal extends EventTarget {
```
kitsonk, line 10: This is now part of lib dom.

cli/lsp/tsc.rs
```diff
@@ -1816,14 +1819,14 @@ impl CallHierarchyOutgoingCall {
...
-  // kind: ScriptElementKind,
-  // kind_modifiers: String,
   display_parts: Vec<SymbolDisplayPart>,
   documentation: Option<Vec<SymbolDisplayPart>>,
   tags: Option<Vec<JsDocTagInfo>>,
+  // name: String,
+  // kind: ScriptElementKind,
+  // kind_modifiers: String,
   // code_actions: Option<Vec<CodeAction>>,
-  // source: Option<Vec<SymbolDisplayPart>>,
```
kitsonk, line 1826: this is deprecated, so we shouldn't ever use it if we need it.

cli/dts/lib.webworker.d.ts
```diff
@@ -695,7 +723,7 @@ interface AbortController {
     /** Returns the AbortSignal object associated with this object. */
     readonly signal: AbortSignal;
     /** Invoking this method will set this object's AbortSignal's aborted flag and signal to any observers that the associated activity is to be aborted. */
-    abort(reason?: any): void;
+    // abort(): AbortSignal; - To be re-added in the future
```
bartlomieju, line 726: Is this on purpose? Looks strange

bartlomieju (approved):
LGTM, should we wait until the week of June 13th before landing this PR? It appears it might be painful to not apply it in v1.22 branch

cli/dts/lib.webworker.d.ts
lucacasonato: Agreed, this seems a bit problematic.

kitsonk: It is what is upstream: https://github.com/microsoft/TypeScript/blob/release-4.7/lib/lib.webworker.d.ts#L726

I really don't want to get into updating their distributables if we can avoid it.  It appears they have fixed it in main, but that is what they issued with 4.7.  The impact on Deno users will be minimal, because it only gets used when users explicitly use it.

kitsonk:
> LGTM, should we wait until the week of June 13th before landing this PR? It appears it might be painful to not apply it in v1.22 branch

I would rather get it in so canary has it so it isn't some big surprise when it makes the release as it may highlight more "breakages" in `std` and in other libraries.

It is mostly atomic except for the LSP changes, and even then, it is stuff we likely won't be touching anytime soon.

Also, I am away the week of June 13th.

bartlomieju:
> > LGTM, should we wait until the week of June 13th before landing this PR? It appears it might be painful to not apply it in v1.22 branch
> 
> I would rather get it in so canary has it so it isn't some big surprise when it makes the release as it may highlight more "breakages" in `std` and in other libraries.
> 
> It is mostly atomic except for the LSP changes, and even then, it is stuff we likely won't be touching anytime soon.
> 
> Also, I am away the week of June 13th.

Okay, let's land it now then. I'll cut the 1.23.2 release and deal with any troubles then.

Issue #13256, opened 2022-01-01
`deno repl --eval-file`


somebody1234:
an `--eval-file` commandline option may be useful? since `--eval <pasted code>` is little different from `deno repl` then pasting the code in.
plus with longer code and/or code with multiple lines, i think it would quickly become quite unwieldy

kitsonk:
How would `--eval-file` differ from `deno run`?

somebody1234:
because it's for `deno repl`?
i.e. it'd function exactly the same as current `deno repl --eval` - as a prelude to set up the repl environment

bartlomieju:
I think this is actually very reasonable request. @dsherret what do you think?

dsherret:
This was discussed a bit in #11317.

How this should work is a bit ambiguous—should local variables from the provided module become available in the repl or should only exports?

There's already a non-abiguous way to do this so I'm not sure how much value having an `--eval-file` flag would add:

```ts
deno repl -eval "import * as myModule from './relative-path-to-file.ts';"
deno repl -eval "import { someValue, otherValue } from 'file:///absolute/path-to-file.ts';"
deno repl -eval "import { someValue, otherValue } from 'https://deno.land/x/something.ts';"
```

somebody1234:
as you said, there's already a non-ambiguous way to do the latter, so it's the former I'm interested in. especially being able to add global variables without the nasty `Object.defineProperty` hacks the node compatibility scripts use.

somebody1234:
basically the idea is, I'm assuming people would seldom use `--eval` since you could just, well `deno repl` as usual and manyally `import` (as usual)

of course, there are added benefits like being IDE support for checking the code now that it's in its own file and not, say, in a string literal in a shell script

somebody1234:
importing members manually could work, but assuming the repl is intended for rapid prototyping generally you'll probably want to, say, import *every* exported member from several modules, which could quickly become tedious to type every time.

... well tbh it'd be tedious to type even with the above, relatively short, examples imo. i guess thats partly because I'm not a very fast typer

but anyway tl;dr i guess usefulness would depend very heavily on what the repl is actually intended for/what are the most common intended usecases

bartlomieju:
Implemented in https://github.com/denoland/deno/pull/14247

Issue #7938, opened 2020-10-11
Deno.repl() feature suggestion 


JeremyJaydan:
It'd be neat if we could trigger the deno repl from within a script and specify the initial input.

Below script here would launch a deno repl that inputs `1 + 1`, then returns the result, and waits for further inputs.
```typescript
Deno.repl(`1 + 1`);
```

A possible use case is for cli tools that have apis, the below script would launch a deno repl with your own prelude:
```typescript
Deno.repl(`const foo = await import('path/to/mod.ts')`);
```
Above could be executed by your main cli command, similar to the deno cli.

Could also be an alternative to making cli tools in general for simple stuff by combining this with `deno install`.

andreespirela:
This could be nice, I can see several use cases such as running external scripts in a dynamic way. It may even work for dynamic automation & having the Deno namespace available

kitsonk:
How would this be better than `eval()` or `new Function()`?

andreespirela:
> How would this be better than `eval()` or `new Function()`?

As far as I am aware, eval doesn’t have acess to the Deno namespace, eval is purely javascript code (is this right?) with `Deno.repl` you could have access to the Deno namespace and execute deno scripts dinamically. 

Note: I didn’t suggest this feature, I’m just commenting from what makes sense to me, since “eval” would be the same thing so I’m guessing Deno.repl is to be used for deno scripts

kitsonk:
@andreespirela `eval()` runs in the same context where it is invoked.  Just try `eval("Deno.cwd()");` (same with `new Function()`)

andreespirela:
You’re right @kitsonk , then I guess my question is the same as yours, how is it different from eval.

JeremyJaydan:
Hey @kitsonk @andreespirela 

What I'm referring to is starting a deno repl via script. When you type the `deno` command in the terminal, that is the deno repl I'm referring to.

1. Imagine having the following in a script called `lodash.ts`
```typescript
Deno.repl(`const foo = await import('https://deno.land/x/lodash@4.17.19/lodash.js')`);
```

2. Then you did `deno install ./lodash.ts`

That would install that script with the command `lodash`.

3. When you type `lodash`, an interactive deno repl starts up and imports the lodash library.

I hope that makes more sense, let me know if it doesn't!

Also I just noticed this pr here, maybe the above could be an alternative solution to #7427

JeremyJaydan:
[...]

Issue #7425, opened 2020-09-11
[Feature Request] CLI option for REPL imports


solson:
### Motivation

Being able to run `deno repl` with module imports would be very useful to me for experimenting with new libraries and for general desktop calculator use (by defining my own module of helper functions). This would be roughly equivalent to `node --require`.

It is currently difficult to import from the REPL, requiring hard-to-discover incantations like `let colors; import("https://deno.land/std@0.68.0/fmt/colors.ts").then((m) => colors = m)` which might allow you to observe `colors` being `undefined` if the import doesn't complete fast enough. Top-level await in the REPL would make this cleaner, but it would still not be easy to make a shell script or alias that invokes the REPL with pre-defined imports, since piping to `deno repl` makes it non-interactive. An explicit import flag can sidestep these problems.

### Design

We should add a new CLI option to `deno repl` tentatively called `-i, --import` which takes a list of `NAME=URL` pairs and makes the given modules available as if by `import * as NAME from "URL";`.

### Use cases

Example 1:

```console
$ deno repl -i hash=https://deno.land/std@0.68.0/hash/mod.ts
Download https://deno.land/std@0.68.0/hash/mod.ts
Download https://deno.land/std@0.68.0/hash/_wasm/hash.ts
Download https://deno.land/std@0.68.0/hash/hasher.ts
Download https://deno.land/std@0.68.0/hash/_wasm/wasm.js
Check https://deno.land/std@0.68.0/hash/mod.ts
Deno 1.3.3
exit using ctrl+d or close()
> { const h = hash.createHash("sha256"); h.update("hello world"); h.digest() }
Uint8Array(32) [
  185,  77,  39, 185, 147,  77,  62,
    8, 165,  46,  82, 215, 218, 125,
  171, 250, 196, 132, 239, 227, 122,
   83, 128, 238, 144, 136, 247, 172,
  226, 239, 205, 233
]
> { const h = hash.createHash("sha256"); h.update("hello world"); h.toString() }
b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9
```

Example 2:

```console
$ cat ~/scripts/calc.ts
import { createHash } from "https://deno.land/std@0.68.0/hash/mod.ts";

export function sha256(s: string): string {
  const h = createHash("sha256");
  h.update(s);
  return h.toString();
}

$ alias calc='deno repl -i calc=$HOME/scripts/calc.ts'

$ calc
Deno 1.3.3
[...]

PR #14247, opened 2022-04-10
feat(repl): add "--eval-file" flag to execute a script file on startup

najumancheril:
<h2>Syntax</h2>
deno repl --script /path/to/script

<h2>Purpose</h2>
The purpose of this change is to provide deno with an option that is analogous to python -i /path/to/source, which runs the specified source file as additional prelude and then shows the user a REPL that can be used to enter more code. Python's -i flag is very useful for prototyping and experimentation. It is also useful for students who want to set up example structures and then examine them in an interactive session to improve their understanding of the language. This feature is different from "deno run" because "deno run" exits once the specified module exits. deno run is not interactive.

<h2>Example</h2>

```
$ cat a.js
let a=5;

$ deno repl --script a.js
Deno 1.20.5
exit using ctrl+d or close()
> a
5
>

```

<h2>Related Issues</h2>
This feature (and closely related features) have been previously discussed:

1. #3173
2. #7425
3. #7938
4. #11317
5. #13256

<h2>Related Work</h2>
Related work includes the implementation of the --eval flag (https://github.com/denoland/deno/pull/11590) which allows the user to specify the prelude code as command-line parameters. This is useful, but may be inconvenient for larger preludes.

<h2>Potential Concerns</h2>

1. --script may be a bad flag name.  #13256 suggests the name --eval-file. --prelude is another possibility, but it may be confusing because cli/tools/repl/session.rs already specifies some initial commands in a PRELUDE variable.
2. There is discussion by @dsherret on #13256 that this feature may not be worth it. It is already possible to expose module exports efficiently via --eval. Do we want a way to expose local variables in an interactive session? I think the answer is "probably." For example, students may find it useful to use preludes which define variables in a global scope because they don't understand modules yet. Other users may find it useful during prototyping when the amount of code is small and not worth structuring as a module.
3. There is discussion by @dsherret on #11317 about possibly modifying the script code to rewrite module import paths, etc. This implementation does not attempt to do that.
4. If the user specifies both --script and --eval, which should be executed first? In this implementation, I execute --script first.
5. Perhaps we should support multiple prelude files? The current implementation supports only one.
6. Perhaps we should support http:// in addition to file system access?
7. This is my first Rust program, so there may be bugs or non-idiomatic patterns in the code.


cli/tools/repl/mod.rs
```diff
@@ -58,9 +60,17 @@ async fn read_line_and_poll(
   }
 }
 
+fn read_script(script_path: String) -> Result<String, AnyError> {
+  let mut file = File::open(script_path)?;
+  let mut data = Vec::new();
+  file.read_to_end(&mut data)?;
+  Ok(String::from_utf8(data)?)
+}
```
bartlomieju, line 68: We should reuse facilities from `FileFetcher` to provide ability to load remote scripts as well

bartlomieju (commented):
This PR was discussed during last CLI design meeting and there's support for landing it. A few changes that were requested:

- the flag should be named `--eval-file`
- the flag should support multiple values
- both local and remote scripts should be supported
- loaded files should be type-stripped if they have `.ts` extension and left as they are for `.js`

@najumancheril this is a good start, we will need to add an integration test for this feature in `cli/tests/integration/repl_tests.rs`. Let me know if you need some pointers.

cli/tools/repl/mod.rs
najumancheril: This is great feedback, thank you. I will make these changes.

najumancheril:
I have implemented the first three requested changes.
1) Flag is now named --eval-file.
2) Multiple value are now supported. I was not sure what syntax to use, so I copied the argument parsing code for allow-lists (e.g. deno repl --eval-file=./a.js,http://localhost:8000/b.js,http://localhost:8000/c.js). Note that equal sign is now required.
3) tools::repl::run takes a ProcState, which has a FileFetcher as member. I am using this to load scripts. It handles both local and remote scripts. It also respects --reload. Due to our use of deno_core::resolve_import, local script paths now need to start with ./, ../, or /.

I will look into type-stripping and tests tomorrow. Let me know if you have any feedback on the above.

bartlomieju (commented):
@najumancheril this looks great! I think @dsherret should be able to provide some pointers regarding transpilation.

najumancheril:
@bartlomieju and @dsherret, I don't think I need to make any additional changes to deno in order to transpile .ts script files passed via --eval-file into JavaScript. The REPL entry point that I am using to inject the script code (ReplSession.evaluate_line_and_get_output) already does TypeScript transpilation. Here is the sequence of events:

1. tools::repl::run loads each eval file's source code (JavaScript or TypeScript code) as a String and invokes ReplSession.evaluate_line_and_get_output on that code String
2. evaluate_line_and_get_output invokes ReplSession.evaluate_line_with_object_wrapping on the code
3. evaluate_line_with_object_wrapping invokes ReplSession.evaluate_ts_expression on the code
4. evaluate_ts_expression creates a ParsedSource object with the code and invokes .transpile(). **This does the TypeScript to JavaScript conversion.**
6. evaluate_ts_expression invokes ReplSession.evaluate_expression on the transpiled code,.

Does that explanation make sense? If so, then I will proceed with adding the integration tests.

bartlomieju:
> @bartlomieju and @dsherret, I don't think I need to make any additional changes to deno in order to transpile .ts script files passed via --eval-file into JavaScript. The REPL entry point that I am using to inject the script code (ReplSession.evaluate_line_and_get_output) already does TypeScript transpilation. Here is the sequence of events:
> 
> 1. tools::repl::run loads each eval file's source code (JavaScript or TypeScript code) as a String and invokes ReplSession.evaluate_line_and_get_output on that code String
> 2. evaluate_line_and_get_output invokes ReplSession.evaluate_line_with_object_wrapping on the code
> 3. evaluate_line_with_object_wrapping invokes ReplSession.evaluate_ts_expression on the code
> 4. evaluate_ts_expression creates a ParsedSource object with the code and invokes .transpile(). **This does the TypeScript to JavaScript conversion.**
> 5. evaluate_ts_expression invokes ReplSession.evaluate_expression on the transpiled code,.
> 
> Does that explanation make sense? If so, then I will proceed with adding the integration tests.

Thanks, that makes sense. I wonder about edge cases, but let's leave that until tests are done (maybe tests will surface some of these cases). I'd like to land this PR as soon as possible, so we could ship it in Deno v1.21 on Wednesday.

najumancheril:
@bartlomieju I have added several tests for the new --eval-file flag. Please let know if you'd like any additional tests and/or changes.

cli/flags.rs
```diff
@@ -1299,6 +1309,16 @@ Ignore linting a file by adding an ignore comment at the top of the file:
...
   runtime_args(Command::new("repl"), false, true)
     .about("Read Eval Print Loop")
+    .arg(
+      Arg::new("eval-file")
+        .long("eval-file")
+        .min_values(1)
+        .takes_value(true)
+        .use_value_delimiter(true)
+        .require_equals(true)
+        .help("Evaluates the provided code file(s) when the REPL starts.")
```
bartlomieju, line 1319:
```suggestion
        .help("Evaluates the provided file(s) as scripts when the REPL starts. Accepts file paths and URLs.")
```

bartlomieju (commented):
Looks well implemented. I will try it locally and give it another review. @dsherret please take a look as well

najumancheril:
I agree this is annoying. I just noticed that resolve_import returns a specific error code (ParseError::RelativeUrlWithoutBase) in this case, so I can check for that and retry with an adjusted eval file path. I will do this now.

bartlomieju:
> I agree this is annoying. I just noticed that resolve_import returns a specific error code (ParseError::RelativeUrlWithoutBase) in this case, so I can check for that and retry with an adjusted eval file path. I will do this now.

I think you can just apply `resolve_url_or_path` function to each argument in that list.

najumancheril:
OK I think we're good

bartlomieju:
@najumancheril please merge `main` branch into your branch and fix lint errors to make CI happy.

dsherret (approved):
This seems reasonable to me and the implementation looks good! I think if someone provides a module to this, they'll just have to make sure they use absolute specifiers if they import anything.

bartlomieju:
> This seems reasonable to me and the implementation looks good! I think if someone provides a module to this, they'll just have to make sure they use absolute specifiers if they import anything.

That's a good point. I think we should add more explanation for the `--eval-flag` help. Also we will need to update https://github.com/denoland/manual/blob/main/tools/repl.md @najumancheril could you look into that?

najumancheril:
I have updated the deno manual with information about --eval-file: https://github.com/denoland/manual/pull/293

bartlomieju (approved):
LGTM

Thank you @najumancheril. Great work for a first contribution, I can see cool use cases for this feature.

Issue #14131, opened 2022-03-27
Consider keeping the scripts in `/tools` from importing network modules


andreubotella:
The `/tools` directory in the Deno repo is home to a number of scripts used for Deno development, cutting releases and other things. For security, it would seem preferable to have only things vendored within the repo as dependencies, to avoid loading modules from the network. This is already done for the standard library, which is vendored in `/test_util/std`, but here are other instances:

- `/tools/upload_wptfyi.js` imports [`compress`](https://deno.land/x/compress@v0.4.1) version 0.4.1 from deno.land/x
- `/tools/wpt/runner.ts` imports [`deno_dom`](https://deno.land/x/deno_dom@v0.1.3-alpha2) version 0.1.3 alpha 2 from deno.land/x
- `/tools/release/deps.ts` imports the [`denoland/automation`](https://github.com/denoland/automation) github, from tag 0.5.0.
- `/tools/bench/rebootstrap.js` imports `path/mod.ts` and `fs/mod.ts` from the latest version of the standard library at deno.land/std.

PR #14251, opened 2022-04-10
chore: use local copy of std in tools/bench/ scripts

bartlomieju:
Ref https://github.com/denoland/deno/issues/14131


kt3k (approved):
LGTM

PR #14253, opened 2022-04-10
chore: use lock file for tooling scripts

bartlomieju:
Closes https://github.com/denoland/deno/issues/14131

I'm not really happy with this solution yet, I believe there's more deduplication to be done.


kt3k:
Does `vendor` include the all dependencies of all tooling scripts? What is the command to update the vendor directory when we change any dependency in any script?

bartlomieju:
> Maybe we could just use a lockfile instead?

I believe that should be enough, I'll rework the PR to do that.

dsherret (approved):
LGTM

Issue #14188, opened 2022-04-02
test: add file paths to the failures section


jespertheend:
from https://github.com/denoland/deno/discussions/14156:
> Print the path above the final list of failures. This could look similar to how the first list of test results look, where it prints the path only once if multiple tests in a single file fail.

![image](https://user-images.githubusercontent.com/2737650/161397183-b9e719c0-6063-44b5-8149-40f0d3c402b5.png)

Issue #14187, opened 2022-04-02
test: add file paths to the errors section


jespertheend:
from https://github.com/denoland/deno/discussions/14156:
> Print the path of a failing test above the error. Currently the name of the test is already printed, but often this is not enough. Say you have a file `calculator.test.js` with test name `"it adds two numbers together"` and only the name is printed, it is not immediately clear that this test is from `calculator.test.js` rather than another file that might have a similar test.

![image](https://user-images.githubusercontent.com/2737650/161397135-a6379f34-239e-48ca-8fd9-21c930b0dfd6.png)

PR #14255, opened 2022-04-11
feat(test): Improve testing report output

bartlomieju:
Following changes were done in this PR:
- remove "test" prefix before each test
- use gray color for "running N tests from ..." prompt
- use relative path or remote URL instead of full URL in "running N tests from ..." prompt
- in "failures" section, add file path/remote URL before the test name

Before:
Screenshot 2022-04-11 at 16 21 25

After:
Screenshot 2022-04-11 at 16 43 58

Closes https://github.com/denoland/deno/issues/14186
Closes https://github.com/denoland/deno/issues/14188
Closes https://github.com/denoland/deno/issues/14187

More PRs to follow, that will improve final report and errors display.


cli/tools/test.rs
```diff
@@ -388,14 +404,36 @@ impl TestReporter for PrettyTestReporter {
...
+          colors::gray(">"),
+          description.name
+        );
         println!("{}", error);
         println!();
       }
 
-      println!("failures:\n");
+      let mut grouped_by_origin: HashMap<String, Vec<String>> =
+        HashMap::default();
```
dsherret, line 420: This output will have unstable sorting. Maybe use a `BTreeMap`?

dsherret (approved):
Looks much better!

cli/tools/test.rs
bartlomieju: Yeah, I'll just do `BTreeMap`

Issue #14190, opened 2022-04-02
test: Better headers for the three `deno test` sections


jespertheend:
Right now `deno test` gets printed with three sections, but it's not immediately clear where one ends and the other starts. And the last two sections both have `failures:` as header.

I think it could be nice to make them a bit more distinguishable and name them `results`, `errors` and `failures` respectively.

Giving them a separate color would also help I think.

![image](https://user-images.githubusercontent.com/2737650/160925198-9b11111b-465b-4574-b4b4-ecf17d31bbd2.png)

bartlomieju:
Ooops, this is actually not fixed...

Issue #13913, opened 2022-03-11
Improve footer in `deno test`


lucsoft:
The current footer is a bit noisy and not easy to identify

A improved footer could look something like this:
<img width="253" alt="image" src="https://user-images.githubusercontent.com/11093382/157891863-08ac8440-af4a-4178-b7bc-8e2392c407d4.png">

Compared to the current one:
<img width="350" alt="image" src="https://user-images.githubusercontent.com/11093382/157891882-db296d5e-05a3-475d-983a-48df48edb0d8.png">

Backstory to this issue:
[vitest on twitter](https://twitter.com/alexlafroscia/status/1501204689278443530)
[discord chat](https://canary.discord.com/channels/684898665143206084/684911491035430919/951840231383007302)

nayeemrmn:
After playing around with the output more, I liked it better as less vertical space was taken up by the summary on an integrated terminal when scrolling up to see failures. It would be nice if we could reduce noise while keeping it to one line, maybe by omitting `ignored` and `filtered out` if they are 0 (what even is `measured`), considering a different delimiter than semicolon, etc.

![image](https://user-images.githubusercontent.com/29990554/166854505-6b94e11d-f230-47bc-b966-43c0d3b6fb9e.png)
->
![image](https://user-images.githubusercontent.com/29990554/166854550-549338d5-8930-466c-8f80-ead3b13a8461.png)

![image](https://user-images.githubusercontent.com/29990554/166854593-26d64ff2-eb99-4f2b-928a-003448b813bb.png)
->
![image](https://user-images.githubusercontent.com/29990554/166854631-505c3832-974d-40b7-86ce-61d17b185077.png)

bartlomieju:
@nayeemrmn that looks good, I think this is better than vertical summary that I proposed in #14262

mrkldshv:
Hey! Is this issue free to pick up? Or there were some work done by @nayeemrmn?

nayeemrmn:
@mrkldshv It's free

mrkldshv:
Right, I'd be happy to work on this issue.

bartlomieju:
> Right, I'd be happy to work on this issue.

Please do!

mrkldshv:
Do we want to keep horizontal summary, but make it easier to read?

PR #14262, opened 2022-04-12
[WIP] feat(test): refreshed test report

bartlomieju:
Screenshot 2022-04-16 at 22 03 38

Closes https://github.com/denoland/deno/issues/14190
Closes https://github.com/denoland/deno/issues/13913


cli/fmt_errors.rs
```diff
@@ -220,25 +228,59 @@ fn format_js_error(js_error: &JsError, is_child: bool) -> String {
...
+          false
+        }
+      })
+      .collect();
+    js_error.frames = frames;
+
+    let pretty_js_error = Self {
+      js_error,
+      color_message: true,
+    };
```
bartlomieju, line 265: @nayeemrmn does this seem fine?

nayeemrmn: IMO this change shouldn't involve any change to `fmt_errors.rs`, we should have a function in `cli/tools/test.rs`:
```rust
fn format_test_error(js_error: &JsError) -> String {
  let mut js_error = js_error.clone();
  let e = colors::red(&js_error.exception_message).to_string();
  js_error.exception_message = e;
  let frames = std::mem::take(&mut js_error.frames);
  let frames = frames
    .into_iter()
    .rev()
    .skip_while(|f| {
      if let Some(file_name) = &f.file_name {
        file_name.starts_with("[deno:") || file_name.starts_with("deno:")
      } else {
        false
      }
    })
    .collect();
  js_error.frames = frames;
  PrettyJsError::create(js_error)
    .to_string()
    .trim_start_matches("Uncaught ")
    .to_string()
}
```
```rust
for (description, js_error) in &summary.failures {
  // ...
  println!("{}", format_test_error(js_error));
  println!();
}
```

bartlomieju: Makes sense, I'll do that in separate PR, so this change can be landed independently.

lucacasonato:
Some feedback:

The `$name\n== output ==\n$output\n== output end ==\n$status ($time ms)\n` formatting seems weird to me. I think we should do something more like Rust does, where we capture and buffer by default, unless the test fails (in which case we dump the output with the error messages). A user could disable this behavior with a `--no-capture` flag or something (just like libtest in Rust).

The `running 3 tests from b_test.ts` messages are still very useless when running with `-j` flag. In parallel mode each test needs to be prefixed by test name, or test completions per file need to be buffered.

I don't like the many colors in the final output. I think there should only be colors for the FAILED / PASS label, not for the number of failures and passes.

---

Overall I think this takes some good inspiration from Vitest, but I think we should be more liberal with adopting some of their nice test output. I think for example that the extra left indentation and line numbers on the per test failure report is very nice:

image

---

I think we should also really consider more "radical" changes to the test runner output. For example I think including test names for each file by default is really not that important. If we instead had an "interactive mode" version of this output that showed the test file name, and then a live updating count of tests already run / left to run that'd be great IMO. That would also work out much better with the `-j` flag.

nayeemrmn:
> image

The red background part in this image is a lot more readable than what we have now, especially on light theme it's not good without bold:
image 
image
or maybe 
image

cli/fmt_errors.rs
bartlomieju: https://github.com/denoland/deno/pull/14302

bartlomieju:
Updated output

nayeemrmn:
I think the right-aligned text is out of place, this seem readable enough: 
image

bartlomieju:
> I think the right-aligned text is out of place, this seem readable enough: image

Applied this suggestion

cli/tools/test.rs
```diff
@@ -508,21 +502,101 @@ impl TestReporter for PrettyTestReporter {
...
+      println!(
+        "{}     {}{}",
+        colors::gray("ignored:"),
+        summary.ignored,
+        get_steps_text(summary.ignored_steps)
+      );
+    }
+    if summary.filtered_out > 0 {
+      println!("{}    {}", colors::gray("filtered:"), summary.filtered_out);
+    }
```
bartlomieju, line 532: Keep in mind, that "failed", "ignored" and "filtered out" counts are conditional and are displayed only if the count of that category is greater than zero. Not sure if this is desirable or not. Options welcome

aapoalas: Probably a good thing. There's very little reason to keep the cluttering messages around when they convey exactly no information.

PR #14265, opened 2022-04-12
experiment(runtime): bypass tokio File

AaronO:
Allowing us to drive FS IO buffer size instead of being bound by tokio's 16kb limit

### DIFF
--- a/runtime/ops/fs.rs
+++ b/runtime/ops/fs.rs
@@ -22,13 +22,14 @@ use serde::Serialize;
 use std::cell::RefCell;
 use std::convert::From;
 use std::env::{current_dir, set_current_dir, temp_dir};
+use std::fs::File as StdFile;
 use std::io;
 use std::io::{Error, Seek, SeekFrom};
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
-use tokio::io::AsyncSeekExt;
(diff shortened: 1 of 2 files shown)


runtime/ops/io.rs
```diff
@@ -227,8 +227,7 @@ impl Resource for ChildStderrResource {
 
 #[derive(Debug, Default)]
 pub struct StdFileResource {
-  pub fs_file:
-    Option<AsyncRefCell<(Option<tokio::fs::File>, Option<FileMetadata>)>>,
+  pub fs_file: Option<AsyncRefCell<(Option<StdFile>, Option<FileMetadata>)>>,
```
dsherret, line 230: Is it possible for us to remove this `AsyncRefCell` now with this change?

dsherret, line 230: Is it possible for us to remove this `AsyncRefCell` now with this change?

runtime/ops/fs.rs
```diff
@@ -246,8 +246,11 @@ async fn op_seek_async(
     .borrow_mut()
     .await;
 
-  let pos = (*fs_file).0.as_mut().unwrap().seek(seek_from).await?;
-  Ok(pos)
+  let std_file = fs_file.0.as_mut().unwrap();
+  let std_file: &'static mut StdFile = unsafe { std::mem::transmute(std_file) };
```
ry, line 250: This is unnecessarily unsafe - use a RefCell or some other type safe way of achieving this.

PR #14266, opened 2022-04-12
perf(core): Deno.core.unsafeBuffer

AaronO:
To create a non bzero'd ArrayBuffer of a given size. Useful when creating a buffer for IO that will be immediately written to

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -1460,6 +1460,8 @@ fn unsafe_buffer(
     v8::Local::<v8::Integer>::try_from(args.get(0)).map_or(0, |x| x.value());
   let len = len as usize;
   let mut bytes = Vec::with_capacity(len);
+  // SAFETY: caveat emptor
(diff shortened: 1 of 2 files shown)


lucacasonato:
This can be used to leak heap contents into the isolate. As such I am not in favour. If we want to do I/O reads using non zeroed buffers then create an op that takes a length to read and does the allocation and reading inside of that op. You can then return the clean buffer to the runtime.

ry:
I agree with Luca. We can just make `Deno.core.opSync("op_fake_read")` return the allocated buf - then it's (probably) even more efficient one opcall instead of two

lucacasonato (commented):
Also, just generally: calling some unsafe code in an op or binding with a safety comment "caveat emptor" (caller beware), but then just directly exposing that to userland is just really very very questionable.

AaronO:
Will replace with an `op_readn`

PR #14269, opened 2022-04-13
refactor(core/error): Clarify JsError message fields

### LINKED ISSUE #13212: Remove `"uncaught"` when creating JsError in core
in https://github.com/denoland/deno/blob/07618c861e27e284cae3a309783ef78dcef611dd/core/error.rs#L192-L200 we add `"uncaught"` to the error message. this feels quite out of place for core, but rather something for CLI to do. also a blocker for #13209

nayeemrmn:
Closes #13212. Also adds `message` and `name` fields to make `Deno.core.destructureError()` nicer to use.

### DIFF
--- a/cli/fmt_errors.rs
+++ b/cli/fmt_errors.rs
@@ -283,7 +283,7 @@ impl fmt::Display for PrettyJsError {
       "{}",
       &format_stack(
         true,
-        &self.0.message,
+        &self.0.exception_message,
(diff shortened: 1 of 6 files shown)


bartlomieju:
Very nice, I hit this problem again in #14262, this should make it more manageable

bartlomieju (approved):
LGTM, thank you @nayeemrmn

Issue #14192, opened 2022-04-02
test: `console.log` calls get printed in between the test results


jespertheend:
Consider the following test:
```js
Deno.test({
  name: "logs a message",
  fn() {
    console.log({bool: true, num: 5})
    console.log("hello");
  }
})
```

When you run it, it shows up with the console messages in between `test logs a message ...` and `ok (8ms)`:

![image](https://user-images.githubusercontent.com/2737650/161397622-ac214aa9-5bfe-42a9-acce-1b94a2f881b7.png)

And also, multiple `console.log` calls have two new lines in between them. If you were to do this with `deno run`, they are printed with only a single new line in between:

![image](https://user-images.githubusercontent.com/2737650/161397703-e5a4ed0f-0066-420c-8fbf-377b043a941b.png)

Not sure what the right behaviour is here, but some things I can think of:
- keep it almost like it is right now, except add a new line before the first `console.log()` call, and then only a single new line between every two calls, rather than the double new lines it currently prints.
- print the `console.log()` messages above/below the test result, and then print `ok (8ms)` retroactively in the right position once the test finishes. Not sure how difficult this would be, or if it even would be possible at all.
- Disable printing `console.log()` messages during tests altogether.

All in all I don't think this issue needs to have a very high priority imo, tests generally shouldn't run code that call `console.log` without mocking it first. But I do have a lot of tests that dynamically import blob urls, which looks kind of weird:

![image](https://user-images.githubusercontent.com/2737650/160927670-d88a76d2-c298-4094-a301-80a5ac80a09c.png)

luk3skyw4lker:
I'm gonna take a look at it and open a PR as soon as I can.

luk3skyw4lker:
I guess I'll go with the first idea of keeping it as it is, but adding a new line before the first console.log call and only a single line break for subsequent calls.

I don't think that disabling the console.log messages during tests is the best way, the messages could be useful to someone. Maybe the case could be adding a flag like "--no-console" to the test options, but I'm not sure of how hard this is.

PR #14271, opened 2022-04-13
feat(test): format user code output

bartlomieju:
Closes https://github.com/denoland/deno/issues/14192

This commit changes "deno test" to better denote user output coming
from test cases. 

This is done by printing "---- output ----" and "---- output end ----" 
markers if an output is produced. The output from "console" and 
"Deno.core.print" is captured, as well as direct writes to "Deno.stdout"
and "Deno.stderr".

To achieve that new APIs were added to "deno_core" crate, that allow
to replace an existing resource with a different one (while keeping resource
ids intact). Resources for stdout and stderr are replaced by pipes.

Screenshot 2022-04-14 at 19 20 44

### DIFF
--- a/cli/lsp/testing/execution.rs
+++ b/cli/lsp/testing/execution.rs
@@ -14,6 +14,7 @@ use crate::lsp::client::TestingNotification;
 use crate::lsp::config;
 use crate::lsp::logging::lsp_log;
 use crate::ops;
+use crate::ops::testing::create_stdout_stderr_pipes;
 use crate::proc_state;
 use crate::tools::test;
 
@@ -183,11 +184,17 @@ async fn test_specifier(
   options: Option<Value>,
 ) -> Result<(), AnyError> {
   if !token.is_cancelled() {
+    let (stdout_writer, stderr_writer) =
+      create_stdout_stderr_pipes(channel.clone());
     let mut worker = create_main_worker(
       &ps,
       specifier.clone(),
       permissions,
-      vec![ops::testing::init(channel.clone())],
+      vec![ops::testing::init(
+        channel.clone(),
+        stdout_writer,
+        stderr_writer,
+      )],
     );
 
     worker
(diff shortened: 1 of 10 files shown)


bartlomieju:
Before landing, the same should be done for benching code.

cli/tools/test.rs
```diff
@@ -314,9 +319,24 @@ impl TestReporter for PrettyTestReporter {
...
+    }
+
+    if !self.did_have_user_output {
+      self.did_have_user_output = true;
+      println!();
+      println!("{}", colors::gray("-------output:-------"));
+    }
+    match output {
+      TestOutput::PrintStdout(line) | TestOutput::PrintStderr(line) => {
+        print!("{}", line)
```
bartlomieju, line 333: This should probably be `print!` for `PrintStdout` and `eprint!` for `PrintStderr`

cli/tools/test.rs
```diff
@@ -81,7 +81,10 @@ pub struct TestDescription {
 #[serde(rename_all = "camelCase")]
 pub enum TestOutput {
   // TODO(caspervonb): add stdout and stderr redirection.
```
bartlomieju, line 83: Remove

cli/lsp/testing/execution.rs
```diff
@@ -753,14 +753,18 @@ impl test::TestReporter for LspTestReporter {
...
+      | test::TestOutput::PrintStderr(value) => {
         self.progress(lsp_custom::TestRunProgressMessage::Output {
           value: value.replace('\n', "\r\n"),
           test,
           // TODO(@kitsonk) test output should include a location
           location: None,
         })
       }
+      test::TestOutput::Stdout(_) | test::TestOutput::Stderr(_) => {
+        // todo(@dsherret): handle
```
bartlomieju, line 766: Fix before landing

bartlomieju:
> Before landing, the same should be done for benching code.

I will actually address this in a follow up PR.

dsherret (approved):
LGTM

PR #14274, opened 2022-04-14
refactor: Move source map lookups to core

nayeemrmn:
This had been blocking lots of other cleanups as well. Namely the following transformations gradually faced by `JsError` have all been moved up front to `JsError::from_v8_exception()`: finding the first non-`deno:` source line [[old code]](https://github.com/denoland/deno/blob/f7e7f548499eff8d2df0872d1340ddcdfa028c45/cli/source_maps.rs#L60-L80); moving `JsError::script_resource_name` etc. into the first error stack in case of syntax errors [[old code]](https://github.com/denoland/deno/blob/f7e7f548499eff8d2df0872d1340ddcdfa028c45/cli/fmt_errors.rs#L259-L273); source mapping `JsError::script_resource_name` etc. when wrapping the error even though the frame locations are source mapped earlier [[old code]](https://github.com/denoland/deno/blob/f7e7f548499eff8d2df0872d1340ddcdfa028c45/cli/main.rs#L132-L133); removing `JsError::{script_resource_name,line_number,start_column,end_column}` entirely in favour of `js_error.frames.get(0)`. We also no longer pass a js-side callback to `core/02_error.js` from cli. I avoided doing this on previous occasions because the source map lookups were in an awkward place.


cli/tests/testdata/inline_js_source_map_2_with_inline_contents.js.out
```diff
@@ -1,4 +1,2 @@
 error: Uncaught Error: Hello world!
-throw new Error("Hello world!" as unknown as string);
-      ^
```
nayeemrmn, line 3: Source lines for inline user source maps are eliminated for now as a side effect of this cleanup, but currently only work in very specific arbitrary cases and aren't important. I think this 'regression' is fine considering.

bartlomieju: What do you mean by inline source maps?

core/bindings.rs
```diff
@@ -1324,6 +1330,42 @@ fn terminate(
...
+      location.column_number.into(),
+      &mut HashMap::new(),
+      source_map_getter.as_ref(),
+    );
+    location.file_name = f;
+    location.line_number = l as u32;
+    location.column_number = c as u32;
+    rv.set(serde_v8::to_v8(scope, location).unwrap());
+  } else {
+    dbg!();
```
bartlomieju, line 1364: Remove

core/error.rs
```diff
@@ -236,25 +236,82 @@ impl JsError {
...
+      }
+
+      let mut source_line = None;
+      let mut source_line_frame_index = None;
+      if let Some(source_map_getter) = &state.source_map_getter {
+        for (i, frame) in frames.iter().enumerate() {
+          if let (Some(file_name), Some(line_number)) =
+            (&frame.file_name, frame.line_number)
+          {
+            if !file_name.starts_with("deno:") {
```
bartlomieju, line 288: I think we should also check for file names starting with `[deno:` (the ones create by `located_script_name!` macro)

core/error.rs
```diff
@@ -236,25 +236,82 @@ impl JsError {
...
+              source_line = source_map_getter
+                .get_source_line(file_name, (line_number - 1) as usize);
+              source_line_frame_index = Some(i);
+              break;
+            }
+          }
+        }
+      } else if let Some(frame) = frames.first() {
+        if let Some(file_name) = &frame.file_name {
+          if !file_name.starts_with("deno:") {
```
bartlomieju, line 299: Ditto

cli/tests/testdata/inline_js_source_map_2_with_inline_contents.js.out
nayeemrmn: `//# sourceMappingURL=` at the bottom. Currently if that source map additionally contains the inline source in the `"sources"` field, and the top stack frame is the one selected for the source line, you would get a source line. You don't anymore with this change which isn't a big deal.

bartlomieju: Yeah I guess that's fine.

bartlomieju (approved):
LGTM, I like how much code is getting removed. That said, I want to get some more eyes on it, as moving source mapping to `core` might be a bit controversial for other team members.

PR #14276, opened 2022-04-14
TRY perf(http): zero copy response body bytes

littledivy:
`Response` clones the buffer JS side. `Bytes::copy_from_slice` did a second copy of the same data.

```
deno run -A https://deno.land/std/http/file_server.ts
```

`main`:
```
$ wrk -t8 -c20 -d10s http://localhost:4500/README.md

Running 10s test @ http://localhost:4500/README.md
  8 threads and 20 connections
    Latency   149.85us   67.68us   6.95ms   96.22%
    Req/Sec    13.17k   555.45    14.67k    97.40%
  1057363 requests in 10.10s, 110.92MB read
Requests/sec: 104697.21
Transfer/sec:     10.98MB
```

this PR:

```
Running 10s test @ http://localhost:4500/README.md
  8 threads and 20 connections
    Latency   146.98us   62.17us   6.45ms   95.38%
    Req/Sec    13.41k   539.74    13.77k    99.01%
  1078083 requests in 10.10s, 113.10MB read
Requests/sec: 106747.55
Transfer/sec:     11.20MB
```

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -34,6 +34,7 @@ use deno_core::Resource;
 use deno_core::ResourceId;
 use deno_core::StringOrBuffer;
 use deno_core::ZeroCopyBuf;
+use deno_core::serde_v8::DetachedBuffer;
 use deno_websocket::ws_create_server_stream;
 use flate2::write::GzEncoder;
 use flate2::Compression;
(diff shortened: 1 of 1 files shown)


ext/http/lib.rs
```diff
@@ -668,7 +669,7 @@ async fn op_http_write_headers(
 async fn op_http_write(
   state: Rc<RefCell<OpState>>,
   rid: ResourceId,
-  buf: ZeroCopyBuf,
+  buf: DetachedBuffer,
```
lucacasonato, line 672: Do we perform a JS side copy too?

PR #14278, opened 2022-04-14
[wip] experiment(http): introduce `requestEvent.sendFile` fast path

littledivy:
A new method to handle sending files to the socket. In the future, this could _also_ use the `sendfile(2)` syscall to avoid file copy in the userspace.

`main`:
```
Running 10s test @ http://localhost:8080/README.md
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   759.55us  121.38us   4.67ms   83.20%
    Req/Sec     2.64k    86.82     2.76k    92.20%
  212156 requests in 10.10s, 558.63MB read
Requests/sec:  21005.52
Transfer/sec:     55.31MB
```

This PR (`sendFile`):
```
Running 10s test @ http://localhost:8080/README.md
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   470.52us  126.24us   3.43ms   77.33%
    Req/Sec     4.27k   199.36     4.68k    80.57%
  343083 requests in 10.10s, 0.88GB read
Requests/sec:  33968.47
Transfer/sec:     89.44MB
```

Node:
```
Running 10s test @ http://localhost:8081/README.md
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   610.48us  782.74us  29.18ms   99.41%
    Req/Sec     3.47k   326.92     3.64k    92.08%
  279390 requests in 10.10s, 748.21MB read
Requests/sec:  27662.49
Transfer/sec:     74.08MB
```

PR #14284, opened 2022-04-14
perf(http): optimize `ReadableStream`s backed by a resource

littledivy:
83% faster than Deno 1.20.6
22% faster than Node 16.13.1

`main`:
```
  8 threads and 20 connections
    Latency    86.79ms    4.41ms 129.53ms   95.15%
    Req/Sec    23.10      7.35    40.00     83.75%
  1856 requests in 10.10s, 9.09GB read
Requests/sec:    183.78
Transfer/sec:      0.90GB
```

This PR:
```
  8 threads and 20 connections
    Latency    14.08ms    2.60ms  52.65ms   91.75%
    Req/Sec   142.92     15.35   161.00     56.44%
  11509 requests in 10.10s, 56.24GB read
Requests/sec:   1139.58
Transfer/sec:      5.57GB
```

Node:
```
Running 10s test @ http://localhost:8081/sample.txt
  8 threads and 20 connections
    Latency    18.50ms    3.05ms  74.03ms   96.78%
    Req/Sec   108.82     13.91   121.00     97.03%
  8754 requests in 10.10s, 42.82GB read
Requests/sec:    866.73
Transfer/sec:      4.24GB
```

sample.txt is 5mb

<details>
<summary> Benchmark code </summary>

```js
// Deno
const server = Deno.listen({ port: 8080 });
console.log("File server running on http://localhost:8080/");

for await (const conn of server) {
  handleHttp(conn);
}

async function handleHttp(conn) {
    // Use the request pathname as filepath
    const url = new URL(requestEvent.request.url);
    const filepath = decodeURIComponent(url.pathname);

    const file = await Deno.open("." + filepath, { read: true });
    const response = new Response(file.readable);
    await requestEvent.respondWith(response);
  }
}
```

```js
// Node
const http = require("http");
const fs = require("fs");

http.createServer(function (req, res) {
    const filename = __dirname + req.url;

    const readStream = fs.createReadStream(filename);

    readStream.on("open", function() {
        readStream.pipe(res);
    });

    readStream.on("error", function(err) {
        res.end(err);
    })
}).listen(8081);
```

</details>

It's not as fast as a dedicated `sendFile` (https://github.com/denoland/deno/pull/14278) but definitely better than `main`


server.js
```diff
@@ -0,0 +1,23 @@
...
+  for await (const requestEvent of httpConn) {
+    // Use the request pathname as filepath
+    const url = new URL(requestEvent.request.url);
+    const filepath = decodeURIComponent(url.pathname);
+
+    const file = await Deno.open("." + filepath, { read: true });
+    // const readableStream = readableStreamFromReader(file);
+    const response = new Response(file.readable);
+    await requestEvent.respondWith(response);
+    Deno.core.tryClose(file.rid);
```
littledivy, line 21: The op should try_close.

serde_v8/magic/buffer.rs
```diff
@@ -14,19 +14,26 @@ use crate::magic::transl8::impl_magic;
 pub enum MagicBuffer {
   FromV8(ZeroCopyBuf),
   ToV8(Mutex<Option<Box<[u8]>>>),
+  Temp(Vec<u8>),
```
bartlomieju, line 17: Would be nice to get a comment explaining what this "Temp" means and when it's used

ext/web/06_streams.js
```diff
@@ -633,6 +634,41 @@
     return stream[_disturbed];
   }
 
+  const DEFAULT_CHUNK_SIZE = 16_640;
```
bartlomieju, line 637: Have you experimented with greater size of this buffer?

ext/web/06_streams.js
```diff
@@ -4278,6 +4314,7 @@
     WeakMapPrototypeSet(countSizeFunctionWeakMap, globalObject, size);
   }
 
+  const _maybeRid = Symbol("maybeRid");
```
aapoalas, line 4317: Align with common symbol "language" and use `[[maybeRid]]` perhaps?

ext/web/06_streams.js
littledivy: This change in unrelated to the optimization. It's just moving existing code from `ext/net` to `ext/web/06_streams.js`.  
Though yes, we should try increasing the chunk size to smth like 64KB, once #14265 lands to actually benefit from >16kb sizes.

ext/http/lib.rs
```diff
@@ -664,6 +665,57 @@ async fn op_http_write_headers(
...
+      }
+      HttpResponseWriter::Closed => {
+        return Err(http_error("response already completed"))
+      }
+    };
+
+    let mut vec = vec![0u8; 16 * 1024];
+    let vec_ptr = vec.as_mut_ptr();
+    let buf = ZeroCopyBuf::new_temp(vec);
+    let nread = resource.clone().read(buf).await?;
```
ry, line 694: Is the vec stack allocated here? Wouldn't that be a problem across the await boundary?

maybe use [into_boxed_slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice)?

AaronO: > Is the vec stack allocated here? Wouldn't that be a problem across the await boundary?
> 
> maybe use [into_boxed_slice](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice)?

Vecs are always heap allocated, it's wrapped in a ZeroCopyBuf to satisfy the resource read trait sig.

As previously mentioned ZeroCopyBuf is conceptually similar to a Cow enum of Owned/Borrowed, Temp is hackish but I'll clean that up

ext/http/lib.rs
```diff
@@ -664,6 +665,57 @@ async fn op_http_write_headers(
...
+      HttpResponseWriter::Body(body_tx) => body_tx,
+      HttpResponseWriter::Headers(_) => {
+        return Err(http_error("no response headers"))
+      }
+      HttpResponseWriter::Closed => {
+        return Err(http_error("response already completed"))
+      }
+    };
+
+    let mut vec = vec![0u8; 16 * 1024];
```
ry, line 691: add TODO to increase this to 64k

ext/web/06_streams.js
littledivy: updated! and holy moly we are faster than Node now (see updated benchmarks above)

bartlomieju (approved):
LGTM, great work Divy!

lucacasonato (approved):
LGTM!

lucacasonato (changes requested):
Actually, a few issues I just realized this has that need to be addressed:

a) before attempting to stream the body using the rid, we need to make sure the stream is not locked
b) we need to lock the stream while we are reading
c) we need to close the stream once we are done reading

If we don't do that last thing, the underlying resource will not be automatically closed by the `ReadableStream` implementation. This also means that `op_http_write_resource` should not close the stream resource itself.

We'll need tests for these cases.

lucacasonato:
Can we get a test for the locking scenarios?

lucacasonato (approved):
LGTM!

ext/http/lib.rs
```diff
@@ -664,6 +665,56 @@ async fn op_http_write_headers(
...
+    let mut vec = vec![0u8; 64 * 1024];
+    let vec_ptr = vec.as_mut_ptr();
+    let buf = ZeroCopyBuf::new_temp(vec);
+    let nread = resource.clone().read(buf).await?;
+    if nread == 0 {
+      break;
+    }
+    // SAFETY: ZeroCopyBuf keeps the Vec<u8> alive.
+    let bytes =
+      Bytes::from_static(unsafe { std::slice::from_raw_parts(vec_ptr, nread) });
```
lucacasonato, line 700: Looking over this code again, this isn't safe. `ZeroCopyBuf` is moved into `resource.read()`, and as such can already be dropped at this point (add a `drop` handler to `ZeroCopyBuf` that prints something and observe).

ext/http/01_http.js
```diff
@@ -269,35 +269,50 @@
...
+            if (respBody.locked) {
+              throw new TypeError("ReadableStream is locked.");
             }
+            const _reader = respBody.getReader(); // Aquire JS lock.
+            await core.opAsync(
+              "op_http_write_resource",
+              streamRid,
+              resourceRid,
+            );
+            readableStreamClose(respBody); // Release JS lock.
```
dsherret, line 283: @littledivy if the above statement throws should this still be released?

littledivy: oh nice catch, yeah it should call `reader.cancel`

dsherret: Cool. We can fix it when getting this PR back into main.

littledivy: Fixed in https://github.com/denoland/deno/pull/14345

Issue #13831, opened 2022-03-04
Improve `AggregateError` console output


jaydenseric:
A standard [`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) instance has an `errors` property containing an array of the aggregated errors.

It would be really useful if Deno displayed these errors when printing an `AggregateError` instance in the console, much like how it displays the `cause` property.

For example, with `demo.mjs`:

```js
const aggregateError = new AggregateError([
  new Error("Error message 1."),
  new Error("Error message 2."),
], "Multiple errors.");

console.error(aggregateError);

throw aggregateError;
```

Run with Deno:

```
deno run demo.mjs
```

Currently results in:

```
AggregateError: Multiple errors.
    at file://[redacted]/demo.mjs:1:24
error: Uncaught AggregateError: Multiple errors.
const aggregateError = new AggregateError([
                       ^
    at file://[redacted]/demo.mjs:1:24
```

As you can see, it doesn't surface the messages of the aggregated errors.

bartlomieju:
@crowlKats could you take a look?

PR #14285, opened 2022-04-14
feat: Better formatting for AggregateError

bartlomieju:
Closes https://github.com/denoland/deno/issues/13831

This commit adds "aggregated" field to "deno_core::JsError" that stores
instances of "JsError" recursively to properly handle "AggregateError"
formatting. Appropriate logics was added to "PrettyJsError" and 
"console" API to format AggregateErrors.


bartlomieju:
@nayeemrmn please take a look

nayeemrmn:
I think it's noisy to show source lines for child errors (aggregated, cause). We should only show it for the main error, I implemented this in https://github.com/nayeemrmn/deno/commit/c18499bb047f9b2543f1e1b746096abbe3cf7be along with more fixes/cleanup.

bartlomieju:
> I think it's noisy to show source lines for child errors (aggregated, cause). We should only show it for the main error, I implemented this in [nayeemrmn@c18499b](https://github.com/nayeemrmn/deno/commit/c18499bb047f9b2543f1e1b746096abbe3cf7be) along with more fixes/cleanup.

Thanks, these are great changes. I merged your commit into this PR.

core/error.rs
```diff
@@ -104,6 +104,7 @@ pub struct JsError {
   pub frames: Vec<JsStackFrame>,
   pub source_line: Option<String>,
   pub source_line_frame_index: Option<usize>,
+  pub aggregated: Option<Vec<Box<JsError>>>,
```
littledivy, line 107: Redundant `Option<Vec<Box<T>>>`? It can just be a `Vec<T>`.

bartlomieju: It's not, this option allows distinction between AggregateError and regular Error.

littledivy: Oh, I meant the redundant use of the `Box<T>`. compile-time size shouldn't be an issue since Vec is heap-allocated.

bartlomieju: Good point. I'll remove these boxes

littledivy (approved):

Issue #14189, opened 2022-04-02
test: print error messages with a red color


jespertheend:
When a test fails, there's two things I want to know first:
1. **Which** test failed
2. **Why** the test failed

So I think it makes sense to put some extra emphasis on the error message.

from https://github.com/denoland/deno/discussions/14156:
> I think it's a lot easier to visually parse like this. Not sure how feasible it would be to print error messages in red though. Since error messages might have custom color formatting. Maybe they should only be printed in red if the error message does not contain any ansi codes.

![image](https://user-images.githubusercontent.com/2737650/161397135-a6379f34-239e-48ca-8fd9-21c930b0dfd6.png)

bartlomieju:
Ref https://github.com/denoland/deno/pull/14287 which is a prerequisite for this feature

PR #14287, opened 2022-04-14
feat(test): use structured data for JavaScript errors in tests

bartlomieju:
This commit rewrites test runner to send structured error data from JavaScript
to Rust instead of passing strings. This will allow to customize display of errors
in test report (which will be addressed in follow up commits).


bartlomieju:
@nayeemrmn please take a look

cli/tools/test.rs
```diff
@@ -285,8 +286,13 @@ impl PrettyTestReporter {
...
     );
 
-    if let Some(error_text) = result.error() {
-      for line in error_text.lines() {
+    if let Some(js_error) = result.error() {
+      // TODO(bartlomieju): use structured data here
+      let err_string = js_error
+        .to_string()
+        .trim_start_matches("Uncaught ")
+        .to_string();
```
nayeemrmn, line 294: I think we should implement the `JsError::aggregate` printing in `PrettyJsError` (addressing #13831) and use that formatting here.
```suggestion
      let err_string = PrettyJsError::create(js_error)
        .to_string()
        .trim_start_matches("Uncaught ")
        .to_string();
```
Using our colourful terminal error formatting would be good for test failures. #14189 can still be done on top.

`JsError::fmt()` should remain as it is (a simple implementation around `write!(f, "{}", self.stack)`) rather than being a third formatting implementation that has to be mindful of test output.

bartlomieju: @nayeemrmn shouldn't we do both? What I implemented here is not really specific to tests, it's just a way to print aggregate error without multiple "Uncaught" in the middle. I agree we should implement it for `PrettyJsError`, but not implementing it for `JsError` means that the only information you'll get will be `AggregateError` if it's not formatted by `PrettyJsError` later. 

Anyway, I'll look into doing `PrettyJsError` first.

bartlomieju: Reopened https://github.com/denoland/deno/pull/14285, but it suffers from the same problem of having to manually trim "Uncaught". Please take a look and let me know if that approach looks fine for you.

nayeemrmn: > What I implemented here is not really specific to tests

No I mean that this would become the only place that `JsError::fmt()` is exposed in Deno cli. I believe currently it's only there for consumers of core or debugging. IMO this code here should definitely defer to `PrettyJsError`, and `JsError::fmt()` should probably be kept minimal until there is some use/demand from core users, otherwise it is a third error formatting implementation to maintain.

bartlomieju: Sure, I'll remove that implementation

bartlomieju: Addressed in https://github.com/denoland/deno/pull/14285

nayeemrmn:
I think it's noisy to show source lines for child errors (aggregated, cause). We should only show it for the main error, I implemented this in https://github.com/nayeemrmn/deno/commit/c18499bb047f9b2543f1e1b746096abbe3cf7be along with more fixes/cleanup.

bartlomieju:
@nayeemrmn this is now done with changes from https://github.com/denoland/deno/pull/14285 applied. The next step would be to use different formatting for `PrettyJsError` to get rid of "internal" stack frames that are not useful in tests, but I suppose we can tackle that in a follow up - I'm thinking about adding a helper method on the `PrettyJsError` that would provide this alternative formatting. Please take a look and let me know what you think.

littledivy (approved):
LGTM

Perf nit: cloning `Box<JsError>` doesn't seem nessecary, it could be an (immutable) Rc instead.

nayeemrmn (approved):
LGTM

PR #13929, opened 2022-03-12
[WIP] feat(bench): add percentile output

bartlomieju:

PR #14305, opened 2022-04-17
feat(bench): new console reporter

### LINKED ISSUE #14022: bench: compare implementations
I'd like a way to use `Deno.bench` to define performance "tests" for competing implementations of a solution so that I can have confidence that one implementation remains faster than others when things change.

e g. A new Deno/V8 runtime could cause generators to out perform an array-based solution of an algorithm. A new library dependency version could speed up or slow down a function. etc.

This way I can write multiple implementations of a solution, declare one as the current winner and have it serve as the baseline implementation going forward, and have `deno bench` fail if an implementation performs better than it going forward.

The alternate implementations also serve as references as faster implementations of some algorithms are often harder to read and understand than naive implementations that may be easier to read but do not perform as well.

### LINKED ISSUE #14149: Default Deno.bench warmups may not be high enough for some CI runners
### Discussed in https://github.com/denoland/deno/discussions/14147

<div type='discussions-op-text'>

<sup>Originally posted by **GJZwiers** March 29, 2022</sup>
Recently I have been benchmarking with `Deno.bench` and having fun, although it seems that on some CI runners the default 1000 warmups is not sufficient. On the Ubuntu runner in Github Actions in particular there is a major difference in the measured time of whatever benchmark is run first, compared to other runners and my local machine.

For example, I am benching a JS and WASM implementation of a string distance algorithm. The results are very consistent on my local setup and the MacOS/Windows runners: JS is faster on relatively short strings while WASM is faster on long ones:

```
bench Myers 32bp string JavaScript ... 1000 iterations 1,035 ns/iter (700..59,100 ns/iter) ok (17ms)
bench Myers 32bp string WebAssembly ... 1000 iterations 2,726 ns/iter (2,400..8,900 ns/iter) ok (12ms)
bench Myers 1024bp string JavaScript ... 1000 iterations 425,071 ns/iter (401,300..565,900 ns/iter) ok (884ms)
bench Myers 1024bp string WebAssembly ... 1000 iterations 80,577 ns/iter (78,500..122,900 ns/iter) ok (166ms)
```

On the Ubuntu runner though, there is a time anomaly in the first bench, showing a much higher average `ns/iter`:

```
bench Myers 32bp string JavaScript ... 1000 iterations 10,783 ns/iter (9,300..88,499 ns/iter) ok (35ms)
bench Myers 32bp string WebAssembly ... 1000 iterations 6,350 ns/iter (3,400..60,500 ns/iter) ok (24ms)
bench Myers 1024bp string JavaScript ... 1000 iterations 673,224 ns/iter (556,596..3,331,975 ns/iter) ok (1s)
bench Myers 1024bp string WebAssembly ... 1000 iterations 152,748 ns/iter (118,299..4,364,268 ns/iter) ok (294ms)
```

I can actually shift the extra time in the CI to a different benchmark by having it run first:

```
bench test bench JavaScript ... 1000 iterations 9,218 ns/iter (1,400..96,701 ns/iter) ok (59ms)
bench Myers 32bp string JavaScript ... 1000 iterations 1,322 ns/iter (1,000..5,600 ns/iter) ok (16ms)
bench Myers 32bp string WebAssembly ... 1000 iterations 4,417 ns/iter (4,300..22,500 ns/iter) ok (33ms)
bench Myers 1024bp string JavaScript ... 1000 iterations 750,559 ns/iter (736,909..3,943,152 ns/iter) ok (1s)
bench Myers 1024bp string WebAssembly ... 1000 iterations 141,181 ns/iter (140,301..225,303 ns/iter) ok (290ms)
```

Using more warmups solves this problem, somewhere around 3000-5000 giving results that are more in line with my expectation:

```
bench Myers 32bp string JavaScript ... 1000 iterations 1,470 ns/iter (1,100..64,300 ns/iter) ok (47ms)
bench Myers 32bp string WebAssembly ... 1000 iterations 4,473 ns/iter (4,100..129,601 ns/iter) ok (17ms)
bench Myers 1024bp string JavaScript ... 1000 iterations 739,221 ns/iter (731,106..1,865,416 ns/iter) ok (1s)
bench Myers 1024bp string WebAssembly ... 1000 iterations 144,435 ns/iter (143,501..224,102 ns/iter) ok (298ms)
```

Don't know if anyone else has encountered something similar, but this made me wonder if the default warmups for `Deno.bench` should be higher, assuming more people will be using this in CI environments in the future.
</div>

evanwashere:
This commit replaces basic benchmarks reporter with a pretty table, and adds group and baseline options.

fixes #14022 
fixes #14149 
closes #13929 
closes #14015 

image


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -7,15 +7,10 @@ declare namespace Deno {
   export interface BenchDefinition {
     fn: () => void | Promise<void>;
     name: string;
+    group?: string;
```
bartlomieju, line 10: This could use a docstring to tell what it's useful for

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -7,15 +7,10 @@ declare namespace Deno {
...
-    /** Specify number of iterations benchmark should perform. Defaults to 1000. */
-    n?: number;
-    /** Specify number of warmup iterations benchmark should perform. Defaults
-     * to 1000.
-     *
-     * These iterations are not measured. It allows the code to be optimized
-     * by JIT compiler before measuring its performance. */
-    warmup?: number;
+    /** Benchmark should be used as the baseline for other benchmarks */
+    baseline?: boolean;
```
bartlomieju, line 13: What happens if multiple cases use `baseline: true`? Can you add it to the docstring?

cli/Cargo.toml
```diff
@@ -74,6 +74,7 @@ import_map = "=0.9.0"
 jsonc-parser = { version = "=0.19.0", features = ["serde"] }
 libc = "=0.2.121"
 log = { version = "=0.4.14", features = ["serde"] }
+mitata = '=0.0.5'
```
bartlomieju, line 77: I checked the source code for this crate and it seems there's quite a lot of overlap with code that's already present in Deno (coloring, `BenchReporter`). It's not a big deal but I would suggest to directly write that source code in `cli/tools/bench.rs`, then we could get rid of some of the intermediate abstractions like `BenchReporter`. 

Unless you're planning on developing the crate further and want to potentially introduce more features to `deno bench` in the future? TIOLI

cli/Cargo.toml
```diff
@@ -74,6 +74,7 @@ import_map = "=0.9.0"
...
 node_resolver = "=0.1.1"
 notify = "=5.0.0-pre.14"
 num-format = "=0.4.0"
```
bartlomieju, line 80: I believe this crate can now be removed

cli/tests/testdata/bench/group_baseline.out
```diff
@@ -0,0 +1,15 @@
...
+benchmark         time (avg)             (min … max)       p75       p99      p995
+---------------------------------------------------- -----------------------------
+noop [WILDCARD] [WILDCARD]/iter[WILDCARD]([WILDCARD] … [WILDCARD]) [WILDCARD]
+parse url 2x [WILDCARD] [WILDCARD]/iter[WILDCARD]([WILDCARD] … [WILDCARD]) [WILDCARD]
+parse url 6x [WILDCARD] [WILDCARD]/iter[WILDCARD]([WILDCARD] … [WILDCARD]) [WILDCARD]
+
+summary
+  parse url 2x
+   [WILDCARD]x times slower than noop
+   [WILDCARD]x times faster than parse url 6x
```
bartlomieju, line 15: Nice 👍

cli/tools/bench.rs
```diff
@@ -53,214 +49,311 @@ struct BenchSpecifierOptions {
...
-    current_bench: &BenchMeasures,
-  );
-  fn report_summary(&mut self, summary: &BenchSummary, elapsed: &Duration);
+  fn report_result(&mut self, result: &BenchResult);
 }
 
-struct PrettyBenchReporter {
-  echo_output: bool,
+#[allow(dead_code)]
+struct JSONReporter {}
```
bartlomieju, line 152: It looks like there's no CLI option to enable this reporter, please remove for now, or add a TODO comment

bartlomieju (commented):
Nice work @evanwashere, here's my initial review. The new report looks fantastic!

I want to play more with these changes before landing - especially want to check if we can now rewrite `cli/bench/deno_common.js` to leverage `deno bench`.

cli/Cargo.toml
evanwashere: It's cli and rust port of https://npmjs.com/mitata, but I can move it into deno repo in a follow-up pr.

runtime/js/40_testing.js
```diff
@@ -823,37 +844,227 @@
...
+          continue;
+        }
+
+        c++;
+        wavg += t2;
+        budget -= t2;
+      }
+
+      wavg /= c;
+      break warmup;
```
dsherret, line 897: Could we remove all these `break <label-name>` statements? It's not very standard in the codebase and I don't think they really do anything in this case?

runtime/js/40_testing.js
```diff
@@ -1013,32 +1205,31 @@
       createTestFilter(filter),
     );
 
+    let groups = new Set([undefined]);
```
dsherret, line 1208: Why is undefined added here? I think it will be added if needed on line 1213?

runtime/js/40_testing.js
```diff
@@ -1013,32 +1205,31 @@
...
+
+    for (const bench of benchmarks) {
+      bench.group ||= undefined;
+      groups.add(bench.group);
+    }
+
+    groups = ArrayFrom(groups);
+    ArrayPrototypeSort(
+      benchmarks,
+      (a, b) => groups.indexOf(a.group) - groups.indexOf(b.group),
```
dsherret, line 1219: Is this sorting intended to be stable? I believe it will be unstable because it originated from a `Set`?

runtime/js/40_testing.js
```diff
@@ -823,37 +844,227 @@
...
+      max,
+      p75: all[MathCeil(n * (75 / 100)) - 1],
+      p99: all[MathCeil(n * (99 / 100)) - 1],
+      avg: !t ? (avg / n) : MathCeil(avg / n),
+      p995: all[MathCeil(n * (99.5 / 100)) - 1],
+      p999: all[MathCeil(n * (99.9 / 100)) - 1],
+    };
+  }
+
+  function benchMeasureSync(t, fn, step) {
```
dsherret, line 867: Nitpick: Would be helpful to give `t` a more descriptive name.

runtime/js/40_testing.js
```diff
@@ -823,37 +844,227 @@
...
+      step.warmup = true;
+      let iterations = 20;
+      let budget = 10 * 1e6;
+
+      while (0 < budget || 0 < iterations--) {
+        const t1 = benchNow();
+
+        fn();
+        const t2 = benchNow() - t1;
+        if (0 > t2) {
```
dsherret, line 886: Why is this additional logic necessary? When would the elapsed duration be less than 0? (maybe I'm just misreading this because it's the end of the day and the constant is on the left instead of the right---we should change that)

runtime/js/40_testing.js
```diff
@@ -823,37 +844,227 @@
...
+      let c = 0;
+      step.warmup = true;
+      let iterations = 20;
+      let budget = 10 * 1e6;
+
+      while (0 < budget || 0 < iterations--) {
+        const t1 = benchNow();
+
+        fn();
+        const t2 = benchNow() - t1;
```
dsherret, line 885: Nitpick: `t2` is not a moment in time like `t1`, but instead a duration.

runtime/js/40_testing.js
```diff
@@ -823,37 +844,227 @@
...
+      }
+
+      wavg /= c;
+      break warmup;
+    }
+
+    measure: {
+      step.warmup = false;
+
+      if (wavg > 10_000) {
```
dsherret, line 903: Some of these conditions have the constant on the right like this and others do the opposite ("yoda conditions"—ex. line 907), which is not typical in the codebase. Could we change them all to this style where the constant is on the right? It's harder to read when it's not this way because it's not typical.

runtime/js/40_testing.js
```diff
@@ -823,37 +844,227 @@
...
+      }
+
+      break measure;
     }
 
+    all.sort(compareMeasurements);
+    return benchStats(n, wavg > 10_000, avg, min, max, all);
+  }
+
+  async function benchMeasureAsync(t, fn, step) {
```
dsherret, line 953: There's a lot of duplicate code between `benchMeasureSync` and `benchMeasureAsync`. I think we only need to duplicate the measurement code though?

```
const t1 = benchNow();
await fn();
const elapsed = benchNow() - t1;
```

...and...

```
const t1 = benchNow();
fn();
const elapsed = benchNow() - t1;
```

dsherret (commented):
It looks a lot nicer!

runtime/js/40_testing.js
evanwashere: Using perf.now() can lead to negative numbers, but we use bench time op here so it shouldn't be the case.

runtime/js/40_testing.js
evanwashere: undefined makes sure that ungrouped benchmarks are placed first in sorted array

bartlomieju:
@evanwashere two pieces of feedback from the team that were surfaced internally:
- would it be possible to keep the width of the table to 80 columns?
- would it be possible to keep the same unit in the whole table? (ie. the bottom table changes unit from nanosecs to microsecs, can the unit be fixed in such case?)

runtime/js/40_testing.js
dsherret: Would you be able to add a comment explaining why along with comments answering the other questions? Thanks!

runtime/js/40_testing.js
dsherret: But an array is created from the set. Don't you need to sort the groups array?

bartlomieju:
> @evanwashere two pieces of feedback from the team that were surfaced internally:
> 
> * would it be possible to keep the width of the table to 80 columns?
> * would it be possible to keep the same unit in the whole table? (ie. the bottom table changes unit from nanosecs to microsecs, can the unit be fixed in such case?)

Actually I believe the second point is not feasible, since cases are benched one by one, so we don't know what unit should be used for following cases.

evanwashere:
> - would it be possible to keep the width of the table to 80 columns?

all columns except benchmark name can be disabled through reporter options (eg. sacrificing percentiles column to fit on screen)

unless you meant locking it to 80 width?

> - would it be possible to keep the same unit in the whole table? (ie. the bottom table changes unit from nanosecs to microsecs, can the unit be fixed in such case?)

making this possible would conflict with keeping table width constrained and columns aligned with values.

runtime/js/40_testing.js
evanwashere: Set maintains insertion order

bartlomieju:
> unless you meant locking it to 80 width?

Yes, locking it to 80 width

runtime/js/40_testing.js
evanwashere: https://doc.rust-lang.org/std/time/struct.Instant.html#monotonicity

another issue is losing precision when converting nanoseconds from u128 to f64

evanwashere:
> Yes, locking it to 80 width

this is not possible currently but could be added

runtime/js/40_testing.js
```diff
@@ -823,37 +844,166 @@
...
+          fn();
+          const iterationTime = benchNow() - t1;
+
+          n++;
+          avg += iterationTime;
+          budget -= iterationTime;
+          all.push(iterationTime);
+          if (iterationTime < min) min = iterationTime;
+          if (iterationTime > max) max = iterationTime;
+        }
```
bartlomieju, line 928: Did some measurements for `cli/bench/deno_common.js` and its version written using `Deno.bench`. There's still a huge discrepancy in the taken measurements (about 2-3x). Discussed with @AaronO and he suggested that we could alleviate it twofold:
a) Don't do two measurements per iteration, but instead one measurement per iter - ie. use previous iteration "end time" as a start time for next iteration. There are a bunch of math/array operation afterward, but these should be negligible compared to a syscall to get the current time
b) sample the cost of the time measurement - ie. run some loops of `benchNow` to see how much time it takes

@evanwashere do you think you could incorporate these changes in the PR (or either one of them)?

evanwashere: You likely forgot to disable sanitizers, slow benchmarks loop only has 50-200ns overhead  and fast loop reaches vm speed limit

image

bartlomieju (approved):
LGTM, thank you @evanwashere, this is a great improvement

PR #14306, opened 2022-04-17
fix: panic when trying to pledge permissions before restoring previous pledge

bartlomieju:
This commit fixes and edge case, where testing/benching code could pledge new
permission set before restoring the previous pledge. 

Appropriate panics were added and tests that assert that process is killed
in case of "recursive pledge".


dsherret (approved):
LGTM

PR #14314, opened 2022-04-18
perf(fmt/lint): incremental formatting and linting

### LINKED ISSUE #10906: deno fmt - Incremental formatting backed by a cache
I was thinking we could speed up `deno fmt` by having it backed by a single file of formatted file hashes (stored behind the scenes in the deno user data directory).

So you have the deno fmt cache that stores the version of deno that just ran deno fmt along with a bunch of formatted file hashes in it limited to the X most recent hashes. If the hash of the file appears in the cache then it doesn't need to be formatted again.

This has the negative of being slightly slower on the first run, but in dprint I do something similar and haven't noticed much of a slowdown on the first run compared to not having incremental (it's using https://crates.io/crates/twox-hash to do the hashing). A huge advantage is that most runs will complete almost instantaenously.

dsherret:
## Formatting

Before on deno_std:

```ps
> Measure-Command { deno fmt } | Select-Object -Property TotalMilliseconds
Checked 929 files

TotalMilliseconds
-----------------
          414.208
> Measure-Command { deno fmt } | Select-Object -Property TotalMilliseconds
Checked 929 files

TotalMilliseconds
-----------------
         415.6013
```

After (see second run):

```ps
> Measure-Command { ../deno/target/release/deno fmt } | Select-Object -Property TotalMilliseconds
Checked 929 files

TotalMilliseconds
-----------------
         415.8299
> Measure-Command { ../deno/target/release/deno fmt } | Select-Object -Property TotalMilliseconds
Checked 929 files

TotalMilliseconds
-----------------
         118.6008
```

## Linting

Before on deno_std:

```
> Measure-Command { deno lint } | Select-Object -Property TotalMilliseconds
Checked 873 files

TotalMilliseconds
-----------------
         464.5036
> Measure-Command { deno lint } | Select-Object -Property TotalMilliseconds
Checked 873 files

TotalMilliseconds
-----------------
         461.2106
```

After (see second run... first run is just by chance faster):

```
> Measure-Command { ../deno/target/release/deno lint } | Select-Object -Property TotalMilliseconds
Checked 873 files

TotalMilliseconds
-----------------
         451.739

V:\deno_std [main ↑10 +0 ~1 -0 !]
> Measure-Command { ../deno/target/release/deno lint } | Select-Object -Property TotalMilliseconds
Checked 873 files

TotalMilliseconds
-----------------
          114.018
```

This will be a big performance improvement if people cache their DENO_DIR on the CI too (the 400ms above is with 20 cores which would be much slower with only 2 cores on GH actions). Edit: Actually, the cache would need to be keyed on the CLI version... I'm not sure how that could work.

Closes #10906


cli/file_watcher.rs
```diff
@@ -25,29 +25,39 @@ const CLEAR_SCREEN: &str = "\x1B[2J\x1B[1;1H";
 const DEBOUNCE_INTERVAL: Duration = Duration::from_millis(200);
 
 struct DebouncedReceiver {
+  // The `recv()` call could be used in a tokio `select!` macro,
+  // and so we store this state on the struct to ensure we don't
+  // lose items if a `recv()` never completes
+  received_items: HashSet<PathBuf>,
```
dsherret, line 31: Extracted out this file's change to #14317, so we should merge that first.

bartlomieju:
> This will be a big performance improvement if people cache their DENO_DIR on the CI too (the 400ms above is with 20 cores which would be much slower with only 2 cores on GH actions).

I guess that information should be added to https://deno.land/manual@v1.20.6/continuous_integration#caching-dependencies

cli/tools/fmt/incremental_cache.rs
```diff
@@ -0,0 +1,294 @@
...
+      if let Some(hash) = cache.get_source_hash(path) {
+        previous_hashes.insert(path.to_path_buf(), hash);
+      }
+    }
+
+    let (sender, mut receiver) =
+      tokio::sync::mpsc::unbounded_channel::<ReceiverMessage>();
+
+    // sqlite isn't `Sync`, so we do all the updating on a dedicated thread
+    let handle = tokio::task::spawn(async move {
```
bartlomieju, line 57: `tokio::task::spawn` doesn't create a dedicated thread, it creates a new task that will be run on the same runtime instance (ie. main thread)

cli/tools/fmt/incremental_cache.rs
```diff
@@ -0,0 +1,294 @@
...
+      .ok()?;
+    hash.parse::<u64>().ok()
+  }
+
+  pub fn set_source_hash(
+    &self,
+    path: &Path,
+    source_hash: u64,
+  ) -> Result<(), AnyError> {
+    let mut stmt = self.conn.prepare_cached("INSERT OR REPLACE INTO incrementalcache (file_path, state_hash, source_hash) VALUES (?1, ?2, ?3)")?;
```
bartlomieju, line 145: Ditto

cli/tools/fmt/mod.rs
```diff
@@ -372,6 +412,66 @@ async fn format_source_files(
...
+        count += 1;
+        if count == 5 {
+          bail!(
+            concat!(
+              "Formatting not stable. Bailed after {} tries. Please report this ",
+              "as it indicates a bug in the formatter where it formats the file ",
+              "differently each time."
+            ),
+            count
+          )
```
bartlomieju, line 467: Use panic! here instead? That way we'll get nice report from our panic hook

bartlomieju (commented):
Can this feature be generalized so it could be leveraged by linter as well?

cli/tools/fmt/incremental_cache.rs
dsherret: I think that's ok because all the work is being done on blocking threads? I should update the comment to remove the word thread.

cli/tools/fmt/mod.rs
dsherret: I think it's still nice to let formatting continue, but then someone could ignore this file?

cli/tools/fmt/incremental_cache.rs
bartlomieju: Yes, I think that is fine

cli/tools/fmt/mod.rs
dsherret: I'll just change it to panic actually.

cli/tools/lint.rs
```diff
@@ -147,6 +149,12 @@ pub async fn lint(flags: Flags, lint_flags: LintFlags) -> Result<(), AnyError> {
   };
 
   let operation = |paths: Vec<PathBuf>| async {
+    let incremental_cache = Arc::new(IncrementalCache::new(
+      &ps.dir.lint_incremental_cache_db_file_path(),
+      // use a hash of the rule names in order to bust the cache
+      &lint_rules.iter().map(|r| r.code()).collect::<Vec<_>>(),
+      &paths,
+    )?);
```
dsherret, line 157: Oh, this doesn't work for linting for some reason. My guess would be the lint rules order is not stable and so I will sort it.

dsherret: Fixed now.

bartlomieju (approved):
LGTM

PR #14317, opened 2022-04-18
fix: `--watch` was losing items

dsherret:
We need to ensure this works if the `recv()` never completes due to being used in a `tokio::select`


bartlomieju (approved):
I guess there's no reliable way to test this?

dsherret:
@bartlomieju #14314 will be flaky without this

PR #14320, opened 2022-04-19
Remove the DENO_FUTURE_CHECK warning

lucacasonato:
It has good intentions, but it is a really terrible user experience. As
such we shouldn't print this warning.


bartlomieju (approved):
I don't have a strong opinion about it, but seems to me that mentioning this only in the blog posts might be not enough... That said, LGTM, if anyone has objections, please add "Request changes" comment

dsherret (approved):
👍

PR #14325, opened 2022-04-19
feat(ext/http): stream auto resp body compression

lucacasonato:
This commit adds support for auto response body compression for
streaming bodies.

Alternative to https://github.com/denoland/deno/tree/compression_stream
that uses `async_compression` and tokio's io traits.

cc @AaronO @satyarohith 

Fixes #13835

PR #14326, opened 2022-04-19
experiment(http): tower compression

### LINKED ISSUE #13835: Streaming HTTP response body compression
Fixed length response bodies were done in https://github.com/denoland/deno/pull/13769

AaronO:
Simpler, more composable & adds stream support "for free", fixes #13835

Invariants we may want to test are preserved (across main and alt impls, i.e: #14325, [tree/compression_stream](https://github.com/denoland/deno/tree/compression_stream)):
- http performance / throughput
- etag weakening ? (not done atm in this PR, but could be)
- compression predicates (~same responses are compressed)


ext/http/lib.rs
```diff
@@ -101,6 +102,36 @@ impl From<tokio::net::unix::SocketAddr> for HttpSocketAddr {
   }
 }
 
+// implements tower_http::compression's Predicate
+fn not_uncompressible(
```
lucacasonato, line 106: not_uncompressible == is_compressible? double negation is very weird to read IMO

lucacasonato (commented):
Does this handle `vary` and `etag` headers like we do now?

ext/http/lib.rs
AaronO: Yeah agreed, had fixed this yesterday but hadn't pushed

bartlomieju:
Do we still want to land this PR, given that #14345 landed?

PR #14327, opened 2022-04-20
feat(runtime/workers): add `close` event

duanckham:
The `close` event is emitted once the worker has closed. That will help the developer to handle the worker exit. And it is a little something like https://nodejs.org/api/worker_threads.html#event-exit.


andreubotella:
Deno's implementation of `Worker` follows the HTML spec, and the spec doesn't have such a `close` event. The relevant issue is whatwg/html#5869.

duanckham:
> 

😫

PR #14330, opened 2022-04-20
Revert "perf(http): optimize `ReadableStream`s backed by a resource"

lucacasonato:
There is actually unsafe `unsafe` code in here that I have not been able to trivially "safify".

As such we should revert, and re-land once the unsafe `unsafe` code has been made safe.

Reverts denoland/deno#14284

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -854,45 +854,6 @@ Deno.test({ permissions: { net: true } }, async function httpServerPanic() {
   listener.close();
 });
 
-Deno.test(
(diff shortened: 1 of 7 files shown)


bartlomieju (approved):

littledivy (approved):

lucacasonato:
#14331 invalidates this.

lucacasonato:
@kt3k does it work with #14331?

kt3k:
will try when canary binary uploaded

kt3k:
It seems fixed! (tried with release & debug builds on macos)

PR #14331, opened 2022-04-20
core: introduce `resource.read_return`

littledivy:
Fixes https://github.com/denoland/deno/pull/14284#discussion_r854101629

Comes with a performance penalty but still faster than Deno 1.20.6 and Node:
```
  8 threads and 20 connections
    Latency    16.64ms    1.62ms  39.96ms   83.43%
    Req/Sec   120.46      9.53   141.00     76.24%
  9695 requests in 10.10s, 47.41GB read
Requests/sec:    959.93
Transfer/sec:      4.69GB
```

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -688,16 +688,13 @@ async fn op_http_write_resource(
       }
     };
 
-    let mut vec = vec![0u8; 64 * 1024];
-    let vec_ptr = vec.as_mut_ptr();
+    let mut vec = vec![0u8; 64 * 1024]; // 64KB
     let buf = ZeroCopyBuf::new_temp(vec);
-    let nread = resource.clone().read(buf).await?;
+    let (nread, buf) = resource.clone().read_return(buf).await?;
     if nread == 0 {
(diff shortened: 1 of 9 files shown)


ext/http/lib.rs
```diff
@@ -688,16 +688,13 @@ async fn op_http_write_resource(
...
     let buf = ZeroCopyBuf::new_temp(vec);
-    let nread = resource.clone().read(buf).await?;
+    let (nread, buf) = resource.clone().read_return(buf).await?;
     if nread == 0 {
       break;
     }
-    // SAFETY: ZeroCopyBuf keeps the Vec<u8> alive.
-    let bytes =
-      Bytes::from_static(unsafe { std::slice::from_raw_parts(vec_ptr, nread) });
+    let bytes = Bytes::copy_from_slice(&buf);
```
lucacasonato, line 697: Downcast it to a `Vec<u8>` and then `Bytes::from`. Should save 1 copy.

lucacasonato (approved):

Issue #14226, opened 2022-04-06
Add global `clear` function to the REPL?


dsherret:
Browsers have `console.clear()` available in the dev tools as `clear()`. Perhaps we should do the same as it's a bit more accessible?

Related to #12896 and I think this would be a bit nicer because it aligns with chrome and firefox.

https://discord.com/channels/684898665143206084/684911491035430919/961241552217325639

bartlomieju:
Somewhat related to https://github.com/denoland/deno/issues/12896 ?

PR #14332, opened 2022-04-20
feat(repl): add global clear() function

cjihrig:
This commit adds a `clear()` function in the REPL which works similar to `console.clear()`.

Fixes: https://github.com/denoland/deno/issues/14226

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -761,3 +761,21 @@ fn eval_file_flag_multiple_files() {
   assert!(out.contains("helloFOO"));
   assert!(err.contains("Download"));
 }
+
+#[test]
+fn pty_clear_function() {
+  util::with_pty(&["repl"], |mut console| {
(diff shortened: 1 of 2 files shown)


dsherret (approved):
LGTM!

As mentioned in the issue, I think this makes sense to add because both Firefox and Chrome have this in their REPL.

bartlomieju (approved):
LGTM too

cli/tests/integration/repl_tests.rs
```diff
@@ -776,6 +776,6 @@ fn pty_clear_function() {
     assert!(output.contains("[1;1H"));
     assert!(output.contains("undefined"));
     assert!(output.contains("const clear = 1 + 2;"));
-    assert!(output.contains("3"));
+    assert!(output.contains('3'));
```
lucacasonato, line 779: Wow, clippy is a little aggressive today... :D

lucacasonato (approved):
LGTM!

PR #14335, opened 2022-04-20
fix(compat): canonicalize import paths

rjeli:
Node resolution algo "follows symlinks"
See https://github.com/nodejs/node/blob/61fefe1959a8ab9e796f409346ff619c14036c81/lib/internal/modules/esm/resolve.js#L409

PNPM relies on this behavior to have a "nested node_modules" without tons of bloated duplicated packages, see https://pnpm.io/blog/2020/05/27/flat-node-modules-is-not-the-only-way for full explanation

This PR canonicalizes import path to replicate node resolve.js behavior

### DIFF
--- a/cli/compat/esm_resolver.rs
+++ b/cli/compat/esm_resolver.rs
@@ -261,7 +261,8 @@ fn finalize_resolution(
     ));
   }
 
-  Ok(resolved)
+  // OK to unwrap, path came from to_file_path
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks for the PR, but we will need tests to land it.

rjeli:
The CI tests or more? Can i help?

rjeli:
oh, you mean add tests to the suite. ok 👍🏻

PR #14337, opened 2022-04-20
chore(ci): use Rust 1.60.0

bartlomieju:


lucacasonato:
I already tried in #14260

PR #14341, opened 2022-04-21
chore(scripts): allow running version_bump workflow without releasing deno_std yet

dsherret:
The instructions say to do a deno_std release first and then start with the deno release, but when someone doesn't do that make sure the deno_std version gets bumped anyway.

This just now always increments the minor version.

### DIFF
--- a/tools/release/01_bump_crate_versions.ts
+++ b/tools/release/01_bump_crate_versions.ts
@@ -97,24 +97,19 @@ async function getGitLog() {
 }
 
 async function updateStdVersion() {
-  const newStdVersion = await getLatestStdVersion();
   const compatFilePath = path.join(cliCrate.folderPath, "compat/mod.rs");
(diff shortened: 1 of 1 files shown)


tools/release/01_bump_crate_versions.ts
```diff
@@ -97,24 +97,17 @@ async function getGitLog() {
...
-    "https://raw.githubusercontent.com/denoland/deno_std/main/version.ts";
-  const result = await fetch(url);
-  const text = await result.text();
-  const version = /"([0-9]+\.[0-9]+\.[0-9]+)"/.exec(text);
-  if (version == null) {
-    throw new Error(`Could not find version in text: ${text}`);
-  } else {
-    return version[1];
-  }
-}
```
dsherret, line 120: Reduces some complexity too. If this makes a mistake then someone can correct it in the PR.

bartlomieju (approved):

PR #14345, opened 2022-04-21
Reland "feat(ext/http): stream auto resp body compression"

littledivy:
Relands #14325

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -590,14 +590,14 @@ async fn op_http_write_headers(
         builder = builder.header("etag", value.as_slice());
       }
     }
+    // Drop 'content-length' header. Hyper will update it using compressed body.
+    if let Some(headers) = builder.headers_mut() {
+      headers.remove("content-length");
+    }
   } else if let Some(value) = etag_header {
     builder = builder.header("etag", value.as_slice());
   }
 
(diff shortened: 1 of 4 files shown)


littledivy:
```
./http/file_server_test.ts > file_server file responses includes correct etag
AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   W/f97172fa
+   f97172fa


  throw new AssertionError(message);
        ^
    at assertEquals (file:///Users/divy/gh/deno_std/testing/asserts.ts:179:9)
    at file:///Users/divy/gh/deno_std/http/file_server_test.ts:866:7
```

Only this one `std/http/file_server` test fails.

lucacasonato (approved):
LGTM. That std test just needs to be updated.

ext/http/lib.rs
```diff
@@ -545,56 +547,60 @@ async fn op_http_write_headers(
...
+      }
+    }
+  } else if let Some(value) = etag_header {
+    builder = builder.header("etag", value.as_slice());
+  }
 
+  // Drop 'content-length' header. Hyper will update it using compressed body.
+  if let Some(headers) = builder.headers_mut() {
+    headers.remove("content-length");
+  }
```
lucacasonato, line 600: Should only be done `if should_compress`

PR #14346, opened 2022-04-21
Reland "perf(http): optimize ReadableStreams backed by a resource"

littledivy:
and https://github.com/denoland/deno/pull/14331. 

Now that the std tests are passing after the fix in https://github.com/denoland/deno/pull/14345. Also fixes https://github.com/denoland/deno/pull/14284#discussion_r854658149


ext/http/01_http.js
```diff
@@ -269,16 +269,19 @@
...
             }
+            const reader = respBody.getReader(); // Aquire JS lock.
             try {
-              await core.opAsync("op_http_write", streamRid, value);
+              await core.opAsync(
+                "op_http_write_resource",
+                streamRid,
+                resourceRid,
+              );
+              readableStreamClose(respBody); // Release JS lock.
```
bartlomieju, line 284: This should be done in `finally` block

bartlomieju (changes requested):

cli/tests/unit/http_test.ts
```diff
@@ -854,6 +854,45 @@ Deno.test({ permissions: { net: true } }, async function httpServerPanic() {
...
+    const reader = await file.readable.getReader();
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      assert(value);
+    }
+
+    try {
+      await respondWith(new Response(file.readable));
+      fail("The stream should've been locked");
```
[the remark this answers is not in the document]
littledivy: oh lol, it's done everywhere in the fetch tests: https://github.com/denoland/deno/blob/bfa4ed92bcd119fa5c8fdb6b444b83a092dccca9/cli/tests/unit/fetch_test.ts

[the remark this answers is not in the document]
dsherret: Hah, yeah we should fix those. I think it’s probably best to use `await assertRejects(() => …);`

ext/http/01_http.js
```diff
@@ -290,13 +292,40 @@
               }
               await reader.cancel(error);
               throw error;
+            } finally {
+              readableStreamClose(respBody); // Release JS lock.
```
lucacasonato, line 296: Does this throw if called after `await reader.cancel()`?

ext/http/01_http.js
```diff
@@ -290,13 +292,40 @@
...
+                }
+                await reader.cancel(error);
+                throw error;
+              }
+            }
+
+            try {
+              await core.opAsync("op_http_shutdown", streamRid);
+            } catch (error) {
+              await reader.cancel(error);
```
lucacasonato, line 326: Does this throw if called after `readableStreamClose(respBody)`?

ext/http/01_http.js
littledivy: It's unreachable since `catch` throws after `await reader.cancel()`.

ext/http/01_http.js
littledivy: This is unreachable since `op_http_write_resource` performs the shutdown itself.

ext/http/01_http.js
lucacasonato: No, `finally` will still run after that `throw`: `try { throw "a" } catch { throw "b" } finally { console.log("c") }`.

ext/http/01_http.js
lucacasonato: Oh right. 👍

ext/http/01_http.js
littledivy: ooh TIL. `readableStreamClose` does throw, since `reader.cancel` closes it already. Maybe I should move it back to the `try` block?

lucacasonato: SGTM

lucacasonato (approved):
LGTM

PR #14358, opened 2022-04-21
fix(runtime/js/spawn): Pass stdio options for spawn() and spawnSync()

nayeemrmn:

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1442,6 +1442,7 @@ declare namespace Deno {
   /**
    * Executes a subprocess, waiting for it to finish and
    * collecting all of its output.
+   * Will throw an error if `stdin: "piped"` is passed.
    *
    * ```ts
    * const { status, stdout, stderr } = await Deno.spawn(Deno.execPath(), {
@@ -1463,6 +1464,7 @@ declare namespace Deno {
   /**
    * Synchronously executes a subprocess, waiting for it to finish and
    * collecting all of its output.
+   * Will throw an error if `stdin: "piped"` is passed.
    *
    * ```ts
    * const { status, stdout, stderr } = Deno.spawnSync(Deno.execPath(), {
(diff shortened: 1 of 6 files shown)


crowlKats:
This is intentional, but after working on porting all `Deno.run` usages to the new subprocess API, I agree we should allow it

nayeemrmn:
Oh, I assumed it was unintentional because the types allowed it. I expected this to work as it seems like a reasonable use of the shorthand:
```ts
const { stdout } = await Deno.spawn("cat", { stdin: "inherit" });
```

crowlKats:
@nayeemrmn could you also update the tsdoc on the typings? There is a line that states that stdin options are ignored on spawn/spawnSync

crowlKats:
One question remains: what happens when `stdin` is set to `"piped`"? that will cause issues

bartlomieju:
> One question remains: what happens when `stdin` is set to `"piped`"? that will cause issues

I think it makes no sense to have "piped" setting in `Deno.spawnSync` API as there would be no way to actually write something to stdin, but it could be used in async version.

nayeemrmn:
> I think it makes no sense to have "piped" setting in `Deno.spawnSync` API as there would be no way to actually write something to stdin, but it could be used in async version.

But there's no handle returned on the stdin. We should just map `"piped"` to `"null"` there and note it in the docs.

crowlKats:
> I think it makes no sense to have "piped" setting in `Deno.spawnSync` API as there would be no way to actually write something to stdin, but it could be used in async version.

yes, but what would be the beviour? right now, as the PR is, if someone sets stdin to `"piped"`, a stdin resource will be added that has to be manually closed thorugh `spawnChild`, however the user wont be able to close that resource

bartlomieju:
Ah this is good point. In such case Nayeem's suggestion seems reasonable. IIRC Leo wanted to add `input` option as well that would be passed to stdin and then close it.

nayeemrmn:
Yeah, I think in the future we should have:
```ts
export interface SpawnOptions {
  stdin?: "piped" | "inherit" | "null" | Uint8Array;
}
```

crowlKats:
Could you add a test for disallowed stdin?

crowlKats (approved):
LGTM, thanks

littledivy:
I recently hit this for node:child_process. Piped stdio absolutely makes sense, it means the parent's stdio is piped to the child (not the other way around). 

Removing the stdin check in `outputSync` fixes #23051

PR #14359, opened 2022-04-22
fix(lsp): correct positions in some scenarios

dsherret:
1. Keeps all the script snapshots in JavaScript and let TS handle it 
    - Should be ~~faster~~ essentially the same speed and use less memory. Also, lets the TS language service handle updating a file or not.
    - This `ts.ScriptSnapshot.fromString` doesn't do the incremental parsing like before though. It seems tsserver uses this function so maybe best to copy tsserver? It removes a lot of complex code so I think it's worth it.
2. This had the side effect of exposing a bug with multi-byte characters.

Will wait for Kitson to get back before merging.


cli/tests/integration/lsp_tests.rs
```diff
@@ -4284,7 +4286,7 @@ fn lsp_performance() {
     .unwrap();
   assert!(maybe_err.is_none());
   if let Some(res) = maybe_res {
-    assert_eq!(res.averages.len(), 14);
+    assert_eq!(res.averages.len(), 13);
```
dsherret, line 4289: Lower because I removed some ops.

cli/tests/integration/lsp_tests.rs
```diff
@@ -1444,11 +1446,11 @@ fn lsp_hover_change_mbc() {
       "range": {
         "start": {
           "line": 2,
-          "character": 13,
+          "character": 15,
         },
         "end": {
           "line": 2,
-          "character": 14,
+          "character": 16,
```
dsherret, line 1453: The numbers specified for this test were previously wrong. Please double check what I've done here. Everything is 0-indexed and the language server protocol uses utf16 character indexes to make life complicated.

cli/lsp/tsc.rs
```diff
@@ -2669,13 +2531,22 @@ fn op_is_cancelled(state: &mut OpState) -> Result<bool, AnyError> {
 fn op_load(
   state: &mut OpState,
   args: SpecifierArgs,
-) -> Result<Option<String>, AnyError> {
+) -> Result<Value, AnyError> {
```
dsherret, line 2534: This `op_load` was previously never called. That's why its return type not being correct didn't matter.

dsherret:
By the way, I think there's a chance this might fix #13657.

dsherret:
By the way, I think there's a chance this might fix #13657.

kitsonk:
Seems logical.

It would be good to get a baseline of the big file edits LSP bench test off of main and this PR to "prove" the performance improvement (or at least to prove no performance regression).  We don't measure memory consumption on the benchmark tests for the LSP, maybe we should find a way?

We should implement the `.fromString()` I suspect in our snapshots as well.

dsherret:
> It would be good to get a baseline of the big file edits LSP bench test off of main and this PR to "prove" the performance improvement (or at least to prove no performance regression)

@kitsonk good idea! I just tried it out and it seems the same:

Main:

```
-> Start benchmarking lsp
   - Simple Startup/Shutdown
      (10 runs, mean: 508ms)
   - Big Document/Several Edits 
      (5 runs, mean: 1938ms)
   - Find/Replace
      (10 runs, mean: 84ms)
   - Code Lens
      (10 runs, mean: 610ms)
<- End benchmarking lsp
```

This PR:

```
-> Start benchmarking lsp
   - Simple Startup/Shutdown
      (10 runs, mean: 503ms)
   - Big Document/Several Edits 
      (5 runs, mean: 1901ms)
   - Find/Replace
      (10 runs, mean: 81ms)
   - Code Lens
      (10 runs, mean: 608ms)
```

> We don't measure memory consumption on the benchmark tests for the LSP, maybe we should find a way?

Yeah, I think it would be very good for us to do this so we don't get any regressions. I will look into it tomorrow.

dsherret:
I looked into a way of tracking memory usage, but it seems like a lot of work so I just opened #14467.

> We should implement the `.fromString()` I suspect in our snapshots as well.

@kitsonk could you clarify what you mean here? I believe I removed our snapshots in this PR.

kitsonk:
> > We should implement the `.fromString()` I suspect in our snapshots as well.
> 
> @kitsonk could you clarify what you mean here? I believe I removed our snapshots in this PR.

Yes, sorry... also I thought `.fromString()` was an instance method, but it was actually a status method of `ts.ScriptSnapshot`, so basically ignore me.

kitsonk (approved):
LGTM

PR #14366, opened 2022-04-22
perf(ext/http): fast path for uncompressed bodies

littledivy:
Brings back the nice juicy numbers seen in https://github.com/denoland/deno/pull/14284

sample.txt is 5MB

`main`:
```
  8 threads and 20 connections
    Latency    38.43ms    4.80ms  83.32ms   91.27%
    Req/Sec    51.87      9.99    60.00     62.87%
  4192 requests in 10.10s, 20.53GB read
Requests/sec:    414.91
Transfer/sec:      2.03GB
```

This PR:
```
  8 threads and 20 connections
    Latency    17.58ms    2.53ms  44.05ms   88.56%
    Req/Sec   114.07     12.06   141.00     67.20%
  9179 requests in 10.10s, 44.87GB read
Requests/sec:    908.86
Transfer/sec:      4.44GB
```

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -730,7 +730,6 @@ async fn op_http_write_resource(
       HttpResponseWriter::BodyUncompressed(body) => {
         if let Err(err) = body.send_data(Bytes::from(buf.to_temp())).await {
           assert!(err.is_closed());
-          // Don't return "broken pipe", that's an implementation detail.
           // Pull up the failure associated with the transport connection instead.
           http_stream.conn.closed().await?;
           // If there was no connection error, drop body_tx.
@@ -799,7 +798,6 @@ async fn op_http_write(
           Ok(_) => break Ok(()),
           Err(err) => {
             assert!(err.is_closed());
-            // Don't return "broken pipe", that's an implementation detail.
             // Pull up the failure associated with the transport connection instead.
             stream.conn.closed().await?;
             // If there was no connection error, drop body_tx.
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -718,29 +718,23 @@ async fn op_http_write_resource(
 
     match &mut *wr {
       HttpResponseWriter::Body(body) => {
-        match body.write_all(&buf[..nread]).await {
-          Ok(_) => (),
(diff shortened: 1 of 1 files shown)


ext/http/lib.rs
```diff
@@ -713,17 +715,30 @@ async fn op_http_write_resource(
...
+          // If there was no connection error, drop body_tx.
+          *wr = HttpResponseWriter::Closed;
+        }
       }
-    }
+      HttpResponseWriter::BodyUncompressed(body) => {
+        if let Err(err) = body.send_data(Bytes::from(buf.to_temp())).await {
+          assert!(err.is_closed());
+          // Don't return "broken pipe", that's an implementation detail.
+          // Pull up the failure associated with the transport connection instead.
```
bartlomieju, line 734: This comment seems copy-pasted from previous match arm and seems irrelevant

ext/http/lib.rs
```diff
@@ -754,30 +769,43 @@ async fn op_http_write(
...
+        }
+      }
+      HttpResponseWriter::BodyUncompressed(body) => {
+        let bytes = Bytes::copy_from_slice(&buf[..]);
+        match body.send_data(bytes).await {
+          Ok(_) => break Ok(()),
+          Err(err) => {
+            assert!(err.is_closed());
+            // Don't return "broken pipe", that's an implementation detail.
+            // Pull up the failure associated with the transport connection instead.
```
bartlomieju, line 803: Ditto

bartlomieju (approved):
LGTM, impressive results. Please verify that `std/http` servers work with this change before landing

lucacasonato (commented):
Is Tokio I/O that much slower? I wonder if we can do this some other way so that we do not have to have so many separate code paths. What about a `HttpResponseWriter::send()` method that encapsulates logic shared across the two send functions?

littledivy:
> What about a HttpResponseWriter::send() method that encapsulates logic shared across the two send functions?

Both logics are not exactly duplicated, one deals with an owned ZeroCopyBuf so it could avoid copy. `op_http_write` copies every time because of ownership outside the scope. Let's think of this in follow-ups.

PR #14367, opened 2022-04-22
refactor(lsp): store all the assets in Rust when initializing

dsherret:
The total size of the assets stored in the TS isolate is 0.7MB. I think it's worth it to store these up front in the Rust code when initializing because it simplifies the code a decent amount.

1. Removes the concept of "cached assets" vs "uncached assets" (reduces the chance of making a mistake as well)
2. Gets rid of a lot of asyncness

### DIFF
--- a/cli/lsp/analysis.rs
+++ b/cli/lsp/analysis.rs
@@ -323,14 +323,13 @@ fn is_preferred(
 
 /// Convert changes returned from a TypeScript quick fix action into edits
 /// for an LSP CodeAction.
-pub async fn ts_changes_to_edit(
+pub fn ts_changes_to_edit(
   changes: &[tsc::FileTextChanges],
   language_server: &language_server::Inner,
 ) -> Result<Option<lsp::WorkspaceEdit>, AnyError> {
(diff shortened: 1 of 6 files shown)


cli/lsp/analysis.rs
```diff
@@ -323,14 +323,13 @@ fn is_preferred(
 
 /// Convert changes returned from a TypeScript quick fix action into edits
 /// for an LSP CodeAction.
-pub async fn ts_changes_to_edit(
+pub fn ts_changes_to_edit(
   changes: &[tsc::FileTextChanges],
   language_server: &language_server::Inner,
```
dsherret, line 328: This is a step closer to being able to remove passing this around too.

bartlomieju (approved):
I must admit that I don't have the full context on why there was this split, this PR significantly simplifies the code and all tests are passing. 

@dsherret if you verify locally that the extension works without the problem with this change then we should land it. LGTM 👍

kitsonk (approved):
LGTM, this makes a lot of sense.

PR #14372, opened 2022-04-23
feat(ext/net): add `NS` DNS record support

### LINKED ISSUE #14355: Deno panicked with Deno.resolveDns trying to resolve NameServer(NS)
Command to reproduce the problem:
await Deno.resolveDns("deno.land","NS")

=========================Message Error===================================

Platform: linux x86_64
Version: 1.19.3
Args: ["deno"]

thread 'main' panicked at 'not yet implemented', /home/runner/work/deno/deno/ext/net/ops.rs:736:12

PatrickChoDev:
Adding `NS` DNS record using `Name` struct. For resolving name server record to delegates a DNS zone to use the given with [RFC 1035](https://datatracker.ietf.org/doc/html/rfc1035#page-12).
Closes #14355

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2953,6 +2953,7 @@ declare namespace Deno {
     | "ANAME"
     | "CNAME"
     | "MX"
+    | "NS"
     | "PTR"
     | "SRV"
     | "TXT";
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2986,7 +2986,7 @@ declare namespace Deno {
 
   export function resolveDns(
     query: string,
-    recordType: "A" | "AAAA" | "ANAME" | "CNAME" | "PTR",
+    recordType: "A" | "AAAA" | "ANAME" | "CNAME" | "NS" | "PTR",
     options?: ResolveDnsOptions,
   ): Promise<string[]>;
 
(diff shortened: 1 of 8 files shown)


bartlomieju:
Looks good, could you add a test case to `test_resolve_dns` integration test?

cli/tests/integration/mod.rs
```diff
@@ -1003,6 +1012,7 @@ async fn test_resolve_dns() {
       .unwrap();
     let err = String::from_utf8_lossy(&output.stderr);
     let out = String::from_utf8_lossy(&output.stdout);
+    println!("{}", err);
```
bartlomieju, line 1015: Remove

bartlomieju (commented):
Looks good, let's see if CI passes

cli/dts/lib.deno.ns.d.ts
```diff
@@ -2953,6 +2953,7 @@ declare namespace Deno {
     | "ANAME"
     | "CNAME"
     | "MX"
+    | "NS"
```
bartlomieju, line 2956: A new overload for `resolveDns` should be added to this file as well

bartlomieju (changes requested):

cli/dts/lib.deno.ns.d.ts
bartlomieju: Actually, a new overload is probably not needed, but "NS" variant should be added to one of existing overloads.

PatrickChoDev: resolved in 6ccb2ce03ca1ffe3d298dfda028124525d7848a4

cli/tests/integration/mod.rs
PatrickChoDev: removed in 6ccb2ce03ca1ffe3d298dfda028124525d7848a4

bartlomieju (approved):
LGTM, nice work @PatrickChoDev!

PR #14378, opened 2022-04-23
refactor: Remove PrettyJsError and js_error_create_fn

### LINKED ISSUE #14377: Worker error events missing location info
```ts
const worker = new Worker(
  new URL("data:,throw new Error('foo')", import.meta.url).href,
  { type: "module" },
);
worker.addEventListener("error", (e) => {
  console.log({
    "message": e.message,
    "filename": e.filename,
    "lineno": e.lineno,
    "colno": e.colno,
  });
});
```
Expected
```
{ filename: "data:,throw new Error('foo')", lineno: 1, colno: 7 }
```
Actual
```
{ filename: "", lineno: 0, colno: 0 }
```
This is because the [code](https://github.com/denoland/deno/blob/2eb8c3b82fd54027c35c87ccc94797e34e2e95fd/runtime/web_worker.rs#L97-L102) that gets these properties depends on downcasting to `JsError` which fails because of wrappers such as `ErrWithV8Handle` and `PrettyJsError`. I think we should refactor these.

nayeemrmn:
Blocked on #14394.
Fixes #14377. Also adds pretty-printed errors and source mapping for `deno compile`d executables.


cli/main.rs
```diff
@@ -1442,10 +1442,14 @@ fn unwrap_or_exit<T>(result: Result<T, AnyError>) -> T {
   match result {
     Ok(value) => value,
     Err(error) => {
+      let error_string = match error.downcast_ref::<JsError>() {
+        Some(e) => format_js_error(e),
+        None => format!("{:?}", error),
+      };
```
bartlomieju, line 1448: Why is it needed?

cli/tests/integration/compile_tests.rs
```diff
@@ -103,11 +103,15 @@ fn standalone_error() {
     .wait_with_output()
     .unwrap();
   assert!(!output.status.success());
+  println!("{:#?}", &output);
```
bartlomieju, line 106: Remove

core/modules.rs
```diff
@@ -711,6 +709,12 @@ enum SymbolicModule {
   Mod(ModuleId),
 }
 
+#[derive(Debug)]
+pub(crate) enum ModuleError {
+  Exception(v8::Global<v8::Value>),
+  Other(Error),
+}
```
bartlomieju, line 716: I quite like this approach

bartlomieju (commented):
There's a lot going on in this PR @nayeemrmn, would it be possible to split it into two separate PRs? One for changes to `core/` and the other one that updates workers and compile subcommand

nayeemrmn:
> There's a lot going on in this PR @nayeemrmn, would it be possible to split it into two separate PRs? One for changes to `core/` and the other one that updates workers and compile subcommand

I can split it it into one for removing `ErrWithV8Handle` (core-only refactor) and one for removing `PrettyJsError` along with `create_js_error_fn` from core's API.

cli/main.rs
nayeemrmn: The only places errors are pretty printed are `cli/main.rs`, `cli/tools/test.rs` and `runtime/web_worker.rs`. This is where we apply `format_js_error()`. In fact, the `WorkerOptions::format_js_error_fn` in runtime is only for the third one.

bartlomieju:
> > There's a lot going on in this PR @nayeemrmn, would it be possible to split it into two separate PRs? One for changes to `core/` and the other one that updates workers and compile subcommand
> 
> I can split it it into one for removing `ErrWithV8Handle` (core-only refactor) and one for removing `PrettyJsError` along with `create_js_error_fn` from core's API.

That sounds good 👍

bartlomieju (approved):
LGTM, great refactor @nayeemrmn

PR #14384, opened 2022-04-24
feat(ext/console): Add string abbreviation size option for "Deno.inspect"

bcheidemann:
Implements an option on `Deno.inspect` to set the maximum string size before it is truncated with an ellipsis.

This is required to fix [this issue in denoland/deno_std](https://github.com/denoland/deno_std/issues/2140).

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2491,6 +2491,8 @@ declare namespace Deno {
     getters?: boolean;
     /** Show an object's non-enumerable properties. Defaults to false. */
     showHidden?: boolean;
+    /** The maximum length of a string before it is truncated with an ellipsis */
+    strAbbreviateSize?: number;
   }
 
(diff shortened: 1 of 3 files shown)


bcheidemann:
@bartlomieju @kt3k could you advise on where to add a test for this?

bartlomieju:
> @bartlomieju @kt3k could you advise on where to add a test for this?

That would be in `cli/tests/unit/console_test.ts`

bcheidemann:
This property currently doesn't apply to strings because strings are not currently truncated by `Deno.inspect`. Should this property also apply to `Deno.inspect(someLongString)`?

kt3k (approved):
LGTM

kt3k:
> This property currently doesn't apply to strings because strings are not currently truncated by Deno.inspect. Should this property also apply to Deno.inspect(someLongString)?

I don't think we need it at this point.

bartlomieju (approved):
LGTM

PR #14387, opened 2022-04-24
test(ext/console): Compact empty iterables when calling Deno.inspect with compact false

bcheidemann:
When `Deno.inspect` is called with the `compact` option set to `false`, empty iterables will still be compacted.

**Before**
```ts
Deno.inspect([], { compact: false });
// [
// 
// ]

Deno.inspect([], { compact: false, trailingComma: true });
// [
//   ,
// ]
```

**Now**
```ts
Deno.inspect([], { compact: false });
// []

Deno.inspect([], { compact: false, trailingComma: true });
// []
```

This is required to fix [this issue in denoland/deno_std](https://github.com/denoland/deno_std/issues/2144).


kt3k:
@bcheidemann 
Thanks for the PR. I see this fixes the issue, but ideal fix would be like the below in my view:

```js
Deno.inspect([], { compact: false, trailingComma: true });
// [
//   
// ]
```

(This has less effect to the existing code. Also this will produce better diff when comparing, for example, `[]` with `[1]` in snapshot testing)

What do you think?

bcheidemann:
> @bcheidemann Thanks for the PR. I see this fixes the issue, but ideal fix would be like the below in my view:
> 
> ```js
> Deno.inspect([], { compact: false, trailingComma: true });
> // [
> //   
> // ]
> ```
> 
> (This has less effect to the existing code. Also this will produce better diff when comparing, for example, `[]` with `[1]` in snapshot testing)
> 
> What do you think?

I can see the benefit of this from both perspectives - better snapshot diffs and minimising impact on existing code.

However, I would suggest that the following would make more sense.

```js
Deno.inspect([], { compact: false, trailingComma: true });
// [
// ]
```

This way the diff would be:

```
Expected + / Actual -

   [
-    3
   ]
```

Instead of:

```
Expected + / Actual -

   [
+
-    3
   ]
```

Let me know which you would prefer @kt3k - happy to implement either 🙂

ext/console/02_console.js
```diff
@@ -465,12 +465,18 @@
...
+      level,
+    );
+    const closingIndentation = `${
+      inspectOptions.trailingComma ? "," : ""
+    }\n${closingDelimIndentation}`;
 
     let iContent;
-    if (options.group && entries.length > MIN_GROUP_LENGTH) {
+    if (entries.length === 0 && !inspectOptions.compact) {
+      iContent = `\n${closingDelimIndentation}`;
```
kt3k, line 478: 👍

kt3k (approved):
LGTM!

bartlomieju (approved):
LGTM

PR #14388, opened 2022-04-24
fix(ext): disable bytes mode on streams that don't handle it correctly

### LINKED ISSUE #14219: ReadableStream in bring-your-own-buffer mode doesn't handle EOF
`ReadableStreamBYOBReader.read()` should return a Promise that resolves to `{done: true, value: undefined}` when the stream hits EOF, but instead the returned Promise never resolves.

```ts
const req = await fetch("http://example.com");
const reader = req.body!.getReader({ mode: "byob" });

let buf = new Uint8Array(4096);
const result = await reader.read(buf);
console.log(result);

buf = new Uint8Array(result.value!.buffer);
const result2 = await reader.read(buf);
console.log(result2);
```

```
% deno run --allow-net byob_fail.ts
Check file:///Users/macil/Coding/streaming-zip/byob_fail.ts
{
  value: Uint8Array(1256) [
     60,  33, 100, 111,  99, 116, 121, 112, 101,  32, 104, 116, 109, 108,
     62,  10,  60, 104, 116, 109, 108,  62,  10,  60, 104, 101,  97, 100,
     62,  10,  32,  32,  32,  32,  60, 116, 105, 116, 108, 101,  62,  69,
    120,  97, 109, 112, 108, 101,  32,  68, 111, 109,  97, 105, 110,  60,
     47, 116, 105, 116, 108, 101,  62,  10,  10,  32,  32,  32,  32,  60,
    109, 101, 116,  97,  32,  99, 104,  97, 114, 115, 101, 116,  61,  34,
    117, 116, 102,  45,  56,  34,  32,  47,  62,  10,  32,  32,  32,  32,
     60, 109,
    ... 1156 more items
  ],
  done: false
}
error: Module evaluation is still pending but there are no pending ops or dynamic imports. This situation is often caused by unresolved promises.
```

Every read() call works except for the final one that's supposed to resolve to `done: true`. (If you adjust the buffer size to be smaller here so that the first read doesn't reach EOF, then the second read() works fine.)

Macil:
Fixes #14219

Several ReadableStream instances were marked `type: "bytes"` which means that they're marked as supporting bring-your-own-buffer-mode, but then they did not actually do anything to support byob mode. They failed to handle closing correctly in byob mode. This PR makes them into regular non-bytes-mode ReadableStreams, which will act identically to before except in the case that `.getReader({ mode: "byob" })` is called on them,  which will now throw an error. This behavior is consistent with browsers like Chrome. Previously, that would return a ReadableStreamBYOBReader that didn't properly support bring-your-own-buffer mode and didn't handle end-of-file at all.

This solution (removing byob support) was picked instead of fixing the byob support because of https://github.com/denoland/deno/issues/14219#issuecomment-1107874094.


ext/web/09_file.js
```diff
@@ -320,8 +320,6 @@
       webidl.assertBranded(this, BlobPrototype);
       const partIterator = toIterator(this[_parts]);
       const stream = new ReadableStream({
-        type: "bytes",
-        /** @param {ReadableByteStreamController} controller */
```
Macil, line 324: (comment removed because removing `type: "bytes"` changes the type of `controller` to be ReadableStreamDefaultController instead which is inferred fine, and not specified in other similar bits of code)

bartlomieju:
@lucacasonato please take a look, is it something we want to land?

lucacasonato (commented):
@crowlKats PTAL

crowlKats:
This would be a large breaking change, so respectfully closing

Macil:
The current behavior is unusably broken and also inconsistent with browsers like Chrome. It seems unlikely anyone depends on these streams supporting bytes-mode because they're broken there.

Would a PR that instead makes these ReadableStreams properly support bytes-mode be accepted?

PR #14389, opened 2022-04-25
fix(ext/http): truncate read bytes when streaming bodies

littledivy:


lucacasonato:
Test?

ry (commented):
does this fix a bug? should it have a test?

PR #14393, opened 2022-04-25
refactor(ops/process): add `StdioOrRid` enum

dsherret:
I was confused by this code while working on piping process output to test output so I refactored it. Makes #14395 a lot easier too.


bartlomieju (approved):
LGTM

PR #14395, opened 2022-04-25
fix(test): capture inherited stdout and stderr for subprocesses in test output

dsherret:
Waiting on #14393 first


bartlomieju:
Looks like it's almost working, I spotted that some outputs are not enclosed between the markers:
Screenshot 2022-04-26 at 14 27 52

dsherret:
@bartlomieju weird. It works for me
image

Also, it's strange that the caret is so high up for you. I will try on linux.

bartlomieju:
It works locally for me too:
Screenshot 2022-04-26 at 15 47 32

One more test that looks broken:
Screenshot 2022-04-26 at 15 48 01

bartlomieju (approved):
LGTM, unless you want to address the issue with workers in this PR as well.

dsherret:
Yeah, I'm going to do it in a separate PR. I haven't found a nice way of doing it yet.

PR #14396, opened 2022-04-26
experiment(core): Move the promise ring to Rust (REBASED)

andreubotella:
Towards #13943.

---------------------------------------------------

This is a rebased version of #13906 which fully works with `deno_ops`. It also fixes the following issues with the original PR:

  - The original PR removed `promiseIdSymbol` and made it so a promise's ID isn't accessible to JS. This, however, makes `Deno.core.{un,}refOp` useless, since there is no way for the JS code calling them to access a promise's ID. In this PR, calling an async op returns a `[promise, id]` tuple, and `Deno.core.opAsync` sets that ID on the promise object with `promiseIdSymbol`, so the promise ID is once again accessible to JS code with access to that symbol.

  - The original PR also removed functionality for op call tracing from `01_core.js` but didn't implement it in Rust. This PR restores that functionality to the JS side for the time being.

  - This PR also fixes the issues in `PromiseRing::has` that I pointed out in https://github.com/denoland/deno/pull/13906#discussion_r837142758.

  - #13861 removed the `sync_cb` callback, and moving the promise ring to Rust removes the `recv_cb` callback. With both callbacks gone, the `JsRuntime::init_cbs` method is now useless and has been removed.

  - Rather than making microtasks explicit all the time, this PR makes them explicit only inside of `JsRuntime::resolve_async_ops`. After the promises have been resolved, a microtask checkpoint is performed inside a `v8::TryCatch` so unhandled exceptions are correctly handled in the event loop.

cc @AaronO


bartlomieju:
@andreubotella is there visible performance difference?

core/bindings.rs
```diff
@@ -531,7 +531,7 @@ fn ref_op<'s>(
   let mut state = state_rc.borrow_mut();
 
   let promise_id = match v8::Local::<v8::Integer>::try_from(args.get(0))
-    .map(|l| l.value() as PromiseId)
+    .map(|l| l.uint32_value(scope).unwrap_or_default() as PromiseId)
```
bartlomieju, line 534: This is default to 0 if no arg is passed, we should probably throw an error explicitly instead of defaulting to id 0

core/promise_ring.rs
```diff
@@ -0,0 +1,82 @@
+use crate::PromiseId;
```
bartlomieju, line 1: Please add copyright header

andreubotella:
> Does that mean that `Deno.core.(un)refOp()` doesn't work anymore?

With @AaronO's PR, the promise ID's weren't exposed to JS at all, so there was no way to use `Deno.core.{,un}refOp()`, since they take the promise ID. This PR instead returns both the created promise and the new promise ID to the JS side, so it can still be exposed with the `promiseIdSymbol` symbol.

littledivy:
This PR is slower than main, I feel we should optimize this to be on par with the JS promise ring before landing.

`main`:
```
cargo bench -p deno_bench_util --bench op_baseline

test bench_op_async   ... bench:     258,880 ns/iter (+/- 4,305)
```

This patch:
```
cargo bench -p deno_bench_util --bench op_baseline

test bench_op_async   ... bench:     322,266 ns/iter (+/- 5,645)
```

littledivy:
We should be able to avoid `promiseIdSymbol` in JS altogether with embedder internal fields. See https://github.com/denoland/rusty_v8/pull/967

andreubotella:
> We should be able to avoid `promiseIdSymbol` in JS altogether with embedder internal fields. See [denoland/rusty_v8#967](https://github.com/denoland/rusty_v8/pull/967)

Ideally the op's `v8_func` would return a promise with the embedder field set. However, currently we handle result unwrapping and op call tracing by creating new promises with `Promise.prototype.then` and similar in JS. In order to use this, we'd need a binding to set the returned promise's embedder field, which doesn't sound like it'd be much of an improvement, performance-wise.

It doesn't seem like op call tracing would be that hard to move to Rust. `unwrapOpResult`, though, would be a problem. I'll experiment with exporting `unwrapOpResult` and calling it from Rust, though.

ops/lib.rs
```diff
@@ -157,14 +157,18 @@ fn codegen_v8_async(
...
-    quote! {}
-  };
-  let rust_i0 = if uses_opstate { 1 } else { 0 };
+  let MacroArgs { is_v8, .. } = margs;
+  let special_args = f
+    .sig
+    .inputs
+    .iter()
+    .map_while(|a| {
+      (if is_v8 { scope_arg(a) } else { None }).or_else(|| opstate_arg(a))
```
aapoalas, line 166: `if is_v8 { scope_arg(a) } else { opstate_arg(a) }` possibly?

core/bindings.rs
andreubotella: With the move to `#[op(v8)]` non-number arguments will throw, but negative numbers might be truncated to 0. That's probably not worth fixing though.

ops/lib.rs
andreubotella: That's not something I changed – it's there on `master`.

core/promise_ring.rs
andreubotella: Done

andreubotella:
This PR gets us some of the way towards supporting async ops in realms, but it would still need additional changes, some of which are already implemented in the alternative PR #14734. Considering that, and the performance loss, I think it's better to close this PR and focus on the other one.

PR #14410, opened 2022-04-26
fix(test): capture worker stdout and stderr in test output

dsherret:
Also improves the runtime API for this scenario.

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -22,7 +22,6 @@ use crate::graph_util::graph_valid;
 use crate::located_script_name;
 use crate::lockfile;
 use crate::ops;
-use crate::ops::testing::create_stdout_stderr_pipes;
 use crate::proc_state::ProcState;
 use crate::resolver::ImportMapResolver;
 use crate::resolver::JsxResolver;
@@ -41,6 +40,8 @@ use deno_core::serde_json::json;
 use deno_core::url::Url;
 use deno_core::ModuleSpecifier;
 use deno_graph::ModuleKind;
+use deno_runtime::ops::io::Stdio;
+use deno_runtime::ops::io::StdioPipe;
 use deno_runtime::permissions::Permissions;
 use deno_runtime::tokio_util::run_basic;
 use log::Level;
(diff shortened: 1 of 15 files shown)


cli/tools/test.rs
```diff
@@ -588,17 +590,17 @@ async fn test_specifier(
...
-      stdout_writer,
-      stderr_writer,
-    )],
+    vec![ops::testing::init(channel.clone())],
+    Stdio {
+      stdin: StdioPipe::Inherit,
+      stdout: StdioPipe::File(stdout),
+      stderr: StdioPipe::File(stderr),
+    },
   );
```
bartlomieju, line 604: I think it's high time we introduce `MainWorkerOptions` struct, 5 arguments become unwieldy without field names; I can do that in a follow up

cli/tools/test.rs
```diff
@@ -1452,3 +1454,56 @@ pub async fn run_tests_with_watch(
 
   Ok(())
 }
+
+/// Creates the stdout and stderr pipes and returns the writers for stdout and stderr.
+pub fn create_stdout_stderr_pipes(
+  sender: UnboundedSender<TestEvent>,
+) -> (std::fs::File, std::fs::File) {
```
bartlomieju, line 1461: I need to finish https://github.com/denoland/deno/pull/14299 which would mean that this function will be required in `cli/tools/bench.rs`. I'm thinking it should be moved to `runtime/ops/io.rs` at that point, WDYT?

bartlomieju (approved):
LGTM, great cleanup

cli/tools/test.rs
dsherret: I think it could still remain in the CLI, but moved to a common place that `cli/tools/bench.rs` also uses? `runtime/ops/bench.rs` could mostly be reverted other than `op_print`.

bartlomieju: > I think it could still remain in the CLI, but moved to a common place that `cli/tools/bench.rs` also uses? `runtime/ops/bench.rs` could mostly be reverted other than `op_print`.

Currently there's no such module. What do you mean by reverted?

dsherret: Oh, I meant cli/ops/bench.rs then.

PR #14412, opened 2022-04-27
experiment(serde_v8): bytes::Bytes support

AaronO:
Allows ~zero-cost conversions of `serde_v8` buffer-like types into `bytes::Bytes`, improving efficient interoperability with the hyper/tokio ecosystem

(It does require a single heap-alloc: `V8Slice -> Rc<V8Slice>` but avoids copying the original bytes)

## Tasks

- [x] Bounds check & support len/ptr in clone "constructor" (used when Bytes reslices)


bartlomieju (approved):
LGTM

PR #14417, opened 2022-04-27
fix(ext/http): explicitly flush compressed streaming bodies

littledivy:


cli/tests/unit/http_test.ts
```diff
@@ -1355,6 +1355,66 @@ Deno.test({
   },
 });
 
+Deno.test({
+  name: "http server compresses streaming body - check body",
```
lucacasonato, line 1359:
```suggestion
  name: "http server compresses streamed body (direct resource fast path) - check body",
```

And move down to the other "compresses streaming body" tests.

PR #14419, opened 2022-04-27
fix: `deno task` should forward double hyphens

dsherret:
When someone runs:

```
deno task node:test -- test-assert.js
```

It should forward everything after the task name like so:

```
Task node:test deno test --unstable --allow-all node/_tools/test.ts "--" "test-assert.js"
```

Currently it does:

```
Task node:test deno test --unstable --allow-all node/_tools/test.ts "test-assert.js"
```

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -512,14 +512,10 @@ To evaluate code in the shell:
 ";
 
 /// Main entry point for parsing deno's command line flags.
-pub fn flags_from_vec<I, T>(args: I) -> clap::Result<Flags>
-where
-  I: IntoIterator<Item = T>,
-  T: Into<std::ffi::OsString> + Clone,
-{
+pub fn flags_from_vec(args: Vec<String>) -> clap::Result<Flags> {
   let version = crate::version::deno();
   let app = clap_root(&version);
-  let matches = app.clone().try_get_matches_from(args)?;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, thanks for a quick fix!

jsejcksn:
@dsherret With this change, is it no longer possible to utilize the first element of `Deno.args` for any other value than `--` in a script that is run as a task command?

If that is the case, can you provide some clarity on why this change was requested and implemented?

kt3k:
@jsejcksn
> With this change, is it no longer possible to utilize the first element of Deno.args for any other value than "--" in a script that is run as a task command?

Right, we can't pass the 1st arg to the internal script, and that looks like a bug to me.

I think we shouldn't parse anything in `<task_args>` and just pass everything to the internal command (like `yarn` does)

> deno task log_args -1

I think this should invoke `deno run log_args.ts -1` instead of throwing an error

What do you think? @dsherret @bartlomieju

jsejcksn:
In additional testing, this change also appears to break the [`std/flags`](https://deno.land/std@0.137.0/flags) parser module. 😕

dsherret:
It's correct that it's forwarding `--` now, but it should allow running `deno run log_args.ts -1`

Issue #14404, opened 2022-04-26
Help wraps on 80 char terminal


tdillon:
This line is a few characters too long to be rendered on an 80 character wide terminal window.

https://github.com/denoland/deno/blob/f07f246ae8a158e33e81aa4ccf225cd536795f50/cli/flags.rs#L485

![image](https://user-images.githubusercontent.com/12434812/165329182-01db9c1b-5663-4652-a809-12992d35a523.png)

PR #14422, opened 2022-04-28
fix(cli): Wrap long line of the env variables help

johnmanjiro13:
fix https://github.com/denoland/deno/issues/14404

image


bartlomieju (approved):
Thanks!

PR #14429, opened 2022-04-28
fix(test/bench): Accept file protocol module specifier

### LINKED ISSUE #14375: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', cli\fs_util.rs:273:52
```
 deno test "file://C:/Documents/gitee/leetcode-test/deps.ts"




Platform: windows x86_64
Version: 1.21.0
Args: ["C:\\Users\\Administrator\\.deno\\bin\\deno.exe", "test", "file://C:/Documents/gitee/leetcode-test/deps.ts"]

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', cli\fs_util.rs:273:52
```

```
deno bench "file:///C:/Documents/gitee/leetcode-test/deps.ts"

Platform: windows x86_64
Version: 1.21.0
Args: ["C:\\Users\\Administrator\\.deno\\bin\\deno.exe", "bench", "file:///C:/Documents/gitee/leetcode-test/deps.ts"]

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', cli\fs_util.rs:273:52

```

```ts
export {
    assert,
    assertAlmostEquals,
    assertEquals,
    assertStrictEquals,
    equal,
} from "https://deno.land/std@0.136.0/testing/asserts.ts";
export { default as random } from "https://cdn.skypack.dev/lodash@4.17.21/random?dts";

export { default as uniqBy } from "https://cdn.skypack.dev/lodash@4.17.21/uniqBy?dts";

```

sirgallifrey:
This PR fixes problem reported on #14375 where deno cli `test` and `bench` panic if given a module specifier as file URL

On MacOs it would not panic, but it fails with error saying it's not possible to read the file.

I was able to reproduce the issue with integration tests, and they are now passing with the fix.

A note about the solution: It will work only for files as File protocol URL, but not for directories. A directory will return an error saying it's not possible to read the file. I decided not implement the directory for 2 reasons:
1 - this is the simpler approach, and it's better to increment if you folks want than removing it.
2 - this is probably how http:// and https:// will work for `test` and `bench`. I didn't tested this, but by reading the code, they will not be able to handle a directory given by URL

If you would like, I can change the approach and make so it will check the URL (only file protocol), if it's a directory then we can recursively get all files, otherwise we don't change it.

`cargo test` failed with 4 errors, but they are unrelated to my change and I believe that is some problem with my setup, since they seem related to symlinks
<details>
  <summary>See failures here</summary>
failures:

        ./cli/tests/unit/stat_test.ts
        statSyncSuccess
        lstatSyncSuccess
        statSuccess
        lstatSuccess

test result: FAILED. 1003 passed (6 steps); 4 failed; 59 ignored; 0 measured; 0 filtered out (1m42s)
</details>

code was formatted with `./tools/format.js`
`./tools/lint.js` passed successfully with no errors
closes #14375


sirgallifrey:
side note: so far I tested only on my windows machine, tomorrow I'll run the tests on macOs

dsherret (approved):
Looks great. Thanks!

Issue #14344, opened 2022-04-21
`AbortSignal.reason` not available in `lib.dom.d.ts`


jaydenseric:
Something has changed in deno [v1.21.0](https://github.com/denoland/deno/releases/tag/v1.21.0), as upgrading from v1.20.6 has caused TS errors to fail tests that used to run fine:

```sh
deno test \
  --unstable \
  --allow-env \
  --allow-net \
  --allow-run \
  --allow-read \
  --allow-write \
  --import-map=importMap.json
```

```
Check file://[redacted]/ruck/Effect.test.mjs
Check file://[redacted]/ruck/HeadManager.test.mjs
Check file://[redacted]/ruck/HeadManagerContext.test.mjs
Check file://[redacted]/ruck/Html.test.mjs
Check file://[redacted]/ruck/LinkCss.test.mjs
Check file://[redacted]/ruck/NavigateContext.test.mjs
Check file://[redacted]/ruck/RouteContext.test.mjs
Check file://[redacted]/ruck/TransferContext.test.mjs
Check file://[redacted]/ruck/assertImportMap.test.mjs
Check file://[redacted]/ruck/createPseudoNode.test.mjs
Check file://[redacted]/ruck/documentHasStyleSheet.test.mjs
Check file://[redacted]/ruck/hydrate.test.mjs
Check file://[redacted]/ruck/publicFileResponse.test.mjs
Check file://[redacted]/ruck/readImportMapFile.test.mjs
Check file://[redacted]/ruck/routePlanForContentWithCss.test.mjs
Check file://[redacted]/ruck/scrollToHash.test.mjs
Check file://[redacted]/ruck/serve.test.mjs
Check file://[redacted]/ruck/useOnClickRouteLink.test.mjs
error: TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
    return Promise.reject(createAbortError(signal.reason));
                                                  ~~~~~~
    at https://deno.land/std@0.136.0/async/abortable.ts:27:51

TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
  const abort = () => waiter.reject(createAbortError(signal.reason));
                                                            ~~~~~~
    at https://deno.land/std@0.136.0/async/abortable.ts:30:61

TS2339 [ERROR]: Property 'reason' does not exist on type 'AbortSignal'.
    throw createAbortError(signal.reason);
                                  ~~~~~~
[...]

Issue #12558, opened 2021-10-26
`URLPattern` not available in `lib.dom.d.ts`


lucacasonato:
We should add it now, even if TSC has not quite caught up yet.

lucacasonato:
I think we should do this even though it may sound like a slippery slope. Using `URLPattern` with `lib.dom.d.ts` is currently really painful.

bartlomieju:
I agree with @lucacasonato

kitsonk:
We should do a `lib.dom.extras.d.ts` then versus patching the file, like we do for `lib.dom.asynciterable.d.ts`. We could in theory warn about including it, like we do with unstable.

sant123:
@kitsonk so in a future could exist this?

``` json
{
  "compilerOptions": {
    "lib": ["dom", "dom.asynciterable", "deno.ns", "dom.extra"],
    "jsxFactory": "h"
  }
}

```

kitsonk:
Yes.

PR #14430, opened 2022-04-29
fix(cli): add dom.extras lib

kitsonk:
Closes: #12558
Fixes: #14344

This adds the `lib.dom.extras.d.ts` file and modifies the `lib.dom.d.ts` file to include it.  This solves the problem where users use the `lib: ["dom", "deno.ns"]` configuration when compiling their isomorphic TypeScript and the distributed version of the DOM library does not include every web standard API that Deno supports, and therefore causes confusing errors to the user.

This means we do make a one line edit to the distributed `lib.dom.d.ts` to include `lib.dom.extras.d.ts`.  This means that users don't have to explicitly include `"dom.extras"` when using the library, so for most users previous errors will just "go away".

This does establish a pattern now where we can add items Deno supports in advance of the TypeScript distribution supporting it while allowing more isomorphic code to type check safely.

### DIFF
--- a/cli/tests/testdata/lib_dom_extras.ts
+++ b/cli/tests/testdata/lib_dom_extras.ts
@@ -1,0 +1,17 @@
+const { diagnostics, files } = await Deno.emit("/main.ts", {
+  compilerOptions: {
+    target: "esnext",
+    lib: ["esnext", "dom"],
+  },
+  sources: {
+    "/main.ts": `const as = new AbortSignal();
+    console.log(as.reason);
+    
+    const up = new URLPattern("https://example.com/books/:id");
+    console.log(up);
+    `,
+  },
+});
+
+console.log(diagnostics);
+console.log(Object.keys(files).sort());
--- a/cli/dts/lib.dom.d.ts
+++ b/cli/dts/lib.dom.d.ts
@@ -16,7 +16,7 @@ and limitations under the License.
 
 
 /// <reference no-default-lib="true"/>
-
+/// <reference lib="dom.extras" />
 
 /////////////////////////////
 /// Window APIs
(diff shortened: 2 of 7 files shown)


cli/tests/testdata/lib_dom_extras.ts
```diff
@@ -0,0 +1,17 @@
...
+    console.log(as.reason);
+    
+    const up = new URLPattern("https://example.com/books/:id");
+    console.log(up);
+    `,
+  },
+});
+
+console.log(diagnostics);
+console.log(Object.keys(files).sort());
```
dsherret, line 17: Is this test file used? It would be good to add a test for this (Maybe I’m missing something?)

kitsonk: sorry, forgot to add it as an integration test... 🤦  tested it manually though

dsherret (approved):
Nice! This will be good to just work now. I think we should consider this same solution for any similar problems (if they exist)

jaydenseric:
So if you are publishing a `.mjs` module that has TypeScript JSDoc types, that is intended to work with type safety both in Node.js and Deno TypeScript projects, that uses DOM globals, how are you supposed to support that?

Do you use triple slash references in the published module? If so, what are the right ones to use now? Do you only add this?

```js
/// <reference lib="dom" />
```

For deno, do you also need to add this:

```js
/// <reference lib="deno.ns" />
```

If you do, that will break Node.js TypeScript:

Screen Shot 2022-05-09 at 12 05 13 pm

dsherret:
@jaydenseric would you be able to open a new issue? It’s a better medium than old PRs that are closed.

PR #14434, opened 2022-04-29
fix(task): allow hyphen values after task name

dsherret:
See https://github.com/denoland/deno/pull/14419#issuecomment-1113127332

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1434,6 +1434,7 @@ fn task_subcommand<'a>() -> Command<'a> {
       Arg::new("task_args")
         .multiple_values(true)
         .multiple_occurrences(true)
+        .allow_hyphen_values(true)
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

jsejcksn:
@dsherret Referring to [the comment](https://github.com/denoland/deno/pull/14419#issuecomment-1113127332) you linked:

With this change, I expect the commands from 2 and 5 to produce the same result (the result from 2), and for the commands from 3 and 7 to produce the same result (the result from 3). Is that the case?

dsherret:
@jsejcksn it does for 2, 3, and 7, but I missed 5. Opened #14436. Thanks!

jsejcksn (commented):
Thanks!

PR #14435, opened 2022-04-29
fix(test): actually capture stdout and stderr in web workers

dsherret:
Fixes flaky captured_output on main: https://github.com/denoland/deno/runs/6229873309?check_suite_focus=true

While investigating this, I discovered that stdout and stderr weren't actually being captured properly in web workers... the previous test passed by coincidence.

Follow up to #14410


runtime/ops/io.rs
```diff
@@ -419,6 +423,24 @@ impl Resource for StdFileResource {
...
+pub fn op_print(
+  state: &mut OpState,
+  msg: String,
+  is_err: bool,
+) -> Result<(), AnyError> {
+  let rid = if is_err { 2 } else { 1 };
+  StdFileResource::with(state, rid, move |r| match r {
+    Ok(std_file) => {
+      std_file.write_all(msg.as_bytes())?;
+      std_file.flush().unwrap();
```
dsherret, line 437: Copied from https://github.com/denoland/deno/blob/12d28dffc6c8d08e6620751e55d8a382c66443a2/core/ops_builtin.rs#L103-L112

cli/ops/testing.rs
```diff
@@ -90,15 +85,3 @@ fn op_dispatch_test_event(
...
-pub fn op_print(
-  state: &mut OpState,
-  msg: String,
-  _is_err: bool,
-) -> Result<(), AnyError> {
-  let sender = state.borrow::<UnboundedSender<TestEvent>>().clone();
-  let msg = TestOutput::String(msg);
-  sender.send(TestEvent::Output(msg)).ok();
-  Ok(())
-}
```
dsherret, line 104: CI failed because this change to only use the pipes exposed that what we're doing here is racy. I'm not exactly sure yet how we can fix this.

dsherret: It's unfortunate, but I'm going to have to implement some manual flushing for these :/

Converted to draft

cli/lsp/testing/execution.rs
```diff
@@ -754,19 +755,14 @@ impl test::TestReporter for LspTestReporter {
...
         .get(origin)
         .and_then(|v| v.last().map(|td| td.into()))
     });
-    let value = match output {
-      test::TestOutput::String(value) => value.replace('\n', "\r\n"),
-      test::TestOutput::Bytes(bytes) => {
-        String::from_utf8_lossy(bytes).replace('\n', "\r\n")
-      }
-    };
+    let value = String::from_utf8_lossy(output).replace('\n', "\r\n");
```
dsherret, line 765: This is actually somewhat of a bug because the data is sent over buffered and so this might be between a multi-byte character... I'm not sure how to fix this.

cli/ops/testing.rs
dsherret: Ok, this looks good now and seems to have been solved by doing a flush of the stdout and stderr pipes before sending certain events.

runtime/ops/io.rs
```diff
@@ -111,6 +111,10 @@ pub fn init_stdio(stdio: Stdio) -> Extension {
   let stdio = Rc::new(RefCell::new(Some(stdio)));
 
   Extension::builder()
+    .middleware(|op| match op.name {
+      "op_print" => op_print::decl(),
+      _ => op,
+    })
```
bartlomieju, line 117: This seems fishy - effectively we're always overriding `op_print` provided by `deno_core`. I guess it makes sense though, other consumers/embedders besides CLI might not want to do that (due to missing stdio pipes).

bartlomieju (approved):
LGTM

runtime/ops/io.rs
dsherret: Yeah, I thought so too, but it would require pushing all this code down to deno_core.

PR #14438, opened 2022-04-30
experiment(ops): sync Rc<RefCell<OpState>>

AaronO:


littledivy (approved):
LGTM. Two borrows happening now when arg is `&mut OpState`, we can reuse the borrow_mut after `Self::call` - not very important tho.

bartlomieju (approved):

PR #14439, opened 2022-04-30
experiment(ops): faster OpCtx ptr via f64

AaronO:
Marginally faster than using v8::External due to embedder pointer indirection

### DIFF
--- a/core/bindings.rs
+++ b/core/bindings.rs
@@ -297,7 +297,8 @@ pub fn set_func_raw(
   external_data: *const c_void,
 ) {
   let key = v8::String::new(scope, name).unwrap();
-  let external = v8::External::new(scope, external_data as *mut c_void);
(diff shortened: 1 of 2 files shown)


core/bindings.rs
```diff
@@ -297,7 +297,8 @@ pub fn set_func_raw(
   external_data: *const c_void,
 ) {
   let key = v8::String::new(scope, name).unwrap();
-  let external = v8::External::new(scope, external_data as *mut c_void);
+  let f64ptr = unsafe { std::mem::transmute(external_data) };
```
aapoalas, line 300: Explicit transmute types.

littledivy (commented):
`f64` to pointer cast is sus (for an everyday Rust core reviewer). Please add it to the SAFETY comment above.

bartlomieju (commented):
Thank you for the PR, we've now doing translation to Fast API and I believe this PR is no longer needed.

PR #14450, opened 2022-05-01
perf(serde_v8): fast path for large strings

AaronO:
```
Before:
test de_sob_str_6b        ... bench:          48 ns/iter (+/- 2)
test de_sob_str_64b       ... bench:         105 ns/iter (+/- 4)
test de_sob_str_512b      ... bench:         327 ns/iter (+/- 19)
test de_sob_str_1kb       ... bench:         460 ns/iter (+/- 5)
test de_sob_str_4kb       ... bench:       1,571 ns/iter (+/- 27)

After:
test de_sob_str_6b        ... bench:          49 ns/iter (+/- 3)
test de_sob_str_64b       ... bench:          66 ns/iter (+/- 9)
test de_sob_str_512b      ... bench:         119 ns/iter (+/- 3)
test de_sob_str_1kb       ... bench:          84 ns/iter (+/- 3)
test de_sob_str_4kb       ... bench:         181 ns/iter (+/- 8)
```


serde_v8/de.rs
```diff
@@ -661,3 +661,57 @@ fn bigint_to_f64(b: v8::Local<v8::BigInt>) -> f64 {
...
+  to_utf8_slow(s, scope)
+}
+
+fn to_utf8_fast(
+  s: v8::Local<v8::String>,
+  scope: &mut v8::HandleScope,
+) -> Option<std::string::String> {
+  // Over-allocate by 20% to avoid checking string twice
+  let len = s.length();
+  let capacity = (s.length() as f64 * 1.2) as usize;
```
littledivy, line 681:
```suggestion
  let capacity = (len as f64 * 1.2) as usize;
```

serde_v8/de.rs
```diff
@@ -661,3 +661,57 @@ fn bigint_to_f64(b: v8::Local<v8::BigInt>) -> f64 {
...
+    Some(&mut nchars),
+    v8::WriteOptions::NO_NULL_TERMINATION
+      | v8::WriteOptions::REPLACE_INVALID_UTF8,
+  );
+  unsafe {
+    buf.set_len(length);
+  }
+  if nchars < len {
+    return None;
+  }
```
littledivy, line 697: nit: early exit
```suggestion
  if nchars < len {
    return None;
  }
  unsafe {
    buf.set_len(length);
  }
```

serde_v8/de.rs
```diff
@@ -661,3 +661,57 @@ fn bigint_to_f64(b: v8::Local<v8::BigInt>) -> f64 {
...
+}
+
+fn to_utf8_slow(
+  s: v8::Local<v8::String>,
+  scope: &mut v8::HandleScope,
+) -> std::string::String {
+  let capacity = s.utf8_length(scope);
+  let mut string = std::string::String::with_capacity(capacity);
+  let data = string.as_mut_ptr();
+  std::mem::forget(string);
```
littledivy, line 708: Maybe we can avoid `forget` and the `String::from_raw_parts` re-construct by just returning `string` at the end? (mutable reference does not live longer)

littledivy (commented):
looks neat! needs a tiny safety comment on top of `set_len`

serde_v8/de.rs
AaronO: That code was pretty much `v8::String::to_rust_string_lossy` untouched, I've now aligned it follow the same general design as the fast path and maybe we'll upstream these changes (or in part) to rusty_v8

littledivy (approved):
LGTM!

PR #14453, opened 2022-05-01
fix(vendor): do not panic on relative specifier with scheme-like folder name

dsherret:
I tried to vendor something for debugging purposes today and got a panic with jspm because it had some relative specifiers that looked like this:

```
import _regex2 from './npm:uc.micro@1.0.6/categories/Cc/regex!cjs';
```

```

Platform: windows x86_64
Version: 1.21.1
Args: ["C:\\Users\\david\\AppData\\Local\\bvm\\binaries\\denoland\\deno\\1.21.1\\bin\\deno.exe", "vendor", "--force", "_config.ts"]

thread 'main' panicked at 'Could not find local path for npm:uc.micro@1/categories/Cc/regex!cjs', cli\tools\vendor\mappings.rs:143:11
```

The reason was a join was being done on `npm:uc.micro@1.0.6/categories/Cc/regex!cjs`, which would make it go absolute because "npm:" looks like a scheme.

### DIFF
--- a/cli/tools/vendor/build.rs
+++ b/cli/tools/vendor/build.rs
@@ -619,6 +619,55 @@ mod test {
     );
   }
 
+  #[tokio::test]
+  async fn remote_relative_specifier_with_scheme_like_folder_name() {
+    let mut builder = VendorTestBuilder::with_default_setup();
+    let output = builder
+      .with_loader(|loader| {
+        loader
+          .add("/mod.ts", "import 'https://localhost/mod.ts';")
+          .add(
+            "https://localhost/mod.ts",
+            "import './npm:test@1.0.0/test/test!cjs';",
+          )
+          .add_with_headers(
+            "https://localhost/npm:test@1.0.0/test/test!cjs",
+            "console.log(5);",
+            &[("content-type", "application/javascript")],
+          );
+      })
+      .build()
+      .await
+      .unwrap();
+
+    assert_eq!(
+      output.import_map,
+      Some(json!({
+        "imports": {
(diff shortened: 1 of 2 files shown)


cli/tools/vendor/import_map.rs
```diff
@@ -226,7 +226,8 @@ fn handle_dep_specifier(
       let mut local_base_specifier = mappings.local_uri(base_specifier);
       local_base_specifier.set_query(unresolved_specifier.query());
       local_base_specifier = local_base_specifier
-        .join(&unresolved_specifier.path()[1..])
+        // path includes "/" so make it relative
+        .join(&format!(".{}", unresolved_specifier.path()))
```
dsherret, line 230: Makes the path be joined with `./` instead of whatever is after the `/`.

cli/tools/vendor/build.rs
```diff
@@ -619,6 +619,55 @@ mod test {
...
+  #[tokio::test]
+  async fn remote_relative_specifier_with_scheme_like_folder_name() {
+    let mut builder = VendorTestBuilder::with_default_setup();
+    let output = builder
+      .with_loader(|loader| {
+        loader
+          .add("/mod.ts", "import 'https://localhost/mod.ts';")
+          .add(
+            "https://localhost/mod.ts",
+            "import './npm:test@1.0.0/test/test!cjs';",
```
dsherret, line 631: Example of specifier that would cause a bug is here.

bartlomieju (approved):
LGTM

PR #14462, opened 2022-05-02
feat(unstable/task): resolve the current executable for the deno command

### LINKED ISSUE #14427: Deno task should resolve the Deno executable path
Hi. Let's say I have a task that executes a deno script like this:

```json
{
  "tasks": {
    "my-task": "deno run script.ts"
  }
}
```

I'd like `deno task my-task` to work in all environments, even if `deno` command doesn't exist. For example, netlify doesn't have Deno installed by default, so it's needed to install it before run it:

```
curl -fsSL https://deno.land/x/install/install.sh | sh && /opt/buildhome/.deno/bin/deno task my-task
```

But due `deno` doesn't exist in the `PATH`, this commands returns an error. I think `deno` keyword should be replaced by `Deno.execPath()` value.

dsherret:
Closes #14427

Has https://github.com/denoland/deno_task_shell/pull/38

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1074,9 +1074,9 @@ dependencies = [
 
 [[package]]
 name = "deno_task_shell"
-version = "0.2.1"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

Issue #14461, opened 2022-05-02
Remove Deno.emit


ry:
Should be handled by external code, not built into runtime. deno_ast, deno_graph largely contain this functionality 

Aiming for 1.22.0

dsherret:
I think we should create a wasm version of `Deno.emit` before doing this or have some documented alternatives? That way users can transition over to it.

ry:
We've been wanting to do this for over a year now - nothing has yet materialized in third party. We'll remove in 1.22 and that will provide a good impetus. Of course if it's done ahead of time, that would be better, but I don't want to delay this removal for more months.

AnInternetTroll:
A deprecated tag on the jsdoc would have been nice to show that you intend to remove Deno.emit. Otherwise it's just out of the blue

As of now there is nothing indicating any intention of removal on https://doc.deno.land/deno/unstable/~/Deno.emit and unless you follow all conversations on discord there never really was any outspoken intention of removal

AnInternetTroll:
That is true, but a bit more communication couldn't hurt

KnorpelSenf:
There's no indication at https://deno.land/manual/typescript/runtime that the feature will be removed, either. I wasn't aware that this is being considered.

Now that we suddenly have to rewrite `Deno.emit` very fast in user land before 1.22 hits, can you suggest where to start with this? For example, can we wrap `ts_morph`?

What do you think about moving this to `/std` instead of dropping it completely?

lowlighter:
> it’s unstable so anything can change at any time.

The [documentation about stability](https://deno.land/manual/runtime/stability) doesn't explicit state that said feature is subject to removal though:

> However, not all of Deno's features are ready for production yet. Features which are not ready, because they are still in draft phase, are locked behind the `--unstable` command line flag.
>
> You should be aware that many unstable APIs have **not undergone a security review**, are likely to have **breaking API changes** in the future, and are **not ready for production**.

It only suggest that a feature usage may change or is unsecure but will *eventually* stabilize, *not* that its support could be entirely dropped. 

I think a full removal implies a lot more than just "breaking API changes", even for unstable features, especially if there are no alternatives yet. 

Though it could be that I misunderstand what the documentation was trying to said too 🤔

buttercubz:
> This is a widely used API and not having an alternative ready to go will cause a huge amount of distrust in our user base. Opening an issue to removing something widely used in a release a couple weeks will break a lot of existing solutions, causing people to not be able to upgrade.
[...]

PR #14463, opened 2022-05-02
BREAKING: Remove unstable Deno.emit and Deno.formatDiagnostics APIs

bartlomieju:
Closes https://github.com/denoland/deno/issues/14461


dsherret (approved):
LGTM

kitsonk:
I strongly believe we shouldnt remove it until we have an alternative upgrade path. Removing it on a whim us irresponsible and not a sane plan.

ry (approved):
LGTM

Unfortunately after almost a year of planning we've still been unable to provide a userland version of this. We need to not stagnate on stuff like this - this is a small API used by a small number of projects. Removing it will force development of https://github.com/denoland/deno_emit

This is an unstable API - there was never a promise of continuity.

PR #14465, opened 2022-05-02
fix(vendor): handle relative imports when mapped local folder name differs from remote's

dsherret:
I was using deno vendor this weekend and found this bug.

```ts
// https://localhost/npm:test/file.js
export function test() {}

// https://localhost/npm:other/file.js
import "../npm:test/file.js";
```

In this case, the `npm:x` folders will be mapped to `npm_x`. So when `../npm:test/file.js?test` is resolved it can't find the folder because it's looking for `../npm_test`. See the tests for more details.

Maybe this is an import map bug though and shouldn't be fixed in deno vendor? Also, perhaps there is a better way than this fix?


bartlomieju (approved):
LGTM!

Issue #14458, opened 2022-05-02
Memory leak when using Deno.open and streams with serve()


sant123:

lucacasonato:
This is a bug. Thanks for reporting. @littledivy  probably the rust2rust streams optimization?

littledivy:
Yes, a bug...unfortunate that tests didn't catch this. 

@sant123 as a quick workaround: call `try { file.close() } catch { /** Pass **/ }` after sending the response.

sant123:
> Yes, a bug...unfortunate that tests didn't catch this.
> 
> @sant123 as a quick workaround: call `try { file.close() } catch { /** Pass **/ }` after sending the response.

Thank you! This is my workaround for now:

``` ts
import { serve } from "https://deno.land/std@0.137.0/http/server.ts";

serve(async (request) => {
  const url = new URL(request.url);

  if (url.pathname === "/") {
    const file = await Deno.open("./file.txt");
    console.log(Deno.resources());
    try {
      return new Response(file.readable, {
        headers: { "Content-Type": "text/plain; charset=utf-8" },
      });
    } finally {
      file.close();
    }
  } else {
    return new Response(null, {
      status: 404,
    });
  }
});

```

Master-Hash:
Well, in deno 1.21.2, [the workaround](#issuecomment-1114581926) breaks.

The Response body is empty, much like the file has been closed before the response is returned.

Why is that? The behavior is quite confusing... @sant123 @littledivy

littledivy:
> Well, in deno 1.21.2, [the workaround](#issuecomment-1114581926) breaks.
> 
> The Response body is empty, much like the file has been closed before the response is returned.
> 
> Why is that? The behavior is quite confusing... @sant123 @littledivy

The bug has been fixed, so you can get rid of the workaround now.

PR #14471, opened 2022-05-03
fix(ext/http): explicitly close resource after reading

littledivy:
Closes https://github.com/denoland/deno/issues/14458

### DIFF
--- a/ext/http/01_http.js
+++ b/ext/http/01_http.js
@@ -282,7 +282,7 @@
                 resourceRid,
               );
               readableStreamClose(respBody); // Release JS lock.
-              core.close(resourceRid);
+              core.tryClose(resourceRid);
             } catch (error) {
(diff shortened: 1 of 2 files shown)


piscisaureus (approved):

ext/http/01_http.js
```diff
@@ -282,6 +282,7 @@
                 resourceRid,
               );
               readableStreamClose(respBody); // Release JS lock.
+              core.tryClose(resourceRid);
```
lucacasonato, line 285: I'd do them the other way around, but otherwise LGTM

bartlomieju: Shouldn't both of them happen in `finally` block anyway?

littledivy: @bartlomieju  The `catch` block calls `reader.cancel()` which should close the resource.

Issue #14469, opened 2022-05-03
Remove unstable `Deno.applySourceMap()`


nayeemrmn:
We have `Deno.core.applySourceMap()` now which is the same thing, if we're using this for something in denoland (looks like just some very old broken and ignored [test](https://github.com/denoland/deno/blob/256dcb058ac2dd2a935076922265278c56264050/cli/tests/unit/error_stack_test.ts#L59)). This function is not fit for public API because it requires the target module to be `deno cache`d and that the input module specifier refers to the typescript file that was transpiled. A good but still rarely used API would work for any module by fetching it independently of the cache and be like:
```ts
Deno.applySourceMap({ fileName: "code_with_source_map_at_the_bottom.js", lineNumber: 3, columnNumber: 4 });
// { fileName: "filename_referenced_in_map.ts", lineNumber: 2, columnNumber: 2 }
```

dsherret:
It also looks like nobody uses this on github except forks of deno: https://cs.github.com/?scopeName=All+repos&scope=&q=Deno.applySourceMap

PR #14473, opened 2022-05-03
BREAKING: Remove unstable Deno.applySourceMap API

bartlomieju:
Closes https://github.com/denoland/deno/issues/14469

### DIFF
--- a/cli/diagnostics.rs
+++ b/cli/diagnostics.rs
@@ -39,7 +39,6 @@ const UNSTABLE_DENO_PROPS: &[&str] = &[
   "UnixConnectOptions",
   "UnixListenOptions",
   "addSignalListener",
-  "applySourceMap",
   "bench",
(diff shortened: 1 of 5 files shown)


dsherret (approved):
LGTM

cli/diagnostics.rs
```diff
@@ -39,7 +39,6 @@ const UNSTABLE_DENO_PROPS: &[&str] = &[
   "UnixConnectOptions",
   "UnixListenOptions",
   "addSignalListener",
-  "applySourceMap",
```
dsherret, line 42: I think we need to update the declaration file too?

bartlomieju: Good catch, I had `cli/dts` excluded in my search box and didn't spot it.

dsherret (approved):

Issue #14464, opened 2022-05-02
op_format_file_name should probably be moved to deno_core


bartlomieju:
This op is responsible for providing abbrevations for data URLs and is used by "pretty" formatting of errors in the CLI.

It seems to me that there's no need for it to be in the CLI, and could be contained in `deno_core` to simplify providing source maps even more.

@nayeemrmn thoughts on this?

PR #14474, opened 2022-05-03
refactor: add core.formatLocationFilename, remove op_format_filename

bartlomieju:
Closes https://github.com/denoland/deno/issues/14464

This commit moves "op_format_location" to "core/ops_builtin.rs"
and removes "Deno.core.createPrepareStackTrace" in favor of
"Deno.core.prepareStackTrace".

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -435,28 +435,22 @@ pub(crate) fn is_instance_of_error<'s>(
 const DATA_URL_ABBREV_THRESHOLD: usize = 150;
 
 pub fn format_file_name(file_name: &str) -> String {
-  if file_name.len() > DATA_URL_ABBREV_THRESHOLD {
-    if let Ok(url) = Url::parse(file_name) {
-      if url.scheme() == "data" {
-        let data_path = url.path();
-        if let Some(data_pieces) = data_path.split_once(',') {
-          let data_length = data_pieces.1.len();
-          if let Some(data_start) = data_pieces.1.get(0..20) {
-            if let Some(data_end) = data_pieces.1.get(data_length - 20..) {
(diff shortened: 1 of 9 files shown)


bartlomieju:
@nayeemrmn please take a look

AaronO (commented):
Moving it into core makes sense, however it looks like this can remain an op (in `ops_builtin.rs`) and doesn't need to be promoted to a v8 binding

core/error.rs
```diff
@@ -431,6 +432,33 @@ pub(crate) fn is_instance_of_error<'s>(
...
+                data_end
+              );
+            }
+          }
+        }
+      }
+    }
+  }
+  file_name.to_string()
+}
```
AaronO, line 460: Technically your PR doesn't change this code but moves it around, I think we can avoid the deep nested ifs with something like:

```suggestion
pub fn format_file_name(file_name: &str) -> String {
  abbrev_file_name(file_name).unwrap_or_else(|| file_name.to_string())
}

fn abbrev_file_name(file_name: &str) -> Option<String> {
  if file_name.len() <= DATA_URL_ABBREV_THRESHOLD {
    return None;
  }
  let url = Url::parse(file_name).ok()?;
  if url.scheme() != "data" {
    return None;
  }
  let (head, tail) = url.path().split_once(',')?;
  let len = tail.len();
  let start = tail.get(0..20)?;
  let end = tail.get(len - 20..)?;
  Some(format!("{}:{},{}......{}", url.scheme(), head, start, end))
}
```

AaronO (commented):
LGTM

AaronO (approved):

PR #14475, opened 2022-05-03
feat(core): Re-export v8 use_custom_libcxx feature

arthurprs:
Re-exports rusty v8 `use_custom_libcxx` feature added in https://github.com/denoland/rusty_v8/pull/924 so users of deno core can disable if if needed (e.g. when also bundling rocksdb in the same binary).


core/Cargo.toml
```diff
@@ -12,6 +12,10 @@ description = "A modern JavaScript/TypeScript runtime built with V8, Rust, and T
 [lib]
 path = "lib.rs"
 
+[features]
+default = ["v8_use_custom_libcxx"]
```
bartlomieju, line 16: Why do you want to make it a default feature?

arthurprs: It is the current default. I think it could be contentious to change it.

bartlomieju: Is it? Adding `default-features = false` to `v8` imports, suggests that it's not, or maybe I'm misunderstanding something?

arthurprs: It's a default feature of v8 https://github.com/denoland/rusty_v8/pull/924, so as of today it's the default of deno core.

With that said, the only way to make it optional to users of deno core is to make it `default-features=false` at the dependency level but enable via a feature. This new feature can then be enabled by default (preserving the status quo) but can be disabled by deno core users.

arthurprs:
Gentle ping, it would be amazing to get this and #13915 merged

bartlomieju:
Hey @arthurprs sorry for late response, please rebase and we'll merge it

arthurprs:
No problem. The PR is now merged with the upstream repo.

bartlomieju (approved):
LGTM, thanks!

PR #14477, opened 2022-05-03
chore(core/core.js): remove window prefix

linbingquan:
Remove `window` unused code for `core/01_core.js`.

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -25,10 +25,10 @@
     ObjectAssign,
     SymbolFor,
   } = window.__bootstrap.primordials;
-  const ops = window.Deno.core.ops;
+  const ops = Deno.core.ops;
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thank you, but this is done on purpose, there's a lot of places where `window.` is superfluous, but it helps with readability where various namespaces come from. Closing without merge.

linbingquan:
@bartlomieju Thank you for your reply, I agree with you. TBH, I compared `01_core.js` with `02_error.js` to make this PR.

I saw `Deno.core` without `window.` in `code/02_error.js`. 

https://github.com/denoland/deno/blob/253fbf9d2a15018cbba9c9ea2829bd70e9723362/core/02_error.js#L5

Issue #14478, opened 2022-05-03
steps_output_within is flaky


dsherret:
The test output capturing at the moment is flaky by design unfortunately. Essentially, the JS code is logging to stdout and stderr, which is piped via pipes to two threads that do blocking reads on the output pipes of stdout and stderr. This test output data is then sent across a channel to the test reporter.

What can occurr (especially on the CI), is the data is written to a stdout or stderror pipe, but the threads reading do not read and send the output message in time before the other thread sends the "test complete" notification message.

https://github.com/denoland/deno/runs/6277042335?check_suite_focus=true

To fix this, I believe we'll need a way to unblock the stdout and stderr reader threads and have them to signal back that they don't have any more data to read. I haven't figured out a way to wake up these reader threads though... initially I thought sending zero bytes over would work, but it seems like on Windows doing `WriteFile` on the anonymous pipe write handle with zero bytes does not wake up the read end even when flushing (I'm using all low level win api for testing purposes).

dsherret:
Resolved in #14479

cjihrig:
@dsherret I'm not sure that the flakiness is gone. While working on the release, I just had this test fail for me locally and then it passed on a second run.

dsherret:
That makes me sad.

PR #14479, opened 2022-05-03
chore: fix flaky `steps_output_within` test

dsherret:
See #14478 for more details on why this is required :(

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -36,6 +36,7 @@ use deno_core::futures::future;
 use deno_core::futures::stream;
 use deno_core::futures::FutureExt;
 use deno_core::futures::StreamExt;
+use deno_core::parking_lot::Mutex;
 use deno_core::serde_json::json;
 use deno_core::url::Url;
 use deno_core::ModuleSpecifier;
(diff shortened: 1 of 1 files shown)


cli/tools/test.rs
```diff
@@ -1525,19 +1560,32 @@ fn pipe_writer_to_file(writer: os_pipe::PipeWriter) -> std::fs::File {
...
+    if !data.is_empty()
+      && sender
+        .send(TestEvent::Output(buffer[0..size].to_vec()))
+        .is_err()
     {
       break;
     }
+
+    // Always respond back if this was set. Ideally we would also check to
+    // ensure the pipe reader is empty before sending back this response.
```
dsherret, line 1586: It seems this is possible on Windows (works with anonymous pipes even though the function name is deceiving: https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe), but I haven't found a way to peek at a pipe with libc.

kitsonk (approved):
LGTM

The little old lady who swallowed thr fly...

PR #14483, opened 2022-05-04
fix(bundle): source map encoding

bartlomieju:
Ref https://github.com/denoland/deno_ast/pull/88

Still need to add a test.


bartlomieju:
This is most likely not needed.

PR #14485, opened 2022-05-04
WIP: wasm modules in core

lucacasonato:


core/modules.rs
```diff
@@ -493,10 +497,11 @@ impl RecursiveModuleLoad {
     module_request: &ModuleRequest,
     module_source: &ModuleSource,
   ) -> Result<(), ModuleError> {
-    if module_request.expected_module_type != module_source.module_type {
+    let asserted_module_type = module_source.module_type.into();
```
bartlomieju, line 500:
```suggestion
    let expected_asserted_module_type = module_source.module_type.into();
```

aapoalas (changes requested):
This draft presumably would require a sizable number of changes, especially in light of semi-recent deno_graph refactors etc. Is the expectation still that this will eventually be finished, or should this be closed in favour of a later, fresh start?

bartlomieju:
This was blocked by V8 APIs from the get-go. I'm gonna close this PR for now since there's some movement on the proposal in TC39, which would make V8 implement this properly without workarounds on our side.

Issue #2552, opened 2019-06-20
Support importing wasm modules 


ry:
```
import * as foo from "./foo.wasm";
```
where foo will correspond to the exports of the wasm module.

kitsonk:
For reference, here is the [README](https://github.com/WebAssembly/esm-integration/blob/master/proposals/esm-integration/README.md) about the integration.  It is a Stage 1 proposal (and the WebAssembly committee follows the TC39 stage process), which means it is still quite early.  I think this could be important to Deno, so maybe getting involved early would be a good thing, but is likely to be a bit of a rough road.

This will present interesting challenges for TypeScript and trying to understand the exports of a WASM module, but I suspect this would be an iterative process, where it would be just an untyped import, but eventually could try to figure out a way to evaluate/type on the fly the imports and exports of a WASM Module.  It is going to be fairly complex I suspect, with some mediation between V8 and the compiler.

kitsonk:
Just an update, the standard has moved to Stage 2.  An important note in the proposal is that [TLA is potentially](https://github.com/WebAssembly/esm-integration/tree/master/proposals/esm-integration#why-does-this-proposal-depend-on-top-level-await) needed to evaluate the module and hand it back to the ESM importer, so it is likely we need a version of TLA in V8 before we can really address this.

sh7dm:
I have some ideas and I probably will work on this one. Compiler for that should produce smth like this:
```js
const { instance } = await WebAssembly.instantiate(Uint8Array.from(atob("<WASM as base64>"), c => c.charCodeAt(0)));
module.exports = instance.exports;
```

kevinkassimo:
We should be able to close this with #3328

hayd:
It'd be great to have a toy example in the docs/manual.

kevinkassimo:
@hayd that could be a good first issue 😄

kevinkassimo:
@sh7dm Ooops, did not realize that you are working on this too... Only noticed this issue just now.
But basically I also thought of the base64 solution (and I guess it might work well with bundling). I implemented the thing by generating a dynamic script.

Would love to hear your feedback since currently in #3348 we have double compilation issue (and I did a hack to placate TS compiler... which might not be the most elegant thing)

hayd:
@kevinkassimo I will add an example to the manual.

hayd:
@kevinkassimo am i missing something here:

```
 $ cat fib.as.ts
export function fib(n: i32): i32 {
[...]

PR #14487, opened 2022-05-04
refactor(core): use Box<u8> for ModuleSource.code instead of a String

bartlomieju:
Towards https://github.com/denoland/deno/issues/2552
Towards https://github.com/denoland/deno/pull/14485

### DIFF
--- a/cli/proc_state.rs
+++ b/cli/proc_state.rs
@@ -592,7 +592,7 @@ impl ProcState {
           }
         };
         Ok(ModuleSource {
-          code: code.as_bytes().to_vec(),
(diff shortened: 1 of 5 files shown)


cli/proc_state.rs
```diff
@@ -592,7 +592,7 @@ impl ProcState {
           }
         };
         Ok(ModuleSource {
-          code,
+          code: code.as_bytes().to_vec().into_boxed_slice(),
```
lucacasonato, line 595: `code.into_bytes().into_boxed_slice()` doesn't require a clone.

lucacasonato (approved):

PR #14490, opened 2022-05-04
feat(test): change "failures:" headers in test report

bartlomieju:
Closes https://github.com/denoland/deno/issues/14190

### DIFF
--- a/cli/tools/test.rs
+++ b/cli/tools/test.rs
@@ -452,7 +452,7 @@ impl TestReporter for PrettyTestReporter {
   fn report_summary(&mut self, summary: &TestSummary, elapsed: &Duration) {
     if !summary.failures.is_empty() {
       let mut failure_titles = vec![];
-      println!("\nfailures:\n");
+      println!("\n{}\n", colors::white_bold_on_red(" ERRORS "));
       for (description, js_error) in &summary.failures {
         let failure_title = format!(
           "{} {}",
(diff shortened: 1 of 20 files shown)


cli/tools/test.rs
```diff
@@ -452,7 +452,7 @@ impl TestReporter for PrettyTestReporter {
   fn report_summary(&mut self, summary: &TestSummary, elapsed: &Duration) {
     if !summary.failures.is_empty() {
       let mut failure_titles = vec![];
-      println!("\nfailures:\n");
+      println!("\n{}\n", colors::white_bold_on_red("  ERRORS  "));
```
nayeemrmn, line 455: Is it intentional that `ERRORS` has double spaces around it while `FAILURES`  has single spaces? I would prefer single like jest for both...

bartlomieju: Yeah, I wanted them to be the same width

nayeemrmn: I believe this style is supposed to be more like red-highlighted text rather than uniform red text boxes. The spaces are to add fixed horizontal padding since ansi backgrounds happen to look ugly without it. At least, it looks nicer that way IMO

bartlomieju: Can you explain a bit more what you are proposing?

nayeemrmn:
```suggestion
      println!("\n{}\n", colors::white_bold_on_red(" ERRORS "));
```
And to shift the mental model to 'red-highlighted text with single-space padding' rather than worrying about the red boxes being the same width.

kt3k (approved):
LGTM

Before
スクリーンショット 2022-05-09 13 33 38

After
スクリーンショット 2022-05-09 13 33 01

I like the new format. It looks more visually clear where the summary starts

PR #14501, opened 2022-05-05
refactor(core): add AssertedModuleType enum

bartlomieju:
Towards https://github.com/denoland/deno/issues/2552 
Towards https://github.com/denoland/deno/pull/14485


core/bindings.rs
```diff
@@ -343,7 +343,8 @@ pub extern "C" fn host_import_module_dynamically_callback(
       resolver.reject(tc_scope, e);
     }
   }
-  let module_type = get_module_type_from_assertions(&assertions);
+  let expected_module_type =
```
lucacasonato, line 346:
```suggestion
  let asserted_module_type =
```

core/bindings.rs
```diff
@@ -358,7 +359,7 @@ pub extern "C" fn host_import_module_dynamically_callback(
       module_map_rc,
       &specifier_str,
       &referrer_name_str,
-      module_type,
+      expected_module_type,
```
lucacasonato, line 362:
```suggestion
      asserted_module_type,
```

core/modules.rs
```diff
@@ -680,13 +683,42 @@ impl Stream for RecursiveModuleLoad {
...
+// probably be removed once WASM module support lands.
+#[allow(clippy::from_over_into)]
+impl Into<AssertedModuleType> for ModuleType {
+  fn into(self) -> AssertedModuleType {
+    match self {
+      ModuleType::JavaScript => AssertedModuleType::JavaScriptOrWasm,
+      ModuleType::Json => AssertedModuleType::Json,
+    }
+  }
+}
```
lucacasonato, line 704:
```suggestion
impl From<ModuleType> for AssertedModuleType {
  fn from(module_type: ModuleType) -> AssertedModuleType {
    match module_type {
      ModuleType::JavaScript => AssertedModuleType::JavaScriptOrWasm,
      ModuleType::Json => AssertedModuleType::Json,
    }
  }
}
```

I think you are confusing something here. Regardless of if you use `From` or `Into`, you will only be able to do `ModuleType -> AssertedModuleType`, not the other way around.

The only difference is that with `Into` you can only do `ModuleType::into`, while with `From` you can do `ModuleType::into` and `AssertedModuleType::from`.

Neither of them allow for `AssertedModuleType::into` or `ModuleType::from` (I think this is the one you are concerned about.

core/modules.rs
```diff
@@ -680,13 +683,42 @@ impl Stream for RecursiveModuleLoad {
...
+    match self {
+      Self::JavaScriptOrWasm => write!(f, "JavaScriptOrWasm"),
+      Self::Json => write!(f, "JSON"),
+    }
+  }
+}
+
 /// Describes what is the expected type of module, usually
 /// it's `ModuleType::JavaScript`, but if there were import assertions
 /// it might be `ModuleType::Json`.
```
lucacasonato, line 717:
```suggestion
/// Describes a request for a module as parsed from the source code.
/// Usually executable (`JavaScriptOrWasm`) is used, except when an
/// import assertions explicitly constrains an import to JSON, in
/// which case this will have a `AssertedModuleType::Json`.
```

lucacasonato (approved):
LGTM with nits

PR #14502, opened 2022-05-05
refactor(ext/url): URLPattern structured matchers

lucacasonato:
This commit updates URLPattern to use a structured matcher instead of
always using a RegExp. This results in an approximately 0% change in performance
 in the cases I tested. The real value of this PR is that it allows for
 very fast matching in the upcoming URLPatternList, because this
interface can be implemented to use a radix tree matcher internally.

The new matcher still falls back to RegExp matching if the pattern is
too complicated to be handled by the simple cases (for example if
backtracking is required to match).

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -638,7 +638,7 @@ fn clap_root(version: &str) -> Command {
 }
 
 fn bench_subcommand<'a>() -> Command<'a> {
-  runtime_args(Command::new("bench"), true, false)
+  runtime_args(Command::new("bench"), true, true)
     .trailing_var_arg(true)
     .arg(
       Arg::new("ignore")
@@ -2090,7 +2090,7 @@ fn unsafely_ignore_certificate_errors_arg<'a>() -> Arg<'a> {
 }
 
 fn bench_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
-  runtime_args_parse(flags, matches, true, false);
+  runtime_args_parse(flags, matches, true, true);
 
   // NOTE: `deno bench` always uses `--no-prompt`, tests shouldn't ever do
   // interactive prompts, unless done by user code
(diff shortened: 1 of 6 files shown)


ext/url/Cargo.toml
```diff
@@ -17,7 +17,7 @@ path = "lib.rs"
 deno_core = { version = "0.132.0", path = "../../core" }
 serde = { version = "1.0.136", features = ["derive"] }
 serde_repr = "0.1.7"
-urlpattern = "0.1.6"
+urlpattern = { git = "https://github.com/lucacasonato/rust-urlpattern", branch = "fast_matcher" }
```
crowlKats, line 20: update this

crowlKats (approved):
LGTM besides the git import

cli/flags.rs
```diff
@@ -638,7 +638,7 @@ fn clap_root(version: &str) -> Command {
 }
 
 fn bench_subcommand<'a>() -> Command<'a> {
-  runtime_args(Command::new("bench"), true, false)
```
bartlomieju, line 641: Unrelated change?

cli/tools/bench.rs
```diff
@@ -402,7 +402,7 @@ async fn bench_specifier(
         &specifier.to_file_path().unwrap().display().to_string(),
         false,
       )?;
-      worker.run_event_loop(false).await?;
+      worker.run_event_loop(true).await?;
```
bartlomieju, line 405: Ditto

bartlomieju (changes requested):

cli/tools/bench.rs
bartlomieju: Good catch, but let's do that in a separate PR. I thought I did exclude `--inspect` flags from `deno bench`

bartlomieju (approved):
LGTM, if you revert changes to `cli/flags.rs` and `cli/tools/bench.rs`

PR #14505, opened 2022-05-05
feat(cli): add deno version to manual links

### LINKED ISSUE #14250: All manual links in help text should include version number
For example, the `--config` help text includes the link https://deno.land/manual/getting_started/configuration_file but ought to specify the current version: https://deno.land/manual@v1.20.5/getting_started/configuration_file

hint use `crate::version::deno()`

Examples of where it should be updated:
https://github.com/denoland/deno/blob/a4eee007ef28f918db9165c52a19bc30bd65bad3/cli/flags.rs#L457
https://github.com/denoland/deno/blob/a4eee007ef28f918db9165c52a19bc30bd65bad3/cli/flags.rs#L1205
https://github.com/denoland/deno/blob/a4eee007ef28f918db9165c52a19bc30bd65bad3/cli/flags.rs#L1764

mrkldshv:
Adds current `deno` version to all manual links in CLI.

Closes #14250.

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -35,6 +35,14 @@ static LONG_VERSION: Lazy<String> = Lazy::new(|| {
   )
 });
 
+static SHORT_VERSION: Lazy<String> = Lazy::new(|| {
+  crate::version::deno()
+    .split('+')
+    .next()
+    .unwrap()
+    .to_string()
+});
+
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct BenchFlags {
   pub ignore: Vec<PathBuf>,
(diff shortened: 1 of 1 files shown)


cli/flags.rs
```diff
@@ -510,7 +512,10 @@ To execute a script:
 To evaluate code in the shell:
 
   deno eval \"console.log(30933 + 404)\"
-";
+",
+    crate::version::deno()
```
bartlomieju, line 516: This won't work for canary releases, because `version::deno()` returns semver + 7 characters from git hash. I think we should always split version on `+` sign to get latest version and call it a day

bartlomieju (changes requested):

cli/flags.rs
mrkldshv: Alright, didn't know about this for canary version. Will fix it!

mrkldshv: Should be fixed now

mrkldshv:
@bartlomieju thanks for review!

PR #14515, opened 2022-05-06
refactor(core): simplify json_module_evaluation_steps

bartlomieju:
When working on https://github.com/denoland/deno/pull/14485 and digging through
V8 code, we noticed that the return value of synthetic module evaluation steps callback
doesn't matter, and the promise is not used in any way. Changed to just returned
`undefined` so we're not getting confused. I originally took this idea from `d8`.


bartlomieju:
Maybe it is needed after all 🤔

Issue #13452, opened 2022-01-21
Disable auto-detection of config file


kitsonk:
I don't see anything in the [release notes](https://deno.com/blog/v1.18#auto-discovery-of-the-config-file) about how to disable this feature. How can I prevent deno from walking the directory hierarchy and discovering configs?

_Originally posted by @jsejcksn in https://github.com/denoland/deno/issues/13313#issuecomment-1018222760_

crowlKats:
maybe we should have `--config` take an *optional* value, so it would act as such:
`deno run foo.ts` -> no config detection
`deno run --config foo.ts` -> config detection
`deno run --config=./bar.json foo.ts` -> manual config

jsejcksn:
I like @crowlKats suggestion. Upon reading the release notes for `1.18.0`, I was very surprised to see this behavior enabled: it appears to be a deviation in the project's historical philosophy of preferring explicitness over opaque/automatic/magic defaults.

hayd:
This is also a problem with udd (and presumably other CLI scripts) which break if run in a directory with a incompatible deno.json: https://github.com/hayd/deno-udd/issues/59

kitsonk:
Several of us have talked about support remote configuration files before, I believe we should.  I have opened #13488 to cover that.

kt3k:
Summarizing hayd/deno-udd#59

1. They develop frontend app, and `deno.json` has `"lib": ["dom"]`
2. They try to update dependencies by `udd`. They run `deno run https://deno.land/x/udd/main.ts`
3. Here `deno.json` for frontend app is detected and applied to `udd` and causes `Cannot find name 'Deno'`

I feel the auto detection at step 3 is a questionable behavior. `deno.json` is usually for local files in my view. Is it correct to always auto-detect deno.json even when the entry script is a remote file?

kitsonk:
@kt3k in that scenario, I tend to agree.  If the main specifier is a remote specifier, then we shouldn't attempt to auto-resolve a configuration file.  The vision for the resolution was to resolve relative to the target of the script.  So either we attempt to resolve the remote target, or better for now, don't auto resolve.

jsejcksn:
> This issue has been [automatically marked as stale](https://probot.github.io/apps/stale/#how-it-works) because it has not had recent activity. It will be closed in 7 days if no further activity occurs. Thank you for your contributions.

This is still quite relevant.

PR #14520, opened 2022-05-07
[WIP] feat: support "permissions" in config file

### LINKED ISSUE #12763: Proposal: add permissions to config file

bartlomieju:
Closes #12763

TODO:
- [x] When a configuration file is applied and the "permissions" section is parsed and the permissions are applied from the "permissions" section. Any other flags on the command line are ignored.
- [x] If there are flags on the command line and a config file is being applied, and the config file contains "permissions" a warning should be issued to stderr that permissions from the config file are being applied.
- [ ] <s>Remote configuration files are supported, but a summary of the permissions is prompted before hand and requires user configuration to continue.</s> (see https://github.com/denoland/deno/issues/12763#issuecomment-1121199270)
- [ ] In situations where a base path is needed for relative paths (for --allow-read and --allow-write) the config file is used as a base, versus the cwd.

### DIFF
--- a/cli/config_file.rs
+++ b/cli/config_file.rs
@@ -531,6 +531,7 @@ pub struct ConfigFileJson {
   pub lint: Option<Value>,
   pub fmt: Option<Value>,
   pub tasks: Option<Value>,
+  pub permissions: Option<Value>,
 }
 
 #[derive(Clone, Debug)]
@@ -647,6 +648,10 @@ impl ConfigFile {
     } else {
       Ok(None)
     }
+  }
+
+  pub fn to_permissions(&self) -> Result<(), AnyError> {
+    todo!()
   }
 
   /// Return any tasks that are defined in the configuration file as a sequence
@@ -821,6 +826,13 @@ mod tests {
       "tasks": {
         "build": "deno run --allow-read --allow-write build.ts",
         "server": "deno run --allow-net --allow-read server.ts"
+      },
+      "permissions": {
+        "read": true,
+        "write": ["some/dir", "file.ts"],
+        "env": false,
+        "ffi": false,
+        "run": ["deno", "echo"]
       }
     }"#;
     let config_dir = ModuleSpecifier::parse("file:///deno/").unwrap();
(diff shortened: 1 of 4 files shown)


jsejcksn:
I think it is important for #13452 to be completed before this lands. It is core to Deno's security model that code is run in a sandbox with only explicitly granted permissions, and if config files are automatically detected (with no way of disabling that behavior), and config files include permissions, then that means that programs will be run with implicit permissions (with no way of disabling that). ⚠️

bartlomieju:
> I think it is important for #13452 to be completed before this lands. It is core to Deno's security model that code is run in a sandbox with only explicitly granted permissions, and if config files are automatically detected (with no way of disabling that behavior), and config files include permissions, then that means that programs will be run with implicit permissions (with no way of disabling that). ⚠️

@jsejcksn opened https://github.com/denoland/deno/pull/14555

bartlomieju:
There's no consensus about this feature in the core team. Closing for now.

Issue #14521, opened 2022-05-07
Map u64s and i64s from the FFI to bigints instead of numbers


joeltg:
First I just wanna say that using the FFI overall has been awesome, it's really really great to have this for Deno and I hope it stays around!

Right now FFI functions that return u64s or i64s get coerced to javascript numbers, losing precision for values greater than `Number.MAX_SAFE_INTEGER` or less than `Number.MIN_SAFE_INTEGER`.

In the other direction, it's not possible to pass a number

- greater than `2 ^ 32 - 1` (!!) as a `u64` parameter
- greater than `2 ^ 32 - 1` or less than `-2 ^ 31` (!!) as a `i64` parameter

I'm testing this with just

```c
// example.c
#include <stdint.h>

int64_t max_i64() { return 9223372036854775807; }

uint64_t max_u64() { return 18446744073709551615U; }

uint64_t identity_u64(uint64_t i) { return i; }

int64_t identity_i64(int64_t i) { return i; }
```

```
$ gcc -dynamiclib -o example.dylib example.c
```

```ts
const example = Deno.dlopen("example.dylib", {
	max_i64: { parameters: [], result: "i64" },
	max_u64: { parameters: [], result: "u64" },
	identity_u64: { parameters: ["u64"], result: "u64" },
	identity_i64: { parameters: ["i64"], result: "i64" },
})

console.log(example.symbols.max_i64()) // 9223372036854776000
console.log(example.symbols.max_u64()) // 18446744073709552000
console.log(example.symbols.identity_u64(4294967295)) // 4294967295
console.log(example.symbols.identity_u64(4294967296)) // Uncaught TypeError: Expected FFI argument to be an unsigned integer, but got Number(4294967296.0)
console.log(example.symbols.identity_i64(-2147483648)) // -2147483648
console.log(example.symbols.identity_i64(-2147483649)) // Uncaught TypeError: Expected FFI argument to be a signed integer, but got Number(-2147483649.0)
```

There are a few ways to improve the situation here:

- accept entire range of integers representable as javascript numbers, rounding when necessary
- return javascript numbers when possible without rounding, return bigints for values outside the safe integer range, accept bigints in addition to numbers
- require/always use bigints for all `u64` and `i64` values

They all have drawbacks but in my opinion the last option (always use bigints) is the cleanest and most consistent. Having an FFI that silently rounds is just not that useful.

PR #14523, opened 2022-05-07
feat(ext/ffi): support passing and returning bigints

### LINKED ISSUE #11641: ffi panic on integer parameters over 32bits

eliassjogreen:
Fixes #11641 and implements #14521. This changes it so that all integers can also be passed as `number | bigint`, internally using the `U32x2` type for conversions. All native types over 32 bits (`u64`, `i64`, `usize`, `isize`) now result in a `bigint` being returned.


aapoalas:
On a related note, why was the `U32x2` type needed again? Is there no way to handle the conversion cleanly at serde_v8 level from `v8::Local<v8::Value>` standing for a BigInt into `u64`, `i64` or `f64` directly?

bartlomieju:
@littledivy please review

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -360,17 +360,25 @@ declare namespace Deno {
   }
 
   /** All possible number types interfacing with foreign functions */
-  type StaticNativeNumberType = Exclude<NativeType, "void" | "pointer">;
+  type StaticNativeNumberType = Exclude<
+    NativeType,
+    "void" | "u64" | "i64" | "usize" | "isize" | "pointer"
+  >;
```
littledivy, line 366:
```suggestion
  type StaticNativeNumberType = Exclude<
    NativeType,
    "void" | "pointer" | StaticNativeBigIntType
  >;
```

littledivy (commented):
lgtm, can you rebase? that should fix the rustc cache keys issue

PR #14525, opened 2022-05-07
fix(ext/http): make serveHttp compress for Accept-Encoding: deflate, gzip

### LINKED ISSUE #14524: serveHttp does not compress response for Accept-Encoding: deflate, gzip
Given `Accept-Encoding: deflate, gzip`, which `curl --compressed` uses by default, `Deno.serveHttp` chooses to not compress the response, although it can do gzip compression.

```js
for await (const conn of Deno.listen({ port: 4500 })) {
  (async () => {
    for await (const { respondWith } of Deno.serveHttp(conn)) {
      respondWith(new Response("x".repeat(100000)));
    }
  })();
}
```

```
$ deno run -A helloweb.ts

$ curl --compressed -sI 'http://localhost:4500' | grep ^content
content-type: text/plain;charset=UTF-8
content-length: 100000
$ curl --compressed -vs 'http://localhost:4500' 2>&1 | grep -i accept-encoding:
> Accept-Encoding: deflate, gzip
$ curl --compressed -H 'accept-encoding: gzip' -sI 'http://localhost:4500' | grep ^content
content-type: text/plain;charset=UTF-8
content-encoding: gzip
content-length: 494
```

andy-k:
Closes #14524.

Half the code in this PR came from [an existing test case](https://github.com/denoland/deno/blob/bcd875030a2774a74051ef006c4a66436d43b3fa/cli/tests/unit/http_test.ts#L1857-L1919).

Half the code in this PR came from [fly-accept-encoding](https://github.com/superfly/accept-encoding/blob/69ef982a8cb6bdfa6c3f32ed028fa812842d4276/src/lib.rs#L86-L101), which understands Deflate.

### DIFF
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -1958,8 +1958,8 @@ Deno.test({
         "GET",
         "--url",
         url,
-        "--compressed",
-        "--header", // included for clarity although implied by --compressed
+        // "--compressed", // Windows curl does not support --compressed
+        "--header",
         "Accept-Encoding: deflate, gzip",
       ];
       const proc = Deno.run({ cmd, stdout: "piped", stderr: "null" });
--- a/cli/tests/unit/http_test.ts
+++ b/cli/tests/unit/http_test.ts
@@ -1918,6 +1918,66 @@ Deno.test({
   },
 });
 
+Deno.test({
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
This makes sense, thanks @andy-k!

AaronO (approved):
LGTM

PR #14535, opened 2022-05-08
[WIP] feat(unstable): add Deno.openPty()

### LINKED ISSUE #3994: Support a pty interface
Deno should support forking of pty interfaces (psuedo terminals) which would allow Deno workloads to be able to "pretend" to be a terminal, which can allow Deno workloads to be a terminal emulator or allow certain programs to think a Deno workload is an terminal to send it control sequences.

Node.js Prior art: https://github.com/microsoft/node-pty (which is used to provide the terminal interface in VSCode)
Likely this crate can be used: https://docs.rs/portable-pty/0.2.0/portable_pty/ which provides Linux/Mac and Windows support.

cc/ @Lerc

ihack2712:
Fixes #3994.

Uses [portable-pty](https://github.com/wez/wezterm/tree/main/pty) which supports Windows, Linux and Mac OS.

Ops:
- [ ] `op_openpty(args): { rid, bid, }`
  - [ ] Implement `read(b)` on Resource for reading output.
  - [ ] Implement `write(b)` on Resource for writing to input.
  - [ ] Closes on kill.
  - [ ] Closes when process dies.
- [ ] `op_kill(rid)`
- [ ] `op_resize(rid, console_size)`
- [ ] `op_size(rid) -> ConsoleSize`
- [ ] `op_wait(rid) -> Status`
- [ ] `op_status(id) -> Option<Status>`

Interface:

```ts
declare namespace Deno {
  interface Pty {
    rid: number;
    pid: number;
    output: ReadableStream<Uint8Array>;
    input: WritableStream<Uint8Array>;
    kill(): void;
    status(): undefined | number;
    wait(): Promise<number>;
    resize(consoleSize: ConsoleSize): void;
    size(): ConsoleSize;
    done(): boolean;
  }
  interface ConsoleSize {
    columns: number;
    rows: number;
  }
  interface PtyOptions extends ConsoleSize {
    cmd: string[];
    clearEnv?: boolean;
    env?: Record<string, string>;
  }
  openPty(options: PtyOptions): Promise<Pty>;
}
```


ihack2712:
@kitsonk Any feedback to the current structure?

nayeemrmn:
To be inline with the new `spawn()` API it should be more like:
```ts
declare namespace Deno {
  interface PtyChild {
    readonly pid: number;
    readonly status: Promise<ChildStatus>;
    kill(signo: Signal): void;
    size(): ConsoleSize;
    resize(consoleSize: ConsoleSize): void;
  }
  interface ConsoleSize {
    columns: number;
    rows: number;
  }
  interface SpawnPtyOptions extends ConsoleSize {
    args: string[];
    clearEnv?: boolean;
    env?: Record<string, string>;
  }
  /** Requires --allow-run */
  function spawnPty(command: string | URL, options: SpawnPtyOptions): PtyChild;
}
```

ihack2712:
> To be inline with the new `spawn()` API it should be more like:
> 
> ```ts
> declare namespace Deno {
>   interface PtyChild {
>     readonly readable: ReadableStream<Uint8Array>;
>     readonly writable: WritableStream<Uint8Array>;
>     readonly pid: number;
>     readonly status: Promise<ChildStatus>;
>     kill(signo: Signal): void;
>     size(): ConsoleSize;
>     resize(consoleSize: ConsoleSize): void;
>   }
>   interface ConsoleSize {
>     columns: number;
>     rows: number;
>   }
>   interface SpawnPtyOptions extends ConsoleSize {
>     args: string[];
>     clearEnv?: boolean;
>     env?: Record<string, string>;
>   }
>   /** Requires --allow-run */
>   function spawnPty(command: string | URL, options: SpawnPtyOptions): PtyChild;
> }
> ```

Agreed! @nayeemrmn 

1. AFAIK `portable_pty` doesn't offer `signal` arg on `kill()`.
2. `portable_pty` offers a way to wait for a pty to exit, and a way to get the return code, or none if the process hasn't exited yet. I think both should be available?

nayeemrmn:
> and a way to get the return code, or none if the process hasn't exited yet

A way to poll for the return code, yeah that was a discussion with `spawn()` as well and was decided as kind of an antipattern in JS.

ihack2712:
> > and a way to get the return code, or none if the process hasn't exited yet
> 
> A way to poll for the return code, yeah that was a discussion with `spawn()` as well and was decided as kind of an antipattern in JS.

I see, I guess there's a point there! Let's keep the `await .status` syntax ;)

ihack2712:
> the usecase for this is when you need to start a process in a tty other than the currently running tty. Instead of output being sent to the main processes stdout and stderr it is sent to the PTY's stdout and stderr. And it won't receive any input unless you specifically write to the PTY. This can allow you to run two child processes at once in the same Deno session, showing two outputs, switching between which process get data from stdin. You can send the WINCH signal to resize the amount of rows and columns the PTY has and so on. With this you can essentially make tmux on Deno
https://discord.com/channels/684898665143206084/688040863313428503/973217262913867786

runtime/Cargo.toml
```diff
@@ -82,6 +82,7 @@ sys-info = "0.9.1"
 termcolor = "1.1.3"
 tokio = { version = "1.17", features = ["full"] }
 uuid = { version = "1.0.0", features = ["v4"] }
+portable-pty = "0.7.0"
```
dsherret, line 85: I wonder if we could instead refactor and reuse the code in test_util/src/pty.rs

imjamesb: > I wonder if we could instead refactor and reuse the code in test_util/src/pty.rs

Maybe, but it needs to be async on the input and output, so some tokio magic would have to be used. I'll take a look and see what I find out!

imjamesb: Seems doable, I'm gonna try to get some work done on it, but most likely won't really get to it until next week.

PR #14538, opened 2022-05-09
feat(runtime/spawn): add `AbortSignal` support

crowlKats:
Equivalent of https://github.com/denoland/deno/pull/11579 for the new API

### DIFF
--- a/cli/tests/unit/spawn_test.ts
+++ b/cli/tests/unit/spawn_test.ts
@@ -264,7 +264,11 @@ Deno.test(
     queueMicrotask(() => ac.abort());
     const status = await child.status;
     assertEquals(status.success, false);
-    assertEquals(status.code, 143);
+    if (Deno.build.os === "windows") {
+      assertEquals(status.code, 1);
+    } else {
+      assertEquals(status.code, 143);
+    }
     assertEquals(status.signal, 15);
   },
 );
--- a/cli/tests/unit/spawn_test.ts
+++ b/cli/tests/unit/spawn_test.ts
@@ -264,11 +264,7 @@ Deno.test(
     queueMicrotask(() => ac.abort());
     const status = await child.status;
     assertEquals(status.success, false);
-    if (Deno.build.os === "windows") {
-      assertEquals(status.code, 1);
-    } else {
-      assertEquals(status.code, 143);
-    }
+    assertEquals(status.code, 143);
     assertEquals(status.signal, 15);
   },
 );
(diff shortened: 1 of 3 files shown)


runtime/js/40_spawn.js
```diff
@@ -119,8 +125,12 @@
         this.#stderr = readableStreamForRid(stderrRid);
       }
 
+      const onAbort = () => this.kill("SIGTERM");
+      signal?.[add](onAbort);
+
       this.#status = core.opAsync("op_spawn_wait", this.#rid).then((res) => {
         this.#rid = null;
+        signal?.[remove](onAbort);
```
dsherret, line 133: Should this be done in a `.finally(...)` so this gets unsubscribed if it ever throws?

cli/tests/unit/spawn_test.ts
```diff
@@ -248,6 +248,27 @@ Deno.test(
...
+        "setTimeout(console.log, 1e8)",
+      ],
+      signal: ac.signal,
+      stdout: "null",
+      stderr: "null",
+    });
+    queueMicrotask(() => ac.abort());
+    const status = await child.status;
+    assertEquals(status.success, false);
+    assertEquals(status.code, 143);
```
dsherret, line 267: Windows doesn't seem to use special exit codes based on the signal sent, but will just have a non-zero exit code (seems to be application specific).

One of the kill tests does the following:

```suggestion
  if (Deno.build.os === "windows") {
    assertEquals(status.code, 1);
    assertEquals(status.signal, undefined);
  } else {
    assertEquals(status.success, false);
    assertEquals(status.code, 143);
  }
```

cli/tests/unit/spawn_test.ts
```diff
@@ -248,6 +248,27 @@ Deno.test(
...
+    assertEquals(status.signal, 15);
```
dsherret, line 268: Windows doesn't seem to use special exit codes based on the signal sent, but will just have a non-zero exit code (seems to be application specific).

One of the kill tests does the following:

```suggestion
  if (Deno.build.os === "windows") {
    assertEquals(status.code, 1);
    assertEquals(status.signal, undefined);
  } else {
    assertEquals(status.success, false);
    assertEquals(status.code, 143);
  }
```

runtime/js/40_spawn.js
crowlKats: maybe, though I dont think there is a scenario this would throw, and if it does, then i'd consider that a bug to some degree

Issue #14491, opened 2022-05-05
Inconsistency in API between `Child.kill()` and `ChildStatus.signal`


lowlighter:
To kill a child you need to send a `Deno.Signal`, but the child status report a `number` instead
It may make more sense to use the same type for both

https://doc.deno.land/deno/unstable/~/Deno.ChildStatus
```ts
type ChildStatus = {
  success: true;
  code: 0;
  signal: null;
} | {
  success: false;
  code: number;
  signal: number | null;
};
```

https://doc.deno.land/deno/unstable/~/Deno.Child#kill
```ts
class Child<T extends SpawnOptions> {
  kill(signo: Signal): void;
}
```

PR #14539, opened 2022-05-09
refactor(runtime): change from signal_str_to_int function to enum

crowlKats:
Towards https://github.com/denoland/deno/issues/14491


runtime/ops/signal.rs
```diff
@@ -61,181 +63,173 @@ impl Resource for SignalStreamResource {
...
+  SIGBUS = 10,
+  SIGSEGV = 11,
+  SIGSYS = 12,
+  SIGPIPE = 13,
+  SIGALRM = 14,
+  SIGTERM = 15,
+  SIGUSR1 = 16,
+  SIGUSR2 = 17,
+  SIGCLD = 18,
+  SIGCHLD = 18,
```
crowlKats, line 197: this is a problem, unsure what to do about this

runtime/ops/signal.rs
```diff
@@ -61,181 +63,173 @@ impl Resource for SignalStreamResource {
...
+  SIGPWR = 19,
+  SIGWINCH = 20,
+  SIGURG = 21,
+  SIGPOLL = 22,
+  SIGIO = 22,
```
crowlKats, line 202: Ditto

crowlKats:
@ichizok I am removing the `SIGCHLD` & `SIGIO` signals as they cause issues when turning signals into an enum. Is this going to be a problem for Solaris?

ichizok:
@crowlKats I think should keep `SIGCHLD` since other platforms also have it, and rather remove `SIGCLD`.

crowlKats:
Sure, makes sense

runtime/ops/process.rs
```diff
@@ -291,16 +291,15 @@ async fn op_run_status(
...
   use nix::sys::signal::{kill as unix_kill, Signal};
   use nix::unistd::Pid;
-  let sig = Signal::try_from(signo)?;
+  let sig = Signal::try_from(signal as libc::c_int)?;
   unix_kill(Pid::from_raw(pid), Option::Some(sig)).map_err(AnyError::from)
 }
 
 #[cfg(not(unix))]
-pub fn kill(pid: i32, signal: &str) -> Result<(), AnyError> {
+pub fn kill(pid: i32, signal: super::signal::Signal) -> Result<(), AnyError> {
```
dsherret, line 302: `signal` is no longer used?

kt3k:
The error message looks less intelligible with this change to me.

Before
```
error: Uncaught TypeError: Invalid signal: SIGXYZ
Deno.kill(Deno.pid, "SIGXYZ");
     ^
```

After
```
error: Uncaught TypeError: Error parsing args at position 1: unknown variant `SIGXYZ`, expected one of `SIGHUP`, `SIGINT`, `SIGQUIT`, `SIGILL`, `SIGTRAP`, `SIGABRT`, `SIGEMT`, `SIGFPE`, `SIGKILL`, `SIGBUS`, `SIGSEGV`, `SIGSYS`, `SIGPIPE`, `SIGALRM`, `SIGTERM`, `SIGURG`, `SIGSTOP`, `SIGTSTP`, `SIGCONT`, `SIGCHLD`, `SIGTTIN`, `SIGTTOU`, `SIGIO`, `SIGXCPU`, `SIGXFSZ`, `SIGVTALRM`, `SIGPROF`, `SIGWINCH`, `SIGINFO`, `SIGUSR1`, `SIGUSR2`
Deno.kill(Deno.pid, "SIGXYZ");
     ^
```

What do you think?

Is the change to enum absolutely necessary for addressing #14491 ?

crowlKats:
ok that error message is atrocious, there ought to be a way to improve that... 

> Is the change to enum absolutely necessary for addressing https://github.com/denoland/deno/issues/14491 ?

well, if we dont go the enum route, we need to have a `signal_str_to_int` and `signal_int_to_str`, which is just not great for maintainability

kt3k:
@crowlKats 
> well, if we dont go the enum route, we need to have a signal_str_to_int and signal_int_to_str, which is just not great for maintainability

That kind of code bloat isn't that harmful to maintainability in my opinion. Because they are just data, it doesn't add much complexity to code base. Also they don't need to be updated that often.

crowlKats:
Well, we are looking exactly for behaviour of an enum, so it just makes sense we use an enum imo. Also, thought more about the error, and yes it isnt the most beautiful thing, but i think its fine

bartlomieju (approved):

Issue #14494, opened 2022-05-05
thread 'main' panicked at 'assertion failed: self.wr_state >= State::TlsClosed', ext\net\ops_tls.rs:426:5


masx200:
```
============================================================
Deno has panicked. This is a bug in Deno. Please report this
at https://github.com/denoland/deno/issues/new.
If you can reliably reproduce this panic, include the
reproduction steps and re-run with the RUST_BACKTRACE=1 env
var set and include the backtrace in your report.

Platform: windows x86_64
Version: 1.21.1
Args: ["deno", "run", "-A", "--unstable", "./deno/index.ts"]

thread 'main' panicked at 'assertion failed: self.wr_state >= State::TlsClosed', ext\net\ops_tls.rs:426:5
stack backtrace:
   0:     0x7ff7b29bbf44 - <unknown>
   1:     0x7ff7b1f365aa - <unknown>
   2:     0x7ff7b29af8a7 - <unknown>
   3:     0x7ff7b29bfe2b - <unknown>
   4:     0x7ff7b29bf9e1 - <unknown>
   5:     0x7ff7b1ebf9ea - <unknown>
   6:     0x7ff7b29c087d - <unknown>
   7:     0x7ff7b29c0683 - <unknown>
   8:     0x7ff7b29c0619 - <unknown>
   9:     0x7ff7b29c05d4 - <unknown>
  10:     0x7ff7b3a8da40 - CrashForExceptionInNonABICompliantCodeRange
  11:     0x7ff7b3a8db9c - CrashForExceptionInNonABICompliantCodeRange
  12:     0x7ff7b2261c5c - <unknown>
  13:     0x7ff7b226219f - <unknown>
  14:     0x7ff7b2b824b6 - <unknown>
  15:     0x7ff7b1a96628 - <unknown>
  16:     0x7ff7b1a9230b - <unknown>
  17:     0x7ff7b1a91508 - <unknown>
  18:     0x7ff7b1a91041 - <unknown>
  19:     0x7ff7b1a91154 - <unknown>
  20:     0x7ff7b3a402bc - CrashForExceptionInNonABICompliantCodeRange
  21:     0x7ff82a037034 - BaseThreadInitThunk
  22:     0x7ff82a4c2651 - RtlUserThreadStart

```

Running this will throw an error.
```ts

const c=await Deno.connectTls({port:443,hostname:'www.baidu.com'});c.close()

```

Running this will not throw an error.

```ts
[...]

PR #14547, opened 2022-05-09
fix: finish TLS handshake before shutting down

cjihrig:
This commit ensures that the TLS handshake completes before
attempting to close the connection.

Fixes: https://github.com/denoland/deno/issues/14494

This fixes the bug in the linked issue, but creating a reproducible unit test has been... frustrating.


lucacasonato (approved):
Unfortunate that there is no test, but let's land anyway. Many folks seem to be running into this panic.

bartlomieju (approved):

PR #14549, opened 2022-05-09
refactor: remove unused `Option`s on `StdFileResource.fs_file`

dsherret:
The option wasn't being used and adding some complexity.

### DIFF
--- a/runtime/ops/io.rs
+++ b/runtime/ops/io.rs
@@ -301,11 +301,10 @@ impl Resource for ChildStderrResource {
   }
 }
 
-type MaybeSharedStdFile = Option<Arc<Mutex<StdFile>>>;
-
 #[derive(Default)]
 pub struct StdFileResource {
-  fs_file: (MaybeSharedStdFile, RefCell<FileMetadata>),
+  fs_file: Option<Arc<Mutex<StdFile>>>,
+  metadata: RefCell<FileMetadata>,
   cancel: CancelHandle,
   name: String,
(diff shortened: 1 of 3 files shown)


runtime/ops/tty.rs
```diff
@@ -91,20 +90,9 @@ fn op_set_raw(state: &mut OpState, args: SetRawArgs) -> Result<(), AnyError> {
...
-      let std_file = file.lock().unwrap();
-      let raw_handle = std_file.as_raw_handle();
-      Ok(raw_handle)
-    } else {
-      Err(resource_unavailable())
-    };
-
-    let handle = handle_result?;
+    let std_file = resource.std_file();
+    let std_file = std_file.lock().unwrap(); // hold the lock
```
dsherret, line 94: I think it's more correct to hold the lock here if use of this is meant to be exclusive? Not sure though.

runtime/ops/io.rs
```diff
@@ -305,66 +305,64 @@ type MaybeSharedStdFile = Option<Arc<Mutex<StdFile>>>;
 
 #[derive(Default)]
 pub struct StdFileResource {
-  pub fs_file: Option<(MaybeSharedStdFile, Option<RefCell<FileMetadata>>)>,
+  fs_file: (MaybeSharedStdFile, RefCell<FileMetadata>),
```
dsherret, line 308: Actually, we can remove this tuple now.

PR #14552, opened 2022-05-10
cleanup(ext/http): simpler http write ops

AaronO:
Facilitates making `op_http_write_headers` sync and thus faster (will do in follow-up, already experimented locally)

### DIFF
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -622,7 +622,7 @@ fn http_response(
 // ensure it is a Weak Etag header ("W/").
 fn weaken_etag(hmap: &mut hyper::HeaderMap) {
   if let Some(etag) = hmap.get_mut(hyper::header::ETAG) {
-    if etag.as_bytes().starts_with(b"W/") {
+    if !etag.as_bytes().starts_with(b"W/") {
       let mut v = Vec::with_capacity(etag.as_bytes().len() + 2);
       v.extend(b"W/");
       v.extend(etag.as_bytes());
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -669,7 +669,7 @@ fn should_compress(headers: &hyper::HeaderMap) -> bool {
   !content_range
     && !cache_control_no_transform(headers).unwrap_or_default()
     && headers
-      .get(hyper::header::CONTENT_ENCODING)
+      .get(hyper::header::CONTENT_TYPE)
       .map(compressible::is_content_compressible)
       .unwrap_or_default()
 }
--- a/ext/http/lib.rs
+++ b/ext/http/lib.rs
@@ -39,6 +39,8 @@ use flate2::write::GzEncoder;
 use flate2::Compression;
 use fly_accept_encoding::Encoding;
 use hyper::body::Bytes;
+use hyper::header::HeaderName;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, much cleaner

lucacasonato:
@AaronO Please don't mark PRs as pure "cleanup" if they rewrite code to cause behavioral changes. Some examples:

- much more loose about invalid headers
- compresses irregardless of if `content-encoding` is already set on the response
- handles invalid UTF8 vary headers differently
- has an early return **true** inside of the `should_cache` function

At least half of this should have been caught in review.

Please revert this PR, or fix the above issues.

AaronO:
> * compresses irregardless of if `content-encoding` is already set on the response

My bad, submitted a regression fix here https://github.com/denoland/deno/pull/14641

> much more loose about invalid headers

Could you elaborate ?

> handles invalid UTF8 vary headers differently

Not sure this is true, logic is effectively (which I believe is preserved in both cases):
```rs
vary =
  | utf8 && lowecase -> add prefix,
  | utf8 && !lowercase -> untouched,
  | _ -> "Accept-Encoding",
```

> has an early return true inside of the should_cache function when a header is invalid
I believe this behaviour is also unchanged. We're ultimately impacting the predicate of compressing based on headers, which is negatively controlled by `cache_control_no_transform`:

```rs
cache_control_prevents_compressing =
  | utf8 && Some(c) => c.no_transform,
  | utf8 && None => false,
  | !utf8 => true,
  | _ -> false,
  
headers_allow_compression = !cache_control_prevents_compressing && ...;
```

lucacasonato:
> has an early return true inside of the should_cache function when a header is invalid

I mean this early return. If the `cache-control` header is invalid UTF-8, we compress irregardless of if `content-range` and friends are set. https://github.com/denoland/deno/pull/14552/files#diff-90dfe8e7faf01dad4a43d5f274862d01e5c6c53c07ec35ec28f6d73189b2f52fR659

> Not sure this is true, logic is effectively (which I believe is preserved in both cases):

Sorry, you're right I misread the code here.


> > much more loose about invalid headers
>
> Could you elaborate ?

I mean this `filter_map` here. We just skip over invalid header names and values now, rather than propagating the error of the header being invalid like we did previously. https://github.com/denoland/deno/pull/14552/files#diff-90dfe8e7faf01dad4a43d5f274862d01e5c6c53c07ec35ec28f6d73189b2f52fR507-R513

AaronO:
> I mean this early return. If the `cache-control` header is invalid UTF-8, we compress irregardless of if `content-range` and friends are set. https://github.com/denoland/deno/pull/14552/files#diff-90dfe8e7faf01dad4a43d5f274862d01e5c6c53c07ec35ec28f6d73189b2f52fR659

The early return is in an inner function not the outer, will clarify these predicates in a follow-up PR so they're easier to grok

> I mean this filter_map here. We just skip over invalid header names and values now, rather than propagating the error of the header being invalid like we did previously. https://github.com/denoland/deno/pull/14552/files#diff-90dfe8e7faf01dad4a43d5f274862d01e5c6c53c07ec35ec28f6d73189b2f52fR507-R513

Discarding invalid headers could indeed be fair, but I'll revert back to erroring on invalid headers

PR #14555, opened 2022-05-10
feat: add --no-config flag

bartlomieju:
Closes https://github.com/denoland/deno/issues/13452

This flag disables loading of configuration file, ie. it will not be
automatically discovered and loaded. Of course this flag conflicts
with "--config" flag and they cannot be used together.


cli/flags.rs
```diff
@@ -1068,6 +1070,7 @@ Ignore formatting a file by adding an ignore comment at the top of the file:
   // deno-fmt-ignore-file",
     )
     .arg(config_arg())
+    .arg(no_config_arg())
```
dsherret, line 1073: It's slightly error prone to have to provide both of these each time. Is there a way we could combine them into one function that adds both args at the same time and force the use of that? I think we do that elsewhere IIRC.

cli/flags.rs
```diff
@@ -298,6 +298,8 @@ pub struct Flags {
   pub future_type_check_mode: FutureTypeCheckMode,
   pub has_check_flag: bool,
   pub config_path: Option<String>,
+  // Disable loading of configuration file
+  pub no_config: bool,
```
dsherret, line 302: Nitpick: Maybe we could represent this in the struct as a single property in order to force it to be checked? Right now someone could access `config_path` and maybe forget to check `no_config`.

dsherret (approved):
LGTM

cli/flags.rs
bartlomieju: I've added it, but that's a lot of additional code. Not sure if it's worth it.

bartlomieju:
Note to self: Remember to land https://github.com/denoland/manual/pull/316 when landing this PR

cli/flags.rs
dsherret: Looks good now with the enum 👍

Issue #14431, opened 2022-04-29
FFI performance improvements (JIT compiler)


littledivy:
- [x] Value (de)serialization is slow. It has to go through JS -> `serde_json::Value` -> Rust, a faster path would be to convert to/fro a `serde_v8::Value` instead. https://github.com/denoland/deno/pull/14663
- [x] Each call has to go through the (already very thin) op layer. A possible solution is to make v8 callbacks during registration. This will get rid of the overhead of OpState. https://github.com/denoland/deno/pull/14945
- [x] JIT compiler for FFI calls. https://github.com/denoland/deno/pull/15006

littledivy:
Resolved with https://github.com/denoland/deno/pull/15125 and https://github.com/denoland/deno/pull/15139

PR #14556, opened 2022-05-10
[wip] perf(ext/ffi): Low overhead v8::Value <-> FFI (de)serialization

littledivy:
Fixes https://github.com/denoland/deno/issues/14431

TODO: 
- [x] Get's rid of `serde_json::Value`.
- [x] FFI calls are no longer 'ops' (not auto-generated with deno_ops)
- [ ] Throw errors in JS (don't `unwrap` lol)
- [ ] `tokio::spawn_blocking` for async calls.

Benchmarks -

Deno 1.12.1:
```
cpu: Apple M1
runtime: deno 1.21.1 (aarch64-apple-darwin)

file:///Users/divy/gh/deno/test_ffi/bench.js
benchmark      time (avg)             (min … max)       p75       p99      p995
------------------------------------------------- -----------------------------
nop()      456.78 ns/iter (440.37 ns … 903.48 ns) 445.07 ns 727.85 ns 903.48 ns
add(1,2)   678.34 ns/iter (662.85 ns … 694.42 ns) 681.14 ns 694.42 ns 694.42 ns
```

This PR:
```
cpu: Apple M1
runtime: deno 1.21.2 (aarch64-apple-darwin)

file:///Users/divy/gh/deno/test_ffi/bench.js
benchmark      time (avg)             (min … max)       p75       p99      p995
------------------------------------------------- -----------------------------
nop()      152.12 ns/iter  (144.67 ns … 261.4 ns) 150.43 ns 232.65 ns  252.9 ns
add(1,2)   260.75 ns/iter (253.98 ns … 268.91 ns) 261.96 ns 267.75 ns 267.81 ns
```


lucacasonato (commented):
Why can this not be done with ops + [`serde_v8::Value`](https://docs.rs/serde_v8/latest/serde_v8/struct.Value.html)?

littledivy:
> Why can this not be done with ops + [`serde_v8::Value`](https://docs.rs/serde_v8/latest/serde_v8/struct.Value.html)?

@lucacasonato We need the handle scope too.

ext/ffi/lib.rs
```diff
@@ -756,44 +558,421 @@ fn op_ffi_get_static(
...
+  mut rv: deno_core::v8::ReturnValue,
+) {
+  // SAFETY: deno_core guarantees args.data() is a v8 External pointing to an OpCtx for the isolates lifetime
+  let ctx = unsafe {
+    &*(v8::Local::<v8::External>::cast(args.data().unwrap_unchecked()).value()
+      as *const deno_core::_ops::OpCtx)
+  };
+  let rid = args.get(0).uint32_value(scope).unwrap();
+  let symbol = args.get(1).to_string(scope).unwrap();
+  let symbol = symbol.to_rust_string_lossy(scope);
```
aapoalas, line 672: Hmmm, probably a totally irrelevant performance optimisation but I wonder if the symbols could be kept in a Map keyed by the V8's internal UTF-16 representation to avoid the need to convert between V8 strings and Rust strings...

ext/ffi/lib.rs
```diff
@@ -756,44 +558,421 @@ fn op_ffi_get_static(
...
+  // Track async call & get copy of get_error_class_fn
+  let get_class = {
+    let state = state.borrow();
+    state.tracker.track_async(op_id);
+    state.get_error_class_fn
+  };
+
+  let pointer: U32x2 = serde_v8::from_v8(scope, args.get(1)).unwrap();
+  let fn_def: ForeignFunction = serde_v8::from_v8(scope, args.get(2)).unwrap();
+  let symbol = get_symbol(pointer.into(), fn_def);
```
aapoalas, line 782: Definitely not a thing for this PR, but eventually it might be useful to divide the function pointer symbol creation and calling into separate ops, where the `UnsafeFnPointer` constructor would already create the symbol and use garbage collectable resources to bind the symbol's lifetime to the `UnsafeFnPointer` instance, after which calling would be nearly 100% equivalent to a normal symbol call. The only difference would be the resource to symbol getting.

aapoalas (commented):
While the exposing and usage of more direct deno_core / rusty_v8 APIs is regrettable, this stuff and the performance improvement achieved is pretty amazing stuff.

Of course, this'll also then nicely pave the way for more reasonable functions for FFI callbacks, where the scope / isolate really becomes a pain in the butt to get a handle on without this more direct API.

littledivy:
This has mostly been done in #14663 and follow ups.

PR #14559, opened 2022-05-10
fix: stdout and stderr encoding on Windows

### LINKED ISSUE #6001: windows doesn't print emojis correctly
Hi! This is my first time looking at Deno, and it seems to be very exciting :)

I was playing around with the examples from the tutorial and noticed this inconsistency: (using powershell and windows terminal):

```
> deno run https://deno.land/std/examples/welcome.ts
Welcome to Deno 🦕
> deno run --allow-net https://deno.land/std/examples/curl.ts https://deno.land/std/examples/welcome.ts
console.log("Welcome to Deno ­ƒªò");
```

I found that editing `curl.ts` to add the line 
```ts
import { decode } from "https://deno.land/std/encoding/utf8.ts";
```
and replacing the `Deno.stdout.write` line with
```ts
console.log(decode(body));
```
seems to make the downloaded file print out correctly.

### LINKED ISSUE #14516: The console display is abnormal on Windows when console encoding is not utf-8
implement:
```js
console.log("你好，世界")
```
output:
```
浣犲ソ锛屼笘鐣?
```
Deno:1.21.2
OS:Windows 11

### LINKED ISSUE #14545: regression: `confirm` sometimes doesn't show prompt before receiving input on Windows

dsherret:
Given:

```ts
const encoder = new TextEncoder();

// stdout
{
  console.log("你好，世界");
  const size = await Deno.stdout.write(encoder.encode("🦕\n"));
  if (size !== 5) {
    throw new Error();
  }
}

// stderr
{
  console.error("你好，世界");
  const size = await Deno.stderr.write(encoder.encode("🦕\n"));
  if (size !== 5) {
    throw new Error();
  }
}
```

Before:

```
õ¢áÕÑ¢´╝îõ©ûþòî
­ƒªò
õ¢áÕÑ¢´╝îõ©ûþòî
­ƒªò
```

After:

```
你好，世界
🦕
你好，世界
🦕
```

Closes #6001
Closes #14516
Closes #14545

### DIFF
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -207,6 +207,20 @@ fn pty_assign_global_this() {
 }
 
 #[test]
+fn pty_emoji() {
+  // windows was having issues displaying this
+  util::with_pty(&["repl"], |mut console| {
+    console.write_line("console.log('🦕');");
+    console.write_line("close();");
+
+    let output = console.read_all_output();
+    // one for input, one for output
+    let emoji_count = output.chars().filter(|c| *c == '🦕').count();
+    assert_eq!(emoji_count, 2);
+  });
+}
+
+#[test]
 fn console_log() {
   let (out, err) = util::run_and_collect_output(
     true,
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -164,9 +164,7 @@ fn pty_complete_imports() {
     let output = console.read_all_output();
     assert!(output.contains("Hello World"));
     if cfg!(windows) {
-      if !output.contains("testing output\u{1b}") {
-        panic!("Output did not contain expected:\n{:?}", output);
-      }
+      assert!(output.contains("testing output\u{1b}"));
     } else {
       assert!(output.contains("\ntesting output"));
     }
--- a/cli/tests/integration/repl_tests.rs
+++ b/cli/tests/integration/repl_tests.rs
@@ -164,7 +164,9 @@ fn pty_complete_imports() {
     let output = console.read_all_output();
     assert!(output.contains("Hello World"));
     if cfg!(windows) {
-      assert!(output.contains("testing output\u{1b}"));
+      if !output.contains("testing output\u{1b}") {
+        panic!("Output did not contain expected:\n{:?}", output);
+      }
     } else {
       assert!(output.contains("\ntesting output"));
     }
--- a/runtime/ops/io.rs
+++ b/runtime/ops/io.rs
@@ -1,10 +1,8 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-use deno_core::error::bad_resource_id;
-use deno_core::error::not_supported;
-use deno_core::error::resource_unavailable;
 use deno_core::error::AnyError;
 use deno_core::op;
+use deno_core::parking_lot::Mutex;
 use deno_core::AsyncMutFuture;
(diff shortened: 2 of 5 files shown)


dsherret:
I tried to add a test for this, but whatever I tried passed before and after (I think probably because our test utilitity collects data with a pipe and so it's not a console and doesn't get this special behaviour applied)

runtime/ops/io.rs
```diff
@@ -301,33 +293,119 @@ impl Resource for ChildStderrResource {
...
+#[derive(Clone)]
+enum StdFileResourceInner {
+  // Ideally we would store stdio as an StdFile, but we get some Windows
+  // specific functionality for free by using Rust std's wrappers. So we
+  // take a bit of a complexity hit here in order to not have to duplicate
+  // the functionality in Rust's std/src/sys/windows/stdio.rs
+  Stdin,
+  Stdout,
+  Stderr,
+  File(Arc<Mutex<StdFile>>),
```
dsherret, line 305: Question: why do we store this in a mutex? Can't we just store `Arc<StdFile>` and clone it out when we need it or does this need exclusive access?

runtime/ops/io.rs
```diff
@@ -50,14 +50,6 @@ static STDOUT_HANDLE: Lazy<StdFile> =
...
   Lazy::new(|| unsafe { StdFile::from_raw_fd(2) });
 
-/// Due to portability issues on Windows handle to stdout is created from raw
-/// file descriptor.  The caveat of that approach is fact that when this
-/// handle is dropped underlying file descriptor is closed - that is highly
-/// not desirable in case of stdout.  That's why we store this global handle
-/// that is then cloned when obtaining stdio for process. In turn when
-/// resource table is dropped storing reference to that handle, the handle
-/// itself won't be closed (so Deno.core.print) will still work.
-// TODO(ry) It should be possible to close stdout.
```
ry, line 60: regarding the TODO is it possible to close stdout after this patch?

runtime/ops/io.rs
ry: I don't know off hand. I guess you'd have to remove it and see what happens.

ry (approved):
LGTM

runtime/ops/io.rs
dsherret: No, not possible still since they're still held globally here (and not stored in a `Mutex<Option<T>>`).

I removed the todo because I didn't think it was worth keeping it in the code since I think doing this will arise from the need to do it.

piscisaureus (approved):

runtime/ops/io.rs
dsherret: Discussed with Bert and the reason for this is because Read & Write both require `&mut` so storing this in a mutex is faster than duplicating the handle which also could fail.

cli/tests/integration/repl_tests.rs
```diff
@@ -206,6 +206,20 @@ fn pty_assign_global_this() {
...
+  // windows was having issues displaying this
+  util::with_pty(&["repl"], |mut console| {
+    console.write_line("console.log('🦕');");
+    console.write_line("close();");
+
+    let output = console.read_all_output();
+    // one for input, one for output
+    let emoji_count = output.chars().filter(|c| *c == '🦕').count();
+    assert_eq!(emoji_count, 2);
+  });
```
bartlomieju, line 220: Nice 👍

PR #14567, opened 2022-05-11
fix(task): accept double hyphen arg immediately following task name

### LINKED ISSUE #14514: deno task error passing values
Hi, I have the following task in `deno.json`:

```json
{
  "tasks": {
    "build": "deno run -A https://deno.land/x/lume@v1.8.0/ci.ts"
  }
}
```

So I can run `deno task build` to build a website with Lume.

I want to pass options to the build task, for example `-s` (to start a local server): `deno task build -s` works fine.

But if I pass a value to change the output url with `deno task build --location="https://example.com"` I get the following error:
```
error: Found argument '--location' which wasn't expected, or isn't valid in this context

        If you tried to supply `--location` as a value rather than a flag, use `-- --location`

USAGE:
    deno task <task>

```

If the argument `-s` is passed before `--location`:
```
deno task build -s --location="https://example.com"
```
it works fine, but if it's passed after:
```
deno task build --location="https://example.com" -s
```
I get that error. Maybe there's an issue parsing the double dash `--`.

dsherret:
This is unfortunate, but provides a temporary workaround for a bug in clap.

Closes #14514

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -1448,14 +1448,15 @@ fn task_subcommand<'a>() -> Command<'a> {
   Command::new("task")
     .trailing_var_arg(true)
     .arg(config_arg())
-    .arg(Arg::new("task").help("Task to be executed"))
-    .arg(
-      Arg::new("task_args")
+    // Ideally the task name and trailing arguments should be two separate clap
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

Issue #10995, opened 2021-06-15
Performance global is not an EventTarget


lucacasonato:

MierenManz:
this is just as simple as doing `class Performance extends EventTarget` right?
Cuz then I can easily make a pr for it

bartlomieju:
@MierenManz sorry for late response. Yes, that's right. We need some tests as well that ensure you can dispatch events on `Performance`.

PR #14548, opened 2022-05-09
feat(ext/web): add performance.toJSON

GJZwiers:
Closes #10838. Needs https://github.com/denoland/deno_std/pull/2209

PR #14573, opened 2022-05-12
fix: make Performance global an EventTarget

cjihrig:
This commit updates the `Performance` global to extend `EventTarget`.

Fixes: https://github.com/denoland/deno/issues/10995

The CI should fail for the same reason (and with the same test failures) as #14548.


crowlKats:
The amin reason wh we didnt do this yet is because the `resourcetimingbufferfull` event should be implemented, though i guess just doing this should be fine

crowlKats (approved):

PR #14574, opened 2022-05-12
feat(ops): allow passing scope handle to sync ops

littledivy:
To be able to leverage `serde_v8::Value` APIs (as required in #14556)


bartlomieju (approved):
LGTM, would be great if you added a test for this change

littledivy:
> LGTM, would be great if you added a test for this change

Opened an issue to add compile tests for the proc macro https://github.com/denoland/deno/issues/14577

PR #14582, opened 2022-05-12
experiment(ops): #[op(v8)]

AaronO:
Explicitly tag v8 ops, would allow many/most "bindings" to be converted to `#[op(v8)]`

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -17,6 +17,7 @@ pub struct OpDecl {
   pub enabled: bool,
   pub is_async: bool, // TODO(@AaronO): enum sync/async/fast ?
   pub is_unstable: bool,
+  pub is_v8: bool,
 }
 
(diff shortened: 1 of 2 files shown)


core/extensions.rs
```diff
@@ -17,6 +17,7 @@ pub struct OpDecl {
   pub enabled: bool,
   pub is_async: bool, // TODO(@AaronO): enum sync/async/fast ?
   pub is_unstable: bool,
+  pub is_v8: bool,
```
bartlomieju, line 20: `is_async`/`is_v8` could actually be an enum - these are mutually exclusive

AaronO: Yeah, it's hinted at in the above TODO, will look into enforcing it once I land fast ops which I've mostly prototyped locally.

PR #14583, opened 2022-05-12
feat(lsp): enable linting by default

bartlomieju:
Ref https://github.com/denoland/vscode_deno/pull/665

### DIFF
--- a/cli/lsp/config.rs
+++ b/cli/lsp/config.rs
@@ -165,6 +165,10 @@ impl Default for TestingSettings {
   }
 }
 
+fn default_to_true() -> bool {
+  true
+}
+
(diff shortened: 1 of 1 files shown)


lucacasonato:
Shouldn't the default in `vscode_deno` be updated instead?

bartlomieju:
> Shouldn't the default in `vscode_deno` be updated instead?

We need to update this in both places, as we want to change extension to not write to `settings.json` if the selected setting is default value.

kitsonk (approved):
LGTM

Yes it needs to be set in vscode_deno, which is more important.  This is only important when a client doesn't provide a configuration value.

PR #14585, opened 2022-05-12
feat(ops): infallible / result-free ops

AaronO:
Allows ops to be `-> T` instead of `-> Result<T, AnyError>` if they are indeed infallible

NOTE: removed Result returns from infallible ops in `ops_builtin.rs`, but there are likely more in the codebase

### DIFF
--- a/ops/lib.rs
+++ b/ops/lib.rs
@@ -142,6 +142,11 @@ fn codegen_v8_async(core: &TokenStream2, f: &syn::ItemFn) -> TokenStream2 {
   let (arg_decls, args_tail) = codegen_args(core, f, rust_i0, 1);
   let type_params = &f.sig.generics.params;
 
+  let result_wrapper = match is_result(&f.sig.output) {
+    true => quote! {},
+    false => quote! { let result = Ok(result); },
+  };
+
   quote! {
     use #core::futures::FutureExt;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):

ops/lib.rs
```diff
@@ -308,7 +314,14 @@ fn codegen_sync_ret(
     }
   };
 
+  let result_wrapper = match is_result(&**ret_type) {
```
aapoalas, line 317: While this does allow ops to avoid returning `Ok(ret)` in favour of just `ret`, it seems to me like this leaves a lot of dead code lying around.

If I'm not badly mistaken, this should be enough for a non-result type:
```rs
if !is_result(&**ret_type) {
  if is_unit(&**ref_type) {
    return quote! {};
  }
  return ok_block;
}
```
Might be that some customisation is required, but basically that should be enough I think. In the actual generated Rust it'd then be something like this:

```rs
fn op_void_decl(scope, args, rv) {
  let ret: () = op_void();
  // Nothing to do
}

fn op_not_void_decl(scope, args, rv) {
  let not_void: bool = op_not_void();
  match deno_core::serde_v8::to_v8(scope, v) {
    Ok(ret) => rv.set(ret),
    Err(err) => ...
  };
}
```

AaronO: Resolved by #14587

PR #14587, opened 2022-05-12
cleanup(ops): Remove dead code from codegen ops

aapoalas:
Minor further improvement on #14586: There's no need to wrap `result` into `Ok(result)` just to match against `Result`.

Additionally, there's no need to create `let ret = ()` for return value-less ops.


AaronO (approved):
LGTM

Issue #14593, opened 2022-05-13
EventListener does not throw with "null" signal


lucacasonato:
There is a failing WPT in dom/events/AddEventListenerOptions-signal.any.html

Issue #14254, opened 2022-04-11
EventTarget#addEventListener mishandles nullish "signal" option


Garcia101:
Currently, Deno accepts this example snippet without error:
```js
window.addEventListener("load", null, {
 signal: null
});
```
According to <https://dom.spec.whatwg.org/#eventtarget>, EventTarget#addEventListener is defined as
```
interface EventTarget {
 ...
 undefined addEventListener(..., ..., optional (AddEventListenerOptions or boolean) options = {});
 ...
}

dictionary EventListenerOptions {
  boolean capture = false;
};

dictionary AddEventListenerOptions : EventListenerOptions {
  boolean passive = false;
  boolean once = false;
  AbortSignal signal;
};
```
Note the `AbortSignal signal`, which doesn't allow the `signal` value to be null, if the object property is present.

From a quick skim, it seems to be that this branch isn't being reached, even though it's intended to be?
https://github.com/denoland/deno/blob/8ae17026cbb30ab23fd79ac5cf5c207af5ad90a3/ext/web/02_event.js#L932-L934

randomicon00:
From the spec, it says that the signal property can be null.

`signal (null or an [AbortSignal] object)`

F3n67u:
> From the spec, it states that the signal property can be null.
> 
> `signal (null or an [AbortSignal] object)`

@randomicon00  I think null is not the javascript `null` value, it's more like undefined on spec text.


there is a wpt test which states a `TypeError` should throw when `signal` is `null`.

https://github.com/web-platform-tests/wpt/blob/3532a4b4082fbbb32438ae8670252e664aaa2efe/dom/events/AddEventListenerOptions-signal.any.js#L135-L138

Garcia101:
> From the spec, it states that the signal property can be null.
> 
> `signal (null or an [AbortSignal] object)`

If anything, the code snippet that I provide yields different results than Deno when evaluated in a browser.
I expect the null is internal, and intended to mark handlers that aren't storing an AbortSignal reference.

F3n67u:
@Garcia101 what's your deno version? "AddEventListenerOptions.signal shouldn't be nullable" fix is landed at version v1.12.0， so if you using version lower than version 1.12.0, then the problem will exist.

I am using version 1.20.6, the problem is not exist

```bash
$ cat test.js                                          
const et = new EventTarget();
et.addEventListener("foo", () => {}, { signal: null });

$ deno run test.js
error: Uncaught TypeError: signal must be non-null
[...]

PR #14601, opened 2022-05-14
fix(ext/web): throw if listener and signal are null

cjihrig:
This commit fixes a failing WPT test by making `EventTarget`'s `addEventListener()` method throw if both the `listener` and the `signal` option are `null`.

Fixes: https://github.com/denoland/deno/issues/14593
Fixes: https://github.com/denoland/deno/issues/14254

### DIFF
--- a/ext/web/02_event.js
+++ b/ext/web/02_event.js
@@ -944,6 +944,18 @@
       if (!(ReflectHas(listeners, type))) {
         listeners[type] = [];
       }
+
+      for (const listener of listeners[type]) {
+        if (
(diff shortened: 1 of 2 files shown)


crowlKats:
shouldnt we be rather using a webidl converter?

lucacasonato:
> shouldnt we be rather using a webidl converter?

In the limit, yes. This is the only web api that we haven't converted over to use webidl handlers yet.

cjihrig:
I've updated this to use a webidl converter. If this lands, I'll work on porting the rest of `EventTarget` over as well.

PR #14603, opened 2022-05-14
fix(test): Format thrown non-errors properly

### LINKED ISSUE #14596: `Deno.test` does not render POJO errors nicely
```js
Deno.test("foo", async () => {
  throw { foo: "bar" };
})
```

prints as the error message:

```
./test.ts > foo
#<Object>
```

`#<Object>` is not a helpful serialization of the thrown value.

nayeemrmn:
Fixes #14596.

I realised we also have this problem with exceptions in general. This only addresses it for testing. It would be good if we always applied nice formatting for `JsError::exception_message` (which is a different approach from this PR) but that would involve either moving inspect logic to core or the usual of adding a callback to core runtime options. But that callback would need to make a JS call to `Deno[Deno.internal].inspectArgs([error])`. I guess the callback could be registered on the JS-side, making it a lot easier? I have to look into it.

### DIFF
--- a/ext/console/02_console.js
+++ b/ext/console/02_console.js
@@ -326,7 +326,7 @@
   }
 
   function maybeColor(fn, inspectOptions) {
-    return inspectOptions.colors ? fn : (s) => s;
+    return inspectOptions.colors && !colors.getNoColor() ? fn : (s) => s;
   }
 
   function inspectFunction(value, level, inspectOptions) {
@@ -2322,5 +2322,6 @@
     inspect,
     wrapConsole,
     createFilteredInspectProxy,
+    quoteString,
   };
 })(this);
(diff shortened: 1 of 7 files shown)


bartlomieju:
> It would be good if we always applied nice formatting for JsError::exception_message (which is a different approach from this PR) but that would involve either moving inspect logic to core or the usual of adding a callback to core runtime options. But that callback would need to make a JS call to Deno[Deno.internal].inspectArgs([error]). I guess the callback could be registered on the JS-side, making it a lot easier? I have to look into it.

Are you talking about moving console inspection logic to `deno_core`?

ext/console/02_console.js
```diff
@@ -326,7 +326,7 @@
   }
 
   function maybeColor(fn, inspectOptions) {
-    return inspectOptions.colors ? fn : (s) => s;
+    return inspectOptions.colors && !colors.getNoColor() ? fn : (s) => s;
```
bartlomieju, line 329: I believe this might have been on purpose, to always be able to force using colors, though I might be wrong...

nayeemrmn:
> I guess the callback could be registered on the JS-side, making it a lot easier?

I got this to work! It applies to all exceptions as well. Closing in favour of #14604.

PR #14604, opened 2022-05-14
fix: Format non-error exceptions

### LINKED ISSUE #14596: `Deno.test` does not render POJO errors nicely

nayeemrmn:
Fixes #14596.


core/error.rs
```diff
@@ -191,6 +191,20 @@ impl JsError {
...
+    let state = state_rc.borrow();
+    if let Some(format_exception_cb) = &state.js_format_exception_cb {
+      let format_exception_cb = format_exception_cb.open(scope);
+      let this = v8::undefined(scope).into();
+      let formatted = format_exception_cb.call(scope, this, &[exception]);
+      if let Some(formatted) = formatted {
+        if formatted.is_string() {
+          exception_message = Some(formatted.to_rust_string_lossy(scope));
+        }
+      }
```
bartlomieju, line 205: What will be the formatting if the callback is not set?

nayeemrmn: It will hit the `exception_message.unwrap_or_else()`s below to give the current behaviour

Issue #10936, opened 2021-06-11
Inaccurate test coverage might be caused by emojis


eibens:
Hi, I am having an issue where the output of `deno coverage` appears inaccurate. I was able to narrow it down to a minimal example; instructions below. I hope it is reproducible. The problem goes away if the emojis or the unused function `h` are removed. Removing a single emoji removes line 2 from the coverage report, but still inaccurately keeps line 3. I have not tested it with other emojis :relaxed:

This is the `mod.ts` file:

~~~ts
export default function () {
  return "📣❓";
}

function h() {}
~~~

This is the `mod_test.ts` file:

~~~ts
import f from "./mod.ts";

Deno.test("test", () => {
  f()
});
~~~

These are the shell commands:

~~~sh
deno test --coverage=.cov
deno coverage .cov
rm -rf .cov
~~~

This is the output of `deno coverage`. As far as I understand it, line 2 and 3 should be covered:

~~~txt
 cover file:///home/mod.ts ... 25.000% (1/4)
   2 |   return "📣❓";
   3 | }
-----|-----
   5 | function h() {}
~~~

This is the output of `deno --version`:

~~~txt
deno 1.11.0 (release, x86_64-unknown-linux-gnu)
v8 9.1.269.35
typescript 4.3.2
~~~

cdaringe:
I get a rust panic.

```
thread 'main' panicked at 'byte index 1090 is not a char boundary; it is inside '💣' (bytes 1088..1092) of `export class RadError extends Error {
    static message;
    static code;
    static emoji;
    reason;
}
RadError.code = "ERADERROR";
RadError.emoji = "🚨";
export class RadMissingRadFile extends RadError {
}
RadMissingRadFile.message = "unable to find`[...]', cli/tools/coverage.rs:230:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Seems pretty clear that the coverage checker is doing char mapping expecting byte chars, vs graphemes

bartlomieju:
I got a panic as well:
```

============================================================
Deno has panicked. This is a bug in Deno. Please report this
at https://github.com/denoland/deno/issues/new.
If you can reliably reproduce this panic, include the
reproduction steps and re-run with the RUST_BACKTRACE=1 env
var set and include the backtrace in your report.

Platform: macos x86_64
Version: 1.17.2+1fb5858
[...]

PR #14607, opened 2022-05-14
fix: base64 encoding of source maps with emojis

bartlomieju:
Fixes https://github.com/denoland/deno/issues/10936


cli/emit.rs
```diff
@@ -192,7 +192,8 @@ pub fn get_ts_config(
         ts_config.merge(&json!({
           "emitDecoratorMetadata": false,
           "importsNotUsedAsValues": "remove",
-          "inlineSourceMap": true,
+          "inlineSourceMap": false,
+          "sourceMap": true,
```
bartlomieju, line 196: I'm a bit confused - from my understanding this should only have an impact on TSC emit for `deno` subcommand and `Deno.emit()` API shouldn't be altered at all and yet tests that use `Deno.emit()` started failing.

cli/emit.rs
```diff
@@ -463,9 +465,29 @@ pub fn check_and_maybe_emit(
...
       }
     }
-    for emit in response.emitted_files.into_iter() {
+
+    // NOTE(bartlomieju): here we're doing a bit of magic - TSC emits
+    // broken inlined source maps if original source code contains
+    // emojis. To alleviate this problem, we're gonna manually edit
+    // emitted files and append a source map that we'll encode manually.
+    // This is only happening for the emits for "deno" subcommand, the
+    // runtime API (`Deno.emit()`) is excluded from this.
```
dsherret, line 474: Could we identify if this behaviour happens with tsc and if so open a bug in their repo before merging this PR? Then we could open an issue in our repo to eventually remove this code.

bartlomieju: I created a sample TSC project with following config:
```ts
export function hello(): string {
  return "📣❓";
}

```
```
{
    "compilerOptions": {
        "allowJs": true,
        "allowSyntheticDefaultImports": true,
        "emitDecoratorMetadata": false,
        "experimentalDecorators": true,
        "importsNotUsedAsValues": "remove",
        "incremental": true,
        "inlineSourceMap": true,
        "inlineSources": true,
        "isolatedModules": true,
        "jsx": "react",
        "module": "esnext",
        "removeComments": true,
        "sourceMap": false,
        "strict": true,
        "target": "esnext",
        "useDefineForClassFields": true,
        "useUnknownInCatchVariables": false,
    }
}
```

The output is as follows:
```js
export function hello() {
    return "📣❓";
}
//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiZW1vamkuanMiLCJzb3VyY2VSb290IjoiIiwic291cmNlcyI6WyJlbW9qaS50cyJdLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiQUFBQSxNQUFNLFVBQVUsS0FBSztJQUNuQixPQUFPLEtBQUssQ0FBQztBQUNmLENBQUMiLCJzb3VyY2VzQ29udGVudCI6WyJleHBvcnQgZnVuY3Rpb24gaGVsbG8oKTogc3RyaW5nIHtcbiAgcmV0dXJuIFwi8J+To+Kdk1wiO1xufVxuIl19
```
It does contain a proper source map that can be decoded. It looks like this is somehow caused by our implementation of TS host. I wonder if `tsc` available for node uses some kind of custom `base64` encoding function that is different than the function in default compiler host implementation (that we use).

dsherret: Probably. They're doing:

```ts
    /**
     * Converts a string to a base-64 encoded ASCII string.
     */
    function convertToBase64(input) {
        var result = "";
        var charCodes = getExpandedCharCodes(input);
        var i = 0;
        var length = charCodes.length;
        var byte1, byte2, byte3, byte4;
        while (i < length) {
            // Convert every 6-bits in the input 3 character points
            // into a base64 digit
            byte1 = charCodes[i] >> 2;
            byte2 = (charCodes[i] & 3) << 4 | charCodes[i + 1] >> 4;
            byte3 = (charCodes[i + 1] & 15) << 2 | charCodes[i + 2] >> 6;
            byte4 = charCodes[i + 2] & 63;
            // We are out of characters in the input, set the extra
            // digits to 64 (padding character).
            if (i + 1 >= length) {
                byte3 = byte4 = 64;
            }
            else if (i + 2 >= length) {
                byte4 = 64;
            }
            // Write to the output
            result += base64Digits.charAt(byte1) + base64Digits.charAt(byte2) + base64Digits.charAt(byte3) + base64Digits.charAt(byte4);
            i += 3;
        }
        return result;
    }
```

bartlomieju: Oh, good to know, I guess we could just provide our own JS version to fix this problem? Do you happen to now how it's "injected" into the compiler host?

dsherret: You could assign a new function to `ts.base64encode`. Basically, copy this, but use something else than `convertToBase64`:

````
   function base64encode(host, input) {
        if (host && host.base64encode) {
            return host.base64encode(input);
        }
        return convertToBase64(input);
    }
    ts.base64encode = base64encode;
```

dsherret: Opened https://github.com/microsoft/TypeScript/issues/49150

We can just overwrite `ts.convertToBase64` I think.

dsherret (approved):
LGTM!

PR #14610, opened 2022-05-14
fix(ext/tls): ability to ignore IP-address certificate errors

### LINKED ISSUE #14609: Ability to ignore IP-address certificate errors
## Problem

[rustls 0.20.5](https://crates.io/crates/rustls/0.20.5) has been recently released and have landed [this PR](https://github.com/rustls/rustls/pull/1032).

It feels that `--unsafely-ignore-certificate-errors` option could just force Deno to ignore verification of IP-addresses as stated in rustls PR. This means that [this issue](https://github.com/denoland/deno/issues/7660) could be handled at least by unsafe CLI option.

However, bumping rustls is not a complete solution, since after doing that I got `Unknown ServerName variant` error. This is most likely because rustls added `IpAddress` to [ServerName](https://docs.rs/rustls/0.20.5/rustls/enum.ServerName.html) enum while [this part](https://github.com/denoland/deno/blob/c496639d5dc190b107222bc30462d67ddb86c223/ext/tls/lib.rs#L53) of Deno code is not ready for that.

## Potential Solution

Bump rustls to v0.20.5, then update `ext/tls/lib.rs` to recognise IP-addresses.

albnnc:
Fixes #14609.

- rustls was updated to v0.20.5
- `ext/tls/lib.rs` was changed to understand IP-addresses
- TS TLS test of invalid host handling was updated to check truly invalid host

### DIFF
--- a/ext/tls/lib.rs
+++ b/ext/tls/lib.rs
@@ -50,6 +50,9 @@ impl ServerCertVerifier for NoCertificateVerification {
     ocsp_response: &[u8],
     now: SystemTime,
   ) -> Result<ServerCertVerified, Error> {
+    if self.0.is_empty() {
+      return Ok(ServerCertVerified::assertion());
+    }
     let dns_name_or_ip_address = match server_name {
       ServerName::DnsName(dns_name) => dns_name.as_ref().to_owned(),
       ServerName::IpAddress(ip_address) => ip_address.to_string(),
@@ -59,7 +62,7 @@ impl ServerCertVerifier for NoCertificateVerification {
         return Err(Error::General("Unknown `ServerName` variant".to_string()));
       }
     };
-    if self.0.is_empty() || self.0.contains(&dns_name_or_ip_address) {
+    if self.0.contains(&dns_name_or_ip_address) {
       Ok(ServerCertVerified::assertion())
     } else {
       let root_store = create_default_root_cert_store();
--- a/ext/tls/lib.rs
+++ b/ext/tls/lib.rs
@@ -12,10 +12,12 @@ use deno_core::error::AnyError;
 use deno_core::parking_lot::Mutex;
 use deno_core::Extension;
 
+use rustls::client::HandshakeSignatureValid;
 use rustls::client::ServerCertVerified;
 use rustls::client::ServerCertVerifier;
 use rustls::client::StoresClientSessions;
 use rustls::client::WebPkiVerifier;
+use rustls::internal::msgs::handshake::DigitallySignedStruct;
 use rustls::Certificate;
 use rustls::ClientConfig;
 use rustls::Error;
@@ -76,6 +78,24 @@ impl ServerCertVerifier for NoCertificateVerification {
         now,
       )
     }
+  }
+
+  fn verify_tls12_signature(
+    &self,
+    _message: &[u8],
+    _cert: &rustls::Certificate,
+    _dss: &DigitallySignedStruct,
+  ) -> Result<HandshakeSignatureValid, Error> {
+    Ok(HandshakeSignatureValid::assertion())
+  }
+
+  fn verify_tls13_signature(
+    &self,
+    _message: &[u8],
+    _cert: &rustls::Certificate,
+    _dss: &DigitallySignedStruct,
+  ) -> Result<HandshakeSignatureValid, Error> {
+    Ok(HandshakeSignatureValid::assertion())
   }
 }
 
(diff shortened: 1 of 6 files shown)


bartlomieju (dismissed):
Please add a unit test that exercises this change (connecting to an ip address)

albnnc:
> Please add a unit test that exercises this change (connecting to an ip address)

Not sure if I understood the idea correctly, since it seems that `--unsafely-ignore-certificate-errors` option is being tested apart from common SSL unit tests which are located in TS-files. However, I added integration test of IP-address unsafe connectability. Will it be enough?

ext/tls/lib.rs
```diff
@@ -50,26 +50,28 @@ impl ServerCertVerifier for NoCertificateVerification {
...
-        )
+    let dns_name_or_ip_address = match server_name {
+      ServerName::DnsName(dns_name) => dns_name.as_ref().to_owned(),
+      ServerName::IpAddress(ip_address) => ip_address.to_string(),
+      _ => {
+        // NOTE(bartlomieju): `ServerName` is a non-exhaustive enum
+        // so we have this catch all errors here.
+        return Err(Error::General("Unknown `ServerName` variant".to_string()));
       }
+    };
```
lucacasonato, line 61: We should skip computing this when `self.0.is_empty()` (99% case).

albnnc: Good point, I've moved empty case to the top of branching.

lucacasonato (commented):
Looks good, can you add a test for an IPv6 address too? Then we can land this.

albnnc:
Eventually, I added [these lines](https://github.com/seanmonstar/reqwest/blob/d24a5286c7bd101f82c90e39cb2ccc21ae854da8/src/tls.rs#L405) from reqwest to handle a few more cert errors when unsafe option presented.
 
I think we can't map webpki errors, since they are getting [mapped](https://github.com/rustls/rustls/blob/4fc3a3af8cadd4d414560f79b3428c6c93aa11eb/rustls/src/verify.rs#L579) by rustls anyway and, as you said in Discord, seem to occur only if we skipped validation for invalid cert previously.

I failed to add ipv6 tests, since reqwest seems to fail such requests.

<details>
<summary>Example</summary>

Example uses `danger_accept_invalid_certs`, since it's the default way to make completely unsafe client. I think we won't be able to handle things that couldn't be handled with that option. Not without upstream fix at least. 

```rust
use reqwest::blocking::Client;
use reqwest::Method;

fn main() {
  let client = Client::builder()
    .danger_accept_invalid_certs(true)
    .build()
    .unwrap();
  let resp = client
    .request(Method::GET, "https://[::1]:3000")
    .send()
    .unwrap();
  println!("resp {:?}", resp.status());
}
```

- The code listed above fails with error.
- Target address (`https://[::1]:3000`) exists and available in browser and postman.
- The same code with request to `https://localhost:3000`, `https://127.0.0.1:3000` and `http://[::1]:3000` works. 

The trace is

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv6(::1)), port: Some(3000), path: "/", query: None, fragment: None }, source: hyper::Error(Connect, Custom { kind: Other, error: "invalid dnsname" }) }', main.rs:12:6
```

</details>

I think that for now we can just handle ipv4 addresses – this is by far the most common case and it works only with unsafe cli option anyway. Are we going to be happy with that?

lucacasonato:
> I think we can't map webpki errors, ... [and] as you said in Discord, seem to occur only if we skipped validation for invalid cert previously.

That's not quite what I said. There are two types of relevant WebPKI errors here: ones resulting from invalid signature verification (these are **not** OK to ignore), and ones resulting from invalid certificates (these are OK to ignore).

I think we should selectively ignore `webpki::Error::BadDER` derived errors, or all errors if the "unsafe" flag was passed with no arguments.

albnnc:
> That's not quite what I said. There are two types of relevant WebPKI errors here: ones resulting from invalid signature verification (these are not OK to ignore), and ones resulting from invalid certificates (these are OK to ignore).

Misunderstood, sorry. Now I hopefully got it.

> I think we should selectively ignore webpki::Error::BadDER derived errors, or all errors if the "unsafe" flag was passed with no arguments.

I've updated PR.

lucacasonato (approved):
LGTM, thanks @albnnc

PR #14614, opened 2022-05-15
Revert "core: don't include_str extension js code (#10786)"

### LINKED ISSUE #11030: `deno_core` prints cargo related stuff if used as a dependency

AaronO:
This reverts commit 10e50a120744de71d6915af4ae93f8231607573d

Alternative to #13217, IMO the tradeoffs made by #10786 aren't worth it.

It breaks abstractions (crates being self-contained, deno_core without snapshotting etc...) and causes pain points / gotchas for both embedders & devs for a relatively minimal gain in incremental build time ...

Closes #11030


littledivy (commented):
`webgpu/src` changes seems unrelated. https://github.com/denoland/deno/pull/14614/files#diff-924b4b7acc08e4f385a8158fcef447fc3006dd533da7a6dc6b2039d3ab818e28

AaronO:
> `webgpu/src` changes seems unrelated. https://github.com/denoland/deno/pull/14614/files#diff-924b4b7acc08e4f385a8158fcef447fc3006dd533da7a6dc6b2039d3ab818e28

It's not, we either need to move the files (as I did) or change the include path to add `../`, this happens because when https://github.com/gfx-rs ingested `deno_webgpu` they restored a `src/` folder in the crate (unlike our pattern) but didn't move the JS files into thus breaking our assumption that JS files live alongside rust files

littledivy (approved):
LGTM

andreubotella:
This PR closes #11030.

Issue #14498, opened 2022-05-05
--no-check=remote on by default for all subcommands


ry:
like `deno test`

`--check` to enable

nayeemrmn:
> `--check` to enable

Currently it's:
- `deno run main.ts` - Check all
- `deno run --check main.ts` - Check all
- `deno run --check=all main.ts` - Check all

I think 1.22 was supposed to be:
- `deno run main.ts` - Check with no remote
- `deno run --check main.ts` - Check with no remote
- `deno run --check=all main.ts` - Check all

(+ deprecate `--no-check=remote`)

And at 1.23 or currently with `DENO_FUTURE_CHECK`:
- `deno run main.ts` - No check
- `deno run --check main.ts` - Check with no remote
- `deno run --check=all main.ts` - Check all

(+ deprecate `--no-check`)

Is this correct?

lucacasonato:
@nayeemrmn  that looks right!

PR #14623, opened 2022-05-15
feat: subcommands type-check only local files by default

bartlomieju:
Closes https://github.com/denoland/deno/issues/14498


cli/flags.rs
```diff
@@ -2041,10 +2046,13 @@ fn no_check_arg<'a>() -> Arg<'a> {
...
-    .help("Skip type checking modules")
+    .help("DEPRECATED: Skip type checking modules")
     .long_help(
-      "Skip type-checking. If the value of '--no-check=remote' is supplied, \
-      diagnostic errors from remote modules will be ignored.",
+      "DEPRECATED: Skip type-checking. If the value of '--no-check=remote' is supplied,
+diagnostic errors from remote modules will be ignored.
+
+This option will be removed in Deno 1.23, see `--check` instead.
+      ",
```
ry, line 2055: this seems to be doing more than just "type-check only local files by default"

ry (commented):
I would expect a test that shows, for example, that `deno test` can have type errors in remote modules that do not cause error.

cli/flags.rs
bartlomieju: I don't understand what you mean here

bartlomieju:
@ry do you want tests for all subcommands?

ry:
@bartlomieju which subcommands are effected? `deno test`... ? Yes I think we should probably have tests for each.

bartlomieju:
Following commands are affected by this PR:
- `deno bench`
- `deno bundle`
- `deno cache`
- `deno compile`
- `deno eval`
- `deno install`
- `deno run`
- `deno test`

ry (approved):
LGTM

dsherret (approved):
LGTM

littledivy (approved):

PR #14624, opened 2022-05-15
feat(ext/net): add `CAA` DNS record support

cmorten:
Add `CAA` record type support for `Deno.resolveDns()`.

### DIFF
--- a/ext/net/ops.rs
+++ b/ext/net/ops.rs
@@ -34,6 +34,7 @@ use std::rc::Rc;
 use tokio::net::TcpListener;
 use tokio::net::TcpStream;
 use tokio::net::UdpSocket;
+use trust_dns_proto::rr::rdata::caa::Value;
 use trust_dns_proto::rr::record_data::RData;
 use trust_dns_proto::rr::record_type::RecordType;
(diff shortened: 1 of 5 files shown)


ext/net/ops.rs
```diff
@@ -740,6 +746,38 @@ fn rdata_to_return_record(
...
         .map(DnsReturnRecord::Aname),
+      CAA => r.as_caa().map(|caa| {
+        let value = caa.value();
+
+        DnsReturnRecord::Caa {
+          critical: caa.issuer_critical(),
+          tag: caa.tag().to_string(),
+          value: match value {
+            Value::Issuer(name, key_values) => {
+              let mut s = String::new();
```
cmorten, line 757: TODO: utilise `name.to_string()` and `key_values.to_string()` instead of handrolling.

Unfortunately can’t just use `value.to_string()` as `trust-dns` wraps in unwanted additional double quotes - https://github.com/bluejekyll/trust-dns/blob/95b2dee327ade007bf317ca98bbd3b24c0bdd096/crates/proto/src/rr/rdata/caa.rs#L842

bartlomieju (approved):
LGTM, nice work!

Issue #14432, opened 2022-04-29
Support address to name translation


cmorten:
Feature request to support address to name translation ("reverse DNS"), a.k.a. wrapper around https://man7.org/linux/man-pages/man3/getnameinfo.3.html (e.g. using https://docs.rs/dns-lookup/latest/dns_lookup/#getnameinfo) or equivalent, so that users of Deno can translate an IP and port into a hostname and service.

E.g. 

```ts
interface Service {
  hostname: string;
  service: string;
}

function Deno.resolveService(ip: string, port: number): Promise<Service>;
```

_API shape and names just as example - suspect we can devise better alternatives!_

```ts
const { hostname, service } = await Deno.resolveDns('127.0.0.1', 22);

console.log(hostname, service); // "localhost" "ssh"
```

This would provide a counterpart to the existing [`Deno.resolveDns()`](https://doc.deno.land/deno/stable/~/Deno.resolveDns) API which offers "the reverse" name to address resolution currently.

This would increase parity with existing Node APIs:

- https://nodejs.org/api/dns.html#dnslookupserviceaddress-port-callback
- https://nodejs.org/api/dns.html#dnspromiseslookupserviceaddress-port

And is required to support the `dns` and `dns/promises` modules in the Deno std library compat layer.

cmorten:
Semi working PoC - https://github.com/denoland/deno/compare/main...cmorten:feat/support-reverse-dns-api

(First rust code... so yeah, probably needs work 😅)

cmorten:
Example implementation in https://github.com/denoland/deno/pull/14634

bartlomieju:
This is covered now by `node:dns` / `node:dns/promises` `lookupService`, which is the `getnameinfo` wrapper this issue asked for:

```js
import dnsp from "node:dns/promises";

const { hostname, service } = await dnsp.lookupService("127.0.0.1", 22);
console.log(hostname, service); // localhost ssh
```

That returns the exact `{ hostname, service }` shape and result the request described, and the callback form (`dns.lookupService`) works too. The original motivation, supporting the `dns`/`dns/promises` compat modules, is therefore met.

We don't need a dedicated `Deno.resolveNameInfo()` for this when `node:dns.lookupService` already provides it, so closing as resolved.

PR #14634, opened 2022-05-16
feat(ext/net): New unstable `Deno.resolveNameInfo()` API

cmorten:
Proposal for resolving #14432 

Provides a `Deno` namespace API for wrapping `getnameinfo` for hostname and service resolution from ip and port.


bartlomieju (changes requested):
We discussed this PR during latest design meeting. The team believes that it should be an async API, ie. the actual API call should happen in side a threadpool.

There's also a question regarding return value, @cmorten can you give some context why you went with this interface? Is it required for Node compatbility?

cmorten:
> We discussed this PR during latest design meeting. The team believes that it should be an async API, ie. the actual API call should happen in side a threadpool.
> 
> There's also a question regarding return value, @cmorten can you give some context why you went with this interface? Is it required for Node compatbility?

Agree, think should mirror the interface of resolveDns if implemented, be async etc.

Please regard this PR as more of a PoC than anything, wouldn’t expect this to be the final implementation - it was just to add some colour to the issue (find easier to critique when there’s something to contrast against!).

Primary reason for suggestion is for Node compat (e.g. https://nodejs.org/api/dns.html#dnslookupserviceaddress-port-callback, covered in the issue). If team feel it’s not a priority / something worth covering in compat then happy to close the issue out 🙂

bartlomieju:
> Please regard this PR as more of a PoC than anything, wouldn’t expect this to be the final implementation - it was just to add some colour to the issue (find easier to critique when there’s something to contrast against!). Sadly not quite there with rust to be able to offer the best / most advised implementation 😅

Agreed, it was good to have a PoC implementation to discuss it.

> Primary reason for suggestion is for Node compat (e.g. https://nodejs.org/api/dns.html#dnslookupserviceaddress-port-callback, covered in the associated issue). If team feel it’s not a priority / something worth covering in compat then happy to close the issue out 🙂

Makes sense! I'm glad you were looking into that. I see that return type matches what Node does so that's fine for me. If the API became async I think we could land it then.

Issue #12666, opened 2021-11-07
EventTarget's methods don't validate [this] arguments


Phosra:
Currently, `EventTarget`'s prototype methods don't check the type (prototype) of the `this` argument passed to them.

E.g. given this code,
```js
const this_ = {};
const event_name = "";
const event_handler = null;
const options = {};
EventTarget.prototype.addEventListener.call(this_, event_name, event_handler, options);
```
Firefox and Chrome, respectively throw the following:
```
TypeError: 'addEventListener' called on an object that does not implement interface EventTarget.
```
```
TypeError: Illegal invocation
```

In contrast, on the current canary (1.15.3+26a5471), it does nothing, and returns `undefined`.

lucacasonato:
I have a very old branch somewhere that fixes this. I'll assign it to myself - might get this fixed for 1.16.1.

Garcia101:
> I have a very old branch somewhere that fixes this. I'll assign it to myself - might get this fixed for 1.16.1.

Any update on this?
I've since accidentally encountered another bug related to the code in question: #14254.

PR #14637, opened 2022-05-17
fix(ext/web): add EventTarget brand checking

cjihrig:
This commit adds brand checking to `EventTarget`. It also fixes a bug where deno would crash if an abort signal was aborted via the global `addEventListener()`.

Fixes: https://github.com/denoland/deno/issues/12666

### DIFF
--- a/ext/web/02_event.js
+++ b/ext/web/02_event.js
@@ -917,6 +917,7 @@
   class EventTarget {
     constructor() {
       this[eventTargetData] = getDefaultTargetData();
+      this[webidl.brand] = webidl.brand;
     }
 
     addEventListener(
@@ -924,6 +925,8 @@
       callback,
       options,
     ) {
+      const self = this ?? globalThis;
+      webidl.assertBranded(self, EventTargetPrototype);
       const prefix = "Failed to execute 'addEventListener' on 'EventTarget'";
 
       webidl.requiredArguments(arguments.length, 2, {
@@ -939,7 +942,7 @@
         return;
       }
 
-      const { listeners } = (this ?? globalThis)[eventTargetData];
+      const { listeners } = self[eventTargetData];
 
       if (!(ReflectHas(listeners, type))) {
         listeners[type] = [];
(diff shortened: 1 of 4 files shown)


kt3k:
The code change looks good to me. How about adding the example in #12666 as a test case?

ext/web/02_event.js
```diff
@@ -917,13 +917,16 @@
...
       this[eventTargetData] = getDefaultTargetData();
+      this[webidl.brand] = webidl.brand;
     }
 
     addEventListener(
       type,
       callback,
       options,
     ) {
+      const self = this ?? globalThis;
```
lucacasonato, line 928: I have a feeling like the global `addEventListener` methods are not bound correctly if this is necessary.

kt3k (approved):
LGTM

cjihrig:
This is currently blocked by a failing WPT (webidl/ecmascript-binding/global-immutable-prototype.any.html) that modifies the `window` object's prototype. This causes the brand checking added in this PR to fail.

kt3k:
> This is currently blocked by a failing WPT (webidl/ecmascript-binding/global-immutable-prototype.any.html) that modifies the window object's prototype.

Oh, that's tricky..

I can think of 2 ways to fix it:
- make global prototype chain immutable (as tested in that test case)
- store global prototype chain somewhere and restore them before `Deno.exit` is called in `tools/wpt/testharnessreport.js`

cjihrig:
I opened https://github.com/denoland/deno/pull/14659 to try freezing the prototype chain. I didn't want to bake the fix into `tools/wpt/testharnessreport.js` because the problem would still exist elsewhere.

cjihrig:
@lucacasonato and @kt3k - I paired with @crowlKats today on https://github.com/denoland/deno/pull/14637/commits/9d25d5d44bb4da696f2220d1c5ae489b18627d73. Since we might not be able to freeze the `globalThis` prototype anytime real soon, we addressed the issue with the test harness. Does this still look good to you?

kt3k (approved):
> Since we might not be able to freeze the globalThis prototype anytime real soon, we addressed the issue with the test harness.

This makes sense to me.

LGTM

PR #14640, opened 2022-05-17
feat(ops): 'hybrid' ops - sync returning future

littledivy:
Support for 'hybrid' (sync + async) ops. Ops can have a sync phase to do things, like creating V8 handles and using values that aren't movable to the async block.

```rust
#[op]
fn op_sum_async(nums: Vec<f64>) -> impl Future<Output = f64> + 'static {
  // Sum inputs
  let sum = nums.iter().fold(0.0, |a, v| a + v);
  async move { sum }
}
```


aapoalas:
One thing that this reminds me is the pain of these sorts of JS functions:
```ts
function hybrid_func(nums) {
  const stringifiedNumbers = nums.toString(); // Throws if nums happens to be null
  return fetch(getSomeAddress(stringifiedNumbers));
}

hybrid_func(null).catch(err => {}); // .catch() won't work because the function throws synchronously
```

To put this in context with this PR: The only way these hybrid functions can do synchronous fallible work is by using `.unwrap()`. If these ops are intended to be equally powerful, then the return value should support something like `Result<impl Future<Output = T> + 'static, Error>` if that is possible. This will make the ops macro quite a bit more complicated.

bartlomieju:
@littledivy what is the use case for this PR? AFAIK ops are already eagerly polled, so you should be able to run some code in the same tick of the event loop, until await point is encountered.

bartlomieju (approved):

PR #14641, opened 2022-05-17
fix(ext/http): skip auto-compression if content-encoding present

AaronO:
Regression from #14552


lucacasonato (approved):

lucacasonato:
Actually, can you add a test so this does not regress again in the future?

lucacasonato (approved):
Thanks

PR #14642, opened 2022-05-17
fix(ext/http): error on invalid headers

AaronO:
Minor regression/change-in-behaviour from #14552 that filtered out invalid http headers in rust vs error-ing back to JS


lucacasonato (approved):
LGTM

PR #14643, opened 2022-05-17
 feat: return a signal string instead number on ChildStatus 

### LINKED ISSUE #14491: Inconsistency in API between `Child.kill()` and `ChildStatus.signal`

crowlKats:
Closes #14491

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1459,7 +1459,7 @@ declare namespace Deno {
     | {
       success: false;
       code: number;
-      signal: number | null;
+      signal: string | null;
     };
 
   export interface SpawnOutput<T extends SpawnOptions> {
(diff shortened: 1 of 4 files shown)


kt3k (dismissed):
LGTM. Nice!

kt3k:
Just saw @AaronO's comment in the chat room. Looks like we're not in agreement in general direction?

kt3k:
BTW node.js looks returning string for this https://nodejs.org/api/child_process.html#subprocesssignalcode

Node.js accepts both string and number as input of `kill()` https://nodejs.org/api/child_process.html#subprocesskillsignal

crowlKats:
> Just saw @AaronO's comment in the chat room. Looks like we're not in agreement in general direction?

well, during the meeting there was a general consensus, especially because of the node compat

kt3k (approved):
Ok. Then LGTM 👍

PR #14645, opened 2022-05-17
feat: stabilize deno.spawn API

crowlKats:
Waiting on #14643


lucacasonato:
If we change API surface in a release, we shouldn't stabilize in that same release IMO (https://github.com/denoland/deno/pull/14643). Stabilization should occur if we have gone through one release cycle without encountering any API issues.

crowlKats:
Closing for now, as we have decided to do this next release

Issue #14649, opened 2022-05-17
window prototype chain should be frozen


cjihrig:
The global-immutable-prototype.any WPT currently fails because Deno does not freeze the window prototype chain. This makes it difficult to [add webidl brand checking to `EventTarget`](https://github.com/denoland/deno/pull/14637) because `window` should be an `EventTarget`. This also makes it more difficult to [harden certain APIs against tampering](https://github.com/denoland/deno/issues/14648).

PR #14659, opened 2022-05-18
fix(runtime): freeze globalThis prototype chain

cjihrig:
The `globalThis` prototype chain should be frozen. There is a WPT for this behavior. This is also desirable in order to add webidl brand checking to `EventTarget`.

Fixes: https://github.com/denoland/deno/issues/14649
Refs: https://github.com/denoland/deno/pull/14637

### DIFF
--- a/runtime/js/99_main.js
+++ b/runtime/js/99_main.js
@@ -17,9 +17,11 @@ delete Object.prototype.__proto__;
     ObjectDefineProperty,
     ObjectDefineProperties,
     ObjectFreeze,
+    ObjectGetPrototypeOf,
     ObjectPrototypeIsPrototypeOf,
     ObjectSetPrototypeOf,
     PromiseResolve,
+    SafeSet,
     Symbol,
     SymbolFor,
     SymbolIterator,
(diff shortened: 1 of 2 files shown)


kt3k:
This solution looks fine to me, but V8 seems having `is_immutable_proto` bit field dedicated for this purpose https://cs.github.com/v8/v8/blob/a9f802859bc31e57037b7c293ce8008542ca03d8/src/objects/map.h?q=repo%3Av8%2Fv8+Immutable+Proto+bits#L162 and chromium seems using it (according to the error message)

Do we need to follow it? cc @lucacasonato @bartlomieju

cjihrig:
@kt3k I think to do this in Rust, we need access to `SetImmutableProto()`. I opened https://github.com/denoland/rusty_v8/pull/977 to add a binding for it. This is blocked until a new rusty_v8 is available with that binding.

bartlomieju:
@cjihrig you should now be unblocked on this PR

cjihrig:
@Phosra it does still need to be done, but not like this. The work should ideally happen in Rust, but I think it will take a large refactor to implement.

PR #14663, opened 2022-05-18
feat(ffi): Explicit callbacks

aapoalas:
Tbd


aapoalas:
Benchmarks:

```
> deno bench --unstable --allow-ffi --allow-read test_ffi/tests/bench.js
benchmark                              time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------- -----------------------------
nop()                               601.4 ns/iter (542.62 ns … 734.39 ns) 608.11 ns 734.39 ns 734.39 ns
nop_u8()                           862.84 ns/iter   (791.34 ns … 1.03 µs) 882.96 ns   1.03 µs   1.03 µs
nop_i8()                           859.52 ns/iter (820.53 ns … 924.55 ns) 877.61 ns 924.55 ns 924.55 ns
nop_u16()                          858.13 ns/iter (790.73 ns … 982.01 ns) 877.41 ns 982.01 ns 982.01 ns
nop_i16()                          860.38 ns/iter       (793.1 ns … 1 µs) 873.15 ns      1 µs      1 µs
nop_u32()                          879.65 ns/iter  (844.15 ns … 977.7 ns)  902.1 ns  977.7 ns  977.7 ns
nop_i32()                             875 ns/iter (824.41 ns … 953.11 ns) 887.28 ns 953.11 ns 953.11 ns
nop_u64()                           878.1 ns/iter (839.11 ns … 982.81 ns) 886.43 ns 982.81 ns 982.81 ns
nop_i64()                          844.61 ns/iter (790.26 ns … 921.08 ns) 855.65 ns 921.08 ns 921.08 ns
nop_usize()                         845.8 ns/iter (824.03 ns … 967.26 ns) 849.53 ns 967.26 ns 967.26 ns
nop_isize()                        846.07 ns/iter (810.77 ns … 985.96 ns) 853.98 ns 985.96 ns 985.96 ns
nop_f32()                          858.54 ns/iter  (812.26 ns … 966.1 ns) 866.04 ns  966.1 ns  966.1 ns
nop_f64()                          857.69 ns/iter (810.77 ns … 959.37 ns) 867.07 ns 959.37 ns 959.37 ns
nop_buffer()                         1.09 µs/iter   (999.69 ns … 1.23 µs)   1.12 µs   1.23 µs   1.23 µs
return_u8()                        910.34 ns/iter   (875.06 ns … 1.06 µs)  919.9 ns   1.06 µs   1.06 µs
return_i8()                         887.8 ns/iter   (835.07 ns … 1.15 µs) 894.66 ns   1.15 µs   1.15 µs
return_u16()                       871.13 ns/iter   (830.61 ns … 1.07 µs) 874.69 ns   1.07 µs   1.07 µs
return_i16()                       877.43 ns/iter   (803.81 ns … 1.11 µs) 899.52 ns   1.11 µs   1.11 µs
return_u32()                          691 ns/iter (654.08 ns … 973.55 ns) 698.52 ns 973.55 ns 973.55 ns
return_i32()                       682.13 ns/iter (656.45 ns … 805.61 ns) 687.23 ns 805.61 ns 805.61 ns
return_u64()                       688.34 ns/iter (659.23 ns … 763.29 ns)  694.8 ns 763.29 ns 763.29 ns
return_i64()                       680.85 ns/iter (622.17 ns … 766.63 ns) 689.65 ns 766.63 ns 766.63 ns
return_usize()                     688.65 ns/iter   (627.74 ns … 1.02 µs) 689.36 ns   1.02 µs   1.02 µs
return_isize()                     674.71 ns/iter (654.17 ns … 758.42 ns) 680.98 ns 758.42 ns 758.42 ns
return_f32()                       676.53 ns/iter (655.17 ns … 776.51 ns) 683.17 ns 776.51 ns 776.51 ns
return_f64()                       678.97 ns/iter (658.08 ns … 743.51 ns) 684.24 ns 743.51 ns 743.51 ns
return_buffer()                    985.09 ns/iter   (929.87 ns … 1.16 µs) 992.66 ns   1.16 µs   1.16 µs
nop_nonblocking()                  897.27 ns/iter   (817.11 ns … 1.22 µs) 899.49 ns   1.22 µs   1.22 µs
nop_u8_nonblocking()                 1.13 µs/iter     (1.08 µs … 1.33 µs)   1.14 µs   1.33 µs   1.33 µs
nop_i8_nonblocking()                 1.12 µs/iter     (1.07 µs … 1.35 µs)   1.15 µs   1.35 µs   1.35 µs
nop_u16_nonblocking()                1.11 µs/iter     (1.06 µs … 1.23 µs)   1.12 µs   1.23 µs   1.23 µs
nop_i16_nonblocking()                 1.1 µs/iter     (1.07 µs … 1.29 µs)   1.11 µs   1.29 µs   1.29 µs
nop_u32_nonblocking()                1.11 µs/iter     (1.05 µs … 1.26 µs)   1.13 µs   1.26 µs   1.26 µs
nop_i32_nonblocking()                1.12 µs/iter     (1.07 µs … 1.27 µs)   1.13 µs   1.27 µs   1.27 µs
nop_u64_nonblocking()                1.11 µs/iter      (1.07 µs … 1.3 µs)   1.11 µs    1.3 µs    1.3 µs
nop_i64_nonblocking()                1.15 µs/iter     (1.08 µs … 1.44 µs)   1.16 µs   1.44 µs   1.44 µs
nop_usize_nonblocking()              1.15 µs/iter      (1.09 µs … 1.4 µs)   1.15 µs    1.4 µs    1.4 µs
nop_isize_nonblocking()              1.22 µs/iter     (1.11 µs … 1.86 µs)   1.25 µs   1.86 µs   1.86 µs
nop_f32_nonblocking()                1.33 µs/iter     (1.09 µs … 2.16 µs)   1.43 µs   2.16 µs   2.16 µs
nop_f64_nonblocking()                1.24 µs/iter     (1.08 µs … 2.33 µs)   1.23 µs   2.33 µs   2.33 µs
nop_buffer_nonblocking()             1.37 µs/iter      (1.3 µs … 1.57 µs)   1.39 µs   1.57 µs   1.57 µs
return_u8_nonblocking()              1.12 µs/iter     (1.03 µs … 1.92 µs)   1.12 µs   1.92 µs   1.92 µs
return_i8_nonblocking()              1.13 µs/iter     (1.05 µs … 1.31 µs)   1.14 µs   1.31 µs   1.31 µs
return_u16_nonblocking()             1.15 µs/iter     (1.05 µs … 1.57 µs)   1.19 µs   1.57 µs   1.57 µs
return_i16_nonblocking()             1.21 µs/iter     (1.06 µs … 2.07 µs)   1.23 µs   2.07 µs   2.07 µs
return_u32_nonblocking()           961.55 ns/iter      (897 ns … 1.09 µs) 989.35 ns   1.09 µs   1.09 µs
return_i32_nonblocking()           976.18 ns/iter   (905.17 ns … 1.12 µs)      1 µs   1.12 µs   1.12 µs
return_u64_nonblocking()           943.17 ns/iter   (919.66 ns … 1.04 µs) 944.27 ns   1.04 µs   1.04 µs
return_i64_nonblocking()           961.59 ns/iter   (906.27 ns … 1.05 µs) 978.97 ns   1.05 µs   1.05 µs
return_usize_nonblocking()         944.81 ns/iter   (898.06 ns … 1.04 µs)  969.5 ns   1.04 µs   1.04 µs
return_isize_nonblocking()          962.6 ns/iter   (892.48 ns … 1.18 µs) 987.95 ns   1.18 µs   1.18 µs
return_f32_nonblocking()           959.11 ns/iter   (908.63 ns … 1.15 µs) 974.95 ns   1.15 µs   1.15 µs
return_f64_nonblocking()           964.69 ns/iter   (904.63 ns … 1.19 µs) 983.65 ns   1.19 µs   1.19 µs
return_buffer_nonblocking()          1.34 µs/iter     (1.26 µs … 1.43 µs)   1.37 µs   1.43 µs   1.43 µs
```

````
> ./target/release/deno bench --unstable --allow-ffi --allow-read test_ffi/tests/bench.js
benchmark                              time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------- -----------------------------
nop()                              249.01 ns/iter  (216.6 ns … 353.91 ns)    255 ns 319.77 ns 339.51 ns
nop_u8()                           409.41 ns/iter (387.72 ns … 583.33 ns)  414.5 ns 473.22 ns 583.33 ns
nop_i8()                           409.05 ns/iter  (373.2 ns … 539.88 ns) 420.89 ns 530.71 ns 539.88 ns
nop_u16()                           426.8 ns/iter (392.35 ns … 576.27 ns) 433.18 ns 556.07 ns 576.27 ns
nop_i16()                          397.76 ns/iter (366.27 ns … 539.16 ns)  405.3 ns 481.61 ns 539.16 ns
nop_u32()                          417.71 ns/iter (387.85 ns … 492.84 ns) 429.83 ns 489.38 ns 492.84 ns
nop_i32()                          394.82 ns/iter (358.37 ns … 493.79 ns) 401.07 ns 446.93 ns 493.79 ns
nop_u64()                           414.2 ns/iter (386.05 ns … 493.97 ns) 422.51 ns 490.09 ns 493.97 ns
nop_i64()                          402.25 ns/iter  (380.4 ns … 470.69 ns) 405.24 ns 456.24 ns 470.69 ns
nop_usize()                        406.93 ns/iter (389.14 ns … 470.14 ns) 411.52 ns 428.72 ns 470.14 ns
nop_isize()                        414.02 ns/iter (392.85 ns … 481.47 ns) 419.27 ns 469.06 ns 481.47 ns
nop_f32()                           399.7 ns/iter (367.51 ns … 467.07 ns) 405.78 ns 455.17 ns 467.07 ns
nop_f64()                           403.4 ns/iter (375.11 ns … 469.95 ns) 411.11 ns 466.77 ns 469.95 ns
nop_buffer()                       530.51 ns/iter  (498.95 ns … 626.2 ns) 541.91 ns 577.58 ns  626.2 ns
return_u8()                        473.93 ns/iter (413.87 ns … 624.84 ns) 490.25 ns 623.47 ns 624.84 ns
return_i8()                        455.67 ns/iter (413.66 ns … 607.31 ns)  455.6 ns 600.93 ns 607.31 ns
return_u16()                       461.85 ns/iter (426.42 ns … 696.58 ns) 465.35 ns 621.59 ns 696.58 ns
return_i16()                       460.31 ns/iter (426.46 ns … 632.57 ns) 467.44 ns 582.07 ns 632.57 ns
return_u32()                       285.06 ns/iter (254.29 ns … 401.45 ns) 292.33 ns 344.46 ns 401.45 ns
return_i32()                       286.85 ns/iter (255.72 ns … 363.39 ns) 294.96 ns 340.89 ns 363.39 ns
return_u64()                       288.76 ns/iter (259.19 ns … 357.66 ns) 293.05 ns 337.74 ns 357.66 ns
return_i64()                       286.58 ns/iter (260.18 ns … 375.12 ns) 289.87 ns 369.55 ns 375.12 ns
return_usize()                      289.7 ns/iter (266.03 ns … 357.58 ns) 294.42 ns 354.82 ns 357.58 ns
return_isize()                     292.02 ns/iter (265.16 ns … 343.05 ns) 299.06 ns 331.61 ns 343.05 ns
return_f32()                       290.09 ns/iter (257.52 ns … 365.24 ns) 297.21 ns 351.57 ns 365.24 ns
return_f64()                       289.57 ns/iter (262.19 ns … 378.06 ns) 294.77 ns 371.24 ns 378.06 ns
return_buffer()                    291.94 ns/iter (263.67 ns … 363.63 ns) 296.79 ns  354.6 ns 363.63 ns
nop_nonblocking()                  451.69 ns/iter  (427.7 ns … 656.19 ns) 449.93 ns 599.08 ns 656.19 ns
nop_u8_nonblocking()               591.24 ns/iter  (537.87 ns … 752.1 ns) 598.11 ns  752.1 ns  752.1 ns
nop_i8_nonblocking()                599.2 ns/iter (574.47 ns … 714.06 ns) 602.65 ns 714.06 ns 714.06 ns
nop_u16_nonblocking()              607.42 ns/iter (547.25 ns … 764.59 ns) 617.75 ns 764.59 ns 764.59 ns
nop_i16_nonblocking()               591.5 ns/iter  (561.1 ns … 690.28 ns) 598.27 ns 690.28 ns 690.28 ns
nop_u32_nonblocking()               600.5 ns/iter  (576.15 ns … 708.5 ns) 606.62 ns  708.5 ns  708.5 ns
nop_i32_nonblocking()              600.21 ns/iter (562.85 ns … 825.15 ns) 603.44 ns 825.15 ns 825.15 ns
nop_u64_nonblocking()              620.04 ns/iter (575.22 ns … 791.98 ns) 632.96 ns 791.98 ns 791.98 ns
nop_i64_nonblocking()              688.28 ns/iter    (579.85 ns … 1.2 µs) 722.55 ns    1.2 µs    1.2 µs
nop_usize_nonblocking()            661.15 ns/iter   (584.53 ns … 1.35 µs)  661.3 ns   1.35 µs   1.35 µs
nop_isize_nonblocking()            700.14 ns/iter    (577.35 ns … 1.1 µs) 740.77 ns    1.1 µs    1.1 µs
nop_f32_nonblocking()              637.73 ns/iter   (560.25 ns … 1.12 µs) 650.55 ns   1.12 µs   1.12 µs
nop_f64_nonblocking()               662.1 ns/iter (600.78 ns … 899.25 ns) 679.85 ns 899.25 ns 899.25 ns
nop_buffer_nonblocking()           785.01 ns/iter   (719.74 ns … 1.34 µs) 794.57 ns   1.34 µs   1.34 µs
return_u8_nonblocking()            708.15 ns/iter (664.87 ns … 988.47 ns) 720.11 ns 988.47 ns 988.47 ns
return_i8_nonblocking()            707.84 ns/iter   (643.63 ns … 1.07 µs) 711.22 ns   1.07 µs   1.07 µs
return_u16_nonblocking()           712.21 ns/iter   (644.67 ns … 1.35 µs) 710.15 ns   1.35 µs   1.35 µs
return_i16_nonblocking()           741.59 ns/iter    (638.2 ns … 1.12 µs) 762.08 ns   1.12 µs   1.12 µs
return_u32_nonblocking()           511.36 ns/iter   (467.51 ns … 1.04 µs) 500.86 ns   1.03 µs   1.04 µs
return_i32_nonblocking()           507.54 ns/iter   (451.44 ns … 1.05 µs) 495.64 ns   1.04 µs   1.05 µs
return_u64_nonblocking()           489.62 ns/iter (450.33 ns … 817.62 ns) 491.33 ns 714.77 ns 817.62 ns
return_i64_nonblocking()           499.26 ns/iter (453.79 ns … 750.28 ns) 510.89 ns 678.76 ns 750.28 ns
return_usize_nonblocking()         502.79 ns/iter (465.02 ns … 654.34 ns) 509.94 ns 612.79 ns 654.34 ns
return_isize_nonblocking()          477.7 ns/iter (445.63 ns … 574.18 ns) 481.72 ns 559.35 ns 574.18 ns
return_f32_nonblocking()           483.16 ns/iter (450.88 ns … 548.27 ns) 489.12 ns 532.94 ns 548.27 ns
return_f64_nonblocking()           490.64 ns/iter (446.03 ns … 754.32 ns) 491.53 ns 718.02 ns 754.32 ns
return_buffer_nonblocking()        524.71 ns/iter (496.63 ns … 599.54 ns) 532.38 ns 593.09 ns 599.54 ns
nop_many_parameters()                3.95 µs/iter     (3.77 µs … 4.38 µs)   3.93 µs   4.38 µs   4.38 µs
nop_many_parameters_nonblocking()   23.67 µs/iter      (7.8 µs … 1.93 ms)   23.7 µs   48.9 µs   64.7 µs
````

Note `nop_many_parameters` and `nop_many_parameters_nonblocking` are missing from the baseline: 1.22.3 release panics on those with `thread 'tokio-runtime-worker' panicked at 'serde_v8: unknown ValueType for v8::Value', serde_v8/payload.rs:38:5`

File: test_ffi/tests/bench.js [hunk 1]
```diff
@@ -0,0 +1,504 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// deno-lint-ignore-file
+
+//const targetDir = Deno.execPath().replace(/[^\/\\]+$/, "");
+const targetDir = "/home/tiffany/dev/rust/deno/target/debug";
```
littledivy, line 5: should be pointing to a release build instead?

aapoalas: This... shouldn't be here. The above `execPath()` is the proper way to get it, I think.

aapoalas: Removed the explicit path.

ext/ffi/lib.rs
```diff
@@ -535,322 +564,792 @@ where
...
+  let isolate = &mut *info.isolate;
+  let callback = v8::Global::from_raw(isolate, info.callback);
+  let context = std::mem::transmute::<
+    NonNull<v8::Context>,
+    v8::Local<v8::Context>,
+  >(info.context);
+  IS_EVENT_LOOP_THREAD.with(|is_event_loop_thread| {
+    if !(*is_event_loop_thread.borrow()) {
+      // Call from another thread, PANIC IN THE DISCO!
+      todo!("Call from another thread");
```
littledivy, line 830: UB to unwind panic from across FFI. Maybe we can change this to an exit(1)

ext/ffi/lib.rs
```diff
@@ -535,322 +564,792 @@ where
...
+  // Call from main thread. If this callback is being triggered due to a
+  // function call coming from Deno itself, then this callback will build
+  // ontop of that stack.
+  // If this callback is being triggered outside of Deno (for example from a
+  // signal handler) then this will either create an empty new stack if
+  // Deno currently has nothing running and is waiting for promises to resolve,
+  // or will (very incorrectly) build ontop of whatever stack exists.
+  // The callback will even be called through from a `while (true)` liveloop, but
+  // it somehow cannot change the values that the loop sees, even if they both
+  // refer the same `let bool_value`.
```
littledivy, line 842: I do not follow. Can you elaborate?

ext/ffi/lib.rs
```diff
@@ -535,322 +564,792 @@ where
...
+      isolate,
+    }))
+  };
+  let cif = Cif::new(
+    args.parameters.into_iter().map(libffi::middle::Type::from),
+    libffi::middle::Type::from(args.result),
+  );
+
+  let closure = libffi::middle::Closure::new(cif, deno_ffi_callback, info);
+  let u3x2 = U32x2::from(*closure.code_ptr() as usize as u64);
```
littledivy, line 1063: Add a TODO to use BigInt later.

ext/ffi/lib.rs
```diff
@@ -535,322 +564,792 @@ where
...
+        } else {
+          return Err(type_error("Invalid FFI pointer type, expected null, BigInt, ArrayBuffer, or ArrayBufferView"));
+        }
+      }
+      NativeType::Function {} => {
+        if value.is_null() {
+          let value: *const u8 = ptr::null();
+          ffi_args.push(NativeValue { pointer: value })
+        } else if value.is_big_int() {
+          // Do we support this? This would be a foreign function pointer given to us by an FFI library.
```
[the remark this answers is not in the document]
aapoalas: Yes, especially now since I moved to using the pointer value even for registered callbacks. This is now any function pointer, including own registered one.

ext/ffi/lib.rs
aapoalas: I guess an `eprintln!` or something like that would be useful before exiting?

ext/ffi/lib.rs
aapoalas: Ah, this has to do with the sick example I showed in Discord that used a signal handler on the FFI Rust side to call the registered callback.

So a code like this:
```js
let shouldContinueLoop = true;
const callback = () => {
  console.log("Stop looping!");
  shouldContinueLoop = false;
};
const regCb = new Deno.RegisteredCallback({ parameters: [], result: "void" }, callback);
dylib.symbols.store_callback(regCb);
dylib.symbols.call_stored_callback_after_delay(); // Sets up a SIGALRM interrupt handler that calls the stored callback, and then sets up a SIGALRM interrupt signal to be fired in 2 seconds time.

let count = 0;
while(shouldContinueLoop) {
  if (count++ == 10 ** 6) {
  console.log("No I won't!");
 }
}
```
will output:
> Stop looping!
> No I won't!
and will never exit.

If you add a `performance.now()` call into the loop, then it will exit properly.
If you add a `Math.random()` call into the loop, then it will panic.
Also, if you add a `console.trace()` to the callback you'll see the callback emanating from inside the while-loop.
So, if an interrupt handler is setup to call back into JS and Deno currently has an active stack, the callback will simply step on top of that stack and do its work. That does not mean that the data that Deno (or V8, rather) uses will be affected as expected.

This is, I think, something that should be carefully documented but trying to fight against this would probably not be a good idea, as I don't think that at present we have the tools to do anything except accept this.

ext/ffi/lib.rs
aapoalas: I could also change this code to manually create a `v8::Array` and add the two items?

ext/ffi/lib.rs
littledivy: sure

ext/ffi/lib.rs
```diff
@@ -535,322 +569,809 @@ where
...
+  fn name(&self) -> Cow<str> {
+    "registeredcallback".into()
+  }
+
+  fn close(self: Rc<Self>) {
+    let info = unsafe { Box::from_raw(self.info as *mut CallbackInfo) };
+    let isolate = unsafe { info.isolate.as_mut().unwrap() };
+    unsafe { v8::Global::from_raw(isolate, info.callback) };
+    unsafe { v8::Global::from_raw(isolate, info.context) };
+    drop(self)
```
littledivy, line 809: drop is implicit already
```suggestion
```

ext/ffi/lib.rs
```diff
@@ -535,322 +569,809 @@ where
...
+impl Resource for UnsafeCallbackResource {
+  fn name(&self) -> Cow<str> {
+    "registeredcallback".into()
+  }
+
+  fn close(self: Rc<Self>) {
+    let info = unsafe { Box::from_raw(self.info as *mut CallbackInfo) };
+    let isolate = unsafe { info.isolate.as_mut().unwrap() };
+    unsafe { v8::Global::from_raw(isolate, info.callback) };
+    unsafe { v8::Global::from_raw(isolate, info.context) };
```
littledivy, line 808: `close()` is essentially an unsafe operation - It's UB to drop the Global and CallbackInfo while `deno_ffi_callback` references it. Either rename `close` in JS to `unsafeClose` or indicate `deno_ffi_callback` about close() using a bool field in CallbackInfo?

test_ffi/tests/bench.js
```diff
@@ -0,0 +1,503 @@
...
+    344657895n,
+    -2436732n,
+    135.26437e3,
+    264.3576468623546834,
+    buffer2,
+  );
+});
+
+Deno.bench("nop_many_parameters_nonblocking()", () => {
+  dylib.symbols.nop_many_parameters_nonblocking(
```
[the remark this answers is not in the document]
aapoalas: I wanted to test beyond the absolute limites of what might be reasonable :)

ext/ffi/lib.rs
littledivy: Ah yes, in that case `close()` is fine. An invalid pointer is a predictable consequence - let's just document this anyways.

ext/ffi/lib.rs
```diff
@@ -535,322 +569,808 @@ where
...
+        } else if value.is_big_int() {
+          let value = v8::Local::<v8::BigInt>::try_from(value).unwrap();
+          let value = value.u64_value().0 as *const u8;
+          ffi_args.push(NativeValue { pointer: value });
+        } else if value.is_array_buffer() || value.is_array_buffer_view() {
+          let value: ZeroCopyBuf = serde_v8::from_v8(scope, value)?;
+          let value: &[u8] = &value[..];
+          ffi_args.push(NativeValue {
+            pointer: value.as_ptr(),
+          });
```
littledivy, line 708: UB: `value` is dropped at the end of this function, what you push to `ffi_args` is a dangling pointer

aapoalas: ZeroCopyBuf, as its name suggests, does not copy the data from the `ArrayBuffer` but instead acts as a wrapper that can be used to reference into the memory. ZeroCopyBuf is dropped, but the `ArrayBuffer` lives until V8 decides it's time for it to be GC'd.

So, this definitely can be UB if the user passes in an AB that is then immediately dropped and a GC triggers. Otherwise it's totally fine.

ext/ffi/lib.rs
```diff
@@ -535,322 +562,806 @@ where
...
-#[op]
-fn op_ffi_call_ptr<FP>(
+#[derive(Deserialize)]
+struct RegisterCallbackArgs {
+  parameters: Vec<NativeType>,
+  result: NativeType,
+}
+
+#[op(v8)]
+fn op_ffi_unsafe_callback_create(
```
littledivy, line 1059: Missing unstable & permissions check

littledivy (approved):
LGTM!

littledivy:
We're going to further iterate on the API and performance. This is a great first pass

aapoalas:
Good to merge on my part.

ext/ffi/lib.rs
```diff
@@ -189,6 +230,7 @@ enum NativeType {
   F32,
   F64,
   Pointer,
+  Function {},
```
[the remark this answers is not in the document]
piscisaureus: > Can just be `Function` instead of an enum with named fields?

🤞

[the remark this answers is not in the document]
piscisaureus: > Can just be `Function` instead of an enum with named fields?

🤞

piscisaureus:
> * The format for declaring function type parameters has two objects. I'd like to turn this into a tagged object eg. `{ "type": "function", "parameters": [], "result": "void" }` but didn't find a way to serde work with me on that.

Did `#[serde(tag = "type")]` not work?

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -346,9 +341,14 @@ declare namespace Deno {
     | "f64"
     | "pointer";
 
+  type NativeParameterType =
```
piscisaureus, line 344: I was thinking, maybe it makes things more explicit if we had `NativeParameterType` and `NativeReturnType`, as in:
* `NativeParameterType` can be sent from JS to C
* `NativeReturnType` can be sent from C to JS

Of course in case of callbacks the types would be reversed: parameters would be `NativeReturnType` and the return value would have type `NativeParameterType`.

So maybe we should consider`ToNativeType` and `FromNativeType` or something.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -513,6 +521,83 @@ declare namespace Deno {
...
+    StaticNativeBigIntType ? bigint
+    : T extends StaticNativeNumberType ? number
+    : T extends "pointer" ? UnsafePointer
+    : never;
+
+  type UnsafeCallbackResult<T extends NativeParameterType> = T extends "void"
+    ? void
+    : T extends StaticNativeBigIntType ? number | bigint
+    : T extends StaticNativeNumberType ? number
+    : T extends "pointer" ? UnsafePointer | TypedArray | null
```
piscisaureus, line 559: If you use TypedArray, who ends up owning the memory? How is it released?

aapoalas:
> Did `#[serde(tag = "type")]` not work?

I'll have to recheck, but when I tried originall serde did not like mixing untagged (plain strings) and internally tagged (`tag = "type"`) enum members, or so it seemed.

cli/dts/lib.deno.unstable.d.ts
aapoalas: I agree, and do plan on doing a rework on the FFI TS types. f.ex. Currently `NativeType` is the "base type" and `NativeNumberType` is a narrowed down type from that. I will do a follow-up to rework the types into something that hopefully resembles a reasonable whole.

A `ToNativeType` and `FromNativeType` sounds splendid for that.

cli/dts/lib.deno.unstable.d.ts
aapoalas: I am by no means an expert, but from my understanding (or lack of it) of the C ABI (or lack of it :D ) there is no way for a function call to pass actual ownership of memory.

1. Pointers are of course purely pointers to the memory. The memory itself is still owned by whoever allocated it, and the caller and callee must agree on what is found at the other end of the pointer.
2. Structs (which the FFI API currently does not actually support) pass struct parts through various registers (and others on some platforms), but these are still copies of the actual data in memory, the memory itself is still allocated and released by the same party.

Of course with structs (and even with pointers) there's a possibility that whatever is contained in the struct has pointers that should be considered to carry ownership with it. An example would be the result of Rust Box's `Box::to_raw_parts()` which returns essentially a `(*const u8, usize, usize)` struct. This struct would likely get passed in through FFI in a bunch of registers. From Deno FFI point of view this would be a `new BigIntArray(3)`. Now the question is, what happens to the three BigInts when this struct gets passed through an FFI boundary?
1. The values of the usize integers get copied to the other side of the FFI boundary.
2. The ownership of the (implied) `Box` is transferred to the other side of the FFI boundary.
3. The receiver of ownership is expected to call any required APIs when it is done with the ownership. Of course just removing the usize integers isn't going to be enough. Releasing whatever is found on the other side of pointers needs to be done according to the original allocators preferences. Generally this means calling a destructor / drop implementation.
4. The values on the original side of the FFI boundary, what happens to them? They are not changed, of course. Three (essentially) usize integers are not going to changed or deallocated. What they represent, that is the important part but it's also beyond what FFI can represent: The understanding that the first `*const u8` is a pointer referring to a contiguous memory of some length determined by the value of one of the `usize` parameters, and that whoever has last received that pointer should carry the responsibility of calling a destructor when they're done with it is plainly a contract, not one built into the ABI itself.

So, when a `TypedArray` is used as a pointer, the ownership of the memory (as a physical address space) remains always with Deno, and by extension the user of the FFI library. The user must make sure that the `TypedArray` does not go out of scope and get garbage collected for as long as the pointer is given to an FFI library. Deno's FFI can only provide the means by which the contract of memory usage is shared (passing pointers to memory), it cannot (with the current API at least) accurately represent passing of ownership.

If / once structs are supported (this is another planned followup of mine), those will likely operate on a different logic. Whereas pointers are just referring to memory, structs are generally understood to pass ownership. On a technical level the data is still copied, but eg. if a struct contains a pointer (like with the above `Box` example) then it is generally understood that the ownership is passed along with the struct. For this reason, I think the proper model for struct passing is definitely detaching ArrayBuffers.

So, where passing a `TypedArray` as pointer means that Deno must keep the buffer alive then on the other hand passing a `TypedArray` as struct would generally mean that any ownership implied by the buffer are transferred to the other side of the FFI boundary. As such, it's generally more apt that the buffer ownership is taken away from Deno and passed onto Rust (which will then deallocate it according to the ArrayBuffer's deleter callback). If then user code truly wants to retain part-ownership of the struct contained in the `TypedArray` they'll need to either plainly copy the buffer or (more likely) call some native library's copy constructor with another `TypedArray` argument into which the copy will be created.

aapoalas:
> Did `#[serde(tag = "type")]` not work?

Okay, I tested this and indeed there doesn't seem to be a clear way to do this.

We'd need the same enum to handle both of these: `"u8"` and `{ "function": { "parameters": [...], "result": "..." } }` or `{ "type": "function", "parameters": [...], "result": "..." }`. If `serde(tag = "type")` is used then it would imply that `U8` should be presented in JSON as `{ type: "u8" }`. On the other hand, If we want `{ type: "function", parameters: [...], result: "..." }` to get parsed then that would imply the `serde(tag = "type")` setting.

So, these two obviously do not coexist. Or at least I am not aware of a way for the two to coexist. Luckily there's also a benefit for the current predicament. `Function {}` requires the JSON format to be `{ "function": { ... } }`, where we of course use the `{ ... }` to be `{ "parameters": [...], "result": "..." }`. That object format then happens to exactly align with the declaration that we expect `UnsafeCallback` to be given, meaning that if so wanted FFI users can reuse the same objects for both callback parameter declaration AND the `UnsafeCallback` declaration that will eventually get passed as the parameter.

littledivy:
Yeah, cannot mix different tag representations in serde.

cli/dts/lib.deno.unstable.d.ts
littledivy: tl;dr ownership always resides with the backing store managed by V8.

nayeemrmn:
Even for the API, mixing `"u8"` and `{ type: "function", ... }` doesn't seem desirable at all, since it's an inconsistent tagging scheme. That would be the third best option after (1) what we have now aligning with serde's defaults and (2) internally tagged as `{ type: "u8" }` + `{ type: "function", ... }`.

aapoalas:
> Even for the API, mixing `"u8"` and `{ type: "function", ... }` doesn't seem desirable at all, since it's an inconsistent tagging scheme. That would be the third best option after (1) what we have now aligning with serde's defaults and (2) internally tagged as `{ type: "u8" }` + `{ type: "function", ... }`.

Yeah, I originally wanted to avoid the extra object needed for `{ function: {} }`, but now I'm very happy with it, since as said it allows the inner object to be reused as-is for both the parameter and the `UnsafeCallback`'s definition.

PR #14670, opened 2022-05-19
feat: add 32 bits support for serde_v8

PerfectLaugh:
This is the initial effort for enabling 32 bits support for deno, and this PR doesn't even compile if you don't modify the part which isn't relevant to this PR.

There's still two test cases failed after the modification.
image
image

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,6 +1,9 @@
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
 
+[target.i686-pc-windows-msvc]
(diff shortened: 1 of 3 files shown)


PerfectLaugh:
Related PR in Rusty_V8: https://github.com/denoland/rusty_v8/pull/976

littledivy (commented):
There are also assumptions made in `ext/ffi` that the target architecture is 64bit.

PerfectLaugh:
Should we make a type alias for pointer then?
Will be cumbersome if we have to determine u64/u32 all the time.

PR #14671, opened 2022-05-19
fix(core): rethrow exception during structured cloning serialization

### LINKED ISSUE #14296: It's wrong to transform all exceptions thrown during structured cloning to `DataCloneError` exceptions
Deno's implementation of the structured clone algorithm is divided into a JS part, and a native part which doesn't have access to `DOMException`. Since serialization errors must be represented by throwing a `DataCloneError` `DOMException`, the native part [throws a `TypeError` with the required message](https://github.com/denoland/deno/blob/main/core/bindings.rs#L958), and the JS part [rethrows any exception thrown by the serialization as a `DataCloneError`](https://github.com/denoland/deno/blob/main/ext/web/13_message_port.js#L278).

As shown by web-platform-tests/wpt#33663, however, this behavior is wrong, because exceptions thrown from an object's property getters should be surfaced without changes:

```js
const testObject = {
  get testProperty() {
    throw new Error();
  }
};
try {
  structuredClone(testObject);
} catch(err) {
  console.log(err instanceof DOMException);  // should be false; Deno prints true
}
```

One way to fix this would be to have an optional third parameter to the internal API `Deno.core.serialize` that would be a `(string) => any` callback, returning the exception to throw. Or possibly `(string) => never`, having the callback itself throw the exception.

mrkldshv:
* Introduced optional callback for `Deno.core.serialize` API, that returns cloning error if there is one. 
* Removed `try/catch` in seralize structured clone function and throw error from callback.
* Removed "Object with a getter that throws" assertion from WPT.

Closes  #14296.


lucacasonato:
LGTM. @andreubotella do you want to take a look too? I remember you were originally looking into this.

core/bindings.rs
```diff
@@ -937,6 +937,7 @@ fn decode(
 
 struct SerializeDeserialize<'a> {
   host_objects: Option<v8::Local<'a, v8::Array>>,
+  cb: Option<v8::Local<'a, v8::Function>>,
```
andreubotella, line 940: This field could use a more descriptive name.

core/bindings.rs
```diff
@@ -1078,6 +1088,17 @@ fn serialize(
       }
     };
 
+  let arg2_to_cb = match !args.get(2).is_undefined() {
```
andreubotella, line 1091: This variable too.

andreubotella (approved):
LGTM

core/bindings.rs
mrkldshv: Maybe something like `error_callback`?

andreubotella: That works.

mrkldshv: Alright, thanks! I'll rename it.

PR #14672, opened 2022-05-19
fix(bench): update typo in bench summary

GJZwiers:
This PR updates a small issue in the benchmark summary output. Previously it would print e.g. `2x times faster` which is a duplication because it would read as `2 times times faster`. It was updated in `mitata` 0.0.7 to `2x faster` instead,

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2498,9 +2498,9 @@ dependencies = [
 
 [[package]]
 name = "mitata"
-version = "0.0.6"
+version = "0.0.7"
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):

GJZwiers:
Do I have to re-sign the CLA? I didn't get a notification I don't think.

bartlomieju:
> Do I have to re-sign the CLA? I didn't get a notification I don't think.

It's a flake, I'll restart CI and it should pass

Issue #14675, opened 2022-05-19
Permissions APIs do not properly handle bad descriptor inputs


cjihrig:
```js
await Deno.permissions.query();
```

This throws `TypeError: Cannot read properties of undefined (reading 'name')`.

PR #14676, opened 2022-05-19
fix(runtime): improve permission descriptor validation

cjihrig:
This commit improves the permission descriptor validation by
explicitly checking for object types and using optional
chaining when creating error messages in case the descriptor
is not an object.

Fixes: https://github.com/denoland/deno/issues/14675

### DIFF
--- a/runtime/js/10_permissions.js
+++ b/runtime/js/10_permissions.js
@@ -149,7 +149,7 @@
    * @returns {desc is Deno.PermissionDescriptor}
    */
   function isValidDescriptor(desc) {
-    return desc && desc !== null &&
+    return typeof desc === "object" && desc !== null &&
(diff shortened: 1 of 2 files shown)


runtime/js/10_permissions.js
```diff
@@ -185,7 +186,8 @@
       if (!isValidDescriptor(desc)) {
         return PromiseReject(
           new TypeError(
-            `The provided value "${desc.name}" is not a valid permission name.`,
+            `The provided value "${desc
+              ?.name}" is not a valid permission name.`,
```
bartlomieju, line 190: Really strange formatting... CC @dsherret

bartlomieju (approved):
LGTM

PR #14678, opened 2022-05-19
refactor: strip utf8 BOM in file fetcher

dsherret:
Right now the BOM is being stripped in deno_ast (being moved to deno graph), but instead I believe we could remove it right at the start and not have to deal with it.

This would also reduce a copy of the source text that previously needed to be made when a file contained a BOM.

### DIFF
--- a/cli/file_fetcher.rs
+++ b/cli/file_fetcher.rs
@@ -8,6 +8,7 @@ use crate::http_util::CacheSemantics;
 use crate::http_util::FetchOnceArgs;
 use crate::http_util::FetchOnceResult;
 use crate::text_encoding;
+use crate::text_encoding::strip_bom_mut;
 use crate::version::get_user_agent;
 
 use data_url::DataUrl;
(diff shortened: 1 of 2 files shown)


lucacasonato (commented):
Considering that deno_graph will take in byte slices instead of strings in the not to distant future, I'd like to avoid adding more text manipulation into the loader in CLI. Wdyt?

dsherret:
@lucacasonato I think it's worth it to do this at the source because then we don't need to deal with this encoding detail anywhere else for the most part and we do this before it gets put in an `Arc<str>` so we reduce cloning. I think once we have the source text in an `Arc<str>` then we shouldn't touch it anymore to reduce memory usage since files can be quite large.

lucacasonato:
We can't do this in the CLI source fetcher though, as we don't know at this point if a given asset is WASM or JS/JSON. We don't want to strip BOM out of WASM, as that isn't UTF8

dsherret:
> We can't do this in the CLI source fetcher though, as we don't know at this point if a given asset is WASM or JS/JSON. We don't want to strip BOM out of WASM, as that isn't UTF8

All Wasm files start with a module which has a four byte magic value that's `\0asm` so I believe we are ok doing this.

https://webassembly.github.io/spec/core/binary/modules.html#binary-module

image

dsherret:
I think we'll have to update the code in deno graph to specify if it's loading binary data or utf8 data and have it up to the `Loader` implementation to convert to utf8 w/o BOM. That way we don't duplicate files in memory, which could be expensive (ex. this caches files).

dsherret:
We're going to explore an alternate solution in deno_graph.

PR #14690, opened 2022-05-20
fix(cli/dts): change `ChildStatus.signal` from `string` to `Deno.Signal`

lowlighter:
Since #14643 the `ChildStatus.signal` changed from `number` to `string` which is great, but a more accurate typing would be `Deno.Signal`

Currently if you try to pass it down to a type/interface which expect a `Deno.Signal`, you must force the type assertion to avoid TypeScript complaining:
image

Not complaining with assertion:
image

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -1117,7 +1117,7 @@ declare namespace Deno {
     | {
       success: false;
       code: number;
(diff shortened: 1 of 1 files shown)


crowlKats (approved):
LGTM, apologies about this

lowlighter:
No worries, thanks a lot for your awesome work !

Issue #9995, opened 2021-04-04
🙏🏻 [feat/req] Implement signal handlers for Windows


rivy:
Currently, all the available signal handling API throw "unimplemented" errors for Windows platforms.

It would be great to have cross-platform signal handling along the lines of [NodeJS Process Signal Events](https://nodejs.org/api/process.html#process_signal_events).

rivy:
There is a stalled partial implementation ([Windows CTRL-C/CTRL-BREAK (SIGINT/SIGBREAK) signal handler](https://github.com/denoland/deno/pull/10258)).

### related

- [Signal Handlers/discussion](https://github.com/denoland/deno/issues/2339)
- [Initial signal handling (non-Windows) implementation](https://github.com/denoland/deno/pull/3757)
- [Signal syntactic sugar implementation](https://github.com/denoland/deno/pull/4696)

rivy:
😃 
Yeah! There's some revived signal handling work going on in #14694.

BlackGlory:
I encountered this error on deno 1.39.1.

```ts
import puppeteer from 'npm:puppeteer-core@21.7.0'

await puppeteer.launch({
  executablePath: 'chrome'
})
```

```
error: Uncaught (in promise) TypeError: Windows only supports ctrl-c (SIGINT) and ctrl-break (SIGBREAK).
```

guillaume86:
The `handleSIGTERM: false` is not required, only handleSIGHUP ;).

bartlomieju:
Done in https://github.com/denoland/deno/issues/26331.

PR #14694, opened 2022-05-21
feat(runtime/signal): implement SIGINT and SIGBREAK for windows

### LINKED ISSUE #10236: Support for windows ctrl-c signal

GJZwiers:
Revival of https://github.com/denoland/deno/pull/10258. Adds support for SIGINT and SIGBREAK signals on Windows platform.

- [x] update [`std/node/`](https://github.com/denoland/deno_std/blob/6069bfe4137864214daf43637069fa533f04ba82/node/internal_binding/constants.ts#L91) with SIGBREAK

Lots of credits to @juzi5201314 for the initial efforts.

Closes #10236.


bartlomieju:
> An update to [std/node/](https://github.com/denoland/deno_std/blob/6069bfe4137864214daf43637069fa533f04ba82/node/internal_binding/constants.ts#L91) is needed to resolve a couple of failing tests.

@GJZwiers can you handle that first?

GJZwiers:
@bartlomieju The type errors for sigbreak are gone now, but other tests are failing. Maybe that's because of the std update?
https://github.com/denoland/deno/runs/6597255925?check_suite_focus=true#step:32:1160

bartlomieju:
> @bartlomieju The type errors for sigbreak are gone now, but other tests are failing. Maybe that's because of the std update? https://github.com/denoland/deno/runs/6597255925?check_suite_focus=true#step:32:1160

Yes, that's true. I'll make sure to upgrade `deno_std` in a separate PR to unblock you.

bartlomieju:
Opened https://github.com/denoland/deno/pull/14722

bartlomieju:
@GJZwiers after merging with `main` you should be good to go

GJZwiers:
@bartlomieju There is a test I want to include here, which works well when run with `deno test` but somehow it does not when added to the unit tests and run with `cargo test`. 
The test looks like this:

```ts
Deno.addSignalListener("SIGINT", () => {
  console.log("parent process received SIGINT");
});

Deno.test("signal listener handles signal", async () => {
  const child = Deno.spawnChild(Deno.execPath(), {
    stdin: "null",
    stdout: "inherit",
    stderr: "inherit",
    args: [
      "eval",
      `console.log("child process started");
    Deno.addSignalListener("SIGINT", () => {
      console.log("child process received SIGINT");
      Deno.exit(0);
    });
    let n = 0;
    setInterval(() => {
      console.log(n + 1);
      if (n === 2) {
        clearInterval();
        Deno.kill(Deno.pid, "SIGINT");
      }
      n++;
    }, 500);
    `,
    ],
  });

  const output = await child.output();

  assertEquals(output.status.success, true);
  assertEquals(output.status.code, 0);
});
```

Output of `deno test -A --unstable` then becomes:

```
signal listener handles signal ...
------- output -------
child process started
1
2
3
child process received SIGINT
parent process received SIGINT
----- output end -----
signal listener handles signal ... ok (2s)
```

With `cargo test js_unit_tests` though, the test will interrupt the whole test suite, maybe because the default action for SIGINT somehow remains? I have run it together with other tests in `deno test` as well, and that works well too, they just keep executing after the signal test.

Do you have any ideas on what the problem could be with `cargo test`? I don't think the problem is with the windows implementation, because the same thing happened in the CI on Linux too.

runtime/ops/process.rs
```diff
@@ -301,37 +301,66 @@ pub fn kill(pid: i32, signal: &str) -> Result<(), AnyError> {
...
+
+  if matches!(signal, "SIGINT" | "SIGBREAK") {
+    let is_generated = unsafe {
+      GenerateConsoleCtrlEvent(
+        match signal {
+          "SIGINT" => CTRL_C_EVENT,
+          "SIGBREAK" => CTRL_BREAK_EVENT,
+          _ => unreachable!(),
+        },
+        pid as u32,
```
piscisaureus, line 330: @GJZwiers Did you check that processes actually receive the `CTR_C_EVENT`? 

According to MSDN (which is often wildly inaccurate, but still) you can only send it to `CTR_C_EVENT` to all processes in the terminal.
From: https://docs.microsoft.com/en-us/windows/console/generateconsolectrlevent

> This signal cannot be limited to a specific process group. If dwProcessGroupId is nonzero, this function will succeed, but the CTRL+C signal will not be received by processes within the specified process group.

runtime/ops/process.rs
```diff
@@ -301,37 +301,66 @@ pub fn kill(pid: i32, signal: &str) -> Result<(), AnyError> {
...
 
-  if !matches!(signal, "SIGKILL" | "SIGTERM") {
-    Err(type_error(format!("Invalid signal: {}", signal)))
-  } else if pid <= 0 {
-    Err(type_error("Invalid pid"))
-  } else {
-    let handle = unsafe { OpenProcess(PROCESS_TERMINATE, FALSE, pid as DWORD) };
+  if pid <= 0 {
+    return Err(type_error(
+      "Process id (pid) cannot be negative or zero on Windows.",
```
piscisaureus, line 318: Wrt `CTRL_C_EVENT` and `CTRL_BREAK_EVENT`: Windows supports sending them to `pid 0`.

runtime/ops/signal.rs
```diff
@@ -60,6 +49,53 @@ impl Resource for SignalStreamResource {
...
+  }
+}
+
+#[cfg(windows)]
+impl WindowsSignal {
+  pub async fn recv(&mut self) -> Option<()> {
+    match self {
+      WindowsSignal::Sigint(ctrl_c) => ctrl_c.recv().await,
+      WindowsSignal::Sigbreak(ctrl_break) => ctrl_break.recv().await,
+    }
```
piscisaureus, line 78: We could also listen to `CTRL_CLOSE_EVENT` and map it to `SIGHUP`. Node does this too.

runtime/ops/process.rs
GJZwiers: I checked some of the examples on https://deno.land/manual/examples/os_signals and they are working with the current changes. A Deno child process also receives sigint/sigbreak on windows now, with e.g. something like this:

```js
Deno.addSignalListener("SIGINT", () => {
    console.log("parent process received SIGINT");
});

const child = Deno.spawnChild(Deno.execPath(), {
  stdin: "null",
  stdout: "inherit",
  stderr: "inherit",
  args: [
    "eval",
    `console.log("child process started");
    Deno.addSignalListener("SIGINT", () => {
      console.log("child process received SIGINT");
      Deno.exit(128);
    });
    let n = 0;
    setInterval(() => {
      console.log(n + 1);
      if (n === 2) {
        clearInterval();
        Deno.kill(Deno.pid, "SIGINT");
      }
      n++;
    }, 500);
    `,
  ],
});
```

runtime/ops/process.rs
GJZwiers: TIL :) made some changes

piscisaureus (approved):
LGTM, thank you @GJZwiers!

rivy:
@GJZwiers , thanks for this work!

> Closes #9995 and #10236.

This only partially fixes #9995 (which is requesting broad signal handler parity for WinOS). It definitely goes a long way toward the goal by implementing the most commonly needed signals ('SIGBREAK' and 'SIGINT'). But, I wouldn't close #9995 based on the partial implementation.

GJZwiers:
> @GJZwiers , thanks for this work!
> 
> > Closes #9995 and #10236.
> 
> This only partially fixes #9995 (which is requesting broad signal handler parity for WinOS). It definitely goes a long way toward the goal by implementing the most commonly needed signals ('SIGBREAK' and 'SIGINT'). But, I wouldn't close #9995 based on the partial implementation.

Ah my bad, I'll update the description

runtime/ops/process.rs
dsherret: I was trying to write some tests for SIGINT and SIGBREAK in this PR and ran into this issue where sending a SIGINT to the child process also kills the parent in the terminal.

rivy: For WinOS, new "child" processes, created from a console process using `CreateProcess()`, all share the same console (*unless* created with `CREATE_NEW_CONSOLE` or `DETACHED_PROCESS`). <kbd>CTRL+C</kbd> and <kbd>CTRL+BREAK</kbd> events are sent to all processes which share that console, so all processes in the console process group will receive the event/signal.

For situations requiring shared I/O with a "parent", such as a shim, I'll just use null handlers for signals in the parent while waiting for the "child" process to finish.

ref: [MS ~ Creation of a Console](https://docs.microsoft.com/en-us/windows/console/creation-of-a-console) @@ <https://archive.is/yFRT5>
ref: [MS ~ CTRL+C and CTRL+BREAK Signals](https://docs.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals) @@ <https://archive.is/6T140>

PR #14697, opened 2022-05-21
Create root directory implicitly if `DENO_INSTALL_ROOT` does not exist

oefd:
If you install deno and attempt to install something when the deno root does not exist and is set with `DENO_INSTALL_ROOT` you get an incredibly unhelpful error:

    error: No such file or directory (os error 2)

This is because inside the `resolve_shim_data` call an attempt is made to canonicalize the path which doesn't yet exist.

I'm not sure this PR is the right approach to fix the issue, it might be better to try and replicate what I assume happens elsewhere in the codebase to ensure `.deno` exists, because I can't replicate this issue when I unset `DENO_INSTALL_ROOT`

### DIFF
--- a/cli/tools/installer.rs
+++ b/cli/tools/installer.rs
@@ -105,7 +105,7 @@ exec deno {} "$@"
 fn get_installer_root() -> Result<PathBuf, io::Error> {
   if let Ok(env_dir) = env::var("DENO_INSTALL_ROOT") {
     if !env_dir.is_empty() {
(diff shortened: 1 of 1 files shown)


dsherret (changes requested):
Thanks! Would you be able to sign the CLA and add a test for this in cli/tests/integration/install_tests.rs? After that should be good to merge. (I think you could just copy install_custom_dir_env_var and use a sub directory of the temp dir)

PR #14698, opened 2022-05-21
feat(ext/net): return [] when no DNS record found

### LINKED ISSUE #14695: `Deno.resolveDns` return when there's no record found
**Actual:**
```
Uncaught NotFound: no record found for name: www.google.com. type: SOA class: IN
    at async <anonymous>:2:1
```
**Ideal/Expected:**
```js
[]
/* or */ null
/* or */ undefined
```
Should `Deno.resolveDns` return `[]`, `null`, or `undefined` instead throw an error when there're no records found for the name.

This issue is related to #14373 which causes Deno apps won't fail or panic.

PatrickChoDev:
This make `Deno.resolveDns` return empty `Vec<DnsReturnRecord>` instead throw an error when there're no records found for the name.

Closes #14695.

### DIFF
--- a/cli/tests/testdata/resolve_dns.ts
+++ b/cli/tests/testdata/resolve_dns.ts
@@ -53,12 +53,6 @@ console.log(JSON.stringify(srv));
 console.log("TXT");
 console.log(JSON.stringify(txt));
 
-try {
-  await Deno.resolveDns("not-found-example.com", "A", nameServer);
-} catch (e) {
-  console.log(
-    `Error ${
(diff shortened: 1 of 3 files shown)


cjihrig:
FWIW, Node.js throws in this case as well. IMHO we should keep the current behavior.

PatrickChoDev:
> FWIW, Node.js throws in this case as well. IMHO we should keep the current behavior.

Oh, I missed that, Thank you very much.

Issue #14683, opened 2022-05-20
`deno coverage` prints `NaN%` for files that only declare an interface


GJZwiers:
Link: https://discord.com/channels/684898665143206084/689420767620104201/976976786900549672

Steps to reproduce:
- Create `User.ts`:

```ts
export interface User {
   id: string;
}
```
- Create `deps.ts`:

```ts
export {
   assertEquals,
} from 'https://deno.land/std@0.139.0/testing/asserts.ts';
```

- Create `index.ts`:

```ts
export * from './deps.ts';
export * from './User.ts';
```

- Create `User_test.ts`:

```ts
import {
   assertEquals,
   User,
} from './index.ts';

Deno.test('User', () => {
   const user: User = {
      id: 'id',
   };

   assertEquals(user.id, 'id');
});
```

- Run `deno test --coverage=cov/`
- Run `deno coverage cov/`

Output:

```
cover file:///C:/Users/GJZwiers/repos/rpg-ts/User.ts ... NaN% (0/0)
cover file:///C:/Users/GJZwiers/repos/rpg-ts/deps.ts ... 100.000% (3/3)
cover file:///C:/Users/GJZwiers/repos/rpg-ts/index.ts ... 100.000% (2/2)
```

The `NaN%` becomes `100%` if a variable declaration like `const foo = "bar";` is added to the file with the interface `User.ts`

PR #14699, opened 2022-05-21
fix(coverage): do not report transpiled files with no lines 

cjihrig:
This commit omits files from the coverage report that have no lines of code to report coverage for.

Fixes: https://github.com/denoland/deno/issues/14683


bartlomieju (approved):
LGTM

PR #14701, opened 2022-05-22
feat(ext/net): `Deno.resolveDns()` won't panic for not supported type

### LINKED ISSUE #14373: Should `Deno.resolveDns` handle error instead of panicked?
From #14355. For others invalid or typo in code, should Deno fall to `_ => {_}` case instead of panicking the application? The way to implements `_` case I thought is..

1. Print info as missing typo in code?
2. Print info as this is unsupported DNS record type and link to opening issue on this repo.

After that `_` case would return for `None` instead of `Some`.

PatrickChoDev:
Closes #14373


cjihrig (commented):
Are you able to add a test?

ext/net/ops.rs
```diff
@@ -704,7 +705,16 @@ where
     .filter_map(rdata_to_return_record(record_type))
     .collect();
 
-  Ok(results)
+  match results[0] {
+    DnsReturnRecord::Unsupported => {
+      let var: &'static str = record_type.into();
+      Err(custom_error(
+        "NotSupported",
+        var.to_owned() + " records are not supported for now",
```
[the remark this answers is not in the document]
PatrickChoDev: Done

PatrickChoDev:
> Are you able to add a test?
Since It's an unsopprted case. I think this'll be an integration test only?

cjihrig:
> I think this'll be an integration test only?

I think that makes sense.

bartlomieju (changes requested):
This looks good, but we really need a test to land it

cli/tests/testdata/resolve_dns.ts.out
```diff
@@ -23,3 +23,4 @@ SRV
 TXT
 [["I","am","a","txt","record"],["I","am","another","txt","record"],["I am a different","txt record"],["key=val"]]
 Error NotFound thrown for not-found-example.com
+Error NotFound thrown for not-found-example.com
```
cjihrig, line 26: I would expect to see something different here. This test should be validating unsupported record type behavior.

ext/net/ops.rs
```diff
@@ -704,7 +705,16 @@ where
     .filter_map(rdata_to_return_record(record_type))
     .collect();
 
-  Ok(results)
+  match results[0] {
+    DnsReturnRecord::Unsupported => {
+      let var: &'static str = record_type.into();
+      Err(custom_error(
+        "NotSupported",
+        var.to_owned() + " records are not supported",
```
cjihrig, line 713: Are either of these showing up in the test added by this PR?

cli/tests/testdata/resolve_dns.ts.out
PatrickChoDev: I'm not sure what the message should look like, could you provide me some?

ext/net/ops.rs
PatrickChoDev: This won't be appear in ts file but will display on Deno CLI, as I checked.

cli/tests/testdata/resolve_dns.ts.out
cjihrig: The text here was created on lines 69-73 of the test added in this PR. Based on what this PR is supposed to be changing, I would expect the message to indicate that the record type is not supported. The message should ideally come from the runtime itself and not a string in the test file.

cli/Cargo.toml
```diff
@@ -120,11 +120,12 @@ winapi = { version = "=0.3.9", features = ["knownfolders", "mswsock", "objbase",
 [dev-dependencies]
 deno_bench_util = { version = "0.51.0", path = "../bench_util" }
 flaky_test = "=0.1.0"
+openssl = { version = "0.10", features = ["vendored"] }
```
ry, line 123: We are not going to add OpenSSL dependency.

lucacasonato: More helpful feedback: use `rustls` to generate the RSA key. You can view some code in `ext/crypto` for how this is done there.

PatrickChoDev: I've found that `trust-dns` also have builtin function to generate RSA keys but it's on openssl lib inside.

Cargo.lock
```diff
@@ -1088,6 +1088,7 @@ dependencies = [
  "deno_core",
  "deno_tls",
  "log 0.4.17",
+ "openssl",
```
cjihrig, line 1091: I think all of these OpenSSL related changes need to be reverted before this can land.

cjihrig (changes requested):

ext/net/Cargo.toml
```diff
@@ -17,8 +17,9 @@ path = "lib.rs"
 deno_core = { version = "0.140.0", path = "../../core" }
 deno_tls = { version = "0.45.0", path = "../tls" }
 log = "0.4.16"
+openssl = { version = "0.10", features = ["vendored"] }
 serde = { version = "1.0.136", features = ["derive"] }
 socket2 = "0.4.4"
 tokio = { version = "1.17", features = ["full"] }
-trust-dns-proto = "=0.21.2"
+trust-dns-proto = { version = "=0.21.2" }
```
cjihrig, line 24: It looks like this change is no longer necessary.

cli/tests/testdata/resolve_dns.ts
```diff
@@ -62,3 +62,15 @@ try {
...
+  await Deno.resolveDns("example.com", "DNSKEY", nameServer);
+} catch (e) {
+  console.log(e.message);
+}
+
+try {
+  await Deno.resolveDns("example.com", "HINFO", nameServer);
+} catch (e) {
+  console.log(e.message);
+}
```
cjihrig, line 76: I think we only need one of these.

cjihrig (commented):
It looks like this is still pulling in OpenSSL?

cli/tests/testdata/resolve_dns.ts
PatrickChoDev: I'll remove DNSKEY then

ext/net/Cargo.toml
PatrickChoDev: I've forgetten to remove from main cargo.toml, sorry.

PR #14704, opened 2022-05-23
fix: read raw stdin to prevent buffering (regression)

dsherret:
A test for this started failing in the deno_task repo with the latest Deno. This is an edge case that probably won't affect anyone.

It turns out, Rust is buffering `std::io::stdin()` and we don't want that, so use raw stdin (unlike stdout and stderr where we want to use rust's helpers).


bartlomieju (approved):
LGTM

PR #14706, opened 2022-05-23
fix(cli): add config flags to `deno info`

### LINKED ISSUE #14703: `deno info` does not support `--config` or `--no-config`
It should work the same way as it does in `deno run`. The auto discovered config file is used by `deno info` for the `importMap` field.

mrkldshv:
Add config flags (`--config`, `--no-config`) to `deno info` command.

Closes #14703.


lucacasonato (commented):
Can you add an integration test to check that the override flags for `deno info` actually work?

bartlomieju (approved):
Thanks!

PR #14716, opened 2022-05-24
fix(core): op metrics op_names mismatch

AaronO:


cli/tests/unit/metrics_test.ts
```diff
@@ -76,3 +76,14 @@ Deno.test(function metricsForOpCrates() {
...
 });
+
+// Test that op_names == Objects.keys(Deno.core.ops)
+// since building the per-op metrics depends on op_names being complete
+Deno.test(function opNamesMatch() {
+  assertEquals(
+    // @ts-ignore: Deno.core allowed
+    Deno.core.opNames().sort(),
+    // @ts-ignore: Deno.core allowed
+    Object.keys(Deno.core.ops).sort(),
```
bartlomieju, line 87: Why do we need `Deno.core.opNames()` if you can get names of ops via `Deno.core.ops`?

AaronO: The order of op names from `opNames()` aligns with the order of metrics (essentially idx == op_id), the `Object.keys(Deno.core.ops)` doesn't preserve order so we can't form the correct pairs

Issue #14708, opened 2022-05-23
Stabilize "Deno.sleepSync" API


bartlomieju:
It's been unstable long enough, there's really not much to it.

It would allow to refactor and remove some code from `deno_web` extension, namely `TimerPermissions` trait wouldn't need to have `check_unstable` method.

lucacasonato:
@ry might have objections here. I know he didn't really like the API back when.

lucacasonato:
Also, we should check that the same functionality isn't already possible with just `Atomics`.

dsherret:
Apparently the issue with `Atomics.wait` is that it doesn't currently work in the main worker context?

We discussed this super briefly in the design meeting and if there's a way we could provide a helper in deno.land/std for doing this with `Atomics.wait`, we think that would be preferable.

I do really think we should have a way to sleep synchronously though as it can provide a more accurate sleep mechanism for some non-web server applications (ex. when using hardware).

ry:
sleepSync has no utility - it's very easy to misuse - let's remove it not stabilize it.

dsherret:
> sleepSync has no utility

...for web servers

PR #14719, opened 2022-05-25
Remove unstable Deno.sleepSync

ry:
closes https://github.com/denoland/deno/issues/14708


dsherret (approved):
LGTM considering `Atomics.wait` seems to be a good substitute.

cli/tests/testdata/worker_drop_handle_race_terminate.js
```diff
@@ -16,8 +16,9 @@ const WORKER2 = getCodeBlobUrl(`
   // We sleep for slightly under 2 seconds in order to make sure that worker 1
   // has closed, and that this worker's thread finishes normally rather than
   // being killed (which happens 2 seconds after calling terminate).
-  Deno.sleepSync(1800);
```
dsherret, line 19: I think you might have to use `Atomics.wait` for some of these and using `setTimeout` might possibly make some tests flakier.

bartlomieju: CC @andreubotella

lucacasonato: The synchronous sleep here is intentional (the thread should be blocked, and the event loop should not be executing). As such you can not just replace this with `setTimeout`.

PR #14725, opened 2022-05-25
fix(fmt): prevent infinite loop when formatting certain binary expressions

dsherret:
Main fix is here: https://github.com/dprint/dprint/commit/4dceb874de60f18958aee0d43c5fcacf3abb1de9#diff-1b5f5a0428033aa5ae0d4f2cbb13980abe7edd73d58bbc6efe0af720e186ea87R515-R516


ext/web/06_streams.js
```diff
@@ -569,10 +569,14 @@
...
-        stream[_inFlightCloseRequest] = stream[_pendingAbortRequest] =
-          undefined;
+    stream[_storedError] =
+      stream[_writer] =
+      stream[_controller] =
+      stream[_inFlightWriteRequest] =
+      stream[_closeRequest] =
+      stream[_inFlightCloseRequest] =
+      stream[_pendingAbortRequest] =
+        undefined;
```
dsherret, line 579: Oh yeah, it also formats this kind of stuff better now.

bartlomieju (approved):
LGTM

Issue #14315, opened 2022-04-18
Deno.test with output: Show ok on the same line as the test name


AnInternetTroll:
It would be nice for the "ok" to be shown on the same line as the test name. Here it feels like the "ok" is disconnected from the test, and the test name is incomplete
![image](https://user-images.githubusercontent.com/40470127/163869166-1df09ab0-e024-4c6b-86e6-3d0684cc3675.png)

nayeemrmn:
We should apply this if there is anything in between the test start and result, including sub-steps as well.

PR #14729, opened 2022-05-26
refactor: Allocate IDs for tests

nayeemrmn:
Blocked on #14757.

Allows for better state-keeping of tests in Rust:
- Prints ` ... cancelled` for tests that were registered and cancelled due to uncaught errors:
  ```
  running 3 tests from ./temp.ts
  testWithUncaughtRejection ...
  Uncaught error from ./temp.ts FAILED
  testWithUncaughtRejection ... cancelled (0ms)
  anotherTest ... cancelled (0ms)
  anotherOne ... cancelled (0ms)
  ```
- Always know whether or not to reprint `someTest ...` when a result arrives (addresses https://github.com/denoland/deno/issues/14315#issuecomment-1138141249).
- Allows lots of future improvements to output e.g. more self-contained result lines for concurrent runs:
  ```
  (./temp.ts) t1 ... ok
  (./temp.ts) t2 ... t2-step1 ... FAILED
  (./temp.ts) t2 ... t2-step2 ... ok
  (./temp.ts) t2 ... FAILED
  ```

Adds `op_register_test()` to register tests, allocate IDs to them and potentially other info:
- Move filtering to rust, `{ id, filteredOut }` is returned by this op.
-  Tests are known about in Rust earlier, allowing for the `... cancelled` output above for tests that haven't started yet.

Cleanups in LSP test runner and more functional approach in `40_testing.js`. Bring bench runner error formatting up to date, it was still using `formatError()` in JS.


dsherret (approved):
LGTM. Probably @kitsonk should review the LSP changes before merge though.

bartlomieju (approved):
LGTM, nice work Nayeem. I verified that LSP changes work properly in vscode_deno.

cli/lsp/testing/collectors.rs
```diff
@@ -42,7 +37,7 @@ fn fn_to_steps(
   parent: &str,
   level: usize,
   function: &ast::Function,
-) -> Option<Vec<TestDefinition>> {
+) -> Vec<TestDefinition> {
```
kitsonk, line 40: Again, an arbitrary change.

cli/lsp/testing/definitions.rs
```diff
@@ -137,27 +138,21 @@ impl TestDefinitions {
     })
   }
 
+  pub fn inject(&mut self, data: lsp_custom::TestData) -> bool {
```
kitsonk, line 141: some doc as to what this is any why

cli/lsp/testing/execution.rs
```diff
@@ -61,12 +61,12 @@ fn as_queue_and_filters(
           if let Some(test) = test_definitions.get_by_id(id) {
             let filter =
               filters.entry(item.text_document.uri.clone()).or_default();
-            if let Some(include) = filter.maybe_include.as_mut() {
+            if let Some(include) = filter.include.as_mut() {
```
kitsonk, line 64: arbitrary renaming, why?  Our convention is that options are named "maybe_".

cli/lsp/testing/execution.rs
```diff
@@ -131,14 +122,14 @@ fn as_test_messages<S: AsRef<str>>(
 }
 
 #[derive(Debug, Clone, Default, PartialEq)]
-struct TestFilter {
-  maybe_include: Option<HashMap<String, TestDefinition>>,
-  maybe_exclude: Option<HashMap<String, TestDefinition>>,
+struct LspTestFilter {
```
kitsonk, line 125: Arbitrary renaming, why?

cli/lsp/testing/execution.rs
```diff
@@ -131,14 +122,14 @@ fn as_test_messages<S: AsRef<str>>(
...
+  include: Option<HashMap<String, TestDefinition>>,
+  exclude: HashMap<String, TestDefinition>,
```
kitsonk, line 127: Why is exclude now not optional?

cli/lsp/testing/execution.rs
```diff
@@ -131,14 +122,14 @@ fn as_test_messages<S: AsRef<str>>(
...
 }
 
-impl TestFilter {
+impl LspTestFilter {
```
kitsonk, line 130: arbitrary renaming, why?

cli/lsp/testing/execution.rs
```diff
@@ -991,14 +947,5 @@ mod tests {
           .to_string()
       ]
     );
-    assert_eq!(
```
kitsonk, line 994: Why remove a test assertion?  Specifically testing the serialization of JSON, where else is this tested?

cli/lsp/testing/lsp_custom.rs
```diff
@@ -92,8 +93,9 @@ pub enum TestRunKind {
 pub struct TestRunRequestParams {
   pub id: u32,
   pub kind: TestRunKind,
-  #[serde(skip_serializing_if = "Option::is_none")]
-  pub exclude: Option<Vec<TestIdentifier>>,
+  #[serde(skip_serializing_if = "Vec::is_empty")]
+  #[serde(default)]
+  pub exclude: Vec<TestIdentifier>,
```
kitsonk, line 98: Changing things like this that come from the client changes the API with the client.  This message is specifically sent from the client, and you are now making an optional property mandatory. 😢   Why make a breaking change with the API?

cli/lsp/testing/lsp_custom.rs
```diff
@@ -21,8 +21,9 @@ pub struct TestData {
   pub id: String,
   /// The human readable test to display for the test.
   pub label: String,
-  #[serde(skip_serializing_if = "Option::is_none")]
-  pub steps: Option<Vec<TestData>>,
+  #[serde(skip_serializing_if = "Vec::is_empty")]
+  #[serde(default)]
+  pub steps: Vec<TestData>,
```
kitsonk, line 26: This changes the deserialization of this structure.  I can't remember off hand if this is ever sent from the client, but why arbitrarily break the contract with the client?

kitsonk (commented):
I'm concerned some of the changes to the lsp aren't "cleanup" but arbitrary changes.  Specifically for the messages that go to and from the client that are changed should have their types changed in `https://github.com/denoland/vscode_deno/blob/main/client/src/lsp_extensions.ts` to ensure there isn't a knock on effect.  Especially when we remove `Option` it becomes a mandatory field on deserializing.

cli/lsp/testing/collectors.rs
```diff
@@ -15,7 +15,7 @@ fn arrow_to_steps(
   parent: &str,
   level: usize,
   arrow_expr: &ast::ArrowExpr,
-) -> Option<Vec<TestDefinition>> {
+) -> Vec<TestDefinition> {
```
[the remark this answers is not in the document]
nayeemrmn: This is a redundant `Option` unless I missed something, if there are no steps that is represented as an empty vec

cli/lsp/testing/execution.rs
nayeemrmn: This is another redundant `Option`, blocklists are inherently non-optional because that's the same thing as being empty

cli/lsp/testing/execution.rs
nayeemrmn: That's not true, I saw https://github.com/denoland/deno/blob/7d9e8baebb6c2534346e0823e2cc6408ebb8b1eb/cli/lsp/testing/lsp_custom.rs#L90-L99 and plenty of other places. It was attractive to switch here because `exclude` was made non-optional.

In my experience, prefixes like `try_foo()` and `maybe_foo()` are used for _methods_ to distinguish from their counterparts which panic on failure/absence. It's usually not needed for fields and variables thanks to editors, it's sort of Hungarian

cli/lsp/testing/execution.rs
nayeemrmn: There is another `TestFilter` added in `cli/tools/test.rs`, it's used in this file

cli/lsp/testing/execution.rs
nayeemrmn: `.as_test_options()` doesn't exist anymore, filters aren't passed to JS since it's moved to Rust

cli/lsp/testing/lsp_custom.rs
nayeemrmn: See https://serde.rs/field-attrs.html#default

cli/lsp/testing/definitions.rs
nayeemrmn: Sure, though it's already clearer than before... we had `TestDefinitions::injected` already, I extracted a method to populate it. I suppose the return-if-insertion-succeeded thing could be doc'd

cli/tests/testdata/test/uncaught_errors.out
```diff
@@ -53,6 +54,6 @@ bar 2 => ./test/uncaught_errors_2.ts:3:6
 bar 3 => ./test/uncaught_errors_2.ts:6:6
 ./test/uncaught_errors_3.ts (uncaught error)
 
-FAILED | 2 passed | 5 failed ([WILDCARD])
```
dsherret, line 56: @nayeemrmn seems like this is flaky now: https://github.com/denoland/deno/runs/7362087847?check_suite_focus=true

PR #14733, opened 2022-05-26
BREAKING: remove deprecated Deno APIs

bartlomieju:
Ref https://github.com/denoland/deno/issues/9795


cli/dts/lib.deno.ns.d.ts
```diff
@@ -710,115 +710,6 @@ declare namespace Deno {
...
-   *
-   * ```ts
-   * const source = await Deno.open("my_file.txt");
-   * const bytesCopied1 = await Deno.copy(source, Deno.stdout);
-   * const destination = await Deno.create("my_file_2.txt");
-   * const bytesCopied2 = await Deno.copy(source, destination);
-   * ```
-   *
-   * @deprecated Use `copy` from https://deno.land/std/streams/conversion.ts
-   * instead. `Deno.copy` will be removed in Deno 2.0.
```
dsherret, line 726: "Will be removed in Deno 2.0"

dsherret (commented):
The comments in the file say that this functionality will be removed in Deno 2.0 so I think we should close this PR and wait for Deno 2.0 unless this PR is for Deno 2.0. The other changes we've been doing are for unstable APIs, but these are stable APIs.

lucacasonato (changes requested):
I agree with David. Not at all semver compatible. There is still a lot of code out there that uses these APIs. There is no rush to remove them.

cli/dts/lib.deno.ns.d.ts
```diff
@@ -946,7 +839,7 @@ declare namespace Deno {
    *
    * Resolves to the number of bytes written.  This function is one of the lowest
    * level APIs and most users should not work with this directly, but rather use
-   * Deno.writeAll() instead.
+   * `writeAll()` from https://deno.land/std/streams/conversion.ts instead.
```
dsherret, line 842: We should merge this in a separate PR.

ry:
we'll wait...

PR #14743, opened 2022-05-27
fix(scripts/release): add unstable flag when running ./tools/format.js

### LINKED ISSUE #14739: version_bump should run the format command to format the Releases.md
I had a version_bump generated PR fail CI because the Releases.md needed to be formatted because of rather long commit messages.  The process should run formatting as part of generating the PR.

dsherret:
Need to use the unstable flag when launching ./tools/format.js

Closes #14739

### DIFF
--- a/tools/release/01_bump_crate_versions.ts
+++ b/tools/release/01_bump_crate_versions.ts
@@ -8,10 +8,6 @@ const repo = workspace.repo;
 const cliCrate = workspace.getCliCrate();
 const originalCliVersion = cliCrate.version;
 
-// update the std version used in the code
-console.log("Updating std version...");
-await updateStdVersion();
-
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):

tools/release/01_bump_crate_versions.ts
```diff
@@ -8,10 +8,6 @@ const repo = workspace.repo;
 const cliCrate = workspace.getCliCrate();
 const originalCliVersion = cliCrate.version;
 
-// update the std version used in the code
-console.log("Updating std version...");
-await updateStdVersion();
```
dsherret, line 13: Moved this below because on line 23 (now 19) it may prompt when running the script manually and the user might cancel the prompt.

PR #14744, opened 2022-05-27
feat(core): support creating snapshots from existing snapshots

bartlomieju:


crowlKats (approved):

core/bindings.rs
```diff
@@ -186,24 +183,22 @@ pub fn set_func(
...
+  snapshot_options: SnapshotOptions,
 ) {
   let key = v8::String::new(scope, name).unwrap();
   let external = v8::External::new(scope, external_data as *mut c_void);
   let builder =
     v8::FunctionTemplate::builder_raw(callback).data(external.into());
   let templ = if let Some(fast_function) = fast_function {
     // Don't initialize fast ops when snapshotting, the external references count mismatch.
-    if snapshot_loaded && !will_snapshot {
+    if snapshot_options == SnapshotOptions::Load {
```
bartlomieju, line 201: I think this was wrong originally - probably should check if snapshot was loaded or there was no snapshot at all

crowlKats: ok

core/bindings.rs
```diff
@@ -190,19 +190,22 @@ pub fn set_func_raw(
...
   let builder =
     v8::FunctionTemplate::builder_raw(callback).data(external.into());
   let templ = if let Some(fast_function) = fast_function {
     // Don't initialize fast ops when snapshotting, the external references count mismatch.
-    if !snapshot_loaded {
-      builder.build(scope)
-    } else {
+    if matches!(
+      snapshot_options,
+      SnapshotOptions::Load | SnapshotOptions::None
```
bartlomieju, line 203: This seems wrong comparing with the left side

core/runtime.rs
```diff
@@ -2104,6 +2172,14 @@ impl JsRealm {
     v8::HandleScope::with_context(isolate, &self.0)
   }
 
+  pub fn context_scope<'s>(
+    &self,
+    scope: &'s mut v8::HandleScope<'s>,
+  ) -> v8::ContextScope<'s, v8::HandleScope<'s>> {
+    let context = v8::Context::new(scope);
+    v8::ContextScope::new(scope, context)
+  }
```
bartlomieju, line 2181: Not needed?

core/bindings.rs
crowlKats: your previous review said to change it to this

bartlomieju: err, okay, you're right

PR #14750, opened 2022-05-28
test(core): Test that sync ops return/throw objects in the right realm

andreubotella:
This behavior was introduced in #14019 but wasn't properly tested in that PR.


cjihrig (approved):

bartlomieju (approved):
LGTM

PR #14754, opened 2022-05-29
chore(docs): use cross-platform example in deno.run jsdoc

GJZwiers:
This PR changes the `Deno.run` JSDoc to use a `curl` example that works the same on all the major platforms. Also adds `await p.status` so the program doesn't exit before the output is shown on the terminal.

### DIFF
--- a/cli/dts/lib.deno.ns.d.ts
+++ b/cli/dts/lib.deno.ns.d.ts
@@ -2432,7 +2432,7 @@ declare namespace Deno {
    *
    * ```ts
    * const p = Deno.run({
-   *   cmd: ["echo", "hello"], // Use ["pwsh", "-c", "echo", "hello"] on Windows.
+   *  cmd: ["curl", "https://example.com"],
    * });
    * const status = await p.status();
    * ```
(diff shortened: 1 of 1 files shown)


cli/dts/lib.deno.ns.d.ts
```diff
@@ -2427,13 +2427,14 @@ declare namespace Deno {
...
   }
 
-  /** Spawns new subprocess.  RunOptions must contain at a minimum the `opt.cmd`,
+  /** Spawns new subprocess. RunOptions must contain at a minimum the `opt.cmd`,
    * an array of program arguments, the first of which is the binary.
    *
    * ```ts
    * const p = Deno.run({
-   *   cmd: ["echo", "hello"],
+   *   cmd: ["echo", "hello"], // Use ["pwsh", "-c", "echo", "hello"] on Windows.
```
dsherret, line 2435: This seems to suggest that people should use `Deno.run` with `pwsh` on Windows. To reduce confusion, it would be good to use a command that works on all three operating systems such as `curl`.

dsherret (changes requested):
Thanks for this, but it would be good to instead use a single command that works on all three operating systems.

GJZwiers:
Sure thing, I will rewrite it.

dsherret (approved):
LGTM. Thanks @GJZwiers!

PR #14758, opened 2022-05-30
wip serialize ModuleMap

bartlomieju:
It seems to be crashing on M1 with SIGSEGV :( `cargo test -p deno_core snapshot`

### DIFF
--- a/core/modules.rs
+++ b/core/modules.rs
@@ -751,16 +751,22 @@ pub(crate) enum ModuleError {
   Other(Error),
 }
 
-/// A collection of JS modules.
-pub(crate) struct ModuleMap {
-  // Handling of specifiers and v8 objects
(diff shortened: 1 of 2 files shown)


bartlomieju:
This approach won't work. We can only serialize things that are `v8::Value` using `serde_v8` and `v8::Module` is `v8::Data` and not `v8::Value`. 

Other approach to explore is to manually serialize `ModuleMapInner` into a V8 object - since all `v8::Value` are also `v8::Data` we could then embed it into the snapshot using `SnapshotCreator::AddData` method. This will require some manual labor for serialization, but seems doable. The disadvantage is that when creating a snapshot from an existing snapshot all embedded data is lost, so we'll need to figure out how to make it work with #14744

PR #14764, opened 2022-06-01
feat(ext/crypto): export elliptic keys as "raw"

diachedelic:
This pull request chips away at the WebCrypto roadmap (#11690). It adds support for the "raw" format when exporting public ECDH/ECDSA keys via the SubtleCrypto.exportKey method.

The WebCrypto specification [explicitly states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-exportKey) that support for the "raw" format is encouraged for interoperability.

As a newbie contributor, are there tests for ext/crypto? I could not find any.


lucacasonato:
> As a newbie contributor, are there tests for ext/crypto? I could not find any.

There are two types of tests for Web Crypto. Some simple ones in https://github.com/denoland/deno/blob/main/cli/tests/unit/webcrypto_test.ts, and an extensive test suite in the [Web Platform Tests](https://web-platform-tests.org/). You probably just need to update the WPT expectations file to match this new feature. See [Manual: Contributing > Web Platform Tests](https://deno.land/manual/contributing/web_platform_tests).

diachedelic:
I've added some unit tests. Oddly enough, the WPT suite does seem to test exportKey, only importKey.

diachedelic:
I've signed the CLA three times in the last week, and it's still pending.

bartlomieju:
> I've signed the CLA three times in the last week, and it's still pending.

@diachedelic are you sure you used the same GH account to sign the CLA?

jamesdiacono:
Right you are! I had to sign it with my other GitHub account, because the CLAssistant looks at the email address  in the actual commit.

ext/crypto/shared.rs
```diff
@@ -144,7 +144,7 @@ impl RawKeyData {
             )
           })?;
         let point = p384::EncodedPoint::from_bytes(&ec_key.encoded_point)
-          .map_err(|_| data_error("expected valid public EC key"))?;
+          .map_err(|_| data_error("expected valid private EC key"))?;
```
littledivy, line 147: This is the public encoded point - the error should not be for private key decoding.

littledivy (approved):
LGTM, just one minor nit.

ext/crypto/shared.rs
diachedelic: Oops, fixed.

PR #14769, opened 2022-06-01
fix(ext/fetch): remove deprecation of `URL` in deno `fetch`

### LINKED ISSUE #14620: Undeprecate the URL parameter of fetch
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1269 allows lib.dom to accept URL, so I think the reasoning in #14113 is now invalid:

>Deno's fetch and lib.dom's fetch have slightly different declarations. (Deno's fetch accept URL as the first arg, but lib.dom's fetch doesn't)

mrkldshv:
Remove deprecation of `URL` as the first argument in deno `fetch`.

Closes #14620.


kt3k (approved):
LGTM

lib.dom's `fetch` declaration is updated in #14242 (TypeScript 4.7 update), which includes `URL` in the 1st arg. We should align this again

bartlomieju (approved):
Thanks

PR #14771, opened 2022-06-01
chore: fix flaky `integration::bench::multifile_summary` test

### LINKED ISSUE #14732: `integration::bench::multifile_summary` is flaky
https://github.com/denoland/deno/runs/6608476505

GJZwiers:
Fixes #14732. The test is flaky because sometimes "parse url 6x" runs faster than "parse url 2x". I think it may be because the execution time difference between the two benches is quite small and if a CI runner has a small performance dip, it results in the parse url 2x being slower than the parse url 6x on occasion. 

This PR changes the 6x to 200x creating a bigger difference. An alternative could also be to change the slower/faster to wildcards instead.


bartlomieju (approved):
LGTM, thanks!

PR #14775, opened 2022-06-01
[WIP]: feat: File watcher with dynamic imports

bartlomieju:

### DIFF
--- a/check_watcher.ts
+++ b/check_watcher.ts
@@ -1,0 +1,2 @@
+console.log("hello there121233!");
+await import("./imported.ts");
--- a/cli/proc_state.rs
+++ b/cli/proc_state.rs
@@ -358,6 +358,12 @@ impl ProcState {
       reload: reload_on_watch,
     };
 
+    // let maybe_reporter = if false {
+    //   Some(GraphReporter)
+    // } else {
+    //   None
+    // };
+
     let graph = create_graph(
       roots.clone(),
       is_dynamic,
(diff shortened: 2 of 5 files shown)


cli/proc_state.rs
```diff
@@ -358,6 +402,13 @@ impl ProcState {
       reload: reload_on_watch,
     };
 
+    let maybe_file_watcher_reporter: Option<&dyn deno_graph::source::Reporter> =
```
bartlomieju, line 405: I'm not sure if we should pass this reporter to `create_graph` or to `CliModuleLoader` - the latter will ensure all local files that V8 execute will get watched

check_watcher.ts
```diff
@@ -0,0 +1,2 @@
+console.log("Hopefully dynamic import will be watched...");
```
lucacasonato, line 1: TODO: remove

lucacasonato (approved):
LGTM! Thanks 🎉

cli/tests/integration/watcher_tests.rs
```diff
@@ -496,15 +505,21 @@ fn run_watch() {
     .unwrap();
   let (mut stdout_lines, mut stderr_lines) = child_lines(&mut child);
 
-  assert_contains!(stdout_lines.next().unwrap(), "Hello world");
```
dsherret, line 499: @bartlomieju I think to make the other tests not flaky we will have to switch them to use `wait_for`/`wait_contains` as well.

PR #14776, opened 2022-06-02
fix(lsp): handle get diagnostic errors better

kitsonk:
This handles errors in `getDiagnostics` request where there are transient errors in getting the diagnostics which cause an output of:

```
ERROR TSLS = {}
```

In the debug console, which is distracting to users and doesn't impact the usage of the extension.  Also, when one specifier causes an issue, all requested specifiers get dropped, which isn't ideal.  This will handle the whole failure of getting diagnostics better.

### DIFF
--- a/cli/tsc/99_main_compiler.js
+++ b/cli/tsc/99_main_compiler.js
@@ -752,25 +752,18 @@ delete Object.prototype.__proto__;
           /** @type {Record<string, any[]>} */
           const diagnosticMap = {};
           for (const specifier of request.specifiers) {
-            let diagnostics = [];
-            try {
-              diagnostics = [
-                ...languageService.getSemanticDiagnostics(specifier),
-                ...languageService.getSuggestionDiagnostics(specifier),
-                ...languageService.getSyntacticDiagnostics(specifier),
-              ];
-            } catch {
-              // just swallow errors here, as they are meaningless
-            }
-            diagnosticMap[specifier] = fromTypeScriptDiagnostic(
-              diagnostics.filter(({ code }) =>
-                !IGNORED_DIAGNOSTICS.includes(code)
-              ),
-            );
+            diagnosticMap[specifier] = fromTypeScriptDiagnostic([
+              ...languageService.getSemanticDiagnostics(specifier),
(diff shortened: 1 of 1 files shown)


cli/tsc/99_main_compiler.js
```diff
@@ -752,11 +752,21 @@ delete Object.prototype.__proto__;
...
-              ...languageService.getSyntacticDiagnostics(specifier),
-            ].filter(({ code }) => !IGNORED_DIAGNOSTICS.includes(code)));
+            let diagnostics = [];
+            try {
+              diagnostics = [
+                ...languageService.getSemanticDiagnostics(specifier),
+                ...languageService.getSuggestionDiagnostics(specifier),
+                ...languageService.getSyntacticDiagnostics(specifier),
+              ];
+            } catch {
```
dsherret, line 762: Huh, interesting. Would it make sense to do this per method call? I guess not since these are transient errors? What kind of errors get thrown?

dsherret (approved):
LGTM thanks for looking into this!

cli/tsc/99_main_compiler.js
dsherret: It would be nice to know exactly what’s happening, but this seems fine to do.

kitsonk: Looking at all the code, I suspect it is some sort of debug logging/failure that is occurring, related to an attempt to find a source file, but it being in a bad state it non-critically debug errors, but because we don't fully implement the debug host, instead of it being something meaningful handled by the TypeScript internal debugger, it bubbles up as a `{}`.

One thing we don't do is validate on the `getDiagnostics` that all the specifiers have source files... actually that is an interesting point, I will look into that.  It could be we just need to go grab any missing source files, where we are asking for diagnostics for source files that aren't in tsc yet.

kitsonk: ok, found it, it was `ts.OperationCanceledException` being thrown, which we should just swallow like we do with the `OperationCanceledError`, which explains the transient nature of the error.  I have refactored it again to better handle that error.

dsherret: Perfect! Thanks for looking into it.

PR #14780, opened 2022-06-02
refactor: Delete unused field from OpDecl

espindola:
This was user visible via .middleware, but it is not clear if that was
intentional.


AaronO:
This is intentional, `#[op(v8)]` isn't currently used in the CLI codebase, it makes sense to track that in the declaration

PR #14793, opened 2022-06-03
refactor(core): Move Deno.core bindings to ops

nayeemrmn:
Move all `Deno.core` bindings which are defined in Rust to ops e.g. `Deno.core.encode(foo)` -> `Deno.core.opSync("op_encode", foo)` using `#[op(v8)]`. Ref #14574, #14582. One more change to the macro was required to avoid late-bound lifetime parameters when returning a value with a lifetime compatible based on `scope`'s.

This breaks any third party uses of these `Deno.core` bindings. The following are used in std and need to be updated in synchrony with this change:
- `Deno.core.setMacrotaskCallback()`
- `Deno.core.setNextTickCallback()`
- `Deno.core.runMicrotasks()`
- `Deno.core.hasTickScheduled()`
- `Deno.core.setHasTickScheduled()`
- `Deno.core.evalContext()`
- `Deno.core.encode()`

Or we could preserve these as wrappers defined in JS to avoid breakage... but why and which ones and how long to keep them? Not a fan.


AaronO (changes requested):
Great work overall, pretty much what I had in mind following https://github.com/denoland/deno/pull/14582

However I would suggest the following changes for this first pass:
1. Scope changes to `core/`, have outside callers still use `Deno.core.foo()` by adding light wrappers in `01_core.js`.
2. Keep the "v8 ops" separate from the other "regular ops", e.g: in an `ops_v8.rs` or `ops_builtin_v8.rs` or just in bindings for now

nayeemrmn:
> 1. Scope changes to `core/`, have outside callers still use `Deno.core.foo()` by adding light wrappers in `01_core.js`.

Okay, I did that for the ones used in std but it's fairer to do it with all of them after all :-) I think in a couple of versions we should remove all of those basic wrappers, including existing `Deno.core.read()` etc. in favour of the ops.

AaronO:
> > 1. Scope changes to `core/`, have outside callers still use `Deno.core.foo()` by adding light wrappers in `01_core.js`.
> 
> Okay, I added wrappers for the ones used in std but it's fairer to do it with all of them after all :-) I think in a couple of versions we should remove all of those basic wrappers, including existing `Deno.core.read()` etc. in favour of the ops.

Both `opSync` and `opAsync` will likely disappear soon, and so calls might be called as `Deno.core.ops.op_foo` or `Deno.ops.op_foo` or `Deno.ops.foo`

ops/lib.rs
```diff
@@ -379,9 +382,21 @@ fn is_rc_refcell_opstate(arg: &syn::FnArg) -> bool {
...
+    || tokens(arg).ends_with(": & mut v8 :: HandleScope < 'a >")
     || tokens(arg).ends_with(": & mut deno_core :: v8 :: HandleScope")
+    || tokens(arg).ends_with(": & mut deno_core :: v8 :: HandleScope < 'a >")
 }
 
 fn tokens(x: impl ToTokens) -> String {
   x.to_token_stream().to_string()
 }
+
+fn exclude_lifetime_params(
```
littledivy, line 394: nice

core/ops_builtin_v8.rs
```diff
@@ -0,0 +1,770 @@
+use crate::error::is_instance_of_error;
```
littledivy, line 1: Add copyright header

littledivy (commented):
great stuff

core/bindings.rs
```diff
@@ -1,144 +1,26 @@
...
-        function: terminate.map_fn_to(),
-      },
-      v8::ExternalReference {
-        function: apply_source_map.map_fn_to(),
-      },
-      v8::ExternalReference {
-        function: op_names.map_fn_to(),
-      },
-    ])
+    v8::ExternalReferences::new(&[v8::ExternalReference {
```
bartlomieju, line 19: Cool that you limited number of external references, that should simplify snapshotting a bit. Ideally we'd remove all of them, but that's outside of scope for this PR.

AaronO: In the future I would like to make ops-snapshottable, avoiding the teardown/setup we currently have and allowing code to hold long lived references avoid the indirection and IC costs of lookups.

AaronO: So we'll likely never provide zero external refs, rather than a fixed count it will become variable (per isolate)

core/runtime.rs
```diff
@@ -1806,6 +1806,25 @@ impl JsRuntime {
   }
 }
 
+pub(crate) fn script_origin<'a>(
```
bartlomieju, line 1809: Technically this is breaking change - it was in `bindings.rs` and it was `pub` - I suggest to revert this change, it's just moving code around.

cli/tests/testdata/eval_context_throw_with_conflicting_source.ts.out
```diff
@@ -1,4 +1,5 @@
 [WILDCARD]error: Uncaught Error: foo
 Warning Couldn't format source line: Column 31 is out of bounds (source may have changed at runtime)
     at file:///[WILDCARD]/eval_context_conflicting_source.ts:1:31
+    at [WILDCARD]
```
bartlomieju, line 4: What's changed here?

ext/web/13_message_port.js
```diff
@@ -205,7 +205,7 @@
     const transferables = [];
     const hostObjects = [];
     const arrayBufferIdsInTransferables = [];
-    const transferedArrayBuffers = [];
```
bartlomieju, line 208: These are technically unrelated, but thanks for fixing it

bartlomieju:
Blocked by update after https://github.com/denoland/deno/pull/14604 was merged

cli/tests/testdata/eval_context_throw_with_conflicting_source.ts.out
nayeemrmn: There's a new stack frame for `opSync`

bartlomieju (approved):
LGTM, I'll wait for one more approval before landing

PR #14794, opened 2022-06-03
refactor(snapshots): to their own crate

AaronO:
A revival of #10472, which is generally cleaner and facilitates cross-compiling deno, thus unblocking arm64 CI builds


piscisaureus (approved):

core/runtime.rs
```diff
@@ -53,6 +53,16 @@ pub enum Snapshot {
   Boxed(Box<[u8]>),
 }
 
+impl Clone for Snapshot {
```
bartlomieju, line 56: Why is it needed?

runtime/web_worker.rs
```diff
@@ -442,7 +445,7 @@ impl WebWorker {
 
     let mut js_runtime = JsRuntime::new(RuntimeOptions {
       module_loader: Some(options.module_loader.clone()),
-      startup_snapshot: Some(js::deno_isolate_init()),
+      startup_snapshot: options.startup_snapshot.clone(),
```
bartlomieju, line 448: Use `.take()` instead of clone?

runtime/worker.rs
```diff
@@ -154,14 +155,16 @@ impl MainWorker {
...
+      js::init(),
       // Permissions ext (worker specific state)
       perm_ext,
     ];
     extensions.extend(std::mem::take(&mut options.extensions));
 
     let mut js_runtime = JsRuntime::new(RuntimeOptions {
       module_loader: Some(options.module_loader.clone()),
-      startup_snapshot: Some(js::deno_isolate_init()),
+      startup_snapshot: options.startup_snapshot.clone(),
```
bartlomieju, line 167: Ditto

snapshots/build_runtime.rs
```diff
@@ -0,0 +1,168 @@
...
+  let js_runtime = JsRuntime::new(RuntimeOptions {
+    will_snapshot: true,
+    extensions,
+    ..Default::default()
+  });
+  write_runtime_snapshot(js_runtime, snapshot_path);
+}
+
+// TODO(bartlomieju): this module contains a lot of duplicated
+// logic with `cli/build.rs`, factor out to `deno_core`.
```
bartlomieju, line 63: Seems like this comment is no longer valid?

bartlomieju (commented):
@AaronO can you provide some rationale for this change? How does it impact build times and CI times?

bartlomieju (approved):
LGTM

littledivy (approved):

PR #14796, opened 2022-06-04
Poc/snaphack

AaronO:
A PoC demonstrating the feasibility of a sidecar binary to produce snapshots when cross-compiling


msarahan:
@aslilac can this get some of your time any time soon? I would like to package deno for conda-forge (https://conda-forge.org/), but I am restricted on the CI services that are available there. Cross-compiling is the only way I'll be able to get packages built there. Solving this snapshot issue would be very helpful for unblocking my efforts.

This sidecar binary seems like a good approach. I initially tried to refactor the snapshot generation out of cli/build.rs, such that the snapshot could be generated and loaded from a file outside of the build process. The approach here is definitely cleaner than my efforts to refactor cli/build.rs, moving much of it into core/runtime.ts.

I'm not a typescript expert, but I would like to help here if there are ways for me to take work off your hands.

littledivy:
Closing because stale 🧹

IIRC this idea was also explored in https://github.com/denoland/deno/pull/14794 but then got reverted due to cargo publish issues.

PR #14808, opened 2022-06-06
refactor(core): SourceMapGetter doesn't need Send + Sync

bartlomieju:
Removes unnecessary trait bounds for `SourceMapGetter` trait.


dsherret (approved):
LGTM

Issue #14280, opened 2022-04-14
Uncaught websocket error - bad resource ID


jakeg:
```
error: Uncaught (in promise) BadResource: Bad resource ID
    at async deno:ext/websocket/01_websocket.js:465:11
```

Running on latest deno, version 1.20.5 so it's coming from this line:

https://github.com/denoland/deno/blob/d2c09408380191721ba7dfd971f802f5639eb1f7/ext/websocket/01_websocket.js#L465

I'm assuming maybe I'm trying to send a message to a websocket that's just closed?

crowlKats:
@jakeg do you have reproducible code?

jakeg:
Afraid not easily. I managed to get it to happen twice while doing some throttling via Chrome's dev tools. Assume it's related to a timeout.

I assume related to #14283 ?

crowlKats:
It might be related, though not fully sure yet. will look into it once I am back

jtoppine:
Are you sure you are catching all socket send() calls? This happened to me a lot, especially on "production" (demo) servers. I just made sure to `try { await ws.send(...); } catch(er) { ... };` everywhere, and it seems to have solved those issues. (tbh not even sure if send is async, the await may not be needed).

Indeed it appears to happen when sending to a socket that is already closed or maybe not yet acknowledged to be closed, but in unresponsive state - like when the client internet connection is dropped or unstable. It can often be triggered by repeatedly, aggressively refreshing a client browser, or detaching a physical cable connection. I might be wrong but I suspect it may happen also when a client machine goes standby, which may briefly leave the connection in limbo, unlike when a browser or tab is closed, which may result in more graceful connection closure.  Just some observations and guesses I have made trying to track down these very undeterministic issues for some time.

crowlKats:
> tbh not even sure if send is async, the await may not be needed

It isn't, the `WebSocket` API isn't promise based, but event based. Any errors caused asynchronously should be emitted through the `error` event.

jakeg:
This is now a recurring problem for me in production. Yes, I'm doing a `try { ... } catch { ...}` around any `socket.send()`.

crowlKats:
@jakeg sorry about this, I am currently looking into the stockpile of websocket issues. I just merged a PR that fixes a bunch of issues, could you check against a canary build, and tell me if it still persists?

jakeg:
[...]

PR #14813, opened 2022-06-07
fix(http/upgradewebsocket): check for open state for idle timeout

crowlKats:
might be a fix for #14280. currently untestable, very flaky behaviour.


bartlomieju (approved):
LGTM, discussed offline, we have no way to test this because `tungstenite` always responds to ping/pong messages

Issue #13934, opened 2022-03-13
Panic when creating a worker with Deno namespace permissions from a worker without them


andreubotella:
```js
// main.js
new Worker(
    new URL("./worker1.js", import.meta.url).href,
    { type: "module" }
);
```

```js
// worker1.js
console.log("Worker 1");
new Worker(
    new URL("./worker2.js", import.meta.url).href,
    { type: "module", deno: { namespace: true } }
);
```

```js
// worker2.js
console.log("Worker 2");
```

```
$ deno run --allow-read --unstable main.js
Worker 1

============================================================
Deno has panicked. This is a bug in Deno. Please report this
at https://github.com/denoland/deno/issues/new.
If you can reliably reproduce this panic, include the
reproduction steps and re-run with the RUST_BACKTRACE=1 env
var set and include the backtrace in your report.

Platform: linux x86_64
Version: 1.19.3
Args: ["deno", "run", "--allow-read", "--unstable", "main.js"]

thread 'worker-0' panicked at 'Worker has access to OS ops but exit code was not passed.', runtime/web_worker.rs:420:52
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This panic happens because worker 1 doesn't have access to the Deno namespace, but it's trying to start worker 2 with Deno namespace permissions. Creating the worker should instead throw an error in that case.

PR #14814, opened 2022-06-08
refactor: ensure exit code reference is passed to all workers

bartlomieju:
Closes https://github.com/denoland/deno/issues/13934
Prerequisite for https://github.com/denoland/deno/pull/14787


runtime/ops/os.rs
```diff
@@ -105,12 +103,12 @@ fn op_delete_env(state: &mut OpState, key: String) -> Result<(), AnyError> {
 
 #[op]
 fn op_set_exit_code(state: &mut OpState, code: i32) {
-  state.borrow_mut::<Arc<AtomicI32>>().store(code, Relaxed);
+  state.borrow_mut::<ExitCode>().0.store(code, Relaxed);
```
dsherret, line 106: The `Relaxed` could be buried in some implementation methods of `ExitCode` rather than being spread around in multiple files.

- `state.borrow_mut::<ExitCode>().set(code);`
- `let exit_code = state.borrow_mut::<ExitCode>().get();`

dsherret (approved):
LGTM, but consider making the interal `Arc<AtomicI32>` on `ExitCode` private.

PR #14815, opened 2022-06-08
chore(bench): generalized HTTP benchmarks framework

littledivy:
Makes it easier to add new wrk benchmarks.

- auto discoverable benchmarks sources. Just place a JS file in `cli/bench/http` and `cargo bench` will collect it.
- configuring `wrk` with lua scripts. (eg: `deno_http.lua` for `deno_http.js`)

Future improvements:
- configure wrk command with comment directives in JS


lucacasonato (approved):
Yay!

PR #14816, opened 2022-06-08
perf(core): Cache source lookups

nayeemrmn:
Keep a cache for source maps and source lines. We sort of already had a cache argument for source map lookup functions but we just passed an empty map instead of storing it (even before moving to core). I extended it to cache source line lookups as well and plugged it into runtime state.

cc @bartlomieju This runtime state cache extends beyond caching across frame lookups for a single stack, and we don't need to accept a list of locations in `op_apply_source_map`. ([from discord](https://discord.com/channels/684898665143206084/684911491035430919/983518187570229369))

### DIFF
--- a/core/source_map.rs
+++ b/core/source_map.rs
@@ -37,8 +37,6 @@ pub fn apply_source_map<G: SourceMapGetter + ?Sized>(
   let default_pos = (file_name.clone(), line_number, column_number);
   let maybe_source_map =
     cache.maps.entry(file_name.clone()).or_insert_with(|| {
-      // TODO(kitsonk) parsed source maps should probably be cached in state in
-      // the module meta data.
       getter
         .get_source_map(&file_name)
         .and_then(|raw_source_map| SourceMap::from_slice(&raw_source_map).ok())


nayeemrmn:
> I'm not sure what will be the memory impact, but we should be careful about that.

We already cache all sources in memory, this makes it also cache the source map for the subset of those which are referenced in any stack trace. I think programs have a finite and small amount of unique rendered stack traces. With the source line cache I made sure to move the 150 character cap to core so we don't cache anything ridiculous like from minified code.

bartlomieju:
> > I'm not sure what will be the memory impact, but we should be careful about that.
> 
> We already cache all sources in memory, this makes it also cache the source map for the subset of those which are referenced in any stack trace. I think programs have a finite and small amount of unique rendered stack traces. With the source line cache I made sure to move the 150 character cap to core so we don't cache anything ridiculous like from minified code.

Do we actually cache source maps in memory? Could you point me to relevant code?

nayeemrmn:
> Do we actually cache source maps in memory?

We cache _sources_ already, I'm saying adding the source map for a small (and ~constant IMO) subset of those wouldn't be a big deal

core/error.rs
```diff
@@ -192,15 +192,20 @@ impl JsError {
...
-    let state_rc = JsRuntime::state(scope);
-    let state = state_rc.borrow();
-    if let Some(format_exception_cb) = &state.js_format_exception_cb {
-      let format_exception_cb = format_exception_cb.open(scope);
-      let this = v8::undefined(scope).into();
-      let formatted = format_exception_cb.call(scope, this, &[exception]);
-      if let Some(formatted) = formatted {
-        if formatted.is_string() {
-          exception_message = Some(formatted.to_rust_string_lossy(scope));
+    // Nest this state borrow. A mutable borrow can occur when accessing `stack`
```
bartlomieju, line 195: What do you mean by "next this state borrow"?

core/source_map.rs
```diff
@@ -17,83 +17,77 @@ pub trait SourceMapGetter {
...
-    match maybe_source_map {
-      None => default_pos,
-      Some(source_map) => {
-        match source_map.lookup_token(line_number as u32, column_number as u32)
-        {
+  let default_pos = (file_name.clone(), line_number, column_number);
+  let maybe_source_map =
+    cache.maps.entry(file_name.clone()).or_insert_with(|| {
+      // TODO(kitsonk) parsed source maps should probably be cached in state in
+      // the module meta data.
```
bartlomieju, line 41: This comment can now be removed

core/error.rs
nayeemrmn: `nest`, just explaining why the block scope is there.

core/source_map.rs
nayeemrmn: I think this comment means caching them in FS in the `.meta` file

core/error.rs
bartlomieju: Ooops, typo. I meant "nest". This is fine then

core/source_map.rs
bartlomieju: Parsed implies that JSON representation is converted to `SourceMap`. I would suggest to remove this comment after all

bartlomieju:
@nayeemrmn this PR looks fine to me, there's one more quirk that would be great to be addressed.

In we're converting source code to a string, then iterating over the lines to find a source map and create another string out of it to be decoded. I believe all of this could be done on a vector of bytes instead to avoid many round-trip in string <-> bytes conversion. Could you look into that?
https://github.com/denoland/deno/blob/60869c2598321386f57e55537e8c99ed011fbb95/cli/proc_state.rs#L759-L777

bartlomieju (approved):
LGTM, thank you @nayeemrmn, great clean up

PR #14819, opened 2022-06-08
chore(scripts): always run workflows against Deno's most recent version.

lucasmichot:
`denoland/setup-deno` allows to install the [most recent stable version](https://github.com/denoland/setup-deno#latest-stable-for-a-major).

It seems to me that it makes sense to use this here as:

1. this notation is already used across most `denoland` organization
2. this enforces always testing against the most stable version
3. frequent update of the wokflow files can be avoided


bartlomieju (approved):
Thanks!

PR #14821, opened 2022-06-08
refactor(bench): continuous benchmarking improvements

littledivy:
This commit improves Deno's benchmarking workflow like:

- benchmark filtering. `cargo bench --bench deno_bench -- bundle`
- cron workflow to run benchmarks.
- upload metrics to a GCS bucket anhd use grafana for visualization instead of `deno.land/benchmarks`


littledivy:
https://datastudio.google.com/reporting/7778e3b2-5a8b-429f-9f32-6e6dfde4c58f

.github/workflows/bench_cron.yml
```diff
@@ -0,0 +1,42 @@
+name: bench
+
+on:
+  schedule:
+    - cron: 0 0,6,12 * * *
```
ry, line 5: Add a comment describing this in english. Also maybe use 0,6,12,18.

.github/workflows/ci.yml
```diff
@@ -496,26 +496,6 @@ jobs:
...
-              gh-pages
-          ./target/release/deno run --allow-all --unstable \
-              ./tools/build_benchmark_jsons.js --release
-          cd gh-pages
-          git config user.email "propelml@gmail.com"
-          git config user.name "denobot"
-          git add .
-          git commit --message "Update benchmarks"
-          git push origin gh-pages
-
```
ry, line 518: Is it possible for both of these to live simultaneously for some time? How much work would that be. I'd rather not make a hot switch all at once.

ry (commented):
Update the cli/bench/README.md file with examples of how to run it.

.github/workflows/ci.yml
littledivy: Added back

Cargo.lock
```diff
@@ -4977,16 +5103,27 @@ version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a156c684c91ea7d62626509bce3cb4e1d9ed5c4d978f7b4352658f96a4c26b4a"
 
+[[package]]
```
bartlomieju, line 5106: It's concerning how many changed happened in this file... Is there a way we could reuse the same version (for example by `cargo update --package` that imports this version)

bartlomieju: Thanks

PR #14823, opened 2022-06-08
feat(task): add `--cwd` flag for configuring the working directory

### LINKED ISSUE #14820: Ability to set start cwd of deno task
`deno task` has a characteristic that if you execute a task anywhere in the directory tree it will use the deno.json's dir as the cwd. This is nice because you can define tasks that use relative paths and have the tasks work when you execute them from any directory.

Sometimes this is not nice though and for those situations there should be an opt-in way to set the cwd that a deno task uses when it starts up. For example, the following is not possible at the moment:

```jsonc
{
  "tasks": {
    // doesn't work, because `deno task wasmbuild` will start in this deno.json's directory
    "build:crypto": "cd _wasm_crypto && deno task wasmbuild",
    "build:hash": "cd hash/_wasm && deno task wasmbuild",
    "build:varint": "cd _wasm_varint && deno task wasmbuild",
    "wasmbuild": "deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts"
  }
}
```

Instead you need to do the following, which creates multiple sources of truth:

```jsonc
{
  "tasks": {
    "build:crypto": "cd _wasm_crypto && deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts",
    "build:hash": "cd hash/_wasm && deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts",
    "build:varint": "cd _wasm_varint && deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts"
  }
}
```

To fix this, it would be useful to have a way to set the cwd of a deno task. Probably a `--cwd <path>` option:

```jsonc
{
  "tasks": {
    "build:crypto": "deno task --cwd _wasm_crypto wasmbuild",
    "build:hash": "deno task --cwd hash/_wasm wasmbuild",
    "build:varint": "deno task --cwd _wasm_varint wasmbuild",
    "wasmbuild": "deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts"
  }
}
```

People could also do stuff like `deno task --cwd $PWD wasmbuild` in order to use the cwd as the starting point of a deno task.

Ref: https://github.com/denoland/deno_std/pull/2319#discussion_r892434783

aslilac:
Fixes #14820

Adds a `--cwd` flag to `deno task` that can control the directory the task will be executed in

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -163,6 +163,7 @@ pub struct RunFlags {
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct TaskFlags {
+  pub cwd: Option<String>,
   pub task: String,
 }
 
@@ -1465,6 +1466,14 @@ fn task_subcommand<'a>() -> Command<'a> {
   Command::new("task")
     .trailing_var_arg(true)
     .args(config_args())
+    .arg(
+      Arg::new("cwd")
+        .long("cwd")
+        .value_name("DIR")
+        .help("Specify the directory to run the task in")
+        .takes_value(true)
+        .value_hint(ValueHint::DirPath)
+    )
     // Ideally the task name and trailing arguments should be two separate clap
     // arguments, but there is a bug in clap that's preventing us from doing
     // this (https://github.com/clap-rs/clap/issues/1538). Once that's fixed,
@@ -2567,7 +2576,15 @@ fn task_parse(
 ) {
   config_args_parse(flags, matches);
 
-  let mut task_name = "".to_string();
+  let mut task_flags = TaskFlags {
+    cwd: None,
+    task: String::new(),
+  };
+
+  if let Some(cwd) = matches.value_of("cwd") {
+    task_flags.cwd = Some(cwd.to_string());
+  }
+
   if let Some(mut index) = matches.index_of("task_name_and_args") {
     index += 1; // skip `task`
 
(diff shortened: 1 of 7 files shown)


dsherret (approved):
Looks great. Thanks @aslilac!

cli/flags.rs
```diff
@@ -163,6 +163,7 @@ pub struct RunFlags {
 
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct TaskFlags {
+  pub cwd: Option<String>,
```
bartlomieju, line 166: Consider using `Option<PathBuf>` instead?

dsherret: The config flag uses `Option<String>` 🤷‍♂️

ry:
I think we should be cautious about adding new flags. Otherwise we're going to end up in a situation where there are many many thousands of flags to learn about.
```
{
  "tasks": {
    "build:crypto": "deno task --cwd _wasm_crypto wasmbuild",
    "build:hash": "deno task --cwd hash/_wasm wasmbuild",
    "build:varint": "deno task --cwd _wasm_varint wasmbuild",
    "wasmbuild": "deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts"
  }
}
```
Not sure how much better that is than
```
{
  "tasks": {
    "build:crypto": "cd _wasm_crypto && deno task wasmbuild",
    "build:hash": "cd hash/_wasm && deno task wasmbuild",
    "build:varint": "cd _wasm_varint && deno task wasmbuild",
    "wasmbuild": "deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts"
  }
}
```

If we're really just making a `make` replacement maybe just use the same flags? This is called `-C` in make. 

I'm skeptical that Deno should be implementing a shell interpreter and make replacement.

aslilac:
It'd also be pretty simple to add a `-C` short form

PR #14824, opened 2022-06-08
feat: make `deno vendor` work with an existing import map by modifying it

### LINKED ISSUE #13697: `deno vendor` does not work with existing import map with bare specifier mapping
```
~/p/g/d/d/test ❯❯❯ cat import_map.json
{
  "imports": {
    "std/": "https://deno.land/std@0.125.0/"
  }
}
~/p/g/d/d/test ❯❯❯ cat main.ts 
import "std/examples/welcome.ts";
~/p/g/d/d/test ❯❯❯ deno vendor --import-map import_map.json main.ts -f
Vendored 1 module into vendor/ directory.

To use vendored modules, specify the `--import-map` flag when invoking deno subcommands:
  deno run -A --import-map vendor/import_map.json main.ts
~/p/g/d/d/test ❯❯❯ deno run -A --import-map vendor/import_map.json main.ts
error: Relative import path "std/examples/welcome.ts" not prefixed with / or ./ or ../ and not in import map from "file:///mnt/starship/Projects/github.com/denoland/dotcom/test/main.ts"
    at file:///mnt/starship/Projects/github.com/denoland/dotcom/test/main.ts:1:8
```

dsherret:
This is based on an idea from @rtpg:

1. Moves the import_map.json to be in the root directory instead of vendor folder.
2. Specific usages of remote dependencies in local code is added to the "imports" (ex. `"https://localhost/mod.ts": "./vendor/localhost/mod.ts"`) instead of a general one (ex. `"https://localhost/": "./vendor/localhost/"`)... so when a new remote specifier is used then `deno vendor` needs to be run again.
3. Strips all vendor output folder entries from the import map before vendoring in order to allow re-vendoring.

Closes #13697


dsherret:
This sadly does not work because we need to modify import map entries like `"$twind": "https://esm.sh/twind@0.16.16"` to point to the vendored location. We're going to continue using a `vendor/import_map.json` file. I will open a new PR that makes these changes.

Issue #14818, opened 2022-06-08
Disallow "Deno.exit()" inside web worker context


bartlomieju:
@lucacasonato proposed we disallow `Deno.exit()` API in web worker context, or more precisely throw on call to that API to keep backward compatibility by not removing the API. One of the reasons is that in browsers there's no way to close main window from within the worker. Users could still achieve that by sending a message to main worker that could handle it and call `Deno.exit()`, example:
```js
// worker.js

self.onmessage = (e) => {
   // something going on here...
   if (someCondition) {
      self.postMessage("exit");
   }
};
```

```js
// main.js

const worker = new Worker("./worker.js", { type: "module" });

// something going on here

worker.onmessage = (e) => {
  if (e.data == "exit") {
    Deno.exit(10);
  }
};
```

If user tries to call `Deno.exit()` in worker context we would then throw an error like: `Deno.exit() is not allowed in worker context`.

This change would greatly simplify implementing following issues:
- https://github.com/denoland/deno/pull/14814
- https://github.com/denoland/deno/issues/7590
- https://github.com/denoland/deno/issues/13093
- https://github.com/denoland/deno/issues/8713
- https://github.com/denoland/deno/issues/12653
- https://github.com/denoland/deno/issues/14795

cjihrig:
In Node.js, `process.exit()` does not exit the whole process, only that worker thread. I'm +1 to either doing what Node does or throwing an error as suggested here.

PR #14826, opened 2022-06-09
Deno.exit() is an alias to self.close() in worker contexts

bartlomieju:
Closes https://github.com/denoland/deno/issues/14818

<s>This is technically a breaking change. If we want to be more conservative we could alias `Deno.exit()` to `self.terminate()` in worker contexts similarly to Node.js as pointed by Colin below.</s>

This commit changes `Deno.exit()` to be an alias to `self.close()` in worker contexts,
and setting an exit code becomes a no-op in such case.


bartlomieju:
@cjihrig is `process.exitCode` forwarded from the worker to main context in Node? I just noticed that I added such a case, but I'm not sure if that's the correct behavior.

cjihrig:
> setting an exit code becomes a no-op in such case

This part seems like it could confuse users. Is it possible to delete this functionality in a Worker context (or have it be a stub that throws, or show a warning, etc.)? For example, in Node Workers, some functionality gets [stubbed out like this](https://github.com/nodejs/node/blob/78b410ff4bfb9a967668a6895b0f55c82eab3d49/lib/internal/bootstrap/switches/does_not_own_process_state.js#L8-L9).

bartlomieju:
> > setting an exit code becomes a no-op in such case
> 
> This part seems like it could confuse users. Is it possible to delete this functionality in a Worker context (or have it be a stub that throws, or show a warning, etc.)? For example, in Node Workers, some functionality gets [stubbed out like this](https://github.com/nodejs/node/blob/78b410ff4bfb9a967668a6895b0f55c82eab3d49/lib/internal/bootstrap/switches/does_not_own_process_state.js#L8-L9).

I can throw an error in `Deno.core.opSync("op_set_exit_code");` but it's not really a public API, OTOH the code you pass to `Deno.exit()` is not saved/passed in anyway (ie. it's ignored). What should I adjust?

cjihrig:
Hm, if it's not intended to be public API and the worker itself doesn't have an "exit code" then it's probably fine to leave as is, and just document the behavior.

bartlomieju:
> Hm, if it's not intended to be public API and the worker itself doesn't have an "exit code" then it's probably fine to leave as is, and just document the behavior.

I'll update the manual accordingly and mention it in a blog post.

dsherret (approved):

PR #14827, opened 2022-06-09
docs: Improve `mod_evaluate` documentation

cd-work:
All existing documentation and examples for `mod_evaluate` seem to point
towards its return value being relatively insignificant, mostly just
discarding it directly. However evaluating that result is necessary to
validate execution success.

To help users understand how execution failure should be handled and in
which order these futures must be awaited to avoid stalling, the
examples have been updated and some additional documentation on
`mod_evaluate` was added.

In the spirit of incremental improvement, some existing documentation
links have also been driveby-fixed.


bartlomieju:
@cd-work please sign the CLA

cd-work:
> @cd-work please sign the CLA

Will do as soon as I get the okay.

Would you mind if I add the following patch to this PR, rather than opening a new one? It fits somewhat well because it's also a documentation/example improvement, but it's otherwise completely unrelated.

```diff
diff --git a/runtime/examples/hello_runtime.rs b/runtime/examples/hello_runtime.rs
index 07e42f0ff..1ce7cde4b 100644
--- a/runtime/examples/hello_runtime.rs
+++ b/runtime/examples/hello_runtime.rs
@@ -70,7 +70,5 @@ async fn main() -> Result<(), AnyError> {
     permissions,
     options,
   );
-  worker.execute_main_module(&main_module).await?;
-  worker.run_event_loop(false).await?;
-  Ok(())
+  worker.execute_main_module(&main_module).await
 }
```

The patched code should be more idiomatic as far as I can tell, since `execute_main_module` already calls `run_event_loop` internally, thus doing it twice is unnecessary. I'm not 100% certain on this though, because I'm not sure if there's any reason to call `run_event_loop` when module loading failed before execution was possible.

bartlomieju:
> > @cd-work please sign the CLA
> 
> Will do as soon as I get the okay.
> 
> Would you mind if I add the following patch to this PR, rather than opening a new one? It fits somewhat well because it's also a documentation/example improvement, but it's otherwise completely unrelated.
> 
> ```diff
> diff --git a/runtime/examples/hello_runtime.rs b/runtime/examples/hello_runtime.rs
> index 07e42f0ff..1ce7cde4b 100644
> --- a/runtime/examples/hello_runtime.rs
> +++ b/runtime/examples/hello_runtime.rs
> @@ -70,7 +70,5 @@ async fn main() -> Result<(), AnyError> {
>      permissions,
>      options,
>    );
> -  worker.execute_main_module(&main_module).await?;
> -  worker.run_event_loop(false).await?;
> -  Ok(())
> +  worker.execute_main_module(&main_module).await
>  }
> ```
> 
> The patched code should be more idiomatic as far as I can tell, since `execute_main_module` already calls `run_event_loop` internally, thus doing it twice is unnecessary. I'm not 100% certain on this though, because I'm not sure if there's any reason to call `run_event_loop` when module loading failed before execution was possible.

This is actually correct - `Worker::evaluate_module` returns result when the last top-level promise resolves, but there can be more asynchronous work to do (eg. code in the module started timeouts) that will be done only if event loop is run.

cd-work:
> This is actually correct - Worker::evaluate_module returns result when the last top-level promise resolves, but there can be more asynchronous work to do (eg. code in the module started timeouts) that will be done only if event loop is run.

Good thing I asked, that sounds interesting. By "top-level promise" I'd assume you mean that once the module itself is executed and awaited, `evaluate_module` will return, but `setTimeout(..., "1000")` inside that main module code might cause future side-effects that require running the event loop again until everything is done?

And I'd assume the `evaluate_module` function itself doesn't just await both because a user might actually want to run more stuff and it's unnecessary to await all futures when the goal is to just start running another module?

bartlomieju:
> Good thing I asked, that sounds interesting. By "top-level promise" I'd assume you mean that once the module itself is executed and awaited, evaluate_module will return, but setTimeout(..., "1000") inside that main module code might cause future side-effects that require running the event loop again until everything is done?

This is exactly right.

> And I'd assume the evaluate_module function itself doesn't just await both because a user might actually want to run more stuff and it's unnecessary to await all futures when the goal is to just start running another module?

Again, correct.

core/examples/fs_module_loader.rs
```diff
@@ -28,9 +28,9 @@ fn main() -> Result<(), Error> {
 
   let future = async move {
     let mod_id = js_runtime.load_main_module(&main_module, None).await?;
-    let _ = js_runtime.mod_evaluate(mod_id);
+    let result = js_runtime.mod_evaluate(mod_id);
     js_runtime.run_event_loop(false).await?;
-    Ok(())
+    result.await?
```
bartlomieju, line 33: To be completely previse, we should use Tokio's `select!` macro here, but this was supposed to be a simple example, so I guess we could punt on that

PR #14830, opened 2022-06-09
feat(web): add beforeunload event

cjihrig:
This commit adds the 'beforeunload' event.


bartlomieju:
I changed this PR to require `e.preventDefault()` to be called to keep running the event loop. I'm not sure if that's desirable though - it poses a challenge how to polyfill `process.on("beforeExit")` in Node compat layer - we can check if there are active listeners and prevent the event, but I'm worried that it could cause infinite loops. @lucacasonato I'm interested in hearing your thoughts on the topic of this API.

bartlomieju:
It appears there are no WPT tests that we can enable. @lucacasonato please take a look

bartlomieju (approved):

andreubotella:
Is there any reason why this PR added `op_event_loop_has_more_work` and `JsRuntime::event_loop_has_work`, given that they're not used in the PR?

nayeemrmn:
> Is there any reason why this PR added `op_event_loop_has_more_work` and `JsRuntime::event_loop_has_work`, given that they're not used?

The former to use in https://github.com/denoland/deno_std/pull/2331, I guess the latter was supposed to be used in the op but was forgotten..

andreubotella:
In that case, given that that same code shows up three times (in `JsRuntime::poll_event_loop` plus in those two functions), I think `event_loop_has_work` should return a struct of booleans which can be used by both the event loop and this op.

PR #14836, opened 2022-06-09
feat(vendor): support using an existing import map

dsherret:
Take 2 (initial attempt was #14824).

1. Keeps the import map in the `./vendor` folder.
2. Includes any necessary mappings from the original import map in the output.

Closes https://github.com/denoland/deno/issues/13697


cli/tools/vendor/mod.rs
```diff
@@ -33,29 +32,33 @@ pub async fn vendor(ps: ProcState, flags: VendorFlags) -> Result<(), AnyError> {
...
-      .unwrap_or("main.ts"),
   );
+  if vendored_count > 0 {
+    eprintln!(
+      concat!(
+        "\nTo use vendored modules, specify the `--import-map {}` flag when ",
+        r#"invoking deno subcommands or add an `"importMap": "<path_to_vendored_import_map>"` "#,
+        "entry to your deno.json file.",
+      ),
+      raw_output_dir.join("import_map.json").display(),
```
lucacasonato, line 59: Why is there an actual specifier for `--import-map`, but not for `<path_to_vendored_import_map>`?

lucacasonato (approved):
Tested this in three repositories that use import maps, and it works correctly in all three. Very nice!

cli/tools/vendor/mod.rs
dsherret: It's more complicated and I didn't think it was worth it. It would require checking where the config is and then doing different logic for whether the config existed or didn't exist.

bartlomieju:
@dsherret any chance we can land that for v1.23?

Issue #14846, opened 2022-06-11
Suggestion: Allow `--compat` to be used with `check` and `bundle` subcommands


arendjr:
Currently, the `--compat` option is only allowed in combination with the `run` command.

I would like to extend its use so it can also be passed to the `check` and `bundle` commands. This would allow people targeting Node.js to use Deno as a `tsc` replacement that is compatible with the ESM resolution as implemented by Deno.

bartlomieju:
> Currently, the `--compat` option is only allowed in combination with the `run` command.
> 
> I would like to extend its use so it can also be passed to the `check` and `bundle` commands. This would allow people targeting Node.js to use Deno as a `tsc` replacement that is compatible with the ESM resolution as implemented by Deno.

I'm pretty sure adding `--compat` to `deno check` won't be enough to make it a replacement for `tsc` - Deno has different built-in libraries and doesn't allow numerous options for TS configuration that are popular in Node world. 

`--compat` in `deno bundle` might make sense, but we need to discuss how it would behave - should `deno_std/node` be included in such the output bundle or not?

`--compat` is meant to allow running code authored for Node.js, but it hasn't worked in terms of type checking.

arendjr:
Yeah, I realize it wouldn’t be a perfect replacement yet, but I think it might be in the direction, even if more work would ideally be done as a follow-up. I was creating a small PoC to see how feasible it would be to use Deno as a `tsc` replacement and I think for that small PoC, the lack of `--compat` was the main blocker, but yeah, real projects would have more requirements.

Out of curiosity, what options that `tsc` supports do you think would be most important for such a use case?

One more I was thinking of was a `--no-std` option for `deno check`. This would then disable the `Deno` namespace, which would not only be useful for people targeting Node, but also people that want to target both Deno and web with a single codebase.

arendjr:
> --compat in deno bundle might make sense, but we need to discuss how it would behave - should deno_std/node be included in such the output bundle or not?

I asked exactly this question in the draft PR :) https://github.com/denoland/deno/pull/14848

arendjr:
> --compat is meant to allow running code authored for Node.js, but it hasn't worked in terms of type checking.

It actually worked for my PoC (happy to publish it, if you like), but indeed with the caveat the user is now responsible for making sure they don’t accidentally use Deno APIs in their codebase. But that’s something an additional `--no-std` option could solve.

[...]

PR #14847, opened 2022-06-11
feat(compat): Allow --compat to be used with check command

arendjr:
Fixes the first part of https://github.com/denoland/deno/issues/14846.


bartlomieju:
@arendjr thank you for this PR, but I'm going to close it without a merge. We're removing `--compat` mode in https://github.com/denoland/deno/pull/15678 and using `npm:` specifiers is the way forward for npm compatibility. Thanks for the work and sorry we can't merge it.

PR #14848, opened 2022-06-11
feat(compat): Allow --compat to be used with bundle command

arendjr:
This is the second part of https://github.com/denoland/deno/issues/14846.

The test won't succeed yet, because while this PR allows for passing `--compat` now, the generated bundle contains Deno's `std/node` library, which I think should be omitted for this use case, so that the bundle will become runnable on a real Node runtime. This means leaving the imports to Node's built-ins unaltered in the generated bundle, but I haven't found out exactly how to accomplish this. A pointer by one of the Deno maintainers on how to achieve this would be appreciated. Thanks!

### DIFF
--- a/cli/compat/esm_resolver.rs
+++ b/cli/compat/esm_resolver.rs
@@ -25,6 +25,10 @@ impl NodeEsmResolver {
       maybe_import_map_resolver,
     }
   }
+
+  pub fn as_resolver(&self) -> &dyn Resolver {
+    self
+  }
 }
 
 impl Resolver for NodeEsmResolver {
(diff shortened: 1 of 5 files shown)


arendjr:
Also I’d be curious to hear if we should introduce another CLI flag (`--omit-node-std`, maybe?) so that users can choose whether the `std/node` lib is included or not. If it’s included, the resulting bundle should work in Deno (without needing to pass the `--compat` flag anymore). Without it, it should work in Node.

bartlomieju:
@arendjr thank you for this PR, but I'm going to close it without a merge. We're removing `--compat` mode in https://github.com/denoland/deno/pull/15678 and using `npm:` specifiers is the way forward for npm compatibility. Thanks for the work and sorry we can't merge it.

PR #14860, opened 2022-06-13
refactor: Add default implementation for WorkerOptions

cd-work:
This adds an implementation of `Default` for both `WorkerOptions` and
`BootstrapOptions`. Since both of these structs are rather big, this
should make it easier for people unfamiliar with the internals to focus
on the options relevant to them.

As a user of `deno_runtime` I feel like these should serve as good
defaults, getting people them started without having to tweak the
runtime. Additionally even if some changes are made, the usage of
`..Default::default()` will significantly help with code clarity and
verbosity.

---

Some of these defaults might be somewhat questionable, but I think the convenience of `..Default::default()` still outweighs that. However there is the option of using a builder pattern, which would allow to still have **some** mandatory options while the rest could be relied upon as fallbacks.

I think the majority of `deno_runtime` consumers would rely on the defaults of most `WorkerOptions`, so having *some* methodology that doesn't require setting every single one of them makes sense to me and would likely clean up consumer code significantly.


runtime/worker.rs
```diff
@@ -83,6 +84,32 @@ pub struct WorkerOptions {
   pub stdio: Stdio,
 }
 
+impl Default for WorkerOptions {
+    fn default() -> Self {
+        Self {
+            web_worker_preload_module_cb: Arc::new(|_| unimplemented!("web workers are not supported")),
+            create_web_worker_cb: Arc::new(|_| unimplemented!("web workers are not supported")),
```
cd-work, line 91: Crashing from a library is certainly always problematic, but this basically seems to be the recommended implementation for `deno_runtime` consumers based on what the CLI does.

runtime/worker.rs
```diff
@@ -83,6 +84,32 @@ pub struct WorkerOptions {
...
+            module_loader: Rc::new(FsModuleLoader),
```
cd-work, line 92: Currently a module loader is required. It might be possible to just transform this field into an `Option` and use `Default::default()` here, then do the translation to the default loader further down the line. But I think the `FsModuleLoader` should work fine for most and `deno_core` will use a default loader anyway if it's not specified.

runtime/worker.rs
```diff
@@ -83,6 +84,32 @@ pub struct WorkerOptions {
...
+            seed: None,
```
cd-work, line 93: I've explicitly specified seed as `None` since it could potentially be a really bad idea if its type is changed and it just falls back to a numeric default like `0`.

runtime/worker_bootstrap.rs
```diff
@@ -23,6 +27,31 @@ pub struct BootstrapOptions {
   pub user_agent: String,
 }
 
+impl Default for BootstrapOptions {
+  fn default() -> Self {
+    let cpu_count = thread::available_parallelism()
+        .map(|p| p.get())
+        .unwrap_or(1);
+
+    let runtime_version = env!("CARGO_PKG_VERSION").into();
```
cd-work, line 36: Having the runtime version the version of the `deno_runtime` crate is probably a better idea than whatever the consumer can specify here.

runtime/worker_bootstrap.rs
```diff
@@ -23,6 +27,31 @@ pub struct BootstrapOptions {
...
+    let user_agent = format!("Deno/{}", runtime_version);
```
cd-work, line 37: Might want to change this to `Deno-runtime` or something, but I think most people would likely be fine with just a generic `Deno` user agent if they don't care much about the defaults.

cd-work (commented):
Explaining my reasoning behind the defaults chosen here:

bartlomieju:
Interesting PR, I think we could land as is (or with some minor changes). Please make sure your code your code passes lint and format.

mlafeldt:
Personally, I'd love to see this being merged. 👍

Having no default makes creating workers relatively cumbersome. As an example, in the case of this demo, it makes up over half of the code:

https://github.com/denoland/roll-your-own-javascript-runtime/compare/main...mlafeldt:roll-your-own-javascript-runtime:deno_runtime

All defaults except for `runtime_version` make sense to me. I don't know how it's being used, but an empty string or "n/a" might be less confusing than a wrong version (Deno vs deno_runtime)?

mlafeldt:
Would also reduce the boilerplate in tests and examples, e.g.

- https://github.com/denoland/deno/blob/25a1cc1b28ee2f37deb6305c0feb66d4eec50804/runtime/worker.rs#L402
- https://github.com/denoland/deno/blob/25a1cc1b28ee2f37deb6305c0feb66d4eec50804/runtime/examples/hello_runtime.rs#L29

bartlomieju:
@cd-work could you please rebase this PR? I think we should land it after all

cd-work:
I've rebased it, I'm not sure the CI failure is related to this PR?

bartlomieju:
> I've rebased it, I'm not sure the CI failure is related to this PR?

It's not, CI should pass now

PR #14864, opened 2022-06-14
BREAKING: remove `Intl.v8BreakIterator`

lucacasonato:
This is a non-standard API that is mostly replaced by `Intl.Segmenter`.

### DIFF
--- a/runtime/js/99_main.js
+++ b/runtime/js/99_main.js
@@ -1,9 +1,12 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+"use strict";
+
 // Removes the `__proto__` for security reasons.  This intentionally makes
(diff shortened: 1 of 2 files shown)


runtime/js/99_main.js
```diff
@@ -1,10 +1,13 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+"use strict";
+
 // Removes the `__proto__` for security reasons.  This intentionally makes
 // Deno non compliant with ECMA-262 Annex B.2.2.1
```
petamoriken, line 5: Since `Object.prototype._proto__` has been moved from Annex B in the current spec, it would be a good time to update the comments while we are at it.
https://tc39.es/ecma262/#sec-object.prototype.__proto__

PR #14865, opened 2022-06-14
refactor(runtime/signal): revert SIGINT and SIGBREAK `Deno.kill` Windows changes

dsherret:
Reverts the `Deno.kill` changes in https://github.com/denoland/deno/pull/14694/files

FYI @GJZwiers - I tried to [add some tests](https://gist.github.com/dsherret/89d05d333bb8dca3475b257d48bec269) for that PR, but got unexpected behaviour due to: https://github.com/denoland/deno/pull/14694/files#r885888645 so I think we should revert the `Deno.kill` changes in that PR for now.

Opened #14866

### DIFF
--- a/cli/tests/unit/signal_test.ts
+++ b/cli/tests/unit/signal_test.ts
@@ -204,10 +204,10 @@ Deno.test(
   function windowsThrowsOnNegativeProcessIdTest() {
     assertThrows(
       () => {
-        Deno.kill(-1, "SIGINT");
+        Deno.kill(-1, "SIGKILL");
       },
       TypeError,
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):

GJZwiers:
There was a feature request over at Node at some point for this behaviour, because it would allow sending a SIGINT to a (parent) process and all its subprocesses, which could have some uses. But after re-reading a bit I see that I missed a couple of points later on in the discussion where they retracted the proposal to add it to Node core, because of the problem that it would send the signal to all processes attached to the same console. So I agree with the revert.

Issue #14171, opened 2022-04-01
Logging a URL in an array (or object) prints without consistent tabbing


fabiancook:
## Code

```typescript
console.log([new URL("https://example.com/path")])
console.log({ url: new URL("https://example.com/path") })
console.log({ url: [new URL("https://example.com/path")] })
```

## Expected

```shell
[
  URL {
    href: "https://example.com/path",
    origin: "https://example.com",
    protocol: "https:",
    username: "",
    password: "",
    host: "example.com",
    hostname: "example.com",
    port: "",
    pathname: "/path",
    hash: "",
    search: ""
  }
]
{
  url: URL {
    href: "https://example.com/path",
    origin: "https://example.com",
    protocol: "https:",
    username: "",
    password: "",
    host: "example.com",
    hostname: "example.com",
    port: "",
    pathname: "/path",
    hash: "",
    search: ""
  }
}
{
  url: [
    URL {
      href: "https://example.com/path",
      origin: "https://example.com",
      protocol: "https:",
      username: "",
      password: "",
      host: "example.com",
      hostname: "example.com",
      port: "",
      pathname: "/path",
      hash: "",
      search: ""
    }
  ]
}
```

## Node

```shell
[
  URL {
    href: 'https://example.com/path',
    origin: 'https://example.com',
    protocol: 'https:',
    username: '',
    password: '',
    host: 'example.com',
    hostname: 'example.com',
    port: '',
    pathname: '/path',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  }
]
{
  url: URL {
    href: 'https://example.com/path',
    origin: 'https://example.com',
    protocol: 'https:',
    username: '',
    password: '',
    host: 'example.com',
    hostname: 'example.com',
    port: '',
    pathname: '/path',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  }
}
{
  url: [
    URL {
      href: 'https://example.com/path',
      origin: 'https://example.com',
      protocol: 'https:',
      username: '',
      password: '',
      host: 'example.com',
      hostname: 'example.com',
      port: '',
      pathname: '/path',
      search: '',
      searchParams: URLSearchParams {},
      hash: ''
    }
  ]
}
```

## Actual

```shell
[
  URL {
  href: "https://example.com/path",
  origin: "https://example.com",
  protocol: "https:",
  username: "",
  password: "",
  host: "example.com",
  hostname: "example.com",
  port: "",
  pathname: "/path",
  hash: "",
[...]

PR #14867, opened 2022-06-14
fix(url): properly indent when inspecting URLs

cjihrig:
This PR contains two commits, which I think should land separately (unsquashed):

##### fix(console): constrol inspect() indent with option

This commit updates the `Deno.inspect()` logic to use the
(currently unused) `indentLevel` option to control indentation instead
of passing around separate indent/level parameters internally.

Refs: https://github.com/denoland/deno/issues/8099
Refs: https://github.com/denoland/deno/issues/14171

##### fix(url): properly indent when inspecting URLs

This commit updates the custom inspect function for URL objects
to pass the inspect options through so that the context is
propagated and the resulting indentation is correct.

Fixes: https://github.com/denoland/deno/issues/14171

### DIFF
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -2704,3 +2704,8 @@ itest!(error_name_non_string {
   output: "error_name_non_string.js.out",
   exit_code: 1,
 });
+
+itest!(custom_inspect_url {
+  args: "run custom_inspect_url.js",
+  output: "custom_inspect_url.js.out",
+});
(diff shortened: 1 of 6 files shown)


lucacasonato:
@bartlomieju I think we sometimes land things with rebase commits in cases like this. As long as we don't use merge commits, I'm happy.

cjihrig:
@bartlomieju I added the issue number to the title of both commits. I'll land this once the CI passes again.

PR #14869, opened 2022-06-15
feat(ops): support `Result<impl Future<Output = Result<T, AnyError>> + 'static, AnyError>`

littledivy:
Support returning a `Result<impl Future<Output = Result<T, AnyError>> + 'static, AnyError>`. An hybrid op where both sync and async phases are fallible.

### DIFF
--- a/ops/lib.rs
+++ b/ops/lib.rs
@@ -73,7 +73,14 @@ pub fn op(attr: TokenStream, item: TokenStream) -> TokenStream {
   let MacroArgs { is_unstable, is_v8 } = margs;
   let func = syn::parse::<syn::ItemFn>(item).expect("expected a function");
   let name = &func.sig.ident;
-  let generics = &func.sig.generics;
+  let mut generics = func.sig.generics.clone();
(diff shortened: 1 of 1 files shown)


ops/lib.rs
```diff
@@ -398,6 +427,8 @@ fn is_handle_scope(arg: &syn::FnArg) -> bool {
     || tokens(arg).ends_with(": & mut v8 :: HandleScope < 'a >")
     || tokens(arg).ends_with(": & mut deno_core :: v8 :: HandleScope")
     || tokens(arg).ends_with(": & mut deno_core :: v8 :: HandleScope < 'a >")
+    || tokens(arg).contains("mut v8 :: HandleScope")
+    || tokens(arg).ends_with(": & mut v8 :: HandeScope < 'scope >")
```
nayeemrmn, line 431: This case and the first two are covered by the `.contains("mut v8 :: HandleScope")` above, but I don't know if that should be there.

Should we just use regex in this function for the lifetime parameter?

PR #14875, opened 2022-06-15
fix(check): Use "moduleDetection": "force"

### LINKED ISSUE #9593: dynamically imported module without import / export is treated as script by tsc
(Note: This probably doesn't affect any meaningful use cases, but I create issue for record.)

When we have the following scripts:

a.ts
```ts
const a = 1;
await import("./b.ts");
```

b.ts
```ts
const a = 2;
```

Then execute `a.ts` and it causes the following error.

```shellsession
$ deno run --allow-read a.ts
error: TS2451 [ERROR]: Cannot redeclare block-scoped variable 'a'.
const a = 2;
      ^
    at file:///Users/kt3k/0/di-test/b.ts:1:7

    'a' was also declared here.
    const a = 1;
          ^
        at file:///Users/kt3k/0/di-test/a.ts:1:7
```

This doesn't happen when:
- b.ts has any import or export statement
- import of b.ts is not dynamic import

---

```
$ deno --version
deno 1.7.5 (release, x86_64-apple-darwin)
v8 9.0.123
typescript 4.1.4
```

### LINKED ISSUE #13924: Determine correct `moduleDetection` compiler option (TypeScript 4.7)
Ref: https://github.com/microsoft/TypeScript/pull/47495

The next minor release of TypeScript (4.7) should include a new `moduleDetection` option, which has to deal with how modules are detected by TypeScript when type checking.  The default is `"auto"`, which might be ok for us, but also there is `"force"` which is more the behaviour that we would expect for code running under Deno.  We need to do a bit of experiment and testing.

The issue mostly manifested itself around issues in the lsp when having a file open with no top level import or export statement being treated a global script for type checking purposes.  There were also some issues with emitting `@jsxImportSource` with modules that appeared to be scripts.

nayeemrmn:
Fixes #9593.
Closes #13924.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#control-over-module-detection

### DIFF
--- a/cli/tests/testdata/lsp/code_action_response_no_disabled.json
+++ b/cli/tests/testdata/lsp/code_action_response_no_disabled.json
@@ -1,5 +1,25 @@
 [
   {
+    "title": "Extract to function in module scope",
+    "kind": "refactor.extract.function",
+    "isPreferred": false,
+    "data": {
+      "specifier": "file:///a/file.ts",
+      "range": {
+        "start": {
+          "line": 0,
+          "character": 0
+        },
+        "end": {
+          "line": 14,
+          "character": 0
+        }
+      },
+      "refactorName": "Extract Symbol",
+      "actionName": "function_scope_0"
+    }
+  },
+  {
     "title": "Move to a new file",
     "kind": "refactor.move.newFile",
     "isPreferred": false,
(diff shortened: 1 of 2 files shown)


dsherret (approved):
Nice! Thanks @nayeemrmn. I looked at the TS changes and this looks good for us to do.

PR #14882, opened 2022-06-16
fix(ext/fetch): add `accept-language` default header to fetch

### LINKED ISSUE #10987: Default `accept-language` header in `fetch`
The fetch spec mandates we add a `accept-language` header to all outgoing responses. Should we? If we do, should we choose the platform language (the one also used by default for Intl APIs), or should we hard-code it to english?

mrkldshv:
Add default `Accept-Language: *` header to fetch request.

Closes #10987.


satyarohith (approved):
LGTM. Thanks!

littledivy (approved):

Issue #14862, opened 2022-06-14
Ambiguous error with `Deno.bench`


danteissaias:
Steps to reproduce:
```ts
// file.ts
Deno.bench("console.log", () => {
    console.log("Hello, World!")
})
```

```
anon@mini:~/repos/fast$ deno run file.ts 
error: Uncaught TypeError: ops[opName] is not a function
Deno.bench({
     ^
    at Object.opSync (deno:core/01_core.js:172:38)
    at Object.bench (deno:runtime/js/40_testing.js:657:10)
    at file:///Users/anon/repos/fast/benchmarks.ts:11:6
```

jmj0502:
I would like to work on this issue. How can I get more details about it?

danteissaias:
Relevant discussion on discord: https://discord.com/channels/684898665143206084/684911491035430919/986936454225727519

PR #14883, opened 2022-06-16
fix(bench): don't throw on missing ops

danteissaias:


danteissaias:
Fixes https://github.com/denoland/deno/issues/14862

runtime/js/40_testing.js
```diff
@@ -749,15 +748,19 @@
...
-    benchDesc.filteredOut = filteredOut;
+      const { id, filteredOut } = core.opSync("op_register_bench", benchDesc);
+      benchDesc.id = id;
+      benchDesc.filteredOut = filteredOut;
 
-    ArrayPrototypePush(benchDescs, benchDesc);
+      ArrayPrototypePush(benchDescs, benchDesc);
+    } catch {
+      // Ignore missing ops
+    }
```
bartlomieju, line 763: This is brittle, it will ignore all errors that could happen during registration

bartlomieju (commented):
@danteissaias thank you for the PR, however I think this approach is not correct. I did open https://github.com/denoland/deno/pull/14946 with alternative approach that I think is simpler and more bullet proof

danteissaias:
> @danteissaias thank you for the PR, however I think this approach is not correct. I did open https://github.com/denoland/deno/pull/14946 with alternative approach that I think is simpler and more bullet proof

@bartlomieju I agree your approach is better however bench should be moved to the unstable ns and `op_bench_check_unstable` should be removed. Should I open a new PR with only these changes?

bartlomieju:
> > @danteissaias thank you for the PR, however I think this approach is not correct. I did open #14946 with alternative approach that I think is simpler and more bullet proof
> 
> @bartlomieju I agree your approach is better however bench should be moved to the unstable ns and `op_bench_check_unstable` should be removed. Should I open a new PR with only these changes?

I think that's not necessary - if we do that, then you will get errors saying that `Deno.bench` is undefined (unless you specify `--unstable` flag). I believe it's better to print error and exit when first bench is run. And for cases where you have `Deno.bench()` registered but use `deno run` would still work as expected. WDYT?

danteissaias:
> > > @danteissaias thank you for the PR, however I think this approach is not correct. I did open #14946 with alternative approach that I think is simpler and more bullet proof
> > 
> > 
> > @bartlomieju I agree your approach is better however bench should be moved to the unstable ns and `op_bench_check_unstable` should be removed. Should I open a new PR with only these changes?
> 
> I think that's not necessary - if we do that, then you will get errors saying that `Deno.bench` is undefined (unless you specify `--unstable` flag). I believe it's better to print error and exit when first bench is run. And for cases where you have `Deno.bench()` registered but use `deno run` would still work as expected. WDYT?

This isn’t consistent with other unstable APIs which are undefined without `--unstable`. I guess we could treat it as a special case but I don’t see much benefit.

bartlomieju:
> This isn’t consistent with other unstable APIs which are undefined without --unstable. I guess we could treat it as a special case but I don’t see much benefit.

That's a valid point, but I think it's fine band-aid solution, since we would like to stabilize `deno bench` soon anyway.

danteissaias:
> > This isn’t consistent with other unstable APIs which are undefined without --unstable. I guess we could treat it as a special case but I don’t see much benefit.
> 
> That's a valid point, but I think it's fine band-aid solution, since we would like to stabilize `deno bench` soon anyway.

Fair

PR #14890, opened 2022-06-16
fix(docs): update description of `--check` flag

### LINKED ISSUE #14889: CLI help information is outdated
Some of the information produced by `deno help` is not up to date. For example, here's a snippet of ouput from `deno help run` which says that type checking is still the default behavior for `deno run` in v`1.23.0`:

```
% deno --version
deno 1.23.0 (release, x86_64-apple-darwin)
v8 10.4.132.5
typescript 4.7.2

% deno help run
deno-run 
Run a JavaScript or TypeScript program

--- snip ---

OPTIONS:
--- snip ---

        --check[=<CHECK_TYPE>...]
            Type check modules.
            Currently this is a default behavior to type check modules, but in
            future releases
            Deno will not automatically type check without the --check flag.
            
            If the value of '--check=all' is supplied, diagnostic errors from
            remote modules
            will be included.

--- snip ---

```

I'm not sure what else is outdated, but a review and update is in order.

GJZwiers:
Fixes #14889. Updates help text for `--check` with the new behaviour.

Edit: I changed the spelling to type-check/type-checking which is consistent with what is used over at [typescript](https://www.typescriptlang.org/docs/handbook/2/basic-types.html#static-type-checking).

### DIFF
--- a/cli/flags.rs
+++ b/cli/flags.rs
@@ -234,12 +234,12 @@ impl Default for DenoSubcommand {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum TypeCheckMode {
-  /// Type check all modules.
+  /// type-check all modules.
   All,
-  /// Skip type checking of all modules. The default value for "deno run" and
+  /// Skip type-checking of all modules. The default value for "deno run" and
   /// several other subcommands.
   None,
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, thanks

PR #14892, opened 2022-06-16
refactor: make MainWorker::evaluate_module public

mlafeldt:
I noticed that it's not possible to pass any source code as a string to `execute_main_module`. That means I end up reimplementing the code inside it with slightly different params (passing code to `load_main_module` instead
of None). That's mostly trivial except for `evaluate_module`, which is currently private. Making it public allows to load modules from strings without copying any code.

Alternatively, I might add a param to `execute_main_module` and `preload_module` but that would break backward compat, which is something I wanted to avoid.

Happy to add a test if you think this change is acceptable.

### DIFF
--- a/runtime/worker.rs
+++ b/runtime/worker.rs
@@ -236,7 +236,11 @@ impl MainWorker {
     }
   }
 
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

PR #14896, opened 2022-06-17
add missing `use serde_v8;` (one line change)

jeff-hykin:
```
40 |             let deserialized_value = serde_v8::from_v8::<serde_json::Value>(scope, local);
    |                                      ^^^^^^^^ use of undeclared crate or module `serde_v8`
```

### DIFF
--- a/core/examples/eval_js_value.rs
+++ b/core/examples/eval_js_value.rs
@@ -9,6 +9,7 @@
 use deno_core::v8;
 use deno_core::JsRuntime;
 use deno_core::RuntimeOptions;
+use serde_v8;
 
 fn main() {
   let mut runtime = JsRuntime::new(RuntimeOptions::default());


littledivy (commented):
`use <crate>;` is redundant syntax. The error message means you don't have serde_v8 dependency in your Cargo.toml.

bartlomieju:
@jeff-hykin when are you getting this error? The CI passes fine for all examples, so as pointed out by @littledivy this change is not necessary.

jeff-hykin:
In a `main.rs` where I create a standalone Deno core runtine, thats the error I get. And that's pretty much the entire file (I based it on this example). I believe I'm using rust 1.63 nightly 

I don't understand how it doesn't give an error in the CI though haha. I mean `serde_v8` not defined anywhere in the file, where does it come from?

jeff-hykin:
In a `main.rs` where I create a standalone Deno core runtine. It's pretty empty aside from what is in the example. I was missing serde_v8 at first but even after using `cargo add` which should've edited the cargo.toml I got that error.

I didn't realize rust would import things automatically, crates really don't have to be explicitly imported?

I believe I'm using rust 1.63 nightly

bartlomieju:
> I didn't realize rust would import things automatically, crates really don't have to be explicitly imported?

Yes, if you have a crate added in your `Cargo.toml` you don't need to explicitly import it if you use fully qualified specifiers (eg. `serde_v8::to_v8`).

Can you provide your full code?

jeff-hykin:
Well I went through git history to find which commit, and none of them reproduce the error.

I'm just going to assume my memory was wrong and I got that error before installing serde instead of after. The example code runs for me once serde_v8 is installed.

Thanks for taking it seriously!

PR #14897, opened 2022-06-17
docs(lsp): correct header name in comments

ascorbic:
The comments refer to the header `x-deno-warn`, but [the actual header name](https://github.com/denoland/deno/blob/53dac7451bbdd527aa91e01653b678547624fc39/cli/lsp/cache.rs#L142) is `x-deno-warning`.


dsherret (approved):
LGTM. Thanks!

PR #14907, opened 2022-06-18
fix(fmt): should fail `--check` on parse error

dsherret:
https://discord.com/channels/684898665143206084/832257813492269088/987705741429342218


lucacasonato (approved):
LGTM

PR #14914, opened 2022-06-20
[WIP] feat(unstable): add Deno.openPty and friends

imjamesb:
Original PR: #14535, github repo got deleted so I had to reopen.


cli/dts/lib.deno.unstable.d.ts
```diff
@@ -203,10 +203,67 @@ declare namespace Deno {
...
+   * ```
+   */
+  export function openPty(options: ConsoleSize): Pty;
+
+  /** A Pty fork with readable and writable streams. */
+  export interface Pty {
+    rid: number;
+    readable: ReadableStream<Uint8Array>;
+    writable: WritableStream<Uint8Array>;
+  }
```
nayeemrmn, line 260: Add a `close()` method like other resources have
```suggestion
  export class Pty {
    readonly rid: number;
    readonly readable: ReadableStream<Uint8Array>;
    readonly writable: WritableStream<Uint8Array>;
    close(): void;
  }
```

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -203,10 +203,67 @@ declare namespace Deno {
...
+   *
+   * await Deno.spawn(Deno.execPath(), {
+   *   args: [
+   *     "eval",
+   *     `
+   *       const size = Deno.consoleSize(Deno.stdout.rid);
+   *       console.log("%sx%s", size.columns, size.rows);
+   *     `
+   *   ],
+   *   tty: pty.rid,
```
nayeemrmn, line 244: Accept the `Pty` object directly, unwrap the rid when calling the op
```suggestion
   *   tty: pty,
```

nayeemrmn (commented):
Deno's later APIs, while not hiding the `rid`, are designed to be usable without referencing them. See `Deno.HttpClient`:
```ts
const client = Deno.createHttpClient({ caCerts: [ caCert ] });
const req = await fetch("https://myserver.com", { client });
client.close();
```

cli/dts/lib.deno.unstable.d.ts
crowlKats: @nayeemrmn not all resources have that, ie the new subprocess API does not have a close method. also on that note, the `rid` should not be exposed either.

nayeemrmn: Spawn has alternate ways of closing / being waited on, things that are closed primarily by `Deno.close(res.rid)` should have a wrapper method.

Hiding `rid`s (even if they're not used for normal API usage) might be a problem for metrics and resource leakage tracking, but I'm not sure if that's practically a problem especially with our better sanitizer messages. Sure, I agree with hiding `rid` in a weak map for now.

crowlKats: > I agree with hiding `rid` in a weak map for now.

instead of a weakmap, it should be a symbol (as we do with web apis), for consistency.

overall, from what i remember discussing a few months back, we would want to move away from allowing manual resource management (so closing or resource and other usages of resource ids)

imjamesb: @nayeemrmn I suppose that this method will use `core.tryClose()` under the hood?

imjamesb:
So I think the PR should be complete Unix wise, maybe some tests is needed first. I'll start getting work done on the Windows features ASAP.

imjamesb:
I got my network card working, just doing some research on [conpty](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/), just need to learn the infrastructure around it for Windows.

imjamesb:
For reference:

image
image
image
image

https://discord.com/channels/684898665143206084/688040863313428503/1009118094632951879

imjamesb:
Another thing that should be taken into consideration if PTY's are moved into spawn API is that consoleSize() should get the PTY console size for the attached process, which will cause issues if implemented incorrectly on Linux, but will work ok on Windows. And then consoleResize which should resize the PTY, so in this case the operation should be applied to the PTY and not the process. I think that some changes are required in jsland to support those new features.

imjamesb:
Just because I want this so bad I will take another shot at implementing this into the spawn API and also introduce a new tty function, `Deno.consoleResize()`.

PR #14915, opened 2022-06-20
BREAKING(ext/ffi): Remove `Deno.UnsafePointer` indirection

littledivy:
This commits removes the `Deno.UnsafePointer` 
wrapper around FFI pointers. A pointer is now 
represented as a bigint primitive. 

There still exist the `Deno.UnsafePointer.of`
API to get pointer value for a TypedArray and 
`Deno.UnsafeCallback`.

```ts
const ptr = Deno.UnsafePointer.of(buf);
```

This commit also removes the extra processing for
FFI parameters. Hence, `Deno.UnsafeCallback` must be 
passed as a pointer.

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -397,10 +397,7 @@ declare namespace Deno {
     : T extends StaticNativeNumberType | StaticNativeBigIntType
       ? number | bigint
     : T extends "pointer" ? bigint | TypedArray | null
-    : T extends NativeCallbackType<
-      infer U extends readonly NativeType[],
-      infer V extends NativeParameterType
-    > ? UnsafeCallback<U, V> | bigint | null
+    : T extends NativeCallbackType ? bigint | null
     : unknown;
 
   /** Infers a foreign function parameter list. */
@@ -551,10 +548,7 @@ declare namespace Deno {
     : T extends StaticNativeBigIntType ? number | bigint
     : T extends StaticNativeNumberType ? number
     : T extends "pointer" ? bigint | TypedArray | null
-    : T extends NativeCallbackType<
-      infer U extends readonly NativeType[],
-      infer V extends NativeParameterType
-    > ? UnsafeCallback<U, V> | bigint | null
+    : T extends NativeCallbackType ? bigint | null
     : never;
 
   type UnsafeCallbackFunction<
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -390,17 +390,17 @@ declare namespace Deno {
     ? void
     : T extends StaticNativeBigIntType ? bigint
     : T extends StaticNativeNumberType ? number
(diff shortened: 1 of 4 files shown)


ext/ffi/00_ffi.js
```diff
@@ -120,141 +115,28 @@
...
-          ArrayPrototypePush(parameters, arg.value);
-        } else {
-          throw new TypeError(
-            "Expected FFI argument to be UnsafeCallback, UnsafeFnPointer, UnsafePointer or null",
-          );
-        }
-      } else {
-        throw new TypeError(`Invalid FFI argument type '${type}'`);
+    static of(value) {
+      if (ObjectPrototypeIsPrototypeOf(UnsafeCallbackPrototype, value)) {
```
aapoalas, line 127: Supporting UnsafeCallbacks for `UnsafePointer.of` seems unnecessary.

ext/ffi/00_ffi.js
```diff
@@ -489,22 +290,13 @@
             return promise;
           };
         } else {
-          const mustWrap = isPointerType(resultType);
-          fn = (...args) => {
-            const parameters = prepareArgs(types, args);
-
-            const result = core.opSync(
+          fn = (...parameters) => {
```
aapoalas, line 293: Unnecessary block around direct return.

aapoalas:
LGTM. Did you check performance improvement?

ext/ffi/00_ffi.js
littledivy: We will get rid of the UnsafePointer class completely. I was thinking that we could turn this into a more common interface until it's removed.

bartlomieju (approved):
LGTM, please update the manual before landing the PR

aapoalas:
Oh yeah, one thing to note publicly in the PR on this implementation is that after this there will be no difference in typing for:
* uint64
* int64
* usize
* isize
* pointer
* function
with the exception that pointer allows for TypedArrays in addition to bigints, and that number type arguments also allow for plain numbers instead of only bigints.

So, a lot of the possible benefits of types are lost, unfortunately.

littledivy:
yeah, I think the most notable will be for callbacks. 64bit numeric types + pointers are bigint in JS.

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -390,17 +390,17 @@ declare namespace Deno {
...
   type StaticForeignFunctionParameter<T> = T extends "void" ? void
     : T extends StaticNativeNumberType | StaticNativeBigIntType
       ? number | bigint
-    : T extends "pointer" ? UnsafePointer | TypedArray | null
+    : T extends "pointer" ? bigint | TypedArray | null
     : T extends NativeCallbackType<
       infer U extends readonly NativeType[],
       infer V extends NativeParameterType
-    > ? UnsafeCallback<U, V> | UnsafePointer | null
+    > ? UnsafeCallback<U, V> | bigint | null
```
aapoalas, line 403: With the removal of the wrapping, passing `UnsafeCallback<U, V>` as a parameter will not work.

PR #14917, opened 2022-06-20
fix: do not panic running .d.cts and .d.mts files

dsherret:
There was a match clause that was not correct, probably due to these being added and this code previously not throwing an error, so moved away from `_ =>` matching here too.


bartlomieju (approved):
LGTM

PR #14919, opened 2022-06-20
BREAKING(unstable): Improve Deno.spawn() stdio API

### LINKED ISSUE #14918: Proposed changes to `Deno.spawn()` API
- `SpawnOutput` should extend `ChildStatus` instead of composing it, discussed favourably on Discord a while ago.
- `SpawnOutput::stdout`, `SpawnOutput::stderr`, `Child::stdin`, `Child::stdout` and `Child::stderr` shouldn't be optional, instead make them getters that throw at runtime if that stream wasn't set to `"piped"`. Remove the complicated `<T extends SpawnOptions = SpawnOptions>` which we currently need to give proper type hints for the availability of these fields. Their typings for these would get increasingly complex if it became dependent on more options (e.g. `SpawnOptions::pty` which if set should make the stdio streams unavailable), so we should make them always present.

nayeemrmn:
~~Blocked on #15098.~~
Blocked on `test_util/std` update.

Closes #14918.


nayeemrmn:
Not sure how we can make changes to unstable APIs with these dependency loop-the-loops between cli and std workflows...

nayeemrmn:
We're running benchmarks on `test_util/std/examples/chat/server_test.ts` which uses the current spawn API. I think we could temporarily revert that to `Deno.run()` and bump the submodule, I need to check if that's the only one.

kt3k (approved):
I'm in favor of this design. LGTM

cc @crowlKats What do you think?

bartlomieju (changes requested):
I'm marking this as "Request changes" because I'd like to discuss it during CLI design meeting before landing. There are some use cases (like "unrefing" of child processes that need to be taken into account with such changes).

CC @cjihrig can you take a look as well?

.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
 
       - name: Install Deno
-        if: matrix.job == 'lint'
+        if: matrix.job == 'lint' || matrix.job == 'test'
```
bartlomieju, line 90: Could changes to this file be done in a separate PR?

bartlomieju (commented):
Why upgrade `test_util/std/` submodule? Could it also be done in a separate PR?

.github/workflows/ci.yml
nayeemrmn: #15098

bartlomieju (approved):
LGTM, thank you @nayeemrmn

PR #14920, opened 2022-06-20
feat(ext/ffi): BREAKING(unstable): Simplify FFI types

aapoalas:
This PR simplifies the TypeScript types used for interacting with Deno FFI. The basis is that types are now first grouped into logical wholes, `NativeNumberType`, `NativeBigIntType` etc. These wholes are combined into the `NativeType` and `NativeResultType` general types.

Then, for both "To" and "From" direction (viewed from Deno's point of view), mapper types are set up to convert the type declaration to the actual TypeScript type.

So, as an example for the type declaration `"usize"` the `ToNativeType` result is `number | bigint`, as we'll allow either a number of BigInt to be passed as a usize into FFI. The `FromNativeType` result however is `bigint`, since FFI will only ever return `usize` type values as BigInts from FFI.

**BREAKING:** Additionally, this PR removes the `{ function: { parameters: [], result: "void" } }` type declaration from parameters (and result types) as PR #14915 made the type information gained from these declarations unusable. Now functions are merely passed and returned as `"function"`. For `FromNativeType` types there is no difference between `"pointer"` and `"function"`, both will always be given to users as BigInts. However, on the `ToNativeType` side there is a difference as `"pointer"` parameters and callback return values can be given TypedArrays whereas `"function"`s cannot accept those, and passing those would not work (luckily, as this would be a MASSIVE safety risk).

### DIFF
--- a/cli/dts/lib.deno.unstable.d.ts
+++ b/cli/dts/lib.deno.unstable.d.ts
@@ -360,7 +360,7 @@ declare namespace Deno {
     | NativePointerType
     | NativeFunctionType;
 
-  type NativeResultType = NativeType | NativeVoidType;
+  export type NativeResultType = NativeType | NativeVoidType;
 
   /** Type conversion for foreign symbol parameters and unsafe callback return types */
   type ToNativeType<T extends NativeType = NativeType> = T extends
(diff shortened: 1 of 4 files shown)


littledivy (approved):
LGTM, the new "function" type isn't a breaking change, ffi callbacks aren't part of a release yet.

aapoalas:
Oh yeah, good point :D

PR #14924, opened 2022-06-20
chore(ext/ffi): Improve FFI static handling

aapoalas:
Minor improvement to FFI static symbols. Mostly just code cleanup.

### DIFF
--- a/ext/ffi/lib.rs
+++ b/ext/ffi/lib.rs
@@ -1203,37 +1203,36 @@ fn op_ffi_get_static<'scope>(
     NativeType::U8 => {
       let result = unsafe { ptr::read_unaligned(data_ptr as *const u8) };
       let number: v8::Local<v8::Value> =
-        v8::Number::new(scope, result as f64).into();
+        v8::Integer::new_from_unsigned(scope, result as u32).into();
       number.into()
     }
     NativeType::I8 => {
       let result = unsafe { ptr::read_unaligned(data_ptr as *const i8) };
       let number: v8::Local<v8::Value> =
-        v8::Number::new(scope, result as f64).into();
+        v8::Integer::new(scope, result as i32).into();
       number.into()
     }
     NativeType::U16 => {
       let result = unsafe { ptr::read_unaligned(data_ptr as *const u16) };
       let number: v8::Local<v8::Value> =
-        v8::Number::new(scope, result as f64).into();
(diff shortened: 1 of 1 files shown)


ext/ffi/lib.rs
```diff
@@ -1179,76 +1179,89 @@ fn op_ffi_get_static<'scope>(
...
-      serde_v8::from_v8(scope, number.into())?
+      let number: v8::Local<v8::Value> =
+        v8::Number::new(scope, result as f64).into();
+      number.into()
     }
     NativeType::U64 => {
       let result = unsafe { ptr::read_unaligned(data_ptr as *const u64) };
-      let big_int = v8::BigInt::new_from_u64(scope, result);
-      serde_v8::from_v8(scope, big_int.into())?
+      let big_int: v8::Local<v8::Value> =
```
bartlomieju, line 1226: These type declarations seem to be superfluous, any chance we could remove them?

aapoalas: They're there to facilitate creating the BigInt and immediately calling `.into()`to get the `Local<Value>`, which is in turn needed to call `.into()` for the actual return type `serde_v8::Value`.

ext/ffi/lib.rs
```diff
@@ -1194,76 +1194,89 @@ fn op_ffi_get_static<'scope>(
...
   Ok(match static_type {
     NativeType::Void => {
       return Err(type_error("Invalid FFI static type 'void'"));
     }
     NativeType::U8 => {
       let result = unsafe { ptr::read_unaligned(data_ptr as *const u8) };
-      let number = v8::Number::new(scope, result as f64);
-      serde_v8::from_v8(scope, number.into())?
+      let number: v8::Local<v8::Value> =
+        v8::Number::new(scope, result as f64).into();
```
piscisaureus, line 1206: Why not use v8::Integer::new() ?

Converting to float and then back to int has a lot of overhead.

aapoalas: Ah, this is a copy-paste error indeed. I think this was what I was originally supposed to fix in this PR (I had a note to self to give statics some love) but forgot all about it.

aapoalas: Massive thanks for noticing this! Leaving this in would've been the height of idiocy and I was completely blind to it.

littledivy (approved):
LGTM

PR #14925, opened 2022-06-20
refactor: add `EmitCache` trait

dsherret:
This PR adds a new `EmitCache` trait to move towards an sqlite database as part of #13302.

### DIFF
--- a/cli/cache.rs
+++ b/cli/cache.rs
@@ -8,7 +8,6 @@ use deno_core::error::AnyError;
 use deno_core::futures::FutureExt;
 use deno_core::serde::Deserialize;
 use deno_core::serde::Serialize;
-use deno_core::serde_json;
 use deno_core::ModuleSpecifier;
 use deno_graph::source::CacheInfo;
 use deno_graph::source::LoadFuture;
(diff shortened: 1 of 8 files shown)


cli/cache.rs
```diff
@@ -81,30 +71,6 @@ impl FetchCacher {
...
-  fn get_emit_metadata(
-    &self,
-    specifier: &ModuleSpecifier,
-  ) -> Option<EmitMetadata> {
-    let filename = self
-      .disk_cache
-      .get_cache_filename_with_extension(specifier, "meta")?;
-    let bytes = self.disk_cache.get(&filename).ok()?;
-    serde_json::from_slice(&bytes).ok()
-  }
```
dsherret, line 94: All of this is moved to disk_cache as it wasn't necessary here.

cli/emit.rs
```diff
@@ -453,6 +564,13 @@ pub fn check_and_maybe_emit(
         }
       }
     }
+
+    // now insert these items into the cache
+    for (specifier, data) in emit_data_items.into_iter() {
+      if let Some(cache_data) = data.into_cache_data() {
+        cache.set_emit_data(specifier, cache_data)?;
```
dsherret, line 571: In the future this will be a single sqlite db update where it updates a row all at once, so that's why the code was changed to have a single update method.

cli/tools/test.rs
```diff
@@ -1453,7 +1452,7 @@ pub async fn run_tests_with_watch(
           .collect(),
         false,
         maybe_imports,
-        cache.as_mut_loader(),
```
bartlomieju, line 1456: I really didn't like this `as_mut_loader` method, glad to see it gone.

bartlomieju (approved):
LGTM, looking forward to a follow up PR

dsherret:
@bartlomieju I'll do a follow up PR closer to the next minor release (and have the deno_cache changes ready at the same time).

PR #14927, opened 2022-06-21
chore: fix pty_tab_handler test on windows

dsherret:
This fixes the test, but adding a tab is a little broken on Windows (the output gets offset). I think it's a rustyline bug... doesn't seem worth tracking though until someone complains because it's cosmetic I think.

### DIFF
--- a/cli/tools/repl/editor.rs
+++ b/cli/tools/repl/editor.rs
@@ -411,7 +411,10 @@ impl ConditionalEventHandler for TabEventHandler {
     _: bool,
     ctx: &EventContext,
   ) -> Option<Cmd> {
-    debug_assert_eq!(*evt, Event::from(KeyEvent(KeyCode::Tab, Modifiers::NONE)));
+    debug_assert_eq!(
+      *evt,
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):

sigmaSd:
Do you mean that tabbing doesn't work on windows correctly (I will try to get a windows machine to test myself)?  if so what about in windows we replace tab with 4 spaces?

dsherret:
> Do you mean that tabbing doesn't work on windows correctly (I will try to get a windows machine to test myself)? if so what about in windows we replace tab with 4 spaces?

@sigmaSd that might be a good idea. Do you want to try it out on your windows machine to see if there are any issues and want to follow up with that pr? I will merge this for now because it's causing the tests to hang on Windows locally (this test does not run on the CI).

sigmaSd:
@dsherret I can't send a pr because I can't compile deno on my windows machine but I tested on rustyline repo and using 4 spaces seems to work very well. I think its an ok compromise until its fixed in rustyline (I'll open a bug report).

PR #14928, opened 2022-06-21
chore(test_util): add new string assertion macros

dsherret:
Adds some string assertion macros so we can get more details about why a certain test fails.

### DIFF
--- a/cli/tests/integration/bundle_tests.rs
+++ b/cli/tests/integration/bundle_tests.rs
@@ -2,6 +2,8 @@
 
 use crate::itest;
 use test_util as util;
+use test_util::assert_contains;
+use test_util::assert_ends_with;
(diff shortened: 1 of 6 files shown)


cli/tests/integration/bundle_tests.rs
```diff
@@ -39,10 +41,10 @@ fn bundle_exports() {
     .output()
     .unwrap();
   // check the output of the test.ts program.
-  assert!(std::str::from_utf8(&output.stdout)
-    .unwrap()
-    .trim()
-    .ends_with("Hello"));
+  assert_ends_with!(
```
dsherret, line 44: Macros are used instead of functions because they say what line the test failed on.

bartlomieju (approved):

Issue #9585, opened 2021-02-23
Add --jobs flag to relevant subcommands


caspervonb:
Add a `-j, --jobs N`(most common name in the wild) flag to `deno lint` and `deno fmt` to let users control how many concurrent threads it will consume.

Might be applicable for other subcommands, like deno test soon'ish.
Not sure if we should apply this to the runtime commands like run, eval, etc.

caspervonb:
Not stale, deno test has this in a pending PR and got one for lint locally, will raise that soon.

lucacasonato:
I don't get why we should have this for `deno lint` or `deno fmt`. They should have good defaults, and not have any configuration for this.

@dsherret WDYT?

caspervonb:
Going back a bit here, but iirc lint spins wild with threads and used to die on my Mac mini a lot.

dsherret:
I don’t think it should be a flag. If anything just an environment variable that users can set for their system like what rust does.

mrkldshv:
@lucacasonato @bartlomieju WDYT about this?

bartlomieju:
I agree with David, there's no point adding `--jobs` to `fmt` and `lint` since they use multiple threads by default.

nayeemrmn:
I think
- Deprecate `--jobs` flag by printing a warning when it's used
- Add a `DENO_JOBS` env var to replace it
- `DENO_JOBS` default should be 1 for `deno test`
- `DENO_JOBS` default should be system cores for `deno fmt` and `deno lint` (separate PR from above)

This is good because `--jobs` syntax was a bit broken anyway: #14533.

caspervonb:
Can always do `DENO_TEST_THREADS` and less than zero as a condition to fallback to `num_cpus`

lucacasonato:
We are using the `DENO_JOBS` env var for this now, in combination with a `--parallel` flag.

Issue #14533, opened 2022-05-08
`deno test --jobs` flag should require equals when passing a value


nayeemrmn:
The `--jobs` flag optionally doesn't take a number value to default to the number of available cpus. So it should require an equal sign. Currently `deno test --jobs tests/` doesn't work because `tests/` is parsed as the number of threads. Unfortunately this is a breaking change, so this was a bad oversight.

dsherret:
When fixing this (aka, doing this breaking change) we should consider whether we should just deprecate it in favour of the DENO_JOBS env variable. See discussion in https://github.com/denoland/deno/pull/14929#discussion_r919079601

nayeemrmn:
Note that we also wanted _some_ binary flag to activate `DENO_JOBS` so that the test runner remained single-threaded by default. Other than retracting that expectation, a sound option would be to still deprecate `--jobs` and replace it with `--parallel` which does the same thing. It's a slightly better name for a binary flag and wouldn't have this bug.

dsherret:
I think replacing it with `--parallel` makes sense given that fixing `--jobs` is a breaking change and it's not a great name. Also, `"parallel": true` in the configuration file might be a nice option in the future.

PR #14929, opened 2022-06-21
feat(cli/test): add `DENO_JOBS` env variable for `test` subcommand

mrkldshv:
All changes in this MR concern `test` subcommand. The implemented behaviour for running tests is: 
* Use one thread by default
* Deprecate usage of `--jobs` flag with numeric values (e.g. `--jobs=2`)
* Introduce `DENO_JOBS` env variable that is used with `--jobs` flag (usage: `DENO_JOBS=2 deno test --jobs`)
* Use number of available threads when `--jobs` flag is provided without value and there is no `DENO_JOBS` env variable set

Addresses part of https://github.com/denoland/deno/issues/9585#issuecomment-1160432543.


cli/flags.rs
```diff
@@ -2654,12 +2657,20 @@ fn test_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
...
+    );
     if let Some(value) = matches.value_of("jobs") {
       value.parse().unwrap()
     } else {
       std::thread::available_parallelism()
         .unwrap_or(NonZeroUsize::new(1).unwrap())
     }
+  } else if let Ok(value) = env::var("DENO_JOBS") {
+    value
+      .parse::<NonZeroUsize>()
```
mrkldshv, line 2672: Here we just ignore error when parsing value fails. Instead we can print error message asking to use valid value. What do you think? @dsherret @nayeemrmn

dsherret: Makes sense to me. Probably not worth any extra complexity as someone putting in a non-number value would be extremely rare.

cli/flags.rs
```diff
@@ -2654,12 +2657,20 @@ fn test_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
...
+      "{}",
+      crate::colors::yellow("Warning: --jobs flag is deprecated. Use 'DENO_JOBS' environment variable instead."),
+    );
     if let Some(value) = matches.value_of("jobs") {
       value.parse().unwrap()
     } else {
       std::thread::available_parallelism()
         .unwrap_or(NonZeroUsize::new(1).unwrap())
     }
+  } else if let Ok(value) = env::var("DENO_JOBS") {
```
dsherret, line 2670: I don't think we should do this because not every codebase can handle tests running at the same time. We should probably have a `--parallel` flag like @nayeemrmn suggested and DENO_JOBS would then only be used to inform the amount of parallelism. I will discuss it with some people first though to see if there's support for this.

nayeemrmn: Another reason to have a distinct `--parallel` flag is because we will likely have different result formatting in parallel mode, which suggests it should be API rather than an env config.

mrkldshv: Alright, so if I understand correctly:
- Deprecate `--jobs` flag
- Introduce `--parallel` flag and default number of used threads to number of available CPUs
- Use `DENO_JOBS` env variable to override number of used threads
- Default number of used threads to 1 if `--jobs` and `--parallel` flags are not passed

nayeemrmn: SGTM

dsherret: @mrkldshv we discussed this in our design meeting and we want to keep the jobs flag, but warn when someone provides a number and suggest they use DENO_JOBS env var with --jobs instead. So what you said, but no parallel flag. Also it should prefer the flag value over the env var.

mrkldshv: Thanks for the information. Keeping `--jobs` flag sounds good to me. I'll update this PR accordingly.

cli/args/flags.rs
```diff
@@ -1548,9 +1552,10 @@ fn test_subcommand<'a>() -> Command<'a> {
       Arg::new("jobs")
         .short('j')
         .long("jobs")
-        .help("Number of parallel workers, defaults to # of CPUs when no value is provided. Defaults to 1 when the option is not present.")
+        .help("Number of parallel workers, defaults to number of available CPUs when no value is provided. Defaults to 1 when the option is not present.")
         .min_values(0)
         .max_values(1)
+        .require_equals(true)
```
dsherret, line 1558: I think this is a breaking change, so we should leave it as-is and not require equals.

cli/args/flags.rs
```diff
@@ -2665,7 +2670,15 @@ fn test_parse(flags: &mut Flags, matches: &clap::ArgMatches) {
 
   let concurrent_jobs = if matches.is_present("jobs") {
     if let Some(value) = matches.value_of("jobs") {
+      println!(
+          "{}",
+          crate::colors::yellow("Warning: --jobs flag with numeric value is deprecated. Use 'DENO_JOBS' environment variable with --jobs flag instead."),
+        );
```
dsherret, line 2676: I'm surprised this passes the formatting check for `cargo fmt`. I guess because it's in a macro.

cli/tests/testdata/test/pass-jobs-flag.out
```diff
@@ -0,0 +1,24 @@
+Warning: --jobs flag is deprecated. Use 'DENO_JOBS' environment variable instead.
```
dsherret, line 1: I think this file is not used anymore?

dsherret (changes requested):
Overall, looks good to me, but just a couple small changes.

cli/args/flags.rs
dsherret, line 2676: One scenario I'm wondering about is when people really want to limit the parallelism to say `2` due to the tests using some common resource that needs to be limited. I think probably doing so is probably an anti-pattern because it affects all the tests and not just the ones that use the common resource. In this case, the resource usage should be limited by the tests themselves or the user should run the command in a deno task setting the DENO_JOBS env var first (ex. `DENO_JOBS=2 deno test`).

cli/args/flags.rs
mrkldshv: Yep, I've added this to resolve the issue that now we can specify `--jobs` without a value now. So, if you write `deno test --jobs test.ts`, `test.ts` will be a value for `--jobs` flag. Do you have an idea on how this should be handled?

cli/tests/testdata/test/pass-jobs-flag.out
mrkldshv: Yep, will remove it.

cli/args/flags.rs
mrkldshv: Do we want to make this breaking change and fix the issue, or keep it as it was allowing to provide value with the equals sign?

dsherret: Hmmm... that is really annoying. I'm thinking we should just remove the deprecation notice on lines 2673-2676 then deprecate it once resolving #14533. So we would only be adding the DENO_JOBS env var in this PR. Sound ok?

mrkldshv: Sure, sounds good to me. I'll apply change to this PR accordingly.

dsherret (approved):
LGTM. Thanks @mrkldshv!

PR #14942, opened 2022-06-22
feat(ext/ffi): Thread safe callbacks

aapoalas:
Fairly okay working solution to thread safe FFI callbacks, complete with tests!

A new `FfiState` is saved into the `OpState`, containing the sender and receiver for a `std::mpsc::sync_channel` which is used for message passing between threads. Isolate threads (main and worker) are now identified using a thread-local Isolate pointer which gets assigned to the thread on `UnsafeCallback` creation. This allows an incoming callback to not only figure out if the thread it is being called on is an isolate thread but to also tell apart its own isolate thread from others, thus making it possible for workers to register and trigger callbacks originating in any thread while still ensuring that those callbacks are called on the appropriate isolate thread.

If the callback is found to be coming from a foreign thread (foreign isolate thread or completely foreign), it will create a new response `SyncChannel` and sends a message using its own copy of the sender side of the main `SyncChannel`. The message is a boxed closure which will call the actual `deno_ffi_callback` internal implementation and then send a message into the newly created response `SyncChannel`. The foreign thread will then block on the response `SyncChannel` until it receives the response message after which it unblocks the thread and deallocates the response channel and returns control to the caller of the callback.

### Open questions:
 - [x] Channel type decided in the favour of an `UnboundedChannel`
 - [x] Isolate pointer comparison deemed apt, or at least not objected to
 - [x] Response `sync_channel` bound set to 0, making it a rendezvous channel
 - [ ] Error handling for message passing?


ext/ffi/lib.rs
```diff
@@ -842,21 +891,55 @@ unsafe extern "C" fn deno_ffi_callback(
...
+        result,
+        args,
+        info.callback,
+        info.context,
+        info.isolate,
       );
-      std::process::exit(1);
+    } else {
+      let async_work_sender = &info.async_work_sender;
+      // safeish, since we block until the message has been received and a response transmitted.
```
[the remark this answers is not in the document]
aapoalas: Yeah, very true :D

cli/dts/lib.deno.unstable.d.ts
```diff
@@ -554,6 +554,9 @@ declare namespace Deno {
    * as C function pointers to ffi calls.
    *
    * The function pointer remains valid until the `close()` method is called.
+   *
+   * The callback can be explicitly referenced and dereferenced to stop Deno's
```
bartlomieju, line 558: I believe `dereferenced` is misleading in this context

ext/ffi/lib.rs
```diff
@@ -204,10 +212,50 @@ pub fn init<P: FfiPermissions + 'static>(unstable: bool) -> Extension {
...
+      }
+
+      maybe_scheduling
+    })
     .state(move |state| {
       // Stolen from deno_webgpu, is there a better option?
       state.put(Unstable(unstable));
+
+      let (async_work_sender, async_work_receiver) =
+        mpsc::sync_channel::<PendingFfiAsyncWork>(100);
```
bartlomieju, line 251: Using a buffer of size 100 here seems rather arbitrary, what's the reasoning behind it?

ext/ffi/lib.rs
```diff
@@ -178,6 +178,14 @@ impl DynamicLibraryResource {
   }
 }
 
+type PendingFfiAsyncWork = Box<dyn FnOnce()>;
+
+struct FfiState {
+  async_work_sender: mpsc::SyncSender<PendingFfiAsyncWork>,
+  async_work_receiver: mpsc::Receiver<PendingFfiAsyncWork>,
+  active_refd_functions: usize,
```
bartlomieju, line 186:
```suggestion
  active_refed_functions: usize,
```

ext/ffi/lib.rs
```diff
@@ -1096,11 +1179,16 @@ where
   let cb = v8::Local::<v8::Function>::try_from(v8_value)?;
 
   let isolate: *mut v8::Isolate = &mut *scope as &mut v8::Isolate;
+  LOCAL_ISOLATE_POINTER.with(|s| s.replace(isolate));
```
bartlomieju, line 1182: Maybe add a check that we've replaced a null pointer here?

ext/ffi/lib.rs
aapoalas: It is an absolute "number out of the hat".

cli/dts/lib.deno.unstable.d.ts
aapoalas: Used `ref'ed` and `unref'ed` for now.

ext/ffi/lib.rs
aapoalas: Added a check to only replace the null pointer once.

ext/ffi/lib.rs
```diff
@@ -204,10 +212,50 @@ pub fn init<P: FfiPermissions + 'static>(unstable: bool) -> Extension {
...
+      }
+
+      maybe_scheduling
+    })
     .state(move |state| {
       // Stolen from deno_webgpu, is there a better option?
       state.put(Unstable(unstable));
+
+      let (async_work_sender, async_work_receiver) =
+        mpsc::sync_channel::<PendingFfiAsyncWork>(100);
```
littledivy, line 251: Use a `futures::channel::mpsc::UnboundedChannel` since there is no hard requirement that the async work sender should block.

ext/ffi/lib.rs
aapoalas: This is now a `futures::channel::unbounded`.

aapoalas:
One question that I thought of and tested that _locally_ this caused no issues is the response `sync_channel`. Currently I create it with a bound of 1, meaning that if somehow theoretically the isolate thread manages to call the `do_ffi_callback` and then send a response before the "async" thread calls `response_receiver.recv().unwrap()` then the response can be buffered.

This seems unnecessary, as the `sync_channel` receiver side is literally this:
```rs
async_work_sender.unbounded_send(fut).unwrap(); // Send the message
response_receiver.recv().unwrap(); // Block on a response
```
The isolate thread would need to be lightning fast to poll the async work channel and do the work, and the calling thread would need to be absolutely massively busy to not get to call `.recv()` before the response sent. This does not seem to be a particularly realistic case.

As such the bound could perhaps be lowered to 0, making the channel a "rendezvous channel" which brings with it some perf optimisations I believe.

Opinions?

littledivy (approved):
LGTM. This introduces an event loop middleware - let's watch for any unrelated perf regressions.

aapoalas:
> LGTM. This introduces an event loop middleware - let's watch for any unrelated perf regressions.

Good point. The vector allocation might cause perf regression.

PR #14945, opened 2022-06-23
perf(ext/ffi): optimize synchronous calls 

littledivy:
This commit implements an optimization
for blocking FFI calls. 

`op_ffi_load` creates JS functions for 
synchronous symbols. This avoids 
reallocations that existed in `op_ffi_call`
and allows future V8 fast API optimizations.

Each function owns an allocation to corresponding
the Symbol, which is freed on GC.

Nonblocking ops & statics are still registered JS
side and use ops to make calls.

image


bartlomieju (approved):
LGTM, these are very impressive results

lucacasonato:
I want to discuss this in CLI design meeting, because it bypasses ops.

lucacasonato:
Also unintentional SWC downgrade?

ext/ffi/lib.rs
```diff
@@ -541,24 +501,111 @@ where
...
+  let func = v8::Function::builder(
+    |scope: &mut v8::HandleScope,
+     args: v8::FunctionCallbackArguments,
+     mut rv: v8::ReturnValue| {
+      let external: v8::Local<v8::External> =
+        args.data().unwrap().try_into().unwrap();
+      // SAFETY: The pointer will not be deallocated until the function is
+      // garbage collected.
+      let symbol = unsafe { &*(external.value() as *const Symbol) };
+      let result = ffi_call_sync(scope, args, symbol).unwrap();
```
aapoalas, line 588: Should handle errors.

bartlomieju:
> Also unintentional SWC downgrade?

How so?

ext/ffi/lib.rs
bartlomieju: Good point, I've missed that

bartlomieju (changes requested):
https://github.com/denoland/deno/pull/14945#discussion_r904866038

lucacasonato:
> How so?

Look at cli/Cargo.toml

bartlomieju:
> > How so?
> 
> Look at cli/Cargo.toml

Screenshot 2022-06-23 at 12 42 49

I see only two changed files

lucacasonato:
Huh, on the GitHub mobile app I see a bunch more changed files. I guess must be a bug. It looks correct on the website... sorry for the noise.

bartlomieju (approved):

PR #14946, opened 2022-06-23
fix: don't error if Deno.bench() or Deno.test() are used in run subcommand

bartlomieju:
Fixes https://github.com/denoland/deno/issues/14862

### DIFF
--- a/cli/tests/integration/run_tests.rs
+++ b/cli/tests/integration/run_tests.rs
@@ -2728,3 +2728,8 @@ fn running_declaration_files() {
     assert!(output.status.success());
   }
 }
+
+itest!(test_and_bench_are_noops_in_run {
+  args: "run test_and_bench_in_run.js",
+  output_str: Some(""),
+});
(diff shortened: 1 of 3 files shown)


dsherret (approved):
LGTM.

I kind of wonder if it would have been good to throw with Deno.test originally in order to guide people to using the `deno test` sub command in case they use `deno run`, but it's probably pretty obvious when someone accidentally does that. Maybe not perfect for beginners, but not a big deal.

bartlomieju:
> LGTM.
> 
> I kind of wonder if it would have been good to throw with Deno.test originally in order to guide people to using the `deno test` sub command in case they use `deno run`, but it's probably pretty obvious when someone accidentally does that. Maybe not perfect for beginners, but not a big deal.

Some people want to have testing code in the same file as the actual code. I think it makes sense for some one-off scripts or CLI tools.

nayeemrmn:
An error can still be seen for `Deno.test()` in `deno bench` and vice versa, since the switch is unified here.

